PR #3645, opened 2019-10-26
Retrieve port-id in LLDP tab

### LINKED ISSUE #3629: Display port-id in LLDP panel
### Environment
* Python version:  3.7.3
* NetBox version:  2.5.11

### Proposed Functionality
Currently, when displaying the LLDP tab on a JunOS device, the "port description" is displayed if present in the equipment.
image

The reason for that is that Napalm used function is "get_lldp_neighbors". This function returns "port description" in the "port" field if available. If not it returns "port-id".
This does not seem the best behaviour as the LLDP panel with its color code (white, blue, green and red) is used to compare Netbox declared connectivity to real (LLDP) connectivity.

The proposed enhancement consists in using "get_lldp_neighbors_detail" Napalm function. The use of this function allows to explicitly retrieve LLDP "port-id".
image

### Use Case
Using "port-id" instead of "port description" allows to easily check connectivity consistency.

### Database Changes
None

### External Dependencies
None

BegBlev:
### Fixes: #3629 
This fix allows to retrieve, LLDP port-id from NAPALM driver instead of port description.

### DIFF
--- a/netbox/templates/dcim/device_lldp_neighbors.html
+++ b/netbox/templates/dcim/device_lldp_neighbors.html
@@ -52,10 +52,10 @@
 <script type="text/javascript">
 $(document).ready(function() {
     $.ajax({
-        url: "{% url 'dcim-api:device-napalm' pk=device.pk %}?method=get_lldp_neighbors",
+        url: "{% url 'dcim-api:device-napalm' pk=device.pk %}?method=get_lldp_neighbors_detail",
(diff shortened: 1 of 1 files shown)


kobayashi (commented):
LGTM. I will test myself and merge later

DanSheps (approved):
Just checked it myself with a Cisco device, looks good to me as well.

Issue #3538, opened 2019-09-24
Add REST API endpoint for executing custom scripts


jeremystretch:
### Environment
* Python version: 3.6.8
* NetBox version: 2.6.4

### Proposed Functionality
Add a REST API endpoint for executing custom scripts, similar to the endpoint which exists for executing validation reports (`/api/extras/reports/`).

### Use Case
This will allow custom scripts to be triggered by remote systems in addition to web UI users.

### Database Changes
None

### External Dependencies
None

chaomodus:
One minor request here, if there were a way to pass some custom data not as a log message back, such as a dictionary that, in the API form, was converted to part of the JSON response, but pretty printed as part of the result in the web UI response.

PR #3654, opened 2019-10-29
3538: Add custom script API endpoints

jeremystretch:
### Fixes: #3538

Introduces REST API endpoints to list, retrieve, and run custom scripts. Includes tests for retrieving and running a script via the API.


jeremystretch:
Still need to fix a bug where scripts within a module which has a custom name defined are not found.

Issue #1814, opened 2018-01-09
Ability to use object store for images


jpds:
### Issue type
[x] Feature request 
[ ] Bug report      
[ ] Documentation   

### Environment
* NetBox version: v2.2.9

### Description

I'm running Netbox in a Kubernetes cluster. It currently runs as a single container as I have a block device mounted into the container so that there's persistent storage for the image attachments which were introduced in #152.

This works OK, but I cannot run multiple instances of netbox in a clustered fashion as it's tied to the block device.

Could there be some integration with https://github.com/jschneier/django-storages so that S3 can be used as a backend for the image attachments?

jeremystretch:
Per the discussion in PR #2466, the configuration scheme for this needs to be fleshed out. While I don't have a strong preference, whatever solution we decide on must be platform- and vendor-agnostic, and easily extendable to accommodate similar services.

nward:
Hi, I've been directed here from #3095 - thanks for that @jeremystretch.

I have implemented something as a PoC for this.
Additionally, I implemented a way to optionally load another application for display. "django-storages" doesn't require this, as it directs the user for the application to a direct download URL. This has the effect that images are not behind the NetBox access control anymore. Writing an application for this would allow an administrator to use S3/etc. for only storage, not for serving the files.

For the configuration, I toyed with several approaches:

1. Have configuration for all the various options (S3_* etc.). This doesn't work, as in NetBox we have only specific settings passed through from configuration.py to settings.py, and we would need settings.py to support every possible option - doesn't really fly.
2. Have a dictionary defined in configuration.py which the "external" applications, storage classes etc. can use for settings. This requires that these applications, storage classes etc. are written custom for NetBox (or at the very least, with NetBox support). This is not untenable as potentially some sort of wrapper classes could be written but it's not ideal.
3. Have a dictionary defined in configuration.py which NetBox's settings.py takes and translates in to settings.py parameters.
4. Have settings in configuration.py which are found by settings.py with a prefix, and then the prefix stripped off and defined as settings.py parameters. This is OK, but a bit clunky and hard to read.

In my implementation, I opted for (3) which I found this to be a good balance between (1) and (2). It has some things I'd like to improve. Primarily reduce or eliminate the ability for a configuration parameter to be named the same as an existing NetBox parameter and override it - perhaps it needs to scan a list of known settings used by NetBox and error if the administrator attempts to pass them in this dictionary. That seems fairly easy to implement.

[...]

PR #3666, opened 2019-11-03
Add support for S3 storage for media

steffann:
### Fixes: #1814

I implemented the following configuration structure:

```
MEDIA_STORAGE = {
  # Required configuration
  'BACKEND': 'S3',
  'ACCESS_KEY_ID': 'Key ID',
  'SECRET_ACCESS_KEY': 'Secret',
  'BUCKET_NAME': 'netbox',

  # Optional configuration, defaults are shown
  'REGION_NAME': '',
  'ENDPOINT_URL': None,
  'AUTO_CREATE_BUCKET': False,
  'BUCKET_ACL': 'public-read',
  'DEFAULT_ACL': 'public-read',
  'OBJECT_PARAMETERS': {
    'CacheControl': 'max-age=86400',
    'CacheControl': 'max-age=86400',
  },
  'QUERYSTRING_AUTH': True,
  'QUERYSTRING_EXPIRE': 3600,
}
```

Feedback appreciated.

### DIFF
--- a/netbox/extras/models.py
+++ b/netbox/extras/models.py
@@ -611,11 +611,18 @@ class ImageAttachment(models.Model):
     @property
     def size(self):
         """
-        Wrapper around `image.size` to suppress an OSError in case the file is inaccessible.
+        Wrapper around `image.size` to suppress an OSError in case the file is inaccessible. When S3 storage is used
+        ClientError is suppressed instead.
         """
+        from django.conf import settings
+        if settings.MEDIA_STORAGE and settings.MEDIA_STORAGE['BACKEND'] == 'S3':
+            from botocore.exceptions import ClientError as AccessError
+        else:
+            AccessError = OSError
+
         try:
             return self.image.size
-        except OSError:
+        except AccessError:
             return None
 
 
--- a/netbox/extras/models.py
+++ b/netbox/extras/models.py
@@ -611,21 +611,20 @@ class ImageAttachment(models.Model):
     @property
     def size(self):
         """
-        Wrapper around `image.size` to suppress an OSError in case the file is inaccessible. When S3 storage is used
-        ClientError is suppressed instead.
+        Wrapper around `image.size` to suppress an OSError in case the file is inaccessible. Also opportunistically
+        catch other exceptions that we know other storage back-ends to throw.
         """
-        from django.conf import settings
-        if settings.MEDIA_STORAGE and settings.MEDIA_STORAGE['BACKEND'] == 'S3':
-            # For S3 we need to handle a different exception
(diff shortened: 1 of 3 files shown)


robbat2:
@steffann reading this, you have configuration, but nothing else. Maybe re-title the PR to be `configuration structure for alternate image storage`

Some ideas that I think would work for S3 storage:
- Browser uses CORS upload to the bucket w/ a presigned URL
- Netbox serves images w/ presigned URLs that are good for 6 hours (cache the entire presigned URL, so that it doesn't change on the browser side and the cache is good)

steffann:
> @steffann reading this, you have configuration, but nothing else.

I don't understand what you mean by that. It's a fully working solution, the limitation is that only S3 (and compatible) systems are supported for now / in this patch.

> Netbox serves images w/ presigned URLs that are good for 6 hours

This is exactly what happens with django-storages and S3.

lampwins:
> I don't understand what you mean by that. It's a fully working solution, the limitation is that only S3 (and compatible) systems are supported for now / in this patch.

This has always been my issue with this FR. Would it be possible to abstract the config options so it is not locked to S3? Meaning other storage providers could be added in a generic fashion without having to add additional config options.

steffann:
Storage providers are so diverse that I don't see a good way of implementing something generic just for storage providers.

If we want the flexibility I'd suggest using something like what @wmfgerrit proposed in #1814. See my comment at the bottom of that thread 🙂

steffann:
PS: this patch supports S3 and compatible, so it is slightly more flexible than just S3 but not by much.

netbox/extras/models.py
```diff
@@ -750,11 +750,18 @@ def delete(self, *args, **kwargs):
...
         """
-        Wrapper around `image.size` to suppress an OSError in case the file is inaccessible.
+        Wrapper around `image.size` to suppress an OSError in case the file is inaccessible. When S3 storage is used
+        ClientError is suppressed instead.
         """
+        from django.conf import settings
+        if settings.MEDIA_STORAGE and settings.MEDIA_STORAGE['BACKEND'] == 'S3':
+            from botocore.exceptions import ClientError as AccessError
+        else:
+            AccessError = OSError
```
jeremystretch: Although trivial, the reassignment of one exception class to another seems almost certain to frustrate future troubleshooting efforts. Can we split this out into two separate return conditions under `if settings.MEDIA_STORAGE ...` instead?

steffann: Sure! You want me to change the PR, or are you already ahead of me? :)

steffann: Never mind, already done

PR #3680, opened 2019-11-07
Closes #3663: API filter by created, last_updated

### LINKED ISSUE #3663: API filter by created, last_updated
### Environment
* Python version:  3.7.3
* NetBox version:  2.6.6

### Proposed Functionality

The date fields (created, last_updated) added in Version 2.3.0 should also be filterable.

### Use Case

Get a List of all Racks that have been changed > yyyy-mm-dd

### Database Changes

none

### External Dependencies

none added

struppinet:
### Fixes: #3663: API filter by created, last_updated
Creating a new class that the applicable existing FilterSets inherit. I think it makes sense to have exact, gte, and lte filters for the created and last_updated fields.

### DIFF
--- a/netbox/extras/filters.py
+++ b/netbox/extras/filters.py
@@ -241,3 +241,23 @@ class ObjectChangeFilter(django_filters.FilterSet):
             Q(user_name__icontains=value) |
             Q(object_repr__icontains=value)
         )
+
+class ChangeLoggedFilter(django_filters.FilterSet):
+    created = django_filters.DateFilter()
+    created__gte = django_filters.DateFilter(
+        field_name='created',
+        lookup_expr='gte'
+    )
+    created__lte = django_filters.DateFilter(
+        field_name='created',
+        lookup_expr='lte'
+    )
+    last_updated = django_filters.DateTimeFilter()
+    last_updated__gte = django_filters.DateTimeFilter(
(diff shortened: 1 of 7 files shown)


netbox/extras/filters.py
```diff
@@ -241,3 +241,23 @@ def search(self, queryset, name, value):
...
+    )
+    last_updated = django_filters.DateTimeFilter()
+    last_updated__gte = django_filters.DateTimeFilter(
+        field_name='last_updated',
+        lookup_expr='gte'
+    )
+    last_updated__lte = django_filters.DateTimeFilter(
+        field_name='last_updated',
+        lookup_expr='lte'
+    )
```
lampwins: This must have a single blank line at the end of the file.

lampwins (changes requested):
I like this, just a small PEP8 fix.

I think we should use a different name for the filter. Change logging is a concept we use in a different context. Perhaps `CreatedUpdatedFilter`?

Also, it would be nice to have a couple of tests for this.

struppinet:
Fixed PEP8 warnings, renamed class. (used the name because of the model inheritance "class ChangeLoggedModel(models.Model):"

struppinet:
I will check Tests with a friend and maybe add them on the weekend.

PR #3690, opened 2019-11-13
Switch from "pycryptodome" to "pycryptodomex"

### LINKED ISSUE #3689: Switch from pycryptodome to its library independent variant pycryptodomex
### Environment
* Python version:  3.6.9
* NetBox version:  2.6.7

### Proposed Functionality

Change the requirement from `pycryptodome` to its library independent variant `pycryptodomex`.

This shouldn't lead to any specific changes to data models, user interfaces, workflows, etc. 

### Use Case

When Netbox is used outside a virtualenv it can happen that there's a Python package conflict between `pycryptodome`, which is required by NetBox, and `pycrypto`, which another package requires (e.g. Salt in conjunction with ZeroMQ).

This is because `pycryptodome` is meant as drop-in replacement for `pycrypto` and both packages use `Crypto` as namespace.

Switching from `pycryptodome` to `pycryptodomex` will solve the issue as the latter one uses `Cryptodome` as package name.

### Database Changes

This shouldn't require any database changes.

### External Dependencies

Replace the existing external dependency `pycryptodome` with `pycryptodomex`.

knobix:
### Fixes: #3689

When NetBox is used outside an virtualenv it can happen that there's a
Python package conflict between "pycryptodome", which is required by
NetBox, and "pycrypto", which some other package requires.

Thus switch from "pycryptodome" to its library independent variant
"pycryptodomex". The latter one uses "Cryptodome" as package name
instead of "Crypto".

My tests were so far:

- Login/Logoff -> OK
- Unlock/Lock secrets via private key -> OK
- Create new user key pair -> OK
- Activate new user key pair -> OK
- Unlock/Lock secrets with new user key pair -> OK

### DIFF
--- a/base_requirements.txt
+++ b/base_requirements.txt
@@ -81,4 +81,4 @@ py-gfm
 
 # Extensive cryptographic library (fork of pycrypto)
 # https://github.com/Legrandin/pycryptodome
-pycryptodome
+pycryptodomex
--- a/netbox/secrets/api/views.py
+++ b/netbox/secrets/api/views.py
@@ -1,6 +1,6 @@
 import base64
 
-from Crypto.PublicKey import RSA
+from Cryptodome.PublicKey import RSA
 from django.db.models import Count
 from django.http import HttpResponseBadRequest
 from rest_framework.exceptions import ValidationError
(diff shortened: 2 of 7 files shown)


jeremystretch:
Please see my comment on #3689

Issue #2248, opened 2018-07-18
Render rack elevations using HTML5 SVG


jeremystretch:
### Issue type
[x] Feature request 
[ ] Bug report      
[ ] Documentation   
[ ] Housekeeping    

### Environment
* Python version:  3.5.2
* NetBox version:  2.3.6

### Description
Currently, rack elevations are rendered in HTML as multi-layered lists of divs, which is exactly as terrible as it sounds. While this approach has been reasonably sufficient, an elevation's fixed size and limited screen space has led to several limitations:

* #160 - Rack elevations are not printer-friendly
* #1911 - Difficult to show child devices within a parent
* #1998 - Elevations do not include status information

The idea was raised in #1529 to adopt HTML5's canvas feature and render racks as free-form drawings. While this will take a substantial amount of effort on the backend, it should result in a _much_ cleaner and more flexible illustration of rack elevations.

Further research into the feasibility of this feature is needed, but I wanted to get it on the board for reference.

jeremystretch:
[fabric.js](http://fabricjs.com/) looks pretty promising.

andre-lx:
Look at the svg.js. 

Uses pure svg instead of canvas (fabric.js) and seems ideal so that in future other features can be added, like childs inside parents on rack and cable visualization.

tb-killa:
Hi @jeremystretch i build a simple Sample via "svg.js".
https://jsfiddle.net/tb_killa/1wu5stgj/

It´s not perfect but it shows you that this could work.

candlerb:
We have had good success doing Netbox visualizations using the [d3 js](https://github.com/d3/d3) library

hellerve:
Have we decided on a framework yet? I’d volunteer to work on this unless anyone has started to work already!

jeremystretch:
I believe there's a strong desire for the ability to return SVG files directly from the API, without requiring any UI-centric processing (e.g. something like fabric.js). This would enable the diagrams to be rendered on-demand and embedded in other applications. (I'm not sure about the details concerning how we'd deliver non-JSON content yet.)

hellerve:
Alright, I can look into that. I don’t have a ton of experience building SVGs, especially on the backend, but I’ll let you know what I can come up with!

hellerve:
[...]

Issue #3569, opened 2019-10-04
Replace numeric constants with slug values


jeremystretch:
### Proposed Changes
Replace the numeric values used for many fields with human-friendly slug values. For example:

```
RACK_STATUS_RESERVED = 0
RACK_STATUS_AVAILABLE = 1
RACK_STATUS_PLANNED = 2
RACK_STATUS_ACTIVE = 3
RACK_STATUS_DEPRECATED = 4
RACK_STATUS_CHOICES = [
    [RACK_STATUS_ACTIVE, 'Active'],
    [RACK_STATUS_PLANNED, 'Planned'],
    [RACK_STATUS_RESERVED, 'Reserved'],
    [RACK_STATUS_AVAILABLE, 'Available'],
    [RACK_STATUS_DEPRECATED, 'Deprecated'],
]
```

would become

```
RACK_STATUS_RESERVED = 'active'
RACK_STATUS_AVAILABLE = 'planned'
RACK_STATUS_PLANNED = 'reserved'
RACK_STATUS_ACTIVE = 'available'
RACK_STATUS_DEPRECATED = 'deprecated'
RACK_STATUS_CHOICES = [
    [RACK_STATUS_ACTIVE, 'Active'],
    [RACK_STATUS_PLANNED, 'Planned'],
    [RACK_STATUS_RESERVED, 'Reserved'],
    [RACK_STATUS_AVAILABLE, 'Available'],
    [RACK_STATUS_DEPRECATED, 'Deprecated'],
]
```

### Justification
Employing human-friendly slug values make consuming the REST API more convenient. It also allows more human-friendly representations of the pertinent field values in other formats, such as YAML (see #451).

candlerb:
Would this also change the database column types to enum?

jeremystretch:
I don't think Django has built-in support for enumeration types, and adding it probably wouldn't get us much. It would also mean incurring a new migration each time a field had its set of valid choices modified.

candlerb:
In that case, surely the numeric values for RACK_STATUS_RESERVED, RACK_STATUS_AVAILABLE etc still need to be preserved in the Python code?

You could instead have RACK_STATUS_DISPLAY_CHOICES and RACK_STATUS_API_CHOICES ?

jeremystretch:
The point of the FR is to replace the numeric values with strings entirely. This will occur within a migration and be completely transparent to the user.

candlerb:
So the database field will be a varchar, not an enum? Got it.

[...]

PR #3702, opened 2019-11-21
Rack elevations API

hellerve:
### Fixes: #2248

This PR introduces the rack elevations API. It’s a bit special because it returns SVG results. It replaces the current inline `div`-based solution. I marked it as a draft because there are a few open questions.

I guess an overview of what its capabilities are are in order. Basically, you call `/api/dcim/rack-elevations/<rack-pk>`, and you get back a SVG of that rack’s elevations. There are three URL parameters here:

- `face`: specifies which face to render, `0` for front, `1` for rear. Front is the default.
- `width`: specifies the desired width of the SVG in pixels. 230 is the default.
- `slot_height`: specifies the height of one slot (i.e. one rack unit). I opted for this rather than a total height because the height of racks is usually not as important visually as the height of the devices/components. (I’m not sure about the naming of this param and/or you’d rather like to have total height. Input is appreciated)

The look and feel of the rendered SVG is pretty much identical to that of the DIV, save any bugs or open questions.

### Open Questions

The big one first: what about those tooltips? They’re super useful and I kind of don’t want to lose them, but they’re really not meant to be ported to SVGs. If you twist my arm I can port them, but if we can find a better solution that’d be great!

Then there are a few open questions related to the code climate:
- For the foreground color I’m currently reusing the `fgcolor` utility,w hich is a templatetag. That’s kind of bad style, and I’d opt for extracting that functionality into a helper that is shared by the tempalte tag and the elevation API if desired (it’s desired by me, anyway).
- The stylesheet that we’re using for the rendered SVG is inline, which I think is good (so that the SVG is freestanding). But it’s also inline in the sense of `within the source code`, which is... less ideal. Where should we put it?
- The viewset is fairly readable (to me anyway), but I’m sure it could be improved. Any feedback here is appreciated. A few things that I’m not sure of myself:
  - The GET param validation happens fairly manually. It’s alright I guess, but maybe REST framework has something fancy that we could use?
  - While most of the functions are local to the view set, they are generic helpers that don’t use `self` at all. Should they be factored out somewhere?
  - I’m getting the elevations myself instead of using `rack.get_*_elevation` because I always need both anyway and I was able to simplify the code tremendously by specializing it. At first I tried to reuse the functions, but this leads to more queries and a bunch of glue code to fuse everything. In fact, I was able to remove the `rack.get_*_elevation` functions alltogether, and the `remove_redundant` keyword argument, since they’re now unused. Is the handful of lines of kind-of-but-not-really duplicated code acceptable if we get rid of basically the same amount of code in return?
  - I was liberal in removing legacy code, including stylesheets only used by the elevations and such, and I’m pretty sure nothing broke, but you never know. Is there anything I missed?

All in all, this was equally fun and frustrating, and I hope this is either already useful or can be made useful. I’m looking forward to your review!

Cheers


hellerve:
The linter was not amused when I forgot to add spaces around my operators. Should be fixed now.

lampwins:
@hellerve this looks really good! I have a couple of high-level comments at this point.

- I think it would be really cool to use this to also expose rack elevations as a dictionary in JSON. Some of your helper methods would actually make this rather trivial.
- With that mind, we should use the `Accept` header to either return an application/json response or the svg. This is more of a pet peeve of mine when REST API's "randomly" change content types without client negotiation.
- If we are only going to render elevations one at a time, would it not make sense to make this a detail action route on the RackViewSet instead of a standalone APIView?
- I understand why you chose to move the helper functions out of the Rack model, but I would actually argue against this because it then limits this low-level API to the REST API only. I.e. I would not be able to use this logic in the nbshell, or my reports, custom scripts, etc, which are officially supported NetBox interfaces.

hellerve:
@lampwins These are all helpful comments! Comment 1 and 2 I’ll try and implement as is.

As for your third comment, I think it makes sense that this is standalone, similarly to how `rack-reservations` is standalone as well. I think they’re either both their own endpoints, or both part of racks. Does that make sense?

As for the fourth comment, I’ll try and put that back into the `Rack` model, though I’m not sure how generally useful it is. Though I understand that even if it’s only useful for one person it’d make sense to have it somewhere where they can use it. I’ll think about naming and whether we can maybe even merge this again with other helper functions, although I’m pretty sure I looked at this when I first worked on it.

hellerve:
@lampwins Something that I didn’t consider is that HTTP Accept headers can’t be manipulated inside tags like `<object>` (and it doesn’t get set by Firefox at least). This means that we need to set `image/svg+xml` as the default value—which would be a little bit annoying in a JSON API—, unless I’m missing an alternative. I vote against loading the SVG in a request made by JS (where we would have that kind of control, but I don’t want to need to enable JS to see the rendered racks).

What do you think?

hellerve:
I drafted something based on `rest_framework` renderers, but sadly this doesn’t solve the more general issue of headers not being set “correctly” by the browser. Once we’ve decided on an alternative I’d be happy to commit and add this to the branch.

For reference, these are the possible solutions that I see:
- Request the SVG from JS.
- Make SVG the default renderer for this view.
- Separate the two endpoints, i.e. making one for JSON and one for SVG.
- Drop the idea of supporting both for now.

jeremystretch:
Great work! I'm going to list my thoughts in no particular order...

* I agree with @lampwins that the SVG rendering code should not be restricted to the API. Ideally it should be accessible directly from the Rack model. (Due to the amount of code required, I'd opt to offload this to a separate class which gets inherited by the `dcim.Rack` model.)

* Per #3569, the rack faces have just recently changed from `0` and `1` to `front` and `rear`, respectively. Just needs a find & replace.

* I would rename `width` and `slot_height` to `unit_width` and `unit_height` for consistency. (NetBox favors the term "unit" for rack openings.)

* The `fgcolor` template tag is a wrapper; you want the underlying utility function at `utilities.util.foreground_color`.

* No problem with using inline style for the SVG but we should move that whole blob of CSS into `dcim/constants.py`.

* I'm not 100% on SVG, but I believe if we embed the SVG code into the HTML directly (as opposed to making an API call for it), we should be able to extend it with tooltips/pop-ups. IIRC that was my original plan. We would just need a toggle on the render method to enable/disable the inclusion of the additional code. (Again, this is just my theory.)

Hoping to share more feedback once I've had the opportunity to try it out. Again, great work!

jeremystretch:
I've re-targeted this PR against a new dedicate branch off of `develop-2.7` (since this is scheduled to be incorporated in the v2.7 release).

I am going to release the v2.7 beta next week, so it would be fantastic to get this wrapped up in the next few days. I'm okay running with it from here if you don't have time @steffann.

steffann:
> I'm okay running with it from here if you don't have time @steffann.

I think you meant @hellerve there :)

lampwins:
@hellerve 

> I think it makes sense that this is standalone, similarly to how rack-reservations is standalone as well.

The difference is rack-reservations is its own model which needs to support creating/modifying/deleting. To me, this elevation feature's implementation is sort of like the napalm integration in the API. The napalm detail route on the device viewset only allows GET operations as a convenience method.

You raise a good point about the accept header method surrounding the actual use case of this endpoint. I would actually be in favor of a query parameter for this. The DRF browsable API view actually does this exact same thing with a query param that allows you to see raw JSON with `?format=json` or the default which renders HTML.

With that in mind I would implement this as a detail action route on the rack model view set:
```
/api/dcim/racks/<id>/elevation/
```
Would return JSON, while passing `?format=svg` would render an SVG:
```
/api/dcim/racks/<id>/elevation/?format=svg
```
That would cover my concerns about client negotiation.

hellerve:
@jeremystretch I might be able to work on this tomorrow (to avoid timezone confusion, that would be European sunday). I’ll keep you posted if I can’t make it, but I should be able to at least do some of the smaller tasks.

hellerve:
So, somehow the retargetting messed up the state of this branch. I’m not quite sure what happened, but I’m going to try and fix it.

netbox/dcim/api/serializers.py
```diff
@@ -157,7 +157,7 @@ class RackUnitSerializer(serializers.Serializer):
     """
     id = serializers.IntegerField(read_only=True)
     name = serializers.CharField(read_only=True)
-    face = serializers.IntegerField(read_only=True)
```
hellerve: It seems like this was forgotten in #3569?

netbox/dcim/api/views.py
```diff
@@ -181,7 +183,7 @@ def units(self, request, pk=None):
         List rack units (by rack)
         """
         rack = get_object_or_404(Rack, pk=pk)
-        face = request.GET.get('face', 0)
+        face = request.GET.get('face', 'front')
```
hellerve: This one also had to change to fix the tests.

hellerve:
Okay, so I updated this branch with the feedback that you gave me, @jeremystretch. I didn’t manage to also put in the new JSON API or move this to a different class, and I think I won’t have time for that next week, so if you want to finish work on this that’s totally fine with me.

Sorry I wasn’t able to put in all of my changes. I also realized that there is already a way to get rack units in the API and now I’m a little confused as to what the added value of the JSON version of this endpoint is. I’m excited to see what you end up building on top of my janky foundations!

lampwins:
@hellerve great work on this! I am going to merge this into the 2248 branch and clean a few things up. Ultimately my plan is to deprecate the current `units` detail route (in 2.8) in favor of this.

Issue #822, opened 2017-01-20
Bulk import of device components


dav3860:
I would to nice to be able to import interfaces in devices in CSV format, like VLANs, etc.

jeremystretch:
Could you describe your use case for this? Interfaces can be created in bulk and should typically be inherited at the time of creation from a device's device type.

dav3860:
We have a bunch of devices with dozens of VLAN interfaces. It was a pain to add them to netbox as the VLANs were randomly numbered. Indeed, we haven't included the VLAN interfaces to the corresponding device types as these are virtual interfaces, so we had to add them manually.
I have created a PR referencing #821 to support commas in bulk creation. But it doesn't help to create descriptions in bulk too. 
So, having a way to import interfaces with descriptions would be nice.

killcity:
+1 on this.

mike-19:
Interfaces inherits from their device type however their MAC details needs to be manually added per device it becomes tedious for bulk of machines 

We will need a API to ADD MAC associated with a interface for a host as well UPDATE MAC for an associated host interface whenever NIC  replacement takes place

ankerstal:
We would also need this. We have 2000 servers in a exisiting DC and need a some way to import "current state" including interfaces, their vlans and mac addresses.

Grokzen:
I would second this feature but from a much more simple use-case, simply to add the ability to move all data from one netbox installation to another without modify it. Simply put just to export all the data for all Interfaces and to what VM:s they are connected. Then import it again in a new Netbox instance.

candlerb:
The original counter-argument said:

> Interfaces can be created in bulk and should typically be inherited at the time of creation from a device's device type

However, VMs do not have templates or device types, and don't even get a single interface by default (#1492).  So bulk import of interfaces would be especially useful for VMs.

Ricaz:
This would be _so_ helpful! Adding big core routers with many interfaces of different types and statuses can be a pain manually.

chonty:
BIG +1 for this.  Our use case: several routers terminating hundreds of L2 customer connections as dot1q subinterfaces.  Without the ability to import, I need to manually enter these hundreds of connections with unique interface names and descriptions.  

I can very easily create a CSV from router output which would turn a day+ long job into a 5 minute job.

jeremystretch:
Extending the scope from interfaces to all device components.

candlerb:
Another use case for importing interfaces is for routers with line cards.

[...]

Issue #3564, opened 2019-10-03
Add an interfaces list view


darcynz:
### Environment
* Python version:  
* NetBox version:  

### Proposed Functionality
Add a GUI page for interfaces -> https://netbox/api/dcim/interfaces/. Filtering by site/region and tags would allow easier management of interfaces.

Export template functionality would also add.

### Use Case

Enable holistic management of interfaces. 

We would use this for our Switch port register (currently done in a spreadsheet to keep track of allocated interfaces in the datacenter). 

In general Netbox would be ideal for holistic management of interfaces, making it easy to view them and. It can also be leveraged by adding metadata via tags. Eg monitoring groups.

**Noted** this was raised before  ( https://github.com/netbox-community/netbox/issues/291) but wanted to show the relevance with our use case.

### Database Changes
None that I'm aware off

### External Dependencies
None that I'm aware off

jeremystretch:
I don't think it would make sense to do this for only interfaces and not for the six other types of device components (console ports, power ports, etc.).

wols:
I could export all interface _descriptions_ (before the connection) to fill a device configuration template.

darcynz:
I agree. I can see benefits seeing the the other components holistically eg one use case could be to find devices with only one power port that need another. It would also enable each component type as a source of truth domain, as users able to see the items holistically easily and any mistakes / omissions. Currently this is done via reports.

Perhaps a multi discipline components page could work. The main difficulty would be filtering, I suppose component specific tags could be listed under their own headings. A check box filter could be used to hide/add component types. 

Curious to see community thoughts on this...

darcynz:
We've managed to find a work around to recent post on exporting associated interfaces via device model. Thanks Jeremy!!!  Would still be great to see interfaces via the GUI but helps provide the overview we wanted.

I've adjusted the template to give a report like format and posted below in case any one else stumbles across this.

devicename, deviceid, interfacename,  IP Address(es), description, connected device, connected endpoint, formfactor, mtu, lag, enabled?,  tags
[...]

PR #3711, opened 2019-11-23
822 bulk import of device components

steffann:
### Fixes: #822 

This PR implements bulk import forms for device components. The only thing missing is a decent place to link to/from these import forms. I suggest adding those links after #3564 has been implemented. Those pages would be the perfect place for a link to the import.


lampwins:
This is looking good. I noticed several of your __init__ methods have some duplicated logic. Do you think there is any way we can reduce some of that to make it more DRY?

Also, you have some TODOs for URLs, should we wait to merge this until after that is complete and added here?

steffann:
I have thought about abstracting the logic in the init methods a bit, but although they are similar they are not identical. Creating a helper function/method would make the code more compact but possibly harder to read (although that is very personal). In short: i have a slight preference for leaving it as it is, but don't mind either way.

I think it's ready to merge. One of the merges has to go first, and I don't think the order really matters as long as the URLs are put in place before 2.7 is released.

PR #3719, opened 2019-11-26
Add list view for device components

DanSheps:
### Fixes: #3564 

* Adds list views for: Interfaces, Front Ports, Rear Ports, Console Ports, Console Server Ports, Power Ports, Power Outlets & Device Bays
* Adds new filter
* Adds new section in nav menu for device component list view


DanSheps:
Are we good, as a start, with the generic columns (parent/device, name, type, description, cable)?

lampwins:
@DanSheps I think it is fine for all models _except_ interface. The interface model is just to core to the NetBox mission that I think it justifies overriding the table definition with more relevant fields. But if that becomes a line in the sand blocking this from moving forward, I will concede.

netbox/dcim/views.py
```diff
@@ -1365,6 +1401,16 @@ class PowerOutletBulkDeleteView(PermissionRequiredMixin, BulkDeleteView):
 # Interfaces
 #
 
+class InterfaceListView(PermissionRequiredMixin, ObjectListView):
+    type = 'Interface'
```
jeremystretch: Not sure where this came from; is it needed?

netbox/dcim/tables.py
```diff
@@ -643,41 +643,95 @@ class Meta(BaseTable.Meta):
...
     class Meta(BaseTable.Meta):
         model = Interface
         fields = ('name', 'type', 'lag', 'enabled', 'mgmt_only', 'description')
 
 
+class InterfaceDetailTable(DeviceComponentDetailTable):
+    component_type = 'Interfaces'
+    parent = tables.LinkColumn(order_by=('device', 'virtual_machine'))
+
+    class Meta(InterfaceTable.Meta, DeviceComponentDetailTable.Meta):
```
jeremystretch: Is there a need to inherit from other Meta classes here, since we're setting the parameters anyway? If there is, inheritance order should be consistent with the other component tables.

netbox/dcim/tables.py
```diff
@@ -643,41 +643,95 @@ class Meta(BaseTable.Meta):
...
 
+class DeviceComponentDetailTable(BaseTable):
+    component_type = ''
+
+    pk = ToggleColumn()
+    cable = tables.LinkColumn()
+
+    class Meta(BaseTable.Meta):
+        order_by = ('device', 'name')
+        fields = ('pk', 'device', 'name', 'type', 'description', 'cable')
```
jeremystretch: I'd recommend removing `type` from this table and adding it only to the tables for models where it's a field (similar to how we're handling `device`).

netbox/templates/inc/nav_menu.html
```diff
@@ -183,6 +183,26 @@
                         <li{% if not perms.dcim.view_interface %} class="disabled"{% endif %}>
                             <a href="{% url 'dcim:interface_connections_list' %}">Interface Connections</a>
                         </li>
+                        <li class="divider"></li>
+                        <li class="dropdown-header">Device Components</li>
```
jeremystretch: Console ports and console server ports are missing from the navigation menu.

netbox/templates/dcim/device_component_list.html
```diff
@@ -0,0 +1,19 @@
+{% extends '_base.html' %}
+{% load buttons %}
+
+{% block content %}
+<div class="pull-right noprint">
+    {% export_button content_type %}
+</div>
+<h1>{% block title %}{{ table.component_type }}{% endblock %}</h1>
```
jeremystretch: We should be able to reference the table's model here instead of setting `component_type` on each table. (You can use the `|model_name` filter from the `helpers` library to access the model's verbose name.)

jeremystretch (commented):
Excellent work! Thanks for doing all this heavy lifting. Just had a few comments/questions regarding cleanup/consistency.

netbox/dcim/views.py
DanSheps: Can be removed, thanks, will do it shortly.

netbox/dcim/tables.py
DanSheps: I believe we are taking the model from some and the columns from another, but for this one most likely not.

I believe I set it mostly for a "in the future case" and to maintain consistency with the other models but it looks like I goofed on the ordering.

netbox/templates/inc/nav_menu.html
DanSheps: Will fix this up right away

netbox/templates/dcim/device_component_list.html
DanSheps: This is exactly what I was looking for.  ^_^

netbox/dcim/tables.py
DanSheps: I have adjusted Interface and DeviceBay to instead not inherit the Component meta, as those are the only ones that will have a differing field list.

DanSheps:
> @DanSheps I think it is fine for all models _except_ interface. The interface model is just to core to the NetBox mission that I think it justifies overriding the table definition with more relevant fields. But if that becomes a line in the sand blocking this from moving forward, I will concede.

What other fields do you think should be added?

PR #3726, opened 2019-12-05
Fix race condition in available-prefix/ip APIs

### LINKED ISSUE #2519: Race condition when POSTing to /api/ipam/prefixes/<id>/available-ips/
### Environment
* Python version: 3.5.3
* NetBox version:  2.4.6

### Steps to Reproduce
Create a normal prefix, then run this code against the API (replace `$token`, `$prefixid` and `$host`):

`for i in $(seq 16); do curl -X POST -so/dev/null -H "Authorization: Token $token" -H "Accept: application/json; indent=4" https://$host/api/ipam/prefixes/$prefixid/available-ips/ & done`

This code requests 16 IP addresses simultaneously from Netbox triggering the race condition bug.

### Expected Behavior
Netbox assigns IP addresses from .1 up to .16: [Example screenshot](http://xor.meo.ws/8b820a1a/ab06/4ecc/b226/260e6076c512.png)

### Observed Behavior
Netbox assigned IP addresses from .1 up to .5, assigning some of the IPs multiple times: [Example screenshot](http://xor.meo.ws/b993ccfc/632f/4372/b7b3/c0f2398e9ddb.png)

mattolenik:
### Fixes: #2519
Use an [advisory lock](https://www.postgresql.org/docs/9.4/explicit-locking.html#ADVISORY-LOCKS) to create mutual exclusion and prevent these APIs from being called in parallel. If the API is called in one session, other sessions will have to wait for the first to finish. This prevents simultaneous calls that trample each other and insert duplicate records, which would result in multiple callers receiving the same prefix/IP.

Tests work by making many parallel calls and ensuring that only the expected number of records are actually created.

### DIFF
--- a/netbox/ipam/api/views.py
+++ b/netbox/ipam/api/views.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.db.models import Count
 from django.shortcuts import get_object_or_404
+from django_pglocks import advisory_lock
 from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.exceptions import PermissionDenied
@@ -10,6 +11,7 @@ from extras.api.views import CustomFieldModelViewSet
 from ipam import filters
 from ipam.models import Aggregate, IPAddress, Prefix, RIR, Role, Service, VLAN, VLANGroup, VRF
 from utilities.api import FieldChoicesViewSet, ModelViewSet
+from utilities.constants import ADVISORY_LOCK_KEYS
 from utilities.utils import get_subquery
 from . import serializers
 
@@ -86,9 +88,13 @@ class PrefixViewSet(CustomFieldModelViewSet):
     filterset_class = filters.PrefixFilterSet
 
     @action(detail=True, url_path='available-prefixes', methods=['get', 'post'])
+    @advisory_lock(ADVISORY_LOCK_KEYS['available-prefixes'])
     def available_prefixes(self, request, pk=None):
         """
         A convenience method for returning available child prefixes within a parent.
+
+        The advisory lock decorator uses a PostgreSQL advisory lock to prevent this API from being
+        invoked in parallel, which results in a race condition where multiple insertions can occur.
         """
         prefix = get_object_or_404(Prefix, pk=pk)
         available_prefixes = prefix.get_available_prefixes()
@@ -180,11 +186,15 @@ class PrefixViewSet(CustomFieldModelViewSet):
             return Response(serializer.data)
 
     @action(detail=True, url_path='available-ips', methods=['get', 'post'])
+    @advisory_lock(ADVISORY_LOCK_KEYS['available-ips'])
     def available_ips(self, request, pk=None):
         """
         A convenience method for returning available IP addresses within a prefix. By default, the number of IPs
         returned will be equivalent to PAGINATE_COUNT. An arbitrary limit (up to MAX_PAGE_SIZE, if set) may be passed,
         however results will not be paginated.
+
+        The advisory lock decorator uses a PostgreSQL advisory lock to prevent this API from being
+        invoked in parallel, which results in a race condition where multiple insertions can occur.
         """
         prefix = get_object_or_404(Prefix, pk=pk)
 
--- a/netbox/utilities/constants.py
+++ b/netbox/utilities/constants.py
@@ -27,3 +27,14 @@ COLOR_CHOICES = (
     ('111111', 'Black'),
     ('ffffff', 'White'),
 )
+
+# Keys for PostgreSQL advisory locks. These are arbitrary bigints used by
+# the advisory_lock contextmanager. When a lock is acquired,
+# one of these keys will be used to identify said lock.
+#
+# When adding a new key, pick something arbitrary and unique so
+# that it is easily searchable in query logs.
+ADVISORY_LOCK_KEYS = {
+    'available-prefixes': 100100,
+    'available-ips': 100200,
+}
--- a/base_requirements.txt
+++ b/base_requirements.txt
@@ -22,6 +22,10 @@ django-filter
 # https://github.com/django-mptt/django-mptt
 django-mptt
 
+# Context managers for PostgreSQL advisory locks
+# https://github.com/Xof/django-pglocks
+django-pglocks
+
 # Prometheus metrics library for Django
 # https://github.com/korfuri/django-prometheus
 django-prometheus
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ django-cors-headers==3.2.1
 django-debug-toolbar==2.1
 django-filter==2.2.0
 django-mptt==0.9.1
+django-pglocks==1.0.4
 django-prometheus==1.1.0
 django-rq==2.2.0
 django-tables2==2.2.1


lampwins:
@mattolenik NetBox still supports py3.5 so f-strings cannot be used yet. Please change these to `format()`.

netbox/utilities/utils.py
```diff
@@ -174,3 +178,39 @@ def to_meters(length, unit):
...
+        cursor = connection.cursor()
+        cursor.execute(f'SELECT pg_advisory_lock({lock_key});')
+        yield
+    finally:
+        cursor.close()
+
+
+def with_pg_advisory_lock(string_key):
+    """
+    Decorator that wraps a function with pg_advisory_lock
```
lampwins: Please provide more documentation here as to why this is needed. I would also like to see the link the Postgres docs here too.

netbox/utilities/utils.py
```diff
@@ -174,3 +178,39 @@ def to_meters(length, unit):
...
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kw):
+            with pg_advisory_lock(string_key):
+                return func(*args, **kw)
+        return wrapper
+    return decorator
+
+
+@lru_cache()
```
lampwins: Why the cache decorator?

netbox/ipam/tests/test_api.py
```diff
@@ -662,6 +667,77 @@ def test_create_multiple_available_ips(self):
...
+class ParallelPrefixTest(APITransactionTestCase):
+
+    CallResults = namedtuple('CallResults', ['successes', 'failures'])
+
+    def setUp(self):
+
+        super().setUp()
+        self.prefix1 = Prefix.objects.create(prefix=IPNetwork('192.168.2.0/24'))
+
+    def test_create_multiple_available_prefixes_parallel(self):
```
lampwins: Please write a meaningful docstring to explain the specifics of this test.

netbox/ipam/tests/test_api.py
```diff
@@ -662,6 +667,77 @@ def test_create_multiple_available_ips(self):
...
+        # should succeed, all others after that should "fail" with 201.
+        num_calls = 12
+        expected_count = 4
+
+        results = self.do_parallel_requests(url, data, num_calls)
+
+        self.assertEqual(results.successes, expected_count)
+        self.assertEqual(results.failures, num_calls - expected_count)
+
+    def test_create_multiple_available_ips_parallel(self):
```
lampwins: Please write a meaningful docstring to explain the specifics of this test.

lampwins (commented):
An elegant solution overall, but we need some more docs/comments for future NetBox historians. Also please don't use f-strings (yet :) )

netbox/utilities/utils.py
mattolenik: Just to avoid calculating the hash for every API call. However, SHA1 is really quite fast, so maybe this isn't worth the bother? Happy to just remove it.

netbox/utilities/utils.py
mattolenik: Done

netbox/utilities/utils.py
mattolenik: I ended up removing all of this and just using bigints that are hand-created in `constants.py`. This makes it much easier to correlate the lock IDs seen in query logs with the actual code.

mattolenik:
I made several changes that simplified the code and made test cases easier to understand. Thanks for the feedback so far!

netbox/utilities/utils.py
```diff
@@ -1,11 +1,16 @@
 import datetime
 import json
 from collections import OrderedDict
+from contextlib import contextmanager
+from functools import lru_cache, wraps
+from hashlib import blake2b
```
DanSheps: This appear to be unused, it is also causing a build failure.  What is this doing?

netbox/ipam/tests/test_api.py
```diff
@@ -604,6 +607,7 @@ def test_create_multiple_available_prefixes(self):
         self.assertHttpStatus(response, status.HTTP_201_CREATED)
         self.assertEqual(len(response.data), 4)
 
+
```
DanSheps: This extra line is causing a PEP8, likewise could you clean up the other PEP8 errors?

DanSheps (changes requested):

netbox/utilities/utils.py
mattolenik: Oops, that's leftover from some experimentation. Removed

DanSheps (approved):

DanSheps:
Looks like you fixed the errors, I will go through and re-check tomorrow during the day.

mattolenik:
Whoa, what's going on? How can a PR that fixes the issue, and been approved, be marked as wontfix? @DanSheps @lampwins any assistance here?

netbox/ipam/tests/test_api.py
```diff
@@ -662,6 +665,107 @@ def test_create_multiple_available_ips(self):
         self.assertEqual(len(response.data), 8)
 
 
+class ParallelAPITests(APITransactionTestCase):
```
jeremystretch: Let's avoid deviating too much from the existing test structure. I propose:

1. Move this class to `utilities.testing.ParallelAPITestCase` (keep `APITransactionTestCase` as it is)
2. Create `PrefixParallelTest` and `IPAddressParallelTest` in `ipam/tests/test_api.py`
3. Move the two `test_*` methods to their respective test classes

This will ensure the new test cases are immediately distinguishable, and allow for potential reuse of `ParallelAPITestCase` in the future. (Ideally, we'd incorporate these methods within the existing test cases but that doesn't seem feasible without converting the TestCase to a TransactionTestCase.) What do you think?

netbox/ipam/tests/test_api.py
```diff
@@ -662,6 +665,107 @@ def test_create_multiple_available_ips(self):
...
+    Due to the non-deterministic nature of race conditions, it is possible that these tests
+    may succeed on rare occasions, even when a race condition is present. These occurrences are
+    rare, however, and were difficult to reproduce during development of these tests. Regular
+    runs of these test cases should provide a high degree of certainty that no races are present.
+    """
+
+    # Convenience tuple for passing around request data for tests.
+    Request = namedtuple('Request', ['url', 'data'])
+
+    def setUp(self):
```
jeremystretch: Looks like an extraneous call to `setUp()`

netbox/utilities/testing.py
```diff
@@ -1,10 +1,12 @@
+from unittest import TestCase
 from django.contrib.auth.models import Permission, User
+from django.db import connection
```
jeremystretch: Can be omitted

netbox/utilities/constants.py
```diff
@@ -27,3 +27,17 @@
...
 )
+
+# Keys for PostgreSQL advisory locks. These are arbitrary bigints used by
+# the pg_advisory_lock contextmanager/decorator. When a lock is acquired,
+# one of these keys will be used to identify said lock.
+#
+# When adding a new key, pick something arbitrary and unique so
+# that it is easily searchable in query logs.
+ADVISORY_LOCK_KEYS = {
+    'available-prefixes': 123456710,
```
jeremystretch: Is there any reason that these specific integers were chosen? I realize that they are arbitrary, but would it make sense to use somewhat more consecutive values (similar to what we've done for other constants)?

netbox/utilities/constants.py
```diff
@@ -27,3 +27,17 @@
...
+    'available-ips': 234567820,
+}
+
+MAX_INT64 = 2**63 - 1
```
jeremystretch: We probably don't need to define constants for these; should be find to reference the absolute values directly within the function.

jeremystretch (changes requested):

jeremystretch:
In addition to the feedback above, I'm getting intermittent failures while running the new tests. It's an issue with `ObjectChangeMiddleware`, though I'm not sure how/why the assignment of `_thread_locals.changed_objects` is getting bypassed.

```
test_create_multiple_available_ips_parallel (ipam.tests.test_api.ParallelAPITests) ... ok
test_create_multiple_available_prefixes_parallel (ipam.tests.test_api.ParallelAPITests) ... ERROR

ERROR: test_create_multiple_available_prefixes_parallel (ipam.tests.test_api.ParallelAPITests)
  File "/home/jstretch/netbox/netbox/ipam/tests/test_api.py", line 689, in setUp
    super().setUp()
  File "/home/jstretch/netbox/netbox/utilities/testing.py", line 15, in setUp
    self.user = User.objects.create(username='testuser', is_superuser=True)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/db/models/query.py", line 422, in create
    obj.save(force_insert=True, using=self.db)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/contrib/auth/base_user.py", line 66, in save
    super().save(*args, **kwargs)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/db/models/base.py", line 741, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/db/models/base.py", line 790, in save_base
    update_fields=update_fields, raw=raw, using=using,
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/dispatch/dispatcher.py", line 175, in send
    for receiver in self._live_receivers(sender)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/dispatch/dispatcher.py", line 175, in <listcomp>
    for receiver in self._live_receivers(sender)
  File "/home/jstretch/netbox/netbox/extras/middleware.py", line 27, in handle_changed_object
    _thread_locals.changed_objects.append(
AttributeError: '_thread._local' object has no attribute 'changed_objects'
```

So far, this has only occurred under `test_create_multiple_available_prefixes_parallel`.

juan-vg:
@jeremystretch I bet this error is not happening due to the changes made, but due to the tests showing up an already existant error. I've experienced some issue like that one on previous versions when hitting the API several times in a fast way

jeremystretch:
The error does occur when I remove the decorator from the API views, however this may or may not be a symptom of the test environment (i.e. using `ThreadPoolExecutor`) as opposed to running normally. I say this because we have not had any previous reports of this error. Unfortunately I don't have the time to dig into this any further at present.

At any rate, we won't be able to merge tests that sporadically fail.

mattolenik:
Hmm, interesting! I'd like to repro that test failure. How are you running the tests? I've been using `./manage.py test ipam.tests` within a container using the `netbox-docker` project.

jeremystretch:
The errors occur sporadically when running `./manage.py test ipam.tests.test_api.ParallelAPITests` in my normal development environment (based on the v2.6 release).

juan-vg:
From the traceback posted by @jeremystretch I can infer that this error is not related to this particular fix/PR. I bet it's way more related to Django and the cache system.

Having a look at the _"netbox/extras/middleware.py"_ file you will be able to see that the '_changed_objects_' attrib **is separately set for each thread** environment (_threading.local()_) **which calls first the class** _ObjectChangeMiddleware_ (indirectly calling the method _\_\_call\_\__). The problem appears when a thread calls the function _handle_changed_object_ without calling first the class. In order to illustrate this behaviour I've added some logging to the file.
```bash
$ ./manage.py test ipam.tests.test_api.ParallelAPITests
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
(ThreadPoolExecutor-0_0)::__init__ :  
(ThreadPoolExecutor-0_1)::__init__ :  
(ThreadPoolExecutor-0_1)::__call__ :  
(ThreadPoolExecutor-0_2)::__call__ :  
(ThreadPoolExecutor-0_3)::__call__ :  
(ThreadPoolExecutor-0_5)::__call__ :  
(ThreadPoolExecutor-0_8)::__call__ :  
(ThreadPoolExecutor-0_0)::__call__ :  
(ThreadPoolExecutor-0_13)::__call__ :  
(ThreadPoolExecutor-0_6)::__call__ :  
(ThreadPoolExecutor-0_4)::__call__ :  
(ThreadPoolExecutor-0_10)::__call__ :  
(ThreadPoolExecutor-0_14)::__call__ :  
(ThreadPoolExecutor-0_15)::__call__ :  
(ThreadPoolExecutor-0_7)::__call__ :  
(ThreadPoolExecutor-0_11)::__call__ :  
(ThreadPoolExecutor-0_12)::__call__ :  
(ThreadPoolExecutor-0_9)::__call__ :  
E(MainThread)::handle_changed_object :  not hasattr(_thread_locals, "changed_objects")
E
ERROR: test_create_multiple_available_prefixes_parallel (ipam.tests.test_api.ParallelAPITests)
...
  File "./netbox/extras/middleware.py", line 33, in handle_changed_object
    _thread_locals.changed_objects.append(
AttributeError: '_thread._local' object has no attribute 'changed_objects'
```
As you can see above, _MainThread_ never calls the _\_\_call\_\__ method, so the attrib is never initialized. 

From my POV this behaviour is out of the scope of the solution provided by @mattolenik, but must be fixed prior to the PR acceptance and merge

jeremystretch:
> As you can see above, _MainThread_ never calls the _\_\_call\_\__ method, so the attrib is never initialized. 

So just to clarify, this is a symptom of how the test itself executes to simulate parallel requests, correct? Or am I missing something? (Sorry, it's difficult to follow the thread [pun!] here over time.)

mattolenik:
Thank you @juan-vg! Any idea why I've _never_ seen this when running the entire IPAM test suite, as opposed to the single test class? I also noticed that the first test executed in `ParallelAPITests` will succeed but the second will fail, every time. If separated out into two test classes (which I did as refactoring per the feedback), I see the same behavior: first test succeeds, second fails. If other tests are run before both, they all succeed.

juan-vg:
>Any idea why I've never seen this when running the entire IPAM test suite, as opposed to the single test class?

@mattolenik I bet when you run the entire IPAM test suite you give the chance to the system to create the attribute '_changed_objects_' on the MainThread before running this new tests, which doesn't mean it's properly handling the object update but at least it doesn't lead in an exception.

A quick workaround would be to create that missing attribute if it's not present, but it could have some not desired side effect that must be checked carefully as it seems to be a core thing used by the entire system

juan-vg:
The following changes do the trick and make the tests end successfully

```diff
diff --git a/netbox/extras/middleware.py b/netbox/extras/middleware.py
index 57f8f37d..a42f86ef 100644
--- a/netbox/extras/middleware.py
+++ b/netbox/extras/middleware.py
@@ -18,12 +18,29 @@ from .webhooks import enqueue_webhooks
 _thread_locals = threading.local()
 
 
+def _init_thread_locals():
+    """
+    Initialize the thread local env
+    """
+    _thread_locals.changed_objects = []
+
+
+def _check_thread_locals():
+    """
+    Check if the thread locals need to be initialized
+    """
+    if not hasattr(_thread_locals, "changed_objects"):
+        _init_thread_locals()
+
+
 def handle_changed_object(sender, instance, **kwargs):
     """
     Fires when an object is created or updated.
     """
     # Queue the object for processing once the request completes
     action = OBJECTCHANGE_ACTION_CREATE if kwargs['created'] else OBJECTCHANGE_ACTION_UPDATE
+
+    _check_thread_locals()
     _thread_locals.changed_objects.append(
         (instance, action)
     )
@@ -44,6 +61,7 @@ def handle_deleted_object(sender, instance, **kwargs):
     if hasattr(instance, 'tags'):
         copy.tags = DummyQuerySet(instance.tags.all())
 
+    _check_thread_locals()
     # Queue the copy of the object for processing once the request completes
     _thread_locals.changed_objects.append(
         (copy, OBJECTCHANGE_ACTION_DELETE)
@@ -54,6 +72,7 @@ def purge_objectchange_cache(sender, **kwargs):
     """
     Delete any queued object changes waiting to be written.
     """
+    _check_thread_locals()
     _thread_locals.changed_objects = []
 
 
@@ -71,13 +90,13 @@ class ObjectChangeMiddleware(object):
     have been created. Conversely, deletions are acted upon immediately, so that the serialized representation of the
     object is recorded before it (and any related objects) are actually deleted from the database.
     """
+
     def __init__(self, get_response):
         self.get_response = get_response
 
     def __call__(self, request):
-
         # Initialize an empty list to cache objects being saved.
-        _thread_locals.changed_objects = []
+        _init_thread_locals()
 
         # Assign a random unique ID to the request. This will be used to associate multiple object changes made during
         # the same request.

```

mattolenik:
Thank you, @juan-vg! Your fix work. I've updated the PR with that as well as other feedback.

Unfortunately, I've discovered a test isolation problem. I've found that, when I run the entire `ipam` test suite, everything works fine, but if I run the individual tests (both in this recent commit as well as before the changes), only the first test run will succeed, the second will fail. It appears to be a test isolation issue, the second run fails due to the data from the previous test data still being in the database.

I have a feeling this is related to the use of `TransactionTestCase` rather than just `TestCase`. `TestCase` wraps test cases in transactions to provide isolation, but `TransactionTestCase` (I think) is meant to tear down the test database or test tables. This doesn't seem to happen, _unless_ the entire test suite is run, then it's fine.

Any thoughts? 😕

netbox/utilities/constants.py
mattolenik: I changed them to some simpler, but also easy to search for numbers, let me know what you think

mattolenik:
Scratch that, I'm only seeing that behavior when under `netbox-docker`. When run execute tests in a virtualenv on my host machine I don't see this problem.

Please ignore the above comment :) I think everything's finally working!

mattolenik:
Alright, one more change! I figured out the above issue. It has nothing to do with Docker itself, the issue is all with caching. The `netbox-docker` project spins up a redis container with compose, and so models were being cached automatically. This lead to `IPAddress.objects.all()` returning models that were created in the previous test run, even though the entire test database was being torn down and recreated. I didn't see it locally because I wasn't using redis.

After rebasing today, I found that tests wouldn't run until I had redis running. I had copied my `configuration.py` from the example config, and due to [this change](https://github.com/netbox-community/netbox/commit/47fefbec078305a57bc620af7706fbca2182a659), it was now requiring redis. After installing and running redis, I saw the failures again.

I fixed it by just flushing the redis cache in `tearDown`. Does this seem like the correct fix?

juan-vg:
@mattolenik for me looks perfect! Erasing the test data you've generated after using it is the good practice! :+1: 
@jeremystretch Could you have a quick look? Everything seems to be solved and running fine

jeremystretch:
Please see my request for clarification above. It doesn't make sense to me to modify core middleware if the only purpose is to accommodate these specific tests. If rather forgo the tests entirely than risk introducing an issue with the change logging middleware (which has already been a source of problems).

mattolenik:
But it's literally just checking if a variable is set and if not, set it to an empty array. I think we can effectively reason about such a simple piece of code and be confident in what it does.

jeremystretch:
I understand it seems like a trivial change when examined in a vacuum, but consider that it will need to be maintained indefinitely along with the rest of the code base outside the context of this specific issue, and its necessity is not readily apparent. This issue has gone from effecting a SQL lock for a specific API endpoint to manipulating core middleware so it's prudent to be critical of the changes, no offense intended. I'm not comfortable accepting the responsibility for maintaining a change I can't 100% explain and defend. (This stance is borne not out of stubbornness but unpleasant experience.) Maybe some of the other maintainers can weigh in with their opinions, as admittedly I haven't been able to dedicate much time to this issue.

mattolenik:
I understand that, but at the same time, if some change reintroduces the race condition somehow, that breaking change will now go unnoticed. I think it's a lot harder to debug a race condition than a variable being set to `[]`. If some code is relying on an attribute of a private variable being null, I would say it's a bug in _that_ code.

jeremystretch (changes requested):
Let's walk this back and focus on the specific reason #2519 was raised, because the PR in its current form has too much going on. I realize the changes here may seem trivial to you, but please bear in mind that there are many, many other issues and PRs competing for attention. The simpler we can make this, the better a chance it has at getting merged.

Please remove all changes to the middleware, as well as the tests for now. I want to focus on why this is being done, and the justification for adopting a custom context manager as opposed to using something provided by either Django or an external package. Once we have that nailed down, we can discuss any issues there may be with testing. Sound good?

mattolenik:
@jeremystretch when you say justification for adopting a custom context manager, are you referring to the solution itself? Without that, there is nothing.

jeremystretch:
Versus using something else, yes. [django-pglocks](https://github.com/Xof/django-pglocks), for example. Normally I prefer to avoid introducing new dependencies, but for such a low-level feature it's probably best to offload as much of the long-term maintenance as we can.

(Django provides `select_for_update()`, though that is specifically for row locking. I haven't dug into Django deep enough to see if there's anything built-in that could be re-purposed here.)

rekahsoft:
@DanSheps do you share the insane opinion of @jeremystretch? I see a well written and documented solution to the problem we seek to resolve, but there has been (in my opinion) unreasonable push-back. For example:

> Let's walk this back and focus on the specific reason #2519 was raised, because the PR in its current form has too much going on

Has too much going on? Seriously?

> I'd rather forgo the tests entirely than risk introducing an issue with the change logging middleware (which has already been a source of problems).

Really? (jaw drops)

> Please remove all changes to the middleware, as well as the tests for now.

Jaw drops further.

> . I'm not comfortable accepting the responsibility for maintaining a change I can't 100% explain and defend

Sounds like @jeremystretch does not have the prerequisite knowledge to be a maintainer of this project.

@DanSheps, @juan-vg, @lampwins and other maintainers of this project: Do you agree with the dialog of @jeremystretch blocking this PR? We have been spending company resources on getting this PR submitted and reviewed, and are going to have to fork if we can't find a way to work together on this.

@juan-vg, @lampwins and @DanSheps though this comment is somewhat negative, I wanted to highlight that we really appreciate your support and assistance so far on this PR. Lets get this through already!

mattolenik:
I'd be happy to switch to the existing package (I just didn't want to introduce a dependency but it's a small one), it doesn't alleviate the need to test the race condition itself, however.

jeremystretch:
@rekahsoft You are welcome to share your opinions, but if you cannot do so in a courteous and professional manner you will be asked to refrain from engaging in discussion here. If you don't agree with the decisions being made you are absolutely free to fork the project and do as you see fit.

@mattolenik I think adopting django-pglocks is the way to go for now, provided there are no issues with it.

If changes to the project outside the scope of #2519 are necessary to accommodate testing of the change, this will need to be marked as blocked and a _new_ issue will need to opened to propose those changes (e.g. any modifications to the middleware) with the justification for doing so. But IMO it's not a _huge_ deal if we don't test specifically for parallel requests, since we obviously don't test for that anywhere else, so long as the current tests remain intact. Other maintainers might disagree though.

juan-vg:
@rekahsoft You are right, we want to solve this. But we (or at least me) don't want to do it at any cost. I completely agree with @jeremystretch that this PR has completely gone out of its initial scope. Despite this, @jeremystretch has offered a solution to the problem. From my point of view he has evaluated the risks of adding some code we're not able to assure if it's going to perform well 100% of the time, and he is acting in consequence. To be honest I would do the same, and keep in mind that I'm strongly interested on this PR being merged.

On the other hand, you're not paying to either @jeremystretch nor the project, so you should improve your manners. You've been a little bit rude. Consider carefully that you're benefiting from something free. Every effort from their side must be always welcomed (maybe discussed, of course), but never blamed.

@mattolenik is right when saying that some (other) code could be a bug itself if it relies on another bug, BUT:
- Introducing collateral changes to a very specific PR is not a good practice.
- Even being a bug, it could be a deliverated one to workaround the current status. You could be breaking that without directly notifying on the release notes.

Finally focusing on the current status, I have a question for @jeremystretch: ¿Does this context manager belong to Django core, or it's Netbox stuff? If it belongs to Django core it should be fixed on its main project, not here. If it's Netbox stuff we should hold this, open a new PR addressing the context-manager fix, merge it ASAP and then continue with this one.

@mattolenik could the tests rely on a different threading approach?. Maybe we can get the tests working without any other changes. I personally don't agree with the option of not including tests (even if I thank @jeremystretch for being polite and offering it)

lampwins:
As another maintainer, I will also add my support in saying that modifications to the middleware need to be handled outside of this PR.

We need to implement as much testing as we can, but in this case, modifications came about as a means to satisfy a new test. Arguably that is simply the nature of test-driven development, however, these modifications do not have anything to do with the core of this PR, which is the point @jeremystretch has raised.

Personally I believe the issues with the middleware warrant further investigation as a likely bug but this needs to be done as a separate issue.

DanSheps:
@rekahsoft 

It is absolutely on point asking for a unrelated model's change to be handled outside of the scope of this specific issue by creating a new issue.

As far as whether or not he has the requisite knowledge, I can assure you he most certainly does.  Lets not forget, if not for him Netbox wouldn't exist as this came out of his work at digital ocean and their need for a DCIM/IPAM there.  I imagine that he is in the same boat as me on this one, do I know what a async operation is?  Yes.  Am I familiar with async operations in python?  No.  If I need to troubleshoot this, can I?  Perhaps, but most likely not.

I also think there is another aspect you are not taking into account.  As a large number of us are not able to devote a large amount of time to the project (a couple of hours a week, sometimes a full day a week) it is difficult for us when we add a large change to the code base when it might be more easily accomplished by a a library that can handle certain tasks for us.  It also lifts a burden off the maintainers as we do not have to know & understand the underlying code as it is abstracted away for us.

It seems may also be mis-understanding blocking, blocking a issue/PR doesn't mean it goes away, it means that it can't be merged/fixed until the underlying issue is fixed.

mattolenik:
I appreciate the feedback, everyone, I'll be updating this to use `django-pglocks` shortly!

mattolenik:
Alright, updated to use `django-pglocks`!

juan-vg:
@jeremystretch Could you have a look please?

netbox/ipam/api/views.py
```diff
@@ -86,6 +88,7 @@ class PrefixViewSet(CustomFieldModelViewSet):
     filterset_class = filters.PrefixFilterSet
 
     @action(detail=True, url_path='available-prefixes', methods=['get', 'post'])
+    @advisory_lock(ADVISORY_LOCK_KEYS['available-prefixes'])
     def available_prefixes(self, request, pk=None):
         """
         A convenience method for returning available child prefixes within a parent.
```
lampwins: Can you please add a sentence or two in here about why we are now using the advisory lock decorator. Just so future NetBox historians understand it is in an effort to avoid a race condition.

netbox/ipam/api/views.py
```diff
@@ -180,6 +183,7 @@ def available_prefixes(self, request, pk=None):
             return Response(serializer.data)
 
     @action(detail=True, url_path='available-ips', methods=['get', 'post'])
+    @advisory_lock(ADVISORY_LOCK_KEYS['available-ips'])
     def available_ips(self, request, pk=None):
         """
         A convenience method for returning available IP addresses within a prefix. By default, the number of IPs
```
lampwins: Same as above.

lampwins (changes requested):
LGTM, but can you please add the two comments in the method docstrings.

netbox/ipam/api/views.py
mattolenik: Certainly! Done

netbox/ipam/api/views.py
mattolenik: Done

mattolenik:
Sure thing, done!

jeremystretch (approved):

jeremystretch:
LGTM! Thanks for hanging in there with this.

Issue #3664, opened 2019-11-03
Config context by tags


itmicus:
### Environment
* Python version:   3.7.3
* NetBox version:  2.6.6

### Proposed Functionality

Add to the config context assignment field with tags

### Use Case
I use the tag "monitored_by_zabbix" for devices and VMs wich will be monitoring in our Zabbix system.
I want to create config contex with config for Zabbix for this devices/vms

### Database Changes
Yes

### External Dependencies
None

kobayashi:
Could you give us more detail of your usecase step-by-step?

wfrced:
I believe I have a somewhat similar usecase, so will chime in.
With these changes I'd be able to:  

1. Get a list of hosts using Netbox plugin for Ansible;  
2. Create Ansible groups using tag names and add my hosts to these groups;  
3. Add tag's Config Context to group variables;
2. Create/Update hosts in Zabbix using Zabbix plugin for Ansible.  

Basically, it would allow me to use Config Contexts as Ansible groups.  
Or I could simply store a more general Zabbix config for, say, a Tenant or a Site, and use tag's CC to store a boolean - whether to enable monitoring for this host.

kobayashi:
I am taking this

PR #3752, opened 2019-12-11
implement 3664

kobayashi:
### Fixes: #3664

This is a function to add config context by tags. but still need work for tag selection box rendering.
The `tags` select option shows as same as other tags fields provided with taggit. If any items are alocated there, can be added, deleted. To solve this, avoiding model item name `tags`. Is there better way for this?


kobayashi:
The tag selection field shows like this.
Screenshot from 2019-12-11 04-51-01

netbox/utilities/views.py
```diff
@@ -144,7 +144,7 @@ def get(self, request):
             table.columns.show('pk')
 
         # Construct queryset for tags list
-        if hasattr(model, 'tags'):
+        if hasattr(model, 'tags') and type(model.tags).__name__ is not 'ManyToManyDescriptor':
```
jeremystretch: How about:

```
from django.db.models.manager import Manager
...
if hasattr(model, 'tags') and issubclass(model.tags, Manager):
```

That's probably conveys our intention a little better. Admittedly, just using `hasattr` was lazy on my part.

PR #3754, opened 2019-12-11
2248 svg rack elevations

lampwins:
### Fixes: #2248
Cleanup of the new rack elevation endpoint


netbox/dcim/api/serializers.py
```diff
@@ -157,7 +157,7 @@ class RackUnitSerializer(serializers.Serializer):
     """
     id = serializers.IntegerField(read_only=True)
     name = serializers.CharField(read_only=True)
-    face = serializers.IntegerField(read_only=True)
+    face = ChoiceField(choices=DeviceFaceChoices, required=False, allow_null=True)
```
jeremystretch: Should be `read_only=True`

docs/release-notes/version-2.7.md
```diff
@@ -38,6 +38,28 @@ components can be imported in CSV-format.
...
+/api/dcim/racks/<id>/elevation/
+```
+
+By default, this endpoint returns a paginated JSON response representing each rack unit in the given elevation. This is the same response returned by the rack units detail endpoint and for this reason the rack units endpoint has been deprecated and will be removed in v2.8 (see [#3753](https://github.com/netbox-community/netbox/issues/3753)):
+
+```
+/api/dcim/racks/<id>/units/
+```
+
+All internal use of the rack units endpoint has been updated to use the new rack elevation endpoint.
```
jeremystretch: This isn't necessary to point out in the user-facing release notes.

docs/release-notes/version-2.7.md
```diff
@@ -38,6 +38,28 @@ components can be imported in CSV-format.
...
+
+In order to render the elevation as an SVG, include the `render_format=svg` query parameter in the request. You may also control the width of the elevation drawing in pixels with `width=<width in pixels>` and the height of each rack unit with `unit_height=<height in pixels>`. The `width` defaults to `230` and the `unit_height` default to `20` which produces elevations the same size as those that appear in the NetBox Web UI. The query parameter `face` is used to request either the `front` or `rear` of the elevation and defaults to `front`.
```
jeremystretch: An example request URL would be helpful here.

netbox/dcim/models.py
```diff
@@ -451,7 +454,215 @@ def to_csv(self):
...
+                )
+
+            unit_cursor += height
+
+        # Wrap the drawing with a border
+        drawing.add(drawing.rect((0, 0), (width, self.u_height * unit_height), class_='rack'))
+
+        return drawing
+
+    def get_rack_units(self, face=DeviceFaceChoices.FACE_FRONT, exclude=None, expand_devices=True):
```
jeremystretch: IMO generic methods like `get_rack_units()`, `get_available_units()`, and `get_reserved_units()` should stay on the Rack model as they don't pertain specifically to rendering elevations (the unit lists are also used for validation in other areas).

netbox/templates/dcim/rack_elevation_list.html
```diff
@@ -18,9 +18,9 @@ <h1>{% block title %}Rack Elevations{% endblock %}</h1>
                             <p><small class="text-muted">{{ rack.facility_id|truncatechars:"30" }}</small></p>
                         </div>
                         {% if face_id %}
-                            {% include 'dcim/inc/rack_elevation.html' with primary_face=rack.get_rear_elevation secondary_face=rack.get_front_elevation face_id=1 reserved_units=rack.get_reserved_units %}
+                            {% include 'dcim/inc/rack_elevation.html' with face=1 %}
```
jeremystretch: Should be `face='front'` and `face='rear'`

netbox/dcim/api/serializers.py
```diff
@@ -171,6 +171,16 @@ class Meta:
         fields = ['id', 'rack', 'units', 'created', 'user', 'tenant', 'description']
 
 
+class RackElevationDetailFilterSerializer(serializers.Serializer):
+    face = serializers.ChoiceField(choices=['front', 'rear'], default='front')
```
jeremystretch: `choices` and `default` should reference `dcim.choices.DeviceFaceChoices`. We should never be setting static choice values directly on fields.

netbox/dcim/api/serializers.py
```diff
@@ -171,6 +171,16 @@ class Meta:
...
+    render_format = serializers.ChoiceField(choices=['json', 'svg'], default='json')
```
jeremystretch: Same as above. The choices should be defined in a ChoiceSet.

netbox/dcim/api/serializers.py
```diff
@@ -171,6 +171,16 @@ class Meta:
...
+    width = serializers.IntegerField(default=230)
```
jeremystretch: This conflicts with `RackFilter.width`. As I mentioned earlier it would be good to use `unit_width` and `unit_height` to make obvious the relationship between these two fields.

netbox/dcim/api/views.py
```diff
@@ -200,6 +203,39 @@ def units(self, request, pk=None):
...
+            return HttpResponse(drawing.tostring(), content_type='image/svg+xml')
+
+        else:
+            elevation = rack.get_rack_units(
+                face=data['face'],
+                exclude=data['exclude'],
+                expand_devices=data['expand_devices']
+            )
+            if data['q']:
+                elevation = [u for u in elevation if data['q'] in str(u['id'])]
```
jeremystretch: I don't understand the purpose of this. It seems to be matching `q` against the numeric unit number, which doesn't seem useful. I'm not sure why we would ever return a partial elevation anyway. I could be misunderstanding.

jeremystretch (changes requested):
There are some minor issues I've flagged, and some items might need further discussion.

netbox/dcim/api/views.py
jeremystretch: As @lampwins pointed out, this behavior is present in the current `units` endpoint, apparently for filtering by unit ID. I don't believe this functionality is used any longer.

jeremystretch (approved):

PR #3768, opened 2019-12-14
Change gunicorn.conf remains to gunicorn.py

markkuleinio:


DanSheps:
Thanks for catching this

PR #3775, opened 2019-12-17
Add request to Custom Script run, if receiver supports it

### LINKED ISSUE #3705: Make current user available in Custom Scripts
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.7

### Proposed Functionality

Make the currently logged in user available to Custom Scripts. It may be beneficial to make the entire `request` context available, but all I need currently is the user.

### Use Case

In my use case, we have a policy of commenting on Prefixes with the user who allocated and assigned the prefix to a site. I'm attempting to automate this process using a Custom Script, but currently need to ask the user for their username, despite them being logged in.

### Database Changes

None

### External Dependencies

None

steffann:
### Fixes: #3705 

- Check whether the script.run() method accepts `request` as a keyword argument
- If it does, pass the current request to the method

### DIFF
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -235,6 +235,9 @@ class BaseScript:
         # Initiate the log
         self.log = []
 
+        # Declare the placeholder for the current request
+        self.request = None
+
         # Grab some info about the script
         self.filename = inspect.getfile(self.__class__)
         self.source = inspect.getsource(self.__class__)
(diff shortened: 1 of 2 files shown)


steffann:
I implemented this as a keyword-only option to prevent any clashes with existing scripts that already have a `run()` method with multiple arguments.

jeremystretch:
Rather than altering the signature of `run()`, my plan was to make the current request available as a context in the Script object. For example, the script author would access `self.request.user` to get the current user (if any). This avoids the need to adapt existing scripts, and allows for cleanly running scripts without a request present (e.g. when executed locally via cron job).

steffann:
No problem, I considered both approaches but wasn't sure what the lifetime of a Script object was, and I didn't want to keep a long-lived reference to the request hanging around in a property. If that's not a problem then I agree your approach is cleaner.

I'll adapt the implementation!

jeremystretch:
@steffann Can you re-base this against the `develop` branch please? I don't think there's any reason to wait for v2.7 as it won't be a disruptive change.

PR #3793, opened 2019-12-25
Closes #3663: add Filter Tests

### LINKED ISSUE #3663: API filter by created, last_updated

struppinet:
### Amends: #3680

added the wished Filter Tests

### DIFF
--- a/netbox/extras/tests/test_api.py
+++ b/netbox/extras/tests/test_api.py
@@ -1,11 +1,11 @@
-import time
 import datetime
 
 from django.contrib.contenttypes.models import ContentType
 from django.urls import reverse
+from django.utils import timezone
 from rest_framework import status
 
-from dcim.models import Device, DeviceRole, DeviceType, Manufacturer, Platform, Region, Site, RackGroup, RackRole, Rack
+from dcim.models import Device, DeviceRole, DeviceType, Manufacturer, Platform, Rack, RackGroup, RackRole, Region, Site
 from extras.constants import GRAPH_TYPE_SITE
 from extras.models import ConfigContext, Graph, ExportTemplate, Tag
 from tenancy.models import Tenant, TenantGroup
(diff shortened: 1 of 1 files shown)


DanSheps:
1. Build is failing
2. No open issue, what is this currently fixing?

struppinet:
lampwins wished for tests for this (https://github.com/netbox-community/netbox/pull/3680) so I added them.

netbox/extras/tests/test_api.py
```diff
@@ -520,3 +523,76 @@ def test_render_configcontext_for_object(self):
...
+        self.rackrole1 = RackRole.objects.create(name='Test Rack Role 1', slug='test-rack-role-1', color='ff0000')
+        self.rack1 = Rack.objects.create(
+            site=self.site1, group=self.rackgroup1, role=self.rackrole1, name='Test Rack 1', u_height=42,
+        )
+        self.rack2 = Rack.objects.create(
+            site=self.site1, group=self.rackgroup1, role=self.rackrole1, name='Test Rack 2', u_height=42,
+        )
+
+        # change the created and last_updated of one
+        time.sleep(2)  # to have a difference on the last_updated
```
jeremystretch: Because tests are run on every merge, they need to execute as fast as reasonably possible. Using `sleep()` here introduces 14 seconds of unnecessary delay (two seconds for each of the seven individual test methods). It also doesn't make sense because you're manually setting an arbitrary create time in the next line.

Instead, you can simply use the `update()` queryset method to assign arbitrary datetimes for the objects' `created` and `last_updated` fields.

netbox/extras/tests/test_api.py
```diff
@@ -520,3 +523,76 @@ def test_render_configcontext_for_object(self):
...
+        self.rack2.created = datetime.datetime(2001, 2, 3)
+        self.rack2.save()
+
+    def test_get_rack_meta(self):
```
jeremystretch: This method should be omitted from this test case as it does not test a filter.

netbox/extras/tests/test_api.py
```diff
@@ -520,3 +523,76 @@ def test_render_configcontext_for_object(self):
...
+        response = self.client.get(url, **self.header)
+
+        self.assertEqual(response.data['created'], self.rack1.created.strftime("%Y-%m-%d"))
+        self.assertEqual(response.data['last_updated'], self.rack1.last_updated.strftime("%Y-%m-%dT%H:%M:%S.%fZ"))
+
+    def test_get_rack_created(self):
+        url = reverse('dcim-api:rack-list')
+        response = self.client.get('{}?created=2001-02-03'.format(url), **self.header)
+
+        self.assertEqual(response.data['count'], 1)
```
jeremystretch: We should also verify that the single rack returned is the rack we're expecting, by asserting that the rack's ID is equal to `self.rack2.pk`. (This applies to each of the methods below as well.)

netbox/extras/tests/test_api.py
```diff
@@ -1,8 +1,11 @@
+import time
+import datetime
+
 from django.contrib.contenttypes.models import ContentType
 from django.urls import reverse
 from rest_framework import status
 
-from dcim.models import Device, DeviceRole, DeviceType, Manufacturer, Platform, Region, Site
+from dcim.models import Device, DeviceRole, DeviceType, Manufacturer, Platform, Region, Site, RackGroup, RackRole, Rack
```
jeremystretch: Minor nit, but imported models should be in alphabetical order.

jeremystretch (changes requested):

jeremystretch:
I've tweaked this to indicate that it amends the work contributed under #3680. Please also see my feedback above.

struppinet:
Feedback has been applied. (tried save and failed on the automatic updated last_updated. didn't know about the update() method)

jeremystretch:
> didn't know about the update() method)

You generally want to avoid using `update()` as it bypasses most validation, but it can be very helpful in niche cases like this.

PR #3806, opened 2019-12-29
Fixes 3805

### LINKED ISSUE #3805: Upgrade Select2
### Proposed Changes
Upgrade the Select2 library to the latest patch (4.0.12)

### Justification
Mainly for https://github.com/select2/select2/pull/5222 (keep select open if the CTRL/CMD key is held down, which lowers the number of queries to the server by not reloading the available options then force-scrolling after the query finishes – gets proportionately more annoying the higher the latency), but also to keep up with all of the improvements.

hSaria:
### Fixes: #3805
Bumped Select2 library to version 4.0.12. It's only a few files that have changed, but because the directory was updated, the diff is pretty long.


hSaria:
Static libraries bumped on minor version releases (i.e. v2.7). PR no longer needed.

PR #3814, opened 2020-01-01
Fixes #3589: Interface VLAN filtering

### LINKED ISSUE #3589: Missing error message when Tagged VLAN with incompatible Site is set
### Environment
* Python version: Python version: 3.6.8
* NetBox version: NetBox version: 2.6.5

### Steps to Reproduce
1. Create VLAN 123 associated with Site AAA
2. Create Device at Site BBB and Interface
3. Edit interface and set 802.1Q Mode to Tagged
4. Select VLAN 123 as Tagged VLAN
5. Apply

### Expected Behavior
An Error message.

### Observed Behavior
No error message, and the VLAN does not get set.

### Comment
A error message is correctly produced if the Mode is Access and the VLAN is selected in the Untagged input field instead. Then we get _"The untagged VLAN (123) must belong to the same site as the interface's parent device/VM, or it must be global"_. The same message should be show there too.

hSaria:
### Fixes: #3589

* Perform validation on the tagged VLANs similar to that which is already performed on the untagged VLAN
* Removes the legacy choices filter that is no longer used because the API-based select widgets override the choices
* Limit the select filter to only global VLANs or VLANs in the same site as the device

> Because there currently isn't a way to specify multiple values for the same parameter, I had to work around it.

### DIFF
--- a/netbox/dcim/forms.py
+++ b/netbox/dcim/forms.py
@@ -2236,36 +2236,6 @@ class InterfaceForm(InterfaceCommonForm, BootstrapMixin, forms.ModelForm):
                 device__in=[self.instance.device, self.instance.device.get_vc_master()], type=IFACE_TYPE_LAG
             )
 
-        # Limit VLan choices to those in: global vlans, global groups, the current site's group, the current site
-        vlan_choices = []
-        global_vlans = VLAN.objects.filter(site=None, group=None)
-        vlan_choices.append(
-            ('Global', [(vlan.pk, vlan) for vlan in global_vlans])
-        )
-        for group in VLANGroup.objects.filter(site=None):
-            global_group_vlans = VLAN.objects.filter(group=group)
-            vlan_choices.append(
-                (group.name, [(vlan.pk, vlan) for vlan in global_group_vlans])
-            )
-
-        site = getattr(self.instance.parent, 'site', None)
-        if site is not None:
-
-            # Add non-grouped site VLANs
-            site_vlans = VLAN.objects.filter(site=site, group=None)
-            vlan_choices.append((site.name, [(vlan.pk, vlan) for vlan in site_vlans]))
-
-            # Add grouped site VLANs
-            for group in VLANGroup.objects.filter(site=site):
-                site_group_vlans = VLAN.objects.filter(group=group)
-                vlan_choices.append((
-                    '{} / {}'.format(group.site.name, group.name),
-                    [(vlan.pk, vlan) for vlan in site_group_vlans]
-                ))
-
-        self.fields['untagged_vlan'].choices = [(None, '---------')] + vlan_choices
-        self.fields['tagged_vlans'].choices = vlan_choices
-
 
 class InterfaceCreateForm(InterfaceCommonForm, ComponentForm, forms.Form):
     name_pattern = ExpandableNameField(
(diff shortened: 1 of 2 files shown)


netbox/dcim/forms.py
```diff
@@ -74,6 +74,17 @@ def clean(self):
         elif self.cleaned_data['mode'] == IFACE_MODE_TAGGED_ALL:
             self.cleaned_data['tagged_vlans'] = []
 
+        # Validate tagged VLANs; must be a global VLAN or in the same site
+        else:
```
jeremystretch:
```suggestion
        elif self.cleaned_data['mode'] == IFACE_MODE_TAGGED:
```
No need to check for this on access ports with no tagged VLANs.

netbox/dcim/forms.py
```diff
@@ -2227,35 +2244,9 @@ def __init__(self, *args, **kwargs):
...
-                site_group_vlans = VLAN.objects.filter(group=group)
-                vlan_choices.append((
-                    '{} / {}'.format(group.site.name, group.name),
-                    [(vlan.pk, vlan) for vlan in site_group_vlans]
-                ))
-
-        self.fields['untagged_vlan'].choices = [(None, '---------')] + vlan_choices
-        self.fields['tagged_vlans'].choices = vlan_choices
+        # Add the current site to the list of filtered VLANs
+        self.fields['untagged_vlan'].widget.attrs['1-data-additional-query-param-site_id'] = self.instance.device.site.pk
```
jeremystretch:
```suggestion
        self.fields['untagged_vlan'].widget.add_additional_query_param('site_id', self.instance.device.site.pk)
```
The widget has convenience methods for more cleanly applying these attributes.

netbox/dcim/forms.py
```diff
@@ -2227,35 +2244,9 @@ def __init__(self, *args, **kwargs):
...
+        self.fields['tagged_vlans'].widget.attrs['1-data-additional-query-param-site_id'] = self.instance.device.site.pk
```
jeremystretch:
```suggestion
        self.fields['tagged_vlans'].widget.add_additional_query_param('site_id', self.instance.device.site.pk)
```

netbox/dcim/forms.py
```diff
@@ -2337,35 +2334,9 @@ def __init__(self, *args, **kwargs):
...
-                site_group_vlans = VLAN.objects.filter(group=group)
-                vlan_choices.append((
-                    '{} / {}'.format(group.site.name, group.name),
-                    [(vlan.pk, vlan) for vlan in site_group_vlans]
-                ))
-
-        self.fields['untagged_vlan'].choices = [(None, '---------')] + vlan_choices
-        self.fields['tagged_vlans'].choices = vlan_choices
+        # Add the current site to the list of filtered VLANs
+        self.fields['untagged_vlan'].widget.attrs['1-data-additional-query-param-site_id'] = self.parent.site.pk
```
jeremystretch:
```suggestion
        self.fields['untagged_vlan'].widget.add_additional_query_param('site_id', self.parent.site.pk)
```

netbox/dcim/forms.py
```diff
@@ -2337,35 +2334,9 @@ def __init__(self, *args, **kwargs):
...
+        self.fields['tagged_vlans'].widget.attrs['1-data-additional-query-param-site_id'] = self.parent.site.pk
```
jeremystretch:
```suggestion
        self.fields['tagged_vlans'].widget.add_additional_query_param('site_id', self.parent.site.pk)
```

netbox/dcim/forms.py
```diff
@@ -2449,35 +2426,9 @@ def __init__(self, *args, **kwargs):
...
-                    site_group_vlans = VLAN.objects.filter(group=group)
-                    vlan_choices.append((
-                        '{} / {}'.format(group.site.name, group.name),
-                        [(vlan.pk, vlan) for vlan in site_group_vlans]
-                    ))
-
-        self.fields['untagged_vlan'].choices = [(None, '---------')] + vlan_choices
-        self.fields['tagged_vlans'].choices = vlan_choices
+        # Add the current site to the list of filtered VLANs
+        self.fields['untagged_vlan'].widget.attrs['1-data-additional-query-param-site_id'] = self.parent_obj.site.pk
```
jeremystretch:
```suggestion
        self.fields['untagged_vlan'].widget.add_additional_query_param('site_id', self.parent_obj.site.pk)
```

netbox/dcim/forms.py
```diff
@@ -2449,35 +2426,9 @@ def __init__(self, *args, **kwargs):
...
+        self.fields['tagged_vlans'].widget.attrs['1-data-additional-query-param-site_id'] = self.parent_obj.site.pk
```
jeremystretch:
```suggestion
        self.fields['tagged_vlans'].widget.add_additional_query_param('site_id', self.parent_obj.site.pk)
```

jeremystretch (changes requested):

netbox/dcim/forms.py
hSaria: This wouldn't work because it will override

```
additional_query_params={
    'site_id': 'null'
},
```

hSaria:
@jeremystretch FYI, this is currently waiting on your input regarding multiple values for the same field. The current implementation will override the field I've put in the `additional_query_params` of the widget.

I'm hoping there's a better way, but I can think of only two ways to do it:
* Key-based: use a different key which will map to the same field once the code in `forms.js` processes it (what I did, but perhaps in a better way than just prepending `1-`)
* Value-based: Use some character to indicate multiple values in the field (e.g. `add_additional_query_param('site_id', 'null,{}'.format(self.parent_obj.site.pk))`) and have `forms.js` split on that character.

hSaria:
I have removed the select widget filtering changes from this PR as they will be covered in a different PR for #3840.

This PR just adds validation for the tagged VLANs of an interface.

jeremystretch (approved):

PR #3817, opened 2020-01-01
Closes #2988

### LINKED ISSUE #2988: LDAP authentication documentation: add paragraph about group creation required.
### Change Type
[X] Addition
[ ] Correction
[ ] Deprecation

### Proposed Changes
On [the netbox ldap configuration page](https://netbox.readthedocs.io/en/stable/installation/4-ldap/) it would be great to explain that the three users groups (is_active, is_staff, is_superuser) must exist in the LDAP directory otherwise authentication will fail while trying to retrieve the user group mapping.

Reason; Netbox is trying to retrieve permissions using an LDAP compareRequest, however if one of the group is not present in the LDAP directory, the LDAP server may reply with a success response instead of compareResponse. Because of the implementation of the Python ldap module, such a response would lead to an exception and authentication will fail.

hSaria:
### Fixes: #2988
Add a small warning that the LDAP groups must exist.

### DIFF
--- a/docs/installation/4-ldap.md
+++ b/docs/installation/4-ldap.md
@@ -80,6 +80,7 @@ AUTH_LDAP_USER_ATTR_MAP = {
 ```
 
 # User Groups for Permissions
+
 !!! info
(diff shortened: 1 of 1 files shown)


jeremystretch:
Wrong button :roll_eyes:

Issue #3294, opened 2019-06-24
Introduce a mechanism for storing individual user preferences


jeremystretch:
### Environment
* Python version: 3.6.8
* NetBox version: 2.6.0

### Proposed Functionality
Introduce a model for storing the preferences of an individual user. A simple model providing arbitrary JSON data storage associated one-to-one with user accounts is probably sufficient. This will allow a user to set a preferred item (e.g. the default status for new prefixes) that will persist indefinitely.

### Use Case
This has been highlighted as a requirement for several other features:

* #492 - Customized table column ordering
* #1426 - Customized default statuses

### Database Changes
Introduce a new model under the `users` app, e.g. `UserPreferences`.

### External Dependencies
None

tb-killa:
Maybe this could help? https://pypi.org/project/django-jsonstore/

PR #3821, opened 2020-01-02
Fixes #2598: Toggle for showing available prefixes/ip addresses

### LINKED ISSUE #2365: Toggle suggested/Available prefixes in Child Prefix tab
### Environment
* Python version:  3.4.3
* NetBox version: 2.4.1

### Proposed Functionality
Currently, when clicking on the Child Prefix tab of a given prefix, it shows 'Active' and 'Available' prefixes. A button to show/hide the 'Available' prefixes would be useful.

### Use Case

Daily prefix view

### LINKED ISSUE #2598: Toggle the display of available child prefixes
**Environment**
NetBox version: 2.4.6

**Proposed Functionality**
Under Child prefixes add a toggle button to view available prefixes or not.
Default option in my opinion should be to have the available prefix hidden.

If you add a 10.0.0.0/8 as prefix you will automatically get several hundreds of available lines when viewing it under child prefixes that interfere with the user experience of the GUI.

image

**Use Case**
User friendliness

**Database Changes**

**External Dependencies**

hSaria:
### Fixes: #2598

Adds a cookie-based toggle to showing/hiding available prefixes and IP addresses

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -2,7 +2,6 @@
 
 ## Enhancements
 
-* [#2589](https://github.com/netbox-community/netbox/issues/2589) - Toggle for showing available prefixes/ip addresses
 * [#2892](https://github.com/netbox-community/netbox/issues/2892) - Extend admin UI to allow deleting old report results
 * [#3062](https://github.com/netbox-community/netbox/issues/3062) - Add `assigned_to_interface` filter for IP addresses
 * [#3461](https://github.com/netbox-community/netbox/issues/3461) - Fail gracefully on custom link rendering exception
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -2,7 +2,7 @@
 
 ## Enhancements
 
-* [#2365](https://github.com/netbox-community/netbox/issues/2365) - Toggle for showing available prefixes/ip addresses
+* [#2589](https://github.com/netbox-community/netbox/issues/2589) - Toggle for showing available prefixes/ip addresses
 * [#2892](https://github.com/netbox-community/netbox/issues/2892) - Extend admin UI to allow deleting old report results
 * [#3062](https://github.com/netbox-community/netbox/issues/3062) - Add `assigned_to_interface` filter for IP addresses
 * [#3461](https://github.com/netbox-community/netbox/issues/3461) - Fail gracefully on custom link rendering exception
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -1,3 +1,9 @@
+# v2.6.12 (FUTURE)
+
+## Enhancements
+
+* [#2589](https://github.com/netbox-community/netbox/issues/2589) - Toggle for showing available prefixes/ip addresses
+
 # v2.6.11 (2020-01-03)
 
 ## Bug Fixes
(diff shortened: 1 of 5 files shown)


hSaria:
@jeremystretch I had the idea of creating a `{{ header_options }}` in the table template to allow this to be more dynamic. If you think we'll have more instances of table-based actions, it might be worth while. This particularly shines in the Aggregate page because the table is at the end of the page and the options are at the top. I'll leave it up to you if you want me to do something like that or just keeping where it is right now.

netbox/ipam/views.py
```diff
@@ -333,7 +333,14 @@ def get(self, request, pk):
         ).annotate_depth(
             limit=0
         )
-        child_prefixes = add_available_prefixes(aggregate.prefix, child_prefixes)
+
+        # Update the ipam_show_available cookie if request specifies it
+        if request.GET.get('show_available') is not None:
+            request.session['ipam_show_available'] = request.GET.get('show_available') == 'true'
```
jeremystretch: We need to put some development guidance in place before we start storing session data, specifically around naming convention and user experience expectations. (We've somehow avoided it thus far.) I'll raise it with the other maintainers to see what seems prudent. IMO we should hold off on leveraging session state until we have a mechanism for persistence across logins (see #3294).

netbox/templates/ipam/aggregate.html
```diff
@@ -40,6 +40,14 @@
     </div>
     <h1>{% block title %}{{ aggregate }}{% endblock %}</h1>
     {% include 'inc/created_updated.html' with obj=aggregate %}
+    {% if show_available is not None %}
```
jeremystretch: Can you move this code to a separate template (e.g. `ipam/inc/toggle_available.html`) and use an `{% include %}` tag to avoid the repetition?

jeremystretch (commented):
The button itself is fine. Going to ask the other maintainers for their thoughts on using session state.

netbox/ipam/views.py
hSaria: Yeah, I was just as surprised when I noticed that we've not been explicitly dealing with cookies before (ignoring the Django-based ones).

docs/release-notes/version-2.6.md
```diff
@@ -2,6 +2,7 @@
 
 ## Enhancements
 
+* [#2365](https://github.com/netbox-community/netbox/issues/2365) - Toggle for showing available prefixes/ip addresses
```
jeremystretch: I closed #2365 as a duplicate in favor of #2598. If you can tweak this great, otherwise I'll fix it during merge.

hSaria: Sure. I'll fix it along with the template thing.

netbox/ipam/views.py
hSaria: Should settings like these (i.e. visual) exist for users in the first place? Someone that is not logged in might also be interested in toggling the view.

With that said, I do agree that we need a developer interface for abstracting cookies and how they're handled, though I don't think it should rely on the `users` model.

jeremystretch:
@hSaria I discussed session state with the other maintainers today. Let's hold off on it until we have a formal plan for namespacing and documentation to manage user expectations (which should be covered by #3294).

If you can remove the session storage bits but leave the toggle button, I'm happy to merge this.

hSaria:
@jeremystretch, sure thing. Shall I leave the current default of listing the available prefixes or keep them hidden until someone explicitly requests them?

jeremystretch:
@hSaria Please leave them visible by default to preserve the current behavior.

hSaria:
Alright. It's ready.

PR #3823, opened 2020-01-02
Fixes #3822:  device bay edit

### LINKED ISSUE #3822: "'DeviceBayForm' has no field named 'installed_device'" when trying to edit device bay details with an installed device in it
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.9

### Steps to Reproduce
1) Create Device Type "Parent" with Parent/child status = Parent and add one device bay "Bay 1"
2) Create Device Type "Child" with Parent/child status = Child
3) Create Devices "Parent" and "Child"
4) Assign device "Child" to bay of device "Parent"
5) Navigate to device "Parent" and select the "Edit device bay" icon for "Bay 1"
6) Do not change anything in the form, simply click on "Update"

Error message appears:
2020-01-02 11_48_16-Server Error

### Expected Behavior
No error message when trying to update the details of an installed device in a device bay.

### Observed Behavior
Changing details of a device bay only seems to be possible as long as the device bay has a status of "Vacant". Once a device is installed in the device bay, the above error message appears.

hSaria:
### Fixes: #3822

The edit form for the device bay was missing `installed_device` as a hidden input field.

When fixing the above issue, I noticed that there was a logic issue in the model of `DeviceBay` in that it did not allow updates once created and a device is installed because of the `current_bay` of the installed device condition. This was fixed by allowing the update to go through if the current_bay is the bay to be updated, or, stated differently, fail if current_bay is not the bay to be updated.

> You could also exclude `self` from the query for `current_bay`. Not sure which is the better approach.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -14,6 +14,7 @@
 * [#3780](https://github.com/netbox-community/netbox/issues/3780) - Fix AttributeError exception in API docs
 * [#3809](https://github.com/netbox-community/netbox/issues/3809) - Filter platform by manufacturer when editing devices
 * [#3811](https://github.com/netbox-community/netbox/issues/3811) - Fix filtering of racks by group on device list
+* [#3822](https://github.com/netbox-community/netbox/issues/3822) - Fix failure to update device bay from edit page
 
 ---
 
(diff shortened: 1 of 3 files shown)


hSaria:
No longer needed; fixed in https://github.com/netbox-community/netbox/commit/395f23e1d3f50015b0dfe62481ed324e3b1e4025.

PR #3826, opened 2020-01-02
Fixes #3122: Select2 for device field

### LINKED ISSUE #3122: Use Select2 UI to filter cables by connected device
### Environment
* Python version:  3.5.3
* NetBox version:  2.5.11

### Proposed Functionality
Enhancing the user experience of the new filter added by #3023, allowing cables to be filtered by connected device, by implementing the Select2 User Interface introduced by #2772.

### Use Case
A user who wants to filter the /dcim/cables view by connected device, should be able to use the search-box input field labelled "Device" as a search tool by itself - similar to the "Device" input field used by the "Connect" dialog.

### Database Changes
None

### External Dependencies
Nothing new

hSaria:
### Fixes: #3122

I didn't put this in the changelog because it's nothing notable

### DIFF
--- a/netbox/dcim/forms.py
+++ b/netbox/dcim/forms.py
@@ -3171,9 +3171,12 @@ class CableFilterForm(BootstrapMixin, forms.Form):
         required=False,
         widget=ColorSelect()
     )
-    device = forms.CharField(
+    device = forms.ChoiceField(
         required=False,
-        label='Device name'
+        widget=APISelect(
+            api_url='/api/dcim/devices/',
+            value_field='text',
+        )
     )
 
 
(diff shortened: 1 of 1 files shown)


jeremystretch:
I've noticed an issue with this approach: After submitting the filter form, the object list comes back as expected however the field value is lost. This is due to searching by device name rather than ID. We should change `device` to `device_id` to remain consistent with the other filter forms/fields.

`device_id` exists on `CableFilter` and `InventoryItemFilter` however it needs to be added to the other three filter sets. I can take care of this in a separate FR.

jeremystretch:
Added the necessary `device_id` filters under #3827.

hSaria:
Awesome! Thank you. I’ll update the PR to work with that.

hSaria:
The PR is ready now. I've also fixed up the Site select widgets for the 3 "_something_ Connections".

PR #3838, opened 2020-01-03
Fixes #3090: Interface filtering

### LINKED ISSUE #3090: Allow filtering interfaces by type on device view
### Environment
* Python version:  3.5.6
* NetBox version:  2.5.10

### Proposed Functionality
Add a dropdown menu at the top of the device interfaces tables (next to "show IPs") to allow filtering interfaces by form factor/type. ~We don't have any sufficient UI framework in place yet to do this dynamically, so for now we're looking at a full page refresh: A link to something like `/dcim/devices/123/?interface_type=800`.~ A search box with some jQuery to display only interfaces with a matching name should suffice.

### Use Case
Some switches, especially large chassis switches, can have hundreds of interfaces. Sometimes a user may want to edit all interfaces of a certain type in bulk. Currently, this requires selecting each interface individually.

### Database Changes
None

### External Dependencies
None

hSaria:
### Fixes: #3090

Add a filter for device interfaces that updates the selection as changed. It is RegEx based just because it is convenient for actions that span multiple chassises, like creating an MC-LAG.

interface-filter

### DIFF
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -557,7 +557,7 @@
                             </button>
                         </div>
                         <div class="col-md-2 pull-right noprint">
-                            <input class="form-control interface-filter" type="text" placeholder="Filter (RegExp)" style="height: 24px" />
+                            <input class="form-control interface-filter" type="text" placeholder="Filter (RegExp)" style="height: 23px" />
                         </div>
                     </div>
                     <table id="interfaces_table" class="table table-hover table-headings panel-body component-list">
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -557,7 +557,7 @@
                             </button>
                         </div>
                         <div class="col-md-2 pull-right noprint">
-                            <input class="form-control interface-filter" type="text" placeholder="Filter (RegExp)" style="height: 23px" />
+                            <input class="form-control interface-filter" type="text" placeholder="Filter" title="RegEx-enabled" style="height: 23px" />
                         </div>
                     </div>
                     <table id="interfaces_table" class="table table-hover table-headings panel-body component-list">
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -903,35 +903,8 @@ function toggleConnection(elem) {
 $(".cable-toggle").click(function() {
     return toggleConnection($(this));
 });
-// Toggle the display of IP addresses under interfaces
-$('button.toggle-ips').click(function() {
-    var selected = $(this).attr('selected');
-    if (selected) {
-        $('#interfaces_table tr.ipaddresses').hide();
-    } else {
-        $('#interfaces_table tr.ipaddresses').show();
-    }
-    $(this).attr('selected', !selected);
-    $(this).children('span').toggleClass('glyphicon-check glyphicon-unchecked');
-    return false;
-});
-$('input.interface-filter').on('input', function() {
-    var filter = new RegExp(this.value);
-
-    for (interface of $(this).closest('form').find('tbody > tr')) {
-        // Slice off 'interface_' at the start of the ID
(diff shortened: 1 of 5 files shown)


netbox/templates/dcim/device.html
```diff
@@ -556,6 +556,9 @@ <h1>{{ device }}</h1>
                                 <span class="glyphicon glyphicon-check" aria-hidden="true"></span> Show IPs
                             </button>
                         </div>
+                        <div class="col-md-2 pull-right noprint">
+                            <input class="form-control interface-filter" type="text" placeholder="Filter (RegExp)" style="height: 23px" />
```
jeremystretch: IMO we should leave the placeholder text as "Filter" since some people won't know what RegExp refers to. It might be nice to add a little question mark icon (`fa-question-circle`) next to the filter field that would pop up a little explanation, but not really necessary.

netbox/templates/virtualization/virtualmachine.html
```diff
@@ -325,5 +328,21 @@ <h1>{% block title %}{{ virtualmachine }}{% endblock %}</h1>
     $(this).children('span').toggleClass('glyphicon-check glyphicon-unchecked');
     return false;
 });
+$('input.interface-filter').on('input', function() {
```
jeremystretch: Can you move this code somewhere so that it can be referenced by either template without being repeated? Maybe `templates/dcim/inc/interface_toggles.js`? (The "toggle IP" bit above it can be moved too.) I just know that if we keep two separate copies of the code we'll inevitably forget to update one at some point.

jeremystretch (changes requested):
Awesome work! One minor issue I noticed: If I filter interfaces and _then_ check the "select all" checkbox, it will also select the hidden interfaces. I think we just need to tweak [this line](https://github.com/netbox-community/netbox/blob/develop/netbox/project-static/js/forms.js#L10) in `forms.js` to select only visible items.

hSaria:
> I think we just need to tweak [this line](https://github.com/netbox-community/netbox/blob/develop/netbox/project-static/js/forms.js#L10) in `forms.js` to select only visible items.

Agreed. My original code added `:visible` to the selector but then I opted against it figuring I'd avoid modifying existing code, but I suppose it makes sense in all cases as the user would want to select things that he/she are seeing.

> It might be nice to add a little question mark icon (fa-question-circle) next to the filter field that would pop up a little explanation

Good point. I'll use the jQuery `.tooltip()` as it works pretty well with Bootstrap, like so
```html
<i class="fa fa-info-circle" data-toggle="tooltip" title="Sup!"></i>
```
```javascript
$('[data-toggle="tooltip"]').tooltip(); 
```

(I used `fa-info-circle` because the comments section uses it for Markdown note.)

> I just know that if we keep two separate copies of the code we'll inevitably forget to update one at some point.

You read my mind; I had the same idea when I was copying the code over and wanted to use a template, but I hesitated because I remembered the first line the [style guide](https://netbox.readthedocs.io/en/stable/development/style-guide/#general-guidance) regarding consistency (it did trigger my consistent-but-consistently-wrong alarm, but I suppressed it).

netbox/templates/virtualization/virtualmachine.html
hSaria: For the sake of performance (caching), I'm gonna put it as a static `.js` file.

hSaria:
I've fixed up everything, but I couldn't add an icon next to the Filter in a clean way so I just used to title attribute (if you hover long enough, you'll know)

This is due to the fact that we're using size `xs` on those additional options in the table header but there doesn't exist an `xs` input class (smallest is `sm`). I could add an `xs` class to `base.css`, but that would make the transition to Bootstrap 4 even more difficult.

I could also just take out the RegEx note all together and let people figure it out; those who know RegEx will give it a try and – much to their delight – know of its support.

I have no preference here, so you let me know what you think is best.

jeremystretch (approved):

PR #3844, opened 2020-01-03
Develop 2.7 - New Connection Types

### LINKED ISSUE #3841: Add California Style Power connection types
### Environment
* Python version: 3.6.3
* NetBox version: 2.7

### Proposed Functionality
California Style Connectors seem to be missing from the current power_port and power_outlet types.  

### Use Case
Many PDUs use these connector type.

### Database Changes
Add a new data model type with the fields mentioned in the proposed functionality.

### External Dependencies
N/A

### LINKED ISSUE #3842: Add 802.11ax wireless connection type
### Environment
* Python version: 3.6.3
* NetBox version: 2.7

### Proposed Functionality
802.11ax seems to be missing from the wireless interface connection types.  

### Use Case
New wifi6

### Database Changes
NA

### External Dependencies
N/A

### LINKED ISSUE #3843: Add RJ-12 console connection type
### Environment
* Python version: 3.6.3
* NetBox version: 2.7

### Proposed Functionality
RJ-12 seems to be missing from the console/serial interface connection types.  

### Use Case
Needed mostly for PDUs

### Database Changes
NA

### External Dependencies
N/A

ryanmerolle:
Fixes #3841 (PowerPort & PowerOutlet)
Fixes #3842 
Fixes #3843


hSaria:
I don't think migration are supposed to be modified.

ryanmerolle:
For wireless I just added 802.11ax wherever I saw 802.11ad, and that included the migrations.  I can remove.

DanSheps:
You should not directly modify previous migrations.

Instead, you should make the changes to trigger a migration and run manage.py makemigrations.

ryanmerolle:
Noob contribution on my part.  I can remove, I just figured it would be needed for any where I saw similar wireless choices.  How should I proceed?

I guess I did not also fully understand the migrations.

DanSheps:
@ryanmerolle You would revert the changes to the migration files then run the python command:

python manage.py makemigrations

This will generate a new migration file, if required, that has the changes.

DanSheps:
If the migration is not quite right, you can delete the migration file and re-create it with the same command.

jeremystretch:
Addressing three different FRs requires three different pull requests. I realize this may seem excessive, but experience has proven that trying to batch changes leads to problems and confusion.

> You should not directly modify previous migrations.

This is generally correct. However, choice fields are a bit odd given that the resulting migration has no real effect on the underlying database scheme: it merely tracks changes to the Django model field. I have in the past merged these changes into prior migrations, but doing so takes considerable forethought and attention to detail, and I wouldn't think it fair to expect this of anyone who is not intimately familiar with NetBox's migration history.

So as far as PRs are concerned, it's best to submit each with its own discrete migration. I'm happy to condense them as the opportunity arises prior to a release.

@ryanmerolle I'm going to close out this PR to avoid confusion; could you please submit the changes proposed here as separate PRs?

PR #3846, opened 2020-01-04
Fixes #2050: Image preview

### LINKED ISSUE #2050: Images preview
### Issue type
[X] Feature request 
[ ] Bug report      
[ ] Documentation   

### Environment
* Python version:  3.5.4
* NetBox version:  2.3.2

### Description
We used to attach a variety of picture (setup, custumers install proof etc ...) to each rack, and it's not too easy to find the good one.
It could be very nice to display a gallery of preview instead of just a table with the name/size/date.

hSaria:
### Fixes: #2050

Image preview when hovering over an attachment's link. The preview is displayed only if the image can be rendered. 

I've avoided using a library for this because it seemed they either require additional supporting code or are too overkill for the task, and also didn't want to introduce another dependency.

It works fine even when loading is taking a long time or the image in unavailable; I've put a 1s delay to my VM to test and the image would appear once loaded and is hidden until then.

Might get a PR conflict regarding the changelog; I can change that once the v2.6.12 headers are in.

image-preview

> Cat not included with the PR. Cat sold separately.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -1,3 +1,11 @@
+# v2.6.12 (FUTURE)
+
+## Enhancements
+
+* [#2050](https://github.com/netbox-community/netbox/issues/2050) - Preview image attachments when hovering the link
+
+---
+
 # v2.6.11 (2020-01-03)
 
 ## Bug Fixes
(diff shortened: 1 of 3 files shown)


hSaria:
@jeremystretch Thanks for fixing the conflict. For future reference, do you prefer that I write the changelog myself or leave it to you? I'm fine with either.

jeremystretch:
@hSaria Good to include it if you remember; if not don't worry about it.

jeremystretch:
10/10 image, would preview again.

PR #3847, opened 2020-01-05
Fixes #3525: Filter muiltiple ipaddress terms

### LINKED ISSUE #3525: Allow multiple search terms when searching for IP addresses
### Environment
* Python version: Python 3.6.8
* NetBox version:  NetBox 2.6.3

### Proposed Functionality

It would be useful to perform a bulk attribute update on a range of IP's within a subnet. For example, adding (or removing) tag `tester-tom` to addresses `10.20.102.[50-100]`. Right now the IP search is string based (#1620), so searching for `10.20.102.5` will find addresses `10.20.102.5,[10.20.102.50-10.20.102.59], [10.20.102.500-599]`. In this scenario, to assign hosts `10.20.102.50/24-10.20.102.100/24` to Tester Tom, we have to perform five searches: `10.20.102.[5,6,7,8,9]`: for each search, select all addresses matching the query -> Edit All -> Add/Remove Tags -> Save.  One suggestion is to allow the range syntax that is supported on the IP bulk add UI: i.e. `10.20.102.[50-100]`.  Yet another suggestion, allow for multiple comma-separated search strings e.g.: `10.20.102.5,10.20.102.6`.

### Use Case

We oftentimes designate blocks of IP addresses within a subnet for certain purposes. For example, within a given subnet we may assign fifty contiguous addresses for use by a particular developer/tester, or for exclusive use by VMs on a given host system. We plan on "assigning" the IPs using tags, as I don't see a concept of IP "ownership" within netbox (although maybe I can use the custom data feature). These designations change fairly rapidly, so the multiple/search update workflow will become very tedious.

This request seems similar to #1410 (but is different), and is somewhat related to #834

kobayashi:
### Fixes: #3525 

This makes it possible to filter by mpultiple address terms for ipaddress even w/ or w/o subnet.

```
$ curl -X GET -H "Accept: application/json; indent=4" -H "Authorization: Token 0123456789abcdef0123456789abcdef01234567" "http://127.0.0.1:8000/api/ipam/ip-addresses/?address=192.168.1.1/24&address=192.168.1.2/24&address=192.168.1.3/24" 
{
    "count": 3,
    "next": null,
    "previous": null, 
    "results": [...]
}
```


jeremystretch:
It seems like we're mixing up two different function here:

1. Enable querying for multiple `address` values; e.g. `?address=192.0.2.1&address=192.0.2.2`
2. Enable querying for all addresses within a given prefix, e.g. `?address=192.0.2.0/24`

#3525 was scoped specifically to the first function; any other feature would need a new FR first. (However, I think it would make sense to extend this same behavior to the `prefix` field on `PrefixFilter` for consistency.)

Regarding the second function, this would need to be provided via a separate filter, as filtering by `address` implies that you are searching for a specific address. For example, a user might not expect a query for `address=192.0.2.1/24` to return anything but that one specific IP.

`PrefixFilter` is a good reference for this: `prefix` performs an exact match, whereas `within` and `within_include` allowing searching by a given containing prefix. IMO it would make sense to introduce a `within` filter for the IPAddress model (but again, that would be a separate FR).

kobayashi:
The PR does not provide prefix filter function. This change extends the current ipaddress filtering method.
If we make a query like `address=192.0.2.1/24&address=192.0.2.2`, it returnes a specific 2 items.

netbox/ipam/fields.py
```diff
@@ -32,6 +32,8 @@ def to_python(self, value):
     def get_prep_value(self, value):
         if not value:
             return None
+        if isinstance(value, list):
+            return [str(self.to_python(v)).split('/')[0] for v in value]
```
jeremystretch: I believe this is causing the filter to ignore prefix lengths. For example, assume I have an IP defined as 192.0.2.1/24. In the current release, both `address=192.0.2.1` and `address=192.0.2.1/24` will match this. However, `address=192.0.2.1/25` will not.

We need to preserve the ability to match the mask length, if specified.

jeremystretch (changes requested):
I just added a whole bunch of tests for filtersets. Can you please expand the test case `IPAddressFilter` for to test the new multi-value behavior? Per my comment inline, make sure that it checks three cases for mask length: matching, not matching, and not specified.

kobayashi:
fixes issue of filtering ipaddress with subnet and adds ipaddress filter tests.

netbox/ipam/fields.py
kobayashi: Solved this by the latest push.

jeremystretch (approved):

PR #3860, opened 2020-01-08
Fixes #1982: Swagger NAPALM documentation

### LINKED ISSUE #1982: Enhance SwaggerUI to support NAPALM methods
### Issue type
[X] Feature request

### Environment
* Python version:  3.6
* NetBox version:  2.3.1

### Description
I know how to use the NetBox API directly from the CLI to get info from devices using `NAPALM` (and the works great!).  However, I can't do the same thing in the SwaggerUI and I'm hoping a small change could enable me to do that.

What I'm trying to accomplish is to programatically query the devices I have in NetBox (via `NAPALM`) to see if their `model` matches the `device_type` (or specifically `part_number`) that I have defined in NetBox.  

From what I can tell, `pynetbox` doesn't have this functionalty so I'm trying to use `netbox-pyswagger` (from @jeremyschulman) to accomplish this.  However, that "client" is solely based on the output gathered from `/api/docs/?format=openapi`.

To clarify, under the `dcim` section of Swagger, there is this option:
```
GET /api/dcim/devices/{id}/napalm/
```
And when you expand it, you can only pass the `id` field.

After entering an `id` there and clicking the "Try It Out!" button, I can confirm (with command logging on the device in question) that `NAPALM` **does** connect just fine, but doesn't run anything else (due to not be told otherwise) and then disconnects.

The, the response seen in Swagger is the following:
```
Response Body
{}

Response Code
200
```

I suppose what I'm looking for is the option to pass other `method`s (via Swagger) to get the same information as I can using the CLI.  I hope that makes sense.  If not, let me know and I can try to explain further.

hSaria:
### Fixes: #1982 <ISSUE NUMBER GOES HERE>

The NAPALM endpoint in Swagger is now documented. In addition, a call with no methods will raise an exception.

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -374,9 +374,6 @@ class DeviceViewSet(CustomFieldModelViewSet):
         """
         Execute a NAPALM method on a Device
         """
-        if not request.GET.get('method'):
-            raise ServiceUnavailable('No NAPALM methods were specified.')
-
         device = get_object_or_404(Device, pk=pk)
         if not device.primary_ip:
             raise ServiceUnavailable("This device does not have a primary IP address configured.")
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -327,13 +327,6 @@ class DeviceViewSet(CustomFieldModelViewSet):
     )
     filterset_class = filters.DeviceFilter
 
-    _method = Parameter(
-        name='method',
-        in_='query',
-        required=True,
-        type=openapi.TYPE_STRING
-    )
-
     def get_serializer_class(self):
         """
         Select the specific serializer based on the request context.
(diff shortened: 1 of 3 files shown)


netbox/dcim/api/views.py
```diff
@@ -327,6 +327,13 @@ class DeviceViewSet(CustomFieldModelViewSet):
     )
     filterset_class = filters.DeviceFilter
 
+    _method = Parameter(
```
jeremystretch: Let's avoid using a generic variable name on DeviceView for something that's specific to the NAPALM endpoint. Honestly, I'd be okay with just defining the Parameter list directly inside the `@swagger_auto_schema()` decorator.

netbox/dcim/api/views.py
```diff
@@ -358,11 +365,15 @@ def graphs(self, request, pk):
 
         return Response(serializer.data)
 
+    @swagger_auto_schema(manual_parameters=[_method], responses={'200': serializers.DeviceNAPALMSerializer})
     @action(detail=True, url_path='napalm')
     def napalm(self, request, pk):
         """
         Execute a NAPALM method on a Device
         """
+        if not request.GET.get('method'):
```
jeremystretch: I see why you did this, but this is technically out of scope for #1982 as it modifies the endpoint behavior. It needs a separate issue; I'd argue for a bug report (rather than a FR).

I say this because in a scenario where a user has a script making malformed calls to the NAPALM endpoint, this change might cause it to start throwing errors. We want to make sure the user can quickly discern the suspect reason from reviewing the release notes.

jeremystretch (changes requested):

netbox/dcim/api/views.py
hSaria: Moved parameter to decorator

netbox/dcim/api/views.py
hSaria: Took it out. To be honest, I'm not at all bothered by the fact that it doesn't do anything. I just misunderstood the original ticket when the person was referring to the call not "running anything".

jeremystretch (approved):

PR #3863, opened 2020-01-08
Fixes #2113: NAPALM driver settings

### LINKED ISSUE #2113: Dynamically pass credentials to the Napalm module
### Issue type
[x] Feature request 
[ ] Bug report      
[ ] Documentation   

### Environment
* Python version: 3.5.4 
* NetBox version:  2.1.3 

### Description
Please implement a way to pass SSH credentials to the Napalm module via the web interface or login creds.

I require a dynamic way to pass the --user and --password Naplam values to the module. My goal is to hookup NetBox to LDAP therefore allowing users to auth with NetBox and query devices using Napalm with the same creds.

Thanks,

hSaria:
### Fixes: #2113

* Allow the username and password for NAPALM to be overridden for an API call.
* Expose NAPALM's `optional_args` using headers prefixes with `NAPALM-`. Because they can contain sensitive information like `secret`, I've opted to use headers.
* Documented how NetBox works with NAPALM and other odd and ends.

### DIFF
--- a/docs/additional-features/napalm.md
+++ b/docs/additional-features/napalm.md
@@ -17,15 +17,15 @@ GET /api/dcim/devices/1/napalm/?method=get_environment
 
 ## Authentication
 
-By default, the [`NAPALM_USERNAME`](../../configuration/optional-settings/#napalm_username) and [`NAPALM_PASSWORD`](../../configuration/optional-settings/#napalm_password) are used for NAPALM authentication. They can be overridden for an individual API call through the `NAPALM-Username` and `NAPALM-Password` headers.
+By default, the [`NAPALM_USERNAME`](../../configuration/optional-settings/#napalm_username) and [`NAPALM_PASSWORD`](../../configuration/optional-settings/#napalm_password) are used for NAPALM authentication. They can be overridden for an individual API call through the `X-NAPALM-Username` and `X-NAPALM-Password` headers.
 
 ```
 $ curl "http://localhost/api/dcim/devices/1/napalm/?method=get_environment" \
 -H "Authorization: Token f4b378553dacfcfd44c5a0b9ae49b57e29c552b5" \
 -H "Content-Type: application/json" \
 -H "Accept: application/json; indent=4" \
--H "NAPALM-Username: foo" \
--H "NAPALM-Password: bar"
+-H "X-NAPALM-Username: foo" \
+-H "X-NAPALM-Password: bar"
 ```
 
 ## Method Support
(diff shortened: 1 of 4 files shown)


jeremystretch:
I'm wondering if we should prefix the header names with `X-`, e.g. `X-NAPALM-USERNAME`.

hellerve:
> I'm wondering if we should prefix the header names with `X-`, e.g. `X-NAPALM-USERNAME`. Here's [some discussion](https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions) on the topic.

I’m in favor of prefixing with `X-`. I’m usually not a fan of in-band signalling like that, but it’s become enough of a standard that I think respecting it makes sense. I understood IETF’s reasoning to deprecate it, but I do not think that those headers _should_ ever be part of the standard, since they’re extremely specific.

docs/additional-features/napalm.md
```diff
@@ -0,0 +1,65 @@
+# NAPALM
```
jeremystretch: Thank you for adding this documentation! It's been missing for a long time, yet somehow was never directly addressed.

netbox/dcim/api/views.py
```diff
@@ -396,13 +396,29 @@ def napalm(self, request, pk):
...
         ip_address = str(device.primary_ip.address.ip)
+        username = settings.NAPALM_USERNAME
+        password = settings.NAPALM_PASSWORD
         optional_args = settings.NAPALM_ARGS.copy()
         if device.platform.napalm_args is not None:
             optional_args.update(device.platform.napalm_args)
+
+        # Update NAPALM parameters according to the request headers
+        for header in request.headers:
+            if header[:7].lower() != 'napalm-':
```
jeremystretch: Waiting on feedback right now to see if we should opt for a `X-` prefix on the header names. (We've used custom headers elsewhere in NetBox without the prefix, but I wanted to give it some more thought.)

docs/additional-features/napalm.md
hSaria: Right there with ya! If there's anything else burning that you'd like written, I'm here to help :). It might not be how you want it from the get go, but I accept constructive criticism with open arms.

netbox/dcim/api/views.py
hSaria: Now that you mention, yea, it should be additionally prefixed with `X-`. I didn't think of that.

hSaria:
I've changed the prefix to `X-NAPALM-`

PR #3865, opened 2020-01-08
Fixes #3623: Word expansion for interfaces

### LINKED ISSUE #3623: Improve interfaces templates: allow word expansion
### Environment
* Python version:  3.7
* NetBox version:  2.6.6

### Proposed Functionality
In the template interface, allow user to expand names. For example `Interface/[FRONT|BACK]/Eth[1-10]` would produce:

```
Interface/FRONT/Eth1
Interface/FRONT/Eth2
Interface/FRONT/Eth3
Interface/FRONT/Eth4
Interface/FRONT/Eth5
Interface/FRONT/Eth6
Interface/FRONT/Eth7
Interface/FRONT/Eth8
Interface/FRONT/Eth9
Interface/FRONT/Eth10
Interface/BACK/Eth1
Interface/BACK/Eth2
Interface/BACK/Eth3
Interface/BACK/Eth4
Interface/BACK/Eth5
Interface/BACK/Eth6
Interface/BACK/Eth7
Interface/BACK/Eth8
Interface/BACK/Eth9
Interface/BACK/Eth10
```

### Use Case

We need such expansion pattern for our device templates. Maybe others do as well!

### Database Changes
None

### External Dependencies
None

hSaria:
### Fixes: #3623 

Word expansion for interfaces. Just needed accept non-ranges as values. Also added graceful handling to ranges with more than a single character (currently throws an `ord` error because it has more than one character in a range's beginning or end).

> I could not, for the life of me, think of a better example of word expansion than `[xe,ge]-0/0/0`. I'm open for suggestions.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -4,6 +4,7 @@
 
 * [#2050](https://github.com/netbox-community/netbox/issues/2050) - Preview image attachments when hovering the link
 * [#3187](https://github.com/netbox-community/netbox/issues/3187) - Add rack selection field to rack elevations
+* [#3623](https://github.com/netbox-community/netbox/issues/3623) - Add word expansion during interface creation
 
 ## Bug Fixes
 
--- a/netbox/utilities/forms.py
+++ b/netbox/utilities/forms.py
@@ -65,6 +65,9 @@ def parse_alphanumeric_range(string):
                 values.append(begin)
             # Range-based
             else:
+                # Not a valid range (more than a single character)
+                if not len(begin) == len(end) == 1:
+                    raise forms.ValidationError('Range "{}" is invalid.'.format(dash_range))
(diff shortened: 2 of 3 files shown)


jeremystretch:
Let's come up with a TestCase for this function (which is already overdue) before we make any further changes to it. `utilities/tests/test_forms.py` is probably where it should go. @hSaria let me know if you're interested in taking this on or if you'd prefer I handle it. I'm good either way.

hSaria:
Sure thing! I'll follow what convention already exists for other test cases.

jeremystretch:
@hSaria Thanks! Let me know if you run into any issues.

hSaria:
@jeremystretch, I've added an exhaustive list of tests which should cover the possible types of input (including invalid) and the excepted output, but let me know if I missed any or if you'd like something changed.

PR #3867, opened 2020-01-09
Fixes #3668: use `q` to search when assigning IP

### LINKED ISSUE #3009: Search also by description field when assigning addresses
### Environment
* Python version:  3.6.5
* NetBox version:  2.5.8

### Proposed Functionality
Hello, I have a small usability proposal I've been thinking about.

Occasionally, if an IP address is allocated before assigning it to a device, it would be slightly easier to locate afterwards if the search form in "Assign an IP address" page also filtered by the address description field.

This is a small change to the Django filter and does not break existing use cases. Ready to submit a PR if deemed worthy.

### Use Case

1. Reserve IP address for future use and document the name/usage/etc. in the description field
2. Some time passes and the previously reserved numerical address is no longer fresh in your memory
3. Create a device which will be using the previously reserved address and start adding addresses using the Assign an IP Address page.
4. Enter part of the description in the search field and the address will show up without having to remember it numerically.

Previously one would have to first search for the address separately, take note of it and then return to the "Assign an IP address" page or switch between browser tabs, disrupting the workflow.

### LINKED ISSUE #3668: Ability to search by DNS when 'Assigning new IP' address to device
### Environment
* Python version: 3.7.3 
* NetBox version: 2.6.6 

### Proposed Functionality
When assigning an IP address to an interface that exists on the device pagevia the green "+" sign you are presented with a page with two options, 'new' and 'assign IP'. 

We have all usable address in IPAM, so would 'assign' one. In the 'Assign an IP Address' page there is only the option to search via IP address and not DNS. Many of the infra would like the ability to search for an applicable IP address via DNS name and not IP address.

### Use Case
Other non network team members should be able to update their IP address details of their infrastructure and wish to use DNS rather than IP address

### Database Changes
No known changes required.

### External Dependencies
No known changes required.

hSaria:
### Fixes: #3668

Search by address (starting with), DNS name (icontains), or description (icontains) when assigning an IP address.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -4,6 +4,7 @@
 
 * [#2050](https://github.com/netbox-community/netbox/issues/2050) - Preview image attachments when hovering the link
 * [#3187](https://github.com/netbox-community/netbox/issues/3187) - Add rack selection field to rack elevations
+* [#3668](https://github.com/netbox-community/netbox/issues/3668) - Search by DNS name when assigning IP address
 
 ## Bug Fixes
 
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -4,6 +4,7 @@
 
 * [#2050](https://github.com/netbox-community/netbox/issues/2050) - Preview image attachments when hovering the link
 * [#2589](https://github.com/netbox-community/netbox/issues/2589) - Toggle for showing available prefixes/ip addresses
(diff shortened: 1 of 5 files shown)


hSaria:
Hold off on this for a bit.

hSaria:
You can take a look now.

I resolved #3009 in here too. Usually I'd create a separate PR, but then I noticed that the `q` filter is perfect for this.
https://github.com/netbox-community/netbox/blob/92ec4bd22febc63642e8d3872fd7efaeb5bf4842/netbox/ipam/filters.py#L353-L359

PR #3869, opened 2020-01-09
Fixes #3864: Limit number of related IPs

### LINKED ISSUE #3864: IPAM allows creation of IPAddresses with '0' mask length
### Environment
* Python version:  3.6.8
* NetBox version:  v2.6.7

### Steps to Reproduce
1. Create an IP address with a `/0` mask

### Expected Behavior
Validation error returned in form when trying to create an IP address with a mask of 0.

### Observed Behavior
Netbox GUI times out and returns HTTP 504 while trying to load IPAddress detail page. You can see the IP is successfully created in the changelog, but any attempt to view the detail page times out with a reasonable nginx/gunicorn worker timeout (<60 seconds).

hSaria:
### Fixes: #3864 

Limited the related IPs table to 50 entries (currently loads/renders all related IPs). No preference on the number; it just seemed that 50 was good enough for functionality without making the page too long.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -10,6 +10,7 @@
 * [#3589](https://github.com/netbox-community/netbox/issues/3589) - Fix validation on tagged VLANs of an interface
 * [#3853](https://github.com/netbox-community/netbox/issues/3853) - Fix device role link on config context view
 * [#3856](https://github.com/netbox-community/netbox/issues/3856) - Allow filtering VM interfaces by multiple MAC addresses
+* [#3864](https://github.com/netbox-community/netbox/issues/3864) - Limit number of related IPs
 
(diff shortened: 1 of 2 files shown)


jeremystretch:
This isn't a resolution for #3864. NetBox should not allow IP addresses to be created with a `/0` mask.

hSaria:
I see. Didn't realise that was the case. Shall I create a separate ticket for the limit bit? It's still possible to create an IP with a subnet mask that makes it related to excessively many IPs.

jeremystretch:
> Shall I create a separate ticket for the limit bit?

Sure, if you want. Just needs to be called out as a separate issue from the validation bug.

hSaria:
Sure. I'll raise a new PR for the /0 mask thing.

Issue #3393, opened 2019-08-05
Add Pagination to Provider Detail View


tyler-8:
### Environment
* Python version:  3.6.8 
* NetBox version:  2.5.13 

### Proposed Functionality

Similar to https://github.com/netbox-community/netbox/issues/3314, the provider detail circuit listing would benefit from pagination when the object count is higher than a few hundred entries.

Edit: Same goes for IP Address detail page (Related and Duplicate IPs) (`/ipam/ip-addresses/<id>/`)

### Use Case

When viewing Providers (e.g., `/circuits/providers/att/`) with a high number of circuits, the page loads very slow (and performs a larger DB query). Adding pagination will reduce the load on the webserver and the DB.

Edit: Same goes for IP Address detail page (Related and Duplicate IPs) (`/ipam/ip-addresses/<id>/`)

### Database Changes
 NA

### External Dependencies
NA

tyler-8:
Title needs to be changed to reflect the IP address detail page to be covered in this too.

tb-killa:
I think in this framework a global pagination should be introduced for all possible tables that exist as APPS in Netbox.

shuichiro-makigaki:
Regarding "Related IP Address" table in the IP Address view, it becomes too heavy to load when the related IP address list is big. It is because the table does not have pagination feature (all items are listed on one page even if the list size is >9999...).

I think it is challenging to enable pagination on the table only because it is implemented as one of the subpanels on this page. Moving to a separate tab is easier in terms of implementation.

PR #3873, opened 2020-01-09
Fixes #3872: Limit related IPs table

### LINKED ISSUE #3872: No limit on related IP addresses
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.11

### Steps to Reproduce
1. Create IP address in `10/8` space, like `10.1.1.1/8`
2. Repeat step 1 numerous times with a different address each time, but the same mask
3. The list of related IPs for any of those addresses grows with every new related address

### Expected Behavior

The table of related IPs is limited (i.e. a maximum to number of related IPs are displayed). 

### Observed Behavior

The related IPs table is has no bound on its size, which slows down page loading further with every new related IP.

hSaria:
### Fixes: #3872

Limited the related IPs table to 50 entries (currently loads/renders all related IPs). No preference on the number; it just seemed that 50 was good enough for functionality without making the page too long.

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -676,8 +676,15 @@ class IPAddressView(PermissionRequiredMixin, View):
             address=str(ipaddress.address)
         ).filter(
             vrf=ipaddress.vrf, address__net_contained_or_equal=str(ipaddress.address)
-        )[:50]
-        related_ips_table = tables.IPAddressTable(list(related_ips), orderable=False)
+        )
+
+        related_ips_table = tables.IPAddressTable(related_ips, orderable=False)
+
+        paginate = {
+            'paginator_class': EnhancedPaginator,
+            'per_page': request.GET.get('per_page', settings.PAGINATE_COUNT)
+        }
+        RequestConfig(request, paginate).configure(related_ips_table)
 
         return render(request, 'ipam/ipaddress.html', {
             'ipaddress': ipaddress,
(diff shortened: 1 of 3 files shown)


netbox/ipam/views.py
```diff
@@ -676,7 +676,7 @@ def get(self, request, pk):
             address=str(ipaddress.address)
         ).filter(
             vrf=ipaddress.vrf, address__net_contained_or_equal=str(ipaddress.address)
-        )
+        )[:50]
```
jeremystretch: Two requests:

1. Maybe use `settings.PAGINATE_COUNT` instead of hard-coding this value (it defaults to 50 anyway). This variable should be directly accessible from within the template.

2. Check whether the total number of records exceeds the displayed length. If so, include a link to the filtered IP addresses list for convenience.

jeremystretch (changes requested):

jeremystretch:
Tangentially related to #3393. (Just want to capture the relation.)

hSaria:
Yeah, I'm in the process in resolving that ticket. However, for that one, I'll do paging because it completeness of the data is important in that context. For this one, I doubt someone actively uses the related IP address functionality for large data sizes (i.e. >50).

hSaria:
On second thought, I'm going with paginating. I didn't get the optimization that I wanted during my first attempted because I didn't notice `list(related_ips)` which was getting all of the data *then* filtering it down. Pagination is pretty much equal in terms of performance.

netbox/ipam/views.py
hSaria: Done and done.

jeremystretch (approved):

PR #3879, opened 2020-01-09
Fixes #3876: Min/max values for ASN field

### LINKED ISSUE #3876: Increment/Decrement Integer on Site AS Number Does not Function (Chrome/Safari)
### Environment
* Python version:  3.6.8
* NetBox version:  v2.7 beta1 & v2.7 beta2

### Steps to Reproduce
A. Open Google Chrome 79.0.3945.88 (MacOS)
1. Go to sites
2. Edit a site
3. Click on the integer increment button

Firefox works just fine. Safari has a whole different issue (report coming soon). Not sure if this is a browser function or a function of a library Netbox uses? 

### Expected Behavior
Expected to have the integers increment and decrement based on number of mouse clicks.

### Observed Behavior
The number stays the same

hSaria:
### Fixes: #3876 

`ASNField` inherits from `models.BigIntegerField` which is defaulting to long numbers in the `min` and `max` attributes of the input field.

### DIFF
--- a/netbox/dcim/fields.py
+++ b/netbox/dcim/fields.py
@@ -3,16 +3,18 @@ from django.core.validators import MinValueValidator, MaxValueValidator
 from django.db import models
 from netaddr import AddrFormatError, EUI, mac_unix_expanded
 
+from .constants import *
+
 
 class ASNField(models.BigIntegerField):
     description = "32-bit ASN field"
     default_validators = [
-        MinValueValidator(1),
-        MaxValueValidator(4294967295),
+        MinValueValidator(BGP_ASN_MIN),
+        MaxValueValidator(BGP_ASN_MAX),
     ]
 
     def formfield(self, **kwargs):
-        defaults = {'min_value': 1, 'max_value': 4294967295}
+        defaults = {'min_value': BGP_ASN_MIN, 'max_value': BGP_ASN_MAX}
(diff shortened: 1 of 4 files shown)


netbox/dcim/fields.py
```diff
@@ -11,6 +11,11 @@ class ASNField(models.BigIntegerField):
         MaxValueValidator(4294967295),
     ]
 
+    def formfield(self, *args, **kwargs):
```
jeremystretch: Minor nit, but `formfield()` doesn't take positional args; only kwargs.

jeremystretch (changes requested):

netbox/dcim/fields.py
hSaria: My bad. I added it for "completeness" but I should've just stuck to what the documentation said.

lampwins:
I would like us to move away from "magic numbers" and use constants.py, especially in this case because they are being defined twice.

hSaria:
@lampwins makes sense. I'll push another commit for that.

hSaria:
I've replaced them with constants. The max is set to `2**32 - 1` instead of that seemingly arbitrary value (well, arbitrary unless you know your powers of 2).

jeremystretch (approved):

PR #2949, opened 2019-03-01
CSV Import & Export Custom Fields

anthonyeden:
### Fixes: #568 

This pull request is my first attempt at resolving ticket #568.

With these changes, you can now import Custom Fields to Devices. The UI has been updated, and a new version of CSVChoiceField has been created to map the select-field labels to the DB values when performing an import.

It also adds custom fields to the CSV Exporter, using the same column headings as the importer.

If these changes are correct, we need to enable Custom Field importing for all other models by replacing 'forms.ModelForm' with 'CustomFieldForm' in the inheritance of the remainder of the CSVForm models.

PR #3885, opened 2020-01-10
Fixes #568: CSV import/export of custom fields

### LINKED ISSUE #568: Bulk import custom fields
I'd like the possibility, one some custom fields are defined, to have them importable in bulk.

hSaria:
### Fixes: #568

Allows the custom fields to be imported and exported via CSV.

> This is a refactor of the code @anthonyeden wrote in his original PR #2949; the credit goes to him.


netbox/circuits/forms.py
```diff
@@ -144,7 +144,7 @@ class Meta:
         ]
 
 
-class CircuitTypeCSVForm(forms.ModelForm):
+class CircuitTypeCSVForm(CustomFieldForm):
```
jeremystretch: Only models which support the assignment of custom fields (those identified in `extras.constants.CUSTOMFIELD_MODELS`) would need to have their import forms changed.

netbox/extras/forms.py
```diff
@@ -71,7 +71,7 @@ def get_custom_fields_for_model(content_type, filterable_only=False, bulk_edit=F
                     default_choice = cf.choices.get(value=initial).pk
                 except ObjectDoesNotExist:
                     pass
-            field = forms.TypedChoiceField(
+            field = CSVCustomFieldChoiceField(
```
jeremystretch: This is a general-purpose function for retrieving custom fields. There should be nothing specifically related to CSV import here.

netbox/utilities/templatetags/helpers.py
```diff
@@ -149,6 +149,9 @@ def example_choices(field, arg=3):
             break
         if not value or not label:
             continue
+        # Handling for custom fields
+        if hasattr(label, 'value'):
```
jeremystretch: This should be handled in the form widget, not the template tag. The widget should be modified to return a set of choices in the standard tuple format.

netbox/utilities/views.py
```diff
@@ -88,15 +88,28 @@ def queryset_to_csv(self):
         Export the queryset of objects as comma-separated value (CSV), using the model's to_csv() method.
         """
         csv_data = []
+        custom_fields = []
 
         # Start with the column headers
         headers = ','.join(self.queryset.model.csv_headers)
+
+        # Add custom field headers
+        content_type = ContentType.objects.get_for_model(self.queryset.model)
```
jeremystretch: This should probably be wrapped by an `if` statement to clarify that the relevant logic applies only to models with custom fields.

You can use call `get_custom_fields()` on the model to return all applicable CustomFields.

netbox/utilities/views.py
```diff
@@ -88,15 +88,28 @@ def queryset_to_csv(self):
...
+
+        for custom_field in CustomField.objects.filter(obj_type=content_type):
+            headers += ',cf_{}'.format(custom_field.name)
```
jeremystretch: It's a bit tidier to keep `headers` as a list and call `.append()` to add the custom field headers. You can move the `join()` logic below this to generate the string.

jeremystretch (changes requested):
In addition to above, we'll need to add tests for custom field import. IMO a single new test class somewhere under `extras` using the Site object as an example would be sufficient; no need to extend the CSV import test for every model.

netbox/utilities/views.py
hSaria: Can't use `get_custom_fields` here as it's specific to an instance of a model. Any alternatives?

netbox/extras/forms.py
hSaria: Okay, I've named it `CustomFieldChoiceField` and did some minor cleanup on that class.

netbox/utilities/views.py
jeremystretch: `self.queryset.model().get_custom_fields()`

We should really turn that into a class method, though.

hSaria: Thanks for that. I'll update the PR now.

Not sure about changing it to class method; the current implementation returns the custom fields' values of the current element, if any. Though to be honest, `get_custom_fields` is only used once (twice with this change), so changing it isn't that taxing.

docs/release-notes/version-2.7.md
```diff
@@ -2,6 +2,7 @@
 
 ## Enhancements
 
+* [#568](https://github.com/netbox-community/netbox/issues/568) - Allow custom fields to be imported and exported using CSV
```
jeremystretch: This is fine as-is, but don't worry about adding changelog notes for future PRs. It's not worth the hasssle trying to avoid merge conflicts given all the recent activity. I'll make sure they get added.

hSaria: Thank you.

hSaria:
Added the test you requested (sorry for the delay; I completely forget about it).

jeremystretch:
I've merged this into `568-customfield-csv-import` for now because I want to refactor some of the custom field bits before we merge it into `develop`.

PR #3893, opened 2020-01-11
Fixes #3310: Preselect cable fields

### LINKED ISSUE #3310: Make connecting cables quicker by pre-selecting B end Site and Rack
### Environment
* Python version:  3.5.2
* NetBox version:  2.6.1

### Proposed Functionality

When creating a new cable, pre-select the Site and Rack of the B-end to be the Site and Rack of the A-end.

Anybody who wants to connect elsewhere would still have to select the correct Rack or Site/Rack anyway, so you might as well pre-select the most likely choice.

### Use Case

In 2.6.1 there has been a regression in the UI, which means that connecting cables has become more tiresome.

After a data center move, I now have hundreds of cables to connect up.  CSV import is a possibility, but due to heterogenous devices and interface names, selecting them from the available interfaces in the UI would make more sense.

However, the interface for adding a cable has changed between 2.5 and 2.6.  Previously there were two tabs for the B end: under the first tab, you could just type the name of any device (or part of it) to select it.  Under the second tab, you could choose to drill down via site and rack.

It appears the first option has been removed.  Now what happens is:

1. You click on the "Connect" button and you get a drop-down.  I know I'm connecting to an Interface, so that's OK.

    image

2. At the B side you must drill down via site, rack and device.

    image

3. If you try to type into the "device" box without selecting Site/Rack, you are only allowed to choose *unracked* devices (across all sites).  So you are forced to select Site and Rack (or just Rack) first, before typing in part of the target device name.

Under 2.5 it was going to be simple - working through the devices in rack X, I would have the name of the switch in rack X in my clipboard and just paste it in to the target device, then just have to select the interface. I can't do this now, without also selecting Site/Rack every time.

There are a bunch of ways the UI for connecting devices could be improved, but pre-selecting the Site and Rack would be minimally invasive and yet speed up the 80/20 case where connections are made within a rack (e.g. device to top-of-rack switch).

Another option would be that if the site and rack are unselecting, then typing into the Device field could search across all devices in the database.  However I understand that is going to become less feasible when naming uniqueness rules are relaxed.

### Database Changes

None

### External Dependencies

None

hSaria:
### Fixes: #3310

Preselect the site and rack fields when connecting a cable. Also noticed that the status, type, and length unit weren't using Select2, so fixed that one up as well.

> Because of how chain filtering cannot span multiple model relationships (AFAIK), I cannot filter the list of interfaces based on the site and rack; this change only assists with the devices field.


netbox/dcim/forms.py
```diff
@@ -2725,6 +2725,23 @@ class Meta:
...
+
+    def __init__(self, *args, **kwargs):
+
+        super().__init__(*args, **kwargs)
+
+        # Initialize site and rack fields to match that of the existing temination
+        if self.instance and self.instance.termination_a:
+            termination_a_parent = self.instance.termination_a.parent
+            if hasattr(termination_a_parent, 'site'):
+                self.initial['termination_b_site'] = termination_a_parent.site
```
jeremystretch: Be careful with initial data. As written, this will override any initial data passed as a URL query parameter (e.g. `?termination_b_site=123`). We only want to set `self.initial['termination_b_site']` if it doesn't already exist.

netbox/dcim/forms.py
```diff
@@ -2725,6 +2725,23 @@ class Meta:
...
+            if hasattr(termination_a_parent, 'rack'):
+                self.initial['termination_b_rack'] = termination_a_parent.rack
```
jeremystretch: The rack isn't getting pre-selected in the form widget, even though the initial data is being passed correctly. I'm sure this has something to do with the APISelect widget's behavior. Not sure it's worth tracking down but that's up to you.

jeremystretch (changes requested):

netbox/dcim/forms.py
hSaria: Thanks for pointing that out. I've fixed this and will keep it in mind in the future; I've never touched Django before in my life, so I appreciate your patience with my rookie mistakes.

netbox/dcim/forms.py
hSaria: This is happening because of this and is affecting any chained field. Currently trying to figure out what's the purpose of that.
https://github.com/netbox-community/netbox/blob/66daeda85f5cf1e5b2b84dbb1027653af5af9425/netbox/utilities/forms.py#L724

hSaria: Correct me if I'm wrong, but I think `ChainedFieldsMixin` is no longer needed as it's been replaced with the `filter_for` functionality of APISelect? If that's the case, should we cleanup the chaining? If so, may I 😄 ?

hSaria: The rack issue will be resolved as part of #3912

jeremystretch:
The root issue here was trying to set initial data in the form's `__init__()` _after_ it had already been modified by the mixin. Ideally, initial data should be applied from within the view rather than from within the form itself. I've created PR #3987 to resolve this, which incorporates your other changes here. Thank you for your work on this!

Issue #3377, opened 2019-07-30
Power utilization not calculated over cascaded devices/PDU


A888R:
# Environment
* Python version:  3.6.8
* NetBox version:  2.6.1

### Steps to Reproduce
1. create power panel
2. create power feed
3. create 3 devices (example: 'pdu_1level', 'pdu_2level', 'device')
4. connect 'pdu_1level' to created feed (over power port)
5. connect 'pdu_2level' to 'pdu_1level' (over power port + power outlet)
6. connect 'device' to 'pdu_2level'
7. set 'Allocated draw' for 'device' (example: 300)

### Expected Behavior
In 'pdu_2level' view 'Power Utilization -> Allocated' - 300;
In 'pdu_1level' view 'Power Utilization -> Allocated' - 300;
In power feed view 'Utilization (Allocated)' - 300.

### Observed Behavior
In 'pdu_2level' view 'Power Utilization -> Allocated' - 300;
In 'pdu_1level' view 'Power Utilization -> Allocated' - **0**;
In power feed view 'Utilization (Allocated)' - **0**.

larsuhartmann:
Same behaviour on our Side - It would be nice to have this Feature - it is not uncommon to add PDUs to USVs and USVs to Power Feeds especially in smaller Edge Cabinets..

roblarose:
Agreed.  In a "server room" (versus a large datacenter?) we often see multiple PDUs on a single UPS (backed by a single circuit/feed), so at least 3 hops for most equipment.

Feed -> UPS -> PDU -> Device

jeremystretch:
Issues like this illustrate why participation in beta releases is so important. Modifying the current approach to support recursive calculations efficiently would require a substantial amount of work. Do we have any volunteers?

larsuhartmann:
We could switch our Netbox instance to a beta branch to test it out - no problem!

jeremystretch:
@larsuhartmann not beta testing, actually committing to doing the work.

andrewcleveland:
@jeremystretch Is a recursive SQL query the right way to go about this?
[nested_power_draw_legs.patch.txt](https://github.com/netbox-community/netbox/files/3884733/nested_power_draw_legs.patch.txt)
Just a proof of concept (individual legs only).

chas0rde:
This definetly is also an issue for us. Sadly we do not have internal coding-capabilities we could supply but we would be happy if anyone could support this think

hSaria:
[...]

PR #3897, opened 2020-01-11
Fixes #3782: Prevent power loops

### LINKED ISSUE #3782: Disallow creating power loops/cycles
### Environment
* Python version:  3.7.5
* NetBox version:  2.6.7

### Proposed Functionality
It is currently possible to connect multiple devices in a way that creates a loop in the power flow. For example:

Device A power outlet <-> Device B power port
Device B power outlet <-> Device C power port
Device C power outlet <-> Device A power port

Disallowing creating a loop like this would accomplish the following:
* Prevent user error- a physical device layout as above with power connected in a loop would be nonfunctional (effectively the same as plugging a PDU into itself), so a user would never intentionally create such a layout in Netbox.
* Ease implementation of #3377 - calculating power utilization accurately with multiple chained downstream devices.

This could be implemented by adding a verification step when creating a cable between a power port and power outlet.
It would start at the power outlet to be connected and follow all upstream paths (power outlet -> device -> power port -> power outlet -> ...), verifying that the power port to be connected is not present in any of them.

### Use Case
A user who inadvertently attempts to connect a power port and power outlet such that the connection would create a power loop would be prevented from doing so.

### Database Changes
None.

### External Dependencies
None.

hSaria:
### Fixes: #3782

Prevent power loops from being created. It ended up being a derivative of the Bellman–Ford algorithm, though it didn't start as one. 

> Please note that this is not necessarily something we'd want to do; I'm not the expert on power so I don't know if loops inherently exist out there. I created this PR as a potential solution to the ticket, but apart from #3377, it makes no difference if a loop is preset (AFAIK), so it might be worth to continue allowing power loops and just figure out a different way to solve #3377. To be honest, I'd like to keep power loops in case someone needs them. 

It runs in a ring-like structure (like the rings on a piece of wood), expanding on every iteration. The ring starts with the source node (i.e. `termination_a.device` of the new connection). On the second iteration, it will include devices on hop away from those processed in the iteration before. It will stop once all vertices (power cables in this case) that connect to the source node (directly or indirectly) have been scanned, or when a loop gets detected (an iteration containing the destination node `termination_b.device`).

```
1st iteration/ring:    A -)- B --- C --- D
2nd iteration/ring:    A --- B -)- C --- D
3rd iteration/ring:    A --- B --- C -)- D
```

> I did my best to in-line document the operation and to keep it as simple as I could, though a cup of coffee when reading it would definitely ease your experience.

Some notes:
* You should probably look at the tests first to understand specifically what this is preventing against.
* _Relating to the number of the database operations_, this implementation has an upper bound of _O(2n)_, where _n_ is the longest number of hops involving the source node. 
* A vertex between the nodes must be of type power outlet; there's an in-line explanation for this (tldr; it is __always__ one end of the cable what would cause a power loop). Because of this and how power is modelled, power feeds improve the search by acting as boundaries.
* Existing power loops will be ignored, but new ones will be prevented.
* Going back to the algorithm, an alternative solution would be to use recursive queries which would have a O(1) time complexity from the perspective of the ORM. However, I didn't go with that implementation because I'm not all too comfortable with it as it'll involve a raw query, but I'm open to refactoring the code to use that instead.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -19,6 +19,7 @@
 ## Bug Fixes
 
 * [#3589](https://github.com/netbox-community/netbox/issues/3589) - Fix validation on tagged VLANs of an interface
+* [#3782](https://github.com/netbox-community/netbox/issues/3782) - Prevent power loops
 * [#3849](https://github.com/netbox-community/netbox/issues/3849) - Fix ordering of models when dumping data to JSON
 * [#3853](https://github.com/netbox-community/netbox/issues/3853) - Fix device role link on config context view
 * [#3856](https://github.com/netbox-community/netbox/issues/3856) - Allow filtering VM interfaces by multiple MAC addresses
--- a/netbox/dcim/models.py
+++ b/netbox/dcim/models.py
@@ -3099,6 +3099,43 @@ class Cable(ChangeLoggedModel):
         if self.termination_a == self.termination_b:
             raise ValidationError("Cannot connect {} to itself".format(self.termination_a_type))
 
+        # A power loop is prohibited. Power feeds cannot create loops as they're downstream-only. Cannot use power
+        # ports as one end may not have a device, like a power feed. This leaves power outlets which are always a part
+        # of a loop since they exist on a device and connect to another.
+        if 'poweroutlet' in [type_a, type_b]:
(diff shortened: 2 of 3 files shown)


hSaria:
I'm gonna close this PR as I'm not opposed to it. I'll work on a solution for #3377 instead and make sure it tolerates loops.

Issue #2984, opened 2019-03-08
Changelog for Cable Creation Shows Object as "#None"


hSaria:
### Environment
* Python version:  3.6.6
* NetBox version:  2.5.7

### Steps to Reproduce
1. Create a Cable
2. Go to /extras/changelog/
3. The changelog entry there shows `#None` in the Object column

### What did you expect to happen?
As with the short changelog on the main page, the extras/changelog/ page should show the correct object ID for the cable.

### What happened instead?
On the extras/changelog/ page, it will show `#None` in the Object column, instead of `#123` (123 being the ID of the cable). (the changelog on the main page shows the correct ID)

DanSheps:
I haven't dove into this yet, but it appears to be related to using a null label on the cable

PR #3899, opened 2020-01-12
Fixes #3898: Call str of cable on delete to save PK in id_string

### LINKED ISSUE #3898: Message when disconnecting a cable not showing ID
### Environment
* Python version: 3.6.8
* NetBox version: 2.6.12-dev

### Steps to Reproduce
1. Go to a device (e.g. `dcim/devices/1/`)
2. Disconnect a cable

### Expected Behavior
The message says `Deleted cable #5` where 5 is the ID of the cable.

### Observed Behavior
The message says `#None`

This is similar to https://github.com/netbox-community/netbox/issues/2984

hSaria:
### Fixes: #3898

Called the `__str__` method of the cable to create the `self.id_string` attribute if not there. An alternative solution would've been to move the `msg` line before `delete()` is called, but I checked all models' `__str__` method and it's only `Cable` that uses the PK as part of the string.
https://github.com/netbox-community/netbox/blob/b7e78028ce6e998dc9172bd41bc602045daeb1ea/netbox/utilities/views.py#L289-L295

### DIFF
--- a/netbox/dcim/models.py
+++ b/netbox/dcim/models.py
@@ -3040,6 +3040,11 @@ class Cable(ChangeLoggedModel):
     def get_absolute_url(self):
         return reverse('dcim:cable', args=[self.pk])
 
+    def delete(self, *args, **kwargs):
+        # Trigger the __str__ method to save the pk into `self.id_string`
+        str(self)
+        super().delete(*args, **kwargs)
+
     def clean(self):
 
         # Validate that termination A exists
--- a/netbox/dcim/models.py
+++ b/netbox/dcim/models.py
@@ -3027,23 +3027,17 @@ class Cable(ChangeLoggedModel):
             ('termination_b_type', 'termination_b_id'),
         )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # A copy of the PK to be used by __str__ in case the object is deleted
+        self._pk = self.pk
+
     def __str__(self):
-        if self.label:
-            return self.label
-
-        # Save a copy of the PK on the instance since it's nullified if .delete() is called
-        if not hasattr(self, 'id_string'):
-            self.id_string = '#{}'.format(self.pk)
-
-        return self.id_string
+        return self.label or '#{}'.format(self._pk)
 
     def get_absolute_url(self):
         return reverse('dcim:cable', args=[self.pk])
-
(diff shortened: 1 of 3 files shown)


jeremystretch:
My fix for this under #2984 was admittedly naive, since it requires calling `str()` on the cable instance first. We can do this more efficiently by setting `self._pk = self.pk` on init, and having `__str__` return `self.label or self._pk`.

hSaria:
Yep, that'd be cleaner. Let me adjust the PR.

netbox/dcim/models.py
```diff
@@ -3027,15 +3027,14 @@ class Meta:
...
 
-        # Save a copy of the PK on the instance since it's nullified if .delete() is called
-        if not hasattr(self, 'id_string'):
-            self.id_string = '#{}'.format(self.pk)
+        # A copy of the PK to be used by __str__ in case the object is deleted
+        self._pk = self.pk
 
-        return self.id_string
+    def __str__(self):
+        return self.label or '#{}'.format(self._pk)
```
jeremystretch:
```suggestion
        return self.label or '#{}'.format(self.pk or self._pk)
```

Sorry, my earlier suggestion ignored the Cable creation workflow. This checks for `self.pk` first and falls back to `self._pk` only if the instance has been deleted.

jeremystretch (changes requested):

netbox/dcim/models.py
hSaria: I could change that if you'd like, but it doesn't help as` __init__` will always be called in order to instantiate `self`.

If you're referring to the test, it will fail with or without the suggested change. It's the fact that the cable object is not fetched after being saved and is only getting a PK afterwards, init is not being called again so `_pk` is not being updated, and `str` is being called after the cable is being deleted, so `pk` is already nulled.

jeremystretch: Consider this:

```python
c = Cable(...)
c.save()
```

`c._pk` would be set to `None` when init is called on line 1, since the object hasn't been created in the database yet. `c.pk` is assigned during `save()`, but `self._pk` won't get updated on the instance.

hSaria: Yeah, so my solution was to add this during `save` (ignore the force-push; I messed up while copying)
```
self._pk = self.pk
```

jeremystretch (approved):

PR #3901, opened 2020-01-13
Fixes #3900: component template related object existance not checked

### LINKED ISSUE #3900: Deleting DeviceType with Component raises DeviceType.DoesNotExist exception
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.11

### Steps to Reproduce
1. Create a Device Type
2. Add any component to the device type template (example two Power ports)
3. Delete the Device Type

### Expected Behavior
The Device Type to be deleted and being redirected to the Device Type list.

### Observed Behavior
The Device Type is deleted, but an exception is raised:
```
<class 'dcim.models.DeviceType.DoesNotExist'>

DeviceType matching query does not exist.
```

The URL with the exception is https://netbox-server/dcim/device-types/3/delete/

It seems like deleting the device type components are trying to redirect to the Device Type after deletion, but the Device Type has already been deleted. So some way of differentiating the `return_url ` when deleting the Device Type (which includes deleting all components) contra when deleting only components from the Device Type is needed.

hSaria:
### Fixes: #3900

Needed to check whether the `related_object` existed for the component template.

I've checked all other instances of `ObjectChange` and this seems to be the only one missing this check.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -28,6 +28,7 @@
 * [#3872](https://github.com/netbox-community/netbox/issues/3872) - Paginate related IPs of an address
 * [#3876](https://github.com/netbox-community/netbox/issues/3876) - Fixed min/max to ASN input field at the site creation page
 * [#3882](https://github.com/netbox-community/netbox/issues/3882) - Fix filtering of devices by rack group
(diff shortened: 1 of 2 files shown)


hSaria:
Closing this PR as there is a discussion to be had regarding the whether to keep the delete record of child when the parent is deleted. Once the maintainers have made the decision, I'm happy to take another look.

PR #3906, opened 2020-01-13
Fixes #3905: divide by zero on power feeds with low values

### LINKED ISSUE #3905: Divide by zero error with very low power feed values
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.11

### Steps to Reproduce
1. Create power feed with voltage, amperage, and max utilization set to 1
2. Create device with a power port with no power attributes (e.g. maximum_draw) that connects to the power feed created in step 1
3. Browse to the power feed details page

### Expected Behavior
Page loads

### Observed Behavior
`ZeroDivisionError` exception because the percentage calculation is using `available_power` which was rounded to 0 with
```python
self.voltage * self.amperage * (self.max_utilization / 100) =
1            * 1             * 0.01                         = 0.01
```

hSaria:
### Fixes: #3905

The `available_power` attribute had the possibility of being set to 0. It'll now be set to 1 if the calculation returns 0.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -28,6 +28,7 @@
 * [#3872](https://github.com/netbox-community/netbox/issues/3872) - Paginate related IPs of an address
 * [#3876](https://github.com/netbox-community/netbox/issues/3876) - Fixed min/max to ASN input field at the site creation page
 * [#3882](https://github.com/netbox-community/netbox/issues/3882) - Fix filtering of devices by rack group
+* [#3905](https://github.com/netbox-community/netbox/issues/3905) - Fix divide-by-zero on power feeds with low power values
 
 ---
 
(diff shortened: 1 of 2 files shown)


netbox/dcim/models.py
```diff
@@ -2923,9 +2923,9 @@ def save(self, *args, **kwargs):
         # Cache the available_power property on the instance
         kva = self.voltage * self.amperage * (self.max_utilization / 100)
         if self.phase == POWERFEED_PHASE_3PHASE:
-            self.available_power = round(kva * 1.732)
+            self.available_power = round(kva * 1.732) or 1
         else:
-            self.available_power = round(kva)
+            self.available_power = round(kva) or 1
```
jeremystretch: While this does prevent the error, we're essentially overriding user-input values with an arbitrary static to avoid returning zero. Although I can't imagine a real-world scenario that would result in `available_power` being set to 0 kVA, IMO a more elegant approach would be to address this in the template (`dcim/powerfeed.html`).

jeremystretch:
Per my comment above, we can wrap this line in `dcim/powerfeed.html` with an `if` statement that ensures `powerfeed.available_power` is greater than zero:

```{% utilization_graph utilization.allocated|percentage:powerfeed.available_power %}```

hSaria:
Thank you! This does indeed make it better. Let me update the PR.

jeremystretch (approved):

PR #3910, opened 2020-01-13
Fixes #3909: Regression from #3898

### LINKED ISSUE #3909: Regression from #3898 on device detail
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.12

### Steps to Reproduce
1. Create two devices with an interface
2. Connect the two interfaces
3. Go to device detail page

### Expected Behavior
Page loads

### Observed Behavior
Exception `'Cable' object has no attribute '_pk'`

hSaria:
### Fixes: #3909 

Cable `str` method is being called during the `super()` in `__init__` which is causing an exception when accessing `_pk`. It will now fallback to `self.pk` even if it's None (I didn't go with `self.pk or self._pk` as it may trigger the same issue on a new cable).

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -1,3 +1,11 @@
+# v2.6.13 (FUTURE)
+
+## Bug Fixes
+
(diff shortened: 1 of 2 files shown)


hSaria:
No longer required. See #3909

PR #3913, opened 2020-01-13
Fixes #3912: Remove chaining (now handled via APISelect)

### LINKED ISSUE #3912: Remove ChainedFieldsMixin
### Proposed Changes
Remove the `ChainedFieldsMixin`, `ChainedModelChoiceField`, and `ChainedModelMultipleChoiceField`.

### Justification
Chaining has been replaced with APISelect's `filter_for` and is no longer required. It updates the queryset, but it is entirely ignored by the API call. And ever since #3812, only the selected option (if any) in the queryset is rendered, so limiting makes very little difference.

hSaria:
### Fixes: #3912

Chaining has been superseded by APISelect. Apart from the initial loading, the queryset has no barring on the operation of APISelect – those widgets use the `api_url`, `filter_for`, and `additional_query_params`.

As you scroll down the diff list, you'll notice that every field and its chain that was changed has an APISelect widget.

> I don't know whether or not to create a changelog entry for a housekeeping change.


jeremystretch:
Please see my comments on #3912. Thanks for your work toward this end, but in the future it's best to wait on submitting any PRs until the related FR has been accepted.

PR #3915, opened 2020-01-14
Fixes #3914: Interface filter field when unauthenticated

### LINKED ISSUE #3914: Interface filter in device view doesn't work when logged out
### Environment
* Python version: 3.6
* NetBox version: (v2.6.12)

While logged out, attempting to filter for a device doesn't trigger the dynamic filtering.

### Steps to Reproduce
1. Ensure you're logged out.
2. Browse to a device with at least some interfaces defined.
3. Click into the filter box and search for an interface

### Expected Behavior
The list should dynamically filter based on the entered search params.

### Observed Behavior
No change to the list.

hSaria:
### Fixes: #3914 

The selector relied on the `form` element is isn't present when unauthenticated. Fixed this by using the `div.panel` instead.

### DIFF
--- a/docs/release-notes/version-2.6.md
+++ b/docs/release-notes/version-2.6.md
@@ -1,3 +1,11 @@
+# v2.6.13 (FUTURE)
+
+## Bug Fixes
+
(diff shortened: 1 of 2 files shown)


jeremystretch:
Ohhhh. That makes sense. This was a weird bug.

PR #3916, opened 2020-01-14
Fixes #3377: Recursive power calculation

### LINKED ISSUE #3377: Power utilization not calculated over cascaded devices/PDU

hSaria:
### Fixes: #3377

Calculate the power utilization with the use of a recursive query. It's a single query from the perspective of the ORM (recursion happens in Postgres which is far faster than doing it in the app).

A note regarding the tests: they'll detect a regression in the calculation of a power or if loops are occurring, but if someone makes a change and the loop test fails... well... let's just say it won't "technically" fail as the DB would loop forever. Just wanted to make you aware of that.

I left the outlet count to only local ones (current behaviour). I can change that to include all children if need be.

> Thanks for @andrewcleveland for writing the query. Apart from the loop-prevention mechanism that I added from the [Postgres docs](https://www.postgresql.org/docs/9.6/queries-with.html), it was pretty much ready for use.


hSaria:
Hold off on this for a bit. I thought of a way to reduce 4 queries to 1.

hSaria:
Much better now. I think that's as efficient as it'll get.

netbox/dcim/models/device_components.py
```diff
@@ -30,6 +30,35 @@
     'RearPort',
 )
 
+QUERY_POWER_DRAW = """
```
jeremystretch: We need to avoid raw SQL wherever possible as it is very difficult to maintain long term. (It is completely opaque to schema migrations and exempt from IDE introspection.) If we can't find a way to do it efficiently using Django's ORM, we have to ask whether the effort is worth it. I'll defer to the other maintainers, but personally I'm opposed to the introduction of any new raw SQL, anywhere in the project. (We have some bits we're still trying to get rid of.)

hSaria: I agree, though I simply couldn't find a native way of performing recursion through the ORM. You'll notice the affects of this when you're on the racks list and have lots of power feeds assigned to them as each of those power feeds' attached power port is called with `get_power_draw`.

We can do this through the ORM, but the best that I got is _O(n)_ with _n_ being the depth (i.e. hops from the origin).

jeremystretch: After some discussion with the other maintainers, we've agreed to reject the introduction of any new raw SQL queries if they can be avoided, for the reasons I mentioned above. It's simply not an acceptable development burden given the current resources afforded to the project. If this feature is to be implemented, it will need to be accomplished using the Django ORM.

jeremystretch:
@hSaria per my comment above, do you want to try solving this using the ORM, or should we close this out?

hSaria:
I'll solve it in the ORM; it's pretty easy.

netbox/dcim/models/device_components.py
```diff
@@ -373,37 +373,67 @@ def connected_endpoint(self, value):
...
+            visited_power_ports = PowerPort.objects.none()
+
+            # Power outlets assigned to the current power port
+            power_outlets = PowerOutlet.objects.filter(power_port=self)
+            if leg is not None:
+                power_outlets = power_outlets.filter(feed_leg=leg)
+
+            # Cannot be cached as it will otherwise not update the per-leg stats when an outlet's leg changes.
+            connected_power_ports = PowerPort.objects.exclude(pk__in=visited_power_ports).filter(
+                _connected_poweroutlet__in=power_outlets).nocache()
```
hSaria: I'd preferably like to avoid `nocache` and simply invalidate this query when the leg is changed on one of the `power_outlets`. Not sure if this is possible in a clean way, though I'm doing a bit of research on this.

jeremystretch (commented):
Seems to work okay, but I'm not satisfied with the performance hit when viewing the racks list. When viewing 50 racks with two PDUs each (and not even that many connections in most racks), it takes 494 SQL queries and 11 seconds to load the page. I realize your initial approach employing custom SQL queries is likely more efficient, but as discussed that's not an approach we can maintain.

I think if we're going to implement this, we'll need to either come up with a way to improve query efficiency (maybe caching locally on each power port?), or ditch the power utilization stats on the rack list. I'd be okay with implementing this as-is on the individual rack views only.

hSaria:
I agree. I think it would make sense to cache on each power port as that would also give us the ability to expose it via the API as a read-only attribute.

As you mentioned, we can also remove the power utilization stats from the list. This is clearly the simpler solution.

 Let me know what you and the other maintainers prefer and I'm happy to do the work.

kobayashi:
This is great solution to get power utilization for rack view! I agree with the api way @hSaria mentioned. I am thinking which timing is the best to update the cached data if implemented. :thinking:

For me, I don't have any objection to remove power utilization from rack list view.

hSaria:
I'm working on solution for that. So far, it'll only take two DB calls for each feed leg (one for `.count` and another for `.aggregate(Sum('allocated_draw'), Sum('maximum_draw'))`, so both are very quick).

For each power port, a `ManyToManyField` is maintained which keeps track of all power ports that are indirectly pulling power (downstream) from that port. The field is updated whenever the _topology is changed_. (Working on this last bit)

I've had different ideas regarding the cached value:

* Storing the sums directly: Highly expensive whenever performing updates because the entire topology needs to be calculated. Has the same ORM performance when getting the statistics as the solution above.
* Storing the downstream power ports in the `PowerOutlet`: While this may make more sense at first thought and works just about the same as what I'm working on, it will span the code for handling topology changes over two models which would get quite confusing as you go back and forth while reading
* Using Django MPTT's `TreeForeignKey` to handle the hierarchy: Doesn't work as a power loop is a possibility; if a database already contains one, the migration would fail.

hSaria:
I've reworked how this calculation is done; the downstream power ports (supplied power) are now cached in a field in the outlet, called `downstream_powerports`. The field is updated when a **topology change** occurs, like so:
1. Before the topology change takes effect on power outlet `x`, the current (old) downstream power ports are removed from all of the power outlets above power outlet `x`. This ensures that they don't have any stale information in the case the topology change causes them to no longer be upstream power outlets.
2. The topology change is applied.
3. The new downstream power ports are calculated and cached in `x.downstream_powerports`. In addition, these power ports are added to the cached field of any new (post topology-change) upstream power outlets.

A topology change is always captured by `PowerOutlet.save` as creating or deleting a cable will trigger an update on the PowerOutlet object.

I've improved the tests to accommodate for these changes. There's a simple topology graph in the test class doc string as well as in-line comments to help you understand what they're testing.

There is one little thing that needs sorting: when a loop is a broken (loop-causing cable is removed), the old `downstream_powerports` in the loop are erroneously cleared from the outlets in that loop as all of those outlets are considered old parents (step 1). You can checkout `PowerCalculationTestCase.test_power_loop` for an example of this.

With this implementation `PowerPort.get_power_draw` has the same performance as before – only 2 queries (count and sum) for each leg.

While I work on this last thing, I'd welcome feedback. Also, if @jeremystretch would kindly confirm the performance, namely in the rack list view, I'd greatly appreciate it.

hSaria:
This PR has gotten too dirty with the all the changes. I'm gonna close it and open up a new one later which will have all relevant details in a concise form. Now that I have gotten a better idea of how we'd like this achieved, the new one shouldn't get too many revisions.

PR #3924, opened 2020-01-15
Fixes #3509: IPAddressVar for custom scripts

### LINKED ISSUE #3509: Add a custom script variable for IP addresses
### Environment
* Python version: 3.6.8
* NetBox version: 2.6.3

### Proposed Functionality
Add a variable type for representing IP addresses in custom scripts. This would be similar to the existing `IPNetworkVar`, but for individual IPs rather than prefixes.

### Use Case
Adding this variable type will allow for validation of IP addresses.

### Database Changes
None

### External Dependencies
None

hSaria:
### Fixes: #3509

* Added `IPAddressFormField` which `IPAddressVar` uses. Updated documentation and added test.
* Renamed `IPFormField` to `IPNetworkFormField` (it'd otherwise be confusing with the one above).
* Updated the documentation with the `min_prefix_length` and `max_prefix_length` arguments for `IPNetworkVar`.


jeremystretch:
After some time debating this further, I ended up introducing *two* new variables in #3988: `IPAddressVar` and `IPAddressWithMaskVar`. I left `IPNetworkVar` alone. I figure that's the most obvious, least disruptive way to handle the three use cases I mentioned in #3509. Again, thank you for your work on this!

PR #3925, opened 2020-01-15
Fixes #3840: Only show valid interface VLAN choices

### LINKED ISSUE #3840: Expand Edit Interface 802.1Q VLAN form search query
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.7

### Proposed Functionality
When using the GUI Select2 fields for adding untagged/tagged VLANs to a device/VM interface, expand the search parameters to also search by site and VLAN group. Currently it appears that only the VLAN Name and VID are used. The dropdown currently displays the sites in the dropdown list, but they're not part of the search parameters.

### Use Case
When you have a cookiecutter design for a particular type of site, the VIDs and VLAN names are repeated identically, and added to their respective sites and VLAN groups, if applicable.

When you go to add those VLANs to an interface and search for vlan `5` for example, you may get dozens or hundreds of results because that VID and name are repeated lots of times. This makes it nearly impossible to add the VLAN through the GUI. 

A potential workaround with the current state is to make all the VLAN names unique by adding the site name to the VLAN name, but this seems repetitive and less flexible, should the site name ever change for example.

A cursory glance makes this appear to be an issue with the REST API (and the `q` param) but there is also a `site` filter param that could be used in this case.

### Database Changes
N/A

### External Dependencies
N/A

hSaria:
### Fixes: #3840

The list of VLANs when editing an interface will now only show valid choices (i.e. now matches the form validation).

In order to achieve this, I made it so the `data-additional-query-param-*` is now a JSON-encoded list:
* Allow multiple values by calling `add_addtional_query_param` multiple times over the same key (before, it would overwrite whatever was there).
* It works well with jQuery [`.data`](https://api.jquery.com/data/#data-html5) as it attempts to cast the value to JSON by default

When testing this, do a reload-from-source, else you may get a cached version of `forms.js`.

> This is a follow up to #3589


netbox/dcim/forms.py
```diff
@@ -2826,6 +2852,10 @@ def __init__(self, *args, **kwargs):
                 device__in=[device, device.get_vc_master()],
                 type=InterfaceTypeChoices.TYPE_LAG
             )
+
+            # Add current site to VLANs query params
+            self.fields['untagged_vlan'].widget.add_additional_query_param('site_id', device.site.pk)
```
jeremystretch: This approach doesn't seem very DRY: We're setting additional query parameters in two places, per field, per form. It's inevitable that we'll missing some of these references should the parameters need to be changed in the future.

Additionally, these fields need to be updated for the interface forms in `virtualization.forms` as well.

netbox/utilities/forms.py
```diff
@@ -311,12 +311,17 @@ def add_filter_for(self, name, value):
 
     def add_additional_query_param(self, name, value):
         """
-        Add details for an additional query param in the form of a data-* attribute.
+        Add details for an additional query param in the form of a data-* JSON-encoded list attribute.
```
jeremystretch: Wouldn't it be simpler to use whitespace separation for multiple values when assigning the attribute on the form element? With the JSON approach, you end up with stuff like this:

```
data-additional-query-param-site_id="[&quot;null&quot;, 6]"
```

jeremystretch (changes requested):
We probably want to look at consolidating the interface forms somewhat before we extend them any further.

netbox/utilities/forms.py
hSaria: Primarily to work nicely with jQuery's [`.data`](https://api.jquery.com/data/#data-html5) which attempts to deserialize the value as JSON if possible. Apart from that, I have no preference on which separator to use.

netbox/dcim/forms.py
hSaria: That's a valid concern, and – as you pointed out – virtualization is clear example of this inconsistency. #4057 should improve the implementation of this; I'm happy to close this PR until that issue is closed; let me know what's your preference on this.

jeremystretch: Let's leave this PR open for now. I don't have a read yet on just how involved merging those forms will be: If it's too much effort to take on at the moment, we can push this through.

hSaria: Fixed `virtualization.forms`.

netbox/virtualization/forms.py
```diff
@@ -691,51 +697,12 @@ class Meta:
...
-                site_group_vlans = VLAN.objects.filter(group=group)
-                vlan_choices.append((
-                    '{} / {}'.format(group.site.name, group.name),
-                    [(vlan.pk, vlan) for vlan in site_group_vlans]
-                ))
-
-        self.fields['untagged_vlan'].choices = [(None, '---------')] + vlan_choices
-        self.fields['tagged_vlans'].choices = vlan_choices
-
-    def clean(self):
```
jeremystretch: I believe we still need `clean()` on this form. It's missing from `dcm.forms.InterfaceForm` because it's provided by `InterfaceCommonForm` (which I'd like to work on removing).

jeremystretch (changes requested):
Consolidating the various interface forms is going to take some time, so I'm okay with merging this. Just need to restore `clean()` on the virtualization InterfaceForm and I think it will be good to go!

netbox/virtualization/forms.py
hSaria: Thanks for catching it.

PR #3969, opened 2020-01-21
Fixes: #3960 legacy device status

### LINKED ISSUE #3960: Status tab claims device is inactive
### Environment
* Python version: 3.7.6
* NetBox version: 2.7.1

### Steps to Reproduce
1. Have DCIM devices which are `active` (in the sense of the `dcim_device.status` column in the database) and have a valid database navigation
2. Navigate to the device overview

### Expected Behavior
The `status` tab should be enabled and clickable.

### Observed Behavior
The `status` tab is not clickable, instead it reports `device must be in active status`.
However, manually navigating to `status/` works perfectly fine and displays all desired data.

### Probable reason
The [template](https://github.com/netbox-community/netbox/blob/v2.7.1/netbox/templates/dcim/device.html#L87) still checks `device.status == 1` using the legacy numerical value (although this might be plain wrong because I don't know the codebase).

kobayashi:
### Fixes: #3960 
device status still refers legacy choice (integer).

https://github.com/netbox-community/netbox/blob/295d4f0394b431351c0cb2c3ecc791df68c6c2fb/netbox/templates/dcim/device.html#L86-L88

### DIFF
--- a/docs/release-notes/version-2.7.md
+++ b/docs/release-notes/version-2.7.md
@@ -10,6 +10,7 @@
 * [#3721](https://github.com/netbox-community/netbox/issues/3721) - Allow Unicode characters in tag slugs
 * [#3951](https://github.com/netbox-community/netbox/issues/3951) - Fix exception in webhook worker due to missing constant
 * [#3953](https://github.com/netbox-community/netbox/issues/3953) - Fix validation error when creating child devices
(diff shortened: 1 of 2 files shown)


wk:
Successfully tested against v2.7.1. Resolves issue described as well as availability of other conditional tabs (such as `LLDP Neighbors`).

qaxi:
Successfully tested against v 2.7.1 .

PR #3971, opened 2020-01-21
Display occupied rack units correctly

### LINKED ISSUE #3965: SVG rack elevations: racks seem unoccupied when other side is occupied
### Environment
* Python version:  Any
* NetBox version:  >=2.7.0

Quoting from #3958:
> When you view the front face of a rack, but there is a device in the corresponding rear position, you used to get some shading to indicate this. This no longer appears - it looks like the rack U is completely unoccupied.

I remember that this used to work, but it seems like it doesn’t anymore (haven’t been able to verify). We’ll have to investigate what happened there.

Cheers

hellerve:
### Fixes: #3965

This PR reintroduces code to mark a slot in the rack elevation as occupied if the other side is occupied. This problem was introduced when @lampwins took over the initial draft that I made, because I didn’t correctly explain the differences of my version of `get_rack_units` to the “vanilla” one. `get_rack_units` now has a flag named `always_show_device` that will even put the device in the elevation even if the face is different.

Cheers

### DIFF
--- a/netbox/dcim/models/__init__.py
+++ b/netbox/dcim/models/__init__.py
@@ -468,7 +468,7 @@ class RackElevationHelperMixin:
         :param unit_height: Height of each rack unit for the rendered drawing. Note this is not the total
             height of the elevation
         """
-        elevation = self.get_rack_units(face=face, expand_devices=False)
+        elevation = self.get_rack_units(face=face, expand_devices=False, always_show_device=True)
         reserved_units = self.get_reserved_units().keys()
 
         return self._draw_elevations(elevation, reserved_units, face, unit_width, unit_height)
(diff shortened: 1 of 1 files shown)


jeremystretch:
This doesn't seem to handle a scenario where a rack has two devices installed back-to-back in the same position, one on either face of the rack. In testing, when I add two devices back-to-back, only one or the other displays (presumably determined by the order in which they are returned in the Queryset); the opposing rack unit appears to be empty.

I recall that the old approach was to draw each rack in two "layers:" first the opposing face was rendered, followed by the near (selected) face, ensuring that the near device always obscured the device behind it.

hellerve:
Hm, that makes sense, good catch! I wanted to avoid drawing twice, but we can do that. I’d rather solve this in the query if possible. Does that make sense?

I’ll ponder it for a moment and get back to you.

jeremystretch:
It's probably sufficient to order the list devices by face such that the _oppposite_ face comes first, and then "bump" that device from the drawing if the next device is in the same unit. I haven't dug into the logic much, just pondering as well.

hellerve:
It would be fairly straightforward if all devices had a height of 1, but because you could potentially have a 3U device that’s not full depth and 2 1U devices that aren’t full depth on the other side it’s getting a little wonky (since we also “prune” the list if a device has more than 1U).

hellerve:
I’m just about to push a solution to this. I’m not quite sure whether it works, but it should; I tested it with a bunch of setups like the one below. Would you be able to check?

Bildschirmfoto 2020-01-21 um 16 54 35

hellerve:
The build broke because sometimes I forget whitespaces around operators. Bear with me.

jeremystretch:
LGTM

PR #3973, opened 2020-01-21
dcim: fix tooltips in svg rack display

### LINKED ISSUE #3963: SVG rack elevations don’t render tooltips anymore
### Environment
* Python version:  Any
* NetBox version:  >=2.7.0

As discussed in #3958, the tooltip isn’t rendered anymore inside rack elevations SVG models on hover. @candlerb pointed out that this could potentially be fixed with titles.

This needs some investigation as to how to fix this. I actually pointed this out in #3754, but we must’ve forgotten it somewhere along the process.

Cheers

hellerve:
### Fixes: #3963

This PR fixes #3963 by adding a `title` to the individual slots of the element. Those are rendered as follows:

Bildschirmfoto 2020-01-21 um 16 19 01

So they aren’t quite as pretty and feature-rich as the old ones, but they show the same information. To emulate the old ones we have to tinker a little with JavaScript I fear, and I don’t feel equipped to do that in the SVG context.

@candlerb can you confirm that this is what you expected?

Cheers

### DIFF
--- a/netbox/dcim/models/__init__.py
+++ b/netbox/dcim/models/__init__.py
@@ -395,13 +395,22 @@ class RackElevationHelperMixin:
                 fill='black'
             )
         )
+        link.set_desc('{} — {} ({}U) {} {}'.format(
+            device.device_role, device.device_type.display_name,
+            device.device_type.u_height, device.asset_tag or '', device.serial or ''
+        ))
         link.add(drawing.rect(start, end, fill='#{}'.format(color)))
         hex_color = '#{}'.format(foreground_color(color))
         link.add(drawing.text(str(device), insert=text, fill=hex_color))
 
     @staticmethod
     def _draw_device_rear(drawing, device, start, end, text):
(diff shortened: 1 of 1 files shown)


candlerb:
I don't have an old instance left to compare with :-)  However role and manuf/model seems reasonable to me.

hellerve:
I took the template from pre-2.6 and recreated the text. The style of course is different, but the information on it should be the same!

jeremystretch:
I don't think this _fully_ resolves the issue, since I'd really love to support popovers with formatted text as we had in prior releases, however this works well as a temporary solution. I think it's best to close #3963 with this; I'll open a new FR for the improved popovers to be undertaken in the future. Thanks for the quick turnaround on these @hellerve!

hellerve:
Of course. I broke it, I fix it.

I also agree that the fidelity is a little low, but I think that working on improving the popovers for too long might not have the right ROI at the moment.

PR #3990, opened 2020-01-23
Fix incorrect file references in documentation

### LINKED ISSUE #3961: Reference to incorrect files in migrating-to-systemd.md
### Change Type
[ ] Addition
[X] Correction
[ ] Deprecation

### Proposed Changes
In `migrating-to-systemd.md`:
- there is a section referring to non existent `contrib/netbox.env`
- references to `gunicorn.conf` should be to `gunicorn.py` 

`3-http-daemon.md` has the correct instructions

mikewright:
### Fixes: #3961 

updated documentation to reference correct files
copied largely from 3-http-daemon.md

### DIFF
--- a/docs/installation/migrating-to-systemd.md
+++ b/docs/installation/migrating-to-systemd.md
@@ -12,82 +12,22 @@ Migration is not required, as supervisord will still continue to function.
 
 ### systemd configuration:
 
-Copy or link contrib/netbox.service and contrib/netbox-rq.service to /etc/systemd/system/netbox.service and /etc/systemd/system/netbox-rq.service
+Copy or link `contrib/netbox.service` and `contrib/netbox-rq.service` to the `/etc/systemd/system/` directory:
 
 ```no-highlight
-# cp contrib/netbox.service /etc/systemd/system/netbox.service
-# cp contrib/netbox-rq.service /etc/systemd/system/netbox-rq.service
+# cp contrib/*.service /etc/systemd/system/
 ```
 
-Edit /etc/systemd/system/netbox.service and /etc/systemd/system/netbox-rq.service. Be sure to verify the location of the gunicorn executable on your server (e.g. `which gunicorn`).  If using CentOS/RHEL.  Change the username from `www-data` to `nginx` or `apache`:
+!!! note
+    These service files assume that gunicorn is installed at `/usr/local/bin/gunicorn`. If the output of `which gunicorn` indicates a different path, you'll need to correct the `ExecStart` path in both files.
+
+Copy `contrib/gunicorn.py` to `/opt/netbox/gunicorn.py`. We make a copy of this file to ensure that any changes to it do not get overwritten by a future upgrade.
 
 ```no-highlight
-/usr/local/bin/gunicorn --pid ${PidPath} --pythonpath ${WorkingDirectory}/netbox --config ${ConfigPath} netbox.wsgi
+# cp contrib/gunicorn.py /opt/netbox/gunicorn.py
 ```
 
-```no-highlight
-User=www-data
-Group=www-data
-```
-
-Copy contrib/netbox.env to /etc/sysconfig/netbox.env
-
-```no-highlight
-# cp contrib/netbox.env /etc/sysconfig/netbox.env
-```
-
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
I'm fine with fixing the mistakes but let's not gut the examples without further discussion. #3961 pertains only to the correction of errors.

docs/installation/migrating-to-systemd.md
```diff
@@ -12,82 +12,22 @@ Migration is not required, as supervisord will still continue to function.
...
 
 ```no-highlight
-# cp contrib/netbox.service /etc/systemd/system/netbox.service
-# cp contrib/netbox-rq.service /etc/systemd/system/netbox-rq.service
+# cp contrib/*.service /etc/systemd/system/
 ```
 
-Edit /etc/systemd/system/netbox.service and /etc/systemd/system/netbox-rq.service. Be sure to verify the location of the gunicorn executable on your server (e.g. `which gunicorn`).  If using CentOS/RHEL.  Change the username from `www-data` to `nginx` or `apache`:
+!!! note
+    These service files assume that gunicorn is installed at `/usr/local/bin/gunicorn`. If the output of `which gunicorn` indicates a different path, you'll need to correct the `ExecStart` path in both files.
```
kobayashi: Let us keep the deleted line and add your note lines as supplement.

kobayashi (changes requested):

DanSheps:
Looking at the PR, it looks like @kobayashi suggestion is all that is needed.  The examples that are in there are from the old systemd configuration that we never used so I think we are fine to remove those as part of this.

docs/installation/migrating-to-systemd.md
```diff
@@ -98,3 +38,20 @@ Finally, start the `netbox` and `netbox-rq` services and enable them to initiate
...
+   Active: active (running) since Thu 2019-12-12 19:23:40 UTC; 25s ago
+     Docs: https://netbox.readthedocs.io/en/stable/
+ Main PID: 11993 (gunicorn)
+    Tasks: 6 (limit: 2362)
+   CGroup: /system.slice/netbox.service
+           ├─11993 /usr/bin/python3 /usr/local/bin/gunicorn --pid /var/tmp/netbox.pid --pythonpath /opt/netbox/...
+           ├─12015 /usr/bin/python3 /usr/local/bin/gunicorn --pid /var/tmp/netbox.pid --pythonpath /opt/netbox/...
+           ├─12016 /usr/bin/python3 /usr/local/bin/gunicorn --pid /var/tmp/netbox.pid --pythonpath /opt/netbox/...
+...
+```
```
DanSheps: I don't think we need to include this.  Someone who is relatively proficient with systemd will know how to do this and someone who is not won't really know what is going on (beyond the running) and there are other ways to verify that.

docs/installation/migrating-to-systemd.md
```diff
@@ -12,82 +12,22 @@ Migration is not required, as supervisord will still continue to function.
...
-Copy or link contrib/netbox.service and contrib/netbox-rq.service to /etc/systemd/system/netbox.service and /etc/systemd/system/netbox-rq.service
+Copy or link `contrib/netbox.service` and `contrib/netbox-rq.service` to the `/etc/systemd/system/` directory:
 
 ```no-highlight
-# cp contrib/netbox.service /etc/systemd/system/netbox.service
-# cp contrib/netbox-rq.service /etc/systemd/system/netbox-rq.service
+# cp contrib/*.service /etc/systemd/system/
 ```
 
-Edit /etc/systemd/system/netbox.service and /etc/systemd/system/netbox-rq.service. Be sure to verify the location of the gunicorn executable on your server (e.g. `which gunicorn`).  If using CentOS/RHEL.  Change the username from `www-data` to `nginx` or `apache`:
```
DanSheps: We may want to keep the username portion, as it is not always the case that www-data is used or available.  However I would just tack it onto the end.

DanSheps (changes requested):

DanSheps:
There has been no activity on this for a few days, I am going to close this out and resolve myself as we have had a few issues opened up in the last few days.

PR #3994, opened 2020-01-23
3193 front rear mapping

### LINKED ISSUE #3193: Front/Rear port mapping breaks when multiple cables are between devices
### Environment
* Python version:  
* NetBox version:  2.5.10

**<class 'django.db.utils.IntegrityError'>

duplicate key value violates unique constraint "dcim_interface__connected_circuittermination_id_key"
DETAIL:  Key (_connected_circuittermination_id)=(22) already exists.**

When I have two devices that I am mapping multiple front ports to a rear port on each and I join the rear ports of the two devices together, I can put a device on front port 1 on device A and it will connect all the way through to a device placed on front port 1 on device B.  Same for any device I connect to a front port on Device A, it will show as connected to the corresponding port on device B.
```
F 1 ---\                        /---F 1
 .  ----\                      /---- .
 .  -----\____________________/----- .
 .  -----/rear 1:1    rear 1:1\----- .
 .  ----/     .          .     \---- .
F 10---/      .          .      \---F 10  
          rear 1:10   rear 1:10
```		  
Now, when I put any other cables via panels that are mapped as one front port to one back port (like a standard patch panel would be) between the rear ports of device A & B, I can again put a device on front port 1 on device A and it will connect all the way through to a device placed on front port 1 on device B.  However, when I add a second device via a cable to the next front port on device A, I get the above error.  Also, the device port will show up as cabled, even after the error, but the trace will not go to the intended device on the other end and will instead go to the first device built.
```
F 1  ---\                                            /---F 1
   . ----\                                          /---- .
   . -----\_________o_________o__________o_________/----- .
   . -----/rear 1:1  cable 101  cable 102  rear 1:1\----- .
   . ----/    .                               .     \---- .
F 10 ---/     .                               .      \---F 10  
           rear 1:10                       rear 1:10 
```
### Steps to Reproduce
1.  Create two devices of the same device type with multiple front ports mapping to a single back port.
2.  Join the two device's rear ports together via panels that simply map one front port to one back port.
3.  Start putting cables between the front ports on both devices.  The first one should trace through, any subsequent ones will generate and error, appear to have been added anyway, but will not trace through properly.

### Expected Behavior
I was expecting a something connected on device A port 2, to be able to connect to Device B port 2 regardless of whether there are additional cables in between.

### Observed Behavior
Instead I get an error when adding a second connection (first connection will work) and the resulting cable's trace will not map correctly.

steffann:
### Fixes: #3193

I reimplemented `CircuitTermination.trace()` to properly deal with rear ports that have 1 position and front port (just continue with the next cable) and with nested front/rear ports. This supports scenarios like:

```
Intf <-> FP                               FP <-> Intf
         ^                                 ^
       [mux1]                           [mux4]
         v                                 v
         RP <-> FP                 FP <-> RP
                ^                   ^
              [mux2]             [mux3]
                v                   v
                RP <-> Circuit <-> RP
```

### DIFF
--- a/netbox/dcim/models/device_components.py
+++ b/netbox/dcim/models/device_components.py
@@ -113,16 +113,33 @@ class CableTermination(models.Model):
                 (FrontPort, cable, Interface)
             ]
 
-        And an example of a trace where we stop at a rear port:
+        +-----------------+                                               +-----------------+
+        |                 |                                               |                 |
+        |     Device      |                                               |     Device      |
+        |                 |   +-----------------+    +-----------------+  |                 |
+        |  +-------------+|   |+-------------+  |    |  +-------------+|  |+-------------+  |
+        |  |             ||   ||             |  |    |  |             ||  ||             |  |
+        |  | Interface A |-----| FrontPort B |  |    |  | FrontPort E |----| Interface F |  |
+        |  |             ||   ||             |  |    |  |             ||  ||             |  |
+        |  +-------------+|   |+------+------+  |    |  +------+------+|  |+-------------+  |
+        +-----------------+   |       |         |    |         |       |  +-----------------+
+                              |  +----+--------+|    |+--------+----+  |
+                              |  |             ||    ||             |  |
+                              |  |  RearPort C |------| RearPort D  |  |
+                              |  |             ||    ||             |  |
+                              |  +-------------+|    |+-------------+  |
+                              +-----------------+    +-----------------+
 
-            [
-                (RearPort, Cable, FrontPort),
-                (RearPort, Cable, RearPort),
-                (FrontPort, cable, RearPort)
-            ]
+        When tracing from Interface A the trace method will reach RearPort C. If that RearPort only has 1 FrontPort it
+        is basically a cable extender and the trace continues. However when RearPort C has multiple FrontPorts the
+        trace will pause and start a separate trace from RearPort C.
 
-        In this last example we start tracing at a rear port, so the only thing that makes sense is to end at the
-        corresponding rear port.
+        That (sub)trace will reach RearPort D, at which point it doesn't know which FrontPort to continue with, so it
+        ends the (sub)trace there and returns the path from RearPort C to RearPort D. In more complicated examples
+        there can be a whole set of ports between RearPort C and D, which will be (sub)traced accordingly.
+
+        The initial trace resumes, and because it knows which FrontPort it came from, it can resume with FrontPort E,
+        and complete the trace to Interface F.
         """
         if not self.cable:
             return [(self, None, None)]
--- a/netbox/dcim/models/device_components.py
+++ b/netbox/dcim/models/device_components.py
@@ -87,7 +87,14 @@ class CableTermination(models.Model):
     class Meta:
         abstract = True
 
-    def trace(self, position=1, follow_circuits=False, cable_history=None):
+    def get_peer_port(self):
+        """
+        Base method for getting the next port on a trace. Using the example of `trace()` this is for getting from
+        termination B to termination C.
+        """
+        return None
+
+    def trace(self, follow_circuits=False, cable_history=None):
         """
         Return a list representing a complete cable path, with each individual segment represented as a three-tuple:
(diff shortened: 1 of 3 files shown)


netbox/dcim/models/device_components.py
```diff
@@ -139,22 +134,42 @@ def get_peer_port(termination, position=1, follow_circuits=False):
...
 
-        if next_segment is None:
-            return path + [(peer_port, None, None)]
+        # What comes after the other end? (like termination C)
+        other_end_peer_port = other_end.get_peer_port()
+        while other_end_peer_port:
+            path += other_end_peer_port.trace(follow_circuits=follow_circuits, cable_history=cable_history)
+
+            if isinstance(other_end, FrontPort) and \
+                    isinstance(other_end_peer_port, RearPort) and other_end_peer_port.positions > 1:
```
jeremystretch: `other_end_peer_port` in this case can _only_ be a RearPort, correct? Because we know `other_end` is a FrontPort?

netbox/dcim/tests/test_traces.py
```diff
@@ -0,0 +1,250 @@
...
+            site=self.site
+        )
+        self.circuit_term_z = CircuitTermination.objects.create(
+            circuit=self.circuit,
+            port_speed=1,
+            term_side=CircuitTerminationSideChoices.SIDE_Z,
+            site=self.site
+        )
+
+        # And connect everything together in a way that covers all combinations:
```
jeremystretch: Can you add an ASCII topology diagram here for clarity? http://asciiflow.com/ works great for this. Or if you can sketch something out in Visio or similar I'd be happy to convert it.

netbox/dcim/models/device_components.py
```diff
@@ -87,48 +87,43 @@ class CableTermination(models.Model):
     class Meta:
         abstract = True
 
-    def trace(self, position=1, follow_circuits=False, cable_history=None):
+    def get_peer_port(self):
```
jeremystretch: I'm on the fence about this, but I'm leaning toward _not_ declaring this on all device components. IMO it could be confusing to have a `get_peer_port()` method on a model like ConsolePort, which can never have a peer port (though people might understandably expect it to return a ConsoleServerPort). I do appreciate that it makes for cleaner calls, though.

jeremystretch (changes requested):
Still trying to wrap my head around the tests. If you could include a simple topology drawing that would be awesome.

jeremystretch:
Go home stalebot, you're drunk.

netbox/dcim/models/device_components.py
steffann: I included it to cover the case where `other_end_peer_port` is None, but I realise that that is impossible because a front port already has a non-nullable ForeignKey to its rear port. I'd prefer to leave the check in: it doesn't hurt, makes it explicit what type `other_end_peer_port` is, and may catch unforeseen situations if the data model changes in the future.

But you are right: currently `isinstance(other_end_peer_port, RearPort)` will always succeed

netbox/dcim/models/device_components.py
steffann: Rename it to `comes_out_again_on_port()`? :)

steffann: The main reason I added it in the superclass is to cover situations where people connect things together that we currently don't expect. I am sure reality is weirder than I can imagine… With this in the base class the trace method will not malfunction but just stop.

Actually: we want to be able to trace console connections. They can be patched through patch panels as well, so a trace is valuable. I strongly feel we should keep the method but possible rename it.

netbox/dcim/tests/test_traces.py
steffann: Included in the next commit

jeremystretch:
@steffann Thanks for all your work on this. I did end up going with a slightly different approach, however your effort helped me get there.

Unfortunately there remain other problems with the cable model that need to be solved. I'll capture those in a new issue once I've had time to thoroughly review the other outstanding bug reports relating to cables.

PR #3997, opened 2020-01-23
Fixes #3995: Navbar scroll when overflowing

### LINKED ISSUE #3995: Navigation bar too tall
### Environment
* Python version: 3.6.8
* NetBox version: 2.7.2

### Steps to Reproduce
1. Go to a page that is not too long (`/dcim/manufacturers/` or `/dcim/regions/` are good examples as long as you only have a couple of objects)
2. Expand one of the long drop down menus, like `Devices` and have it extend beyond the page's contents
3. Attempt to reach the lower options in the drop down menu 

### Expected Behavior

The navigation bar and its dropdown menus are scrollable

### Observed Behavior

The overflowing elements are unreachable

Dropdown menu
image

Navbar is not scrollable
image

hSaria:
### Fixes: #3995 

Added a max-height to the dropdown menus (80% of the horizontal viewport) and made them scrollable.


jeremystretch:
Closing this as #3995 has not been accepted yet. We can re-open it when/if #3995 is accepted.

netbox/project-static/css/base.css
```diff
@@ -62,8 +62,20 @@ footer p {
     }
 }
 
+/* Scroll the drop-down menus at or above 768px wide to match bootstrap's behavior for hiding dropdown menus */
+@media (min-width: 768px) {
+    .navbar-nav>li>ul {
+        max-height: 80vh;
+        overflow-y: scroll;
```
DanSheps: Perhaps change this to auto instead

netbox/project-static/css/base.css
```diff
@@ -62,8 +62,20 @@ footer p {
...
+    }
+}
+
 /* Collapse the nav menu on displays less than 980px wide */
 @media (max-width: 979px) {
+    #navbar {
+        max-height: 80vh;
+        overflow-y: scroll;
```
DanSheps: Same as above, with auto instead of scroll

netbox/project-static/css/base.css
```diff
@@ -62,8 +62,20 @@ footer p {
...
+@media (min-width: 768px) {
+    .navbar-nav>li>ul {
+        max-height: 80vh;
+        overflow-y: scroll;
+    }
+}
+
 /* Collapse the nav menu on displays less than 980px wide */
 @media (max-width: 979px) {
+    #navbar {
```
DanSheps: We should perhaps use the same selectors if possible.  I don't see any issues with using:  .navbar-nav>li>ul for this as well

netbox/project-static/css/base.css
```diff
@@ -62,8 +62,20 @@ footer p {
     }
 }
 
+/* Scroll the drop-down menus at or above 768px wide to match bootstrap's behavior for hiding dropdown menus */
+@media (min-width: 768px) {
+    .navbar-nav>li>ul {
+        max-height: 80vh;
```
DanSheps: 80vh does have an issue once you get to 150px but I don't believe it is anything to worry about really, just more of a FYI.

DanSheps (changes requested):

netbox/project-static/css/base.css
hSaria: Thanks for pointing it out. I'm on a mac so the scroll bar is only shown when scrolling.

netbox/project-static/css/base.css
hSaria: It's because I wanted the navbar to be scrollable when in the mobile view (the navbar itself should be scrollable, otherwise it'll extend beyond the window).

mobile-navbar-scroll

netbox/project-static/css/base.css
hSaria: The root of the issue is that I'm not subtracting the header height like so
```
max-height: calc(80vh - 50px);
```

If we're confident that we won't be changing the height of the header anytime soon, I'm fine with hardcoding `50px`. Preferably, I'd like to add this in.

hSaria: I've gone and added the above. If anyone objects to hard-coding 50px, speak up. [previously](https://github.com/netbox-community/netbox/blob/98de88de901f5eab752fd1f93cd94a08eb73d281/netbox/project-static/js/forms.js#L396), I've opted to dynamically calculating the header height, but that's not possible in CSS.

Issue #3952, opened 2020-01-17
Create tests for webhooks


jeremystretch:
### Proposed Changes
Create new unit tests for:
1. Create/updating/deleting webhooks
2. Placing webhook events in the queue
3. Transmitting webhooks

### Justification
We have a huge gap in our test coverage around webhooks currently, leading to bugs such as #3951.

jeremystretch:
I've added tests for the enqueuing process in c6eb40daa8dbc3ed7c70c4835089e24ccb8301eb, but I'm not sure how to go about testing the RQ worker process. @lampwins any ideas?

hSaria:
`unittest.mock.patch` the `requests.send` and checking it is called with the expected arguments is an option (I was testing with `requests.post`, but you get the idea).

```python
>>> from unittest.mock import patch
>>> import requests
>>>
>>> def send():
...     requests.post('http://localhost')
... 
>>> @patch('requests.post')
... def test_send(mock_post):
...     send()
...     mock_post.assert_called_with('http://localhost')
... 
>>> test_send()
>>> 
>>> def send():
...     requests.post('http://not_localhost')
... 
>>> test_send()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py", line 1255, in patched
    return func(*args, **keywargs)
  File "<stdin>", line 4, in test_send
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py", line 873, in assert_called_with
    raise AssertionError(_error_message()) from cause
AssertionError: Expected call: post('http://localhost')
Actual call: post('http://not_localhost')
>>> 
```

PR #4001, opened 2020-01-23
Fixes: #738: Automatically check for new versions

### LINKED ISSUE #738: Automatically check for new releases
NetBox should periodically check for a new release and inform the user if one exists. PR #737 was submitted to introduce this feature, but I'd like to see a more efficient approach. This will require implementing a database model to store temporary configuration information. We'll also need to introduce a permanent configuration variable which allows a user to disable this feature.

steffann:
### Fixes: #738

This fetches the latest release from the GitHub API and uses the now mandatory Redis cache to cache the result for 8 hours. Settings to change the URL, change the cache timeout or to disable the feature are provided.


steffann:
I figured I might work on the backlog and take care of the oldest open issue ;)

netbox/utilities/versions.py
```diff
@@ -0,0 +1,27 @@
...
+            'Accept': 'application/vnd.github.v3+json',
+        }
+        try:
+            response = requests.get(url, headers=headers)
+            versions = [(version.parse(release['tag_name']), release.get('html_url'))
+                        for release in response.json()
+                        if 'tag_name' in release]
+            if versions:
+                return max(versions)
+        except:
```
DanSheps: This is throwing a PEP8 error, could you change instead to:  ```except Exception:```

DanSheps (changes requested):

netbox/utilities/versions.py
steffann: Done

netbox/utilities/versions.py
```diff
@@ -0,0 +1,27 @@
...
+            if versions:
+                return max(versions)
+        except Exception:
+            pass
+
+        return 'unknown', None
+
+else:
+    def get_latest_version():
+        return None
```
hSaria: This will throw a `cannot unpack non-iterable NoneType object` error when it is called.
```python
github_latest_version, github_url = get_latest_version()
```

Maybe move the condition to inside the function 

```python
def get_latest_version():
    if settings.GITHUB_VERSION_TIMEOUT and settings.GITHUB_REPOSITORY:
        # check the version and attempt to return the newest one
        ...

    return 'unknown', None
```

netbox/utilities/versions.py
```diff
@@ -0,0 +1,27 @@
...
+        headers = {
+            'Accept': 'application/vnd.github.v3+json',
+        }
+        try:
+            response = requests.get(url, headers=headers)
+            versions = [(version.parse(release['tag_name']), release.get('html_url'))
+                        for release in response.json()
+                        if 'tag_name' in release]
+            if versions:
+                return max(versions)
```
hSaria: This will return a beta version, like `2.7.2 > 2.8-beta1`. Something like this would address that, but there might be a better way.
```
max([v for v in versions if not v[0].is_devrelease and not v[0].is_prerelease])
```

netbox/utilities/versions.py
steffann: The reason I put the `if` outside the function is that using the decorator with `timeout=0` will raise an exception. I also don't want to cache whether the setting is turned on/off as that would confuse users.

As the added complexity is caused by using a decorator I'll rewrite it to use the `cache` object directly. This makes the code a lot more readable.

netbox/utilities/versions.py
steffann: Good catch! I used your solution.

netbox/utilities/versions.py
steffann: I switched to another solution: split the function into two: `get_releases`, which uses the decorator with a minimum `timeout=1` and `get_latest_release` which is used by the context processor.

netbox/utilities/versions.py
```diff
@@ -0,0 +1,35 @@
...
+        releases = [(version.parse(release['tag_name']), release.get('html_url'))
+                    for release in response.json()
+                    if 'tag_name' in release]
+    except Exception:
+        releases = []
+
+    if not pre_releases:
+        releases = [(release, url)
+                    for release, url in releases
+                    if not release.is_devrelease and not release.is_prerelease]
```
jeremystretch: This conditional logic can be moved up to avoid looping through the list of releases twice:

```python
if 'tag_name' in release and (not release.is_devrelease and not release.is_prerelease) or pre_releases
```

However, it's probably worth ditching list comprehension in favor of a regular `for ... in` stanza for the sake of clarity.

netbox/utilities/versions.py
```diff
@@ -0,0 +1,35 @@
+import requests
+from cacheops import cached
+from django.conf import settings
+from packaging import version
+
+
+@cached(timeout=settings.GITHUB_VERSION_TIMEOUT if settings.GITHUB_VERSION_TIMEOUT > 0 else 1)
```
jeremystretch: It doesn't seem prudent to default to a 1 second timeout here, as that would result in a new check for virtually every page load. This isn't a great place for validation, anyway: `GITHUB_VERSION_TIMEOUT` should be validated immediately upon import in `settings.py`, so that any code which references it can be sure to receive a valid value.

netbox/netbox/settings.py
```diff
@@ -77,6 +78,8 @@
 EMAIL = getattr(configuration, 'EMAIL', {})
 ENFORCE_GLOBAL_UNIQUE = getattr(configuration, 'ENFORCE_GLOBAL_UNIQUE', False)
 EXEMPT_VIEW_PERMISSIONS = getattr(configuration, 'EXEMPT_VIEW_PERMISSIONS', [])
+GITHUB_REPOSITORY = getattr(configuration, 'GITHUB_REPOSITORY', 'netbox-community/netbox')
+GITHUB_VERSION_TIMEOUT = getattr(configuration, 'GITHUB_VERSION_TIMEOUT', 8 * 3600)
```
jeremystretch: Per my comment below, let's do validation on `GITHUB_VERSION_TIMEOUT` here. I want to be especially cautious with this setting because it effects calls to an external system.

```suggestion
GITHUB_VERSION_TIMEOUT = int(getattr(configuration, 'GITHUB_VERSION_TIMEOUT', DEFAULT_GITHUB_VERSION_TIMEOUT))

try:
    if GITHUB_VERSION_TIMEOUT and GITHUB_VERSION_TIMEOUT < MINIMUM_GITHUB_VERSION_TIMEOUT:
        raise ImproperlyConfigured("GITHUB_VERSION_TIMEOUT must be at least {} seconds".format(MINIMUM_GITHUB_VERSION_TIMEOUT)
    except ValueError:
        raise ImproperlyConfigured("GITHUB_VERSION_TIMEOUT must be an integer")
```

`DEFAULT_GITHUB_VERSION_TIMEOUT` and `MINIMUM_GITHUB_VERSION_TIMEOUT` should be defined in `extras/constants.py`.

jeremystretch: IMO `MINIMUM_GITHUB_VERSION_TIMEOUT` should be several (4?) hours. It might make sense to introduce a simple management command to manually force a version check, too. This would be exempt from the minimum timeout.

docs/configuration/optional-settings.md
```diff
@@ -135,6 +135,22 @@ Enforcement of unique IP space can be toggled on a per-VRF basis. To enforce uni
 
 ---
 
+## GITHUB_REPOSITORY
```
jeremystretch: We'll also need a mechanism to specify the remote server, for organizations which maintain a fork on a local GitHub Enterprise server. (I'm assuming the same API is available.)

jeremystretch: Whether it makes sense to introduce a separate config parameter or just change this one to convey the full URL is up for discussion; I vote for the later.

netbox/templates/_base.html
```diff
@@ -50,7 +50,16 @@ <h4><i class="fa fa-exclamation-triangle"></i> Maintenance Mode</h4>
...
+                    <p class="text-muted">
+                        {{ settings.HOSTNAME }} (v{{ settings.VERSION }})
+                        {% if latest_version %}
+                            {% if latest_version_url %}<a target="_blank" href="{{ latest_version_url }}">{% endif %}
+                            <span class="label label-default" style="color: #ffffff; background-color: #ff9800">
+                                New version: {{ latest_version }}
+                            </span>
+                            {% if latest_version_url %}</a>{% endif %}
+                        {% endif %}
+                    </p>
```
jeremystretch: I'm not sure it makes sense to include this on _every_ page. I would think a notification on the main page would be sufficient. It would also be helpful to include the date and time at which the latest version was last fetched.

It might also be appropriate to display the message only to administrative (`is_staff` or `is_superuser`) users.

netbox/utilities/context_processors.py
```diff
@@ -8,3 +11,23 @@ def settings(request):
     return {
         'settings': django_settings,
     }
+
+
+def latest_version(request):
```
jeremystretch: This will need a unit test. We'll need to ensure the test uses dummy data and does not actually reach out to GitHub for every run.

jeremystretch (changes requested):
Some of these points probably deserve more discussion, but here are my thoughts at the moment.

netbox/netbox/settings.py
tb-killa: Please also keep an eye on the "rate-limits" of github: https://developer.github.com/v3/#rate-limiting
``For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests.``

steffann: This makes one request every 8 hours, so no problems there

netbox/utilities/versions.py
steffann: The result is cached for hours, so iterating over the list twice is negligible. I personally feel that list comprehension is clearer than a loop, but that's personal.

netbox/utilities/versions.py
jeremystretch: hSaria actually posted a great example of how to work around this in a test here: https://github.com/netbox-community/netbox/issues/3952#issuecomment-575777381

steffann:
I implemented most of the suggestions. Still to do:
- tests
- showing when the version was released
- some way to dismiss the notice
- optimisation

Feedback appreciated before I continue

docs/configuration/optional-settings.md
```diff
@@ -135,6 +135,22 @@ Enforcement of unique IP space can be toggled on a per-VRF basis. To enforce uni
 
 ---
 
+## GITHUB_REPOSITORY_API
```
jeremystretch: Let's avoid using `GITHUB` in the parameter names: We might extend this at some point in the future to work with other services, and it would be nice not to have to rename them.

Maybe `UPDATE_REPO_URL` and `UPDATE_CACHE_TIMEOUT`? Open to other suggestions.

docs/configuration/optional-settings.md
```diff
@@ -135,6 +135,22 @@ Enforcement of unique IP space can be toggled on a per-VRF basis. To enforce uni
...
+
+Default: 'https://api.github.com/repos/netbox-community/netbox'
+
+The releases of this repository are checked to detect new releases, which are shown on the home page of the web interface. You can change this to your own fork of the NetBox repository, or set it to `None` to disable the check.
```
jeremystretch: We should call out that the URL provided MUST adhere to GitHub's API. (I haven't checked but I assume this would not work against GitLab, for example.)

docs/configuration/optional-settings.md
```diff
@@ -135,6 +135,22 @@ Enforcement of unique IP space can be toggled on a per-VRF basis. To enforce uni
...
+
+---
+
+## GITHUB_CACHE_TIMEOUT
+
+Default: 24 * 3600
```
jeremystretch:
```suggestion
Default: 86,400 (24 hours)
```

netbox/netbox/settings.py
```diff
@@ -288,6 +293,30 @@ def _setting(name, default=None):
...
+        raise ImproperlyConfigured("GITHUB_REPOSITORY_API must be a valid http:// or https:// URL")
+
+    if not re.fullmatch(r'/repos/[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+', path):
+        raise ImproperlyConfigured(
+            "GITHUB_REPOSITORY must contain the base URL of the GitHub API in a form like "
+            "'https://api.github.com/repos/<owner>/<repository>'"
+        )
+
+    if query or fragment:
+        raise ImproperlyConfigured("GITHUB_REPOSITORY_API may not contain a query or fragment")
```
jeremystretch: This error message is unlikely to make sense to the user out of context. I don't even know what would trigger it and I'm looking at the code right now. What are we checking for?

netbox/netbox/settings.py
```diff
@@ -288,6 +293,30 @@ def _setting(name, default=None):
...
+    GITHUB_REPOSITORY_API = GITHUB_REPOSITORY_API.rstrip('/')
+    try:
+        scheme, netloc, path, query, fragment = urlsplit(GITHUB_REPOSITORY_API)
+    except ValueError:
+        raise ImproperlyConfigured("GITHUB_REPOSITORY_API must be a valid URL")
+
+    if scheme not in ('http', 'https'):
+        raise ImproperlyConfigured("GITHUB_REPOSITORY_API must be a valid http:// or https:// URL")
+
+    if not re.fullmatch(r'/repos/[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+', path):
```
jeremystretch: Might this be a little _too_ strict? I'm all for validation but I do worry that we might be shooting ourselves in the foot. For instance, does git/GitHub allow Unicode in repo names? I think if we keep this in place we should include a comment pointing to an authoritative spec for the naming format.

netbox/netbox/views.py
```diff
@@ -248,11 +251,25 @@ def get(self, request):
 
         }
 
+        new_release = None
+        new_release_url = None
+
+        if request.user.is_staff:
```
jeremystretch: Probably want to check for superuser status also, since superuser status oddly enough does not imply `is_staff`:

```
>>> jstretch = User.objects.get(username='jstretch')
>>> jstretch.is_superuser
True
>>> jstretch.is_staff
False
```

netbox/utilities/releases.py
```diff
@@ -0,0 +1,35 @@
...
+    headers = {
+        'Accept': 'application/vnd.github.v3+json',
+    }
+    try:
+        response = requests.get(url, headers=headers)
+        releases = [(version.parse(release['tag_name']), release.get('html_url'))
+                    for release in response.json()
+                    if 'tag_name' in release]
+    except Exception:
+        releases = []
```
jeremystretch: We should log the request failure here.

netbox/templates/home.html
```diff
@@ -1,6 +1,18 @@
 {% extends '_base.html' %}
 {% load helpers %}
 
+{% block header %}
+    {% if new_release %}
```
jeremystretch: I think we need to include `{{ block.super }}` here to ensure we don't overwrite any existing content (e.g. an error message from a previous request).

jeremystretch (changes requested):
Awesome work so far! Just some minor comments and questions.

lampwins:
Looking good so far! In speaking with the other maintainers, we feel we should try to do the actual request in the background RQ worker. It is reasonable to think that a large number of NetBox installs are in environments in which outbound internet access is restricted. Although we have already limited the impact of the actual request, it will still come as a surprise on some user requests. Using the cache makes this super simple. Basically, if you get a cache miss, enqueue a request to the background worker to make the call to github. The task will set the value in the cache so on subsequent user requests that result in a cache hit, you display the result to them.

To do this, I would create a new file `utilities/background_tasks.py`. Move your `get_releases()` method here and apply the `@job('default')` decorator from `django_rq`. You can see an example today in [webhooks](https://github.com/netbox-community/netbox/blob/develop/netbox/extras/webhooks_worker.py#L11). You likely will have switch the caching mechanism to use the manual interface that cacheops exposes, instead of the decorator:

```python
from cacheops import cache


cache.set('netbox_releases', release_data)
```

You would then modify your `get_latest_release()` method to handle cache misses and enqueue new requests:

```python
from cacheops import cache
from django_rq import get_queue


def get_latest_release(pre_releases=False):
    if settings.GITHUB_REPOSITORY_API:
        releases = cache.get('netbox_releases')
        if releases:
            return max(releases)
        else:
            background_queue = get_queue('default')
            background_queue.enqueue(
                'utilities.background_tasks.get_releases',
                pre_releases=pre_releases
            )
```

netbox/netbox/settings.py
```diff
@@ -77,6 +79,9 @@
 EMAIL = getattr(configuration, 'EMAIL', {})
 ENFORCE_GLOBAL_UNIQUE = getattr(configuration, 'ENFORCE_GLOBAL_UNIQUE', False)
 EXEMPT_VIEW_PERMISSIONS = getattr(configuration, 'EXEMPT_VIEW_PERMISSIONS', [])
+GITHUB_REPOSITORY_API = getattr(configuration, 'GITHUB_REPOSITORY_API',
```
jeremystretch: This behavior should be disabled by default to avoid any surprises for existing installations. Let's set `GITHUB_REPOSITORY_API` to `None` by default, which will require users to take a direct action to enable the functionality.

jeremystretch:
@steffann Are you still actively working on this? If not, should we move it to a local branch so one of the maintainers can wrap it up?

steffann:
I'll pick this up later this week!

netbox/netbox/settings.py
steffann: I got this regex from GitHub docs somewhere, but I don't remember where… :/

steffann: I have tested all characters that I could type on my keyboard (including a whole bunch of unicode) and every single character except those in the regex above are either replaced with `-` automatically or ignored by GitHub. I am therefore confident that the regex is accurate.

netbox/netbox/settings.py
steffann: This checks that the URL provided doesn't contain a query (`URL?q=bla`) or a fragment (`URL#somewhere`).

steffann:
Rebased to develop and addressed all the comments except for the unit tests

jeremystretch:
I've moved this into a local branch to finish up some implementation details.

PR #4007, opened 2020-01-24
Closes #3880: Define constants for arbitrary values

### LINKED ISSUE #3880: Replace magic numbers with constants
### Proposed Changes
We have a number of instances in which we define "magic numbers" with little or no provided context. Particularly surrounding numeric min/max validators. We should properly define these in constants.py instead.

### Justification
Defining these values as constants will allow for better descriptive use and also proper reuse throughout the codebase.

jeremystretch:
### Fixes: #3880 

Replace statically-defined arbitrary values with named constants for better maintainability. This might not catch everything, but IMO it's sufficient to close #3880 if we're vigilant about the use of constants from here forward.


lampwins (approved):
LGTM

hSaria:
Just thought I'd mention the 1.732 regarding three-phase power calculation. It's no big deal, but I noticed it while working on the recursive power calculation PR. I'm not even sure what that constant would be called (it's the square root of 3 AFAIK).

PR #4012, opened 2020-01-25
Fixes #3338: Added termination A and Z to the circuit

### LINKED ISSUE #3338: Return site in circuit-list API
### Environment
* Python version:  3.6.8
* NetBox version:  2.5.13

### Proposed Functionality
Add `a_site` and `z_site` keys to output of `/circuits/circuits`

### Use Case
In the GUI when viewing the circuit listing, the table has a column that shows the site names for each termination (if it exists). This is extremely useful for filtering by site and generally viewing the circuit list.  The API for circuits does not return this matched data however. In order to get the site, you also have to call `/circuits/circuit-terminations/` as well, then pair the data up. This is a slow workflow within a script, and even more-so when trying to represent the data in an external tool.

The circuit API does allow you to filter by sites so it would also make it more manageable to know which circuit belongs to which site if you're filtering on multiple sites, without having to do even more API calls.

### Database Changes
N/A

### External Dependencies
N/A

hSaria:
### Fixes: #3338 <ISSUE NUMBER GOES HERE>

Added termination A and Z to the circuit API (i.e. `/api/circuits/circuits`).

Removed the `circuit` and `term_side` from `NestedCircuitTerminationSerializer` because they're redundant.

Added `site`, `connected_endpoint`, `port_speed`, `upstream_speed`, and `xconnect_id` as they seem relevant, though I've not preference on that front.

> `NestedCircuitTerminationSerializer` was never used before, so it won't be breaking any existing functionality.

### DIFF
--- a/netbox/circuits/api/serializers.py
+++ b/netbox/circuits/api/serializers.py
@@ -3,11 +3,11 @@ from taggit_serializer.serializers import TaggitSerializer, TagListSerializerFie
 
 from circuits.choices import CircuitStatusChoices
 from circuits.models import Provider, Circuit, CircuitTermination, CircuitType
-from dcim.api.nested_serializers import NestedCableSerializer, NestedSiteSerializer
+from dcim.api.nested_serializers import NestedCableSerializer, NestedInterfaceSerializer, NestedSiteSerializer
 from dcim.api.serializers import ConnectedEndpointSerializer
 from extras.api.customfields import CustomFieldModelSerializer
 from tenancy.api.nested_serializers import NestedTenantSerializer
-from utilities.api import ChoiceField, ValidatedModelSerializer
+from utilities.api import ChoiceField, ValidatedModelSerializer, WritableNestedSerializer
 from .nested_serializers import *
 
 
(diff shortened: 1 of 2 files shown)


netbox/circuits/api/nested_serializers.py
```diff
@@ -47,8 +48,9 @@ class Meta:
 
 class NestedCircuitTerminationSerializer(WritableNestedSerializer):
```
jeremystretch: `NestedCircuitTerminationSerializer` is used by other views. We cannot modify it arbitrarily. Extending the `CircuitSerializer` to include a circuit's terminations will likely require the introduction of a new serializer specifically for this endpoint.

netbox/circuits/api/serializers.py
```diff
@@ -44,13 +44,15 @@ class CircuitSerializer(TaggitSerializer, CustomFieldModelSerializer):
...
+    termination_a = NestedCircuitTerminationSerializer(read_only=True)
+    termination_z = NestedCircuitTerminationSerializer(read_only=True)
     tags = TagListSerializerField(required=False)
 
     class Meta:
         model = Circuit
         fields = [
             'id', 'cid', 'provider', 'type', 'status', 'tenant', 'install_date', 'commit_rate', 'description',
-            'comments', 'tags', 'custom_fields', 'created', 'last_updated',
+            'termination_a', 'termination_z', 'comments', 'tags', 'custom_fields', 'created', 'last_updated',
```
jeremystretch: The API view's queryset needs to be extended to prefetch these values. See `circuits.views.CircuitListView` for reference.

jeremystretch (changes requested):

netbox/circuits/api/nested_serializers.py
hSaria: > `NestedCircuitTerminationSerializer` is used by other views.
Really? I didn't find any places where it was referenced. Had I known it was used, I wouldn't have removed anything from it to avoid breaking existing functionality (i.e. only addition of fields).

jeremystretch: It unfortunately escapes IDE attribution, but it is used dynamically by CableSerializer (to represent a cable connected to a CircuitTermination). It may be used elsewhere.

hSaria: Aha! Good catch. Would you prefer that I restore the removed fields and have it so this same serializer is shared between the two, or just create another one as you suggested? Now that I know it's already used, I'd go with the latter just to be safe.

jeremystretch: Best to create a new one. Silly as it sounds, you should name it `CircuitCircuitTerminationSerializer`. We've had to do this in other places as well, such as `IPAddressNestedInterfaceSerializer`.

hSaria: PR updated accordingly.

netbox/circuits/api/serializers.py
hSaria: Done. Thanks for catching this.

jeremystretch (approved):

PR #4013, opened 2020-01-25
Fixes #3249:  Description length

### LINKED ISSUE #3249: Interface Description max_length
### Environment
* Python version:  
* NetBox version:  2.5.13

### Proposed Functionality
Allow for 256 characters in interface descriptions

### Use Case
For devices with 'ifalias long' enabled we're able to exceed the current maximum of 100 characters and we use interface descriptions for aggregating pertinent data tech would need to know when generating alerts

### Database Changes
I know and understand the argument for maintaining close control over interface name lengths etc. from #1817 but the description is more like a custom field in many respects

### External Dependencies
None AFAIK

hSaria:
### Fixes: #3249

Increased the description `max_length` from 100 to 256 (see #3249 for info on why that number).

In the process, I've replaced that number with a constant which addresses an inconsistency in `ExportTemplate` where description had 200.


jeremystretch:
More discussion is needed around standardizing field lengths before any work can proceed on this. However it's not a priority at the moment.

PR #4014, opened 2020-01-26
Fixes #3886: Config context cluster (group)

### LINKED ISSUE #3886: Cluster assignment under Config Context
### Environment
* Python version:  3
* NetBox version:  2.6.11

### Proposed Functionality
It is currently not possible to assign config context to cluster.
I would expect to have the possibility to assign cluster under `/extras/config-contexts/1/`.
Devices and virtual machines(?) would have the assigned config context.

### Use Case
Devices in a cluster could have their own cluster configuration. If you have various cluster configuration you could easily assign them to the cluster which affects to all devices and maybe virtual machines.

### Database Changes
I think this will need some small database changes. New cluster assignment field and references must be stored.

### External Dependencies
None

hSaria:
### Fixes: #3886

Added cluster and cluster group to config context.

> I may need to update the migration depending on whether this goes before or after #4013.


netbox/extras/forms.py
```diff
@@ -340,6 +346,21 @@ class ConfigContextFilterForm(BootstrapMixin, forms.Form):
             value_field="slug",
         )
     )
+    cluster_group = FilterChoiceField(
+        queryset=TenantGroup.objects.all(),
```
kobayashi: This should be `queryset=ClusterGroup.objects.all()`

netbox/extras/forms.py
```diff
@@ -340,6 +346,21 @@ class ConfigContextFilterForm(BootstrapMixin, forms.Form):
...
+        to_field_name='slug',
+        widget=APISelectMultiple(
+            api_url="/api/virtualization/cluster-groups/",
+            value_field="slug",
+        )
+    )
+    cluster_id = FilterChoiceField(
+        queryset=Tenant.objects.all(),
```
kobayashi: also `queryset=Cluster.objects.all()`

kobayashi (changes requested):

netbox/extras/forms.py
hSaria: Thanks. I missed during my copy-foo.

jeremystretch (approved):

PR #4022, opened 2020-01-27
Fixes #4010: Fixes IP addresses table when filtering interfaces

### LINKED ISSUE #4010: IP Addresses not shown on filtered Interface list
### Environment
* Python version:  3.5.3
* NetBox version:  2.7.2

### Steps to Reproduce
1. Choose a Device 
2. Scroll down to Interfaces and set a filter
3. Check 'Show IPs'

### Expected Behavior
The IP Addresses for the filtered Interface will be displayed.

### Observed Behavior
Only the Headers for the IP Information are displayed (IP Address | Status/Role | VRF | Description)

hSaria:
### Fixes: #4010

The IP addresses table and its toggle were not being considered by the filter.

### DIFF
--- a/netbox/project-static/js/interface_toggles.js
+++ b/netbox/project-static/js/interface_toggles.js
@@ -2,9 +2,9 @@
 $('button.toggle-ips').click(function() {
     var selected = $(this).attr('selected');
     if (selected) {
-        $('#interfaces_table tr.ipaddresses').hide();
+        $('#interfaces_table tr.interface:visible + tr.ipaddresses').hide();
     } else {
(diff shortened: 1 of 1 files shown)


jeremystretch:
I've found a small problem:

1. Apply an interface filter
2. Uncheck "show IPs"
3. Clear the interface filter

After doing so, the IP address re-appear and the button is left unchecked.

hSaria:
Thanks for spotting this. I've restricted the selector for the interfaces to only rows with the `interface` class and then added handling around the next row (IP addresses) if it's present.

PR #4023, opened 2020-01-27
Fixes #4024: Update nginx documentation

### LINKED ISSUE #4024: Update the nginx documentation
### Change Type
[ ] Addition
[X] Correction
[ ] Deprecation

### Proposed Changes

In the nginx configuration, the line below:

```
proxy_set_header X-Forwarded-Host $server_name;
```

should be replaced by this line to take account of port used to reach netbox:

```
proxy_set_header X-Forwarded-Host $http_host;
```

In the previous configuration, nginx assume that you are calling it on port 80. So the next field on the JSON response when you call the API only contains port 80.

smutel:
### Fixes: #4024

With the current nginx configuration, the server port cannot be changed. If you change the port, the first call to the API will work but the second one (from next field of JSON response) will not work because the port used for this request will be always 80.

More details on this discussion: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/netbox-discuss/Jyefr2hO14E/93s6SdvoEQAJ

### DIFF
--- a/docs/installation/3-http-daemon.md
+++ b/docs/installation/3-http-daemon.md
@@ -29,7 +29,7 @@ server {
 
     location / {
         proxy_pass http://127.0.0.1:8001;
-        proxy_set_header X-Forwarded-Host $server_name;
+        proxy_set_header X-Forwarded-Host $http_host;
         proxy_set_header X-Real-IP $remote_addr;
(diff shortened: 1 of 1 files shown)


jeremystretch:
#4024 has not been accepted. This may be re-opened if it is.

PR #4044, opened 2020-01-29
Fix custom script fields being required if they're not

### LINKED ISSUE #4043: Setting required=False on a custom script variable has no effect
Using NetBox 2.7.3 all fields in custom scripts are required (in the html input the required flag is set, when sending it anyways NetBox reports that the field is required).

# Environment
* Python version:  3.7.6
* NetBox version:  2.7.3

### Steps to Reproduce
1. Have a custom script with a field with `required = False`
2. Try leaving the field empty
3. Fail

### Expected Behavior
The html input is not required and a submit with the field being empty succeeds.

### Observed Behavior
The html input is required and a submit with the field being empty fails.

LinzerToertchen:
Fixes #4043

### DIFF
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -60,7 +60,7 @@ class ScriptVariable:
         if default:
             self.field_attrs['initial'] = default
         if required:
-            self.field_attrs['required'] = True
+            self.field_attrs['required'] = required
         if 'validators' not in self.field_attrs:
             self.field_attrs['validators'] = []
 


hSaria:
This doesn't fix it. It's set to `True` only `if required` (see the condition above that statement).

hSaria:
Something like this will
```diff
diff --git a/netbox/extras/scripts.py b/netbox/extras/scripts.py
index bd7e864..a48269b 100644
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -59,10 +59,9 @@ class ScriptVariable:
             self.field_attrs['label'] = label
         if default:
             self.field_attrs['initial'] = default
-        if required:
-            self.field_attrs['required'] = True
         if 'validators' not in self.field_attrs:
             self.field_attrs['validators'] = []
+        self.field_attrs['required'] = bool(required)
 
     def as_field(self):
         """
```

netbox/extras/scripts.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, label='', description='', default=None, required=True):
         if default:
             self.field_attrs['initial'] = default
         if required:
-            self.field_attrs['required'] = True
+            self.field_attrs['required'] = required
         if 'validators' not in self.field_attrs:
             self.field_attrs['validators'] = []
```
DanSheps: As per @hSaria, please make the changes to this code

DanSheps (changes requested):

jeremystretch:
Whoops, looks like we all jumped on this at once. Apologies!

PR #4053, opened 2020-01-30
Closes #4051: Disable the makemigrations management command

### LINKED ISSUE #4051: Override Django's management commands to suppress suggestion to create migrations
### Proposed Changes
Alter or remove some of Django's built-in management commands to simplify the process of applying schema migrations for users. It might be best to introduce a "developers only" configuration parameter, and simply remove the `makemigrations` command from the list of available commands unless that parameter is set.

### Justification
When a user does not follow the documented upgrade procedure, they may encounter a scenario where Django warns them of missing migrations, and prompts them to create new migrations. As this should *never* be required of an end user, we should remove that ability to help preserve the integrity of the installation environment.

jeremystretch:
### Fixes: #4051

Disable the `manage.py makemigrations` command unless `configuration.DEVELOPER` evaluates to True.


jeremystretch:
I've intentionally omitted the new configuration parameter from the documentation, since some people (raises hand) like to enable this sort of thing for fun.

Issue #2562, opened 2018-11-07
IP Prefix in VRF showing in all parent global container prefixes as well as master prefix list


DanSheps:
### Environment
* Python version:  3.4
* NetBox version:  2.3.7

Prefixes that are within a global container also visible on the main prefix list page, as well as subsequent child prefix lists within a prefix.

Related: #1073 

### Steps to Reproduce
1. Create a IP Prefix in the **Global** as a **container**: 10.0.0.0/8
1. Create a IP Prefix in the **Global** as a **any status**: 192.16.0.0/16
2. Create a Child IP Prefix of 10.0.0.0/8 in the **Global** as a **container**: 10.0.0.0/16
2. Create a Child IP Prefix of 10.0.0.0/8 in the **Global** as a **active**: 10.1.0.0/16
3. Create a Child IP Prefix of 10.0.0.0/16 in **VRF** as a **active** network: 10.0.0.0/24
3. Create a Child IP Prefix of 10.0.0.0/16 in **VRF** as a **active** network: 10.1.0.0/24
3. Create a Child IP Prefix of 10.0.0.0/16 in **VRF** as a **container** network: 10.0.1.0/24
4. Create a Child IP Prefix of 10.1.0.0/24 in **VRF** as a **active** network: 10.1.0.0/25
4. Create a Child IP Prefix of 10.1.0.0/16 in **VRF** as a **active** network: 10.1.1.0/24

### Expected Behavior

* 10.0.0.0/24 visible as a child of 10.0.0.0/16
* 10.1.0.0/24 visible as a child of 10.0.0.0/16
* 10.1.0.0/25 visible as a child of 10.1.0.0/24

### Observed Behavior

Prefix 10.0.0.0/24 is visible in (incorrect):

* The main prefix list page (/ipam/prefixes) (incorrect)
* 10.0.0.0/8 Child Prefixes (incorrect)
* 10.0.0.0/16 Child Prefixes (correct)

Prefix 10.1.0.0/24 is visible in:

* The main prefix list page (/ipam/prefixes) (incorrect)
* 10.0.0.0/8 Child Prefixes (incorrect)
* 10.0.0.0/16 Child Prefixes (correct)

Prefix 10.1.0.0/25 is visible in (this behaviour is correct, just included as a test case):
* 10.1.0.0/24 child prefixes (correct)

DanSheps:
Standby on this, I am unable to duplicate this on a clean install.

DanSheps:
[...]

PR #4068, opened 2020-01-31
Fixes: #4034 - Adds Prefix & IPAddress Ordering Tests

### LINKED ISSUE #4034: Add tests for Prefix and IPAddress ordering
### Proposed Changes

This issue is related to adding tests to ensure the data integrity of the Prefix and IPAddress model ordering behaviour.

This will add the following tests:

* test_prefix_ordering
* test_ipaddress_ordering
* test_prefix-vrf_ordering
* test_ipaddress-vrf_ordering
* test_prefix_complex_ordering

The last test will be expected to have the ordering that results in a "problem" for #2562

### Justification

This is tangently related to #2562 however the main goal is to ensure data integrity.

DanSheps:
### Fixes: #4034

Adds various tests for Prefix ordering and IPAddress tests, specifically to help address #2562

### DIFF
--- a/netbox/ipam/tests/test_ordering.py
+++ b/netbox/ipam/tests/test_ordering.py
@@ -5,6 +5,7 @@ from ipam.models import IPAddress, Prefix, VRF
 
 import netaddr
 
+
 class OrderingTestBase(TestCase):
     vrfs = None
 
@@ -97,7 +98,7 @@ class PrefixOrderingTestCase(OrderingTestBase):
     """
     This function tests a compex ordering of interwoven prefixes and vrfs.  This is the current expected ordering of VRFs
     This includes the testing of the Container status.
-    
+
     The proper ordering, to get proper containerization should be:
         None:10.0.0.0/8
         None:10.0.0.0/16
--- a/netbox/ipam/tests/test_ordering.py
+++ b/netbox/ipam/tests/test_ordering.py
@@ -5,279 +5,177 @@ from ipam.models import IPAddress, Prefix, VRF
 
 import netaddr
 
-
-class PrefixOrderingTestCase(TestCase):
+class OrderingTestBase(TestCase):
     vrfs = None
 
+    """
+    Setup the VRFs for the class as a whole
+    """
     def setUp(self):
-        vrfa = VRF(name="VRF A")
-        vrfb = VRF(name="VRF B")
-        vrfc = VRF(name="VRF C")
-        VRF.objects.bulk_create([vrfa, vrfb, vrfc])
-        self.vrfs = (vrfa, vrfb, vrfc)
+        self.vrfs = (VRF(name="VRF A"), VRF(name="VRF B"), VRF(name="VRF C"))
+        VRF.objects.bulk_create(self.vrfs)
 
-    def _create_prefix(self, prefixes):
-        prefixobjects = []
-        for pfx in prefixes:
-            status, vrf, prefix = pfx
-            family = 4
-            if not netaddr.valid_ipv4(prefix):
-                family = 6
-            pfx = Prefix(prefix=prefix, family=family, vrf=vrf, status=status)
-            prefixobjects.append(pfx)
-
-        return prefixobjects
-
-    def _compare_prefix(self, queryset, prefixes):
-
+    """
+    Perform the comparison of the queryset object and the object used to instantiate the queryset.
+    """
+    def _compare(self, queryset, objectset):
         for i, obj in enumerate(queryset):
-            status, vrf, prefix = prefixes[i]
-            self.assertEqual((obj.vrf, obj.prefix), (vrf, prefix))
+            if isinstance(obj, Prefix):
+                self.assertEqual((obj.vrf, obj.prefix), (objectset[i]['vrf'], objectset[i]['prefix']))
+            elif isinstance(obj, IPAddress):
+                self.assertEqual((obj.vrf, obj.address), (objectset[i]['vrf'], objectset[i]['address']))
 
+
+class PrefixOrderingTestCase(OrderingTestBase):
+
+    """
+    This is for comparing the complex ordering test case
+    """
     def _compare_complex(self, queryset, prefixes):
         qsprefixes, regprefixes = [], []
         for i, obj in enumerate(queryset):
             qsprefixes.append(obj.prefix)
         for pfx in prefixes:
-            regprefixes.append(pfx[2])
-        return (qsprefixes, regprefixes)
+            regprefixes.append(pfx['prefix'])
+        self.assertEquals(qsprefixes, regprefixes)
 
-
-
-    def test_prefix_ordering(self):
-        # Setup Prefixes
-        prefixes = (
-            (PrefixStatusChoices.STATUS_CONTAINER, None, netaddr.IPNetwork('10.0.0.0/8')),
-            (PrefixStatusChoices.STATUS_CONTAINER, None, netaddr.IPNetwork('10.0.0.0/16')),
-            (PrefixStatusChoices.STATUS_ACTIVE, None, netaddr.IPNetwork('10.0.0.0/24')),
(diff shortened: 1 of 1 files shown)


netbox/ipam/tests/test_ordering.py
```diff
@@ -0,0 +1,290 @@
...
+from ipam.models import IPAddress, Prefix, VRF
+
+import netaddr
+
+
+class PrefixOrderingTestCase(TestCase):
+    vrfs = None
+
+    def setUp(self):
+        vrfa = VRF(name="VRF A")
```
jeremystretch: Should be able to simplify this as:

```python
self.vrfs = (
    VRF(name="VRF A"),
    VRF(name="VRF B"),
    VRF(name="VRF C"),
)
VRF.objects.bulk_create(self.vrfs)
```

netbox/ipam/tests/test_ordering.py
```diff
@@ -0,0 +1,290 @@
...
+        vrfb = VRF(name="VRF B")
+        vrfc = VRF(name="VRF C")
+        VRF.objects.bulk_create([vrfa, vrfb, vrfc])
+        self.vrfs = (vrfa, vrfb, vrfc)
+
+    # Function to create all prefixes
+    def _create_prefix(self, prefixes):
```
jeremystretch: Two issues with this:

1. `prefix` implies a singular instance, but it deals with a list of instances.
2. It's not actually _creating_ the prefixes, just instantiating them.

Something like `prefix_from_args()` would be more accurate. IMO it would be fine to have this deal with a single object, and call it inside list comprehension during creation:

```
Prefix.objects.bulk_create([prefix_from_args(args) for args in prefixes])
```

netbox/ipam/tests/test_ordering.py
```diff
@@ -0,0 +1,290 @@
...
+            qsprefixes.append(obj.prefix)
+        # Loop and build list of prefixes to compare
+        for pfx in prefixes:
+            regprefixes.append(pfx[2])
+
+        # Return two lists as a tuple
+        return (qsprefixes, regprefixes)
+
+    def test_prefix_ordering(self):
+        # Setup Prefixes
```
jeremystretch: Let's add docstrings to each test method explaining why each one exists.

netbox/ipam/tests/test_ordering.py
```diff
@@ -0,0 +1,290 @@
...
+            (PrefixStatusChoices.STATUS_ACTIVE, vrfb, netaddr.IPNetwork('172.17.3.0/24')),
+            (PrefixStatusChoices.STATUS_ACTIVE, vrfb, netaddr.IPNetwork('172.17.4.0/24')),
+        )
+        Prefix.objects.bulk_create(self._create_prefix(prefixes))
+
+        # Test
+        self._compare_prefix(Prefix.objects.all(), prefixes)
+
+    def test_prefix_complex_ordering(self):
+        # Setup VRFs
```
jeremystretch: Why not just take this approach with the first two tests? Intermixing the ranges makes for a more robust test and obviates the need for a separate method.

netbox/ipam/tests/test_ordering.py
```diff
@@ -0,0 +1,290 @@
...
+            family = 4
+            if not netaddr.valid_ipv4(address):
+                family = 6
+            addressobj = IPAddress(address=address, vrf=vrf, status=status, family=family)
+            addressobjects.append(addressobj)
+
+        return addressobjects
+
+    # Function to compare all addresses
+    def _compare_address(self, queryset, addresses):
```
jeremystretch: Can we replace this and `_compare_prefix()` above with a single common function?

netbox/ipam/tests/test_ordering.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def setUp(self):
+        vrfa = VRF(name="VRF A")
+        vrfb = VRF(name="VRF B")
+        vrfc = VRF(name="VRF C")
+        VRF.objects.bulk_create([vrfa, vrfb, vrfc])
+        self.vrfs = (vrfa, vrfb, vrfc)
+
+    # Function to build all addresses
+    def _create_address(self, addresses):
```
jeremystretch: Same feedback as `_create_prefix()` above. (Also, can we combine them?)

jeremystretch (changes requested):

netbox/ipam/tests/test_ordering.py
DanSheps: Yeah, it should.  I will get on it later today

netbox/ipam/tests/test_ordering.py
DanSheps: We can definitely do this, my thought was this gives us three tests to see what exactly we broke.  I could wrap the three tests into 1 function however.

netbox/ipam/tests/test_ordering.py
DanSheps: I will see what I can do.

netbox/ipam/tests/test_ordering.py
DanSheps: I have wrapped the first two up into one class, but I think I want to keep this one, as this will help us sort out the nesting of prefix problem in #2562

jeremystretch:
Looks like this has some PEP8 failures.

DanSheps:
Yeah, I missed those

On Feb. 11, 2020 13:24, Jeremy Stretch <notifications@github.com> wrote:

Looks like this has some PEP8 failures.

netbox/ipam/tests/test_ordering.py
```diff
@@ -0,0 +1,182 @@
...
+        for i, obj in enumerate(queryset):
+            if isinstance(obj, Prefix):
+                self.assertEqual((obj.vrf, obj.prefix), (objectset[i]['vrf'], objectset[i]['prefix']))
+            elif isinstance(obj, IPAddress):
+                self.assertEqual((obj.vrf, obj.address), (objectset[i]['vrf'], objectset[i]['address']))
+
+
+class PrefixOrderingTestCase(OrderingTestBase):
+
+    """
```
jeremystretch: Docstrings should go inside their respective methods:

```python
def test_foo(self):
    """
    Docstring goes here
    """

File: netbox/ipam/tests/test_ordering.py [hunk 8]
```diff
@@ -0,0 +1,182 @@
...
+    This is for comparing the complex ordering test case
+    """
+    def _compare_complex(self, queryset, prefixes):
```
jeremystretch: It probably makes sense to move this to `OrderingTestBase` with `_compare()` even if it's only used in the one subclass for now.

netbox/ipam/tests/test_ordering.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    This is a very basic test, which tests both prefixes without VRFs and prefixes with VRFs
+    """
+    def test_prefix_vrf_ordering(self):
+        # Setup VRFs
+        vrfa, vrfb, vrfc = self.vrfs
+
+        # Setup Prefixes
+        prefixes = (
+            {"status": PrefixStatusChoices.STATUS_CONTAINER, "vrf": None, "family": 4, "prefix": netaddr.IPNetwork('192.168.0.0/16')},
```
jeremystretch: Instead of defining all the attrs in dicts first, you can create `Prefix` instances directly in a list, and then pass that list to `Prefix.objects.bulk_create()`. This cuts down on boilerplate somewhat.

```python
prefixes = (
    Prefix(status=PrefixStatusChoices.STATUS_CONTAINER, vrf=None, family=4, prefix=netaddr.IPNetwork('192.168.0.0/16'),
    ...
)
Prefix.objects.bulk_create(prefixes).
```

Plus, you can reference the prefixes by their index on the `prefixes` list even after creation in the database if needed:

```python
prefixes[4].pk
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #4075, opened 2020-02-02
Fixes #3507: Filtering IP by multiple devices

### LINKED ISSUE #3507: API filtering 'OR' logic not working when filtering IP addresses by device ID
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.3

### Steps to Reproduce
Issue an API query with filter parameters specifying two different device IDs, for example:
https://netbox/api/ipam/ip-addresses/?device_id=13441&device_id=13442

### Expected Behavior
API will return IP addresses associated with the devices having ids 13441 and 13442.

### Observed Behavior
API returned only the IP addresses associated with the last given device id, in this case 13442.

kobayashi:
### Fixes: #3507

Filtering IP addresses by device id or device name

### DIFF
--- a/netbox/ipam/filters.py
+++ b/netbox/ipam/filters.py
@@ -304,14 +304,15 @@ class IPAddressFilterSet(TenancyFilterSet, CustomFieldFilterSet, CreatedUpdatedF
         to_field_name='rd',
         label='VRF (RD)',
     )
-    device = django_filters.CharFilter(
-        method='filter_device',
-        field_name='name',
-        label='Device',
+    device = django_filters.ModelMultipleChoiceFilter(
+        field_name='interface__device__name',
+        queryset=Device.objects.all(),
+        to_field_name='name',
+        label='Device (name)',
     )
-    device_id = django_filters.NumberFilter(
-        method='filter_device',
-        field_name='pk',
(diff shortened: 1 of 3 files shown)


netbox/ipam/filters.py
```diff
@@ -304,14 +304,15 @@ class IPAddressFilterSet(TenancyFilterSet, CustomFieldFilterSet, CreatedUpdatedF
...
-    device = django_filters.CharFilter(
-        method='filter_device',
-        field_name='name',
-        label='Device',
-    )
-    device_id = django_filters.NumberFilter(
-        method='filter_device',
-        field_name='pk',
+    device = django_filters.ModelMultipleChoiceFilter(
+        field_name='interface__device__name',
```
jeremystretch: This does not account for IP addresses assigned to interfaces belonging to virtual chassis members, which is why the local `filter_device()` method was added. The filter/method need to be adapted to pass multiple values.

jeremystretch (changes requested):

netbox/ipam/filters.py
kobayashi: Thanks the review, I will fix later.

kobayashi: Enabled filtering virtual chassis members also.

PR #4101, opened 2020-02-06
Fixes #4074: DNS regex validator matches 0 or more

### LINKED ISSUE #4074: Regex validator in field `dns_name` in model `IPAddress` conflict with empty string value
### Environment
* Python version:   3.6.8
* NetBox version:   2.6.6, develop

### Steps to Reproduce
1. Run netbox with default configuration, create an IP Address, example `192.168.10.15/24`, with `dns name`  field is empty
2. Get netbox swagger schema document from url `/api/swagger.yaml`.Create python netbox client by `swagger-codegen` from swagger document
3. Use client to get ipaddress list from netbox api, example:

```python
import netbox_client

netbox_client_config = netbox_client.Configuration()
netbox_client_config.host = #NETBOX_API_URL
netbox_client_config.api_key["Authorization"] = "Token {}".format(
    #TOKEN
)

api_client = netbox_client.ApiClient(netbox_client_config)
ipam_app_api = netbox_client.IpamApi(api_client)
api_resp = ipam_app_api.ipam_ip_addresses_list()
print('done')

```

### Expected Behavior

python client return IP Address object list. No Exceptions raised.

### Observed Behavior

python netbox client raise Exception:

```log
ValueError: Invalid value for `dns_name`, must be a follow pattern or equal to `/^[0-9A-Za-z._-]+$/`
```

when at least one IP Address has empty `dns_name`

### Reason

With current iplementation:

```python

DNSValidator = RegexValidator(
    regex='^[0-9A-Za-z._-]+$',
    message='Only alphanumeric characters, hyphens, periods, and underscores are allowed in DNS names',
    code='invalid'
)

class IPAddress(ChangeLoggedModel, CustomFieldModel):
    #...
    dns_name = models.CharField(
        max_length=255,
        blank=True,
        validators=[DNSValidator],
        verbose_name='DNS Name',
        help_text='Hostname or FQDN (not case-sensitive)'
    )

```

`dns_name` field will generated in swagger document like this:

```yaml
  IPAddress:
    required:
      - address
      dns_name:
        title: DNS Name
        description: Hostname or FQDN (not case-sensitive)
        type: string
        pattern: ^[0-9A-Za-z._-]+$
        maxLength: 255
```

this definition when converted to python client:

```python
    @dns_name.setter
    def dns_name(self, dns_name):
        """Sets the dns_name of this IPAddress.

        Hostname or FQDN (not case-sensitive)  # noqa: E501

        :param dns_name: The dns_name of this IPAddress.  # noqa: E501
        :type: str
        """
        if dns_name is not None and len(dns_name) > 255:
            raise ValueError("Invalid value for `dns_name`, length must be less than or equal to `255`")  # noqa: E501
        if dns_name is not None and not re.search(r'^[0-9A-Za-z._-]+$', dns_name):  # noqa: E501
            raise ValueError(r"Invalid value for `dns_name`, must be a follow pattern or equal to `/^[0-9A-Za-z._-]+$/`")  # noqa: E501

        self._dns_name = dns_name
```

when dns_name is empty or missing, Django automatically set its value in database to empty string `''`. In this scenario, when python client get ipaddress info from netbox api, dns_name will be validated in `dns_name.setter` method and not pass regex validator, and ValueError will be raised.

## Proposal Change

Because we allow dns_name is empty, could we change validator restriction from `1 or many`: `'^[0-9A-Za-z._-]+$'` to `0 or many`: `'^[0-9A-Za-z._-]*$'` ?

kobayashi:
### Fixes: #4074

If the issue is accepted, this can be merged.

### DIFF
--- a/docs/release-notes/version-2.7.md
+++ b/docs/release-notes/version-2.7.md
@@ -16,6 +16,7 @@
 * [#3313](https://github.com/netbox-community/netbox/issues/3313) - Toggle config context display between JSON and YAML
 * [#3886](https://github.com/netbox-community/netbox/issues/3886) - Enable assigning config contexts by cluster and cluster group
 * [#4051](https://github.com/netbox-community/netbox/issues/4051) - Disable the `makemigrations` management command
+* [#4074](https://github.com/netbox-community/netbox/issues/4074) - DNS regex validation matches 0 or more
 
 ## Bug Fixes
 
--- a/netbox/ipam/validators.py
+++ b/netbox/ipam/validators.py
@@ -24,7 +24,7 @@ class MinPrefixLengthValidator(BaseValidator):
 
 
 DNSValidator = RegexValidator(
-    regex='^[0-9A-Za-z._-]+$',
+    regex='^[0-9A-Za-z._-]*$',
     message='Only alphanumeric characters, hyphens, periods, and underscores are allowed in DNS names',
     code='invalid'
 )


jeremystretch:
This is not needed; the current implementation is correct per Django. Please see my comment on #4074.

PR #4102, opened 2020-02-06
Fixes #4093: Additional status choices for vms

### LINKED ISSUE #4093: Status Decommissioning missing for virtual machines
### Environment
* Python version:  3.6.9
* NetBox version:   2.7.4

### Proposed Functionality
Currently a virtual machine can only have as status:
* Active
* Offline
* Staged

I'm missing here at least Decommissioning as status.

### Use Case
I would distinguish between Offline and Decommissioning in this way:

**Offline** is a VM that is temporary powered off and will be later powered on again.
**Decommissioning** is a VM that is powered off and will be later deleted.

Our decommissioning process is actually to power off a VM and then wait a reasonable time if still someone came up who needs this VM.
Being able to set Decommissioning as status would help to track VM's  planned for decommissioning.

Maybe it also makes sense to use for virtual machines the same status as for devices.
If you think of devices as hardware servers, the same processes are often used for hardware servers and VM's. Having the same status choices for both of them would help here.

### Database Changes
None

### External Dependencies
None

kobayashi:
### Fixes: #4093

When the issue is accepted, this PR can be merged.

### DIFF
--- a/netbox/virtualization/choices.py
+++ b/netbox/virtualization/choices.py
@@ -8,15 +8,19 @@ from utilities.choices import ChoiceSet
 
 class VirtualMachineStatusChoices(ChoiceSet):
 
-    STATUS_ACTIVE = 'active'
     STATUS_OFFLINE = 'offline'
+    STATUS_ACTIVE = 'active'
+    STATUS_PLANNED = 'planned'
     STATUS_STAGED = 'staged'
+    STATUS_FAILED = 'failed'
     STATUS_DECOMMISSIONING = 'decommissioning'
 
     CHOICES = (
-        (STATUS_ACTIVE, 'Active'),
         (STATUS_OFFLINE, 'Offline'),
+        (STATUS_ACTIVE, 'Active'),
+        (STATUS_PLANNED, 'Planned'),
         (STATUS_STAGED, 'Staged'),
+        (STATUS_FAILED, 'Failed'),
         (STATUS_DECOMMISSIONING, 'Decommissioning'),
     )
 
--- a/netbox/virtualization/choices.py
+++ b/netbox/virtualization/choices.py
@@ -24,7 +24,6 @@ class VirtualMachineStatusChoices(ChoiceSet):
         STATUS_OFFLINE: 0,
         STATUS_ACTIVE: 1,
         STATUS_STAGED: 3,
-        STATUS_DECOMMISSIONING: 4,
     }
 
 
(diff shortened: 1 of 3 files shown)


netbox/virtualization/choices.py
```diff
@@ -11,17 +11,20 @@ class VirtualMachineStatusChoices(ChoiceSet):
...
         (STATUS_OFFLINE, 'Offline'),
         (STATUS_STAGED, 'Staged'),
+        (STATUS_DECOMMISSIONING, 'Decommissioning'),
     )
 
     LEGACY_MAP = {
         STATUS_OFFLINE: 0,
         STATUS_ACTIVE: 1,
         STATUS_STAGED: 3,
+        STATUS_DECOMMISSIONING: 4,
```
jeremystretch: This is not needed. `LEGACY_MAP` exists only for backward-compatibility with pre-2.7 values; there is no need to populate new values.

netbox/virtualization/models.py
```diff
@@ -267,9 +267,10 @@ class VirtualMachine(ChangeLoggedModel, ConfigContextModel, CustomFieldModel):
     ]
 
     STATUS_CLASS_MAP = {
-        'active': 'success',
-        'offline': 'warning',
-        'staged': 'primary',
+        VirtualMachineStatusChoices.STATUS_ACTIVE: 'success',
+        VirtualMachineStatusChoices.STATUS_OFFLINE: 'warning',
+        VirtualMachineStatusChoices.STATUS_STAGED: 'primary',
+        VirtualMachineStatusChoices.STATUS_DECOMMISSIONING: 'warning',
```
jeremystretch: We should use different classes for the offline and decommissioning statuses. Might want to change offline to `danger` (red).

jeremystretch (changes requested):

netbox/virtualization/choices.py
kobayashi: Thx, removed the line.

netbox/virtualization/models.py
kobayashi: That is better. Changed the color of "OFFLINE" to `danger`

jeremystretch:
@kobayashi Can you take a look at the latest comments on #4093 and let us know what you think about adding the additional statuses?

kobayashi:
Set additional choices for virtual machine as the issue discussion.

netbox/virtualization/models.py
kobayashi: color label of vm status is set the same color as that of device.

Issue #4110, opened 2020-02-06
Rack SVG splits into two at certain screen widths


candlerb:
### Environment
* Python version:  3.5.2
* NetBox version:  2.7.4

### Steps to Reproduce
1. Create a rack with height 47U
2. Go to `/dcim/racks/NNN/`
3. Adjust the width of the window

### Expected Behavior
SVGs to display normally, regardless of screen width.

### Observed Behavior
At a screen width between 992px and 1183px, the rack splits into two, with U labels 1-47 at the top and the actual U slots below. 

![image](https://user-images.githubusercontent.com/44789/73960849-7d0db000-4903-11ea-9c59-bedaa2221905.png)

This is with Chrome 79 under macOS 10.14.6

DanSheps:
Duplicate of #3986

PR #4144, opened 2020-02-12
Fixes: #3986 - Convert rack units to part of SVG rendered document

### LINKED ISSUE #3986: Printing a rack is not as expected
### Environment
* Python version:  3.6.9
* NetBox version:  2.7.2

Hi

Printing a rack on paper and/or pdf show devices under U1.
Here an example of a rack with version 2.7.2
rack-2 7 2

Here, the same rack with version 2.6.11
rack-2 6 11

### Steps to Reproduce
1. choose a rack with devices
2. print it on pdf (or use preview on MacOS)
3. All devices appears under the floor :)

### Expected Behavior
Devices are in the rack

### Observed Behavior
Devices are under the rack

DanSheps:
### Fixes: #3986 & #4110 

Converts the rack legend into part of the SVG rendered document


netbox/dcim/models/__init__.py
```diff
@@ -435,9 +435,20 @@ def _draw_empty(drawing, rack, start, end, text, id_, face_id, class_, reservati
 
     def _draw_elevations(self, elevation, reserved_units, face, unit_width, unit_height):
 
-        drawing = self._setup_drawing(unit_width, unit_height * self.u_height)
+        # Setup vars for the rack unit
+        position_width = 30
```
jeremystretch: We should probably make this a keyword argument on `_draw_elevations()` to ensure flexibility. We can set it to 30 as a default, but it should be defined as a constant like we do with `RACK_ELEVATION_UNIT_WIDTH_DEFAULT` and `RACK_ELEVATION_UNIT_HEIGHT_DEFAULT`.

Also, what do you think about calling this something like `legend_width` instead? Position seems a bit confusing given the context.

netbox/dcim/models/__init__.py
```diff
@@ -471,7 +482,7 @@ def _draw_elevations(self, elevation, reserved_units, face, unit_width, unit_hei
             unit_cursor += height
 
         # Wrap the drawing with a border
-        drawing.add(drawing.rect((0, 0), (unit_width, self.u_height * unit_height), class_='rack'))
+        drawing.add(drawing.rect((30, 0), (unit_width, self.u_height * unit_height), class_='rack'))
```
jeremystretch: `drawing.rect()` should reference a variable instead of `30`.

netbox/dcim/models/__init__.py
```diff
@@ -435,9 +435,20 @@ def _draw_empty(drawing, rack, start, end, text, id_, face_id, class_, reservati
...
-        drawing = self._setup_drawing(unit_width, unit_height * self.u_height)
+        # Setup vars for the rack unit
+        position_width = 30
+
+        drawing = self._setup_drawing(unit_width + position_width, unit_height * self.u_height)
 
         unit_cursor = 0
+        for ru in range(0, self.u_height):
+            start_y = ru * unit_height
+            position_coordinates = (15, start_y + unit_height / 2 + 2)
```
jeremystretch: Where did the `15` come from?

netbox/dcim/models/__init__.py
```diff
@@ -435,9 +435,20 @@ def _draw_empty(drawing, rack, start, end, text, id_, face_id, class_, reservati
...
+            unit = ru + 1 if self.desc_units else self.u_height - ru
+            drawing.add(
+                drawing.text("U" + str(unit), position_coordinates, class_="unit")
```
jeremystretch: We can omit the "U" from the unit number. (The current rendering does not include it.)

jeremystretch (changes requested):

netbox/dcim/models/__init__.py
DanSheps: Yeah, I think legend might be better.

Will do the constants ASAP

netbox/dcim/models/__init__.py
DanSheps: Half of 30, since we are using a variable, changed it to that

jeremystretch (approved):

PR #4152, opened 2020-02-12
Add email testing example 

joshcorr:

This adds documentation on how to test email settings configured in they configuration.py file.


docs/configuration/optional-settings.md
```diff
@@ -109,6 +109,16 @@ In order to send email, NetBox needs an email server configured. The following i
 * TIMEOUT - Amount of time to wait for a connection (seconds)
 * FROM_EMAIL - Sender address for emails sent by NetBox
 
+Email is sent from NetBox only for critical events. If you would like to test the email server configuration please use the following:
+
+```
+# python ./manage.py nbshell
+### NetBox interactive shell (jstretch-workstation)
```
jeremystretch: You can omit the header lines from the output to keep it a bit cleaner.

docs/configuration/optional-settings.md
```diff
@@ -109,6 +109,16 @@ In order to send email, NetBox needs an email server configured. The following i
 * TIMEOUT - Amount of time to wait for a connection (seconds)
 * FROM_EMAIL - Sender address for emails sent by NetBox
 
+Email is sent from NetBox only for critical events. If you would like to test the email server configuration please use the following:
```
jeremystretch: We probably want to include a link to Django's documentation for the [`send_mail()` utility](https://docs.djangoproject.com/en/stable/topics/email/#send-mail) in case the reader wants to explore all its options.

jeremystretch (changes requested):

docs/configuration/optional-settings.md
joshcorr: Removed the headers

docs/configuration/optional-settings.md
joshcorr: Included a link to send_mail() documentation

joshcorr:
@jeremystretch Should be good to go. Let me know if you need anything else changed.

Issue #4159, opened 2020-02-13
Caching utilizes webhooks configuration


DanSheps:
### Environment
* Python version:  2.6
* NetBox version:  develop

### Steps to Reproduce
1. Configure a separate redis server for caching

### Expected Behavior
The separate server is used

### Observed Behavior
The redis server for WEBHOOKS is used

### Technical details
In settings.py, redis for caching uses WEBHOOKS_REDIS instead of CACHING_REDIS

```
CACHING_REDIS = REDIS.get('caching', {})
CACHING_REDIS_HOST = WEBHOOKS_REDIS.get('HOST', 'localhost')
CACHING_REDIS_PORT = WEBHOOKS_REDIS.get('PORT', 6379)
CACHING_REDIS_PASSWORD = WEBHOOKS_REDIS.get('PASSWORD', '')
CACHING_REDIS_DATABASE = WEBHOOKS_REDIS.get('DATABASE', 0)
CACHING_REDIS_DEFAULT_TIMEOUT = WEBHOOKS_REDIS.get('DEFAULT_TIMEOUT', 300)
CACHING_REDIS_SSL = WEBHOOKS_REDIS.get('SSL', False)
```

PR #4161, opened 2020-02-13
Fixes #3984: Allow for Redis Sentinel Connection Configuration

### LINKED ISSUE #3984: Allow Redis Connections to Use Sentinel
### Environment
* Python version:  3.7.3
* NetBox version:  2.7.2

### Proposed Functionality

My proposal is to allow `SENTINELS` and `SERVICE` as optional keys to the `REDIS` dictionaries, like so. If these are both provided and non-empty, then the application would attempt to use the Redis Sentinel connection information. Note that the current way of configuring normal Redis would not change.

```python
# Using the current method 
REDIS = {
    'webhooks': {
        'HOST': 'webhooks_redis',
        'PORT': 6379,
        'PASSWORD': '',
        'DATABASE': 0,
        'DEFAULT_TIMEOUT': 300,
        'SSL': False,
    },
    'caching': {
        'HOST': 'cache_redis',
        'PORT': 6379,
        'PASSWORD': '',
        'DATABASE': 0,
        'DEFAULT_TIMEOUT': 300,
        'SSL': False,
    }
}

# Using a Sentinel instance
REDIS = {
    'webhooks': {
        'SENTINELS': (
            ('sentinel.redis.example.com', 1234),  # Each is defined as (HOST, PORT) tuple
        ),
        'SERVICE': 'webhooks_redis',                   # Sentinel service to connect to
        'PASSWORD': '',
        'DATABASE': 0,
        'DEFAULT_TIMEOUT': 300,
        'SSL': False,
    },
    'caching': {
        'SENTINELS': (
            ('sentinel.redis.example.com', 1234),  # Each is defined as (HOST, PORT) tuple
        ),
        'SERVICE': 'webhooks_redis',                   # Sentinel service to connect to
        'PASSWORD': '',
        'DATABASE': 0,
        'DEFAULT_TIMEOUT': 300,
        'SSL': False,
    }
}
```

### Use Case

It was nice that the webhooks and cache Redis connection settings were separated, but it would be ever nicer to be able to configure one (or both) to use Redis Sentinel connections for the sake of High Availability. Not only may this be preferred, but in some environments, it might be necessary due to the availability of developer environments.

### Database Changes
None.

### External Dependencies
None.

dstarner:
### Fixes: #3984 

Adds functionality and documentation to allow Redis for both the webhook and caching layers to use [Redis Sentinel](https://redis.io/topics/sentinel).


DanSheps:
One quick comment, we should perhaps not have the docs under requirements.md.  Maybe in a separate sentinel configuration doc or something (will wait for @jeremystretch for this), I just think if we include this in requirements it will be viewed as a requirement.

dstarner:
I thought about the location for the docs a little, and I decided to place them under the initial Redis docs since 1) using sentinel for redis does not change the fact that it is required and 2) it reduces searching around the documentation if someone needs to use Sentinel from the get-go

I will fix the linting error (as well as need to update the connection check in the `extras.apps.AppConfig`).

jeremystretch:
Hold off on this until #4159 is fixed please as it will introduce a conflict.

netbox/netbox/settings.py
```diff
@@ -394,28 +407,35 @@ def _setting(name, default=None):
 #
 # Caching
 #
-
-if CACHING_REDIS_SSL:
-    REDIS_CACHE_CON_STRING = 'rediss://'
+if CACHING_REDIS_USING_SENTINEL:
+    CACHEOPS_SENTINEL = {
+        'locations': CACHING_REDIS_SENTINELS,
```
DanSheps: This line appears to be throwing the PEP8 error

DanSheps (changes requested):
Could you also fix up the PEP8 error?

dstarner:
Oops, I missed you comment about holding off...I will remove that last commit and wait for the merge of #4159. I'll fix the linting errors once I rebase and remove any conflicts

dstarner:
Alright @jeremystretch this should be good to go now, thanks for getting the other one in so quickly

jeremystretch:
I've asked @lampwins to sanity-check this as well as he's probably most familiar with the caching configuration.

dstarner:
That's fine. For more context, I am currently running with the same settings changes in our production instance(s) and we haven't had any problems

DanSheps:
In addition to.that, I do think the docs for this particular piece should go in the optional settings doc file

dstarner:
Yea, I mean, it doesn't matter too much to me as long as its *somewhere*, so I don't mind moving it. I'll update it :)

DanSheps:
We can leave it as is, I was just worried some people might see it and think it is something that is required

dstarner:
I guess I'll leave it as is unless anyone has a strong opinion one way or another.

PR #4169, opened 2020-02-14
[Corollary to #4161] Redis Connection Check when Using Sentinel

### LINKED ISSUE #3985: collectstatic now requires a redis server running
### Environment
* Python version:  3.7.6
* NetBox version:  2.7.1

### Steps to Reproduce
1. Just run `python3 manage.py collectstatic --no-input`

### Expected Behavior
Previously (at least with 2.6.5), this commands generated all static files needed.

### Observed Behavior
With the latest version available in container (aka 2.7.1), the script tries to establish a connection with the redis server. As we try to use this command during a containter build, the given redis server is not accessible, making the build impossible.

Here is the stack error given with this command:

```
python3 manage.py collectstatic --no-input
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 539, in connect
    sock = self._connect()
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 596, in _connect
    raise err
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 584, in _connect
    sock.connect(socket_address)
OSError: [Errno 99] Address not available

  File "/opt/netbox/netbox/extras/apps.py", line 23, in ready
    rs.ping()
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 1106, in ping
    return self.execute_command('PING')
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 836, in execute_command
    conn = self.connection or pool.get_connection(command_name, **options)
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 1073, in get_connection
    connection.connect()
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 544, in connect
    raise ConnectionError(self._error_message(e))
redis.exceptions.ConnectionError: Error 99 connecting to localhost:6379. Address not available.

  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 357, in execute
    django.setup()
  File "/usr/local/lib/python3.7/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python3.7/site-packages/django/apps/registry.py", line 122, in populate
    app_config.ready()
  File "/opt/netbox/netbox/extras/apps.py", line 26, in ready
    "Unable to connect to the Redis database. Check that the Redis configuration has been defined in "
django.core.exceptions.ImproperlyConfigured: Unable to connect to the Redis database. Check that the Redis configuration has been defined in configuration.py.
```

dstarner:
### Fixes: #3985, #3984, and a bug introduced in #4161

Redis Sentinel support was adding to the settings and configuration, but the `extras` AppConfig was not updated to account for Sentinel when starting the application. Without this fix **it shouldn't cause the application to crash**, but to throw a false positive connection issue. This fixes the check to account for when sentinel is configured

### DIFF
--- a/netbox/extras/apps.py
+++ b/netbox/extras/apps.py
@@ -13,13 +13,23 @@ class ExtrasConfig(AppConfig):
 
         # Check that we can connect to the configured Redis database.
         try:
-            rs = redis.Redis(
-                host=settings.WEBHOOKS_REDIS_HOST,
-                port=settings.WEBHOOKS_REDIS_PORT,
-                db=settings.WEBHOOKS_REDIS_DATABASE,
-                password=settings.WEBHOOKS_REDIS_PASSWORD or None,
-                ssl=settings.WEBHOOKS_REDIS_SSL,
-            )
+            if settings.WEBHOOKS_REDIS_USING_SENTINEL:
+                sentinel = redis.sentinel.Sentinel(
+                    settings.WEBHOOKS_REDIS_SENTINELS,
+                    socket_timeout=settings.WEBHOOKS_REDIS_DEFAULT_TIMEOUT
+                )
+                rs = sentinel.master_for(
+                    settings.WEBHOOKS_REDIS_SENTINEL_SERVICE,
+                    socket_timeout=settings.WEBHOOKS_REDIS_DEFAULT_TIMEOUT
+                )
+            else:
+                rs = redis.Redis(
+                    host=settings.WEBHOOKS_REDIS_HOST,
+                    port=settings.WEBHOOKS_REDIS_PORT,
+                    db=settings.WEBHOOKS_REDIS_DATABASE,
+                    password=settings.WEBHOOKS_REDIS_PASSWORD or None,
+                    ssl=settings.WEBHOOKS_REDIS_SSL,
+                )
             rs.ping()
         except redis.exceptions.ConnectionError:
             raise ImproperlyConfigured(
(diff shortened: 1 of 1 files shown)


lampwins:
I would actually be okay with ripping this check out. It isn't necessarily adding any value over a sanity check on application upstart. For instance, it does nothing to check that the connection is still open over time. Cacheops is already set to safely degrade if the connection fails and error messages will be logged for failed webhook requests on account of the redis connection.

@jeremystretch thoughts? This would also resolve #3985

dstarner:
I feel like if a check is ever needed (which i believe it shouldnt be) then it also shouldn't be in a blocking path that prevents local-only commands from running, like mentioned in #3985. If it's believed that the check can be fully removed, I don't mind reworking this PR to just remove the original check.

dstarner:
Or, I guess if we wanted to keep the check, in a similar fashion to `IS_TESTING` in the configuration, we could add either a `REQUIRES_REDIS` or `LOCAL_ONLY` variable that wraps the check...but that seems even hackier and it has a slight code smell to it.

jeremystretch:
I have no objection to removing the check entirely. I agree that it is not necessary.

jeremystretch:
We probably do want to add some error handling around webhooks at least, though. Without this check in place, taking any action which results in the queuing of a webhook will cause a server error if the Redis service is unreachable. I'm happy to work on that; just wanted to make a note of it here.

jeremystretch (approved):

Issue #3451, opened 2019-08-26
WebHook should provide old data together with new data


robertpenz:
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.2

### Proposed Functionality

We've implemented a Webhook which gets informed for IP address and service changes, the problem now is that we only have the new values and not the old values. e.g. the dns_name field is changed from bla1.example.com to bla2.example.com we are only seeing the second name in the webhook call. This makes it impossible to delete the old DNS record and add the new one. We also took a look at the change log and even there only the new name is shown.

This makes is quit impossible to trigger stuff directly from the netbox which needs to know the old value to clean up. Please provide the old values along with the new ones in the webhook call.

### Use Case
User creates Device/VM in Netbox  and assigns IP address with DNS Name --> DNS Name gets written into the DNS Server. User changes the DNS Name, old DNS Name gets deleted in the DNS and the new one created. 

### Database Changes
No, the old value should be read from database before writing new and than send the webhook with both.

### External Dependencies
none

tb-killa:
We are also currently testing whether the webhooks will meet our future requirements.

Since I don't know if I should ask a new FR for this, I'm happy to give my topic under this FR otherwise.

For our tests we use the change for the IP addresses.
If I go to an address, but it doesn't make any changes but only clicks on Update, a webhook is generated directly. In fact, in my opinion, a webhook should only take place when a change is made.
Since an internal check between the cache and the current FORM value has to take place first, this requires a little more work.

candlerb:
I would very much like this too.

This could be done in a backwards-compatible way by adding a "data-before" member to the webhook message.

Another use case is for ignoring unimportant changes: you may wish to take action if field A changes, but not field B.

candlerb:
A little research into how it could be implemented: this [SO post](https://stackoverflow.com/questions/1355150/django-when-saving-how-can-you-check-if-a-field-has-changed) has some pointers.

* [FieldTracker](https://django-model-utils.readthedocs.io/en/latest/utilities.html#field-tracker) from django-model-utils is a plugin which can do this - but that adds another dependency
* Use the [pre-init](https://docs.djangoproject.com/en/2.2/ref/signals/#pre-init) signal to take a snapshot of fields at the point the model is instantiated
* Re-read the object from the database just before saving it (ugh)

I think it would also be useful to have previous state in the ObjectChange history - making it easier to revert an object to its previous state.

candlerb:
[...]

PR #4181, opened 2020-02-15
Fixes #2511: Compare object change to the previous change

### LINKED ISSUE #2511: Provide a diff showing the changes between two logged changes
### Environment
NetBox version:  2.4.6

### Proposed Functionality
As a netbox user I would like to have the ability to compare the changes in object data between logged changes. This feature would be available within the changelog context of an object in the UI and have the ability to select two or more log entries and compare them against each other.

### Use Case
This feature would be beneficial to users that want to quickly identify changes that have been made to an object rather than pulling the full object data and using it in an external tool to do a comparison.

As an example I created a screenshot that shows the differences between two versions of a device object. In this instance updating a serial number after it has been swapped out because it hypothetically had some sort of unrecoverable issue.

image

hSaria:
### Fixes: #2511

Add a difference section to the object change (i.e. `/extras/changelog/{pk}/`) with previous and next buttons.

image

> While going with a full-blown diff library might provide a nicer-looking diff, like that of GitHub, I've opted for a simpler approach to avoid the additional overhead of another dependency. And it seemed a bit overkill.

### DIFF
--- a/docs/release-notes/version-2.7.md
+++ b/docs/release-notes/version-2.7.md
@@ -2,6 +2,7 @@
 
 ## Enhancements
 
+* [#2511](https://github.com/netbox-community/netbox/issues/2511) - Compare object change to the previous change
 * [#3840](https://github.com/netbox-community/netbox/issues/3840) - Enhance search function when selecting VLANs for interface assignment
 * [#4170](https://github.com/netbox-community/netbox/issues/4170) - Improve color contrast in rack elevation drawings
 
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -12,6 +12,7 @@ from django_tables2 import RequestConfig
 
 from utilities.forms import ConfirmationForm
 from utilities.paginator import EnhancedPaginator
+from utilities.utils import shallow_compare_dict
 from utilities.views import BulkDeleteView, BulkEditView, ObjectDeleteView, ObjectEditView, ObjectListView
 from . import filters, forms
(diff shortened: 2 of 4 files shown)


tb-killa:
Doesn't we could use this type of implementation for fixing https://github.com/netbox-community/netbox/issues/3451 (closed but  allowed) too?

hSaria:
I've look into that, though I've got some concerns:

1. The additional database read call used to read the previous change (this would happen for every object change)
2. Inconsistency in the webhook data due to `settings.CHANGELOG_RETENTION`; the previous change might've been deleted if it was too old and thus the webhook data will not have a diff.
3. Even more inconsistency in the webhook data because not all models log their changes, so some webhooks will have a diff, whilst others won't.

At that point, I'm not even sure it's worth the effort from the perspective of the application that's listening to webhook because it would need to have the measures necessary to handle a webhook without a diff, anyway. And to be honest, a webhook listener should only ever perform idempotent actions, where running the same action multiple times will result in the same state (usually requires extra effort on the action's side), so a diff wouldn't be needed to begin with.

jeremystretch:
Nice!

neufeind:
If I update an object with the same like existing data could this here also make sure to only create a new "version" if at least something changed? Like if I update a server with it's primary ip-address - and if it didn't change to what was already in the database simply don't actually create a "save" for it?

jeremystretch:
@neufeind that would be outside the scope of the change proposed in the issue this PR addresses (#2511)

PR #4182, opened 2020-02-15
Fixes #4179: Set site field as required in rack group create form

### LINKED ISSUE #4179: IntegrityError when creating rack group or power panel without site
### Environment
* Python version:  3.6.9
* NetBox version:  2.7.6

### Steps to Reproduce
1. Go to rack group create form (`/dcim/rack-groups/add/`)
2. Enter name and slug, but leave site empty
3. Submit

### Expected Behavior
The form shouldn't allow me to submit without selecting a site.

### Observed Behavior
Exception
```

null value in column "site_id" violates not-null constraint
DETAIL:  Failing row contains (6, Rack Group 1, rg1, null, 2020-02-10, 2020-02-10 13:27:20.503952+00).
```

hSaria:
### Fixes: #4179 

During #3912, the site field in the rack group create form was accidentally set as not required.

### DIFF
--- a/docs/release-notes/version-2.7.md
+++ b/docs/release-notes/version-2.7.md
@@ -10,6 +10,7 @@
 * [#2519](https://github.com/netbox-community/netbox/issues/2519) - Avoid race condition when provisioning "next available" IPs/prefixes via the API
 * [#4168](https://github.com/netbox-community/netbox/issues/4168) - Role is not required when creating a virtual machine
 * [#4175](https://github.com/netbox-community/netbox/issues/4175) - Fix potential exception when bulk editing objects from a filtered list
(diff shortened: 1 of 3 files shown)


hSaria:
I went through #4136 and have also caught the following (included in the PR):
* `PowerPanelForm`: `site` field should be required
* `SecretFilterForm`: `role` field should not be required (it's a filter form)

hSaria:
Closed by https://github.com/netbox-community/netbox/commit/a456cbb26ce93507f810514ef7c60ae3971e8933.

jeremystretch:
Thanks for putting this in. I've used your PR as reference for a fix I just pushed for the first two issues. I wanted to open a separate bug (#4194) for the secret role problem though to ensure it's tracked separately.

PR #4162, opened 2020-02-13
Fixes: #4159 - Corrects settings.py to use CACHING_REDIS

DanSheps:
### Fixes: #4159 

Changes WEBHOOKS_REDIS to CACHING_REDIS under the caching settings in settings.py

PR #4189, opened 2020-02-18
4121 filter lookup expressions

### LINKED ISSUE #4121: Implement standard lookup expressions for all filter fields
### Environment
* Python version:  3.7
* NetBox version:  2.7.4

### Proposed Functionality
#3482 proposes implementing negation filters in the API. To realistically implement this, we need to do so in a dynamic way in a base filter class. In this issue, I propose a way to easily do this, but in doing so, to also support a number of other standard lookup expressions, beyond just negation. 

I propose the support of these lookup expressions on field types for which is makes logical sense:
- `n` - negated exact
- `isw` - case insensitive starts with
- `nisw` - negated case insensitive starts with
- `iew` - case insensitive ends with
- `niew` - negated insensitive ends with
- `ic` - insensitive contains
- `nic` - negated insensitive contains
- `lt` - less than
- ~`nlt` - negated less than~
- `lte` - less than or equal
- ~`nlte` - negated less than or equal~
- `gt` - greater than
- ~`ngt` - negated greater than~
- `gte` - greater than or equal
- ~`ngte` - negated greater than or equal~

It may seem like a lot, but they are cheap to add and maintain. Only certain fields will support certain lookups. For example, only numeric based fields will get less than/greater than/etc.

These lookup expressions are exposed as `<field_name>__<lookup_expr>`.

This is implemented by using a custom base filterset class which overrides the `get_filters()` method to extend fields with more lookup expressions.

### Use Case
Today most all filters offer only exact or iexact lookup expressions and only on querset `filter()` clauses, which also does not allow for negation. Many users often want to lookup based on partial matches of individual fields (instead of the aggregated nature of the `q` filter). Also pointed out in #3482 is a strong desire to support negation filters (queryset `exclude()` clauses).

As an example, in networking often a lot of logic is put into a device's hostname. It would be useful to do lookups based on the hostname string starting or ending with a certain string, without having to deal with the potential for unwanted results from the use of the `q` filter.

### Database Changes
None

### External Dependencies
This will require the use of a common base filterset class used across the board. This should not present any actual issues, but should still be noted.

lampwins:
### Fixes: #4121 

Adds support for dynamic filter lookup expressions set out in #4162 

This does not yet add support for custom fields or `isnull` lookups. It also needs docs to explain the available filters.


hSaria:
Should this be documented?

netbox/utilities/filters.py
```diff
@@ -111,6 +118,152 @@ def __init__(self, *args, **kwargs):
 # FilterSets
 #
 
+class BaseFilterSet(django_filters.FilterSet):
+    """
+    A base filterset which provides common functionaly to all NetBox filtersets
+    """
+    FILTER_DEFAULTS = deepcopy(django_filters.filterset.FILTER_FOR_DBFIELD_DEFAULTS)
```
jeremystretch: Why was this moved to inside the FilterSet?

jeremystretch (commented):
LGTM. Do you want to tackle custom field filtering as part of this, or break that into a separate issue? I'm fine with either.

netbox/utilities/filters.py
lampwins: Today we are simply doing this:
```
FILTER_DEFAULTS = django_filters.filterset.FILTER_FOR_DBFIELD_DEFAULTS
```
At the root of the module, which is actually just a weak reference to the original dictionary which is then also referenced in this same way in `django_filters.FilterSet`. So while it does work today because of the global nature in which we use these defaults, it is a form of monkey patch. Overriding it in the base class is the "more correct" way to do it and would allow us to use alternate functionality in some other case should the need ever arise. It also seems to be the unofficial recommendation of django-filters when looking at other solutions where people are using base classes like this.

lampwins:
@jeremystretch I just pushed some more tests and documentation.

After playing around with custom fields, it will require a separate implementation that has nothing to do with this one, so I will open a new issue for that.

netbox/utilities/filters.py
```diff
@@ -111,6 +117,152 @@ def __init__(self, *args, **kwargs):
...
+                            lookup_expr=lookup_expr,
+                            label=existing_filter.label,
+                            exclude=existing_filter.exclude,
+                            distinct=existing_filter.distinct,
+                            **existing_filter.extra
+                        )
+                    else:
+                        # The filter field is listed in Meta.fields so we can safely rely on default behaviour
+                        new_filter = cls.filter_for_field(field, field_name, lookup_expr)
+                except django_filters.exceptions.FieldLookupError:
```
jeremystretch: Wondering if we should move the `try`/`except` inside `if existing_filter_name in cls.declared_filters`. Seems like if `cls.filter_for_field()` fails on a default field it should make some noise.

jeremystretch (commented):
Looks good overall I think, just had some clarifying questions. Also, curious whether you have any insight to the performance impact.

netbox/utilities/filters.py
lampwins: `cls.filter_for_field()` will also raise `FieldLookupError` if a field explicit does not support a lookup expression we are trying to apply to it. Internally it is calling `resolve_field()`. I have added a comment to make this clear.

netbox/utilities/filters.py
```diff
@@ -111,6 +117,152 @@ def __init__(self, *args, **kwargs):
...
+
+            # Get properties of the existing filter for later use
+            field_name = existing_filter.field_name
+            field = get_model_field(cls._meta.model, field_name)
+
+            # Create new filters for each lookup expression in the map
+            for lookup_name, lookup_expr in lookup_map.items():
+                new_filter_name = '{}__{}'.format(existing_filter_name, lookup_name)
+                if existing_filter.lookup_expr == 'in':
+                    lookup_expr = 'in'  # 'in' lookups must remain to avoid unwanted slicing on certain querysets
```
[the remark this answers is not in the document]
lampwins: I suppose that is a possibility, however, the only filters we use `in` on are `region` are `id__in` (which we should deprecate anyway). This exists to create negated region filters.

Open to suggestions on how to better refactor it though.

netbox/utilities/filters.py
```diff
@@ -111,6 +117,152 @@ def __init__(self, *args, **kwargs):
...
+                # These filter types support only negation
+                lookup_map = FILTER_NEGATION_LOOKUP_MAP
+
+            elif isinstance(existing_filter, (
+                django_filters.filters.CharFilter,
+                django_filters.MultipleChoiceFilter,
+                MultiValueCharFilter,
+                MultiValueMACAddressFilter
+            )):
+                lookup_map = FILTER_CHAR_BASED_LOOKUP_MAP
```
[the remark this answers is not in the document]
lampwins: Done

netbox/utilities/filters.py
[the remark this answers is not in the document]
jeremystretch: How about adding a separate lookup map for MultipleChoiceFilters?

```python
FILTER_MULTICHOICE_LOOKUP_MAP = dict(
    n='in'
)
```

netbox/utilities/tests/test_filters.py
```diff
@@ -60,3 +65,196 @@ def test_filter_combined(self):
         self.assertEqual(qs.count(), 2)
         self.assertEqual(qs[0], self.site1)
         self.assertEqual(qs[1], self.site3)
+
+
+class DynamicFilterLookupExpressionTest(TestCase):
```
jeremystretch: IMO we should also include a generic test for BaseFilterSet itself. Just an instantiation with one each of every filter type, with tests to ensure all the expected filters are present. I'm going to take a shot at this myself (unless there are objections) to help get more familiar with the logic.

netbox/utilities/filters.py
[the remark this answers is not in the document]
lampwins: This is only a problem with our region filters because we do this to explicitly set the lookup expression to `in`:
```
region_id = TreeNodeMultipleChoiceFilter(
    queryset=Region.objects.all(),
    field_name='region',
    lookup_expr='in',
    label='Region (ID)',
)
region = TreeNodeMultipleChoiceFilter(
    queryset=Region.objects.all(),
    field_name='region',
    lookup_expr='in',
    to_field_name='slug',
    label='Region (slug)',
)
```

I have fixed this by applying your solution to `TreeNodeMultipleChoiceFilter` and move it before the lookup map matching for `MultipleChoiceFilters` which was somewhat hiding this because of subclassing.

netbox/utilities/tests/test_filters.py
lampwins: Yeah sounds good. I will hold off doing it unless you ask me.

lampwins:
> Also, curious whether you have any insight to the performance impact.

`django_filters.FilterSet.get_filters()` is called from `__new__` so this logic actually only runs once per filterset class when that class is imported into the runtime for the first time. Once the filter fields are created, they are no different from any other filter field, so there is really no discernible impact to speak of here.

netbox/utilities/tests/test_filters.py
```diff
@@ -60,3 +72,447 @@ def test_filter_combined(self):
...
+    def test_tag_filter(self):
+        self.assertIsInstance(self.filters['tagfield'], TagFilter)
+        self.assertEqual(self.filters['tagfield'].lookup_expr, 'exact')
+        self.assertEqual(self.filters['tagfield'].exclude, False)
+        self.assertEqual(self.filters['tagfield__n'].lookup_expr, 'exact')
+        self.assertEqual(self.filters['tagfield__n'].exclude, True)
+
+    def test_tree_node_multiple_choice_filter(self):
+        self.assertIsInstance(self.filters['treeforeignkeyfield'], TreeNodeMultipleChoiceFilter)
+        # TODO: lookup_expr different for negation?
```
jeremystretch: @lampwins Can you sanity-check this please?

PR #4190, opened 2020-02-18
Fixes #1830: Support assgin an existing device to a rack

### LINKED ISSUE #1830: Assign existing device to rack from rack layout window
[x] Feature request
[ ] Bug report
[ ] Documentation
Environment

    Python version: 3.5.2
    NetBox version: 2.2.8

Description

Would love to be able to assign existing devices to a rack slot when you click Add Device from the rack layout Window.

I see it being similar to how you can search for an IP address when assigning an IP to a management interface.

kobayashi:
### Fixes: #1830
Add "Assign Device" tab in the device create page if  clicking "add device" button in rack view.
Screenshot from 2020-02-18 02-39-26

### DIFF
--- a/docs/release-notes/version-2.7.md
+++ b/docs/release-notes/version-2.7.md
@@ -2,6 +2,7 @@
 
 ## Enhancements
 
+* [#1830](https://github.com/netbox-community/netbox/issues/1830) - Support Assgin existing device to a rack with 'add device' button
 * [#3840](https://github.com/netbox-community/netbox/issues/3840) - Enhance search function when selecting VLANs for interface assignment
 * [#4170](https://github.com/netbox-community/netbox/issues/4170) - Improve color contrast in rack elevation drawings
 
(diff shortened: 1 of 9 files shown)


jeremystretch:
IMO this introduces way too many new components to maintain, especially given the minimal functionality it provides. We have too many forms and tables and templates as it is; we should be working toward consolidating what's there already instead of adding more.

And ultimately, what are we gaining by clicking the rack unit first and selecting a device, versus editing a device and selecting the rack unit? (Bear in mind that in both instances, the device has already been assigned to the current rack.) This would just introduce a redundant workflow to accomplish the same thing.

I appreciate your work here @kobayashi, but I think we should reconsider whether implementing #1830 is worthwhile.

kobayashi:
Thanks to your comment @jeremystretch ,

>IMO this introduces way too many new components to maintain, especially given the minimal functionality it provides. We have too many forms and tables and templates as it is; we should be working toward consolidating what's there already instead of adding more.

That make sense. It is better to clean the code before adding more like this PR.

> This would just introduce a redundant workflow to accomplish the same thing.

Yes, that's exactly it. I just picked it up from accepted issues to close.
I can close this PR and will change the label of the issue from accepted to `gathering feedback`.

jeremystretch:
Thanks. Sorry for the runaround. I think I had assumed earlier that this would require much less work to implement.

PR #4233, opened 2020-02-21
Fixes #4228: Fix display of device images in rack elevations

### LINKED ISSUE #4228: Images in rack view not formatted, right hand side is cropped
### Environment
* Python version: 3.6.8 
* NetBox version: 2.7.7 

### Steps to Reproduce
1. Upload an image formatted to the correct aspect ratio (226x19 was suggested) to a front or rear image of a device type. Save.
2. Go to a rack view containing device which has had an image uploaded.

### Expected Behavior
Image is shown correctly, scaled so that the entirety of it fits the width (and appropriate number of RU height).

### Observed Behavior
No matter what I try, the right hand side of the image is always cut off. This happens regardless of the image size / aspect ratio.

Searching into this a bit more, I think the issue is related to the CSS code which enforces a width of 260px when the rectangle box representing the rack is 229px wide.

`<image class="device-image" height="20" preserveAspectRatio="none" width="260" x="30" xlink:href="/media/deicetype-images/<filename>" t="0"></image>`

Upon further experimenting, if you set the preserveAspectRatio to "yes", then the image's aspect ratio is respected and squeezed to fit inside the width of 230 / 30x number of RU of device *and* it is also centered. This might be a good way of working around issue of certain images having the rack ears in them (19") or only the "opening" (circa 17.5") width.

jeremystretch:
### Fixes: #4228

- Correct the width of the SVG device element
- Change the default width and height of a rack unit from `(230, 20)` to `(220, 22)` to achieve a proper aspect ratio


tfboy:
If you wanted a tiny bit more width than the 10:1 aspect ratio (for example to include rack ears!) To help with longer device names, I'm sure that would be fine. It's easy to pad out non-eared 10:1 equipment images with white or other to fit the designated recommended image aspect ratio.

PR #4242, opened 2020-02-22
InterfaceIPAddressTable to use non-edit link

### LINKED ISSUE #4241: /dcim/interfaces/xxxx/ IP address table uses edit URL instead of view URL
### Environment
* Python version:  3.6.9
* NetBox version:  2.7.7

### Steps to Reproduce
1. Visit any interface URL which has an IP address assigned (IPv4 or IPv6, doesn't matter)
    - Example: https://netboxdemo.com/dcim/interfaces/531/
2. Click on an IP address in the IP Address table at the bottom of the page
3. See that the hyperlink opens up the edit page for that IP address (instead of the view page)

### Expected Behavior
It is expected that the hyperlink would point towards the view page of the IP address (`/ipam/ip-addresses/xxxx/`). The VLANs table (in the same page) does point towards the view URL which is expected.

### Observed Behavior
Instead, the link points towards the edit URL of the IP address (`/ipam/ip-addresses/xxx/edit/?interface=yyy&return_url=/dcim/interfaces/yyy/`)

PackeTsar:
InterfaceIPAddressTable was using IPADDRESS_ASSIGN_LINK (edit URL with return_url), change to using view-only URL with IPADDRESS_LINK

### Fixes: #4241
Adjusted InterfaceIPAddressTable link object to use the view link to view the IP address page, instead of the edit link to edit the IP address

### DIFF
--- a/netbox/ipam/tables.py
+++ b/netbox/ipam/tables.py
@@ -385,7 +385,7 @@ class InterfaceIPAddressTable(BaseTable):
     """
     List IP addresses assigned to a specific Interface.
     """
-    address = tables.TemplateColumn(IPADDRESS_ASSIGN_LINK, verbose_name='IP Address')
+    address = tables.TemplateColumn(IPADDRESS_LINK, verbose_name='IP Address')
     vrf = tables.TemplateColumn(VRF_LINK, verbose_name='VRF')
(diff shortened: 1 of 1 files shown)


jeremystretch:
Sorry, just saw this. Ended up just using a regular `LinkColumn()`. Thanks though!

PR #4253, opened 2020-02-24
Fixes #3612: Update screenshots

### LINKED ISSUE #3612: Update old screenshots in README.md
### Change Type
[ ] Addition
[X] Correction
[ ] Deprecation

### Proposed Changes
The three screenshots in the README.md file is getting stale/old. The date on them point to 2016 and much have changed since then and they needs to be updated as they do not reflect how exactly how netbox looks and feels currently.

hSaria:
### Fixes: #3612

Updated the screenshots with ones from version v2.7.7. The content of the existing screenshots were fine (main page for summary, rack view for DCIM, prefix list for IPAM), so I didn't add anything else to avoid bloating the main page.


jeremystretch:
> 3176 × 1968 pixels

Probably don't need them quite this large. I'm guessing you're using a very high DPI display?

hSaria:
Yeah, only _slightly_. GitHub sizes them according to the width of the page (why I didn't notice any difference), but I can downscale them if you'd like.

jeremystretch:
@hSaria Yeah, like 50% of that would be good I think.

hSaria:
Done

Issue #3474, opened 2019-09-03
Napalm / device status problem when facts not returned correctly


fknorn:
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.2

### Steps to Reproduce
1. Set up napalm as per docs
2. Have a device that for some reason doesn't correctly return the device facts (an older Juniper SRX in my case)
3. Click device's status tab
4. Site never finishes loading ...

### Expected Behavior
Get more graceful handling of the case where no facts are returned. 

### Observed Behavior
No error message, the page just never finishes loading (the spinner remains forever)

### Diagnosis
What occurs under the hood (and can currently only be seen in the console) is a javascript error:

`Uncaught TypeError: Cannot read property 'used_ram' of undefined`

The problematic line is

` $('#memory').after("<tr><td>Used</td><td>" + json['get_environment']['memory']['used_ram'] + "</td></tr>");`

This is what the API actually returns in my case:

```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "get_facts": {
        "error": "Method get_facts failed: RpcError(severity: error, bad_element: None, message: error: usp_ipc_client_open: failed to connect to the server after 1 retries(65)\nerror: usp_ipc_client_open: failed to connect to the server after 1 retries(65)\nerror: 
[... line repeated several times ..]
usp_ipc_client_open: failed to connect to the server after 1 retries(65))"
    },
    "get_environment": {}
}
```

So the code a little higher up that performs some sanity checks on the returned data

```
if (!json['get_environment']['error']) {
```
near that place should probably be extended with something along the lines of

```
if (json['get_environment'].length || !json['get_environment']['error']) {
```

Thanks!

PR #4258, opened 2020-02-24
Fixes: #4255 - Modify script ObjectVars to utilize DynamicModelChoiceField

### LINKED ISSUE #4255: ObjectVar and MultiObjectVar should utilize DynamicModelChoiceField
### Proposed Changes

ObjectVar currently uses forms.ModelChoiceField, this should be changed to utilities.forms.DynamicModelChoiceField
MultiObjectVar currently uses forms.ModelMultipleChoiceField.  this should be changed to use utilities.forms.DynamicModelMultipleChoiceField.

```
from utilities.forms import DynamicModelChoiceField, DynamicModelMultipleChoiceField

class ObjectVar(ObjectVar):
    form_field = DynamicModelChoiceField
    
    
class MultiObjectVar(MultiObjectVar):
    form_field = DynamicModelMultipleChoiceField
```

### Justification

The two above dynamic model fields limits the pre-population of the select fields to improve performance when using the Select2 widget on scripts.

DanSheps:
### Fixes: #4255 

Adds two new fields that extend their related fields:

* DynamicObjectVar
* DynamicMultiObjectVar

Adds tests for those two new fields

### DIFF
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -170,8 +170,8 @@ class ObjectVar(ScriptVariable):
     """
     form_field = DynamicModelChoiceField
 
-    def __init__(self, queryset, widget, *args, **kwargs):
-        super().__init__(widget=widget, *args, **kwargs)
+    def __init__(self, queryset, *args, **kwargs):
+        super().__init__(*args, **kwargs)
 
         # Queryset for field choices
         self.field_attrs['queryset'] = queryset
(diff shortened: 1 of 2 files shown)


jeremystretch:
@DanSheps please see my comment on #4255

DanSheps:
@jeremystretch I pushed a new commit, does that look cleaner?

jeremystretch:
@DanSheps With #3474 in place, I think all we need to do is tweak the form field.

PR #4263, opened 2020-02-24
Closes #4237: Enable custom templating for webhook request content

### LINKED ISSUE #4237: Add support for customizable webhook payloads
### Environment
* Python version: 3.6.9
* NetBox version: 2.7.7

### Proposed Functionality
Extend the webhook model to support custom payload formats and HTTP methods. The proposal is to add two new fields to the model:

* `http_method` - A static choice of valid HTTP methods `GET`, `POST`, `DELETE`, etc. (I'm curious to know if there's an argument for allowing arbitrary custom methods here.)
* `template` - Jinja2 template code for forming the request body. All of the existing payload data will be passed as context available within the template. If the template is not defined for a webhook, it will use the current format.

No changes to the manner in which webhooks are currently enqueued should be needed, so there should not be any performance penalty noticeable to the user. Upon extraction of a queued webhook action, the worker process will render the request body using the provided template code stored in the database.

### Use Case
Currently, a webhook can only make an HTTP POST request with a [particular data structure](https://netbox.readthedocs.io/en/stable/additional-features/webhooks/#requests). While useful, this typically requires some type of middleware to translate the data from NetBox into something actionable by the receiver. (For example, to inject a Slack message in response to the creation of a new site.) Providing the option of crafting an arbitrary request body is likely to mitigate the need for interim middleware for many (but certainly not all) use cases.

### Database Changes
Introduce the two fields described above to the `extras.Webhook` model.

### External Dependencies
None

jeremystretch:
# Fixes: #4237

- Change the `http_content_type` field from a choice field to a free-form string. This allows the user to specify an arbitrary content type.
- Add the `body_template` field for defining a Jinja2 template.
- Enable logging for the webhooks worker

Screenshot_2020-02-24 Change webhook NetBox

The above configuration will generate a request similar to what follows:

```
127.0.0.1 - - [24/Feb/2020 15:55:24] "POST / HTTP/1.1" 200 -
------------
Host: localhost:9000
Accept-Encoding: identity
Content-Length: 34

Someone created a site named Foo!
```


netbox/extras/webhooks_worker.py
```diff
@@ -33,10 +41,22 @@ def process_webhook(webhook, data, model_name, event, timestamp, username, reque
...
+
+    # Construct the request body. If a template has been defined, use it. Otherwise, dump the context as either JSON
+    # or form data.
+    if webhook.body_template:
+        try:
+            params['data'] = render_jinja2(webhook.body_template, context)
+        except TemplateError as e:
+            logger.error("Error rendering request body: {}".format(e))
+            return
+    elif webhook.http_content_type == HTTP_CONTENT_TYPE_JSON:
```
lampwins: I wonder if we should add an additional sanity check here to catch invalid JSON from a templated body. Basically check if the content type is `HTTP_CONTENT_TYPE_JSON` and a jinja2 rendered body, attempt a json.loads() and log a helpful message on failure. I imagine this would be very helpful to users who are writing JSON templates (because they absolutely will).

jeremystretch: An opportunistic validation of the outgoing JSON? We can, I guess. But are you proposing that an outgoing request with invalid JSON data gets aborted or just logged as a warning?

PR #4299, opened 2020-02-28
Closes #2328: External user authentication

### LINKED ISSUE #2328: External authentication
### Environment
* Python version:  3.5.4
* NetBox version:  Example: v2.3.2

Ability to use external authentication via a proxy (eg. Keycloak-proxy, Nginx, Apache) by setting the appropriate headers. It was already described as a possibility [here](https://github.com/digitalocean/netbox/issues/889) but I find no documentation or anything about this topic.

### Proposed Functionality
It should be possible to pass authentication headers to netbox and either add a user automatically to the local database and make it possible to give the correct permissions in netbox or use the permissions passed via a header.

### Use Case
This would be very useful because it wouldn't be necessary to implement and maintain dozens of different authentication mechanisms directly to Netbox but just pass the headers.
 
There is already an example for this in the wild but it requires some code changes and patching Netbox on every update is in my opinion not a good idea. So it would be really cool if something like this could be implemented in the codebase.

Example:
https://groups.google.com/d/msg/netbox-discuss/BTB8q8CzmrA/2BcnbectAQAJ

jeremystretch:
### Fixes: #2328

Introduces a set of new configuration parameters that can be used to support external user authentication:

* REMOTE_AUTH_ENABLED = False
* REMOTE_AUTH_BACKEND = 'utilities.auth_backends.RemoteUserBackend'
* REMOTE_AUTH_HEADER = 'HTTP_REMOTE_USER'
* REMOTE_AUTH_AUTO_CREATE_USER = True
* REMOTE_AUTH_DEFAULT_GROUPS = []
* REMOTE_AUTH_DEFAULT_PERMISSIONS = []


sdktr:
Great work! If understood correctly the grouplist (the authorization part basically) is alsays fetched from the netboxDB. The default grouplist is used only for newly created accounts right?

If this is the case, should we provide another http header to pass down a grouplist as well? Or should that be a typical use case for a custom auth middleware?

jeremystretch:
> If understood correctly the grouplist (the authorization part basically) is always fetched from the netboxDB

Correct; the backend is just indicating to NetBox what user is logged in (and creating a new user if necessary). Permissions are still calculated based on the local user instance.

> The default grouplist is used only for newly created accounts right?

Correct; groups/permission assignment only happens in the event a new user is created.

> If this is the case, should we provide another http header to pass down a grouplist as well? Or should that be a typical use case for a custom auth middleware?

I'd argue that should be handled by a custom backend (not technically middleware), because people are going to have different needs and expectations as to its specific behavior.

sdktr:
While an http based grouplist would come in handy, I understand it’s one of these things that tend to grow out of hand quickly when people start tweaking it to their environment.

LGTM, needs testing though🙂

jeremystretch:
> needs testing though

Yeah, for sure. Docs & testing to come, just hoping to get the rough model nailed down initially.

ThomasADavis:
How does this affect API tokens?  Does this mean you would have to put a separate path for the API tokens vs. the rest of netbox url's?

jeremystretch:
@ThomasADavis This has no effect on API authentication at all. Tokens use a separate authentication backend.

jeremystretch:
Still need to figure out how to modify the remote auth configuration for tests.

davama:
@jeremystretch 

>Still need to figure out how to modify the remote auth configuration for tests.

Im assuming that's why this PR has not made it to a release, yes?

I've honestly been waiting for this feature for some time now. Happy to see it now :+1: 

Thank you,
Dave

PR #4304, opened 2020-03-02
Set local_context_data serializer on Device and VM to method

### LINKED ISSUE #4288: Incorrect Type for Device & VM `local_context_data`
### Environment
* Python version:  3.7
* NetBox version:  2.7.8

### Steps to Reproduce
1. Add config context to a device
1. Using any generated Swagger client - we used go-netbox - try to hit one of the list routes `/devices` or `/virtual-machines`
1. Watch the strongly-typed swagger client explode as it tries to convert a dict/map into a string

### Expected Behavior

`local_context_data` should be defined in the Swagger documentation as either `null` or as an object/dict/map so that it can be handled correctly by strongly typed swagger clients.

### Observed Behavior

Screen Shot 2020-02-26 at 2 12 26 PM

`local_context_data` is type-unaware and its converted into a string, meaning that Swagger clients will try to parse it out as a string, which will cause type errors on strongly typed systems

dstarner:
### Fixes: #4288 Incorrect Type for Device & VM `local_context_data`

For some reason, `local_context_data` fields were being parsed by Swagger as strings, which broke any swagger clients that attempted to operate against devices or VMs containing this data. This fixes that issue by explicitly denoting `local_context_data` to be of type `DictField` in the serializer

### DIFF
--- a/netbox/dcim/api/serializers.py
+++ b/netbox/dcim/api/serializers.py
@@ -376,7 +376,6 @@ class DeviceSerializer(TaggitSerializer, CustomFieldModelSerializer):
     cluster = NestedClusterSerializer(required=False, allow_null=True)
     virtual_chassis = NestedVirtualChassisSerializer(required=False, allow_null=True)
     tags = TagListSerializerField(required=False)
-    local_context_data = serializers.SerializerMethodField()
 
     class Meta:
         model = Device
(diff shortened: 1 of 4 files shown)


jeremystretch:
Rather than replicating a custom method for every instance of a JSON field, I suggest we introduce a [custom inspector](https://github.com/netbox-community/netbox/blob/develop/netbox/utilities/custom_inspectors.py) for the serializer field. This should avoid the need to extend the serializers.

dstarner:
hm, alright, thanks for the suggestion. I will give it a look and implement one

jeremystretch:
I haven't played with it, but it should look something like:

```python
class JSONFieldInspector(FieldInspector):
    def process_result(self, result, method_name, obj, **kwargs):

        if isinstance(result, openapi.Schema) and isinstance(obj, JSONField):
            result.type = 'dict'

        return result
```

And then the class just needs to be added to `netbox.settings.SWAGGER_SETTINGS['DEFAULT_FIELD_INSPECTORS']`.

dstarner:
Thanks! Yea, I haven't looked too deep into the Inspector code for drf-yasg, but that should do it. I'll double check

PR #4327, opened 2020-03-07
4078: Standardize 'comments' and 'description' Fields

### LINKED ISSUE #4078: Standardize the use of name, description, and comments fields
### Proposed Changes
We need to do an audit of all existing `description` and `comments` fields and adjust where necessary to ensure consistency across all models. I'm not sure this has ever been formally defined, but in my mind the distinction is:

* `description`: A one-liner of limited length describing the object, suitable for inclusion when viewing a list of objects
* `comments`: Long-form arbitrary remarks regarding an object, complete with Markdown support

### Justification
This ensures consistency in the way the fields are handled across all models.

dstarner:
### Fixes: #4078 

**NOTE**: This is merging into the `develop-2.8` branch, because this is a 2.8 milestone issue.

Implements a `DescriptionModel` and `CommentModel` that standardizes the `comments` and `description` fields for any/all models. They are broken out into model extensions / mixins to make it easier to standardize across the application(s) for current *and* future models.

The migration files were added as `<id>_standarize_description_comment_fields.py` which I felt was an appropriate name for them.

---
I initially called them `...Mixins` but when being appended to some of the more basic models, this caused a [method resolution order exception](https://stackoverflow.com/questions/3003053/metaclass-multiple-inheritance-inconsistency) so I called them `Model`s because `MyModel(models.Model, CommentModel)` had to become `MyModel(CommentModel)`.

### DIFF
--- a/docs/release-notes/version-2.8.md
+++ b/docs/release-notes/version-2.8.md
@@ -18,6 +18,8 @@ If further customization of remote authentication is desired (for instance, if y
 
 ### Enhancements
 
+* [#1754](https://github.com/netbox-community/netbox/issues/1754) - Added support for nested rack groups
+* [#3939](https://github.com/netbox-community/netbox/issues/3939) - Added support for nested tenant groups
 * [#4195](https://github.com/netbox-community/netbox/issues/4195) - Enabled application logging (see [logging configuration](../configuration/optional-settings.md#logging))
 
 ### API Changes


dstarner:
I can add a comment to the changelog before this gets reviewed / lands. I left that out

dstarner:
I also need to fix this branch to go into `develop-2.8`

netbox/utilities/models.py
```diff
@@ -33,9 +32,37 @@ def to_objectchange(self, action):
...
+        from extras.models import ObjectChange
         return ObjectChange(
             changed_object=self,
             object_repr=str(self),
             action=action,
             object_data=serialize_object(self)
         )
+
+
+class CommentModel(models.Model):
```
jeremystretch: Using abstract models to add individual fields is overkill. At most we would implement a custom field class, and even that seems excessive. All we need to do is standardize and document the field attributes.

jeremystretch (changes requested):

netbox/utilities/models.py
DanSheps: Perhaps just go with a simple constant for length, that looks to be all that we need to standardize on.

dstarner:
Fair, overengineered, will fix

jeremystretch:
@dstarner Hold off on this for now, please. There may be some additional work to do: see my comment on #4078.

PR #4379, opened 2020-03-18
Refactor registry to behave as a dictionary

jeremystretch:
- Tweaks the new Registry object to behave as a sort of write-once dictionary
- Adds basic tests

### DIFF
--- a/netbox/extras/registry.py
+++ b/netbox/extras/registry.py
@@ -1,0 +1,21 @@
+class Registry(dict):
+    """
+    Central registry for registration of functionality. Once a store (key) is defined, it cannot be overwritten or
+    deleted (although its value may be manipulated).
+    """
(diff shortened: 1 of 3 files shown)


netbox/extras/registry.py
```diff
@@ -0,0 +1,21 @@
...
+        except KeyError:
+            raise Exception("Invalid store: {}".format(key))
+
+    def __setitem__(self, key, value):
+        if key in self:
+            raise Exception("Store already set: {}".format(key))
+        super().__setitem__(key, value)
+
+    def __delitem__(self, key):
+        raise Exception("Cannot delete stores from registry")
```
lampwins: Same thing here, would raise KeyError.

lampwins (commented):
LGTM, just the two notes about exceptions.

Issue #3351, opened 2019-07-17
Introduce support for NetBox plugins


jeremystretch:
### Environment

* Python version:  3.6
* NetBox version:  2.6.1

### Proposed Functionality

This feature request proposes modifying NetBox to support the addition of arbitrary third party plugin applications. The plugins can be installed and configured by end users to extend core NetBox functionality to better suit particular use cases. Plugins will allow users to introduce custom models, API endpoints, and additional business logic.

The plugin architecture is envisioned as follows:

* Plugin source code is pulled into a new `/plugins/` directory within the NetBox application root.
* Individual plugins are enabled by adding them to the new `PLUGINS` parameter in `configuration.py`.
* Each plugin will package a default configuration, elements of which can be overridden in `configuration.py`.
* Plugins can introduce new UI and API URLs as `/plugins/<plugin>/` and `/api/plugins/<plugin/`, respectively.
* Each plugin is responsible for its own database schema migrations (if any). Plugins will not be permitted to modify the schema of any models outside the plugin.
* Plugins will be able to register arbitrary links in NetBox's navigation menu.
* Plugins will be able to act on events within NetBox by receiving [Django signals](https://docs.djangoproject.com/en/2.2/topics/signals/).
* Plugins may optionally specify a minimum and/or maximum version of NetBox which which they are compatible.

Obviously, there's a lot to figure out here. @lampwins has already done some work in this area, but I wanted to capture it in a feature request to kick off the community discussion.

### Use Case

This feature would introduce several benefits:

* Remove the need for users to fork the core NetBox code base in order to extend its functionality
* Provide a standardized, documented path for users to implement customizations
* Relieve demand on the core developer team for new features
* Potentially useful as a "proving ground" for new features: Popular plugins may be adopted into the core

However, there are also certain drawbacks which should be considered:

* The additional care necessary to accommodate plugins may slow general feature development
* General support may be more difficult: Plugins will need to be disabled for troubleshooting core issues
* End users will need to thoroughly vet the plugins they choose to install

### Database Changes

This probably won't require any substantial database schema changes, though more exploration is needed to make a clear determination.

### External Dependencies

Probably none

lampwins:
I am happy to see us looking down this path :)

In the name of transparency, here is the PoC I did for a plugin's architecture to which Jeremy alluded. It follows many of the points made above.
https://github.com/lampwins/netbox/tree/plugins

A couple of thoughts:
[...]

PR #4385, opened 2020-03-18
Implements 3351 plugins

lampwins:
### Fixes: #3351

This is the first pass at the core plugins implementation. This is lacking docs and the maintainers have agreed we will implement those as a separate PR.


netbox/netbox/urls.py
```diff
@@ -68,6 +71,41 @@
 
 ]
 
+# Plugins
+plugin_patterns = []
+plugin_api_patterns = []
+for app in apps.get_app_configs():
```
jeremystretch: Should we move this logic out of the root URLs file? Maybe into `extras/plugins/urls.py`?

netbox/netbox/urls.py
```diff
@@ -68,6 +71,41 @@
...
+        if importlib.util.find_spec('{}.urls'.format(app.name)):
+            urls = importlib.import_module('{}.urls'.format(app.name))
+            url_slug = getattr(app.NetBoxPluginMeta, 'url_slug', app.label)
+            if hasattr(urls, 'urlpatterns'):
+                # Mount URLs at `<url_slug>/<path>`
+                plugin_patterns.append(
+                    path('{}/'.format(url_slug), include((urls.urlpatterns, app.label)))
+                )
+        # Check if the plugin specifies any API URLs
+        if importlib.util.find_spec('{}.api'.format(app.name)):
```
jeremystretch: Is the first call to `find_spec()` needed?

jeremystretch (approved):

netbox/netbox/urls.py
lampwins: Yeah I like it, will do.

docs/plugins/development.md
```diff
@@ -106,6 +106,7 @@ class AnimalSoundsConfig(PluginConfig):
 * `max_version`: Maximum version of NetBox with which the plugin is compatible
 * `middleware`: A list of middleware classes to append after NetBox's build-in middleware.
 * `caching_config`: Plugin-specific cache configuration
+* `template_content`: The dotted path to the list of template content classes (default: `template_content.template_contnet`)
```
lampwins: `template_contnet` > `template_content`

docs/plugins/development.md
```diff
@@ -305,3 +306,40 @@ A `PluginNavMenuButton` has the following attributes:
...
+
+* `left_page()` - Inject content on the left side of the page
+* `right_page()` - Inject content on the right side of the page
+* `full_width_page()` - Inject content across the entire bottom of the page
+* `buttons()` - Add buttons to the top of the page
+
+Each of these methods must return HTML content suitable for inclusion in the object template. Two instance attributes are available for context:
+
+* `self.obj` - The object being viewed
+* `self.context` - The current template context
```
lampwins: While this is true, it might be worth mentioning this "original" context is available in the `render()` method as `obj_context`

docs/plugins/development.md
```diff
@@ -305,3 +306,40 @@ A `PluginNavMenuButton` has the following attributes:
...
+        return self.render('netbox_animal_sounds/site.html')
+
+class AddRackAnimal(PluginTemplateContent):
+    model = 'dcim.rack'
+
+    def left_page(self):
+        extra_data = {'foo': 123}
+        return self.render('netbox_animal_sounds/rack.html', extra_data)
+
+template_content_classes = [AddSiteAnimal, AddRackAnimal]
```
lampwins: Looks like you have renamed this `template_content`

netbox/extras/plugins/__init__.py
```diff
@@ -48,10 +48,18 @@ class PluginConfig(AppConfig):
 
     # Default integration paths. Plugin authors can override these to customize the paths to
     # integrated components.
+    template_content_classes = 'template_content.template_content_classes'
```
lampwins: Seems to be some confusion over `template_content` vs `template_content_classes`? I don't have a strong opinion either way, but `template_content` is shorter.

netbox/extras/plugins/signals.py
```diff
@@ -3,7 +3,9 @@
 
 
 class PluginSignal(django.dispatch.Signal):
```
lampwins: If we are not using this anymore do you think it makes sense to leave it?

docs/plugins/development.md
jeremystretch: Why are we passing the original context as a single variable, as opposed to just passing it directly (after adding the `obj` variable)?

netbox/extras/plugins/signals.py
jeremystretch: I felt bad wiping it out. We'll probably need it or some variation of it, but yeah we should probably remove it given that it's not currently in use.

docs/plugins/development.md
jeremystretch: Looking at this further, we should restrict the template context to a prescribed set of variables, since we make no guarantees that any of the context data present in the current release will always be available. I think it makes sense to limit the context available to a plugin to:

* The object being viewed
* The current request
* The plugin's configuration

netbox/extras/templatetags/plugins.py
```diff
@@ -0,0 +1,71 @@
...
+        # Update context with plugin-specific configuration parameters
+        plugin_name = template_extension.__module__.split('.')[0]
+        context['config'] = settings.PLUGINS_CONFIG.get(plugin_name)
+
+        instance = template_extension(context)
+        try:
+            content = getattr(instance, method)()
+        except NotImplementedError:
+            # This content renderer class does not define content for this method
+            continue
```
jeremystretch: Can we optimize this? Currently we're populating context data, instantiating a class, and calling a method just to determine if the method is actually defined. This approach does allow for a tidy plugin development experience, but it also imposes a lot of unnecessary overhead when rendering templates.

I'm not sure what a better approach would look like. I think ideally we'd call `hasattr()` on the class first to determine whether any further action is needed.

lampwins: Personally I think it is a premature optimization. The time complexity of this is the same either way: O(N) over the list of registered classes. We could create a larger dict in registry per class method, but is it really worth it?

jeremystretch: Under the current approach, each applicable PluginTemplateExtension is instantiated four times per view (one for each available method). It's almost certain we'll add more methods at some point in the future, so we might as well get ahead of it now.

jeremystretch: Came up with a simple shortcut: if the sublcass' method is the same as the parent class', skip it.

```python
if getattr(template_extension, method) == getattr(PluginTemplateExtension, method):
    continue
```

DanSheps (approved):
LGTM

PR #4397, opened 2020-03-23
Add documentation for the plugins framework

jeremystretch:
Tacks on documentation for #3351


docs/plugins/development.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Plugin Structure
+
+Although the specific structure of a plugin is largely left to the discretion of its authors, a typical NetBox plugin looks something like this:
+
+```no-highlight
+plugin_name/
+  - plugin_name/
+    - templates/
+      - *.html
```
lampwins: I think we should recommend the practice of nesting templates within an app directory as that seems to be best practice and helps prevent namespace collisions when django searches for templates:
```
templates/
  plugin_name/
    *.html
```

docs/plugins/development.md
```diff
@@ -0,0 +1,243 @@
...
+    required_settings = []
+    default_settings = {
+        'loud': False
+    }
+```
+
+#### PluginConfig Attributes
+
+* `name` - Raw plugin name; same as the plugin's source directory
+* `author_name` - Name of plugin's author
```
lampwins: I added `author_email` recently, as it seemed pertinent.

docs/plugins/development.md
```diff
@@ -0,0 +1,243 @@
...
+    sound = models.CharField(max_length=50)
+
+    def __str__(self):
+        return self.name
+```
+
+Once you have defined the model(s) for your plugin, you'll need to create the necessary database schema migrations as well. This can be done using the Django `makemigrations` management command:
+
+```no-highlight
+$ ./manage.py makemigrations netbox_animal_sounds
```
lampwins: They will need to install the plugin package as it exists before this will work.

docs/plugins/development.md
```diff
@@ -0,0 +1,243 @@
...
+@admin.register(Animal)
+class AnimalAdmin(admin.ModelAdmin):
+    list_display = ('name', 'sound')
+``` 
+
+This will display the plugin and its model in the admin UI. Staff users can create, change, and delete model instances via the admin UI without needing to create a custom view.
+
+![NetBox plugin in the admin UI](../media/plugins/plugin_admin_ui.png)
+
+## Views
```
lampwins: It might be worth linking to our existing development docs on the base view classes NetBox uses. That section needs some work, but we can do that over time.

lampwins (approved):
This is good. One general note, you briefly cover it, but it would be good to discuss that if you don't want to include certain functionalities, those areas need not be included in a plugin at all. For example, if you don't want models, you don't need models.py or if you don't have an api in you don't need the api directory at all.

Also, we need a section on installing a plugin as an end user.

docs/configuration/optional-settings.md
```diff
@@ -299,6 +299,24 @@ Determine how many objects to display per page within each list of objects.
...
+
+Default: Empty
+
+This parameter holds configuration settings for individual NetBox plugins. It is defined as a dictionary, with each key using the name of an installed plugin. The specific parameters supported are unique to each plugin: Reference the plugin's documentation to determine the supported parameters.
+
+Note that `PLUGINS_ENABLED` must be set to True for this to take effect.
+
+---
+
+## PLUGINS_ENABLED
```
lampwins: We should add a big red warning here, something like:
```
Warning: plugins extend NetBox by allowing external code to be run, and have access to all NetBox data. Only install and run plugins from trusted sources.
```

docs/plugins/development.md
jeremystretch: I'm going to take a hard line against providing any sort of API for our internal view functions. They are for internal use only and indefinitely subject to change. Plugin authors need to either create their own views or accept that integration might randomly break in any release.

docs/plugins/development.md
jeremystretch: IMO that seems redundant. Can you give an example of a problem that would prevent?

lampwins: If I develop a plugin that injects content into say the device and IP address pages, it is reasonable that I might do this:
```
templates/
    dcim/
        device.html
    ipam
        ipaddress.html
```
When django resolves the render command with the string `"dcim/device.html"` this created a namespace collision within core and in fact, in this case, django tries to render our core template a second time but fails due to lack of context.

I can't find any specific django docs but it does seem to be a general convention with a purpose. All of these apps to it: [oauth2_provider](https://github.com/jazzband/django-oauth-toolkit/tree/master/oauth2_provider), [debug_toolbar](https://github.com/jazzband/django-debug-toolbar/tree/master/debug_toolbar), [pipeline](https://github.com/jazzband/django-pipeline/tree/master/pipeline)

jeremystretch: Oh, I see. I was thinking between core and plugin, rather than between two plugins.

PR #4403, opened 2020-03-25
Closes #4401: Simplify registration process for pluin menu items

### LINKED ISSUE #4401: Simplify approach for registering plugin navigation items
### Proposed Changes
I'd like a to adopt a simpler mechanism than what currently exists in the `3351-plugins` branch for plugins to register navigation menu items. We have a few options, which I'll outline here.

### Current approach: Signals

The current approach requires three components. First, we declare the menu items in a file:

**navigation.py**
```python
class NavLink1(PluginNavMenuLink):
    link = 'plugins:my_plugin:view1'
    link_text = 'Link A'

class NavLink2(PluginNavMenuLink):
    link = 'plugins:my_plugin:view2'
    link_text = 'Link B'
```

Next, we create a function to return these classes, and register it as a signal receiver:

**signals.py**
```python
@receiver(register_nav_menu_link_classes)
def nav_menu_link_classes(**kwargs):
    return [NavLink1, NavLink2]
```

Finally, we extend our PluginApp to import this function on `ready()`. (We could probably build this part into the PluginApp class.)

**__init__.py**
```python
class MyPluginConfig(PluginConfig):
    ...

    def ready(self):
        from . import signals
```

### Proposal: Import of a known variable

The first option I propose is to direct plugin authors to define a single variable within a standard file, i.e. `navigation.py`, which contains all navigation menu items for their plugin. (This is very similar in concept to the way a plugin currently declares URL patterns in `urls.py`.)

**navigation.py**
```python
class NavLink1(PluginNavMenuLink):
    link = 'plugins:my_plugin:view1'
    link_text = 'Link A'

class NavLink2(PluginNavMenuLink):
    link = 'plugins:my_plugin:view2'
    link_text = 'Link B'

menu_items = (NavLink1, NavLink2)
```

NetBox would automatically import and register the classes in `menu_items` (if defined by the plugin). This could be handled by the PluginConfig's `ready()` method, like how we're currently handling signals.

With a minor tweak to the PluginNavMenuLink class to accept attributes on initialization, the above example could be further simplified to skip naming the individual instances:

**navigation.py**
```python
menu_items = (
    PluginNavMenuLink(
        link='plugins:my_plugin:view1',
        link_text='Link A'
    ),
    PluginNavMenuLink(
        link='plugins:my_plugin:view2',
        link_text='Link B'
    ),
)
```

### Justification
I prefer this approach for four reasons:

1. It minimizes the amount of code needed from the plugin author.
2. It is much easier to understand what's happening, especially for someone unfamiliar with Django signals.
3. It provides the simplest possible integration point: A single variable.
4. This approach is already well-established by Django's URL patterns mechanism (wherein each app provides a `urlpatterns` variable).

jeremystretch:
### Closes: #4401

- Automatically import `menu_items` from a plugin's `navigation.py` (if it exists)
- Declare `get_menu_items()` on PluginConfig to allow a plugin author to override the default behavior
- Require instantiating PluginNavMenuLink to create menu items to enforce requires attributes

An example `navigation.py` within a plugin looks like this:

```python
from extras.plugins import PluginNavMenuButton, PluginNavMenuLink
from utilities.choices import ButtonColorChoices

menu_items = (
    PluginNavMenuLink(
        link='plugins:plugin_name:view_a',
        link_text='Link 1',
        buttons=(
            PluginNavMenuButton('view_foo', 'Button A', 'fa-info', ButtonColorChoices.BLUE),
            PluginNavMenuButton('view_bar', 'Button B', 'fa-warning', ButtonColorChoices.GREEN),
        )
    ),
    PluginNavMenuLink(
        link='plugins:plugin_name:view_b',
        link_text='Link 2'
    ),
)
```

Edit: Will also update the docs accordingly if this is agreed upon.

### DIFF
--- a/netbox/extras/plugins/__init__.py
+++ b/netbox/extras/plugins/__init__.py
@@ -46,22 +46,18 @@ class PluginConfig(AppConfig):
     # Caching configuration
     caching_config = {}
 
+    # Default integration paths. Plugin authors can override these to customize the paths to
+    # integrated components.
+    menu_items = 'navigation.menu_items'
+
     def ready(self):
 
         # Register navigation menu items (if defined)
-        register_menu_items(self.verbose_name, self.get_menu_items())
-
-    def get_menu_items(self):
-        """
-        Default method to import navigation menu items for a plugin from the default location (menu_items in a
-        file named navigation.py). This method may be overridden by a plugin author to import menu items from
-        a different location if needed.
-        """
         try:
-            menu_items = import_string(f"{self.__module__}.navigation.menu_items")
-            return menu_items
+            menu_items = import_string(f"{self.__module__}.{self.menu_items}")
+            register_menu_items(self.verbose_name, menu_items)
         except ImportError:
-            return []
+            pass
 
 
 #
--- a/netbox/extras/plugins/__init__.py
+++ b/netbox/extras/plugins/__init__.py
@@ -1,11 +1,16 @@
 import collections
 import inspect
 
-from django.apps import AppConfig, apps
+from django.apps import AppConfig
 from django.template.loader import get_template
+from django.utils.module_loading import import_string
 
 from extras.registry import registry
-from .signals import register_detail_page_content_classes, register_nav_menu_link_classes
+from .signals import register_detail_page_content_classes
+
+
+# Initialize plugin registry stores
+registry['plugin_nav_menu_links'] = {}
 
 
 #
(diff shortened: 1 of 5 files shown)


netbox/extras/plugins/__init__.py
```diff
@@ -41,6 +46,23 @@ class PluginConfig(AppConfig):
     # Caching configuration
     caching_config = {}
 
+    def ready(self):
+
+        # Register navigation menu items (if defined)
+        register_menu_items(self.verbose_name, self.get_menu_items())
+
+    def get_menu_items(self):
```
lampwins: Rather than having an author potentially override a method, what if we just let them specify the module name and have it default to `navigation`? This is similar to how django handles the models module https://docs.djangoproject.com/en/3.0/ref/applications/#django.apps.AppConfig.models_module

lampwins (commented):
Any idea if the order Django runs AppConfig.ready() methods is deterministic? I would think it would be based on the order of `INSTALLED_APPS` but is that for sure?

lampwins:
I found the answer and it is yes :)
> The application registry is initialized in three stages. At each stage, Django processes all applications in the order of INSTALLED_APPS.
https://docs.djangoproject.com/en/3.0/ref/applications/#how-applications-are-loaded

netbox/extras/plugins/__init__.py
jeremystretch: That's fine too. Do you want to make it an attribute of PluginConfig?

lampwins: Yeah, that's what I would do. Something like `navigation_module = 'navigation'`?

PR #4412, opened 2020-03-26
Add tags example to custom scripts documentation

### LINKED ISSUE #4410: Site(tags) are not created when creating a Site using NetBox scripts function
### Environment
* Python version:  3.6.8
* NetBox version:  2.7.10

### Steps to Reproduce
Using the below script in the scripts directory for netbox & entering a new site name, creates the site, but without tags.
```python
from django.utils.text import slugify

from dcim.choices import DeviceStatusChoices, SiteStatusChoices
from dcim.models import Device, DeviceRole, DeviceType, Site, Region
from extras.scripts import *


class NewColoScript(Script):

    class Meta:
        name = "New Colo"
        description = "Provision a new colo site"
        field_order = ['site_name']

    site_name = StringVar(
        label="Site Name",
        description="Site Code Per the Colo Provider",
        required=True
    )


    def run(self, data, commit):


        # Create the new site
        site = Site(
            name=data['site_name'].upper(),
            slug=slugify(data['site_name']),
            tags=["colo", "test"]
        )
        site.save()
        self.log_success("Created new site: {}".format(site))

```

I have even tried by pre-creating the tags and referencing the name(s) as well as referencing the tag id.

### Expected Behavior
Create the site and populate tags.

### Observed Behavior
Create the site without tags.

ryanmerolle:
### Fixes: #4410
Adds to the sample script with an example of the proper way how to add tags to objects since it uses the django-taggit functionality.

### DIFF
--- a/docs/additional-features/custom-scripts.md
+++ b/docs/additional-features/custom-scripts.md
@@ -237,6 +237,7 @@ class NewBranchScript(Script):
             status=SiteStatusChoices.STATUS_PLANNED
         )
         site.save()
+        site.tags.add("colo", "test") # Unlike other object attributes, any time you want to add tags to an object, you must do so AFTER the object has been saved.
         self.log_success("Created new site: {}".format(site))
 
         # Create access switches


jeremystretch:
Per my comment on #4410, this is out of scope for documentation on custom scripts. I am not going to establish a precedent for people to start tacking on unrelated topics.

ryanmerolle:
I figured this was fine because you included a sample script, and I just added a line to the sample script to prevent people from getting tripped up on this single unique attribute for objects.  I also figured this was not quite documenting and setting a precedent as I know you would have concern with.


I’ll move some examples to the reports repo, and we can link to it in the documentation.

PR #4422, opened 2020-03-29
update WS2012 references

athompson-merlin:
Eliminates potential confusion during install, clarifying that these instructions would also apply to e.g. WS2016, etc.


athompson-merlin:
Argh... ffs, that's NOT what I wanted to include in a pull request! Sorry for the noise..

PR #4433, opened 2020-04-01
Change PLUGINS_ENABLED to a list of specific plugins (PLUGINS)

jeremystretch:
- Changes `PLUGINS_ENABLED` from a boolean to a list of plugin names; renamed to `PLUGINS`.
- Requires a plugin to declare a `config` pointing to a specific PluginConfig (similar to Django's `default_app_config`). We _could_ find PluginConfig subclasses automatically, but this is a) more explicit, and b) allows a plugin author to change out configs dynamically in response to environment variables or other factors.
- Removes the need for declaring entry points in `setup.py`.

### DIFF
--- a/docs/configuration/optional-settings.md
+++ b/docs/configuration/optional-settings.md
@@ -299,6 +299,17 @@ Determine how many objects to display per page within each list of objects.
 
 ---
 
+## PLUGINS
+
+Default: Empty
+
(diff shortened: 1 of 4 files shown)


lampwins (approved):

DanSheps (approved):
LGTM

PR #4457, opened 2020-04-06
Assign to Power Device and Device Resources

CobbyJ:
Hello,

You have a great system, but there is a lot to be desired for example there is no way to add recsources on a physical device, no idea why you didnt add that and only for VM's

There is no way to assign a PDU and PORT to the specific PDU so you have to hunt for it.

The IP system is way to frustratoring i cant seem to find anyway to assign single ip's from a prefix to a device.

Also would be great to add custom fields, and a URL link for PDU's and switches for quick access.


jeremystretch:

PR #4478, opened 2020-04-13
Fixes #4474: Populating Devive Type Attr in Device Bulk Edit view

### LINKED ISSUE #4474: Device type does not populate when bulk editing devices.
### Environment
* Python version:  3.6.9
* NetBox version:  2.7.12

### Steps to Reproduce
On the device list page, select a few devices, then click "Edit Selected". 

This is reproducible on DanShep's [demo site](https://master.netbox.dansheps.com/) running 2.7.11, as well as our instance running 2.7.12.

### Expected Behavior
The "Device type" field in the form should populate when selected.

### Observed Behavior
The "Device type" field remains empty. Other fields populate correctly.

kobayashi:
### Fixes: 4474

I know this is for pre-accepted issue. If accepted, we can go further. 

`form.js` in Bulk Edit view cannot not set record.text because  `element.getAttribute('value-field')` in js cannot find the attribute and Device Type does not have name attribute.

https://github.com/netbox-community/netbox/blob/c593eca9364e0fcca54b7cd4986470da795cb065/netbox/project-static/js/forms.js#L223-L225

### DIFF
--- a/netbox/project-static/js/forms.js
+++ b/netbox/project-static/js/forms.js
@@ -221,7 +221,7 @@ $(document).ready(function() {
                 var results = data.results;
 
                 results = results.reduce((results,record,idx) => {
-                    record.text = record[element.getAttribute('display-field')] || record.name;
+                    record.text = record[element.getAttribute('display-field')] || record.name || record.model;
                     record.id = record[element.getAttribute('value-field')] || record.id;
                     if(element.getAttribute('disabled-indicator') && record[element.getAttribute('disabled-indicator')]) {
                         // The disabled-indicator equated to true, so we disable this option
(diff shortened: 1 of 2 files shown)


netbox/project-static/js/forms.js
```diff
@@ -221,7 +221,7 @@ $(document).ready(function() {
                 var results = data.results;
 
                 results = results.reduce((results,record,idx) => {
-                    record.text = record[element.getAttribute('display-field')] || record.name;
+                    record.text = record[element.getAttribute('display-field')] || record.name || record.model;
```
jeremystretch, line 224: Instead of appending individual fields, we should ensure that the `display-field` attribute is being set on the device type form field.

jeremystretch:
Fixed this in 67f2cdc9 by restoring `display_name` on the APISelect widget.

PR #4494, opened 2020-04-15
Make sure systemd doesn't try to start NetBox before Redis is ready

### LINKED ISSUE #4493: The contrib systemd unit file can cause first NetBox WSGI start to fail
### Environment
* Python version: 3.6.8
* NetBox version: 2.8.0

### Steps to Reproduce
1. Configure systemd with `contrib/netbox.service`.
2. Make sure Redis takes a few seconds to start.
3. systemd attempts to start NetBox before Redis has completed starting up and loading data and NetBox will error.

### Expected Behavior
NetBox starts only after Redis is ready.

### Observed Behavior
NetBox systemd unit file does not depend on Redis so it will not wait Redis to signal to systemd that it has completed startup.

bluikko:
### Fixes: #4493

In some cases Redis will take some seconds to become ready.
If NetBox is started with systemd unit file before Redis is ready it will fail with an error.

Changing the After/Wants to redis.service will make sure NetBox start is not attempted before Redis signals it is ready. This needs in `redis.conf`:
```
supervised systemd
```

### DIFF
--- a/contrib/netbox.service
+++ b/contrib/netbox.service
@@ -1,8 +1,8 @@
 [Unit]
 Description=NetBox WSGI Service
 Documentation=https://netbox.readthedocs.io/en/stable/
-After=network-online.target
-Wants=network-online.target
+After=redis.service
+Wants=redis.service
 
 [Service]
 Type=simple


jeremystretch:
Thank you for your interest in contributing to NetBox, however it appears there is no accepted issue that correlates to this pull request.  Before pull requests are opened, we require an accepted issue as per our [contributing guide](https://github.com/digitalocean/netbox/blob/master/CONTRIBUTING.md).  Please wait for #4493 to be accepted before further work is done on this pull request.

PR #4509, opened 2020-04-20
Fixes #4463 Describe APISelect for custom scripts object filter

### LINKED ISSUE #4463: ObjectVar queryset returns all items - does not filter
### Environment
* Python version:  3.7 
* NetBox version: 2.7.11  

### Steps to Reproduce
1.  Added following code as per the Netbox New Branch Script example

```
    class Meta:
        name = "New Branch"
        description = "Provision a new branch site"
        fields = ['site_name', 'switch_count', 'switch_model']

    site_name = StringVar(
        description="Name of the new site"
    )
    switch_count = IntegerVar(
        description="Number of access switches to create"
    )
    switch_model = ObjectVar(
        description="Access switch model",
        queryset = DeviceType.objects.filter(
            manufacturer__name='Cisco',
            model__in=['Catalyst 3560X-48T', 'Catalyst 3750X-48T']
        ),
        widget=APISelect(display_field='model')
    )
```

2. Go the script page and click the model type drop down.

### Expected Behavior
Only 'Catalyst 3560X-48T', 'Catalyst 3750X-48T'] are displayed.

### Observed Behavior
All model types are displayed.

FURTHER INFO ---
I've been having this issue on a Site.objects.Filter() i've created that i could not get to work (it would display all records.

Tried to replicate via the site script example as above. But no fields would display initially.

Read the previous article related to widget=APISelect(display_field='model') and added that. Then saw it also displayed all device types instead of just two.

kobayashi:
### Fixes: #4463
To utilize custom scripts object filtering, Add `APISelect` description in custom scripts page. The description is from the [docstring](https://github.com/netbox-community/netbox/blob/5aadfff1de2ba7a07faa327c519df4038bab6e7e/netbox/utilities/forms.py#L252-L270). This also modify the sample script.

### DIFF
--- a/docs/additional-features/custom-scripts.md
+++ b/docs/additional-features/custom-scripts.md
@@ -156,9 +156,27 @@ direction = ChoiceVar(choices=CHOICES)
 
 ### ObjectVar
 
-A NetBox object. The list of available objects is defined by the queryset parameter. Each instance of this variable is limited to a single object type.
+A NetBox object. The list of available objects is defined by the `APISelect` widget. Each instance of this variable is limited to a single object type.
 
 * `queryset` - A [Django queryset](https://docs.djangoproject.com/en/stable/topics/db/queries/)
+* `widget` - A form widget for APISelect
+
+    APISelect is a select widget populated via an API call
+
+    - `api_url` - API endpoint URL. Required if not set automatically by the parent field.
+    - `display_field` (Optional) - Field to display for child in selection list. Defaults to `name`.
+    - `value_field` (Optional) - Field to use for the option value in selection list. Defaults to `id`.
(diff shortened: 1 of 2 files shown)


jeremystretch:
Feels like we've jumped the gun a bit here. The APISelect widget in its current form is confusing even to me, let alone a casual user. IMO we should work on simplifying its signature and options before we commit to a published API for it.

kobayashi:
Agreed.

I think `additional_query_params` is the most used field to filter objects, so how about just add only this parameter and example in docs right now? Then, we can go ahead to improve usability.

kobayashi:
This documentation is not good for usage of objectvar. I should make it suitable for script before this documentation, so for now I closes this PR.

Issue #4504, opened 2020-04-18
Troubleshooting LDAP instructions still mention supervisord


markkuleinio:
### Change Type
[ ] Addition
[X] Correction
[ ] Deprecation
[ ] Cleanup (formatting, typos, etc.)

### Area
[X] Installation instructions
[ ] Configuration parameters
[ ] Functionality/features
[ ] REST API
[ ] Administration/development
[ ] Other

### Proposed Changes
The LDAP authentication configuration troubleshooting section in https://netbox.readthedocs.io/en/stable/installation/5-ldap/ contains references to supervisord. They should be replaced by systemd conventions.

PR #4568, opened 2020-05-04
Replace supervisord with systemd

koratfood:
Update LDAP troubleshooting steps so that they are consistent with the rest of the documentaiton, which nowadays expects us to be running netbox via systemd instead of supervisord.


koratfood:
Forgot to attach this one to issue #4504. Will re-submit.

Issue #4519, opened 2020-04-21
Multiple interfaces connected to a single circuit termination


steffann:
### Environment
* Python version: 3.8
* NetBox version:  2.8

### Proposed Functionality
The current implementation cannot handle the following situation:

```
       +--------------+
INT1 ----- FP1        |
       |         RP ----- CT --- Circuit
INT2 ----- FP2        |
       +--------------+
```

WhereINT is an interface, FP is a front port, RP is a rear port and CT is a circuit termination.

Such a situation can occur for example when modelling a DWDM multiplexer connected to a leased line, when creating all the cabling on one end of the circuit before connecting anything to the other side of the circuit.

When creating the setup shown above both INT1 and INT2 will consider CT their connected endpoint. In the current implementation this is not allowed (`_connected_circuittermination` is a `OneToOneField(CircuitTermination)` on the interface, and `connected_endpoint` is a `OneToOneField(Interface)` on the circuit termination).

There are two possible ways to deal with this that I can see:
1. Allow the CT to be considered a valid connected endpoint for multiple interfaces> This would mean: change the `_connected_circuittermination` on the interface to a `ForeignKey` with a `related_name` of `connected_endpoints` and remove the `connected_endpoint` field from the circuit. Adapting the UI and API to match.
2. Don't allow this, and check if there is a multiplexer in the path that would allow connecting multiple interfaces to the same CT. If that is the case, don't store the CT as the connected endpoint. This would require no changes to the UI and API.

Which choice is the best? Discuss…

### Use Case
Choice 1 would be useful in cases where having multiple interfaces connected to the same CT is the final goal. For example when buying a service from an ISP that involves multiple wavelengths for different services. I know of cases where the ISP provides plain internet on one wavelength and private interconnects on others, so plausible scenario.

Choice 2 would be fine if we don't need to support such wavelength services. In that case there will be a corresponding multiplexer on the other side of the circuit, and the connected endpoint of the interface will be filled when something is connected at the other end.

### Database Changes
When choosing option 1: changing a one-to-one relationship into a one-to-many.

### External Dependencies
None

iunderwood:
Additional use case: using single strands in a fiber, e.g. BiDi optics.

Without being too familiar the backend of the system, perhaps allow a circuit to be flexibly usable as a cable connecting rear ports.

semaja2:
I have a need for this as well in regards to leased dark fibre circuits and using BiDi modules

jeremystretch:
[...]

PR #4592, opened 2020-05-07
4591: make single front/rear port work when between panels

### LINKED ISSUE #4591: 1:1 patch panels fail when between 1:n panels
### Environment
* Python version: 3.6.9
* NetBox version: 2.8.2

### Steps to Reproduce
   `configuration.py`.
2. Create the following setup:

```
                1                                           2
[Device 1] -----------+                               +----------- [Device 2]
      Iface1          |                               |          Iface1
                  FP1 |       3               4       | FP1
                  [Panel 1] ----- [Panel 5] ----- [Panel 2]
                  FP2 |  RP1     RP1     FP1     RP1  | FP2
      Iface1          |                               |          Iface1
[Device 3] -----------+                               +----------- [Device 4]
                5                                           6
```

3. get `Exception: Invalid position for Rear Port 1 (1 positions): 2)`

When writing this issue I created the cables in numerical order, but the order of creation of the cables doesn't really matter. As soon as cables 3 and 5 are in place, NetBox will try to trace the connection from device 3 to determine what its connected endpoint is, and it will raise the exception upon reaching panel 5.

### Expected Behavior
I expected the front ports of panel 1 to correspond to the front ports of panel 2.

### Observed Behavior
NetBox tries to map the front ports of panel 1 to the front ports of panel 5, which doesn't work because it is a 1:1 pass-though panel.

steffann:
### Fixes: #4591

This patch considers 1:1 patch panels as cable extenders and will not push/pop the front port position (which is always 1) on the `position_stack`. That way 1:1 panels don't interfere with 1:n patch panels.

### DIFF
--- a/netbox/dcim/tests/test_models.py
+++ b/netbox/dcim/tests/test_models.py
@@ -501,9 +501,12 @@ class CablePathTestCase(TestCase):
             Device(device_type=devicetype, device_role=devicerole, name='Panel 2', site=site),
             Device(device_type=devicetype, device_role=devicerole, name='Panel 3', site=site),
             Device(device_type=devicetype, device_role=devicerole, name='Panel 4', site=site),
+            Device(device_type=devicetype, device_role=devicerole, name='Panel 5', site=site),
         )
         Device.objects.bulk_create(patch_panels)
-        for patch_panel in patch_panels:
+
+        # Create patch panels with 4 positions
+        for patch_panel in patch_panels[:4]:
             rearport = RearPort.objects.create(device=patch_panel, name='Rear Port 1', positions=4, type=PortTypeChoices.TYPE_8P8C)
             FrontPort.objects.bulk_create((
                 FrontPort(device=patch_panel, name='Front Port 1', rear_port=rearport, rear_port_position=1, type=PortTypeChoices.TYPE_8P8C),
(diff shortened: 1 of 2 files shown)


netbox/dcim/tests/test_models.py
```diff
@@ -512,6 +512,11 @@ def setUpTestData(cls):
                 FrontPort(device=patch_panel, name='Front Port 4', rear_port=rearport, rear_port_position=4, type=PortTypeChoices.TYPE_8P8C),
             ))
 
+        # Create a 1-on-1 patch panel
+        patch_panel = Device.objects.create(device_type=devicetype, device_role=devicerole, name='Panel 5', site=site)
+        rearport = RearPort.objects.create(device=patch_panel, name='Rear Port 1', positions=1, type=PortTypeChoices.TYPE_8P8C)
+        FrontPort.objects.create(device=patch_panel, name='Front Port 1', rear_port=rearport, rear_port_position=1, type=PortTypeChoices.TYPE_8P8C)
```
jeremystretch, line 518: IMO it would be cleaner to do the panel creation with the others on line 499, then modify the for loop to only act on the first four. That way it's more clear that you're creating five patch panels, four with multiposition rear ports and one with a single position rear port.

It also ensures that the last patch panel is included in the set should we need to assign `patch_panels` to the instance for reference by a test method.

netbox/dcim/tests/test_models.py
```diff
@@ -554,17 +559,17 @@ def test_connection_via_single_rear_port(self):
         Test a connection which passes through a single front/rear port pair.
 
                      1               2
-        [Device 1] ----- [Panel 1] ----- [Device 2]
+        [Device 1] ----- [Panel 5] ----- [Device 2]
```
jeremystretch, line 562: This test needs to stay the way it is to ensure we test against a multi-position rear port; a new test is needed for the topology with a single-position rear port. I'd rename the current test to `test_connection_via_multi_position_rear_port` and name the new one `test_connection_via_single_position_rear_port`.

I think we're actually missing this condition in the cable validation logic: connecting an interface to a multi-position port probably should not be allowed. (The validation currently only checks if both terminations are rear ports.) This hits on the same conceptual question as #4519: How to treat multiple "A side" terminations mapped to a single "B side" termination.

netbox/dcim/tests/test_models.py
```diff
@@ -592,6 +597,155 @@ def test_connection_via_single_rear_port(self):
         self.assertIsNone(endpoint_a.connection_status)
         self.assertIsNone(endpoint_b.connection_status)
 
+        # Recreate cable 1 to test creating the cables in reverse order (RP first, FP second)
+        cable1.save()
```
jeremystretch, line 601: It may be preferable to recreate the cable explicitly in the interest of clarity. Although it technically works, calling `save()` on an object after calling `delete()` may be confusing to the reader.

jeremystretch (changes requested):

netbox/dcim/tests/test_models.py
steffann: I restored the original `test_connection_via_single_rear_port` code and added a TODO item for when #4604 is going to be merged.

PR #4623, opened 2020-05-11
Notes on multiprocessing metrics and gunicorn vs uwsgi

### LINKED ISSUE #3779: Add configuration to prevent inaccurate prometheus metrics
### Environment
* Python version:  3.6.8
* NetBox version:  2.6.7

### Proposed Functionality
When running `django-prometheus` in a multiprocessed environment (as with `gunicorn` or `uwsgi`), the metrics files are generated using the `pid` of each worker process. This becomes an issue as worker processes recycle (by whatever cause) and get new a new `pid` and the old metrics files persist - and those numbers are scraped by `prometheus_client`. This topic has been discussed further in the main `prometheus_client` repo: https://github.com/prometheus/client_python/issues/275 & https://github.com/prometheus/client_python/issues/204

I have combined the two fixes documented (from [here](https://github.com/prometheus/client_python/issues/275#issuecomment-504755024) and [here](https://github.com/korfuri/django-prometheus/blob/master/documentation/exports.md#exporting-metrics-in-a-wsgi-application-with-multiple-processes-globally)). The below block would be added near the top of the Netbox `settings.py`. There's no impact to users not running with metrics enabled. And if users do have metrics enabled, the below block is only executed if the [prometheus_multiproc_dir](https://netbox.readthedocs.io/en/stable/additional-features/prometheus-metrics/) environment variable is set (which is essentially mandatory to get accurate metrics with a multiprocessed wsgi server).

```
if "prometheus_multiproc_dir" in os.environ:
    try:
        import prometheus_client
        import uwsgi
        prometheus_client.values.ValueClass = prometheus_client.values.MultiProcessValue(
            _pidFunc=uwsgi.worker_id)
    except ImportError:
        pass  # not running in uwsgi
```

### Use Case
Any user that wants to take advantage of the prometheus metrics will eventually have performance issues (as the number of metrics files grows) and inaccurate metrics.

### Database Changes
N/A

### External Dependencies
N/A

tyler-8:
### Fixes: https://github.com/netbox-community/netbox/issues/3779

This adds a paragraph recommending usage of `uwsgi` over `gunicorn` for accurate metrics when using multiprocessing. It includes a link to my comments in the original issue rather than including the nitty-gritty details in the docs themselves.


docs/additional-features/prometheus-metrics.md
```diff
@@ -32,3 +32,7 @@ This can be setup by first creating a shared directory and then adding this line
 ```
 environment=prometheus_multiproc_dir=/tmp/prometheus_metrics
 ```
+
+#### Accuracy
+
+If having long-term-accurate metrics in a multiprocess environment is important to you then it's recommended you use the `uwsgi` library instead of `gunicorn`. The issue lies in the way `gunicorn` tracks worker processes (vs `uwsgi`) which helps manage the metrics files created by the above configurations. If you're using Netbox w/ `gunicorn` in a containerized enviroment following the 1-process-per-container methodology, then you will likely not need to change to `uwsgi`. More details can be found in [this issue](https://github.com/netbox-community/netbox/issues/3779#issuecomment-590547562).
```
jeremystretch, line 38: Needs a bit of cleanup.

> long-term-accurate metrics

"accurate long-term metrics"

> Netbox w/ `gunicorn`

"NetBox with gunicorn"

> 1-process-per-container

"one-process-per-container"

> More details can be found in [this issue]

Please cite the issue number in the link text. (This retains the reference in case the markdown gets stripped for some reason.)

tyler-8: Done!

PR #4630, opened 2020-05-12
fix get parameters lost on create and add another

### LINKED ISSUE #4629: Interface assignment get lost after IP address "create and add another"
### Environment
* Python version: 3.7 (should be on any)
* NetBox version: 2.8.3

### Steps to Reproduce
   `configuration.py`.
2. Navigate to any device with an interface
3. Click the + at the and of the interface row in the interface table to add an IP address to an interface
4. Fill out the form and click "Create and add another"

### Expected Behavior
- Another IP address form with filled out interface assignment appears.
- `interface` GET-parameter is set to the appropriate interface-ID.

### Observed Behavior
- Another IP address form without filled out interface assignment appears.
- `interface` GET-parameter is missing.

### Further information
With the implementation of clone able models in the `ObjectEditView.post()` function the redirection GET-parameters are build only from clone_fields and other maybe existing GET-parameters are fully ignored.

JulianJacobi:
### Fixes: #4629

- add get_clone_fields to get clone fields get parameters as QueryDict
- change prepare_clone_fields to use get_clone_fields and build
querystring with urlencode function instead of manual building.
- add non clone_field GET-parameters to params QueryDict on
ObjectEditView.post()

### DIFF
--- a/netbox/utilities/utils.py
+++ b/netbox/utilities/utils.py
@@ -195,12 +195,12 @@ def render_jinja2(template_code, context):
     return Environment().from_string(source=template_code).render(**context)
 
 
-def prepare_cloned_fields(instance):
+def get_cloned_fields(instance):
     """
-    Compile an object's `clone_fields` list into a string of URL query parameters. Tags are automatically cloned where
+    Compile an object's `clone_fields` list into a QueryDict. Tags are automatically cloned where
     applicable.
     """
-    params = {}
+    params = QueryDict(mutable=True)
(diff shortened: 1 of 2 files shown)


DanSheps:
Thank you for your interest in contributing to Netbox, however it appears there is no accepted issue that correlates to this pull request.  Before pull requests are opened, we require an accepted issue as per our [contributing guide](https://github.com/digitalocean/netbox/blob/master/CONTRIBUTING.md).  Please first open an issue and wait for it to be accepted before further work is done on this pull request.

Issue #1590, opened 2017-10-13
Reload report code without restarting server


candlerb:
### Issue type
[X] Feature request 
[ ] Bug report      
[ ] Documentation   

### Environment
* Python version:  3.5.2
* NetBox version:  2.2.1

### Description
* You run a report from the web UI
* There's a bug in the python code (e.g. `self.log_fail` instead of `self.log_failure`) which causes an exception, or you just want to change the behaviour
* You fix the code, and re-test it from the command line (`python manage.py runreport <name>`)
* However if you re-run it from the UI, it still fails

At the moment you need to do a full shutdown/restart of netbox.  It would be nice if it could pick up that the report file has changed since it was last loaded, and reload it automatically.

RyanBreaker:
@jeremystretch In what way are reports implemented? Is it a from-scratch feature or is it something built upon from a module?

Yakelixir:
Help still wanted? I have been working on this issue in my own

arsenakishev:
I just tested this recently. After fixing a bug that causes the exception, the reports page UI shows a passing result. Can anyone comment on this issue? It seems like it has been resolved.

zelfix:
@jeremystretch  It's not an issue of NetBox. Users can use the option "reload" of gunicorn server for auto restarting gunicorn workers when source code changes. http://docs.gunicorn.org/en/19.0/settings.html#reload

TakeMeNL:
Just a small update, running Netbox via docker-compose (ninech/netbox-docker) and i can edit report scripts live, not reload whatsoever.

candlerb:
Yes, I can confirm the behaviour is correct now - tested with v2.5.2. I can break a report, run it in UI (get exception), fix it, run it in UI (and it works), without any restarts.

Maybe the fix happened in Netbox, maybe in Django.

DanSheps:
I don't think this should be closed out yet.  gunicorn is not the only way to establish the workers.  For example mod_wsgi with Apache or uWSGI.

I will let @jeremystretch weigh in though perhaps.

PR #4638, opened 2020-05-13
Add support for script and report packaging in plugins (#4573)

### LINKED ISSUE #4573: Support plugins as a delivery mechanism for reports and custom scripts
### Environment
* Python version:  3.7
* NetBox version:  2.8.1

### Proposed Functionality
This proposal is to add two new interfaces to the plugins framework to support the packaging and delivery of reports and custom scripts through a plugin. This proposal does not intend to change in any way the existing API for either reports or custom scripts, just the ability to make them a part of a plugin.

The proposed mechanism follows plugin menu items and template content classes, whereby a new plugin interface will expose scripts and reports modules with the respective attribute lists.

While this new method is intended to replace the local reports and scripts directories, it is not necessary that we deprecate these means immediately, if ever. It would, however, allow us to reduce the instances of code that performs dynamic imports for one reason or another.

### Use Case
With the addition of the plugins framework, NetBox now has three separate interfaces for extensibility: reports, custom scripts, and plugins. If a user wishes to make use of all three, they must navigate the different interfaces. Granted, reports and custom scripts are similar but require using different directories.

Allowing reports and custom scripts to be bundled in a plugin would offer a simplified delivery interface for users and developers.

There are two primary use cases this is born from:
1. An organization that uses a plugin to contain org specific business logic and functionality, can bundle their reports and scripts into the same plugin for ease of development and management.
2. Aid the publishing and reuse of reports and scripts by allowing them to be a part of python packages supported by a formal NetBox interface.

### Database Changes
None

### External Dependencies
None

glennmatthews:
### Fixes: #4573 

Plugins can now include custom scripts and reports as part of their packaging. If a plugin registers these with the application registry (following the established design for registering template extensions and menu items), they will be automatically discovered and listed alongside the scripts and reports provided locally as per the existing implementation.

Additionally I've added the ability for report modules to register a custom `name` string to be displayed as a section header in the Reports view (just as script modules already can do for the Scripts view), as the baseline behavior of displaying just the raw module name was a bit ugly for plugin reports (for example, **Plugin_name.reports**).


netbox/extras/reports.py
```diff
@@ -22,23 +23,18 @@ def get_report(module_name, report_name):
...
-    except FileNotFoundError:
-        return None
+    return None
 
-    report = getattr(module, report_name, None)
-    if report is None:
-        return None
 
-    return report()
-
```
glennmatthews, line 39: An explanatory note: this is a rework of the `get_report` method as the prior implementation would only work for local reports (as it was hardcoded to import relative to the `REPORTS_ROOT` directory). Instead of re-importing and re-instantiating the requested `Report` subclass, this now takes advantage of the fact that `get_reports()` already returns a dict of lists of `Report` subclass instances, and simply locates and returns the desired instance. This works for both local and plugin-packaged reports as `get_reports()` already handles both.

lampwins (approved):
LGTM. I did notice it seems we don't have any existing tests for the local report/script import logic? I am not sure of the effort involved in it but I would assume it would look somewhat similar to the structure of the dummy_plugin setup for testing. @jeremystretch do you have any thoughts on this? Not sure if it is even worth it if we think the local imports is something we may deprecate in the future.

jeremystretch:
I've targeted this PR to `develop-2.9` as it introduces new functionality beyond the current plugins framework. Will take a look through it when time allows, hopefully sometime next week.

glennmatthews:
I've rebased my branch onto `develop-2.9` and moved the changelog from `version-2.8.md` to `version-2.9.md`.

jeremystretch:
@glennmatthews Thanks. Ignore the CI failure for now: It's a bug in an upstream package that will be fixed when we merge v2.8.4 into `develop-2.9`.

jeremystretch:
Using the application registry for scripts/reports may not be the best approach, because the registry is intended to be initialized once at run time and to remain static for the life of the process. Relying on it to register the presence of scripts and reports means the WSGI worker process must be restarted to acknowledge any changes in those modules on disk. I want to ensure we don't do anything that will prevent us from dynamically reloading script and report modules without restarting the WSGI worker.

Feature request #1590 was opened some time ago to support the automatic detection of new and changed scripts/reports. It has been closed, as a workaround was provided through gunicorn configuration, however I believe it should have been kept open. (Obviously, restarting the WSGI workers is much more disruptive than simply reloading the affected modules. It also remains an issue where this functionality is not supported by the chosen WSGI service.)

I understand that scripts and reports delivered via plugins are less subject to change than those which are locally derived, however we owe it to the user to ensure functional parity for both. That is, if we were to support dynamic reloading of a script in response to a change on disk for a _local_ script, we should support it for a plugin-delivered script as well.

Have you evaluated other options for registering scripts/reports that might be more accommodating to this ability? Django signals are the first thing to pop into my head as a potential solution, though I haven't thought much about how that would work.

jeremystretch:
Also, holding off on merging this until #4710 is addressed and merged into `develop-2.9`.

jeremystretch:
@glennmatthews Have you had a chance to consider my feedback above? Would like to get some movement on this PR before it gets too stale.

glennmatthews:
I haven't had the bandwidth to look into alternate approaches in depth. The simplest approach that comes to my mind would be to just call `importlib.reload(script_or_report)` before each time we invoke it, but that would potentially have undesirable side effects if the script/report does anything that it expects to be one-time setup when first loaded.

jeremystretch:
I spoke to @glennmatthews and he agreed to mothball this one until we've taken a shot at #4735 (currently earmarked for v2.10), which seeks to merge the functionality of reports into custom scripts. That should save a lot of effort on this front.

jeremystretch:
@glennmatthews @lampwins Would it make sense to include this in v2.9 but only provide support for scripts? That might actually be a good way to help signal that reports are being deprecated. I'm a bit hesitant to make the call prior to figuring out the detailed migration path for existing reports, though. What do you think?

PR #4663, opened 2020-05-19
Do not fail silently when ldap_config yields an ImportError itself

### LINKED ISSUE #4662: Fails silently on failed import inside ldap_config.py
### Environment
* Python version: 3.7.3
* NetBox version: 2.8.4

### Steps to Reproduce
1. Try and configure LDAP authentication
2. Have a typo in `ldap_config.py`s import statements

### Expected Behavior
An error message being shown.

### Observed Behavior
Netbox starts without error, but LDAP authentication does not work (obviously) and there was no hint what went wrong whatsoever.

sur5r:

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -382,8 +382,11 @@ CSRF_TRUSTED_ORIGINS = ALLOWED_HOSTS
 
 try:
     from netbox import ldap_config as LDAP_CONFIG
-except ImportError:
-    LDAP_CONFIG = None
+except ImportError as ie:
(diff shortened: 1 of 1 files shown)


DanSheps:
Thank you for your interest in contributing to Netbox, however it appears there is no accepted issue that correlates to this pull request.  Before pull requests are opened, we require an accepted issue as per our [contributing guide](https://github.com/digitalocean/netbox/blob/master/CONTRIBUTING.md).  Please first open an issue and wait for it to be accepted before further work is done on this pull request.

PR #4678, opened 2020-05-22
4525: better support for ScriptVariable fields with multiple values

### LINKED ISSUE #4525: Custom Script MultiObjectVar cannot be filled from URL parameter
### Environment
* Python version: 3.7
* NetBox version: 2.7.12

I'm building a custom script with multiple fields that I want to pre-fill from a a URL (that I will generate from a Custom Link).
It works, except for MultiObjectVar (Multiple Select) variable type.

### Steps to Reproduce
1. Create Custom Script with a MultiObjectVar field (named "test")
```
test = MultiObjectVar(
    queryset=Device.objects.filter(status="active"),
    required=True,
    label="Test field", 
)
```
2. Try to put a valid device ID in URL parameter ( http://netbox/extra/scripts/myscript/Myclass/?test=1 )
3. Or list-type format (?test=[1,2])

### Expected Behavior
Field should contain the value of ID in specified in URL parameter test (GET)

### Observed Behavior
with step 1 and 2: It doesn't fill any field. MultiObjectVar is blank/null. No error.

With step 3, as a MultiObjectVar should contain a list object, it gives:
class 'ValueError'
invalid literal for int() with base 10: ']'

If I add "default=[1,2]" to step 1 MultiObjectVar, it will fill it with the right values.
"default", for this MultiObjectVar variable, needs a list object. int() will give an error.

If you replace MultiObjectVar with ObjectVar and repeat step 2, it will fill the field with the right value.

Filling fields from GET parameters seems to work for all other field type I tried.

We should be able to pass multiple values using GET requests for such field type.

steffann:
### Fixes: #4525 

Implemented by adding `multiple_values` as a class property on `ScriptVariable` to determine which fields require a single value and which fields support multiple values and adding code to `ScriptForm.__init__` to provide corresponding `initial` values.

### DIFF
--- a/netbox/extras/forms.py
+++ b/netbox/extras/forms.py
@@ -1,6 +1,7 @@
 from django import forms
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
+from django.utils.datastructures import MultiValueDict
 from mptt.forms import TreeNodeMultipleChoiceField
 from taggit.forms import TagField as TagField_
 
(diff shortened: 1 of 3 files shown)


jeremystretch:
Ended up addressing this deeper under the hood in #4692. Thank you for getting it it started though!

Issue #4619, opened 2020-05-11
Cable traces do not display `via` anymore


hellerve:
# Environment
* Python version: 3.7 (should be on any)
* NetBox version: 2.8

### Steps to Reproduce
1. Disable any installed plugins by commenting out the `PLUGINS` setting in
   `configuration.py`.
2. Add two devices with interfaces, let’s call them A and B.
3. Create a circuit with two terminations, one on the A side and one on the Z side.
4. Connect interfaces on device A to the A side of the circuit, and device B on the Z side.

### Expected Behavior

Historically, the circuit termination was the connected endpoint, meaning that you were able to see the property `via`.

Here’s an interface connected before the update:

<img width="1408" alt="Bildschirmfoto 2020-05-11 um 12 51 32" src="https://user-images.githubusercontent.com/7725188/81554317-dc59e100-9386-11ea-9f58-46962b196396.png">

This is the trace:

<img width="1432" alt="Bildschirmfoto 2020-05-11 um 12 52 02" src="https://user-images.githubusercontent.com/7725188/81554362-f1cf0b00-9386-11ea-964d-dae4569d87ad.png">

### Observed Behavior

Due to the new tracing logic, circuit terminations gets resolved in there, meaning that the template never ends up in the code path that renders the `via`,  and only interfaces are the connected endpoints.

Here’s an interface connected after the update:

<img width="1428" alt="Bildschirmfoto 2020-05-11 um 12 52 11" src="https://user-images.githubusercontent.com/7725188/81554475-25aa3080-9387-11ea-9eeb-5a83d7e5cd3e.png">

Here’s the trace:

<img width="1440" alt="Bildschirmfoto 2020-05-11 um 12 52 17" src="https://user-images.githubusercontent.com/7725188/81554486-2c38a800-9387-11ea-89a9-3cff35c252cb.png">

This behavior change is a bit tricky to resolve, since it involves two parts of the Netbox code: the signal that handles cables and connected endpoints in `netbox.dcim.signals:update_connected_endpoints`, which after a bit of indirection calls `trace`, and the template `netbox/templates/dcim/inc/interface`, specifically the part that triggers on  `iface.connected_endpoint.term_side`.

I hope that this is understandable. I’d be happy to provide any additional info that you might need.

Cheers

jeremystretch:
> Create a cable from an interface to another, through a circuit.

Please provide the exact steps you are taking to reproduce this issue.

hellerve:
[...]

Issue #4604, opened 2020-05-07
IntegrityError when connecting multiple interfaces to one via a multi-position rear port


jeremystretch:
### Environment
* Python version: 3.6.9
* NetBox version: 2.8.3

### Steps to Reproduce
1. Disable any installed plugins by commenting out the `PLUGINS` setting in
   `configuration.py`.
2. Create or locate three devices (A, B, and C) each with at least one non-connected interface.
3. Create a device with two front ports mapped to a single rear port.
4. Connect interface A to the rear port.
5. Connect interface B to the first front port.
6. Connect interface C to the second rear port.

### Expected Behavior
The third cable should be created.

### Observed Behavior
An IntegrityError exception is raised:

```
Traceback (most recent call last):
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)

The above exception (duplicate key value violates unique constraint "dcim_interface__connected_interface_id_key"
DETAIL:  Key (_connected_interface_id)=(18403) already exists.
) was the direct cause of the following exception:
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/jstretch/netbox/netbox/dcim/views.py", line 2101, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/contrib/auth/mixins.py", line 85, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/home/jstretch/.virtualenvs/netbox/lib/python3.6/site-packages/django/views/generic/base.py", line 97, in dispatch
    return handler(request, *args, **kwargs)
[...]

PR #4706, opened 2020-06-02
4708: more flexible checks on RearPort usage

### LINKED ISSUE #4708: ValidationError when connecting multiple interfaces to one via a multi-position rear port
### Environment
* Python version: 3.7.3
* NetBox version: 2.8.5

### Steps to Reproduce
   `configuration.py`.
2. Create three devices (A, B, and C) each with at least one non-connected interface.
3. Create a device (X) with two front ports mapped to a single rear port with 2 positions.
4. Connect interface A to the first front port of X.
5. Connect interface B to the second front port of X.
6. Create a device (Y) with one rear port with one position mapped to a single front port
7. Connect interface C to the rear port of Y.
8. Connect the rear port of X to the front port of Y

Alternative that should produce the same result:
7. Connect interface C to the front port of Y.
8. Connect the rear port of X to the rear port of Y

### Expected Behaviour
No errors expected. When this would succeed, I would expect the connected endpoints of all three devices to be empty, as there is no single endpoint for the interface on device C.

### Observed Behavior
In step 8 you get ValidationError "Rear ports with multiple positions may only be connected to other rear ports". 

In the alternative version the ValidationError is "RP has 2 position(s) but RP has 1. Both terminations must have the same number of positions.".

Both are not expected, as the 1-on-1 RearPort/FrontPort is basically a cable extender.

steffann:
### Fixes: #4708

Alternative solution to #4708 that allows users to use RearPort in more complex ways without causing an IntegrityError with connected endpoints.

This implementation changes `trace()` to return the `position_stack` to the caller, which allows the caller to determine whether the returned path is end-to-end (nothing left on the stack) or whether it went through a FrontPort->RearPort without going through the corresponding RearPort->FrontPort (stack is not empty).

Adapted the `update_connected_endpoints` signal handler to consider an endpoint unconnected when the stack is not empty. This prevents the IntegrityError from occurring as connected endpoints are now only established for one-to-one paths.

### DIFF
--- a/netbox/circuits/models.py
+++ b/netbox/circuits/models.py
@@ -300,6 +300,9 @@ class CircuitTermination(CableTermination):
         blank=True
     )
 
+    # Paths do not end on cable terminations, they continue at the other end of the circuit
+    is_path_endpoint = False
+
     class Meta:
         ordering = ['circuit', 'term_side']
         unique_together = ['circuit', 'term_side']
--- a/netbox/circuits/models.py
+++ b/netbox/circuits/models.py
@@ -303,6 +303,9 @@ class CircuitTermination(CableTermination):
     # Paths do not end on cable terminations, they continue at the other end of the circuit
     is_path_endpoint = False
 
+    # But they are a possible connected endpoint
+    is_connected_endpoint = True
+
     class Meta:
         ordering = ['circuit', 'term_side']
         unique_together = ['circuit', 'term_side']
--- a/netbox/circuits/models.py
+++ b/netbox/circuits/models.py
@@ -300,9 +300,6 @@ class CircuitTermination(CableTermination):
         blank=True
     )
 
-    # Paths do not end on cable terminations, they continue at the other end of the circuit
-    is_path_endpoint = False
-
     # But they are a possible connected endpoint
     is_connected_endpoint = True
 
--- a/netbox/circuits/models.py
+++ b/netbox/circuits/models.py
@@ -300,9 +300,6 @@ class CircuitTermination(CableTermination):
         blank=True
     )
 
-    # But they are a possible connected endpoint
-    is_connected_endpoint = True
-
     class Meta:
         ordering = ['circuit', 'term_side']
         unique_together = ['circuit', 'term_side']
--- a/netbox/dcim/models/__init__.py
+++ b/netbox/dcim/models/__init__.py
@@ -2191,17 +2191,18 @@ class Cable(ChangeLoggedModel):
                 f"Incompatible termination types: {self.termination_a_type} and {self.termination_b_type}"
             )
 
-        # A RearPort with multiple positions must be connected to a RearPort with an equal number of positions
+        # A RearPort with multiple positions must be connected to a RearPort with an equal number of positions, or a
+        # FrontPort
         for term_a, term_b in [
             (self.termination_a, self.termination_b),
             (self.termination_b, self.termination_a)
         ]:
             if isinstance(term_a, RearPort) and term_a.positions > 1:
-                if not isinstance(term_b, RearPort):
+                if not isinstance(term_b, (FrontPort, RearPort)):
                     raise ValidationError(
-                        "Rear ports with multiple positions may only be connected to other rear ports"
+                        "Rear ports with multiple positions may only be connected to other pass-through ports"
                     )
-                elif term_a.positions != term_b.positions:
+                if isinstance(term_b, RearPort) and term_a.positions != term_b.positions:
                     raise ValidationError(
                         f"{term_a} has {term_a.positions} position(s) but {term_b} has {term_b.positions}. "
                         f"Both terminations must have the same number of positions."
--- a/netbox/dcim/models/__init__.py
+++ b/netbox/dcim/models/__init__.py
@@ -2191,20 +2191,20 @@ class Cable(ChangeLoggedModel):
                 f"Incompatible termination types: {self.termination_a_type} and {self.termination_b_type}"
             )
 
-        # A RearPort with multiple positions must be connected to a RearPort with an equal number of positions, or a
-        # FrontPort
+        # Check that a RearPort isn't connected to something silly
         for term_a, term_b in [
             (self.termination_a, self.termination_b),
             (self.termination_b, self.termination_a)
         ]:
             if isinstance(term_a, RearPort) and term_a.positions > 1:
-                if not isinstance(term_b, (FrontPort, RearPort)):
+                if term_b.is_path_endpoint:
                     raise ValidationError(
                         "Rear ports with multiple positions may only be connected to other pass-through ports"
                     )
-                if isinstance(term_b, RearPort) and term_a.positions != term_b.positions:
+                if isinstance(term_b, RearPort) and term_b.positions > 1 and term_a.positions != term_b.positions:
                     raise ValidationError(
-                        f"{term_a} has {term_a.positions} position(s) but {term_b} has {term_b.positions}. "
+                        f"{term_a} of {term_a.device} has {term_a.positions} position(s) but "
+                        f"{term_b} of {term_b.device} has {term_b.positions}. "
                         f"Both terminations must have the same number of positions."
                     )
 
--- a/netbox/dcim/models/__init__.py
+++ b/netbox/dcim/models/__init__.py
@@ -2191,7 +2191,8 @@ class Cable(ChangeLoggedModel):
                 f"Incompatible termination types: {self.termination_a_type} and {self.termination_b_type}"
             )
 
-        # Check that a RearPort isn't connected to something silly
+        # Check that a RearPort with multiple positions isn't connected to an endpoint
+        # or a RearPort with a different number of positions.
         for term_a, term_b in [
             (self.termination_a, self.termination_b),
             (self.termination_b, self.termination_a)
--- a/netbox/dcim/models/__init__.py
+++ b/netbox/dcim/models/__init__.py
@@ -2129,6 +2129,7 @@ class Cable(ChangeLoggedModel):
         return reverse('dcim:cable', args=[self.pk])
 
     def clean(self):
+        from circuits.models import CircuitTermination
 
         # Validate that termination A exists
         if not hasattr(self, 'termination_a_type'):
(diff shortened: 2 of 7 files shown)


steffann:
Fixing build errors…

steffann:
I realised that this fix only works after merging `4591-single-position-rearports`, so rebased on that branch.

jeremystretch:
Please open a new issue before submitting a PR. #4604 has been closed.

netbox/dcim/tests/test_models.py
```diff
@@ -549,24 +604,27 @@ def test_direct_connection(self):
         self.assertIsNone(endpoint_a.connection_status)
         self.assertIsNone(endpoint_b.connection_status)
 
-    def test_connection_via_single_rear_port(self):
```
steffann, line 552: The test in `test_connection_via_single_rear_port` should fail because Panel 1's rear port has multiple positions. This is already testen in `test_singlepos_rearport_connections` and `test_multipos_rearport_connections` above.

netbox/templates/dcim/cable_trace.html
```diff
@@ -88,6 +88,10 @@ <h4 class="text-muted">No Cable</h4>
                     </table>
                 </div>
             </div>
+        {% elif position_stack %}
+            <div class="col-md-11 col-md-offset-1">
+                <h3 class="text-warning text-center">Multiple possible paths end at this point. No connection established!</h3>
+            </div>
```
steffann, line 94: When tracing from FrontPort to RearPort there may be an endpoint, but it can/will be the endpoint from multiple traces. It is therefore not considered a connected endpoint, so explain this to the user.

netbox/dcim/tests/test_models.py
jeremystretch: The fact that this test exists implies that this is a supported topology. We can't simply change it without explicit documentation that this connection is no longer going to be supported.

netbox/dcim/tests/test_models.py
```diff
@@ -592,6 +650,171 @@ def test_connection_via_single_rear_port(self):
         self.assertIsNone(endpoint_a.connection_status)
         self.assertIsNone(endpoint_b.connection_status)
 
+        # Recreate cable 1 to test creating the cables in reverse order (RP first, FP second)
```
jeremystretch, line 653: This needs to be moved to its own test (if it's even needed).

netbox/dcim/tests/test_models.py
```diff
@@ -549,24 +604,27 @@ def test_direct_connection(self):
         self.assertIsNone(endpoint_a.connection_status)
         self.assertIsNone(endpoint_b.connection_status)
 
-    def test_connection_via_single_rear_port(self):
+    def test_connection_via_one_on_one_port(self):
```
jeremystretch, line 607: There is no such thing as a "one-on-one" port. The test method name should describe exactly what's being tested to avoid confusion.

netbox/dcim/tests/test_models.py
```diff
@@ -439,6 +453,37 @@ def test_cable_cannot_terminate_to_an_existing_connection(self):
         with self.assertRaises(ValidationError):
             cable.clean()
 
+    def test_singlepos_rearport_connections(self):
```
jeremystretch, line 456: Needs a topology diagram like the other tests. Name should be `test_connection_via_single_position_rearport`.

netbox/dcim/tests/test_models.py
```diff
@@ -439,6 +453,37 @@ def test_cable_cannot_terminate_to_an_existing_connection(self):
...
+        A RearPort with one position can be connected to anything as it is just a
+        cable extender.
+        """
+        # Connecting a single-position RearPort to a multi-position RearPort is ok
+        Cable(termination_a=self.rear_port1, termination_b=self.rear_port2).full_clean()
+
+        # Connecting a single-position RearPort to an Interface is ok
+        Cable(termination_a=self.rear_port1, termination_b=self.interface3).full_clean()
+
+    def test_multipos_rearport_connections(self):
```
jeremystretch, line 467: Needs a topology diagram like the other tests. Name should be `test_connection_via_multi_position_rearport`.

netbox/dcim/tests/test_models.py
```diff
@@ -439,6 +453,37 @@ def test_cable_cannot_terminate_to_an_existing_connection(self):
         with self.assertRaises(ValidationError):
             cable.clean()
 
+    def test_singlepos_rearport_connections(self):
+        """
+        A RearPort with one position can be connected to anything as it is just a
+        cable extender.
```
jeremystretch, line 459: No such thing as a "cable extender." Be sure to explain _exactly_ what the test is doing and why it's there.

netbox/dcim/models/__init__.py
```diff
@@ -2191,19 +2191,20 @@ def clean(self):
                 f"Incompatible termination types: {self.termination_a_type} and {self.termination_b_type}"
             )
 
-        # A RearPort with multiple positions must be connected to a RearPort with an equal number of positions
+        # Check that a RearPort isn't connected to something silly
```
jeremystretch, line 2194: What is "something silly?" Be explicit.

netbox/dcim/models/device_components.py
```diff
@@ -115,26 +117,32 @@ def get_peer_port(termination):
 
             # Map a front port to its corresponding rear port
             if isinstance(termination, FrontPort):
-                position_stack.append(termination.rear_port_position)
                 # Retrieve the corresponding RearPort from database to ensure we have an up-to-date instance
                 peer_port = RearPort.objects.get(pk=termination.rear_port.pk)
+
+                # Don't use the stack for 1-on-1 ports, they don't have to come in pairs
```
jeremystretch, line 123: What does it mean to "come in pairs?"

netbox/dcim/models/device_components.py
```diff
@@ -115,26 +117,32 @@ def get_peer_port(termination):
...
+
+                    position = position_stack.pop()
+
+                    # Validate the position
+                    if position not in range(1, termination.positions + 1):
+                        raise Exception("Invalid position for {} ({} positions): {})".format(
+                            termination, termination.positions, position
+                        ))
+                else:
+                    # Don't use the stack for 1-on-1 ports, they don't have to come in pairs
```
jeremystretch, line 144: Same as above.

netbox/dcim/tests/test_models.py
```diff
@@ -592,6 +650,171 @@ def test_connection_via_single_rear_port(self):
...
+        endpoint_a.refresh_from_db()
+        endpoint_b.refresh_from_db()
+
+        # Check that connections have been nullified
+        self.assertIsNone(endpoint_a.connected_endpoint)
+        self.assertIsNone(endpoint_b.connected_endpoint)
+        self.assertIsNone(endpoint_a.connection_status)
+        self.assertIsNone(endpoint_b.connection_status)
+
+    def test_connection_via_nested_one_on_one_port(self):
```
jeremystretch, line 684: `test_connections_via_nested_single_position_rearport`

netbox/dcim/tests/test_models.py
```diff
@@ -592,6 +650,171 @@ def test_connection_via_single_rear_port(self):
...
+        self.assertIsNone(endpoint_a.connected_endpoint)
+        self.assertIsNone(endpoint_b.connected_endpoint)
+        self.assertIsNone(endpoint_c.connected_endpoint)
+        self.assertIsNone(endpoint_d.connected_endpoint)
+        self.assertIsNone(endpoint_a.connection_status)
+        self.assertIsNone(endpoint_b.connection_status)
+        self.assertIsNone(endpoint_c.connection_status)
+        self.assertIsNone(endpoint_d.connection_status)
+
+        # Recreate cable 3 to test reverse order (Panel 5 FP first, RP second)
```
jeremystretch, line 775: Break into separate test if needed.

netbox/templates/dcim/cable_trace.html
jeremystretch: As with `split_ends`, the user should be presented options at this point. How do we convey the possible next paths?

jeremystretch (changes requested):
Try to keep in mind when writing comments, test names, etc., the reader is likely operating with far less context than you have. Be very explicit in what you're doing and why. Avoid resorting to shorthand (e.g. "one-on-ports") that can't be referenced in the documentation.

netbox/templates/dcim/cable_trace.html
steffann: If there is something left on `position_stack` that indicates that there are no possible next paths. There is no choice to be made. We can only notify the user that the path ends sooner than expected.

netbox/dcim/models/device_components.py
steffann: I have struggled to describe this succinctly (and apparently failed). The pairs I meant are:

`[FPx] [RP] --some-path-- [RP][FPx]` where multiplexing multiple FPs onto one RP needs to demux them again through an RP to an FP.  What would be better wording?

netbox/dcim/models/__init__.py
steffann: Sorry, should have added to TODO marker for myself there :)
Will fix!

netbox/dcim/tests/test_models.py
steffann: This will fail on this test because Panel 1 has 4 positions:
https://github.com/netbox-community/netbox/blob/develop/netbox/dcim/models/__init__.py#L2199-L2200

But this test never caught it because `full_clean()` was never called.

DanSheps:
Looks good, I want to do a pull on this later today and dive into everything more though.

DanSheps:
Tested, looks like it might resolve #4619 on merge as well

DanSheps (approved):

netbox/circuits/models.py
```diff
@@ -300,6 +300,12 @@ class CircuitTermination(CableTermination):
         blank=True
     )
 
+    # Paths do not end on cable terminations, they continue at the other end of the circuit
+    is_path_endpoint = False
```
jeremystretch, line 304: Let's avoid setting arbitrary flags on model classes. The values aren't going to change among instances, so there's no need to store it on the class.

Instead, we can create e.g. `PATH_ENDPOINT_MODELS` and `CONNECTION_ENDPOINT_MODELS` constants, and check for `type(instance) in constants.PATH_ENDPOINT_MODELS`.

jeremystretch (changes requested):

PR #4713, opened 2020-06-03
Closes #4348: Introduce LDAPBackend

### LINKED ISSUE #4348: Move LDAP authentication into a standalone authentication backend
### Proposed Changes
With the introduction of support for custom authentication backends (#2328) in NetBox v2.8, it no longer makes sense to maintain LDAP authentication as part of NetBox's core.

### Justification
Offloading this functionality removes the burden of maintenance from the core maintainers, encourages more community participation in the development of the LDAP backend itself, and should serve as a strong reference implementation for custom authentication backends. It will also help guide the development of remote authentication support in NetBox.

jeremystretch:
### Closes: #4348

- Remove `AUTH_LDAP_*` configuration parameters from `settings.py`
- Introduce LDAPBackend to wrap django-auth-ldap's backend and read configuration parameters directly from `ldap_config.py`
- Update installation and configuration docs


lampwins:
What do you think of renaming `ldap_config.py` to `remote_auth_config.py`? I think other remote auth classes can use this same pattern when they need user-defined configuration options, and this keeps things consistent within the ecosystem.

jeremystretch:
@lampwins My initial thought was to minimize the disruption for current LDAP users upgrading to v2.9. I'm not opposed to renaming it, though I wonder if it might be less disruptive to use a unique filename per backend so that evaluating different backends doesn't require modifying any config files. E.g. FooBackend and BarBackend would use `foo_config.py` and `bar_config.py` respectively.

lampwins:
@jeremystretch all good points. The one remaining thing to call out in favor of a single file is the easier ability to add it to gitignore to help reduce accidental contributions, but that is likely not a good enough sole reason to do it. Also, there is nothing stopping us from doing this down the road if we come up with a more compelling reason.

PR #4718, opened 2020-06-04
Fixes #4674 - Fix available-ips and available-prefixes swagger definitions

### LINKED ISSUE #4674: API Definitions for available-ips and available-prefixes are incorrect
### Change Type
[ ] Addition
[x] Correction
[ ] Deprecation

### Area
[ ] Installation instructions
[ ] Configuration parameters
[ ] Functionality/features
[x] REST API
[ ] Administration/development
[ ] Other

### Proposed Changes
I'm writing a ipam project with go-netbox(https://github.com/kobayashi/go-netbox for v2.8), which generate the client from swagger. And I'm using the prefix/available-ips api to create unused ip address. I notice that the api documents in http://netbox/api/docs/ has difference with actually code:
* The request model should be IPAddress rather than WritablePrefix
  _ref: https://github.com/netbox-community/netbox/blob/v2.8.3/netbox/ipam/api/views.py#L211_
* The return code should be 201 rather than 200
  _ref: https://github.com/netbox-community/netbox/blob/v2.8.3/netbox/ipam/api/views.py#L218_

These mismatches make go-netbox client unable to work correctly. I would appreciate if this can be fixed.

DanSheps:
### Fixes: #4674 

Modifies the swagger definitions for the post and get methods to provide proper return codes and body serializer for post methods.

### DIFF
--- a/netbox/ipam/api/views.py
+++ b/netbox/ipam/api/views.py
@@ -75,7 +75,7 @@ class PrefixViewSet(CustomFieldModelViewSet):
     filterset_class = filters.PrefixFilterSet
 
     @swagger_auto_schema(method='get', responses={200: serializers.AvailablePrefixSerializer(many=True)})
-    @swagger_auto_schema(method='post', responses={201: serializers.AvailablePrefixSerializer(many=False)})
+    @swagger_auto_schema(method='post', responses={201: serializers.AvailablePrefixSerializer(many=True)})
     @action(detail=True, url_path='available-prefixes', methods=['get', 'post'])
     @advisory_lock(ADVISORY_LOCK_KEYS['available-prefixes'])
     def available_prefixes(self, request, pk=None):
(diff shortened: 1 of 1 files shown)


netbox/ipam/api/views.py
```diff
@@ -155,7 +155,7 @@ def available_prefixes(self, request, pk=None):
             return Response(serializer.data)
 
     @swagger_auto_schema(method='get', responses={200: serializers.AvailableIPSerializer(many=True)})
-    @swagger_auto_schema(method='post', responses={201: serializers.AvailableIPSerializer(many=False)},
+    @swagger_auto_schema(method='post', responses={201: serializers.AvailableIPSerializer(many=True)},
                          request_body=serializers.AvailableIPSerializer(many=False))
```
DanSheps, line 159: The many=False is required due to the way the body schema generation works as it cannot handle a list serializer.

jeremystretch: Why does `available_ips()` require specifying `request_body` but `available_prefixes()` does not?

DanSheps: available_prefixes already uses the prefix serializer for the body.  With available_ips, unless you specify AvailableIPSerializer, it defaults to WriteablePrefix.  We can do the Prefix serializer too, if we want to be explicit, but it is redundant here I think

jeremystretch (approved):

PR #4797, opened 2020-06-26
Fixes #4695 - Add Metadata class that returns content type choices

### LINKED ISSUE #4695: Expose choices of cable:termination_a_type and cable:termination_b_type
### Environment
* Python version:  3.8.3
* NetBox version:  2.8.4

### Proposed Functionality
In older Netbox versions it was possible to get choices for cable:termination_a_type via this api-request:
```
GET '/api/dcim/_choices/'
...
"cable:termination_a_type": [
    {
      "value": "dcim.consoleport",
      "label": "console port"
    },
    {
      "value": "dcim.consoleserverport",
      "label": "console server port"
    },
    {
      "value": "dcim.interface",
      "label": "interface"
    },
...
```

But since #3416 it is not possible to get the choices for termination_a_type with the new OPTIONS request, as there are no choices listed:
```
OPTIONS '/api/dcim/cables/'
...
"termination_a_type": {
  "type": "field",
  "required": true,
  "read_only": false,
  "label": "Termination a type"
},
...
```

### Use Case
API users/developers would know which terminations exist.

### Database Changes
None

### External Dependencies
None

DanSheps:
### Fixes: #4695 

Introduce a ~~LimitedMetadata~~ContentTypeMetadata class that will return ContentType choices

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -29,6 +29,7 @@ from utilities.api import (
     get_serializer_for_model, IsAuthenticatedOrLoginNotRequired, ModelViewSet, ServiceUnavailable,
 )
 from utilities.utils import get_subquery
+from utilities.metadata import LimitedMetaData
 from virtualization.models import VirtualMachine
 from . import serializers
 from .exceptions import MissingFilterException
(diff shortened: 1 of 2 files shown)


steffann (approved):

jeremystretch (changes requested):
Looks good. Could you also add it to the other serializers which utilize `ContentTypeField` please?

PR #4813, opened 2020-07-01
Don't ignore ImportErrors raised when loading a plugin. Fixes #4805

### LINKED ISSUE #4805: Improve handling of plugin loading errors
### Environment
* Python version:  3.7.7
* NetBox version:  v2.8.4

### Proposed Functionality

Currently, when developing a plugin, errors in the plugin code under development can cause undesirable NetBox behavior. If the plugin code is severely broken (such as a `SyntaxError`, then NetBox as a whole will die when attempting to load the plugin. In other cases, such as where the plugin code throws an `ImportError` due to a missing or incorrectly specified dependency, NetBox will silently ignore the error, the admin "Installed Plugins" page will still show the broken plugin, but attempts to access the plugin later (via its URLs, views, etc.) will report a confusing error such as `NoReverseMatch at /
'plugin_name' is not a registered namespace inside 'plugins'`.

See also #4628.

Currently NetBox loads plugins via three primary functions: `importlib.import_module()`, `django.apps.apps.get_app_config()` and `django.utils.module_loading.import_string()`. 

#### importlib.import_module()

This is called in `netbox/netbox/settings.py`:

```python
for plugin_name in PLUGINS:

    # Import plugin module
    try:
        plugin = importlib.import_module(plugin_name)
    except ImportError:
        raise ImproperlyConfigured(
            "Unable to import plugin {}: Module not found. Check that the plugin module has been installed within the "
            "correct Python environment.".format(plugin_name)
        )
```

In the case of an `ImportError`, this should re-raise an `ImproperlyConfigured` error, but in practice it appears that Django is silently ignoring that exception. (Possible NetBox bug of some sort?) This then causes errors downstream when `get_app_config` is called - see below.

In the case of any other exception being thrown by the plugin, the entire Django startup process aborts. 

*Proposal*: add a blanket `except Exception` to this `try` block, and log the exception and continue without the plugin, so that NetBox itself remains operational.

#### django.apps.apps.get_app_config()

This is called in `netbox/extras/plugins/urls.py` and `netbox/extras/plugins/views.py`, and will raise a `LookupError` if the requested plugin is not present (such as due to an earlier error in `settings.py` calling `import_module()` above), which is not handled by NetBox, causing the entire application to abort. 

*Proposal*: catch the LookupError in each case, print an appropriate error message to the Django console, and skip over the affected plugin. This will fix the case where broken plugins are still listed under the admin "Installed Plugins" view and the `api/plugins/installed-plugins/` API view. 

*Proposal*: A possible further enhancement would be for these two views to, rather than simply skipping broken plugins, separately report any plugins that are configured but could not be loaded successfully.

#### django.utils.module_loading.import_string()

This method raises an `ImportError` if:
1. the specified string is not a valid dotted path specification (`module.submodule.object`)
2. importing the specified module/submodule raises an `ImportError`, because either:
    a. it does not exist
    b. it exists but fails to load, perhaps because it itself contains an `import` statement that fails (the case described in this issue)
3. the specified module/submodule exists but does not contain the specified named object.

In the case of plugin loading where we are using `import_string` (`extras/plugins/__init__.py`, `extras/plugins/views.py`, `extras/plugins/urls.py`) we avoid case 1. by properly constructing the string ourselves. We want case 2.a. (and 3.?) to fail silently because, for example, a plugin very well may not implement a `plugin_name.urls` module (or have a `urlpatterns` declaration in said module?). 

*Proposal*: Case 2.b. should result in a reported error message, and possibly also this failure to be reported in some fashion under the /admin/plugins/installed-plugins/ view. This could probably be accomplished by rolling our own implementation of `import_string`, which is a relatively simple function.

### Use Cases

1. Plugin developers will be more readily able to diagnose and repair errors in the plugin code.
2. Deployments of NetBox will be more resilient to errors in plugin code that are outside NetBox's direct control.

### Database Changes

None are proposed.

### External Dependencies

None identified.

glennmatthews:
### Fixes: #4805

Django's `import_string` function is convenient to use, but it's difficult to distinguish between an `ImportError` raised by this function (because the string describes a module/object that does not exist) versus an `ImportError` raised by incorrect code in the module being imported. When NetBox is loading a plugin, the former should be ignored (as a plugin is not required to implement all possible entry points supported by the plugins API) but we are incorrectly ignoring the latter as well (which represent a real bug in the plugin code, and should not be ignored).

Solution: move away from the use of `import_string` to use more granular functions from `importlib` that let us distinguish between "module does not exist" and "module exists but throws an `ImportError` while being loaded".

Example of a plugin error now being correctly raised rather than caught:

```
netbox_1    | Exception in thread django-main-thread:
netbox_1    | Traceback (most recent call last):
netbox_1    |   File "/usr/local/lib/python3.7/threading.py", line 926, in _bootstrap_inner
netbox_1    |     self.run()
netbox_1    |   File "/usr/local/lib/python3.7/threading.py", line 870, in run
netbox_1    |     self._target(*self._args, **self._kwargs)
netbox_1    |   File "/usr/local/lib/python3.7/site-packages/django/utils/autoreload.py", line 53, in wrapper
netbox_1    |     fn(*args, **kwargs)
netbox_1    |   File "/usr/local/lib/python3.7/site-packages/django/core/management/commands/runserver.py", line 109, in inner_run
netbox_1    |     autoreload.raise_last_exception()
netbox_1    |   File "/usr/local/lib/python3.7/site-packages/django/utils/autoreload.py", line 76, in raise_last_exception
netbox_1    |     raise _exception[1]
netbox_1    |   File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 357, in execute
netbox_1    |     autoreload.check_errors(django.setup)()
netbox_1    |   File "/usr/local/lib/python3.7/site-packages/django/utils/autoreload.py", line 53, in wrapper
netbox_1    |     fn(*args, **kwargs)
netbox_1    |   File "/usr/local/lib/python3.7/site-packages/django/__init__.py", line 24, in setup
netbox_1    |     apps.populate(settings.INSTALLED_APPS)
netbox_1    |   File "/usr/local/lib/python3.7/site-packages/django/apps/registry.py", line 122, in populate
netbox_1    |     app_config.ready()
netbox_1    |   File "/opt/netbox/netbox/extras/plugins/__init__.py", line 77, in ready
netbox_1    |     spec.loader.exec_module(navigation)
netbox_1    |   File "<frozen importlib._bootstrap_external>", line 728, in exec_module
netbox_1    |   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
netbox_1    |   File "/plugins/myplugin/navigation.py", line 6, in <module>
netbox_1    |     import foo
netbox_1    | ModuleNotFoundError: No module named 'foo'
```

### DIFF
--- a/netbox/extras/plugins/__init__.py
+++ b/netbox/extras/plugins/__init__.py
@@ -1,7 +1,5 @@
 import collections
-import importlib
 import inspect
-import sys
 from packaging import version
 
 from django.apps import AppConfig
@@ -11,6 +9,8 @@ from django.template.loader import get_template
 
 from extras.registry import registry
 from utilities.choices import ButtonColorChoices
+
+from extras.plugins.utils import import_object
 
 
 # Initialize plugin registry stores
@@ -61,26 +61,14 @@ class PluginConfig(AppConfig):
     def ready(self):
 
         # Register template content
-        module, attr = f"{self.__module__}.{self.template_extensions}".rsplit('.', 1)
-        spec = importlib.util.find_spec(module)
-        if spec is not None:
-            template_content = importlib.util.module_from_spec(spec)
-            sys.modules[module] = template_content
-            spec.loader.exec_module(template_content)
-            if hasattr(template_content, attr):
-                template_extensions = getattr(template_content, attr)
-                register_template_extensions(template_extensions)
+        template_extensions = import_object(f"{self.__module__}.{self.template_extensions}")
+        if template_extensions is not None:
+            register_template_extensions(template_extensions)
 
         # Register navigation menu items (if defined)
-        module, attr = f"{self.__module__}.{self.menu_items}".rsplit('.', 1)
-        spec = importlib.util.find_spec(module)
-        if spec is not None:
-            navigation = importlib.util.module_from_spec(spec)
-            sys.modules[module] = navigation
-            spec.loader.exec_module(navigation)
-            if hasattr(navigation, attr):
-                menu_items = getattr(navigation, attr)
-                register_menu_items(self.verbose_name, menu_items)
+        menu_items = import_object(f"{self.__module__}.{self.menu_items}")
+        if menu_items is not None:
+            register_menu_items(self.verbose_name, menu_items)
 
     @classmethod
     def validate(cls, user_config):
--- a/netbox/extras/plugins/__init__.py
+++ b/netbox/extras/plugins/__init__.py
@@ -1,12 +1,13 @@
 import collections
+import importlib
 import inspect
+import sys
 from packaging import version
 
 from django.apps import AppConfig
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.template.loader import get_template
-from django.utils.module_loading import import_string
 
 from extras.registry import registry
 from utilities.choices import ButtonColorChoices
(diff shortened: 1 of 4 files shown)


lampwins:
I am fine with this approach, but it is a lot of repeated logic/code. @glennmatthews do you think we can sanely refactor out a utility function that manages looking for a spec and manages the import?

glennmatthews:
That's a fair point. I'll see what I can do.

lampwins (approved):
LGTM

PR #4844, opened 2020-07-10
Adds name lookup to NAPALM if no primary IP address exists for device

### LINKED ISSUE #4831: NAPALM - Support connecting via Hostname
### Environment
* Python version:  3.7.7
* NetBox version:  2.8.6

### Proposed Functionality

Support device name NAPALM connectivity when no primary IP address is defined

### Use Case

- Some orgs do not use for IPAM but still would like NAPALM functionality
  - IP Addresses can change and if NetBox is not that source of truth, would be out of date
- Capability exists with minimal additional code to complete a DNS lookup using socket
- Would require name to be reachable by name service within the server NetBox is using

### Database Changes

No changes

### External Dependencies

- Python socket library, built in
- NetBox server ability to query appropriate name server (not a NetBox dependency)

jvanderaa:
### Fixes: #4831

- For NAPALM connection, uses Python socket library to attempt to resolve a name to IP address when a primary IP address is not defined.
- Name resolution must be done without adding any suffixes to the name such as: `rtr01` would be supplied back to name resolution. If `rtr01` does not resolve to anything on the NetBox host, no translation will be found.
  - If hosts exist in *.infra.example.com, but the host can only lookup against *.example.com, then the name resolution will fail
- Remaps IP address to use "host" for the name to indicate could be an IP address or could be a device name

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -1,3 +1,4 @@
+import socket
 from collections import OrderedDict
 
 from django.conf import settings
@@ -371,8 +372,6 @@ class DeviceViewSet(CustomFieldModelViewSet):
         Execute a NAPALM method on a Device
         """
         device = get_object_or_404(Device, pk=pk)
-        if not device.primary_ip:
-            raise ServiceUnavailable("This device does not have a primary IP address configured.")
         if device.platform is None:
             raise ServiceUnavailable("No platform is configured for this device.")
         if not device.platform.napalm_driver:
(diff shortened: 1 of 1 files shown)


netbox/dcim/api/views.py
```diff
@@ -371,8 +372,6 @@ def napalm(self, request, pk):
         Execute a NAPALM method on a Device
         """
         device = get_object_or_404(Device, pk=pk)
-        if not device.primary_ip:
```
jeremystretch, line 374: Device names are optional, so we still want to ensure that a device has a primary IP _or_ name defined.

netbox/dcim/api/views.py
```diff
@@ -402,7 +401,18 @@ def napalm(self, request, pk):
         # Connect to the device
```
jeremystretch, line 401: Minor nit, but this comment should be moved down to the driver initialization since we've inserted a new comment below it.

jeremystretch (changes requested):
Looks good! Just one minor nit about validation.

netbox/dcim/api/views.py
jvanderaa: 👍  Updated to have a check for device.name and raise exception if not found.

netbox/dcim/api/views.py
jvanderaa: Updated the comment location. I then also moved two other variable assignments to below the logic to assist in the aesthetics of the code and where the work is getting completed.

netbox/dcim/api/views.py
jeremystretch: We want to keep this check at the beginning of the method, both to keep it grouped with the other checks and to ensure that failures occur quickly. (There's no point in initializing the NAPALM driver only to then discover that we have no way of connecting to the device.)

PR #4847, opened 2020-07-13
CircuitTermination cleanup (Fixes #4812)

### LINKED ISSUE #4812: Clean up code for showing CircuitTermination on interfaces
This template code relies on `trace()` not traversing circuits:
https://github.com/netbox-community/netbox/blob/43d610405f625f2044ebdab0c21df01b75dc856d/netbox/templates/dcim/inc/interface.html#L94-L123

Since https://github.com/netbox-community/netbox/commit/e143158f121c7d147bec820af227388deca6dfb5 this is no longer the case, which makes the template code partially obsolete. This needs to be cleaned up.

# Example 1

Consider the following topology:

```
+--------+     +------+  +-----------+  +------+     +--------+
| Intf X +-----+ CT A +--+ Circuit C +--+ CT Z +-----+ Intf Y |
+--------+  1  +------+  +-----------+  +------+  2  +--------+
```

## Old implementation

Before https://github.com/netbox-community/netbox/commit/e143158f121c7d147bec820af227388deca6dfb5 the signal handler to update `connected_endpoint`s would call `trace(follow_circuits=False)`. This would cause the `connected_endpoint` for Interface X to be CircuitTermination A (and vice versa), and the `connected_endpoint` for Interface Y to be CircuitTermination Z (and vice versa).

When the template is showing the connected endpoint it will try to call `get_peer_termination` on it. Because the connected endpoint is a CircuitTermination this may succeed (depending on whether both CircuitTerminations exist for the Circuit, in this example it exists) and if it does the CircuitTermination on the other end of the Circuit will be used. If that CircuitTermination has a connected endpoint then that connected endpoint will be shown as the connected endpoint of the interface, and the circuit information will be displayed as "via …".

So in the example above, for Interface X, the old code would have CircuitTermination A stored as the connected endpoint in the database, but it would display Interface Y via Circuit C in the user interface.

## Current implementation

Since https://github.com/netbox-community/netbox/commit/e143158f121c7d147bec820af227388deca6dfb5 the `follow_circuits` parameter to `trace()` no longer exists, and the implementation always follows circuits. So now the `connected_endpoint` of Interface X will be Interface Y (and vice versa). The code that displays the connected endpoint of the peer of a circuit termination if the original connected endpoint is a circuit termination is therefore obsolete (and considering the length of that sentence probably rightfully so).

# Example 2

Consider the following topology:

```
+--------+     +------+  +-----------+  +------+
| Intf X +-----+ CT A +--+ Circuit C +--+ CT Z |
+--------+  1  +------+  +-----------+  +------+
```

## Old implementation

The old implementation would stop at CircuitTermination A and store that as the connected endpoint. The user interface would then find the other CircuitTermination through `get_peer_termination` and display that.

## Current implementation

Because circuits are now always traversed `trace` will now return these steps in the path:

| CableTermination     | Cable | CableTermination     |
|----------------------|-------|----------------------|
| Interface X          | 1     | CircuitTermination A |
| CircuitTermination Z | None  | None                 |

Because the last CableTermination is empty the `connected_endpoint` will be empty and the user interface will show "Not connected"

# Example 3

Consider the following topology:

```
+--------+     +------+  +-----------+
| Intf X +-----+ CT A +--+ Circuit C |
+--------+  1  +------+  +-----------+
```

Both the old and the new implementation handle this in the same way. What makes this a special case is that this is the only remaining situation where the `connected_endpoint` of CircuitTermination A is updated. When CircuitTermination Z is later added and connected to an interface the `connected_endpoint` of Interface X will be updated to reflect that but the `connected_endpoint` of CircuitTermination A will still be Interface X.

This behaviour is inconsistent because it only occurs when there is only one CircuitTermination connected to the Circuit. When both CircuitTerminations are created before connecting the cable the `connected_endpoint` of CableTermination A will always remain empty.

steffann:
### Fixes: #4812

I made the following choices:
- When an interface has a connected endpoint show it. Do not show "via …" anymore. The part of the template that handled that (no longer existing) situation has been removed. (see #4812 example 1)
- When a trace ends at the far end of a circuit store the near circuit termination. This was the "old" behaviour before the `follow_circuits` parameter to `trace()` was removed. This was useful to show the circuit and remote site in the user interface, so that has been restored. (see #4812 example 2)
- When a trace no longer leads anywhere make sure the connected endpoint is cleared (see #4812 example 3)

Each of these has been separately committed.


jeremystretch:
We should not have a management command for retracing cables en masse. If we need a one-time schema migration to handle these changes, that's fine, but the existence of a manually-invoked `retrace` command implies that the tracing logic is broken and requires user intervention. If that's the case, we should focus on fixing the root issue instead.

steffann:
Fair enough. I got the idea from renaturalize.

I'm sure there are some inconsistencies in the connected endpoints, with potential old dsta left over from before the recent fixes. Rerunning the current implementation makes sure everything is in a consistent state. So this should Indeed be a one-off action.

What is the best way to write a migration for that? I don't think we can use the trace method in a migration 😕  Copy and paste the current implementation into a migration so it can run on bare models?

jeremystretch:
Have a look at some of the recent schema migrations which call `migrationsRunPython()`: It allows for running custom code as part of a migration (with some exceptions). It should be possible to call `trace()` from there, just have to load the model first:

```python
def do_something_with_cables(apps, schema_editor):
    Cable = apps.get_model('dcim', 'Cable')
    for cable in Cable.objects.all():
        ...

class Migration(migrations.Migration):
    ...
    operations = [
        migrations.RunPython(
            code=do_something_with_cables,
            reverse_code=migrations.RunPython.noop
        ),
    ]
```

steffann:
IIRC apps.get_model gives back a class without the custom methods. But I guess my knowledge may be a bit outdated. Will figure it out!

steffann:
Working on this today. Apologies for the delay

jeremystretch:
Given that the entirety of the cabling logic in NetBox is being gutted and reimplemented under #4900 in v2.10, let's pass on these changes. I worry that merging them in v2.9 is going to introduce difficult conflicts, particularly the migration. I have taken all of the above points to heart in my work under #4900 however, and I believe v2.10 will address all of these concerns and more.

Issue #2006, opened 2018-04-04
Run reports/scripts in the background


arionl:
### Issue type
[X] Feature request 
[ ] Bug report      
[ ] Documentation   

### Environment
* Python version:  3.6.4
* NetBox version:  2.3.2

### Description
I have some reports that take a long time to run. They are usually reports that are comparing NetBox data with data in another system. For example, I'm using a custom field to track if a device or VM should be actively monitored and then I have a report that runs every day to see if the object exists or if it is missing from the monitoring platform. With thousands of objects, these reports take a while to run. They work fine when they run on the daily schedule or if I run them via `manage.py` but if I use the "Run Report" button within the UI, I usually get a timeout message from my nginx reverse proxy. I could increase the timeouts, but I think it would be more useful if a report could be marked as "Running" and then I could refresh the page later to check for results. Could this be managed as a DB flag or something?

Grokzen:
Would it not be better to have this type of task outsourced to a celery runner and run the report async to free up the call to the GUI and avoid having to increase the timeout?

arionl:
I'm not sure of the proper way to tackle this from within Django. For me, I'm running all reports via a cron job that just does `python3 manage.py runreport REPORTNAME` every day.. but in this feature request I was just pointing out that in many cases, the GUI will be unresponsive or timeout when a report takes a long time to run and it would be nice to allow my users to run reports on demand if they really don't want to wait until the next day when a report is run via cron. Not a high priority request, but I think others might have a similar use case and there may be other tasks lurking within NetBox that could be offloaded like this..

jeremystretch:
Beginning with v2.6, NetBox will require a Redis database, so I think it makes sense to also require `django-rq` to enable asynchronous processing of reports. (`django-rq` is currently an optional component, required only to enable webhooks.)

tb-killa:
Maybe this FR could be used by some adapted Fragments from https://github.com/netbox-community/netbox/issues/3415 ?

jeremystretch:
We need to figure out the optimal workflow when running reports via the REST API. If the request is queued for background processing, we need to return some kind of confirmation to the client.

It also raises the question of whether we should immediately delete the result from the previous run, or wait until the new run has completed.

lampwins:
[...]

PR #4852, opened 2020-07-14
Some additional fixes to scripts/reports background running (2.9)

glennmatthews:
### Fixes: #2006 

Follow-up to #4799. Fixes two issues:

1) `success` log result was misspelled as `sucess`, which was causing display issues in the result detail view (since `label-sucess` (sic) is not a defined Bootstrap style).
2) If a script was run with `commit=False`, its JobResult status would remain as `running` indefinitely even after the script execution had actually completed.

### DIFF
--- a/netbox/extras/choices.py
+++ b/netbox/extras/choices.py
@@ -127,7 +127,7 @@ class TemplateLanguageChoices(ChoiceSet):
 class LogLevelChoices(ChoiceSet):
 
     LOG_DEFAULT = 'default'
-    LOG_SUCCESS = 'sucess'
+    LOG_SUCCESS = 'success'
     LOG_INFO = 'info'
     LOG_WARNING = 'warning'
     LOG_FAILURE = 'failure'
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -415,8 +415,6 @@ def run_script(data, request, commit=True, *args, **kwargs):
     try:
         with transaction.atomic():
             script.output = script.run(**kwargs)
-            job_result.data = ScriptOutputSerializer(script).data
-            job_result.set_status(JobResultStatusChoices.STATUS_COMPLETED)
 
             if not commit:
                 raise AbortTransaction()
@@ -434,6 +432,10 @@ def run_script(data, request, commit=True, *args, **kwargs):
         job_result.set_status(JobResultStatusChoices.STATUS_ERRORED)
 
     finally:
+        if job_result.status != JobResultStatusChoices.STATUS_ERRORED:
+            job_result.data = ScriptOutputSerializer(script).data
+            job_result.set_status(JobResultStatusChoices.STATUS_COMPLETED)
+
         if not commit:
             # Delete all pending changelog entries
             purge_changelog.send(Script)


lampwins:
I am fine with this, but I wonder if we could refactor the script execution logic to avoid nested try blocks.

glennmatthews:
I'm open to suggestions on the refactoring. I *think* the following alternative would work but I personally think it's less readable:

```python
    try:
        with transaction.atomic():
            script.output = script.run(**kwargs)

            if not commit:
                raise AbortTransaction()

    except AbortTransaction:
        job_result.data = ScriptOutputSerializer(script).data
        job_result.set_status(JobResultStatusChoices.STATUS_COMPLETED)

    except Exception as e:
        stacktrace = traceback.format_exc()
        script.log_failure(
            "An exception occurred: `{}: {}`\n```\n{}\n```".format(type(e).__name__, e, stacktrace)
        )
        logger.error(f"Exception raised during script execution: {e}")
        commit = False
        job_result.set_status(JobResultStatusChoices.STATUS_ERRORED)

    else:
        job_result.data = ScriptOutputSerializer(script).data
        job_result.set_status(JobResultStatusChoices.STATUS_COMPLETED)

    finally:
        if not commit:
            # Delete all pending changelog entries
            purge_changelog.send(Script)
            script.log_info(
                "Database changes have been reverted automatically."
            )
```

glennmatthews:
Or a second option:

```python
    try:
        with transaction.atomic():
            script.output = script.run(**kwargs)

            if not commit:
                raise AbortTransaction()

    except AbortTransaction:
        pass

    except Exception as e:
        stacktrace = traceback.format_exc()
        script.log_failure(
            "An exception occurred: `{}: {}`\n```\n{}\n```".format(type(e).__name__, e, stacktrace)
        )
        logger.error(f"Exception raised during script execution: {e}")
        commit = False
        job_result.set_status(JobResultStatusChoices.STATUS_ERRORED)

    finally:
        if job_result.status != JobResultStatusChoices.STATUS_ERRORED:
            job_result.data = ScriptOutputSerializer(script).data
            job_result.set_status(JobResultStatusChoices.STATUS_COMPLETED)

        if not commit:
            # Delete all pending changelog entries
            purge_changelog.send(Script)
            script.log_info(
                "Database changes have been reverted automatically."
            )
```

jeremystretch:
I think [this approach](https://github.com/netbox-community/netbox/pull/4852#issuecomment-658391818) is fine. It might be cleaner to only set `STATUS_COMPLETED` if no status has been set, but either way.

glennmatthews:
The status is set to `STATUS_RUNNING` at the start of the function - I could change the `status != STATUS_ERRORED` check to `status == STATUS_RUNNING` if you think that would be cleaner?

jeremystretch:
Sorry, forgot about setting it to `RUNNING` first. What you have should be fine.

glennmatthews:
Weird Travis failure for Python 3.7, looks unrelated to my changes.

jeremystretch:
@glennmatthews yeah ignore those, something wonky is going on.

jeremystretch:
FYI tests are fixed

glennmatthews:
Is this blocked on "com[ing] up with a more elegant approach to intentionally rolling back a transaction"? Or can we move ahead with this as-is?

glennmatthews:
Looks like [`set_rollback`](https://docs.djangoproject.com/en/3.0/topics/db/transactions/#django.db.transaction.set_rollback) may be the approach to consider.

jeremystretch:
> Is this blocked on "com[ing] up with a more elegant approach to intentionally rolling back a transaction"?

No; that was just a general comment/complaint.

> Looks like set_rollback may be the approach to consider.

Yes! I knew there had to be some mechanism for this. Not sure how I missed this. You can either add this to the PR, or leave it as-is and I'll take it as an action item to implement application-wide.

Issue #3672, opened 2019-11-05
Device_count is not updated on device roles if a device is removed/add to the role


soer7022:
### Environment
* Python version:  v3.6.8
* NetBox version:  v2.6.5 

### Steps to Reproduce
1. Create a new device role
2. Add an existing device to it
3. Check the api response on `device_count`
4. Remove the device from it
5. Check the api response on `device_count`

### Expected Behavior
I expect the `device_count` to be correctly updated

### Observed Behavior
When adding an existing device to the role, the role's `device_count` stays at null.
When a new device is created and added to the role directly, the `device_count` is still null.

jeremystretch:
I'm not able to reproduce this using the development server alone. It smells like a caching problem.

jeremystretch:
This likely stems from the use of [subqueries](https://github.com/netbox-community/netbox/blob/5eb5c4bac551d9875a3688aad36388af8f61e3a6/netbox/utilities/utils.py#L74) to count the related devices and VMs. @lampwins you've done some work around this; can you take a look?

hSaria:
Ran into this issue with `/dcim/regions`. Is disabling caching on that count query an option or would you rather invalidate the count on updates of the children?

lampwins:
@jeremystretch this is related to caching and the use of subqueries. Basically cacheops just doesn't support them (see 7 [here](https://github.com/Suor/django-cacheops#caveats))

There are a couple of ways we can work around this but want to know your thoughts?

lampwins:
FYI I am looking into what it will take to solve this upstream in cacheops by supporting subqueries natively.

lampwins:
After looking into it, implementing support for subqueries upstream appears pretty involved and over my head, so I am now investigating doing this with annotations.

steffann:
Actually, the [comment on subqueries](https://github.com/Suor/django-cacheops#caveats) "one can just break queries into simpler ones, which cache better" made me think. Usually subqueries are used to make things faster by having fewer separate queries, but with cacheops having several separately cacheable queries may make more sense.

Something to test and play with…

steffann:
I tried removing the subqueries

```
class DeviceRoleViewSet(ModelViewSet):
    queryset = DeviceRole.objects.all()
    ...
```

And adding the counts as simple properties:

```python
class DeviceRole(ChangeLoggedModel):
    ...

    @property
    def device_count(self):
        return self.devices.count()

    @property
    def virtualmachine_count(self):
[...]

PR #4861, opened 2020-07-16
Upgrade to Django 3.1 (v2.9)

jeremystretch:
This is a draft PR to track the work necessary for upgrading to Django 3.1. Most issues have been addressed, however there remains a bug regarding queryset ordering when annotations are used. This was called out as [an intentional change in Django 2.2](https://docs.djangoproject.com/en/3.1/releases/2.2/#model-meta-ordering-will-no-longer-affect-group-by-queries), however I'm not sure we have a reasonable workaround (further comments below).


jeremystretch:
Under Django 3.1, a model's `Meta.ordering` is ignored when a `GROUP BY` SQL clause is in use. This means, for example, that

```python
DeviceType.objects.prefetch_related('manufacturer')
```

will order correctly, but

```python
DeviceType.objects.prefetch_related('manufacturer').annotate(instance_count=Count('instances'))
```

will be unordered.

jeremystretch:
Here are the resulting SQL queries for `DeviceType.objects.prefetch_related('manufacturer').annotate(instance_count=Count('instances'))`:

```python
# Django 3.0.8
SELECT ••• FROM "dcim_devicetype"
    LEFT OUTER JOIN "dcim_device" ON ("dcim_devicetype"."id" = "dcim_device"."device_type_id")
    INNER JOIN "dcim_manufacturer" ON ("dcim_devicetype"."manufacturer_id" = "dcim_manufacturer"."id")
    GROUP BY "dcim_devicetype"."id", "dcim_manufacturer"."name"
    ORDER BY "dcim_manufacturer"."name" ASC, "dcim_devicetype"."model" ASC

# Django 3.1b1
SELECT ••• FROM "dcim_devicetype"
    LEFT OUTER JOIN "dcim_device" ON ("dcim_devicetype"."id" = "dcim_device"."device_type_id")
    INNER JOIN "dcim_manufacturer" ON ("dcim_devicetype"."manufacturer_id" = "dcim_manufacturer"."id")
    GROUP BY "dcim_devicetype"."id", "dcim_manufacturer"."name"
```

jeremystretch:
Some rough benchmarking suggests that moving from `Count()` to [subqueries](https://github.com/netbox-community/netbox/blob/develop/netbox/utilities/utils.py#L67) may actually be the way to go:

```Provider.objects.annotate(count_circuits=Count('circuits'))```
~ 120msec, no ordering

```Provider.objects.annotate(count_circuits=get_subquery(Circuit, 'provider'))```
~ 50msec, preserved ordering

I'm not 100% confident in this assumption yet though; need to do some more testing. It would be nice to standardize on the way we annotate related object counts project-wide, though.

lampwins:
The issue we have hit with subqueries for counting aggregates is cacheops doesn't support them for invalidation. There is some discussion on this in https://github.com/netbox-community/netbox/issues/3672

jeremystretch:
I've opened Suor/django-cacheops#365 to see if we can get an idea how feasible this would be to work around.

jeremystretch:
Given that we cannot immediately accommodate the use of subqueries, I've opened and implemented #4871 to manually specify ordering on the affected querysets. Eventually, this will be removed, but it is necessary now to allow moving to Django 3.1 for the v2.9 release.

Issue #4877, opened 2020-07-22
Create a REST API endpoint for user accounts and groups


jeremystretch:
### Environment
* Python version: 3.6.9
* NetBox version: 2.8.8

### Proposed Functionality
Establish REST API endpoints at `/api/users/users/` and `/api/users/groups/` for managing local users and groups, respectively.

### Use Case
Enables programmatic CRUD operations on the built-in user and group models. Presently, user accounts can be manipulated only via the admin UI or user control panel.

### Database Changes
None

### External Dependencies
None

PR #4879, opened 2020-07-22
4877 users api endpoint

jeremystretch:
### Closes: #4877
- Introduce REST API endpoints for Django's `User` and `Group` models
- Extend the nested serializers for these models to reference the individual object views
- Exclude these models from wildcard view exemption

### DIFF
--- a/netbox/users/api/views.py
+++ b/netbox/users/api/views.py
@@ -1,7 +1,27 @@
+from django.contrib.auth.models import Group, User
+from django.db.models import Count
+
+from users import filters
+from users.models import ObjectPermission
 from utilities.api import ModelViewSet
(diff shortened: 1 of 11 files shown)


netbox/users/api/views.py
```diff
@@ -1,7 +1,27 @@
...
-from users.models import ObjectPermission
+
+#
+# Users and groups
+#
+
+class UserViewSet(ModelViewSet):
+    queryset = RestrictedQuerySet(model=User).prefetch_related('groups')
+    serializer_class = serializers.UserSerializer
+    filterset_class = filters.UserFitlerSet
```
glennmatthews, line 18: nit: s/Fitler/Filter/g

Issue #4903, opened 2020-07-27
Member count incorrect for Virtual Chassis members in search results


steffann:
### Environment
* Python version: 3.6.9
* NetBox version: 2.9-beta1

### Steps to Reproduce
1. Create a Virtual Chassis with multiple members
2. Search for the name of the virtual chassis
3. Inspect the number of reported members

For example: create virtual chassis `cr1.ams2` with members `cr1.ams2` (master), `fusion1.ams1`, `fusion1.ams2` and `fusion2.ams3`:
![image](https://user-images.githubusercontent.com/509689/88511969-103bbe80-cfe6-11ea-8971-113ba502d75a.png)

Searching for `cr1.ams2` shows:
![image](https://user-images.githubusercontent.com/509689/88512049-2d708d00-cfe6-11ea-917b-9c6598dba3a5.png)

### Expected Behavior
I expected to see Members = 4.

### Observed Behavior
It shows Members = 16.

steffann:
I discovered that this can be fixed by explicitly telling Django to count distinct devices with `member_count=Count('members', distinct=True)` in `SEARCH_TYPES`. I will provide a PR.

jeremystretch:
It looks like this is triggered specifically when _searching_ for virtual chassis. The `q` filter matches against both the virtual chassis name (in v2.9) and the name of the associated master device (if any). This results in joining the `dcim_device` table twice:

```sql
SELECT "dcim_virtualchassis"."id",
       "dcim_virtualchassis"."created",
       "dcim_virtualchassis"."last_updated",
       "dcim_virtualchassis"."master_id",
       "dcim_virtualchassis"."name",
       "dcim_virtualchassis"."domain",
       COUNT("dcim_device"."id") AS "member_count"
  FROM "dcim_virtualchassis"
  LEFT OUTER JOIN "dcim_device"
    ON ("dcim_virtualchassis"."id" = "dcim_device"."virtual_chassis_id")
  LEFT OUTER JOIN "dcim_device" T3
    ON ("dcim_virtualchassis"."id" = T3."virtual_chassis_id")
 WHERE (UPPER("dcim_virtualchassis"."name"::text) LIKE UPPER('%MyVC1%') OR UPPER(T3."name"::text) LIKE UPPER('%MyVC1%') OR UPPER("dcim_virtualchassis"."domain"::text) LIKE UPPER('%MyVC1%'))
 GROUP BY "dcim_virtualchassis"."id"
[...]

PR #4904, opened 2020-07-27
Make sure counts in search results are for distinct objects

steffann:
### Fixes: #4903

Virtual Chassis members were counted multiple times in search results. I added `distinct=True` to `Count` to make sure it returns the number of unique members. I added this to all counts in `SEARCH_TYPES` to make sure the same counting error doesn't occur for other object types as well,

### DIFF
--- a/netbox/netbox/views.py
+++ b/netbox/netbox/views.py
@@ -17,16 +17,12 @@ from dcim.filters import (
     CableFilterSet, DeviceFilterSet, DeviceTypeFilterSet, PowerFeedFilterSet, RackFilterSet, RackGroupFilterSet,
     SiteFilterSet, VirtualChassisFilterSet,
 )
-from dcim.models import (
-    Cable, ConsolePort, Device, DeviceType, Interface, PowerPanel, PowerFeed, PowerPort, Rack, RackGroup, Site,
-    VirtualChassis,
-)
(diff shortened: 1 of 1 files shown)


netbox/netbox/views.py
```diff
@@ -47,7 +43,7 @@
     # Circuits
     ('provider', {
         'queryset': Provider.objects.annotate(
-            count_circuits=Count('circuits')
+            count_circuits=Count('circuits', distinct=True)
```
jeremystretch, line 46: We should probably focus only on virtual chassis for this PR/issue, and open separate issues for similar issues that have been identified.

netbox/netbox/views.py
```diff
@@ -100,7 +96,7 @@
     }),
     ('virtualchassis', {
         'queryset': VirtualChassis.objects.prefetch_related('master').annotate(
-            member_count=Count('members')
+            member_count=Count('members', distinct=True)
```
jeremystretch, line 99: This is probably okay, but per my comment on #4903, we should ensure that the fix is applied consistently everywhere VirtualChassisFilterSet is used.

jeremystretch (changes requested):

jeremystretch:
Fixed #4903 in 09015c9

PR #4936, opened 2020-08-03
Closes #4869: Enable to search by MAC Address

### LINKED ISSUE #4869: Searching by MAC address
### Environment
* Python version:  3.7.6
* NetBox version:  2.8.7

### Proposed Functionality

Allow global searching of objects via MAC address.

### Use Case

This is one of the most common search mechanisms used by networking administrators when tracking down an unknown device or a device that may not be operating properly. 

### Database Changes

Possibly an index on the mac field for interfaces

kobayashi:
### Fixes: #4869 

This enables to search interface objects by mac address in global search box.

### DIFF
--- a/docs/release-notes/version-2.8.md
+++ b/docs/release-notes/version-2.8.md
@@ -4,6 +4,7 @@
 
 ### Enhancements
 
+* [#4869](https://github.com/netbox-community/netbox/issues/4869) - Allow global search by MAC address
 * [#4898](https://github.com/netbox-community/netbox/issues/4898) - Add MAC address search field to interfaces list
 * [#4899](https://github.com/netbox-community/netbox/issues/4899) - Add MAC address column to interfaces table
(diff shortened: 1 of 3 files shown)


jeremystretch:
Let's not extend global search to include device components, as that's going to add a significant performance impact (to an already very inefficient function) and commonly return a lot of irrelevant results.

As I understand it, #4869 calls for the ability to search devices/VMs by the MAC address of an assigned interface, so it should only be necessary to extend the `search()` methods on those filter sets to match against assigned interfaces.

jeremystretch:
@kobayashi are you still working on this?

kobayashi:
Sorry for late response. Yeah, this will make the result slower. Is it okay to search MAC address of assigned interface in global search bar?

kobayashi:
Close this to prevent performance impact.

PR #5039, opened 2020-08-23
Use correct serializer for available-prefixes POST response

### LINKED ISSUE #5002: Swagger spec for available-prefixes are incorrect
### Environment
* Python version: N/A
* NetBox version: 103a0e2b32d1 (v2.8.9)

### Steps to Reproduce
The swagger spec states that POST `/ipam/prefixes/{id}/available-prefixes/` returns list of `AvailablePrefix` objects, but it seems to actually return list of `Prefix` objects or single `Prefix` object depending on what is requested.  

### Expected Behavior
Expected the following command to return list of `AvailablePrefix` objects:
```
curl -X POST "https://localhost:8000/api/ipam/prefixes/<prefix-id>/available-prefixes/" -H "Authorization: Token <token>" -H "Content-Type: application/json" -d "{ \"prefix_length\": 24}"
```

### Observed Behavior
But instead returned what looks like a single `Prefix` object:
```
{"id":169,"family":{"value":4,"label":"IPv4"},"prefix":"10.0.1.0/24","site":null,"vrf":null,"tenant":null,"vlan":null,"status":{"value":"active","label":"Active","id":1},"role":null,"is_pool":false,"description":"","tags":[],"custom_fields":{},"created":"2020-08-14","last_updated":"2020-08-14T12:27:01.701782Z"}
```

Reference to:
https://github.com/netbox-community/netbox/blob/v2.8.9/netbox/ipam/api/views.py#L83

Happy to create pull create to fix this.

joakimhellum:
This pull request updates swagger spec for POST `/ipam/prefixes/{id}/available-prefixes/` endpoint to return single `Prefix` object, not list of `AvailablePrefix` objects.

### DIFF
--- a/netbox/ipam/api/views.py
+++ b/netbox/ipam/api/views.py
@@ -88,7 +88,7 @@ class PrefixViewSet(CustomFieldModelViewSet):
         return super().get_serializer_class()
 
     @swagger_auto_schema(method='get', responses={200: serializers.AvailablePrefixSerializer(many=True)})
-    @swagger_auto_schema(method='post', responses={201: serializers.AvailablePrefixSerializer(many=True)})
+    @swagger_auto_schema(method='post', responses={201: serializers.PrefixSerializer(many=False)})
     @action(detail=True, url_path='available-prefixes', methods=['get', 'post'])
     @advisory_lock(ADVISORY_LOCK_KEYS['available-prefixes'])
     def available_prefixes(self, request, pk=None):


netbox/ipam/api/views.py
```diff
@@ -88,7 +88,7 @@ def get_serializer_class(self):
         return super().get_serializer_class()
 
     @swagger_auto_schema(method='get', responses={200: serializers.AvailablePrefixSerializer(many=True)})
-    @swagger_auto_schema(method='post', responses={201: serializers.AvailablePrefixSerializer(many=True)})
+    @swagger_auto_schema(method='post', responses={201: serializers.PrefixSerializer(many=False)})
```
jeremystretch, line 91: Shouldn't `many` be True here, since it's possible to create multiple prefixes simultaneously?

joakimhellum: @jeremystretch I did this initially, but realized after some testing that it doesn't match the request body, which is a single `PrefixLength` object and suggests that only single `Prefix` object will be returned. 
If we set `many` to be `True` here for the response, then we also need to update the request body to be list of `PrefixLength` objects. But then it doesn't really match rest of the swagger spec, since most of the POST reqeust bodies and responses in the swagger spec are defined as single objects and not list of objects. 
Have used the `available-ips` endpoint as a reference to this (https://github.com/netbox-community/netbox/blob/v2.9.1/netbox/ipam/api/views.py#L166-L168).
FYI have built NetBox from this branch and tested generating a NetBox client using `go-swagger`.

PR #5071, opened 2020-08-29
Redirect users away from /login, if they are already logged in 

### LINKED ISSUE #4977: Redirect users away from /login, if they are already logged in
### Environment
* Python version: 3.8.5 
* NetBox version: 2.8.9 

### Proposed Functionality
When a already logged-in user access the `/login` url, they should be redirected to home or to the page from the parameter `next`.

I created a simple PoC: https://github.com/n0emis/netbox/commit/061549b0f0b82d8defaad0fbc2fb21ff9765d6a4

### Use Case
This would be especially usefull when using a Remote-Auth-Backend like [`django.contrib.auth.middleware.PersistentRemoteUserMiddleware`](https://docs.djangoproject.com/en/3.0/ref/middleware/#django.contrib.auth.middleware.PersistentRemoteUserMiddleware), where the HTTP-Header only get's send when accessing the login-page.

### Database Changes
No database-changes required

### External Dependencies
No new external Dependencies required

n0emis:
### Fixes: #4977 

Before returning the login form, the corresponding method now checks if the user is already logged in. when this is the case, the user get's redirected.

### DIFF
--- a/netbox/users/views.py
+++ b/netbox/users/views.py
@@ -36,6 +36,15 @@ class LoginView(View):
         return super().dispatch(*args, **kwargs)
 
     def get(self, request):
+        if request.user.is_authenticated:
+            # Already logged-in, determine where to redirect
+            redirect_to = request.GET.get('next', reverse('home'))
+            if redirect_to and not is_safe_url(url=redirect_to, allowed_hosts=request.get_host()):
+                logger.warning(f"Ignoring unsafe 'next' URL passed to login form: {redirect_to}")
+                redirect_to = reverse('home')
+
+            return HttpResponseRedirect(redirect_to)
+
         form = LoginForm(request)
 
         return render(request, self.template_name, {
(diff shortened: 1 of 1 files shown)


netbox/users/views.py
```diff
@@ -36,6 +36,15 @@ def dispatch(self, *args, **kwargs):
         return super().dispatch(*args, **kwargs)
 
     def get(self, request):
+        if request.user.is_authenticated:
+            # Already logged-in, determine where to redirect
+            redirect_to = request.GET.get('next', reverse('home'))
+            if redirect_to and not is_safe_url(url=redirect_to, allowed_hosts=request.get_host()):
+                logger.warning(f"Ignoring unsafe 'next' URL passed to login form: {redirect_to}")
+                redirect_to = reverse('home')
```
jeremystretch, line 44: This duplicates code within `post()` below. This logic should be moved into a separate method that can be used for both requests.

jeremystretch (changes requested):
Avoid duplicating code

netbox/users/views.py
n0emis: I've moved the code to an own method, please recheck.

Issue #1503, opened 2017-09-18
Allow the assignment of secrets to objects other than devices


elmmare:
### Issue type
[X] Feature request 
[ ] Bug report      
[ ] Documentation   

### Environment
* Python version: 2.7.5
* NetBox version: 2.1.4

### Description
Secret cannot be attached to Sites and Racks but only to devices.

Usage cases:
*Store combination lock codes for racks or cage 
*Store  Site access codes

jeremystretch:
Extending secrets from devices to multiple objects requires replacing the ForeignKey field tied to the Device model with a GenericForeignKey. This in itself is straightforward, however we have to consider how this relationship should be displayed in the API.

Currently, a secret retrieved via the API looks like this:

```
{
    "id": 1234,
    "device": {
        "id": 5678,
        "url": "http://localhost:8000/api/dcim/devices/5678/",
        "name": "MyDevice",
        "display_name": "MyDevice"
    },
    "role": <...>,
    "name": "root",
    "plaintext": null,
    "hash": "<hash>",
    ...
}
```

If we switch to a generic relationship, we need some way to indicate the type of the parent object, both while retrieving existing secrets and while creating new ones. The [DRF documentation](http://www.django-rest-framework.org/api-guide/relations/#generic-relationships) talks about this briefly, but does not suggest any particular approach.

darkstar:
How about integrating with HashiCorp's Vault (https://www.vaultproject.io) as backend storage for secrets? If it is even remotely possible/interesting I can open a feature request issue for it.

rlaneyjr:
I agree with darkstar.  I had thought about making an attempt to integrate it in myself.  Then realized I am a novice at best with Python/Django and lost with GO.

jeremystretch:
I looked into Vault back when I was first working on secrets storage. While it can certainly be used in conjunction with NetBox (with Vault serving as a replacement for NetBox's built-in storage), I don't think it makes sense for NetBox to integrate directly with Vault. Doing so would incur a substantial new dependency while not providing any significant value as opposed to accessing Vault directly. Let's stick with the functionality we have now.

linasjan:
Secrets for tenants also would be appreciated.

larsuhartmann:
This feature would be highly apreciated!

c0dyhi11:
Hello,

[...]

PR #5151, opened 2020-09-18
#1503: Extend secrets assignment to virtual machines

jeremystretch:
### Closes: #1503

- Replaces the `device` field on the Secret model with a GenericForeignKey
- Requires the assignment of a secret to a device _or_ a virtual machine


AndyGauge:
This is very exciting to see!

Issue #3436, opened 2019-08-19
API support for bulk object deletion


lamoni:
### Environment
* Python version:  Python 3.7
* NetBox version:  2.6.2

### Proposed Functionality
Allow bulk deletion of sites and devices (preferably anything) through API end-point.

### Use Case
We have a use case that requires bulk deletion of sites and devices for automated deployment and testing of a Netbox instance. Currently, we have to iterate through the list of objects and delete them 1 by 1 (i.e. 1 HTTP request per device/site/whatever), this can take upwards of minutes. We can bulk create just fine (which reduced time to finish by a lot), but can't bulk delete. The GUI has the ability, but there isn't an API endpoint to do so.

lampwins:
This has been in the back of my mind for a while, and I am surprised it has not come up before now. As we look to make UI improvements down the road, this will be needed.

We should be able to do this with a non-detail action route on our modified ModelViewSet class.

I would want to be able to pass a list of PKs or a list of objects that use the same queryset field lookup logic we implemented not long ago, as the quest body.

For example for sites, you could pass either:
```
[1, 4, 6]
```
Or
```
[{"name": "Site A"}, {"name": "Site B"}, {"name": "Site C"}]
```

gunnertwin:
I also have a requirement for this. I'm pulling data from Azure and want my netbox to match my Azure estate exactly. So if someone deletes a vnet in Azure, that vnet also needs to be removed in Netbox. So I would like to run my pipeline which will initially bulk delete everything and then add the current state of Azure

jeremystretch:
I've made good progress on this in the branch `3436-api-bulk-update`, however I've hit a wall troubleshooting an OpenAPI schema generation issue. The introduction of the bulk delete view and subsequent support for the `DELETE` HTTP action under a list endpoint results in a duplicate operation ID when generating the schema:

```
SwaggerValidationError at /api/docs/

spec validation failed: {'ssv': 'Duplicate operationId: circuits_circuit-terminations_delete'}
```

jeremystretch:
[...]

PR #5163, opened 2020-09-22
#3436: Support for bulk deletion via REST API

jeremystretch:
### Closes: #3436

- Extend ModelViewSet with BulkDestroyModelMixin to support bulk deletion of objects
- Modify OrderedDefaultRouter to accept `DELETE` operations on list routes
- Monkey-patch DRF's `is_custom_action()` to avoid SwaggerValidationError resulting from duplicate operation IDs
- Tests & documentation

### DIFF
--- a/docs/rest-api/overview.md
+++ b/docs/rest-api/overview.md
@@ -529,3 +529,17 @@ Note that `DELETE` requests do not return any data: If successful, the API will 
 
 !!! note
     You can run `curl` with the verbose (`-v`) flag to inspect the HTTP response codes.
+
+### Deleting Multiple Objects
+
+NetBox supports the simultaneous deletion of multiple objects of the same type by issuing a `DELETE` request to the model's list endpoint with a list of dictionaries specifying the numeric ID of each object to be deleted. For example, to delete sites with IDs 10, 11, and 12, issue the following request:
+
+```no-highlight
+curl -s -X DELETE \
+-H "Authorization: Token $TOKEN" \
+http://netbox/api/dcim/sites/ \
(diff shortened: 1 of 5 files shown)


docs/rest-api/overview.md
```diff
@@ -529,3 +529,17 @@ Note that `DELETE` requests do not return any data: If successful, the API will
...
+
+```no-highlight
+curl -s -X DELETE \
+-H "Authorization: Token $TOKEN" \
+http://netbox/api/dcim/sites/ \
+--data '[{"id": 10}, {"id": 11}, {"id": 12}]'
+```
+
+!!! note
+    The bulk deletion of objects is an all-or-none operation, meaning that if NetBox fails to delete any of the specified objects (e.g. due a dependency by a related object), the entire operation will be aborted and none of the objects will be deleted.
```
glennmatthews, line 545: `...due to a dependency...`

netbox/netbox/api.py
```diff
@@ -4,11 +4,22 @@
...
+from rest_framework.schemas import coreapi
 from rest_framework.utils import formatting
 
 from users.models import Token
 
 
+def is_custom_action(action):
+    return action not in {
+        'retrieve', 'list', 'create', 'update', 'partial_update', 'destroy', 'bulk_destroy'
+    }
```
glennmatthews, line 16: Wonder if it would be better to do something like:

```python
from restframework.schemas.coreapi import is_custom_action as base_is_custom_action

def is_custom_action(action):
    return base_is_custom_action(action) and action not in {'bulk_destroy'}
```

So that we are clearly extending the base function, just in case it ever changes?

Issue #4882, opened 2020-07-23
API support for bulk object updates/edits (put/patch)


ttarabula:
### Environment
* Python version:  Python 3.6
* NetBox version:  Netbox 2.8.8

### Proposed Functionality

Allow for bulk editing (put/patch) of devices (preferably any object) through API end-points.

Related issue (bulk deletes): #3436 

### Use Case

Currently, iterative updates to objects doesn't scale well and API support for a single call to edit multiple objects via the API could streamline various access patterns. More generally though, since bulk create (post) is already supported, and the bulk delete enhancement proposal was accepted, it would be nice to have this as well for completeness' sake. Similar to bulk delete, since the web interface also supports this, I can see official support in the API simplifying the web interface code as well.

jeremystretch:
Marking as blocked by #3436 (still need to address the schema generation issue)

PR #5166, opened 2020-09-22
#4882: Support bulk updates via REST API

jeremystretch:
### Closes: #4882

- Introduce BulkUpdateModelMixin to support `PATCH` via object list endpoints
- Tests & documentation

### DIFF
--- a/netbox/netbox/api.py
+++ b/netbox/netbox/api.py
@@ -12,7 +12,10 @@ from users.models import Token
 
 def is_custom_action(action):
     return action not in {
-        'retrieve', 'list', 'create', 'update', 'partial_update', 'destroy', 'bulk_update', 'bulk_destroy'
+        # Default actions
+        'retrieve', 'list', 'create', 'update', 'partial_update', 'destroy',
+        # Bulk operations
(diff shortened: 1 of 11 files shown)


netbox/netbox/api.py
```diff
@@ -12,7 +12,10 @@
 
 def is_custom_action(action):
     return action not in {
-        'retrieve', 'list', 'create', 'update', 'partial_update', 'destroy', 'bulk_destroy'
+        # Default actions
+        'retrieve', 'list', 'create', 'update', 'partial_update', 'destroy',
+        # Bulk operations
+        'bulk_update', 'bulk_partial_update', 'bulk_destroy',
     }
```
glennmatthews, line 19: Wonder if it would be better to do something like:

```python
from restframework.schemas.coreapi import is_custom_action as base_is_custom_action

def is_custom_action(action):
    return base_is_custom_action(action) and action not in {
        'bulk_update', 'bulk_partial_update', 'bulk_destroy',
    }
```

So that we are clearly extending the base function, just in case it ever changes?

jeremystretch: If the built-in operations change, we'll have much, much bigger signs than the schema generation failing. :smile:

Issue #5192, opened 2020-09-29
Excessive response times for LDAP user accounts


gdanielson:
### Environment
* Python version: 3.7
* NetBox version: 2.9.4

### Steps to Reproduce
1. Enable LDAP auth, REMOTE_AUTH_BACKEND = 'netbox.authentication.LDAPBackend'

2. As a standard user or with Staff status enabled, perform standard
activities via the web browser interface.

3. All transacations are observed to take an excessive amount of time.

### Expected Behavior
Expect that SuperUser and non-SuperUser accounts to have approximately similar
response time experience.

### Observed Behavior
A browser refresh of the NetBox homepage takes approx. 20 seconds.

Observing the logfile for the above transaction we see many hundreds of lines
logged saying `Initiating TLS` followed by many hundreds of lines of ldap
`search_s(...` activity.

Changing django console handler to level ERROR the above response time is
reduced to approx. 10 seconds

When the same user account is enabled for SuperUser a browser refresh of the
NetBox home page takes approx. 1 second.

These extended response times have been observed since upgrading to v2.9

gdanielson:
With the logs showing LDAP activity for every transaction it seems that LDAP caching has mysteriously stopped working.
The settings we have for LDAP caching are:

`AUTH_LDAP_CACHE_GROUPS = True`
`AUTH_LDAP_GROUP_CACHE_TIMEOUT = 1800`

The django-auth doc for LDAP caching says that these params are deprecated but do still work.
https://django-auth-ldap.readthedocs.io/en/latest/reference.html?#auth-ldap-cache-timeout
> Changed in version 1.6.0: Previously caching was controlled by the settings AUTH_LDAP_CACHE_GROUPS and AUTH_LDAP_GROUP_CACHE_TIMEOUT. If AUTH_LDAP_CACHE_GROUPS is set, the AUTH_LDAP_CACHE_TIMEOUT value is derievd from these deprecated settings.

I will try changing to the new param `AUTH_LDAP_CACHE_TIMEOUT` then test.

jeremystretch:
@gdanielson Could you have a look at the report in #5194 please and see if it's related to your issue?

gdanielson:
> @gdanielson Could you have a look at the report in #5194 please and see if it's related to your issue?

We are not seeing obvious increase of CPU consumption over time (we are running in a container in AWS ECS so observability of this is not always so clear cut).

[...]

PR #5195, opened 2020-09-29
Fixes 5194 logger should be initialized elsewhere.

nniehoff:
### Fixes: 5194

After digging into the CPU usage in our environment I generated several flamegraphs to determine what function was using the CPU time.  I found most of the time being spent in addHandler on line 177.  The addHandler and setLevel I believe should be handled by the LOGGING definition in configuration.py or as documented [here](https://netbox.readthedocs.io/en/stable/installation/6-ldap/) in ldap_config.py.  I also moved the ldap_logger definition to the beginning of the __new__ method as this follow existing standards.  Please let me know if something else, like docs, should be changed and I'd be happy to make those changes.

### DIFF
--- a/netbox/netbox/authentication.py
+++ b/netbox/netbox/authentication.py
@@ -136,6 +136,9 @@ class RemoteUserBackend(_RemoteUserBackend):
 class LDAPBackend:
 
     def __new__(cls, *args, **kwargs):
+        # Enable logging for django_auth_ldap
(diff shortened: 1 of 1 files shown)


glennmatthews:
It looks to me like `ldap_logger` isn't even used in the function after your changes - it might be more correct to delete it entirely rather than relocating it.

Possibly also related: #5192.

PR #5208, opened 2020-10-02
5197: Limit main IPAddress view to a max of 10 duplicate addresses; add new duplicates view

### LINKED ISSUE #5197: Long Response Time for Viewing IP Addresses with A Large Number of Duplicate Assignments
### Environment
* Python version: 3.6.8 
* NetBox version: 2.8.6

### Steps to Reproduce
1. Create any IP Address in Netbox and assign it to a large number of interfaces (1000+)
2. View the IP Address object

Capture

### Expected Behavior
- The page to load faster
- The list of duplicate IP addresses to be paginated (similar to the changelog view on the home page)

### Observed Behavior
- Long response time (3m15s for 16K duplicates)
   - We have a lot of access points....
- Gunicorn worker at 100% for the duration of the response

glennmatthews:
### Fixes: #5197 

- Limit number of duplicate addresses in main IPAddressView to a maximum of 10
- If more than 10 duplicate addresses exist, add a "View all..." link to the duplicate address table
- This link leads to a new IPAddressDuplicatesView which provides a paginated table of all duplicate IPs.

Tested with 1000 duplicate addresses; main view loads quickly as does the new view with default pagination settings.

image

image

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -527,7 +527,8 @@ class IPAddressView(ObjectView):
         # Exclude anycast IPs if this IP is anycast
         if ipaddress.role == IPAddressRoleChoices.ROLE_ANYCAST:
             duplicate_ips = duplicate_ips.exclude(role=IPAddressRoleChoices.ROLE_ANYCAST)
-        duplicate_ips_table = tables.IPAddressTable(list(duplicate_ips), orderable=False)
+        # Limit to a maximum of 10 duplicates displayed here
+        duplicate_ips_table = tables.IPAddressTable(duplicate_ips[:10], orderable=False)
 
         # Related IP table
         related_ips = IPAddress.objects.restrict(request.user, 'view').exclude(
@@ -547,6 +548,7 @@ class IPAddressView(ObjectView):
             'ipaddress': ipaddress,
             'parent_prefixes_table': parent_prefixes_table,
             'duplicate_ips_table': duplicate_ips_table,
+            'more_duplicate_ips': duplicate_ips.count() > 10,
             'related_ips_table': related_ips_table,
         })
 
--- a/netbox/templates/ipam/ipaddress.html
+++ b/netbox/templates/ipam/ipaddress.html
@@ -3,6 +3,7 @@
 {% load custom_links %}
 {% load helpers %}
 {% load plugins %}
+{% load render_table from django_tables2 %}
 
 {% block header %}
     <div class="row noprint">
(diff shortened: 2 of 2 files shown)


netbox/ipam/views.py
```diff
@@ -547,10 +548,42 @@ def get(self, request, pk):
             'ipaddress': ipaddress,
             'parent_prefixes_table': parent_prefixes_table,
             'duplicate_ips_table': duplicate_ips_table,
+            'more_duplicate_ips': duplicate_ips.count() > 10,
             'related_ips_table': related_ips_table,
         })
 
 
+class IPAddressDuplicatesView(ObjectListView):
```
jeremystretch, line 556: We shouldn't need a separate view for this. Just linking to the IP addresses list with the appropriate `address` and `vrf_id` filters should suffice.

netbox/ipam/views.py
```diff
@@ -527,7 +527,8 @@ def get(self, request, pk):
         # Exclude anycast IPs if this IP is anycast
         if ipaddress.role == IPAddressRoleChoices.ROLE_ANYCAST:
             duplicate_ips = duplicate_ips.exclude(role=IPAddressRoleChoices.ROLE_ANYCAST)
-        duplicate_ips_table = tables.IPAddressTable(list(duplicate_ips), orderable=False)
+        # Limit to a maximum of 10 duplicates displayed here
+        duplicate_ips_table = tables.IPAddressTable(list(duplicate_ips[:10]), orderable=False)
```
jeremystretch, line 531: I don't think you need `list()` here; slicing returns a list.

netbox/templates/ipam/ipaddress.html
```diff
@@ -159,7 +160,18 @@ <h1>{% block title %}{{ ipaddress }}{% endblock %}</h1>
...
         {% include 'panel_table.html' with table=parent_prefixes_table heading='Parent Prefixes' %}
         {% if duplicate_ips_table.rows %}
-            {% include 'panel_table.html' with table=duplicate_ips_table heading='Duplicate IP Addresses' panel_class='danger' %}
+            {# Custom version of panel_table.html #}
+            <div class="panel panel-danger">
+                <div class="panel-heading">
+                    <strong>Duplicate IP Addresses</strong>
+                    {% if more_duplicate_ips %}
+                    <div class="pull-right">
+                        <a href="{% url 'ipam:ipaddress_duplicates' pk=ipaddress.pk %}">View all&hellip;</a>
```
jeremystretch, line 169: We try to stay away from legacy HTML elements since they can be difficult to read. I'd recommend using a FontAwesome icon (e.g. `<i class="fa fa-arrow-right"></i>`) or just omitting the ellipsis.

jeremystretch (changes requested):
Thanks for your work! I apologize if I implied that we needed a new view for this; just linking to the filtered IP addresses list should be fine.

netbox/ipam/views.py
glennmatthews: 🤦 I don't know why I didn't think of that. Will do.

netbox/templates/ipam/ipaddress.html
glennmatthews: Changed it to a button:

image

glennmatthews:
Much simpler now - thanks for catching my over-complication. :-)

Issue #4900, opened 2020-07-24
Implement new mechanism for tracking end-to-end connections across cable paths


jeremystretch:
### Proposed Changes
NetBox tracks end-to-end connections over pass-through ports. When a cable is saved, traces are done in both directions, and if a complete path has been established, the peer termination information is recorded at either of its endpoints.

This proposal seeks to devise and implement a new mechanism for tracking these end-to-end paths separately from the cable termination models. It may be desirable to introduce a new model (e.g. CablePath) dedicated to holding this information. Although more research is needed to identify the ideal solution, I've opened this issue now so that we can start referencing it as a blocker.

### Justification
The association of end-to-end path data with specific device components and circuit terminations has proved problematic and inefficient (see the linked issues for examples).

jeremystretch:
Although many of the details need to be refined, the high-level approach will be to model discrete cable paths individually for each connected object. For example, consider the following topology:

```
                1                           2
[Device 1] -----------+               +----------- [Device 2]
      Iface1          |               |          Iface2
                  FP1 |       3       | FP2
                  [Panel 1] ----- [Panel 2]
                  FP3 |   RP1   RP2   | FP4
      Iface3          |               |          Iface4
[Device 3] -----------+               +----------- [Device 4]
                4                           5
```

There are two end-to-end connections present, both via cable 3:

* Iface 1 <---> Iface 2
* Iface 3 <---> Iface 4

Currently, these connections are conveyed only on the four individual interfaces. A more robust approach would record each of the four unidirectional end-to-end connections individually, as:

1. (Iface1, Cable 1, FP1, RP1, Cable 3, RP2, FP2, Cable 2, Iface2)
2. (Iface2, Cable 2, FP2, RP2, Cable 3, RP1, FP1, Cable 1, Iface1)
3. (Iface3, Cable 4, FP3, RP1, Cable 3, RP2, FP4, Cable 5, Iface4)
4. (Iface4, Cable 5, FP4, RP2, Cable 3, RP1, FP3, Cable 4, Iface3)

This approach ensures that we can immediately query all potentially affected points along a path any time any single component within the path changes. For example, if cable 2 is deleted, we know immediately that paths 1 and 2 need to be updated.

steffann:
I have experimented a bit with something similar: https://gist.github.com/steffann/b0cf1340fe0825bbffeba73822867d3b

Feel free to use!

seismiccollision:
[...]

PR #5212, opened 2020-10-05
#4900: New model for cable paths

jeremystretch:
### Closes: #4900

- Introduce a new CablePath model for recording the origin, destination, intermediate nodes, and status of each end-to-end path
- Add a ForeignKey to CablePath on each of the seven path endpoint models
- Drop the old fields for `connected_endpoint` and `connection_status`
- Cache peer termination on CableTermination instances
- Remove `trace` API endpoint from FrontPort and RearPort
- Remove `connection_status` field from nested API serializers for device components
- Add `cable_peer` field to API serializer for all CableTermination models
- Overhaul cable tracing logic with the new `trace_paths()` utility function
- Completely rewrite all cable path tests
- Add the `retrace_paths` management command (temporarily?)


netbox/dcim/api/views.py
```diff
@@ -469,49 +469,43 @@ def napalm(self, request, pk):
...
 
 
-class InterfaceViewSet(CableTraceMixin, ModelViewSet):
-    queryset = Interface.objects.prefetch_related(
-        'device', '_connected_interface', '_connected_circuittermination', 'cable', 'ip_addresses', 'tags'
-    ).filter(
-        device__isnull=False
-    )
+class InterfaceViewSet(PathEndpointMixin, ModelViewSet):
+    queryset = Interface.objects.prefetch_related('device', '_path', 'cable', 'ip_addresses', 'tags')
```
glennmatthews, line 497: No longer filtering out null devices?

netbox/dcim/management/commands/retrace_paths.py
```diff
@@ -0,0 +1,68 @@
...
+)
+
+
+class Command(BaseCommand):
+    help = "Recalculate natural ordering values for the specified models"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'args', metavar='app_label.ModelName', nargs='*',
+            help='One or more specific models (each prefixed with its app_label) to retrace',
```
glennmatthews, line 27: If this is "one or more", should `nargs='+'` instead?

jeremystretch:
@glennmatthews I appreciate your enthusiasm, however this is a draft PR, and not yet open for review.

jeremystretch:
To avoid jamming any complex logic into schema migrations, I've opted to retain the `trace_paths` management command and call it within `upgrade.sh`. This will, by default, automatically create any missing CablePaths. In theory, this should be necessary only when upgrading to v2.10, however I've also included a `--force` option to force the recalculation of all CablePath objects should the need arise.

I'm still mulling options for conveying the presence of multiple cable paths through a RearPort with multiple FrontPorts mapped to it in the REST API. I've dropped the `/trace/` endpoint from pass-through ports for now.

PR #5253, opened 2020-10-19
Limit parent to same site when importing RackGroup

### LINKED ISSUE #5247: Rack import fails when rack group is not globally unique
### Environment
* Python version: 3.6.9
* NetBox version: 2.9.3

### Steps to Reproduce
1. In site A create rackgroup RG1
2. In site B create rackgroup RG1
3. Use the import function to import a rack into site A rackgroup RG1

```
site,parent,name,slug,description
A,RG1,R1,r1,
```
One can create the rack with the same information in the ui which works,
also an export of the manually imported rack will give the above CSV which 
cannot be imported.

### Expected Behavior
The rack is imported correctly. 

### Observed Behavior
Following error is displayed:
```
Row 1 parent: "Cage 1 (DH01)" is not a unique value for this field; multiple objects were found
Row 1 __all__: Rack group with this Site and Name already exists.
```

martink2:
### Fixes: #5247

If multiple sites have the same RackGroup name used within them,
import fails, this limits the parent queryset to parents in the site 
the group is added in.

### DIFF
--- a/netbox/dcim/forms.py
+++ b/netbox/dcim/forms.py
@@ -370,6 +370,15 @@ class RackGroupForm(BootstrapMixin, forms.ModelForm):
             'site', 'parent', 'name', 'slug', 'description',
         )
 
+    def __init__(self, data=None, *args, **kwargs):
+        super().__init__(data, *args, **kwargs)
+
+        if data:
+
+            # Limit parent queryset by assigned site
+            params = {f"site__{self.fields['site'].to_field_name}": data.get('site')}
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thank you for your interest in contributing to NetBox, however it appears there is no accepted issue that correlates to this pull request.  Before pull requests are opened, we require an **accepted** issue as per our [contributing guide](https://github.com/digitalocean/netbox/blob/master/CONTRIBUTING.md).

PR #5266, opened 2020-10-25
4559 config context rendering

### LINKED ISSUE #4559: Improve Performance of Generating Config Contexts
### Environment
* Python version:  3.7
* NetBox version:  2.7.12, but issue exists on <= 2.8.1

### Proposed Functionality

Improve the [`get_config_context` serializer method](https://github.com/netbox-community/netbox/blob/develop/netbox/dcim/api/serializers.py#L428) that is observed to have slow performance that builds up as the number of fetched devices at a time increases.

I'm not sure exactly where the bottleneck is, but it does perform a very complex query that may be the cause of the overhead. I performed some profiling awhile back and noticed that the SQL queries used an acceptable amount of request time, but that it was mostly using CPU execution time. I'm not sure if this is still the case across the board, but it may be worthwhile to profile and optimize whatever calls are made under this `get_config_context` serializer method.

### Use Case

Fetching 1000 devices by default takes ~60s. _Without_ config context - using `?excludes=config_context` - the response takes ~5 seconds. This large overhead was determined to be from the generation & merging of the config contexts that exist on devices / virtual machines. Reducing this performance overhead would make NetBox's usage much more friendly and acceptable.

### Database Changes

This context is regenerated on every request on via [this serialize](https://github.com/netbox-community/netbox/blob/develop/netbox/dcim/api/serializers.py#L428). This then calls [this complex query](https://github.com/netbox-community/netbox/blob/d8cb58c74653da88d9aade40548b146a9311f5e6/netbox/extras/querysets.py#L24) which either in the query, object generation, or serialization takes an enormous amount of time.

We would need to evaluate the performance of this query / method and determine where the bottleneck is occurring, and how to make it faster.

### External Dependencies
N/A

---

Note that this is a revised version of #4544

lampwins:
### Fixes: #4559 

This implements a new queryset manager for the device and virtual machine models which includes an `annotate_config_context_data()` method. This method uses a dynamically generated subquery with the `JSONBAgg` function to annotate the list of config context json objects directly on the device and virtual machine querysets.

There are a couple of things to note:

- Django has included the postgres `JSONBAgg` function for a while but in 3.2 it includes support for aggregation ordering. My implementation includes a custom subclass which implements the same until we move to 3.2. Ordering is a critical component of being able to do this as a subquery.
- The original `get_for_object()` method is still present and actively used. This is still needed to query for the actual `ConfigContext` instances for a given object, and in certain cases where the annotation is not available.
- Because the annotation method is optional and not added to the queryset by default, the existing ability to exclude config context data from the API response is still in place.

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -24,7 +24,7 @@ from dcim.models import (
     VirtualChassis,
 )
 from extras.api.serializers import RenderedGraphSerializer
-from extras.api.views import CustomFieldModelViewSet
+from extras.api.views import ConfigContextQuerySetMixin, CustomFieldModelViewSet
 from extras.models import Graph
 from ipam.models import Prefix, VLAN
 from utilities.api import (
@@ -336,27 +336,12 @@ class PlatformViewSet(ModelViewSet):
 # Devices
 #
 
-class DeviceViewSet(CustomFieldModelViewSet):
+class DeviceViewSet(CustomFieldModelViewSet, ConfigContextQuerySetMixin):
     queryset = Device.objects.prefetch_related(
         'device_type__manufacturer', 'device_role', 'tenant', 'platform', 'site', 'rack', 'parent_bay',
         'virtual_chassis__master', 'primary_ip4__nat_outside', 'primary_ip6__nat_outside', 'tags',
     )
     filterset_class = filters.DeviceFilterSet
-
-    def get_queryset(self):
-        """
-        Build the proper queryset based on the request context
-
-        If the `brief` query param equates to True or the `exclude` query param
-        includes `config_context` as a value, return the base queryset.
-
-        Else, return the queryset annotated with config context data
-        """
-
-        request = self.get_serializer_context()['request']
-        if request.query_params.get('brief') or 'config_context' in request.query_params.get('exclude', []):
-            return self.queryset
-        return self.queryset.annotate_config_context_data()
 
     def get_serializer_class(self):
         """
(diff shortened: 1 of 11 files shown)


netbox/dcim/api/views.py
```diff
@@ -343,6 +343,21 @@ class DeviceViewSet(CustomFieldModelViewSet):
     )
     filterset_class = filters.DeviceFilterSet
 
+    def get_queryset(self):
```
jeremystretch, line 346: Can we move this to a mixin to avoid replicating the code in `virtualization/api/views.py`?

netbox/extras/querysets.py
```diff
@@ -57,3 +61,78 @@ def get_for_object(self, obj):
...
+        )
+
+        if self.model._meta.model_name == 'device':
+            base_query.add((Q(roles=OuterRef('device_role')) | Q(roles=None)), Q.AND)
+            base_query.add(
+                (Q(
+                    regions__tree_id=OuterRef('site__region__tree_id'),
+                    regions__level__lte=OuterRef('site__region__level'),
+                    regions__lft__lte=OuterRef('site__region__lft'),
+                    regions__rght__gte=OuterRef('site__region__rght'),
```
jeremystretch, line 117: Do you think it makes sense to merge the region MPTT stuff for devices/VMs, and use a format string for `OuterRef`? E.g. set `region_field` depending on the model and call `OuterRef(f'{region_field}__tree_id')`. Or is it cleaner to keep as-is? Primarily I'm just worried about ensuring that any changes to the region filters get replicated for both devices and VMs.

jeremystretch (commented):
Great work! And yes, much faster. Just had a couple minor points. If you don't feel the need to act on either, I'm fine with merging this as-is. :+1:

PR #5268, opened 2020-10-26
update: autocomplete attr on login fields

ndom91:
Updated the login field autocomplete values in accordance with modern security best practices. 

According to MDN browsers ignore `autocomplete='off'` on password fields, but `autocomplete='new-password'` should do the trick. 

Source: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion


jeremystretch:

PR #5276, opened 2020-10-27
Update upgrading.md

systeembeheerder:
Fix paths to prevent creating a scripts dir within scripts. 
Fix paths to prevent creating a reports dir within reports.
Fix path for media to relative just as all other cp commands

### Fixes: 5275

fix paths


DanSheps:

PR #5286, opened 2020-10-29
Fixes #5285: Include plugin API endpoints in root view without needing app_name set

### LINKED ISSUE #5285: Plugins must define app_name in api/urls.py to be included in API root view
### Environment
* Python version: 3.6.9
* NetBox version: 2.9.7

### Steps to Reproduce
I'm actually not certain whether this is better classified as a bug or a lapse in the documentation, but I'm opting to propose it as a bug because I think it can be addressed by a code change.

1. Create/install a plugin which provides at least one REST API endpoint. (Assume the plugin's name is `myplugin`.)
2. Within `api/urls.py`, set `app_name = myplugin-api`
3. Visit `/api/plugins/` in the browser. The plugin should be included.
4. Remove the `app_name` definition from `api/urls.py`.
5. Visit `/api/plugins/` in the browser again. The plugin no longer appears.

### Expected Behavior
Plugin API endpoints should be detected automatically. AFAICT there's no reason a plugin should need to declare `app_name`. (The core apps do this, but only because they're being included dynamically.)

### Observed Behavior
Plugin API endpoints disappear from the list when `app_name` is not declared.

It seems like the need for the `app_name` variable could be obviated within `PluginsAPIRootView._get_plugin_entry()` by changing

```
api_app_name = import_object(f"{plugin}.api.urls.app_name")
```

to

```
api_app_name = f'{app_config.name}-api'
```

This actually raises another point: the full URL name for a plugin API endpoint is currently in the format `plugins-api:myplugin-api:url-name`, which seems a bit unwieldy: `plugins-api:myplugin:url-name` should suffice. However, this would be a breaking change for any plugins which rely on reverse URL resolution to their REST API endpoints.

jeremystretch:
### Fixes: #5285

Include plugin REST API endpoints in the root view without requiring the plugin to define `app_name` in its `api/urls.py`.

### DIFF
--- a/netbox/extras/plugins/views.py
+++ b/netbox/extras/plugins/views.py
@@ -10,8 +10,6 @@ from rest_framework.response import Response
 from rest_framework.reverse import reverse
 from rest_framework.views import APIView
 
-from extras.plugins.utils import import_object
-
 
 class InstalledPluginsAdminView(View):
     """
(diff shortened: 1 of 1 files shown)


lampwins (approved):
LGTM

PR #5290, opened 2020-10-29
Fixing #5282 - moving logging configuration from to configuration.py

### LINKED ISSUE #5282: Netbox Consuming 100% of CPU with LDAP authentication enabled
### Environment
* Python version: 3.7.9
* NetBox version: 2.9.7

### Steps to Reproduce
1. Configure LDAP authentication, following the documentation [here](https://netbox.readthedocs.io/en/stable/installation/6-ldap/) using a Microsoft Active Directory server as the LDAP server.  Some key configurations to note:
1a.  Our `AUTH_LDAP_SERVER_URI` is an ldaps endpoint.
1b. `AUTH_LDAP_GROUP_TYPE` is set to `NestedActiveDirectoryGroupType()`
1c. `LDAP_IGNORE_CERT_ERRORS = True`
2.  Start NetBox using uwsgi
3. Wait for a while and the CPU Usage will hit near 100%.

To be more explicit about our LDAP configuration the entire contents are here with some redactions:

```
import os
import logging

from django_auth_ldap.config import LDAPGroupQuery, LDAPSearch, NestedActiveDirectoryGroupType
import ldap

# Instantiate logger
my_logger = logging.getLogger('django_auth_ldap')


# Server URI
# AUTH_LDAP_SERVER_URI = "ldaps://ad1-ldaps.example.com"
AUTH_LDAP_SERVER_URI = "ldaps://10.10.10.10"

# The following may be needed if you are binding to Active Directory.
AUTH_LDAP_CONNECTION_OPTIONS = {ldap.OPT_REFERRALS: 0}

# Set the DN and password for the NetBox service account.
AUTH_LDAP_BIND_DN = f"{os.environ.get('LDAP_USERNAME', '')}@example.com"
AUTH_LDAP_BIND_PASSWORD = os.environ.get("LDAP_PASSWORD", "")

# Include this setting if you want to ignore certificate errors. This might be needed to accept a self-signed cert.
# Note that this is a NetBox-specific setting which sets:
#     ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_NEVER)
LDAP_IGNORE_CERT_ERRORS = True

# This search matches users with the sAMAccountName equal to the provided username. This is required if the user's
# username is not in their DN (Active Directory).
AUTH_LDAP_USER_SEARCH = LDAPSearch(
    "dc=example,dc=com", ldap.SCOPE_SUBTREE, "(&(objectClass=user)(sAMAccountName=%(user)s))"
)

# If a user's DN is producible from their username, we don't need to search.
# AUTH_LDAP_USER_DN_TEMPLATE = "sAMAccountName=%(user)s,dc=example,dc=com"

# You can map user attributes to Django attributes as so.
AUTH_LDAP_USER_ATTR_MAP = {
    "first_name": "givenName",
    "last_name": "sn",
    "email": "mail",
}

# This search ought to return all groups to which the user belongs. django_auth_ldap uses this to determine group
# hierarchy.
AUTH_LDAP_GROUP_SEARCH = LDAPSearch(
    "dc=example,dc=com", ldap.SCOPE_SUBTREE, "(objectClass=group)"
)
AUTH_LDAP_GROUP_TYPE = NestedActiveDirectoryGroupType()

# Define a group required to login.
# AUTH_LDAP_REQUIRE_GROUP = "CN=Netbox-ReadOnly,OU=Resource_Groups,OU=Support - Netbox,OU=Infrastructure"

# Mirror LDAP group assignments.
AUTH_LDAP_MIRROR_GROUPS = True

# Define special user types using groups. Exercise great caution when assigning superuser status.
AUTH_LDAP_USER_FLAGS_BY_GROUP = {
    "is_active": (
        LDAPGroupQuery("CN=Netbox-ReadOnly,OU=Resource_Groups,OU=Support - Netbox,OU=Infrastructure Services,DC=example,DC=com") |
        LDAPGroupQuery("CN=Netbox-StandardAccess,OU=Resource_Groups,OU=Support - Netbox,OU=Infrastructure Services,DC=example,DC=com") |
        LDAPGroupQuery("CN=Netbox-Admins,OU=Resource_Groups,OU=Support - Netbox,OU=Infrastructure Services,DC=example,DC=com")
    ),
    "is_staff": "CN=Netbox-Admins,OU=Resource_Groups,OU=Support - Netbox,OU=Infrastructure Services,DC=example,DC=com",
    "is_superuser": "CN=DNetbox-Admins,OU=Resource_Groups,OU=Support - Netbox,OU=Infrastructure Services,DC=example,DC=com",
}

# For more granular permissions, we can map LDAP groups to Django groups.

# Cache groups for one hour to reduce LDAP traffic
AUTH_LDAP_CACHE_TIMEOUT = 3600
```

I believe the issue is from [these](https://github.com/netbox-community/netbox/blob/3047208e6a635441863a648a2a2993d9b0841ae6/netbox/netbox/authentication.py#L176-L178) lines as I have investigated several flame graphs (see #5194) and have seen the issue specifically line 177 using nearly all of the CPU time.  Moving the logging configuration for authentication.py to the LOGGING parameter in configuration.py seems to address this issue in my environment.

### Expected Behavior

NetBox should idle with very little CPU usage

### Observed Behavior

NetBox idles with near 100% CPU usage

nniehoff:
### Fixes: #5282

This PR removes the logging configuration from authentication.py and moves it to configuration.py with all other logging config.  The troubleshooting LDAP docs have also been updated to reflect the change.

### DIFF
--- a/docs/installation/6-ldap.md
+++ b/docs/installation/6-ldap.md
@@ -143,17 +143,29 @@ AUTH_LDAP_CACHE_TIMEOUT = 3600
 
 `systemctl restart netbox` restarts the Netbox service, and initiates any changes made to `ldap_config.py`. If there are syntax errors present, the NetBox process will not spawn an instance, and errors should be logged to `/var/log/messages`.
 
-For troubleshooting LDAP user/group queries, add the following lines to the start of `ldap_config.py` after `import ldap`.
+For troubleshooting LDAP user/group queries, add or merge the following [logging](/configuration/optional-settings.md#logging) configuration to `configuration.py`:
 
 ```python
-import logging, logging.handlers
 logfile = "/opt/netbox/logs/django-ldap-debug.log"
-my_logger = logging.getLogger('django_auth_ldap')
-my_logger.setLevel(logging.DEBUG)
-handler = logging.handlers.RotatingFileHandler(
-    logfile, maxBytes=1024 * 500, backupCount=5
-)
-my_logger.addHandler(handler)
+LOGGING = {
+    "version": 1,
+    "disable_existing_loggers": False,
+    "handlers": {
+        'netbox_auth_log': {
+            'level': 'DEBUG',
+            'class': 'logging.handlers.RotatingFileHandler',
+            'filename': logfile,
+            'maxBytes': 1024 * 500,
+            'backupCount': 5,
+        },
+    },
+    'loggers': {
+        'django_auth_ldap': {
+            'handlers': ['netbox_auth_log'],
+            'level': 'DEBUG',
+        },
+    },
+}
 ```
 
 Ensure the file and path specified in logfile exist and are writable and executable by the application service account. Restart the netbox service and attempt to log into the site to trigger log entries to this file.
--- a/docs/installation/6-ldap.md
+++ b/docs/installation/6-ldap.md
@@ -146,15 +146,14 @@ AUTH_LDAP_CACHE_TIMEOUT = 3600
 For troubleshooting LDAP user/group queries, add or merge the following [logging](/configuration/optional-settings.md#logging) configuration to `configuration.py`:
 
 ```python
-logfile = "/opt/netbox/logs/django-ldap-debug.log"
 LOGGING = {
-    "version": 1,
-    "disable_existing_loggers": False,
-    "handlers": {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'handlers': {
         'netbox_auth_log': {
             'level': 'DEBUG',
             'class': 'logging.handlers.RotatingFileHandler',
-            'filename': logfile,
+            'filename': '/opt/netbox/logs/django-ldap-debug.log',
             'maxBytes': 1024 * 500,
             'backupCount': 5,
         },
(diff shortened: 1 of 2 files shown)


docs/installation/6-ldap.md
```diff
@@ -143,17 +143,29 @@ AUTH_LDAP_CACHE_TIMEOUT = 3600
 
 `systemctl restart netbox` restarts the Netbox service, and initiates any changes made to `ldap_config.py`. If there are syntax errors present, the NetBox process will not spawn an instance, and errors should be logged to `/var/log/messages`.
 
-For troubleshooting LDAP user/group queries, add the following lines to the start of `ldap_config.py` after `import ldap`.
+For troubleshooting LDAP user/group queries, add or merge the following [logging](/configuration/optional-settings.md#logging) configuration to `configuration.py`:
 
 ```python
-import logging, logging.handlers
 logfile = "/opt/netbox/logs/django-ldap-debug.log"
```
lampwins, line 149: A small nitpick, but can we get rid of this variable to just set the filename directly in the dict? I think it is slightly less confusing as we don't really do this sort of thing anywhere else in configuration.py (not that users aren't free to do whatever they like that is).

nniehoff: Removed

PR #5321, opened 2020-11-08
Adds Title to Prefix Utilization to create Tooltip

### LINKED ISSUE #4611: Add tally of remaining Available IPs in web UI views
### Environment
* Python version:  3.6.8
* NetBox version:  2.7.11

Note: I'm aware that 2.7.11 is not the latest stable version, but the functionality being requested is not in the latest version of the code either.  If I need to deploy 2.8.3 to see the same and update this issue, I can do that.

### Proposed Functionality
Add a count of the number of remaining Available IPs in Prefix views (and any others that seem logical).  For instance, this could be under https://netbox.example.com/ipam/prefixes/1/ in a field, but it would also be nice to have a tally next to the Utilization percentage bar in the https://netbox.example.com/ipam/prefixes/ listing view as well.

### Use Case
Showing a prefix to a particular user or tenant with a tally of remaining available IPs would make understanding the data easier and simpler without the need to look through the Prefix definition to find out how many are available.  I understand that this is easily done in the API, but the UI is all that some of my consumers see.

### Database Changes
As far as I'm aware, this would not require a database change.

### External Dependencies
As far as I'm aware, this would not require any external dependencies.

jvanderaa:
### Fixes: #4611 

Adds a multi-line tooltip to include:
  - Addresses Used count
  - Total available addresses

- Modifies the return of the get_utilization function to include the count used, and total available calculated
- Updates the utilization utility function to pass these values into the template
- Update the template to include a title to create a tooltip

Screen Shot 2020-11-07 at 21 29 31


netbox/ipam/models.py
```diff
@@ -559,14 +559,14 @@ def get_utilization(self):
                 vrf=self.vrf
             )
             child_prefixes = netaddr.IPSet([p.prefix for p in queryset])
-            return int(float(child_prefixes.size) / self.prefix.size * 100)
+            return (int(float(child_prefixes.size) / self.prefix.size * 100), child_prefixes.size, self.prefix.size)
```
jeremystretch, line 562: Returning the numerator, denominator, and quotient is redundant: We can return just the child prefixes' size and parent prefix size, and let the caller calculate the percentage.

Also, since we're changing the return signature, be sure to update the method's docstring.

netbox/ipam/tests/test_models.py
```diff
@@ -179,7 +179,8 @@ def test_get_utilization(self):
             Prefix(prefix=netaddr.IPNetwork('10.0.0.0/26')),
             Prefix(prefix=netaddr.IPNetwork('10.0.0.128/26')),
         ))
-        self.assertEqual(prefix.get_utilization(), 50)
+        self.assertEqual(prefix.get_utilization()[0], 50)
```
jeremystretch, line 182:
```suggestion
```
This check is duplicated by L183.

netbox/ipam/tests/test_models.py
```diff
@@ -188,7 +189,8 @@ def test_get_utilization(self):
             # Create 32 IPAddresses within the Prefix
             [IPAddress(address=netaddr.IPNetwork('10.0.0.{}/24'.format(i))) for i in range(1, 33)]
         )
-        self.assertEqual(prefix.get_utilization(), 12)  # ~= 12%
+        self.assertEqual(prefix.get_utilization()[0], 12)  # ~= 12%
```
jeremystretch, line 192:
```suggestion
```
This check is duplicated by L193.

netbox/utilities/templatetags/helpers.py
```diff
@@ -234,10 +234,19 @@ def utilization_graph(utilization, warning_threshold=75, danger_threshold=90):
     """
     Display a horizontal bar graph indicating a percentage of utilization.
     """
+    if isinstance(utilization, int):
```
jeremystretch, line 237: The function's return signature needs to be consistent and predictable; that is, it should always return the same context. A better approach might be to require the numerator and denominator, and calculate the resulting percentage internally. (Note that this would entail updating all other usages of this function as well.)

netbox/templates/utilities/templatetags/utilization_graph.html
```diff
@@ -1,4 +1,4 @@
-<div class="progress text-center">
+<div title="Used: {{ util_count }}&#013Available: {{ num_available }}" class="progress text-center">
```
jeremystretch, line 1: We should only be including the tooltip if the `util_count` and `num_available` context variables are present.

netbox/utilities/templatetags/helpers.py
```diff
@@ -234,10 +234,19 @@ def utilization_graph(utilization, warning_threshold=75, danger_threshold=90):
...
+            'warning_threshold': warning_threshold,
+            'danger_threshold': danger_threshold,
+        }
+
     return {
-        'utilization': utilization,
+        'utilization': utilization[0],
         'warning_threshold': warning_threshold,
         'danger_threshold': danger_threshold,
+        'util_count': utilization[1],
```
jeremystretch, line 248: Bit of a nit, but I'd prefer to use more verbose names for these. `util_count` isn't immediately clear as to its purpose (we often use "util" for "utility"). Maybe `utilization_count` and `utilization_total`?

jeremystretch (changes requested):

netbox/utilities/templatetags/helpers.py
```diff
@@ -230,14 +230,22 @@ def querystring(request, **kwargs):
...
+        utilization = 0
+    else:
+        utilization = int(float(data_input[0]) / data_input[1] * 100)
+
     return {
         'utilization': utilization,
         'warning_threshold': warning_threshold,
         'danger_threshold': danger_threshold,
+        'utilization_count': data_input[0],
+        'num_available': data_input[1],
```
jeremystretch, line 248: Should `num_available` indicate the _total_ number of objects, or the remaining number after accounting for `utilization_count`? It's confusing because the tooltip will currently show e.g. "Available: 254" for a /24 IPv4 prefix regardless of its `utilization_count`.

jvanderaa: Yes, this is off as I look at it here again.

Should we expand this to 3 items, including with an example of Prefix from the screenshot above:

| Var Name | Represents | Example | Label |
| ---------- | ----------- | -------- | ------ |
| utilization_count | Number of used address | 21 | Used |
| *No variable*, handle with math | Number of available to be used after utilization accounted for | 233 (254-21) | Available |
| total_possible (rename from num_available) | Total addresses within the prefix | 254 | Total Count |

netbox/dcim/models/racks.py
```diff
@@ -542,9 +542,9 @@ def get_utilization(self):
                 available_units.remove(u)
 
         occupied_unit_count = self.u_height - len(available_units)
-        percentage = int(float(occupied_unit_count) / self.u_height * 100)
 
-        return percentage
+        # Return the numerator and denominator as percentage is to be calculated later where needed
+        return (occupied_unit_count, self.u_height)
```
glennmatthews, line 547: From a readability and usability perspective this seems like it'd be a great use case for `collections.namedtuple`.

netbox/utilities/templatetags/helpers.py
jeremystretch: This only needs to convey the total possible and the number consumed/created.

netbox/dcim/models/racks.py
```diff
@@ -562,8 +562,8 @@ def get_power_utilization(self):
         if power_stats:
             allocated_draw_total = sum(x['allocated_draw_total'] or 0 for x in power_stats)
             available_power_total = sum(x['available_power'] for x in power_stats)
-            return int(allocated_draw_total / available_power_total * 100) or 0
-        return 0
+            return int(allocated_draw_total / available_power_total * 100) or (0, 0)
```
jeremystretch, line 565: Different return signatures

netbox/utilities/templatetags/helpers.py
```diff
@@ -230,14 +230,22 @@ def querystring(request, **kwargs):
 
 
 @register.inclusion_tag('utilities/templatetags/utilization_graph.html')
-def utilization_graph(utilization, warning_threshold=75, danger_threshold=90):
+def utilization_graph(data_input, warning_threshold=75, danger_threshold=90):
```
jeremystretch, line 233: `data_input` should be split into two discrete arguments (utilization and total). Note that this will involve updating all usages of the tag within the project.

jeremystretch (changes requested):

candlerb:
The image shows "Used: 24, Available: 254", and I think that's ambiguous: it could imply that 254 addresses are available for allocation.

I suggest "Used: 24, Total: 254", or simply "Used: 24 / 254"

jvanderaa:
@jeremystretch @DanSheps this should be all set to go and is ready for merging.

jeremystretch:
@jvanderaa while I appreciate the work you've put into this, I can't accept the level of complexity that has been introduced merely to display a count of IP addresses. There also remain some questionable choices, such as using `namedtuple` where it's not needed and accessing elements by numeric index within a template.

It would probably make more sense to introduce a separate table column to convey this information, rather than trying to inject it into the utilization graph. That would avoid the need to intermingle logic and rendering with the utilization graph.

jvanderaa:
Agreed on the numeric index on the template. Seeing that, it is definitely a mistake there. I will get the update in that is this:

```
                    <td>Space Utilization</td>
                    <td>{% utilization_graph object.get_utilization %}</td>
                </tr>
                <tr>
                    <td>Power Utilization</td>
                    <td>{% utilization_graph object.get_power_utilization %}</td>
```

jeremystretch:
@jvanderaa the much larger issue is the overall complexity being introduced: it's just trying to do too much at once. As I said, I'm not able to accept this PR, however I sincerely appreciate the effort.

Are you interested in pursuing an alternative approach using a separate table column? If so, I'll leave #4611 assigned to you. I'm afraid we have to close out this PR though.

PR #5343, opened 2020-11-12
Update 6-ldap.md - change ldap config example

Harrison-S1:
When going through the set up I kept getting the error (from logging) "Authentication failed for USERNAME: failed to map the username to DN" I could login with the service account 'netbox' I created to bind. I changed to password of the user to match the service account and it went straight through. Looks like the bind password was being used to login and not the users password. Checked the doc on django https://django-auth-ldap.readthedocs.io/en/latest/reference.html and they have single quotes.


jeremystretch:
Python doesn't care whether strings are encapsulated in single or double quotes. There's no need for this change.

PR #5362, opened 2020-11-19
Add: USB power outlet/ports

### LINKED ISSUE #5319: Power ports for USB
### Environment
* Python version: 3.6
* NetBox version: 2.9.8

### Proposed Functionality
USB style power ports/plugs/etc

Would want the same ones offered in console ports:

USB Type A
USB Type B
USB Type C
USB Mini A
USB Mini B
USB Micro A
USB Micro B

(I'm not sure, is that actually all the USB types these days?)

### Use Case
Many smaller devices these days use things like USB-A, MicroUSB, mini-usb, USB-C etc etc for power.  While a single PI might be powered directly by a wall-wart adapter, multiple ones might be powered by something like a USB charger, which has one NEMA style plug, and then 10 USB ports. You can code this into netbox by just saying everything is type OTHER, but that's not very precise.

It's very difficult for me to accurately track IOT style devices or PI-Clusters, etc etc accurately without this.

ndom91:
### Fixes: #5319 

Adds USB Ports to Power Outlets / Power Ports


netbox/dcim/choices.py
```diff
@@ -278,6 +278,15 @@ class PowerPortTypeChoices(ChoiceSet):
     TYPE_ITA_M = 'ita-m'
     TYPE_ITA_N = 'ita-n'
     TYPE_ITA_O = 'ita-o'
+    # USB
```
jeremystretch, line 281: Minor nit, but can we keep these in the same order as `ConsolePortTypeChoices`? Or update that choice set to match this one? I don't have a preference for the order, just want them to be consistent.

netbox/dcim/choices.py
```diff
@@ -278,6 +278,15 @@ class PowerPortTypeChoices(ChoiceSet):
...
+    TYPE_USB_A = 'usb-a'
+    TYPE_USB_MINIA = 'usb-mini-a'
+    TYPE_USB_MICROA = 'usb-micro-a'
+    TYPE_USB_B = 'usb-b'
+    TYPE_USB_MINIB = 'usb-mini-b'
+    TYPE_USB_MICROB = 'usb-micro-b'
+    TYPE_USB_3_MICROB = 'usb-3-micro-b'
```
jeremystretch, line 288: Is it worth calling out USB 3.0 micro B as its own form factor? Seems like a very rare connector that might just introduce confusion. (If it _is_ worth adding, we probably need to add a USB 3.0 variant for the type B plug as well.)

jeremystretch (changes requested):

netbox/dcim/choices.py
ndom91: Yeah no problem, i'll change that now

netbox/dcim/choices.py
ndom91: The usb3.0 microB is pretty prevelant on external HDD's with USB3.0 afaik. I've seen them there a few times. Doesn't hurt to include it, right? I'll add the usb3.0 type B variant then as well :+1:

PR #5434, opened 2020-12-09
Added rack_elevation.show_images to user.preferences

### LINKED ISSUE #5203: Remember user preference to show/hide device type images in rack views
### Environment
* Python version:  3.6.8
* NetBox version:  2.9.3

### Proposed Functionality
It would be nice if the value of the "Show images" setting on a rack view could be persisted as a user preference (using the mechanism introduced in #3294).

### Use Case
We are making use of device type front/rear images on most of our devices in Netbox. By default, Netbox will render these images in a rack view, except when the user toggles the "Show images" button.
However, this is only valid for the current view -- as soon as another rack view is opened, or the current page is refreshed, Netbox falls back to the default mode, i.e. showing images.

Some users may not see any use in the images, and prefer to always see the name and device type color instead. Having a dedicated user preference key for this setting would allow for this functionality.

### Database Changes

none
### External Dependencies
none

ogorov:
### Fixes: [5203](https://github.com/netbox-community/netbox/issues/5203)

- Changed rack_elevation.js to show/hide images in case of preference is set and to send preference to server when user toggle it
- Added post method to RackElevationListView (I can't find better place to add method that set new preference)
- Made toggle-button glyph and attributes dependent on new property


ogorov:
I don't sure if it matches the style of NetBox. But I'm ready to make changes if in will needed.

netbox/dcim/views.py
```diff
@@ -341,8 +342,21 @@ def get(self, request):
             'reverse': reverse,
             'rack_face': rack_face,
             'filter_form': forms.RackElevationFilterForm(request.GET),
+            'show_images': request.user.config.get('rack_elevation.show_images', False)
```
jeremystretch, line 345: Images should be enabled by default to preserve the existing behavior.

netbox/dcim/views.py
```diff
@@ -341,8 +342,21 @@ def get(self, request):
...
         })
 
+    def post(self, request):
```
jeremystretch, line 348: Any such dynamic updates would need to use the REST API. (Adding individual view methods for all of the various settings does not scale.) Unfortunately, we don't yet have a REST API endpoint for user preferences. I believe this work is blocked until one can be introduced.

netbox/dcim/views.py
ogorov: It will be easy to change

netbox/dcim/views.py
ogorov: Has a suitable problem been created or do I need to create a new one?

jeremystretch: The upcoming v2.10 release introduces an endpoint for this. If you want, you can rebase this PR off the `develop-2.10` branch and adapt to use the new endpoint, or just wait for the release (which should be sometime next week).

ogorov: OK, I'll do something about it soon.

jeremystretch:
With v2.10 released, the `/api/users/config/` endpoint is now available for use.

ogorov:
> With v2.10 released, the `/api/users/config/` endpoint is now available for use.

I tried to set preference with new api endpoint. But I can't set values as a hierarchical dictionary structure like other preferences. This is because PATCH method uses the deepmerge function, which adds the raw received params to the user preferences. The raw data can only be a string, but I want to set dictionary. 
There is also no protection from passing bad data (non-existent field) to the user preferences. 
Is changing the logic of the PATCH method a good idea or am I missing something?

ogorov:
Are there any comments about this pull request?

jeremystretch:
It seems like this has gotten pretty far off track. Maybe it makes sense to defer this work until the user preferences API endpoint becomes a bit more mature. We shouldn't be trying to address the limitations you cite above as part of this FR.

netbox/dcim/views.py
DanSheps: Do you have an update for this?

ogorov:
Hi
Is there any sense on this issue?

jeremystretch:
Looks like this should have been closed out a while back. As noted above, we need more work around the preferences architecture before we consider extending it any further. Additionally, the UI has been completed reworked for v3.0.

PR #5440, opened 2020-12-11
Adds support for multi select custom fields

TomGrozev:
### Fixes: #5451
Since in 2.10 custom fields are now represented as JSON, it is quite easy to allow multiple selection option custom fields. Before the JSON change, this feature would have required a many-to-many relationship.

(This may have already been planned but I saw it could be done so I did it.)

For CSV importing the field, I'm not too sure how feasibile it would be for multiple choices, looking for some suggestions on this.

### DIFF
--- a/docs/release-notes/version-2.10.md
+++ b/docs/release-notes/version-2.10.md
@@ -5,6 +5,7 @@
 ### Enhancements
 
 * [#5411](https://github.com/netbox-community/netbox/issues/5411) - Include cable tags in trace view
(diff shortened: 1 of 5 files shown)


jeremystretch:
Both of the referenced issues have been closed. Per the [contributing guide](https://github.com/netbox-community/netbox/blob/develop/CONTRIBUTING.md), please open a new feature request for the proposed work. No pull requests can be accepted until the feature request has been approved.

PR #5459, opened 2020-12-14
Adds support for multi select custom fields

### LINKED ISSUE #5451: Custom field multiple selection
### Environment
* Python version: 3.8.6
* NetBox version: 2.10

### Proposed Functionality
The ability to be able to select multiple values in custom fields of type "Selection". This is similar to how tags work but the ability to do it for custom selection fields. A new option to allow multiple options to be selected for custom selection fields would be added and using the Django multiple selection form, this can be added.

### Use Case
We use ansible to deploy configurations and NetBox as the source of truth. We need to be able to select multiple applications for a virtual machine and then ansible can use that to deploy the application. In order to do this and install multiple applications, we need to be able to select multiple values from a custom field of type "Selection".

### Database Changes
A new multiple selection field would be added to the custom field model. Since the changes in version 2.10
 store custom field data as JSON, storing a list of values is simple.

### External Dependencies
None

TomGrozev:
### Fixes: #5451

Since in 2.10 custom fields are now represented as JSON, it is quite easy to allow multiple selection option custom fields. Before the JSON change, this feature would have required a many-to-many relationship.

For CSV importing the field, I'm not too sure how feasibile it would be for multiple choices, looking for some suggestions on this.

### DIFF
--- a/docs/release-notes/version-2.10.md
+++ b/docs/release-notes/version-2.10.md
@@ -62,6 +62,7 @@ All end-to-end cable paths are now cached using the new CablePath backend model.
 
 ### Enhancements
 
+* [#5451](https://github.com/netbox-community/netbox/issues/5451) - Add ability to create multi selection custom fields
 * [#609](https://github.com/netbox-community/netbox/issues/609) - Add min/max value and regex validation for custom fields
 * [#1503](https://github.com/netbox-community/netbox/issues/1503) - Allow assigment of secrets to virtual machines
 * [#1692](https://github.com/netbox-community/netbox/issues/1692) - Allow assigment of inventory items to parent items in web UI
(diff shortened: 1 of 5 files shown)


jeremystretch:
Thank you for your contribution, however we are not prepared to accept this work at this time. Note that #5451 has been designated as "needs milestone:" a complete implementation still needs to be identified and scheduled among other initiatives for an upcoming release. Because this feature requires an API change, it must be bundled in a major release, the earliest of which would be v2.11.

TomGrozev:
I don't understand what the required API change is, it is a simple change and not complex.

jeremystretch:
Adding a new type of custom field changes the API signature, which forces a version increment.

PR #5495, opened 2020-12-18
Fixes: #5301 - Add positions when rack-face is not set

### LINKED ISSUE #5301: Misleading error when trying to rack device from non-racked state when rack face not specified
### Environment
* Python version: Python 3.6.8
* NetBox version: 2.9.7

### Steps to Reproduce
1. Create a device and under Location only select the Site and Rack. Leave all other fields empty or null (------). You should now have a non-racked device assigned to the rack.
2. Edit the device and select an available Position in the rack. Click Update
3. Observe error saying the rack unit is not available ```The lowest-numbered unit occupied by the device
Select a valid choice. 16 is not one of the available choices.```

### Expected Behavior
I expected an error highlighting the rack face field. Similar to changing a racked device's position but nulling out the rack face field. ```Must specify rack face when defining rack position.```

Screen Shot 2020-11-02 at 11 23 17 AM

### Observed Behavior
Netbox gave an error saying my RU was not available but really it was erroring because I did not select a Rack face now that I am trying to rack the device coming from a non-racked state.

Screen Shot 2020-11-02 at 11 23 43 AM

DanSheps:
### Fixes: #5301

When rack face is not set, but positions are, this will result in a position_choices list being generated with a union between front and rear position choices.

### DIFF
--- a/netbox/dcim/forms.py
+++ b/netbox/dcim/forms.py
@@ -27,6 +27,7 @@ from utilities.forms import (
     NumericArrayField, SelectWithPK, SmallTextarea, SlugField, StaticSelect2, StaticSelect2Multiple, TagFilterField,
     BOOLEAN_WITH_BLANK_CHOICES,
 )
+from utilities.utils import union_list
 from virtualization.models import Cluster, ClusterGroup
 from .choices import *
 from .constants import *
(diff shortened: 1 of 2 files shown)


jeremystretch:
Can we just add a check under `DeviceForm.clean()` to check that `face` is specified if `position` is set? Admittedly it's redundant to the model's validation but seems like the simplest fix given the nature of position assignment.

DanSheps:
Sorry, I did test this, with using clean, since position is a null array it unsets position for clean() as there is no valid position

jeremystretch:
@DanSheps thanks for your work on this. Ultimately I was able to resolve it in fce6129 by changing `position` to an IntegerField, which also allowed me to accomplish a bit of overdue cleanup in the form logic (from before we had the APISelect widget).

Issue #5506, opened 2020-12-20
Angle brackets in markdown


DeemGit:
### Environment
* Python version:  3.9
* NetBox version:  2.10.1

### Steps to Reproduce
1.  Create/edit a device
2.  Add comment with markdown syntax, for e.g.: 
`External Link <http://google.com>`

### Expected Behavior
Markdown should render `<http://google.com>` to `<a href="http://google.com">http://google.com</a>`

### Observed Behavior
No link rendered.

### Caused by
netbox/utilities/templatetags/helpers.py:
def render_markdown(value):
...
    # Strip HTML tags
    value = strip_tags(value)

`strip_tags` cuts  any angle brackets despite on valid markdown syntax.

marcus-crane:
Hey DeemGit,

The standard Markdown format for links looks like this: `[link title](https://example.com)` which would render the following HTML: `<a href="https://example.com">link title</a>`

While I believe it's a feature of Github Flavoured Markdown to turn any plain text URL into a link automatically, that is a custom extension that isn't part of the standard Markdown specification

I just tested it on our internal instance though and using the above `[x](y)` format does generate a link within the device comments

PR #5508, opened 2020-12-21
Fix for #5506

DeemGit:
Corrects behavior of angle brackets in markdown.


jeremystretch:
Rewriting one form of Markdown into another should be considered an anti-pattern and avoided wherever possible as it's very likely to induce future bugs. As discussed in #5506, a workaround to the proposed behavior already exists, and I don't see sufficient justification for further work in this area.

PR #5632, opened 2021-01-17
Implemented WebRequestContext for emulating HTTP requests for ORM operations

### LINKED ISSUE #4932: Add support for Webhooks when accessing objects via the ORM
### Environment
* Python version:  3.7.7
* NetBox version: 2.8.8

### Proposed Functionality
Support a solution to be able to generate Webhooks when objects are changed via the ORM (Create/update/delete). Either automatically or manually.
As I understand, currently webhooks are managed in the class ObjectChangeMiddleware that is only executed in the context of a HTTP request. 

### Use Case
Support Webhooks for all changes done within scripts & plugins executed asynchronously in a worker. (#2006)

### Database Changes
No change to the database

### External Dependencies
I don't think so

lampwins:
### Fixes: #4932

This implements a new `WebRequestContext` context manager for emulating HTTP requests for ORM use. The primary need is for easily enabling change logging and webhooks when interacting with the ORM directly, and outside of the context of an HTTP request.

You will likely note the use of `RequestFactory` for creating the "fake" request object. I feel this is the safest way of doing this because it ensures we don't break anything in the future if we require more context from the request than just the user. When looking at the internals of `RequestFactory`, there is nothing specific to its primary usage in testing that should cause any issues here. I started down the path of doing pretty much the same thing it does to create a bare-bones `WSGIRequest` object and then decided this would be a better, cleaner approach.

### DIFF
--- a/docs/administration/netbox-shell.md
+++ b/docs/administration/netbox-shell.md
@@ -201,3 +201,19 @@ To delete multiple objects at once, call `delete()` on a filtered queryset. It's
 
 !!! warning
     Deletions are immediate and irreversible. Always consider the impact of deleting objects carefully before calling `delete()` on an instance or queryset.
+
+## Change Logging and Webhooks
+
+Note that NetBox's change logging and webhook processing features operate under the context of an HTTP request. As such, these functions do not work automatically when using the ORM directly, either through the nbshell or otherwise. A special context manager is provided to allow these features to operate under an emulated HTTP request context. This context manager must be explicitly invoked for change log entries and webhooks to be created when interacting with objects through the ORM. Here is an example using the `WebRequestContext` context manager within the nbshell:
+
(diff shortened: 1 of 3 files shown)


jeremystretch:
I like it, but why create a separate class from the `change_logging` context manager as opposed to replacing it? We should be able to have ObjectChangeMiddleware call WebRequestContext the same way a user would on the console.

lampwins:
Sure, makes sense. I think to make that work, I would make `user` a kwarg with a default of `None` in addition to `request` and then enforce a check inside the function to make sure either user _or_ a request object with a user set has been passed. This way we don't have to do any preprocessing to satisfy the existing signature `(user, request=None)` within `ObjectChangeMiddleware` and elsewhere internally.

PR #5670, opened 2021-01-22
Proposal to add QSFP+ (64GFC) FibreChannel Interface option re FR #5630

### LINKED ISSUE #5630: Request to add QSFP+ (64GFC) interface type option
### Environment
* Python version: 3.8.3
* NetBox version: 2.10.3

### Proposed Functionality
Request is to add the "QSFP+ (64GFC)" interface type as an option.

### Use Case
These interface types are used for Inter-Chassis Link connections for the Brocade DCX SAN directors.
These are 4x16G FC bundled adapters, like the QSFP28 (128GFC) are for 4x32G FC ones which are available in the application.
Here is an example of the part number and description:
57-0000090-01 | QSFP+ SWL 4x16G FC (64G)

### Database Changes
I don't believe this should require database changes.

### External Dependencies
None

lastwednesday:
### Fixes: #5630 

This proposal will add the QSFP+ (64GFC) interface type for FibreChannel into the netbox/dcim/choices.py script

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -686,7 +686,7 @@ class InterfaceTypeChoices(ChoiceSet):
     TYPE_8GFC_SFP_PLUS = '8gfc-sfpp'
     TYPE_16GFC_SFP_PLUS = '16gfc-sfpp'
     TYPE_32GFC_SFP28 = '32gfc-sfp28'
-    TYPE_64GFC_SFP_PLUS = '64gfc-sfpp'
+    TYPE_64GFC_QSFP_PLUS = '64gfc-sfpp'
     TYPE_128GFC_QSFP28 = '128gfc-sfp28'
 
     # InfiniBand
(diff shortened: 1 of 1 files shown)


netbox/dcim/choices.py
```diff
@@ -686,6 +686,7 @@ class InterfaceTypeChoices(ChoiceSet):
     TYPE_8GFC_SFP_PLUS = '8gfc-sfpp'
     TYPE_16GFC_SFP_PLUS = '16gfc-sfpp'
     TYPE_32GFC_SFP28 = '32gfc-sfp28'
+    TYPE_64GFC_QSFP_PLUS = '64gfc-sfpp'
```
jeremystretch, line 689: The slug for this value should indicate QSFP+, not SFP+.

lastwednesday: Ok I will adjust that. I was following the example of the existing 128GFC QSFP28. Should that one be updated as well?

Currently it is:
TYPE_128GFC_QSFP28 = '128gfc-sfp28'

jeremystretch: Do **not** alter any other values, as that would break existing installations.

lastwednesday: I have updated the slug to be 64gfc-qsfpp now as mentioned.

PR #5672, opened 2021-01-22
Indicate when cable length is not definitive

### LINKED ISSUE #5650: Cable length calculation incorrect when segment lengths are missing
### Environment
* Python version: 2.10.3
* NetBox version: 3.9

### Steps to Reproduce
1. Create two devices
2. Create two 'patch panel' devices with front and rear ports
3. Connect the rear ports of the patch panels without specifying the length
4. Connect the two devices to each of the patch panels and specify the length in one or both
5. Trace the cable between the devices

### Expected Behavior
Ideally, the total length should indicate that it is 'at least' n meters long or in some better way indicate that only some of the lengths were populated and therefore the calculated length is somewhat short of the actual.

### Observed Behavior
With only some of the cable lengths specified - perhaps the patch cables between the patch panels and the devices and not the length of the horizontal cable between the patch panels, NetBox outputs a calculated length which is the correct sum of the lengths that it has available in the database. This is obviously incorrect due to missing data, but no indication of this is given to the user who might (not unreasonably) think that the total length as calculated by NetBox is correct.

cpmills1975:
indicate to users whether the cable length is
accurate or not.

### Fixes: #5650

CablePath.get_total_length() retrieves Cable objects where _abs_length is not null as well as the id being in the list of IDs for the path. If the number of cables returned is equal to the number of cable IDs in the path, all cables have a length and the length is determined to be definitive.

The total length and whether it is definitive is returned in a tuple.

PathTraceView.get_extra_context handles the new is_definitive flag and returns that to the template.

The cable_trace template checks whether is_definitive is false and if it is, prefixes the total length with a greater than sign indicting that the length of the cable is at least as long as the length shown.

When cable length is definitive:

Screenshot 2021-01-22 at 16 57 44
 
When one or more cables have no length defined:

Screenshot 2021-01-22 at 16 58 08

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -480,13 +480,17 @@ class CablePath(models.Model):
 
     def get_total_length(self):
         """
-        Return the sum of the length of each cable in the path.
+        Return a tuple containing the sum of the length of each cable in the path
+        and a flag indicating whether the length is definitive.
         """
         cable_ids = [
             # Starting from the first element, every third element in the path should be a Cable
             decompile_path_node(self.path[i])[1] for i in range(0, len(self.path), 3)
         ]
-        return Cable.objects.filter(id__in=cable_ids).aggregate(total=Sum('_abs_length'))['total']
+        cables = Cable.objects.filter(id__in=cable_ids, _abs_length__isnull=False)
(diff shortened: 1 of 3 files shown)


cpmills1975:
PR raised.

PR #5704, opened 2021-01-27
Fixes: #5315 - Make "null_option" on DynamicModelChoiceField also null the value on the model.

### LINKED ISSUE #5315: Cannot unassign a site from a vlan
### Environment
* Python version: 3.6.9
* NetBox version:  2.9.8

### Steps to Reproduce
1. Create a VLAN, say ID "123" name "foo" and save. (No site assigned)
2. Edit the VLAN. Set the site to some existing site, and save.
3. Edit the VLAN. Set the site to "None" and save.

### Expected Behavior

To be able to set the site to "None"

### Observed Behavior

After selecting "None" and clicking Save, the update is rejected:

image

DanSheps:
### Fixes: #5315

Alter the DynamicModelChoiceField to remove any value that is a string of 'null' if the 'null_option' and then call the parent clean() method.


netbox/utilities/forms/fields.py
```diff
@@ -355,7 +355,15 @@ class DynamicModelChoiceField(DynamicModelChoiceMixin, forms.ModelChoiceField):
...
     rendered only with choices set via bound data. Choices are populated on-demand via the APISelect widget.
     """
-    pass
+
+    def clean(self, value):
+        """
+        When null option is enabled and "None" is sent as part of a form to be submitted, it is sent as the
+        string 'null'.  This will check for that condition and gracefully handle the conversion to a NoneType.
+        """
+        if self.null_option is not None and value == 'null':
```
jeremystretch, line 364: We should check `value` against `settings.FILTERS_NULL_CHOICE_VALUE` rather than replicating the string for consistency.

jeremystretch (changes requested):
Other than my one note I think this is fine. It feels a _bit_ hacky, but it's certainly preferable to overhauling the choice widgets at the moment.

jeremystretch (approved):

PR #5764, opened 2021-02-07
Closes #5503: ISO 8601 date in UI and alternative format as tooltip

### LINKED ISSUE #5503: Default to ISO 8601 date format in UI and add "June 26, 2016" as tooltip
### Environment
* Python version: 3.7.3
* NetBox version: 2.10.1

### Proposed Functionality

The current approach seems to be to use ISO 8601

```python
SHORT_DATE_FORMAT = 'Y-m-d'          # 2016-06-26
```

where the logical order matters, for example in the changelog view that is sorted by time. For "free text" dates,

```python
DATE_FORMAT = 'N j, Y'               # June 26, 2016
```

is used. I am proposing to use `SHORT_DATE_FORMAT` everywhere in the UI except in tooltips. And then augment dates as tooltip in `DATE_FORMAT`. This is because ISO 8601 is *the* standard for dates in written form. If the user ever needs to communicate a date via voice, they are expected to translate 2016-06-26 in their head or peek in the tooltip and say "June 26, 2016".

@jeremystretch commented on ISO 8601 in https://github.com/netbox-community/netbox/issues/1412#issuecomment-321318039 already so I assume there is some background to the current default. I still would like to discuss this.

### Use Case

Consistency.

### Database Changes

No.

### External Dependencies

No.

ypid:
### Fixes: #5503

With this PR all dates in the UI are now consistently displayed.

I changed the long date format as suggested by @xkilian and confirmed by my own research.

* DATETIME_FORMAT
  * Before: July 20, 2020 4:52 p.m.
  * Now: 20th July, 2020 16:52

"20th July, 2020" would be spoken as "the 20th of July, 2020" but the "the" and "of" are never written.

I also switch from the illogical a.m./p.m. time format to military time format that can be pronounced as "sixteen hundred and fifty-two".

### DIFF
--- a/netbox/templates/base.html
+++ b/netbox/templates/base.html
@@ -67,7 +67,7 @@
                     <p class="text-muted">{{ settings.HOSTNAME }} (v{{ settings.VERSION }})</p>
                 </div>
                 <div class="col-xs-4 text-center">
-                    <p class="text-muted">{% now 'Y-m-d H:i:s T' %}</p>
+                    <p class="text-muted">{% annotated_now %} {% now 'T' %}</p>
                 </div>
                 <div class="col-xs-4 text-right noprint">
                     <p class="text-muted">
--- a/netbox/templates/generic/object.html
+++ b/netbox/templates/generic/object.html
@@ -43,7 +43,7 @@
   <p>
     <small class="text-muted">
       Created {{ object.created|annotated_date }} &middot;
-      Updated <span title="{{ object.last_updated|date:'SHORT_DATETIME_FORMAT' }} ({{ object.last_updated|date:'DATETIME_FORMAT' }})">{{ object.last_updated|timesince }}</span> ago
+      Updated {{ object.last_updated|annotated_date }} ({{ object.last_updated|timesince }} ago)
     </small>
     <span class="label label-default">{{ object|meta:"app_label" }}.{{ object|meta:"model_name" }}:{{ object.pk }}</span>
   </p>
--- a/netbox/templates/generic/object.html
+++ b/netbox/templates/generic/object.html
@@ -42,8 +42,8 @@
   <h1 class="title">{% block title %}{{ object }}{% endblock %}</h1>
   <p>
     <small class="text-muted">
-      Created {{ object.created }} &middot;
-      Updated <span title="{{ object.last_updated }}">{{ object.last_updated|timesince }}</span> ago
+      Created {{ object.created|annotated_date }} &middot;
+      Updated <span title="{{ object.last_updated|date:'SHORT_DATETIME_FORMAT' }} ({{ object.last_updated|date:'DATETIME_FORMAT' }})">{{ object.last_updated|timesince }}</span> ago
     </small>
     <span class="label label-default">{{ object|meta:"app_label" }}.{{ object|meta:"model_name" }}:{{ object.pk }}</span>
   </p>
--- a/netbox/templates/users/userkey.html
+++ b/netbox/templates/users/userkey.html
@@ -22,7 +22,7 @@
         <p>
             <small class="text-muted">
               Created {{ object.created|annotated_date }} &middot;
-              Updated <span title="{{ object.last_updated|date:'SHORT_DATETIME_FORMAT' }} ({{ object.last_updated|date:'DATETIME_FORMAT' }})">{{ object.last_updated|timesince }}</span> ago
+              Updated {{ object.last_updated|annotated_date }} ({{ object.last_updated|timesince }} ago)
         </p>
         {% if not object.is_active %}
             <div class="alert alert-warning" role="alert">
--- a/netbox/templates/users/userkey.html
+++ b/netbox/templates/users/userkey.html
@@ -1,4 +1,5 @@
 {% extends 'users/base.html' %}
+{% load helpers %}
 
 {% block title %}User Key{% endblock %}
 
@@ -19,7 +20,9 @@
             {% endif %}
         </h4>
         <p>
-            <small class="text-muted">Created {{ object.created }} &middot; Updated <span title="{{ object.last_updated }}">{{ object.last_updated|timesince }}</span> ago</small>
+            <small class="text-muted">
+              Created {{ object.created|annotated_date }} &middot;
+              Updated <span title="{{ object.last_updated|date:'SHORT_DATETIME_FORMAT' }} ({{ object.last_updated|date:'DATETIME_FORMAT' }})">{{ object.last_updated|timesince }}</span> ago
         </p>
         {% if not object.is_active %}
             <div class="alert alert-warning" role="alert">
@@ -37,7 +40,7 @@
                 </a>
             </div>
             <h4>Session key: <span class="label label-success">Active</span></h4>
-            <small class="text-muted">Created {{ object.session_key.created }}</small>
+            <small class="text-muted">Created {{ object.session_key.created|annotated_date }}</small>
         {% else %}
             <h4>No active session key</h4>
         {% endif %}
--- a/netbox/utilities/templatetags/helpers.py
+++ b/netbox/utilities/templatetags/helpers.py
@@ -162,10 +162,6 @@ def annotated_date(date_value):
     if not date_value:
         return ''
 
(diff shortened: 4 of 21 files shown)


DanSheps:
There seems to be a lot of repetition here, do you think it might be possible to create an include and pass the date, it would also make future modifications easier.

ypid:
Thanks for the pointer (am was/am new to Django), done.

The only place (that I know of, and I also greped the source and so on) with dates without tooltip is `object_list.html`. I tried it but there it does not work so easily because the dates are passed to Jinja as SafeString. Also, I am not sure if it would be good UI design to have tooltips in the list.

ypid:
I rebased and updated this patch to the latest develop branch. Some new date fields were introduced with NetBox 2.11. I hope I found all of them by greping the source. If not, it is not the end of the world. I will come across them eventually and fix them (potentially in a new PR). I applied this patch (including the dependency patch 2618dde1e2837fd4c587bed89d2358199176f01e) on my instance now for testing during regular use. I tested all modified pages manually already. Feel free to have a look.

ypid:
I have been using this ever since and did not run into dates that I missed nor other issues. I would say this is ready. @jeremystretch Can you have a look?

jeremystretch:
@ypid Adding inclusion tags to display a value (i.e. `{% include 'inc/date_now_span.html' %}`) should be considered poor practice. Instead, we should use a template filter for this (e.g. `{{ mydate|with_current_time }}`). We should also only need the one filter; the `parse_date()` and `datatype_str()` are unnecessary since we can process the value directly.

ypid:
I improved it, thanks. I agree with you that this approach is better. Now with the `date_span` filter, it is very easy to just append that if new date files are added.

netbox/netbox/configuration.example.py
```diff
@@ -261,7 +261,7 @@
 # https://docs.djangoproject.com/en/stable/ref/templates/builtins/#date
 DATE_FORMAT = 'N j, Y'
 SHORT_DATE_FORMAT = 'Y-m-d'
-TIME_FORMAT = 'g:i a'
+TIME_FORMAT = 'H:i:s'
```
jeremystretch, line 264: Why have you changed the default formats?

netbox/templates/base.html
```diff
@@ -67,7 +67,7 @@ <h4><i class="mdi mdi-alert"></i> Maintenance Mode</h4>
                     <p class="text-muted">{{ settings.HOSTNAME }} (v{{ settings.VERSION }})</p>
                 </div>
                 <div class="col-xs-4 text-center">
-                    <p class="text-muted">{% now 'Y-m-d H:i:s T' %}</p>
+                    <p class="text-muted">{% now_span %} {% now 'T' %}</p>
```
jeremystretch, line 70: The template tag should be able to render the entire date string.

netbox/utilities/templatetags/helpers.py
```diff
@@ -137,6 +139,39 @@ def tzoffset(value):
     return datetime.datetime.now(value).strftime('%z')
 
 
+@register.filter(expects_localtime=True)
+def date_span(date_value):
```
jeremystretch, line 143: I thought I mentioned this previously, but maybe not. Let's avoid the term "span" in the filter name, since it naturally implies a span of time between a start and end time (rather than an HTML `<span>` element). `annotated_date` maybe?

netbox/utilities/templatetags/helpers.py
```diff
@@ -137,6 +139,39 @@ def tzoffset(value):
...
+    """
+    Returns date as HTML span with short date format as the content and the
+    (long) date format as the title.
+    """
+    if not date_value:
+        return ''
+
+    if type(date_value) == str:
```
jeremystretch, line 151: I wouldn't think we'd ever want to apply this tag to a string; we should always be working with datetime objects, no?

jeremystretch (changes requested):

netbox/templates/base.html
ypid: The problem with the old code was that the date format was hard coded. I decided to use two template tags to respect the configured date format while preserving the time zone which is not intended to be part of the configurable date format strings. What do you have in mind? Do you want to keep the hard coded format?

netbox/utilities/templatetags/helpers.py
ypid: Here the date is a string: https://github.com/netbox-community/netbox/blob/develop/netbox/templates/inc/custom_fields_panel.html

netbox/utilities/templatetags/helpers.py
ypid: Makes sense. I appreciate that you care about such things!

netbox/netbox/configuration.example.py
ypid: I separated this change into it’s own commit to describe my intend:

> a.m./p.m. is illogical. Military time format can be pronounced as "sixteen hundred and fifty-two".

But that is only the format that appears on hover now. Should I drop that change or the whole commit? Are you fine with `SHORT_DATETIME_FORMAT = 'Y-m-d H:i:s'`?

netbox/utilities/templatetags/helpers.py
jeremystretch: It'd be better to have the custom field return a date object than to accommodate string values in the template filter. Let's just omit custom field dates for now to keep this from getting any more complex.

netbox/netbox/configuration.example.py
jeremystretch: Let's just leave the defaults in the example alone. If there's a case to change them, it should be addressed under a separate issue.

netbox/templates/users/userkey.html
```diff
@@ -19,7 +20,9 @@ <h4>
             {% endif %}
         </h4>
         <p>
-            <small class="text-muted">Created {{ object.created }} &middot; Updated <span title="{{ object.last_updated }}">{{ object.last_updated|timesince }}</span> ago</small>
+            <small class="text-muted">
+              Created {{ object.created|annotated_date }} &middot;
+              Updated <span title="{{ object.last_updated|date:'SHORT_DATETIME_FORMAT' }} ({{ object.last_updated|date:'DATETIME_FORMAT' }})">{{ object.last_updated|timesince }}</span> ago
```
jeremystretch, line 25:
```suggestion
              Updated {{ object.last_updated|annotated_date }} ({{ object.last_updated|timesince }} ago)
```

netbox/templates/generic/object.html
```diff
@@ -42,8 +42,8 @@
   <h1 class="title">{% block title %}{{ object }}{% endblock %}</h1>
   <p>
     <small class="text-muted">
-      Created {{ object.created }} &middot;
-      Updated <span title="{{ object.last_updated }}">{{ object.last_updated|timesince }}</span> ago
+      Created {{ object.created|annotated_date }} &middot;
+      Updated <span title="{{ object.last_updated|date:'SHORT_DATETIME_FORMAT' }} ({{ object.last_updated|date:'DATETIME_FORMAT' }})">{{ object.last_updated|timesince }}</span> ago
```
jeremystretch, line 46:
```suggestion
      Updated {{ object.last_updated|annotated_date }} ({{ object.last_updated|timesince }} ago)
```

netbox/utilities/templatetags/helpers.py
```diff
@@ -4,8 +4,10 @@
 
 import yaml
 from django import template
+from django.template.defaultfilters import date
```
jeremystretch, line 7: This looks to be unused?

netbox/netbox/configuration.example.py
ypid: I dropped the commit.

netbox/utilities/templatetags/helpers.py
ypid: It is used by this line for example:

```Python
long_ts = date(date_value, 'DATE_FORMAT')
```

netbox/utilities/templatetags/helpers.py
ypid: Fine for me, done.

ypid:
Done.

PR #5787, opened 2021-02-10
Update /netbox/tenancy/api/serializers.py to allow nullable group

### LINKED ISSUE #5786: Error while setting tenant group to null via API
### Environment
* Python version: v3.6.8
* NetBox version: v2.10.4

Attempting to save a tenant object via the API with the group property set to null results in an error.  The Swagger documentation specifies that this property is nullable.

### Steps to Reproduce
1. Send the following JSON body via PATCH method to $netbox_url/api/tenancy/tenants/$tenant_id/ :
`{
      "group": null
}`

### Expected Behavior
200 OK response and a JSON object containing the updated tenant object.

### Observed Behavior
400 Bad Request response and an error message:
`{
  "group": [
    "This field may not be null."
  ]
}`

pgnuta:
Group should be nullable via API to match frontend functionality and Swagger documentation.

### Fixes: #5786
Added 'allow_null=True' to group property definition in /netbox/tenancy/api/serializers.py

### DIFF
--- a/netbox/tenancy/api/serializers.py
+++ b/netbox/tenancy/api/serializers.py
@@ -24,7 +24,7 @@ class TenantGroupSerializer(ValidatedModelSerializer):
 
 class TenantSerializer(TaggedObjectSerializer, CustomFieldModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='tenancy-api:tenant-detail')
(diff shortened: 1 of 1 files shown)


DanSheps:
Could you add some tests for this?

jeremystretch:
I'll add a test.

Issue #5943, opened 2021-03-08
Template tags do not take into account plugins


DanSheps:
### Proposed Changes

Wrap reverse() in try/catch statement to test against both plugins and base namespace.

### Justification

Currently, these button tags do not test against the plugins namespace.  This would add testing against the namespace with and preserve the original.

Sample code:

```
    try:
        url = reverse(viewname, kwargs={'pk': instance.pk})
    except NoReverseMatch as e:
        try:
            url = reverse(f'plugins:{viewname}', kwargs={'pk': instance.pk})
        except NoReverseMatch:
            raise NoReverseMatch(e)
```

jeremystretch:
I don't follow. Can you point to an example where this is breaking something in NetBox? IMO it's bad practice to try and "brute force" the URL resolution: The tag should know when it's being called what URL it intends to resolve.

DanSheps:
An example would be ButtonColumn where you resolve the URL automatically for edit/delete/changelog, the button does not take into account the fact that a URL might be in a sub-namespace (plugins:xxx for example) as the resolution is simply `f'{model._meta.app_label}:{model._meta,model_name}'`

jeremystretch:
These tags aren't intended for use within plugins, so we don't check the plugins space. Supporting the use of these internal components within a plugin is a much greater undertaking, involving formally extending the plugins API and documenting their usage.

PR #5958, opened 2021-03-10
Closes: #5943 - Takes plugin namespace into account in certain template tags

DanSheps:
### Closes: #5943

This PR will take into account plugin namespaces for templatetags, specifically buttons.

### DIFF
--- a/netbox/utilities/templatetags/helpers.py
+++ b/netbox/utilities/templatetags/helpers.py
@@ -4,6 +4,7 @@ import re
 
 import yaml
 from django import template
+from django.apps import apps
 from django.conf import settings
 from django.urls import NoReverseMatch, reverse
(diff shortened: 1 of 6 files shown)


netbox/utilities/templatetags/helpers.py
```diff
@@ -13,9 +14,21 @@
...
+
 register = template.Library()
 
 
+def _resolve_namespace(instance):
+    """
+    Get the appropriate namespace for the app based on whether it is a Plugin or base application
+    """
+    app = apps.get_app_config(instance._meta.app_label)
+    if isinstance(app, PluginConfig):
```
DanSheps, line 27: Perhaps I should check if this is a base app as well.  Because this could allow people to override base app behaviour, thoughts?

Issue #5893, opened 2021-02-27
Overhaul UI and upgrade to Bootstrap 5


jeremystretch:
### Proposed Changes
NetBox currently uses the outdated Bootstrap 3 CSS/JS framework for its user interface. [Bootstrap 5](https://getbootstrap.com/) is currently in late beta and should be released in the near future. This issue proposes adopting BS5, and in the process giving the entire UI a substantial refresh.

However, as @checktheroads notes in a [recent discussion](https://github.com/netbox-community/netbox/discussions/5782#discussioncomment-407811), it's worth noting that BS5 explicitly [drops support for Internet Explorer](https://getbootstrap.com/docs/5.0/getting-started/browsers-devices/) (in favor of Edge). If the community is opposed to this, we can consider adopting BS4 instead.

If accepted, I am willing to take the lead on this initiative. I've already been experimenting with some layouts and templates offline, and am encouraged by my progress thus far.

### Justification
- Modernizes the UI
- BS5 introduces a suite of utility classes that allows for more flexible customization

ryanmerolle:
I say death to IE, but seriously most people using NetBox have other means of browsers these days.  Nearly all shops allow Chrome.

A drop for internet explorer may not even kill the entire usability but just greatly hamper the new functionality.  Which might not be the end of the world.

jeremystretch:
I also have to think that organizations stuck on Internet Explorer _probably_ aren't running the latest and greatest NetBox release right away either.

checktheroads:
To add some clarity, the list of browsers Bootstrap 5 supports is [here](https://getbootstrap.com/docs/5.0/getting-started/browsers-devices/). TL;DR: Everything except IE. It's worth noting that I (re)built [hyperglass](https://github.com/checktheroads/hyperglass) with React and several other dependencies, and inadvertently dropped IE support. I've yet to hear a word from any users that this has been a problem (and this is an app meant to be end-user facing!). Additionally, my organization's website has been IE-free for over a year with nary a complaint.

It's also worth noting that one of the other major CSS frameworks out there — [Tailwind](https://tailwindcss.com/) — also no longer supports IE. I tend to agree with both @jeremystretch & @ryanmerolle — it's likely that either Netbox users are already using a modern browser, or if not, they're probably not running modern Netbox.

[...]

PR #6013, opened 2021-03-18
Closes #5429: Paginator on top of tables

### LINKED ISSUE #5429: Paginator on top of tables
### Environment
* Python version: 3.6.8
* NetBox version: 2.9.9

### Proposed Functionality
Move or add paginator at top of any tables (e.g. prefixes, ip address, prefix ip addresses)

### Use Case

At least there is no search/filter in prefix ip addresses view so the only way to find IP address is to scroll down many pages to find next-page button.

Futhermore, I think it is better to navigate between table pages via page navigator that on the top of page. When it is in bottom of page you have to scroll down all pages to find next-button page until reach object you want

### Database Changes
N/A

### External Dependencies
N/A

cpmills1975:
### Fixes: #5429

* I've re-worked the structure of the top of the page to show the page buttons right aligned with the table.
* I've restructured the grids on the page such that the bottom paginator controls are always right aligned with the table. There are a few side effects to this.
  * The responsiveness of the application as a whole is, IMO, better - especially for small screens.
  * The select all n objects confirmation no longer pushes the table below the sidebar (an unreported bug?)
  * For some very wide (e.g. 4K) screen widths, the side bar now appears a bit large. This will be trivial to fix by using the col-xxl-* classes from Bootstrap 5 (when implemented in #5893) or could be back-ported to the NetBox CSS now. I stopped short of doing this as part of this change)

Screenshot 2021-03-18 at 22 12 21

Screenshot 2021-03-18 at 22 12 43

Screenshot 2021-03-18 at 22 15 10

Screenshot 2021-03-18 at 22 13 11

### DIFF
--- a/netbox/project-static/css/base.css
+++ b/netbox/project-static/css/base.css
@@ -33,11 +33,6 @@ footer p {
 #navbar_search {
     padding: 0 8px;
 }
-.right-side-panel {
-    max-width: 400px;
-    min-width: 300px;
-    padding-right: 0;
-}
 
 /* Hide the username in the navigation menu on displays less than 1400px wide */
 @media (max-width: 1399px) {
@@ -168,7 +163,7 @@ td.min-width {
 
 /* Paginator */
 div.paginator {
-    margin-bottom: 20px;
+    margin-bottom: 1em;
 }
 div.paginator form {
     margin-bottom: 6px;
--- a/netbox/templates/circuits/provider.html
+++ b/netbox/templates/circuits/provider.html
@@ -129,7 +129,7 @@
                 </div>
             {% endif %}
         </div>
-    {% include 'inc/paginator.html' with paginator=circuits_table.paginator page=circuits_table.page %}
+    {% include 'inc/paginator.html' with paginator=circuits_table.paginator page=circuits_table.page show_count=True %}
     {% plugin_right_page object %}
     </div>
 </div>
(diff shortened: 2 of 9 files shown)


jeremystretch:
Thanks for your work on this. I hate to ask, but do you mind tabling this for now? We have #5893 coming up for v2.12 and I'm trying to minimize the amount of flux in the templates in the immediate term. Let's see how that pans out and then examine what changes remain to be made to accommodate the top paginator.

cpmills1975:
@jeremystretch no worries. Are there plans for a big overhaul in the UI/UX for 2.12 or is NetBox likely to look similar for now, but underpinned with the BS5 framework? I'm just trying to work out whether I need to start prepping my users for a major overhaul or not.

FWIW, moving the main body of the page in to a properly defined column and removing the explicit width constraint from the side bar adds most value here as it makes the app much more responsive. I'm not sure how many sadists will be trying to use NetBox on a phone, but mobile-first is the best way forward right? Removing the width constraint on the side-bar does make it a bit too wide IMO on wide screens (I'm on a 32" 4K monitor at the moment) but this will be easier to fix with the new col-xxl-* classes that come with BS5.

jeremystretch:
>  Are there plans for a big overhaul in the UI/UX for 2.12 or is NetBox likely to look similar for now, but underpinned with the BS5 framework?

We expect a major overhaul, though that work is still in progress.

jeremystretch:
Going to close this out as we've ended up overhauling a ton under the `feature` branch (with more work to come). However, it's something we should revisit before the v3.0 release.

PR #6036, opened 2021-03-23
Add support for customfield static choices with multiple select

### LINKED ISSUE #6035: Add support for customfield static choices with multiple select
### NetBox version

v2.10.4

### Feature type

New functionality

### Proposed functionality

Add support for custom field static choices with multiple select.

This resembles #2522 which was rejected before but with the revamp of custom fields this use case is much easier to support. Effectively it only adds a new type that uses a MultipleChoiceField and the StaticSelect2Multiple widget.

Note that CSV import is not yet supported because of #3637 and the multiple select custom fields are explicitly excluded.

### Use case

It adds custom fields with multiple selection to allow for feature flags and similar properties while reducing the selection scope to the options given. In the past this was only supported with tags. But the downside of tags are that if your choices are different per model you need to validate them, add tags per model or train your users to only use choice X for model Y.

Now you can add a field with a specific set of choices and those are the only ones available to the user.

### Database changes

No database changes required

### External dependencies

No external dependencies required

Urth:
### Fixes: #6035

Implementation to go along with #6035 that adds a new type that uses a MultipleChoiceField and the
StaticSelect2Multiple widget.

### DIFF
--- a/netbox/extras/choices.py
+++ b/netbox/extras/choices.py
@@ -13,6 +13,7 @@ class CustomFieldTypeChoices(ChoiceSet):
     TYPE_DATE = 'date'
     TYPE_URL = 'url'
     TYPE_SELECT = 'select'
+    TYPE_SELECT_MULTIPLE = 'select-multiple'
 
     CHOICES = (
         (TYPE_TEXT, 'Text'),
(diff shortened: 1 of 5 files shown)


jeremystretch:

Issue #6052, opened 2021-03-25
Accessing Fully Rendered Config Context Data is not Clear When Using Export Templates


tcaiazza:
### Change Type

Addition

### Area

- [ ] Installation instructions
- [ ] Configuration parameters
- [ ] Functionality/features
- [ ] REST API
- [ ] Administration/development
- [ ] Other

### Proposed Changes

The combination of export templates and config context data are extremely powerful, but use the two in combination isn't as intuitive as accessing ofter attributes within an object.  In order to access the fully rendered config context for an object, you need to use the function `get_config_context()` in the jinja template, which does not appear to be documented anywhere.

tcaiazza:
I'd be happy to write the documentation myself, I just need the issue be accepted.

PR #6053, opened 2021-03-25
Update export-templates.md

tcaiazza:
### Fixes: https://github.com/netbox-community/netbox/issues/6052

Both export templates and config context data are extremely powerful and useful, but it isn't immediately clear how use the two together.  This change adds details on how to use the two together.


jeremystretch:

Issue #5830, opened 2021-02-17
Expand Export Template to allow for non-attachment response


rodvand:
### Environment
* Python version: 3.6.8
* NetBox version: 2.10.4

### Proposed Functionality
Add an option to the Export Template functionality where it does not download the export template as an attachment, but instead presents it right away.

### Use Case
Allow rendered content to be opened straight away. Currently I am generating a link using the Export Template and am required to open the downloaded attachment and click on the link to access the page. Being able to adjust this behaviour would make for a smoother workflow. 

### Database Changes
Boolean field on the ExportTemplate model for deciding whether to set the Content-Disposition to attachment. Suggests to set this default as True to keep the current workflow as standard.

### External Dependencies
None

rodvand:
I can take a stab at it if you assign it to me.

jeremystretch:
I'm assuming this will require the addition of a new boolean field on the ExportTemplate model? In that case, this needs to be wait for a milestone assignment. It's a minor change though, obviously; we can tag it for v2.11.

rodvand:
Will we manage to get this into 2.11? If needed I can try to work on this.

jeremystretch:
@rodvand go for it! I would like to get the v2.11 beta out this week, though, if possible. Let me know if you'd rather me knock it out.

We can either add a boolean field to control the content disposition header, or add it only when a file extension is set.

PR #6072, opened 2021-03-30
#5830: Expand ExportTemplate model with attachment choice

rodvand:
### Fixes: #5830
Expand ExportTemplate model with the Boolean as_attachment to allow the user to not require a local download of the export template. Instead, with as_attachment unchecked, the export template is presented directly in the user's browser.


jeremystretch:
Nice, thanks!

PR #6091, opened 2021-04-05
Update export-templates.md

### LINKED ISSUE #6052: Accessing Fully Rendered Config Context Data is not Clear When Using Export Templates

tcaiazza:
### Fixes: https://github.com/netbox-community/netbox/issues/6052

Both export templates and config context data are extremely powerful and useful, but it isn't immediately clear how use the two together. This change adds details on how to use the two together.

### DIFF
--- a/docs/additional-features/export-templates.md
+++ b/docs/additional-features/export-templates.md
@@ -18,6 +18,14 @@ Height: {{ rack.u_height }}U
 
 To access custom fields of an object within a template, use the `cf` attribute. For example, `{{ obj.cf.color }}` will return the value (if any) for a custom field named `color` on `obj`.
 
+Since local_context_data is part of an object, it is immediately available to use in the export template, however if you need the fully rendered context config you'll need to use the function `get_config_context`. For example:
+```
+{% for server in queryset %}
+{% set data = server.get_config_context() %}
+{{data.syslog}}
+{% endfor %}
+```
+
 A MIME type and file extension can optionally be defined for each export template. The default MIME type is `text/plain`.
 
 ## Example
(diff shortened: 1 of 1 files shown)


docs/additional-features/export-templates.md
```diff
@@ -18,6 +18,14 @@ Height: {{ rack.u_height }}U
 
 To access custom fields of an object within a template, use the `cf` attribute. For example, `{{ obj.cf.color }}` will return the value (if any) for a custom field named `color` on `obj`.
 
+Since local_context_data is part of an object, it is immediately available to use in the export template, however if you need the fully rendered context config you'll need to use the function `get_config_context`. For example:
```
jeremystretch, line 21: We should not encourage accessing `local_context_data` directly: `get_config_context()` should always be used to ensure the complete configuration data is returned.

docs/additional-features/export-templates.md
```diff
@@ -18,6 +18,14 @@ Height: {{ rack.u_height }}U
...
+```
+{% for server in queryset %}
+{% set data = server.get_config_context() %}
+{{data.syslog}}
```
jeremystretch, line 25:
```suggestion
{{data.syslog}}
```
```suggestion
{{ data.syslog }}
```

jeremystretch (changes requested):

docs/additional-features/export-templates.md
tcaiazza: I've updated the comment here to drop completely the reference to `local_context_data` and said to use the config_context data you should use the `get_config_context` function.

PR #6111, opened 2021-04-08
Fix 5952

### LINKED ISSUE #5952: Cable Export current format is not able to be used for import
### NetBox version

v2.10.5

### Feature type

### Proposed functionality

When we export the cable from the URL /dcim/cables/
We get a file with this headers:
"termination_a_type	termination_a_id	termination_b_type	termination_b_id	type	status	label	color	length	length_unit"
This is not able to be used directly for the import feature. 
I propose to have also an export option, following with the same headers used in the import option.
This are 

### Use case

To simplify the population of cabling having an export that provides a correct template to be used is extremely valuable. This enables all users to do a bulk creation of the cables.
It also allows us to export the information, and if we have the name of the devices we can used this to provide additional information to other field teams. 

This allow to have a more friendly and intuitive UI

### Database changes

No DB Changes required

### External dependencies

No additional Dependencies

hellerve:
### Fixes: #5952

This PR unifies the import and export CSV formats for cables. As of now it is a mix of both. The new headers are:

```
side_a_parent_type
side_a_parent_id
side_b_parent_type
side_b_parent_id
```

They replace the headers `side_a_device` and `side_b_device`. An ID instead of a name was chosen for two reasons:
- Circuits do not have names, just IDs.
- Device names are not necessarily unique across sites.

The value for `side_a_name` and `side_b_name` also underwent a semantic change: if the component type does not have a name (such as for circuit terminations), the value of `side_<side>_name` will be assumed to be a PK instead. This is also exported like this.

I tried testing this with a good amount of cables (I dumped about ~1500 cables and reimported them), but I cannot guarantee that this code was tested with all possible combinations of types and values.

I’d be happy to rework this a bit if the approach used is not quite as intended.

Cheers

### DIFF
--- a/netbox/dcim/constants.py
+++ b/netbox/dcim/constants.py
@@ -59,6 +59,17 @@ POWERFEED_MAX_UTILIZATION_DEFAULT = 80  # Percentage
 # Cabling and connections
 #
 
+# Cable endpoint parent types
+CABLE_TERMINATION_PARENT_MODELS = Q(
+    Q(app_label='circuits', model__in=(
+        'circuit',
+    )) |
+    Q(app_label='dcim', model__in=(
+        'device',
+        'powerpanel',
+    ))
+)
+
 # Cable endpoint types
 CABLE_TERMINATION_MODELS = Q(
     Q(app_label='circuits', model__in=(
--- a/netbox/dcim/forms.py
+++ b/netbox/dcim/forms.py
@@ -23,6 +23,7 @@ from tenancy.models import Tenant, TenantGroup
 from utilities.forms import (
     APISelect, APISelectMultiple, add_blank_choice, BootstrapMixin, BulkEditForm, BulkEditNullBooleanSelect,
     ColorSelect, CommentField, CSVChoiceField, CSVContentTypeField, CSVModelChoiceField, CSVModelForm,
+    DeferredCSVModelChoiceField,
     DynamicModelChoiceField, DynamicModelMultipleChoiceField, ExpandableNameField, form_from_model, JSONField,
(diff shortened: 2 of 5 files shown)


jeremystretch:
I'm not sure this approach is practical. When people want to import cables, they generally want to be able to refer to devices by their names. Requiring them to first ascertain NetBox's unique numeric ID for each device imposes an additional burden on the user. We should support referencing devices by a user-specified attribute, as we do elsewhere for CSV imports. For example, we should support

```
side_a_parent_type,side_a_parent.id
dcim.Device,123
```

or

```
side_a_parent_type,side_a_parent.name
dcim.Device,router1
```

or

```
side_a_parent_type,side_a_parent.cid
circuits.Circuit,39400271
```

hellerve:
Fair enough, I can rework this to support that after the weekend. Should we offer the same for the side itself, instead of using `side_<side>_name`, then?

jeremystretch:
Yes, we should try to keep it consistent wherever possible.

hellerve:
As I see this, this would require a custom CSV field type (`CSVModelChoiceField` doesn’t cut it, since it’s used in conjunction with a `ContentType`). This is definitely possible, but makes this PR somewhat bigger, I expect.

I’ll try and implement that, but let me know if you have any other ideas or if I’m missing something!

hellerve:
I came up with a solution with a new field type, `DeferredCSVModelChoiceField`, which defers getting the value until it is passed a queryset. While that’s a little unconventional, it’s the best way I could come up with making the field’s queryset be depnendent on another field’s selection.

hellerve:
I’d be happy to reopen this and adjust if necessary if we find time to review it. It’s ready for a review when time allows for it IIRC.

PR #6164, opened 2021-04-15
#5822: Increase Virtual Machine Model Name Length to 255

### LINKED ISSUE #5822: Increase virtual machine "name" field maximum length from 64 characters
### Environment
* Python version: 3.6.11
* NetBox version: 2.10.4

### Proposed Functionality
Tracking AWS EC2 instances in Netbox, but AWS allows `Name` tags up to 256 characters and Netbox only allows 64 characters:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions

Ideally the limit in Netbox would be increased to allow for tracking EC2 instances with their full `Name`.

### Use Case
EC2 instance with name longer than 64 characters cannot be easily tracked in Netbox, and causes conflicts with tracking other EC2 instances/Netbox vms if the 256 character name shortened to 64 characters is not unique.

### Database Changes
Increase character limit for Name field on Virtual Machine model to 256 characters

### External Dependencies
N/A

Joezeppe:


jeremystretch:

PR #6316, opened 2021-04-30
Closes #2434: Show 'Create & Assign IP Address' Button when Creating Interfaces

### LINKED ISSUE #2434: Add a "Create and assign IP" button to interface creation dialog
### Environment
* Python version: 2.7.13
* NetBox version: 2.4.4

### Proposed Functionality
Add another blue "Create and assign IP" button to the interface creation dialog that results in the same view as pressing the green "+" on a interface.

### Use Case
Often when adding an interface, users also want to assign an IP address to it. Also the distinction between interface itself and ip addresses configured on the interface (that makes perfectly sense) can be a difficulty to grasp. They have to find the mouseover on the green plus of the created interface to find out how to configure an address to the interface. Having a shortcut directly from the interface creation dialog would result in a much more streamlined experience.

### Database Changes
None.

### External Dependencies
None.

thatmattlove:
#### #2434 Changes:
- Show a 'Create & Assign IP Address' button, permissions permitting
- Handle `_assignip` form field (set by above button) by choosing first created object and redirecting the user to the Create IP Address View with the newly created object pre-populated.

#### UI Changes:
- Fix an issue in dark mode where object deletion confirmation buttons were the wrong color.
- Fix an issue in dark mode where select elements' search input fields' color was the wrong color.
- Set the `$link-hover-color` Sass variable in dark mode so `<a/>` tags are lightened on hover instead of darkened.
- Fix an issue where API-backed select elements used the default pagination limit of 50 objects, causing incomplete data and predefined selections not to work properly.
- Set smaller table padding-x and font-size so that table data fits a little better in the viewport.


checktheroads:
Closing this because I'm terrible at git — will reopen and implement properly shortly.

PR #6323, opened 2021-04-30
Closes #2434: Show 'Create & Assign IP Address' Button when Creating …

### LINKED ISSUE #2434: Add a "Create and assign IP" button to interface creation dialog

thatmattlove:
#### Changes
- Show a 'Create & Assign IP Address' button, permissions permitting
- Handle `_assignip` form field (set by above button) by choosing first created object and redirecting the user to the Create IP Address View with the newly created object pre-populated.


netbox/templates/dcim/device_component_add.html
```diff
@@ -20,6 +20,11 @@ <h4>{{ component_type|title }}</h4>
         <div class="col-md-8 text-end">
         {% block buttons %}
         <a class="btn btn-outline-danger" href="{{ return_url }}">Cancel</a>
+        {% if perms.ipam.add_ipaddress %}
```
DanSheps, line 23: Only concern, and I haven't pulled this down myself to test, does this show for all components or only interface?

checktheroads: You make a good point, I’ll review and adjust if necessary.

checktheroads: 04cc43b should fix this

netbox/dcim/views.py
```diff
@@ -1908,6 +1911,73 @@ class InterfaceCreateView(generic.ComponentCreateView):
     model_form = forms.InterfaceForm
     template_name = 'dcim/device_component_add.html'
 
+    def post(self, request):
```
jeremystretch, line 1914: This is quite a bit to replicate from ComponentCreateView. Can we instead refactor ComponentCreateView to introduce a `on_success()` method or something to that effect? That would allow us to override just that portion of the view logic.

checktheroads: I agree, and I didn't really like copying and pasting the method over — I was concerned about introducing an API change to the base class just to implement a slightly different logical change in a subclass. However, let me take a look and see if I can determine a functional way to break up the logic in a more reusable way.

checktheroads: 2d2719c should resolve this.

PR #6327, opened 2021-04-30
Add Basic Support for Smaller Screens in New UI

thatmattlove:
#### Changes

- Restructure navbar/sidebar so that collapsible version is visible on smaller screens, and invisible otherwise.
- Add `col` class to all existing `col-md-*` elements, so that `col` styles are present on smaller screens.
- Adjust navbar toggler styles for dark mode.


DanSheps:
LGTM

PR #6537, opened 2021-06-03
feat: markdown support in report's description

### LINKED ISSUE #6527: Allow to use markdown on report description
### NetBox version

v2.11.3

### Feature type

### Proposed functionality

I'd like to propose an enhancement on the look & feel of the r[eport_list.html](https://github.com/netbox-community/netbox/blob/develop/netbox/templates/extras/report_list.html) page.

Each report is currently presented in a List of 4 columns (_Name_ | _Status_ | _Description_ | _Last Run_). I would propose to render the _Description_ column using markdown support.

The change would be as simple as modifying the report_list.html template and add the class `rendered-markdown` and use the `markdown` django template function on the following line

https://github.com/netbox-community/netbox/blob/8d0a3c8e6932f4c523919180d3e75608cdf54a9b/netbox/templates/extras/report_list.html#L32

The markdown support has already been implemented for each report log messages so it won't be difficult to use the same feature for the Report description (check this [commit](https://github.com/netbox-community/netbox/commit/7c48e3632db052924c80213fc07faa8d3059fe7c))

I can handle this issue and propose a Pull Request.

### Use case

Adding markdown support to each Report description allows to structure the this description easily if we want to explain in more detail the whole test case covered.

### Database changes

None

### External dependencies

None

maximumG:
### Fixes: #6527

* markdown support in report list and report result pages
* Add notes in the documentation regarding markdown

### DIFF
--- a/docs/additional-features/reports.md
+++ b/docs/additional-features/reports.md
@@ -80,13 +80,10 @@ class DeviceConnectionsReport(Report):
                 self.log_success(device)
 ```
 
-As you can see, reports are completely customizable. Validation logic can be as simple or as complex as needed.
+As you can see, reports are completely customizable. Validation logic can be as simple or as complex as needed. Also note that the `description` attribute support markdown syntax. It will be rendered in the report list page.
 
 !!! warning
     Reports should never alter data: If you find yourself using the `create()`, `save()`, `update()`, or `delete()` methods on objects within reports, stop and re-evaluate what you're trying to accomplish. Note that there are no safeguards against the accidental alteration or destruction of data.
-
-!!! note 
-    The `description`attribute support markdown syntax. You can use markdown to print fency description in the report list page.
 
 The following methods are available to log results within a report:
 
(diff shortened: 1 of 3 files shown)


docs/additional-features/reports.md
```diff
@@ -85,6 +85,9 @@ As you can see, reports are completely customizable. Validation logic can be as
 !!! warning
     Reports should never alter data: If you find yourself using the `create()`, `save()`, `update()`, or `delete()` methods on objects within reports, stop and re-evaluate what you're trying to accomplish. Note that there are no safeguards against the accidental alteration or destruction of data.
 
+!!! note
```
jeremystretch, line 88: We don't need call-outs specifically for this; just a note within the text mentioning Markdown support is sufficient.

docs/additional-features/reports.md
```diff
@@ -95,6 +98,9 @@ The following methods are available to log results within a report:
 
 The recording of one or more failure messages will automatically flag a report as failed. It is advised to log a success for each object that is evaluated so that the results will reflect how many objects are being reported on. (The inclusion of a log message is optional for successes.) Messages recorded with `log()` will appear in a report's results but are not associated with a particular object or status.
 
+!!! note
```
jeremystretch, line 101: Same as above

jeremystretch (changes requested):
Thanks for this!

docs/additional-features/reports.md
maximumG: Removed in commit 10c9954

docs/additional-features/reports.md
maximumG: Removed in commit 10c9954

jeremystretch (approved):

PR #6545, opened 2021-06-04
 Add ipv4 /32 and ipv6 /128 prefix support #6455

### LINKED ISSUE #6455: Add ipv4 /32 and ipv6 /128 prefix support
### NetBox version

v2.10.3

### Feature type

### Proposed functionality

Add ipv4 /32 and ipv6 /128 prefix support

### Use case

While testing netbox I've noticed that it does not allow user to create single-host networks. In my case I would like to use such feature to track end-user IP-unnumbered allocations while also keeping those /32 networks linked to a vlan

### Database changes

_No response_

### External dependencies

_No response_

crafty-ua:

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -340,16 +340,6 @@ class Prefix(PrimaryModel):
                     'prefix': "Cannot create prefix with /0 mask."
                 })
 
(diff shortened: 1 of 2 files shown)


jeremystretch:
Great work, thanks!

Issue #6544, opened 2021-06-04
ValueError exception when migrating to v2.11.5 with one or more VRFs defined


jeremystretch:
### NetBox version

v2.11.5

### Python version

3.7

### Steps to Reproduce

1. Start with NetBox v2.11.4 or earlier
2. Create a VRF
3. Create one or more prefixes within that VRF
4. Upgrade to NetBox v2.11.5

### Expected Behavior

The upgrade should complete successfully.

### Observed Behavior

A ValueError exception is raised due to a bug in the way VRFs are retrieved in the migration:

```
ValueError: Cannot query "VRF object (1)": Must be "VRF" instance.
```

It seems that Django does not like using the "faked" VRF instance returned by `apps.get_model()` for filtering. It's being cast as a string by `filter()` rather than its numeric ID beign evaluated.

```python
VRF = apps.get_model('ipam', 'VRF')
...
    # Iterate through all VRFs, rebuilding each
    for vrf in VRF.objects.all():
        rebuild_prefixes(vrf)
```

Altering the migration to reference the ID directly resolves the issue:

```python
        rebuild_prefixes(vrf.pk)
```

```
$ ./manage.py migrate
Operations to perform:
  Apply all migrations: admin, auth, circuits, contenttypes, dcim, extras, ipam, secrets, sessions, taggit, tenancy, users, virtualization
Running migrations:
  Applying ipam.0048_prefix_populate_depth_children...
Updating 3157 prefixes...
 OK
```

maznu:
Thanks for fixing this one so quickly, @jeremystretch!

PR #6548, opened 2021-06-04
Release v2.11.6

jeremystretch:
### Bug Fixes

* [#6544](https://github.com/netbox-community/netbox/issues/6544) - Fix migration error when upgrading with VRF(s) defined


maznu:
We too got hit by #6544. Happy to confirm that this fixed the migration for us.

PR #6561, opened 2021-06-07
CSV file import

### LINKED ISSUE #6560: CSV file import
### NetBox version

v2.11.6

### Feature type

New functionality

### Proposed functionality

Allows users to upload a csv file of new objects into netbox in addition to the current textbox for bulk import. 

### Use case

Provides an easier way for users to upload large amounts of data into netbox at one time. 

### Database changes

_No response_

### External dependencies

_No response_

abigley:
### Fixes: #6560

An example / first draft of a solution for #6560 .

In the current implementation, if users try to use both the `CSVFileField` and the `CSVDataField`, the `CSVDataField` will be ignored and only the objects in the file will be imported. It should be unusual that both fields are populated, but let me know if you prefer to handle that case differently, such as by raising a validation error. The csv file is checked for errors (incorrect formatting, objects already exist, invalid columns) before importing.

This solution involves some duplication between `CSVFileField` and `CSVDataField`. If you prefer, I could remove most of the duplication by refactoring them to both inherit from a single class - say, `CSVParsingField`.

### DIFF
--- a/netbox/netbox/views/generic.py
+++ b/netbox/netbox/views/generic.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import csv
 from copy import deepcopy
 
 from django.contrib import messages
@@ -7,7 +8,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import FieldDoesNotExist, ObjectDoesNotExist, ValidationError
 from django.db import transaction, IntegrityError
 from django.db.models import ManyToManyField, ProtectedError
-from django.forms import Form, ModelMultipleChoiceField, MultipleHiddenInput, Textarea, FileField, CharField
+from django.forms import Form, ModelMultipleChoiceField, MultipleHiddenInput, Textarea, FileField
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, redirect, render
 from django.utils.html import escape
@@ -668,7 +669,6 @@ class BulkImportView(GetReturnURLMixin, ObjectPermissionRequiredMixin, View):
             Upload_CSV = FileField(
                 required=False
             )
-        # this is where the import form is created -- add csv upload here or create new form?
         return ImportForm(*args, **kwargs)
 
     def _save_obj(self, obj_form, request):
@@ -693,7 +693,6 @@ class BulkImportView(GetReturnURLMixin, ObjectPermissionRequiredMixin, View):
         logger = logging.getLogger('netbox.views.BulkImportView')
         new_objs = []
         form = self._import_form(request.POST)
-        # this is where the csv data is handled
 
         if form.is_valid():
             logger.debug("Form validation was successful")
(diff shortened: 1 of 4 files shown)


netbox/netbox/views/generic.py
```diff
@@ -665,7 +665,10 @@ class ImportForm(BootstrapMixin, Form):
                 from_form=self.model_form,
                 widget=Textarea(attrs=self.widget_attrs)
             )
-
+            upload_csv = CSVFileField(
```
jeremystretch, line 668: This field needs a human-friendly label (e.g. "CSV file")

netbox/netbox/views/generic.py
```diff
@@ -689,15 +692,18 @@ def get(self, request):
...
+        form = self._import_form(request.POST, request.FILES)
 
         if form.is_valid():
             logger.debug("Form validation was successful")
 
             try:
                 # Iterate through CSV data and bind each row to a new model form instance.
                 with transaction.atomic():
-                    headers, records = form.cleaned_data['csv']
+                    if request.FILES:
```
jeremystretch, line 703: We should add a validation method to the form to check for both `csv` and `upload_csv` set, and raise a ValidationError if so as a sanity check.

netbox/utilities/forms/fields.py
```diff
@@ -221,6 +222,77 @@ def validate(self, value):
         return value
 
 
+class CSVFileField(forms.FileField):
```
jeremystretch, line 225: Most of the content of this class seems to have been duplicated from CSVDataField: We always want to minimize redundant code. Instead of duplicating the logic, we should move it out of CSVDataField into a separate class or functions that can be utilized by both CSVDataField and CSVFileField.

netbox/netbox/views/generic.py
jeremystretch: Also, `csv_file` might be a better name than `upload_csv`.

jeremystretch (changes requested):
This is a great start, thank you for your contribution. Just needs a bit of cleaning up before we can merge it.

netbox/netbox/views/generic.py
abigley: Renamed.

netbox/netbox/views/generic.py
abigley: Added a method for this within the _import_form definition

netbox/utilities/forms/fields.py
abigley: I made two functions, parse_csv and validate_csv, within forms/utils.py. Both CSV fields use these functions to reduce redundancy.

jlost (approved):
Looks good to me!

jlost:
@jeremystretch This would be useful to me, and it looks like all requested changes have been addressed. Could you give this another peek when you have a chance?

jeremystretch:
Might need a bit of cleanup (e.g. stripping whitespace from uploaded files) but I can take care of that. Thanks!

jeremystretch (approved):

PR #6592, opened 2021-06-11
Closes #5264: REST API endpoint for tokens

### LINKED ISSUE #5264: Extend the Users REST API Endpoint to support managing API Tokens
### Environment
* Python version: 3.7
* NetBox version: 2.9.7

### Proposed Functionality
Establish REST API endpoint at `/api/users/tokens/` for managing tokens via the API.

### Use Case
Enables the ability to programmatically manage the user tokens that have access to the netbox API. Currently the management of this is only via the Admin UI or user control panel. 

There are a couple potential use cases that can be benefitted from.

**1) Creation of API users in configuration management (IaC) platform.** 
 - We use a configuration management tool to deploy our Netbox instance and manage it, including user and token management. Currently we are using the `nbshell` programmatically to create the API users that we need for our automated processes. Since the creation of the Users and Groups endpoint, we are almost able to do it all via the API. However since Token's is not supported, we will still have to shell out the the `nbshell` to add our users.

**2) Dynamic API user creation**
- We can see a use case where ephemeral users/tokens can be generated by a tool such as hashicorp vault which would allow us to have short lived secrets that can be revoked and/or deleted after a certain point in time or if the secret was breached. This would improve security and allow us to have better auditing and compliance with the usage of a tool like vault. If token management via the API were supported, we could programmatically do these operations in a safe a repeatable way.

### Database Changes
None

### External Dependencies
None

jeremystretch:
### Closes: #5264

- Add a REST API endpoint for tokens (`/api/users/tokens/`)
- Add a child endpoint `provision/` to facilitate the provisioning of new tokens for users without an existing token

### DIFF
--- a/docs/release-notes/version-3.0.md
+++ b/docs/release-notes/version-3.0.md
@@ -8,6 +8,23 @@
 
 ### New Features
 
+### REST API Token Provisioning ([#5264](https://github.com/netbox-community/netbox/issues/5264))
+
+This release introduces the `/api/users/tokens/` REST API endpoint, which includes a child endpoint that can be employed by a user to provision a new REST API token. This allows a user to gain REST API access without needing to first create a token via the web UI.
+
+```
+$ curl -X POST \
+-H "Content-Type: application/json" \
+-H "Accept: application/json; indent=4" \
+https://netbox/api/users/tokens/provision/
+{
+    "username": "hankhill",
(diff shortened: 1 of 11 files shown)


jeremystretch:
Still need to tweak TokenPermissions to work around non-model-based API views.

Issue #6217, opened 2021-04-19
Wrong value types on custom fields


exiz01:
### NetBox version

v2.11.0

### Python version

3.6

### Steps to Reproduce

1. Create a custom field for a dcim > device type.
2. Select a value type as an integer.
3. Create a new device type through the API and set a string value equal to '1234' for the new custom field in the same request.

### Expected Behavior

The new device type should have an integer value on the custom field.
If a device type is created through the GUI this works fine.

### Observed Behavior

The new device type has a string value on the custom field

PR #6641, opened 2021-06-21
Add type check for custom fields with "text" type

### LINKED ISSUE #6640: Wrong value types on custom fields
### NetBox version

v2.11.7

### Python version

3.6

### Steps to Reproduce

1. Create a custom field for a ipam > prefix.
2. Select a value type as a text.
3. Update prefix through the API and set a integer value equal to 51 for the new custom field in the same request

### Expected Behavior

Error should be returned.
If a prefix updated through the GUI this works fine.

### Observed Behavior

The prefix has an integer value on the custom field

avisom:
### Fixes: #6640

Add type check for custom fields with "text" type
Same as #6217

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -281,6 +281,8 @@ class CustomField(BigIDModel):
 
             # Validate text field
             if self.type == CustomFieldTypeChoices.TYPE_TEXT and self.validation_regex:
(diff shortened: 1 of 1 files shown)


avisom:
not actual after https://github.com/netbox-community/netbox/commit/efa0fc2b0996cea4c6b823f76aec52b9b339ef8e

PR #6645, opened 2021-06-22
Update installation

hanserasmus:
Just separated it so the user can easily click the copy button, and only be presented with the command.

### DIFF
--- a/docs/installation/3-netbox.md
+++ b/docs/installation/3-netbox.md
@@ -73,6 +73,9 @@ Next, clone the **master** branch of the NetBox GitHub repository into the curre
 
 ```no-highlight
 $ sudo git clone -b master https://github.com/netbox-community/netbox.git .
+```
+The screen below should be the result:
+```
 Cloning into '.'...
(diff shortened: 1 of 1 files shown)


docs/installation/3-netbox.md
```diff
@@ -73,6 +73,9 @@ Next, clone the **master** branch of the NetBox GitHub repository into the curre
 
 ```no-highlight
 $ sudo git clone -b master https://github.com/netbox-community/netbox.git .
+```
+The screen below should be the result:
```
jeremystretch, line 77: Please add newlines above and below this text to remain consistent with the rest of the document.

jeremystretch:
@hanserasmus are you still interested in pursuing this PR?

docs/installation/3-netbox.md
hanserasmus: Added.

hanserasmus:
@jeremystretch Sorry for the delay. I forgot about this one completely and it got lost in the mail. Apologies.

PR #6703, opened 2021-07-07
Update nginx.conf

### LINKED ISSUE #6702: Fix example nginx config to listen on IPv6 alongside IPv4
### NetBox version

v2.11.8

### Python version

3.8

### Steps to Reproduce

1. Install netbox using the docs, opting for nginx
2. try to connect using dns name that only has a AAAA record

### Expected Behavior

Web server responds, netbox is displayed

### Observed Behavior

A connection refused error appears.

aaroneg:
extend example nginx config to listen on IPv6 addresses as well as IPv4.

### Fixes: #6702

update example nginx config to also listen on all ipv6 addresses since the current option results in only listening to ipv4 on nginx 1.18.0 (Ubuntu)

### DIFF
--- a/contrib/nginx.conf
+++ b/contrib/nginx.conf
@@ -1,6 +1,7 @@
 server {
     listen 443 ssl;
-
+    listen [::]:443 ssl;
+    
     # CHANGE THIS TO YOUR SERVER'S NAME
     server_name netbox.example.com;
 
@@ -24,6 +25,7 @@ server {
 server {
     # Redirect HTTP traffic to HTTPS
     listen 80;
+    listen [::]:80;
     server_name _;
     return 301 https://$host$request_uri;
 }


contrib/nginx.conf
```diff
@@ -1,6 +1,7 @@
 server {
     listen 443 ssl;
-
+    listen [::]:443 ssl;
```
jeremystretch, line 3: Only one line is needed: `listen [::]:443 ssl ipv6only=off;`

contrib/nginx.conf
```diff
@@ -24,6 +25,7 @@ server {
 server {
     # Redirect HTTP traffic to HTTPS
     listen 80;
+    listen [::]:80;
```
jeremystretch, line 28: Same as above (`listen [::]:80 ipv6only=off;`)

jeremystretch (changes requested):

jeremystretch:
Pushed the suggested changes above under 72aaf76cf475245ef399ad076606b0b4be776ee3 as there has been no further activity on this PR.

Issue #6639, opened 2021-06-21
Remove queryset caching support


jeremystretch:
### Proposed Changes

Remove the support for query caching that was introduced in v2.6 under #2647. This would remove django-cacheops as a dependency and eliminate the ~`REDS['caching']` and~ `CACHE_TIMEOUT` configuration parameters.

### Justification

Since its implementation, we have been constantly chasing [numerous bugs](https://github.com/netbox-community/netbox/issues?q=is%3Aissue+caching+label%3A%22type%3A+bug%22+) attributable to limitations in the caching engine. I have addressed some of these issues within django-cacheops itself, however we continue to run into problems, particularly where complex workflows are involved.

In my opinion, it does not make sense to continue burning time chasing these bugs to support a feature with relatively little benefit. I would prefer to optimize NetBox itself where necessary to meet performance expectations rather than trying to keep the caching layer intact. I've opened this issue for discussion to see what the community thinks.

jhujhiti:
Are there any quantitative numbers on how much the removal of caching would hurt specific API endpoints, or a way to disable the cache in an existing 2.11 environment to evaluate what the impact of this change would be? As a user, I'm certainly in favor of reducing the complexity of the stack, but my deployment has a number of identical queries run by external monitoring repeatedly and losing the cache there could put a lot more easily-cached load on the database. (Although I do suspect Redis isn't making much difference, I don't have a way to prove it.)

jeremystretch:
@jhujhiti You can test on your own deployment by toggling the caching feature on and off. Setting `CACHE_TIMEOUT = 0` in `configuration.py` and restarting the NetBox service will disable caching.

jhujhiti:
Great! No discernible impact on the 95th or 50th percentile latency on the API endpoints I've been monitoring. Consider this a vote in favor.

dreng:
Very low impact in our environment. Not noticeable in every day use. I'd say: kick it!

maximumG:
As mentioned in issue #6651. The [check_release](https://github.com/netbox-community/netbox/blob/develop/netbox/utilities/background_tasks.py) process will be impacted by cache removal.

However _check_release_ can easily rely on RQ [result caching](https://python-rq.org/docs/results/) (`result_ttl`) and RQ [registries](https://python-rq.org/docs/job_registries/) to achieve the same result as the current implementation.

[...]

PR #6731, opened 2021-07-09
Fixes #6651:  Add plugin's queueing system

### LINKED ISSUE #6651: Allow plugin to create their own django-rq queues
### NetBox version

v2.11.7

### Feature type

New functionality

### Proposed functionality

I'd like to propose the ability for netbox plugin to define some dedicated _django-rq_ queue(s). This new feature has already been discussed in #4908 but was missing a owner back in time.

The idea would be to store queue names in a new attribute inside the `PluginConfig` class. This attributes would be a `list` in order to defined several queue to create by django-rq.

In order to avoid queue name collision, the actual queue names will be prepended by the plugin name.

As an exemple the following config will lead to 3 queues being created using django-rq:

* _my_super_plugin.high_
* _my_super_plugin.medium_
* _my_super_plugin.low_

```python
MyPluginConfig(PluginConfig):
    name = "my_super_plugin"
    queues = [
        "low"
        "medium"
        "high"
    ]
```

The creation of the queues will need to override django settings ([RQ_QUEUES](https://github.com/netbox-community/netbox/blob/607039f043f0a43437db05aae18bae6f1ef49f85/netbox/netbox/settings.py#L571)) and can be done in the `ready` method of the `PluginConfig` class.

We can also override django RQ_QUEUES settings directly in the `settings.py`, during [init of all plugins](https://github.com/netbox-community/netbox/blob/4292d88a9230789ef0c5af627e7245d2d3dd4bda/netbox/netbox/settings.py#L597).

If you are interested in such feature I can take care of a PR.

### Use case

All netbox plugins have the possibility to use the _default_ queue to do some background/housekeeping tasks. If these tasks are long-running tasks they will block all other netbox-native background jobs, including netbox-reports & netbox-script.

### Database changes

_No response_

### External dependencies

_No response_

maximumG:
### Fixes: #6651

* Create Netbox core RQ queues : *high*, *default* and *low*. For now on, Netbox still uses the *default* queue for all background tasks (mainly webhooks).
* Add plugin's dedicated queues: plugins can now define their own dedicated queues
* Add some documentation in netbox-doc regarding core queues and plugin queues

### Warning

Also note that netbox default RQ worker must explicitly specify the 3 core queues to listen to. Without this, the default RQ worker will listen to every queues, including plugin's dedicated queues.

I reflect this change in `contrib/netbox-rq.service` systemD unit file, but this should also be modified in the [netbox-docker](https://github.com/netbox-community/netbox-docker/blob/676ca798cd684ce30fca2692180e680d2becf76e/docker-compose.yml#L27) project


jeremystretch:
Thanks for this. I went ahead and resolved the conflicts resulting from my recent work on #6639.

jeremystretch:
> Also note that netbox default RQ worker must explicitly specify the 3 core queues to listen to. Without this, the default RQ worker will listen to every queues, including plugin's dedicated queues.

Is it better for rqworker to default to _all_ defined queues, or to default to only the three "built-in" queues? I'm thinking about the likely scenario where a user installs a plugin with it own queue, but does not take any action to modify the rqworker invocation. The current approach (defaulting to all queues) probably work without issue initially, but could eventually cause performance issues. But if we only service the three built-in queues, it will be immediately evident to the user that rqworker needs to be adjusted, since the plugin's queue won't be serviced automatically.

maximumG:
> Is it better for rqworker to default to all defined queues, or to default to only the three "built-in" queues?

From my point of view it is better having the default RQ worker to explicitly listen to netbox core queues. I see the following benefits from it:

* Netbox core background tasks (report, scripts, webhook) won't be impacted by any plugin which requires dedicated queues
* If a user install a plugin which requires queues without a dedicated RQ worker, the plugin will break explicitly
* User could introduce performance issue when long running tasks from plugin overload netbox core tasks, and from experience it is hard to debug 😄 
* Plugin's developer will create a quick install guide and Plugin's user will follow it (carefully)

jeremystretch:
I guess what I'm getting at is what we should do with [the management command](https://github.com/netbox-community/netbox/blob/develop/netbox/extras/management/commands/rqworker.py) itself. Currently, if `manage.py rqworker` is invoked with no arguments, it defaults to servicing _all_ queues:

```python
    # If no queues have been specified on the command line, listen on all configured queues.
    if len(args) < 1:
        args = settings.RQ_QUEUES
```

So we have three viable options when `rqworker` is called without any queues specified:

1. Service all queues, both built-in and plugins (current behavior)
2. Service only the built-in queues
3. Return an error indicating that queues must be specified

I don't have a strong opinion, but feel like #3 might be the best approach to avoid confusion.

maximumG:
I do agree with you, as option 3 will avoid any confusion 👍

jeremystretch:
Alright, I'll take care of the rqworker bit separately. This looks good to me!

Issue #6785, opened 2021-07-22
Add the choice for AC Direct Wire for power types


bellwood:
### NetBox version

v2.11.9

### Feature type

Change to existing functionality

### Proposed functionality

Add the ability to choose `AC Direct Wire` to power types, similar to `DC Terminal`.

This would be a simple adder to `netbox/dcim/choices.py`

### Use case

Not all devices are connectorized coming off power panels and indicating an end device is directly wired is currently not possible. 

A simply example would be a UPS output breaker wired to a power panel input breaker and then further wired to a PDU input breaker (which would all be direct wired) or a power panel wired to a control panel, annunciator panel, etc.

### Database changes

None

### External dependencies

None

jeremystretch:
I could swear this came up in conversation under a previous issue, but I can't find it now.

At any rate, I think it makes sense.

PR #6788, opened 2021-07-22
Add AC Hardwire option to PowerPortTypeChoices

bellwood:
Resolves FR #6785


netbox/dcim/choices.py
```diff
@@ -441,6 +443,9 @@ class PowerPortTypeChoices(ChoiceSet):
             (TYPE_USB_3_B, 'USB 3.0 Type B'),
             (TYPE_USB_3_MICROB, 'USB 3.0 Micro B'),
         )),
+        ('AC', (
```
jeremystretch, line 446: Not sure it makes sense to group this under "AC," since most of the preceding options are also AC power. Maybe just an "other" category at the end of the list?

PR #6789, opened 2021-07-22
Add AC Hardwire option to PowerPortTypeChoices

### LINKED ISSUE #6785: Add the choice for AC Direct Wire for power types

bellwood:
### Closes: #6785

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -341,6 +341,8 @@ class PowerPortTypeChoices(ChoiceSet):
     TYPE_DC = 'dc-terminal'
     # Proprietary
     TYPE_SAF_D_GRID = 'saf-d-grid'
+    # Other
+    TYPE_OTHER = 'other'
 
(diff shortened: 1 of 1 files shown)


jeremystretch:
Oh, I meant we could just have an "other" group, and still call this "hard-wired."

(Also just FYI, there's no need to re-create pull requests: You can just push your changes to the same branch and the PR will be updated.)

PR #6799, opened 2021-07-23
style(sidebar): fix sidebar dual scroll contexts

ndom91:

Currently large sections, like IPAM, create new scroll contexts inside themselves.

image

I think its more standard to fully open the dropdown and just scroll it with the master list of categories, i.e. extending the original scroll context.

Sorry, wasn't sure if you were going against `feature` or `develop` at the moment..

### DIFF
--- a/netbox/project-static/styles/netbox.scss
+++ b/netbox/project-static/styles/netbox.scss
@@ -535,8 +535,7 @@ div.content-container {
   }
 
   .accordion-body {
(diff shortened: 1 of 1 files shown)


jeremystretch:
Let's hold off on this for a just a bit. I know @thatmattlove has been working on an alternative navigation menu that might do away with the accordion effect entirely. Appreciate the contribution though!

Issue #6797, opened 2021-07-23
v3.0-beta1 UI tweaks


jeremystretch:
### Proposed Changes

This issue serves as an umbrella for small changes to the desktop UI which don't warrant their own bug reports. Please comment below with any issues you encounter within the UI and we'll add it here.

For the sake of limiting scope, let's focus on issues observed on non-mobile platforms in this thread. Please note any issues with the mobile UI under #6828. Thanks!

- [x] Fix "flashing" from light mode when loading pages with dark mode enabled (@bellwood)
- [x] Duplicate search fields w/same ID throws error in browser dev console (@bellwood)
- [x] Login banner text colors not consistent with light/dark theme (@jeremystretch)
- [x] Sidebar footer disappears after login in Chrome (@michaelklose)
- [x] Sidebar "button" links render incorrectly under Safari (and other browsers?) (@bellwood)
- [x] "Refine search" buttons in global search results not properly spaced (@hoalex)
- [x] Double scrolling inside expanded nav menu (@ndom91)
- [x] Link color inside toasts can be difficult to read (@n0emis)
- [x] Utilization bars with low percentages are hard to read (@nahun)
- [x] Add cable length & type to SVG traces (@Jackbennett)
- [x] Device type images are not scaled to fit available space (@JeffTadashi)
- [x] Remove fade effect when hovering over dropdown items (@bluikko)
- [x] Visually distinguish custom links from built-in buttons (@rodvand)
- [x] Expanding a new menu section should close others (@jeremystretch)

### Justification

Clean up the user interface and achieve a greater overall user experience.

bellwood:
Adding media attributes to the dark/light CSS references should prevent "flashing" between modes:

`<link rel="stylesheet" href="/static/netbox-light.css" onerror="window.location='/media-failure/?filename=netbox-light.css'" media="(prefers-color-scheme: light)”
>
<link rel="stylesheet" href="/static/netbox-dark.css" onerror="window.location='/media-failure/?filename=netbox-dark.css'" media="(prefers-color-scheme: dark)"
>`

bellwood:
Search functionality in the sidebar and content pane use the same DOM `id="id_q"` and `id="search-button"` which renders errors in dev console

```<input name="q" id="id_q" type="text" aria-label="Search" placeholder="Search" class="form-control" value="" aria-describedby="search-button">```


```<button class="btn btn-primary" type="submit" id="search-button"><i class="mdi mdi-magnify"></i></button>```

[...]

PR #6860, opened 2021-08-02
Re-Implement Filter Panel & Support Custom API Select Fetch Triggers

thatmattlove:
## Re-Implement Filter Panel

From a comment in #6797, it was mentioned that the advanced search collapsible panel is a bit cumbersome, and I agree. Originally, it was implemented because the static sidebar reduced the screen real-estate such that the previous filter panel combined with the object list table made things feel cramped. However, with the new collapsible sidebar, this is no longer a concern. This re-implements the previous filter panel and adds some responsiveness.

## Support Custom API Select Fetch Event Types

In order to do the above, I implemented the ability to selectively (pun intended?) specify trigger events for which API selects will load their data — `load` (data is immediately loaded; current and default behavior), `open` (data is loaded every time the element is opened), and `collapse` (data is only loaded when a parent collapsing element is opened). This can be (optionally) set at the field level with the new `fetch_trigger` kwarg on `DynamicModelChoiceMixin`. Since the only current use-case is for the object list filter panels, I added `fetch_trigger='open'` to all filter form API select fields only.


thatmattlove:
@jeremystretch - I can handle the merge on this, I more or less wanted to get a thumbs up or thumbs down on this as a whole, since it ended up touching more than just front end stuff.

jeremystretch:
Looks great to me, nice work @thatmattlove!

My only comment is that it _might_ be overkill. v2.11 and earlier used Select2 for dynamic dropdowns, which IIRC only ever populate upon being clicked; synonymous with the `open` trigger here. I don't think that was ever problematic. It's nice to have some of these dropdowns populate automatically (e.g. when adding a new object), though I'm not sure is the additional load placed on the backend is significant. Either way, I'm happy to see this merged as-is; we can always tweak it moving forward if need be, and it's nice to preserve the option.

jeremystretch (approved):

PR #6880, opened 2021-08-04
Display device names in front of device front/rear images

### LINKED ISSUE #6879: Front/rear image obscures device name
### NetBox version

v2.11.10

### Feature type

### Proposed functionality

I have just started using the front/rear images functionality.  I found that the SVG is built up in the following sequence:

* background rectangle
* device name
* device image

This means that once you add images, you no longer can see the device names.

I propose simply reordering the rendering sequence so that the text is added after the image.  Before:

image

After:

image

Doing the same for rear suffers from the default text colour being black.  Before:

image

After:

image

I would therefore change the colour to white when there's an image, on the assumption that most images are "dark-like".  It then becomes:

image

Or else have setting(s) in the device type to select the text colour for the front and rear text when images are present.

### Use case

Being able to see at-a-glance the positions and the *names* of all the devices in a rack is an important use for rack elevations.

I do realise that there's a button to turn images on and off.  However, images without device names is pretty but not very useful.  This means that generally I would have to turn the images off all the time, which in turn means there is no point uploading images in the first place.

Potentially the button could be changed to a 3-state (show names and images; show names only; show images only).  But for me, the "images without names" variation is not something I'd use.

### Database changes

None

### External dependencies

None

candlerb:
### Fixes: #6879

Display device names in white text in front of device front/rear background images

### DIFF
--- a/netbox/dcim/elevations.py
+++ b/netbox/dcim/elevations.py
@@ -89,10 +89,9 @@ class RackElevationSVG:
         )
         link.set_desc(self._get_device_description(device))
         link.add(drawing.rect(start, end, style='fill: #{}'.format(color), class_='slot'))
-        hex_color = '#{}'.format(foreground_color(color))
-        link.add(drawing.text(str(name), insert=text, fill=hex_color))
 
         # Embed front device type image if one exists
+        text_color = '#{}'.format(foreground_color(color))
         if self.include_images and device.device_type.front_image:
             image = drawing.image(
                 href=device.device_type.front_image.url,
@@ -103,11 +102,14 @@ class RackElevationSVG:
             image.fit(scale='slice')
             link.add(image)
 
+        link.add(drawing.text(str(name), insert=text, stroke='#{}'.format(color),
+                 stroke_width='0.2em', stroke_linejoin='round', class_='device-label'))
+        link.add(drawing.text(str(name), insert=text, fill=text_color, class_='device-label'))
+
     def _draw_device_rear(self, drawing, device, start, end, text):
         rect = drawing.rect(start, end, class_="slot blocked")
         rect.set_desc(self._get_device_description(device))
         drawing.add(rect)
-        drawing.add(drawing.text(str(device), insert=text))
 
         # Embed rear device type image if one exists
         if self.include_images and device.device_type.rear_image:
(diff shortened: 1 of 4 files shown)


candlerb:
This is a proof-of-concept PR.

It doesn't quite work correctly, because if you have images and then turn them off (which is presumably done via CSS), the text colour may be wrong for the new background:

image

So it may be necessary to force the text colour to white via CSS only if the image is being shown, rather than forcing the text colour to white in Python as I did here.

Serazio:
Maybe it would be a better solution to display the names beside of the rack?  In our particular case we save devices names as fqdn, which sometimes lead into cut off hostnames in rack view.

candlerb:
> Maybe it would be a better solution to display the names beside of the rack? In our particular case we save devices names as fqdn, which sometimes lead into cut off hostnames in rack view.

I think that should be a different feature request, since the fqdn's will overrun the rack width whether or not you use front/rear images.

jeremystretch:
You;d still have the same issue on lighter-colored images though. How about adding an outline effect to the text using CSS?

candlerb:
The outline is ugly with small text.  The top one has `stroke="black" stroke-width="1px" stroke-linejoin="round"` (note: captured on retina macbook, without scaling, so shows double-size here)

image

And this is adding `font-weight="bold"`

image

(This was editing the SVG directly in-situ).

It might be possible to put an opaque background behind the text, but apparently that would require drawing another rectangle:
https://stackoverflow.com/questions/15500894/background-color-of-text-in-svg

That thread also suggests a solution using a second text element to make a "splodge" behind:

```
<text x="142.0" y="409.0" stroke="#000" stroke-width="0.6em">ldex-l3sw1</text>
<text fill="#ffffff" x="142.0" y="409.0">ldex-l3sw1</text>
```
renders as
image

Change the first text stroke to `#ccc` and stroke-width to `0.3em`:
image

Change second text fill to `#000000`:
image

Change the first stroke to `#eee`:
image

Adding `stroke-linejoin="round"` fixes the spike:
image

Reduce stroke-width to `0.2em`:
image

That's not too bad, although it still looks a bit "art shop".  You'd definitely only want to use it when the images are visible.

candlerb:
I have pushed a new version.  With images on, it looks like this:

image

With images off:

image

There's a bit of a blue halo around the text with images on (because the role colour is blue), but I can live with this.

candlerb:
Example of what the new version looks like with a very dark image (Netgate) and a very light image (Mikrotik)

image

candlerb:
This PR now implements separate toggles for Images and Labels:

image

You still cannot get *exactly* the same behaviour as before, which was mixed:
* devices with images did not show the label
* devices without images did show the label

However I prefer it this way, as the images-without-labels view is now uncluttered:

image

Are you OK with this?

Otherwise, it would be possible to get the old behaviour, by not adding the CSS class to labels if there's no image. But then you wouldn't be able to get the uncluttered view. At that point it probably makes sense to change to a 4-way dropdown: images+labels, mixed, images only, labels only.

candlerb:
Closing this out as I now have a version for Netbox 3.  No point implementing this twice.

candlerb:
New PR is #6960

PR #6884, opened 2021-08-04
Expose powerfeed details for racks in API

### LINKED ISSUE #6770: Expose Power Panel/Feed Details From UI in /api/dcim/racks
### NetBox version

v2.10.10

### Feature type

### Proposed functionality

Currently, via the API, the only information about power exposed in `/api/dcim/racks/` is `powerfeed_count` via `powerfeed_count=count_related(PowerFeed, 'rack')` in `netbox/dcim/api/views.py`

It would be beneficial to have the same related power panels/feeds data exposed via the API that is also visible in the UI, specifically:

-power panel: id, name
-power feed: id, name, status, type

### Use case

Having the above exposed for the rack endpoint gives an API caller a quick glance at power related details and provides the relationship details needed to make additional REST queries should further details about those panels/feeds be desired.

### Database changes

None

### External dependencies

None

bellwood:
Resolves #6770

### DIFF
--- a/netbox/dcim/api/serializers.py
+++ b/netbox/dcim/api/serializers.py
@@ -173,6 +173,7 @@ class RackSerializer(PrimaryModelSerializer):
     outer_unit = ChoiceField(choices=RackDimensionUnitChoices, allow_blank=True, required=False)
     device_count = serializers.IntegerField(read_only=True)
     powerfeed_count = serializers.IntegerField(read_only=True)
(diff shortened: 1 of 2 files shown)


bellwood:
Le sigh, these squashed. LMK if you want them broken apart.

jeremystretch:
@bellwood These absolutely need to be separate PRs. See my note on #6770 as to why we won't accept a PR for it.

PR #6891, opened 2021-08-05
feat: add cable reconnection

### LINKED ISSUE #4881: Please implement moving ends of a cable
### Environment
* Python version:  unknown - using a pre-installed system
* NetBox version:  2.8.7

### Proposed Functionality
Haven't found a way to move one end of a cable from one port/interface of a device to another one. The cable needs to be deleted and another one needs to be added, as it seems.

### Use Case
One could have made a mistake while adding a cable or the change is to reflect something that happens in pratice - it would be a real time-saver if one end or both of a cable could be reassigned to some other port/interface without havin to deleted and re-add the cable, which is a some work at least if you care about type, color, label and all those details.

### Database Changes
This appears to be a UI enhancement, so none expected.

### External Dependencies
None.

hellerve:
### Fixes: #4881

**Note:** This is marked as a draft PR, since it is rather big and makes some tradeoffs that we might have to talk about and reconsider. It also has no automated tests, since there are no cable form test cases. Should I try to add them anyway? My guess is “yes”, but I’d like to verify before doing that work :)

This PR adds, on the face of it, a button labelled `Reconnect` to a cable detail. The form that it leads to allows for both sides of the cable to move. As such, we needed to implement all the missing forms—18 in total. Another option would be to be able to edit just one side at a time, and do that from the device component table. This would possibly allow us to reuse the forms we already have. Since this is the implementation I had at hand (in a plugin, I can’t really edit the table), I stuck with this solution for now. Happy to change it.

It also, under the hood, deletes the cable and recreates it. This means that the PK changes on reconnection, which should be a non-issue in most cases. Otherwise we’d have to touch the cable model, which I refrained from doing for now. Again, happy to change that.

The use case of moving ends of a cable seems to work, though, so that’s good :)

Cheers

### DIFF
--- a/netbox/dcim/forms.py
+++ b/netbox/dcim/forms.py
@@ -4269,6 +4269,574 @@ class CableCSVForm(CustomFieldModelCSVForm):
         return length_unit if length_unit is not None else ''
 
 
+class ReconnectCableToDeviceForm(BootstrapMixin, forms.ModelForm):
+    """
+    Base form for connecting a Cable to a Device component
+    """
+    termination_a_site = DynamicModelChoiceField(
+        queryset=Site.objects.all(),
+        label='Site',
+        required=False,
+    )
+    termination_a_rack = DynamicModelChoiceField(
+        queryset=Rack.objects.all(),
+        label='Rack',
+        required=False,
+        display_field='display_name',
+        null_option='None',
+        query_params={
+            'site_id': '$termination_a_site'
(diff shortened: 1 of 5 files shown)


jeremystretch:
@hellerve no worries. TBH it might be more efficient for us to clean up some of the existing cabling forms/views before moving forward on this anyway. Thanks again for your effort.

PR #6898, opened 2021-08-05
Fixes #6887 Add Examples in the Lookup Expression Docs

### LINKED ISSUE #6887: Add Examples to the Lookup Expressions Section.
### Change Type

Addition

### Area

REST API

### Proposed Changes

When I first used lookup expressions I kind of puzzled on how exactly they worked. It still happens each time I try to use them. I'd like to add some examples to them to make it clearer how they are used. 

Add: 

`GET/api/ipam/vlans/?vid__gt=900`

after the Numeric Fields subsection

and add:

`GET api/dcim/devices/?name__ic=switch`

after the String Fields subsection

and add:

`GET api/ipam/vlans/?group_id__n=3203`

after the Foreign Keys & Other Fields

I'd like to have them formatted like the example under the subsection "Filtering by Custom Field" looks.

I'm happy to submit a PR on this if approved.

jmcguir:
### Fixes: #6887 

This adds some examples in the lookup expressions sections of the docs. I added them in each subsection since that seemed fit in more with the rest of the docs. 

I can move this around if you would like. I'm not exactly sure how you want the docs to look. I could either make an new subsection with examples or add it to the bottom of the main section but then it's kind of wonky because the examples are before the reference tables. Let me know.

### DIFF
--- a/docs/rest-api/filtering.md
+++ b/docs/rest-api/filtering.md
@@ -69,6 +69,12 @@ Numeric based fields (ASN, VLAN ID, etc) support these lookup expressions:
 | `gt` | Greater than |
 | `gte` | Greater than or equal to |
 
+Here is an example of a Numeric Field lookup expression that will return all VLANs with a VLAN ID greater than 9000:
+
+```no-highlight
+GET /api/ipam/vlans/?vid__gt=900
+```
+
 ### String Fields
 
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks!

PR #6904, opened 2021-08-06
fixes #6895: Some fields not exporting to CSV right

### LINKED ISSUE #6895: Prefix Roles not exporting right (CSV Current View)
### NetBox version

v2.11.9

### Python version

3.8

### Steps to Reproduce

1. Create a Prefix
2. "Export from current view"

### Expected Behavior

An actual empty value (eg ,,) on prefixes that do not have an assigned Role

### Observed Behavior

The exported CSV contains "&mdash;" on empty Role fields
image

martijnremmen:
### Fixes: #6895

* Adds a custom column for VRF Targets and Prefixes. that export to csv space seperated
* Replaced vlan/prefix templatecolumn with a normal column. **Note:** This removes old functionality that by clicking the role, one would filter the list on vlans/prefixes that have this specific role. If this functionality is desired, another custom column could be made. Although in my opinion the UX isn't clear on this (when clicking other properties we navigate to the view of that model, except for the role, which removes all current filters and applies a new one)

### DIFF
--- a/netbox/ipam/tables.py
+++ b/netbox/ipam/tables.py
@@ -21,14 +21,6 @@ PREFIX_LINK = """
 <a href="{% if record.pk %}{% url 'ipam:prefix' pk=record.pk %}{% else %}{% url 'ipam:prefix_add' %}?prefix={{ record }}{% if object.vrf %}&vrf={{ object.vrf.pk }}{% endif %}{% if object.site %}&site={{ object.site.pk }}{% endif %}{% if object.tenant %}&tenant_group={{ object.tenant.group.pk }}&tenant={{ object.tenant.pk }}{% endif %}{% endif %}">{{ record.prefix }}</a>
 """
 
-PREFIX_ROLE_LINK = """
-{% if record.role %}
-    <a href="{% url 'ipam:prefix_list' %}?role={{ record.role.slug }}">{{ record.role }}</a>
-{% else %}
-    &mdash;
-{% endif %}
-"""
-
 IPADDRESS_LINK = """
 {% if record.pk %}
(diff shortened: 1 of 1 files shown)


jeremystretch:
As discussed in the #netbox Slack channel yesterday, this does _not_ require custom columns. We just need to adjust the handling of empty values so that a dash is displayed in the table, but the CSV export returns an empty string. If you would like to work on this, I can assign #6895 to you. Otherwise, this PR will be closed.

Accendit:
After we discussed it I looked at the 'prefixes' field from vlans and the import & export target fields from vrfs (these can all display multiple values) and thought these might also not be handled correctly because the default behavior renders the template and just throws the html away.

This results in some junk characters (whitespace) left behind in the csv export (see image). I believed this pr to be a better solution to this. So we can manage how it is displayed (html template) and how it is exported (value method)
image

As for assigning me to this issue, I'll be having a week off starting Monday. I really want to contribute to NetBox but if you'd like this resolved within one week then someone else might be a better option.

Accendit:
Please feel free to close this if this isn't the right solution or if there is a better one. I'm not sure what the appropriate export behavior should be for fields that can contain multiple values but I found this solution to make sense.

I'd be happy to make changes to this if desired. But I'm not sure how to implement the solution you suggested (and it doesn't seem to take into account the multiple values per field scenario)

jeremystretch:
We shouldn't need to create custom columns just to ultimately override the placeholder text, since that won't scale. Let's see if we can come up with a more elegant approach.

jeremystretch:
I was able to address this in 1f1a05dc6 by subclassing the stock TemplateColumn to render the placeholder for empty values where appropriate, as well as switching to `Template(linkify=True)` for a few columns.

PR #6933, opened 2021-08-10
Fixes #6912: Fix static asset references when BASE_PATH is in use

### LINKED ISSUE #6912: some static files not found when `BASE_PATH` is changed
### NetBox version

v3.0-beta1

### Python version

3.8

### Steps to Reproduce

1. Set `BASE_PATH = 'netbox/'`
2. Run `upgrade.sh`.
3. Load site in browser.

### Expected Behavior

All static asset files should be found.

### Observed Behavior

Some static files cannot be found (i.e., 404 errors).

For example, the site tries to load `/static/netbox.js.map` instead of `/netbox/static/netbox.js.map`. The problem appears to be this line: https://github.com/netbox-community/netbox/blob/6845fb0f005f0bdc15a258ae2ebd7f9be5341f76/netbox/project-static/dist/netbox.js#L67

Perhaps due to this?
https://github.com/netbox-community/netbox/blob/6845fb0f005f0bdc15a258ae2ebd7f9be5341f76/netbox/project-static/bundle.js#L11

Changing `sourceMappingURL=/static/netbox.js.map` to `sourceMappingURL=netbox.js.map` appears to work, at least for this particular file.

jeremystretch:
### Fixes: #6912

- Remove the `publicPath` bundle option so that relative paths are used for all static assets


thatmattlove (approved):
From what I can tell in my local tests, this doesn't cause any issues. It's odd, I remember _having_ to add it in order to get source maps to be referenced properly, but hey, it works.

Issue #6962, opened 2021-08-15
Developer doc for v3: point out "yarn typecheck"


candlerb:
### Change Type

Addition

### Area

Administration/development

### Proposed Changes

Mention `yarn typecheck` in netbox/project-static/README.md

`yarn bundle:scripts` [does not do type checking](https://esbuild.github.io/content-types/#typescript) so an extra step is required to validate typescript.

thatmattlove:
c221b9b should address this — there's now a Web UI section in the docs that explains much of the front end stack and documents available commands. Additionally, #7024 proposes running these checks via standard development processes (pre-commit hooks and CI).

PR #6963, opened 2021-08-15
Reference "yarn typecheck" in developer documentation

candlerb:
### Fixes: #6962

Point out the existence of "yarn typecheck", and that this needs to be run as a separate step


jeremystretch:
The proposed note doesn't explain what the command does or why it's necessary. But rather than call it out as a separate step, it should be run automatically. Let's let @thatmattlove weigh in on #6962 since he wrote the bundler script.

candlerb:
The disadvantage with running typecheck within the bundle step is that it's slooooow, and that would make the modify/test/debug cycle more painful.

Issue #7035, opened 2021-08-25
Interface VLAN assignment selection not filtered by selected VLAN group


jeremystretch:
### NetBox version

v3.0-beta2

### Python version

3.7

### Steps to Reproduce

1. Create VLAN groups named A and B
2. Create a VLAN in group A named "Foo" with VLAN ID 100
3. Create a VLAN in group B named "Bar" with VLAN ID 100
4. Create or edit a device interface
5. Select "access" for the interface's 802.1Q mode
6. Select VLAN group A
7. View the available VLANs

### Expected Behavior

Only VLAN Foo (in group A) should be listed

### Observed Behavior

Both VLANs Foo and Bar are listed. This is because the API request is made with the query parameter `?vlan_group_id=` rather than `?group_id=`.

PR #7040, opened 2021-08-26
Improve APISelect query parameter handling

### LINKED ISSUE #7034: Changing VLAN group scope type does not toggle selection fields correctly
### NetBox version

v3.0-beta2

### Python version

3.7

### Steps to Reproduce

1. Navigate to IPAM > VLAN Groups > Add
2. Select several different scope types and observe the selection fields below

### Expected Behavior

Initially, none of the selection fields below scope type should be present. Once a scope type has been selected, only the relevant selection fields should be displayed.

For example, if the scope type has been set to "site," only the region, site group, and site fields should be visible.

### Observed Behavior

Initially, all scope selection fields are visible instead of hidden. After selecting a scope type, the form widgets are hidden, however the parent form "rows" (with labels) are not.

thatmattlove:
### Fixes: #7034, #7035

- Replaced `query_params` on `DynamicModelChoiceMixin` with `filter_fields`, which carries a more strict data model to define how the field's related fields should be handled. The resulting data is serialized to JSON and added as a single `data-filter-fields` attribute on the `<select/>` element, and then appropriately handled in the UI.
- Refactor APISelect TypeScript class to handle new logic — parsing of the `data-filter-fields` into an expected structure seems to be much simpler than the previous method. Or at least, it's easier for me follow :)
- `query_params` is still available on the widget to add **static** values only.
- Restructured former `netbox/project-static/src/select/api.ts` to a directory to break up the 1000+ LOC into multiple files for better readability.
- Fixed #7034 which was primarily broken due to how we've restructured the form layout (we needed to hide the field's parent's parent, not the immediate parent). Secondarily, the `filter_fields` changes also fix the filtering itself.
- Updated `extras.ObjectVar` docs to reflect addition of `filter_fields`.

All the testing I can think to do passes perfectly, from what I can tell.


jeremystretch (changes requested):

thatmattlove:
Probably a hella messy commit history, but, this should contain the fixes we discussed. Nutshell: the current `query_params` API is maintained, all the transformation logic occurs in the widget.

jeremystretch (approved):

jeremystretch:
Awesome work, thanks @thatmattlove!

PR #7053, opened 2021-08-27
Close #7022

### LINKED ISSUE #7022: Add new power port CEE 7/16 (Type C) a.k.a. Europlug
### NetBox version

v2.11.12

### Feature type

### Proposed functionality

Add a new power connector type for selection in netbox: CEE 7/16, Type C (also known as Europlug).
This connector type is very common across Europe for low-power devices like DSL modems, USB PowerSupply, etc.

Details can be found on https://en.wikipedia.org/wiki/CEE_7_standard_AC_plugs_and_sockets#CEE_7/16_plugs

If accepted, I'll try to create the proper PR for it.

### Use case

Support the selection of CEE 7/16 (Type C) "Europlug" for power connections in netbox.

### Database changes

none

### External dependencies

none

tb-killa:
### Fixes: #7022
Add a new power connector type for selection in netbox: CEE 7/16, Type C (also known as Europlug).

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -246,6 +246,8 @@ class ConsolePortSpeedChoices(ChoiceSet):
 
 class PowerPortTypeChoices(ChoiceSet):
 
+    # CEE 7/16
+    TYPE_CEE_7_16_TYPE_C = 'cee-7-16-type-c'
     # IEC 60320
     TYPE_IEC_C6 = 'iec-60320-c6'
     TYPE_IEC_C8 = 'iec-60320-c8'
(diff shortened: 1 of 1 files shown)


jeremystretch (commented):
A couple thoughts:

1. CEE should be a group (not specifically CEE 7/16) if we're going to add this.
2. The CEE group should come _after_ the IEC types.
3. We already have a few CEE types listed under their ITA nomenclature. I'm not sure whether it makes sense to move or replicate these, if either.

jeremystretch:
I just added this under the "ITA/international heading" since it has an ITA designation.

PR #7110, opened 2021-09-01
When handling an exception during an API query, send an error response containing structured information about the error rather than HTML

### LINKED ISSUE #7109: Uncaught Exceptions from within `ModelViewSet` are returned as HTML
### NetBox version

v3.0.0

### Python version

3.8

### Steps to Reproduce

This can be tricky - something needs to done to trigger an uncaught exception from within `ModelViewSet`. I suppose you could add this...

```python
raise Exception('test')
```

...[here](https://github.com/netbox-community/netbox/blob/d743dc160ab701cb6b92579b5ff164b5ec2eaa9b/netbox/netbox/api/views.py#L199), and then make an API request.

### Expected Behavior

At a minimum, I think we should append useful information to the error. Ideally, however, no HTML would be returned to an API client.

### Observed Behavior

The returned value is something along the lines of:

```json
{
  "error": "<giant unreadable HTML string>"
}
```

thatmattlove:
### Fixes: #7109

See #7109 for more details. There seemed to be an already-written handler for such cases, this simply reuses it in the `ModelViewSet` view as a fallback.


jeremystretch:
Thanks for this. Even simpler, we can just check for the requested content type in the middleware and act accordingly (see my comment on #7109). Implemented an alternative fix in d2fe59ae8f9c34c1f8a7d51c0d113348d7bed937

PR #7172, opened 2021-09-03
Fixes #7162: Fix API-populated form fields when BASE_PATH is defined

### LINKED ISSUE #7162: Failed to load resource, 404 while using BASE_PATH
### NetBox version

v3.0.1

### Python version

3.9

### Steps to Reproduce

1. Go to IPAM -> VLANs
2. Click "Add"

### Expected Behavior

Get all of searching fields like: "Tags", "Role", "VLAN Group"

### Observed Behavior

I used BASE_PATH = 'netbox/'
Get a red box and api error. 
netbox.js:8 GET https://localhost/netbox/netbox/api/ipam/roles/?brief=true 404 (Not Found)
I think the problem is in the link being created, it is duplicated.

When i removed BASE_PATH and use netbox like https://localhost/. 
Link be like: https://localhost/api/ipam/roles/?brief=true and no problems further

jeremystretch:
### Fixes: #7162

- Remove the call to `buildUrl()` from `apiRequest`
- Pass the table config API endpoint through `buildUrl()` within `submitFormConfig`


thatmattlove:
Will this not break all other requests when `BASE_PATH` is defined? I’m not able to test in detail at the moment, but I believe all requests need to go through `buildUrl()`, otherwise `BASE_PATH` won’t be picked up.

jeremystretch:
It's not needed for the APISelect form fields, because `BASE_PATH` is already accounted for when resolving the API endpoint URL (which is why it's getting duplicated in the path currently).

thatmattlove:
That's a great point, I didn't have any logic to exclude `BASE_PATH` if it was already present in the URL. While APISelect shouldn't break, I believe this implementation _will_ break Device Status, LLDP, & Config, as well as Job statuses as all of those call `apiGetBase()` directly, which will now not include the `BASE_PATH` on those requests. I'd rather not have to include the extra `buildUrl()` function call on every function that also has to call the API without a server-supplied URL. I think I can modify the current logic to ensure that it doesn't matter if the supplied URL contains `BASE_PATH` or not, so that it just works. I'll commit a few changes to this PR that we can review shortly.

thatmattlove:
I confirmed in 70cd096 that the issue from #7162 is still fixed, and that Device Status/LLDP/Configs still work (I did not directly test Reports/Jobs, but the logic is identical).

netbox/project-static/src/util.ts
```diff
@@ -162,10 +163,15 @@ function buildUrl(destination: string): string {
 
   // If the `origin` exists in the API path (as in the case of paginated responses), remove it.
   const origin = new RegExp(window.location.origin, 'g');
-  const path = pathname.replaceAll(origin, '');
+  let path = pathname.replaceAll(origin, '');
 
   const basePath = getBasePath();
 
+  // If the `BASE_PATH` already exists in the URL, remove it.
+  if (basePath !== '' && path.includes(basePath)) {
```
jeremystretch, line 171: I'm not a fan of this approach because IMO it's too vulnerable to false positives. For example, what if a user sets `BASE_PATH = 'ipam/'` (because NetBox is their IPAM app)? That would break IPAM API endpoints.

I'd much prefer to either _always_ prepend the base path, or _never_ prepend it.

thatmattlove: > I'd much prefer to either always prepend the base path, or never prepend it.

I completely agree with this. I didn't consider there being a naming conflict between netbox paths and the base path, let me take another swing at accounting for that.

jeremystretch: Was there an issue with the approach I took? It seemed to take a very light touch, and invokes `buildUrl()` only where it's needed.

netbox/project-static/src/util.ts
```diff
@@ -163,17 +163,21 @@ function buildUrl(destination: string): string {
...
+    .split('/')
+    .filter(p => p);
 
   const basePath = getBasePath();
 
-  // If the `BASE_PATH` already exists in the URL, remove it.
-  if (basePath !== '' && path.includes(basePath)) {
-    path = path.replaceAll(basePath, '');
+  // If the `BASE_PATH` already exists in the URL, and it is the first element, remove it.
+  if (basePath !== '' && path[0].includes(basePath)) {
```
jeremystretch, line 174: Like my previous comment though, this should never be necessary. We should be able to assume at this point that the URL being passed has already been processed to accommodate `BASE_PATH` and proceed accordingly.

thatmattlove:
> Was there an issue with the approach I took? It seemed to take a very light touch, and invokes buildUrl() only where it's needed.

The problem with the original approach is that it didn't cover all API requests made by the UI. Device Status/LLDP/Config and Reports/Jobs also communicate with the API and would need to call `buildUrl()`. In fact, the only request made from the API that _doesn't_, as far as I can tell, is `APISelect`.

To me, it makes more sense to ensure that we don't have to care. The implementation from c35563d ensures that regardless of the input URL, the output URL is correct. For example:

| Input URL          | NetBox Path                   | Output URL                                    |
| ------------------ | ----------------------------- | --------------------------------------------- |
| `/api/ipam/vlans/` | `''`                          | `/api/ipam/vlans/`                            |
| `/api/ipam/vlans/` | `netbox/`                     | `/netbox/api/ipam/vlans/`                     |
| `/api/ipam/vlans/` | `ipam/`                       | `/ipam/api/ipam/vlans/`                       |
| `/api/ipam/vlans/` | `ipam/netbox/vlans/are/cool/` | `/ipam/netbox/vlans/are/cool/api/ipam/vlans/` |

IMO, calling `buildUrl()` before _every_ API call except one is a heavy touch. Over time, I would expect the number of UI → API calls to increase, not decrease, and I can _easily_ see bugs being introduced by contributors who don't think about `BASE_PATH` and forget to add `buildUrl()` before an API request. Personally, I think we should make that irrelevant by handling it transparently.

However, if you're still not satisfied with the approach from c35563d, here are a couple other ideas:

1. Have a separate API function for `APISelect` that doesn't call `buildUrl()`.
2. Add an option to `apiRequest` and all other API functions to _not_ build the URL, when needed (which should be exclusively in `APISelect`).

jeremystretch:
> The problem with the original approach is that it didn't cover all API requests made by the UI. Device Status/LLDP/Config and Reports/Jobs also communicate with the API and would need to call buildUrl().

These other calls are using a URL that has already been processed by the backend to include `BASE_PATH`. For example, with `BASE_PATH = 'foo/'` configured, the device config page renders with

```
<span data-object-url="/foo/api/dcim/devices/14/napalm/?method=get_config"></span>
```

Anywhere we're calling `reverse()` to resolve a URL in the backend, it's taking `BASE_PATH` into account.

thatmattlove:
> These other calls are using a URL that has already been processed by the backend to include `BASE_PATH`. For example, with `BASE_PATH = 'foo/'` configured, the device config page renders with

My mistake, I forgot about that implementation for Device Status/Config/LLDP. However, there are three implementations that do not use this approach:

1. [Job Results](https://github.com/netbox-community/netbox/blob/d40d1638aff12af31ee015279fe4a9b673384b15/netbox/project-static/src/jobs.ts#L65)
2. [Table Config](https://github.com/netbox-community/netbox/blob/d40d1638aff12af31ee015279fe4a9b673384b15/netbox/project-static/src/tableConfig.ts#L57)
3. [Connection Toggles](https://github.com/netbox-community/netbox/blob/d40d1638aff12af31ee015279fe4a9b673384b15/netbox/project-static/src/buttons/connectionToggle.ts#L16)

Now, if we want to build in server-side solutions for these to pick up the post-processed URL from the server and just remove all logic around `BASE_PATH` from the frontend entirely, I'm good with that. But if there's _ever_ going to be a case where that's not feasible, then we need something in place to handle it.

I still believe an arbitrary extra function call before is a form of technical debt, but, the project's benevolent dictator can make the final call 😉

jeremystretch:
Booting this in favor of #7218 (for the same issue).

PR #7183, opened 2021-09-04
feat: add yarn dev watcher script

ndom91:

Added a javascript script in the root of `/netbox/project-static` in order to make the frontend development workflow a bit nicer.

You can now run `yarn dev` and it'll begin watching `../project-static/src/**/*.ts` and `../project-static/styles/**/*.scss` for changes. When any of the above files is detected to have changed, it'll automatically run the appropriate transpile step (`bundle.js`) and also the `manage.py collectstatic` step. 

Therefore, the developer just makes the changes to the sass / ts files, waits a second or so and reloads their development netbox tab to see the updated changes. This gets us a bit closer to the normal modern js development flow of hot reloading everything :tada: 

This does require 1 new dependency, `chokidar` a node file watcher, but it's only a dev dependency and so is only installed during development and is not part of the bundle that gets sent to netbox users. 

Example output:
image

CC: @thatmattlove


jeremystretch:
> and also the `manage.py collectstatic` step

I'm not sure this is piece is needed. When running the Django development server (`manage.py runserver`), it will automatically serve static assets directly from their original location. I don't think it's necessary to run `collectstatic`, at least for my workflow, but would like to hear from @thatmattlove and anyone else with input.

ndom91:
Ah so the django dev server serves the bundled js / css files from `../netbox/project-static/dist` ? I was always under the impression they had to be copied back to the normal static directory, i.e. run `collectstatic`.

jeremystretch:
> I was always under the impression they had to be copied back to the normal static directory

They do, ultimately, but only for serving via a production HTTP daemon (e.g. nginx/Apache). Django's `runserver` is clever enough to look in the various static directories for files when processing a request for assets within the static path.

ndom91:
Ah gotcha so yuo actually don't even need the nginx proxy in front during dev? I dont know why but I thought that was required.

Okay great, I'll remove that part then. My dev flow has been significantly simplified after this chat haha

ndom91:
@thatmattlove looping ya back in here, does this make sense to you? Would you find it helpful as well?

thatmattlove:
I'm of two minds on this.

On one hand, it would be nice not to have to run the build commands and re-run `collectstatic` (I know @jeremystretch said this isn't necessary, however I found that I had to run it for some reason, and now it's just part of my workflow).

On the other hand, due primarily to the tooling itself, I can also see how this could get cumbersome. In many cases, it may be frustrating to have to wait on the bundle process (particularly for SCSS, which is significantly longer than JS/TS bundling) every time a file is changed, when you may or may not be ready to preview your changes. Unfortunately, we don't (yet) have a snazzy React app with CSS-in-JS and all the fancy DX-focused toolchains to hot-reload changes made in real-time. However, that's not to say it's not worth trying out.

That said, should we choose to accept the proposal to begin with, my primary concern with the PR is: is it *absolutely necessary* to add the chokidar dependency? I wholeheartedly agree with NetBox's [dependency policy](https://netbox.readthedocs.io/en/stable/development/style-guide/#introducing-new-dependencies), and I don't think the policy should exclude developer toolsets. esbuild has a [watch feature](https://esbuild.github.io/api/#watch), is there any reason we couldn't use it? I haven't personally used it, and I'm not sure how it would work with plugins (SCSS). That may eliminate the need for a separate file entirely, which would be nice.

ndom91:
Hey thanks for the detailed answer @thatmattlove. Im in agreement, thats the point of this PR, to get us closer to one of thsoe snazzy hot reloading setups. Regarding if its necessary to have the `collectstatic` part, I'm not positive. It doesn't seem to be techncially although I've had to run it often and have kind of built it into my workflow as well.

In regard to `chokidar` being absolutely necessary - probably not if esbuild also has a nice watch feature. Upon a cursory glance of their docs, it doesn't seem you can watch individual files programmatically though so upon changes to any file, it'll reexecute the `bundleAll` cmd. Chokidar would give us a bit more granular control, plus its a dev dependency only - not shipped to end users. But obviously I can understand if yall don't want it, just let me know and I can remove it and try to work out a nice version using esbuilds watch feature :+1:

thatmattlove:
Hey @ndom91, I think if we remove chokidar and use esbuild natively, I don't have a problem merging this. Can you take a swing at it and see how it goes?

ndom91:
Hey sorry for taking so long yall, I'll have a stab at removing `chokidar` and using `esbuild`'s watch natively this week :+1:

ndom91:
@thatmattlove sorry this took me forever to around to! I ended up adjusting the existing `bundle.js` and added a `--watch` flag to enable the builtin esbuild file watching. Hope this makes it a bit easier for yall!

thatmattlove (commented):
I've exposed this to `manage.py` to further improve the experience. Now, running `manage.py dev` runs `yarn dev`, `manage.py collectstatic --noinput`, and `manage.py runserver` and inherits all the arguments from `django.contrib.staticfiles.management.commands.runserver.Command`

Edit: I also rebased this with `develop` to make things a little cleaner

thatmattlove:
Hey @jeremystretch just a reminder about this one. Since I ended up adding this to `manage.py`, I figured you'd want to take a look

netbox/extras/management/commands/dev.py
```diff
@@ -0,0 +1,22 @@
+import typing as t
+import threading
+from django.core.management import call_command
+from django.contrib.staticfiles.management.commands import runserver
+
+
+class Command(runserver.Command):
```
jeremystretch, line 7: Rather than introduce a command that's largely redundant to `runserver`, we can override `runserver` (as we do with `migrate` and `makemigrations`, under `utilities`) to add an optional command line argument that spawns the UI dev server. E.g. `manage.py runserver --with-dev-ui` or something like that.

netbox/extras/management/commands/dev.py
```diff
@@ -0,0 +1,22 @@
...
+    watches UI source files for changes.
+    """
+
+    def handle(self, *args: t.Any, **options: t.Any) -> t.Optional[str]:
+        # Run the UI development server in a separate thread so file watching doesn't block the
+        # Django development server running in the main thread.
+        ui_dev = threading.Thread(name="uidev", target=lambda: call_command("uidev"))
+        ui_dev.start()
+        # Run `collectstatic`.
+        call_command("collectstatic", interactive=False)
```
jeremystretch, line 20: Per the previous discussion, `collectstatic` should not be needed, as the `runserver` daemon [automatically serves static files](https://docs.djangoproject.com/en/4.0/ref/contrib/staticfiles/#staticfiles-runserver) from their source locations. (If this does not appear to be working locally, further investigation is needed to determine the root cause.)

jeremystretch (changes requested):

PR #7214, opened 2021-09-08
Update Apache setup

hanserasmus:
If you don't disable the default apache site, you get the default Ubuntu Apache site. This may confuse users.


jeremystretch:
I've never needed to do this when installing NetBox using Apache. If you believe it's necessary, please submit an issue for discussion.

PR #7218, opened 2021-09-08
Fixes #7162: Decouple base path rendering from API request logic

### LINKED ISSUE #7162: Failed to load resource, 404 while using BASE_PATH

jeremystretch:
### Fixes: #7162

This is a more fleshed out implementation of my initial proposal under PR #7172. This removes the need to ever detect or calculate the `BASE_PATH` (if configured) in the front end. All URLs are rendered in the back end and passed via context in the template for reference by the front end.


thatmattlove (approved):
Based on semi-limited testing, looks good to me!

PR #7307, opened 2021-09-20
Fixes #7228: Round °F temperature to one decimal place

### LINKED ISSUE #7228: Rounding required for °F temperature conversion in Napalm device status tab
### NetBox version

v3.0.2

### Python version

3.8

### Steps to Reproduce

1. Configure napalm
2. Go to a napalm-enabled device which reports temperature (mine's a Mikrotik RB4011)
3. Go to the Status tab: `/dcim/devices/<N>/status/`
4. Look at the °F temperature conversion

### Expected Behavior

°F temperature to one decimal place at most.

### Observed Behavior

I see `42 °C 107.60000000000001 °F`

image

This isn't always the case - e.g. on another device I see `51 °C 123.8 °F`

royreznik:
### Fixes: #7228 
Rounding the °F temperature to the first decimal place

before:
image

after:
temp


netbox/project-static/src/util.ts
```diff
@@ -409,7 +409,7 @@ export function createElement<
  * @returns Degrees in Fahrenheit.
  */
 export function cToF(celsius: number): number {
-  return celsius * (9 / 5) + 32;
+  return Math.round((celsius * (9 / 5) + 32 + Number.EPSILON) * 10) / 10;
```
jeremystretch, line 412: This seems a bit much. Shouldn't calling `round()` be sufficient?

royreznik: `round()` will just round the number to the next integer, don't we want to save that one decimal place?

also using `Number.EPSILON` to be more presise, take a look at this example:
```javascript
Math.round((1.005 + Number.EPSILON) * 100) / 100 // == 1.01
Math.round((1.005) * 100) / 100 // == 1
```

candlerb, line 412: Does this actually work? Certain numbers cannot be represented exactly using floating point (0.1 is one example); rounding doesn't change this.

I would expect the problem is at the point of rendering it back to decimal.

royreznik: Can you please provide an example for this?

I checked it with `0.1` and `0.3` and still got the expected value
```javascript
Math.round((0.1 + Number.EPSILON) * 10) / 10; # 0.1
Math.round((0.3 + Number.EPSILON) * 10) / 10; # 0.3
```

candlerb: OK, I think I can replicate via Javascript console in Chrome:

image

The problem is indeed fixed by your method, but actually it would be simpler to multiply by 9 and then divide by 5, rather than multiply by (9/5) which is an inexact floating point value; it appears that the error in (9/5) is magnified by multiplying by the degrees C value.

candlerb: There's no guarantee that doing *9/5 will give the right answer though. I believe the correct solution is to convert the number to a string with the correct number of decimal places, using `toFixed`:

```
> (42 * 1.8)
< 75.60000000000001
> (42 * 1.8).toFixed(2)
< '75.60'
> (42 * 1.8).toFixed(1)
< '75.6'

Comment (royreznik) [reply netbox/project-static/src/util.ts:412]: As you said, `There's no guarantee that doing *9/5 will give the right answer though` i don't think it's a good idea

using only `toFixed` can result a different inacurate like this:
```javascript
(1.005).toFixed(2) # 1.00
```

candlerb: Yes you're right:

```
> (1.006).toFixed(2)
< '1.01'
> (1.005).toFixed(2)
< '1.00'
> (1.00501).toFixed(2)
< '1.01'
```

and yet:

```
> (1.05).toFixed(1)
< '1.1'
```

Floating point is [broken](https://0.30000000000000004.com/), and Javascript especially so :-(

However I think toFixed is OK here.  I just want never to see a number with spurious 13 places of decimals.

royreznik: I see what did you mean. Floating point is the evil :-(

also `toFixed` will do this:
```javascript
(1).toFixed(1) // "1.0 "
Math.round((1 + Number.EPSILON) * 10) / 10; //  1
````
this will add `.0` for integers.

Anyway, I think my solution fix the bug we had and maybe different approch can be achived in other PR :)

PR #7308, opened 2021-09-20
Fix #7118: Make custom field URLs clickable in tables

### LINKED ISSUE #7118: Clickable URLs in custom field
### NetBox version

v3.0.0

### Feature type

### Proposed functionality

When viewing the `ipam/services/` table with a custom URL field on services I want to be able to click the URL to open it in a new tab.

### Use case

Currently, you have to go to the service details to have a clickable URL for custom fields.
It would be faster if I could go to the custom field URL directly from the services table.

### Database changes

_No response_

### External dependencies

_No response_

Paktosan:
### Fixes: #7118

These changes set `linkify=True` for the table column when the custom field is of the URL type.

### DIFF
--- a/netbox/utilities/tables.py
+++ b/netbox/utilities/tables.py
@@ -11,6 +11,7 @@ from django_tables2 import RequestConfig
 from django_tables2.data import TableQuerysetData
 from django_tables2.utils import Accessor
 
+from extras.choices import CustomFieldTypeChoices
 from extras.models import CustomField
 from .utils import content_type_name
 from .paginator import EnhancedPaginator, get_paginate_count
@@ -344,11 +345,15 @@ class CustomFieldColumn(tables.Column):
     """
     Display custom fields in the appropriate format.
     """
+
     def __init__(self, customfield, *args, **kwargs):
         self.customfield = customfield
         kwargs['accessor'] = Accessor(f'custom_field_data__{customfield.name}')
         if 'verbose_name' not in kwargs:
             kwargs['verbose_name'] = customfield.label or customfield.name
+
+        linkify = bool(customfield.type == CustomFieldTypeChoices.TYPE_URL)
+        kwargs.setdefault('linkify', linkify)
 
         super().__init__(*args, **kwargs)
 


Paktosan:
At this point I would like to thank @jeremystretch for his Coding Livestreams which gave me the insight needed for doing this.

netbox/utilities/tables.py
```diff
@@ -344,13 +345,17 @@ class CustomFieldColumn(tables.Column):
...
     """
+
     def __init__(self, customfield, *args, **kwargs):
         self.customfield = customfield
         kwargs['accessor'] = Accessor(f'custom_field_data__{customfield.name}')
         if 'verbose_name' not in kwargs:
             kwargs['verbose_name'] = customfield.label or customfield.name
 
-        super().__init__(*args, **kwargs)
+        if customfield.type == CustomFieldTypeChoices.TYPE_URL:
```
jeremystretch, line 355: You've got the right idea! I would just suggest a minor tweak in execution. Ideally, we want to default to linkifying this value for URL fields, but still ultimately allow it to be disabled upon instantiation if desired.

So instead of setting the `linkify` kwarg directly, we can do something like this:

```
linkify = bool(customfield.type == CustomFieldTypeChoices.TYPE_URL)
kwargs.setdefault('linkify', linkify)
```

`setdefault` will set `linkify` to True only if the argument hasn't been specified; otherwise, it will be left alone.

Then we can just call `super().__init__(*args, **kwargs)` normally.

jeremystretch (changes requested):
Thanks for digging into this!

Paktosan:
For some reason, when I worked on it today on a different machine it was not functioning properly.
While the custom column looks completely fine, the link seems to be malformed and always show to the address of the object.
Example: `<a href="/virtualization/clusters/3/">https://google.com</a>`
I am pretty sure I have not seen this behavior on the machine I worked on yesterday.
Unfortunately, I am clueless why this is happening now. :(

jeremystretch:
Sorry, I think I lost sight of what you were trying to accomplish in my initial review. Setting `linkify=True` on the table column indeed links to the object itself. To embed an arbitrary hyperlink, we need to generate the hyperlink ourselves within the `render()` method. I can take care of this real quick. Thanks again for giving it a shot!

PR #7440, opened 2021-10-05
Added permission check /api/dcim/connected-devices

xraurp:
### Fixes: netbox-community/netbox#7051

Affects `/api/dcim/connected-device/`

Api now checks `dcim.view_device` and `dcim.view_interface` permissions when login is required.
If login is not required, endpoint is available without permissions as before.
Set `LOGIN_REQUIRED = True` in configuration for this change to take effect.

Error message when permission is not sufficient:
```
HTTP 403 Forbidden
Allow: GET, HEAD, OPTIONS
Vary: Accept

{
    "detail": "You do not have permission to perform this action."
}
```

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -8,6 +8,7 @@ from drf_yasg import openapi
 from drf_yasg.openapi import Parameter
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework.decorators import action
+from rest_framework.exceptions import PermissionDenied
 from rest_framework.response import Response
 from rest_framework.routers import APIRootView
 from rest_framework.viewsets import ViewSet
@@ -672,6 +673,12 @@ class ConnectedDeviceViewSet(ViewSet):
         peer_device_name = request.query_params.get(self._device_param.name)
         peer_interface_name = request.query_params.get(self._interface_param.name)
 
+        # check permissions of user, if login is required
+        if settings.LOGIN_REQUIRED:
+            if not request.user.has_perm('dcim.view_device') or not request.user.has_perm('dcim.view_interface'):
+                raise PermissionDenied('You do not have permission to perform this action.')
+
+        # check query parameters
         if not peer_device_name or not peer_interface_name:
             raise MissingFilterException(detail='Request must include "peer_device" and "peer_interface" filters.')
 


netbox/dcim/api/views.py
```diff
@@ -672,6 +673,12 @@ def list(self, request):
         peer_device_name = request.query_params.get(self._device_param.name)
         peer_interface_name = request.query_params.get(self._interface_param.name)
 
+        # check permissions of user, if login is required
+        if settings.LOGIN_REQUIRED:
```
jeremystretch, line 677: Evaluation of `LOGIN_REQUIRED` is handled by middleware, so this line isn't needed. (We also want to check permissions regardless of how `LOGIN_REQUIRED` is set.)

netbox/dcim/api/views.py
```diff
@@ -672,6 +673,12 @@ def list(self, request):
...
+            if not request.user.has_perm('dcim.view_device') or not request.user.has_perm('dcim.view_interface'):
```
jeremystretch, line 678: Instead of calling `has_perm()` twice, we can pass a list of both permissions to `request.user.has_perms()`.

netbox/dcim/api/views.py
```diff
@@ -672,6 +673,12 @@ def list(self, request):
...
+                raise PermissionDenied('You do not have permission to perform this action.')
```
jeremystretch, line 679: No need to specify an error message when raising `PermissionDenied`. It includes a suitable default message.

jeremystretch (changes requested):
Thanks for taking this on! You're on the right track, just needs to be simplified a bit.

netbox/dcim/api/views.py
xraurp: OK, thanks.

netbox/dcim/api/views.py
xraurp: Ok, thanks. I didn't notice that.

netbox/dcim/api/views.py
jeremystretch: It depends how NetBox is configured. It's simply not necessary to check `LOGIN_REQUIRED` when evaluating permissions.

xraurp: I think that, if the `LOGIN_REQUIRED` should be checked here depends on, if the unauthenticated users should have access or not. It allows them to pass without permission check, when login is not required. Otherwise they are blocked by permission class `IsAuthenticatedOrLoginNotRequired` before they even get here.

If this line is deleted and only 2 following lines are kept, unauthenticated user will always get:
1. `Authentication credentials were not provided` when `LOGIN_REQUIRED = True` (this is ok)
2. `You do not have permission to perform this action` when `LOGIN_REQUIRED = False` (this is the problem)

No way for him to access the resource.

So i think the options are:
1. Users should be always authenticated.
2. `LOGIN_REQUIRED` should be checked here before permissions are checked.
3. I missed something in settings and documentation, that allows setting permissions to unauthenticated users.

If you have other suggestion, please let me know.

PR #7575, opened 2021-10-18
Closes #1344: Contact objects

### LINKED ISSUE #1344: Contact Objects
### Issue type: Feature Request

This may well be out of scope for Netbox as a technical application but thought i'd gauge interest in this feature.

I find myself managing several sites with the same contact. When that contact for example changes an email address, phone number, or someone takes over their role i am updating all the sites with that contact listed.

If this is changed to a contact object that we then attach to these sites we have a central place to manage these people without duplicating the data in separate sites.

From here we could add functionality such as adding several contacts to a site (primary, secondary), we could stipulate between a business contact or a technical contact, add which company they work for, have an api get request that lists contacts for a child object such as devices (useful for managing outages and notifications).

Let me know your thoughts.

jeremystretch:
### Closes: #1344

Introduces the following models within the tenancy app:

* ContactGroup - Self-nesting organizational model for contacts (similar to SiteGroup)
* Contact - Represents a reusable set of contact info (name, title, phone, email, etc.)
* ContactRole - Functional role of the relationship between a contact and an object in NetBox
* ContactAssignment - Binds a mapping of object, contact, and role; also conveys priority


DanSheps (approved):
LGTM, will test in an instance later hopefully but nothing glaring.

PR #7629, opened 2021-10-25
Fix #7589: Rename 128gfc-sfp28 to 128gfc-qsfp28

### LINKED ISSUE #7589: Rename optic type 128gfc-sfp28 to 128gfc-qsfp28 (Fibrechannel)
### NetBox version

v3.0.7

### Python version

3.8

### Steps to Reproduce

Currently the 128G FC optic type is named `128gfc-sfp28`, which is wrong as it's not a SFP28 formfactor but a QSFP28.

### Expected Behavior

The optic type should be `128gfc-qsfp28` to be correct with the type of optic it is. 

### Observed Behavior

Problem was noticed when working on [netbox-community/devicetype-library#503](https://github.com/netbox-community/devicetype-library/pull/503).

suom1:
### Fixes: #7589
Renames 128gfc-sfp28 to 128gfc-qsfp28, this because QSFP28 is the right name for the optic type.

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -760,7 +760,7 @@ class InterfaceTypeChoices(ChoiceSet):
     TYPE_16GFC_SFP_PLUS = '16gfc-sfpp'
     TYPE_32GFC_SFP28 = '32gfc-sfp28'
     TYPE_64GFC_QSFP_PLUS = '64gfc-qsfpp'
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
Merely changing the string value is insufficient. A database migration is needed as well, to ensure any saved values are updated appropriately.

jeremystretch:
This has been fixed with a migration in in b7b5a5788

Issue #6926, opened 2021-08-10
LDAP group caching no longer works after upgrade to v2.11.10


koratfood:
### NetBox version

v2.11.10

### Python version

3.8

### Steps to Reproduce

1. Install a new instance of Netbox v2.11.9 (or earlier) with LDAP authentication. Use LDAP config params towards a working LDAP server as suggested by official Netbox documentation, except the following:

```
AUTH_LDAP_FIND_GROUP_PERMS = False
AUTH_LDAP_CACHE_GROUPS = True
AUTH_LDAP_CACHE_TIMEOUT = 60
AUTH_LDAP_GROUP_CACHE_TIMEOUT = 60
AUTH_LDAP_REQUIRE_GROUP = "CN=Netbox Users,OU=Groups,DC=mycompany,DC=com"
AUTH_LDAP_MIRROR_GROUPS = ["Netbox Admins", "Netbox Users"]
AUTH_LDAP_USER_FLAGS_BY_GROUP = {
        "is_superuser": ["CN=Netbox Admins,OU=Groups,DC=mycompany,DC=com"]
}
```

Substitute domain and group names as appropriate, and make sure the LDAP account you use for testing is a member of 'Netbox Admins' as well as 'Netbox Users'.

2. Log in and create an API token for yourself, with "write enabled" and no expiration date.
3. POST an object via the REST-API (just to rule out potential hiccups with initial POST-request after upgrade)
```
curl -X POST \
-H "Authorization: Token <insert your token here>" \
-H "Content-Type: application/json" \
-H "Accept: application/json; indent=4" \
https://my-netbox-server/api/ipam/ip-addresses/ \
--data '{
    "address": "10.0.0.1/24",
	"description": "My IP Host 1"
}'
```
4. POST another object via the REST-API, and take note of how long it takes to complete
```
curl -X POST \
-H "Authorization: Token <insert your token here>" \
-H "Content-Type: application/json" \
-H "Accept: application/json; indent=4" \
https://my-netbox-server/api/ipam/ip-addresses/ \
--data '{
    "address": "10.0.0.2/24",
	"description": "My IP Host 2"
}'
```
5. Upgrade Netbox instance to v2.11.10 in accordance with official documentation
6. POST an object via the REST-API (just to rule out potential hiccups with initial POST-request after upgrade)
```
curl -X POST \
-H "Authorization: Token <insert your token here>" \
-H "Content-Type: application/json" \
-H "Accept: application/json; indent=4" \
[...]

PR #7676, opened 2021-10-28
Fix #7399: LDAP excessive CPU usage when AUTH_LDAP_FIND_GROUP_PERMS is enabled

### LINKED ISSUE #7399: gunicorn spends 80% of each request in get_object_permissions()
### NetBox version

v3.0.3

### Python version

3.9

### Steps to Reproduce

1. Configure Netbox with AD authentication (a few thousands users + heavy nesting)
2. Create a bunch of devices (~500 servers in my case, ~500 cables, ~50 VLANs, ~500 IP addresses, ~40 prefixes, ~40 racks in 3 sites)
3. Try to access https://netbox.example.com/dcim/devices/1234
4. See that gunicorn process is suddenly eats 100% CPU and the request takes more than a second

### Expected Behavior

The request should be completed way faster.

### Observed Behavior

For me the situtation looks like this: the HTTP request takes ~1.6s to complete and ~1.3s of that request was spent in authentication.py:get_object_permissions(). I am terrible with Python, but I came up with a patch:

```
@@ -11,6 +11,9 @@ from django.db.models import Q
 from users.models import ObjectPermission
 from utilities.permissions import permission_is_exempt, resolve_permission, resolve_permission_ct

+import time
+import syslog
+
 UserModel = get_user_model()


@@ -38,11 +42,18 @@ class ObjectPermissionMixin():

         # Create a dictionary mapping permissions to their constraints
         perms = defaultdict(list)
+        i = 0
+        start = time.time()
         for obj_perm in object_permissions:
             for object_type in obj_perm.object_types.all():
                 for action in obj_perm.actions:
                     perm_name = f"{object_type.app_label}.{action}_{object_type.model}"
                     perms[perm_name].extend(obj_perm.list_constraints())
+                    syslog.syslog(perm_name)
+                    i += 1
+        end = time.time()
+        t = end - start
+        syslog.syslog("took " + str(t) + " seconds with " + str(i) + " iterations")

         return perms

```

The output:

```
Sep 29 22:55:18 netbox.example.com gunicorn[78629]: took 1.325146198272705 seconds with 10397 iterations
```

Wow. That's a lot. Okay, more dirty patching:

```
@@ -24,7 +27,7 @@ class ObjectPermissionMixin():
         return user_obj._object_perm_cache

     def get_permission_filter(self, user_obj):
-        return Q(users=user_obj) | Q(groups__user=user_obj)
+        return Q(users=user_obj)

     def get_object_permissions(self, user_obj):
         """
```

And  now I have this:

```
Sep 29 22:57:51 netbox.example.com gunicorn[78889]: took 0.17675495147705078 seconds with 1126 iterations
```

Still a lot (the request itself took ~400ms), but a lot better.

I'm not familiar with Netbox internals at all, so I'm not sure where to start digging.

kkthxbye-code:
### Fixes: #7399

This should fix the excessive CPU usage when enabling AUTH_LDAP_FIND_GROUP_PERMS and having a combination of many object permissions and many users in the same group.

First, this is not an optimal fix, feel free to close the PR if a better fix is suggested. The reason for the bad performance is the filter added in 82300990ec79f8ce6b2fca953d5548028a14aaea:

https://github.com/netbox-community/netbox/blob/82300990ec79f8ce6b2fca953d5548028a14aaea/netbox/netbox/authentication.py#L174-L180

When added to the original query:

https://github.com/netbox-community/netbox/blob/82300990ec79f8ce6b2fca953d5548028a14aaea/netbox/netbox/authentication.py#L23-L34

The resulting SQL is essentially:

```sql
SELECT "users_objectpermission".*
FROM "users_objectpermission"
LEFT OUTER JOIN "users_objectpermission_users" ON ("users_objectpermission"."id" = "users_objectpermission_users"."objectpermission_id")
LEFT OUTER JOIN "users_objectpermission_groups" ON ("users_objectpermission"."id" = "users_objectpermission_groups"."objectpermission_id")
LEFT OUTER JOIN "auth_group" ON ("users_objectpermission_groups"."group_id" = "auth_group"."id")
LEFT OUTER JOIN "auth_user_groups" ON ("auth_group"."id" = "auth_user_groups"."group_id")
WHERE (("users_objectpermission_users"."user_id" = 1
        OR "auth_user_groups"."user_id" = 1
        OR "auth_group"."name" IN (
                                   'netbox-user',
                                   'group1',
                                   'group2'
                                   ))
       AND "users_objectpermission"."enabled")
ORDER BY "users_objectpermission"."name" ASC;
```

The outer join for auth_user_group essentially duplicates each users_objectpermission row once for each other user that is allocated to one of the groups not filtered out in the auth_group join.

In my test instance, with 327 objectpermissions in the netbox-user group, 30 other users has that same group, which makes the query return 30 duplicates per objectpermission totalling 9810 permissionobjects. The fix caused a load of the rack page with 50 objects per page to decrease from 1400 ms to 600 ms.

I'm not sure how to fix it in a smart way, as I have a hard time seing how the relationship is supposed to work. Even the original query is iffy I think, as I'm pretty sure it can return duplicate permissions as well if user bound objectpermissions are used.

The fix just makes the postgres remove the duplicates, which fixes the performance regression when using AUTH_LDAP_FIND_GROUP_PERMS. The query time increase should be negligible even in extreme cases.

Still someone should probably rework the queries.

Note: While the fix is pretty straightforward and I can't see how it could mess anything up, I would appreciate if someone else could test it before deciding to merge.

### DIFF
--- a/netbox/netbox/authentication.py
+++ b/netbox/netbox/authentication.py
@@ -34,7 +34,7 @@ class ObjectPermissionMixin():
         object_permissions = ObjectPermission.objects.filter(
             self.get_permission_filter(user_obj),
             enabled=True
-        ).prefetch_related('object_types')
+        ).order_by('id').distinct('id').prefetch_related('object_types')
 
         # Create a dictionary mapping permissions to their constraints
         perms = defaultdict(list)


buzzingbren:
Hi @kkthxbye-code and @jeremystretch, thanks for linking to this.
Unfortunately In my case (using https://github.com/bb-Ricardo/netbox-sync to fill our Netbox) this fix does not resolve my issue. But I'm also not using AUTH_LDAP_FIND_GROUP_PERMS setting. But since I only have issues when the Nebox-Sync is using the API to access Netbox, a feature in Netbox to set a user as "local account only" or "LDAP-User" and forgoing the LDAP check if the API-token is linked to a local user would help in my case. Bypassing the LDAP problem altogether (for me then, but probably there are other use cases as well).
Just for authentication purposes for the web-interface, I had no problems there with the LDAP authentication.

tobiasge:
My first thought was "Why `oder_by` and `disctinct`? But after reading about the additional capabilities of Postgres in `disctinct` I think this is a really good solution. Nice 👍

kkthxbye-code:
@buzzingbren - Yeah, the API performance issue when using LDAP is not related to this fix. The closest issue might be #6926 - but a new issue with your specific case might be in order. I haven't noticed any performance issue with the API, but I'm not really hammering it. If you create an issue or add a comment to #6926 with more details about what your setup is and what you are doing when experiencing slowdown, I'll be happy to try to debug it.

This issue fixed in this PR will manifest if the following is true:

* AUTH_LDAP_FIND_GROUP_PERMS enabled
* The user is an LDAP user
* More than one user is in the same group as you
* The group has a fair amount of ObjectPermissions

If this is true, the number of ObjectPermissions returned by the query will be equal to ObjectPermissions * Other users in the same group as you - so if you have 200 object permissions and 100 other netbox users are in the same group as you, the query will return 20000 ObjectPermissions, causing the CPU usage when it's iterating them here:

https://github.com/netbox-community/netbox/blob/82300990ec79f8ce6b2fca953d5548028a14aaea/netbox/netbox/authentication.py#L36-L42

jeremystretch:
Sorry this fell off the radar for a bit. Seems like there's been a bit of overlap between #7399 and one or more other LDAP issues, bit I'm happy to merge this with the understanding that it addresses the specific issue raised in #7399.

PR #7677, opened 2021-10-28
#6529 - Add CLI to run scripts

### LINKED ISSUE #6529: Run script in command line (without redis)
### Proposed Changes

Add command to run script in command line without redis.

### Justification

Useful for testing, developing and debugging scripts

DanSheps:
### Closes: #6529

Add CLI to extras to allow running scripts

### DIFF
--- a/docs/customization/custom-scripts.md
+++ b/docs/customization/custom-scripts.md
@@ -264,10 +264,12 @@ http://netbox/api/extras/scripts/example.MyReport/ \
 Scripts can be run on the CLI by invoking the management command:
 
 ```
-python3 manage.py runscript [--commit] [--loglevel {debug,info,warning,error,critical}] --script <module>.<script> <data>
+python3 manage.py runscript [--commit] [--loglevel {debug,info,warning,error,critical}] [--data "<data>"] <module>.<script> 
 ```
 
-The required ``--script <module>.<script>`` argument is the report to run where ``<module>`` is the name of the python file in the ``scripts`` directory without the ``.py`` extension and ``<script>`` is the name of the script class in the ``<module>`` to run.
+The required ``<module>.<script>`` argument is the script to run where ``<module>`` is the name of the python file in the ``scripts`` directory without the ``.py`` extension and ``<script>`` is the name of the script class in the ``<module>`` to run.
+
+The optional ``--data "<data>"`` argument is the data to send to the script
 
 The optional ``--loglevel`` argument is the desired logging level to output to the console.
 
--- a/docs/customization/custom-scripts.md
+++ b/docs/customization/custom-scripts.md
@@ -259,6 +259,20 @@ http://netbox/api/extras/scripts/example.MyReport/ \
 --data '{"data": {"foo": "somevalue", "bar": 123}, "commit": true}'
 ```
 
+### Via the CLI
+
+Scripts can be run on the CLI by invoking the management command:
+
+```
+python3 manage.py runscript [--commit] [--loglevel {debug,info,warning,error,critical}] --script <module>.<script> <data>
+```
+
+The required ``--script <module>.<script>`` argument is the report to run where ``<module>`` is the name of the python file in the ``scripts`` directory without the ``.py`` extension and ``<script>`` is the name of the script class in the ``<module>`` to run.
+
+The optional ``--loglevel`` argument is the desired logging level to output to the console.
+
+The optional ``--commit`` argument will commit any changes in the script to the database.
+
 ## Example
 
 Below is an example script that creates new objects for a planned site. The user is prompted for three variables:
--- a/netbox/extras/management/commands/runscript.py
+++ b/netbox/extras/management/commands/runscript.py
@@ -6,14 +6,14 @@ import uuid
 
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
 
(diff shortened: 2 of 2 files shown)


netbox/extras/management/commands/runscript.py
```diff
@@ -0,0 +1,160 @@
...
+            logger.setLevel(logging.WARNING)
+        elif loglevel == 'error':
+            logger.setLevel(logging.ERROR)
+        else:
+            logger.setLevel(logging.INFO)
+
+        # Get the script
+        script = self._get_script(module, name)
+        # Parse the parameters
+        form = script.as_form(data, None)
```
jeremystretch, line 111: Why use a form here? It should be possible to just pass the data directly to the script, as we do with the REST API endpoint for scripts. IMO we shouldn't use forms outside of UI views.

netbox/extras/management/commands/runscript.py
```diff
@@ -0,0 +1,160 @@
...
+            'user': user,
+            'path': '',
+            'id': job_result.job_id
+        })
+
+        if form.is_valid():
+            job_result.status = JobResultStatusChoices.STATUS_RUNNING
+            job_result.save()
+
+            commit = form.cleaned_data.pop('_commit')
```
jeremystretch, line 137: Seems like it would be more convenient to expose `commit` as a command argument, rather than extract it from the input data.

netbox/extras/management/commands/runscript.py
```diff
@@ -0,0 +1,160 @@
...
+            # Execute the script. If commit is True, wrap it with the change_logging context manager to ensure we process
+            # change logging, webhooks, etc.
+            if commit:
+                with change_logging(request):
+                    _run_script()
+            else:
+                _run_script()
+
+            # Delete any previous terminal state results
+            JobResult.objects.filter(
```
jeremystretch, line 151: We should probably do this before creating a new JobResult for the current run.

netbox/extras/management/commands/runscript.py
```diff
@@ -0,0 +1,160 @@
...
+class Command(BaseCommand):
+    help = "Run a script in Netbox"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--loglevel',
+            help="Logging Level (default: info)",
+            dest='loglevel',
+            default='info',
+            choices=['debug', 'info', 'warning', 'error'])
```
jeremystretch, line 30:
```suggestion
            choices=['debug', 'info', 'warning', 'error', 'critical'])
```

netbox/extras/management/commands/runscript.py
```diff
@@ -0,0 +1,160 @@
...
+            job_result.save()
+
+            commit = form.cleaned_data.pop('_commit')
+
+            logger.info(f"Running script (commit={commit})")
+            script.request = request
+
+            # Execute the script. If commit is True, wrap it with the change_logging context manager to ensure we process
+            # change logging, webhooks, etc.
+            if commit:
```
jeremystretch, line 144: It might be prudent to _always_ run inside `change_logging()`, to catch any potential bug that are present only with the context active.

jeremystretch (changes requested):

netbox/extras/management/commands/runscript.py
DanSheps: We do exclude the current JobResult, I took a copy of this from the current script, and the reason it is done this way is I believe there may be a failure mode running under RQ which does not change the job state, however that shouldn't be an issue here so I can make this change I think.

netbox/extras/management/commands/runscript.py
DanSheps: The reason is a form seems to resolve ModelObjects to their object, passing the raw data does not.

So if a script is expecting a DeviceObject because someone used a ModelVar and we passed the straight PK it wouldn't work.  Perhaps there is a easier way to do this.

netbox/extras/management/commands/runscript.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--loglevel',
+            help="Logging Level (default: info)",
+            dest='loglevel',
+            default='info',
+            choices=['debug', 'info', 'warning', 'error', 'critical'])
+        parser.add_argument('--script', help="Script to run", dest='script', required=True)
+        parser.add_argument('--commit', help="Commit this script to database", dest='commit')
```
hSaria, line 32: You could convert the `--commit` argument to a flag by using `action='store_true'` then use `if options['commit']:`; it'll default to `False` when not specified in `sys.argv`.

You can also skip the `dest` keyword since it default to the first arg of `add_argument`.

docs/customization/custom-scripts.md
```diff
@@ -259,6 +259,20 @@ http://netbox/api/extras/scripts/example.MyReport/ \
...
 
+### Via the CLI
+
+Scripts can be run on the CLI by invoking the management command:
+
+```
+python3 manage.py runscript [--commit] [--loglevel {debug,info,warning,error,critical}] --script <module>.<script> <data>
+```
+
+The required ``--script <module>.<script>`` argument is the report to run where ``<module>`` is the name of the python file in the ``scripts`` directory without the ``.py`` extension and ``<script>`` is the name of the script class in the ``<module>`` to run.
```
jeremystretch, line 270: `--foo` format arguments are typically optional. It would probably make more sense to make the script a positional argument (similar to `runreport`).

DanSheps: Swapped them around, should be good now.

PR #7697, opened 2021-10-30
Update install docs with Python3.9 instruction

### LINKED ISSUE #7687: CentOs installation instructions
### Change Type

Correction

### Area

Installation instructions

### Proposed Changes

In the CentOs instructions we are requestion python36 to be installed. 

I guess this needs to be updated as 3x is not supported anymore

image

apellini:

### DIFF
--- a/docs/installation/3-netbox.md
+++ b/docs/installation/3-netbox.md
@@ -18,8 +18,28 @@ Begin by installing all system packages required by NetBox and its dependencies.
 === "CentOS"
 
     ```no-highlight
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thank you for your effort, however the installation of Python on the underlying operating system is firmly outside the scope of NetBox's documentation. We can't be expected to maintain documentation on behalf of the distribution. If the OS has not made the necessary Python release available via its native package management system, the user will need to seek instruction from the distribution's own documentation.

PR #7785, opened 2021-11-09
Fixes #7783 - Site location visual changes

### LINKED ISSUE #7783: Indent child locations in Site view
### NetBox version

3.0.9

### Feature type

### Proposed functionality

When viewing a site the locations widget lists all parent and child locations with the same level of indentation (none). This can make it difficult to understand the relationship between parent/child locations. 

A few possible proposals:
* Intend child locations. The list is already ordered based on parent/child relationship so intending would improve usability.
* Change the icon of a child location to make it easily distinguishable from a parent.
* Make the location widget expand/collapse. Initially show only parents and allow the row to be clicked to expand to display children

### Use case

We operate a campus with multiple buildings, each building has multiple floors with 1-2 IDF's per floor. In Netbox, each building is defined as location with a child location per IDF. Visually the list when looking at the Site view is difficult to follow as all locations appear with the same hierarchy / indentation level.

### Database changes

_No response_

### External dependencies

_No response_

jasonyates:
### Fixes: #7783 
Updating site location list to visually match the /dcim/locations list where child locations are "indented" with mdi-circle-small.

Also removes the padding-left attribute on each row as it is no longer functional.

### DIFF
--- a/netbox/templates/dcim/site.html
+++ b/netbox/templates/dcim/site.html
@@ -219,8 +219,8 @@
                     </tr>
                     {% for location in locations %}
                         <tr>
-                            <td style="padding-left: {{ location.level }}8px">
-                                <i class="mdi mdi-folder-open"></i>
+                            <td>
+                                {% for i in location.level|as_range %}<i class="mdi mdi-circle-small"></i>{% endfor %}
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks!

PR #7826, opened 2021-11-13
Add filter for optionally including assigned prefixes

### LINKED ISSUE #7665: Add filter to container prefixes to show only avaliable child Prefixes
### NetBox version

v3.0.8

### Feature type

### Proposed functionality

When looking at the child prefixes of a container prefix there are two views
1. "Show Available" Prefixes
2. "Hide Available" Prefixes

image

This FR proposes a third option that filters only by available prefixes in the container. "Show Only Available"

This filter removes all currently allocated prefixes from the list, and only displays currently unassigned "available" prefixes within the container prefix

I should be able to put a PR in for this if accepted.

### Use case

We currently use the "Show Available Prefixes" filter to assign new child prefixes from the container. 
This works well when the container is mostly empty, but when you have a /20 container full of /31's that's 99% full you need to run through pages and pages of allocated prefixes to find the ones that are unallocated.

### Database changes

None - this functionality is currently available via the API

### External dependencies

None

WillIrvine:
### Fixes: #7665

- Added new filtering boolean "show_assigned" for optionally including assigned prefixes in filtered table, note: this defaults to true so the previous functionality is maintained. you just now have the option of using false and filtering these out
- Adjusted "add_available_prefixes" function to only return available prefixes, rather than adding to prefix_list, addition is now done in the view
- Moved sorting logic to the view so that we only sort once after adding all requested prefixes to the list.

Adding a second boolean seems like a much better option than trying to add a third option to the current boolean (as was proposed in the original issue).


DanSheps (changes requested):
Rather then using a True/False for Show Available and Show Assigned, have instead 3 options:

* Show Available
* Show Assigned
* Show All/Both (default and fallback for invalid values)

The reason being is, I can see a potential issue here with what if someone has both "hide assigned" and "hide available".  You get nothing back, and I don't think that should be an option in the UI.

Additionally, I would leave the sorting and appending logic in the helper function and instead modify that to take an additional parameter to append the assigned or not.  It will be cleaner and less repetition in the end.

WillIrvine:
Hi @DanSheps ,

Thanks for the review, happy to do that way. However id like to point out the reasons I took this approach

- Adding as a second boolean doesn't break existing logic as the "show_available" option retains its original function, we would need to rename this and adjust its options away from true / false should we give it a third option, meaning any links / favorites, etc will need to be updated to give the same results.

- Adding as a second boolean makes the process of adding a potential third filter much easier as you only need to take into account the unique options each individual filter performs. if you try to account for every scenario in a single filter this exponentially increases the options required as they are added.

- I can see the issue with having no prefixes displayed, however this is pretty standard for filter logic (filtering yourself out of any results) Both queries default to true so it would be an active decision by the user to filter both out.

WillIrvine:
Hi @DanSheps ,

Please see latest revisions.

I've moved the appending and sorting logic back to utils
I've also kept the two filters separate in the View, but adjusted the buttons to correspond with both filters.

i.e

Show Assigned = show_assigned=true&show_available=false
Show Available = show_assigned=false&show_available=true
Show All = show_assigned=true&show_available=true

image

Let me know if this works. This wont break current filters for people which is the ideal outcome.
Icons are up for debate though, ha.

DanSheps:
I think I like the second set of icons better, but let me check MDI to see if there is another option instead.

DanSheps:
Thinking about it, I think the first set of icons is fine for now.  I will review tonight and let you know.

WillIrvine:
Sounds good, one more for thought..

image

DanSheps (approved):

WillIrvine:
Fixed conflicts with new develop 3.1 branch

netbox/ipam/utils.py
```diff
@@ -4,20 +4,30 @@
 from .models import Prefix, VLAN
 
 
-def add_available_prefixes(parent, prefix_list):
+def add_requested_prefixes(parent, prefix_list, request):
```
jeremystretch, line 7: My only nit here is that we're passing the whole request object. This should instead be two optional kwargs (e.g. `show_available` and `show_assigned`) which default to True. This will ensure that the function can be used outside the context of a specific request if needed.

jeremystretch: I'm happy to tackle this myself since this has already been approved.

PR #7827, opened 2021-11-14
Fix #7424: Add virtual_chassis_id filter for device components

### LINKED ISSUE #7424: Filter interfaces/ports on virtualchassis_id
### NetBox version

V2.11.3

### Feature type

New functionality

### Proposed functionality

For the interfaces API endpoint, it would be nice if one could filter on virtualchassis_id, beyond device_id. I imagine it would be fairly simply include, but it would need to be mutually exclusive with filtering on device_id.

### Use case

Efficiency in loading virtualchassis interfaces via one api request vs per-device requests

### Database changes

None

### External dependencies

None

rhyser9:
### Fixes: #7424 

- Added `virtual_chassis` and `virtual_chassis_id` filters to device components
- Added virtual chassis field to device component filter forms
- Filter device field based on value of virtual_chassis field
- Added test to interface filterset test case for virtual_chassis_id filter

image
image

### DIFF
--- a/netbox/dcim/api/nested_serializers.py
+++ b/netbox/dcim/api/nested_serializers.py
@@ -340,7 +340,7 @@ class NestedVirtualChassisSerializer(WritableNestedSerializer):
 
     class Meta:
         model = models.VirtualChassis
-        fields = ['id', 'name', 'url', 'master', 'member_count']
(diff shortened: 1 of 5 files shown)


jeremystretch:
Great work, thanks @rhyser9!

PR #7829, opened 2021-11-14
Fix #7542: Add VLAN Group column to IP Prefix table

### LINKED ISSUE #7542: VLAN Group information in the Prefixes overview would help
### NetBox version

v3.0.7

### Feature type

### Proposed functionality

Extend the VLAN information on the Prefixes overview by also showing the VLAN Group information.

### Use case

VLANs can exist multiple times in an IPAM, and the distinguisher is the VLAN Group.
While in an individual Prefixes view the VLAN shows both the VLAN Group and the VLAN information, separated by a slash, in the Prefixes overview the VLAN column only shows the VLAN.
A similar display of VLAN Group/VLAN like in the individual Prefix would greatly improve the overview.

### Database changes

none

### External dependencies

_No response_

rhyser9:
### Fixes: #7542

Adds a non-default VLAN Group column to the IP Prefix table in order to disambiguate which VLAN instance the prefix is assigned to, particularly when the table is exported.

image

### DIFF
--- a/netbox/ipam/tables/ip.py
+++ b/netbox/ipam/tables/ip.py
@@ -206,6 +206,11 @@ class PrefixTable(BaseTable):
     site = tables.Column(
         linkify=True
     )
+    vlan_group = tables.Column(
+        accessor='vlan__group',
+        linkify=True,
+        verbose_name='VLAN Group'
+    )
     vlan = tables.Column(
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @rhyser9!

PR #7915, opened 2021-11-23
Add support for L22-30P power port type

### LINKED ISSUE #7892: L22-30P power port type
### NetBox version

v3.0.7

### Feature type

Data model extension

### Proposed functionality

Under power port type, L22-30P is missing. Adding it as an option would be great.

### Use case

Allow for proper documentation as it is a common locking plug type

### Database changes

_No response_

### External dependencies

_No response_

jsenecal:
### Fixes #7892

Adds support for L22-30P power port type to the available choices


DanSheps (approved):

DanSheps (changes requested):
Where is the corresponding receptacle/outlet for L22-30P?

jsenecal:
> Where is the corresponding receptacle/outlet for L22-30P?

Excellent point, I had only focused on one side of the equation - let me fix that

PR #7928, opened 2021-11-25
Fix #7751: LDAP: Only get API user from ldap when FIND_GROUP_PERMS is on

### LINKED ISSUE #7751: Don't fetch LDAP user and groups on all API request when FIND_GROUP_PERMS is disabled
### NetBox version

V3.0.9

### Feature type

### Proposed functionality

Currently when using the LDAP backend for authentication, the AD is queried on every API request, regardless of other settings and regardless if the user is local or has been created by the LDAP backend. Additionally the LDAP cache built into django-auth-ldap does not function when using populate_user.

As the user is not actually authenticated against the AD when using the API (the token is used), I propose that the local user and it's group assignments are used when FIND_GROUP_PERMISSIONS is disabled.

I have a change ready for pull request if the issue is accepted.

For more info, please see the discussion I created: https://github.com/netbox-community/netbox/discussions/7708

This issue would also partly fix #6926 - it will not fix the caching, but the user who reported the issue is not using FIND_GROUP_PERMISSIONS.

### Use case

The end goal is vastly improved API performance when using the LDAP backend in most cases.

The above changes will result in the following changes for users:

**Not using the LDAP backend:**

No changes

**FIND_GROUP_PERMS  = True:**

No changes

**MIRROR_GROUPS = True and FIND_GROUP_PERMS = True:**

No changes

**MIRROR_GROUPS = True and FIND_GROUP_PERMS = False:**

Local user and group assignments will be used when calling the API and the user and groups are never reloaded from the LDAP server during API calls. This means that LDAP users utilizing the API will have to login to the web ui to update group memberships. The change also allows one to use locally created users to call the API with querying the LDAP server.

**MIRROR_GROUPS = False and FIND_GROUP_PERMS = False:**

The user performing the API request has to be locally assigned groups or have local user object permissions.

### Database changes

No database changes

### External dependencies

_No response_

kkthxbye-code:
### Fixes: #7751

When using the LDAP backend for auth, all API requests would refresh the user (local or LDAP created) from the LDAP server, causing performance issues. With this fix, the user is only refreshed when FIND_GROUP_PERMS is enabled.

I tested all combinations I could think of. Not sure if this change in logic warrants a warning in the release notes. The changes for different combinations of settings can be seen in the issue: #7751

### DIFF
--- a/netbox/netbox/api/authentication.py
+++ b/netbox/netbox/api/authentication.py
@@ -29,10 +29,13 @@ class TokenAuthentication(authentication.TokenAuthentication):
         if settings.REMOTE_AUTH_BACKEND == 'netbox.authentication.LDAPBackend':
             from netbox.authentication import LDAPBackend
             ldap_backend = LDAPBackend()
-            user = ldap_backend.populate_user(token.user.username)
-            # If the user is found in the LDAP directory use it, if not fallback to the local user
-            if user:
-                return user, token
+
+            # Load from LDAP if FIND_GROUP_PERMS is active
+            if ldap_backend.settings.FIND_GROUP_PERMS:
+                user = ldap_backend.populate_user(token.user.username)
+                # If the user is found in the LDAP directory use it, if not fallback to the local user
+                if user:
+                    return user, token
 
         return token.user, token
 


kkthxbye-code:
Should also fix #6926

PR #7933, opened 2021-11-25
Fix #7932: Improve performance of the quick filter

### LINKED ISSUE #7932: Quick find search box is slow
### NetBox version

v3.0.10

### Python version

3.9

### Steps to Reproduce

1. Load the data from [netbox-demo-data](https://github.com/netbox-community/netbox-demo-data
2. Go to Devices -> Device Components -> Interfaces
3. Change pagination to 1000
4. Paste "GigabitEthernet" into the Quick Find field

### Expected Behavior

Results in a timely manner without freezing the browser for an extended period of time.

### Observed Behavior

On my i7-9700 it takes roughly 22 seconds before the browser tab is responsive again.

kkthxbye-code:
### Fixes: #7932

Instead of querying the DOM and modifying it in the same loop, maintain a list of matched rows and apply the css classes afterwards.

Before the fix filtering 1000 elements took 22 seconds. After the fix the same query took around 200 ms.


kkthxbye-code:
This is my first UI change, so please tell me if I did anything wrong.

PR #7941, opened 2021-11-26
Add multistandard ITA power outlet type

### LINKED ISSUE #7940: Add multistandard ITA Power Outlet type
### NetBox version

v3.0.11

### Feature type

### Proposed functionality

NetBox has Power Outlet types for all major ITA connector types, but there also exists a "multistandard" outlet type. These are very common in this country - I would say perhaps half to three quarters of all modern outlets are multistandard. We also use exclusively ITA multistandard PDUs in those rare cases where we need ITA outlets (otherwise using 60320 or 60309).

The outlets look similar to this, with some variations on accepted ITA types:
image

Usually they accept A, B, E, F, G, I and M types at least.

### Use case

Model the ITA multistandard Power Outlet.

### Database changes

None.

### External dependencies

None.

bluikko:
### Fixes: #7940

Add the ITA multistandard Power Outlet type.

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -552,6 +552,7 @@ class PowerOutletTypeChoices(ChoiceSet):
     TYPE_ITA_M = 'ita-m'
     TYPE_ITA_N = 'ita-n'
     TYPE_ITA_O = 'ita-o'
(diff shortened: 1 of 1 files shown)


DanSheps (approved):
LGTM

PR #7969, opened 2021-12-02
Fix #6905: Assign existing prefix to VLAN from VLAN view

### LINKED ISSUE #6905: Assign existing prefix to VLAN, starting from VLAN view
### NetBox version

v2.11.10

### Feature type

### Proposed functionality

When attaching a prefix to a VLAN there are two routes in the GUI:
1) Open the prefix and choose a VLAN using the vlan filters OR
2) Open the VLAN and click 'add prefix'

Scenario 2 closely repeats the workflow for assigning an IP to an interface. In that selector there is an option to assign an already known IP instead of creating a new instance. 
Assigning a prefix to a vlan would benefit from an 'assign' option as well. 

### Use case

When starting from a VLAN, trying to attach an existing prefix requires opening a new set of tabs to look up the prefix and starting the association from there. Attaching an already known prefix would save quite some clicks.

### Database changes

N/A

### External dependencies

N/A

rhyser9:
### Fixes: #6905

Adds a prefix assign form, mimicking the IP address assign form from the Interface view. This allows the end user to view a VLAN and assign an existing prefix, instead of only creating a new prefix.

Adds PrefixAssign form, table, view, url, and template.
Adds prefix_edit_header.html, used by prefix_assign.html and prefix_edit.html

### DIFF
--- a/netbox/ipam/forms/models.py
+++ b/netbox/ipam/forms/models.py
@@ -7,6 +7,7 @@ from extras.models import Tag
 from ipam.constants import *
 from ipam.models import *
 from tenancy.forms import TenancyForm
+from tenancy.models import Tenant
 from utilities.forms import (
     BootstrapMixin, ContentTypeChoiceField, DatePicker, DynamicModelChoiceField, DynamicModelMultipleChoiceField,
     NumericArrayField, SlugField, StaticSelect, StaticSelectMultiple,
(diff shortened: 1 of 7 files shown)


jeremystretch:
Thanks for your work on this. However, I'm pretty opposed to adding any more custom views that aren't absolutely necessary. I've been trying hard to replace such views with our generic views to bolster maintainability.

PR #7978, opened 2021-12-04
Fixes #7003: Order image attachments by newest uploaded

### LINKED ISSUE #7003: Order Device images by date
### NetBox version

v2.11.0

### Feature type

### Proposed functionality

tldr - Images are currently shown in the web UI sorted by Name, can this be changed to Created date (newest at the top?)

Uploading images to a Device is a great way to see what it actually looks like in the real live rack -However, over time things change slightly and so  we take new photos and add them - we like to keep old photos as a reference - but usually just want to look at the newest images - so ordering by Created date would be great here.

### Use case

It makes it easier to see the newest images of a Device.

### Database changes

_No response_

### External dependencies

_No response_

rhyser9:
### Fixes: #7003 

Sorts images by date added, so that the most recent images are displayed first.
This might be better implemented by changing the models Meta ordering from `ordering = ('name', 'pk')`  to `ordering = ('created', 'name', 'pk')`, but that triggers a database migration.

### DIFF
--- a/netbox/templates/inc/image_attachments_panel.html
+++ b/netbox/templates/inc/image_attachments_panel.html
@@ -14,7 +14,7 @@
             <th>Created</th>
             <th></th>
           </tr>
-          {% for attachment in images %}
+          {% for attachment in images|dictsortreversed:"created" %}
             <tr{% if not attachment.size %} class="table-danger"{% endif %}>
               <td>
                 <i class="mdi mdi-file-image-outline"></i>


jeremystretch:
It doesn't seem prudent to force ordering by date, but to allow it as an option. I assume many users prefer the current alphabetical ordering.

Implementing this FR will probably involve rendering the attachments as a table to allow column-based ordering. However, some additional consideration is needed where another paginated table is already present within the view.

jeremystretch:
I'm going to close this out as the static approach proposed is untenable. However, we're slowly working on migrating embedded tables to utilize HTMX application-wide. Once we do this for the attachments table, you'll be able to sort by any of the columns without a page reload.

Issue #7401, opened 2021-09-30
NetBox API swagger docs fetch url error


cybarox:
### NetBox version

v3.0.4

### Python version

3.8

### Steps to Reproduce

1. open https://demo.netbox.dev/api/docs/
2. View Fetch Error

### Expected Behavior

API docs opens

### Observed Behavior

API docs show Fetch URL error

tirkarthi:
https://github.com/Julian/jsonschema/issues/845#issuecomment-931282776

4.0.0 has now been yanked and 4.0.1 was released with a fix for `python_requires`

jeremystretch:
Well that was fast! Thanks for the heads up, @tirkarthi! Appreciate the quick fix!

cybarox:
with jsonschema=4.0.1 still the same issue, temp fix works for me

jeremystretch:
Agreed, still seems to be an issue.

tirkarthi:
Is there a traceback showing the error in jsonschema?

jeremystretch:
@tirkarthi here's what I've narrowed it down to:

```
$ python3.7 -m venv testing

$ source testing/bin/activate
(testing) $ pip install swagger_spec_validator
Collecting swagger_spec_validator
  Using cached swagger_spec_validator-2.7.3-py2.py3-none-any.whl (27 kB)
Collecting six
  Using cached six-1.16.0-py2.py3-none-any.whl (11 kB)
Collecting pyyaml
  Using cached PyYAML-5.4.1-cp37-cp37m-manylinux1_x86_64.whl (636 kB)
Collecting jsonschema
  Using cached jsonschema-4.0.1-py3-none-any.whl (69 kB)
Collecting attrs>=17.4.0
  Using cached attrs-21.2.0-py2.py3-none-any.whl (53 kB)
Collecting pyrsistent!=0.17.0,!=0.17.1,!=0.17.2,>=0.14.0
  Using cached pyrsistent-0.18.0-cp37-cp37m-manylinux1_x86_64.whl (119 kB)
Collecting importlib-metadata; python_version < "3.8"
  Using cached importlib_metadata-4.8.1-py3-none-any.whl (17 kB)
Collecting typing-extensions>=3.6.4; python_version < "3.8"
  Using cached typing_extensions-3.10.0.2-py3-none-any.whl (26 kB)
Collecting zipp>=0.5
  Using cached zipp-3.6.0-py3-none-any.whl (5.3 kB)
[...]

PR #8063, opened 2021-12-14
Get_Environment from napalm should not need any decoding

### LINKED ISSUE #7246: Get Environment Failure
### NetBox version

v3.0.2

### Python version

3.8

### Steps to Reproduce

The issue occurs for all my devices (Cisco ISR4321 and WS-C3850) using NApalm Diver "ios".
I can't reproduce this on the demo instance, as it needs live devices with a primary IP.

1.Go to a device detail page
2.Click on "Status"

### Expected Behavior

Both "Device Facts" and "Environment" to be populated.

### Observed Behavior

Only the "Device Facts" gets populated, and no information seems to be pulled for "Environment".
Getting error:
image

Need to mention that "LLDP Neighbors" and "Configuration" works fine.

rizlas:
### Fixes: #7246

As stated already in the issue, napalm method `get_environment` is currently failing. Dictionary coming from napalm driver are already sanitized and therefore param decode_keys should be set to false before calling decode_dict.

https://github.com/napalm-automation/napalm/blob/develop/napalm/base/base.py#L465
https://github.com/napalm-automation/napalm/blob/develop/napalm/ios/ios.py#L2197

This was tested with ios driver.

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -501,7 +501,7 @@ class DeviceViewSet(ConfigContextQuerySetMixin, CustomFieldModelViewSet):
                 response[method] = {'error': 'Only get_* NAPALM methods are supported'}
                 continue
             try:
-                response[method] = decode_dict(getattr(d, method)(), decode_keys=False)
+                response[method] = getattr(d, method)()
             except NotImplementedError:
                 response[method] = {'error': 'Method {} not implemented for NAPALM driver {}'.format(method, driver)}
             except Exception as e:
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -501,7 +501,7 @@ class DeviceViewSet(ConfigContextQuerySetMixin, CustomFieldModelViewSet):
                 response[method] = {'error': 'Only get_* NAPALM methods are supported'}
                 continue
             try:
-                response[method] = decode_dict(getattr(d, method)())
+                response[method] = decode_dict(getattr(d, method)(), decode_keys=False)
             except NotImplementedError:
                 response[method] = {'error': 'Method {} not implemented for NAPALM driver {}'.format(method, driver)}
             except Exception as e:


jeremystretch:
`decode_dict()` was introduced as part of the fix for #7401. I'm not comfortable altering its execution without understanding the ramifications of doing so. @thatmattlove can you comment?

thatmattlove:
I'm fairly certain that while this may fix data from IOS devices, it will break data from Aruba CX devices, thus regressing #7041.

I can see a few options forward:

1. Add some logic to `decode_dict()` determine if keys need to be decoded in the first place (i.e., checking for escape characters)
2. Add some logic to `napalm()` to adjust the `decode_keys` keyword argument as needed, which would require us to maintain a mapping of device types that need their responses decoded (to my knowledge, it's only the Aruba CX platform).
3. Add some logic to `napalm()` to only add the `decode_keys` keyword argument to `get_interfaces()`
4. Submit a PR to [napalm-automation-community/napalm-aruba-cx](https://github.com/napalm-automation-community/napalm-aruba-cx) that implements our `decode_dict` function for `get_interfaces`, however I'm hesitant to do this since I'm not overly familiar with napalm, nor Aruba CX.

4 is probably the best, 3 is probably the easiest :)

jeremystretch:
IMO this looks like a bug in the Arbua NAPALM driver. The below output was cited under #7041:

```
"Port": {
    "1%2F4%2F1": {
        "interfaces": [
            "1%2F4%2F1"
        ],
        "name": "1/4/1"
    },
    "1%2F4%2F2": {
        "interfaces": [
            "1%2F4%2F2"
        ],
        "name": "1/4/2"
    },
```

The interface names are URL-escaped for some reason, which I would not expect. I noticed that the slash in some IP addresses gets escaped as well.

We really shouldn't need to do any processing of the NAPALM response data; `decode_dict()` should not be needed. Unless I'm missing something?

thatmattlove:
I don't disagree. #7041 was originally opened because of `[object Object]` being displayed as the config, and this should have been fixed by the changes made [here](https://github.com/netbox-community/netbox/commit/14d87a3584c53b20ce0be38aeea92b2307f11355#diff-8fd7285bcf022de5a39a1d7729a17480eed64dea250a3cd40956d0311eb034dcR29). During testing I went a step further and fixed the key encoding issue. I recall not being able to determine if the issue was with Aruba CX's response, NAPALM core, or the NAPALM driver.

If we don't care about "breaking" Aruba CX NAPALM interface checks, then I have no issue removing `decode_dict()` from the codebase.

netbox/dcim/api/views.py
```diff
@@ -501,7 +501,7 @@ def napalm(self, request, pk):
                 response[method] = {'error': 'Only get_* NAPALM methods are supported'}
                 continue
             try:
-                response[method] = decode_dict(getattr(d, method)())
+                response[method] = decode_dict(getattr(d, method)(), decode_keys=False)
```
jeremystretch, line 504: Can you try removing the call to `decode_dict()` entirely and test again? Per @thatmattlove's comments I think we can just remove it. Any drivers which don't properly format their output will need to be fixed upstream.

```suggestion
                response[method] = getattr(d, method)()
```

jeremystretch (changes requested):

netbox/dcim/api/views.py
rizlas: Yes, it is working without `decode_dict`. I also agree with the fact that format output should be handled directly by napalm.

schwos:
To report the same error experienced here was impacting EOS Napalm driver as well within Netbox; with the modification to line 504 in view.py of the api resolved the error..

schwos:
I can also confirm that this did not effect the following drivers and the upgrades made to increase their functionality within the napalm tool and more specifically the get_facts and get_environment calls within Napalm; 
napalm_ftos, or 
napalm_dell10os

PR #8121, opened 2021-12-20
Fix #8097: Re-fix markdown table rendering

### LINKED ISSUE #8097: Markdown tables missing CSS style
### NetBox version

v3.0.8

### Python version

3.8

### Steps to Reproduce

1. after upgrading to v3.0, a plugin that I wrote before which has render_markdown in its HTML page can not work properly. I would want to render a table by inputting
Screenshot from 2021-12-17 18-03-18
but what I get is not a table
2. I tried this input in some other places using `render_mardown` like Circuits.providers.comments, they showed the same result:
Screenshot from 2021-12-17 18-00-17

### Expected Behavior

a table would show up

### Observed Behavior

no table show up

kkthxbye-code:
### Fixes: #8097

The CSS for markdown-python table styling (https://github.com/netbox-community/netbox/commit/2fce7ebd8fee511a0fef8b3aada11037e8ef8e9f) was not ported to 3.0. This PR re-adds the CSS and adds the ´rendered-markdown´ CSS class where missing.


jeremystretch:
Rather than encapsulating every instance of `render_markdown` in an element with the `rendered-markdown` class, it may make more sense to just output `<div class="rendered-markdown">{content}</div>` from the filter directly. This would ensure it's not missed in future implementations of the filter.

kkthxbye-code:
Makes sense, I'll try to implement that.

kkthxbye-code:
render_markdown now wraps the output as you suggested. I added a check for empty strings as `render_markdown` is chained with `placeholder` in some spots. 

Removed any other usages of rendered-markdown.

PR #8188, opened 2021-12-30
Update tables.py after moved tag helper

### LINKED ISSUE #8187: Many views broken in 3.1.3 due to moved "templatetags/tag.html"
### NetBox version

v3.1.3

### Python version

3.8

### Steps to Reproduce

1. Update from 3.1.2 to 3.1.3
2. Try to open Devices or Virtual Machines views, there is an error about missing "utilities/templatetags/tag.html"

It seem to stem from https://github.com/netbox-community/netbox/commit/f369b5f588ad29d9de0f36cfdce586bd3769e7eb where `tables.py` was not updated after file moved?

Now it might also be that I screwed up my update, but it is the same automated process that has worked for a number of updates.

### Expected Behavior

The views should work like before.

### Observed Behavior

It seems like the file `netbox/utilities/tables.py` has a wrong path for an include, seems some files have moved around.

It is only result of my quick troubleshooting while trying to get the production instance back up - my fault, should have actually clicked around plenty in the testing instance before updating production....

bluikko:
### Fixes: #8187

May be a naive workaround, I am not familiar with the code base but it gets my instance working.

### DIFF
--- a/netbox/utilities/tables.py
+++ b/netbox/utilities/tables.py
@@ -382,7 +382,7 @@ class TagColumn(tables.TemplateColumn):
     """
     template_code = """
     {% for tag in value.all %}
-        {% include 'utilities/templatetags/tag.html' %}
+        {% include 'helpers/tag.html' %}
     {% empty %}
         <span class="text-muted">&mdash;</span>
     {% endfor %}


jeremystretch:
Thanks for this contribution. I've resolved this in the `develop` branch by calling `{% tag %}` directly (and added a test to catch any future issues).

PR #8284, opened 2022-01-07
Fixes #8262 - Add Cable stat for Tenant

### LINKED ISSUE #8262: Add cable stat for tenant
### NetBox version

v3.1.4

### Feature type

### Proposed functionality

I would like to have a stats field for cables on the tenant screen. Same as for VLAN, circuits, etc.

image

### Use case

 It will make a better overview of the services provided to the tenant.

### Database changes

I believe no change is required in the database.

### External dependencies

_No response_

jasonyates:
### Fixes: #8262
Adds a stats field for cables on the tenant screen along with relevant API output


jeremystretch:
The REST API changes included here will need to be reverted. This were not part of the change proposed in #8262, and should not be included. We also cannot accommodate a modification to the API in patch releases.

MarianRychtecky:
Hi all, sorry for the confusion - I think it makes sense to have cable stats available over API. It is a good point I missed.

PR #8329, opened 2022-01-12
Adding asdot notation to ASN views

### LINKED ISSUE #8293: support for asdot notation for 4 Byte ASN
### NetBox version

v3.1.5

### Feature type

### Proposed functionality

In the netbox url/ipam/asns/ and  netbox ur>/ipam/asns/#/ view the AS Number should be displayed in asplain and asdot notation. 
The db model does not have to be changed as we can still only store the asplain value in the db.

The actual conversion can be done in the web browser with some simple javascript like this:

```
 function conv_asplain2asdot() {
        asplain = document.AScalc.asplain.value;
        var asdot1 = (Math.floor(asplain/65536));
        var asdot2 = asplain%65536;
        document.AScalc.asdot1.value = asdot1;
        document.AScalc.asdot2.value = asdot2;
};

function conv_asdot2asplain() {
        var asdot1 = document.AScalc.asdot1.value;
        var asdot2 = document.AScalc.asdot2.value;
        asplain = ((parseInt(asdot1) * 65536) + parseInt(asdot2));
        document.AScalc.asplain.value = asplain;
};
```

### Use case

4 Byte AS number get more popular these day and the asdot notation seems to be much easier to work with. 

### Database changes

not required, conversion can be done in javascript at runtime as mentioned above.

### External dependencies

none

jasonyates:
### Fixes: #8293 
Adds custom property to asn model to compute asdot notation if required.
Updates asn view to show asdot notation if one exists in the format xxxxx (yyy.yyy)
Adds a custom column renderer to asn table to display asdot notation if one exists

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -130,6 +130,14 @@ class ASN(PrimaryModel):
     def get_absolute_url(self):
         return reverse('ipam:asn', args=[self.pk])
 
+    @property
+    def asdot_notation(self):
+        # Return asdot notation for an ASN larger than 65535
+        if self.asn > 65535:
+            return '{}.{}'.format(self.asn // 65536, self.asn % 65536)
+        else:
+            return None
+
 
 @extras_features('custom_fields', 'custom_links', 'export_templates', 'tags', 'webhooks')
 class Aggregate(GetAvailablePrefixesMixin, PrimaryModel):
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -125,21 +125,18 @@ class ASN(PrimaryModel):
         verbose_name_plural = 'ASNs'
 
     def __str__(self):
-        if self.asn > 65535:
-            return 'AS{} ({}.{})'.format(self.asn, self.asn // 65536, self.asn % 65536)
-        else:
-            return f'AS{self.asn}'
+        return f'AS{self.asn_with_asdot}'
 
     def get_absolute_url(self):
         return reverse('ipam:asn', args=[self.pk])
 
     @property
-    def asdot_notation(self):
-        # Return asdot notation for an ASN larger than 65535
+    def asn_with_asdot(self):
+        # Return asn with asdot notation for an ASN larger than 65535 otherwise return the plain ASN
         if self.asn > 65535:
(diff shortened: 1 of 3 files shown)


jasonyates:
@jeremystretch Soliciting some feedback on this.

Should the page title also be updated to include the asdot notation or not?

E.g.
Screenshot 2022-01-12 at 14 06 36

jeremystretch:
Sure? No strong opinion here.

netbox/ipam/models/ip.py
```diff
@@ -125,11 +125,22 @@ class Meta:
         verbose_name_plural = 'ASNs'
 
     def __str__(self):
-        return f'AS{self.asn}'
+        if self.asn > 65535:
+            return 'AS{} ({}.{})'.format(self.asn, self.asn // 65536, self.asn % 65536)
```
jeremystretch, line 129: This should probably return `self.asdot_notation`.

netbox/ipam/models/ip.py
```diff
@@ -125,11 +125,22 @@ class Meta:
...
+        else:
+            return f'AS{self.asn}'
 
     def get_absolute_url(self):
         return reverse('ipam:asn', args=[self.pk])
 
+    @property
+    def asdot_notation(self):
```
jeremystretch, line 137: Maybe rename this to `asn_with_asdot` or something like that, and return either the ASN with ASDOT notiation or just the ASN. Then you can reference it directly from within the ASNTable and object template.

jeremystretch (changes requested):
Thanks for this! Just had a couple suggestions.

netbox/ipam/models/ip.py
DanSheps: I think we should also stick with fstrings if we can IMO

jasonyates: Updated to use fstring and to return `asn_with_asdot` per below suggestion. Resulting code is much cleaner.

netbox/ipam/models/ip.py
jasonyates: Great suggestion. That works much cleaner. Updated ASNTable & object template.

jeremystretch (approved):

jeremystretch:
Excellent, thanks @jasonyates!

PR #8339, opened 2022-01-12
Fixes: #8295 Render the payload_url of a Webhook with Jinja2

### LINKED ISSUE #8295: Allow jinja2 code in webhook payload_url
### NetBox version

v3.1.5

### Feature type

New functionality

### Proposed functionality

Allow the use of jinja2 code in the webhook payload_url.

### Use case

I have a custom field where I store the ID of an object in my NMS. This allows me to utilise NetBox to control my monitoring (pause monitoring if the device is set to offline as an example). The current setup requires some middleware where I POST to a generic URL and this web server again make a call to the REST API of the monitoring system using the ID in the custom field from the payload data. If I could reference the obj.cf_nms_id directly in the payload_url it would allow me to skip this middleware.

The payload_url would have to be generated on each webhook event.

### Database changes

None

### External dependencies

_No response_

rodvand:
### Fixes: #8295 
Render the payload_url with Jinja2 when triggering a webhook.

- Update markdown documentation
- Expand on the help text for the Webhook model to reference support of Jinja2 templating


jeremystretch:
Thanks @rodvand!

PR #8420, opened 2022-01-20
Fixes #7853 - Add speed and duplex

### LINKED ISSUE #7853: Interface Speed & Duplex
### NetBox version

v3.0.10

### Feature type

Data model extension

### Proposed functionality

Generally NetBox should support most base functions an interface supports. One obvious function missing in the current model is the ability to set interface speed.

None of the maintainers could recall anyone ever requesting this, which is quite surprising.  I completed a rough search through the backlog of issues, and could not find anything obvious existing.

### Use case

There 2 possible use cases I am thinking of, but there could be more:

- People who want to hard set speed for whatever reason like removing auto-negotiation with fickle peer interfaces.
- People who want a way to track actual speed of a port without going into the granularity of tracking the SFP inserted into the port.  Example: SFP28 capable ports with SFP+ optics installed are currently hard to model.  A lot of people likely do not want to deal with the overhead of SFP management in Netbox, though that could become a thing later.

### Database changes

_No response_

### External dependencies

_No response_

DanSheps:
### Fixes: #7853

Add Speed and Duplex tot he interface model


jeremystretch:
Awesome work, thank you!

PR #8441, opened 2022-01-24
Fixes #8391: Install date should appear empty when exported

### LINKED ISSUE #8391: Circuit Install Date set to NULL exports to csv as â€”
### NetBox version

v3.1.6

### Python version

3.9

### Steps to Reproduce

1. Create a circuit with no Install Date
2. Export circuits to CSV
3. Notice Install Date Field value displayed as `â€”`

### Expected Behavior

Install Date shows as NULL or empty

### Observed Behavior

Install Date shows as `â€”`

seulsale:
### Fixes: #8391

Set the default value to export installed_date values as empty when they are null

### DIFF
--- a/netbox/circuits/tables.py
+++ b/netbox/circuits/tables.py
@@ -140,9 +140,6 @@ class CircuitTable(BaseTable):
         template_code=CIRCUITTERMINATION_LINK,
         verbose_name='Side Z'
     )
-    install_date = tables.DateColumn(
-        default=''
-    )
     commit_rate = CommitRateColumn()
     comments = MarkdownColumn()
     tags = TagColumn(
--- a/netbox/circuits/tables.py
+++ b/netbox/circuits/tables.py
@@ -140,6 +140,9 @@ class CircuitTable(BaseTable):
         template_code=CIRCUITTERMINATION_LINK,
         verbose_name='Side Z'
     )
+    install_date = tables.DateColumn(
+        default=''
+    )
     commit_rate = CommitRateColumn()
     comments = MarkdownColumn()
     tags = TagColumn(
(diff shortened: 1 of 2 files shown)


netbox/circuits/tables.py
```diff
@@ -140,6 +140,9 @@ class CircuitTable(BaseTable):
         template_code=CIRCUITTERMINATION_LINK,
         verbose_name='Side Z'
     )
+    install_date = tables.DateColumn(
+        default=''
```
jeremystretch, line 144: Unfortunately this approach removes the default placeholder (&mdash;) in the table when rendered in the UI, which we want to keep.

It looks like DateColumn inherits from TemplateColumn: The ideal fix for this might actually be to fix it upstream in django-tables2 by adding a `value()` method to DateColumn to better handle null values.

seulsale: Yes, I agree in that it might be better to fix it upstream. This behavior will occur again in the case of exporting other null date fields, like date_added in the Aggregate model. 

I can go and open an issue there.

seulsale: The django-tables2 library doesn't seem to very active at the moment.

What do you think of implementing a custom DateColumn and improving the handling of null values within the project? 

It could be added into the utilities/tables.py module.

jeremystretch: I'd rather avoid that, but it was be a reasonable short-term fix. Do you want me to keep #8391 assigned to you?

seulsale: I don't like it too much either, but sure it can be a short-term fix. Yes please keep it assigned to me, I can work on it this weekend.

jeremystretch: Thanks!

seulsale: This is the best way I found to handle null values for all fields of type Date and DateTime, feel free to give comments on the approach.

jeremystretch: Looks great to me! Thanks for your work.

Issue #8486, opened 2022-01-28
update social-core-auth dependency version


oasys:
### NetBox version

v3.1.6

### Feature type

Change to existing functionality

### Proposed functionality

Bump social-auth-core dependency in `requirements.txt` to `4.2.0`.

### Use case

Use Okta SSO with Netbox.   A regression in social_auth that prevented this from working is now fixed in python-social-auth/social-core#588 and included in version 4.2.0.

### Database changes

none

### External dependencies

Update requirements.txt:

```diff
-social-auth-core==4.1.0
+social-auth-core==4.2.0
```

oasys:
For reference, here's the error I was receiving 


For reference, here's the symptom I was seeing:

After successful authentication, netbox redirects to `https://<okta custom url>/oauth2/.well-known/openid-configuration?client_id=xxxxxxx` which is a 404.  If I manually go to `/.well-known/openid-configuration?client_id=xxxxxxxx` (without the /oauth2/) in my browser, I see the correct json data.

If you want, assign to me, and I'll submit the (one-line) PR.

jeremystretch:
This will be bumped in the v3.1.6 release.

davekempe:
@jeremystretch do you mean in 3.1.7?

PR #8503, opened 2022-01-31
Added 'all' extra to social-auth-core in requirements.txt

### LINKED ISSUE #8502: Update requirements.txt for social-auth-core to use 'all' extras
### NetBox version

v3.1.6

### Feature type

### Proposed functionality

In order to support OIDC and SAML in social-auth-core, requirements.txt would just need to be updated with "all" extras:

social-auth-core[all]==4.1.0

This is in base_requirements.txt, but it didn't make its way over to requirements.txt.

### Use case

OIDC and SAML-based SSO

### Database changes

N/A

### External dependencies

Change social-auth-core to social-auth-core[all] in requirements

adamweiss:
### Fixes: #8502 

Changes social-auth-core in requirements.txt to social-auth-core[all], allowing for OIDC and SAML SSO.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ Pillow==8.4.0
 psycopg2-binary==2.9.3
 PyYAML==6.0
 social-auth-app-django==5.0.0
-social-auth-core==4.1.0
+social-auth-core[all]==4.1.0
 svgwrite==1.4.1
 tablib==3.1.0
 


davekempe:
https://github.com/netbox-community/netbox/issues/8486 Don't forget about this - we are waiting on this bump to fix a bug

jeremystretch:
@adamweiss it looks like CI has a problem building the wheel for `xmlsec`. I've run into this as well; presumably because the `xmlsec1` system library is not installed. This library isn't part of our documented installation process either. This reinforces my preference per [my earlier comment](https://github.com/netbox-community/netbox/issues/8502#issuecomment-1027299265) to defer the `[all]` option for only those users who need it.

adamweiss:
Yeah, `xmlsec1` does complicate things a bit. If you end up leaving out the `[all]` option, it should probably also be removed from [base_requirements.txt](https://github.com/netbox-community/netbox/blob/develop/base_requirements.txt#L103). It would also probably help for this to be documented [here](https://demo.netbox.dev/static/docs/administration/authentication/#single-sign-on-sso), since the link to the supported backends seems to indicate OIDC and SAML are supported.

PR #8510, opened 2022-02-01
Closes #8032: Upgrade to Django 4.0

### LINKED ISSUE #8032: Upgrade to Django 4.0
### Proposed Changes

Move from Django 3.2 to 4.0.

### Justification

The most interesting bits from Django 4.0 for us are:

* Functional unique constraints
* Built-in Redis cache backend
* Template-based form rendering

Full release notes are [here](https://docs.djangoproject.com/en/4.0/releases/4.0/).

jeremystretch:
### Closes: #8032

**NOTE:** Because graphene-django does not yet support Django v4.0, this branch is currently using a fork which has added v4.0 support. Once the relevant PR has been merged, this will be returned to the normal package.

- Bumps Django to 4.0
- Tweak old migrations to avoid introducing [no-op migrations](https://docs.djangoproject.com/en/4.0/releases/4.0/#migrations-autodetector-changes)
- Remove dependency on private `is_safe_url()` function


jeremystretch:
We're now pulling `graphene-django` from its `v2` branch in anticipation of the v2.16 release (date TBD).

PR #8518, opened 2022-02-02
feat: plugins can extend the GraphQL schema

### LINKED ISSUE #8405: Enable plugins to extend the GraphQL API
### NetBox version

v3.1.6

### Feature type

New functionality

### Proposed functionality

Provide a mechanism through which a NetBox plugin can extend the GraphQL API with its own query class(es). I don't have a strong opinion on the optimal way to achieve this, but it should be fairly straightforward.

Currently, the GraphQL schema is formed by combining the root query from each app within NetBox:

```python
class Query(
    CircuitsQuery,
    DCIMQuery,
    ExtrasQuery,
    IPAMQuery,
    TenancyQuery,
    UsersQuery,
    VirtualizationQuery,
    WirelessQuery,
    graphene.ObjectType
):
    pass

schema = graphene.Schema(query=Query, auto_camelcase=False)
```

We can extend this approach to allow dynamic inclusions. Something like:

```python
query_classes = [
    CircuitsQuery,
    DCIMQuery,
    ExtrasQuery,
    IPAMQuery,
    TenancyQuery,
    UsersQuery,
    VirtualizationQuery,
    WirelessQuery,
    *get_plugin_queries()
]

class Query(*query_classes, graphene.ObjectType):
    pass

schema = graphene.Schema(query=Query, auto_camelcase=False)
```

`get_plugin_queries()` above is just a placeholder function, but we can probably handle the registration of GraphQL queries the same way we do for template extensions and navigation menu items today (by designating a configurable path to a variable within the plugin which contains these components).

This was originally proposed by @Jenjen1324 in WG83333.

### Use case

This enables plugins to extend NetBox's GraphQL API, e.g. to add their own models.

### Database changes

_No response_

### External dependencies

_No response_

Jenjen1324:
### Fixes: #8405

This change allows plugin to extend the GraphQL schema with queries, mutations, subscriptions and custom types. This allows full flexibility in extending the API for Plugins.

Remaining Tasks:

 - [x] Plugin endpoint for amending the schema
 - [ ] Tests
 - [ ] Plugin Documentation updates

### DIFF
--- a/netbox/extras/plugins/__init__.py
+++ b/netbox/extras/plugins/__init__.py
@@ -1,6 +1,9 @@
 import collections
 import inspect
+import graphene
 from packaging import version
+
+from graphql.type.definition import GraphQLType
 
 from django.apps import AppConfig
 from django.core.exceptions import ImproperlyConfigured
@@ -15,6 +18,10 @@ from extras.plugins.utils import import_object
 # Initialize plugin registry stores
 registry['plugin_template_extensions'] = collections.defaultdict(list)
 registry['plugin_menu_items'] = {}
+registry['graphql_queries'] = []
+registry['graphql_mutations'] = []
+registry['graphql_subscriptions'] = []
+registry['graphql_types'] = []
 
 
 #
@@ -50,6 +57,12 @@ class PluginConfig(AppConfig):
     # Django-rq queues dedicated to the plugin
     queues = []
 
+    # Graphene GraphQL schema extension points
+    graphql_queries = []
+    graphql_mutations = []
+    graphql_subscriptions = []
+    graphql_types = []
+
     # Default integration paths. Plugin authors can override these to customize the paths to
     # integrated components.
     template_extensions = 'template_content.template_extensions'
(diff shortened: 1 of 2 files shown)


Jenjen1324:
@jeremystretch I got a bit tied up at work last week. Would be glad if you could take a peek at the changes and tell me what you think about the stuff that goes a bit beyond what was discussed in the issue.

Also I'm not too sure regarding the loading of the plugin data into the graphql part. I'm happy to change up things if you have any suggestions

jeremystretch:
Hey @Jenjen1324 would you mind rebasing this off of the `feature` branch? That's where we're staging all the changes for v3.2. It shouldn't give you too much trouble but let me know if you run into any conflicts you're unsure about. Thanks!

jeremystretch:
NetBox's GraphQL API doesn't currently support mutations or subscriptions. We can certainly look at adding support for those, however that's beyond the scope of #8405, which merely extends the existing functionality to plugins. Because we're adding so much to the plugins framework in v3.2, I want to be very careful to avoid scope creep.

jeremystretch:
@Jenjen1324 I've just pushed some earlier work on #8405 under PR #8562. Please have a look and let me know if you think anything is missing.

I'd like to get the first v3.2 beta out next week, so it's important that we wrap this up pretty quickly. Thanks!

jeremystretch:
@Jenjen1324 thanks for your work on this. I went ahead with merging my own PR #8562 to ensure that we keep moving forward with this, as it involves work beyond what's included in this PR. I'm going to close this one as I believe the necessary implementation has been covered, however you're welcome to continue evaluating and contributing throughout the v3.2 beta cycle if you see opportunities for improvement. Thanks again!

PR #8565, opened 2022-02-05
Fixes #8398: Add ConfigParam.size to enlarge specific config fields

### LINKED ISSUE #8398: Banner edit fields should be larger
### NetBox version

v3.1.6

### Feature type

### Proposed functionality

Currently the Login, Top and Bottom banner edit fields in Config revisions are small input fields: they only show a few words at the time. It makes editing the banners somewhat hard.

I propose that the input fields are shown larger, preferably as multi-line input fields, to make it easier to edit the HTML-containing banners.

### Use case

Communicating upcoming maintenances or specific login instructions is easier when the banner input fields are easier to edit.

### Database changes

None

### External dependencies

None

markkuleinio:
### Fixes: #8398 

Adds `ConfigParam.size` attribute that can be used for enlarging the text input field for specific configuration parameters.

Also adds `size="80"` for those parameters that seems to benefit from longer input field.

Unfortunately my Django skills did not allow for using textareas for the banner fields, it was not a simple `field=forms.Textarea` change as it resulted as `TypeError: __init__() got an unexpected keyword argument 'required'` in https://github.com/netbox-community/netbox/blob/aff55881df21e0edeef589dbaf4204c1be6a0d42/netbox/extras/forms/config.py#L27
 🤔

### DIFF
--- a/netbox/extras/forms/config.py
+++ b/netbox/extras/forms/config.py
@@ -25,8 +25,6 @@ class FormMetaclass(forms.models.ModelFormMetaclass):
             }
             field_kwargs.update(**param.field_kwargs)
             param_fields[param.name] = param.field(**field_kwargs)
-            if param.size:
-                param_fields[param.name].widget.attrs["size"] = param.size
         attrs.update(param_fields)
 
         return super().__new__(mcs, name, bases, attrs)
--- a/netbox/extras/forms/config.py
+++ b/netbox/extras/forms/config.py
@@ -25,6 +25,8 @@ class FormMetaclass(forms.models.ModelFormMetaclass):
             }
             field_kwargs.update(**param.field_kwargs)
             param_fields[param.name] = param.field(**field_kwargs)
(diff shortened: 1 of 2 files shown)


jeremystretch:

jeremystretch:
Rather than restricting the available parameter to a specific size (which doesn't even apply to most form widgets), IMO it would be preferable to allow the specification of a widget instance with arbitrary parameters. For the banners specifically, I would expect we'd use Django's [`Textarea` widget](https://docs.djangoproject.com/en/stable/ref/forms/widgets/#textarea) rather than just increasing the size of the `TextInput`.

jeremystretch (changes requested):
Per my comment [here](https://github.com/netbox-community/netbox/pull/8565#issuecomment-1031591736), we don't need to add a `size` attribute: We can just specify the widget to employ in `field_kwargs` for the relevant fields.

PR #8573, opened 2022-02-07
Fixes 8553: Fix contacts and ASNs missing in the search dropdown and …

### LINKED ISSUE #8553: Provider network, ASN, and contact options missins from global search selector
### NetBox version

v3.1.7

### Python version

3.8

### Steps to Reproduce

Issue 1:
1. Click on filter selections for global search
2. Look for "Contacts" in the list

Issue 2:
1. ~Search for a name in global search~
2. ~View the "assignments" column~

### Expected Behavior

Issue 1:
"Contacts" should be a selectable option, akin to to all the other searchable models

~Issue 2:~
~Number of assignments should be shown~

### Observed Behavior

Issue 1:
"Contacts" are not listed as a selectable option, although searching for a contact still works (which may or may not include various other results)

Filter List

~Issue 2:~
~Instead of the number of assignments being shown, a hyphen is displayed instead.~

991jo:
…contacts not showing the number of assigned objects

### Fixes: #8553 

The contact and ASN object were not shown in the global object search dropdown.
Also the contant queryset used in the global search was not annotated with the `assigned_count`

### DIFF
--- a/netbox/netbox/constants.py
+++ b/netbox/netbox/constants.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from typing import Dict
 
 from circuits.filtersets import CircuitFilterSet, ProviderFilterSet, ProviderNetworkFilterSet
 from circuits.models import Circuit, ProviderNetwork, Provider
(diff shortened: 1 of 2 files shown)


docs/release-notes/version-3.1.md
```diff
@@ -5,6 +5,7 @@
 ### Bug Fixes
 
 * [#8561](https://github.com/netbox-community/netbox/issues/8561) - Include option to connect a rear port to a console port
+* [#8553](https://github.com/netbox-community/netbox/issues/8553) - Fix contacts and ASNs missing in the search dropdown and contacts not showing the number of assigned objects
```
jeremystretch, line 8: You can omit the changelog to avoid potential merge conflicts with other PRs.

netbox/netbox/constants.py
```diff
@@ -186,7 +186,7 @@
         'url': 'tenancy:tenant_list',
     }),
     ('contact', {
-        'queryset': Contact.objects.prefetch_related('group', 'assignments'),
+        'queryset': Contact.objects.prefetch_related('group', 'assignments').annotate(assignment_count=count_related(ContactAssignment, 'contact')),
```
jeremystretch, line 189: Change is out of scope for #8553. I've asked the OP to submit a separate bug report for this.

netbox/netbox/forms.py
```diff
@@ -7,6 +7,7 @@
     ('Circuits', (
         ('provider', 'Providers'),
         ('circuit', 'Circuits'),
+        ('providernetwork', 'Provider Networks'),
```
jeremystretch, line 10: While this fixes the immediate issue, it does nothing to prevent future models from also being inadvertently omitted from the list. A more complete solution would be to dynamically render the dropdown list from the `SEARCH_TYPES` dictionary. Would you like to take a shot at coming up with something?

jeremystretch (changes requested):

netbox/netbox/constants.py
991jo: okay, I will add remove that and make a new PR when the new issue is opened.

docs/release-notes/version-3.1.md
991jo: okay, I can do that. Do you add things to the changelog as an additional commit before merging or releasing or how does the "getting things into the changelog" process work?

netbox/netbox/constants.py
jeremystretch: The second issue is #8577 FYI

docs/release-notes/version-3.1.md
jeremystretch: Whichever maintainer merges the PR is responsible for following up in the change log to ensure everything is recorded. (Admittedly it's not the best process but it works well enough with incurring any additional overhead.)

netbox/netbox/forms.py
991jo: Okay, I added some code that adds a layer of hierarchy to the `constants.py` and then builds the choices from that.

jeremystretch (changes requested):

netbox/netbox/views/__init__.py
```diff
@@ -167,27 +167,30 @@ def get(self, request):
...
-                table = table(filtered_queryset, orderable=False)
-                table.paginate(per_page=SEARCH_MAX_RESULTS)
-
-                if table.page:
-                    results.append({
-                        'name': queryset.model._meta.verbose_name_plural,
-                        'table': table,
-                        'url': f"{reverse(url)}?q={form.cleaned_data.get('q')}"
-                    })
+                for _, value in SEARCH_TYPE_HIERARCHY.items():
```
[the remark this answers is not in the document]
991jo: I understand that the for loop is not a very nice solution.

Regarding having the hierarchy separate:
I thought the fundamental problem of this bug is that 2 partially redundant data structures have to be maintained by hand.
Separating the hierarchy from the dict containing the search types still results in 2 different places holding the information. Or did I misunderstand you?

I could keep the `SEARCH_TYPE_HIERARCHY` as is, but add code to build a flattened version for building the view. This should solve the loop problem while not having to maintain partially redundant data structures by hand.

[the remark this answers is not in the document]
991jo: I prepared it as described above, the `SEARCH_TYPE_HIERARCHY` still exists, and then `OBJECT_CHOICES` and `SEARCH_TYPES` are generated from that at program start.
The hierarchy is split into per-app dicts to keep in readable. That many layers of brackets and indentation would get to messy. Also the OrderedDicts are required because we might want to preserve the order in `OBJECT_CHOICES`

The result is that I don't even have to touch the code in the view and therefore not introduce a for-loop there.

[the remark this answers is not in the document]
991jo: @jeremystretch This PR is waiting for a response from you.

netbox/netbox/constants.py
```diff
@@ -26,169 +27,212 @@
...
-        'table': RackReservationTable,
-        'url': 'dcim:rackreservation_list',
-    }),
-    ('location', {
-        'queryset': Location.objects.add_related_count(
-            Location.objects.add_related_count(
-                Location.objects.all(),
-                Device,
+
+CIRCUIT_TYPES = OrderedDict(
```
jeremystretch, line 31: FYI there's no need to use `OrderedDict` in this case on Python 3.7+ since dictionary ordering is now guaranteed out of the box. It's fine as-is though, I'll clean it up along with some other stuff in a project-wide sweep at some point.

jeremystretch (approved):

jeremystretch:
Thanks for this!

PR #8621, opened 2022-02-13
Enable tab completion in nbshell

### LINKED ISSUE #8620: Add tab completion to nbshell
### NetBox version

v3.1.7

### Feature type

### Proposed functionality

The [NetBox Shell](https://netbox.readthedocs.io/en/stable/administration/netbox-shell/) is nice for hacking on NetBox. It would be even more convenient if it supported tab completion like the normal Python shell.

### Use case

Makes hacking on NetBox easier.

### Database changes

_No response_

### External dependencies

_No response_

JonathonReinhart:
### Fixes: #8620

This uses the standard library [`readline`](https://docs.python.org/3/library/readline.html) and [`rlcompleter`](https://docs.python.org/3/library/rlcompleter.html) modules to add tab-completion to the NetBox Shell (`mange.py nbshell`).

If `readline` support isn't enabled in the Python installation, then this code does nothing.

### DIFF
--- a/netbox/extras/management/commands/nbshell.py
+++ b/netbox/extras/management/commands/nbshell.py
@@ -70,10 +70,23 @@ class Command(BaseCommand):
         return namespace
 
     def handle(self, **options):
+        namespace = self.get_namespace()
+
         # If Python code has been passed, execute it and exit.
         if options['command']:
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks for the handy addition!

PR #8735, opened 2022-02-24
Add description to tag table search function

### LINKED ISSUE #8629: Search tags by their description
### NetBox version

v3.1.7

### Feature type

### Proposed functionality

Allow the search bar at the Tags page to also search through the Description column.

### Use case

One use case for tags is for identifying VLANs and, in our use case, they are abbreviated, e.g., `NWPT` stands for **New Port**.
We'd like to be able to search for the term **New Port**, which is in the `Description` field, without needing to use the browser's  `Ctrl+F`.

### Database changes

None.

### External dependencies

N/A

minitriga:
### Fixes: #8629 

Add description to tag table search function.

### DIFF
--- a/docs/release-notes/version-3.1.md
+++ b/docs/release-notes/version-3.1.md
@@ -2,6 +2,10 @@
 
 ## v3.1.9 (FUTURE)
 
+### Enhancements
+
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks @minitriga!

PR #8742, opened 2022-02-24
Allow filtering on Core models for Contacts

### LINKED ISSUE #8645: Searchable Contact Assignments per Content-Type in API/UI
### NetBox version

v3.1.7

### Feature type

New functionality

### Proposed functionality

I propose adding query/filter parameters in the API & UI on each Content-Type, that is supported by Contact Assignments.

So it would be possible to just filter by Contact & Contact-Role to receive all matching Objects:
`virtualization/virtual-machines/?contact=Some%20Team&contact_role=administration`

If the direct integration is not the design pattern desired by the maintainers, and it should be kept as a standalone endpoint/functionality, the UI at least requires a combined filterable list on all Contact Assignments, that basically represents the API endpoint `/api/tenancy/contact-assignments/` for users being able to leverage this useful feature.
Like with `/dcim/interfaces/` in the UI.

### Use case

Currently the only places for viewing Contact Assignments in the UI are:
* on the Contact itself
* on the Contact-Role itself
* on the Object itself

All these pages provide no further filtering on shown Contact Assignments.

Consider having a Contact that has more than a handful of Devices, VMs, Virtual-Chassis and Clusters assigned - how would one leverage the UI to find all VMs, the Contact is an "Administrator" for?

Via API it is possible to filter for NetBox Objects by Contact & Contact-Role, e.g.:
`api/tenancy/contact-assignments/?content_type=virtualization.virtualmachine&contact_id=6&role=administration`

### Database changes

None

### External dependencies

None

minitriga:
### Fixes: #8645 

Implements filtering for all core models that have contacts. 

- [x] Filter by contacts
- [x] Filter by contact roles
- [x] Allow contacts to be shown in the table columns


netbox/circuits/api/serializers.py
```diff
@@ -17,12 +17,17 @@
 class ProviderSerializer(PrimaryModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='circuits-api:provider-detail')
     circuit_count = serializers.IntegerField(read_only=True)
+    contacts = NestedContactAssignmentSerializer(
```
jeremystretch, line 20: I'm afraid this approach of following the reverse relationship is not tenable, as it can result in a substantial performance hit. Notice that as a rule we don't follow reverse relations for any model in the REST API (though there may be a handful of exceptions).

REST API consumers should instead query the contact assignments endpoint to retrieve these objects.

minitriga:
@jeremystretch I have removed all changes to the API.

netbox/circuits/api/serializers.py
minitriga: I have removed the API changes from the PR.

jeremystretch (changes requested):
Looks pretty good. Noted a few minor issues for cleanup. Also, it looks like we still need to define `contacts` columns on ProviderTable and CircuitTable to be consistent.

Thanks for your work on this!

netbox/tenancy/forms/forms.py
```diff
@@ -1,12 +1,13 @@
...
 from django.utils.translation import gettext as _
 
-from tenancy.models import Tenant, TenantGroup
+from tenancy.models import *
 from utilities.forms import DynamicModelChoiceField, DynamicModelMultipleChoiceField
 
 __all__ = (
     'TenancyForm',
     'TenancyFilterForm',
+    'ContactModelFilterForm'
```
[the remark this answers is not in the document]
minitriga: I have corrected this in my latest push

jeremystretch:
Excellent work, thank you!

Issue #8764, opened 2022-02-28
Dynamic choice fields don't support plugin API endpoints


jeremystretch:
### NetBox version

v3.2-beta1

### Python version

3.8

### Steps to Reproduce

1. Create a form with a DynamicModelChoiceField referencing a plugin model.
2. Attempt to render the form within a view.

(This bug was originally reported by @peteeckel in WG8333.)

### Expected Behavior

The field should render and populate choices dynamically using the REST API backend.

### Observed Behavior

An exception is raised due the view namespace not being properly resolved:

```
django.urls.exceptions.NoReverseMatch: 'pluginname-api' is not a registered namespace
```

peteeckel:
I just tested it and still get an exception:
```
Traceback (most recent call last):
  File "/opt/netbox/lib/python3.8/site-packages/django/template/base.py", line 870, in _resolve_lookup
    current = getattr(current, bit)
  File "/opt/netbox/lib/python3.8/site-packages/django/utils/functional.py", line 48, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/opt/netbox/lib/python3.8/site-packages/django/forms/forms.py", line 455, in changed_data
    return [name for name, bf in self._bound_items() if bf._has_changed()]
  File "/opt/netbox/lib/python3.8/site-packages/django/forms/forms.py", line 455, in <listcomp>
    return [name for name, bf in self._bound_items() if bf._has_changed()]
  File "/opt/netbox/lib/python3.8/site-packages/django/forms/forms.py", line 154, in _bound_items
    yield name, self[name]
  File "/opt/netbox/lib/python3.8/site-packages/django/forms/forms.py", line 177, in __getitem__
    bound_field = field.get_bound_field(self, name)
  File "/opt/netbox/netbox/utilities/forms/fields/dynamic.py", line 106, in get_bound_field
    widget.attrs['data-url'] = reverse(viewname)
  File "/opt/netbox/lib/python3.8/site-packages/django/urls/base.py", line 77, in reverse
    raise NoReverseMatch(
django.urls.exceptions.NoReverseMatch: 'netbox_dns' is not a registered namespace inside 'plugins-api'
Internal Server Error: /plugins/netbox-dns/records/
Traceback (most recent call last):
  File "/opt/netbox/lib/python3.8/site-packages/django/forms/forms.py", line 168, in __getitem__
    field = self.fields[name]
KeyError: 'changed_data'
```

[...]

PR #8767, opened 2022-03-01
Correct the endpoint for plugin API view names

peteeckel:
### Fixes: #8764

@jeremystretch: Yesterday's commit to issue #8764 does not actually insert the correct label for API endpoints. I had a look at the code and found the missing `-api` suffix for the application label to be the likely culprit.

Tested this with NetBox DNS, it works for me.


jeremystretch:
Ah, not sure how I missed that. Thanks!

PR #8816, opened 2022-03-07
Closes #8232: Add color show full 100% utilization

### LINKED ISSUE #8232: Show full 100% utilization bar in a different color from 99.9%
### NetBox version

v3.1.4

### Feature type

### Proposed functionality

I propose when utilization is full, i.e. there is nothing free left, the utilization bar would not show the same red as it shows for 90% or 99.9%.  
Instead it could show a (slightly) different color, say brighter/darker red or purple.

### Use case

Make it easy for human operator to distinguish from the utilization bar whether utilization is totally full 100%, or not totally full <100%.

Currently the operator cannot just scan the list of utilization bars for whether something is 100% full or almost full because the bar has the same color and may display too similar bar length for 99%/99.9% and 100%.

100% is a slightly special case since there is nothing free left inside, so indicate that to the operator for spotting it easily.

### Database changes

None.

### External dependencies

None.

tranthang2404:

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -248,9 +248,9 @@ class Aggregate(GetAvailablePrefixesMixin, PrimaryModel):
         """
         queryset = Prefix.objects.filter(prefix__net_contained_or_equal=str(self.prefix))
         child_prefixes = netaddr.IPSet([p.prefix for p in queryset])
-        utilization = int(float(child_prefixes.size) / self.prefix.size * 100)
+        utilization = float(child_prefixes.size) / self.prefix.size * 100
 
-        return min(utilization, 100)
+        return utilization
 
 
 @extras_features('custom_fields', 'custom_links', 'export_templates', 'tags', 'webhooks')
@@ -548,7 +548,7 @@ class Prefix(GetAvailablePrefixesMixin, PrimaryModel):
                 vrf=self.vrf
             )
             child_prefixes = netaddr.IPSet([p.prefix for p in queryset])
-            utilization = int(float(child_prefixes.size) / self.prefix.size * 100)
+            utilization = float(child_prefixes.size) / self.prefix.size * 100
         else:
             # Compile an IPSet to avoid counting duplicate IPs
             child_ips = netaddr.IPSet(
@@ -558,9 +558,9 @@ class Prefix(GetAvailablePrefixesMixin, PrimaryModel):
             prefix_size = self.prefix.size
             if self.prefix.version == 4 and self.prefix.prefixlen < 31 and not self.is_pool:
                 prefix_size -= 2
-            utilization = int(float(child_ips.size) / prefix_size * 100)
+            utilization = float(child_ips.size) / prefix_size * 100
 
-        return min(utilization, 100)
+        return utilization
 
 
 @extras_features('custom_fields', 'custom_links', 'export_templates', 'tags', 'webhooks')
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -250,7 +250,7 @@ class Aggregate(GetAvailablePrefixesMixin, PrimaryModel):
         child_prefixes = netaddr.IPSet([p.prefix for p in queryset])
         utilization = float(child_prefixes.size) / self.prefix.size * 100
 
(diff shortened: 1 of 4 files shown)


netbox/utilities/templates/helpers/utilization_graph.html
```diff
@@ -9,13 +9,18 @@
...
+      {% endif %}
     >
       {% if utilization >= 25 %}{{ utilization }}%{% endif %}
     </div>
     {% if utilization < 25 %}
       <span class="ps-1">{{ utilization }}%</span>
     {% endif %}
   </div>
-{% endif %}
+{% endif %}
```
tovin07, line 26: Please add a newline

netbox/utilities/templates/helpers/utilization_graph.html
```diff
@@ -9,13 +9,18 @@
       aria-valuemin="0"
       aria-valuemax="100"
       aria-valuenow="{{ utilization }}"
+      {% if is_full_danger %}
+      class="progress-bar"
+      style="width: {{ utilization }}%; background-color: #800080;"
```
tovin07, line 14: Do you have any demo image for this? Including all four colors in light mode and dark mode?

netbox/utilities/templatetags/helpers.py
```diff
@@ -389,6 +389,9 @@ def utilization_graph(utilization, warning_threshold=75, danger_threshold=90):
     """
     Display a horizontal bar graph indicating a percentage of utilization.
     """
+    is_full_danger = False
+    if utilization == 100:
+        is_full_danger = True
```
jeremystretch, line 394: This logic can be cleaned up considerably. There should be no need to store a variable indicating whether the utilization value is 100: we can just reference utilization directly.

tranthang2404: yeah, i misunderstand

netbox/utilities/templates/helpers/utilization_graph.html
tranthang2404: Sure, I choose constants color for 100% utilization

tranthang2404: Darkmode:
image
Lightmode:
image

netbox/ipam/models/ip.py
```diff
@@ -248,9 +248,9 @@ def get_utilization(self):
         """
         queryset = Prefix.objects.filter(prefix__net_contained_or_equal=str(self.prefix))
         child_prefixes = netaddr.IPSet([p.prefix for p in queryset])
-        utilization = int(float(child_prefixes.size) / self.prefix.size * 100)
+        utilization = float(child_prefixes.size) / self.prefix.size * 100
 
-        return min(utilization, 100)
```
jeremystretch, line 253: See below

netbox/utilities/templates/helpers/utilization_graph.html
```diff
@@ -9,13 +9,18 @@
       aria-valuemin="0"
       aria-valuemax="100"
       aria-valuenow="{{ utilization }}"
+      {% if utilization == 100 %}
+      class="progress-bar"
+      style="width: {{ utilization }}%; background-color: #800080;"
```
jeremystretch, line 14: The bar color needs to be changed by modifying the `bar_class` context var inside the `utilization_graph` template tag. (`bg-secondary` seems like a good choice.) Please avoid hard-coding color values directly in the HTML.

jeremystretch (changes requested):

netbox/ipam/models/ip.py
```diff
@@ -558,9 +558,9 @@ def get_utilization(self):
             prefix_size = self.prefix.size
             if self.prefix.version == 4 and self.prefix.prefixlen < 31 and not self.is_pool:
                 prefix_size -= 2
-            utilization = int(float(child_ips.size) / prefix_size * 100)
+            utilization = float(child_ips.size) / prefix_size * 100
 
-        return min(utilization, 100)
```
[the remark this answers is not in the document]
tranthang2404: Sorry, I forgot that context

jeremystretch:
Thanks for your work on this @tranthang2404!

PR #8857, opened 2022-03-11
Implement UI to handle MB,GB and TB input for VM's

### LINKED ISSUE #8437: Use MB/GB/TB suffix for VM memory input
### NetBox version

v3.1.6

### Feature type

### Proposed functionality

Allow inputting memory with a MB/GB/TB suffix instead of defaulting to MB.

### Use case

The use case is to make it easier for users to input the correct amount of memory for the virtual machine.

It is far more common today to input GBs of memory rather than MBs of memory and having to memorize or calculate the correct amount of MBs the wanted GB represent is tedious.

### Database changes

None

### External dependencies

One method of solving it is by changing the form input field to this https://pypi.org/project/django-sizefield/.

minitriga:
### Fixes: #8437 

- [x] Fix issue with bulk edit wiping previous values when blank


jeremystretch:
I took a stab at fixing bulk editing. My solution was to allow the unit field to remain blank, which avoids returning True when `field.has_changed()` is called. Let me know what you think.

minitriga:
We may want to use this new "MemoryField" for the disk size to because of #8897

DanSheps:
> We may want to use this new "MemoryField" for the disk size to because of #8897

Yes, please do.

netbox/utilities/forms/fields.py
```diff
@@ -150,6 +153,72 @@ def to_python(self, value):
...
+            size, unit = data
+            if size and not unit:
+                raise forms.ValidationError("Memory unit cannot be blank.")
+            elif size and unit:
+                return size * cls.MULTIPLIERS[unit]
+            elif not size and unit:
+                raise forms.ValidationError("Please enter a memory value when unit is selected.")
+
+
+class DiskField(forms.MultiValueField):
```
jeremystretch, line 190: This is essentially a duplicate of `MemoryField`; both the memory and disk fields should use the same form field class. Instead of declaring `MULTIPLIERS` as a constant, you could accept a list of units as a kwarg on init.

netbox/utilities/forms/fields.py
```diff
@@ -150,6 +153,72 @@ def to_python(self, value):
...
+class MemoryField(forms.MultiValueField):
+    widget = widgets.MemoryWidget
+    MULTIPLIERS = {
+        MemoryUnitChoices.UNIT_MB: 1024**0,
+        MemoryUnitChoices.UNIT_GB: 1024**1,
+        MemoryUnitChoices.UNIT_TB: 1024**2,
+    }
+
+    def __init__(self, **kwargs):
+        print(kwargs)
```
jeremystretch, line 165:
```suggestion
```

netbox/utilities/forms/fields.py
```diff
@@ -150,6 +153,72 @@ def to_python(self, value):
...
+    @classmethod
+    def compress(cls, data):
+        if data:
+            size, unit = data
+            if size and not unit:
+                raise forms.ValidationError("Memory unit cannot be blank.")
+            elif size and unit:
+                return size * cls.MULTIPLIERS[unit]
+            elif not size and unit:
+                raise forms.ValidationError("Please enter a memory value when unit is selected.")
```
jeremystretch, line 187: I would simplify this a bit.

```suggestion
            if not size or not unit:
                raise forms.ValidationError("Bot size and unit must be specified.")
            return size * cls.MULTIPLIERS[unit]
```

netbox/utilities/forms/widgets.py
```diff
@@ -310,3 +313,39 @@ def __init__(self, *args, **kwargs):
...
+        super(MemoryWidget, self).__init__(widgets, attrs)
+
+    def decompress(self, value):
+        if value:
+            return [value, MemoryUnitChoices.UNIT_MB]
+        else:
+            return ['', '']
+
+
+class DiskWidget(forms.MultiWidget):
```
jeremystretch, line 336: Same as above; we should only need one widget class.

netbox/virtualization/choices.py
```diff
@@ -5,6 +5,23 @@
 # VirtualMachines
 #
 
+class MemoryUnitChoices(ChoiceSet):
```
jeremystretch, line 8: `ChoiceSet` should always declare a `CHOICES` constant. Rather than trying to combine them, we should create a separate `ChoiceSet` class for each purpose. The `UNIT_*` constants can be defined outside of the classes for reuse to cut down on boilerplate.

jeremystretch (changes requested):

jeremystretch:
@minitriga are you still interested in working on this?

minitriga:
Yes. I will make the changes. Was on holiday for two weeks.

minitriga:
Not to sure how to handle passing the multipliers to the compress function. Do you have any ideas? There are different multipliers depending on if we are storing the RAM or DISK.

```python
class MemoryField(forms.MultiValueField):

    def __init__(self, **kwargs):
        self.multipliers = kwargs['multipliers']
        widget = widgets.MemoryWidget(choices=kwargs['choices'])
        fields = (
            forms.IntegerField(required=False),
            forms.ChoiceField(
                choices=kwargs['choices'],
                required=False
            ),
        )
        super(MemoryField, self).__init__(
            fields=fields, required=False,
            require_all_fields=False, widget=widget,
        )

    @classmethod
    def compress(cls, data):
        if data:
            size, unit = data
            if not size or not unit:
                raise forms.ValidationError("Size and unit must be specified.")
            return size * cls.MULTIPLIERS[unit]
 ```

netbox/virtualization/choices.py
minitriga: Done.

netbox/utilities/forms/widgets.py
minitriga: Done.

netbox/utilities/forms/fields.py
minitriga: I'm having trouble passing the MULTIPLIERS through to the compress function as the compress function class is based on django MultiValueField not MemoryField. Need to work on a solution to get around this.

PR #8873, opened 2022-03-14
Closes: #8457 - Nonracked  Devices Location/Site

### LINKED ISSUE #8457: "Add device" button on location, site views
### NetBox version

v3.1.6

### Feature type

New functionality

### Proposed functionality

It would be nice to be able to create a device in a location without assigning it to a rack. 

### Use case

If you have a device in a room, that is not in a rack, currently, it is not possible to display this in NetBox without creating a fake rack or assigning the device to another rack. 

### Database changes

Right now, if you create a non-rack device, the `position` and `face` attributes of the device are set to `null`. Since the location is stored in the device, the `rack` attribute could be set to null to assign it to a location without assigning it to a rack.

### External dependencies

_No response_

minitriga:
### Fixes: #8457 

Ability to add non-racked devices from location and site page.

### DIFF
--- a/netbox/templates/dcim/inc/nonracked_devices.html
+++ b/netbox/templates/dcim/inc/nonracked_devices.html
@@ -1,0 +1,62 @@
+{% load helpers %}
+
+<div class="card">
+<h5 class="card-header">
+    Non-Racked Devices
+</h5>
+<div class="card-body">
+{% if nonracked_devices %}
+    <table class="table table-hover">
+        <tr>
+            <th>Name</th>
+            <th>Role</th>
+            <th>Type</th>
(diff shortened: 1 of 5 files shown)


netbox/templates/dcim/inc/nonracked_devices.html
```diff
@@ -0,0 +1,62 @@
+{% load helpers %}
+
+<div class="card">
+<h5 class="card-header">
+    Non-Racked Devices
+</h5>
+<div class="card-body">
+{% if nonracked_devices %}
```
jeremystretch, line 8: SiteView needs to be extended to include `nonracked_devices` in its context, otherwise this will never display.

PR #8874, opened 2022-03-14
Closes: #8575 - Show Racks on Cable Table and Cable Page

### LINKED ISSUE #8575: Show racks A and B at cables table
### NetBox version

v3.0.11

### Feature type

New functionality

### Proposed functionality

Cables table (dcim/cables/) shows
- Side A/Side B
- Termination A/Termination B
I propose also showing Rack A/Rack B in which respective Side A/Side B Devices are installed (as a link, same as Sides and Terminations) if it is a device. For Circuits (which do not have a Rack) the field could be empty or null.

Rack A/Rack B fields which will get it's values automatically (and will be not not editable manually) from Side A/Side B of a given cable can also be added to Cable properties.

### Use case

Employees operating cables often need to know from which rack to which it runs. Finding this out now requires extra clicking to view each side. Outputting racks right in the table with cable properties will make things more convenient and faster.

### Database changes

_No response_

### External dependencies

_No response_

minitriga:
### Fixes: #8575 

- [x] Implement table changes for rack_a and rack_b
- [x] add racks to cableview

image


jeremystretch:
Looks good, thanks!

PR #8876, opened 2022-03-15
Assign Device to Position

### LINKED ISSUE #8552: Improve ways to insert devices in rack positions
### NetBox version

v3.1.6

### Feature type

### Proposed functionality

Adding a device to a rack can only be done through the 'Edit' form of a device.
This is simply, but not very inuitive and user friendly.

It would benefit to change the way this is done in the GUI.
1. The edit form's Position field could get a 'Pick from rack' button. This would popup the rack and allow you to select the position visually.
2. From the rack point of view
When hovering over a free position, only 'Add Device' is shown. This links to the creation form of devices.
It would benefit of this give you the option to link an existing device the, rather than creating a new one. I got the feeling that this rarely done from that view.
Same goes for the 'Add non racked device' in this view.

### Use case

This will help to visualize the racks when inserting devices into them.
Sometimes you know it has to go 'below that switch', without knowing the exact location

### Database changes

None

### External dependencies

None

minitriga:
### Fixes: #8552

Allow user to assign a device to an empty slot within the Rack UI. 

image

This also works for non-racked devices.


netbox/dcim/views.py
```diff
@@ -1463,6 +1463,46 @@ class DeviceEditView(generic.ObjectEditView):
...
+        })
+
+    def post(self, request):
+        form = forms.DeviceAssignForm(request.POST)
+        table = None
+
+        if form.is_valid():
+
+            devices = self.queryset.prefetch_related('rack')
+            print(devices[0].rack.pk)
```
minitriga, line 1495: remove debug

jeremystretch:
I'm not comfortable with the pattern used here, as it's going to be very difficult to maintain long-term. I suspect you're replicating the workflow for IP address assignment; we actually need to ditch that because of how "snowflakey" it is.

minitriga:
@jeremystretch Yes the workflow is replicated from the IP address assignment. Happy to work on another workflow if you have any ideas?

abhi1693:
Not sure, why IP assignment and this would ultimately be ditched. But, I have several complaints from my team for device assignment to a rack being extremely difficult especially when they are standing in front of a rack. First, a device is needed to be searched from the device page, go into edit and then fill up the rack-related values whereas clicking on the rack U position and allocating a device directly from the search is much more suitable. This way they don't have to go back and forth between the rack and the device pages to validate the data.

Having something like this is very beneficial.

DanSheps:
> Not sure, why IP assignment and this would ultimately be ditched

This is in reference to the exact method in which this is implemented in both IP Address assignment and here.  The root of the problem is that IPAddress assignments use a M2O GFK (If I am remembering correctly) and as such cannot be assigned directly on the address model and traditional assignment methods do not work as well.

minitriga:
Going to close this as I have not worked on it in ages and don't have time at the moment to work on it. We can reopen in future if needed.

PR #8878, opened 2022-03-15
#8233 Restrict API key usage by source IP

### LINKED ISSUE #8233: Restrict API key usage by source IP
### NetBox version

v3.1.4

### Feature type

### Proposed functionality

Limit the access to the API with a certain API key by source IPs

### Use case

We have API keys that belong to 'service accounts'. The source IP of these services consumers are known.
We should be able to limit the usage of these service account API keys to only the known sources.
This cannot be done by firewalls but needs to go to the application layer

If an API key gets compromised, it cannot be used outside of that scope of source IPs

### Database changes

API Keys
- add multivalue 'Source IP'

### External dependencies

_No response_

PieterL75:


netbox/users/models.py
```diff
@@ -227,6 +238,24 @@ def is_expired(self):
...
+        if not self.allowed_ips:
+            return True
+
+        try:
+            ip_address = ipaddress.ip_address(raw_ip_address)
+        except ValueError:
+            raise ValidationError(f"{raw_ip_address} is an invalid IP address")
+
+        for ipnet in self.allowed_ips:
+            if ip_address in ipaddress.ip_network(ipnet):
```
jeremystretch, line 254: Try to avoid using overly terse variable names; they can impede comprehension.
```suggestion
        for ip_network in self.allowed_ips:
            if ip_address in ipaddress.ip_network(ip_network):
```

jeremystretch (changes requested):

netbox/netbox/api/authentication.py
```diff
@@ -18,6 +23,20 @@ def authenticate_credentials(self, key):
         except model.DoesNotExist:
             raise exceptions.AuthenticationFailed("Invalid token")
 
+        # Verify source IP is allowed
+        request = self.request
+        if token.allowed_ips and request:
```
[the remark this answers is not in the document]
PieterL75: Yes. I do super of the authenicate(), that super calls the authenticate_credentials without a request parameter.
After the super I call the authenticate_credentials with 'request'...

hm.. maybe I need to move the source-ip check in the authenticate()... I needed the token, but I get that from the super...

Going to test that out..

base_requirements.txt
```diff
@@ -113,3 +113,7 @@ svgwrite
 # Tabular dataset library (for table-based exports)
 # https://github.com/jazzband/tablib
 tablib
+
+# It changes comma separated widget to list based in admin panel
+# https://github.com/gradam/django-better-admin-arrayfield
+django_better_admin_arrayfield
```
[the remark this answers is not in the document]
PieterL75: The addon provides an easier way to add objects to a comma separated list.. The only drawback is, that this ony is for the Admin side... not for the Userside.. 
I'll remove it for now, but I think this kind of widget is a good addition to both admin and user gui

netbox/netbox/api/authentication.py
```diff
@@ -18,6 +23,20 @@ def authenticate_credentials(self, key):
...
+            # Replace 'HTTP_X_REAL_IP' with the settings variable choosen in #8867
+            if 'HTTP_X_REAL_IP' in request.META:
+                clientip = request.META['HTTP_X_REAL_IP'].split(",")[0].strip()
+            elif 'REMOTE_ADDR' in request.META:
+                clientip = request.META['REMOTE_ADDR']
+            else:
+                raise exceptions.AuthenticationFailed(f"The request HTTP headers (HTTP_X_REAL_IP, REMOTE_ADDR) are missing or do not contain a valid source IP.")
```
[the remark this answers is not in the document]
PieterL75: I actually don't test if the request has a valid ip. Only if the httpheaders exist..
Removed the " or do not contain a valid source IP" part of the message

base_requirements.txt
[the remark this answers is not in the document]
jeremystretch: I understand; however every dependency that we introduce brings with it additional burdens. And smaller packages like these are highly susceptible to abandonment, which creates more work for us in the long-term.

netbox/netbox/api/authentication.py
```diff
@@ -11,6 +11,28 @@ class TokenAuthentication(authentication.TokenAuthentication):
...
+                    clientip = request.META['REMOTE_ADDR']
+                else:
+                    raise exceptions.AuthenticationFailed(f"A HTTP header containing the SourceIP (HTTP_X_REAL_IP, REMOTE_ADDR) is missing from the request.")
+
+                if not token.validate_client_ip(clientip):
+                    raise exceptions.AuthenticationFailed(f"Source IP {clientip} is not allowed to use this token.")
+
+            return token_user, token
+        else:
+            return None
```
jeremystretch, line 34:
```suggestion
    return authenticationresult
```

netbox/users/models.py
```diff
@@ -227,6 +237,24 @@ def is_expired(self):
...
+        """
+        Checks that an ip address falls within the allowed ips.
+        """
+        if not self.allowed_ips:
+            return True
+
+        try:
+            ip_address = ipaddress.ip_address(raw_ip_address)
+        except ValueError:
+            raise ValidationError(f"{raw_ip_address} is an invalid IP address")
```
jeremystretch, line 250: I feel like we should return a 400 Bad Request response if the IP address is invalid. (We can handle this under `authenticate()` by catching ValidationError. Otherwise, raising `ValidationError` (which AFAICT won't be caught by DRF) is just going to trigger a 500 so we should figure out something a bit cleaner.

netbox/users/models.py
```diff
@@ -227,6 +237,24 @@ def is_expired(self):
             return False
         return True
 
+    def validate_client_ip(self, raw_ip_address):
+        """
+        Checks that an ip address falls within the allowed ips.
```
jeremystretch, line 242:
```suggestion
        Checks that an IP address falls within the allowed IPs.
```

netbox/users/admin/__init__.py
```diff
@@ -58,9 +58,15 @@ def get_inlines(self, request, obj):
...
     form = forms.TokenAdminForm
     list_display = [
-        'key', 'user', 'created', 'expires', 'write_enabled', 'description'
+        'key', 'user', 'created', 'expires', 'write_enabled', 'description', 'list_allowed_ips'
     ]
 
+    def list_allowed_ips(self, obj):
+        if obj.allowed_ips:
+            return obj.allowed_ips
+        return 'Any'
```
jeremystretch, line 67:
```suggestion
        return obj.allowed_ips or 'Any'
```

jeremystretch (changes requested):

netbox/users/admin/__init__.py
PieterL75: Nice ! Whish I knew how to code this clean :-)

netbox/users/models.py
PieterL75: Not sure how to do this.. Any example code that accomplish this ?

jeremystretch: We should be able to catch the `ValidationError` and return a [`HttpResponseBadRequest`](https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpResponseBadRequest) response, with an error message indicating the invalid value.

PieterL75: Hm.. not easy :-)
Do I need to extend the ExceptionHandlingMiddleware to catch the error and generate a 400 ?
a simple `return HttpResponseBadRequest(f"ValidationErrorInfo: {ValidationErrorInfo.message}")` just throws another error that the return of authenticate() expects 2 values, but does not return the 404 error to the client

PieterL75: Found it :
```
except ValidationError as ValidationErrorInfo:
    raise exceptions.ValidationError(f"The Source IP in the HTTP Header {http_header} has a ValidationError: {ValidationErrorInfo.message}")
```

This raises a `HTTP/1.1 400 Bad Request`

PieterL75: @jeremystretch Does this solution seems good ?

jeremystretch:
Should be all set now.

PieterL75:
@jeremystretch its failing on the documentation build.. anything you can fix plz ?

jeremystretch:
@PieterL75 sorry, I'm not sure what happened here. There are definitely still some discrepancies with the branch. I may have messed something up earlier.

Given that there's not a ton of change in this PR, it might be cleanest to create a new fork of the `feature` branch and just copy over the changes by hand from the diff here, then resubmit that as a new PR.

Additionally, the migration file will need to be bumped to `0003_token_allowed_ips` so as not to conflict with the v3.2 migration.

PieterL75:
As suggested by @jeremystretch : Closing this PR and recreating it with a new Fork of the feature branch

PR #8907, opened 2022-03-18
Fixes #8813 Retain search value after submitting

### LINKED ISSUE #8813: Search bar contents not retained after submitting
### NetBox version

v3.1.8

### Python version

3.8

### Steps to Reproduce

1. Enter a search term into the main search bar and submit
2. See empty search bar

### Expected Behavior

1. Enter a search term into the main search bar and submit
2. Search term should still be present

### Observed Behavior

1. Enter a search term into the main search bar and submit
2. Search bar is empty

PieterL75:

### DIFF
--- a/netbox/utilities/templates/search/searchbar.html
+++ b/netbox/utilities/templates/search/searchbar.html
@@ -5,11 +5,7 @@
     aria-label="Search"
     placeholder="Search"
     class="form-control"
-    value=""
-    {% if request.GET.q %}
-      autofocus
-      onfocus="this.value = '{{ request.GET.q|escapejs }}';" 
-    {% endif %}
+    value="{{ request.GET.q|escape }}"
   />
 
   <input name="obj_type" hidden type="text" class="search-obj-type" />
--- a/netbox/utilities/templates/search/searchbar.html
+++ b/netbox/utilities/templates/search/searchbar.html
@@ -5,7 +5,11 @@
     aria-label="Search"
     placeholder="Search"
     class="form-control"
-    value="{{ request.GET.q }}"
+    value=""
+    {% if request.GET.q %}
+      autofocus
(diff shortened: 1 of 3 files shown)


netbox/utilities/templates/search/searchbar.html
```diff
@@ -5,7 +5,11 @@
     aria-label="Search"
     placeholder="Search"
     class="form-control"
-    value="{{ request.GET.q }}"
+    value=""
```
jeremystretch, line 8: Why not leave this as-is?

PieterL75: By setting the value with a javascript, the cursor moves to the end of the input field.. if I just set the value, the cursor is at the start of the input field...

jeremystretch: But it doesn't even render the value until the user clicks the field, which doesn't resolve #8813. Just leaving it as `value="{{ request.GET.q }}"` ensures the value is populated, and the cursor defaults to the end of the string on focus.

PieterL75: The autofocus part takes care of setting the focus. 
If I leave out the onfocus part. then the cursor is at the start of the value.
You need to merge that PR #8908 to make this work. The autofocus is currently overruled by the piece of javascript in the netbox.ts

jeremystretch: Ok. For now, can we change this to `value="{{ request.GET.q }}"` just to close out #8813? I suspect #8908 might need some additional tweaking but I'm waiting for Matt to weigh in. We can add the autofocus bit under that PR to ensure it's correlated with the appropriate FR.

PieterL75: I removed the autocomplete and onfocus from this PR.. Moving that to the #8908

PR #8908, opened 2022-03-18
Fixes #7839 Focus on searchbar

### LINKED ISSUE #7839: Focus on searchbar
### NetBox version

v3.0.10

### Feature type

### Proposed functionality

When opening netbox, I always have to click in the searchbox, before I can type.

It would be easier if the searchbox already has focus by default.
Only when there is a form to be completed, then the first field of the form needs to have the focus

### Use case

Less clicking

### Database changes

None

### External dependencies

None

PieterL75:


jeremystretch:
This doesn't appear to be working for me. If I search using the global search bar from the home page, I'm taken to the results page, but the search bar at top is empty. Only when I click on it does the value populate. It does not appear to be auto-focusing.

PieterL75:
What browser/os are you using? Maybe my JavaScript is too browser native

PieterL75:
I think the problem lies with the fact that both mobile and desktop searchbar are generated, but that the mobile part is not visible. That gives me 2times an input field 'q' with autofocus enabled.
How is the mobile/desktop handled ? I can see to put that logic in the DTF so that it only displays one of the two

PieterL75:
Looking at the implementation of that dual-searchbar, I think we need to remove the autofocus from the nav-mobile section.. and also populate the value (as you suggested earlier)

PieterL75:
I changed the logic, so that the focus is only set on the desktop navbar, and not the mobile one..

PR #8925, opened 2022-03-21
Fix #8924 - Speed up rendering of the script list

### LINKED ISSUE #8924: Script list is slow when having large script files/a lot of scripts
### NetBox version

v3.1.9

### Python version

3.9

### Steps to Reproduce

1. Create script file with following content in the script directory (test1.py): [content](https://gist.github.com/kkthxbye-code/75aa97e2fc5506da999bb6b01b3c1ed8)
2. Duplicate it a bunch of times: `for i in {2..20}; do cp test1.py "test$i.py"; done`
3. Open the script list view - `/extras/scripts/`

### Expected Behavior

Load times should not be multiple seconds

### Observed Behavior

On my i7 9700 the load times are roughly 2.3 seconds for the above example. On our production server (weaker single core performance) the performance impact is larger.

I have implemented a fix (https://github.com/netbox-community/netbox/pull/8925) which reduces the load time for the reproduction example from ~2300 ms to ~160 ms.

kkthxbye-code:
### Fixes: #8924

Loading the script list can become slow when having multiple large scripts. The reason is that django instantiates the Script class for each access in the template. This can be very slow in some cases.

The fix is simply to set `do_not_call_in_templates` = True on the callable before passing them to the template. Please let me know if there is a better option.

The template was changed slightly, as the script variable did not return the name when not instantiated. Name is now taken from Meta instead.

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -753,6 +753,8 @@ class ScriptListView(ContentTypePermissionRequiredMixin, View):
 
         for _scripts in scripts.values():
             for script in _scripts.values():
+                # Prevent django from instantiating the class on all accesses
+                script.do_not_call_in_templates = True
                 script.result = results.get(script.full_name)
 
         return render(request, 'extras/script_list.html', {
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -753,8 +753,6 @@ class ScriptListView(ContentTypePermissionRequiredMixin, View):
 
         for _scripts in scripts.values():
             for script in _scripts.values():
-                # Prevent django from instantiating the class on all accesses
-                script.do_not_call_in_templates = True
                 script.result = results.get(script.full_name)
 
         return render(request, 'extras/script_list.html', {
--- a/netbox/templates/extras/script_list.html
+++ b/netbox/templates/extras/script_list.html
@@ -34,7 +34,7 @@
                 {% for class_name, script in module_scripts.items %}
                   <tr>
                     <td>
-                      <a href="{% url 'extras:script' module=script.module name=class_name %}" name="script.{{ class_name }}">{{ script.Meta.name }}</a>
+                      <a href="{% url 'extras:script' module=script.module name=class_name %}" name="script.{{ class_name }}">{{ script.name }}</a>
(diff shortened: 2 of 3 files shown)


netbox/extras/views.py
```diff
@@ -753,6 +753,8 @@ def get(self, request):
 
         for _scripts in scripts.values():
             for script in _scripts.values():
+                # Prevent django from instantiating the class on all accesses
+                script.do_not_call_in_templates = True
```
jeremystretch, line 757: Seems like this should be set on the base class, rather than being set by the view.

netbox/templates/extras/script_list.html
```diff
@@ -34,7 +34,7 @@ <h5 class="card-header">
                 {% for class_name, script in module_scripts.items %}
                   <tr>
                     <td>
-                      <a href="{% url 'extras:script' module=script.module name=class_name %}" name="script.{{ class_name }}">{{ script }}</a>
+                      <a href="{% url 'extras:script' module=script.module name=class_name %}" name="script.{{ class_name }}">{{ script.Meta.name }}</a>
```
jeremystretch, line 37: `script.Meta.name` is not a required attribute. Instead, this should reference `script.name`. (Incidentally, `name` should default to `self.__name__`, but that's a separate issue.)

jeremystretch, line 37: `script.Meta.name` is not a required attribute. Instead, this should reference `script.name`.

jeremystretch (changes requested):

netbox/templates/extras/script_list.html
kkthxbye-code: Ahh, I totally missed the `name` class attribute. I'll change that shortly.

kkthxbye-code:
@jeremystretch - I moved `do_not_call_in_templates`  to the base class.

I also changed the template to use script.name. I had to change it from `self.__class__.__name__` to `self.__name__`. I tested on our production system (v3.1.5) and if you omit Meta.name the name of the script will be `"type"`. I might be missing how it's supposed to work, but I think that's a bug.

PR #8945, opened 2022-03-23
Fix #8785 - allow wildcard dns records

### LINKED ISSUE #8785: Using wildcards in DNS names
### NetBox version

v3.1.8

### Python version

3.9

### Steps to Reproduce

This is a copy of https://github.com/netbox-community/netbox/issues/4973 because I can't comment there, but the issue is not resolved.

1. Edit IP Address in IPAM
2. Try to save with * in DNS name like: "*.example.local"

Like in https://github.com/netbox-community/netbox/issues/3722 the regex needs to be adjusted.

### Expected Behavior

IP address saved correctly.
https://en.wikipedia.org/wiki/Wildcard_DNS_record

### Observed Behavior

Error: "Only alphanumeric characters, hyphens, periods, and underscores are allowed in DNS names"

fmlshai:
### Fixes: #8785
Added * to the DNSValidator regex to allow wildcard domains like *.example.com

### DIFF
--- a/netbox/ipam/validators.py
+++ b/netbox/ipam/validators.py
@@ -24,7 +24,7 @@ class MinPrefixLengthValidator(BaseValidator):
 
 
 DNSValidator = RegexValidator(
-    regex='^[0-9A-Za-z*._-]+$',
+    regex=r'^([0-9A-Za-z_-]+|\*)(\.[0-9A-Za-z_-]+)*\.?$',
     message='Only alphanumeric characters, asterisks, hyphens, periods, and underscores are allowed in DNS names',
     code='invalid'
 )
--- a/netbox/ipam/validators.py
+++ b/netbox/ipam/validators.py
@@ -24,7 +24,7 @@ class MinPrefixLengthValidator(BaseValidator):
 
 
 DNSValidator = RegexValidator(
-    regex='^[0-9A-Za-z._-]+$',
-    message='Only alphanumeric characters, hyphens, periods, and underscores are allowed in DNS names',
+    regex='^[0-9A-Za-z*._-]+$',
+    message='Only alphanumeric characters, asterisks, hyphens, periods, and underscores are allowed in DNS names',
     code='invalid'
 )


netbox/ipam/validators.py
```diff
@@ -24,7 +24,7 @@ def compare(self, a, b):
 
 
 DNSValidator = RegexValidator(
-    regex='^[0-9A-Za-z._-]+$',
-    message='Only alphanumeric characters, hyphens, periods, and underscores are allowed in DNS names',
+    regex='^[0-9A-Za-z*._-]+$',
```
jeremystretch, line 27: This would allow any number of asterisks, anywhere within the string, which would not be valid. A wildcard must be defined as a single asterisk only.

jeremystretch (changes requested):

netbox/ipam/validators.py
fmlshai: Well, you're right. Didn't think about that case. I'll try to figure it out.

hSaria: This ensures a wildcard can only be present at the beginning of a record `'^(\\*\\.)?[0-9A-Za-z._-]+$'`. With a raw string, it would be `r'^(\*\.)?[0-9A-Za-z._-]+$'`.

Also, it might be more accurate to change the RegEx to `r'^(\*\.)?[0-9A-Za-z_-]+(\.[0-9A-Za-z_-]+)*$'` to prevent a record with multiple consecutive `.`, like `www..example..com`.

kkthxbye-code: >Wildcards can be present in any level

I'm pretty sure this is wrong. Wildcards must be leftmost in the DNS string.

https://datatracker.ietf.org/doc/html/rfc4592#section-2.1.1

Nothing4You: @kkthxbye-code if you read a bit further, to https://datatracker.ietf.org/doc/html/rfc4592#section-2.1.3

> A wildcard domain name can have subdomains.

kkthxbye-code: Does that mean that wildcards can be at any level? The bind maintainer seems to think nothttps://www.mail-archive.com/bind-users@lists.isc.org/msg23505.html - also wikipedia seems to think wildcards in the middle are invalid. Do you have a production example of a DNS server that respects wildcard domains in the middle?

If that was the case, I would prefer if we just remove validation altogether.

Nothing4You: bind9 accepts me adding the following record:
`netbox-8945.*.demo.central-intelligence.agency. 300 IN TXT "https://github.com/netbox-community/netbox/pull/8945"`
it does log the following warning:
> warning: ownername 'netbox-8945.*.demo.central-intelligence.agency' contains a non-terminal wildcard

See also https://kb.isc.org/docs/what-is-an-empty-non-terminal, which suggests to avoid such records, but they do seem to be valid.

To note is however, that it seems that the use of an asterisk in these records does not turn them into wildcard records.
They seem to be used as literal asterisks in these cases.
That is also what the email you linked mentions.
It does not explicitly mention asterisks, but it mentions that only leading asterisk labels are treated as wildcard, without going into detail on asterisk labels in other places.

Try this:
`dig txt "netbox-8945.*.demo.central-intelligence.agency"`

https://datatracker.ietf.org/doc/html/rfc4592#section-2.2.1 specifically lists this type of record as non-wildcard:

```
   To illustrate what is meant by existence consider this complete zone:

      $ORIGIN example.
      example.                 3600 IN  SOA   <SOA RDATA>
      example.                 3600     NS    ns.example.com.
      example.                 3600     NS    ns.example.net.
      *.example.               3600     TXT   "this is a wildcard"
      *.example.               3600     MX    10 host1.example.
      sub.*.example.           3600     TXT   "this is not a wildcard"
      host1.example.           3600     A     192.0.2.1
      _ssh._tcp.host1.example. 3600     SRV   <SRV RDATA>
      _ssh._tcp.host2.example. 3600     SRV   <SRV RDATA>
      subdel.example.          3600     NS    ns.example.com.
      subdel.example.          3600     NS    ns.example.net.
```

For an `A` record, however, I can't create this with `nsupdate`:
```
nsupdate commands queued:
server localhost
zone central-intelligence.agency
update add    netbox-8945.*.demo.central-intelligence.agency. 300 IN A 127.0.0.1
send
answer

[S]end, [e]dit, send and edit [a]gain, [q]uit: [s]

check-names failed: bad owner 'netbox-8945.*.demo.central-intelligence.agency'
syntax error
nsupdate returned 1, press enter to edit again
```

So in the end it seems that this is at least not usable for IPs, therefore likely not useful in netbox.

kkthxbye-code: Thanks for spending the time doing such thorough digging. I guess we'll let Jeremy decide on the final outcome.

jeremystretch: Since the FR does not specify a use case beyond e.g. `*.example.com`, let's keep it simple and just allow terminating wildcards.

@fmlshai are you still interested in making this change?

fmlshai: Yes. Thanks for all the effort you guys put into this. I'll update the PR as soon as I find the time for it.

PR #8957, opened 2022-03-24
Fix #8956: Save old JobResults

### LINKED ISSUE #8956: Retain old JobResults
### NetBox version

v.3.1.9

### Feature type

### Proposed functionality

Currently all old JobResults are deleted when a new JobResult is in a terminal state. I propose that old JobResults are preserved.

### Use case

Deleting old JobResults make it hard to use for automation through the API. Imagine a case where two runs of the same script is triggered via. the API. The API returns an id for the JobResult that can be used to query for the result. The RQ worker will run the scripts one after the other. When the first script finishes, you have the runtime of the second script to get the result. If you do not retrieve the result before the second script finishes, the JobResult is deleted and impossible to retrieve.

Also contributing to the issue is that JobResults are not webhook enabled, which could provide a workaround in some cases. But that's a different feature request.

### Database changes

None

### External dependencies

_No response_

kkthxbye-code:
### Fixes: #8956

Removes all logic that deletes old JobResults. Also tweaks the view querysets to only return the newest terminal result.

Lastly removes unused JobResults in the homepage view, which seems like a leftover from the v2 -> v3 split.

### DIFF
--- a/docs/administration/housekeeping.md
+++ b/docs/administration/housekeeping.md
@@ -4,6 +4,7 @@ NetBox includes a `housekeeping` management command that should be run nightly. 
 
 * Clearing expired authentication sessions from the database
 * Deleting changelog records older than the configured [retention time](../configuration/dynamic-settings.md#changelog_retention)
+* Deleting job result records older than the configured [retention time](../configuration/dynamic-settings.md#jobresult_retention)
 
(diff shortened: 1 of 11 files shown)


kkthxbye-code:
Is it enough to just change the base branch on the PR via. the github ui?

kkthxbye-code:
Should be good to go now then I guess.

jeremystretch:
Thanks for all your work on this @kkthxbye-code!

PR #8962, opened 2022-03-25
#8830 Adding ClusterXL

### LINKED ISSUE #8830: Add ClusterXL protocol to the options of the FHRP Group
### NetBox version

v3.1.9

### Feature type

### Proposed functionality

Currently the FHRP Group does not support ClusterXL protocol used by Checkpoint, and only available option would be to use Other

Available options in v3.1.9 are seen bellow

image

We propose to add another Protocol to the dropdown list that is ClusterXL

### Use case

Improve the modelling of the FHRP when using Checkpoint firewalls

### Database changes

_No response_

### External dependencies

_No response_

apellini:
Adding ClusterCL Choice to FHRP Group

### DIFF
--- a/netbox/ipam/choices.py
+++ b/netbox/ipam/choices.py
@@ -100,6 +100,7 @@ class IPAddressRoleChoices(ChoiceSet):
     ROLE_HSRP = 'hsrp'
     ROLE_GLBP = 'glbp'
     ROLE_CARP = 'carp'
+    ROLE_CLUSTERXL = 'clusterxl'
 
     CHOICES = (
         (ROLE_LOOPBACK, 'Loopback'),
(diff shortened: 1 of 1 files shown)


jeremystretch:
#8830 proposes adding ClusterXL as an FHRP group protocol, not an IP address role. It needs to be added to `FHRPGroupProtocolChoices`.

Additionally, if we're going to start including proprietary protocols, IMO they should be arranged under a subgroup. Something like this:

```
    CHOICES = (
        ('Standard', (
            (PROTOCOL_VRRP2, 'VRRPv2'),
            (PROTOCOL_VRRP3, 'VRRPv3'),
            (PROTOCOL_HSRP, 'HSRP'),
            (PROTOCOL_GLBP, 'GLBP'),
            (PROTOCOL_CARP, 'CARP'),
        ),
        ('Proprietary', (
            (PROTOCOL_CLUSTERXL, 'ClusterXL'),
        ),
        (PROTOCOL_OTHER, 'Other'),
    )
```

apellini:
So consider that also HSRP and GLBP are Cisco Proprietary.

jeremystretch:
True, for GLBP at least; HSRP [straddles the fence](https://www.rfc-editor.org/rfc/rfc2281.html). I'm happy to put them both under proprietary too.

apellini:
I have done, please check if it is feasible

PR #9028, opened 2022-04-04
Fixes #5479 - Allow setting individual timeouts on scripts and reports

### LINKED ISSUE #5479: Allow customization of job timeout for scripts
### Environment
* Python version: 3.6.6
* NetBox version: 2.10.1

### Proposed Functionality
* Add JOB_TIMEOUT to Script base class
* Add job_timeout=Script.JOB_TIMEOUT to enqueue_job kwarg (per docs, this is popped out on job start)

### Use Case
With the move towards background job processing, the time constraint on scripts is no longer as restrictive.  I feel allowing customization of the job_timeout per-script would be beneficial (if a job times out, the job is still also listed as "Running", this should be handled separately as a bug)

### Database Changes
None

### External Dependencies
None

kkthxbye-code:
### Fixes: #5479

Allows setting the job_timeout Meta attribute on scripts and the job_timeout attribute on reports. I tested UI, API and CLI execution. Scripts ran through the CLI do not support the timeout, as those are not ran through the rqworker for some reason.

I had to change enqueue_job slightly as I found no way to pass the timeout via. the `delay` method added by the `@job` decorator.

### DIFF
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -550,7 +550,7 @@ class JobResult(models.Model):
             user=user,
             job_id=uuid.uuid4()
         )
-        
+
         queue = django_rq.get_queue("default")
         queue.enqueue(func, job_id=str(job_result.job_id), job_result=job_result, **kwargs)
 
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -13,6 +13,7 @@ from django.urls import reverse
 from django.utils import timezone
 from django.utils.formats import date_format
 from rest_framework.utils.encoders import JSONEncoder
+import django_rq
 
 from extras.choices import *
 from extras.constants import *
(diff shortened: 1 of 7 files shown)


kkthxbye-code:
Fixed build failing because of pep8 check (I should get those git hooks working). Just squash before merging.

netbox/extras/models/models.py
```diff
@@ -550,7 +551,8 @@ def enqueue_job(cls, func, name, obj_type, user, *args, **kwargs):
             job_id=uuid.uuid4()
         )
 
-        func.delay(*args, job_id=str(job_result.job_id), job_result=job_result, **kwargs)
+        queue = django_rq.get_queue("default")
+        queue.enqueue(func, job_id=str(job_result.job_id), job_result=job_result, **kwargs)
```
DanSheps, line 555: I don't know enough about django_rq to know if there is a reason to use the job decorator over just enqueuing the job.

I do like that doing this also gives us the ability to schedule jobs down the line.

jeremystretch (approved):
Excellent work, thank you!

netbox/extras/models/models.py
jeremystretch: Should have the same net effect either way.

PR #9067, opened 2022-04-07
Limit amount of nonracked devices displayed

### LINKED ISSUE #8920: Limit the number of non-racked devices that are listed on the Site and Location pages
### NetBox version

netbox-84cb8b95fd-dcfw8 (v3.2.0-beta2)

### Feature type

### Proposed functionality

Only show a small number (10 or so) of non-racked devices on site and location pages. Include a link that would take you to a device search with pre-populated search that limits the devices shown to that site/location and have not been assigned to a rack.

### Use case

Currently, when you view a site or a location in NetBox, it will list all of the non-racked devices that have been assigned to that site or location. If you have a lot of non-racked devices, the page can take a very long time to generate and is very unwieldy to use. By linking to a device search, you can use the paging functionality or further narrow the search parameters.

### Database changes

None

### External dependencies

None

jasonyates:
Fixes #8920

Limits the amount of non-racked devices on Site and Location view to 10 and provides a link to the device list this is pre-filtered to the relevant site or location.

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -342,17 +342,12 @@ class SiteView(generic.ObjectView):
             'device_count',
             cumulative=True
         ).restrict(request.user, 'view').filter(site=instance)
+
         nonracked_devices = Device.objects.filter(
             site=instance,
             position__isnull=True,
             parent_bay__isnull=True
-        ).prefetch_related('device_type__manufacturer').order_by('-pk')[:10]
-
-        total_nonracked_devices_count = Device.objects.filter(
-            site=instance,
-            position__isnull=True,
-            parent_bay__isnull=True
-        ).count()
+        ).prefetch_related('device_type__manufacturer')
 
         asns = ASN.objects.restrict(request.user, 'view').filter(sites=instance)
         asn_count = asns.count()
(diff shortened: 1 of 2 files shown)


netbox/dcim/views.py
```diff
@@ -346,7 +346,13 @@ def get_extra_context(self, request, instance):
             site=instance,
             position__isnull=True,
             parent_bay__isnull=True
-        ).prefetch_related('device_type__manufacturer')
+        ).prefetch_related('device_type__manufacturer').order_by('-pk')[:10]
```
jeremystretch, line 349: To avoid duplicating the queryset, I suggest declaring it once and then referencing that to get both the limited set (`[:10]`) and the total count. You can leave the queryset as-is here:

```python
        nonracked_devices = Device.objects.filter(
            site=instance,
            position__isnull=True,
            parent_bay__isnull=True
        ).prefetch_related('device_type__manufacturer')
```

And under the template context, do:

```
            'nonracked_devices': nonracked_devices.order_by('-pk')[:10],
            'total_nonracked_devices_count': nonracked_devices.count(),
```

Same idea with LocationView below. This helps ensure that any future changes to the queryset aren't missed in the second definition.

jeremystretch (changes requested):

netbox/dcim/views.py
jasonyates: Neat. I didn't realise you could do that. Updated FR pushed.

PR #9112, opened 2022-04-11
Closes #4264: Do not allocate subnet router anycast for IPv6 prefixes

### LINKED ISSUE #4264: Non-pool IPv6 prefixes think the 0th address is available
### Environment
* Python version:  3.6.7
* NetBox version:  2.7.7

### Steps to Reproduce
1. Create an IPv6 prefix (tested /64) that is not configured as a pool
2. Add an address to the prefix (::1 is the simplest test case)
3. Visit the ip-addresses view for this prefix

### Expected Behavior
The count of available IPs 'below' the first allocated IP should not include the 0th address, unless the prefix is configured as a pool (not sure address pools make sense in IPv6 context at all, tbh) or is a /127.

### Observed Behavior
The 0th IP is counted as 'available' in the address list. For example, allocating the ::1 address shows a list entry with '1 IP Available' below the ::1 address. The count will be off-by-one for other cases where there are actually available IPs. See screenshot.

Clicking 'Add an IP address' chooses ::2 as the default as expected, and the `available-ips` API call also correctly does not include the 0th address.

image

sc68cal:
This fixes #4264 by removing the first IPv6 address from the available_ips that is returned, in cases where the RFCs prohibit it

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -507,16 +507,20 @@ class Prefix(GetAvailablePrefixesMixin, NetBoxModel):
             child_ranges.add(iprange.range)
         available_ips = prefix - child_ips - child_ranges
 
-        # IPv6, pool, or IPv4 /31-/32 sets are fully usable
-        if self.family == 6 or self.is_pool or (self.family == 4 and self.prefix.prefixlen >= 31):
+        # IPv6 /127's, pool, or IPv4 /31-/32 sets are fully usable
+        if (self.family == 6 and self.prefix.prefixlen >= 127) or self.is_pool or (self.family == 4 and self.prefix.prefixlen >= 31):
             return available_ips
 
-        # For "normal" IPv4 prefixes, omit first and last addresses
-        available_ips -= netaddr.IPSet([
-            netaddr.IPAddress(self.prefix.first),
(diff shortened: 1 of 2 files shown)


sc68cal:
This just needs an additional test where there are some child prefixes, and then the cover message for this PR and the commit message itself needs cleanup, then this is good to go.

sc68cal:
I think I also need to be proactive and reserve the anycast addresses defined in https://www.rfc-editor.org/rfc/rfc2526

sc68cal:
I also see the following note in netaddr

https://netaddr.readthedocs.io/en/latest/api.html#netaddr.IPNetwork.iter_hosts

> The next release (0.9.0) will contain a backwards incompatible change connected to handling of RFC 6164 IPv6 addresses (/127 and /128 subnets). When iterating IPNetwork and IPNetwork.iter_hosts() the first addresses in the networks will no longer be excluded.

We are currently using 0.8.0 - and this is probably something we're going to run into trouble with?

jeremystretch:
Thanks for working on this!

> We are currently using 0.8.0 - and this is probably something we're going to run into trouble with?

I don't think we use `iter_hosts()` anywhere but I'll keep an eye on it.

PR #9150, opened 2022-04-15
Closes #8894: Add first and last name username user api select

### LINKED ISSUE #8894: Show full usernames when filtering
### NetBox version

v3.1.9

### Feature type

### Proposed functionality

When using the filters to filter user-based (e.g. Journal and Changelog), the full user name should be displayed in brackets, instead of just the login name.

### Use case

Makes filtering in big installations easier

### Database changes

None, all data is already in netbox

### External dependencies

None

minitriga:

### DIFF
--- a/netbox/users/api/nested_serializers.py
+++ b/netbox/users/api/nested_serializers.py
@@ -23,10 +23,16 @@ class NestedGroupSerializer(WritableNestedSerializer):
 
 class NestedUserSerializer(WritableNestedSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='users-api:user-detail')
+    display = serializers.SerializerMethodField(read_only=True)
 
     class Meta:
         model = User
         fields = ['id', 'url', 'display', 'username']
+
+    def get_display(self, obj):
+        if obj.first_name and obj.last_name:
+            return f"{obj.username} ({obj.first_name} {obj.last_name})"
+        return obj.username
 
 
 class NestedTokenSerializer(WritableNestedSerializer):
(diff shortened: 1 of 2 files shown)


netbox/users/api/nested_serializers.py
```diff
@@ -23,11 +23,17 @@ class Meta:
...
     url = serializers.HyperlinkedIdentityField(view_name='users-api:user-detail')
+    display = serializers.SerializerMethodField(read_only=True)
 
     class Meta:
         model = User
         fields = ['id', 'url', 'display', 'username']
 
+    def get_display(self, obj):
+        if obj.first_name and obj.last_name:
+            return f"{obj.username} ({obj.first_name} {obj.last_name})"
```
jeremystretch, line 34: We can use `get_full_name()` on the User instance to return the first and/or last name if either is defined:

```suggestion
        if full_name := obj.get_full_name():
            return f"{obj.username} ({full_name})"
```

netbox/users/api/nested_serializers.py
```diff
@@ -23,11 +23,17 @@ class Meta:
 
 class NestedUserSerializer(WritableNestedSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='users-api:user-detail')
+    display = serializers.SerializerMethodField(read_only=True)
```
jeremystretch, line 26: This is defined on the `BaseModelSerializer` class.

```suggestion
```

netbox/users/api/nested_serializers.py
jeremystretch: `get_display()` should be defined on `UserSerializer` as well.

jeremystretch (changes requested):

minitriga:
Updated. Cheers for the review. 

Just to note that `get_full_name` will return first and last name individually if the other is not set. 

For example if the first_name is not defined it will only show surname. 

```
>>> User.objects.get(pk=2).first_name
''
>>> User.objects.get(pk=2).last_name
'Gittings'
>>> User.objects.get(pk=2).get_full_name()
'Gittings'
```

PR #9175, opened 2022-04-19
Closes #8233: Restrict API key access by source IP

### LINKED ISSUE #8233: Restrict API key usage by source IP

PieterL75:
### Fixes: #8233

Restrict API key usage by source IP


jeremystretch:
Thanks for your work on this @PieterL75. I'm going to see if I can move the source IP resolution logic to a generic utility before merging this into `feature`, but it looks good!

PR #9198, opened 2022-04-22
Update 5-http-server.md

LHBL2003:
Add: "sudo nano /etc/apache2/sites-available/netbox.conf"  and Change description
Reason: To point out more strongly the necessary adjustment of the server name. I had overlooked it in the text.

Add: Open in the browser https://Servername/ or    ...     without the port 8000!
Add: I assumed that you still have to call it with port 8000. As in the pre-test. But it runs with 443

### DIFF
--- a/docs/installation/5-http-server.md
+++ b/docs/installation/5-http-server.md
@@ -56,10 +56,12 @@ Begin by installing Apache:
 sudo apt install -y apache2
 ```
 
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thank you for the proposal, however the proposed changes are unnecessary. The existing documentation is sufficient.

PR #9248, opened 2022-04-28
Fixes #9138 - Prevent searching when pressing enter in Quick Search

### LINKED ISSUE #9138: Searching 'front ports' or 'rear ports' results in an attempt to rename
### NetBox version

v3.2.1

### Python version

3.8

### Steps to Reproduce

1) Go to Demo site
2) click on a device that has front ports (Ex: any of the patch panels)
3) click on the front ports tab
4) attempt to search using the 'quick search'
5) if you type 14 and hit 'enter' you are taken to a 'rename' page where netbox wants to rename 0 ports.

I am recreating this bug since it is still present from my last bug report that was incorrectly marked as a duplicate.
https://github.com/netbox-community/netbox/issues/9113

### Expected Behavior

hitting enter should initiate the quick search and/or do nothing. 
Look at the quick search for 'contacts'. Hitting enter on that quick search does not attempt to rename a contact.

### Observed Behavior

hitting enter on quick search within front ports results in an attempt to rename 0 ports.

kkthxbye-code:
### Fixes: #9138

Move all includes of `inc/table_controls_htmx.html` out of the table forms. This matches the existing setup in `object_list.hml`, which is why the bug only occurred in some cases.

I tested quick search in all the tables I changed, and there seem to be no ill effects of the change, however there could be cases I haven't thought about (I wish UI testing was easier to setup).

### DIFF
--- a/netbox/templates/dcim/device/consoleports.html
+++ b/netbox/templates/dcim/device/consoleports.html
@@ -4,9 +4,10 @@
 {% load static %}
 
 {% block content %}
+  {% include 'inc/table_controls_htmx.html' with table_modal="DeviceConsolePortTable_config" %}
+  
   <form method="post">
     {% csrf_token %}
-    {% include 'inc/table_controls_htmx.html' with table_modal="DeviceConsolePortTable_config" %}
 
     <div class="card">
       <div class="card-body" id="object_list">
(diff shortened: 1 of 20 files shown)


jeremystretch (approved):

jeremystretch:
LGTM, thank you!

PR #9281, opened 2022-05-02
Fixes #9280 - Add option to adopt existing DeviceComponents

### LINKED ISSUE #9280: Add an option to adopt existing DeviceComponents when installing modules
### NetBox version

v3.2.2

### Feature type

### Proposed functionality

Add an option to adopt existing DeviceComponents when installing modules. The option would change the module of all DeviceComponents on the target device with the same name as the component being added by the module.

### Use case

Making it easier to migrate to using modules for devices with manually added DeviceComponents.

### Database changes

None.

### External dependencies

None.

kkthxbye-code:
### Fixes: #9280

This PR adds a new boolean field (adopt_components) to the Module form. The creation logic has been changed to check for and change the module field on any existing components.

~~This is still a draft. I need to verify that the API works, and I haven't checked a few cases.~~ Looking for input for the implementation also.

### DIFF
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -1086,13 +1086,17 @@ class Module(NetBoxModel, ConfigContextModel):
             create_instances = []
             update_instances = []
 
+            # Prefetch installed components
+            installed_components = {
+                component.name: component for component in getattr(self.device, component_attribute).filter(module__isnull=True)
+            }
+
             # Get the template for the module type.
             for template in getattr(self.module_type, templates).all():
                 template_instance = template.instantiate(device=self.device, module=self)
 
                 if adopt_components:
-                    existing_item = getattr(self.device, component_attribute).filter(
-                        module__isnull=True, name=template_instance.name).first()
+                    existing_item = installed_components.get(template_instance.name)
 
                     # Check if there's a component with the same name already
                     if existing_item:
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -1091,7 +1091,8 @@ class Module(NetBoxModel, ConfigContextModel):
                 template_instance = template.instantiate(device=self.device, module=self)
 
                 if adopt_components:
-                    existing_item = getattr(self.device, component_attribute).filter(name=template_instance.name).first()
+                    existing_item = getattr(self.device, component_attribute).filter(
+                        module__isnull=True, name=template_instance.name).first()
 
                     # Check if there's a component with the same name already
                     if existing_item:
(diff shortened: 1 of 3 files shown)


kkthxbye-code:
A potential issue I found when testing. When you add a module with component names that are already present in on another installed module, which of the following solutions would make most sense?

* Steal the components. Just changing the module field works fine, but the components will obviously be removed from the existing module.
* Throw an error (would probably be added to my WIP form validation change here https://github.com/kkthxbye-code/netbox/commit/8f765c114fee67dfd3644ebd7b794f4a07dcf456)

kkthxbye-code:
I think this is a good starting point for the feature. 

I mentioned that I would like to verify that the API works as expected (it does). However, I can see from the commit here a2981870ce6911d577dc2af3d6cd2cf5c952aa14 that disabling replication was specifically only added to the UI. I can bring both replication and adoption into the API also if you want @jeremystretch - not sure if that's another feature request after or if it is even a good idea.

When this is merged I'll create a pull request for #9001

jeremystretch:
Did some testing with this, works great! :+1: 

> When you add a module with component names that are already present in on another installed module, which of the following solutions would make most sense?

The simplest approach may be to simply ignore any existing components which are already assigned to a module. I'll add a note to my review w/more detail.

> I can bring both replication and adoption into the API also if you want @jeremystretch - not sure if that's another feature request after or if it is even a good idea.

Probably worth discussing as a separate issue. For now I think introducing the adoption functionality in the UI only is fine.

netbox/dcim/models/devices.py
```diff
@@ -1065,30 +1065,47 @@ def save(self, *args, **kwargs):
...
+        ]:
+            create_instances = []
+            update_instances = []
+
+            # Get the template for the module type.
+            for template in getattr(self.module_type, templates).all():
+                template_instance = template.instantiate(device=self.device, module=self)
+
+                if adopt_components:
+                    existing_item = getattr(self.device, component_attribute).filter(name=template_instance.name).first()
```
jeremystretch, line 1094: To avoid dealing with duplicate components already assigned to another module, it may be reasonable to ignore them completely.

```suggestion
                    existing_item = getattr(self.device, component_attribute).filter(module__isnull=True, name=template_instance.name).first()
```

Also, a minor point of optimization: We could retrieve all of these components in a single query outside the `for` loop and cache them in a dictionary keyed by name. Then we could just look up each template in the dictionary.

kkthxbye-code:
Added prefetching of installed components.

Ignore components with a module set. This will make module installation fail if replicate components is activated. A better error message can be shown when I finish the error message PR.

kkthxbye-code:
I'll fix the test tomorrow.

jeremystretch (approved):

Issue #9307, opened 2022-05-05
Switch with 48 ports and 100 tagged vLAN per port (No bulk configuration possible.)


LHBL2003:
### NetBox version

v3.2.2

### Feature type

New functionality

### Proposed functionality

I have many switches with 48 ports.
There are over 100 tagged VLANs on very many ports (interfaces).
e.g. VLAN 9, 27, 101,1010, 3000-3100 etc.

Currently I don't see any way how to do a bulk configuration for each interface.

### Use case

In the "802.1Q Switching" section (https://netbox-100/dcim/interfaces/25/edit/)
Only 1 VLAN can be selected in the "Tagged VLANs" field. Then the next VLAN must be selected. etc. 

A mass select is helpful. So that I don't have to click 100 times for VLAN 3000 to 3100.

![2022-05-05_18h44_31](https://user-images.githubusercontent.com/46369917/166973302-e46dc64d-8ff1-4d40-b915-4dd54f5f67d3.png)

LHBL2003:
OK, one point I hadn't thought of. I can select multiple interfaces and thus configure multiple ports for this switch at the same time. However, I have to click 100 times for each switch to start VLAN 3000-3100. I would be happy if there was a multi select for this.

kkthxbye-code:
From a quick glance [slim-select](https://github.com/brianvoe/slim-select) does not support shift selection. A band-aid could be disabling closeOnSelect for multiple select fields, saving at least one click per option (I'm probably gonna create a seperate FR for this regardless).

Not sure of the feasibility of adding shift-select on top of slim-select. @thatmattlove you have any input on this?

CroogQT:
I'm fairly confidant that a similar implementation to  #9309 will work in this case as well. If that pull request is approved I'd be glad to implement this case as well.

LHBL2003:
I don't know anything about APIs yet, but I've already managed a transitional solution. I hope that a solution for everyday use will be implemented.

For the general public: 
Below are a few screenshots showing how to achieve the goal via the API.
It makes my work easier and my colleagues won't behead me. If you work with NetBox for the first time :)

Still a very nice tool :)

Overview
<img width="1916" alt="001_Overview" src="https://user-images.githubusercontent.com/46369917/167574410-e615e10b-f621-4967-8190-70781be69c86.png">

Get Netbox object ID from vLAN
[...]

PR #9309, opened 2022-05-05
Multiselect

### LINKED ISSUE #8704: shift click to select multiple boxes
### NetBox version

v3.1.8

### Feature type

New functionality

### Proposed functionality

I am fairly new to this, but I want a function where you click the first tic box, hold down shift and click the last tic box you need. Then you have selected all the tickboxes in between.

### Use case

When you are selecting multiple devices, a feature where you can select all by holding in shift after ticking one box, and then clicking on the last box that you need would save some time instead of clicking a tick box on every single device.

### Database changes

_No response_

### External dependencies

_No response_

CroogQT:
### Fixes:  #8704 

Allows a user to shift click a checkbox to set the "checked" value of all check boxes between their selection and their previous selection to the new value of their selected checkbox.


LHBL2003:
So does this new feature support my feature suggestion?

https://github.com/netbox-community/netbox/issues/9307

Or is something else meant here?

kkthxbye-code:
@LHBL2003 - This is not related to your issue no.

CroogQT:
Hopefully this will help clarify the intent of #8704 as I understand it.

https://user-images.githubusercontent.com/16785887/167200153-11433cbb-c8f5-4842-a565-69ca2824d525.mov

kkthxbye-code:
Tried it out, works great!

Found a very minor issue:

* Click one or more checkboxes
* Search for something in the Quick Search input
* Remove the search string from Quick Search again
* Shift click any checkbox

Everything from the bottom of the table and to where you clicked will be marked. The reason is that when using the Quick Search the table is replaced by HTMX. As such the state of the previous element will be wrong (doesn't exist anymore) and it will mark everything as it never finds the target element.

Maybe you can clear the state whenever a HTMX table [has been rendered](https://htmx.org/events/#htmx:afterSettle).

CroogQT:
> Tried it out, works great!
> 
> Found a very minor issue:
> 
>     * Click one or more checkboxes
> 
>     * Search for something in the Quick Search input
> 
>     * Remove the search string from Quick Search again
> 
>     * Shift click any checkbox
> 
> 
> Everything from the bottom of the table and to where you clicked will be marked. The reason is that when using the Quick Search the table is replaced by HTMX. As such the state of the previous element will be wrong (doesn't exist anymore) and it will mark everything as it never finds the target element.
> 
> Maybe you can clear the state whenever a HTMX table [has been rendered](https://htmx.org/events/#htmx:afterSettle).

Ah! Great catch and great solution! Will implement shortly. Thanks for the feedback! :)

kkthxbye-code:
@CroogQT - Expect getting some time to work on this? I think it's a great usability improvement.

If you don't have the time I can give it a shot fixing the last small bug.

jeremystretch:
Honestly I'm ok with merging this as-is, even with the bug. Happy to open a separate bug report to track the remaining work.

kkthxbye-code:
More than fine by me.

PR #9383, opened 2022-05-18
Add other power, front/rear port types

### LINKED ISSUE #9098: Add additional options to 'rear ports', 'front ports', and 'power' types
### NetBox version

v3.2.0

### Feature type

### Proposed functionality

Type 'other' is available for interfaces, but no such generalized option exists for 'front port', 'rear port', or 'power' connector types.
A list of every single possible connector type for these connections would not be feasible, so unless users have the ability to add options themselves (which is outside the scope of this request) the next best option is to add an 'other' type to each of those options drop down lists.

### Use case

There are many more connector types available than what is listed for rear port, front port, and power type connections. Adding an 'other' option to these drop down lists would allow users to select something if their particular connector is not listed, rather than not filling in the data at all.

### Database changes

_No response_

### External dependencies

_No response_

bluikko:

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -354,6 +354,7 @@ class PowerPortTypeChoices(ChoiceSet):
     TYPE_UBIQUITI_SMARTPOWER = 'ubiquiti-smartpower'
     # Other
     TYPE_HARDWIRED = 'hardwired'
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks!

PR #9411, opened 2022-05-24
Fixes #9166 - Add the option to make custom fields read-only or hidden in UI

### LINKED ISSUE #9166: Allow hiding custom fields in UI
### NetBox version

v3.2

### Feature type

### Proposed functionality

Provide a 'ReadOnly' checkbox to custom fields.
If a checkbox is readonly, then the user is not able to edit/update the field, when editing an object.

### Use case

Fields that are populated by automation, should not be editable by end users.
(ex: vcenter, prefix utilization, ...)

### Database changes

New Boolean field on the custom fields.

### External dependencies

_No response_

kkthxbye-code:
### Fixes: #9166

Provides 3 visibility options:

* Read/Write (Default)
* Read-only
* Hidden

Read/Write is the current behavior.

Read-only shows the value as normal on objects, but makes the form field disabled on create/edit forms.

Hidden hides the field completely in the UI.

No changes to how custom fields are displayed in the API. There is no form validation if the user were to edit the form to remove the disabled option. I didn't see it as necessary as the visibility is not a security feature.

Please let me know if I missed something.


jeremystretch (approved):
Awesome work, thanks!

MalfuncEddie:
Does an update of a hidden/read-only field also trigger a changelog entry? Since it would be strange that this would trigger it?

I'll try and see If I spin up a test instance from the feature branch to test this

jeremystretch:
Yes, it's still going to log a change to the object, because the object has been changed. The UI display of the field is irrelevant.

PR #9435, opened 2022-05-27
Fixes #9430 - Make sure initial data is passed as array for DynamicModelChoiceFields

### LINKED ISSUE #9430: Allow MultiObjectVar to accept PKs higher than 9 when script is called with a single URL param
### NetBox version

v3.2.3

### Python version

3.9

### Steps to Reproduce

0. Start with a clean, empty installation
1. Create a custom script `testscript.py`:
> ```python
> from extras.scripts import *
> from dcim.models import Device
> 
> class MultiObjectVarTest(Script):
>     devices = MultiObjectVar (model=Device)
> 
>     def run(data, commit):
>         pass
> ```
2. Create a dummy `Manufacturer` and `DeviceType`
3. Fill the database with multiple `Device` objects (more than 10)
4. Confirm that devices have at least picked up PKs 1 thru 10
5. Access `http://netbox/extras/scripts/testscript.MultiObjectVarTest/?devices=1` (1 is a valid, existing device pk) -> Confirm that the `devices` field is properly populated with the `Device` which has pk 1
6. Access `http://netbox/extras/scripts/testscript.MultiObjectVarTest/?devices=10` (10 is also a valid, existing device pk) -> Notice that the `devices` field is now empty

### Expected Behavior

Access `http://netbox/extras/scripts/testscript.MultiObjectVarTest/?devices=10` (10 is a valid, existing device pk) and confirm that the `devices` field is properly populated with the `Device` which has pk 10

### Observed Behavior

The devices field is not populated with the `Device` which has PK 10, but can be searched manually.

I believe that this is due to the widget only prefetching the first few items and not knowing what to do with higher-count objects.

kkthxbye-code:
### Fixes: #9430

When pre-filling a MultiObjectVar for custom script execution passing a single pk above 9 would not work. Passing multiple values like: `?devices=1&device=99` works.

The error occurs in line 95 here:

https://github.com/netbox-community/netbox/blob/2e5a5f71ba4622b2989d892d4215e2aefe59b91c/netbox/utilities/forms/fields/dynamic.py#L88-L100

When passed a two or more digit number in a string, the value seems to get split into a string, so instead of looking up pk 123, it looks up pk 1, 2 and 3.

I'm not sure exactly what's going on here so there might be a better place to fix the issue, please let me know and I'll change it.

### DIFF
--- a/netbox/utilities/forms/fields/dynamic.py
+++ b/netbox/utilities/forms/fields/dynamic.py
@@ -88,7 +88,12 @@ class DynamicModelChoiceMixin:
         # Modify the QuerySet of the field before we return it. Limit choices to any data already bound: Options
         # will be populated on-demand via the APISelect widget.
         data = bound_field.value()
+
         if data:
+            # When the field is multiple choice pass the data as a list if it's not already
+            if isinstance(bound_field.field, DynamicModelMultipleChoiceField) and not type(data) is list:
+                data = [data]
+
             field_name = getattr(self, 'to_field_name') or 'pk'
             filter = self.filter(field_name=field_name)
             try:


jsenecal:
Good catch! I was struggling to find where the issue was occurring. 

Could you add a test for that edge-case for good measure?

PR #9463, opened 2022-06-01
Closes #9368: Change interface naturalize function

### LINKED ISSUE #9368: Change interface naturalize function
### NetBox version

v3.2.2

### Feature type

### Proposed functionality

The interface name is naturalized to make it sortable in a proper way. (1,2,10,11,20,21; and not 1,10,11,2,20,21)
The current implementation is not covering all use cases to sort interfaces properly.

By making the function more generic, it could cover more use cases:
```
output = ''
parts = re.split(r'(\d+)',value.lower())
for part in parts:
    if part.isdigit():
        output += part.rjust(4, '0')
    elif part not in ['.',':','/']:
        output += part
```
It splits the interfacename into parts with and without numbers. the numbers are pre-padded with zero's and the chars [dot,colon,slash] are removed from between the numbers.
There might be corner cases with mixed number schema's on one device, but it will keep them grouped.

Unless there was a reason why the slot/subslot/position/.. were prepended to the interfacetype, rather then post ?

### Use case

On Arista we have interfaces that are split in lanes, without slot/pos/subpos identifiers.
```
Ethernet1
Ethernet2
Ethernet3
Ethernet48/1
Ethernet48/2
Ethernet48/3
Ethernet48/4
Ethernet49/1
Ethernet49/2
Ethernet49/3
Ethernet49/4
```

The current rules sorts them as
```
Ethernet48/1
Ethernet48/2
Ethernet48/3
Ethernet48/4
Ethernet49/1
Ethernet49/2
Ethernet49/3
Ethernet49/4
Ethernet1
Ethernet2
Ethernet3

```

because the /1 is prepended to the interfacetype, while with non-slashed interfaces, the prepend is 9999

https://github.com/netbox-community/netbox/blob/a2203da1c6fa674a55d3071a287094d70cb1c502/netbox/utilities/ordering.py#L46

### Database changes

All _name fields of all interfaces need to be re-calculated.

### External dependencies

_No response_

PieterL75:
### Fixes: #9368 Change interface naturalize function

This PR will change the naturalize_interface function to provide a more general but still adoptable way of sorting network interfaces.


PieterL75:
I checked quite some FR's and implemented a logic that would match for all requests.
Few exceptions : 
FastEthernet1, FastEthernet2, FastEthernet10,... will be sorted before GigaEthernet1/1, etc and not after.
I see no reason to place those after the moduled interfaces. If FastEthernet0 is a mgmt_only interface, then the function will place that at the end

This case is also not possible, as it is either by name, either by slot/module/port
```
Fa1/0/1  
Fa1/0/2
Gi1/0/1
Gi1/0/2
```

jeremystretch:
> I see no reason to place those after the moduled interfaces.

They're ordered the way they are for a reason. As I stated [here](https://github.com/netbox-community/netbox/issues/9368#issuecomment-1135977690), we cannot accept any changes which modify the established orderings. I'm not going to re-open those discussions.

PR #9498, opened 2022-06-07
Adds replication and adoption for module import

### LINKED ISSUE #9361: Add replication and association booleans to Module bulk import
### NetBox version

v3.2.3

### Feature type

### Proposed functionality

It would be nice to extend the replication and association checkboxes available in the UI to also be available when doing a module bulk import.

### Use case

Adding a single module at a time is tedious and we are looking to utilize more bulk imports, but have now run into the limitation that device components cannot be automatically associated with that method and would still require manual intervention after the fact to do those associations. Possibly the larger at this point though is that there is no way to disable the component replication, so trying to do an import like this generates an exception for duplicate keys.

### Database changes

None

### External dependencies

None

sleepinggenius2:
### Fixes: #9361

Makes the replicate_components and adopt_components fields available in the UI to also be available when doing a module bulk import.

### DIFF
--- a/netbox/dcim/tests/test_views.py
+++ b/netbox/dcim/tests/test_views.py
@@ -1849,6 +1849,53 @@ class ModuleTestCase(
         self.assertEqual(Interface.objects.filter(device=device).count(), 5)
 
     @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])
+    def test_module_bulk_replication(self):
+        self.add_permissions('dcim.add_module')
+
+        # Add 5 InterfaceTemplates to a ModuleType
+        module_type = ModuleType.objects.first()
+        interface_templates = [
+            InterfaceTemplate(module_type=module_type, name=f'Interface {i}') for i in range(1, 6)
+        ]
+        InterfaceTemplate.objects.bulk_create(interface_templates)
+
+        form_data = self.form_data.copy()
+        device = Device.objects.get(pk=form_data['device'])
+
+        # Create a module *without* replicating components
+        module_bay = ModuleBay.objects.get(pk=form_data['module_bay'])
+        csv_data = [
+            "device,module_bay,module_type,replicate_components",
+            f"{device.name},{module_bay.name},{module_type.model},false"
+        ]
+        request = {
+            'path': self._get_url('import'),
+            'data': {
+                'csv': '\n'.join(csv_data),
+            }
+        }
+
+        initial_count = self._get_queryset().count()
+        self.assertHttpStatus(self.client.post(**request), 200)
+        self.assertEqual(self._get_queryset().count(), initial_count + len(csv_data) - 1)
+        self.assertEqual(Interface.objects.filter(device=device).count(), 0)
+
+        # Create a second module (in the next bay) with replicated components
+        module_bay = ModuleBay.objects.get(pk=(form_data['module_bay'] + 1))
+        csv_data[1] = f"{device.name},{module_bay.name},{module_type.model},true"
+        request = {
+            'path': self._get_url('import'),
+            'data': {
+                'csv': '\n'.join(csv_data),
+            }
+        }
+
+        initial_count = self._get_queryset().count()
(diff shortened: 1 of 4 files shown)


kkthxbye-code:
Thank you for the PR! I'm not a big fan of having that much duplicate code for the clean method.

@jeremystretch is there any clean way to share the `clean` method between the model form and the CSV form? I guess we could create a mixin with a `clean_module` method, just not sure where to put it.

The code functioned as expected when testing.

jeremystretch:
I don't have a strong preference for an approach, but we definitely can't just replicate the existing code. Maybe a utility function within the `dcim.forms` module?

sleepinggenius2:
Updated PR to move clean logic to a common class that both forms can inherit from. This works in my testing, but I'm not very familiar with all of this, so please let me know if this isn't a good solution.

sleepinggenius2:
As we're getting closer to the v3.3 release, I just wanted to follow up on this to see if there's anything else I need to do to be able to get this merged.

sleepinggenius2:
@kkthxbye-code or @jeremystretch, as there have been a number of point releases now since this PR was submitted and we continue to get closer to a v3.3 release, I just wanted to follow up again to see if there is anything I still need to do to get this merged. Thank you.

jeremystretch (dismissed):
@sleepinggenius2 I think the approach looks good, thanks for the revision. Could you extend the import view test to verify the function of the two new fields please?

sleepinggenius2:
> @sleepinggenius2 I think the approach looks good, thanks for the revision. Could you extend the import view test to verify the function of the two new fields please?

I assume that this will need to go in the `ModuleTestCase` class of `netbox/dcim/tests/test_views.py`? I'm not particularly familiar with testing in Python, but I think I have enough information to give it a shot.

sleepinggenius2:
@jeremystretch, I requested a review a month ago after I made the changes that you asked for, but it is still showing *Changes requested* and there has been no movement, so I apologize if there was something I missed in that process.

sleepinggenius2:
@jeremystretch or @kkthxbye-code, what do I need to do to get this merged or at least kept from being closed?

kkthxbye-code (approved):
Sorry this fell through the cracks @sleepinggenius2

I checked out the PR and rebased on the development and everything works exactly as it should, both for CSV import and through the UI for all combinations of replication/adoption. So this looks good to me, can we merge this @jeremystretch ?

netbox/dcim/tests/test_views.py
```diff
@@ -1907,6 +1954,49 @@ def test_module_component_adoption(self):
...
+        device = Device.objects.get(pk=form_data['device'])
+
+        # Create an interface to be adopted
+        interface = Interface(device=device, name=interface_name, type=InterfaceTypeChoices.TYPE_10GE_FIXED)
+        interface.save()
+
+        # Ensure that interface is created with no module
+        self.assertIsNone(interface.module)
+
+        # Create a module with adopted components
```
kkthxbye-code, line 1977: One small thing, this testcase actually fails on my PC. It's trying to install in the first modulebay of the device, which is already occupied from the setup. If I change to another bay it passes. Not sure if it's something with my setup or from the rebase.

jeremystretch: Same. Maybe related to 9ef24d3f43022f61044756741f40e095e9290bc1? In any case, changing it to an empty bay appears to fix the test.

jeremystretch: Pushed a fix for this in [2a55c29](https://github.com/netbox-community/netbox/pull/9498/commits/2a55c29a9a09fe20a61dcc9964e298769ada3e72).

sleepinggenius2:
I appreciate your all's assistance with this.

jeremystretch:
Not sure what's going on with CI today. Tests pass locally, though I'm getting failure notifications from GitHub. Nothing's showing up for the PR currently.

Going to go ahead and merge this into `develop` and deal with any remaining issues there. Thanks for your work on this @sleepinggenius2 and sorry for the delays!

Issue #8230, opened 2022-01-04
Custom script does not correctly escape log output for HTML display


candlerb:
### NetBox version

v3.1.4

### Python version

3.8

### Steps to Reproduce

Write a custom script which uses `log_xxx` and outputs `<` characters. e.g.

```
from extras.scripts import Script, StringVar

class Foo(Script):
    abc = StringVar(required=False)
    name = "Foo"

    def run(self, data, commit):
        self.log_info(f'Type of abc is {type(data["abc"])}')
```

### Expected Behavior

Script to log and display:

```
Type of abc is <class 'str'>
```

### Observed Behavior

![image](https://user-images.githubusercontent.com/44789/148116937-98a05be3-2db3-4c90-8ca8-46863e9bd990.png)

Oddly, *something* is going on.  if I change the script to

```
        self.log_info(f'Type of abc is <b>grr</b>')
```

then I get "Type of abc is grr" (without the bold).  Maybe there's some dubious regexp-type sanitisation instead of proper HTML escaping going on.

jeremystretch:
The HTML tags (or anything that looks like one) are being stripped by the `render_markdown` filter to protect against XSS. I'm not sure there's any action to take here as this should be considered expected behavior.

If you want to print angle brackets, you can wrap the line (or value) using `django.utils.html.escape`:

```
self.log_info(escape(f'Type of abc is {type(data["abc"])}'))
```

candlerb:
> this should be considered expected behavior

I see it's [documented](https://netbox.readthedocs.io/en/stable/customization/custom-scripts/#logging) that "Markdown rendering is supported for log messages", although I can't see that it says that html tags are stripped.  I can't think of any other system which supports markdown in log messages, or what the intended use is - maybe for `**bold**` text?   ISTM there's plenty of risk of special characters in log data being mistaken for markdown-ese, and I think it would be much safer not to do markdown processing.

However, even if there's a markdown pipeline, another option would be to escape html tags _prior_ to markdown rendering?  Obviously this protects against XSS in the same way.

Otherwise, I suppose I can can emit backticks around each log message (ugh).

DanSheps:
I think there are plenty of valid uses for markdown however.  Lets say your script creates a bunch of interfaces and you want to drop a link to them in the logs.  Perfectly valid.

What about a script that pulls data from some external interface you want to link to?

[...]

Issue #8944, opened 2022-03-23
Sequential relative URLs with colons in the text not rendering correctly using the render_markdown templatetag


drygdryg:
### NetBox version

v3.1.9

### Python version

3.10

### Steps to Reproduce

1. Run NetBox shell and import the `render_markdown` templatetag:
```
$ /opt/netbox/venv/bin/python /opt/netbox/netbox/manage.py nbshell
### NetBox interactive shell (ubuntuserver)
### Python 3.10.2+ | Django 3.2.12 | NetBox 3.1.9
### lsmodels() will show available models. Use help(<model>) for more info.
>>> from utilities.templatetags.helpers import render_markdown
```
2. Try to render test case URLs:
```python
>>> render_markdown('[TEST](/dcim/devices/1/)')
'<div class="rendered-markdown"><p><a href="/dcim/devices/1/">TEST</a></p></div>'  # Correct
>>> render_markdown('[TEST:](/dcim/devices/1/)')
'<div class="rendered-markdown"><p><a href="/dcim/devices/1/">TEST:</a></p></div>'  # Correct
>>> render_markdown('[TEST](/dcim/devices/1/) [TEST](/dcim/devices/1/)')
'<div class="rendered-markdown"><p><a href="/dcim/devices/1/">TEST</a> <a href="/dcim/devices/1/">TEST</a></p></div>'  # Correct
>>> render_markdown('[TEST](/dcim/devices/1/) [TEST:](/dcim/devices/1/)')
'<div class="rendered-markdown"><p>[TEST](](/dcim/devices/1/)</p></div>'  # Incorrect
>>> render_markdown('[TEST](https://netbox.site/dcim/devices/1/) [TEST:](https://netbox.site/dcim/devices/1/)')
'<div class="rendered-markdown"><p><a href="https://netbox.site/dcim/devices/1/">TEST</a> <a href="https://netbox.site/dcim/devices/1/">TEST:</a></p></div>'  # Correct
```

### Expected Behavior

The string 
```
[TEST](/dcim/devices/1/) [TEST:](/dcim/devices/1/)
```
should be rendered as
```
<div class="rendered-markdown"><p><a href="/dcim/devices/1/">TEST</a> <a href="/dcim/devices/1/">TEST:</a></p></div>
```

### Observed Behavior

The string 
```
[TEST](/dcim/devices/1/) [TEST:](/dcim/devices/1/)
```
was rendered as
```
[...]

PR #9522, opened 2022-06-11
Fixes #9108 & #8944 - Sanitize HTML after rendering markdown

### LINKED ISSUE #9108: Markdown syntax not working correctly within comment fields
### NetBox version

v3.2.0

### Python version

3.8

### Steps to Reproduce

1. Navigate to any object that has a 'comments' field.
2. Attempt to create a URL using angle brackets < >.
3. Click save.

### Expected Behavior

Netbox should turn the text within the angle brackets into a clickable URL.

### Observed Behavior

Text contained within the angle brackets completely disappears after clicking save.

kkthxbye-code:
### Fixes: #9108 and #8944

This PR removes the stripping of all HTML tags and the workarounds for sanitizing markdown links. Instead the HTML output of python-markdown is sanitized using [bleach](https://github.com/mozilla/bleach). The result is a more correct handling of markdown (HTML is explicitly allowed in the markdown spec) while still preventing user defined HTML that might result in XSS, by sanitizing with a whitelist approach.

From a little unscientific testing of a journal page with extreme markdown usage, the performance penalty is very slight (2-5% of total page load time).

I targeted this at the feature branch as a new dependency is added, not sure if that is correct. If it's not I'll change the target branch.

### DIFF
--- a/base_requirements.txt
+++ b/base_requirements.txt
@@ -125,3 +125,7 @@ tablib
 # Timezone data (required by django-timezone-field on Python 3.9+)
 # https://github.com/python/tzdata
 tzdata
+
+# HTML sanitizer
+# https://github.com/mozilla/bleach
+bleach
--- a/netbox/utilities/templatetags/builtins/filters.py
+++ b/netbox/utilities/templatetags/builtins/filters.py
@@ -11,7 +11,7 @@ from markdown import markdown
 
 from netbox.config import get_config
 from utilities.markdown import StrikethroughExtension
-from utilities.utils import foreground_color
+from utilities.utils import clean_html, foreground_color
 
 register = template.Library()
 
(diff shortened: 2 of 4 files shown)


kkthxbye-code:
Should also fix #8230 which was auto closed.

jeremystretch:
Thanks for digging into this @kkthxbye-code! I had a feeling the eventual fallback to bleach was inevitable.

> I targeted this at the feature branch as a new dependency is added, not sure if that is correct. If it's not I'll change the target branch.

IMO it's fine to push this in the next patch release, since it's not a breaking change. I'll update the PR.

kkthxbye-code:
I think it should be good now @jeremystretch - I'll never learn rebasing properly, always seems like a huge mess.

jeremystretch:
git

PR #9527, opened 2022-06-13
Partially fixes #9374 - Implement a custom paginator for DeviceViewSet to improve config_context load times

### LINKED ISSUE #9374: REST response time does not scale properly with `limit` param while including config_context
### NetBox version

v3.2.2

### Python version

3.9

### Steps to Reproduce

1. GET devices from `/api/dcim/devices`, including config_context, with a resulting `count` field less than 100, and `limit=100` (for example, use a query parameter that would result in a low number of matches)
2. GET devices from `/api/dcim/devices` including config_context, with a resulting `count` field greater than 100,000, and `limit=100` (for example, use a query parameter that would result in a high number of matches)

### Expected Behavior

1. Similar API response times from either query, scaling with `limit`

### Observed Behavior

1. When the `exclude=config_context` parameter is included in a query, there is a similar response time between queries that return a large dataset, and queries that return a small dataset
2. When `exclude=config_context` is omitted (i.e. config context is fetched), queries that return a large dataset increase in response time dramatically vs. smaller queries, regardless of `limit`

kkthxbye-code:
### Fixes: #9374

Running count on the annotated query for loading config_context is slow. The custom paginator removes the annotation before getting the count.

Testing with 35k devices loading the devices list via. the API improved load times from ~3200 ms to ~1600 ms.

This doesn't fix the underlying issue in that the query config_context annotation is slow in general and scales with the number of devices, regardless of number of records fetched.

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -19,6 +19,7 @@ from ipam.models import Prefix, VLAN
 from netbox.api.authentication import IsAuthenticatedOrLoginNotRequired
 from netbox.api.exceptions import ServiceUnavailable
 from netbox.api.metadata import ContentTypeMetadata
+from netbox.api.pagination import StripCountAnnotationsPaginator
 from netbox.api.viewsets import NetBoxModelViewSet
 from netbox.config import get_config
 from utilities.api import get_serializer_for_model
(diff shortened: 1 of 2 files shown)


jeremystretch:
> This doesn't fix the underlying issue in that the query config_context annotation is slow in general and scales with the number of devices, regardless of number of records fetched.

I don't know that we can consider this a _bug_ frankly, since it's more of an optimization issue. The fix your provided here works great to address the most obvious issue, and IMO resolves #9374.

We should probably look into implementing this more widely across the application, as there are instances where other queries could be sped up by shedding annotations from their `COUNT()` statements. I'll open a separate housekeeping item to pursue that once this has been tested more widely.

PR #9543, opened 2022-06-15
Replace None in templates with placeholder filter

### LINKED ISSUE #9537: Inconsistent "lack of value" value on details view
### NetBox version

v3.2.4

### Python version

3.10

### Steps to Reproduce

1. Create any object
2. Navigate to the details page for the created object
3. Observe rows which lack a value (Represented as `–` or `None`)

### Expected Behavior

All rows which lacks a value should use the same displayed value `–`.

### Observed Behavior

Both `–` and `None` are used to represent "lack of value".

hagbarddenstore:
### Fixes: #9537

To be consistent, all uses of &mdash; or None is replaced with the
placeholder filter.


jeremystretch:
Thanks for taking care of this!

PR #9547, opened 2022-06-16
Fixes #9533 - Move markdown documentation to docs

### LINKED ISSUE #9533: Provide Markdown help with links to local documentation pages
### NetBox version

v3.2.4

### Feature type

New functionality

### Proposed functionality

Currently netbox supports a documentation package as part of the main release due to https://github.com/netbox-community/netbox/issues/6328

I propose to change the Markdown assistance available in some text areas ( for example in comments fields) that is currently going to "https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet" to another URL as part of the offline documentation package 
image

I propose that a new section in the documentation package is created, based in the github url above, and use the corresponding link within this assistance fields. 

The final url could be something like, if this section is placed under references. 
https://netboxfqdn/static/docs/reference/markdown/

### Use case

The following use cases are applicable:

    Provide the correct documentation that is always related to the specific version being used, instead of the online version that refers the latest version.
    Provide access to the documentation to system installed in a isolated management environment that do not have internet access.

### Database changes

none

### External dependencies

none

kkthxbye-code:
### Fixes: #9533

Removed unsupported stuff from the docs and moved to internal docs. It is based on the linked cheat sheet.

Note: Not testet properly yet as I don't have a setup that serves static files properly so docs work. I'll get around to it tomorrow probably.

### DIFF
--- a/docs/reference/markdown.md
+++ b/docs/reference/markdown.md
@@ -1,0 +1,353 @@
+---
+hide:
+  - toc
+---
+
+# Markdown
+
+NetBox supports markdown rendering for certain text fields.
+
+## Syntax
+
+##### Table of Contents  
+[Headers](#headers)  
+[Emphasis](#emphasis)  
+[Lists](#lists)  
+[Links](#links)  
+[Images](#images)  
+[Code Blocks](#code)  
+[Tables](#tables)  
+[Blockquotes](#blockquotes)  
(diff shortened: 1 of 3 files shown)


jeremystretch:
I checked this out on a proper production setup w/static docs, works great. Thanks!

PR #9554, opened 2022-06-17
Closes #9536: Last Used field on API tokens

### LINKED ISSUE #9536: Last Used field on API tokens
### NetBox version

v3.2.3

### Feature type

### Proposed functionality

Add a 'Last used' property to the API token

### Use case

I want to know if a token is actively being used.

### Database changes

Add 'Last used' field (datetime) to 'users_token'

### External dependencies

_No response_

PieterL75:
### Fixes: #9536: Last Used field on API tokens

Added a new field to the Token Model : last_used (DateTime)
The field is updated each time a token is authenticated (max 1x per minute to reduce db write load)


jeremystretch:
Since this introduces a new field, could you rebase this on the `feature` branch for potential inclusion in v3.3.

PieterL75:
I noticed that this PR and #9175 both create a 0003 migration file.. 
Little point of attention when merging them both

jeremystretch:
No worries there; I'll take care of renumbering one or the other as they get merged. Thanks for calling that out.

netbox/netbox/api/authentication.py
```diff
@@ -18,6 +22,17 @@ def authenticate_credentials(self, key):
         except model.DoesNotExist:
             raise exceptions.AuthenticationFailed("Invalid token")
 
+        # Update last used, but only once a minute. This reduces the write load on the db
+        if not token.last_used or (timezone.now() - token.last_used).total_seconds() > 60:
+            # If maintenance mode is enabled, assume the database is read-only, and disable updating the token's
+            # last_used time upon authentication.
+            if get_config().MAINTENANCE_MODE:
+                logger = logging.getLogger('netbox.auth.login')
+                logger.warning("Maintenance mode enabled: disabling update of token's last used timestamp")
```
jeremystretch, line 31: This seems like it might result in a lot of log spam. Maybe use debug level instead of warning?

netbox/netbox/api/authentication.py
```diff
@@ -18,6 +22,17 @@ def authenticate_credentials(self, key):
...
+            else:
+                token.last_used = timezone.now()
+                token.save()
```
jeremystretch, line 34: It might be a bit faster to make a raw `update()` query to update the time.

```suggestion
                Token.objects.filter(pk=token.pk).update(last_used=timezone.now())
```

jeremystretch (approved):
Some very minor notes but I like the approach! Per your comment earlier we'll need to merge these changes with the work from #8233, but I'm happy to take care of that.

PR #9562, opened 2022-06-19
Closes: #9525 - Add split button functionality to table rows

### LINKED ISSUE #9525: Faster editing with 1 click on Edit
### NetBox version

v3.2.4

### Feature type

### Proposed functionality

Bulk editing is not always helpful, e.g. when adjusting the VLAN name. Then you have to click on edit. 

I then do this as follows:

- Click on the edit symbol
- Scroll wheel Mouse click on Edit (to open the entry in a new tab.)
- X repetitions on the other lines.
- Edit all open tabs.

The work would be more pleasant if you didn't have to click 2 times each time.
This would work if the Edit button is split:

- Left side directly to edit.
- Right side, selection (Edit, Delete, Changelog etc.)

image

### Use case

Faster and more pleasant work. By easier clicking for editing.

### Database changes

_No response_

### External dependencies

_No response_

huntabyte:
### Fixes: #9525 

Utilizing the _**extra_buttons**_ argument within the _**ActionsColumn**_ class, I added an additional button that takes the user directly to the edit page, removing a click from the workflow.

The edit button was added through the **_actions_** of the **_NetBoxTable_**. The existing **_Edit_** link inside the dropdown is retained but can be removed if requested by maintainers/users.

netbox-1-click-edit-arrow

### DIFF
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -175,6 +175,7 @@ class ActionsColumn(tables.Column):
 
     :param actions: The ordered list of dropdown menu items to include
     :param extra_buttons: A Django template string which renders additional buttons preceding the actions dropdown
+    :param split_actions: When True, converts the actions dropdown menu into a split button with first action as the direct button link and icon (default: True)
     """
     attrs = {'td': {'class': 'text-end text-nowrap noprint'}}
     empty_values = ()
@@ -184,10 +185,11 @@ class ActionsColumn(tables.Column):
         'changelog': ActionsItem('Changelog', 'history'),
     }
 
-    def __init__(self, *args, actions=('edit', 'delete', 'changelog'), extra_buttons='', **kwargs):
+    def __init__(self, *args, actions=('edit', 'delete', 'changelog'), extra_buttons='', split_actions=True, **kwargs):
         super().__init__(*args, **kwargs)
 
         self.extra_buttons = extra_buttons
+        self.split_actions = split_actions
 
         # Determine which actions to enable
         self.actions = {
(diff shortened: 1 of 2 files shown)


jeremystretch:
Per my comment [here](https://github.com/netbox-community/netbox/issues/9525#issuecomment-1159765550), the intent of FR #9525 is to split the dropdown button into separate link and dropdown components as in [this example](https://getbootstrap.com/docs/5.0/components/dropdowns/#split-button), rather than to add another button (which largely detracts from the purpose of the dropdown menu).

huntabyte:
I will implement the changes accordingly. Thanks for the feedback!

huntabyte:
Before pushing the updated changes @jeremystretch - Do you feel it is necessary to retain the first action in the dropdown menu along with it being the direct link, or should it be removed from the dropdown and only accessible through the direct link?

jeremystretch:
@huntabyte the first action should only appear as the direct link and not in the dropdown menu.

huntabyte:
Added split button functionality to the `ActionsColumn.render()` method. 

Loom_BuK03h9w6m
&nbsp;<br>
The use of a split button is controlled by the argument `split_actions` which defaults to `True`.

```python
def __init__(self, *args, actions=('edit', 'delete', 'changelog'), extra_buttons='', split_actions=True, **kwargs):

        self.extra_buttons = extra_buttons
        self.split_actions = split_actions

        # Determine which actions to enable
        self.actions = {
            name: self.actions[name] for name in actions
        }

```
&nbsp;<br>
The first modification that was made is within the for loop. In order to gain access to the index which will be used to access the first action item, I enumerated the object to gain access to the counter of each item.

```python
for idx, (action, attrs) in enumerate(self.actions.items()):
```

&nbsp;<br>
Inside the loop, the number of action items is checked, if there is only one action, a standard button will be rendered with a direct link to that action's view. This reduces unnecessarily rendering dropdowns or split buttons for a single action. This was implemented to better support the single-action objects, such as **VLAN Device Interfaces**.

```python
# If only a single action exists, render a regular button
if len(self.actions.items()) == 1:
    html += (
        f'<a class="btn btn-sm btn-secondary" href="{url}{url_appendix}" type="button">'
        f'<i class="mdi mdi-{attrs.icon}"></i></a>'
    )

```

Loom_fHJjCdm3ld

&nbsp;<br>
If there is more than one action, the method checks to see if `split_actions` is `True` and that the `idx` is `0` *(idx being the index of the current action item).* If the conditions are met, the direct link button is appended to the `html` variable with the appropriate icon and URL. The `dropdown` class must exist alongside the `btn-group` class to prevent an overflow issue.


```python
# Creates split button for the first action with direct link and icon
elif self.split_actions and idx == 0:
    html += (
        f'<span class="btn-group dropdown">'
        f'<a class="btn btn-sm btn-secondary" href="{url}{url_appendix}" type="button">'
        f'<i class="mdi mdi-{attrs.icon}"></i></a>'
    )
```

&nbsp;<br>
Finally, for all other actions, a standard list item with the appropriate URL, icon, and title is appended to the `links` list.

```python
# Creates standard action dropdown menu items
else:
    links.append(
        f'<li><a class="dropdown-item" href="{url}{url_appendix}">'
        f'<i class="mdi mdi-{attrs.icon}"></i> {attrs.title}</a></li>'
    )
```

&nbsp;<br>
Lastly, I updated the following conditional which includes variables that are empty if `self.split_actions is True` else contain the necessary HTML to render the regular dropdown menu. I added these variables in an attempt to reduce the HTML duplication - if there is a better way of doing so, I am all ears.

```python
# Create the actions dropdown menu
if links:
    dropdown_icon = '' if self.split_actions else '<i class="mdi mdi-wrench"></i>'
    dropdown_class = '' if self.split_actions else '<span class="dropdown">'
    html += (
        f'{dropdown_class}'
        f'<a class="btn btn-sm btn-secondary dropdown-toggle" type="button" data-bs-toggle="dropdown">'
        f'{dropdown_icon}'
        f'<span class="visually-hidden">Toggle Dropdown</span></a>'
        f'<ul class="dropdown-menu">{"".join(links)}</ul></span>'
    )
```

jeremystretch:
Excellent work, @huntabyte! Thanks for this!

PR #9576, opened 2022-06-21
Closes #7702: Add Power feed defaults to user configurations

### LINKED ISSUE #7702: Allow configuration of Netbox powerfeed defaults
### NetBox version

v3.0.7

### Feature type

### Proposed functionality

At the moment the POWERFEED_VOLTAGE_DEFAULT, POWERFEED_AMPERAGE_DEFAULT and POWERFEED_MAX_UTILIZATION_DEFAULT are picked up from dcim/constants.py and set for typical US datacentre values.

It would be good to adapt that to allow them to pick up either the defaults in constants.py or defaults supplied from an environment variable or config file instead.

### Use case

This would remove a UX "papercut" for users outside of the US.

### Database changes

_No response_

### External dependencies

_No response_

huntabyte:
### Fixes: #7702

Implemented the ability for users to override the default power feed values via user-configurable preferences.

I ran into a bit of a fork in the road when adding this functionality. Initially, I was replacing the model defaults with config items, for example:

```python
class PowerFeed(NetBoxModel, PathEndpoint, LinkTermination):
    """
    An electrical circuit delivered from a PowerPanel.
    """
    voltage = models.SmallIntegerField(
        default=ConfigItem('POWERFEED_DEFAULT_VOLTAGE')
        validators=[ExclusionValidator([0])]
    )
    amperage = models.PositiveSmallIntegerField(
        validators=[MinValueValidator(1)],
        default=ConfigItem('POWERFEED_DEFAULT_AMPERAGE')
    )
    max_utilization = models.PositiveSmallIntegerField(
        validators=[MinValueValidator(1), MaxValueValidator(100)],
        default=ConfigItem('POWERFEED_DEFAULT_MAX_UTILIZATION')
        help_text="Maximum permissible draw (percentage)"
```

Django's documentation regarding migrations does state the following:

> Django never sets database defaults and always applies them in the Django ORM code.

But after investing the rest of the codebase, I didn't see user-defined values as defaults for any of the other models, so I assumed the standard is to let the view handle it:

```python
class PowerFeedCreateView(generic.ObjectEditView):
    queryset = PowerFeed.objects.all()
    form = forms.PowerFeedForm

    def alter_object(self, obj, request, args, kwargs):
        obj.voltage = ConfigItem('POWERFEED_DEFAULT_VOLTAGE')
        obj.amperage = ConfigItem('POWERFEED_DEFAULT_AMPERAGE')
        obj.max_utilization = ConfigItem('POWERFEED_DEFAULT_MAX_UTILIZATION')
        return obj
```

However, if the maintainers decide otherwise or see no reason not to allow users to define the model defaults, I can make that adjustment.

### DIFF
--- a/netbox/dcim/models/power.py
+++ b/netbox/dcim/models/power.py
@@ -106,7 +106,7 @@ class PowerFeed(NetBoxModel, PathEndpoint, LinkTermination):
         default=PowerFeedPhaseChoices.PHASE_SINGLE
     )
     voltage = models.SmallIntegerField(
-        default=ConfigItem('POWERFEED_DEFAULT_VOLTAGE')
+        default=ConfigItem('POWERFEED_DEFAULT_VOLTAGE'),
         validators=[ExclusionValidator([0])]
     )
     amperage = models.PositiveSmallIntegerField(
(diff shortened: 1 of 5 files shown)


huntabyte:
Yes, I will move these over to the model fields.

PR #9600, opened 2022-06-24
Fixes #8854: Remote auth default groups added to new remote auth users

### LINKED ISSUE #8854: REMOTE_AUTH_DEFAULT_GROUPS not adding new users to group
### NetBox version

v3.1.9

### Python version

3.8

### Steps to Reproduce

1. Create new group called 'Users' in Netbox admin
2. Add settings to configuration.py for remote auth and default group:
```
# Remote authentication support
REMOTE_AUTH_ENABLED = True

REMOTE_AUTH_BACKEND = 'social_core.backends.azuread_tenant.AzureADTenantOAuth2'

REMOTE_AUTH_DEFAULT_GROUPS = ['Users']

REMOTE_AUTH_GROUP_SYNC_ENABLED = False

SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_RESOURCE  = '[redacted]'
SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_KEY       = '[redacted]'
SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_SECRET    = '[redacted]'
SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_TENANT_ID = '[redacted]'
```

3. New user authenticates using the Azure auth provider into Netbox
4. User is added to Netbox, but user is not added to 'Users' group

### Expected Behavior

New users should be added to the group(s) indicted by the REMOTE_AUTH_DEFAULT_GROUPS setting.

### Observed Behavior

User is not added to group on sign-on.

huntabyte:
### Fixes: #8854 

Added a custom pipeline handler function per python-social-auth [docs](https://python-social-auth.readthedocs.io/en/latest/configuration/django.html)

This pipeline handler assigns the default group(s) to remote auth users. Successfully tested with Microsoft Azure AD, however, there is nothing particular related to Azure AD in this configuration.

### DIFF
--- a/netbox/netbox/authentication.py
+++ b/netbox/netbox/authentication.py
@@ -1,5 +1,7 @@
 import logging
 from collections import defaultdict
+import requests
+from rich import print
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


jeremystretch:
Awesome work, thanks for tackling this!

HolySephi:
This change removes the possibility to define my own SOCIAL_AUTH_PIPELINE in configuration.py and thus removes the function to assign roles from auth claims as groups via custom_pipeline. Removing thhis static SOCIAL_AUTH_PIPELINE from settings.py allows my own pipeline to run again.
Bug: SOCIAL_AUTH_PIPELINE from configuration.py does not overwrite this SOCIAL_AUTH_PIPELINE from settings.py


if you are interested - this is the custom_pipeline to set auth claim roles to groups:

`from django.contrib.auth.models import Group

class AuthFailed(Exception):
    pass

def set_role(response, user, backend, *args, **kwargs):
    try:
        role = response['roles'][0]
    except KeyError:
        raise AuthFailed("No role assigned")

    try:
        group = Group.objects.get(name="{}".format(role))
    except Group.DoesNotExist:
        raise AuthFailed("Unknown role")
`
and adding 'netbox.custom_pipeline.set_role' to my own SOCIAL_AUTH_PIPELINE

PR #9607, opened 2022-06-25
Closes #9540: Filter IP addresses by assigned Device/VM

### LINKED ISSUE #9540: IPAM IP Addresses List no Filter Column Device/VM
### NetBox version

v3.2.4

### Feature type

### Proposed functionality

Filter the IPAM IP Addresses List by Device/VM.

### Use case

Intuitively, you open the list of IP addresses and want to check which IP addresses are assigned to certain devices. This is not possible in this way because the filter ignores the Device/VM column.

So you have to go to the device and list the interfaces.
But this is more complicated if you want to check many devices.

### Database changes

_No response_

### External dependencies

_No response_

huntabyte:
### Fixes: #9540

Added the ability within the UI to filter IP addresses by assigned Devices or Virtual Machines.

### DIFF
--- a/netbox/ipam/forms/filtersets.py
+++ b/netbox/ipam/forms/filtersets.py
@@ -1,7 +1,8 @@
 from django import forms
 from django.utils.translation import gettext as _
 
-from dcim.models import Location, Rack, Region, Site, SiteGroup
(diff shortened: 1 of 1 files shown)


jeremystretch:
Great work, thanks!

PR #9631, opened 2022-06-30
Closes: #8157 - Add L2VPN support

### LINKED ISSUE #8157: Add L2VPN Support - ELINE, ELAN, ETREE, Bridge Domain via VPLS, EVPN, VXLAN
### NetBox version

v3.1.0

### Feature type

New functionality

### Proposed functionality

New data model to support tracking of L2VPNs such as VPLS, EVPN, VXLAN, and other tunneling protocols. This would require two major items.
1) New L2VPN data model
2) Ability to link any vlan and/or interface to the L2VPN.

The proposed functionality would rely on a new data model for L2VPNs. Since there are many different methods to deliver L2VPN service I also propose a new model for "L2VPN Types", similar to prefix/vlan roles, otherwise a drop down with pre-defined L2VPN Types would suffice or we could also use custom fields if we did not want to create the additional "L2VPN Types" model. Creating "L2VPN Types" add some context and it's flexibile so the user can define them according to whats important to them. For instance you could simply define ELINE, ELAN, ETREE... On the other hand you might be interested in technologies such as VPLS or EVPN-IRB.

**1) Create new data models under IPAM.**
* L2VPNs
	* Name (required and unique)
	* L2VPN Type (required and select from drop-down from model "L2VPN Types" or create predefined list)
	* Description
	* Status (required and select from drop-down)
		*Active
		*Planned
		*Inactive
	* Tags
	* Tenancy Assignment
	* Comments
	* Journal and Change Log.
	* Allow custom fields (This will be useful for PW-ID, EVI, RTs, etc-depending on your environment)
* L2VPN Types
	* Name (required and unique)
	* Weight
	* Description
	* Tags

**2) Link Interfaces and Vlans**
_**Link Interfaces -**_
There will only be one L2VPN ever assigned to an interface and we could do this in the same way you will be able to assign an interface to a VRF(Request #7852). You should be able to assign a L3 interface to a VRF and also to a L2VPN. In the case where the interface has multiple L2VPNs, such as a trunk, each L2VPN would have different encapsulation so you would not link at the interface level and only link underlying vlans to the L2VPN. There can be multiple physical or virtual interfaces from any device assigned to a single L2VPN instance.
- On the L2VPN page provide a function to search for existing interfaces and add any interface to the L2VPN instance.
- On the interface page provide a row to associate L2VPNs - much like the upcoming Feature Request #7852.  

_**Link Vlans -**_
There will only be one L2VPN ever assigned to a vlan but there can be multiple vlans assigned to one L2VPN.
- On the L2VPN page provide a function to search for existing vlans and add any vlan to the L2VPN instance.
- On the vlan object provide a row to associate L2VPNs - much like how you assign a prefix to a VRF.

The L2VPN page would show you all the information about the L2VPN and all vlans and/or interfaces that are assigned to this L2VPN so you could easily know all the interfaces. In case of vlans you would need to drill down to the Vlan->Device Interfaces Tab to see the interfaces.

Ideally we would be able to look at the Tenant in Netbox and see a number of L2VPNs assigned to the Tenant. We would then click on the L2VPN icon where it list all the L2VPNs assigned to the Tenant. You would then be able to select the L2VPN of interest.

### Use case

Netbox does a great job at documenting L3VPNs. I hope Netbox can do something similar for L2VPNs. Netbox will only model a L2 domain as one flat vlan. In ISP and Data Center environments we use MPLS, EVPN, VXLAN, etc. each having different vlan IDs and we even bridge vlan ranges together. We need Netbox to document the L2VPN attributes and all the end-points. It would be very powerful to be able to find all interfaces within a L2VPN in a few seconds.

We have tenants that have multiple L2VPN services that use a mix of vlans and/or interfaces(untagged). An L2VPN can span across multiple routers as tagged or untagged traffic at the edge ports. These ports may be physical or virtual or there may be a vlan on trunk interface. In an Integrated-Routing-and-Bridging(IRB) scenario there will be L3 virtual interfaces assigned to an L2VPN or even multiple Distributed Anycast gateways(in EVPN-IRB). Netbox has all the underlying pieces of the L2VPN already documented, we just need a way to link these pieces together to construct an L2VPN. We have VPLS and EVPN mainly in our network but I think the proposed functionality will be flexible enough to document any L2VPN technology. 

**NOTE:** I am familiar with the netbox-virtual-circuit-plugin from vapor-ware. Aside from the project being outdated and not maintained, the biggest deal-breaker is it only supports grouping of vlans and not interfaces. Another important item is that we need to assign these L2VPNs to a Tenant.

### Database changes

New tables for L2VPN models.

### External dependencies

None

DanSheps:
### Fixes: #8157

* Adds the L2VPN model
* Adds the L2VPNTermination model
* Adds related apis, forms, filtersets and views


DanSheps:
We could relocate the models and associated items elsewhere if needed.  IPAM seems the best fit ATM.  Virtualization seems more VM centric.  DCIM doesn't exactly fit.

netbox/ipam/models/l2vpn.py
```diff
@@ -0,0 +1,116 @@
...
+            obj_id = self.assigned_object.pk
+            obj_type = ContentType.objects.get_for_model(self.assigned_object)
+            if L2VPNTermination.objects.filter(assigned_object_id=obj_id, assigned_object_type=obj_type).\
+                    exclude(pk=self.pk).count() > 0:
+                raise ValidationError(f'L2VPN Termination already assigned ({self.assigned_object})')
+
+        # Only check if L2VPN is set and is of type P2P
+        if self.l2vpn and self.l2vpn.type in L2VPNTypeChoices.P2P:
+            if L2VPNTermination.objects.filter(l2vpn=self.l2vpn).exclude(pk=self.pk).count() >= 2:
+                raise ValidationError(f'P2P Type L2VPNs can only have 2 terminations; first delete a termination')
```
jeremystretch, line 116:
```suggestion
            terminations_count = L2VPNTermination.objects.filter(l2vpn=self.l2vpn).exclude(pk=self.pk).count()
            if terminations_count >= 2:
                l2vpn_type = self.l2vpn.get_type_display()
                raise ValidationError(
                    f'{l2vpn_type} L2VPNs cannot have more than two terminations; found {terminations_count} already '
                    f'defined.'
                )
```

netbox/ipam/forms/models.py
```diff
@@ -861,3 +866,94 @@ def clean(self):
...
+        ('L2VPN', ('name', 'slug', 'type', 'identifier', 'description', 'tags')),
+        ('Route Targets', ('import_targets', 'export_targets')),
+        ('Tenancy', ('tenant_group', 'tenant')),
+    )
+
+    class Meta:
+        model = L2VPN
+        fields = (
+            'name', 'slug', 'type', 'identifier', 'description', 'import_targets', 'export_targets', 'tenant', 'tags'
+        )
```
jeremystretch, line 897: Need to manually assign the `StaticSelect` widget for static choice fields (otherwise they look funny).

```suggestion
        fields = (
            'name', 'slug', 'type', 'identifier', 'description', 'import_targets', 'export_targets', 'tenant', 'tags'
        )
        widgets = {
            'type': StaticSelect(),
        }
```

netbox/ipam/choices.py
```diff
@@ -170,3 +170,52 @@ class ServiceProtocolChoices(ChoiceSet):
...
+    TYPE_EPLAN = 'ep-lan'
+    TYPE_EVPLAN = 'evp-lan'
+    TYPE_EPTREE = 'ep-tree'
+    TYPE_EVPTREE = 'evp-tree'
+    TYPE_VXLAN = 'vxlan'
+    TYPE_VXLAN_EVPN = 'vxlan-evpn'
+    TYPE_MPLS_EVPN = 'mpls-evpn'
+    TYPE_PBB_EVPN = 'pbb-evpn'
+
+    CHOICES = (
```
jeremystretch, line 189: Is there a logic to the group ordering? Seems random.

netbox/ipam/forms/models.py
```diff
@@ -861,3 +866,94 @@ def clean(self):
...
+
+
+class L2VPNTerminationForm(NetBoxModelForm):
```
jeremystretch, line 900: I think this needs to be extended to allow selecting a VM interface as well.

Also, we should include a "Connect L2VPN" link from the VLAN and interface views for convenience.

netbox/ipam/api/serializers.py
```diff
@@ -433,3 +437,58 @@ class Meta:
...
+        many=True
+    )
+    tenant = NestedTenantSerializer(required=False, allow_null=True)
+
+    class Meta:
+        model = L2VPN
+        fields = [
+            'id', 'url', 'display', 'identifier', 'name', 'slug', 'type', 'import_targets', 'export_targets',
+            'description', 'tenant',
+            # Extra Fields
```
jeremystretch, line 468:
```suggestion
```

netbox/ipam/api/serializers.py
```diff
@@ -433,3 +437,58 @@ class Meta:
...
+        queryset=ContentType.objects.all()
+    )
+    assigned_object = serializers.SerializerMethodField(read_only=True)
+
+    class Meta:
+        model = L2VPNTermination
+        fields = [
+            'id', 'url', 'display', 'l2vpn', 'assigned_object_type', 'assigned_object_id',
+            'assigned_object',
+            # Extra Fields
```
jeremystretch, line 486:
```suggestion
```

netbox/ipam/api/views.py
```diff
@@ -18,6 +18,7 @@
 from utilities.constants import ADVISORY_LOCK_KEYS
 from utilities.utils import count_related
 from . import serializers
+from ..models.l2vpn import L2VPN, L2VPNTermination
```
jeremystretch, line 21:
```suggestion
from ipam.models import L2VPN, L2VPNTermination
```

netbox/ipam/api/views.py
```diff
@@ -157,6 +158,18 @@ class ServiceViewSet(NetBoxModelViewSet):
     filterset_class = filtersets.ServiceFilterSet
 
 
+class L2VPNViewSet(NetBoxModelViewSet):
+    queryset = L2VPN.objects
```
jeremystretch, line 162:
```suggestion
    queryset = L2VPN.objects.prefetch_related('import_targets', 'export_targets', 'tenant', 'tags')
```

netbox/templates/ipam/l2vpn.html
```diff
@@ -0,0 +1,111 @@
...
+                        {% endif %}
+                      </td>
+                    </tr>
+                  {% endfor %}
+                </table>
+              {% else %}
+                <div class="text-muted">None</div>
+              {% endif %}
+            {% endwith %}
+          </div>
```
jeremystretch, line 95: This should be rendered using `L2VPNTerminationTable`.

netbox/ipam/urls.py
```diff
@@ -186,4 +186,26 @@
...
+    path('l2vpn-termination/add/', views.L2VPNTerminationEditView.as_view(), name='l2vpntermination_add'),
+    path('l2vpn-termination/import/', views.L2VPNTerminationBulkImportView.as_view(), name='l2vpntermination_import'),
+    path('l2vpn-termination/edit/', views.L2VPNTerminationBulkEditView.as_view(), name='l2vpntermination_bulk_edit'),
+    path('l2vpn-termination/delete/', views.L2VPNTerminationBulkDeleteView.as_view(), name='l2vpntermination_bulk_delete'),
+    path('l2vpn-termination/<int:pk>/', views.L2VPNTerminationView.as_view(), name='l2vpntermination'),
+    path('l2vpn-termination/<int:pk>/edit/', views.L2VPNTerminationEditView.as_view(), name='l2vpntermination_edit'),
+    path('l2vpn-termination/<int:pk>/delete/', views.L2VPNTerminationDeleteView.as_view(), name='l2vpntermination_delete'),
+    path('l2vpn-termination/<int:pk>/changelog/', ObjectChangeLogView.as_view(), name='l2vpntermination_changelog', kwargs={'model': L2VPNTermination}),
+    path('l2vpn-termination/<int:pk>/journal/', ObjectJournalView.as_view(), name='l2vpntermination_journal', kwargs={'model': L2VPNTermination}),
```
jeremystretch, line 210: These paths should start with `'l2vpns/` and `l2vpn-terminations/`, as we always use the plural form of the model name throughout the rest of NetBox.

netbox/ipam/tables/l2vpn.py
```diff
@@ -0,0 +1,38 @@
...
+from ipam.models.l2vpn import L2VPN, L2VPNTermination
+from netbox.tables import NetBoxTable, columns
+
+__all__ = (
+    'L2VPNTable',
+    'L2VPNTerminationTable',
+)
+
+
+class L2VPNTable(NetBoxTable):
```
jeremystretch, line 13: Should probably add columns for the import and export route targets too. Should be able to replicate what we did for `VRFTable`.

netbox/ipam/tables/l2vpn.py
```diff
@@ -0,0 +1,38 @@
...
+        linkify=True
+    )
+
+    class Meta(NetBoxTable.Meta):
+        model = L2VPN
+        fields = ('pk', 'name', 'description', 'slug', 'type', 'tenant', 'actions')
+        default_columns = ('pk', 'name', 'description', 'actions')
+
+
+class L2VPNTerminationTable(NetBoxTable):
```
jeremystretch, line 25: The `l2vpn` column should be linkified too.

netbox/netbox/navigation_menu.py
```diff
@@ -260,6 +260,13 @@ def get_model_buttons(app_label, model_name, actions=('add', 'import')):
                 get_model_item('ipam', 'vlangroup', 'VLAN Groups'),
             ),
         ),
+        MenuGroup(
+            label='L2VPNs',
+            items=(
+                get_model_item('ipam', 'l2vpn', 'L2VPN'),
```
jeremystretch, line 266:
```suggestion
                get_model_item('ipam', 'l2vpn', 'L2VPNs'),
```

jeremystretch (changes requested):
Great work, @DanSheps! I have some minor notes, mostly around ensuring consistency with the rest of NetBox. We also need to ensure we support L2VPN termination for both device and VM interfaces.

I haven't dug into the tests yet.

docs/models/ipam/l2vpntermination.md
```diff
@@ -0,0 +1,12 @@
+# L2VPN Termination
+
+A L2VPN Termination is the termination point of a L2VPN.  Certain types of L2VPN's may only have 2 termination points (point-to-point) while others may have many terminations (multipoint).
+
+Each termination consists of a L2VPN it is a member of as well as the connected endpoint which can be an interface or a VLAN.
+
+The following types of L2VPN's are considered point-to-point:
```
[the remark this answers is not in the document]
jeremystretch: I see further down that it does limit the number of terminations per L2VPN; we should note this in the docs.

netbox/ipam/models/l2vpn.py
```diff
@@ -0,0 +1,116 @@
...
+        return reverse('ipam:l2vpntermination', args=[self.pk])
+
+    def clean(self):
+        # Only check is assigned_object is set
+        if self.assigned_object:
+            obj_id = self.assigned_object.pk
+            obj_type = ContentType.objects.get_for_model(self.assigned_object)
+            if L2VPNTermination.objects.filter(assigned_object_id=obj_id, assigned_object_type=obj_type).\
+                    exclude(pk=self.pk).count() > 0:
+                raise ValidationError(f'L2VPN Termination already assigned ({self.assigned_object})')
```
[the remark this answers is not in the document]
DanSheps: This results in a IntegrityError which we most likely want to catch before it gets to validate_unique()

netbox/ipam/choices.py
DanSheps: Going to re-arrange that.  I am going to put the "E-XXX" at the bottom as I think the VPLS/VXLAN/L2VPN E-VPN will be more commonly used.  Thoughts?

netbox/ipam/models/l2vpn.py
```diff
@@ -0,0 +1,116 @@
...
+        related_name='l2vpns',
+        blank=True,
+        null=True
+    )
+    contacts = GenericRelation(
+        to='tenancy.ContactAssignment'
+    )
+
+    class Meta:
+        ordering = ('identifier', 'name')
```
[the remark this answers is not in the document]
jeremystretch: Oh my mistake, I was thinking `name` was optional. This should be okay.

PR #9639, opened 2022-06-30
PR for #8723

### LINKED ISSUE #8723: Bulk-rename for devices
### NetBox version

v3.1.6

### Feature type

New functionality

### Proposed functionality

Currently there is no bulk-rename functionality for devices. There is only a bulk-rename for interfaces. I would like to request to get the same bulk-rename functionality with (regex-)search/replace for devices.

### Use case

We are currently facing a renumbering of the rooms and buildings of our organisation to a more unique scheme. As our devices names in netbox contain some location info it would be a big help to rename the devices with the bulk-rename functionality. 

### Database changes

_No response_

### External dependencies

_No response_

cpund:
### Fixes: #8723 

WIP of implementing bulk rename for devices. Devices without a name need special handling implemented still.

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -1786,6 +1786,12 @@ class DeviceBulkDeleteView(generic.BulkDeleteView):
     table = tables.DeviceTable
 
 
+class DeviceBulkRenameView(generic.BulkRenameView):
+    queryset = Device.objects.prefetch_related('tenant', 'site', 'rack', 'device_role', 'device_type__manufacturer')
+    filterset = filtersets.DeviceFilterSet
+    table = tables.DeviceTable
+
+
 #
 # Devices
 #
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -1787,7 +1787,7 @@ class DeviceBulkDeleteView(generic.BulkDeleteView):
 
 
 class DeviceBulkRenameView(generic.BulkRenameView):
-    queryset = Device.objects.prefetch_related('tenant', 'site', 'rack', 'device_role', 'device_type__manufacturer')
+    queryset = Device.objects.all()
     filterset = filtersets.DeviceFilterSet
     table = tables.DeviceTable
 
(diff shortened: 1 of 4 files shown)


netbox/dcim/views.py
```diff
@@ -1786,6 +1786,12 @@ class DeviceBulkDeleteView(generic.BulkDeleteView):
     table = tables.DeviceTable
 
 
+class DeviceBulkRenameView(generic.BulkRenameView):
+    queryset = Device.objects.prefetch_related('tenant', 'site', 'rack', 'device_role', 'device_type__manufacturer')
```
jeremystretch, line 1790: We can omit `prefetch_related()` here since we're only dealing with the object names.

```suggestion
    queryset = Device.objects.all()
```

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -676,6 +676,9 @@ def post(self, request):
         else:
             form = self.form(initial={'pk': request.POST.getlist('pk')})
             selected_objects = self.queryset.filter(pk__in=form.initial['pk'])
+            for object in selected_objects:
+                # Do something to raise error message to user
+                pass
```
jeremystretch, line 681: The view should already handle the necessary validation. Did you run into something that isn't covered?

jeremystretch (changes requested):
This should be sufficient for devices, however we should consider extending the bulk rename function to other objects as well.

jeremystretch:
Hey @cpund are you still available to work on this?

cpund:
> Hey @cpund are you still available to work on this?

Apologies - I never got the github notification for your previous comments. I'll work on getting to this later.

netbox/netbox/views/generic/bulk_views.py
cpund: I believe I was just trying to raise an error when selecting a device without a name. Going over things, I definitely prefer your previous suggestion of `obj.name or ''`. Removing these lines.

netbox/dcim/views.py
cpund: Done.

cpund:
> This should be sufficient for devices, however we should consider extending the bulk rename function to other objects as well.

I definitely agree, just currently working on learning the Netbox code-base to a greater extent before I try and perform something as wide-reaching. I haven't spent much time on it since June going over it and I've forgotten a lot.

cpund:
@jeremystretch I believe I've reviewed and corrected the changes you requested. If you want to just go straight to adding this to bulk rename other objects as well, I can spend more time reviewing the code base and finding candidates where it makes sense that don't currently have the info.

jeremystretch:
Looks great, thanks @cpund!

PR #9695, opened 2022-07-10
Add Tenant Group column to tables #9686

### LINKED ISSUE #9686: Add Tenant Group column to tables
### NetBox version

v3.2.5

### Feature type

New functionality

### Proposed functionality

Add a new column to display the Tenant Group of the records Tenant.

This would require a new column `tenant_group` on all tables which also has the `tenant` column.

It would not be a default column.

It would also require (For performance reasons) that the tenant.group is prefetched in all views.

### Use case

Our use case is that we represent a customer as a Tenant Group and theirs different contracts as Tenants. This means that we end up with a contract number as the Tenant name, which makes it a bit hard to figure out at a glance which customer a thing belongs to.

If we could add the Tenant Group (Which contains the customer name) to the table, it would solve our user experience issue.

### Database changes

None

### External dependencies

None

hagbarddenstore:
### Fixes: #9686

Add a column to display Tenant Group on tables which already displays the Tenant.

### DIFF
--- a/netbox/tenancy/tables/columns.py
+++ b/netbox/tenancy/tables/columns.py
@@ -3,6 +3,7 @@ import django_tables2 as tables
 __all__ = (
     'TenantColumn',
     'TenantGroupColumn',
+    'TenancyColumnsMixin',
 )
 
 
@@ -50,3 +51,7 @@ class TenantGroupColumn(tables.TemplateColumn):
     def value(self, value):
         return str(value) if value else None
 
+
+class TenancyColumnsMixin(tables.Table):
+    tenant_group = TenantGroupColumn()
+    tenant = TenantColumn()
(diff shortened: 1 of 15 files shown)


netbox/tenancy/tables/columns.py
```diff
@@ -24,3 +26,32 @@ def __init__(self, *args, **kwargs):
...
     def value(self, value):
         return str(value) if value else None
+
+
+class TenantGroupColumn(tables.TemplateColumn):
+    """
+    Include the tenant group description.
+    """
+    template_code = """
+    {% if record.tenant and record.tenant.group %}
```
jeremystretch, line 36: This is where we need to be careful with regard to prefetching. Because the column itself doesn't define an accessor, the tenant group doesn't get prefetched in the queryset. Instead, a separate query is made for `tenant.group` as needed for each unique tenant. I think we can resolve this by specifying `tenant__group` as the accessor for the column and tweaking the template code, but I'll merge this as-is and try a few things.

jeremystretch:
Thanks for taking this!

PR #9696, opened 2022-07-10
Fixes #9632 - Focus on select field after populating data

### LINKED ISSUE #9632: Filter fields on drop downs don't gain focus automatically
### NetBox version

v3.1.11

### Python version

3.9

### Steps to Reproduce

1. Go to the sites list
2. Select the filter tab
3. Click the "Region", "Tenant", "Site Group" or any other drop-down option that prompts you with a "filter" search box - the cursor isn't automatically focussed to this filter text field.

### Expected Behavior

Like the "Status" and "Tags" filter fields that present a "search" input text box, all filtering text inputs should automatically gain keyboard focus for 0-click text entry.

### Observed Behavior

The "filter" field text boxes remain unfocussed and require an additional mouse-click.

kkthxbye-code:
### Fixes: #9632

Adds a call to `focus()` on the search field in select boxes when data is loaded from the API. This prevents cases where focus is not set if the API doesn't respond immediately.


jeremystretch:
Nice, thanks!

PR #9700, opened 2022-07-11
Fixes: #9656 Re-order journal list and form

PieterL75:
### Fixes: #9656 Re-order journal list and form

On the the Journal tab, the 'table' of entries is placed above the 'add journal' card.


jeremystretch:
Thanks!

PR #9712, opened 2022-07-13
Include Network information in Prefix Template Issue:#9505

### LINKED ISSUE #9505: Include Network information in Prefix/Aggregate view
### NetBox version

v3.2.4

### Feature type

### Proposed functionality

Inlcude additinal information in the Prefix/Aggregate view to make it easier for admins to get more knowledge to the network.
The example would be the output of sipcalc. For me the most valueable information is boradcast address, quick view on the usable range and cisco wildcard address.

```
$ sipcalc 10.14.0.0/19
-[ipv4 : 10.14.0.0/19] - 0

[CIDR]
Host address		- 10.14.0.0
Host address (decimal)	- 168689664
Host address (hex)	- A0E0000
Network address		- 10.14.0.0
Network mask		- 255.255.224.0
Network mask (bits)	- 19
Network mask (hex)	- FFFFE000
Broadcast address	- 10.14.31.255
Cisco wildcard		- 0.0.31.255
Addresses in network	- 8192
Network range		- 10.14.0.0 - 10.14.31.255
Usable range		- 10.14.0.1 - 10.14.31.254
```

### Use case

Get additional network information more quickly.

### Database changes

_No response_

### External dependencies

_No response_

renatoalmeidaoliveira:
### Fixes: #9505
Add Network Mask,  Broadcast Address, and Wildcard Mask to prefix template

### DIFF
--- a/netbox/templates/ipam/prefix.html
+++ b/netbox/templates/ipam/prefix.html
@@ -143,6 +143,32 @@
               {% endwith %}
             </td>
           </tr>
+	  {% if object.prefix.version == 4 %}
+	  <tr>
+	    <td>
+		    Network Mask
+            </td>
+	    <td>
+		    {{ object.prefix.netmask }}		
+            </td>
+	  </tr>
+	  <tr>
+	    <td>
+		    Broadcast Address
+            </td>
+	    <td>
+		    {{ object.prefix.broadcast }}		
+            </td>
+	  </tr>
+	  <tr>
+	    <td>
+		    Wildcard Mask
+            </td>
+	    <td>
+		    {{ object.prefix.hostmask }}		
+            </td>
+	  </tr>
+	  {% endif %}
         </table>
       </div>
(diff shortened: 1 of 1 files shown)


jeremystretch:
Rather than adding this information to the main panel where it will take up valuable screen space, a pop-up modal might be more appropriate. I'd expect that this information isn't of interest in the vast majority of cases, so it would be best to tuck it away where it can be summoned as needed (e.g. by clicking a "details" button next to the prefix value).

renatoalmeidaoliveira:
Do you think in something like the Delete modal?

renatoalmeidaoliveira:
@jeremystretch what do you think? this time with modal.
image

jeremystretch:
I'd like to reposition the button, but this looks good, thanks!

jeremystretch:
CI failure is unrelated to changes.

PR #9716, opened 2022-07-13
Closes #9599: Add cursor pagination mode

### LINKED ISSUE #9599: Provide CursorPagination as an alternative to LimitOffsetPagination
### NetBox version

v3.2.5

### Feature type

### Proposed functionality

The proposed functionality and use cases are well documented in https://github.com/netbox-community/netbox/issues/3699 which was closed by the stale bot. This creates a new issue for it. I am willing to implement and own the changes once it gets accepted. (I've actually already implemented a working version of it. I'll be happy to submit a PR and refine it. )

To reiterate the things that were discussed in the original issue:

Provide an option to use **CursorPagination** instead of the default **LimitOffsetPagination** for all GET endpoints that retrieve a list of objects. This will allow clients to retrieve a consistent list of objects even if elements are changed/removed during iteration.

### Proposed changes:
- Interface: provide a new query parameter `pagination_mode` so clients can choose to use **CursorPagination**. When in **CursorPagination** mode, `limit` can be used to control page size (it has the same functionality as `limit` in **LimitOffsetPagination** mode). **LimitOffsetPagination** is still the default so the change will not effect any existing clients.
- Implementation: implement a custom Pagination class that uses either **CursorPagination** or **LimitOffsetPagination** depending on `pagination_mode`. Internally, it still uses the pagination implementation provided by Django REST. When in **CursorPagination** mode, records will be returned in "-create" order.

#### 1. Get/Delete/Update all records
With **LimitOffsetPagination**, it is difficult to do this consistently. Take getting all records of a data model as an example: if I have a loop that fetches records page by page, it may end up missing valid records because a record is deleted while the script is running and the offsets are no longer valid.

To work around problems like this, the script needs to iterate through the list multiple times until it gets a "stable" list.

This also applies to attempts to delete or update all records. It is even more difficult to deal with if the update is not idempotent as it is not an option to apply the update multiple times.

#### 2. Pagination on a large dataset
As mentioned in [Django REST framework documentation](https://www.django-rest-framework.org/api-guide/pagination/#cursorpagination:~:text=Supports%20usage%20with%20very%20large%20datasets.%20With%20extremely%20large%20datasets%20pagination%20using%20offset%2Dbased%20pagination%20styles%20may%20become%20inefficient%20or%20unusable.%20Cursor%20based%20pagination%20schemes%20instead%20have%20fixed%2Dtime%20properties%2C%20and%20do%20not%20slow%20down%20as%20the%20dataset%20size%20increases.), **LimitOffsetPagination** may become inefficient on extremely large datasets and **CursorPagination** has fixed-time properties.

### Database changes

None

### External dependencies

None

lyuyangh:
### Fixes: #9599
Implemented `TwoModePagination` to support both [Cursor Pagination](https://www.django-rest-framework.org/api-guide/pagination/#cursorpagination) and Limit Offset Pagination (default). The client can now specify `pagination_mode=cursor` to use Cursor Pagination. Please see Swagger UI for more information.
- `CursorPaginationWithNoLimit` follows `OptionalLimitOffsetPagination` and allows the client to turn off limit by specifying `limit=0`. Most of the code is copied from the stock implementation of `CursorPagination`.
- When in Cursor Pagination mode, it returns the results in "-create"
Screen Shot 2022-08-10 at 2 12 49 PM


netbox/netbox/api/pagination.py
```diff
@@ -83,3 +87,167 @@ def get_queryset_count(self, queryset):
...
+                if MAX_PAGE_SIZE:
+                    return MAX_PAGE_SIZE if page_size == 0 else min(page_size, MAX_PAGE_SIZE)
+                return page_size
+            except (KeyError, ValueError):
+                pass
+
+        return self.default_page_size
+
+
+class TwoModePagination(BasePagination):
```
jeremystretch, line 199: I understand why this class was introduced, given that DRF does not support dynamic paginator selection, but I'm afraid this approach isn't tenable. We need to keep the paginator construction relatively simple to ensure its long-term maintainability, and using a sort of proxy class to choose between two different paginators is far from ideal.

Instead, we should be able to override the `paginator()` method of GenericAPIView to select one class or the other based on the value of the `pagination_mode` parameter. (`self.request` is accessible when `paginator()` is called.) I

If that ends up not being feasible, perhaps a reasonable compromise would be expose the paginator as a NetBox configuration parameter. This would allow an administrator to designate either pagination strategy application-wide, however it would not be user-selectable.

jeremystretch:
Thanks for you work on this @lyuyangh! Per my comment above, we'll need to come up with a simpler way to handle the selection of the paginator class, but I think I can solve that. I'm going to merge this PR into an intermediary branch for now to continue work on it.

lyuyangh:
Hi @jeremystretch , thanks for reviewing the PR and taking this over! Now that you are working on it, I know it is in good hands.

jeremystretch:
Hi @lyuyangh, do you have any idea about the first issue I mentioned [here](https://github.com/netbox-community/netbox/pull/10764#issuecomment-1292640255)? Seems that it was an issue with the original PR that arose sometime after it was opened, maybe due to a change in DRF version.

Issue #9713, opened 2022-07-13
Can't clear Device field Rack in UI


JulijusBertasius:
### NetBox version

v3.2.6

### Python version

3.8

### Steps to Reproduce

1. In UI press any racked device edit button
2. Press x in rack field

I tried it on 2 different computers and 3 different browsers. Problems also happens on devices I tried in demo.netbox.dev.

### Expected Behavior

Rack field is empty

### Observed Behavior

Page freezes

kkthxbye-code:
Like #9734 this is also caused by #9696 - should be fixed in #9735.

I have no idea how it manages to freeze the browser, it might be a Chrome bug, as it doesn't happen in firefox. Seems to be related to setting focus on an element while it is being manipulated in another way.

jeremystretch:
I'm going to close this as covered by #9734, but please ask to have it reopened if you determine that #9735 does not address the problem.

PR #9735, opened 2022-07-15
Fixes #9734 & #9713 - Only set focus on select field search boxes if the select is open

### LINKED ISSUE #9734: Other select fields steal focus if the choice filters other select fields (caused by #9632)
### NetBox version

v3.2.6

### Python version

3.10

### Steps to Reproduce

1. Create manufacturer test
2. Create devicetype test with manufacturer test
3. Go to create device form
4. Select test under manufacturer

### Expected Behavior

Select field closes.

### Observed Behavior

The devicetype field opens and take focus.

kkthxbye-code:
### Fixes: #9734 and #9713

This was caused by me missing that selecting fields like manufacturer would pre-load the dependant field (devicetype).

I have no idea if there's a smarter way of knowing if the select field is open. Checking for visibility in javascript seems like a nightmare, and I was unable to find a built-in property or method in slim-select that tells you if a select field is open.

I did try with `this.slim.data.contentOpen`, however it will return false when the select is in the process of opening, as it is only set on animation completion. Not sure if that is an actual issue though, as slim-select will set the focus correctly if the API returns fast.


jeremystretch:
Thanks for jumping on this so quickly!

PR #9755, opened 2022-07-17
Fixes #9754 - Fix chrome deadlock when opening some select fields

### LINKED ISSUE #9754: Selecting the position field on device creation page without selecting a rack crashes the chrome tab
### NetBox version

v3.2.7-dev

### Python version

3.10

### Steps to Reproduce

1. Navigate to the device creation form
2. Click on the "Position" select field

### Expected Behavior

Select field opens

### Observed Behavior

Chrome tab crashes. Firefox works fine.

I have absolutely no idea what is going on here. Must be some kind of race condition with slim-selects own call to focus and netbox.

kkthxbye-code:
### Fixes: #9754

This PR tries to fix an unidentified deadlock in chrome when opening certain slim-select fields. I tested as much as I could, but as I am still not sure what actually causes this, I think another valid (maybe safer) solution would be to revert #9735 and #9696 and reopen #9632 until someone who can untangle the hell of async events being fired when opening a slim-select field - I'll let you decide @jeremystretch .


kkthxbye-code:
After having a look around I think the issue might be an infinite loop stemming from the eventlistener:

https://github.com/brianvoe/slim-select/blob/e6b42a9c11ba743fab63b5f909d8499239362d98/src/slim-select/slim.ts#L420

As such I am not confident in this fix, or any fix that calls focus() on the input. Not sure how to fix it with messing with unregistering the listener, and as we are already kinda misusing slim-select internals, that's not something I'm comfortable handling correct.

I created a new PR to revert the previous changes. When that is done, we need to re-open #9632 and waiting for someone with more js experience to show up.

PR #9807, opened 2022-07-20
Add DEVELOPER setting to example config

### LINKED ISSUE #9806: Add DEVELOPER setting to configuration.example.py
### Proposed Changes

Adding the `DEVELOPER = False` setting to the configuration.example.py would make it easier to find and better documented (in addition to the developer setup documentation, of course).

### Justification

I'm an idiot and set `DEVELORERS = True` and was very confused for several minutes when I tried to create database migrations. Just creating issue + PR here on the offchance of saving anybody else the trouble in the future.

rixx:
### Fixes: #9806

As described, adds a default setting so people with similarly bad typing skills won't be as confused.

### DIFF
--- a/netbox/netbox/configuration_example.py
+++ b/netbox/netbox/configuration_example.py
@@ -103,6 +103,9 @@ CORS_ORIGIN_REGEX_WHITELIST = [
 # on a production system.
 DEBUG = False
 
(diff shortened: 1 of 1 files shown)


jeremystretch:
Please see my comment on #9806 for why this won't be merged.

PR #9854, opened 2022-07-27
Fixes #9758: Fix path end device and provider network information on CableTable and DeviceInterfaceTable

### LINKED ISSUE #9758: v3.3-beta1: Connection field in interfaces view of device only shows name of connection not device
### NetBox version

v3.3-beta1

### Python version

3.10

### Steps to Reproduce

1. Open device show connection field in table
2. Connect interface to another device
3. 
image

### Expected Behavior

V3.2.6 connection field shows devicename + port assigned
image

### Observed Behavior

Connection field only shows portname instead of device_name > portname

jaylik:
### Fixes: #9758 

Please note that this won't fix existing CablePath objects that are marked with is_complete = False when path is initially connected

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -576,6 +576,7 @@ class CablePath(models.Model):
                         [object_to_path_node(circuit_termination)],
                         [object_to_path_node(circuit_termination.provider_network)],
                     ])
+                    is_complete = True
                     break
                 elif circuit_termination.site and not circuit_termination.cable:
                     # Circuit terminates to a Site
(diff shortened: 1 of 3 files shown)


netbox/dcim/models/cables.py
```diff
@@ -576,6 +576,7 @@ def from_origin(cls, terminations):
                         [object_to_path_node(circuit_termination)],
                         [object_to_path_node(circuit_termination.provider_network)],
                     ])
+                    is_complete = True
```
jeremystretch, line 579: This isn't needed for this change.

jeremystretch (commented):
Apologies, I forgot we had a PR in for this and resolved #9758 while knocking out some other stuff for v3.3.

PR #9855, opened 2022-07-27
Fixes #9789: Fix svg width calculation

### LINKED ISSUE #9789: Cable Trace for interface terminated on circuit not working
### NetBox version

v3.3-beta1

### Python version

3.9

### Steps to Reproduce

1. Create circuit
2. Create router with interface
3. Terminate circuit on router interface
4. Try to run Cable Trace

### Expected Behavior

Tracing should work like in v3.2.6

### Observed Behavior

Trace returns empty.

jaylik:

### DIFF
--- a/netbox/dcim/svg/cables.py
+++ b/netbox/dcim/svg/cables.py
@@ -178,7 +178,7 @@ class CableTraceSVG:
         """
         Draw a set of parent objects.
         """
-        width = self.width / len(obj_list)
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks for the fix @jaylik, however I'm afraid the root issue goes a bit deeper than just the SVG rendering. (It concerns how attachments to provider networks are rendered from a path.) I think I have a fix worked out, and will merge something soon.

PR #9870, opened 2022-07-28
Closes #9637: Add existing fields to the Rack Reservation user interface pages

### LINKED ISSUE #9637: Add existing fields to the Rack Reservation user interface pages
### NetBox version

v3.2.4

### Feature type

### Proposed functionality

Proposing to add already existent fields to the Rack Reservation create/edit/list user interface pages.

1. Add the **Site Group** field to the Rack Reservation **add** and **edit** pages to filter the **Site** input.
image
image
2. Add the **Location** field as a column option for the Rack Reservation **list** page.
image
3. Change the **Rack** section name in the **filter** tab for the Rack Reservation **list** page to **Location**.
image

**Site group** and **Location** are already options as filters in the Rack Reservation list page.

### Use case

The proposed changes would better align the user interface for the RackReservation model to other similar model pages (Rack) and provide additional filtering options.

### Database changes

_No response_

### External dependencies

_No response_

atownson:
Closes #9637

### DIFF
--- a/netbox/dcim/forms/filtersets.py
+++ b/netbox/dcim/forms/filtersets.py
@@ -287,7 +287,7 @@ class RackReservationFilterForm(TenancyFilterForm, NetBoxModelFilterSetForm):
     fieldsets = (
         (None, ('q', 'tag')),
         ('User', ('user_id',)),
-        ('Rack', ('region_id', 'site_group_id', 'site_id', 'location_id')),
+        ('Location', ('region_id', 'site_group_id', 'site_id', 'location_id')),
         ('Tenant', ('tenant_group_id', 'tenant_id')),
     )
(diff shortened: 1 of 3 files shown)


netbox/dcim/forms/filtersets.py
```diff
@@ -287,7 +287,7 @@ class RackReservationFilterForm(TenancyFilterForm, NetBoxModelFilterSetForm):
     fieldsets = (
         (None, ('q', 'tag')),
         ('User', ('user_id',)),
-        ('Rack', ('region_id', 'site_group_id', 'site_id', 'location_id')),
```
jeremystretch, line 290: This should be left as-is, but we need to add a rack filter field.

PR #9877, opened 2022-07-29
Resolves #9876 Conditions for webhook process tags

### LINKED ISSUE #9876: Improve support for referencing tags in webhook conditions
### NetBox version

v3.2.7

### Feature type

### Proposed functionality

Execute webhook only when we add tag to object. 
    {
      "attr": "tags",
      "value": "exempt",
      "op": "contains"
    }

Now, it does not work because tags are list of OrderedDict. It is not the bug #9797. 

### Use case

Only when we add tag to object (devices, ip-addresses and etc)  webhook is executed.

beginin:
### Fixes: #9876
Add compare in webhook condition for list of dict

### DIFF
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -1,5 +1,6 @@
 import functools
 import re
+from collections import OrderedDict
 
 __all__ = (
     'Condition',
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks for the proposal, but I'm afraid this sort of approach isn't going to work. It's extremely error-prone and will be difficult to maintain long-term. Please also note that #9876 has not yet been accepted, because no workable solution has been proposed. Happy to discuss further under that issue.

PR #9951, opened 2022-08-08
Fixes #9950 Open redirect fix after login

### LINKED ISSUE #9950: Open redirect to another domain after login with next query
### NetBox version

v3.2.7

### Python version

3.9

### Steps to Reproduce

On login the query string next can redirect to another domain with //www.google.com 
/login/?next={path} 
/login/?next=/home
/login/?next=//www.google.com 

### Expected Behavior

Only redirect to a Path of the current domain.
/login/?next=/home (success)
/login/?next=//www.google.com  (should fail and redirect to home)

### Observed Behavior

/login/?next=/home (success)
/login/?next=//www.google.com  (success)

pellepelle3:
### Fixes: #9950

This is just to make sure after login you are just redirecting to a path not another domain.
Otherwise if some one clicks on a link to Login, it could redirect them to malicious site without realizing it.

### DIFF
--- a/netbox/users/views.py
+++ b/netbox/users/views.py
@@ -91,7 +91,7 @@ class LoginView(View):
         data = request.POST if request.method == "POST" else request.GET
         redirect_url = data.get('next', settings.LOGIN_REDIRECT_URL)
 
-        if redirect_url and redirect_url.startswith('/'):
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks for this @pellepelle3. Turns out Django has a `url_has_allowed_host_and_scheme()` utility function (introduced in v3.0) we can use which covers this as well as some additional corner cases.

PR #9964, opened 2022-08-09
Add a "clear" button for quick search

### LINKED ISSUE #9857: Add a "clear" button for object list quick search fields
### NetBox version

v3.2.7

### Feature type

New functionality

### Proposed functionality

Add a button next to the quick search field on object lists to clear the search value and reload the entire object list.

### Use case

Provides a slightly more convenient user experience when utilizing the quick search feature.

### Database changes

_No response_

### External dependencies

_No response_

jsenecal:
### Fixes: #9857
**Proposed functionality**
Add a button next to the quick search field on object lists to clear the search value and reload the entire object list.

**Use case**
Provides a slightly more convenient user experience when utilizing the quick search feature.


netbox/templates/inc/table_controls_htmx.html
```diff
@@ -2,31 +2,21 @@
...
-        type="text"
-        name="q"
-        class="form-control"
-        placeholder="Quick search"
-        hx-get="{{ request.full_path }}"
-        hx-target="#object_list"
-        hx-trigger="keyup changed delay:500ms"
-      />
+    <div class="input-group input-group-sm quicksearch hide-last-child">
+      <input type="search" results=5 name="q" id="quicksearch" class="form-control" placeholder="Quick search"
```
jeremystretch, line 6: Is `results=5` needed?

netbox/templates/dcim/device/interfaces.html
```diff
@@ -4,84 +4,85 @@
...
-            placeholder="Quick search"
-            hx-get="{{ request.full_path }}"
-            hx-target="#object_list"
-            hx-trigger="keyup changed delay:500ms"
-        />
-      </div>
+<div class="row mb-3 justify-content-between">
+  <div class="col col-12 col-lg-4 my-3 my-lg-0 d-flex noprint table-controls">
+    <div class="input-group input-group-sm quicksearch hide-last-child">
+      <input type="search" results=5 name="q" id="quicksearch" class="form-control" placeholder="Quick search"
```
jeremystretch, line 10: Same question as below re: `results=5`

jeremystretch (approved):
Awesome work @jsenecal, thank you! Had one question about a particular element but this works great!

netbox/templates/inc/table_controls_htmx.html
jsenecal: Switching the input field type to search enables HTML5 functionalities. 
One of those functionalities is to present the user a drop-down of recent searches, results limits this field to a fixed number of recent searches.
[results documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#results)

To be honest it's not really needed but makes the UI a bit easier to work with on certain browsers such as Safari (Computers and Mobiles)

netbox/templates/dcim/device/interfaces.html
jsenecal: Same answer :)

netbox/templates/inc/table_controls_htmx.html
jeremystretch: TIL, thanks!

PR #9981, opened 2022-08-11
re-enable markdown in custom columns

### LINKED ISSUE #9979: Markdown doesn't render in custom fields in table view
### NetBox version

v3.2.8

### Python version

3.8

### Steps to Reproduce

1. Create a custom field of type Text(long) for any content type, e.g. IPAM->IP addresses
2. Add some Markdown or raw HTML to the custom column for an IP address object in the IPAM
3. See that the Markdown/raw HTML is not rendering in the IPAM->IP addresses table view

This was working in v3.2.7, so this is probably due to this merge request:
https://github.com/netbox-community/netbox/commit/38350a102320ea4b1f6bdb81b454c452f9779d2b 

### Expected Behavior

Markdown should render in custom columns in the table view for any content type if field type Text(long) was chosen for the custom column.

### Observed Behavior

Markdown/raw HTML is not rendering for Text(long) custom columns in the table view of any content type like IPAM->IP addresses.

chcon:
### Fixes: #9979

Introduced a new column class _CustomFieldMarkdownColumn_ that combines the default markdown rendering behavior of the _MarkdownColumn_ with the custom column specific constructor parameters of the _CustomFieldColumn_.
That way markdown in custom columns should be both XSS safe and functional.


chcon:
@jeremystretch agreed, please check it again.

PR #9982, opened 2022-08-11
Closes #8595: Added new PON interface types

### LINKED ISSUE #8595: Add PON ports to interface type
### NetBox version

v3.1.7

### Feature type

Data model extension

### Proposed functionality

Add EPON, GPON, X-GPON and other type of PON ports to interface type.

### Use case

These ports are used for OLTs in FTTH scenarios.

### Database changes

_No response_

### External dependencies

_No response_

DorianBourgeoisat:
### Closes: #8595:
Added the interface types as listed in #8595 for PON technologies.

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -814,6 +814,14 @@ class InterfaceTypeChoices(ChoiceSet):
     # ATM/DSL
     TYPE_XDSL = 'xdsl'
 
+    # PON
+    TYPE_GPON = 'gpon'
+    TYPE_XG_PON = 'xg-pon'
+    TYPE_XGS_PON = 'xgs-pon'
+    TYPE_TWDM_PON = 'twdm-pon'
+    TYPE_EPON = 'epon'
+    TYPE_10G_EPON = '10g-epon'
+
     # Stacking
     TYPE_STACKWISE = 'cisco-stackwise'
     TYPE_STACKWISE_PLUS = 'cisco-stackwise-plus'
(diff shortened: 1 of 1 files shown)


jsenecal:
What about  NG-PON2?

DorianXGH:
What do you mean ?
It is included, with 
`(TYPE_TWDM_PON, 'TWDM-PON (NG-PON2) (4x10 Gbps)'),`
NG-PON2 and TWDM-PON are the same thing, according to wikipedia https://en.wikipedia.org/wiki/NG-PON2
Maybe we should invert NG-PON2 and TWDM-PON as it true that NG-PON2 is more used than TWDM-PON, but it depends on the manufacturer, Nokia/Alcatel uses TWDM-PON for example.

jsenecal:
> What do you mean ? It is included, with `(TYPE_TWDM_PON, 'TWDM-PON (NG-PON2) (4x10 Gbps)'),` NG-PON2 and TWDM-PON are the same thing, according to wikipedia https://en.wikipedia.org/wiki/NG-PON2 Maybe we should invert NG-PON2 and TWDM-PON in the name as it true that NG-PON2 is the most prevalent one, but it depends on the manufacturer, Nokia/Alcatel uses TWDM-PON for example.

I actually had deleted my comment before you replied ahaha, I didn't see NG-PON2 at first ben then saw it and removed my comment.

These look good to me, paired with the new cabling functionality in v3.3 they will be awesome.

vista-:
NG-PON2 is definitely the more well-known name of the technology, I glanced over it the first time as well.

Just 2-cents

vista-:
There's also XG-PON (10G down, 2.5G up) out there, but I'm not sure how widespread is the adoption of that standard, as it's been largely succeeded by XGS-PON.

jeremystretch:
Keep in mind that none of this is set in stone; we can add additional types in the future if the need arises. What I'm more concerned with is ensuring that whatever we _do_ add is accurate, to avoid having to alter the existing choices.

DorianXGH:
I've changed it from TWDM PON to NG PON2 as the main name, with TWDM in parentheses.
It is the only one with a 2nd name used in real life as far as I know.
But yeah, I've seen the deleted reply asking about the asymmetrical XG-PON, it's also in there :)

jeremystretch:
CI failures are unrelated to the change being made and have already been resolved in `develop`.

jeremystretch:
Thanks @DorianXGH and @jsenecal!

PR #10065, opened 2022-08-18
Fixes #10055: NAT Outside evaluations and links

### LINKED ISSUE #10055: Device Management IPv4 Address shows erroneous text "(NAT:)" when IP has no NAT associations
### NetBox version

v3.3.0

### Python version

3.8

### Steps to Reproduce

This is a **minor** / Display Bug

I am seeing this on *all* Device instances in my installation (which was upgraded from 3.2.8 to 3.3.0). 
It is also reproducible on a newly-created Device / Interface / IP combo.

1. Create a Device with at least one Interface
2. Set (or create) an IP Address on that Interface, and check the box to make it the "Primary IP for this device" (i.e. management address)
3. View the Device's overview page, note that in the "Management" box there is the text "(NAT:)" after the IP address

Screenshot from my own installation with a pre-existing device:
image

### Expected Behavior

Would not expect to see any information about NAT when the IP address has no NAT relationships defined. (Suspect this is related to the change in the `nat_outside` relation to return zero or more IPs instead of the past "None or one"?)

### Observed Behavior

UI displays spurious "(NAT:)" text when it shouldn't

atownson:
Fixes: #10055

Added loops for NAT Outside in IP Address (ipaddress.html), Device (device.html), and Virtual Machine (virtualmachine.html) templates, and fixed 'ipam.IPAddress.None' text in IP table (ip.py).

### DIFF
--- a/netbox/ipam/tables/ip.py
+++ b/netbox/ipam/tables/ip.py
@@ -56,14 +56,6 @@ VRF_LINK = """
 {% endif %}
 """
 
-NAT_OUTSIDE_LINK = """
-{% if record.nat_outside.count > 0 %}
-    {% for nat in record.nat_outside.all %}<a href="{{ nat.get_absolute_url }}">{{ nat.address.ip }}</a>{% if not forloop.last %}, {% endif %}{% endfor %}
-{% else %}
-    &mdash;
-{% endif %}
-"""
-
 
 #
 # RIRs
@@ -368,8 +360,8 @@ class IPAddressTable(TenancyColumnsMixin, NetBoxTable):
         orderable=False,
         verbose_name='NAT (Inside)'
     )
-    nat_outside = tables.TemplateColumn(
-        template_code=NAT_OUTSIDE_LINK,
+    nat_outside = tables.ManyToManyColumn(
+        linkify_item=True,
         orderable=False,
         verbose_name='NAT (Outside)'
     )
--- a/netbox/ipam/tables/ip.py
+++ b/netbox/ipam/tables/ip.py
@@ -56,6 +56,12 @@ VRF_LINK = """
 {% endif %}
 """
 
+NAT_OUTSIDE_LINK = """
+{% if record.nat_outside.count > 0 %}
+    {% for nat in record.nat_outside.all %}<a href="{{ nat.get_absolute_url }}">{{ nat.address.ip }}</a>{% if not forloop.last %}, {% endif %}{% endfor %}
+{% endif %}
+"""
+
 
 #
 # RIRs
@@ -360,8 +366,8 @@ class IPAddressTable(TenancyColumnsMixin, NetBoxTable):
         orderable=False,
         verbose_name='NAT (Inside)'
     )
-    nat_outside = tables.Column(
-        linkify=True,
+    nat_outside = tables.TemplateColumn(
+        template_code=NAT_OUTSIDE_LINK,
         orderable=False,
         verbose_name='NAT (Outside)'
     )
--- a/netbox/ipam/tables/ip.py
+++ b/netbox/ipam/tables/ip.py
@@ -59,6 +59,8 @@ VRF_LINK = """
 NAT_OUTSIDE_LINK = """
 {% if record.nat_outside.count > 0 %}
     {% for nat in record.nat_outside.all %}<a href="{{ nat.get_absolute_url }}">{{ nat.address.ip }}</a>{% if not forloop.last %}, {% endif %}{% endfor %}
+{% else %}
+    &mdash;
 {% endif %}
 """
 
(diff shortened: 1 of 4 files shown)


netbox/templates/dcim/device.html
```diff
@@ -179,8 +179,8 @@ <h5 class="card-header">
                                 <a href="{{ object.primary_ip4.get_absolute_url }}">{{ object.primary_ip4.address.ip }}</a>
                                 {% if object.primary_ip4.nat_inside %}
                                   (NAT for {{ object.primary_ip4.nat_inside.address.ip|linkify }})
-                                {% elif object.primary_ip4.nat_outside %}
-                                  (NAT: {{ object.primary_ip4.nat_outside.address.ip|linkify }})
+                                {% elif object.primary_ip4.nat_outside.count > 0 %}
```
jeremystretch, line 182: When checking whether an object exists, it's best practice to use `.exists()` rather than `.count()` as it's slightly more efficient and better conveys intent.

```suggestion
                                {% elif object.primary_ip4.nat_outside.exists %}
```

netbox/templates/dcim/device.html
```diff
@@ -194,8 +194,8 @@ <h5 class="card-header">
                                 <a href="{{ object.primary_ip6.get_absolute_url }}">{{ object.primary_ip6.address.ip }}</a>
                                 {% if object.primary_ip6.nat_inside %}
                                   (NAT for {{ object.primary_ip6.nat_inside.address.ip|linkify }})
-                                {% elif object.primary_ip6.nat_outside %}
-                                  (NAT: {{ object.primary_ip6.nat_outside.address.ip|linkify }})
+                                {% elif object.primary_ip6.nat_outside.count > 0 %}
```
jeremystretch, line 197:
```suggestion
                                {% elif object.primary_ip6.nat_outside.exists %}
```

netbox/templates/virtualization/virtualmachine.html
```diff
@@ -45,8 +45,8 @@ <h5 class="card-header">
                             <a href="{% url 'ipam:ipaddress' pk=object.primary_ip4.pk %}">{{ object.primary_ip4.address.ip }}</a>
                             {% if object.primary_ip4.nat_inside %}
                               (NAT for <a href="{{ object.primary_ip4.nat_inside.get_absolute_url }}">{{ object.primary_ip4.nat_inside.address.ip }}</a>)
-                            {% elif object.primary_ip4.nat_outside %}
-                              (NAT: <a href="{{ object.primary_ip4.nat_outside.get_absolute_url }}">{{ object.primary_ip4.nat_outside.address.ip }}</a>)
+                            {% elif object.primary_ip4.nat_outside.count > 0 %}
```
jeremystretch, line 48:
```suggestion
                            {% elif object.primary_ip4.nat_outside.exists %}
```

netbox/templates/virtualization/virtualmachine.html
```diff
@@ -60,8 +60,8 @@ <h5 class="card-header">
                             <a href="{% url 'ipam:ipaddress' pk=object.primary_ip6.pk %}">{{ object.primary_ip6.address.ip }}</a>
                             {% if object.primary_ip6.nat_inside %}
                               (NAT for <a href="{{ object.primary_ip6.nat_inside.get_absolute_url }}">{{ object.primary_ip6.nat_inside.address.ip }}</a>)
-                            {% elif object.primary_ip6.nat_outside %}
-                              (NAT: <a href="{{ object.primary_ip6.nat_outside.get_absolute_url }}">{{ object.primary_ip6.nat_outside.address.ip }}</a>)
+                            {% elif object.primary_ip4.nat_outside.count > 0 %}
```
jeremystretch, line 63: Careful with the copy/pastes. :slightly_smiling_face: 

```suggestion
                            {% elif object.primary_ip6.nat_outside.exists %}
```

netbox/ipam/tables/ip.py
```diff
@@ -360,8 +368,8 @@ class IPAddressTable(TenancyColumnsMixin, NetBoxTable):
         orderable=False,
         verbose_name='NAT (Inside)'
     )
-    nat_outside = tables.Column(
-        linkify=True,
+    nat_outside = tables.TemplateColumn(
```
jeremystretch, line 371: We can use [ManyToManyColumn](https://github.com/netbox-community/netbox/blob/646272f9b39aec3edf58f1c59c9f72821ba099ac/netbox/netbox/tables/columns.py#L79) here instead of creating a custom template for the column.

jeremystretch (changes requested):
Thanks for working on this! Found a couple minor things, but looks great!

netbox/ipam/tables/ip.py
atownson: The ManyToManyColumn doesn't seem to function correctly here.
```
nat_outside = tables.ManyToManyColumn(
        linkify=True,
        orderable=False,
        verbose_name='NAT (Outside)'
    )
```
This produces the correct value in the table, but the links are not correct. They link to the IP address instead of the NAT address clicked.
image

atownson: Ah, got it -> linkify_item=True,

jeremystretch (approved):

jeremystretch:
Thanks @atownson!

PR #10103, opened 2022-08-22
Fix #10094 - Contact Assignments Create and Add Another

### LINKED ISSUE #10094: Contact Assignments Create and Add Another
### NetBox version

v3.3.0

### Python version

3.8

### Steps to Reproduce

1. Click 'Add a Contact' on a Device or Virtual Machine (possibly others) detail page.
2. Enter Contact and Role.
3. Click 'Create & Add Another'

### Expected Behavior

Returns to the Contact Assignment Page for the previous Device/Virtual Machine.

### Observed Behavior

Contact is created but redirects to /tenancy/contact-assignments/add/ without the content_type and object_id parameters resulting in Page Not Found.
image

arthanson:
### Fixes: #10094

Fixes for Contact Assignment create and Add Another (previously returning a 404 error:

Steps to Reproduce
Click 'Add a Contact' on a Device or Virtual Machine (possibly others) detail page.
Enter Contact and Role.
Click 'Create & Add Another'

adds a get_extra_addanother_params() to ObjectEditView that can be overridden to set extra query params for the button.

### DIFF
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -327,7 +327,7 @@ class ObjectEditView(GetReturnURLMixin, BaseObjectView):
         """
         return obj
 
-    def get_extra_addanother_params(self, request, params):
+    def get_extra_addanother_params(self, request):
         """
         Return a dictionary of extra parameters to use on the Add Another button.
         """
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -327,7 +327,10 @@ class ObjectEditView(GetReturnURLMixin, BaseObjectView):
         """
         return obj
 
-    def get_extra_addanother_params(self, request, params: dict):
+    def get_extra_addanother_params(self, request, params):
+        """
+        Return a QueryDict of extra params to use on the Add Another button.
+        """
         return params
 
     #
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -327,6 +327,9 @@ class ObjectEditView(GetReturnURLMixin, BaseObjectView):
         """
         return obj
 
+    def get_extra_addanother_params(self, request, params: dict):
+        return params
+
     #
     # Request handlers
     #
@@ -399,6 +402,7 @@ class ObjectEditView(GetReturnURLMixin, BaseObjectView):
 
                     # If cloning is supported, pre-populate a new instance of the form
                     params = prepare_cloned_fields(obj)
+                    params = self.get_extra_addanother_params(request, params)
                     if params:
                         if 'return_url' in request.GET:
                             params['return_url'] = request.GET.get('return_url')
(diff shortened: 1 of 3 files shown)


netbox/netbox/views/generic/object_views.py
```diff
@@ -399,6 +405,7 @@ def post(self, request, *args, **kwargs):
 
                     # If cloning is supported, pre-populate a new instance of the form
                     params = prepare_cloned_fields(obj)
+                    params = self.get_extra_addanother_params(request, params)
```
jeremystretch, line 408: Rather than passing `params` for each call, we can just update `params` with the output of `get_extra_addanother_params()`:

```suggestion
                    params.update(self.get_extra_addanother_params(request))
```

netbox/netbox/views/generic/object_views.py
```diff
@@ -327,6 +327,12 @@ def alter_object(self, obj, request, url_args, url_kwargs):
         """
         return obj
 
+    def get_extra_addanother_params(self, request, params):
+        """
+        Return a QueryDict of extra params to use on the Add Another button.
+        """
+        return params
```
jeremystretch, line 334:
```suggestion
    def get_extra_addanother_params(self, request):
        """
        Return a dictionary of extra parameters to use on the Add Another button.
        """
        return {}
```

netbox/tenancy/views.py
```diff
@@ -365,6 +366,15 @@ def alter_object(self, instance, request, args, kwargs):
...
         return instance
 
+    def get_extra_addanother_params(self, request, params: dict):
+        if not params:
+            params = QueryDict(mutable=True)
+
+        params['content_type'] = request.GET.get('content_type')
+        params['object_id'] = request.GET.get('object_id')
+
+        return params
```
jeremystretch, line 376:
```suggestion
    def get_extra_addanother_params(self, request):
        return {
            'content_type': request.GET.get('content_type'),
            'object_id': request.GET.get('object_id'),
        }
```

jeremystretch (changes requested):
I like the approach of adding a `get_extra_addanother_params()` method to the view, but we want to keep it as simple as possible because it will need to be added to the [plugins API](https://docs.netbox.dev/en/stable/plugins/development/views/#netbox.views.generic.object_views.ObjectEditView) for the view.

netbox/netbox/views/generic/object_views.py
```diff
@@ -327,6 +327,12 @@ def alter_object(self, obj, request, url_args, url_kwargs):
         """
         return obj
 
+    def get_extra_addanother_params(self, request, params):
```
jeremystretch, line 330: Need to drop `params` from the default signature.

```suggestion
    def get_extra_addanother_params(self, request):
```

jeremystretch (changes requested):

netbox/netbox/views/generic/object_views.py
arthanson: doh! should have caught that.

PR #10115, opened 2022-08-23
#10033 disable Add a Termination button if 2 terminations on L2VPN P2P

### LINKED ISSUE #10033: Disable "Add termination" button for P2P L2VPNs which already have 2 terminations
### NetBox version

v3.3-beta2

### Feature type

### Proposed functionality

Netbox enforces at model level that a P2P L2VPN can only have two terminations.

However in the GUI, if you are viewing a P2P L2VPN e.g. `/ipam/l2vpns/1/`, and it already has 2 terminations, then clicking "Add Termination" still takes you to the Add Termination page, allowing you to fill in termination information. It's only when you try to save it that you're told you can't have more than 2 terminations.

I propose that either you should get an error immediately on clicking the Add Termination button, or that the button should be greyed out (with a tooltip saying that you cannot have more than 2 terminations on a P2P L2VPN)

### Use case

User experience: not prompting the user to enter data which you know in advance will be rejected.

### Database changes

None

### External dependencies

None

arthanson:
### Fixes: #10033 

Disables the "Add a Termination" button if there are already 2 terminations for an L2VPN P2P

### DIFF
--- a/netbox/ipam/models/l2vpn.py
+++ b/netbox/ipam/models/l2vpn.py
@@ -3,6 +3,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.urls import reverse
+from django.utils.functional import cached_property
 
 from ipam.choices import L2VPNTypeChoices
 from ipam.constants import L2VPN_ASSIGNMENT_MODELS
@@ -67,6 +68,7 @@ class L2VPN(NetBoxModel):
     def get_absolute_url(self):
         return reverse('ipam:l2vpn', args=[self.pk])
 
+    @cached_property
     def can_add_termination(self):
         if self.type in L2VPNTypeChoices.P2P and self.terminations.count() >= 2:
             return False
--- a/netbox/ipam/models/l2vpn.py
+++ b/netbox/ipam/models/l2vpn.py
@@ -67,6 +67,12 @@ class L2VPN(NetBoxModel):
     def get_absolute_url(self):
         return reverse('ipam:l2vpn', args=[self.pk])
 
+    def can_add_termination(self):
+        if self.type in L2VPNTypeChoices.P2P and self.terminations.count() >= 2:
+            return False
+        else:
+            return True
+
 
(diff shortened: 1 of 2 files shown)


netbox/ipam/models/l2vpn.py
```diff
@@ -67,6 +67,12 @@ def __str__(self):
     def get_absolute_url(self):
         return reverse('ipam:l2vpn', args=[self.pk])
 
+    def can_add_termination(self):
```
jeremystretch, line 70: We probably want to wrap this with [`@cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) to avoid repeated database queries for each call.

netbox/templates/ipam/l2vpn.html
```diff
@@ -59,7 +59,7 @@ <h5 class="card-header">Terminations</h5>
       </div>
       {% if perms.ipam.add_l2vpntermination %}
         <div class="card-footer text-end noprint">
-          <a href="{% url 'ipam:l2vpntermination_add' %}?l2vpn={{ object.pk }}&return_url={{ object.get_absolute_url }}" class="btn btn-primary btn-sm">
+          <a href="{% url 'ipam:l2vpntermination_add' %}?l2vpn={{ object.pk }}&return_url={{ object.get_absolute_url }}" class="btn btn-primary btn-sm{% if not object.can_add_termination %} disabled{% endif %}"{% if not object.can_add_termination %} aria-disabled="true"{% endif %}>
```
jeremystretch, line 62: Can we simplify this a bit to avoid repeating the `{% if %}` block? I know the HTML is a little ugly but it eliminates the second evaluation. Alternatively, we could repeat the entire button element for better readability (once enable and once disabled).

```suggestion
          <a href="{% url 'ipam:l2vpntermination_add' %}?l2vpn={{ object.pk }}&return_url={{ object.get_absolute_url }}" class="btn btn-primary btn-sm{% if not object.can_add_termination %} disabled" aria-disabled="true{% endif %}">
```

jeremystretch (changes requested):

PR #10116, opened 2022-08-23
#9935 add new wireless choices for interfaces

### LINKED ISSUE #9935: Add 802.11ay and "Other Wireless" interface types
### NetBox version

v3.2.7

### Feature type

Data model extension

### Proposed functionality

I am proposing the addition of two types to the Interface model.

[IEEE 802.11ay](https://en.wikipedia.org/wiki/IEEE_802.11ay) is a defined standard (it's an evolution of 802.11ad, which already exists in the list).

I would also like to propose the addition of some sort of "Other (Wireless)" placeholder type, to enable users to document that an interface is some form of "Wireless" rather than just "Other" (which implies one of the cabled types, rather than an RF interface).

### Use case

802.11ay will benefit anybody who has this kind of equpment.

"Other (Wireless)" will give a user a way to cover any other type of proprietary RF interface that isn't already covered by NetBox (e.g. Siklu 8010 P2P radios, which run in the 70-80GHz E-band spectrum at 10Gbps). Without an "Other (Wireless)" option for Interface Type, the "Other" interface must be used which is treated like a cabled interface rather than an RF one.

### Database changes

Uncertain but suspect nothing major is required here - two new types that are both classified as RF/Wireless instead of Cabled.

### External dependencies

n/a

arthanson:
### Fixes: #9935 

Adds 802.11ay and "Other Wireless" interface types.


netbox/dcim/choices.py
```diff
@@ -790,7 +790,9 @@ class InterfaceTypeChoices(ChoiceSet):
     TYPE_80211AC = 'ieee802.11ac'
     TYPE_80211AD = 'ieee802.11ad'
     TYPE_80211AX = 'ieee802.11ax'
+    TYPE_80211AY = 'ieee802.11ay'
     TYPE_802151 = 'ieee802.15.1'
+    TYPE_OTHER_WIRELESS = 'other-Wireless'
```
jeremystretch, line 795: Raw values for the types should typically be lowercase.

```suggestion
    TYPE_OTHER_WIRELESS = 'other-wireless'
```

jeremystretch (changes requested):
Could you add this type to `WIRELESS_IFACE_TYPES` in `constants.py` too please?

netbox/dcim/choices.py
arthanson: @jeremystretch updated and added to constants.py.  Did generate a migration, but you said it shouldn't have?

jeremystretch: Disregard; the migration is expected because we're referencing the `WIRELESS_IFACE_TYPES` constant from `limit_choices_to` under the WirelessLink model. I've opened #10136 to address this, but this PR is fine as-is.

Issue #8113, opened 2021-12-19
Make slug fields optional for all models


jeremystretch:
### NetBox version

v3.1.1

### Feature type

Change to existing functionality

### Proposed functionality

For all 22 models which currently have a required [slug field](https://docs.djangoproject.com/en/4.0/ref/models/fields/#slugfield), make the field optional and remove the uniqueness requirement. This will make the slug field no longer required in UI forms and REST API serializers.

### Use case

Slugs are no longer used as unique identifiers in NetBox URLs, and no longer serve any native function. #8036 proposed removing slug fields entirely, however some users have expressed opposition to this idea. As an alternative, making slugs optional will preserve their utility for those users while alleviating the undue burden they place on users who do not.

Users who wish to still enforce slugs as required fields can employ simple [custom validation rules](https://netbox.readthedocs.io/en/stable/customization/custom-validation/) to do so.

### Database changes

A migration will be introduced within each app altering all `models.SlugField` instances to set `blank=True`.

candlerb:
I have no objection, except I think they should remain unique (per model).

jeremystretch:
@candlerb please elaborate on why.

candlerb:
If a slug has any value, then it's because it's a unique identifier. If I know object X has slug Y, then I can search for `/foo/?slug=Y` and be sure I get exactly zero or one.

The field already has a unique constraint, so why remove it?  People who don't want to use slugs can set them to null.

kkthxbye-code:
For automation we have the id, which is a unique identifier. So the usecase is a human searchable unique key with constraints regarding charset? And name with a custom constraint is not enough?

sol1-matt:
@kkthxbye-code 
> For automation we have the id, which is a unique identifier. So the use case is a human searchable unique key with constraints regarding charset? And name with custom constraints is not enough?

ID's are good for machines but not as good for people, often automation is building something that humans will look at. 

Example 1
If the automation you are using does something like use templates to create relationships then you end up with a bunch of templates named `123` which are terrible at describing the template itself. The current slug means your unique link between objects is also human readable eg: `the-earth`.

Example 2
In the filter example from @candlerb a human will likely know a slug name but they won't know the id. So using a slug in a filter is straight forward where as using the id will require something to lookup the id. 

I know I've run into this with id's in Netbox and why I've built a lot of automation around slugs, it simplifies things greatly with the same safety.


@candlerb 
> except I think they should remain unique (per model).

[...]

PR #10119, opened 2022-08-23
fixes for #10070 make l2vpn slug unique

### LINKED ISSUE #10070: Slug for L2VPN has no unique constraint
### NetBox version

v3.3.0

### Python version

3.9

### Steps to Reproduce

1. Create a L2VPN with name `Test` and slug `test`
2. Create a L2VPN with name `Test2` and slug `test`

### Expected Behavior

I expect a warning to be displayed when trying to create the second L2VPN with the same slug as the one before. The slug should be an unique attribute. It is unique for other models i.e. Tenants.

### Observed Behavior

The second L2VPN is created without a warning or an error. Two L2VPN with the same slug exist at the same time.

arthanson:
### Fixes: #10070

Adds unique constraint to L2VPN slug field along with max_length.  Note: Would cause issues migrating if someone already has non-unique slugs defined (probably fairly unlikely).


jeremystretch:
Looks like a few of the tests are failing simply because they're creating L2VPNs without slug values:

https://github.com/netbox-community/netbox/blob/deee36651d1a2cf464a2ed2ec932f066d8cbeab7/netbox/ipam/tests/test_api.py#L975-L980

jeremystretch:
@arthanson I've just pushed a tweak to `develop` to add slugs for the affected tests. If you rebase this branch it should pass CI now.

DanSheps:
Would it make sense to make slugs unique but non-required?

jeremystretch:
Every model which includes a `slug` field currently requires that field to be populated. Barring any changes per #8113, let's keep consistent with that.

PR #10137, opened 2022-08-24
Fix SSO signon for SAML with idp

### LINKED ISSUE #10135: SSO doesn't work for SAML accounts that require idp
### NetBox version

v3.3.0

### Python version

3.10

### Steps to Reproduce

1. Setup SAML SSO integration that has idp (see: https://python-social-auth.readthedocs.io/en/latest/backends/saml.html#basic-usage)
2. Use the SSO signin link on the main NetBox login page

### Expected Behavior

The SSO link should work and contain an idp query param

### Observed Behavior

The provider will show an error page as the idp isn't given in the QueryParam

arthanson:
### Fixes: #10135 

SAML signin usually requires passing an idp in the query param, this updates so if doing SAML sign-in and have SOCIAL_AUTH_SAML_ENABLED_IDPS defined it will add those to the sing-in (see screenshot).  Also refactors the auth_backends context to be more extensible.

saml

### DIFF
--- a/netbox/users/views.py
+++ b/netbox/users/views.py
@@ -17,7 +17,7 @@ from social_core.backends.utils import load_backends
 
 from extras.models import ObjectChange
 from extras.tables import ObjectChangeTable
-from netbox.authentication import get_auth_backend_display
+from netbox.authentication import get_auth_backend_display, get_saml_idps
 from netbox.config import get_config
 from utilities.forms import ConfirmationForm
 from .forms import LoginForm, PasswordChangeForm, TokenForm, UserConfigForm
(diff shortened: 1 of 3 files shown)


netbox/users/views.py
```diff
@@ -46,9 +46,18 @@ def get(self, request):
             logger = logging.getLogger('netbox.auth.login')
             return self.redirect_to_next(request, logger)
 
-        auth_backends = {
-            name: get_auth_backend_display(name) for name in load_backends(settings.AUTHENTICATION_BACKENDS).keys()
-        }
+        auth_backends = {}
+        for name in load_backends(settings.AUTHENTICATION_BACKENDS).keys():
```
jeremystretch, line 50: Let's refactor this to add a new backend to `auth_backends` for each IDP. Then we can just iterate through the backends list in the template without the additional complexity of checking for `saml_idps`.

It's probably also worth rewriting `get_auth_backend_display()` to just return the dictionary of display name and icon directly.

jeremystretch (changes requested):

Issue #10039, opened 2022-08-16
Interfaces table, dropdown menu: "Add FHRP Group" option


candlerb:
### NetBox version

v3.3-beta2

### Feature type

Change to existing functionality

### Proposed functionality

In the interfaces table, "Add" is now a drop-down:
![image](https://user-images.githubusercontent.com/44789/184943179-18b99669-2b72-4513-a9b2-fe69454ed806.png)

I propose adding another option to this list, "Add FHRP Group"

Similar to IP addresses, ideally this would take you to a page with two tabs:
* One for Create FHRP Group (and assign it to this interface)
* One for Assign existing FHRP Group to this interface

An alternative (and slightly simpler) feature is to add option "Assign FHRP Group" instead. This would take you to another page, which would only give you the option to Assign an existing FHRP group to this interface.

I think the latter option is fine, because normally you'd create an FHRP group, before you go around assigning it to interfaces.

### Use case

There is already a way to assign FHRP groups to interfaces but it's awkward and non-obvious (you have to *View* the interface, at which point one of the panels is a table of FHRP groups, with a button to assign another one)

Having the option to do it directly on the interface "Add" dropdown makes it consistent with assigning other things (IP addresses, Inventory items etc)

### Database changes

None

### External dependencies

None

PR #10140, opened 2022-08-24
Add "Assign FHRP Group" to device-interfaces context menu

### LINKED ISSUE #10038: Interfaces table, dropdown menu: "Add L2VPN Termination" option
### NetBox version

v3.3-beta2

### Feature type

### Proposed functionality

In the interfaces table, "Add" is now a drop-down:
image

I propose adding another option to this list, "Add L2VPN termination"

This will take you to the "Add L2VPN Termination" page, with the current device (or VM) interface selected.

### Use case

UX: easier way to add a (sub)interface to an L2VPN, e.g. as a pseudowire endpoint.

### Database changes

None

### External dependencies

None

arthanson:
### Fixes: #10038

Adds an "Assign FHRP Group" to the context menu of devices-interfaces (see screenshot).  Couple potential changes/questions:

1. This is an add button, so I'm not sure of putting an "assignment" in here if there is a convention?  Potentially rename it to "FHRP Group - Assign"
2. The feature request also had "Add FHRP Group" going to a separate page that had tabs for each of add and assign.  I didn't see any existing convention for this type of page currently?  Would require refactoring of the code to avoid code-duplication.
3. could add "Add FHRP Group" as a separate menu item - but that isn't tied at all to the particular interface so seems like an over-loading of the context menu?

assign-fhrp

### DIFF
--- a/netbox/dcim/tables/template_code.py
+++ b/netbox/dcim/tables/template_code.py
@@ -241,6 +241,9 @@ INTERFACE_BUTTONS = """
       {% if perms.dcim.add_interface %}
         <li><a class="dropdown-item" href="{% url 'dcim:interface_add' %}?device={{ record.device_id }}&parent={{ record.pk }}&name_pattern={{ record.name }}.&type=virtual&return_url={% url 'dcim:device_interfaces' pk=object.pk %}">Child Interface</a></li>
       {% endif %}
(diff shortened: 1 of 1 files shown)


jeremystretch:
Is this supposed to be for #10038 or #10039?

PR #10171, opened 2022-08-26
#10170 optimize change log queries

### LINKED ISSUE #10170: Change Log list view has inefficient query
### Proposed Changes

need to add some select_related and prefetch_related to the Change Log list view to reduce the number of SQL queries.

### Justification

On the demo site noticed the Change log list view was somewhat slow, checking in the debug toolbar with demo data saw it was generating over 100 SQL queries.  For longer lists it will just get slower.

arthanson:
### Fixes:  #10170

Adds some better accessors to ObjectChangeTable to make the Change Log better performing.


jeremystretch (changes requested):
Hopefully this makes sense. I realize it's a bit convoluted.

Issue #9329, opened 2022-05-10
Netbox api call /dcim/devices/ virtual chassis does not have "member_count"


PaulWestphal:
### NetBox version

v3.2.2

### Python version

3.8

### Steps to Reproduce

1. Create any number of devices and a virtual chassis.
2. Add the devices to the virtual chassis.
3. Send a GET HTTP request to http://netbox-address-here/api/dcim/devices/

### Expected Behavior

According to the documentation, the "virtual_chassis" of each of the created devices should have a field "member_count".

### Observed Behavior

There is no "member_count".

DanSheps:
You need to query the virtual chassis, not the device.

PaulWestphal:
The example value from /dcim/devices/ has the member_count in it.

```
{

  "count": 0,

  "next": "string",

  "previous": "string",

  "results": [

    {

      "id": 0,

      "url": "string",

      "display": "string",

      "name": "string",

      "device_type": {

        "id": 0,

        "url": "string",

        "display": "string",

        "manufacturer": {

          "id": 0,

          "url": "string",

          "display": "string",

          "name": "string",

          "slug": "string",

          "devicetype_count": 0

       [...]

      "virtual_chassis": {

        "id": 0,

        "url": "string",

        "display": "string",

        "name": "string",

        "master": {

          "id": 0,

          "url": "string",

          "display": "string",

          "name": "string"

        },

        "member_count": 0  <-----------

      },

      "vc_position": 0,

      "vc_priority": 0,

      "comments": "string",

      "local_context_data": "string",

     [...]

      },

      "created": "2022-05-10T14:42:55.190Z",

      "last_updated": "2022-05-10T14:42:55.190Z"

    }

  ]

}

```

DanSheps:
You are correct, it is on the serializer.

I am currently looking into this as it may be related to the field not being annotated when the query resolves.

jeremystretch:
This isn't a bug. The `member_count` field is omitted from the serializer because that data is not available when querying the devices list. (Annotating it would require unduly expensive queries.) It is intended for use and present when retrieving virtual chassis directly in either full or brief mode:

```
GET /api/dcim/virtual-chassis/?brief=True

{
    "count": 4,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 1,
            "url": "http://netbox:8000/api/dcim/virtual-chassis/1/",
            "display": "r105-torsw",
            "name": "r105-torsw",
            "master": {
[...]

PR #10186, opened 2022-08-29
Fixes #9592: Remove member_count from NestedVirtualChassisSerializer

amhn:
### Fixes: #9592

This removes the member_count field from the NestedVirtualChassisSerializer and the corresponding test.


jeremystretch (commented):
It would not be appropriate to remove the `member_count` field from NestedVirtualChassisSerializer, as this is a valid and documented part of the API when retrieving virtual chassis. The problem raised in #9329 is that the count data is not available under other views (e.g. when listing devices), so a separate nested serializer needs to be defined for that purpose.

jeremystretch:
Thanks for your interest in taking this on. However, per my comment [here](https://github.com/netbox-community/netbox/issues/9592#issuecomment-1230387051) I'd like to try a different approach to resolve the underlying issue.

PR #10187, opened 2022-08-29
9895 remove json field inspector

### LINKED ISSUE #9895: swagger/ipam/ip-addresses/: Invalid type for "assigned_object"
### NetBox version

v3.2.6

### Python version

3.8

### Steps to Reproduce

Hi there,

the netbox swagger.yml states that the IP address list endpoint `/ipam/ip-addresses/` returns a value of type map[string]string for key `assigned_object ` instead of the correct type object. This causes issues for me since I use an auto-generated api client which will error out in case it receives something _not string-like_ - in this case e.g. a number as id.

Another similar issue has been opened here already: https://github.com/netbox-community/go-netbox/issues/133

What can we do about that? :)

Thanks a lot!

### Expected Behavior

Correct type stated in swagger.yml

### Observed Behavior

Incorrect type stated in swagger.yml

amhn:
### Fixes: #9895

This removes the (unused) custom JSONFieldInspector in favor of the one in drf_yasg.

Fields with serializers.DictField are changed to serializers.JSONField. This change fixes #9895 for ipam/ip-addresses and other objects. This allows an arbitrary object as the respective field instead of a dict of strings.

The third commit fixes the type of the _occupied field for NestedSerializers as mentioned in https://github.com/netbox-community/netbox/issues/9895#issuecomment-1203166847

### DIFF
--- a/netbox/dcim/api/nested_serializers.py
+++ b/netbox/dcim/api/nested_serializers.py
@@ -316,6 +316,7 @@ class NestedModuleSerializer(WritableNestedSerializer):
 class NestedConsoleServerPortSerializer(WritableNestedSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='dcim-api:consoleserverport-detail')
     device = NestedDeviceSerializer(read_only=True)
+    _occupied = serializers.BooleanField(required=False, read_only=True)
 
     class Meta:
         model = models.ConsoleServerPort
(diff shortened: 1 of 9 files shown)


jeremystretch:
Here's a diff between the API spec before and after the change.

[10187_diff.txt](https://github.com/netbox-community/netbox/files/9455509/10187_diff.txt)

jeremystretch (approved):
Thanks for taking this on @amhn!

PR #10215, opened 2022-08-30
#10060 add journal entry to global search

### LINKED ISSUE #10060: Include journal entries in global search
### NetBox version

v3.2.7

### Feature type

### Proposed functionality

When a search term is entered in the top-global search function, the contents of the journal should be searched as well.

### Use case

When I leave (for example) an issue number in a journal entry, it would be nice if I later were able to find all devices with that reference number in their journal. This way one can quickly see which devices were involved in an issue/incident.

### Database changes

As far as I can tell: no changes to the DB.

### External dependencies

No dependencies either.

arthanson:
### Fixes: #10060

Include journal entries in global search. @jeremystretch best guess as to how to add this following the other items, let me know if I missed something.
jsearch

### DIFF
--- a/netbox/netbox/search.py
+++ b/netbox/netbox/search.py
@@ -2,6 +2,8 @@ import circuits.filtersets
 import circuits.tables
 import dcim.filtersets
 import dcim.tables
+import extras.filtersets
+import extras.tables
 import ipam.filtersets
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Great, thanks!

Issue #9903, opened 2022-08-02
Implement a mechanism for maintaining denormalized (cached) database fields


jeremystretch:
### NetBox version

v3.3-beta1

### Feature type

New functionality

### Proposed functionality

Establish a mechanism for registry denormalized model fields within the application so that they can be kept up-to-date. (These are fields which store cached values from elsewhere in the database.) The registry can be used to track these fields, and we can leverage Django signals to trigger automatic updates in the event of a change.

#9902 could also help with this effort, although I don't believe it's strictly required.

### Use case

Sometimes it is necessary to denormalize certain database fields for performance optimization and/or to enable filtering. For example, the CableTermination model introduced in v3.3 caches the device, rack, location, and site associated with its termination object.

### Database changes

This feature should not require any database changes itself.

Issue #10196, opened 2022-08-29
Identify and standardize all instances of database denormalization


jeremystretch:
### Proposed Changes

A [new set of utilities](https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/denormalized.py) was introduced under #9102 to better facilitate the caching of statistical data based on foreign key relationships. Currently, these are employed only for some fields on the CableTermination model, but they can be readily used elsewhere. This issue seeks to track the identification of such uses for the purposes of standardizing how we denormalize in NetBox.

Current and future instances of denormalization are tracked below:

* circuits.Circuit: `termination_a`, `termination_b`
* dcim.Interface: `wireless_link`
* #6347
* #7845

### Justification

Moving to a standardized denormalization implementation will ease long-term maintenance and reduce the likelihood of introducing new bugs.

PR #10246, opened 2022-09-01
#9481 update child location whem move parent

### LINKED ISSUE #9481: Child location does not update when Parent location changes
### NetBox version

v3.2.2

### Python version

3.9

### Steps to Reproduce

1. Create more than one location, and populate each location with at least one rack with non-zero height of any type.
2. Create a parent-capable device-type with non-zero height. Add a device-bay to the parent-capable device-type.
3. Create an instance of the parent-capable device-type, and insert the device into a rack of sufficient height.
4. Create a child-capable device-type with zero height. Then create an instance of the child-capable type.
5. Insert the child-capable device into the rack and location of the parent-device, without a position.
6. In the Device-Bays tab of the parent-device, associate the child-device with the open device-bay.
7. Move the parent device into a rack of sufficient height in some other location.

### Expected Behavior

The child-device's rack and location should update according to the move associated with the parent-device.

### Observed Behavior

The child-device's rack updated according to the move associated with the parent-device, but the location of the child-device remains as it was before the parent-device was moved.

Example result: The child could have assigned-Location A but be found in a rack associated with Location B.

Example impact: The child-device cannot be updated in the UI until the discrepancy between Assigned-Location and Rack-Location is resolved:
```
Errors: Rack - Rack A15 (A15) does not belong to location B.
```

arthanson:
### Fixes: #9481

See bug report - when parent device is moved to a new rack the child device location wasn't getting updated.  @jeremystretch you had mentioned "This may be a use case for https://github.com/netbox-community/netbox/issues/9903 (available in v3.3)." in the bug report, am I missing something for this - this seems to fix the issue?

### DIFF
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -868,6 +868,7 @@ class Device(NetBoxModel, ConfigContextModel):
         for device in devices:
             device.site = self.site
             device.rack = self.rack
+            device.location = self.location
             device.save()
 
     @property


jeremystretch:
Yep your fix is good. Re: #9903 I was just tagging this as an instance where we might be able to leverage the new denormalization tools (see #10196). Still something I need to look into further.

jeremystretch (approved):

PR #10267, opened 2022-09-06
Documentation entry for customers on different cloud endpoints (GCC High, DoD, etc)

Ldogg123:
Add an entry for customers on GCC High / DoD or other endpoints (China, Germany, etc.)


jeremystretch:
Thanks for the suggestion, however the documentation is not intended to capture all potential configurations. (The `python-social-auth` docs are already great.) It provided merely to serve as a reference for basic configuration. We cannot commit to maintaining documentation for extended configurations, particularly to services we're unable to test.

PR #10290, opened 2022-09-07
Fixes #10258 - Allow running scripts nested in modules/packages

### LINKED ISSUE #10258: Not possible to use custom scripts in nested packages
### NetBox version

v3.3.2

### Python version

3.10

### Steps to Reproduce

Create the following folder structure:

```
./scripts/__init__.py
./scripts/testscript.py
./scripts/subscriptfolder/__init__.py 
./scripts/subscriptfolder/myscript1.py
./scripts/subscriptfolder/myscript2.py
```

`testscript.py` contains script `class Test(Script):`

`myscript1.py` contains script `class MyScript1(Script):`

`myscript2.py` contains script `class MyScript2(Script):`

__init__.py contains:

```
from myscript1 import MyScript1
from myscript1 import MyScript2
```

### Expected Behavior

Three scripts should be shown on the script page:

TestScript under the testscript heading and MyScript1 and MyScript2 under the subscriptfolder heading.

All scripts should work as expected.

### Observed Behavior

All scripts show in the list as expected, but clicking on the subscriptsfolder scripts will cause a 404 error as the URL points to subscriptsfolder.myscript1.MyScript1 but the returned dict from `get_scripts()` is in the following format:

```
{
  "subscriptsfolder": {"MyScript1": MyScript1() }
}
```

Noticably missing the `myscript1` part of the module path.

This makes it hard to organize scripts as there is no way to get scripts under the same heading without having them in the same file.

kkthxbye-code:
### Fixes: #10258

The purpose of this FR is to allow a structure like this:

```
scripts
├── __init__.py
├── script0.py
└── testscriptfolder
    ├── __init__.py
    ├── script1.py
    ├── script2.py
    └── testnestedfolder
        ├── __init__.py
        └── script3.py
```

All scripts contain (with the numbers replaced):

```
from extras.scripts import *

class Script0(Script):
    def run(self, data, commit):
        self.log_success("Script0")
```

The `__init__.py` file in the `testscriptfolder` contains:

```
from .script1 import Script1
from .script2 import Script2
from .testnestedfolder.script3 import Script3
```

This setup results in the following script list with the PR:

image

This is achieved by splitting import paths so that the module_name contains the base module only, while the script name contains the rest of the path. For Script3 the import path would look like:

`testscriptfolder.testnestedfolder.script3` which would be split to:

```
module_name = "testscriptfolder"
script_name = "testnestedfolder.script3"
```

The advantage is that scripts become much easier to manage as before the only way to have scripts grouped under the same module was to have them all in the same file. With long scripts this instantly becomes unwieldy.

I tested every combination I could come up with, and as far as I can tell defining custom names for the scripts still work. Defining custom names for the module still works (done in the `__init__.py` file). Calling scripts from the API still works. Script ordering still works.

I'm not sure if this is the right approach. As far as I can tell it doesn't break anything, but it's hard to guarantee that people haven't found crazy ways around the limitations of script loading. I would really like some feedback and help testing here. Maybe there's a better way to achieve better script organization that I'm missing.

Another limitation is that I haven't touched reports here.

### DIFF
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -257,7 +257,7 @@ class ScriptViewSet(ViewSet):
     lookup_value_regex = '[^/]+'  # Allow dots
 
     def _get_script(self, pk):
-        module_name, script_name = pk.split('.')
+        module_name, script_name = pk.split('.', maxsplit=1)
         script = get_script(module_name, script_name)
         if script is None:
             raise Http404
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -159,7 +159,7 @@ class ReportViewSet(ViewSet):
         # Read the PK as "<module>.<report>"
         if '.' not in pk:
             raise Http404
-        module_name, report_name = pk.split('.', 1)
+        module_name, report_name = pk.split('.', maxsplit=1)
 
         # Raise a 404 on an invalid Report module/name
         report = get_report(module_name, report_name)
@@ -183,8 +183,8 @@ class ReportViewSet(ViewSet):
         }
 
         # Iterate through all available Reports.
-        for module_name, reports in get_reports():
-            for report in reports:
+        for module_name, reports in get_reports().items():
+            for report in reports.values():
 
                 # Attach the relevant JobResult (if any) to each Report.
                 report.result = results.get(report.full_name, None)
--- a/netbox/extras/management/commands/runreport.py
+++ b/netbox/extras/management/commands/runreport.py
@@ -21,8 +21,8 @@ class Command(BaseCommand):
         reports = get_reports()
 
         # Run reports
-        for module_name, report_list in reports:
-            for report in report_list:
+        for module_name, report_list in reports.items():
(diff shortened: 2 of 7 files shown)


jeremystretch:
LGTM! Thanks for sorting this out @kkthxbye-code.

> Another limitation is that I haven't touched reports here.

Do we want to expand the scope of #10258 to cover reports as well, and replicate the necessary changes in this PR, or treat that work as a separate issue?

kkthxbye-code:
@jeremystretch - We might as well get it all fixed. I'll investigate if there are any blockers for just doing the same fixes for reports, if not I'll amend the issue and PR to include that as well. I'll ping you when I got it all figured out.

kkthxbye-code:
I think this is good to go @jeremystretch unless I missed something. I really tried to thoroughly test everything, but there's a lot of leeway with regards to creating scripts. If you have some scripts/reports lying around, feel free to test it out quick before merge.

jeremystretch (approved):
Haven't found any issues loading either script or reports using the new approach. Thanks!

PR #10291, opened 2022-09-07
#8580 add interface filters for connected

### LINKED ISSUE #8580: Ability to filter interfaces by 'mark_connected' or those with a connected cable
### NetBox version

3.1.7

### Feature type

New functionality

### Proposed functionality

Add the ability to filter interfaces under /dcim/interfaces/ by those that are marked / not marked as 'mark_connected' 
Add the ability to filter interfaces under /dcim/interfaces/ by those that have / have not got a cable connected

### Use case

A large switch with multiple interfaces can be trick to sort/filter in the current GUI. E.g. I have a 9410 switch with 358 interfaces, the /dcim/devices/<device>/interfaces/ view allows me to fairly quickly see which are connected / not connected albeit with no way to filter.

However when looking at /dcim/interfaces/ I can add the Mark Connected and Cable columns but i'm unable to filter by interfaces that are marked connected or have a cable attached

### Database changes

_No response_

### External dependencies

_No response_

arthanson:
### Fixes: #8580

Adds filters to interface for:

**cabled** - Boolean filter returning only models which have a cable assigned
**connected** - Boolean filter for path endpoints which returns only models with a complete end-to-end path assigned
**is_occupied** - Boolean filter for objects which are cabled or which have mark_connected set to True

Added them as a new group to the filters page "Connection" at the end (see screenshot):

Monosnap Interfaces | NetBox 2022-09-07 15-44-18

@jeremystretch I didn't see any place in the docs that need to be updated for a new filter.  Do the API specs / GraphQL stuff need to be separately updated if adding a new filter?


netbox/dcim/filtersets.py
```diff
@@ -1304,6 +1304,9 @@ class InterfaceFilterSet(
         to_field_name='rd',
         label='VRF (RD)',
     )
+    is_occupied = django_filters.BooleanFilter(
```
jeremystretch, line 1307: This should go under CabledObjectFilterSet so that the filter is available for all object types to which a cable can be attached (those which inherit from CabledObjectModel).

netbox/dcim/forms/filtersets.py
```diff
@@ -1087,6 +1088,24 @@ class InterfaceFilterForm(DeviceComponentFilterForm):
...
+        widget=StaticSelect(
+            choices=BOOLEAN_WITH_BLANK_CHOICES
+        )
+    )
+    is_occupied = forms.NullBooleanField(
+        required=False,
+        widget=StaticSelect(
+            choices=BOOLEAN_WITH_BLANK_CHOICES
+        )
+    )
```
jeremystretch, line 1108: These should be added to the filter form for all cabled objects to be consistent.

jeremystretch (changes requested):
Thanks for this! Could you also add tests for the new filters in `dcim/tests/test_filtersets.py` and `circuits/test/test_filtersets.py`?

arthanson:
tests added.

PR #10312, opened 2022-09-09
Fixes #10247: Allow changing selected device/VM when creating a new component

### LINKED ISSUE #10247: Virtual machine selection on Add a new interface not recognised
### NetBox version

v3.3.1

### Python version

3.8

### Steps to Reproduce

When adding an interface to a Virtual Machine, the Virtual Machine selection drop down does not appear to register a change.

These are the steps I followed, there may be other permutations.

1. From the Virtual Machine list, Select a VM.
2. Click '+ Add Interfaces' button.
3. In the Virtual Machine dropdown list, select a different VM from the one pre-populated.
6. Enter a name in the 'name' field.
7. Click Create.
8. The created interface will be appear on the virtual machine that the form was instantiated from.

### Expected Behavior

If it's possible to change the virtual machine in the field, I would expect the interface to be created on VM selected in the 'Virtual Machine' field and not the VM from where the Add an interface was instantiated.

The process I followed was to create several VMs, via clone. The interface doesn't get cloned so I then went to create multiple interfaces using the "Create & Add Another" selecting another VM in the Add new interface form, without having to exit the form, select the next VM and then add an interface on the next VM.

### Observed Behavior

After creating the interface on a Virtual Machine, using the 'create and add another' selecting a different VM in the drop down, and entering the same interface name, a notification dialog popped up 'Name: this field is required'. I thought originally that the name entry was not being accepted, but then realised it was the Virtual Machine selection that was not functioning. The warning message was confusing, and I would have expected either the selected virtual machine to be recognised, an error similar to "Name: duplicate interface name" , rather than 'Name: this field is required' as the name field was populated.

jeremystretch:
### Fixes: #10247

This work affects all device components, all device component templates, and VM interfaces. At a high level, we've switched from rendering two individual forms for component creation to a single new per-model creation form, which inherits from both `ComponentCreateForm` as well as the pertinent model form. This enables us to efficiently render the form and ditch a number of custom templates.

`patterned_fields` on the component creation views has been replaced with `replication_fields`, which is now defined on the form for more robust validation. Additionally, `name_pattern` and `label_pattern` have been renamed to `name` and `label`, and now replace the model form fields of the same names.


jeremystretch:
I still need to fix bulk component creation (e.g. for multiple objects).

Issue #10172, opened 2022-08-26
Upgrade to Django 4.1


jeremystretch:
### Proposed Changes

Move from Django 4.0 to 4.1.

### Justification

The most appealing feature for us is probably automatic constraint validation, however this release also brings improved async support. Full release notes are [here](https://docs.djangoproject.com/en/4.1/releases/4.1/).

arthanson:
Probably need to wait for Django 4.1.1 the tests fail I think related to this issue: https://mirror.cnhub.dev/django/django/pull/15995

PR #10341, opened 2022-09-12
10172 upgrade django

arthanson:
### Fixes: #10172 

Updated to Django 4.1.1 - fixes several references in clean functions that 4.1 reports as errors now.

### DIFF
--- a/netbox/dcim/models/device_components.py
+++ b/netbox/dcim/models/device_components.py
@@ -953,12 +953,13 @@ class RearPort(ModularComponentModel, CabledObjectModel):
         super().clean()
 
         # Check that positions count is greater than or equal to the number of associated FrontPorts
-        frontport_count = self.frontports.count()
-        if self.positions < frontport_count:
-            raise ValidationError({
-                "positions": f"The number of positions cannot be less than the number of mapped front ports "
-                             f"({frontport_count})"
-            })
+        if self.pk:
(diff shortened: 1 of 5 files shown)


netbox/dcim/models/device_components.py
```diff
@@ -953,12 +953,13 @@ def clean(self):
         super().clean()
 
         # Check that positions count is greater than or equal to the number of associated FrontPorts
-        frontport_count = self.frontports.count()
-        if self.positions < frontport_count:
-            raise ValidationError({
-                "positions": f"The number of positions cannot be less than the number of mapped front ports "
-                             f"({frontport_count})"
-            })
+        if self.pk:
```
jeremystretch, line 956: Is this change related to [this Django bug](https://code.djangoproject.com/ticket/33952)? I don't have any problem with the change itself but I suspect there are similar instances elsewhere in the code that will need to be addressed.

arthanson: I don't think it is a Django "bug" it is more being more explicit about reporting an invalid condition.  This should probably only occur in the clean methods in our codebase - we are querying the m2m on the model when the pk isn't set yet, which doesn't really make any sense.  I think previously Django probably defaulted this to be None or 0 for count(), but now is explicitly raising that this shouldn't be done.

Issue #5507, opened 2020-12-21
Custom fields for IP addresses, secrets are not working/storing the data in v2.10.1


991jo:
### Environment
* Python version: 3.6.8
* NetBox version: 2.10.1

### Steps to Reproduce
1. Create a custom field for IP addresses (I have tested boolean and text fields)
2. add/edit an IP address and populate the field

### Expected Behavior
The custom fields are saved and shown in the Web UI/Django object

### Observed Behavior
The custom fields are empty. (The Django objects `.cf` is also an empty dict)

### Further Observations
Custom fields for other objects (e.g. Prefixes) work as intended.

bmhughes:
Getting the same result here as well.

jeremystretch:
Looks like secrets are also affected.

The problem is a missing call to `super().clean()` on the model forms; an easy fix. Thanks for the report!

PR #10347, opened 2022-09-12
5509 add test cases for custom fields

### LINKED ISSUE #5509: Extend view tests to include custom field data
### Proposed Changes
Extend the tests for UI views to include custom field data.

### Justification
This will provide more comprehensive testing and avoid bugs such as #5507.

arthanson:
### Fixes: #5509

Adds test case for checking custom fields, refactors some of the ViewTestCase base classes as the extras views do not include custom fields.


jeremystretch:
Thanks for digging into this @arthanson. The primary driver behind #5509 was to extend the view tests to include custom field data for each model that supports it, as part of the existing create & edit view tests. So for example, extending `dcim.tests.test_views.SiteTestCase` to create some custom fields along with the test sites, and populate data for these via the class' `form_data`. The intent is simply to ensure that custom field data is getting applied correctly for each object type (to avoid bugs such as #5507).

The most efficient way to achieve this IMO would probably be to have the relevant test classes inherit from a common parent that overrides `setUpTestData()` to automatically create one or more custom fields for each model and defines some values for these fields which can be included in `form_data` and `bulk_edit_data`. What do you think of this approach?

PR #10349, opened 2022-09-13
Fixes #10305: Allows null master in VirtualChassis APIs

### LINKED ISSUE #10305: Virtual Chassis master field cannot be null according to the API
### NetBox version

v3.2.9

### Python version

3.10

### Steps to Reproduce

1. Submit API write request to the VirtualChassis with "master": null in it

### Expected Behavior

Requests is accepted and the master is unset for the specific VirtualChassis

### Observed Behavior

HTTP 400 is returned with the message: `[{"master":["This field may not be null."]}]`

sasschary:
### Fixes: #10305 

VirtualChassis API calls were not allowing the `master` field to be `null`. This PR allows the field to be null and adds a test to verify that the field can be set to `null`.

### DIFF
--- a/netbox/dcim/api/serializers.py
+++ b/netbox/dcim/api/serializers.py
@@ -1076,7 +1076,7 @@ class CablePathSerializer(serializers.ModelSerializer):
 
 class VirtualChassisSerializer(NetBoxModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='dcim-api:virtualchassis-detail')
-    master = NestedDeviceSerializer(required=False)
+    master = NestedDeviceSerializer(required=False, allow_null=True, default=None)
     member_count = serializers.IntegerField(read_only=True)
 
     class Meta:
--- a/netbox/dcim/tests/test_api.py
+++ b/netbox/dcim/tests/test_api.py
@@ -2057,6 +2057,7 @@ class VirtualChassisTest(APIViewTestCases.APIViewTestCase):
 
         cls.bulk_update_data = {
             'domain': 'newdomain',
+            'master': None
         }
 
 


netbox/dcim/tests/test_api.py
```diff
@@ -2040,6 +2040,11 @@ def setUpTestData(cls):
             'master': devices[1].pk,
         }
 
+        cls.update_data = {
+            'name': 'Virtual Chassis 1',
+            'master': None,
+        }
+
```
jeremystretch, line 2047: `update_data` is already defined above, so this isn't needed. To test setting `master` to None, we can instead add `'master': None` to one or more of the object definitions under `bulk_update_data`.

netbox/dcim/api/serializers.py
```diff
@@ -1076,7 +1076,7 @@ def get_path(self, obj):
 
 class VirtualChassisSerializer(NetBoxModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='dcim-api:virtualchassis-detail')
-    master = NestedDeviceSerializer(required=False)
+    master = NestedDeviceSerializer(required=False, allow_null=True)
```
jeremystretch, line 1079: I think this also needs `default=None` as similar fields have defined.

```suggestion
    master = NestedDeviceSerializer(required=False, allow_null=True, default=None)
```

jeremystretch (changes requested):
Thanks for taking this on!

netbox/dcim/tests/test_api.py
sasschary: You'll have to excuse my ignorance when it comes to Django and how it does tests, I'm mostly familiar with Flask (plus am new to this codebase). Looking back now it should have been pretty obvious that this would just reassign `update_data` rather than adding an additional test that updates data...

Issue #4347, opened 2020-03-11
Add JSON/YAML import support for all objects


jeremystretch:
### Proposed Changes
Allow users to import objects in YAML or JSON format in addition to the existing support for CSV. YAML/JSON support was introduced for device types in #451; this extends that same functionality to all models.

### Justification
JSON and YAML offer greater flexibility than CSV in incorporating related instances, as well as referencing parent objects.

BarbarossaTM:
This is really cool, thanks!

Are there any plans to allow import of a full export with multiple items (for examples device types) at once? This would be really awesome!

Thanks and best
Max

kjetilho:
One use case for this feature: The changelog holds the JSON object for deleted objects, and it would be very convenient if this could be fed directly into an Import Device dialog to get an undo feature.  (Since the JSON contains numeric ids for stuff like `site` or `device_type`, it is not so convenient to recreate the device by hand.)

sol1-matt:
The import/export of device types already supports YAML but has limitations (discussion #6228). It works differently from normal import/export so don't know if would fall under the umbrella of this issue. 

The ability to automatically create the linked objects is nice but the inability to bulk import isn't.

If there is to be a refactor consideration of device types would be useful.

tom0010:
> One use case for this feature: The changelog holds the JSON object for deleted objects, and it would be very convenient if this could be fed directly into an Import Device dialog to get an undo feature. (Since the JSON contains numeric ids for stuff like `site` or `device_type`, it is not so convenient to recreate the device by hand.)

Also looking for this type of thing. 

In this release, would it be possible to do that via the API too?

jsenecal:
> > One use case for this feature: The changelog holds the JSON object for deleted objects, and it would be very convenient if this could be fed directly into an Import Device dialog to get an undo feature. (Since the JSON contains numeric ids for stuff like `site` or `device_type`, it is not so convenient to recreate the device by hand.)
> 
> Also looking for this type of thing.
> 
> In this release, would it be possible to do that via the API too?

What do you mean by "do that via the API too" ?

tom0010:
Bulk import NetBox information in json format via the API as well as in the GUI.

In this issue, my understanding is that you could bulk import yaml AND json, therefore I am assuming it's just the GUI.
Even though yaml and json follow the same model as such, from how I am reading it, it seems like this is a just GUI functionality. So I'd like to have the ability to do that via the API too if possible.

PR #10367, opened 2022-09-14
4347 Add JSON/YAML import support for all objects

arthanson:
### Fixes: #4347

Add JSON/YAML import support for all objects.  Changes DeviceType and ModuleType import to use the standard BulkImportView.  Changes BulkImportView code and forms to incorporate the JSON and YAML import code.


jeremystretch:
One thing we still need to consider is how to replicate the field accessor function from the CSV import to the JSON and YAML imports. For example, you can specify `region.id` when import sites to reference regions by numeric ID rather than by name. Ideally, we'd support the same mechanism so that both of the example below work:

```json
{
    "name": "Site 1",
    "slug": "site-1",
    "status": "active",
    "region": 4
}
```
```json
{
    "name": "Site 1",
    "slug": "site-1",
    "status": "active",
    "region": {
        "name": "Africa"
    }
}
```

PR #10369, opened 2022-09-14
#10359 add region column to site table

### LINKED ISSUE #10359: Missing Region column on Devices object
### NetBox version

v3.3.1

### Feature type

Data model extension

### Proposed functionality

Hi All,

Greetings!

I noticed that Region column is missing in Table configuration for Devices object. But inside the device the Region is there. Need help on how to populate the Region column.

image

image

image

### Use case

So it will be part of export and table field and it can easily identify the device's region rather.

### Database changes

N/A

### External dependencies

N/A

arthanson:
### Fixes: #10359 

Adds region column to Device table.


jeremystretch:
This one requires an explicit column definition on the DeviceTable class to a) specify the accessor for the related region and b) linkify the object. We should also add a similar column for site groups.

Issue #9045, opened 2022-04-06
Remove legacy ASN field from provider model


ryanmerolle:
### NetBox version

v3.2.0

### Feature type

Change to existing functionality

### Proposed functionality

Remove legacy ASN from provider model

### Use case

Currently there is an ASNs text field and an ASNs association field.

### Database changes

Requires a migration to remove the ASN field from providers.

Issue #9046, opened 2022-04-06
Remove legacy contact fields from provider model


ryanmerolle:
### NetBox version

v3.2.0

### Feature type

Change to existing functionality

### Proposed functionality

Remove legacy contact fields from provider model:

- Customer Portal
- NOC Contact
- Admin Contact

### Use case

Currently there is the above contact text fields and a contact association field.

### Database changes

Requires a migration to remove the contact fields from providers.

jeremystretch:
Blocked by #8234

jeremystretch:
Disregard; #8234 has already been addressed.

PR #10377, opened 2022-09-15
#9045 #9046 - remove legacy fields from Provider

arthanson:
### Fixes: #9045 #9046 

Remove legacy ASN from provider model
Remove legacy contact fields from provider model:

Customer Portal
NOC Contact
Admin Contact


jeremystretch:
@arthanson I've added a safeguard to the migration to check for any remaining legacy data (similar to what we did when removing the ASN and contact fields from the Site model in v3.2). However, there are a lot of places still where the legacy fields need to be removed (e.g. bulk edit forms, templates, etc.). You can search for all instances of each field name within the code base to ensure everything is accounted for.

arthanson:
@jeremystretch updated - removed all occurrences

netbox/extras/tests/test_customvalidator.py
```diff
@@ -67,22 +53,6 @@ def validate(self, instance):
...
-    @override_settings(CUSTOM_VALIDATORS={'circuits.provider': [min_validator]})
-    def test_min(self):
-        with self.assertRaises(ValidationError):
-            Provider(name='Provider 1', slug='provider-1', asn=1).clean()
-
-    @override_settings(CUSTOM_VALIDATORS={'circuits.provider': [max_validator]})
-    def test_max(self):
-        with self.assertRaises(ValidationError):
-            Provider(name='Provider 1', slug='provider-1', asn=65535).clean()
-
```
jeremystretch, line 85: These tests need to stay; they evaluate custom validation functionality. (We're just using the Provider model as an example.) I'll tweak them to use the ASN model instead.

netbox/utilities/tests/test_filters.py
```diff
@@ -454,22 +454,6 @@ def test_site_slug_endswith_negation(self):
...
-        self.assertEqual(ProviderFilterSet(params, Provider.objects.all()).qs.count(), 2)
-
-    def test_provider_asn_gt(self):
-        params = {'asn__lt': [65101]}
-        self.assertEqual(ProviderFilterSet(params, Provider.objects.all()).qs.count(), 1)
-
-    def test_provider_asn_gte(self):
-        params = {'asn__gte': [65101]}
-        self.assertEqual(ProviderFilterSet(params, Provider.objects.all()).qs.count(), 2)
-
```
jeremystretch, line 472: Same as above; need to rewrite these to use the ASN model instead.

jeremystretch (approved):

PR #10410, opened 2022-09-19
10408 add error message if already exists

### LINKED ISSUE #10408: Exception when adding the same contact multiple times
### NetBox version

v3.3.4

### Python version

3.8

### Steps to Reproduce

1. Create any contact, say "Contact1", and one Contact role, say "Admin"
2. Go to any Location
3. Under the Contacts panel, click "Add a contact"
4. Select Contact1 and Admin
5. Click Create and Add Another
6. Select Contact1 and Admin again
7. Click Create

### Expected Behavior

A form validation message saying that the contact already exists.

Optionally: existing contacts being disabled or hidden in the drop-down menu of available contacts.

### Observed Behavior

An exception was shown in the UI.

```

duplicate key value violates unique constraint "tenancy_contactassignmen_content_type_id_object_i_0764a1ef_uniq"
DETAIL:  Key (content_type_id, object_id, contact_id, role_id, priority)=(34, 1, 1, 1, ) already exists.

Python version: 3.8.10
NetBox version: 3.3.4
```

arthanson:
### Fixes: #10408

Adds error message if contact already exists.  This catches IntegrityError in ObjectEditView - @jeremystretch can you think of any instances where an IntegrityError here wouldn't be from a duplicate?  Not sure if this is too broad.

### DIFF
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -3,7 +3,7 @@ from copy import deepcopy
 
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
-from django.db import transaction
+from django.db import transaction, IntegrityError
 from django.db.models import ProtectedError
 from django.forms.widgets import HiddenInput
 from django.shortcuts import redirect, render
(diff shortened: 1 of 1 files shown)


jeremystretch:
Merged this PR by mistake, sorry.

jeremystretch:
This validation needs to be handled under the model, not the view. If an IntegrityError is raised to the view, we definitely want to know about it.

Issue #8927, opened 2022-03-21
Search integration in plugins


johannwagner:
### NetBox version

v3.1.6

### Feature type

New functionality

### Proposed functionality

We would like to introduce a possibility to search within added models of plugins. We would probably add a `search.py` within the plugin root repository that exposes  a similar formed structure as in https://github.com/netbox-community/netbox/blob/6c686af1b70acdc56a8fdbe84451f7012e2b4197/netbox/netbox/search.py#L241, similar to menu items. Therefore, this can be included into `SEARCH_TYPES` on plugin load to to enable a search in plugins.

We would be happy to provide a PR.

### Use case

We do some heavy lifting within Netbox plugins that is going to be extended after the next v3.2 release. Therefore, we need to search in our custom models to ensure a good usability. 

### Database changes

None

### External dependencies

None

ecraven:
This would be relevant for us too.

ryanmerolle:
@jeremystretch If @johannwagner  details out his approach, would you be happy to assign him this issue?

jeremystretch:
This is blocked until we come up with a new global search implementation (currently tracked under #7016).

jeremystretch:
We're flipping this around a bit. The plan now is to introduce a dynamic registration for model search first, and then to implement an improved search engine under #7016.

arthanson:
was already fixed and merged into feature https://github.com/netbox-community/netbox/pull/10489 so closing.

Issue #7016, opened 2021-08-23
Enable PostgreSQL full text searching


jeremystretch:
### NetBox version

v2.11.11

### Feature type

New functionality

### Proposed functionality

Extend NetBox's search functionality to employ PostgreSQL's native full text search capability. This is already supported by Django: https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/search/

I haven't dug into this too much yet, but it appears to be pretty straightforward. I expect the bulk of the work will involve replacing our sets of `filter()` parameters with `SearchQuery` and `SearchVector` instances, and adding a common `SearchVectorField` to all appropriate models.

I'm also open to alternative strategies for improving NetBox's global search performance, however any such proposals should be opened as separate feature requests and referenced here.

### Use case

This change should provide much more robust search functionality, and allow us to rank search results according to customizable rankings per model.

### Database changes

We'll likely need to add a new SearchVectorField to each searchable model.

tom-ua:
> Would this enable global searching based off of any object fields? That would be a great feature to have to find things that we might not know the name of 100%

Including MAC addresses. :-)

hagbarddenstore:
From the plugin framework workgroup:

Be able to add models defined in plugins to the global / full text search capabilities mentioned here.

PieterL75:
Looking forward to v3.3 with this full-text search !

jeremystretch:
Unfortunately it doesn't look like we'll be able to tackle this in v3.3 without additional help. A lot of work is still needed to determine exactly how this should be implemented, and any overlap with a future indexer (yet to be determined) must be evaluated.

PieterL75:
Is there a way to introduce a wildcard in the search as an alternative? So that we can do *switch00* or dev*sw* ? 
This is really one of the biggest remarks I get from new users that migrate from our old cmdb

ITJamie:
Would it be possible to sponsor this as a feature?

jeremystretch:
@ITJamie we don't allow direct sponsorship of features in NetBox, however you're certainly welcome to compensate individuals who want to help explore the feasibility of this proposal and perhaps develop a proof of concept.

arthanson:
Dug into this more:

1. watson - is tailored to a title-description-content search which I'm not sure really maps to what NetBox needs.  But does offer a multi-model search and does a lot of the work for you.
[...]

PR #10411, opened 2022-09-19
DRAFT: 7016 postgresql search

arthanson:


jeremystretch:
@arthanson per our discussion yesterday, let's re-target this PR to resolving #8927, and handle #7016 as a separate initiative.

PR #10417, opened 2022-09-20
Fixes #8366 - Add job scheduling

### LINKED ISSUE #8366: Add ability to schedule scripts & reports
### NetBox version

v3.1.5

### Feature type

New functionality

### Proposed functionality

This feature request is to add the ability to schedule report and script runs.

One idea is to create a new page/section for these schedules and possibly associated results that allows users to view all scheduled operations.

Another idea is to just add this functionality right into the existing scripts and reports page/section

### Use case

As a user, I want to be able to schedule recurring scripts and reports in NetBox itself.

Alternatively, a user could setup a schedule outside of UI using an external tool to execute an API call / POST or via the cli.  The challenge is that if you are not a user who is aware of the NetBox setup, its hard to follow how and what triggered the script or report run.

### Database changes

_No response_

### External dependencies

_No response_

kkthxbye-code:
### Fixes: #8366

This is a draft PR for adding job scheduling. Feel free to provide input.

Changes:

* Add a "Schedule at" form field to both the report and script form
* Add JobResult list view
* Add the `form-object-edit` class to both forms to match the width of forms in the rest of netbox (previously they were full width)
* Update the js `flatpickr` dependency to fix issue with the DateTimeField automatically populating when losing focus.
* Modify JobResult to delete RQ tasks to avoid running scheduled tasks with no JobResult associated
* Added new JobResultStatusChoice `Scheduled`
* Added new field `scheduled_time` to JobResult model
* Show scheduled time when viewing a job result
* Add --with-scheduler option to rqworker management command
* Added small snippet to docs about scheduling jobs
* REST API changes:
  * Added scheduled_time to JobResult

To do:
 
* Add job scheduling support to the API as well?
* `limit_choices_to=FeatureQuery('job_results'),` doesn't work at all in filter forms and it doesn't seem to work in other filterforms either.


jeremystretch:
@kkthxbye-code thanks for taking this on. I wanted to note that since this has been tagged for v3.4, the PR should be based on the `feature` branch.

kkthxbye-code:
@jeremystretch - Thanks, didn't notice when I opened the PR that it was the wrong target branch, the actual branch was made off of feature though, so it shouldn't be an issue. I changed the target now :)

kkthxbye-code:
I think this is ready for initial review now @jeremystretch

There's some stuff I'm unsure about:

* I moved job results out of the admin panel. Do we want another list view called scheduled jobs? It would be the same exact table as the job results table, just filtered to only contain job results with the scheduled status.
* Should I add support for scheduling jobs to the API as well?
* Currently the way to cancel a job is to delete the associated job result. Do we want to add another method that just cancels the schedule but lets the job result instance be?

I probably forgot some stuff, but let me know if something is not clear when testing.

netbox/extras/filtersets.py
```diff
@@ -434,8 +435,9 @@ class JobResultFilterSet(BaseFilterSet):
         method='search',
         label='Search',
     )
-    created = django_filters.DateTimeFilter()
-    completed = django_filters.DateTimeFilter()
+    created = django_filters.DateTimeFromToRangeFilter()
+    completed = django_filters.DateTimeFromToRangeFilter()
```
jeremystretch, line 439: Is this a breaking change, or is the behavior of the original filter retained?

netbox/extras/management/commands/housekeeping.py
```diff
@@ -81,7 +81,7 @@ def handle(self, *args, **options):
                         ending=""
                     )
                     self.stdout.flush()
-                JobResult.objects.filter(created__lt=cutoff)._raw_delete(using=DEFAULT_DB_ALIAS)
+                JobResult.objects.filter(created__lt=cutoff).delete(using=DEFAULT_DB_ALIAS)
```
jeremystretch, line 84: Should be fine to omit `using` here; it's only needed for the raw method. (Should also delete the import above.)
```suggestion
                JobResult.objects.filter(created__lt=cutoff).delete()
```

netbox/extras/management/commands/rqworker.py
```diff
@@ -14,6 +14,8 @@ class Command(_Command):
     of only the 'default' queue).
     """
     def handle(self, *args, **options):
+        # Run the worker with scheduler functionality
+        options['with_scheduler'] = True
```
jeremystretch, line 18: We might want to do something with the `--with-scheduler` command argument, since it will no longer have any effect. Not sure if there's an easy way to do that though. I doubt it's worth any significant effort.

netbox/extras/views.py
```diff
@@ -788,3 +798,29 @@ def get(self, request, job_result_pk):
             'result': result,
             'class_name': script.__class__.__name__
         })
+
+
+#
+# Job results
+#
+
+class JobResultView(generic.ObjectView):
```
jeremystretch, line 807: This doesn't appear to be used?

jeremystretch:
> Do we want another list view called scheduled jobs?

Can we just add a filter for the primary list?

> Currently the way to cancel a job is to delete the associated job result. Do we want to add another method that just cancels the schedule but lets the job result instance be?

Possibly; I'm happy to defer to feedback we receive during the beta cycle.

netbox/extras/views.py
```diff
@@ -605,6 +606,12 @@ def post(self, request, module, name):
         if report is None:
             raise Http404
 
+        schedule_at = None
+        form = ReportForm(request.POST)
+
+        if form.is_valid():
+            schedule_at = form.cleaned_data.get("schedule_at")
```
jeremystretch, line 613: This seems like it will run the report even if the form is invalid. We probably want to follow a pattern similar to `ScriptView.post()`, where we fall back to returning a view of the rendered form.

jeremystretch (approved):
Awesome work @kkthxbye-code! A couple minor comments but overall this is excellent! Thanks for owning this.

netbox/extras/filtersets.py
kkthxbye-code: I looked into this now, and I'm not sure what is the right solution here and I'm frankly a little confused.

Current behavior (develop branch) is that there's a created and completed DateTimeFilter.

https://github.com/netbox-community/netbox/blob/96c469641796b2b12ce643363d41d823670b68db/netbox/extras/filtersets.py#L437-L438

In practice this means that you can filter on the exact datetime value, so like `created=2022-10-21T09:23:57.593918`. Neither gte or lte works here. I think for those to work the filterset had to look like this?

https://github.com/netbox-community/netbox/blob/96c469641796b2b12ce643363d41d823670b68db/netbox/users/filtersets.py#L79-L87

Now with my change to DateTimeFromToRangeFilter created_after, created_before, completed_after, completed_before are added. So there's an issue here that the base created/completed actually doesn't exist anymore. So how should this be added to the filterset to make the swagger definition correct?

If we look at another place where DateTimeFromToRange is used, it shows the same issue:

https://github.com/netbox-community/netbox/blob/96c469641796b2b12ce643363d41d823670b68db/netbox/extras/filtersets.py#L399

The `time` filter that shows up in swagger doesn't actually work at all, but `time_before` and `time_after` does.

I'm really not sure what to do here, it needs to be DateTimeFromToRangeFilter for the filter form to work like for example the object change filter form. But the filters need to be different for the API for the swagger docs to match. I don't think it makes sense to have a filter that only matches on exact date, but __gte and __lte works some places like the TokenFilterSet I linked above, should we use a normal DateTimeFilter with those instead?

netbox/extras/management/commands/housekeeping.py
kkthxbye-code: Removed the using parameter.

> Should also delete the import above

Could you elaborate, not sure I follow.

kkthxbye-code:
> Can we just add a filter for the primary list?

It is already possible to filter by status in the JobResult view. We could add a menu item that links to the JobResult list with the status pre-filtered to `Schedules` but not sure if it is necessary. 

Thank you for the review, I adressed what I could. Regarding the filtersets I'm gonna need some guidance.

netbox/extras/management/commands/housekeeping.py
jeremystretch: Sorry, I mean the the import of `DEFAULT_DB_ALIAS` can be removed.

kkthxbye-code: It is still being used for the cleanup of ObjectChange's, so I'll let it stay for now at least. Not sure how _raw_delete actually functions, but if it doesn't respect cascade deletions, maybe you want to use the nomal delete method for ObjectChanges as well to ensure search indexes are deleted? Just a thought, I'm gonna resolve this for now.

netbox/extras/filtersets.py
jeremystretch: Agreed, this is frustrating.

> The time filter that shows up in swagger doesn't actually work at all, but time_before and time_after does.

AFAICT this is because DateTimeFromToRangeFilter doesn't actually register the before/after lookups on the filterset, so they can't be inspected. `get_filters()` returns a single filter for each:

```
created <django_filters.filters.DateTimeFromToRangeFilter object at 0x7fcf591b2c40>
completed <django_filters.filters.DateTimeFromToRangeFilter object at 0x7fcf591b2c70>
```

It's not ideal, but the simplest approach until we can come up with something more elegant would be to add the before & after lookups explicitly, like under TokenFilterSet. However I do think we should stick with the before/after nomenclature as they make more sense in the context of time than lte/gte. Something like this:

```
created = django_filters.DateTimeFilter()
created_before = django_filters.DateTimeFilter(
    field_name='created',
    lookup_expr='lte'
)
created_after = django_filters.DateTimeFilter(
    field_name='created',
    lookup_expr='gte'
)
```

It's probably worth auditing how we handle datetime filtering in NetBox overall, but this approach would let us knock out this PR without breaking either of the existing filters. What do you think?

jeremystretch: Looks perfect, thanks!

> Do note that it is now inconsistent with the ObjectChangeFilterSet and FilterForm

Yep, and I think that's fine for now. Ultimately (possibly in v3.4) I expect we'll bring them in line with what you have here, to be more consistent with other filter lookups in general.

jeremystretch:
Awesome work @kkthxbye-code! :raised_hands:

Issue #10348, opened 2022-09-12
Implement Decimal and/or Float as a custom field data type


eronlloyd:
### NetBox version

v3.3.2

### Feature type

Change to existing functionality

### Proposed functionality

While the Integer custom field data type is usage for many use cases, there are others that require decimal or float precision. 

I propose that one or both of these numeric types be added as choices for custom fields to enable such use cases. @jeremystretch pointed out that the underlying JSON serialization would have to be taken into consideration for the different types as JSON doesn't distinguish between numerical types. This constraint may make sticking with Decimal the better option to preserve precision.

### Use case

As an example, I'm trying to create custom fields for optical insertion loss (IL) for ports and attenuation for cables, but this needs to be in decimal format to accurately represent the decibel values.

### Database changes

Stored as JSON in the DB, so shouldn't affect the data model. Requires changes when serialized/de-serialized into memory.

### External dependencies

None

PR #10422, opened 2022-09-20
10348 add decimal custom field

arthanson:
### Fixes: #10348

Adds custom decimal field type and associated tests.


jeremystretch (changes requested):
Great work! The only issue I see at the moment is that decimal values are being stored as strings, rather than decimals/floats. Need to dig into this some more.

jeremystretch:
> The only issue I see at the moment is that decimal values are being stored as strings, rather than decimals/floats.

Ok, tracked this down. The `custom_field_data` JSONField on each model is using DjangoJSONEncoder, which explicitly [encodes decimals as strings](https://github.com/django/django/blob/19e6efa50b603af325e7f62058364f278596758f/django/core/serializers/json.py#L103).

Ideally, we should save these as raw JSON numbers. I'm not sure of the impact on precision, however it's probably acceptable.

If storing decimal values as JSON numbers is not tenable, an alternative would be to store them as strings but convert to decimals under `CustomField.deserialize()`.

jeremystretch:
This should be about ready to go. We still need to resolve the validation issue with MultiValueDecimalFilter that @arthanson called out in ad6d4f892d5e2f113fa09528dd32f30fc0168313, but that should be addressed separately as it may have implications for other filters.

PR #10441, opened 2022-09-22
9071 add header to plugin menu

### LINKED ISSUE #9071: Allow plugins to inject top-level navigation menu items
### NetBox version

v3.2.0

### Feature type

### Proposed functionality

Currently, plugins can define menu items only under the "Plugins" header within the navigation menu. Several plugin authors have proposed the ability for a plugin to "register" its own top-level section in the menu.

IMO the simplest approach would be to allow a plugin to define a menu heading with a title and icon. If defined, its menu items will appear there; if not, they will appear under the general "Plugins" heading.

We'd also want to give some thought to influencing the ordering of top-level menu items; assigning a "weight" to each item might suffice.

### Use case

Greater visibility for plugin functions; control over order of placement (potentially?)

### Database changes

_No response_

### External dependencies

_No response_

arthanson:
### Fixes: #9071

Allows plugins to define a menu_heading (see below) that will make it a top-level menu.  This still needs to edit the doc files for plugins.  @jeremystretch the original issue had a request to allow items to be appended to existing menus, however this would require the menu.groups to be an array instead of an immutable tuple so items could be appended on - not sure if this is desired or not?

menu_heading = {
    "title": "MyPlugin",
    "icon": "mdi-puzzle"
}


netbox/extras/plugins/__init__.py
```diff
@@ -70,9 +71,14 @@ def ready(self):
             register_template_extensions(template_extensions)
 
         # Register navigation menu items (if defined)
+        try:
+            menu_header = import_object(f"{self.__module__}.{self.menu_header}")
+        except AttributeError:
+            menu_header = None
```
jeremystretch, line 77: This doesn't need to be wrapped in a `try` block: `import_object()` will return None if the attribute doesn't exist.

netbox/extras/plugins/__init__.py
```diff
@@ -58,6 +58,7 @@ class PluginConfig(AppConfig):
     # integrated components.
     graphql_schema = 'graphql.schema'
     menu_items = 'navigation.menu_items'
+    menu_header = 'navigation.menu_heading'
```
jeremystretch, line 61: We should stick with either "header" or "heading" and use it consistently.

netbox/netbox/navigation_menu.py
```diff
@@ -408,18 +408,28 @@ def get_model_buttons(app_label, model_name, actions=('add', 'import')):
...
     plugin_menu_groups = []
 
-    for plugin_name, items in registry['plugins']['menu_items'].items():
-        plugin_menu_groups.append(
-            MenuGroup(
-                label=plugin_name,
-                items=items
+    for plugin_name, data in registry['plugins']['menu_items'].items():
+        if data['header']:
+            menu_groups = [MenuGroup(label=plugin_name, items=data["items"])]
```
jeremystretch, line 413: We should probably give some more thought to handling groups within a top-level menu. Ideally, plugins should be able to declare their own groups of items. I'm going to think about this some more.

PR #10442, opened 2022-09-22
10435 check if vm.cluster in qs

### LINKED ISSUE #10435: VM Interface: Error Fetching Options for field 'untagged_vlan'
### NetBox version

v3.3.4

### Python version

3.10

### Steps to Reproduce

1. Create a VM and assign it only to a Site, not a Cluster
2. Create a Interface for the VM
3. Scroll down, select  802.1Q Mode: "Access" 
4. Open the "Untagged VLAN" Dropdown

### Expected Behavior

A short error message with a hint on the error. Or the possibility to select a VLAN when no Cluster is assigned to the VM (if that makes any sense). 

### Observed Behavior

A very long error message with visible html code appears. 
Screenshot

arthanson:
### Fixes: #10435 

@jeremystretch  This fixes the issues but not sure if this is the correct fix - this will return all VLANs if there is no cluster defined, not sure if there should be some separate checks for site?

### DIFF
--- a/netbox/ipam/querysets.py
+++ b/netbox/ipam/querysets.py
@@ -81,30 +81,31 @@ class VLANQuerySet(RestrictedQuerySet):
 
         # Find all relevant VLANGroups
         q = Q()
-        if vm.cluster.site:
-            if vm.cluster.site.region:
+        if vm.cluster:
+            if vm.cluster.site:
+                if vm.cluster.site.region:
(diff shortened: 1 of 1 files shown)


jeremystretch:
> not sure if there should be some separate checks for site?

Yes, we should be filtering against `vm.site` directly. I can add that bit.

PR #10458, opened 2022-09-26
10458 add "Location" to CircuitTermination

### LINKED ISSUE #9604: Add Location to Circuit Termination
### NetBox version

v3.2.5

### Feature type

Data model extension

### Proposed functionality

When editing a circuit termination, i can specify "Region", "Site group", and "Site", but not "Location".

I would like to specify the "Location" field (optionally) as part of the termination.

### Use case

We have cross-connects between our cage and cages of other tenants in Equinix.  Each tenant's cage is identified by placing the Equinix cage ID in the "Location" field, and this works well in the "Region / Site Group / Site / Location" containment hierarchy.

Right now I am having to specify the tenant's cage as the "Site", because "Location" is not available to me when editing the circuit termination, which makes for something of an impedance mismatch.

### Database changes

add foreign key from `CircuitTerminations` to `Location`, (where `null=True`)

### External dependencies

None

gjvc:
### Fixes: #9604

added location field to ProviderForm
added location foreign key to CircuitTermination
added location field to circuittermination_edit.html template
updated circuits table to display Side A/B value via this logic:
- if location is set, show location
- if site only is set, show site (as before)
- if provider_network only is set, show provider_network (as before)


jeremystretch:
Thanks for your interest in working on this, however please note that #9604 has not been assigned a release milestone (required as this would be a breaking change to the API) and thus is not yet ready for work.

gjvc:
Sure thing; I will keep an eye on the original issue and this PR and will resubmit when you are ready.  Thank you for your consideration.

PR #10475, opened 2022-09-27
Issue10352 removegetvariables

### LINKED ISSUE #10352: Remove 'empty' variables from url after search/filter
### NetBox version

v3.3.2

### Feature type

### Proposed functionality

When a search or filter is done, the resulting page has an url with all possible fields.
It would be neater if only the field that has data is added to the url variable.
Especially when there are a lot of custom fields.

ex : 
https://netbox/dcim/devices/?q=test&serial=&asset_tag=&mac_address=&device_type_id=79&console_ports=&console_server_ports=&power_ports=&power_outlets=&interfaces=&pass_through_ports=&has_primary_ip=&virtual_chassis_member=&local_context_data=&cf_hostname=&cf_ltag=&cf_observium=&cf_observium_id=&cf_reference=&cf_u_bios_uuid=&cf_warrantydate=

could be reduced to 

https://netbox/dcim/devices/?q=test&device_type_id=79

### Use case

When URLs are shared or stored, it gives a better view in what is going to be filtered.

### Database changes

_No response_

### External dependencies

_No response_

PieterL75:


PieterL75:
The script works, but I fail to get it to work with the typescript files. Could you check what I'm doing wrong ?

netbox/project-static/src/util.ts
```diff
@@ -477,3 +477,31 @@ export function replaceAll(input: string, pattern: string | RegExp, replacement:
 
   return input.replace(pattern, replacement);
 }
+
```
kkthxbye-code, line 480:
```suggestion
export function cleanGetUrl(targetform: HTMLFormElement): void {
  const form_elements = targetform.elements;

  for (const element of form_elements) {
    // The SELECT statement requires a different approach. It depends on the selectedIndex, rather that the value.
    switch (element.nodeName) {
      case 'SELECT': {
        const selectElement = element as HTMLSelectElement;
        if (
          selectElement.selectedIndex == null ||
          selectElement.selectedIndex == -1 ||
          selectElement[selectElement.selectedIndex].getAttribute('value') == ''
        ) {
          element.setAttribute('disabled', '');
        }
        break;
      }
      default:
        if (element.getAttribute('value') == null) {
          element.setAttribute('disabled', '');
        }
    }
  }
}

```

kkthxbye-code: This change will fix the ts and linter warnings, however it doesn't work at all for me currently.

netbox/templates/inc/filter_list.html
```diff
@@ -1,7 +1,7 @@
 {% load form_helpers %}
 {% load helpers %}
 
-<form action="." method="get">
+<form action="." method="get" onsubmit="return cleanGetUrl(this)">
```
kkthxbye-code, line 4: This is putting javascript in the templates, which as Jeremy said, we don't do. Use event listeners instead.

Docs: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

netbox/templates/search.html
```diff
@@ -66,7 +66,7 @@ <h3 class="text-muted text-center">No results found</h3>
     {% else %}
         <div class="row">
             <div class="col col-12 col-lg-6 offset-lg-3">
-                <form action="{% url 'search' %}" method="get" class="form form-horizontal">
+                <form action="{% url 'search' %}" method="get" class="form form-horizontal" onsubmit="return cleanGetUrl(this)">
```
kkthxbye-code, line 69: Same as above.

netbox/utilities/templates/search/searchbar.html
```diff
@@ -1,4 +1,4 @@
-<form class="input-group" action="{% url 'search' %}" method="get">
+<form class="input-group" action="{% url 'search' %}" method="get" onsubmit="return util.cleanGetUrl(this)">
```
kkthxbye-code, line 1: Same as above.

netbox/project-static/src/netbox.ts
```diff
@@ -14,6 +14,9 @@ import { initRackElevation } from './racks';
 import { initLinks } from './links';
 import { initHtmx } from './htmx';
 
+// @ts-ignore
+import { cleanGetUrl } from './util';
```
kkthxbye-code, line 18: The linter will not allow this, as it is unused. Should go away when you implement this properly using event listeners. Remember to remove @ts-ignore

kkthxbye-code (changes requested):
Not sure why lldp.js is changed, I don't think it should with the current changes. The relevant dist files should be updated before merge though.

Your current approach needs some work, please see comments.

jeremystretch:
Works well for me. @kkthxbye-code any lingering  concerns?

kkthxbye-code:
@jeremystretch - I don't think the approach works in general. It breaks the form when navigating back as the elements are still disabled, at least for me in both chrome and firefox.

Maybe an alternative approach would be to modify the history with window.history.pushState()/window.history.replaceState(). I'm not really a frontend guy though.

The new code also has some other issue. The switch is not needed for two outcomes, there should not be a default case as that will break if we ever add other form elements to a get form, there's redunant checks when checking for falsy values, the function returns a boolean for no reason, the prototype of the object is compared by the nodeName string instead of Object.getPrototypeOf.

PieterL75:
`I'm not really a frontend guy though.`
Me neither :-D

`The switch is not needed for two outcomes, there should not be a default case as that will break if we ever add other form elements to a get form`
I had to change the element type to HTMLSelectElement when using the TypeScript, In regular Javascript it just ignores the fact that the selectedItems does not exist.

`there's redunant checks when checking for falsy values`
Indeed. I had to do that to catch all cases. Just comparing to only null, empty string, ... did not catch all 'empty' cases

`the function returns a boolean for no reason`
In my experience, the 'onsubmit' event function has to return a true or false. In case of false the form submit is canceled.

`the prototype of the object is compared by the nodeName string instead of Object.getPrototypeOf.`
I see that Object.getPrototypeOf return the class, but I have no idea how work with it to typescript validating properly

kkthxbye-code:
> I had to change the element type to HTMLSelectElement when using the TypeScript, In regular Javascript it just ignores the fact that the selectedItems does not exist.

I realize that, but you didn't have to treat everything else as an HTMLInputElement.

> Indeed. I had to do that to catch all cases. Just comparing to only null, empty string, ... did not catch all 'empty' cases

```
!inputElement.value ||
inputElement.value == null ||
inputElement.value == ''
```

The first !inputElement.value covers the last two cases in your code here. That's what I meant.

> In my experience, the 'onsubmit' event function has to return a true or false. In case of false the form submit is canceled.

I didn't see that in my testing. Returning false might break it, but when I tested making the function return void did not.

> I see that Object.getPrototypeOf return the class, but I have no idea how work with it to typescript validating properly

Just validate first that `Object.getPrototypeOf(element) === HTMLInputElement.prototype`, then cast it to the type.

Regardless, as I said, I think breaking back button behavior is a no-go here, so the chosen approach seems to be a non-starter.

jeremystretch:
> I don't think the approach works in general. It breaks the form when navigating back as the elements are still disabled, at least for me in both chrome and firefox.

I'm inclined to agree. I suspect we'd need to handle the form submission ourselves, ignoring empty values, rather than relying on the default submission process. But I'm _also_ not a front end person. :laughing:

PieterL75:
Well, you can't say I didn't try :-D

Another option that I read, is to use the FormData, going to test that out

jsenecal:
We could `preventDefault()` and submit the form via `get` only the fields with data manually.

Is that your intent @PieterL75 ?

PieterL75:
https://developer.mozilla.org/en-US/docs/Web/API/FormDataEvent

  const documentForms = document.forms
  for (var documentForm of documentForms) {
    if (documentForm.method.toUpperCase() == 'GET') {
      documentForm.addEventListener('formdata', function(event) {
        let formData = event.formData;
        for (let [name, value] of Array.from(formData.entries())) {
          if (value === '') formData.delete(name);
        }
      });
    }
  }
src: https://stackoverflow.com/questions/8029532/how-to-prevent-submitting-the-html-forms-input-field-value-if-it-empty/64029534#64029534

This works pretty awesome.. too bad typescript doesnt know the formdata event

netbox/project-static/src/netbox.ts
```diff
@@ -42,7 +41,16 @@ function initWindow(): void {
...
+// @ts-ignore: // formdata is not yet supported by TS https://github.com/microsoft/TypeScript/issues/36217
+      documentForm.addEventListener('formdata', function(event) {
+// @ts-ignore: // formdata is not yet supported by TS https://github.com/microsoft/TypeScript/issues/36217
+        let formData = event.formData;
+// @ts-ignore: // formdata is not yet supported by TS https://github.com/microsoft/TypeScript/issues/36217
+        for (let [name, value] of Array.from(formData.entries())) {
+// @ts-ignore: // formdata is not yet supported by TS https://github.com/microsoft/TypeScript/issues/36217
+          if (value === '') formData.delete(name);
+        }
+      });
```
kkthxbye-code, line 53:
```suggestion
      // @ts-ignore: Our version of typescript seems to be too old for FormDataEvent
      documentForm.addEventListener('formdata', function (event: FormDataEvent) {
        let formData: FormData = event.formData;
        for (let [name, value] of Array.from(formData.entries())) {
          if (value === '') formData.delete(name);
        }
      });
```

kkthxbye-code: I spent a little time trying to get this working. The suggested change works on my PC.

FormDataEvent is actually in typescript, our version is just too old. I tried updating typescript but as usual you just enter a rabbithole of broken stuff. Updating typescript -> Now popperjs has type errors -> Try to update popperjs -> Didn't work -> Google popperjs -> It changed name to floating-ui -> Start reading migration guide -> Give up.

I think we need to spend a little time soon getting all our JS dependencies up-to-date. Our typescript release seems to be from june 2021, which might as well be a decade in web-framework terms.

Could you try my suggestion? See if CI builds it, and then maybe I'll take a look at the dependency situation soon-ish.

kkthxbye-code (changes requested):
Works as expected, see suggested change.

netbox/project-static/src/netbox.ts
PieterL75: My tsc throws an error : 
```
tsc --noEmit
src/netbox.ts(46,30): error TS2339: Property 'formData' does not exist on type 'Event'.
src/netbox.ts(47,18): error TS2488: Type 'unknown' must have a '[Symbol.iterator]()' method that returns an iterator.
```
the ts-ignore is only for the next lines... so it also fails for the other items it does not know

kkthxbye-code: If you push the bundles so that the CI doesn't fail and it builds I think it's fine for now. I'll fix it up when I get to updating our JS dependencies.

I'm a little worried that my suggesting didn't work for you, so please remove your node_modules folder and do a clean `yarn install` afterwards to make sure you have the correct dependencies. Do this before bundling.

kkthxbye-code:
I created https://github.com/netbox-community/netbox/pull/10526 to test if my suggestion really was only working for me, but it seems like the CI builds it fine, so you should check your environment to make sure it is up to date.

You can see the change here:

https://github.com/netbox-community/netbox/pull/10526/files

I would prefer if you could try again to make sure we do it properly. Sorry for all the issues, the solution you came up with is great!

PieterL75:
> but it seems like the CI builds it fine, so you should check your environment to make sure it is up to date.

It's actually my local tsc (v4.8.4, latest build) that is throwing the errors.
I can override it (as is, disable the tsc check) for this build, but not sure how it will handle it in future builds
```
$ tsc --noEmit
src/netbox.ts(46,28): error TS2339: Property 'formData' does not exist on type 'Event'.
src/netbox.ts(47,16): error TS2488: Type 'unknown' must have a '[Symbol.iterator]()' method that returns an iterator.
```

kkthxbye-code:
You are running a globally installed tsc. You need to use yarn validate instead, which uses the correct typescript version installed locally.

PieterL75:
I did follow that guide to setup all things.. worked perfect till now :-(
its part of the output of the pre-commit https://docs.netbox.dev/en/stable/development/getting-started/#3-enable-pre-commit-hooks, so not ran manually..
yarn v1.22.19, tsc v4.8.4
Not sure what is causing the difference in interpretation by the CI build and the pre-commit script.
What versions are the CI builds running ?

kkthxbye-code:
Remove the typescript you have globally installed with something like.

`yarn global remove typescript`

No you don't need ignore on all lines, please see the CI build in the PR I linked, it runs validate fine. Nevertheless you need to fix your environment, you are currently using a newer version of typescript than netbox.

PR #10478, opened 2022-09-27
Fixes: #10460 Fix view for connection details

### LINKED ISSUE #10460: Missing connection details below cable on several objects views since v3.3
### NetBox version

v3.3.4

### Python version

3.10

### Steps to Reproduce

As Netbox 3.3 introduced multi-connection cables some parts weren't updated to the new code and the following templates are not showing any connection detail below "Cable". 

- consoleport
- consoleserverport
- interface
- powerfeed
- poweroutlet
- powerport

The powerfeed view additionally lacks to power usage and the capacity graph.

I already have a PR ready and are happy to submit if this bug is accepted.

### Expected Behavior

Connection details are again visible below Cable on the views/templates for:

- consoleport
- consoleserverport
- interface
- powerfeed
- poweroutlet
- powerport

Power usage and the capacity graph are visible on the powerfeed again.

Examples for fixed views:

powerfeed_fixed

powerport_fixed

### Observed Behavior

The mentioned templates/view only show the cable itself, but no connection details and status.
Powerfeed doesn't show the connected device and only show N/A as power usage.

Examples of broken views:

powerfeed_broken

powerport_broken

phurrelmann:
### Fixes: #10460

* assumes only one connection for power related connections
* consoleport and consoleserver allow multiple connections on the view
* interface allows multiple connections and old circuit details removed (circuit is broken since at least 2.7 and this is probably a remnant of the first cable-model overhaul)

### DIFF
--- a/netbox/templates/dcim/consoleport.html
+++ b/netbox/templates/dcim/consoleport.html
@@ -72,22 +72,22 @@
                                     </a>
                                 </td>
                             </tr>
-                            {% if object.connected_endpoint %}
+                            {% for endpoint in object.connected_endpoints %}
                                 <tr>
                                     <th scope="row">Device</th>
-                                    <td>{{ object.connected_endpoint.device|linkify }}</td>
+                                    <td>{{ endpoint.device|linkify }}</td>
                                 </tr>
                                 <tr>
                                     <th scope="row">Name</th>
-                                    <td>{{ object.connected_endpoint|linkify:"name" }}</td>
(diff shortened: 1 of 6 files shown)


jeremystretch:
Hey @phurrelmann, thanks for volunteering for this! I think your approach is reasonable, however, given how much pre-existing duplicate code there is, this is probably a great opportunity to do some cleanup. I'm going to see if I can reduce the amount of redundant code via an inclusion template. Thanks again!

PR #10479, opened 2022-09-27
convert customfield multiobject to ids at save()

### LINKED ISSUE #10241: CustomField MultiObject returns dict but expects id's
### NetBox version

v3.2.9

### Python version

3.9

### Steps to Reproduce

1. Create a custom field of the multiobject type
2. create a device that uses that custom field and populate with some data
3. retrieve the custom field through the API
4. update the custom field with a new ID
5. save the device

### Expected Behavior

The device with customs fields is updated with the new value's

### Observed Behavior

The save of the object fails with the message that a dict is not acceptable, only ID's are allowed

This is because
- the GET API returns the multiobject custom fields as an array of dicts.
- the POST API expects only ID's and not dicts

Reason:
API GET call returns the expanded dicts for custom_fields multiobjects
when adding a new object to the custom_fields, we use the ID of the object
This causes the custom_fields to be a mix of dict and id's
But the deserializer of the CustomField class expects only ID's

PieterL75:
### Fixes: #10241 CustomField MultiObject API returns dict at read but expects id's at save

This change will convert the mixed array of dict and id's into id's only at deserialize

### DIFF
--- a/netbox/extras/api/customfields.py
+++ b/netbox/extras/api/customfields.py
@@ -69,6 +69,12 @@ class CustomFieldsDataField(Field):
                 "values."
             )
 
+        for cf in self._get_custom_fields():
+            if cf.name in data.keys():
+                # Convert the mixed array of dict and id's into id's only
+                if cf.type == CustomFieldTypeChoices.TYPE_MULTIOBJECT:
+                    data[cf.name] = [obj['id'] if isinstance(obj, dict) else obj for obj in data[cf.name]] or None
+
         # If updating an existing instance, start with existing custom_field_data
         if self.parent.instance:
             data = {**self.parent.instance.custom_field_data, **data}
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -293,8 +293,9 @@ class CustomField(CloningMixin, ExportTemplatesMixin, WebhooksMixin, ChangeLogge
             model = self.object_type.model_class()
             return model.objects.filter(pk=value).first()
         if self.type == CustomFieldTypeChoices.TYPE_MULTIOBJECT:
+            valueids = [obj['id'] if isinstance(obj, dict) else obj for obj in value] or None
             model = self.object_type.model_class()
-            return model.objects.filter(pk__in=value)
+            return model.objects.filter(pk__in=valueids)
         return value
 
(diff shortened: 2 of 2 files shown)


jeremystretch:
While I agree with the approach, I don't think `serialize()` is the best place to implement it. The `serialize()` and `deserialize()` methods are core methods on CustomField, entirely decoupled from the REST API, and they should be symmetric.

As this change relates specifically to the REST API, the logic should probably go in the [`to_internal_value()`](https://github.com/netbox-community/netbox/blob/669e86f96e5866124db04e7efc1542b84966987a/netbox/extras/api/customfields.py#L65) method of `CustomFieldsDataField`.

PieterL75:
Looks like that the to_internal_value() does not know about the customfield type. 
so I'll have to query the db for the types, itterate over all customefields and change dicts to ids...

or do you see a better approach ?

(I hope you're not spending more time in explaining me, then self fixing it.. )

jeremystretch:
> so I'll have to query the db for the types, itterate over all customefields and change dicts to ids... or do you see a better approach ?

That's what I'm thinking. I'm sure there's probably a more elegant way to tackle it, but I'd have to go digging pretty deeply into DRF again to see what our other options might be. Open to any other ideas.

PieterL75:
After committing.. I thought that his might be even better : 
Keep the login in the deserialize function of the model, and call that deserialize in the to_internal_value.
Similar to that is done in the to_presentation calling the serialize function

Let me know what you prefer.

jeremystretch:
Considering this further, the related objects should be mapped using [`WritableNestedSerializer`](https://github.com/netbox-community/netbox/blob/d1efbf662005aeed91fa21fa26e34de3e307bb50/netbox/netbox/api/serializers/nested.py#L15), to ensure we provide the same functionality as is provided for built-in relationships. We can probably pass the custom field's object type to `get_serializer_for_model()` to accomplish this dynamically. Let me play with it for a bit.

jeremystretch:
I've come up with a solution in https://github.com/netbox-community/netbox/commit/d5538c1ca3d7ef7ca2d457ce22bbc645b8b65505 per my comment above. I believe this satisfies #10241 but please let me know if you find any further limitations. Thanks!

PR #10488, opened 2022-09-27
Closes #9249: Ignore case for device/VM names

### LINKED ISSUE #9249: Make Device/VM unique name enforcement case insensitive
### NetBox version

v3.2.0

### Feature type

### Proposed functionality

Presently, Netbox validates device/VM names to ensure they are unique, but the enforcement of this is case sensitive. This allows _device1_ and _DEVICE1_ to coexist but not _DEVICE2_ and _DEVICE2_

I propose that the existing validation is extended to be case _insensitive_ to prevent duplicate devices/VMs being created. This could be a change to the default behaviour or perhaps a setting in configuration.py

### Use case

Extending the existing validation would prevent duplicate devices/VMs being created where the user is unaware that the intended device already exists within Netbox.

I can't think of a use case for the existing behaviour over the proposed behaviour

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #9249

- Adjust the uniqueness constraints on the Device and VirtualMachine models to use the `Lower('name')` expression rather than the raw `name` value.
- Adjust the `name` filters for Device and Virtual Machine to use the case-insensitive `iexact` lookup.
- Add relevant model and filter tests.

### DIFF
--- a/docs/release-notes/version-3.4.md
+++ b/docs/release-notes/version-3.4.md
@@ -3,8 +3,13 @@
 !!! warning "PostgreSQL 11 Required"
     NetBox v3.4 requires PostgreSQL 11 or later.
 
+### Breaking Changes
+
+* Device and virtual machine names are no longer case-sensitive. Attempting to create e.g. "device1" and "DEVICE1" will raise a validation error.
+
 ### Enhancements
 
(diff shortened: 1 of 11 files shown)


arthanson (approved):
Not an issue - just be people should be aware, if you do have existing duplicates and edit the record it will show the error so you will have to update the name, might surprise someone as it was already existing.

PR #10489, opened 2022-09-27
8927 plugin search

### LINKED ISSUE #8927: Search integration in plugins

arthanson:
### Fixes: #8927

Refactors and adds to the search code to allow two things:

1. Allow plugins to have search capabilities
2. Make the search system 'pluggable' so a new Postgres search can be added.

The major changes are moving the search definitions to be class based (from the current dictionary), adding in a pluggable search backend system and making the search registration auto-scan the apps (which requires the class based definitions) to dynamically pick up new apps and plugins.


jeremystretch:
This has gone through a few iterations, building on @arthanson's initial work, but ultimately I ended up settling on the following for _this_ FR (with additional work remaining to be done under #10560):

- The `SearchIndex` class enables global search for each relevant model
- Indexes are registered to the application registry using the new `register_search()` decorator
- FilterSetSearchBackend retains the existing legacy global search functionality
- Add the `search_indexes` config parameter to PluginConfig

We'll undoubtedly need to keep adjusting the search backends implementation under #10560, but this feels like a solid start, and meets the primary objective of enabling global search for plugin models.

PR #10525, opened 2022-09-30
10491 improve error message for ProtectedError on contact assignment

### LINKED ISSUE #10491: Wrong dependent objects when deleting contacts
### NetBox version

v.3.3.4

### Python version

3.9

### Steps to Reproduce

1. Open a contact that has assignments, like https://demo.netbox.dev/tenancy/contacts/3/
2. Press delete button
3. Notice the error box  in the lower right corner that says, that the contact depends on itself.

### Expected Behavior

The error box should link/list names of assigned objects.

### Observed Behavior

Instead of assigned objects, the contact itself is listed as many times as number of assignments exist. In this example it is "Dwight Schrute" 2 times while it should be "vm1" and "DM-Scranton".

grafik

arthanson:
### Fixes: #10491 

@jeremystretch I wasn't sure if there were any side-effects from this str change.  The error handler is at https://github.com/netbox-community/netbox/blob/develop/netbox/utilities/error_handlers.py#L12 For contacts since this is the side the error happens on the protected_objects str is showing the wrong end of the link.  I think changing str to show self.object would break it in other places so showing both sides of the connection in the str seemed like the best approach?

### DIFF
--- a/netbox/tenancy/models/contacts.py
+++ b/netbox/tenancy/models/contacts.py
@@ -163,8 +163,8 @@ class ContactAssignment(WebhooksMixin, ChangeLoggedModel):
 
     def __str__(self):
         if self.priority:
-            return f"{self.contact} ({self.get_priority_display()})"
-        return str(self.contact)
+            return f"{self.contact} ({self.get_priority_display()}) -> {self.object}"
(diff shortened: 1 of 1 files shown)


jeremystretch:
The change should be fine as you have it.

Issue #8424, opened 2022-01-21
Location of the rack in the device menu


KosenkoIvan:
### NetBox version

nexbox (v3.1.5)

### Feature type

Change to existing functionality

### Proposed functionality

It would be convenient if the racks were displayed not only in the racks menu but also in the devices menu. This would simplify switching between devices. That is, we will not need to go to the racks menu and select the object we need again, but instantly select the desired device.
![Screenshot_4](https://user-images.githubusercontent.com/97949308/150517231-104b59d1-10bb-4ed3-bee3-654af3145efe.png)
![Screenshot_3](https://user-images.githubusercontent.com/97949308/150517234-d84009f8-2169-46ce-8b0c-8cd2f961bd21.png)

### Use case

This would simplify switching between devices. That is, we will not need to go to the racks menu and select the object we need again, but instantly select the desired device.

jeremystretch:
Probably related to #4434

DanSheps:
I don't think devices should be showing the rack they are in, at least in my opinion.  Conflating two different views, is something I think we should avoid, as it will add to the maintenance burden.

KosenkoIvan:
Before NetBox, we used RackTables and racks were shown there. From my own experience I can say about the convenience.

cpmills1975:
How about adding the elevations button to the "Rack" row on the left hand "Device" pane which redirects to the elevations view pre-filtered to just one rack?

KosenkoIvan:
I do not think that such an implementation will be correct, because the advantage of a solid display of the rack is that it would be possible not only to view the appearance of the rack, but also to switch the next device in 1 click.

jeremystretch:
I think it makes sense to embed the rack elevation, however in doing so we'd certainly want to highlight the device itself within the SVG drawing. I'm going to mark this as blocked by #4434, which seeks to do just that.

jeremystretch:
If pursued, this work needs to be done on a v3.3 or later release (to enable device highlighting).

SimonSchmitz:
Is there a way to display the racks on the left? Because now all plug-in extensions are displayed under the rack.

![Bildschirmfoto 2022-10-13 um 18 36 30](https://user-images.githubusercontent.com/29935460/195654990-1b65469b-75cf-4fd3-be2a-c05c23a74734.png)

PR #10531, opened 2022-10-01
8424 device location

arthanson:
### Fixes: #8424

Adds rack SVG to device detail page. 

Monosnap dmi01-albany-sw01 | NetBox 2022-09-30 17-23-13


netbox/templates/dcim/device.html
```diff
@@ -286,6 +286,31 @@ <h5 class="card-header">
             </div>
             {% include 'inc/panels/contacts.html' %}
             {% include 'inc/panels/image_attachments.html' %}
+
+            <div class="text-end mb-4">
+              <select class="btn btn-sm btn-outline-dark rack-view">
+                <option value="images-and-labels" selected="selected">Images and Labels</option>
+                <option value="images-only">Images only</option>
+                <option value="labels-only">Labels only</option>
+              </select>
```
jeremystretch, line 295: Not sure we need this toggle for the device view, but if we want to keep it we should move the whole thing into a separate inclusion template to avoid duplicate template code. My vote would be to just omit the selector though.

jeremystretch (changes requested):

PR #10549, opened 2022-10-04
9669 sanitize usernames

### LINKED ISSUE #9669: Remote Auth does not sanitize usernames
### NetBox version

v3.2.2

### Python version

3.10

### Steps to Reproduce

1. Setup netbox with remote auth based on netbox.authentication.RemoteUserBackend / Header
2. Auto Create User must be enabled, see configuration example below
3. Now open netbox and set the define AUTH_HEADER to a username that contains invalid characters like `:` which is used for example by Google IAP
4. This creates a new user in netbox, now open the admin interface of netbox and try to change anything in the user (/admin/auth/user/)
5. The user edit dialog validates the username if you change something, the frontend does not allow saving any changes there because of an invalid character in the username.

```
REMOTE_AUTH_ENABLED: true
REMOTE_AUTH_BACKEND: "netbox.authentication.RemoteUserBackend"
REMOTE_AUTH_HEADER: "HTTP_X_GOOG_AUTHENTICATED_USER_EMAIL"
REMOTE_AUTH_AUTO_CREATE_USER: true                                                                                                                                                                                                                                                                                                                                                 
```

### Expected Behavior

Netbox should sanitize the usernames that are provided via REMOTE_AUTH_HEADER to create valid usernames.




### Observed Behavior

Google IAPs headers have a value of: `accounts.google.com:example@gmail.com` and it creates a new user based on this in netbox.
But the admin dialogs to change or update those user accounts validates the username and rejects any changes because of invalid characters in the username:

`Enter a valid username. This value may contain only letters, numbers, and @/./+/-/_ characters.`

arthanson:
### Fixes: #9669 

Ad discussed, haven't really tested, but should work.  The clean_usersnames is from social_auth, just added the line 8 replace to it.

### DIFF
--- a/netbox/users/utils.py
+++ b/netbox/users/utils.py
@@ -1,0 +1,9 @@
+from social_core.storage import NO_ASCII_REGEX, NO_SPECIAL_REGEX
+
+
+def clean_username(value):
+    """Clean username removing any unsupported character"""
+    value = NO_ASCII_REGEX.sub('', value)
+    value = NO_SPECIAL_REGEX.sub('', value)
+    value = value.replace(':', '')
+    return value
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -498,7 +498,7 @@ for param in dir(configuration):
 
 # Force usage of PostgreSQL's JSONB field for extra data
 SOCIAL_AUTH_JSONFIELD_ENABLED = True
-
+SOCIAL_AUTH_CLEAN_USERNAME_FUNCTION = 'netbox.users.utils.clean_username'
 
 #
 # Django Prometheus


netbox/users/utils.py
```diff
@@ -0,0 +1,9 @@
+from social_core.storage import NO_ASCII_REGEX, NO_SPECIAL_REGEX
+
+
+def clean_username(value):
+    """Clean username removing any unsupported character"""
+    value = NO_ASCII_REGEX.sub('', value)
+    value = NO_SPECIAL_REGEX.sub('', value)
```
jeremystretch, line 7: We still want to allow `@` and `.` per Django's [`UnicodeUsernameValidator`](https://github.com/django/django/blob/a142edcc4981bb27f4356acad0c7fdfb2d5f9849/django/contrib/auth/validators.py#L19) validator. We can probably just reference its `regex` attribute to strip out any non-confirming characters.

PR #10554, opened 2022-10-04
Fixes: #10356 backplane connections

### LINKED ISSUE #10356: Add new interface and cable types to model direct interface connections to backplanes
### NetBox version

v3.3.0

### Feature type

New functionality

### Proposed functionality

Blades (and router/switch modules) often provide network connections via direct physical connections on their backplane. The interface then is directly wired over the backplane. There are proper IEEE standards defined for 1G (1000BASE-KX, see https://en.wikipedia.org/wiki/Gigabit_Ethernet), for 10G (10GBASE-KX4 and 10GBASE-KR, see https://en.wikipedia.org/wiki/10_Gigabit_Ethernet#10GBASE-KR), for 40G and 100G (40GBASE-KR4, 100GBASE-KP4, 100GBASE-KR2 and 100GBASE-KR4, see https://en.wikipedia.org/wiki/100_Gigabit_Ethernet).

To model and trace backplane network interfaces I'd also like to add the generic cable type "Backplane" to document, that the cable is indeed a direct physical copper connection.

**Summary:**

1. Add new cable type "Backplane" to section of copper cables
2. Add new interface types to section "Ethernet (fixed)"

Interface types to be added:

- 1000BASE-KX
- 10GBASE-KR
- 10GBASE-KX4
- 40GBASE-KR4
- 100GBASE-KP4
- 100GBASE-KR2
- 100GBASE-KR4

If accepted I'll try to add/implement those new interface and cable type myself and submit a PR.

### Use case

Model physical interfaces connected directly to a backplane and make this visible with the correct cable of type "Backplane".
We have lots of bladecenters and blades where the interfaces of blades blade are directly wired to the backplane. Currenlty this physical connection can't be properly modeled and its not visible for the users that the cable is indeed a backplane connection.

This feature could also be handy when using the new modules feature in combination with modular switches where some internal interfaces are realized with direct backplane connectivity.

### Database changes

none, only new interface and cable type definitions are needed.

### External dependencies

none

phurrelmann:
### Fixes: #10356

- Added new interface-types for backplane connections. The interface types are grouped as "Ethernet (backplane)'. This is inline with other Ethernet types. I refrained from using "IEE 802.3ap" in the group description as its not adequate.  the SIG 802.3ap only was involved for the 1G and 10G speeds, but the newer standards were developed in different SIG.

- Added new port-type backplane to enable modeling of bldecenter-passthru modules (e.g. SBM-25G-P10 which hast 10/25 backplane front ports and 40G/100G fiber rear-ports and acts like a passive media converter).

- Added new cable-type backplane to model the connection for backplane ports/interfaces.

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -1178,7 +1178,6 @@ class PortTypeChoices(ChoiceSet):
     TYPE_OTHER = 'other'
     TYPE_BACKPLANE = 'backplane'
 
-
     CHOICES = (
         (
             'Copper',
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -1176,7 +1176,7 @@ class PortTypeChoices(ChoiceSet):
     TYPE_URM_P4 = 'urm-p4'
     TYPE_URM_P8 = 'urm-p8'
     TYPE_OTHER = 'other'
-    TYPE_BACKPLAME = 'backplane'
+    TYPE_BACKPLANE = 'backplane'
 
 
     CHOICES = (
@@ -1271,7 +1271,7 @@ class CableTypeChoices(ChoiceSet):
     TYPE_SMF_OS2 = 'smf-os2'
     TYPE_AOC = 'aoc'
     TYPE_POWER = 'power'
-    TYPE_BACKPLAME = 'backplane'
+    TYPE_BACKPLANE = 'backplane'
 
     CHOICES = (
         (
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -1268,7 +1268,6 @@ class CableTypeChoices(ChoiceSet):
     TYPE_SMF_OS2 = 'smf-os2'
     TYPE_AOC = 'aoc'
     TYPE_POWER = 'power'
-    TYPE_BACKPLANE = 'backplane'
 
     CHOICES = (
         (
@@ -1285,7 +1284,6 @@ class CableTypeChoices(ChoiceSet):
                 (TYPE_DAC_PASSIVE, 'Direct Attach Copper (Passive)'),
                 (TYPE_MRJ21_TRUNK, 'MRJ21 Trunk'),
                 (TYPE_COAXIAL, 'Coaxial'),
-                (TYPE_BACKPLANE, 'Directly connected (Backplane)'),
             ),
         ),
         (
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -1176,6 +1176,8 @@ class PortTypeChoices(ChoiceSet):
     TYPE_URM_P4 = 'urm-p4'
     TYPE_URM_P8 = 'urm-p8'
     TYPE_OTHER = 'other'
+    TYPE_BACKPLAME = 'backplane'
+
 
     CHOICES = (
         (
@@ -1199,6 +1201,7 @@ class PortTypeChoices(ChoiceSet):
                 (TYPE_F, 'F Connector'),
                 (TYPE_N, 'N Connector'),
                 (TYPE_MRJ21, 'MRJ21'),
+                (TYPE_BACKPLANE, 'Backplane'),
             ),
         ),
         (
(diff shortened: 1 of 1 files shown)


netbox/dcim/choices.py
```diff
@@ -1151,6 +1176,7 @@ class PortTypeChoices(ChoiceSet):
     TYPE_URM_P4 = 'urm-p4'
     TYPE_URM_P8 = 'urm-p8'
     TYPE_OTHER = 'other'
+    TYPE_BACKPLANE = 'backplane'
```
jeremystretch, line 1179: The scope of #10356 is limited to the introduction of new interface types. Any new port/cable types would need to be proposed separately.

netbox/dcim/choices.py
```diff
@@ -911,6 +922,20 @@ class InterfaceTypeChoices(ChoiceSet):
                 (TYPE_400GE_OSFP, 'OSFP (400GE)'),
             )
         ),
+        (
+            'Ethernet (backplane)',
```
jeremystretch, line 926: Per your comment [here](https://github.com/netbox-community/netbox/issues/10356#issuecomment-1245588171) we should call these out as 802.3ap types.
```suggestion
            'Backplane (802.3ap)',
```

jeremystretch (changes requested):

netbox/dcim/choices.py
phurrelmann: Fair enough. I forgot to include the new port type initially and only were missing them when testing and modeling my requirements. I'll create a new ticket for that.
But the new cable type is included in the ticket description.

jeremystretch: Sorry, I missed the cable type. Would "stacking" maybe be a better name for this? I've just never heard the term "backplane cable;" open to other opinions.

phurrelmann: I removed the new port type and will create a followup ticket to have that added.

A backplane connection is no stacking (like in virtual-chassis, etc). It is merely a direct physical connection without any cable :/ Nevertheless the cable description was adjusted to hopefully better describe it. If you come up with a better/more appropriate name feel free to change that. As said there actually  is no cable in use, when a backplane connection is established. It's just a shortcoming of the connection/cable model in netbox and I'm not questioning that nor do I propose any changes. I'd just like to have a way to filter and search for backplane connections/cables, so it needs an identifier. The type "other" wouldn't match reality too, imho.

cpund: What's the difference between a backplane connection and a module? For something like modular Cisco chassis (Nexus 7700 and Catalyst 9500), they just use slots that slide into the backplace, but I'm curious as to the reasoning for modeling the backplane?

phurrelmann: Imho a module is more or less limited to networking equipment (at least that's what it's designed for). My use case comes from modular server systems e.g. bladecenters. You don't want to model a blade as a module. A blade is a fully fledged child device that goes into device-bays of a chassis.

E.g. https://www.supermicro.com/en/products/superblade/module/SBI-4129P-C2N.cfm
This blade has 2x 10G interfaces that are backplane connections to the switch/passthru/whatever is plugged into the bladecenter chassis.

An ethtool output of one interface on those blades show:

```
user# ethtool eno1
Settings for eno1:
	Supported ports: [ Backplane ]
	Supported link modes:   1000baseKX/Full
	                        10000baseKR/Full
	Supported pause frame use: Symmetric Receive-only
	Supports auto-negotiation: Yes
	Supported FEC modes: Not reported
	Advertised link modes:  1000baseKX/Full
	                        10000baseKR/Full
	Advertised pause frame use: No
	Advertised auto-negotiation: Yes
	Advertised FEC modes: Not reported
	Speed: 10000Mb/s
	Duplex: Full
	Auto-negotiation: on
	Port: None
	PHYAD: 0
	Transceiver: internal
	Supports Wake-on: g
	Wake-on: g
        Current message level: 0x00000007 (7)
                               drv probe link
	Link detected: yes
```

This whole PR is for modeling those kind of devices.

jeremystretch: > A blade is a fully fledged child device that goes into device-bays of a chassis.

These are modeled in NetBox as child devices. There is no need to model a discrete connection as it's implied by the parent/child relationship.

I think we should omit the "backplane" connection type as it's too ambiguous, and easily confused with e.g. passive direct attach cabling.

jeremystretch:
I've omitted the "backplane" cable type per the comments above just to unblock this PR and get the interface types added.

PR #10555, opened 2022-10-04
Fixes #10527 - Update JS dependencies

### LINKED ISSUE #10527: Update javascript dependencies
### Proposed Changes

Update our javascript dependencies for all packages where it's possible with minimal work. It is yet to be decided if this is going in a patch release or 3.4.

### Justification

Our packages are a year plus old. It's not a problem in itself, but makes new development cumbersome, as seen in the case of #10475. #10417 also needed to update the flatpickr dependency to get around bugs with regards datetime fields.

kkthxbye-code:
### Fixes: #10527

Mostly done, just need to figure out why the graphql playground breaks.

* All dependencies upgraded except bootstrap. 
  * Going from bootstrap 5.0.x to 5.1.x or 5.2.x both broke various stuff, such as the divider between the table header and the badges in dark mode.
* Pinned dependencies in `resolutions` were removed.
* A few smaller fixes to support newer versions of eslint and typescript.

I went through the release notes I could find, and for the packages with changelogs looked for breaking changes. I also tried to verify that the functionality the packages provided still worked.


kkthxbye-code:
So, esbuild 0.14 or newer breaks our graphiql implementation. Not quite sure why, but it silently breaks `subscriptions-transport-ws` which we import in our hacky fix to bundle graphiql dependencies. It's probably caused by one of the two first breaking changes here, but alas I was unable to get it to work: https://github.com/evanw/esbuild/blob/master/CHANGELOG-2021.md#0140

I've pinned it to the newest working version and then someone smarter will have to re-evaluate our graphiql integration in the future.

Otherwise I think this is ready for review. I fixed up the type stuff that was broken in #10475. Please give it a thorough test, I did my best, but I could have missed something. I also currently have no way to test LLDP/Napalm stuff, so that's untested.

jeremystretch:
I've gone through and tested all the individual functionality that I can glean from the TypeScript libraries, and haven't found any issues. Of course that's no guarantee it's perfect, but I'm comfortable merging this. We'll continue to test under `develop` until the next patch release. Thanks for all your work on this @kkthxbye-code!

Issue #9072, opened 2022-04-07
Allow plugins to inject sub-tabs on object detail views


jeremystretch:
### NetBox version

v3.2.0

### Feature type

New functionality

### Proposed functionality

Object detail views typically have several tabs: the primary object details, the object change log, and the journal, for example. Some objects have additional tabs. This proposal (borne from discussion in WG8333) seeks to allow plugins to inject arbitrary additional tabs for particular objects.

The exact implementation is open for discussion, but it seems reasonable to assume that a plugin could register one or more (name, view) tuples to specific NetBox models identified by their app label and model name.

One likely blocker here is that we'll need to come up with a way to programmatically declare sub-tabs under a view, but this would serve the existing changelog & journal views as well.

### Use case

Provides a greater degree of integration with core NetBox models and more convenient navigation for the user.

kkthxbye-code:
Such a good addition to the plugin API!

jeremystretch:
Not only that, it also allows us to clean up quite a bit of cruft and redundancy internally!

PR #10597, opened 2022-10-06
#9072: Custom model view tabs for plugins

jeremystretch:
### Closes: #9072

- Establish a new `views` registry root
- Introduce `register_model_view()` for registering arbitrary views for core models, optionally with tab details for UI rendering
- Introduce `get_model_urls()` utility function for creating URL path entries from registered views
- Automatically register the appropriate views for models inheriting from `JournalingMixin` and/or `ChangeLoggingMixin`
- Add a `model_view_tabs` template tag for rendering the tabs of registered views
- Replace legacy template cruft for journal/changelog tabs with `model_view_tabs`
- Replace all the static journaling and changelog URL paths with the dynamic resolution mechanism
- Introduce ViewTab class for registering tabs
- Extend plugins tests & docs


jeremystretch:
Really happy with how this has turned out so far. I'd still like to look at introducing a decorator that can be used to wrap views for convenience. Also on the fence about whether to expose `register_model_view()` directly for plugins, or use an intermediate wrapper to give us some buffer between the plugins and core API.

kkthxbye-code:
@jeremystretch - Doing some work on another issue I noticed that the journal tab is not visible anymore unless there are any journal entries already, thus making it impossible to add journal entries without manually going to `/journal`.

Let me know if you want me to create an issue or you want to handle it.

jeremystretch:
@kkthxbye-code thanks, I missed that. I've been bouncing between a couple different approaches for allowing control over tab rendering, trying to find the sweet spot between simplicity and flexibility.

PR #10630, opened 2022-10-12
9584 add device type (slug) to filter list

### LINKED ISSUE #9584: GraphQL: Add device_type as a filter for device_list / device
### NetBox version

v3.2.5

### Feature type

### Proposed functionality

Currently, you can only filter device_type_id in the graphql filters.  It would be good to filter on either the slug or the name of device_type.

Current way to filter on device_type:
```
{
  device_list(device_type_id: "11") {
    id
    name
    device_type {
      id
    }
  }
}
​```

Requested approach:
```
{
  device_list(device_type: "<insert slug OR model>") {
    id
    name
    device_type {
      id
    }
  }
}
​```

### Use case

Often individuals do not know the id, so its hard to pass said value if you do not know it.

Frankly this just further sets graphql API standardization since this is already supported for device_role (you can filter on it like below):
```
{
  device_list(role: "patch-panel") {
    id
    name
    device_type {
      id
      model
      slug
    }
    device_role {
      id
      slug
    }
  }
}
```

### Database changes

N/A

### External dependencies

N/A

arthanson:
### Fixes: #9584

Added as a filter (GraphQL, REST).  Wasn't sure if we wanted this in the UI as well?  Under "Hardware"?

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -800,6 +800,12 @@ class DeviceFilterSet(NetBoxModelFilterSet, TenancyFilterSet, ContactModelFilter
         to_field_name='slug',
         label='Manufacturer (slug)',
     )
+    device_type = django_filters.ModelMultipleChoiceFilter(
+        field_name='device_type__slug',
+        queryset=DeviceType.objects.all(),
+        to_field_name='slug',
+        label='Device type (slug)',
+    )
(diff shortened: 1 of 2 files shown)


jeremystretch:
@arthanson could you add a test for the new filter to `dcim/tests/test_filtersets.py` please?

arthanson:
@jeremystretch done.

PR #10660, opened 2022-10-15
Closes #10659: Import signals.py from Plugin

### LINKED ISSUE #10659: Add django signals functionality to Plugins
### NetBox version

v3.3.5

### Feature type

New functionality

### Proposed functionality

Currently a developer has to copy the `ready(self)` function from `PluginConfig` to get signals running. 
Developers should be able to use signals just out of the box with a `signals.py` file in the root of the plugin directory. 

I tried to implement it myself and I will create a pull request as it's just two/three lines in `__init__.py` from `extras.plugins`. 

```
from importlib import import_module

...

class PluginConfig(AppConfig):

...

    signals = 'signals'

    def ready(self):
        plugin_name = self.name.rsplit('.', 1)[-1]

        # import signals if defined
        import_module(f"{self.__module__}.{self.signals}")

...
```

### Use case

Plugins could react to actions from the main part of NetBox. 

Example:
Someone develops a YubiKey plugin and wants to associate a new YubiKey everytime someone adds a new Device.  
This could be reached super handy by signals in django. 

### Database changes

_No response_

### External dependencies

_No response_

Selfmade-RuLeZ:
This pull request uses the importlib.import_module function from Python to import the signals module from a plugin. This registers the receiver functions to receive signals from django. Maybe someone should add a small documentation page for signal development within plugins. 

### Fixes: #10659

- [X]  Implement changes in `extras.plugins.__init__.py`
- [ ] Update documentation

### DIFF
--- a/netbox/extras/plugins/__init__.py
+++ b/netbox/extras/plugins/__init__.py
@@ -2,6 +2,8 @@ import collections
 import inspect
 from packaging import version
 
+from importlib import import_module
+
 from django.apps import AppConfig
 from django.core.exceptions import ImproperlyConfigured
 from django.template.loader import get_template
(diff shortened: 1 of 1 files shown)


DanSheps:

PR #10673, opened 2022-10-18
Fix LDAP auth: user never updated if inactive

### LINKED ISSUE #10666: LDAP: Disabled user never re-enabled when doing API calls
### NetBox version

v3.2.9

### Python version

3.9

### Steps to Reproduce

* set FIND_GROUP_PERMS to True
* set LDAP as the authentication backend
* use token to do any API requests continuously (either DRF or GraphQL) like this:
  `curl -H "Authorization: Token $NETBOX_TOKEN" 127.0.0.1:8001/api/users/users/`
* have the LDAP responding an empty responses. This is difficult to reproduce as it happens during too high load on the LDAP server, which is an issue itself
  * Easier alternative: disable the user manually via the Django admin panel

### Expected Behavior

When calling the API, an inactive user should always be synced with the LDAP (using `populate_user()` method) to recover from such issue.

It should happen whether FIND_GROUP_PERMS is enabled or disabled, to be able to recover from such issue.

If the following behavior is acceptable for you, I have a code ready to be submitted as a PR:

<details><summary>proposed fix</summary>

_**netbox/netbox/api/authentication.py:**_

    class TokenAuthentication(authentication.TokenAuthentication):
        """
        A custom authentication scheme which enforces Token expiration times.
        """
        model = Token

        def authenticate_credentials(self, key):
            model = self.get_model()
            try:
                token = model.objects.prefetch_related('user').get(key=key)
            except model.DoesNotExist:

            # Enforce the Token's expiration time, if one has been set.
            if token.is_expired:
                raise exceptions.AuthenticationFailed("Token expired")

            user = token.user
            # When LDAP authentication is active try to load user data from LDAP directory
            if settings.REMOTE_AUTH_BACKEND == 'netbox.authentication.LDAPBackend':
                from netbox.authentication import LDAPBackend
                ldap_backend = LDAPBackend()

                # Load from LDAP if FIND_GROUP_PERMS is active
                # Always query LDAP when user is not active, otherwise it is never activated again
                if ldap_backend.settings.FIND_GROUP_PERMS or not token.user.is_active:
                    ldap_user = ldap_backend.populate_user(token.user.username)
                    # If the user is found in the LDAP directory use it, if not fallback to the local user
                    if ldap_user:
                        user = ldap_user

            if not user.is_active:
                raise exceptions.AuthenticationFailed("User inactive")

            return user, token
</details>

### Observed Behavior

When using LDAP and FIND_GROUP_PERMS, if the LDAP server gives an empty response, the user is disabled.

The issue is that the user is never resynced again with the LDAP when using API calls. `populate_user()` is never called because the is_active condition exits the function before.

The only current solution is to resync the user by asking him to connect to the UI.

kpetremann:
### Fixes: #10666

Before this PR, a disabled LDAP user were never enabled again when using API endpoints (DRF or GraphQL).

A LDAP user can be disabled when:
- the LDAP returns an empty 200 when `populate_user() `is called
- someone disables the user in the admin

`populate_user()` were never called for a disabled user.

With this commit, it is always called for a disabled user.

### DIFF
--- a/netbox/netbox/api/authentication.py
+++ b/netbox/netbox/api/authentication.py
@@ -22,22 +22,24 @@ class TokenAuthentication(authentication.TokenAuthentication):
         if token.is_expired:
             raise exceptions.AuthenticationFailed("Token expired")
 
-        if not token.user.is_active:
-            raise exceptions.AuthenticationFailed("User inactive")
-
+        user = token.user
         # When LDAP authentication is active try to load user data from LDAP directory
         if settings.REMOTE_AUTH_BACKEND == 'netbox.authentication.LDAPBackend':
             from netbox.authentication import LDAPBackend
             ldap_backend = LDAPBackend()
 
             # Load from LDAP if FIND_GROUP_PERMS is active
-            if ldap_backend.settings.FIND_GROUP_PERMS:
-                user = ldap_backend.populate_user(token.user.username)
+            # Always query LDAP when user is not active, otherwise it is never activated again
+            if ldap_backend.settings.FIND_GROUP_PERMS or not token.user.is_active:
+                ldap_user = ldap_backend.populate_user(token.user.username)
                 # If the user is found in the LDAP directory use it, if not fallback to the local user
-                if user:
-                    return user, token
+                if ldap_user:
+                    user = ldap_user
 
-        return token.user, token
+        if not user.is_active:
+            raise exceptions.AuthenticationFailed("User inactive")
+
+        return user, token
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @kpetremann!

kpetremann:
Thanks for merging this PR :tada:

kpetremann:
I saw :) but having it in the new release is the most important to me.
We can deploy it right now in production! :rocket:

PR #10687, opened 2022-10-19
Fixes #10580

### LINKED ISSUE #10580: link 'assigned' checkbox to device rather than ip
### NetBox version

v3.3.5

### Feature type

### Proposed functionality

When searching for an IP, the result shows the 'assigned' green checkbox if an IP address is assigned to a device/vm/...
The current link of the green check is to the IP address object.
I like to propose to change that to the device/vm/... the IP is assigned to.

### Use case

When the checkbox is clicked, it is more intuitive to go to the device it is assigned to.
Now it requires 2 additional clicks to go to the device.

### Database changes

_No response_

### External dependencies

_No response_

cpund:
### Fixes: #10580 

Changes accessor in IPAddressTable for the assigned column to switch it from linking to the IP address page to the linked interface's page.

### DIFF
--- a/netbox/ipam/tables/ip.py
+++ b/netbox/ipam/tables/ip.py
@@ -374,7 +374,7 @@ class IPAddressTable(TenancyColumnsMixin, NetBoxTable):
         verbose_name='NAT (Outside)'
     )
     assigned = columns.BooleanColumn(
-        accessor='assigned_object_id',
+        accessor='assigned_object__parent_object',
         linkify=True,
         verbose_name='Assigned'
     )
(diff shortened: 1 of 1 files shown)


netbox/ipam/tables/ip.py
```diff
@@ -374,7 +374,7 @@ class IPAddressTable(TenancyColumnsMixin, NetBoxTable):
         verbose_name='NAT (Outside)'
     )
     assigned = columns.BooleanColumn(
-        accessor='assigned_object_id',
+        accessor='assigned_object',
         linkify=True,
```
jeremystretch, line 378: Unfortunately this breaks ordering for the column (because the accessor is no longer tied to a discrete field). What we can do instead is pass a callable to `linkify` which will dynamically resolve the URL for the assigned object.

```suggestion
        accessor='assigned_object_id',
        linkify=lambda record: record.assigned_object.get_absolute_url(),
```

jeremystretch (changes requested):

cpund:
@jeremystretch applied. thanks again for the advice!

PR #10706, opened 2022-10-20
Set *_COOKIE_PATH according to BASE_PATH

### LINKED ISSUE #10639: Improve coexistence with other django-apps on the same host by setting *_COOKIE_PATH
### NetBox version

v3.2.4

### Feature type

New functionality

### Proposed functionality

While netbox provides one configuration option to specify a different `BASE_PATH`, It does not set the `CSRF_COOKIE_PATH`, `LANGUAGE_COOKIE_PATH` or `SESSION_COOKIE_PATH`. All Values could be set directly to `BASE_PATH or "/"`, as `BASE_PATH` already contains a trailing slash or empty string.

### Use case

Improve coexistence with other Django-based-projects hosted on the same host. While this can (and probably should) also be achieved by using different cookie-names, setting the path of these cookies probably decreases confusion.

### Database changes

None

### External dependencies

None

prauscher:
### Fixes: #10639

All three `COOKIE_PATH`s should be set accordingly with the netbox-`BASE_PATH` to improve coexistance with other Django-projects probably hosted on the same Host


jeremystretch:
CI failure is unrelated; see #10712.

candlerb:
Unfortunately this change sets an invalid cookie path which prevents login: #10837

PR #10715, opened 2022-10-20
7961 CSV bulk update

arthanson:
### Fixes: #7961 

Adds bulk update to CSV import if "id" is included in the CSV data.


netbox/utilities/testing/views.py
```diff
@@ -583,6 +598,55 @@ def test_bulk_import_objects_with_permission(self):
             self.assertHttpStatus(self.client.post(self._get_url('import'), data), 200)
             self.assertEqual(self._get_queryset().count(), initial_count + len(self.csv_data) - 1)
 
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])
+        def test_bulk_update_objects_with_permission(self):
+            if not self.csv_update_data:
```
jeremystretch, line 603: We probably want to raise an error if `csv_update_data` hasn't been set, as it likely means that someone forgot to add it.

netbox/utilities/testing/views.py
```diff
@@ -583,6 +598,55 @@ def test_bulk_import_objects_with_permission(self):
...
+                return
+
+            data = {
+                'csv': self._get_csv_data(),
+            }
```
jeremystretch, line 608: This is defined below.
```suggestion
```

netbox/utilities/testing/views.py
```diff
@@ -583,6 +598,55 @@ def test_bulk_import_objects_with_permission(self):
...
+            obj_perm.object_types.add(ContentType.objects.get_for_model(self.model))
+
+            # need to track ids so we know what new ids were added by csv_data so we can
+            # do the updates on the appropriate ids
+            prev_ids = list(self._get_queryset().values_list('id', flat=True).order_by('id'))
+            self.assertHttpStatus(self.client.post(self._get_url('import'), data), 200)
+            count = self._get_queryset().count()
+            new_ids = list(self._get_queryset().values_list('id', flat=True).order_by('id'))
+            diff_ids = [x for x in new_ids if x not in prev_ids]
+            start_id = diff_ids[0]
```
jeremystretch, line 626: I'm not comfortable baking this much logic into the test method itself. My fear is that hiding the logic responsible for retrieving the IDs of objects to be updated will inevitably lead to confusion for test authors. IMO it's much safer to specify the IDs explicitly within each test. For example, in ProvderTestCase we first create three providers:

```python
providers = (
    Provider(name='Provider 1', slug='provider-1'),
    Provider(name='Provider 2', slug='provider-2'),
    Provider(name='Provider 3', slug='provider-3'),
)
Provider.objects.bulk_create(providers)
```

We can reference these objects directly when defining the CSV update data:

```python
cls.csv_update_data = (
    "id,name,comments",
    "{providers[0].pk},Provider 7,New comment7",
    "{providers[1].pk},Provider 8,New comment8",
    "{providers[2].pk},Provider 9,New comment9",
)
```

Granted it's a bit more verbose, but with the benefit of conveying to the reader _exactly_ what data is being fed to the form without any digging.

I'm happy to make these adjustments if you're tired of dealing with the test suites at this point; I realize how mind numbing it gets. :slightly_smiling_face:

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -321,13 +321,47 @@ def clean(self):
...
 
+        return headers, records
+
+    def _update_objects(self, form, request, headers, records):
+        from utilities.forms import CSVModelChoiceField
+        new_objs = []
+
+        for row, data in enumerate(records, start=1):
+            try:
+                obj = self.queryset.model.objects.get(pk=data["id"])
```
jeremystretch, line 338: We should extract the IDs list and retrieve all of the instances in a single query. We can create a dictionary mapping each ID to its instance and quickly check for any missing keys.

jeremystretch (changes requested):

netbox/utilities/testing/views.py
arthanson: it's different data - the first is self._get_csv_data, in the second instance it is the update_csv_data

netbox/ipam/tests/test_views.py
```diff
@@ -804,7 +920,8 @@ class L2VPNTerminationTestCase(
     def setUpTestData(cls):
         device = create_test_device('Device 1')
         interface = Interface.objects.create(name='Interface 1', device=device, type='1000baset')
-        l2vpn = L2VPN.objects.create(name='L2VPN 1', type=L2VPNTypeChoices.TYPE_VXLAN, identifier=650001)
+        l2vpn = L2VPN.objects.create(name='L2VPN 1', slug='l2vpn-1', type=L2VPNTypeChoices.TYPE_VXLAN, identifier=650001)
+        l2vpn2 = L2VPN.objects.create(name='L2VPN 2', slug='l2vpn-2', type=L2VPNTypeChoices.TYPE_VXLAN, identifier=650002)
```
jeremystretch, line 924: Seems this is unused.
```suggestion
```

netbox/utilities/testing/views.py
```diff
@@ -583,6 +590,49 @@ def test_bulk_import_objects_with_permission(self):
             self.assertHttpStatus(self.client.post(self._get_url('import'), data), 200)
             self.assertEqual(self._get_queryset().count(), initial_count + len(self.csv_data) - 1)
 
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])
+        def test_bulk_update_objects_with_permission(self):
+            if not self.csv_update_data:
+                raise NotImplementedError("The test must define csv_update_data.")
```
jeremystretch, line 596: This will ensure we raise the NotImplementedError rather than an AttributeError.
```suggestion
            if not hasattr(self, 'csv_update_data'):
                raise NotImplementedError("The test must define csv_update_data.")
```

netbox/utilities/testing/views.py
```diff
@@ -583,6 +590,49 @@ def test_bulk_import_objects_with_permission(self):
...
+            # Assign model-level permission
+            obj_perm = ObjectPermission(
+                name='Test permission',
+                actions=['add']
+            )
+            obj_perm.save()
+            obj_perm.users.add(self.user)
+            obj_perm.object_types.add(ContentType.objects.get_for_model(self.model))
+
+            self.assertHttpStatus(self.client.post(self._get_url('import'), data), 200)
```
jeremystretch, line 611: This step shouldn't be necessary. We've already created some objects under `setUpTestData()` for each test, and we're modifying those. For instance, under ProviderTestCase, we have:

```python
        providers = (
            Provider(name='Provider 1', slug='provider-1'),
            Provider(name='Provider 2', slug='provider-2'),
            Provider(name='Provider 3', slug='provider-3'),
        )
        Provider.objects.bulk_create(providers)
```

and then:

```python
        cls.csv_update_data = (
            "id,name,comments",
            f"{providers[0].pk},Provider 7,New comment7",
            f"{providers[1].pk},Provider 8,New comment8",
            f"{providers[2].pk},Provider 9,New comment9",
        )
```

This will rename e.g. Provider 1 to Provider 7.

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -321,13 +321,52 @@ def clean(self):
...
         if request.FILES:
             headers, records = form.cleaned_data['csv_file']
         else:
             headers, records = form.cleaned_data['csv']
 
+        return headers, records
+
+    def _update_objects(self, form, request, headers, records):
+        from utilities.forms import CSVModelChoiceField
+        new_objs = []
```
jeremystretch, line 334: Might want to rename this just to avoid confusion.
```suggestion
        updated_objs = []
```

jeremystretch (approved):
Awesome work!

stavr666:
> How do I know which version of netbox contain is amazing update?
I's marked (and closed) as 3.4 milestone.

PR #10744, opened 2022-10-25
10716 add left-right plugins to tags page

### LINKED ISSUE #10716: Add left_page/right_page plugin template extensions to Tag page template
### NetBox version

v3.2.9

### Feature type

### Proposed functionality

The [tag page template](https://github.com/netbox-community/netbox/blob/develop/netbox/templates/extras/tag.html) does not include `{% plugin_left_page object %}` and `{% plugin_right_page object %}` like most other models, but only has the full width option. I propose to add these to the tag page as well.

### Use case

Automation systems might rely a lot on tags for its functionality, and being able to have a clearly visible plugin on the tag page for custom interaction/documentation could be very helpful. The full width option is placed way below the large Tagged Objects table, somewhat hiding the plugin if you don't know to look for it.

### Database changes

_No response_

### External dependencies

_No response_

arthanson:
### Fixes: #10716 

Adds left/right plugins template tag to tags page

### DIFF
--- a/netbox/templates/extras/tag.html
+++ b/netbox/templates/extras/tag.html
@@ -77,6 +77,7 @@
           {% include 'inc/paginator.html' with paginator=taggeditem_table.paginator page=taggeditem_table.page %}
         </div>
       </div>
+      {% plugin_full_width_page object %}
     </div>
   </div>
 {% endblock %}
--- a/netbox/templates/extras/tag.html
+++ b/netbox/templates/extras/tag.html
@@ -39,6 +39,7 @@
           </table>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


netbox/templates/extras/tag.html
```diff
@@ -75,7 +77,6 @@ <h5 class="card-header">Tagged Objects</h5>
           {% include 'inc/paginator.html' with paginator=taggeditem_table.paginator page=taggeditem_table.page %}
         </div>
       </div>
-      {% plugin_full_width_page object %}
```
jeremystretch, line 78: This tag should stay.

PR #10747, opened 2022-10-25
10719 check ip permission on fhrpgroup form save

### LINKED ISSUE #10719: User without sufficient permission can create an IP address via FHRP group creation
### NetBox version

v3.3.5

### Python version

3.8

### Steps to Reproduce

1. Create a user with permission to view and create FHRP groups but not IP addresses.
2. Authenticate as that user.
3. Create an FHRP group with any valid values, and also specify an IP address and status.

### Expected Behavior

The creation of an IP address should not be permitted, as the user lacks sufficient permissions.

### Observed Behavior

The IP address is created.

arthanson:
### Fixes: #10719 

Moved the save / permission check out of the ipam form and into the view.
Created a save_related_data to make it more generic when using the generic edit form, otherwise would need to copy-pasta the whole post function.
This also showed an existing issue (fixed) where if the form got a permission error it would switch to an edit view (noticeable on the fhrpgroup form) as the obj.pk would be set even though the transaction would be rolled-back.

### DIFF
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -362,6 +362,13 @@ class ObjectEditView(GetReturnURLMixin, BaseObjectView):
             **self.get_extra_context(request, obj),
         })
 
+    def save_related_data(self, request, form, objd):
+        """
+        Optionally override to save model specific related data after the form is saved.
+        Raise exception (PermissionsViolation) or such if error.
+        """
+        return
+
     def post(self, request, *args, **kwargs):
         """
         POST request handler.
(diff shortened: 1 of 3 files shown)


netbox/netbox/views/generic/object_views.py
```diff
@@ -362,6 +362,13 @@ def get(self, request, *args, **kwargs):
             **self.get_extra_context(request, obj),
         })
 
+    def save_related_data(self, request, form, objd):
```
jeremystretch, line 365: I like the approach, however we can't add new methods to generic views arbitrarily as they're part of the [plugins API](https://docs.netbox.dev/en/stable/plugins/development/views/), and adding a new method could conflict with a custom method defined by a plugin. It might be worth starting a FR to introduce this method in a future release, but for now it's probably sufficient to implement a one-off just for #10719.

jeremystretch:
I resolved this using a hack in https://github.com/netbox-community/netbox/commit/7b3ef2ade5266ec0738c4e210cf1c6f9b85fbee7 for now, though we should give more thought to this pattern of secondary object creation overall.

PR #10764, opened 2022-10-26
9599 cursor pagination

jeremystretch:
### Closes: #9599

This builds on the work by @lyuyangh in PR #9716 to implement cursor-pased pagination for REST API requests.


jeremystretch:
Two outstanding issues:

- REST API requests currently raise an exception when using cursor pagination. (This was true for the original PR #9716 as well.)

```
AssertionError: Using cursor pagination, but filter class OrderingFilter returned a `None` ordering.
```

- We need to figure out how best to handle the tentative regression under DeviceViewSet. For now, I've removed the StripCountAnnotationsPaginator introduced in #9374, but we'll need a strategy to replace it. (It's also worth noting that it probably should have been added to VirtualMachineViewSet as well, since VMs can also have config contexts assigned.)

jeremystretch:
Seems like there's a conflict using CursorPagination with OrderingFilter. Unfortunately I don't have time to track it down, so I'm going to close this.

Issue #8354, opened 2022-01-14
Python 3.9 is available in AppStream.


joubbi:
### Change Type

Correction

### Area

Installation instructions

### Proposed Changes

The installation instructions here: https://netbox.readthedocs.io/en/stable/installation/3-netbox/
mentions that CentOS 8 does not provide Python 3.7 or later via its native package manager.

It is possible to install Python 3.9 from AppStream.
I installed Netbox on the latest RHEL8 and was able to do `sudo yum install python39`.

If installing Python 3.9 this way the LDAP installation described here https://netbox.readthedocs.io/en/stable/installation/6-ldap/ will fail.
The `pip3 install django-auth-ldap` will fail unless you have done this `sudo yum install python39-devel`.

I think the instructions should be updated with this.

ryanmerolle:
@joubbi would you please take a stab at the documentation update?

PR #10767, opened 2022-10-27
Update installation instructions for RHEL.

joubbi:
CentOS is RHEL based so the heading makes more sense to be RHEL rather than CentOS. Also I don't think anyone will install netbox on CentOS anymore. RHEL is more likely.

I have installed netbox on RHEL8. The default version of Python 3 is too old, but it's possible to install Python 3.9 from a package. The default python3 is not replaced, so you have to change the pointer to point at python3.9 for the upgrade script to work.


jeremystretch:

joubbi:
https://github.com/netbox-community/netbox/issues/8354

PR #10774, opened 2022-10-27
Closes #9410: adding HEIF image format support for upload

### LINKED ISSUE #9410: HEIC image support
### NetBox version

v2.9.8

### Feature type

### Proposed functionality

Allow support for uploading heic format images to netbox devices. 

### Use case

Most modern devices (phones, tablets, etc) used by technicians to take photos in the data center / build rooms take pictures in the HEIC format, https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format

By supporting HEIC uploads, it can save a step required to first convert to a JPG or other currently supported file type.

### Database changes

_No response_

### External dependencies

Will require `convert` or imagemagick or another utility to convert the heic to jpg. This can be an optional depdency and only allow heic uploads if imagemagick is available.

bigcat88:
Fixes: #9410

I added an import to `netbox/settings.py` because `configuration.py` doesn't seem like the right place for it.

### DIFF
--- a/netbox/dcim/constants.py
+++ b/netbox/dcim/constants.py
@@ -3,7 +3,7 @@ from django.db.models import Q
 from .choices import InterfaceTypeChoices
 
 # Exclude SVG images (unsupported by PIL)
-DEVICETYPE_IMAGE_FORMATS = 'image/bmp,image/gif,image/jpeg,image/png,image/tiff,image/webp'
+DEVICETYPE_IMAGE_FORMATS = 'image/bmp,image/gif,image/jpeg,image/png,image/tiff,image/webp,image/heic'
 
 
 #
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -15,7 +15,7 @@ from django.urls import reverse
 from django.utils.safestring import mark_safe
 
 try:
-    from pillow_heif import HeifImageFile, HeifImagePlugin
+    from pi_heif import HeifImageFile, HeifImagePlugin
 except ImportError:
     HeifImagePlugin = HeifImageFile = None
 
@@ -28,6 +28,7 @@ from netbox.models import OrganizationalModel, NetBoxModel
 from utilities.choices import ColorChoices
 from utilities.fields import ColorField, NaturalOrderingField
 from .device_components import *
+
 
 __all__ = (
     'Device',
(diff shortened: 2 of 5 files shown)


kkthxbye-code:
I'm not sure this actually implements the requested feature. It allows the upload, but no conversion is made, so uploaded images are only viewable on browsers supporting HEIC, which currently is none according to google.

I might be missing something though.

bigcat88:
Safari supports. Do you need a converting to jpeg before saving? Can add that if needed.

kkthxbye-code:
> Safari supports.

Safari on my iPhone 13 Pro does not show HEIC images on websites, just tested. caniuse seems to agree: https://caniuse.com/?search=heic

> Do you need a converting to jpeg before saving? Can add that if needed.

From the original issue:

> Will require convert or imagemagick or another utility to convert the heic to jpg. This can be an optional depdency and only allow heic uploads if imagemagick is available.

I certainly do not think that it makes sense to allow uploads in a format that cannot be displayed in any browser, at least not without conversion.

bigcat88:
@kkthxbye-code done!

kkthxbye-code:
@bigcat88 - Thank you! Did a quick test and it works great here.

@jeremystretch - I'll request a review from you as well, not sure if this can be put into a patch release or not. Do note that @bigcat88 is the maintainer of [pillow_heif](https://github.com/bigcat88/pillow_heif) - not that I think it matters.

jeremystretch:
I'm sorry but we won't be able to support this approach. Any changes necessary to accommodate a new image format need to be made to the field itself, rather than to individual models. Ideally, this should be done upstream in Django rather than by introducing a new custom field.

bigcat88:
@jeremystretch did not see any new custom fields here, but no problems, close PR then.

jeremystretch:
> @jeremystretch did not see any new custom fields here

Yes, that's my point. You've attached to the models logic which pertains only to the specific field. This should be done in a custom subclass of Django's ImageField, if not natively within Django itself. Have you pursued introducing HEIF support upstream in Django?

bigcat88:
Didn't even have that idea.

jeremystretch:
Thanks for the effort @bigcat88, however as explained [here](https://github.com/netbox-community/netbox/issues/9410#issuecomment-1303532910) this isn't something we're willing to own in NetBox itself.

Issue #9623, opened 2022-06-28
Saved Filters


jasonyates:
### NetBox version

3.2.5

### Feature type

New functionality

### Proposed functionality

After building a filter it would be useful to be able to save it for future re-use, give it a name and the option to make it public. Public filters would then be viewable by all logged in to Netbox.

Proposing a 3rd tab alongside Results | Filters - perhaps 'Saved Filters'. Or possibly a section within the existing Filters tab to show any saved filters.

UI Suggestion for model list view

<img width="693" alt="Screenshot 2022-06-29 at 10 45 03" src="https://user-images.githubusercontent.com/242483/176406541-0055dc0d-d326-45f2-ae12-474fcaa5cd20.png">

Filters would not need to be modifiable, once saved, the user that created the filter can delete it from the saved filters list. When a user is removed from the system, the filters they created should also be removed.

### Use case

I often find myself having to build the same filter repeatedly in Netbox as well as having to remember to share links with the team when I build a useful filter. Having the option to save a filter in Netbox and share it with my users would save time.

### Database changes

A new model would need to be created to store the saved filters. Model would require a relation to the object type e.g. dcim.devices, filter attributes, friendly name, is public and a relation to the user that created it (to allow that user to delete it).

jeremystretch:
@jasonyates it sounds like a neat idea, but could you spend some more time fleshing out the proposal please? You haven't noted anything under database changes, and I'd assume this would require at least a new model.

jasonyates:
@jeremystretch Done. I included a proposed mockup of a piece of the UI too.

jeremystretch:
Expanding on the proposed model a bit, we probably need:

* Name
* User (nullable FK)
* Description (optional)
* Object type(s?)
* Query parameters (JSON data)
* Shared (boolean)

If `shared` is true, the filter will be made available for other users; otherwise, only the owner will see it listed as an available option when filtering an object list. However, it should still appear as visible in the saved filters list.

I've proposed making the `user` field nullable, so that the saved filter remains intact if the owning user is deleted.

We still need to determine how a saved filter would be applied in practice. I see two options:

* Reference the saved filter directly when making a query, e.g. `GET /dcim/sites/?filter=my_saved_filter`
* Use the filter to generate a query URL in the UI, so that when the user clicks "My saved filter," a request is made for e.g. `GET /dcim/sites/?status=planned&tenant=blah`

[...]

PR #10787, opened 2022-10-28
Closes: #7854 - Add VDC/Instances/etc

### LINKED ISSUE #7854: Virtual Device Context
### NetBox version

v3.0.10

### Feature type

New functionality

### Proposed functionality

For lack of a better term, it would be ideal to support the common model of virtual device context.  This should **NOT** model a specific vendor, but the common functionality of virtual device context in that interfaces can be assigned to them (in addition to VRFs if implemented #7852)

### Use case

Platform specific implementations to look at to understand the commonly shared functionality:
- Fortinet Virtual Domains (vDOMs)
- Juniper Virtual Router Instances
- Cisco Virtual Device Context
- Palo Alto Virtual Systems

### Database changes

_No response_

### External dependencies

_No response_

DanSheps:
### Closes: #7854

This PR adds VDC/Instance/etc support to NetBox

* Adds VDC Model
* Adds vdc M2M field on Interfaces
* Adds vdc_type field to device_type

Not stuck on the names, could be renamed "VirtualDeviceInstance" easily enough to be more encompassing and less Ciscoy


netbox/dcim/models/devices.py
```diff
@@ -1129,3 +1136,88 @@ def delete(self, *args, **kwargs):
...
+    )
+    status = models.CharField(
+        max_length=50,
+        blank=True,
+        choices=VirtualDeviceContextStatusChoices,
+    )
+    identifier = models.PositiveSmallIntegerField(
+        blank=True,
+        null=True,
+        validators=[MaxValueValidator(255)]
```
jeremystretch, line 1160: Do we want to be really strict about the maximum? Just wondering if other implementations might use higher IDs.

netbox/dcim/models/devices.py
```diff
@@ -1129,3 +1136,88 @@ def delete(self, *args, **kwargs):
...
+    )
+    primary_ip4 = models.OneToOneField(
```
jeremystretch, line 1162: Do we need primary IPv4/v6 attributes for a VDC? Seems like maybe overkill.

netbox/dcim/models/devices.py
```diff
@@ -128,6 +129,12 @@ class DeviceType(NetBoxModel, WeightMixin):
         choices=DeviceAirflowChoices,
         blank=True
     )
+    vdc_type = models.CharField(
```
jeremystretch, line 132: Not sure it makes sense to tie the VDC type to the DeviceType. I see two potential scenarios:

1. The VDC type is defined by the platform, and thus not configurable, in which case there is no need to have an attribute.
2. The VDC type _is_ configurable, which would be done per instantiated device rather than on the parent type.

netbox/dcim/choices.py
```diff
@@ -1399,3 +1399,74 @@ class PowerFeedPhaseChoices(ChoiceSet):
...
+    STATUS_OFFLINE = 'offline'
+
+    CHOICES = [
+        (STATUS_PLANNED, 'Planned', 'cyan'),
+        (STATUS_ACTIVE, 'Active', 'green'),
+        (STATUS_OFFLINE, 'Offline', 'red'),
+    ]
+
+
+class VirtualDeviceContextTypeChoices(ChoiceSet):
```
jeremystretch, line 1421: If we do end up needing to track the VDC type, it would probably make sense for these to be user-defined, as we can't reasonably anticipate every vendor implementation. (However as I commented elsewhere we can probably just infer the VDC type from the assigned platform.)

netbox/dcim/signals.py
```diff
@@ -123,3 +125,15 @@ def nullify_connected_endpoints(instance, **kwargs):
...
     for cablepath in CablePath.objects.filter(_nodes__contains=instance.cable):
         cablepath.retrace()
+
+
+@receiver(m2m_changed, sender=Interface.vdcs.through)
+def enforce_vdc_type_restrictions(instance, **kwargs):
+    if 'action' == 'post_add':
+        device = instance.device
+        if device.device_type.vdc_type not in [VirtualDeviceContextTypeChoices.CISCO_ASA_CONTEXT, VirtualDeviceContextTypeChoices.CISCO_FTD_INSTANCE] \
+                and len(instance.vdcs) > 1:
```
jeremystretch, line 135: This sort of validation should take place in the model's `clean()` method.

jeremystretch:
As noted in #7854, some VDC implementations permit the assignment of an interface to multiple VDCs whereas others do not. This might be something we want to track per VDC instance as a configurable attribute.

netbox/dcim/models/devices.py
DanSheps: So, from my experience (Cisco), we have:

ASA: "Context"
FTD: Instance
Nexus: VDC

In each one, both the Chassis has a IP (Device primary_ip4) as well as each instance/virtual context (VDC: manangement IP, ASA: management IP, FTD: Instance management IP & Diagnostic IP).

Since each have a separate management plane, I think it would be realistic to have the primary IP.

netbox/dcim/signals.py
DanSheps: M2M set operations are performed after the model's save(), so it wouldn't really hit the mdoels clean() method either.  You would see old data, but if you are adding a interface it would bypass the check.

We could remove this, as the form clean should catch it, we would have to do something different for the API however.

netbox/dcim/choices.py
DanSheps: Instead of using this..  Could we instead have a field for "Allow interface sharing within virtual device contexts" on the device type?  Default it to none, which would be either Yes or No.

netbox/dcim/models/devices.py
DanSheps: I think if we go with scenario 1, the only problem is we need to have a field to define whether interface sharing is allowed, unless we just don't worry about interface sharing enforcement.

With #2, I thought about having VDC type on the VDC model, (I didn't think about on the device model), however **most** platforms only use a certain technology, as seen in the Issue and those device types are pigeon holed into that VDC technology.  I think #1 is most applicable.

DanSheps:
> As noted in https://github.com/netbox-community/netbox/issues/7854, some VDC implementations permit the assignment of an interface to multiple VDCs whereas others do not. This might be something we want to track per VDC instance as a configurable attribute.

I can see a couple of options for tracking this, by a atribute on:

1. Device Type
2. Device
3. VDC
4. Interface
5. Interface.vdcs.through

For DeviceType, I think this makes the most sense.  It allows you to define your Device as allowing interface sharing.  Device seems like an awkward place to track this type of setting.  VDC, typically either the platform (Nexus 7700 for example, which doesn't) allows VDC interface sharing or not.  Interface, I think would be a great place for a "Shared Interface" checkbox but I don't think it would make sense to use this to track whether the platform allows sharing.  Similar to "Device", I don't think the through model is the best place to track this

I think a combination of "Can interface share" on the device type and a "Shared inteface" attribute on the Interface is the best way go.  We could infer whether it is a shared interface too, just based on how many VDC's are assigned (>1 == Shared), but there could be a case where the interface is set to allow sharing but currently only one device context is created.

Thoughts?

netbox/dcim/models/devices.py
jeremystretch: I suspect just sticking with PositiveSmallIntegerField and no additional validators is sufficient.

netbox/dcim/models/devices.py
jeremystretch: IMO we should keep things simple and go with option one, eschewing the concept of VDC type entirely until such need arises.

> the only problem is we need to have a field to define whether interface sharing is allowed, unless we just don't worry about interface sharing enforcement

For now we could ignore it, allowing an interface to be allocated to any number of VDCs, and rely on feedback from the beta to determine whether anything more robust is needed.

jeremystretch:
I think we should avoid over-rotating on this, and just stick with permitting the assignment of an interface to multiple VDCs. If there's a strong need for some conditional logic there, I'm sure we'll hear of it during the beta evaluation.

netbox/dcim/models/devices.py
```diff
@@ -1062,3 +1063,83 @@ def delete(self, *args, **kwargs):
...
+        related_name='vdcs',
+        blank=True,
+        null=True
+    )
+    name = models.CharField(
+        max_length=64
+    )
+    status = models.CharField(
+        max_length=50,
+        blank=True,
```
jeremystretch, line 1081: IMO `status` should be a required field, as with other models.

netbox/dcim/models/devices.py
```diff
@@ -1062,3 +1063,83 @@ def delete(self, *args, **kwargs):
...
+            models.UniqueConstraint(
+                fields=('device', 'identifier',),
+                name='%(app_label)s_%(class)s_device_identifiers',
+                violation_error_message="A VDC with this identifier already exists on this device."
+            ),
+            models.UniqueConstraint(
+                fields=('device', 'name',),
+                name='%(app_label)s_%(class)s_name',
+                violation_error_message="A VDC with this name already exists on this device."
+            ),
```
jeremystretch, line 1127: Let's avoid adding custom error messages except where needed.
```suggestion
            models.UniqueConstraint(
                fields=('device', 'identifier',),
                name='%(app_label)s_%(class)s_device_identifiers'
            ),
            models.UniqueConstraint(
                fields=('device', 'name',),
                name='%(app_label)s_%(class)s_name'
            ),
```

netbox/dcim/signals.py
```diff
@@ -1,10 +1,12 @@
...
 
-from django.db.models.signals import post_save, post_delete, pre_delete
+from django import forms
+from django.db.models.signals import post_save, post_delete, pre_delete, m2m_changed
 from django.dispatch import receiver
 
 from .choices import CableEndChoices, LinkStatusChoices
-from .models import Cable, CablePath, CableTermination, Device, PathEndpoint, PowerPanel, Rack, Location, VirtualChassis
+from .models import Cable, CablePath, CableTermination, Device, PathEndpoint, PowerPanel, Rack, Location, \
+    VirtualChassis, VirtualDeviceContext, Interface
```
jeremystretch, line 9: Looks like some extraneous imports were left here.

netbox/netbox/navigation/menu.py
```diff
@@ -58,6 +58,7 @@
             label='Devices',
             items=(
                 get_model_item('dcim', 'device', 'Devices'),
+                get_model_item('dcim', 'virtualdevicecontext', 'Virtual Device Contexts'),
```
jeremystretch, line 61: IMO this should go somewhere else in the menu. Seems random to have a virtual resource listed between two physical ones.

netbox/dcim/models/devices.py
```diff
@@ -1062,3 +1063,83 @@ def delete(self, *args, **kwargs):
...
+    )
+    name = models.CharField(
+        max_length=64
+    )
+    status = models.CharField(
+        max_length=50,
+        blank=True,
+        choices=VirtualDeviceContextStatusChoices,
+    )
+    identifier = models.PositiveSmallIntegerField(
```
jeremystretch, line 1084: Should probably add a `help_text` here to explain the field's intended purpose.

netbox/templates/dcim/virtualdevicecontext.html
```diff
@@ -0,0 +1,76 @@
+{% extends 'generic/object.html' %}
+{% load helpers %}
+{% load plugins %}
+{% load render_table from django_tables2 %}
+
+{% block breadcrumbs %}
+  <li class="breadcrumb-item"><a href="{% url 'dcim:devicerole_list' %}">Device Roles</a></li>
```
jeremystretch, line 7: Breadcrumb link is incorrect

netbox/templates/dcim/virtualdevicecontext.html
```diff
@@ -0,0 +1,76 @@
...
+{% endblock %}
+
+{% block extra_controls %}
+  {% if perms.dcim.add_interface %}
+    <a href="{% url 'dcim:interface_add' %}?device={{ object.device.pk }}&vdc={{ object.pk }}" class="btn btn-sm btn-primary">
+      <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add Interface
+    </a>
+  {% endif %}
+{% endblock extra_controls %}
```
jeremystretch, line 16: This seems confusing, as interfaces are _added_ to devices and _assigned_ to VDCs. It seems unlikely that a user will want to create a new interface on a device while viewing a VDC. (Also the `vdc` parameter in the URL should be `vdcs`.)

netbox/dcim/forms/bulk_edit.py
```diff
@@ -1398,3 +1399,20 @@ class InventoryItemRoleBulkEditForm(NetBoxModelBulkEditForm):
...
+        required=False
+    )
+    status = forms.ChoiceField(
+        required=False,
+        choices=add_blank_choice(VirtualDeviceContextStatusChoices),
+        widget=StaticSelect()
+    )
+    model = VirtualDeviceContext
+    fieldsets = (
+        (None, ('device',)),
```
jeremystretch, line 1416: Missing status and tenant fields

netbox/dcim/tests/test_filtersets.py
```diff
@@ -4254,4 +4254,121 @@ def test_connected(self):
...
+        interfaces[5].vdcs.set([vdcs[4], vdcs[5]])
+        interfaces[6].vdcs.set([vdcs[0]])
+        interfaces[7].vdcs.set([vdcs[1]])
+        interfaces[8].vdcs.set([vdcs[2]])
+        interfaces[9].vdcs.set([vdcs[3]])
+        interfaces[10].vdcs.set([vdcs[4]])
+        interfaces[11].vdcs.set([vdcs[5]])
+
+
+class VirtualDeviceContextTestCase(VirtualDeviceContextBaseTestCase, TestCase, ChangeLoggedFilterSetTests):
```
jeremystretch, line 4342: Needs tests for `device_id` and `has_primary_ip`

netbox/dcim/tests/test_filtersets.py
```diff
@@ -4254,4 +4254,121 @@ def test_connected(self):
...
+    def test_device(self):
+        params = {'device': ['Device 1', 'Device 2']}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 6)
+
+    def test_status(self):
+        params = {'status': ['active']}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 3)
+
+
+class VirtualDeviceContextInterfaceTestCase(VirtualDeviceContextBaseTestCase, TestCase, ChangeLoggedFilterSetTests):
```
jeremystretch, line 4355: I'm not sure about adding these here. IMO every test for interface filters should go under the `InterfaceTestCase` class. I realize the advantage of making a separate class here is to reuse the object creation code, however I worry that this will needlessly complicate potential future evaluations. Also, it should be fairly simple to extend InterfaceTestCase with some VDC assignments.

netbox/dcim/tests/test_filtersets.py
```diff
@@ -4254,4 +4254,121 @@ def test_connected(self):
...
+        )
+        for location in locations:
+            location.save()
+
+        racks = (
+            Rack(name='Rack 1', site=sites[0], location=locations[0]),
+            Rack(name='Rack 2', site=sites[1], location=locations[1]),
+            Rack(name='Rack 3', site=sites[2], location=locations[2]),
+        )
+        Rack.objects.bulk_create(racks)
```
jeremystretch, line 4282: We can omit location and rack assignment for these tests.

netbox/dcim/forms/filtersets.py
```diff
@@ -728,6 +729,37 @@ class DeviceFilterForm(
     tag = TagFilterField(model)
 
 
+class VirtualDeviceContextFilterForm(
+    TenancyFilterForm,
+    NetBoxModelFilterSetForm
+):
+    model = VirtualDeviceContext
+    fieldsets = (
+        (None, ('q', 'tag')),
```
jeremystretch, line 738: For #9623
```suggestion
        (None, ('q', 'filter', 'tag')),
```

netbox/dcim/forms/filtersets.py
```diff
@@ -728,6 +729,37 @@ class DeviceFilterForm(
     tag = TagFilterField(model)
 
 
+class VirtualDeviceContextFilterForm(
+    TenancyFilterForm,
+    NetBoxModelFilterSetForm
+):
+    model = VirtualDeviceContext
+    fieldsets = (
```
jeremystretch, line 737: Missing the `status` field

netbox/dcim/tables/devices.py
```diff
@@ -884,3 +885,44 @@ class Meta(NetBoxTable.Meta):
...
+class VirtualDeviceContextTable(TenancyColumnsMixin, NetBoxTable):
+    name = tables.Column(
+        linkify=True
+    )
+    indentifier = tables.Column()
+    device = tables.TemplateColumn(
+        order_by=('_name',),
+        template_code=DEVICE_LINK
+    )
+    status = tables.Column()
```
jeremystretch, line 899:
```suggestion
    status = columns.ChoiceFieldColumn()
```

netbox/dcim/tables/devices.py
```diff
@@ -884,3 +885,44 @@ class Meta(NetBoxTable.Meta):
             'last_updated',
         )
         default_columns = ('pk', 'name', 'domain', 'master', 'member_count')
+
+
+class VirtualDeviceContextTable(TenancyColumnsMixin, NetBoxTable):
+    name = tables.Column(
+        linkify=True
+    )
+    indentifier = tables.Column()
```
jeremystretch, line 894:
```suggestion
```

netbox/dcim/tests/test_models.py
```diff
@@ -588,3 +588,30 @@ def test_cable_cannot_terminate_to_a_wireless_interface(self):
         cable = Cable(a_terminations=[self.interface2], b_terminations=[wireless_interface])
         with self.assertRaises(ValidationError):
             cable.clean()
+
+
+class CableTestCase(TestCase):
```
jeremystretch, line 593: I assume this was added by mistake? (CableTestCase is declared above.)

netbox/dcim/tests/test_views.py
```diff
@@ -3076,3 +3076,89 @@ def test_trace(self):
 
         response = self.client.get(reverse('dcim:powerfeed_trace', kwargs={'pk': powerfeed.pk}))
         self.assertHttpStatus(response, 200)
+
+# TODO: VDC Test Cases
```
jeremystretch, line 3080: Is this still a TODO item?

netbox/dcim/tests/test_views.py
```diff
@@ -3076,3 +3076,89 @@ def test_trace(self):
...
+        )
+        DeviceRole.objects.bulk_create(deviceroles)
+
+        platforms = (
+            Platform(name='Platform 1', slug='platform-1'),
+            Platform(name='Platform 2', slug='platform-2'),
+        )
+        Platform.objects.bulk_create(platforms)
+
+        devices = (
```
jeremystretch, line 3124: You can use the `create_test_device` utility to avoid much of this boilerplate.

netbox/dcim/views.py
```diff
@@ -3561,3 +3570,61 @@ class PowerFeedBulkDeleteView(generic.BulkDeleteView):
...
+    queryset = VirtualDeviceContext.objects.all()
+
+
+class VirtualDeviceContextBulkImportView(generic.BulkImportView):
+    queryset = VirtualDeviceContext.objects.all()
+    model_form = forms.VirtualDeviceContextCSVForm
+    table = tables.VirtualDeviceContextTable
+
+
+class VirtualDeviceContextBulkRenameView(generic.BulkRenameView):
```
jeremystretch, line 3614: This doesn't appear to be accessible via the UI. (There's no "rename" button.)

netbox/templates/dcim/virtualdevicecontext.html
```diff
@@ -0,0 +1,76 @@
...
+            <td>
+              {{ object.primary_ip6|placeholder }}
+            </td>
+          </tr>
+        </table>
+      </div>
+    </div>
+    {% plugin_left_page object %}
+	</div>
+	<div class="col col-md-6">
```
jeremystretch, line 57:
```suggestion
  </div>
  <div class="col col-md-6">
```

jeremystretch (changes requested):

netbox/dcim/models/devices.py
DanSheps: Going to make this change now

netbox/netbox/navigation/menu.py
DanSheps: Moved it just below virtual chassis

netbox/dcim/models/devices.py
DanSheps: > Unique identifier provided by the platform being virtualized (Example: Nexus VDC Identifier)

netbox/templates/dcim/virtualdevicecontext.html
DanSheps: We can remove it for now, however adding Vlan interfaces could definitely be 1 way this could be useful.  We can re-add based on feedback.

netbox/dcim/tests/test_filtersets.py
DanSheps: Good point, I wrapped it into the InterfaceTestCase.

netbox/dcim/tests/test_models.py
DanSheps: Eh, I think I half started on the tests and forgot to finish them.  Should be updated soon.

netbox/dcim/tests/test_views.py
DanSheps: Thanks, did not know about this one.

DanSheps:
Once thing I thought of...  We may have a situation now where an interface needs to be created in a VDC but is **not** shared, we might need to investigate the uniqueness parameters for interfaces now and add VDC when VDC is not null as a uniqueness component.

Thoughts @jeremystretch?

jeremystretch:
Interfaces simply need to have unique names per device/VM; no way around that, and I think that's a reasonable limitation.

jeremystretch (changes requested):

netbox/dcim/forms/bulk_edit.py
```diff
@@ -1398,3 +1399,20 @@ class InventoryItemRoleBulkEditForm(NetBoxModelBulkEditForm):
...
+        required=False
+    )
+    status = forms.ChoiceField(
+        required=False,
+        choices=add_blank_choice(VirtualDeviceContextStatusChoices),
+        widget=StaticSelect()
+    )
+    model = VirtualDeviceContext
+    fieldsets = (
+        (None, ('device', 'status', 'tenant')),
```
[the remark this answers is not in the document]
DanSheps: This should be fixed to change the failing test.

jeremystretch (approved):
Great work, thanks @DanSheps!

Issue #7376, opened 2021-09-27
Support the assignment of tags to objects during bulk CSV import


jeremystretch:
### NetBox version

v3.0.3

### Feature type

New functionality

### Proposed functionality

Enable the assignment of tags to objects during bulk import of CSV data. This was originally proposed specifically for IP addresses under #6420, but should apply to all objects which support tags.

### Use case

This would enable users to assign tags directly to objects at the time of creation, and avoid needing to edit objects in bulk later to add the tags.

sdktr:
Probably above my knowledge level, but trying to get this to work and will submit fore review here

beezed1:
Hi,

Are there any thoughts on when (or if) importing of tags would be added to Netbox?

Thanks
Phil

sdktr:
Any hints @jeremystretch why my attempt at solving this results in an 'always required' tags column in the CSVimport?
https://github.com/sdktr/netbox/commit/1a73410d8d840f8c22da0852b1f04144fc1c94e9

PR #10802, opened 2022-10-31
7376 csv tags

arthanson:
### Fixes: #7376 

Adds tags as an allowed (optional) field to CSV import.  Tried import and update on several different models and it worked.  Note: Have not tested on all the effected models as it should be common code.

### DIFF
--- a/netbox/netbox/forms/base.py
+++ b/netbox/netbox/forms/base.py
@@ -6,9 +6,8 @@ from django.db.models import Q
 from extras.choices import CustomFieldFilterLogicChoices, CustomFieldTypeChoices, CustomFieldVisibilityChoices
 from extras.forms.mixins import CustomFieldsMixin, SavedFiltersMixin
 from extras.models import CustomField, Tag
-from taggit.forms import TagField
 from utilities.forms import BootstrapMixin, CSVModelForm
-from utilities.forms.fields import DynamicModelMultipleChoiceField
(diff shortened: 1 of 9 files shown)


jeremystretch (changes requested):
A couple issues:

- Looks like CSV forms in the circuits app were not updated.
- Using taggit's `TagField` automatically creates invalid tags, which is not desired behavior. Attempting to import an object with one or more invalid tags should trigger a validation failure.
- This needs a test for both valid and invalid imports.

arthanson:
@jeremystretch these issues should be resolved now.

netbox/netbox/forms/base.py
```diff
@@ -71,6 +76,15 @@ def _get_custom_fields(self, content_type):
     def _get_form_field(self, customfield):
         return customfield.to_form_field(for_csv_import=True)
 
+    def clean_tags(self):
+        data = self.cleaned_data['tags']
+        existing_tags = Tag.objects.values_list('slug', flat=True)
```
jeremystretch, line 81: This is less than ideal as it will query all tags from the database for each row being imported. Instead of using TagField in the CSV form, we should use a model choice field, which handles the validation of existing objects natively. We have CSVModelChoiceField for individual objects; the same approach can likely be used for selecting multiple objects. I'll take a look.

jeremystretch: Introduced CSVModelMultipleChoiceField to handle this. Will be suitable for the assignment of other M2M relationships as well.

PR #10843, opened 2022-11-04
10653 log failed login attempts on INFO

### LINKED ISSUE #10653: Make it possible to log failed login attempts
### NetBox version

v3.2.9

### Feature type

New functionality

### Proposed functionality

I would like to have netbox log every login attempt that is made, however this doesn't seem possible, i currently have this logging config

```
LOGGING = {
    'version': 1,
    'disable_existing_loggers': False,
    'handlers': {
        'console': {
            'class': 'logging.StreamHandler',
        },
    },
    'loggers': {
        'netobx.*': {
        'handlers': ['console'],
        'level': 'INFO',
        }
    },
}
```
This config logs every time a user logs in successfully but not when a user fails to authenticate.

### Use case

This would make it possible to see if netbox is currently being bruteforced, this would be especially helpful to deployments in the cloud where it is accessible to the web.

### Database changes

N/A

### External dependencies

N/A

arthanson:
### Fixes: #10653 

@jeremystretch Not completely sure of what is the best way here.  The log was done for failed login attempts but at DEBUG level which would be fairly spammy to have on, changed to INFO level also included failed username - I don't think that would violate personal-identifiable-information security issues...  Probably up for discussion if this makes sense to change (from a logging standpoint if that is too much info for some people, and security if the username should be shown).

### DIFF
--- a/netbox/users/apps.py
+++ b/netbox/users/apps.py
@@ -1,7 +1,7 @@
 from django.apps import AppConfig
 
 
-class HomeConfig(AppConfig):
+class UsersConfig(AppConfig):
     name = 'users'
 
     def ready(self):
--- a/netbox/users/apps.py
+++ b/netbox/users/apps.py
@@ -1,0 +1,8 @@
+from django.apps import AppConfig
+
+
+class HomeConfig(AppConfig):
+    name = 'users'
+
+    def ready(self):
+        import users.signals
--- a/netbox/users/signals.py
+++ b/netbox/users/signals.py
@@ -6,5 +6,5 @@ from django.contrib.auth.signals import user_login_failed
 @receiver(user_login_failed)
 def log_user_login_failed(sender, credentials, request, **kwargs):
     logger = logging.getLogger('netbox.auth.login')
-    username = credentials.get("username", None)
+    username = credentials.get("username")
     logger.info(f"Failed login attempt for username: {username}")
--- a/netbox/users/signals.py
+++ b/netbox/users/signals.py
@@ -1,0 +1,10 @@
+import logging
+from django.dispatch import receiver
+from django.contrib.auth.signals import user_login_failed
+
+
+@receiver(user_login_failed)
+def log_user_login_failed(sender, credentials, request, **kwargs):
+    logger = logging.getLogger('netbox.auth.login')
+    username = credentials.get("username", None)
+    logger.info(f"Failed login attempt for username: {username}")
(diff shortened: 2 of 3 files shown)


jeremystretch:
Let's keep the debug line as it is, I think. I would think it's already possible to log authentication failures reported by Django using the native [logging functionality](https://docs.netbox.dev/en/stable/configuration/system/#logging); is this not the case?

jeremystretch:
I did a bit of digging and surprisingly it seems that Django _doesn't_ log authentication failures out of the box. Still, we'd want to log the success/failure result somewhere closer to the backend itself, rather than on the login form. (The exists logging statements are merely for debugging the form itself.)

arthanson:
Changed to use the user_login_failed signal issued from within Django.

arthanson:
@jeremystretch ready for re-review.

netbox/users/signals.py
```diff
@@ -0,0 +1,10 @@
+import logging
+from django.dispatch import receiver
+from django.contrib.auth.signals import user_login_failed
+
+
+@receiver(user_login_failed)
+def log_user_login_failed(sender, credentials, request, **kwargs):
+    logger = logging.getLogger('netbox.auth.login')
+    username = credentials.get("username", None)
```
jeremystretch, line 9:
```suggestion
    username = credentials.get("username")
```

netbox/users/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+
+
+class HomeConfig(AppConfig):
```
jeremystretch, line 4: To keep consistent with the other apps:
```suggestion
class UsersConfig(AppConfig):
```

netbox/users/views.py
```diff
@@ -106,7 +106,7 @@ def post(self, request):
             return self.redirect_to_next(request, logger)
 
         else:
-            logger.debug("Login form validation failed")
+            logger.debug(f"Login form validation failed for username: {form['username'].value()}")
```
jeremystretch, line 109: We should be able to get the username from the cleaned data:
```suggestion
            logger.debug(f"Login form validation failed for username: {form.cleaned_data['username']}")
```

jeremystretch (changes requested):

jeremystretch:
> Changed to use the user_login_failed signal issued from within Django.

Nice!

netbox/users/views.py
arthanson: Not sure on this one - it is in a case where the form is not valid, from the docs "If your data does not validate, the cleaned_data dictionary contains only the valid fields:" so it is potential that the username is not in the cleaned_data in this case.

PR #10905, opened 2022-11-13
Closes #10904: Added Colors to SVG for Front and Rear Ports

### LINKED ISSUE #10904: Render Specified Front and Rear Port Colors During Cable Trace
### NetBox version

v3.3.7

### Feature type

### Proposed functionality

Currently it is possible to specify colors for Front and Rear Ports. However, when running a cable trace, the port colors are not reflected in the rendered SVG.  Cable colors can be specified and are already rendered during cable traces, it only makes sense that the same logic should apply for the Front and Rear Ports. Additionally, when no color is specified, the port color should revert to the current default. 

## Current Functionality
Current Functionality Netbox

## Proposed Functionality
Proposed Functionality Netbox

### Use case

Sometimes specific patch panel ports and/or network drop keystone colors are used to help identify specific traces in the real world. Adding this functionality would not only be simple but would further help users visualize and identify traces.

### Database changes

None

### External dependencies

None

dvigne:
Fix for feature request 10904 thanks to @TheZackCodec

### Fixes: #10904

When rendering the SVG, checks if a custom color is defined and renders that color on the termination.

### DIFF
--- a/netbox/dcim/svg/cables.py
+++ b/netbox/dcim/svg/cables.py
@@ -166,6 +166,8 @@ class CableTraceSVG:
         """
         if hasattr(instance, 'parent_object'):
             # Termination
+            if(hasattr(instance, 'get_color') and instance.get_color()):
+                return instance.get_color()
             return 'f0f0f0'
         if hasattr(instance, 'device_role'):
             # Device
--- a/netbox/dcim/svg/cables.py
+++ b/netbox/dcim/svg/cables.py
@@ -166,9 +166,7 @@ class CableTraceSVG:
         """
         if hasattr(instance, 'parent_object'):
             # Termination
-            if(hasattr(instance, 'get_color') and instance.get_color()):
-                return instance.get_color()
-            return 'f0f0f0'
+            return getattr(instance, 'color', 'f0f0f0') or 'f0f0f0'
         if hasattr(instance, 'device_role'):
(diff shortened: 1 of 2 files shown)


netbox/dcim/svg/cables.py
```diff
@@ -166,6 +166,8 @@ def _get_color(cls, instance):
         """
         if hasattr(instance, 'parent_object'):
             # Termination
+            if(hasattr(instance, 'get_color') and instance.get_color()):
+                return instance.get_color()
             return 'f0f0f0'
```
jeremystretch, line 171: No need for a method here. You can just do:
```suggestion
            return getattr(instance, 'color', 'f0f0f0')
```

dvigne: This is much cleaner and I like the syntax a lot better, but what I'm finding is that the model has `blank=True` set for both the `FrontPort` and `RearPort` models and this leads to an empty string being stored in the database. This logic fails because it always has the `color` attribute, it just happens to be a blank string.

A quick solution for this would be change that one line to something like below, but I'm open to suggestions.

```suggestion
            return getattr(instance, 'color', 'f0f0f0') or 'f0f0f0'
```

jeremystretch: That works.

PR #10913, opened 2022-11-14
10902 add location to power feed form

### LINKED ISSUE #10902: Add location field to "Add Power Feed" form
### NetBox version

v3.3.5

### Feature type

### Proposed functionality

When creating a new Power Feed via the UI, the form only lets you select a rack, it would be great if the form would also expose a Location selector and then filter the available Racks based on that selection (similar to how the Site choices are limited by a selected Region).

### Use case

We have racks with the same name in the same site but within a different location. That makes it non-obvious which is the right rack to select.

### Database changes

Being able to specify a Location is already possible when using the Bulk Import feature, so we don't think this should require any database changes.

### External dependencies

n/a

arthanson:
### Fixes: #10902 

Adds location field to power feed form and filters rack dropdown based on it:

Monosnap Add a new power feed | NetBox 2022-11-14 09-38-47

### DIFF
--- a/netbox/dcim/forms/models.py
+++ b/netbox/dcim/forms/models.py
@@ -877,10 +877,18 @@ class PowerFeedForm(NetBoxModelForm):
             'site_id': '$site'
         }
     )
+    location = DynamicModelChoiceField(
+        queryset=Location.objects.all(),
+        required=False,
+        initial_params={
+            'racks': '$rack'
+        }
+    )
     rack = DynamicModelChoiceField(
         queryset=Rack.objects.all(),
         required=False,
         query_params={
+            'location_id': '$location',
             'site_id': '$site'
         }
     )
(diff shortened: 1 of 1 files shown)


netbox/dcim/forms/models.py
```diff
@@ -877,25 +877,33 @@ class PowerFeedForm(NetBoxModelForm):
             'site_id': '$site'
         }
     )
+    location = DynamicModelChoiceField(
+        queryset=Location.objects.all(),
+        required=False,
+        initial_params={
+            'racks': '$rack'
+        }
```
jeremystretch, line 885: Sorry, missed this earlier when we discussed: This should have `query_params` on it too, to limit the locations by the selected site:

```python
        query_params={
            'site_id': '$site'
        }
```

PR #10939, opened 2022-11-16
10748 Add 'Provider' to the circuit termination edit/view

### LINKED ISSUE #10748: Add 'Provider' to the circuit termination edit/view
### NetBox version

v3.3.5

### Feature type

### Proposed functionality

On the 'circuit' detail page, the Termination show only the Provider Network.
It would benefit if the 'Provider' is shown too
ex : https://demo.netbox.dev/circuits/circuits/14/  
The Demo repeated the Provider in the name of the ProviderNetwork.
Maybe that Provider network should return "ProviderNetwork (Provider)" as display string.

When editing an Termination, we can set the type to 'Provider Network', but we can't select the provider.
https://demo.netbox.dev/circuits/circuit-terminations/42/edit/?return_url=/circuits/circuits/14/

### Use case

The ProviderNetwork itself is not that descriptive. It only shows for who that network is used, but not by whom it is provided.
Adding this info to the circuit termination and the circuit list, would help to make that clear in one view

I have some ProviderNetworks that have the same name, but from different Providers. 
When selecting it from the dropdown, I have to guess for what Provider that is.

### Database changes

_No response_

### External dependencies

_No response_

PieterL75:

### DIFF
--- a/netbox/circuits/forms/models.py
+++ b/netbox/circuits/forms/models.py
@@ -158,6 +158,14 @@ class CircuitTerminationForm(NetBoxModelForm):
         },
         required=False
     )
+    provider_network_provider = DynamicModelChoiceField(
+        queryset=Provider.objects.all(),
+        required=False,
+        label='Provider',
+        initial_params={
+            'networks': 'provider_network'
+        }
+    )
     provider_network = DynamicModelChoiceField(
         queryset=ProviderNetwork.objects.all(),
         query_params={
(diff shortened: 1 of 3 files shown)


jeremystretch:
Thanks for this @PieterL75. I went ahead and tweaked the form fields a bit to fix initial loading of the assigned network's provider.

PieterL75:
Nice, I didn't git that to work, but the order is important apparently...

Those labels were added to distinct the purpose better. Otherwise you have 2 times the provider field in the form

jeremystretch:
> Those labels were added to distinct the purpose better. Otherwise you have 2 times the provider field in the form
@jeremystretch

Yeah, that's fine; the second one is nested under the "provider network" tab so I think it's purpose is clear.

PR #10954, opened 2022-11-17
Fixes #10950 - Add filtering and validation to primary_ip4/primary_ip6 VirtualDeviceContext model

### LINKED ISSUE #10950: Primary IPv6 for VDC filters IPv4
### NetBox version

v3.4-beta1

### Python version

3.10

### Steps to Reproduce

1. Click on Add on the VDC page
2. Choose `ncsu-coreswitch1` on the device drop-down
3. Click on Primary IPv6 dropdown

### Expected Behavior

Only IPv6 addresses should be displayed

### Observed Behavior

IPv4 addresses are displayed
image

DanSheps:
### Fixes: #10950

* Adds filtering to VirtualDeviceContext model_form
* Adds validation to VirtualDeviceContext model
* Adds validation to IPAddress model

### DIFF
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -1705,6 +1705,22 @@ class VirtualDeviceContextForm(TenancyForm, NetBoxModelForm):
             'rack_id': '$rack',
         }
     )
+    primary_ip4 = DynamicModelChoiceField(
+        queryset=IPAddress.objects.all(),
+        required=False,
+        query_params={
+            'device_id': '$device',
+            'family': 4,
+        }
+    )
+    primary_ip6 = DynamicModelChoiceField(
+        queryset=IPAddress.objects.all(),
+        required=False,
+        query_params={
+            'device_id': '$device',
+            'family': 6,
+        }
+    )
 
     fieldsets = (
         ('Assigned Device', ('region', 'site_group', 'site', 'location', 'rack', 'device')),
@@ -1720,6 +1736,4 @@ class VirtualDeviceContextForm(TenancyForm, NetBoxModelForm):
         ]
         widgets = {
             'status': StaticSelect(),
-            'primary_ip4': StaticSelect(),
-            'primary_ip6': StaticSelect(),
         }
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -1705,22 +1705,6 @@ class VirtualDeviceContextForm(TenancyForm, NetBoxModelForm):
             'rack_id': '$rack',
         }
     )
-    primary_ip4 = DynamicModelChoiceField(
-        queryset=IPAddress.objects.all(),
-        required=False,
-        query_params={
-            'device_id': '$device',
-            'family': 4,
-        }
-    )
-    primary_ip6 = DynamicModelChoiceField(
-        queryset=IPAddress.objects.all(),
-        required=False,
-        query_params={
-            'device_id': '$device',
-            'family': 6,
-        }
-    )
 
     fieldsets = (
         ('Assigned Device', ('region', 'site_group', 'site', 'location', 'rack', 'device')),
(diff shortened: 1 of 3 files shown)


netbox/dcim/forms/model_forms.py
```diff
@@ -1705,6 +1705,22 @@ class VirtualDeviceContextForm(TenancyForm, NetBoxModelForm):
             'rack_id': '$rack',
         }
     )
+    primary_ip4 = DynamicModelChoiceField(
```
jeremystretch: Need to set labels on the form fields to ensure their proper display.

netbox/dcim/models/devices.py
```diff
@@ -1175,3 +1175,25 @@ def primary_ip(self):
             return self.primary_ip4
         else:
             return None
+
+    def clean(self):
+        super().clean()
+        vc_interfaces = self.device.vc_interfaces(if_master=False)
```
jeremystretch: Shouldn't we limit this to interfaces assigned to the VDC?

netbox/ipam/models/ip.py
```diff
@@ -866,7 +866,7 @@ def clean(self):
 
         # Check for primary IP assignment that doesn't match the assigned device/VM
         if self.pk:
-            for cls, attr in ((Device, 'device'), (VirtualMachine, 'virtual_machine')):
+            for cls, attr in ((Device, 'device'), (VirtualMachine, 'virtual_machine'), (VirtualDeviceContext, 'device')):
```
jeremystretch:
```suggestion
            for cls, attr in ((Device, 'device'), (VirtualMachine, 'virtual_machine'), (VirtualDeviceContext, 'vdc')):
```

jeremystretch (changes requested):

netbox/dcim/models/devices.py
DanSheps: Yes, I have went ahead and made this change.

netbox/ipam/models/ip.py
DanSheps: The reverse attribute is vdcs and is a manytomany manager field.  I had to do some thing different but perhaps you can take a look and maybe make a suggestion to clean this up better.

netbox/ipam/models/ip.py
```diff
@@ -866,9 +866,16 @@ def clean(self):
 
         # Check for primary IP assignment that doesn't match the assigned device/VM
         if self.pk:
-            for cls, attr in ((Device, 'device'), (VirtualMachine, 'virtual_machine')):
+            for cls, attr in ((Device, 'device'), (VirtualMachine, 'virtual_machine'), (VirtualDeviceContext, 'vdcs')):
```
arthanson, line 869: This might be a bit cleaner - I think line 872 comparison of .all() != parent might be wrong, should be first()?  I'm assuming the m2m check only needs to be done if 'vdcs', device and virtual_machine attrs aren't m2m.
```
        if self.pk:
            for cls, attr in ((Device, 'device'), (VirtualMachine, 'virtual_machine'), (VirtualDeviceContext, 'vdcs')):
                parent = cls.objects.filter(Q(primary_ip4=self) | Q(primary_ip6=self)).first()
                if parent and hasattr(self.assigned_object, attr):
                    if attr == 'vdcs'and getattr(self.assigned_object, attr).first() != parent:
                        raise ValidationError({
                            'interface': f"IP address is primary for {cls._meta.model_name} {parent} but "
                                         f"not assigned to it!"
                        })
                    elif getattr(self.assigned_object, attr, None) != parent:
                        # Check for a NAT relationship
                        if not self.nat_inside or getattr(self.nat_inside.assigned_object, attr, None) != parent:
                            raise ValidationError({
                                'interface': f"IP address is primary for {cls._meta.model_name} {parent} but "
                                             f"not assigned to it!"
                            })

```

arthanson (changes requested):

netbox/ipam/models/ip.py
arthanson: also, move the parent = cls.objects outside the loop:
```
        if self.pk:
            parent = cls.objects.filter(Q(primary_ip4=self) | Q(primary_ip6=self)).first()
            for cls, attr in ((Device, 'device'), (VirtualMachine, 'virtual_machine'), (VirtualDeviceContext, 'vdcs')):

```

Issue #8749, opened 2022-02-25
Clone should also prepopulate custom fields


skvedo:
### NetBox version

v3.1.8

### Feature type

Change to existing functionality

### Proposed functionality

When cloning entity, values from custom fields should be also populated in form.

### Use case

Our organization have defined custom fields for prefix with abuse contact to particular tenant IT support staff. When assigning new prefix you usually set these "same as that other segment" so using clone button is obvious approach.

### Database changes

Might be good idea add this fields to prefix form.
We defined 2 fields:
* security abuse contact for case of incident which should be taken case ASAP. Usually creates an incident ticket
* administrative contact if you need contact for particular tenant. Usually this is pointed to teams mailing list.

IdRatherStand:
I also have users coming to me within our org asking if there is an option to turn this on (we have a very similar use case) so gets a thumbs up from me (y)

jeremystretch:
Blocked by #9414

arthanson:
@jeremystretch I'm thinking we wouldn't want this for all custom fields, what about an "is_clonable" flag added to custom fields (default False) if set it will copy over when cloning the related object?

hagbarddenstore:
Arthur, that would suit our use cases perfectly. Some custom fields are used to determine which SLA we’re using, which could be cloned. Other fields are used to store external identifiers, which shouldn’t be cloned as they are unique per object in NetBox.On 4 Nov 2022, at 00:53, Arthur Hanson ***@***.***> wrote:﻿
@jeremystretch I'm thinking we wouldn't want this for all custom fields, what about an "is_clonable" flag added to custom fields (default False) if set it will copy over when cloning the related object?

PR #10959, opened 2022-11-17
8749 clone custom fields

arthanson:
### Fixes: #8749 

Allows to flag custom fields to be cloneable, these will be copied over when objects linked to the custom fields are cloned.


netbox/netbox/models/features.py
```diff
@@ -118,6 +118,15 @@ def clone(self):
         if is_taggable(self):
             attrs['tags'] = [tag.pk for tag in self.tags.all()]
 
+        # check custom fields
+        if hasattr(self, 'custom_field_data'):
+            from extras.models import CustomField
+
+            for field in CustomField.objects.get_for_model(self):
```
jeremystretch, line 125: We should be able to use `get_custom_fields()` here.
```suggestion
            for field in model.get_custom_fields():
```

jeremystretch (changes requested):
Do you think it makes sense to add a filter for `is_cloneable`? I can see a case for wanting to filter custom fields on this (e.g. to check for any fields that aren't).

PR #11036, opened 2022-11-28
11014 fix rack elevation name sorting

### LINKED ISSUE #11014: Elevations View Rack Name Number Sort
### NetBox version

v3.3.8

### Python version

3.10

### Steps to Reproduce

Using rack names such as 1/1, 1/2, 1/3... 1/10, 1/11, 1/12: When viewing the rack elevations view, the racks are not sorted left to right in numerical order when single digit numbers are used, they show up as 1/1, 1/11, 1/12, etc. whereas in the rack view they sort correctly, showing 1/1, 1/2, 1/3, etc. see below:

image

image

### Expected Behavior

In Elevations view, racks are sorted left to right in numerical order, e.g., 1/1, 1/2, 1/3, 1/4, 1/5, 1/6, 1/7, 1/8, 1/9, 1/10, 1/11, etc. like they sort in the rack view. 

### Observed Behavior

In elevations view, racks are sorted left to right as follows, 1/1, 1/10, 1/11, 1/12... etc.

arthanson:
### Fixes: #11014 

Fixes rack elevation sorting order for name, use _name instead of name to sort by.

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -590,14 +590,14 @@ class RackElevationListView(generic.ObjectListView):
         total_count = racks.count()
 
         ORDERING_CHOICES = {
-            'name': 'Name (A-Z)',
-            '-name': 'Name (Z-A)',
+            '_name': 'Name (A-Z)',
+            '-_name': 'Name (Z-A)',
             'facility_id': 'Facility ID (A-Z)',
             '-facility_id': 'Facility ID (Z-A)',
         }
-        sort = request.GET.get('sort', "name")
+        sort = request.GET.get('sort', "_name")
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
We probably want to avoid exposing the `_name` field (with the leading underscore) to the user. It might be cleaner to keep `name` and `-name` in place, but swap them for `_name` and `-_name` respectively when calling `order_by()`.

arthanson:
> We probably want to avoid exposing the `_name` field (with the leading underscore) to the user. It might be cleaner to keep `name` and `-name` in place, but swap them for `_name` and `-_name` respectively when calling `order_by()`.

Done

PR #11088, opened 2022-12-02
Fixes: #10720 Add small delay to the menu onmouseover 

### LINKED ISSUE #10720: Add small delay to the menu onmouseover
### NetBox version

v3.3.4

### Feature type

### Proposed functionality

I would like to see a (possibly optional) delay for collapsed menu fly-out.

When using a collapsed menu, the mouseover delay is effectively 0, meaning that if you move your mouse just a little too far and click, you end up navigating to whatever your mouse happens to be over once the menu expands.  Adding a small delay would reduce the number of accidental navigation errors and delays.

I acknowledge that this is a minor annoyance, but I hope that it's a minor fix.

### Use case

Reduces delay and mis-clicks for those that use a collapsed menu.

### Database changes

None

### External dependencies

None

PieterL75:
Fixes: #10720 Add small delay to the menu onmouseover

### DIFF
--- a/docs/release-notes/version-3.4.md
+++ b/docs/release-notes/version-3.4.md
@@ -40,6 +40,7 @@
 * [#11497](https://github.com/netbox-community/netbox/issues/11497) - Enforce `run_script` permission when executing scripts via REST API
 * ~[#11516](https://github.com/netbox-community/netbox/issues/11516) - Prevent text highlight utility from interpreting match as regex~
 * [#11522](https://github.com/netbox-community/netbox/issues/11522) - Correct tag links under contact & tenant list views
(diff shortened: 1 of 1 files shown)


kkthxbye-code:
I'm still not convinced that this is a good idea, however I'll review it regardless.

I don't think your approach is the correct one. The animation of the sidenav is controlled by CSS and as such the delay should be there too (it's also just the most common way of handling animations). Adding `transition-delay: 200ms` in all the relevant spots should be equivalent.

The css is a little messy, so it could use a cleanup as well, but I guess that's outside of the purview of this issue.

PieterL75:
@kkthxbye-code 
git n00b problem.. 

But, you say that the JS delay is not a good way to go.
but issue #7495 was solved using JS.. 

I'll try to get it working with CSS, maybe the solution for #7495 also needs to be CSS'd ?

PR #11093, opened 2022-12-02
11087 fix bottom banner background

### LINKED ISSUE #11087: Inconsistency of the background color when the bottom banner is active
### NetBox version

v3.3.9

### Python version

3.8

### Steps to Reproduce

1. Enable Bottom Banner in Config Revisions or go to demo.netbox.dev
2. Click on any sidemenu
3. Compare before and after

I think the background color is inherited from the body styles instead of the container-fluid class. Which is true for the top banner but not for the bottom banner.

### Expected Behavior

expected-banner-bottom-dark
expected-banner-bottom-light

### Observed Behavior

observed-banner-bottom-dark
observed-banner-bottom-light

arthanson:
### Fixes: #11087 

Fixes are just in netbox.scss and layout.html to change background color on bottom banner for config-context, other files are from running the yarn bundle.  Had to change the mx-3 to px-3 (margin to padding) to make the background not have white gaps on the margins.


jeremystretch:
I realized what the root problem was. The bottom banner needs to be included within the "content" container, rather than appearing as a separate element below it (fixed in d2d60c060). This ensures that it receives the proper background color regardless of the page's styling. Thanks for tackling this though!

PR #11096, opened 2022-12-02
Closes #10945: Enable recurring execution of scheduled reports & scripts

### LINKED ISSUE #10945: Periodic scheduling of Reports/Scripts
### NetBox version

v3.4-beta1

### Feature type

### Proposed functionality

It would be nice to not only be able to schedule scripts/reports once but to set a period the script/report should run, like every hour, every day, etc.

### Use case

This is very useful if you have scripts that e.g. sync with other systems or reports with long execution times. This way you don't need to wait for the report result, because you have the data from e.g. an hour ago. 

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #10945

- Add an `interval` field to the JobResult model
- Extend the UI & API views to record recurrence interval (in minutes) when scheduling jobs
- Extend workers to automatically reschedule jobs with an interval defined

### DIFF
--- a/netbox/extras/forms/reports.py
+++ b/netbox/extras/forms/reports.py
@@ -15,3 +15,8 @@ class ReportForm(BootstrapMixin, forms.Form):
         label=_("Schedule at"),
         help_text=_("Schedule execution of report to a set time"),
     )
+    interval = forms.IntegerField(
+        required=False,
+        label=_("Recurs every"),
+        help_text=_("Interval at which this report is re-run (in minutes)")
+    )
--- a/netbox/extras/forms/reports.py
+++ b/netbox/extras/forms/reports.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.utils import timezone
 from django.utils.translation import gettext as _
 
 from utilities.forms import BootstrapMixin, DateTimePicker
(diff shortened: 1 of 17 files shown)


jsenecal:
I may have missed it but where/how is a scheduled/recurring task is cancelled/stopped ?

netbox/extras/forms/reports.py
```diff
@@ -15,3 +15,8 @@ class ReportForm(BootstrapMixin, forms.Form):
         label=_("Schedule at"),
         help_text=_("Schedule execution of report to a set time"),
     )
+    interval = forms.IntegerField(
```
kkthxbye-code, line 18: Setting the interval in minutes is maybe not the most userfriendly way. Supporting something like m/h/d suffixes might be an option, shouldn't be too involved.

kkthxbye-code (commented):
Messed around with it a little and it works great. Some small stuff:

* If the user schedules a script in the past with an interval, the script will execute repeatedly. If the script has a short runtime it can be hard cancelling it, because you have to delete the newest non-executed scheduled JobResult. A solution could be to either disallow executing scripts in the past or setting the scheduled date to the current timestamp if it is in the past.
* Interval can be set to 0 which just doesn't work. It should probably just be caught in validation as invalid. Allowing a 0 interval (scheduling the job instantly) would cause the same issues as the above point.

kkthxbye-code:
@jsenecal - By deleting the JobResult.

PR #11106, opened 2022-12-05
Fixes #11022 - Allow redefining internally used background queues 

### LINKED ISSUE #11022: Allow redefining internally used background queues
### NetBox version

v3.3.8

### Feature type

### Proposed functionality

Allow changing the used RQ queue for webhook scheduling, custom scripts and reports to a specific queue.

### Use case

Currently script/report and webhook execution uses the same queue ("default"). One RQ worker will only ever run [one job at a time](https://python-rq.org/docs/workers/):

> Each worker will process a single job at a time. Within a worker, there is no concurrent processing going on. If you want to perform jobs concurrently, simply start more workers.

For webhooks we want to only ever run one job at the time, as the order is very important to avoid a webhook referencing a related object that hasn't been sent in a webhook yet.

For scripts/reports the user might not care about the order of executions, but as it stands now if the user runs one rqworker, long running scripts will block both webhooks and other scripts from execution before it is done. 

If the user starts multiple rqworkers that will allow multiple scripts to run at a time, but it will also allow multiple webhooks to be sent at a time, which in turn can create situations where a newer webhook is done processing before an older one - imagine three workers all picking up a job at the same time, the webhook will potentially send the HTTP request at roughly the same time, which one goes first depends on CPU scheduling or other external factors.

Allowing webhooks to be configured to use a seperate queue will allow the user to start one worker for webhooks and multiple workers for script/report execution, without risking data consistency issues with webhooks.

### Database changes

None

### External dependencies

None

kkthxbye-code:
### Fixes: #11022

Adds the optional setting `QUEUE_MAPPINGS`.

Usage:

```
QUEUE_MAPPINGS = {
    'webhook': 'low',
    'report': 'high',
    'script': 'custom-script-queue',
}
```

I also added a change so that RQ_QUEUES gets updated if the queue defined in QUEUE_MAPPINGS is not defined already.

### DIFF
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -141,6 +141,22 @@ When determining the primary IP address for a device, IPv6 is preferred over IPv
 
 ---
 
+## QUEUE_MAPPINGS
+
+Allows changing which queues are used internally for background tasks.
+
+```python
+QUEUE_MAPPINGS = {
+    'webhook': 'low',
+    'report': 'high',
+    'script': 'high',
+}
+```
+
+If no queue is defined the queue named `default` will be used.
+
+---
+
 ## RELEASE_CHECK_URL
 
 Default: None (disabled)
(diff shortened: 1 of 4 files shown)


jeremystretch (approved):
Great work, thanks!

PR #11143, opened 2022-12-09
11077 use formatting for custom field date

### LINKED ISSUE #11077: Date format for custom field does not respect values in configuration.py
### NetBox version

v.3.3.8

### Python version

3.8

### Steps to Reproduce

I'm not sure if this is a bug or my own ignorance on where this field gets its formatting.

I have created a custom field with type 'date'.
In the netbox configuration.py I have adjusted the date/time formatting for short date format to m-d-Y.
When populating a date in the custom field, the short date format is still Y-m-d.

>
# Date/time formatting.
DATE_FORMAT = 'N j, Y' # Dec 31, 2022
SHORT_DATE_FORMAT = 'm-d-Y' # 12-31-2022
TIME_FORMAT = 'g:i a' # 4:06 p.m.
SHORT_TIME_FORMAT = 'g:i:s a' # 4:06:32 p.m.
DATETIME_FORMAT = 'N j, Y g:i a' # Dec 31, 2022 4:06 p.m.
SHORT_DATETIME_FORMAT = 'm-d-Y g:i a' # 12-31-2022 4:06 p.m.

# Original date/time format
#DATE_FORMAT = 'N j, Y' # Dec 31, 2022
#SHORT_DATE_FORMAT = 'Y-m-d' # 2022-12-31
#TIME_FORMAT = 'g:i a' # 4:06 p.m.
#SHORT_TIME_FORMAT = 'H:i:s' # 16:06:02
#DATETIME_FORMAT = 'N j, Y g:i a' # Dec 31, 2022 4:06 p.m.
#SHORT_DATETIME_FORMAT = 'Y-m-d H:i' # 2022-12-31 16:06

### Expected Behavior

Expected behavior after changing
`SHORT_DATE_FORMAT = 'Y-m-d' # 2022-12-31`
to
`SHORT_DATE_FORMAT = 'm-d-Y' # 12-31-2022`

would be that the 'date' in the custom field would be displayed as m-d-Y.

### Observed Behavior

Date displayed in the custom field is still Y-m-d.

arthanson:
### Fixes: #11077 

Updates custom field display in tables to use SHORT_DATE_FORMAT specifier, also updates the override for DateColumn to also use SHORT_DATE_FORMAT. Is assuming Custom field date will always be in the Y-m-d format, would need an extra date-format specifier if in another format.

### DIFF
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -51,10 +51,12 @@ class DateColumn(tables.DateColumn):
     tables and null when exporting data. It is registered in the tables library to use this class instead of the
     default, making this behavior consistent in all fields of type DateField.
     """
-    def value(self, value):
+    def render(self, value):
         if value:
             return date_format(value, format="SHORT_DATE_FORMAT")
-        return None
+
+    def value(self, value):
+        return value
 
     @classmethod
     def from_field(cls, field, **kwargs):
@@ -444,7 +446,6 @@ class CustomFieldColumn(tables.Column):
         return escape(item)
 
     def render(self, value):
-        print("render CustomFieldColumn")
         if self.customfield.type == CustomFieldTypeChoices.TYPE_BOOLEAN and value is True:
             return mark_safe('<i class="mdi mdi-check-bold text-success"></i>')
         if self.customfield.type == CustomFieldTypeChoices.TYPE_BOOLEAN and value is False:
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -7,6 +7,7 @@ from django.contrib.auth.models import AnonymousUser
 from django.db.models import DateField, DateTimeField
 from django.template import Context, Template
 from django.urls import reverse
+from django.utils.dateparse import parse_date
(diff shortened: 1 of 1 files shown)


netbox/netbox/tables/columns.py
```diff
@@ -51,7 +52,9 @@ class DateColumn(tables.DateColumn):
     default, making this behavior consistent in all fields of type DateField.
     """
     def value(self, value):
-        return value
+        if value:
+            return date_format(value, format="SHORT_DATE_FORMAT")
+        return None
```
jeremystretch, line 57: For `value()`, we want to return the raw ISO-standard date regardless of how the UI display is configured. Instead, this should be applied to the column's `render()` method.

netbox/netbox/tables/columns.py
```diff
@@ -441,6 +444,7 @@ def _linkify_item(item):
         return escape(item)
 
     def render(self, value):
+        print("render CustomFieldColumn")
```
jeremystretch, line 447:
```suggestion
```

jeremystretch (changes requested):

PR #11155, opened 2022-12-11
Fixes #10447 - Allow re-assigning inventoryitems to other devices

### LINKED ISSUE #10447: Possibility to change device when editing inventory item.
### NetBox version

v3.3.4

### Feature type

### Proposed functionality

Return ability to change device when editing inventory item.

### Use case

We are using inventory items to account transceivers. When transceiver is moved to another device it should be reflected in netbox. [Here](https://github.com/netbox-community/netbox/issues/10437) it is said that this "feature" was removed for some reason, but if we have to recreate items we will lose history and the hole process of moving items becomes very burdensome.

inventory_item

### Database changes

_No response_

### External dependencies

_No response_

kkthxbye-code:
### Fixes: #10447

This is a suggested fix for allowing InventoryItems to be re-assigned to other devices. We had a discussion at one of the latest maintainers meetings about this, and my takeaway was that it would be fine to allow InventoryItems to move device. The alternative solution is to disallow it entirely, which requires disabling the device field for both the API and bulkedit.

There are two limitations to avoid inconsistant state:

* Validation prevents re-assigning the device if the InventoryItem has a parent or has children.
* When moving an InventoryItem the assigned component is removed.

An alternative to the second item would be to just deny the move, however it is currently not possible to de-couple an inventoryitem from a component (should also be fixed at some point), it seems like that would be too limiting.

### DIFF
--- a/netbox/dcim/forms/models.py
+++ b/netbox/dcim/forms/models.py
@@ -1610,6 +1610,13 @@ class InventoryItemForm(DeviceComponentForm):
         ('Hardware', ('manufacturer', 'part_id', 'serial', 'asset_tag')),
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Specifically allow editing the device of IntentoryItems
+        if self.instance.pk:
+            self.fields['device'].disabled = False
+
     class Meta:
         model = InventoryItem
         fields = [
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks for taking this on @kkthxbye-code! The double-override to re-enable the `device` field is obviously less than ideal, but I couldn't come up with a reasonable alternative. (IIRC that primary motivation for preventing reassignment of other components to different devices involves implications of the cabling model; maybe we can revisit those in the future.) This works for me!

PR #11231, opened 2022-12-19
Fixes #11210 - Exception is thrown when trying to CSV import a cable in with occupied terminations

### LINKED ISSUE #11210: Cannot import or clone a Cable
### NetBox version

v3.4.0

### Python version

3.9

### Steps to Reproduce

1. Import Cables
2. Add data to import a cable
3. Click Import

### Expected Behavior

A cable should be created.

### Observed Behavior

A ValueError was raised.
Creating a cable through the UI is done successfully.
Using the API creates the same error as importing.

```
<class 'ValueError'>

'Cable' instance needs to have a primary key value before this relationship can be used.

Python version: 3.9.2
NetBox version: 3.4.0
```

There seems to also be an issue with Cloning a Cable:
When you click Clone on a Cable you recieve this error:

```
<class 'AttributeError'>

Python version: 3.9.2
NetBox version: 3.4.0
```

kkthxbye-code:
### Fixes: #11210

I believe this is caused by the following django change:

https://docs.djangoproject.com/en/4.1/releases/4.1/#reverse-foreign-key-changes-for-unsaved-model-instances

This is the validation logic that is called for new cables that are bulk imported:
https://github.com/netbox-community/netbox/blob/3675ad2539400038af6797140ddca3679b8bca30/netbox/dcim/models/cables.py#L148-L149

In the case of occupied terminations, when the following code is reached the self.terminations queryset is called without the cable being saved yet:

https://github.com/netbox-community/netbox/blob/3675ad2539400038af6797140ddca3679b8bca30/netbox/dcim/models/cables.py#L111-L118

I'm not sure this is the correct fix. After the fix the following validation errors are provided:

image

Tests passes, but please give this a good look before merging.

Bonus question: Why does the validation failing in CableImportForm not "cancel" the request? Why are the clean methods of the cable still called afterwards (or before)?

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -112,6 +112,10 @@ class Cable(PrimaryModel):
     def a_terminations(self):
         if hasattr(self, '_a_terminations'):
             return self._a_terminations
+
+        if not self.pk:
+            return []
+
         # Query self.terminations.all() to leverage cached results
         return [
             ct.termination for ct in self.terminations.all() if ct.cable_end == CableEndChoices.SIDE_A
@@ -126,6 +130,10 @@ class Cable(PrimaryModel):
     def b_terminations(self):
         if hasattr(self, '_b_terminations'):
             return self._b_terminations
+
+        if not self.pk:
+            return []
+
         # Query self.terminations.all() to leverage cached results
         return [
             ct.termination for ct in self.terminations.all() if ct.cable_end == CableEndChoices.SIDE_B


jeremystretch:
This looks like the best solution for the Django 4.1 change. Thanks for tracking it down!

> Why does the validation failing in CableImportForm not "cancel" the request? Why are the clean methods of the cable still called afterwards (or before)?

IIRC the workflow will continue processing forms and collecting errors to report.

PR #11256, opened 2022-12-20
Fixes #11156 - Allow InventoryItem component reassignment

### LINKED ISSUE #11156: Allow modifying the assigned component for InventoryItems in the UI
### NetBox version

v3.3.9

### Feature type

### Proposed functionality

Currently it is possible to change the assigned component of an InventoryItem via. the API. It is however not possible to change the associated InventoryItem via. the UI. This means that the only way to create the association is by going to the component details view and clicking "Add Inventory Item". It is not possible to change the relationship afterwards and it is not possible to remove the relationship either.

A component field should be added to InventoryItem form to allow changing the associated component. The challenge is that the component field is a generic foreign key to any of the component types, so we need to figure out a slick way to present this to the user. We do it the following way for IP assignment currently:

image

However as there are currently 7 types of device components that support associating an InventoryItem, this might be a little much in the UI.

### Use case

Changing the associated component might be useful adding the inventoryitem after the component has been added to the device. There might also be times where a component should be removed but the inventory item should stay. Similarly it might be useful to change the associated component, if an SFP is moved to another port for example.

Lastly it will allow the user to do in the UI what is already possible in the API.

### Database changes

None

### External dependencies

None

kkthxbye-code:
### Fixes: #11156

This is a draft for adding component selection to the InventoryItem form. The approach here is the same as interface assignment for the IPAddress form. While I still think we should probably find a better way of doing generic relationship assignment, I believe that the feature is important enough to forgo that for now. We can easily refactor this if we find a better way in the future.

image

@jeremystretch - let me know what you think. If you agree on the approach I'll ready the PR for review.

### DIFF
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -1550,60 +1550,62 @@ class InventoryItemForm(DeviceComponentForm):
         required=False
     )
 
+    # Assigned component selectors
     consoleport = DynamicModelChoiceField(
         queryset=ConsolePort.objects.all(),
         required=False,
         query_params={
             'device_id': '$device'
-        }
+        },
+        label=_('Console port')
     )
-
     consoleserverport = DynamicModelChoiceField(
         queryset=ConsoleServerPort.objects.all(),
         required=False,
         query_params={
             'device_id': '$device'
-        }
+        },
+        label=_('Console server port')
     )
-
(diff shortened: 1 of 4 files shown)


jeremystretch:
Sorry for the delay in getting to this. IMO this approach is fine in lieu of a more elegant frontend mechanism. Thanks for tackling this!

PR #11278, opened 2022-12-21
pin pytz version to 2022.7 in requirements.txt

### LINKED ISSUE #11277: tzdata error 'No time zone found with key America/Ciudad_Juarez'
### NetBox version

v3.3.8

### Python version

3.9

### Steps to Reproduce

1. pip install -r requirements.txt
2. launch manage.py migrate with a plugin that has date related fields
3. error happens:
``zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key America/Ciudad_Juarez'``

seems related to many previous raised bugs, usually fixed by pinning tzdata in requirements.txt
due to the new release of tzdata/pytz 2022.7 

### Expected Behavior

application shall not crash

### Observed Behavior

application crash

nicoovh:
### Fixes: #11277

When pytz/tzdata release a new version together, if both are not pinned together, we end up with a mixed, non compatible versions. It happens this month with netbox version 3.3.8, because a fresh install would get ``tzdata==2022.6``, as pinned in requirements.txt, and ``pytz==2022.7``.

Which may lead to following error:

``zoneinfo._common.ZoneInfoNotFoundError: 'No time zone found with key America/Ciudad_Juarez'``

I suggest to pin both versions in the requirements.txt file.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,6 +31,7 @@ social-auth-core[openidconnect]==4.3.0
 svgwrite==1.4.3
 tablib==3.3.0
 tzdata==2022.7
+pytz==2022.7
 
 # Workaround for #7401
 jsonschema==3.2.0


jeremystretch:

PR #11285, opened 2022-12-21
Fixes #11280 - Fix exporting interfaces and FHRP group rows with multiple IP's assigned

### LINKED ISSUE #11280: VM Interface CSV Export have empty lines if two ip adresses assigned to VM Interface
### NetBox version

v3.4.1

### Python version

3.10

### Steps to Reproduce

VM Interface Export
open export.csv

### Expected Behavior

no empty lines should displayed
second lines should have the same information about device / interface as first line

### Observed Behavior

Between first interface Export with IP one and second Export IP three empty lines and second line only second IP Adress is displayed. no device / interface information in this line

kkthxbye-code:
### Fixes: #11280

Adds `value_ip_address` method to BaseInterfaceTable and FHRPGroupTable.

@jeremystretch - Let me know if we should create a new column extending TemplateColumn for IPAddresses instead. I felt like it might be overkill, but there is slight replication of code in the current fix.

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -506,6 +506,9 @@ class BaseInterfaceTable(NetBoxTable):
         verbose_name='Tagged VLANs'
     )
 
+    def value_ip_addresses(self, value):
+        return ",".join([str(obj.address) for obj in value.all()])
+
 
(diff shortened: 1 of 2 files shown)


jeremystretch:
> Let me know if we should create a new column extending TemplateColumn for IPAddresses instead.

Might be something to tackle in the near future, and not just for IP addresses, but IMO your solution is fine for now.

PR #11321, opened 2022-12-27
Fixes #11244: Elevations: Filter badge missing

### LINKED ISSUE #11244: Elevations: Filter badge missing
### NetBox version

v3.4.1

### Python version

3.9

### Steps to Reproduce

1. Go to menu => Elevations
2. Click tab "Filters"
3. Type any search term and press Enter

### Expected Behavior

I would expect a filter badge and a save button to be displayed as in other views. If you have a look at the "Filters" tab you can see that there is a dropdown for "Saved Filter" which doesn't make sense if you are not able to save a filter.

### Observed Behavior

Neither filter badge nor save button is shown.

dreng:

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -652,6 +652,7 @@ class RackElevationListView(generic.ObjectListView):
 
         racks = filtersets.RackFilterSet(request.GET, self.queryset).qs
         total_count = racks.count()
+        self.model = self.queryset.model
 
         # Ordering
         ORDERING_CHOICES = {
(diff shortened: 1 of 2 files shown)


netbox/dcim/views.py
```diff
@@ -691,6 +692,7 @@ def get(self, request):
             'sort_choices': ORDERING_CHOICES,
             'rack_face': rack_face,
             'filter_form': forms.RackElevationFilterForm(request.GET),
+            'model': self.model,
```
jeremystretch, line 695:
```suggestion
            'model': self.queryset.model,
```

netbox/dcim/views.py
```diff
@@ -652,6 +652,7 @@ def get(self, request):
 
         racks = filtersets.RackFilterSet(request.GET, self.queryset).qs
         total_count = racks.count()
+        self.model = self.queryset.model
```
jeremystretch, line 655: We can avoid setting the model on the instance and just pass it directly as context for the template.
```suggestion
```

jeremystretch (changes requested):

PR #11323, opened 2022-12-27
11271 flag to disable localization

### LINKED ISSUE #11271: Issues relating to the enabling of localization and LocaleMiddleware
### NetBox version

v3.4.1

### Python version

3.11

### Steps to Reproduce

This issue is intended to track all the issues related to the enabling of USE_L10N and LocaleMiddleware.

### Some words are now translated, causing a mix of English and the browser language

https://github.com/netbox-community/netbox/issues/11189#issuecomment-1356326763 - Humanized dates are output with localized `minutes/hours/days`
https://github.com/netbox-community/netbox/issues/11240 - The word "Tags" is displayed as the German word "Schlagwörter"

Possible solution is to disable USE_I18N as netbox is not yet translated itself.

### Floats are output with localized decimal separator breaking progress bars and other functionality

https://github.com/netbox-community/netbox/issues/11272 - Device position dropdown gets the U value with a comma, causing device save to fail.

https://github.com/netbox-community/netbox/discussions/11270 - When browser language is set to any language using comma as a decimal separator it breaks the progressbars:

https://github.com/netbox-community/netbox/issues/11273
https://github.com/netbox-community/netbox/blob/3675ad2539400038af6797140ddca3679b8bca30/netbox/utilities/templates/helpers/utilization_graph.html#L6

HTML doesn't support commas there for displaying progressbars. Possible solution is to disable the LocaleMiddleware again until we have a better understanding of the effects. Alternatively find all the places where comma as a decimal separator is invalid and find a way to output the unlocalized float.

### It is not possible to configure the date and datetime format anymore
https://github.com/netbox-community/netbox/issues/11189#issuecomment-1356326763
https://github.com/netbox-community/netbox/issues/11268

The way to configuring a global date and datetime format with LocaleMiddleware installed seems to be the following:
https://docs.djangoproject.com/en/4.1/ref/settings/#format-module-path

Currently we do not expose this and frankly I think it's a bad UX to have to create formats.py for each language you want to control localization for. It is also not great for people running containerized as it forces either mounts or building a custom image. Maybe there's a better way I'm missing. My suggestion is to revert the `USE_L10N` and `LocaleMiddleware` changes until we have a better solution for all the above problems. Django 5 is not slated to come out until december 2023, so we have plenty of time. Currently I do not see any benefits to the changes.

An alternative to reverting would be allowing the disabling of the `LocaleMiddleware` by a netbox setting.

arthanson:
### Fixes: #11271 

Add configuration.py flag to ENABLE_LOCALIZATION (default False) so this will disable localization by default.  There are a couple issues with the localization that we will need to address in the future to enable this:

1. provide localization language files if we want to turn on USE_I18N
2. Fix a couple UI issues with localized commas in numbers breaking progress bars
3. Allow an override to DATE_FORMAT and such as USE_L10N will take the browser setting as precedence over DATE_FORMAT.  It sounds like users want to be able to control this directly so will probably need to be another setting (CUSTOM_DATE_FORMAT)

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -69,7 +69,7 @@ Email is sent from NetBox only for critical events or if configured for [logging
 
 Default: False
 
-Determines if localization features are enabled or not.  This should only be enabled for development or testing purposes as the product is not fully localized and there can be issues in certain locales with turning this on (mostly in locales where commas are used as numeric separators instead of periods). Turning this on will localize numeric and date formats (overriding what is set for DATE_FORMAT) based on the browser settings as well as translate some strings within the product.
+Determines if localization features are enabled or not. This should only be enabled for development or testing purposes as netbox is not yet fully localized. Turning this on will localize numeric and date formats (overriding what is set for DATE_FORMAT) based on the browser locale as well as translate certain strings from third party modules.
 
 ---
 
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -65,6 +65,14 @@ Email is sent from NetBox only for critical events or if configured for [logging
 
 ---
 
+## ENABLE_LOCALIZATION
+
+Default: False
+
+Determines if localization features are enabled or not.  This should only be enabled for development or testing purposes as the product is not fully localized and there can be issues in certain locales with turning this on (mostly in locales where commas are used as numeric separators instead of periods). Turning this on will localize numeric and date formats (overriding what is set for DATE_FORMAT) based on the browser settings as well as translate some strings within the product.
+
+---
+
 ## HTTP_PROXIES
 
 Default: None
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -137,6 +137,7 @@ STORAGE_BACKEND = getattr(configuration, 'STORAGE_BACKEND', None)
 STORAGE_CONFIG = getattr(configuration, 'STORAGE_CONFIG', {})
 TIME_FORMAT = getattr(configuration, 'TIME_FORMAT', 'g:i a')
 TIME_ZONE = getattr(configuration, 'TIME_ZONE', 'UTC')
+ENABLE_LOCALIZATION = getattr(configuration, 'ENABLE_LOCALIZATION', False)
 
 # Check for hard-coded dynamic config parameters
 for param in PARAMS:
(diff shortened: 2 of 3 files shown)


netbox/netbox/settings.py
```diff
@@ -340,7 +341,14 @@ def _setting(name, default=None):
     'django_prometheus.middleware.PrometheusBeforeMiddleware',
     'corsheaders.middleware.CorsMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
-    'django.middleware.locale.LocaleMiddleware',
+]
+
+if ENABLE_LOCALIZATION:
```
kkthxbye-code, line 346: Minor thing. Maybe instead of conditionally inserting it, we could just remove it if ENABLE_LOCALIZATION is False:

```python
if not ENABLE_LOCALIZATION:
    MIDDLEWARE.remove("django.middleware.locale.LocaleMiddleware")
```

Seems a little cleaner then splitting the list.

arthanson: Good call, changed.

docs/configuration/system.md
```diff
@@ -65,6 +65,14 @@ Email is sent from NetBox only for critical events or if configured for [logging
 
 ---
 
+## ENABLE_LOCALIZATION
+
+Default: False
+
+Determines if localization features are enabled or not.  This should only be enabled for development or testing purposes as the product is not fully localized and there can be issues in certain locales with turning this on (mostly in locales where commas are used as numeric separators instead of periods). Turning this on will localize numeric and date formats (overriding what is set for DATE_FORMAT) based on the browser settings as well as translate some strings within the product.
```
kkthxbye-code, line 72:
```suggestion
Determines if localization features are enabled or not. This should only be enabled for development or testing purposes as netbox is not yet fully localized. Turning this on will localize numeric and date formats (overriding what is set for DATE_FORMAT) based on the browser locale as well as translate certain strings from third party modules.
```

I shortened this a bit. Also, I think we should refer to netbox as `netbox` instead of `product`.

kkthxbye-code (approved):
Apart from the one comment about the documentation it looks good to me. Tested it out and functions like advertised.

There's an argument to be made that the setting is misnamed as it disables both internationalization (i18n) and localization (L10n), but I personally think it's fine, as it's a temporary setting intended for development mostly.

jeremystretch:
> There's an argument to be made that the setting is misnamed as it disables both internationalization (i18n) and localization (L10n), but I personally think it's fine, as it's a temporary setting intended for development mostly.

Agreed.

Issue #11248, opened 2022-12-19
Initial NetBox 3.4.x migration fails when a plugin using SearchIndex is installed


peteeckel:
### NetBox version

3.4.1

### Python version

3.8

### Steps to Reproduce

1. Install NetBox 3.4.1
2. Install any plugin that uses SearchIndex functionality (e.g. netbox-dns) and configure it in PLUGINS
3. Run the initial migration

### Expected Behavior

The migration succeeds

### Observed Behavior

The migration fails with ProgrammingError exception:
```
Operations to perform:
  Apply all migrations: admin, auth, circuits, contenttypes, dcim, django_rq, extras, ipam, netbox_dns, sessions, social_django, taggit, tenancy, users, virtualization, wireless
Running migrations:
  Applying extras.0083_search...Reindexing 67 models.
Clearing cached values... 0 entries deleted.
Indexing models
  netbox_dns.nameserver... Traceback (most recent call last):
  File "/opt/netbox/lib/python3.8/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UndefinedTable: relation "netbox_dns_nameserver" does not exist
LINE 1: ...name", "netbox_dns_nameserver"."description" FROM "netbox_dn...
                                                             ^


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/netbox/netbox/manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/netbox/lib/python3.8/site-packages/django/core/management/__init__.py", line 446, in execute_from_command_line
    utility.execute()
  File "/opt/netbox/lib/python3.8/site-packages/django/core/management/__init__.py", line 440, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/netbox/lib/python3.8/site-packages/django/core/management/base.py", line 402, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/netbox/lib/python3.8/site-packages/django/core/management/base.py", line 448, in execute
    output = self.handle(*args, **options)
  File "/opt/netbox/lib/python3.8/site-packages/django/core/management/base.py", line 96, in wrapped
    res = handle_func(*args, **kwargs)
  File "/opt/netbox/lib/python3.8/site-packages/django/core/management/commands/migrate.py", line 349, in handle
    post_migrate_state = executor.migrate(
[...]

PR #11331, opened 2022-12-28
Fixes #11223 & #11248 - Reindex app label

### LINKED ISSUE #11223: Reindex command does not accept app_label as argument
### NetBox version

v3.4.1

### Python version

3.10

### Steps to Reproduce

1. Look at the help of the reindex management command: it accepts ```app_label[.ModelName]```
```
$ python /opt/netbox/netbox/manage.py reindex -h
[...]
positional arguments:
  app_label[.ModelName]
                        One or more apps or models to reindex
```
2. Reindex all models of a plugin or an app:
```
$ python /opt/netbox/netbox/manage.py reindex dcim
CommandError: Invalid model: dcim. Model names must be in the format <app_label>.<model_name>.
```

The use case is reindexing all the models of a plugin after adding SearchIndex for my models.

### Expected Behavior

The command reindexes all the the models of the plugin/app corresponding to app_label.

### Observed Behavior

```
CommandError: Invalid model: dcim. Model names must be in the format <app_label>.<model_name>.
```

The current code does not allow selecting all the models of an app. ```registry['search']``` could be iterated to find all the models of the app if there is no dot in model_names, or simpler, the args of the command could be updated to remove that possibility.

https://github.com/netbox-community/netbox/blob/3675ad2539400038af6797140ddca3679b8bca30/netbox/extras/management/commands/reindex.py#L29-L40

Alef-Burzmali:
### Fixes: #11223 and #11248 

Add the possibility to reindex all the models of an app_label via the management command reindex.

The second commit modifies the reindex call in migration extras.0083_search to reindex only the models of core NetBox, to prevent plugins from blocking the migration as in #11248. Plugins should add their own reindex migration when they add a SearchIndex, such as:

```python
import sys

from django.core import management
from django.db import migrations

def reindex(apps, schema_editor):
    # Build the search index (except during tests)
    if "test" not in sys.argv:
        management.call_command("reindex", "<plugin_label>")

class Migration(migrations.Migration):
    dependencies = [
        ("<plugin_label>", "<NNNN_previous_migration>"),
    ]

    operations = [
        migrations.RunPython(
            code=reindex, reverse_code=migrations.RunPython.noop
        ),
    ]
```

### DIFF
--- a/netbox/extras/management/commands/reindex.py
+++ b/netbox/extras/management/commands/reindex.py
@@ -27,17 +27,28 @@ class Command(BaseCommand):
         # Return only indexers for the specified models
         else:
             for label in model_names:
-                try:
-                    app_label, model_name = label.lower().split('.')
-                except ValueError:
+                labels = label.lower().split('.')
+
+                # Label specifies an exact model
+                if len(labels) == 2:
+                    app_label, model_name = labels
+                    try:
+                        idx = registry['search'][f'{app_label}.{model_name}']
+                        indexers[idx.model] = idx
+                    except KeyError:
+                        raise CommandError(f"No indexer registered for {label}")
+
+                # Label specifies all the models of an app
+                elif len(labels) == 1:
+                    app_label = labels[0] + '.'
+                    for indexer_label, idx in registry['search'].items():
+                        if indexer_label.startswith(app_label):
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks for taking this on @Alef-Burzmali!

PR #11337, opened 2022-12-29
Fixes #11259 - add username for redis authentication

### LINKED ISSUE #11259: Add Redis username parameter
### NetBox version

v.3.4.0

### Feature type

### Proposed functionality

Allow for specifying an username in the Redis configuration, to be able to connect to Redis servers, which have an redis-username configured.

### Implementation:
* Add optional username parameter to redis-tasks and redis-caching in configuration.py
* Add TASKS_REDIS_USERNAME and CACHING_REDIS_USERNAME to settings.py
* Add username to cache-options and RQ_PARAMS in setting.py

### Use case

When you have an Redis instance with username configured, you are unable to connect.  E.g. Managed-Redis at OVH-cloud forces you to set an redis-user, so this service can't be used for NetBox.

If you could specify an redis user, it would allow users to connect to more redis-instances.

### Database changes

_No response_

### External dependencies

_No response_

christianpinger:
### Fixes: #11259 

Add an optional configuration parameter `username` to redis (tasks and caching). This allows to connect to redis instances which require a username.


jeremystretch:
Thanks @christianharendt!

PR #11343, opened 2022-12-30
10201 cable terminations2

### LINKED ISSUE #10201: Remove cable terminations cause AssertionError
### NetBox version

v3.3.2-dev (also on v3.3.1 demo site)

### Python version

3.8

### Steps to Reproduce

1. Load netbox-demo-v3.3.json
2. Select devices > dmi01-akron-sw01 > interfaces > GigabitEthernet0/1/3 > connect cable > interface
3. Select B side interfaces
    - dmi01-akron-rtr01  GigabitEthernet0/1/5
    - dmi01-akron-rtr01  GigabitEthernet0/1/6
    - dmi01-akron-rtr01  GigabitEthernet0/1/7
    - dmi01-akron-sw01  GigabitEthernet1/0/14
    - dmi01-akron-sw01  GigabitEthernet1/0/15
4. Click Connect Button
5. Go back to edit the cable
6. Remove B Side interfaces
    - dmi01-akron-rtr01  GigabitEthernet0/1/6
    - dmi01-akron-rtr01  GigabitEthernet0/1/7
    - dmi01-akron-sw01  GigabitEthernet1/0/15
7. Click Connect Button

### Expected Behavior

Edit success and remove these three interfaces from cable B side

### Observed Behavior

Get Assertion Error

image
image

arthanson:
### Fixes: #10201

Fix cable termination checks if edit and remove terminations.  

@jeremystretch can you please double-check the logic here, not 100% sure if I'm missing any edge cases or anything.  **Note:** the lines removed in signals.py for CableTermination delete:

```
    for cablepath in CablePath.objects.filter(_nodes__contains=instance.cable):
        cablepath.retrace()
```
This assumes that CableTermination can only be removed on a Cable save (which I believe is correct), so the path retrace is handled in the update_connected_endpoints signal and this is not needed here.  If we can directly delete a CableTermination without going through Cable save then this would be an issue.


kkthxbye-code:
> This assumes that CableTermination can only be removed on a Cable save

What about the cable-termination API endpoint? Or is that logic seperate? Also custom scripts I guess.

jeremystretch:
Thanks @arthanson.

>  If we can directly delete a CableTermination without going through Cable save then this would be an issue.

I ended up implementing a narrower fix in 758c5347f, as I believe this workflow also needs to be supported. I've also included a test for the condition outlined in #10201.

PR #11346, opened 2022-12-30
Fixes #11345 - Fix module validation

### LINKED ISSUE #11345: Module Bulk Import: Server error page comes up if data is incorrect
### NetBox version

v3.4.1

### Python version

3.8

### Steps to Reproduce

1. Navigate to "Module Bulk Import" page (/dcim/modules/import/)
2. In the text box, add the following top line: "device,module_bay,module_type,status"
3. In the text box, add the following second line: "_non-device_,slot 2,Micro 2000,active"
4. Ensure the 'non-device' field does not match a device name that exists in NetBox
5. Click 'submit'
6. Observe how a 'Server Error' page comes up

net
net2

### Expected Behavior

If an issue exists with the inputted data, the page should direct the user to where the issue exists. In the provided example, the issue would be in row 1, column 1. This should be shown to the user to instead of failing. This would align it with other bulk import pages. This issue seems to exist with all fields that uses netbox stored data such as module_bay and module_type. 

Example from the device import page of expected behavior:

net3

### Observed Behavior

If inputted data is incorrect, a server error page pulls up instead of telling the user where the issue is. This leads to frustration on which line is causing the issue.

kkthxbye-code:
### Fixes: #11345

The issue was that validation of module component replication/adoption was executed even if field validation failed. The model also had a validation step that would throw an exception when no module_bay was set (which was the case when trying to import to a non-existant module bay).

I'm not sure if this is actually the preferred way to do this:

```
cleaned_data = super().clean()

# Skip form validation if field validation already found errors.
if self.errors:
    return cleaned_data
```

Please let me know if there is a smarted way to bail out if field validation (non-existant device for example) has already failed.

### DIFF
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -54,23 +54,20 @@ class InterfaceCommonForm(forms.Form):
 class ModuleCommonForm(forms.Form):
 
     def clean(self):
-        cleaned_data = super().clean()
+        super().clean()
 
-        # Skip form validation if field validation already found errors.
-        if self.errors:
-            return cleaned_data
-
-        replicate_components = cleaned_data.get("replicate_components")
-        adopt_components = cleaned_data.get("adopt_components")
-        device = cleaned_data.get('device')
-        module_type = cleaned_data.get('module_type')
-        module_bay = cleaned_data.get('module_bay')
+        replicate_components = self.cleaned_data.get('replicate_components')
+        adopt_components = self.cleaned_data.get('adopt_components')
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks for tackling this @kkthxbye-code. I just pushed a minor tweak to move the check for `self.errors`.

An aside: We don't usually return anything from a form's `clean()` method as it shouldn't be necessary to, though I noticed when reviewing this that the parent class' method _does_ return `self.cleaned_data`. That might be something we should clean up application-wide.

PR #11347, opened 2022-12-30
Fixes #11342 - Fix component traces all pointing to the interface trace URL

### LINKED ISSUE #11342: UI: Cable Trace button has incorrect URL for power-port & power-outlet models
### NetBox version

v3.4.1

### Python version

3.8

### Steps to Reproduce

1. Create a device with at least one power port
2. Create another device with at least one power outlet
3. Connect the power port to the power outlet with a cable
4. Click on either the power port or the power outlet to get the main "overview" page of the port or outlet (URL should be `/dcim/power-ports/{id}` or `/dcim/power-outlets/{id}`)
5. In the "Connection" table on the page, to the right of the "Cable" cell, the cable's ID # will be shown [correct], and the cable trace button to the right of the ID will have an incorrect URL, to `/dcim/interfaces` instead of `/dcim/power-ports` or `/dcim/power-outlets`

### Expected Behavior

Expect the cable trace button to trace the power cable, and not some other random interface (which may or may not actually be connected, and hence might throw an error).

### Observed Behavior

The Cable Trace button should be referencing `/dcim/(power-ports|power-outlets)/{id}/trace/` but instead it references `/dcim/interfaces/{id}/trace/`. I poked through the code a little bit but I couldn't figure out where/how it is building the URL (possibly due to my very limited Python skill).

kkthxbye-code:
### Fixes: #11342

Alternatively we could add a property to the PathEndpoint abstract model called `trace_url` or something of that effect. Not sure if it's worth it, but let me know and I'll change it.

### DIFF
--- a/netbox/templates/dcim/consoleport.html
+++ b/netbox/templates/dcim/consoleport.html
@@ -60,7 +60,7 @@
               {% if object.mark_connected %}
                 <span class="text-success"><i class="mdi mdi-check-bold"></i></span> Marked as connected
               {% elif object.cable %}
-                {% include 'dcim/inc/connection_endpoints.html' %}
+                {% include 'dcim/inc/connection_endpoints.html' with trace_url='dcim:consoleport_trace' %}
               {% else %}
                 <div class="text-muted">
                   Not Connected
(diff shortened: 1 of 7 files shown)


jeremystretch:
Thanks @kkthxbye-code!

PR #11377, opened 2023-01-03
#11371 - Add various 100Mb Interface Types

### LINKED ISSUE #11371: Interface Type Addition: Add 100Base-FX and other 100Mbps Types
### NetBox version

v3.4.1

### Feature type

Data model extension

### Proposed functionality

We have some legacy hardware that runs on 100Base-FX (Fast Ethernet over Fiber). We would like to request the interface type be added

### Use case

We have some Allied Telesis AT-FS202's in the field that we are looking to record. These have 100Base-FX connection for their Port1 and 10Base-T/100Base-TX for their Port 2

### Database changes

_No response_

### External dependencies

_No response_

danner26:
### Fixes: #11371 

Added 3 types of interfaces. 100base-fx (fast ethernet over fiber), 100base-lfx (proprietary fast ethernet over fiber), and 100base-t1 (10/100 single pair) for completeness.

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -921,6 +921,8 @@ class InterfaceTypeChoices(ChoiceSet):
         (
             'Ethernet (fixed)',
             (
+                (TYPE_100ME_FX, '100BASE-FX (10/100ME FIBER)'),
+                (TYPE_100ME_LFX, '100BASE-LFX (10/100ME FIBER)'),
                 (TYPE_100ME_FIXED, '100BASE-TX (10/100ME)'),
                 (TYPE_100ME_T1, '100BASE-T1 (10/100ME Single Pair)'),
                 (TYPE_1GE_FIXED, '1000BASE-T (1GE)'),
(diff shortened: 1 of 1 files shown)


netbox/dcim/choices.py
```diff
@@ -929,6 +933,8 @@ class InterfaceTypeChoices(ChoiceSet):
         (
             'Ethernet (modular)',
             (
+                (TYPE_100ME_FX, '100BASE-FX (10/100ME FIBER)'),
+                (TYPE_100ME_LFX, '100BASE-LFX (10/100ME FIBER)'),
```
jeremystretch, line 937: These should probably go under the "fixed" group, as the standard applies to the physical transmission rather than the interface to a transceiver.

danner26: That makes sense, I updated it

netbox/dcim/choices.py
```diff
@@ -929,7 +935,7 @@ class InterfaceTypeChoices(ChoiceSet):
         (
             'Ethernet (modular)',
             (
-                (TYPE_1GE_GBIC, 'GBIC (1GE)'),
+                (TYPE_1GE_GBIC, 'GBIC (1GE)')
```
jeremystretch, line 938:
```suggestion
                (TYPE_1GE_GBIC, 'GBIC (1GE)'),
```

PR #11382, opened 2023-01-04
#11150: Add a filter for device.primary_ip4 and primary_ip6

### LINKED ISSUE #11150: Add a filter for device.primary_ip4 and primary_ip6 to the REST API / graphQL
### NetBox version

v3.3.9

### Feature type

### Proposed functionality

I would like to be able to fetch a device based on it's primary ipv4 (and probably also ipv6). I understand IP addresses aren't unique (you can them in multiple VRF's), but it should still be posible to get the device based on the ID of the IP Address.

Related to https://github.com/netbox-community/netbox/issues/10058

Mentioned here https://groups.google.com/g/netbox-discuss/c/Di5VMTtK5lg

### Use case

I'm pushing devices to netbox trough a script with the API, and it happens that a device already exists with the same primary IP address. I want to handle these cases (probably by removing the old device).

When adding the device, I currently only get the error `Device with this Primary IPv4 already exists`, but there is no way to find this actual device. 

I assign the IP address to an interface of the new device before I can configure the primary IP, but somehow while doing this, the IP address is still linked to an old device, causing that device to have a primary IP4 that isn't linked to any of its interfaces.

Once this happens, it's not possible anymore to find the actual device to remove it.

### Database changes

_No response_

### External dependencies

_No response_

michaelarnauts:
### Fixes: #11150

* Add `primary_ip4` and `primary_ip6` to the DeviceFilterSet
* Add tests to query based on those fields

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -3,7 +3,7 @@ from django.contrib.auth.models import User
 from django.utils.translation import gettext as _
 
 from extras.filtersets import LocalConfigContextFilterSet
-from ipam.models import ASN, VRF
+from ipam.models import ASN, IPAddress, VRF
 from netbox.filtersets import (
     BaseFilterSet, ChangeLoggedModelFilterSet, OrganizationalModelFilterSet, NetBoxModelFilterSet,
 )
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks @michaelarnauts! I've extended your tests just a bit to test querying multiple IDs, but this is great!

PR #11411, opened 2023-01-05
NetBox should always be referred to as NetBox [DATALAD RUNCMD]

ypid:
### Fix to simple for issue. Ref to docs see below.

Fix all instances of "Netbox" except the one that is used as an example how not to write it.

Ref: https://netbox.readthedocs.io/en/stable/development/style-guide/#branding
Follow up for:  #5754

See the commit body for the command I used. You may put this into your CI tests or something ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ employed by thousands of organizations around the world.
 
 ## About NetBox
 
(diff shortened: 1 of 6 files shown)


jeremystretch:
Thanks for cleaning these up!

PR #11414, opened 2023-01-05
Closes #10486: Add buttons to edit cables

### LINKED ISSUE #10486: Add buttons to edit cables
### NetBox version

v3.3.4

### Feature type

### Proposed functionality

The device view provides tabs for interfaces, power ports etc. On every line we have a button to connect a cable or, if a cable is connected already, a button to disconnect the cable.

I've got a PR ready that adds a button to all kinds of connection overviews (console ports, power ports, front ports, interfaces etc.). It looks like this:

grafik

### Use case

What I miss here is a button to go directly to the edit page of the cable including a redirect to the page we came from. Best thing you can do at the moment is to click the link in the "Cable" column (if not hidden), edit the cable and go back to the connections overview manually. This approach is kinda cumbersome.

### Database changes

None

### External dependencies

None

dreng:


jeremystretch:
Great work, thanks @dreng!

PR #11445, opened 2023-01-09
Fixes #11444 - Snapshot DeviceBay before populating/depopulating.

### LINKED ISSUE #11444: Missing pre-change data for device bay
### NetBox version

v3.3.10

### Python version

3.8

### Steps to Reproduce

1. Add a device to a device bay
2. Delete the device from device bay
3. Check change log for device bay

### Expected Behavior

Change Log and webhook should show previous installed device under pre-change data

### Observed Behavior

When deleting a device from device bay the pre-change data is null.

DeviceBayPopulateView and DeviceBayDepopulateView doesn't call snapshot before disassociating the device.

kkthxbye-code:
### Fixes: #11444

Since we overwrite the post method of `ObjectEditView` the snapshot method is not called. This is just the obvious fix, not sure if there's a smarter way of ensuring that snapshot is called.

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -2820,7 +2820,7 @@ class DeviceBayPopulateView(generic.ObjectEditView):
         form = forms.PopulateDeviceBayForm(device_bay, request.POST)
 
         if form.is_valid():
-
+            device_bay.snapshot()
             device_bay.installed_device = form.cleaned_data['installed_device']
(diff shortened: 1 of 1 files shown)


jeremystretch:
> This is just the obvious fix, not sure if there's a smarter way of ensuring that snapshot is called.

LGTM, at least until we can do something to better standardize these views.

PR #11465, opened 2023-01-11
Unique ip prefix

Mister-IM:
Greetings, I need the ability to prevent the creation of duplicate prefixes or ip addresses.

Either at the time of creating a prefix or ip address, receive a warning that such a prefix or IP address already exists.


kkthxbye-code:
Please follow the contributer guidelines.

https://github.com/netbox-community/netbox/blob/develop/CONTRIBUTING.md

PR #11475, opened 2023-01-12
#9386 Buttons on forms and lists made sticky to top/bottom

### LINKED ISSUE #9386: Make form/pagination button stick at the bottom
### NetBox version

v3.2.3

### Feature type

### Proposed functionality

The pagination and edit button are currently hidden at the bottom of the page.
One needs to scroll down to reach the 'Save' or 'Page 5' like buttons.

It would be very handy if those buttons 'stick' to the bottom of the page, and the do not scroll.

image
image

### Use case

This way, when you only need to edit the 'name' of a device for example, you have that 'Save' button right at hand.
Or of you open a prefix list, you have the 'Page 5' button right at hand, rather having to scroll down to navigate

### Database changes

None

### External dependencies

None

PieterL75:


PieterL75:
First shot, please see if this according policy

jeremystretch:
Thanks @PieterL75; this is a great start! We probably need to add a container element of some type around the paginator & buttons to help with padding. Currently there's no space between the "hovering" controls and the table items:

Screenshot 2023-01-20 at 10-20-57 Virtual Machines NetBox

Do you want to take a shot at tweaking this? If not I can play around with it when I have time.

PieterL75:
@jeremystretch  Added some extra y-padding
image

PieterL75:
@jeremystretch Is this ok ? or does it need improvement

jeremystretch (changes requested):
Thanks for working on this @PieterL75, however I don't think this is quite where it needs to be. The floating element interferes with the shadow and border of the table. I'm not sure what the best approach is here, but I don't think we can merge this without some aesthetic improvements.

jeremystretch:
Closing this out as there doesn't seem to be any further movement on this. Happy to revisit in the future.

PR #11481, opened 2023-01-12
Fixed swagger schema for bulk operation

### LINKED ISSUE #10922: Bulk update data type is incorrect in swagger spec
### NetBox version

v3.3.7

### Python version

3.10

### Steps to Reproduce

As per the swagger doc, the bulk payload requires a dict
image

However, when trying to run the same, it throws an error

image

The issue is seen for all models in the spec for PUT and PATCH using the bulk operations

### Expected Behavior

The spec should display that the required payload should be a List of items/dictionaries

### Observed Behavior

The spec displays it needs a dictionary

abhi1693:
### Fixes: #10922

Using django `method_decorator` I have added the `BulkOperation` to present list of objects

### DIFF
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -1,5 +1,7 @@
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
+from django.utils.decorators import method_decorator
+from drf_yasg.utils import swagger_auto_schema
 from rest_framework import status
 from rest_framework.response import Response
 
@@ -30,6 +32,9 @@ class BulkUpdateModelMixin:
         }
     ]
     """
+    @method_decorator(decorator=swagger_auto_schema(
+        request_body=BulkOperationSerializer(many=True),
+    ))
     def bulk_update(self, request, *args, **kwargs):
         partial = kwargs.pop('partial', False)
         serializer = BulkOperationSerializer(data=request.data, many=True)
@@ -61,6 +66,9 @@ class BulkUpdateModelMixin:
 
             return data_list
 
+    @method_decorator(decorator=swagger_auto_schema(
+        request_body=BulkOperationSerializer(many=True),
+    ))
     def bulk_partial_update(self, request, *args, **kwargs):
         kwargs['partial'] = True
         return self.bulk_update(request, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


jeremystretch:
I'm not sure this is a complete solution. While the schema does now specify a list, the object definition includes only the `id` field. Shouldn't it list the optional attributes as well?

Screenshot 2023-01-17 at 21-17-21 NetBox API

abhi1693:
I'll see if I can figure out the optional parameter as well. I used the serializer that was in the method.

abhi1693:
@jeremystretch I'm not able to figure out how to get the optional attributes in the schema

jeremystretch:
I'm not sure there's any need for this change any longer, given that we've moved to OpenAPI 3.0 for NetBox v3.5. @arthanson can you weigh in?

arthanson:
yes, will no longer work for OpenAPI 3.

jeremystretch:
That said, is there any harm in applying this for v3.4? Or should we consider the 2.0 spec "frozen" at this point, since we've already begun the migration?

arthanson:
Can apply, but would then need to retroactively apply it to the pre OpenAPI 3 merge of 3.5 and produce another frozen yaml image for it to work going forward.

jeremystretch:
I'm going to try the following:

1. Generate a new schema file from this branch
2. Generate a diff between that and the current file
3. Apply that diff to the snapshot we're shipping with v3.5

If that works, I think it makes sense to merge this and ship that updated snapshot. Otherwise, it's probably not worth the effort to adjust.

jeremystretch:
It seems that the new schema, while properly indicating bulk support, loses the object-specific schema definition:

```
613c613,616
<                             "$ref": "#/definitions/WritableCircuitTermination"
---
>                             "type": "array",
>                             "items": {
>                                 "$ref": "#/definitions/BulkOperation"
>                             }
```

@abhi1693 is that intentional?

abhi1693:
Well, wasn't indending to do that but since the `BulkOperation` only specifies a list of ID (required), the schema generated loses the object specific definition. One way could be to somehow merge the object serializer with the `BulkOperationSerializer` to generate the schema, but I'm not sure how that would be done here.

jeremystretch:
It seems at this point that it's probably not worth pursuing for v3.4 given that v3.5 will be out by end of month, although we should check where this ended up for the new OpenAPI 3.0 spec. I'm going to close this out, but thank you for working on it!

PR #11499, opened 2023-01-13
added import loggers for plugins

### LINKED ISSUE #11267: ImportErrors inside plugins are currently ignored → add logging
### NetBox version

v3.4.0

### Feature type

### Proposed functionality

This has been requested before, see #4628. I'm currently writing my first plugin and just spent about a day trying to figure out why it won't get loaded. No matter how far I turned up the log level, this `pass` in `extras/plugins/urls.py` always caught the `ImportError` inside my plugin:

```python

    # Check if the plugin specifies any base URLs
    try:
        urlpatterns = import_string(f"{plugin_path}.urls.urlpatterns")
        plugin_patterns.append(
            path(f"{base_url}/", include((urlpatterns, app.label)))
        )
    except ImportError:
        pass
```

I propose replacing the `pass` with a log message that says something like "could not load plugin base URL {plugin_path}.urls.urlpatterns" and also log a traceback.

Since this mechanism is used to allow having plugins without API URLs (or base URLs), this is not neccessarily an error and should probably be log level `INFO` and the traceback should be `DEBUG`, I assume?

I have a patch ready, and can submit a PR, if this seems like a good idea. Log would look something like this:

```
gunicorn[1337]: Plugin base URL patterns could not be loaded: my_little_plugin.urls.urlpatterns

gunicorn[1337]: Traceback (most recent call last):
gunicorn[1337]: File "/opt/netbox/netbox/extras/plugins/urls.py", line 33, in <module>
gunicorn[1337]: urlpatterns = import_string(f"{plugin_path}.urls.urlpatterns")
gunicorn[1337]: File "/opt/netbox-3.4.0/venv/lib64/python3.8/site-packages/my_little_plugin/forms.py", line 2, in <module>
[...]
gunicorn[1337]: from dcim.models import Devic, Interface
gunicorn[1337]: ImportError: cannot import name 'Devic' from 'dcim.models' (/opt/netbox/netbox/dcim/models/__init__.py)
```

### Use case

I hope this saves plugin developers some head-scratching.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/extras/plugins/urls.py
+++ b/netbox/extras/plugins/urls.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.apps import apps
 from django.conf import settings
 from django.conf.urls import include
@@ -6,6 +8,8 @@ from django.urls import path
 from django.utils.module_loading import import_string
 
 from . import views
+
+logger = logging.getLogger('netbox.extras.plugins')
 
 # Initialize URL base, API, and admin URL patterns for plugins
 plugin_patterns = []
@@ -30,7 +34,7 @@ for plugin_path in settings.PLUGINS:
             path(f"{base_url}/", include((urlpatterns, app.label)))
         )
     except ImportError:
-        pass
+        logger.error(f"Plugin {plugin_name} does not define any base URLs")
 
     # Check if the plugin specifies any API URLs
     try:
@@ -39,4 +43,4 @@ for plugin_path in settings.PLUGINS:
             path(f"{base_url}/", include((urlpatterns, f"{app.label}-api")))
         )
     except ImportError:
-        pass
+        logger.error(f"Plugin {plugin_name} does not define any API URLs")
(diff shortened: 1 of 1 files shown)


kkthxbye-code:
I think the intention was to log the exception.

abhi1693:
@kkthxbye-code I made slight change to add the error as debug, otherwise you will always atleast see a warning.

kkthxbye-code:
The original FR states:

> I propose replacing the pass with a log message that says something like "could not load plugin base URL {plugin_path}.urls.urlpatterns" and *also log a traceback*.

You can add `exc_info=True` to the log calls. As the import error can be nested multiple levels, I think it makes sense to include it.

jeremystretch:
I think we just need to be a bit smarter about how we check for the presence of `urlpatterns` modules. Bear in mind that a plugin might not provide any URL patterns, which is perfectly valid and should not trigger any log messages (other than for debugging purposes).

Rather than employing `import_string()` (which assumes that the specified module should exist), we could first call `module_has_submodule()` to check for its existence. If found, then we can attempt to import it using `import_string()` without catching any `ImportErrors`.

PR #11500, opened 2023-01-13
added device and vm tab on device role

### LINKED ISSUE #11228: Device Roles: Add Virtual Maschines view
### NetBox version

v3.3.9

### Feature type

### Proposed functionality

Under Device Roles, only the Devices are displayed in the detailed view of the individual roles.

But not the virtual machines associated with them.

This does not look complete. Maybe this was forgotten.

vms

### Use case

I think the use case is obvious or?

So that you can see immediately in the respective role which VMs have this role 

### Database changes

?

### External dependencies

?

abhi1693:

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -1755,6 +1755,7 @@ class DeviceRoleDevicesView(generic.ObjectChildrenView):
     def get_children(self, request, parent):
         return Device.objects.restrict(request.user, 'view').filter(device_role=parent)
 
+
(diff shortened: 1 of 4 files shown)


jeremystretch:
IMO if we're going to have a tab for VMs, we should move the devices table to a tab as well, to keep things orderly. However, I also just recently opened #11517, which asserts that the device role view should not list _either_type, for feedback. If accepted, it would render this issue moot.

abhi1693:
I'll still go ahead and make the required changes in case it's needed to be merged.

jeremystretch:
Thanks @abhi1693!

PR #11501, opened 2023-01-13
added l2vpn_termination on vlan filterset

### LINKED ISSUE #11118: Allow VLAN filtering based on L2VPN termination
### NetBox version

v3.3.9

### Feature type

New functionality

### Proposed functionality

Allow (API) filtering VLAN's based on L2VPN termination. Currently i can filter VLAN's based on description, ID etc. But not with their corresponding VXLAN VNI identifier.

If this functionality is already present in a different way (get all L2VPN's with linked VLAN's) then that would also fulfill our use-case.

### Use case

This would allow us to get a list of all VLAN's which would be active in our VXLAN environment with the API call "L2VPN_termination__n='null'".

### Database changes

No DB change required

### External dependencies

No new external dependencies.

abhi1693:

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -852,10 +852,15 @@ class VLANFilterSet(NetBoxModelFilterSet, TenancyFilterSet):
         queryset=VirtualMachine.objects.all(),
         method='get_for_virtualmachine'
     )
+    l2vpn_termination = django_filters.ModelMultipleChoiceFilter(
+        field_name='l2vpn_terminations',
+        queryset=L2VPNTermination.objects.all(),
+        label=_('L2VPN termination (ID)'),
+    )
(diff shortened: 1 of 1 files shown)


jeremystretch:
From the use case cited in #11118, it seems the goal here is to enable filtering VLANs by associated L2VPN, rather than by their directly associated terminations. Filtering by the termination ID probably isn't very useful, because 1) it will only return one VLAN per termination, and 2) if you know the termination ID you likely already have the VLAN ID.

abhi1693:
Alright, I guess I misunderstood the issue, I'll try again on this

abhi1693:
@jeremystretch I have made the change as per your comment. Hopefully, I got it right this time

jeremystretch:
Thanks!

PR #11514, opened 2023-01-16
Added db read-only mode implementation

### LINKED ISSUE #11233: Webfrontend with database in readonly mode
### NetBox version

v3.3.9

### Feature type

### Proposed functionality

The current 'Maintenance mode' ensures that no writes are done to the database. It also displays a banner that the instance is in maintenance mode.

I would like to have a 'ReadOnly' mode that
- like maintenance mode, prevents updates by logins.
- don't display a banner (custom banner can be set with the BANNER_TOP)
- don't do any database updates with the upgrade.sh script
- optionally: provide a message to the user if he tries to do a write operation, with info to the read/write
- optionally: use a separate database for session logs

### Use case

We use postgres WAL to create a almost-realtime readonly copy of our production database.
This can be used to poll data from with API calls and the GUI.

In order to maintain this node, we need to patch the upgrade.sh (remove database operation), but that breaks the git pull.
Also, that banner that is show is not relevant to the state of our instance. It is not in maintenance, it is in readonly.
The sessionpath is an option, but I would rather like to send that to a separate database (different postgresql server) to maintain the session state across webfrontends

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:
### Fixes: #11233

Added a wrapper to prevent write operations by raising an exception and catching it to gracefully inform the user.

### DIFF
--- a/netbox/netbox/cursor.py
+++ b/netbox/netbox/cursor.py
@@ -1,7 +1,6 @@
 import logging
 import time
 
-from django.conf import settings
 from django.db.backends.utils import CursorWrapper as _CursorWrapper
 
 from netbox.exceptions import DatabaseWriteDenied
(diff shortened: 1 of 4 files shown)


jeremystretch:
Thanks for taking this on @abhi1693. Unfortunately I don't think we can accept this in its current form, as a large portion of the PR seems to have been lifted directly from the [django-db-readonly](https://github.com/streeter/django-db-readonly) project. There's no issue using the same technical approach that that project does, however all of the code for the implementation needs to be original.

jeremystretch:
@abhi1693 are you interested in working on this further? If not, I understand.

abhi1693:
I dont think I'll be able to work on this sooner.

jeremystretch:
Ok, thanks. In that case, I'll close this out. Please feel free to take another stab at it in the future.

PR #11526, opened 2023-01-17
Update Database Privilege command for PostgreSQL

navjotjsingh:
The updated command will make Netbox work with PostgreSQL 15.

Fixes: #11314

### DIFF
--- a/docs/installation/1-postgresql.md
+++ b/docs/installation/1-postgresql.md
@@ -52,9 +52,9 @@ sudo -u postgres psql
 Within the shell, enter the following commands to create the database and user (role), substituting your own value for the password:
 
 ```postgresql
-CREATE DATABASE netbox;
 CREATE USER netbox WITH PASSWORD 'J5brHrAXFLQSif0K';
-GRANT ALL PRIVILEGES ON DATABASE netbox TO netbox;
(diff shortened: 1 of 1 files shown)


abhi1693:
This is already documented at https://docs.netbox.dev/en/stable/installation/1-postgresql/ under the section `Database Creation`

image

navjotjsingh:
That doesn't work with Postgresql 15.

DanSheps:
There is a schema permissions change as of PG15:

https://www.crunchydata.com/blog/be-ready-public-schema-changes-in-postgres-15

There are two options as I see it:

1. Alter the database owner
2. Alter the permissions on the schema

I think either are equally as valid but there needs to be more discussion on this first and it should be done through a proper bug report, not a discussion opened that none of the maintainer team has had a chance to comment on or in the PR itself.

Please raise this as a proper BR.

navjotjsingh:
I opened it as a bug report which got converted to discussion.

jeremystretch:
Please resubmit this as a proposed [documentation change](https://github.com/netbox-community/netbox/issues/new?assignees=&labels=type%3A+documentation&template=documentation_change.yaml) and we can discuss.

Issue #11454, opened 2023-01-10
IPAddress model accepts inconsistent data for GenericForeignKey "assigned_object"


moonrail:
### NetBox version

v3.4.2

### Python version

3.10

### Steps to Reproduce

1. Have a Device/VM with Interface
2. Have an IP
3. Assign IP to Interface
4. PATCH /ipam/ip-addresses/{id}/ and send only either `assigned_object_type` or `assigned_object_id` with value `null`

### Expected Behavior

Request either fails with HTTP 400 and an explanation why the client sent inconsistent data or the backend sets the other attribute, that is not sent, to null/None as well.

Or - if the current behaviour is accepted - the UI does not fail with HTTP 500 `'NoneType' object has no attribute 'get_absolute_url'` like described in https://github.com/netbox-community/netbox/discussions/11112

### Observed Behavior

Backend does exactly what the client requests and sets only the sent attribute to the value.
So now IP-Address has one `assigned_object_` value referencing something and one `assigned_object_` value not referencing something.

Causes this issue in the UI: https://github.com/netbox-community/netbox/discussions/11112

'NoneType' object has no attribute 'get_absolute_url'

jeremystretch:
> 4. PATCH /ipam/ip-addresses/{id}/ and send only either assigned_object_type or assigned_object_id with value null

Please share the complete API request you're sending.

moonrail:
@jeremystretch 
Here you go, straight from Swaggerui:
```bash
curl -X 'PATCH' \
  'https://netbox.domain.tld/api/ipam/ip-addresses/123/' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -H 'X-CSRFToken: XYZ' \
  -d '{
  "assigned_object_id": null
}'
```

jeremystretch:
This seems to overlap quite a bit with #10221. I'm not sure yet whether the fix for one will effectively resolve the other, but it's likely.

abhi1693:
@jeremystretch I'd like to take this

DanSheps:
@abhi1693 Going to assign this to you, feel free to proceed.

abhi1693:
@DanSheps Can I also get #10221? My solution fixes these both together

PR #11550, opened 2023-01-20
Fixes GenericForeignKey validation

### LINKED ISSUE #10221: IP address model does not check if assigned_object exists
### NetBox version

v3.3.1

### Python version

3.10

### Steps to Reproduce

1. Create IP address "127.0.0.1/8"
2. Create a PATCH request to /api/ipam/ip-addresses/:
```
curl -X PATCH \
       -H  "accept: application/json" -H  "Authorization: Token $NETBOX_TOKEN" \
      --data '{"assigned_object_type": "virtualization.vminterface", "assigned_object_id":4711}' \
     "http://localhost:8000/api/ipam/ip-addresses/1/"
```

### Expected Behavior

An error is returned that a vminterface with ID 4711 does not exist.

### Observed Behavior

assigned_object_id and assigned_object_type are saved in the model and returned on subsequent GET requests.
```
{
  "id": 1,
  "url": "http://localhost:8000/api/ipam/ip-addresses/1/",
[...]
  "address": "127.0.0.1/8",
[...]
  "assigned_object_type": "virtualization.vminterface",
  "assigned_object_id": 4711,
  "assigned_object": null,
[...]
}
```

abhi1693:
### Fixes: #10221, #11454

Added a validation on `NetBoxModel` `clean()` method for GenericForeignKey fields to ensure that the content type and object ID exist.

### DIFF
--- a/netbox/netbox/models/__init__.py
+++ b/netbox/netbox/models/__init__.py
@@ -67,6 +67,15 @@ class NetBoxModel(CloningMixin, NetBoxFeatureSet, models.Model):
 
         for field in self._meta.get_fields():
             if isinstance(field, GenericForeignKey):
+                if getattr(self, field.ct_field) is None and getattr(self, field.fk_field) is not None:
+                    raise ValidationError({
+                        field.ct_field: "This field cannot be null.",
+                    })
+                if getattr(self, field.fk_field) is None and getattr(self, field.ct_field) is not None:
+                    raise ValidationError({
+                        field.fk_field: "This field cannot be null.",
+                    })
+
                 if getattr(self, field.ct_field) and getattr(self, field.fk_field):
                     klass = getattr(self, field.ct_field).model_class()
                     try:
--- a/netbox/netbox/models/__init__.py
+++ b/netbox/netbox/models/__init__.py
@@ -67,22 +67,23 @@ class NetBoxModel(CloningMixin, NetBoxFeatureSet, models.Model):
 
         for field in self._meta.get_fields():
             if isinstance(field, GenericForeignKey):
-                if getattr(self, field.ct_field) is None and getattr(self, field.fk_field) is not None:
+                ct_field = getattr(self, field.ct_field)
+                fk_field = getattr(self, field.fk_field)
+
+                if ct_field is None and fk_field is not None:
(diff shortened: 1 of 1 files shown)


abhi1693:
@DanSheps Could you please review this as you had assigned me the issues? If there are changes needed, I'd be happy to make them

DanSheps (approved):
Looks good to me, however I would like @jeremystretch to review it.

netbox/netbox/models/__init__.py
```diff
@@ -58,6 +59,32 @@ class NetBoxModel(CloningMixin, NetBoxFeatureSet, models.Model):
...
 
+    def clean(self):
+        """
+        Validate the model for GenericForeignKey fields to ensure that the content type and object ID exist.
+        """
+        super().clean()
+
+        for field in self._meta.get_fields():
+            if isinstance(field, GenericForeignKey):
+                if getattr(self, field.ct_field) is None and getattr(self, field.fk_field) is not None:
```
jeremystretch, line 70: Instead of calling `getattr()` for either field each time, we can just call it twice to define `ct_field` and `fk_field` at the beginning and then reference those values.

netbox/netbox/models/__init__.py
```diff
@@ -58,6 +59,32 @@ class NetBoxModel(CloningMixin, NetBoxFeatureSet, models.Model):
...
+                    })
+                if getattr(self, field.fk_field) is None and getattr(self, field.ct_field) is not None:
+                    raise ValidationError({
+                        field.fk_field: "This field cannot be null.",
+                    })
+
+                if getattr(self, field.ct_field) and getattr(self, field.fk_field):
+                    klass = getattr(self, field.ct_field).model_class()
+                    try:
+                        klass.objects.get(pk=getattr(self, field.fk_field))
```
jeremystretch, line 82: Calling `get()` retrieves and instantiates the entire object from the database. Since we only need to check for its existence, we can do:

```suggestion
                        klass.objects.filter(pk=getattr(self, field.fk_field)).exists()
```

jeremystretch (changes requested):
Thanks @abhi1693, and sorry for the delay in getting to this.

PR #11552, opened 2023-01-20
11487 remove set null from read-only custom fields bulk edit

arthanson:
### Fixes: #11487 

Removes the "Set Null" when bulk editing items that have custom fields when the custom fields UI visibility is not set to read-write (the other choices are read-only and hidden). 

Also downgrades sentry-sdk to 1.13.0 as tests were complaining that 1.13.1 is not available? Which is strange, but maybe pypi issue or the removed it.

### DIFF
--- a/netbox/netbox/forms/base.py
+++ b/netbox/netbox/forms/base.py
@@ -131,7 +131,7 @@ class NetBoxModelBulkEditForm(BootstrapMixin, CustomFieldsMixin, forms.Form):
 
     def _extend_nullable_fields(self):
         nullable_custom_fields = [
-            name for name, customfield in self.custom_fields.items() if not customfield.required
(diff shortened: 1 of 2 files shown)


netbox/netbox/forms/base.py
```diff
@@ -131,7 +131,7 @@ def _get_form_field(self, customfield):
 
     def _extend_nullable_fields(self):
         nullable_custom_fields = [
-            name for name, customfield in self.custom_fields.items() if not customfield.required
+            name for name, customfield in self.custom_fields.items() if (not customfield.required and customfield.ui_visibility == CustomFieldVisibilityChoices.VISIBILITY_READ_WRITE)
```
jeremystretch, line 134:
```suggestion
            name for name, customfield in self.custom_fields.items()
            if not customfield.required and customfield.ui_visibility == CustomFieldVisibilityChoices.VISIBILITY_READ_WRITE
```

PR #11564, opened 2023-01-23
Fixes: #10720 Add small delay to the menu onmouseover

### LINKED ISSUE #10720: Add small delay to the menu onmouseover

PieterL75:


netbox/project-static/styles/sidenav.scss
```diff
@@ -78,6 +78,7 @@
   background-color: var(--nbx-sidebar-bg);
   border-right: 1px solid $border-color;
   transition: $transition-100ms-ease-in-out;
+  transition-delay: 200ms;
```
jeremystretch, line 81: After some brief experimentation, this feels a bit sluggish. 100ms feels more responsive IMO while still introducing sufficient delay in trigger the transition.

netbox/project-static/src/sidenav.ts
```diff
@@ -27,6 +27,12 @@ class SideNav {
    */
   private sections: Section[] = [];
 
+  /**
+   * Tracker if the mouse on the navbar or not
+   */
+  // private mouseOver: boolean = false;
+
+
```
jeremystretch, line 35: Should be free to remove all these commented-out script lines.

jeremystretch (changes requested):
This approach works well. Can we also impose the same delay to the other nav menu items with transitions (the logo and menu headings)?

netbox/project-static/src/sidenav.ts
PieterL75: Yep.. I left them in, just in case the CSS was not doing it as we want

netbox/project-static/styles/sidenav.scss
PieterL75: 100ms it is

jeremystretch:
I think it may be as simple as introducing the delay wherever `transition: $transition-100ms-ease-in-out;` is applied. Or better yet, we can probably include the delay as part of that variable's definition in `theme-base.scss`.

PieterL75:
I tried adding the delay in several places, but the activateLink() is just doing it's thing... 
Funny thing: when I add a CSS for the navlink to open slow, it works when click the menuitems itself (like devices/ipam/...). But when the navbar shows, it just goes fast...

Need to check on this later..

jeremystretch:
@PieterL75 did you want to keep digging into this? We'll likely look at migrating the UI to Tabler in the near future (see #12128), so I'm not sure it'll be worth the effort.

jeremystretch:
Yeah, my hope is to overhaul the entire navigation menu with something built into Tabler, which will reduce our UI development burden significantly.

Going to close this out. Thanks!

PR #11566, opened 2023-01-23
Fixes #11267: Avoid catching ImportError exceptions when loading plugins

### LINKED ISSUE #11267: ImportErrors inside plugins are currently ignored → add logging

jeremystretch:
### Fixes: #11267

- Employ `module_has_submodule()` to check for presence of plugin URL patterns
- Remove default paths from PluginConfig class
- If plugin resource path has not been explicitly set, use `find_spec()` to check default path before importing

### DIFF
--- a/netbox/extras/plugins/__init__.py
+++ b/netbox/extras/plugins/__init__.py
@@ -1,4 +1,5 @@
 import collections
+from importlib.util import find_spec
 
 from django.apps import AppConfig
 from django.conf import settings
@@ -19,6 +20,15 @@ registry['plugins'] = {
     'menu_items': {},
     'preferences': {},
     'template_extensions': collections.defaultdict(list),
+}
+
+DEFAULT_RESOURCE_PATHS = {
+    'search_indexes': 'search.indexes',
+    'graphql_schema': 'graphql.schema',
+    'menu': 'navigation.menu',
+    'menu_items': 'navigation.menu_items',
+    'template_extensions': 'template_content.template_extensions',
+    'user_preferences': 'preferences.preferences',
 }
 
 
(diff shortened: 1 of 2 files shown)


jsenecal (approved):
Looks good to me, nice reuse of django builtins

PR #11567, opened 2023-01-23
11432 device field

### LINKED ISSUE #11432: The device field on device-components (except InventoryItem) should not be writeable via. the API
### NetBox version

v3.4.2

### Python version

3.10

### Steps to Reproduce

1. Create a device with database id 1 and an interface with database id 1
2. Create a device with database id 2
3. Send the following request to the API:
URL: http://127.0.0.1:8000/api/dcim/interfaces/1/
JSON data:
```
{
  "device": 2
}
```

### Expected Behavior

The interface should not be able to be moved to another device. We do not allow this in the UI. There are potential issue regarding cables and caching of paths.

### Observed Behavior

The interface is moved to the other device.

We should disallow this for all devicecomponents except InventoryItem, as this is the way it works in the UI now.

arthanson:
### Fixes: #11432 

Makes the device field for interface read-only if a patch request.  The 'normal' way of doing this would be by override get_extra_kwargs, but that doesn't work if the field is explicitly declared (which this is).  As read-only it is just ignored if the patch is done which is the DRF convention.


jeremystretch:
This may be better addressed in the model itself. I know the bug was opened specifically for the API, however it's something we should prevent via other means as well. Once an interface has been assigned a PK, its assigned device should not be allowed to change.

kkthxbye-code:
The itention was to match the UI, which only allows reassigning the device of `InventoryItem`, but not the other device components listed here:

    'ConsolePort',
    'ConsoleServerPort',
    'DeviceBay',
    'FrontPort',
    'Interface',
    'ModuleBay',
    'PowerOutlet',
    'PowerPort',
    'RearPort',

arthanson:
@jeremystretch should be ready now, tests fixed.

netbox/dcim/api/serializers.py
```diff
@@ -31,6 +31,31 @@
...
+    def get_field_name(self):
+        return "device"
+
+
+class ReadonlyDeviceTypeMixin(BaseReadonlyDeviceMixin):
+
+    def get_field_name(self):
+        return "device_type"
+
+
```
jeremystretch, line 58: Can we simplify this a bit by doing something like this? Also, I'd really like to avoid introducing a new public method on the serializer as it could result in confusion.

```suggestion
class ReadonlyDeviceMixin:
    _field_name = 'device'

    def get_fields(self, *args, **kwargs):
        fields = super().get_fields(*args, **kwargs)

        if self.instance:
            fields[self._field_name].read_only = True

        return fields


class ReadonlyDeviceTypeMixin(ReadonlyDeviceMixin):
    _field_name = 'device_type'


```

jeremystretch:
I think this still needs to be addressed in the model. With the field set to read-only, any attempt to change the parent device/device type of an existing component is silently ignored, because the field is ignored. However, any other valid attributes in the request will be applied successfully. I think we should instead return a validation error, so that the client is aware the entire proposed change did not succeed.

arthanson:
@jeremystretch changed to model based validator, as discussed this will not be officially correct in the OpenAPI spec.  Looked at doing it as read_only in the serializer and overriding validate_device, but that is never called if read_only is set.  Can also be done fairly cleanly in the validate_device if no read_only is set, but that will only address it for the API call and you had mentioned wanting at model level so moved it there.

PR #11621, opened 2023-01-30
Closes #11152: Add support to abort custom script gracefully

### LINKED ISSUE #11152: Allow custom scripts to fail gracefully on errors detected by the scripts
### NetBox version

v3.3.9

### Feature type

### Proposed functionality

It would be nice if a custom script (and I guess a report, too), would have the option to throw a special kind of Exception (say `AbortScript (msg: str)`, which stops the script, aborts the DB transaction, and stores the error `msg` so the caller can see the fact that the script was aborted and why.

Happy to open a PR to add this into the `_run_script()` function. If there is a way to do this, which I didn't find, I'm more than happy for a pointer :-)

### Use case

I wrote a scripts which should aid in provisioning things inside our network, which will do some sanity checks and if all is good set up a number of things (prefixes, IPs, sub-interfaces, IPs on them, etc.). The reason to do this inside a script is to do this in an atomic operation, so either the full provision process is done, or nothing is changed at all.

Now I was looking into exiting the script on an error and only found the option to throw any `Exception` which will produce a stack trace of all my internal Exceptions I caught and handled. When I throw a `AbortTransaction` Exception from the script I can't pass any message.

### Database changes

None.

### External dependencies

None.

BarbarossaTM:
### Fixes: #11152

This PR adds a new `AbortScript` exception which a custom script can throw in case the script should be terminated cleanly. It allows specifying an error message which will be reported to the user, but no stack trace is shown. Any other unhandled exception will still raise a stacktrace.

### DIFF
--- a/docs/customization/custom-scripts.md
+++ b/docs/customization/custom-scripts.md
@@ -142,6 +142,19 @@ obj.full_clean()
 obj.save()
 ```
 
+## Error handling
+
+Sometimes things go wrong and a script will run into an `Exception`. If that happens and an uncaught exception is raised by the custom script, the execution is aborted and a full stack trace is reported.
+
+Although this is helpful for debugging, in some situations it might be required to cleanly abort the execution of a custom script (e.g. because of invalid input data) and thereby make sure no changes are performed on the database. In this case the script can throw an `AbortScript` exception, which will prevent the stack trace from being reported, but still terminating the script's execution and reporting a given error message.
+
+```python
+from utilities.exceptions import AbortScript
+
+if some_error:
+    raise AbortScript("Some meaningful error message")
+```
+
 ## Variable Reference
 
 ### Default Options
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -21,7 +21,7 @@ from extras.models import JobResult
 from extras.signals import clear_webhooks
 from ipam.formfields import IPAddressFormField, IPNetworkFormField
 from ipam.validators import MaxPrefixLengthValidator, MinPrefixLengthValidator, prefix_validator
-from utilities.exceptions import AbortTransaction
+from utilities.exceptions import AbortScript, AbortTransaction
 from utilities.forms import add_blank_choice, DynamicModelChoiceField, DynamicModelMultipleChoiceField
 from .context_managers import change_logging
 from .forms import ScriptForm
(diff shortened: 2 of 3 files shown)


netbox/extras/scripts.py
```diff
@@ -470,6 +470,14 @@ def _run_script():
         except AbortTransaction:
             script.log_info("Database changes have been reverted automatically.")
             clear_webhooks.send(request)
+        except AbortScript as a:
```
jeremystretch, line 473: This is a _super_ minor nit, but could you name the captured exception `e` rather than `a`? This is purely a convention thing.

```suggestion
        except AbortScript as e:
```

netbox/utilities/exceptions.py
```diff
@@ -24,6 +24,14 @@ def __init__(self, message):
         self.message = message
 
 
+class AbortScript(Exception):
+    """
+    Raised to cleanly abort a script.
+    """
+    def __init__(self, message):
+        self.message = message
```
jeremystretch, line 32: Message assignment is actually handled for us by the base class, so we don't need to replicate it explicitly.
```suggestion
    pass
```

docs/customization/custom-scripts.md
```diff
@@ -142,6 +142,19 @@ obj.full_clean()
 obj.save()
 ```
 
+## Error handling
+
+Sometimes things go wrong and a script will run into an `Exception`. If that happens and an uncatched Exception is raised by the custom script, the execution is aborted and a full stacktrace reported.
```
jeremystretch, line 147:
```suggestion
Sometimes things go wrong and a script will run into an `Exception`. If that happens and an uncaught exception is raised by the custom script, the execution is aborted and a full stack trace is reported.
```

docs/customization/custom-scripts.md
```diff
@@ -142,6 +142,19 @@ obj.full_clean()
...
+
+Although this is helpful for debugging in some situations it might be required to cleanly abort the execution of a custom script (e.g. because of invalid input data) and thereby make sure no changes are performed on the database. In this case the script can throw a `AbortScript` exception, which will prevent the stacktrace from being reported, but still terminating the script's execution and reporting a given error message.
```
jeremystretch, line 149:
```suggestion
Although this is helpful for debugging, in some situations it might be required to cleanly abort the execution of a custom script (e.g. because of invalid input data) and thereby make sure no changes are performed on the database. In this case the script can throw an `AbortScript` exception, which will prevent the stack trace from being reported, but still terminating the script's execution and reporting a given error message.
```

jeremystretch (changes requested):
A couple very minor notes, but this is great work! Thank you!

jeremystretch:
Thanks @BarbarossaTM!

BarbarossaTM:
Thanks for the native speaker language polish :-)

I haven't used report myself yet, but I assume it might be useful there too, I guess the place to do it there would be in

```
def run_report(job_result, *args, **kwargs):
```

Do you want to me look into this and file another PR?

jeremystretch:
If you'd like to. Or we can wait and see if this gets any feedback first.

Issue #11658, opened 2023-02-03
Migrations fail on the feature branch because search indexing is run before later schema changes


kkthxbye-code:
### NetBox version

v3.5.0-dev

### Python version

3.10

### Steps to Reproduce

1. Checkout the feature branch
2. Run ./upgrade.sh on a clean database

### Expected Behavior

The upgrade script should finish.

### Observed Behavior

Migration log and exception:

https://gist.github.com/kkthxbye-code/4dd23a639bf0af94e856032fbac83920

The error is caused because `extras.0083_search` runs before `dcim.0167_module_status`. The search migration runs the search indexing command, which queries the Module model, which fails because the Model doesn't match the database yet. We might need to remove the indexing command and re-think how initial search indexing should be executed.

PR #11686, opened 2023-02-07
Fixes #11685 - Include containing prefixes and aggregates when searching for IP addresses

### LINKED ISSUE #11685: Show containing Prefixes and Aggregates when doing a site-wide partial search
### NetBox version

v3.4.4

### Feature type

### Proposed functionality

Relevant bug report: https://github.com/netbox-community/netbox/issues/11315
Relevant discussion: https://github.com/netbox-community/netbox/discussions/11665

In short the goal is to find prefixes and aggregates based on the input, if the input is parsable as an netaddr.IPNetwork.

### Use case

It is useful to get the parent prefix if searching for both existing and non-existent IPAddresses and more closely matches the previous global search. More reasoning can be found in the linked bug report.

### Database changes

While no actual database changes should be necessary, changing the CachedValue.value field to a custom field type inherating TextField generates a migration, but there should be no actual SQL schema changes. See the PR.

### External dependencies

None

kkthxbye-code:
### Fixes: #11685

* Changes the CachedValue types of IPAddressField and IPNetworkField to `inet` and `cidr`.
* Creates a child class of TextField called CachedValueField (not sure about the name). Reasoning as discussed on slack is to prevent adding special lookup types to the global TextField field.
* A NetContainsOrEquals lookup has been added to CachedValueField which casts the string value to inet and does a >>= lookup on it.
* Changing the type create a new migration, however as far as I can tell, it's pretty much a NOP migration, there should be no change to the database schema. Not sure if this is preventable.
* If the search query in partial mode parses as a IP address, all CachedValues of type `cidr` are queried to check if they contain the IP being searched for.

There's some screenshots here: https://github.com/netbox-community/netbox/discussions/11665#discussioncomment-4882273

There's still time to evaluate if we want to do something special about searching for `inet` type fields (IPAddresses), which we can handle in this PR or later.


kkthxbye-code:
Forgot to mention, for this to function for existing objects the user needs to run the reindex mangement command.

jeremystretch:
Great work @kkthxbye-code, thank you!

> Changing the type create a new migration, however as far as I can tell, it's pretty much a NOP migration, there should be no change to the database schema. Not sure if this is preventable.

In such cases where a generated migration doesn't actually effect a material schema change, I like to just alter the field's definition in the most recent relevant migration. (Happy to handle this myself.)

> the user needs to run the reindex mangement command

Any thoughts on how to best handle this? Should we add `reindex` to the upgrade script? (That would help fix #11658 too.)

netbox/extras/lookups.py
```diff
@@ -14,4 +15,18 @@ def as_sql(self, qn, connection):
...
+    """
+    This lookup has the same functionality as the one from the ipam app except lhs is cast to inet
+    """
+    lookup_name = 'net_contains_or_equals'
+
+    def as_sql(self, qn, connection):
+        lhs, lhs_params = self.process_lhs(qn, connection)
+        rhs, rhs_params = self.process_rhs(qn, connection)
+        params = lhs_params + rhs_params
+        return 'CAST(%s as inet) >>= %s' % (lhs, rhs), params
```
jeremystretch, line 28: Just for readability
```suggestion
        return 'CAST(%s AS INET) >>= %s' % (lhs, rhs), params
```

netbox/extras/lookups.py
```diff
@@ -14,4 +15,18 @@ def as_sql(self, qn, connection):
         return 'CAST(LENGTH(%s) AS BOOLEAN) != %s' % (lhs, rhs), params
 
 
+class NetContainsOrEquals(Lookup):
+    """
+    This lookup has the same functionality as the one from the ipam app except lhs is cast to inet
+    """
+    lookup_name = 'net_contains_or_equals'
```
jeremystretch, line 22: Have you considered using a different name for this lookup? Maybe something like `contains_ipaddress`? I don't know if we _should_ differentiate it from the existing `net_contains_or_equals` lookup; just throwing it out there.

jeremystretch (commented):
Awesome work!

jeremystretch:
We might also want to consider reassigning fields weight. Currently, containing prefixes/aggregates have the same precedence (100) as an exact IP address match (see screenshot). We could tweak the weights a bit across the three models so that IPs < prefixes < aggregates (e.g. 100, 110, and 120).

Screenshot 2023-02-17 at 16-10-14 Search NetBox

kkthxbye-code:
>>the user needs to run the reindex mangement command

>Any thoughts on how to best handle this? Should we add reindex to the upgrade script? (That would help fix https://github.com/netbox-community/netbox/issues/11658 too.)

My only worry about doing it in the upgrade script, is that the reindexing can take a while. Maybe we can conditionally run it if there has been applied any migration earlier in the script?

Thanks for the rest of the feedback, I'll address them shortly!

netbox/extras/lookups.py
kkthxbye-code: I think the current name makes sense as it's the same lookup just working with a different datatype. I'll leave it for now, if you feel like it would be better with another name, feel free to change it when merging 👍

kkthxbye-code:
I addressed the feedback @jeremystretch. Only outstanding issue is how to make the user reindex. I see the following options:

* reindex on all runs of ./upgrade.sh
* reindex when running ./upgrade.sh, but only if migrations have been applied in the same run. This would require that we leave the "empty" migration in.
* Just note it in the release notes that reindex should be executed. We might also add a notice to the end of the ./upgrade.sh output telling the user to run reindex if the release notes say so. The search still works without reindexing, just not with the new feature.

jeremystretch:
What if we delete the cache in a new migration, and update `upgrade.sh` to run migrations if none already exist. That way we're only ever rebuilding the whole thing if there's a need to do so. This would solve for #11658 as well.

kkthxbye-code:
> What if we delete the cache in a new migration, and update `upgrade.sh` to run migrations if none already exist. That way we're only ever rebuilding the whole thing if there's a need to do so. This would solve for #11658 as well.

That's a great idea, do you want to implement it or should I add it to the PR?

jeremystretch:
> That's a great idea, do you want to implement it or should I add it to the PR?

I'll handle it under #11787; let's leave this as-is.

PR #11696, opened 2023-02-08
Fixes #11582: Fix missing VC form errors

### LINKED ISSUE #11582: Modifying a Virtual Chassis fails without error message when member device is missing a required custom field
### NetBox version

v3.4.3

### Python version

3.10

### Steps to Reproduce

1. Create a device
2. Create a custom field marking it as required and choosing DCIM > Device as a Content Type
3. Create a virtual chassis adding the device created in step 1 as a member.  Save the virtual chassis
4. Edit the virtual chassis created in step 4. Make any change except removing the member device and hit OK

Same behavior occurs in the Add Member page.
1. Create a device
2. Create a custom field marking it as required and choosing DCIM > Device as a Content Type
3. Create a virtual chassis without adding a member device. Save the virtual chassis
4. Try to add the member device created in step 1 to the virtual chassis created in step 3 via the add member page.

### Expected Behavior

Either the change should be applied and the browser should return the virtual chassis view page for the modified chassis or there should be an error popup indicating that a required field is missing for the member device.   Given that it isn't obvious that the member device is being modified while editing the virtual chassis (especially when only trying to add a comment), applying the change would be more intuitive but the device missing a required field is an error state that could have further impact in the future so failing with a clear error message might be more beneficial.

### Observed Behavior

The Edit Virtual Chassis page reloads without error message or any indication of what the issue is.  The change is not applied.

It appears the missing required field on the member device is causing the formset.is_valid() check to fail in the post methods of the VirtualChassisEditView and VirtualChassisAddMemberView

kkthxbye-code:
### Fixes: #11582

Not sure if this is the correct fix or not. The reason that the custom field errors were not shown is that messages.html only shows non_field_errors if the form passed to the context is named form. This is probably an issue in more places, but not sure how to make it generic. A change to messages.html would also need to support formsets.

Any input appreciated @jeremystretch or @arthanson

### DIFF
--- a/netbox/templates/dcim/virtualchassis_add_member.html
+++ b/netbox/templates/dcim/virtualchassis_add_member.html
@@ -5,6 +5,8 @@
 
 {% block content %}
     <form action="" method="post" enctype="multipart/form-data" class="form-object-edit">
+        {% render_errors membership_form %}
+        
         {% csrf_token %}
         <div class="card">
             <h5 class="card-header">Add New Member</h5>
--- a/netbox/templates/dcim/virtualchassis_edit.html
+++ b/netbox/templates/dcim/virtualchassis_edit.html
@@ -8,6 +8,10 @@
   <div class="tab-content">
     <div class="tab-pane show active" id="edit-form" role="tabpanel" aria-labelledby="object-list-tab">
       <form action="" method="post" enctype="multipart/form-data" class="form-object-edit">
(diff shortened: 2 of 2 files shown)


jeremystretch:
LGTM. Virtual chassis in general is a bit of a snowflake. Hopefully we can achieve better standardization with these templates/views in the near future, to avoid issues such as this one.

PR #11697, opened 2023-02-08
Fixes #11601 - Add partial lookup to IPRangeFilterSet

### LINKED ISSUE #11601: IP Range search not working as expected
### NetBox version

v3.4.3

### Python version

3.11

### Steps to Reproduce

1. Log into https://demo.netbox.dev
2. [Go to IP-Ranges](https://demo.netbox.dev/ipam/ip-ranges/)
3. Search for one of the IP ranges, eg. `192.168.1.50`, or simply `192`

### Expected Behavior

All IP ranges with a matching substring of the search shows, kinda like `?q=` works on `ip_addresses`. (It's the same for the API and pynetbox)

### Observed Behavior

No results found.
Search in GUI and `?q=`  does not show any results at all.
`?contains=` seems to work when searching for an IP (with /cidr), and URL encodes the `/`. `192.168.1.66/22` does not work, but `192.168.1.66%2f22` works. But I can't get contains to work via pynetbox.

kkthxbye-code:
### Fixes: #11601 

Allows partial lookup in the search method for IPRange. Also removes the `.cidr` from the parsed IP lookup as I really don't see how it makes sense.

The changes makes it so that the following matches (192.168.0.1/25-192.168.0.25/25, which didn't before:
192
192.16
192.168.0.1/25 <- even the full IP with mask didn't match before because of using the cidr.
192.168.0.25/25

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -441,9 +441,9 @@ class IPRangeFilterSet(TenancyFilterSet, NetBoxModelFilterSet):
     def search(self, queryset, name, value):
         if not value.strip():
             return queryset
-        qs_filter = Q(description__icontains=value)
+        qs_filter = Q(description__icontains=value) | Q(start_address__contains=value) | Q(end_address__contains=value)
         try:
-            ipaddress = str(netaddr.IPNetwork(value.strip()).cidr)
+            ipaddress = str(netaddr.IPNetwork(value.strip()))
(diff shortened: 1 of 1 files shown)


jeremystretch:
> Also removes the `.cidr` from the parsed IP lookup as I really don't see how it makes sense.

My guess is it was mistakenly copied over from PrefixFilterSet back when IP ranges were introduced. (It's used there to match based on a contained IP address.)

PR #11698, opened 2023-02-08
Fixes #11032 - Replication fields broken in custom validation

### LINKED ISSUE #11032: Replication fields broken in custom validation
### NetBox version

v3.3.8

### Python version

3.9

### Steps to Reproduce

After updating NetBox 3.2.x to 3.3.8 the custom validation logic that used to work is now broken.

1. Create a custom validator that tries to validate a replication field, for example for `VMInterface` model try to validate field `name`.
2. Try to add a new interface to a VM.

### Expected Behavior

The custom validator should have access to `name` field.

### Observed Behavior

The `name` field is always empty when creating a new object and the custom validation logic fails.

As per https://github.com/netbox-community/netbox/discussions/11031 this is apparently due to "replication fields (think interface name expansion) was refactored in that PR and seems to have caused the name field to be blank at the time of validation", possibly due to https://github.com/netbox-community/netbox/pull/10312.

kkthxbye-code:
### Fixes: #11032

The issue here is that validation is triggered on the base instance when the form uses `replication_fields`. So if the user creates a vminterface with name test[1-2] validation is triggered for:

VMInterface with no name
VMInterface with name test1
VMInterface with name test2

For the normal validation this works okay, but for custom validation it results in the validator being triggered for the first VMInterface which is never actually saved. Not sure if there is a smarter way to fix this, my approach is to tag the instance that is never going to be saved with a private variable and make sure we don't run custom validators for these instances.

### DIFF
--- a/netbox/netbox/models/features.py
+++ b/netbox/netbox/models/features.py
@@ -258,7 +258,7 @@ class CustomValidationMixin(models.Model):
         super().clean()
 
         # If the instance is a base for replications, skip custom validation
-        if hasattr(self, '_replicated_base'):
+        if getattr(self, '_replicated_base', False):
             return
 
         # Send the post_clean signal
--- a/netbox/netbox/models/features.py
+++ b/netbox/netbox/models/features.py
@@ -257,6 +257,10 @@ class CustomValidationMixin(models.Model):
     def clean(self):
         super().clean()
 
+        # If the instance is a base for replications, skip custom validation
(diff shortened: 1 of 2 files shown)


jeremystretch:
To really resolve this, we'll need to overhaul the manner in which component replication forms are built and processed; I've opened #11707 to address this. For now, this seems like a reasonable stop-gap. Thanks!

PR #11722, opened 2023-02-10
Fixes #11459 - Allow using null in conditions

### LINKED ISSUE #11459: Webhook condition for having a primary IP doesn't appear to be working
### NetBox version

v3.4.2

### Python version

3.10

### Steps to Reproduce

1. Create a webhook, content type 'DCIM > Device', all events, with the condition:
`{
    "and": [
        {
            "attr": "primary_ip",
            "value": "",
            "negate": true
        }
    ]
}`
2. Create a site, all defaults.
3. Create a device role, all defaults.
4. Create a manufacturer, all defaults.
5. Create a device type, all defaults.
6. Create a device.
7. Go to the admin panel, check the jobs in the default queue.

### Expected Behavior

The webhook does not fire due to the condition of requiring a primary IP.

### Observed Behavior

The webhook fires (despite the condition).

kkthxbye-code:
### Fixes: #11459

Issue here was two-fold I believe. First issue is that primary_ip is always None when evaluating the conditions as computed properties are not serialized when queuing webhooks.

The other issue is that comparing "" with None is False. Allowing conditions to contain `null`/`None` allows to compare against None instead of empty string which was not functional.

### DIFF
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -44,7 +44,8 @@ class Condition:
         bool: (EQ, CONTAINS),
         int: (EQ, GT, GTE, LT, LTE, CONTAINS),
         float: (EQ, GT, GTE, LT, LTE, CONTAINS),
-        list: (EQ, IN, CONTAINS)
+        list: (EQ, IN, CONTAINS),
+        type(None): (EQ)
     }
 
     def __init__(self, attr, value, op=EQ, negate=False):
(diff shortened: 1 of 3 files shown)


netbox/extras/conditions.py
```diff
@@ -44,7 +44,8 @@ class Condition:
         bool: (EQ, CONTAINS),
         int: (EQ, GT, GTE, LT, LTE, CONTAINS),
         float: (EQ, GT, GTE, LT, LTE, CONTAINS),
-        list: (EQ, IN, CONTAINS)
+        list: (EQ, IN, CONTAINS),
+        type(None): (EQ)
```
jeremystretch, line 48: I think this needs the comma to be evaluated as a tuple. It's not causing a problem right now because `'eq' in 'eq'` evaluates to True.
```suggestion
        type(None): (EQ,)
```

Issue #11040, opened 2022-11-29
Default values in custom fields applied to imported data


ejstover:
### NetBox version

v3.2.7

### Feature type

Change to existing functionality

### Proposed functionality

A custom field with a default value should always be applied to all future new objects, including objects that are imported. 

### Use case

I personally use the DeviceTypeLibrary, plus a [Zabbix sync script](https://github.com/TheNetworkGuy/netbox-zabbix-sync) to sync NetBox inventory to Zabbix monitoring via API. Part of the requirements of the sync script is a custom field in the dcim.devicetype that specifies the Zabbix monitoring template to use. I've defined a default value that is applied when manually creating a device type, but not when imported via YAML/JSON.
![2022-11-28 20_54_34-Add a new device type _ NetBox — Mozilla Firefox](https://user-images.githubusercontent.com/19699277/204419331-83c157c3-7a82-4224-9846-3fece3943f9f.png)
![2022-11-28 20_55_02-Editing device type MX84 _ NetBox — Mozilla Firefox](https://user-images.githubusercontent.com/19699277/204419333-78f33295-2a56-45e7-9ae4-f025c25b8348.png)

I could see reasoning against this functionality. I have patch panel device types that don't apply to the use case I've listed above.

atownson:
I have found if you define a required custom field for Interfaces with a defined default value, when instantiating a new Device, the value of that custom field for the created Interfaces is `null`. Could we please expand the scope of this request to more than the bulk import functionality and ensure all means of creating models are covered (device templates, scripts, etc.)? Thanks.

jeremystretch:
This can likely be addresses using the same approach as #11565.

abhi1693:
@jeremystretch The issue that you have mentioned as a reference, should that not be applied commonly to all models?

jeremystretch:
Treating this as a bug per #16073

PR #11724, opened 2023-02-10
Fixes #11565 - Populate custom field defaults when creating FHRP groups with VIP

### LINKED ISSUE #11565: IP-addresses created via addition of FHRP group do not inherit default custom field values
### NetBox version

3.1.10

### Python version

3.9

### Steps to Reproduce

1. Create a boolean custom field with a certain default value, for example **False**, and assign it the `ipam | IP address`-model.
2. Add a new FHRP group (IPAM -> FHRP Groups) with a new, unused IP-address under **Virtual IP Address**.
3. The new IP-address object attached to the VHRP-group does not have the custom field set, which generates an error when running reports.

### Expected Behavior

I'd expect the new IP-address to have the default custom field value.

### Observed Behavior

An exception occurred: KeyError: 'my_custom_field' Traceback (most recent call last): File "/usr/lib/netbox/app/extras/reports.py", line 234, in run test_method() File "/usr/local/lib/netbox/reports/my_report.py", line 11, in test_for_field if ip.custom_field_data["my_custom_field"]: 'my_custom_field'

kkthxbye-code:
### Fixes: #11565

Creates a new method in CustomFieldMixin for populating defaults. It is utilized when creating the IP during FHRP creation.

I created a method on the model because it seems likely there's other places where we manually create objects during the creation of other objects. We might look into just always applying the defaults whenever an object with CustomFieldMixin is instantiated.

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -578,6 +578,7 @@ class FHRPGroupForm(NetBoxModelForm):
                 role=FHRP_PROTOCOL_ROLE_MAPPINGS.get(self.cleaned_data['protocol'], IPAddressRoleChoices.ROLE_VIP),
                 assigned_object=instance
             )
+            ipaddress.populate_custom_field_defaults()
             ipaddress.save()
 
             # Check that the new IPAddress conforms with any assigned object-level permissions
--- a/netbox/netbox/models/features.py
+++ b/netbox/netbox/models/features.py
@@ -216,6 +216,13 @@ class CustomFieldsMixin(models.Model):
 
         return dict(groups)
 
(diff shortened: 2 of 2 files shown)


jeremystretch:
I'm not sure about adding a model method for this if it's currently a one-off concern, though I'm also not sure it _is_ a one-off concern.

kkthxbye-code:
@jeremystretch - The same issue is at least present when instantiating devices as the components do not get the default custom field values applied.

https://github.com/netbox-community/netbox/issues/11040

Same should apply to components instantiated when installing a module. I did a quick audit, and these might be the only other places currently applicable. Let me know if you still want it removed and I'll fix it right up!

PR #11730, opened 2023-02-10
Fixes #9653 - Add default_platform to DeviceType

### LINKED ISSUE #9653: Add platform to device type
### NetBox version

v3.2.3

### Feature type

New functionality

### Proposed functionality

Hello,

is it possible to add the point "management - platform" to the device type, so that this is inherited by the device as soon as it is created?

### Use case

We use this option to document the running system of the devices. For example: Juniper -> Junos

Since this is the same for each device type, you could save the step of entering it manually. 
Another point would be that less mistakes are made when entering data.

### Database changes

_No response_

### External dependencies

_No response_

kkthxbye-code:
### Fixes: #9653 

Add default_platform to DeviceType. I think I got all the places updated and tests created. If anything is missing please let me know.

Only thing I was a little in doubt about is if the new field on the DeviceType model should have a related_name or not. In the current PR it doesn't, let me know if I should change it.


jeremystretch:
Great work, thanks @kkthxbye-code!

> Only thing I was a little in doubt about is if the new field on the DeviceType model should have a related_name or not.

I don't see a need for it; we can certainly add one later if the need arises.

PR #11763, opened 2023-02-16
Fixes #11226 - Change the way we invalidate the module cache to support reloading code from subpackages

### LINKED ISSUE #11226: Custom scripts in sub packages should not be cached
### NetBox version

v3.4.1

### Python version

3.10

### Steps to Reproduce

Create the structure described in the following issue:

https://github.com/netbox-community/netbox/issues/10258

Change the `Script` class name or the variable name of any `ScriptVariable`.

### Expected Behavior

The new code should be dynamically reloaded.

### Observed Behavior

Code is not reloaded.

We have had discussions about this at the last maintainers meeting. Dynamically loading code in python is not a pleasent experience, and as such there is a bunch of pitfalls. Currently we delete the module entry in `sys.modules` when loading the script list, this prevent caching.  This doesn't work for nested packages, which this issue seeks to address.

In addition to this, there's other issues that might be also be addressed:

* The way we currently load scripts is to load any module/package in the `SCRIPTS_ROOT` folder in a way that makes the module path be in the root path. Essentially this means that if we load a file called ipam.py in the scripts folder, it will clash with the already existing ipam module.
* `load_module` is to be deprecated in python 3.12, so we need to refactor the script/report loading code at some point soon regardless.

I'll be looking into solutions to these problems, but if anyone has experience with dynamically loading python code in a long running python process, please feel free to give input.

For reference, the loading code is here:

https://github.com/netbox-community/netbox/blob/3675ad2539400038af6797140ddca3679b8bca30/netbox/extras/scripts.py#L513-L543

kkthxbye-code:
### Fixes: #11226

This PR intends to fix cases where script and utility code loaded from subpackages wasn't getting reloaded after changes were made.

The way this is done, is by taking the base module path (the string before the first dot), so if a script module import path is `my_submodule.my_script_file.MyScriptClass` we would delete any module from sys.modules where the base `my_submodule` matches. 

Furthermore we remove any module from sys.modules where the base path equals `scripts`. This was added, because utility functions imported via. the normal import statement, would be imported like:

`from scripts.util.util_file import util_function`

Which would be present in sys.modules as:

`scripts.util.util_file`

Removing all modules from the cache which starts with `scripts` solves this issue, and from my testing causes no other issues. I tested this as well as I could, with the main focus being the default value for `SCRIPT_ROOT`. Subpackage importing when using a non-default `SCRIPT_ROOT` is still not great, but outside the scope of this PR.

In short you can now modify the following without reloading netbox or the rqworker:
* The name of the script class in subpackages
* Script variables in subpackages
* Imported utility functions

### DIFF
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -524,27 +524,39 @@ def get_scripts(use_names=False):
     defined name in place of the actual module name.
     """
     scripts = {}
-    # Iterate through all modules within the scripts path. These are the user-created files in which reports are
+
+    # Get all modules within the scripts path. These are the user-created files in which scripts are
     # defined.
-    for importer, module_name, _ in pkgutil.iter_modules([settings.SCRIPTS_ROOT]):
-        # Use a lock as removing and loading modules is not thread safe
-        with lock:
-            # Remove cached module to ensure consistency with filesystem
-            if module_name in sys.modules:
+    modules = list(pkgutil.iter_modules([settings.SCRIPTS_ROOT]))
+    modules_bases = set([name.split(".")[0] for _, name, _ in modules])
+
+    # Deleting from sys.modules needs to done behind a lock to prevent race conditions where a module is
+    # removed from sys.modules while another thread is importing
+    with lock:
+        for module_name in list(sys.modules.keys()):
+            # Everything sharing a base module path with a module in the script folder is removed.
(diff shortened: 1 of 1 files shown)


jeremystretch:
Awesome, kicked this around a bit and it seems to work great. Thanks @kkthxbye-code!

PR #11779, opened 2023-02-17
Fixes #10987: Show rack-list dropdown in rack 

### LINKED ISSUE #10987: Show rack-list dropdown in rack view
### NetBox version

v3.3.8

### Feature type

### Proposed functionality

When viewing a rack, you have the 'previous' and 'next' button.
image

Add a 'dropdown' between those 2 buttons, so that all racks of a location (or site if no location is set) are selectable
image

### Use case

This will provide a faster navigation between the racks.
Another option, is to display all racks in a full-width bar with horizontal scrolling
image

### Database changes

_No response_

### External dependencies

_No response_

PieterL75:
Fixes #10987: Show rack-list dropdown in rack

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -738,6 +738,7 @@ class RackView(generic.ObjectView):
             'next_rack': next_rack,
             'prev_rack': prev_rack,
             'svg_extra': svg_extra,
+            'peer_racks': peer_racks,
         }
 
 
--- a/netbox/templates/dcim/rack.html
+++ b/netbox/templates/dcim/rack.html
@@ -20,26 +20,26 @@
 {% block extra_controls %}
   <div class="btn-group" role="group" aria-label="RackNavigation">
     {% if prev_rack %}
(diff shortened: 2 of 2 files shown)


PieterL75:
There are currently 2 types of navigation.
Maybe we can make it dynamic. So that the short one is shown on small devices, and the bar on desktop views ?

PieterL75:
I got the resize working (using BS5 classes)
image
and
image

PieterL75:
The dropdown is shown when the view goes in 'mobile' mode.
The static list is show in desktop mode..

I do like that static list, as it navigates faster than the dropdown list..
I might make it slide to 2 rows when the screen gets to small...

jeremystretch:
I'm afraid the static list is a nonstarter. It looks reasonable in your example with extremely short rack names, but is not tenable with even moderately long rack names, or with many peer racks. The only workable solution is a dropdown selector.

jeremystretch:
Thanks @PieterL75!

PR #11785, opened 2023-02-18
Fixes #11775: Skip clearing cache when handling new objects

### LINKED ISSUE #11775: Creating devices takes an inordinate amount of time
### NetBox version

v3.4.4

### Python version

3.10

### Steps to Reproduce

This issue will be used to collect what I find while investigating the issue, as such the scenario is based on my test database and test devicetype.

1. Create a devicetype with 62 interfaces
2. Create a device based on the interface
3. Notice the responsetime

### Expected Behavior

The device should be create within a reasonable time period.

### Observed Behavior

The device takes many seconds to create.

Do note, that while the performance was never great, it was made worse by both the search indexing and the fact that we fixed a bug where changelog entries were not created for instantiated components.

jeremystretch:
### Fixes: #11775

When receiving `post_save`, check whether the object is newly created. If so, skip deleting any pre-existing cache records.

### DIFF
--- a/netbox/netbox/search/backends.py
+++ b/netbox/netbox/search/backends.py
@@ -52,11 +52,11 @@ class SearchBackend:
         """
         raise NotImplementedError
 
-    def caching_handler(self, sender, instance, **kwargs):
+    def caching_handler(self, sender, instance, created, **kwargs):
         """
         Receiver for the post_save signal, responsible for caching object creation/changes.
         """
-        self.cache(instance)
+        self.cache(instance, remove_existing=not created)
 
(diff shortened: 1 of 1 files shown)


kkthxbye-code:
I tried it out and can confirm that it has the same effect on device creation time as adding the index. In general the change just makes sense and will probably improve query time across the board for all creations. I think this is an acceptable first step, if you want you can merge and resolve #11775 - and I'll create a new issue if I find other improvements.

Thanks for handling this :)

jeremystretch:
Sounds good. I've just opened #11790 to perform an application-wide audit for missing GFK indexes.

PR #11788, opened 2023-02-20
Closes #11638: add http redirect to apache

### LINKED ISSUE #11638: Add HTTP redirect to reference Apache configuration
### Change Type

Addition

### Area

Installation/upgrade

### Proposed Changes

Extend the [reference Apache server configuration](https://github.com/netbox-community/netbox/blob/develop/contrib/apache.conf) that ships with NetBox to include a rule that redirects all HTTP requests to HTTPS. (This is already included in the nginx config.)

rganascim:
### Fixes: #11638

Add the Virtualhost for HTTP (using the same ServerName from https). Uses the mod_rewrite to redirect any access to http port to https session. Updates the doc to activate the rewrite module.

### DIFF
--- a/contrib/apache.conf
+++ b/contrib/apache.conf
@@ -1,3 +1,12 @@
+<VirtualHost *:80>
+    # CHANGE THIS TO YOUR SERVER'S NAME
+    ServerName netbox.example.com
+
+    RewriteEngine On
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks @rganascim!

PR #11814, opened 2023-02-23
Closes #11011: Hide virtual interfaces

### LINKED ISSUE #11011: Add the ability to hide virtual interfaces
### NetBox version

v3.3.5

### Feature type

New functionality

### Proposed functionality

Expand on https://github.com/netbox-community/netbox/issues/4247 to add the option to hide virtual interfaces.

Currently there are two options: Hide Enabled and Hide Disabled. I propose a third option of Hide Virtual. 
On button toggle, every interface in InterfaceTable of type "Virtual" should be hidden or shown based on the current state.

Files needing updated:
[netbox/project-static/src/tables/interfaceTable.ts](https://github.com/netbox-community/netbox/commit/e004f872f9b6c5ba1c4ad792b8b10adfc3659c7e#diff-783f2e6ec565238f5d454ad36dc74eb8efb43032bdfbbc7034fc62682c96224b)

[netbox/templates/dcim/device/interfaces.html](https://github.com/netbox-community/netbox/commit/e004f872f9b6c5ba1c4ad792b8b10adfc3659c7e#diff-735e4e23856b0bf848f6d54a14ca359dfc21eda805d680a468842b2f59fc0397)

### Use case

In my network, many of my locations have a Juniper MX204 providing PE services to that region. The MX204 only has 4x100G and 8x10G, however, we may have hundreds or thousands of sub interfaces based on how many circuits we sell. In the interfaces tab of a device, this becomes pretty cumbersome to navigate.

### Database changes

None.

### External dependencies

None.

aronjohanns:
### Closes: #11011

Adds the ability to show/hide virtual interfaces in device's interface table.


aronbj20:
I had to refactor some of the button state logic. Since the enabled/disabled and virtual type are not the same columns the button state would get out of sync with the actual visibility state. Please let me know if there is something I should add/fix/explain or if you have a suggestion for a better way of solving this problem.

jeremystretch:
Seems to work fine for me. Thanks for picking this up @aronbj20!

PR #11815, opened 2023-02-23
Fixes #10058: Allow searching device by primary IP address

### LINKED ISSUE #10058: Allow searching on device page by IP address
### NetBox version

v3.3.0

### Feature type

### Proposed functionality

We often use the devices page. It has a great layout and shows the info we need most of the time. We also frequently search for a machine by IP address. In order to do this we have to navigate to the IP Addresses page and then search there, get the machine name, then go back to the devices page.

It would be great if the 'search' on the devices pages can also search IP addresses on the same page (or perhaps other info as well).

### Use case

Use case described in the functionality text.

### Database changes

Don't believe db changes are needed.

### External dependencies

None.

PieterL75:
Fixes #10058: Allow searching device by primary IP address

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -981,7 +981,9 @@ class DeviceFilterSet(NetBoxModelFilterSet, TenancyFilterSet, ContactModelFilter
             Q(serial__icontains=value.strip()) |
             Q(inventoryitems__serial__icontains=value.strip()) |
             Q(asset_tag__icontains=value.strip()) |
-            Q(comments__icontains=value)
+            Q(comments__icontains=value) |
+            Q(primary_ip4__address__startswith=value) |
+            Q(primary_ip6__address__startswith=value)
         ).distinct()
 
     def _has_primary_ip(self, queryset, name, value):
--- a/netbox/dcim/search.py
+++ b/netbox/dcim/search.py
@@ -41,6 +41,8 @@ class DeviceIndex(SearchIndex):
         ('asset_tag', 50),
         ('serial', 60),
         ('name', 100),
+        ('primary_ip4', 110),
+        ('primary_ip6', 110),
         ('description', 500),
         ('comments', 5000),
     )
--- a/netbox/dcim/search.py
+++ b/netbox/dcim/search.py
@@ -41,8 +41,6 @@ class DeviceIndex(SearchIndex):
         ('asset_tag', 50),
         ('serial', 60),
         ('name', 100),
-        ('primary_ip4', 110),
-        ('primary_ip6', 110),
         ('description', 500),
         ('comments', 5000),
     )


netbox/dcim/search.py
```diff
@@ -41,6 +41,8 @@ class DeviceIndex(SearchIndex):
         ('asset_tag', 50),
         ('serial', 60),
         ('name', 100),
+        ('primary_ip4', 110),
```
kkthxbye-code, line 44: This will not work, primary_ip4 is a foreignkey to the IPAddress model and as such cannot be indexed. Furthermore changing indexes would probably require dumping the existing index in a migration.

netbox/dcim/filtersets.py
```diff
@@ -981,7 +981,9 @@ def search(self, queryset, name, value):
             Q(serial__icontains=value.strip()) |
             Q(inventoryitems__serial__icontains=value.strip()) |
             Q(asset_tag__icontains=value.strip()) |
-            Q(comments__icontains=value)
+            Q(comments__icontains=value) |
+            Q(primary_ip4__address__startswith=value) |
+            Q(primary_ip6__address__startswith=value)
```
kkthxbye-code, line 986: I don't have any strong feelings about this. There's a performance penalty (on my test system, ~10k devices and ~14k interfaces the query time goes from ~45 ms to ~70 ms. However I can see how it could be very useful and since the q search is no longer part of the global search, the performance penalty might be worth it?

kkthxbye-code (changes requested):

netbox/dcim/search.py
PieterL75: Works fine. When adding an IP to a device and checking the 'make primary' it seems to be indexed.
I can search that IP immediately and find the ip, device, prefix, aggregate

netbox/dcim/filtersets.py
PieterL75: It is indeed quite a hit.. Did not test that... but still acceptable for the extra searchability of a device.
I often get an IP where I have to find the device for, this will make that search much more direct

netbox/dcim/search.py
kkthxbye-code: I see why it works, the device is for some reason saved when the primary ip is modified in the UI (empty changelog though). In general we can't index related objects, because it's complex to update the index when related objects change.

In this case, try to modify the primary ip via. the API and change the address. You will end up with an inconsistent cache. Here I change the IPAddress object assigned as primary ip from 199.199.199.3/25 to 199.199.199.3/26. The index for the ip address was updated, but not for the device as none of the fields on the device model changed.

image

So no, it doesn't work.

kkthxbye-code: > Isn't a the reindex is done automatically during the housekeeping ?

No, a full reindex can take 10+ minutes on large instances and during the reindex the search index would be incomplete.

> A note can be added to run the reindex manually if needed.

Currently, as I said, the way we do it is to drop the index for the specific model in a migration and then we have added a reindex --lazy command which is run during ./upgrade.sh. This avoids having to refresh the entire index every upgrade.

PieterL75: That's a bummer.. I really liked the device in those results. I hope this might make it in a different FR, as there is a request to add related objects to the results

I'll remove this from the PR

netbox/dcim/filtersets.py
jeremystretch: > but still acceptable for the extra searchability of a device.

We don't really have any way of knowing this. A user with 100K IPs defined is going to experience a _far_ more severe performance hit than someone with a few hundred (hence my comment [here](https://github.com/netbox-community/netbox/issues/10058#issuecomment-1222922659)).

IMO we should evaluate potential strategies for caching the IP addresses directly on the device model rather than imposing a potentially substantial burden on the user.

jeremystretch: Another consideration: It might be more efficient to perform a separate query against IPAddress (for relevant search queries), and then match on the ID for any device that points to matching IPAddresses. This could be a union with devices matching on other fields.

Additionally, `startswith` probably isn't a sufficient lookup for IP addresses as it's constrained to matching on the octet/nibble boundary,

PieterL75: The search is on the primary ip assigned to a device, not to the list of IPs.
And this is for the quick search/device search, not for the new cached search.

jeremystretch: > The search is on the primary ip assigned to a device, not to the list of IPs.

As it's currently constructed, the query is triggering a SQL `JOIN` with the IP addresses table (two actually), which is very expensive. I'm saying we could instead query the IP addresses table directly for the value, and then work backward to device the IDs of any relevant devices:

1. Does the search term look like an IP address? (If no, skip these steps.)
2. Find all matching IP addresses.
3. For all matching IPs, find all devices pointing to them as primary IPs.
4. Add the IDs of these devices to the base query.

jeremystretch: I built a test bed for this with 10 thousand devices, each with an interface and both IPv4 and IPv6 primary IPs. The resulting SQL query when searching for e.g. `?q=10.1.2.3` takes only 40-50ms on average, which seems very reasonable and probably wouldn't benefit substantially from further optimization. I'm open to potential future work in the event of complaints, but this seems acceptable as-is. Thanks for your work on this @PieterL75!

jeremystretch:
I've replicated this code for virtual machines as well in 6ea30798b.

PR #11820, opened 2023-02-24
Fixes #11817 - Add connected_endpoints property to graphql

### LINKED ISSUE #11817: Add connected_enpoints property to graphql api
### NetBox version

v.3.4.5

### Feature type

### Proposed functionality

The netbox graphql api should expose the same connected_endpoints property as the rest api.
This makes it easier to resolve complete paths via graphql.

A query would look like:

```{
  device(id:xx) {
    id
    name
    interfaces {
      name
      connected_endpoints {
        ... on InterfaceType {
          ip_addresses {
            address
          }
          device {
            name
            device_role {
              slug
            }
          }
        }
      }
    }
  }
}
```

### Use case

This function helps on making graphql information analog to rest-Api information.

### Database changes

_No response_

### External dependencies

_No response_

basti1508:
### Fixes: #11817

add conntected_enpoints property to graphql

### DIFF
--- a/netbox/dcim/graphql/mixins.py
+++ b/netbox/dcim/graphql/mixins.py
@@ -10,3 +10,11 @@ class CabledObjectMixin:
 
     def resolve_link_peers(self, info):
         return self.link_peers
+
+
+class PathEndpointMixin:
+    connected_endpoints = graphene.List('dcim.graphql.gfk_mixins.LinkPeerType')
+
+    def resolve_connected_endpoints(self, info):
+        # Handle empty values
+        return self.connected_endpoints or None
--- a/netbox/dcim/graphql/types.py
+++ b/netbox/dcim/graphql/types.py
@@ -7,7 +7,7 @@ from extras.graphql.mixins import (
 from ipam.graphql.mixins import IPAddressesMixin, VLANGroupsMixin
 from netbox.graphql.scalars import BigInt
 from netbox.graphql.types import BaseObjectType, OrganizationalObjectType, NetBoxObjectType
-from .mixins import CabledObjectMixin
+from .mixins import CabledObjectMixin, PathEndpointMixin
 
 __all__ = (
     'CableType',
@@ -117,7 +117,7 @@ class CableTerminationType(NetBoxObjectType):
         filterset_class = filtersets.CableTerminationFilterSet
 
 
-class ConsolePortType(ComponentObjectType, CabledObjectMixin):
+class ConsolePortType(ComponentObjectType, CabledObjectMixin, PathEndpointMixin):
 
     class Meta:
         model = models.ConsolePort
@@ -139,7 +139,7 @@ class ConsolePortTemplateType(ComponentTemplateObjectType):
         return self.type or None
 
 
-class ConsoleServerPortType(ComponentObjectType, CabledObjectMixin):
+class ConsoleServerPortType(ComponentObjectType, CabledObjectMixin, PathEndpointMixin):
 
     class Meta:
         model = models.ConsoleServerPort
@@ -241,7 +241,7 @@ class FrontPortTemplateType(ComponentTemplateObjectType):
         filterset_class = filtersets.FrontPortTemplateFilterSet
 
 
-class InterfaceType(IPAddressesMixin, ComponentObjectType, CabledObjectMixin):
+class InterfaceType(IPAddressesMixin, ComponentObjectType, CabledObjectMixin, PathEndpointMixin):
 
     class Meta:
         model = models.Interface
@@ -354,7 +354,7 @@ class PlatformType(OrganizationalObjectType):
         filterset_class = filtersets.PlatformFilterSet
 
 
-class PowerFeedType(NetBoxObjectType, CabledObjectMixin):
+class PowerFeedType(NetBoxObjectType, CabledObjectMixin, PathEndpointMixin):
 
     class Meta:
         model = models.PowerFeed
@@ -362,7 +362,7 @@ class PowerFeedType(NetBoxObjectType, CabledObjectMixin):
         filterset_class = filtersets.PowerFeedFilterSet
 
 
-class PowerOutletType(ComponentObjectType, CabledObjectMixin):
+class PowerOutletType(ComponentObjectType, CabledObjectMixin, PathEndpointMixin):
 
     class Meta:
         model = models.PowerOutlet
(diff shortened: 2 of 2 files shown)


ryanmerolle:
@basti1508 let me know when ready and I will test.  Currently it does not look like this PR is complete.

basti1508:
@ryanmerolle What are you missing?
I've tested it successfully with all these types.

ryanmerolle:
Maybe I was half asleep last night, but I pulled your change into netbox-docker and it was not working.  I’ll test again today.

basti1508:
> Maybe I was half asleep last night, but I pulled your change into netbox-docker and it was not working. I’ll test again today.

I've just tested again against development (since I only tested against 3.4.5) and it's still working. If it keeps failing for you, it would be good to know, what you have tested :+1:

ryanmerolle:
maybe I am misunderstanding the PR, but here are my notes:

I tested in the graphiql page, and I noticed no tooltips/autocomplete  (except for `__typename`)come up for connected_endpoints.  I would expect many of the rest api fields to display:

I tested with:
```{
  device_list {
    name
    device_role {
      name
    }
    site {
      name
    }
    interfaces {
      name
      type
      connected_endpoints {
        __typename
        device
      }
    }
  }
}

```

result:
```json
{
  "errors": [
    {
      "message": "Cannot query field 'device' on type 'LinkPeerType'. Did you mean to use an inline fragment on 'ConsolePortType', 'ConsoleServerPortType', 'FrontPortType', 'InterfaceType', or 'PowerOutletType'?",
      "locations": [
        {
          "line": 15,
          "column": 9
        }
      ]
    }
  ]
}
```

if I remove the device field like below:
```
{
  device_list {
    name
    device_role {
      name
    }
    site {
      name
    }
    interfaces {
      name
      type
      connected_endpoints {
        __typename
      }
    }
  }
}

```

I get 
```json
{
  "data": {
    "device_list": [
      {
        "name": "test1",
        "device_role": {
          "name": "router"
        },
        "site": {
          "name": "test1"
        },
        "interfaces": [
          {
            "name": "Ethernet1",
            "type": "A_1000BASE_T",
            "connected_endpoints": [
              {
                "__typename": "InterfaceType"
              }
            ]
          }
        ]
      },
      {
        "name": "test2",
        "device_role": {
          "name": "router"
        },
        "site": {
          "name": "test2"
        },
        "interfaces": [
          {
            "name": "Ethernet1",
            "type": "A_1000BASE_T",
            "connected_endpoints": [
              {
                "__typename": "InterfaceType"
              }
            ]
          }
        ]
      }
    ]
  }
}
```

basti1508:
This may be since I have took this analog to link_peers. This returns LinkPeerType (the same as this funktction, since there are many different Endpoint types).

So to access it, you need a inline Fragment there.
like

```
interfaces {
  connected_endpoints {
    ... on InterfaceType {
      device {
        id
        name
      }
    }
  }
}
```

ryanmerolle:
@arthanson might have more input, but I think the user interaction should be similar to the way users interact with similar relationships now.  Most users are just becoming used to graphql as it is.  I’d expect something more to the query I provided above to work.

**TLDR** I’ll test your approach, but it does not seem consistent to the greater approach already in place.

ryanmerolle:
query:
```
{
  device_list {
    name
    interfaces {
      connected_endpoints {
        ... on InterfaceType {
          device {
            id
            name
          }
        }
      }
    }
  }
}
```

response
```json
{
  "errors": [
    {
      "message": "Cannot query field 'connected_endpoints' on type 'InterfaceType'.",
      "locations": [
        {
          "line": 5,
          "column": 7
        }
      ]
    }
  ]
}
```

kkthxbye-code:
@ryanmerolle 

Works just fine here.

Input (copy/pasted from your post):

```
{
  device_list {
    name
    interfaces {
      connected_endpoints {
        ... on InterfaceType {
          device {
            id
            name
          }
        }
      }
    }
  }
}
```

Output:

```
{
  "data": {
    "device_list": [
      {
        "name": "device1",
        "interfaces": [
          {
            "connected_endpoints": [
              {
                "device": {
                  "id": "2",
                  "name": "testdevice2"
                }
              }
            ]
          },
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": null
          }
        ]
      },
      {
        "name": "testdevice2",
        "interfaces": [
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": [
              {
                "device": {
                  "id": "1",
                  "name": "device1"
                }
              }
            ]
          },
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": null
          },
          {
            "connected_endpoints": null
          }
        ]
      }
    ]
  }
}
```

Are you sure you have the PR checked out correctly?

ryanmerolle:
Yep I checked out the users branch via netbox docker and the connected endpoints do show up as an option in graphiql.  Regardless the approach is not consistent to the current approach for other graphql relationships.

kkthxbye-code:
> Yep I checked out the users branch via netbox docker and the connected endpoints do show up as an option in graphiql.

I have a hard time seeing how the error you got could in any way show when running the code he provided.

> Regardless the approach is not consistent to the current approach for other graphql relationships.

He copied the approach for the link_peers field:

https://github.com/netbox-community/netbox/blob/561f1eadfc2386bd266b941d019d830c37c53cf3/netbox/dcim/graphql/mixins.py#L5

The approach is also used for other generic foreign key fields, such as assigned_object on IPAddress. I think it is actually used for all other generic foreign keys, so what you are suggesting would be the inconsistent option I believe.

Not saying it's a bad idea to change it, but not sure it's actually possible? When googling examples of using Union types, they all use inline fragments in the query to specify the expected return type.

basti1508:
For me connected_endpoints was the subset of all linkpeers in between. (only the last one)  So returning LinkPeerType was the obvious solution.
It would also be possible to check for the type and return the corresponding one based on the source.
But this is only possible as long as there are never multiple endpoints with different LinkPeer Types.

ryanmerolle:
I found my issue with net box-docker build for PR validation.  User stupidity.  I can confirm the PR works.

With regards to my comments on consistency, I just meant with user experience wise.  I did not realize this was used for other gfk.  To better articulate my concern, again more long term, is that the fragments feature does not as easily auto-completion within graphiql.  I do not know how you solve this at this point.  @basti1508 has successfully implemented the feature, and has followed existing approaches in netbox.

Apologies for all my confusion.  I hope I did not cause you any frustration.  Thanks for contributing!

ryanmerolle (approved):

arthanson (approved):

PR #11831, opened 2023-02-25
Fixes #11758 - replace unsafe chars in menu label

### LINKED ISSUE #11758: Plugin menu label consisting of url unsafe characters breaks unfolding menu
### NetBox version

v3.4.4

### Python version

3.10

### Steps to Reproduce

Create a plugin menu with any url unsafe character (eg. { } / \ # < > ~ | ^) in its label:
```
menu = PluginMenu(
    label='menu with /slash',
    groups=(
        ('menu item', [menu_item]),
    ),
    icon_class='mdi mdi-format-list-text'
)
```
After restarting NetBox, open the GUI at the main menu
Click on the plugin menu name

### Expected Behavior

The plugin menu opens, showing the menu items.

### Observed Behavior

The plugin menu is displayed in the side bar, but does not unfold when clicking on it.

This issue is the same as https://github.com/netbox-community/netbox/issues/10984, only with url unsafe characters instead of spaces, eg ` { } / \ # < > ~ | ^`

rganascim:
### Fixes: #11758

The proposed changes replace any kind of special characters with underscores (_). 
I chose not to use regular expressions and instead opted for a simple check to replace the special characters (although there may be other invalid characters beyond the scope of this change). If you would like me to use regular expressions instead, please let me know

### DIFF
--- a/netbox/extras/plugins/navigation.py
+++ b/netbox/extras/plugins/navigation.py
@@ -21,7 +21,18 @@ class PluginMenu:
 
     @property
     def name(self):
-        return self.label.replace(' ', '_')
+        special_chars = [
+            ' ', '*', '/', '?', '>', '<', '\\', '|', '"', ':',
+            ';', ',', '.', '#', '+', '=', '!', '@', '$', '%',
+            '^', '&', '(', ')', '[', ']', '{', '}', '`', '~'
+        ]
+        new_name = ''
+        for char in self.label:
+            if char in special_chars:
+                new_name += '_'
+            else:
+                new_name += char
+        return new_name
 
 
 class PluginMenuItem:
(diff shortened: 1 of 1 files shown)


kkthxbye-code (changes requested):
Thanks for the PR, however the approach is not really optimal. Without having test it I think just using django.utils.text.slugify should suffice instead.

For reference `slugify` does the following:

```
Converts a string to a URL slug by:
* Converting to ASCII if allow_unicode is False (the default).
* Converting to lowercase.
* Removing characters that aren’t alphanumerics, underscores, hyphens, or whitespace.
* Replacing any whitespace or repeated dashes with single dashes.
* Removing leading and trailing whitespace, dashes, and underscores
```

Not sure if the name property is used for anything else though.

jeremystretch:
Thanks @rganascim!

PR #11836, opened 2023-02-26
Change Interpreter in shebang to python3

MephistoaD:
Dear netbox community, thank you soo much for this great software :heart_eyes: 
While installing I noticed the interpreter in the shebang of this file varies from the interpreter which is suggested in your documentation. And indeed a ./generate_secret_key.py told me that python3 is required to run this file. 

Since this is no big deal I decided not to open a dedicated issue for this, I hope you allow this change anyway.


jeremystretch:
I think this makes sense. I can't think of a supported scenario where `python` would be available but `python3` would not (given we dropped support for Python 2 years ago).

Do you want to make the same update to `netbox/manage.py`?

PR #11858, opened 2023-02-28
Fixes #11470: Validation and user friendly message on invalid address query param

### LINKED ISSUE #11470: Invalid "address" query param causes server error
### NetBox version

v3.2.9 (confirmed on v3.4.2 as well)

### Python version

3.10

### Steps to Reproduce

1. Via API or GUI, do an HTTP GET on the `IPAddress` endpoint with `?address=/32` (For example: `https://demo.netbox.dev/ipam/ip-addresses/?address=/32`)

### Expected Behavior

I would expect an empty results list, or perhaps a more gracefully handled error message like `Invalid IP address`.

### Observed Behavior

An error is generated that references the raw SQL query that is attempted

```
Internal Server Error: /api/ipam/ip-addresses/

DataError at /api/ipam/ip-addresses/
invalid input syntax for type inet: "/32"
LINE 1: ..."ipam_ipaddress" WHERE "ipam_ipaddress"."address" IN ('/32')
```

Example from the demo site

```Server Error

<class 'django.db.utils.DataError'>

invalid input syntax for type inet: "/32"
LINE 1: ..."ipam_ipaddress" WHERE "ipam_ipaddress"."address" IN ('/32')
                                                                 ^
Python version: 3.8.10
NetBox version: 3.4.2
```

rmanyari:
### Fixes: #11470 

* Adds address query param validation to support valid network ranges and IP address.
* Raises user-friendly error message on invalid input.
* Additional unit tests for edge case coverage and invalid input

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -16,6 +16,7 @@ from virtualization.models import VirtualMachine, VMInterface
 from .choices import *
 from .models import *
 
+from rest_framework import serializers
 
 __all__ = (
     'AggregateFilterSet',
@@ -599,8 +600,24 @@ class IPAddressFilterSet(NetBoxModelFilterSet, TenancyFilterSet):
                 return queryset.none()
         return queryset.filter(q)
 
+    def parse_inet_addresses(self, value):
+        try:
+            parsed = []
+            for addr in value:
+                if netaddr.valid_ipv4(addr) or netaddr.valid_ipv6(addr):
+                    parsed.append(addr)
+                    continue
+                network = netaddr.IPNetwork(addr)
+                parsed.append(str(network))
+            return parsed
+        except (AddrFormatError, ValueError):
+            raise serializers.ValidationError({
+                'address': f'Invalid address {addr}. It must be a valid IPv4 or IPv6 address or network'
+            })
+
     def filter_address(self, queryset, name, value):
         try:
+            value = self.parse_inet_addresses(value)
             return queryset.filter(address__net_in=value)
         except ValidationError:
             return queryset.none()
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -601,23 +601,33 @@ class IPAddressFilterSet(NetBoxModelFilterSet, TenancyFilterSet):
         return queryset.filter(q)
 
     def parse_inet_addresses(self, value):
-        try:
-            parsed = []
-            for addr in value:
-                if netaddr.valid_ipv4(addr) or netaddr.valid_ipv6(addr):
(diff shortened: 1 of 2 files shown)


netbox/ipam/tests/test_filtersets.py
```diff
@@ -851,6 +852,32 @@ def test_filter_address(self):
...
+        params = {'address': ['10.0.0.1/255.255.255.0', '10.0.0.1/25']}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+
+        # Check for invalid input.
+        params = {'address': ['/24']}
+        with self.assertRaises(serializers.ValidationError) as cm:
+            self.filterset(params, self.queryset).qs.count()
+        self.assertRegex(cm.exception.detail['address'], r'^Invalid address.*')
+
+        params = {'address': ['10.0.0.1/255.255.555.0']}
```
jeremystretch, line 871: A comment would be helpful here. This took me _way_ too long to tell apart from line 860. :laughing: 
```suggestion
        params = {'address': ['10.0.0.1/255.255.555.0']}  # Invalid subnet mask
```

netbox/ipam/filtersets.py
```diff
@@ -599,8 +600,24 @@ def search_by_parent(self, queryset, name, value):
...
+            parsed = []
+            for addr in value:
+                if netaddr.valid_ipv4(addr) or netaddr.valid_ipv6(addr):
+                    parsed.append(addr)
+                    continue
+                network = netaddr.IPNetwork(addr)
+                parsed.append(str(network))
+            return parsed
+        except (AddrFormatError, ValueError):
+            raise serializers.ValidationError({
```
jeremystretch, line 614: We probably need to avoid raising any validation errors inside the filter. This works for API requests, but the filter is used for UI requests as well, which don't catch the exception. In the event an invalid address is found, we can just exclude it from the list of values. (If the invalid address is the only value provided, this should result in returning an empty results list.)

jeremystretch (changes requested):
Great work, thank you for tackling this!

netbox/ipam/filtersets.py
rmanyari: Got it - Will change to silently drop invalid addresses and potentially return an empty list if all the addresses provided are invalid.

netbox/ipam/tests/test_filtersets.py
rmanyari: Welp - updating and will change the invalid subnet mask to something easier to catch

rmanyari:
Addressed the comments, should be good to go.

jeremystretch:
Thanks @rmanyari!

PR #11860, opened 2023-02-28
Fixes #11812: Use object based permissions to render the right actions on tables

### LINKED ISSUE #11812: Different buttons in a list when a user is part of more than one group
### NetBox version

v3.4.5

### Feature type

### Proposed functionality

In a list, is it possible to have both "can view" and "can change" buttons when a user is a member of several groups and has different permissions depending on the constraints? At the moment, the buttons are all identical but some point to an error page when the user does not have the rights for the selected object.

### Use case

In a list, the user would have for each line the button corresponding to his rights.
Screen Shot 2023-02-23 at 12 38 36

### Database changes

_No response_

### External dependencies

_No response_

rmanyari:
### Fixes: #11812

* Use object based permissions to render the right actions on table rows. In particular when a user has mixed permissions that involve object constraints.

See #11812 for more background. After the fix, we can see that the edit button no longer shows on rows the user isn't allowed to change.

image

### DIFF
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -245,7 +245,7 @@ class ActionsColumn(tables.Column):
         user = getattr(request, 'user', AnonymousUser())
         for idx, (action, attrs) in enumerate(self.actions.items()):
             permission = f'{model._meta.app_label}.{attrs.permission}_{model._meta.model_name}'
-            if attrs.permission is None or user.has_perm(permission):
+            if attrs.permission is None or user.has_perm(permission, obj=record):
                 url = reverse(get_viewname(model, action), kwargs={'pk': record.pk})
 
                 # Render a separate button if a) only one action exists, or b) if split_actions is True


jeremystretch:
Unfortunately validating each object in this manner is trigger two additional SQL queries per object displayed (one for the `edit` permission and one for the `delete` permission), which imposes a significant performance penalty. This is a limitation of how the object permissions framework is constructed rather than this specific usage. We'll need to dig further into the permissions framework itself to determine whether this can be optimized.

rmanyari:
Interesting, let me dig a bit deeper see if I can come up with a more efficient check.

Out of curiosity, how did you realize that this change would trigger two additional queries? Any tool you'd recommend to keep tabs on these types of regressions for future changes?

jeremystretch:
I rely pretty heavily on the SQL tab of the Django debug toolbar. Makes it very easy to spot suboptimal patterns.

rmanyari:
I've been looking a bit more into detail at the code and it seems like this is pretty much the access pattern that the django auth packages enable.

One optimization that crosses my mind is the following: Instead of running an exists query on each column render call

`return model.objects.filter(qs_filter, pk=obj.pk).exists()`

We could, on the first render of the column, run a similar query but taking into account all the pks (it's doable since we have access to `table.rows` in the render function signature), cache it and use it on subsequent render calls. Once we hit the last row in the table, we clear the cache. That way we'd reduce the roundtrips to the DB, the tradeoff is obviously a more expensive query and a more complicated logic, but it might be worth it.

I can prototype something along those lines and post the git diff if this sounds reasonable.

Thoughts?

rmanyari:
Another thing to consider on this approach is how pagination is handled and how that affects the render calls. The proposed alternative feels a bit hacky but would achieve the goal of reducing DB roundtrips and calls.

rmanyari:
Alright - here's a prototype. The idea is the following:

On each render call, we add the `batch_pks` attribute to the record and then continue evaluating permissions as usual. In the `ObjectPermissionMixin`, we check whether the object has the `batch_pks` attribute, if it does, we take the `_batched_has_perm` path instead of doing the original filter DB call.

In `_batched_has_perm` we check whether we've already evaluated the permissions for the record `(permission, pk)` tuple. If it exists, we return that deleting it from our local cache. If it doesn't exist, we evaluate it along with everything else in that same batch.

This does the trick as far as reducing the number of DB roundtrips, and also works with pagination. The current downside is the increased code complexity and also the fact that the implementation is not thread-safe. I'd imagine some weirdness when running behind gunicorn.

```
diff --git a/netbox/netbox/tables/columns.py b/netbox/netbox/tables/columns.py
index 3162e80a2..e02089cd8 100644
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -228,11 +228,23 @@ class ActionsColumn(tables.Column):
     def header(self):
         return ''
 
+    def _collect_page_pks(self, table):
+        pks = []
+        for r in table.rows[table.page.start_index() - 1:table.page.end_index()]:
+            if hasattr(r.record, 'pk'):
+                pks.append(r.record.pk)
+        return pks
+
     def render(self, record, table, **kwargs):
         # Skip dummy records (e.g. available VLANs) or those with no actions
         if not getattr(record, 'pk', None) or not self.actions:
             return ''
 
+        # By setting this attribute, we allow the ObjectPermissionMixin to evaluate
+        # permissions in batches instead of making many round-trips to the DB for
+        # each render call.
+        setattr(record, 'batch_pks', self._collect_page_pks(table))
+
         model = table.Meta.model
         request = getattr(table, 'context', {}).get('request')
         url_appendix = f'?return_url={escape_uri_path(request.get_full_path())}' if request else ''
```

```
diff --git a/netbox/netbox/authentication.py b/netbox/netbox/authentication.py
index 798cb80e2..a10845f06 100644
--- a/netbox/netbox/authentication.py
+++ b/netbox/netbox/authentication.py
@@ -62,6 +62,8 @@ def get_saml_idps():
 
 class ObjectPermissionMixin:
 
+    batched_permissions = defaultdict(dict)
+
     def get_all_permissions(self, user_obj, obj=None):
         if not user_obj.is_active or user_obj.is_anonymous:
             return dict()
@@ -92,6 +94,20 @@ class ObjectPermissionMixin:
 
         return perms
 
+    def _batched_has_perm(self, model, perm, qs_filter, obj):
+        if not self.batched_permissions[perm]:
+            self.batched_permissions[perm] = dict()
+
+        if obj.pk not in self.batched_permissions[perm]:
+            for k in obj.batch_pks:
+                self.batched_permissions[perm][k] = False
+            for r in model.objects.filter(qs_filter, pk__in=obj.batch_pks):
+                self.batched_permissions[perm][k] = True
+
+        r = self.batched_permissions[perm][obj.pk]
+        del self.batched_permissions[perm][obj.pk]
+        return r
+
     def has_perm(self, user_obj, perm, obj=None):
         app_label, action, model_name = resolve_permission(perm)
 
@@ -129,6 +145,9 @@ class ObjectPermissionMixin:
         }
         qs_filter = qs_filter_from_constraints(object_permissions[perm], tokens)
 
+        if hasattr(obj, 'batch_pks'):
+            return self._batched_has_perm(model, perm, qs_filter, obj)
+
         # Permission to perform the requested action on the object depends on whether the specified object matches
         # the specified constraints. Note that this check is made against the *database* record representing the object,
         # not the instance itself.
```

@jeremystretch if this is in a direction that you like, I'm glad to dig a bit deeper and tweak the implementation.

rmanyari:
I prototyped this idea (basically doing a query with all the pks in the table instead of one query per row on each render) but it defeats the purpose because the simple action of getting the pks out of the table triggers a bunch of DB calls.

Essentially any version of this (I tried both with `table.data` and `table.rows`), from what I've seen so far, makes the problem we're trying to fix even worse:

```
    def _collect_page_pks(self, table):
        pks = []
        for r in table.data[table.page.start_index() - 1:table.page.end_index()]:
            pks.append(r.id)
        return pks
```

Unless there's a better way of getting this information without more DB calls, I don't think it makes sense to pursue this option. If anyone can chime in, it would be appreciated.

Another option I'm playing with at the moment is the following: Prefetch all the rows that match the `qs_filter` (essentially doing a filter query without the pk, caching that result for `PREFETCHED_OBJECT_PERMISSIONS_TTL_MS` (configurable through the settings) and then checking against that in `has_perm` refreshing the result based on the TTL. I'm personally not a huge fan of propagation delays in authz systems but it could also do the trick.

It goes without saying that I'm open to explore other options if others with a better understanding of the system have suggestions :)

rmanyari:
Here's a prototype of the prefetching based approach. It's a pretty simple idea, prefetch the objects that match the `qs_filter` and keep them around for some configurable amount of time. It reduces the DB roundtrips, it's all contained in the authentication file but the downside is that it makes authz propagation slower (configurable but still, not immediate like it is right now). If the cache is not enabled, it simply fallsback to the original check which does many DB roundtrips, if this is accepted, I'd recommend releasing with the cache configured to 250ms TTL.

```
diff --git a/netbox/netbox/authentication.py b/netbox/netbox/authentication.py
index 798cb80e2..e1e84816e 100644
--- a/netbox/netbox/authentication.py
+++ b/netbox/netbox/authentication.py
@@ -1,3 +1,4 @@
+import time
 import logging
 from collections import defaultdict
 
@@ -59,9 +60,77 @@ def get_auth_backend_display(name):
 def get_saml_idps():
     return getattr(settings, "SOCIAL_AUTH_SAML_ENABLED_IDPS", {}).keys()
 
+class PrefetchedObjectPermissions:
+
+    def __init__(self, model, perm, qs_filter, ttl_ms):
+        self.model = model
+        self.perm = perm
+        self.qs_filter = qs_filter
+        self.ttl_ms = ttl_ms
+        self._refresh(qs_filter)
+
+    def _time_ms(self):
+        return int(time.time() * 1000)
+
+    def _is_expired(self):
+        return self._time_ms() > self.expires_at
+
+    def _filter_changed(self, qs_filter):
+        return self.qs_filter != qs_filter
+
+    def _refresh(self, qs_filter):
+        self.qs_filter = qs_filter
+        s = set()
+        for r in self.model.objects.filter(self.qs_filter):
+            s.add(r.id)
+        self.allowed = s
+        self.expires_at = self._time_ms() + self.ttl_ms
+
+    def _has_perm(self, obj):
+        return obj.pk in self.allowed
+
+    def has_perm_or_refresh(self, qs_filter, obj):
+        if self._is_expired() or self._filter_changed(qs_filter):
+            self._refresh(qs_filter)
+        return self._has_perm(obj)
+
+
+class PrefetchedObjectPermissionsCache:
+    '''
+    Cache manager for object based permissions.
+
+    We keep a map of user.id -> perm -> PrefetchedObjectPermission instance
+    which holds the set of object IDs that permissioned to do perm action.
+
+    This cache is valid for PREFETCHED_OBJECT_PERMISSIONS_TTL_MS and refreshed
+    automatically each time it's accessed if needed.
+    '''
+
+    def __init__(self):
+        self.ttl_ms = self._prefetched_object_permissions_ttl()
+        self.m = defaultdict()
+
+    def _prefetched_object_permissions_ttl(self):
+        try:
+            return int(getattr(settings, 'PREFETCHED_OBJECT_PERMISSIONS_TTL_MS', 0))
+        except:
+            return 0
+
+    def is_enabled(self):
+        return self.ttl_ms > 0
+
+    def has_perm(self, model, perm, user, qs_filter, obj):
+        if user.id not in self.m:
+            self.m[user.id] = defaultdict()
+        user_cache = self.m[user.id]
+        if perm not in user_cache:
+            user_cache[perm] = PrefetchedObjectPermissions(model, perm, qs_filter, self.ttl_ms)
+        perm_cache = user_cache[perm]
+        return perm_cache.has_perm_or_refresh(qs_filter, obj)
 
 class ObjectPermissionMixin:
 
+    prefetched_obj_perm_cache = PrefetchedObjectPermissionsCache()
+
     def get_all_permissions(self, user_obj, obj=None):
         if not user_obj.is_active or user_obj.is_anonymous:
             return dict()
@@ -129,6 +198,9 @@ class ObjectPermissionMixin:
         }
         qs_filter = qs_filter_from_constraints(object_permissions[perm], tokens)
 
+        if self.prefetched_obj_perm_cache.is_enabled():
+            return self.prefetched_obj_perm_cache.has_perm(model, perm, user_obj, qs_filter, obj)
+
         # Permission to perform the requested action on the object depends on whether the specified object matches
         # the specified constraints. Note that this check is made against the *database* record representing the object,
         # not the instance itself.
diff --git a/netbox/netbox/settings.py b/netbox/netbox/settings.py
index 967766763..2f062c2a7 100644
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -401,6 +401,8 @@ AUTHENTICATION_BACKENDS = [
     'netbox.authentication.ObjectPermissionBackend',
 ]
 
+PREFETCHED_OBJECT_PERMISSIONS_TTL_MS = 250
+
 # Time zones
 USE_TZ = True
 ```

jeremystretch:
Hey @rmanyari, thanks for digging into this so much! I think it's clear from your experiments that we can't efficiently implement the FR without significant changes to the underlying permissions framework (which would need to be captured and discussed in a separate FR).

Given the degree of effort, I'm inclined to shelve this effort until improvements to the permissions framework can be explored.

rmanyari:
Fair enough - I'd actually be curious to understand how `django.tables2` calls `render` on each column without making DB calls for each row. I assume there's something at the table level that fetches things in bulk and if that can be extended it would maybe be the right place to also pull object permissions.

Although interesting, it certainly seems out of scope for this FR. Feel free to close it and maybe we can look into that later if there's enough people asking to fix the UI issue (edit button showing but sending you to an error page).

rmanyari:
Should I go ahead and close this for now?

jeremystretch:
Unfortunately yes, I don't think we're going to be able to implement this efficiently without further work to the permissions backend.

PR #11863, opened 2023-03-01
using cable.get_trace_url

### LINKED ISSUE #11622: Cable trace in cable view
### NetBox version

v3.4.3

### Feature type

### Proposed functionality

It would be awesome to have the possibility to click on tracing a cable while viewing the cable (e.g. https://localhost/dcim/cables/1/)

image

as well as on the list view

image

### Use case

Being in the data center, seeing a cable without knowing whats connected would involve searching for the cable ID printed on the cable label. After finding it in Netbox it would be nice to see a trace icon image in the list overview as well as in the details page of each cable to be able to directly get a trace to each end without the need of click of one connected interface to click on trace there (see example screenshots above in the proposal).

### Database changes

_No response_

### External dependencies

_No response_

PieterL75:

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -219,6 +219,12 @@ class Cable(PrimaryModel):
     def get_status_color(self):
         return LinkStatusChoices.colors.get(self.status)
 
+    def get_trace_url(self):
+        for cabletermination in self.terminations.all():
+            if cabletermination.termination_type.model in ['interface', 'rearport', 'frontport', 'consoleport', 'consoleserverport', 'powerport', 'poweroutlet']:
+                return reverse(f'dcim:{cabletermination.termination_type.model}_trace', kwargs={'pk': cabletermination.termination.pk})
+        return ''
+
 
 class CableTermination(models.Model):
     """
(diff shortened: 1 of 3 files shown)


netbox/dcim/models/cables.py
```diff
@@ -219,6 +219,12 @@ def save(self, *args, **kwargs):
     def get_status_color(self):
         return LinkStatusChoices.colors.get(self.status)
 
+    def get_trace_url(self):
+        for cabletermination in self.terminations.all():
+            if cabletermination.termination_type.model in ['interface', 'rearport', 'frontport', 'consoleport', 'consoleserverport', 'powerport', 'poweroutlet']:
```
jeremystretch, line 224: This is generating a database query against ContentType for each termination. We'll have to devise a more efficient approach to resolving the trace URL.

jeremystretch (changes requested):
I'm not sure how practical it is to provide a button that only traces the cable in one (non-deterministic) direction. That seems like it would only generate confusion.

jeremystretch:
@PieterL75 are you interested in pursuing this any further? IMO the proposed implementation is problematic for a few reasons, but I unfortunately don't have any recommendations.

netbox/dcim/models/cables.py
PieterL75: A littlebit higher in the code, there is a remark
```        # Query self.terminations.all() to leverage cached results```

So I assumed it will take the data from the cache in my call to

jeremystretch:
Closing this out as no further progress has been made.

PR #11870, opened 2023-03-01
Closes #11851: Add family field to IPAddress queries in GraphQL

### LINKED ISSUE #11851: Show IP address version in GraphQL queries
### NetBox version

v3.4.5

### Feature type

Data model extension

### Proposed functionality

I would like GraphQL queries to able to show the address family of an IP Address as the REST API currently does.

REST via https://netbox.acronis.work/api/ipam/ip-addresses/ endpoint:
```
{
  "family": {
    "value": 4,
    "label": "IPv4"
  }
}
```
GraphQL has no concept of the "family" field.

Per Jeremy on Slack:
Currently "family" is an attribute of the IPAddress model rather than an actual database field. The REST API serializer has an explicit field defined to convey this value whereas the GraphQL API serializer does not.

### Use case

When pulling data from Netbox to configure network devices, I need to be able to distinguish between address families. I could theoretically do this in the Jinja template but I prefer to minimise data manipulation at that level. 

### Database changes

I believe this would require a new field.

### External dependencies

_No response_

rmanyari:
### Closes: #11851 

* Use a graphene resolver to include the family information in GraphQL queries.

### DIFF
--- a/netbox/ipam/graphql/types.py
+++ b/netbox/ipam/graphql/types.py
@@ -27,6 +27,28 @@ __all__ = (
 )
 
 
+class IPAddressFamilyType(graphene.ObjectType):
+
+    value = graphene.Int()
+    label = graphene.String()
+
+    def __init__(self, value):
+        self.value = value
+        self.label = f'IPv{value}'
+
+
+class BaseIPAddressFamilyType:
+    '''
+    Base type for models that need to expose their IPAddress family type.
+    '''
+    family = graphene.Field(IPAddressFamilyType)
+
+    def resolve_family(self, _):
+        # Note that self, is an instance of models.IPAddress
+        # thus resolves to the address family value.
+        return IPAddressFamilyType(self.family)
+
+
 class ASNType(NetBoxObjectType):
     asn = graphene.Field(BigInt)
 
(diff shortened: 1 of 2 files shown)


netbox/ipam/graphql/types.py
```diff
@@ -64,14 +64,30 @@ class Meta:
...
 
 
+class IPAddressFamilyType(graphene.ObjectType):
+
+    value = graphene.Int()
+    label = graphene.String()
+
+    def __init__(self, value):
+        self.value = value
+        self.label = 'IPv4' if value == 4 else 'IPv6'
```
jeremystretch, line 74: Could do
```suggestion
        self.label = f'IPv{value}'
```

jeremystretch:
I like the approach. Seems like it would make sense to replicate for the Aggregate and Prefix models as well (both already provide a `family` property). Would you like to tackle that as part of this PR?

netbox/ipam/graphql/types.py
rmanyari: Nice, will update.

rmanyari:
Yep, let me take a look

netbox/utilities/testing/api.py
```diff
@@ -460,6 +462,8 @@ def _build_query(self, name, **filters):
                     # TODO: Come up with something more elegant
                     # Temporary hack to support automated testing of reverse generic relations
                     fields_string += f'{field_name} {{ id }}\n'
+                elif inspect.isclass(field.type) and issubclass(field.type, IPAddressFamilyType):
```
rmanyari, line 465: Not really happy about this one here, if there's a better way of doing this I'm happy to change it.

jeremystretch:
Thanks @rmanyari!

PR #11894, opened 2023-03-02
Closes #11294: Markdown Preview

### LINKED ISSUE #11294: Markdown input field preview option
### NetBox version

v3.3.9

### Feature type

New functionality

### Proposed functionality

Similar to Github, a web UI option to preview the formatted result of the comments/journal fields without the need to save the object.

### Use case

Currently users are modifying the same object multiple times when trying to format with markdown on device comments or journal entries.  
This causes unnecessary database updates and change log records.
A preview option to the comments field would mitigate this.

### Database changes

_No response_

### External dependencies

_No response_

aronjohanns:
### Closes: #11294

- New markdown widget that has preview button
- Render markdown endpoint in extras
- CommentField uses new markdown widget


netbox/extras/views.py
```diff
@@ -885,3 +886,24 @@ class JobResultBulkDeleteView(generic.BulkDeleteView):
...
     filterset = filtersets.JobResultFilterSet
     table = tables.JobResultTable
+
+
+#
+# Markdown
+#
+
+class MarkdownRenderView(View):
+    # _ignore_model_permissions = True
```
jeremystretch, line 896: Rather than going through the extra steps of defining and rendering a template, we can call the `render_markdown` template filter directly on the input value.

Additionally, if the value is empty, we should return an empty string (rather than the placeholder) to ensure the view is suitable for general use.

netbox/extras/urls.py
```diff
@@ -92,4 +92,6 @@
     path('scripts/results/<int:job_result_pk>/', views.ScriptResultView.as_view(), name='script_result'),
     re_path(r'^scripts/(?P<module>.([^.]+)).(?P<name>.(.+))/', views.ScriptView.as_view(), name='script'),
 
+    # Markdown
+    path('markdown/', views.MarkdownRenderView.as_view(), name="markdown_render")
```
jeremystretch, line 96: We should consider potential future use cases, where we might need to render other types of markup as well. I'd suggest creating a Markdown-specific view under a more general-purpose path:
```suggestion
    path('render/markdown/', views.MarkdownRenderView.as_view(), name="render_markdown")
```

netbox/utilities/forms/widgets.py
```diff
@@ -116,6 +118,17 @@ class SelectDurationWidget(forms.NumberInput):
     template_name = 'widgets/select_duration.html'
 
 
+class MarkdownWidget(forms.Textarea):
+    template_name = 'widgets/markdown_input.html'
+
+
+class SmallMarkdownWidget(MarkdownWidget):
```
jeremystretch, line 125: I don't think this is actually needed. The `Smalltextarea` widget was a remnant of the old UI that I've coincidentally already ripped out for v3.5. We should be able to just stick with a single `MarkdownWidget` class.

jeremystretch (changes requested):
Nice work!

netbox/utilities/forms/widgets.py
aronbj20: Should CommentFields that used SmallTextArea then just use the MarkdownWidget or should I change them back to using SmallTextArea?

jeremystretch (approved):

Issue #11291, opened 2022-12-22
Slow queries in GraphQL after upgrading to v3.4.1


glesys-andreas:
### NetBox version

v3.4.1

### Python version

3.9

### Steps to Reproduce

import/create >= 800 vlans
[{"vid": "1", "name": "vlan_1", "status": "active"} .....  {"vid": "800", "name": "vlan_800", "status": "active"}]

run query:
```
query vlan {
	vlan_list {
		prefixes {
			prefix
			description
			tenant {
				name
			}
			vlan {
				group {
					name
				}
				vid
			}
		}
	}
}
```

### Expected Behavior

Response in less than 10 seconds.

### Observed Behavior

In version 3.3 this query took less than 10s, after upgrading to 3.4 this takes between 30-40 seconds.
This is reproduced at the demo site: https://demo.netbox.dev/ipam/vlans/

arthanson:
This is more of a feature request than a bug,  The issue is we aren't doing any prefetch_related or select_related on the GraphQL queries.  I replicated the above and 2046 queries are done.  I'm not certain why it got worse from 3.3 to 3.4 but the root of the issue is the DB calls.  

The bigger issue with GraphQL is we don't know what will be included in the query, so you can either optimize for most common queries or dynamically optimize queries.  There is a package https://github.com/tfoxy/graphene-django-optimizer to dynamically optimize queries, unfortunately it doesn't work with the latest Graphene.

emil-nasso:
I think this is a standard problem in graphql APIs that is usually solved using batching/data-loaders. Basically, it's a n+1 problem/lazy loading problem. There is one query to the database to fetch all prefixes and then when it is time to load the vlan for a prefix, there is no batching implemented, so it performs lazy-loading/one query to the database per prefix.

The solution is to batch it. When trying to load vlans, instead of directly loading them, their keys are "cached" and finally a list of all prefixes that we need to load vlans for is sent to the batch-method and we can load them all at once, using one query to the database.

I have implemented this in several graphql APIs but I'm not "fluent in python or graphene". :)

It seems like graphene has a solution for it though: https://docs.graphene-python.org/en/latest/execution/dataloader

It seems like it could possible be a solution for this to use that?

neuro42:
After upgrading to 3.4.3 from 3.3.6, we've also noticed our graphql query jump from ~1s to ~7s.  Looking at the CPU usage with py-spy, it doesn't seem like the DB queries are being slow, but the django bits around them that are:
[...]

PR #11943, opened 2023-03-09
11291 optimize GraphQL queries

arthanson:
### Fixes: #11291 

This optimizes the GraphQL queries by automatically putting in prefetch_related to queried fields.  Results:

I made a bunch of VLANs and a query that shows the query issue resulting in a 10x or 20x reduction of queries, here are the query counts:
- old code (before filtering at all levels): 985 queries
- current code (with filtering at all levels): 1852 queries
- with query optimizer: 84 queries

Tested with sample queries for nested types, nested filters, fragments (triple dot notation) and they all worked successfully.

**Couple Notes on implementation:**

There is some interaction between graphene / FilterSets and the optimizer.  The optimizer works fine with graphene but without FilterSets and works fine outside of graphene with FilterSets, but combining Graphene, FilterSets and the optimizer produces extra queries.  So [netbox/netbox/graphql/fields.py](https://github.com/netbox-community/netbox/pull/11943/files#diff-453a32882215ad6bdb3c106adeeff83bcf4d2667375fe4130a8643f10be5e872) if there are no filter params then it doesn't instantiate the FilterSet.

Propose this as a temporary solution as it results in 10x less queries in the non-filtered case and figuring out how to fix the interaction between graphql / filterset will be a big time sink I think so can create a separate issue for that.

There is a separate issue where this code: https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/filtersets.py#L249 is adding in extra queries for custom-fields even if they are not referenced in the query-set.  This is what causes the number of queries to double with the filtering at all levels feature that was added in NetBox 3.4.  It isn't a bug in that feature, rather an existing bug that was just exposed by this change as it now applies filters at all levels in the schema. If the custom fields issue is fixed the querycount comes back down to 985 with filtering.

In the schmea files the lines:
```
    def resolve_vrf_list(root, info, **kwargs):
        return gql_query_optimizer(models.VRF.objects.all(), info)
```
were needed as unfortunately the optimizer code could not be put into fields.py which generically handles getting the schema class and filterset.  The reason for this is two fold:

- The optimizer has to be inserted in code in the parent list_resolver initializer after it queries the class but before it passes it down into Django.
- second: fields.py list_resolver gets called for all lists, but you only want the optimizer called for the root resolver so it needs to go into the resolver functions in the schema files.

### DIFF
--- a/netbox/circuits/graphql/schema.py
+++ b/netbox/circuits/graphql/schema.py
@@ -1,21 +1,38 @@
 import graphene
 
+from circuits import models
 from netbox.graphql.fields import ObjectField, ObjectListField
 from .types import *
+from utilities.graphql_optimizer import gql_query_optimizer
 
 
 class CircuitsQuery(graphene.ObjectType):
     circuit = ObjectField(CircuitType)
     circuit_list = ObjectListField(CircuitType)
 
+    def resolve_circuit_list(root, info, **kwargs):
+        return gql_query_optimizer(models.Circuit.objects.all(), info)
+
     circuit_termination = ObjectField(CircuitTerminationType)
     circuit_termination_list = ObjectListField(CircuitTerminationType)
+
+    def resolve_circuit_termination_list(root, info, **kwargs):
+        return gql_query_optimizer(models.CircuitTermination.objects.all(), info)
 
     circuit_type = ObjectField(CircuitTypeType)
     circuit_type_list = ObjectListField(CircuitTypeType)
 
+    def resolve_circuit_type_list(root, info, **kwargs):
+        return gql_query_optimizer(models.CircuitType.objects.all(), info)
+
     provider = ObjectField(ProviderType)
     provider_list = ObjectListField(ProviderType)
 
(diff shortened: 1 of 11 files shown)


v0tti:
I know this is just a draft so sorry if this is premature feedback. There seems to be a problem with accessing assigned objects in ip_address_list (maybe even in more cases but this is what I have run into).

Example query:
```
{
  ip_address_list(assigned_to_interface: true) {
    assigned_object {
      ... on InterfaceType {
        id
      }
      ... on VMInterfaceType {
        id
      }
    }
  }
}

```

Results in:
```
{
  "errors": [
    {
      "message": "name 'store' is not defined",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "ip_address_list"
      ]
    }
  ],
  "data": {
    "ip_address_list": null
  }
}
```

arthanson:
@v0tti not premature at all, that really helps. Looking into it now...

Issue #8137, opened 2021-12-21
Add new field to Device/VM models for out-of-band management IP address 


bitcollector1:
### NetBox version

v3.1.2

### Feature type

Data model extension

### Proposed functionality

BMC IP address is a key management item we track as a custom field, it would be great if the Device Management Box had an option for this information. 

   

### Use case

Currently I need to drill down into the interfaces to see the BMC IP address. We are creating custom fields so this information is available on the device page for easy reference, same as the primary IPv4 address. 

<img width="656" alt="Screen Shot 2021-12-21 at 10 14 00 AM" src="https://user-images.githubusercontent.com/50723251/146979456-d3c5ca46-958d-4158-8e7f-6b4a77f330ea.png">

jeremystretch:
I think what you're referring to as BMC is more generally known as an out-of-band (OOB) management IP (but correct me if I'm mistaken). NetBox has the concept of management-only interfaces (via the `mgmt_only` boolean attribute), but as you note we don't currently provide a direct link from a device/VM to this IP address.

As seen in the screenshot above, NetBox has two primary IP fields on the device and VM models: `primary_ip4` and `primary_ip6`. The generic `primary_ip` attribute included in the REST API representation of a device or VM points to one or the other depending on the value of the [`PREFER_IPV4`](https://netbox.readthedocs.io/en/stable/configuration/dynamic-settings/#prefer_ipv4) configuration parameter.

I'm not sure we'd want to fully replicate this structure for an OOB IP address, but it might make sense to introduce only a single `oob_ip` ForeignKey field pointing to the desired OOB IP address. I'm open to other opinions.

sdktr:
Seems like a great use case for 'computed fields'. A type of custom field that is rendered when accessed.
Assuming that the IP can be retrieved from the device object, by filtering on type 'mgmt_only', or a user defined tag.

jeremystretch:
> Seems like a great use case for 'computed fields'.

It would need to be a discrete field, as there could be more than one OOB IP address.

bitcollector1:
it's nothing critical, once you explained the model it seemed like it was more effort than it was worth. Thanks for consideration, love everything about NetBox. Hope you have a Merry X-Mas!!

bluikko:
Wouldn't BMC be modeled as just another Interface on a Device, perhaps following a naming convention for a uniform name?

jeremystretch:
[...]

PR #11953, opened 2023-03-10
Fixes #11934 - Clear previous primary ip assignment on edit

### LINKED ISSUE #11934: Moving primary ip address for a device to a new device generates Integrity error
### NetBox version

v3.4.5

### Python version

3.8

### Steps to Reproduce

We encountered this on our local version (3.4.2) but were able to reproduce on the demo instance (3.4.5)

1. Browse to the edit page for an ipv4 address marked as primary ip for a device
2. Change the device and interface to another device and interface, leave the "Make this the primary IP for the device/VM" box ticked 
3. Press Save

### Expected Behavior

 IP address is updated to be set as primary ip for the new device, and assigned to the new interface, and is no longer primary ip address for the old device and is no longer assigned to the old interface.

### Observed Behavior

The web server displays:

duplicate key value violates unique constraint "dcim_device_primary_ip4_id_key"
DETAIL:  Key (primary_ip4_id)=(185) already exists.

kkthxbye-code:
### Fixes: #11934

Kinda icky fix, but we are reassigning the assigned object in the clean method, so didn't see any other way that wouldn't add more queries.

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -464,6 +464,7 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
                 selected_objects[1]: "An IP address can only be assigned to a single object."
             })
         elif selected_objects:
+            self._old_parent = self.instance.assigned_object.parent_object
             self.instance.assigned_object = self.cleaned_data[selected_objects[0]]
         else:
             self.instance.assigned_object = None
@@ -483,10 +484,19 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
         if type(interface) in (Interface, VMInterface):
             parent = interface.parent_object
             if self.cleaned_data['primary_for_parent']:
+                # Clear any existing primary IP assignment
+                if old_parent := getattr(self, '_old_parent', None):
+                    if ipaddress.address.version == 4:
+                        old_parent.primary_ip4 = None
+                    else:
+                        old_parent.primary_ip6 = None
+                    old_parent.save()
+
                 if ipaddress.address.version == 4:
                     parent.primary_ip4 = ipaddress
                 else:
                     parent.primary_ip6 = ipaddress
+
                 parent.save()
             elif ipaddress.address.version == 4 and parent.primary_ip4 == ipaddress:
                 parent.primary_ip4 = None


kkthxbye-code:
This fix is actually not enough. Editing the primary IP for device1, removing the primary ip checkbox and assigning the ip to  device2 will not clear the primary IP of device1 leaving it referencing an IP not assigned to any of its interfaces.

The code is already messy imo, so not sure what the right approach is here. @jeremystretch suggested in the issue to throw validation errors instead, although I think the usability suffers a little it might be the right way to go.

jeremystretch:
It seems like we need to devise a mechanism to check for existing primary IP assignments (from all possible objects) when validating the updated IP address. I'm sure we could hack something together, but this might be an opportunity to re-think the underlying data model.

Currently, we have `primary_ip4` and `primary_ip6` ForeignKeys defined on the Device and VirtualMachine models. (In its current form, #8137 would also add an `oob_ip` ForeignKey for Device.) We could instead model this relationship via an intermediary table, mapping objects to individual IP addresses. This would allow for more robust management of primary IP assignments, however further consideration is needed.

ITJamie:
an intermediary relationships table could be very useful... it could allow for the possibility of other "service"  ip's (probably not the best term) to be assigned. 

for example some san's have multiple different ip's for difference types of management, firmware level (drac/oob), 
a hypervisor/ha manager level, then the actual "os" being having the primary_ip.

DanSheps:
> We could instead model this relationship via an intermediary table, mapping objects to individual IP addresses. This would allow for more robust management of primary IP assignments, however further consideration is needed.

I like this approach.  It would allow us to extend things such as oob_ip, primary_ip, etc.

With an intermediary table you can also allow user defined "roles" with a Choice field (For the intermediary table you have your uniqueness validate on device, role and ip family, v4 vs v6) that is extensible such that you can have an oob ip, a primary ip and perhaps users could add things like ipmi, console, whatever.

ITJamie:
combined with custom validators that would that would be very powerful for business logic

jeremystretch:
As a more thorough solution to the root issue here is expected to involve an almost certainly breaking change to the data model, I've opted for now to simply disable the reassignment of an IP address that has been designated as primary for its parent object in b64b19a3f. (This is not currently possible anyway as it triggers the exception documented above.)

PR #11956, opened 2023-03-10
Fixes #11929 - Strip whitespace from csv headers

### LINKED ISSUE #11929: Remove whitespace on header columns for CSV import
### NetBox version

v3.4.5

### Feature type

New functionality

### Proposed functionality

When performing a CSV import, whitespace on the column headers will cause that field to not import.  I propose the trimming of whitespace on each of the column headers.

### Use case

This avoids the chances of inadvertent missing data on import, which can happen for users particularly when exporting direct from spreadsheet applications and importing directly into NetBox and not carefully checking the header does not contain erroneous whitespace first in a text editor.

### Database changes

Nil.

### External dependencies

Nil.

kkthxbye-code:
### Fixes: #11929

Not sure if there's a better way of doing this. Both the headers and records vars contain the headers. Tests passes, but not sure if there's some weird edge cases.

### DIFF
--- a/netbox/utilities/forms/forms.py
+++ b/netbox/utilities/forms/forms.py
@@ -226,6 +226,10 @@ class ImportForm(BootstrapMixin, forms.Form):
         reader = csv.reader(stream)
         headers, records = parse_csv(reader)
 
+        # Strip whitespace from headers
+        headers = {k.strip(): v for k, v in headers.items()}
+        records = [{k.strip(): v for k, v in record.items()} for record in records]
+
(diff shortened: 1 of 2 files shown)


jeremystretch:
Is there any reason not to implement this a bit closer to the source, inside `parse_csv()`? That way we'd only need to clean the headers once.

jeremystretch:
I moved the calls to `strip()` under `parse_csv()` and it seems to work without issue.

PR #11958, opened 2023-03-10
Fixes #11914 - Add parameters to the SavedFilterTable as an optional field

### LINKED ISSUE #11914: Export Saved Filter does not include parameters
### NetBox version

v3.4.5

### Python version

3.10

### Steps to Reproduce

1. Browse to /extras/saved-filters with at least one saved filter defined
2. Export saved filters to CSV via "Export" > "All Data (CSV)"
3. Observe that the CSV does not contain the "parameters" field which is the core data to define/import/clone a saved filter

### Expected Behavior

We would expect that the data export contains the "parameters" JSON, as this is the core value of the saved filter (all other fields are effectively metadata/housekeeping)

### Observed Behavior

The parameters of the saved filter are not present in the export, meaning their subsequent import into NetBox would be incomplete.  Parameters is a required field for import, we would expect it to be present in the export.

kkthxbye-code:
### Fixes: #11914

Seems to display decently in the table and in csv export.

### DIFF
--- a/netbox/extras/tables/tables.py
+++ b/netbox/extras/tables/tables.py
@@ -1,3 +1,5 @@
+import json
+
 import django_tables2 as tables
 from django.conf import settings
 from django.utils.translation import gettext as _
@@ -109,6 +111,9 @@ class SavedFilterTable(NetBoxTable):
     content_types = columns.ContentTypesColumn()
     enabled = columns.BooleanColumn()
     shared = columns.BooleanColumn()
+
+    def value_parameters(self, value):
+        return json.dumps(value)
 
     class Meta(NetBoxTable.Meta):
         model = SavedFilter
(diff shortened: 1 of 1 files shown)


jeremystretch:
My only nit with the export display is that it appears to be raw Python rather than JSON (single quotes vs. double quotes):

```
Name,Content types,User,Description,Enabled,Shared,ID,Slug,Weight,Created,Last updated,Parameters
Filter1,dcim.site,admin,,True,True,2,filter1,100,2023-03-13 12:18,2023-03-13 12:18,"{'status': ['staging'], 'region_id': ['5']}"
```

IMO we should ensure that the export data is in some standardized format (ideally JSON) so that it is readily usable by the client.

kkthxbye-code:
I totally missed that it wasn't JSON, I'll look into exporting as JSON.

kkthxbye-code:
Made a change, can you check if the solution is the correct way to do it? It results in an export like this:

```
Name,Content types,User,Description,Enabled,Shared,ID,Slug,Weight,Created,Last updated,Parameters
test,dcim.device,admin,,True,True,1,test,100,2023-03-10 22:44:47,2023-03-10 22:44:47,"{""status"": [""active""], ""role_id"": [""1""], ""site_id"": [""1""]}"
```

Double quotes are the way the CSV write escapes quotes.

PR #11966, opened 2023-03-13
11955 remove csvdatafield csvfilefield

### LINKED ISSUE #11955: CSVDataField and CSVFileField are unused and should be removed
### Proposed Changes

It appears that CSVDataField and CSVFileField are not used anymore. Should probably be removed.

### Justification

Unused code should be removed.

arthanson:
### Fixes: #11955 

Removes unused CSVDataField, CSVFileField


jeremystretch:
Although these classes are no longer used, we should defer their removal to a minor release (i.e. v3.5) just in case someone is making use of them "unofficially." Could you rebase this PR to the `feature` branch?

Issue #11871, opened 2023-03-01
Add PoE Type 2 IEEE 802.3az


danner26:
### NetBox version

v3.4.5

### Feature type

Data model extension

### Proposed functionality

We have some switches that utilize PoE Type 2 IEEE 802.3az. We would like to request the PoE Type be added

### Use case

Current gen Datto Networking hardware utilizes IEEE 802.3az for PoE delivery

PR #11985, opened 2023-03-15
Removal of type2-ieee802.3az

### LINKED ISSUE #11984: Remove PoE Type type2-ieee802.3az
### Proposed Changes

Remove PoE Type `type2-ieee802.3az`

### Justification

As per my conversation with Braden McGrath, IEEE 802.3az is not a PoE standard but rather an Energy Efficient Ethernet standard and this does not belong here. The devices referenced in #11871 are actually `type2-ieee802.3at` based upon the PoE output rating

danner26:
### Fixes: #11984

As per my conversation with Braden McGrath, IEEE 802.3az is not a PoE standard but rather an Energy Efficient Ethernet standard and this does not belong here. The devices referenced in #11871 are actually `type2-ieee802.3at` based upon the PoE output rating

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -1137,7 +1137,6 @@ class InterfacePoETypeChoices(ChoiceSet):
 
     TYPE_1_8023AF = 'type1-ieee802.3af'
     TYPE_2_8023AT = 'type2-ieee802.3at'
-    TYPE_2_8023AZ = 'type2-ieee802.3az'
     TYPE_3_8023BT = 'type3-ieee802.3bt'
     TYPE_4_8023BT = 'type4-ieee802.3bt'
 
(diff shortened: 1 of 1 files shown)


ZPrimed:
👋  I'm the guy who chatted with danner26 on Slack about this 🤓 

802.3az standard = https://en.wikipedia.org/wiki/Energy-Efficient_Ethernet
This is what allows a device to set a given interface/PHY to a lower power-draw mode, in cases of shorter ethernet runs or similar. Doesn't _directly_ have anything to do with power-over-ethernet, though.

Fairly certain that the vendor's spec sheet he was working from is just... poorly written. 😉

ZPrimed (commented):
Guessing maybe your commit comment/reason should reference 802.3 `az` and not `at`...
[edit] nevermind, the comment on the actual netbox PR looks accurate, I think it is just the commit comment on your own git repo that has the wrong letters? I kinda suck at git. 😛 

The diff seems to be showing the correct lines removed though.

danner26:
Yeah it was a mistype, If you look at the PR history I renamed the PR since I mistyped that. Sorry about the confusion

PR #11996, opened 2023-03-16
Fixes #11991 - Add vdcs to InterfaceImportForm and InterfaceBulkEditForm

### LINKED ISSUE #11991: add VDC selection to Bulk Edit of interfaces and interface imports
### NetBox version

v3.4.6

### Feature type

New functionality

### Proposed functionality

It's not possible to select a VDC while bulk editing or importing interfaces. Please add a VDC selector in that form and in import model

### Use case

Mass edit of VDCs would be much easier. In cases where a vdc owns a lot interfaces, you need a lot clicks to add all interfaces to the right VDC.

### Database changes

_No response_

### External dependencies

_No response_

kkthxbye-code:
### Fixes: #11991

Add vdcs to InterfaceImportForm and InterfaceBulkEditForm

### DIFF
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -1175,6 +1175,14 @@ class InterfaceBulkEditForm(
         },
         label=_('LAG')
     )
+    vdcs = DynamicModelMultipleChoiceField(
+        queryset=VirtualDeviceContext.objects.all(),
+        required=False,
+        label='Virtual Device Contexts',
+        query_params={
+            'device_id': '$device',
+        }
+    )
     speed = forms.IntegerField(
         required=False,
         widget=SelectSpeedWidget(),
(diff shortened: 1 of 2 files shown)


kkthxbye-code:
I just realized that for the bulk import, the name cannot actually be used as the `to_field_name` value as it is not unique (two vdcs can have the same name if they are on other devices).

Not sure if there are other options than using ids instead?

jeremystretch:
> Not sure if there are other options than using ids instead?

We'll need to extend the init method of `InterfaceImportForm` to automatically set the queryset of of the `vdcs` fields as we do for several other relationships:

https://github.com/netbox-community/netbox/blob/96eb89a469b255a209d80555d487b1cdddd6062f/netbox/dcim/forms/bulk_import.py#L716-L724

kkthxbye-code:
@jeremystretch - Fixed, thanks for the pointer!

PR #12006, opened 2023-03-19
Updated _schedule_at to use local time when _interval is set

### LINKED ISSUE #11645: Custom script scheduling when only interval is set
### NetBox version

v3.4.3

### Feature type

### Proposed functionality

Auto set the current value of `_scheduled_at` when only `_interval` is set.

### Use case

From https://github.com/netbox-community/netbox/issues/11574#issuecomment-1402584632 and https://github.com/netbox-community/netbox/issues/11574#issuecomment-1413348595, it would be more user friendly if the script can run only using interval field as the scheduled field shows current datetime. 

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:
### Closes: #11645 

When recurs every is set but schedule at is left empty, local time displayed on the form will be used instead.

### DIFF
--- a/netbox/extras/forms/reports.py
+++ b/netbox/extras/forms/reports.py
@@ -25,12 +25,16 @@ class ReportForm(BootstrapMixin, forms.Form):
         help_text=_("Interval at which this report is re-run (in minutes)")
     )
 
-    def clean_schedule_at(self):
+    def clean(self):
         scheduled_time = self.cleaned_data['schedule_at']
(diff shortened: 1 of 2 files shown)


jeremystretch:
Could you replicate this change to ReportForm as well please?

abhi1693:
Is there any specific reason that `ScriptForm` has fields with `underscore` and `ReportForm` does not?

jeremystretch:
Thanks for taking this!

PR #12013, opened 2023-03-20
11617 validate csv column headers

arthanson:
### Fixes: #11617 

Refactors the bulk import code to use common validation of headers for csv, yaml and json - this is done on each line item as yaml and json have separate headers on each item.  The validation for fields that aren't in the form is skipped for (DeviceTypeImportForm, ModuleTypeImportForm) as these allow importing of related objects.

Extra tests were added to test yaml and json bulk import along with the existing csv bulk import tests, as well as test for invalid field in the import.

**Note:** #13500 will need to be merged before this to fix the failing test.


netbox/utilities/forms/forms.py
```diff
@@ -226,6 +230,13 @@ def _clean_csv(self, data):
         reader = csv.reader(stream)
         headers, records = parse_csv(reader)
 
+        if self.model_form:
+            form_fields = self.model_form().fields
+            required_fields = [
+                name for name, field in form_fields.items() if field.required
+            ]
+            validate_csv(headers, form_fields, required_fields)
```
jeremystretch, line 238: Rather than passing the model form to the parent and instantiating it separately, can we just call `validate_csv()` in the model form itself?

arthanson: @jeremystretch  Not sure if I completely follow - the model_form is called/saved per line-item of the CSV, I didn't want to do this validation per line-item as it is really validating just he header line - but the fields need to be pulled from the model_form which I think need to be instantiated to access fields.  I could move the field checking into function in CSVModelForm but would still need to call it here...  Am I missing something obvious here?

jeremystretch: IIRC validation will fail on the first item if there's an invalid header, so there shouldn't be any penalty for evaluating it per-object (though I need to double-check this).

jeremystretch:
I've rebased this against the current `develop` branch to bring it current, which unsurprisingly has broken the tests. Further review is still needed.

jeremystretch:
I'm going to close this now that #13826 has addressed #11617 specifically, however I'll keep the branch intact as it will serve as a reference when/if we extend the header validation logic to JSON & YAML imports as well.

PR #12031, opened 2023-03-22
Propagate Tags to components when adding component

jclbc:
I'm using tags on all objects (for permissions), I would like to propagate tags from object model to components.

e.g.: when adding module or device is it possible to add:
   -checkbox "propagate Tags to components"
or
   -new field "tags for components"

thank you


kkthxbye-code:
Please follow the contributer guide. We do not accept random PR's without an accepted issue.

PR #12039, opened 2023-03-22
Closes #9047 - Add Provider Accounts

### LINKED ISSUE #9047: Allow more than 1 Account per provider
### NetBox version

v3.2.0

### Feature type

New functionality

### Proposed functionality

Create an account model allowing more than one account per provider.  The model could include:
- Account Name (where to track where account numbers result due to regional entities being different IE Network Client UK vs Network Client US for only billing/AP purposes) - mandatory
- Account ID- mandatory
- Account Description ( add any additional useful context) - optional

### Use case

Sometimes providers require a different account and circuit id per region, sometimes your firm may have different billed entities per region for AP purposes and the provider cannot link the accounts all together, etc

### Database changes

_No response_

### External dependencies

_No response_

DanSheps:
### Closes: #9047

* Add ProviderAccount model and associated views
* Link Circuit to ProviderAccount model instead of provider model (supported automatically through migration)


DanSheps:
Going to re-base this tomorrow against the latest feature I think...

PR #12054, opened 2023-03-24
Closes #11623: hide Wi-Fi PSK

### LINKED ISSUE #11623: Hide WIFI password
### NetBox version

v3.4.3

### Feature type

### Proposed functionality

Is it possible to hide the WIFI password as well? As now when you add the password it is added as plain-text.
Can this be a hidden password with "*********" and icon on the side to make it visible?

Or does it needs to be moved to "secret(store) plugin?

### Use case

Well i think passwords dont need to be visible until you want it to be. 
The best part is that you ofcourse need to be logged in to see this. 
But some think this needs to be secure and im not sure if this is also plain text in the database itself? 

### Database changes

Probally yes encrypted and hidden

### External dependencies

Maybe the secret plugin.

decoupca:
### Fixes: #11623

Changes wifi PSK field to a toggle-able password field


jeremystretch (changes requested):
I don't believe this satisfies the proposal in #11623.

- It only addresses WirelessLANForm and not WirelessLinkForm
- It does nothing to obscure the display of the PSK in the object templates

decoupca:
@jeremystretch Something like this what you have in mind? also: how do you want to handle list views?

Screenshot 2023-04-05 at 11 05 37 AM

Screenshot 2023-04-05 at 11 05 46 AM

jeremystretch:
I believe that's the goal, but it's not my FR. I don't have a suggestion for handling list views I'm afraid; maybe the FR author does.

decoupca:
@deku-m can you comment on this?

deku-m:
> @deku-m can you comment on this?

Yes that what i like to see. The password is hidden and can be visible again after clicking the button.
Dont know what is meant with the list views?

decoupca:
> > @deku-m can you comment on this?
> 
> Yes that what i like to see. The password is hidden and can be visible again after clicking the button. Dont know what is meant with the list views?

When you add Pre-shared key to the columns in list view, you see the plaintext PSK:

image

Does your FR propose/request a way of obfuscating PSKs in that context too?

deku-m:
> > > @deku-m can you comment on this?
> > 
> > 
> > Yes that what i like to see. The password is hidden and can be visible again after clicking the button. Dont know what is meant with the list views?
> 
> When you add Pre-shared key to the columns in list view, you see the plaintext PSK:
> 
> image
> 
> Does your FR propose/request a way of obfuscating PSKs in that context too?

I would say keep it equal as possible by hiding it. As if using the listview you also immediatly see all the PSKs from all SSIDs but you need to add the column first to see it. But for export its very handy to have it visible. So bit of double feeling. Or it is visible during export but not on the list?

decoupca:
@deku-m by default, PSK is not visible in the list view. If someone explicitly adds PSKs to the list view, I think we should consider that equivalent to un-hiding them. Also, it might be difficult to obfuscate PSKs in list view due to the way tables are rendered.

deku-m:
> @deku-m by default, PSK is not visible in the list view. If someone explicitly adds PSKs to the list view, I think we should consider that equivalent to un-hiding them. Also, it might be difficult to obfuscate PSKs in list view due to the way tables are rendered.

Yes true, I agree on this.

decoupca:
Sounds good. I'll leave the list view unmodified and only use obfuscation on the object/edit views. @jeremystretch does this sound appropriate?

decoupca:
Mucked up this PR with an accidental merge. Re-opened with requested changes here: https://github.com/netbox-community/netbox/pull/12244

PR #12075, opened 2023-03-28
Closes #12068: Establish a direct relationship from jobs to objects

### LINKED ISSUE #12068: Establish a concrete relationship from JobResult to related objects
### NetBox version

v3.4.6

### Feature type

Data model extension

### Proposed functionality

The JobResult model current has a ForeignKey field to ContentType, which enables associating an instance with a particular type of NetBox model. It also has a `name` field to indicate the object's name, however this is not a concrete relationship in the database. This is due to the limitation of scripts and reports not being tracked as database objects.

This FR proposes the additional of an object ID field to complement the object type field, which will enable GenericForeignKey assignments for JobResults. This is possible thanks to the implementation of #11890, which introduces "real" database models representing reports and scripts.

The following changes are proposed for the JobResult model:

1. Rename `obj_type` to `object_type` for consistency with other NetBox models using GFK.
2. Add an `object_id` integer field.
3. Add a GenericForeignKey field leveraging both of the above. Object types will be limited to those models inheriting from `JobResultsMixin`.

We'll keep the `name` field as an optional field: It can be used to differentiate among jobs pertaining to the same object. For example, we can expect a ScriptModule containing multiple scripts to have separate jobs for each script.

### Use case

Introducing a database relationship between a job and its assigned object enables more robust and efficient correlation and management of jobs. For example, we'll be able to pull all jobs pertaining to a specific script via it's module:

```
ScriptModule.jobs.filter(name='myscript1')
```

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #12068

- Establish a GFK relationship on Job (`object_type`, `object_id`)
- Rewrite the `enqueue_job()` method (now `enqueue()`) to use instances
- Rewrite report & script workflows to work with jobs assigned to module objects
- Add "jobs" views for data sources, scripts, and reports

### DIFF
--- a/netbox/core/api/serializers.py
+++ b/netbox/core/api/serializers.py
@@ -67,6 +67,6 @@ class JobSerializer(BaseModelSerializer):
     class Meta:
         model = Job
         fields = [
-            'id', 'url', 'display', 'status', 'created', 'scheduled', 'interval', 'started', 'completed', 'name',
-            'object_type', 'user', 'data', 'job_id',
+            'id', 'url', 'display', 'object_type', 'object_id', 'name', 'status', 'created', 'scheduled', 'interval',
+            'started', 'completed', 'user', 'data', 'job_id',
         ]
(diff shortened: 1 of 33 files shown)


minitriga:
OOOO I think this could be very useful for me 👍

PR #12090, opened 2023-03-29
 12084 saved filters

### LINKED ISSUE #12084: Server Error when saving Filters since 3.4.7
### NetBox version

v.3.4.7

### Python version

3.9

### Steps to Reproduce

Since v3.4.7 it is not possible to save Filters.

You can reproduce it on https://demo.netbox.dev as well:
- Select Sites
- Filter for specific tag, for example
- try to save this Filter
- Server Error is displayed

For example from the netbox Demo:
https://demo.netbox.dev/extras/saved-filters/add/?content_types=50&parameters=tag%3Dbravo
https://demo.netbox.dev/extras/saved-filters/add/?content_types=44&parameters=region_id%3D5

### Expected Behavior

Filter should be saved like in previous versions

### Observed Behavior

```Server Error

<class 'json.decoder.JSONDecodeError'>

Expecting value: line 1 column 1 (char 0)

Python version: 3.9.14
NetBox version: 3.4.7
```

arthanson:
### Fixes: #12084 

Reverts back to saving saved filters as dict, adds test case in for saved filter form save.

### DIFF
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -2,6 +2,7 @@ import json
 
 from django import forms
 from django.contrib.contenttypes.models import ContentType
+from django.http import QueryDict
 from django.utils.translation import gettext as _
 
 from dcim.models import DeviceRole, DeviceType, Location, Platform, Region, Site, SiteGroup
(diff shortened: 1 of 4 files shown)


netbox/extras/forms/model_forms.py
```diff
@@ -129,10 +130,10 @@ class Meta:
 
     def __init__(self, *args, initial=None, **kwargs):
 
-        # Convert any parameters delivered via initial data to JSON data
+        # Convert any parameters delivered via initial data to dict
```
jeremystretch, line 133: This needs to accept JSON data in order for cloning to work. Rather than changing the form back to taking a QueryDict, we should change the originating link to send JSON.

PR #12092, opened 2023-03-29
11746 fix delete custom field

### LINKED ISSUE #11746: Custom Field still exists after deleted
### NetBox version

v3.4.4

### Python version

3.9

### Steps to Reproduce

1. Create "Custom Field" dcim.device with name "123"
2. Create "Custom Field" dcim.device with name "aaa"
3. Update and add some value on custom field "aaa"
4. Remove custom field "123"
5. Reupdate custom field "aaa", and showing error, like https://demo.netbox.dev/dcim/devices/148/edit/
image
6. custom field has been deleted
image

### Expected Behavior

- Custom field deleted and not pop up error

### Observed Behavior

- I just check documentation, its correct just "should consist only of alphanumeric characters and underscores" , but as human error suddenly we can create numeric as customfield name
image

- Maybe adding checking to reject numeric on customfield name
- And how to force remove custom_field "123", causing error every update value ?

Thank youu

arthanson:
### Fixes: #11746 

Cleans up a deleted custom field.  Previous code would check if isnull, but this would leave empty custom-fields in the database and cause the issue on save, new code removes if it exists.

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -215,7 +215,7 @@ class CustomField(CloningMixin, ExportTemplatesMixin, WebhooksMixin, ChangeLogge
         """
         for ct in content_types:
             model = ct.model_class()
-            instances = model.objects.filter(**{f'custom_field_data__{self.name}__isnull': False})
+            instances = model.objects.exclude(**{f'custom_field_data__contains': self.name})
             for instance in instances:
                 del instance.custom_field_data[self.name]
             model.objects.bulk_update(instances, ['custom_field_data'], batch_size=100)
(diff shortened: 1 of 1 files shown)


netbox/extras/models/customfields.py
```diff
@@ -215,7 +215,7 @@ def remove_stale_data(self, content_types):
         """
         for ct in content_types:
             model = ct.model_class()
-            instances = model.objects.filter(**{f'custom_field_data__{self.name}__isnull': False})
+            instances = model.objects.exclude(**{f'custom_field_data__contains': self.name})
```
jeremystretch, line 218: I think we want to look up by the presence of the field's name as a key.
```suggestion
            instances = model.objects.filter(custom_field_data__has_key=self.name)
```

arthanson: @jeremystretch  yeah - that would be better. fixed.

PR #12130, opened 2023-03-31
12117 remove clone from cable

### LINKED ISSUE #12117: Clone cable causes server error 'str' object has no attribute 'label'
### NetBox version

v3.4.7

### Python version

3.8

### Steps to Reproduce

1. view a cable.
2. On cable view pane, click [clone] button.
3. Server error reported:

Server Error

<class 'AttributeError'>

Python version: 3.8.10
NetBox version: 3.4.7

### Expected Behavior

Expected to see a new cable page.

### Observed Behavior

Page displayed:

Server Error

<class 'AttributeError'>

Python version: 3.8.10
NetBox version: 3.4.7

arthanson:
### Fixes: #12117 

Removes clone from cable detail page until we allow adding cables from the list page.  Added in template overrides for all the buttons on the generic_object page just in-case some page might need to disable or change them.


jeremystretch:
Instead of solving this at the template layer, what about tweaking the button itself to only render a link when at least one clone parameter is presented?

https://github.com/netbox-community/netbox/blob/3264636b7a3cec47646abda1e8d9fb5f3dbd6176/netbox/utilities/templatetags/buttons.py#L16-L26

We can modify the tag and its template (`buttons/clone.html`) to skip the button if `prepare_cloned_fields()` returns an empty QueryDict.

arthanson:
@jeremystretch changed.

jeremystretch (approved):

PR #12150, opened 2023-04-02
Fixed interface mode hx query

abhi1693:


jeremystretch:
Thanks for this @abhi1693. I think we're going to need a more robust approach to handling HTMX form regeneration in general, to ensure that this sort of issue is not overlooked again. I've already begun work on new widget classes that will simplify the implementation of this feature.

abhi1693:
It did sort of seem hacky to me as well when doing the change.

Issue #12148, opened 2023-04-01
Netbox API doc pages should not depend on external ressouces (JS, CSS)


amhn:
### NetBox version

v3.5-beta

### Feature type

Change to existing functionality

### Proposed functionality

This only affects the current beta with the openapi spec. The swagger spec in 3.4.7 is served completely by the netbox instance.

The needed resources to access the API documentation pages should (at least optionally) be added as dependency.

This requires a new dependency drf-spectacular-sidecar.

Settings that need to be added:
```python
SPECTACULAR_SETTINGS = {
    [...]
    'SWAGGER_UI_DIST': 'SIDECAR',
    'SWAGGER_UI_FAVICON_HREF': 'SIDECAR',
    'REDOC_DIST': 'SIDECAR', 
    [...]
```

See https://drf-spectacular.readthedocs.io/en/latest/faq.html?highlight=ENUM_NAME_OVERRIDES#my-swagger-ui-and-or-redoc-page-is-blank

### Use case

In a scenario where netbox is running in an isolated network it should still be possible to access the API docs, e.g. disaster recovery with non working internet access.

### Database changes

None

### External dependencies

drf-spectacular-sidecar

PR #12159, opened 2023-04-03
12148 add swagger-sidecar for self-hosted swagger docs

arthanson:
### Fixes: #12148 

Adds drf-spectacular-sidecar to allow self-hosted swagger docs (see bug report).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,8 @@ django-tables2==2.5.3
 django-taggit==3.1.0
 django-timezone-field==5.0
 djangorestframework==3.14.0
-drf-spectacular==0.25.1
+drf-spectacular==0.26.1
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -15,7 +15,8 @@ django-tables2==2.5.3
 django-taggit==3.1.0
 django-timezone-field==5.0
 djangorestframework==3.14.0
-drf-spectacular==0.25.1
+drf-spectacular==0.26.1
+drf-spectacular-sidecar==2023.4.1
```
jeremystretch, line 19: We also need to add `drf-spectacular-sidecar` to `base_requirements.txt` with a URL and a brief description, to ensure it gets updated for future releases.

arthanson: @jeremystretch done.

Issue #12149, opened 2023-04-01
Collisions of enum fields in openapi spec


amhn:
### NetBox version

v3.5-beta

### Python version

3.10

### Steps to Reproduce

Run `./manage.py spectacular --file ../../terraform/go-netbox-openapi/schema.json --validate --format openapi-json` with the beta version.

### Expected Behavior

Names for enums are predictable and meaningful.

### Observed Behavior

Names like Type666Enum, Type2fcEnum, Type6e5Enum, Type653Enum and Type209Enum are chosen for some of the ChoiceFields named Type.

```
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type666Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type2fcEnum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type6e5Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type653Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type209Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
Warning: enum naming encountered a non-optimally resolvable collision for fields named "type". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Type261Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming.
```

amhn:
To resolve this this the enums shoud be added to ENUM_NAME_OVERRIDES in SPECTACULAR_SETTINGS. Preferably before the 3.5 release to prevent a breaking change after the release.

The following settings should be set. Names are of course only suggestions. For the 3 enums with XX in the name I am not sure what they should be named.
```
SPECTACULAR_SETTINGS = {
    [...]
    'ENUM_NAME_OVERRIDES': {
[...]

PR #12161, opened 2023-04-03
12149 fix enum collision for OpenApi swagger

arthanson:
### Fixes: #12149 

Fix enum collision for OpenAPI swagger


jeremystretch:
Per comments on #12149, I'm afraid this approach isn't tenable, for a few reasons:

1. It seems that every (?) choice field would need a static enum mapping.
2. The mapping definition lives apart from any related objects (even outside the relevant app).
3. This approach precludes a plugin from defining its own mappings.

I'm going to close this PR until a suitable alternative can be identified.

PR #12171, opened 2023-04-04
12011 fix module bay bulk create

### LINKED ISSUE #12011: KeyError when adding Module Bays to devices
### NetBox version

3.4.6

### Python version

3.10

### Steps to Reproduce

1. Add device
2. Goto devices overview 
3. mark new device
4. click on "Add Components"
5. click on "Module Bays"
6. Fill in "Name"
7. click "Create"

### Expected Behavior

New Module Bay should be added.

### Observed Behavior

Error Message:
Server Error

class 'KeyError'

'position'

Python version: 3.10.6
NetBox version: 3.4.6

arthanson:
### Fixes: #12011 

Fixes bulk adding module bays.
@jeremystretch I couldn't see why it would be named position_pattern instead of just position, the other expandable fields (name, label) are also referenced without a _pattern.  Just wanting to double-check I'm not missing something here.

### DIFF
--- a/netbox/dcim/forms/bulk_create.py
+++ b/netbox/dcim/forms/bulk_create.py
@@ -103,9 +103,9 @@ class RearPortBulkCreateForm(
 
 class ModuleBayBulkCreateForm(DeviceBulkAddComponentForm):
     model = ModuleBay
-    field_order = ('name', 'label', 'position_pattern', 'description', 'tags')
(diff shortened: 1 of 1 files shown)


jeremystretch:
>  Just wanting to double-check I'm not missing something here.

No, I think it really is just that simple. LGTM

PR #12206, opened 2023-04-07
Added optional user and group on custom field

### LINKED ISSUE #10600: Allow custom object fields to reference a user or group
### NetBox version

v3.3.5

### Feature type

Data model extension

### Proposed functionality

Being able to create object type custom fields with Netbox users.

### Use case

Being able to point a specific user which is responsible for a given object, like a device or VM.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -1,6 +1,7 @@
 import json
 
 from django import forms
+from django.db.models import Q
 from django.contrib.contenttypes.models import ContentType
 from django.utils.translation import gettext as _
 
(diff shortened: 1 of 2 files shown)


kkthxbye-code:
I don't think this is what the FR was suggesting. They just wanted the User and Group model being available as content types for object type custom fields.

abhi1693:
Since there are no views for user or group, where would this be displayed in the UI?

kkthxbye-code:
As a string would be my best guess, that's how we do for rack reservations as Jeremy mentioned in the FR.

abhi1693:
I thought he was referring to the literal implementation of the user field on the rack reservation. Basically, the ability to assign the reservation to a user iteself. I can look into the other implementation as well

kkthxbye-code:
> I thought he was referring to the literal implementation of the user field on the rack reservation.

I'm not sure what the purpose of a user and group field on the CustomField model would be. Re-reading the proposed functionality and use-case of the FR again, I can't see any way to interpret it other than being able to choose User/Group for object type custom fields.

PR #12222, opened 2023-04-11
7503 do device validate-create in serial

### LINKED ISSUE #7503: API Bulk-Create of Devices does not check Rack-Space in its own context
### NetBox version

v3.0.6

### Python version

3.9

### Steps to Reproduce

1. Create Rack with 42U
2. Create Device-Type with Height of 2U
3. POST /dcim/devices/:
```json
[
  {
    "device_role": 1,
    "device_type": <your_id>,
    "site": 1,
    "rack": <your_id>,
    "face": "front",
    "position": 1
  },
  {
    "device_role": 1,
    "device_type": <your_id>,
    "site": 1,
    "rack": <your_id>,
    "face": "front",
    "position": 2
  }
]
```

### Expected Behavior

A ValidationError with >400 HTTP Status, as both Devices overlap each other in the Rack.

### Observed Behavior

API returns 201 HTTP Status and both Devices are there.
Further Handling other than deleting them or moving them to another position is not possible.
This also breaks SVG-rendering of Racks:
```python
  File "/opt/netbox/netbox/dcim/api/views.py", line 218, in elevation
    drawing = rack.get_elevation_svg(
  File "/opt/netbox/netbox/dcim/models/racks.py", line 392, in get_elevation_svg
    return elevation.render(face, unit_width, unit_height, legend_width)
  File "/opt/netbox/netbox/dcim/svg.py", line 192, in render
    for unit in self.merge_elevations(face):
  File "/opt/netbox/netbox/dcim/svg.py", line 156, in merge_elevations
    elevation = self.rack.get_rack_units(face=face, expand_devices=False)
  File "/opt/netbox/netbox/dcim/models/racks.py", line 313, in get_rack_units
    elevation[device.position]['device'] = device
KeyError: 2
```

This Issue does not occur on API POST /dcim/devices/ with only one Device.
This Issue does not occur in Frontend-Import-View for Devices.
In both these circumstances the transaction is rolled back due to insufficient Space.

arthanson:
### Fixes: #7503 

Makes Device ViewSet to object validation and creation serial, instead of validating all at once then creating all at once - to catch conditions like overlapping rack locations.

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -416,7 +416,9 @@ class DeviceViewSet(ConfigContextQuerySetMixin, ConfigTemplateRenderMixin, NetBo
 
     def create(self, request, *args, **kwargs):
         # do validate / create for each item in serial instead of validating all data at once
-        for data in request.data:
+        data_list = request.data if isinstance(request.data, list) else [request.data, ]
+
+        for data in data_list:
             serializer = self.get_serializer(data=data)
             serializer.is_valid(raise_exception=True)
             self.perform_create(serializer)
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -1,12 +1,13 @@
 from django.http import Http404, HttpResponse
 from django.shortcuts import get_object_or_404
-from drf_spectacular.utils import extend_schema, extend_schema_view, OpenApiParameter
 from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import extend_schema, OpenApiParameter
+from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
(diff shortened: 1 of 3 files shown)


netbox/dcim/api/views.py
```diff
@@ -414,6 +415,25 @@ def get_serializer_class(self):
 
         return serializers.DeviceWithConfigContextSerializer
 
+    @transaction.atomic
+    def create(self, request, *args, **kwargs):
```
jeremystretch, line 419: IMO we should move this to a mixin class in case we need it for other views in the future.

jeremystretch:
I've moved the `create()` method into a separate mixin class to enable its future re-use, but kept it applied only to the device viewset for now. @arthanson can you give this a quick sanity check please to make sure I didn't miss anything?

arthanson:
@jeremystretch looks good.

PR #12231, opened 2023-04-12
Fixes 12040 incorrect tab displayed after failed file bulk import

### LINKED ISSUE #12040: Import of uploaded file validation failure does not display selected tab
### NetBox version

v3.4.6

### Python version

3.8

### Steps to Reproduce

1. Navigate to Organization > Sites > Import
2. Select the "Upload File" tab
3. Upload a CSV file that fails validation

### Expected Behavior

The relevant error message(s) should be displayed and the user returned to the file upload view.

### Observed Behavior

The error message(s) are displayed, however the reloaded page shows the default "Data Import" tab rather than the original "File Upload" form. The user must select the proper tab.

decoupca:
### Fixes: #12040

Bulk import tabs will now be preserved across window refreshes.

### DIFF
--- a/netbox/templates/generic/bulk_import.html
+++ b/netbox/templates/generic/bulk_import.html
@@ -15,12 +15,12 @@ Context:
 {% block tabs %}
   <ul class="nav nav-tabs px-3">
     <li class="nav-item" role="presentation">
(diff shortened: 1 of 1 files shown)


jeremystretch:
Nice and clean, thank you!

PR #12244, opened 2023-04-13
Fixes #11623: obfuscate Wi-Fi PSKs

### LINKED ISSUE #11623: Hide WIFI password

decoupca:
### Fixes: #11623

* Changes PSK fields to `PasswordInput`s in edit views
* Obfuscates PSK on object view with show/hide toggle


decoupca:
@jeremystretch I'm not clear on what I need to push to pass CI checks, please advise

abhi1693:
Try running the following and commit. You are basically missing generated static files

```
cd netbox/project-static
yarn
yarn bundle
```

decoupca:
> Try running the following and commit. You are basically missing generated static files
> 
> ```
> cd netbox/project-static
> yarn
> yarn bundle
> ```

When I do that and run `git status`:

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   dist/config.js.map
        modified:   dist/graphiql.js.map
        modified:   dist/lldp.js.map
        modified:   dist/netbox.js
        modified:   dist/netbox.js.map
        modified:   dist/status.js.map
        modified:   package.json
        modified:   src/racks.ts
        modified:   src/search.ts
        modified:   src/util.ts
        modified:   styles/overrides.scss
        modified:   styles/rack-elevation.scss
        modified:   yarn.lock

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        .yarn/
        .yarnrc.yml
```

If I do `git add .` then `git commit` it adds a bunch of stuff, tens of megabytes of what looks like yarn build environment stuff. Do I include all that?

kkthxbye-code:
@decoupca - these are the ones you are supposed to push:

        modified:   dist/netbox.js
        modified:   dist/netbox.js.map

However as you've modified package.json it will probably still fail. The .yarn folder and .yarnrc.yml files are probably because you have installed global yarn packages, also not something you should push.

You can revert all that, delete node_modules or whatever it's called and just simply do a yarn and yarn bundle and push the two files I specified.

decoupca:
@kkthxbye-code I started from a clean working tree, reran yarn and yarn bundle, ended up with the same git status as before. But pushed only your 2 files, all good. Not sure where all the other changes came from.

jeremystretch:
Thanks @decoupca!

PR #12253, opened 2023-04-14
Closes #12226: Add Profile Data Headers to Remote Authentication Middleware

### LINKED ISSUE #12226: Allow Full Name to be set by remote authentication
### NetBox version

v3.4.7

### Feature type

### Proposed functionality

Add a variable `REMOTE_AUTH_FULL_NAME` or similar which would define a header that provides the user's full name. This would then update the Full Name field in the user's profile.

### Use case

Netbox displays the full name in various places since https://github.com/netbox-community/netbox/issues/8556 and https://github.com/netbox-community/netbox/issues/8894, so it would be good to be able to automatically populate that field.

### Database changes

n/a

### External dependencies

n/a

stuntguy3000:
### Fixes: #12226

This change introduces a new set of optional headers which update the user's profile when using Remote Authentication.

The original issue references specifically first names, however this change looks at the whole user's profile.


heapdavid:
Quick note from the original request, it was Full Name rather than First Name - are they separate fields in the db? Most auth providers send the full name as one field.

stuntguy3000:
That's my understanding of the data structure. Perhaps we could add a full name field and transpose it? But that the leaves it up to interpretation of how to split a name.

kkthxbye-code:
> it was Full Name rather than First Name - are they separate fields in the db?

https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.models.User.first_name

They are separate fields.

heapdavid:
Ah yeah that looks to be the case. That's a tough one isn't it - so many different formats of name!

heapdavid:
Apologies, it looks like we can get first and last name fields from our auth providers separately, so should be fine to proceed as is.

netbox/netbox/middleware.py
```diff
@@ -87,7 +87,17 @@ def process_request(self, request):
...
-            # User is valid.  Set request.user and persist user in the session
+            # User is valid.
+            # Update the User's Profile if set by request headers
+            if settings.REMOTE_AUTH_USER_FIRST_NAME in request.META:
+                user.first_name = request.META[settings.REMOTE_AUTH_USER_FIRST_NAME]
+            if settings.REMOTE_AUTH_USER_LAST_NAME in request.META:
+                user.last_name = request.META[settings.REMOTE_AUTH_USER_LAST_NAME]
+            if settings.REMOTE_AUTH_USER_EMAIL in request.META:
+                user.email = request.META[settings.REMOTE_AUTH_USER_EMAIL]
+            user.save()
```
jeremystretch, line 98: We want to avoid updating the user instance unless there's a reason to so; otherwise we'd be forcing an update for each request.

jeremystretch (approved):

jeremystretch:
Thanks @stuntguy3000!

PR #12262, opened 2023-04-14
Closes #12129: Enable automatic synchronization of objects when DataFiles are updated

### LINKED ISSUE #12129: Add the option to automatically push changes to objects when synchronizing data sources
### NetBox version

v3.5-beta1

### Feature type

### Proposed functionality

#11558 introduced the ability to populate certain object data from remote data sources. Currently, this is a two-step process:

1. Run a synchronization job to create/update local DataFile objects from the source.
2. Individually sync NetBox objects to reflect the new data in those DataFiles.

This separation of tasks is intentional, and ensures a decoupling of synchronization jobs (which are expected to be automated) from effecting a change to production data. However, there may be cases where it's desirable to effect changes immediately upon synchronization with the remote source.

This FR proposes the addition of an `auto_sync` boolean field to the `SyncedDataMixin` class. Any object with `auto_sync` set to true shall have its `sync_data()` method called automatically upon its associated DataFile being updated.

### Use case

This provides an alternative direct coupling approach to employing synchronized object data, while preserving the existing loose coupling as the default option.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #12129

- Introduce the `AutoSyncRecord` model to cache objects that need to be auto-synced
- Add the `auto_sync_enabled` boolean field to `SyncedDataMixin`
- Extend `save()` on `SyncedDataMixin` to automatically create/delete auto-sync records
- Create `auto_sync()` receiver for the `post_sync` signal
- Add a `save` kwarg to `sync()` on `SyncedDataMixin` (for convenience)

### DIFF
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -17,5 +17,5 @@ def auto_sync(instance, **kwargs):
     """
     from .models import AutoSyncRecord
 
-    for autosync in AutoSyncRecord.objects.filter(datafile__source=instance):
+    for autosync in AutoSyncRecord.objects.filter(datafile__source=instance).prefetch_related('object'):
         autosync.object.sync(save=True)
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -1,4 +1,4 @@
-import django.dispatch
+from django.dispatch import Signal, receiver
 
 __all__ = (
     'post_sync',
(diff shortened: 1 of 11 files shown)


jeremystretch:
I considered adding a GenericForeignKey to DataFile to track the related object, but that would preclude us from ever having two objects referencing the same DataFile. (I'm not sure if that's a probable scenario, but didn't want to rule it out.)

netbox/core/signals.py
```diff
@@ -1,10 +1,21 @@
...
+
+
+@receiver(post_sync)
+def auto_sync(instance, **kwargs):
+    """
+    Automatically synchronize any DataFiles with AutoSyncRecords after synchronizing a DataSource.
+    """
+    from .models import AutoSyncRecord
+
+    for autosync in AutoSyncRecord.objects.filter(datafile__source=instance):
```
arthanson, line 20: I think you can prefetch_related('object') here?

netbox/netbox/models/features.py
```diff
@@ -421,13 +449,15 @@ def resolve_data_file(self):
             except DataFile.DoesNotExist:
                 pass
 
-    def sync(self):
+    def sync(self, save=False):
```
arthanson, line 452: document the save param?

PR #12263, opened 2023-04-15
Adds utilization on vlan groups

### LINKED ISSUE #11738: Provide utilization on vlan groups
### NetBox version

v3.4.4

### Feature type

New functionality

### Proposed functionality

Add `utlization` property which adds to the following views

- Object view
- Table column
- API

This similar to `prefixes` and `aggregates` functionality.

### Use case

As an engineer, I'd like to see at a glance how many VLANs are utilized in a group. Using this information via the API, I'd also like to perform certain maintenance tasks in my automation workflow.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:


netbox/ipam/models/vlans.py
```diff
@@ -114,6 +114,15 @@ def get_next_available_vid(self):
             return available_vids[0]
         return None
 
+    def get_utilization(self):
+        """
+        Return the percentage of utilization for this vlan group.
+        """
+        assigned_vlan = VLAN.objects.filter(group=self.id).count()
```
jeremystretch, line 121: Just to reiterate what we touched on in today's maintainer meeting: We can't call `count()` on each instance as it will result in a database query for each object being returned as part of an API response. Instead, we can annotate the queryset under the view to include the count of assigned VLANs per VLAN range.

abhi1693: I have made the changes as per the suggestion

netbox/ipam/api/views.py
```diff
@@ -146,7 +148,8 @@ class FHRPGroupAssignmentViewSet(NetBoxModelViewSet):
 
 class VLANGroupViewSet(NetBoxModelViewSet):
     queryset = VLANGroup.objects.annotate(
-        vlan_count=count_related(VLAN, 'group')
+        vlan_count=count_related(VLAN, 'group'),
+        utilization=Round(F('vlan_count') / (F('max_vid') - F('min_vid') + 1.0) * 100, 2)
```
jeremystretch, line 152: The annotation logic should be moved to a queryset method to avoid duplication. Also, `Round()` doesn't seem to be applied consistently.

jeremystretch (changes requested):

jeremystretch:
Resolved the conflicts noted above but ended up pushing to a local branch; not sure why. Opened PR #13108 to replicate this one.

Issue #12272, opened 2023-04-17
Use dulwich as Git client in GitBackend


tobiasge:
### Proposed Changes

Use the Python Git library [dulwich](https://pypi.org/project/dulwich/) in the GitBackend class instead of invoking the external Git binary.

### Justification

- More control over the dependency as Git is usually installed by the OS
- Smaller image size with Netbox Docker image

tobiasge:
The needed changes can be found here: https://github.com/tobiasge/netbox/commit/ff6e057edd23452ca898c85b8ad3ef1c1b291ad5

jeremystretch:
Makes sense to me. Dulwich seems like a mature product, and I'd love to obviate use of the binary client. @tobiasge would you like to put in a PR with your proposed changes?

PR #12274, opened 2023-04-17
Use dulwich as Git client

tobiasge:
### Fixes: #12272

Adds dulwich as dependency and uses it in GitBackend.


jeremystretch:
Thanks @tobiasge!

Issue #12278, opened 2023-04-17
Warning on schema generation for IPAddressField


arthanson:
### Proposed Changes

Should silence warning for IPAddressField on drf-spectacular openAPI 3.

Generate schema using, e.g.

```
./netbox/manage.py spectacular --color --file schema.json --validate --format openapi-json
```

Warning [IPRangeViewSet > IPRangeSerializer]: model field "IPAddressField" has no mapping in ModelSerializer. It may be a deprecated field. Defaulting to "string"

### Justification

There should be no warnings produced by default if possible.

PR #12285, opened 2023-04-18
12278 add ipaddressfield serializer for OpenAPI spectacular typing

arthanson:
### Fixes: #12278 

Creates a serializer for the IPAddressField so typing works correctly for OpenAPI eliminating warnings over the type (the default schema overrides weren't working for this).  It's defined in nestedserializers due to scoping (serializers include nestedserializers).


arthanson:
@jeremystretch this should be ready, moved IPAddressField serializer to separate file.  I didn't find an auto field mapper for DRF, the closet I found was this discussion (https://github.com/encode/django-rest-framework/issues/3244) but doesn't look like something we would want to do (monkey patching ModelSerializer) as "No, the recommended way is to just specify the field on the serializer class - not a big deal."

netbox/ipam/api/field_serializers.py
```diff
@@ -0,0 +1,38 @@
...
+    """IPAddressField with mask"""
+
+    default_error_messages = {
+        'invalid': _('Enter a valid IPv4 or IPv6 address with optional mask.'),
+    }
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        validator = validate_ipaddress_with_mask
+        self.validators.append(validator)
```
jeremystretch, line 27: I don't think we need to add the validator since we're already validating the address format under `to_internal_value()`. My impression is that the validators are more intended for the application of conditional rules (e.g. a string's max length) as opposed to the data format. The logic is redundant anyway: The validator doesn't provide any further validation beyond what you have under `to_internal_value()` already, so I think it's safe to ditch, unless the validator plays a role in generating the schema somehow.

arthanson: done

jeremystretch:
> I didn't find an auto field mapper for DRF

Ack. I must have been thinking about something else; probably django-filters. Apologies for the wild goose chase!

Issue #12288, opened 2023-04-18
OpenAPI spec should include servers property


abhi1693:
### NetBox version

v3.5.0-beta2

### Feature type

Change to existing functionality

### Proposed functionality

Add `servers` property to the spec

### Use case

Having `servers` allows the client to easily run on targetted systems as a sane default value.

Log snippet from OpenAPI client generator

```
[main] INFO  o.o.codegen.utils.URLPathUtils - 'host' (OAS 2.0) or 'servers' (OAS 3.0) not defined in the spec. Default to [http://localhost] for server URL [http://localhost/]
```

PR #12300, opened 2023-04-19
12288 add servers to spectacular settings

arthanson:
### Fixes: #12288 

Include servers in spectacular settings

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -581,13 +581,13 @@ REST_FRAMEWORK = {
 #
 
 SPECTACULAR_SETTINGS = {
-    "TITLE": "NetBox API",
-    "DESCRIPTION": "API to access NetBox",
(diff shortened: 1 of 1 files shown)


netbox/netbox/settings.py
```diff
@@ -586,9 +586,10 @@ def _setting(name, default=None):
     "LICENSE": {"name": "Apache v2 License"},
     "VERSION": VERSION,
     'COMPONENT_SPLIT_REQUEST': True,
+    'REDOC_DIST': 'SIDECAR',
```
jeremystretch, line 589: Can you clean this up so that the dictionary keys and values use either single or double quotes consistently please?

arthanson: done

arthanson:
Test were failing, but it looks like in feature as well, not from this PR... checking...

PR #12311, opened 2023-04-20
12255 inventory item device change

### LINKED ISSUE #12255: Inventory item cannot be reassgined to another device
### NetBox version

v3.4.8

### Python version

3.8

### Steps to Reproduce

1. Click on Edit on any inventory item in the UI
2. Edit Device field
3. Click save

### Expected Behavior

The inventory item should be assigned to the selected device.

### Observed Behavior

"Components cannot be moved to a different device." validation error is shown
This validation is introduced in #11567 , but I don't think the change that even inventory items cannot be reassigned to another device via the UI was intended.

arthanson:
### Fixes: #12255 

Allows inventory item to change device.
Fixes template models setting device type - no exception needed for inventory_item_template as that is restricted from changing device_type in the UI as well.

### DIFF
--- a/netbox/dcim/models/device_components.py
+++ b/netbox/dcim/models/device_components.py
@@ -67,6 +67,7 @@ class ComponentModel(NetBoxModel):
         max_length=200,
         blank=True
     )
+    _can_switch_device = False
 
     class Meta:
         abstract = True
@@ -97,7 +98,7 @@ class ComponentModel(NetBoxModel):
     def clean(self):
         super().clean()
 
-        if self.pk is not None and self._original_device != self.device_id:
+        if (not self._can_switch_device) and (self.pk is not None) and (self._original_device != self.device_id):
             raise ValidationError({
                 "device": "Components cannot be moved to a different device."
             })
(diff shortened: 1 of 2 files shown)


netbox/dcim/models/device_components.py
```diff
@@ -97,7 +98,7 @@ def to_objectchange(self, action):
     def clean(self):
         super().clean()
 
-        if self.pk is not None and self._original_device != self.device_id:
+        if (not self._can_switch_device) and (self.pk is not None) and (self._original_device != self.device_id):
```
jeremystretch, line 101: I'm not sure setting a variable to indicate whether a component is moveable is the ideal pattern. Instead, we could define a list of exempt classes:

```suggestion
        if type(self) not in [InventoryItem] and self.pk is not None and self._original_device != self.device_id:
```

netbox/dcim/models/device_components.py
```diff
@@ -67,6 +67,7 @@ class ComponentModel(NetBoxModel):
         max_length=200,
         blank=True
     )
+    _can_switch_device = False
```
jeremystretch, line 70:
```suggestion
```

netbox/dcim/models/device_components.py
```diff
@@ -1128,6 +1129,7 @@ class InventoryItem(MPTTModel, ComponentModel):
     objects = TreeManager()
 
     clone_fields = ('device', 'parent', 'role', 'manufacturer', 'part_id',)
+    _can_switch_device = True
```
jeremystretch, line 1132:
```suggestion
```

jeremystretch (changes requested):

PR #12335, opened 2023-04-24
Fixes color mode selection

### LINKED ISSUE #12330: Color mode does not take system theme in consideration after initial load
### NetBox version

v3.4.8

### Python version

3.10

### Steps to Reproduce

1. Set system theme to dark
2. Clear brower cache and local storage
3. Notice the UI shows dark
4. Change the system theme to light
5. No change is seen on the UI

### Expected Behavior

Theme should work with system theme if user prefers it

### Observed Behavior

Theme discards user preference

abhi1693:


jeremystretch:
Thanks @abhi1693! I'm going to hold off on this until v3.5, just because it touches a lot of Javascript that may be difficult to fully test across various browsers & operating systems, and v3.4.9 is the last planned release for v3.4

abhi1693:
This is supposed to be released under v3.5 only due to the reasons you have mentioned and the introduction of a new varriable.

jeremystretch:
I haven't dug into this, but it appears to break the ability to toggle light/dark mode using the link under the user menu (top right corner).

abhi1693:
Ok, so the functionality only works when changed via the Profile option. Via that it still works as expected. Do we need the extra options in the footer and the top-right drop-down? Because the whole idea was to get rid of local storage to avoid further issues.

jeremystretch:
> Do we need the extra options in the footer and the top-right drop-down?

Yes, I believe we want to keep those, so that users have the option of easily toggling light/dark mode directly within the UI.

abhi1693:
I'll redo the implementation using local storage only and I think I should be able to clean up the complex logic that's in there.

PR #12344, opened 2023-04-25
Closes 10759: Enable markdown support for custom field descriptions

### LINKED ISSUE #10759: Enable markdown in custom field description fields and custom validator error messages
### NetBox version

v3.3.6

### Feature type

New functionality

### Proposed functionality

Enable markdown syntax on custom field descriptions and error messages in custom validators. Not all syntax options need to be supported; at least boldface, italics, and links.

### Use case

The ability to link to outside content (e.g. documentation) in either descriptions or error messages would be helpful. Formatting options like boldface and italics can make error messages easier to understand.

### Database changes

Possibly expand the maximum length of custom field description fields, as markdown syntax may significantly increase the value stored.

### External dependencies

None

decoupca:
### Fixes: #10759

Enables markdown support for:
* Custom field descriptions
* ~~Custom validator error messages~~


jeremystretch:
I agree with supporting Markdown rendering for custom field descriptions, but I see no need to implement it for form errors: These values are not user-configurable. (It would need to be a separate FR anyway, as the change would impact far more than just custom fields.)

decoupca:
@jeremystretch 

> I agree with supporting Markdown rendering for custom field descriptions, but I see no need to implement it for form errors: These values are not user-configurable. (It would need to be a separate FR anyway, as the change would impact far more than just custom fields.)

Custom validator error messages are user-configurable, which in my case would benefit from markdown support (linking to internal KBs, etc.) Code in this PR achieves that, which didn't look like it had a large impact. Maybe I missed something?

decoupca:
@jeremystretch copy that, reverted changes.

jeremystretch:
Thanks @decoupca!

RadaofTheLight (approved):

PR #12357, opened 2023-04-26
12252 allow sorting on object in search

### LINKED ISSUE #12252: Sorting search results by Object column does not work
### NetBox version

v3.4.8

### Python version

3.8

### Steps to Reproduce

1. Run any search that returns more than 1 result
2. Click the Object header to sort

### Expected Behavior

Results table should be sorted lexicographically by object name.

### Observed Behavior

Results table is sorted by some other unclear criteria

arthanson:
### Fixes: #12252 

Adds a name column (str of object) to allow sorting of the object column in search results:

1. tables2 looks like it only supports sorting based on items in the query results.
2. There is not a consistent field on object to sort on that matches the str of the object.
3. There are max 1,000 search results returned that are already in a list, so going through all the items as doing a str should not be too bad performance wise

The alternative would be restricting this field from being sortable, but it sounds like people want to sort on this field.

### DIFF
--- a/netbox/netbox/search/backends.py
+++ b/netbox/netbox/search/backends.py
@@ -145,12 +145,11 @@ class CachedValueSearchBackend(SearchBackend):
         )
 
         # Omit any results pertaining to an object the user does not have permission to view
-        ret = [
-            r for r in results if r.object is not None
-        ]
-        for r in ret:
-            r.name = str(r.object)
-
+        ret = []
+        for r in results:
+            if r.object is not None:
+                r.name = str(r.object)
+                ret.append(r)
         return ret
 
     def cache(self, instances, indexer=None, remove_existing=True):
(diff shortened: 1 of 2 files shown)


netbox/netbox/search/backends.py
```diff
@@ -145,9 +145,13 @@ def search(self, value, user=None, object_types=None, lookup=DEFAULT_LOOKUP_TYPE
...
 
         # Omit any results pertaining to an object the user does not have permission to view
-        return [
+        ret = [
             r for r in results if r.object is not None
         ]
+        for r in ret:
+            r.name = str(r.object)
+
+        return ret
```
jeremystretch, line 154:
```suggestion
        ret = []
        for r in results:
            if r.object is not None:
                r.name = str(r.object)
                ret.append(r)

        return ret
```

arthanson: done

PR #12358, opened 2023-04-26
Add additional characters to exclude from url encode

### LINKED ISSUE #12355: Certain characters are incorrectly escaped when rendering custom links
### NetBox version

v3.4.9

### Python version

3.8

### Steps to Reproduce

1. Create a custom link with the URL `http://example.com/?foo=1&bar=2`
2. View an object where the custom link appears and inspect its source

### Expected Behavior

The unmodified URL should be present.

### Observed Behavior

The link is incorrectly escaped to `http://example.com/?foo%3D1&bar%3D2`. This is a regression resulting from 89fa546a1481e56d14cfc4d666a55482b0c85fd1.

x64x6a:
### Fixes: #12355

Excludes additional characters from being encoded.  This prevents valid URI characters (such as `=`) from being encoded and encoded values from being encoded again (such as `%DD` or `+`).

### DIFF
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -279,7 +279,7 @@ class CustomLink(CloningMixin, ExportTemplatesMixin, WebhooksMixin, ChangeLogged
         text = clean_html(text, allowed_schemes)
 
         # Sanitize link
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @x64x6a!

PR #12382, opened 2023-04-28
Fixes #12380 - Add ObjectChange as a valid option for ObjectListWidget and ObjectCountsWidget

### LINKED ISSUE #12380: Dashboard - Changelog can't be displayed
### NetBox version

v3.5.0

### Python version

3.8

### Steps to Reproduce

After an upgrade from 3.4.8, Default Widget displays "Change Log" in blue. I wanted to change the color

From demo website
image

Editing the entry shows 
image

I can't find extras > changelog in "model"

### Expected Behavior

Extra > Changelog should be in the menu

### Observed Behavior

image

kkthxbye-code:
### Fixes: #12380

Adds the objectchange contenttype to the valid objects for the widgets. Not sure if there's a prettier way.

### DIFF
--- a/netbox/extras/dashboard/widgets.py
+++ b/netbox/extras/dashboard/widgets.py
@@ -8,6 +8,7 @@ from django import forms
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.core.cache import cache
+from django.db.models import Q
 from django.template.loader import render_to_string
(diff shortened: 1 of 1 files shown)


jeremystretch:
> Not sure if there's a prettier way.

None that I can think of. We really need a canonical mechanism for qualifying "real" content types.

PR #12408, opened 2023-05-01
Fixes #12402 - When bulk-importing related objects, add the errors to the base form instead of the model_form

### LINKED ISSUE #12402: JSON import of circuit with terminations fails without error
### NetBox version

v3.4.10

### Python version

3.8

### Steps to Reproduce

1. Create circuit type and provider (I am using "Internet" and "Cerberus" in the example below)
2. Go to Circuits > Import
3. Paste in the following:

```
[
    {
        "cid": "ID-TEST",
        "commit_rate": 200000,
        "description": "Testing Import",
        "provider": "Cerberus",
        "status": "decommissioned",
        "terminations": [
            {
                "site": {
                    "name": "Test"
                }
            }
        ],
        "type": "Internet"
    }
]
```

### Expected Behavior

*Either* the import to complete successfully, *or* an error message to be displayed.

### Observed Behavior

Netbox returns to the import page, without showing any error, and without completing the import.

Aside 1: if I give a provider name or circuit type that doesn't exist, this *is* correctly reported as an error.

Aside 2: I can find no documented example of the correct format for this, so I'm just going by guesswork - but if my guess is wrong, I think Netbox should at least say what's wrong.

Related issues: #11469, #11204, #815.

kkthxbye-code:
### Fixes: #12402

The errors of related models are currently added to the `model_form`. If there's any errors in a related object form, an AbortTransaction is thrown. The errors saved on the model form are never replicated to the base form and thus are never shown on the form.

I'm really unsure of the fix in this PR, but it seems to work. The code is pretty complex imo, a lot of different forms being passed around, so not sure if there's a better way. Feel free to just close this PR and fix it another way, just wanted to have it written down.

### DIFF
--- a/netbox/netbox/views/generic/bulk_views.py
+++ b/netbox/netbox/views/generic/bulk_views.py
@@ -313,7 +313,7 @@ class BulkImportView(GetReturnURLMixin, BaseMultiObjectView):
         """
         return data
 
-    def _save_object(self, model_form, request):
+    def _save_object(self, model_form, request, form):
 
         # Save the primary object
         obj = self.save_object(model_form, request)
@@ -342,7 +342,7 @@ class BulkImportView(GetReturnURLMixin, BaseMultiObjectView):
                     for subfield_name, errors in f.errors.items():
                         for err in errors:
                             err_msg = "{}[{}] {}: {}".format(field_name, i, subfield_name, err)
-                            model_form.add_error(None, err_msg)
+                            form.add_error(None, err_msg)
                     raise AbortTransaction()
 
             # Enforce object-level permissions on related objects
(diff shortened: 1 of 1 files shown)


jeremystretch:
I think the root issue here is that the related object forms are not validated until _after_ we've begun to save the parent object. We'll need to unwind some of the bulk logic to correct this.

kkthxbye-code:
> the related object forms are not validated until after we've begun to save the parent object

I'm assuming this is because the clean methods do validation on the reference between the related object and the parent, so we need the pk of the parent before we can validate the related object?

Regardless, feel free to close this PR and un-assign me, it was mostly just a way to get the ball rolling 👍

PR #12432, opened 2023-05-02
Replicate origin hierarchy when creating managed files

### LINKED ISSUE #12431: Creating a report or script from a remote data source does not retain the origin's directory hierarchy
### NetBox version

v3.5.0

### Python version

3.8

### Steps to Reproduce

1. Create a git repository which stores a report and/or script within a subdirectory of the repo root.
2. Create a remote data source in NetBox and synchronize the repo.
3. Create a report or script from one of the files in the repo's subdirectory.

### Expected Behavior

The report or script should be created within a local directory of the same name as the origin subdirectory.

### Observed Behavior

The report or script is created within the corresponding root path locally, without replication the origin hierarchy.

jeremystretch:
### Fixes: #12431

WIP

### DIFF
--- a/netbox/core/models/files.py
+++ b/netbox/core/models/files.py
@@ -75,7 +75,9 @@ class ManagedFile(SyncedDataMixin, models.Model):
 
     def sync_data(self):
         if self.data_file:
-            self.file_path = os.path.basename(self.data_path)
+            self.file_path = self.data_path
+            path, _ = os.path.split(self.full_path)
+            os.makedirs(path, exist_ok=True)
             self.data_file.write_to_disk(self.full_path, overwrite=True)
 
     def delete(self, *args, **kwargs):
--- a/netbox/extras/models/mixins.py
+++ b/netbox/extras/models/mixins.py
@@ -13,6 +13,11 @@ class PythonModuleMixin:
         return os.path.splitext(self.file_path)[0]
 
     @property
+    def url_path(self):
+        path, _ = os.path.split(self.data_path)
+        return f"{path}/{self.python_name}"
+
+    @property
     def python_name(self):
         path, filename = os.path.split(self.full_path)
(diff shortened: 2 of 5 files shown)


kkthxbye-code:
The following issues were found by using https://github.com/kkthxbye-code/netboxscriptexample as a testcase. The repo was added as a datasource, and ScriptModules were added one by one for all files, including `__init__.py` files.

First issue is that relative imports do not seem to work, changing:

https://github.com/kkthxbye-code/netboxscriptexample/blob/a09ad436ab50440723a9f20fc633c3daddc1bb18/nestedmodule/nested_script1.py#L3

To:

`from scripts.nestedmodule.base_script import BaseScript`

Makes it work.

After this the script page looks like this:

image

The first obvious issue is that non-script modules need to be imported to be available in the source path and will show up as empty. Not sure how to fix this, as it seems at odds with the new script system.

Taking the Scripts from the top.

**nestedmodule:**

* Title should be "My Nested Sripts" which is defined in the __init__.py file.
* Opening the Scripts generate a URL like `/extras/scripts/nestedmodule/nestedmodule/nested_script2.NestedScript2/` which 404's. nestedmodule appears twice, removing it nets another error: `get() returned more than one ScriptModule -- it returned 4!`

**nested_script1 and nested_script2:**

Should not be shown, but since we have to add it as a ScriptModule it has to. Opening and running works, but using the Jobs or Source tab loads a 404 URL of this format `/extras/scripts/nested_script1/NestedScript1/jobs/`

**root_script**

Do not work, generates URLs like these: `/extras/scripts//root_script/RootScript1/` - removing the extra slash makes it load and run. Source and jobs tabs do not work, generates URL's like these: `/extras/scripts/root_script/RootScript1/source/`. Also seems like the `name` does not function anymore.

For reference this is what the example repo looks like in netbox 3.4:

image

PR #12434, opened 2023-05-02
12433 update object list widget to correctly parameterize urls

### LINKED ISSUE #12433: Dashboard widgets incorrectly concatenate params
### NetBox version

v3.5.0

### Python version

3.8

### Steps to Reproduce

1. create a new widget of object list configuration
2. choose model "DCIM > Site"
3. Put "10" in Page size
4. put {"tenant_id": 5} in URL parameters
5. click save

### Expected Behavior

widget will show list based on filtered params.

### Observed Behavior

All sites are shown. Debugging shows the URL passed to the view is "?tenant_id=5?page_size=10" (note the two "?") correct url should have "&" as the param separator.

arthanson:
### Fixes: #12433 

Moves the page_size concatenation for object list view widget from the template to the widget code so parameter concatenation is done correctly.

### DIFF
--- a/netbox/extras/dashboard/widgets.py
+++ b/netbox/extras/dashboard/widgets.py
@@ -229,7 +229,13 @@ class ObjectListWidget(DashboardWidget):
             htmx_url = reverse(viewname)
         except NoReverseMatch:
             htmx_url = None
-        if parameters := self.config.get('url_params'):
+        parameters = self.config.get('url_params')
+        if page_size := self.config.get('page_size'):
+            if not parameters:
+                parameters = {}
+            parameters['page_size'] = page_size
+
+        if parameters:
             try:
                 htmx_url = f'{htmx_url}?{urlencode(parameters, doseq=True)}'
             except ValueError:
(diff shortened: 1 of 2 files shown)


netbox/extras/dashboard/widgets.py
```diff
@@ -229,7 +229,13 @@ def render(self, request):
             htmx_url = reverse(viewname)
         except NoReverseMatch:
             htmx_url = None
-        if parameters := self.config.get('url_params'):
+        parameters = self.config.get('url_params')
+        if page_size := self.config.get('page_size'):
+            if not parameters:
+                parameters = {}
+            parameters['page_size'] = page_size
```
jeremystretch, line 236: We can clean this up a bit by passing `{}` as the default for null `url_params`.
```suggestion
        parameters = self.config.get('url_params', {})
        if page_size := self.config.get('page_size'):
          parameters['page_size'] = page_size
```

jeremystretch (changes requested):

arthanson:
updated

PR #12435, opened 2023-05-02
12363 update paragraph spacing on journal table

### LINKED ISSUE #12363: New paragraph line breaks do not render in journal entry table comments
### NetBox version

v3.3.7

### Python version

3.8

### Steps to Reproduce

1. Click on **Devices** in the side menu and then select **Devices** from the Devices submenu
2. Click on a Device object and select the **Journal** tab
3. Create a new journal entry and enter the following in the comments section:

image

### Expected Behavior

The journal entry comments attribute within the device object's journal entry table should render a line break between the two paragraphs.

This behaviour is correctly demonstrated when clicking into the Journal Entry object and observing the **Comments** attribute:
image

### Observed Behavior

The journal entry comments attribute within the device object's journal entry table does not render a line break between the two paragraphs:
image

Prior to the user interface updates in v3, the journal entry comments attribute within the device object's journal entry table would correctly render new paragraph line breaks.

arthanson:
### Fixes: #12363 

Updates paragraph spacing on journal table so comments are shown correctly spaced.  Only effects journal-table as other tables should display with condensed paragraphs.


netbox/templates/extras/object_journal.html
```diff
@@ -6,7 +6,7 @@
 
 {% block content %}
   <div class="card">
-    <div class="card-body table-responsive">
+    <div class="card-body table-responsive journal-table">
```
jeremystretch, line 9: Big :-1: on bespoke CSS classes. Instead, can we either a) restore the margin universally, or b) introduce a separate utility class that removes the margin where needed?

arthanson:
changed to a generic spaced-paragraph-table utility class. let me know if you want any further changes.

jeremystretch:
Rather than adding a one-off class for this, let's do some digging to determine why and how the margin is being removed in the first place. We can attribute the style to this rule in `netbox.scss`:

https://github.com/netbox-community/netbox/blob/78ec3a6411c57b29b7da3d885845c78f9b3062da/netbox/project-static/styles/netbox.scss#L232-L235

Running git blame on this file show that the change was part of a [large array of changes](https://github.com/netbox-community/netbox/commit/1de46f592cabc4239246154dca1adb430340e1d9) two years ago, likely when the v3.0 UI was still being refined. So I think it's reasonable to just remove it entirely. If complaints arise that this results in suboptimal rendering elsewhere in the UI, we can then add a "condensed" class to remove the margin where needed.

kkthxbye-code:
> wonder if we can easily apply the rule only to the last `<p>` element in the cell.

I'm not really a CSS aficionado, but the :last-child pseudo-class would probably work.

arthanson:
That historical context helps.  Reverted and made the last-child to 0 and the regular p table spacing to .5em - this allows you to still see paragraph spacing, but gets rid of it at the bottom and reduces the spacing to conserve line height:
Monosnap Devices | NetBox 2023-05-03 08-21-41

jeremystretch:
Thanks @arthanson!

PR #12440, opened 2023-05-02
12219 adjust widget header contrast

### LINKED ISSUE #12219: New dashboard widgets have insufficient header contrast in dark mode
### NetBox version

v3.5.0-dev

### Python version

3.8

### Steps to Reproduce

1. Navigate to the dashboard screen with the new widgets.

### Expected Behavior

The text and icon colors within the card header should maintain at least the same contrast ratio between light and dark modes. You can see this done within the search and log in buttons in the screenshots below.

### Observed Behavior

Currently, the colors used for the text and icons for card headers in dark mode do not provide much contrast against the background, making them very difficult to see. Screenshots below for reference.

beta-demo netbox dev_light
beta-demo netbox dev_dark

arthanson:
### Fixes: #12219 

1. updates foreground_color template tag to work with color values "red" instead of hex values.
2. change widget template to use this for text and icons
3. adds some utility functions to lighten (and darken) a color, used to make gear and close icons lighter then text

### DIFF
--- a/netbox/templates/extras/dashboard/widget.html
+++ b/netbox/templates/extras/dashboard/widget.html
@@ -1,4 +1,5 @@
 {% load dashboard %}
+{% load helpers %}
 
 <div
     class="grid-stack-item"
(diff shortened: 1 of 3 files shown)


arthanson:
@jeremystretch went through all the colors in dark mode and it works there as well.  However "Default" on dark mode doesn't work (black on dark-grey) however IMHO we should look at a separate issue/bug to remove or replace Default as it currently uses outline-dark which is not a color and is just he widget background color which is not very useful in either dark or light mode.  I don't think it is possible to fix this for Default as it is the background widget color that changes between dark and light mode which isn't detectable from python.  I think removing default should be a separate issue as it will require some migrations and we may want to refactor widget color more.

This fixes the text color for widgets and I think is useful, it just doesn't fix it for dark mode default (which should probably be removed).

jeremystretch:
I'm to close this out in favor of the simpler fix in PR #13753. We may reconsider something more elegant as part of the upcoming UI refresh.

PR #12448, opened 2023-05-03
Fixes: #11715 - Fix Parent Prefix table display

### LINKED ISSUE #11715: Prefix in different VRF wrongly sees global prefix as its parent
### NetBox version

v3.4.4

### Python version

3.8

### Steps to Reproduce

1. Create VRF "testvrf"
2. Create prefix 192.168.0.0/24 in global VRF
3. Create prefix 192.168.0.0/28 in "testvrf"
4. Browse to 192.168.0.0/28 (which is where you'll be right after creation)

### Expected Behavior

192.168.0.0/28@testvrf and 192.168.0.0/24 should be independent from each other, as they are in different VRFs.

### Observed Behavior

192.168.0.0/28@testvrf sees 192.168.0.0/24 as its parent.

image

However, 192.168.0.0/24 does *not* see 192.168.0.0/28@testvrf as its child.

This is inconsistent: this child<->parent relationship should either work in both ways, or neither.

In this case, I think it should be neither: otherwise 192.168.0.0/24 could see many different prefixes and IP addresses in different VRFs as its children, which would mix lots of VRFs together in the same view.

EDIT: the current behaviour *does* appear to be intentional in the [code](https://github.com/netbox-community/netbox/blob/v3.4.4/netbox/ipam/views.py#L461-L463):

```
        # Parent prefixes table
        parent_prefixes = Prefix.objects.restrict(request.user, 'view').filter(
            Q(vrf=instance.vrf) | Q(vrf__isnull=True)
```

DanSheps:
### Fixes: #11715 

Fixes the display of a prefix's parent(s) in the prefix detailed view when the prefix is in a VRF and there are global prefixes that could be parents.  If the "parent" prefixes have an active status, they will also show even though they should only show when the status is "container"

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -496,7 +496,7 @@ class PrefixView(generic.ObjectView):
 
         # Parent prefixes table
         parent_prefixes = Prefix.objects.restrict(request.user, 'view').filter(
-            Q(vrf=instance.vrf) | Q(Q(vrf__isnull=True) & Q(status=PrefixStatusChoices.STATUS_CONTAINER))
+            Q(vrf=instance.vrf) | Q(vrf__isnull=True, status=PrefixStatusChoices.STATUS_CONTAINER)
         ).filter(
             prefix__net_contains=str(instance.prefix)
         ).prefetch_related(
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -14,6 +14,7 @@ from utilities.views import ViewTab, register_model_view
 from virtualization.filtersets import VMInterfaceFilterSet
 from virtualization.models import VMInterface
 from . import filtersets, forms, tables
+from .choices import PrefixStatusChoices
 from .constants import *
 from .models import *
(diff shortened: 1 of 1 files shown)


DanSheps:
Going to check the IP table to ensure there are no issues there either.

netbox/ipam/views.py
```diff
@@ -495,7 +496,7 @@ def get_extra_context(self, request, instance):
 
         # Parent prefixes table
         parent_prefixes = Prefix.objects.restrict(request.user, 'view').filter(
-            Q(vrf=instance.vrf) | Q(vrf__isnull=True)
+            Q(vrf=instance.vrf) | Q(Q(vrf__isnull=True) & Q(status=PrefixStatusChoices.STATUS_CONTAINER))
```
jeremystretch, line 499: We should be able to combine these into a single Q object.

```suggestion
            Q(vrf=instance.vrf) | Q(vrf__isnull=True, status=PrefixStatusChoices.STATUS_CONTAINER)
```

jeremystretch:
Thanks @DanSheps!

PR #12452, opened 2023-05-03
Fixes: #10757 - Change IP interface assignment to use new selector

### LINKED ISSUE #10757: Edit ip address - change device assignment, doesnt clear interface assignment. causes confusion
### NetBox version

v3.3.5

### Python version

3.10

### Steps to Reproduce

1. goto an ip address with an existing assignment, eg https://demo.netbox.dev/ipam/ip-addresses/186/
2. edit the ip address, change the vm or device assignment, eg from vm1 to vm2.
3. Notice that interface field below still says eth0
(its not obvious but that eth0 field is still set to eth0 on vm1, not eth0 on vm2 which is selected above)
4. user saves form, result is accepted. however interface assignment is still set to vm1 - eth0.

### Expected Behavior

I would expect that at step 2 once the device or vm relationship was changed that the interfaces field would empty out

OR

that during the save process data validation would fail and tell the user that the device / virtual machine assignment + interface are not part of the same device.

OR

maybe the interface assignment (eth0) should also have the vm or device name listed after it? making it obvious that the field above didnt change anything and is just a filter?

### Observed Behavior

assignment does not change in netbox. user gets no warning that there was a conflict or no change made

### LINKED ISSUE #11477: IP Address Interface Assignment silently discards selected Device if no Interface is selected
### NetBox version

v3.4.2

### Python version

3.8

### Steps to Reproduce

1. Create a Device
2. Fill out required fields
3. Save Device
4. Create IP Address
5. Fill out required fields
6. Choose the created Device in the Interface Assignment
7. Do not choose any Interface in the Interface Assignment
8. Save IP Address

### Expected Behavior

When saving the IP Address object the form gets validated and at this point I expect to get a warning which tells me to choose an interface in case I want to assign this IP Address to the Device, else to clear the device field.

### Observed Behavior

The device selection gets silently discarded. Users can't find the IP Address they have "assigned" to a Device anywhere. Only on editing the IP Address again shows that the value never got stored.

DanSheps:
### Fixes: #10757

Convert the interface/FHRP assignment on IPAddress edit form to use the new selector system.  This should reduce the confusion as to whether a "device" is assigned or not.

This will also resolve #11477 in a similar manner.

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -262,30 +262,18 @@ class IPRangeForm(TenancyForm, NetBoxModelForm):
 
 
 class IPAddressForm(TenancyForm, NetBoxModelForm):
-    device = DynamicModelChoiceField(
-        queryset=Device.objects.all(),
-        required=False,
-        initial_params={
-            'interfaces': '$interface'
-        }
-    )
     interface = DynamicModelChoiceField(
         queryset=Interface.objects.all(),
         required=False,
+        selector=True,
         query_params={
             'device_id': '$device'
-        }
-    )
-    virtual_machine = DynamicModelChoiceField(
-        queryset=VirtualMachine.objects.all(),
-        required=False,
-        initial_params={
-            'interfaces': '$vminterface'
         }
     )
     vminterface = DynamicModelChoiceField(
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks @DanSheps!

PR #12455, opened 2023-05-03
Adds module status to module bay table

### LINKED ISSUE #11652: Add Status column to Module Bays tables
### NetBox version

v3.4.3

### Feature type

### Proposed functionality

A *Status* field was added to *Modules* in #10371. I am proposing making that field available as a column on the *Module Bays* tables (full list and tab under a device). There is existing precedent for *Module* fields to be added as column options for those tables, like *Serial number* and *Asset tag*. I would additionally propose adding that column as one of the defaults on those tables, as was done for the *Modules* table.

### Use case

Adding this as a table column would increase its visibility and make it easier to understand at a glance the status of all the modules in a shelf. Currently, a user would need to click into a given module in the list to see its status, which becomes cumbersome on a shelf with many modules.

### Database changes

None

### External dependencies

None

abhi1693:

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -785,17 +785,17 @@ class ModuleBayTable(DeviceComponentTable):
     tags = columns.TagColumn(
         url_name='dcim:modulebay_list'
     )
-    status = columns.TemplateColumn(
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @abhi1693!

PR #12456, opened 2023-05-03
12416 warning for missing script file

### LINKED ISSUE #12416: Cannot add or remove scripts from Custom Scripts after updating to 3.5.0
### NetBox version

v.3.5.0

### Python version

3.8

### Steps to Reproduce

1. Add any custom Python script into /opt/netbox/netbox/scripts
2. Access Scripts tab on Web UI, newly added script that was put into /opt/netbox/netbox/scripts does not show up

OR

1. Remove any existing script from /opt/netbox/netbox/scripts that was present before upgrading to v3.5.0
2. Access Scripts tab on web UI, this will result in an error message on web UI.

~~Originally I thought this was due to a bug with the netbox-secrets plugin (not supported on 3.5.0 yet, hence why the Changelog doesn't work) so I have removed the netbox-secrets plugin and dropped the related tables from the psql DB according to the Remove a Plugin documentation and restarted NetBox, however the issue with the custom scripts.~~

### Expected Behavior

New scripts should be present in the "Scripts" tab in NetBox, removed script should not be present / not error.

### Observed Behavior

Server Error

<class 'FileNotFoundError'>

[Errno 2] No such file or directory: '/opt/netbox/netbox/scripts/add_device_type_components.py'

Python version: 3.8.10
NetBox version: 3.5.0

arthanson:
### Fixes: #12416 

Fixes the case of missing script files causing the script list view to error out.  Now shows a warning message for the admin to check the file (see screenshot below).  Also removed a left-over print statement in the code.

Monosnap Scripts | NetBox 2023-05-03 14-26-20

### DIFF
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -56,7 +56,7 @@ class ScriptModule(PythonModuleMixin, JobsMixin, ManagedFile):
         module = None
         try:
             module = self.get_module()
-        except OSError as e:
+        except Exception as e:
             pass
 
         scripts = {}
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -53,7 +53,12 @@ class ScriptModule(PythonModuleMixin, JobsMixin, ManagedFile):
             # For child objects in submodules use the full import path w/o the root module as the name
             return cls.full_name.split(".", maxsplit=1)[1]
 
-        module = self.get_module()
+        module = None
+        try:
+            module = self.get_module()
+        except OSError as e:
+            pass
+
         scripts = {}
         ordered = getattr(module, 'script_order', [])
 
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -53,11 +53,10 @@ class ScriptModule(PythonModuleMixin, JobsMixin, ManagedFile):
             # For child objects in submodules use the full import path w/o the root module as the name
             return cls.full_name.split(".", maxsplit=1)[1]
 
-        module = None
         try:
             module = self.get_module()
         except Exception as e:
-            pass
+            module = None
 
         scripts = {}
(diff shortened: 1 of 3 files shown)


netbox/extras/models/scripts.py
```diff
@@ -53,7 +53,12 @@ def _get_name(cls):
             # For child objects in submodules use the full import path w/o the root module as the name
             return cls.full_name.split(".", maxsplit=1)[1]
 
-        module = self.get_module()
+        module = None
+        try:
+            module = self.get_module()
+        except Exception as e:
+            pass
```
jeremystretch, line 60:
```suggestion
        try:
            module = self.get_module()
        except Exception as e:
            module = None
```

jeremystretch: We also might want to log the missing file.

netbox/templates/extras/script_list.html
```diff
@@ -37,43 +37,51 @@ <h5 class="card-header" id="module{{ module.pk }}">
...
-                  </tr>
-                {% endfor %}
-              {% endwith %}
-            </tbody>
-          </table>
+          {% if not module.scripts %}
+            <div class="alert alert-warning d-flex align-items-center" role="alert">
+              <div>
+                Warning: Script file at: <strong>{{ module.full_path }}"</strong> is either missing or has a problem with the script code, please check the file.
+              </div>
```
jeremystretch, line 44:
```suggestion
              <i class="mdi mdi-alert"></i> Script file at <code>{{ module.full_path }}</code> could not be loaded.
```

jeremystretch (changes requested):
Suggested some tweaks, let me know what you think.

netbox/templates/extras/script_list.html
arthanson: left out the <code> tag as it was really messing up spacing and formatting when narrow window.

PR #12457, opened 2023-05-03
Adds housekeeping systemd files

### LINKED ISSUE #11190: Add netbox housekeeping systemd units
### Change Type

Addition

### Area

Administration

### Proposed Changes

In the [housekeeping documentation](https://github.com/netbox-community/netbox/blob/develop/docs/administration/housekeeping.md) only the way over the antiquated Cron procedure was described. This is very irritating if you consider that the rest of the stored services are already realized via Service Unit.
I have therefore already provided in a fork, a branch, in which I have established service and timer unit, as well as documentation.
For a Pull-Request I need however still one Issue assigned to me, which I request hereby.  At least that's what the texts suggest to me when I want to make a pull request.

https://github.com/YoSiJo/netbox/tree/systemd-housekeeping-docs

abhi1693:

### DIFF
--- a/contrib/netbox-housekeeping.service
+++ b/contrib/netbox-housekeeping.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=NetBox Housekeeping Service
+Documentation=https://docs.netbox.dev/
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=simple
+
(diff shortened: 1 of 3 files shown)


jeremystretch:
Thanks @abhi1693!

PR #12458, opened 2023-05-03
Adds ENGINE to database config

### LINKED ISSUE #11791: Implement DATABASE_ENGINE Configuration Parameter
### NetBox version

v3.4.4

### Feature type

New functionality

### Proposed functionality

I am proposing introduction of a DATABASE_ENGINE parameter, which would allow overwriting the default database engine of 'django.contrib.gis.db.backends.postgresql' to a custom database backend.

This would be an optional configuration parameter within the configuration.py file. In the event METRICS_ENABLED was also set, I propose we give the DATABASE_ENGINE option priority. If DATABASE_ENGINE is not set, and METRICS_ENABLED is set to True, then the database engine would be set to 'django_prometheus.db.backends.postgresql'

### Use case

This would allow extending the netbox datamodel with different Postgres database extensions, for example to support PostGIS to allow geospatial data models.

### Database changes

No database changes.

### External dependencies

No external dependencies until a database engine was overwritten, in which case that engine may potentially have external dependencies.

abhi1693:


jeremystretch (changes requested):
Could you annotate the `ENGINE` setting in the documentation as well please? We should include a note that it must be a PostgreSQL-compatible backend.

jeremystretch:
Thanks @abhi1693!

PR #12485, opened 2023-05-04
Adds bulk import for journal entry

### LINKED ISSUE #12122: Allow Import of Journal Entries
### NetBox version

v3.4.3

### Feature type

New functionality

### Proposed functionality

I would like to have the ability to import Journal Entries via the web UI. (at /extras/journal-entries/import/ ) the same way we import other items in Netbox. 

Currently, seems feasible via API, and only export is implemented via web UI. 

### Use case

When our technicians are doing preventive maintenance of a site, they come back with a list of remarks that we save as Journal entries. They arrive to us in batch, and would like to be able to upload them as CSV via UI.

I'm sure other people would have similar use cases that would apply to other sections of NB. 

Thanks! :)

### Database changes

None

### External dependencies

None

abhi1693:


jeremystretch:
Thanks @abhi1693!

PR #12487, opened 2023-05-04
Adds image attachment list view

### LINKED ISSUE #11932: Sort images by date
### NetBox version

v3.4.1

### Feature type

New functionality

### Proposed functionality

It would be great to be able to sort the "images" table based on the "created" date

[Image](https://imgur.com/a/rsu0Nil)

### Use case

Easily find the latest pictures

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/extras/forms/filtersets.py
+++ b/netbox/extras/forms/filtersets.py
@@ -11,7 +11,7 @@ from extras.utils import FeatureQuery
 from netbox.forms.base import NetBoxModelFilterSetForm
 from tenancy.models import Tenant, TenantGroup
 from utilities.forms import BOOLEAN_WITH_BLANK_CHOICES, FilterForm, add_blank_choice
-from utilities.forms.fields import ContentTypeMultipleChoiceField, DynamicModelMultipleChoiceField, TagFilterField
+from utilities.forms.fields import ContentTypeChoiceField, ContentTypeMultipleChoiceField, DynamicModelMultipleChoiceField, TagFilterField
 from utilities.forms.widgets import APISelectMultiple, DateTimePicker
 from virtualization.models import Cluster, ClusterGroup, ClusterType
 from .mixins import SavedFiltersMixin
@@ -22,6 +22,7 @@ __all__ = (
     'CustomFieldFilterForm',
     'CustomLinkFilterForm',
     'ExportTemplateFilterForm',
+    'ImageAttachmentFilterForm',
     'JournalEntryFilterForm',
     'LocalConfigContextFilterForm',
     'ObjectChangeFilterForm',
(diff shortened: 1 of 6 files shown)


netbox/extras/forms/filtersets.py
```diff
@@ -137,6 +138,22 @@ class ExportTemplateFilterForm(SavedFiltersMixin, FilterForm):
     )
 
 
+class ImageAttachmentFilterForm(SavedFiltersMixin, FilterForm):
+    fieldsets = (
+        (None, ('q', 'filter_id')),
+        ('Attributes', ('content_type', 'name',)),
+    )
+
```
jeremystretch, line 146: Please remove extra newlines between fields

netbox/extras/tables/tables.py
```diff
@@ -1,8 +1,7 @@
-import json
-
 import django_tables2 as tables
 from django.conf import settings
 
+import json
```
jeremystretch, line 4: The `json` import should be at top as it's a standard Python library.

```suggestion
import json

import django_tables2 as tables
from django.conf import settings

```

netbox/extras/tables/tables.py
```diff
@@ -86,6 +86,30 @@ class Meta(NetBoxTable.Meta):
         )
 
 
+class ImageAttachmentTable(NetBoxTable):
+    id = tables.Column(
+        linkify=False
+    )
+
```
jeremystretch, line 93: Please remove extra newlines between columns

netbox/netbox/navigation/menu.py
```diff
@@ -290,6 +290,7 @@
                 get_model_item('extras', 'customfield', _('Custom Fields')),
                 get_model_item('extras', 'customlink', _('Custom Links')),
                 get_model_item('extras', 'exporttemplate', _('Export Templates')),
+                get_model_item('extras', 'imageattachment', _('Image Attachments'), actions=()),
```
jeremystretch, line 293: I'm not sure it makes sense to put this under customizations, though I also can't think of anywhere more fitting. It should probably come at the end of the list though.

netbox/extras/forms/filtersets.py
```diff
@@ -137,6 +138,22 @@ class ExportTemplateFilterForm(SavedFiltersMixin, FilterForm):
     )
 
 
+class ImageAttachmentFilterForm(SavedFiltersMixin, FilterForm):
+    fieldsets = (
+        (None, ('q', 'filter_id')),
+        ('Attributes', ('content_type', 'name',)),
```
jeremystretch, line 144: This needs to match the exact filter name, or it raises an exception.

```suggestion
        ('Attributes', ('content_type_id', 'name',)),
```

jeremystretch (changes requested):
Thanks for taking this @abhi1693!

jeremystretch:
One quick note: I set `orderable=False` on the `size` column, because this value isn't actually stored in the database. (Maybe it should be, but that's a separate discussion.)

PR #12488, opened 2023-05-04
Adds description to elevation device tooltip

### LINKED ISSUE #11801: Tooltip extension - Show the description in the rack for the device.
### NetBox version

v3.4.5

### Feature type

### Proposed functionality

For a rack, you can display a tooltip text for the device.
It would be good if you could also see the description.

image

### Use case

With many devices or splice boxes, one does not know what the purpose of the device is. The description contains information such as: Mailstore server, Active Directory, splice box according to building X, etc.
The device name Server-1, Server-2, Server-3 is not equally meaningful. So you have to click on each device to know whether you have selected the right one.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/dcim/svg/racks.py
+++ b/netbox/dcim/svg/racks.py
@@ -46,14 +46,14 @@ def get_device_description(device):
     Device Type: Device Type Manufacturer Device Type Model (Device Type U Height) (Device Asset Tag) (Device Serial)
     """
     description = f'Name: {device.name}'
-    description += f'\nRole: {device.device_role}'
-    if device.description:
-        description += f'\nDescription: {device.description}'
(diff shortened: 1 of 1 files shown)


kkthxbye-code:
Currently it looks something like this:

image

With a 200 char description, it looks like this (with devtools, so might be wrong):

image

I find it a little busy, it's hard to distinguish where the description starts/ends.

Maybe we could use some whitespace? 

image

It helps legibility imo, but I'm not really a great UX'er :)

abhi1693:
@kkthxbye-code I have implemented the way you have suggested.

jeremystretch:
I really like this, much cleaner! I went ahead and broke out the asset tag and serial to separate lines as well to improve readability.

PR #12490, opened 2023-05-04
Adds db read-only middleware

### LINKED ISSUE #11233: Webfrontend with database in readonly mode

abhi1693:
### Fixes: #11233

image

### DIFF
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -180,7 +180,7 @@ class MaintenanceModeMiddleware:
         self.get_response = get_response
 
     def __call__(self, request):
-        if settings.MAINTENANCE_MODE:
+        if getattr(settings, 'MAINTENANCE_MODE', False):
             self._prevent_db_write_operations()
 
         return self.get_response(request)
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -11,7 +11,7 @@ from django.db.utils import InternalError
 from django.http import Http404, HttpResponseRedirect
 
 from extras.context_managers import change_logging
-from netbox.config import clear_config
+from netbox.config import clear_config, get_config
 from netbox.views import handler_500
 from utilities.api import is_api_request, rest_api_server_error
 
@@ -180,7 +180,7 @@ class MaintenanceModeMiddleware:
         self.get_response = get_response
 
     def __call__(self, request):
-        if getattr(settings, 'MAINTENANCE_MODE', False):
+        if get_config().MAINTENANCE_MODE:
             self._prevent_db_write_operations()
 
         return self.get_response(request)
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -3,10 +3,11 @@ import uuid
 from urllib import parse
 
 from django.conf import settings
-from django.contrib import auth
+from django.contrib import auth, messages
 from django.contrib.auth.middleware import RemoteUserMiddleware as RemoteUserMiddleware_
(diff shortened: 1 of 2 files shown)


jeremystretch:
I like the approach here, however I don't think it satisfies FR #11233, which specifically asks for a dedicated "read-only" mode separate from maintenance mode. There's definitely a lot of overlap between the two concepts though, and I'm not sure whether having both makes sense.

jeremystretch:
Maybe we could keep this as is, and introduce a `BANNER_MAINTENANCE` configuration parameter which, if set to None, would disable the maintenance banner. Seems like that would be sufficient.

abhi1693:
Would it not be confusing to have 2 different modes? Also, currently maintenance mode only displays a banner that does not do anything functionaly which I believe this PR fixes by putting the app in read only.

In maintenance mode, I as an would want to continue to do whatever I need in the background but still keep the application up in read-only which IMO is a better option.

abhi1693:
> introduce a BANNER_MAINTENANCE configuration parameter

I think that's also a good idea. Do you want to do this implementation as a part of this PR itself?

jeremystretch:
No, I think it deserves its own FR. I'll put one in.

jeremystretch:
Opened #12554 for the `BANNER_MAINTENANCE` config parameter.

abhi1693:
@jeremystretch Should I update the error message with the `BANNER_MAINTENANCE` value here? That seems logical to me.

jeremystretch:
I'm not sure about that, because we can't be certain that whatever text the user configures for the banner is going to be relevant. IMO it's best to stick with a static error message that explicitly explains that write operations aren't currently supported. (If there's a maintenance banner value, they'll see it in the UI anyway.)

netbox/netbox/middleware.py
```diff
@@ -166,3 +168,43 @@ def _get_groups(self, request):
...
+    """
+    Middleware that checks if the application is in maintenance mode
+    and restricts write-related operations to the database.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if getattr(settings, 'MAINTENANCE_MODE', False):
```
jeremystretch, line 183: We need to reference this setting from `get_config()` because it's a dynamic parameter, which can be set either in `configuration.py` or a config revision in the database.

```suggestion
        if get_config().MAINTENANCE_MODE:
```

jeremystretch (changes requested):

jeremystretch:
Thanks @abhi1693!

PR #12491, opened 2023-05-04
12489 Use HTMX for Location and Non-Racked Devices in Site detail view

### LINKED ISSUE #12489: Limit amount of locations shown on site view
### NetBox version

3.5.0

### Feature type

### Proposed functionality

When opening a site there currently is no limitations in the locations table on the amount of locations shown. Similairly to the limitation of max 10 non-racked devices shown in the site view (implemented in #8920), there should also be a limit to 10 locations shown.

### Use case

On some sites we have 1k+ locations in NetBox. This creates issues like having to scroll for 10 seconds to get to images or the non-racked device table, plus increased unnecessary load times for the site page. 

### Database changes

_No response_

### External dependencies

_No response_

arthanson:
### Fixes: #12489 

Changes site view to use HTMX for Locations and Non-Racked Devices in Site detail view. See before and after screenshots below.  Also removed the now unused queries from the view.

Monosnap MDF | NetBox 2023-05-04 14-39-25
Monosnap MDF | NetBox 2023-05-04 14-39-56


netbox/templates/dcim/inc/nonracked_devices.html
jeremystretch, line 1: This template file is used by the rack and location views as well, so we can't make changes to it unless those views are also updated. Instead, we could just copy the new content directly into `site.html` and remove the include from that template.

arthanson: updated rack and location templates as well.

jeremystretch (changes requested):

netbox/templates/dcim/rack.html
```diff
@@ -190,14 +190,31 @@ <h4>Rear</h4>
...
   </div>
   <div class="row">
       <div class="col col-md-12">
-          {% plugin_full_width_page object %}
+        <div class="card">
+            <h5 class="card-header">
+                Non-Racked Devices
+            </h5>
+              <div class="card-body htmx-container table-responsive"
+                hx-get="{% url 'dcim:device_list' %}?rack_id={{ object.pk }}&position=null&parent_bay_id=null"
```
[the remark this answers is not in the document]
arthanson: went with adding has_position filter.

netbox/dcim/filtersets.py
```diff
@@ -909,6 +909,10 @@ class DeviceFilterSet(NetBoxModelFilterSet, TenancyFilterSet, ContactModelFilter
         queryset=Rack.objects.all(),
         label=_('Rack (ID)'),
     )
+    has_position = django_filters.BooleanFilter(
```
jeremystretch, line 912: We shouldn't need to introduce a new filter for this; the `position` filter should support a null (empty) lookup. I've opened #13585 to address this.

PR #12503, opened 2023-05-05
Adds parent device and bay position to table

### LINKED ISSUE #12223: Add support for a "parent device" column in device table views
### NetBox version

v3.1.7

### Feature type

### Proposed functionality

The table view for devices (e.g. https://&lt;NETBOX&gt;/dcim/devices) allows configuring a wide range of columns, but the parent device/device bay position is not one of them (the detail view of any given device shows this in the "Position" field, but the table view only allows for the "Position (U)" selection). This is to request that the column selections be extended to support inclusion of the parent-device/device-bay position of the devices.

### Use case

This is useful for seeing the location/relationship of a set of child devices to their parent devices at a glance, instead of having to select each device one at a time to ascertain that detail.

### Database changes

Nothing expected.

### External dependencies

None.

abhi1693:


jeremystretch:
Thanks @abhi1693!

PR #12504, opened 2023-05-05
Adds device type and role to device component filter

### LINKED ISSUE #12015: Add device type/role to Interface filter
### NetBox version

v3.4.4

### Feature type

New functionality

### Proposed functionality

Add the ability to filter interfaces by device types or device roles. I would like to review a set of devices that all fall under one device type or device role, and check whether a connection has been made to the interface. However, to do this, I have to search for each device individually and select every one into the filter to display what would be the same result if I could just filter by device type/role.

### Use case

The use case for this would be reviewing and checking interfaces easily for specific device types or roles. More specifically, the ability to see cables and their connections for device types and roles.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1219,6 +1219,28 @@ class DeviceComponentFilterSet(django_filters.FilterSet):
         to_field_name='name',
         label=_('Device (name)'),
     )
+    device_type_id = django_filters.ModelMultipleChoiceFilter(
+        field_name='device__device_type',
+        queryset=DeviceType.objects.all(),
+        label=_('Device type (ID)'),
+    )
+    device_type = django_filters.ModelMultipleChoiceFilter(
+        field_name='device__device_type__model',
+        queryset=DeviceType.objects.all(),
+        to_field_name='model',
+        label=_('Device type (model)'),
+    )
+    device_role_id = django_filters.ModelMultipleChoiceFilter(
+        field_name='device__device_role',
+        queryset=DeviceRole.objects.all(),
+        label=_('Device role (ID)'),
+    )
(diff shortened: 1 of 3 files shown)


netbox/dcim/forms/filtersets.py
```diff
@@ -1141,8 +1157,8 @@ class InterfaceFilterForm(PathEndpointFilterForm, DeviceComponentFilterForm):
         ('Addressing', ('vrf_id', 'l2vpn_id', 'mac_address', 'wwn')),
         ('PoE', ('poe_mode', 'poe_type')),
         ('Wireless', ('rf_role', 'rf_channel', 'rf_channel_width', 'tx_power')),
-        ('Device', ('region_id', 'site_group_id', 'site_id', 'location_id', 'rack_id', 'virtual_chassis_id',
-                    'device_id', 'vdc_id')),
+        ('Location', ('region_id', 'site_group_id', 'site_id', 'location_id', 'rack_id', 'virtual_chassis_id', 'vdc_id')),
```
jeremystretch, line 1160: Same as above for `vdc_id`.

netbox/dcim/filtersets.py
```diff
@@ -1219,6 +1219,28 @@ class DeviceComponentFilterSet(django_filters.FilterSet):
         to_field_name='name',
         label=_('Device (name)'),
     )
+    device_type_id = django_filters.ModelMultipleChoiceFilter(
```
jeremystretch, line 1222: Could you add filtersets tests for these under the component filterset test cases please?

jeremystretch (changes requested):

PR #12505, opened 2023-05-05
Adds tooltip on custom field

### LINKED ISSUE #12131: Custom Field information/hint button
### NetBox version

v3.4.7

### Feature type

New functionality

### Proposed functionality

In my opinion, an information description for a specific `custom field` is missing. If the `Custom Field` is very specific, an expandable information text would be very helpful.
The Custom Field `inBackup` serves as an example. In this case, a functional description can be quite helpful.

netbox-information
collapsed information button
netbox-information-open
folded out information button
netbox-information-edit
Custom Field setting

### Use case

Especially new and inexperienced NetBox users would get a better overview of the special functions and backgrounds of this custom field. Users with reading rights also have no possibility to view the description during the editing process or under the tab `Other`.

### Database changes

Nothing

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/templates/circuits/inc/circuit_termination.html
+++ b/netbox/templates/circuits/inc/circuit_termination.html
@@ -132,9 +132,16 @@
             </tr>
             {% for field, value in fields.items %}
               <tr>
-                <td>
-                  <span title="{{ field.description|escape }}">{{ field }}</span>
-                </td>
+                <th scope="row">{{ field }}
+                  {% if field.description %}
+                    <i
+                      class="mdi mdi-information text-primary"
+                      data-bs-toggle="tooltip"
+                      data-bs-placement="right"
+                      title="{{ field.description }}"
+                    ></i>
+                 {% endif %}
+                </th>
                 <td>
                   {% customfield_value field value %}
                 </td>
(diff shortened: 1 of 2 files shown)


netbox/templates/inc/panels/custom_fields.html
```diff
@@ -12,11 +12,16 @@ <h6>{{ group_name }}</h6>
           <table class="table table-hover attr-table">
             {% for field, value in fields.items %}
               <tr>
-                <th scope="row">
-                  <span title="{{ field.description|escape }}">{{ field }}</span>
-                </th>
```
jeremystretch, line 17: I think we still want to show the field name, but replace the `title` attribute with an icon within the same cell.

abhi1693: Could you please clarify a bit more? The field name is still showing up here

image
image

netbox/templates/circuits/inc/circuit_termination.html
```diff
@@ -132,9 +132,16 @@ <h5>Termination {{ side }}</h5>
...
-                <td>
-                  <span title="{{ field.description|escape }}">{{ field }}</span>
-                </td>
+                <th scope="row">{{ field }}
+                  {% if field.description %}
+                    <i
+                      class="mdi mdi-information text-primary"
+                      data-bs-toggle="tooltip"
+                      data-bs-placement="right"
+                      title="{{ field.description }}"
```
jeremystretch, line 141: The `escape` filter is still needed here AFAICT. Any reason to remove it?

PR #12506, opened 2023-05-05
Adds stroke to the reservation

### LINKED ISSUE #11900: Improve readability of rack reservation in elevations
### NetBox version

3.4.3

### Feature type

### Proposed functionality

Add some borders in SVG around rack reservations to better distinguish them (and maybe a sort of highlight on hover).
`path` tag can be used to solve this problem. For example, for the case described just after, this solution works : 

```HTML
<a xmlns="http://www.w3.org/2000/svg" target="_parent" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="xxx">
  <title>Reservation #14: CUSTOMER 1</title>
  <path d="M256,530 L271,530" style="stroke:  black;stroke-width: 2px;"/>
  <rect class="reservation" height="66" width="15" x="256" y="530"/>
  <path d="M256,596 L271,596" style="stroke:  black;stroke-width: 2px;"/>
</a>
<a xmlns="http://www.w3.org/2000/svg" target="_parent" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="xxx"> 
  <title>Reservation #13: CUSTOMER 2</title>
  <path d="M256,596 L271,596" style="stroke:  black;stroke-width: 2px;"/>
  <rect class="reservation" height="66" width="15" x="256" y="596"/>
  <path d="M256,662 L271,662" style="stroke:  black;stroke-width: 2px;"/>
</a>
```

Another idea (but I should maybe open another ticket for that) is to have a slighty different background on reservation representation when looking at it : 
image

### Use case

When there are many rack-reservation on a rack, elevation is not really readable because all reservations representations are sticked.
For example, in this one, there is one reservation between 13 and 15U and an other, just after, between 16 and 19U.

image

With borders (paths) :
image

### Database changes

None.

### External dependencies

_No response_

abhi1693:
### Fixes: #11900

Preview of how it looks
image

### DIFF
--- a/netbox/dcim/svg/racks.py
+++ b/netbox/dcim/svg/racks.py
@@ -252,7 +252,7 @@ class RackElevationSVG:
                 link = Hyperlink(href=f'{self.base_url}{reservation.get_absolute_url()}', target='_parent')
                 link.set_desc(f'Reservation #{reservation.pk}: {reservation.description}')
                 link.add(
-                    Rect(coords, size, class_='reservation')
+                    Rect(coords, size, class_='reservation', stroke='red', stroke_width=2)
                 )
                 self.drawing.add(link)
 
--- a/netbox/dcim/svg/racks.py
+++ b/netbox/dcim/svg/racks.py
@@ -22,6 +22,11 @@ __all__ = (
     'RackElevationSVG',
 )
 
+GRADIENT_RESERVED = '#b0b0ff'
+GRADIENT_OCCUPIED = '#d7d7d7'
+GRADIENT_BLOCKED = '#ffc0c0'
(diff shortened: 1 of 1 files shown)


abhi1693:
`0.1` subtraction may look wierd in the code but was needed to make the borders look symmetrical

image

jeremystretch:
I tweaked the stroke style and position a bit further to ensure that we're rendering the full 2px width on both sides, changed the stroke color to match the gradient, and added a bit of space between the rack border and the reservation block.

Screenshot 2023-05-09 at 09-57-24 Reservation for rack IDF128 NetBox

PR #12508, opened 2023-05-05
Disables map button when MAPS_URL is none

### LINKED ISSUE #12498: Disable the Map button
### NetBox version

v3.4.10

### Feature type

### Proposed functionality

I would like to be able to disable the "Map" button which is auto-generated on the physical address and GPS fields, for example by setting `MAPS_URL=None` (or empty string or false) - or if that doesn't work with inheritance of ConfigParam values, then to some sentinel value.

This would be a small change to netbox/templates/dcim/site.html in two places: e.g.

```
              {% if object.physical_address %}
```

becomes something like

```
              {% if object.physical_address and config.MAPS_URL %}
```

### Use case

See #12399: you may want to make a Custom Link which gives a better map reference (by combining multiple fields).

(Making the physical and GPS map links fully templatable, like Custom Links are, would be nice for this use case, but I'm just putting the simplest case of disabling the map button)

### Database changes

None

### External dependencies

None

abhi1693:

### DIFF
--- a/netbox/templates/dcim/site.html
+++ b/netbox/templates/dcim/site.html
@@ -87,11 +87,13 @@
             <th scope="row">Physical Address</th>
             <td class="position-relative">
               {% if object.physical_address %}
-                <div class="position-absolute top-50 end-0 translate-middle-y noprint">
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @abhi1693!

PR #12509, opened 2023-05-05
Adds dimensions card to device view

### LINKED ISSUE #12286: Add Total Weight/Dimensions to Device view
### NetBox version

v3.4.8

### Feature type

New functionality

### Proposed functionality

When #9654 was completed, the `total_weight` property was added to the `Device` and `Rack` models, but only the `Rack` model current displays it in the UI. I propose adding this to the view for a `Device` as well. Ideally, this would be in a separate *Dimensions* card, like under the `Rack` view, that could also include the Height from the device type, as well as width and depth values, if those are ever added.

### Use case

The `total_weight` field can already be accessed programmatically, so it makes sense to include it in the UI for visibility and there is already precedent for the `Rack` model.

### Database changes

None

### External dependencies

None

abhi1693:

### DIFF
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -300,6 +300,29 @@
             </div>
             {% include 'inc/panels/contacts.html' %}
             {% include 'inc/panels/image_attachments.html' %}
+            <div class="card">
+                <h5 class="card-header">Dimensions</h5>
+                <div class="card-body table-responsive">
+                    <table class="table table-hover attr-table">
+                        <tr>
+                            <th scope="row">Height</th>
+                            <td>
+                                {{ object.device_type.u_height }}U ({% if object.desc_units %}descending{% else %}ascending{% endif %})
+                            </td>
+                        </tr>
+                        <tr>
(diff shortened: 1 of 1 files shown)


netbox/templates/dcim/device.html
```diff
@@ -300,6 +300,29 @@ <h5 class="card-header">Services</h5>
...
             {% include 'inc/panels/contacts.html' %}
             {% include 'inc/panels/image_attachments.html' %}
+            <div class="card">
+                <h5 class="card-header">Dimensions</h5>
+                <div class="card-body table-responsive">
+                    <table class="table table-hover attr-table">
+                        <tr>
+                            <th scope="row">Height</th>
+                            <td>
+                                {{ object.device_type.u_height }}U ({% if object.desc_units %}descending{% else %}ascending{% endif %})
```
jeremystretch, line 310: I don't think the "descending" indicator makes sense here. (This was probably copied from the rack template by mistake.)

```suggestion
                                {{ object.device_type.u_height }}U
```

jeremystretch (changes requested):

PR #12522, opened 2023-05-08
Replaced device type weight with device total weight

### LINKED ISSUE #12286: Add Total Weight/Dimensions to Device view

abhi1693:
### Fixes: #12286

OP mentioned modules should also be included in the weight display.

### DIFF
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -313,7 +313,7 @@
                         <tr>
                             <th scope="row">Weight</th>
                             <td>
-                                {% if object.device_type.weight %}
+                                {% if object.total_weight %}
                                     {{ object.total_weight|floatformat }} {{ object.device_type.get_weight_unit_display }}
                                 {% else %}
                                     {{ ''|placeholder }}
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -314,7 +314,7 @@
                             <th scope="row">Weight</th>
                             <td>
                                 {% if object.total_weight %}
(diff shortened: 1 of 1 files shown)


netbox/templates/dcim/device.html
```diff
@@ -313,8 +313,8 @@ <h5 class="card-header">Dimensions</h5>
                         <tr>
                             <th scope="row">Weight</th>
                             <td>
-                                {% if object.device_type.weight %}
-                                    {{ object.device_type.weight|floatformat }} {{ object.device_type.get_weight_unit_display }}
+                                {% if object.total_weight %}
+                                    {{ object.total_weight|floatformat }} {{ object.device_type.get_weight_unit_display }}
```
jeremystretch, line 317: I believe the total weight will always be displayed in kilograms.
```suggestion
                                    {{ object.total_weight|floatformat }} Kilograms
```

jeremystretch (changes requested):

netbox/templates/dcim/device.html
abhi1693: Ok, I wasn't sure of that. I'll make this change now

PR #12527, opened 2023-05-08
Adds full_clean in examples

### LINKED ISSUE #11689: Missing `full_clean()` from custom script examples
### Change Type

Correction

### Area

Customization

### Proposed Changes

The sample script at https://docs.netbox.dev/en/stable/customization/custom-scripts/#example should add `site.full_clean()` and `switch.full_clean()` before the respective `save()` calls.

A similar issue exists for the nbshell examples at https://docs.netbox.dev/en/stable/administration/netbox-shell/#creating-and-updating-objects (but may need a slightly bigger change, since I don't think you can use a one-liner `blah(...).full_clean().save()`

abhi1693:
### Fixes: #11689

One question, since it's not possible to provide a 1-liner along with `full_clean`, should I remove this from the doc?

https://github.com/netbox-community/netbox/blob/develop/docs/administration/netbox-shell.md?plain=1#L159-L163

```
Alternatively, the above can be performed as a single operation. (Note, however, that `save()` does _not_ return the new instance for reuse.)

>>> VLAN(vid=123, name='MyNewVLAN', site=Site.objects.get(pk=7)).save()
```

### DIFF
--- a/docs/administration/netbox-shell.md
+++ b/docs/administration/netbox-shell.md
@@ -157,12 +157,6 @@ New objects can be created by instantiating the desired model, defining values f
 >>> myvlan.save()
 ```
 
-Alternatively, the above can be performed as a single operation. (Note, however, that `save()` does _not_ return the new instance for reuse.)
-
-```
->>> VLAN(vid=123, name='MyNewVLAN', site=Site.objects.get(pk=7)).save()
-```
-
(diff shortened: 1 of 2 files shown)


kkthxbye-code:
I think it would be fine to remove that to avoid confusion.

PR #12528, opened 2023-05-08
Adds CXP (100GE) under Ethernet (modular)

### LINKED ISSUE #12323: Add CXP 100GE Interface Type
### NetBox version

v3.4.8

### Feature type

### Proposed functionality

I see that the CXP form factor has been mentioned in a few issues over the years, most recently with respect to infiniband, but it does not seem to have ever been implemented for Ethernet. Therefore, I request the addition of `CXP (100GE)` under the `Ethernet (modular)` section of interface types.

### Use case

We have quite a few service cards in our network ([datasheet](https://www.cisco.com/c/en/us/products/collateral/optical-networking/ons-15454-series-multiservice-provisioning-platforms/data_sheet_c78-713296.html)) that utilize a CXP pluggable slot to support 100GE connectivity and we would like to be able to accurately model them.

### Database changes

None

### External dependencies

None

abhi1693:


jeremystretch:
Thanks @abhi1693!

PR #12551, opened 2023-05-10
Fixes #10686 - Import cables using VC master device

### LINKED ISSUE #10686: CSV upload doesn't respect chassis master as accessor for all interfaces in stack
### NetBox version

v3.3.5

### Python version

3.10

### Steps to Reproduce

1. Create a virtual chassis and two stack members
2. Ensure 1st stack member has interfaces 1/0/X
3. Ensure 2nd stack member has interfaces 2/0/X
4. Set 1st stack member to be primary for the virtual chassis
5. Use CSV import for Connections, and ensure one of the terminations goes to the 2/0/X, with the device selected as the 1st (primary) stack member
6. Don't profit

### Expected Behavior

In the UI, all interfaces in a stack are exposed by the primary member. We can manually create a connection from one of the other member's interfaces to a device, while viewing these interfaces as the primary.

### Observed Behavior

Import states that termination doesn't exist, which is sort of true since it "doesn't" exist on that switch.

DanSheps:
### Fixes: #10686

Allow importing cables using the master device as a target for the interface device

### DIFF
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -1078,7 +1078,11 @@ class CableImportForm(NetBoxModelImportForm):
 
         model = content_type.model_class()
         try:
-            termination_object = model.objects.get(device=device, name=name)
+            if device.virtual_chassis and device.virtual_chassis.master == device and \
+                    model.objects.filter(device=device, name=name).count() == 0:
+                termination_object = model.objects.get(device__in=device.virtual_chassis.members.all(), name=name)
+            else:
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):
Could you extend the existing import test to also evaluate this please?

DanSheps:
I have no idea why the PR request is failing...  Going to try and re-run it

jeremystretch:
> I have no idea why the PR request is failing... Going to try and re-run it

There were widespread issues with GitHub Actions yesterday.

PR #12575, opened 2023-05-12
Adding Abnormal choice to IP Address

apellini:
Could be useful to identify IPs that are alive but you don't recognize in CMDB


abhi1693:

PR #12584, opened 2023-05-12
Adds copy content button

### LINKED ISSUE #12499: Copy-to-clipboard functionality for IPs
### NetBox version

v3.5.0

### Feature type

New functionality

### Proposed functionality

It would be great if there would be a function that when you hover over the IP addresses with the mouse a copy button appears which copies the IP without subnet mask.

Copy button like:

Screenshot 2023-05-05 at 11 27 40

### Use case

I think it would be a relief for most users, at the moment you always have to tediously mark the IP and then copy it to connect to a system e.g. via SSH.

### Database changes

-

### External dependencies

-

abhi1693:
### Fixes: #12499

I have created a new `copy_content` template that I have used in several places to provide the copy to clipboard functionality. I have also refactored `copy-token` code to utlize this new template.


netbox/utilities/templates/builtins/copy_content.html
```diff
@@ -0,0 +1,3 @@
+<a class="btn btn-sm {{ color }} copy-content" data-clipboard-target="{{ target }}" title="Copy to clipboard">
+  <i class="mdi mdi-content-copy"></i>
+</a>
```
bluikko, line 3: nit: happened to notice that this change seems to make the file miss the final newline?

jeremystretch:
I wonder if it might be a more scalable approach to introduce a CSS class that display a "copy" button on hover, appended to the element? That would obviate the need for any substantial changes to the underlying tables or templates. It should be feasible, but I'll need to track down an example.

jeremystretch:
Thanks @abhi1693!

PR #12593, opened 2023-05-12
Fixes incorrectly handled type error when list of objects is found in data

### LINKED ISSUE #9876: Improve support for referencing tags in webhook conditions

abhi1693:

### DIFF
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -67,7 +67,7 @@ class Condition:
         """
         def _get(obj, key):
             if isinstance(obj, list):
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @abhi1693!

PR #12618, opened 2023-05-15
Closes #9068: validate addresses assigned to interfaces

### LINKED ISSUE #9068: Validation of IP Addresses when of type network or broadcast
### NetBox version

v3.1.11

### Feature type

New functionality

### Proposed functionality

One of the many ways we use netbox is to assign IP addresses to device interfaces, which are used to push config changes to device.

Currently, one can add any ip address, including the network or broadcast address, to an interface. This is not a valid config from any devices standpoint. There are some obvious exceptions to this:

* /31 - rfc3021
* /32 - In our case, we set these to Role Loopback
* /127 - rfc6164

Are there other use cases other than above where an IP object on a broadcast or network address should be added? 

I would propose that some sort of validation happens when 'invalid' IP addresses are added to netbox. It's not entirely clear what the best validation here is but a few possibilities would be:

* don't accept the IP
* throw a big scary warning, but accept the IP

One could also consider a config knob to disable this sanity check and revert to current behavour.

It sounds like there may be a way to handle this via a custom validator, but it seems to be enough of a common issue that it seems worth discussion built in functionality. 

There is some discussion about this in [this slack thread.](https://netdev-community.slack.com/archives/C01P0FRSXRV/p1649339405458799)

### Use case

Should help with user friendliness of end users using netbox, preventing humans clumsy entering invalid data.

### Database changes

_No response_

### External dependencies

_No response_

decoupca:
### Fixes: #9068

Disallows assigning a broadcast address or network ID to an interface.

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -350,7 +350,7 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
             self.add_error(
                 'primary_for_parent', "Only IP addresses assigned to an interface can be designated as primary IPs."
             )
-        
+
         # Do not allow assigning a network ID or broadcast address to an interface.
         if interface:
             if address := self.cleaned_data.get('address'):
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -352,19 +352,16 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
             )
 
         # Do not allow assigning a network ID or broadcast address to an interface.
-        if interface:
-            if address := self.cleaned_data.get('address'):
-                if address.ip == address.network:
-                    msg = f"{address} is a network ID, which may not be assigned to an interface."
-                    if address.version == 4:
-                        if address.prefixlen not in (31, 32):
-                            raise ValidationError(msg)
-                    if address.version == 6:
-                        if address.prefixlen not in (127, 128):
-                            raise ValidationError(msg)
-                if address.ip == address.broadcast:
-                    msg = f"{address} is a broadcast address, which may not be assigned to an interface."
+        if interface and (address := self.cleaned_data.get('address')):
+            if address.ip == address.network:
+                msg = f"{address} is a network ID, which may not be assigned to an interface."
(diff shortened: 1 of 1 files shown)


netbox/ipam/forms/model_forms.py
```diff
@@ -351,6 +351,18 @@ def clean(self):
                 'primary_for_parent', "Only IP addresses assigned to an interface can be designated as primary IPs."
             )
 
+        # Do not allow assigning a network ID or broadcast address to an interface.
+        if interface:
+            if address := self.cleaned_data.get('address'):
+                if address.ip == address.network and address.prefixlen not in (31, 32, 127, 128):
+                    self.add_error(
```
jeremystretch, line 358: Validation failures should raise `ValidationError` by convention. (I'm not sure why the check above is calling `add_error()`.

netbox/ipam/forms/model_forms.py
```diff
@@ -351,6 +351,18 @@ def clean(self):
                 'primary_for_parent', "Only IP addresses assigned to an interface can be designated as primary IPs."
             )
 
+        # Do not allow assigning a network ID or broadcast address to an interface.
+        if interface:
+            if address := self.cleaned_data.get('address'):
+                if address.ip == address.network and address.prefixlen not in (31, 32, 127, 128):
```
jeremystretch, line 357: For the sake of completeness, we should split this into separate checks for IPv4 and IPv6 addresses.

jeremystretch (changes requested):
Thanks for taking this!

netbox/ipam/forms/model_forms.py
decoupca: Done!

kkthxbye-code:
Doing validation like this in the form has the side-effect of only applying to the UI. This is an issue with large parts of the netbox codebase, but I think we should try to avoid it if possible.

With the PR it is now possible to create 10.0.0.0/24 in the API and assign it to an interface. When editing the new IP and saving without changes it gets the validation error. Not sure what the right solution is. Either the validation logic has to be in both the model form and the serializer, or the validation needs to be done on the model.

@jeremystretch - Any input? I would classify this fix as being buggy, as it makes the UI and API behave differently.

Issue #6347, opened 2021-05-04
Cache the number of each component type assigned to devices/VMs


jeremystretch:
### Proposed Changes

Add the following integer fields to dcim.Device:

* `_console_port_count`
* `_console_server_port_count`
* `_power_port_count`
* `_power_outlet_count`
* `_interface_count`
* `_front_port_count`
* `_rear_port_count`
* `_device_bay_count`
* `_inventory_item_count`

Add the following integer fields to virtualization.VirtualMachine:

* `_interface_count`

These fields would be automatically populated upon creating/modifying/deleting the relevant child components.

### Justification

* Enables efficient filters of devices/VMs based on whether a certain type of component has been created.
* Eliminates the need for additional queries e.g. when display the device view.

jeremystretch:
This might also be a use case for CTE; see #6587.

AndMrzv:
I also like to verify that devices match their device type templates (you know something's up if a 24port 1PSU switch suddenly has 29 interfaces and 3 power ports). So maybe you'll find it useful to add same data to device types, maybe expand the idea to strict matching devices to their types or syncing (now I do this with API and Bash).

ryanmerolle:
pushed to 3.6 milestone

PR #12632, opened 2023-05-16
6347 Cache the number of each component type assigned to devices/VMs

arthanson:
### Fixes: #6347 

Adds counter fields to cache related model counts.  Does this by adding a CachedCounterField type that hooks up post-save and post-delete signals on the related model, these hooked up by calling connect_counter in apps.py

also has change tracking, there is a TrackingModelMixin which I added to the base classes - by default it does nothing but if you set change_tracking_fields on the model to the fields to track it will do change tracking on them.  The connect_counter method automatically does the correct fields (generally device_id) and then checks this on post_save signal  to see if it has been moved and updates the counts on the old and new parent models appropriately.

The change tracking is probably useful elsewhere - you can check model.tracker.changed to see any changed fields.  Note: these fields must be listed in change_tracking_fields.


kkthxbye-code:
I know this is a draft, so apologies if this is premature. Looking at the code it doesn't look like it currently take into account that InventoryItems can be moved between devices.

It might also be an idea to add a management command to re-calculate counts, if they ever get out of sync.

arthanson:
@kkthxbye-code could definitely use a more thorough review on this one, any feedback welcome.

netbox/utilities/counter.py
```diff
@@ -0,0 +1,91 @@
...
+
+        self.connect()
+
+    def validate(self):
+        counter_field, _, _, _ = self.parent_model._meta.get_field_by_name(self.counter_name)
+        if not isinstance(counter_field, CounterCacheField):
+            raise TypeError(
+                f"{self.counter_name} should be a CounterCacheField on {self.parent_model}, but is {type(counter_field)}"
+            )
+        if not isinstance(parent_model, TrackingModelMixin):
```
jeremystretch, line 38:
```suggestion
        if not isinstance(self.parent_model, TrackingModelMixin):
```

netbox/utilities/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+class Tracker(object):
+    def __init__(self, instance):
+        self.instance = instance
+        self.newly_created = False
+        self.changed = {}
+        self.tracked_fields = self.instance.change_tracking_fields
+
+
+class TrackingModelMixin(object):
+    change_tracking_fields = []
```
jeremystretch, line 13: Because this is declared as a class attribute, the list is being shared across _all_ models:

```python
>>> ConsolePort.change_tracking_fields
['device_id', 'virtual_machine_id']
```

Instead of defining this on the model, could we create a registry entry for mapping models to their tracked fields? This would probably be a bit cleaner.

netbox/utilities/fields.py
```diff
@@ -143,3 +143,12 @@ def gfk_key(obj):
             self.name,
             False,
         )
+
+
+class CounterCacheField(models.BigIntegerField):
```
jeremystretch, line 148: Need to add this to `__all__`

netbox/utilities/mixins.py
arthanson: made an instance var

netbox/utilities/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+        self.change_tracking_fields = []
+        self._initialized = True
+
+    @property
+    def tracker(self):
+        if hasattr(self._state, "_tracker"):
+            tracker = self._state._tracker
+        else:
+            tracker = self._state._tracker = Tracker(self)
+        return tracker
```
jeremystretch, line 25:
```suggestion
        if not hasattr(self._state, "_tracker"):
            self._state._tracker = Tracker(self)
        return self._state._tracker
```

netbox/utilities/mixins.py
```diff
@@ -0,0 +1,54 @@
+from django.db.models.query_utils import DeferredAttribute
+
+
+class Tracker(object):
+    def __init__(self, instance):
+        self.instance = instance
+        self.newly_created = False
```
jeremystretch, line 7: Do we need the `newly_created` attribute? It doesn't appear to ever be read.

netbox/utilities/mixins.py
```diff
@@ -0,0 +1,54 @@
...
+        self.changed = {}
+        self.tracked_fields = self.instance.change_tracking_fields
```
jeremystretch, line 9: This seems redundant. Won't the value of a Tracker instance's `tracked_fields` attribute always be equal to the value of `change_tracking_fields` on its instance?

netbox/utilities/management/commands/calculate_cached_counts.py
```diff
@@ -0,0 +1,36 @@
+from django.core.management.base import BaseCommand
+from django.core.management.color import no_style
+
+from dcim.models import Device
+from virtualization.models import VirtualMachine
+
+
+def recalculate_device_counts():
```
jeremystretch, line 8: Ideally we'd have some mechanism to discover all the necessary fields, rather than defining them explicitly. Again, this is where utilizing the registry can help.

jeremystretch (changes requested):
Excellent work! Just working on simplifying and cleaning things up a bit at this point. Also, could you add some simple tests for the counters? A single relationship (e.g. device interfaces) should be sufficient.

arthanson:
@jeremystretch I added auto-discovery and cleaned up how you register / connect the fields, now you just supply it as params to the field:
```
    _interface_count = CounterCacheField(to_model='virtualization.VMInterface', to_field='virtual_machine')
```
and in apps.py ready function call:
```
        connect_counters(self)
```
I think this solves your issues with auto-discovery and the app registry?  Please let me know if any additional changes.

It would be nice if there was a global app init, then could just run through all models and wouldn't have to remember to put connect_counters in each application apps.py.

Should I document using this in the models doc?  Sort of an internal thing so not sure if this should go into the docs, but does effect people if they want to use this and add a field that should be cached.

PR #12708, opened 2023-05-25
Fix #12690 - Add Cluster/ClusterGroup Filtering to VLANGroup QuerySet

### LINKED ISSUE #12690: From the GUI we can't assign vlans from a VLAN Group with Virtualization Cluster as scope (works with the API)
### NetBox version

v3.3.10

### Python version

3.9

### Steps to Reproduce

0. Create a device in a site. <-- Will be referred to "the device" from now on.
1. Create a Virtualization Cluster with at least the device.
2a. Create in the IPAM a VLAN Group attached to the site of the device (created in step 0).
2b. Create in the IPAM a VLAN Group attached to the just created Virtualization Cluster.
3. Create at least 1 vlan in that newly created VLAN Groups (site and cluster based).
4a. From the web GUI try to attach this Virtualization Cluster VLAN (from step 3) to an interface on a device in the group (from step 1). <-- The vlan don't show up in the drop down list.
4b. From the API, we can assign the vlan that is impossible to assign from the GUI.

### Expected Behavior

The WEB Gui dropdown list should pull from the appropriate VLAN group. It looks like the the list that is pulled by default is the "site" vlan group and we can't override the behaviour.

### Observed Behavior

When using the API, this limitation si not present. In fact, since it's working the WEB Gui display the _correct_ vlan until we try to edit the interface, then when we select the vlans, the Virtualization vlan disappear from the list and it is replaced by the list of vlans from the site group instead.

stuntguy3000:
### Fixes: #12690 

This changes adds additional filtering to the VLANGroup QuerySet to account for Cluster and Cluster Group scopes - enabling them to be returned by the `/api/ipam/vlans/?available_on_device` endpoint.

This has been validated against various test cases.

### DIFF
--- a/netbox/ipam/querysets.py
+++ b/netbox/ipam/querysets.py
@@ -64,6 +64,16 @@ class VLANQuerySet(RestrictedQuerySet):
                 scope_type=ContentType.objects.get_by_natural_key('dcim', 'rack'),
                 scope_id=device.rack_id
             )
+        if device.cluster:
+            q |= Q(
+                scope_type=ContentType.objects.get_by_natural_key('virtualization', 'cluster'),
+                scope_id=device.cluster.id
+            )
+            if device.cluster.group:
+                q |= Q(
+                    scope_type=ContentType.objects.get_by_natural_key('virtualization', 'clustergroup'),
+                    scope_id=device.cluster.group.id
+                )
 
         # Return all applicable VLANs
         return self.filter(


abhi1693:
Thank you for your interest in contributing to NetBox, however, it appears there is no accepted issue that correlates to this pull request. Before pull requests are opened, we require an accepted issue as per our [contributing guide](https://github.com/netbox-community/netbox/blob/master/CONTRIBUTING.md). Please wait for the issue to be accepted before further work is done on this pull request.

PR #12718, opened 2023-05-25
DRAFT: 12552 replace mptt

arthanson:


netbox/wireless/migrations/0009_remove_wirelesslangroup_level_and_more.py
jeremystretch, line 1: Rename migration file

netbox/dcim/models/device_components.py
```diff
@@ -1136,7 +1128,7 @@ class InventoryItem(MPTTModel, ComponentModel):
     clone_fields = ('device', 'parent', 'role', 'manufacturer', 'part_id',)
 
     class Meta:
-        ordering = ('device__id', 'parent__id', '_name')
```
jeremystretch, line 1139: We can't arbitrarily drop ordering by parent Device, as it has ramifications unrelated to the tree hierarchy; for example, the order of results returned by the REST API endpoint.

jeremystretch (changes requested):

jeremystretch:
I might be missing something, but it seems like there are some pieces still missing from the implementation.

Screenshot 2023-06-22 at 15-02-04 Regions NetBox

Some observations, using the Region model as an example:

1. Regions are not implicitly ordered deterministically by name.
2. Attempting to explicitly order regions by name (by clicking the table column header) appears to have no effect.
3. Top-level regions do not display a cumulative count of all descendant regions. For example, the North America region should show 24 sites assigned.

netbox/dcim/models/device_components.py
arthanson: @jeremystretch can you provide the repo scenarios that this would effect?  CTE can only order on a single field (technical limitation) if we need to do two I'm not sure how to work around that.

arthanson:
Sorting tables does seem to be broken, checking...

jeremystretch:
@arthanson given the limitations of the CTE approach we discussed, can this be closed out?

PR #12721, opened 2023-05-25
Fixes return_url for image attachment

### LINKED ISSUE #12538: Delete image attachment on device (type) should return to device not to image attachments list
### NetBox version

v3.5.1

### Feature type

### Proposed functionality

When deleting an attached image (picture) from a detail view of a device or device type you should return to the device / device type from which you started this.
A.t.m. you are coming to the list of image attachments and need to navigate back to the device.

### Use case

Usability improvement as it is most likely that you are in the context of a specific device / device type you are maintaining and not the list of all image attachments.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:
### Fixes: #12538

I have added support for parsing `return_url` from the request in the table if it exists. The old functionality is still retained if the parameter is not provided. I could not find any other clean way of redirecting the user back to where the request was initiated.

### DIFF
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -235,7 +235,7 @@ class ActionsColumn(tables.Column):
 
         model = table.Meta.model
         request = getattr(table, 'context', {}).get('request')
-        return_url = request.GET.get('return_url', request.get_full_path())
+        return_url = request.GET.get('return_url', request.get_full_path() if request else '')
         url_appendix = f'?return_url={quote(return_url)}' if return_url else ''
 
         html = ''
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -236,9 +236,9 @@ class ActionsColumn(tables.Column):
         model = table.Meta.model
         if request := getattr(table, 'context', {}).get('request'):
             return_url = request.GET.get('return_url', request.get_full_path())
+            url_appendix = f'?return_url={quote(return_url)}'
         else:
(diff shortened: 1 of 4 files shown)


netbox/netbox/tables/columns.py
```diff
@@ -235,7 +235,12 @@ def render(self, record, table, **kwargs):
 
         model = table.Meta.model
         request = getattr(table, 'context', {}).get('request')
-        url_appendix = f'?return_url={quote(request.get_full_path())}' if request else ''
+        query_params = request.GET.copy() if request else {}
+        if return_url := query_params.pop('return_url', None):
```
jeremystretch, line 239: I could be missing something but this would this not be sufficient?

```suggestion
        if return_url := request.GET.get('return_url'):
```

jeremystretch: We can probably simplify this to something like:

```python
return_url = request.GET.get('return_url', request.get_full_path())
url_appendix = f'?return_url={quote(return_url)}' if return_url else ''
```

jeremystretch (changes requested):

netbox/netbox/tables/columns.py
abhi1693: > I could be missing something but this would this not be sufficient?

I tested this using services on device and that worked just it was before. Then did the same test using image attachment and post edit and delete, I landed up on the device page.

While implementing this, I was thinking the same solution should be implemented application wide though.

If you can think of a use-case, that I ay have missed let know.

PR #12723, opened 2023-05-25
Adds hide-if-unset to custom field

### LINKED ISSUE #12597: Custom Field UI visibility option "Hidden if unset"
### NetBox version

v3.5.1

### Feature type

### Proposed functionality

On the custom-fields edit page, add "Hidden if unset" option to the "UI visibility" selection, and on the object view page, if that's set on definition of a custom field _and_ the field is unset hide that custom field (instead of showing `-`).

### Use case

We have devices managed by a few groups with different needs, and we're using custom fields to help with some of that. This would keep the display of a device (or virtual-machine) cleaner when custom fields aren't used on a specific device.

Use cases are similar to #12423, but I suspect the code would be much simpler.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/netbox/models/features.py
+++ b/netbox/netbox/models/features.py
@@ -197,11 +197,15 @@ class CustomFieldsMixin(models.Model):
         data = {}
 
         for field in CustomField.objects.get_for_model(self):
-            # Skip fields that are hidden if 'omit_hidden' is set
-            if omit_hidden and field.ui_visibility == CustomFieldVisibilityChoices.VISIBILITY_HIDDEN:
-                continue
-
             value = self.custom_field_data.get(field.name)
+
+            # Skip fields that are hidden if 'omit_hidden' is set
+            if omit_hidden:
+                if field.ui_visibility == CustomFieldVisibilityChoices.VISIBILITY_HIDDEN:
+                    continue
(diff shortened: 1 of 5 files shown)


jeremystretch:
I think it would be cleaner to tackle the conditional logic under [`get_custom_fields_by_group()`](https://github.com/netbox-community/netbox/blob/5a5fcf7d377d61e114a738c59edb9e015f465efe/netbox/netbox/models/features.py#L209) since we're already handling the omission of fully hidden fields there. (Plus it avoid having to deal with the duplicated template code.)

abhi1693:
@jeremystretch I have updated the code as per your suggestion

netbox/netbox/models/features.py
```diff
@@ -227,6 +231,8 @@ def get_custom_fields_by_group(self):
 
         for cf in visible_custom_fields:
             value = self.custom_field_data.get(cf.name)
+            if not value and cf.ui_visibility == CustomFieldVisibilityChoices.VISIBILITY_HIDDEN_IFUNSET:
```
ITJamie, line 234: this hides boolean custom fields where the value is set to false.

this works and shows boolean fields that are set to false
``` python
if value == None and cf.ui_visibility == CustomFieldVisibilityChoices.VISIBILITY_HIDDEN_IFUNSET:
```

PR #12724, opened 2023-05-25
Changes render config card with accordian

### LINKED ISSUE #12470: Make "Context Data" collapsed by default on "Render config" view
### NetBox version

v3.5.0

### Feature type

### Proposed functionality

Looking at the "Render Config" tab always show first the "Context Data", this can be quite big. So i always need to scroll down very far to see the rendered config. I think it would be more user friendly if the "Context Data" would be collapsed by default.

### Use case

Getting faster to rendered config without need to scroll for a long time.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:
### Fixes: #12470

Collapsed view
image

Open view
image

### DIFF
--- a/netbox/templates/dcim/device/render_config.html
+++ b/netbox/templates/dcim/device/render_config.html
@@ -28,10 +28,24 @@
     </div>
     <div class="col-7">
       <div class="card">
-        <h5 class="card-header">Context Data</h5>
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @abhi1693!

PR #12725, opened 2023-05-25
Adds ip to failed logs

### LINKED ISSUE #12562: Add IP address to auth log output
### NetBox version

v3.5.1

### Feature type

### Proposed functionality

Change the Netbox logger (netbox.auth.login) for authentication failure to include the remote IP address, this.

### Use case

The change would allow for parsing of the logs easier to detect brute force authentication attacks (i.e. Fail2Ban etc).

`May 11 12:58:20 netbox NetBox: level=INFO,logger=netbox.auth.login,msg=Failed login attempt for username: test-fail`
to
`May 11 12:58:20 netbox NetBox: level=INFO,logger=netbox.auth.login,msg=Failed login attempt for username: test-fail from x.y.1.1`

### Database changes

NA.

### External dependencies

This may have to use the X-Real-IP header?

abhi1693:
### Fixes: #12562

Should I raise an exception if `client_ip=None`?

### DIFF
--- a/netbox/users/signals.py
+++ b/netbox/users/signals.py
@@ -1,10 +1,11 @@
 import logging
 from django.dispatch import receiver
 from django.contrib.auth.signals import user_login_failed
+from utilities.request import get_client_ip
 
 
 @receiver(user_login_failed)
 def log_user_login_failed(sender, credentials, request, **kwargs):
     logger = logging.getLogger('netbox.auth.login')
     username = credentials.get("username")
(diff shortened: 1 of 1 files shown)


jeremystretch:
I don't think an exception is necessary, but in the even the client IP cannot be determined (e.g. due to omitted HTTP headers), we should report that in the log message.

abhi1693:
@jeremystretch I have added a separate warning now to let the user know about a possible mis-configuration

PR #12726, opened 2023-05-25
Adds filter query parameter on dashboard widget for ObjectCountWidget

### LINKED ISSUE #12599: Carry filters over when clicking on object list in dashboard widget
### NetBox version

v3.5.1

### Feature type

### Proposed functionality

I have an "Object Count" widget with a Devices/Cables/Racks/Sites and tenant filter `"{"tenant_id": "19"}`. I'd like if I click (for example) on "Devices" that I get `/dcim/devices/?tenant_id=19` instead of `/dcim/devices/`.

### Use case

Makes widgets with filters more useful if clicking on something gives a page with the actual items in the count instead of _all_ objects of that type. And can make the dashboard widgets a handy way to get to what that user needs to see.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/extras/dashboard/widgets.py
+++ b/netbox/extras/dashboard/widgets.py
@@ -1,6 +1,7 @@
 import uuid
 from functools import cached_property
 from hashlib import sha256
+from itertools import chain
 from urllib.parse import urlencode
 
 import feedparser
@@ -121,6 +122,14 @@ class DashboardWidget:
             'config': self.config,
         }
 
+    def filter_query(self):
+        query_string = ''
+        if filters := self.config.get('filters', None):
+            filter_items = [(k, v) if not isinstance(v, list) else zip([k] * len(v), v) for k, v in filters.items()]
+            query_string = urlencode(filter_items, doseq=True)
+
+        return query_string
+
 
 @register_widget
 class NoteWidget(DashboardWidget):
@@ -182,6 +191,7 @@ class ObjectCountsWidget(DashboardWidget):
 
         return render_to_string(self.template_name, {
             'counts': counts,
+            'filters': self.filter_query(),
         })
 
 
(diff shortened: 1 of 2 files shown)


kkthxbye-code:
I'm not sure about this one. I get the intention, but the ObjectCountWidget filters are queryset filters. Blindly converting these to URL FilterSet filters seems like a bad idea. It will work in some cases, but there's a lot where it wont.

Example (IPAM > Prefixes):

{
    {"tenant__name": "Dunder-Mifflin, Inc."}
}

Resulting URL:

`http://127.0.0.1:8000/ipam/prefixes/?tenant__name=Dunder-Mifflin%2C+Inc.`

`tenant__name` is of course not an actual FilterSet filter on PrefixFilterSet so it does nothing.

I'm not sure there is a clean solution here, QuerySet filters and FilterSet filters will never be compatible, so we would need to switch to using the API, however there's no "count" API endpoints to my knowledge.

abhi1693:
Good catch, I had completely forgotten that this runs different. The other solution (not preferable) is to split at `__` and query the object ID to transform the query params into IDs. That may have bad performance overall though.

jeremystretch:
I've opened #12742 to address this, which IMO should be treated as a bug as it can be very confusing to end users.

abhi1693:
Closing in favour of the new approach discussed here.

jeremystretch:
Well, I think this PR is still needed, and will suffice once we've solved for #12742.

netbox/extras/dashboard/widgets.py
```diff
@@ -180,8 +181,14 @@ def render(self, request):
             else:
                 counts.append((model, None))
 
+        filter_query_string = ''
+        if filters := self.config.get('filters', None):
+            filter_items = [(k, v) if not isinstance(v, list) else zip([k] * len(v), v) for k, v in filters.items()]
+            filter_query_string = urlencode(filter_items, doseq=True)
```
jeremystretch, line 187: We can leverage Django's QueryDict for this. Something like:

```
qd = QueryDict(mutable=True)
qd.update(filters)
filter_query_string = qd.urlencode()
```

PR #12743, opened 2023-05-26
12625 hide datasource password

### LINKED ISSUE #12625: Datasource passwords are displayed in plaintext
### NetBox version

v3.5.1

### Python version

3.8

### Steps to Reproduce

1. Create a new Datasource with a token or user/password
2. Open `core/data-sources/<ID>/`
3. Password is displayed in plaintext and also visible for readonly users

### Expected Behavior

Password should be hidden.

### Observed Behavior

Password is displayed as plaintext for all users with access to datasources.

aroth01:
### Fixes: #12625 

Change datasource.html to hide password until the "show secret" button is clicked. I have the permissions to view set for update/change/add instead of view. As if a user doesn't have view for core/datasource, they wouldn't be able to see the page anyways. Unless I'm misunderstanding something.

### DIFF
--- a/netbox/templates/core/datasource.html
+++ b/netbox/templates/core/datasource.html
@@ -89,11 +89,13 @@
             {% for name, field in object.get_backend.parameters.items %}
               <tr>
                 <th scope="row">{{ field.label }}</th>
-                {% if field.label == "Password" %}
+                {% if field.label == "Password" and object.parameters|get_key:name %}
                   <td>
                     {% if request.user|can_add:object or request.user|can_change:object or request.user|can_delete:object %}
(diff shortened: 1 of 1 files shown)


netbox/templates/core/datasource.html
```diff
@@ -88,7 +89,18 @@ <h5 class="card-header">Backend</h5>
             {% for name, field in object.get_backend.parameters.items %}
               <tr>
                 <th scope="row">{{ field.label }}</th>
-                <td>{{ object.parameters|get_key:name|placeholder }}</td>
+                {% if field.label == "Password" and object.parameters|get_key:name %}
```
jeremystretch, line 92: I don't think this is going to be a viable approach. We can't rely on the field label to infer whether the parameter should be displayed. (And even if we could, hard-coding this in the template should be considered poor practice.) We should devise some approach to explicitly set some attribute on the parameter itself to control its display.

PR #12755, opened 2023-05-30
Fixes #12745 - Escape text passed as display values to slim-select

### LINKED ISSUE #12745: XSS in DynamicModelChoiceField
### NetBox version

v3.5.2

### Python version

3.10

### Steps to Reproduce

1. Create a tenant group with the name ``
2. Go to the site creation form
3. Open the tenant group dropdown

### Expected Behavior

No XSS

### Observed Behavior

The name is not escaped and the browser displays an alert with the cookie. Originally found here: https://github.com/anhdq201/netbox/issues

This probably applies to all DynamicModelChoiceField where the user can set the display value.

The person responsible for that repo chose not to report it. The same report is also duplicated 15 times and one report is invalid (the graphql one).

kkthxbye-code:
### Fixes: #12745

Adds js dependencies html-entities (yes, there really is no built-in method for this is js) and escapes display text passed to slim-select.


kkthxbye-code:
Should fix all 15 of these CVE's: https://github.com/anhdq201/netbox/issues however there might be other places where we pass unescaped HTML from the API to a library. Ideally I believe slim-select should do this escaping, but slim-select 1.x is deprecated and seemingly completely abandoned. slim-select 2.x is a complete rewrite and not compatible at all.

I've requested that MITRE fold the 15 CVE's into one CVE as their rules state that they must be:

https://www.cve.org/ResourcesSupport/AllResources/CNARules

Section 7.2

>The CVE Program expects separate CVE IDs to be assigned to
**independently fixable vulnerabilities**. If one vulnerability can be
fixed without fixing the other, then the vulnerabilities should
receive separate CVE IDs.

As these are not independently fixable, one CVE should have been created instead of 15.

jeremystretch:
Thanks for digging into this @kkthxbye-code!

PR #12778, opened 2023-05-31
12175 rack with starting unit > 1

arthanson:


arthanson:
@jeremystretch this one is blocked on https://github.com/netbox-community/netbox/pull/12632 as I'd like to use the TrackingModelMixin from that PR in order to track changes to the starting_unit as it needs to update all devices in the rack if that is changed.

netbox/dcim/models/racks.py
```diff
@@ -235,13 +240,26 @@ def clean(self):
...
                         'u_height': "Rack must be at least {}U tall to house currently installed devices.".format(
                             min_height
                         )
                     })
+            # Validate the Rack starting_unit > position of any installed devices
+            first_device = Device.objects.filter(
+                rack=self
+            ).exclude(
+                position__isnull=True
+            ).order_by('position').first()
```
jeremystretch, line 255: We should be able to reuse the above queryset and just call `first()` and `last()` to get the top and bottom devices, respectively.

netbox/dcim/models/racks.py
```diff
@@ -235,13 +240,26 @@ def clean(self):
...
             if top_device:
-                min_height = top_device.position + top_device.device_type.u_height - 1
+                min_height = top_device.position + top_device.device_type.u_height - self.starting_unit
                 if self.u_height < min_height:
                     raise ValidationError({
                         'u_height': "Rack must be at least {}U tall to house currently installed devices.".format(
                             min_height
                         )
                     })
+            # Validate the Rack starting_unit > position of any installed devices
```
jeremystretch, line 250:
```suggestion
            # Validate that the Rack's starting_unit is <= the position of the lowest mounted device (if any)
```

netbox/dcim/models/racks.py
```diff
@@ -269,8 +287,8 @@ def units(self):
         Return a list of unit numbers, top to bottom.
         """
         if self.desc_units:
-            return drange(decimal.Decimal(1.0), self.u_height + 1, 0.5)
-        return drange(self.u_height + decimal.Decimal(0.5), 0.5, -0.5)
+            return drange(decimal.Decimal(self.starting_unit), self.u_height + 1, 0.5)
```
jeremystretch, line 290: This fixes the unit list for racks with descending unit numbering.

```suggestion
            return drange(decimal.Decimal(self.starting_unit), self.u_height + self.starting_unit, 0.5)
```

jeremystretch (changes requested):
Nice work!

Issue #9077, opened 2022-04-07
Set `alters_data=True` on model methods as appropriate


jeremystretch:
### Proposed Changes

Audit the code base and set `alters_data=True` on all model methods which alter object data.

### Justification

The [Django documentation](https://docs.djangoproject.com/en/4.0/ref/templates/api/#variables-and-lookups) recommends this to avoid accidentally calling potentially destructive methods when rendering templates.

PR #12781, opened 2023-05-31
9077 audit alters_data=True

arthanson:
### Fixes: #9077 

adds alters_data=True to model functions with no params that modify data also sets some do_not_call_in_templates for several that should just be skipped.

Tested by manually setting string_if_invalid in settings and temporarily calling the method in a template to make sure it was skipped. easiest one is adding a call to sync_data in exporttemplate.html as it will throw an Attribute Error if the alters_data setting is removed.


jeremystretch:
Thanks @arthanson!

PR #12782, opened 2023-05-31
11305 Add GPS coordinates to device

arthanson:


jeremystretch (approved):
Nice work, thanks @arthanson!

Issue #12609, opened 2023-05-15
Netbox 3.5 script load changes cause script module load errors


Omripresent:
### NetBox version

v3.5.1

### Python version

3.10

### Steps to Reproduce

1. Create a custom script module using a subdirectory and __init__.py and a script class within this file
2. Mount the module under the scripts/ directory
3. Try to access the script via its API endpoint
4. Receive DoesNotExist at /api/extras/scripts/<module>.<class name>

### Expected Behavior

We should be able to make an API call to start a custom script execution when deployed as a module.

### Observed Behavior

A ScriptModule matching query does not exist exception raised.

Traceback:
```
Environment:


Request Method: GET
Request URL: https://netbox-dev/api/extras/scripts/config.Render/

Django Version: 4.1.9
Python Version: 3.10.6
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.humanize',
 'corsheaders',
 'debug_toolbar',
 'graphiql_debug_toolbar',
 'django_filters',
 'django_tables2',
 'django_prometheus',
 'graphene_django',
 'mptt',
 'rest_framework',
 'social_django',
 'taggit',
 'timezone_field',
 'core',
 'circuits',
 'dcim',
 'ipam',
 'extras',
 'tenancy',
 'users',
 'utilities',
 'virtualization',
 'wireless',
 'django_rq',
 'drf_spectacular',
 'drf_spectacular_sidecar',
 'netbox_bgp.BGPConfig']
Installed Middleware:
['graphiql_debug_toolbar.middleware.DebugToolbarMiddleware',
 'django_prometheus.middleware.PrometheusBeforeMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'netbox.middleware.RemoteUserMiddleware',
 'netbox.middleware.CoreMiddleware',
[...]

PR #12801, opened 2023-06-02
#12624 and #12609 | Custom scripts and reports local file sync command

### LINKED ISSUE #12624: Upload custom scripts outside of GUI
### NetBox version

v3.5.1

### Feature type

### Proposed functionality

Versions prior to Netbox 3.5 you had the option to upload custom scripts to the path **/opt/netbox/netbox/scripts**. I would propose to bring this option back. While still having the option to upload via the web GUI.
After updating to Netbox 3.5 you have to upload scripts from the Netbox web GUI, this makes it impossible for us to upload scripts since we have a **read-only** file system on our Netbox deployment.

### Use case

This feature would make it possible to upload scripts outside of the GUI which is a necessity for deployments which have a read-only filesystem.

An example of a deployment with a read-only filesystem is our own which runs on Kubernetes, where it is very common to have such a setup.

### Database changes

N/A

### External dependencies

N/A

Omripresent:
### Fixes: #12624 and #12609

Adding a command script to synchronize the `ScriptModule` and `ReportModule` models with the unmanaged files under `SCRIPTS_ROOT` and `REPORTS_ROOT`.

Also updates the lookup of `ScriptModule` and `ReportModule` Django and API views to account for single file and python package variations.


DanSheps:
@Omripresent

Please review the Contributing guidelines.  We cannot accept a PR until the underlying issue has been accepted.

PR #12820, opened 2023-06-06
Fixes #11209 - Fix PrefixIPAddress view with saved sort preferences

### LINKED ISSUE #11209: IPAM: Child address list for IP prefix is garbled
### NetBox version

v3.4.0

### Python version

3.8

### Steps to Reproduce

1. View a prefix with multiple non-sequentially assigned child addresses 
2. Click on "IP Addresses"

### Expected Behavior

The list of IP address is displayed, with "xx IPs available" rows in rows between assigned IP address ranges

### Observed Behavior

All "xx IPs available" placeholders are displayed at the top of the list. 

Screenshot 2022-12-16 at 09 58 40

After clicking on the row header in order to sort the list, the "xx IPs available" placeholders are all gone.

DanSheps:
### Fixes: #11209 - Do not add available ips when IPAddressTable sort preferences are saved

* Exclude the IPAddress ViewTab from adding available IP annotations when table is sorted.

Checked other views (PrefixPrefixes, PrefixIPRanges) for this and is only present in the IPAddress view (Prefixes are annotated with actual prefixes, IPRanges have no such annotation)

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -952,7 +952,9 @@ class VLANGroupVLANsView(generic.ObjectChildrenView):
         )
 
     def prep_table_data(self, request, queryset, parent):
-        return add_available_vlans(parent.get_child_vlans(), parent)
+        if not get_table_ordering(request, self.table):
+            return add_available_vlans(parent.get_child_vlans(), parent)
+        return queryset
 
 
 #
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -1,7 +1,6 @@
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import F, Prefetch
 from django.db.models.expressions import RawSQL
-from django.db.models.functions import Round
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.utils.translation import gettext as _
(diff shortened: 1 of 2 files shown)


jeremystretch:
This removes the annotations when an ordering is specified, however it does not seem possible to restore as the ordering is saved in the user's config as `['']`. This may require a separate bug report to address.

DanSheps:
Do you need anything cleaned up on this one?

jeremystretch:
Unfortunately I'm not sure this approach is viable. Even with the fix for #12914 in, `prep_table_data()` is called before the prior ordering is cleared from the UserConfig data. So, if you order the table by IP and then clear the ordering, the "available" annotations aren't added on the second request, but they _are_ added on successive requests because by then the ordering has been cleared.

DanSheps:
> Thanks for your work on on this @DanSheps but IMO we should find a cleaner way to evaluate the table's ordering logic. This might entail refactoring `prep_table_data()` to pass the table itself.

I think this is a chicken and egg problem.  You need to generate the table before you can configure() it which is where the ordering changes happen, but you can't make the ordering changes without first initializing the table (which includes the data as part of initialization).

The only thing I could think of if we refactor would be to call get_table() before with an empty dataset then take the data from get_table_data() and overwrite the data in the table class.  Something like:

```
table = self.get_table(table_data, request, has_bulk_actions)
table_data = self.prep_table_data(request, child_objects, instance, table)
table.data = TableData.from_data(data=table_data)
table.data.set_table(table)
```

To me, that seems messy and could have unintended consequences, it does do a lot with that data within the init() method.

It may make more sense to have the prep_table_data be part of the table class, but I can't see that working really well either unless we have a different table class for ObjectChildrenView.

PR #12834, opened 2023-06-07
add elevation button to location view

### LINKED ISSUE #12748: locations view - request for link to "View Elevations" on single location view
### NetBox version

v3.5.2

### Feature type

restore old functionality

### Proposed functionality

on a location (eg a row) with no child locations. there is no link to view elevations of that location. in previous versions there was a "view elevations" button to the right of the rack count, which was on the left hand side list. this was lost when the list of object types was moved to the right

eg: https://demo.netbox.dev/dcim/locations/1/
this location has 8 racks but theres no single click way to get from the location to the elevation of this location.

### Use case

Single click link from a single location view to elevations view

### Database changes

_No response_

### External dependencies

_No response_

ITJamie:
### Fixes: #12748

elevation view button was removed from the locations view when racks counter was moved.
previous view elevations link

new elevation button

### DIFF
--- a/netbox/templates/dcim/location.html
+++ b/netbox/templates/dcim/location.html
@@ -16,6 +16,9 @@
       <span class="mdi mdi-plus-thick" aria-hidden="true"></span> Add Child Location
     </a>
   {% endif %}
+  <a href="{% url 'dcim:rack_elevation_list' %}?site={{ object.site.slug }}&location_id={{ object.pk }}" class="btn btn-sm btn-primary" title="View elevations">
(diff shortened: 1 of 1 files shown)


jeremystretch:
The top row is intended for buttons that perform actions relevant to the object. IMO it does not make sense to place a link to a rack elevations list here.

jeremystretch:
As #12834 has not yet been accepted I'm going to close this PR, however you're welcome to post the above screenshot as a comment on the FR for feedback.

PR #12864, opened 2023-06-11
Fixes #12847 - Include Missing Add buttons to Views

### LINKED ISSUE #12847: Include "add" button on page heading for virtual machine interfaces list
### NetBox version

v3.5.3

### Python version

3.9

### Steps to Reproduce

1. Click on Virtualization
2. Click on Interfaces

### Expected Behavior

I don't know if it would be exactly a bug or if something like that has already been opened (at least I didn't find it), but I believe that the "add" button previously existed in that part marked in red. At least in this version, there is this button on several other pages of the system. Sorry the English.
image

### Observed Behavior

That the button existed, since items can be added.

sudheesh001:
### Fixes: #12847 

- Includes missing `Add` button in Views for `Device Components`
- Includes missing `Add` button in Views for `Virtualization > Interfaces`

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -2193,7 +2193,6 @@ class ConsolePortListView(generic.ObjectListView):
     filterset = filtersets.ConsolePortFilterSet
     filterset_form = forms.ConsolePortFilterForm
     table = tables.ConsolePortTable
-    actions = ('import', 'export', 'bulk_edit', 'bulk_delete')
 
 
 @register_model_view(ConsolePort)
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):

sudheesh001:
Thanks for the review @jeremystretch I just updated the PR based on your feedback. Please let me know if this would be a better patch.

jeremystretch:
That'll work, thanks!

jeremystretch (approved):

PR #12874, opened 2023-06-12
assigning broadcast ip error fixes for ipv6 and /31/32

### LINKED ISSUE #12687: Change rules for 'Disallow assigning bcast/networks to interfaces'
### NetBox version

v3.5.2

### Feature type

### Proposed functionality

[#9068](https://github.com/netbox-community/netbox/issues/9068) introduced a new limitation to adding IPs to device interfaces.
The parameters chosen for this FR are limiting how IPs are used.

### Use case

IPv6 does not have broadcast or network addresses. All is handled with linklocal addresses and multicast.. You can used to :: or :ffff ip's without problems. We tend to use the lowest address (aaaa:bbbb:1234:5678::) as gateway.

For IPv4, there are use cases for the network and broadcast address. When I route an entire subnet to a firewall, we can use all ips for NAT translations. Those IPs are added to the device on a virtual interface. 
I tend to flag that prefix as a 'pool' (All IP addresses within this prefix are considered usable). Can that 'pool' be taken into account in the validation ?

### Database changes

_No response_

### External dependencies

_No response_

ITJamie:
### Fixes: #12687

adding appropriate logic to raise assigning broadcast error on the correct ipv4 ip's

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -370,7 +370,7 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
                     raise ValidationError(msg)
                 if address.version == 6 and address.prefixlen not in (127, 128):
                     raise ValidationError(msg)
-            if address.ip == address.broadcast:
+            if address.version == 4 and address.ip == address.broadcast and address.prefixlen not in (31, 32):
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @ITJamie!

PR #12894, opened 2023-06-14
Change link parsing from quote_plus to quote

### LINKED ISSUE #12822: Custom links with variables use improper URL encoding when spaces are included in the site name
### NetBox version

v3.5.3

### Python version

3.10

### Steps to Reproduce

1. Create site with space in it - Example: "Test Site"
2. Create custom link with content type "DCIM: Site".
3. In the link url, use format: https://www.website.com/anything/{{object.name}}

### Expected Behavior

When the link is selected, the variable should be replaced with "test site".  proper URL encoding would look like this:

https://www.website.com/anything/test%20site

### Observed Behavior

The link sends me to this URL which is what would be used when submitting a query.  This may work for some URLs but with sharepoint all our links are now invalid.

https://www.website.com/anything/test+site

abhi1693:

### DIFF
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -285,7 +285,7 @@ class CustomLink(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
         text = clean_html(text, allowed_schemes)
 
         # Sanitize link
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @abhi1693!

PR #12901, opened 2023-06-14
Fixes #12885: Permit mounting of devices in U100

### LINKED ISSUE #12885: Can't add 1-unit device to position "U100" of rack with height 100
### NetBox version

v3.5.3

### Python version

3.11

### Steps to Reproduce

1. Create rack with height "100"
2. Create device_type with height "1"
3. Create device with this device_type and position "U100" 

### Expected Behavior

Added device

### Observed Behavior

Error:

- Position
Ensure this value is less than or equal to 99.5.

jeremystretch:
### Fixes: #12885

- Fixes the `MaxValueValidator` applied to `Device.position` to allow 100 and 100.5 as valid values


netbox/dcim/migrations/0154_half_height_rack_units.py
```diff
@@ -18,6 +18,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='device',
             name='position',
-            field=models.DecimalField(blank=True, decimal_places=1, max_digits=4, null=True, validators=[django.core.validators.MinValueValidator(1), django.core.validators.MaxValueValidator(99.5)]),
+            field=models.DecimalField(blank=True, decimal_places=1, max_digits=4, null=True, validators=[django.core.validators.MinValueValidator(1), django.core.validators.MaxValueValidator(100.5)]),
```
arthanson, line 21: We should probably have this defined as a constant, it is sort of a magic number right now.  Also wondering in regards to https://github.com/netbox-community/netbox/pull/12778 would people potentially want racks to go beyond this (i.e. start at 200 up to 300?)

arthanson (changes requested):

netbox/dcim/migrations/0154_half_height_rack_units.py
jeremystretch: We only support rack heights up to 100, which already exceeds any practical real-world instance. If anyone voices a need for larger racks it's almost certainly because they're misusing the model.

I'll define a constant for the upper bound, though I doubt it would have mitigated this bug anyway as it's a classic off-by-one error.

Issue #12591, opened 2023-05-12
Introduce a UI view for altering NetBox's dynamic configuration parameters


jeremystretch:
### NetBox version

v3.5.1

### Feature type

New functionality

### Proposed functionality

Introduce a UI view which allows administrators to alter NetBox's [dynamic configuration parameters](https://docs.netbox.dev/en/stable/configuration/#dynamic-configuration-parameters), and remove this functionality from the Django admin UI.

### Use case

The admin UI is being removed in NetBox v4.0 (see https://github.com/netbox-community/netbox/issues/12325).

emersonfelipesp:
Hi @jeremystretch, will this page include support to change `PLUGINS` and `PLUGINS_CONFIG` variables?

deku-m:
Would be nice to also have the LDAP/SSO/SAML config in there maybe?

jeremystretch:
The scope of this issue is limited to moving the existing view from the admin UI to the regular frontend.

PR #12904, opened 2023-06-14
12591 config params admin

arthanson:
### Fixes: #12591 

Adds standard list, add, edit views for ConfigRevisions so can mange them within NetBox UI instead of the admin.  Did not add the bulk edit or import views as those did not seem to make any sense for ConfigRevision.  Left the functionality within the Admin, not sure if we wanted to remove those right away.

**Note:** Kept in the one-line admin.py as without it will cause a circular import issue.  Opened up #12919 to track as this would require more refactoring that is worth a separate ticket.

Monosnap Config Revisions | NetBox 2023-06-14 12-44-20
Monosnap Editing config revision Config revision #1 (2023-06-13 22:21:04 520611+00:00) | NetBox 2023-06-14 12-44-37
Monosnap Restore: Config revision #3 (2023-06-16 15:18:54 162509+00:00) | NetBox 2023-06-16 13-10-51


jeremystretch:
> Left the functionality within the Admin, not sure if we wanted to remove those right away.

IMO we should remove it from the admin: It can be confusing having the same functionality accessible in two places.

netbox/extras/forms/__init__.py
```diff
@@ -4,5 +4,4 @@
 from .bulk_import import *
 from .misc import *
 from .mixins import *
-from .config import *
```
jeremystretch, line 7: We should either move ConfigRevisionForm into `model_forms.py` or leave it here, but not have two copies of it.

netbox/extras/forms/model_forms.py
```diff
@@ -374,3 +378,99 @@ class Meta:
             'assigned_object_type': forms.HiddenInput,
             'assigned_object_id': forms.HiddenInput,
         }
+
+
+EMPTY_VALUES = ('', None, [], ())
+
+
+class FormMetaclass(forms.models.ModelFormMetaclass):
```
jeremystretch, line 386: We should rename this to avoid confusion with the stock class.

```suggestion
class ConfigRevisionFormMetaclass(forms.models.ModelFormMetaclass):
```

netbox/netbox/navigation/menu.py
```diff
@@ -344,6 +344,22 @@
...
+ADMIN_MENU = Menu(
+    label=_('Admin'),
+    icon_class='mdi mdi-account-multiple',
+    groups=(
+        MenuGroup(
+            label=_('Configuration'),
+            items=(
+                MenuItem(
+                    link='extras:configrevision',
+                    link_text=_('Config Revision'),
```
jeremystretch, line 356:
```suggestion
                    link_text=_('Config Revisions'),
```

netbox/templates/extras/configrevision.html
```diff
@@ -0,0 +1,213 @@
...
+          </table>
+        </div>
+      </div>
+
+      <div class="card">
+        <h5 class="card-header">Banners</h5>
+        <div class="card-body">
+          <table class="table table-hover attr-table">
+            <tr>
+              <th scope="row">Banner login:</th>
```
jeremystretch, line 108: We can rename some of these parameters to be more human-friendly:

```suggestion
              <th scope="row">Login banner:</th>
```

netbox/utilities/views.py
```diff
@@ -93,7 +94,7 @@ def dispatch(self, request, *args, **kwargs):
                 'a base queryset'.format(self.__class__.__name__)
             )
 
-        if not self.has_permission():
+        if isinstance(self.queryset, RestrictedQuerySet) and not self.has_permission():
```
jeremystretch, line 97: We don't want to bypass `has_permission()` entirely because it's responsible for calling `user.has_perms()`. If it's necessary to avoid calling `restrict()` on querysets which don't support it, we should account for that inside `has_permission()`.

jeremystretch (changes requested):

arthanson:
All updated @jeremystretch - I noticed I forgot to add the restore command, so that is in there now.

netbox/extras/admin.py
jeremystretch, line 1: This file can probably be deleted.

netbox/extras/views.py
```diff
@@ -1189,3 +1190,73 @@ def post(self, request):
...
+
+@register_model_view(ConfigRevision)
+class ConfigRevisionView(generic.ObjectView):
+    queryset = ConfigRevision.objects.all()
+
+
+@register_model_view(ConfigRevision, 'edit')
+class ConfigRevisionEditView(generic.ObjectEditView):
+    queryset = ConfigRevision.objects.all()
+    form = forms.ConfigRevisionForm
```
jeremystretch, line 1215: It shouldn't ever be necessary to modify a ConfigRevision, so this view is not needed. Instead, we create a new one to replace it.

netbox/extras/views.py
```diff
@@ -1189,3 +1190,73 @@ def post(self, request):
...
         return HttpResponse(rendered)
+
+
+#
+# Config Revision
+#
+
+
+class ConfigRevisionListView(generic.ObjectListView):
+    queryset = ConfigRevision.objects.all()
```
jeremystretch, line 1201: ConfigRevisions are listed with the oldest first; it would make more sense to list the newest (current) revision first. Ideally this should be set by `Meta.ordering` on the model.

netbox/utilities/views.py
```diff
@@ -78,8 +79,9 @@ def has_permission(self):
         if user.has_perms((permission_required, *self.additional_permissions)):
 
             # Update the view's QuerySet to filter only the permitted objects
-            action = resolve_permission(permission_required)[1]
-            self.queryset = self.queryset.restrict(user, action)
+            if isinstance(self.queryset, RestrictedQuerySet):
```
jeremystretch, line 82: We should probably capture this change under a separate issue for visibility.

jeremystretch (changes requested):

netbox/utilities/views.py
jeremystretch: Thinking about this further, we should leave the mixin alone. Instead, we can override `has_permission()` the specific views. We'll need to do this anyway to check for `is_superuser`.

jeremystretch: Disregard; we can just employ RestrictedQuerySet for the ConfigRevision model.

PR #12909, opened 2023-06-15
Fix #12865 - Include Add Nav Buttons for Report and Script Objects

### LINKED ISSUE #12865: Include "add" buttons on navigation menu for reports and scripts
### NetBox version

v3.5.3

### Python version

3.8

### Steps to Reproduce

1. Once the netbox server instance is running, login into the application with the user credentials
2. In the sidebar, uncollapse the `Customization` section and scroll to the `Reports and Scripts`
3. The quick `Add` buttons are missing for both `Reports` and `Scripts`, while their corresponding views contain `Add` buttons.

Screen Shot 2023-06-11 at 1 15 04 PM

### Expected Behavior

The elements for `Reports` and `Scripts` should contain the quick `Add` buttons.

### Observed Behavior

The buttons are missing.

stuntguy3000:
### Fixes: #12865 

This change introduces "Add" Navigation Buttons for Report and Script objects. It also changes the internal URL names for Report and Script Add paths to enable this functionality like other paths.

This change has been tested against permissions sets to validate functionality.

### DIFF
--- a/netbox/netbox/navigation/menu.py
+++ b/netbox/netbox/navigation/menu.py
@@ -301,12 +301,14 @@ CUSTOMIZATION_MENU = Menu(
                 MenuItem(
                     link='extras:report_list',
                     link_text=_('Reports'),
-                    permissions=['extras.view_report']
+                    permissions=['extras.view_report'],
+                    buttons=get_model_buttons('extras', "reportmodule", actions=['add'])
                 ),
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):

jeremystretch:
Thanks @stuntguy3000!

PR #12911, opened 2023-06-15
Close #12870 - Expand the TokenProvision API to include All Token values

### LINKED ISSUE #12870: Support for setting expiration time when using the API token provisioning endpoint (/api/users/tokens/provision/)
### NetBox version

v3.5.3

### Python version

3.10

### Steps to Reproduce

API ignores "expires" attribute when generating a token for a user and sets value to null.

```
curl 'https://<netbox hostname>/api/users/tokens/provision/' \
-H 'Accept: application/json' \
-d '{
    "username": "admin",
    "password": "<admin password>",
    "expires": "2023-06-12T12:47:47.245Z"
}'
```

`{"id":40,"url":"https://<netbox hostname>/api/users/tokens/40/","display":"<key>","user":{"id":1,"url":"https://<netbox hostname>/api/users/users/1/","display":"admin","username":"admin"},"created":"2023-06-12T12:51:57.222273Z","expires":null,"last_used":null,"key":"<key>","write_enabled":true,"description":"","allowed_ips":null}`

If an invalid date string is set to the value of "expires" the same thing happens.

### Expected Behavior

The response should be that the "expires" is set to that which is in the request payload, or an error if it is invalid and not setting it to null which has the result that the token has unlimited validity.

### Observed Behavior

All tokens set via the API have unlimited validity.

stuntguy3000:
### Fixes: #12870 

This change allows the TokenProvision API to accept all values applicable to the Token field, providing more user flexibility and use cases.

### DIFF
--- a/netbox/users/api/views.py
+++ b/netbox/users/api/views.py
@@ -3,7 +3,7 @@ from django.contrib.auth.models import Group, User
 from django.db.models import Count
 from drf_spectacular.utils import extend_schema
 from drf_spectacular.types import OpenApiTypes
-from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.exceptions import AuthenticationFailed, ValidationError
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.routers import APIRootView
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
We should avoid using two different serializers for the same data. Instead, `TokenProvisionSerializer` should be extended to support the necessary fields.

Issue #12237, opened 2023-04-12
Update to Django 4.2 and psycopg3


arthanson:
### Proposed Changes

Django 4.2 has been released and it has support for the new psycopg3 (instead of the current psycopg2).

Also note: https://docs.djangoproject.com/en/4.2/releases/4.2/#index-together-option-is-deprecated-in-favor-of-indexes index-together is deprecated for indexes and we have this in a couple places in the code.

### Justification

From the Django release notes: "Support for psycopg2 is likely to be deprecated and removed at some point in the future.

Be aware that psycopg 3 introduces some breaking changes over psycopg2. As a consequence, you may need to make some changes to account for differences from psycopg2."

jeremystretch:
It's worth calling out that Django 4.2 also [drops support for PostgreSQL 11](https://docs.djangoproject.com/en/4.2/releases/4.2/#dropped-support-for-postgresql-11).

PR #12916, opened 2023-06-15
12237 update to Django 4.2 / psycopg3

arthanson:
### Fixes: #12237 

Updates requirements to Django 4.2.2 and pyscopyg3.
this broke tobytes accesss to DataFile so fixed that for test cases.
changed index_together to indexes as this was deprecated.


requirements.txt
```diff
@@ -27,7 +27,9 @@ mkdocs-material==9.1.15
 mkdocstrings[python-legacy]==0.22.0
 netaddr==0.8.0
 Pillow==9.5.0
-psycopg2-binary==2.9.6
+psycopg==3.1.9
+psycopg-binary==3.1.9
+psycopg-pool==3.1.7
```
jeremystretch, line 32: Are all three of these necessary?

netbox/extras/migrations/0093_rename_taggeditem_content_type_object_id_extras_tagg_content_717743_idx.py
jeremystretch, line 1: We should give the migration a more readable name rather than relying on the auto-generated one.

jeremystretch (changes requested):

requirements.txt
jeremystretch: [Found this article](https://learndjango.com/tutorials/psycopg3-binary-and-django-42-installation-quick-t) which partially answers my question. And it seems that pooling has also been split out from the main `psycopg` package.

arthanson: made to one line

netbox/extras/migrations/0093_rename_taggeditem_content_type_object_id_extras_tagg_content_717743_idx.py
arthanson: changed.

arthanson:
@jeremystretch updated and merged latest feature.  Condensed the requirements.txt changes to one line so the base_requirements should be good now without additional changes.

PR #12928, opened 2023-06-19
Closes #12518: Add 'asset tag' to rack view

### LINKED ISSUE #12518: Show 'asset_tag' in the rack view
### NetBox version

v3.4.10

### Feature type

Data model extension

### Proposed functionality

Add 'Asset tag' to the dropdown of the 'rack view'
image

### Use case

In our datacenter, we only have the assettags labeled on the devices, not the hostnames.
Currently we can see the assettag if we 'hover' over the name of the device, but it would be better to see the asset in the view direct

### Database changes

_No response_

### External dependencies

_No response_

PieterL75:


abhi1693:

PieterL75:
@abhi1693 This PR was for you to evaluate the solution presented. How else can you evaluate the implemented proposal ?
I did not get any response on my questions if that was needed or not..

PR #12929, opened 2023-06-19
Fixes #11622 Cable trace in cable view

### LINKED ISSUE #11622: Cable trace in cable view

PieterL75:

### DIFF
--- a/netbox/templates/dcim/cable.html
+++ b/netbox/templates/dcim/cable.html
@@ -18,16 +18,11 @@
             </tr>
             <tr>
               <th scope="row">{% trans "Status" %}</th>
-              <td>{% badge object.get_status_display bg_color=object.get_status_color %}</td>
-            </tr>
-            {% if object.get_trace_url %}
-            <tr>
-              <th scope="row">Trace</th>
               <td>
-              <a href="{{ object.get_trace_url }}" class="btn btn-primary btn-sm" title="Trace"><i class="mdi mdi-transit-connection-variant" aria-hidden="true"></i></a>
+                {% badge object.get_status_display bg_color=object.get_status_color %}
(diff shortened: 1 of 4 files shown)


jeremystretch (changes requested):

netbox/templates/dcim/cable.html
```diff
@@ -18,7 +18,10 @@ <h5 class="card-header">{% trans "Cable" %}</h5>
             </tr>
             <tr>
               <th scope="row">{% trans "Status" %}</th>
-              <td>{% badge object.get_status_display bg_color=object.get_status_color %}</td>
+              <td>
+                {% badge object.get_status_display bg_color=object.get_status_color %}
+                {% if object.get_trace_url %} <a href="{{ object.get_trace_url }}" class="btn btn-primary btn-sm" title="Trace"><i class="mdi mdi-transit-connection-variant" aria-hidden="true"></i></a>{% endif %}
```
jeremystretch, line 23: This isn't ideal because the trace's origin is non-deterministic. Instead, we can add an individual trace button to each termination listed on the righthand side of the page.

Issue #7845, opened 2021-11-16
Automatically create direct assignments of IP addresses to their parent prefixes


jeremystretch:
### NetBox version

v3.0.10

### Feature type

Change to existing functionality

### Proposed functionality

Currently, NetBox determines the relationships between prefixes and IP addresses dynamically, leveraging PostgreSQL query logic to find individual IPs which exist "within" each prefix (filtered by VRF): there is no direct relationship in the database from an IP address to its parent prefix.

This issue proposes effecting a direct ForeignKey field named `prefix` on the IPAddress model relating to the Prefix model. This will be for internal use only, and not editable via any form or API. The relationship will be set automatically whenever a prefix or IP address is created, modified, or deleted. (Django signals will be leveraged to detect these events.) It should be possible to perform these updates via a single bulk query for each object being modified.

For example, when creating the prefix 192.0.2.0/24 in VRF A, a bulk query will automatically update the `prefix` field of any IP addresses within this prefix and VRF to reference the new prefix. Similarly, when a new IP address is created, NetBox will assign it to the closest matching prefix.

It may make sense to extend this concept to the Prefix model as well, automatically defining the parent prefix for each child prefix.

(This proposal is related to #7451 but more involved. If this proposal is accepted, #7451 will likely be closed.)

### Use case

The introduction of a ForeignKey relationship from IPAddress to Prefix will make querying child IPs and calculating prefix utilization much more efficient.

### Database changes

Add a `prefix` ForeignKey field to the IPAddress model relating to Prefix.

IdRatherStand:
Would this allow us to reference an IP's prefix (and related prefix values) within the IP Address permissions constraints? As an example at the moment we are unable to grant IP Permissions based on the parent prefix, an example of the constraint we would look to use (on IPAM | IP Address) could be:

{"prefix__custom_field_data__fieldA":"XXX"}

At present we can't find a way to do this so if the answer is yes a big thumbs up from me.

DanSheps:
It most likely would, yes

jeremystretch:
Yes, but you'd have to be extremely careful about it. For example, if you create 192.168.0.0/16 and 192.168.0.0/24, the former will be assigned as the parent of the later. However, if someone later creates 192.168.0.0/20, it will become the new parent for 192.168.0.0/24.

> At present we can't find a way to do this so if the answer is yes a big thumbs up from me.

[...]

PR #12937, opened 2023-06-19
prefix is_pool check for ip assignments

### LINKED ISSUE #12962: allow assigning of prefix network and broadcast ips to devices if prefix is_pool
### NetBox version

v3.5.4

### Python version

3.10

### Steps to Reproduce
attempt to assign a broadcast ip or network IP to a device. (eg 192.168.0.0/24 or 192.168.0.255/24)

### Expected Behavior
if the prefix the ip belongs to has is_pool set to true, it should be possible to assign the network and broadcast ips to an interface

### Observed Behavior
error is raised about not being allowed to assign the network or broadcast ips to an interface
currently this is only allowed for /31,/32,/127,/128 network or broadcast ips.

### additional info
note this is a follow on from #12687

ITJamie:
### Fixes: #12962

some users pointed out in #12687 that I missed checking for prefixes where is_pool is true with my previous patch.
I added some logic to find a prefix matching the IP and checks to see if is_pool is set.
It is also safe for ip's that have no matching prefix in netbox.

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -365,12 +365,12 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
             prefix_str = f"{self.instance.address.network}/{self.instance.address.prefixlen}"
             allow_assignment_error = True
             if self.instance.vrf is None:
-                prefix_obj =  Prefix.objects.get(prefix=prefix_str)
+                prefix_obj = Prefix.objects.get(prefix=prefix_str)
             else:
-                prefix_obj =  Prefix.objects.get(prefix=prefix_str, vrf=self.vrf)
+                prefix_obj = Prefix.objects.get(prefix=prefix_str, vrf=self.vrf)
             if prefix_obj and prefix_obj.is_pool:
                 allow_assignment_error = False
-            
+
             if allow_assignment_error:
                 if address.ip == address.network:
                     msg = f"{address} is a network ID, which may not be assigned to an interface unless the prefix is a pool."
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -362,13 +362,12 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
 
         # Do not allow assigning a network ID or broadcast address to an interface.
         if interface and (address := self.cleaned_data.get('address')):
-            prefix_str = f"{address.network}/{address.prefixlen}"
             allow_assignment_error = True
             if self.instance.vrf is None:
-                prefix_obj = Prefix.objects.filter(prefix=prefix_str)
+                prefix_obj = Prefix.objects.filter(prefix=address.cidr)
             else:
-                prefix_obj = Prefix.objects.filter(prefix=prefix_str, vrf=self.vrf)
-            if prefix_obj.exists() and prefix_obj[0].is_pool:
+                prefix_obj = Prefix.objects.filter(prefix=address.cidr, vrf=self.vrf)
+            if prefix_obj.exists() and prefix_obj.first().is_pool:
                 allow_assignment_error = False
 
             if allow_assignment_error:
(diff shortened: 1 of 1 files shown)


netbox/ipam/forms/model_forms.py
```diff
@@ -362,15 +362,25 @@ def clean(self):
 
         # Do not allow assigning a network ID or broadcast address to an interface.
         if interface and (address := self.cleaned_data.get('address')):
-            if address.ip == address.network:
-                msg = f"{address} is a network ID, which may not be assigned to an interface."
-                if address.version == 4 and address.prefixlen not in (31, 32):
+            prefix_str = f"{self.instance.address.network}/{self.instance.address.prefixlen}"
+            allow_assignment_error = True
+            if self.instance.vrf is None:
+                prefix_obj = Prefix.objects.get(prefix=prefix_str)
```
kkthxbye-code, line 368: Wont this throw an exception if no prefix exists?

ITJamie: I will double check tomorrow am but i thought i tested that.

jeremystretch:
Please open a new issue to track these changes, as #12687 has already been closed.

ITJamie:
created #12962

kkthxbye-code (commented):
Personally I think this whole thing should be reverted and users encouraged to use custom validators to enforce it instead. The logic is getting more and more convoluted, there's been a bunch of edgecases and in general we don't really do strict validation of stuff like this for most of netbox.

The validation is also still only present in the form, so bulk editing and the API have no restrictions.

If it must be merged, maybe it would be the time to add some utility methods to the IPAddress model to allow finding the parent prefix(s) more easily. Would be useful for configuration generation and custom scripts as well.

ITJamie:
> Personally I think this whole thing should be reverted and users encouraged to use custom validators to enforce it instead. The logic is getting more and more convoluted, there's been a bunch of edgecases and in general we don't really do strict validation of stuff like this for most of netbox.

Im in agreement actually. That or add a configuration setting to disable this validation chunk completely 


> The validation is also still only present in the form, so bulk editing and the API have no restrictions.
> 
> 
> 
> If it must be merged, maybe it would be the time to add some utility methods to the IPAddress model to allow finding the parent prefix(s) more easily. Would be useful for configuration generation and custom scripts as well.

I would agree but at that point were starting to encroach on #7845 when it comes to relationship finding between prefixes and ips

netbox/ipam/forms/model_forms.py
```diff
@@ -362,15 +362,25 @@ def clean(self):
...
-            if address.ip == address.network:
-                msg = f"{address} is a network ID, which may not be assigned to an interface."
-                if address.version == 4 and address.prefixlen not in (31, 32):
+            prefix_str = f"{address.network}/{address.prefixlen}"
+            allow_assignment_error = True
+            if self.instance.vrf is None:
+                prefix_obj = Prefix.objects.filter(prefix=prefix_str)
+            else:
+                prefix_obj = Prefix.objects.filter(prefix=prefix_str, vrf=self.vrf)
+            if prefix_obj.exists() and prefix_obj[0].is_pool:
```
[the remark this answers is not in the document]
ITJamie: so from a logic POV.

lets say you have two prefixes.
192.168.0.0/24 - with has_pool = False
192.168.1.0/24 - with has_pool = True

in the following cases there should
192.168.0.0/24 - should error if created and attempt to assign to an interface
192.168.0.254/24 - should error if created and attempt to assign to an interface

192.168.1.0/24 - should allow creation and assignment to an interface
192.168.1.254/24 - should allow creation and assignment to an interface


if you then decided to create an assign an IP with NO matching prefix in netbox. it should error if you attempt to assign it to an interface (standard logic as there's nothing to tell it its a pool of ips)
192.168.2.0/24 - should error if created and attempt to assign to an interface
192.168.2.254/24 - should error if created and attempt to assign to an interface


this is making the assumption that a netbox user has correctly made their prefix object with the same prefixlen.
but unless #7845 happens I see no better way, this way is explicit in that the supplied address.cidr has to match an existing prefix if you want the is_pool assignment override logic.

ITJamie:
Closing out per #12962

PR #12989, opened 2023-06-23
Fixes device type and module type tag import

### LINKED ISSUE #12949: modules-type CSV bulk import does not import tags
### NetBox version

dokcer version 3.5.3

### Python version

3.10

### Steps to Reproduce

1. click on the import button under the modules-type section.
2. under the tab 'Direct import' past the following in the 'data' field (make sure the tags and custom fields are already created)
```
manufacturer,model,tags,cf_FABRIC_PROFILE_ACC,cf_FABRIC_PROFILE_NW
Nokia,Test-mda,"pipo",-1,-1
```
3. Format can be either auto-detect or CSV
4. click 'submit'

The module is actually imported, but the tags are not.

### Expected Behavior

That the tags be considered when importing.

### Observed Behavior

no error raised.

abhi1693:


jeremystretch:
Thanks @abhi1693!

PR #12990, opened 2023-06-23
Fixes prechange snapshot

### LINKED ISSUE #12617: Webhook prechange snapshot missing when primary IP assigned or unassigned
### NetBox version

v3.5.1

### Python version

3.10

### Steps to Reproduce

1. Create a webhook - Content type `DCIM > Device`, events Updates, any URL, and the rest blank/defaults.
2. Create a device (and all the prerequisite objects: Site, Role, Manufacturer, Type) - All defaults
3. If no interfaces on the device type, add one.
4. From the interfaces view of the device, add an IP address to the new interface - Check the 'Make this the primary' box.
5. Go to the admin portal, find the webhook in the finished or failed jobs lists of the default queue.

### Expected Behavior

The value of the prechange key of snapshots should have the prechange object data.

### Observed Behavior

The value of the prechange key of snapshots is None.

abhi1693:

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -379,6 +379,7 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
         interface = self.instance.assigned_object
         if type(interface) in (Interface, VMInterface):
             parent = interface.parent_object
+            parent.snapshot()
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @abhi1693!

PR #12991, opened 2023-06-23
Fixes #12977: Fix URL parameters for object count dashboard widgets

### LINKED ISSUE #12977: Object Count Widget filters ignore lists
### NetBox version

v3.5.4

### Python version

3.9

### Steps to Reproduce

1. Create a new widget
- Widget type "Object Counts"
- Models: DCIM > Device
- Object filters: { "manufacturer_id": ["1", "2"] }
(replace "1" and "2" with valid IDs in your environment)

2. Check number of counted objects

### Expected Behavior

I'd expect NetBox to take lists into account like shown above and just like the corresponding URL does: https://netbox.tld/dcim/devices/?manufacturer_id=1&manufacturer_id=2

As per https://github.com/netbox-community/netbox/issues/12742, Object Count filters employ standard UI filters rather than low-level queryset filters since v3.5.3. While this is a good design decision, queryset filters gave the ability to use logical OR filters. Since Object List filters take lists into account, I'd expect Object Count filters to do the same.

### Observed Behavior

The number of counted objects is always equal to the number of all devices. Nothing will be filtered. The filter works if you use a single string and not a list, e.g. { "manufacturer_id": "1" }

This behaviour is not the case with Object Lists.

jeremystretch:
### Fixes: #12977

- Introduce `dict_to_querydict()` utility function to ensure proper handling of QueryDicts
- Call it everywhere we had previously been creating QueryDicts directly

### DIFF
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -9,7 +9,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.cache import cache
 from django.core.validators import ValidationError
 from django.db import models
-from django.http import HttpResponse, QueryDict
+from django.http import HttpResponse
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.formats import date_format
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -26,7 +26,7 @@ from netbox.models.features import (
     CloningMixin, CustomFieldsMixin, CustomLinksMixin, ExportTemplatesMixin, SyncedDataMixin, TagsMixin,
 )
 from utilities.querysets import RestrictedQuerySet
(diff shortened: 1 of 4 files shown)


netbox/extras/models/models.py
```diff
@@ -26,7 +26,7 @@
     CloningMixin, CustomFieldsMixin, CustomLinksMixin, ExportTemplatesMixin, SyncedDataMixin, TagsMixin,
 )
 from utilities.querysets import RestrictedQuerySet
-from utilities.utils import clean_html, render_jinja2
+from utilities.utils import clean_html, dict_to_querydict, render_jinja2
```
arthanson, line 29: can remove the import for QueryDict

arthanson (approved):

PR #12995, opened 2023-06-24
fix bulk edit clear m2m fields #12983

### LINKED ISSUE #12983: Bulk edit of models with M2M fields can clear the M2M field
### NetBox version

v3.5.4 and tested on develop

### Python version

3.10

### Steps to Reproduce

1. Create an interface with one or more tagged_vlans.
2. Use bulk edit on the interface to change something unrelated like the speed.
3. Observe the changelog with the tagged_vlans being removed.

### Expected Behavior

The untagged_vlans should remain unchanged if none are selected and set null is unchecked.

### Observed Behavior

The untagged_vlans are removed from the interface.

Caused by https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/views/generic/bulk_views.py#L554-L555
In https://github.com/netbox-community/netbox/commit/41c92483a0b9b820dd0ca0c43c6c6d6e560043ed the truthy check on the m2m queryset was removed. Django querysets evaluate to False if the queryset is empty.

netopsab:

### DIFF
--- a/netbox/netbox/views/generic/bulk_views.py
+++ b/netbox/netbox/views/generic/bulk_views.py
@@ -551,7 +551,7 @@ class BulkEditView(GetReturnURLMixin, BaseMultiObjectView):
             for name, m2m_field in m2m_fields.items():
                 if name in form.nullable_fields and name in nullified_fields:
                     getattr(obj, name).clear()
-                else:
+                elif name in form.changed_data:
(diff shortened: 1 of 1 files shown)


jeremystretch:
Looks like #12983 has been addressed by #13004. My apologies; I didn't realize we had two PRs for that issue.

PR #12997, opened 2023-06-25
Fixes syntax error on reports

### LINKED ISSUE #12842: Improve handling of faulty reports
### NetBox version

v3.5.3

### Python version

3.8

### Steps to Reproduce

1. Create a file has a python syntax error in it
2. Upload it as a 'report'
3. Now browse to `/extras/reports/`

### Expected Behavior

The invalid report to be shown as such, and to be able at least to delete it via the GUI.

Note, however, if this were to hide the error message, it would then become very difficult to debug.  See #12766 for the behaviour of Custom Scripts.

### Observed Behavior

A Server Error is generated. Unfortunately, this then means it's impossible to delete the report via the web interface, as you can't even get a list of reports.  Furthermore, if you simply rm the file from the reports directory, you get another Server Error ("File Not Found"). See #12841 for example.

#### Workaround

At the CLI:

```
cd /opt/netbox/netbox/reports
mv broken_report.py broken_report.py.xxx
touch broken_report.py
```

A zero-byte report is considered valid, and then can be deleted from within the GUI.

abhi1693:
### Fixes: #12842

I could not figure how to also display the exception on the UI, however this now solves the blocker when the report was unable to load either due to ImportError or SyntaxError. A user can delete the report and upload the fixed version.

image

To ease with the issue, I have added a logger when the exception is caught.

### DIFF
--- a/netbox/extras/models/reports.py
+++ b/netbox/extras/models/reports.py
@@ -1,7 +1,7 @@
 import inspect
+import logging
 from functools import cached_property
 
-from django.contrib.contenttypes.fields import GenericRelation
 from django.db import models
 from django.urls import reverse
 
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks @abhi1693!

jeremystretch (approved):

Issue #11672, opened 2023-02-06
Add the ability to remove user sign in field


K4pper:
### NetBox version

v3.4.4

### Feature type

New functionality

### Proposed functionality

Netbox has the ability to use multiple SSO providers, but there is no way to completely disable the login form even if there are no local users on the Netbox instance.

Example:
![image](https://user-images.githubusercontent.com/76392392/216948555-281c16ed-cabf-4f62-bfa5-d51c330ba064.png)

This could pose an unnecessary way for attackers to brute force the application by spamming the login field, especially if the Netbox instance is run in the cloud.

### Use case

By adding an option to disable the login form and only having the option to login using an SSO provider would decrease the attack surface for a malicious entity.

### Database changes

N/A

### External dependencies

N/A

proudbro:
I propose not to remove the authorization form, just simply hide it and leave available

jeremystretch:
What happens when the login form is disabled and the SSO provider becomes unavailable?

K4pper:
> What happens when the login form is disabled and the SSO provider becomes unavailable?

Just to clarify i am not proposing to have the login form hidden or removed automatically when SSO is configured.
What i am proposing is a setting, that you would manually enable to hide the login form.

Our instance of Netbox has no local users, so if the SSO provider goes down we wouldn't be able to access Netbox anyway. If the worst case scenario happens we would still be able to connect to the netbox instance and create a temporary local user.

Hope this answers the question (:

Kerwood:
What is your justification for not letting the user make the decision if the login form should be showed or not ?

jeremystretch:
As I've said, ensuring that the local login form is always available ensures that a backup authentication path exists in the event the remote authentication provided becomes inaccessible.

Kerwood:
> As I've said, ensuring that the local login form is always available ensures that a backup authentication path exists in the event the remote authentication provided becomes inaccessible.

I am not sure we are on the same page here... 
No one is proposing that the login form should be automatically removed if SSO is enabled. The login form can be available by default, just as it is now. No one will know the difference.
We just want the option to disable it in the settings.

ssza:
In my opinion the local login form should always be available, but in the case of SSO enabled, it should be hidden behind a dropdown menu

K4pper:
@jeremystretch

[...]

PR #13003, opened 2023-06-26
Allow users to define the LOGIN_URL.

### LINKED ISSUE #13002: Allow the Django LOGIN_URL setting to be defined in configuration
### NetBox version

v3.2.9

### Feature type

### Proposed functionality

Allow the users to define the default Django LOGIN_URL variable in the Netbox configuration.

### Use case

For organizations who deploy a single SSO provider it can confuse users that they are presented for both the username/password and the SSO option on the login page.

By allowing the LOGIN_URL to be defined in the configuration users can be redirect directly to the SSO provider, effectively hiding the non-functional username/password fields.

### Database changes

None

### External dependencies

None

slyngshede:
### Fixes: #13002 #11672

If you rely on a single SSO provider, it can
be confusing for users to see the username/password login with the SSO provider below. By allowing the LOGIN_URL to be overwritten users can be redirect
directly to the SSO provider.

Simply check if the LOGIN_URL has already been defined, if not fallback to f'/{BASE_PATH}login/'

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -455,8 +455,11 @@ MESSAGE_TAGS = {
     messages.ERROR: 'danger',
 }
 
-# Authentication URLs
-LOGIN_URL = f'/{BASE_PATH}login/'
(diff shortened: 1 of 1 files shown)


jeremystretch:

PR #13010, opened 2023-06-26
Fix remote auth backend check

### LINKED ISSUE #12849: Remote users (LDAP backend) with limited rights lost ability to use RestAPI
### NetBox version

v3.5.3

### Python version

3.10

### Steps to Reproduce

**Perquisite**

**configuration.py** contains:
```
REMOTE_AUTH_ENABLED = True
REMOTE_AUTH_BACKEND = 'netbox.authentication.LDAPBackend'
REMOTE_AUTH_HEADER = 'HTTP_REMOTE_USER'
REMOTE_AUTH_AUTO_CREATE_USER = True
REMOTE_AUTH_DEFAULT_GROUPS = []
REMOTE_AUTH_DEFAULT_PERMISSIONS = {}
```

ldap_config.py contains:
```
AUTH_LDAP_MIRROR_GROUPS = False
```
_can add full files with masked data paths, if needed._

Our config allow us to have users with rights, assigned by AD group membership. But when checking django group membership, there is none (we ok with it, we have no issues before):
image
If we tun on mirroring, whole bunch 10th thousands groups from AD will be loaded to netbox.

**Issue steps**

1. Add one user (Bob) to admin and staff groups, another (Alice) to access group.
2. Create permissions to view anything (or some limited list of objects, we have both scenario) and assign them to Alice.
3. Both Bob and Alice can login to UI and view any objects.
4. Create 2 (read-only in our case) tokens, one for Alice, other for Bob.

### Expected Behavior

Tokens must work anywhere and return any data for both users.

### Observed Behavior

Both users can use UI, /api/ and /api/schema/swagger-ui/ endpoints while logged in.

Both of them also receive correct answer in Swager, with token used (it's ignored, likely, coz return correct answer with wrong token).

Outside of Netbox app (from Curl, PoSh, Grafana requests) it returns "You do not have permission to perform this action." error.

image

**Problem not existed in 3.0 and 3.1**. I used several accounts in Grafana tests before moving Netbox to production, it worked.

Now I must explicitly define group membership of user for API only.

rmanyari:
### Fixes: #12849 

Support for [multiple remote authentication backend](https://github.com/netbox-community/netbox/pull/12012 ) made `REMOTE_AUTH_BACKEND` an iterable but we forgot to update this check in `api/authentication.py`

### DIFF
--- a/netbox/netbox/api/authentication.py
+++ b/netbox/netbox/api/authentication.py
@@ -60,7 +60,7 @@ class TokenAuthentication(authentication.TokenAuthentication):
 
         user = token.user
         # When LDAP authentication is active try to load user data from LDAP directory
-        if settings.REMOTE_AUTH_BACKEND == 'netbox.authentication.LDAPBackend':
+        if 'netbox.authentication.LDAPBackend' in settings.REMOTE_AUTH_BACKEND:
             from netbox.authentication import LDAPBackend
             ldap_backend = LDAPBackend()
 


jeremystretch:
Thanks @rmanyari!

PR #13028, opened 2023-06-27
Fixes #12760 - Adds Vary header to prevent caching of HTMX requests

### LINKED ISSUE #12760: Sporadically broken HTML formatting since at least 3.5.0
### NetBox version

v3.5.1

### Python version

3.10

### Steps to Reproduce

Spontaneously observed behavior:
1. Open any list
2. Click to any object
3. Press "Go back" (or ALT+←) in browser

### Expected Behavior

List will be rendered as usual.

### Observed Behavior

Loaded only part of page, without sidebar, with empty head tag:

image

DanSheps:
### Fixes: #12760

* Adds Vary header to CoreMiddleware to cause caching to depend on the URL and the HX-Request flag.

### DIFF
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -49,6 +49,9 @@ class CoreMiddleware:
         # Attach the unique request ID as an HTTP header.
         response['X-Request-ID'] = request.id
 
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @DanSheps!

PR #13035, opened 2023-06-28
Closes #8248: User bookmarks

### LINKED ISSUE #8248: Enable pinning objects to a user's home page
### NetBox version

v3.1.4

### Feature type

New functionality

### Proposed functionality

Include a "pin to homepage" button on the view for (almost?) every object. When clicked, a new record will be created associating the object with the current user as a "pinned" or "favorite" object. The NetBox home page will be extended to display a list of all such objects for the current user. Objects can be "unpinned" from this list as needed.

This idea was originally proposed by @bogdancordos in #8235.

### Use case

This allows a user to compile a list of commonly-viewed objects for near- or long-term use.

It's also somewhat related to #7855, which proposes displaying a history of recently viewed objects.

### Database changes

I imagine the introduction of a simple new model within the users app (e.g. `UserFavorite`) mapping the user object to the "pinned" object via a generic foreign key. We would probably want to track creation date, and _possibly_ assign a numeric weight to aid in reordering objects, depending on how much functionality is needed.

### External dependencies

_No response_

jeremystretch:
### Closes: #8248

- Introduce the `extras.Bookmark` model and relevant views to enable the creation and removal of bookmarks
- Add a dashboard widget to display user bookmarks


arthanson (changes requested):
Documentation?

Also - not sure necessary as you have filtering on the widget, but it might be nice to group them by type with header for the dashboard widget:
```
Devices
    PP:B117
    PP:B118
IP addresses
    172.16.0.1/24
```
The profile page shows the type but nothing on the widget which can make it confusing.

jeremystretch:
My intent currently is to keep the widget content minimal, but I'm happy to revisit denoting object types if we get more feedback about it.

PR #13049, opened 2023-06-29
Fixes asn_count on asn range

### LINKED ISSUE #13047: ASN Count is not populated in ASN Ranges view
### NetBox version

v3.5.4

### Python version

3.10

### Steps to Reproduce

1. Create ASN object in 64512-65534 range
2. Create ASN Range object and define it as the 64512-65534 range
3. ASN Count is not populated in ASN Ranges view and shows value of 0
4. In specific ASN Range object view ASN object tab shows ASN numbers included in the range

Same behaviour in the https://demo.netbox.dev instance.

Our instance after some minutes got this error:
```

Cannot resolve keyword 'asn_count' into field. Choices are: created, custom_field_data, description, end, id, journal_entries, last_updated, name, rir, rir_id, slug, start, tagged_items, tags, tenant, tenant_id

Python version: 3.10.6
NetBox version: 3.5.4
```

### Expected Behavior

ASN Count correctly displays number of included ASN objects

### Observed Behavior

SHows 0 even though objects can be seen in the specific ASN Range view

abhi1693:
### Fixes: #13047

- `asn_count` was defined on the serializer but wasn't showing up in the results
- Replaced `LinkedCountColumn` with `TemplateColumn` due to 2 reasons:
  1. As per my understanding, LinkedCountColumn uses filtersets to find the information. I tried different variations but none worked.
  2. TemplateColumn does not produce a linkified cell, as there is no UI filter on the ASN table for the range

### DIFF
--- a/netbox/ipam/api/serializers.py
+++ b/netbox/ipam/api/serializers.py
@@ -1,4 +1,5 @@
 from django.contrib.contenttypes.models import ContentType
+from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
@@ -24,7 +25,7 @@ class ASNRangeSerializer(NetBoxModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='ipam-api:asnrange-detail')
     rir = NestedRIRSerializer()
     tenant = NestedTenantSerializer(required=False, allow_null=True)
-    asn_count = serializers.IntegerField(read_only=True)
+    asn_count = serializers.SerializerMethodField(read_only=True)
 
     class Meta:
         model = ASNRange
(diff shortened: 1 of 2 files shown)


netbox/ipam/api/serializers.py
```diff
@@ -33,6 +34,10 @@ class Meta:
             'custom_fields', 'created', 'last_updated', 'asn_count',
         ]
 
+    @extend_schema_field(OpenApiTypes.INT)
+    def get_asn_count(self, obj):
+        return obj.get_child_asns().count()
+
```
jeremystretch, line 40: There are two issues with this approach:

1. It generates a separate query per object in the table.
2. It prevents us from ordering the table by ASN count.

I've opened #13053 with a more complete (albeit more complex) solution. Please let me know what you think.

jeremystretch:
Superseded by #13053

PR #13098, opened 2023-07-06
13092 allow setnull for bulk edit power port maximum and allocated draw

### LINKED ISSUE #13092: Power port bulk edit: missing "set null" on maximum power draw / allocated power draw
### NetBox version

v3.5.4

### Python version

3.8

### Steps to Reproduce

1. Go to a device which has power ports
2. Click on the power ports tab
3. Click the checkboxes besides one or more (or all) ports
4. Use the bulk "Edit" facility, e.g. e.g. `/dcim/power-ports/edit/?device=5&return_url=/dcim/devices/5/power-ports/`
5. Try to set the Allocated Draw or Maximum Draw to null for all those ports

### Expected Behavior

Be able to click "Set null" to unset the value

### Observed Behavior

The "Set null" checkbox is missing. (And setting zero is not allowed).

WORKAROUND: edit each port individually, blank out the power draw fields individually, and save each one.

arthanson:
### Fixes: #13092 

Allows to set-null on power port maximum and allocated draw when doing a bulk edit.

### DIFF
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -1106,7 +1106,7 @@ class PowerPortBulkEditForm(
         (None, ('module', 'type', 'label', 'description', 'mark_connected')),
         ('Power', ('maximum_draw', 'allocated_draw')),
     )
-    nullable_fields = ('module', 'label', 'description')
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @arthanson!

PR #13117, opened 2023-07-07
Fixes content type lookups when db is uninitialized

### LINKED ISSUE #13116: Netbox won't start with empty database
### NetBox version

v3.5.5

### Python version

3.11

### Steps to Reproduce

1. Create new empty database
2. Configure Netbox to use that database
3. Run "manage.py check" or "manage.py migrate"

### Expected Behavior

Tables are being created

### Observed Behavior

An exception occurs, because the table for the Contenttypes doesn't exists yet.
Logs can be found here: https://github.com/netbox-community/netbox-docker/actions/runs/5483212436/jobs/9989301710#step:7:62
If I revert the PR #11709 Netbox starts as expected.

abhi1693:


jeremystretch:
Thanks @abhi1693!

PR #13123, opened 2023-07-07
Fixes form rendering when scheduling_enabled is disabled

### LINKED ISSUE #13096: Script ignores 'scheduling_enabled' if it has no input parameters
### NetBox version

v3.5.4

### Python version

3.8

### Steps to Reproduce

Upload the following script and run it:

```

class TestScript(Script):
    class Meta:
        name = "Test Script"
        scheduling_enabled = False

    def run(self, data, commit):
        self.log_success("Run OK")
        return "OK\n"
```

### Expected Behavior

Since `scheduling_enabled` is False, Netbox should not offer to run the script at scheduled times.

### Observed Behavior

image

Note however that `commit_default = False` *is* honoured if provided.

abhi1693:

### DIFF
--- a/netbox/extras/forms/scripts.py
+++ b/netbox/extras/forms/scripts.py
@@ -56,10 +56,3 @@ class ScriptForm(BootstrapMixin, forms.Form):
             self.cleaned_data['_schedule_at'] = local_now()
 
         return self.cleaned_data
-
-    @property
-    def requires_input(self):
-        """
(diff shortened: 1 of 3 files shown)


jeremystretch:
Thanks @abhi1693! I've tweaked the template a bit too, to address @kkthxbye-code's [comments](https://github.com/netbox-community/netbox/issues/13096#issuecomment-1622394042) as well.

PR #13125, opened 2023-07-07
Rewrite upgrade guide to distinguish between current and new release

### LINKED ISSUE #13080: More clearly define the current version and target version in upgrade documentation
### Change Type

Correction

### Area

Installation/upgrade

### Proposed Changes

I've done several (successful) upgrades for netbox, but whenever I do them, I always get a little bit confused about what directories I'm supposed to copy and where I'm supposed to copy them and then I realise that the instructions here are rather confusing and that this is isn't just my fault :slightly_smiling_face:
https://demo.netbox.dev/static/docs/installation/upgrading/
The reason for that is that `X.Y.Z` refers both to the newly downloaded version and the current version you're using in two different consecutive code snippets. This refers to the newly downloaded version:
```
sudo ln -sfn /opt/netbox-X.Y.Z/ /opt/netbox
```
And this to the current version:
```
sudo cp /opt/netbox-X.Y.Z/local_requirements.txt /opt/netbox/
```
I mean, sure, it is logical, because you've just created that symlink in `/opt/netbox`, but it's still rather confusing and you have to look and check several times until you figure out :slightly_smiling_face:

It might be one of those small changes that can increase the readability of the documentation, but let me know if I'm reading this wrong :)

abhi1693:

### DIFF
--- a/docs/installation/upgrading.md
+++ b/docs/installation/upgrading.md
@@ -56,29 +56,32 @@ sudo ln -sfn /opt/netbox-X.Y.Z/ /opt/netbox
 Copy `local_requirements.txt`, `configuration.py`, and `ldap_config.py` (if present) from the current installation to the new version:
 
 ```no-highlight
-sudo cp /opt/netbox-X.Y.Z/local_requirements.txt /opt/netbox/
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks for tackling this @abhi1693. Inspired by your approach, I went a bit further in c89193d33 by declaring named shell variables to more clearly denote the placeholders.

Issue #13132, opened 2023-07-10
Update model field definitions to use gettext_lazy


arthanson:
### Proposed Changes

In order to correctly support internationalization, models need to be audited and add in the name as gettext_lazy string and update verbose_name where applicable to use gettext_lazy.  See Django's use at https://github.com/django/django/blob/main/django/contrib/auth/models.py#L356
```
    slug = models.SlugField(
        _("slug"), <- add this
        max_length=100
    )
```

### Justification

Model fields need their name (and where appropriate their verbose_name) as translatable strings.

PR #13148, opened 2023-07-11
13132 updated model fields to use gettext_lazy

arthanson:
### Fixes: #13132 

update model fields to make sure they have verbose_names on the fields that are localized via gettext_lazy, also audit / update help text and validation messages to make sure they are put in gettext_lazy as well.

### DIFF
--- a/netbox/circuits/models/circuits.py
+++ b/netbox/circuits/models/circuits.py
@@ -2,7 +2,7 @@ from django.contrib.contenttypes.fields import GenericRelation
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.urls import reverse
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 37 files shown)


jeremystretch:
Reviewing this raised the question of capitalization for `verbose_name` strings. Per the [Django docs](https://docs.djangoproject.com/en/4.2/topics/db/models/#verbose-field-names):

> The convention is not to capitalize the first letter of the [verbose_name](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.verbose_name). Django will automatically capitalize the first letter where it needs to.

We seem to have a mix of capitalized and non-capitalized strings at the moment (and preceding this work), so we should standardize on non-capitalized.

jeremystretch (approved):

jeremystretch:
Thanks @arthanson!

PR #13156, opened 2023-07-11
Fixes: #11478 - Alters interface serializers get_display to include device to string representation

### LINKED ISSUE #11478: Patching front port to virtual chassis member interface is broken in UI
### NetBox version

v3.4.2

### Python version

3.8

### Steps to Reproduce

1. Create an empty rack (default settings with 42 HUs works).
2. Add a device with 1 front port to rack.
3. Create a virtual chassis with at least two members (pos. 1, 2,...) with at least one interface each, **no master chassis (!)**, add members to rack.
4. Attach a cable from front port to an interface of VC member at pos. 2 or higher.

### Expected Behavior

The cable should be attached between front port and VC member's interface. Only the interfaces of selected VC member are shown.

### Observed Behavior

The interface selection always show all interfaces of all VC members even if selected VC member is no master. This can lead to confusion because you don't expect to see other VC members' interfaces.

We've made the following observations.
- This issue only occurs for VC members, not for single devices in the same rack.
- If VC member's interface X is already patched, interface X is always marked as not available to connect for all other VC members.
- Patching from the VC member's interface view works which suggests that this is indeed a VC related problem.
- When in the cable creation UI and when selecting any VC member, scrolling through the interface selector will always list interfaces of all VC members, even if selected VC member isn't a master chassis (or, as in our case, no VC member is configured as master chassis).

I believe that this is an UI issue. My guess is that in `/dcim/cables/add`, the Interface selector always displays VC member 1's interfaces instead the actual selected VC member.

DanSheps:
### Fixes: #11478 

Alters interface serializers get_display to include device to string representation in brackets to better denote what device the specific interface is connected to.

### DIFF
--- a/netbox/dcim/api/nested_serializers.py
+++ b/netbox/dcim/api/nested_serializers.py
@@ -391,6 +391,9 @@ class NestedInterfaceSerializer(WritableNestedSerializer):
         model = models.Interface
         fields = ['id', 'url', 'display', 'device', 'name', 'cable', '_occupied']
 
+    def get_display(self, obj):
+        return f"{obj.name} ({obj.device})"
+
 
 class NestedRearPortSerializer(WritableNestedSerializer):
     device = NestedDeviceSerializer(read_only=True)
--- a/netbox/dcim/api/serializers.py
+++ b/netbox/dcim/api/serializers.py
@@ -925,6 +925,9 @@ class InterfaceSerializer(NetBoxModelSerializer, CabledObjectSerializer, Connect
             'last_updated', 'count_ipaddresses', 'count_fhrp_groups', '_occupied',
         ]
 
+    def get_display(self, obj):
+        return f"{obj.name} ({obj.device})"
+
     def validate(self, data):
 
         # Validate many-to-many VLAN assignments


DanSheps:
While we could wrap this around frontport, rearport and other things such as PSU's, only interfaces are included from all vc members when viewing the API.  The rest are per device only.

abhi1693 (commented):
As a user, when I see the device name in the list, I'd expect the search to return the more filtered result but it returns empty list instead. So, maybe we should either think of updating the search method to include device names or rethink this solution.

ITJamie:
Id personally rather have the context of the device name in the interface even if the device name search doesnt work in the interface box. Its about additional context now that the device/virtualmachine field has been dropped which used to give that context

DanSheps:
> As a user, when I see the device name in the list, I'd expect the search to return the more filtered result but it returns empty list instead. So, maybe we should either think of updating the search method to include device names or rethink this solution.

I think this would be a separate FR IMO, to enable searching by device name.

ITJamie:
Should we also do the same for the virtualmachine interface serialisers?

DanSheps:
> Should we also do the same for the virtualmachine interface serialisers?

Might not be a bad idea, however this is specifically to deal with connecting things, which we don't do with VMs (yet)

ITJamie:
It would show on the ip address assignment section. This current change would affect the device interface list but not the vm interface list

kkthxbye-code:
@DanSheps - He's saying that your change changes the IP assignment selection field for interfaces but not for vm interfaces, which is inconsistent:

Interface:

image

VM Interface:

image

Another place where it is inconsistent is the InventoryItem form:

image

Only the Inventory dropdown will have the device in parenthesis, the seven other dropdowns will not. One could argue that it's not even needed here as there's a device dropdown at the top of the form that filters the component dropdowns.

And of course with custom scripts, an interface Object field will show the device, but all other components will not.

Imo this shouldn't get merged with that inconsistencies. Either everything should be changed to include the device, or another solution should be figured out.

DanSheps:
After discussion in the maintainer meeting, we are going to go in a different direction, this will be closed out and a new PR will be opened to correct the behaviour.

PR #13160, opened 2023-07-12
汉化menu,table,form

wybaccount:


kkthxbye-code:
Please don't create unsolicited PR's.

PR #13172, opened 2023-07-12
Fixes #12768 - Update the install doc for PostgreSQL 15

### LINKED ISSUE #12768: Postgres 15 requires a Schema during installation
### Change Type

Correction

### Area

Installation/upgrade

### Proposed Changes

With the next Debian release coming with Postgres 15 as the default version, there is an interesting change for the configuration of the database for Django. With the current installation steps, you have a "permission denied for schema public" error during the initial migration.

I'll paraphrase my source: https://gist.github.com/axelbdt/74898d80ceee51b69a16b575345e8457

In Postgres 15 [release announcement](https://www.postgresql.org/about/news/postgresql-15-released-2526/):
> Remove PUBLIC creation permission on the [public schema](https://www.postgresql.org/docs/15/ddl-schemas.html#DDL-SCHEMAS-PUBLIC) (Noah Misch)
> The new default is one of the secure schema usage patterns that [Section 5.9.6](https://www.postgresql.org/docs/15/ddl-schemas.html#DDL-SCHEMAS-PATTERNS) has recommended...

In practice, it means that the configuration block in `configuration_example.py` for the database should be changed to:
```python
DATABASE = {
    'ENGINE': 'django.db.backends.postgresql',  # Database engine
    'NAME': 'netbox',         # Database name
    'USER': '',               # PostgreSQL username
    'PASSWORD': '',           # PostgreSQL password
    'HOST': 'localhost',      # Database server
    'PORT': '',               # Database port (leave blank for default)
    'CONN_MAX_AGE': 300,      # Max database connection age
    'OPTIONS': {
        'options': '-c search_path=netbox',
    }
}
```

and in the instructions to create the database in [PostgreSQL Database Installation](https://docs.netbox.dev/en/stable/installation/1-postgresql/):
```sql
CREATE DATABASE netbox;
CREATE USER netbox WITH PASSWORD 'J5brHrAXFLQSif0K';
\connect netbox;
CREATE SCHEMA netbox AUTHORIZATION netbox;
```

As far as I know, these changes are also compatible with Postgres <15. Also, I have not found any bug or issue with this configuration on my test server with Postgres 15.

I do not have (yet) the systems needed to test what happens during a migration from an older Postgres version to Postgres 15 and if some specific actions would be required.

Alef-Burzmali:
### Fixes: #12768

The PR updates the installation documentation to support PostgreSQL 15 and later, which changes the default permissions on the public schema.

As far as I could test, this is a one-time command during the installation. No further actions are needed on NetBox upgrades.

### DIFF
--- a/docs/installation/1-postgresql.md
+++ b/docs/installation/1-postgresql.md
@@ -55,6 +55,9 @@ Within the shell, enter the following commands to create the database and user (
 CREATE DATABASE netbox;
 CREATE USER netbox WITH PASSWORD 'J5brHrAXFLQSif0K';
 ALTER DATABASE netbox OWNER TO netbox;
+-- the next two commands are needed on PostgreSQL 15 and later
+\connect netbox;
+GRANT CREATE ON SCHEMA public TO netbox;
 ```
 
 !!! danger "Use a strong password"


Alef-Burzmali:
Also, I've tested an upgrade from PostgreSQL 13 to PostgreSQL 15, the old permissions on the schema are retained when the cluster is upgraded, so no specific actions are required.

DanSheps (approved):

jeremystretch:
Thanks @Alef-Burzmali!

PR #13174, opened 2023-07-13
13149 add gettext_lazy to form fields

arthanson:
### Fixes: #13149 

Makes sure all form fields have localizable label, also converts all formset titles to use gettext_lazy.


ITJamie:
Can we write a pre-commit / test script to make sure all fields have a label assigned? To make sure we catch any missing during new PR's

jeremystretch:
I have a similar question here as I did on #13254. How do we want to handle ModelForm fields which are automatically derived from the model? I assume translation isn't handled automatically (but maybe it is?), in which case we'd need to account for those field labels as well.

jeremystretch (approved):

Issue #12194, opened 2023-04-06
Add custom field types for standardized location identifiers


jeremystretch:
### NetBox version

v3.4.7

### Feature type

New functionality

### Proposed functionality

Add two new custom fields types:

- [ISO 3166 country code](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes)
- [UN-LOCODE](https://unece.org/trade/cefact/unlocode-code-list-country-and-territory)

Custom fields of either type would render as selection fields listing only valid values as selectable options. These would be maintained within NetBox core, with no additional setup needed on the part of the administrator.

### Use case

These are well-known values with a wide range of useful applications concerning the assignment of location data. For example, it is common practice to designate the LOCODE for each site on a network.

### Database changes

N/A

### External dependencies

There are a few ways we might source this data, such as the [locode Python package](https://pypi.org/project/locode/). Further evaluation of the available options is needed.

BarbarossaTM:
This would be awesome, especially UN-LOCODE!

Do you have some idea in mind how to model this? Like for example would there be two levels for country and city?

jeremystretch:
My initial thought here was to introduce new custom field types to hold the types of data mentioned above, but in retrospect that seems like a poor approach that won't scale well. Instead, I've just opened #12988 which proposes moving stored custom field choices into a separate model, which will allow for the efficient reuse of choice sets across multiple custom fields.

jeremystretch:
Another set to consider is [CLLI](https://en.wikipedia.org/wiki/CLLI_code).

PR #13195, opened 2023-07-17
Closes #12988: Introduce custom field choice sets

### LINKED ISSUE #12988: Introduce choice sets for custom fields
### NetBox version

v3.5.4

### Feature type

Data model extension

### Proposed functionality

Currently, custom field choices are stored as an array of values on each applicable `CustomField` instance:

```python
class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
    choices = ArrayField(
        base_field=models.CharField(max_length=100),
        blank=True,
        null=True,
        help_text=_('Comma-separated list of available choices (for selection fields)')
    )
```

This issue proposes essentially moving the `choices` field to a separate model, `CustomFieldChoiceSet`:

```python
class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
    choices = models.ForeignKey(
        to='CustomFieldChoiceSet',
        blank=True,
        null=True
    )


class CustomFieldChoiceSet(ChangeLoggedModel):
    name = models.CharField()
    description = models.CharField(blank=True)
    choices = ArrayField(
        base_field=models.CharField(max_length=100),
        help_text=_('Comma-separated list of available choices (for selection fields)')
    )
```

Individual choice values will continue to be managed as flat lists, but now are defined on separate `CustomFieldChoiceSet` instances, which can be referenced by multiple custom fields.

As part of this work, I'd also like to improve the manner in which custom field choices are managed by the end user. We can extend the current view to allow creating, modifying, and deleting choice values as if they were individual objects (as opposed to a list of values in a single JSON form field).

This _will_ result in a breaking API change for custom fields themselves, however custom field data on NetBox objects should not be affected by the change.

### Use case

This change was prompted by #12194, which seeks to introduce reusable choice sets containing well-known values (e.g. UN-LOCODE identifiers). Moving the choice sets into a separate model allows for their efficient reuse across multiple custom fields without introducing significant overhead. (The choice sets can be easily pre-fetched along with the custom fields themselves.)

### Database changes

- Introduce the new CustomFieldChoiceSet model (above)
- Add a ForeignKey from CustomField to CustomFieldChoiceSet
- Migrate all existing `choices` values to new CustomFieldChoiceSet instances
- Drop the old `choices` field from the CustomField model

### External dependencies

N/A

jeremystretch:
### Fixes: #12988

- Introduce the new `CustomFieldChoiceSet` model
- Automatically migrate all CF choices to the new model
- Add a `choice_set` ForeignKey to `CustomField` and remove the `choices` array field
- Introduce `ArrayWidget` to improve UX while editing choices


netbox/extras/models/customfields.py
```diff
@@ -627,3 +631,46 @@ def validate(self, value):
...
+        max_length=100,
+        unique=True
+    )
+    description = models.CharField(
+        max_length=200,
+        blank=True
+    )
+    extra_choices = ArrayField(
+        base_field=models.CharField(max_length=100),
+        help_text=_('Comma-separated list of available choices (for selection fields)')
```
DanSheps, line 650: Should this be new-line instead of CSV?  In the form it mentions new-line.  DB it is stored as json however.

```suggestion
        help_text=_('Line-separated list of available choices (for selection fields)')
```

netbox/extras/forms/model_forms.py
```diff
@@ -78,6 +84,20 @@ def __init__(self, *args, **kwargs):
...
+        widget=ArrayWidget(),
+        help_text=_('Enter one choice per line.')
+    )
+
+    class Meta:
+        model = CustomFieldChoiceSet
+        fields = ('name', 'description', 'extra_choices', 'order_alphabetically')
+
+    def clean_extra_choices(self):
+        return self.cleaned_data['extra_choices'].split('\n')
```
DanSheps, line 98: This does not account for Windows users who get CRLF instead of just LF.

```suggestion
        return self.cleaned_data['extra_choices'].replace('\r\n','\n').split('\n')
```

docs/customization/custom-fields.md
```diff
@@ -60,7 +60,7 @@ NetBox supports limited custom validation for custom field values. Following are
 
 ### Custom Selection Fields
 
-Each custom selection field must have at least two choices. These are specified as a comma-separated list. Choices appear in forms in the order they are listed. Note that choice values are saved exactly as they appear, so it's best to avoid superfluous punctuation or symbols where possible.
+Each custom selection field must designate a [choice set](../models/extras/customfieldchoiceset.md) containing at least two choices. These are specified as a comma-separated list.
```
DanSheps, line 63:
```suggestion
Each custom selection field must designate a [choice set](../models/extras/customfieldchoiceset.md) containing at least two choices. These are specified as a line-separated list.
```

docs/models/extras/customfieldchoiceset.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Fields
+
+### Name
+
+The human-friendly name of the choice set.
+
+### Extra Choices
+
+The list of valid choices, entered as a comma-separated list.
```
DanSheps, line 13:
```suggestion
The list of valid choices, entered as a line-separated list.
```

netbox/extras/forms/bulk_import.py
```diff
@@ -53,8 +57,22 @@ class Meta:
...
+            'search_weight', 'filter_logic', 'default', 'choice_set', 'weight', 'validation_minimum',
+            'validation_maximum', 'validation_regex', 'ui_visibility', 'is_cloneable',
+        )
+
+
+class CustomFieldChoiceSetImportForm(CSVModelForm):
+    extra_choices = SimpleArrayField(
+        base_field=forms.CharField(),
+        required=False,
+        help_text=_('Comma-separated list of field choices')
```
DanSheps, line 69:
```suggestion
        help_text=_('Line-separated list of field choices')
```

DanSheps (changes requested):
Line separated appears to have issues when used on Windows clients.  Quick and dirty fix is to replace CRLF with LF only.

netbox/extras/forms/model_forms.py
jeremystretch: This should use `splitlines()` to catch both cases.

jeremystretch:
Thanks @DanSheps. I'm going to hold off on the documentation updates as I'll need to rewrite the descriptions for `extra_fields` for #12194 anyway.

netbox/extras/forms/filtersets.py
```diff
@@ -37,7 +38,8 @@ class CustomFieldFilterForm(SavedFiltersMixin, FilterForm):
     fieldsets = (
         (None, ('q', 'filter_id')),
         ('Attributes', (
```
arthanson, line 40: add localization to 'Attributes' use gettext_lazy

netbox/extras/forms/model_forms.py
```diff
@@ -50,13 +52,17 @@ class CustomFieldForm(BootstrapMixin, forms.ModelForm):
         required=False,
         help_text=_("Type of the related object (for object/multi-object fields only)")
     )
+    choice_set = DynamicModelChoiceField(
+        queryset=CustomFieldChoiceSet.objects.all(),
+        required=False
+    )
 
     fieldsets = (
         ('Custom Field', (
```
arthanson, line 61: Add localization to these fieldset headers

netbox/extras/models/customfields.py
```diff
@@ -278,22 +286,18 @@ def clean(self):
...
-        ) and not self.choices:
+        ):
+            if not self.choice_set:
+                raise ValidationError({
+                    'choice_set': "Selection fields must define a set of choices."
+                })
+        elif self.choice_set:
             raise ValidationError({
-                'choices': "Selection fields must specify at least one choice."
+                'choice_set': "Choices may be set only for selection fields."
```
arthanson, line 300: internationalize - wrap with _( (gettext_lazy)

netbox/extras/models/customfields.py
```diff
@@ -627,3 +631,46 @@ def validate(self, value):
...
+    name = models.CharField(
+        max_length=100,
+        unique=True
+    )
+    description = models.CharField(
+        max_length=200,
+        blank=True
+    )
+    extra_choices = ArrayField(
+        base_field=models.CharField(max_length=100),
```
arthanson, line 649: Not to change here but in relation to the issue of increasing field sizes - these max length are a bit of a magic number in that the choices and selection fields must match and are 100, where description are 200 - seems like these should be constants in the code. Just mentioning as I'm thinking of it, can open a new issue if you agree.

netbox/templates/extras/customfieldchoiceset.html
```diff
@@ -0,0 +1,64 @@
+{% extends 'generic/object.html' %}
+{% load helpers %}
+{% load plugins %}
+
+{% block content %}
+  <div class="row mb-3">
+    <div class="col col-md-6">
+      <div class="card">
+        <h5 class="card-header">Custom Field Choice Set</h5>
```
arthanson, line 9: Add translate template tags round text.  I'll need to do the one above as it will have merge conflicts.

arthanson (changes requested):
Also noticed when you add a custom field "Choice set" is only valid if you have selection type - we should add a PR to change the form to use HTMX so "Choice Set" is only displayed if you have a selection type.

netbox/extras/forms/model_forms.py
jeremystretch: These weren't changed as part of the PR

netbox/extras/models/customfields.py
jeremystretch: No need to change the field size; we're simply porting what has existed for years.

DanSheps:
> Thanks @DanSheps. I'm going to hold off on the documentation updates as I'll need to rewrite the descriptions for `extra_fields` for #12194 anyway.

Yeah, no problem.

The splitlines() is way better then my solution anyways, and I always forget about it.

PR #13201, opened 2023-07-18
Moves non-racked devices to tab

### LINKED ISSUE #11803: Move non-racked devices to a separate tab under the rack view
### NetBox version

v3.4.5

### Feature type

### Proposed functionality

Set the list "Non-Racked Devices" via the rack or as a separate tab/register

image

Maybe even with a signal colour so that you can recognise more quickly that there is a special feature here, which is not the case with every rack.
image

THX to @abhi1693 for the basic image.

### Use case

I have been thinking about raising this point for months. But I realise again and again that it is important.

Racks with 42 units are so long that you can't see the "non-racked devices" area on a 24" monitor. That's why you don't pay attention to them. Either they are devices that are not sorted or devices that cannot be installed in a unit, e.g. socket outlets, small devices that are screwed to the rack, e.g. temperature monitors. 

Translated with www.DeepL.com/Translator (free version)

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -681,13 +681,6 @@ class RackView(generic.ObjectView):
             (PowerFeed.objects.restrict(request.user).filter(rack=instance), 'rack_id'),
         )
 
-        # Get 0U devices located within the rack
-        nonracked_devices = Device.objects.filter(
-            rack=instance,
-            position__isnull=True,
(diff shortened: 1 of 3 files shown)


jeremystretch:
Thanks @abhi1693!

PR #13202, opened 2023-07-18
Adds faster polling for scripts and reports

### LINKED ISSUE #13097: Faster initial poll for script responses
### NetBox version

v3.5.4

### Feature type

### Proposed functionality

One of the downsides of scripts now being run as background tasks, with the results being polled, is that even a trivial script takes 5 seconds to produce a response.

e.g. even a simple script like this, still takes 5 seconds before its result is shown in the UI:

image

I propose that either:
* the first poll is done 1 second after the job is submitted, and then 5 seconds thereafter (I expect that to be a very simple change)
* the first few polls are done at 0.5 and 1 second intervals, after which it reverts to 5 seconds

Some sort of COMET / long-poll / server-sent event would clearly be a better solution, so that the result is returned as soon as it's available, but that would more work to implement.

### Use case

When scripts are being used to automate data entry tasks, a 5 second delay on every form submission is long enough to become annoying quickly.

### Database changes

None

### External dependencies

None

abhi1693:

### DIFF
--- a/netbox/templates/extras/report_result.html
+++ b/netbox/templates/extras/report_result.html
@@ -4,7 +4,7 @@
 
 {% block content-wrapper %}
   <div class="row p-3">
-    <div class="col col-md-12"{% if not job.completed %} hx-get="{% url 'extras:report_result' job_pk=job.pk %}" hx-trigger="every 5s"{% endif %}>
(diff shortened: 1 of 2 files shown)


jeremystretch:
In testing, I only see HTMX requests at 500ms and at 5sec intervals. It doesn't seem like the 1sec or 2sec delay timers are in effect.

IMO a 1- or 2-second initial delay timer is probably sufficient anyway.

abhi1693:
@jeremystretch Check now, I have removed the extra load definitions

jeremystretch:
Thanks @abhi1693!

PR #13203, opened 2023-07-18
Adds sensitive_parameters to DataBackend

### LINKED ISSUE #12625: Datasource passwords are displayed in plaintext

abhi1693:
### Fixes: #12625

I have added a new parameter `sensitive_parameters` which takes a list of fields that should be considered to contain confidential information like keys and passwords.

@jeremystretch I'm not sure if I should keep the perms check in the template, kindly advise.

### DIFF
--- a/netbox/core/data_backends.py
+++ b/netbox/core/data_backends.py
@@ -41,6 +41,7 @@ def register_backend(name):
 
 class DataBackend:
     parameters = {}
+    sensitive_parameters = []
 
     def __init__(self, url, **kwargs):
         self.url = url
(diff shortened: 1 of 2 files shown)


jeremystretch:
> I'm not sure if I should keep the perms check in the template, kindly advise.

Makes sense to me. Thanks for taking this!

PR #13243, opened 2023-07-23
Fixes #13233 - Remove vlan_group from nullable fields in InterfaceBulkEditForm

### LINKED ISSUE #13233: Interface Bulk edit fails if vlan_group is set to null
### NetBox version

v3.5.6

### Python version

3.8

### Steps to Reproduce

1. Select multiple Interfaces and click on Edit below the list.
2. Activate `Set Null` for the `802.1Q Switching`>` VLAN group` option
3. Click on save

### Expected Behavior

There is no `Set Null` option as the group is used for filtering only or at least no error is raised.

### Observed Behavior

 Server Error  is raised.
```
<class 'AttributeError'>

'NoneType' object has no attribute 'null'

Python version: 3.8.10
NetBox version: 3.5.6
```

kkthxbye-code:
### Fixes: #13233 

vlan_group does not exist on the Interface model and as such can't be nulled.


jeremystretch:
Thanks @kkthxbye-code!

PR #13247, opened 2023-07-23
Fixes #13051 - Allow table cells to be aligned.

### LINKED ISSUE #13051: Unable to Align Markdown Tables Values
### NetBox version

3.5.1 (also tested on 3.5.4)

### Python version

3.10

### Steps to Reproduce

1. Edit any object that supports the Comments field
2. Add a table with colons on one or both sides of the hyphens in header row, thus indicating desired alignment, such as:

```
| Left Alignment| Center Alignment | Right Aligment     |
| :---        |    :----:   |          ---: |
| Words     | Title       | Here's this   |
| Values   | Text        | And more      |
```

3. Either preview the table on the Edit page, or save the object and view the Comments field

### Expected Behavior

I expected the table to be aligned per [the documentation on markdown support](https://docs.netbox.dev/en/stable/reference/markdown/), which would involve it appearing like below:

| Left Alignment| Center Alignment | Right Aligment     |
| :---        |    :----:   |          ---: |
| Words     | Title       | Here's this   |
| Values   | Text        | And more      |

### Observed Behavior

The colons seem to be totally ignored, as shown in the screenshots below (using the same markdown syntax as above):

image

image

kkthxbye-code:
### Fixes: #13051

See reasoning for CSS changes here: https://github.com/netbox-community/netbox/issues/13051#issuecomment-1646956815


kkthxbye-code:
Tried it out briefly. It installs fine, haven't checked for differences in mkdocs though, but it builds. However I wanted the upgrade because the TableExtension would use the style attribute instead, which makes the css changes unnecessary, however with bleach it's not enough to allow the style attribute:

https://bleach.readthedocs.io/en/latest/clean.html#sanitizing-css

You have to install the additional package bleach[css] and pass a CSSSanitizer to the sanitizer. I can do the work, but there's a bunch of other changes to mkdocs that we need to verify doesn't break the docs. On top of that, bleach is deprecated (still going to get security patches I think).

Not sure what the best option is here. However if we keep the current fix, we need to remember that when we upgrade python-markdown to 3.4.x, we need to pass the `use_align_attribute=True` to TableExtension - otherwise tables will break because they made the new way of doing it the default.

Simple stuff is never simple...

jeremystretch:
In that case, I have no problem merging this as-is. Thanks for digging so far into this!

PR #13248, opened 2023-07-23
Fixes #13237 - Allow unauthenticated api access to content-types.

### LINKED ISSUE #13237: Content-Types only fetchable as logged in user
### NetBox version

v3.5.6

### Python version

3.11

### Steps to Reproduce

Config settings:
```
LOGIN_REQUIRED = False
EXEMPT_VIEW_PERMISSIONS = ["*"]
```

Try to access `/api/extras/content-types/`

### Expected Behavior

List of content types is returned.

### Observed Behavior

`{"detail":"Authentication credentials were not provided."}`

This view only uses `IsAuthenticated` while other views use `IsAuthenticatedOrLoginNotRequired`, which I believe is what should be used for `ContentTypeViewSet` as well.

fabi125:
### Fixes: #13237

Allow unauthenticated access to `/api/extras/content-types/` if login is not required for Netbox.

The rest of the file uses lists for the permissions, so I switched this view to a list as well (instead of a tuple).

### DIFF
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -6,7 +6,6 @@ from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.exceptions import PermissionDenied
 from rest_framework.generics import RetrieveUpdateDestroyAPIView
-from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @fabi125!

Issue #13061, opened 2023-06-30
Script Rest API has no results


jiuka:
### NetBox version

v3.5.4

### Python version

3.11

### Steps to Reproduce

1. Save this script into a file called `bug.py`
```
from extras.scripts import *


class Bug13061(Script):

    class Meta:
        name = "Demo for Bug13061"
        description = "Demo for Bug13061"

    def run(self, data, commit):
        self.log_success(f"Demo for Bug13061")
```
2. Go to https://netbox.example.com/extras/scripts/add/
3. Upload the `bug.py` file.
4. Got to https://netbox.example.com/extras/scripts/bug/Bug13061/
5. Click on `Run Script`
6. Go to https://netbox.example.com/api/extras/scripts/bug.Bug13061/

### Expected Behavior

The `result` should be job result like in the reports endpoint.

### Observed Behavior

It is always `null`.

The [job.name](https://github.com/netbox-community/netbox/blob/develop/netbox/extras/api/views.py#L293C13-L293C21) is just the name of the script class. When search for the result the [script.name
](https://github.com/netbox-community/netbox/blob/develop/netbox/extras/api/views.py#L306C41-L306C52) is used and not the `class_name` therefore no results are found, unless the `name` ist set to the same value as the `class_name`.

abhi1693:
Thank you for opening a bug report. Unfortunately, the information you have provided is not sufficient for someone else to attempt to reproduce the reported behavior. Remember, each bug report must include detailed steps that someone else can follow on a clean, empty NetBox installation to reproduce the exact problem you're experiencing. These instructions should include the creation of any involved objects, any configuration changes, and complete accounting of the actions being taken. Also be sure that your report does not reference data on the public NetBox demo, as that is subject to change at any time by an outside party and cannot be relied upon for bug reports.

jiuka:
I hope it is detailed enough.

bitcollector1:
I have the same issue. Here you can see an example of just (1) of my scripts reported correctly, even though all of them completed successfully. 

![Screenshot 2023-07-05 at 09 28 52](https://github.com/netbox-community/netbox/assets/50723251/9df5650c-e361-45c5-8cba-2c1530afcbf1)

jiuka:
Hi @jeremystretch 

This fixed the GUI but not the Rest API I made bug about. Should I create a new bug report or do you want to reopen this one?

Regards and many thanks for Netbox

 Marius

cholyoak:
[...]

PR #13250, opened 2023-07-24
Fixes #13167- Use class_name instead of name to get script results

### LINKED ISSUE #13167: Script Rest API has no results
### NetBox version

v3.5.6

### Python version

3.11

### Steps to Reproduce

As reported in #13061 not only the UI had this issue but the Rest API too.

1. Save this script into a file called bug.py
```
from extras.scripts import *


class Bug13061(Script):

    class Meta:
        name = "Demo for Bug13061"
        description = "Demo for Bug13061"

    def run(self, data, commit):
        self.log_success(f"Demo for Bug13061")
```
2. Go to https://netbox.example.com/extras/scripts/add/
3. Upload the bug.py file.
4. Got to https://netbox.example.com/extras/scripts/bug/Bug13061/
5. Click on Run Script
6. Go to https://netbox.example.com/api/extras/scripts/bug.Bug13061/

### Expected Behavior

The `result` should be job result like in the reports endpoint.

### Observed Behavior

It is always `null`.

The [job.name](https://github.com/netbox-community/netbox/blob/develop/netbox/extras/api/views.py#L293C13-L293C21) is just the name of the script class. When search for the result the [script.name](https://github.com/netbox-community/netbox/blob/develop/netbox/extras/api/views.py#L306C41-L306C52)
is used and not the `class_name` therefore no results are found, unless the `name` ist set to the same value as the `class_name`.

kkthxbye-code:
### Fixes: #13167

This is actually a fix to #13061, but as that was closed by solving a related but different issue, #13167 was opened while technically being a duplicate.

### DIFF
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -303,7 +303,7 @@ class ScriptViewSet(ViewSet):
 
         # Attach Job objects to each script (if any)
         for script in script_list:
-            script.result = results.get(script.name, None)
+            script.result = results.get(script.class_name, None)
 
         serializer = serializers.ScriptSerializer(script_list, many=True, context={'request': request})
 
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @kkthxbye-code!

PR #13290, opened 2023-07-28
13279 update choices to use gettext_lazy

arthanson:


jeremystretch:
While I appreciate the consistency, I'm not sure it makes sense to provide translation support for things like interface types and wireless channels. We might want to stick to the arbitrary strings like statuses and colors.

netbox/dcim/choices.py
```diff
@@ -120,8 +121,8 @@ class RackElevationDetailRenderChoices(ChoiceSet):
     RENDER_SVG = 'svg'
 
     CHOICES = (
```
jsenecal, line 123: Do we need to translate these ?

netbox/utilities/choices.py
```diff
@@ -222,8 +223,8 @@ class ImportFormatChoices(ChoiceSet):
     YAML = 'yaml'
 
     CHOICES = [
-        (AUTO, 'Auto-detect'),
-        (CSV, 'CSV'),
-        (JSON, 'JSON'),
-        (YAML, 'YAML'),
+        (AUTO, _('Auto-detect')),
+        (CSV, _('CSV')),
```
jsenecal, line 227: Again, do we want to support translating file formats ?

jsenecal (commented):
I added a few notes about supporting translation on things that perhaps shouldn't be but otherwise this looks good to me. Thanks @arthanson

netbox/utilities/choices.py
jeremystretch: I've made the presumption elsewhere that acronyms like JSON, YAML, CSV, and so on do _not_ need to be translated. In this specific case, we'd want to translate the "auto-detect" option however.

jsenecal: I agree, my comment started on L227+

jeremystretch (approved):

netbox/dcim/choices.py
jeremystretch: No; these are just file extensions.

PR #13296, opened 2023-07-28
Fix #11478 - Add vc_interfaces flag to control selection of VC interfaces

### LINKED ISSUE #11478: Patching front port to virtual chassis member interface is broken in UI

DanSheps:
### Fixes: #11478 - Add vc_interfaces flag to control selection of VC interfaces

* Create a `vc_interfaces` flag that can be used for controlling device queryset generation.
* Update interface edit form to account for new flag


netbox/dcim/filtersets.py
```diff
@@ -1511,22 +1511,16 @@ class Meta:
...
-            return queryset.filter(pk__in=vc_interface_ids)
-        except Device.DoesNotExist:
-            return queryset.none()
-
-    def filter_device_id(self, queryset, name, id_list):
         # Include interfaces belonging to peer virtual chassis members
         vc_interface_ids = []
         try:
-            devices = Device.objects.filter(pk__in=id_list)
+            devices = Device.objects.filter(**{'{}__in'.format(name): value})
```
jsenecal, line 1517: btw f-strings are faster than str. format() because f-strings are evaluated at compile-time rather than at runtime.

DanSheps: I was just following the established convention.  I would say if we switch to f-strings we should do it on all filtersets at once.

jsenecal (approved):

netbox/dcim/filtersets.py
```diff
@@ -1511,22 +1511,16 @@ class Meta:
...
             for device in devices:
-                vc_interface_ids += device.vc_interfaces(if_master=False).values_list('id', flat=True)
+                # Hack to show all VC member interfaces when requested
+                if self.request is not None and 'vc_interfaces' in self.request.GET.keys():
```
jeremystretch, line 1520: We can't support this pattern. Any separate query parameter needs to have its own filter defined on the filter set for validation and documentation. We'll need to figure out a different mechanism to toggle the inclusion of VC interfaces. Defining an entirely separate filter, while perhaps not ideal, is one option.

jeremystretch (changes requested):

netbox/dcim/filtersets.py
DanSheps: Fair enough,

I think it might be best to consider this a breaking change, rebase to feature and then remove the custom logic for both.  Instead device returns only specific devices.

With that, introduce a new filter, `virtual_chassis` that will return all the interfaces as part of a virtual chassis.

netbox/dcim/filtersets.py
```diff
@@ -1462,17 +1462,16 @@ class InterfaceFilterSet(
     PathEndpointFilterSet,
     CommonInterfaceFilterSet
 ):
-    # Override device and device_id filters from DeviceComponentFilterSet to match against any peer virtual chassis
-    # members
-    device = MultiValueCharFilter(
-        method='filter_device',
-        field_name='name',
-        label=_('Device'),
+    virtual_chassis_for_device = django_filters.ModelMultipleChoiceFilter(
```
jeremystretch, line 1465: This name is a bit confusing IMO. How about something like `virtual_chassis_member`?

jeremystretch (changes requested):

netbox/dcim/filtersets.py
DanSheps: That works, I can get that fixed right away.

netbox/dcim/tests/test_filtersets.py
```diff
@@ -3116,6 +3116,13 @@ def test_device(self):
         params = {'device': [devices[0].name, devices[1].name]}
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
 
+    def test_virtual_chassis_for_device(self):
+        devices = Device.objects.filter(name__in=['Device 1', 'Device 3'])
+        params = {'device_id': [devices[0].pk, devices[1].pk]}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+        params = {'device': [devices[0].name, devices[1].name]}
```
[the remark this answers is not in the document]
DanSheps: Thanks for the catch, I will get that fixed up

netbox/dcim/tests/test_filtersets.py
```diff
@@ -2822,11 +2822,44 @@ def setUpTestData(cls):
...
+
         devices = (
-            Device(name='Device 1', device_type=device_types[0], role=roles[0], site=sites[0], location=locations[0], rack=racks[0]),
-            Device(name='Device 2', device_type=device_types[1], role=roles[1], site=sites[1], location=locations[1], rack=racks[1]),
-            Device(name='Device 3', device_type=device_types[2], role=roles[2], site=sites[2], location=locations[2], rack=racks[2]),
-            Device(name=None, device_type=device_types[2], role=roles[2], site=sites[3]),  # For cable connections
+            Device(
+                name='Device 1',
+                device_type=device_types[0],
+                role=roles[0], site=sites[0],
```
jeremystretch, line 2833: Inconsistent formatting (each argument should go on a new line).

netbox/dcim/tests/test_filtersets.py
```diff
@@ -2822,11 +2822,44 @@ def setUpTestData(cls):
...
+            ),
+            Device(
+                name=None,
+                device_type=device_types[2],
+                role=roles[2],
+                site=sites[3],
+                virtual_chassis=virtual_chassis,
+                vc_position=2,
+                vc_priority=1
+            ),  # For cable connections
```
jeremystretch, line 2862: Should move the comment above the instance definition for clarity

netbox/dcim/tests/test_filtersets.py
```diff
@@ -3116,6 +3144,15 @@ def test_device(self):
         params = {'device': [devices[0].name, devices[1].name]}
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
 
+    def test_virtual_chassis_member(self):
+        # Device 1 & 3 have 1 management interface, Device "None" has 5 interfaces, 2 of which are management and
+        # will not be visible
```
jeremystretch, line 3149: The fourth device exists solely to serve as the opposite end of the various cable connections. It would probably be simpler to avoid using it as a VC member, and instead create a separate device. Then we could have e.g. Device 1A and Device 1B, so it's more readily apparent that a VC is in use. (We also don't want to reference "Device None:" If it needs to be referenced, we should give the device a proper name.)

jeremystretch (changes requested):
Could you also add a note to the v3.6 release notes indicating the breaking change to the `device` & `device_id` filters for interfaces?

netbox/dcim/tests/test_filtersets.py
DanSheps: Fixed this up

netbox/dcim/tests/test_filtersets.py
DanSheps: I am going to rename "Device 1" to "Device 1A" and then add in "Device 1B" to the mix with a new interface.  I am going to try and exclude 1B from as many of the other tests as possible (by using a non-tested role/device_type/etc)

DanSheps:
I think I covered everything @jeremystretch

PR #13301, opened 2023-07-29
Adds assigned bool for IP address API

### LINKED ISSUE #13151: Add assigned parameter to the ipam/ip-addresses API endpoint
### NetBox version

v3.5.5

### Feature type

### Proposed functionality

I propose to add an `assigned` boolean parameter to the ipam/ip-addresses API endpoint to handle filtering IP addresses by assignments to both interfaces and FHRP groups. The `assigned_to_interface` parameter currently supports filtering for assignments to interfaces only.

### Use case

Adding this parameter would allow users to retrieve assigned or unassigned IP addresses via the API with regard to interface and FHRP group assignments. This would help an API integration task we have in place. Currently we are getting all IP addresses and determining if the `assigned_object` key has a value as a workaround. This would also better align the API with the UI table for IP Addresses as it also has an `Assigned` column.

### Database changes

None

### External dependencies

None

abhi1693:

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -591,6 +591,10 @@ class IPAddressFilterSet(NetBoxModelFilterSet, TenancyFilterSet):
         method='_assigned_to_interface',
         label=_('Is assigned to an interface'),
     )
+    assigned = django_filters.BooleanFilter(
+        method='_assigned',
+        label=_('Is assigned'),
+    )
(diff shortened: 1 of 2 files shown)


jsenecal (approved):

jeremystretch:
When adding new filters, don't forget to include a relevant rest inside `test_filtersets.py`. Thanks @abhi1693!

jeremystretch:
CI failures are linter errors that have been fixed in `develop`.

PR #13332, opened 2023-08-01
Closes #11732: Protect against errant overwriting of data via web UI forms

### LINKED ISSUE #11732: Prevent race condition when using edit forms in GUI
### NetBox version

v3.4.4

### Feature type

### Proposed functionality

In edit forms for primary models, when confirming/saving your changes, add a new validation check that the `last_updated` time has not changed from when you first started editing the object.

If the `last_updated` time is captured at the time the edit form is loaded, and check again when the `Save` button is clicked, we can check that the timestamp is the same as before, before finally saving the new change. Otherwise, a validation error is raised.

### Use case

It's possible for two users to be editing an object in the GUI at the same time and overwrite each other's changes. Here's the scenario:

1. Bob creates a Site.
2. Joe clicks “Edit” on that Site and has the Edit Site form open.
3. Bob also clicks Edit on the same Site and makes some changes, clicks “Save”.
4. Joe make changes on his Edit Form and clicks save.
5. Bob’s changes are no longer present, and Joe’s change is - based on outdated data.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #11732

- Introduce CheckLastUpdatedMixin, which adds the optional field `_init_time` and validation to check against an object's `last_updated` value
- NetBoxModelForm now inherits from CheckLastUpdatedMixin

### DIFF
--- a/netbox/netbox/forms/base.py
+++ b/netbox/netbox/forms/base.py
@@ -6,7 +6,7 @@ from django.utils.translation import gettext_lazy as _
 from extras.choices import CustomFieldFilterLogicChoices, CustomFieldTypeChoices, CustomFieldVisibilityChoices
 from extras.forms.mixins import CustomFieldsMixin, SavedFiltersMixin
 from extras.models import CustomField, Tag
-from utilities.forms import BootstrapMixin, CSVModelForm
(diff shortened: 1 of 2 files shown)


jsenecal (approved):
Looks good and elegant to me!

PR #13337, opened 2023-08-01
Fixes: #11079 - Handle cables across multiple rear-port positions

### LINKED ISSUE #11079: cable with multiple connections can't cross module front/rear ports.
### NetBox version

v3.3.9

### Python version

3.9

### Steps to Reproduce

We use a module fiber cassette system - these are cassettes you plug into 1u rack mount.

Trying to plug a MPO to LC breakout cable can cross these modules - the rear ports a single MPO and the front ports are 6 pairs of LC connectors.  So the break out cable we use is a 1 MPO (ie, 100G qsfp) to 4 pair of LC's.  This can cross between two modules.

1) create a device.  aka 'sw-test'
2)  add to this device a network interface.  any type will do, except lag or virtual.

3) create a module_type, with no components add.

3) create a device to use as a patch panel.
4) in that patch panel, create two module bays, 'C1' and 'C2'.
5) populate the module bay with the above module type.

6) for module C1 add rear port of 'MPO-1', positions of 6
7) for module C2 add rear port of 'MPO-2', positions of 6

8) for module C1, add front ports of 'LC-[1-6]', type 'LC', attached to rear-port MPO-1:[1-6]
9) for module C2, add front ports of 'LC-[7-12]', type 'LC', attached to rear-port MPO-2:[1-6]

See attached screen shot for module/cassette/front/rear setup.
Screenshot_2022-12-01_15-02-22

10) go to the test device you created.

11) for the on the network interface, tell it you want to make a connection.  Choose front port.

for the termination_b, choose LC-6 (associated with MPO-1) and LC-7 (associated with MPO-2)

Screenshot_2022-12-01_15-08-24

Get error message in browser.

Screenshot_2022-12-01_15-08-46

Stack dump from our test netbox instance, doing the same thing:

```Internal Server Error: /api/dcim/cables/
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/core/handlers/exception.py", line 55, in inner
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/viewsets.py", line 125, in view
  File "/opt/netbox/netbox/netbox/api/viewsets/__init__.py", line 118, in dispatch
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/mixins.py", line 19, in create
  File "/opt/netbox/netbox/netbox/api/viewsets/__init__.py", line 157, in perform_create
    instance = serializer.save()
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/serializers.py", line 212, in save
    self.instance = self.create(validated_data)
  File "/opt/netbox/netbox/netbox/api/serializers/features.py", line 56, in create
    instance = super().create(validated_data)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/serializers.py", line 962, in create
    instance = ModelClass._default_manager.create(**validated_data)
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/db/models/query.py", line 514, in create
    obj.save(force_insert=True, using=self.db)
  File "/opt/netbox/netbox/dcim/models/cables.py", line 207, in save
    trace_paths.send(Cable, instance=self, created=_created)
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 176, in send
    return [
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 177, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
  File "/opt/netbox/netbox/dcim/signals.py", line 97, in update_connected_endpoints
    create_cablepath(nodes)
  File "/opt/netbox/netbox/dcim/utils.py", line 43, in create_cablepath
    cp = CablePath.from_origin(terminations)
  File "/opt/netbox/netbox/dcim/models/cables.py", line 543, in from_origin
    assert all(rp.positions == 1 for rp in rear_ports)
AssertionError
```

### Expected Behavior

Can attach a breakout cable across modules/frontports

### Observed Behavior

```
Internal Server Error: /api/dcim/cables/
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/core/handlers/exception.py", line 55, in inner
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/viewsets.py", line 125, in view
  File "/opt/netbox/netbox/netbox/api/viewsets/__init__.py", line 118, in dispatch
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/mixins.py", line 19, in create
  File "/opt/netbox/netbox/netbox/api/viewsets/__init__.py", line 157, in perform_create
    instance = serializer.save()
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/serializers.py", line 212, in save
    self.instance = self.create(validated_data)
  File "/opt/netbox/netbox/netbox/api/serializers/features.py", line 56, in create
    instance = super().create(validated_data)
  File "/opt/netbox/venv/lib/python3.10/site-packages/rest_framework/serializers.py", line 962, in create
    instance = ModelClass._default_manager.create(**validated_data)
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/db/models/query.py", line 514, in create
    obj.save(force_insert=True, using=self.db)
  File "/opt/netbox/netbox/dcim/models/cables.py", line 207, in save
    trace_paths.send(Cable, instance=self, created=_created)
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 176, in send
    return [
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 177, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
  File "/opt/netbox/netbox/dcim/signals.py", line 97, in update_connected_endpoints
    create_cablepath(nodes)
  File "/opt/netbox/netbox/dcim/utils.py", line 43, in create_cablepath
    cp = CablePath.from_origin(terminations)
  File "/opt/netbox/netbox/dcim/models/cables.py", line 543, in from_origin
    assert all(rp.positions == 1 for rp in rear_ports)
AssertionError
```

### LINKED ISSUE #11901: Index out of range bug for PATCH requests to cables API
### NetBox version

v3.4.5

### Python version

3.8

### Steps to Reproduce

Hi Everyone, 

Found an issue sending PATCH/PUT API requests to /cables/{id}/

Any request body that requests a change in the cable termination throws an index out of range error.

Steps to reproduce:
- create three identical devices with identical (manufacturer/type etc) configuration and racked  
- Create two 1GE interfaces, one for each device
- create a cable connection between between two of the devices
- request an API PATCH HTTP request to change the b termination of the cable to the third device. 
- this then throws a HTTP500 error with the below html content. 

Below shows the full HTTP request:

```
curl --location --request PATCH 'https://DOMAIN/api/dcim/cables/425/' \
--header 'Authorization: Token MYTOKEN' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--data-raw '{
     "b_terminations": [
        {
          "object_id": 15
        }
    ]
}'
```

See attached file for the backtrace, with some stuff removed for security:
[Internal Server Error netboxapidcim.txt](https://github.com/netbox-community/netbox/files/10884218/Internal.Server.Error.netboxapidcim.txt)

Not sure much else I can provide - let me know :)

TIA

### Expected Behavior

HTTP response 200 with an updates connection with termination changes.

### Observed Behavior

HTTP response 500 with an index out of range error with no changes to the cable.

Further, shows the following HTML file:

[error.txt](https://github.com/netbox-community/netbox/files/10884289/error.txt)

### LINKED ISSUE #12023: PATCH requests to cable-terminations do not update the old endpoint
### NetBox version

v3.4.6

### Python version

3.8

### Steps to Reproduce

I came across bug #11901 (although I am working with front ports not interfaces), and had the idea of sending the PATCH/PUT to /dcim/cable-terminations/{id} directly instead of the cable.

The results are an imporovement, but still has to problem that the old endpoint still thinks it is connected to the cable.

Steps to reproduce:

- create three devices
- Create three rear ports and three front ports, one of each for each device
- create a cable connection between two of the devices' front ports
- request an API PATCH HTTP request to change the device and port of b termination to the third device.

### Expected Behavior

The cable is moved from port two to port three, and there are no references to the cable on port two

### Observed Behavior

Looking at the properties of the cable, it shows that it is connected to ports one and three as expected.

However, from the front port view, port two still reports as being connected to the cable as well as ports one and three.

### LINKED ISSUE #13581: Update terminations in cable by API
### NetBox version

V.3.5.8

### Python version

3.10

### Steps to Reproduce

1. Create 2 devices with one or more network interfaces
2. Create cable between devices
example : DeviceA(mgmt0) to DeviceB (ge-1/0/2)

3. Now use PUT/PATCH API (/api/dcim/cables/{id_of_your_cable}/) (example id_of_your_cable = 23 for exemple)
```bach
curl -X 'PUT' \
  'http://X.X.X.X:8100/api/dcim/cables/23/' \
  -H 'accept: application/json' \
  -H 'X-CSRFTOKEN: XXXXXXXXXXXXXXXXXXXXXXXXXXXX' \
  -d '{
  "a_terminations": [{ "object_type": "dcim.interface", "object_id": 187 }],
  "b_terminations": [{ "object_type": "dcim.interface", "object_id": 66 }],
  "status": "connected",
  "type": "cat7"
}
'
```
In this application, the device A interface identifier remains the same (so 187 is for the device A interface called mgmt0), but the device B interface has changed, 66 is one device interface but another (for example ge-1/0/**3**).

You will get this error : 
```json
{
  "error": "list index out of range",
  "exception": "IndexError",
  "netbox_version": "3.5.8",
  "python_version": "3.10.12"
}
```

### Expected Behavior

I want to modify the cable termination(s) 
return the cable with the new interface(s) 

### Observed Behavior

error 500
```json
{
  "error": "list index out of range",
  "exception": "IndexError",
  "netbox_version": "3.5.8",
  "python_version": "3.10.12"
}
```

DanSheps:
### Fixes: #11079 - Add supported path through modules and multiple rear-ports with different positions on same device

* Add support for pathing through modules.
* Adds some logic to generate split paths when path lengths vary


DanSheps:
Not sure how I feel about the logging, could definitely remove it or modify it if someone has strong opinions.

jsenecal (approved):
LGTM

jsenecal:
> Not sure how I feel about the logging, could definitely remove it or modify it if someone has strong opinions.

What is the behavior on an invalid path besides logging ?

DanSheps:
Don't create the path.   It would fail normally anyways, this just allows the cable to be created but not create the cable path.

jeremystretch:
This is not a sufficient means of addressing the underlying issue. The assertions function as sanity checks: If they are being violated, we need to determine why and resolve accordingly. This might entail removing the assertions, or implementing other logic to accommodate the unexpected case. But merely logging them and avoiding creating the cable path is not a solution.

DanSheps:
I am moving to a bit of a rewrite to properly account for at least this type of cable path.  It should be ready for re-review in a few days.

netbox/dcim/models/cables.py
```diff
@@ -501,9 +498,15 @@ def from_origin(cls, terminations):
             # Terminations must all be of the same type
             assert all(isinstance(t, type(terminations[0])) for t in terminations[1:])
 
+            # All mid-span terminations must all be attached to the same device
```
jsenecal, line 501: Is this required by the current implementation ?

DanSheps: I am doing it to maintain the current status quo and not introduce any new features (beyond those requested in the Bug)

jsenecal:
FWIW I tested the current state of `11079-Cablepath_catch_assertions` and it works as advertised and I'm able to trace through multiple front ports which span multiple rear ports :)

Trying to break it some more but so far great work @DanSheps

DanSheps:
I think this is ready @jeremystretch

netbox/dcim/models/cables.py
```diff
@@ -577,20 +603,26 @@ def from_origin(cls, terminations):
...
-
-                if len(remote_terminations) > 1 or remote_terminations[0].positions == 1:
+                if len(remote_terminations) == 1 and remote_terminations[0].positions == 1:
                     front_ports = FrontPort.objects.filter(
                         rear_port_id__in=[rp.pk for rp in remote_terminations],
                         rear_port_position=1
                     )
+                elif len(remote_terminations) > 1 and position_stack:
+                    positions = position_stack.pop()
+                    assert len(remote_terminations) == len(positions)
```
arthanson, line 619: again comment what is going on here would help.

netbox/dcim/models/cables.py
```diff
@@ -632,18 +664,25 @@ def from_origin(cls, terminations):
...
+                if all(isinstance(t, type(remote_terminations[0])) for t in remote_terminations[1:]):
+                    is_complete = True
+                else:
+                    # Unsupported topology, mark as split and exit
+                    is_complete = False
+                    is_split = True
                 break
 
-        return cls(
+        cablepath = cls(
```
arthanson, line 677: Any reason not to just return this without assigning to a var?

arthanson (changes requested):

netbox/dcim/models/cables.py
```diff
@@ -529,46 +535,66 @@ def from_origin(cls, terminations):
...
+            elif len(links) == 0:
                 # Otherwise, halt the trace if no link exists
                 break
-            assert type(link) in (Cable, WirelessLink)
+            assert all(type(link) in (Cable, WirelessLink) for link in links)
+
+            # Create set of links in path.  Cannot use list(set()) as it does this in a non-deterministic manner
+            links_path = []
+            for link in links:
+                if object_to_path_node(link) not in links_path:
```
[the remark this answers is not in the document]
DanSheps: We just want to silently ignore it here because we aren't checking as a "we need to stop this if this condition is hit" but as a "We already have this node in the list of current links, we don't need to add it again".  More relevant when there are multiple rear-port positions in use for a specific rear-port

netbox/dcim/models/cables.py
```diff
@@ -529,46 +535,66 @@ def from_origin(cls, terminations):
...
                 )
-                # Terminations must all belong to same end of Cable
-                local_cable_end = local_cable_terminations[0].cable_end
-                assert all(ct.cable_end == local_cable_end for ct in local_cable_terminations[1:])
-                remote_cable_terminations = CableTermination.objects.filter(
-                    cable=link,
-                    cable_end='A' if local_cable_end == 'B' else 'B'
-                )
+
+                q_filter = Q()
```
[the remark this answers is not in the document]
DanSheps: Sorry, I don't quite follow this.  It isn't checkcing for anything, we are getting our far end terminations as per the comment for the if statement.

DanSheps:
Something I am thinking about, we should run a force retrace paths after this has completed.  Not sure how best to do that.

arthanson (approved):

netbox/dcim/models/cables.py
```diff
@@ -529,46 +535,66 @@ def from_origin(cls, terminations):
...
+            if len(links) == 0:
+                if len(path) == 1:
+                    # If this is the start of the path and no link exists, return None
+                    return None
                 # Otherwise, halt the trace if no link exists
                 break
-            assert type(link) in (Cable, WirelessLink)
+            assert all(type(link) in (Cable, WirelessLink) for link in links)
+
+            # Create set of links in path.  Cannot use list(set()) as it does this in a non-deterministic manner
```
jeremystretch, line 548: > Cannot use list(set()) as it does this in a non-deterministic manner

The implications of this comment aren't immediately apparent. Please extend it to elaborate on why the list of links must be ordered.

Separately, I'm not sure how it could be valid to ignore redundant links in the list if ordering is important. How do we handle a scenario where two pairs of ports are connected via cable A and another two pairs of ports are connected via cable B?

netbox/dcim/models/cables.py
```diff
@@ -529,46 +535,66 @@ def from_origin(cls, terminations):
...
 
-            # Step 3: Record the link and update path status if not "connected"
-            path.append([object_to_path_node(link)])
-            if hasattr(link, 'status') and link.status != LinkStatusChoices.STATUS_CONNECTED:
+            # Step 3: Record the links
+            path.append(links_path)
+
+            # Step 4: Update the path status if a link is not connected
+            links_status = [
+                link.status for link in links if hasattr(link, 'status') and
```
jeremystretch, line 559: All links (Cables and WirelessLinks) must have a `status` attribute.

netbox/dcim/models/cables.py
```diff
@@ -529,46 +535,66 @@ def from_origin(cls, terminations):
...
+                link.status != LinkStatusChoices.STATUS_CONNECTED
+            ]
+            if len(links_status) and len(links) != len(links_status):
```
jeremystretch, line 562: This can be simplified by using `any()` to check for any values other than `STATUS_CONNECTED`.

netbox/dcim/models/cables.py
```diff
@@ -529,46 +535,66 @@ def from_origin(cls, terminations):
...
                 is_active = False
 
-            # Step 4: Determine the far-end terminations
-            if isinstance(link, Cable):
+            # Step 5: Determine the far-end terminations
+            if isinstance(links[0], Cable):
```
jeremystretch, line 566: Is there a scenario wherein the links may be of different types? Probably not, but we should include an assertion to enforce that assumption.

netbox/dcim/models/cables.py
```diff
@@ -529,46 +535,66 @@ def from_origin(cls, terminations):
...
-                local_cable_end = local_cable_terminations[0].cable_end
-                assert all(ct.cable_end == local_cable_end for ct in local_cable_terminations[1:])
-                remote_cable_terminations = CableTermination.objects.filter(
-                    cable=link,
-                    cable_end='A' if local_cable_end == 'B' else 'B'
-                )
+
+                q_filter = Q()
+                for lct in local_cable_terminations:
+                    q_filter |= Q(cable=lct.cable, cable_end='A' if lct.cable_end == 'B' else 'B')
```
jeremystretch, line 575: Suggest splitting this out just for readability.
```suggestion
                    cable_end = 'A' if lct.cable_end == 'B' else 'B'
                    q_filter |= Q(cable=lct.cable, cable_end=cable_end)
```

netbox/dcim/models/cables.py
```diff
@@ -529,46 +535,66 @@ def from_origin(cls, terminations):
...
+
+                assert q_filter is not Q()
```
jeremystretch, line 577: Is this a valid assertion? It's valid to have only one end of a cable connected, so it's possible that there are no remote terminations present.

netbox/dcim/tests/test_cablepaths.py
```diff
@@ -1695,6 +1717,264 @@ def test_218_interfaces_to_interfaces_via_multiposition_rear_ports(self):
...
+            (
+                interface2, cable3, (frontport2, frontport4), (rearport2, rearport4), (cable2, cable4),
+                (rearport1, rearport3), (frontport1, frontport3), cable1, interface1
+            ),
+            is_complete=True,
+            is_active=True
+        )
+        self.assertEqual(CablePath.objects.count(), 2)
+
+    def test_220_interface_to_interface_duplex_via_multiple_rearports_with_modules(self):
```
jeremystretch, line 1792: We can probably eliminate this test. The only difference between it and the one above (219) is the addition of module associations, which I don't believe has any impact on the tracing logic. (This was probably a red herring from #11079.)

netbox/dcim/tests/test_cablepaths.py
```diff
@@ -1695,6 +1717,264 @@ def test_218_interfaces_to_interfaces_via_multiposition_rear_ports(self):
...
+            (
+                interface2, cable3, (frontport2, frontport4), (rearport2, rearport4), (cable2, cable4),
+                (rearport1, rearport3), (frontport1, frontport3), cable1, interface1
+            ),
+            is_complete=True,
+            is_active=True
+        )
+        self.assertEqual(CablePath.objects.count(), 2)
+
+    def test_221_interface_to_interface_duplex_via_multiple_rearports_with_modules(self):
```
jeremystretch, line 1873: This can probably be simplified by removing the module assignments sacrificing any coverage of the actual tracing logic. (Also it should be renamed if we remove test 220 above.)

netbox/dcim/tests/test_cablepaths.py
```diff
@@ -1845,3 +2125,215 @@ def test_303_remove_termination_from_existing_cable(self):
             is_complete=True,
             is_active=True
         )
+
+    def test_401_non_symmetric_paths(self):
```
jeremystretch, line 2129: These can probably be numbered from the 2XX series. But if we want to keep the 4XX numbering, we should add a note to the docstring for `CablePathTestCase`.

jeremystretch (changes requested):

netbox/dcim/models/cables.py
DanSheps: You are absolutely corect here, I originally did it because of the way it was coded to not lose the cable position and diagram it as a simple fan-in/fan-out, however the cable svg logic should instead be modified to account for this.

netbox/dcim/models/cables.py
DanSheps: THinking about it, I don't think it is possible to hit this, as there shoudl always be a local cable termination.

DanSheps:
I have some fine-tuning to do on a new SVG rendering and then I will push that code as well.

netbox/dcim/models/cables.py
jsenecal:
```python
assert all(isinstance(link, type(links[0])) for link in links)
```
Should work for that and be pretty efficient.

DanSheps:
Things to note:

1. If there are 3 or more cables in the render, it will push the other stuff to the tooltip and only show the Cable ID/Name.
2. I have made it so it only puts unique cables in the node list per node.  It does this in order to preserve termination and cable ordering (otherwise you might have cables ordered incorrectly which results in weird lines

netbox/dcim/models/cables.py
```diff
@@ -518,9 +518,15 @@ def from_origin(cls, terminations):
             # Terminations must all be of the same type
             assert all(isinstance(t, type(terminations[0])) for t in terminations[1:])
 
+            # All mid-span terminations must all be attached to the same device
+            if not isinstance(terminations[0], PathEndpoint):
+                assert all(t.device == terminations[0].device for t in terminations[1:])
```
jeremystretch, line 523: Not all terminations are device components (i.e. CircuitTerminations). This should check that all terminations have the same `parent` rather than `device`.

netbox/dcim/svg/cables.py
```diff
@@ -34,9 +34,13 @@ class Node(Hyperlink):
         radius: Box corner radius, for rounded corners (default: 10)
     """
 
-    def __init__(self, position, width, url, color, labels, radius=10, **extra):
+    object = None
+
+    def __init__(self, position, width, url, color, labels, radius=10, object=object, **extra):
         super(Node, self).__init__(href=url, target='_parent', **extra)
 
+        self.object = object
```
jeremystretch, line 42: What's going on here? Why is the object needed?

netbox/dcim/svg/cables.py
```diff
@@ -238,20 +245,39 @@ def draw_fanout(self, node, connector):
...
-    def draw_cable(self, cable):
-        labels = [
-            f'Cable {cable}',
-            cable.get_status_display()
-        ]
-        if cable.type:
-            labels.append(cable.get_type_display())
-        if cable.length and cable.length_unit:
-            labels.append(f'{cable.length} {cable.get_length_unit_display()}')
+    def draw_cable(self, cable, terminations, cable_count=0):
```
jeremystretch, line 248: `draw_cable()` should be limited to rendering a single cable. Where multiple cables need to be displayed, it should be called multiple times.

jeremystretch (changes requested):

netbox/dcim/svg/cables.py
DanSheps: It is referenced by lines 379/387 and 401/402

netbox/dcim/svg/cables.py
DanSheps: It does only draw a single cable, the terminations are passed to get position data on each termination.

jeremystretch (approved):
This has been a huge undertaking. Thanks for seeing it through @DanSheps!

PR #13338, opened 2023-08-01
Adding interface TYPE_400GE_CFP2/400gbase-x-cfp2

### LINKED ISSUE #12889: Add support for CFP2 400G interface type
### NetBox version

v3.5.3

### Feature type

Data model extension

### Proposed functionality

I would suggest a new interface type being created.

I particularly need support for a CFP2 interface which support 400G.

### Use case

You are able to buy a CFP2 400G DCO optic today, which are both used in routers and DWDM systems.

Users can benefit of this interface type when modeling new systems which support CFP2 400G interfaces.

There are several manufacturers of a CFP2 400G DCO optic

### Database changes

_No response_

### External dependencies

_No response_

stranden:
### Fixes: #12889 

Adding support for CFP2 interface which will be able to run 400G.


jeremystretch:
Thanks @stranden!

PR #13342, opened 2023-08-02
Replaces device_role with role on device model

abhi1693:


netbox/dcim/tables/devices.py
```diff
@@ -285,14 +285,14 @@ class DeviceTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
     class Meta(NetBoxTable.Meta):
         model = models.Device
         fields = (
-            'pk', 'id', 'name', 'status', 'tenant', 'tenant_group', 'device_role', 'manufacturer', 'device_type',
+            'pk', 'id', 'name', 'status', 'tenant', 'tenant_group', 'role', 'manufacturer', 'device_type',
```
jeremystretch, line 288: Renaming the column might cause an issue with users' saved table preferences. We may be able to address this in the database migration, updating any occurrence of `device_role` in `tables.DeviceTable.columns`.

netbox/netbox/settings.py
```diff
@@ -469,7 +469,7 @@ def _setting(name, default=None):
 
 DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
 
-TEST_RUNNER = "django_rich.test.RichRunner"
+# TEST_RUNNER = "django_rich.test.RichRunner"
```
jeremystretch, line 472: Guessing this was intended to be a temporary change? (Happy to make the test runner configurable.)

jeremystretch (changes requested):
Thanks for plowing through all of this!

netbox/netbox/settings.py
abhi1693: Yeah, I do not like the colored output. Makes it difficult to find the issue. I'll revert this.

netbox/dcim/tables/devices.py
jeremystretch: I've added a routine in the migration file to update references to the old `device_role` table column in user configs.

PR #13353, opened 2023-08-02
Feature - Schema Generation

### LINKED ISSUE #10030: Automatically build JSON schema for the device type library with each release
### NetBox version

v3.2.9

### Feature type

New functionality

### Proposed functionality

Implement a mechanism for generating JSON schema for device type and related definitions. This could be a Django management command (e.g. `manage.py buildschema`), the output of which would be included with each NetBox release.

### Use case

The NetBox [devicetype library](https://github.com/netbox-community/devicetype-library) includes [JSON schema](https://github.com/netbox-community/devicetype-library/tree/master/schema) for validating device and module type definitions. These definitions must be kept up-to-date with NetBox itself, which is currently a manual process. Generating the schema automatically and publishing it with each NetBox release will remove this administrative burden.

Additionally, once we implement #4347, schema can be generated for all relevant models as an additional resource.

### Database changes

_No response_

### External dependencies

_No response_

danner26:
Fixes: #10030

Added `manage.py` subcommand (`buildschema`) to build a valid schema file for the devicetype-library. Data within this file will be used directly within the validation schema to ensure choices are synced properly between both repositories. Future revisions of this will be utilized with GitHub Actions to automatically update the devicetype-library (or alternatively we will write the schema to this repository and pull it on the fly).


contrib/generated_schema.json
```diff
@@ -0,0 +1,561 @@
...
+                        "urm-p4",
+                        "urm-p8",
+                        "splice",
+                        "other"
+                    ]
+                }
+            }
+        }
+    }
+}
```
abhi1693, line 561: Should add new line

netbox/extras/management/commands/buildschema.py
```diff
@@ -0,0 +1,60 @@
+from os import path as os_path
+from json import dumps as json_dumps
+from json import loads as json_loads
+from jinja2 import FileSystemLoader, Environment
+
+from django.core.management.base import BaseCommand
+
+from dcim.choices import DeviceAirflowChoices
```
abhi1693, line 8: I think it would be better to clean up the import like `from dcim.choices import *`

netbox/extras/management/commands/buildschema.py
```diff
@@ -0,0 +1,60 @@
...
+        schemas["console_port_type_choices"] = json_dumps(ConsolePortTypeChoices.values())  # console-ports and console-server-ports
+        schemas["power_port_type_choices"] = json_dumps(PowerPortTypeChoices.values())
+        schemas["power_outlet_type_choices"] = json_dumps(PowerOutletTypeChoices.values())
+        schemas["power_outlet_feedleg_choices"] = json_dumps(PowerOutletFeedLegChoices.values())
+        schemas["interface_type_choices"] = json_dumps(InterfaceTypeChoices.values())
+        schemas["interface_poe_mode_choices"] = json_dumps(InterfacePoEModeChoices.values())
+        schemas["interface_poe_type_choices"] = json_dumps(InterfacePoETypeChoices.values())
+        schemas["port_type_choices"] = json_dumps(PortTypeChoices.values())  # front-ports and rear-ports
+
+        template_loader = FileSystemLoader(searchpath=f'{os_path.dirname(__file__)}/../templates')
```
abhi1693, line 47: You should rather use `TEMPLATES_DIR`. In future if this changes, it will not affect the code here

netbox/extras/management/commands/buildschema.py
```diff
@@ -0,0 +1,60 @@
...
+        TEMPLATE_FILE = 'generated_schema.json.j2'
+        template = template_env.get_template(TEMPLATE_FILE)
+        outputText = template.render(schemas=schemas)
+
+        if kwargs['console']:
+            print(json_dumps(json_loads(outputText), indent=4))
+
+        if kwargs['file']:
+            print()
+            with open(f'{os_path.dirname(__file__)}/../../../../contrib/generated_schema.json', 'w') as generated_json_file:
```
abhi1693, line 58: You should use `BASE_DIR` and then do the path traversal to contrib directory

netbox/extras/management/templates/generated_schema.json.j2
```diff
@@ -0,0 +1,93 @@
+{
```
abhi1693, line 1: If `TEMPLATES_DIR` is used, then the file should be moved to the common `templates` directory

netbox/extras/management/templates/generated_schema.json.j2
```diff
@@ -0,0 +1,93 @@
...
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "enum": {{ schemas["port_type_choices"]}}
+        }
+      }
+    }
+  }
+}
```
abhi1693, line 93: New line

netbox/extras/management/commands/buildschema.py
danner26: Absolutely agree, this will be implemented as well

netbox/extras/management/commands/buildschema.py
danner26: I personally disagree with this. To avoid excess data load, aka imports, we should only import what is needed to improve performance. Yes it might be negligable at the end of the day, but any saved CPU cycles are better than nothing

netbox/extras/management/templates/generated_schema.json.j2
danner26: Absolutely agree, this is being implemented

netbox/extras/management/templates/generated_schema.json.j2
danner26: Absolutely agree, this is being implemented

contrib/generated_schema.json
danner26: Absolutely agree, this is being implemented

danner26: This is implemented now.

netbox/extras/management/commands/buildschema.py
danner26: This is implemented now.

netbox/extras/management/commands/buildschema.py
danner26: This is implemented now.

netbox/extras/management/templates/generated_schema.json.j2
danner26: This is implemented now.

netbox/extras/management/templates/generated_schema.json.j2
danner26: This is implemented now.

netbox/extras/management/commands/buildschema.py
jeremystretch: Avoiding the wildcard doesn't really save anything. The entire module must be loaded into memory either way; the explicit imports just limit what gets populated in the namespace.

danner26: Okay, so would you prefer that we change this to `*`?

netbox/extras/management/commands/buildschema.py
```diff
@@ -0,0 +1,57 @@
...
+        schemas["weight_unit_choices"] = json_dumps(WeightUnitChoices.values())
+        schemas["subdevice_role_choices"] = json_dumps(SubdeviceRoleChoices.values())
+        schemas["console_port_type_choices"] = json_dumps(ConsolePortTypeChoices.values())  # console-ports and console-server-ports
+        schemas["power_port_type_choices"] = json_dumps(PowerPortTypeChoices.values())
+        schemas["power_outlet_type_choices"] = json_dumps(PowerOutletTypeChoices.values())
+        schemas["power_outlet_feedleg_choices"] = json_dumps(PowerOutletFeedLegChoices.values())
+        schemas["interface_type_choices"] = json_dumps(InterfaceTypeChoices.values())
+        schemas["interface_poe_mode_choices"] = json_dumps(InterfacePoEModeChoices.values())
+        schemas["interface_poe_type_choices"] = json_dumps(InterfacePoETypeChoices.values())
+        schemas["port_type_choices"] = json_dumps(PortTypeChoices.values())  # front-ports and rear-ports
```
jeremystretch, line 42: We can simplify this to a mapping of schema section name (e.g. `airflow_choices`) to ChoiceSet class (e.g. `DeviceAirFlowChoices`). We can then use a `for` loop to call `json_dumps()` and generate the template context data. This allows us to separate the rote mapping from the generation logic, which will make it easier to adjust in the future if needed.

netbox/extras/management/commands/buildschema.py
```diff
@@ -0,0 +1,57 @@
...
+            action='store_true',
+            help="Print the generated schema to the generated file"
+        )
+
+    def handle(self, *args, **kwargs):
+        schemas = {}
+        schemas["airflow_choices"] = json_dumps(DeviceAirflowChoices.values())
+        schemas["weight_unit_choices"] = json_dumps(WeightUnitChoices.values())
+        schemas["subdevice_role_choices"] = json_dumps(SubdeviceRoleChoices.values())
+        schemas["console_port_type_choices"] = json_dumps(ConsolePortTypeChoices.values())  # console-ports and console-server-ports
```
jeremystretch, line 35: I would create separate keys for `console_port_type_choices` and `console_server_port_type_choices`, pointing to the same ChoiceSet. This makes our intention more explicit and helps prevent confusion.

netbox/extras/management/commands/buildschema.py
```diff
@@ -0,0 +1,57 @@
...
+        schemas["power_outlet_type_choices"] = json_dumps(PowerOutletTypeChoices.values())
+        schemas["power_outlet_feedleg_choices"] = json_dumps(PowerOutletFeedLegChoices.values())
+        schemas["interface_type_choices"] = json_dumps(InterfaceTypeChoices.values())
+        schemas["interface_poe_mode_choices"] = json_dumps(InterfacePoEModeChoices.values())
+        schemas["interface_poe_type_choices"] = json_dumps(InterfacePoETypeChoices.values())
+        schemas["port_type_choices"] = json_dumps(PortTypeChoices.values())  # front-ports and rear-ports
+
+        template_loader = FileSystemLoader(searchpath=f'{settings.TEMPLATES_DIR}/extras/')
+        template_env = Environment(loader=template_loader)
+        TEMPLATE_FILE = 'generated_schema.json'
```
jeremystretch, line 46: A name like `TEMPLATE_FILENAME` might be preferable to avoid confusion with actual file objects. I'd also suggest moving it outside of the `Command` class for better visibility.

netbox/extras/management/commands/buildschema.py
```diff
@@ -0,0 +1,57 @@
...
+
+class Command(BaseCommand):
+    help = "Generate the NetBox validation schemas."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--console',
+            action='store_true',
+            help="Print the generated schema to stdout"
+        )
```
jeremystretch, line 23: It's not necessary to specify this explicitly: We can output the generated content to STDOUT by default, unless an output filename is specified.

netbox/extras/management/commands/buildschema.py
```diff
@@ -0,0 +1,57 @@
...
+        parser.add_argument(
+            '--file',
```
jeremystretch, line 25: A minor nit, but `--output` might be preferable to keep consistent with some native Django management commands (`dumpdata`, at least?).

jeremystretch (changes requested):

netbox/templates/extras/generated_schema.json
```diff
@@ -0,0 +1,93 @@
+{
+  "type": "object",
+  "additionalProperties": false,
+  "definitions": {
+    "airflow": {
+      "type": "string",
+      "enum": {{ schemas["airflow_choices"] }}
```
abhi1693, line 7: If it works, would a dot notation be more clearer approah? Example: `{{ schemas.ariflow_choices }}`

PR #13373, opened 2023-08-04
11508 Add group assignments for Azure SSO

### LINKED ISSUE #11508: Using AzureAD authentication I cannot get the group members for the user that authenticates
### NetBox version

v3.3.8

### Python version

3.9

### Steps to Reproduce

1. I authenticate using AzureAD with my user
2. I get authenticated
3. I check the user groups in admin with a superuser account
4. I don't see any group even if the user has several groups in AzureAD
5. I check in the user id_token in User social auths (inside the meny PYTHON SOCIAL AUTH) and I verify that I get the group id

### Expected Behavior

The expected behaviour would be to see the member groups related to the user that is authenticated using AzureAD

### Observed Behavior

No groups appeared related to the user.

My configurations related with authentication is:

REMOTE_AUTH_BACKEND = 'social_core.backends.azuread.AzureADOAuth2'

SOCIAL_AUTH_AZUREAD_OAUTH2_KEY = 'xxxxxxxxxxxxxxxxxxxxxxxxxxxx'
SOCIAL_AUTH_AZUREAD_OAUTH2_SECRET = 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
SOCIAL_AUTH_REDIRECT_IS_HTTPS = True

arthanson:
### Fixes: #11508 

Creates a custom Social Auth pipeline to allow mapping Azure group-id's to NetBox groups so groups, staff and superuser assignments are automatically done when using Azure SSO.


netbox/netbox/authentication.py
```diff
@@ -386,3 +387,74 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+class AuthFailed(Exception):
+    pass
+
+
+def azuread_map_groups(response, user, backend, *args, **kwargs):
+    '''
+    Map Azure AD group ID to Netbox group
+    Also set is_superuser or is_staff based on config map
+    '''
+    if not getattr(settings, "SOCIAL_AUTH_AZUREAD_MAP_GROUP_PERMS", False):
```
jeremystretch, line 401: We probably don't need this setting. Adding/removing this function to/from the authentication pipeline should be enough to enable or disable the functionality.

netbox/netbox/authentication.py
```diff
@@ -386,3 +387,74 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+        return
+
+    flags_by_group = getattr(settings, "SOCIAL_AUTH_AZUREAD_USER_FLAGS_BY_GROUP", False)
+    if not flags_by_group:
```
jeremystretch, line 405:
```suggestion
    if not hasattr(settings, "SOCIAL_AUTH_AZUREAD_USER_FLAGS_BY_GROUP"):
```

netbox/netbox/authentication.py
```diff
@@ -386,3 +387,74 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+        raise ImproperlyConfigured(
+            "Azure group mapping has been configured, but SOCIAL_AUTH_AZUREAD_USER_FLAGS_BY_GROUP is not defined."
+        )
+
+    group_mapping = getattr(settings, "SOCIAL_AUTH_AZUREAD_GROUP_MAP", False)
+    if not group_mapping:
```
jeremystretch, line 411:
```suggestion
    if not hasattr(settings, "SOCIAL_AUTH_AZUREAD_GROUP_MAP"):
```

netbox/netbox/authentication.py
```diff
@@ -386,3 +387,74 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+    response = requests.get(
+        url,
+        headers=headers,
+    )
+    uid = response.json().get('id')
+    url = f"https://graph.microsoft.com/v1.0/users/{uid}/memberOf"
+    response = requests.get(
+        url,
+        headers=headers,
+    )
```
jeremystretch, line 434: There doesn't seem to be any error handling in place. At a minimum, we should catch any fatal errors and re-raise them with some context for the user. Some comments would also be helpful to explain what requests are being made and why.

Additionally, we should avoid declaring `url` multiple times: If the value isn't being reused, we can pass the string(s) directly to `requests.get()` instead.

netbox/netbox/authentication.py
```diff
@@ -386,3 +387,74 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
             user.groups.add(*group_list)
         else:
             logger.info(f"No valid group assignments for {user} - REMOTE_AUTH_DEFAULT_GROUPS may be incorrectly set?")
+
+
+class AuthFailed(Exception):
+    pass
+
+
+def azuread_map_groups(response, user, backend, *args, **kwargs):
```
jeremystretch, line 396: We should enable logging for everything that happens inside this function.

netbox/netbox/authentication.py
```diff
@@ -386,3 +387,74 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+
+            if group_id in flags_by_group['is_staff']:
+                is_staff = True
+
+            if group_id in group_mapping:
+                group = Group.objects.get(name=group_mapping[group_id])
+                if group:
+                    group.user_set.add(user)
+                else:
+                    logger.info(f"Azure group mapping - group: {group_mapping[group_id]} not found.")
```
jeremystretch, line 456: `logger` has not been defined.

docs/administration/authentication/microsoft-azure-ad.md
```diff
@@ -61,6 +61,43 @@ Restart the NetBox services so that the new configuration takes effect. This is
...
+    'social_core.pipeline.social_auth.associate_user',
+    'netbox.authentication.user_default_groups_handler',
+    'social_core.pipeline.social_auth.load_extra_data',
+    'social_core.pipeline.user.user_details',
+    'netbox.authentication.azuread_map_groups',
+)
+SOCIAL_AUTH_AZUREAD_MAP_GROUP_PERMS = True
+
+# Define special user types using groups. Exercise great caution when assigning superuser status.
+SOCIAL_AUTH_AZUREAD_USER_FLAGS_BY_GROUP = {
```
jeremystretch, line 86: I recall we talked about the parameter naming strategy last week, but I'm having second thoughts. I worry that introducing our own paramters prefixed with `SOCIAL_AUTH_AZUREAD` may cause confusion among users expecting to see it the `python-social-auth` docs. I don't have a better suggestion though; open to other opinions.

jeremystretch (changes requested):

netbox/netbox/authentication.py
arthanson: For error checking I think the error from the API call would contain the most info so I've added extra logging but re-raise the error.

docs/administration/authentication/microsoft-azure-ad.md
abhi1693: Why not prefix with `NETBOX_`?

arthanson: changed to REMOTE_AUTH_BACKEND_ and auto-imported everything starting with this as we will probably be adding more so can group them all together.

docs/administration/authentication/microsoft-azure-ad.md
```diff
@@ -61,6 +61,42 @@ Restart the NetBox services so that the new configuration takes effect. This is
...
+
+REMOTE_AUTH_BACKEND_AZUREAD_GROUP_MAP = {
+    '{AZURE_GROUP_ID}': '{NETBOX_GROUP}',
+}
+```
+**REMOTE_AUTH_BACKEND_AZUREAD_USER_FLAGS_BY_GROUP.is_staff**: users who are in any of the Azure AD group-ids in the array will have staff permission assigned to them.
+
+**REMOTE_AUTH_BACKEND_AZUREAD_USER_FLAGS_BY_GROUP.is_superuser**: users who are in any of the Azure AD group-ids in the array will have superuser permission assigned to them.
+
+**REMOTE_AUTH_BACKEND_AZUREAD_GROUP_MAP**: Any user with the given Azure AD group-id is included in the given NetBox group name.
```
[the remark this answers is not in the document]
arthanson: changed to the following:

```
SOCIAL_AUTH_PIPELINE_CONFIG = {
    # Define special user types using groups. Exercise great caution when assigning superuser status.
    'AZUREAD_USER_FLAGS_BY_GROUP': {
        'is_staff': ['1a36bed9-3bdc-4970-ab66-faf9704e0af4',],
        'is_superuser': ['1a36bed9-3bdc-4970-ab66-faf9704e0af4',]
    },

    'AZUREAD_GROUP_MAP': {
        '1a36bed9-3bdc-4970-ab66-faf9704e0af4': 'tgroup',
    }
}
```

netbox/netbox/settings.py
```diff
@@ -162,6 +162,11 @@
 TIME_ZONE = getattr(configuration, 'TIME_ZONE', 'UTC')
 ENABLE_LOCALIZATION = getattr(configuration, 'ENABLE_LOCALIZATION', False)
 
+# Load all REMOTE_AUTH_BACKEND_* settings from the user configuration
+for param in dir(configuration):
+    if param.startswith('REMOTE_AUTH_BACKEND_'):
+        globals()[param] = getattr(configuration, param)
+
```
jeremystretch, line 169: This is no longer needed and can be removed.

docs/administration/authentication/microsoft-azure-ad.md
```diff
@@ -61,6 +61,45 @@ Restart the NetBox services so that the new configuration takes effect. This is
...
+    'social_core.pipeline.social_auth.load_extra_data',
+    'social_core.pipeline.user.user_details',
+    'netbox.authentication.azuread_map_groups',
+)
+
+# Define special user types using groups. Exercise great caution when assigning superuser status.
+SOCIAL_AUTH_PIPELINE_CONFIG = {
+    'AZUREAD_USER_FLAGS_BY_GROUP': {
+        "is_staff": ['{AZURE_GROUP_ID}',],
+        "is_superuser": ['{AZURE_GROUP_ID}',]
```
jeremystretch, line 88: IMO we should remove the braces from the example just to avoid confusion with the containing Python syntax. It might also be helpful to show an example listing multiple groups.

```suggestion
        "is_staff": ['StaffGroup1', 'StaffGroup2'],
        "is_superuser": ['SuperuserGroup'],
```

docs/administration/authentication/microsoft-azure-ad.md
```diff
@@ -61,6 +61,45 @@ Restart the NetBox services so that the new configuration takes effect. This is
...
+    },
+
+    'AZUREAD_GROUP_MAP': {
+        '{AZURE_GROUP_ID}': '{NETBOX_GROUP}',
+    }
+
```
jeremystretch, line 94: Showing a list of mappings helps convey the pattern to the user.

```suggestion
    'AZUREAD_GROUP_MAP': {
        'AzureADGroup1': 'NetBoxGroup1',
        'AzureADGroup2': 'NetBoxGroup2',
        'AzureADGroup3': 'NetBoxGroup3',
    },
```

netbox/netbox/authentication.py
```diff
@@ -47,6 +48,8 @@
     'salesforce-oauth2': ('Salesforce', 'salesforce'),
 }
 
+BASE_MICROSOFT_GRAPH_URL = 'https://graph.microsoft.com/v1.0/'
```
jeremystretch, line 51: We should probably move this inside `azuread_map_groups()` as its only used there.

netbox/netbox/authentication.py
```diff
@@ -386,3 +389,90 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+def azuread_map_groups(response, user, backend, *args, **kwargs):
+    '''
+    Map Azure AD group ID to Netbox group
+    Also set is_superuser or is_staff based on config map
+    '''
+    logger = logging.getLogger('netbox.auth.azuread_map_groups')
+
+    if not hasattr(settings, "SOCIAL_AUTH_PIPELINE_CONFIG"):
+        raise ImproperlyConfigured(
+            "Azure group mapping has been configured, but SOCIAL_AUTH_PIPELINE_CONFIG is not defined."
```
jeremystretch, line 403: We should be consistent about always referring to "Azure AD" authentication rather than just "Azure" here and below, as the later could very well imply some other authentication mechanism in the future (or even the present, for all I know).

```suggestion
            "Azure AD group mapping has been configured, but SOCIAL_AUTH_PIPELINE_CONFIG is not defined."
```

netbox/netbox/authentication.py
```diff
@@ -386,3 +389,90 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+    config = getattr(settings, "SOCIAL_AUTH_PIPELINE_CONFIG")
+    if "AZUREAD_USER_FLAGS_BY_GROUP" not in config:
+        raise ImproperlyConfigured(
+            "Azure group mapping has been configured, but AZUREAD_USER_FLAGS_BY_GROUP is not defined."
+        )
+
+    if "AZUREAD_GROUP_MAP" not in config:
+        raise ImproperlyConfigured(
+            "Azure group mapping has been configured, but AZUREAD_GROUP_MAP is not defined."
+        )
```
jeremystretch, line 415: Should we require both of these to be defined? I can easily see a case where the user might not care to set `AZUREAD_USER_FLAGS_BY_GROUP`. We should probably check that at least one or the other is defined, though.

netbox/netbox/authentication.py
```diff
@@ -386,3 +389,90 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+            headers=headers,
+        )
+    except Exception as e:
+        logger.error(f"Azure group mapping error getting groups for user {user} from Microsoft Graph API: {e}")
+        raise e
+
+    # Set groups and permissions based on returned group list
+    is_superuser = False
+    is_staff = False
+    values = response.json().get('value', [])
```
jeremystretch, line 447: Probably want to add some error-handling around this in case the response data is in an unexpected format.

netbox/netbox/authentication.py
```diff
@@ -386,3 +389,90 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+
+    user.groups.through.objects.filter(user=user).delete()
+    for value in values:
+        # AD response contains both directories and groups - we only want groups
```
jeremystretch, line 451: Is this something that should be conveyed in the user-facing documentation? Is this a distinction they'll need to be aware of on the AD side? (We're pushing the limits of my AD knowledge.)

netbox/netbox/authentication.py
```diff
@@ -386,3 +389,90 @@ def user_default_groups_handler(backend, user, response, *args, **kwargs):
...
+                try:
+                    group = Group.objects.get(name=group_name)
+                except Group.DoesNotExist:
+                    group = None
+
+                if group:
+                    group.user_set.add(user)
+                    logger.info(f"Azure group mapping - adding group {group_name} to user: {user}.")
+                else:
+                    logger.info(f"Azure group mapping - group: {group_name} not found.")
```
jeremystretch, line 474: Can we combine this block with the one above it to condense the code a bit?

jeremystretch (changes requested):
Thanks for keeping at this! Looks good. I just have some cleanup notes & general questions.

netbox/netbox/authentication.py
arthanson: I don't think so, it is just how it is sent internally, I don't think it has any effect on the user or anything they need to be aware of.

jeremystretch (approved):

jeremystretch:
Thanks @arthanson!

PR #13382, opened 2023-08-04
Adds standardized list API for scripts and reports

abhi1693:

### DIFF
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -243,7 +243,7 @@ class ReportViewSet(ViewSet):
             'request': request,
         })
 
-        return Response(serializer.data)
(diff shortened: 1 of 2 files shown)


netbox/extras/api/views.py
```diff
@@ -243,7 +243,7 @@ def list(self, request):
             'request': request,
         })
 
-        return Response(serializer.data)
+        return Response({'count': len(report_list), 'results': serializer.data})
```
jeremystretch, line 246: Should `display` also include the module? If not, we could see duplicate report/script names.

abhi1693: Check now if you find the format correct or should we use a different one?

Issue #12814, opened 2023-06-05
Expose more objects in the configuration rendering feature


MajesticFalcon:
### NetBox version

v3.5.3

### Feature type

Change to existing functionality

### Proposed functionality

Configuration rendering should expose the route-targets so more advanced configurations can be generated.

### Use case

Unless I am missing something, currently, the implementation of configuration rendering seems to only provide access to the "device" object. Technically, you can loop through the device.interfaces.all() and create a new array of L2vpn terminations, however, I don't believe you can get the route-targets using this method.

### Database changes

None.

### External dependencies

None.

jeremystretch:
Expanded more generally, I think the underlying proposal here is to make NetBox models available directly as context variables when rendering a template. This would enable the template author to do e.g.

```
{% for rt in RouteTarget.objects.all() %}
```

This was a consideration raised during the initial work on the config rendering feature and ultimately punted simply because we weren't sure if it made sense at the time. I don't think there's anything necessarily precluding us from adding this, but would like to leave this open for discussion for a while.

We may also want to consider namespacing the models in some fashion, so that generic models like `Device` and `Site` don't conflict with user-provided context variables. (It may also be necessary to avoid naming collisions among core and plugin models.)

MajesticFalcon:
Sounds good. In the meantime, is there a different method I could to get the RT info usable inside the configuration rendering? The configuration snippet you provided would be great, but I suspect/hope there is a round-a-bout way to get the same information through the device object?

MajesticFalcon:
> Expanded more generally, I think the underlying proposal here is to make NetBox models available directly as context variables when rendering a template. This would enable the template author to do e.g.
> 
> ```
> {% for rt in RouteTarget.objects.all() %}
> ```
> 
> This was a consideration raised during the initial work on the config rendering feature and ultimately punted simply because we weren't sure if it made sense at the time. I don't think there's anything necessarily precluding us from adding this, but would like to leave this open for discussion for a while.
> 
> We may also want to consider namespacing the models in some fashion, so that generic models like `Device` and `Site` don't conflict with user-provided context variables. (It may also be necessary to avoid naming collisions among core and plugin models.)

Sounds good. In the meantime, is there a different method I could to get the RT info usable inside the configuration rendering? The configuration snippet you provided would be great, but I suspect/hope there is a round-a-bout way to get the same information through the device object?

DanSheps:
I am assuming this is for L2VPNs?

Two ways:

[...]

PR #13389, opened 2023-08-04
Exposes all models in device context data

abhi1693:
### Fixes: #12814

To avoid confusion between `device` and `Device`, I have renamed `device` to `object`. The models are exposed using the class model name and are limited to models that have `custom_field` features.

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -1,3 +1,4 @@
+from django.apps import apps
 from django.http import Http404, HttpResponse
 from django.shortcuts import get_object_or_404
 from drf_spectacular.types import OpenApiTypes
@@ -23,6 +24,7 @@ from netbox.api.renderers import TextRenderer
 from netbox.api.viewsets import NetBoxModelViewSet
 from netbox.api.viewsets.mixins import SequentialBulkCreatesMixin
 from netbox.constants import NESTED_SERIALIZER_PREFIX
+from netbox.registry import registry
 from utilities.api import get_serializer_for_model
 from utilities.utils import count_related
 from virtualization.models import VirtualMachine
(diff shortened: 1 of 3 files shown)


netbox/dcim/api/views.py
```diff
@@ -432,7 +434,14 @@ def render_config(self, request, pk):
         # Compile context data
         context_data = device.get_config_context()
         context_data.update(request.data)
-        context_data.update({'device': device})
+        context_data.update({'object': device})
```
jeremystretch, line 437: I realize `object` is more standard, but IMO we should keep this as `device` now that the name collision is no longer a concern.

netbox/dcim/api/views.py
```diff
@@ -432,7 +434,14 @@ def render_config(self, request, pk):
...
+
+        app_ns = registry['model_features']['custom_fields'].keys()
```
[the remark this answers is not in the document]
jeremystretch: I've just opened #13427 to track this effort.

netbox/dcim/api/views.py
```diff
@@ -434,6 +436,13 @@ def render_config(self, request, pk):
         context_data.update(request.data)
         context_data.update({'device': device})
 
+        app_ns = registry['model_features']['custom_fields'].keys()
+        for app in app_ns:
+            context_data.setdefault(app, {})
+            models = apps.get_app_config(app).get_models()
+            for model in models:
+                context_data[app][model.__name__] = model
```
jeremystretch, line 444: @abhi1693 what do you think about moving this logic to `ConfigTemplate.render()`? This would ensure the models are always present in the base template context, and avoid the duplication.

abhi1693: I was thinking to refactor this under device but that's a better place to put this in to have more consistent flow of code.

abhi1693: After refactoring, should the models be displayed in the context data?

abhi1693:
This is ready for another review

PR #13402, opened 2023-08-07
#13388: Misc suggested optimizations

jeremystretch:
A set of proposed changes to #13388. (Apologies for jumping the gun on the draft PR.)

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -367,9 +367,7 @@ class AggregatePrefixesView(generic.ObjectChildrenView):
         return add_requested_prefixes(parent.prefix, queryset, show_available, show_assigned)
 
     def get_extra_context(self, request, instance):
-        context = super().get_extra_context(request, instance)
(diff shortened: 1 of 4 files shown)


netbox/ipam/views.py
```diff
@@ -216,6 +216,7 @@ class ASNRangeASNsView(generic.ObjectChildrenView):
     child_model = ASN
     table = tables.ASNTable
     filterset = filtersets.ASNFilterSet
+    template_name = 'generic/object_tab.html'
```
abhi1693, line 219: Is there a need to define the template again? The same is already defined on `ObjectChildrenView`

jeremystretch: I've removed it from the base class per my comment over at https://github.com/netbox-community/netbox/pull/13388#discussion_r1285898157. I'm probably being overly cautious but I want to be very careful to not introduce a breaking change for any plugins using the view.

PR #13411, opened 2023-08-08
13256 ignore REMOTE_AUTH_BACKEND if REMOTE_AUTH_ENABLED is not set

### LINKED ISSUE #13256: LDAP configuration errors even when REMOTE_AUTH_ENABLED = False
### NetBox version

v3.5.6

### Python version

3.8

### Steps to Reproduce

Set the following in configuration.py

```
REMOTE_AUTH_ENABLED = False
REMOTE_AUTH_BACKEND = 'netbox.authentication.LDAPBackend'
```

and leave ldap_config.py missing, or create an empty file.

### Expected Behavior

Since REMOTE_AUTH_ENABLED is False, no attempt should be made to load ldap config

### Observed Behavior

ldap config errors are shown.  (Reported in discussion #13138)

I suspect the cause is [here](https://github.com/netbox-community/netbox/blob/v3.5.6/netbox/netbox/api/authentication.py#L63):

```
        if 'netbox.authentication.LDAPBackend' in settings.REMOTE_AUTH_BACKEND:
```

which probably should start `if settings.REMOTE_AUTH_ENABLED and ...`

arthanson:
### Fixes: #13256 

Ignores REMOTE_AUTH_BACKEND if REMOTE_AUTH_ENABLED is set to False.  @jeremystretch is easy / short fix, but wasn't sure if we just wanted to say by-design for this one...


netbox/netbox/settings.py
```diff
@@ -418,6 +418,9 @@ def _setting(name, default=None):
 ]
 
 # Set up authentication backends
+if not REMOTE_AUTH_ENABLED:
+    # ignore setting if REMOTE_AUTH_ENABLED not set
+    REMOTE_AUTH_BACKEND = 'netbox.authentication.RemoteUserBackend'
```
jeremystretch, line 423: Let's clean up this logic a bit to avoid referencing RemoteUserBackend again. We should only be extending `AUTHENTICATION_BACKENDS` if `REMOTE_AUTH_ENABLED` is true.

jeremystretch:
Upon closer review, I don't believe this change is warranted. Please see [my comment](https://github.com/netbox-community/netbox/issues/13256#issuecomment-1674849109) on #13256.

PR #13419, opened 2023-08-09
Fixes #13414 - Don't hide HIDDEN_IFUNSET custom fields from bulk import fields

### LINKED ISSUE #13414: Custom Fields omitted from Import "Field Options" Table if UI Visibility set to 'Hidden (if unset)'
### NetBox version

v3.5.7

### Python version

3.9

### Steps to Reproduce

1. Take any Custom Field object and change the UI Visibility to 'Hidden (if unset)
2. Go to a mass import page, eg https://{netbox-host}/circuits/circuits/import/
3. Observe the lack of presence for said Custom Field object in the Field Options table

### Expected Behavior

Expect the Custom Fields to be present in the Field Options table to aid in crafting CSV's with required formatting.

### Observed Behavior

CF options are omitted from the Field Objects table, and so you generally need to gather requisite information in a roundabout manner.
Export creates fields as Label Value, so you generally still need to parse CF's for the right Import formatting; can be a nuisance.

We like the newer capability of hiding unused fields from UI, but we have less skilled users using Netbox and so hope an exception can be made here for restoration of CF visibility in Import scenarios.

kkthxbye-code:
### Fixes: #13414

Bulk import should show Hidden (If unset) custom fields in the field list to match behavior of edit form.

### DIFF
--- a/netbox/netbox/forms/base.py
+++ b/netbox/netbox/forms/base.py
@@ -78,7 +78,10 @@ class NetBoxModelImportForm(CSVModelForm, NetBoxModelForm):
 
     def _get_custom_fields(self, content_type):
         return CustomField.objects.filter(content_types=content_type).filter(
-            ui_visibility=CustomFieldVisibilityChoices.VISIBILITY_READ_WRITE
+            ui_visibility__in=[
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @kkthxbye-code!

PR #13452, opened 2023-08-12
Add 200Gbps & 400Gbps interface speed options

### LINKED ISSUE #13442: Interface Speed: 400Gbps
### NetBox version

v3.5.7

### Feature type

Data model extension

### Proposed functionality

Add 400G to the Speed 

Screenshot 2023-08-10 at 17 24 19

### Use case

We have 400G interfaces and would like to set the speed in NetBox. 

### Database changes

_No response_

### External dependencies

_No response_

tagur87:
### Fixes: #13442

This adds a 200 Gb/s and 400Gb/s speed option for interfaces.

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -1141,6 +1141,8 @@ class InterfaceSpeedChoices(ChoiceSet):
         (25000000, '25 Gbps'),
         (40000000, '40 Gbps'),
         (100000000, '100 Gbps'),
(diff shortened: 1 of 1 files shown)


tagur87:
@jeremystretch  - here's the proposed change. 

While looking at this, I also wondered if I should add any additional Infiniband or Ethernet interface type that would support 200Gbs or 400Gbs??

PR #13474, opened 2023-08-15
upgrading.md: there shouldbe OLDVER instead of NEWVER

jose-d:
documentation fix - OLDVER and NEWVER seems to be mismatched.


jeremystretch:
Thanks @jose-d!

PR #13489, opened 2023-08-16
Remove card management buttons when the dashboard is locked

### LINKED ISSUE #13077: Hide grid stack buttons when dashboard is locked
### NetBox version

v3.5.4

### Python version

3.8

### Steps to Reproduce

1- Navigate to the dashboard page
2- Scroll down to confirm the dashboard is locked with the presence of the following button:
image
3- Click on the "cog" button at the top left corner of any dashboard cards.
4- Make any change and save.
5- Notice that your change was applied even though the dashboard is locked.

### Expected Behavior

The `cog`-style and `close` buttons should be hidden when the dashboard is locked in order to prevent changes to the dashboard in modal forms or from the dashboard itself.

### Observed Behavior

Changes to dashboard cards are not prevented when the dashboard is locked.

jsenecal:
### Fixes: #13077

The cog-style and close buttons should be hidden on the various cards when the dashboard is locked in order to prevent changes to the dashboard in modal forms or from the dashboard itself.

This PR addresses this and hides the buttons when the dashboard is locked.


netbox/project-static/src/dashboard.ts
```diff
@@ -4,15 +4,36 @@ import { apiPatch, hasError } from './util';
...
+    // Show the grid stack items buttons
+    Array.from(document.getElementsByClassName("grid-stack-item-button")).forEach(function (element) {
+      element.classList.remove('invisible');
+    });
+  }
+  if (gridUnlockButton && gridLockButton) {
+    gridUnlockButton.classList.add('invisible');
+    gridLockButton.classList.remove('invisible');
   }
 }
```
kkthxbye-code, line 38:
```suggestion
function lockDashboard(): void {
  const dashboard = document.getElementById('dashboard') as any;
  if (dashboard) {
    dashboard.gridstack.disable();
    // Hide the grid stack items buttons
    document.querySelectorAll(".grid-stack-item-button, #unlock_dashboard, #lock_dashboard").forEach(el => el.classList.toggle('invisible'));
  }
}

function unlockDashboard(): void {
  const dashboard = document.getElementById('dashboard') as any;
  if (dashboard) {
    dashboard.gridstack.enable();
    // Show the grid stack items buttons
    document.querySelectorAll(".grid-stack-item-button, #unlock_dashboard, #lock_dashboard").forEach(el => el.classList.toggle('invisible'));
  }
}
```

Just toggling the class seems simpler and works on my end, but not sure if I'm missing any edge cases.

kkthxbye-code (commented):
I tested this out and it works as advertised. Made a suggestion on simplifying the logic, but LGTM either way.

kkthxbye-code:
As a comment on the FR itself, hiding the x button does make it more tedious to delete multiple widgets as the locked/unlocked state is not preserved between reloads, so you have to unlock after each deletion. Not sure if it's a big deal though.

netbox/project-static/src/dashboard.ts
jsenecal: I really like your approach, I'm not seeing any obvious edge cases other than the browser potentially loosing track of the states (shouldn't happen) and making both the lock and unlock buttons visible? 

In any case I'm going to commit your approach to this PR

jsenecal:
> As a comment on the FR itself, hiding the x button does make it more tedious to delete multiple widgets as the locked/unlocked state is not preserved between reloads, so you have to unlock after each deletion. Not sure if it's a big deal though.

We could persist the state (locked/unlocked) through page reloads to address this.

Issue #13079, opened 2023-07-03
Make the dashboard cards/widget headers taller/bigger


jsenecal:
### NetBox version

v3.5.4

### Feature type

Change to existing functionality

### Proposed functionality

Prior to the implementation of Dashboard widgets introduced by #9416, the dashboards had taller headers which, looked better IMHO. Lets bring them back.

![image](https://github.com/netbox-community/netbox/assets/1564902/a63bbe40-c575-4ca8-ba02-14fda2eb12c1)
Versus:
![image](https://github.com/netbox-community/netbox/assets/1564902/3e288f16-4869-4e7f-a4e2-2dadd2e90936)

(Screenshots include other changes from https://github.com/netbox-community/netbox/issues/13077, sorry for this)

### Use case

Brings a more consistent look and feel to the dashboard page.

### Database changes

None

### External dependencies

None

jeremystretch:
> Prior to the implementation of Dashboard widgets introduced by https://github.com/netbox-community/netbox/issues/9416, the dashboards had taller headers which, looked better IMHO. Lets bring them back.

This proposal would roughly double the amount of valuable screen space consumed by each header while conveying no additional information to the user. But this is a moot point as we're planning to move to [Tabler](https://tabler.io/) in NetBox v4.0, and will be adopting its card styling.

PR #13490, opened 2023-08-16
Allow render_markdown to have custom css classes

### LINKED ISSUE #13055: Allow the "markdown" template filter to add css class names to the rendered div class attribute
### NetBox version

v3.5.4

### Feature type

### Proposed functionality

As implemented in https://github.com/netbox-community/netbox/pull/12664/commits/3e852acc8ec48f5b7d32811f4d9dec0ccdfa5970, this would allow the `div` generated by the `markdown` template filter to have additional custom classes appended to the list of classes.

### Use case

This extends usability of the template filter for use cases where one needs to add CSS classes in a specific rendering context.

### Database changes

None

### External dependencies

None

jsenecal:
(cherry picked from commit 3e852acc8ec48f5b7d32811f4d9dec0ccdfa5970)

### Fixes: #13055

This extends usability of the `markdown` template filter for use cases where one needs to add CSS classes in a specific rendering context.

### DIFF
--- a/netbox/utilities/templatetags/builtins/filters.py
+++ b/netbox/utilities/templatetags/builtins/filters.py
@@ -35,6 +35,7 @@ register = template.Library()
 # General
 #
 
+
 @register.filter()
 def linkify(instance, attr=None):
     """
@@ -129,6 +130,7 @@ def tzoffset(value):
 # Content types
 #
 
+
 @register.filter()
 def content_type(model):
     """
@@ -152,22 +154,30 @@ def content_type_id(model):
 # Rendering
 #
 
+
 @register.filter('markdown', is_safe=True)
-def render_markdown(value):
+def render_markdown(value, classes=""):
     """
     Render a string as Markdown. This filter is invoked as "markdown":
+    Args: optional coma separated list of classes to add to the generated div
+    (stripped of leading and trailing whitespace)
 
-        {{ md_source_text|markdown }}
+        {{ md_source_text|markdown:"classa, classb,classc, classd " }}
+
     """
     if not value:
         return ''
+
+    if classes:
+        # Remove any leading or trailing whitespace from each class name
+        classes = ", ".join(map(str.strip, classes.split(',')))
 
     # Render Markdown
     html = markdown(value, extensions=['def_list', 'fenced_code', 'tables', StrikethroughExtension()])
 
     # If the string is not empty wrap it in rendered-markdown to style tables
     if html:
-        html = f'<div class="rendered-markdown">{html}</div>'
+        html = f'<div class="rendered-markdown{" "+classes if classes else ""}">{html}</div>'
 
     schemes = get_config().ALLOWED_URL_SCHEMES
 
--- a/netbox/utilities/templatetags/builtins/filters.py
+++ b/netbox/utilities/templatetags/builtins/filters.py
@@ -159,7 +159,8 @@ def content_type_id(model):
 def render_markdown(value, classes=""):
     """
     Render a string as Markdown. This filter is invoked as "markdown":
-    Args: optional coma separated list of classes to add to the generated div
+    Args:
+        optional: Comma separated list of classes to add to the generated div
     (stripped of leading and trailing whitespace)
 
         {{ md_source_text|markdown:"classa, classb,classc, classd " }}


netbox/utilities/templatetags/builtins/filters.py
```diff
@@ -152,22 +154,30 @@ def content_type_id(model):
 # Rendering
 #
 
+
 @register.filter('markdown', is_safe=True)
-def render_markdown(value):
+def render_markdown(value, classes=""):
     """
     Render a string as Markdown. This filter is invoked as "markdown":
+    Args: optional coma separated list of classes to add to the generated div
```
kkthxbye-code, line 162:
```suggestion
    Args: 
        optional: Comma separated list of classes to add to the generated div
```

Typo in comma and change to match format used other places for Arg lists.

netbox/utilities/templatetags/builtins/filters.py
```diff
@@ -152,22 +154,30 @@ def content_type_id(model):
...
+        # Remove any leading or trailing whitespace from each class name
+        classes = ", ".join(map(str.strip, classes.split(',')))
+
     # Render Markdown
     html = markdown(value, extensions=['def_list', 'fenced_code', 'tables', StrikethroughExtension()])
 
     # If the string is not empty wrap it in rendered-markdown to style tables
     if html:
-        html = f'<div class="rendered-markdown">{html}</div>'
+        html = f'<div class="rendered-markdown{" "+classes if classes else ""}">{html}</div>'
```
kkthxbye-code, line 180: Now, I'm not really sure I think it's necessary to add class support to the filter, as the caller can just wrap it in another div, but if it is, I think the parsing can be omitted and just let the caller pass something like this:

`{{ md_source_text|markdown:"classa classb classc" }}`

Then the code can be simplified to:

```suggestion
        html = f'<div class="rendered-markdown {classes}">{html}</div>'
```

kkthxbye-code (changes requested):
I think this might be going to far with splitting up the original PR. Looking at the original, passing classes to render_markdown is only used one place and only to add margin to the top of the widget body. Can't that be done in the widget template instead, something like here?

https://github.com/netbox-community/netbox/blob/16e2283d192c9726c5010144df576baa3b16ccd7/netbox/templates/extras/dashboard/widget.html#L33

netbox/utilities/templatetags/builtins/filters.py
```diff
@@ -152,22 +154,30 @@ def content_type_id(model):
...
+        {{ md_source_text|markdown:"classa, classb,classc, classd " }}
 
-        {{ md_source_text|markdown }}
     """
     if not value:
         return ''
 
+    if classes:
+        # Remove any leading or trailing whitespace from each class name
+        classes = ", ".join(map(str.strip, classes.split(',')))
```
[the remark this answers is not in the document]
jsenecal: Filter parameters are passed as a single string in templates, this cleans up the passed argument.

netbox/utilities/templatetags/builtins/filters.py
jsenecal: The way the templates are currently designed, there is no good way to wrap this div in another one. Widgets use a generic template and doing what you suggested requires custom templates for markdown widgets.

Regarding your suggestion, classes may be "None" which is not desirable in the generated HTML.

jsenecal:
> Can't that be done in the widget template instead, something like here?

Addressed here: https://github.com/netbox-community/netbox/pull/13490#discussion_r1297304635

netbox/utilities/templatetags/builtins/filters.py
kkthxbye-code: > Regarding your suggestion, classes may be "None" which is not desirable in the generated HTML.

Classes has a default of empty string though:

`def render_markdown(value, classes=""):`

One would assume we would call the filter with no parameters if there are no extra classes.

> Widgets use a generic template and doing what you suggested requires custom templates for markdown widgets.

I can only go on the usecase in the original PR, and there you only use this to add top margin to all widget bodies, which can easily be done in the generic widget template as I linked earlier:

https://github.com/netbox-community/netbox/blob/16e2283d192c9726c5010144df576baa3b16ccd7/netbox/templates/extras/dashboard/widget.html#L33

kkthxbye-code: Maybe I'm misunderstanding, is there a usecase for this in markdown widgets (I assume that means the note widget?). I don't see how the user can pass classes in that case, even with this PR.

kkthxbye-code: Looking at it, I think it makes more sense to either just use a custom template for the note widget (like for the objectcount and rss widget), or add an optional template variable to the base widget class that allows to set extra classes on card body (just do padding instead).

jsenecal: > Maybe I'm misunderstanding, is there a usecase for this in markdown widgets (I assume that means the note widget?). I don't see how the user can pass classes in that case, even with this PR.

The use case is tightly coupled with changes required for another feature to be supported in the widgets (which is regarding taller headers) other widgets have better support for padding/margins but the note widget rendered improperly and needed that change in order to not introduce additional templates and change more than what I felt was required.

This added functionality could however be used by plugin authors so this is why I felt it could be it's own feature.

If you think otherwise, I would be okay with closing both the PR and FR in favor of addressing those requirements in context of the other FR (#13079) and perhaps working on the custom templates idea instead over there if accepted.

Thanks for your time in evaluating this.

PR #13502, opened 2023-08-17
11272 make position field work correctly when internationalization enabled

### LINKED ISSUE #11272: Error when saving device position
### NetBox version

v3.4.1

### Python version

3.10

### Steps to Reproduce

1. Open Netbox in chrome (it works in firefox)
2. Go to a device 
3. Click edit. The position should say something like UX.0
4. Click save without changing anything

### Expected Behavior

The device should be saved with the position UX.0 or UX

### Observed Behavior

The UX.0 gets transformed to UX,0 which can't be saved and an error occurs.

arthanson:
### Fixes: #11272 

Fixes position form field when turning on ENABLE_INTERNATIONALIZATION and selecting browser language that uses commas as separators.

### DIFF
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -401,12 +401,13 @@ class DeviceForm(TenancyForm, NetBoxModelForm):
     position = forms.DecimalField(
         required=False,
         help_text=_("The lowest-numbered unit occupied by the device"),
+        localize=True,
         widget=APISelect(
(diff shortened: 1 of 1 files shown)


jeremystretch:
This seems like it would affect _every_ DecimalField, no? Should we tackle all of these at once?

arthanson:
> This seems like it would affect _every_ DecimalField, no? Should we tackle all of these at once?

It actually doesn't effect the standard decimal fields from what I can see - I tried it on site lat/long and cable length (both decimal fields) and it works fine.

PR #13509, opened 2023-08-18
12114 show filename and line number on 500 error page

### LINKED ISSUE #12114: Show filename and line number in summary exception report
### NetBox version

v3.4.7

### Feature type

### Proposed functionality

Currently, when there's an internal exception, Netbox shows a very minimal exception message, referring users to the forum for more info: for example

> &lt;class 'TypeError'>
> 
> ObjectVar.\_\_init\_\_() missing 1 required positional argument: 'model'
>
> Python version: 3.10.6
> NetBox version: 3.4.7
> If further assistance is required, please post to the [NetBox discussion forum](https://github.com/netbox-community/netbox/discussions) on GitHub.

I propose that this should include at least the filename and line number where the exception occurred, i.e. the innermost line of backtrace.

I do not see this as a significant information leak or security issue.

### Use case

People come to the forum with these errors, and in many cases it would be much quicker to help them (or for them to help themselves) if the file name and line number were known.

In particular, errors in the user's own custom scripts or reports could easily be identified as such.

Example: discussion #12111

### Database changes

None

### External dependencies

None

arthanson:
### Fixes: #12114 

Adds the filename and line number to 500 page error report that is potentially useful in tracking down the error without exposing a security risk.  Not as useful as a full stack trace, but as pointed out in the ticket it can sometimes help to quickly identify if it is in a plugin.  Note however that it only shows the last line of the stack-trace which can be a library routine (for example a raised exception) - doing the full stack trace would give much better debugging info but is probably too much of a potential security leak?

Note: screenshot is from a raise exception, so in this specific case not very useful.

Monosnap Server Error 2023-08-18 12-43-26

### DIFF
--- a/netbox/netbox/views/errors.py
+++ b/netbox/netbox/views/errors.py
@@ -1,3 +1,4 @@
+import os
 import platform
 import sys
 
@@ -49,10 +50,13 @@ def handler_500(request, template_name=ERROR_500_TEMPLATE_NAME):
     except TemplateDoesNotExist:
         return HttpResponseServerError('<h1>Server Error (500)</h1>', content_type='text/html')
     type_, error, traceback = sys.exc_info()
-
+    fname = os.path.split(traceback.tb_frame.f_code.co_filename)[1]
+    lineno = traceback.tb_lineno
     return HttpResponseServerError(template.render({
         'error': error,
         'exception': str(type_),
+        'fname': fname,
+        'lineno': lineno,
         'netbox_version': settings.VERSION,
         'python_version': platform.python_version(),
         'plugins': get_installed_plugins(),
--- a/netbox/templates/500.html
+++ b/netbox/templates/500.html
@@ -29,6 +29,8 @@
 <pre class="block"><strong>{{ exception }}</strong><br />
 {{ error }}
 
+file: {{ fname }} line: {{ lineno }}
+
 Python version: {{ python_version }}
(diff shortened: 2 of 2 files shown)


netbox/templates/500.html
```diff
@@ -29,6 +29,8 @@ <h5 class="card-header">
 <pre class="block"><strong>{{ exception }}</strong><br />
 {{ error }}
 
+file: {{ fname }} line: {{ lineno }}
+
```
jeremystretch, line 33:
```suggestion
Location: ({{ file_name}}:{{ line_number }})
```

netbox/netbox/views/errors.py
```diff
@@ -49,10 +50,13 @@ def handler_500(request, template_name=ERROR_500_TEMPLATE_NAME):
     except TemplateDoesNotExist:
         return HttpResponseServerError('<h1>Server Error (500)</h1>', content_type='text/html')
     type_, error, traceback = sys.exc_info()
-
+    fname = os.path.split(traceback.tb_frame.f_code.co_filename)[1]
```
jeremystretch, line 53: We may want to expose more of the path than just the filename. If we get an exception in `views.py` for example, the filename alone doesn't help much. We could strip `BASE_DIR` from the file's path and return the remainder.

jeremystretch (changes requested):

PR #13525, opened 2023-08-22
13396 upgrade graphiql

### LINKED ISSUE #13396: Upgrade GraphiQL
### Proposed Changes

Upgrade GraphiQL - the current stack uses "graphiql": "1.4.1" which is several major versions behind.  The new version is 3.0.5.  Also the netbox-graphiql uses an abandoned subscriptions-transport-ws which is superseded by graphql-ws.

### Justification

Using a several major version outdated version of GraphiQL and it is using a package that is no longer maintained.  I also believe there is are several security patches for GraphiQL.

arthanson:
### Fixes: #13396 

Upgrades GraphiQL browser.  There are quite a few more major versions, but this is the highest version that can be updated to without more major changes.  This was requested as there are some security fixes in these versions.


jeremystretch:
Thanks @arthanson!

PR #13538, opened 2023-08-23
Adds vlan child table to vlan group

### LINKED ISSUE #13536: Move VLAN group child VLANs to a separate tab
### NetBox version

v3.5.8

### Feature type

New functionality

### Proposed functionality

When viewing a VLAN group, its list of child VLANs is currently embedded beneath the group's attributes. The proposal is to move this list to a separate "VLANs" tab, similar to how we display child IP addresses for a prefix.

### Use case

This standardizes the display of child VLANs using the generic ObjectChildrenView, and in doing so resolves bug #13494.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/ipam/models/vlans.py
+++ b/netbox/ipam/models/vlans.py
@@ -116,6 +116,12 @@ class VLANGroup(OrganizationalModel):
             return available_vids[0]
         return None
 
+    def get_child_vlans(self):
+        """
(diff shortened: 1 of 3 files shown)


jeremystretch:
Thanks @abhi1693!

PR #13545, opened 2023-08-24
Clear all cache when lazy is not used

### LINKED ISSUE #13544: Remove content type lookup when clearing cached values
### Proposed Changes

```patch
index 9a29c54f5..e072c220a 100644
--- a/netbox/extras/management/commands/reindex.py
+++ b/netbox/extras/management/commands/reindex.py
@@ -69,10 +69,7 @@ class Command(BaseCommand):
         if not kwargs['lazy']:
             self.stdout.write('Clearing cached values... ', ending='')
             self.stdout.flush()
-            content_types = [
-                ContentType.objects.get_for_model(model) for model in indexers.keys()
-            ]
-            deleted_count = search_backend.clear(content_types)
+            deleted_count = search_backend.clear()
             self.stdout.write(f'{deleted_count} entries deleted.')
```

### Justification

When trying to debug the issue for #13507, I wanted to clear my existing cache so I ran `./manage.py reindex` expecting all values to be removed including any plugins that have been uninstalled. However, that was not the case and my search was still broken as the plugin values still existed within `CacheValue`

abhi1693:

### DIFF
--- a/netbox/extras/management/commands/reindex.py
+++ b/netbox/extras/management/commands/reindex.py
@@ -69,10 +69,7 @@ class Command(BaseCommand):
         if not kwargs['lazy']:
             self.stdout.write('Clearing cached values... ', ending='')
             self.stdout.flush()
-            content_types = [
-                ContentType.objects.get_for_model(model) for model in indexers.keys()
-            ]
-            deleted_count = search_backend.clear(content_types)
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks @abhi1693!

PR #13551, opened 2023-08-24
Closes #13415: Pass request context when rendering custom links in a table column

### LINKED ISSUE #13415: Include request object in custom link renderer on tables
### NetBox version

v3.5.7

### Feature type

### Proposed functionality

The custom field renderer should have access to the request object.

### Use case

I'd like to record a console server 'node reference' against a console port and create a custom link that deep links me to an SSH session on that console server.

The deep link needs to look something like ssh://{{ request.user }}:{{ object.cf['node_id'] }}@consoleserver.my.domain

If I create a custom link for console ports and then view the detail page for the console port, the custom link shows up and works fine, but if I add the custom link as a column to the console ports table view, the custom link returns 'Error' as the request object is not exposed to the custom link table column renderer.

### Database changes

None

### External dependencies

None

jeremystretch:
### Fixes: #13415

Extends `CustomLinkColumn` to include the [full context data](https://docs.netbox.dev/en/stable/customization/custom-links/#context-data) supported by custom links.

### DIFF
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -4,6 +4,7 @@ from urllib.parse import quote
 
 import django_tables2 as tables
 from django.conf import settings
+from django.contrib.auth.context_processors import auth
 from django.contrib.auth.models import AnonymousUser
 from django.db.models import DateField, DateTimeField
 from django.template import Context, Template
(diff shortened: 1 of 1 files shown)


netbox/netbox/tables/columns.py
```diff
@@ -510,25 +511,32 @@ def __init__(self, customlink, *args, **kwargs):
...
 
-    def render(self, record):
-        try:
-            rendered = self.customlink.render({
-                'object': record,
-                'obj': record,  # TODO: Remove in NetBox v3.5
+    def _render_customlink(self, record, table):
+        context = {
+            'object': record,
+            'obj': record,  # TODO: Remove in NetBox v3.5
```
abhi1693, line 517: Shouldn't this be removed now?

jeremystretch: It should have been removed in v3.5, yes, but we didn't, so I've retained it in my PR. It _has_ been removed in `feature` already for the upcoming v3.6 release.

arthanson (approved):

PR #13561, opened 2023-08-25
Adds missing status choicefield for vdc

### LINKED ISSUE #13556: Inconsistent status field in API for virtual-device-context
### NetBox version

v3.5.8

### Feature type

### Proposed functionality

When using the API or pynetbox module the status for a virtual-device-context is different from the response for a device or virtual-machine.
This lead to parsing errors and requires different parsing of status fields for a vdc.

For example in pynetbox working with a device or virtual-machine
item.status              will give back "Active"
item.status.label    will give back "Active"
item.status.value    will give back "active"

API Response for a device or virtual-machine looks like this:
   "status": {
        "value": "active",
        "label": "Active"
    },

For a virtual-device-context there are not value an label fields
item.status              will give back "active"

API Response for a device or virtual-device-context looks like this:
    "status": "active",

Simple solution is to use str(item.status).lower()

### Use case

It would improve the overall consistency of the API to align the status field for a VDC with the status fields for other objects.

### Database changes

no

### External dependencies

no

abhi1693:

### DIFF
--- a/netbox/dcim/api/serializers.py
+++ b/netbox/dcim/api/serializers.py
@@ -714,6 +714,7 @@ class VirtualDeviceContextSerializer(NetBoxModelSerializer):
     primary_ip = NestedIPAddressSerializer(read_only=True, allow_null=True)
     primary_ip4 = NestedIPAddressSerializer(required=False, allow_null=True)
     primary_ip6 = NestedIPAddressSerializer(required=False, allow_null=True)
+    status = ChoiceField(choices=VirtualDeviceContextStatusChoices)
 
     # Related object counts
     interface_count = serializers.IntegerField(read_only=True)


netbox/dcim/api/serializers.py
```diff
@@ -714,6 +714,7 @@ class VirtualDeviceContextSerializer(NetBoxModelSerializer):
     primary_ip = NestedIPAddressSerializer(read_only=True, allow_null=True)
     primary_ip4 = NestedIPAddressSerializer(required=False, allow_null=True)
     primary_ip6 = NestedIPAddressSerializer(required=False, allow_null=True)
+    status = ChoiceField(choices=VirtualDeviceContextStatusChoices)
```
jeremystretch, line 717: I think this needs `required=False` to prevent it from being mandatory for PATCH requests.

```suggestion
    status = ChoiceField(choices=VirtualDeviceContextStatusChoices, required=False)
```

abhi1693: Will that not make POST requests optional? Because the field on the model does not have any default value.

jeremystretch: Disregard: I tested it out and it works fine as-is.

PR #13563, opened 2023-08-25
Adds csv dialect detection to bulk import view

### LINKED ISSUE #13239: CSV import: add support for alternate separators
### NetBox version

v3.5.3

### Feature type

### Proposed functionality

- Currently the CSV import only supports `,` as field seperator – it would really help if it would support `;` and `\t` aka `	`, too.

### Use case

* copy&paste tables from spreadsheet software to the import text-area of the web UI, 
* import CSV files create with Numbers, etc.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:


netbox/utilities/forms/bulk_import.py
```diff
@@ -90,8 +90,17 @@ def _clean_csv(self, data):
         """
         Clean CSV-formatted data. The first row will be treated as column headers.
         """
+
+        # Determine the CSV dialect
```
jeremystretch, line 94: We should add some indication to the user that semicolons can also be used as delimiters now. Maybe on the form field's help text? It's also worth calling out in the documentation.

netbox/utilities/forms/bulk_import.py
```diff
@@ -90,8 +90,17 @@ def _clean_csv(self, data):
...
+        try:
+            # This uses a rough heuristic to detect the CSV dialect. If the data is malformed, we'll fall back to
+            # the default Excel dialect.
+            dialect = csv.Sniffer().sniff(data.strip(), delimiters=',; ')
```
jeremystretch, line 98: It would probably be a good idea to provide the user a mechanism for overriding this as well, like what we do for bulk import formats (CSV/JSON/auto).

jeremystretch, line 98: Should `delimiters` include a space? Should this be a tab (`\t`) instead?

jeremystretch (changes requested):

netbox/utilities/forms/bulk_import.py
abhi1693: I tried but the internals for sniff simply iterate over the single char like `[key for key in delimiters]` due to which it's not really possible to pass `\t`. I tried using a list of strings but my IDE was telling me the expected type is `str` or `None` only,

abhi1693: If you want specific selectors in the form we can get rid of this logic and simply pass the `delimiter` to the reader. Using sniff seemed more manageable in case we decide to include more delimiters or even provide ways for the user to set this in the configuration.

kkthxbye-code: Maybe I'm misunderstanding, but seems to work here?

```
>>> data = """test1\ttest2\ttest3
... dss\tdsds\tdsds"""
>>> print(data)
test1   test2   test3
dss     dsds    dsds
>>> dialect = csv.Sniffer().sniff(data.strip(), delimiters=',;\t')
>>> dialect.delimiter
'\t'
```

Also:

```
>>> [key for key in ",;\t"]
[',', ';', '\t']
```

abhi1693: This is odd, when I tried from the Import UI, the code blew up because it could not detect the delimiter

abhi1693: It's my fault I guess to assume the issue originated at the sniff function, rather it was with header checks

image

jeremystretch: We should retain the automatic detection as the default behavior, but also offer the option for the user to override it.

abhi1693: I have made some changes to the code, see if that fits what you were articulating. It may need more cleanup though.

netbox/utilities/forms/bulk_import.py
```diff
@@ -86,12 +93,33 @@ def _detect_format(self, data):
...
         Clean CSV-formatted data. The first row will be treated as column headers.
         """
+
+        if delimiter == CSVDelimiterChoices.AUTO:
+            # Determine the CSV dialect
+            try:
+                # This uses a rough heuristic to detect the CSV dialect. If the data is malformed, we'll fall back to
+                # the default Excel dialect. Note that delimiter can only be one character.
+                dialect = csv.Sniffer().sniff(
+                    data.strip(), delimiters=''.join([CSVDelimiterChoices.COMMA, CSVDelimiterChoices.SEMICOLON])
```
pv2b, line 107: Add CSVDelimiterChoices.TAB to this list. I've tried it locally, and it works fine for me. I know you're concerned that '\t' may be considered as two character by the Sniff internals, but trust me, it works. `'\t'` is turned into a single ascii code already by the Python interpreter's string parsing.

netbox/utilities/tests/test_forms.py
```diff
@@ -331,3 +331,36 @@ def test_format_detection(self):
...
+
+        data = (
+            "a\tb\tc\n"
+            "1\t2\t3\n"
+            "4\t5\t6\n"
+        )
+        self.assertEqual(form._clean_csv(data, delimiter='\t'), [
+            {'a': '1', 'b': '2', 'c': '3'},
+            {'a': '4', 'b': '5', 'c': '6'},
+        ])
```
pv2b, line 366: You could add a test case for the auto-detection of tab-seperated data here. You'll find that auto-detection of CSV format is broken except for tab-seperated ones, because you also need to update the _detect_format function, right now that CSV autodetection works by looking for a , in the first line, which won't really work for ; or tab-seperated lines

netbox/utilities/forms/bulk_import.py
pv2b: This screenshot makes no sense, you're trying to put \t into the web form? That's not the same as a tab character.

Try creating a text file in Notepad with tabs seperating the fields and copy-paste this into the form instead.

netbox/utilities/choices.py
```diff
@@ -230,3 +230,17 @@ class ImportFormatChoices(ChoiceSet):
...
+
+
+class CSVDelimiterChoices(ChoiceSet):
+    AUTO = 'auto'
+    COMMA = ','
+    SEMICOLON = ';'
+    TAB = '\t'
+
+    CHOICES = [
+        (AUTO, _('Auto-detect')),
```
abhi1693, line 242: Maybe add a comment here to always keep `auto` on top since this drives the slicing logic in other places

abhi1693 (commented):
@jeremystretch Looks good to me with the exception of the slicing because that's dangeours if the `auto` is moved from the index `0`

jeremystretch:
> Looks good to me with the exception of the slicing

Good call. I've addressed this by moving the actual delimiting characters to a separate constant, which is wrapped by the choice set (with the addition of the `auto` option).

PR #13564, opened 2023-08-25
Adds rename button on the list page for device components

### LINKED ISSUE #13444: Add a rename button to the device interfaces list
### NetBox version

v3.5.7

### Feature type

New functionality

### Proposed functionality

Rename button under Interfaces, for bulk renaming of interfaces, as available for Devices.

Currently:
image

### Use case

Due to different operating systems, the names of the interfaces may change. A bulk update would facilitate the adaptation process.

### Database changes

I do not think so, not sure

### External dependencies

_No response_

abhi1693:


DanSheps (changes requested):
I am not sure if you saw the comment from Jeremy but this should be extended to all component views.

abhi1693:
@DanSheps Thanks, I really did miss that comment.

DanSheps (approved):
I see a lot of code duplication.  Any thoughts on a way to perhaps deduplicate some of this?

For example, a Mixin for ListView like "DeviceComponentListViewMixin", and then have a single template ("devicecomponent_list.html"). Just a thought, not saying we should or have to do this.

Otherwise, it looks good.

jeremystretch:
I've converged the model-specific templates into a single one following the pattern we have for component lists under the device view.

PR #13568, opened 2023-08-25
Adds parent filter on iprange

### LINKED ISSUE #13313: Add parent filter for IP Range objects in REST API
### NetBox version

v3.5.7

### Feature type

### Proposed functionality

Add the "parent" filter to IP Range objects in the same manner as it is implemented for IP Address objects. Ex:  "/api/ipam/ip-ranges/?parent=172.20.128.160/28"

### Use case

There are many reasons you may want to interact with all ranges defined inside a given prefix, such as cleaning up related records upon removal.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:


jeremystretch:
Thanks @abhi1693!

PR #13570, opened 2023-08-25
Adds related models count on custom field

### LINKED ISSUE #12825: custom-fields - usage counters / related objects box
### NetBox version

v.3.5.3

### Feature type

### Proposed functionality

when viewing a custom field's config details (https://demo.netbox.dev/extras/custom-fields/1/) it would be nice if it had a "related objects" list, showing how many times that field is used/set per object type.

example of related objects view is on the tenants page https://demo.netbox.dev/tenancy/tenants/1/

### Use case

when deprecating a custom field, it would be nice if the ui showed any existing usage of a field has been cleaned up

its also useful to see how popular certain custom fields are

### Database changes

unknown

### External dependencies

unknown

abhi1693:
### Fixes: #12825

Note: The related models cannot have deterministic filtering on the URL and so I have removed the href property from the anchor tag.

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -44,9 +44,7 @@ class CustomFieldView(generic.ObjectView):
     queryset = CustomField.objects.all()
 
     def get_extra_context(self, request, instance):
-        related_models = (
-            # (DataFile.objects.restrict(request.user, 'view').filter(source=instance), 'source_id'),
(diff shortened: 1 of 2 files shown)


kkthxbye-code (approved):
Tested it, works as advertised and performance of the query seems to be just fine.

PR #13586, opened 2023-08-28
Closes #13585: Introduce 'empty' lookup for numeric value filters

### LINKED ISSUE #13585: Enable filtering for null numeric values using the `empty` lookup
### NetBox version

v3.5.8

### Feature type

New functionality

### Proposed functionality

NetBox currently supports checking for empty string fields using the `empty` lookup (e.g. `?description__empty=true`). This FR proposes extending that functionality to numeric fields as well, to match null/non-null values.

### Use case

This enables a user to query numeric fields generally for null or non-null values. For example, `?position__empty=true` would return devices which do not have a rack position defined.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Fixes: #13585

Automatically registers an `empty` lookup to filter non null/not-null numeric values (e.g. device position).

### DIFF
--- a/docs/reference/filtering.md
+++ b/docs/reference/filtering.md
@@ -61,13 +61,14 @@ These lookup expressions can be applied by adding a suffix to the desired field'
 
 Numeric based fields (ASN, VLAN ID, etc) support these lookup expressions:
 
-| Filter | Description |
-|--------|-------------|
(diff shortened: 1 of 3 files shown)


DanSheps (approved):
Filter works as advertisied, LGTM

PR #13617, opened 2023-08-30
Fixes: #11466 - Add `device` field to L2VPNTermination for assigning vlan terminations to devices

### LINKED ISSUE #11466: Assign VLAN to multiple L2VPNs
### NetBox version

v3.3

### Feature type

### Proposed functionality

We document VLANs transferred through a fabric with L2VPNs, e.g. adding the VLAN to the fabric-specific L2VPN.

Currently we are not able to add the VLAN to two different L2VPN instances, which means that we cannot document a VLAN that is transferred through two different fabrics.

Therefore, we propose to remove the requirement, that an object can only be attached to one L2VPN, at least for VLANs. I see that this limit is useful for interfaces.

### Use case

We have two different fabrics, a "Core Fabric" in our data center and a "WAN Fabric" for our wide area network. We provide services to our customers tunneling the same VLAN from a port in the data center to a port in a WAN pop. So this VLAN needs to be transferred through the "Core Fabric" and the "WAN Fabric".

### Database changes

None

### External dependencies

None

DanSheps:
### Fixes: #11466 - Add `device` field to L2VPNTermination for assigning vlan terminations to devices

* Add `device` field to L2VPNTermination model
* Add related fields to relevant forms, tables, filters and views.


jeremystretch:
This needs to be tied to a milestone release as it's introducing a new model field.

DanSheps:
I added this to the 3.7 Milestones

jeremystretch:
I'm not sure this is the best way to solve (what I believe is) the challenge surfaced in #11466. IMO it would make more sense in that scenario to model a separate VLAN at each site, and attach each of them to the L2VPN. Trying to model two attachments to the same VLAN in this manner is going to be challenging and likely error-prone.

Additionally, the modified constraints don't fully validate the assignment of other object types. For instance, under this PR it becomes possible to attach a single interface to multiple L2VPNs.

I don't feel that we have a clear enough understanding of the challenge(s) to be solved here to be comfortable proceeding with these changes.

DanSheps:
I think we are good for a preliminary dive in on this.

Couple things:

* VLANDeviceMapping is attached to ipam.  Just on the off chance it can be used elsewhere I thought it best there.  We can move it to L2VPN as well if you think that is better
* Thinking about this, this morning.  We could remove VLAN as an option to map to and instead have either a "VLAN Mapping", "Interface" or "VMInterface".  This would clean up the filterset for L2VPN vlan filter but it would require allowing device to be empty on the mapping itself.

arthanson:
IMHO I see several issues with this:

First - documentation is missing for this.

Second - This sort of went in the original issue from "lets add a device field to L2VPN termination" to this PR which adds a completely new model VLANDeviceMapping without any real discussion around it.  While this might be the correct solution, I am leery of this without an FR discussion as it is adding a new concept that isn't reflected in the real world - i.e. VLANDeviceMapping is something NetBox makes up and isn't something a network engineer would know about. When trying to test it I was at first confused as I went to the new tab in the UI and there were no items in the list even though I had VLan and devices defined.  Finally saw that I had to go under the VLAN menu and add this new "concept".

Most everything in NetBox is a physical item or concept a network engineer is already familiar with, I just think adding in something like this that is NetBox specific requires more discussion to make sure it is the right and best solution. I'm sure there are ramifications in the future for this new model so we should make sure this is the direction we want to take before adding in a new concept like this.

DanSheps:
Blocked by #14451

DanSheps:
Closing for now - Will resume work after blocking issue is resolved

PR #13618, opened 2023-08-30
Adds additional parameters for token provision api

### LINKED ISSUE #12870: Support for setting expiration time when using the API token provisioning endpoint (/api/users/tokens/provision/)

abhi1693:
### Fixes: #12870

Note: @stuntguy3000 initially submitted #12911 but it had a different approach. I have moved the validation under the serializer from the view and extended the `TokenSerializer` onto `TokenProvisionSeriazlier` which works for both request and response.

### DIFF
--- a/netbox/users/api/serializers.py
+++ b/netbox/users/api/serializers.py
@@ -1,11 +1,12 @@
 from django.conf import settings
+from django.contrib.auth import authenticate
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group
 from django.contrib.contenttypes.models import ContentType
 from drf_spectacular.utils import extend_schema_field
 from drf_spectacular.types import OpenApiTypes
 from rest_framework import serializers
-from rest_framework.exceptions import PermissionDenied
+from rest_framework.exceptions import AuthenticationFailed, PermissionDenied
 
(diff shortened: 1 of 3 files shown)


jeremystretch:
Thanks @abhi1693!

PR #13641, opened 2023-08-31
Fixes #13620: Show admin menu items only for staff users

### LINKED ISSUE #13620: Cannot Remove Authentication > Tokens access with permissions
### NetBox version

v3.6.0

### Python version

3.10

### Steps to Reproduce

1. Create user with very limited access permissions (e.g view DCIM > Circuits)
2. Login with user
3. Admin > Authentication > API Tokens still visible

### Expected Behavior

API Tokens and Admin tabs should be hidden without permissions to view them.

### Observed Behavior

Option is visible in navigation bar

I notice the affected page is /user/tokens/ as opposed to /user/api-tokens and I'm not sure if one is a replacement for the other or what the plans may be here.

jeremystretch:
### Fixes: #13620

Set `staff_only=True` on all menu items under the "administration" heading

### DIFF
--- a/netbox/netbox/navigation/menu.py
+++ b/netbox/netbox/navigation/menu.py
@@ -360,6 +360,7 @@ ADMIN_MENU = Menu(
                     link=f'users:netboxuser_list',
                     link_text=_('Users'),
                     permissions=[f'auth.view_user'],
+                    staff_only=True,
                     buttons=(
                         MenuItemButton(
                             link=f'users:netboxuser_add',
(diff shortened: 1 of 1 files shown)


abhi1693 (commented):
Will this still work for a superuser that does not have the is_staff flag set?

jeremystretch:
No; the menu items will still be hidden for a superuser if the user does not have `is_staff` set to True. IIRC this is consistent with Django's own admin UI.

PR #13667, opened 2023-09-03
Fixes #13666: Fix behavior for reports without test methods

### LINKED ISSUE #13666: Uploading report without test method yields "Server Error"
### NetBox version

v3.6.0 (But reproducible in 3.6.1-dev, too)

### Python version

3.11

### Steps to Reproduce

1.Create `a.py`:
```py
from extras.reports import Report

class DeviceConnectionsReport(Report):
    description = "Validate the minimum physical connections for each device"
    pass
```
(Note, the class doesn't have a test method)

2.Upload it as report

### Expected Behavior

I expected it to either get an error, that there are no test methods in it, and the ability to delete it without using the CLI.

### Observed Behavior

Reality:
```
<class 'Exception'>

A report must contain at least one test method.

Python version: 3.11.4
NetBox version: 3.6.1-dev
Plugins: None installed
```

image

Screenshot from develop branch, but reproducible with stable, too

JCWasmx86:
### Fixes: #13666

If you upload a report file without any test methods, the entire GUI for the reports page fails. This fixes the error by splitting up the `report_modules` into valid and empty ones. The valid `report_module`s are rendered as before, the empty ones are listed with the possibility to delete them. (Before you had to drop into the CLI to delete them)

The list of reports without test methods are at the top of the page to increase the visibility of them, but you could consider moving them into the normal list of reports, but show an error/warning there - similar to e.g. "Failed to load foo.py"

Additionally there is a second commit (But can be either dropped/moved to new PR as it changes the scope of the issue a little bit). It fixes the API by filtering out those reports

*Note: I wasn't able to run black, as `black --config pyproject.toml netbox/` wants to reformat everything, but `pycodestyle --ignore=W504,E501 netbox/` shows no warnings)*

### DIFF
--- a/netbox/extras/reports.py
+++ b/netbox/extras/reports.py
@@ -106,8 +106,6 @@ class Report(object):
                     'failure': 0,
                     'log': [],
                 }
-        if not test_methods:
-            raise Exception("A report must contain at least one test method.")
         self.test_methods = test_methods
 
     @classproperty
--- a/netbox/templates/extras/report.html
+++ b/netbox/templates/extras/report.html
@@ -12,7 +12,11 @@
             {% csrf_token %}
             {% render_form form %}
             <div class="float-end">
-              <button type="submit" name="_run" class="btn btn-primary">
+              <button type="submit" name="_run" class="btn btn-primary"
+              {% if not report.test_methods|length %}
+                disabled
+              {% endif %}
+              >
                 {% if report.result %}
                   <i class="mdi mdi-replay"></i> {% trans "Run Again" %}
                 {% else %}
--- a/netbox/templates/extras/report_list.html
+++ b/netbox/templates/extras/report_list.html
@@ -68,10 +68,16 @@
                           </td>
                         {% else %}
                           <td class="text-muted">{% trans "Never" %}</td>
-                          <td>{{ ''|placeholder }}</td>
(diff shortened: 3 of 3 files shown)


jeremystretch (changes requested):
A much simpler fix would be to simply remove the exception (below) and perhaps add a warning to the UI when viewing a report with no test methods. (It might also make sense to disable the "run" action.)

https://github.com/netbox-community/netbox/blob/0ce2b1b779228befa55d867eacb863b7f2ee1684/netbox/extras/reports.py#L109-L110

JCWasmx86:
Yes I will do that. I hesitated to remove the exception, as I thought it was maybe important (And just not documented, why missing test methods should be an error instead of an warning - thus this more complex solution)

JCWasmx86:
Do you like a design like that:

image

Or should the alert be an entire table row below the ReportClass?

jeremystretch:
I would change the text to something like "Invalid (no test methods found)" and remove the alert styling (it's not intended for table cells). But otherwise yes I think this is great!

JCWasmx86:
Looks like that now:

image


The button here is now disabled:

image

jeremystretch (approved):
Thanks @JCWasmx86!

PR #13677, opened 2023-09-05
Add support for TSV and semicolon-seperated CSV

### LINKED ISSUE #13239: CSV import: add support for alternate separators

pv2b:
### Fixes: #13239

This is an alternative PR for issue #13239. A different PR #13563 exists that attempts to solve the same issue, however I think my approach is better.

1. The other approach creates a new field for CSV seperators. This makes the UI a lot messier and adds more clicks to importing CSV files for the user. In this solution, instead, semicolon-seperated CSV and tab-seperated CSV are added as new formats and the user only has one drop-down to make a selection in.
2. There was some confusion in the code regarding auto-detection of file formats in general and autodetecting different types of CSV file specifically. @abhi1693 was attempting to use csv.Sniff to detect a specific CSV file format, but was running into some problems that I don't understand with making it work with tab-seperated files. Anyway, this approach completely sidesteps csv.Sniff, just adding a couple more cases into _detect_format.
3. The other approach wasn't complete, for example, it was not able to automatically detect tab- or semicolon-seperated CSV files when set to automatic mode. This I think is also due there being two ways of detecting CSV files.

There's one area of concern for breakage here. I've had to disable Django's own built-in stripping of whitespace, because with TSV files, whitespace is significant, and removing trailing tabs can remove an empty column, which isn't what you want. I think I've anticipated every case where this could break something (for example, _detect_format expects the file to be stripped, so I had to add some stripping back into that function) but this could probably use a second pair of eyes to make sure.

### DIFF
--- a/netbox/netbox/tests/test_import.py
+++ b/netbox/netbox/tests/test_import.py
@@ -63,6 +63,61 @@ class CSVImportTestCase(ModelViewTestCase):
         return self._test_valid_tags(csv_data, ImportFormatChoices.CSV)
 
     @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])
+    def test_valid_tags_csv_autodetect(self):
+        csv_data = (
+            'name,slug,tags',
+            'Region 1,region-1,"alpha,bravo"',
+            'Region 2,region-2,"charlie,delta"',
+            'Region 3,region-3,echo',
+            'Region 4,region-4,',
+        )
+        return self._test_valid_tags(csv_data, ImportFormatChoices.AUTO)
+
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])
+    def test_valid_tags_csv_semicolon(self):
+        csv_data = (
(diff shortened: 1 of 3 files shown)


jeremystretch:
As you note, #13239 has already been assigned and is being worked on. Please do not submit competing PRs without first collaborating with the issue's assigned owner. If you are interested in contributing to NetBox, please consider adopting one of [the many unassigned issues](https://github.com/netbox-community/netbox/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+needs+owner%22) currently in need of an owner.

PR #13688, opened 2023-09-05
13674 fix ReportSerializer

### LINKED ISSUE #13674: Server Error accessing specific Report via API
### NetBox version

v3.6.0

### Python version

3.11

### Steps to Reproduce

1. Go to /api/extras/reports/
2. Got to any of the URLs returned in the "url" fields
3. Server Error:
image

### Expected Behavior

The API should return the Details of the specific report.
Working example in NetBox v3.4.4:
image

### Observed Behavior

Accessing  `/api/extras/reports/datacenter_sites.ErrorCheckSummary/` in NetBox v3.6.0 does not return an answer like in v3.4 but instead creates a Server Error.

arthanson:
### Fixes: #13674 

Fixes test_methods under report serializer, makes read-only and initializes it as class attribute (as opposed to just inside init) so that serializer sees it.

### DIFF
--- a/netbox/extras/reports.py
+++ b/netbox/extras/reports.py
@@ -85,9 +85,9 @@ class Report(object):
     description = None
     scheduling_enabled = True
     job_timeout = None
+    test_methods = []
 
     def __init__(self):
-
         self._results = {}
         self.active_test = None
         self.failed = False
--- a/netbox/extras/reports.py
+++ b/netbox/extras/reports.py
@@ -85,9 +85,9 @@ class Report(object):
     description = None
     scheduling_enabled = True
     job_timeout = None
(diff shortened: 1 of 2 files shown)


DanSheps:
I believe this is a duplicate of #13667

netbox/extras/reports.py
```diff
@@ -85,9 +85,9 @@ class Report(object):
     description = None
     scheduling_enabled = True
     job_timeout = None
+    test_methods = []
```
jeremystretch, line 88: This list won't be populated if the report class hasn't been initialized. IMO it would be preferable to omit the attribute entirely rather than to assert that there are no test methods.

PR #13747, opened 2023-09-12
13745 device type migration

### LINKED ISSUE #13745: v3.6 database migration logic is inefficient for large numbers of device types
### NetBox version

v3.6.1

### Python version

3.8

### Steps to Reproduce

This is an extension to bug #13605, which was addressed for device and virtual machines during the v3.6 beta testing period. The issue described therein impacts device types as well, as raised in discussion #13676.

### Expected Behavior

Schema migrations should apply without error, and should not consume excessive disk space or memory.

### Observed Behavior

Attempting to upgrade to v3.6.0 or later with a large number of device types (over 3,000 as tested) can trigger excessive disk space utilization due to inefficiencies in the component template count logic.

We should be able to resolve this in a manner similar to the fix for #13605.

arthanson:
### Fixes: #13745 

Optimize counter field migration on DeviceType for large databases.

### DIFF
--- a/netbox/dcim/migrations/0176_device_component_counters.py
+++ b/netbox/dcim/migrations/0176_device_component_counters.py
@@ -2,47 +2,22 @@ from django.db import migrations
 from django.db.models import Count
 
 import utilities.fields
+from utilities.counters import update_counts
 
 
 def recalculate_device_counts(apps, schema_editor):
     Device = apps.get_model("dcim", "Device")
-    devices = Device.objects.annotate(
-        _console_port_count=Count('consoleports', distinct=True),
-        _console_server_port_count=Count('consoleserverports', distinct=True),
-        _power_port_count=Count('powerports', distinct=True),
-        _power_outlet_count=Count('poweroutlets', distinct=True),
-        _interface_count=Count('interfaces', distinct=True),
-        _front_port_count=Count('frontports', distinct=True),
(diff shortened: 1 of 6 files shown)


jeremystretch:
While this addresses the immediate concern, I'd argue that we should utilize the same approach for all three affected migrations (`dcim.0176_device_component_counters` and `virtualization.0035_virtualmachine_interface_count` as well).

Also, we probably want to move the logic from the management command somewhere internally so that it be can be reused by the migrations.

jeremystretch (changes requested):
While this addresses the immediate concern, I'd argue that we should utilize the same approach for all three affected migrations (`dcim.0176_device_component_counters` and `virtualization.0035_virtualmachine_interface_count` as well).

Also, we probably want to move the logic from the management command (rather than duplicating it) somewhere internally so that it be can be reused by the migrations.

arthanson:
> While this addresses the immediate concern, I'd argue that we should utilize the same approach for all three affected migrations (`dcim.0176_device_component_counters` and `virtualization.0035_virtualmachine_interface_count` as well).
> 
> Also, we probably want to move the logic from the management command (rather than duplicating it) somewhere internally so that it be can be reused by the migrations.

Moved the calculate routine to utilities then updated all migrations to use this pattern.

jeremystretch (approved):

PR #13761, opened 2023-09-14
Disables module_status ordering

### LINKED ISSUE #13756: Sort on module bay status field error
### NetBox version

v3.6.1

### Python version

3.10

### Steps to Reproduce

1. Create a module bay
2. Sort Module bay on Status
3. Click on module bay tab again

### Expected Behavior

Either:
1. Sort on Module Bay Status
2. No sorting on module bay status allowed

### Observed Behavior

`<class 'django.core.exceptions.FieldError'>`

image

abhi1693:

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -871,9 +871,9 @@ class ModuleBayTable(DeviceComponentTable):
         url_name='dcim:modulebay_list'
     )
     module_status = columns.TemplateColumn(
-        verbose_name=_('Module Status'),
+        accessor=tables.A('installed_module__status'),
         template_code=MODULEBAY_STATUS,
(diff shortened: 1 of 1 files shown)


netbox/dcim/tables/devices.py
```diff
@@ -872,7 +872,8 @@ class ModuleBayTable(DeviceComponentTable):
     )
     module_status = columns.TemplateColumn(
         verbose_name=_('Module Status'),
-        template_code=MODULEBAY_STATUS
+        template_code=MODULEBAY_STATUS,
+        orderable=False
```
jeremystretch, line 876: We should be able to support ordering for this column if we define an accessor for it. `A('installed_module__status')` should work, I think.

PR #13793, opened 2023-09-17
13791 rename whitespace fix

### LINKED ISSUE #13791: Bulk rename improperly strips "find" and "replace" fields
### NetBox version

v3.6.1

### Python version

3.11

### Steps to Reproduce

1. On an existing device, create an interface labeled something like "One Two".
2. Go to the interface tab of that device, check the checkbox next to it and hit "rename".
3. In the find field, enter "One " (note the trailing space!). In the replacement field enter "One-".
4. Click the preview button.

### Expected Behavior

NetBox should offer to rename the device to "One-Two".

### Observed Behavior

Netbox offers to rename the device to "One- Two" (note the space).

My guess (I've not looked at the code) is that this happens because django autostrips whitespace from input fields by default, and this needs to be disabled.

Also, the same bug exists on the "replace" field, and is completely analagous to the replace field.

Workaround: For the "find" field, use regex mode and \s to match trailing whitespace. For the "replace" field, I don't know of a general workaround.

pv2b:
### Fixes: #13791

Disables Django's form stripping of the "find" and "replace" fields of the BulkRenameForm. This should fix all bulk rename forms in the project that inherit from this base class. Also added a unit test to make sure this works.

### DIFF
--- a/netbox/utilities/forms/forms.py
+++ b/netbox/utilities/forms/forms.py
@@ -40,8 +40,11 @@ class BulkRenameForm(BootstrapMixin, forms.Form):
     """
     An extendable form to be used for renaming objects in bulk.
     """
-    find = forms.CharField()
+    find = forms.CharField(
+        strip=False
+    )
     replace = forms.CharField(
+        strip=False,
         required=False
     )
     use_regex = forms.BooleanField(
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks @pv2b!

PR #13798, opened 2023-09-18
Feature/#13795 add script hide option

### LINKED ISSUE #13795: option to hide script classes from being visible
### NetBox version

v3.6.1

### Feature type

New functionality

### Proposed functionality

We would like to add a way to hide certain script classes from the UI when they do not have user-facing functionality (no run method).  There seem to be a few ways this could be implemented:
- Hide scripts that do not contain a run() method.  This check could potentially be added into is_script() so that scripts without a run() method are inherently hidden.  That is, codify the fact that a script without a run() method is not actually a script.
- Add an option to the script module similar to script_order that lists classes which should be suppressed (something like "script_hide=(MyScript)), then use that list in the creation of sxripts to exclude those scripts.  Add a check like "if cls not in script_hide" to the logic that creates the list of scripts)
- Add an option to the script module which says that "script_order" is an absolute/complete list, and those scripts not listed should not be visible.    Right now, any script not listed in "script_order" is just appended to the end.  Such an option perhaps would be a boolean script_display_unordered=True/False

### Use case

We have a class hierarchy for our Netbox scripts, to abstract common functionality.  We do not want our intermediate classes from the module(s) being visible in the UI, as those classes do not contain user functionality

### Database changes

None

### External dependencies

None

FliesLikeABrick:
### Fixes: #13795

Adding a possible implementation for hiding scripts from the user-facing listing.


abhi1693:

FliesLikeABrick:
Thanks @abhi1693  - #13795 is open for developer input.

PR #13804, opened 2023-09-18
13757 Fix ConfigContext reference to DeviceType

### LINKED ISSUE #13757: Config Context Fails to assign to "Device types"
### NetBox version

v3.6.0

### Python version

3.11

### Steps to Reproduce

1. Under Provisioning Tab Click on "Config Contexts"
2. Click "Add" to create Config Context fill in any Name and enter valid json in Data field
3. Go to bottom and click "Create" to save
4. Once Config Context is created go back to edit and try to assign to a "Device types" 

### Expected Behavior

it should have assigned config context to "Device types"

### Observed Behavior

<class 'RecursionError'>

maximum recursion depth exceeded in comparison

Python version: 3.11.4
NetBox version: 3.6.0
Plugins: None installed

arthanson:
### Fixes: #13757 

As per Django issue https://code.djangoproject.com/ticket/34847 problem arises as DeviceType has variable references in init method.  Recommendation is to access vars in init with self.__dict as they state this will probably be documented this way as there is no good fix for dereferenced vars except for this.

Changed ConfigContext test case save devicetype to catch if this ever regresses.

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -98,10 +98,10 @@ class Cable(PrimaryModel):
         super().__init__(*args, **kwargs)
 
         # A copy of the PK to be used by __str__ in case the object is deleted
-        self._pk = self.pk
+        self._pk = self.__dict__.get('id')
 
         # Cache the original status so we can check later if it's been changed
-        self._orig_status = self.status
+        self._orig_status = self.__dict__.get('status')
 
         self._terminations_modified = False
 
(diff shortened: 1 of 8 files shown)


abhi1693 (commented):
I'm not sure if it's really possible, but a test case to ensure this doesn't creep in later

jeremystretch:
Thanks @arthanson!

jeremystretch (approved):

Issue #13806, opened 2023-09-18
"Mark Installed" tooltip doesn't properly update to "Mark Planned" 


pv2b:
### NetBox version

v3.6.1

### Python version

3.8

### Steps to Reproduce

1. Go to an enabled non-virtual interface with an installed cable in the device interface view.
2. Edit the cable and mark it as planned. (Important, do it this way, not through the "mark planned" button.)
3. Hover over the "mark installed" button, notice that it says "mark installed".
4. Click the "mark installed" button to mark the cable as installed.
5. WITHOUT navigating away from the page, hover over the button again and read the tooltip on the button.

### Expected Behavior

The tooltip should read "mark planned" since the cable is now marked as installed and the change would be to mark it as planned.

### Observed Behavior

The tooltip actually reads "mark installed".

pv2b:
I found this bug when working on a fix for #13712, going to send in a fix for this in the same PR

DanSheps:
Since the previous issue was closed due to a lack of activity, I am putting this back to needs owner for now.  If someone would like to take this on, just reply here.

pv2b:
Hello,

The lack of activity is due to a lack of feedback from the Netbox developer side to my pull requests.

There's a pull request #13874 that's been avaiting maintainer feedback for the last 6 months that addresses this issue.

I also have other pull requests in for the Netbox projects that you have accepted me as an assignee on and then subsequently have seen no review from your side. It's disappointing since I've put effort into the project only for it to get ignored, which means I've not really been interested in contributing more to the project.

I'm happy to work with you on any concerns you have with any of my open issues or PR's, but I won't be contributing anything more until those existing PR's are addressed in some way other than a Github stalebot.

DanSheps:
@pv2b I will re-assign you.

In  the future, perhaps make sure you make "request review" from at least Jeremy (unless you feel it is better suited for someone else).  To be clear, I am not saying this is your fault at all, but it serves as a reminder for the people you request a review of to review it (it also shows in our notifications as a pending review request)

DanSheps:
Closed by #13874

PR #13807, opened 2023-09-18
Bugfixes for mark installed/mark planned button

### LINKED ISSUE #13712: "Mark planned" doesn't immediately change row highlight colour
### NetBox version

v3.6.1

### Python version

3.11

### Steps to Reproduce

Prerequisites: You need a device with a cable attached to one of its interfaces. The interface should not be disabled.

Since it's a client-side issue it's not impossible that the problem is browser specific. In case it turns out to be browser-specific, I've observed this behaviour on Firefox 117.0.

1. Open the "interfaces" view of the device in question.
2. Locate an interface with a cable connected to it. Note that the background colour of the row in the table is green to indicate it has a connected cable.
3. Click on the "marked planned" button (the third button from the right)
4. Notice that the colour of the button changed from yellow to blue to indicate that the cable is now planned, and no longer installed. However notice that the background colour of the row is still green.
5. Refresh the page, and notice that the background colour of the row is now correctly blue.

### Expected Behavior

The background colour of the row should have changed from blue to green as soon as the "Mark Planned" button was clicked.

### Observed Behavior

The background colour of the row did not change, and a page refresh was required to correctly display the current state.

(P.S. I'm willing to troubleshoot this issue and produce a PR, but as per project rules, I cannot do this before the issue has been accepted and assigned to me.)

pv2b:
### Fixes: #13712 and #13806.

Not super happy with the fix here, because it doesn't address the underlying problem, which is that the toggleConnection() typescript function hardcodes which CSS classes should be added/removed. Probably a more permanent fix would be to stop applying CSS classes on the table view, and instead apply attributes for cable/interface state, and then use CSS to apply colours based on interface state, but this is a quite involved process. But it does at least fix things in the here and now.


pv2b:
Let's hold off on this one for now, since this doesn't properly respect FIELD_CHOICES. I'm going to see if I can't come up with a better solution.

PR #13823, opened 2023-09-19
13813 fix virtual chassis member count

### LINKED ISSUE #13813: Virtual Chassis membercount issue
### NetBox version

v3.6.1

### Python version

3.9

### Steps to Reproduce

For full reproduction create at least 2 member devices as a prerequisite. 

1. Go to Virtual Chassis and press `Add`.
2. Name your Virtual Chassis f.e. `SW0201`
3. Add two member switches f.e. `SW0201.1` and `SW0201.2`
4. Click `Create`

Selecting a site/rack or not during creation does not make a difference in the observed behaviour.

### Expected Behavior

Creating a Virtual Chassis with the membercount being 2.

### Observed Behavior

Virtual Chassis is created with membercount being 0.
VC-0

Adding a new switch member to an already existing Virtual Chassis also doesn't increase the membercount.
VC-extramember

Deleting a member from the Virtual Chassis **does decrease** the membercount!
VC-minus

arthanson:
### Fixes: #13813 

Fixes VirtualChassis member count - issue was in counter fields where the tracked object (Device) existed but had the FK pointer that was being tracked as None which caused the increment to get bypassed.  Added test for this condition.

### DIFF
--- a/netbox/utilities/counters.py
+++ b/netbox/utilities/counters.py
@@ -52,12 +52,13 @@ def post_save_receiver(sender, instance, created, **kwargs):
     for field_name, counter_name in get_counters_for_model(sender):
         parent_model = sender._meta.get_field(field_name).related_model
         new_pk = getattr(instance, field_name, None)
-        old_pk = instance.tracker.get(field_name) if field_name in instance.tracker else None
+        has_old_field = field_name in instance.tracker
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thanks @arthanson!

PR #13825, opened 2023-09-19
13809 fix ConfigRevision edit if custom validators

### LINKED ISSUE #13809: Config revision view error when using custom validators
### NetBox version

v3.6.1

### Python version

3.8

### Steps to Reproduce

1. Create Custom Validator
EX: 

```

class cvReservationMAC(CustomValidator):
    def validate(self, instance):
        if instance.status == 'dhcp' and not instance.cf["MAC_Address"]:
            self.fail("DHCP Reservations must have Mac Address!", field='cf_MAC_Address')

class cvReservationDNSName(CustomValidator):
    def validate(self, instance):
        if instance.status == 'dhcp' and not instance.dns_name:
            self.fail("DHCP Reservations must have DNS Name!", field='dns_name')

CUSTOM_VALIDATORS = {
    'ipam.ipaddress': (
        cvReservationMAC(),
        cvReservationDNSName(),
    )
}
```
3. Open menu "Admin >> Current Config" Click on Edit

### Expected Behavior

Edit config context appears

### Observed Behavior

image

arthanson:
### Fixes: #13809 

Disables the CUSTOM_VALIDATORS field if field is not JSON serializable (because it uses custom functions).

### DIFF
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -532,13 +532,11 @@ class ConfigRevisionForm(BootstrapMixin, forms.ModelForm, metaclass=ConfigFormMe
                     value = ', '.join(value)
                 self.fields[param.name].initial = value
 
-                if param.name == 'CUSTOM_VALIDATORS':
-                    # custom validators can be either dict or functions
-                    try:
-                        json.dumps(value)
-                    except TypeError:
-                        self.fields[param.name].initial = None
-                        self.fields[param.name].disabled = True
+                try:
+                    json.dumps(value)
+                except TypeError:
+                    self.fields[param.name].initial = None
+                    self.fields[param.name].disabled = True
 
             if is_static:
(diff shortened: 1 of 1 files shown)


netbox/extras/forms/model_forms.py
```diff
@@ -531,6 +531,15 @@ def __init__(self, *args, **kwargs):
                 if type(value) in (tuple, list):
                     value = ', '.join(value)
                 self.fields[param.name].initial = value
+
+                if param.name == 'CUSTOM_VALIDATORS':
```
jeremystretch, line 535: Rather than focusing on `CUSTOM_VALIDATORS` specifically, this would be a good check to implement for _all_ parameters.

jeremystretch (changes requested):

jeremystretch (approved):

jeremystretch:
Thanks @arthanson!

PR #13826, opened 2023-09-19
Fixes #11617: Check for invalid CSV headers during bulk import

jeremystretch:
### Fixes: #11617

This is being submitted as a lightweight alternative to PR #12013. I believe it resolves the root issue without significantly refactoring the bulk import logic.


arthanson (changes requested):
Should we add a test case.

arthanson (approved):

PR #13827, opened 2023-09-19
13653 darken code color to work in light and dark modes

### LINKED ISSUE #13653: Some text is unreadable in warning messages when dark mode is active
### NetBox version

v3.6.0

### Python version

3.11

### Steps to Reproduce

1. Add a custom script to NetBox
2. Delete the script file from the file system
3. Open the "Custom Scripts" list view

### Expected Behavior

A readable warning or error message is displayed

### Observed Behavior

Parts of the warning message (in this case, the file name) is displayed in a light gray font on yellow background, which makes it almost impossible to read.

Screenshot 2023-09-01 at 10 19 38

arthanson:
### Fixes: #13653 

Darkens up the warning color for code blocks so they work in both light and dark modes, also added spacing around the code block in the warning message as before it was scrunched together.
Monosnap Scripts | NetBox 2023-09-19 13-46-45
Monosnap Scripts | NetBox 2023-09-19 13-47-03


netbox/templates/extras/script_list.html
```diff
@@ -42,7 +42,7 @@ <h5 class="card-header" id="module{{ module.pk }}">
             <div class="alert alert-warning d-flex align-items-center" role="alert">
               <i class="mdi mdi-alert"></i>
               {% blocktrans trimmed with file_path=module.full_path %}
-                Script file at <code>{{ file_path }}</code> could not be loaded.
+                Script file at &nbsp;<code>{{ file_path }}</code>&nbsp; could not be loaded.
```
jeremystretch, line 45: If additional horizontal space is needed, we should tweak `padding` for the `<code>` element instead.

jeremystretch (changes requested):

netbox/templates/extras/script_list.html
arthanson: It is just here that the spacing is off, I looked at other <code block usages and the spacing looks correct.  I thought the &nbsp; was cleaner solution then a specific css adjustment for .alert code, but can change.

arthanson: changed to use mx-1 spacing.

PR #13828, opened 2023-09-19
12685 use markdown for custom fields added to form

### LINKED ISSUE #12685: Markdown in custom field descriptions not rendered in object edit view
### NetBox version

v3.5.2

### Python version

3.8

### Steps to Reproduce

1. Create a custom field with a description that uses markdown
2. Assign that custom field to an object type
3. Create an object of that type
4. Edit that object

### Expected Behavior

Description below custom field should render markdown.

### Observed Behavior

Markdown not rendered.

I have a PR ready for this one.

arthanson:
### Fixes: #12685 

Renders custom field description (for help text) using markdown for forms.  Did the escape first so it catches html and just process markdown afterwards.

### DIFF
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -76,7 +76,8 @@ class CustomFieldForm(BootstrapMixin, forms.ModelForm):
             'type': _(
                 "The type of data stored in this field. For object/multi-object fields, select the related object "
                 "type below."
-            )
+            ),
+            'description': _("This will be displayed as help text for the form field. Markdown is supported.")
         }
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


kkthxbye-code (commented):
This does not match the markdown supported elsewhere in netbox. Easiest way to demonstrate is to set the description to `<strong>test</strong>`.

How it's displayed in different parts of the UI:

Custom field object view:
image

Target object view:
image

Target object edit view:
image

All other places we render markdown is done with `from utilities.templatetags.builtins.filters import render_markdown`.

When setting the description to `**test**` the custom field object view and the target object edit view works, but not the target object view:

Custom field object view:
image

Target object view:
image

Target object edit view:
image

The cause is that the PR doesn't modify this:

https://github.com/netbox-community/netbox/blob/69215c411b76082412607a0674132ebe4961a95a/netbox/templates/inc/panels/custom_fields.html#L19-L22

PR #13857, opened 2023-09-21
13839 change color and spacing on alert code block

### LINKED ISSUE #13839: Regression in Import form Field Options color
### NetBox version

v3.6.2

### Python version

3.8

### Steps to Reproduce

1. Navigate to any Import form
2. Observe the low contrast on the Field and Accessor columns of the Field Options table

### Expected Behavior

The Field and Accessor columns of the Field Options table of an Import form should have sufficient contrast.

### Observed Behavior

#13827 was introduced in v3.6.2 to fix #13653, but as a side effect, it changed the color of all `<code>` tags, which caused this regression.

arthanson:
### Fixes: #13839 

Revert previous change to global code block, made more specific scss to just affect alert code block setting color and left-right padding.


netbox/project-static/styles/netbox.scss
```diff
@@ -167,6 +167,13 @@ table td > .progress {
   }
 }
 
+.alert {
+  code {
+    color: $gray-600;
+    margin: 0 1em 0 1em;
```
jeremystretch, line 173: IMO we should keep the `mx-1` CSS class in place on the element for this, just to make it easier to adjust if need be in the future.

jeremystretch (changes requested):

PR #13869, opened 2023-09-22
Fix creating config template using rest api

### LINKED ISSUE #13506: Unable to provision a config-template with local data source via the REST API
### NetBox version

v3.6.1

### Python version

3.11.4

### Steps to Reproduce

1. Start from a fresh NetBox instance, configure an API token and configure these 2 environment variables:

```
TOKEN=mynetboxtoken
NETBOX_URL=http://my.netbox.url
```

Please replace with values relevant to your NetBox instance.

2. Create `/etc/netbox/templates` directory on the NetBox instance and create a trivial Jinja2 template to the NetBox instance in it:

```
mkdir -p /etc/netbox/templates
echo 'This is a simple Jinja2 template.' > /etc/netbox/templates/basic.j2
```

4. Provision a new local data source using the REST API:

```
curl -X POST \
-H "Authorization: Token $TOKEN" \
$NETBOX_URL/api/core/data-sources/ \
--data '{
    "name": "templates",
    "type": "local",
    "source_url": "file:///etc/netbox/templates",
    "enabled": true
}'
```

Make sure it is correctly created and note down its id:

```
DATASOURCE_ID=1
```

Please replace with the id of the created object.

5. Sync this new local data source using the REST API:

```
curl -X POST \
-H "Authorization: Token $TOKEN" \
$NETBOX_URL/api/core/data-sources/$DATASOURCE_ID/sync/ \
```

6. Provision a new config template based on the `basic.j2` Jinja2 template in this local data source using the REST API:

```
curl -X POST \
-H "Authorization: Token $TOKEN" \
$NETBOX_URL/api/extras/config-templates/ \
--data '{
    "name": "basic",
    "template_code": "Will be overwritten after data_file is synced",
    "data_source": {"name": "templates"},
    "data_file": {"path": "basic.j2"}
}'
```

Make sure it is correctly created and note down its id:

```
CONFIGTEMPLATE_ID=1
```

Please replace with the id of the created object.

7. Sync this new config template:

```
curl -X POST \
-H "Authorization: Token $TOKEN" \
$NETBOX_URL/api/extras/config-templates/$CONFIGTEMPLATE_ID/sync/
```

Now, the config template can be used on a device or device-role.

### Expected Behavior

At step 6, I expect `data_file` to be correctly populated in the JSON data of the POST response. I should get:

```
{
    ...
    "data_file": {
        ...
        "path": "basic.j2"
    },
    ...
}
```

### Observed Behavior

At step 6, the JSON data I'm really receiving in the POST response is:

```
{
    ...
    "data_file": null,
    ...
}
```

desnoe:


netbox/extras/api/serializers.py
```diff
@@ -454,7 +454,7 @@ class ConfigTemplateSerializer(TaggableModelSerializer, ValidatedModelSerializer
         required=False
     )
     data_file = NestedDataFileSerializer(
-        read_only=True
+        read_only=False
```
abhi1693, line 457: It usually defaults to false is not specified, so this can be removed entirely

abhi1693 (changes requested):

netbox/extras/api/serializers.py
desnoe: I think I should put `required=False`, because `data_file` parameter is not mandatory, as for `data_source` just above ? I suppose it has been set because default value for `required` field is True ?

desnoe: That's confirmed by the CI:
```
AssertionError: 400 != 201 : Expected HTTP status 201; received 400: {'data_file': [ErrorDetail(string='This field is required.', code='required')]}
```
I'll send an update soon.

jeremystretch:
Thanks @desnoe!

PR #13874, opened 2023-09-23
Refactor row coloring logic and simplify mark planned/connected toggle implementation

### LINKED ISSUE #13712: "Mark planned" doesn't immediately change row highlight colour

pv2b:
### Fixes: #13712 and #13806.

As I mentioned in the discussion of #13712, the root cause of the fact that this bug happens in the first place is that the current approach in Netbox to toggling cable state implements styling logic for interface rows in two distinct places, one which is server-side by applying CSS classes to table rows depending on different criteria, and again on the client-side by updating the DOM after the fact.

These two implementation had slipped out of sync at some point, which caused the bug to appear.

To fix the root problem, which is a duplication of logic, I have refactored the row coloring logic to fully happen in CSS. This makes sense, because setting colors of table rows is styling, so it makes sense to do that in CSS. To do this, the following enhancements needed to be made:

- Adding data attributes to table rows representing virtual interface type, mark connected, as well as cable state. This is so that we can define CSS attribute selectors for said interface connection states.
- Creating CSS variables for colors for choices. This is so that these colors can then be referenced by name in the runtime-generated CSS for styling rows.
- Dynamically generating stylesheets based on available choices. This has to happen at runtime because connection states are potentially user-customizable.

This allowed the code implementing the toggle button for cable state to be substantially simplified. Instead of generating a single button at template render time depending on the current cable state, both buttons are now just always generated, and hidden/shown depending on the current state of the cable (as per the attribute on the table row). Visually the effect is the same, and code-wise, it's a lot simpler. The only element the Typescript current twiddles in the DOM is now the data-cable-status attribute of the table row, the CSS does the rest, including showing the correct buttons and coloring the rows.

Now, this pull request is likely *not* acceptable in its current state. A few issues I've already identified:

- Cable connection states are not only togglable in Device Interface Views, but in several other places. For consistency, this should be updated across the board. In fact, I've probably broken these buttons in other views in the current state of the PR.
- I'm not sure if I've added the new CSS styles in the right place. The "choice-based" colour styles by neccessity need to be generated at runtime and putting them in the dcim/device/interfaces.html template's head block was the easiest solution. I'm open for any opinions on whether these CSS styles need to be moved. For now I just did the simplest thing that could possibly work as a proof-of-concept, but ultimately this is a design decision.

Anyway, before I put in any more work into this PR to finish it, I'd like some feedback at least on the above points, and whether you agree with the general direction I'm going with this, because it turns out this was much more than a simple bugfix. If you're looking for a simpler "bugfix" type change that doesn't address the underlying issue, there's also my first stab at it, at #13807, but that's just likely to break again because it doesn't address the repetition of logic.


DanSheps:
@jeremystretch Marked you as review requested as peter would like some feedback before he pushes forward.

pv2b:
> peter would like some feedback before he pushes forward.

Thanks for raising this for me, but my name is Per, not Peter.

jeremystretch:
Any maintainer is capable of reviewing this; there is no need to ping me. However, I'll point out that this is still a draft PR. If it's ready for review, its conflicts with the base branch and all tests must pass, and it needs to be marked as such.

netbox/templates/dcim/device/interfaces.html
```diff
@@ -30,3 +30,29 @@
...
 {% endblock bulk_extra_controls %}
+
+{% block head %}
+    {{ block.super }}
+    <style>
+        {% for status, color in table.Meta.cable_status_styles %}
+            tr[data-cable-status={{ status }}] {
+                background-color: var(--nbx-color-{{ color }}-a15);
+            }
+        {% endfor %}
```
DanSheps, line 41: I don't think we need this, the row should not change depending on cable_status color beyond either green or blue.  Any other changes should be done in the cable form which do not need to be dynamically updated.

netbox/templates/dcim/device/interfaces.html
```diff
@@ -30,3 +30,29 @@
...
+        }
+        tr[data-enabled=disabled] {
+            background-color: var(--nbx-color-danger-a15);
+        }
+        tr[data-cable-status=connected] button.mark-installed {
+            display: none;
+        }
+        tr:not([data-cable-status=connected]) button.mark-planned {
+            display: none;
+        }
```
DanSheps, line 56: This should be in the main scss file and not in the template

netbox/dcim/tables/devices.py
```diff
@@ -670,11 +693,14 @@ class Meta(DeviceComponentTable.Meta):
...
         row_attrs = {
-            'class': get_interface_row_class,
             'data-name': lambda record: record.name,
             'data-enabled': get_interface_state_attribute,
+            'data-virtual': get_interface_virtual_attribute,
+            'data-mark-connected': get_interface_mark_connected_attribute,
+            'data-cable-status': get_interface_cable_status_attribute,
             'data-type': lambda record: record.type,
         }
+        cable_status_styles = [(slug, color) for slug, _, color in LinkStatusChoices.CHOICES]
```
DanSheps, line 703: LinkStatusChoices is a tuple and not user-customizable.  We can likely just add the appropriate settings in the SCSS itself.

netbox/dcim/tables/devices.py
```diff
@@ -670,11 +693,14 @@ class Meta(DeviceComponentTable.Meta):
             'cable', 'connection',
         )
         row_attrs = {
-            'class': get_interface_row_class,
             'data-name': lambda record: record.name,
             'data-enabled': get_interface_state_attribute,
+            'data-virtual': get_interface_virtual_attribute,
```
DanSheps, line 698:
```suggestion
            'data-virtual': 'true' if record.is_virtual else 'false'
```

netbox/dcim/tables/devices.py
```diff
@@ -69,6 +62,36 @@ def get_interface_state_attribute(record):
...
+def get_interface_cable_status_attribute(record):
+    """
+    Get interface enabled state as string to attach to <tr/> DOM element.
+    """
+    if record.cable:
+        return record.cable.status
+    else:
+        return ""
+
+
```
DanSheps, line 94:
```suggestion
```
I think the way you are going is a little "extra" here.

netbox/dcim/tables/devices.py
```diff
@@ -670,11 +693,14 @@ class Meta(DeviceComponentTable.Meta):
             'cable', 'connection',
         )
         row_attrs = {
-            'class': get_interface_row_class,
             'data-name': lambda record: record.name,
             'data-enabled': get_interface_state_attribute,
+            'data-virtual': get_interface_virtual_attribute,
+            'data-mark-connected': get_interface_mark_connected_attribute,
```
DanSheps, line 699:
```suggestion
            'data-mark-connected': 'true' if record.is_connected else 'false'
```

netbox/dcim/tables/devices.py
```diff
@@ -670,11 +693,14 @@ class Meta(DeviceComponentTable.Meta):
...
+            'data-cable-status': get_interface_cable_status_attribute,
```
DanSheps, line 700:
```suggestion
            'data-cable-status': 'true' if record.cable and record.cable.status else 'false'
```

DanSheps (changes requested):

netbox/dcim/tables/devices.py
pv2b: Okay, wow, yeah, if LinkStatusChoices is not user-customizable then a lot of the complexity in this patch is entirely useless.

I'll see if I can pare it down.

netbox/dcim/tables/devices.py
pv2b: I agree, this is a lot briefer and nicer. I was trying to match the code style in the rest of the method, but if you're prefer this more brief style, hey, nobody's paying me by the line, so why not.

netbox/templates/dcim/device/interfaces.html
pv2b: Putting it in the main scss was not possible with this approach since the CSS needed to happen at run-time from my mistaken belief that the link statuses were user-customizable.

But since we've agreed that they're not, this can definitely go into the scss instead.

netbox/dcim/tables/devices.py
DanSheps: Yes, I can understand it is sometimes hard to tell what is and isn't user customizable.  I always take a look at the .CHOICES attribute on the choices class.  If it is a tuple (`(,)`) it is not customizable

pv2b:
Looking back at this after a few months, another bug jumps out at me that I didn't notice, and it probably present both in this new implementation and the existing implementation - if you have a listing with both ends of the cable (e.g. both interfaces where both of them connect to each other) the background colour only gets updated on the interface you actually click, not the other one.

Just putting this here as a note, if it's an easy fix I might throw it in "for free".

pv2b:
I've adjusted this PR to align with the guidance provided by @DanSheps and I think it'll be closer to acceptable now.

As for what I wrote before:

> Cable connection states are not only togglable in Device Interface Views, but in several other places. For consistency, this should be updated across the board. In fact, I've probably broken these buttons in other views in the current state of the PR.

I'm not sure what I was thinking back when I wrote it, but I can't find any other places where interfaces are coloured this way so... I think I'm done? At least functionality-wise.

DanSheps:
Took a quick look and seems to be in-order.  I will pull it down when I can and do a more in-depth dive into it.

DanSheps:
Tested, works very well.

Thanks for the work @pv2b

PR #13892, opened 2023-09-25
13891 fix primary ip assignment if assigning ip

### LINKED ISSUE #13891: Can't assign primary IP when editing and assigning existing IP address
### NetBox version

v3.6.2

### Python version

3.8

### Steps to Reproduce

1. Go to an existing IP address
2. Assign the address to an interface
3. Check "Make this primary IP for device/VM"
4. click the Save button

image

Related to: https://github.com/netbox-community/netbox/issues/12953 - but this is for editing case, not create case.

### Expected Behavior

IP Address will save without error

### Observed Behavior

image

arthanson:
### Fixes: #13891 

Fixes case where you are marking an IP address as primary while also assigning it.


jeremystretch (approved):
Thanks @arthanson!

PR #13893, opened 2023-09-25
13130 dont allow reassigning ipaddress assigned object if primary ip

### LINKED ISSUE #13130: primary IP is set even when IP doesn't belong to the VM
### NetBox version

v3.5.5

### Python version

3.8

### Steps to Reproduce

When I am reassigning an IP to another virtual machine, the IP is still set as a primary IP for the initial virtual virtual machine.
I have two virtual machines, `ceph-mon-001`(network interface id: 40) and `ceph-mon-002` (network interface id: 92)

```
curl -H "Authorization: Token <TOKEN>" -H "Content-Type: application/json" -H "Accept: application/json; indent=4" -X PATCH -d '{"assigned_object_id": 40}' https://netbox-company.com/api/ipam/ip-addresses/735/
```

So in this case I'm reassigning IP id `735` (`10.88.88.176/24`) from the network id of ceph-mon-002 (`92`) to the network id of ceph-mon-001 (`40`).

At this point ceph-mon-002's only network interface has no ip at all. Nonetheless I'm still seeing the ip `10.88.88.176` as primary IP:
ceph-mon-002

Tested with the API:
```
curl -H "Authorization: Token <TOKEN>" -H "Content-Type: application/json" -H "Accept: application/json; indent=4" -X GET https://netbox-company.com/api/virtualization/virtual-machines/72/
```
```json
{
  "id": 735,
  "url": "https://netbox-company.com/api/ipam/ip-addresses/735/",
  "display": "10.88.88.176/24",
  "family": 4,
  "address": "10.88.88.176/24"
}
{
  "id": 735,
  "url": "https://netbox-company.com/api/ipam/ip-addresses/735/",
  "display": "10.88.88.176/24",
  "family": 4,
  "address": "10.88.88.176/24"
}
```

### Expected Behavior

I expect that the primary VM disappears from the initial virtual machine.

### Observed Behavior

It's still there :)

This is seems to be the same issue reported here https://github.com/netbox-community/netbox/issues/12953

Something really similar was reported in 2018 https://github.com/netbox-community/netbox/issues/2342
(related specifically to this: https://github.com/netbox-community/netbox/issues/2399)

So this keeps coming backup.

I'm using this in Ubuntu 22.04.2. Netbox is directly installed on the host. Postgresql 13.10 is installed.

arthanson:
### Fixes: #13130 

Don't allow reassigning IP address if it is a primary IP for a different object.

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -1,7 +1,6 @@
 import netaddr
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
-
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models import F
@@ -786,7 +785,7 @@ class IPAddress(PrimaryModel):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        # Save a copy of u_height for validation in clean()
+        # Denote the original assigned object (if any) for validation in clean()
         self._original_assigned_object_id = self.__dict__.get('assigned_object_id')
         self._original_assigned_object_type_id = self.__dict__.get('assigned_object_type_id')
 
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @arthanson!

PR #13911, opened 2023-09-27
Service Bulk Import Form with IP Address Added

### LINKED ISSUE #13022: Service Bulk Import Form with IP Address
### NetBox version

v3.4.0

### Feature type

New functionality

### Proposed functionality

Add IP Address to the Bulk Import Form of Service

### Use case

The import function of service would be complete

### Database changes

_No response_

### External dependencies

_No response_

yash-pal1:


jeremystretch (changes requested):
The import form needs to validate that any specified IP addresses belong to the device/VM to which the service is being assigned.

jeremystretch:
I've cleaned up the validation logic in the above changes and extended the bulk import test in a new PR, #14230.

PR #13935, opened 2023-09-29
Attached a clonable field multi selection, bug fixes

### LINKED ISSUE #13909: Attaching a cloneable multi selection custom field with custom field choices breaks the Clone function
### NetBox version

v3.6.3

### Python version

3.9

### Steps to Reproduce

1. Create a custom field choices object without base choices selected and with at least 1 extra choice defined.
2. Create a custom field of the type "Multiple Selection" and associate with a content type (I used Virtual Machine, but the problem also appears for all other object types I've tested with such as IP addresses). 
3. Use the previously created custom field choices object as "Choice Set".
4. Select the "Is cloneable" checkbox.
5. Go to any object of the associated content type and try to clone it. It doesn't matter if the custom field has been provided with a value for that object or not.
6. Cloning will fail with the message `<class 'TypeError'> 'in <string>' requires string as left operand, not NoneType`
7. Disable the "Is cloneable" checkbox in the custom field that was created in step 2.
8. Cloning the object will now work (but obviously without cloning the custom field).

### Expected Behavior

I would expect the object to be cloneable when I use custom fields of the multiple selection type with the cloneable option enabled.

### Observed Behavior

`<class 'TypeError'>'in <string>' requires string as left operand, not NoneType` occurs whenever trying to clone any object that uses a multiple selection custom field where "is cloneable" is enabled.

This worked fine in 3.5.9, so it seems it is probably related to the new custom field choices object introduced in 3.6.0.

The workaround for now is to disable "is cloneable" on all multiple selection custom fields until the cause of the error is fixed.

yash-pal1:

### DIFF
--- a/netbox/utilities/forms/fields/dynamic.py
+++ b/netbox/utilities/forms/fields/dynamic.py
@@ -43,7 +43,7 @@ class DynamicMultipleChoiceField(forms.MultipleChoiceField):
 
         if data is not None:
             self.choices = [
-                choice for choice in self.choices if choice[0] in data
+                choice for choice in self.choices if isinstance(self.choices, str) if choice[0] in data
             ]
 
         return bound_field


netbox/utilities/forms/fields/dynamic.py
```diff
@@ -43,7 +43,7 @@ def get_bound_field(self, form, field_name):
 
         if data is not None:
             self.choices = [
-                choice for choice in self.choices if choice[0] in data
+                choice for choice in self.choices if isinstance(self.choices, str) if choice[0] in data
```
abhi1693, line 46:
```suggestion
                choice for choice in self.choices if isinstance(self.choices, str) and choice[0] in data
```

I think `and` should suffice here

jeremystretch (changes requested):
While this avoids the exception, it doesn't actually fix the cloning functionality. The new form should be pre-populated with the custom field value(s) from the object being cloned.

jeremystretch:
@yash-pal1 are you still interested in working on this?

yash-pal1:
> @yash-pal1 are you still interested in working on this?



> @yash-pal1 are you still interested in working on this?

trying to fix this, so sorry for not active on this issue

jeremystretch:
Closing for inactivity. Please see #14612 for a (very similar) alternative fix.

PR #13942, opened 2023-10-02
12336 make region API calls atomic

### LINKED ISSUE #12336: Parallel Region Create API calls can produce top-level regions with the same `tree_id`
### NetBox version

v3.4.8

### Python version

3.8

### Steps to Reproduce

Ref: https://github.com/e-breuninger/terraform-provider-netbox/issues/385

Using a tool which will produce parallel API calls to NetBox (e.g. Terraform), create two or more top-level Region objects with children, via the `/dcim/regions` endpoint.

In Terraform, this might look like this:

```
resource "netbox_region" "uk" {
  name = "United Kingdom"
}

resource "netbox_region" "ireland" {
  name = "Ireland"
}

resource "netbox_region" "london" {
  name             = "London"
  parent_region_id = netbox_region.uk.id
}

resource "netbox_region" "gatwick" {
  name             = "Gatwick"
  parent_region_id = netbox_region.uk.id
}



resource "netbox_region" "northeurope" {
  name             = "Azure North Europe"
  parent_region_id = netbox_region.ireland.id
}

resource "netbox_region" "uksouth" {
  name             = "Azure UK South"
  parent_region_id = netbox_region.london.id
}
```

### Expected Behavior

The Regions are created with the correct tree structure in the NetBox UI

### Observed Behavior

The top-level regions can sometimes be created with the same `tree_id` in the `dcim_region` table. This causes some weird UI issues whereby all the child regions will appear under one of the top-level regions, despite it not being their parent:

image

image

If you follow the child region links, the Parent IDs are set correctly.

To work around this, we can ensure that the top-level regions are created in serial so that they never receive the same `tree_id` value.

arthanson:
### Fixes: #12336 

added mixin to put pglocks (for MPTT) around view calls to create, update, destroy.  used the model name for the lock key (instead of a common 'mptt' one) so parallel calls could be made to different mptt objects (say sitegroups and regions) for better performance.

### DIFF
--- a/netbox/netbox/api/viewsets/__init__.py
+++ b/netbox/netbox/api/viewsets/__init__.py
@@ -168,13 +168,13 @@ class MPTTLockedMixin(GenericViewSet):
     """
 
     def create(self, request, *args, **kwargs):
-        with advisory_lock(ADVISORY_LOCK_KEYS[self.queryset.model._meta.verbose_name.lower()]):
+        with advisory_lock(ADVISORY_LOCK_KEYS[self.queryset.model._meta.verbose_name.lower().replace(" ", "")]):
             return super().create(request, *args, **kwargs)
 
     def update(self, request, *args, **kwargs):
-        with advisory_lock(ADVISORY_LOCK_KEYS[self.queryset.model._meta.verbose_name.lower()]):
+        with advisory_lock(ADVISORY_LOCK_KEYS[self.queryset.model._meta.verbose_name.lower().replace(" ", "")]):
             return super().update(request, *args, **kwargs)
 
     def destroy(self, request, *args, **kwargs):
-        with advisory_lock(ADVISORY_LOCK_KEYS[self.queryset.model._meta.verbose_name.lower()]):
+        with advisory_lock(ADVISORY_LOCK_KEYS[self.queryset.model._meta.verbose_name.lower().replace(" ", "")]):
             return super().destroy(request, *args, **kwargs)
--- a/netbox/netbox/api/viewsets/__init__.py
+++ b/netbox/netbox/api/viewsets/__init__.py
@@ -161,20 +161,20 @@ class NetBoxModelViewSet(
         return super().perform_destroy(instance)
 
 
-class MPTTLockedMixin(GenericViewSet):
+class MPTTLockedMixin:
     """
     Puts pglock on objects that derive from MPTTModel for parallel API calling.
     Note: If adding this to a view, must add the model name to ADVISORY_LOCK_KEYS
     """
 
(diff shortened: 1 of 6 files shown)


netbox/netbox/api/viewsets/__init__.py
```diff
@@ -157,3 +159,22 @@ def perform_destroy(self, instance):
...
+
+
+class MPTTLockedMixin(GenericViewSet):
+    """
+    Puts pglock on objects that derive from MPTTModel for parallel API calling.
+    Note: If adding this to a view, must add the model name to ADVISORY_LOCK_KEYS
+    """
+
+    def create(self, request, *args, **kwargs):
+        with advisory_lock(ADVISORY_LOCK_KEYS[self.queryset.model._meta.verbose_name.lower().replace(" ", "")]):
```
jeremystretch, line 171: The `model_name` meta attribute works for this:
```suggestion
        with advisory_lock(ADVISORY_LOCK_KEYS[self.queryset.model._meta.model_name]):
```

netbox/netbox/api/viewsets/__init__.py
```diff
@@ -157,3 +159,22 @@ def perform_destroy(self, instance):
         logger.info(f"Deleting {model._meta.verbose_name} {instance} (PK: {instance.pk})")
 
         return super().perform_destroy(instance)
+
+
+class MPTTLockedMixin(GenericViewSet):
```
jeremystretch, line 164: Does this need to inherit from `GenericViewSet`? Seems like it would be fine as a base class.

netbox/netbox/constants.py
```diff
@@ -15,4 +15,14 @@
     'available-ips': 100200,
     'available-vlans': 100300,
     'available-asns': 100400,
+
+    # MPTT locks
+    'region': 100500,
```
jeremystretch, line 20: Should we carve out a separate number space for MPTT locks? We might want to add more "available object" endpoints in the future and it would be cleaner to keep those numbers consistent.

jeremystretch (changes requested):

netbox/netbox/constants.py
arthanson: That's why I put them all under 1005xx, if we add more available they can be at 100600, 100700, etc.. We might add more MPTT locks in the future so this allows them to easily co-exist.

jeremystretch: Right; I'm saying maybe we should reserve 100xxx for views and use e.g. 101xxx for MPTT models. It's not  big deal, just trying to plan for future additions.

jeremystretch (approved):
Thanks @arthanson!

PR #13943, opened 2023-10-02
13815 document view permissions for scripts

### LINKED ISSUE #13815: No Scripts Found for non superusers
### NetBox version

v3.6.1

### Python version

3.9

### Steps to Reproduce

1. Upload script in the scripts repository
2. Login as Active user (not superuser) with "extras | script and extras | script module" actions set to run, view.
3. Select Customization > Scripts
image
4. with a superuser we can see the scripts
image

### Expected Behavior

Be able to see the scripts stored on the server

### Observed Behavior

Page reports "No Scripts Found"

This issue was reported in a previous version (v3.5.1) but it seems it was fixed enabling extras script module (see issue #12516)

arthanson:
### Fixes: #13815 

Adds documentation for permission needed to view and run scripts as it requires several non-obvious permissions.

### DIFF
--- a/docs/customization/custom-scripts.md
+++ b/docs/customization/custom-scripts.md
@@ -288,7 +288,7 @@ An IPv4 or IPv6 network with a mask. Returns a `netaddr.IPNetwork` object. Two a
 ## Running Custom Scripts
 
 !!! note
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):
We should probably add the same note for reports as well, as both extend ManagedFile. @arthanson would you mind doing that?

Issue #13944, opened 2023-10-02
Bug in API serialization of report details


Haeki:
### NetBox version

v3.6.3

### Python version

3.11

### Steps to Reproduce

1. Access a report directly using the API for example `/api/extras/reports/example_report.DeviceConnectionsReport/`
2. In the result some values are not serialized correctly.

### Expected Behavior

The fields should be serialized correctly and results should be found.

### Observed Behavior

The name and display fields have a property object that got not serialized.
![image](https://github.com/netbox-community/netbox/assets/6272118/a1742c61-7dfe-4778-aaac-9871dd3a7b2c)
Also the results could not be found

PR #13945, opened 2023-10-02
Fix Bug in report serialization 

Haeki:
### Fixes: #[13944](https://github.com/netbox-community/netbox/issues/13944)

The name field should be `@classproperty` so that it can be serialized correctly.

Im not sure if there are other fields that should be `@classproperty` instead of `@property` but this change is enough to fix the Bug i encountered.

### DIFF
--- a/netbox/extras/reports.py
+++ b/netbox/extras/reports.py
@@ -120,7 +120,7 @@ class Report(object):
     def full_name(self):
         return f'{self.module}.{self.class_name}'
 
-    @property
+    @classproperty
     def name(self):
         """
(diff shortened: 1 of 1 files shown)


jeremystretch:
Hi @Haeki, thank you for working on this, however it seems to have been resolved by @arthanson in PR #13958. While your approach is valid, the alternative solution of instantiating the report prior to serialization is more consistent with how reports are handled under the UI views, and avoids a potentially breaking change to the Report class.

Haeki:
No problem. I'm Happy this got resolved so fast.
Thank you for all the work.

PR #13946, opened 2023-10-02
12328 update GFK object in clean

### LINKED ISSUE #12328: Updating a GFK field via API does not return the new value
### NetBox version

v3.4.8

### Python version

3.10

### Steps to Reproduce

1. Create IP-Address (do not assign object)
2. Create Object to assign (e.g. FHRP Group
3. `curl -H "Authorization: Token $NETBOX_TOKEN" -H "Content-type: application/json" $NETBOX_SCHEME://$NETBOX_URL/api/ipam/ip-addresses/1/ -X PATCH --data '{"assigned_object_type": "ipam.fhrpgroup", "assigned_object_id":1}'`

### Expected Behavior

`assigned_object` in the response is not null.

### Observed Behavior

`assigned_object` is null:
```json
{
    "id": 1,
    "url": "http://localhost:8000/api/ipam/ip-addresses/1/",
    "display": "127.0.0.1/32",
    "family": {
        "value": 4,
        "label": "IPv4"
    },
    "address": "127.0.0.1/32",
    "vrf": null,
    "tenant": null,
    "status": {
        "value": "active",
        "label": "Active"
    },
    "role": null,
    "assigned_object_type": "ipam.fhrpgroup",
    "assigned_object_id": 1,
    "assigned_object": null,
    "nat_inside": null,
    "nat_outside": [],
    "dns_name": "",
    "description": "",
    "comments": "",
    "tags": [],
    "custom_fields": {},
    "created": "2023-04-23T09:32:24.766755Z",
    "last_updated": "2023-04-23T09:35:09.908879Z"
}
```

This is only the case if the assigned_object was not assigned before the request. Changing or removing the assigned object gives the correct response.

arthanson:
### Fixes: #12328 

Potentially a bug in DRF, for the GFK field the _id and _type are set correctly but the content_object isn't set jut on return for a creation on REST API calls.  We are already going through all the GFK fields in the clean method so this just adds the actual setting of the content_object field.  A very slight performance hit changing from checking existence to actually getting the object.  Little bit of a hack to work-around the issue, but I can't think of any side-effects this should have.

### DIFF
--- a/netbox/netbox/models/__init__.py
+++ b/netbox/netbox/models/__init__.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.validators import ValidationError
 from django.db import models
 from django.utils.translation import gettext_lazy as _
--- a/netbox/netbox/models/__init__.py
+++ b/netbox/netbox/models/__init__.py
@@ -85,10 +85,15 @@ class NetBoxModel(NetBoxFeatureSet, models.Model):
 
                 if ct_value and fk_value:
                     klass = getattr(self, field.ct_field).model_class()
-                    if not klass.objects.filter(pk=fk_value).exists():
+                    try:
+                        obj = klass.objects.get(pk=fk_value)
+                    except ObjectDoesNotExist:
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @arthanson!

Issue #10769, opened 2022-10-27
Empty cable objects after deleting last termination


kr3ator:
### NetBox version

v3.3.5

### Python version

3.10

### Steps to Reproduce

1. Create two devices with interfaces.
2. Create a cable between the devices.
3. Delete device A.
4. Either delete device B or the interface of device B that was used in the Cable.

### Expected Behavior

The Cable object should be deleted after deleting the last termination.

### Observed Behavior

The Cable object is not deleted and has no terminations in it.

fggec:
Same here.

jeremystretch:
> The Cable object should be deleted after deleting the last termination.

I think our current sticking point is deciding whether or not the above assertion is true. I don't have a strong opinion either way, but the _current_ implementation permits cables to exist with no terminations at either end. Whether this was by design or by accident I can't recall. :laughing:

CADbloke:
It is physically possible for a cable to exist without terminations so it should probably be possible too in the documentation for that cable. imho.

fggec:
Yes but then you have to reference it somewhere in the Doku. I see 3 possible solutions for this:
1. cables can exist without a termination. -> Reference in Doku 
2. cables cannot exist without a termination. -> should be deleted.
3. let the users choice if 1 or 2.  this could be a separate checkbox or separate window. When you delete a device the user can select every cable that should be deleted or not with checkboxes maybe. Also for single cable if it's disconnected and has only one termination on one device left.

jeremystretch:
We discussed this in today's maintainers meeting, and the consensus was to delete a cable upon the deletion of its last termination. In the UI, this will first prompt the user to confirm the cable's deletion.

CADbloke:
> We discussed this in today's maintainers meeting, and the consensus was to delete a cable upon the deletion of its last termination. In the UI, this will first prompt the user to confirm the cable's deletion.

What will the default behaviour for the API be? Alternatively, can we specify they behaviour in the API call?

arthanson:
Thinking for the API case it should probably maintain default of not deleting the cable, but provide an option "delete_if_unterminated" or something on the termination delete API.

PaulWestphal:
I'm getting an AtributeError when trying to edit a cable after deleting all terminations on one side:


```
<class 'AttributeError'>

'str' object has no attribute 'label'

Python version: 3.10.6
NetBox version: 3.4.6

```

serge-deiss-obt:
@PaulWestphal : 
[...]

PR #13949, opened 2023-10-02
13972 allow filtering of cables if have terminations

### LINKED ISSUE #13972: Allow cables to be filtered if they have no terminations
### NetBox version

v3.6.3

### Feature type

New functionality

### Proposed functionality

Allow cables to be filtered if there are no terminations.

### Use case

As stated in this issue: https://github.com/netbox-community/netbox/issues/10769#issuecomment-1517495764 there is currently no way to filter for orphaned cables.  The orphaned cables can be created by deleting the devices they are connected to, but then there is no good way to find those cables.  Even if functionality is added to automatically remove cables when the last termination is removed - it still means that there can be older cables that were orphaned before these changes.

### Database changes

None.

### External dependencies

None.

arthanson:
### Fixes: #13972  

Changed from bug #10769 to resolving FR #13972 - this allows filtering of unterminated cables (either A, B or both).  Even if #10769 is done to remove cables when the last termination is deleted, this still leaves unterminated cables either from before the functionality or if the deletion is done via the API, hence the FR and this PR.

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1,5 +1,6 @@
 import django_filters
 from django.contrib.auth import get_user_model
+from django.db.models import Q
 from django.utils.translation import gettext as _
 
 from extras.filtersets import LocalConfigContextFilterSet
@@ -1745,13 +1746,9 @@ class CableFilterSet(TenancyFilterSet, NetBoxModelFilterSet):
         method='filter_by_cable_end_b',
         field_name='terminations__termination_id'
     )
-    has_a_terminations = django_filters.BooleanFilter(
-        method='_has_a_terminations',
-        label=_('Has a terminations'),
-    )
-    has_b_terminations = django_filters.BooleanFilter(
-        method='_has_b_terminations',
-        label=_('Has b terminations'),
+    unterminated = django_filters.BooleanFilter(
+        method='_unterminated',
+        label=_('Unterminated'),
     )
     type = django_filters.MultipleChoiceFilter(
         choices=CableTypeChoices
@@ -1820,17 +1817,12 @@ class CableFilterSet(TenancyFilterSet, NetBoxModelFilterSet):
         # Filter by termination id and cable_end type
         return self.filter_by_cable_end(queryset, name, value, CableEndChoices.SIDE_B)
 
-    def _has_a_terminations(self, queryset, name, value):
+    def _unterminated(self, queryset, name, value):
         if value:
-            return queryset.filter(terminations__cable_end=CableEndChoices.SIDE_A)
+            terminated_ids = queryset.filter(terminations__cable_end=CableEndChoices.SIDE_A).filter(terminations__cable_end=CableEndChoices.SIDE_B).values("id")
+            return queryset.exclude(id__in=terminated_ids)
         else:
-            return queryset.exclude(terminations__cable_end=CableEndChoices.SIDE_A)
-
-    def _has_b_terminations(self, queryset, name, value):
-        if value:
-            return queryset.filter(terminations__cable_end=CableEndChoices.SIDE_B)
-        else:
-            return queryset.exclude(terminations__cable_end=CableEndChoices.SIDE_B)
+            return queryset.filter(terminations__cable_end=CableEndChoices.SIDE_A).filter(terminations__cable_end=CableEndChoices.SIDE_B)
 
 
 class CableTerminationFilterSet(BaseFilterSet):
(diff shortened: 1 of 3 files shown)


jeremystretch:
@arthanson while the ability to filter for terminations can be useful, I don't think this resolves #10769. The consensus there seemed to be to automatically delete the cables, so if there's justification not to, we should continue the discussion there.

jeremystretch:
(If you still want to proceed with adding these filters, please open a separate FR for the work.)

arthanson:
@jeremystretch added to the discussion for #10769 and changed this to be the fix for the new FR.

jeremystretch:
Does it make sense to introduce separate filters for A & B terminations, as opposed to a single `has_terminations` filter? I'm having a difficult time imagining a use case for separate filters, since the model's stance is that the A and B designations don't imply and meaning.

arthanson:
@jeremystretch changed to unterminated as per discussion.

jeremystretch:
@arthanson thanks. Could you add a test for the filter as well please?

arthanson:
@jeremystretch test added.

netbox/dcim/forms/filtersets.py
```diff
@@ -908,7 +908,7 @@ class VirtualChassisFilterForm(TenancyFilterForm, NetBoxModelFilterSetForm):
 class CableFilterForm(TenancyFilterForm, NetBoxModelFilterSetForm):
     model = Cable
     fieldsets = (
-        (None, ('q', 'filter_id', 'tag')),
+        (None, ('q', 'filter_id', 'unterminated', 'tag')),
```
jeremystretch, line 911: The `unterminated` field should go under one of the model-specific sections below.

netbox/dcim/tests/test_filtersets.py
```diff
@@ -4289,14 +4290,15 @@ def setUpTestData(cls):
         Cable(a_terminations=[interfaces[9]], b_terminations=[interfaces[10]], label='Cable 5', type=CableTypeChoices.TYPE_CAT6, tenant=tenants[2], status=LinkStatusChoices.STATUS_PLANNED, color='e91e63', length=10, length_unit=CableLengthUnitChoices.UNIT_METER).save()
         Cable(a_terminations=[interfaces[11]], b_terminations=[interfaces[0]], label='Cable 6', type=CableTypeChoices.TYPE_CAT6, tenant=tenants[2], status=LinkStatusChoices.STATUS_PLANNED, color='e91e63', length=20, length_unit=CableLengthUnitChoices.UNIT_METER).save()
         Cable(a_terminations=[console_port], b_terminations=[console_server_port], label='Cable 7').save()
+        Cable(a_terminations=[interfaces[12]], label='Cable 8', type=CableTypeChoices.TYPE_CAT6, tenant=tenants[2], status=LinkStatusChoices.STATUS_DECOMMISSIONING, color='e91e63', length=20, length_unit=CableLengthUnitChoices.UNIT_METER).save()
```
jeremystretch, line 4293: Does this cable need its type, tenant, etc. set? Would be better to simplify if possible. It could also use a comment noting why the cable is necessary.

netbox/dcim/filtersets.py
```diff
@@ -1812,6 +1817,13 @@ def filter_by_cable_end_b(self, queryset, name, value):
         # Filter by termination id and cable_end type
         return self.filter_by_cable_end(queryset, name, value, CableEndChoices.SIDE_B)
 
+    def _unterminated(self, queryset, name, value):
+        if value:
+            terminated_ids = queryset.filter(terminations__cable_end=CableEndChoices.SIDE_A).filter(terminations__cable_end=CableEndChoices.SIDE_B).values("id")
+            return queryset.exclude(id__in=terminated_ids)
+        else:
+            return queryset.filter(terminations__cable_end=CableEndChoices.SIDE_A).filter(terminations__cable_end=CableEndChoices.SIDE_B)
```
jeremystretch, line 1825: These lines exceed the [120 character cut-off](https://docs.netbox.dev/en/stable/development/style-guide/#maximum-line-length-e501) and should be wrapped for better readability.

netbox/dcim/filtersets.py
```diff
@@ -1,5 +1,6 @@
 import django_filters
 from django.contrib.auth import get_user_model
+from django.db.models import Q
```
jeremystretch, line 3: Extraneous import statement?

jeremystretch (changes requested):

DanSheps:
Not sure if you looked at this at all, but this is the script I am using to find orphaned cables:


https://github.com/netbox-community/customizations/pull/92/files

PR #13958, opened 2023-10-03
13944 fix report detail api

### LINKED ISSUE #13944: Bug in API serialization of report details

arthanson:
### Fixes: #13944 

@jeremystretch can you take a second look, this does fix the issue and mirrors the code below and all tests run, just want to make sure there are no other unintentional side effects.

### DIFF
--- a/netbox/extras/reports.py
+++ b/netbox/extras/reports.py
@@ -23,7 +23,7 @@ logger = logging.getLogger(__name__)
 
 def get_module_and_report(module_name, report_name):
     module = ReportModule.objects.get(file_path=f'{module_name}.py')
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @arthanson!

PR #13963, opened 2023-10-03
Fixes #13064: Ensure unchecked checkboxes do not revert to original values upon HTMX form refresh

### LINKED ISSUE #13064: "Enabled" flag is reset during editing of an interface if "802.1Q Mode" is changed
### NetBox version

v3.5.4

### Python version

3.10

### Steps to Reproduce

1. edit an interface of hardware or a VM
2. change the state of the `Enabled` flag, it doesn't matter if it is checked or not, it just needs to be changed
3. change the value for the `802.1Q Mode`

### Expected Behavior

The set value (checked or unchecked) for the `Enabled` flag should be preserved during editing of an interface and not be reset to the current actual state when changing anything.

### Observed Behavior

The state for the `Enabled` flag is reset to the current actual state when `802.1Q Mode` is changed.

jeremystretch:
### Fixes: #13064

- Set `FORM_RENDERER` to `TemplatesSetting`, which allows overriding Django's stock form widget HTML templates
- Embed a hidden form element above each checkbox widget to ensure an empty field value is sent for unchecked checkboxes as part of the form data

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -355,6 +355,7 @@ INSTALLED_APPS = [
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'django.contrib.humanize',
+    'django.forms',
     'corsheaders',
     'debug_toolbar',
     'graphiql_debug_toolbar',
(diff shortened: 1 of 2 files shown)


netbox/netbox/settings.py
```diff
@@ -430,6 +431,9 @@ def _setting(name, default=None):
     },
 ]
 
+# This allows us to override Django's stock form widget templates
```
arthanson, line 434: How about adding a comment for why this here as it's not obvious.

arthanson (approved):

PR #13974, opened 2023-10-04
Adds selection custom field labels to UI

### LINKED ISSUE #13950: Render Custom Field Choices as label instead of value in UI
### NetBox version

v3.6.3

### Feature type

### Proposed functionality

When using the new Custom Field Choices, you are able to set a value and a label for each choice. When editing an object with a custom field that uses a Choice Set that has different labels and values like that, the label is shown in the dropdown, but the value is shown in the object view and in the corresponding table column. My proposal is to use the label value for display purposes everywhere in the UI.

### Use case

This proposal would bring the Custom Field Choices functionality in line with the native choice fields in how they are rendered in the UI and would provide a cleaner experience to the end users. I feel that this is also in alignment with the original feature request #13241.

### Database changes

None

### External dependencies

None

sleepinggenius2:
### Fixes: #13950

Adds a function to the `CustomField` model to get the choice label for a given value, defaulting to the passed value if one cannot be found. Updates the `customfield_value` template tag to leverage the new function, as I could not find a way to call a function with an argument in the template without registering a new tag anyway. Updates the `render` function of the `CustomFieldColumn` class to leverage the new function, so that table columns are rendered properly. Please let me know if there is anywhere else in the UI where custom field values are rendered that I might have missed.

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -232,6 +232,11 @@ class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
             return self.choice_set.choices
         return []
 
+    def get_choice_label(self, value):
+        if not hasattr(self, '_choice_map'):
+            self._choice_map = dict(self.choices)
+        return self._choice_map.get(value, value)
+
     def populate_initial_data(self, content_types):
         """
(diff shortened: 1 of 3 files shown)


jeremystretch (approved):
Thanks @sleepinggenius2!

PR #13984, opened 2023-10-05
13850 add requests to requirements

### LINKED ISSUE #13850: Missing dependency on requirements.txt
### NetBox version

develop

### Python version

3.8

### Steps to Reproduce

On commit d87235af2fc560731802ba475018b13a0c82af5c from 20230426 an import of the module requests was added into netbox/settings.py

It needs to be added into the general requirements.txt because it's missing on default installs

`ModuleNotFoundError: No module named 'requests'`

### Expected Behavior

Update should have taken care of this new dependency

### Observed Behavior

`ModuleNotFoundError: No module named 'requests'`

arthanson:
### Fixes: #13850 

Add requests to requirements, not specifically required but check the bug report, no bad effects from adding it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,6 +27,7 @@ netaddr==0.9.0
 Pillow==10.0.1
 psycopg[binary,pool]==3.1.11
 PyYAML==6.0.1
+requests==2.28.1
 sentry-sdk==1.31.0
 social-auth-app-django==5.3.0
 social-auth-core[openidconnect]==4.4.2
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -27,6 +27,7 @@ netaddr==0.9.0
 Pillow==10.0.1
 psycopg[binary,pool]==3.1.11
 PyYAML==6.0.1
+requests==2.28.1
```
jeremystretch, line 30: The current release is 2.31.0; any reason to stick with the older version?

arthanson: That is what I got doing pip install --upgrade -r requirements.txt so I was assuming it was bound-checked by another library.  Can update it further but I'm not sure if that would have any negative effects.

jeremystretch: That's odd; I have 2.31.0 installed. Either way, it will get bumped during the release process.

PR #14027, opened 2023-10-12
Fix #14026: Only get the needed amount of objects

### LINKED ISSUE #14026: Post to /api/ipam/prefixes/<int:pk>/available-ips/ is slow when the prefix is large
### NetBox version

v3.6.3

### Python version

3.11

### Steps to Reproduce

1. Create large prefix (e. g. 10.0.0.0/8
2. Request an IP from the create prefix with:
 curl -v -X POST -d '{}' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Authorization: Token REDACTED' http://localhost:8787/api/ipam/prefixes/$idOfPrefixFromStep1/available-ips/

### Expected Behavior

Get a fast response

### Observed Behavior

Response takes over 30 seconds

tobiasge:
### Fixes: #14026

Only get the needed amount of objects in the post method.

### DIFF
--- a/netbox/ipam/api/views.py
+++ b/netbox/ipam/api/views.py
@@ -266,6 +266,7 @@ class AvailableObjectsView(ObjectValidationMixin, APIView):
 
         # Normalize request data to a list of objects
         requested_objects = request.data if isinstance(request.data, list) else [request.data]
+        limit = len(requested_objects)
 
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @tobiasge!

PR #14028, opened 2023-10-12
14013 fix device role filter

### LINKED ISSUE #14013: InventoryItem filters - Device Roles are mapped to InventoryItem Roles
### NetBox version

v3.6.3

### Python version

3.10

### Steps to Reproduce

1. Go to Inventory Item list view (netbo-url/dcim/inventory-items/)
2. Click on Filters
3. Scroll to `Device` filter option
4. Select Device Role
    - `Inventory Item` **roles** are displayed instead of `Device` **roles**

image

The InventoryItem role can also be selected in the top of the filters in section **Attributes**

### Expected Behavior

Filtering Device roles at the `Device` section of filters

### Observed Behavior

Filtering InventoryItem roles at the `Device` section of filters

arthanson:
### Fixes: #14013 

InventoryItem had double role_id in the filter as it has one and the device has one defined, refactored the device one from using role_id to device_role_id to make it consistent with the other role filters "device_type_id", "device_id".

### DIFF
--- a/netbox/dcim/forms/filtersets.py
+++ b/netbox/dcim/forms/filtersets.py
@@ -109,7 +109,7 @@ class DeviceComponentFilterForm(NetBoxModelFilterSetForm):
         required=False,
         label=_('Device type')
     )
-    role_id = DynamicModelMultipleChoiceField(
+    device_role_id = DynamicModelMultipleChoiceField(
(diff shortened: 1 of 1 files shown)


DanSheps (approved):
LGTM, I was thinking of doing basically the same thing

PR #14029, opened 2023-10-12
Fixes: #14023 - Fixes bulk disconnecting with multiple components attached to the same cable

### LINKED ISSUE #14023: Error when deleting multiple cables
### NetBox version

v3.6.3

### Python version

3.8

### Steps to Reproduce

1. Registers wiring that would result in an L2 loop to the same device.
2. Select these two to delete the cable.
3. An error occurs and the cable is not deleted.
image
image
image
image

### Expected Behavior

Expect the cable to be removed.

### Observed Behavior

The error shown in the image has been identified.

DanSheps:
Fixes: #14023 - Fixes bulk disconnecting with multiple components attached to the same cable

* Change code to generate a list of cable ID's to then process for deletion.

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -122,13 +122,17 @@ class BulkDisconnectView(GetReturnURLMixin, ObjectPermissionRequiredMixin, View)
             if form.is_valid():
 
                 with transaction.atomic():
-
+                    cables = []
                     count = 0
                     for obj in self.queryset.filter(pk__in=form.cleaned_data['pk']):
                         if obj.cable is None:
                             continue
-                        obj.cable.delete()
+                        if obj.cable.pk not in cables:
+                            cables.append(obj.cable.pk)
                         count += 1
+
+                    if len(cables) > 0:
+                        Cable.objects.filter(pk__in=cables).delete()
 
                 messages.success(request, "Disconnected {} {}".format(
                     count, self.queryset.model._meta.verbose_name_plural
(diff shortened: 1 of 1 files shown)


netbox/dcim/views.py
```diff
@@ -122,14 +122,18 @@ def post(self, request):
             if form.is_valid():
 
                 with transaction.atomic():
-
+                    cables = []
```
jeremystretch, line 125: We can use a set (rather than a list) for this to automatically ignore duplicated IDs.
```suggestion
                    cables = set()
```

netbox/dcim/views.py
```diff
@@ -122,14 +122,18 @@ def post(self, request):
...
                     count = 0
                     for obj in self.queryset.filter(pk__in=form.cleaned_data['pk']):
                         if obj.cable is None:
                             continue
-                        obj.cable.delete()
+                        if obj.cable.pk not in cables:
+                            cables.append(obj.cable.pk)
```
jeremystretch, line 131:
```suggestion
                        cables.add(obj.cable.pk)
```

netbox/dcim/views.py
```diff
@@ -122,14 +122,18 @@ def post(self, request):
...
                         count += 1
 
+                    if len(cables) > 0:
+                        Cable.objects.filter(pk__in=cables).delete()
```
jeremystretch, line 135: Are we sure that calling `delete()` on the queryset (as opposed to each Cable object individually) is safe?

jeremystretch (changes requested):

netbox/dcim/views.py
DanSheps: You are correct, it might be better to call each cable as it skips the delete() method and does a bulk delete anyways.

```suggestion
                    for cable in Cable.objects.filter(pk__in=cables):
                        cable.delete()
```

Do you think this would be a reasonable change instead?

jeremystretch:
Thanks @DanSheps!

PR #14030, opened 2023-10-12
14025 fix script name checking

### LINKED ISSUE #14025: Exception when a script name is equal to the start of the name of another one
### NetBox version

v3.6.2

### Python version

3.11

### Steps to Reproduce

1. create two scripts for netbox with some sample code named `test.py` and `testscript.py`
2. upload them to NetBox scripts
3. try to open the test.py script

### Expected Behavior

The script page should be opened

### Observed Behavior

Error in website
netbox_error
and

Error in logging:
`2023-10-12 13:26:02,131 django.request ERROR: Internal Server Error: /extras/scripts/test/Test/
  File "/opt/netbox/dev/venv/lib64/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/dev/venv/lib64/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/dev/venv/lib64/python3.11/site-packages/django/views/generic/base.py", line 104, in view
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/dev/netbox/utilities/views.py", line 53, in dispatch
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/dev/venv/lib64/python3.11/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/dev/netbox/extras/views.py", line 1160, in get
    module = get_object_or_404(ScriptModule.objects.restrict(request.user), file_path__startswith=module)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/dev/venv/lib64/python3.11/site-packages/django/shortcuts.py", line 85, in get_object_or_404
    return queryset.get(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/dev/venv/lib64/python3.11/site-packages/django/db/models/query.py", line 640, in get
    raise self.model.MultipleObjectsReturned(
extras.models.scripts.ScriptModule.MultipleObjectsReturned: get() returned more than one ScriptModule -- it returned 2!
`

arthanson:
### Fixes: #14025 

@jeremystretch the module name checking is a bit verbose, not sure if there is a simpler way to do this.

The issue is if you have the script "test.py" module comes in as "test" and file_path is "test.py" which is why the startswith is there. If you remove startswith it won't match. I think it would be difficult at this point to change module to keep the full name with .py as I think it is stored that way. 

Ideally if we could remove a potential ".py" from file_path in the filter clause but I don't think there is a way to do that. it would also work to change how the scripts are stored to include the full name but that would require a data migration with potentially incomplete historical information.


netbox/extras/views.py
```diff
@@ -1157,7 +1157,7 @@ def get_required_permission(self):
         return 'extras.view_script'
 
     def get(self, request, module, name):
-        module = get_object_or_404(ScriptModule.objects.restrict(request.user), file_path__startswith=module)
+        module = get_object_or_404(ScriptModule.objects.restrict(request.user), file_path__startswith=module + ".py")
```
DanSheps, line 1160: I would think iif we are doing ".py" at the end it effectively turns it into `file_path=module + ".py"`

Maybe we should remove `__startswith` if that is what we are going for

venomnom: I think it would make sense to remove the `__startswith` as the name should be exactly right.
Also, in the weird case if a file would be named `something.py.py` and a second one with `something.py` is created, it would lead to the same error as is in #14025 again

arthanson: If you have the script "test.py" module comes in as "test" and file_path is "test.py" which is why the startswith is there.  If you remove startswith it won't match.  I think it would be difficult at this point to change module to keep the full name with .py as I think it is stored that way.  Will check into it more...

jeremystretch:
Can we just use a regex to capture the portion of the file path preceding the dot? Seems like that would be sufficient.

netbox/extras/views.py
DanSheps: > If you have the script "test.py" module comes in as "test" and file_path is "test.py" which is why the startswith is there. If you remove startswith it won't match. I think it would be difficult at this point to change module to keep the full name with .py as I think it is stored that way. Will check into it more... Thinking about it, it might make more sense to remove a potential ".py" from file_path if there is one - that should be pretty safe.

Yeah, but we are matching on file_path, so if you are doing `{module}.py` it is either going to match or not with or without `__startswith`.

Not sure if I am explaining this well, if you need more info hit me up.  I still think startrswith is not needed with the proposed code change and it will be looking for partial matches when a partial match is not even possible.

netbox/extras/views.py
```diff
@@ -1151,13 +1151,17 @@ def get(self, request):
         })
 
 
+def get_script_module(module, request):
```
jeremystretch, line 1154: We should replicate these changes for reports as well.

netbox/core/forms/model_forms.py
```diff
@@ -88,14 +90,34 @@ class Meta:
...
+
+        if upload_file and data_file:
+            extension = Path(upload_file.name).suffix[1:].lower()
             raise forms.ValidationError("Cannot upload a file and sync from an existing file")
-        if not self.cleaned_data.get('upload_file') and not self.cleaned_data.get('data_file'):
+        if not upload_file and not data_file:
             raise forms.ValidationError("Must upload a file or select a data file to sync")
 
+        self._validate_file_extension(upload_file)
+        self._validate_file_extension(data_file)
```
jeremystretch, line 119: Sorry, I think I gave you the wrong impression when we spoke about validation earlier. I meant that we could explore enforcing specific file suffixes, however we can't force that change as part of a bug fix; it would need to be proposed & implemented under a separate FR. The changes in `views.py` should be sufficient to resolve the bug.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #14039, opened 2023-10-13
Closes #13962: Add a copy-to-clipboard button for API tokens

### LINKED ISSUE #13962: Add a copy-to-clipboard button to the API token creation form
### NetBox version

v3.6.3

### Feature type

New functionality

### Proposed functionality

On the API token creation form, add a button to conveniently copy the randomly-generated API token value to the system clipboard.

API token creation form

### Use case

Convenience

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Fixes: #13962

- Add support for including a copy-to-clipboard button for form fields with the `data-clipboard` widget attribute
- Extend `initClipboard()` to support buttons in addition to hyperlinks
- Remove obsolete template code


netbox/templates/account/token.html
```diff
@@ -15,11 +15,6 @@
 {% block content %}
   <div class="row">
     <div class="col col-md-12">
-      {% if key and not settings.ALLOW_TOKEN_RETRIEVAL %}
```
arthanson, line 18: I think a warning message is still relevant?  They have to copy it on this form before leaving?

arthanson (changes requested):

netbox/templates/account/token.html
jeremystretch: The key is only shown on the form prior to creation now; it's never passed to the view afterward.

arthanson (approved):

PR #14048, opened 2023-10-16
14042 mptt cache count

### LINKED ISSUE #14042: Removing Inventory Items with Children Items breaks Inventory count
### NetBox version

v3.6.3

### Python version

3.8

### Steps to Reproduce

1. add an inventory item to a device
2. add a child inventory item to the item from 1.
3. remove both at the same time

### Expected Behavior

For the inventory item count to be correct

### Observed Behavior

Inventory count is -1
Screenshot 2023-10-14 at 13 25 25
Screenshot 2023-10-14 at 13 25 35

arthanson:
### Fixes: #14042 

MPTT code is causing two delete signals to come through if nested objects, in this case the instance and origin (from kwargs) are different so put in a check to ignore updating the counter if this is the case.

### DIFF
--- a/netbox/utilities/counters.py
+++ b/netbox/utilities/counters.py
@@ -72,7 +72,10 @@ def post_delete_receiver(sender, instance, **kwargs):
 
         # Decrement the parent's counter by one
         if parent_pk is not None:
-            update_counter(parent_model, parent_pk, counter_name, -1)
+            # MPTT sends two delete signals if child element so guard against multiple decrements
+            origin = kwargs.get('origin')
+            if not origin or origin == instance:
+                update_counter(parent_model, parent_pk, counter_name, -1)
 
 
 #
--- a/netbox/utilities/counters.py
+++ b/netbox/utilities/counters.py
@@ -62,7 +62,7 @@ def post_save_receiver(sender, instance, created, **kwargs):
             update_counter(parent_model, new_pk, counter_name, 1)
 
 
-def post_delete_receiver(sender, instance, **kwargs):
+def post_delete_receiver(sender, instance, origin, **kwargs):
     """
     Update counter fields on related objects when a TrackingModelMixin subclass is deleted.
     """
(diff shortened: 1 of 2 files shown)


netbox/utilities/tests/test_counters.py
```diff
@@ -79,3 +82,33 @@ def test_interface_count_move(self):
         device2.refresh_from_db()
         self.assertEqual(device1.interface_count, 1)
         self.assertEqual(device2.interface_count, 3)
+
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])
+    def test_mptt_child_delete(self):
+        device1, device2 = Device.objects.all()
+        inventory_item1 = InventoryItem.objects.create(device=device1, name='Inventory Item 1')
+        inventory_item2 = InventoryItem.objects.create(device=device1, name='Inventory Item 1', parent=inventory_item1)
```
jeremystretch, line 90:
```suggestion
        inventory_item2 = InventoryItem.objects.create(device=device1, name='Inventory Item 2', parent=inventory_item1)
```

netbox/utilities/tests/test_counters.py
```diff
@@ -79,3 +82,33 @@ def test_interface_count_move(self):
...
+        }
+
+        # Assign unconstrained permission
+        obj_perm = ObjectPermission(
+            name='Test permission',
+            actions=['delete']
+        )
+        obj_perm.save()
+        obj_perm.users.add(self.user.pk)
+        obj_perm.object_types.add(ContentType.objects.get_for_model(InventoryItem))
```
jeremystretch, line 109: You can use the `add_permissions()` test case method for this.
```suggestion
        self.add_permissions('dcim.delete_inventoryitem')
```

netbox/utilities/counters.py
```diff
@@ -72,7 +72,10 @@ def post_delete_receiver(sender, instance, **kwargs):
 
         # Decrement the parent's counter by one
         if parent_pk is not None:
-            update_counter(parent_model, parent_pk, counter_name, -1)
+            # MPTT sends two delete signals if child element so guard against multiple decrements
+            origin = kwargs.get('origin')
```
jeremystretch, line 76: We should declare `origin` as a kwarg in the receiver's signature so it can be referenced directly. Also, it's required as of Django 4.1 so it should be safe to not specify a default value.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #14050, opened 2023-10-16
14033 raise validation error if A and B term go to same object

### LINKED ISSUE #14033: Cable registration to the same Interface results in a server error.
### NetBox version

v3.6.3

### Python version

3.8

### Steps to Reproduce

1. Register the same termination at the time of cable registration by selecting the same termination.
2. A server error is output.
image
image

### Expected Behavior

The UI displays an error that the same location cannot be registered. Or, it is expected that the same IF cannot be selected from IF candidates.

### Observed Behavior

A server error was displayed.
This error is related to the database, and the registration process itself seems to be running.

arthanson:
### Fixes: #14033 

On cable creation, don't allow cables to have A and B terminations pointing to the same object.

### DIFF
--- a/netbox/dcim/forms/connections.py
+++ b/netbox/dcim/forms/connections.py
@@ -103,4 +103,10 @@ def get_cable_form(a_type, b_type):
             self.instance.a_terminations = self.cleaned_data['a_terminations']
             self.instance.b_terminations = self.cleaned_data['b_terminations']
 
+            if a_type == b_type and self.instance.a_terminations and self.instance.b_terminations:
+                if self.instance.a_terminations.intersection(self.instance.b_terminations):
+                    raise forms.ValidationError(
+                        _("A and B terminations cannot connect to the same object.")
+                    )
+
     return _CableForm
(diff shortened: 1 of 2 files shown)


netbox/dcim/forms/connections.py
```diff
@@ -103,4 +103,10 @@ def clean(self):
             self.instance.a_terminations = self.cleaned_data['a_terminations']
             self.instance.b_terminations = self.cleaned_data['b_terminations']
 
+            if a_type == b_type and self.instance.a_terminations and self.instance.b_terminations:
```
abhi1693, line 106: I've not checked but is this an issue on the API as well?

jeremystretch (changes requested):
Per @abhi1693's comment, let's move this logic from the form to `Cable.clean()` to ensure the same validation is applied for cables created via the REST API as well.

netbox/dcim/forms/connections.py
arthanson: moved to cable clean

jeremystretch (approved):

jeremystretch:
Thanks @arthanson!

PR #14062, opened 2023-10-17
Closes #13550: Refactor view action mappings

### LINKED ISSUE #13550: Optimize the declaration of view actions
### Proposed Changes

`ActionsMixin` defines two attributes on a view:

* `actions` - A tuple of support action names
* `action_perms` - A dictionary mapping of action names to a set of required permissions for each

The following actions are defined by default:

```python
    actions = ('add', 'import', 'export', 'bulk_edit', 'bulk_delete')
    action_perms = defaultdict(set, **{
        'add': {'add'},
        'import': {'add'},
        'bulk_edit': {'change'},
        'bulk_delete': {'delete'},
    })
```

These can be combined into a single mapping:

```python
    actions = {
        'add': {'add'},
        'import': {'add'},
        'export': set(),
        'bulk_edit': {'change'},
        'bulk_delete': {'delete'},
    }
```

### Justification

This change will simplify the list of support actions for a view and reduce duplicate code.

jeremystretch:
### Closes: #13550

- Delete `action_perms` map from ActionsMixin
- Convert `actions` from a tuple to a dictionary mapping
- Normalize permissions for all actions (every action should require at least one permission)
- Ensure backward compatibility for plugins

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -20,6 +20,7 @@ from circuits.models import Circuit, CircuitTermination
 from extras.views import ObjectConfigContextView
 from ipam.models import ASN, IPAddress, Prefix, VLAN, VLANGroup
 from ipam.tables import InterfaceVLANTable
+from netbox.constants import DEFAULT_ACTION_PERMISSIONS
 from netbox.views import generic
 from tenancy.views import ObjectContactsView
 from utilities.forms import ConfirmationForm
@@ -47,11 +48,7 @@ CABLE_TERMINATION_TYPES = {
 
 class DeviceComponentsView(generic.ObjectChildrenView):
     actions = {
-        'add': {'add'},
-        'import': {'add'},
-        'export': {'view'},
-        'bulk_edit': {'change'},
-        'bulk_delete': {'delete'},
+        **DEFAULT_ACTION_PERMISSIONS,
         'bulk_rename': {'change'},
         'bulk_disconnect': {'change'},
     }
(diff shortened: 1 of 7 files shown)


netbox/dcim/views.py
```diff
@@ -2005,11 +2019,18 @@ class DeviceDeviceBaysView(DeviceComponentsView):
...
     filterset = filtersets.InventoryItemFilterSet
     template_name = 'dcim/device/inventory.html'
+    actions = {
+        'add': {'add'},
+        'import': {'add'},
+        'export': {'view'},
+        'bulk_edit': {'change'},
+        'bulk_delete': {'delete'},
+        'bulk_rename': {'change'},
+    }
```
arthanson, line 2033: These are all duplicated (lines 1980, 2003, 2026, 2211, 2283, ...), can we make a STD_ACTIONS defined and then just do actions = STD_ACTIONS to be more DRY?

netbox/netbox/views/generic/mixins.py
```diff
@@ -9,25 +9,58 @@
 
 
 class ActionsMixin:
-    actions = ('add', 'import', 'export', 'bulk_edit', 'bulk_delete')
-    action_perms = defaultdict(set, **{
+    # Map action names to the set of required permissions for each
```
arthanson, line 12: IMHO it might be useful here to have a doc-comment of standard actions - basically just bulk_rename and bulk_sync add here.  I think bulk_remove_devices is the only special case / used once.

arthanson (approved):

PR #14065, opened 2023-10-17
14044 handle bulk rename if None name

### LINKED ISSUE #14044: NoneType AttributeError when Bulk Renaming Nameless Devices
### NetBox version

3.6.3

### Python version

3.8

### Steps to Reproduce

1. Create devices without a name
2. Select and Bulk Rename those devices
3. **Disable** Regex
4. Attempt a rename operation

Of note, I am also unable to rename nameless devices with Regex on, but no error is observed.

### Expected Behavior

The renaming of bulk devices operation should be successful.

### Observed Behavior

```python
<class 'AttributeError'>

'NoneType' object has no attribute 'replace'

Python version: 3.8.10
NetBox version: 3.6.3
Plugins: 
  netbox_demo: 0.3.1
```

arthanson:
### Fixes: #14044 

Checks if name empty for bulk_rename - put at top of procedure as there is a bug (beyond the issue) where if you use regex you get '' (instead of None) for the name which can cause postgres duplicate error if in the same site.

### DIFF
--- a/netbox/netbox/views/generic/bulk_views.py
+++ b/netbox/netbox/views/generic/bulk_views.py
@@ -689,14 +689,19 @@ class BulkRenameView(GetReturnURLMixin, BaseMultiObjectView):
 
             find = form.cleaned_data['find']
             replace = form.cleaned_data['replace']
-            if form.cleaned_data['use_regex']:
-                try:
-                    obj.new_name = re.sub(find, replace, obj.name or '')
-                # Catch regex group reference errors
-                except re.error:
-                    obj.new_name = obj.name
+            if obj.name:
+                if form.cleaned_data['use_regex']:
+                    try:
+                        obj.new_name = re.sub(find, replace, obj.name or '')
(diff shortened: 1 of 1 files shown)


netbox/netbox/views/generic/bulk_views.py
```diff
@@ -689,14 +689,17 @@ def _rename_objects(self, form, selected_objects):
...
-                    obj.new_name = re.sub(find, replace, obj.name or '')
+                    obj.new_name = re.sub(find, replace, obj.name)
                 # Catch regex group reference errors
                 except re.error:
                     obj.new_name = obj.name
-            else:
+            elif obj.name:
                 obj.new_name = obj.name.replace(find, replace)
+            else:
+                obj.new_name = obj.name
```
jeremystretch, line 701: This effectively bypasses the rename operation, which may be sufficient, however #14044 seems to expect that unnamed devices would be renamed.

PR #14087, opened 2023-10-19
Closes #8356: Add virtual disk to Virtual Machines

### LINKED ISSUE #8356: Discrete resource tracking for VirtualMachines
### NetBox version

v3.1.5

### Feature type

### Proposed functionality

Device have the option to document inventory items. (cpu, disks, blades, modules, ...)
For virtual machines,this option is not available.

I would like to see the possibility to add inventory items to virtual-machines

### Use case

VM can have additional 'hardware' like disks.
The inventory items on a VM will allow us to document how many disks and what size the disks has.
By default we have no CDROM attached to the VM. with the inventory we could add a CDROM to document that exception

### Database changes

extend the dcim/inventory-items model to be attached to virtualization/virtual-machines

### External dependencies

_No response_

arthanson:
### Fixes: #8356 

introduces a new `VirtualDisk` model with `name` and `size` fields, instances of which must be assigned to a specific virtual machine.  Retains the existing `disk` field on VirtualMachine.

Switched Virtual Machine list view to show Disk Size (from Virtual Disks) instead of Disk - made it default, but user can add disks back if needed. I think this makes it very clear that they should be moved.

Also put in deprecation warnings for Disk in detail / edit view:

Monosnap Editing virtual machine vm1 | NetBox 2023-10-23 11-22-17
Monosnap vm1 | NetBox 2023-10-23 11-24-08

Added Virtual Disks to end of Virtual Machine detail view:
Monosnap vm1 | NetBox 2023-10-24 10-13-32


ITJamie:
Can i suggest that we add config context to the new VirtualDisk model. 
this would be useful for documenting disk type information (disk io settings, adaptor type, disk position ids, storage location info etc).

netbox/virtualization/api/nested_serializers.py
```diff
@@ -72,3 +73,12 @@ class NestedVMInterfaceSerializer(WritableNestedSerializer):
...
         fields = ['id', 'url', 'display', 'virtual_machine', 'name']
+
+
+class NestedVirtualDiskSerializer(WritableNestedSerializer):
+    url = serializers.HyperlinkedIdentityField(view_name='virtualization-api:virtualdisk-detail')
+    virtual_machine = NestedVirtualMachineSerializer(read_only=True)
+
+    class Meta:
+        model = VirtualDisk
+        fields = ['id', 'url', 'virtual_machine', 'name', 'size']
```
jeremystretch, line 84: Nested serializers should include the `display` field, but we can omit the `size` field.

```suggestion
        fields = ['id', 'url', 'display', 'virtual_machine', 'name']
```

netbox/templates/virtualization/virtualdisk.html
```diff
@@ -0,0 +1,59 @@
...
+{% load plugins %}
+{% load render_table from django_tables2 %}
+{% load i18n %}
+
+{% block breadcrumbs %}
+  {{ block.super }}
+  <li class="breadcrumb-item">
+    {% comment %}
+     <a href="{% url 'virtualization:virtualmachine_virtualdisks' pk=object.virtual_machine.pk %}">{{ object.virtual_machine }}</a>
+    {% endcomment %}
```
jeremystretch, line 12: Should be okay to uncomment

netbox/virtualization/tests/test_api.py
```diff
@@ -6,7 +6,7 @@
 from ipam.models import VLAN, VRF
 from utilities.testing import APITestCase, APIViewTestCases, create_test_device
 from virtualization.choices import *
-from virtualization.models import Cluster, ClusterGroup, ClusterType, VirtualMachine, VMInterface
+from virtualization.models import Cluster, ClusterGroup, ClusterType, VirtualDisk, VirtualMachine, VMInterface
```
jeremystretch, line 9:
```suggestion
from virtualization.models import *
```

netbox/virtualization/tests/test_filtersets.py
```diff
@@ -6,7 +6,7 @@
 from utilities.testing import ChangeLoggedFilterSetTests, create_test_device
 from virtualization.choices import *
 from virtualization.filtersets import *
-from virtualization.models import Cluster, ClusterGroup, ClusterType, VirtualMachine, VMInterface
+from virtualization.models import Cluster, ClusterGroup, ClusterType, VirtualDisk, VirtualMachine, VMInterface
```
jeremystretch, line 9:
```suggestion
from virtualization.models import *
```

netbox/virtualization/tests/test_views.py
```diff
@@ -7,7 +7,7 @@
 from ipam.models import VLAN, VRF
 from utilities.testing import ViewTestCases, create_tags, create_test_device
 from virtualization.choices import *
-from virtualization.models import Cluster, ClusterGroup, ClusterType, VirtualMachine, VMInterface
+from virtualization.models import Cluster, ClusterGroup, ClusterType, VirtualDisk, VirtualMachine, VMInterface
```
jeremystretch, line 10:
```suggestion
from virtualization.models import *
```

jeremystretch (changes requested):
Great work, @arthanson! Just have some minor cleanup notes.

Also, we should consider how best to report disk space for a VM given the presence of both VirtualDisks and the `disk` field on VirtualMachine. E.g. it might make sense to introduce a `total_disk_size` property on VirtualMachine.

netbox/virtualization/views.py
```diff
@@ -350,7 +351,8 @@ class VirtualMachineListView(generic.ObjectListView):
 
 @register_model_view(VirtualMachine)
 class VirtualMachineView(generic.ObjectView):
-    queryset = VirtualMachine.objects.all()
+    queryset = VirtualMachine.objects.all().annotate(
+        disk_size=Sum('virtualdisks__size'))
```
jeremystretch, line 355: There isn't much benefit to annotating the aggregate disk size in the queryset. We should add a method or property on VirtualMachine to return the aggregate size of assigned VirtualDisks.

netbox/virtualization/views.py
```diff
@@ -341,7 +341,8 @@ class ClusterContactsView(ObjectContactsView):
 #
 
 class VirtualMachineListView(generic.ObjectListView):
-    queryset = VirtualMachine.objects.prefetch_related('primary_ip4', 'primary_ip6')
+    queryset = VirtualMachine.objects.annotate(
+        disk_size=Sum('virtualdisks__size')).prefetch_related('primary_ip4', 'primary_ip6')
```
jeremystretch, line 345:
```suggestion
    queryset = VirtualMachine.objects.annotate(
        disk_size=Sum('virtualdisks__size')
    ).prefetch_related('primary_ip4', 'primary_ip6')
```

netbox/virtualization/api/serializers.py
```diff
@@ -84,14 +84,15 @@ class VirtualMachineSerializer(NetBoxModelSerializer):
 
     # Counter fields
     interface_count = serializers.IntegerField(read_only=True)
+    virtual_disk_count = serializers.IntegerField(read_only=True)
```
jeremystretch, line 87: We might also want to include the aggregate size of all assigned VirtualDisks.

jeremystretch (changes requested):

netbox/virtualization/views.py
arthanson: left this one in as having the property interferes with the same name for the list view queryset on the table, so would need to have the property and the calculated value in the queryset different which seems overly complicated.  I think using aggregation here is cleanest.

netbox/virtualization/api/serializers.py
```diff
@@ -84,14 +84,16 @@ class VirtualMachineSerializer(NetBoxModelSerializer):
...
+    virtual_disk_space = serializers.IntegerField(read_only=True)
```
jeremystretch, line 88: I'm trying to decide what approach makes the most sense for migration here. I propose we:

1. Override the `disk` field on the serializer to return the value of the concrete `disk` field only if it's set, and otherwise return the aggregate size of all attached VirtualDisks.
2. Possibly also include a list of nested VirtualDisks, indicating the name and size of each. This would be a departure from the current pattern of not including child objects, but might make sense here.

netbox/virtualization/tables/virtualmachines.py
```diff
@@ -84,20 +86,26 @@ class VirtualMachineTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable)
...
         verbose_name=_('Interfaces')
     )
+    virtual_disk_count = tables.Column(
+        verbose_name=_('Virtual Disks')
+    )
     config_template = tables.Column(
         verbose_name=_('Config Template'),
         linkify=True
     )
+    disk_size = tables.Column(
```
jeremystretch, line 96: Similar to the API serializer, it might be cleaner to override the `disk` column to return the concrete field if set, and otherwise fall back to reporting the aggregate size of all attached VirtualDisks. This can be normalized once we've removed the `disk` field in a future release.

netbox/virtualization/api/serializers.py
arthanson: I didn't put in 2 as it was different then all other APIs and sort of a slipper slope.  Thought it might be better to have it like the other APIs and see if there are requests to add it directly to the main serializer.

netbox/virtualization/tables/virtualmachines.py
```diff
@@ -84,10 +86,17 @@ class VirtualMachineTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable)
...
         verbose_name=_('Interfaces')
     )
+    virtual_disk_count = tables.Column(
+        verbose_name=_('Virtual Disks')
+    )
     config_template = tables.Column(
         verbose_name=_('Config Template'),
         linkify=True
     )
+    disk = tables.Column(
```
jeremystretch, line 96: If my congestion above is implemented, we could replace this with `virtual_disk_space`.

jeremystretch (changes requested):

netbox/virtualization/api/serializers.py
```diff
@@ -84,17 +84,25 @@ class VirtualMachineSerializer(NetBoxModelSerializer):
 
     # Counter fields
     interface_count = serializers.IntegerField(read_only=True)
+    virtual_disk_count = serializers.IntegerField(read_only=True)
+    disk = serializers.SerializerMethodField(read_only=True)
```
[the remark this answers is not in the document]
arthanson: kept disk as writeable for backwards compatibility and added disk_space, used disk_space globally instead of virtual_disk_space.

amhn:
Just a minor note:
Using the GUI the URL component is named "disks" (e.g. http://localhost:8000/virtualization/disks/1/) while in the API it is named "virtual-disks" (e.g. http://localhost:8000/api/virtualization/virtual-disks/1/). Probably should be the same in both.

netbox/virtualization/api/serializers.py
```diff
@@ -84,17 +84,24 @@ class VirtualMachineSerializer(NetBoxModelSerializer):
...
-            'interface_count',
+            'interface_count', 'virtual_disk_count', 'disk_space',
         ]
         validators = []
 
+    def get_disk(self, obj):
+        if obj.disk:
+            return obj.disk
+
+        return obj.disk_space
```
jeremystretch, line 103: This doesn't appear to be needed anymore.

netbox/virtualization/models/virtualmachines.py
```diff
@@ -235,6 +242,10 @@ def primary_ip(self):
         else:
             return None
 
+    @property
+    def disk_space(self):
```
jeremystretch, line 246: This should return the value of the `disk` field if no virtual disks have been created.

netbox/virtualization/models/virtualmachines.py
```diff
@@ -235,6 +242,10 @@ def primary_ip(self):
         else:
             return None
 
+    @property
```
jeremystretch, line 245: Would it be safe to make this a cached property?

netbox/virtualization/tables/virtualmachines.py
```diff
@@ -84,27 +86,42 @@ class VirtualMachineTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable)
...
         default_columns = (
-            'pk', 'name', 'status', 'site', 'cluster', 'role', 'tenant', 'vcpus', 'memory', 'disk', 'primary_ip',
+            'pk', 'name', 'status', 'site', 'cluster', 'role', 'tenant', 'vcpus', 'memory', 'disk_space', 'primary_ip',
         )
 
+    def render_disk_space(self, value, record):
+        if record.disk:
+            return record.disk
+        elif record.disk_space:
+            return record.disk_space
```
jeremystretch, line 116: We should move this logic to the `disk_space` property.

netbox/virtualization/api/serializers.py
```diff
@@ -104,7 +111,7 @@ class Meta(VirtualMachineSerializer.Meta):
             'id', 'url', 'display', 'name', 'status', 'site', 'cluster', 'device', 'role', 'tenant', 'platform',
             'primary_ip', 'primary_ip4', 'primary_ip6', 'vcpus', 'memory', 'disk', 'description', 'comments',
             'local_context_data', 'tags', 'custom_fields', 'config_context', 'created', 'last_updated',
-            'interface_count',
+            'interface_count', 'virtual_disk_count',
```
jeremystretch, line 114: We should include `disk_space` here as well.

netbox/virtualization/api/serializers.py
```diff
@@ -84,17 +84,24 @@ class VirtualMachineSerializer(NetBoxModelSerializer):
...
+    virtual_disk_count = serializers.IntegerField(read_only=True)
 
     class Meta:
         model = VirtualMachine
         fields = [
             'id', 'url', 'display', 'name', 'status', 'site', 'cluster', 'device', 'role', 'tenant', 'platform',
             'primary_ip', 'primary_ip4', 'primary_ip6', 'vcpus', 'memory', 'disk', 'description', 'comments',
             'config_template', 'local_context_data', 'tags', 'custom_fields', 'created', 'last_updated',
-            'interface_count',
+            'interface_count', 'virtual_disk_count', 'disk_space',
```
jeremystretch, line 95: We should figure out a way to annotate `disk_space` on each instance, to avoid calling a separate query per VM on list view.

jeremystretch (changes requested):
I'm going to take a stab at annotating the disk size on VM querysets.

jeremystretch (approved):

Issue #13842, opened 2023-09-21
Bookmark delete button text change


HamagenCZE:
### NetBox version

v3.6.2

### Feature type

Change to existing functionality

### Proposed functionality

I suggest changing the notification window when the Delete button is pressed in /user/bookmarks/ for the object.

At the moment the text is "Are you sure you want to delete #OBJECT NAME#?: Which evokes the feeling that the entire object will be deleted and not the bookmark itself.

I suggest changing the text to ""Are you sure you want to delete your bookmark for #OBJECT NAME#?"

### Use case

My colleague and I, who manage Netbox, both had the same feeling that the button would delete the entire object instead of the bookmark. We were forced to make a dummy object, bookmark it and then test the button to be sure. This change, I think, will help the clarity of the button.

ITJamie:
Related to https://github.com/netbox-community/netbox/issues/13690

abhi1693:
@ITJamie could you please explain how it's related? I don't see any relation. This FR is simply asking for a delete message update whereas yours is changing how the delete works by showing related objects.

ITJamie:
See the screenshots i added in that fr. It shows exactly what would be deleted including the primay object itself. 

If you were deleting just a bookmark it would say "bookmark - bookmark name" in the list of items that would be deleted

HamagenCZE:
I dont see a relation too. You are talking about normal delate behavior and about object deletion and relative links to that object, my propose was only for changing of text to clarify function of a bin icon in Bookmarks list which is not a object deletion at all

ITJamie:
Sorry for the delay I was on vacation.
The initial version of a branch I made for listing objects that would be deleted included the primary item. 

My view was that the delete modal right now does not give enough details and that this issue is related heavily to the other issue.

I do think it's odd that the `{{ object_type }}` isn't rendering for bookmark objects though...

here is a screenshot of what the output of the delete form would look like with my current dev branch
<img width="571" alt="Screenshot 2023-09-29 at 21 21 02" src="https://github.com/netbox-community/netbox/assets/1613241/1590db50-9bc5-4f5c-b3f6-025f0e265520">

The text etc could be re-worked but I'm of the view that the delete modal needs to present as much information as possible

ITJamie:
I found the reason as to why `{{ object_type }}` doesn't show on the delete page for bookmarks.
[...]

PR #14089, opened 2023-10-20
Closes #13690: List all objects to be deleted

### LINKED ISSUE #13690: when deleteing an object in webui. list the related objects that would also be deleted
### NetBox version

v3.6.0

### Feature type

### Proposed functionality

when choosing delete on an object (eg device for this example). the popup that "asks are you sure you want to delete device xxx" should be extended to show the list of related objects that will also be deleted (interfaces, cables, etc)

### Use case

some users are not aware that deleting an object will remove some other related objects.
showing the user what additional objects will deleted would be a fantastic safety feature.

### Database changes

n/a

### External dependencies

n/a

ITJamie:
### Fixes: #13690

List all objects to be deleted when deleting an item.

Currently, the list includes the original item being deleted. we can change that pretty easily but I would like feedback first as to whether the list should be exhaustive or just the additional items (see #13842 as to why including the original item in the list might make sense)

Screenshot 2023-10-20 at 07 52 22
Screenshot 2023-10-20 at 07 53 26


netbox/netbox/views/generic/object_views.py
```diff
@@ -334,6 +335,18 @@ def get(self, request, *args, **kwargs):
...
         form = ConfirmationForm(initial=request.GET)
 
+        using = router.db_for_write(obj._meta.model)
+        collector = Collector(using=using)
+        collector.collect([obj])
+        deletion_objects = []
+        if collector.instances_with_model():
+            for model, instance in collector.instances_with_model():
+                # we could ignore the instance == obj so that the list doesnt contain itself...
+                deletion_objects.append({
```
ITJamie, line 345: this append could be gated to not include the original item
```
        if collector.instances_with_model():
            for model, instance in collector.instances_with_model():
                if not instance == obj:
                    deletion_objects.append({
                        "modelname":f"{model.__name__}", 
                        "object": instance,
                    })
```

abhi1693:
I like the general idea but this will turn out to be a bad ux if the similar objects are listed in the new line because it may go beyond the actual page making a huge scroll. I'd prefer the similar objects be comma separated instead.

netbox/netbox/views/generic/object_views.py
```diff
@@ -334,6 +335,18 @@ def get(self, request, *args, **kwargs):
...
+                    "modelname":f"{model.__name__}",
```
abhi1693, line 346: What's the need of the f-string here? Isn't the name already a string?

abhi1693 (changes requested):

ITJamie:
> I like the general idea but this will turn out to be a bad ux if the similar objects are listed in the new line because it may go beyond the actual page making a huge scroll. I'd prefer the similar objects be comma separated instead.

I can absolutely refactor for that. For switches it makes a lot of sense as there can be a lot of ports and make the list long. 

The one advantage to that is that it really reinforces just how much impact the current deletion will have by having to scroll to the delete button 😂

abhi1693:
I just noticed this is to be released in v3.7, so it should go into the feature branch instead.

ITJamie:
Screenshot 2023-10-20 at 12 04 29
@abhi1693 something more like this?

netbox/netbox/views/generic/object_views.py
```diff
@@ -334,6 +335,18 @@ def get(self, request, *args, **kwargs):
         obj = self.get_object(**kwargs)
         form = ConfirmationForm(initial=request.GET)
 
+        using = router.db_for_write(obj._meta.model)
+        collector = Collector(using=using)
+        collector.collect([obj])
+        deletion_objects = []
+        if collector.instances_with_model():
```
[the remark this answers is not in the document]
ITJamie: removed the unneeded if statement

netbox/netbox/views/generic/object_views.py
ITJamie: cleaned up. was a leftover from my first pass

ITJamie:
@kkthxbye-code re the upvotes and downvotes. Is there anything specific you would like me to adjust regards the ui view change from the first screenshots to the adjusted screenshots based on @abhi1693 feedback?

kkthxbye-code:
> @kkthxbye-code re the upvotes and downvotes. Is there anything specific you would like me to adjust regards the ui view change from the first screenshots to the adjusted screenshots based on @abhi1693 feedback?

I think your initial version was much more readble than the version abhi suggested. Showing a large number of varying length list items as comma seperated tends to look very messy.

If the page size is an issue (I personally don't think it is), maybe a scrolling container could solve it, or truncating the output with a "show more" button that extends it.

Also the trailing comma should be removed if the second version is chosen.

netbox/templates/htmx/delete_form.html
```diff
@@ -12,6 +12,18 @@ <h5 class="modal-title">{% trans "Confirm Deletion" %}</h5>
...
+    {% if related_objects %}
+    <p>
+      This will cause deletion of the following items: <br>
+      {% for model,objects in related_objects.items %} 
+        - {{ model }}:
+        {% for object in objects %}
+          {{ object | linkify }}{% if not forloop.last %}, {% endif %}
+        {% endfor %}
+        <br>
+      {% endfor %}
```
jeremystretch, line 24: IMO we should present the list of dependent objects in a scrollable list. Bootstrap's [accordion component](https://getbootstrap.com/docs/5.1/components/accordion/) could work very well here.

jeremystretch (changes requested):

netbox/templates/htmx/delete_form.html
ITJamie: Screenshot 2023-10-31 at 21 14 27
Screenshot 2023-10-31 at 21 14 32

happier with this? (im using verbose_name and verbose_name_plural as needed for the accordion headers, some are capitalised, some are not)

ITJamie:
Current version:
- using bootstrap accordian
- titles are now being passed in + using plurals 
- added `_get_dependent_objects` to `ObjectDeleteView` class
- using 
Screenshot 2023-10-31 at 21 50 47
Screenshot 2023-10-31 at 21 50 58

jeremystretch (approved):
Nice work @ITJamie!

jeremystretch (approved):

PR #14095, opened 2023-10-20
Closes #13381: Enable plugins to register custom data backends

### LINKED ISSUE #13381: Provide a mechanism for plugins to register custom remote data source backends
### NetBox version

v3.5.7

### Feature type

New functionality

### Proposed functionality

Provide and support a mechanism for NetBox plugins to register their own remote data source backends, to complement the two backends (git and Amazon S3) currently provided natively. This will likely entail just formally documenting the `register_backend()` decorator, which is currently an internal resource.

### Use case

This will enable users to introduce support for additional backends by installing NetBox plugins.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Fixes: #13381

- Move base DataBackend class from `core` to `netbox`
- Remove static choices for `DataSource.type`
- Validate `DataSource.type` against registered backends
- Move `register_backend()` wrapper to `core.utils`
- Introduce `get_data_backend_choices()` utility function
- Add a `data_backends` attribute to PluginConfig
- Establish `name`, `label`, and `is_local` attributes on DataBackend
- Remove colors from DataSource types


netbox/core/utils.py
```diff
@@ -0,0 +1,26 @@
+from netbox.registry import registry
```
arthanson, line 1: should this be in netbox/core/utils.py or netbox/netbox/utils.py?

arthanson (changes requested):

netbox/core/utils.py
jeremystretch: I don't have a strong opinion, but I put it here because DataBackend lives in `core`. We should probably have a higher-level discussion about the organization of utility functions in general. I'm okay leaving this as-is for now and potentially moving prior to the v3.7 beta release.

PR #14098, opened 2023-10-20
12216 Add color to circuit-type and add to SVG rendering

### LINKED ISSUE #12216: Add colors to Circuit Types model and use this when rendering cable traces
### NetBox version

v3.4.6

### Feature type

New functionality

### Proposed functionality

As described in [this discussion](https://github.com/netbox-community/netbox/discussions/12035), make it possible to identify the type of circuit that's involved in a cable trace by color-coding circuit types, and including the circuit type's name in the circuit's block in the diagram.

This would involve:
* Extending the circuit type model to include a color value (identical to how device roles are color-coded);
* Updating the cable trace view to reflect the circuit type's color and name.

This FR proposes no changes to cable traces that do not involve a circuit.

### Use case

Currently there is no way to identify the type of a circuit involved in a cable trace just by looking at the diagram. Color-coding and labeling circuit types adds convenience when looking at multiple cable traces, especially if the cable trace is exported or viewed by someone without access to NetBox, and hence can't resolve the circuit type by clicking on it, or looking it up.

### Database changes

Circuit types need a field for color value.

### External dependencies

None.

arthanson:
### Fixes: #12216 

* Extending the circuit type model to include a color value (identical to how device roles are color-coded);
* Updating the cable trace view to reflect the circuit type's color and name.
Monosnap Cable Trace for Circuit Termination 1002840283: Termination Z | NetBox 2023-10-20 15-09-54


netbox/circuits/forms/bulk_edit.py
```diff
@@ -99,9 +103,9 @@ class CircuitTypeBulkEditForm(NetBoxModelBulkEditForm):
 
     model = CircuitType
     fieldsets = (
-        (None, ('description',)),
+        (None, ('color', 'description',)),
```
jeremystretch, line 106:
```suggestion
        (None, ('color', 'description')),
```

netbox/circuits/forms/bulk_edit.py
```diff
@@ -99,9 +103,9 @@ class CircuitTypeBulkEditForm(NetBoxModelBulkEditForm):
...
     )
-    nullable_fields = ('description',)
+    nullable_fields = ('color', 'description',)
```
jeremystretch, line 108:
```suggestion
    nullable_fields = ('color', 'description')
```

netbox/dcim/svg/cables.py
```diff
@@ -181,6 +182,8 @@ def _get_color(cls, instance):
         if hasattr(instance, 'role'):
             # Device
             return instance.role.color
+        elif instance._meta.model_name == 'circuit':
```
jeremystretch, line 185: Need to check whether `color` is actually set. (Device role, for instance, _requires_ a color, so this isn't a concern above.)

jeremystretch (changes requested):
Nice work!

jeremystretch (approved):

Issue #13230, opened 2023-07-21
Allow certain device type to be excluded from rack utilization percentage


thefreakquency:
### NetBox version

v3.5.6

### Feature type

Change to existing functionality

### Proposed functionality

Currently, any device that is assigned to a rack unit counts towards the percentage of use of the given rack. 
As an example, when documenting what has been installed, if someone puts in the blank panels as it's really installed in the rack for airflow management, rack utilization can go up to 100% when in fact, "real" equipment is only partially using the overall space. 

I would like to have a way to discriminate what is counted, so we can have a real idea of the current real use. By having a flag in the Device Type, or on the device itself, that allow the user to specify this should be excluded from utilization calculation. 

### Use case

We document everything that is installed in the rack. This helps planner know that if a 5U filler plate has to be replaced for something smaller, that he takes that into account when he plan a server install. Same goes for hardware removal. 

### Database changes

Adding the new "flag" would most likely require a single new field in either Device Type or Device Table. 

### External dependencies

None

SirVilhelm:
I stumbled across this while googling for this exact solution. We record rack fillers and they make each of our racks show 100% utilization

PR #14099, opened 2023-10-20
13230 Allow Devices to be excluded from Rack utilization

arthanson:
### Fixes: #13230 

Sets flag to exclude items from rack utilization.  Not sure if there is a better (shorter) name that would work `exclude_from_utilization` is descriptive, but a bit verbose.


netbox/dcim/forms/bulk_edit.py
```diff
@@ -445,7 +450,7 @@ class DeviceTypeBulkEditForm(NetBoxModelBulkEditForm):
 
     model = DeviceType
     fieldsets = (
-        (_('Device Type'), ('manufacturer', 'default_platform', 'part_number', 'u_height', 'is_full_depth', 'airflow', 'description')),
+        (_('Device Type'), ('manufacturer', 'default_platform', 'part_number', 'u_height', 'exclude_from_utilization', 'is_full_depth', 'airflow', 'description')),
```
jeremystretch, line 453:
```suggestion
        (_('Device Type'), (
            'manufacturer', 'default_platform', 'part_number', 'u_height', 'exclude_from_utilization', 'is_full_depth',
            'airflow', 'description',
        )),
```

netbox/dcim/models/racks.py
```diff
@@ -357,7 +357,7 @@ def get_rack_units(self, user=None, face=DeviceFaceChoices.FACE_FRONT, exclude=N
 
         return [u for u in elevation.values()]
 
-    def get_available_units(self, u_height=1, rack_face=None, exclude=None):
+    def get_available_units(self, u_height=1, rack_face=None, exclude=None, exclude_from_utilization=False):
```
jeremystretch, line 360: The purpose of the `exclude_from_utilization` parameter isn't immediately clear. Is it really needed? If so, we should rename it to something like `ignore_excluded_devices`.

jeremystretch (changes requested):
This is still missing a few things:

- Add the `exclude_from_utilization` field to the REST API serializer
- Add a filter for `exclude_from_utilization` on DeviceTypeFilterSet

The developer docs has a page on [extending models](https://docs.netbox.dev/en/stable/development/extending-models/) that works well as a checklist for this sort of thing.

jeremystretch (approved):

PR #14106, opened 2023-10-23
Closes #13334: Record error message on failed jobs

### LINKED ISSUE #13334: Record data source sync errors in the job data
### NetBox version

v3.5.7

### Feature type

### Proposed functionality

When a data source sync job fails, record any relevant error message(s) in the job data. (This is handled by the `sync_datasource()` function.)

### Use case

This will make it easier to troubleshoot issues without shell access. Currently, these errors are being logged but not recorded in the job directly.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Fixes: #13334

- Add an optional `error` text field on Job
- Extend the `terminate()` method to support passing an error message for failed jobs
- Update calls to `Job.terminate()` to pass error message where applicable


arthanson (approved):
Q: Would there be any security implications to this as showing the error would include traceback info.  Would we want a permission to show this info on the job results at all?  Probably not, but just wanted to mention it.

jeremystretch:
AFAICT the error just includes the error message tied to the exception itself; it shouldn't include a traceback with it.

PR #14152, opened 2023-10-31
Closes #13427: Public model registration

### LINKED ISSUE #13427: Devise a mechanism for referencing all user-facing models
### Proposed Changes

Devise and implement a mechanism for filtering the content types for all user-facing models in NetBox. This includes essentially all of the models which can be manipulated directly via the UI or API, and excludes those "behind the scenes" models which cannot.

### Justification

There are several instances where we currently work around this limitation. Below are some that I have identified; there are likely others to address as well.

1. Retrieving content types for e.g. the object list dashboard widget:

```python
# extras/dashboard/widgets.py
def get_content_type_labels():
    return [
        (content_type_identifier(ct), content_type_name(ct))
        for ct in ContentType.objects.filter(
            FeatureQuery('export_templates').get_query() | Q(app_label='extras', model='objectchange') |
            Q(app_label='extras', model='configcontext')
        ).order_by('app_label', 'model')
    ]
```

2. Limiting `object_type` choices on CustomField import:

```python
# extras/forms/bulk_import.py
class CustomFieldImportForm(CSVModelForm):
    ...
    object_type = CSVContentTypeField(
        label=_('Object type'),
        queryset=ContentType.objects.all(),
        limit_choices_to=FeatureQuery('custom_fields'),  # Should be generic
        required=False,
        help_text=_("Object type (for object or multi-object fields)")
    )
```

3. Filtering for SavedFilter objects:

```python
# extras/forms/filtersets.py
class SavedFilterFilterForm(SavedFiltersMixin, FilterForm):
    content_types = ContentTypeMultipleChoiceField(
        label=_('Content types'),
        queryset=ContentType.objects.filter(FeatureQuery('export_templates').get_query()),  # Should be generic
        required=False
    )
```

4. Limiting `object_type` choices on CustomField edit:

```python
# extras/forms/model_forms.py
class CustomFieldForm(BootstrapMixin, forms.ModelForm):
    object_type = ContentTypeChoiceField(
        label=_('Object type'),
        queryset=ContentType.objects.all(),
        # TODO: Come up with a canonical way to register suitable models
        limit_choices_to=FeatureQuery('webhooks').get_query() | Q(app_label='auth', model__in=['user', 'group']),
        required=False,
        help_text=_("Type of the related object (for object/multi-object fields only)")
    )
```

jeremystretch:
### Closes: #13427

- Record any models not having `_netbox_private = True` in `registry['models']`
- Introduce a proxy model for ContentType with a custom ContentTypeManager
- ContentTypeManager provides a `public()` method which returns all ContentTypes matching public models
- Update imports of ContentType to use the proxy model where `public()` is needed


docs/development/application-registry.md
```diff
@@ -41,6 +41,10 @@ A dictionary of particular features (e.g. custom fields) mapped to the NetBox mo
 
 Supported model features are listed in the [features matrix](./models.md#features-matrix).
 
+### `models`
```
arthanson, line 44: I think this is a bit confusing and should be stated the opposite - you need to set _netbox_private to True to make it non public.

netbox/extras/dashboard/widgets.py
```diff
@@ -37,10 +35,7 @@
 def get_content_type_labels():
     return [
         (content_type_identifier(ct), content_type_name(ct))
-        for ct in ContentType.objects.filter(
-            FeatureQuery('export_templates').get_query() | Q(app_label='extras', model='objectchange') |
-            Q(app_label='extras', model='configcontext')
-        ).order_by('app_label', 'model')
+        for ct in ContentType.objects.public().order_by('app_label', 'model')
```
arthanson, line 38: The public query looks like it is opening up to a wider number of models then what these existing queries do?  For example this one is does it have objectchange but the new one is not marked private, especially for plugins I think this could be different?   Not sure what effect that would have.

arthanson (changes requested):

netbox/extras/dashboard/widgets.py
jeremystretch: The use of `public()` does return some additional object types as opposed to the original query, but I think that's ultimately what we want. The old query was pieced together by including things we identified as desirable, versus excluding those that we know are not.

PR #14154, opened 2023-10-31
Closes #14134: Display additional object attributes in global search results

### LINKED ISSUE #14134: Display additional object attributes in global search results
### NetBox version

v3.6.4

### Feature type

### Proposed functionality

This FR proposes extending NetBox's global search feature to include certain additional object attributes when display global search results. (This does not impact model-specific filters.) Several extensions are necessary to implement this in a feasible manner.

screenshot

#### 1. Define display attributes on each SearchIndex class

The `display_attrs` attribute will define the additional attributes (if any) to include alongside search results. For example:

```python
@register_search
class DeviceIndex(SearchIndex):
    model = models.Device
    fields = (
        ('asset_tag', 50),
        ('serial', 60),
        ('name', 100),
        ('description', 500),
        ('comments', 5000),
    )
    display_attrs = ('site', 'location', 'rack', 'description')
```

This tells NetBox to also include the site, location, rack, and description for each global search result returning a device. Attributes will be included only if the object has a non-empty value for the specified field.

Although these attributes don't directly impact the search engine itself, IMO we should set them on the indexers (as opposed to the models) to maintain a clean API.

#### 2. Extend search backend to support prefetching of display attributes

Prefetching related objects is necessary to ensure a reasonably performant implementation. Using the example above, each matching device would trigger up to three discrete SQL queries (on each for site, location, and device) without prefetching enabled. (This is necessary only for related objects: Concrete fields will have already been populated while retrieving the object associated with each result.)

We should be able to employ Django's [`prefetch_related_objects()`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.prefetch_related_objects) utility function to attach prefetched objects to the results list after it has been retrieved. However, because global search results are heterogeneous, we must manually attach prefetched objects per result type. This will require calling `prefetch_related_objects()` on subsets of the results list organized by content type.

#### 3. Determine how to display the additional attributes

Assuming we can efficiently retrieve the attributes, we still need to determine the best way to incorporate them into the search results. I see several options:

1. Introduce a single column to hold all attributes per result.
2. List all results in a separate row beneath each result.
3. Dynamically add a table column for each attribute across all results.

There are pros and cons to each approach, however I don't consider option 3 truly viable, as it could generate very large tables when many types of objects are returned, and would be very difficult for the user to read. Option 1 (shown above) is probably the simplest.

### Use case

These attributes are intended to convey additional context to the user where readily distinguishing between similar search results may be difficult. For example, multiple results might each reference a "VLAN 100," each of which is assigned to a different site. This enhancement provides the context necessary for the user to see the site to which each VLAN is assigned without needing to navigate to each object individually.

### Database changes

I don't believe any changes to the database schema are necessary to support this functionality. We probably won't even need to trigger a re-indexing of cached values.

### External dependencies

None

jeremystretch:
### Closes: #14134

- Add `display_attrs` attribute to SearchIndex
- Add `display_attrs` property on CachedValue to render object attributes
- Extend CachedValueSearchBackend to support prefetching related objects

### DIFF
--- a/netbox/extras/models/search.py
+++ b/netbox/extras/models/search.py
@@ -4,7 +4,9 @@ from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
+from netbox.registry import registry
 from utilities.fields import RestrictedGenericForeignKey
+from utilities.utils import content_type_identifier
 from ..fields import CachedValueField
 
 __all__ = (
@@ -56,3 +58,16 @@ class CachedValue(models.Model):
 
     def __str__(self):
         return f'{self.object_type} {self.object_id}: {self.field}={self.value}'
+
+    @property
+    def display_attrs(self):
+        indexer = registry['search'].get(content_type_identifier(self.object_type))
+        attrs = {}
+        for attr in getattr(indexer, 'display_attrs', []):
+            name = self.object._meta.get_field(attr).verbose_name
+            if value := getattr(self.object, attr):
+                if display_func := getattr(self.object, f'get_{attr}_display', None):
+                    attrs[name] = display_func()
+                else:
+                    attrs[name] = value
+        return attrs
--- a/netbox/extras/models/search.py
+++ b/netbox/extras/models/search.py
@@ -4,6 +4,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
+from netbox.search.utils import get_indexer
 from netbox.registry import registry
 from utilities.fields import RestrictedGenericForeignKey
 from utilities.utils import content_type_identifier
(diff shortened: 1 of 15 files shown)


netbox/extras/models/search.py
```diff
@@ -56,3 +59,19 @@ class Meta:
...
         return f'{self.object_type} {self.object_id}: {self.field}={self.value}'
+
+    @property
+    def display_attrs(self):
+        """
+        Render any display attributes associated with this search result.
+        """
+        indexer = get_indexer(self.object_type)
+        attrs = {}
+        for attr in indexer.display_attrs:
```
arthanson, line 70: Does this handle if display_attrs doesn't exist (for example in an old plugin)?

arthanson (changes requested):
I think display_attrs needs to be added to documentation https://docs.netbox.dev/en/stable/plugins/development/search/

netbox/extras/models/search.py
jeremystretch: Indexers should always inherit from NetBox's base `SearchIndex` class, which defines `display_attrs`. IMO we should mandate this inheritance.

arthanson (approved):

PR #14167, opened 2023-11-02
Update attr in conditions example

### LINKED ISSUE #14164: Update Reference>Conditions example code after feature #9876 addition
### Change Type

Correction

### Area

Other

### Proposed Changes

With the addition of feature #9876 to handle conditional tag checks by parsing the `tags` list of dictionaries, the `attr` key in a conditional check requires a key in the tag dict to check against. For example, if checking for a tag with the slug `exempt`, the check would be:

`{
      "attr": "tags.slug",
      "value": "exempt",
      "op": "contains"
    }`

The [current conditional example](https://docs.netbox.dev/en/stable/reference/conditions/) in the documentation appears to have the intention of checking for the presence of a tag, though does not currently work correctly with only `"attr": "tags` and no deeper key selection.

kenny-y-dev:
### Fixes: #14164

Small update to conditions example documentation to allow successful check for presence of tag

### DIFF
--- a/docs/reference/conditions.md
+++ b/docs/reference/conditions.md
@@ -116,7 +116,7 @@ Multiple conditions can be combined into nested sets using AND or OR logic. This
       ]
     },
     {
-      "attr": "tags",
+      "attr": "tags.slug",
       "value": "exempt",
       "op": "contains"
     }


abhi1693 (approved):

jeremystretch:
Thanks @kenny-y-dev!

PR #14175, opened 2023-11-02
Fix #12731: Add custom model validation from forms and serializers

### LINKED ISSUE #12731: Custom validators unable to validate tag presence
### NetBox version

v3.5.1

### Python version

3.10

### Steps to Reproduce

1. Set the Custom Validators setting to validate tag requirements
```
CUSTOM_VALIDATORS = {
        "ipam.ipaddress": [{"tags": {"required": True}}],
        "dcim.site": [{"tags": {"required": True}}],
        "dcim.region": [{"tags": {"required": True}}]
}
```
2. Create a new object defined in the list above with/without tags defined.

_(This error is not exclusive to those object types)_

### Expected Behavior

I expected NetBox to validate if that particular object has tags set.

### Observed Behavior

```
<class 'ValueError'>

IPAddress objects need to have a primary key value before you can access their tags.

Python version: 3.10.6
NetBox version: 3.5.1
```

I'm not sure if I'm asking too much from the Custom Validators here, or if another method to enforce the existence of Tags is available.

Urth:
### Fixes: #12731

This allows for validation of any m2m relation like tags. The CustomValidator receives the cleaned data from the form/serializer. The model instance is not consistently available in the model forms/serializers so is omitted from both.

Note: the available fields can vary per form/serializer for a model. You should always check if a field is defined in the cleaned data. Alternatively the form/serializer class could be given to the validator to help identify the calling stack.

The patch extends the CustomValidator class with 3 validation functions, documentation is still needed.

```python
    def validate_data(self, data):                                              
        """                                                                     
        Custom validation method for model forms and model serializers, to be overridden by the user.
        Validation failures should raise a ValidationError exception.           
        """                                                                     
        return                                                                  
                                                                                
    def validate_form_data(self, data):                                         
        """                                                                     
        Custom validation method for model forms, to be overridden by the user. 
        Validation failures should raise a ValidationError exception.           
        """                                                                     
        return self.validate_data(data)                                         
                                                                                
    def validate_serializer_data(self, data):                                   
        """                                                                     
        Custom validation method for model serializers, to be overridden by the user.
        Validation failures should raise a ValidationError exception.           
        """                                                                     
        return self.validate_data(data)
```

These functions allow users to validate M2M relations with regular attributes in the form/serializer validation stage. Using this system you can make exclusive tags based on model attributes or validate combinations of M2M relations 
```python
class FooSiteValidator(CustomValidator):
    def validate_data(self, data):
        if data['name'] != 'foo':
            for tag in data['tags']:
                if tag.name == 'FOO':
                    self.fail('FOO tag is reserved for site foo', 'tags')
```

### DIFF
--- a/netbox/extras/signals.py
+++ b/netbox/extras/signals.py
@@ -9,7 +9,7 @@ from django_prometheus.models import model_deletes, model_inserts, model_updates
 from extras.validators import CustomValidator
 from netbox.config import get_config
 from netbox.context import current_request, webhooks_queue
-from netbox.signals import post_clean
+from netbox.signals import post_clean, post_form_clean, post_serializer_clean
 from utilities.exceptions import AbortRequest
 from .choices import ObjectChangeActionChoices
 from .models import ConfigRevision, CustomField, ObjectChange, TaggedItem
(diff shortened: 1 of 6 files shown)


jeremystretch:
@Urth thanks for your work on this. I believe I've identified a simpler approach that doesn't require multiple validation paths in #14516. I have still have some work to do on that, but will continue down this path as it appears very promising.

I'm going to close out this PR for now, but if the alternative approach doesn't pan out I'm happy to revisit.

Issue #14173, opened 2023-11-02
Allow plugins to add columns to tables


rixx:
### NetBox version

dev / v3.6.3

### Feature type

New functionality

### Proposed functionality

Allow plugins to register columns for built-in NetBox tables.

### Use case

NetBox comes with admirably configurable tables and filters for all its built-in models, and makes it easy for plugins to provide tables like that, too. Some plugins manage fairly separate data, like DNS. However, when plugins *connect* to built-in models (instead of just providing separate models), the additional data cannot be seen from the built-in tables.

This is especially jarring for users when the data *can* be shown on the detail pages (by way of the `template_extensions` mechanism). Naturally, users tend to ask for ways to use the data displayed on the detail pages in the list/table views. As a plugin developer, I frequently build models that attach to existing NetBox models, and I get those requests a lot ("why do I have to go to the detail page of the Tenant to see the new data, can't you show it in the Tenant list?").

(I realise that there are *technically* ways to add columns to the tables right now: You can subclass/extend the built-in tables and provide them at a different URL, or, worst-of-all-worlds, patch something in. However, I think both of these are terrible in terms of maintainability and generally not good practice.)

An example could be a plugin that extends the customer management side of NetBox with additional data on Contacts, Tenants, or Sites, and would like to a) show the added data in those tables (e.g. the Site table), but also in other tables that can show Site information, e.g. the Device table.

I realise this isn't a trivial request. If you think this is generally worthwhile, I'd be happy to work on this feature!

jeremystretch:
Thank you for your interest in extending NetBox. Unfortunately, the information you have provided does not constitute an actionable feature request. Per our [contributing guide](https://github.com/netbox-community/netbox/blob/develop/CONTRIBUTING.md), a feature request must include a thorough description of the proposed functionality, including any database changes, new views or API endpoints, and so on. It must also include a detailed use case justifying its implementation. If you would like to elaborate on your proposal, please modify your post above. If sufficient detail is not added, this issue will be closed.

rixx:
Apologies – updated the ticket with a longer description.

ITJamie:
to give a very specific example of how this could be used for the netbox-gateways plugin.

the plugin could "register" that It should show related gateways objects on the prefix core module.
so it would add a "gateway" column to the prefix tables and maybe even a "gateway" field on the prefix API serializer for general API use.

[...]

PR #14191, opened 2023-11-06
Closes #14153: Filter ContentTypes by supported feature

### LINKED ISSUE #14153: Replace FeatureQuery with a custom manager method for ContentType
### Proposed Changes

NetBox uses the FeatureQuery class to filter ContentType querysets by support for a specific feature. For example, to retrieve the content types corresponding to all models which support the assignment of jobs:

```python
queryset=ContentType.objects.filter(FeatureQuery('jobs').get_query())
```

This proposal is to move this logic to a custom manager method for ContentType, such that we could instead do:

```python
queryset=ContentType.objects.with_feature('jobs')
```

In addressing #13427, PR #14152 proposes introduce a proxy model and custom manager for ContentType, which would be easily extended to support this implementation.

### Justification

FeatureQuery has always felt cumbersome to use, and is not invoked consistently for all use cases (i.e. sometimes `get_query()` is needed and sometimes not). The proposed approach provides a much cleaner mechanism for filtering ContentTypes by supported feature.

jeremystretch:
### Fixes: #14153

- Add a `with_feature()` method to the manager for our ContentType proxy model
- Replace all uses of FeatureQuery with the new method
- Remove FeatureQuery
- Update all ForeignKey references to ContentType to use a string name (rather than the class itself) to simplify imports and avoid migration errors with GenericForeignKeys

### DIFF
--- a/netbox/core/models/contenttypes.py
+++ b/netbox/core/models/contenttypes.py
@@ -21,6 +21,24 @@ class ContentTypeManager(ContentTypeManager_):
             q |= Q(app_label=app_label, model__in=models)
         return self.get_queryset().filter(q)
 
+    def with_feature(self, feature):
+        """
+        Return the ContentTypes only for models which are registered as supporting the specified feature. For example,
+        we can find all ContentTypes for models which support webhooks with
+
+            ContentType.objects.with_feature('webhooks')
+        """
+        if feature not in registry['model_features']:
+            raise KeyError(
+                f"{feature} is not a registered model feature! Valid features are: {registry['model_features'].keys()}"
+            )
+
+        q = Q()
+        for app_label, models in registry['model_features'][feature].items():
+            q |= Q(app_label=app_label, model__in=models)
+
+        return self.get_queryset().filter(q)
+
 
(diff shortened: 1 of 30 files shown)


netbox/core/models/contenttypes.py
```diff
@@ -21,6 +21,24 @@ def public(self):
             q |= Q(app_label=app_label, model__in=models)
         return self.get_queryset().filter(q)
 
+    def with_feature(self, feature):
```
arthanson, line 24: Not a part of this PR, but I think we should have a FR to document the registry as it is getting a bit complicated now, for example for a model-feature what it is for, you need to do the following steps (like in the clean validate the content-type), etc...

arthanson (approved):

netbox/core/models/contenttypes.py
jeremystretch: We have a page in [the developer docs](https://docs.netbox.dev/en/stable/development/application-registry/) that explains each store. Happy to flesh this out some. (And your comment reminded me to go add the new `tables` store introduced under #14173.)

PR #14196, opened 2023-11-06
Closes #13794: Dynamically populate related objects list under tenant view

### LINKED ISSUE #13794: Make the "Related Objects" pane in the detail view for "Tenant" extensible
### NetBox version

v3.6.1

### Feature type

### Proposed functionality

On the right side of the detail view for a "Tenant" object there is a list of NetBox core models and object counts the tenant is assigned to. Currently the list of models in this view is hardcoded in `tenancy.TenantView.get_extra_content()`.

It would be helpful for plugins supporting tenancy if that list could be extended, e.g. by providing a decorator that could be applied to models using tenancy similar to the mechanism used for the global search index.

### Use case

In many cases it makes sense for plugins to use tenancy to indicate the ownership of or responsibility for objects, making it easier to filter for objects related, for instance, to a specific customer or department. 

While tenancy itself is easy to implement, this does not reflect in the list of related objects for a tenant. The detail view for a tenant only shows NetBox core objects that are assigned to the tenant, not objects of plugin models. 

This is especially annoying when a user wants to delete a tenant object, which is usually related to the target object using the referential integrity option `models.PROTECT`. So even if there are no related objects shown in the tenant view, the tenant still can't be deleted because there are unlisted objects still owned by it in the plugin's models.

### Database changes

Unknown, probably none - depending on how this can be implemented.

### External dependencies

None.

jeremystretch:
### Fixes: #13794

- Introduce the `get_related_models()` utility function
- Replace the static list of related models

### DIFF
--- a/netbox/tenancy/views.py
+++ b/netbox/tenancy/views.py
@@ -2,14 +2,9 @@ from django.contrib.contenttypes.models import ContentType
 from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext as _
 
-from circuits.models import Circuit
-from dcim.models import Cable, Device, Location, PowerFeed, Rack, RackReservation, Site, VirtualDeviceContext
-from ipam.models import Aggregate, ASN, IPAddress, IPRange, L2VPN, Prefix, VLAN, VRF
 from netbox.views import generic
-from utilities.utils import count_related
+from utilities.utils import count_related, get_related_models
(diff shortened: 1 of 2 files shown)


jeremystretch:
Per a conversation with @arthanson, we should look into ordering these by app.

arthanson (approved):

PR #14208, opened 2023-11-08
Adds unit to the power port draw

### LINKED ISSUE #13587: Add unit to power port overview on device view
### NetBox version

v3.5.8

### Feature type

New functionality

### Proposed functionality

On the power ports view within a device, add the unit to the Maximum draw and Allocated draw. 
From this:
Screenshot 2023-08-28 at 21 14 29

To this
Screenshot 2023-08-28 at 21 14 29

### Use case

Just makes it more apparent for users that the unit of power is W. 

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -466,13 +466,11 @@ class PowerPortTable(ModularDeviceComponentTable, PathEndpointTable):
             'args': [Accessor('device_id')],
         }
     )
-    maximum_draw = columns.TemplateColumn(
-        template_code='''{% if record.maximum_draw %}{{ value }} W{% endif %}''',
-        verbose_name=_('Maximum draw')
+    maximum_draw = tables.Column(
+        verbose_name=_('Maximum draw (watts)')
     )
-    allocated_draw = columns.TemplateColumn(
-        template_code='''{% if record.allocated_draw %}{{ value }} W{% endif %}''',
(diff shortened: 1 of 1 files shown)


netbox/dcim/tables/devices.py
```diff
@@ -466,6 +466,14 @@ class PowerPortTable(ModularDeviceComponentTable, PathEndpointTable):
             'args': [Accessor('device_id')],
         }
     )
+    maximum_draw = tables.TemplateColumn(
```
jeremystretch, line 469: We can use `netbox.tables.columns.TemplateColumn` instead of the stock column to automatically render the placeholder for empty values, which allows us to simplify the template code.

netbox/dcim/tables/devices.py
```diff
@@ -466,6 +466,14 @@ class PowerPortTable(ModularDeviceComponentTable, PathEndpointTable):
...
+        template_code='''{% if record.maximum_draw %}{{ record.maximum_draw }} W{% else %}&mdash;{% endif %}''',
```
jeremystretch, line 470: We should be able to reference `value` here instead of `record.maximum_draw` explicitly.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @abhi1693!

PR #14211, opened 2023-11-08
Adds copy button to ip addresses in interface list

### LINKED ISSUE #13697: add copy ip button to ips in `Ip Addresses` column on interface lists
### NetBox version

v3.6.0

### Feature type

### Proposed functionality

when looking at a list of interfaces on a device / VM it would be nice to have the quick copy IP button that has appeared elsewhere recently 

### Use case

ability to copy ip from device/vm interface list without having to select first goto the IP object page

### Database changes

n/a

### External dependencies

n/a

abhi1693:
### Fixes: #13697

I did my best with the UI, but requires a bit more refactor to reduce the horizontal margins. @jeremystretch feel free to adjust if you feel it is required.

### DIFF
--- a/netbox/dcim/tables/template_code.py
+++ b/netbox/dcim/tables/template_code.py
@@ -35,14 +35,26 @@ DEVICEBAY_STATUS = """
 """
 
 INTERFACE_IPADDRESSES = """
-<div class="table-badge-group">
-  {% for ip in value.all %}
-    {% if ip.status != 'active' %}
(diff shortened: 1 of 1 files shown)


jeremystretch:
IMO we should change this to a ManyToManyColumn, and look at implementing copy-to-clipboard support for columns in a more general sense rather than on a case-by-case basis. This would probably entail introduce a new base column class to support a keyword argument like `copy_to_clipboard=True`.

abhi1693:
|> IMO we should change this to a ManyToManyColumn

Are you fine with removing the badge logic that adds coloring and tooltip to the value? This maybe seen as a breaking change if the fuctionality is removed.

jeremystretch:
I personally am, but I'm open to other opinions. It currently renders with excessive whitespace, so it needs ultimately needs to be cleaned up one way or another.

PR #14212, opened 2023-11-08
Adds inventory items to interface

### LINKED ISSUE #13723: Enable Inventory Item column in Device's Interface tab
### NetBox version

v3.5.4

### Feature type

### Proposed functionality

The inventory Items tab inside the device allows to show the interface to which it is linked (column Component)
The proposed functionality is to also enable as a column option the other way around; showing the inventory item linked to the interface

### Use case

Currently in our case, we use a custom field to show the item linked to the interface. Allowing this column would allow a more synchronized data and to be shown in a quick glance.
image

### Database changes

None

### External dependencies

None

abhi1693:


jeremystretch (approved):
Thanks @abhi1693!

PR #14213, opened 2023-11-08
Adds region hierarchy in templates

### LINKED ISSUE #13735: Show full Region hierarchy on all relevant object views
### NetBox version

v3.6.1

### Feature type

### Proposed functionality

There seems to be some inconsistency with the way that the Region information is displayed in different object views. For a Device, there is a separate Region row shown in the UI with the full Region path. But, for Circuit Termination, Prefix, Rack, Rack Reservation, and VLAN (this is everything I could find with `grep -F 'site.region|linkify'`, so maybe I missed something), only the leaf Region is shown on the Site row in the respective object view. My proposal would be to align the other models' object views to the way that Device does it, but I'm also fine with the regions staying on the Site line, as long as they show the full path, and ideally are consistent across all the object views.

I also noticed that models like Power Panels and Virtualization Clusters show the Site and no Region information. I'm not sure how to gather an exhaustive list for that scenario, but I feel like it should be consistent as well. Maybe there's a better way to DRY the site/location/rack and related site groups and regions into a separate template for consistency (understanding that location/rack may not be relevant to all models that support site)?

### Use case

We are trying to implement a Region hierarchy that follows a pattern like `State / City` (which I think is fairly common) and it seems redundant to have to include the state as part of the city name, since the Region model only requires the name to be unique to its parent. There is also inconsistency today between how different models represent the same data in the UI, so that creates some visual dissonance. There are other areas of the UI that currently don't handle nested models very well, like table columns and dropdowns, but that's a separate issue.

### Database changes

None

### External dependencies

None

abhi1693:


netbox/dcim/templatetags/display_region.py
```diff
@@ -0,0 +1,37 @@
...
+def _display_site(obj):
+    """
+    Render a link to the site of an object.
+    """
+    site = getattr(obj, 'site', None)
+    return mark_safe('<a href="{}">{}</a>'.format(site.get_absolute_url(), site)) if site else None
+
+
+@register.simple_tag(takes_context=True)
+def display_region(context, obj, include_site=False):
```
jeremystretch, line 16: IMO we should omit any reference to a site, and limit this tag to only being concerned with rendering the region hierarchy. For example, something like `{% region_tree region %}` might render "US > North Carolina > Raleigh" for the Raleigh region.

In fact, it might be useful to turn this into a general-purpose tag to render other MPTT objects in the same manner.

abhi1693: To keep the current scope limited, I'll update the relevant UI and add a simple region tree tag. We can make it generic as a part of a separate FR.

jeremystretch: I agree on limiting the implementation scope for this FR, but we should give it a generic name now to avoid having to update it later.

abhi1693:
@jeremystretch Are there any other changes you'd like in here?

jeremystretch:
@abhi1693 I've tweaked this to allow its use for more than just regions.

jeremystretch (approved):

PR #14214, opened 2023-11-08
Adds site column to power feeds

### LINKED ISSUE #13743: Add "site" column to power feed list
### NetBox version

v3.6.1

### Feature type

### Proposed functionality

I propose adding "Site" as a selectable column for the power feed view. The power feed model doesn't directly have a "Site" field, so the "Site" column would be populated from the associated Rack.

### Use case

When looking at a list of power feeds, it's helpful to, at a glance, know what site the power feed belongs to. This is especially useful if you have duplicate rack names across sites.

### Database changes

None.

### External dependencies

None.

abhi1693:

### DIFF
--- a/netbox/dcim/tables/power.py
+++ b/netbox/dcim/tables/power.py
@@ -87,8 +87,9 @@ class PowerFeedTable(TenancyColumnsMixin, CableTerminationTable):
         linkify=True,
         verbose_name=_('Tenant')
     )
-    site = tables.TemplateColumn(
-        template_code='''{% if record.rack %}{{ record.rack.site|linkify }}{% else %}&mdash;{% endif %}''',
+    site = tables.Column(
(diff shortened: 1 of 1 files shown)


netbox/dcim/tables/power.py
```diff
@@ -87,6 +87,10 @@ class PowerFeedTable(TenancyColumnsMixin, CableTerminationTable):
         linkify=True,
         verbose_name=_('Tenant')
     )
+    site = tables.TemplateColumn(
```
jeremystretch, line 90: We should be able to do this with a regular column by setting its accessor to `rack__site` and `linkify=True`.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @abhi1693!

PR #14216, opened 2023-11-08
Adds parent to inventory item table

### LINKED ISSUE #14113: inventory item list - add parent column
### NetBox version

v3.6.4

### Feature type

### Proposed functionality

when viewing the list of inventory items (https://netboxinstance/dcim/inventory-items/) please add "parent" as one of the column choices

### Use case

there is currently no option to add "parent" to the displayed list of inventory items. 

note: parent is also missing from the csv export, but is supported on csv import

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -933,7 +933,8 @@ class InventoryItemTable(DeviceComponentTable):
     discovered = columns.BooleanColumn(
         verbose_name=_('Discovered'),
     )
-    parent = tables.LinkColumn(
+    parent = tables.Column(
+        linkify=True,
         verbose_name=_('Parent'),
     )
     tags = columns.TagColumn(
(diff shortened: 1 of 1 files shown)


netbox/dcim/tables/devices.py
```diff
@@ -933,6 +933,9 @@ class InventoryItemTable(DeviceComponentTable):
     discovered = columns.BooleanColumn(
         verbose_name=_('Discovered'),
     )
+    parent = tables.LinkColumn(
```
jeremystretch, line 936: FYI LinkColumn has been [deprecated](https://django-tables2.readthedocs.io/en/latest/pages/api-reference.html#django_tables2.columns.LinkColumn): The recommendation is now to use `Column(linkify=True)` instead.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @abhi1693!

PR #14217, opened 2023-11-08
Adds inventory item children view

### LINKED ISSUE #14112: Inventory Item view - show child objects on parent
### NetBox version

v3.6.4

### Feature type

### Proposed functionality

when viewing an inventory item there should be a table listing child objects of that inventory item

### Use case

when documenting inventory items (eg ssd's, in a raid controller) it would be great if the child items showed in a list when viewing the parent inventory item.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:


jeremystretch (approved):
Thanks @abhi1693!

PR #14218, opened 2023-11-08
Adds mask length filters on ipaddress

### LINKED ISSUE #14101: `mask_length__(lte|gte)` filter for IP addresses
### NetBox version

v3.6.4

### Feature type

New functionality

### Proposed functionality

The `ipam/prefixes` API endpoint has the `mask_length__gte` and `mask_length__lte` filters while the `ipam/ip-addresses` endpoint only has `mask_length`. I propose to add the `gte` and `lte` filters to the IP addresses endpoint as well.

I can make a PR for this if accepted.

### Use case

I would like to get all point to point interfaces (interfaces with an address with a mask length >= 30 for IPv4 or >= 126 for IPv6).

### Database changes

None

### External dependencies

None

abhi1693:

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -266,7 +266,8 @@ class PrefixFilterSet(NetBoxModelFilterSet, TenancyFilterSet):
     )
     mask_length = MultiValueNumberFilter(
         field_name='prefix',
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @abhi1693!

PR #14220, opened 2023-11-08
Order available columns

### LINKED ISSUE #14219: Order available columns alphabetically in table configuration dialog
### NetBox version

v3.6.4

### Feature type

### Proposed functionality

When displaying the list of available columns (at left) under the table configuration form, columns should be listed alphabetically. There is currently no ordering in place:

screenshot

### Use case

This change will make it easier for users to located desired columns in the list.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:


jeremystretch (approved):
Thanks @abhi1693!

PR #14226, opened 2023-11-09
Fix ordering on JobTable. #14223

### LINKED ISSUE #14223: Sorting Jobs table by object causes FieldError
### NetBox version

v3.6.4

### Python version

3.8

### Steps to Reproduce

1. Browse to /core/jobs
2. Click on the 'object' heading in the table, observe nothing happens (HTMX request results in 500 server error, but this is not visible - see below)
3. Try and reload the page, observe that an exception is thrown and all further attempts to view the Job listing page results in the same exception:

> <class 'django.core.exceptions.FieldError'>
>
>Field 'object' does not generate an automatic reverse relation and therefore cannot be used for reverse querying. If it is a GenericForeignKey, consider adding a GenericRelation.
>
>Python version: 3.8.10
>NetBox version: 3.6.4
>Plugins: 
>  netbox_demo: 0.3.1

When attempting to sort by object, Django will throw a FieldError meaning the HTMX request appears to do nothing. The error can be observed in the browser console. Subsequently, the FieldError becomes visible when reloading the page, because sorting by object is now stored in the user's preferences.

In case anyone stumbles across this, the fix is to remove the table configuration for the JobTable from the user preferences or remove the ordering by appending ?sort= to the /core/jobs URL.

### Expected Behavior

The Jobs list should either return a list sorted by the object, or shouldn't allow the table to be sorted by the object field at all.

### Observed Behavior

A FieldError exception is raised.

cpmills1975:
### Fixes: #14223 

Set orderable=False on object column in JobTable

### DIFF
--- a/netbox/core/tables/jobs.py
+++ b/netbox/core/tables/jobs.py
@@ -19,7 +19,8 @@ class JobTable(NetBoxTable):
     )
     object = tables.Column(
         verbose_name=_('Object'),
-        linkify=True
+        linkify=True,
+        orderable=False
     )
     status = columns.ChoiceFieldColumn(
         verbose_name=_('Status'),


jeremystretch (approved):
Thanks @cpmills1975!

PR #14229, opened 2023-11-09
Fix asn view under asn range

### LINKED ISSUE #14195: Not Able to See ASN Tab within ASN Range Without Being a Superuser
### NetBox version

v3.6.4

### Python version

3.8

### Steps to Reproduce

1. Within the IPAM module, create an ASN range that encompasses other/existing ASNs
2. Open the ASN Range object as a user that does NOT have superuser permissions, but does have all other permissions

### Expected Behavior

A user with ALL permissions (at least all IPAM permissions) should see an "ASNs" tab within the ASN Range object that lists all ASNs within the range.

### Observed Behavior

Without superuser permissions, the "ASNs" tab is not shown.

abhi1693:

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -220,7 +220,7 @@ class ASNRangeASNsView(generic.ObjectChildrenView):
     tab = ViewTab(
         label=_('ASNs'),
         badge=lambda x: x.get_child_asns().count(),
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @abhi1693!

PR #14231, opened 2023-11-09
Adds contact group on contact assignment table

### LINKED ISSUE #14221: Add contact group to table view on object
### NetBox version

v3.6.4

### Feature type

New functionality

### Proposed functionality

Add the column Contact Group to the table view on objects detailed view.

### Use case

For certain objects we use a contact to point out that the device/virtual machine is "owned" by this group. As such we wanted to name this contact the same name, but just add it to different contact groups. But since the table view on a specific device does not show the contact group as an available column, we are not able to distinguish between each group's contact. Our workaround is just to add the group name to the contact name, but I think adding the Contact Group to the table view is a valid solution as well.

### Database changes

None

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/tenancy/tables/contacts.py
+++ b/netbox/tenancy/tables/contacts.py
@@ -102,6 +102,11 @@ class ContactAssignmentTable(NetBoxTable):
         verbose_name=_('Role'),
         linkify=True
     )
+    contact_group = tables.Column(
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @abhi1693!

PR #14232, opened 2023-11-09
Raises validation error if file path and root are not unique

### LINKED ISSUE #14187: attempting to add a script from a datasource that already exists throws a server error
### NetBox version

v3.6.4

### Python version

3.11

### Steps to Reproduce

1: create a datasource (eg local) with a script file in it
2: add the script (extras/scripts/add/)
3: attempt to add the same script again

### Expected Behavior

a validation error should show up (or the file should not be available in the dropdown list as its already been added)

### Observed Behavior

django server error is thrown

```

duplicate key value violates unique constraint "core_managedfile_unique_root_path"
DETAIL:  Key (file_root, file_path)=(scripts, ip_grabber.py) already exists.

Python version: 3.11.4
NetBox version: 3.6.4
Plugins: 
  netbox_circuitmaintenance: 0.3.0
  netbox_gateways: 0.4.3
  ``

### PR DESCRIPTION
**Pull request by abhi1693**

### DIFF
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -2,6 +2,7 @@ import inspect
 import logging
 from functools import cached_property
 
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: netbox/extras/models/scripts.py [hunk 1]
```diff
@@ -76,6 +77,14 @@ def _get_name(cls):
 
         return scripts
 
+    def clean(self):
```
jeremystretch, line 80: This should probably go on the ManagedFile class. We can tweak the method to reference the inheriting class itself so that there's no overlap between scripts and reports, for instance.

netbox/extras/models/scripts.py
```diff
@@ -76,6 +77,14 @@ def _get_name(cls):
...
+        super().clean()
+
+        # Ensure that the file root and path make a unique pair
+        if self.file_root and self.file_path:
+            if ScriptModule.objects.filter(file_root=self.file_root, file_path=self.file_path).exclude(pk=self.pk).exists():
```
jeremystretch, line 85:
```suggestion
            if self._meta.model.objects.filter(file_root=self.file_root, file_path=self.file_path).exclude(pk=self.pk).exists():
```

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @abhi1693!

PR #14233, opened 2023-11-09
Restores config revision during cache clear

### LINKED ISSUE #14182: Changelog retention defaults to 90 days after upgrade, all prior changelogs are lost.
### NetBox version

v3.6.4

### Python version

3.8

### Steps to Reproduce

1. Set the changelog retention days to other value than 90 days with a new configuration revision
2. Do upgrade as described in documentation [docs.netbox.dev](https://docs.netbox.dev/en/stable/installation/upgrading/)
3. Log in to your netbox instance and go to /extras/changelog and the changelog retention days have been defaulted to 90 days all prior logs are deleted.

### Expected Behavior

The changelog retention days should not default back to 90 days after upgrade. 

### Observed Behavior

/core/config/
image

/extras/changelog/
image

/extras/config-revisions/add/
image

abhi1693:

### DIFF
--- a/netbox/core/management/commands/clearcache.py
+++ b/netbox/core/management/commands/clearcache.py
@@ -1,11 +1,20 @@
 from django.core.cache import cache
 from django.core.management.base import BaseCommand
 
+from extras.models import ConfigRevision
+
 
 class Command(BaseCommand):
     """Command to clear the entire cache."""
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @abhi1693!

PR #14235, opened 2023-11-09
Handles the port in the ip

### LINKED ISSUE #14085: Fix for header IP with source port
### NetBox version

v3.6.4

### Feature type

### Proposed functionality

Based on a [Slack Discussion](https://netdev-community.slack.com/archives/C01P0FRSXRV/p1697600102751449), I'm submitting what could be a half bug / half feature request/issue... 

When NetBox is behind an IIS reverse proxy, it appears that the header information is different. We can use it without any issue, except if a user enter a wrong credential set, which results in an error message: 
image

The user IP returned has source port, which breaks the `get_client_ip` function defined in [netbox/utilities/request.py ](https://github.com/netbox-community/netbox/blob/5b2f29480a42e9c815ca1f50403c115cc89b6538/netbox/utilities/request.py)

I worked around the issue by adding a single line of code:
`client_ip = client_ip.split(':')[0]` just before `try`

### Use case

To prevent error message, which lead the user to think that server is down instead of him entering a wrong password. 

### Database changes

none

### External dependencies

none

abhi1693:

### DIFF
--- a/netbox/utilities/request.py
+++ b/netbox/utilities/request.py
@@ -17,7 +17,7 @@ def get_client_ip(request, additional_headers=()):
     )
     for header in HTTP_HEADERS:
         if header in request.META:
-            client_ip = request.META[header].split(',')[0]
+            client_ip = request.META[header].split(',')[0].partition(':')[0]
             try:
                 return IPAddress(client_ip)
             except ValueError:


netbox/utilities/request.py
```diff
@@ -17,7 +17,7 @@ def get_client_ip(request, additional_headers=()):
     )
     for header in HTTP_HEADERS:
         if header in request.META:
-            client_ip = request.META[header].split(',')[0]
+            client_ip = request.META[header].split(',')[0].partition(':')[0]
```
jeremystretch, line 20: Just curious, any reason to use `partition()` instead of `split()`?

abhi1693: Both `partition()` and `split()` can be used interchangeably. `partition()` is generally used when we need to split a string at the first occurrence of a separator and we are sure there will be only one separator or are only interested in the first occurrence.

PR #14236, opened 2023-11-09
Allow login and logout in maintenance mode

### LINKED ISSUE #14166: Maintenance mode prevents login
### NetBox version

v3.6.4

### Python version

3.9

### Steps to Reproduce

This issue occurred this morning while I was attempting to upgrade my NetBox database from v3.2.8 to v3.6.4.  Having fixed the issue, I'm reluctant to try and reproduce it but believe the following set of conditions led to an inability for ANY user, even super users, to log in.

As part of my upgrade plan, I needed to upgrade the underlying OS so I built a whole new VM and installed NetBox v3.6.4.

* I enabled MAINTENANCE_MODE on the existing v3.2.8 instance using a config revision.
* I dumped the database using pg_dump and moved it to the new box via scp (I actually wrote a bash script to do this).
* I dropped the existing v3.6.4 database and imported the old data using psql and then ran the upgrade.sh process.
* After starting the netbox and netbox-rq services, I visited the site and attempted to log in using my LDAP backed user account.
* The login didn't fail, but I was immediately redirected back to the login page!
* I've followed this process repeatedly over the last couple of weeks while updating and testing plugins and scripts etc - netbox v3.6.4 has been running on this VM for some weeks, automatically synced to the v3.2.8 instance using the above process every night and it has always worked. The only difference this morning was that the old instance was in maintenance mode prior to migrating the data.
* Thinking that perhaps something was wrong with my LDAP auth, I created a local super user account (python netbox/manage.py createsuperuser). I was unable to log in with that account either.
* Thinking that perhaps nginx was causing the redirect, I stopped nginx and ran the NetBox server directly (python netbox/manage.py runserver 0.0.0.0:8000). This did exactly the same, but I was able to see log messages saying that the last login date wasn't updated due to the application being in maintenance mode.
* Because I couldn't log in, I couldn't change the config revision to disable MAINTENANCE_MODE so I added the option to the configuration.py file to override the contents of the config revision.
* Having then restarted the application (systemctl restart netbox netbox-rq), I was then able to log in successfully using my LDAP account and delete the config revision from the database.

It appears that enabling maintenance mode in the config revision, prevented any user from logging in.

### Expected Behavior

Users should be able to log in even when maintenance mode is enabled.

### Observed Behavior

With maintenance mode enabled via a config revision, I was unable to log in at all.

abhi1693:

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -502,6 +502,9 @@ AUTH_EXEMPT_PATHS = (
 MAINTENANCE_EXEMPT_PATHS = (
     f'/{BASE_PATH}admin/',
     f'/{BASE_PATH}extras/config-revisions/',  # Allow modifying the configuration
+    LOGIN_URL,
+    LOGIN_REDIRECT_URL,
+    LOGOUT_REDIRECT_URL
 )
 
 SERIALIZATION_MODULES = {


netbox/netbox/settings.py
```diff
@@ -502,6 +502,9 @@ def _setting(name, default=None):
 MAINTENANCE_EXEMPT_PATHS = (
     f'/{BASE_PATH}admin/',
     f'/{BASE_PATH}extras/config-revisions/',  # Allow modifying the configuration
+    LOGIN_URL,
+    LOGIN_REDIRECT_URL,
+    LOGOUT_REDIRECT_URL
```
jeremystretch, line 507: Do the redirect URLs need to be included? Only the actual login view should need to write to the database.

abhi1693: I added these thinking maybe it'll be required for 3rd party auth integrations

jeremystretch: Yeah that's fair, good call.

PR #14265, opened 2023-11-14
Closes #14173: Enable plugins to register columns on core tables

### LINKED ISSUE #14173: Allow plugins to add columns to tables

jeremystretch:
### Fixes: #14173

- Introduce `register_table_column()` utility function to register custom columns on existing tables
- Establish new `tables` registry key
- Add documentation & test

### DIFF
--- a/docs/plugins/development/tables.md
+++ b/docs/plugins/development/tables.md
@@ -96,12 +96,13 @@ Plugins can register their own custom columns on core tables using the `register
 
 ```python
 import django_tables2
+from django.utils.translation import gettext_lazy as _
 
 from dcim.tables import SiteTable
 from utilities.tables import register_table_column
 
 mycol = django_tables2.Column(
-    verbose_name='My Column',
+    verbose_name=_('My Column'),
     accessor=django_tables2.A('description')
 )
 
--- a/docs/plugins/development/tables.md
+++ b/docs/plugins/development/tables.md
@@ -87,3 +87,27 @@ The table column classes listed below are supported for use in plugins. These cl
     options:
       members:
         - __init__
+
+## Extending Core Tables
+
+!!! info "This feature was introduced in NetBox v3.7."
+
+Plugins can register their own custom columns on core tables using the `register_table_column()` utility function. This allows a plugin to attach additional information, such as relationships to its own models, to built-in object lists.
+
+```python
+import django_tables2
+
+from dcim.tables import SiteTable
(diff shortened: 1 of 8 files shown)


docs/plugins/development/tables.md
```diff
@@ -87,3 +87,27 @@ The table column classes listed below are supported for use in plugins. These cl
...
+
+Plugins can register their own custom columns on core tables using the `register_table_column()` utility function. This allows a plugin to attach additional information, such as relationships to its own models, to built-in object lists.
+
+```python
+import django_tables2
+
+from dcim.tables import SiteTable
+from utilities.tables import register_table_column
+
+mycol = django_tables2.Column(
```
arthanson, line 103: Just a question, now for best practices do we want to move people to making localizable plugins?  So update this to use the _( localized strings?

arthanson (approved):

docs/plugins/development/tables.md
jeremystretch: Yeah, that's a good call-out.

PR #14289, opened 2023-11-16
Closes #13299: Improve options for controlling custom field visibility

### LINKED ISSUE #13299: Custom Field UI - hidden if unset - independent of rw/ro
### NetBox version

v3.5.7

### Feature type

### Proposed functionality

I have many custom fields which only apply to a part of e.g. devices. With the "UI visibility" = "Hidden(if unset)" you can keep the UI clean for each device and you only see those that are set for the device. The problem and hence the feature reqest the "if unset" option should be independent of "Read/Write" or "Read only" option.

**UI visibility:**
- Read/Write
- Read/Write(if unset)
- Read-only
- Read-only(if unset)
- Hidden

or maybe with a checkbox

**UI visibility:**
- Read/Write
- Read-only
- Hidden

- [ ] Hidden if unset

### Use case

I have custom fields that are Ready-only but I don't want them to appear in the UI on every device as they have no meaning for those devices. So it would be good if I could select read-only and "hidden if unset". Maybe there is another way to achieve this, but I haven't found any.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #13299

- Add two new fields to the CustomField model: `ui_visible` and `ui_editable`
- Migrate existing custom fields, setting the appropriate values depending on the value of the legacy `ui_visibility` field
- Remove the `ui_visibility` field
- Update documentation

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -179,13 +179,6 @@ class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
         blank=True,
         null=True
     )
-    ui_visibility = models.CharField(
-        max_length=50,
-        choices=CustomFieldVisibilityChoices,
-        default=CustomFieldVisibilityChoices.VISIBILITY_READ_WRITE,
-        verbose_name=_('UI visibility'),
-        help_text=_('Specifies the visibility of custom field in the UI')
-    )
     ui_visible = models.CharField(
         max_length=50,
         choices=CustomFieldUIVisibleChoices,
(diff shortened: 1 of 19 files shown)


netbox/extras/models/customfields.py
```diff
@@ -177,12 +177,19 @@ class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
...
-        choices=CustomFieldVisibilityChoices,
-        default=CustomFieldVisibilityChoices.VISIBILITY_READ_WRITE,
-        verbose_name=_('UI visibility'),
-        help_text=_('Specifies the visibility of custom field in the UI')
+        choices=CustomFieldUIVisibleChoices,
+        default=CustomFieldUIVisibleChoices.ALWAYS,
+        verbose_name=_('UI visible'),
+        help_text=_('Specifies whether the custom field is displayed in the UI')
+    )
+    ui_editable = models.CharField(
```
arthanson, line 187: It looks like these two are bound together and it doesn't make sense to have something that is not visible but is editable?  If so I think there should be a check in the save to disallow these conflicting settings.

arthanson (changes requested):

netbox/extras/models/customfields.py
jeremystretch: I don't think either field is necessarily dependent on the other. A user may which to omit a custom field from display in the UI but still allow it to be viewable/editable "behind the scenes" for convenience when editing the object, for example.

PR #14332, opened 2023-11-23
Fixes #14329: Improve diffs for custom fields

JCWasmx86:
### Fixes: #14329

This improves the diff by comparing the changes dicts recursively. It changes the Diff and the Pre/Post-Change Data

image

That's one issue I found, but it seems unrelated, as even without my patch, the type of the floating-point field seems to change:
image

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -1,3 +1,4 @@
+import json
 from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.contenttypes.models import ContentType
@@ -19,7 +20,7 @@ from extras.dashboard.utils import get_widget_class
 from netbox.constants import DEFAULT_ACTION_PERMISSIONS
 from netbox.views import generic
 from netbox.views.generic.mixins import TableMixin
-from utilities.data import deep_compare_dict
+from utilities.data import deep_compare_dict, make_diff
 from utilities.forms import ConfirmationForm, get_field_value
 from utilities.htmx import htmx_partial
 from utilities.paginator import EnhancedPaginator, get_paginate_count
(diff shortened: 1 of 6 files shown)


netbox/utilities/utils.py
```diff
@@ -377,20 +377,32 @@ def prepare_cloned_fields(instance):
...
+    """
+    added_diffs = {}
+    removed_diffs = {}
+
+    for k in old:
+        old_data = old[k]
+        new_data = new[k]
+        # HACK: Sometimes, there are issues with floating point fields
+        # where the type of old_data is `float` and new_data is a `str`.
+        # So we serialize it to a string and compare it.
```
jeremystretch, line 393: I feel this needs much more consideration. A change in data type is a change in data; we should not be hand-waving this. The current implementation does not force string casting when evaluating data, so neither should any improvements.

jeremystretch (changes requested):

jeremystretch:
@arthanson @jeffgdotorg and I talked about this today. The consensus was that we need to address the inconsistency in how the summary difference and the pre-/post-change data are presented. Looking at the screenshot below, note that while the "difference" panel is accurate, the changes highlighted in the pre- & post-change panels still show the entire `custom_fields` as having changed. (@JCWasmx86 does call this out in the original PR description above.)

screenshot

IMO the pre-/post-change panel highlights should exactly match the summary above, or we should remove the highlighting in that section.

netbox/extras/views.py
```diff
@@ -729,21 +729,28 @@ def get_extra_context(self, request, instance):
...
+            custom_fields_added = diff_added['custom_fields'] if 'custom_fields' in diff_added else None
+            custom_fields_removed = diff_removed['custom_fields'] if 'custom_fields' in diff_removed else None
+            cfr_list = []
+            if custom_fields_added:
+                for cf, cf_value in prechange_data['custom_fields'].items():
+                    cfr_list.append((cf, cf_value, cf in custom_fields_added))
+            cfa_list = []
+            if custom_fields_removed:
+                for cf, cf_value in instance.postchange_data['custom_fields'].items():
+                    cfa_list.append((cf, cf_value, cf in custom_fields_removed))
```
jeremystretch, line 742: I'm afraid this sort of approach isn't going to be maintainable long-term. The diffing implementation needs to operate generically; we can't give custom fields specifically any special treatment. (Consider what changes would be needed to support some similarly structured field in a future release.) Ultimately, we want a solution that can follow nested data structures and roll back up their diffs, without any context about the data contained within.

jeremystretch (commented):
Hey @JCWasmx86, apologies for the delay in getting back to this; we've been really backed up on PR reviews.

Unfortunately this approach isn't going to be tenable due to the amount of code it introduces. This PR adds five new utility functions (one of which is 160 lines long) and a new template filter, merely to improve cosmetic highlighting of the differences between two dictionaries.

I appreciate that it's a challenging problem to solve, but we'll need to come up with a more compact, maintainable solution. I'm going to close out this PR for now, but I hope to revisit #14329 soon myself to see if I can come up with anything more clever.

Thanks again for your work on this, and sorry we haven't been able to directly implement it.

PR #14340, opened 2023-11-23
Fixed password was not hashed on REST API update

### LINKED ISSUE #14339: User password update via REST API is not hashed
### NetBox version

v3.6.5

### Python version

3.11

### Steps to Reproduce

1. Create a new User with password (username: test, password: test). For the request use a user with user creation permission:
```
curl --location 'http://netbox/api/users/users/' \
--header 'Authorization: Token 45202..6e9e092df' \
--header 'Content-Type: application/json' \
--data '{
    "username": "test",
    "password": "test"
}'
```
2. Update the new created user's password with a new one. For the request use a user with user change permission and the created user `{id}`:
```
curl --location --request PATCH 'http://netbox/api/users/users/{id}/' \
--header 'Authorization: Token 45202..6e9e092df' \
--header 'Content-Type: application/json' \
--data '{
    "password": "newpassword"
}'
```

### Expected Behavior

The password stored on the database must be hashed, so user with the password updated is able to login on the Netbox instance.

### Observed Behavior

The password stored on the database is not hashed, so user with the password updated is not able to login on the Netbox instance.

Connected on the database, do (where `{id}` is the created user's id) :

`SELECT password FROM auth_user WHERE id={id}`

Result `"newpassword"`

fanshan:
* When we updated a user password with a REST API call the password was stored in clear in plain text in the database.

### Fixes: #14339

* Add an `update` method on the UserSerializer that set properly the new password if provided by the API call.

### DIFF
--- a/netbox/users/api/serializers.py
+++ b/netbox/users/api/serializers.py
@@ -60,9 +60,7 @@ class UserSerializer(ValidatedModelSerializer):
         if password is not None:
             instance.set_password(password)
 
-        super().update(instance, validated_data)
-
-        return instance
+        return super().update(instance, validated_data)
 
     @extend_schema_field(OpenApiTypes.STR)
     def get_display(self, obj):
--- a/netbox/users/api/serializers.py
+++ b/netbox/users/api/serializers.py
@@ -52,6 +52,18 @@ class UserSerializer(ValidatedModelSerializer):
 
         return user
 
+    def update(self, instance, validated_data):
+        """
+        Ensure proper updated password hash generation.
+        """
+        password = validated_data.pop('password', None)
+        if password is not None:
+            instance.set_password(password)
+
+        instance.save()
+
+        return instance
+
     @extend_schema_field(OpenApiTypes.STR)
     def get_display(self, obj):
         if full_name := obj.get_full_name():
--- a/netbox/users/api/serializers.py
+++ b/netbox/users/api/serializers.py
@@ -60,7 +60,7 @@ class UserSerializer(ValidatedModelSerializer):
         if password is not None:
             instance.set_password(password)
 
-        instance.save()
+        super().update(instance, validated_data)
 
         return instance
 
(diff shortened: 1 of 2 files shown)


abhi1693:

netbox/users/api/serializers.py
```diff
@@ -52,6 +52,18 @@ def create(self, validated_data):
...
+        """
+        Ensure proper updated password hash generation.
+        """
+        password = validated_data.pop('password', None)
+        if password is not None:
+            instance.set_password(password)
+
+        instance.save()
+
+        return instance
```
jeremystretch, line 65: We're missing a call to `super().update()` here, which is needed for additional validation and to handle any future many-to-many relationships.

netbox/users/tests/test_api.py
```diff
@@ -55,6 +55,37 @@ def setUpTestData(cls):
         User.objects.bulk_create(users)
 
 
+class ChangeUserPasswordTest(APITestCase):
+
+    user_permissions = ['auth.change_user']
+
+    def test_that_password_is_changed(self):
```
jeremystretch, line 62: This doesn't need its own class; it can go in `UserTest`.

netbox/users/tests/test_api.py
```diff
@@ -55,6 +55,37 @@ def setUpTestData(cls):
...
+        Test that password is changed
+        """
+
+        user_credentials = {
+            'username': 'user1',
+            'password': 'abc123',
+        }
+        user = User.objects.create_user(**user_credentials)
+
+        print(user.id)
```
jeremystretch, line 73:
```suggestion
```

jeremystretch (changes requested):
Thanks @fanshan!

netbox/users/api/serializers.py
fanshan: Removed `instance.save()` in favor of `super().update()` as instance is saved on super method.

fanshan:
> Thanks @fanshan!

Thanks @jeremystretch for your review ;-)

jeremystretch (approved):

Issue #9816, opened 2022-07-21
Support for VPN Tunnels


mlazzarotto:
### NetBox version

3.2.6

### Feature type

New functionality

### Proposed functionality

I suggest to add support for site-to-site VPN tunnels.
In this way I can have a clear view of all of my tunnels

### Use case

I have an Excel with 100+ site-to-site vpn tunnels.
This would help to store all the details about VPN.

### Database changes

This is a list of fields that I'd like to have

- Name 
- status (Active, Offline, Backup, Planned, Deprecated)
- type (route based VTI, policy based)
- nat traversal enabled (true, false)
- interface of the device (linked to device interface)
- local peer ip (one or more, linked to ip in IPAM)
- remote peer ip (one or more, linked to ip in IPAM)
- list of local subnets (linked to Subnets in IPAM)
- list of remote subnets (linked to Subnets in IPAM)
- protocols used (IKEv1/IKEv2 and IPsec)
- date created
- contact field
- description field

### External dependencies

I looked at this https://pypi.org/project/netbox-tunnels-plugin/, I don't know if it can be implemented or not

mlazzarotto:
Please don't close and mark as duplicate. It is indeed a duplicate of #5845 but this is more detailed.

DanSheps:
This may be something that is better off as a plugin, at least until it gets a chance to be properly implemented in core.

Given that the plugin you linked is 2 years old, I doubt it would work with the current iteration of netbox.

ryanmerolle:
Feels like the perfect candidate for a plugin.  If it was well built, it could even make it into the core at some point.

ryanmerolle:
The maintainer @jdrew82 of the example package abandoned that plugin for work on a Nautbot plugin of the same approach.

The approach looks solid if it made it to core, but could get some refinements.

@mlazzarotto missed the MTU.  The maintainer specifically set:

 - Tunnel MTU
 - CLNS MTU (I did not realize you could set MTU for other types of traffic, but this makes sense, I'd probably take a more simple approach and just allow for MTU to be set the same as any other interface)


I feel like the remote details would just be derived from modeling the remote end in NetBox.  I recognize this is not always the case with edge devices, but this issue is not specific to tunnels, and it may be a can of worms addressing that with this feature.

jeremystretch:
> @mlazzarotto missed the MTU

MTU is an attribute of the virtual interface on either end (which we already track), rather than the tunnel itself.

[...]

PR #14358, opened 2023-11-27
Closes #14311: Move L2VPN models from `ipam` to `vpn`

### LINKED ISSUE #14311: Move L2VPN models to new vpn app
### Proposed Changes

FR #9816 (VPN tunnel modeling) has introduced a new `vpn` app. This issue proposes moving the L2VPN and L2VPN Termination models and all their related resources from the `ipam` app to the `vpn` app.

### Justification

There was discussion at the time when L2VPN support was being added about whether it made sense to create a new app for it. Although we opted not to at the time, it would be prudent to move these now as the `vpn` app is a much more suitable home for these models.

Unfortunately, this will effect a breaking change in the REST API, however it should be easily manageable.

jeremystretch:
### Closes: #14311

Relocates the L2VPN and L2VPNTermination models, as well as all their related resources, from the `ipam` app to the new `vpn` app introduced by #9816.


DanSheps:
Testing this out, I am getting an integrity error after upgrading when trying to add a new termination to the L2VPN.

Will have more information shortly.

DanSheps (approved):
Reviewed code, code looks to be in order with no references missing.

Tested locally, no issues with migrations or operations.

PR #14370, opened 2023-11-28
Fixes #14325: Ensure expanded numeric arrays are ordered

### LINKED ISSUE #14325: "service" does not preserve the order of ports when editing
### NetBox version

v3.6.5

### Python version

3.8

### Steps to Reproduce

1. Create a "service", and enter the ports as "9093,9095,9998-9999"
2. View the list of services
3. Edit the service (i.e. click the pencil icon at the end of the service row)

### Expected Behavior

Either the ports to remain in the order originally entered, or to be sorted. (I note that the data type in the underlying Postgres column is `integer[]` which is an ordered list)

### Observed Behavior

When viewing the table of services (`/ipam/services/`), the ports are shown in order:

image

It also shows the same when viewing the details of an individual service (e.g. `/ipam/services/2/`)

However, when editing the service (`/ipam/services/2/edit/`), the ports are in a randomized order:

image

This matches what's in the database, which in the same randomized order:

```
netbox=# select ports from ipam_service where id=2
         ports
-----------------------
 {9999,9093,9998,9095}
(1 row)
```

jeremystretch:
### Fixes: #14325

Call `sorted()` on the return value of `parse_numeric_range()` after casting to a `set()`.

### DIFF
--- a/netbox/utilities/forms/utils.py
+++ b/netbox/utilities/forms/utils.py
@@ -40,7 +40,7 @@ def parse_numeric_range(string, base=10):
         except ValueError:
             raise forms.ValidationError(f'Range "{dash_range}" is invalid.')
         values.extend(range(begin, end))
-    return list(set(values))
+    return sorted(list(set(values)))
 
 
 def parse_alphanumeric_range(string):
--- a/netbox/utilities/forms/utils.py
+++ b/netbox/utilities/forms/utils.py
@@ -40,7 +40,7 @@ def parse_numeric_range(string, base=10):
         except ValueError:
             raise forms.ValidationError(f'Range "{dash_range}" is invalid.')
         values.extend(range(begin, end))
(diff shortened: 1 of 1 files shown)


netbox/utilities/forms/utils.py
```diff
@@ -40,7 +40,7 @@ def parse_numeric_range(string, base=10):
         except ValueError:
             raise forms.ValidationError(f'Range "{dash_range}" is invalid.')
         values.extend(range(begin, end))
-    return list(set(values))
+    return sorted(list(set(values)))
```
candlerb, line 43: I think sorted() should work directly on the set, without having to make a list.

```
>>> sorted(set([3,2,1]))
[1, 2, 3]
>>> sorted(list(set([3,2,1])))
[1, 2, 3]
```

PR #14387, opened 2023-11-30
Fixes #12751 - Usability improvements for object selector

### LINKED ISSUE #12751: ip address - interface assignment section - object selector - usability adjustments
### NetBox version

v3.5.2

### Feature type

### Proposed functionality

re the change in #10757. 

#### There are a few changes that could make this more usable / less cumbersome (as mentioned in #12669 )
1: display the name of the device/vm along with the interface name in the list. this would remove some confusion as to what device/interface pair is currently assigned. it would also make directly selecting a new interface from the list (it is a dropdown selection box after all) easier
2: when clicking the "Open Selector" box:
  - if the interface assignment option is "Device" the selector should pre-open with device selector enabled (or a default saved filter of "device search" or similar enabled)
  - if the interface assignment option is "Virtual Machine" the selector should pre-open with Virtual Machine selector enabled (or a default saved filter of "virtual machine search" or similar enabled)
  - if a device / vm interface is already assigned, that device/vm should be filtered to automatically (allows for easy re-assignment to a different interface on the same device/vm)
  
3: in the selector view, the related device/virtual machine name should be displayed with each interface name in the interface list. by default it 
4: consider not having any interfaces returned until a user triggers a search, otherwise there are objects to select at random and it can cause confusion (like thinking these are interfaces from the device of the existing assigned interface)

### Use case

re the above:
1: we have seen a user change an assignment from eth0 to eth1 in the dropdown list without using the selector, however this ending up being eth1 on a different device, it is not possible to tell from this viewpoint what device an interface is assigned to, but its possible to search for interfaces without using the new selector modal. simply combining the device/vm name with the interface name would solve this.
2: preselecting device / virtual machine filters removes a click and confusion with the new selector
3: the list of interfaces returned does not show the name of the related device/vm. if your searching based on a filter like virtual chassis or similar you will get a list of interfaces likely with matching names, but no idea which device they relate to.
4: having objects show up for assignment without any search criteria is a bit odd

### Database changes

_No response_

### External dependencies

_No response_

kkthxbye-code:
### Fixes: #12751

* Adds preselected filters
* Applies the filter on selection instead of requiring the search button to be clicked.

The typescript change was because of a bug that was also present for the Saved Filter field. If a collapsible field is defaulted to be shown, the list will not populate when opening like it is supposed to, but only when the user starts typing.

I was a bit uncertain about the `selector_fields` name, `selector_default` or `selector_default_fields` might be better. Feedback appreciated.

The default fields were taken from the initial PR adding selector fields. I looked at what fields were removed from the forms and added them as pre-selected fields.


ITJamie:
im wondering should we make a separate issue to mark this as fixing as #12751 lists several issues with the selector. this pr fixes part 2 of the list of issues.

jeremystretch:
@ITJamie this is a great example of why it's so important to ensure each feature request conveys a concise, discrete enhancement and to not bundle multiple initiatives in a single issue. If you don't believe this PR fully resolves the FR, I'd suggest breaking out the remaining items into separate FRs so that this can move forward.

jeremystretch (approved):
Thanks @kkthxbye-code!

c-mege:
@kkthxbye-code : thank you !

Issue #14401, opened 2023-12-01
Not allowed to put at U0


Riprock:
### NetBox version

v3.6.5

### Python version

3.11

### Steps to Reproduce

![image](https://github.com/netbox-community/netbox/assets/8990377/89ba5a51-2aa0-4b74-9f34-2f82a9614d34)
![image](https://github.com/netbox-community/netbox/assets/8990377/462d4852-7598-4f82-bef0-6e43a0b36d95)
![image](https://github.com/netbox-community/netbox/assets/8990377/8b53f34c-6ef1-4792-84cf-f91889e9e58b)

This is all preformed in the UI
1. Create New Rack
2. Create new Device
3. Attempt to assign to to U0 To have device be at the bottom of the rack

### Expected Behavior

What I expected is it to place the Device in U0-U3 Not to have a 1U gap as shown in pictures above

### Observed Behavior

Error was provided in the interface saying I cant put a device at U0 because it has to be 1 or greater

jeremystretch:
0 should not be a permitted value for `Rack.starting_unit`. The bug is not in the placement of the device, but rather in the failure to validate the minimum value for this field.

PR #14425, opened 2023-12-04
14401 validate rack startion position > 0

arthanson:
### Fixes: #14401 

Adds validator that Racks must begin at starting_unit 1.  Added to feature branch as requires migration and changes potential existing values.  Only found one discussion related to this #11420


jeremystretch:
@arthanson this should be a fix for v3.6.x. Please rebase against the `develop` branch.

PR #14439, opened 2023-12-05
14401 validate rack startion position > 0

### LINKED ISSUE #14401: Not allowed to put at U0

arthanson:
### Fixes: #14401 

Adds validator that Racks must begin at starting_unit 1.  Added to feature branch as requires migration and changes potential existing values.  Only found one discussion related to this #11420


netbox/dcim/migrations/0182_alter_rack_starting_unit.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ('dcim', '0181_rename_device_role_device_role'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='rack',
+            name='starting_unit',
+            field=models.PositiveSmallIntegerField(default=1, validators=[django.core.validators.MinValueValidator(1)]),
```
jeremystretch, line 16: Because the implementation of a field validator doesn't actually effect a concrete change in the database, we can move this into the migration file where the field is defined (`0174_rack_starting_unit`). This allows us to avoid introducing a no-op migration file.

(The only check constraint on the `starting_unit` field is for `>= 0`, which is installed by PositiveSmallIntegerField.)

jeremystretch (changes requested):

netbox/dcim/migrations/0182_alter_rack_starting_unit.py
arthanson: Done.

PR #14456, opened 2023-12-06
Fixes IPv6 detection from headers

### LINKED ISSUE #14249: API Requests Via Token WITH "Allowed IPs" Set Broken After Upgrade to 3.6.5
### NetBox version

v3.6.5

### Python version

3.9

### Steps to Reproduce

Unless otherwise noted, leave everything else default:

1. Create a permission with "can view" selected and select every object type.
2. Create a user with a password and the newly created permission assigned.
3. Create an API token assigned to the new user with "Allowed IPs" set to `::/0,0.0.0.0/0` to allow all.
4. Perform an API get request using the new token (should fail).
5. Remove the "Allowed IPs" setting on the API token and retry the request (should succeed).

### Expected Behavior

Setting AllowedIPs should not throw a 500 Internal Server Error.

### Observed Behavior

500: Internal Server Error is raised.  Setting debug logging for django and netbox yield the following:

```
2023-11-10 18:01:57,367 netbox.config DEBUG: No previous configuration found in database; proceeding with default values
2023-11-10 18:01:57,368 netbox.config DEBUG: Initialized configuration
2023-11-10 18:01:57,402 netbox.config DEBUG: Cleared configuration
2023-11-10 18:01:57,402 django.request ERROR: Internal Server Error: /api/ipam/vlans/
```

abhi1693:

### DIFF
--- a/netbox/utilities/request.py
+++ b/netbox/utilities/request.py
@@ -1,4 +1,4 @@
-from netaddr import IPAddress
+from netaddr import AddrFormatError, IPAddress
 
 __all__ = (
     'get_client_ip',
@@ -17,10 +17,15 @@ def get_client_ip(request, additional_headers=()):
     )
     for header in HTTP_HEADERS:
         if header in request.META:
-            client_ip = request.META[header].split(',')[0].partition(':')[0]
+            ip = request.META[header].split(',')[0].strip()
+            # Check if the IP address is v6 or v4
+            if ip.count(':') > 1:
+                client_ip = ip
+            else:
+                client_ip = ip.partition(':')[0]
             try:
                 return IPAddress(client_ip)
-            except ValueError:
+            except (AddrFormatError, ValueError):
                 raise ValueError(f"Invalid IP address set for {header}: {client_ip}")
 
     # Could not determine the client IP address from request headers
--- a/netbox/utilities/request.py
+++ b/netbox/utilities/request.py
@@ -1,4 +1,5 @@
 from netaddr import AddrFormatError, IPAddress
+from urllib.parse import urlparse
 
 __all__ = (
     'get_client_ip',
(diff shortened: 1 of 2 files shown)


netbox/utilities/request.py
```diff
@@ -17,10 +17,15 @@ def get_client_ip(request, additional_headers=()):
...
+            # Check if the IP address is v6 or v4
+            if ip.count(':') > 1:
+                client_ip = ip
+            else:
+                client_ip = ip.partition(':')[0]
             try:
                 return IPAddress(client_ip)
-            except ValueError:
+            except (AddrFormatError, ValueError):
                 raise ValueError(f"Invalid IP address set for {header}: {client_ip}")
```
jeremystretch, line 29: We should be able to lean on `urllib` here to simplify the logic. We just need to pass `f"//{ip}"` to `urlparse()` and take `hostname`:

```
>>> urlparse('//1.2.3.4').hostname
'1.2.3.4'
>>> urlparse('//1.2.3.4:80').hostname
'1.2.3.4'
>>> urlparse('//[2001:db8::1]').hostname
'2001:db8::1'
>>> urlparse('//[2001:db8::1]:80').hostname
'2001:db8::1'
```

netbox/utilities/tests/test_request.py
```diff
@@ -0,0 +1,25 @@
...
+        self.factory = RequestFactory()
+
+    def test_ipv4_address(self):
+        request = self.factory.get('/', HTTP_X_FORWARDED_FOR='192.168.1.1')
+        self.assertEqual(get_client_ip(request), IPAddress('192.168.1.1'))
+
+        request = self.factory.get('/', HTTP_X_FORWARDED_FOR='192.168.1.1:8080')
+        self.assertEqual(get_client_ip(request), IPAddress('192.168.1.1'))
+
+    def test_ipv6_address(self):
```
jeremystretch, line 18: We should add a test for IPv6 with port too, e.g. `[2001:db8::8a2e:370:7334]:8080`.

jeremystretch (changes requested):
Thanks @abhi1693!

netbox/utilities/request.py
jeremystretch: I just noticed that this approach doesn't work for IPv6 addresses without brackets, which may appear as `HTTP_X_FORWARDED_FOR` values. It really seems like there should be a simple, canonical solution for this.

jeremystretch (approved):

PR #14463, opened 2023-12-07
Closes #14436: Add indexes for all GenericForeignKey fields

### LINKED ISSUE #14436: Add PostgreSQL indexes for all generic foreign key fields
### Proposed Changes

For all GenericForeignKey (GFK) relationships in a model, a PostgreSQL index should be created for its two concrete fields (content type and ID).

### Justification

Creating indexes for these fields enables optimized lookup of records by related object. (Django unfortunately does not create these indexes automatically.)

jeremystretch:
### Closes: #14436

Migrations introduced in v3.7 are updated in place where applicable.


jeremystretch (commented):
We should also note in the developer docs to add an index for any future GFKs.

arthanson (approved):

Issue #13983, opened 2023-10-05
Custom Field Choice Set Bulk Import error "Nested arrays must have same length"


jayGG83:
### NetBox version

v3.6.3

### Python version

3.10

### Steps to Reproduce

1. Select Custom Field Choices import
2. Data to import:

CSV:
```
name;description;extra_choices
numbers;Numbers;7,13,18,23,26,32,38,80
```

JSON:
```
{
    "name": "numbers",
    "description": "Numbers",
    "extra_choices": [7,13,18,23,26,32,38,80]
}
```

YAML:
```
---
name: numbers
description: Numbers
extra_choices: 7,13,18,23,26,32,38,80
```

3. Click Submit

### Expected Behavior

It is expected to create the cusotm field choices

### Observed Behavior

Returns this error in all cases: Record 1 extra_choices: Nested arrays must have the same length.

jeremystretch:
The root issue here is that each choice value is being interpreted as an array, such that e.g. `13` becomes `["1", "3"]`. The error message is a bit of a red herring, and presents only if all the choice values are not of the same length. For example, importing a field value of `A1,B2,C3` will result in choices `[(A, 1), (B, 2), (C, 3)]`.

I'm not sure we have a great option for handling nested arrays in CSV data, but we should seek to support them when expressed via JSON/YAML.

sleepinggenius2:
I ran into a similar requirement recently in a plugin I was working on and defining the field like below on the `*ImportForm` class has worked great for CSV data. Admittedly, I have not tried importing JSON or YAML.

```python
field = SimpleArrayField(
    base_field=SimpleArrayField(
        base_field=CharField(),
        min_length=2,
        max_length=2,
    ),
    delimiter=';',
    required=False,
    help_text=_('Semicolon-separated list of comma-separated pairs, encased with double quotes (e.g. "value1,Label 1;value2,Label 2")'),
)
```

arthanson:
We could potentially have a "CSV string delimiter" along with the "CSV delimiter" with a none option default.  If set to semicolon then it could parse string arrays based on it.  But this might be overly-complicated and would need to be made clear in the UI.

There would have to be some validation that the two delimiter fields are not set to the same value.

PR #14469, opened 2023-12-07
14467 change ChoiceField separator from comma to colon

### LINKED ISSUE #14467: Change CustomFieldChoiceSet label separator from comma to colon
### NetBox version

v3.6.6

### Feature type

### Proposed functionality

Change the CustomFieldChoiceSet label separator from a comma to a colon.  Right now for CustomFieldChoiceSets the extra_choices uses a comma to separate out the label from the value for input:

```
        help_text=mark_safe(_(
            'Enter one choice per line. An optional label may be specified for each choice by appending it with a '
            'comma. Example:'
        ) + ' <code>choice1,First Choice</code>')
```
This FR would change it to be a colon.

### Use case

From working on #13983 this allows nested arrays of these label,value in the CSV import, however commas are default separators for both CSV and the ArrayField.  The commas are not stored anywhere they are just used to split the string, so changing to a colon should have no adverse effect and would allow not only easier and less confusing input of data for users, but would make the code simpler as well.

### Database changes

None

### External dependencies

None

arthanson:
### Fixes: #14467 

Changes the ChoiceField (for CustomFieldChoiceSetForm extra_choices) from a comma to colon.  This is to enable a clean fix for #13983

Should have no side-effects as the comma isn't stored anywhere and the change in the widget makes old saved values appear correctly with the new colon.  The only potential for issue is if someone has a colon stored as part of a label/choice which should be very unlikely.

### DIFF
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -1,4 +1,5 @@
 import json
+import re
 
 from django import forms
 from django.conf import settings
@@ -107,7 +108,7 @@ class CustomFieldChoiceSetForm(BootstrapMixin, forms.ModelForm):
         data = []
         for line in self.cleaned_data['extra_choices'].splitlines():
             try:
-                value, label = line.split(':', maxsplit=1)
+                value, label = re.split(r'(?<!\\):', line, maxsplit=1)
             except ValueError:
                 value, label = line, line
             data.append((value, label))
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -95,8 +95,8 @@ class CustomFieldChoiceSetForm(BootstrapMixin, forms.ModelForm):
         required=False,
         help_text=mark_safe(_(
             'Enter one choice per line. An optional label may be specified for each choice by appending it with a '
-            'comma. Example:'
-        ) + ' <code>choice1,First Choice</code>')
+            'colon. Example:'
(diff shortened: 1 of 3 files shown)


netbox/extras/forms/model_forms.py
```diff
@@ -107,7 +107,7 @@ def clean_extra_choices(self):
         data = []
         for line in self.cleaned_data['extra_choices'].splitlines():
             try:
-                value, label = line.split(',', maxsplit=1)
+                value, label = line.split(':', maxsplit=1)
```
jeremystretch, line 110: We probably want to provide a means to escape colons in the choices value or label. We got away without this for commas (so far), but I expect using colons as delimiters will present an issue without this ability. Something like `re.split('(?<!\\\\):', line)` would allow us to split on colons not preceded by a backslash.

netbox/extras/tests/test_views.py
```diff
@@ -119,6 +120,31 @@ def setUpTestData(cls):
             'description': 'New description',
         }
 
+    # These are here as extra_choices field splits on colon, but is returned
+    # from DB as comma separated.  So we exclude them in validation_excluded_fields
+    # but test they are actually set correctly here.
```
jeremystretch, line 125: It might be cleaner to override `assertInstanceEqual()` on the test class, and overwrite `data['extra_choices`] there (before calling `super()`).

jeremystretch (changes requested):

jeremystretch (approved):

PR #14470, opened 2023-12-07
13983 Add nested arrays for extra_choices in CustomFieldChoiceSet 

### LINKED ISSUE #13983: Custom Field Choice Set Bulk Import error "Nested arrays must have same length"

arthanson:
### Fixes: #13983 

Relies on change in #14469 Allows to specify multiple choices and labels for CSV import of extra_choices - just quote the whole field: "choice1:First Choice,choice2:Second Choice"

### DIFF
--- a/netbox/extras/forms/bulk_import.py
+++ b/netbox/extras/forms/bulk_import.py
@@ -1,3 +1,5 @@
+import re
+
 from django import forms
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres.forms import SimpleArrayField
@@ -93,7 +95,7 @@ class CustomFieldChoiceSetImportForm(CSVModelForm):
             data = []
             for line in self.cleaned_data['extra_choices']:
                 try:
-                    value, label = line.split(':', maxsplit=1)
+                    value, label = re.split(r'(?<!\\):', line, maxsplit=1)
                 except ValueError:
                     value, label = line, line
                 data.append((value, label))
--- a/netbox/extras/forms/bulk_import.py
+++ b/netbox/extras/forms/bulk_import.py
@@ -96,8 +96,6 @@ class CustomFieldChoiceSetImportForm(CSVModelForm):
             for line in self.cleaned_data['extra_choices']:
                 try:
                     value, label = re.split(r'(?<!\\):', line, maxsplit=1)
-                    value = value.replace('\\:', ':')
-                    label = label.replace('\\:', ':')
                 except ValueError:
                     value, label = line, line
                 data.append((value, label))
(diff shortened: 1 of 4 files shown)


netbox/extras/forms/bulk_import.py
```diff
@@ -85,6 +90,17 @@ class Meta:
...
 
+    def clean_extra_choices(self):
+        if isinstance(self.cleaned_data['extra_choices'], list):
+            data = []
+            for line in self.cleaned_data['extra_choices']:
+                try:
+                    value, label = re.split(r'(?<!\\):', line, maxsplit=1)
+                except ValueError:
+                    value, label = line, line
+                data.append((value, label))
```
jeremystretch, line 101: We also need to normalize any escaped colons (`\:`) when saving the data.
```suggestion
                value = value.replace('\:', ':')
                label = label.replace('\:', ':')
                data.append((value, label))
```

(This should also be applied to the model form for #14467.)

jeremystretch (changes requested):

netbox/extras/forms/bulk_import.py
arthanson: It needed to be saved with the escaped character as doing the save/edit cycle would replace it.  This has a change to replace the label when it is displayed in the model form using the custom field.

jeremystretch: The escape character should never be saved to the database; it's relevant only for processing raw input. I don't follow what you mean about the save/edit cycle, but we'll have to figure out a different approach to ensure the normalized values are stored.

arthanson: The save/edit cycle meant if you edit and save a choice-set then re-edit it.  You either have to save the escape characters or else re-add them when edit.  I've changed to re-add them when you edit so they aren't saved in the database.

jeremystretch (approved):

Issue #14424, opened 2023-12-04
Remove ChangeLoggedModel from StagedChange


arthanson:
### Proposed Changes

Remove ChangeLoggedModel mixin from the StagedChange model.  Marking this as a deprecation as it would remove functionality.

### Justification

As per #13640 When doing staged change API it can result in numerous ChangeLoggedModels being produced - For example when ingesting some customer information using the staging method there can be up to 100k+ objects being altered. Deleting these takes very long due to the ChangeLoggedModel process.

### Impact

Would regress in that no more ChangeLoggedModels would be produced for StagedChanges.

PR #14476, opened 2023-12-08
14424 Remove ChangeLoggedModel from StagedChange

arthanson:
### Fixes: #14424 

Remove ChangeLoggedModel from StagedChange.  Not sure of the testing to do for this beyond the test suites.


netbox/extras/migrations/0104_remove_stagedchange_created_and_more.py
jeremystretch, line 1: Let's give the migration a meaningful name; something like `0104_stagedchange_remove_change_logging`.

jeremystretch (changes requested):

netbox/extras/migrations/0104_remove_stagedchange_created_and_more.py
arthanson: Done

jeremystretch (approved):

Issue #14147, opened 2023-10-31
Prevent logging to Change Log when no changes are made


thefreakquency:
### NetBox version

v3.6.4

### Feature type

Change to existing functionality

### Proposed functionality

When a record is patched via API and no changes are made, a Change Log entry is created stating "No changes". I would like to be able to either a) prevent those to be created, using a global configuration flag, or b) provide the ability to purge those manually...

Also, if those empty log gentries are kept, i would like to have, _at minima_, a filter in /extra/changelog/ section. Currently, it is not possible to filter by Action type = No changes. 

This request is somewhat related to #13391 and #6599 

### Use case

I have a script that once in a while goes thru a list of 600+ sites and gather information from a corporate database, then patch the NB site with the required information.  When it patches the site data, as expected, it does not update when information is the same between what's in the request and what's already in NetBox. Most of the time, nothing has changed. That said, even if nothing has changed, a change log is being created for each site. That means that every time my script runs, I get over 600 Change Logs entries even if only 20ish of those were really changed... 

### Database changes

None, AFAIK

### External dependencies

none

ITJamie:
another example of an annoying "no change" changelog ive seen is patching / saving with a reordered list of tags or multiselect objects (restapi/pynetbox). this creates a changelog even though nothing has actually really changed.

rizlas:
Sometimes not all the changes are displayed in the UI even if something is changed via script/api/ecc.
Anyhow. imho, this is a must have feature. It'll also keep the database lighter.

PR #14477, opened 2023-12-08
14147 Prevent logging to Change Log when no changes are made

arthanson:
### Fixes: #14147 

Prevents an ObjectChange record from being created if there is an update and no data has changed.

Adds a CHANGELOG_SKIP_EMPTY_CHANGES config param, default False, that allows to skip the logging if no changes. i.e. the default is the old behavior.

If this flag is set, then to_objectchange can return None which needs to be checked in overriden functions.  Did it this way for performance to prevent instantiating an object and serializing data if it is never going to be used.

A side-effect of this change, which should probably be default, is the `last_updated` field is removed from the serialized data of the changelog.  The reason for this is this would be counted as a changed field when comparing the serialized records (pre / post update).

Added to feature branch as it seems a big enough feature change to be in a point release.

### DIFF
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -82,7 +82,7 @@ changes in the database indefinitely.
 
 ## CHANGELOG_SKIP_EMPTY_CHANGES
 
-Default: False
+Default: True
 
 Enables skipping the creation of logged changes on updates if there were no modifications to the object.
 
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -84,10 +84,10 @@ changes in the database indefinitely.
 
 Default: True
 
-Enables skipping the creation of logged changes on updates if there were no modifications to the object.
+If enabled, a change log record will not be created when an object is updated without any changes to its existing field values.
 
 !!! note
-    As a side-effect of turning this on, the `last_updated` field will not be included in the change log record.
+    The object's `last_updated` field will always reflect the time of the most recent update, regardless of this parameter.
 
 ---
 
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -80,6 +80,17 @@ changes in the database indefinitely.
 
 ---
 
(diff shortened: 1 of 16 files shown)


netbox/netbox/models/features.py
```diff
@@ -84,6 +85,15 @@ def to_objectchange(self, action):
         by ChangeLoggingMiddleware.
         """
         from extras.models import ObjectChange
+
+        postchange_data = None
```
jeremystretch, line 89: This probably isn't the ideal place to hook in, because it requires modifying all the overridden calls to `to_objectchange()` inside individual models with additional logic.

Instead, we wrap the call to `save()` in the `handle_changed_object()` signal receiver, to ensure the ObjectChange instance is saved only if the pre- and post-change data differ. To further simplify this, we can add a property on ObjectChange which returns essentially `self.prechange_data == self.postchange_data`.

netbox/utilities/utils.py
```diff
@@ -159,6 +159,9 @@ def serialize_object(obj, resolve_tags=True, extra=None):
         for field in ['level', 'lft', 'rght', 'tree_id']:
             data.pop(field)
 
+    if get_config().CHANGELOG_SKIP_EMPTY_CHANGES and 'last_updated' in data:
+        data.pop('last_updated')
+
```
jeremystretch, line 164: IMO we should move this logic into the model itself. We might use `serialize_object()` for other purposes irrelevant to change logging.

docs/configuration/miscellaneous.md
```diff
@@ -80,6 +80,17 @@ changes in the database indefinitely.
 
 ---
 
+## CHANGELOG_SKIP_EMPTY_CHANGES
+
+Default: False
```
jeremystretch, line 85: How do we feel about having this default to True? IMO it makes more sense to enable by default, as most people won't have any reason to enable it.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #14524, opened 2023-12-14
Fixes: #14515 - UserConfig creation

### LINKED ISSUE #14515: UserConfig is not being created when pre-creating a account for use with social auth
### Deployment Type

Self-hosted

### NetBox Version

v3.6.6

### Python Version

3.11

### Steps to Reproduce

1. Connect to a social auth system
2. Pre-create a user account
3. Login with user account

### Expected Behavior

Login succeeds

### Observed Behavior

Exception: User.config.RelatedObjectDoesNotExist

DanSheps:
### Fixes: #14515

* Add NetBoxUser to recevier for create_userconfig

### DIFF
--- a/netbox/users/models.py
+++ b/netbox/users/models.py
@@ -218,6 +218,7 @@ class UserConfig(models.Model):
 
 
 @receiver(post_save, sender=User)
+@receiver(post_save, sender=NetBoxUser)
 def create_userconfig(instance, created, raw=False, **kwargs):
     """
     Automatically create a new UserConfig when a new User is created. Skip this if importing a user from a fixture.


netbox/users/models.py
```diff
@@ -218,6 +218,7 @@ def clear(self, path, commit=False):
 
 
 @receiver(post_save, sender=User)
```
abhi1693, line 220: Is this really needed as it seems redundant to me?

DanSheps: It is used in a number of tests, as well as Django internals might use it in places (not sure), that would require a deeper dive.

Simply adding the additional recevier doesn't impact much but it will capture any corner cases.

abhi1693: What I meant was since `NetBoxUser` is a proxy model to `User`, just having a post save for `NetBoxUser` is not enough?

DanSheps: > What I meant was since `NetBoxUser` is a proxy model to `User`, just having a post save for `NetBoxUser` is not enough?

A call to NetBoxUser.save() won't trigger the User.post_save signal (and likewise User.save() won't trigger NetBoxUser.post_save signal) so I don't think so.

We could go through and replace all instances of User with NetBoxUser but if User is used anywhere inside or outside of NetBox code that signal will get missed and it will through an error in the context_processor middleware if the user tries to login without the config built.

I suppose we could also just add a check in the context_processor middleware but I figured this was the better way to handle it.

jeremystretch (approved):
Thanks @DanSheps!

PR #14547, opened 2023-12-16
Add NetBox version output in `upgrade.sh`

### LINKED ISSUE #14507: Print NetBox version when running upgrade.sh
### NetBox version

v3.7-beta1

### Feature type

New functionality

### Proposed functionality

When running upgrade.sh it would be beneficial to see what is the actual NetBox version that is being prepared. 

Something like

```
echo "NetBox version is $(grep ^VERSION netbox/netbox/settings.py | cut -d\' -f2)"
NetBox version is 3.7-beta1
```

would maybe be enough somewhere in the beginning of upgrade.sh.

### Use case

Sometimes there is confusion when symlinking or pulling the versions here and there and running upgrade.sh, what is the actual codebase that is being activated.

### Database changes

None

### External dependencies

None

markkuleinio:
### Fixes: #14507

NetBox version is fetched from `netbox/netbox/settings.py` (the script already `cd`'d to the location of `upgrade.sh` = the top-level directory of NetBox).

### DIFF
--- a/upgrade.sh
+++ b/upgrade.sh
@@ -7,6 +7,10 @@
 # Python 3.8 or later.
 
 cd "$(dirname "$0")"
+
+NETBOX_VERSION="$(grep ^VERSION netbox/netbox/settings.py | cut -d\' -f2)"
+echo "You are installing (or upgrading to) NetBox version ${NETBOX_VERSION}"
+
 VIRTUALENV="$(pwd -P)/venv"
 PYTHON="${PYTHON:-python3}"
 


DanSheps:
This might make sense to rebase to feature, as we are making a pretty big change to how the upgrade script runs (pause).  Thoughts?

markkuleinio:
Sounds reasonable, I rebased the PR to `feature` (unless I failed it somehow but looks good I think).

upgrade.sh
```diff
@@ -7,6 +7,15 @@
 # Python 3.8 or later.
 
 cd "$(dirname "$0")"
+
+NETBOX_VERSION="$(grep ^VERSION netbox/netbox/settings.py | cut -d\' -f2)"
+echo "You are installing (or upgrading to) NetBox version ${NETBOX_VERSION}"
+# Pause if running interactively
+if [ -t 0 ]; then
```
jeremystretch, line 14: In what scenarios might this check return a false positive? I'd rather not risk breaking automated workflows merely to report on the version. Maybe this should be an optional behavior instead?

markkuleinio: I now realized that checking also `stdout` is needed, otherwise `./upgrade.sh > logfile.txt` will just hang there waiting for input, so that would be `[ -t 0 -a -t 1 ]`.

In that case:
- When run interactively (either directly or via another script), pause is triggered if both `stdin` and `stdout` are on the terminal
- When run via cron job or similar, pause is skipped
- I'm not able to figure out any situations where `stdin` and `stdout` are on the terminal but there is nobody to read and respond (but frankly this is the first time I'm figuring this out)

But yes, if you feel the pause is not desirable after all, let me know, I'll then remove that code. I don't know if adding `--confirm` or similar option would help those that accidentally run `upgrade.sh` with the wrong codebase.

jeremystretch: Let's not turn this into something more complicated than it needs to be. Let's do this:

* Remove the confirmation prompt and just display the version to which we're upgrading (per the original FR)
* Rebase to the `develop` branch as this is not a v3.7 milestone FR

jeremystretch (approved):
Thanks @markkuleinio!

PR #14565, opened 2023-12-19
Closes #14536: Enable ENFORCE_GLOBAL_UNIQUE by default

### LINKED ISSUE #14536: Enable `ENFORCE_GLOBAL_UNIQUE` by default
### NetBox version

v3.6.6

### Feature type

### Proposed functionality

Change the default value of the [`ENFORCE_GLOBAL_UNIQUE`](https://docs.netbox.dev/en/stable/configuration/miscellaneous/#enforce_global_unique) configuration parameter from false to true. This will enforce unique IP addressing within the global (non-VRF) table by default.

This change would affect only deployments without any configuration already defined.

Please use a :+1: or :-1: response below to indicate your support for/opposition to this proposed change.

### Use case

Enforcing unique IP space by default is more restrictive and thus safer than the current default. Obviously, the parameter can still be toggled as needed.

This change would also be consistent with the `enforce_unique` field on the VRF model, which defaults to True.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Fixes: #14536

Change the default value for the `ENFORCE_GLOBAL_UNIQUE` configuration parameter from False to True.

### DIFF
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -103,9 +103,12 @@ The maximum size (in bytes) of an incoming HTTP request (i.e. `GET` or `POST` da
 
 !!! tip "Dynamic Configuration Parameter"
 
-Default: False
+Default: True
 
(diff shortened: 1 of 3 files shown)


DanSheps (approved):
Looks good to me.  Cleans up the tests a little with those override settings too.

Issue #14550, opened 2023-12-18
Changing Event Rule Action Type only works one way


rodvand:
### Deployment Type

Self-hosted

### NetBox Version

v3.7-beta1

### Python Version

3.8

### Steps to Reproduce

1. Create a webhook (anything goes)
2. Create an Event Rule with Webhook as Action Type and select the new Webhook in the dropdown
3. Save the Event rule
4. Edit the Event rule and try changing the Action Type from Webhook to Script

### Expected Behavior

The changing of the Action Type should change the input field to allow selection of a script.

### Observed Behavior

It does not change. If you do it the other way around, changing from Script to Webhook seems to work as expected.

PR #14566, opened 2023-12-19
14550 fix EventRule HTMX change of action_type

arthanson:
### Fixes: #14550 

@jeremystretch if you can double-check added a get_initial_data to the base view which allows resetting form fields on an HTMX change which I thought would be the most flexible fix.


jeremystretch:
I think there might be a simpler fix for this. Please have a look at #14571 and let me know if I've missed something.

jeremystretch:
Closing this out in favor of #14571

PR #14568, opened 2023-12-20
Add PUBLIC_SWAGGER configuration option

### LINKED ISSUE #14555: Add config option to not expose SWAGGER schemas to unauthenticated users
### NetBox version

3.2

### Feature type

### Proposed functionality

Swagger metadata are by default opened to unauthenticated users.
I'd like to add a configuration option to be able to only expose them to authenticated users.

Even though it's only the metadata (and not read or write operations) public Netbox instances attract a significant amount of "Google traffic" and people experimenting with the API, slowing down Netbox. This is a real world issue we're having at Wikimedia.

An alternative would be to disable it by default (and not configuration option) to keep the config more learn, but this might be a breaking change for existing users.

I already have a working change on https://github.com/XioNoX/netbox/commit/14472c0cf4bc48f066af6da1c8f8b05084bd78d4 (tested on 3.6) so I can send a PR if this is approved.

### Use case

See above.

### Database changes

None.

### External dependencies

None.

XioNoX:
### Fixes: #14555

Swagger metadata are by default opened to unauthenticated users. Add a configuration option to be able to only expose them to authenticated users.

Even though it's only the metadata (and not read or write operations) public Netbox instances attract a significant amount of "Google traffic" and people experimenting with the API, slowing down Netbox.

### DIFF
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -180,14 +180,6 @@ When determining the primary IP address for a device, IPv6 is preferred over IPv
 
 ---
 
-## PUBLIC_SWAGGER
-
-Default: True
-
-Swagger exposes by default the schemas to unauthenticated users. Set this to False to only expose them to authenticated users.
-
----
-
 ## QUEUE_MAPPINGS
 
 Allows changing which queues are used internally for background tasks.
@@ -242,3 +234,11 @@ This parameter controls how frequently a failed job is retried, up to the maximu
 Default: `0` (retries disabled)
 
 The maximum number of times a background task will be retried before being marked as failed.
+
+---
+
+## SWAGGER_PUBLIC
+
+Default: True
+
+The [REST API](../integrations/rest-api.md) interactive documentation (Swagger) exposes the schemas by default to unauthenticated users. As this UI is ressource intensive, it can be preferable to only expose the schemas to authenticated users, by setting this variable to False. Especially for public instances.
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -180,6 +180,14 @@ When determining the primary IP address for a device, IPv6 is preferred over IPv
 
 ---
 
+## PUBLIC_SWAGGER
+
+Default: True
+
+Swagger exposes by default the schemas to unauthenticated users. Set this to False to only expose them to authenticated users.
+
+---
+
 ## QUEUE_MAPPINGS
 
 Allows changing which queues are used internally for background tasks.
--- a/netbox/netbox/configuration_example.py
+++ b/netbox/netbox/configuration_example.py
@@ -190,9 +190,6 @@ PLUGINS = []
 #     }
 # }
 
-# Expose the SWAGGER UI schemas to unauthenticated users
-PUBLIC_SWAGGER = True
-
 # Remote authentication support
 REMOTE_AUTH_ENABLED = False
 REMOTE_AUTH_BACKEND = 'netbox.authentication.RemoteUserBackend'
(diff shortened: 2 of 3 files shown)


docs/configuration/miscellaneous.md
```diff
@@ -180,6 +180,14 @@ When determining the primary IP address for a device, IPv6 is preferred over IPv
 
 ---
 
+## PUBLIC_SWAGGER
```
jeremystretch, line 183: I'm going to lightly suggest flipping this to `SWAGGER_PUBLIC`, so that it would fit better with potential future configuration options for Swagger.

docs/configuration/miscellaneous.md
```diff
@@ -180,6 +180,14 @@ When determining the primary IP address for a device, IPv6 is preferred over IPv
...
+
+Default: True
+
+Swagger exposes by default the schemas to unauthenticated users. Set this to False to only expose them to authenticated users.
```
jeremystretch, line 187: This probably needs a bit more context, as the casual reader might not know what Swagger is or where the schemas can be accessed. (Feel free to link to other areas of the documentation.)

netbox/netbox/configuration_example.py
```diff
@@ -190,6 +190,9 @@
 #     }
 # }
 
+# Expose the SWAGGER UI schemas to unauthenticated users
+PUBLIC_SWAGGER = True
+
```
jeremystretch, line 195: We can omit this from the example configuration. It's not meant to be an exhaustive list.

netbox/netbox/settings.py
```diff
@@ -653,6 +653,7 @@ def _setting(name, default=None):
     'VERSION': VERSION,
     'COMPONENT_SPLIT_REQUEST': True,
     'REDOC_DIST': 'SIDECAR',
+    'SERVE_PUBLIC': getattr(configuration, 'PUBLIC_SWAGGER', True),
```
jeremystretch, line 656: The config parameter should first be defined toward the top of `settings.py` for consistency with other parameters, then referenced here.

jeremystretch (changes requested):
Thanks @XioNoX!

netbox/netbox/settings.py
XioNoX: ack, done in the new commit.

netbox/netbox/configuration_example.py
XioNoX: ack, done in the new commit.

docs/configuration/miscellaneous.md
XioNoX: ack, done in the new commit.

jeremystretch (approved):
Thanks @XioNoX!

jeremystretch (changes requested):
Might have jumped the gun on this. The `SERVE_PUBLIC` setting doesn't appear to actually restrict serving the schema. Rather, I believe it has to do with the context in which the schema is rendered.

XioNoX:
I'm not sure I fully understand your comment and what "context" means here. Should I change something else in this PR or does it mean more investigation ?
Thanks

XioNoX:
Hi, @jeremystretch just want to make sure nobody forgot about this :) Or if I can be of any help. Thanks

jeremystretch:
@XioNoX your proposed change does not address #14555, as indicated above. Do you still intend to work on it?

XioNoX:
> @XioNoX your proposed change does not address #14555, as indicated above. Do you still intend to work on it?

Ideally yes, do you have any pointers on the best path for it? Thanks

jeremystretch:
> Ideally yes, do you have any pointers on the best path for it?

I'm sorry, I don't. I'm going to close this PR out as it does not address #14555. You are welcome to submit a new PR if you are able to identify a solution.

PR #14569, opened 2023-12-20
Fix custom script job results while executing the call command

### LINKED ISSUE #14549: Job results are not available when executing custom scripts via the runscript command
### Deployment Type

Self-hosted

### NetBox Version

v3.6.6

### Python Version

3.11

### Steps to Reproduce

1. Create/upload any custom script (e.g. the example script from documentation)
2. Try to execute the script via `runscript` command 
```
/opt/netbox/venv/bin/python /opt/netbox/netbox/manage.py runscript example.NewBranchScript --commit --loglevel debug --data '{"site_name": "testing", "switch_count": 1, "switch_model": 11}'
```

### Expected Behavior

Script is completed and job results are available

### Observed Behavior

Script gets completed, but a KeyError exception gets raised when trying to access the job results
image

It seems that an incorrect name tries to get accessed when the script is ran through the `runscript` command, as compared to running the script via the GUI

image

kprince28:

### DIFF
--- a/netbox/extras/management/commands/runscript.py
+++ b/netbox/extras/management/commands/runscript.py
@@ -114,7 +114,7 @@ class Command(BaseCommand):
         # Create the job
         job = Job.objects.create(
             object=module,
-            name=script.name,
+            name=script.class_name,
             user=User.objects.filter(is_superuser=True).order_by('pk')[0],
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @kprince28!

PR #14580, opened 2023-12-21
Closes #14579: Add user language preference

### LINKED ISSUE #14579: Allow the user to designate a preferred language for UI translation
### NetBox version

v3.7-beta1

### Feature type

New functionality

### Proposed functionality

Introduce a dropdown field on the user preferences form where the user can select their preferred language. If defined, this selection will override the `Accept-Language` request header when determining the language to use.

I'm not sure whether it makes sense to list _all_ of the available languages under [ISO 639 (set 1)](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes), or just those for which NetBox includes translations (probably the later).

### Use case

This will allow the user to designate a translation language other than what their browser is configured to request. For example, a German-speaking user's browser would typically be configured to prefer German, but they might prefer to work in NetBox using English (as a complete German translation is unavailable).

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Fixes: #14579

- Declare the list of supported languages in `settings.LANGUAGES` (these are being added)
- Introduce a new UserConfig parameter: `locale.language`
- Extend UserConfigView to set/clear the language cookie depending on the value of this parameter

### DIFF
--- a/netbox/account/views.py
+++ b/netbox/account/views.py
@@ -13,6 +13,7 @@ from django.shortcuts import render, resolve_url
 from django.urls import reverse
 from django.utils.decorators import method_decorator
 from django.utils.http import url_has_allowed_host_and_scheme, urlencode
+from django.utils.translation import gettext_lazy as _
 from django.views.decorators.debug import sensitive_post_parameters
 from django.views.generic import View
 from social_core.backends.utils import load_backends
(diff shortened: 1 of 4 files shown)


DanSheps (approved):
Looks good!

Issue #14672, opened 2024-01-03
Add support for Python 3.12


jeremystretch:
### Proposed Changes

Document and test for Python 3.12 support in addition to 3.10 and 3.11.

### Justification

Declare support for running NetBox v4.0 on Python 3.12. Django 5.0 also [supports Python 3.12](https://docs.djangoproject.com/en/5.0/releases/5.0/#python-compatibility).

PR #14673, opened 2024-01-03
Closes #14638: Drop support for Python 3.8 and 3.9

### LINKED ISSUE #14638: Drop support for Python 3.8 and 3.9
### Proposed Changes

Increase the minimum supported Python version from 3.8 to 3.10. Remove tests for Python <3.10.

### Justification

We'll be adopting Django 5.0 in NetBox v4.0 (see #14637), which [requires Python 3.10 or later](https://docs.djangoproject.com/en/5.0/releases/5.0/#python-compatibility).

Additionally, Python versions earlier than 3.11 are no longer receiving bug fixes per the [Python release cycle](https://devguide.python.org/versions/).

### Impact

NetBox deployments running Python 3.9 or earlier will need to upgrade their Python version to 3.10 or later.

jeremystretch:
### Fixes: #14638

- Remove CI runners for Python 3.8 and 3.9
- Update documented minimum required Python version to 3.10

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       NETBOX_CONFIGURATION: netbox.configuration_testing
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.10', '3.11']
         node-version: ['14.x']
     services:
       redis:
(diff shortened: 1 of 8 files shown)


arthanson (approved):
Should we add Python 3.12 to this as well to the python-version line?

jeremystretch:
I actually just opened #14672 to track Python 3.12 as a separate initiative. (Wasn't sure if it's going to require any other changes so I wanted to handle it separately.)

markkuleinio:
Note that https://github.com/netbox-community/netbox/blob/feature/upgrade.sh still references Python 3.8

jeremystretch:
@markkuleinio good catch, thanks!

Issue #14637, opened 2023-12-29
Upgrade to Django 5.0


jeremystretch:
### Proposed Changes

Upgrade Django from v4.2 to v5.0.

### Justification

Makes available the [new features provided by Django 5.0](https://docs.djangoproject.com/en/5.0/releases/5.0/).

PR #14675, opened 2024-01-03
14637 update to Django 5

arthanson:
### Fixes: #14637 

Update to use Django 5


netbox/dcim/models/cables.py
```diff
@@ -317,10 +317,14 @@ def clean(self):
...
-        existing_termination = CableTermination.objects.exclude(cable=self.cable).filter(
+        qs = existing_termination = CableTermination.objects.filter(
             termination_type=self.termination_type,
             termination_id=self.termination_id
-        ).first()
+        )
+        if self.cable.pk:
+            qs = qs.exclude(cable=self.cable)
+
+        existing_termination = qs.first()
```
jeremystretch, line 327: Is this change related to supporting Django 5.0? (Also, `existing_termination` is declared twice.)

netbox/dcim/models/devices.py
```diff
@@ -1098,7 +1098,10 @@ def vc_interfaces(self, if_master=True):
 
         :param if_master: If True, return VC member interfaces only if this Device is the VC master.
         """
-        filter = Q(device=self)
+        if self.pk:
+            filter = Q(device=self)
+        else:
+            filter = Q()
```
jeremystretch, line 1104: What prompted this change? Is filtering by None no longer supported? If that's the case I suspect there are many other cases that may need to be updated.

netbox/extras/models/configs.py
```diff
@@ -182,7 +182,7 @@ def get_config_context(self):
 
         if not hasattr(self, 'config_context_data'):
             # The annotation is not available, so we fall back to manually querying for the config context objects
-            config_context_data = ConfigContext.objects.get_for_object(self, aggregate_data=True)
+            config_context_data = ConfigContext.objects.get_for_object(self, aggregate_data=True) or []
```
jeremystretch, line 185: Looks like a potential bug fix? Maybe worth capturing in a separate issue.

netbox/dcim/models/cables.py
arthanson: fixed duplicate assignment. This is from this issue this issue https://forum.djangoproject.com/t/unsaved-model-instances-to-related-filters-in-tests-for-django-5-0/26586/2 from "Passing unsaved model instances to related filters is no longer allowed."

netbox/dcim/models/devices.py
arthanson: this issue https://forum.djangoproject.com/t/unsaved-model-instances-to-related-filters-in-tests-for-django-5-0/26586/2 from "Passing unsaved model instances to related filters is no longer allowed."

netbox/extras/models/configs.py
arthanson: With Django 5 the tests won't pass without this change so included it here.  I wasn't completely sure what change caused this to fail, but it looked like a valid fix so didn't dig into what Django 5 issue caused it.

jeremystretch (approved):

PR #14754, opened 2024-01-09
Fixes #14572 - Constrains JobView (and related views) badge to specific named job

### LINKED ISSUE #14572: Script job count is incorrect
### Deployment Type

Self-hosted

### NetBox Version

v3.6.7

### Python Version

3.10

### Steps to Reproduce

1. Create a script file:

```python


class TestScript(Script):
    name = 'Test'
    description = ''

    def run(self, data, commit=True):
        self.log_success('Finished Script')


class TestScriptNew(Script):
    name = 'Test New'
    description = ''

    def run(self, data, commit=True):
        self.log_success('Finished Script 2')
```
2. Upload to NetBox
3. Run `Test`
5. View `Test New`

### Expected Behavior

Job tab badge has "0"

### Observed Behavior

Job tab badge has >0

DanSheps:
### Fixes: #14572 - Constrains JobView to specific job

* Create new method to return relevant jobs
* Call method to provide 'jobs' key for rendered context data

### DIFF
--- a/netbox/extras/models/mixins.py
+++ b/netbox/extras/models/mixins.py
@@ -9,6 +9,11 @@ __all__ = (
 class PythonModuleMixin:
 
     def get_jobs(self, name):
+        """
+        Returns a list of Jobs associated with this specific script or report module
+        :param name: The class name of the script or report
+        :return: List of Jobs associated with this
+        """
         return self.jobs.filter(
             name=name
         )
--- a/netbox/extras/models/mixins.py
+++ b/netbox/extras/models/mixins.py
@@ -8,6 +8,11 @@ __all__ = (
 
 class PythonModuleMixin:
 
+    def get_jobs(self, name):
+        return self.jobs.filter(
+            name=name
+        )
+
     @property
     def path(self):
         return os.path.splitext(self.file_path)[0]
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -1057,7 +1057,7 @@ class ReportView(ContentTypePermissionRequiredMixin, View):
     def get(self, request, module, name):
         module = get_report_module(module, request)
         report = module.reports[name]()
-        jobs = module.get_jobs(name)
+        jobs = module.get_jobs(report.class_name)
 
         report.result = jobs.filter(
             status__in=JobStatusChoices.TERMINAL_STATE_CHOICES
@@ -1076,7 +1076,7 @@ class ReportView(ContentTypePermissionRequiredMixin, View):
 
         module = get_report_module(module, request)
         report = module.reports[name]()
-        jobs = module.get_jobs(name)
+        jobs = module.get_jobs(report.class_name)
         form = ReportForm(request.POST, scheduling_enabled=report.scheduling_enabled)
 
         if form.is_valid():
@@ -1118,7 +1118,7 @@ class ReportSourceView(ContentTypePermissionRequiredMixin, View):
     def get(self, request, module, name):
         module = get_report_module(module, request)
         report = module.reports[name]()
-        jobs = module.get_jobs(name)
+        jobs = module.get_jobs(report.class_name)
 
         return render(request, 'extras/report/source.html', {
             'job_count': jobs.count(),
@@ -1136,8 +1136,7 @@ class ReportJobsView(ContentTypePermissionRequiredMixin, View):
     def get(self, request, module, name):
         module = get_report_module(module, request)
         report = module.reports[name]()
-        jobs = module.get_jobs(name)
-
+        jobs = module.get_jobs(report.class_name)
 
         jobs_table = JobTable(
             data=jobs,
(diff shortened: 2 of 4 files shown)


netbox/extras/views.py
```diff
@@ -1007,6 +1007,23 @@ def post(self, request, id):
         return redirect(reverse('home'))
 
 
+#
+# Function for returning job data - used in scripts and reports
+#
+
+# Return jobs filtered for job model type (Example: `scriptmodule`)
+def get_jobs(module, job_object, jobtype):
```
jeremystretch, line 1015: Can we move this to a method on the Report & Script classes? That would keep the code closer to the models, and obviate needing to pass `jobs` as context under every view.

(If not, the function should at least be moved to `utils.py`.)

netbox/templates/extras/script/base.html
```diff
@@ -33,7 +33,7 @@
     </li>
     <li class="nav-item" role="presentation">
       <a class="nav-link{% if tab == 'jobs' %} active{% endif %}" href="{% url 'extras:script_jobs' module=script.module name=script.class_name %}">
-        {% trans "Jobs" %} {% badge module.jobs.count %}
+        {% trans "Jobs" %} {%  badge jobs.count %}
```
jeremystretch, line 36:
```suggestion
        {% trans "Jobs" %} {% badge jobs.count %}
```

jeremystretch (changes requested):

netbox/extras/views.py
DanSheps: Yup, definitely, I figured we wanted to reduce the reproduction.  Are you good if I make a Mixin for the classes?

netbox/extras/views.py
```diff
@@ -1057,16 +1057,14 @@ def get_required_permission(self):
...
         module = get_report_module(module, request)
         report = module.reports[name]()
+        jobs = module.get_jobs(name)
 
-        object_type = ContentType.objects.get(app_label='extras', model='reportmodule')
-        report.result = Job.objects.filter(
-            object_type=object_type,
-            object_id=module.pk,
-            name=report.name,
+        report.result = jobs.exclude(
```
jeremystretch, line 1062: Shouldn't this be `filter()`?
```suggestion
        report.result = jobs.filter(
```

netbox/extras/views.py
```diff
@@ -1057,16 +1057,14 @@ def get_required_permission(self):
     def get(self, request, module, name):
         module = get_report_module(module, request)
         report = module.reports[name]()
+        jobs = module.get_jobs(name)
```
jeremystretch, line 1060: I think this should pass `report.class_name` as below.

netbox/extras/views.py
```diff
@@ -1133,13 +1136,9 @@ def get_required_permission(self):
...
 
         object_type = ContentType.objects.get(app_label='extras', model='reportmodule')
```
jeremystretch, line 1141: Doesn't appear to be needed anymore.
```suggestion
```

netbox/extras/views.py
```diff
@@ -1133,13 +1136,9 @@ def get_required_permission(self):
     def get(self, request, module, name):
         module = get_report_module(module, request)
         report = module.reports[name]()
+        jobs = module.get_jobs(name)
```
jeremystretch, line 1139: The prior code is passing `report.class_name`, which IIRC is correct.

netbox/extras/views.py
```diff
@@ -1309,13 +1310,7 @@ def get_required_permission(self):
...
         module = get_script_module(module, request)
         script = module.scripts[name]()
-
-        object_type = ContentType.objects.get(app_label='extras', model='scriptmodule')
-        jobs = Job.objects.filter(
-            object_type=object_type,
-            object_id=module.pk,
-            name=script.class_name
-        )
+        jobs = module.get_jobs(name)
```
jeremystretch, line 1313:
```suggestion
        jobs = module.get_jobs(script.class_name)
```

netbox/extras/views.py
```diff
@@ -1232,19 +1232,16 @@ def get_required_permission(self):
     def get(self, request, module, name):
         module = get_script_module(module, request)
         script = module.scripts[name]()
+        jobs = module.get_jobs(name)
```
jeremystretch, line 1235:
```suggestion
        jobs = module.get_jobs(script.class_name)
```

netbox/extras/models/mixins.py
```diff
@@ -8,6 +8,11 @@
 
 class PythonModuleMixin:
 
+    def get_jobs(self, name):
```
jeremystretch, line 11: Could you add a short docstring here indicating the method's purpose?

jeremystretch (changes requested):
Looks good, but we need to be careful about passing the report/script _class_ name rather than its proper name when fetching jobs.

jeremystretch (approved):
Thanks @DanSheps!

Issue #12851, opened 2023-06-09
Replace Bleach (which is deprecated) with NH3


arthanson:
### Proposed Changes

bleach is deprecated (see: https://github.com/mozilla/bleach/issues/698)
https://github.com/messense/nh3 is a nice replacement

### Justification

Bleach is deprecated, pre notice: "Bleach sits on top of--and heavily relies on--html5lib which is no longer in active development. It is increasingly difficult to maintain Bleach in that context and I think it's nuts to build a security library on top of a library that's not in active development. "

PR #14767, opened 2024-01-10
12851 replace bleach with nh3

arthanson:
### Fixes: #12851 

Replaces the deprecated bleach with NH3.


jeremystretch (approved):
Thanks @arthanson! I took this opportunity to also move the constant definitions outside of `clean_html()`.

Issue #14728, opened 2024-01-08
Move the installed plugins list from the admin UI to the primary UI


jeremystretch:
### NetBox version

v3.7.0

### Feature type

Change to existing functionality

### Proposed functionality

NetBox currently lists all installed plugins under the admin UI (`/admin/plugins/installed-plugins/`). Replicate this view in the primary UI, under the "admin" menu section.

### Use case

Users will be able to inspect plugins without leaving the primary UI, and unblocks #12325.

### Database changes

None

### External dependencies

None

PR #14768, opened 2024-01-11
 14728 Move installed plugins list from admin UI to NetBox UI

arthanson:
### Fixes: #14728

Monosnap Plugins | NetBox 2024-01-10 16-29-56


netbox/netbox/navigation/menu.py
```diff
@@ -454,6 +454,17 @@
                 ),
             ),
         ),
+        MenuGroup(
+            label=_('Plugins'),
+            items=(
+                MenuItem(
+                    link='extras:plugin_list',
+                    link_text=_('Plugins'),
+                    permissions=['core.view_configrevision'],
```
jeremystretch, line 463: We don't have a permission for viewing plugins. It should be sufficient to set `staff_only=True` below.
```suggestion
```

netbox/extras/views.py
```diff
@@ -1373,3 +1375,21 @@ def post(self, request):
...
+#
+# Plugins
+#
+
+class PluginListView(LoginRequiredMixin, View):
+
+    def get(self, request):
+        plugins = [apps.get_app_config(plugin) for plugin in settings.PLUGINS]
+        table = tables.PluginTable(plugins)
+        # table.configure(request)
```
jeremystretch, line 1389: Ideally we should support sorting plugins by name, version, etc.

netbox/extras/urls.py
```diff
@@ -136,4 +136,8 @@
 
     # Markdown
     path('render/markdown/', views.RenderMarkdownView.as_view(), name="render_markdown"),
+
+    # Plugins
+    path('plugins/', views.PluginListView.as_view(), name='plugin_list'),
```
jeremystretch, line 141: This should go under `core` rather than `extras` (we just moved all the plugin resources out of `extras` in v3.7).

netbox/extras/tables/tables.py
```diff
@@ -507,3 +508,17 @@ class Meta(NetBoxTable.Meta):
         default_columns = (
             'pk', 'created', 'created_by', 'assigned_object_type', 'assigned_object', 'kind', 'comments'
         )
+
+
+class PluginTable(tables.Table):
```
jeremystretch, line 513: This should inherit from BaseTable to support reconfiguration.

jeremystretch (changes requested):
Looks good, but we should support ordering & customization of the plugins table.

jeremystretch (approved):

PR #14783, opened 2024-01-12
Add support for displaying port and interface information on a virtual chassis

jschewebbn:
This is in support of fixing the issues discussed at https://github.com/netbox-community/netbox/discussions/12525

In addition to seeing the interfaces and ports on the master device in a virtual chassis this change allows one to view the interfaces and ports for all members on the virtual chassis device itself. This makes it easier to make cable connections from a virtual device as one can find all of the information on the virtual device itself.

By adding ports support to the virtual device one can create virtual patch panels in addition to stacked switches. This is useful when there are a number of physical patch panels in a rack where all ports are uniquely numbered and one needs to connect a cable from that stack of patch panels to another device. One doesn't need to remember which physical patch panel the port is on, only the label of the port.

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1576,6 +1576,17 @@ class FrontPortFilterSet(
     NetBoxModelFilterSet,
     CabledObjectFilterSet
 ):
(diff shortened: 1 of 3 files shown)


abhi1693:

PR #14794, opened 2024-01-15
fix: add missing DH group 15 (3072)

### LINKED ISSUE #14793: DH Group 15 (3072) not selectable in UI
### Deployment Type

Self-hosted

### NetBox Version

v3.7.0

### Python Version

3.11

### Steps to Reproduce

1. Open vpn/ike proposals in UI
2. Try to select DH group 15

### Expected Behavior

A dropdown containing group 15

### Observed Behavior

DH group 15 is missing

jonkerj:
### Fixes #14793

DH group 15 was not selectable in the UI, and I strongly suspect this PR will fix that, as that particular choices was missing in `choices.py`.


jeremystretch (approved):
Thanks @jonkerj!

PR #14804, opened 2024-01-16
Fixes #14755: ValueError in web UI after REST API accepts invalid cus…

### LINKED ISSUE #14755: ValueError in web UI after REST API accepts invalid custom-field choice-set data
### Deployment Type

Self-hosted

### NetBox Version

v3.7.0

### Python Version

3.10

### Steps to Reproduce

1. Start with a fresh NetBox 3.7.0 installation and database
2. Post a custom-field choice-set whose choices are lists with just one element each (should be two elements each):

```
curl -H "Authorization: Token ${NETBOX_TOKEN}" \
  'http://192.168.64.3/api/extras/custom-field-choice-sets/' \
  -H 'accept: application/json'   -H 'Content-Type: application/json' \
  -d '[ { "name": "test1", "extra_choices": [ ["choice1"], ["choice2"] ] } ]'
```

3. Log in to the web UI, navigate to _Customization_ -> _Custom Field Choices_, and click on the __test1__ entry

### Expected Behavior

The REST API should reject the POST data as invalid

**AND/OR**

The web UI should render a dialog showing that the choice set's choices have undefined labels

### Observed Behavior

The REST API cheerfully creates the choice set with the choice labels undefined.

**AND**

The web UI displays an unstyled `ValueError`:

```
<class 'ValueError'>

Need 2 values to unpack in for loop; got 1. 

Python version: 3.10.12
NetBox version: 3.7.0
Plugins: None installed
```

This problem was originally reported by @jordanrvillarreal.

Julio-Oliveira-Encora:
…tom-field choice-set data

### Fixes: #14755
Currently, the endpoint "custom-field-choice-sets" accepts one element each (should be two elements each). I changed the Serializer to check the payload and return 400 if it's malformed.


netbox/extras/api/serializers.py
```diff
@@ -179,6 +179,15 @@ class Meta:
             'choices_count', 'created', 'last_updated',
         ]
 
+    def validate_extra_choices(self, value):
```
jeremystretch, line 182: While this approach provides validation, it doesn't allow for easy introspection. We should declare `extra_choices` as a [ListField](https://www.django-rest-framework.org/api-guide/fields/#listfield) on the serializer, with a child ListField which requires exactly two items.

netbox/extras/api/serializers.py
```diff
@@ -374,7 +383,8 @@ def validate(self, data):
 
     @extend_schema_field(serializers.JSONField(allow_null=True))
     def get_assigned_object(self, instance):
-        serializer = get_serializer_for_model(instance.assigned_object_type.model_class(), prefix=NESTED_SERIALIZER_PREFIX)
+        serializer = get_serializer_for_model(instance.assigned_object_type.model_class(),
```
jeremystretch, line 386: Please revert this unrelated change.

netbox/extras/tests/test_api.py
```diff
@@ -14,14 +14,12 @@
 from extras.scripts import BooleanVar, IntegerVar, Script, StringVar
 from utilities.testing import APITestCase, APIViewTestCases
 
-
```
jeremystretch, line 17: Please revert all these arbitrary formatting changes. No changes to this file are needed.

netbox/extras/tests/test_custom_field_choice_sets_endpoint.py
```diff
@@ -0,0 +1,59 @@
+from django.contrib.auth import get_user_model
+from rest_framework.test import APITestCase
+
+from users.models import Token
+
+User = get_user_model()
+
+
+class CustomFieldChoiceSetsEndpointTest(APITestCase):
```
jeremystretch, line 9: All tests for this endpoint should go in `extras.tests.test_api.CustomFieldChoiceSetTest`.

jeremystretch (changes requested):

jeremystretch (approved):

PR #14812, opened 2024-01-17
Fixes #14655: Document raw text configuration render

### LINKED ISSUE #14655: Document method for rendering configurations as raw text
### Change Type

Addition

### Area

Features

### Proposed Changes

The current documentation for [configuration rendering](https://docs.netbox.dev/en/stable/features/configuration-rendering/) explains how a REST API request can be made to render a configuration template, which will return a JSON response. This are should be extended to also note how the rendered configuration can be returned in plaintext (by setting the `Accept: text/plain` HTTP header).

bluikko:
### Fixes: #14655

Add simple explanation how to use HTTP headers to render configuration
templates as either JSON or plaintext.

Also while there fix a nearby typo where "object" was typed as "objet".

### DIFF
--- a/docs/features/configuration-rendering.md
+++ b/docs/features/configuration-rendering.md
@@ -39,7 +39,7 @@ When rendered for a specific NetBox device, the template's `device` variable wil
 
 ### Context Data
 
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @bluikko!

PR #14868, opened 2024-01-19
Limits ip addresses on interface tables

### LINKED ISSUE #14645: Add option to limit IP Addresses displayed in interface tables
### NetBox version

v3.6.9

### Feature type

### Proposed functionality

Allow to limit the list of IP Addresses to e.g. 10, so interfaces with many assigned IPs won't clog the table.

### Use case

When having e.g. NAT interfaces with 256 assigned IPs on a central firewall on a virtual interface this will clog the table, e.g. making it hard to find the next interface and also the name for the current interface. So an option to limit the viewed addresses would be handy.

Possible look, when limited to 10 addresses:
image

### Database changes

N/A

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/dcim/tables/template_code.py
+++ b/netbox/dcim/tables/template_code.py
@@ -36,13 +36,17 @@ DEVICEBAY_STATUS = """
 
 INTERFACE_IPADDRESSES = """
 <div class="table-badge-group">
-  {% for ip in value.all %}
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @abhi1693!

PR #14869, opened 2024-01-19
Cascades event rule when the associated script is deleted

### LINKED ISSUE #14852: Event Rule breaks when script is deleted
### Deployment Type

Self-hosted

### NetBox Version

v3.7.0

### Python Version

3.8

### Steps to Reproduce

1. Create a Script
2. Create a EventRule linked with the created Script
3. Delete the Script
4. Try to open the EventRule

### Expected Behavior

The Script deletion cascades the deletion of the EventRule

### Observed Behavior

```
<class 'django.urls.exceptions.NoReverseMatch'>

Reverse for 'script' with keyword arguments '{'module': '', 'name': 'EventRule'}' not found. 1 pattern(s) tried: ['extras/scripts/(?P<module>[^/]+)/(?P<name>[^/]+)/\\Z']

Python version: 3.10.13
NetBox version: 3.7.0
```

abhi1693:

### DIFF
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -2,6 +2,7 @@ import inspect
 import logging
 from functools import cached_property
 
+from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
@@ -79,3 +80,8 @@ class ScriptModule(PythonModuleMixin, JobsMixin, ManagedFile):
     def save(self, *args, **kwargs):
         self.file_root = ManagedFileRootPathChoices.SCRIPTS
         return super().save(*args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        # Delete all associated event rules
+        ContentType.objects.get(app_label=self._meta.app_label, model=self._meta.model_name).eventrule_actions.all().delete()
+        super().delete(*args, **kwargs)


jeremystretch (changes requested):
Rather than handling this manually, a generic relation should be added to the ScriptModule class to automate the cascading deletion:

```python
event_rules = GenericRelation(
    to='extras.EventRule',
    content_type_field='action_object_type',
    object_id_field='action_object_id',
    for_concrete_model=False
)
```

This ensures that the user is informed that the event rule will be deleted when confirming deletion of the script module.

abhi1693:
I have added the `events_rules` like suggested but there is an issue that I'm facing with the code. The `handle_deleted_object` signal gets called when I try to delete the script and the code fails with the following error

```
  File "/home/asaharan/PycharmProjects/netbox/netbox/extras/api/serializers.py", line 90, in get_action_object
    script = instance.action_object.scripts[script_name]()
KeyError: 'AddDeviceTypeComponents'
```

I checked the that `instance.action_object` points to the `ScriptModule` but calling `scripts` just returns an empty dictionary which is the reason for the KeyError. How do you want me to handle this? This issue was not with the delete overide I did though

arthanson:
@abhi1693 I haven't dug too deeply into this bug, but normally scripts should have the script_name in it, however if this is from the delete case I can think of two potential solutions:
1. in the serializer check if script_name is in scripts and if not return a null object for the nested serializer.
2. I think you said this comes from the call to handle_deleted_object? If that is the case then possibly not do the serializer call if there are no script objects?  (not sure on this one).

If you could message me the repro and the line of code in the handle_deleted_object where this is getting called it would help to get context.  I think the main issue is why is scripts returning none if the action_parameter is set, which shouldn't really be happening.  If the scripts is returning none from deletion. then the action_parameter should be empty as well.  The action_parameter is the name of the actual method to call as the action_object just points to the script model and not the actual method to call.  I hope that helps.

abhi1693:
@arthanson The repro steps are simple:

1. Add the code Jeremy has suggested
2. Upload a script. I chose [add_device_type_components](https://github.com/netbox-community/customizations/blob/master/scripts/add_device_type_components.py)
3. Create an event rule with this script (choose whatever required options)
4. Then try deleting the script. You will get a nice widget showing the event and the relationship will be deleted, proceed further and you will get hit with the issue

arthanson:
closing as addressed in #15888

PR #14870, opened 2024-01-19
Fixes #14829 Simple condition (without and/or) does not work in event rule

### LINKED ISSUE #14829: Simple condition (without and/or) does not work in event rule
### Deployment Type

Self-hosted

### NetBox Version

v3.7.0

### Python Version

3.11

### Steps to Reproduce

1. Create webhook: Name = Test, URL = http://127.0.0.1:9000 (doesn't matter in this case, it won't be triggered but is required to configure event rule)
2. Go to **Event rules - Add**:
- Name = Test
- Content types = Circuit
- select Updates
- set Conditions:
```
{
    "attr": "status.value",
    "value": "active"
}
```

- Action type = Webhook
- Webhook = Test
- **Create**

### Expected Behavior

Event rule is created

### Observed Behavior

Error is shown about the condition:

**Ruleset must have exactly one logical operator (found 2)** 

The examples in https://docs.netbox.dev/en/stable/reference/conditions/ look the same: simple JSON object with attributes `attr` and `value`.

Julio-Oliveira-Encora:
### Fixes: #14829

This change allows the addition of simple conditions in the Event Rule. Currently, it only allows to add conditions with logical operators.


netbox/extras/conditions.py
```diff
@@ -132,20 +132,26 @@ class ConditionSet:
     def __init__(self, ruleset):
         if type(ruleset) is not dict:
             raise ValueError(f"Ruleset must be a dictionary, not {type(ruleset)}.")
-        if len(ruleset) != 1:
+        if len(ruleset) < 1:
             raise ValueError(f"Ruleset must have exactly one logical operator (found {len(ruleset)})")
```
markkuleinio, line 136: This need some adjustment (the check effectively catches empty dicts)

Julio-Oliveira-Encora: I think this validation could be removed, could not?

markkuleinio: I see in https://github.com/netbox-community/netbox/blob/develop/netbox/extras/models/models.py there are already checks for empty conditions, so I'd just remove the `len(ruleset)` check.

netbox/extras/conditions.py
```diff
@@ -130,22 +130,30 @@ class ConditionSet:
...
-            raise ValueError(f"Invalid logic type: {logic} (must be '{AND}' or '{OR}')")
-        self.logic = logic.lower()
-
-        # Compile the set of Conditions
-        self.conditions = [
-            ConditionSet(rule) if is_ruleset(rule) else Condition(**rule)
-            for rule in ruleset[self.logic]
-        ]
+
+        self.logic = None
```
jeremystretch, line 137: It should not be necessary to set `logic` as an instance attribute.

netbox/extras/conditions.py
```diff
@@ -130,22 +130,30 @@ class ConditionSet:
...
+
+        # If logic type use it, else return the ruleset
+        if len(ruleset) == 1:
```
jeremystretch, line 140: We can use the `is_ruleset()` function, which provides some additional validation, to determine how to treat the root dictionary.

netbox/extras/conditions.py
```diff
@@ -130,22 +130,30 @@ class ConditionSet:
...
-
-        # Compile the set of Conditions
-        self.conditions = [
-            ConditionSet(rule) if is_ruleset(rule) else Condition(**rule)
-            for rule in ruleset[self.logic]
-        ]
+
+        self.logic = None
+
+        # If logic type use it, else return the ruleset
```
jeremystretch, line 139: This comment doesn't quite make sense. It should explain the distinction being made between an individual rule and a rule set.

netbox/extras/tests/test_event_rules.py
```diff
@@ -376,3 +377,93 @@ def dummy_send(_, request, **kwargs):
         # Patch the Session object with our dummy_send() method, then process the webhook for sending
         with patch.object(Session, 'send', dummy_send) as mock_send:
             send_webhook(**job.kwargs)
+
+    def test_event_rule_conditions_without_logic_operator(self):
```
jeremystretch, line 381: This test case is for testing the processing of event rules. Any tests related to conditions or condition sets specifically should go in `extras.tests.test_conditions.py`. Note that there are two cases in that file: One for individual conditions and one for condition sets.

jeremystretch (changes requested):

netbox/extras/conditions.py
Julio-Oliveira-Encora: I removed it.

netbox/extras/conditions.py
Julio-Oliveira-Encora: I keep  "self.logic" as an instance because it is used in the "eval" method.

netbox/extras/conditions.py
Julio-Oliveira-Encora: I thought to use it, but we could lose the granularity to deal with the errors and the correct error message.

netbox/extras/tests/test_event_rules.py
Julio-Oliveira-Encora: What am I supposed to do with them? Move them to extras.tests.test_conditions.py?

jeremystretch:
@Julio-Oliveira-Encora would you mind tackling the merge conflicts here? It should just be changes from enabling translation of the error messages.

Julio-Oliveira-Encora:
> @Julio-Oliveira-Encora would you mind tackling the merge conflicts here? It should just be changes from enabling translation of the error messages.

Done!

netbox/extras/conditions.py
```diff
@@ -135,23 +135,24 @@ class ConditionSet:
...
-        # Compile the set of Conditions
-        self.conditions = [
-            ConditionSet(rule) if is_ruleset(rule) else Condition(**rule)
-            for rule in ruleset[self.logic]
-        ]
+        if len(ruleset) == 1:
+            self.logic = (list(ruleset.keys())[0]).lower()
+            if self.logic not in (AND, OR):
+                raise ValueError(
+                    f"Invalid logic type: {self.logic} (must be '{AND}' or '{OR}'). Please check documentation.")
```
arthanson, line 143: Should be internationalized, also can just hard-code in the and, or, doesn't need to be an f-string

netbox/extras/conditions.py
```diff
@@ -135,23 +135,24 @@ class ConditionSet:
...
+            self.conditions = [
+                ConditionSet(rule) if is_ruleset(rule) else Condition(**rule)
+                for rule in ruleset[self.logic]
+            ]
+        else:
+            try:
+                self.logic = None
+                self.conditions = [Condition(**ruleset)]
+            except TypeError:
+                raise ValueError(f"Incorrect key(s) informed. Please check documentation.")
```
arthanson, line 155: internationalize this, doesn't need to be an f-string

netbox/extras/tests/test_event_rules.py
```diff
@@ -13,6 +13,7 @@
 from dcim.models import Site
 from extras.choices import EventRuleActionChoices, ObjectChangeActionChoices
 from extras.events import enqueue_object, flush_events, serialize_for_event
+from extras.forms import SavedFilterForm, EventRuleForm
```
arthanson, line 16: These aren't used anywhere

netbox/extras/conditions.py
```diff
@@ -135,23 +135,24 @@ class ConditionSet:
...
-                logic=logic, op_and=AND, op_or=OR
-            ))
-        self.logic = logic.lower()
 
-        # Compile the set of Conditions
-        self.conditions = [
-            ConditionSet(rule) if is_ruleset(rule) else Condition(**rule)
-            for rule in ruleset[self.logic]
-        ]
+        if len(ruleset) == 1:
```
arthanson, line 139: Is this the correct order below - it seems like they are reversed, the and-or processing is being done if the len(ruleset) == 1 which seems like it should be in the else clause?

arthanson (changes requested):

netbox/extras/conditions.py
Julio-Oliveira-Encora: It's in the correct order because of the ruleset below, for example, has length = 1 and this case is deal inside the if. All other conditions must be deal in the else. Does it make sense?
```
{
    "or": [
        {
            "and": [
                {
                    "attr": "status.value",
                    "value": "active"
                },
                {
                    "attr": "primary_ip4",
                    "negate": true,
                    "value": null
                }
            ]
        },
        {
            "attr": "tags.slug",
            "op": "contains",
            "value": "exempt"
        }
    ]
}
```

netbox/extras/conditions.py
```diff
@@ -135,23 +135,24 @@ class ConditionSet:
...
+            self.logic = (list(ruleset.keys())[0]).lower()
+            if self.logic not in (AND, OR):
+                raise ValueError(_(
+                    f"Invalid logic type: {self.logic} (must be 'AND' or 'OR'). Please check documentation."))
```
arthanson, line 143: remove the f-string here, f strings can't be internationalized

arthanson (changes requested):

arthanson (approved):

Issue #12795, opened 2023-06-01
Implement a custom User and Group Model


arthanson:
### NetBox version

v3.5.2

### Feature type

Data model extension

### Proposed functionality

Implement and use a custom User and Group model. 

### Use case

This would allow us to extend the User and Group model in the future (like adding a description to the Group model, which would be useful).  It would also make them more compatible with the rest of the codebase allowing better use of the custom list and bulk views as we are moving the admin of those items from Django admin to NetBox.

### Database changes

would need custom migrations for these new models.

arthanson:
blocking on #13647

jeremystretch:
Re-openining this to tackle a custom group model as well.

PR #14881, opened 2024-01-20
Fixes: #14840 - Forces API to use django user model instead of proxy model

### LINKED ISSUE #14840: Reverse for 'netboxuser-list' not found when editing
### Deployment Type

Self-hosted

### NetBox Version

3.7.0

### Python Version

3.9

### Steps to Reproduce

Create a new custom field: `type:object, object-type: users>user`.

### Expected Behavior

When editing a device/inventory-item, you can select a netbox user within the specified custom field.

### Observed Behavior

When editing to sleect a netbox user, I get the following error message:

```
<class 'django.urls.exceptions.NoReverseMatch'>

Reverse for 'netboxuser-list' not found. 'netboxuser-list' is not a valid view function or pattern name.

Python version: 3.9.2
NetBox version: 3.7.0
Plugins: None installed
```

DanSheps:
### Fixes: #14840 - Forces API to use django user model instead of proxy model

* Swap the model name in get_viewname for only NetBoxUser and NetBoxGroup

### DIFF
--- a/netbox/users/api/urls.py
+++ b/netbox/users/api/urls.py
@@ -9,6 +9,7 @@ router.APIRootView = views.UsersRootView
 
 # Users and groups
 router.register('users', views.UserViewSet)
+router.register('netboxusers', views.NetBoxUserViewSet)
 router.register('groups', views.GroupViewSet)
 
 # Tokens
--- a/netbox/users/api/urls.py
+++ b/netbox/users/api/urls.py
@@ -9,7 +9,6 @@ router.APIRootView = views.UsersRootView
 
 # Users and groups
 router.register('users', views.UserViewSet)
-router.register('netboxusers', views.NetBoxUserViewSet)
 router.register('groups', views.GroupViewSet)
 
 # Tokens
--- a/netbox/users/api/views.py
+++ b/netbox/users/api/views.py
@@ -15,7 +15,7 @@ from rest_framework.viewsets import ViewSet
 
 from netbox.api.viewsets import NetBoxModelViewSet
 from users import filtersets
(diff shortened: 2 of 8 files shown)


DanSheps:
I think this will be a much more acceptable resolution to the problem, it does introduce another API endpoint (but maintains all the filters from `/users/`, I couldn't find a clean way of customizing the "basename" without modifying NetBoxRouter).

Alternatively, we could swap the proxy model in the ContentTypeManager for the actual user model

jeremystretch:
We shouldn't add a redundant API endpoint for the same model. Instead, we can tweak the logic in `get_viewname()` to resolve the view for the concrete model (User) instead of the proxy model (NetBoxUser). There are a few other places where we use proxy models though, so we'll need to check that those are not broken as a result of this change.

Ultimately this problem should be side-stepped by the introduction of a proper custom user model in v4.0 under #12795.

DanSheps:
> We shouldn't add a redundant API endpoint for the same model. Instead, we can tweak the logic in `get_viewname()` to resolve the view for the concrete model (User) instead of the proxy model (NetBoxUser). There are a few other places where we use proxy models though, so we'll need to check that those are not broken as a result of this change.

You are absolutely correct, that is a much easier way to handle that.  Made a tweak to swap the proxy model to the user model and it looks to be good now.

jeremystretch:
This doesn't appear to fully resolve #14840. Although you can now edit an object with a custom field pointing to the User model, attempting to save it raises a `SerializerNotFound` exception:

```
Could not determine serializer for users.NetBoxUser with prefix 'Nested'
```

jeremystretch (approved):
Thanks @DanSheps!

PR #14901, opened 2024-01-23
Fixes #14838: JSONField should treat initial form data as JSON

### LINKED ISSUE #14838: Changing event rule action type messes up Conditions field
### Deployment Type

Self-hosted

### NetBox Version

v3.7.1

### Python Version

3.11

### Steps to Reproduce

1. Go to **Operations - Event Rules - Add**
2. Set Conditions = `{ "and": [{"attr": "status.value", "value": "deprecated"}]}`
3. Select Action type = Script

### Expected Behavior

Conditions field stays as-is

### Observed Behavior

Conditions field is rewritten as invalid JSON:

`"{ \"and\": [{\"attr\": \"status.value\", \"value\": \"deprecated\"}]}"`

The quoting is added with every change of the Action type.

This also happens if editing an existing event rule and changing the Action type.

### Workaround

- Copy the Conditions field somewhere
- Change the Action type
- Paste the Conditions field back

jeremystretch:
### Fixes: #14838

If the field's initial value is a string, attempt to load it as JSON. Also updates the call to `dumps()` afterward with additional arguments from the parent class' implementation.

### DIFF
--- a/netbox/utilities/forms/fields/fields.py
+++ b/netbox/utilities/forms/fields/fields.py
@@ -105,7 +105,12 @@ class JSONField(_JSONField):
             return value
         if value in ('', None):
             return ''
-        return json.dumps(value, sort_keys=True, indent=4)
+        if type(value) is str:
+            try:
+                value = json.loads(value, cls=self.decoder)
+            except json.decoder.JSONDecodeError:
+                return value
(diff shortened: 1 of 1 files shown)


DanSheps (approved):
Looks reasonable.

Tested and resolves the issue

LGTM

PR #14903, opened 2024-01-23
Fixes #14703: Catch exceptions when rendering dashboard and revert to default

### LINKED ISSUE #14703: User Dashboard can become permanently broken when missconfiguring DEFAULT_DASHBOARD
### Deployment Type

Self-hosted

### NetBox Version

v3.7.0

### Python Version

3.9

### Steps to Reproduce

1. User johndoe modifies their dashboard over the GUI
2. Administrator changes DEFAULT_DASHBOARD in configuration.py with a broken configuration (Assigned a wrong value in a bookmarks widget)
3. Bookmark widget looks fine without logging in since it won't load any data
4. After johndoe logs in, they reset their dashboard
5. Site crashes with an Field Error Exception since the dashboard is misconfigured
6. johndoe can't reset their dashboard anymore, even after the error in DEFAULT_DASHBOARD is corrected

### Expected Behavior

When resetting the dashboard, I would expect it to now always follow the changes of DEFAULT_DASHBOARD.

### Observed Behavior

Resetting the dashboard seems to copy the current state of DEFAULT_DASHBOARD. If that state is broken, a user cannot reset it anymore.
Only the main page with the dashboard crashes.

jeremystretch:
### Fixes: #14703

- Extend `get_default_dashboard()` to allow passing a specific dashboard configuration
- Wrap the call to `get_layout()` in a `try`/`except` and fall back to the (hard-coded) default config

### DIFF
--- a/netbox/extras/dashboard/utils.py
+++ b/netbox/extras/dashboard/utils.py
@@ -53,13 +53,13 @@ def get_dashboard(user):
     return dashboard
 
 
-def get_default_dashboard():
+def get_default_dashboard(config=None):
     from extras.models import Dashboard
 
     dashboard = Dashboard()
-    default_config = settings.DEFAULT_DASHBOARD or DEFAULT_DASHBOARD
+    config = config or settings.DEFAULT_DASHBOARD or DEFAULT_DASHBOARD
 
-    for widget in default_config:
+    for widget in config:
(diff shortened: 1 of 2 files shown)


DanSheps (approved):
Looks good to me.

Tested followinng repro steps in #14703

PR #14976, opened 2024-01-30
12510 Merge Scripts and Reports

### LINKED ISSUE #12510: Remove support for reports in NetBox v4.0
### Proposed Changes

Remove all support for custom reports. Support for custom scripts will remain in place.

This change is being proposed for NetBox v4.0 to allow users ample time to prepare for the migration.

### Justification

There's a lot of overlap in functionality between reports and scripts, and dropping reports in favor of scripts will allow us to simplify the user experience and remove a substantial amount of largely redundant code.

### Impact

All reports will need to be converted to scripts, which should be fairly straightforward. We can produce supporting documentation to assist users with this process.

arthanson:
### Fixes: #12510 

Combined scripts and reports and adds the report summary of logs to scripts.  

**Note:** used unassigned for logs that are in the main or have no method assigned (report logs all are tied to the test_methods).  Not really liking the 'unassigned' name, alternative suggestions welcome.

The old job data record is different for reports and scripts, created a new one for the combined and put in legacy_ templates for the older ones.

Monosnap Scripts | NetBox 2024-01-30 08-25-38


netbox/core/migrations/0011_job_report_to_script.py
jeremystretch, line 1: Can we move this logic to the migration in `extras`? IMO there's no need for it to live in `core`.

netbox/extras/api/serializers.py
```diff
@@ -593,22 +592,6 @@ def validate_interval(self, value):
         return value
 
 
-class ScriptLogMessageSerializer(serializers.Serializer):
```
jeremystretch, line 596: This is still used by the `runscript` management command.

netbox/extras/migrations/0106_delete_report_delete_reportmodule.py
jeremystretch, line 1: What is the migration path for existing reports? Shouldn't we convert them?

netbox/extras/utils.py
```diff
@@ -67,3 +67,15 @@ def is_report(obj):
         return issubclass(obj, Report) and obj != Report
     except TypeError:
         return False
+
+
+def is_script_or_report(obj):
```
jeremystretch, line 72: This name is confusing, as reports will no longer exist with this change in place. IMO we should retain & expand the original `is_script()` function instead of introducing a new function.

Also, the `is_report()` function is now obsolete and can be removed.

netbox/extras/views.py
```diff
@@ -1343,11 +1163,21 @@ def get(self, request, job_pk):
         module = job.object
         script = module.scripts[job.name]()
 
+        legacy_script = False
+        legacy_report = False
+        if job.data and ('logs' not in job.data):
+            if 'log' in job.data:
+                legacy_script = True
+            else:
+                legacy_report = True
```
jeremystretch, line 1172: I'm not clear on the concept of a legacy script: We should avoid introducing this distinction if at all possible.

netbox/extras/scripts.py
```diff
@@ -257,7 +257,7 @@ def __init__(self, min_prefix_length=None, max_prefix_length=None, *args, **kwar
 # Scripts
 #
 
-class BaseScript:
+class BaseScript(object):
```
jeremystretch, line 260:
```suggestion
class BaseScript:
```

netbox/extras/scripts.py
```diff
@@ -399,25 +434,54 @@ def as_form(self, data=None, files=None, initial=None):
 
     # Logging
 
-    def log_debug(self, message):
-        self.logger.log(logging.DEBUG, message)
-        self.log.append((LogLevelChoices.LOG_DEFAULT, str(message)))
+    def _log(self, message, obj=None, log_level=LogLevelChoices.LOG_DEFAULT, level=logging.INFO):
```
jeremystretch, line 437: It's very confusing to pass both `log_level` and `level`. Can't we infer one from the other?

netbox/extras/scripts.py
```diff
@@ -446,6 +510,44 @@ def load_json(self, filename):
 
         return data
 
+    def run_test_scripts(self, job):
```
jeremystretch, line 513: Maybe `run_tests()` would be a better name for this? There's only one script involved (the class).

netbox/extras/scripts.py
```diff
@@ -446,6 +510,44 @@ def load_json(self, filename):
...
+        """
+        Run the report and save its results. Each test method will be executed in order.
+        """
+        self.logger.info(f"Running report")
+
+        # Perform any post-run tasks
+        self.pre_run()
```
jeremystretch, line 520: Let's move `pre_run()` and `post_run()` out of this method and into `run()`, to better separate functions.

netbox/extras/scripts.py
```diff
@@ -500,6 +602,61 @@ def run_script(data, job, request=None, commit=True, **kwargs):
     # Add the current request as a property of the script
     script.request = request
 
+    def signature(fn):
+        """
+        Taken from django.tables2 (https://github.com/jieter/django-tables2/blob/master/django_tables2/utils.py)
```
jeremystretch, line 607: We can't reuse code from other projects. Let's rewrite these functions from scratch to serve our specific needs.

jeremystretch (changes requested):
Did a quick first pass on this. Please also add some preliminary documentation noting the move from reports to scripts and providing a high-level migration guide (bulletpoints are fine).

netbox/extras/migrations/0106_delete_report_delete_reportmodule.py
arthanson: The ContentType is moved so they appear under Scripts and scripts can display legacy reports since we can't change the users code so they are auto-converted.

netbox/extras/views.py
arthanson: Added in the summary (success, failure, error) of reports to scripts as part of the merge, the old scripts data is in a truncated format so isn't compatible without doing this.

netbox/extras/migrations/0106_delete_report_delete_reportmodule.py
jeremystretch: Ah right, because `ReportModule` and `ScriptModule` are both proxies for `core.ManagedFile`; they already live in the same table.

netbox/extras/scripts.py
arthanson: For legacy reports log function (just plain log not log_debug or log_exception) if we change that from going to [logger.info](http://logger.info/) to logger.debug then I can make a map.  Right now both log and log_debug use LogLevelChoices.LOG_DEFAULT but log goes to info and log_debug goes to debug so there isn't a 1:1 mapping, but if we can change this then can make one map to the other.

netbox/extras/scripts.py
```diff
@@ -331,6 +354,21 @@ def job_timeout(self):
...
+    @property
+    def source(self):
+        return inspect.getsource(self.__class__)
+
+    @property
+    def is_valid(self):
+        """
+        Indicates whether the report can be run.
+        """
+        return bool(self.test_methods)
```
jeremystretch, line 370: I'm not sure this makes sense to keep around. IIRC `is_valid` was used to determine whether a Report had at least one test method defined. This is not a prerequisite for running a script.

netbox/extras/scripts.py
```diff
@@ -446,6 +511,41 @@ def load_json(self, filename):
...
+            self.post_run()
+            self._current_method = ''
+            raise e
+
+        self._current_method = ''
+
+    def run(self, data, commit):
+        self.pre_run()
+        self.run_tests()
+        self.post_run()
```
jeremystretch, line 535: We should move `pre_run()` and `post_run()` to the `run_script()` function, to ensure that they are always called if `run()` is overridden.

netbox/extras/utils.py
```diff
@@ -51,9 +51,13 @@ def is_script(obj):
     """
     Returns True if the object is a Script.
     """
+    """
+    Returns True if the given object is a Report.
+    """
```
jeremystretch, line 56:
```suggestion
    """
    Returns True if the object is a Script or Report
    """
```

jeremystretch (changes requested):

netbox/extras/scripts.py
jeremystretch: On second thought, let's leave these as-is. They should be considered legacy report functions and ultimately will be phased out.

jeremystretch (approved):
I think we've gotten this to a point where it's reasonable to merge.

FYI I've opened #15072 to improve the script execution logic and address flow issues that existed before and are outside the scope of this FR.

PR #14986, opened 2024-01-30
14947 fix for missing changelog if only update m2m

### LINKED ISSUE #14947: Changelog entry missing for tag change
### Deployment Type

Self-hosted

### NetBox Version

v3.7.1

### Python Version

3.11

### Steps to Reproduce

1. Edit an Interface / Device / Site
2. Add / remove a tag

Reproducible within demo.netbox.dev.

### Expected Behavior

After adding / removing the tag I would expect to see a changelog entry. 

### Observed Behavior

No changelog entry is created.

arthanson:
### Fixes: #14947 

Fix for missing changelog if only update m2m (tag).  Issue is if you only update tags (no other edits) then the changelog entry for the post_save (not m2m_changed) is not created because there are no changes at that point.  When the m2m_changed comes in the update does not work because the query is empty.

Fix is if there is an m2m change and there isn't an existing objectchange then create one, otherwise do the update.

### DIFF
--- a/netbox/extras/signals.py
+++ b/netbox/extras/signals.py
@@ -69,6 +69,7 @@ def handle_changed_object(sender, instance, **kwargs):
         return
 
     # Record an ObjectChange if applicable
+    objectchange = instance.to_objectchange(action)
     if m2m_changed:
         qs = ObjectChange.objects.filter(
             changed_object_type=ContentType.objects.get_for_model(instance),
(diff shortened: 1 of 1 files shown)


netbox/extras/signals.py
```diff
@@ -70,13 +70,21 @@ def handle_changed_object(sender, instance, **kwargs):
...
+        if not qs:
+            objectchange = instance.to_objectchange(action)
+            if objectchange and objectchange.has_changes:
+                objectchange.user = request.user
+                objectchange.request_id = request.id
+                objectchange.save()
+        else:
+            qs.update(
+                postchange_data=instance.to_objectchange(action).postchange_data
+            )
```
jeremystretch, line 87: We can clean this up a bit using the [`update_or_create()`](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#update-or-create) queryset method to automatically create a new change record where needed. We can also pull `objectchange = instance.to_objectchange(action)` up to the top, as it appears to be needed under every condition.

jeremystretch (changes requested):
We should also extend the change logging tests to check for a condition where _only_ the tags of an item are changed.

netbox/extras/signals.py
arthanson: I moved the objectchange = line.

For update_or_create I don't think that would be any cleaner as on create you are doing objectchange.save and not specifying all the fields.  For update_or_create you would need to specify all the fields for defaults and create_defaults which is sort of duplicating instance.to_objectchange.

jeremystretch (approved):

PR #14992, opened 2024-01-31
Issue #14962 VM to merge directly related site context

### LINKED ISSUE #14962: VM Rendered Config Context does not consider directly related site
### Deployment Type

Self-hosted

### NetBox Version

v3.7.1

### Python Version

3.8

### Steps to Reproduce

1. Create a site
2. Create a Config Context and apply it to new site (step 1)
3. Create a VM and set its site to the one created above
4. DO NOT assign the VM to a cluster
5. Inspect the rendered context in the UI.  Note that the site config shows as a source context but it is not merged into the rendered context.
6. Create a VM Cluster and assign its site to that created in step 1.
7. Update the VM to be a member of that cluster
8. Note that the rendered context for the VM now includes the sites config context

### Expected Behavior

The VMs config context should include the sites context when it is directly related via the VM's site attribute with out requiring the cluster.

### Observed Behavior

The VM context does not include the site context when it is directly related and not accessible via a cluster

ChrisPortman:
### Fixes: #14962 

Update the code such that site attached context data is accessed via the site attribute on the virtual machine rather than the site attribute on the optionally associated cluster.
Noting that the `full_clean` of a virtual machine will transpose the cluster's site directly onto the virtual machine if it is not already set.

### DIFF
--- a/netbox/extras/tests/test_models.py
+++ b/netbox/extras/tests/test_models.py
@@ -270,7 +270,12 @@ class ConfigContextTest(TestCase):
         tag = Tag.objects.first()
         cluster_type = ClusterType.objects.create(name="Cluster Type")
         cluster_group = ClusterGroup.objects.create(name="Cluster Group")
-        cluster = Cluster.objects.create(name="Cluster", group=cluster_group, type=cluster_type)
+        cluster = Cluster.objects.create(
+            name="Cluster",
+            group=cluster_group,
+            type=cluster_type,
+            site=site,
+        )
 
         region_context = ConfigContext.objects.create(
             name="region",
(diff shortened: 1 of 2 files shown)


netbox/extras/tests/test_models.py
```diff
@@ -354,6 +359,106 @@ def test_annotation_same_as_get_for_object_virtualmachine_relations(self):
...
+            site=site,
+            tenant=tenant,
+            platform=platform,
+            role=DeviceRole.objects.first()
+        )
+        virtual_machine.tags.add(tag)
+
+        annotated_queryset = VirtualMachine.objects.filter(name=virtual_machine.name).annotate_config_context_data()
+
+        self.assertEqual(virtual_machine.get_config_context(), annotated_queryset[0].get_config_context())
```
jeremystretch, line 460: This doesn't actually validate that the rendered config context is correct for the VM, just that its consistent between the two retrieval methods.

jeremystretch (approved):
Thanks @ChrisPortman!

PR #14995, opened 2024-01-31
Fixes: #14839 - Check for tunnel termination type in data and instance in addition to intially passed data.

### LINKED ISSUE #14839: Unable to create tunnel termination for virtual machine
### Deployment Type

Self-hosted

### NetBox Version

v3.7.1

### Python Version

3.10

### Steps to Reproduce

1. Create tunel without any terminations at first
2. In the tunnel creation form, choose the newly selected tunnel with type virtual machine and any of its interface

### Expected Behavior

A termination for the virtual machine should be created

### Observed Behavior

Error of the form is raised

```
Select a valid choice. That choice is not one of the available choices.
```

I have debugged this a bit further, the `clean` method sets the `termination` which is always picking the Device interface no matter the selection made for the VM. Furthermore, there is an issue with the selector which always remains set for device.

So clearly, any selection made for the VM does show up in the UI but is not taken forward in any requests.

DanSheps:
### Fixes: #14839 

* Check for tunnel termination type in the form data and instance in addition to intially passed data.

### DIFF
--- a/netbox/vpn/forms/model_forms.py
+++ b/netbox/vpn/forms/model_forms.py
@@ -265,14 +265,12 @@ class TunnelTerminationForm(NetBoxModelForm):
     def __init__(self, *args, initial=None, **kwargs):
         super().__init__(*args, initial=initial, **kwargs)
 
+        if (get_field_value(self, 'type') is None and
+                self.instance.pk and isinstance(self.instance.termination.parent_object, VirtualMachine)):
+            self.fields['type'].initial = TunnelTerminationTypeChoices.TYPE_VIRTUALMACHINE
+
         # If initial or self.data is set and the type is a VIRTUALMACHINE type, swap the field querysets.
-        if (
-            initial and initial.get('type') == TunnelTerminationTypeChoices.TYPE_VIRTUALMACHINE
-        ) or (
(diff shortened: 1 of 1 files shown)


netbox/vpn/forms/model_forms.py
```diff
@@ -265,9 +265,17 @@ class Meta:
...
 
-        if initial and initial.get('type') == TunnelTerminationTypeChoices.TYPE_VIRTUALMACHINE:
+        # If initial or self.data is set and the type is a VIRTUALMACHINE type, swap the field querysets.
+        if (
+            initial and initial.get('type') == TunnelTerminationTypeChoices.TYPE_VIRTUALMACHINE
+        ) or (
+            self.data and self.data.get('type') == TunnelTerminationTypeChoices.TYPE_VIRTUALMACHINE
+        ) or (
+            self.instance.pk and isinstance(self.instance.termination.parent_object, VirtualMachine)
+        ):
```
jeremystretch, line 275: Run-on conditionals like this are not viable from a maintainability standpoint. Please have a look at how we've used the `get_field_value()` utility function in other forms to avoid this.

jeremystretch (changes requested):

PR #14998, opened 2024-01-31
Closes: #14570 - Remove extra query for job under scripts and reports detailed view

### LINKED ISSUE #14570: Remove results query from extras/views.py under ScriptView.get()
### Proposed Changes

Remove:

```
        # Look for a pending Job (use the latest one by creation timestamp)
        object_type = ContentType.objects.get(app_label='extras', model='scriptmodule')
        script.result = Job.objects.filter(
            object_type=object_type,
            object_id=module.pk,
            name=script.class_name,
        ).exclude(
            status__in=JobStatusChoices.TERMINAL_STATE_CHOICES
        ).first()
```

### Justification

This appears to be extraneous code that was left over from when results were handled inline.  This is no longer needed as the result is a separate view.

`result` is not referenced anywhere in the template.

DanSheps:
### Closes: #14570

* Remove extra query for job under scripts and reports detailed view
  * This query supports populating the .results property on the script model, however this property is no longer used as the results have been moved to their own view.

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -1059,10 +1059,6 @@ class ReportView(ContentTypePermissionRequiredMixin, View):
         report = module.reports[name]()
         jobs = module.get_jobs(report.class_name)
 
-        report.result = jobs.filter(
-            status__in=JobStatusChoices.TERMINAL_STATE_CHOICES
-        ).first()
-
         return render(request, 'extras/report.html', {
             'job_count': jobs.count(),
             'module': module,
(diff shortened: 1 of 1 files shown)


netbox/extras/views.py
```diff
@@ -1059,10 +1059,6 @@ def get(self, request, module, name):
         report = module.reports[name]()
         jobs = module.get_jobs(report.class_name)
 
-        report.result = jobs.filter(
```
jeremystretch, line 1062: `report.result` is referenced by the `report.html` template to display the time at which the report was most recently run.

jeremystretch (changes requested):

jeremystretch:
@arthanson please have a look at the proposed change and let us know if it will interfere with your work on #12510.

netbox/extras/views.py
DanSheps: Sorry, I must have missed that one, putting it back in.

arthanson (approved):
@jeremystretch okay to include this change, it's not needed even in the new code.

PR #15005, opened 2024-02-01
12795 custom user model

arthanson:
### Fixes: #12795 

Did custom user model - dumped and compared SQL schemas and they look correct.  Any extra testing suggestions welcome. 

Did not do groups for several reasons:

1. All suggestions (SO mostly) suggest just sub-classing Group: https://stackoverflow.com/questions/2181039/how-do-i-extend-the-django-group-model, https://www.reddit.com/r/django/comments/lvakzm/how_should_i_customize_the_group_model/ - Not sure if there is anything in Django that would be an issue with a custom group model (I don't think so off hand) but Django doesn't have any get_group_model like it does for User.
2. There is no BaseGroup or AbstractBaseGroup - although Group model is very simple.
3. There is an old Django ticket (https://code.djangoproject.com/ticket/29748) to allow swapping group model, although nothings been done in 2 years.  Bit worried if we do a custom group and Django does theirs it will cause issues (though more minor concern).
4. There is no pressing current need to have a custom Group, then only potential suggestion is for adding a description.
5. Adding a description would not require anything besides subclassing it as it wouldn't be a perf hit, description would'nt be needed in perf critical areas.

Wouldn't be a big thing to convert Groups bug just some potential issues and not a big reason to do it right now IMHO.  Also, an option is to potentially do the work for Django and submit a PR - though more work to handle the relation between customizable User and Group models.


netbox/users/migrations/0002_squashed_0004.py
```diff
@@ -59,20 +59,4 @@ class Migration(migrations.Migration):
                 ('objects', django.contrib.auth.models.GroupManager()),
             ],
         ),
-        migrations.CreateModel(
```
jeremystretch, line 62: Removing this model leaves an orphaned ContentType. A custom field may reference this as its object type; it also might be referenced by plugins. I'm not sure offhand what the best strategy is here. Ideally I suppose we'd merge it into the new CT for the custom User model.

netbox/users/models.py
```diff
@@ -73,6 +73,7 @@ class NetBoxGroup(Group):
     objects = NetBoxGroupManager()
 
     class Meta:
+        # db_table = 'auth_group'
```
jeremystretch, line 76:
```suggestion
```

netbox/users/models.py
```diff
@@ -23,7 +24,7 @@
 
 __all__ = (
     'NetBoxGroup',
-    'NetBoxUser',
+    'User',
```
jeremystretch, line 27: Should re-alphabetize the models list.

netbox/users/migrations/0005_alter_user_table.py
```diff
@@ -0,0 +1,31 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('users', '0002_squashed_0004'),
+    ]
+
+    operations = [
+        migrations.AlterModelTable(
+            name='user',
+            table=None,
```
jeremystretch, line 25: I assume setting `table=None` reverts to Django's default name resolution for the table (i.e. `users_user`)? If so, we should add a quick note explaining this as it's a bit confusing without context. (And if not I'm lost :laughing:)

netbox/netbox/settings.py
```diff
@@ -592,6 +594,8 @@ def _setting(name, default=None):
 SOCIAL_AUTH_JSONFIELD_ENABLED = True
 SOCIAL_AUTH_CLEAN_USERNAME_FUNCTION = 'users.utils.clean_username'
 
+SOCIAL_AUTH_USER_MODEL = 'users.User'
```
jeremystretch, line 597: The social-auth [documentation](https://python-social-auth.readthedocs.io/en/latest/configuration/settings.html#user-model) doesn't say much about this option, or when it's necessary to specify. (I assume it defaults to `contrib.auth.models.User`.) Curious if you uncovered anything in testing.

Should we link it directly to `AUTH_USER_MODEL` for simplicity's sake?

```suggestion
SOCIAL_AUTH_USER_MODEL = AUTH_USER_MODEL
```

jeremystretch (commented):
Nice work! On first pass this seems to work just fine, however I'm extremely cautious of running across an unforeseen landmine. Will continue to do some further experimentation. We also need to figure out what to do with the NetBoxUser content type.

jeremystretch:
One other nit: We'll need to ensure that the user and group content types are available in `ContentType.objects.public()`.

netbox/users/migrations/0002_squashed_0004.py
arthanson: That should be addressed in netbox/netbox/settings.py it removes and migrates the Content Type...

netbox/users/migrations/0005_alter_user_table.py
arthanson: Not sure? Was from a auto created migration when changed db_table value on the model...

arthanson:
> One other nit: We'll need to ensure that the user and group content types are available in `ContentType.objects.public()`.

@jeremystretch updated the registry code and updated the migration to migrate the CustomField from any existing ct pointers to the new one.

jeremystretch (approved):

netbox/users/apps.py
```diff
@@ -6,3 +6,11 @@ class UsersConfig(AppConfig):
...
     def ready(self):
         import users.signals
+        from .models import NetBoxGroup, ObjectPermission, Token, User, UserConfig
+        from netbox.models.features import _register_features
+
+        _register_features(NetBoxGroup)
+        _register_features(ObjectPermission)
+        _register_features(Token)
+        _register_features(User)
+        _register_features(UserConfig)
```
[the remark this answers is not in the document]
arthanson: added comment

jeremystretch (approved):

PR #15027, opened 2024-02-02
Fixes #14945: Fix "select all" button for device type components

### LINKED ISSUE #14945: Select All check box in device Types Interfaces fails to check all boxes
### Deployment Type

Self-hosted

### NetBox Version

v3.7.0

### Python Version

3.8

### Steps to Reproduce

1. Create a device Template with more than 50 interfaces
2. Go to Device Types
3. Click name of device-type
4. Click on Interfaces
5. Click on per page widget to allow 100 lines to be displayed
6. Click on Check box next to name column

### Expected Behavior

All the checkboxes should be selected.

### Observed Behavior

Only the signle box next to name is selected.

jeremystretch:
### Fixes: #14945

This updates the `component_templates.html` template to inherit from the standard template for subclasses of `generic.ObjectChildrenView`.

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -58,7 +58,11 @@ class DeviceComponentsView(generic.ObjectChildrenView):
         return self.child_model.objects.restrict(request.user, 'view').filter(device=parent)
 
 
-class DeviceTypeComponentsView(DeviceComponentsView):
+class DeviceTypeComponentsView(generic.ObjectChildrenView):
+    actions = {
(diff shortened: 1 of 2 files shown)


arthanson (changes requested):
I'm not sure if the buttons are correct here - the edit and rename buttons aren't separated and the rename is light-yellow text on white background which isn't great for readability (light background).

Monosnap APC AP7901 | NetBox 2024-02-05 09-38-55

jeremystretch:
Yep, this is correct; it's the same way action buttons for device components appear.

PR #15041, opened 2024-02-05
fix: performance for get__available_ips for prefix

### LINKED ISSUE #14500: Displaying the page of a prefix with many child prefixes (each having a child range) runs into gateway timeout
### Deployment Type

Self-hosted

### NetBox Version

v3.6.6

### Python Version

3.11

### Steps to Reproduce

The page `/ipam/prefixes/1/`  on our Netbox instance takes more than 20mins to display and runs into a timeout when not increasing timeout values.

My test system is a Debian 12 VM (VMware VM) with netbox 3.6.6. The system has two CPU cores and 4 GB Memory. Initial state is an empty database.

I can trigger the 504 Gateway Time-out (nginx error) when I create a prefix with 1000 child prefixes where each child prefix has one child range.

This is how I create the objects in nbshell:

```
mainpref = Prefix(prefix='10.128.0.0/11', status='active')
mainpref.full_clean()
mainpref.save()

mainpref = Prefix.objects.filter(prefix='10.128.0.0/11')[0]

prefsize = 256
prefcount = 1000

for i in range(0, prefcount*prefsize, prefsize):
    newpref = Prefix(prefix=mainpref.prefix[i].format() + '/24', status='active')
    newpref.full_clean()
    newpref.save()

for cpref in mainpref.get_child_prefixes():
    newrange = IPRange(start_address=cpref.prefix[100].format() + '/24' , end_address=cpref.prefix[250].format() + '/24' , status='active')
    newrange.full_clean()
    newrange.save()
```

If you apply this to a new installation, this URL triggers the timeout:

```
/ipam/prefixes/1/
```

Or click the result of this search:

```
/ipam/prefixes/?q=10.128.0.0%2F11
```

There is a related https://github.com/netbox-community/netbox/discussions/13471 where it appears that we can't do much about it but hope that this going to be addressed in Netbox itself.

### Expected Behavior

The page should be rendered within reasonable time.

### Observed Behavior

We get a 504 when trying to visit the particular page.

Smixi:
### Fixes: #14500

Use a list to add child rangers, then merge convert it into ipset. This is much faster because how netaddr.IPSet works.

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -427,10 +427,10 @@ class Prefix(GetAvailablePrefixesMixin, PrimaryModel):
 
         prefix = netaddr.IPSet(self.prefix)
         child_ips = netaddr.IPSet([ip.address.ip for ip in self.get_child_ips()])
-        child_ranges = netaddr.IPSet()
+        child_ranges = []
         for iprange in self.get_child_ranges():
-            child_ranges.add(iprange.range)
-        available_ips = prefix - child_ips - child_ranges
+            child_ranges.append(iprange.range)
+        available_ips = prefix - child_ips - netaddr.IPSet(child_ranges)
 
         # IPv6 /127's, pool, or IPv4 /31-/32 sets are fully usable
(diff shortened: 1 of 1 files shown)


arthanson (approved):

jeremystretch:
Thanks @Smixi!

PR #15047, opened 2024-02-05
15029 check if duplicate FHRP group assignment

### LINKED ISSUE #15029: Assigning duplicate FHRP group to interface results in Integrity Error
### Deployment Type

Self-hosted

### NetBox Version

v3.7.1

### Python Version

3.8

### Steps to Reproduce

View an interface
Click "Assign Group" for FHRP
Input "Group" and "Priority"
Click "Create & Add Another"
Input the same "Group" and a "Priority"
Error

```
IntegrityError at /ipam/fhrp-group-assignments/add/
duplicate key value violates unique constraint "ipam_fhrpgroupassignment_unique_interface_group"
```

### Expected Behavior

Should show "duplicate assignment" on edit form

### Observed Behavior

IntegrityError was thrown.

arthanson:
### Fixes: #15029 

Check for duplicate FHRP group assignment so IntegrityError isn't raised.

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -521,6 +521,24 @@ class FHRPGroupAssignmentForm(BootstrapMixin, forms.ModelForm):
         for ipaddress in ipaddresses:
             self.fields['group'].widget.add_query_param('related_ip', ipaddress.pk)
 
+    def clean_group(self):
+        group = self.cleaned_data['group']
+
+        conflicting_assignments = FHRPGroupAssignment.objects.filter(
+            interface_type=self.instance.interface_type,
+            interface_id=self.instance.interface_id,
+            group=group
+        )
+        if self.instance.id:
+            conflicting_assignments = conflicting_assignments.exclude(id=self.instance.id)
+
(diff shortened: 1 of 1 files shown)


jeremystretch:
Have you been able to determine why this exception isn't caught, whereas others are? For instance, attempting to create a device with the same name, site, and tenant as an existing device violates a unique constraint on the Device model, but it is caught and raised cleanly as a validation error:

screenshot

The same thing should be happening for FHRP group assignments, without needing any custom code. I suspect it might have to do with a difference in the model and/or form classes used.

arthanson:
The error happens from https://github.com/django/django/blob/main/django/forms/models.py#L390 as the fields aren't in the form so Django skips validating the unique constraints for them, dcim ones are in the form.  Several different ways to fix this:

* I looked at putting them into the form as hidden fields, but the code is also needed to set them - IMHO not obvious what the code is for doing it this way.
* Looked at overriding _get_validation_exclusions but had side-effects.

I think having it in the clean is the most obvious and direct way to fix?

jeremystretch (approved):

PR #15061, opened 2024-02-06
14438 database representation of scripts

### LINKED ISSUE #14438: Maintain database representations of reports & scripts
### NetBox version

v3.6.6

### Feature type

Data model extension

### Proposed functionality

Promote the "dummy" models in the `extras` that currently represent reports and scripts to "real" models corresponding to database tables. Each Report or Script instance will correlate to a Python class belonging to a ReportModule or ScriptModule, respectively. Report and Script objects will be created, updated, and deleted automatically in response to content changes in their assigned module.

### Use case

The lack of a concrete database object representing individual reports and scripts limits our ability to efficiently represent relationships with other NetBox objects. Two prime examples are:

* Bug #14382 - It is not currently possible to associate a specific report or script with a Webhook
* FR #14132 (v3.7) - Specifying a report or script as the action of an EventRule requires us to store the report/script class name in addition to the ForeignKey relationship to the parent module

Essentially, we reduce the current need for a tuple of (module, class) to a simple ForeignKey relationship.

### Database changes

Promoting the Report and Script classes to concrete models will effect the creation of two new PostgreSQL tables:

* `extras_report`
* `extras_script`

One new database record will be automatically created for each report and script currently installed.

### External dependencies

None

arthanson:
### Fixes: #14438 

Adds database representation of scripts.  There are several parts of this:

* Adds the actual Script model
* Update all jobs that point to the module.script to point to the new Script object and update associated code
* Update all event rules that point to the module.script to point to the new Script object and update associated code

The action_parameter on EventRule is removed as it was only used to save the name of the script as the action_object pointed to the script module.  It now points directly to the Script model.  If the script named in the action_parameter no longer exists then a dummy Script object is created with is_valid=False.

### API

The API was kept as compatible as possible, mostly adding in the new is_valid flag.  As the old API was driven off of the module, it would only return scripts that were actually in the module.  Now if a script is deleted in the code behind the covers we just mark the script as is_valid=False and it is up to the user to actually delete it, so now the API can return these 'ghost' objects.  

The API also still allows referencing by the string 'module.script_name' for backwards compatibility, where the new one just uses id for the script object.

Also, as backwards compatibility was preferred - there are some updates that could be done to the API that would be more breaking changes to keep it more in-line with other APIs, but were not done.  For example - the jobs could be returned on the Script object but wasn't as that would be breaking and it could introduce a lot of bloat to the returned data if there were a lot of previously run jobs for a script.  

**Note:* I had to make two migrations - if the remove event_rule.actions_parameters was done in the first migration it would error with "django.db.utils.OperationalError: cannot ALTER TABLE "extras_eventrule" because it has pending trigger events"


renatoalmeidaoliveira:
Maybe adding TagsMixin, ChangeLoggingMixin and JournalingMixin may give user some flexibility and better experience in the usage of Scripts models,
The TagsMixin may allow users to create granular constrais to Scripts Models allowing/restricting user from run/edit some Script
The ChangeLoggingMixin might be usefull when running the datasource sync because it may create changelogs when the script changes in the data source giving the user a better understanding of what is running
The JournalingMixin may allow users to track development or comunnicate with each other within the Script Model

arthanson:
@jeremystretch can you take a look at for review, I'm still going to do some testing on it as there are several pieces (jobs, events) and conversion from old to new, but having someone else double-check the basics would be great.  Also - let me know if you think the TagsMixin, ChangeLoggingMixin, etc.. should be added to Scripts or done as a separate FR.

netbox/extras/migrations/0108_script_models.py
```diff
@@ -0,0 +1,92 @@
...
+                    code=update_event_rules,
+                    reverse_code=migrations.RunPython.noop
+                ),
+            ],
+            state_operations=[
+                migrations.RemoveField(
+                    model_name='eventrule',
+                    name='action_parameters',
+                ),
+            ]
```
jeremystretch, line 90: Why separate state here? With the current approach, it leaves the `action_parameters` on the table; don't we want to remove it?

netbox/extras/migrations/0108_script_models.py
```diff
@@ -0,0 +1,92 @@
...
+            defaults={'is_valid': False}
+        )
+
+        rule.action_object_type = ct
+        rule.action_object_id = obj.id
+        rule.save()
+
+
+def update_scripts(apps, schema_editor):
+    from extras.models import ScriptModule
```
jeremystretch, line 29: Importing models directly can cause issues with future releases as code changes over time. Unfortunately, the safest option here is to forklift any necessary logic (e.g. `get_module_scripts()`) into the migration itself to protect it from future changes to the model.

netbox/extras/models/scripts.py
```diff
@@ -22,12 +23,50 @@
...
+    name = models.CharField(
+        verbose_name=_('name'),
+        max_length=79,
+    )
+    module = models.ForeignKey(
+        to='extras.ScriptModule',
+        on_delete=models.CASCADE,
+        related_name='scripts'
+    )
+    is_valid = models.BooleanField(
```
jeremystretch, line 36: We might want to rename this for clarity as "valid" is pretty ambiguous. I think specifically we want to convey whether the script can be executed; is that correct? Maybe something like `is_executable` or `can_run`?

jeremystretch (changes requested):

netbox/extras/api/serializers.py
```diff
@@ -510,79 +507,65 @@ class Meta:
...
+            }
+        else:
+            return {}
 
     @extend_schema_field(serializers.CharField())
     def get_display(self, obj):
         return f'{obj.name} ({obj.module})'
 
+    @extend_schema_field(serializers.CharField())
+    def get_description(self, obj):
```
jeremystretch, line 542: Any reason not to save `description` (and maybe `vars`) to the database?

netbox/extras/api/serializers.py
```diff
@@ -510,79 +507,65 @@ class Meta:
...
+            return None
+
+    @extend_schema_field(NestedJobSerializer())
+    def get_result(self, obj):
+        job = obj.jobs.all().order_by('-created').first()
+        context = {
+            'request': self.context['request']
+        }
+        data = NestedJobSerializer(job, context=context).data
+        return data
```
jeremystretch, line 555: `result` should be made accessible directly on the script instance so that the serializer can reference JobSerializer directly without needing an intermediate method. (`ScriptDetailSerializer` should also be updated.)

netbox/extras/api/views.py
```diff
@@ -9,7 +9,7 @@
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
 from rest_framework.routers import APIRootView
-from rest_framework.viewsets import ReadOnlyModelViewSet, ViewSet
+from rest_framework.viewsets import ModelViewSet, ReadOnlyModelViewSet, ViewSet
```
jeremystretch, line 12: Unused import
```suggestion
from rest_framework.viewsets import ModelViewSet, ReadOnlyModelViewSet
```

netbox/extras/api/views.py
```diff
@@ -9,7 +9,7 @@
...
 from rq import Worker
 
 from core.choices import JobStatusChoices
```
jeremystretch, line 15: Unused import
```suggestion
```

netbox/extras/api/views.py
```diff
@@ -214,53 +214,36 @@ def render(self, request, pk):
 # Scripts
 #
 
-class ScriptViewSet(ViewSet):
+class ScriptViewSet(ModelViewSet):
     permission_classes = [IsAuthenticatedOrLoginNotRequired]
+    queryset = Script.objects.all().prefetch_related('jobs')
+    serializer_class = serializers.ScriptSerializer
+    # filterset_class = filtersets.ScriptFilterSet
```
jeremystretch, line 221: Filtering should be supported

netbox/extras/api/views.py
```diff
@@ -214,53 +214,36 @@ def render(self, request, pk):
 # Scripts
 #
 
-class ScriptViewSet(ViewSet):
+class ScriptViewSet(ModelViewSet):
     permission_classes = [IsAuthenticatedOrLoginNotRequired]
+    queryset = Script.objects.all().prefetch_related('jobs')
```
jeremystretch, line 219:
```suggestion
    queryset = Script.objects.prefetch_related('jobs')
```

netbox/extras/api/views.py
```diff
@@ -214,53 +214,36 @@ def render(self, request, pk):
...
+            try:
+                module_name, script_name = pk.split('.', maxsplit=1)
+            except ValueError:
+                raise Http404
+
+            module, script = get_module_and_script(module_name, script_name)
+            if script is None:
+                raise Http404
+        else:
+            pk = int(pk)
```
jeremystretch, line 239: This can be removed; with the new URL redirects, the PK will always be an integer.

netbox/extras/forms/model_forms.py
```diff
@@ -299,18 +299,17 @@ class Meta:
     def init_script_choice(self):
```
jeremystretch, line 299: This should be updated to use an API-backed DynamicModelChoiceField like we do for webhooks.

netbox/extras/models/scripts.py
```diff
@@ -22,12 +23,54 @@
...
+    def __str__(self):
+        return self.name
+
     class Meta:
-        managed = False
+        ordering = ('module', 'name')
+        constraints = (
+            models.UniqueConstraint(
+                fields=('name', 'module'),
+                name='%(app_label)s_%(class)s_unique_name_module'
```
jeremystretch, line 54: We can hard-code this as the model is not inherited.
```suggestion
                name='extras_script_unique_name_module'
```

netbox/extras/models/scripts.py
```diff
@@ -22,12 +23,54 @@
...
+        verbose_name_plural = _('scripts')
+
+    def get_absolute_url(self):
+        return reverse('extras:script', args=[self.pk])
+
+    @cached_property
+    def python_class(self):
+        return self.module.get_module_scripts.get(self.name)
+
+    def delete_if_no_jobs(self):
```
jeremystretch, line 67: Maybe it would be preferable to override the `delete()` method?

netbox/extras/models/scripts.py
```diff
@@ -22,12 +23,54 @@
...
+        if self.jobs.all():
```
jeremystretch, line 68:
```suggestion
        if self.jobs.exists():
```

netbox/extras/models/scripts.py
```diff
@@ -78,6 +121,35 @@ def _get_name(cls):
 
         return scripts
 
+    def sync_classes(self):
```
jeremystretch, line 124: This should have a docstring

netbox/extras/models/scripts.py
```diff
@@ -78,6 +121,35 @@ def _get_name(cls):
...
+            )
+
+    def sync_data(self):
+        super().sync_data()
+        self.sync_classes()
+
     def save(self, *args, **kwargs):
         self.file_root = ManagedFileRootPathChoices.SCRIPTS
-        return super().save(*args, **kwargs)
+        super().save(*args, **kwargs)
```
jeremystretch, line 154: Be careful to still return the result of calling `save()`. We might need to invoke `sync_classes()` via the `post_save` signal.

netbox/extras/models/scripts.py
```diff
@@ -78,6 +121,35 @@ def _get_name(cls):
 
         return scripts
 
+    def sync_classes(self):
+        db_classes = {}
+        for obj in self.scripts.filter(module=self):
+            db_classes[obj.name] = obj
```
jeremystretch, line 127:
```suggestion
        db_classes = {
            script.name: script for script in self.scripts.all()
        }
```

netbox/extras/models/scripts.py
```diff
@@ -78,6 +121,35 @@ def _get_name(cls):
...
+
+        db_classes_set = {k for k in db_classes.keys()}
```
jeremystretch, line 129:
```suggestion
        db_classes_set = set(db_classes.keys())
```

netbox/extras/models/scripts.py
```diff
@@ -55,7 +98,7 @@ def __str__(self):
         return self.python_name
 
     @cached_property
-    def scripts(self):
+    def get_module_scripts(self):
```
jeremystretch, line 101: If this will be a property, its name shouldn't start with `get_`.

netbox/extras/models/scripts.py
```diff
@@ -78,6 +121,35 @@ def _get_name(cls):
...
+    def sync_classes(self):
+        db_classes = {}
+        for obj in self.scripts.filter(module=self):
+            db_classes[obj.name] = obj
+
+        db_classes_set = {k for k in db_classes.keys()}
+
+        module_scripts = self.get_module_scripts
+
+        module_classes_set = {k for k in module_scripts.keys()}
```
jeremystretch, line 133:
```suggestion
        module_classes_set = set(self.get_module_scripts.keys())
```

netbox/extras/views.py
```diff
@@ -1042,116 +1042,177 @@ def get_script_module(module, request):
     return get_object_or_404(ScriptModule.objects.restrict(request.user), file_path__regex=f"^{module}\\.")
 
 
-class ScriptView(ContentTypePermissionRequiredMixin, View):
+class BaseScriptView(ContentTypePermissionRequiredMixin, generic.ObjectView):
```
jeremystretch, line 1045: These views should be extended to support full object-based permissions now that scripts can be referenced as database objects

netbox/extras/views.py
```diff
@@ -1042,116 +1042,177 @@ def get_script_module(module, request):
...
+                return redirect('extras:script_jobs', pk=self.script.id)
+
+        self.jobs = self.script.jobs.all()
+        return None
+
+    def get_script(self, request, pk):
+        self.script = get_object_or_404(Script.objects.all(), pk=pk)
+        return self._init_vars(request)
+
+    def get_script_by_module_name(self, request, module, name):
```
jeremystretch, line 1072: No longer used?

netbox/extras/views.py
```diff
@@ -1042,116 +1042,177 @@ def get_script_module(module, request):
...
-        form = script.as_form(initial=normalize_querydict(request.GET))
+        self.script = get_object_or_404(Script.objects.all(), module=module, name=name)
+        return self._init_vars(request)
+
+
+class ScriptView(BaseScriptView):
+
+    def get(self, request, pk):
+        if ret := self.get_script(request, pk):
+            return ret
```
jeremystretch, line 1082: This is really confusing. It looks like we're using `get_script()` to return a redirect? I would expect it to return the script object.

jeremystretch (changes requested):

netbox/extras/views.py
arthanson: renamed to init_vars_or_redirect, hopefully that is more clear

PR #15063, opened 2024-02-06
Enable Event Rules to process and send data to Scripts using Event Data

### LINKED ISSUE #14884: Event Rule Action Data passed to Script
### NetBox version

v3.7.0

### Feature type

### Proposed functionality

Currently EventRules passes a serialized version of the model as data to the linked Script, I think it would be better if the Action Data was processed like the Body Template of Webhoocks and passed to the Script.

### Use case

It would allow users to build Script to react NetBox changes with the change context, currently the user cannot get directly the changelog. And by passing a parsed data with Event Data gives the plugin developer a lot of flexibility to build automations 

### Database changes

None

### External dependencies

None

renatoalmeidaoliveira:
Fixes: https://github.com/netbox-community/netbox/issues/14884

Passes Event Rule a  jinja2 processed `action_data` to script execution as its input.

### DIFF
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -119,13 +119,22 @@ def process_event_rules(event_rules, model_name, event, data, username=None, sna
             script_name = event_rule.action_parameters['script_name']
             script = script_module.scripts[script_name]()
 
+            context = {
+                'event': event,
+                'timestamp': timezone.now().isoformat(),
+                'model': model_name,
+                'username': username,
+                'request_id': request_id,
+                'model': data,
+            }
+
             # Enqueue a Job to record the script's execution
             Job.enqueue(
                 "extras.scripts.run_script",
                 instance=script_module,
                 name=script.class_name,
                 user=user,
-                data=data
+                data=event_rule.render_script_data(context)
             )
 
         else:
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -1,4 +1,5 @@
 import logging
+import json
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -14,7 +15,7 @@ from netbox.constants import RQ_QUEUE_DEFAULT
 from netbox.registry import registry
 from utilities.api import get_serializer_for_model
 from utilities.rqworker import get_rq_retry
-from utilities.utils import serialize_object
+from utilities.utils import serialize_object, render_jinja2
 from .choices import *
 from .models import EventRule, ScriptModule
 
@@ -119,14 +120,21 @@ def process_event_rules(event_rules, model_name, event, data, username=None, sna
             script_name = event_rule.action_parameters['script_name']
             script = script_module.scripts[script_name]()
 
-            context = {
-                'event': event,
-                'timestamp': timezone.now().isoformat(),
-                'model': model_name,
-                'username': username,
-                'request_id': request_id,
-                'model': data,
-            }
+            # Process Action Data
+
+            if event_rule.action_data:
+                context = {
+                    'event': event,
+                    'timestamp': timezone.now().isoformat(),
+                    'model': model_name,
+                    'username': username,
+                    'request_id': request_id,
+                    'model': data,
+                }
+                rendered_data = render_jinja2(json.dumps(event_rule.action_data), context)
+                form = script.as_form(json.loads(rendered_data))
+                if form.is_valid():
+                    data = form.cleaned_data
 
             # Enqueue a Job to record the script's execution
             Job.enqueue(
@@ -134,7 +142,7 @@ def process_event_rules(event_rules, model_name, event, data, username=None, sna
                 instance=script_module,
                 name=script.class_name,
                 user=user,
-                data=event_rule.render_script_data(context)
+                data=data
             )
 
         else:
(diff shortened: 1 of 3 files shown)


renatoalmeidaoliveira:
@jeremystretch, updated the PR, and the changes in the code reflects only the task of passing action_data to the script just like the other methods, i.e the script develop may access the input the same way he access the forms.

jeremystretch:
@renatoalmeidaoliveira did you push your changes? I don't see any updates.

renatoalmeidaoliveira:
I may have missundertud what you asked. The changes I've made was:
1 create a method to Jinja2 process the action_data with the same context passed to the webhocks
2 passes this data to the script form to process it
3 passes the validated form to the script

You mean to just pass `action_data` as created by the user to the Script as a String without any processing? I think that might not be very usefull for the user because of the lack of context of the modified object

renatoalmeidaoliveira:
I may have missundertud what you asked. The changes I've made was:
1 create a method to Jinja2 process the action_data with the same context passed to the webhocks
2 passes this data to the script form to process it
3 passes the validated form to the script

You mean to just pass `action_data` as created by the user to the Script as a String without any processing? I think that might not be very usefull for the user because of the lack of context of the modified object

renatoalmeidaoliveira:
I can put all code inside process_event_rules too,
Or keep process_action_data method in EventRules model but process the Script form inside it and remove the changes in the run_script method

renatoalmeidaoliveira:
@jeremystretch updated the code to restrict the changes only to process_event_rule inside the Script context.

aharrisson:
Is there a reason why this implementation only affect EventRuleActionChoices.SCRIPT? My use-case is to be able to use more data in a webhook payload than what I get from the default event rule "data" dict.

renatoalmeidaoliveira:
@aharrisson imo the implementation should be the same to both Webhoocks and Scripts, giving the user more flexibility to pass data to his actions.
But anyways  the Scripts it requires a different implementation to allow the user to use the `data` variable along with the input variable forms.
I can make further changes in that PR but I'm waiting some guidance from the maintainers.

aharrisson:
@renatoalmeidaoliveira , Thank you for the quick reply.
Indeed keeping the capabilities consistent (as far as possible) across both action choices would make sense. I hope the maintainers can extend the boundaries of this PR to also include webhooks, otherwise the action_data form field would be confusing for event rules triggering webhooks.

renatoalmeidaoliveira:
@llamafilm short answer, in the implementation of this PR yes.
My idea with this PR is to enable the user to build the input passed to the Script Input vars, in a similar way the webhocks body template does

llamafilm:
I would also like to see this implemented, please don’t close it.

arthanson:
@renatoalmeidaoliveira after reviewing with Jeremy, the jinja2 processing shouldn't be here, the data should just be passed as a json blob without going through jinja2.  Can you please update the PR for this?

netbox/extras/events.py
```diff
@@ -119,6 +120,22 @@ def process_event_rules(event_rules, model_name, event, data, username=None, sna
...
+            if event_rule.action_data:
+                context = {
+                    'event': event,
+                    'timestamp': timezone.now().isoformat(),
+                    'model': model_name,
+                    'username': username,
+                    'request_id': request_id,
+                    'model': data,
+                }
+                rendered_data = render_jinja2(json.dumps(event_rule.action_data), context)
```
arthanson, line 134: Sorry, I don't see why the render_jinja2 is needed here - event_rule.action_data should just be the json (dictionary) which should be able to get directly passed to the script.  I'm thinking you could just change the Job.enqueue to pass event_rule.action_data as the data param.

arthanson (changes requested):

netbox/extras/events.py
renatoalmeidaoliveira: @arthanson  If we don't render the jinja2 how the user gonna pass variables to the script, for example:
1 - For example the Script form expects an Interface ID, or some model attribute.
If It didn't run the jinja2 how the user gonna be able to pass models variables to be processed by the script form?

renatoalmeidaoliveira: The user doens't have the model params when he defines the event rule, so I think there must be some way for him to reference the models and the changelogs, to build the data that gonna be passed to the Script for

jeremystretch: The `action_data` field on EventRule holds a dictionary of arbitrary additional data. This doesn't require any mutation; it just needs to be passed to the script, and the script needs to accommodate whatever data is being passed as variables.

I think you're trying to accomplish something different than what I agreed to [here](https://github.com/netbox-community/netbox/issues/14884#issuecomment-1943820352). There's no reason to manipulate the data before sending it to the script; the script just needs to be written to accept the data as passed.

renatoalmeidaoliveira: @jeremystretch even if I pass the `event_data` as is to the script how the Script developer gonna be able to reference the event object like the changed object, since it doesn't have that data inside the Script? another thing is that way breaks the common script logic where the developer expects the data dict to return a parsed form

renatoalmeidaoliveira: Going further if the action_data is a static value what's the meaning of having it at all, the Script dev could just create the script with that data hard coded

PR #15073, opened 2024-02-07
Closes: #9583 - Add column specific search field to tables

### LINKED ISSUE #9583: Add column-specific search fields to tables
### NetBox version

v3.2.5

### Feature type

New functionality

### Proposed functionality

Add a search field to the top of all relevant columns in each table, to match on values in that column specifically. For example, the search field for the "site" column in the devices table will match only on site name/slug.

### Use case

Provides a convenient path for the user to filter results on specific fields rather than relying on the general purpose "quick search" field, which may return undesired results, and may not filter on all fields being displayed.

### Database changes

_No response_

### External dependencies

_No response_

DanSheps:
### Closes: #9583 - Add column specific search field to tables

The plan is to use the existing filterset to generate form fields on the table column headers to allow for inline-filtering using HTMX requests

#### Progress

* Fields are generated for <select> type fields
* HTMX requests are sent and account for both <select> fields
* Integrate with "quick search"
* Works with custom fields
* Allow dropdown to float outside of the parent div

#### Todo

* ~~Add input fields for "name"~~ Not needed - Quick search is suitable for this
* ~~Fix filter float (Move before field name perhaps)~~ Done
* ~~Add show/hide filter (Optional)~~ Outside of scope


netbox/netbox/views/generic/bulk_views.py
```diff
@@ -160,6 +160,11 @@ def get(self, request):
         # Render the objects table
         table = self.get_table(self.queryset, request, has_bulk_actions)
 
+        filterset_form = self.filterset_form(request.GET, label_suffix='') if self.filterset_form else None
+
+        if hasattr(self, 'filterset_form'):
+            table.filterset_form = filterset_form
+
```
DanSheps, line 167:
```suggestion
        # Check if this viiew has a filterset form and apply that form to the table
        if hasattr(self, 'filterset_form'):
            filterset_form = self.filterset_form(request.GET, label_suffix='') if self.filterset_form else None
            table.filterset_form = filterset_form

```

DanSheps, line 167:
```suggestion
        # Check for filterset_form on this view, if a form exists, apply to context and table, otherwise set to None
        filterset_form = None
        if hasattr(self, 'filterset_form'):
            filterset_form = self.filterset_form(request.GET, label_suffix='')
            table.filterset_form = filterset_form

```

netbox/templates/inc/table_controls_htmx.html
```diff
@@ -5,7 +5,7 @@
   <div class="col-auto d-print-none">
     <div class="input-group input-group-flat me-2 quicksearch">
       <input type="search" results="5" name="q" id="quicksearch" class="form-control px-2 py-1" placeholder="Quick search"
-        hx-get="{{ request.full_path }}" hx-target="#object_list" hx-trigger="keyup changed delay:500ms, search" />
+        hx-get="" hx-target="#object_list" hx-trigger="keyup changed delay:500ms, search" />
```
DanSheps, line 8: Removed full path, as this will allow taking int account filters on the form (to facilitate pushing of the filters from the column as well).

Should not impact functionality

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -1,14 +1,17 @@
 from django import template
 
+
 __all__ = (
     'getfield',
+    'getfilterfield',
     'render_custom_fields',
     'render_errors',
     'render_field',
     'render_form',
```
DanSheps, line 10:
```suggestion
    'render_filter_field',
    'render_form',
```

DanSheps (commented):
I think this is mostly there

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -46,13 +58,33 @@ def widget_type(field):
 #
 
 @register.inclusion_tag('form_helpers/render_field.html')
-def render_field(field, bulk_nullable=False, label=None):
+def render_field(field, bulk_nullable=False, label=None, table=None, request=None):
```
jeremystretch, line 61: This is too much to cram into the general-purpose `render_field` tag. If a custom template tag is necessary (and it might not be), we should create a separate one specifically for table headers.

jeremystretch (commented):
I'm seeing a lot of errors when trying to load the filter fields.

screenshot

This is likely due to having multiple form fields with the same ID (one on the table and one on the "filters" tab).

netbox/utilities/templatetags/form_helpers.py
DanSheps: I will pull it out into a separate tag.  It originally was but I didn't want to duplicate too much.

I will re-break it out as bulk_nullable and label aren't required here.

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -120,13 +131,48 @@ def render_field(field, bulk_nullable=False, label=None):
...
+
     return {
         'field': field,
         'label': label or field.label,
         'bulk_nullable': bulk_nullable or getattr(field, '_nullable', False),
     }
 
 
+@register.inclusion_tag('form_helpers/render_field.html')
+def render_table_filter_field(field, table=None, request=None):
```
jeremystretch, line 143: Why would `table` or `request` ever be None?

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -32,6 +34,15 @@ def getfield(form, fieldname):
         return None
 
 
+@register.filter()
+def getfilterfield(form, fieldname):
+    field = getfield(form, f'{fieldname}')
+    if field is not None:
+        return field
+    else:
+        return getfield(form, f'{fieldname}_id')
```
jeremystretch, line 43: We can simplify this, but I'm not sure how confident we can be in the logic. An explicit mapping of columns to form fields would be preferable.
```suggestion
    return getfield(form, f'{fieldname}') or getfield(form, f'{fieldname}_id')
```

netbox/templates/inc/table_htmx.html
```diff
@@ -16,14 +17,22 @@
                   ><i class="mdi mdi-close"></i></a>
                 </div>
               {% endif %}
+              {%  if table.filterset_form %}
```
jeremystretch, line 20:
```suggestion
              {% if table.filterset_form %}
```

netbox/templates/inc/table_htmx.html
```diff
@@ -16,14 +17,22 @@
...
               <a href="#"
                  hx-get="{{ table.htmx_url }}{% querystring table.prefixed_order_by_field=column.order_by_alias.next %}"
                  hx-target="closest .htmx-container"
                  {% if not table.embedded %}hx-push-url="true"{% endif %}
               >{{ column.header }}</a>
             </th>
           {% else %}
-            <th {{ column.attrs.th.as_html }}>{{ column.header }}</th>
+            <th {{ column.attrs.th.as_html }}>
+              {%  if table.filterset_form %}
```
jeremystretch, line 31:
```suggestion
              {% if table.filterset_form %}
```

netbox/templates/inc/table_header_filter_dropdown.html
```diff
@@ -0,0 +1,11 @@
+{% load form_helpers %}
+{% if form_field %}
+<div class="column-filter dropdown">
+    <a href="#" classs="btn dropdown-toggle" data-bs-toggle="dropdown" data-bs-auto-close="outside">
```
jeremystretch, line 4:
```suggestion
    <a href="#" class="btn dropdown-toggle" data-bs-toggle="dropdown" data-bs-auto-close="outside">

```

netbox/project-static/styles/custom/_misc.scss
```diff
@@ -28,3 +28,9 @@ span.color-label {
   visibility: hidden;
   opacity: 0;
 }
+
+// Override column filter form dropdown
```
jeremystretch, line 32: What is this overriding? It looks like a custom class; the comment should explain its purpose.

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -161,6 +161,12 @@ def get(self, request):
         # Render the objects table
         table = self.get_table(self.queryset, request, has_bulk_actions)
 
+        # Check for filterset_form on this view, if a form exists, apply to context and table, otherwise set to None
+        filterset_form = None
+        if hasattr(self, 'filterset_form') and self.filterset_form:
+            filterset_form = self.filterset_form(request.GET, label_suffix='')
+            table.filterset_form = filterset_form
```
jeremystretch, line 168: `filterset_form` is already initialized to `None` on the view instance. (Also, I don't believe `label_suffix` is used anymore.)
```suggestion
        if self.filterset_form:
            filterset_form = self.filterset_form(request.GET)
            table.filterset_form = filterset_form
```

netbox/templates/inc/table_controls_htmx.html
jeremystretch: > Should not impact functionality

Do we know this for sure? IIRC we specify the full path for a reason, I just can't recall why.

netbox/templates/inc/table_header_filter_dropdown.html
```diff
@@ -0,0 +1,11 @@
+{% load form_helpers %}
+{% if form_field %}
+<div class="column-filter dropdown">
+    <a href="#" classs="btn dropdown-toggle" data-bs-toggle="dropdown" data-bs-auto-close="outside">
+        <i class="mdi mdi-filter-settings" style="font-size: 1.25rem;"> </i>
```
jeremystretch, line 5: Need to remove the inline CSS.

netbox/templates/inc/table_htmx.html
jeremystretch, line 20: Let's refactor the conditional logic to avoid the duplicate `{% include %}`.

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -32,6 +34,15 @@ def getfield(form, fieldname):
         return None
 
 
+@register.filter()
+def getfilterfield(form, fieldname):
```
jeremystretch, line 38: We should rename this for readability. Also please add a docstring explaining its function.
```suggestion
def get_filter_field(form, fieldname):
```

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -120,13 +131,48 @@ def render_field(field, bulk_nullable=False, label=None):
...
+        # Set HTMX args
+
+    if hasattr(field.field, 'widget'):
+        field.field.widget.attrs.update({
+            'id': f'table_filter_id_{field.name}',
+            'hx-get': url if url else '#',
+            'hx-push-url': "true",
+            'hx-target': '#object_list',
+            'hx-trigger': 'hidden.bs.dropdown from:closest .dropdown'
+        })
```
jeremystretch, line 167: I'm not sure this approach is tenable: We're using the same form field for both the column and the regular filter form (behind the second tab). Setting HTMX attributes on a non-HTMX form field should be avoided, and overriding its ID may cause problems. IMO we should try to identify a cleaner approach.

jeremystretch (changes requested):

jeremystretch:
Another concern: There's no way to quickly discern when a column filter has been applied.

netbox/templates/inc/table_htmx.html
```diff
@@ -16,14 +17,22 @@
                   ><i class="mdi mdi-close"></i></a>
                 </div>
               {% endif %}
+              {%  if table.filterset_form %}
+                {% include 'inc/table_header_filter_dropdown.html' with form_field=table.filterset_form|getfilterfield:column.name %}
```
DanSheps, line 21:
```suggestion
                {% include 'inc/table_header_filter_dropdown.html' with form_field=table.filterset_form|get_filter_field:column.name %}
```

netbox/templates/inc/table_htmx.html
```diff
@@ -16,14 +17,22 @@
...
                  hx-get="{{ table.htmx_url }}{% querystring table.prefixed_order_by_field=column.order_by_alias.next %}"
                  hx-target="closest .htmx-container"
                  {% if not table.embedded %}hx-push-url="true"{% endif %}
               >{{ column.header }}</a>
             </th>
           {% else %}
-            <th {{ column.attrs.th.as_html }}>{{ column.header }}</th>
+            <th {{ column.attrs.th.as_html }}>
+              {%  if table.filterset_form %}
+                {% include 'inc/table_header_filter_dropdown.html' with form_field=table.filterset_form|getfilterfield:column.name %}
```
DanSheps, line 32:
```suggestion
                {% include 'inc/table_header_filter_dropdown.html' with form_field=table.filterset_form|get_filter_field:column.name %}
```

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -6,13 +6,15 @@
 
 __all__ = (
     'getfield',
+    'getfilterfield',
```
DanSheps, line 9:
```suggestion
    'get_filter_field',
```

DanSheps:
> Another concern: There's no way to quickly discern when a column filter has been applied.

May be able to do a swap on the icon and change it out to indicate it is filtered.

netbox/utilities/templatetags/form_helpers.py
DanSheps: I have went ahead and made a possible modification to this.

In bulk views, I initialize two copies of the form:

1. A copy of the form for the filtering tab
2. A copy of the form for the table column filtering

Since this render_table_filter_field tag is only used by the table coumn filtering logic, this shouldn't be a problem anymore

jeremystretch: AFAICT this should work fine, but do we need a full form instance for the columns? It might suffice to make a copy of the relevant fields from the filter form and attach them to their respective columns.

Though with either approach, I feel like we're missing a step. The current implementation depends on the column name and the field name matching (`form_field=table.filterset_form|get_filter_field:column.name`). This should work ~95% of the time but inevitably we're going to find exceptions. I'd like to see if we can devise a more robust approach, short of duplicating the field definitions.

arthanson:
> > Another concern: There's no way to quickly discern when a column filter has been applied.
> 
> May be able to do a swap on the icon and change it out to indicate it is filtered.

It does have the filtering chits if you refresh the page, it just doesn't show them when you select the filters - it should be changed to show the filtering chits when you select the filters.
Monosnap Devices | NetBox 2024-03-29 10-13-18

arthanson:
There is a minor UI issue:

1. Click one of the filters so it drops-down.
2. Click on another of the filters so it drops-down.

Both filters get un-selected (not dropped down).  It opens up the drop-down for the second filter, then closes the drop-down for the first filter, then closes the drop-down fro the second filter.

arthanson:
Another minor UI issue - if you select multiple items it will get clipped in the window - the multi-select on edit pages (like tags) handles this by going to multi-line which should work here as well.

Monosnap Devices | NetBox 2024-03-29 10-23-49
Monosnap Add a new device | NetBox 2024-03-29 10-23-14
Monosnap Devices | NetBox 2024-03-29 10-26-38

arthanson:
Suggestion - could we make the funnel icon smaller (use a smaller font-size for it) I think it would be just as visible and would reduce the header height to what it was before.

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -120,13 +127,48 @@ def render_field(field, bulk_nullable=False, label=None):
...
+    # Handle filter forms
+    if table:
+        # Build kwargs for querystring function
+        kwargs = {field.name: None}
+        # Build request url
+        if request and table.htmx_url:
+            url = table.htmx_url + querystring(request, **kwargs)
+        elif request:
+            url = querystring(request, **kwargs)
+        # Set HTMX args
```
arthanson, line 154: minor - what code is this comment referring to?  If the code below can it be indented with it.

netbox/utilities/templatetags/form_helpers.py
DanSheps: > AFAICT this should work fine, but do we need a full form instance for the columns? It might suffice to make a copy of the relevant fields from the filter form and attach them to their respective columns.

Instead of initializing the form right away, we could copy the form definition and then remove fields that not present on the table before intitialization.  This feels like excess code for very little gain however

> Though with either approach, I feel like we're missing a step. The current implementation depends on the column name and the field name matching (`form_field=table.filterset_form|get_filter_field:column.name`). This should work ~95% of the time but inevitably we're going to find exceptions. I'd like to see if we can devise a more robust approach, short of duplicating the field definitions.

I think we should handle this on a case-by-case.  The only other option would be looking at the table and building a completely new form class dynamically based on the table field types, however we then would likely lose all the custom filtering unless we tie it with the existing filter form somehow.

DanSheps:
> It does have the filtering chits if you refresh the page, it just doesn't show them when you select the filters - it should be changed to show the filtering chits when you select the filters.

This is outside of the htmx container, so it doesn't get refreshed, I am not sure if there is an easy way to do this.  Perhaps extracting the filtering chits in the same request, but any suggestions are appreciated.



> There is a minor UI issue:
> 
> 1. Click one of the filters so it drops-down.
> 2. Click on another of the filters so it drops-down.
> 
> Both filters get un-selected (not dropped down). It opens up the drop-down for the second filter, then closes the drop-down for the first filter, then closes the drop-down fro the second filter.

This is just the way the dropdown works.  When the dropdown closes, it reloads the htmx container, which includes all of these fields.

> Also: Suggestion (not sure if this is just the way the components work) when you drop-down a filter it shows the small selection box which you then have to click to see the list. IMHO it would be nicer if when you first drop-down the filter it would show the list as well.

I have fixed this.  Removing btn did the trick



> Another minor UI issue - if you select multiple items it will get clipped in the window - the multi-select on edit pages (like tags) handles this by going to multi-line which should work here as well.

Should be fixed via a max-width CSS attribute on the dropdown menu.



> Suggestion - could we make the funnel icon smaller (use a smaller font-size for it) I think it would be just as visible and would reduce the header height to what it was before.

Done

netbox/templates/inc/table_controls_htmx.html
DanSheps: From my testing, I haven't found any issues with it.

DanSheps:
> Another concern: There's no way to quickly discern when a column filter has been applied.

Should pop the badges now

arthanson:
Still some weirdness - If I select a filter then I do a page refresh I get doubling up of the chits (see below) I tried it on feature branch and this doesn't happen.

Monosnap Devices | NetBox 2024-04-18 09-13-33

arthanson:
If I select a filter (works correctly) then click on the same filter again (as I want to change what it is filtering on) I get a blank drop-down so I can't actually select anything else:

Monosnap Devices | NetBox 2024-04-18 09-16-13

arthanson:
If you do select a filter (I did site -> Butler Communication) then click the checkbox next to Name (for select all in list) it doesn't select anything in the list.  I also tried on feature and it works correctly.

arthanson (changes requested):
See comments - several UI issues still

DanSheps:
Will take a peek when I can.

DanSheps:
> If I select a filter (works correctly) then click on another filter again (as I want to change what it is filtering on) I get a blank drop-down so I can't actually select anything else. Sort of weird, possibly a timing thing as if I click around to other filters after a couple seconds they start rendering correctly

I think this is simply a timing thing as the HTMX re-renders the whole table, including headers.

Not sure if there is a way around this, an "easy" way would be to return the whole table but only swap the tbody on operations surrounding the column filters/search.  Let me poke around with this a bit.

DanSheps:
> If you do select a filter (I did site -> Butler Communication) then click the checkbox next to Name (for select all in list) it doesn't select anything in the list. I also tried on feature and it works correctly.

I can't recreate this.  I am able to select the full list as intended.  We might need some additional testing on this one.

DanSheps:
> Still some weirdness - If I select a filter then I do a page refresh I get doubling up of the chits (see below) I tried it on feature branch and this doesn't happen.

This one should be fixed.  See comments for other two.  Open to suggestions on those.

arthanson:
This is looking really good Dan, I just see one minor issue - if you select a filter, then click the select all (next to name) it will show the "Select all xx devices matching query" the count is all the items in the table, not the filtered count.   If you go through the filters tab and do the filter that way, then click the select all it will not show the "Select all xx..." message.

Monosnap Devices | NetBox 2024-04-23 08-42-03

DanSheps:
> This is looking really good Dan, I just see one minor issue - if you select a filter, then click the select all (next to name) it will show the "Select all xx devices matching query" the count is all the items in the table, not the filtered count. If you go through the filters tab and do the filter that way, then click the select all it will not show the "Select all xx..." message.

Just checked, this bug is present in both feature and develop (it is a long standing bug I believe, just maybe no issue for it).

I think it would be pretty easy to fix but we should track it as a separate bug IMO.

arthanson:
@DanSheps I'm not seeing the behavior with the select-all on feature (see screenshot), I'm only seeing it on your branch. I just go to filter tab, select a filter and then do the select-all button.  If I do select-all after doing the filtering by the drop-downs, that is where I see it.  What are the steps you are using when seeing it on feature?

Monosnap Sites | NetBox 2024-04-25 08-34-09

DanSheps:
Do a quick search, not a filter.  The quick search does an HTMX request on the table only.

The issue is the edit div doesn't get updated because it is outside the scope of the htmx request.

We need to do an OOB swap for that as well.

arthanson (approved):

arthanson:
@DanSheps can you open a separate bug for the select-all issue?

DanSheps:
Let me rebuild the CSS and then we can take a look at this again.

arthanson:
@DanSheps any progress on this?

DanSheps:
> @DanSheps any progress on this?

This week I should have something

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -32,6 +34,14 @@ def getfield(form, fieldname):
         return None
 
 
+@register.filter()
+def get_filter_field(form, fieldname):
+    # Check for a table form column map attribute and use that to map form fields if set
+    if hasattr(form, '_table_form_column_map') and form._table_form_column_map.get(fieldname):
+        return getfield(form, form._table_form_column_map.get(fieldname))
```
DanSheps, line 41: Not set on this, but also wondering if this should be a default on NetBoxTable

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -170,15 +171,40 @@ def get(self, request):
...
+            table.filterset_form = table.filterset_form(request.GET)
+        elif self.filterset_form and not table.filterset_form:
+            filterset_form = self.filterset_form(request.GET)
+            table.filterset_form = self.filterset_form(request.GET)
+        elif not self.filterset_form and table.filterset_form:
+            filterset_form = None
+            table.filterset_form = table.filterset_form(request.GET)
+        else:
+            filterset_form = None
+            table.filterset_form = None
```
jeremystretch, line 192: Can we ditch the association of a filterset form on the table directly, and just use the one associated with the view? That would make things a lot cleaner.

netbox/utilities/templatetags/form_helpers.py
jeremystretch: Where is `_table_form_column_map` getting set? If it's not yet needed, let's ditch it for now.

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -170,15 +171,40 @@ def get(self, request):
...
         if htmx_partial(request):
             if request.GET.get('embedded', False):
                 table.embedded = True
                 # Hide selection checkboxes
                 if 'pk' in table.base_columns:
                     table.columns.hide('pk')
+            filter_chits = render_to_string('inc/applied_filters_pane.html', {
+                'model': model,
+                'filter_form': filterset_form,
+            }, request)
```
jeremystretch, line 204: IMO all rendering logic should go within the table itself, and should be consistent for both natural and HTMX requests.

netbox/templates/inc/table_htmx.html
```diff
@@ -15,12 +16,17 @@
...
                      hx-get="{{ table.htmx_url }}{% querystring table.prefixed_order_by_field='' %}"
                      class="text-danger"
+                     {% if not table.embedded %}hx-push-url="true"{% endif %}
                      title="{% trans "Clear ordering" %}"
                   ><i class="mdi mdi-close"></i></a>
                 </div>
               {% endif %}
+              {% if table.filterset_form %}
+                {% include 'inc/table_header_filter_dropdown.html' with form_field=table.filterset_form|get_filter_field:column.name %}
+              {% endif %}
```
jeremystretch, line 26: Can we associate filter field names directly with their corresponding table columns? That would simplify things a bit.

jeremystretch (changes requested):
Sorry it's taken so long to revisit this PR. We're working now on cleaning up the PR backlog.

netbox/netbox/views/generic/bulk_views.py
DanSheps: I added this to `{% applied_filters %}` instead and adjusted all locations where this template was used and deleted the template.

netbox/templates/inc/table_htmx.html
DanSheps: Definitely

DanSheps:
Interesting note:

`yarn:bundle` on windows outputs CSS with RGB color values
`yarn:bundle` on linux outputs CSS with HEX color values

DanSheps:
@arthanson @bctiemann I just updated the bundle.   Could you two give this a look?

arthanson:
@DanSheps build is failing?

DanSheps:
Thanks @arthanson, wil let you know when it passes.

DanSheps:
I think I only update the styles, this is my bad.  Pushing the latest bundle now.

DanSheps:
@arthanson good to go it looks like

arthanson:
Just a note - not sure if this is a bootstrap type behavior but if you click a filter dropdown (leave it open) then click another filter dropdown - it closes the first one, opens the new one, then closes the new one.  If you close the first one and then select the new one it works okay, but if you don't close the first one it will auto-close the second one after a brief delay.

arthanson:
Sites -> Locations table there is a dropdown for "FACILITY" that drops down for a text entry box, but it doesn't seem to search correctly.  I added faculty "aaa" to one and "bbb" to another location then in the dropdown searched for "bbb" and it errors "No Locations were selected".

arthanson:
Contact Assignments "OBJECT TYPE" filter, because of the long name it causes UI issues - if it was a bit longer the x at the end of object would not be selectable, it doesn't scroll either.
filter1

arthanson:
Height is the same as facility above - it allows me to select a number but when I do an press return it say "No racks were selected" even if there are Racks with that height.
filter2

arthanson:
Is this by design or a bug? If I do filters by the filter drop-down then I get the filter chit but the "Filters" tab doesn't show any filter applied, and if I go to the filters tab it doesn't show any filters.  This is different behavior then if I select the filter via the filter tab. It seems a bit confusing that I have the filter chit but the filters are working two different ways. Say I use the filter dropdown to filter something then I want to filter something more complex and go to the filter tab, but now I don't have that filter. Seems like they should work inter-changeably? 
filter3

arthanson:
Device Types - "PART NUMBER" also doesn't work:
filter 5

arthanson (changes requested):
Still seems like (minor) UI issues, I didn't test every table dropdown as several issues (like certain entry fields aren't filtering correctly) seem to be a general issue.  It's a nice feature, but I'm a bit worried that it could be a support issue (minor UX nits popping up)

I haven't done a thorough code review again as these UX issues will cause code changes.

DanSheps:
> Still seems like (minor) UI issues, I didn't test every table dropdown as several issues (like certain entry fields aren't filtering correctly) seem to be a general issue. It's a nice feature, but I'm a bit worried that it could be a support issue (minor UX nits popping up)
> 
> I haven't done a thorough code review again as these UX issues will cause code changes.

I will have to check but if there isn't a corresponding filter it shouldn't allow filtering.  This seems like it is allowing filtering even if the filter itself doesn't exist.  I will double check though and see what I can find.

arthanson:
Great work, however as per discussion, closing this one for now as it needs another round of refactoring and fixing some UI issues. @DanSheps can you re open it after changes are complete.  The major areas to look at are:

* Look into an alternative solution that doesn't require column names to match the filter names per Jeremy's earlier comment "The current implementation depends on the column name and the field name matching (form_field=table.filterset_form|get_filter_field:column.name). This should work ~95% of the time but inevitably we're going to find exceptions. I'd like to see if we can devise a more robust approach, short of duplicating the field definitions."
* Investigate / fix the issue "allowing filtering even if the filter itself doesn't exist" found in the UI testing. (Sites -> Locations table dropdown for "FACILITY" )
* Filters tab getting updated when filter Is selected

I think a final UI run-through of every table / dropdown needs to be done to double-check the functionality.

PR #15080, opened 2024-02-07
Closes #14917: Replace slim-select with tom-select

### LINKED ISSUE #14917: Replace slim-select with tom-select
### Proposed Changes

We currently employ the [slim-select](https://slimselectjs.com/) library to facilitate dynamic, API-backed dropdown widgets in the user interface. We are currently using v1.27, which is far out of date. This issue proposes replacing slim-select with the alternative [tom-select](https://tom-select.js.org/) library.

If accepted, this proposal would negate #14734.

### Justification

There are a few reasons I'm leaning toward tom-select:

1. The Tabler framework ships with native styling for tom-select inputs
2. Tom-select seems to provide more thorough documentation
3. Initial experimentation with tom-select has been very promising, and it feels like it would easily support #13283
4. Upgrading slim-select to v2.8 would essentially require rewriting the current implementation (and there's apparently no migration guide provided)

jeremystretch:
### Fixes: #14917

- Drop `slim-select` as a dependency & introduce `tom-select`
- Port custom functionality from legacy APISelect to TomSelect


arthanson:
The little 'x' to remove selections is missing styling in the new one.  - screenshot shows old and new:

Monosnap Editing device PP:B117 | NetBox 2024-02-08 09-03-22
Monosnap Editing device PP:B117 | NetBox 2024-02-08 09-04-41

arthanson:
This is a bit of a strange one - found on editing an IP Address, for the NAT IP (inside) dropdown There are two issues:

1. type secondary part of an existing group of ip addresses (for example I have 172.16.0.2, 172.16.0.3..) type '.16' for the filtered list the old select showed all the ip's containing the .16.  For the new tom-select it shows 'No Results Found'.
2. What is worse is if you click off of the dropdown while it is displaying this (click to another dropdown or off to the side) it will now show '----' and if you expand it it will still show 'No Results Found', you have to press the delete key to start getting results again.

Monosnap Editing IP address 10 0 10 2:32 | NetBox 2024-02-08 09-37-36
Monosnap Editing IP address 172 16 0 1:24 | NetBox 2024-02-08 09-38-00

arthanson (changes requested):

jeremystretch:
> 1. type secondary part of an existing group of ip addresses (for example I have 172.16.0.2, 172.16.0.3..) type '.16' for the filtered list the old select showed all the ip's containing the .16. For the new tom-select it shows 'No Results Found'.

Here's what's happening in the current (v3.7) UI:

1. User opens the dropdown
2. APISelect loads and sends its initial REST API request for `/api/ipam/ip-addresses/`, populating its first 50 entries
3. User types a value, e.g. `.16`
4. A second REST API request is issued, this time for `/api/ipam/ip-addresses/?q=.16`
5. An **empty** response is received (because partial match on IPs in this manner is not currently supported)
6. APISelect falls back to returning any of the _previously retrieved_ IP addresses matching the query

It's simply coincidence if any matching IP addresses are displayed. We can confirm this by attempting to match on an IP address that is _not_ returned as part of the initial set of values. The only difference in behavior is that the TomSelect implementation does not fall back to matching against cached values.

This issue can only be resolved by implementing support for partial matching on the `address` field for IP addresses; we'll need to raise a separate bug/FR for that.

jeremystretch:
> 2. What is worse is if you click off of the dropdown while it is displaying this (click to another dropdown or off to the side) it will now show '----' and if you expand it it will still show 'No Results Found', you have to press the delete key to start getting results again.

The issue here seems to be that the query is not cleared when the field loses focus, nor is it re-displayed when the field regains focus.

PR #15104, opened 2024-02-09
Closes #13283: Add context to dropdown options

### LINKED ISSUE #13283: Ability to display additional relevant context data within the select UI
### NetBox version

v3.5.6

### Feature type

### Proposed functionality

Add the ability to display additional context relevant to each select dropdown

This display would be:

* Flexible
* Form specific (for example a interface form might show the context but perhaps the IP<>vlan assignment form it might omit the context)

Optionally, it could include:

* Ability to configure the context data (code or perhaps even user configurable to suit needs)
* Ability to show/hide as part of the actual final display of the field, and not just when searching

### Use case

When you go to select a VLAN for an interface, you might be in a situation where you may have x number of vlans even if you search by a specific VID you might get a result from global vlans, site specific vlans, global group vlans, and group scoped (site, site group, location, etc) vlans.  It currently is impossible to differentiate between all the different vlans

### Database changes

None required

### External dependencies

_No response_

jeremystretch:
### Fixes: #13283

- Introduce support for populating the following attributes within a dropdown option:
  - Value
  - Label
  - Disabled indicator
  - Description
  - Depth in hierarchy (for indentation)
  - Parent object
- Expose a new `option_attrs` keyword argument on dynamic choice form fields that can be used to override default/undefined fields


netbox/dcim/forms/model_forms.py
DanSheps, line 1: One thing I noticed (I don't think it is a problem but something we might want to keep in mind) is when something is selected already it doesn't have those option_attrs populated (ex: editing a device, the current device type is missing the parent)

DanSheps (approved):
Looks great!  Tested it out locally and apart from the one piece I identified above it works amazingly.

Should we extend this to the plugin forms as well?  I don't think the optional_attrs mechanism would work with that until we extend it.

jeremystretch:
> Should we extend this to the plugin forms as well? I don't think the optional_attrs mechanism would work with that until we extend it.

Not sure I follow; why would this work with plugins? It's the same `DynamicModelChoiceField` class.

jeremystretch:
Ah that's a good point; yes, we should expose it somehow. To be honest I'm not a fan of the name `option_attrs` though. I'll see if I can come up with something a bit more elegant.

docs/customization/custom-scripts.md
```diff
@@ -304,6 +304,7 @@ A particular object within NetBox. Each ObjectVar must specify a particular mode
 
 * `model` - The model class
 * `query_params` - A dictionary of query parameters to use when retrieving available options (optional)
+* `context` - A custom dictionary mapping template context variables to fields, used when rendering `<option>` elements within the dropdown menu (optional)
```
arthanson, line 307: Should we document here the current choices?:
Value
Label
Disabled indicator
Description
Depth in hierarchy (for indentation)
Parent object

netbox/project-static/src/select/static.ts
```diff
@@ -17,13 +17,18 @@ export function initStaticSelects(): void {
 
 // Initialize color selection fields
 export function initColorSelects(): void {
+  function renderColor(item: TomOption, escape: typeof escape_html) {
```
arthanson, line 20: This is really nice!

arthanson (approved):

docs/customization/custom-scripts.md
jeremystretch: Yeah, I was just pondering _where_ we should document these. We'll undoubtedly want to retain a reference for DynamicModelChoiceField more generally in addition to ObjectVar.

netbox/project-static/src/select/static.ts
jeremystretch: Thanks! I really like how it turned out.

PR #15112, opened 2024-02-11
Fixes #15111: Correct typo in error message

### LINKED ISSUE #15111: Typo in Device model
### Deployment Type

Self-hosted

### NetBox Version

v3.7.2

### Python Version

3.8

### Steps to Reproduce

error message is defined as "A U0 device type ...", but I think the correct definition is "A 0U device type ...".

https://github.com/netbox-community/netbox/blob/011bc5bd78229122341bf21f0774662413dd8ba0/netbox/dcim/models/devices.py#L873-L880

### Expected Behavior

A 0U device type ({device_type}) cannot be assigned to a rack position.

### Observed Behavior

A U0 device type ({device_type}) cannot be assigned to a rack position.

grncbg:
Fixes #15111: Correct typo in error message

### DIFF
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -875,7 +875,7 @@ class Device(
             if self.position and self.device_type.u_height == 0:
                 raise ValidationError({
                     'position': _(
-                        "A U0 device type ({device_type}) cannot be assigned to a rack position."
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @grncbg!

PR #15122, opened 2024-02-12
Closes #15087: Support for specifying a subset of API serializer fields

### LINKED ISSUE #15087: Allow dynamic specification of fields to return in REST API
### NetBox version

v3.7.2

### Feature type

### Proposed functionality

Allow dynamically specifying which fields to return in REST APIs, something like ?fields=a,b,c or ?exclude=e,f.  Similar to GraphQL how you can only return the fields you specifically need.  There are several projects that do this - they are at different levels of maintainership so might make sense to just roll our own:

* https://stackoverflow.com/questions/27935558/dynamically-exclude-or-include-a-field-in-django-rest-framework-serializer
* https://github.com/rsinger86/drf-flex-fields
* https://github.com/csdenboer/drf-dynamic-serializers
* https://github.com/AltSchool/dynamic-rest
* https://github.com/imbokov/drf_dynamics
* https://timmytango.com/notes/excluding-fields-in-drf-serializers/

These offer differing levels of functionality:

* specify which fields to include: `?fields=`
* specify which fields to exclude: `?exclude=` we probably don't need this
* allow specifying field inclusion on FK references via double-underscore reference: `?fields='front_port__name,...` we probably don't need this 

### Use case

This has several advantages:

* Allow getting rid of brief API
* more flexible and performant as can only return fields you specifically want
* Works better / more flexible for the API for dropdown queries where the existing brief APIs don't quite fit

### Database changes

None

### External dependencies

Potential to use external library, but we may want to just roll our own.

jeremystretch:
### Fixes: #15087

- Introduces support for specifying a subset of serializer fields to be returned in a REST API request (e.g. `GET /api/dcim/sites/?fields=id,name,status,region`)
- Introduces the `get_prefetches_for_serializer()` utility function
- Extends the base API viewset to automatically resolve the necessary queryset prefetches for a serializer based on the requested fields
- Removes all now-unnecessary explicit calls to `prefetch_related()` on API views (a few remain for various reasons)
- Removes the `brief_prefetch_fields` property from API views

### DIFF
--- a/netbox/circuits/api/views.py
+++ b/netbox/circuits/api/views.py
@@ -58,7 +58,6 @@ class CircuitTerminationViewSet(PassThroughPortMixin, NetBoxModelViewSet):
     queryset = CircuitTermination.objects.all()
     serializer_class = serializers.CircuitTerminationSerializer
     filterset_class = filtersets.CircuitTerminationFilterSet
-    brief_prefetch_fields = ['circuit']
 
 
 #
--- a/netbox/circuits/api/views.py
+++ b/netbox/circuits/api/views.py
@@ -21,7 +21,7 @@ class CircuitsRootView(APIRootView):
 #
 
 class ProviderViewSet(NetBoxModelViewSet):
-    queryset = Provider.objects.prefetch_related('asns', 'tags').annotate(
+    queryset = Provider.objects.annotate(
         circuit_count=count_related(Circuit, 'provider')
     )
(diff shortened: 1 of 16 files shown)


jeremystretch:
The test failure is due to a bug in the nested serializer for FHRP group assignments. (It should be returning `group` rather than `group_id`.)

abhi1693 (approved):

PR #15141, opened 2024-02-13
9856 Replace graphene with Strawberry

### LINKED ISSUE #9856: Replace graphene-django with Strawberry
### NetBox version

v3.2.7

### Feature type

### Proposed functionality

Replace NetBox's existing GraphQL implementation, which was built using [graphene-django](https://github.com/graphql-python/graphene-django), with [Strawberry](https://strawberry.rocks/docs/integrations/django). graphene-django unfortunately is [no longer being maintained](https://github.com/graphql-python/graphene-django/issues/1324), and is no longer compatible with Django 4.0 and later releases (something which we currently work around via a monkey patch).

### Use case

The goal here will be to keep the new GraphQL API as backward-compatible as possible, and to leverage any new features of Strawberry where they might be beneficial. Further research is needed to determine the migration strategy.

### Database changes

No changes to the database schema are anticipated

### External dependencies

Replace graphene-django with strawberry

arthanson:
### Fixes: #9856 

Replaces django-graphene with strawberry-django a new engine for handling GraphQL queries.  Notable changes:

* Converts all GraphQL types to Strawberry Types
* Includes much more updated GraphiQL browser
* Decorator to auto-generate Strawberry required data-classes for filters as Strawberry doesn't support django-filter currently.

Example Queries:
```
{
  circuit_termination_list(filters: {cabled: false}) {
    id
    display
    cable {
      id
    }
  }
}
```
and
```
{
  location_list {
    image_attachments {
      id
    }
    contacts {
      id
    }
    changelog {
      id
    }
    display
    class_type
    tags {
      id
    }
    custom_fields
    id
    created
    last_updated
    custom_field_data
    name
    slug
    description
    site {
      id
    }
    status
    tenant {
      id
    }
    lft
    rght
    tree_id
    level
    children {
      id
    }
    powerpanel_set {
      id
    }
    cabletermination_set {
      id
    }
    racks {
      id
    }
    vlan_groups {
      id
    }
    parent {
      id
    }
    devices {
      id
    }
  }
}
```
### Filtering lookups are different between the two libraries:

**Graphene**
```
user_list(is_superuser: true) {
    id
}
```
**Strawberry**
```
user_list(filters: {is_superuser: true}) {
   id
}
```
The filtering lookup syntax (greater than, less than, case-insensitive compare) is also different:
**Graphene**
```
{
  user_list(username__ic: "bob") {
    username
  }
}
```
**Strawberry**
```
{
  user_list(filters: {username: {i_contains: "bob"}}) {
    username
  }
}
```

For stawberry, allowing lookups on fields forces the lookup to be present, so it forces this somewhat awkward syntax:
**Graphene**
```
{
  user_list(id: "3") {
    id
  }
}
```
**Strawberry**
```
{
  user_list(filters: {id: {exact: 3}}) {
    id
  }
}
```
## Performance optimizer

Below is a ridiculous nested query that can be used to show optimization

**old graphene:** 55 queries
**strawberry:** 24 queries

```
{
  location_list {
    image_attachments {
      id
    }
    contacts {
      id
    }
    changelog {
      id
    }
    display
    class_type
    tags {
      id
    }
    custom_fields
    id
    created
    last_updated
    custom_field_data
    name
    slug
    description
    site {
      id
      name
      locations {
        id
        name
      }
    }
    status
    tenant {
      id
      sites {
        id
        name
      }
    }
    lft
    rght
    tree_id
    level
    children {
      id
    }
    powerpanel_set {
      id
    }
    cabletermination_set {
      id
    }
    racks {
      id
      site {
        id
        name
      }
    }
    vlan_groups {
      id
    }
    parent {
      id
    }
    devices {
      id
    }
  }
}
```


arthanson:
Remaining work:

- [ ] Validate Circuits Types
- [ ] Validate Circuits Fitlersets
- [ ] Auto generate Filtersets definitions
- [ ] Choice field handling
- [ ] UsersQuery
- [ ] CoreQuery
- [ ] DCIMQuery
- [ ] ExtrasQuery
- [ ] IPAMQuery
- [ ] TenancyQuery
- [ ] VirtualizationQuery
- [ ] VPNQuery
- [ ] WirelessQuery
- [ ] Fix GraphQL tests auto-discovery of fields
- [ ] Update netbox/graphql/views.py GraphQLView
- [ ] netbox/netbox/graphql/fields.py (ObjectField, ObjectListfield)? - I think this is obsolete from optimizer
- [ ] netbox/netbox/graphql/scalars.py (BigInt)?
- [ ] netbox/netbox/graphql/utils.py (get_graphene_type)?
- [ ] netbox/dcim/graphql/gfk_mixins.py (gfk mixins)?

arthanson:
Including some GraphQL queries here I used for testing as they are handy for reference - just got some GFK stuff working! :tada: :

```
  {
    circuit_termination_list {
      id
      link_peers {
        __typename
        ... on InterfaceType {
          id
          name
        }
        ... on RearPortType {
          id
          display
          device {
            pk
          }
        }
      }
    }
  }
```

netbox/utilities/testing/api.py
```diff
@@ -447,34 +450,34 @@ def _build_query(self, name, **filters):
...
-                elif inspect.isclass(field.type) and issubclass(field.type, GQLUnion):
-                    # Union types dont' have an id or consistent values
+
+            for field in type_class.__strawberry_definition__.fields:
+                if (
+                    field.type in (
+                        strawberry_django.fields.types.DjangoFileType, strawberry_django.fields.types.DjangoImageType) or
+                    type(field.type) is StrawberryOptional and field.type.of_type in (
+                        strawberry_django.fields.types.DjangoFileType, strawberry_django.fields.types.DjangoImageType)
+                ):
```
jeremystretch, line 460: We can clean this up by defining a list of fields to check for.
```suggestion
                file_fields = (strawberry_django.fields.types.DjangoFileType, strawberry_django.fields.types.DjangoImageType)
                if (
                    field.type in file_fields or (
                        type(field.type) is StrawberryOptional and field.type.of_type in file_fields
                    )
                ):
```

netbox/core/graphql/types.py
```diff
@@ -1,21 +1,34 @@
...
 
 __all__ = (
     'DataFileType',
     'DataSourceType',
 )
 
 
+@strawberry_django.type(
+    models.DataFile,
+    exclude=('data',),
```
jeremystretch, line 18: `strawberry_django` seems to demand a list here.
```suggestion
    exclude=['data'],
```

netbox/circuits/graphql/filters.py
```diff
@@ -0,0 +1,50 @@
+import strawberry
+import strawberry_django
+from circuits import filtersets, models
+
+from netbox.graphql.filter_mixins import autotype_decorator, BaseFilterMixin
+
+__all__ = (
```
jeremystretch, line 7: Should we look into automating the generation of these filters? Or would you prefer to leave it explicit for now to avoid potentially complicating things?

netbox/dcim/graphql/types.py
```diff
@@ -67,447 +79,944 @@ class ComponentObjectType(
...
-        filterset_class = filtersets.CableFilterSet
 
-    def resolve_type(self, info):
-        return self.type or None
-
-    def resolve_length_unit(self, info):
-        return self.length_unit or None
+@strawberry_django.type(
+    models.CableTermination,
+    exclude=('termination_type', 'termination_id'),
```
jeremystretch, line 118: `exclude` needs to be a list to make type checking happy.
```suggestion
    exclude=['termination_type', 'termination_id'],
```

netbox/dcim/graphql/types.py
```diff
@@ -67,447 +79,944 @@ class ComponentObjectType(
...
+    @strawberry_django.field
+    def inventory_item_templates(self) -> List[Annotated["InventoryItemTemplateType", strawberry.lazy('dcim.graphql.types')]]:
+        return self.inventory_item_templates.all()
 
-    def resolve_rf_channel(self, info):
-        return self.rf_channel or None
 
+@strawberry_django.type(
+    models.Location,
+    # fields='__all__',
```
jeremystretch, line 569:
```suggestion
```

netbox/dcim/graphql/types.py
```diff
@@ -67,447 +79,944 @@ class ComponentObjectType(
...
-        filterset_class = filtersets.RearPortTemplateFilterSet
+    @strawberry_django.field
+    def frontport_templates(self) -> List[Annotated["FrontPortTemplateType", strawberry.lazy('dcim.graphql.types')]]:
+        return self.frontport_templates.all()
 
 
+@strawberry_django.type(
+    models.Region,
+    exclude=('parent',),
+    # fields='__all__',
```
jeremystretch, line 899:
```suggestion
```

netbox/dcim/graphql/types.py
```diff
@@ -67,447 +79,944 @@ class ComponentObjectType(
...
+        return self.clusters.all()
+
+    @strawberry_django.field
+    def vlans(self) -> List[Annotated["VLANType", strawberry.lazy('ipam.graphql.types')]]:
+        return self.vlans.all()
+
+
+@strawberry_django.type(
+    models.SiteGroup,
+    # fields='__all__',
```
jeremystretch, line 976:
```suggestion
```

netbox/netbox/graphql/filter_mixins.py
```diff
@@ -0,0 +1,194 @@
...
+
+    For other filter fields a function needs to be created for Strawberry with the
+    naming convention `filter_{fieldname}` which is auto detected and called by
+    Strawberry, this function uses the filterset to handle the query.
+    """
+    def create_attribute_and_function(cls, fieldname, attr_type, should_create_function):
+        if fieldname not in cls.__annotations__ and attr_type:
+            cls.__annotations__[fieldname] = attr_type
+
+        fname = f"filter_{fieldname}"
```
jeremystretch, line 39: Rename `fname` to `filter_name` for clarity?

netbox/netbox/graphql/filter_mixins.py
```diff
@@ -0,0 +1,194 @@
...
+
+                create_attribute_and_function(cls, fieldname, attr_type, should_create_function)
+
+        declared_filters = filterset.declared_filters
+        for fieldname, field in declared_filters.items():
+            should_create_function = False
+            attr_type = None
+
+            # NetBox Filter types - put base classes after derived classes
+            if isinstance(field, ContentTypeFilter):
```
jeremystretch, line 71: Might be cleaner to move this swath of logic into a separate function.

netbox/tenancy/graphql/types.py
```diff
@@ -14,64 +18,177 @@
 )
 
 
+@strawberry.type
 class ContactAssignmentsMixin:
```
jeremystretch, line 22: Should we move this to `tenancy/mixins.py` for consistency with other apps?

netbox/users/graphql/types.py
```diff
@@ -1,38 +1,45 @@
...
+    Group,
+    fields=['id', 'name'],
+    filters=GroupFilter
+)
+class GroupType:
     @classmethod
-    def get_queryset(cls, queryset, info):
-        return RestrictedQuerySet(model=Group).restrict(info.context.user, 'view')
+    def get_queryset(cls, queryset, info, **kwargs):
+        return RestrictedQuerySet(model=Group).restrict(info.context.request.user, 'view')
```
jeremystretch, line 27: Is a custom `get_queryset()` method still needed for the User & Group models now that we have our own custom models in place?

requirements.txt
```diff
@@ -31,6 +30,7 @@ PyYAML==6.0.1
 requests==2.31.0
 social-auth-app-django==5.4.0
 social-auth-core[openidconnect]==4.5.3
+strawberry-graphql-django==0.33.0
```
jeremystretch, line 33: We should add `strawberry` itself too as a direct dependency.

jeremystretch (changes requested):
Excellent work! Noticed a few minor cleanup items on initial review and had some general questions.

netbox/circuits/graphql/filters.py
arthanson: I think leave it explicit for now as you may want to possibly add something to the filter beyond the auto-generated response.

netbox/circuits/graphql/types.py
```diff
@@ -15,48 +20,93 @@
...
+)
+class ProviderType(NetBoxObjectType, ContactsMixin):
 
-    class Meta:
-        model = models.CircuitTermination
-        fields = '__all__'
-        filterset_class = filtersets.CircuitTerminationFilterSet
+    @strawberry_django.field
+    def networks(self) -> List[Annotated["ProviderNetworkType", strawberry.lazy('circuits.graphql.types')]]:
+        return self.networks.all()
```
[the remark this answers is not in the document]
arthanson: Its going to ProviderNetworkType, this is ProviderType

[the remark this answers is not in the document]
jeremystretch: What I mean is `ProviderType.networks` seems to return itself, for instance, because of the decorator. It's not a problem, just a bit confusing at first glance.

base_requirements.txt
```diff
@@ -136,8 +131,16 @@ social-auth-core
 # https://github.com/python-social-auth/social-app-django/blob/master/CHANGELOG.md
 social-auth-app-django
 
+# Strawberry GraphQL
+# https://github.com/strawberry-graphql/strawberry/blob/main/CHANGELOG.md
+strawberry-graphql
+
+# Strawberry GraphQL Django extension
+# https://github.com/strawberry-graphql/strawberry-django/blob/main/CHANGELOG.md
+strawberry-django
```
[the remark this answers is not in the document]
arthanson: oops, yes

jeremystretch (changes requested):
The GraphiQL view isn't using our custom template, and static resources are being pulled from unpkg.com rather than being served locally.

Also, it looks like maybe we should swap `django-graphiql-debug-toolbar` for `django-graphiql-strawberry-debug-toolbar`, although AFAICT the current middleware seems functional.

arthanson:
Fixed up the GraphiQL view, the old method of bundling the assets wouldn't work as the new GraphiQL has some include packages that ESBuild can't translate to ES6, see: (https://github.com/graphql/graphiql/discussions/2525) - we could switch to webpack which does handle it, but as it is just for GraphiQL it seemed fine to just statically link the separate files.  Changed the ESBuild script to copy the files out of the node-packages.

django-graphical-debug-toolbar is no longer needed, so removed that.  The strawberry one is handled by middleware already.

netbox/netbox/settings.py
```diff
@@ -474,6 +473,10 @@ def _setting(name, default=None):
     os.path.join(BASE_DIR, 'project-static', 'img'),
     os.path.join(BASE_DIR, 'project-static', 'js'),
     ('docs', os.path.join(BASE_DIR, 'project-static', 'docs')),  # Prefix with /docs
+    # os.path.join(ROOT_DIR, 'node_modules', 'graphiql-explorer'),
+    # os.path.join(ROOT_DIR, 'node_modules', 'react', 'cjs'),
+    # os.path.join(ROOT_DIR, 'node_modules', 'react_dom', 'cjs'),
+    # os.path.join(ROOT_DIR, 'node_modules', 'js-cookie', 'dist'),
```
jeremystretch, line 479: Safe to remove these?
```suggestion
    # os.path.join(ROOT_DIR, 'node_modules', 'graphiql-explorer'),
    # os.path.join(ROOT_DIR, 'node_modules', 'react', 'cjs'),
    # os.path.join(ROOT_DIR, 'node_modules', 'react_dom', 'cjs'),
    # os.path.join(ROOT_DIR, 'node_modules', 'js-cookie', 'dist'),
```

netbox/project-static/bundle.js
```diff
@@ -14,24 +17,49 @@ const options = {
...
-      define: {
-        global: 'window',
-      },
-    });
-    if (result.errors.length === 0) {
-      console.log(`✅ Bundled source file 'netbox-graphiql/index.ts' to 'graphiql.js'`);
+  fileMap = [
+    {
+      source: './node_modules/react/umd/react.production.min.js',
+      dest: './dist/react.production.min.js'
```
jeremystretch, line 30: Can we move these to a subdirectory ( e.g. `static/graphiql/`) to keep them isolated from "native" static resources?

netbox/project-static/bundle.js
```diff
@@ -14,24 +17,49 @@ const options = {
...
+      dest: './dist/graphiql.min.css'
+    },
+    {
+      source: './node_modules/@graphiql/plugin-explorer/dist/style.css',
+      dest: './dist/plugin-explorer-style.css'
     }
-  } catch (err) {
-    console.error(err);
-  }
+  ]
```
jeremystretch, line 56:
```suggestion
  ];
```

netbox/project-static/bundle.js
```diff
@@ -14,24 +17,49 @@ const options = {
...
-        graphiql: 'netbox-graphiql/index.ts',
-      },
-      target: 'es2016',
-      define: {
-        global: 'window',
-      },
-    });
-    if (result.errors.length === 0) {
-      console.log(`✅ Bundled source file 'netbox-graphiql/index.ts' to 'graphiql.js'`);
+  fileMap = [
```
jeremystretch, line 27:
```suggestion
  let fileMap = [
```

jeremystretch (changes requested):
Almost there!

PR #15155, opened 2024-02-14
15094 Add missing gettext to error strings for internationalization

### LINKED ISSUE #15094: Internationalize raised / thrown Error messages
### Proposed Changes

There are a handful of ValidationError that are not internationalized, they can be found searching for:

Error("

In the project.  The main ones are ValidationError

### Justification

These will show to the user so should be internationalized.

arthanson:
### Fixes: #15094 

Was obviously not able to test each individual error message so another set of eyes to double check the formatting would be appreciated.

### DIFF
--- a/netbox/utilities/views.py
+++ b/netbox/utilities/views.py
@@ -36,7 +36,8 @@ class ContentTypePermissionRequiredMixin(AccessMixin):
         Return the specific permission necessary to perform the requested action on an object.
         """
         raise NotImplementedError(_("{self.__class__.__name__} must implement get_required_permission()").format(
-            class_name=self.__class__.__name__))
+            class_name=self.__class__.__name__
+        ))
 
     def has_permission(self):
         user = self.request.user
@@ -71,7 +72,8 @@ class ObjectPermissionRequiredMixin(AccessMixin):
         Return the specific permission necessary to perform the requested action on an object.
         """
         raise NotImplementedError(_("{class_name} must implement get_required_permission()").format(
-            class_name=self.__class__.__name__))
+            class_name=self.__class__.__name__
+        ))
 
     def has_permission(self):
         user = self.request.user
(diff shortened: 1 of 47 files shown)


netbox/utilities/views.py
```diff
@@ -89,8 +92,8 @@ def dispatch(self, request, *args, **kwargs):
 
         if not hasattr(self, 'queryset'):
             raise ImproperlyConfigured(
-                '{} has no queryset defined. ObjectPermissionRequiredMixin may only be used on views which define '
-                'a base queryset'.format(self.__class__.__name__)
+                _('{} has no queryset defined. ObjectPermissionRequiredMixin may only be used on views which define a base queryset').format(
+                    self.__class__.__name__)
```
jeremystretch, line 96:
```suggestion
                _(
                    '{class} has no queryset defined. ObjectPermissionRequiredMixin may only be used on views which '
                    'define a base queryset'
                ).format(class=self.__class__.__name__)
```

netbox/wireless/utils.py
```diff
@@ -22,6 +23,6 @@ def get_channel_attr(channel, attr):
         'width': Decimal(channel_values[3]),
     }
     if attr not in attrs:
-        raise ValueError(f"Invalid channel attribute: {attr}")
+        raise ValueError(_("Invalid channel attribute: {attr}").format(attr=attr))
```
jeremystretch, line 26:
```suggestion
        raise ValueError(_("Invalid channel attribute: {name}").format(name=attr))
```

jeremystretch (changes requested):
Thanks for combing through all of these. I added a lot of changes, but there are really just two points of feedback:

1. Try to keep in mind the perspective of the translator when naming string variables. Often we're guilty of using a shorthand like `{f}` or `{attr}` because it makes sense in the code, but won't when read out of context by someone attempting to translate the string.
2. Be sure to avoid hanging indents (opening and closing parentheses should always be at the same depth).

jeremystretch (approved):

PR #15158, opened 2024-02-14
Closes #14736: Enable HTMX navigation globally

### LINKED ISSUE #14736: Use HTMX for general UI navigation
### NetBox version

v3.7.0

### Feature type

New functionality

### Proposed functionality

NetBox has employed [HTMX](https://htmx.org/) in several areas of the UI for a while now, to improve responsiveness. These areas include:

* Sorting/filtering lists of objects
* Re-rendering object edit forms
* Providing deletion confirmation dialogs

This proposal seeks to extend the use of HTMX across the entire application, leveraging it for nearly all views. For instance, clicking a navigation menu link would replace only the main portion of the page with new content, avoiding a full page refresh.

### Use case

HTMX strikes a very natural balance between traditional full-page rendering and reactive SPA frameworks, returning only the portion of HTML content necessary to update the page. This provides a much more seamless user experience, and can significantly reduce page load times. These gains can be achieved with relatively little backend work, leveraging existing HTML templates.

### Database changes

None

### External dependencies

We'll use the [`django-htmx` package](https://github.com/adamchainz/django-htmx) per #14735.

jeremystretch:
### Fixes: #14736

- Introduce the `ui.htmx_navigation` user preference for toggling HTMX navigation per user
- Introduce the `render_partial()` utility function for determining when to return a partial HTML response
- Add `htmx_navigation` (boolean) to the default template context
- Add a `page-content` element in the base layout template
- Introduce the `{% htmx_boost %}` template tag to enable [boosting](https://htmx.org/attributes/hx-boost/) certain areas of the page
- Introduce the `{% formaction %}` template tag to enable HTMX support for form controls with a `formaction` attribute
- Merge `inc/user_menu.html` into `base/layout.html`


netbox/core/views.py
```diff
@@ -320,7 +321,7 @@ def get(self, request, queue_index, status):
         table = self.get_table(data, request, False)
 
         # If this is an HTMX request, return only the rendered table HTML
-        if request.htmx:
+        if render_partial(request):
```
arthanson, line 324: IMHO this function should be renamed, the name makes it sound like it is rendering something not just for a true/false test.  maybe is_htmx_request?  I realize this is already in the codebase, but wanted to point it out.

netbox/project-static/src/htmx.ts
```diff
@@ -1,11 +1,12 @@
...
 import { initButtons } from './buttons';
+import { initClipboard } from './clipboard'
 import { initSelects } from './select';
 import { initObjectSelector } from './objectSelector';
 import { initBootstrap } from './bs';
+import { initMessages } from './messages';
 
 function initDepedencies(): void {
-  for (const init of [initButtons, initSelects, initObjectSelector, initBootstrap]) {
+  for (const init of [initButtons, initClipboard, initSelects, initObjectSelector, initBootstrap, initMessages]) {
```
arthanson, line 9: is initClipboard and initMessages required?  Wasn't obvious they were part of this PR.

netbox/utilities/templatetags/navigation.py
```diff
@@ -40,6 +40,22 @@ def nav(context: Context) -> Dict:
...
     return {
-        "nav_items": nav_items,
-        "request": context["request"]
+        'nav_items': nav_items,
+        'htmx_navigation': context['htmx_navigation']
     }
+
+
+@register.simple_tag(takes_context=True)
+def htmx_boost(context, target='#page-content', select='#page-content'):
```
arthanson, line 49: Add a comment what this does, not obvious.

arthanson (changes requested):

netbox/core/views.py
DanSheps: maybe:

`should_render_partial()` ?

jeremystretch: Yeah, that's a good call-out. The logic is bit more nuanced than "is it ~~cake~~ HTMX?" because we're considering the target element as well, but we also want to keep the name concise.

jeremystretch: I went with `htmx_partial()`.

arthanson (approved):

PR #15199, opened 2024-02-20
Added index on cachevalue

### LINKED ISSUE #14966: Add index on extras_cachedvalue.object_id for better update performance with many interfaces
### NetBox version

v3.7.0

### Feature type

Data model extension

### Proposed functionality

Add an index to the object_id column on the extras_cachedvalue table.

This was suggested before in #11775 but for inserting new interfaces, and that was solved in #11785 by skipping the clearing of the cache on inserts, but that logic does of course not apply to updating interfaces.
Then there was an implementation of indexes for all generic foreign keys in #11790, but that did not include this object_id as it is something that can refer to multiple tables and is therefore not a formal foreign key.

### Use case

When there are many interfaces (we are currently working with 300k+, times 7 attributes gives 2.2 million rows in the extras_cachedvalue table), updating (many of) them takes a lot of time. Looking at the PostgreSQL stats, this is the query that takes a lot of time:
```
DELETE FROM "extras_cachedvalue" WHERE ("extras_cachedvalue"."object_id" = 268011 AND "extras_cachedvalue"."object_type_id" = 38)
```
where in this instance 38 is the ID of dcim.interface, and that object_id is just an example.

Adding the suggested index made a big improvement in the time it takes to process the updates.

### Database changes

CREATE INDEX extras_cachedvalue_object_id_TBD ON extras_cachedvalue USING btree(object_id);

### External dependencies

_No response_

abhi1693:


netbox/extras/models/search.py
```diff
@@ -57,6 +57,9 @@ class Meta:
         ordering = ('weight', 'object_type', 'value', 'object_id')
         verbose_name = _('cached value')
         verbose_name_plural = _('cached values')
+        indexes = (
+            models.Index(fields=('object_type', 'object_id')),
```
jeremystretch, line 61: We can specify a name for the index to avoid Django attaching random cruft to it.
```suggestion
            models.Index(fields=('object_type', 'object_id'), name='extras_cachedvalue_object'),
```

netbox/extras/migrations/0107_cachedvalue_extras_cach_object__d92213_idx.py
jeremystretch, line 1: We should give the migration a more meaningful name; something like `0107_cachedvalue_add_object_index`.

jeremystretch (changes requested):

jeremystretch (approved):

PR #15201, opened 2024-02-20
14405 render link_peer to CSV

### LINKED ISSUE #14405: Multiple "All Data (CSV)" Exports contain extra lines
### NetBox version

v3.6.6

### Python version

3.8

### Steps to Reproduce

1. Create at least two devices, one containing a rear port, one containing an interface
2. Connect the rear port to an interface
3. Navigate to dcim/rear-ports/ ; use filters if you want to
4. Export via "Export" -> "All Data (CSV)"
5. Notice that the file contains extra lines where the ">" was probably not handled correctly: 
image
(Optional: 6. Be annoyed enough by the bug to write a report)

### Expected Behavior

No extra lines

### Observed Behavior

Lines added

arthanson:
### Fixes: #14405 

Provides a value_link_peer to render link_peer fields to CSV files similar to what the template_code does for HTML rendering.  Instead of appearing as extra lines in the CSV (which is still valid) puts a ">" between the parent and child instead of a new line.

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -359,6 +359,17 @@ class CableTerminationTable(NetBoxTable):
         verbose_name=_('Mark Connected'),
     )
 
+    def value_link_peer(self, value):
+        string = ""
+        for termination in value:
+            if value and not str:
+                string += " "
+            if termination.parent_object:
+                string += f"{termination.parent_object} > "
+            string += str(termination)
+
+        return string
+
 
 class PathEndpointTable(CableTerminationTable):
     connection = columns.TemplateColumn(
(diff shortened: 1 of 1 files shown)


netbox/dcim/tables/devices.py
```diff
@@ -359,6 +359,17 @@ class CableTerminationTable(NetBoxTable):
...
+    def value_link_peer(self, value):
+        string = ""
+        for termination in value:
+            if value and not str:
+                string += " "
+            if termination.parent_object:
+                string += f"{termination.parent_object} > "
+            string += str(termination)
+
+        return string
```
jeremystretch, line 371: We can condense this (it's safe to assume every termination object has a `parent_object` attribute).
```suggestion
        return ', '.join([
            f"{termination.parent_object} > {termination}" for termination in value
        ])
```

jeremystretch (changes requested):

PR #15203, opened 2024-02-20
Closes #15042: Move model registration logic to AppConfigs

### LINKED ISSUE #15042: Rearchitect the logic for registering models & model features
### Proposed Changes

Currently, NetBox automatically registers all models and their features by hooking into Django's [`class_prepared`](https://docs.djangoproject.com/en/5.0/ref/signals/#class-prepared) signal. This issue proposes moving the registration triggers into the `ready()` method of the AppConfig class for each app.

### Justification

Although the current approach has generally worked well, issues arose while introducing a custom user model under #12795. Because Django must load the (now custom) user model before the signal handler, we must reorder the logic. This will also enable cleaner separation of functions.

jeremystretch:
### Fixes: #15042

- Move code from `extras.utils.register_features()` into `netbox.models.features._register_features()`
- Rename `netbox.models.features._register_features()` to `register_model()`
- Decouple model registration from the `class_prepared` signal
- Call `register_model()` for each model under `AppConfig.ready()` for each app
- Extend `PluginConfig.ready()` to also call `register_model()` for each model
- Add a plugin test for model registration

### DIFF
--- a/netbox/circuits/apps.py
+++ b/netbox/circuits/apps.py
@@ -6,4 +6,9 @@ class CircuitsConfig(AppConfig):
     verbose_name = "Circuits"
 
     def ready(self):
+        from netbox.models.features import register_model
         from . import signals, search
+
+        # Register models
+        for model in self.get_models():
+            register_model(model)
--- a/netbox/circuits/apps.py
+++ b/netbox/circuits/apps.py
@@ -6,9 +6,8 @@ class CircuitsConfig(AppConfig):
     verbose_name = "Circuits"
 
     def ready(self):
-        from netbox.models.features import register_model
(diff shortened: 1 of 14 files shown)


netbox/circuits/apps.py
```diff
@@ -6,4 +6,9 @@ class CircuitsConfig(AppConfig):
     verbose_name = "Circuits"
 
     def ready(self):
+        from netbox.models.features import register_model
```
arthanson, line 9: For DRY can we create a base NetBoxAppClass and have a base ready that registers models?  Or in lieu of that have a register_all_models(instance) that does the loop / register_model.

arthanson (changes requested):

PR #15215, opened 2024-02-21
Added a setting to mark read-only as not required

### LINKED ISSUE #14953: Fields conveying annotated values are missing from REST API serialization of newly-created objects
### Deployment Type

Self-hosted

### NetBox Version

v3.7.1

### Python Version

3.11

### Steps to Reproduce

1. Run the server
2. Generate the openapi spec by hitting api/schema endpoint
(observe "tenant_count" field listed as required for the TenantGroup spec)
3. Run a creation query for TenantGroup (observe the response does not contain tenant_count field)

This can also be reproduced for many of the other readonly fields in the schema.

### Expected Behavior

Expected the generated spec and behavior to match, in one of two ways:
- Creation query response should contain tenant_count field
- "tenant_count" should not appear in the required fields for TenantGroup in the openapi spec
(Similarly for many other readonly fields in the schema)

I suspect it is easier to change the spec than the behavior. It can be done by adding the drf-spectacular setting COMPONENT_NO_READ_ONLY_REQUIRED in the default settings.py, and marking all readonly fields that **are** required explicitly as required=True.

Relevant line of code specifying serializer properties: https://github.com/netbox-community/netbox/blob/develop/netbox/tenancy/api/serializers.py#L22

Note that setting required='False' in the line^ above does NOT change the generated spec because drf-spectacular defaults to applying a heuristic that readonly fields are required, unless COMPONENT_NO_READ_ONLY_REQUIRED is set. That is why we think the solution is to flip this default and label readonly fields as required _except_ for the ones that will sometimes not be returned.
(More details here: https://github.com/tfranzel/drf-spectacular/issues/1155).

Furthermore I think the fields that are sometimes not returned are the ones added to responses via annotations, such as the one here: https://github.com/netbox-community/netbox/blob/b9cac97b73c1c08213db6272de11c03711491486/netbox/tenancy/api/views.py#L41

For full context, the use case here is deriving generated code from the generated openapi spec, which would be easier if the spec matched with server behavior. Currently it requires manually editing the spec to adjust the aforementioned fields after observing errors.

### Observed Behavior

See "Steps to Reproduce"

abhi1693:


jeremystretch:
I'm unclear on why this change is needed; see my comment [here](https://github.com/netbox-community/netbox/issues/14953#issuecomment-1957683614).

PR #15251, opened 2024-02-23
Fixes: #14634 - Bypass Write permissions for render endpoint

### LINKED ISSUE #14634: API access to render-config needs add permissions and write enable for the api token
### Deployment Type

Self-hosted

### NetBox Version

v3.6.8

### Python Version

3.10

### Steps to Reproduce

1. Create a User with View Permissions to all Object Types
2. Create a API key for this User with out `Wrire Enabled`
3. Create a Device (No configured render config is needed to trigger this bug)
4. Try to use the [dcim_devices_render_config_create](https://demo.netbox.dev/api/schema/swagger-ui/#/dcim/dcim_devices_render_config_create) API Endpoint with the created API key / user

### Expected Behavior

The User with Read Access to the Device should be able to get the render config from the API. As there is no config set a `No config template found for this device` error is expected.
```
user@host:~$ curl -H 'Authorization: Token <YOURTOKEN>' https://netbox/api/dcim/devices/<ID>/render-config/
{"error":"No config template found for this device."}
```

### Observed Behavior

The User with Read Access to the Device has no permission to access the render config from the API.
```
user@host:~$ curl -H 'Authorization: Token <YOURTOKEN>' https://netbox/api/dcim/devices/<ID>/render-config/
{"detail":"You do not have permission to perform this action."}
```

If the user is granted add permissions on the `DCIM > Device` Object Types and the API Token is set to `Write Enabled` the access works as expected.

```
user@host:~$ curl -H 'Authorization: Token <YOURTOKEN>' https://netbox/api/dcim/devices/<ID>/render-config/
{"error":"No config template found for this device."}
```

DanSheps:
### Fixes: #14634 - Bypass Write permissions for render endpoint

* Bypass permission for render endpoint by swapping the permissions map and overriding the restrict() action.


jeremystretch:
@DanSheps are you still working on this?

DanSheps:
Yes, sorry, I will get back on it.  Trying to find an elegant way to do this.

netbox/extras/api/mixins.py
```diff
@@ -61,14 +63,23 @@ def render_configtemplate(self, request, configtemplate, context):
 
 
 class RenderConfigMixin(ConfigTemplateRenderMixin):
+    """
+    Override initial() to save a copy of the queryset for "un-restricting" the queryset when rendering.
+    """
+    def initial(self, request, *args, **kwargs):
+        self.original_queryset = self.queryset
+        super().initial(request, *args, **kwargs)
```
DanSheps, line 71: We could alternatively modify:

```python
class BaseViewSet(GenericViewSet):
    """
    Base class for all API ViewSets. This is responsible for the enforcement of object-based permissions.
    """
    def initial(self, request, *args, **kwargs):

        # Restrict the view's QuerySet to allow only the permitted objects
        if request.user.is_authenticated:
            if action := HTTP_ACTIONS[request.method]:
                self.queryset = self.queryset.restrict(request.user, action)
```
to account for permissions using:

self.get_permissions() and using the resulting perms_map to determine what permission should be applied.

This is a down the road thing IMO though.

jeremystretch: We'll definitely need to identify a more maintainable solution. This will probably entail digging into DRF permissions logic a bit.

DanSheps: As mentioned above, I think we can figure out something by modifying the `BaseViewSet` and then adjusting the queryset based on that.

I will take a second crack at this on the weekend.

DanSheps:
Need to get back to this.

netbox/netbox/api/authentication.py
```diff
@@ -123,6 +134,33 @@ def has_object_permission(self, request, view, obj):
...
+        elif action := HTTP_ACTIONS[method]:
+            return action
+        return None
+
+
+class ViewOnlyPermissions(TokenPermissions):
+    """
+    Override the stock perm_map to require only view permissions
+    """
+    perms_map = {
```
jeremystretch, line 154: Rather than overriding `perms_map`, it might be cleaner to just override `get_action()` to return `'%(app_label)s.view_%(model_name)s'` if the request method is not `OPTIONS`.

netbox/netbox/api/authentication.py
```diff
@@ -123,6 +134,33 @@ def has_object_permission(self, request, view, obj):
...
+        if permission:
+            # Remove app and model label
+            action = permission.replace('%(app_label)s.', '').replace('_%(model_name)s', '')
+            return action
+        elif action := HTTP_ACTIONS[method]:
+            return action
+        return None
+
+
+class ViewOnlyPermissions(TokenPermissions):
```
jeremystretch, line 150: On the fence about the name. This might be read as "view-only," implying that a view to which it is attach does not support write operations. Maybe something like `RequireViewPermission`? I'm not sure that's any better.

netbox/netbox/api/authentication.py
```diff
@@ -123,6 +134,33 @@ def has_object_permission(self, request, view, obj):
 
         return super().has_object_permission(request, view, obj)
 
+    # Helper function to obtain the desired action from the first permission in the permission map
+    def get_action(self, method):
+        # Get the first permission
+        permission = self.perms_map.get(method)[0] if len(self.perms_map.get(method)) > 0 else None
+        if permission:
+            # Remove app and model label
+            action = permission.replace('%(app_label)s.', '').replace('_%(model_name)s', '')
```
jeremystretch, line 143: We can use `utilities.permissions.resolve_permission()` to get the action from the permission string.

netbox/netbox/api/viewsets/__init__.py
```diff
@@ -41,8 +33,11 @@ def initial(self, request, *args, **kwargs):
 
         # Restrict the view's QuerySet to allow only the permitted objects
         if request.user.is_authenticated:
-            if action := HTTP_ACTIONS[request.method]:
-                self.queryset = self.queryset.restrict(request.user, action)
+            for permission in self.get_permissions():
+                if hasattr(permission, 'get_action') and (action := permission.get_action(request.method)):
+                    self.queryset = self.queryset.restrict(request.user, action)
+                elif action := HTTP_ACTIONS[request.method]:
+                    self.queryset = self.queryset.restrict(request.user, action)
```
jeremystretch, line 40: Can we just call `permission.get_action()`? Is there a scenario where we wouldn't be using TokenPermissions?

jeremystretch (changes requested):
Great work @DanSheps!

netbox/netbox/api/authentication.py
DanSheps: Thanks, seems like you always have a helper for something.  😄

netbox/netbox/api/viewsets/__init__.py
DanSheps: We could, I was primarily worried there might be a case where there wouldn't be, but I was even on the fence about this.

netbox/netbox/api/authentication.py
DanSheps: What about `RequireViewOnlyPermissions`?

jeremystretch (commented):
@DanSheps can you address the CI failures please?

DanSheps:
> @DanSheps can you address the CI failures please?

I had to make some changes, mainly going back to overriding the permissions map as I couldn't cleanly reslolve certain issues without overriding a number of methods to return view-only permissions.

Reason being is DJango checks to make sure the method has permissions before proceeding to check the more specific permissions and before hitting inital() within the ViewSet.

I found myself overriding get_required_permissions as well as a few others just to get where I needed to be and it wouldn't be what I would deem "clean".  (FWIW, I tried to alter the perms_map in `__init__()`, but it looks like unless it is a local variable to that specific class any changes to it within `__init__()` trickles down to the parent class as well)

netbox/netbox/api/authentication.py
```diff
@@ -123,6 +124,32 @@ def has_object_permission(self, request, view, obj):
...
+            return action
+        return None
+
+
+class RequireViewOnlyPermissions(TokenPermissions):
+    """
+    Overrides permission map to return only view permissions as required
+    """
+    # Override the stock perm_map to enforce view permissions
+    perms_map = {
```
jeremystretch, line 143: This approach isn't sufficient as it doesn't permit the use of a write-disabled API token. We need to override `has_permission()` and `has_object_permission()` on the permission class.

jeremystretch:
Thanks for all your work on this @DanSheps; sorry it turned out to be more trouble than it's worth. I'm going to close this PR as #14634 has been superseded by #16681 per our discussion yesterday. Thanks again!

DanSheps:
> Thanks for all your work on this @DanSheps; sorry it turned out to be more trouble than it's worth. I'm going to close this PR as https://github.com/netbox-community/netbox/issues/14634 has been superseded by https://github.com/netbox-community/netbox/issues/16681 per our discussion yesterday. Thanks again!

It was a fun learning exercise with the Django permissions framework so it isn't all bad.

Issue #12766, opened 2023-05-31
Improve debugging for faulty scripts


candlerb:
### NetBox version

v3.5.2

### Feature type

Change to existing functionality

### Proposed functionality

If you upload a script but there's a problem in it, there's no information displayed beyond this:

![image](https://github.com/netbox-community/netbox/assets/44789/7057fa2d-2335-41f9-8576-899224617fa5)

Nothing is sent to Netbox logs either. I would like the actual exception to be displayed.

### Use case

It's very hard to develop a script if you don't know what's wrong with it!

Using `runscript` from the command line doesn't help: it just returns the same as if you'd specified the wrong name for the class.

```
# python3 /opt/netbox/netbox/manage.py runscript --loglevel debug add_device_type_components.AddDeviceTypeComponents
Traceback (most recent call last):
  File "/opt/netbox/netbox/manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/netbox/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 446, in execute_from_command_line
    utility.execute()
  File "/opt/netbox/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 440, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/netbox/venv/lib/python3.8/site-packages/django/core/management/base.py", line 402, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/netbox/venv/lib/python3.8/site-packages/django/core/management/base.py", line 448, in execute
    output = self.handle(*args, **options)
  File "/opt/netbox/netbox/extras/management/commands/runscript.py", line 93, in handle
    logger = logging.getLogger(f"netbox.scripts.{script.full_name}")
AttributeError: 'NoneType' object has no attribute 'full_name'
```

Trying to run the script directly from the command line, it doesn't have the right environment set up:

```
# python3 add_device_type_components.py
Traceback (most recent call last):
  File "add_device_type_components.py", line 5, in <module>
    from dcim.models import (Manufacturer, DeviceType, Device,
ModuleNotFoundError: No module named 'dcim'
```

Setting PYTHONPATH by itself is insufficient. In the end I used [this recipe](https://gist.github.com/candlerb/195ecd0fcc6aca6f7218d6e312dd0ee1) to run the script, by inserting the following lines at the top:

```
#!/opt/netbox/venv/bin/python
import django, os, sys
[...]

PR #15268, opened 2024-02-24
fix: warning for script errors

llamafilm:
### Improves: #12766

I don't see an easy way to bubble this up to the UI but at least logging with the correct severity makes debugging much easier.


abhi1693:
Please open a new BR for this change. Moreover, it doesn't solve the issue you have raised this for.

PR #15301, opened 2024-02-28
Fixes: #13722 - Correct range expansion code when a numeric set is used

### LINKED ISSUE #13722: Range "568" is invalid when creating interface with names vlan [568,570]

DanSheps:
### Fixes: #13722 - Correct range expansion code when a numeric set is used

* Correct range expansion to account for a numeric set ([1,2])
* Added test

### DIFF
--- a/netbox/utilities/forms/utils.py
+++ b/netbox/utilities/forms/utils.py
@@ -60,29 +60,28 @@ def parse_alphanumeric_range(string):
                 return []
         except ValueError:
             begin, end = dash_range, dash_range
+
+        # Value-based
         if begin == end:
             values.append(begin)
+        # Numeric range-based
         elif begin.isdigit() and end.isdigit():
             if int(begin) >= int(end):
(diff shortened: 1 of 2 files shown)


DanSheps:
This will close #13730

jeremystretch (approved):
Thanks @DanSheps! I think this is even a bit simpler: We can skip the range logic entirely for items without a hyphen in them.

PR #15303, opened 2024-02-28
14799 Fix sync of scripts from data source

arthanson:
### Fixes: #15440  

There were a couple issues with loading scripts from a data source from the DB representation of Scripts which also fix this issue:

1. setting up a local data-source and creating a Script module from it errored as the script objects were getting instantiated within the sync_data.  This was a problem as it was called before the save on ScriptModule so there was no PK for the M2M to save. 
2. module_scripts was also getting called before the M2M was setup so there is now a check for this to return an empty dict.
3. module_scripts was cached which is done only on creation so a sync of the data source would not change the internal classes.  module_scripts now just needs to be called when the ScriptModule potentially changes so the cache can be removed.


jeremystretch:
It looks like this has only been addressed for scripts in v4.0. This needs to be addressed for reports in v3.7 as well.

netbox/extras/models/scripts.py
```diff
@@ -108,7 +108,7 @@ def get_absolute_url(self):
     def __str__(self):
         return self.python_name
 
-    @cached_property
+    @property
```
jeremystretch, line 111: This change and the one below are probably sufficient to fix this in `develop`.

jeremystretch (changes requested):

arthanson:
@jeremystretch openend a new PR #15439 to fix this for reports in NB3.x and created a new issue #15440 to fix the issues in the NB 4.0 branch.

netbox/extras/models/scripts.py
```diff
@@ -137,9 +137,13 @@ def sync_classes(self):
...
-        db_classes = {
-            script.name: script for script in self.scripts.all()
-        }
+        if self.id:
+            db_classes = {
+                script.name: script for script in self.scripts.all()
+            }
+        else:
+            db_classes = {}
+
```
jeremystretch, line 146: Is this change necessary? If `self.id` is null, `self.scripts.all()` should always return an empty queryset IIRC, resulting in an empty dict.

arthanson: Yes, it's because you can't ref an M2M before you save, if this isn't here and the ScriptModule hasn't been saved (no id) then you will get an 'ValueError: 'ScriptModule' instance needs to have a primary key value before this relationship can be used.'

jeremystretch (approved):

PR #15318, opened 2024-02-29
Fixes: #15194 - Check the whole queue for matching queued webhooks

### LINKED ISSUE #15194: Erroneous invocations of Event Rules/Webhooks with M2M relationships
### Deployment Type

Self-hosted

### NetBox Version

v3.7.3-dev

### Python Version

3.11

### Steps to Reproduce

This is probably the most obscure issue I've filed for a long time, so I apologise up front for the difficulties in reproducing it. I did my best to minimise it, but "minimise" in that case means "bring it down to a really involved scenario with non-obvious circumstances". Nevertheless, it's a real issue and it finds one consequence in [this issue](https://github.com/peteeckel/netbox-plugin-dns/issues/89) for the NetBox DNS plugin.

In fact, after some analysis I could trace it down to a very specific situation: There is a problem with incorrectly created Webhooks when an M2M field is updated and the `save()` function of one of the models involved in the M2M relation causes other `save()` operations to occur after `super().save()`. This results in multiple invocations of a Webhook (or, in NetBox >= 4.0, an Event Rule) defined for that model, where the first invocation has wrong data.

### Expected Behavior

Only one invocation of the Webhook with correct data

### Observed Behavior

Two invocations of the Webhook are observed, the first of which has incorrect M2M object data

peteeckel:
### Fixes: #15194 

This PR changes the logic how changes to M2M relationships are matched to changes in the base object to which the M2M relationship is related. The current code assumes that the `post_save` is immediately followed by the `m2m_changed` signal for the former, but that is not generally true as any `save()` operation for a different object occurring in between will separate the two entries in the queue. This results in two webhooks being triggered, the first one of which has incorrect data in its payload.

The new code finds the position of the webhook data for the object change in the queue and updates it with the M2M change data and ensures only one webhook is triggered, and with the correct data.

### DIFF
--- a/netbox/extras/signals.py
+++ b/netbox/extras/signals.py
@@ -55,16 +55,21 @@ def run_validators(instance, validators):
 clear_events = Signal()
 
 
-def is_same_object(instance, webhook_data, request_id):
+def last_matching_index(instance, queue, request_id):
     """
-    Compare the given instance to the most recent queued webhook object, returning True
-    if they match. This check is used to avoid creating duplicate webhook entries.
+    Find the latest queued webhook object matching the given instance, returning its index.
+    If no object is found, return None. This check is used to avoid creating duplicate webhook
+    entries.
     """
-    return (
-        ContentType.objects.get_for_model(instance) == webhook_data['content_type'] and
-        instance.pk == webhook_data['object_id'] and
-        request_id == webhook_data['request_id']
-    )
+    try:
+        return max(
+            index_ for index_, webhook_data in enumerate(queue)
+            if ContentType.objects.get_for_model(instance) == webhook_data['content_type'] and
+            instance.pk == webhook_data['object_id'] and
+            request_id == webhook_data['request_id']
+        )
+    except ValueError:
+        return None
 
 
 @receiver((post_save, m2m_changed))
@@ -112,12 +117,12 @@ def handle_changed_object(sender, instance, **kwargs):
         objectchange.request_id = request.id
         objectchange.save()
 
-    # If this is an M2M change, update the previously queued webhook (from post_save)
+    # Update the previously queued webhook from a preceeding post_save
     queue = events_queue.get()
-    if m2m_changed and queue and is_same_object(instance, queue[-1], request.id):
+    if queue and (last_index := last_matching_index(instance, queue, request.id)) is not None:
         instance.refresh_from_db()  # Ensure that we're working with fresh M2M assignments
-        queue[-1]['data'] = serialize_for_event(instance)
-        queue[-1]['snapshots']['postchange'] = get_snapshots(instance, action)['postchange']
+        queue[last_index]['data'] = serialize_for_event(instance)
+        queue[last_index]['snapshots']['postchange'] = get_snapshots(instance, action)['postchange']
     else:
         enqueue_object(queue, instance, request.user, request.id, action)
     events_queue.set(queue)
--- a/netbox/extras/tests/test_event_rules.py
+++ b/netbox/extras/tests/test_event_rules.py
@@ -16,6 +16,8 @@ from extras.events import enqueue_object, flush_events, serialize_for_event
 from extras.models import EventRule, Tag, Webhook
 from extras.webhooks import generate_signature, send_webhook
 from utilities.testing import APITestCase
+from netbox.tests.dummy_plugin.models import DummyModel
+from netbox.context import events_queue
 
 
 class EventRuleTest(APITestCase):
@@ -31,6 +33,7 @@ class EventRuleTest(APITestCase):
     def setUpTestData(cls):
 
         site_type = ObjectType.objects.get_for_model(Site)
+        dummy_type = ObjectType.objects.get_for_model(DummyModel)
         DUMMY_URL = 'http://localhost:9000/'
         DUMMY_SECRET = 'LOOKATMEIMASECRETSTRING'
 
@@ -65,7 +68,7 @@ class EventRuleTest(APITestCase):
             ),
         ))
         for event_rule in event_rules:
-            event_rule.object_types.set([site_type])
+            event_rule.object_types.set([site_type, dummy_type])
 
         Tag.objects.bulk_create((
             Tag(name='Foo', slug='foo'),
@@ -377,3 +380,45 @@ class EventRuleTest(APITestCase):
         # Patch the Session object with our dummy_send() method, then process the webhook for sending
         with patch.object(Session, 'send', dummy_send) as mock_send:
             send_webhook(**job.kwargs)
+
+    def test_webhook_double_save_create(self):
+        data = {
+            'name': 'Dummy',
+        }
+        url = reverse('plugins-api:dummy_plugin-api:dummymodel-list')
+        self.add_permissions('dummy_plugin.add_dummymodel')
+        response = self.client.post(url, data, format='json', **self.header)
+        self.assertHttpStatus(response, status.HTTP_201_CREATED)
+        self.assertEqual(DummyModel.objects.count(), 1)
+
+        dummy = DummyModel.objects.first()
+
+        self.assertEqual(self.queue.count, 1)
+        job = self.queue.jobs[0]
+        self.assertEqual(job.kwargs['event_rule'], EventRule.objects.get(type_create=True))
+        self.assertEqual(job.kwargs['event'], ObjectChangeActionChoices.ACTION_CREATE)
+        self.assertEqual(job.kwargs['model_name'], 'dummymodel')
+        self.assertEqual(job.kwargs['data']['id'], dummy.pk)
+        self.assertEqual(job.kwargs['snapshots']['postchange']['name'], 'Dummy')
+
+    def test_webhook_double_save_update(self):
+        dummy = DummyModel.objects.create(name='Dummy')
+
+        data = {
+            'name': 'New Dummy',
+            'number': 42,
+        }
+        url = reverse('plugins-api:dummy_plugin-api:dummymodel-detail', kwargs={'pk': dummy.pk})
+        self.add_permissions('dummy_plugin.change_dummymodel')
+        response = self.client.patch(url, data, format='json', **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+        self.assertEqual(DummyModel.objects.count(), 1)
+
+        self.assertEqual(self.queue.count, 1)
+        job = self.queue.jobs[0]
+        self.assertEqual(job.kwargs['event_rule'], EventRule.objects.get(type_update=True))
+        self.assertEqual(job.kwargs['event'], ObjectChangeActionChoices.ACTION_UPDATE)
+        self.assertEqual(job.kwargs['model_name'], 'dummymodel')
+        self.assertEqual(job.kwargs['data']['id'], dummy.pk)
+        self.assertEqual(job.kwargs['snapshots']['postchange']['name'], 'New Dummy')
+        self.assertEqual(job.kwargs['snapshots']['postchange']['number'], 42)
--- a/netbox/netbox/tests/dummy_plugin/api/serializers.py
+++ b/netbox/netbox/tests/dummy_plugin/api/serializers.py
@@ -2,7 +2,7 @@ from rest_framework.serializers import ModelSerializer
 from netbox.tests.dummy_plugin.models import DummyModel
 
 
-class DummySerializer(ModelSerializer):
+class DummyModelSerializer(ModelSerializer):
 
     class Meta:
         model = DummyModel
--- a/netbox/netbox/tests/dummy_plugin/api/views.py
+++ b/netbox/netbox/tests/dummy_plugin/api/views.py
@@ -1,8 +1,8 @@
 from rest_framework.viewsets import ModelViewSet
 from netbox.tests.dummy_plugin.models import DummyModel
-from .serializers import DummySerializer
+from .serializers import DummyModelSerializer
 
 
 class DummyViewSet(ModelViewSet):
     queryset = DummyModel.objects.all()
-    serializer_class = DummySerializer
+    serializer_class = DummyModelSerializer
--- a/netbox/netbox/tests/dummy_plugin/migrations/0001_initial.py
+++ b/netbox/netbox/tests/dummy_plugin/migrations/0001_initial.py
@@ -15,7 +15,7 @@ class Migration(migrations.Migration):
                 ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False)),
                 ('name', models.CharField(max_length=20)),
                 ('number', models.IntegerField(default=100)),
-                ('created', models.DateField(auto_now_add=True, null=True)),
+                ('created', models.DateTimeField(auto_now_add=True, null=True)),
                 ('last_updated', models.DateTimeField(auto_now=True, null=True)),
             ],
             options={
--- a/netbox/netbox/tests/dummy_plugin/migrations/0001_initial.py
+++ b/netbox/netbox/tests/dummy_plugin/migrations/0001_initial.py
@@ -15,6 +15,8 @@ class Migration(migrations.Migration):
                 ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False)),
                 ('name', models.CharField(max_length=20)),
                 ('number', models.IntegerField(default=100)),
+                ('created', models.DateField(auto_now_add=True, null=True)),
+                ('last_updated', models.DateTimeField(auto_now=True, null=True)),
             ],
             options={
                 'ordering': ['name'],
--- a/netbox/netbox/tests/dummy_plugin/models.py
+++ b/netbox/netbox/tests/dummy_plugin/models.py
@@ -1,7 +1,8 @@
 from django.db import models
 
+from netbox.models import EventRulesMixin, ChangeLoggingMixin
 
-class DummyModel(models.Model):
+class DummyModel(EventRulesMixin, ChangeLoggingMixin, models.Model):
     name = models.CharField(
         max_length=20
     )
--- a/netbox/netbox/tests/dummy_plugin/models.py
+++ b/netbox/netbox/tests/dummy_plugin/models.py
@@ -13,3 +13,7 @@ class DummyModel(EventRulesMixin, ChangeLoggingMixin, models.Model):
 
     class Meta:
         ordering = ['name']
+
+    def save(self, *args, **kwargs):
+        super().save()
+        super().save()
--- a/netbox/netbox/tests/dummy_plugin/models.py
+++ b/netbox/netbox/tests/dummy_plugin/models.py
@@ -8,6 +8,7 @@ class DummyModel(models.Model):
     number = models.IntegerField(
         default=100
     )
+    serializer_label = 'netbox.tests.dummy_plugin'
 
     class Meta:
         ordering = ['name']
--- a/netbox/netbox/tests/dummy_plugin/models.py
+++ b/netbox/netbox/tests/dummy_plugin/models.py
@@ -2,6 +2,7 @@ from django.db import models
 
 from netbox.models import EventRulesMixin, ChangeLoggingMixin
 
+
 class DummyModel(EventRulesMixin, ChangeLoggingMixin, models.Model):
     name = models.CharField(
         max_length=20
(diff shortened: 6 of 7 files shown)


peteeckel:
Hi @jeremystretch, I just had a look at the code and it seems that converting the queue to a hash map is not a trivial change.

To me, there seems to be a showstopper in `netbox/extras/events.py` function `flush_events`, which not only expects a queue but also hands it off to the the methods listed in `EVENTS_PIPELINE`, which might contain other code (plugin code, maybe?) that needs to be prepared to handle the hash map instead of the queue.

So while changing it is rather simple, it will definitely break documented or at least established interfaces, which would probably be possible with the switch to NetBox 4 but not earlier, correct?

peteeckel:
Hi @jeremystretch, as this seems to be stalled on the question of the interface to methods in the `EVENTS_PIPELINE` - do you have a preference how to proceed with this? 

I think moving from a queue to a hash map would require changing the (documented?) interface, so the question is whether or not this is an option (or whether I am missing something). If it gets changed, it would probably be best with NetBox 4.0 as a major version.

Or were you suggesting converting the queue to a hash map just for the sake of the check whether an item is already in the queue, without changing the queue structure as such? I doubt this would be an efficient use of resources, as repeatedly copying the data into a hash map is more effort than the linear search I implemented.

If not, do you require any changes to my PR?

jeffgdotorg:
@peteeckel Please add a test that illustrates the behavior you're describing, so that the developer who picks this up can be confident in their understanding of the problem. Additionally, given the sensitivity of the code path that this PR targets, we need to include tests which confirm the continued correct operation of the rest of the code path's behavior. For example, there's some concern about the possibility of race conditions when multiple actors try to modify the same object.

peteeckel:
Hi @jeffgdotorg, that's an interesting request :-)

1. I can't add a test within NetBox because the behaviour can only be reproduced with a specifically crafted `save()` operation that calls a `save()` operation for a different model. AFAIK such a constellation does not exist within NetBox (but within NetBox DNS).
2. I agree that the code path is critical, which is why I am assuming that the existing NetBox test suite already covers it - and the existing test suite passes without a flaw with the patch in place.
3. The code (the existing as well as the patched version) runs within the context of a single request, so I don't see how concurrency could come into play. Is there an existing test in the NetBox test suite that I can use as an example for how that can be done?

peteeckel:
By the way, I **can** add a test to the NetBox DNS test suite. Not sure if that would help, though.

jeremystretch:
@peteeckel my biggest concern, given the nature of the bug, is that some future change to the logging logic could reintroduce the error. If you can walk me through exactly what's happening in the plugin, I might be able to rig up a test to replicate it in NetBox.

peteeckel:
HI @jeremystretch, thanks for the quick reaction. 

I constructed some code to reproduce the behviour in the underlying issue. It can be triggered in multiple ways, all requiring a specifically crafted `save()` function that may or may not affect a model with an associated many-to-many foreign key.

The easiest way to trigger it is to call `super().save()` twice within that save() function - that will cause a double invocation of the event hook, one with a correct pre- and one with a correct post-snapshot. The code in `events.py` does not handle this correctly because it only associates an `m2m_changed` event with an immediately preceding `post_save` event, not two consecutive `post_save` events. 

When there is an m2m relationship, the issue is also triggered when the `save()` method of the model calls the `save()` method of another model, because that also breaks the order of `post_save` immediately followed by `m2m_changed` because `post_save` for the second model is called before the `m2m_changed` event.

The only thing my patch does is to check the whole previous queue instead of only the last event in the queue for events for the same object, and not only `m2m_changed` after `post_save`, which covers both cases.

With your earlier suggestion of changing the data structure of the queue to a hash map, that would be much simpler and more elegant, however at the cost of breaking the interface for the `EVENTS_PIPELINE`. In general, however, nothing would change much.

peteeckel:
I invested best part of today in trying to create a test for this, based on the `dummy_plugin` plugin class in `netbox/netbox/tests`. First I created a custom `save()` method to `DummyModel`,
```
        super().save()
        super().save()
```
then I tried to catch the resulting event queue entries with a new test in `netbox/extras/tests/test_event_rules.py`:
```
    def test_webhook_double_save(self):
        data = {
            'name': 'Dummy',
        }
        url = reverse('plugins-api:dummy_plugin-api:dummymodel-list')
        self.add_permissions('dummy_plugin.add_dummymodel')
        response = self.client.post(url, data, format='json', **self.header)
        self.assertHttpStatus(response, status.HTTP_201_CREATED)
        self.assertEqual(DummyModel.objects.count(), 1)
```
So far this works great, but it doesn't create any event log entries. The culprit is that the model needs to inherit from `ChangeLoggedMixin`, which provides the `to_objectchange()` method needed by the signal handlers. OK, this isn't much of a problem, adding `ChangeLoggedMixin` and `EventRuleMixin` for good measure, and extending the migration for `dummy_plugin`
```
            fields=[
                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False)),
                ('name', models.CharField(max_length=20)),
                ('number', models.IntegerField(default=100)),
                ('created', models.DateField(auto_now_add=True, null=True)),
                ('last_updated', models.DateTimeField(auto_now=True, null=True)),
```
gets me one step further. 

However, now I ran into an old enemy: `get_serializer_for_model()` from `/netbox/utilities/api.py` does not work for plugins. This is also one of the major problems with adapting the test suite for plugins, just that it can be worked around there and not (so far) for the event rule processing. 
```
FAIL: test_webhook_double_save (extras.tests.test_event_rules.EventRuleTest.test_webhook_double_save)
  File "/opt/netbox/netbox/extras/tests/test_event_rules.py", line 391, in test_webhook_double_save
    self.assertHttpStatus(response, status.HTTP_201_CREATED)
  File "/opt/netbox/netbox/utilities/testing/base.py", line 72, in assertHttpStatus
    self.assertEqual(response.status_code, expected_status, err_message)
AssertionError: 500 != 201 : Expected HTTP status 201; received 500: b'{"error": "Could not determine serializer for dummy_plugin.DummyModel with prefix \'\'", "exception": "SerializerNotFound", "netbox_version": "4.0.3", "python_version": "3.11.5"}'
```
This is how far I got today, and now I'm pretty much at the end of my wits for the time being - as it's not a class method but a utility function I can't easily override it.

The problem with `get_serializer_for_model()` can be reproduced easily even without jumping through all the hoops:
```
(netbox) [root@dns netbox]# /opt/netbox/netbox/manage.py nbshell 
### NetBox interactive shell (dns.dev.hindenburgring.com)
### Python 3.11.5 | Django 5.0.6 | NetBox 4.0.3
### lsmodels() will show available models. Use help(<model>) for more info.
>>> from netbox.tests.dummy_plugin.models import DummyModel
>>> from utilities.api import get_serializer_for_model
>>> get_serializer_for_model(DummyModel)
  File "/opt/netbox/netbox/utilities/api.py", line 36, in get_serializer_for_model
    return import_string(serializer_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/lib64/python3.11/site-packages/django/utils/module_loading.py", line 30, in import_string
    return cached_import(module_path, class_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/lib64/python3.11/site-packages/django/utils/module_loading.py", line 15, in cached_import
    module = import_module(module_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'dummy_plugin'

  File "/usr/lib64/python3.11/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/opt/netbox/netbox/utilities/api.py", line 38, in get_serializer_for_model
    raise SerializerNotFound(
netbox.api.exceptions.SerializerNotFound: Could not determine serializer for dummy_plugin.DummyModel with prefix ''
```
This can be reproduced with any plugin. 

I would open an issue for this, but I don't have enough enthusiasm left.

peteeckel:
Sorry, I accidentally deleted the last comment instead of editing it, but never mind - I have some more data now, and a better approach to creating a test for the problem. Essentially there is a little change to the model of the `dummy_plugin` in `netbox/tests`, and some little tests to check for the correct invocation if event rules with this change. The change to the model is actually trivial:
```
        super().save()
        super().save()
```
This is enough to trigger the erroneous invocations of the event rules for `create` and `change` events. Now for the tests:
```
    def test_webhook_double_save_create(self):
        data = {
            'name': 'Dummy',
        }
        url = reverse('plugins-api:dummy_plugin-api:dummymodel-list')
        self.add_permissions('dummy_plugin.add_dummymodel')
        response = self.client.post(url, data, format='json', **self.header)
        self.assertHttpStatus(response, status.HTTP_201_CREATED)
        self.assertEqual(DummyModel.objects.count(), 1)

        dummy = DummyModel.objects.first()

        self.assertEqual(self.queue.count, 1)
        job = self.queue.jobs[0]
        self.assertEqual(job.kwargs['event_rule'], EventRule.objects.get(type_create=True))
        self.assertEqual(job.kwargs['event'], ObjectChangeActionChoices.ACTION_CREATE)
        self.assertEqual(job.kwargs['model_name'], 'dummymodel')
        self.assertEqual(job.kwargs['data']['id'], dummy.pk)
        self.assertEqual(job.kwargs['snapshots']['postchange']['name'], 'Dummy')

    def test_webhook_double_save_update(self):
        dummy = DummyModel.objects.create(name='Dummy')

        data = {
            'name': 'New Dummy',
            'number': 42,
        }
        url = reverse('plugins-api:dummy_plugin-api:dummymodel-detail', kwargs={'pk': dummy.pk})
        self.add_permissions('dummy_plugin.change_dummymodel')
        response = self.client.patch(url, data, format='json', **self.header)
        self.assertHttpStatus(response, status.HTTP_200_OK)
        self.assertEqual(DummyModel.objects.count(), 1)

        self.assertEqual(self.queue.count, 1)
        job = self.queue.jobs[0]
        self.assertEqual(job.kwargs['event_rule'], EventRule.objects.get(type_update=True))
        self.assertEqual(job.kwargs['event'], ObjectChangeActionChoices.ACTION_UPDATE)
        self.assertEqual(job.kwargs['model_name'], 'dummymodel')
        self.assertEqual(job.kwargs['data']['id'], dummy.pk)
        self.assertEqual(job.kwargs['snapshots']['postchange']['name'], 'New Dummy')
        self.assertEqual(job.kwargs['snapshots']['postchange']['number'], 42)
```
I appended the tests appended to `extras/tests/test_event_rules.py`, which seems to be the most fitting place to me.

The problem is: This does not work **at all**.

peteeckel:
The first problem is that the updates to `DummyModel` objects do not trigger any EventRule. The reason is that in order to do that, the object needs to have the `to_objectchange()` method, and `DummyModel` does not have it. So I changed the model to inherit from `EventRuleMixin` und `ChangeLoggedMixin` (not sure if the former is required, but I threw it in for good measure).
```
diff --git a/netbox/netbox/tests/dummy_plugin/models.py b/netbox/netbox/tests/dummy_plugin/models.py
index 9bd39a46b..0e818d8ec 100644
--- a/netbox/netbox/tests/dummy_plugin/models.py
+++ b/netbox/netbox/tests/dummy_plugin/models.py
@@ -1,7 +1,8 @@
 from django.db import models
 
+from netbox.models import EventRulesMixin, ChangeLoggingMixin
 
-class DummyModel(models.Model):
+class DummyModel(EventRulesMixin, ChangeLoggingMixin, models.Model):
     name = models.CharField(
         max_length=20
     )
@@ -11,3 +12,7 @@ class DummyModel(models.Model):
 
     class Meta:
         ordering = ['name']
+
+        super().save()
+        super().save()
```
This is where the trouble **really** started.

peteeckel:
The change logging functionality relies on serialising objects, and that requires to find the serializer for the model in the first place. This is done via `utilities.api.get_serializer_for_model()`, and that just fails.
```
### NetBox interactive shell (dns.dev.hindenburgring.com)
### Python 3.11.5 | Django 5.0.6 | NetBox 4.0.4-dev
### lsmodels() will show available models. Use help(<model>) for more info.
>>> from netbox.tests.dummy_plugin.models import DummyModel
>>> from utilities.api import get_serializer_for_model
>>> get_serializer_for_model(DummyModel)
  File "/opt/netbox/netbox/utilities/api.py", line 36, in get_serializer_for_model
    return import_string(serializer_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/lib64/python3.11/site-packages/django/utils/module_loading.py", line 30, in import_string
    return cached_import(module_path, class_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/lib64/python3.11/site-packages/django/utils/module_loading.py", line 15, in cached_import
    module = import_module(module_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'dummy_plugin'

  File "/usr/lib64/python3.11/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/opt/netbox/netbox/utilities/api.py", line 38, in get_serializer_for_model
    raise SerializerNotFound(
netbox.api.exceptions.SerializerNotFound: Could not determine serializer for dummy_plugin.DummyModel with prefix ''
```
After a while I found out that the problem is with the way `get_serializer_for_model()` constructs the name of the serializer. The `model._meta.label` is split at the dot, then `api.serializers` and an optional `prefix` is inserted between them, and then `Serializer` is added to the end. This gives `dummy_plugin.DummyModelSerializer`. Unfortunately, due to the model path of the dummy plugin, that is not the path where the serializer can be found, hence the exception. 

To be honest, I did not find a real solution for this. I have a very dirty hack in place that just works and doesn't break anything else, but it's ugly as hell and definitely can't be used like this.

```
(netbox) [root@dns netbox]# git diff 0bfb9777bed4fcf22f5b684504b1e62d03cdfb9b 3c38f51250710c6637667ef626f0bc484760ead9 
diff --git a/netbox/netbox/tests/dummy_plugin/models.py b/netbox/netbox/tests/dummy_plugin/models.py
index 9bd39a46b..ade46f768 100644
--- a/netbox/netbox/tests/dummy_plugin/models.py
+++ b/netbox/netbox/tests/dummy_plugin/models.py
@@ -8,6 +8,7 @@ class DummyModel(models.Model):
     number = models.IntegerField(
         default=100
     )
+    serializer_label = 'netbox.tests.dummy_plugin'
 
     class Meta:
         ordering = ['name']
diff --git a/netbox/utilities/api.py b/netbox/utilities/api.py
index 11b914811..3cedfc4a3 100644
--- a/netbox/utilities/api.py
+++ b/netbox/utilities/api.py
@@ -30,7 +30,12 @@ def get_serializer_for_model(model, prefix=''):
     """
     Return the appropriate REST API serializer for the given model.
     """
-    app_label, model_name = model._meta.label.split('.')
+    if hasattr(model, 'serializer_label'):
+        app_label = model.serializer_label
+        model_name = model._meta.label.split('.')[1]
+    else:
+        app_label, model_name = model._meta.label.split('.')
+
     serializer_name = f'{app_label}.api.serializers.{prefix}{model_name}Serializer'
     try:
         return import_string(serializer_name)
```

For the time being I left that hack in place. One way to do it properly would probably be to add the `serializer_path` as an official `Meta` variable, but given the specialty of this use case it's probably not the best way to handle it either. @jeremystretch hopefully has a better idea.

peteeckel:
The next problem, a trivial one, was that the serializer for `DummyModel` was named `DummySerializer`, which doesn't work either (same reason as above). But unlike the last issue this is easy to fix.
```python
diff --git a/netbox/netbox/tests/dummy_plugin/api/serializers.py b/netbox/netbox/tests/dummy_plugin/api/serializers.py
index 239d7d998..0a6412095 100644
--- a/netbox/netbox/tests/dummy_plugin/api/serializers.py
+++ b/netbox/netbox/tests/dummy_plugin/api/serializers.py
@@ -2,7 +2,7 @@ from rest_framework.serializers import ModelSerializer
 from netbox.tests.dummy_plugin.models import DummyModel
 
 
-class DummySerializer(ModelSerializer):
+class DummyModelSerializer(ModelSerializer):
 
     class Meta:
         model = DummyModel
diff --git a/netbox/netbox/tests/dummy_plugin/api/views.py b/netbox/netbox/tests/dummy_plugin/api/views.py
index 58f221285..3fd51bcf2 100644
--- a/netbox/netbox/tests/dummy_plugin/api/views.py
+++ b/netbox/netbox/tests/dummy_plugin/api/views.py
@@ -1,8 +1,8 @@
 from rest_framework.viewsets import ModelViewSet
 from netbox.tests.dummy_plugin.models import DummyModel
-from .serializers import DummySerializer
+from .serializers import DummyModelSerializer
 
 
 class DummyViewSet(ModelViewSet):
     queryset = DummyModel.objects.all()
-    serializer_class = DummySerializer
+    serializer_class = DummyModelSerializer
```

And that's pretty much all ... with these changes in place there is a test that fails if the proposed patch is not in place and succeeds if it is. 

The only major issue is the problem with `get_serializer_for_model()`. If there is a way to have the method find the serializer even if it is not in the top level directory, we have a test.

By the way, I experimented very briefly with setting `app_label` to `netbox.tests.dummy_plugin` .... failed miserably, `app_label` is, among other puroposes, also used to construct the table names for SQL queries.

peteeckel:
> By the way, I experimented very briefly with setting app_label to netbox.tests.dummy_plugin .... failed miserably, app_label is, among other puroposes, also used to construct the table names for SQL queries.

And, as a side note, it also requires changing `get_serializer_for_model()` ever so slightly - `model._meta.label.split('.')` needs to become `model._meta.label.rsplit('.', 1)`. But anyway, changing `app_label` gets you nowhere anyway.

jeremystretch:
@peteeckel I think you've been on the right path with this approach, but employing unique IDs to map each object in the queue offers better protection against duplicated events. We can also simplify it a good deal by invoking the `event_tracking()` context manager manually in the test to avoid any modifications to existing models.

Please have a look at the PR I just put in (#16366) and let me know what you think!

PR #15319, opened 2024-02-29
Fixes #15220: Move IP mask validation logic from form to model

### LINKED ISSUE #15220: Assigned network IP possible via bulk edit
### Deployment Type

Self-hosted

### NetBox Version

v3.7.2

### Python Version

3.11

### Steps to Reproduce

1. Create multiple IP addresses in 192.168.0.0/24, including 192.168.0.128/24 assigned to an interface
2. Mark the IP addresses in the overview and click „Edit selected“
3. Set „Mask length“ to 25 and click „Apply“

### Expected Behavior

I would expect the same behavior like creating 192.168.0.128/25 directly. This is not possible because it is a network address. NetBox complains about that with a warning popup and doesn't save the IP address.

### Observed Behavior

There now is a IP address 192.168.0.128/25

screenshot_ip_addresses

jeremystretch:
### Fixes: #15220

- Moves the logic for validating IP address mask from the form to the model

### DIFF
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -367,20 +367,6 @@ class IPAddressForm(TenancyForm, NetBoxModelForm):
                 'primary_for_parent', _("Only IP addresses assigned to an interface can be designated as primary IPs.")
             )
 
-        # Do not allow assigning a network ID or broadcast address to an interface.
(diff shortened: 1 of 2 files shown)


DanSheps (approved):
Tested and confirmed bug is fixed.  Double checked logic to ensure network ips are not impacted when assigning to an interface with a /31|/32 or /127|/128

Looks good!

Issue #10587, opened 2022-10-06
Implement pagination, filtering, and export functionality for report and script results


atownson:
### NetBox version

v3.3.5

### Feature type

New functionality

### Proposed functionality

My proposal is to implement pagination, filtering, and export options for the report Report Results panel and the script Script Log panel.

For pagination, implement pagination for reports (Report Results panel) and scripts (Script Log panel) in the same manner as the model table pages, with a **Per Page** input.

As far as filtering for reports, the current **Report Methods** panel provides hyperlinks to the method results in the **Report Results** panel. Instead of grouping the results by **Method** in the **Report Results** panel, create a new "Method" column which could be filtered. Perhaps the **Report Methods** panel hyperlinks could be transitioned to provide quick filters rather than hyperlinks. My suggestion for filtering options would be to provide dropdown inputs to filter the **Method** and **Level** fields and textboxes to filter the **Object** and **Message** fields. Filters could be provided in a **Filters** tab, similar to the model table pages, to conserve real estate on the page.

My suggestion for filtering the **Script Log** for scripts would be to provide a dropdown input to filter the **Level** field and a textbox to search the **Message** field. 

For export, implement exporting for reports (Report Results panel) and scripts (Script Log panel) in a similar manner as the model table pages, with an **Export** dropdown. I don't see a need for export template functionality. But rather simply "Export All" and "Export Filtered" options to export all results and the currently filtered results respectively.

### Use case

Pagination would prevent the page content from becoming too large if many results are returned.

Filtering would allow users to better organize or search the results.

Export functionality would allow the results to be used outside the NetBox environment (i.e. emailed to users to correct certain modeling mistakes).

### Database changes

None

### External dependencies

None

atownson:
Sure, here's an illustration for Scripts:
![image](https://user-images.githubusercontent.com/52260120/194366299-8d829610-a6c0-4fb0-bf1f-b5d7c15daf5f.png)

atownson:
@kkthxbye-code, here's an illustration for Reports:
![image](https://user-images.githubusercontent.com/52260120/194371854-516aed62-f08c-4483-a106-c9602bcd2d8e.png)

jeremystretch:
Related: #9376

renatoalmeidaoliveira:
@atownson since you are working in scripts, maybe some kind of "script hook" could be a nice feature.
That way the user can setup a script to be runned when something changes in Netbox

jeremystretch:
[...]

PR #15343, opened 2024-03-04
10587 script pagination

arthanson:
### Fixes: #10587 

Makes the script logs use a tables2 table which allows pagination and sorting of columns.  There were a couple bugs in the scripts migrations which this fixes (was causing jobs not to get correctly mapped from old script/report runs).

Original FR asked for filtering and export, however the logs are stored as a dict and filtering requires a queryset so filtering is not possible without converting the individual logs lines to be stored as regular DB objects.
Table configuration is implemented but not tested as table configuration is currently broken across all views.

Monosnap DeviceConnectionsReport | NetBox 2024-03-04 13-44-08

### DIFF
--- a/netbox/extras/tables/tables.py
+++ b/netbox/extras/tables/tables.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.utils.translation import gettext_lazy as _
 
 from extras.models import *
-from netbox.tables import NetBoxTable, columns
+from netbox.tables import BaseTable, NetBoxTable, columns
 from .template_code import *
 
 __all__ = (
@@ -21,6 +21,8 @@ __all__ = (
     'JournalEntryTable',
     'ObjectChangeTable',
     'SavedFilterTable',
+    'ReportResultsTable',
+    'ScriptResultsTable',
     'TaggedItemTable',
     'TagTable',
     'WebhookTable',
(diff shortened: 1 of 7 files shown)


netbox/extras/tables/tables.py
```diff
@@ -507,3 +509,67 @@ class Meta(NetBoxTable.Meta):
...
+    )
+
+    class Meta(BaseTable.Meta):
+        empty_text = _('No results found')
+        fields = (
+            'index', 'method', 'time', 'status', 'object', 'url', 'message',
+        )
+        default_columns = (
+            'index', 'method', 'time', 'status', 'object', 'url', 'message',
+        )
```
jeremystretch, line 575:
```suggestion
```

netbox/extras/tables/tables.py
```diff
@@ -507,3 +509,67 @@ class Meta(NetBoxTable.Meta):
...
+    status = tables.TemplateColumn(
+        template_code="""{% load log_levels %}{% log_level record.status %}""",
+        verbose_name=_('Level')
+    )
+
+    object = tables.Column(
+        verbose_name=_('Object')
+    )
+    url = tables.Column(
+        verbose_name=_('Url')
```
jeremystretch, line 562:
```suggestion
        verbose_name=_('URL')
```

netbox/extras/views.py
```diff
@@ -1143,19 +1145,71 @@ def get(self, request, module, name, path=''):
...
+                        data.append(result)
+
+            table = ReportResultsTable(data, user=request.user)
+            table.configure(request)
+
+        return table
+
     def get(self, request, **kwargs):
         job = get_object_or_404(Job.objects.all(), pk=kwargs.get('job_pk'))
+        table_logs = table_tests = None
```
jeremystretch, line 1202: Unused?
```suggestion
```

netbox/extras/views.py
```diff
@@ -1143,19 +1145,71 @@ def get(self, request, module, name, path=''):
...
+
+        if job.completed:
+            table = self.get_table(job, request, bulk_actions=False)
 
         context = {
             'script': job.object,
             'job': job,
+            'table': table,
```
jeremystretch, line 1210: `table` will be undefined if `job.completed` is False.

netbox/extras/views.py
```diff
@@ -1143,19 +1145,71 @@ def get(self, request, module, name, path=''):
...
+        tests = None
+        table = None
+        index = 0
+        if job.data:
+            if 'log' in job.data:
+                if 'tests' in job.data:
+                    tests = job.data['tests']
+
+                for log in job.data['log']:
+                    index += 1
```
jeremystretch, line 1165: We can avoid having to track the index manually by wrapping `job.data['log']` with `enumerate()`.

netbox/extras/views.py
```diff
@@ -1143,19 +1145,71 @@ def get(self, request, module, name, path=''):
...
+                    result = {
+                        'index': index,
+                        'time': log.get('time', None),
+                        'status': log.get('status', None),
+                        'message': log.get('message', None),
```
jeremystretch, line 1170: Specifying `None` as the default is redundant if `log` is a dict.
```suggestion
                        'time': log.get('time'),
                        'status': log.get('status'),
                        'message': log.get('message'),
```

netbox/extras/views.py
```diff
@@ -1143,19 +1145,71 @@ def get(self, request, module, name, path=''):
...
+                    }
+                    data.append(result)
+
+                table = ScriptResultsTable(data, user=request.user)
+                table.configure(request)
+            else:
+                tests = job.data
```
jeremystretch, line 1177: This should have a comment explaining the accommodation for legacy reports.

netbox/extras/views.py
```diff
@@ -1143,19 +1145,71 @@ def get(self, request, module, name, path=''):
...
+
+        if tests:
+            for method, test_data in tests.items():
+                if 'log' in test_data:
+                    for time, status, obj, url, message in test_data['log']:
+                        index += 1
```
jeremystretch, line 1183: Same as above; we can call `enumerate()` and pass it the previous index value (if any) as the starting value.

jeremystretch (changes requested):

netbox/extras/views.py
arthanson: I didn't do this one as this second one potentially relies on the count of the first one, so you would need a start_counter defined at top then conditionally set the enumeration start to either the start_counter or the end of the previous list, so I thought index was just more readable as to what is going on.

netbox/extras/views.py
arthanson: see comment below

jeremystretch (approved):

PR #15393, opened 2024-03-10
add REMOTE_AUTH_DISPLAYNAME setting

### LINKED ISSUE #15375: Allow for custom names for Identity Provider text in the Login link
### NetBox version

v3.7.3

### Feature type

### Proposed functionality

Recently `oidc` was added to the auth backend (https://github.com/netbox-community/netbox/issues/14587) to make the login link more user friendly.

In the issue, @llamafilm also requested the ability to customize this text, as there are many different OIDC IdP. (https://github.com/netbox-community/netbox/issues/14587#issuecomment-1877804701)

This request is to add that functionality to specify a custom name in the login link for any social login provider.

### Use case

There are many different social IdPs and users will appreciate knowing which one they are going to be authenticating to.

### Database changes

None

### External dependencies

None

llamafilm:
### Fixes: #15375

Add optional setting for `REMOTE_AUTH_DISPLAYNAME`.  If set, it will override the remote auth provider name shown on the login page.

image


jeremystretch (changes requested):
This change does not account for using multiple authentication backends. A much more flexible approach would be to allow an administrator to override the contents of `AUTH_BACKEND_ATTRS` on a per-backend basis.

llamafilm:
How is it possible to have multiple authentication backends? `REMOTE_AUTH_BACKEND` is a single string variable.

jeremystretch:
> `REMOTE_AUTH_BACKEND` is a single string variable.

Correct, but ideally we should support multiple backends. (I thought there was an FR for this, but maybe not.) Also consider that the active backend may be changed.

llamafilm:
Changing the backend should work fine with this.

Since that feature hasn't been defined yet, I don't think it makes sense to add extra complexity here, guessing how it might be implemented.  For example, someone might wants to use 2 different backends of the same type.

jeremystretch:
I'm sorry but this is not acceptable in its current form. Please extend it to allow overriding the values in `AUTH_BACKEND_ATTRS` as requested.

jeremystretch:
@llamafilm do you intend to continue working on this, or should we release you from the FR?

llamafilm:
I provided a very simple solution that works, and you haven't pointed out any problems with it.  I'm not going to spend any more time on this to support some potential new use case that _might_ be added in the future.

jeremystretch:
As I said [above](https://github.com/netbox-community/netbox/pull/15393#pullrequestreview-1930999335):

> This change does not account for using multiple authentication backends.

I'll go ahead and close this out.

Issue #15237, opened 2024-02-23
Audit FilterSet definitions application-wide


jeremystretch:
### Proposed Changes

 Review all currently defined FilterSet classes and check for:

- Missing filters
- Inconsistent names or behavior
- Suboptimal performance
- Opportunities for code reduction

### Justification

The FilterSet classes defined in NetBox have evolved organically over the past several years and we are overdue for a structured audit of the existing filter sets. This will help us identify inconsistent or suboptimal implementations, and may yield additional work to help standardize and shrink the code base.

PR #15403, opened 2024-03-11
Closes #15383: Standardize filtering logic for the parents of recursively-nested models

### LINKED ISSUE #15383: Standardize filtering logic for the parents of recursively-nested models
### NetBox version

v3.7.3

### Feature type

### Proposed functionality

I'm opening this FR to address an inconsistency in the filtering of parent objects of self-recursive models (e.g. regions, site groups, locations, etc.).

For most of these objects, filtering by the parent object will return only the immediate children of the specified parent. For example, filtering for all regions belonging to North America (PK 1) will return only the countries assigned to that continent; it will _not_ include child states or cities within each country.

```
# Returns immediate children only
GET /api/dcim/regions/?parent_id=1
```

However, this is not true for locations. Per #6124, filtering locations by a parent location will return _all_ of the parent object's descendants. For example, if dealing with a hierarchy of buildings, floors, and rooms, filtering for a specific building will return both rooms and floors within it.

```
# Returns *all* descendants in the hierarchy
GET /api/dcim/locations/?parent_id=1
```

I am proposing a two-part change:

1. Standardize the behavior of the `parent` & `parent_id` filters so that they behave consistently across all nested models. (This will likely mean modifying the current behavior of the location filter, as it would be the smallest breaking change.)
2. Introducing an alternative filters for all models to provide the complementary filtering behavior (perhaps named `ancestor`).

**Edit:** For example, you could then do

```
GET /api/dcim/regions/?parent_id=1
```

to return only the _immediate_ children of region 1, OR

```
GET /api/dcim/regions/?ancestor_id=1
```

to return _all_ regions within region 1.

### Use case

Neither behavior described above is objectively correct or incorrect: Both have valid use cases. However, we should address the inconsistency and prescribe a clear mechanism for achieving either behavior for all relevant models.

### Database changes

None expected

### External dependencies

None

jeremystretch:
### Fixes: #15383

- Modify the `parent` and `parent_id` filters for Location to return only immediate children
- Add `ancestor` and `ancestor_id` filters for all recursively-nested models to return _all_ children

### DIFF
--- a/netbox/wireless/filtersets.py
+++ b/netbox/wireless/filtersets.py
@@ -25,6 +25,17 @@ class WirelessLANGroupFilterSet(OrganizationalModelFilterSet):
         queryset=WirelessLANGroup.objects.all(),
         to_field_name='slug'
     )
+    ancestor_id = TreeNodeMultipleChoiceFilter(
+        queryset=WirelessLANGroup.objects.all(),
+        field_name='parent',
+        lookup_expr='in'
+    )
+    ancestor = TreeNodeMultipleChoiceFilter(
+        queryset=WirelessLANGroup.objects.all(),
+        field_name='parent',
+        lookup_expr='in',
+        to_field_name='slug'
+    )
 
     class Meta:
         model = WirelessLANGroup
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -89,6 +89,19 @@ class RegionFilterSet(OrganizationalModelFilterSet, ContactModelFilterSet):
         to_field_name='slug',
         label=_('Parent region (slug)'),
     )
+    ancestor_id = TreeNodeMultipleChoiceFilter(
+        queryset=Region.objects.all(),
+        field_name='parent',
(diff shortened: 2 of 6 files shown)


netbox/wireless/filtersets.py
```diff
@@ -25,6 +25,17 @@ class WirelessLANGroupFilterSet(OrganizationalModelFilterSet):
         queryset=WirelessLANGroup.objects.all(),
         to_field_name='slug'
     )
+    ancestor_id = TreeNodeMultipleChoiceFilter(
```
arthanson, line 28: All of these in this file look like they are missing the label definition, don't know if you want to address it here or in a separate PR as fields beyond just PR are missing them.

arthanson (approved):
Note comment on labels for wireless..

netbox/wireless/filtersets.py
jeremystretch: Not in scope for #15383; it'll need to be addressed separately (probably under #15237).

PR #15405, opened 2024-03-11
Closes #15401: Rename PostgreSQL tables & indexes for L2VPN models

### LINKED ISSUE #15401: Rename PostgreSQL resources for L2VPN models
### Proposed Changes

Rename the following sequence tables:

* `ipam_l2vpn_export_targets_id_seq`
* `ipam_l2vpn_id_seq`
* `ipam_l2vpn_import_targets_id_seq`
* `ipam_l2vpntermination_id_seq`

As well as the following indexes & constraints:

* `ipam_l2vpn_pkey`
* `ipam_l2vpn_name_5e1c080f_like`
* `ipam_l2vpn_name_key`
* `ipam_l2vpn_slug_24008406_like`
* `ipam_l2vpn_slug_24008406_uniq`
* `ipam_l2vpn_tenant_id_bb2564a6`
* `ipam_l2vpn_tenant_id_bb2564a6_fk_tenancy_tenant_id`
* `ipam_l2vpntermination_pkey`
* `ipam_l2vpntermination_assigned_object_type_id_3923c124`
* `ipam_l2vpntermination_l2vpn_id_9e570aa1`
* `ipam_l2vpntermination_assigned_object_id_check`
* `ipam_l2vpnterminatio_assigned_object_type_3923c124_fk_django_co`
* `ipam_l2vpntermination_l2vpn_id_9e570aa1_fk_ipam_l2vpn_id`

### Justification

The `L2VPN` and `L2VPNTermination` models were migrated from the `ipam` app to the `vpn` app in NetBox v3.7, and their corresponding tables in the PostgreSQL database were renamed accordingly. However, Django does not automatically rename their accompanying sequence tables, indexes, or constraints.

Although retaining the original names for these resources has no functional impact, they should be renamed to match their tables to avoid potential confusion.

jeremystretch:
### Fixes: #15401

Introduces a migration which manually renames the sequences, constraints, and indexes for the L2VPN & L2VPN terminations that were originally created in the `ipam` app and moved to `vpn` under #14311.

### DIFF
--- a/netbox/vpn/migrations/0005_rename_indexes.py
+++ b/netbox/vpn/migrations/0005_rename_indexes.py
@@ -22,8 +22,11 @@ class Migration(migrations.Migration):
         migrations.RunSQL("ALTER INDEX ipam_l2vpn_name_5e1c080f_like RENAME TO vpn_l2vpn_name_8824eda5_like"),
         migrations.RunSQL("ALTER INDEX ipam_l2vpn_name_key RENAME TO vpn_l2vpn_name_key"),
         migrations.RunSQL("ALTER INDEX ipam_l2vpn_slug_24008406_like RENAME TO vpn_l2vpn_slug_76b5a174_like"),
-        migrations.RunSQL("ALTER INDEX ipam_l2vpn_slug_24008406_uniq RENAME TO vpn_l2vpn_slug_76b5a174_uniq"),
         migrations.RunSQL("ALTER INDEX ipam_l2vpn_tenant_id_bb2564a6 RENAME TO vpn_l2vpn_tenant_id_57ec8f92"),
+        # The unique index for L2VPN.slug may have one of two names, depending on how it was created,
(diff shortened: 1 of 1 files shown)


jeremystretch:
This is odd. The index for L2VPN slug seems to get two different names under different conditions. If I check out `develop` and run `manage.py migrate`, the index is created with the name `ipam_l2vpn_slug_24008406_uniq`. I can then check out the PR branch (`15401-rename-sql-indexes`), and applying the migration succeeds.

However, if I start with an empty database in branch `15401-rename-sql-indexes` and run migrations all the way through, the migration fails because the index it's trying to rename is named `ipam_l2vpn_slug_key`. This happens using Django 4.2 for both conditions.

We can easily work around this in the migration file, however I want to ensure we understand _why_ this is happening first. Will dig into it some more.

arthanson (approved):

PR #15415, opened 2024-03-12
Closes #14739: Advanced form fieldset functionality

### LINKED ISSUE #14739: Extend `fieldsets` to support rendering complex form layouts
### NetBox version

v3.7.0

### Feature type

New functionality

### Proposed functionality

Introduce several new classes that would enable more complex form rendering via the `fieldsets` attribute on a form. For example, the following is defined on `SiteForm.Meta`:

```python
fieldsets = (
    (_('Site'), (
        'name', 'slug', 'status', 'region', 'group', 'facility', 'asns', 'time_zone', 'description',
        'tags',
    )),
    (_('Tenancy'), ('tenant_group', 'tenant')),
    (_('Contact Info'), ('physical_address', 'shipping_address', 'latitude', 'longitude')),
)
```

This causes the form to render three discrete groups of fields with the assigned headings:

site_form

This proposal seeks to extend this ability by enabling several forms of more complex rendering, detailed below.

#### Static Attribute Rendering

Display an object attribute (not necessarily a form field) of an existing object, rendered as a read-only field.

```python
fieldsets = (
    (None, (
        ObjectAttribute('parent'), 'name', 'image',
    )),
)
```

image_attachment

#### Side-by-Side Fields

Display several fields inline on a single row within the form.

```python
fieldsets = (
    (_('Dimensions'), (
        'type', 'width', 'u_height', InlineFields('outer_width', 'outer_depth', 'outer_unit'),
    )),
)
```

rack_dimensions

#### Tabbed Groups

Allow fields to be arranged in tabbed subgroups. Selecting the tab for a group will display its fields and hide fields from its peers.

```python
fieldsets = (
    (
        _('Interface Assignment'),
        TabbedFieldGroups(
            (_('Device'), 'interface'),
            (_('VirtualMachine'), 'vminterface'),
            (_('FHRP Group'), 'fhrpgroup'),
        ), 'primary_for_parent,
    ),
)
```

ipaddress_assignment

### Use case

There are several instances across NetBox where we use custom HTML templates to render object edit forms, because we need to deviate from the standard form rendering and accommodate a more complex layout. Some instances include:

* Displaying an attribute as a read-only field (e.g. the object to which an image attachment is assigned)
* Grouping several fields under tabs (e.g. IP address interface/FHRP group)
* Displaying several fields inline (e.g. rack width, depth, and unit)

Moving this logic into the form fieldset definitions will allow us to ditch most of our custom templates.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #14739

- Introduces `ObjectAttribute` to display read-only object attributes on a form
- Introduces `InlineFields` to lay out multiple fields side-by-side
- Introduces `TabbedFieldGroups` to organize fields within tabs
- Define `fieldsets` on the following forms & remove their custom templates:
  - `circuits.CircuitTerminationForm`
  - `dcim.InventoryItemForm`
  - `dcim.RackForm`
  - `extras.ImageAttachmentForm`
  - `ipam.FHRPGroupAssignmentForm`
  - `ipam.IPAddressForm`
  - `ipam.ServiceForm`
  - `ipam.ServiceCreateForm`
  - `ipam.VLANForm`
  - `tenancy.ContactAssignmentForm`
  - `vpn.L2VPNTerminationForm`


netbox/utilities/forms/rendering.py
```diff
@@ -0,0 +1,49 @@
...
+        return [
+            {
+                'id': f'{self.id}_{i}',
+                'title': group.label,
+                'fields': group.field_names,
+            } for i, group in enumerate(self.groups, start=1)
+        ]
+
+
+class ObjectAttribute:
```
arthanson, line 46: I wonder if instead of ObjectAttribute you could use readonly_fields and just display those as values, seems more consistent with Django and django-admin usage.

netbox/utilities/forms/rendering.py
```diff
@@ -0,0 +1,49 @@
...
+from functools import cached_property
+
+__all__ = (
+    'InlineFields',
+    'ObjectAttribute',
+    'TabbedFieldGroups',
+)
+
+
+class FieldGroup:
```
arthanson, line 12: Add docstrings to all of these

netbox/utilities/templates/form_helpers/render_fieldset.html
```diff
@@ -0,0 +1,65 @@
...
+            {{ items.0|linkify }}
+          </div>
+        </div>
+      </div>
+
+    {% elif layout == 'inline' %}
+      {# Multiple form fields on the same line #}
+      <div class="row mb-3">
+        <label class="col col-3 col-form-label text-lg-end">{{ title|default:'' }}</label>
+        {% for field in items %}
```
arthanson, line 30: Maybe overkill, but this seems set for 3 wide, might be more flexible if could specify maybe 2 wide for some.  But could be future FR

arthanson (changes requested):
Very nice, would request docstrings and adding pages to the documentation on usage.

netbox/utilities/forms/rendering.py
jeremystretch: The `ObjectAttribute` class is needed to display values that are not tied to actual form fields, but exist as attributes on its related instance.

netbox/utilities/templates/form_helpers/render_fieldset.html
jeremystretch: We want to ensure that all row labels render with the same column width to effect a nice vertical alignment.

arthanson (approved):

PR #15456, opened 2024-03-17
Closes: #13918 - Add facility field

### LINKED ISSUE #13918: Add the field Facility to the Location model
### NetBox version

v3.6.3

### Feature type

Data model extension

### Proposed functionality

Add the field `facility` to the `Location` model the same way it's done on [the `Site` model](https://github.com/netbox-community/netbox/blob/v3.6.3/netbox/dcim/models/sites.py#L180-L185) 

### Use case

Most of the time the _Facility_ is not a whole building but rather a Suite (Location) in a building (Site). You may also have more than one Facility inside the same Site or have a single facility in a suite in the building but also other tenants in their own Locations which are not part of that Facility.

For example: 
**151 Front St., Toronto, ON Canada; Canada's largest Carrier Hotel.**
Equinix TR1: Suite 706
Neutral Data Toronto: Suite 800
BELL Canada: Suite 605
Coloware YTZ1: Suite 504
Cologix TOR1: Suite 602
eStruxture TOR-1: Suite 341
Aptum: Suite 402
... and many others

Being able to document this Facility name/ID in a consistent manner across NetBox seems only reasonable.

### Database changes

A new CharField, `facility` needs to be added to the `Location` model.

We can use [the `Site` model](https://github.com/netbox-community/netbox/blob/v3.6.3/netbox/dcim/models/sites.py#L180-L185)  for reference.

### External dependencies

None

DanSheps:
### Closes: #13918 - Add facility field to Location model

* Add facility field to model
* Add filter field
* Add form fields
* Add tests
* Add serializer field
* Adjust template
* Adjust docs


jeremystretch (approved):
Nicely done! Thanks @DanSheps

Issue #15153, opened 2024-02-14
Customize theme & styling of browseable REST API


jeremystretch:
### NetBox version

v3.7.2

### Feature type

Change to existing functionality

### Proposed functionality

When requested by a browser, NetBox's REST API endpoints return a human-readable HTML rendering of the relevant data (a nice bonus feature of [Django REST framework](https://www.django-rest-framework.org/topics/browsable-api/)), as seen below.

![screenshot](https://github.com/netbox-community/netbox/assets/13487278/32ebc83b-ed51-4ad2-a9e0-ab21a7a1a88d)

In light of the [complete UI refresh]() for NetBox v4.0, we should take some time to incorporate elements of the new theme for the browseable API as well. This can be done by overriding and extending DRF's default HTML templates.

### Use case

Customization of the browseable REST API will provide a more consistent feel and better UX.

### Database changes

N/A

### External dependencies

N/A

PR #15474, opened 2024-03-19
15153 customize styling browseable rest api

arthanson:
### Fixes: #15153 

Limits on what can be changed without copying and modifying the base template into api.html, but made tweaks to bring it more in-line with app.  Just kept inline css as fairly minor changes needed and IMHO not worth tooling for separate css file.

Monosnap Circuits | NetBox REST API 2024-04-01 09-35-12


netbox/templates/rest_framework/api.html
```diff
@@ -2,6 +2,33 @@
 {% load static %}
 {% load i18n %}
 
+{% block bootstrap_theme %}
+  <link rel="stylesheet" type="text/css" href="{% static "rest_framework/css/bootstrap.min.css" %}"/>
+  <style>
```
jeremystretch, line 7: We shouldn't have inline CSS anywhere. Any custom styles should be defined as SASS under `project-static/styles/` along with the native stylesheets. We can build and package a custom CSS just for the REST web UI.

jeremystretch (changes requested):
It's a good start, but I see a few issues:

1. The navbar is not aligned properly
2. No visual contrast between the navbar and page content
3. Too much white space

rest_web_ui

arthanson:
Changed to dark header and adjusted spacing / alignment of items

jeremystretch:
We're going to punt on this one for now and revisit once we've spent more time improving the overall UI design.

PR #15475, opened 2024-03-19
Fixes #15340: Fix flicker on page load with dark mode enabled

### LINKED ISSUE #15340: Fix flicker on page load when color mode is set to dark
### Deployment Type

NetBox Cloud

### NetBox Version

v4.0

### Python Version

3.11

### Steps to Reproduce

1. Open the NetBox UI as an anonymous user (not logged in)
2. Click the toggle button at top right to switch to dark mode
3. Navigate among pages several times

### Expected Behavior

The application of the dark mode theme should persistent without interruption between page loads.

### Observed Behavior

A "flicker" is sometimes visible when a new page loads, as the DOM is initially rendered in light mode and then switched to dark mode. This occurs because the user's preferred theme is being read from local storage _after_ the DOM has finished loading.

Note that this does _not_ occur when authenticated as a known user with the color preference set to dark mode. (In this case, `data-bs-theme="dark"` is set on each page's `<body>` tag directly in the page source.)

jeremystretch:
### Fixes: #15340

- Call `initMode()` on page load
- Set `data-bs-theme` attribute on `<html>` element

### DIFF
--- a/netbox/project-static/js/setmode.js
+++ b/netbox/project-static/js/setmode.js
@@ -5,10 +5,11 @@
  * @param inferred {boolean} Value is inferred from browser/system preference.
  */
 function setMode(mode, inferred) {
-    document.documentElement.setAttribute("data-netbox-color-mode", mode);
+    document.documentElement.setAttribute("data-bs-theme", mode);
     localStorage.setItem("netbox-color-mode", mode);
     localStorage.setItem("netbox-color-mode-inferred", inferred);
 }
+
 /**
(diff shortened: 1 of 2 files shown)


arthanson (changes requested):
As logged in user I set the dark-mode but when I navigate to a new page it goes back to light mode.  Switched back to feature and it stays dark mode (with flicker) but with this branch it keeps switching back.

arthanson (approved):

PR #15480, opened 2024-03-19
Closes #15339: Consume entire viewport

### LINKED ISSUE #15339: Limit screen width for object detail views but use full screen for list views
### NetBox version

v3.7.3

### Feature type

### Proposed functionality

The v3 UI currently renders all views using the full screen width. On very wide viewports, this can result in unnatural layout using too much horizontal whitespace for object detail views. Conversely, the new v4.0 UI in development limits all views to a maximum width; this should be extended for list views which typically benefit from the additional horizontal space.

### Use case

Utilizing different maximum content widths depending on the context provides better UX.

### Database changes

None

### External dependencies

None

jeremystretch:
### Fixes: #15339

- Tweak the base templates to render to the full width of the viewport ("fluid") by default
- Introduce the `page_width` context variable to override this for object detail views

I'm not super happy with this solution, but it seems the only feasible choice without upending our entire template structure.

### DIFF
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -73,7 +73,6 @@ class ObjectView(BaseObjectView):
         return render(request, self.get_template_name(), {
             'object': instance,
             'tab': self.tab,
-            'page_width': 'xl',
             **self.get_extra_context(request, instance),
         })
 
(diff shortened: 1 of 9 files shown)


arthanson (approved):
I'd vote for wide for all screens - I have a very wide monitor and I'd still prefer it wide then narrowed down, although I realize this is very subjective.

jeremystretch:
Should we just leave everything at full width? I don't have a strong opinion.

arthanson:
IMHO full width looks better - the right-left dead space borders looks strange
Monosnap PP:B117 | NetBox 2024-03-19 13-33-49

Monosnap PP:B117 | NetBox 2024-03-19 13-35-29

jeremystretch:
I've modified the PR to just use full width (fluid) containers for all views.

PR #15491, opened 2024-03-20
Closes #14279: Pass current request to custom validators

### LINKED ISSUE #14279: Extend customvalidators with user request information
### NetBox version

v.3.6.5

### Feature type

### Proposed functionality

Extend customvalidators with an additional variable - request user details

### Use case

It would be great it customvalidators had a way to check what users are making the request. in some situations, it can make sense to allow one user/usergroup/API integration the ability to skip a specific validation.

for example we wish to use an ipaddress validator to make sure rfc 1918 ip's don't end up in the global vrf, however there is one particular api integration that is going to take quite awhile to update to support vrf lookup's, in the meantime we want to stop any other rfc 1918 ip's getting added to the global vrf. being able to write the validator with an exclusion for a particular user would be great.

I understand that sometimes the supplied user would also need to be None/False as changes from nbshell wouldn't have an attached user. 

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Fixes: #14279

- Move `run_validators()` to `extras.signals`
- Extend `run_validators()` to pass the current request when running a validator
- Extend the `__call__()` method on CustomValidator to accept and process a request in addition to an instance
- Add a test for request validation

### DIFF
--- a/netbox/extras/signals.py
+++ b/netbox/extras/signals.py
@@ -1,3 +1,4 @@
+import importlib
 import logging
 
 from django.contrib.contenttypes.models import ContentType
@@ -13,7 +14,6 @@ from core.signals import job_end, job_start
 from extras.constants import EVENT_JOB_END, EVENT_JOB_START
 from extras.events import process_event_rules
 from extras.models import EventRule
-from extras.validators import run_validators
 from netbox.config import get_config
 from netbox.context import current_request, events_queue
 from netbox.models.features import ChangeLoggingMixin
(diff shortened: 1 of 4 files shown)


netbox/extras/signals.py
```diff
@@ -22,6 +22,27 @@
...
+    for validator in validators:
+
+        # Loading a validator class by dotted path
+        if type(validator) is str:
+            module, cls = validator.rsplit('.', 1)
+            validator = getattr(importlib.import_module(module), cls)()
+
+        # Constructing a new instance on the fly from a ruleset
+        elif type(validator) is dict:
+            validator = CustomValidator(validator)
```
arthanson, line 42: should we put in an assert here if neither type?

arthanson (approved):

netbox/extras/signals.py
jeremystretch: The third option to pass a CustomValidator instance directly. I'll add a check that `validator` is a subclass of CustomValidator.

PR #15508, opened 2024-03-22
Fixes: #15502 - Correct exception on IP import form when VM is target

### LINKED ISSUE #15502: Cannot set Primary IP for VM when Importing IP with CSV
### Deployment Type

Self-hosted

### NetBox Version

v3.7.4

### Python Version

3.8

### Steps to Reproduce

1. Attempt to assign an IP to a VM interface by importing it in CSV format, as an example on the online demo:
```
address,id,virtual_machine,interface,is_primary,status
172.16.0.3/24,33,vm6,eth0,true,active
```
2. Receive the following error:
```
<class 'KeyError'>

'device'

Python version: 3.8.10
NetBox version: 3.7.4
Plugins: 
  netbox_demo: 0.3.1
```

### Expected Behavior

The IP Address to be updated and assigned to the interface on that VM, as well as set as the primary IP. The GUI lists the Primary IP checkbox as being for both devices and VMs, and there is no other listed import field specifically for VMs.

### Observed Behavior

Returned a KeyError about missing the "device" field.

DanSheps:
### Fixes: #15502 - Correct exception on IP import form when VM is target

* Modify save() to use .get() around cleaned data when determining parent

### DIFF
--- a/netbox/ipam/forms/bulk_import.py
+++ b/netbox/ipam/forms/bulk_import.py
@@ -378,7 +378,7 @@ class IPAddressImportForm(NetBoxModelImportForm):
 
         # Set as primary for device/VM
         if self.cleaned_data.get('is_primary'):
-            parent = self.cleaned_data['device'] or self.cleaned_data['virtual_machine']
+            parent = self.cleaned_data.get('device') or self.cleaned_data.get('virtual_machine')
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @DanSheps!

Issue #15154, opened 2024-02-14
Replace gunicorn with uWSGI as NetBox's default WSGI service


jeremystretch:
### Change Type

Addition

### Area

Installation/upgrade

### Proposed Changes

In light of [recent issues](https://github.com/netbox-community/netbox/issues/14691) with gunicorn that don't seem likely to be resolved soon, I wanted to float the idea of replacing gunicorn with [uWSGI](https://uwsgi-docs.readthedocs.io/) in NetBox's reference deployment stack. This would not effect any changes to NetBox itself, but would modify the installation & upgrade documentation and replace the reference [gunicorn config](https://github.com/netbox-community/netbox/blob/develop/contrib/gunicorn.py).

For more context around the proposed change, please see discussion #14780. (Thank you @markkuleinio and @tyler-8 for sharing your experiences!)

PR #15523, opened 2024-03-25
15154 Replace gunicorn with uWSGI as NetBox default WSGI server

arthanson:
### Fixes: #15154 

I kept gunicorn documentation around as an alternative.  uwsgi.ini file is a minimal config (no workers defined, just using defaults).

@jeremystretch re-did from scratch on a clean Ubuntu system, but probably a good idea if you can install with the steps given just to double-check.

**Note:** Couple related issues / questions:
* The upgrade script text refers to gunicorn - I currently assume we don't want to switch them to uWSGI if they already have gunicorn up and working?
* I've removed gunicorn from requirements, but if we do that it will never get updated when they update NetBox if they are using it.


jeremystretch:
> The upgrade script text refers to gunicorn - I currently assume we don't want to switch them to uWSGI if they already have gunicorn up and working?

We should extend the script text to include the binary locations for both gunicorn and uWSGI for reference.

> I've removed gunicorn from requirements, but if we do that it will never get updated when they update NetBox if they are using it.

IMO the cleanest path would be to have the admin add either gunicorn or uWSGI to `local_requirements.txt`, but I'm not sure how best to handle that. We may want to rethink our packaging a bit.

Issue #15532, opened 2024-03-26
autotype_decorator fails on annotated fields


arthanson:
### Deployment Type

Self-hosted

### NetBox Version

v4.0

### Python Version

3.8

### Steps to Reproduce

1. Create an annotated field for a model
2. Add it to the filterset for the model
3. Use the autotype_decorator for the filterset

Run into by peteeckel when migrating the DNS plugin to NB 4

**For example:**

Definition of the annotation:
```
    def get_queryset(self):
        return (
            super()
            .get_queryset()
            .annotate(
                active=ExpressionWrapper(
                    Q(status__in=Zone.ACTIVE_STATUS_LIST), output_field=BooleanField()
                )
            )
        )
```
Definition in the filterset:
```
    active = django_filters.BooleanFilter(
        label="Zone is active",
    )
```

### Expected Behavior

autotype_decorator should not throw an error

### Observed Behavior

autotype_decorator throws a FieldDoesNotExist error.

peteeckel:
I'm not sure a fix is necessary after all, at least not for the annotation I added.

After removing the decorator from `fields` and just leaving a `BooleanFilter` for the annotated field in place, the `@autotype_decorator` created a functional filter without any further measures:

```
$ curl -s -H "Authorization: Token xxxxxxxxxxxxxxxxxxxxx" -H "Content-Type: application/json" -H "Accept: application/json"  https://netbox.example.com/graphql/ --data '{"query": "query {netbox_dns_zone_list(filters: {active: false}) {view {name}, name, status, active}}" }' | json_pp
{
   "data" : {
      "netbox_dns_zone_list" : [
         {
            "active" : false,
            "name" : "zone13.example.com",
            "status" : "reserved",
            "view" : null
         }
      ]
   }
}
```

So after removing the annotation from `fields` all seems to be OK.

jeremystretch:
@arthanson can you confirm the above? And if not, could you please assign a severity to the bug?

PR #15533, opened 2024-03-26
15532 fix autotype_decorator for method fields

arthanson:
### Fixes: #15532 

Fixes the case found in migrating the DNS plugin to NB 4 where an annotated field was declared (therefore not appearing in the model) and filtered on.  This causes an error as the current code expects fields to be in the model definition. 

Fix is to skip any undefined fields, they will need to be explicitly defined in the GraphQL filter class.

### DIFF
--- a/netbox/netbox/graphql/filter_mixins.py
+++ b/netbox/netbox/graphql/filter_mixins.py
@@ -4,6 +4,7 @@ from typing import List
 import django_filters
 import strawberry
 import strawberry_django
+from django.core.exceptions import FieldDoesNotExist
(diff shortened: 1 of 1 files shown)


jeremystretch:
Are we sure this change is needed? Per @peteeckel's comment [here](https://github.com/netbox-community/netbox/issues/15532#issuecomment-2020872838) it seems not to be needed.

peteeckel:
At least for my use case it is **not** required provided the annotation is not listed in the `fields` list of the filter.

jeremystretch (approved):

PR #15599, opened 2024-04-02
Fixes #15597: Document the value behind "Default" for the Custom Link "Button Class"

### LINKED ISSUE #15597: Document the value behind "Default" for the Custom Link "Button Class"
### Change Type

Addition

### Area

Customization

### Proposed Changes

I would like to propose that aside from the possible non-default values for "Button Class"/"button_class" of Custom Links, documentation is updated to include the information that the value displayed as "Default" in the Netbox UI is actually called "outline-dark". "Default" is as well included in CSV exports of Custom Links from already populated instances, which makes it harder for non-experienced users to determine what value to provide when mass-importing Custom Links with a button_class of "Default"(/"outline-dark").

I originally raised this in #15596 which hopefully provides enough reason why I think this should be included in the documentation.

maddin8:
…nt the value behind "Default" for the Custom Link "Button Class"

### Fixes: #15597 

Add documentation about available button classes for custom links to provide understanding especially for the case of bulk-imports.

### DIFF
--- a/docs/customization/custom-links.md
+++ b/docs/customization/custom-links.md
@@ -66,23 +66,23 @@ Button classes can be specified for custom links which will make the dropdown bu
 
 The following values are possible when bulk-importing custom links: 
 
-| Button class  | Value for bulk imports                                                                                                       |
-|-----------|-------------------------------------------------------------------------------------------------------------------|
-| `Default`  | `outline-dark`                                                                                 |
-| `Blue`   | `blue`                                                                 |
-| `Indigo` | `indigo`                                                                                          |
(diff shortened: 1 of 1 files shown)


jeremystretch:
This is not a valid resolution; please see my comment [here](https://github.com/netbox-community/netbox/issues/15597#issuecomment-2034375458). Please also note this is why the PR template contains the following warning:

> Please note that our contribution policy requires that a feature request or bug report **be approved and assigned prior to opening a pull request**. This helps avoid waste time and effort on a proposed change that we might not be able to accept.

PR #15623, opened 2024-04-04
Closes #14690: Pretty-format JSON fields in the config form

markkuleinio:
### Fixes: #14690

~~Adds `encoder` attribute in `ConfigParam` that is set to a custom JSON encoder if field is a `JSONField`.~~

Replaces the Django-original `JSONField` params with our own `JSONField`s, to get pretty JSON printing in the configuration editor.

### DIFF
--- a/netbox/core/forms/model_forms.py
+++ b/netbox/core/forms/model_forms.py
@@ -207,6 +207,9 @@ class ConfigRevisionForm(BootstrapMixin, forms.ModelForm, metaclass=ConfigFormMe
                 help_text += _(' (default)')
             self.fields[param.name].help_text = help_text
 
+            # Use the parameter-specific encoder
+            self.fields[param.name].encoder = param.encoder
+
     def save(self, commit=True):
         instance = super().save(commit=False)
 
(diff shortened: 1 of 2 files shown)


jeremystretch:
We actually have a custom [JSONField class](https://github.com/netbox-community/netbox/blob/5b50920c618d20a770ddfce217b974cefbd28212/netbox/utilities/forms/fields/fields.py#L92) that we use elsewhere to effect cleaner rendering, so ideally we'd reuse that for the relevant config parameters. However, importing the class directly raises an AppRegistryNotReady exception, because the class needs to be imported before the `utilities` app can be registered.

One option would be to swap out the field class in `ConfigFormMetaclass.__new__()`, changing `param.field` to our custom JSON field wherever it's set to Django's stock JSONField. It's a bit of a hack, but IMO reasonably clean given the constraints.

markkuleinio:
I needed to chew this a while 😁 Totally missed the custom `JSONField` earlier. Let's try again

jeremystretch (approved):

jeremystretch:
Thanks @markkuleinio!

PR #15624, opened 2024-04-04
Closes #14707: Change 'Interface' to 'Tunnel interface' in VPN tunnel forms

### LINKED ISSUE #14707: Change "Interface" to "Tunnel interface" in tunnel termination form
### NetBox version

v3.7.0

### Feature type

### Proposed functionality

I'm proposing changing the "**Interface**" field to "**Tunnel interface**" in the VPN tunnel termination form.

### Use case

The current terminology talking about "tunnel termination" and then asking for **interface** hints the user that this is the actual tunnel starting point (outside interface).

However, the real purpose of the field is to designate the **route-based tunnel interface** (inside interface), not the tunnel source/destination (outside) interface. This is somewhat confusing, especially when the next field is called "Outside IP" that definitely talks about the tunnel outside interface. (That's logical thinking in other cases for me at least: first you select interface, then you select the IP of that interface)

Example issue of the current confusion is #14681.

Renaming the interface field to "**Tunnel interface**" would guide the user to use the tunnel (inside) interface, not the outside interface.

### Database changes

Not sure if this naming change should involve changing some internal structures as well

### External dependencies

_No response_

markkuleinio:

### DIFF
--- a/netbox/vpn/forms/model_forms.py
+++ b/netbox/vpn/forms/model_forms.py
@@ -91,7 +91,7 @@ class TunnelCreateForm(TunnelForm):
     termination1_termination = DynamicModelChoiceField(
         queryset=Interface.objects.all(),
         required=False,
-        label=_('Interface'),
+        label=_('Tunnel interface'),
         query_params={
             'device_id': '$termination1_parent',
         }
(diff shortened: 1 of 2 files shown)


markkuleinio:
If I understand correctly, the translation entry is generated automatically in `django.po` at some point. Let me know if something is needed for that anyway.

jeremystretch (approved):

PR #15626, opened 2024-04-04
Closes #15039: Add Clone button in API token

### LINKED ISSUE #15039: Add possibility to clone API tokens
### NetBox version

v3.7.1

### Feature type

New functionality

### Proposed functionality

I'm proposing adding a **Clone** function in API tokens, the similar way the Clone button works for example in prefixes.

Desired functionality:
- Open an API token
- Click Clone (the new button proposed here)
- The resulted form is basically the same as in adding a new token, but:
    - User, Write enabled, Expires, Description and Allowed IPs fields are prepopulated with the previous values (user can change them if needed)
    - Key field is a new randomly generated key, just like when creating a new API token

### Use case

The use case is rotating API tokens. Currently it is not possible to edit the Key field of an API token, so you have to use the Add button when rotating the tokens, and this requires you to manually enter the identical/relevant User, Expires, Description and Allowed IPs fields to match the previous API token, most often using two browser windows/tabs.

Adding the cloning possibility removes the need to copy+paste the fields manually to the new token.

### Database changes

None

### External dependencies

None

markkuleinio:

### DIFF
--- a/netbox/users/models.py
+++ b/netbox/users/models.py
@@ -17,6 +17,7 @@ from netaddr import IPNetwork
 from core.models import ContentType
 from ipam.fields import IPNetworkField
 from netbox.config import get_config
+from netbox.models.features import CloningMixin
 from utilities.querysets import RestrictedQuerySet
 from utilities.utils import flatten_dict
 from .constants import *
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @markkuleinio!

PR #15659, opened 2024-04-08
Fixes: #14567 - Export current view of IP Addresses

### LINKED ISSUE #14567: Export current view of IP Addresses
### Deployment Type

Self-hosted

### NetBox Version

v3.6.7

### Python Version

3.8

### Steps to Reproduce

Tested with demo.netbox.dev
1. Select IPAM -> IP Addresses (on the top Results shows 183)
image
2.  Quick search 172.16  (Results stays on 183)
image
4. Export Current view 
image

### Expected Behavior

Shouldn't it be the case that when filtering by addresses, the number of results is reduced and accordingly the export of the current view only contains the filtered addresses?

### Observed Behavior

When exporting, all addresses are always exported.

Julio-Oliveira-Encora:
### Fixes: #14567 

It added a script in "table_controls_htmx.html" to add the parameters to the URL (changing the htmx behavior).

Another script was created in "export.html" to append the parameters to the HREF to export the current view. 

It was tested exporting devices, IP Address, and IP Range, with and without a quick search.


netbox/templates/inc/table_controls_htmx.html
```diff
@@ -2,22 +2,49 @@
...
+    function getParams() {
+        const quickSearchParameters = htmx.values(htmx.find("#quick_search"));
+        let link = document.getElementById("current_view")
+        if (quickSearchParameters != null) {
+            const search_parameter = `q=${quickSearchParameters.q}`
+            link = link.href + "&" + search_parameter
+        }
+        document.getElementById("current_view").setAttribute("href", link)
+    }
+</script>
```
jeremystretch, line 50: All scripting should be written in TypeScript and should live in a module under `project-static/src/` with the rest of the frontend scripting.

jeremystretch (changes requested):

netbox/project-static/src/search.ts
```diff
@@ -9,36 +10,59 @@ function quickSearchEventHandler(event: Event): void {
...
+ * @param event
+ */
+
+function handleQuickSearchParams(event: Event): void {
+  console.log('getParams');
+  const quickSearchParameters = event.currentTarget as HTMLInputElement;
+  const link = document.getElementById('current_view') as HTMLLinkElement;
+  if (quickSearchParameters != null) {
+    const search_parameter = `q=${quickSearchParameters.value}`;
+    const linkUpdated = link?.href + '&' + search_parameter;
```
jeremystretch, line 32: This appears to append successive queries, e.g. `?q=foo&q=bar`. Any previous query should be replaced by the new one.

jeremystretch (changes requested):
Getting closer!

netbox/project-static/src/search.ts
```diff
@@ -9,36 +10,59 @@ function quickSearchEventHandler(event: Event): void {
...
-    quicksearch.addEventListener("search", quickSearchEventHandler, {
-      passive: true
-    })
+    quicksearch.addEventListener('keyup', quickSearchEventHandler, {
+      passive: true,
+    });
+    quicksearch.addEventListener('search', quickSearchEventHandler, {
+      passive: true,
+    });
+    quicksearch.addEventListener('change', handleQuickSearchParams);
```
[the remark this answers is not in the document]
Julio-Oliveira-Encora: I created a function to clear the query parameters and used it in the "quickSearchEventHandler"; this way, when the Quick Search field is cleaned up, the parameters are also removed.

The same function is used when the search is changed, cleaning the current parameters before adding the new ones.

netbox/project-static/src/search.ts
```diff
@@ -25,20 +55,26 @@ export function initQuickSearch(): void {
...
+        async () => {
+          const search = new Event("search");
+          quicksearch.value = "";
+          await new Promise(f => setTimeout(f, 100));
+          quicksearch.dispatchEvent(search);
+        },
+        {
+          passive: true,
+        },
+      );
```
jeremystretch, line 77: Please revert any changes that aren't pertinent to the new functionality to limit the scope of PR reviews. (If cleanup is needed, we can tackle those items in a separate housekeeping issue.)

netbox/project-static/src/search.ts
```diff
@@ -11,12 +11,42 @@ function quickSearchEventHandler(event: Event): void {
...
+  const link = document.getElementById("current_view") as HTMLLinkElement;
+  let linkUpdated = link?.href.split("&")[0];
+  link.setAttribute("href", linkUpdated);
+}
+
+/**
+ * Update the Export View link to add the Quick Search parameters.
+ * @param event
+ */
+
```
jeremystretch, line 34:
```suggestion
```

netbox/project-static/src/search.ts
```diff
@@ -11,12 +11,42 @@ function quickSearchEventHandler(event: Event): void {
...
       clearbtn.classList.remove("invisible");
     }
   }
 }
 
+/**
+ * Clear the existing search parameters in the link to export Current View.
+ */
+function clearLinkParams(): void {
+  const link = document.getElementById("current_view") as HTMLLinkElement;
```
jeremystretch, line 25: Could we give this element a slightly more descriptive name, to help prevent confusion?
```suggestion
  const link = document.getElementById("export_current_view") as HTMLLinkElement;
```

jeremystretch (changes requested):

netbox/project-static/src/search.ts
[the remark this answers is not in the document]
jeremystretch: This does not appear to be working. After entering a search term in the quicksearch field and clicking the clear button, the term is still attached to the export link.

jeremystretch (changes requested):

arthanson (changes requested):
I'm still getting the whole table when I'm exporting.  See screenshot, I only have one Address that is in the table, but export I get all.
Monosnap IP Addresses | NetBox 2024-05-30 12-42-52

Julio-Oliveira-Encora:
I tested it. When I exported, I got only one IP address.

ip_addr
export_IP_addr

Julio-Oliveira-Encora:
I checked out to another branch and exported the current view. Then, I went back to the branch that fixed the export current view. Refreshing the page was necessary for it to work correctly.

jeremystretch (approved):

Issue #15671, opened 2024-04-09
Exception trying to add a custom script


peteeckel:
### Deployment Type

Self-hosted

### NetBox Version

v4.0-beta1

### Python Version

3.11

### Steps to Reproduce

1. Create a file with this content:
```python
from extras.scripts import Script


class Test(Script):

    class Meta:
        name = "Test"
        description = "This is a test script without any function"
        commit_default = True

    def run(self, data, commit):
        self.log_info("Script was executed")
```
2. Go to the NetBox GUI and navigate to 'Customization/Scripts/Add' to add the script
3. Go to 'Upload File/Choose File' and select the local file with the script for upload, then click 'Create'

### Expected Behavior

The script is added to the list of installed scripts. This works perfectly in v3.7.5.

### Observed Behavior

```
Traceback (most recent call last):
  File "/opt/netbox/lib64/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/lib64/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/lib64/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/netbox/views/generic/object_views.py", line 175, in dispatch
    return super().dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/netbox/views/generic/base.py", line 26, in dispatch
    return super().dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/utilities/views.py", line 108, in dispatch
    return super().dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #15675, opened 2024-04-09
15671 save module before sync_classes

arthanson:
### Fixes: #15671 

save Module before doing sync_classes as it can cause M2M error: `ValueError: save() prohibited to prevent data loss due to unsaved related object 'module'.`

### DIFF
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -165,9 +165,8 @@ class ScriptModule(PythonModuleMixin, JobsMixin, ManagedFile):
 
     def save(self, *args, **kwargs):
         self.file_root = ManagedFileRootPathChoices.SCRIPTS
-        obj = super().save(*args, **kwargs)
+        super().save(*args, **kwargs)
         self.sync_classes()
-        return obj
 
 
 @receiver(post_save, sender=ScriptModule)
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -165,8 +165,9 @@ class ScriptModule(PythonModuleMixin, JobsMixin, ManagedFile):
 
     def save(self, *args, **kwargs):
         self.file_root = ManagedFileRootPathChoices.SCRIPTS
+        obj = super().save(*args, **kwargs)
         self.sync_classes()
-        return super().save(*args, **kwargs)
+        return obj
 
 
 @receiver(post_save, sender=ScriptModule)


peteeckel:
Hi @arthanson, pardon me if this is a stupid question, but does `save()` really return a value? The Django docs say no, but is there any place in NetBox where this is overridden and the returned value is actually used?

```python
        self.file_root = ManagedFileRootPathChoices.SCRIPTS
        obj = super().save(*args, **kwargs)
        self.sync_classes()
        return obj
```

arthanson:
> Hi @arthanson, pardon me if this is a stupid question, but does `save()` really return a value? The Django docs say no, but is there any place in NetBox where this is overridden and the returned value is actually used?
> 
> ```python
>     def save(self, *args, **kwargs):
>         self.file_root = ManagedFileRootPathChoices.SCRIPTS
>         obj = super().save(*args, **kwargs)
>         self.sync_classes()
>         return obj
> ```

Good catch, fixed.

peteeckel:
There are some other places in the NetBox code where `save()` also returns a value, which made me think that this could be intentional after all.
```
# git grep -n 'return super().save' 
netbox/core/forms/model_forms.py:86:        return super().save(*args, **kwargs)
netbox/core/forms/model_forms.py:120:        return super().save(*args, **kwargs)
netbox/dcim/models/device_components.py:567:        return super().save(*args, **kwargs)
netbox/extras/models/change_logging.py:131:        return super().save(*args, **kwargs)
netbox/extras/models/customfields.py:784:        return super().save(*args, **kwargs)
netbox/extras/models/scripts.py:169:        return super().save(*args, **kwargs)
netbox/users/forms/bulk_import.py:34:        return super().save(*args, **kwargs)
netbox/users/forms/model_forms.py:98:        return super().save(*args, **kwargs)
netbox/users/models/tokens.py:92:        return super().save(*args, **kwargs)
netbox/vpn/forms/bulk_import.py:140:        return super().save(*args, **kwargs)
```
(in all likelihood not a complete list)

jeremystretch:
Django's `Model.save()` always returns None, so technically nothing needs to be returned. However, when invoking the method of a parent class using `super()`, it's common to always return its original output (if any) just out of convention. (Also, many of the instances above are actually forms, not models, and `ModelForm.save()` _does_ return an instance.)

jeremystretch (approved):

peteeckel:
> Django's `Model.save()` always returns None, so technically nothing needs to be returned. However, when invoking the method of a parent class using `super()`, it's common to always return its original output (if any) just out of convention.

I would totally agree with you (I love consistency) if the convention in NetBox wouldn't seem to be to **not** return the value of `super().save()`:

```
netbox/dcim/models/cables.py:        super().save(*args, **kwargs)
netbox/dcim/models/cables.py:        super().save(*args, **kwargs)
netbox/dcim/models/cables.py:        super().save(*args, **kwargs)
netbox/dcim/models/device_components.py:        return super().save(*args, **kwargs)
netbox/dcim/models/device_components.py:        super().save(*args, **kwargs)
netbox/dcim/models/devices.py:        ret = super().save(*args, **kwargs)
netbox/dcim/models/devices.py:        super().save(*args, **kwargs)
netbox/dcim/models/devices.py:        super().save(*args, **kwargs)
netbox/dcim/models/mixins.py:        super().save(*args, **kwargs)
netbox/dcim/models/power.py:        super().save(*args, **kwargs)
netbox/dcim/models/racks.py:        super().save(*args, **kwargs)
netbox/extras/models/change_logging.py:        return super().save(*args, **kwargs)
netbox/extras/models/customfields.py:        return super().save(*args, **kwargs)
netbox/extras/models/scripts.py:        super().save(*args, **kwargs)
netbox/ipam/models/ip.py:        super().save(*args, **kwargs)
netbox/ipam/models/ip.py:        super().save(*args, **kwargs)
netbox/ipam/models/ip.py:        super().save(*args, **kwargs)
netbox/netbox/models/features.py:        ret = super().save(*args, **kwargs)
netbox/users/models/tokens.py:        return super().save(*args, **kwargs)
netbox/virtualization/models/virtualmachines.py:        super().save(*args, **kwargs)
netbox/wireless/models.py:        super().save(*args, **kwargs)
```

That's 6 "return" vs. 15 "don't return" - for models, I was mistaken about the forms, sorry.

Issue #15541, opened 2024-03-26
Unable to Edit Component-Associations on InventoryItemTemplates


chadhpaine:
### Deployment Type

Self-hosted

### NetBox Version

v3.7.4

### Python Version

3.9

### Steps to Reproduce

1. Create a device-type.
2. Add an InventoryItemTemplate on the device-type created in step 1.
2a. Observe the lack of a 'Component' subsection in the Create Template view.
3. Observe that the table for the InventoryItemTemplates shows a 'Component' column.
3a. Observe that the value of the 'Component' column is empty.
5. Edit the InventoryItemTemplate created in step 2.
4a. Observe the lack of a 'Component' subsection in the Edit Template view.

### Expected Behavior

It should be possible to edit the 'Component' field of an InventoryItemTemplate from the UI.

### Observed Behavior

It doesn't appear that the 'Component' field is exposed for editing within the UI of the InventoryItemTemplate type.

arthanson:
This actually shows as non-editable fields on NB4
![Monosnap Editing inventory item template InvTemplate1 | NetBox 2024-03-26 15-16-38](https://github.com/netbox-community/netbox/assets/99642/e29b08ba-c576-439b-bfdf-b5293b2086af)

PR #15691, opened 2024-04-10
15541 Add component selector to InventoryItemTemplate

arthanson:
### Fixes: #15541 

Adds component assignment to InventoryItemTemplateForm to match InventoryItemForm - currently it just shows 
Monosnap Add a new inventory item template | NetBox 2024-04-10 13-24-31


jeremystretch:
@arthanson #15541 affects NetBox v3.7; shouldn't this PR target `develop`?

arthanson:
> @arthanson #15541 affects NetBox v3.7; shouldn't this PR target `develop`?

@jeremystretch because of the change to FieldSet this change if done in 3.7 would break 4 on merge and need to be re-worked (changing to use fieldset and removing items from form) So I thought it was better to just change / test it once by doing in 4?

jeremystretch:
It's a bug in v3.7 so it should be fixed there. We'll have to rework the solution for v4.0, but the same is true for a lot of changes made during the v4.0 beta.

jeremystretch (approved):

Issue #13649, opened 2023-08-31
Zero length cable should be treated as a 0 length cable and not an unknown length cable


DanSheps:
### NetBox version

v3.6.0

### Python version

3.8

### Steps to Reproduce

1. Create all prerequesite objects to have 2 interfaces to connect together (Manufacturer, Type, Role, Device, InterfaceA and InterfaceB)
2. Connect InterfaceC and InterfaceD together using 0 meters

### Expected Behavior

* The cable detail view should show a distance of 0 meters
* A cable trace should show a distance of 0 meters

### Observed Behavior

* Detail view shows placeholder data (--)
* Cable trace considers the distance an unknown distance

DanSheps:
I came across this while setting up splice boxes using rear ports and front ports.  When you set a "0" value for the length, the cable is treated as unknown value and the cable trace shows "approximately".

While there is a nominal (1µm perhaps) of distance added to a splice, it doesn't need to be taken into account, however the distance being accurate is important for reporting purposes (did we miss putting in a length or is it just the "0 length" item).

Underlying this is the API, which reports "None" instead of "0" for the CablePath.get_total_length() method.

jeremystretch:
I think the underlying issue here is that a cable should not be permitted to have a zero length.

DanSheps:
I think there are instances (like I said, when you aren't doing actual cables, but documenting "splices" where a 0-length is going to be desireable.  Until we start adding the ability to account for splicing, I think we should allow "0 length" cables and switch to checking for null instead of "not-empty (catches zero as a value)" in 3 places:

`dcim/models/cables.py:193`
`dcim/tables/template_code.py:15`
`templates/dcim/cable.html:53`

The API is fine as it returns "0.0 m"

jeremystretch:
> Until we start adding the ability to account for splicing, I think we should allow "0 length" cables

But a splice isn't a cable, and attempting to model it as one violates our tenet of replicating the real world. You're certainly welcome to submit a feature request proposing this ability, but as far as this bug goes IMO the solution is to invalidate zero-length cables.

jsenecal:
I wanted to chime in on this discussion, and I find myself in alignment with @DanSheps regarding the documentation of splice trays and their interconnections. This approach, using "0" and "None," brings clarity to network configurations involving devices with pigtails and splice ports. Let me elaborate further:

Device Configuration:
[...]

PR #15703, opened 2024-04-11
15685 Allow decimal for cable length filter form

### LINKED ISSUE #15685: dcim/cables/Filters/Attributes/Length > UI Field wrong Validation
### Deployment Type

Self-hosted

### NetBox Version

340fef3a5305 (v3.7.5)

### Python Version

3.11

### Steps to Reproduce

1. https://domain/dcim/cables/
2. Filters
3. Attributes
4. Length > it is not possible to filter for decimal lengths like 1.5 m Cables but they can be set as such.

### Expected Behavior

It should be possible to use decimal as not all cables come in full numbers and i doubt people drop to a lower metric to describe 150cm instead of 1.5m. Especially since you can set these values on the Cables themselve.

### Observed Behavior

image

arthanson:
### Fixes: #15685 

Cable length is decimal so allow filtering on decimal length.

### DIFF
--- a/netbox/dcim/forms/filtersets.py
+++ b/netbox/dcim/forms/filtersets.py
@@ -979,8 +979,6 @@ class CableFilterForm(TenancyFilterForm, NetBoxModelFilterSetForm):
     )
     length = forms.DecimalField(
         label=_('Length'),
-        required=False,
-        min_value=0.01
     )
     length_unit = forms.ChoiceField(
         label=_('Length unit'),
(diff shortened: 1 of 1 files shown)


netbox/dcim/forms/filtersets.py
```diff
@@ -977,9 +977,10 @@ class CableFilterForm(TenancyFilterForm, NetBoxModelFilterSetForm):
         label=_('Color'),
         required=False
     )
-    length = forms.IntegerField(
+    length = forms.DecimalField(
         label=_('Length'),
-        required=False
+        required=False,
+        min_value=0.01
```
jeremystretch, line 983: Per #13649 we allow cable lengths to be set to zero, so the filter should support this as well.
```suggestion
        required=False
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #15704, opened 2024-04-11
15582 check permissions on specific object when sync request

### LINKED ISSUE #15582: Permission constraint doesn't work on `sync` action
### Deployment Type

Self-hosted

### NetBox Version

v3.7.3

### Python Version

3.11

### Steps to Reproduce

1. Create a user permission with object type: `Core > Data Source`.  Allow `add`, `view`, and `sync` actions.  Add constraint: `{"id": 4}`.
2. Create an API token for this user
3. Sync a different data source: `curl -X POST -H "Authorization: Token $TOKEN" -H "Accept: application/json" http://localhost:8001/api/core/data-sources/7/sync/ `

### Additional Context
I'd like to sync this git data source it as a post-commit hook when I make changes to the scripts.  So I want to restrict this user permission to only one data source.

The constraint works properly on the `view` action.  If I GET `/api/core/data-sources/7/` I get a response: `{"detail":"Not found."}`.

### Expected Behavior

I should get a permission denied error.

### Observed Behavior

It works.

arthanson:
### Fixes: #15582 

Check permissions on specific object when sync request.  I kept the original permission check as that is before the object is loaded and the get_object_or_404 can return 404 if the object doesn't exist and the user doesn't have sync permission which is a security edge case of showing which ones exist.  

Once the object is actually loaded it needs to be re-checked in case there are constraints or perms tied to the specific object.

### DIFF
--- a/netbox/core/api/views.py
+++ b/netbox/core/api/views.py
@@ -37,6 +37,11 @@ class DataSourceViewSet(NetBoxModelViewSet):
             raise PermissionDenied("Syncing data sources requires the core.sync_datasource permission.")
 
         datasource = get_object_or_404(DataSource, pk=pk)
+
+        # have to check perms again against this specific object as there could be constraints
+        if not request.user.has_perm('core.sync_datasource', datasource):
+            raise PermissionDenied("User does not have the core.sync_datasource permission for this object.")
+
         datasource.enqueue_sync_job(request)
         serializer = serializers.DataSourceSerializer(datasource, context={'request': request})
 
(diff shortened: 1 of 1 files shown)


netbox/core/api/views.py
```diff
@@ -37,6 +37,11 @@ def sync(self, request, pk):
             raise PermissionDenied("Syncing data sources requires the core.sync_datasource permission.")
 
         datasource = get_object_or_404(DataSource, pk=pk)
+
+        # have to check perms again against this specific object as there could be constraints
+        if not request.user.has_perm('core.sync_datasource', datasource):
```
jeremystretch, line 42: The call to `request.user.has_perm()` above should be moved down and extended by specifying the object being synced (e.g. `has_perm('core.sync_datasource', obj=datasource)`) to properly utilize the permissions backend and avoid the need for a custom error message.

jeremystretch (changes requested):

PR #15705, opened 2024-04-11
#15548 - False Positive Related Object in Custom Fields

### LINKED ISSUE #15548: false positive related object in custom fields
### Deployment Type

Self-hosted

### NetBox Version

v3.7.4

### Python Version

3.10

### Steps to Reproduce

1. Create new custom field.
2. Try to delete the newly created custom field.
3. There is a message that the related object "CustomField_content_types object (n)" will also get deleted.

### Expected Behavior

Newly created custom field isn't bound to any object, so user should get simple delete confirmation dialog without related objects.

### Observed Behavior

User gets delete confirmation dialog with related objects even though those kinds of objects don't exist.

Another observation, don't know if it is this issue related:
During the upgrade procedure to the feature branch, I get the following error mentioning the customfield_content_types: 

```
File "/opt/netbox/venv/lib/python3.10/site-packages/psycopg/cursor.py", line 732, in execute
    raise ex.with_traceback(None)
django.db.utils.ProgrammingError: relation "extras_customfield_content_types_id_seq" does not exist

```

Julio-Oliveira-Encora:
### Fixes: #15548

I created the class CustomObjectDeleteView, which was inherited from ObjectDeleteView, and overwritten the "get" method and its dependent_objects variable. This way, the modal from delete_form.html  will not display messages about dependent objects.

Also, I changed the inherited class for CustomFieldDeleteView to use generic.CustomObjectDeleteView.

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -69,7 +69,7 @@ class CustomFieldEditView(generic.ObjectEditView):
 
 
 @register_model_view(CustomField, 'delete')
-class CustomFieldDeleteView(generic.ObjectDeleteView):
+class CustomFieldDeleteView(generic.CustomObjectDeleteView):
     queryset = CustomField.objects.select_related('choice_set')
 
 
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -31,6 +31,7 @@ __all__ = (
     'ObjectDeleteView',
     'ObjectEditView',
     'ObjectView',
+    'CustomObjectDeleteView',
 )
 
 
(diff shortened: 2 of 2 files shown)


jeremystretch:
Thanks for tackling this @Julio-Oliveira-Encora. I think there are a few issues with this approach.

First, we shouldn't need to introduce a custom class: ObjectDeleteView already has a `_get_dependent_objects()` method which we can override if needed to tweak which objects are returned.

Second, we can't blindly assume that _no_ dependent objects will exist. The bug pertains only to the custom field's association with content types; it's very likely that other dependencies may exist now or in the future.

Finally, we want to consider whether the reported behavior affects other models in NetBox too. For instance, should we also ignore content type associations when deleting a custom link or export template? Probably. We might even want to ignore many-to-many associations entirely, as the deletion of these associations doesn't actually impact the related objects.

jeremystretch (changes requested):
Thanks for tackling this @Julio-Oliveira-Encora. I think there are a few issues with this approach.

First, we shouldn't need to introduce a custom class: ObjectDeleteView already has a `_get_dependent_objects()` method which we can override if needed to tweak which objects are returned.

Second, we can't blindly assume that _no_ dependent objects will exist. The bug pertains only to the custom field's association with content types; it's very likely that other dependencies may exist now or in the future.

Finally, we want to consider whether the reported behavior affects other models in NetBox too. For instance, should we also ignore content type associations when deleting a custom link or export template? Probably. We might even want to ignore many-to-many associations entirely, as the deletion of these associations doesn't actually impact the related objects.

Issue #15637, opened 2024-04-05
Links within embedded tables do not load when HTMX is enabled


julianstolp:
### Deployment Type

Self-hosted

### NetBox Version

v4.0-beta1

### Python Version

3.10

### Steps to Reproduce

1. Go to user preferences and enable HTMX Navigation
2. Go to a site and attach an image
3. Click on the parent of that image, which should be the site itself
4. Opens the card `Site` and `Custom Fields` within `Images`

### Expected Behavior

It opens the site in the same tab like every other link.

### Observed Behavior

![error1](https://github.com/netbox-community/netbox/assets/41152397/3c85c7bb-d8c2-4324-a394-81e101828ad5)

julianstolp:
And in addition, the edit button does not work. The dropdown menu is also displayed in the card body instead of beyond the edge of the card.

![image](https://github.com/netbox-community/netbox/assets/41152397/039d5320-7cf9-40a8-9fb2-2ae712a32d71)

jeremystretch:
> The dropdown menu is also displayed in the card body instead of beyond the edge of the card.

Tagging #15536 as the same fix is likely needed here.

jeremystretch:
Changed the issue title as this seems to impact all embedded tables, not just image attachments.

PR #15724, opened 2024-04-15
Fixes 15652 - PROTECTION_RULES: Custom Validator does not show error message on object deletion

Julio-Oliveira-Encora:
### Fixes: #15652 

The issue was solved by replacing the line 433 of "netbox/netbox/views/generic/object_views.py" with "return redirect(self.get_return_url(request))".

**_Please let me know if I should also replace line 428 of the same file._**


jeremystretch (changes requested):
@Julio-Oliveira-Encora this change does not appear to address the bug. The returned URL is correct.

jeremystretch:
I'll take ownership of #15652 as I believe it arises from a regression introduced in the [fix](https://github.com/netbox-community/netbox/commit/d34f188d40b11d7d70d5d46cbc3f9accb2d876a8) for #15637.

PR #15734, opened 2024-04-16
15524 round iprange utilization

### LINKED ISSUE #15524: IP range utilization bar being rounded down
### Deployment Type

Self-hosted

### NetBox Version

v3.7.4

### Python Version

3.10

### Steps to Reproduce

The utilization bar for IP ranges is not behaving the same way as for prefixes. Decimal digits are apparently being rounded down to the next integer.

1. Create an IP range, let's say 1.1.1.1-254/24.
2. Create an IP in that range.

### Expected Behavior

Utilization shows 0.4% (and indeed this is shown for the **prefix** without the range and only the solitary IP).

2 IPs 0.8%
3 IPs 1.2%
etc.

### Observed Behavior

Utilization shows 0.0%.

2 IPs 0.0%
3 IPs 1.0%
etc.

arthanson:
### Fixes: #15524 

Change utilization for IPRange from truncating to rounding.

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -692,7 +692,7 @@ class IPRange(PrimaryModel):
             ip.address.ip for ip in self.get_child_ips()
         ]).size
 
-        return int(float(child_count) / self.size * 100)
+        return min(float(child_count) / self.size * 100, 100)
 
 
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @arthanson!

PR #15742, opened 2024-04-17
15427 Use endpoint_url in S3Backend

### LINKED ISSUE #15427: Data source for Amazon S3 being able to accept S3 compatible server
### NetBox version

v3.7.4

### Feature type

### Proposed functionality

Data sources of type `Amazon S3` do not work with compatible S3 servers. The proposal is to allow S3 compatible servers to be used as data sources.

Currently, there is a datasource of type "Amazon S3", that uses Amazon S3 in AWS. When creating the data source, the URL protocol and domain/netloc are not taken into account (except to obtain the region), so by default only AWS can be used.

The proposal is to add the argument `endpoint_url` to construct the boto3.resource, so S3 compatible servers can be used. The proposed change will be located in here: https://github.com/netbox-community/netbox/blob/4adb44f60d91fec19c800ea0e03800e2eb338f06/netbox/core/data_backends.py#L147

As a workaround, environment variables can be used (like `AWS_ENDPOINT_URL`), but that limits all S3 data sources to reside in that url.

### Use case

This feature will allow NetBox users to use other cloud providers or to host their own S3 compatible servers instead of using AWS to store their scripts and reports for `Amazon S3` datasources.

### Database changes

None

### External dependencies

None

javierdelapuente:
### Closes: #15427

This PR includes the `endpoint_url` parameter to the call to `boto3.resource` for the `S3Backend`. This allows to use external S3 servers. 

I have tested the code against AWS S3 and and also against a local server using Minio. These url formats work correctly:
`https://s3.amazonaws.com/<bucket-name>`
`https://s3.<region-name>.amazonaws.com/<bucket-name>`
`https://<your-s3-server>/<bucket-name>`

[Virtual-host-style](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#virtual-hosted-style-access) and s3 protocol (s3://<bucket-name>) do not work, but they were not working before this PR.

### DIFF
--- a/netbox/core/data_backends.py
+++ b/netbox/core/data_backends.py
@@ -149,7 +149,8 @@ class S3Backend(DataBackend):
             region_name=self._region_name,
             aws_access_key_id=aws_access_key_id,
             aws_secret_access_key=aws_secret_access_key,
-            config=self.config
+            config=self.config,
+            endpoint_url=self._endpoint_url
         )
         bucket = s3.Bucket(self._bucket_name)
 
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @javierdelapuente!

PR #15758, opened 2024-04-17
Closes #15736: Remove annotated_date() template filter and annotated_now() template tag

### LINKED ISSUE #15736: Remove the `annotated_date()` and `annotated_now()` template filters
### Proposed Changes

Remove the following template filters from `utilities.templatetags.helpers`:

- `annotated_date()`
- `annotated_now()`

### Justification

With the shift to using ISO 8601-formatted dates & times everywhere in #15735, these filters are no longer needed.

### Impact

All uses of these filters must be updated. These can generally be replaced with the new `isodatetime()` template filter.

jeremystretch:
### Closes: #15736

- Replace uses of `{% annotated_now %}` with Django's `{% now %}` tag (using ISO 8601 format)
- Remove the `annotated_date()` template filter
- Remove the `annotated_now()` template tag

### DIFF
--- a/netbox/templates/base/layout.html
+++ b/netbox/templates/base/layout.html
@@ -179,9 +179,9 @@ Blocks:
             {# /Footer links #}
 
             {# Footer text #}
-            <ul class="list-inline list-inline-dots mb-0" id="footer-stamp" hx-swap-oob="true">
+            <ul class="list-inline list-inline-dots fs-5 mb-0" id="footer-stamp" hx-swap-oob="true">
               <li class="list-inline-item">
-                {% annotated_now %} {% now 'T' %}
+                {% now 'Y-m-d H:i:s T' %}
               </li>
               <li class="list-inline-item">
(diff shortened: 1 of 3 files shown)


netbox/templates/base/layout.html
```diff
@@ -179,9 +179,9 @@ <h1 class="navbar-brand navbar-brand-autodark">
             {# /Footer links #}
 
             {# Footer text #}
-            <ul class="list-inline list-inline-dots mb-0" id="footer-stamp" hx-swap-oob="true">
+            <ul class="list-inline list-inline-dots fs-5 mb-0" id="footer-stamp" hx-swap-oob="true">
               <li class="list-inline-item">
-                {% annotated_now %} {% now 'T' %}
+                {% now 'Y-m-d H:i:s T' %}
```
arthanson, line 184: Will that format be a 'standard' potentially used elsewhere in the future? If so we should put it into a filter or constant.

arthanson (approved):

PR #15763, opened 2024-04-18
Fixes #15717 - Unable to assign a VM in Site to Cluster without Site

### LINKED ISSUE #15717: Unable to assign VM/Device in one site to cluster that has no site
### Deployment Type

Self-hosted

### NetBox Version

v3.7.5

### Python Version

3.11

### Steps to Reproduce

1. create vm/device and assign it to a site
2. create a cluster, don't assign any site
3. attempt to assign the created device/vm in step 1 to the cluster

Error message `The selected cluster (<cluster name>) is not assigned to this site (<site name>)`

### Expected Behavior

If a cluster has no site set, then a VM/device from any site should be assignable to it. Without being able to do this, it's not possible to track a cluster that spans multiple DC.

### Observed Behavior

Can't assign a device/vm with a defined site to a cluster with no defined site (`null`)

nopg:
### Fixes: #15717 

Adds to existing IF statement for VirtualMachine.clean()  (to match existing check for Device.clean()) to allow for VM to be assigned to a Cluster without a Site. 

Updated existing VM test to check for this new use case, and added Device tests for Clusters with Sites matching/not matching as well.

### DIFF
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -465,7 +465,10 @@ class DeviceForm(TenancyForm, NetBoxModelForm):
         label=_('Cluster'),
         queryset=Cluster.objects.all(),
         required=False,
-        selector=True
+        selector=True,
+        query_params={
+            'site_id': ['$site', 'null']
+        },
     )
     comments = CommentField()
     local_context_data = JSONField(
--- a/netbox/dcim/tests/test_models.py
+++ b/netbox/dcim/tests/test_models.py
@@ -560,9 +560,6 @@ class DeviceTestCase(TestCase):
         # Device with site, cluster non-site should pass
         Device(name='device1', site=sites[0], device_type=device_type, role=device_role, cluster=clusters[2]).full_clean()
 
-        # Device with non-site cluster only should pass
-        Device(name='device1', site=sites[0], device_type=device_type, role=device_role, cluster=clusters[2]).full_clean()
-
         # Device with mismatched site & cluster should fail
         with self.assertRaises(ValidationError):
             Device(name='device1', site=sites[0], device_type=device_type, role=device_role, cluster=clusters[1]).full_clean()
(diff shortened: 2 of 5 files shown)


arthanson:
@nopg can you please fix the merge conflicts

arthanson (changes requested):
@nopg I tried this PR - 

I go to edit a VM that has a site assigned and I can't get the unassigned cluster in the dropdown (see screenshot below) I have created a 'tcluster' that doesn't appear in the selection.  I think the queryset for the dropdown needs to be updated. 
Monosnap Editing virtual machine child vm | NetBox 2024-05-30 12-04-54

Note: without the PR I can assign a cluster (not assigned to a site) to a device that is assigned to a device fine, so I think this only applies to VM.

nopg:
I swear i checked that myself and it was working (at least back on 3.x)! But yes I see that issue as well. I've got a lot going on the next few days but plan on updating this as soon as I can, just wanted to reply here first.

nopg:
Ok, done. As you mentioned this doesn't apply to Devices just VM's. I removed the query_params from VM's/Cluster, which is how it looks on the Device side as well.

arthanson:
@nopg The dropdown works now but there is a validation error if you try to save
Monosnap Editing virtual machine vm9 | NetBox 2024-06-07 08-37-46

nopg:
@arthanson I can't reproduce that...I can save or create a VM-with-site into my Cluster without a site now, screenshot below. The validation should have been part of the initial change I made to virtualmachine.py in this PR.

vms

arthanson:
@nopg I see the issue, I actually had TC assigned to a different site, the problem is that the dropdown is now displaying Clusters in all sites so you can select an invalid cluster and get the error message above.  I have 3 clusters to test: TC is unassigned, TC2 is a different site then the VM and TC3 is the same site.  Only TC and TC3 should be showing in the dropdown (see screenshot). 
Monosnap Editing virtual machine vm9 | NetBox 2024-06-11 11-21-45

nopg:
Ok @arthanson, sounds good. I had initially noticed the behavior you are describing, but saw that this same behavior also already existed with Devices, so decided to just 'match' the Devices behavior and leave it be.

After looking into it though, I was able to produce the desired behavior (showing only same-site Clusters in the list, or Clusters not attached to any Site) with query_params, so I also updated the associated cluster query_params for Device as well as VirtualMachine. I assume at this point it's ok to include that in this PR. We are hopefully good to go at this point.

jeremystretch (approved):
Thanks @nopg!

PR #15766, opened 2024-04-18
15761 filter IKE Proposals on IKE Policy detail view

### LINKED ISSUE #15761: Display bug - all the IKE proposals are display in the IKE policy even if not selected
### Deployment Type

Self-hosted

### NetBox Version

3.7.5

### Python Version

3.11

### Steps to Reproduce

1. Create mutliple IKE proposals
2. Create a IKE-policy and select some IKE proposals
3. Go into the IKE-policies

And there you will see all the proposal event the one that are not selected

### Expected Behavior

Display only the proposals selected as on the first view http://127.0.0.1:8000/vpn/ike-policies/ 

### Observed Behavior

On the view http://127.0.0.1:8000/vpn/ike-policies/ everythings is ok
But when I select a specific policy, for ex on http://127.0.0.1:8000/vpn/ike-policies/1/ there is too much proposals

image

image

image

arthanson:
### Fixes: #15761 

IKEProposals were not getting correctly filtered on the IKEPolicy detail view because the matching filters were missing from the Filterset.

### DIFF
--- a/netbox/vpn/filtersets.py
+++ b/netbox/vpn/filtersets.py
@@ -136,6 +136,17 @@ class IKEProposalFilterSet(NetBoxModelFilterSet):
     group = django_filters.MultipleChoiceFilter(
         choices=DHGroupChoices
     )
+    ike_policy_id = django_filters.ModelMultipleChoiceFilter(
+        field_name='ike_policies',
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Don't forget to include a test when introducing a new filter.

PR #15772, opened 2024-04-19
resolve #15712 -  add imageattachments to virtualmachines

### LINKED ISSUE #15712: Add Image Attachments to Virtual Machines
### NetBox version

4.0

### Feature type

New functionality

### Proposed functionality

Extend virtual machines with the ability to add image attachments.

1. Inherit the ImageAttachmentsMixin in the  VirtualMachine model.
2. Add the required template to the object view for VMs

https://github.com/Hedius/netbox/tree/15712-add_imageattachments_to_virtualmachines

### Use case

I acknowledge that this feature (according to the FAQ) is or was initially meant for uploading photos of hardware and racks.

Our NetBox partly evolved into us uploading diagrams etc. into devices to extend the comment section with them.
(We document there what a host / server is doing etc.)
It would be a nice to have if we can do the same in virtual machines.
Main advantage is that you have those images directly linked to the host.

I do not know how other organizations use this feature, but having the same abilities for bare metal servers and virtual machines seems right to me.

Would this change be acceptable to you?

Adding the modal to VMs would not extend the complexity of NetBox.

### Database changes

None

### External dependencies

None

Hedius:
### Fixes: #15712

In order to resolve #15712 this commit adds the Image Attachment template to the virtualmachine template. Furthermore, the VirtualMachine model inherits the ImageAttachmentsMixin to add support for image attachmentes to the model.

Screenshot of the added modal:
image

image

### DIFF
--- a/netbox/templates/virtualization/virtualmachine.html
+++ b/netbox/templates/virtualization/virtualmachine.html
@@ -158,6 +158,7 @@
           </h5>
           {% htmx_table 'ipam:service_list' virtual_machine_id=object.pk %}
         </div>
+        {% include 'inc/panels/image_attachments.html' %}
         {% plugin_right_page object %}
     </div>
 </div>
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @Hedius!

Issue #15764, opened 2024-04-18
Increase maximum value for VC position on the Device model


sleepinggenius2:
### NetBox version

v3.7.5

### Feature type

Change to existing functionality

### Proposed functionality

The VC position on the Device model is currently limited to 255. My proposal would be to increase this to at least 65534, but it may need to be higher in other environments.

### Use case

I recently ran into an issue within our network due to the 255 limit on VC position. Cisco ASR9K devices allow specifying a satellite ID up to 65534, which seems like the exact use case for the VC model, a master chassis and a subtended one, but that means we want to set position 0 for the main router and use the satellite's ID for it's position in the VC, as that is used in provisioning and in its port naming. I was able to change the maximum value in the model within our NetBox instance and it didn't seem to cause any issues for a value < 32767, but 65534 would overflow a small int, as the type is currently configured, due to its signed nature.

Example from the router configuration:
```
#nv satellite ?
  <100-65534>  Satellite ID
```

### Database changes

The `vc_position` field on the Device model would need to be changed from a *smallint* to an *integer* type to support a value of 65534.

### External dependencies

None

Issue #7564, opened 2021-10-18
virtual-chassis position zero


PieterL75:
### NetBox version

v3.0.7

### Python version

3.7

### Steps to Reproduce

1. Create 2 devices
2. Create a new virtual chassis
3. Set the Initial position to zero
4. add the 2 devices
5. Create the Virtual Chassis

Next one, also related
1. Edit the virtual chassis
2. Set the posistions to 0 an 1
3. Save the VC
4. check the overview of devices in the VC

### Expected Behavior

A new VC is created with 2 members. with position 0 en 1.
The GUI shows the 0 and 1 position in a gray box

### Observed Behavior

The members are created with position 1 and 2.
![image](https://user-images.githubusercontent.com/74899468/137729793-9af7e016-45e8-4665-84e4-7e34b9ef8d0d.png)

After correcting: The position 0 is not displayed
![image](https://user-images.githubusercontent.com/74899468/137729634-9271bfcf-418a-4884-aeb6-cbfbc6bc712f.png)

PieterL75:
First issue seems to be related to this : 
https://github.com/netbox-community/netbox/blob/16f5e233d0d779a9bb65d32ee5f2f83a3467d5f3/netbox/dcim/forms/object_create.py#L125
This will cause 'initial_position' to become '1' if the get('initial_position') equals to {int}0
a=0
b=a or 1
b becomes 1
It can be fixed with
```    initial_position = self.cleaned_data.get('initial_position') if initial_position != None else 1```


The second can be solved with this : 
https://github.com/netbox-community/netbox/blob/a1f271d7d9cb1728a5f97379c3af10cae6f3b633/netbox/utilities/templatetags/helpers.py#L378
Change that to 
```        'value': str(value),```

PieterL75:
Junos has 0 for the first chassis id..

Ex
https://www.juniper.net/documentation/us/en/software/junos/virtual-chassis-mx/topics/ref/command/show-virtual-chassis-status-mx-series.html

Pieter


Op ma 18 okt. 2021 22:43 schreef Daniel Sheppard ***@***.***>:

jeremystretch:
Argh. Fixed in 3cb8c5db2 but I typoed the commit message.

PR #15779, opened 2024-04-19
15764 change vc_position from PositiveSmallInteger to PositiveInteger

arthanson:
### Fixes: #15764

Change vc_position from PositiveSmallInteger to PositiveInteger - form field and serializer are already IntegerField so no changes needed there. 

Note: We had talked about min-value of 1, but there was no min-value currently on the field and the serializer has min-value of 0 so it's possible someone already has 0 in the DB so didn't want to change.


jeremystretch (approved):

sleepinggenius2:
> Note: We had talked about min-value of 1, but there was no min-value currently on the field and the serializer has min-value of 0 so it's possible someone already has 0 in the DB so didn't want to change.

Just wanted to add this for completeness. We absolutely use position 0, as the ASR9K chassis in my example is referred to as Rack 0 within the equipment and uses 0 in its interface naming. An example for Junos was discussed in #7564 previously as well.

PR #15813, opened 2024-04-23
Fixes #11460 - Fix unterminated cable exception when editing cable

### LINKED ISSUE #11460: Editing an unterminated cable throws exception
### NetBox version

v3.4.0

### Python version

3.9

### Steps to Reproduce

1. Create Device-A with Interface
2. Create Device-B with Interface
3. Connect Device-A to Device-B via their interfaces.  Cable-1 is automatically created
4. Delete Device-A
5. Navigate to view for Cable-1 and hit the "Edit" button

### Expected Behavior

The "Edit" view for the cable is displayed

### Observed Behavior

<class 'AttributeError'>

DanSheps:
### Fixes: #11460 - Fix unterminated cable exception when editing cable

* Fixes the exception when a single cable is not present
* Fixes the exception when both cables aren't present (but does not allow user to proceed)


DanSheps:
There are other options here for this, right now, this is the path of least resistance without a re-work of the form rendering logic to only partially render specific fields.

Another caveat, for some reason `<button>` does not send in the param for a hx-get so it requires the get url to be specified.

netbox/dcim/views.py
```diff
@@ -3221,6 +3231,51 @@ def get_extra_addanother_params(self, request):
...
+        POST request handler.
+
+        Args:
+            request: The current request
+        """
+        # Set termination types if available
+        if request.GET.get('a_terminations_type'):
+            kwargs.update({'a_terminations_type': request.GET.get('a_terminations_type')})
+        if request.GET.get('b_terminations_type'):
+            kwargs.update({'b_terminations_type': request.GET.get('b_terminations_type')})
```
jeremystretch, line 3275: Same as above; we shouldn't need a custom `post()`.

jeremystretch (changes requested):
Nice work @DanSheps. However, as this is an open bug in v3.7 we need to fix it in the `develop` branch (which shouldn't require much change).

If you haven't already, look at how we handle type selection for remote data sources dynamically; we probably want to follow the same pattern here.

netbox/dcim/views.py
```diff
@@ -3191,7 +3194,7 @@ def dispatch(self, request, *args, **kwargs):
 
         return super().dispatch(request, *args, **kwargs)
 
-    def get_object(self, **kwargs):
+    def get_object(self, a_terminations_type=None, b_terminations_type=None, **kwargs):
```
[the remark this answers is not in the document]
DanSheps: If we move to alter_object, we likely don't need to adjust for signature either, will take a look when I have time.

netbox/dcim/views.py
```diff
@@ -3221,6 +3231,51 @@ def get_extra_addanother_params(self, request):
...
+        GET request handler.
+
+        Args:
+            request: The current request
+        """
+        # Set termination types if available
+        if request.GET.get('a_terminations_type'):
+            kwargs.update({'a_terminations_type': request.GET.get('a_terminations_type')})
+        if request.GET.get('b_terminations_type'):
+            kwargs.update({'b_terminations_type': request.GET.get('b_terminations_type')})
```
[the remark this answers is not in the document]
DanSheps: Unfortunately we still need it to handle the htmx portion, but...  I think I have a way around that too.

DanSheps:
Updated to develop but I also switched around the rending a bit.  Instead of using the hacky method I was using and pulling down the full page as an HTMX and then selecting I did the following:

1. Modify `netbox/views/generic/object_views.py` to have an `htmx_template_name` attribute for the ObjectEditView.
2. Defaulted that to the regular htmx form (`htmx/form.html`)
3. Overrode the `htmx_template_name` in the CableEditView
4. Broke the form (`dcim/cable_edit.html`) out into an new file in a new "htmx" folder (form is at `dcim/htmx/cable_edit.html`)
5. Changed the htmx around slightly

arthanson:
@DanSheps the edit form comes up, but if you just select save without any changes an exception is thrown:

```
 File "/Users/ahanson/dev/work/netbox/netbox/utilities/forms/mixins.py", line 55, in is_valid
    is_valid = super().is_valid()
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.10/site-packages/django/forms/forms.py", line 201, in is_valid
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.10/site-packages/django/forms/forms.py", line 196, in errors
    self.full_clean()
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.10/site-packages/django/forms/forms.py", line 434, in full_clean
    self._clean_form()
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.10/site-packages/django/forms/forms.py", line 455, in _clean_form
    cleaned_data = self.clean()
  File "/Users/ahanson/dev/work/netbox/netbox/dcim/forms/connections.py", line 103, in clean
    self.instance.a_terminations = self.cleaned_data['a_terminations']
KeyError: 'a_terminations'
```

DanSheps:
Thanks Arthur, didn't think to check the next step.  Will try and figure out something for that.

arthanson:
@jeremystretch looks like your changes fixed the issue I had previously

arthanson:
One weirdness - go to edit a cable and click the delete 'x' next to the items in the "B Side", when all are deleted the form auto-refreshes and all the items are back.

netbox/dcim/forms/connections.py
```diff
@@ -82,14 +83,22 @@ def __new__(mcs, name, bases, attrs):
 
     class _CableForm(CableForm, metaclass=FormMetaclass):
 
-        def __init__(self, *args, **kwargs):
+        def __init__(self, *args, initial=None, **kwargs):
+
+            initial = initial or {}
```
arthanson, line 88: Would it work to just do initial={} in the function def?

jeremystretch: It would, but we generally want to [avoid passing mutable objects as arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

jeremystretch:
This implementation isn't quite perfect, but it does at least resolve #11460 without (AFAICT) breaking any existing functionality. I'm going to merge this as-is to resolve the immediate issues, and we can iterate on improvements separately from there.

jeremystretch (approved):

PR #15822, opened 2024-04-23
Feature #15428 - Show all devices with configuration template attached

### LINKED ISSUE #15428: Show all devices with configuration template attached
### NetBox version

v3.7.3

### Feature type

New functionality

### Proposed functionality

Under Provision -> Configuration Template the list of templates should include information about the amount of devices that have the specific template attached. And the option to click on the number to get a search result with the devices where the template is attached.

The same feature as under Devices -> Device Roles. The list contains a count of the devices with this specific role.
The same feature is available for Device Types, where there is an Instance Counter.

### Use case

It will make it easy to figure out what devices have a specific Configuration Template attached, and show if the template is not used at all.

### Database changes

_No response_

### External dependencies

_No response_

Julio-Oliveira-Encora:
### Closes: #15428 

I added a Device column for Config Template to display the devices using a specific Template.

### DIFF
--- a/netbox/extras/tables/tables.py
+++ b/netbox/extras/tables/tables.py
@@ -417,7 +417,7 @@ class ConfigTemplateTable(NetBoxTable):
     instance_count = columns.LinkedCountColumn(
         viewname='dcim:device_list',
         url_params={'config_template_id': 'pk'},
-        verbose_name=_('Instances')
+        verbose_name=_('Devices')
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @Julio-Oliveira-Encora! I added columns for VMs, platforms, and device roles too for consistency.

Issue #15815, opened 2024-04-23
Server error with certain dashboard widgets after upgrading to v4.0


julianstolp:
### Deployment Type

Self-hosted

### NetBox Version

v4.0-beta2

### Python Version

3.10

### Steps to Reproduce

1. Add Widget `Object Count` with Models `Users > Group`, `Users > User` in v3.7.5
2. Upgrade to v4.0-beta2
3. Restart NetBox
4. Go to https://netbox-url/

Most likley caused by https://github.com/netbox-community/netbox/pull/15304

### Expected Behavior

Dashboard view loads as usual.

### Observed Behavior

Server Error

```
Environment:


Request Method: GET
Request URL: https://netbox-url/

Django Version: 5.0.4
Python Version: 3.10.12
Installed Applications:
['django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.humanize',
 'django.forms',
 'corsheaders',
 'debug_toolbar',
 'django_filters',
 'django_htmx',
 'django_tables2',
 'django_prometheus',
 'strawberry_django',
 'mptt',
 'rest_framework',
 'social_django',
 'taggit',
 'timezone_field',
 'core',
 'account',
 'circuits',
 'dcim',
 'ipam',
 'extras',
 'tenancy',
 'users',
 'utilities',
 'virtualization',
 'vpn',
 'wireless',
 'django_rq',
 'drf_spectacular',
 'drf_spectacular_sidecar']
Installed Middleware:
['strawberry_django.middlewares.debug_toolbar.DebugToolbarMiddleware',
 'django_prometheus.middleware.PrometheusBeforeMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django_htmx.middleware.HtmxMiddleware',
 'netbox.middleware.RemoteUserMiddleware',
 'netbox.middleware.CoreMiddleware',
[...]

PR #15839, opened 2024-04-24
15815 convert dashboard widgets for users/groups

arthanson:
### Fixes: #15815 

Convert dashboard widgets using `users.netboxuser` -> `users.user` and `users.netboxgroup` -> `users.group`


netbox/extras/migrations/0115_convert_dashboard_widgets.py
```diff
@@ -0,0 +1,31 @@
...
+
+from django.db import migrations
+
+
+def update_dashboard_widgets(apps, schema_editor):
+    Dashboard = apps.get_model('extras', 'Dashboard')
+
+    for dashboard in Dashboard.objects.all():
+        for key, widget in dashboard.config.items():
+            breakpoint()
```
jeremystretch, line 11:
```suggestion
```

netbox/extras/migrations/0115_convert_dashboard_widgets.py
```diff
@@ -0,0 +1,31 @@
...
+            if 'models' in widget['config']:
+                models = widget['config']['models']
```
jeremystretch, line 13:
```suggestion
            if models := widget['config'].get('models'):
```

jeremystretch (changes requested):
It feels like we should also take this opportunity to improve error handling in `extras.dashboard.widgets.get_models_from_content_types()`. IMO it's reasonably to fail silently when attempting to resolve a content type that doesn't exist.

netbox/extras/migrations/0115_convert_dashboard_widgets.py
arthanson: doh!

jeremystretch (approved):
Thanks @arthanson!

Issue #15826, opened 2024-04-24
EXEMPT_EXCLUDE_MODELS contains old user and group models


tobiasge:
### Deployment Type

Self-hosted

### NetBox Version

v4.0-beta2 (e05ca710ae47db49f8552a77b996ab4f69abaedc)

### Python Version

3.11

### Steps to Reproduce

1. None; `EXEMPT_EXCLUDE_MODELS` is a hard coded setting

### Expected Behavior

`EXEMPT_EXCLUDE_MODELS` has the new user and group model

### Observed Behavior

`EXEMPT_EXCLUDE_MODELS` contains old models

arthanson:
I'm not certain if those still need to be there, can probably be removed but needs to be checked.

PR #15827, opened 2024-04-24
Fix #15826: Added new group and user models

tobiasge:
### Fixes: #15826

Added new group and user models; sorted list alphabetically.

PR #15840, opened 2024-04-24
15826 update EXEMPT_EXCLUDE_MODELS with new user/group models

arthanson:
### Fixes: #15826 

Updated EXEMPT_EXCLUDE_MODELS for new user/group model.  @jeremystretch is it correct to remove these?  I tried switching them but it makes several tests fail.


tobiasge:
@arthanson I had already created a PR for this, that also fixes the tests. See #15827

jeremystretch:
Closing this in favor of #15827. The model references need to be renamed but they still need to be included in `EXEMPT_EXCLUDE_MODELS`.

Issue #15802, opened 2024-04-22
Dark mode anchor contrast to low


alehaa:
### Deployment Type

Self-hosted

### NetBox Version

v4.0-beta1

### Python Version

3.11

### Steps to Reproduce

1. Activate dark mode
2. Navigate to Site list
3. Inspect first columns site name with the browser debug toolbar
4. Check text color and background color attributes
5. Calculate contrast ratio.

### Expected Behavior

According to WCAG 2.1 accessibility standard, a bare minimum contrast ratio of 3:1 is required, 4.5:1 would be better.

### Observed Behavior

A ratio of 2.92:1 was calculated.

Responsible for the link color is the CSS variable `--tblr-link-color-rgb`. I found that the [Tabler Demo](https://preview.tabler.io/tables.html) uses slightly different colors. However, I can't find the cause why, as [netbox.scss](https://github.com/netbox-community/netbox/blob/v4.0-beta1/netbox/project-static/styles/netbox.scss) doesn't seem to change related Tabler variables.

jeffgdotorg:
Thanks for your report. We're aware that 4.0-beta1 has many low visual contrast problems and we hope to address them before the final.

PR #15841, opened 2024-04-24
15802 change table anchor color

arthanson:
### Fixes: #15802 

Made a lighter version of the current link color for tables

Monosnap Devices | NetBox 2024-04-29 10-22-58


jeremystretch (changes requested):
> However it doesn't have any contrast between a link and a non-link (see screenshot) not sure if we want this or to go with a different tabler link color...

We definitely still want to render links in a different color. #15802 is just to increase the contrast of links under dark mode.

jeremystretch (changes requested):
The color change should only apply to dark mode.

screenshot

arthanson:
I tried using different media queries but they didn't work properly when switching between light and dark, even specifically defining media queries for both.  As a compromise I picked a filter to slightly lighten the link which seems to look fairly good in both light and dark.

netbox/project-static/styles/overrides/_tabler.scss
```diff
@@ -27,3 +27,7 @@ pre {
   // Tabler sets display: flex
   display: inline-block;;
 }
+
+table a {
+  filter: brightness(110%);
```
jeremystretch, line 32: Please add a comment explaining why the rule exists.

jeremystretch (changes requested):

jeremystretch (approved):

Issue #15348, opened 2024-03-05
Quick access to SavedFilters


PieterL75:
### NetBox version

v3.7.3

### Feature type

Change to existing functionality

### Proposed functionality

The Saved Filters #9623 are a nice feature. But I would to a more intuitive access in the UI to that
- Show the saved filters in a dropdown on the model page
![image](https://github.com/netbox-community/netbox/assets/74899468/76bb3497-a7b1-4e4b-9671-06a7f4d12211)

### Use case

In orde to access the saved filter, I have to
- open de model
- click 'filter'
- select the filter from the dropdown
- scroll all the way down
- click search

Having the saved filters on the model search page will make it a single-click action

jeremystretch:
I've opened this for an owner, with the caveat that any work be based on the `feature` branch due to the UI overhaul for the upcoming v4.0 release.

Julio-Oliveira-Encora:
Could you please assign it to me?

Julio-Oliveira-Encora:
@jeremystretch Should I work on it using the v4?

jeremystretch:
Thanks for all your work on this @Julio-Oliveira-Encora!

![screenshot](https://github.com/netbox-community/netbox/assets/13487278/97a801ce-55d1-463f-b992-9da14bb5e745)

PR #15862, opened 2024-04-26
Feature 15348 - Quick Access Saved Filters

Julio-Oliveira-Encora:
### Feature: #15348

A dropdown was added to list and apply the saved filters to the current view.


netbox/templates/inc/table_controls_htmx.html
```diff
@@ -4,22 +4,34 @@
 <div class="row mb-3">
   <div class="col-auto d-print-none">
     <div class="input-group input-group-flat me-2 quicksearch" hx-disinherit="hx-select hx-swap">
-      <input type="search" results="5" name="q" id="quicksearch" class="form-control px-2 py-1" placeholder="Quick search"
-        hx-get="{{ request.full_path }}" hx-target="#object_list" hx-trigger="keyup changed delay:500ms, search" />
+      <input type="search" results="5" name="q" id="quicksearch" class="form-control px-2 py-1"
```
jeremystretch, line 7: We need to figure out how to make the new dropdown visually consistent with the quick search box.

screenshot

netbox/project-static/src/savedFiltersSelect.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Handle saved filter change event.
+ *
+ * @param event "change" event for the saved filter select
+ */
+function handleSavedFilterChange(event: Event): void {
+  const savedFilter = event.currentTarget as HTMLSelectElement;
+  let baseUrl = savedFilter.baseURI.split("?")[0];
+  let preFilter = "?";
```
jeremystretch, line 11:
```suggestion
  const preFilter = '?';
```

netbox/project-static/src/savedFiltersSelect.ts
```diff
@@ -0,0 +1,27 @@
+import {isTruthy} from "./util";
```
jeremystretch, line 1: This and some other lines within the file trigger complaints from `prettier`.

```suggestion
import { isTruthy } from './util';
```

netbox/project-static/src/savedFiltersSelect.ts
jeremystretch, line 1: We should probably move this under `forms/`.

jeremystretch (changes requested):
Although functional, displaying the active saved filters in both the original tag form and in the selector widget may be confusing to the user. IMO we should display applied filters only in the selector widget, and show only individual filter values as removable colored tags.

screenshot

arthanson:
@Julio-Oliveira-Encora here are the changes after discussion on this:

* Remove the blue chits for the saved filters only, so normal filters will still show there (The functionality for saving your filters and such should remain).  So the selected saved filters will only show in the save filters dropdown.
* We don't need to change the saved filters to use a placeholder, having the label there is fine.
* Make the vertical height of the dropdown match the Quick Search so at start they aren't different sizes.  You can adjust the text height and padding of the Saved filters, you may also need to tweak the quick search if needed.  Note: If you select many saved filters it will wrap and grow the select box, this is fine - we just care about the initial page look when you have no filters (or one filter) selected.

Julio-Oliveira-Encora:
@arthanson I changed the CSS for quick search, because I did get to change the CSS for Dropdown. Please, check them.

arthanson (changes requested):
@Julio-Oliveira-Encora I added a saved filter and then selected it, however it didn't actually filter the table - see screenshots.  The first is the filter applied manually, the second is selecting it from the saved filter list, they number of results should match but don't.
Monosnap IP Addresses | NetBox 2024-05-28 09-19-39
Monosnap IP Addresses | NetBox 2024-05-28 09-20-03

arthanson:
@Julio-Oliveira-Encora I just re-tested this, it is happening when I don't have HTMX navigation enabled

Julio-Oliveira-Encora:
I tested with HTMX Navigation disabled:

Preferences


For IP Address I created a "Saved Filter" for IPV4 and Active. It worked:

ip_saved_filter
ip_filters


For sites I tested too:

sites
Screenshot from 2024-06-03 12-17-17

arthanson:
@Julio-Oliveira-Encora almost there - can you remove the saved-filters from the filters tab. With both the dropdown and it being in filters tab you can get the UI showing both which isn't great:
Monosnap Devices | NetBox 2024-06-07 08-47-39

netbox/templates/inc/table_controls_htmx.html
```diff
@@ -1,25 +1,37 @@
...
       <span class="input-group-text py-1">
         <a href="#" id="quicksearch_clear" class="invisible text-secondary"><i class="mdi mdi-close-circle"></i></a>
       </span>
-    {% block extra_table_controls %}{% endblock %}
+      {% block extra_table_controls %}{% endblock %}
     </div>
   </div>
+
+  <div class="col-auto d-print-none">
+    <label class="my-2" for="id_filter_id">Saved Filter</label>
```
arthanson, line 18: Also put {% trans around this

netbox/project-static/src/appliedFilter.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ */
+export function initAppliedFilter(): void {
+  const appliedFilter = document.getElementById('appliedfilters');
+  if (isTruthy(appliedFilter)) {
+    const divResults = document.getElementById('results');
+    if (isTruthy(divResults)) {
+      const savedFilterSelect = divResults.getElementsByTagName('select')[0];
+      if (savedFilterSelect.selectedOptions.length > 0) {
+        appliedFilter.hidden = true;
```
jeremystretch, line 14: We should not rely on Javascript to hide this element; rather it should simply not be rendered in the template.

@arthanson can you help Julio with this please?

jeremystretch (changes requested):

arthanson:
@Julio-Oliveira-Encora if you need any help with hiding the column in the table, please let me know.

Julio-Oliveira-Encora:
@arthanson  and @jeremystretch  
Could the function `applied_filters` be modified to check if the form only contains the  `filter_id` and return an empty `applied_filters` to avoid using Javascript?

jeremystretch (approved):
I've omitted saved filters from the "applied filters" list and did some cosmetic cleanup of the selector widget.

Julio-Oliveira-Encora:
@jeremystretch Should I make some additional changes to it?

arthanson (approved):

Issue #15496, opened 2024-03-21
Add a list view for circuit terminations


PieterL75:
### NetBox version

v3.7.3

### Feature type

Data model extension

### Proposed functionality

The model Circuit-terminations has no view in the GUI. 
Can a table view be added ? 
https://netbox/circuits/circuit-terminations/

### Use case

Review the content of circuit-terminations

PR #15876, opened 2024-04-28
15794 Make "related objects" dynamic

### LINKED ISSUE #15794: Make the "Related Objects" pane of all models dynamic
### NetBox version

v3.7.5

### Feature type

### Proposed functionality

In #13794, extending the "Related Objects" by other showing other models was requested and implemented by dynamically listing counts for all `ManyToMany` reverse relationships. Now, I'd like to copy this functionality to all other uses of the "Related Objects" pane,

### Use case

As described in #13794, it might be useful for plugins to link their data to existing NetBox core models. However, this is not only useful for tenancy, but any object relationship could benefit from this. 

A real life example is covered by the [NetBox inventory](https://github.com/ArnesSI/netbox-inventory) plugin: Hardware assets are related to `dcim.Device`, `dcim.Manufacturer`, `dcim.Site` and `dcim.Location`. They would benefit from quick access to a list of related assets for these models. Right now an extra pane needs to be implemented, so there are actually two "Related Objects" panes per details page.

Also, this would make this feature work consistently across all views in NetBox, so the maintenance overhead should theoretically be reduced.

### Database changes

None

### External dependencies

None

alehaa:
### Fixes: #15794

Instead of hardcoding relationships between models for the detail view, they are now dynamically generated.

#### Known limitations

1. If a related model doesn't have a `list` view, rendering of the [related objects pane](https://github.com/netbox-community/netbox/blob/85db007ff585370b03b2d88d8f43e6fa9b94bc8e/netbox/templates/inc/panels/related_objects.html#L8) will fail. This limitation was already present before, but related models listed were explicitly chosen to do so. For all views shipped with NetBox I've added a filter if needed (`omit` parameter). However, external plugins might trigger this unintentionally, when named view is not available, but the relationship has not been setup with `related_name='+'` parameter. Switching to `verified_view` in the panel's template should resolve this, but might decrease performance slightly.

2. If a model introduces several relations to another model, they will all be listed with the same model name but different URL query parameters. This is not the case for models included in NetBox. As before, plugin developers can still use an extra panel for this special purpose and extra context about the different types of relationships.


netbox/utilities/views.py
```diff
@@ -140,6 +144,51 @@ def get_return_url(self, request, obj=None):
         return reverse('home')
 
 
+class GetRelatedModelsMixin:
+    """
+    Provides logic for collecting all related models for the currently viewed model.
+    """
+
+    def get_extra_related_models(self, request, instance):
```
jeremystretch, line 152: What is the purpose of this method? It seems like maybe a hook for monkey patching, which IMO should be discouraged. If needed, we could potentially extend the underlying `get_related_models()` utility function (not the class method) to collect additional types of relationships.

Or, if it's intended to be overridden naturally, we could introduce an `extra` keyword argument to the `get_related_models()` method to pass these.

alehaa: You are right, its intended to be used as a hook by view classes, e.g. when not only direct relations should be collected.

Although I prefer a generic approach that implements some logic in the utility function, I don't know how we should determine which transient relations are interesting for object views. So for now, I'd remove this method and add an `extra` keyword argument to the class method instead.

netbox/circuits/views.py
```diff
@@ -156,19 +147,22 @@ class ProviderNetworkListView(generic.ObjectListView):
...
-                'provider_network_id',
-            ),
-        )
-
         return {
-            'related_models': related_models,
+            'related_models': self.get_related_models(
+                request,
+                instance,
+                [CircuitTermination],
```
jeremystretch, line 158: We should be able to include circuit terminations now that #15496 has been completed.

netbox/dcim/views.py
```diff
@@ -380,31 +386,23 @@ class SiteListView(generic.ObjectListView):
...
-            (Circuit.objects.restrict(request.user, 'view').filter(terminations__site=instance).distinct(), 'site_id'),
-        )
-
         return {
-            'related_models': related_models,
+            'related_models': self.get_related_models(
+                request,
+                instance,
+                [CableTermination, CircuitTermination],
+                (
```
jeremystretch, line 398: Looks like we need to pull in ASNs too.

netbox/dcim/views.py
```diff
@@ -912,16 +889,16 @@ class DeviceTypeListView(generic.ObjectListView):
...
     queryset = DeviceType.objects.all()
 
     def get_extra_context(self, request, instance):
-        related_models = (
-            (Device.objects.restrict(request.user).filter(device_type=instance), 'device_type_id'),
-        )
-
         return {
-            'related_models': related_models,
+            'related_models': self.get_related_models(request, instance, [
```
jeremystretch, line 897: Including the keyword improves clarity.
```suggestion
            'related_models': self.get_related_models(request, instance, omit=[
```

netbox/dcim/views.py
```diff
@@ -1151,16 +1128,16 @@ class ModuleTypeListView(generic.ObjectListView):
...
     queryset = ModuleType.objects.all()
 
     def get_extra_context(self, request, instance):
-        related_models = (
-            (Module.objects.restrict(request.user).filter(module_type=instance), 'module_type_id'),
-        )
-
         return {
-            'related_models': related_models,
+            'related_models': self.get_related_models(request, instance, [
```
jeremystretch, line 1136:
```suggestion
            'related_models': self.get_related_models(request, instance, omit=[
```

netbox/ipam/views.py
```diff
@@ -52,7 +47,7 @@ def get_extra_context(self, request, instance):
         )
 
         return {
-            'related_models': related_models,
+            'related_models': self.get_related_models(request, instance),
```
jeremystretch, line 50: This returns two entries for "Interfaces," because we're querying both device and VM interfaces. The simplest solution would be to exclude both, which IMO is reasonable.

netbox/ipam/views.py
```diff
@@ -272,17 +263,12 @@ class ASNListView(generic.ObjectListView):
...
 
     def get_extra_context(self, request, instance):
-        related_models = (
-            (Site.objects.restrict(request.user, 'view').filter(asns__in=[instance]), 'asn_id'),
-            (Provider.objects.restrict(request.user, 'view').filter(asns__in=[instance]), 'asn_id'),
-        )
-
         return {
-            'related_models': related_models,
+            'related_models': self.get_related_models(request, [instance]),
```
jeremystretch, line 271: Site and Provider need to be pulled explicitly as they are both M2M relationships.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @alehaa!

Issue #15832, opened 2024-04-24
Multiselect has no 'delete' option on the values


PieterL75:
### NetBox version

v4.0-beta

### Feature type

Data model extension

### Proposed functionality

Re-add the X on the multiselect. It got dropped in the UI rehaul 

![image](https://github.com/netbox-community/netbox/assets/74899468/d20b7381-871e-4da4-87e6-f11bcf1bac6c)

vs 

![image](https://github.com/netbox-community/netbox/assets/74899468/4e92060e-eabd-4240-8282-f8f6d802058f)

### Use case

remove a value from the multiselect, without using the keyboard

arthanson:
Looks like this optional from the remove_button plugin (https://tom-select.js.org/plugins/remove-button/):

```
plugins: ['remove_button'],
```

Julio-Oliveira-Encora:
Could you please add it to me?

PR #15883, opened 2024-04-29
Feature 15832 -  Multiselect has no "delete" option on the values

Julio-Oliveira-Encora:
### Feature: #15832

Added "remove_button" plugin in config.ts.


jeremystretch (changes requested):
This works well for multi-select fields (such as tag assignment), but we end up with duplicate "clear" buttons" for single selection fields:

screenshot

IMO it would be ideal to add the new button on individual items only for multi-select fields, and leave the "clear all" button in place for other fields. Alternatively, we could remove the current "clear all" button that appears on hover and use only the individual button introduced by the plugin, however that feels less natural to me.

This should be doable by enabling the plugin under `initDynamicSelects()` and `initStaticSelects()` depending on the field type, rather than globally for all selects.

jeremystretch (approved):

PR #15886, opened 2024-04-29
Added assigned_object_type in prefetch for api view IPAddressViewSet

### LINKED ISSUE #15845: N+1 issue on the /api/ipam/ip-addresses/ endpoint
### Deployment Type

Self-hosted

### NetBox Version

v3.7.6

### Python Version

3.10

### Steps to Reproduce

Call the  `/api/ipam/ip-addresses/` endpoint with more than 1 assigned ip address in the database (important that the ip addresses are assigned and you are listing assigned ip addresses, for example by using `assigned_to_interface=true` filter)

### Expected Behavior

No N+1 issue.

I can submit I PR if wanted (I think the issue is that `IPAddressViewSet` are missing `assigned_object_type` in the `prefetch_related`)

### Observed Behavior

N+1 issue with the following query:

`SELECT ••• FROM "django_content_type" WHERE "django_content_type"."id" = X LIMIT 21`

mulmat:
### Fixes: #15845

Adds `assigned_object_type` to prefetch in order to avoid N+1 issue in `/api/ipam/ip-addresses/`

### DIFF
--- a/netbox/ipam/api/views.py
+++ b/netbox/ipam/api/views.py
@@ -119,7 +119,7 @@ class IPRangeViewSet(NetBoxModelViewSet):
 
 class IPAddressViewSet(NetBoxModelViewSet):
     queryset = IPAddress.objects.prefetch_related(
-        'vrf__tenant', 'tenant', 'nat_inside', 'nat_outside', 'tags', 'assigned_object'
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):

jeremystretch:
Thanks @mulmat!

PR #15897, opened 2024-04-30
Added caching on /api/schema/ endpoint

### LINKED ISSUE #15894: Cache /api/schema/ endpoint
### NetBox version

v3.7.6

### Feature type

### Proposed functionality

Add caching on the `/api/schema/ ` endpoint with a sane default (maybe 300 seconds?) and with the possibility to configure it in configuration.py

### Use case

Generating openapi schema takes some time, caching it would:

- Decrease response time when doing multiple calls to `/api/schema/`
- Lower the resources needed from application server (since less openapi schemas will be generated)

### Database changes

None

### External dependencies

None

mulmat:
### Fixes: #15894

Re-added caching on api schema generation endpoint, similar to what was implemented before in https://github.com/netbox-community/netbox/commit/257c0afdb5b62c7ea84159d2b46d53ce5a43f3a3. Also added netbox version in cache-key in order to bust cache immediately when upgrading netbox.

### DIFF
--- a/netbox/netbox/urls.py
+++ b/netbox/netbox/urls.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.conf.urls import include
 from django.urls import path
+from django.views.decorators.cache import cache_page
 from django.views.decorators.csrf import csrf_exempt
 from django.views.static import serve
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Confirmed response time drops from ~4 seconds for initial schema request to ~100ms for successive requests locally. Thanks @mulmat!

Issue #14692, opened 2024-01-04
Allow disk size to be smaller than 1GB


v0tti:
### NetBox version

v3.7.0

### Feature type

Change to existing functionality

### Proposed functionality

I would like to be able to create virtual disks with a size that is smaller than 1GB. Ideally, disk sizes could specified in MB (like memory) or decimal values could be allowed.

### Use case

We operate a lot of small VMs that are just PBX appliances. These only need disks of a few MB in size. At the moment we can not model these VMs properly since the smallest disk size possible is 1GB.

### Database changes

If the disk size field is changed to represent MB no change to the database is necessary, but existing data would have to be converted during migration. Alternatively, the field could be converted to a decimal field (like vCPUs).

jeremystretch:
Either approach, changing the atomic unit or converting the value to a decimal, would be a breaking API change. It's unfortunate that you didn't raise this suggestion during the evaluation period ahead of the v3.7 release. We could have considered the change then, but now that v3.7 has been released this would need to wait for a new minor version bump.

That said, even before the introduction of the new VirtualDisk model, the `disk` attribute on VirtualMachine was always measured in gigabytes, so I'm not sure how much demand there is for the proposed change.

ITJamie:
one suggestion that could allow this usecase without it making a major breaking change.

add an extra choice field for disk_size_measurement with options of gb/mb/tb with a default of gb. that would allow people who need to use smaller measurements to select mb as the type and use the existing integer field

jose-d:
Just noting, that small HDDs are very common for standalone EFI partitions. Checking my datastores and I see lot of ~5MB disks for EFI part. Proxmox/Qemu.

abrahamvegh:
Came here to file this while, glad to see I’m not the only one, and I have an additional use case:

I use NetBox to document Linode (_Akamai_) VMs, amongst other things. The storage quotas there are defined in MB, so while in theory documenting a drive in GB is mostly correct, being able to document the actual MB value would be more accurate, and allowing sub-GB values would allow accurate tracking of swap disk size, which is currently impossible (most are 256/512MB).

MalfuncEddie:
What would be the impact if it was just allowed to be a float number and keep the GB status.

jeremystretch:
It would still be a breaking change as the data type would change.

PR #15900, opened 2024-04-30
15873 - Make Cluster resource counters more readable

### LINKED ISSUE #15873: make cluster resource counters more readable (add commas or convert amounts to more readable measurements)
### NetBox version

v4.0-beta1

### Feature type

### Proposed functionality

adjust the resource counters on the cluster view page to add comma's or change from MB -> GB -> TB 
Screenshot 2024-04-26 at 23 47 29

right now it can be kind of unreadable for larger clusters

### Use case

when viewing clusters with a large amount of assigned resources (high amounts of ram , CPU or storage) it can be a little hard to read (especially memory as its in MB)

### Database changes

n/a

### External dependencies

n/a

Julio-Oliveira-Encora:
### Fixes: #15873 

One method was created to normalize the VM's disk and memory sizes according to their units. That method is applied when the "Allocated Resources" is generated.


netbox/virtualization/views.py
```diff
@@ -172,7 +173,17 @@ class ClusterView(generic.ObjectView):
     queryset = Cluster.objects.all()
 
     def get_extra_context(self, request, instance):
-        return instance.virtual_machines.aggregate(vcpus_sum=Sum('vcpus'), memory_sum=Sum('memory'), disk_sum=Sum('disk'))
```
jeremystretch, line 175: We need to keep the aggregation logic at the database layer in the interest of performance.

netbox/virtualization/views.py
```diff
@@ -172,7 +173,17 @@ class ClusterView(generic.ObjectView):
...
+        vm_memory = [convert_byte_size(item.memory, 'mega') for item in instance.virtual_machines.all() if item.memory]
```
jeremystretch, line 176: It's sufficient to capture and then round the raw aggregate value. Also, we generally want to handle this sort of massaging of numbers in the template (via a template tag) rather than in the view to afford maximum flexibility.

jeremystretch (changes requested):
I really think we just need to make some small adjustments to the existing `humanize_megabytes()` function to address the goals of #15873.

Julio-Oliveira-Encora:
I was analyzing it, and there is no point in trying to convert and predict any incorrect input from the customer.
The "memory" field in the VM is already informed that the value must be in MB, so if the client enters, for example, the value "68000" for one VM and "64" for another, the result of the sum will be "68064 ", but this is not Netbox's "fault", but rather the incorrect input.

We should discuss it.

In this commit, I changed "humanize_megabytes" to convert non-divisible values to 1024.

Screenshot from 2024-05-09 16-16-17

DO-AMS3

netbox/templates/virtualization/cluster.html
```diff
@@ -61,7 +61,7 @@ <h5 class="card-header">{% trans "Allocated Resources" %}</h5>
                   {% if memory_sum %}
                       {{ memory_sum|humanize_megabytes }}
                   {% else %}
-                      {{ ''|placeholder }}
+                      {{ ''|placeholder }}
```
jeremystretch, line 64: Remove errant trailing whitespace

netbox/utilities/templatetags/helpers.py
```diff
@@ -100,7 +100,7 @@ def humanize_megabytes(mb):
         return f'{int(mb / 1048576)} TB'
     if not mb % 1024:
         return f'{int(mb / 1024)} GB'
-    return f'{mb} MB'
+    return f'{int(mb) / (1048576):.2f} MB'
```
jeremystretch, line 103: This logic needs to be rewritten as our intent is now to display the human-friendly value using the largest appropriate unit. Values of 1TB or larger should be displayed in fractional terabytes; values of 1GB or larger should be in gigabytes; and anything smaller should be in megabytes

jeremystretch, line 103: Additionally, because we're now rounding to the nearest hundredth of a TB or GB, we also need to also convey the exact value in megabytes. This can be achieved by wrapping the value in a `<span>` element with a `title` attribute indicating the raw value, which will be displayed on mouseover.

jeremystretch (changes requested):

netbox/utilities/templatetags/helpers.py
Julio-Oliveira-Encora: I rewrite the logic. I considered the MB to be coming in "Megabytes" because, in the VMs, the unit for memory is MB.

netbox/utilities/templatetags/helpers.py
```diff
@@ -93,14 +93,29 @@ def humanize_speed(speed):
...
     """
     Express a number of megabytes in the most suitable unit (e.g. gigabytes or terabytes).
+    It considers the mb value as megabytes and converts it to the most suitable unit.
     """
+
+    # Factors in bytes
+    factors = {
+        "mega": 1024 ** 2,
+        "giga": 1024 ** 3,
+        "tera": 1024 ** 4,
```
jeremystretch, line 103: I think we should change this logic to work on powers of 10 rather than powers of 2, given that the [commonly accepted definition of gigabyte](https://en.wikipedia.org/wiki/Gigabyte) is 1,000 megabytes.

arthanson (changes requested):
@Julio-Oliveira-Encora I think we should also make the same change to the virtual machine detail page to make it consistent
Monosnap child vm | NetBox 2024-05-30 12-12-01

arthanson:
@Julio-Oliveira-Encora yes, probably should be consistent.

arthanson:
actually never-mind about the disk, there is a 4.1 issue [#14692](https://github.com/netbox-community/netbox/issues/14692) where we need to change the data-field anyways, so will do it there.

jeremystretch (approved):

Issue #15831, opened 2024-04-24
Mirroring LDAP groups doesn't work with v4.0-beta2


Cornelicorn:
### Deployment Type

NetBox Cloud

### NetBox Version

v4.0-beta2

### Python Version

3.11

### Steps to Reproduce

Setup netbox with ldap authentication and set `AUTH_LDAP_MIRROR_GROUPS=True`.
This was tested using `django-auth-ldap==4.8.0` on Debian 12.

Then try to login with an LDAP user.
This Setup is working on v4.0-beta1

### Expected Behavior

The Groups get added to the user and you can login.

### Observed Behavior

An error `500` is raised with the following traceback:
```
Internal Server Error: /login/
Traceback (most recent call last):
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/db/models/fields/__init__.py", line 2117, in get_prep_value
    return int(value)
           ^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'Group'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/utils/decorators.py", line 48, in _wrapper
    return bound_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/views/decorators/debug.py", line 143, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #15902, opened 2024-04-30
15831 monkeypatch LDAP _mirror_group function for NB4

arthanson:
### Fixes: #15831 

Mokey-patches _mirror_groups from django-auth-ldap to use NetBox Group instead of the Django.auth one.  Code is directly from django-auth-ldap, no changes except for the import statement above the function definition.

### DIFF
--- a/netbox/netbox/authentication.py
+++ b/netbox/netbox/authentication.py
@@ -333,7 +333,6 @@ try:
         user's membership.
         """
         target_group_names = frozenset(self._get_groups().get_group_names())
-        target_group_names = frozenset("testgroup",)
         current_group_names = frozenset(
             self._user.groups.values_list("name", flat=True).iterator()
         )
--- a/netbox/netbox/authentication.py
+++ b/netbox/netbox/authentication.py
@@ -313,7 +313,8 @@ class RemoteUserBackend(_RemoteUserBackend):
 
 # Create a new instance of django-auth-ldap's LDAPBackend with our own ObjectPermissions
 try:
-    from django_auth_ldap.backend import LDAPBackend as LDAPBackend_
+    from django_auth_ldap.backend import _LDAPUser, LDAPBackend as LDAPBackend_
+    from users.models import Group
 
     class NBLDAPBackend(ObjectPermissionMixin, LDAPBackend_):
         def get_permission_filter(self, user_obj):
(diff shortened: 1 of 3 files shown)


netbox/netbox/authentication.py
```diff
@@ -323,6 +324,52 @@ def get_permission_filter(self, user_obj):
...
+
+    # Monkey-patch _mirror_groups, code is from django-auth-ldap.backends._LDAPUser
+    # There are no changes to this routine, the 'fix' is the import of Group above.
+    def _mirror_groups(self):
+        """
+        Mirrors the user's LDAP groups in the Django database and updates the
+        user's membership.
+        """
+        target_group_names = frozenset(self._get_groups().get_group_names())
+        target_group_names = frozenset("testgroup",)
```
Cornelicorn, line 336:
```suggestion
```

netbox/netbox/authentication.py
```diff
@@ -323,6 +324,52 @@ def get_permission_filter(self, user_obj):
...
+            )
+            existing_group_names = frozenset(group.name for group in existing_groups)
+
+            new_groups = [
+                Group.objects.get_or_create(name=name)[0]
+                for name in target_group_names
+                if name not in existing_group_names
+            ]
+
+            self._user.groups.set(existing_groups + new_groups)
```
Cornelicorn, line 369: I'd guess that this would be also OK (performance wise) if you don't do a single query for the existing groups, but simplify this to
```suggestion
            target_groups = [
                Group.objects.get_or_create(name=name)[0]
                for name in target_group_names
            ]

            self._user.groups.set(target_groups)
```

Cornelicorn (commented):
Works both on fresh installs and updates for me after dropping the constant `target_group_names` assignment.

No noticable performance difference when simplifying the groups update, but I haven't measured that quantitatively.

netbox/netbox/authentication.py
arthanson: Yes, it's just exactly the same code as in django-auth-ldap, I didn't want to change it so it is easy to merge in any changes in django-auth-ldap in the future.  The only difference is the import statement above it to switch to the NetBox Group instead of the Django.auth one.

Cornelicorn (commented):
Still works.

PR #15909, opened 2024-05-01
Fixes #15852: Update total object counts when filtering object lists

### LINKED ISSUE #15852: Edit/Delete all when filtering with quick search selects all objects, not only the filtered objects
### Deployment Type

Self-hosted

### NetBox Version

v4.0.0-beta2

### Python Version

3.11

### Steps to Reproduce

1. Create dependant devices/interfaces in order to have a sufficiently long enough interface list (2k interfaces is safe)
2. Filter interfaces with the quick search so that you return ~200
3. Hit the "Select All" checkbox
4. Prompt will come up with option to edit all interfaces returned as result of query

### Expected Behavior

* Query should report ~200 interfaces
* Editing/Deleting should only delete the 200 interfaces

### Observed Behavior

* Query reports initial interface list count (~2000)
* Editing/Deleting performs bulk action on all interfaces

jeremystretch:
### Fixes: #15852

Employ HTMX out-of-band swaps to dynamically update the total object counts when filtering a list of objects.


DanSheps (approved):
Looks good!

PR #15918, opened 2024-05-02
Fixes #15917: slim-select-pagination-bug-fix : fixed several bugs related to slim select

### LINKED ISSUE #15917: slim-select-pagination-bug : user not able to load more than one page of data after a text search in slim select search box
### Deployment Type

NetBox Cloud

### NetBox Version

v3.7.7-dev

### Python Version

3.11

### Steps to Reproduce

1. insert enough entries in extras_tag table such that for a search string "x" there are more than 50 (default max page size) items returned
2. open edit page for a dcim device interface
3. open tag dialogue box which uses a slim select GUI control under the hood, enter search string "x" in the filter search box
4. scroll down to the bottom of the returned list 
5.  notice you cannot get more items behind the first page of returned items although an api call to the next page is already made

### Expected Behavior

When user reaches the end of currently returned list of items more items will be extended to the list if there are more.  

### Observed Behavior

Current behaviour is that user will not be able to go behind the first page of returned items.

MaxBear:
### Fixes: #15917 

1. introduce internal state variable searchKey
2. add render() in handleSearch() function
3. call handleSearch() when atbottom event is triggered
4. call handleSearch() after refresh button is clicked


arthanson:
This library has been completely removed in NB v4, so this would no longer apply. 

@MaxBear Also, FYI - While we greatly appreciate the work here, please open the issue before submitting a PR, there may be discussion needed around any proposed solution or it may be superseded by other work.

MaxBear:
But the issue and the fix still apply for users not upgraded to beta v4 release. I have already opened the issue , otherwise, i won't have the fix id . By this 
```
here may be discussion needed around any proposed solution or it may be superseded by other work.
```
Do you mean , is should only submit a pr after you have decided to submit a pr for this fix ?

arthanson:
@MaxBear Sorry, you submitted the issue at just the wrong time (no fault of yours) we just did the last release of 3.x before the 4.x release so there wasn't a release to get this in.  That said, we are currently evaluating whether we might cut one more 3.x release and if we do we will re-evaluate potentially including your PR assuming there are not any changes needed.

jeremystretch (approved):

PR #15950, opened 2024-05-03
Closes #15915: Replace plugins list with an overall system status view

### LINKED ISSUE #15915: Replace the plugins list view with a more comprehensive system view
### NetBox version

v4.0-beta2

### Feature type

### Proposed functionality

NetBox currently provides a list of installed plugins under Admin > Plugins. We can extend this view to provide a more general "status" view, including other attributes such as the NetBox & Python versions, database size, and other data in addition to the plugins list.

### Use case

This would provide a convenient summary of the NetBox deployment.

### Database changes

None

### External dependencies

None

jeremystretch:
### Fixes: #15915

- Create a new system status view
- Move the plugins list
- Move the current config view
- Clean up configuration data display


arthanson (approved):
Sweet - this is really nice.  Could also be handy for bug reports, easy way to get full information from a reporter.

jeremystretch:
Good call. I've added an export button to download the status data as a JSON file.

jeremystretch:
screenshot

PR #15953, opened 2024-05-03
Fixes: #15948 - Fixes cable fanin/fanout when both are required

DanSheps:
### Fixes: #15948 - Fixes cable display when both fanin and fanout are required in cable trace

* Adjust conditional to check for both fanin and fanout condition
* Use my sucky maths to figure out a x_center for the fan-in/fan-out

### DIFF
--- a/netbox/dcim/svg/cables.py
+++ b/netbox/dcim/svg/cables.py
@@ -405,7 +405,17 @@ class CableTraceSVG:
                     end = far[0].top_center
                     text_offset = 0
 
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Confirmed that this has restored the original functionality. Thanks @DanSheps! I think it would be beneficial for us to spend some more time on the cable tracing logic as it could stand to be cleaned up a good deal, but for now this is sufficient.

screenshot

PR #15954, opened 2024-05-03
15148 add copy button to config context

### LINKED ISSUE #15148: UX enhencement: Copy Button on context
### NetBox version

v.3.7.2

### Feature type

New functionality

### Proposed functionality

on device or virtual-machine context page, would be nice to have "copy to clipboard" button for local context data, as most of times I need to copy it to text editor for big context data.

### Use case

Currently coping current device's context is not comfortable and if context is so big that requires scrolling, you may accidentally copy some other parts of page also. 

### Database changes

_No response_

### External dependencies

_No response_

arthanson:
### Fixes: #15148
Adds a copy button to the config context area - Added it to the panel title area next to the json/yaml format buttons (see screenshot)

Monosnap Dist-Switch-1 | NetBox 2024-05-06 07-36-09

### DIFF
--- a/netbox/templates/extras/configcontext.html
+++ b/netbox/templates/extras/configcontext.html
@@ -77,15 +77,9 @@
       </div>
     </div>
     <div class="col col-md-7">
+      {% include 'inc/sync_warning.html' %}
       <div class="card">
-        <h5 class="card-header d-flex justify-content-between">
-          {% trans "Data" %}
-          {% include 'extras/inc/configcontext_format.html' %}
-        </h5>
-        <div class="card-body">
-          {% include 'inc/sync_warning.html' %}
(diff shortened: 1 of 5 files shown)


jeremystretch (changes requested):
Can we refactor the inclusion templates into one? It seems very fragile to have the logic split apart like this. We should be able to move the content from `configcontext_format.html` into `configcontext_data.html` and wrap it in a conditional that looks for `include_header` or similar.

jeremystretch:
@arthanson I merged the content of `configcontext_format.html` into `configcontext_data.html` and introduced a separate inclusion template for the JSON/YAML toggle. Let me know what you think.

arthanson:
@jeremystretch looks good.

Issue #15492, opened 2024-03-20
Copying of Admin->Authentication->Permissions


Dominik9798:
### NetBox version

v3.7.4

### Feature type

New functionality

### Proposed functionality

Add a button and function "clone" on Admin/Authentication/Permission

### Use case

At the moment there is no option to copy an existing authorization with all set rights.

Julio-Oliveira-Encora:
Could you please assign it to me?

Dominik9798:
How?

jeremystretch:
@Dominik9798 sorry, that wasn't directed at you.

@Julio-Oliveira-Encora I've assigned this to you.

jeremystretch:
Marking this as blocked until we resolve a circular import error arising from the import of CloningMixin; see #16058.

pheus:
I'd like to contribute this feature. Can you please assign it to me? Thanks

PR #15976, opened 2024-05-06
Feature 15492 - Copying of Admin->Authentication->Permissions

Julio-Oliveira-Encora:
Fixes: #15492

It was created a copy of CloningMixin in users/features.py and used it in the ObjectPermission.


netbox/users/models/features.py
```diff
@@ -0,0 +1,53 @@
...
+from django.db import models
+
+from extras.utils import is_taggable
+
+
+class CloningUserMixin(models.Model):
+    """
+    Provides the clone() method used to prepare a copy of existing objects.
+    The same code from netbox/users/models/features.py (CloningMixin) is used here.
+    It was necessary to avoid circular imports.
```
jeremystretch, line 12: It's not acceptable to duplicate code from other modules, as this poses serious challenges to long-term maintainability. If employing the original mixin class results in a circular dependency, some other solution must be identified.

jeremystretch (changes requested):

netbox/users/models/features.py
Julio-Oliveira-Encora: One solution is to create a "clone"  method in the ObjectPermission, removing the support for "tags" and "custom_fields" (they do not exist on the Users module).
I tested multiple solutions to avoid duplicate code, but they did not work. All of them caused circular imports.

jeremystretch:
Going to close this out as #15492 is blocked by #16058.

PR #16007, opened 2024-05-07
#15981 - Make device description a clickable link

### LINKED ISSUE #15981: Make device description a clickable link
### NetBox version

v3.7.3

### Feature type

### Proposed functionality

When viewing the `/dcim/devices` webpage, I would like an option to make the `Description` field a hyperlink to the device page, just like the `ID` and `Name` columns behave now.

Implementation details are open for discussion.  It could be always on, or a configuration.py parameter, or a user preference.

### Use case

We don't use the `Name` field at all because the uniqueness constraints don't work for our use case.  All of our devices are "unnamed", so I like to hide that column while browsing the UI.  For now, I'm keeping the `ID` column visible so I can click the link to the device.  I would prefer to hide that column and click on the `Description` instead, because the ID number is meaningless for most users, and it's also requires more mouse precision to click the smaller link area.

Current view:
image

Desired change:
image

This might be an edge case, but it would help a lot for usability in our situation and I hope it would be trivial to implement.

### Database changes

_No response_

### External dependencies

none

Julio-Oliveira-Encora:
### Fixes: #15981

Added hyperlink to the Description field on the device page

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -141,6 +141,10 @@ class DeviceTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
         template_code=DEVICE_LINK,
         linkify=True
     )
+    description = tables.Column(
+        verbose_name=_('Description'),
+        linkify=True
+    )
     status = columns.ChoiceFieldColumn(
         verbose_name=_('Status'),
     )


jeremystretch:
Closing this for now as the proposed behavior is not tenable; see my comment [here](https://github.com/netbox-community/netbox/issues/15981#issuecomment-2107620983).

PR #16008, opened 2024-05-07
Fixes #15995: Permit nullable fields referenced by unique constraints to be omitted from REST API requests

### LINKED ISSUE #15995: 'group' required on create API call in tenant
### Deployment Type

Self-hosted

### NetBox Version

v4.0.0

### Python Version

3.10

### Steps to Reproduce

1. Try to create a tenant with only a name and a slug (which works through the GUI)
```
curl -s -X POST \
-H "Authorization: Token 0123456789abcdef0123456789abcdef01234567" \
http://localhost:32768/api/tenancy/tenants/ \
--data '{"name": "Tenant 2", "slug": "tenant-2"}' | jq '.'
```

### Expected Behavior

Tenant to be created

### Observed Behavior

```
{
  "group": [
    "This field is required."
  ]
}
```

jeremystretch:
### Fixes: #15995

- Set a default value on each of the relevant fields in API serializers
- Modify tests to omit these fields in some object representations

### DIFF
--- a/netbox/dcim/api/serializers_/devices.py
+++ b/netbox/dcim/api/serializers_/devices.py
@@ -122,6 +122,7 @@ class DeviceWithConfigContextSerializer(DeviceSerializer):
 class VirtualDeviceContextSerializer(NetBoxModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='dcim-api:virtualdevicecontext-detail')
     device = DeviceSerializer(nested=True)
+    identifier = serializers.IntegerField(allow_null=True, max_value=32767, min_value=0, required=False, default=None)
     tenant = TenantSerializer(nested=True, required=False, allow_null=True, default=None)
     primary_ip = IPAddressSerializer(nested=True, read_only=True, allow_null=True)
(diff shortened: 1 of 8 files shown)


netbox/dcim/api/serializers_/devices.py
```diff
@@ -122,6 +122,7 @@ def get_config_context(self, obj):
 class VirtualDeviceContextSerializer(NetBoxModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='dcim-api:virtualdevicecontext-detail')
     device = DeviceSerializer(nested=True)
+    identifier = serializers.IntegerField(allow_null=True, max_value=32767, min_value=0, required=False, default=None)
```
arthanson, line 125: was this missing previously?

arthanson (approved):
Should we add a test case to catch some of these?

netbox/dcim/api/serializers_/devices.py
jeremystretch: I needed to define the field to set `default=None` on it.

jeremystretch:
I've modified existing tests to intentionally omit the affected fields in certain test data. This triggers the exception without the fix in place.

PR #16028, opened 2024-05-08
Fixes #16027: Correct typo in error message

### LINKED ISSUE #16027: Typo in IPRange model
### Deployment Type

Self-hosted

### NetBox Version

v4.0.0

### Python Version

3.10

### Steps to Reproduce

The error message is defined as "Ending address must be **lower** than ...", but I think the correct definition is "Ending address must be **greater** than ...".
https://github.com/netbox-community/netbox/blob/195dbaed000edd7f832695b17b7a10329fbed57f/netbox/ipam/models/ip.py#L573-L579

### Expected Behavior

Ending address must be greater than the starting address ({start_address})

### Observed Behavior

Ending address must be lower than the starting address ({start_address})

grncbg:
Fixes #16027: Correct typo in error message

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -574,7 +574,7 @@ class IPRange(PrimaryModel):
             if not self.end_address > self.start_address:
                 raise ValidationError({
                     'end_address': _(
-                        "Ending address must be lower than the starting address ({start_address})"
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @grncbg!

PR #16040, opened 2024-05-08
Closes #16034: Disable uWSGI logging

### LINKED ISSUE #16034: Disable uWSGI logging by default
### NetBox version

v4.0.0

### Feature type

### Proposed functionality

Currently the NetBox-default uWSGI configuration causes all requests to be logged in uWSGI level, in addition to the normal reverse proxy level, thus consuming the journal log in systemd configuration.

I propose defaulting to uWSGI request logging disabled, while still logging application errors.

### Use case

Current uWSGI request logging consumes the systemd journal log capacity, possibly causing other application logs to purge when journal log limits are exceeded.

Since the default NetBox installation already logs the requests in the reverse proxy logs, uWSGI-level logging can be limited to error logging only.

If someone needs the uWSGI logging as well, they can easily change the configuration setting present in the default NetBox `uwsgi.ini`.

### Database changes

None

### External dependencies

None

markkuleinio:

### DIFF
--- a/contrib/uwsgi.ini
+++ b/contrib/uwsgi.ini
@@ -22,3 +22,7 @@ chdir = netbox
 
 ; specify the WSGI module to load
 module = netbox.wsgi
+
+; only log internal messages and errors (reverse proxy already logs the requests)
+disable-logging = true
+log-5xx = true


jeremystretch (approved):
Thanks @markkuleinio!

PR #16044, opened 2024-05-08
Fixes: #16039 - Fix row highlighting on device components and VM interfaces

### LINKED ISSUE #16039: Front and Rear Ports do not highlight connected rows
### Deployment Type

Self-hosted

### NetBox Version

v4.0.0

### Python Version

3.10

### Steps to Reproduce

1. Create 2 interfaces
2. Create 2 Rear Ports
3. Create 2 Front Ports
4. Connect 2 interfaces to 2 Front Ports
5. Connect 2 rear ports together
6. View Front Ports and Rear Ports

### Expected Behavior

Row to be highlighted

### Observed Behavior

Row is not highlighted

DanSheps:
Fixes: #16039 - Fix row highlighting on device components and VM interfaces

* Convert device component tables to utilize the same row highlighting technology that interfaces use
* Convert VMInterfaceTable to utilize the same row highlighting technology that device interfaces use

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -390,7 +390,10 @@ class DeviceConsolePortTable(ConsolePortTable):
         )
         default_columns = ('pk', 'name', 'label', 'type', 'speed', 'description', 'cable', 'connection')
         row_attrs = {
-            'class': get_cabletermination_row_class
+            'data-name': lambda record: record.name,
+            'data-mark-connected': lambda record: "true" if record.mark_connected else "false",
+            'data-cable-status': lambda record: record.cable.status if record.cable else "",
+            'data-type': lambda record: record.type
         }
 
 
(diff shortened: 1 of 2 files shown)


netbox/dcim/tables/devices.py
```diff
@@ -390,7 +390,10 @@ class Meta(DeviceComponentTable.Meta):
         )
         default_columns = ('pk', 'name', 'label', 'type', 'speed', 'description', 'cable', 'connection')
         row_attrs = {
-            'class': get_cabletermination_row_class
```
jeremystretch, line 393: If this is no longer used, let's remove it.

netbox/dcim/tables/devices.py
```diff
@@ -390,7 +390,10 @@ class Meta(DeviceComponentTable.Meta):
...
+            'data-name': lambda record: record.name,
+            'data-mark-connected': lambda record: "true" if record.mark_connected else "false",
+            'data-cable-status': lambda record: record.cable.status if record.cable else "",
+            'data-type': lambda record: record.type
```
jeremystretch, line 396: Lots of static duplication here. Can we replace these with a function that returns the dict for easy reuse?

jeremystretch (changes requested):
Thanks @DanSheps!

netbox/dcim/tables/devices.py
DanSheps: Yeah, I will see what I can do.

jeremystretch (approved):

PR #16045, opened 2024-05-08
Closes #16043: Add 'die-on-term = true' to fix stopping uWSGI

### LINKED ISSUE #16043: Add "die-on-term" to fix uWSGI shutdown
### Change Type

Addition

### Area

Installation/upgrade

### Proposed Changes

When using "`systemctl stop netbox`", systemd sends **SIGTERM** to the process (see https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillSignal=).

However, as documented in https://uwsgi-docs.readthedocs.io/en/latest/Management.html#signals-for-controlling-uwsgi, for uWSGI the SIGTERM signal means:

> brutally **reload** all the workers and the master process

= it is not a "stop" signal for uWSGI.

(_... "this bad choice has been fixed in uWSGI 2.1" as mentioned by the maintainer in https://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html, but it is apparently not happening any time soon._)

Thus, I'm proposing adding `die-on-term = true` to be added in the NetBox-default `uwsgi.ini`.

That makes the default `systemctl` behavior to work with uWSGI as well, instead of changing the kill signal in `netbox.service`, to keep it simplier for also Gunicorn use.

With that addition NetBox shuts down correctly with "`systemctl stop netbox`":
```
May 08 18:42:53 netbox-lab systemd[1]: Stopping netbox.service - NetBox WSGI Service...
May 08 18:42:53 netbox-lab uwsgi[572422]: SIGINT/SIGTERM received...killing workers...
May 08 18:42:54 netbox-lab uwsgi[572422]: worker 1 buried after 1 seconds
May 08 18:42:54 netbox-lab uwsgi[572422]: goodbye to uWSGI.
May 08 18:42:54 netbox-lab systemd[1]: netbox.service: Deactivated successfully.
May 08 18:42:54 netbox-lab systemd[1]: Stopped netbox.service - NetBox WSGI Service.
```

---

**Note:** With the current default configuration, "systemctl stop netbox" _just happens to currently work_, because:

```
May 08 18:33:06 netbox-lab systemd[1]: Stopping netbox.service - NetBox WSGI Service...
May 08 18:33:06 netbox-lab uwsgi[572209]: ...brutally killing workers...
May 08 18:33:07 netbox-lab uwsgi[572209]: worker 1 buried after 1 seconds
May 08 18:33:07 netbox-lab uwsgi[572209]: binary reloading uWSGI...
May 08 18:33:07 netbox-lab uwsgi[572209]: chdir() to /opt/netbox
May 08 18:33:07 netbox-lab uwsgi[572209]: closing all non-uwsgi socket fds > 2 (max_fd = 1024)...
May 08 18:33:07 netbox-lab uwsgi[572209]: found fd 3 mapped to socket 0 (127.0.0.1:8001)
May 08 18:33:07 netbox-lab uwsgi[572209]: running /usr/bin/python3.11
May 08 18:33:07 netbox-lab uwsgi[572209]: Traceback (most recent call last):
May 08 18:33:07 netbox-lab uwsgi[572209]:   File "/opt/netbox/venv/bin/uwsgi", line 5, in <module>
May 08 18:33:07 netbox-lab uwsgi[572209]:     from pyuwsgi import run
May 08 18:33:07 netbox-lab uwsgi[572209]: ModuleNotFoundError: No module named 'pyuwsgi'
May 08 18:33:07 netbox-lab systemd[1]: netbox.service: Main process exited, code=exited, status=1/FAILURE
May 08 18:33:07 netbox-lab systemd[1]: netbox.service: Failed with result 'exit-code'.
May 08 18:33:07 netbox-lab systemd[1]: Stopped netbox.service - NetBox WSGI Service.
```
= it tries to reload uWSGI but a module import error prevents that happening.

I don't currently know why that happens, I haven't used `pyuwsgi` before, and this module import problem does not occur with `uwsgi` package. (But the "reload vs. stop" problem is still valid with `uwsgi` as well, so `die-on-term = true` is needed anyway.)

markkuleinio:

### DIFF
--- a/contrib/uwsgi.ini
+++ b/contrib/uwsgi.ini
@@ -11,7 +11,7 @@ master = true
 ; clear environment on exit
 vacuum = true
 
-; make SIGTERM to stop the app (instead of reload by default)
+; make SIGTERM stop the app (instead of reload)
 die-on-term = true
 
 ; exit if no app can be loaded
--- a/contrib/uwsgi.ini
+++ b/contrib/uwsgi.ini
@@ -11,6 +11,9 @@ master = true
 ; clear environment on exit
 vacuum = true
 
+; make SIGTERM to stop the app (instead of reload by default)
+die-on-term = true
+
 ; exit if no app can be loaded
 need-app = true
 


jeremystretch (approved):
Thanks @markkuleinio!

PR #16065, opened 2024-05-09
Adds extra column for result on the script job table

### LINKED ISSUE #15983: Display Script Results Instead of Job Logs
### NetBox version

v4.0.0

### Feature type

Restore functionality back to the way it was prior to NetBox 4.0; so you can get a link back to the results, from the list of jobs.

### Proposed functionality

From the list of completed jobs, provide link to the script "results" page instead of the jobs "logs" page.

Screenshot 2024-05-06 at 18 09 14

Script Results: (right after the script runs)
Screenshot 2024-05-06 at 18 10 51

Job Log: (Prior to NetBox 4.0 clicking on the "Name" would bring you to the results page. Now both "ID" and "Name" bring you to log page. 
Screenshot 2024-05-06 at 18 12 57

### Use case

Restore functionality back to the way it was prior to NetBox 4.0; so you can get a link back to the results. 

Change this link:
/core/jobs/41699/

to this:
/extras/scripts/results/41699/

It's much easier to read the results with all the formatting and color coding. 

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -17,6 +17,7 @@ from dcim.models import Device, DeviceRole, Platform
 from extras.dashboard.forms import DashboardWidgetAddForm, DashboardWidgetForm
 from extras.dashboard.utils import get_widget_class
 from netbox.constants import DEFAULT_ACTION_PERMISSIONS
+from netbox.tables import columns
 from netbox.views import generic
 from netbox.views.generic.mixins import TableMixin
 from utilities.data import shallow_compare_dict
@@ -1126,9 +1127,15 @@ class ScriptJobsView(generic.ObjectView):
         jobs_table = JobTable(
             data=script.jobs.all(),
             orderable=False,
-            user=request.user
+            user=request.user,
+            extra_columns=[(
+                'result', columns.TemplateColumn(
+                    template_code='<a href="{% url "extras:script_result" job_pk=record.pk %}">{{ record.completed }}</a>'
+                )
+            ),]
         )
         jobs_table.configure(request)
+        jobs_table.columns.show('result')
 
         return render(request, 'extras/script/jobs.html', {
             'script': script,
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -1130,7 +1130,8 @@ class ScriptJobsView(generic.ObjectView):
             user=request.user,
             extra_columns=[(
                 'result', columns.TemplateColumn(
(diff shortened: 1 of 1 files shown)


netbox/extras/views.py
```diff
@@ -1126,9 +1127,15 @@ def get(self, request, **kwargs):
         jobs_table = JobTable(
             data=script.jobs.all(),
             orderable=False,
-            user=request.user
+            user=request.user,
+            extra_columns=[(
+                'result', columns.TemplateColumn(
+                    template_code='<a href="{% url "extras:script_result" job_pk=record.pk %}">{{ record.completed }}</a>'
```
jeremystretch, line 1133: We've started presenting all dates & times in ISO 8601 format now using the `isoformat()` template filter for consistency.

screenshot

jeremystretch, line 1133: We also need to explicitly pass the verbose name for each column now to support translation (e.g. `verbose_name=_('Result')`).

jeremystretch (changes requested):
Thanks @abhi1693! Just a couple minor nits.

bitcollector1:
much appreciated folks!

bitcollector1:
This is much appreciated. Is there anyway to get the results from the main job page? the one with ALL the jobs? 

I use this page the most often. took me awhile to figure out why the results were not posting here:
https://netdevbox.hwe.ebay.com/core/jobs/

but I do see the results now listed here:
https://netdevbox.hwe.ebay.com/extras/scripts/17/jobs/

bitcollector1:
can we add this same info to the core jobs table?  That is the page I use the most to check the status of all my various jobs.

abhi1693:
> can we add this same info to the core jobs table

It's out of scope. Results column only makes sense for scripts and jobs table can have other types in it

jeremystretch:
Rather than adding another column to the table, it might make more sense to override the "completed" column so that it links to the script result. We could do this in a generic way so that the "completed" column in any jobs table links to the relevant record where applicable.

bitcollector1:
> Rather than adding another column to the table, it might make more sense to override the "completed" column so that it links to the script result. We could do this in a generic way so that the "completed" column in any jobs table links to the relevant record where applicable.


Thanks Jeremy, 

All I'm looking for us parity with 3.7, where you can get to the results from the main page. Appreciate the support and excellent work on a great product.

arthanson:
@abhi1693 Are you looking at overriding the completed column as Jeremy mentioned or do you think this is good to go as is?

jeremystretch:
Closing this for inactivity.

PR #16094, opened 2024-05-12
Closes #16090: Show NetBox version if plugin validation fails

### LINKED ISSUE #16090: Add current NetBox version in the `PluginConfig.validate()` error output
### NetBox version

v4.0.1

### Feature type

New functionality

### Proposed functionality

When [`PluginConfig.validate()`](https://github.com/netbox-community/netbox/blob/4a64a3f6e0d0edf27996422eb2dbe0e197a6bea5/netbox/netbox/plugins/__init__.py#L133) determines that the current NetBox version does not meet the plugin requirements, also print out the current NetBox version in the exception message.

### Use case

Currently the error message only prints the version that the plugin mandates, but due to possible installation directory confusion it would be good to also print the current NetBox version. Example case: https://github.com/netbox-community/netbox/issues/16088

### Database changes

None

### External dependencies

None

markkuleinio:

### DIFF
--- a/netbox/netbox/plugins/__init__.py
+++ b/netbox/netbox/plugins/__init__.py
@@ -138,13 +138,15 @@ class PluginConfig(AppConfig):
             min_version = version.parse(cls.min_version)
             if current_version < min_version:
                 raise ImproperlyConfigured(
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @markkuleinio!

PR #16096, opened 2024-05-12
add ENABLE_TRANSLATION setting to optionally turn translation off

### LINKED ISSUE #16055: Option to turn translation off
### NetBox version

v4.0.0

### Feature type

### Proposed functionality

Since #15752 there is no way to turn localization off.

I propose to bring this feature back to users.
The simplest way is to just move [USE_I18N](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-USE_I18N) Django setting inside `configuration.py`

### Use case

1. Many old users of NetBox (including me) are used to default English translation even though they are not English natives. They may get confused seeing translated UI.
2. Having one unified UI using one same language for all users may be beneficial for international teams. Consider this example: newbie NetBox user asks experienced one about the way of creating a Custom Field. If both of them see different versions of UI (due to different languages), one can't just tell another something like "go to Customization, then Custom Fields" and so on.

### Database changes

_No response_

### External dependencies

_No response_

amyasnikov:

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -446,6 +446,7 @@ LOGIN_REDIRECT_URL = f'/{BASE_PATH}'
 # Use timezone-aware datetime objects
 USE_TZ = True
 
+# Toggle language translation support
 USE_I18N = ENABLE_TRANSLATION
 
 # WSGI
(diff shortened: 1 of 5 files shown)


netbox/netbox/settings.py
```diff
@@ -156,6 +156,7 @@
 STORAGE_BACKEND = getattr(configuration, 'STORAGE_BACKEND', None)
 STORAGE_CONFIG = getattr(configuration, 'STORAGE_CONFIG', {})
 TIME_ZONE = getattr(configuration, 'TIME_ZONE', 'UTC')
+USE_I18N = getattr(configuration, 'USE_I18N', True)
```
jeremystretch, line 159: While an argument can be made for preserving Django setting names, I expect the term i18n won't make sense to most users. IMO we should call this something like `ENABLE_TRANSLATION`.

jeremystretch (changes requested):
While this implements the root translation control, it does not address UX: A user can still change his or her language, and would naturally assume that preference to have an effect. Please extend the PR to also disable the user's language preference when translation has been disabled (or at a minimum indicate as much to the user).

amyasnikov:
@jeremystretch fixed, pls take a look

jeremystretch (approved):
Thanks @amyasnikov!

Issue #15569, opened 2024-03-28
Add better tests for GraphQL filtering and lookup


arthanson:
### Proposed Changes

The GraphQL tests currently test all model fields, but are very light on testing any filtering of objects or sub-objects.  We should add tests for checking different primary filters:
```
{
  site_list(filters: {name: {i_contains:"DM"}}) {
    id
    name
  }
}
```
and sub-object filtering:
```
{
  site_list {
    id
    name
    tags(filters: {name: {i_starts_with: "P"}}) {
      name
    }
  }
}
```

### Justification

The filtering logic for Strawberry is undergoing some major refactors and we could easily miss issues with filtering without tests in-place.

loulecrivain:
Heya, just a heads up from a Netbox-using organization regarding this proposed change.

I've seen and experienced some issues/regressions regarding GraphQL filters following latest release and refactors.
Now I know GraphQL isn't used much internally in Netbox itself. However for lots of people writing exporters, crawlers, you name it, GraphQL API is actually the go-to.

In my opinion, if we want to be on par with the level of quality of the REST API, ensuring filters are working as expected should be a priority. +1 for automated testing here

loulecrivain:
Commenting on this issue again, since we had another regression / breaking change in a minor release update (4.1.x -> 4.2.x).

#18433 broke in-production software depending on the GraphQL API. I'm going to be honest, I get the need for making the models evolve according to users needs, but why do so in a minor release?

I urge you to consider prioritizing better end-to-end tests for the GraphQL API. IMO if it's not as stable as the REST API, then there's no point in having this feature available for the users.

cnemo-cenic:
Just adding that my employer pays for NetBox cloud and we recently saw breakage in a critical production service that depends on the graphql API due to a minor update.

jeremystretch:
@cnemo-cenic could you please link the bug report for reference?

cnemo-cenic:
I don't have a bug report. You want me to create one in this repo?

jeremystretch:
@cnemo-cenic it likely won't do much good now I'm afraid. If a bug report had been submitted soon after the release was made, we could have potentially shipped a fix to reverse any unexpected breaking changes.

In the absence of a bug report, could you share specifically what unexpected change(s) were made (and in what release) for consideration in the development of additional tests?

cnemo-cenic:
Yes we were specifically using a graphql query for ASNs:

```
query {
[...]

PR #16115, opened 2024-05-13
16078 make GraphQL NumberFilter optional

### LINKED ISSUE #16078: GraphQL: Required fields that probably shouldn't be
### Deployment Type

Self-hosted

### NetBox Version

v4.0.1

### Python Version

3.11

### Steps to Reproduce

1. Go to GraphQL console
2. Enter this query
```graphql
query {
  ip_address_list(filters: {tenant: "my-tenant"}){
    address
  }
}
```
3. You'll have an error that some required fields are missing

### Expected Behavior

The `family`, `mask_length__gte` and `mask_length__lte` filters shouldn't be mandatory.

Probably the same for other filters.
Also broken on https://demo.netbox.dev/

### Observed Behavior

```json
{
  "data": null,
  "errors": [
    {
      "message": "Field 'IPAddressFilter.family' of required type 'Int!' was not provided.",
      "locations": [
        {
          "line": 2,
          "column": 28
        }
      ]
    },
    {
      "message": "Field 'IPAddressFilter.mask_length__gte' of required type 'Int!' was not provided.",
      "locations": [
        {
          "line": 2,
          "column": 28
        }
      ]
    },
    {
      "message": "Field 'IPAddressFilter.mask_length__lte' of required type 'Int!' was not provided.",
      "locations": [
        {
          "line": 2,
          "column": 28
        }
      ]
    }
  ]
}
```

arthanson:
### Fixes: #16078 

Make NumberFilter not required in GraphQL.

### DIFF
--- a/netbox/utilities/testing/api.py
+++ b/netbox/utilities/testing/api.py
@@ -441,6 +441,10 @@ class APIViewTestCases:
             return getattr(self, 'graphql_base_name', base_name)
 
         def _build_query_with_filter(self, name, filter_string):
+            """
+            Called by either _build_query or _build_filtered_query - construct the actual
+            query given a name and filter string
+            """
             type_class = get_graphql_type_for_model(self.model)
 
             # Compile list of fields to include
@@ -488,6 +492,9 @@ class APIViewTestCases:
             return query
 
         def _build_filtered_query(self, name, **filters):
+            """
+            Create a filtered query: i.e. ip_address_list(filters: {address: "1.1.1.1/24"}){.
+            """
             if filters:
                 filter_string = ', '.join(f'{k}: "{v}"' for k, v in filters.items())
                 filter_string = f'(filters: {{{filter_string}}})'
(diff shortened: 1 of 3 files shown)


jeremystretch (changes requested):
@arthanson could you add a few basic GrapQL filter tests to catch this sort of error?

arthanson:
@jeremystretch added test used graphql_filter in test-case so it can be extended.  Kept this current test narrow in scope as there is #15569 that can expand on this in the future - will also want to look at adding lookups to this in the future.

netbox/utilities/testing/api.py
```diff
@@ -440,13 +440,8 @@ def _get_graphql_base_name(self):
             base_name = self.model._meta.verbose_name.lower().replace(' ', '_')
             return getattr(self, 'graphql_base_name', base_name)
 
-        def _build_query(self, name, **filters):
+        def _build_query_with_filter(self, name, filter_string):
```
jeremystretch, line 443: Could you add docstrings for these three methods please? They all look very similar and it's not immediately clear how they differ.

- `_build_query_with_filter()`
- `_build_filtered_query()`
- `_build_query()`

jeremystretch (changes requested):

netbox/utilities/testing/api.py
arthanson: done

jeremystretch (approved):
Thanks @arthanson!

PR #16118, opened 2024-05-13
Fixes #16101: Fix initial loading of pagination widget for dynamic object tables

### LINKED ISSUE #16101: htmx_table is missing pagination
### Deployment Type

Self-hosted

### NetBox Version

v4.0.1

### Python Version

3.11

### Steps to Reproduce

1. Set Page length to 25 under user preferences
2. Create site X
2. Create more than 25 locations within site X
3. Navigate to site detail view for X

### Expected Behavior

Locations table should show the first 25 locations and pagination below to be able to navigate to the rest of locations.

### Observed Behavior

Only the first 25 locations are shown with no indication that more exist.

image

Curiously if you sort by any column, the pagination appears and is usable until the next full page refresh.

jeremystretch:
### Fixes: #16101

Refactor HTMX table structure to swap the entire container, ensuring that the paginator is always included where appropriate.

**Please test thoroughly.** It seems to work fine AFAICT but there may be some corner cases lingering that I have not considered.

### DIFF
--- a/netbox/templates/htmx/table.html
+++ b/netbox/templates/htmx/table.html
@@ -2,15 +2,17 @@
 {% load helpers %}
 {% load render_table from django_tables2 %}
 
-{% with preferences|get_key:"pagination.placement" as paginator_placement %}
-  {% if paginator_placement == 'top' or paginator_placement == 'both' %}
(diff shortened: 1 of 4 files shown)


arthanson (approved):
Seems to be working, completely separate issue it looks like site list isn't paginated

PR #16152, opened 2024-05-15
16138 fix user/group permissions

### LINKED ISSUE #16138: User and Group object types missing from permissions in 4.0
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.11

### Steps to Reproduce

In the UI:
1. Admin > Permissions
2. Add or Edit a permission
3. Inspect list of selectable Object types.
4. `Authentication and Authorization  > Group` still exists, but it does nothing (Admin > Groups is visible on the navigation bar, but accessing it returns 403 permission denied).
5. `Authentication and Authorization > User` is gone, and users who used to have that permission can no longer view/edit the Users page.

In Postgresql:
1. Upgrade from 3.7.8 to 4.0.2 and run migrations 
2. `psql -d netbox`
3. `\d`
4. Observe that `auth_user` and all other related tables are gone (suspect one of the [migrations](https://github.com/netbox-community/netbox/blob/develop/netbox/users/migrations/0005_alter_user_table.py) did it but it's unclear which).
5. Observe that `auth_group` and `auth_group_permissions` remains but are empty (it's more clear that [this is the migration](https://github.com/netbox-community/netbox/blob/develop/netbox/users/migrations/0006_custom_group_model.py#L66-L74), no apparent deletion of the table is present).

### Expected Behavior

I see that there's a couple migration scripts. I suspect that the DB migration from `auth_user` to `users_user` is incomplete but don't understand enough to definitively say so.

The ability to manipulate Users and Groups should be possible by users _without_ superuser powers.  This is more obviously broken since 4.0.x.

### Observed Behavior

Users with permission to view/change users and groups get 403 permission denied when accessing those pages.

arthanson:
### Fixes: #16138 

Fixes several issues with user/group permissions:
1. in menu - update to check the correct permissions
2. in constants - change so correct permissions are shown in the Permissions form "Object Type" field
3. add a migration to change any permissions set for auth.groups to users.group.  Note: users does not need to be changed as the content-type was migrated.


jeremystretch (approved):
Thanks @arthanson!

PR #16170, opened 2024-05-16
16145 Use module.ScriptName to call Script API instead of PK

### LINKED ISSUE #16145: 4.0.2 scripts can only be launched by api with ID and no longer with a name.
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.10

### Steps to Reproduce

see #16123

In the past I could run the script with "https://netbox-test/api/extras/scripts/scriptA.class"
This does not work anymore and I need an ID ;(

Exception Value: | Field 'id' expected a number but got 'ProcessLogging'.

### Expected Behavior

it works lik in 3.7.7

### Observed Behavior

ValueError at /api/extras/scripts/ProcessLogging/

launching it with /api/extras/scripts/6/

Does work.

arthanson:
### Fixes: #16145 

Changes the script api back to using the "module.ScriptName" instead of id (this is the way it was in NB3.x).  The test was a bit of a mess to fix, the issue is the get_vars function in the Script class calls the python_class property which calls module_scripts on on the module - this tries to load the module which fails as the script doesn't actually exist on disk.

```
curl -X 'GET' \
  'http://localhost:8000/api/extras/scripts/testscript.NewBranchScript/' \
  -H 'accept: application/json; indent=4' \
  -H "Authorization: Token 6f22af1129a164f92e4abdfa716cb54df9684f5e"

{
    "id": 3,
    "url": "http://localhost:8000/api/extras/scripts/3/",
    "module": 7,
    "name": "NewBranchScript",
    "description": "Provision a new branch site",
    "vars": {
        "site_name": "StringVar",
        "switch_count": "IntegerVar",
        "switch_model": "ObjectVar",
        "manufacturer": "ObjectVar"
    },
    "result": {
        "object_id": null,
        "name": "",
        "scheduled": null,
        "interval": null,
        "started": null,
        "completed": null,
        "data": null,
        "job_id": null
    },
    "display": "NewBranchScript (testscript)",
    "is_executable": true
}
```

### DIFF
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -14,7 +14,7 @@ from rq import Worker
 from core.models import Job, ObjectType
 from extras import filtersets
 from extras.models import *
-from extras.scripts import get_module_and_script, run_script
+from extras.scripts import run_script
 from netbox.api.authentication import IsAuthenticatedOrLoginNotRequired
 from netbox.api.features import SyncedDataMixin
 from netbox.api.metadata import ContentTypeMetadata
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):
Could we retain the numeric PK form of the URL as well, as an alternative? That would be more consistent with the manner by which other objects in NetBox are referenced.

jeremystretch:
Thanks @arthanson!

PR #16185, opened 2024-05-17
16050 Show script python_class name and description

### LINKED ISSUE #16050: Custom Scripts - Module & Script Attributes
### Deployment Type

Self-hosted

### NetBox Version

v4.0.0

### Python Version

3.10

### Steps to Reproduce

1. Create a custom script module including multiple scripts
2. Define a "name" for the module
3. Define a "name" and "description" for the scripts.
4. Add the script module to NetBox

### Expected Behavior

When the script module is added, NetBox should reflect all configured attributes on the GUI interface.

image

### Observed Behavior

- NetBox uses the script module file name as the name instead of the defined "name" attribute
- NetBox uses the script class name as the name of the script instead of the defined "name" attribute
- NetBox does not show the defined "description" attribute.

Custom_Scripts

arthanson:
### Fixes: #16050 

Removed module name customization from docs as this doesn't look like it worked in v3.7 after discussion with Jeremy.


jeremystretch (changes requested):
This looks to address the script name & description but not the [module name](https://docs.netbox.dev/en/stable/customization/custom-scripts/#name). Can we fix that as well?

samk-acw:
I tried these changes manually on my server and had issue with `script.python_class.description` - If a script has a field called "description" it will throw an exception:

><class 'AttributeError'>
>'StringVar' object has no attribute 'strip'

Seems like the description attribute (probably name and any other attributes) get overwritten with the field vars if there's a field with the same name
Changing it to `script.python_class.Meta.description` resolved it for me.

jeremystretch (approved):

PR #16192, opened 2024-05-18
16173 - Fix pagination when `pagination.per_page` is `""`

### LINKED ISSUE #16173: Clearing page length user preference breaks pagination
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.10

### Steps to Reproduce

1. Create a new user account
2. Visit a page that uses pagination. It should load properly.
3. Open `/user/preferences/` and click the "X" on `Page length`.
4. Click `Save`.
5. Visit the page in step 2. An error should occur.

### Expected Behavior

The page should load without an error.

### Observed Behavior

```
<class 'TypeError'>

'<' not supported between instances of 'int' and 'str'

Python version: 3.10.14
NetBox version: 4.0.2
Plugins: 
  netbox_demo: 0.5.0
```

I believe this is because the `paginate.per_page` preference is an empty string:

```
GET /api/users/config/

{
    "ui": {
        "htmx_navigation": ""
    },
    "locale": {
        "language": ""
    },
    "pagination": {
        "per_page": "",
        "placement": "bottom"
    },
    "data_format": "json"
}
```

Therefore, `per_page` is set to `""` instead of `config.PAGINATE_COUNT`:
https://github.com/netbox-community/netbox/blob/4b2f26a8002a8f01e186211c648768892cc199c8/netbox/utilities/paginator.py#L90

devon-mar:
### Fixes: #16173

Fixes `TypeError` when `paginate.per_page` user preference is `""`.
This happens when the user clears the per page preference which sets it to the empty string, causing
`per_page` to be a string instead of an integer.

### DIFF
--- a/netbox/utilities/paginator.py
+++ b/netbox/utilities/paginator.py
@@ -87,7 +87,7 @@ def get_paginate_count(request):
             pass
 
     if request.user.is_authenticated:
-        per_page = request.user.config.get('pagination.per_page', config.PAGINATE_COUNT)
+        per_page = request.user.config.get('pagination.per_page') or config.PAGINATE_COUNT
         return _max_allowed(per_page)
 
     return _max_allowed(config.PAGINATE_COUNT)


jeremystretch (approved):
Thanks @devon-mar!

PR #16206, opened 2024-05-20
14010 add IEC 60130-12 power port connetors

### LINKED ISSUE #14010: Add  "IEC 60130-10" (DC) to power port types
### NetBox version

v3.6.3

### Feature type

Data model extension

### Proposed functionality

The existing model for "power ports" includes a very large amount of AC-type ports, but not much in the way of DC power.
It would be nice to have better documentation about what specific type of connector(s) a device has.

I would like to propose these additions:
* Barrel connector (ideally, a list of various ID & ODs for the barrel would be nice, but this gets to be a mess very quickly - there is currently not even a generic "barrel plug" option and I'd be happy with that)
* Bolt-through terminal (is this what "hardwired" under "Other" is meant to be? or is hardwired actually intended for a "permanent, requires electrician" connection?)
* [Screw-down wire-clamp terminal](https://mou.sr/45Q6xC9) 
* [Removable Terminal Block connector](https://mou.sr/3ZMNMye) (there are a slew of these as well, I guess technically "DC Terminal" would cover this as a generic, but documenting at least some of the more common connector types would be "better")

### Use case

Attempting to correctly model real-world equipment requires the port types to be defined. 😉 

### Database changes

(no db changes expected?)

### External dependencies

_No response_

arthanson:


jeremystretch (changes requested):

jeremystretch (commented):
This adds the port (male) types; what about the outlet (female) types? Presumably we need to add those as well.

PR #16209, opened 2024-05-20
14948 add has_virtual_device_contexts filter to device

### LINKED ISSUE #14948: Add "has_virtual_device_context" filter to device filters in UI and API
### NetBox version

v3.7.1

### Feature type

### Proposed functionality

The only way to find devices that have virtual device context is to get all virtual-device-contexts and extract the list of related devices from that. Similar to the has_primary_ip filter.

### Use case

Get the list of devices with have or have not a virtual-device-context with a simple boolean filter.

### Database changes

Not required

### External dependencies

None

arthanson:
### Fixes: #14948 

Adds a has_virtual_device_contexts filter to devices.

Monosnap Devices | NetBox 2024-05-20 13-45-04

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1100,9 +1100,9 @@ class DeviceFilterSet(
         queryset=IPAddress.objects.all(),
         label=_('OOB IP (ID)'),
     )
-    has_virtual_device_contexts = django_filters.BooleanFilter(
-        method='_has_virtual_device_contexts',
-        label=_('Has virtual device contexts'),
+    has_virtual_device_context = django_filters.BooleanFilter(
+        method='_has_virtual_device_context',
+        label=_('Has virtual device context'),
     )
 
     class Meta:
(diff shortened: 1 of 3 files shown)


netbox/dcim/filtersets.py
```diff
@@ -1100,6 +1100,10 @@ class DeviceFilterSet(
         queryset=IPAddress.objects.all(),
         label=_('OOB IP (ID)'),
     )
+    has_virtual_device_contexts = django_filters.BooleanFilter(
```
jeremystretch, line 1103: The filter name should be singular, as we're not checking for the presence of _multiple_ VDCs. (The corresponding method should be renamed too.)
```suggestion
    has_virtual_device_context = django_filters.BooleanFilter(
```

netbox/dcim/forms/filtersets.py
```diff
@@ -813,6 +814,13 @@ class DeviceFilterForm(
             choices=BOOLEAN_WITH_BLANK_CHOICES
         )
     )
+    has_virtual_device_contexts = forms.NullBooleanField(
```
jeremystretch, line 817:
```suggestion
    has_virtual_device_context = forms.NullBooleanField(
```

jeremystretch (changes requested):
Please include a test for the new filter.

jeremystretch (approved):
Thanks @arthanson!

PR #16212, opened 2024-05-20
Capitalize Filter Form Labels

### LINKED ISSUE #16104: Ensure Consistency In Multi Word Filters
### NetBox version

v4.0.1

### Feature type

### Proposed functionality

Currently, in the User Interface, the models have Filter forms that are multi-word that are named like "Site group" but for the Side Navigation, multi word labels are capitalized to "Site Groups." This request is to ensure consistency in the filters.

### Use case

It would ensure consistent grammar without breaking any functionality. It's a minor change, but improves readability and ensures consistency throughout the entire program. Thus improving the user experience 

### Database changes

_No response_

### External dependencies

_No response_

danner26:
### Fixes: #16104 

Adds `text-transform: capitalize;` to the `col-form-label` so that all filter form labels are capitalized. This will also help with capitalization for addition tables as well (i.e. adding a device).

### DIFF
--- a/.github/workflows/close-incomplete-issues.yml
+++ b/.github/workflows/close-incomplete-issues.yml
@@ -30,4 +30,3 @@ jobs:
             This is a reminder that additional information is needed in order to further
             triage this issue. If the requested details are not provided, the issue will
             soon be closed automatically.
-          start-date: 2024-05-14


jeremystretch:
We don't do this in CSS because there are corner cases where we might want to preserve lowercase lettering. For example, consider a field named "eBGP peer."

danner26:
Fair enough, okay I will revert this and start manually updating the labels

PR #16213, opened 2024-05-20
Capitalize Attribute Table Labels

### LINKED ISSUE #16108: Capitalize Multi Word Table Columns For Consistency
### NetBox version

v4.0.1

### Feature type

### Proposed functionality

In the core, the last_updated field column shows up as "Last updated" when other columns have both the 1st characters in both words capitalized. This is to help ensure consistency. There are probably other fields like this, and this is to help maintain consistency.

### Use case

In the core, the last_updated field column shows up as "Last updated" when other columns have both the 1st characters in both words capitalized. This is to help ensure consistency. There are probably other fields like this, and this is to help maintain consistency. It just looks weird now, without a change

### Database changes

_No response_

### External dependencies

_No response_

danner26:
### Fixes: #16108

Adds `text-transform: capitalize;` to the `tr > th` inside of `table.attr-table` so that all attribute form labels are capitalized.


jeremystretch:
Same issue as with #16212: We can't assume that attribute names should _always_ be capitalized.

danner26:
Fair enough, okay I will revert this and start manually updating the labels

PR #16221, opened 2024-05-21
Proposed updates to incomplete issue auto-closure messages

jeffgdotorg:
This was my last task of the day yesterday so it may need revisions.


DanSheps (approved):
I am liking the new messages.

jeremystretch (changes requested):
Let's keep these responses short and to the point. (Remember, they're going to be replicated many times.) Rather than including several paragraphs a text, a succinct message with a link to a more detailed explanation elsewhere (perhaps the wiki) will suffice.

PR #16233, opened 2024-05-21
16232 only remove checkbox on embedded tables

### LINKED ISSUE #16232: Netbox UI. Checkbox disappearing in any Table (Site, Rack, etc) after clicking to next page
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.11

### Steps to Reproduce

1. Open Sites or Rack
2. Create more then default pagination (50) records
3. Move between pages or change per page number

### Expected Behavior

Moving between pages in Tables should not hide checkbox element
image

### Observed Behavior

Checkbox in first column disappeared. Not possible to select object
image

arthanson:
### Fixes: #16232 

Not 100% sure of this solution, relies on the 'return_url' which will be for bulk operations not being the same as the request path.  As this routine tries to set if the table is embedded, I'm not sure if there is another flag to look at that would be different if it is a main table or embedded.

### DIFF
--- a/netbox/netbox/views/generic/bulk_views.py
+++ b/netbox/netbox/views/generic/bulk_views.py
@@ -112,6 +112,13 @@ class ObjectListView(BaseMultiObjectView, ActionsMixin, TableMixin):
             query_params.pop('export')
             return redirect(f'{request.path}?{query_params.urlencode()}')
 
+    def _is_embedded_path(self, request):
+        if 'return_url' in request.GET:
+            if request.GET['return_url'] != request.path:
+                return True
+
+        return False
+
     #
     # Request handlers
     #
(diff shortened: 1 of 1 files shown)


netbox/netbox/views/generic/bulk_views.py
```diff
@@ -112,6 +112,13 @@ def export_template(self, template, request):
             query_params.pop('export')
             return redirect(f'{request.path}?{query_params.urlencode()}')
 
+    def _is_embedded_path(self, request):
+        if 'return_url' in request.GET:
```
jeremystretch, line 116: We can't rely on `return_url` as an indicator for the nature of the request.

jeremystretch:
I ended up taking a different, more concrete approach in #16239. We should still dedicate some time at some point to doing some general cleanup around embedded tables, but I think this suffices for now.

PR #16269, opened 2024-05-23
16252 only show results count if paginator

### LINKED ISSUE #16252: Elevations results count span contains a dot instead of a number
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.10

### Steps to Reproduce

0. Start with a fresh, empty install
1. Navigate to Organization -> Elevations
2. Note appearance of the Results tab
3. Create a site
4. Create a rack
5. Navigate again to Organization -> Elevations
6. Note again the appearance of the Results tab

### Expected Behavior

The Results tab should hava a badge containing a number: either `0` in step 2 or `1` in step 6

### Observed Behavior

The Results tab is badged with a dot. Upon inspecting the document, the `<span>` element in question is completely empty.

arthanson:

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -651,7 +651,6 @@ class RackElevationListView(generic.ObjectListView):
             'rack_face': rack_face,
             'filter_form': forms.RackElevationFilterForm(request.GET),
             'model': self.queryset.model,
-            'table': {'page': page},  # hack to show count in Result tab
         })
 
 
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):
This doesn't address the root issue: RackElevationListView does not provide a `table` object in the template context, so the number of objects is never reported.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #16270, opened 2024-05-23
16202 fix mapit button for internationalized decimal seperator

### LINKED ISSUE #16202: Google Map It Button Link is broken when used a foreign language that uses comma as decimal separator
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.10

### Steps to Reproduce

1. Change user language preference to "Portuguese"
2. Create or edit a site and fill the latitude / longitude with decimal values, separated by *COMMA* and save
3. Press the 'map it' button

### Expected Behavior

The google maps tab opens with the correct latitude / longitude spotted. 

### Observed Behavior

The google maps site will apper in a new tab, but with error message "O Google Maps não encontrou xxxx,yyyy".
This appears to be caused by Google Maps expecting the latitude / longitude values to use PERIOD decimal separator, instead of the COMMA separator. Maybe a simple conversion to US locale when generating the Map It button link can solve this.

arthanson:
### Fixes: #16202 

Used a filter function to get DECIMAL_SEPARATOR from Django as it could be something other then a comma.

### DIFF
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -63,7 +63,7 @@
                         {% if object.latitude and object.longitude %}
                           {% if config.MAPS_URL %}
                             <div class="position-absolute top-50 end-0 translate-middle-y d-print-none">
-                              <a href="{{ config.MAPS_URL }}{{ object.latitude }},{{ object.longitude }}" target="_blank" class="btn btn-primary">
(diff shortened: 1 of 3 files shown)


jeremystretch (changes requested):
Any reason not to use Django's stock [`unlocalize()`](https://docs.djangoproject.com/en/5.0/topics/i18n/formatting/#unlocalize) template filter here?

jeremystretch (approved):

PR #16271, opened 2024-05-23
16149 add (optional) obj hyperlink to script list table

### LINKED ISSUE #16149: Custom script logs are missing hyperlinks
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.11

### Steps to Reproduce

1. Create a custom script with logs using `self.log_info(some_object, message)`
2. Run the script to write the logs
3. View the logs in the jobs log view

### Expected Behavior

Prior to 4.0 the object passed to `log_info` was displayed with a hyperlink to quicly reach the details view.

### Observed Behavior

Now we have a URL column instead wich we have to copy and paste into the browser or search for the name of the object.

arthanson:
### Fixes: #16149 

* Fix documentation to refer to obj instead of object
* Add (optional) obj, url to script table (was only on reports table)
* hyperlink the obj, url to the respective object when display
* add the obj to the log message as documentation states and backwards compatability


jeremystretch (changes requested):
This does not appear to resolve #16149; I see the same output in the script result before and after the change. Additionally, the table configuration button does not work on the script result view.

arthanson:
> This does not appear to resolve #16149; I see the same output in the script result before and after the change. Additionally, the table configuration button does not work on the script result view.

@jeremystretch I'm not seeing those issues?  With the branch I get the configuration button, which you need to do as the object is an optional parameter.  If you do add the parameter then you do get a clickable link (second screenshot).  Is there a data condition I'm missing that is causing the config screen not to appear on yours?

Monosnap NewBranchScript | NetBox 2024-06-10 11-24-25
Monosnap NewBranchScript | NetBox 2024-06-10 11-24-40

jeremystretch:
@arthanson looks like the table configuration button is left in a broken state when the script first completes. It's functional on successive reloads. The modal content is likely missing on the first load.

jeremystretch (changes requested):
Table configuration button is broken on completion of script execution. Also, we should probably make `object` a default table column.

arthanson:
Made object default, the table configuration button wasn't working as the filter list is only included when the script finishes, so made the button dependent on that as well.

netbox/extras/scripts.py
```diff
@@ -500,12 +500,16 @@ def _log(self, message, obj=None, level=LogLevelChoices.LOG_DEFAULT):
...
+                'url': obj.get_absolute_url() if hasattr(obj, 'get_absolute_url') else None,
             })
 
             # Record to the system log
             if obj:
                 message = f"{obj}: {message}"
-            self.logger.log(LogLevelChoices.SYSTEM_LEVELS[level], message)
+                self.logger.log(LogLevelChoices.SYSTEM_LEVELS[level], f"{obj}: {message}")
+            else:
+                self.logger.log(LogLevelChoices.SYSTEM_LEVELS[level], message)
```
jeremystretch, line 512: This appears to be redundant to the line above (which is already prefixing the message with `{obj}: `).

netbox/extras/tables/tables.py
```diff
@@ -545,15 +546,30 @@ class ScriptResultsTable(BaseTable):
         template_code="""{% load log_levels %}{% log_level record.status %}""",
         verbose_name=_('Level')
     )
+    object = tables.Column(
+        verbose_name=_('Object')
+    )
+    url = tables.Column(
+        verbose_name=_('URL')
+    )
```
jeremystretch, line 554: I would ditch the URL column. IMO it's poor form to print a raw URL as a value, and it's redundant to the `object` column.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #16287, opened 2024-05-24
16281 internationalize dashboard headers

### LINKED ISSUE #16281: Colum headlines in dashboard are not translated in foreign languages
### Deployment Type

Self-hosted

### NetBox Version

v4.0.3

### Python Version

3.10

### Steps to Reproduce

1. Make a screenshot of the English Dashboard.
2. Goto User Preferences and Change Language to what ever
3. Go back to the Dashboard: The Headlines will not have been translated

### Expected Behavior

Headlines shown as translated.

### Observed Behavior

Only native english language is shown
The content is shown in the right translation

arthanson:

### DIFF
--- a/netbox/extras/constants.py
+++ b/netbox/extras/constants.py
@@ -1,3 +1,5 @@
+from django.utils.translation import gettext_lazy as _
+
 # Events
 EVENT_CREATE = 'create'
 EVENT_UPDATE = 'update'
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
If we enable translation support for widget headers, it should be done at render time. Attempting to store a wrapped string as a constant is likely to lead to issues when replicating/saving the data. (Also, we don't need to translate note content as it's user-generated.)

PR #16319, opened 2024-05-28
Extend STORAGE_BACKEND config to support Swift

### LINKED ISSUE #14554: Extend STORAGE_BACKEND to support swift
### NetBox version

3.2

### Feature type

### Proposed functionality

Netbox currently supports `django-storages` through the `STORAGE_CONFIG` and `STORAGE_BACKEND`.

I'd like to extend it to support https://github.com/dennisv/django-storage-swift which work overall the same way as `django-storages` so the changes should be minimal.

I already have a PoC over there https://gerrit.wikimedia.org/r/c/operations/software/netbox/+/980908 its not tested yet but will be if this feature request is accepted.

(And yes, we need to upgrade to 3.7).

### Use case

`django-storages` unfortunately doesn't support Openstack SWIFT as backend. As it's a popular storage platform it seems beneficial to all Netbox users.
We're currently using SWIFT through some custom/internal Netbox patches, so it works. Here we would like to streamline it and make it available to all.

### Database changes

_No response_

### External dependencies

If used, `django-storage-swift` will obviously be needed :)

XioNoX:
### Fixes: #14554

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -177,7 +177,7 @@ The dotted path to the desired search backend class. `CachedValueSearchBackend` 
 
 Default: None (local storage)
 
-The backend storage engine for handling uploaded files (e.g. image attachments). NetBox supports integration with the [`django-storages`](https://django-storages.readthedocs.io/en/stable/) and [`django-storage-swift`](https://github.com/dennisv/django-storage-swift) packages, which provides backends for several popular file storage services. If not configured, local filesystem storage will be used.
+The backend storage engine for handling uploaded files (e.g. image attachments). NetBox supports integration with the [`django-storages`](https://django-storages.readthedocs.io/en/stable/) and [`django-storage-swift`](https://github.com/dennisv/django-storage-swift) packages, which provide backends for several popular file storage services. If not configured, local filesystem storage will be used.
 
 The configuration parameters for the specified storage backend are defined under the `STORAGE_CONFIG` setting.
 
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -187,7 +187,7 @@ The configuration parameters for the specified storage backend are defined under
 
 Default: Empty
 
-A dictionary of configuration parameters for the storage backend configured as `STORAGE_BACKEND`. The specific parameters to be used here are specific to each backend; see the [`django-storages` documentation](https://django-storages.readthedocs.io/en/stable/) or [`django-storage-swift` documentation](https://github.com/dennisv/django-storage-swift) for more detail.
+A dictionary of configuration parameters for the storage backend configured as `STORAGE_BACKEND`. The specific parameters to be used here are specific to each backend; see the documentation for your selected backend ([`django-storages`](https://django-storages.readthedocs.io/en/stable/) or [`django-storage-swift`](https://github.com/dennisv/django-storage-swift)) for more detail.
 
 If `STORAGE_BACKEND` is not defined, this setting will be ignored.
 
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -177,7 +177,7 @@ The dotted path to the desired search backend class. `CachedValueSearchBackend` 
 
 Default: None (local storage)
 
-The backend storage engine for handling uploaded files (e.g. image attachments). NetBox supports integration with the [`django-storages`](https://django-storages.readthedocs.io/en/stable/) package, which provides backends for several popular file storage services. If not configured, local filesystem storage will be used.
+The backend storage engine for handling uploaded files (e.g. image attachments). NetBox supports integration with the [`django-storages`](https://django-storages.readthedocs.io/en/stable/) and [`django-storage-swift`](https://github.com/dennisv/django-storage-swift) packages, which provides backends for several popular file storage services. If not configured, local filesystem storage will be used.
 
 The configuration parameters for the specified storage backend are defined under the `STORAGE_CONFIG` setting.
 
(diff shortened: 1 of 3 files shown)


XioNoX:
@arthanson hello, is it possible to prevent the automatic closure? Thanks

netbox/netbox/configuration_example.py
```diff
@@ -230,6 +230,17 @@
...
+# STORAGE_BACKEND = 'swift.storage.SwiftStorage'
+# STORAGE_CONFIG = {
+#     'SWIFT_AUTH_URL': 'https://swift.netbox.dev/auth/v1.0',
+#     'SWIFT_USERNAME': 'netbox',
+#     'SWIFT_PASSWORD': 'Secret',
+#     'SWIFT_CONTAINER_NAME': 'netbox',
+#     'SWIFT_USE_TEMP_URLS': True,
+#     'SWIFT_TEMP_URL_KEY': 'Secret',
+#     'SWIFT_BASE_URL': 'https://demo.netbox.dev/',
+# }
```
jeremystretch, line 243: Please remove this from the example config. This file is meant merely to provide a rough starting point.

docs/configuration/system.md
```diff
@@ -177,7 +177,7 @@ The dotted path to the desired search backend class. `CachedValueSearchBackend`
 
 Default: None (local storage)
 
-The backend storage engine for handling uploaded files (e.g. image attachments). NetBox supports integration with the [`django-storages`](https://django-storages.readthedocs.io/en/stable/) package, which provides backends for several popular file storage services. If not configured, local filesystem storage will be used.
+The backend storage engine for handling uploaded files (e.g. image attachments). NetBox supports integration with the [`django-storages`](https://django-storages.readthedocs.io/en/stable/) and [`django-storage-swift`](https://github.com/dennisv/django-storage-swift) packages, which provides backends for several popular file storage services. If not configured, local filesystem storage will be used.
```
jeremystretch, line 180:
```suggestion
The backend storage engine for handling uploaded files (e.g. image attachments). NetBox supports integration with the [`django-storages`](https://django-storages.readthedocs.io/en/stable/) and [`django-storage-swift`](https://github.com/dennisv/django-storage-swift) packages, which provide backends for several popular file storage services. If not configured, local filesystem storage will be used.
```

docs/configuration/system.md
```diff
@@ -187,7 +187,7 @@ The configuration parameters for the specified storage backend are defined under
 
 Default: Empty
 
-A dictionary of configuration parameters for the storage backend configured as `STORAGE_BACKEND`. The specific parameters to be used here are specific to each backend; see the [`django-storages` documentation](https://django-storages.readthedocs.io/en/stable/) for more detail.
+A dictionary of configuration parameters for the storage backend configured as `STORAGE_BACKEND`. The specific parameters to be used here are specific to each backend; see the [`django-storages` documentation](https://django-storages.readthedocs.io/en/stable/) or [`django-storage-swift` documentation](https://github.com/dennisv/django-storage-swift) for more detail.
```
jeremystretch, line 190:
```suggestion
A dictionary of configuration parameters for the storage backend configured as `STORAGE_BACKEND`. The specific parameters to be used here are specific to each backend; see the documentation for your selected backend ([`django-storages`](https://django-storages.readthedocs.io/en/stable/) or [`django-storage-swift`](https://github.com/dennisv/django-storage-swift)) for more detail.
```

netbox/netbox/settings.py
```diff
@@ -219,12 +224,24 @@
             raise e
 
         # Monkey-patch django-storages to fetch settings from STORAGE_CONFIG
-        def _setting(name, default=None):
-            if name in STORAGE_CONFIG:
-                return STORAGE_CONFIG[name]
-            return globals().get(name, default)
         storages.utils.setting = _setting
 
+    elif STORAGE_BACKEND == 'swift.storage.SwiftStorage':
```
jeremystretch, line 229:
```suggestion
    # django-storage-swift
    elif STORAGE_BACKEND == 'swift.storage.SwiftStorage':
```

netbox/netbox/settings.py
```diff
@@ -219,12 +224,24 @@
...
+        except ModuleNotFoundError as e:
+            if getattr(e, 'name') == 'swift':
+                raise ImproperlyConfigured(
+                    f"STORAGE_BACKEND is set to {STORAGE_BACKEND} but django-storage-swift is not present. "
+                    "It can be installed by running 'pip install django-storage-swift'."
+                )
+            raise e
+
+        # Monkey-patch django-storage-swift to fetch settings from STORAGE_CONFIG
+        swift.utils.setting = _setting
```
jeremystretch, line 242: Not a big fan of monkey-patching if it can be avoided. I'd rather handle this like [we do for django-social-auth](https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/settings.py#L584-L587), by importing any global variables prefixed with `SWIFT_` from `configuration.py`.

jeremystretch (changes requested):

netbox/netbox/settings.py
XioNoX: Indeed ! PR updated. I guess we could do the same with `django-storages` but I didn't want to risk breaking something external to my change as I don't use that lib.

XioNoX:
Hi @jeremystretch I still see "Changes requested", so I want to make sure nothing is blocked on me. Thanks.

jeremystretch (changes requested):
@XioNoX this should be good to go except for the inclusion of `django.po`; this file should be removed from the PR. (I tried to handle this myself but I don't have permission to push changes to your branch.)

XioNoX:
> @XioNoX this should be good to go except for the inclusion of `django.po`; this file should be removed from the PR. (I tried to handle this myself but I don't have permission to push changes to your branch.)

I think it's good now :) thanks

jeremystretch (approved):
Thanks @XioNoX!

PR #16324, opened 2024-05-28
16315 - Cant filter changelog by object type (no results found)

### LINKED ISSUE #16315: Cant filter changelog by object type (no results found)
### Deployment Type

Self-hosted

### NetBox Version

v4.0.3

### Python Version

3.11

### Steps to Reproduce

goto changelog
goto filters tab
select object type dropdown.

### Expected Behavior

list of netbox object types to be selectable

### Observed Behavior

 "no results found" appears in dropdown list
 
Screenshot 2024-05-28 at 12 44 26

Julio-Oliveira-Encora:
### Fixes: #16315

Replaced `/api/extras/content-types/`  with `/api/extras/object-types/` for `JournalEntryFilterForm` and `ObjectChangeFilterForm`.


jeremystretch (changes requested):
Ideally we don't want to return an unfiltered list of object types. Instead, we can change these fields to instances of ContentTypeMultipleChoiceField; see CustomLinkFilterForm for an example:

https://github.com/netbox-community/netbox/blob/fdad59c8cc7bad24261e3f265c1843a9056b44f8/netbox/extras/forms/filtersets.py#L114-L118

PR #16329, opened 2024-05-28
16218 copy script on data-source sync

### LINKED ISSUE #16218: Scripts issue post v4.0.2 upgrade
### Deployment Type

Self-hosted

### NetBox Version

4.0.2

### Python Version

3.12

### Steps to Reproduce

This bug occurs with a 3.7 -> 4.x upgrade and cannot be resolved without removing affected scripts.  It only occurs when the upgrade occurs with missing scripts.

Replicate empty 'file' environment whereas DB has been restored but files NOT:

- Create Script using git datastore
- Manually remove file from netbox scripts dir ; breaking things and replicating an unsynced environment

Go to Script:

- Expected: NetBox 3.7: 'error: missing file'.
- Observed: Missing file

Run Upgrade to 4.0.2 - upgrade succeeds 

Go to scripts - 

- expected: runnable (if data source has synced)
- Observed: unable to run scripts, run button missing, clicking edit yields to empty form

Sync datasource (UI, or manage.py syncdatasource --all) manually

- Expected: script recreated, script can run
- Observed: no change in behaviour as above - also no files created in scripts/ directory

REPO used: https://github.com/cruse1977/test-scripts

### Expected Behavior

- Following a git sync scripts should become runnable and exist in the /script directory

### Observed Behavior

- following a git sync (successful), no change in behaviour; no runs button, no files created in /scripts; editting the script gives an empty form with no pre-populated items (ie: the data source).  Manually copying the affected script to /scripts also does not resolve this.

arthanson:
### Fixes: #16218 

Script module needs sync_data called when data source is synced to make sure any updated file is copied over.  Also was an error in data.py where the underscore was used later so it was counted as a local variable which was causing RQ to crash.

### DIFF
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -11,7 +11,8 @@ from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
 from core.choices import ManagedFileRootPathChoices
-from core.models import ManagedFile
+from core.models import DataSource, ManagedFile
+from core.signals import post_sync
 from extras.utils import is_script
 from netbox.models.features import JobsMixin, EventRulesMixin
 from utilities.querysets import RestrictedQuerySet
@@ -182,3 +183,10 @@ class ScriptModule(PythonModuleMixin, JobsMixin, ManagedFile):
 @receiver(post_save, sender=ScriptModule)
 def script_module_post_save_handler(instance, created, **kwargs):
     instance.sync_classes()
+
+
+@receiver(post_sync, sender=DataSource)
+def script_data_source_sync_handler(instance, **kwargs):
+    modules = ScriptModule.objects.filter(data_source=instance)
+    for module in modules:
+        module.sync_data()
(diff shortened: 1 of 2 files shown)


netbox/extras/models/scripts.py
```diff
@@ -182,3 +183,10 @@ def save(self, *args, **kwargs):
 @receiver(post_save, sender=ScriptModule)
 def script_module_post_save_handler(instance, created, **kwargs):
     instance.sync_classes()
+
+
+@receiver(post_sync, sender=DataSource)
+def script_data_source_sync_handler(instance, **kwargs):
```
jeremystretch, line 189: We do not want to automatically sync files for which auto-sync has not been explicitly enabled. (That is the purpose of AutoSyncRecord.)

jeremystretch (commented):
Per my [comment](https://github.com/netbox-community/netbox/issues/16218#issuecomment-2176268637) on #16218, I'm not sure any changes are actually needed at this time.

PR #16371, opened 2024-05-31
Allow plugins to extend objectchange view

### LINKED ISSUE #16353: Allow plugins to extend objectchange.html (changeview) template
### NetBox version

v4.0.3

### Feature type

### Proposed functionality

Allow plugins to use the `PluginTemplateExtension` feature to extend the objectchange.html (changeview) template with additional panels. (/netbox/netbox/templates/extras/objectchange.html)

I've tested this locally and it is as simple as adding `{% load plugins %}` to the template along with the panel placements
eg:
```
<div class="row">
  <div class="col col-md-12">
    {% plugin_full_width_page object %}
  </div>
</div>
```

### Use case

I have an internal plugin that I wish to add some panels to the changelog view that will point out some business-related information about changes.

### Database changes

N/A

### External dependencies

N/A

ITJamie:
### Fixes: #16353

Allows plugins to extend objectchange view with panels

### DIFF
--- a/netbox/templates/extras/objectchange.html
+++ b/netbox/templates/extras/objectchange.html
@@ -146,12 +146,12 @@
     </div>
 </div>
 <div class="row mb-3">
-	<div class="col col-md-6">
+  <div class="col col-md-6">
     {% plugin_left_page object %}
   </div>
-	<div class="col col-md-6">
+  <div class="col col-md-6">
     {% plugin_right_page object %}
-	</div>
+  /div>
 </div>
 <div class="row mb-3">
     <div class="col col-md-12">
--- a/netbox/templates/extras/objectchange.html
+++ b/netbox/templates/extras/objectchange.html
@@ -1,5 +1,6 @@
 {% extends 'generic/object.html' %}
 {% load helpers %}
+{% load plugins %}
 {% load i18n %}
 
 {% block title %}{{ object }}{% endblock %}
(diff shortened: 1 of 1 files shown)


netbox/templates/extras/objectchange.html
```diff
@@ -144,6 +145,14 @@ <h5 class="card-header">{% trans "Post-Change Data" %}</h5>
         </div>
     </div>
 </div>
+<div class="row mb-3">
+  <div class="col col-md-6">
+    {% plugin_left_page object %}
+  </div>
+  <div class="col col-md-6">
+    {% plugin_right_page object %}
+  /div>
```
jeremystretch, line 154: Fix HTML tag.
```suggestion
  </div>
```

netbox/templates/extras/objectchange.html
```diff
@@ -144,6 +145,14 @@ <h5 class="card-header">{% trans "Post-Change Data" %}</h5>
         </div>
     </div>
 </div>
+<div class="row mb-3">
```
jeremystretch, line 148: The `<div>` elements with class `row` should not have class `mb-3` applied as it results in extraneous vertical margin between cards. (It should have been removed from the existing elements during the v4.0 UI overhaul.)

jeremystretch (changes requested):

netbox/templates/extras/objectchange.html
ITJamie: Would you like me to remove `mb-3` from all the others in this template?

jeremystretch: Yes, but I'll do it real quick.

jeremystretch (approved):
Thanks @ITJamie!

PR #16384, opened 2024-06-03
Fixes: #15016 - Catch AssertionError from cable trace and throw ValidationError

### LINKED ISSUE #15016: Error when adding new connection to front port.
### Deployment Type

Self-hosted

### NetBox Version

v3.7.1

### Python Version

3.8

### Steps to Reproduce

You can reproduce same behaviour in demo.netbox.dev. All devices was created in NewYork site.
1. Log-in to public demo site.
2. Locate NewYork devices: https://demo.netbox.dev/dcim/devices/?site_id=29
3. Add port connection from NY-Router from port et-0/0/1 to Panel1.1-NY port 1 and port 24.
image
4. Press create and get an error))
image

Crosses connection looks like this:
image

### Expected Behavior

I expect a new fancy connection to the patch panel.

### Observed Behavior

Server Error
<class 'AssertionError'>

Python version: 3.8.10
NetBox version: 3.7.1
Plugins: 
  netbox_demo: 0.3.1

DanSheps:
### Fixes: #15016 - Catch AssertionError from cable trace and throw ValidationError

* Catch any assertion errors thrown by the Cable save signal and raise a validation error
* Catch validation error in ObjectEditView and report error in UI while re-rendering form


netbox/dcim/models/cables.py
```diff
@@ -237,7 +237,10 @@ def save(self, *args, **kwargs):
                 if not termination.pk or termination not in b_terminations:
                     CableTermination(cable=self, cable_end='B', termination=termination).save()
 
-        trace_paths.send(Cable, instance=self, created=_created)
+        try:
+            trace_paths.send(Cable, instance=self, created=_created)
+        except ValidationError as e:
+            raise ValidationError(f'{e}')
```
jeremystretch, line 243: Why would we catch and re-raise the same exception? Was this meant to catch AssertionErrors?

netbox/dcim/models/cables.py
```diff
@@ -649,14 +657,16 @@ def from_origin(cls, terminations):
...
+                        _("All positions counts within the path on opposite ends of links must match")
+                    )
 
                     # Get our front ports
                     q_filter = Q()
                     for rt in remote_terminations:
                         position = positions.pop()
                         q_filter |= Q(rear_port_id=rt.pk, rear_port_position=position)
-                    assert q_filter is not Q()
+                    assert q_filter is not Q(), _("Remote termination query filter is empty, please open a bug report")
```
jeremystretch, line 669: Error messages should not directly prescribe a resolving action.

jeremystretch (changes requested):
The use of `assert` statements in this logic was primarily for sanity-checking during initial development; they shouldn't be relied upon as true validation checks. (They aren't even run when `__debug__` is False.)

If the use case in #15016 is not supported, we need to add specific validation logic to check for and report on it.

netbox/dcim/models/cables.py
DanSheps: That should have been an assertion error.  I changed this code a few times and must have accidentally changed it to ValidationError before commiting.

jeremystretch (changes requested):
I'm afraid this still doesn't address my primary concern: Validation of cable state needs to be performed _prior_ to the point where we're generating cable paths. (Again, the original assertions were just sanity checks.) This should ideally be performed on each model under `clean()`.

Additionally, as a rule generic views should not check for and handle any model-specific exceptions.

DanSheps:
> I'm afraid this still doesn't address my primary concern: Validation of cable state needs to be performed _prior_ to the point where we're generating cable paths. (Again, the original assertions were just sanity checks.) This should ideally be performed on each model under `clean()`.

The issue is there isn't really a way to validate the cable state prior to the cable being saved, without effectively duplicating all the work done in the cable path generation.

The main reason for this is because a lot of these assertion checks are dependant on sibling cable state further down the path, not just part of the existing cable save, which isn't known until you generate the full path.

We would effectively be running the from_origin twice in this instance, once to validate the cable state and once to save it.

netbox/dcim/models/cables.py
```diff
@@ -533,8 +534,8 @@ def from_origin(cls, terminations):
             return None
 
         # Ensure all originating terminations are attached to the same link
-        if len(terminations) > 1:
-            assert all(t.link == terminations[0].link for t in terminations[1:])
+        if len(terminations) > 1 and not all(t.link == terminations[0].link for t in terminations[1:]):
+            raise UnsupportedCablePath(_("All originating terminations must start must be attached to the same link"))
```
jsenecal, line 538: Please review the message content, the sentence is incorrect in its current state

jeremystretch (dismissed):

netbox/dcim/models/cables.py
DanSheps: Fixed this, thanks for the catch

netbox/netbox/views/generic/object_views.py
```diff
@@ -319,6 +320,12 @@ def post(self, request, *args, **kwargs):
                 form.add_error(None, e.message)
                 clear_events.send(sender=self)
 
+            # Catch any validation errors thrown in the model.save() or form.save() methods
+            except UnsupportedCablePath as e:
```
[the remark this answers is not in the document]
DanSheps: Should be ready for your review again, I think this is how you envisioned it.

DanSheps:
This is now ready.  I stupidly missed an UnsupportedCablePath in the test instead of AbortRequest

arthanson:
pinging here so it doesn't close

arthanson (changes requested):
Dan, can you please rebase to main

arthanson:
One more @DanSheps when I switch to this branch NetBox won't start with the following error - can you please sync and try:
```
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/urls.py", line 7, in <module>
    from account.views import LoginView, LogoutView
  File "/Users/ahanson/dev/work/netbox/netbox/account/views.py", line 28, in <module>
    from netbox.views import generic
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/views/generic/__init__.py", line 2, in <module>
    from .feature_views import *
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/views/generic/feature_views.py", line 12, in <module>
    from extras.forms import JournalEntryForm
  File "/Users/ahanson/dev/work/netbox/netbox/extras/forms/__init__.py", line 4, in <module>
    from .bulk_import import *
  File "/Users/ahanson/dev/work/netbox/netbox/extras/forms/bulk_import.py", line 184, in <module>
    class EventRuleImportForm(NetBoxModelImportForm):
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.12/site-packages/django/forms/models.py", line 312, in __new__
    fields = fields_for_model(
             ^^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.12/site-packages/django/forms/models.py", line 207, in fields_for_model
    raise FieldError(
django.core.exceptions.FieldError: 'action_object' cannot be specified for EventRule model form as it is a non-editable field
```

DanSheps:
Merged in main.  Started clean for me either way but hopefully this corrects whatever you are hitting.

arthanson (changes requested):
dcim.model.cables.CablePath.from_origin() is called from dcim.utils.create_cablepath which is called from wireless.signals.update_connected_interfaces in a **post_save** handler.  This  would be an issue as it could raise an error post_save so it would already be saved in this incorrect state.

I agree with Jeremy's prior comment that I think we need to validate it before save. Seems like we would want to create the connection graph in memory and then validate that... which sounds like a big pain.

DanSheps:
> I agree with Jeremy's prior comment that I think we need to validate it before save. Seems like we would want to create the connection graph in memory and then validate that... which sounds like a big pain.

Short of re-writing the whole thing, I don't know if we can write it in-memory as-is due to some of the queries that are run against it.

The signals will all fire inside of a transaction which should be rolled back automatically if there is an error.  I don't think there is a data consistency concern here, it would be treated the same as if the exception is raise directly in save (the transaction would roll back and you would be presented with the form again).

arthanson:
You are right on the transaction prevent save, but the raised error won't be caught in this case - to make it easy to test I just put into wireless.signals to force the raise:
```
@receiver(post_save, sender=WirelessLink)
def update_connected_interfaces(instance, created, raw=False, **kwargs):
    """
    When a WirelessLink is saved, save a reference to it on each connected interface.
    """
    from dcim.exceptions import UnsupportedCablePath
    raise UnsupportedCablePath("All mid-span terminations must have the same parent object")
```
Then created a wireless link, resulting in
image

So it's about the same as the assert in this specific case.  Not sure if that is a separate issue.

DanSheps:
I think catching that is a separate issue and I think we can wrap that in this PR as it is similar.

Let me go add that and see if I can adjust to catch that as well.

arthanson:
also, do you have an easier repro scenario for the original cable error?

DanSheps:
> also, do you have an easier repro scenario for the original cable error?

Unfortunately no

arthanson (approved):

PR #16403, opened 2024-06-04
16307 - Restore ability for reports to be able to log_success without generating a message

### LINKED ISSUE #16307: Restore ability for scripts to be able to log_success without generating a message
### NetBox version

v4.0.3

### Feature type

### Proposed functionality

Make `log_success(None, None)` and/or `log_success("", None)` increment the success counter without generating a log message

### Use case

Reports in Netbox <= v3.7 were able to do this:

```
self.log_success(obj)
```

This would increment the success counter for the report without generating any message: e.g.

image

However, in Netbox 4.0, when upgrading a Report to a Script, you need to change it to:

```
self.log_success("", obj)
or
self.log_success(None, obj)
or
self.log_success(None, None)
```

All of these log a message: a dash in the first case, or the string "None" in the second and third. (The third log line has no URL).

Either way, if you have thousands of successful checks, this generates thousands of log lines to wade through.

I note from the documentation of v3.7 that omitting the message for `log_success()` in Reports was explicitly permitted, and defaulted to `None`:

```
* log(message)
* log_success(object, message=None)
* log_info(object, message)
* log_warning(object, message)
* log_failure(object, message)
```

Maybe also `log_success()` with no arguments should be allowed, since the object identity isn't actually being used if no message is being recorded. It would be equivalent to `log_success(None, None)`

-----

Possibly relates to #8984, although I think that this feature request is orthogonal.

### Database changes

None

### External dependencies

None

Julio-Oliveira-Encora:
### Fixes: #16307

It was added to bypass the log message when the status is success, and the message is None.

It was tested using `log_success(None, None)`,  `log_success("", None)`, and `log_sucess()`.


netbox/extras/signals.py
```diff
@@ -118,9 +118,7 @@ def handle_changed_object(sender, instance, **kwargs):
 
 @receiver(pre_delete)
 def handle_deleted_object(sender, instance, **kwargs):
-    """
-    Fires when an object is deleted.
-    """
+    """Fires when an object is deleted."""
```
arthanson, line 121: Can you remove this from the PR?  Nothing to do with the chagne.

netbox/extras/views.py
```diff
@@ -1195,6 +1195,7 @@ def get_table(self, job, request, bulk_actions=True):
                     tests = job.data['tests']
 
                 for log in job.data['log']:
+                    print(log)
```
arthanson, line 1198: Remove this print statement

arthanson (changes requested):

arthanson:
@Julio-Oliveira-Encora it looks like log_success() works now with this change, can you please add a note to the documentation on scripts about this.

arthanson (approved):

jeremystretch (changes requested):
@Julio-Oliveira-Encora have you tested this before submitting the PR? You haven't actually updated [the method's signature](https://github.com/netbox-community/netbox/blob/eb3d4230778f26fa5be38c7d6b3e1205f905fc93/netbox/extras/scripts.py#L513) to make `message` an optional argument. When calling `log_success()` in a script, an exception is raised:

```
  File "/home/jstretch/projects/netbox/netbox/extras/scripts.py", line 658, in _run_script
  File "/home/jstretch/projects/netbox/netbox/scripts/test2.py", line 19, in run
    self.log_success()
TypeError: BaseScript.log_success() missing 1 required positional argument: 'message'
```

Julio-Oliveira-Encora:
I tested it before submitting the PR. I will check the method's signature and test again.

jeremystretch (changes requested):
I'm unclear on what the intended goal is here. As it stands, calling `log_success(obj=thing)` in a custom script does not generate any output. This seems completely pointless, and confusing.

#16307 seems limited to incrementing success counters in _reports_ only; if that's the case, the `log_*()` methods on the Report class should be modified to call `self._log()` directly rather than proxying through the `log_*()` methods on Script, to avoid changing them.

Julio-Oliveira-Encora:
@jeremystretch About the log methods on the Report class, should I change them this way?
   ```
 def log_success(self, obj=None, message=None):
        self._log(message, obj, level=LogLevelChoices.LOG_SUCCESS)
```
Or follow the argument positional from scripts, where `message` is before `obj`?
   ```
 def log_success(self, message, obj=None):
        self._log(message, obj, level=LogLevelChoices.LOG_SUCCESS)
```

Julio-Oliveira-Encora:
We could discuss it more because the user is asking for the change for Reports, but maybe he wants to use it in all types of scripts.

jeremystretch:
@Julio-Oliveira-Encora I think we went down the wrong path with this a bit. I've submitted an alternate PR (#16697) which I believe addresses the proposal in #16307.

PR #16406, opened 2024-06-04
Fixes #16376: Log changes on terminating objects when attaching a cable

### LINKED ISSUE #16376: Change log for terminating object is not updated when a cable is connected
### Deployment Type

NetBox Cloud

### NetBox Version

v4.0.3

### Python Version

3.12

### Steps to Reproduce

1. Create a device with two interfaces
2. Connect a cable between these two interfaces
3. Edit the description of the first interface (merely to trigger the creation of a new change record)
4. View the change log for the first interface

### Expected Behavior

The interface's change log should contain three entries:

1. The creation of the interface
2. The association of the interface with the cable (the `cable` and `cable_end` fields)
3. The modification of the `description` field

### Observed Behavior

Only two entries exist in the change log: There is no record indicating the change of the associated cable. In the first entry, the `cable` and `cable_end` fields are empty; in the second, they have been populated.

jeremystretch:
### Fixes: #16376

Call `snapshot()` and `save()` on the terminating object when creating a CableTermination, to ensure a changelog record is generated.

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -355,11 +355,11 @@ class CableTermination(ChangeLoggedModel):
         super().save(*args, **kwargs)
 
         # Set the cable on the terminating object
-        termination_model = self.termination._meta.model
-        termination_model.objects.filter(pk=self.termination_id).update(
-            cable=self.cable,
-            cable_end=self.cable_end
-        )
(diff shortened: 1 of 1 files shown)


jeremystretch:
screenshot

arthanson (approved):

Issue #7537, opened 2021-10-13
Serial number field for Virtual machines


pmusolino-rms:
### NetBox version

v3.0.7

### Feature type

Data model extension

### Proposed functionality

Would it be possible to add a serial number field to virtual machines?  I know it could certainly be added as a custom field, but many network devices as they exist in either public/private cloud or vm clusters still retain serial numbers for licensing and seems to be something that could be a standard but not required field.

### Use case

Exporting data and inventory/licensing tracking of non physical devices.

### Database changes

New field to Virtual machine table

rodvand:
I can see a use case for this. More and more functionality is moving from hardware appliances to virtual appliances and the vendors tend to stick to their "everything needs a license" model. So my virtual Palo Alto firewall has a serial number, and I can see the value in recording this. Not sure if the use case is "wide" enough to warrant a separate field encompassing all virtual machines, but I've definitely thought about it.

aauerswald:
Hello,

please add the serial field to virtual machines. At the moment we use a custom field, but this is not searchable.

ryanmerolle:
This could be a good candidate for 3.6

tagur87:
I also agree that this would be a helpful field for virtual machines. We are starting to migrate our VPN devices from physical hardware to virtual appliances. These virtual appliances (Ivanti Connect Secure) have serial numbers that are used for registration with the vendor. We have other vendors that also do this...(kemp, clearpass, etc).

I would be willing to try to add this field, since it seems fairly straightforward to add to the VM model.

tagur87:
Anyone opposed to aiming at 3.7 for a milestone? I worked up a simple prototype yesterday mirroring the field used in devices. 

Any concerns with me working on that?

PR #16407, opened 2024-06-04
7537 add serial number to virtual machines

arthanson:
### Fixes: #7537

Added serial number field, kept it as non-unique?  Wasn't sure if multiple vendors would have conflicting serial numbers if they aren't UUID, can change to unique if needed?

Didn't add to bulk edit as it didn't make sense to set multiple ones the same.


ITJamie:
I could see Serials being set to unique for a specific cluster. but non-unique outside the cluster

jeremystretch (changes requested):
Missing a few items:
- Add the serial field to the VM template
- Add the serial field to the search indexer

(Remember there's a [checklist](https://netboxlabs.com/docs/netbox/en/stable/development/extending-models/) for adding new fields to models.)

netbox/virtualization/models/virtualmachines.py
```diff
@@ -127,6 +127,11 @@ class VirtualMachine(ContactsMixin, ImageAttachmentsMixin, RenderConfigMixin, Co
         null=True,
         verbose_name=_('disk (GB)')
     )
+    serial_number = models.CharField(
```
jeremystretch, line 130: I just noticed that we've named the field `serial_number` here but it's called `serial` on the Device and Rack models. IMO we should keep the name consistent across all models to avoid confusion. (Apologies for not registering this earlier.)

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #16412, opened 2024-06-04
15156 Add display_url to REST API

### LINKED ISSUE #15156: Add URL field on API responses that provides the NetBox UI URL to an object
### NetBox version

4.0.2

### Feature type

New functionality

### Proposed functionality

When using the Rest API the URL field on objects has direct API URLs.
It would be great if there was a `display_url` or `ui_url` to use when referencing the object in external tools.

Removing `/api` from the the existing `url` field on API response doesn't always work to get to a valid UI URL.

This could be an optional non-default field that could be requested if needed now that #15087 is possible.

examples of what this could look like:

VM API response:
```
{
    "id": 361,
    "url": "https://demo.netbox.dev/api/virtualization/virtual-machines/361/",
    "ui_url": "https://demo.netbox.dev/virtualization/virtual-machines/361/",
    "display": "vm1",
    "name": "vm1",
 <truncated>
 ```
 
 VM Disk response:
```
{
    "id": 1,
    "url": "https://demo.netbox.dev/api/virtualization/virtual-disks/1/",
    "ui_url: "https://demo.netbox.dev/virtualization/disks/1/",
    "display": "disk1",
 <truncated>
 ```
 
 

### Use case

When using the API sometimes we send slack alerts or emails for specific issues we find with data for a human to visit and interfact with / fix.

With most objects we can strip `/api` from the URL and give that URL to users, however not all objects follow that (e.g. virtual disks on VMs)

Example:
- UI URL: `http://netbox_instance/virtualization/disks/1/`
- API URL: `http://netbox_instance/api/virtualization/virtual-disks/1/`

### Database changes

N/A

### External dependencies

N/A

arthanson:


jeremystretch (changes requested):
Can we subclass `HyperlinkedIdentityField` to define a custom field which dynamically resolves the view name from the serializer's model? We could then include this field on `BaseModelSerializer`, saving a ton of boilerplate. It could likewise replace the existing `url` field on all serializers too.

netbox/netbox/api/serializers/base.py
```diff
@@ -36,6 +56,15 @@ def __init__(self, *args, nested=False, fields=None, **kwargs):
...
 
+        # don't override the field if the class already defines these so can set lookup_field
+        if ("url" in self.fields and not isinstance(self.fields["url"], serializers.HyperlinkedIdentityField) and
+                isinstance(self.fields["url"], serializers.RelatedField)):
+            self.fields["url"] = NetBoxApiHyperlinkedIdentityField(self.Meta.model)
+        if ("display_url" in self.fields and not
+                isinstance(self.fields["display_url"], serializers.HyperlinkedIdentityField) and
+                isinstance(self.fields["display_url"], serializers.RelatedField)):
+            self.fields["display_url"] = NetBoxUrlHyperlinkedIdentityField(self.Meta.model)
+
```
jeremystretch, line 67: Instead of this, we should be able to override the `get_url()` method on our both of our HyperlinkedIdentityField subclasses to infer the model from the serializer (`self.model.parent`) automatically. (We can pass a dummy value to `view_name` to avoid the assertion checks.)

jeremystretch (changes requested):
Nice work!

netbox/dcim/api/serializers_/cables.py
```diff
@@ -42,16 +42,16 @@ class TracedCableSerializer(serializers.ModelSerializer):
     Used only while tracing a cable path.
     """
     url = serializers.HyperlinkedIdentityField(view_name='dcim-api:cable-detail')
+    display_url = serializers.HyperlinkedIdentityField(view_name='dcim:cable')
```
jeremystretch, line 45: This should be updated to inherit from BaseModelSerializer, and add a `display` field. (Or, alternatively, leaving it in its original state, without adding a `display_url` field.)

netbox/extras/api/nested_serializers.py
```diff
@@ -22,99 +22,89 @@
...
 
     class Meta:
         model = models.SavedFilter
-        fields = ['id', 'url', 'display', 'name', 'slug']
+        fields = ['id', 'url', 'display_url', 'display', 'name', 'slug']
 
 
 class NestedBookmarkSerializer(WritableNestedSerializer):
-    url = serializers.HyperlinkedIdentityField(view_name='extras-api:bookmark-detail')
+    display_url = serializers.CharField(allow_null=True, read_only=True)
```
jeremystretch, line 88: We should omit `display_url` for models which don't have a dedicated UI view.

netbox/extras/api/nested_serializers.py
```diff
@@ -22,99 +22,89 @@
...
 
     class Meta:
         model = models.Bookmark
-        fields = ['id', 'url', 'display', 'object_id', 'object_type']
+        fields = ['id', 'url', 'display_url', 'display', 'object_id', 'object_type']
 
 
 class NestedImageAttachmentSerializer(WritableNestedSerializer):
-    url = serializers.HyperlinkedIdentityField(view_name='extras-api:imageattachment-detail')
+    display_url = serializers.CharField(allow_null=True, read_only=True)
```
jeremystretch, line 96: Omit

netbox/extras/api/serializers_/objecttypes.py
```diff
@@ -10,7 +10,8 @@
 
 class ObjectTypeSerializer(BaseModelSerializer):
     url = serializers.HyperlinkedIdentityField(view_name='extras-api:objecttype-detail')
+    display_url = serializers.CharField(allow_null=True, read_only=True)
```
jeremystretch, line 13: Omit

netbox/tenancy/api/nested_serializers.py
```diff
@@ -47,36 +45,33 @@ class Meta:
...
 
     class Meta:
         model = Contact
-        fields = ['id', 'url', 'display', 'name']
+        fields = ['id', 'url', 'display_url', 'display', 'name']
 
 
 class NestedContactAssignmentSerializer(WritableNestedSerializer):
-    url = serializers.HyperlinkedIdentityField(view_name='tenancy-api:contactassignment-detail')
+    display_url = serializers.CharField(allow_null=True, read_only=True)
```
jeremystretch, line 71: Omit

netbox/tenancy/api/serializers_/contacts.py
```diff
@@ -19,45 +19,43 @@
...
-            'comments', 'tags', 'custom_fields', 'created', 'last_updated',
+            'id', 'url', 'display_url', 'display', 'group', 'name', 'title', 'phone', 'email', 'address', 'link',
+            'description', 'comments', 'tags', 'custom_fields', 'created', 'last_updated',
         ]
         brief_fields = ('id', 'url', 'display', 'name', 'description')
 
 
 class ContactAssignmentSerializer(NetBoxModelSerializer):
-    url = serializers.HyperlinkedIdentityField(view_name='tenancy-api:contactassignment-detail')
+    display_url = serializers.CharField(allow_null=True, read_only=True)
```
jeremystretch, line 58: Omit

netbox/netbox/api/serializers/base.py
jeremystretch, line 1: Can we move BaseNetBoxHyperlinkedIdentityField and its two subclasses to a separate module, `fields.py`, just to keep things a bit more organized?

netbox/netbox/api/serializers/base.py
```diff
@@ -13,7 +13,48 @@
...
+class NetBoxURLHyperlinkedIdentityField(BaseNetBoxHyperlinkedIdentityField):
+
+    def get_view_name(self, app_name, model_name):
+        return f'{app_name}:{model_name}'
+
+
 class BaseModelSerializer(serializers.ModelSerializer):
+    url = NetBoxAPIHyperlinkedIdentityField(view_name="")
+    display_url = NetBoxURLHyperlinkedIdentityField(view_name="")
```
jeremystretch, line 57: Since we're using a custom class anyway, let's override its `__init__()` method to avoid having to pass a dummy value for `view_name`.

netbox/netbox/api/serializers/base.py
```diff
@@ -13,7 +13,48 @@
 )
 
 
+class BaseNetBoxHyperlinkedIdentityField(serializers.HyperlinkedIdentityField):
```
jeremystretch, line 16: The parent class should define a `get_view_name()` method. This could raise a NotImplementedError, or it could obviate the need for a separate NetBoxURLHyperlinkedIdentityField class.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #16420, opened 2024-06-05
Closes #15908: Establish canonical & local sources for release info

### LINKED ISSUE #15908: Replace `settings.VERSION` with a set of release data
### Proposed Changes

Introduce a canonical module (e.g. `netbox/edition.py`) containing the relevant metadata for each release, including:

* NetBox edition (Community, Cloud, or Enterprise)
* NetBox version
* Release date
* Beta/experimental designation
* Anything else that seems relevant

This object would replace the `VERSION` variable currently defined in `settings.py`.

### Justification

This approach enables us to convey more information about each release.

jeremystretch:
### Fixes: #15908

- Establish locations where both canonical & local release attributes are defined
- Introduce the ReleaseInfo dataclass for conveying this information
- Update the base template footer

### DIFF
--- a/netbox/release.yaml
+++ b/netbox/release.yaml
@@ -1,0 +1,2 @@
+version: "4.1.0"
+designation: "dev"
--- a/netbox/core/management/commands/nbshell.py
+++ b/netbox/core/management/commands/nbshell.py
@@ -18,7 +18,7 @@ BANNER_TEXT = """### NetBox interactive shell ({node})
     node=platform.node(),
     python=platform.python_version(),
     django=get_version(),
-    netbox=settings.VERSION
+    netbox=settings.RELEASE.name
 )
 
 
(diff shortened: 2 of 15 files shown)


netbox/release.yaml
```diff
@@ -0,0 +1,2 @@
+version: "4.1.0"
```
arthanson, line 1: This is fine, but wondering if we should just use pyproject.toml as it should already have the version info in it - the designation would just be from the version string - I think there are standard libraries to parse the versioning.

arthanson (approved):
just the one extraneous import that I think isn't needed.

netbox/release.yaml
arthanson: yes, but not set on it - whatever is easiest for Nat to work with.

jeremystretch: AFAICT `importlib` can only glean metadata from a project that's actually been _installed_ as a package; it can't be used for local development. I feel like this is leading to a much deeper discussion around packaging and distribution, which I'm happy to have but I'd prefer not to block this particular FR.

arthanson: okay, just keep it in the yaml file then.

markkuleinio:
Note that this didn't address `upgrade.sh`

PR #16426, opened 2024-06-06
16256 - Allow alphabetical ordering of bookmarks on dashboard

### LINKED ISSUE #16256: Allow alphabetical ordering of bookmarks on dashboard
### NetBox version

v4.0.2

### Feature type

### Proposed functionality

Add additional order by options for bookmarks on the dashboard. 

### Use case

Alphabetical ordering is useful, and also matches the default table ordering of other sections.

### Database changes

_No response_

### External dependencies

_No response_

Julio-Oliveira-Encora:
### Fixes: #16256

Added alphabetical ordering option for bookmarks.


arthanson (approved):

netbox/extras/dashboard/widgets.py
```diff
@@ -381,7 +381,12 @@ def render(self, request):
         if request.user.is_anonymous:
             bookmarks = list()
         else:
-            bookmarks = Bookmark.objects.filter(user=request.user).order_by(self.config['order_by'])
+            if self.config['order_by'] == BookmarkOrderingChoices.ORDERING_NAME_AZ:
+                bookmarks = sorted(Bookmark.objects.filter(user=request.user),
```
jeremystretch, line 385: We should define the base queryset (`Bookmark.objects.filter(user=request.user)`) once, and then effect ordering one way or the other depending on the selected option. This ensures that any future modifications to the base queryset are needed only once.

netbox/extras/choices.py
```diff
@@ -117,10 +117,12 @@ class BookmarkOrderingChoices(ChoiceSet):
 
     ORDERING_NEWEST = '-created'
     ORDERING_OLDEST = 'created'
+    ORDERING_NAME_AZ = 'name'
 
     CHOICES = (
         (ORDERING_NEWEST, _('Newest')),
         (ORDERING_OLDEST, _('Oldest')),
+        (ORDERING_NAME_AZ, _('Name (A-Z)')),
```
jeremystretch, line 125: This ordering isn't technically by name, as not all model have a `name` attribute (circuits, for instance). Let's change this to "alphabetical" to be more accurate.

jeremystretch, line 125: I'm not sure how useful it would be, but it feels like we should add an option for reverse alphabetical ordering as well for completeness.

jeremystretch (changes requested):

jeremystretch (approved):

PR #16434, opened 2024-06-06
14692 Convert VM disk size

### LINKED ISSUE #14692: Allow disk size to be smaller than 1GB

arthanson:
### Fixes: #14692 

Monosnap Virtual Machines | NetBox 2024-06-06 11-45-36
Monosnap vm2 | NetBox 2024-06-06 11-46-55


netbox/virtualization/migrations/0039_convert_disk_size.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 5.0.6 on 2024-06-06 17:46
+
+from django.db import migrations
+from django.db.models import F
+
+
+def convert_disk_size(apps, schema_editor):
+    VirtualMachine = apps.get_model('virtualization', 'VirtualMachine')
+    VirtualMachine.objects.update(disk=F('disk') * 1000)
```
jeremystretch, line 9: Can we exclude records with null values, just for efficiency?
```suggestion
    VirtualMachine.objects.filter(disk__isnull=False).update(disk=F('disk') * 1000)
```

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #16435, opened 2024-06-06
16423 - Inconsistency Swagger schema vs API (model Interface)

### LINKED ISSUE #16423: Inconsistency Swagger schema vs API (model Interface)
### Deployment Type

Self-hosted

### NetBox Version

v4.0.3

### Python Version

3.11

### Steps to Reproduce

1. create Manufacturer, Device type, Device role, Site and Device+Interface
2. call GET to /api/dcim/interfaces/$InterfaceID/

### Expected Behavior

return property "device" as "model.NestedDevice" in Swagger schema

### Observed Behavior

by Swagger schema object type of "device" is "model.Device", but API return "model.NestedDevice"

Julio-Oliveira-Encora:
### Fixes: #16423
Changed `nested` to `false` for `device` in `InterfaceSerializer` in device_components.py.


arthanson:
@Julio-Oliveira-Encora can you please fix the tests that are failing.

Julio-Oliveira-Encora:
> @Julio-Oliveira-Encora can you please fix the tests that are failing.

Fixed.

arthanson:
@Julio-Oliveira-Encora I think this is backwards - I don't think we want to change the serializer. The serializer should be returning the nested object.  We want to change the REST schema to correctly reflect that the nested object is being returned.

Issue #15106, opened 2024-02-09
Wireless Link Distance


Nexis81:
### NetBox version

v3.7.2

### Feature type

Change to existing functionality

### Proposed functionality

Add a field for wireless links to show the distance between wireless links that will appear on a trace similar to a cable length trace. 

### Use case

When doing a cable trace on a wired connection that runs through a wireless link, The total cable length stops when it reaches a wireless link. It would be useful to show the complete wired length on both sides of the wireless link and total including the wireless link distance. 

![image](https://github.com/netbox-community/netbox/assets/9299655/1dae6080-81f6-4fd2-956c-906333b99a97)

### Database changes

Table: wireless_wirelesslink
Add Columns: length - numeric(8,2), length_unit - character varying(50), _abs_length - numeric(10,4)

PR #16528, opened 2024-06-12
15106 Add Length Field to Wireless Link

arthanson:
### Fixes: #15106 

Screenshot 2024-06-12 at 11 56 13 AM


docs/models/wireless/wirelesslink.md
```diff
@@ -40,3 +40,7 @@ The security cipher used to apply wireless authentication. Options include:
 ### Pre-Shared Key
 
 The security key configured on each client to grant access to the secured wireless LAN. This applies only to certain authentication types.
+
+### Length
```
jeremystretch, line 44: IMO "length" doesn't really make sense here, as there's no physical object. "Distance" feels much more natural. FR #15106 proposes "length" for the field name but its title refers to "distance." I realize renaming the fields essentially requires a find-and-replace on the whole PR, but I think it's worthwhile to use the most appropriate terminology.

netbox/wireless/models.py
```diff
@@ -224,6 +245,15 @@ def clean(self):
             })
 
     def save(self, *args, **kwargs):
+        # Store the given length (if any) in meters for use in database ordering
+        if self.length is not None and self.length_unit:
+            self._abs_length = to_meters(self.length, self.length_unit)
+        else:
+            self._abs_length = None
```
jeremystretch, line 252: We should move this logic into a utility function that can be employed by both WirelessLink and Cable. Something like

```python
self._abs_length = to_absolute_length(self.length, self.length_unit)
```

netbox/wireless/models.py
```diff
@@ -224,6 +245,15 @@ def clean(self):
             })
 
     def save(self, *args, **kwargs):
```
jeremystretch, line 247: `clean()` also needs to validate `length` & `length_unit`, as on Cable:

```python
        # Validate length and length_unit
        if self.length is not None and not self.length_unit:
            raise ValidationError(_("Must specify a unit when setting a cable length"))
```

netbox/wireless/forms/model_forms.py
```diff
@@ -181,3 +181,8 @@ class Meta:
             'auth_type': 'Type',
             'auth_cipher': 'Cipher',
         }
+        error_messages = {
+            'length': {
+                'max_value': _('Maximum length is 32767 (any unit)')
+            }
+        }
```
jeremystretch, line 188: I see that this was copied from Cable, but I don't believe it's necessary anymore. (`length` and `_abs_length` are decimals fields.) It should be removed from Cable as well.

netbox/wireless/tables/wirelesslink.py
```diff
@@ -36,6 +37,10 @@ class WirelessLinkTable(TenancyColumnsMixin, NetBoxTable):
         verbose_name=_('Interface B'),
         linkify=True
     )
+    length = columns.TemplateColumn(
+        template_code=WIRELESS_LINK_LENGTH,
+        order_by=('_abs_length', 'length_unit')
```
jeremystretch, line 42: Absolute length is stored in meters; it doesn't make sense to also order by the original unit. (I don't know why this is set on the column in CableTable; it should also be removed.)
```suggestion
        order_by=('_abs_length')
```

netbox/wireless/tables/wirelesslink.py
```diff
@@ -36,6 +37,10 @@ class WirelessLinkTable(TenancyColumnsMixin, NetBoxTable):
         verbose_name=_('Interface B'),
         linkify=True
     )
+    length = columns.TemplateColumn(
```
jeremystretch, line 40: We should replace these two TemplateColumns (here and on CableTable) with a new custom column to avoid the duplicated template code.

jeremystretch (changes requested):

netbox/wireless/tables/wirelesslink.py
arthanson: renamed fields so can no longer use the same

jeremystretch (approved):

PR #16545, opened 2024-06-13
Closes #16388: Move change logging resources from `extras` to `core`

### LINKED ISSUE #16388: Move all change logging resources from `extras` to `core`
### Proposed Changes

Move all resources related to change logging from the `extras` app to `core`. This will entail moving the following:

- The `ObjectChange` model
- Relevant views and API endpoints
- The `event_tracking()` context manager

### Justification

Although originally introduced as an optional "extra" feature in NetBox, change logging quickly become a core function of the application as it matured. It no longer makes sense for it to live under `extras`.

jeremystretch:
### Fixes: #16388

- Move the ObjectChange model & related resources (views, forms, tests, etc.) from the `extras` app to the `core` app
- Write migrations to rename the existing table and avoid altering existing data
- Move the `event_tracking()` context manager into `netbox`


arthanson (approved):
As separate Housekeeping - should we move object-types to core as well?  Seem they fit there?

PR #16552, opened 2024-06-13
Per-Script Worker Queue Setting

### LINKED ISSUE #16516: Per Script Worker Queue Setting
### NetBox version

v3.7.8 and v4.0.5

### Feature type

New functionality

### Proposed functionality

Looking to enhance the script subsystem to add support to specify what worker queue each script should be run in.  This will allow critical scripts to run before background or scheduled scripts.  This feature would add a new field in the scripts’ Meta possibly called “rq_queue_name” that can bypass the logic in get_queue_for_model and just use the user defined queue.  If “rq_queue_name” is not set then the current function can be called to determine what queue the script should be map too.

### Use case

We have a few CI/DI pipeline that hit the API to generate control files in a centralized location versus the logic be distributed across different project code.  This simplifies management of the project files but these critical scripts must wait in the queue with other background and user run scripts delaying the running of them. 

Some of our non-critical script that update data can take 5 minutes to run and they are not time critical like the CI/DI pipeline.

### Database changes

I don’t think any models will need to be updated to support this functionality. 

### External dependencies

n/a

jchambers2012:
### Fixes: #16516

This PR add a new Meta field to the scripts called `rq_queue_name`.  This new field will allow time critical scripts to route to the `high` queue and run before scripts that are designed to be background jobs and run at a lower priority.  If the script programmer tries to route the script to a non-existing queue, the current logic will route the script to the queue set to for scripts.

The PR was tested via the GUI and via the API.  This was assessed by creating 4 Custom scripts that each route to diffrent queues. Three scripts were set to route to the standard `high`, `default`, `low` queues and a fourth was set to route to a queue called `None` that did not exist in configuration.  The three scripts that had queue ran in the correct order with high script running first, followed by default then low.  The fourth script that routed to `None`  was redirected to the default queue.

### GUI Tests:
image

image

### API Tests:
image

### Script Used for testing
```python 
from extras.scripts import *
import time


class testLowQue(Script):
    class Meta:
        name = "Log Test  - Que Low"
        scheduling_enabled = False
        description = ""
        rq_queue_name = "low"

    def run(self, data, commit):
        for i in range(6):
            self.log_info(f"Log Test - Sleeping for 10 Sec - Que Low: {i}")
            time.sleep(10)
        return str("")


class testDefaultQue(Script):
    class Meta:
        name = "Log Test  - Que Default"
        scheduling_enabled = False
        description = ""
        rq_queue_name = "default"

    def run(self, data, commit):
        for i in range(6):
            self.log_info(f"Log Test - Sleeping for 10 Sec - Que Default: {i}")
            time.sleep(10)
        return str("")


class testHighQue(Script):
    class Meta:
        name = "Log Test  - Que high"
        scheduling_enabled = False
        description = ""
        rq_queue_name = "high"

    def run(self, data, commit):
        for i in range(6):
            self.log_info(f"Log Test - Sleeping for 10 Sec - Que High: {i}")
            time.sleep(10)
        return str("")


class testNoneQue(Script):
    class Meta:
        name = "Log Test  - Que None"
        scheduling_enabled = False
        description = ""
        rq_queue_name = "None"

    def run(self, data, commit):
        for i in range(6):
            self.log_info(f"Log Test - Sleeping for 10 Sec - Que None: {i}")
            time.sleep(10)
        return str("")
```


netbox/core/models/jobs.py
```diff
@@ -208,11 +213,21 @@ def enqueue(cls, func, instance, name='', user=None, schedule_at=None, interval=
             name: Name for the job (optional)
             user: The user responsible for running the job
             schedule_at: Schedule the job to be executed at the passed date and time
+            rq_queue_name: Queue name to route the job to for processing
             interval: Recurrence interval (in minutes)
         """
         object_type = ObjectType.objects.get_for_model(instance, for_concrete_model=False)
-        rq_queue_name = get_queue_for_model(object_type.model)
-        queue = django_rq.get_queue(rq_queue_name)
+        if rq_queue_name is None:
```
arthanson, line 220: You can refactor this logic to avoid the duplicate code, something like:

```
queue = None
if rq_queue_name:
    queue = django_rq.get_queue(...
if not queue:
    rq_queue_name = get_queue_for_model(...
    queue = django_rq.get_queue(...
```

arthanson (changes requested):

jchambers2012:
Updated the code: https://github.com/netbox-community/netbox/pull/16552/commits/5676c4abadc928ba00943dcd99329173c67c01ee

netbox/core/models/jobs.py
jchambers2012: Updated the code: https://github.com/netbox-community/netbox/commit/5676c4abadc928ba00943dcd99329173c67c01ee

jchambers2012:
Posting comment to prevent closure

docs/customization/custom-scripts.md
```diff
@@ -108,6 +108,10 @@ commit_default = False
 
 By default, a script can be scheduled for execution at a later time. Setting `scheduling_enabled` to False disables this ability: Only immediate execution will be possible. (This also disables the ability to set a recurring execution interval.)
 
+### `rq_queue_name`
+
+This will override the standard `QUEUE_MAPPINGS` setting for the script to be process by the worker.  If the queue is not found then the default logic found in the settings file will be used.
```
arthanson, line 113: Maybe mention here that you will need to add any queues to your rq service start

arthanson (changes requested):
Also, I'm not sure if this is missing something, what mechanism would be used to service the extra queue?  Would you need to make a plugin with the queues defined or should there be a change to settings/config.py to allow an additional queue to be defined?  Currently NetBox only defines / services the high, default,low queue.

jchambers2012:
> Also, I'm not sure if this is missing something, what mechanism would be used to service the extra queue? Would you need to make a plugin with the queues defined or should there be a change to settings/config.py to allow an additional queue to be defined? Currently NetBox only defines / services the high, default,low queue.

Ya this was mainly to support custom queue without the need to do a later PR to support those cases.  If we dont think they are heavily used by the user base then it could be removed and limited to High, Default, Low queue.

https://netboxlabs.com/docs/netbox/en/stable/plugins/development/background-tasks/

```Plugins can also add custom queues for their own needs by setting the queues attribute under the PluginConfig class. An example is included below:```

File: docs/customization/custom-scripts.md (hunk 2 above)
Comment (jchambers2012) [reply docs/customization/custom-scripts.md]: Done: https://github.com/netbox-community/netbox/pull/16552/commits/d34a87f5912c2073c5bc41db216f9435ab97b1dd

File: netbox/core/models/jobs.py [hunk 3]
```diff
@@ -208,11 +213,20 @@ def enqueue(cls, func, instance, name='', user=None, schedule_at=None, interval=
...
         """
         object_type = ObjectType.objects.get_for_model(instance, for_concrete_model=False)
-        rq_queue_name = get_queue_for_model(object_type.model)
-        queue = django_rq.get_queue(rq_queue_name)
+        queue = None
+        if rq_queue_name:
+            try:
+                queue = django_rq.get_queue(rq_queue_name)
+            except Exception:
+                # User defined queue casued an error - return to default logic
```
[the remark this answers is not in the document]
jchambers2012: Done: https://github.com/netbox-community/netbox/pull/16552/commits/d34a87f5912c2073c5bc41db216f9435ab97b1dd

docs/customization/custom-scripts.md
```diff
@@ -108,6 +108,10 @@ commit_default = False
 
 By default, a script can be scheduled for execution at a later time. Setting `scheduling_enabled` to False disables this ability: Only immediate execution will be possible. (This also disables the ability to set a recurring execution interval.)
 
+### `rq_queue_name`
```
alehaa, line 111: In #16927 I'm introducing a general purpose job API used by NetBox and made available for plugins. A minor background for this is abstraction to decouple plugins from using Redis directly, making future changes possible. Therefore I'd like to suggest not using parameter names related to Redis queues, i.e. just using something like `queue` instead of `rq_queue_name`.

jchambers2012: I'm good with whatever Netbox wants to use for name, be it `queue`, `worker_queue` or `job_queue`.  But I won't make any changes right now till the Netbox maintainers pick the language that is preferred for this.

jchambers2012:
@arthanson please see https://github.com/netbox-community/netbox/pull/16552#discussion_r1685148598 and let me know if there is a preference here on what the variable should be called?

jeremystretch:
I like the general idea, but I'm not sure explicitly specifying a queue name inside each script is the optimal approach. For one, it could become highly repetitive when dealing with a large number of scripts. Second, renaming the queues would require modifying each individual script.

IMO the use case would be better served by the ability to configure a policy mapping scripts (by module, name, or some other attribute) to the desired queues. This would allow for more flexibility to adjust the mapping over time without needing to modify individual scripts. A configuration parameter would probably suffice for this purpose.

Separately, we should probably also include the ability for the user to manually select the queue when executing a script (via the web UI or REST API), perhaps limiting the available options to a subset of approved queues.

jchambers2012:
@jeremystretch 

>Parden my ignorance if configuration is dynamically loaded each time there is a change while the application is running. 

Just to defend the pro-per-script setting.  At our shop, we need to do change control for making application-level changes and restarting of the platform as this would make it unavailable (even briefly) for the users and CI/DI pipelines. The outage could be longer if someone makes a typo in the config (IE not a restart but a start-debug-retry). We use Data files only for all our NetBox script so they go through their own CI/DI automated and manual testing and validation process.  So, to make this a “global“ configuration change for mappings, at least for our shop, would make it more paperwork to move the setting their vs just syncing in the scripts. 

I would also be an Anti-User Selected Queue setting as everyone will just think their job should be the highest priority, though it is not a bad idea to have the option available and exposed if needed.  Would you think this be a new per-script setting to enable queue selection like scheduling (`scheduling_enabled`) has a flag today? Something like `user_queue_selection_enables = True`?  This setting might also need to be a per-script option as the script writer may not want their script to have queue changes much like some job should not be scheduled and may/may not have the ability to make changes to the platform configuration to quickly flip this flag on/off.

PR #16553, opened 2024-06-13
Fixes 16536 - Fix filtering of device component by device role

### LINKED ISSUE #16536: Filtering device components per device role fails
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5

### Python Version

3.10

### Steps to Reproduce

1. Create a device role "role_1" and "role_2"
2. Create a device "device_1" with role "role_1" and a device "device_2" with "role_2"
3. Create an interface on device_1 and another on device_2
4. In the Interface list, use the Filters tab to filter on device role role_1

### Expected Behavior

Only interfaces of devices with role_1 are displayed.

### Observed Behavior

All the interfaces are displayed, the list is not filtered.

Alef-Burzmali:
### Fixes: #16536 

Fix the filtering of device components by device role by aligning the name of the fields from the DeviceComponentFilterForm and the DeviceComponentFilterSet.

I saw two solutions to implement the fix:
1. Rename the fields in the DeviceComponentFilterSet to `device_role_id` and `device_role` (that's the one I've implemented)
2. Rename the field in the DeviceComponentFilterForm to `role_id` but create an exception for InventoryItemFilterSet and InventoryItemFilterForm, which already have a field named `role` for the InventoryItemRole
 
I've chosen solution 1 because:
* We are not filtering according to the component's role but according to its parent device's role, so `device_role` makes more sense as `role` could be confused with the role of the component.
* It avoids creating an exception for InventoryItemFilterSet, which would have its `role` field override the role field from DeviceComponentFilterSet, and would have needed a `device_role` field or equivalent to preserve the feature. The role filter of DeviceComponent would have a different meaning between an Interface or a InventoryItem, which would have been more confusing.
* While the release note for 4.0 clearly states in breaking changes that the device_role field of devices has been removed after a depreciation period, there are no mentions of DeviceComponent in the notes or in #13342, so it may have been an undocumented or unintended change
* There is no equivalent filtering on VMInterface, it's not possible to filter them by their VM's role at the moment, so the reason behind the renaming of "device_role" to "role" is not currently present for DeviceComponent. However, if this feature is added, a more neutral field name such as "parent_role" or "parent_object_role" could be used instead.
* It's the least amount of changes required for this fix. Solution 2 would have required modifying InventoryItemFilterSet, InventoryItemFilterForm, DeviceComponentFilterForm and the fieldsets of all inheriting FilterForms.

However, solution 1 also changes the API to filter the device components by roles. This was an undocumented change in 4.0, so maybe that's acceptable.

Please let me know if you agree with that approach or if you prefer another solution.

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1378,12 +1378,12 @@ class DeviceComponentFilterSet(django_filters.FilterSet):
         to_field_name='model',
         label=_('Device type (model)'),
     )
-    role_id = django_filters.ModelMultipleChoiceFilter(
+    device_role_id = django_filters.ModelMultipleChoiceFilter(
         field_name='device__role',
         queryset=DeviceRole.objects.all(),
         label=_('Device role (ID)'),
     )
-    role = django_filters.ModelMultipleChoiceFilter(
+    device_role = django_filters.ModelMultipleChoiceFilter(
         field_name='device__role__slug',
         queryset=DeviceRole.objects.all(),
         to_field_name='slug',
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
This seems like the right way to go. Thanks @Alef-Burzmali!

PR #16627, opened 2024-06-17
Fixes #16624: Set allow_null=True on method fields that can return None

### LINKED ISSUE #16624: parent_device on Device is not marked as nullable in swagger api docs
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5

### Python Version

3.10

### Steps to Reproduce

1. Go to https://demo.netbox.dev/api/schema/swagger-ui/
2. Scroll down to `Device` Schema and look at the `parent_device` field

### Expected Behavior

It should contain an annotation `nullable: true`.

### Observed Behavior

It does not contain that annotation.

The fix for this should be to add `allow_null=True` to `SerializerMethodField` here: https://github.com/netbox-community/netbox/blob/c8aac13ceebf851368f6c7c2fbc6a783fc2e7bda/netbox/dcim/api/serializers_/devices.py#L60

Happy to put up a PR but it might be easier for a maintainer to just make the change and directly merge it.

fabi125:
### Fixes: #16624

Noticed that Device.parent_device was not marked as nullable in the swagger api docs. Went through all uses of `SerializerMethodField` and added `allow_null=True` if the method could return `None`.

I opted to add it to the `@extend_schema_field` decorator where possible because that's how `allow_null` was set on other methods that could return `None`.

### DIFF
--- a/netbox/dcim/api/serializers_/base.py
+++ b/netbox/dcim/api/serializers_/base.py
@@ -13,7 +13,7 @@ class ConnectedEndpointsSerializer(serializers.ModelSerializer):
     """
     Legacy serializer for pre-v3.3 connections
     """
-    connected_endpoints_type = serializers.SerializerMethodField(read_only=True)
+    connected_endpoints_type = serializers.SerializerMethodField(read_only=True, allow_null=True)
     connected_endpoints = serializers.SerializerMethodField(read_only=True)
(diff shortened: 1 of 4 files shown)


fabi125:
bump, because I still very much want to merge this

jeremystretch (approved):
Thanks @fabi125!

PR #16629, opened 2024-06-17
Fixes #16452: Fix sizing of buttons within object attribute panels

### LINKED ISSUE #16452: "Map"-button on site addresses with multiple lines is extended
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5

### Python Version

3.11

### Steps to Reproduce

When an address is added to a site object in NetBox and the address contains new lines, the **Map** button is messed up.

### Expected Behavior

In NetBox pre-version 4, the button height was normalized and in center.
image

### Observed Behavior

Button height is extended and the text is not in center.
image

jeremystretch:

### DIFF
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -28,7 +28,7 @@
                     </tr>
                     <tr>
                         <th scope="row">{% trans "Rack" %}</th>
-                        <td class="d-flex justify-content-between">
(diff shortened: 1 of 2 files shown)


arthanson (approved):

RasmusThing:
Thanks for the fix. May I suggest using **align-items-center** instead of **align-items-start**?
That way, the button will be centered in the column, just like in NetBox pre-version v4.x.x.

PR #16632, opened 2024-06-17
16460 remove spaces from telephone dialing

### LINKED ISSUE #16460: Phone numbers in tel: URLs contain spaces in contravention of RFC3966
### Deployment Type

Self-hosted

### NetBox Version

v4.0.3

### Python Version

3.11

### Steps to Reproduce

1. Set the phone number for a contact to a number with spaces, e.x. `+1 800 123 45 67`
2. Go to the contact detail page and hover over the phone number, and observe the link generated. (This also applies to the contact list table.)

### Expected Behavior

The URL generated should be `tel:+18001234567`, removing spaces as per [RFC3966 section 5.1.1](https://www.rfc-editor.org/rfc/rfc3966#section-5.1.1):

> However, even though ITU-T E.123 [[E.123](https://www.rfc-editor.org/rfc/rfc3966#ref-E.123)] recommends the use of space characters as visual separators in printed telephone numbers, "tel" URIs MUST NOT use spaces in visual separators to avoid excessive escaping.

### Observed Behavior

The URL generated is actually `tel:+1 800 123 45 67` which neither strips nor URL-encodes the spaces.

Inspecting the HTML code would show something like:

```
<a href="tel:+1 800 123 45 67">+1 800 123 45 67</a>
```

arthanson:
### Fixes: #16460 

Just remove spaces from link to dial phone number.  There is more to RFC3966 but removing any more characters can be country specific - we would probably want to use something like https://pypi.org/project/phonenumbers/ but removing space is safe and should fix most issues.

### DIFF
--- a/netbox/utilities/tables.py
+++ b/netbox/utilities/tables.py
@@ -29,7 +29,7 @@ def linkify_phone(value):
     """
     if value is None:
         return None
-    return f"tel:{value}"
+    return f"tel:{value.replace(' ', '')}"
 
 
 def register_table_column(column, name, *tables):


jeremystretch (approved):
Thanks @arthanson!

PR #16674, opened 2024-06-20
16424 - Allow filtering of Devices by Cluster and Cluster Group

### LINKED ISSUE #16424: Allow filtering of Devices by Cluster and Cluster Group
### NetBox version

v4.0.3

### Feature type

### Proposed functionality

Allow filtering Devices in the UI by Cluster and or Cluster Group.

### Use case

Being able to filter the device list view by members of a cluster group or individual cluster would be very useful.
It would allow for mass updating devices based on their clusters.

### Database changes

N/A

### External dependencies

N/A

Julio-Oliveira-Encora:
### Fixes: #16424

It was added to filtering Devices by Cluster Group and Cluster.


jeremystretch (changes requested):
@Julio-Oliveira-Encora could you please add filterset tests for these?

jeremystretch (approved):
Note that for models with a `slug` field, we need filters for both the ID and slug. I went ahead and added the second filter & updated the test. Thanks @Julio-Oliveira-Encora!

PR #16708, opened 2024-06-24
16704 - Define a default help_text for ColorField

### LINKED ISSUE #16704: Define a default `help_text` for ColorField
### Proposed Changes

Set a default `help_text` value on the ColorField model/form field.

### Justification

There are many places where we currently define a help text for these fields as

```
mark_safe(_('RGB color in hexadecimal. Example:') + ' <code>00ff00</code>')
```

Creating a a reusable form field for these will eliminate this redundant code.

Julio-Oliveira-Encora:
### Fixes: #16704 

Added `help_text` to ColorField.


jeremystretch (changes requested):
Ok, but the motivation for the change is that it allows us to remove the redundant `help_text` from all the forms where ColorField is used. Could you tackle that as well please?

jeremystretch (approved):

Issue #9627, opened 2022-06-29
VlanGroup - ArrayField rather than a start/end range


PieterL75:
### NetBox version

v3.2.5

### Feature type

Change to existing functionality

### Proposed functionality

a VLAN Group has the option to limit the vlans that can be used in that group.
This FR is to make the vlan-member a comma-separated value and allow more that one range or separate vlans in a vlan group.

### Use case

We have use cases where one group consists of 2 different ranges (1000-1099, 2500-2599 for example).
Currently we have to create a second group, or extend the vlangroup with vlans that should not be used in that group.

### Database changes

- Remove the min_vid, max_vid values
- Replace with a ArrayField of Vlan/VlanRanges ?

jeremystretch:
#9083 proposed something similar, but the transition to a single ArrayField seems more efficient.

jeremystretch:
Need to put a bit more thought into how best to implement this. We do something similar with rack reservation units, however in that case we're actually storing a discrete integer for each rack unit to which the reservation applies. This doesn't scale well to ~4K VLAN IDs. We may be able to leverage an array of PostgreSQL [range fields](https://docs.djangoproject.com/en/4.0/ref/contrib/postgres/fields/#range-fields), but further research is needed.

While I would have liked to pursue this in v3.3, we're running a bit behind with this release as it is.

jcralbino:
Hello Jeremy,
I wanted to know if this scenarios can be supported in this next iteration this range functionality. :
1. Creation of several vlan ranges, that can be assigned to the a vlan pool. 
2. Each vlan range, can be defined with a separate status field to be configurable ( active, reserved.. ) 
3. A vlan pool, can include several different ranges

I would say that the vlan range can be a separate model itself where additional information can also be added to this object, like comments, or even the change log to monitor changes in the range definition.

abhi1693:
Is it possible to take this up in v3.5?

jeremystretch:
It might be possible to employ an ArrayField of RangeFields.

jeremystretch:
Just noting for posterity: PostgreSQL 14 introduces native [multirange types](https://www.postgresql.org/docs/current/rangetypes.html), which would be ideal. However, we would need to drop support for PostgreSQL 12 and 13 to utilize these. Additionally, these types appear [unlikely to be supported by Django](https://code.djangoproject.com/ticket/33238), so we would need to device our own implementation for the ORM.

PR #16710, opened 2024-06-24
9627 use ArrayField on VLANGroup for allowed VLANs

arthanson:
### Fixes: #9627 

Note: The min_vid and max_vid filters were removed, It might be possible to do some PostGres fu to filter inside the arrays for min / max.  Could also make internal fields storing the array min and max and filter against that.  Tried storing the expanded array as individual items, but that runs into the PostGres row size limit so is not tenable.

Currently using `vlan_id_ranges` for the name, wondering if allowed_vlans or allowed_vlan_ids would be a better name?


netbox/ipam/models/vlans.py
```diff
@@ -45,23 +52,16 @@ class VLANGroup(OrganizationalModel):
...
-    min_vid = models.PositiveSmallIntegerField(
-        verbose_name=_('minimum VLAN ID'),
-        default=VLAN_VID_MIN,
-        validators=(
-            MinValueValidator(VLAN_VID_MIN),
-            MaxValueValidator(VLAN_VID_MAX)
-        ),
-        help_text=_('Lowest permissible ID of a child VLAN')
+    vlan_id_ranges = ArrayField(
+        BigIntegerRangeField(),
```
jeremystretch, line 56: This can be an `IntegerRangeField`; VLAN IDs are limited to 12 bits (1-4094).

netbox/ipam/models/vlans.py
```diff
@@ -45,23 +52,16 @@ class VLANGroup(OrganizationalModel):
...
+        verbose_name=_('min/max VLAN IDs'),
+        default=get_default_vlan_ids,
```
jeremystretch, line 58: Does this field need a default value?

netbox/ipam/forms/filtersets.py
jeremystretch, line 1: We need to devise a filter that matches VLAN groups which contain a given VLAN ID. The following SQL works:

```sql
SELECT * FROM ipam_vlangroup, unnest(vlan_id_ranges) vid_range WHERE 123::int8 <@ vid_range;
```

But we need to convert it to an ORM query.

netbox/templates/ipam/vlangroup.html
```diff
@@ -40,7 +40,7 @@ <h5 class="card-header">{% trans "VLAN Group" %}</h5>
         </tr>
         <tr>
           <th scope="row">{% trans "Permitted VIDs" %}</th>
-          <td>{{ object.min_vid }} - {{ object.max_vid }}</td>
+          <td>{{ object.vlan_ranges }}</td>
```
jeremystretch, line 43: These should be rendered in a list.

netbox/ipam/api/serializers_/vlans.py
```diff
@@ -21,6 +22,14 @@
 )
 
 
+class NumericRangeArraySerializer(serializers.CharField):
+    def to_internal_value(self, data):
+        return string_to_range_array(data)
+
+    def to_representation(self, instance):
+        return ranges_to_string(instance)
```
jeremystretch, line 30: I'd think we want to represent each range as an item within a list, to enable easy parsing.

jeremystretch (changes requested):

netbox/ipam/models/vlans.py
arthanson: It was returning an error on a comparison or something, forget the exact error.

netbox/ipam/forms/filtersets.py
arthanson: Add an annotate_vlan_ranges and queried off of that but returns a set error.

Researching, so it looks like https://forum.djangoproject.com/t/django-4-2-behavior-change-when-using-arrayagg-on-unnested-arrayfield-postgresql-specific/21547/4 there is an issue in Django that prevents the straight-forward solution, there is a ticket to support this https://code.djangoproject.com/ticket/35586 but not implemented yet.  Going to try variations of the Unnest function in the thread:
```
class Unnest(Func):
    subquery = True
    function = "unnest"
```

jeremystretch: > Add an annotate_vlan_ranges

The other problem with this method is that it results in duplicate VLANGroup objects unless we call `distinct()`. I've remove it as I don't _think_ it's needed.

jeremystretch: I've added the `contains_vid` filter which I think gets us what we want for now. It can likely be optimized for performance, but shouldn't be too bad.

arthanson:
@jeremystretch looks good to me

jeremystretch (approved):

PR #16713, opened 2024-06-24
16640 fix JSON custom field save nul

### LINKED ISSUE #16640: UI corrupts read-only JSON fields
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5

### Python Version

3.11

### Steps to Reproduce

This issue can be reproduced in a couple different workflows.  This was originally reported in #16328 after we saw this behavior, which we believe was introduced between 3.6.3 and 3.7.7.  We have now reproduced this in a clean install of 4.0.5 running inside docker using netbox-docker 2.9.1.

Edits through the API are not impacted.

Option 1 - corruption during edit:
1) Create a JSON custom field - set "UI Editable" to "No".  Leave the default value as null.
2) Create a prefix, it is not necessary to populate the JSON field with a value -- it will default to null
3) View the prefix, observe that the JSON custom field shows a value of "-" in the UI due to it being null
4) Edit the prefix and click "save" without making any changes
5) Observe that the custom value now changes to the quoted string "null"

Option 2 - corruption on creation:
1) Create a JSON custom field - set "UI Editable" to "No".  Set the default to JSON such as `{"key1":"value1"}`
2) Create a prefix
3) View the prefix, observe that the JSON custom field now shows an escaped string value of  `"{\"key1\": \"value1\"}"`
4) Edit the prefix and click "save" without making any changes
5) Observe that the custom value is now escaped again to something like `"\"{\\\"key1\\\": \\\"value1\\\"}\""`
6) Repeat edit+save (without making changes) and observe that the prefix keeps getting escaped more and more as a JSON string instead of any valid JSON data structure.

### Expected Behavior

The JSON value should not be changed due to the UI being set to read-only for this field

### Observed Behavior

The server is setting a new and incorrect/corrupt value for this field in the database.  The native JSON is being converted to an escaped string

arthanson:
### Fixes: #16640 

This will fix new saves, doesn't go back and fix old values - could set it to UI editable and edit it to fix.  If you have saved multiple times then it will keep escaping so not sure if more dangerous to auto-fix with a validate as JSON and if not valid then revert to ''... 

It looks like as it is UI_EDITABLE = FALSE the field isn't getting converted in clean as a JSON field, this works if UI_EDITABLE = True and the field is returned as a dict, if it is not UI editable the fields get returned as a string.


netbox/netbox/forms/base.py
```diff
@@ -39,7 +39,7 @@ def _get_form_field(self, customfield):
             form_field = customfield.to_form_field(set_initial=False)
             initial = self.instance.custom_field_data.get(customfield.name)
             if customfield.type == CustomFieldTypeChoices.TYPE_JSON:
-                form_field.initial = json.dumps(initial)
+                form_field.initial = json.dumps(initial) if initial else ''
```
jeremystretch, line 42: I think we need to check against None explicitly, because e.g. `0` would be a valid value.
```suggestion
                form_field.initial = json.dumps(initial) if initial is not None else ''
```

jeremystretch (changes requested):

netbox/netbox/forms/base.py
arthanson: But that wouldn't be valid JSON, I think you have to have key:value for valid JSON which would never evaluate to False?

jeremystretch: AFAIK you can store arbitrary scalars like `123` or `"foo"` in the field too.

FliesLikeABrick: Hi all, I am the reported of the original issue.  This PR fixes the issue when the field is null, but existing data is still ending up escaped.

FliesLikeABrick:
Hi all, I am the reported of the original issue. This PR appears to fix the issue for null values only.  Editing existing values, or creating a prefix with a default value both lead to escaped strings being placed in the field

jeremystretch:
@FliesLikeABrick could you please give instructions for reproducing a case which the PR does not address?

FliesLikeABrick:
> @FliesLikeABrick could you please give instructions for reproducing a case which the PR does not address?

Any non-null custom value is still corrupted via either of the instructions in the original problem report, such as :

     

Option 2 - corruption on creation:

    Create a JSON custom field - set "UI Editable" to "No". Set the default to JSON such as {"key1":"value1"}
    Create a prefix
    View the prefix, observe that the JSON custom field now shows an escaped string value of "{\"key1\": \"value1\"}"
    Edit the prefix and click "save" without making any changes
    Observe that the custom value is now escaped again to something like ""{\"key1\": \"value1\"}""

arthanson:
@FliesLikeABrick Can you please try the updated PR, I think it should fix the problem for newly created custom fields.  It won't change those that are already have incorrect data.

FliesLikeABrick:
@arthanson  @jeremystretch  this PR appears to pass our tests for reproducing the issue, thank you

FliesLikeABrick:
Understood that this does not fix existing corrupt/damaged/impacted data - I have written a tool at https://github.com/FliesLikeABrick/netbox_fix_json which can be used to repair impacted data, for anyone else who comes across this.

rbcollins123:
Any thoughts on when this will be merged?  While we can ignore the human-readability aspect of it, it's becoming progressively more of a problem in some of our real-world uses.   We've started having to write in bug-fix wrappers to integration code that reads these fields and recursively decode the values of all JSON fields to get the original data out.  Luckily, it's not horrific to fix on the fly as you read it via API, but the longer this bug remains the more of an issue it will be in our object fields over time as people touch/edit objects.  Thanks!

FliesLikeABrick:
> Any thoughts on when this will be merged? While we can ignore the human-readability aspect of it, it's becoming progressively more of a problem in some of our real-world uses. We've started having to write in bug-fix wrappers to integration code that reads these fields and recursively decode the values of all JSON fields to get the original data out. Luckily, it's not horrific to fix on the fly as you read it via API, but the longer this bug remains the more of an issue it will be in our object fields over time as people touch/edit objects. Thanks!

100% agreed, we are having to add fixes to our tooling to catch bad json and repair it; and occasionally running the tool I published to go through and fix our fields of data altogether.

jeremystretch:
@rbcollins123 @FliesLikeABrick please understand that NetBox is an _extremely_ active project with very few people actually doing the work. For reference, we have had [73 PRs opened](https://github.com/netbox-community/netbox/pulls?q=is%3Apr+created%3A%3E2024-07-16) in the past 30 days, and each one of them consumes time.

As a reminder, you always have the option of making changes locally yourself ahead of an official release if there's a truly pressing issue.

jeremystretch (approved):

PR #16732, opened 2024-06-26
SSO: custom name for identity providers

### LINKED ISSUE #15375: Allow for custom names for Identity Provider text in the Login link

tobast:
Fixes:  #15375 -- allows a user to customize the display aspect of social login providers (SSO and such).

This PR is separated into two commits on purpose.

The first one fixes  #15375 and adds relevant documentation. Users may specify in their configuration `SOCIAL_AUTH_BACKEND_ATTRS`, which override the display name of a social login provider, based on the `name` attribute of the `REMOTE_AUTH_BACKEND` class.

The second one allows to use an arbitrary URL as the identity provider's image. Using mdi icons works well for eg. Github social auth, but does not allow for corporate branding of an internal SSO (which is often expected by users).

### DIFF
--- a/netbox/account/views.py
+++ b/netbox/account/views.py
@@ -44,10 +44,20 @@ class LoginView(View):
         return super().dispatch(*args, **kwargs)
 
     def gen_auth_data(self, name, url, params):
-        display_name, icon_name = get_auth_backend_display(name)
+        display_name, icon_source = get_auth_backend_display(name)
+
+        icon_name = None
+        icon_img = None
+        if icon_source:
+            if '://' in icon_source:
+                icon_img = icon_source
+            else:
+                icon_name = icon_source
+
         return {
             'display_name': display_name,
             'icon_name': icon_name,
+            'icon_img': icon_img,
             'url': f'{url}?{urlencode(params)}',
         }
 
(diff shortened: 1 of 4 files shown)


netbox/account/views.py
```diff
@@ -44,10 +44,19 @@ def dispatch(self, *args, **kwargs):
         return super().dispatch(*args, **kwargs)
 
     def gen_auth_data(self, name, url, params):
-        display_name, icon_name = get_auth_backend_display(name)
+        display_name, icon_source = get_auth_backend_display(name)
+
+        icon_name = None
+        icon_img = None
+        if '://' in icon_source:
```
jerrykan, line 51: This will raise an exception if `icon_source` is `None`. It should probably be:
```
if icon_source and '://' in icon_source:
```

jerrykan:
I've tested this with a patched v4.0.6 instance and it seems to do the job (ignoring the one bug related to `icon_source`). Adding support for external images is a nice addition as well.

It would be good to see this merged.

netbox/account/views.py
tobast: Good catch, indeed. Patched in my amended commit.

arthanson (approved):
Works great!
Home | NetBox 2024-07-10 12-41-44

Issue #12826, opened 2023-06-07
dcim - racks - manufacturer/model option?


ITJamie:
### NetBox version

v3.5.3

### Feature type

Change to existing functionality

### Proposed functionality

add manufacturer + model option to racks

### Use case

some rack models are better suited to some purposes than others. it would be nice if netbox made it native for tracking this information. some racks for example are soundproofed models, some have better cable routing space on the sides etc.

this (could) potentially speed up adding of new racks as some of the existing add new rack fields could move to a manufacturer/model section (dimensions)

### Database changes

additional models needed for manufacturer / models for racks (similar to devices)

justin-davisibm:
For an example:

7014-T42 is a 7014-T42 regardless of how many times I use it- 
- Manufacturer: IBM
- 42U Ascending
- 19" between the rails
- 4 Post rack
- 25.4" wide
- 79.3" tall
- 43.3" deep 
- 636lbs empty 
- 3521lb capacity

7965-S42
- Manufacturer: IBM
- 42U Ascending
- 19" between the rails
- 4 Post rack
- 23.6" wide
- 79.5" tall
- 44.6" deep 
- 391lbs empty 
- 3700lb capacity

In addition- standardizing these values would likely assist with floor planning (https://netboxlabs.com/blog/announcing-the-netbox-plugin-bounty-program/) - consistent modeling of rack size would make the rendering of floorplans substantially easier.

jeremystretch:
Am I correct in understanding that the proposal here is to essentially replicate for racks how the DeviceType model functions for devices, by introducing a new RackType model?

ITJamie:
> Am I correct in understanding that the proposal here is to essentially replicate for racks how the DeviceType model functions for devices, by introducing a new RackType model?

Correct. 

Right now there is a lot of duplication of data across racks that could be handled with a simple rackmodel.

jeremystretch:
It's not clear to me how the attributes of existing racks are meant to be handled when introducing a new RackType model. Consider the DeviceType model as an analog to RackType: Certain device attributes (manufacturer, U height, etc.) are defined on the device type and cannot be altered on individual devices. However, as the RackType model does not yet exist, all rack attributes (height, width, etc.) which we'd expect to define on RackType are obviously defined uniquely on each existing Rack instance.

I see two options, neither of which is particularly appealing:

[...]

PR #16739, opened 2024-06-26
12826 Add Rack Type

arthanson:
### Fixes: #12826 

Add's a RackType and can set a Rack to use a Rack Type and on creation will get the fields from it. 

Rack fields associated with a rack_type are disabled on Rack form if rack_type is set.

If rack_type is set on a rack and fields associated with a rack_type are modified (via API) then a validation error is thrown.


arthanson:
@jeremystretch updated this to disable rack fields on edit if rack_type is set, also raises error if any of the fields are modified (potentially from API).

netbox/dcim/models/racks.py
```diff
@@ -29,9 +29,183 @@
...
+        target_field='name',
+        max_length=100,
+        blank=True
+    )
+    slug = models.SlugField(
+        verbose_name=_('slug'),
+        max_length=100,
+        unique=True
+    )
+    type = models.CharField(
```
jeremystretch, line 70: IMO we should rename `type` to avoid confusion with RackType. Maybe `form_factor`, or something similar?

jeremystretch: I've renamed it to `form_factor`.

netbox/dcim/models/racks.py
```diff
@@ -271,6 +396,19 @@ def clean(self):
                     })
 
     def save(self, *args, **kwargs):
+        if (not self.pk) and self.rack_type:
```
arthanson, line 399: I think will have to take out the (not self.pk) and just copy as if you edit an existing rack that did not have a rack-type and set a rack-type it doesn't get updated correctly - can see below I edited IDF128 and to have rack type same as track, but the weight is still the old one.
Racks | NetBox 2024-07-16 14-10-05

jeremystretch (approved):

PR #16741, opened 2024-06-26
fix: allow cloning field value of 0

### LINKED ISSUE #16714: Cloning DeviceType with 0U height doesn't work
### Deployment Type

NetBox Cloud

### NetBox Version

v4.0.5

### Python Version

3.12

### Steps to Reproduce

1. Import any Device Type with 0U height ([example](https://github.com/netbox-community/devicetype-library/blob/4b7eb415fb14aa35fe7c00d77fcdd20c47e3b88b/device-types/HPE/t740.yaml))
2. Clone it

### Expected Behavior

The mandatory `height` field should be pre-filled with the number 0.

### Observed Behavior

The `height` field is blank.

The Clone button URL is missing the number: `https://localhost/dcim/device-types/add/?manufacturer=173&u_height=&is_full_depth=&airflow=right-to-left&weight=2.93&weight_unit=lb`

llamafilm:
### Fixes: #16714

Allow cloning fields where value is `0`

### DIFF
--- a/netbox/utilities/querydict.py
+++ b/netbox/utilities/querydict.py
@@ -55,7 +55,7 @@ def prepare_cloned_fields(instance):
     for key, value in attrs.items():
         if type(value) in (list, tuple):
             params.extend([(key, v) for v in value])
-        elif value not in (False, None) or value == 0:
+        elif value is not False and value is not None:
             params.append((key, value))
         else:
             params.append((key, ''))
--- a/netbox/utilities/querydict.py
+++ b/netbox/utilities/querydict.py
@@ -55,7 +55,7 @@ def prepare_cloned_fields(instance):
     for key, value in attrs.items():
         if type(value) in (list, tuple):
             params.extend([(key, v) for v in value])
-        elif value not in (False, None):
+        elif value not in (False, None) or value == 0:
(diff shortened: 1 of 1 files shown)


netbox/utilities/querydict.py
```diff
@@ -55,7 +55,7 @@ def prepare_cloned_fields(instance):
     for key, value in attrs.items():
         if type(value) in (list, tuple):
             params.extend([(key, v) for v in value])
-        elif value not in (False, None):
+        elif value not in (False, None) or value == 0:
```
jeremystretch, line 58: The second condition has no effect; `0` is neither False nor None.

jeremystretch (changes requested):

netbox/utilities/querydict.py
llamafilm: 0 is False. I tested this code and it fixes the issue for me.

jeremystretch: It also breaks boolean fields, such as `enforce_unique` on the VRF model, which is why we specifically test for False.

jeremystretch (approved):
Thanks @llamafilm!

PR #16742, opened 2024-06-26
16660 - Use ngettext for strings with count

### LINKED ISSUE #16660: Use ngettext for strings with count
### Proposed Changes

`netbox/dcim/views.py` has  `_("Disconnected {count} {type}")` where `count` is as far as I can tell an arbitrary number and `type` a pre-baked plural. This cannot be properly translated in some languages, because plurals don’t work the same way as they do in English.

You need to:

1. use [`ngettext`](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#pluralization) for the count part, so plural forms can be handled properly.
2. use one full string per type, because that’s the only way nouns can be declined as necessary.

This would result in something along the lines of

```
…
ngettext("Disconnected 1 device", "Disconnected {count} devices", count)
…
ngettext("Disconnected 1 client", "Disconnected {count} clients", count)
…
```

with something like a switch-case to match a given type with a given ngettext call.

Sorry for using client and device instead of actual types, I couldn’t easily find the type list.

### Justification

You shouldn’t make up your own plural handling. There are tools for that, and good reasons to use them.

See #16659 for a similar but different situation.

Julio-Oliveira-Encora:
### Fixes: #16660

Added `ngettext` for strings with `count`, `positions`, and `rows`.

Please, feel free to indicate more places to add the `ngettext`.

### DIFF
--- a/netbox/dcim/models/device_component_templates.py
+++ b/netbox/dcim/models/device_component_templates.py
@@ -2,7 +2,7 @@ from django.contrib.contenttypes.fields import GenericForeignKey
 from django.core.exceptions import ValidationError
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext_lazy as _, ngettext
 from mptt.models import MPTTModel, TreeForeignKey
 
 from dcim.choices import *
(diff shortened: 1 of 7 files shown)


arthanson:
Closing this as this doesn't address the issue in BulkDisconnectView:
```
                messages.success(request, _(ngettext(
                    "Disconnected 1 {type}",
                    "Disconnected {count} {type_plural}",
                    count,
                )).format(
                    count=count,
                    type=self.queryset.model._meta.verbose_name,
                    type_plural=self.queryset.model._meta.verbose_name_plural
                ))
```
see frank's comment on the bug report.  I don't think it is viable to create a line for each type, so this will have to be researched further.

PR #16746, opened 2024-06-26
16523 re-enable table row background colors

### LINKED ISSUE #16523: Cannot distinguish between master and standby device of Virtual Chassis
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5

### Python Version

3.11

### Steps to Reproduce

1. Have a virtual chassis with a master and standby device.
2. Open one (or both) of the devices.
3. Check in the Virtual Chassis section of the device.
4. See that the current device isn't highlighted there.

### Expected Behavior

A visible visual distinction/highligting of the current device in the virtual chassis section

### Observed Behavior

No highlighting of the current device in the Virtual Chassis section.
This is a regression in Netbox v4.

arthanson:
### Fixes: #16523 

r105-torsw:1 (98) | NetBox 2024-06-26 14-35-30
r105-torsw:1 (98) | NetBox 2024-06-26 14-35-49


jeremystretch:
We really just need to update the class name in the HTML template to fix this. I've submitted PR #16769 to knock this out and do some additional cleanup of the VC members panel.

PR #16749, opened 2024-06-27
Fixes #16689: Load correct configuration

### LINKED ISSUE #16689: "Current Configuration" of /core/system/ does not take configuration file into account
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5

### Python Version

3.11

### Steps to Reproduce

1. Set up a netbox instance.
2. Change some value via the configuration.py, e.g. the login banner

### Expected Behavior

Expect the changes in the config file to affect the "Current Configuration" part of NETBOXURL/core/system/

### Observed Behavior

Changes are not displayed and the default values are used.
However when creating a new config revision and changing exactly nothing, the config file is parsed

tobiasge:
### Fixes: #16689

Load correct configuration on `/core/system/`.


jeremystretch (approved):
Thanks @tobiasge!

PR #16753, opened 2024-06-27
16735 - Empty object list does not show object count in "results" tab

### LINKED ISSUE #16735: Empty object list does not show object count in "results" tab
### Deployment Type

Self-hosted

### NetBox Version

v4.0.6

### Python Version

3.10

### Steps to Reproduce

1. View an empty object list

### Expected Behavior

The "results" tab at top should show either a count of zero or no badge.

### Observed Behavior

An empty badge (which renders as a dot) is displayed:

screenshot

This is a regression introduced under #16252.

Julio-Oliveira-Encora:
### Fixes: #16735

Added default:"0" to total_count in object_list.html


jeremystretch (approved):

PR #16754, opened 2024-06-27
Updated the documentation section about removing plugins

### LINKED ISSUE #16748: Documentation regarding the removal of plugin models from the database is incomplete
### Change Type

Addition

### Area

Administration

### Proposed Changes

https://netboxlabs.com/docs/netbox/en/stable/plugins/removal/ describes the removal of a plugin's models from the database.

What's missing however is the information how to remove a plugin's resources from the database so that it can be reinstalled at a later time. The problem is that the plugin's migrations are not deleted from `django_migrations`, so any attempt to install the plugin at a later time will fail because Django thinks that the migrations have already been executed.

The documentation should describe how to remove the migration entries from `django_migrations` as well.

peteeckel:

### DIFF
--- a/docs/plugins/removal.md
+++ b/docs/plugins/removal.md
@@ -70,3 +70,19 @@ DROP TABLE
 netbox=> DROP TABLE pluginname_bar;
 DROP TABLE
 ```
+
+### Remove the Django Migration Records
+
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @peteeckel!

PR #16762, opened 2024-06-28
16725 - The admin section should always come last in the navigation menu

### LINKED ISSUE #16725: The admin section should always come last in the navigation menu
### NetBox version

v4.0.6

### Feature type

### Proposed functionality

Tweak the navigation menu logic to ensure that the admin section always comes last, after any installed plugins.

### Use case

This approach provides more consistency and feels more natural, as most users will probably utilize plugins more frequently than administrative functions.

### Database changes

None

### External dependencies

None

Julio-Oliveira-Encora:
### Fixes: #16725

I replaced `append` with `insert` into menu.py to make the admin section appear last in the navigation menu.

### DIFF
--- a/netbox/netbox/navigation/menu.py
+++ b/netbox/netbox/navigation/menu.py
@@ -470,7 +470,7 @@ MENUS = [
 #
 
 for menu in registry['plugins']['menus']:
-    MENUS.append(menu)
+    MENUS.insert(-1, menu)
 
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @Julio-Oliveira-Encora. I tweaked the ordering logic a bit so that it's easier to follow.

Issue #14731, opened 2024-01-08
Integrate NetBox UI with plugins catalog


jeremystretch:
### NetBox version

v3.7.0

### Feature type

New functionality

### Proposed functionality

Provide an interface for users to compare their installed plugins against the [public database](https://netbox.dev/plugins/), and potentially to discover new plugins

### Use case

Providing a single interface where NetBox administrators can compare the installed version of each plugin to the latest available greatly simplifies the task of keeping plugins up-to-date.

### Database changes

This can probably be done without introducing any new models, as plugin details live in memory. Some caching may be needed.

### External dependencies

TBD, but no new dependencies are expected.

jeremystretch:
Blocked by #14728

thefreakquency:
Could this be done on client side? In our case (and I'm sure it's the same for many others), the Netbox infrastructure does not have access to internet...

peteeckel:
For this to work properly (and for plugin authors to know how to make sure the information in the plugin database is correct) it would really be helpful to have some documentation on how the database is created. Additionally, the database needs to be kept in a more current state if it's supposed to be useful.

Example: In the case of NetBox DNS (`netbox-plugin-dns`), the last version according to the database is 0.18.2, which is almost 9 months old, and the development status is "unknown". The current version is 0.22.1, which was released last week, and the development status is, as far as I'm concerned, "supported" ... on the other hand, the old `netbox-dns` is still in there, whose status is definitely "abandoned" (the last commit is almost a year old, and there will be no new ones, as the project was forked by me due to the sudden demise of its former maintainers).

jeffgdotorg:
**Details of the public plugin catalog endpoint**
URL: `https://api.netbox.oss.netboxlabs.com//v1/plugins`
Allowed methods: `GET`
Allowed query parameters:
- `page`: Page number to fetch
- `per_page`: Number of entries per page

At the time of this comment, it appears the endpoint expects an authentication token, which I assume will be removed soon.

Here is a sample of the endpoint's output.

```json
{
  "metadata": {
    "duration": 1,
    "pagination": {
      "current_page": 1,
      "first_page": true,
      "last_page": true,
      "page_size": 50,
      "total_items": 2,
      "total_pages": 1
    }
  },
  "data": [
    {
      "id": "77525ee1-838e-4b99-9d0a-ff3d891a9049",
      "status": "active",
      "title_short": "Netbox plugin for BGP related objects documentation.",
[...]

PR #16763, opened 2024-06-28
14731 plugins catalog

arthanson:
### Fixes: #14731 

Changed from the design of cards to a standard table as per comments.  Held off on writing tests as it is info display only and it sounds like we want to change to cards fairly soon which would require re-writing a lot of the tests.  We could mock the API and check the display based on that, but not sure what really to test beyond that.

ACLs | NetBox 2024-06-28 08-28-16
ACLs | NetBox 2024-06-28 08-27-44
ACLs | NetBox 2024-06-28 08-27-59
Plugins | NetBox 2024-07-15 21-33-43


jeffgdotorg:
The screenshots look great, and I'm finally getting around to checking out the branch and taking it for a spin.

At the moment, the plugins catalog is bombing out for me with:

```
<class 'ValueError'>

Invalid isoformat string: '2024-01-10T18:11:37.268905Z'

Python version: 3.10.12
NetBox version: 4.1.0-dev
Plugins: None installed
```

That date expression is the value of the `created_at` property for the first plugin in the list returned from the API.

jeremystretch:
Seems like it would be a lot cleaner (and easier) to list the available plugins in a table rather than in the grid layout pictured above. A table provides more efficient use of space and enables easy ordering by column. It would also make it much easier to compare the installed version of each plugin to its latest release.

jeffgdotorg:
I'm fine with a table layout for the initial implementation. We'll revisit the card / grid design down the road, during or after the 4.1 beta period.

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+    plugins = get_catalog_plugins(plugins)
+
+    cache.set('plugins-catalog-feed', plugins, 3600)
+    return plugins
+
+
+class PluginListView(UserPassesTestMixin, View):
+
+    def test_func(self):
+        return self.request.user.is_staff
```
jeremystretch, line 751: I think we want to move away from using `is_staff` for anything. Maybe we should just restrict these views to superusers?

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+
+    return plugins
+
+
+def get_catalog_plugins(plugins):
+    url = 'https://api.netbox.oss.netboxlabs.com/v1/plugins'
+    session = requests.Session()
+
+    def get_pages():
+        payload = {'page': '1', 'per_page': '25'}
```
jeremystretch, line 695: AFAICT pagination is currently broken on the catalog API. Not something we can fix here, just something to be aware of.

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+            'tag_line': plugin_config.description,
+            'description_short': plugin_config.description,
+            'author': plugin_config.author or _('Unknown Author'),
+            'created': None,
+            'updated': None,
+            'is_local': True,
+            'is_installed': True,
+            'is_certified': False,
+            'is_community': False,
+            'versions': [],
```
jeremystretch, line 684: Let's define a dataclass for these attributes to ensure we have a canonical structure to reference.

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+        }
+
+    return plugins
+
+
+def get_catalog_plugins(plugins):
+    url = 'https://api.netbox.oss.netboxlabs.com/v1/plugins'
```
jeremystretch, line 691: The API URL should probably be defined as a (non-configurable) constant in `settings.py`.

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+
+def get_plugins():
+    if plugins := cache.get('plugins-catalog-feed'):
+        return plugins
+
+    plugins = {}
+    plugins = get_local_plugins(plugins)
+    plugins = get_catalog_plugins(plugins)
+
+    cache.set('plugins-catalog-feed', plugins, 3600)
```
jeremystretch, line 744: Only the catalog data should be cached; local plugin data should always reflect what's currently installed & running.

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
             'plugins_table': plugins_table,
             'config': config,
         })
+
+
+#
+# Plugins
+#
+
+def get_local_plugins(plugins):
```
jeremystretch, line 665: The utility functions should be moved outside of `views.py`.

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+        return self.request.user.is_staff
+
+    def get(self, request):
+        q = request.GET.get('q', None)
+
+        plugins = get_plugins()
+        if q:
+            plugins = [v for k, v in plugins.items() if q.casefold() in v['name'].casefold()]
+        else:
+            plugins = [v for k, v in plugins.items()]
```
jeremystretch, line 760: These can be simplified:
```suggestion
            plugins = plugins.values()
```

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+
+class PluginListView(UserPassesTestMixin, View):
+
+    def test_func(self):
+        return self.request.user.is_staff
+
+    def get(self, request):
+        q = request.GET.get('q', None)
+
+        plugins = get_plugins()
```
jeremystretch, line 756: The plugins list should probably show installed plugins first, followed by catalog plugins, each sorted alphabetically.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,92 @@
...
+  <ul class="nav nav-tabs" role="tablist">
+    <li class="nav-item" role="presentation">
+      <a class="nav-link active" id="overview-tab" data-bs-toggle="tab" data-bs-target="#overview" type="button" role="tab" aria-controls="edit-form" aria-selected="true">
+        {% trans "Overview" %}
+      </a>
+    </li>
+    {% if not plugin.is_local %}
+    <li class="nav-item" role="presentation">
+      <button class="nav-link" id="version-history-tab" data-bs-toggle="tab" data-bs-target="#version-history" type="button" role="tab" aria-controls="object-list" aria-selected="false">
+        {% trans "Version history" %}
```
jeremystretch, line 38: Seems unnecessary to split this onto a separate tab. We can easily show both the plugin's description and version history on the same page.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,92 @@
...
+      </button>
+    </li>
+    {% endif %}
+  </ul>
+
+{% endblock tabs %}
+
+{% block content %}
+
+  {# Component creation form #}
```
jeremystretch, line 53:
```suggestion
```

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,92 @@
...
+    </li>
+    {% if not plugin.is_local %}
+    <li class="nav-item" role="presentation">
+      <button class="nav-link" id="version-history-tab" data-bs-toggle="tab" data-bs-target="#version-history" type="button" role="tab" aria-controls="object-list" aria-selected="false">
+        {% trans "Version history" %}
+      </button>
+    </li>
+    <li class="nav-item" role="presentation">
+      <button class="nav-link" id="install-tab" data-bs-toggle="tab" data-bs-target="#install" type="button" role="tab" aria-controls="object-list" aria-selected="false">
+        {% trans "Install" %}
```
jeremystretch, line 43: We probably want to display local installation instructions only for self-hosted deployments (where `settings.RELEASE.edition` is `community`).

netbox/core/tables/plugins.py
```diff
@@ -37,3 +41,68 @@ class Meta(BaseTable.Meta):
...
+        default_columns = (
+            'version', 'last_updated', 'min_version', 'max_version',
+        )
+        orderable = False
+
+    def render_last_updated(self, value, record):
+        return naturalday(datetime.fromisoformat(value))
+
+
+class CertifiedPluginTable(BaseTable):
```
jeremystretch, line 77: This should be renamed, as not all catalog plugins have been certified.
```suggestion
class CatalogPluginTable(BaseTable):
```

netbox/core/tables/plugins.py
```diff
@@ -37,3 +41,68 @@ class Meta(BaseTable.Meta):
...
+        fields = (
+            'version', 'last_updated', 'min_version', 'max_version',
+        )
+        default_columns = (
+            'version', 'last_updated', 'min_version', 'max_version',
+        )
+        orderable = False
+
+    def render_last_updated(self, value, record):
+        return naturalday(datetime.fromisoformat(value))
```
jeremystretch, line 74: The release timestamp should be deserialized to a datetime object on ingestion.
```suggestion
```

netbox/netbox/navigation/menu.py
```diff
@@ -430,6 +430,11 @@
         MenuGroup(
             label=_('System'),
             items=(
+                MenuItem(
```
jeremystretch, line 433: Should probably move this after the system item.

netbox/core/tables/plugins.py
```diff
@@ -1,13 +1,17 @@
...
 __all__ = (
-    'PluginTable',
+    'CertifiedPluginTable',
+    'InstalledPluginTable',
+    'PluginVersionTable',
 )
 
 
-class PluginTable(BaseTable):
+class InstalledPluginTable(BaseTable):
```
jeremystretch, line 14: IMO we can ditch the plugins list on the system view now that we have a dedicated view for them.

jeremystretch (changes requested):
Thanks for taking this on!

netbox/core/views.py
jeffgdotorg: Further asks here:
- On deployments where `settings.RELEASE.edition != 'community'`, default to showing only catalog plugins for which `release_latest.is_netboxlabs_supported == True`.
- Provide a way to toggle this behavior from the table view

One way to achieve this would be to add a query parm to this view and set it to a truthy value on non-community installations, but do what makes sense to you. Fancy client-side toggling would be cool, but only if it's easy.

If this comment needs to be a FR issue, let me know.

netbox/templates/core/plugin.html
jeffgdotorg: I'd like to keep the version history on its own tab for now, so that we can experiment with longer and richer content in the Overview tab.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,92 @@
...
+    <div class="card">
+
+      <div class="htmx-container table-responsive" id="object_list">
+        {% include 'htmx/table.html' %}
+      </div>
+    </div>
+  </div>
+  <div class="tab-pane" id="install" role="tabpanel" aria-labelledby="install-tab">
+    <p>You can install this plugin from the command line with PyPi.</p>
+    <p>The following commands may be helpful; always refer to the plugin's own documentation and the Installing a Plugin unit of the NetBox documentation.</p>
```
jeffgdotorg, line 68:
```suggestion
    <p>The following commands may be helpful; always refer to <a href="{{ plugin. homepage_url }}" target="_blank">the plugin's own documentation <i class="mdi mdi-launch"></i></a> and the <a href="https://netboxlabs.com/docs/netbox/en/stable/plugins/installation/" target="_blank">Installing a Plugin unit <i class="mdi mdi-launch"></i></a> of the NetBox documentation.</p>
```

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+        for page in range(2, num_pages + 1):
+            payload['page'] = page
+            next_page = session.get(url, params=payload).json()
+            yield next_page
+
+    for page in get_pages():
+        for data in page['data']:
+
+            versions = []
+            versions.append(data['release_latest'])
```
jeffgdotorg, line 709: Don't include this item in the table, as it's a duplicate of an entry in `release_recent_history` and is resulting in multiple appearances in the rendered table.

netbox/core/views.py
```diff
@@ -642,11 +648,145 @@ def get(self, request):
...
+
+    def test_func(self):
+        return self.request.user.is_staff
+
+    def get(self, request, name):
+
+        plugins = get_plugins()
+        plugin = plugins[name]
+
+        table = PluginVersionTable(plugin['versions'], user=request.user)
```
jeffgdotorg, line 786: The versions currently come out of the API ordered from oldest to newest. Before we display this data, we should either:
- Reverse the array's ordering (lazy way); or
- Sort the array in descending order by `date`

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,92 @@
...
+    {% endif %}
+  </ul>
+
+{% endblock tabs %}
+
+{% block content %}
+
+  {# Component creation form #}
+  <div class="tab-pane show active" id="overview" role="tabpanel" aria-labelledby="overview-tab">
+    {{ plugin.description_short }}
```
jeffgdotorg, line 55: This attribute needs to be rendered with Markdown support.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,92 @@
...
+{% block breadcrumbs %}
+  <li class="breadcrumb-item"><a href="{% url 'core:plugin_list' %}">{% trans "Plugins" %}</a></li>
+{% endblock breadcrumbs %}
+
+{% block subtitle %}
+  <div class="text-secondary fs-5">
+    {{ plugin.tag_line }}
+  </div>
+{% endblock subtitle %}
+
```
jeffgdotorg, line 20: Please add the following additional items below the tagline subtitle:
- `<a href="{{ plugin.homepage.url }}" target="_blank">Learn more <i class="mdi mdi-launch"></i></a>`
- `License: {{ plugin.license_type }}`

jeffgdotorg (changes requested):
Looking really good, I think these changes get us to b1.

netbox/core/views.py
arthanson: @jeffgdotorg can we make the last one a FR for post beta?

netbox/core/views.py
jeremystretch: I'm actually going to take this back, since we're currently using `is_staff` as a litmus for access to other admin views. This isn't ideal, but I'd rather be consistent. This can be addressed under a separate FR.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,97 @@
...
+        {% trans "Overview" %}
+      </a>
+    </li>
+    {% if not plugin.is_local %}
+      <li class="nav-item" role="presentation">
+        <button class="nav-link" id="version-history-tab" data-bs-toggle="tab" data-bs-target="#version-history" type="button" role="tab" aria-controls="object-list" aria-selected="false">
+          {% trans "Version history" %}
+        </button>
+      </li>
+      {% if plugin.is_community %}
```
jeremystretch, line 43: It's not the plugin itself we need to check here, but rather the edition of NetBox. Only the Community edition should include instructions for local installation; Cloud & Enterprise will have different means of installing the plugin.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,97 @@
...
+
+{% block breadcrumbs %}
+  <li class="breadcrumb-item"><a href="{% url 'core:plugin_list' %}">{% trans "Plugins" %}</a></li>
+{% endblock breadcrumbs %}
+
+{% block subtitle %}
+  <div class="text-secondary fs-5">
+    {{ plugin.tag_line }}
+    <a href="{{ plugin.homepage.url }}" target="_blank">Learn more <i class="mdi mdi-launch"></i></a><br />
+    <strong>License:</strong> {{ plugin.license_type }}
```
jeremystretch, line 19: We're jamming a lot of content into the header. Let's move these details down into the body of the page.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,97 @@
...
+  </div>
+{% endblock subtitle %}
+
+{% block controls %}
+<div class="text-secondary">
+From
+</div>
+{{ plugin.author }}
+{% endblock %}
```
jeremystretch, line 28: This belongs in the page body. The `controls` block should only contain action elements.

netbox/core/plugins.py
```diff
@@ -0,0 +1,141 @@
...
+    description_short: str = ''
+    author: str = ''
+    homepage_url: str = ''
+    license_type: str = ''
+    created: datetime.datetime = None
+    updated: datetime.datetime = None
+    is_local: bool = False
+    is_installed: bool = False
+    is_certified: bool = False
+    is_community: bool = False
```
jeremystretch, line 42: This appears to be unnecessary; it's just the inverse of `is_certified`?

netbox/core/plugins.py
```diff
@@ -0,0 +1,141 @@
...
+        plugins[plugin_config.name] = Plugin(
+            slug=plugin_config.name,
+            name=plugin_config.verbose_name,
+            tag_line=plugin_config.description,
+            description_short=plugin_config.description,
+            author=plugin_config.author or _('Unknown Author'),
+            is_local=True,
+            is_installed=True,
+            is_certified=False,
+            is_community=False,
```
jeremystretch, line 62: These can be omitted.
```suggestion
            is_certified=False,
            is_community=False,
```

netbox/core/plugins.py
```diff
@@ -0,0 +1,141 @@
...
+                title_long=data['title_long'],
+                tag_line=data['tag_line'],
+                description_short=data['description_short'],
+                author=data['author']['name'] or _('Unknown Author'),
+                homepage_url=data['homepage_url'],
+                license_type=data['license_type'],
+                created=datetime_from_timestamp(data['created_at']),
+                updated=datetime_from_timestamp(data['updated_at']),
+                is_local=False,
+                is_installed=False,
```
jeremystretch, line 117: These can be omitted.
```suggestion
                is_local=False,
                is_installed=False,
```

jeremystretch (changes requested):

netbox/core/tables/plugins.py
```diff
@@ -1,39 +1,76 @@
...
+        fields = (
+            'version', 'last_updated', 'min_version', 'max_version',
+        )
+        default_columns = (
+            'version', 'last_updated', 'min_version', 'max_version',
+        )
+        orderable = False
+
+
+class CatalogPluginTable(BaseTable):
```
jeremystretch, line 41: There is currently no way to displayed the version of each installed plugin.

netbox/templates/core/plugin_list.html
```diff
@@ -0,0 +1,43 @@
...
+
+{% block tabs %}
+  <ul class="nav nav-tabs px-3">
+    <li class="nav-item" role="presentation">
+      <a class="nav-link active" role="tab">{% trans "Plugins" %}</a>
+    </li>
+  </ul>
+{% endblock tabs %}
+
+{% block content %}
```
jeremystretch, line 20: There is no configuration button for the table. This should just use the standard display as with other object lists.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,114 @@
...
+            <td>{{ plugin.license_type }}</td>
+          </tr>
+          <tr>
+            <th scope="row">{% trans "Description" %}</th>
+            <td>{{ plugin.description_short|markdown }}</td>
+          </tr>
+        </table>
+      </div>
+    </div>
+  </div>
```
jeremystretch, line 73: We should be showing all the data we have about a plugin here: author, URL, etc.

netbox/core/views.py
```diff
@@ -642,11 +634,56 @@ def get(self, request):
...
+
+class PluginView(UserPassesTestMixin, View):
+
+    def test_func(self):
+        return self.request.user.is_staff
+
+    def get(self, request, name):
+
+        plugins = get_plugins()
+        plugin = plugins[name]
```
jeremystretch, line 681: An invalid plugin name should return a 404.

jeremystretch (changes requested):
Could you please add some basic tests to validate the API processing and views?

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,114 @@
...
+{% endblock object_identifier %}
+
+{% block breadcrumbs %}
+  <li class="breadcrumb-item"><a href="{% url 'core:plugin_list' %}">{% trans "Plugins" %}</a></li>
+{% endblock breadcrumbs %}
+
+{% block subtitle %}
+  <div class="text-secondary fs-5">
+    {{ plugin.tag_line }}
+    <a href="{{ plugin.homepage.url }}" target="_blank">Learn more <i class="mdi mdi-launch"></i></a>
```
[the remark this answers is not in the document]
jeffgdotorg: This typo might have originated from me, in which case I apologize for the trouble.

jeffgdotorg (changes requested):
Nice progress here. BTW the Markdown rendering in `plugin.description_short` is now working as expected.

netbox/core/plugins.py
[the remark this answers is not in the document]
arthanson: It's because the table is sorted with those items.  It is part of the standard table functionality so adding a tooltip or title should be a separate FR as it would affect all tables in NB

arthanson:
@jeffgdotorg Detail View changed quite a bit from Jeremy's feedback - please take a re-review:
IP Fabric | NetBox 2024-07-23 15-20-23

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,154 @@
...
+
+{% block content %}
+
+  <div class="tab-pane show active" id="overview" role="tabpanel" aria-labelledby="overview-tab">
+
+  <div class="row">
+    <div class="col col-md-12">
+
+      <div class="card">
+        <h5 class="card-header">{% trans "Author" %}</h5>
```
jeffgdotorg, line 58: Because the `org_id` appears to be a non-user-facing field, and `url` currently seems to be tethered to `plugin.homepage_url`, let's just move the author info to a row in the _Plugin_ table. If `author.url` gets a life of its own down the road, we can linkify the author name with it.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,154 @@
...
+
+      <div class="card">
+        <h5 class="card-header">{% trans "Plugin" %}</h5>
+        <table class="table table-hover attr-table">
+          <tr>
+            <th scope="row">{% trans "Name" %}</th>
+            <td>{{ plugin.title_short }}</td>
+          </tr>
+          <tr>
+            <th scope="row">{% trans "Tag line" %}</th>
```
jeffgdotorg, line 83: The tag line has limited value in this context; it's meant for display in a summary list of plugins and serves as a TL;DR for `plugin.description_short`.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,154 @@
...
+            <td>{{ plugin.tag_line|placeholder }}</td>
+          </tr>
+          <tr>
+            <th scope="row">{% trans "License" %}</th>
+            <td>{{ plugin.license_type }}</td>
+          </tr>
+          <tr>
+            <th scope="row">{% trans "Description" %}</th>
```
jeffgdotorg, line 91: In the medium term, this field may include images (loaded only from a trusted source) and render somewhat tall. That's the reason my prototypes dedicated the entire _Overview_ tab to it. This change is fine for the short term, but we should expect to iterate.

netbox/templates/core/plugin.html
```diff
@@ -0,0 +1,154 @@
...
+            <td>{{ plugin.description_short|markdown }}</td>
+          </tr>
+          <tr>
+            <th scope="row">{% trans "URL" %}</th>
```
jeffgdotorg, line 95: This row is redundant with the _Learn more_ link at the top of the page. Please remove.

jeffgdotorg (changes requested):
Just a couple of small asks here. Otherwise looks good.

netbox/core/plugins.py
```diff
@@ -0,0 +1,195 @@
+import datetime
+import importlib
+import importlib.util
+from dataclasses import dataclass, field
+from typing import Optional
+
+import requests
```
jeffgdotorg, line 7: @natm points out that the way we're using `requests` does not honor the system proxy settings, which might result in unnecessary catalog load failures and possibly also data leaks. Can we update the client to respect this setting?

netbox/core/plugins.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+
+    return plugins
+
+
+def get_catalog_plugins():
+    """
+    Return a dictionary of all entries in the plugins catalog, mapped by name.
+    """
+    session = requests.Session()
```
jeffgdotorg, line 95: For catalog API requests can we set a custom `User-Agent` header of a form similar to:

`NetBox/{settings.RELEASE.version} (Python {platform.python_version()}; {platform.system()} {platform.machine()}) {settings.RELEASE.edition}`

Example value:

`NetBox/v4.1.0 (Python 3.10.12; Linux x86_64) Community`

jeffgdotorg (changes requested):
A couple additional flags from talking through this PR with @natm

arthanson:
Added proxy header and user-agent string to API call.

jeremystretch (approved):
Thanks @arthanson!

PR #16764, opened 2024-06-28
Fixes #16758: Create language cookie if required

### LINKED ISSUE #16758: Language setting is lost for new browser sessions
### Deployment Type

Self-hosted

### NetBox Version

v4.0.6

### Python Version

3.12

### Steps to Reproduce

1. Be sure to have `LOGIN_TIMEOUT > 0` set
2. Login with a user which has language set to auto
3. Netbox will show in the language determined from browser language
4. Go to user preferences and set an alternate language
5. Netbox will show in alternate language (`django_language` cookie is set)
6. Close your browser and navigate to your Netbox again with a new browser instance (of course with the same browser)
7. The user is still logged in but language is again detected by browser language

This is because `django_language` cookie's lifetime is set to browser session's lifetime. Each time the browser is closed, the cookie is deleted. Because no new login happens, the language is not set (related #16512).

### Expected Behavior

The user's language setting is applied

### Observed Behavior

Browser language is applied.

m2martin:
### Fixes: #16758

If a new browser session has no language cookie set, check if it should be set and redirect to the requested page.

### DIFF
--- a/netbox/account/views.py
+++ b/netbox/account/views.py
@@ -111,7 +111,7 @@ class LoginView(View):
 
             # Set the user's preferred language (if any)
             if language := request.user.config.get('locale.language'):
-                response.set_cookie(settings.LANGUAGE_COOKIE_NAME, language)
+                response.set_cookie(settings.LANGUAGE_COOKIE_NAME, language, max_age=request.session.get_expiry_age())
 
             return response
 
@@ -206,7 +206,7 @@ class UserConfigView(LoginRequiredMixin, View):
 
             # Set/clear language cookie
             if language := form.cleaned_data['locale.language']:
-                response.set_cookie(settings.LANGUAGE_COOKIE_NAME, language)
+                response.set_cookie(settings.LANGUAGE_COOKIE_NAME, language, max_age=request.session.get_expiry_age())
             else:
                 response.delete_cookie(settings.LANGUAGE_COOKIE_NAME)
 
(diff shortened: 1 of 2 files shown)


m2martin:
I'd like to get rid of the redirection, but `django.utils.translation.activate()` seems worthless here. Does someone have a better idea?

jeremystretch (changes requested):
This probably isn't the best approach to ensure translation is active. Instead, we should update the calls to `set_cookie()` in [LoginView](https://github.com/netbox-community/netbox/blob/c7dcded74fdf8d846ac12f40d03b036015db7ce6/netbox/account/views.py#L114) and [UserConfigView](https://github.com/netbox-community/netbox/blob/c7dcded74fdf8d846ac12f40d03b036015db7ce6/netbox/account/views.py#L209) to set an expiration time on the language cookie (which should presumably match the expiration time of the session cookie, which was derived from `LOGIN_TIMEOUT`).

m2martin:
I totally agree that this approach is not very elegant. After looking into the way how Django handles cookie renewals, I found a very simple way to accomplish the same for our language cookie. Please have a look at the new commit.

dreng:
IMO, instead of dealing with cookies, those kind of user preferences should be stored in the database. Otherwise you will have to set the language preference in every browser on every computer again (not even mentioning private browsing). In our case you have to set it each day actually, because cookies will be automatically removed by a group policy. That can be really annoying.

jeremystretch (approved):
Thanks @m2martin, great work!

PR #16769, opened 2024-06-30
Fixes #16523: Restore highlighting of current device in virtual chassis members panel

### LINKED ISSUE #16523: Cannot distinguish between master and standby device of Virtual Chassis

jeremystretch:
### Fixes: #16523

- Change the class name on `<tr>` from `info` to `table-primary`
- Utilize the `checkmark` template tag
- Misc cleanup

### DIFF
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -125,28 +125,30 @@
                       </div>
                     </h5>
                     <table class="table table-hover attr-table">
-                        <tr>
-                            <th>{% trans "Device" %}</th>
-                            <th>{% trans "Position" %}</th>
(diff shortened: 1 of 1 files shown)


jeremystretch:
screenshot

PR #16800, opened 2024-07-03
Closes #15621: User notifications

### LINKED ISSUE #15621: Ability to "watch" an object and be alerted to changes
### NetBox version

v3.7.4

### Feature type

New functionality

### Proposed functionality

Provide the ability for a user to designate specific objects and receive notifications of any changes to those objects. Users should see a notification icon in the top right corner of the page when a new notification is present.

The notifications mechanism could piggyback off of the change log. Users should be able to designate specific objects as well as defining rules to match discrete sets of objects (e.g. "all devices in site A with role B").

### Use case

Users may often desire to be alerted to specific types of changes in NetBox, such as when preparing for planned work or when informed of an ongoing issue.

### Database changes

A new model would likely be required to define notification rules and associate them with a particular user (or group?). The rule format would likely mimic the one used to define [permission constraints](https://docs.netbox.dev/en/stable/administration/permissions/#constraints).

### External dependencies

N/A

jeremystretch:
### Closes: #15621

- Implement dynamic registration for event types
- Add event types for job failures and errors
- Alter event handling logic to enqueue events by canonical type rather than by the corresponding ObjectChange action type
- Introduce three new models: Subscription, NotificationGroup, and Notification
- Add `post_save` signal receiver to automatically generate notification for subscribed users
- Extend EvenRule to support sending notifications
- Add notifications dropdown in top navigation bar
- Provide user views for subscriptions & notifications
- Add model documentation


arthanson:
Getting the following when checkout the code:
```
django.core.management.base.SystemCheckError: SystemCheckError: System check identified some issues:

ERRORS:
extras.Notification.event_type: (fields.E004) 'choices' must be an iterable (e.g., a list or tuple).
```

netbox/extras/api/serializers_/notifications.py
```diff
@@ -0,0 +1,82 @@
...
+    object_type = ContentTypeField(
+        queryset=ObjectType.objects.with_feature('notifications'),
+    )
+    object = serializers.SerializerMethodField(read_only=True)
+    user = UserSerializer(nested=True)
+
+    class Meta:
+        model = Notification
+        fields = [
+            'id', 'url', 'display', 'object_type', 'object_id', 'object', 'user', 'created', 'read', 'event_type',
```
arthanson, line 29: should there be a display_url here as well

netbox/core/events.py
```diff
@@ -0,0 +1,33 @@
+from django.utils.translation import gettext as _
```
arthanson, line 1: This is only ever used in extras, should it be in extras?

arthanson (changes requested):
See previous comment on getting SystemCheckError wasn't able to actually run it to test.

Also, should add model docs for the new models?

netbox/extras/api/serializers_/notifications.py
jeremystretch: There's no dedicated UI view for a notification or subscription, but I'll add `display_url` for notification groups.

netbox/core/events.py
jeremystretch: I haven't submitted any FRs yet, but I want to 1) allow plugins to register custom event types and 2) extend EventRule to allow mapping to any event type(s). So I opted to put this in `core` in anticipation of that work.

arthanson:
Went to user and clicked [Subscribe] button at top of profile and got: 
```
  File "/Users/arthurhanson/dev/work/netbox/venv/lib/python3.11/site-packages/django/forms/forms.py", line 192, in errors
    self.full_clean()
  File "/Users/arthurhanson/dev/work/netbox/venv/lib/python3.11/site-packages/django/forms/forms.py", line 329, in full_clean
    self._post_clean()
  File "/Users/arthurhanson/dev/work/netbox/venv/lib/python3.11/site-packages/django/forms/models.py", line 496, in _post_clean
  File "/Users/arthurhanson/dev/work/netbox/venv/lib/python3.11/site-packages/django/db/models/base.py", line 1531, in full_clean
    self.clean()
  File "/Users/arthurhanson/dev/work/netbox/netbox/extras/models/notifications.py", line 221, in clean
    if self.object_type not in ObjectType.objects.with_feature('notifications'):
       ^^^^^^^^^^^^^^^^
  File "/Users/arthurhanson/dev/work/netbox/venv/lib/python3.11/site-packages/django/db/models/fields/related_descriptors.py", line 264, in __get__
    raise self.RelatedObjectDoesNotExist(
extras.models.notifications.Subscription.object_type.RelatedObjectDoesNotExist: Subscription has no object_type.
```

arthanson:
Notifications table, click on object to sort the table, get error:
```
  File "/Users/arthurhanson/dev/work/netbox/venv/lib/python3.11/site-packages/django/db/models/query.py", line 1701, in order_by
    obj.query.add_ordering(*field_names)
  File "/Users/arthurhanson/dev/work/netbox/venv/lib/python3.11/site-packages/django/db/models/sql/query.py", line 2253, in add_ordering
    self.names_to_path(item.split(LOOKUP_SEP), self.model._meta)
  File "/Users/arthurhanson/dev/work/netbox/venv/lib/python3.11/site-packages/django/db/models/sql/query.py", line 1748, in names_to_path
    raise FieldError(
django.core.exceptions.FieldError: Field 'object' does not generate an automatic reverse relation and therefore cannot be used for reverse querying. If it is a GenericForeignKey, consider adding a GenericRelation.
```
Notifications | NetBox 2024-07-12 21-15-11

netbox/extras/forms/bulk_import.py
```diff
@@ -247,3 +250,28 @@ class Meta:
...
+        help_text=mark_safe(
+            _('User names separated by commas, encased with double quotes')
+        )
+    )
+    groups = CSVModelMultipleChoiceField(
+        label=_('Groups'),
+        queryset=Group.objects.all(),
+        required=False,
+        to_field_name='name',
+        help_text=mark_safe(
```
arthanson, line 270: why mark_safe?

netbox/extras/signals.py
```diff
@@ -280,4 +292,39 @@ def process_job_end_event_rules(sender, **kwargs):
...
+    if ct.model not in registry['model_features']['notifications'].get(ct.app_label, []):
+        return
+
+    # Find all subscribed Users
+    subscribed_users = Subscription.objects.filter(object_type=ct, object_id=instance.pk).values_list('user', flat=True)
+    if not subscribed_users:
+        return
+
+    # Delete any existing Notifications for the object
+    Notification.objects.filter(object_type=ct, object_id=instance.pk, user__in=subscribed_users).delete()
```
arthanson, line 324: So this would be one notification for an object regardless of what happens, so if I have different events one for create and one for update and I do both, I should only get a single notification?  We probably want to document that.

arthanson (changes requested):
Several crashes

arthanson:
There is no confirmation of delete on a notification.  Not very critical if you delete it and you don't want, but everything else in NetBox has a confirmation on delete so not sure if this was by design.

netbox/extras/forms/bulk_import.py
jeremystretch: Copy/paste error :\

jeremystretch:
> There is no confirmation of delete on a notification.

Yeah, I modeled it after GitHub notifications: Clicking through to an object will mark the notification as "read," but not delete it. Deletion is a manual task, but does not require confirmation, so that notifications can be easily dismissed.

netbox/extras/signals.py
jeremystretch: I'm of the opinion that a user should generally see no more than one notification per object, but I think this depends somewhat on the event type. For example, you probably don't need to see separate notifications that an object was created _and_ later updated. But this assumption might not hold when we consider custom event types (being introduced in #16886).

I'm going to leave this as-is for now, with the expectation that we'll reassess the strategy once we get some real-world feedback during the beta evaluation.

arthanson (approved):
There is just one minor UI in the notification list there is an info icon that I think isn't supposed to be there (see screenshot) otherwise looks good:
Notifications | NetBox 2024-07-15 21-03-18

PR #16803, opened 2024-07-03
Allowed configuration of Sentry send_default_pii parameter

### LINKED ISSUE #16802: Allow configuration of the Sentry send_default_pii parameter
### NetBox version

v4.0.6

### Feature type

### Proposed functionality

Allow configuration of the Sentry `send_default_pii` parameter.

Also default to not send default pii data - to avoid sending auth tokens and session-ids.

### Use case

For our use case, we do not want to log sensitive data such as auth tokens or session-ids from the cookies. By allowing configuration of the `send_default_pii` parameter this can be accomplished.
I would also like to change the default behavior so scrubbing these data points are the default behavior (a.k.a `send_default_pii=False` - just like it is in the Sentry sdk).

### Database changes

_No response_

### External dependencies

_No response_

mulmat:
### Fixes: #16802

- Allowed configuration of Sentry send_default_pii parameter.
- Also changed default value of send_default_pii to False to avoid sending sensitive data to Sentry.


jeremystretch (approved):
Thanks @mulmat!

PR #16804, opened 2024-07-03
Closes #16716, add NAT IP to device view for OOB IP

### LINKED ISSUE #16716: Add NAT outside address for out-of-band IP in GUI on device page
### NetBox version

v4.0.6

### Feature type

### Proposed functionality

Make it as easy to see the NAT address for a device's out-of-band IP as it is to see the NAT address of the device's primary IP
image

### Use case

Simple time-saving feature for users who also do NAT on iDRAC/LOM OOB, it would make life easier for technicians to just copy the IP directly when looking at an device without having to click on the IP.

### Database changes

None

### External dependencies

None

RobertH1993:
### Fixes: #16716

Adds the NAT IP to the device view for OOB IPs


jeremystretch (approved):
Thanks @RobertH1993!

Issue #16648, opened 2024-06-18
Dashboard contrast issues


andrewgormley:
### NetBox version

v4.1.0-dev

### Feature type

Change to existing functionality

### Proposed functionality

Fix contrast issues on the dashboard:

- [ ] (Dashboard) Module headers low contrast text [both, single]
- [ ] (Dashboard) Body copy low contrast on NetBox news [dark, single]

### Use case

Improve ux for users with color vision deficiency

PR #16759, opened 2024-06-28
16649 general contrast issues

andrewgormley:
### Fixes: #16649

- [x]  (Footer) Date and version text low contrast [dark, global]
- [x]  (Header) Secondary text low contrast [dark, global]
- [ ]  (Racks) Low contrast labels [both, multiple]
Leaving labels off this task for now, per our last discussion about labels needing to stay the same colors.

- [x]  (Object) Low contrast on object title sub items eg. created 2020-10-10 [dark, multiple]
- [x]  (Misc) Low contrast on form error text [both, multiple] 
**Fixed for light. Dark mode is not possible with the current fontsize, and colors.**

- [x]  (Toast) Low contrast for warning toast [light, multiple] 
**Improved the WCAG contrast score but not quite to AA standard as orange on light bg are both classed high contrast colors according to WCAG but this is okay with the warning visual aids**

- [x]  (Misc) Low contrast on placeholder module placeholder text [dark, multiple]

Issue #16385, opened 2024-06-03
status color background is not distinguishable


opericgithub:
### Deployment Type

Self-hosted

### NetBox Version

v4.0.3

### Python Version

3.10

### Steps to Reproduce

Not actually a real bug, but more of a suggestion. 
When changing the background color values ​​for the prefix statuses in the "FIELD_CHOICES" section of the configuration.py file, it is very difficult to distinguish black, gray and white colors. in the previous version (v3.x) the difference was much more obvious.

### Expected Behavior

v3 black:
![image](https://github.com/netbox-community/netbox/assets/65983735/f0437843-7884-482b-9cfa-21c497f4afa8)

v3 gray:
![image](https://github.com/netbox-community/netbox/assets/65983735/89cffa3c-c2c8-4657-9828-9b51d71aa276)

v3 white:
![image](https://github.com/netbox-community/netbox/assets/65983735/6e45f663-aa5b-4102-a3b0-17d083e59106)

### Observed Behavior

v4 black:
![image](https://github.com/netbox-community/netbox/assets/65983735/2f12f7ee-0281-4a76-a74e-f0701503e22d)

v4 gray:
![image](https://github.com/netbox-community/netbox/assets/65983735/43830d32-8d9d-46fb-b8e4-f47371c72f76)

v4 white:
![image](https://github.com/netbox-community/netbox/assets/65983735/63f43524-91a4-4d5a-9636-6d5095f501e5)

jeffgdotorg:
Thanks for reporting this low visual contrast issue. I'm putting it through and assigning low severity, partly because there's a comprehensive rework of contrast planned for v4.1 under #16390.

jeremystretch:
Looks like we lost the black, white, and gray styles during the move to Bootstrap v5.3. We probably just need to create aliases for the native styles:

* `black` -> `dark`
* `white` -> `light`
* `gray` -> `secondary`

jeremystretch:
[Bootstrap docs](https://getbootstrap.com/docs/5.3/utilities/colors/) for reference. Custom colors need to be defined in `styles/overrides/_bootstrap.scss`.

PR #16824, opened 2024-07-04
16648 dashboard contrast issues

andrewgormley:
### Fixes: #16648 

- [ ]  (Dashboard) Module headers low contrast text [both, single]
I've changed default card header colors to be more in-line with the theme selected as per the design and increased the contrast. Some card header colors still score low on the WCAG scale due to the customizable colors we allow users to select, however in my opinion they are still legible, especially with the ability to select to preference. 
- [x] (Dashboard) Body copy low contrast on NetBox news [dark, single]
The Netbox News sub text was fixed as secondary text color was changed globally on a prior PR.
- [x] Fix bug where user sets widget color to gray on dark mode making card header text illegible


andrewgormley:
@jeremystretch I've also re-implemented a change we took out of #16759 as I discovered it was not redundant and affected the color of table headers on dark mode. Please let me know if you prefer it to stay out of this PR.

netbox/project-static/styles/overrides/_tabler.scss
```diff
@@ -66,6 +66,19 @@ table th.orderable a {
...
+    // Default color of card header changes depending on theme
+    &.bg-default {
+      background: var(--tblr-bg-surface-secondary)!important;
+    }
+    a {
+      color: inherit!important;
+    }
+  }
+}
+
```
jeremystretch, line 81: Let's move these to a new file, `overrides/_gridstack.css`, and include it in `netbox.scss`, as these classes come from the [gridstack.js](https://gridstackjs.com/) package rather than Tabler.

netbox/templates/extras/dashboard/widget.html
```diff
@@ -9,14 +9,14 @@
     gs-id="{{ widget.id }}"
 >
   <div class="card grid-stack-item-content">
-    <div class="card-header text-{{ widget.fg_color }} bg-{{ widget.color|default:"secondary" }} px-2 py-1 d-flex flex-row">
+    <div class="card-header {% if widget.color and widget.color != 'gray' %} text-{{ widget.fg_color }} {% endif %} bg-{{ widget.color|default:'default' }} px-2 py-1 d-flex flex-row">
```
jeremystretch, line 12: We want to avoid checking against specific values for maintainability reasons. It looks like the root issue might be related to #16385, which is being handled as a bug in the current release. If so, we can probably ignore this issue for now.

jeremystretch (changes requested):

netbox/templates/extras/dashboard/widget.html
jeremystretch: The foreground color is [explicitly mapped](https://github.com/netbox-community/netbox/blob/22348cdbfc2b3b6ea200100704c03e3b82a3b63c/netbox/extras/dashboard/widgets.py#L135) to individual background colors; we can easily change these as needed.

netbox/project-static/styles/overrides/_tabler.scss
andrewgormley: fixed.

netbox/templates/extras/dashboard/widget.html
andrewgormley: I've removed the 'if gray' statement in order for a better fix with #16385. I've kept in the if statement to see if a color is assigned or not in order for the header text color to change with the theme. Please let me know if there are any isues with this approach @jeremystretch

jeremystretch (approved):
Thanks @andrewgormley!

Issue #16650, opened 2024-06-18
Button contrast issues


andrewgormley:
### NetBox version

v4.1.0-dev

### Feature type

Change to existing functionality

### Proposed functionality

Fix button contrast issues throughout the app:

- [x] (List page) Low contrast on buttons [both, multiple]
- [x] (Racks) Low contrast download svg button [dark, multiple]
- [x] (Buttons) Focus state on buttons is missing focus ring causing contrast and CVD issue [both, global]
- [x] (Misc) Low contrast on cancel button [dark, multiple]
- [x] (Misc) Low contrast on create combo button [dark, multiple]

### Use case

Improve ux for users with color vision deficiency

PR #16845, opened 2024-07-08
16650 button contrast issues

andrewgormley:
### Fixes: #16650

- [x]  (List page) Low contrast on buttons [both, multiple]
- [x]  (Racks) Low contrast download svg button [dark, multiple]
- [x]  (Buttons) Focus state on buttons is missing focus ring causing contrast and CVD issue [both, global]
- [x]  (Misc) Low contrast on cancel button [dark, multiple]
- [x]  (Misc) Low contrast on create combo button [dark, multiple]


arthanson:
@andrewgormley just thought I'd mention - I actually find the black text on some of the buttons much harder to read, for example the export button with black on fuscia (see screenshot) sort of strains/hurts my eyes.  Also the delete button (black on red) isn't as bad but still seems less readable than before.  Not sure if it's just my eyes.

Devices | NetBox 2024-07-09 11-42-23
Devices | NetBox 2024-07-09 11-43-47
PP:B117 | NetBox 2024-07-09 11-44-46

jsenecal:
I'm sorry if I come late to this discussion, but wouldn't it be best to address some of these color scheme issues as a user preference `Use high-contrast colors and buttons` ? 

This black on [green,orange,yellow,blue,purple] looks very harsh for the non-color impaired IMHO.

netbox/project-static/styles/overrides/_tabler.scss
```diff
@@ -86,12 +133,17 @@ body[data-bs-theme=dark] {
   table {
     a {
       color: $bright-teal;
+
+      &.dropdown-item {
+        color: inherit;
+      }
```
jeremystretch, line 139: Why is this needed?

netbox/project-static/styles/overrides/_tabler.scss
```diff
@@ -70,6 +76,47 @@ table th.orderable a {
...
+
+// Fix secondary combo button contrast
+.btn-outline-secondary {
+  &.active {
+    color: var(--tblr-primary-fg);
+  }
+  &:hover {
+    color: var(--tblr-primary-fg);
+  }
+}
```
jeremystretch, line 93: This doesn't seem to have any effect? A button with `btn-outline-secondary` appears to render the same whether or not this rule is in place AFAICT.

jeremystretch (changes requested):

jeremystretch:
I realize this is subjective, but I'm not a fan of the brighter colors. As other have noted, it makes the text more difficult to read IMO.

Original:
screenshot1

Modified:
screenshot2

jeremystretch (approved):

jeremystretch:
I've attached the [reverted changes](https://github.com/user-attachments/files/16255067/pr16845_reverted.txt) for future reference.

PR #16846, opened 2024-07-08
Fixes #16843: potential fix for IKE policy mode issue with version2

### LINKED ISSUE #16843: VPN IKE policy API POST requires mode in the request body for IKE version 2
### Deployment Type

Self-hosted

### NetBox Version

v4.0.6

### Python Version

3.10

### Steps to Reproduce

1. Attempt to create a VPN IKE policy via API with version '2' in the request body and no 'mode' field specified or 'null' for the field's value.
2. Witness the response "The request failed with code 400 Bad Request: {'mode': ['This field is required.']"
3. OR receive a response The request failed with code 400 Bad Request: {'mode': ['null is not a valid choice.']}

If you attempt to include version 2 with a mode field specified, you will get this response:
{
  "__all__": [
    "Mode cannot be used for selected IKE version"
  ]
}

This is further reinforced as an issue by the web interface clearly stating these fields are not meant to be combined when IKE version 2 is specified:

image

### Expected Behavior

I expected a VPN IKE policy to be supported in creation via API via a POST request without specifying the 'mode' since IKEv2 does not support a 'main' or 'aggressive' mode like IKEv1.

IKEv2 as a protocol has nothing analogous to 'main mode' and 'aggressive mode' and therefore it shouldn't be required in the request when version 2 is provided in the request body. This is obviously understood in the web interface but not the API at this time.

### Observed Behavior

When attempting to create an IKEv2 policy via the API, if you do not include the mode, you will receive this in an error message from your server:

The request failed with code 400 Bad Request: {'mode': ['This field is required.']

or 

{
  "__all__": [
    "Mode cannot be used for selected IKE version"
  ]
}

locklearxd:
### Fixes: #16843

This change would support mode being an optional field and allow IKEv2 policies to be created via API POST requests.

### DIFF
--- a/netbox/vpn/api/serializers_/crypto.py
+++ b/netbox/vpn/api/serializers_/crypto.py
@@ -50,7 +50,8 @@ class IKEPolicySerializer(NetBoxModelSerializer):
         choices=IKEVersionChoices
     )
     mode = ChoiceField(
-        choices=IKEModeChoices
+        choices=IKEModeChoices,
+        required=False
     )
     proposals = SerializedPKRelatedField(
         queryset=IKEProposal.objects.all(),


jeremystretch (approved):
Thanks @locklearxd!

PR #16848, opened 2024-07-08
fixed #16847: updated font

### LINKED ISSUE #16847: Update font
### NetBox version

v4.1.0-dev

### Feature type

### Proposed functionality

Update font used throughout app to match brand guidelines.

### Use case

Continuity.

### Database changes

_No response_

### External dependencies

_No response_

andrewgormley:
### Fixes: #16847

Updated font to Plus Jakarta Sans


sleepinggenius2:
It is really bad practice to not define any fallback fonts. Instead, I would suggest using `--tblr-font-sans-serif: 'Plus Jakarta Sans', system-ui, sans-serif;` at the very least. Also, fonts are all currently served locally. It seems to be a regression to add a new dependency on an external CDN. We have contractors that use VDIs with very limited network access and they would not be able to load the font as currently defined, coupled with not having a fallback, the results would be indeterminate and undesirable from a UX standpoint.

arthanson (changes requested):
As mentioned by sleepinggennius, the font needs to be local - NetBox must be able to run without outside internet connection.

netbox/project-static/styles/_variables.scss
```diff
@@ -1,7 +1,7 @@
 // Global variables
 
 // Set base fonts
-$font-family-sans-serif: 'Inter';
+$font-family-sans-serif: 'Plus Jakarta Sans Variable', 'Inter', system-ui, sans-serif;
```
arthanson, line 4: If we are using and including Plus Jakarta, do we still need Inter as a fallback?  And if not can we remove it from package.json as a dependency?

jeremystretch: We can remove it. Inter was selected as a custom font over whatever Tabler provides as a default IIRC. If we're using a different custom font, Inter can be removed.

jeremystretch (commented):
I've noticed some odd spacing issues with the Jakarta font. For example, there is considerably less vertical space between the title and subtitle texts in the example below, even though element margin & padding remain unchanged. The subtitle text is also considerably larger in Jakarta, even though both are computed as 12px.

Original (Inter):
screenshot1

Modified (Jakarta):
screenshot2

netbox/project-static/styles/_variables.scss
andrewgormley: Done.

andrewgormley:
I believe the slight size different in the font faces is okay, but I have added some padding to the page header title to give it some breathing room.

jeremystretch (approved):

Issue #16849, opened 2024-07-08
Document hierarchy issues


andrewgormley:
### NetBox version

v4.1.0-dev

### Feature type

Change to existing functionality

### Proposed functionality

- [ ] (Misc) Lists do not contain only li elements [Multiple]
- [ ] (Misc) Heading elements not in sequential descending order (h5 sub-header) [Multiple]

### Use case

Fixing web accessibility issues ensures NetBox is best placed to serve users using screen readers and other accessibility tools.

PR #16875, opened 2024-07-10
16849 document hierarchy issues

andrewgormley:
### Fixes: #16849

- [x]  (Misc) Lists do not contain only li elements [Multiple]
The two main areas that were addressed were the tab navigation for the profile and object area and the related objects module on an object page. Both issues were fixed utilising `role="presentation"`. It's worth noting this issue does not exist for tab navs on object list pages due to them utilising ids, roles and arias correctly due to the navigation happening in-page rather than linking elsewhere.

- [x]  (Misc) Heading elements not in sequential descending order (h5 sub-header) [Multiple]
This issue was occurring because pages were going from `<h2>` > `<h5>` and as far as I can tell the h tags were only being used for styling reasons. I've fixed this by changed the header block to use `<h1>` and changing the `<h5>` card headings to `<h2>` and added the existing class `.h5` to ensure the styling remains the same.


netbox/templates/500.html
```diff
@@ -14,9 +14,9 @@
         <div class="row">
             <div class="col col-md-6 offset-md-3">
                 <div class="card border-danger mt-5">
-                    <h5 class="card-header">
+                    <h2 class="card-header h5">
```
jeremystretch, line 17: You're correct in your assumption that the heading elements are used in cases such as this one solely for styling. If there's a need to enforce sequential ordering of headings (and I'm not sure there is?), it would make more sense IMO to change this to a `<div>`. Keeping these as `<h2>` tags with the `h5` class is confusing.

More to the point, for card headers specifically, we should just tweak the `card-header` class as needed to obviate the need for the extra tag.

andrewgormley: In the same way that sighted users can glance at a page and get a sense of its contents, users of screen readers can do the same by navigating through headings. Clear and sequential headings save those who use screen readers time and frustration. [Source](https://dequeuniversity.com/rules/axe/4.9/heading-order) provided by google lighthouse.

I do agree having a `<h2>` and `.h5` together is odd, it was the best way to ensure the cards I was targeting will keep the correct style where as altering `.card-header` may be used in other areas where we do not want to make styling changes such as the dashboard cards.

jeremystretch (approved):

PR #16881, opened 2024-07-11
16819 highlight parent device in rack

### LINKED ISSUE #16819: "Highlight device in rack" button should link to parent device for child-type devices
### Deployment Type

Self-hosted

### NetBox Version

4.0.6

### Python Version

3.11

### Steps to Reproduce

1. Select a (racked) device of Parent/Child type Child.
2. Click "Highlight device in rack".

### Expected Behavior

Button links to <rack URL>?device=<parent device ID>, i.e. rack elevation view with parent device highlighted.

### Observed Behavior

Button links to <rack URL>?device=<child device ID>, i.e. rack elevation view with nothing highlighted, since child device is not racked directly.

arthanson:

### DIFF
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -31,9 +31,17 @@
                         <td class="d-flex justify-content-between align-items-start">
                             {% if object.rack %}
                                 {{ object.rack|linkify }}
-                                <a href="{{ object.rack.get_absolute_url }}?device={{ object.pk }}" class="btn btn-primary btn-sm d-print-none" title="{% trans "Highlight device in rack" %}">
-                                  <i class="mdi mdi-view-day-outline"></i>
-                                </a>
+                                {% if object.parent_bay %}
+                                    {% with object.parent_bay.device as parent %}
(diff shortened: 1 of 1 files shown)


netbox/templates/dcim/device.html
```diff
@@ -31,9 +31,17 @@ <h5 class="card-header">{% trans "Device" %}</h5>
...
-                                  <i class="mdi mdi-view-day-outline"></i>
-                                </a>
+                                {% if object.parent_bay %}
+                                    {% with object.parent_bay.device as parent %}
+                                        <a href="{{ object.rack.get_absolute_url }}?device={{ parent.pk }}" class="btn btn-primary btn-sm d-print-none" title="{% trans "Highlight device in rack" %}">
+                                          <i class="mdi mdi-view-day-outline"></i>
+                                        </a>
+                                    {% endwith %}
+                                {% else %}
+                                    <a href="{{ object.rack.get_absolute_url }}?device={{ object.pk }}" class="btn btn-primary btn-sm d-print-none" title="{% trans "Highlight device in rack" %}">
```
jeremystretch, line 41: Could we just change this part

```
?device={{ object.pk }}
```

to this?

```
?device={{ object.parent_bay.device.pk or object.pk }}
```

It would be preferable to avoid the redundant HTML is possible.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

PR #16908, opened 2024-07-12
Fix incorrect import in rest-api.md

eoswald:
No existing issue but it will save people some headache


jeremystretch (approved):
Thanks @eoswald!

PR #16923, opened 2024-07-16
Update CONTRIBUTING.md

dfex:
The GitHub reactions icon has been moved from the top right to the bottom left of messages in Issues - I was going insane trying to find it, so this might help someone in the future ; )

Wasn't sure if it was worth opening a bug report for two words to be changed in the documentation, but I'm happy to open one if this Pull gets squashed


jeremystretch (approved):
Thanks for catching this @dfex!

Issue #16394, opened 2024-06-03
Distinguish product types visually (OSS, Cloud, Enterprise)


andrewgormley:
### NetBox version

v4.0.3

### Feature type

New functionality

### Proposed functionality

Distinguish between OSS, Cloud and Enterprise products through visual changes when users are logged in. This should occur through slightly different branded experiences such as adding a label below the logo with the name of the product type. Also using an alternate colour/gradient to subtly alter the theme which is bespoke to that product type.

### Use case

Users will immediately know which application they are using. By adding this feature it will also open the door to future theming and branded experiences.

jeremystretch:
> This should occur through slightly different branded experiences such as adding a label below the logo with the name of the product type.

@andrewgormley how would you like to accomplish this? Do we need to package alternative logos (images), or will this be text-based?

Also, marking this as blocked by #15908, which will introduce support for declaring a NetBox "edition" (e.g. Cloud, Enterprise, community).

jeremystretch:
Just noting that this is no longer blocked as #15908 has been completed. Let's prioritize this work for next week.

PR #16924, opened 2024-07-16
16394 distinguish product edition

andrewgormley:
### Fixes: #16394

1. Added SVG files for light and dark netbox labs logos and geometric motif
2. Altered navigation to show specific logo and motif depending on netbox edition
3. Added subtext below logo to display edition for any edition that is not 'Community'


jeremystretch:
@andrewgormley looks like there's an issue with the SVG logo on top of the background image on mobile view:

screenshot

andrewgormley:
Changes:
- Removed labs logos
- Added sub text for all editions
- Cleaned up motif implementation
- Removed motif for mobile view (it's difficult to implement this in the limited space without it looking like a bug in an already small crowded area. Assuming the majority of users view netbox on desktop this would be a minimal issue.

Outstanding issues:
- in dark mode when navigating to a new page the side navigation flashes white
- on cloud/enterprise editions the motif is incorrectly positioned when opening a sub menu that causes the container to overflow and scrolling down. This issue is solved by using a background-image instead of a standalone  tag, however the background gradient needs to change based on the theme selected. I have not yet found a way to implement the background image and gradient together while also being able to change the gradient with the theme.

Screenshot 2024-07-17 at 23 22 05

andrewgormley:
the motif bug is now fixed

jeremystretch:
I've pushed a small change that should resolve the navbar flickering. @andrewgormley please let me know if this works for you.

jeremystretch (approved):

andrewgormley:
@jeremystretch I've just edited the style of the edition text to better match the now lowercase type. I think we're good to go on this one now.

Issue #15692, opened 2024-04-10
Add background jobs to plugin API


alehaa:
### NetBox version

v3.7.5

### Feature type

Change to existing functionality

### Proposed functionality

NetBox already includes the functionality to schedule jobs using either the `django-rq` library or the `core.Job` model combined with the `JobsMixin` class. Making the abstract job functionality publicly available to plugins allows decoupling from the backend of `django-rq` and a consistent experience in NetBox across all functionality. For this I propose:

1. Add `netbox.models.JobsMixin` to the list of API available model mixins [in the documentation](https://github.com/netbox-community/netbox/blob/b7668fbfc3a81abf2c6a8ace98047647fd9244c9/docs/plugins/development/models.md). This allows plugins to implement new models with Jobs enabled.

2. Add a new `BackgroundJob` class to implement the execution of the job's code, i.e. for consistency in calling `start`, `terminate` and setting error messages. This class should also be used in existing NetBox functionality to run background jobs for consistency. Below is a sample implementation from a plugin of mine for demonstration purposes.

    <details>

    ```Python
    import logging
    from abc import ABC, abstractmethod

    from rq.timeouts import JobTimeoutException

    from core.choices import JobStatusChoices
    from core.models import ObjectType, Job


    class BackgroundJob(ABC):
        """
        Background Job helper class.

        This class handles the execution of a background job. It is responsible for
        maintaining its state, reporting errors, and scheduling recurring jobs.
        """

        @classmethod
        @abstractmethod
        def run(cls, *args, **kwargs) -> None:
            """
            Run the job.

            A `BackgroundJob` class needs to implement this method to execute all
            commands of the job.
            """
            pass

        @classmethod
        def handle(cls, job: Job, *args, **kwargs) -> None:
            """
            Handle the execution of a `BackgroundJob`.

            This method is called by the Job Scheduler to handle the execution of
            all job commands. It will maintain the job's metadata and handle errors.
            For periodic jobs, a new job is automatically scheduled using its
            `interval'.


            :param job: The job to be executed.
            """
            try:
                job.start()
                cls.run(job, *args, **kwargs)
                job.terminate()

            except Exception as e:
                job.terminate(status=JobStatusChoices.STATUS_ERRORED, error=repr(e))
                if type(e) == JobTimeoutException:
                    logging.error(e)

            # If the executed job is a periodic job, schedule its next execution at
            # the specified interval.
            finally:
                if job.interval:
[...]

PR #16927, opened 2024-07-16
15692: Introduce background jobs

alehaa:
### Fixes: #15692 

This PR includes three new background job class to be used in NetBox and by NetBox plugins. They handle the necessary logic for starting and stopping jobs, including exception handling and rescheduling of recurring jobs.

This PR also implements the use of this framework for existing data source and script jobs, eliminating duplicate code and ensuring consistency.


netbox/core/management/commands/syncdatasource.py
```diff
@@ -34,13 +34,12 @@ def handle(self, *args, **options):
...
             self.stdout.flush()
-            try:
-                datasource.sync()
-                self.stdout.write(datasource.get_status_display())
-                self.stdout.flush()
-            except Exception as e:
-                DataSource.objects.filter(pk=datasource.pk).update(status=DataSourceStatusChoices.FAILED)
-                raise e
+
+            datasource.enqueue_sync_job()
```
jeremystretch, line 38: IMO we should leave this as-is and avoid the use of a background job entirely. We want the management command to block on the call to `sync()`.

netbox/core/models/jobs.py
```diff
@@ -224,6 +226,12 @@ def enqueue(cls, func, instance, name='', user=None, schedule_at=None, interval=
             job_id=uuid.uuid4()
         )
 
+        # Optionally, the job can be run immediately without being scheduled to run in the background.
+        if run_now:
+            func(job_id=str(job.job_id), job=job, **kwargs)
+            return job
```
jeremystretch, line 232: I could be missing something, but this seems like it defeats the purpose of using a job. Or if the goal is to reuse logic defined within a job, could we just call its `run()` method directly?

netbox/utilities/jobs.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        Enqueue a new `BackgroundJob`.
+
+        This method is a wrapper of `Job.enqueue()` using `handle()` as function callback. See its documentation for
+        parameters.
+        """
+        return Job.enqueue(cls.handle, *args, **kwargs)
+
+
+class ScheduledJob(BackgroundJob):
```
jeremystretch, line 80: Could we just collapse the functionality delivered by this class into BackgroundJob by adding an alternative `enqueue_once()` method to it? This would replicate the functionality of the `schedule()` method without requiring a separate class.

netbox/utilities/jobs.py
```diff
@@ -0,0 +1,196 @@
...
+            # If the job parameters haven't changed, don't schedule a new job and keep the current schedule. Otherwise,
+            # delete the existing job and schedule a new job instead.
+            if job.interval == interval:
+                return job
+            job.delete()
+
+        return cls.enqueue(instance=instance, interval=interval, *args, **kwargs)
+
+
+class SystemJob(ScheduledJob):
```
jeremystretch, line 153: Would it remove the need for this class if we allowed a Job to be enqueued without defining an object? I don't actually think it was intended to be required: The Job model currently requires an `object_type` but not an `object_id`. I'm open to removing that requirement if it simplifies this effort.

jeremystretch (commented):
@alehaa thank you for such a thorough and well-documented PR! Apologies again for the delay in review. At a high level, this all looks doable. I just have a few questions about the implementation details.

netbox/core/management/commands/syncdatasource.py
alehaa: When `enqueue_sync_job()` gets no request passed as argument, the job will be run in foreground and block until `sync()` has finished. See netbox/core/models/data.py line 169. Maybe a dedicated `background=False` parameter would make this more clear than just using some automation?

netbox/core/models/jobs.py
alehaa: My intention was to run a job in the foreground, but also create a `Job` object for documentation. Otherwise, jobs would run on the console, but there would be no audit trail. Now there's at least the information that a job was run, although the changelog objects are missing (maybe a topic for a future FR).

I agree that the method name `enqueue()` is a bit misleading. I could rename it to `run()` with the default parameter `background=True` to run jobs in the background, but also give the option to run it now. Since the `Job` API is not yet public for plugins, this should not break compatibility.

netbox/utilities/jobs.py
alehaa: Yep. The sole purpose of this class is to circumvent the required `instance` parameter of `Job.enqueue()` and adding the `setup()` method. I'll add a commit to make the `instance` parameter optional and thus removing the need for this class.

netbox/utilities/jobs.py
alehaa: I was hesitant at first because `ScheduledJob` uses a slightly different logic by hardcoding job names. However, I think it can be done and I give it a try.

netbox/utilities/jobs.py
```diff
@@ -0,0 +1,196 @@
...
+    job is already scheduled. Like a regular `BackgroundJob`, this class also accepts intervals.
+
+    The purpose of this class is to decouple jobs from the usual request-based approach. A practical example of this is
+    to schedule a periodic synchronization job for a particular object. All that matters is that the job is scheduled
+    and executed periodically. However, a new periodic job does not need to be scheduled every time the object is saved.
+    Calling the `schedule()` method of this class will ensure that the job's schedule is set up no matter how often the
+    method is called.
+    """
+
+    ENQUEUED_STATUS = [
```
alehaa, line 95: Should be moved into netbox/netbox/core/choices.py

netbox/utilities/jobs.py
```diff
@@ -0,0 +1,196 @@
...
+        JobStatusChoices.STATUS_PENDING,
+        JobStatusChoices.STATUS_SCHEDULED,
+        JobStatusChoices.STATUS_RUNNING,
+    ]
```
alehaa, line 99: Should be moved into netbox/netbox/core/choices.py

netbox/core/management/commands/syncdatasource.py
jeremystretch: Right, but there's no reason to enqueue a job at all; the management command intends to call `sync()` directly. This change can be reverted.

alehaa:
@jeremystretch I‘m a bit confused about your commits. You requested changes, I‘ve worked a day on them and just before uploading commits apear in this PR. 🧐

jeremystretch:
@alehaa apologies for the confusion; it had been several days with no response so I assumed you did not have time to make the changes. I've undertaken the work we discussed above as well as some general cleanup. We need to have all the v4.1 work merged by end of week.

jeremystretch:
@alehaa do you want to handle renaming ScheduledJob from here, or should I just wrap it up?

alehaa:
@jeremystretch no worries, but I’d like to keep it up from here. I see some of your code handles parts I‘ve not finished yet. Please give me till tomorrow to integrate your changes into mine and I‘ll upload the combination from both of our sources.

netbox/utilities/jobs.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @classmethod
+    def setup(cls, *args, **kwargs):
+        """
+        Setup a new `BackgroundJob` during plugin initialization.
+
+        This method should be called from the plugins `ready()` function to set up the schedule as early as possible.
+        For interactive setup of schedules (e.g. on user requests), use either `enqueue()` or `enqueue_once()` instead.
+        """
+        connection_created.connect(lambda sender, **signal_kwargs: cls.enqueue_once(*args, **kwargs))
```
alehaa, line 145: A minor problem with using the `connection_created' signal is that it's fired on every database connection. This includes every time a job is run by the rqworker. In general this should not have any side effects, but performance will be slightly degraded as the schedules of system background jobs will be checked every time.

To improve performance, I suggest adding an exit-early if `setup()` is not called by a WSGI process, and perhaps only for the first thread. That way, only a subset of NetBox processes will handle the setup of system background jobs. However, I'm not familiar enough with NetBox's WSGI internals to know how to detect these conditions.

alehaa:
@jeremystretch I've finished my work and added everything to this PR and its ready for review now. Please note that I did need to do a force push, as some of our works did conflict and I didn't want to open a third PR for this issue. Therefore you might need to to a `git fetch` followed by a `git reset --hard origin/15692-background-jobs` to get the current state on your local machine.

Despite the open tasks and some bug fixing, I've 
* integrated additional tests for the `BackgroundJob` framework
* did some cleanup of legacy code in 309ad29
* implemented to use queue `low` for system background jobs by default
* switched `run()` from a `@classmethod` to a regular method to simplify implementation and give options for possible future extensions

Also note the open comment on the `setup()` method above, maybe you can take a look at it? Feel free to add or request changes to the code as you see fit.

netbox/netbox/settings.py
```diff
@@ -124,7 +124,9 @@
 METRICS_ENABLED = getattr(configuration, 'METRICS_ENABLED', False)
 PLUGINS = getattr(configuration, 'PLUGINS', [])
 PLUGINS_CONFIG = getattr(configuration, 'PLUGINS_CONFIG', {})
-QUEUE_MAPPINGS = getattr(configuration, 'QUEUE_MAPPINGS', {})
+QUEUE_MAPPINGS = getattr(configuration, 'QUEUE_MAPPINGS', {
+    None: 'low',
+})
```
jeremystretch, line 129: We should leave this as-is, because it would effect a change to the default configuration, which would require a separate FR. I assume this is for system jobs? If so, we could instead explicitly queue the job in `RQ_QUEUE_LOW`.

netbox/utilities/jobs.py
jeremystretch: I may not be clear on the exact functionality here. Would it suffice to just call `enqueue_once()` directly from the AppConfig's `ready()` method, rather than attaching it to a signal?

The `connection_created` signal isn't really tenable for this, as database connections can be created quite frequently depending on the setup, especially when a large number of WSGI workers are in use.

netbox/netbox/settings.py
alehaa: I've reverted b17b2050df953427769ab6a337c6995ac2a60615. For specifying a custom queue I would not hardcode the low queue, but wait for #16516 to be resolved. As all enqueue methods of `BackgroundJob` pass through `kwargs`, a future `queue` parameter (as suggested in this FR) will be passed through as well.

netbox/utilities/jobs.py
jeremystretch: True, this is a bit tricky. It feels like there should be a canonical convention for this, but I'm not sure what it would be.

Unfortunately we're on a clock here, as the first v4.1 beta release is due August 5th, and I'd rather not hold up the PR just for this issue. How would you feel about splitting out the auto-run functionality for now? We can merge the remainder of the PR largely as-is (it probably deserves one final review), and revisit the auto-run piece individually. If we arrive at a workable solution, I'd be comfortable slipping that in during the beta period ahead of the final release.

netbox/core/models/jobs.py
```diff
@@ -197,25 +199,34 @@ def terminate(self, status=JobStatusChoices.STATUS_COMPLETED, error=None):
         job_end.send(self)
 
     @classmethod
-    def enqueue(cls, func, instance, name='', user=None, schedule_at=None, interval=None, **kwargs):
+    def enqueue(cls, func, instance=None, name='', user=None, schedule_at=None, interval=None, immediate=False, **kwargs):
```
arthanson, line 202: I think it might be cleaner to break this into a base function with two other functions enqueue and run? to separate out the immediate keyword and schedule_at / interval.   In the immediate case you are not really enqueueing anything. I'm not sure interval is supported when using Immediate?  It is not used by the management command.

netbox/utilities/jobs.py
alehaa: That's fine by me. I've removed the `setup()` method and documentation for system jobs. I think we can close this FR without auto-registration and solve this task with #16971.

The code is ready to be used as a system background job, only the auto-registration isn't solved yet. Although not recommended, anyone can already call `enqueue_once()` in the `ready()` method of AppConfig and it *should* work (no guarantee).

netbox/utilities/tests/test_jobs.py
```diff
@@ -0,0 +1,134 @@
...
+    def test_enqueue_once_after_enqueue(self):
+        instance = Job()
+        job1 = TestBackgroundJob.enqueue(instance, schedule_at=self.get_schedule_at())
+        job2 = TestBackgroundJob.enqueue_once(instance, schedule_at=self.get_schedule_at(2))
+
+        self.assertNotEqual(job1, job2)
+        self.assertRaises(Job.DoesNotExist, job1.refresh_from_db)
+        self.assertEqual(TestBackgroundJob.get_jobs(instance).count(), 1)
+
+    def test_enqueue_system(self):
```
arthanson: Would be good to add a test with the name parameter.

netbox/core/models/jobs.py
```diff
@@ -197,25 +199,34 @@ def terminate(self, status=JobStatusChoices.STATUS_COMPLETED, error=None):
...
             user: The user responsible for running the job
             schedule_at: Schedule the job to be executed at the passed date and time
             interval: Recurrence interval (in minutes)
+            immediate: Run the job immediately without scheduling it in the background. Should be used for interactive
+                management commands only.
         """
-        object_type = ObjectType.objects.get_for_model(instance, for_concrete_model=False)
-        rq_queue_name = get_queue_for_model(object_type.model)
+        if schedule_at and immediate:
+            raise ValueError("enqueue() cannot be called with values for both schedule_at and immediate.")
```
arthanson, line 217: Should be internationalized with _(

arthanson (changes requested):

netbox/extras/management/commands/runscript.py
```diff
@@ -120,40 +76,31 @@ def _run_script():
...
             for field, errors in form.errors.get_json_data().items():
                 for error in errors:
                     logger.error(f'\t{field}: {error.get("message")}')
-            job.status = JobStatusChoices.STATUS_ERRORED
-            job.save()
+            raise CommandError()
+
+        # Execute the script.
+        ScriptJob = import_string("extras.jobs.ScriptJob")
+        job = ScriptJob.enqueue(
```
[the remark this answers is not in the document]
alehaa: Seems I forgot one call with this parameter in 257976d815605206686ebbf1c43db149131fe0fe. Fixed with ecf8e79fddcfd5b885fac3c29231edbcd73a12d8.

netbox/utilities/tests/test_jobs.py
alehaa: @arthanson can you elaborate on this please? Passing `name` in `enqueue()` and `enqueue_once()` is not supported as documented. Test for automatic setting of the job's `name` are implemented in `BackgroundJobTest`.

netbox/core/models/jobs.py
alehaa: There was an earlier discussion about naming the parameter and the consequences. You could split this into two methods, both reusing a third internal `_create()` method to make the parameters and behavior a bit clearer to the user. This would also require an additional proxy method in `BackgroundJob`. I'd like to leave the decision to @jeremystretch.

netbox/extras/api/views.py
```diff
@@ -273,10 +273,9 @@ def post(self, request, pk):
             raise RQWorkerNotRunningException()
 
         if input_serializer.is_valid():
-            Job.enqueue(
-                run_script,
+            ScriptJob = import_string("extras.jobs.ScriptJob")
```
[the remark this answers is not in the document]
jeremystretch: I think this is needed to avoid a circular import, however it should be fine to import `ScriptJob` directly (`from extras.jobs import ScriptJob`) at this point.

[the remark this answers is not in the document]
alehaa: This is a relict from using `BackgroundJob` for syncing data sources to avoid circular imports. For `ScriptJob` this is not necessary, so I've changed as requested in 7f0a4e3c34c06891d225535c36e3389fa6a8705d.

netbox/utilities/tests/test_jobs.py
arthanson: Partly it was used in runscript but I guess that might have been a mistake as it looks like it is removed now, but I'm not understanding why it is a parameter if it can't be passed in?

arthanson:
BTW @alehaa IMHO excellent work on this, both the writeup in the issue and the PR itself is very high quality and impressive.  Nice work!

netbox/utilities/tests/test_jobs.py
alehaa: One can pass a `name` to a `Job` object when using it standalone. `BackgroundJob` augments `Job` and sets the `name` parameter automatically, so it must not be set when calling `BackgroundJob.enqueue()`.

I forgot removing the parameter in `runscript`. This was fixed with ecf8e79fddcfd5b885fac3c29231edbcd73a12d8.

jeremystretch:
Seems like we're just about there, however I still have one concern: As it stands, I expect the average user will have difficulty distinguishing between the Job database model and the BackgrounJob runner class. I'd like to rename `BackgroundJob` to `JobRunner` for clarity. IMO this better conveys the relationship between classes which embody the programming logic and the database records which track its execution.

jeremystretch (approved):

jeremystretch:
Thanks again for all your work on this @alehaa!

PR #16945, opened 2024-07-19
7025 circuit redundancy

arthanson:


netbox/templates/circuits/circuitgroupassignment.html
```diff
@@ -0,0 +1,41 @@
...
+            <th scope="row">{% trans "Priority" %}</th>
+            <td>{{ object.priority }}</td>
+          </tr>
+        </table>
+      </div>
+      {% include 'inc/panels/tags.html' %}
+      {% plugin_left_page object %}
+    </div>
+    <div class="col col-md-6">
+      {% include 'inc/panels/related_objects.html' %}
```
jeremystretch, line 37: This panel is not needed.

jeremystretch, line 37: Missing the custom fields panel.

jeremystretch (changes requested):
Some other notes:

- The CircuitGroup view should include a "circuits" tab to easily list assigned circuits.

- There should be some convenient path to assign a circuit to a group from the circuit view.

- Looks like we're missing a [search index](https://netbox.readthedocs.io/en/stable/development/adding-models/#9-create-a-searchindex-subclass) for CircuitGroup.

netbox/circuits/models/circuits.py
```diff
@@ -151,6 +153,71 @@ def clean(self):
...
+
+    def __str__(self):
+        return self.name
+
+    def get_absolute_url(self):
+        return reverse('circuits:circuitgroup', args=[self.pk])
+
+
+class CircuitGroupAssignment(CustomFieldsMixin, ExportTemplatesMixin, TagsMixin, ChangeLoggedModel):
+    circuit = models.ForeignKey(
```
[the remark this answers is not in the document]
arthanson: I looked at it but it is never used anywhere and adds complexity to the related_counts.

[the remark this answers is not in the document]
arthanson: RE "The CircuitGroup view should include a "circuits" tab to easily list assigned circuits.

There should be some convenient path to assign a circuit to a group from the circuit view.
"
For the detail view the "Related Objects" I thought took the place of that, that is the way all the other group ones do it from what I can see. 

Also, I don't see a paradigm for the other group objects having an assignment on the object page (Tunnel Group, Tenant Group)?  Do we have a UI pattern for this somewhere?

netbox/templates/circuits/circuitgroupassignment.html
jeremystretch: Disregard; we'll keep this in place instead of adding a separate tab view.

Issue #16392, opened 2024-06-03
Collapse side navigation


andrewgormley:
### NetBox version

v4.0.3

### Feature type

New functionality

### Proposed functionality

Add the ability to collapse the the side navigation, this will occur when a user clicks on a 'menu' icon, clicking the same icon will also expand the menu to it's default state. In the collapsed state the text will be hidden, showing the icon only. On hover the menu will be temporarily expand, when the user's cursor moves away from the menu it will return to it's collapsed state. The collapsed menu's sub-items should still be accessible via tabbing. The user's chosen state for the menu should be saved for future sessions.

Sub tasks
- [ ] Change top navigation to be full width and sit on top of side nav
- [ ] Move logo to top navigation
- [ ] Increase width of search bar
- [ ] Add menu icon which collapses and expands the side navigation

### Use case

Enabling users the ability to collapse the side menu will create more space to view content such as large tables with more ease.

jeremystretch:
We've had to bump this from v4.1 due to technical challenges. I'll leave this open for anyone else who would like to make an attempt.

mugglettuk:
It might also be a good idea to be able to just re-size the windows on the fly. There's so much white space in the left had window that is not needed most of the time. It would then allow more columns to be added to the hand pane that contain more information that other users may need to display instead of going in to the device to get the info. Small screens need the ability to be able to re-size.

philmargetts:
> We've had to bump this from v4.1 due to technical challenges. I'll leave this open for anyone else who would like to make an attempt.

Hi there - any details of the challenges faced? Keen to see this feature :)

flakrat:
Looking forward to this feature, specifically when viewing rack elevations!

PieterL75:
Maybe include 2 other requests in this one ?
- pad sub menu items a little https://github.com/netbox-community/netbox/issues/17675
- set the 'top' menu items (org, racks, devices, ...) background a little different (darker/lights) than the rest of the menu, making them easier to distinguish when they are expanded (could not find that FR though)

pheus:
Hi folks! Quick upstream update on this highly requested feature to help keep it from going stale:

- Tabler (the UI framework NetBox uses) is tracking sidebar improvements here: https://github.com/tabler/tabler/issues/1803
- A maintainer recently indicated they’re targeting this work for **v1.5**: https://github.com/tabler/tabler/issues/1803#issuecomment-3220852507

[...]

PR #16948, opened 2024-07-19
16392 collapse side navigation

andrewgormley:
### Fixes: #16392

1. Created toggle button
2. Utilised existing script to implement 'pin' and 'hover' behaviour to side navigation
3. Moved logo to top bar
4. Top bar now sits on top of side navigation
5. Increased width of search input
6. Added padding to bottom of navigation to ensure links are not covered by native browser link hover state

Outstanding issues:
- [ ] Issue - when sub menu != current page then the sub menu does not collapse, Ideally we want to mirror the behaviour of the active sub menu.
- [ ] Issue - You need to click the toggle button twice to activate
- [ ] Issue - Active state is only applied to collapsed menu item after a user has hovered on the side nav, ideally this is implemented on page load.
- [ ] Issue - Tabbing through menu items does not expand the side navigation
- [ ] Issue - Changing the side nav from un-pinned to pinned does not update the state of the active nav item if there is one available.


jeremystretch:
Closing this out as we appear to have hit a roadblock.

PR #16951, opened 2024-07-19
Closes #16926 Adds various USB cable types to front/rear ports and cable choices

### LINKED ISSUE #16926: Add USB type to Front/Rear Ports choices
### NetBox version

v4.0.7

### Feature type

Data model extension

### Proposed functionality

I propose to add USB choices for Front/Rear ports.
Changes proposed include modifying dcim/choices.py to add USB front and rear ports.
No other data or UI changes required beyond the addition of USB choices

### Use case

Use Cases:
* Patch Panels with modular/keystone ports and pass-through USB keystones. 
* USB WiFi and Network Adapters

Specific Scenario:
Firewalla sells a 1u rack adapter/shelf for their Firewalla GOLD line of firewall appliances.  It has an optional USB Wifi adapter that they provide keystone slots to move from the rear of the firewalla appliance to the front of the 1u rack adapter.

General other benefits:
Netbox already supports USB cable types for consoles, but does not have a way to model using a usb cable and a patch panel to relocate the port from the rear/inside the rack to the front/outside of the rack.  These cables should be treated as any other patch cable.
This modification could also support USB attached celluar backup modems, and USB attached SMS gateways.

Code changes already completed in fork, just waiting on approval to submit a PR.

### Database changes

n/a -- only requires adding lines to dcim/choices.py

### External dependencies

n/a

bchow000:
### Fixes: #16926

Add front/rear port and cable choices for various USB styles.

USB-A
USB-B
USB-C
USB-Mini-A
USB-Mini-B
USB-Micro-A
USB-Micro-B
USB-Micro-AB

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -1315,6 +1315,14 @@ class PortTypeChoices(ChoiceSet):
     TYPE_URM_P2 = 'urm-p2'
     TYPE_URM_P4 = 'urm-p4'
     TYPE_URM_P8 = 'urm-p8'
+    TYPE_USB_A = 'usb-a'
+    TYPE_USB_B = 'usb-b'
+    TYPE_USB_C = 'usb-c'
+    TYPE_USB_MINI_A = 'usb-mini-a'
+    TYPE_USB_MINI_B = 'usb-mini-b'
+    TYPE_USB_MICRO_A = 'usb-micro-a'
+    TYPE_USB_MICRO_B = 'usb-micro-b'
+    TYPE_USB_MICRO_AB = 'usb-micro-ab'
     TYPE_OTHER = 'other'
 
     CHOICES = (
@@ -1375,6 +1383,19 @@ class PortTypeChoices(ChoiceSet):
             ),
         ),
         (
+            _('USB'),
+            (
+                (TYPE_USB_A, 'USB-A'),
+                (TYPE_USB_B, 'USB-B'),
+                (TYPE_USB_C, 'USB-C'),
+                (TYPE_USB_MINI_A, 'USB-Mini-A'),
+                (TYPE_USB_MINI_B, 'USB-Mini-B'),
+                (TYPE_USB_MICRO_A, 'USB-Micro-A'),
+                (TYPE_USB_MICRO_B, 'USB-Micro-B'),
+                (TYPE_USB_MICRO_AB, 'USB-Micro-AB'),
+            ),
+        ),
+        (
             _('Other'),
             (
                 (TYPE_OTHER, _('Other')),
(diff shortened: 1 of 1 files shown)


Brian-Chow:
Can we get someone to touch the PR to prevent it from going stale, please?

netbox/dcim/choices.py
```diff
@@ -1412,6 +1433,14 @@ class CableTypeChoices(ChoiceSet):
...
     TYPE_AOC = 'aoc'
     TYPE_POWER = 'power'
+    TYPE_USB_A = 'usb-a'
+    TYPE_USB_B = 'usb-b'
+    TYPE_USB_C = 'usb-c'
+    TYPE_USB_MINI_A = 'usb-mini-a'
+    TYPE_USB_MINI_B = 'usb-mini-b'
+    TYPE_USB_MICRO_A = 'usb-micro-a'
+    TYPE_USB_MICRO_B = 'usb-micro-b'
+    TYPE_USB_MICRO_AB = 'usb-micro-ab'
```
jeremystretch, line 1443: These are not valid cable types. The A, B, and C designation refer to terminations only. (For instance, a cable can have a USB-A termination on one end and a USB-B termination on the other.)

netbox/dcim/choices.py
```diff
@@ -1374,6 +1382,19 @@ class PortTypeChoices(ChoiceSet):
...
+            _('USB'),
+            (
+                (TYPE_USB_A, 'USB-A'),
+                (TYPE_USB_B, 'USB-B'),
+                (TYPE_USB_C, 'USB-C'),
+                (TYPE_USB_MINI_A, 'USB-Mini-A'),
+                (TYPE_USB_MINI_B, 'USB-Mini-B'),
+                (TYPE_USB_MICRO_A, 'USB-Micro-A'),
+                (TYPE_USB_MICRO_B, 'USB-Micro-B'),
+                (TYPE_USB_MICRO_AB, 'USB-Micro-AB'),
```
jeremystretch, line 1395: The human-friendly labels for these types should match what is already defined for console port types in the interest of consistency.

jeremystretch (changes requested):

netbox/dcim/choices.py
Brian-Chow: I will resolve/adjust and resubmit.  Thank you.

Brian-Chow:
Comments addressed and updated.  Thanks!

netbox/dcim/choices.py
```diff
@@ -1444,6 +1473,18 @@ class CableTypeChoices(ChoiceSet):
...
+        (
+            _('USB'), (
+                (TYPE_USB_1, 'USB 1.0'),
+                (TYPE_USB_1_1, 'USB 1.1'),
+                (TYPE_USB_2, 'USB 2.0'),
+                (TYPE_USB_3, 'USB 3.0'),
+                (TYPE_USB_3_1, 'USB 3.1'),
+                (TYPE_USB_3_2, 'USB 3.2'),
+                (TYPE_USB_4, 'USB 4'),
+                (TYPE_USB_4_2, 'USB 4 2.0'),
```
jeremystretch, line 1485: Not all of these specify a new cable type. I spent a few minutes looking over the [Wikipedia page](https://en.wikipedia.org/wiki/USB) but it wasn't clear to me what the different classes of USB cabling actually are. Let's just add a single generic "USB" type for now. It doesn't need to be in a separate subgroup; it can be listed before "power.")

jeremystretch (changes requested):

netbox/dcim/choices.py
Brian-Chow: @jeremystretch I'm ook going to generic USB overall, but I would like to make a case for distinction between at least USB 2 and USB 3 series as the bandwidth capacity is vastly different with USB2 capping out at 480gb and USB3 series ranging between 5Gb and 20gb.  I would think it would be more in line with ok there various data capacity cables, but if the distinction isn't deemed important, I'll go ahead and make the suggested generic modifications.

bctiemann (approved):

Issue #11969, opened 2023-03-13
Airflow attribute for PSU and fan modules


jbemmel:
### NetBox version

latest

### Feature type

Data model extension

### Proposed functionality

The schema at https://github.com/netbox-community/devicetype-library/blob/master/schema/moduletype.json currently does not allow 'airflow' attributes.

Assuming that is in line with what's currently in Netbox, I'd like to propose adding 'airflow' support for modules and racks

In typical systems, fan and psu modules have a specific airflow direction. It is not a property of the base system - it depends on the modules being used. Moreover, data centers are designed with a specific airflow in mind - this 'design intent' should be captured and used to reduce provisioning errors, by disallowing combinations of devices with incompatible airflow in the same rack (and similar for modules in the same device)

### Use case

Supporting 'airflow' at the module and rack level allows for proper modeling of the system. The rack level 'airflow' would allow users to specify a filter for the devices and modules to be installed in that rack

### Database changes

Adds 'airflow' field per module/rack (as a mixin, like weight), plus logic to validate that all modules added support the same airflow direction, and filtering of module_types presented given a device in a rack with specified airflow

### External dependencies

none

danner26:
I absolutely agree, we do need to add `airflow` as a property to modules as well

PieterL75:
Maybe a check also ? that if the device is has a certain airflow, that the modules have to have the same ?
Not sure if that is always a good idea...

jeremystretch:
How would these values map to the options for device airflow?

danner26:
As a rough idea, we could have an airflow attribute on devices called something like `airflow: module-regulated` which then on the NetBox backend is determined by the installed modules which have the airflow attribute to specify the direction

Working with the devicetype-library I have noticed a lot of switches chassis' will have a sku for no PSU/FANs, and then depending on which PSU is installed that determins the direction of the airflow

arthanson:
@jeremystretch Would this then have the same airflow choices as Device (below), and if-so, if Mixed would the filter allow all device airflow in, also for devices without airflow defined, would they be allowed to be in the rack if airflow was set on the rack?  Would the logic be that airflow for rack / device has to match unless it is Mixed or Passive?

```
    CHOICES = (
        (AIRFLOW_FRONT_TO_REAR, _('Front to rear')),
        (AIRFLOW_REAR_TO_FRONT, _('Rear to front')),
        (AIRFLOW_LEFT_TO_RIGHT, _('Left to right')),
[...]

PR #16967, opened 2024-07-24
11969 airflow

arthanson:
### Fixes: #11969 

Add airflow to:
* ModuleType
* RackType
* Rack


jeremystretch (approved):
Thanks @arthanson!

PR #16974, opened 2024-07-24
Closes #16928: Add help center button to top nav

### LINKED ISSUE #16928: Add extra UI element to non-community instances
### NetBox version

v4.0.7

### Feature type

### Proposed functionality

On instances where `settings.RELEASE.edition != 'community'`, add a [question-answer icon](https://fonts.google.com/icons?selected=Material+Icons+Outlined:question_answer:&icon.query=chat&icon.set=Material+Icons&icon.size=24&icon.color=%23e8eaed) element In the scroll-sticky portion of the base template. Ideal placement is to the left of the light mode / dark mode toggle element. The element's `id` attribute should be stable, and similar to `launch-help-center`.

### Use case

In some NetBox product editions, this element will serve as a trigger to launch an interactive help center.

### Database changes

None

### External dependencies

None

jeremystretch:
### Closes: #16928

Embed a `help-center` anchor element if the `help-center` feature is enabled.

### DIFF
--- a/netbox/templates/inc/user_menu.html
+++ b/netbox/templates/inc/user_menu.html
@@ -1,11 +1,18 @@
 {% load i18n %}
 {% load navigation %}
 
+{% if 'help-center' in settings.RELEASE.features %}
+  {# Help center control #}
+  <a href="#" class="nav-link px-1" aria-label="{% trans "Help center" %}">
+    <i class="mdi mdi-forum-outline"></i>
+  </a>
+{% endif %}
+
 {% if request.user.is_authenticated %}
   {# Notifications #}
   {% with notifications=request.user.notifications.unread.exists %}
-    <div class="nav-item dropdown">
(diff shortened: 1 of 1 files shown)


netbox/templates/inc/user_menu.html
```diff
@@ -1,11 +1,18 @@
 {% load i18n %}
 {% load navigation %}
 
+{% if 'help-center' in settings.RELEASE.features %}
```
arthanson, line 4: I don't see features in release.yaml? Is that added in enterprise, cloud and can we get it documented somewhere (and what features there are) if it isn't already?

arthanson (approved):

netbox/templates/inc/user_menu.html
jeremystretch: This was part of #15908. We didn't document the `features` as it should only be used internally.

PR #16982, opened 2024-07-25
Closes #16964: Validate password when creating a new user or updating password for an existing user

### LINKED ISSUE #16964: Configured password validators are not executing when creating a new user or changing the password for existing user
### Deployment Type

NetBox Cloud

### NetBox Version

v4.0.7

### Python Version

3.10

### Steps to Reproduce

1. Configure password validation in the `netbox/netbox/configuration.py` file. Refer below code for the same:
```
AUTH_PASSWORD_VALIDATORS = [
    {
        'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
        'OPTIONS': {
            'min_length': 8,
        }
    }
]
```
2. Login as a superuser.
3. Click "Admin" -> "Authentication" -> "Users"
4. Click on "+ Add".
5. Fill the details for the user and set password and confirm password as `pass1`.
6. Click "Create".
7. User gets created successfully.

Instead of adding a new user from step 4 to 6, we can also just update the password for an existing users, and set it as `pass1`. The result would be the same.

### Expected Behavior

User creation or password updation should fail from "Admin" -> "Authentication" -> "Users".

### Observed Behavior

User is being created without adhering to the configured password policy.

nishant131:
### Fixes: #16964

- Call `validate_password()` on the configured password validators, while cleaning the `UserForm` data.

### DIFF
--- a/netbox/users/forms/model_forms.py
+++ b/netbox/users/forms/model_forms.py
@@ -1,6 +1,6 @@
 from django import forms
 from django.conf import settings
-from django.contrib.auth import get_user_model
+from django.contrib.auth import get_user_model, password_validation
 from django.contrib.postgres.forms import SimpleArrayField
 from django.core.exceptions import FieldError
 from django.utils.safestring import mark_safe
@@ -226,6 +226,8 @@ class UserForm(forms.ModelForm):
         # Check that password confirmation matches if password is set
         if self.cleaned_data['password'] and self.cleaned_data['password'] != self.cleaned_data['confirm_password']:
             raise forms.ValidationError(_("Passwords do not match! Please check your input and try again."))
+        if self.cleaned_data['password'] and self.cleaned_data['confirm_password']:
+            password_validation.validate_password(self.cleaned_data['confirm_password'], self.instance)
 
 
 class GroupForm(forms.ModelForm):


jeremystretch:
Hi @nishant131, thanks for digging into this! In addition to the form validation, we also need to enforce the validation under the REST serializer for the User model.

I went ahead and did this (and added the relevant tests), but when I pushed my commit it ended up creating a new branch under the `netbox` repo. I'm not sure what happened. :confused: 

I've opened PR #16990, which includes your initial commit, to replace this. Sorry for the mix-up.

nishant131:
Hi @jeremystretch, I think the new branch got created on `netbox` repo as I had unchecked `Allow edits by maintainers` in the PR. Since, it was my first contribution to netbox, I wasn't sure whether to allow it or not 😕.
Maybe, we can add this in the PR submission guideline doc.

Thank you for adding the validation under the REST serializer, I'll incorporate that in my setup as well 🙂

jeremystretch:
>  I think the new branch got created on netbox repo as I had unchecked Allow edits by maintainers in the PR

That's my suspicion as well, however in the past GitHub has rejected the push in such cases. Maybe they've changed things. :man_shrugging:

PR #16983, opened 2024-07-25
Closes #10500: Introduce support for nested modules

### LINKED ISSUE #10500: Allow modules to contain module bays and modules to be parents of other modules.
### NetBox version

v3.3.4

### Feature type

### Proposed functionality

Some network devices have line cards inserted into slots that provide sub-slots.

Example:  Juniper MX Chassis
FPC0 - Line card module
-- MIC0 - Sub-module
-- -- PIC0 - Sub-Sub-module
-- -- -- Interface et-0/0/0
-- MIC1 - Sub-module
-- -- PIC2 - Sub-Sub-module
-- -- -- Interface et-0/2/0
-- -- -- Interface et-0/2/1

Thought would need to be put into using the {module} variable in a nested fashion to permit automatic addition of interfaces based on the module template.  In some cases, allowing a limited degree of math in the calculation of interface #s would be needed.  But from a pure inventory standpoint, having the ability to populate these the way some network devices see them would be very useful.

### Use case

Inventory, Generation of interfaces from templates, modeling device configurations.

### Database changes

_No response_

### External dependencies

_No response_

arthanson:


jeremystretch:
@arthanson I've fleshed this out some more, but there's still some work to do. Specifically, the validation in place to prevent recursive module assignment does not appear to be working; please test this. Additionally, please update the tests to incorporate the addition of module bays assigned to modules. Thanks!

jeremystretch (approved):

Issue #16850, opened 2024-07-08
Table Issues


andrewgormley:
### NetBox version

v4.1.0-dev

### Feature type

Change to existing functionality

### Proposed functionality

- [ ] (Misc) Checkbox missing label [Multiple]
- [ ] (Misc) Missing link text on hidden table column [Multiple]
- [x] (Tables) unlabelled column header for checkbox column [Multiple]
- [ ] (Tables) Ensure tables use table and row headers to describe cells [Multiple]
- [x] (Tables) Row actions missing accessible names [Multiple]
- [x] (Tables) When table column is sorted the 'x' icon is missing link text [Multiple]

### Use case

Fixing web accessibility issues ensures NetBox is best placed to serve users using screen readers and other accessibility tools.

andrewgormley:
-  (Misc) Missing link text on hidden table column [Multiple]
-  (Tables) unlabelled column header for checkbox column [Multiple]

These two issues appear to be related to the actions column, the missing link is due to the sortable option being set to true even though there is no label. I believe to fix both issues we should set the sortable option to false and label the column as 'actions' or similar. 

![Screenshot 2024-07-24 at 12 06 02](https://github.com/user-attachments/assets/5fc4a61e-3b78-434c-9fad-c3f5645a8b45)

andrewgormley:
- (Misc) Checkbox missing label [Multiple]

This issue is due to the checkbox not having an associated label, however being in a table this is a bit of an edge case, this might be fixable with an aria label

![Screenshot 2024-07-24 at 12 06 47](https://github.com/user-attachments/assets/f52ff78d-008a-4532-9346-f73cb4418edb)

andrewgormley:
I've moved more table issues from #16852  to here.

jeremystretch:
Some of the above items (checked) were addressed in PR #16986.

PR #16986, opened 2024-07-25
fixes issues in #16850

andrewgormley:
### Fixes: #16850

- [ ]  (Misc) Checkbox missing label [Multiple]
- [ ]  (Misc) Missing link text on hidden table column [Multiple]
- [x]  (Tables) unlabelled column header for checkbox column [Multiple]
- [ ]  (Tables) Ensure tables use table and row headers to describe cells [Multiple]
- [x]  (Tables) Row actions missing accessible names [Multiple]
- [x]  (Tables) When table column is sorted the 'x' icon is missing link text [Multiple]

### DIFF
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -173,6 +173,7 @@ class ToggleColumn(tables.CheckBoxColumn):
             kwargs['attrs'] = {
                 'th': {
                     'class': 'w-1',
+                    'aria-label': 'Select all',
                 },
                 'td': {
                     'class': 'w-1',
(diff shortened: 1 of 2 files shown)


netbox/netbox/tables/columns.py
```diff
@@ -173,6 +173,7 @@ def __init__(self, *args, **kwargs):
             kwargs['attrs'] = {
                 'th': {
                     'class': 'w-1',
+                    'aria-label': 'Select all',
```
jeremystretch, line 176: Would it be better to put the label on the checkbox itself, rather than the `<th>` element?

andrewgormley: The error thrown was for a missing label in the table header itself rather than the checkbox within the table header

jeremystretch (approved):

PR #16994, opened 2024-07-26
Fixes #16782: Add object filtering for custom fields

### LINKED ISSUE #16782: Custom fields - Allow use of query params for filtering object selection
### NetBox version

v4.0.6

### Feature type

### Proposed functionality

Object type custom fields are very useful for creating new relationships between objects, but it has no ability to filter the selection list, any object of that type can be selected which may not be desirable depending on the intended purpose of the custom field.

The DynamicModelChoice mixins already support filtering the API requests using the query_params dict, I propose that this be added as an option to specify in the custom field definition.

### Use case

Allowing filtering of the object choice selection would allow additional business logic to be applied and provide a more intuitive experience for the end user who is using the field.
For example, I have a custom field defined on Circuits to select an IP Address - with a query params definition I could limit it to show only 'active' IP addresses, and chain it to tenant field with $tenant to only list IPs that are assigned to the tenant of that circuit.

### Database changes

CustomField model will need an additional field to store the parameter

### External dependencies

none

samk-acw:
### Fixes: #16782

(new pr rebased off feature branch)
- adds 'related_object_filter' field to CustomField model and relevant form, template and api serializer
- updated CustomField's to_form_field method to load the query_param data
- added documentation
- currently I've chosen to not to include any additional validation as it feels fairly fail-safe already; json syntax is already validated by the field type, and entering an invalid query_params definition results in the filter returning all values, which is the default behavior anyway


jeremystretch (approved):
@samk-acw thanks for jumping on this so quickly!

PR #17003, opened 2024-07-26
Use `provider_id` instead of `account_id` in url_params

### LINKED ISSUE #16963: Providers list view "ACCOUNT COUNT" column using wrong attribute name in link
### Deployment Type

Self-hosted

### NetBox Version

v4.0.7

### Python Version

3.10

### Steps to Reproduce

1. Create at least two Providers
2. Create at least two Provider Accounts linked to multiple different Providers
3. Navigate to Providers list view and click on a non-zero "ACCOUNT COUNT" (notice the Provider name)
4. Notice that the url filter is using `account_id` instead of `provider_id` which results in the list not being properly filtered for that Provider

### Expected Behavior

The resulting Provider Account list should be filtered to only display Accounts belonging to that Provider

### Observed Behavior

All the Accounts are displayed, unfiltered, due to the url parameter applied is `account_id` instead of `provider_id`

jsenecal:
### Fixes: #16963

Use `provider_id` instead of `account_id` to properly filter Accounts belonging to a specific Provider when navigating Accounts from the Providers list view

### DIFF
--- a/netbox/circuits/tables/providers.py
+++ b/netbox/circuits/tables/providers.py
@@ -25,7 +25,7 @@ class ProviderTable(ContactsColumnMixin, NetBoxTable):
     account_count = columns.LinkedCountColumn(
         accessor=tables.A('accounts__count'),
         viewname='circuits:provideraccount_list',
-        url_params={'account_id': 'pk'},
+        url_params={'provider_id': 'pk'},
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @jsenecal!

PR #17015, opened 2024-07-29
16946 return empty queryset  if filterset is not valid

### LINKED ISSUE #16946: GraphQL API returns all object if given filter is invalid
### Deployment Type

Self-hosted

### NetBox Version

v4.0.7

### Python Version

3.10

### Steps to Reproduce

My specific use case is vlan_list filtered by site id, but the behavior can be observed across all objects (eg site_list filtered by region_id)
Add VLANs to netbox and perform a GraphQL query similar to below (where 1234567890 is not a valid site)
```
{
  vlan_list(filters: {site_id: "1234567890"}){
    id
    vid
    site {
      id
    }
    tenant {
      id
    }
  }
}
```

### Expected Behavior

Return an empty list or error, as the filter is not satisfied

### Observed Behavior

Returns all objects

arthanson:


jeremystretch (changes requested):
Could you please add a test for this fix to `GraphQLTestCase`?

netbox/netbox/graphql/filter_mixins.py
```diff
@@ -201,4 +201,11 @@ def wrapper(cls):
 class BaseFilterMixin:
 
     def filter_by_filterset(self, queryset, key):
-        return self.filterset(data={key: getattr(self, key)}, queryset=queryset).qs
+        filterset = self.filterset(data={key: getattr(self, key)}, queryset=queryset)
+        if not filterset.is_valid():
+            # filterset.errors is errorDict - return first error as exception
+            k, v = next(iter(filterset.errors.items()))
```
jeremystretch, line 207: Can these lines be omitted?

netbox/netbox/tests/test_graphql.py
```diff
@@ -34,3 +40,40 @@ def test_graphiql_interface(self):
...
         with disable_warnings('django.request'):
             self.assertHttpStatus(response, 302)  # Redirect to login page
+
+
+class GraphQLAPITestCase(APITestCase):
+
+    @override_settings(LOGIN_REQUIRED=True)
+    @override_settings(EXEMPT_VIEW_PERMISSIONS=['*', 'auth.user'])
+    def test_graphql_filter_objects(self):
+        from dcim.models import Site, Location
```
jeremystretch, line 50: This could use a docstring explaining why the test has been included.

jeremystretch (changes requested):

Issue #17034, opened 2024-07-31
Add API Error Handling to Plugin Catalog


arthanson:
### Proposed Changes

Add error handling around the Plugin Catalog API

### Justification

The plugin catalog API is an external API that can be down and also NetBox might not be connected to the Internet, so proper error handling and display is required.

PR #17035, opened 2024-07-31
17034 Add error handling to Plugins Catalog API

arthanson:
### Fixes: #17034 

Put error handling around plugins catalog API, added a toast if the API could not be loaded with a 15 minute timeout so you don't keep getting spammed by toasts.  Not sure if this is acceptable for an offline installation as when you go back to the page (after 15 minutes) you will get another toast.

### DIFF
--- a/netbox/core/views.py
+++ b/netbox/core/views.py
@@ -651,18 +651,19 @@ class SystemView(UserPassesTestMixin, View):
 #
 
 class BasePluginView(UserPassesTestMixin, View):
+    CACHE_KEY_CATALOG_ERROR = 'plugins-catalog-error'
 
     def test_func(self):
         return self.request.user.is_staff
 
     def get_cached_plugins(self, request):
         catalog_plugins = {}
-        catalog_plugins_error = cache.get('plugins-catalog-error', default=False)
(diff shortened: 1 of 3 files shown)


netbox/core/views.py
```diff
@@ -658,7 +658,7 @@ def test_func(self):
     def get(self, request):
         q = request.GET.get('q', None)
 
-        plugins = get_plugins().values()
+        plugins = get_plugins(request).values()
```
jeremystretch, line 661: We probably want to avoid passing the request through to a utility function solely for the purpose of collecting a potential error message. (We should preserve the option to fetch catalog plugins absent a WSGI request.)

Instead, I'd suggest refactoring the functions:

- Move the caching logic from `get_plugins()` into `get_catalog_plugins()`
- `get_catalog_plugins()` should return an empty dict if fetching fails
- Extend `get_local_plugins()` to accept an optional dictionary of plugins, to which local plugins are added/updated
- Check in the view for a failure to retrieve catalog plugins

Something like this:

```python
plugins = get_catalog_plugins()
if not plugins:
    messages.warning(request, _("Plugins catalog could not be loaded"))
plugins = get_local_plugins(plugins)
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #17067, opened 2024-08-01
Closes #17051: Introduce the `ISOLATED_DEPLOYMENT` config parameter

### LINKED ISSUE #17051: Introduce a configuration parameter for signifying an offline or air-gapped deployment
### NetBox version

v4.0.8

### Feature type

New functionality

### Proposed functionality

Introduce a new boolean configuration parameter, e.g. `OFFLINE_DEPLOYMENT`, which is false by default. This should be set to true is NetBox is installed on a system without Internet connectivity, which may be the case in certain high-security environments.

### Use case

When this parameter is true, NetBox will omit any built-in external links, such as the footer links to GitHub and Slack. There are likely other instances of external references that may be addressed using this functionality.

### Database changes

None

### External dependencies

None

jeremystretch:
### Fixes: #17051

- Introduce the `ISOLATED_DEPLOYMENT` config parameter
- Disable the following functionality when True:
  - Automatic release checks (housekeeping task)
  - Census reporting
  - Plugins catalog integration
  - External footer links

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -87,6 +87,17 @@ addresses (and [`DEBUG`](#debug) is true).
 
 ---
 
+## ISOLATED_DEPLOYMENT
+
+Default: False
+
+Set this configuration parameter to True for NetBox deployments which do not have Internet access. This will disable miscellaneous functionality which depends on access to the Internet.
+
+!!! note
+    If Internet access is available via a proxy, set [`HTTP_PROXIES`](#http_proxies) instead.
+
+---
+
 ## JINJA2_FILTERS
 
 Default: `{}`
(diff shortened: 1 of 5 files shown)


arthanson (approved):
ISOLATED_NETWORK ?

m2martin:
I think it is worth to differentiate between the origin of requests. Some requests may originate from Netbox itself (e.g. release checks) which may fail due to isolation, others are originated from clients like the footer links which can be functional as the client is able to reach external ressources.

jeremystretch:
@m2martin yeah that's a good point. I'm going to revert the footer link changes as those are client-facing.

trrunde:
what about the rss feed on the default frontpage? I just installed netbox on a labserver where I use proxy to reach internet, the domainname where not allowed, so rss feed timed out making the front page slow to load. If its an airgapped installation the rss feed widget should be removed.

PR #17068, opened 2024-08-01
Closes #14656: Dynamic form fields for CustomField

### LINKED ISSUE #14656: Dynamically render the custom field form based on the selected type
### NetBox version

v3.7.0

### Feature type

New functionality

### Proposed functionality

Employ HTMX to re-render the custom field form, displaying only fields relevant to the selected field type. For instance, the "object type" field should be displayed only for object selection types, and the "choices" field should be displayed only for select and multi-select fields.

### Use case

This will simplify the form, omitting irrelevant fields.

### Database changes

None

### External dependencies

None

jeremystretch:
### Fixes: #14656

Dynamically include or omit form fields depending on the selected CustomField type

### DIFF
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -45,31 +45,34 @@ __all__ = (
 class CustomFieldForm(forms.ModelForm):
     object_types = ContentTypeMultipleChoiceField(
         label=_('Object types'),
-        queryset=ObjectType.objects.with_feature('custom_fields')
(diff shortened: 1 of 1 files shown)


arthanson (approved):
One annoyance noticed more on this one (common to the HTMX switching not this PR) is that is changes focus to the end of the document, so you switch the type and you jump to the bottom of the document and have to scroll back up to get back to the fields below the selection

jeremystretch:
@arthanson I haven't noticed that, but it's definitely worth capturing as a bug if it's happening consistently across multiple forms.

PR #17078, opened 2024-08-05
16670 fix OpenAPI schema generation with nested serializers

### LINKED ISSUE #16670: Full definitons of swagger responses are possibly being replaced with brief definitions due to nested serializers
### Deployment Type

Self-hosted

### NetBox Version

v4.0.3

### Python Version

3.11

### Steps to Reproduce

Check swagger JSON, all model representations of responses are only using `brief` definitions.

### Expected Behavior

Swagger responses should `always` match real responses.

### Observed Behavior

Brief representations are being returned, making it effectivelly impossible to use the swagger json.

arthanson:
### Fixes: #16670 

Adds 'Brief' to serializer name if it is called with nested=True, Adding 'Nested' conflicts with existing Nested Serializers, so went with Brief.


jeremystretch (changes requested):
This appears to incorrectly modify certain filters for related objects. For instance, the type for the `region` filter for circuits (which matches on a slug) is now reported as an integer rather than a string.

```diff
2510c2512,2513
<             type: string
---
>             type: integer
>         description: Region (slug)
2518c2521,2522
<             type: string
---
>             type: integer
>         description: Region (slug)
```

arthanson:
This looks like develop having a fix for the id -> string that wasn't merged into this PR.  I merged in the latest develop and  diff'd the .yaml files and they look correct from what I see.  Also note when switching/checking you may need to remove the cache line here: https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/urls.py#L60

jeremystretch (approved):

PR #17093, opened 2024-08-07
16176 Add UI for multi-termination cables

### LINKED ISSUE #16176: Multi-termination no longer working in web UI since 4.0
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.11

### Steps to Reproduce

Click on a front port or interface
Create a cable and connect to other interface/frontport
Try to add second interface (you should be able to have multiple terminations)

### Expected Behavior

That you should be able to have multiple termiantion points

### Observed Behavior

Each time you type in the second device interface it just replaces the orignhal temination point. You can only have 1.
Old "cables" made from 3.7.x still show the multi terminations but it you edit them and add or edit a new termination point, all others are wiped out and you have jsut 1 termination point

arthanson:
### Fixes: #16176 

This will still reset the final dropdown when the device is changed, but opening up a FR: #17105 to address the fields getting reset as that is in Javascript and will take extra work to correctly determine when to reset the field.

Editing cable #1 | NetBox 2024-08-07 14-16-09

### DIFF
--- a/netbox/dcim/forms/connections.py
+++ b/netbox/dcim/forms/connections.py
@@ -19,7 +19,7 @@ def get_cable_form(a_type, b_type):
                 # Device component
                 if hasattr(term_cls, 'device'):
 
-                    attrs[f'termination_{cable_end}_device'] = DynamicModelChoiceField(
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @arthanson!

Issue #17086, opened 2024-08-06
Exception for job details if object not set


alehaa:
### Deployment Type

Self-hosted

### NetBox Version

v4.1-beta1

### Python Version

3.11

### Steps to Reproduce

1. Create a new `JobRunner`.
2. Schedule the job runner without an instance being passed to `enqueue()`.
3. Navigate to jobs view
4. Open List item for newly created job

### Expected Behavior

Job details are shown.

### Observed Behavior

An exception is raised due to accessing `_meta` of the Job's `object`. However `object` is `None` as scheduled above.

This is due to 3028f26 introducing breadcrumbs for jobs. However, d6432fbcb886a8d900b2b24df85512371fa78d80 introduced the job object being optional around the same time.

arthanson:
@alehaa do you want to take this one?

alehaa:
@arthanson yep! :)

PR #17101, opened 2024-08-07
17086 Fix exception on job details view

alehaa:
### Fixes: #17086

If a job is not bound to an object, the job name will be used for breadcrumbs on the job details view instead. Clicking the breadcrumb will lead to the list of jobs with the same name (i.e. the same `JobRunner`).


jeremystretch (approved):
Thanks @alehaa!

PR #17113, opened 2024-08-09
17096 Do not omit "False" values from the request URL for cloned fields

### LINKED ISSUE #17096: Cloning boolean custom fields does not work if their value is `False`
### Deployment Type

Self-hosted

### NetBox Version

v4.0.8

### Python Version

3.11

### Steps to Reproduce

1. Create a custom field 'test' with type 'Boolean', 'Is clonable' checked for an arbitrary object type, say `ipam.IPAddress`
2. Set the custom field to `False` for an arbitrary `IPAddress` object and click on 'Save'
3. Click on 'Clone'

### Expected Behavior

The cloned object has its 'test' CF set to `False`

### Observed Behavior

The cloned object has its 'test' CF set to `None`

If you look at the parameters of the URL for the 'Clone' button, the relevant part is `test=`. This is identical whether the CF 'test' was `False` for the original object or if it was undefined.

This is particularly relevant for required fields, as it makes it necessary to set the field explicitly for cloned objects even if cloning the field was enabled.

peteeckel:
### Fixes: #17096

The current implementation leaves the URL parameters for boolean fields empty if their value is actually `False`, which leads to fields with `False` values not being cloned correctly.

### DIFF
--- a/netbox/netbox/forms/base.py
+++ b/netbox/netbox/forms/base.py
@@ -1,6 +1,7 @@
 import json
 
 from django import forms
+from django.forms.fields import BooleanField, NullBooleanField
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import Q
 from django.utils.translation import gettext_lazy as _
@@ -30,6 +31,15 @@ class NetBoxModelForm(CheckLastUpdatedMixin, CustomFieldsMixin, TagsMixin, forms
             the rendered form (optional). If not defined, the all fields will be rendered as a single section.
     """
     fieldsets = ()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        for key, value in self.initial.items():
+            if key not in self.fields:
+                continue
+            if isinstance(self.fields[key], (BooleanField, NullBooleanField)) and self.initial[key] == "False":
+                self.initial[key] = False
 
     def _get_content_type(self):
         return ContentType.objects.get_for_model(self._meta.model)
--- a/netbox/utilities/querydict.py
+++ b/netbox/utilities/querydict.py
@@ -55,7 +55,7 @@ def prepare_cloned_fields(instance):
     for key, value in attrs.items():
         if type(value) in (list, tuple):
             params.extend([(key, v) for v in value])
-        elif value is not None and (key.startswith('cf_') or value is not False):
+        elif value is not None:
             params.append((key, value))
         else:
             params.append((key, ''))
--- a/netbox/utilities/querydict.py
+++ b/netbox/utilities/querydict.py
@@ -55,7 +55,7 @@ def prepare_cloned_fields(instance):
     for key, value in attrs.items():
         if type(value) in (list, tuple):
             params.extend([(key, v) for v in value])
-        elif value is not False and value is not None:
+        elif value is not None:
             params.append((key, value))
         else:
             params.append((key, ''))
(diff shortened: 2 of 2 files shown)


jeremystretch (changes requested):
Have you tested this change using other models? A VRF, for instance, will not have its `enforce_unique` field cloned properly because the string "False" is interpreted as true.

peteeckel:
I ran the NetBox test suite in the assumption that would secure the fix doesn't break anything. Obviously wrong ...

Given the fact that there is either the CF problem or the problem with cloning the `enforce_unique` field, the question is where exactly the problem is. As long as the string is constructed the was it currently is, CF cloning doesn't work properly, but presumably any non-null version is interpreted by the code that's responsible for cloning `enforce_unique` as `True`, which will break that code. That is definitely an inconsistency in the code base.

I can have a look at why the VRF code interprets "False" as `True`, but that might take some time.

peteeckel:
@jeremystretch I had a look at the problem and there is a solution, but probably not a perfect one.

The interpretation of the value from the URL parameters is a Django utility function, and is applied to all boolean fields on initialisation. This results in the problem you mentioned. Custom boolean fields are not affected by that, because they aren't proper model fields after all, so it's safe to assume that all URL parameters starting with `cf_` can (and need to) be passed the 'False' string if the value to be cloned is `False`. The new version of the fix reflects that.

However, I'm not sure what what will happen if someone choses to use fields of type `NullBooleanField` instead of `BooleanField`. NetBox currently doesn't, neither do I in my plugin, and so I don't have a proper testing environment. I think this could be handled as well by looking at the field type via `_meta` and doing special treatment if a `NullBooleanField` is detected (which is not the case for CFs, so the check for `cf_` in the field name is still necessary), but for issue #17096 this patch is a sufficient solution.

netbox/utilities/querydict.py
```diff
@@ -55,7 +55,7 @@ def prepare_cloned_fields(instance):
     for key, value in attrs.items():
         if type(value) in (list, tuple):
             params.extend([(key, v) for v in value])
-        elif value is not False and value is not None:
+        elif value is not None and (key.startswith('cf_') or value is not False):
```
jeremystretch, line 58: I'm afraid this isn't a tenable change: We cannot make arbitrary exceptions for certain types of fields. Instead, the form initialization logic may need to be updated to acknowledge explicit False (as opposed to undefined) values.

jeremystretch (changes requested):

netbox/utilities/querydict.py
peteeckel: The problem with updating the form initialisation logic is that we can't do that indiscriminately either. There may well be a `CharField` that receives the string `"False"`, in which case it isn't a very useful idea to modify it to the boolean value `False`.

So even when we modify the initialisation logic we still need to discriminate between field types that receive the data. Which is what I did now, this time on field type level.

peteeckel:
@jeremystretch Did you notice that I made some changes to this PR - it's now pending closure and the bug is really annoying ...

jeremystretch (changes requested):

netbox/netbox/forms/base.py
```diff
@@ -31,6 +32,15 @@ class NetBoxModelForm(CheckLastUpdatedMixin, CustomFieldsMixin, TagsMixin, forms
     """
     fieldsets = ()
 
+    def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
peteeckel: > it's probably reasonably safe to assume the string "False" implies the boolean value False.

While this may be true in most cases I would not implement code that relies on this assumption. It's just that, an assumption, after all, and errors resulting from applications actually relying on "False" being a string will be pretty hard to debug. That kind of reasoning is exactly where the referenced issue came from, and the solution of just assuming "False" will always refer to a boolean value will only hide the issue better.

The question is: Are there any forms not derived from `NetBoxModelForm` that are used in cloning objects?

PR #17122, opened 2024-08-10
16995 Move language handling to Middleware

### LINKED ISSUE #16995: Language setting not applied when signing in via SSO
### Deployment Type

Self-hosted

### NetBox Version

v4.0.7

### Python Version

3.11

### Steps to Reproduce

1. Set your user language to something that differs from the automatically detected language
2. Clear cookies
3. Log into Netbox via **Google** SSO

Bonus: if you log in via password everything works correctly.

### Expected Behavior

4\. Language from user settings is applied

### Observed Behavior

4\. The automatically detected language is used

You can see that the language cookie is not being created:
grafik

m2martin:
### Fixes: #16995

Decouples language activation from login form to have it handled by `CoreMiddleware` which will also trigger language activation on any type of login (e.g. SSO).

### DIFF
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -8,6 +8,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.db import connection, ProgrammingError
 from django.db.utils import InternalError
 from django.http import Http404, HttpResponseRedirect
+from django.utils import translation
 
 from netbox.config import clear_config, get_config
 from netbox.context_managers import event_tracking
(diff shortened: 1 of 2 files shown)


netbox/netbox/middleware.py
```diff
@@ -43,12 +44,30 @@ def __call__(self, request):
...
             return HttpResponseRedirect(login_url)
 
+        # Check if the language should be activated before the language cookie has been set.
+        # This will happen for instance on the first request after login.
+        # The language will have been already activated by the framework if the cookie exists.
+        if (
+            request.user.is_authenticated and
+            settings.LANGUAGE_COOKIE_NAME not in request.COOKIES and
+            hasattr(request.user, 'config')
+        ):
```
jeremystretch, line 54: Are we sure this is needed? IIRC the first authenticated response should always be a redirect, so I'd expect the language cookie should already be set.

m2martin: Your assumption is true when the login form or even some types of SSO is used, but not for `netbox.authentication.RemoteUserBackend`.

If you're using header-based authentication without that code fragment, you'll end up with the first page being displayed in the browser's default language, despite having a correct cookie.

jeremystretch: Ah, right.

bctiemann:
@m2martin Could you please rebase this PR? I tried to resolve what seemed like a trivial merge commit and found that there are a large number of tests failing due to references to `settings.AUTH_EXEMPT_PATHS` which does not exist, as well as missing imports (i.e. `parse`).

jeremystretch:
I've rebased this against `develop`, but it's still not clear to me what the recommended convention is for setting and renewing the language cookie.

m2martin:
What exactly are you missing? Do you mean the "Django recommended way"?

m2martin:
Hmm, I understand your concerns but I think I can't provide any satisfying answer...

AFAIK, Django has had language and sessions coupled in the past but now tries to de-couple them (remove language from sessions and fully rely on the cookie as detection mechanism).

I've found no other way to implement all possible login cases.
There are two code points where we can kick in - before and after `get_response()`. Because of this, there are two code blocks. On the latter, we can't change the language anymore because it has already been set and activated by other instances.
On the first, we can follow the Django-recommended way of using `translation.activate()`.

It could be simplified if it is acceptable to have the first page-load being displayed in the wrong language.

PR #17129, opened 2024-08-12
16756 fix script result pagination issue

### LINKED ISSUE #16756: script page elements are duplicated when log output is paginated and browsed
### Deployment Type

Self-hosted

### NetBox Version

4.0.5

### Python Version

3.11

### Steps to Reproduce

Create a custom script that logs at least 500 messages, such as the attached:

```
from extras.scripts import *
class CreateMessagesScript(Script):
    class Meta:
        name = "Show Message Pagination"
        description = "Generate over 500 messages"
    def run(self, data, commit):
        for i in range(1,555):
            self.log_success(i)
        output=[]
        return '\n'.join(output)
```

Execute the script

Page through the output.  For each page that is viewed, rendered page elements around the log table are duplicated.

### Expected Behavior

Paging through logs should not impact static page elements surrounding the output table

### Observed Behavior

The "output" box is duplicated

image

and so are elements above the log table, such as the script status, runtime, etc
image

arthanson:

### DIFF
--- a/netbox/templates/extras/htmx/script_result.html
+++ b/netbox/templates/extras/htmx/script_result.html
@@ -1,6 +1,7 @@
 {% load helpers %}
 {% load log_levels %}
 {% load i18n %}
+{% load render_table from django_tables2 %}
 
 <div class="htmx-container">
   <p>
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):
We can't just copy code out of the included template and tweak it, as that results in duplicated code that assuredly will not be maintained over time. You'll need to dig deeper and find a solution that allows efficient reuse of the template.

netbox/templates/extras/htmx/script_result.html
```diff
@@ -41,7 +41,7 @@ <h5 class="card-header">{% trans "Test Summary" %}</h5>
     <div class="card">
       <div class="table-responsive" id="object_list">
         <h5 class="card-header">{% trans "Log" %}</h5>
-        {% include 'htmx/table.html' %}
+        {% include 'htmx/table.html' with no_htmx_container=True%}
```
jeremystretch, line 44: I don't think this is a pattern we want to encourage. I'm going to see if I can come up with a cleaner fix for this.

jeremystretch:
Closing this in favor of #17290

PR #17130, opened 2024-08-12
16733 fix bulk edit/delete buttons with quick search

### LINKED ISSUE #16733: editing selection of devices via quick search
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5

### Python Version

3.12

### Steps to Reproduce

1. Click on devices
2. type your searchwords in quick search
3. click Select all ... devices matching query
image

4. click edit selected devices
5. At the left upper corner you will see another amount of devices which are going to be edited. In the selected devices tab you will see other devices selected than you searched for.
image

### Expected Behavior

I would expect that only the devices I searched for should be edited.

### Observed Behavior

All devices were selected.

arthanson:

### DIFF
--- a/netbox/templates/generic/object_list.html
+++ b/netbox/templates/generic/object_list.html
@@ -84,10 +84,14 @@ Context:
               <div class="card-body">
                 <div class="float-end">
                   {% if 'bulk_edit' in actions %}
+                    <div class="bulk-edit-button">
                     {% bulk_edit_button model query_params=request.GET %}
+                    </div>
                   {% endif %}
                   {% if 'bulk_delete' in actions %}
+                    <div class="bulk-delete-button">
                     {% bulk_delete_button model query_params=request.GET %}
+                    </div>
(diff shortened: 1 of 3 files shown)


netbox/templates/generic/object_list.html
```diff
@@ -84,10 +84,14 @@
               <div class="card-body">
                 <div class="float-end">
                   {% if 'bulk_edit' in actions %}
+                    <span class="bulk-edit-button">
                     {% bulk_edit_button model query_params=request.GET %}
+                    </span>
```
jeremystretch, line 89: We should use classes on the buttons directly, rather than wrapping each in a `<span>`.

jeremystretch (changes requested):
This fixes the bulk edit & delete buttons within the top panel, but the buttons at the bottom of the list are still broken. If we tweak this to render the bulk edit & delete buttons with classes directly it should fix both the top and bottom buttons.

PR #17142, opened 2024-08-13
add 'vlan' to prefix bulk edit

### LINKED ISSUE #16692: Prefixes: add 'vlan' to bulk edit
### NetBox version

v4.0.5

### Feature type

### Proposed functionality

When editing a Prefix, you can set the VLAN/Group
When using the Bulk-Edit option, there is no 'VLAN' field to change

My proposal is to update the Bulk Edit form, so that VLANs can be assigned in bulk to prefixes

### Use case

Update the VLAN assignment of multiple prefixes in one go

### Database changes

_No response_

### External dependencies

_No response_

PieterL75:


jeremystretch (approved):
Thanks @PieterL75. I've just split out the new fields into a separate field set for VLAN assignment specifically.

PR #17154, opened 2024-08-14
16959 Fix filter reset button

### LINKED ISSUE #16959: Filter reset button does not reset filter
### Deployment Type

Self-hosted

### NetBox Version

v4.0.7

### Python Version

3.10

### Steps to Reproduce

1. Click on Devices (I think any model that can be filtered will do)
2. Click on filter
3. Apply any filter (I used Status PLANNED for device)
4. Press Reset
5. Observe filter is still applied

### Expected Behavior

The filter should have been removed / reset

### Observed Behavior

The filter did not get removed / reset

arthanson:


jeremystretch (changes requested):

jeremystretch (changes requested):
Curious why this was added to `dynamicTomSelect.ts`. Could we move it to `/src/forms/` as it's specific to the tom-select widgets?

PR #17161, opened 2024-08-14
Fixes #17097: Record static object representation when calling `NotificationGroup.notify()`

### LINKED ISSUE #17097: `NotificationGroup.notify()` does not capture static representation of associated object
### Deployment Type

NetBox Cloud

### NetBox Version

v4.1-beta1

### Python Version

3.10

### Steps to Reproduce

1. Create a notification group
2. Create an event rule (e.g. for object changes) targeting that notification group
3. Trigger the event rule

### Expected Behavior

The generated notification should include a static representation of the object (`object_repr`).

### Observed Behavior

The notification's `object_repr` attribute is empty.

jeremystretch:
### Fixes: #17097

Pass the `display` string from the serialized object when bulk-creating notification for a group. This isn't ideal, but the event handling logic doesn't currently have access to the object instance.

### DIFF
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -139,6 +139,7 @@ def process_event_rules(event_rules, object_type, event_type, data, username=Non
             event_rule.action_object.notify(
                 object_type=object_type,
                 object_id=data['id'],
+                object_repr=data.get('display'),
                 event_type=event_type
             )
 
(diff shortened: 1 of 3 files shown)


arthanson (approved):

arthanson (changes requested):
One issue, if you delete an object (I did circuits) the notification dropdown shows a link to it that when clicked will result in an error.  On the notifications page it is not clickable, just need to make this one non-clickable as well.

Circuits | NetBox 2024-08-15 17-41-21

arthanson (approved):

PR #17192, opened 2024-08-16
17174 add version to plugin catalog

arthanson:

### DIFF
--- a/netbox/core/tables/plugins.py
+++ b/netbox/core/tables/plugins.py
@@ -56,6 +56,12 @@ class CatalogPluginTable(BaseTable):
     is_certified = columns.BooleanColumn(
         verbose_name=_('Certified')
     )
+    created_at = columns.DateTimeColumn(
+        verbose_name=_('Published')
+    )
+    updated_at = columns.DateTimeColumn(
+        verbose_name=_('Updated')
+    )
     installed_version = tables.Column(
         verbose_name=_('Installed Version')
     )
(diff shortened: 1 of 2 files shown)


netbox/core/tables/plugins.py
```diff
@@ -56,11 +56,12 @@ class CatalogPluginTable(BaseTable):
...
-        verbose_name=_('Published')
+    installed_version = tables.Column(
+        verbose_name=_('Installed Version')
     )
-    updated_at = columns.DateTimeColumn(
-        verbose_name=_('Updated')
+    latest_version = tables.Column(
+        accessor=tables.A('release_latest__version'),
+        verbose_name=_('Latest Version')
     )
```
jeremystretch, line 65: We should keep these columns as optional, as they still convey data from the catalog API.

netbox/core/tables/plugins.py
```diff
@@ -56,11 +56,12 @@ class CatalogPluginTable(BaseTable):
     is_certified = columns.BooleanColumn(
         verbose_name=_('Certified')
     )
-    created_at = columns.DateTimeColumn(
-        verbose_name=_('Published')
+    installed_version = tables.Column(
+        verbose_name=_('Installed Version')
```
jeremystretch, line 60: There is already an `installed_version` column below.

jeremystretch (changes requested):

jeremystretch (approved):

PR #17201, opened 2024-08-18
Fixes: #14606 - For Cable API add PK for cable terminations to serializer

### LINKED ISSUE #14606: Terminations listed for a cable in the REST API should include their numeric IDs
### Deployment Type

Self-hosted

### NetBox Version

v3.6.7

### Python Version

3.10

### Steps to Reproduce

1. Create a cable with one or more terminations on its A and B sides.
2. View the cable's representation in the REST API (e.g. `/api/dcim/cables/123/`).

### Expected Behavior

Each termination object represented under the cable's `a_terminations` and `b_terminations` attributes should include its own numeric primary key (ID).

### Observed Behavior

This attribute is missing from the terminations. This is because the `id` field is not included on GenericObjectSerializer.

DanSheps:
### Fixes: #14606

* Add pk field for cable termination
* Add test


jeremystretch:
This one is probably a bit stickier than I realized in the original bug report.

We should really be using a nested CableTerminationSerializer here, omitting its `cable` and `cable_end` fields (as they're redundant in this context). That would address the primary issue.

However, the catch is that the generic serializer currently in use returns the fields `object_type`, `object_id` and `object`, whereas CableTerminationSerializer returns `termination_type`, `termination_id`, and `termination`. Swapping out the serializer would be a breaking change.

We _could_ address this in v4.1, however I'm hesitant to slip in a breaking change so late in the beta period. For now, we could settle with just adding the `id` field to GenericObjectSerializer, and capture this as a potential change for v4.2.

netbox/dcim/api/serializers_/cables.py
```diff
@@ -16,14 +16,27 @@
     'CableSerializer',
     'CableTerminationSerializer',
     'CabledObjectSerializer',
+    'GenericObjectTerminationSerializer',
     'TracedCableSerializer',
 )
 
 
+class GenericObjectTerminationSerializer(GenericObjectSerializer):
```
jeremystretch, line 24: This doesn't require a new serializer: We can just add the `id` field to GenericObjectSerializer.

netbox/dcim/tests/test_api.py
```diff
@@ -2038,6 +2038,14 @@ def setUpTestData(cls):
             },
         ]
 
+    def test_cable_termination_pk_exist(self):
```
jeremystretch, line 2041: This test probably isn't necessary.

jeremystretch (changes requested):

netbox/dcim/api/serializers_/cables.py
DanSheps: Unfortunately since a_terminations/b_terminations is a property function that includes the actual termination (Interface/CircuitTermination/etc) and not the CableTermination it won't be the proper `id` (it will be the interface id for example, not the cable termination id).

This is why I opted to go this route, as as you mention, it would be a breaking change.  Perhaps we should aim for 4.1 instead and I just rebase and swap the serializer.

It may also make more sense to maintain this a_terminations/b_terminations and add a generic "terminations" which operates on `Cable.terminations` which is I think where it would be appropriate to use the CableTerminations serializer.

DanSheps:
* Rebased to feature
* Modified serializer to add `terminations` related objects

Let me know if you would rather replace `a_terminations` / `b_terminations` (will be non-trivial as they currently only return the Terminated Objects (Interface, etc)

netbox/dcim/api/serializers_/cables.py
```diff
@@ -20,24 +20,8 @@
...
-    class Meta:
-        model = Cable
-        fields = [
-            'id', 'url', 'display_url', 'display', 'type', 'a_terminations', 'b_terminations', 'status', 'tenant',
-            'label', 'color', 'length', 'length_unit', 'description', 'comments', 'tags', 'custom_fields', 'created',
-            'last_updated',
-        ]
-        brief_fields = ('id', 'url', 'display', 'label', 'description')
-
+class TracedCableSerializer(serializers.ModelSerializer):
```
DanSheps, line 23:
```suggestion
class TracedCableSerializer(BaseModelSerializer):
```

netbox/dcim/api/serializers_/cables.py
```diff
@@ -20,24 +20,8 @@
...
-        fields = [
-            'id', 'url', 'display_url', 'display', 'type', 'a_terminations', 'b_terminations', 'status', 'tenant',
-            'label', 'color', 'length', 'length_unit', 'description', 'comments', 'tags', 'custom_fields', 'created',
-            'last_updated',
-        ]
-        brief_fields = ('id', 'url', 'display', 'label', 'description')
-
-
 class TracedCableSerializer(BaseModelSerializer):
+
```
DanSheps, line 24:
```suggestion
```

jeremystretch:
This is going to be more involved than I originally anticipated, because the `a_terminations` and `b_terminations` attributes of a cable instance are populated with the actual terminating objects, rather than CableTermination instances. We don't want to just tack these on in addition to the existing attributes because that would be redundant and inefficient. We'll have to explore our options, which likely include tweaking the underlying querysets for the Cable model.

jeremystretch:
Appreciate your work on this @DanSheps. I'm going to close it out for now; let's re-evaluate our options for a potential change in v4.2.

PR #17211, opened 2024-08-20
Fixes: #15924 - Prevent API payload from allowing tagged_vlans while interface mode is set to tagged-all

### LINKED ISSUE #15924: Updating Interface.tagged_vlans via API improperly allowed on interface with mode: tagged-all
### Deployment Type

Self-hosted

### NetBox Version

v3.7.5, 4.0.3

### Python Version

3.9

### Steps to Reproduce

- Choose any interface.
- Set the interface's _801.1Q Mode_ to _Tagged (all)_ (in API terms, `"mode": "tagged-all"`)
- Use a `PATCH` request to set a non-empty value for the interface's `tagged_vlans` property
- Use a second `PATCH` request to change only the interface's `description` property, leaving `tagged_vlans` and `untagged_vlans` out of the patch
- Retrieve the interface via the API.

```
$ curl -X PATCH -H "Content-type: application/json"  -H "Accept: application/json" -H "Authorization: Token $TOKEN"  http://localhost/api/dcim/interfaces/1/  --data '{"description": "Patching the description and tagged_vlans", "tagged_vlans": [ 1 ] }'

$ curl -X PATCH -H "Content-type: application/json"  -H "Accept: application/json" -H "Authorization: Token $TOKEN"  http://localhost/api/dcim/interfaces/1/  --data '{"description": "Patching the description only" }'
```

### Expected Behavior

The first `PATCH` should be rejected with an API error message indicating that setting `tagged_vlan` is not allowed on an interface with `"mode": "tagged_all"`.

### Observed Behavior

The first (invalid) `PATCH` is accepted, and reflected as specified. The second (valid) `PATCH` is accepted, and results in the unexpected obliteration of the `tagged_vlan` result.

### LINKED ISSUE #17249: Adding the untagged_vlan property to dcim.interface using the API that isn't in mode tagged fails without error
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5, v4.0.9

### Python Version

3.10

### Steps to Reproduce

1. Create site
2. Create vlan at site, don't set 802.1Q Mode (leave it unset)
3. Create device at site
4. Create interface on device
5. Try adding the vlan to the interface using the netbox api with the following options
   1. dict containing dict `{'untagged_vlan':{vid, name, site}}` values from vlan
   2. dict containing dict `{'untagged_vlan':{id}}`, value from vlan
   3. dict contaning value `{'untagged_vlan': id}` value from vlan

## Test script using pynetbox 
`pip install logruru pynetbox`
loguru because I like the colours 
```
import pynetbox
import requests

from loguru import logger

netbox_url = "https://demo.netbox.dev"
token = "I assume my token will be deleted so add your own :)"
headers = {
    'Authorization': f'Token {token}',
    'Content-Type': 'application/json',
    'Accept': 'application/json',
}

device_id = 129 
iface_id = 1776
vlan_id = 21

nb = pynetbox.api(netbox_url, token=token)

device = nb.dcim.devices.get(id=device_id)
logger.debug(dict(device))

iface = nb.dcim.interfaces.get(id=iface_id)
logger.debug(dict(iface))

vlan = nb.ipam.vlans.get(id=vlan_id)
logger.debug(dict(vlan))



## Pynetbox tests
try:
    logger.error("pynetbox: API doc required vid and name")
    logger.info(f"loaded: {iface.untagged_vlan}")
    iface.untagged_vlan = {"vid": vlan.vid, "name": vlan.name, "site": vlan.site.id}
    logger.info(f"local: {iface.untagged_vlan}")
    iface.save()
    iface = nb.dcim.interfaces.get(id=iface_id)
    logger.info(f"reloaded: {iface.untagged_vlan}")
except Exception as e:
    logger.critical(e)

try:
    logger.error("pynetbox: vlan id")
    iface.untagged_vlan = {"id": vlan.id}
    logger.info(f"local: {iface.untagged_vlan}")
    iface.save()
    iface = nb.dcim.interfaces.get(id=iface_id)
    logger.info(f"reloaded: {iface.untagged_vlan}")
except Exception as e:
    logger.critical(e)

try:
    logger.error("pynetbox: vlan object")
    iface.untagged_vlan = vlan
    logger.info(f"local: {iface.untagged_vlan}")
    iface.save()
    iface = nb.dcim.interfaces.get(id=iface_id)
    logger.info(f"reloaded: {iface.untagged_vlan}")
except Exception as e:
    logger.critical(e)
```

### Expected Behavior

The api call would return a error message telling me I can't add a untagged_vlan with first setting a 802.1Q Mode.

### Observed Behavior

The api works successfully and the interface object is returned unchanged.

DanSheps:
### Fixes: #15924 and #17249 - Prevent API payload from allowing tagged_vlans while interface mode is set to taged-all and prevent API from allowing untagged_vlan when no 802.1q mode is set.

* Prevent API (or any API write function) which includes untagged_vlan with interface 802.1q mode is not set
* Prevent API (or any API write function) which includes tagged_vlans with interface mode set to tagged-all


netbox/dcim/models/device_components.py
```diff
@@ -914,6 +914,10 @@ def save(self, *args, **kwargs):
 
         super().save(*args, **kwargs)
 
+        # Clear any tagged vlans set when mode is tagged-all
+        if self.mode == InterfaceModeChoices.MODE_TAGGED_ALL and self.tagged_vlans.count():
+            self.tagged_vlans.set([])
```
jeremystretch, line 919: IMO we should rely on validation to prevent this, and require the user to explicitly remove all VLANs from the interface. Otherwise, this opens the possibility that all VLAN assignments can be cleared by accidentally changing the 802.1Q mode.

(Also, we can call `self.tagged_vlans.clear()` instead of `set([])`.)

jeremystretch (changes requested):

netbox/dcim/models/device_components.py
```diff
@@ -904,6 +904,12 @@ def clean(self):
                 ).format(untagged_vlan=self.untagged_vlan)
             })
 
+        # Validate that tagged-all payload does not include tagged_vlans
+        if self.mode != InterfaceModeChoices.MODE_TAGGED and self.tagged_vlans:
+            raise ValidationError({
+                'tagged_vlans': "Interface mode does not support including tagged vlans"
+            })
+
```
jeremystretch, line 912: We should move this to `clean()` on BaseInterface to ensure it applies to VM interfaces as well.

jeremystretch (changes requested):

netbox/dcim/models/device_components.py
```diff
@@ -904,6 +904,12 @@ def clean(self):
                 ).format(untagged_vlan=self.untagged_vlan)
             })
 
+        # Validate that tagged-all payload does not include tagged_vlans
+        if self.mode != InterfaceModeChoices.MODE_TAGGED and self.tagged_vlans:
```
jeremystretch, line 908: I think this needs to go on the serializer, as we can't check against many-to-many assignments under the model's `clean()` method. (We should ensure that it applies for VM interfaces too.)

jeremystretch (changes requested):

DanSheps:
Per maintainer meeting, need to refactor logic to try and streamline the checks.

DanSheps:
Blocked by #17794

netbox/dcim/forms/common.py
```diff
@@ -52,8 +56,10 @@ def clean(self):
             })
 
         # Remove all tagged VLAN assignments from "tagged all" interfaces
-        elif self.cleaned_data['mode'] == InterfaceModeChoices.MODE_TAGGED_ALL:
-            self.cleaned_data['tagged_vlans'] = []
+        elif self.cleaned_data['mode'] == InterfaceModeChoices.MODE_TAGGED_ALL and tagged_vlans:
+            raise forms.ValidationError({
+                'mode': _("An tagged-all interface cannot have tagged VLANs assigned.")
```
arthanson, line 61: "An" should be "A"

arthanson: Should this be the same logic as above for the API?  In the API case you are checking for  `mode != InterfaceModeChoices.MODE_TAGGED` and here you are testing for `self.cleaned_data['mode'] == InterfaceModeChoices.MODE_TAGGED_ALL`

netbox/dcim/models/device_components.py
```diff
@@ -934,6 +934,8 @@ def clean(self):
                 raise ValidationError({'rf_channel_width': _("Cannot specify custom width with channel selected.")})
 
         # VLAN validation
+        if not self.mode and self.untagged_vlan:
```
arthanson, line 937: I didn't see a test case for this condition?

arthanson (changes requested):

netbox/dcim/forms/common.py
DanSheps: Funny enough, after reviewing this PR I realised that the invalid data is just removed in __init__ because we delete that item.

I am not sure if this is the best way to do this IMO.

arthanson (commented):
@DanSheps I know a bit outside the scope of this issue, but we have validation here going on in 3 different spots and I think (haven't checked) if you just created an Interface model and saved it (directly in code) it would accept an invalid state for this.  Would it make sense to consolidate this in the model clean or at least a common function?  We should really have a code standard for this - validation in one central place.  Note: Asking, if you don't think it's feasible please push back.

DanSheps:
> We should really have a code standard for this - validation in one central place. Note: Asking, if you don't think it's feasible please push back.

You cannot check the many to many relationship on the clean in the model, which is the main driver of this change.  You wouldn't set the many to many field and save on the model anyways, you would need to set() or add().

Now, we could make a Mixin which handles the validation and call that from both the API and form and at least consolidate the logic there (I did briefly think about this), but the datastructure in the API and the form is slightly different.

I would have preferred to do this on the database too but alas there is no way to perform the check.

DanSheps:
FYI: Tested this, the code between the form validation and the serializer validation is unfortunately too different to really support a common check (ex: form uses `self.get_intial_for_field` to obtain the field data from the instance, whereas api just checks the instance.  Changing this to only use instance might have worked, but form as a non "None"  instance for creation which means you need to check for pk in the validating function vs api which just has a `None`.

There are other differences too, I think, unfortunately, it makes more sense to just check on the serializer/form.

jeremystretch:
After some discussion, we've decided to ship this fix in v4.2 as it doesn't quite rise to the threshold of a breaking change. I'll merge this into `main` as well.

PR #17212, opened 2024-08-20
Fixes: #14044 - Allow regex renaming of unnamed devices

### LINKED ISSUE #14044: NoneType AttributeError when Bulk Renaming Nameless Devices

DanSheps:
### Fixes: #14044 - Allow regex renaming of unnamed devices

* Allow regex renaming of unnamed devices (already allowed actually)
* Catch errors relating to unnamed devices or integrity errors as a result of the rename process

Note: I removed the ^$ requirement as, in my view, as long as there is a valid regex and the regex results in a unique set, we don't need to enforce the ^$ requirement.  For example: `^` or `.*` or `.` would all be equally as valid.

The main changes here catch both IntegrityError's from the DB and a validation error to ensure regex is used.


netbox/dcim/views.py
```diff
@@ -2122,6 +2122,17 @@ class DeviceBulkRenameView(generic.BulkRenameView):
     filterset = filtersets.DeviceFilterSet
     table = tables.DeviceTable
 
+    def _rename_objects(self, form, selected_objects):
```
jeremystretch, line 2125: Private view methods should not be overridden. Additionally, it's not clear why this logic would apply to devices but not other objects being renamed in bulk.

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -776,6 +776,14 @@ def post(self, request):
...
                             return redirect(self.get_return_url(request))
 
+                except IntegrityError as e:
+                    messages.error(self.request, ", ".join(e.args))
+                    clear_events.send(sender=self)
+
+                except ValidationError as e:
+                    messages.error(self.request, ", ".join(e.messages))
+                    clear_events.send(sender=self)
+
```
jeremystretch, line 786: Are these changes strictly relevant to #14044? If not, we should break them out into a separate bug report.

jeremystretch (changes requested):
Might it be simpler to just replace original values of None with an empty string?

netbox/netbox/views/generic/bulk_views.py
DanSheps: They are.  This will throw a validation error if we don't use regex (We can easily dump that though if we want to go to `None == ''`)

The Integrity error will catch any associated integrity errors.  I tried to fine a way to separate the integrity error out and only capture an integrity error that relates to the name field but it wasn't easily do-able.

jeremystretch:
AFAICT the only change actually needed to address #14044 is to ensure that a null name is always treated as an empty string. (Currently this is true only when regex is enabled.) Have I missed something?

DanSheps:
> AFAICT the only change actually needed to address #14044 is to ensure that a null name is always treated as an empty string. (Currently this is true only when regex is enabled.) Have I missed something?

I can get that fixed up right away then

jeremystretch (approved):

PR #17241, opened 2024-08-22
Closes #17221: Extend ObjectEditView to support HTMX requests

### LINKED ISSUE #17221: Extend ObjectEditView to support HTMX requests
### NetBox version

v4.1-beta1

### Feature type

New functionality

### Proposed functionality

Extend NetBox's generic [ObjectEditView](https://netboxlabs.com/docs/netbox/en/stable/plugins/development/views/#netbox.views.generic.object_views.ObjectEditView) to support modifying objects via HTMX `POST` requests. This view currently supports HTMX only for `GET` requests.

This will entail checking whether a `POST` request is HTMX-enabled, and if so returning the applicable rendered content directly (rather than returning a 302 redirect).

### Use case

This change is necessary to support editing objects in place within an HTML template, and will unlock future functionality.

### Database changes

None

### External dependencies

None

jeremystretch:
### Fixes: #17221

Extends `ObjectEditView` to return an `HX-Location` redirect in response to HTMX POST requests

### DIFF
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -234,6 +234,8 @@ class ObjectEditView(GetReturnURLMixin, BaseObjectView):
         # If this is an HTMX request, return only the rendered form HTML
         if htmx_partial(request):
             return render(request, self.htmx_template_name, {
+                'model': model,
+                'object': obj,
                 'form': form,
             })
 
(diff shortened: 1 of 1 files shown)


netbox/netbox/views/generic/object_views.py
```diff
@@ -234,6 +234,8 @@ def get(self, request, *args, **kwargs):
         # If this is an HTMX request, return only the rendered form HTML
         if htmx_partial(request):
             return render(request, self.htmx_template_name, {
+                'model': model,
```
arthanson, line 237: What are these used for?

arthanson (approved):

netbox/netbox/views/generic/object_views.py
jeremystretch: Just additional context for the template, should they be needed.

PR #17244, opened 2024-08-22
Fixes: #16292 - Properly restrict GraphQL queries for querys with pk set

### LINKED ISSUE #16292: Authentication Bypass in GraphQL Queries for Users/Tokens Lacking Permissions
### Deployment Type

Self-hosted

### NetBox Version

v4.0.3

### Python Version

3.11

### Steps to Reproduce

Test user does not have any permissions associated with it (I am using the admin token to make the query):
```bash
% curl -s -H "Authorization: Token 28a079fb8aa5e107583583dbeb7deb027121a15a" \
-H "Accept: application/json" \
http://localhost:8000/api/users/users/\?username\=test | jq -r '.results[] | .username, .permissions'
test
[]
```

Test user token:
```bash
% curl -s -H "Authorization: Token 28a079fb8aa5e107583583dbeb7deb027121a15a" \
-H "Accept: application/json" \
http://localhost:8000/api/users/tokens/ | jq -r '.results[] | select(.user.username == "test") | .user.username, .key'
test
4f745a0b1c13168c95883ee004f17df7ef96e42a
```

Now using the test user token without any permissions via Graphql:
```bash
% curl -s -H "Authorization: Token 4f745a0b1c13168c95883ee004f17df7ef96e42a" \
-H "Accept: application/json" \
http://localhost:8000/graphql/ \
--data '{"query": "query {asn(id:1) {asn}}"}'
{"data": {"asn": {"asn": 666}}}%
```

```bash
% curl -s -H "Authorization: Token 4f745a0b1c13168c95883ee004f17df7ef96e42a" \
-H "Accept: application/json" \
http://localhost:8000/graphql/ \
--data '{"query": "query {provider(id:1) {name}}"}'
{"data": {"provider": {"name": "test-p"}}}%
```

In my understanding, if a user/token has no permissions, it should reject by default.

Version:
```bash
% curl -s -H "Authorization: Token 28a079fb8aa5e107583583dbeb7deb027121a15a" \
-H "Accept: application/json" \
http://localhost:8000/api/status/ | jq -r '.'
{
  "django-version": "5.0.6",
  "installed-apps": {
    "debug_toolbar": "4.3.0",
    "django_filters": "24.2",
    "django_prometheus": "2.3.1",
    "django_rq": "2.10.2",
    "django_tables2": "2.7.0",
    "drf_spectacular": "0.27.2",
    "drf_spectacular_sidecar": "2024.5.1",
    "mptt": "0.16.0",
    "rest_framework": "3.15.1",
    "social_django": "5.4.1",
    "taggit": "5.0.1",
    "timezone_field": "6.1.0"
  },
  "netbox-version": "4.0.3",
  "plugins": {},
  "python-version": "3.11.6",
  "rq-workers-running": 2
}
```

This may be related with the issue#16228.

### Expected Behavior

The system should reject GraphQL queries from users or tokens that do not have the necessary permissions.

### Observed Behavior

During testing, it was discovered that if a user or token lacks permissions, the system does not enforce authentication for GraphQL queries. This was observed using a test user with no permissions, where GraphQL queries still returned sensitive data.

GraphQL queries are processed and data is returned even when the user or token has no permissions.

Further investigation is needed to confirm the scope and cause of the authentication bypass. Additional details will be provided upon request.

DanSheps:
### Fixes: #16292 - Properly restrict GraphQL queries for querys with pk set

* Set name on Schema type to "Query" (required as part of the default resolver generation)
* Change resolver to regular Field with appropriate type
* Set DEFAULT_PK_FIELD_NAME in settings.py to `id`


DanSheps:
Added bonus, pycharm find and replace regex for this change:

```regex
    @strawberry.field
    def (.*)\(self, id: int\) -> (.*):
        return models\..*\.objects\.get\(pk=id\)
```

```regex
    $1: $2 = strawberry_django.field()
```

netbox/netbox/settings.py
```diff
@@ -764,6 +764,7 @@ def _setting(name, default=None):
 #
 STRAWBERRY_DJANGO = {
     "TYPE_DESCRIPTION_FROM_MODEL_DOCSTRING": True,
+    "DEFAULT_PK_FIELD_NAME": "id",
     "USE_DEPRECATED_FILTERS": True,
```
DanSheps, line 768:
```suggestion
    "DEFAULT_PK_FIELD_NAME": "id",
    "TYPE_DESCRIPTION_FROM_MODEL_DOCSTRING": True,
    "USE_DEPRECATED_FILTERS": True,
```

jeremystretch:
I've replicated all of Dan's changes application-wide and it seems to work well. We still need to add a test for this though.

jeremystretch:
I've extended the existing GraphQL API tests to check for this.

Also, it seems that GraphQL is logging an exception for requests which query an invalid object ID (but still returning the expected response). I've compensated for this in the test to reduce noise, but it's something we should look into separately.

arthanson (approved):
I'm still a bit leery of the [name="Query") as I'm not sure if it's needed or has side-effects, but in testing I couldn't find any difference it causes so seems okay.

PR #17246, opened 2024-08-22
17117 change caret to html entity to fix safari performacne issue

### LINKED ISSUE #17117: Poor performance in Safari with long lists of items
### Deployment Type

Self-hosted

### NetBox Version

4.0.8

### Python Version

3.11

### Steps to Reproduce

1. Access any long list of items (reproducible for me on the demo instance, [example](https://demo.netbox.dev/ipam/ip-addresses/))
2. Scroll the list

### Expected Behavior

The list should scroll as performantly as it did prior to 4.x, with no stuttering and without hanging the browser. The browser scroll bar should update with the scroll position.

### Observed Behavior

Screenshot below has part of the story, but there's just some really strange performance issues with lists that cause my browser (non-beta Safari on non-beta macOS, latest) to hang when scrolling, but also to hang when I switch from another tab back to the tab with the list in it.

Screenshot 2024-08-09 at 3 22 17 PM

I’ve been seeing this since 4.x with the UI refresh, haven't had time to write it up until now/was hoping somebody else was running into this and it would get solved, but it seems like it's ongoing. 😅

arthanson:
### Fixes: #17117 

Tabler was creating the caret for the sidebar and buttons via a border and then doing a transform rotation of 45-degrees.  This was causing a performance issue on safari. Changed the css to get rid of the transform and use the material icon for the caret.

In screenshot left is old, right is new:

caret3


jeremystretch (changes requested):
The icon isn't aligned vertically with the menu item text:

screenshot

jeremystretch (approved):

PR #17247, opened 2024-08-22
17219 fix custom validator display if function

### LINKED ISSUE #17219: System page - Server Error if Custom Validation Logic is used
### Deployment Type

Self-hosted

### NetBox Version

4.0.9

### Python Version

3.11

### Steps to Reproduce

This is also described at https://github.com/netbox-community/netbox/discussions/17214

If I in configuration.py add a custom validator like:

```

class UniqueSerial(CustomValidator):

    def validate(self, instance):
        from dcim.models import Device

        if instance.serial and Device.objects.exclude(pk=instance.pk).filter(serial=instance.serial).exists():
            self.fail("The Serial number already exists in NetBox", field='serial')

CUSTOM_VALIDATORS = {
    'dcim.device': (
        UniqueSerial(),
    )
}
```

the menu item "Sytem" at /core/system/ fails with:

Object of type UniqueSerial is not JSON serializable

Python version: 3.11.6
NetBox version: 4.0.9
Plugins: 
  netbox_secrets: 2.0.3
  netbox_topology_views: 4.0.0

### Expected Behavior

The page should be rendered correct and show system data.

### Observed Behavior

If I remove 

`CUSTOM_VALIDATORS = {
    'dcim.device': (
        UniqueSerial(),
    )
}`

from configuration.py, the page renders nicely

if seems to me that the page render tries to see  UniqueSerial() as a json, and not code

arthanson:
### Fixes: #17219 

Tested with function custom validator:

```

class UniqueSerial(CustomValidator):

    def validate(self, instance):
        from dcim.models import Device

        if instance.serial and Device.objects.exclude(pk=instance.pk).filter(serial=instance.serial).exists():
            self.fail("The Serial number already exists in NetBox", field='serial')

CUSTOM_VALIDATORS = {
    'dcim.device': (
        UniqueSerial(),
    )
}
```
System | NetBox 2024-08-22 11-39-30

and with dictionary based:
```
CUSTOM_VALIDATORS = {
    "dcim.device": [
        {
            "name": {
                "regex": "[a-z]+\d{3}"
            },
            "asset_tag": {
                "required": True
            }
        }
    ]
}
```
System | NetBox 2024-08-22 11-38-56


netbox/core/views.py
```diff
@@ -560,6 +560,12 @@ def get(self, request):
             # Fall back to using the active config data if no record is found
             config = get_config()
 
+        # If Custom Validators is function (in configuration.py) get function name
+        if hasattr(config, 'CUSTOM_VALIDATORS') and config.CUSTOM_VALIDATORS:
+            for k, v in config.CUSTOM_VALIDATORS.items():
+                if isinstance(v, tuple):
```
jeremystretch, line 566: The type of every item in the list of configured validators needs to be checked.

jeremystretch (changes requested):
This fails when configuring both class- and dictionary-based validators:

```python
CUSTOM_VALIDATORS = {
    'dcim.device': [
        {
            "name": {
                "min_length": 5
            }
        },
        UniqueSerial(),
    ]
}
```

jeremystretch (approved):

PR #17248, opened 2024-08-22
17186 change custom link button color from outline-dark to outline-se…

### LINKED ISSUE #17186: Default class custom links invisible in dark mode
### Deployment Type

NetBox Enterprise

### NetBox Version

v4.0.8

### Python Version

3.12

### Steps to Reproduce

1. Create a working custom link
2. Enable dark mode

### Expected Behavior

Button should be visible

### Observed Behavior

Button is invisible until you mouse over.
mouseover: Screenshot 2024-08-15 at 10 50 07 PM
regular: Screenshot 2024-08-15 at 10 50 03 PM

A similar bug was fixed in #15616 but for some reason it only fixes non-working links.  Working links are still invisible.

arthanson:
### Fixes: #17186 

The link button choices has default as `outline-dark` which it appends to btn- to get the button color.  outline-dark will not show in dark mode, so as previous changes to other buttons change to outline-secondary.  Migration was needed to change this is in the database as it uses the database char value to create the button class.

123 | NetBox 2024-08-22 12-47-15


netbox/netbox/choices.py
```diff
@@ -84,7 +84,7 @@ class ButtonColorChoices(ChoiceSet):
     """
     Map standard button color choices to Bootstrap 3 button classes
     """
-    DEFAULT = 'outline-dark'
+    DEFAULT = 'outline-secondary'
```
jeremystretch, line 87: IMO we should default to an empty value or the string "default," and avoid hard-coding a specific color/style name (so we don't end up in this situation again down the road).

jeremystretch (changes requested):

netbox/netbox/choices.py
arthanson: changed to default, the link isn't rendered in the template so had to add some code to netbox/extras/templatetags/custom_links.py to change from default to outline-secondary.

jeremystretch (approved):

PR #17260, opened 2024-08-23
Fixes: #16905 - Allow filtering on Device Status in InventoryItemTable

### LINKED ISSUE #16905: Allow filtering of Inventory Items by Device Status
### NetBox version

3.7.8

### Feature type

New functionality

### Proposed functionality

Extend the filterable options for Inventory Items to include Device Status

### Use case

Currently the Filter tab on Inventory Items allows you to filter by Device Type, Device Role & Device. We use the Device Status attribute to determine if a server is actively provisioned in our environment or not so it would be useful to be able to filter Inventory Items by device status to assist in locating an unused server with a given inventory. 

i.e. 
Site: London
Device Status: Inventory
Role: CPU
Label Contains: Gold

### Database changes

_No response_

### External dependencies

_No response_

bctiemann:
### Fixes: #16905 Allow filtering on Device Status in InventoryItemTable

Adds `device_status` as a filter field in Inventory Item list view, and associated filter in `InventoryItemFilterSet`.

Screenshot 2024-08-23 at 5 27 48 PM


netbox/dcim/filtersets.py
```diff
@@ -1795,6 +1795,10 @@ class InventoryItemFilterSet(DeviceComponentFilterSet, NetBoxModelFilterSet):
         to_field_name='slug',
         label=_('Role (slug)'),
     )
+    device_status = django_filters.MultipleChoiceFilter(
```
jeremystretch, line 1798: We can move this to `DeviceComponentFilterSet`, to ensure it's supported for _all_ device components (console ports, interfaces, etc.).

netbox/dcim/forms/filtersets.py
```diff
@@ -1461,6 +1461,11 @@ class InventoryItemFilterForm(DeviceComponentFilterForm):
             choices=BOOLEAN_WITH_BLANK_CHOICES
         )
     )
+    device_status = forms.MultipleChoiceField(
```
jeremystretch, line 1464: Likewise, this can be moved to `DeviceComponentFilterForm`. We'll need to add the field to a fieldset in each of the individual components forms though.

netbox/dcim/models/device_components.py
```diff
@@ -1265,6 +1265,9 @@ class Meta:
     def get_absolute_url(self):
         return reverse('dcim:inventoryitem', kwargs={'pk': self.pk})
 
+    def get_device_status_color(self):
```
jeremystretch, line 1268: We probably want to avoid introducing "proxy" methods that only exist to return the value of an attribute of some related object. Instead, maybe we can refactor `ChoiceFieldColumn` to accept a custom function for determining the appropriate color. Then we could do something like this:

```
device_status = columns.ChoiceFieldColumn(
    accessor=tables.A('device__status'),
    verbose_name=_('Device Status'),
    color=lambda x: x.device.get_status_color()
)
```

jeremystretch (changes requested):

netbox/dcim/filtersets.py
bctiemann: Ah, good call, I was going to ask if there was a central place where this stuff can live so we aren't redundantly implementing it on every similar object type.

netbox/dcim/models/device_components.py
bctiemann: I like where this is going. Let me know if you think the preference order I've got here looks right to you -- it takes the `color` callable first if provided, then falls back to the try-except of `get_FOO_color` as before.

netbox/dcim/filtersets.py
```diff
@@ -1479,7 +1483,7 @@ class ConsolePortFilterSet(
 
     class Meta:
         model = ConsolePort
-        fields = ('id', 'name', 'label', 'speed', 'description', 'mark_connected', 'cable_end')
+        fields = ('id', 'name', 'label', 'speed', 'description', 'mark_connected', 'cable_end', 'device_status')
```
jeremystretch, line 1486: Filters defined on the filterset class don't need to be added here; `fields` just denotes the model fields for which filters should be generated automatically.

jeremystretch (changes requested):

jeremystretch (changes requested):
We should include tests for the `device_status` filters for all relevant models in `dcim/tests/test_filtersets.py`.

jeremystretch (approved):

PR #17268, opened 2024-08-26
Fixes: #16973 - Resolve $user token to User.id for use in permissions based on custom fields

### LINKED ISSUE #16973: Object of type User is not JSON serializable
### Deployment Type

Self-hosted

### NetBox Version

v3.7.6

### Python Version

3.12

### Steps to Reproduce

I created a custom_field for device roles content type named "created_by" .
 It accepts Users > User as object type. 
After I have created a permission with {"custom_field_data__created_by": "$user"} as constraint, I get 'Object of type User is not JSON serializable' when I try to either change or delete a device role created by myself. Putting my user id instead of "$user" works as expected.

### Expected Behavior

Constraint with {"custom_field_data__created_by": "$user"} should recognize the User token

### Observed Behavior

The user token "$user" is not recognized and I get 'Object of type User is not JSON serializable'

bctiemann:
### Fixes: #16973 - Resolve $user token to User.id for use in permissions based on custom fields

Adds special handling for `CONSTRAINT_TOKEN_USER` ("$user" token) when used in custom fields in object-based permissions.

### DIFF
--- a/netbox/utilities/permissions.py
+++ b/netbox/utilities/permissions.py
@@ -97,7 +97,7 @@ def qs_filter_from_constraints(constraints, tokens=None):
         if type(value) is list:
             return list(map(lambda v: tokens.get(v, v), value))
         User = apps.get_model('users.User')
-        if value == CONSTRAINT_TOKEN_USER and type(tokens[CONSTRAINT_TOKEN_USER] is User):
+        if value == CONSTRAINT_TOKEN_USER and isinstance(tokens.get(CONSTRAINT_TOKEN_USER), User):
             return tokens[CONSTRAINT_TOKEN_USER].id
         return tokens.get(value, value)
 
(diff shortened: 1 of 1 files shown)


netbox/utilities/permissions.py
```diff
@@ -93,6 +96,9 @@ def qs_filter_from_constraints(constraints, tokens=None):
     def _replace_tokens(value, tokens):
         if type(value) is list:
             return list(map(lambda v: tokens.get(v, v), value))
+        User = apps.get_model('users.User')
+        if value == CONSTRAINT_TOKEN_USER and isinstance(tokens.get(CONSTRAINT_TOKEN_USER), User):
+            return tokens[CONSTRAINT_TOKEN_USER].id
```
jeremystretch, line 101: We need to catch the presence of the token inside a list of values as well (above). It might be simpler to just swap out the token value with an integer before we start iterating through the constraints.

jeremystretch (changes requested):

netbox/utilities/permissions.py
bctiemann: Agreed, this is much nicer.

jeremystretch (approved):
Nice work @bctiemann!

PR #17270, opened 2024-08-26
Closes #17256: Fix translation support in VLAN group scope assignment form

### LINKED ISSUE #17256: VLAN group scope selection fails for languages other than English
### Deployment Type

Self-hosted

### NetBox Version

v4.0.9

### Python Version

3.10

### Steps to Reproduce

1.  try Netbox in Portuguese
2. go to VLAN Groups (Grupos de VLANS)
3. add a new group
4. In the "Scope" area, select "DCIM > Location" (DCIM > localização)

### Expected Behavior

It should show the options to select the region, site, and location.

### Observed Behavior

The options are not displayed, and when trying to save, the following error appears: _It is not possible to set scope_type without scope_id._

jeremystretch:
### Fixes: #17256

- Convert VLANGroupForm to use a single dynamically-rendered `scope` field
- Ditch an entire TypeScript module dedicated to displaying the old form :tada:


arthanson (changes requested):
It's not working for me, I switched branches to this, cleared cache and reloaded then go to add a VLan Group, selecting scope-type to location doesn't change the form:

Add a new VLAN group | NetBox 2024-08-26 15-18-57

jeremystretch:
Changing the scope type should reset the list of choices in the "scope" field below it. (We're only using a single field now with a selector modal, regardless of the type.)

arthanson (approved):
Not sure if an issue - but if you edit and change the scope-type the scope is not cleared, so if you have id's between the different tables that are the same you will automatically get the item in the new table with the same id.  Not sure if it would be better to clear the scope on scope-type change?

PR #17293, opened 2024-08-28
Fixes: 17292 - Detect infinite loop condition while iterating terminations in CablePath.from_origin

### LINKED ISSUE #17292: Detect infinite loop in cable paths
### Deployment Type

NetBox Cloud

### NetBox Version

v4.0.9

### Python Version

3.10

### Steps to Reproduce

Currently it is possible to create an infinite loop in cabling topology (for instance, a Device with Interface -> Patch Panel Front Port -> Patch Panel Rear Port). This causes `CablePath.from_origin` to loop infinitely while trying to resolve the termination path.

Create Cable connections as follows:

Device (DA): Application Server, APC AP7901
Create Interface 01 (fixed Ethernet)

Patch Panel (PA): Generic 48-pair fiber panel
```
Device (DA) [Interface]
        | (Cable 1)
PatchPanel (PA) Front Port 1
PatchPanel (PA) Front Port 2
        | (Cable 2)
PatchPanel (PA) Rear Splice
```
(Issue also occurs if you create a Rear Port 1 on PA and connect Front Port 2 to Rear Port 1.)

When creating the Cable 2 object, clicking Save leads to an infinitely spinning/wedged browser as the app continuously iterates.

### Expected Behavior

Cable should be created. If the cable path is not valid, the trace diagram should not be expected to redraw correctly.

### Observed Behavior

Cable object fails to save, or is partially saved, while the `from_origin` method iterates infinitely. User has to refresh the browser and the app potentially suffers degraded performance while the session survives.

bctiemann:
### Fixes: #17292

Adds a check to Step 4 of `CablePath.from_origin` to make sure the path length is not exceeding `CABLE_TRACE_MAX_LENGTH` (a new setting).

Also adds an optional `max_length` parameter to `from_origin` for use as a testing hook, to determine whether the loop detection is working.


netbox/dcim/models/cables.py
```diff
@@ -714,6 +718,14 @@ def from_origin(cls, terminations):
                     is_split = True
                 break
 
+            # Detect infinite loop in cabling topology
+            num_elements_added = len(path) - len(prev_path)
+            prev_elements_added = prev_path[-num_elements_added:]
+            elements_added = path[-num_elements_added:]
+            if elements_added == prev_elements_added:
+                logger.warning('Infinite loop detected while updating cable path trace')
+                break
```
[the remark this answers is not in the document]
bctiemann: Please re-review; I moved the check to earlier in the process and made it more straightforward. It only checks cables in the body of the trace, not the beginning and end terminations, but I doubt there's risk there.

netbox/dcim/tests/test_cablepaths.py
```diff
@@ -2343,3 +2344,41 @@ def test_401_exclude_midspan_devices(self):
             is_active=True
         )
         self.assertEqual(CablePath.objects.count(), 0)
+
+    def test_detect_infinite_loop(self):
```
DanSheps, line 2348: Is this test suppose to fail or succeed?  As it stands I don't see a loop in the topology unuless I am missing something.

DanSheps (changes requested):

netbox/dcim/tests/test_cablepaths.py
bctiemann: If there's an infinite loop then the test will just keep executing. I'm not sure of an elegant way to test the infinite looping condition, but as long as it executes without the whole test process hanging then the test passes.

I've tried a few different approaches for making it assert if the execution takes too long or loops too many times, mostly involving timeouts and/or threading, but I don't know if it's worth going to that extreme. What would you suggest?

In any case if you comment out lines 600-602 in cables.py, the test will hang indefinitely.

DanSheps: IMO, for testing purposes we may want to add a parameter that sets a maximum path length (might not be bad anyways to set it to an unreasonable number) and still create a loop.

IDK if it is worth it though TBH.

jeremystretch (changes requested):
It seems like this condition can be avoided by checking for the presence of each encountered termination in the path so far: If we see the same termination twice, we're in a loop. Could this be implemented under `Cable.trace()`?

bctiemann:
Short answer -- no, I don't think so. The `from_origin` method that has the infinite loop possibility in it is called from a variety of places, including `create_cablepath` and `retrace_cable_paths`, and defending against it at that level would have to be done in all those locations. Similarly `Cable.trace` is not called in the post-save signal flow that triggers the infinite loop condition I found; it goes straight to `from_origin` where it falls into the loop. I think the only reasonable place to handle this is in `from_origin` itself.

bctiemann:
I'm going to close this PR, as it's sucking up a lot of time that is better spent elsewhere. I believe the solution here is the correct and least invasive way to deal with this issue, so if anyone picks up this issue in the future this is a potential valid place to start from.

Issue #17289, opened 2024-08-28
Enforce a standard policy for local passwords by default


jeremystretch:
### NetBox version

v4.0.9

### Feature type

Change to existing functionality

### Proposed functionality

Although NetBox supports the enforcement for [configurable password policy](https://netboxlabs.com/docs/netbox/en/stable/configuration/security/#auth_password_validators), it does not assert any policy by default. This FR proposes defining a simple default compliance policy for local passwords. The following criteria are proposed:

- Minimum length of 12 characters
- At least one each of lowercase characters, uppercase characters, numeric digits, and symbols

(This new _default_ policy can be disabled by setting `AUTH_PASSWORD_VALIDATORS = []` in the NetBox configuration.)

### Use case

This will provide a reasonable baseline for ensuring the use of strong local passwords.

RangerRick:
FYI, this is overkill for defaults, but I ended up making a custom auth plugin that implements OWASP's recommendations for a complex password, using a regex. Putting it here in case anyone comes looking for something similar. :)

```python
from django.core.exceptions import ValidationError
from django.utils.translation import gettext as _
import re

class OWASPValidator:
  def __init__(self):
    self.owasp_regex = re.compile('^(?:(?=.*\d)(?=.*[A-Z])(?=.*[a-z])|(?=.*\d)(?=.*[^A-Za-z0-9])(?=.*[a-z])|(?=.*[^A-Za-z0-9])(?=.*[A-Z])(?=.*[a-z])|(?=.*\d)(?=.*[A-Z])(?=.*[^A-Za-z0-9]))(?!.*(.)\1{2,})[A-Za-z0-9!~<>,;:_=?*+#."&§%°()\|\[\]\-\$\^\@\/]{12,128}$')

  def validate(self, password, user=None):
    if not self.owasp_regex.match(password):
      raise ValidationError(
        _("Password must meet the current OWASP recommendation for complex passwords: 12 to 128 characters, requiring at least 3 out 4 of uppercase and lowercase letters, numbers and special characters, and no more than 2 equal characters in a row."),
        code="password_too_weak",
      )

  def get_help_text(self):
    return _(
      "Password must meet the current OWASP recommendation for complex passwords: 12 to 128 characters, requiring at least 3 out 4 of uppercase and lowercase letters, numbers and special characters, and no more than 2 equal characters in a row."
    )
```

PR #17299, opened 2024-08-29
17289 enforce minimum password strength

arthanson:
### Fixes: #17289 

Enforce default minimum password strength:
* 12 characters minimum
* similarities to user attributes (first name, last name, email)
* common password check
* must contain one number, one upper-case letter and one lower-case letter.

Change Password | NetBox 2024-08-29 08-58-46

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -76,7 +76,7 @@ AUTH_PASSWORD_VALIDATORS = getattr(configuration, 'AUTH_PASSWORD_VALIDATORS', [
         "NAME": "django.contrib.auth.password_validation.CommonPasswordValidator",
     },
     {
-        "NAME": "utilities.password_validation.NumericAlphaPasswordValidator",
+        "NAME": "utilities.password_validation.AlphanumericPasswordValidator",
     },
 ])
 BASE_PATH = trailing_slash(getattr(configuration, 'BASE_PATH', ''))
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -64,9 +64,6 @@ ALLOW_TOKEN_RETRIEVAL = getattr(configuration, 'ALLOW_TOKEN_RETRIEVAL', True)
 ALLOWED_HOSTS = getattr(configuration, 'ALLOWED_HOSTS')  # Required
 AUTH_PASSWORD_VALIDATORS = getattr(configuration, 'AUTH_PASSWORD_VALIDATORS', [
     {
-        "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",
-    },
-    {
         "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
         "OPTIONS": {
             "min_length": 12,
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -21,7 +21,6 @@ from netbox.plugins import PluginConfig
 from utilities.release import load_release_data
 from utilities.string import trailing_slash
 
-
 #
 # Environment setup
 #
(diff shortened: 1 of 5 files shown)


netbox/users/tests/test_views.py
```diff
@@ -60,19 +60,32 @@ def setUpTestData(cls):
             'last_name': 'newlastname',
         }
 
-    @override_settings(AUTH_PASSWORD_VALIDATORS=[{
-        'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
-        'OPTIONS': {'min_length': 8}
-    }])
+    @override_settings(AUTH_PASSWORD_VALIDATORS=[
```
jeremystretch, line 63: Why not just leave the default policy in place?

netbox/users/tests/test_api.py
jeremystretch, line 1: Same notes here as for `test_views.py`.

netbox/utilities/password_validation.py
```diff
@@ -0,0 +1,27 @@
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext as _
+
+
+class NumericAlphaPasswordValidator:
```
jeremystretch, line 5:
```suggestion
class AlphanumericPasswordValidator:
```

netbox/utilities/password_validation.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Validate that the password has at least one numeral, one uppercase letter and one lowercase letter.
+    """
+
+    def validate(self, password, user=None):
+        if not any(char.isdigit() for char in password):
+            raise ValidationError(
+                _("Password should have at least one numeral"),
```
jeremystretch, line 13:
```suggestion
                _("Password must have at least one numeral."),
```

netbox/utilities/password_validation.py
```diff
@@ -0,0 +1,27 @@
...
+            )
+
+        if not any(char.isupper() for char in password):
+            raise ValidationError(
+                _("Password should have at least one uppercase letter"),
```
jeremystretch, line 18:
```suggestion
                _("Password must have at least one uppercase letter."),
```

netbox/utilities/password_validation.py
```diff
@@ -0,0 +1,27 @@
...
+            )
+
+        if not any(char.islower() for char in password):
+            raise ValidationError(
+                _("Password should have at least one lowercase letter"),
```
jeremystretch, line 23:
```suggestion
                _("Password must have at least one lowercase letter."),
```

netbox/users/tests/test_views.py
```diff
@@ -84,10 +97,30 @@ def test_password_validation_enforced(self):
...
+
+        # Password no uppercase
+        data['password'] = 'foobar123abc'
+        data['confirm_password'] = 'foobar123abc'
+        self.assertHttpStatus(self.client.post(**request), 200)
+
+        # Password no lowercase
+        data['password'] = 'FOOBAR123ABC'
+        data['confirm_password'] = 'FOOBAR123ABC'
+        self.assertHttpStatus(self.client.post(**request), 200)
```
jeremystretch, line 122: We can move the individual validation logic tests to a separate test case user `users`; there's no need to test the validators for both UI and API requests.

jeremystretch (changes requested):

netbox/netbox/settings.py
```diff
@@ -63,7 +62,23 @@
 ADMINS = getattr(configuration, 'ADMINS', [])
 ALLOW_TOKEN_RETRIEVAL = getattr(configuration, 'ALLOW_TOKEN_RETRIEVAL', True)
 ALLOWED_HOSTS = getattr(configuration, 'ALLOWED_HOSTS')  # Required
-AUTH_PASSWORD_VALIDATORS = getattr(configuration, 'AUTH_PASSWORD_VALIDATORS', [])
+AUTH_PASSWORD_VALIDATORS = getattr(configuration, 'AUTH_PASSWORD_VALIDATORS', [
+    {
+        "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",
```
jeremystretch, line 67: This doesn't seem to validate when creating a new user. For example, I can create a user with the name & password of `NewUser123456`. However, if I then edit the user and set the same password, I get a validation error complaining that the password is too similar to the username.

AFAICT this is all stock Django functionality so I'm not sure what's going on here.

netbox/netbox/settings.py
```diff
@@ -63,7 +62,23 @@
...
+    },
+    {
+        "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
+        "OPTIONS": {
+            "min_length": 12,
+        },
+    },
+    {
+        "NAME": "django.contrib.auth.password_validation.CommonPasswordValidator",
```
jeremystretch, line 76: IMO we should omit this validator, for two reasons:

1. We don't control the list of "common" passwords, and it doesn't appear to be formally documented.
2. Our minimum length validation will disqualify nearly all of [these strings](https://gist.github.com/roycewilliams/226886fd01572964e1431ac8afc999ce) anyway.

netbox/users/tests/test_api.py
```diff
@@ -111,10 +111,30 @@ def test_password_validation_enforced(self):
...
+
+        # Password no uppercase
+        data['password'] = 'foobarfoo1'
+        response = self.client.post(url, data, format='json', **self.header)
+        self.assertEqual(response.status_code, 400)
+
+        # Password no lowercase
+        data['password'] = 'FOOBARFOO1'
+        response = self.client.post(url, data, format='json', **self.header)
+        self.assertEqual(response.status_code, 400)
```
jeremystretch, line 136: This set of logic tests is duplicated for the UI and API. Can we move it to a single test module dedicated to testing password validation on the User model itself?

jeremystretch (changes requested):

netbox/users/tests/test_api.py
arthanson: IMHO: I think it needs to be here as https://github.com/netbox-community/netbox/blob/develop/netbox/users/forms/model_forms.py#L231 specifically calls validate_password so these checks make sure it is getting called appropriately in the form.

netbox/netbox/settings.py
arthanson: that was tricky, needs to be in _post_clean as in clean self.instance isn't set on create.

jeremystretch (approved):

Issue #17301, opened 2024-08-29
Incorrect styling on edit & delete buttons for script module


jeremystretch:
### Deployment Type

NetBox Cloud

### NetBox Version

v4.1-beta1

### Python Version

3.10

### Steps to Reproduce

1. Upload a custom script module

### Expected Behavior

The edit & delete buttons for the script module should be consistent with the rest of the UI.

### Observed Behavior

The buttons have dark text rather than light (see below).

![screenshot](https://github.com/user-attachments/assets/260a1cf5-4248-4732-8410-27f149be91fe)

PR #17311, opened 2024-08-30
17301 script buttons

arthanson:
### Fixes: #17301 

Scripts | NetBox 2024-08-29 18-09-39

### DIFF
--- a/netbox/templates/extras/script_list.html
+++ b/netbox/templates/extras/script_list.html
@@ -22,21 +22,28 @@
   {% for module in script_modules %}
     {% include 'inc/sync_warning.html' with object=module %}
     <div class="card">
-      <h2 class="card-header justify-content-between" id="module{{ module.pk }}">
-        <div>
-          <i class="mdi mdi-file-document-outline"></i> {{ module }}
+
+      <div class="card-header container-fluid">
+        <div class="col col-md-10">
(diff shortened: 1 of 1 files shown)


netbox/templates/extras/script_list.html
```diff
@@ -22,23 +22,28 @@
...
-          {% if perms.extras.delete_scriptmodule %}
-            <a href="{% url 'extras:scriptmodule_delete' pk=module.pk %}" class="btn btn-danger btn-sm">
-              <i class="mdi mdi-trash-can-outline" aria-hidden="true"></i> {% trans "Delete" %}
-            </a>
-          {% endif %}
+        <div class="col col-md-2 d-print-none">
+          <div class="btn-list justify-content-end mb-2">
+            {% if perms.extras.edit_scriptmodule %}
+              {% url 'extras:scriptmodule_edit' pk=module.pk as edit_url %}
+              {% include 'buttons/edit.html' with url=edit_url %}
```
jeremystretch, line 38: These templates exist solely for the button template tags; we should avoid including them directly.

jeremystretch (changes requested):
We want to keep the buttons the same size; just the text color needs to be fixed. This probably needs to be adjusted in the CSS.

screenshot

jeremystretch:
This was addressed under #17328.

PR #17312, opened 2024-08-30
Fixes: #17310 - Properly restrict GraphQL related object queries

### LINKED ISSUE #17310: GraphQL related objects are not properly restricted
### Deployment Type

NetBox Cloud

### NetBox Version

v4.0.10

### Python Version

3.12

### Steps to Reproduce

1. Create a site
2. Create an ASN in the site (Ex: 64511, pk=1)
3. Create a user with permissions to only view ASN's (not sites)
4. Run the following GQL:  `{ asn(id: 1) { id, sites {} } }`

### Expected Behavior

Will show the ASN but not the related site

### Observed Behavior

Related site is shown

DanSheps:
### Fixes: #17310

* Properly restrict GraphQL related object queries
* Update tests to handle additional permission requirements

### DIFF
--- a/netbox/circuits/tests/test_api.py
+++ b/netbox/circuits/tests/test_api.py
@@ -96,6 +96,7 @@ class CircuitTest(APIViewTestCases.APIViewTestCase):
     bulk_update_data = {
         'status': 'planned',
     }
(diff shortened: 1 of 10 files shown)


jeremystretch (approved):
Thanks @DanSheps!

PR #17322, opened 2024-08-30
runscript: use the previously defined "user"

### LINKED ISSUE #17321: manage.py runscript --user ignored
### Deployment Type

Self-hosted

### NetBox Version

v4.0.8

### Python Version

3.11

### Steps to Reproduce

1. Create two superuser users (user1, user2)
2. run a script through `python manage.py runscript` while specifying `--user user2`

### Expected Behavior

The job should have been ran as the "user2".

### Observed Behavior

The job is ran by the user with the lowest ID In our case user1.

XioNoX:
Oops, only seeing now that this might be closed automatically as the linked ticket is not triaged yet.

### Fixes: #17321

There is already a logic set earlier in the code to define "user" if --user is passed as parameter, and default to the user with the lowest ID no none is provided.
This patch uses this "user" to run the job instead of always applying the default.

### DIFF
--- a/netbox/extras/management/commands/runscript.py
+++ b/netbox/extras/management/commands/runscript.py
@@ -125,7 +125,7 @@ class Command(BaseCommand):
         job = Job.objects.create(
             object=module,
             name=script.class_name,
-            user=User.objects.filter(is_superuser=True).order_by('pk')[0],
+            user=user,
             job_id=uuid.uuid4()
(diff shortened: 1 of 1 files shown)


DanSheps:
In the future please wait for the bug to be triaged.

:D

jeremystretch (approved):

Issue #8198, opened 2021-12-30
Add the ability to require unique input to custom field


K0HAX:
### NetBox version

v3.1.1

### Feature type

New functionality

### Proposed functionality

Currently, custom fields are never checked for unique content.  
It would be nice to be able to add checks for uniqueness inside custom fields as an option to that specific custom field.

### Use case

I am using custom fields to add a unique name similar to a Common Language Location Identifiers (CLLI) name for each site.  
No two sites should have the same unique name from this field.

### Database changes

The database would probably need a new column to track if the custom field needs a uniqueness check, and in the general case, what the scope of the uniqueness check should be (example: global, regional, site, none).

### External dependencies

None

sdktr:
Maybe this can already be solved with a 'custom validator'?

The uniqueness scoping that you seam to have in mind are not easily applied to all other netbox users imo

DanSheps:
As pointed out, this should be solvable with a custom validator.

The logic implemented would be the same as we would implement.  It would need to be a custom validation class, so perhaps adding a new option to the base custom validation of "unique" might be the best way forward here.

jeremystretch:
> Maybe this can already be solved with a 'custom validator'?

It can be, although [custom validation](https://netbox.readthedocs.io/en/stable/customization/custom-validation/) doesn't currently support uniqueness checking as a standard validator: It would require the introduction of a custom class. (This might be an FR on its own.)

I do think there's value in providing optional uniqueness enforcement for custom fields, though, just as we provide min/max and regex validation. Actually enforcing the validation could get tricky, though.

I'm going to tag this for milestone assignment as I think it's worth exploring.

ryanmerolle:
This could be a good candidate for 3.6

PR #17325, opened 2024-08-30
#8198: Rename CustomField.validation_unique to unique

jeremystretch:
Rename the `validation_unique` field added on CustomField in #8198 to `unique`, so that it is more consistent with the `required` boolean field.


arthanson (changes requested):
https://github.com/netbox-community/netbox/blob/feature/docs/models/extras/customfield.md should be updated as well

PR #17329, opened 2024-08-30
Closes #17288: Limit the number of aliases within a GraphQL API requests to 10

### LINKED ISSUE #17288: Limit the number of aliases within a GraphQL API request
### NetBox version

v4.0.9

### Feature type

New functionality

### Proposed functionality

Employ Strawberry's [`MaxAliasesLimiter`](https://strawberry.rocks/docs/extensions/max-aliases-limiter) to limit the number of aliases defined within a single request.

### Use case

Mitigates potential GraphQL API attacks which leverage computationally expensive queries.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #17288

Limit a GraphQL API request to a maximum of 10 aliases.

### DIFF
--- a/docs/configuration/graphql-api.md
+++ b/docs/configuration/graphql-api.md
@@ -1,0 +1,17 @@
+# GraphQL API Parameters
+
+## GRAPHQL_ENABLED
+
+!!! tip "Dynamic Configuration Parameter"
+
+Default: True
+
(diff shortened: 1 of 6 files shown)


arthanson (approved):
I would think we probably want that as a configurable parameter?  Potentially some people might want larger queries?

Issue #17320, opened 2024-08-30
Update UI screenshots for v4.1


jeremystretch:
### Change Type

Correction

### Area

Features

### Proposed Changes

Update all screenshots featuring the user interface to reflect the changes in v4.1.

PR #17331, opened 2024-08-30
Fixes: #17320 - New screenshots for 4.1 UI

bctiemann:
### Fixes: #17320

New screenshots for the 4.1 UI/colors. Also normalizes all the screenshots to the same width.


bctiemann:
This should probably be squashed.

docs/media/screenshots/cable-trace.png
jeremystretch, line 1: This is much taller than the original. It can be cropped to stop at the bottom of the trace diagram.

docs/media/screenshots/home-dark.png
jeremystretch, line 1: The side navigation bar doesn't appear to render correctly in these screenshots. It doesn't extend to the bottom of the page.

jeremystretch (changes requested):

docs/media/screenshots/home-dark.png
bctiemann: Hmm, yeah -- I was using the built-in Chrome screenshot tool, I might have to do it manually in Photoshop if it doesn't trigger the right rendering stuff this way.

docs/media/screenshots/cable-trace.png
bctiemann: Recaptured this; note that it's taller but also wider, so it can be scaled down 1:1 if it needs to render at a certain height.

jeremystretch (approved):

PR #17342, opened 2024-09-03
Hide exception in ObjectCountsWidget for models without a `xxx_list` view function

### LINKED ISSUE #17341: ObjectCountsWidget can brick Netbox home page
### Deployment Type

Self-hosted

### NetBox Version

v4.0.10

### Python Version

3.12

### Steps to Reproduce

1. Go to home page
2. Click on settings cog next to an Object Counts widget
3. Add "DCIM > Console Server Port Template" to the list of objects to count, and click Save

image

### Expected Behavior

Netbox continues to function, or the change is rejected.

### Observed Behavior

Netbox web UI crashes and is unusable. There is no way to recover from the web UI.

image

This is because there is a hard-coded assumption in the widget that all models have a `foo_list` endpoint, but template components don't.

Originally raised in discussion #17336

candlerb:
### Fixes #17341

This is a quick workaround specifically to stop ObjectCountsWidget raising an exception for models without a `xxx_list` view function.

(A better, more generally useful solution would be to catch exceptions raised by *any* widget, and allow the widget pane to be displayed with an error message. That would avoid Netbox becoming unusable for any widget problem, and would also allow the widget settings to be edited to correct the problem)

### DIFF
--- a/netbox/extras/dashboard/widgets.py
+++ b/netbox/extras/dashboard/widgets.py
@@ -185,11 +185,11 @@ class ObjectCountsWidget(DashboardWidget):
             if request.user.has_perm(permission):
                 try:
                     url = reverse(get_viewname(model, 'list'))
-                except Exception:
-                    url = '#'
+                except NoReverseMatch:
+                    url = None
                 qs = model.objects.restrict(request.user, 'view')
                 # Apply any specified filters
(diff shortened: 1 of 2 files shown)


bctiemann (approved):

jeremystretch (approved):
Thanks @candlerb!

PR #17368, opened 2024-09-04
17354 fix import with custom-field

### LINKED ISSUE #17354: Import page breaks when an object custom field exists
### Deployment Type

Self-hosted

### NetBox Version

v4.1.0

### Python Version

3.10

### Steps to Reproduce

This has been reproduced on https://demo.netbox.dev/
1. Create a custom field with the following parameters:
a) Object types: DCIM > Device
b) Name: test1
c) Type: Object
d) Related object type: DCIM > Site
2. Browse to https://demo.netbox.dev/dcim/devices/import/

### Expected Behavior

Page should load without error

### Observed Behavior

Page fails to load with message:
<class 'TypeError'>

Field.__init__() got an unexpected keyword argument 'query_params'

arthanson:
### Fixes: #17354 

Only DynamicModelChoiceField has query_params, CSVModelChoiceField does not so this only includes the parameter if the field is of the correct type.

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -524,36 +524,30 @@ class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
         # Object
         elif self.type == CustomFieldTypeChoices.TYPE_OBJECT:
             model = self.related_object_type.model_class()
-            if for_csv_import:
-                field = CSVModelChoiceField(
-                    queryset=model.objects.all(),
-                    required=required,
-                    initial=initial,
-                )
-            else:
-                field = DynamicModelChoiceField(
-                    queryset=model.objects.all(),
-                    required=required,
-                    initial=initial,
-                    query_params=self.related_object_filter
-                )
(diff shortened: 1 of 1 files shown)


netbox/extras/models/customfields.py
```diff
@@ -524,24 +524,36 @@ def to_form_field(self, set_initial=True, enforce_required=True, enforce_visibil
...
+                    required=required,
+                    initial=initial,
+                )
+            else:
+                field = DynamicModelChoiceField(
+                    queryset=model.objects.all(),
+                    required=required,
+                    initial=initial,
+                    query_params=self.related_object_filter
+                )
```
jeremystretch, line 539: We can avoid duplicating these lines by preparing the field class' keyword arguments separately. (This needs to be done for the multi-object field type as well.)

```suggestion
            kwargs = {
                'queryset': model.objects.all(),
                'required': required,
                'initial': initial,
            }
            if not for_csv_import:
                kwargs['query_params'] = self.related_object_filter
            field = CSVModelChoiceField(**kwargs)
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #17391, opened 2024-09-05
Fixes: #17358 - Ensure correct comparison of overlapping IPRanges

### LINKED ISSUE #17358: Overlapping IPRanges are accepted if prefix lengths are different
### Deployment Type

Self-hosted

### NetBox Version

v4.0.11

### Python Version

3.12

### Steps to Reproduce

1. Create an IPRange with start 1.2.3.100/24 and end 1.2.3.200/24
2. Create an IPRange with start 1.2.3.123/26 and end 1.2.3.124/26

### Expected Behavior

The second IPRange should be rejected, as it overlaps with the first. (There is code in `IPRange.clean()` which is clearly intended to do this)

### Observed Behavior

The overlapping range is accepted.

The problem occurs because the Django ORM lte/gte filter for overlapping ranges only works if the prefix lengths match:

```
>>> ip = "1.2.3.101/24"
>>> IPRange.objects.filter(start_address__lte=ip, end_address__gte=ip, vrf=None)
<RestrictedQuerySet [<IPRange: 1.2.3.100-200/24>]>
>>> ip = "1.2.3.101/26"
>>> IPRange.objects.filter(start_address__lte=ip, end_address__gte=ip, vrf=None)
<RestrictedQuerySet []>
```

I cannot see an obvious solution here, apart from brute force iterating over all possible prefix lengths:

```
>>> [IPRange.objects.filter(start_address__lte=f"1.2.3.101/{pl}", end_address__gte=f"1.2.3.101/{pl}", vrf=None).count() for pl in range(33)]
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0]
```

Alternatively, changing the data model so that a range uses IPAddress instead of IPNetwork for the start/end addresses would eliminate this problem.

bctiemann:
### Fixes: #17358

Adds new Lookup classes for `lt/gt/lte/gte` comparisons with INET objects, ignoring the mask. This enables correct comparison of adjacent IPRanges when looking for overlaps, and prevents acceptance of an invalid new IPRange where the entered start_address and end_address masks do not match the existing ranges.

### DIFF
--- a/netbox/ipam/lookups.py
+++ b/netbox/ipam/lookups.py
@@ -207,6 +207,11 @@ class Host(Transform):
     lookup_name = 'host'
 
 
+class HostAsInet(Transform):
+    lookup_name = 'host_as_inet'
+    template = 'CAST(HOST( %(expressions)s ) AS INET)'
+
+
 class Inet(Transform):
     function = 'INET'
     lookup_name = 'inet'
--- a/netbox/ipam/lookups.py
+++ b/netbox/ipam/lookups.py
@@ -154,44 +154,49 @@ class NetHostContained(Lookup):
         return 'CAST(HOST(%s) AS INET) <<= %s' % (lhs, rhs), params
 
 
-class NetHostGreaterThan(Lookup):
+class NetHostComparison(Lookup):
+
+    @property
+    def comparison_sql(self):
+        raise NotImplementedError
+
+    def as_sql(self, qn, connection):
+        lhs, lhs_params = self.process_lhs(qn, connection)
+        rhs, rhs_params = self.process_rhs(qn, connection)
+        params = lhs_params + rhs_params
+        return self.comparison_sql % (lhs, rhs), params
+
+
+class NetHostGreaterThan(NetHostComparison):
     lookup_name = 'net_host_gt'
 
-    def as_sql(self, qn, connection):
-        lhs, lhs_params = self.process_lhs(qn, connection)
-        rhs, rhs_params = self.process_rhs(qn, connection)
-        params = lhs_params + rhs_params
-        return 'CAST(HOST(%s) AS INET) > INET %s' % (lhs, rhs), params
+    @property
+    def comparison_sql(self):
+        return 'CAST(HOST(%s) AS INET) > INET %s'
 
 
-class NetHostLessThan(Lookup):
+class NetHostLessThan(NetHostComparison):
     lookup_name = 'net_host_lt'
 
-    def as_sql(self, qn, connection):
-        lhs, lhs_params = self.process_lhs(qn, connection)
-        rhs, rhs_params = self.process_rhs(qn, connection)
-        params = lhs_params + rhs_params
-        return 'CAST(HOST(%s) AS INET) < INET %s' % (lhs, rhs), params
+    @property
+    def comparison_sql(self):
+        return 'CAST(HOST(%s) AS INET) < INET %s'
 
 
-class NetHostGreaterThanOrEqual(Lookup):
+class NetHostGreaterThanOrEqual(NetHostComparison):
     lookup_name = 'net_host_gte'
 
-    def as_sql(self, qn, connection):
-        lhs, lhs_params = self.process_lhs(qn, connection)
-        rhs, rhs_params = self.process_rhs(qn, connection)
-        params = lhs_params + rhs_params
-        return 'CAST(HOST(%s) AS INET) >= INET %s' % (lhs, rhs), params
+    @property
+    def comparison_sql(self):
+        return 'CAST(HOST(%s) AS INET) >= INET %s'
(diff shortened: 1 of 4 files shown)


netbox/ipam/lookups.py
```diff
@@ -154,6 +154,46 @@ def as_sql(self, qn, connection):
         return 'CAST(HOST(%s) AS INET) <<= %s' % (lhs, rhs), params
 
 
+class NetHostGreaterThan(Lookup):
```
arthanson, line 157: Would it make sense to DRY the code here as the CAST looks like the only difference, so move:
```
         lhs, lhs_params = self.process_lhs(qn, connection)
         rhs, rhs_params = self.process_rhs(qn, connection)
         params = lhs_params + rhs_params
```
to a get_params call?

arthanson (approved):
Seems good, just wondering if it makes sense to make the code a bit more DRY.

netbox/ipam/lookups.py
bctiemann: Yeah, it's a tricky question because a lot of the existing `Lookup` subclasses in this file all use the same `process_lhs` and `process_rhs` library calls, but they do different things with them where these four new ones all just assign them to `params` and then use `lhs` and `rhs` with it in the query builder. I agree it's repetitive but I would be wanting to generalize this logic in some way (i.e. in a mixin) that could be applied to the existing lookups as well but I'm not sure I see an obvious way to do that.

Maybe a more sensible way would be to refactor these into four subclasses of a base class where the part that differs is just the actual final line of `as_sql`. Let me try that...

jeremystretch (changes requested):
Rather than introduce new lookups for these specific comparisons, we should be able to leverage Postgres' `HOST()` function. Unfortunately, the [`NetHost()` lookup](https://github.com/netbox-community/netbox/blob/5f32b23c3533060c310cbd75015705f69825e739/netbox/ipam/lookups.py#L89) won't support this, but we can instead register the [`Host()` transform](https://github.com/netbox-community/netbox/blob/5f32b23c3533060c310cbd75015705f69825e739/netbox/ipam/lookups.py#L175) on `IPAddressField`:

```python
>>> from ipam.fields import IPAddressField
>>> from ipam import lookups
>>> IPAddressField.register_lookup(lookups.Host)
<class 'ipam.lookups.Host'>
>>> IPRange.objects.filter(start_address__host__gte="1.2.3.100", start_address__host__lte="1.2.3.200")
<RestrictedQuerySet [<IPRange: 1.2.3.123-124/26>]>
```

(Ultimately, it probably makes sense to retire some of these lookups in favor of the transforms, but that's a separate discussion.)

bctiemann:
I don't think that works actually; the `Host` transform only casts the input value to a `HOST` (i.e. `HOST(%s)`), but in order to compare to a bare IP address on the rhs you have to cast it to an `INET` e.g. `CAST(HOST(%s) AS INET)`. I don't think `Transform` supports that kind of structure unless I'm missing something.

At any rate the comparisons in my unit test fail and it throws an overlapping exception when trying to add `1.2.3.100/24-1.2.3.199/24` (it says it overlaps with `1.2.3.1-99/24`).

That's why I created the four specific comparison lookups in the first place, because using `HOST` on its own leads to incorrect comparisons.

... (30 minutes later) ...

After some more digging into the `Transform` class, I've added a `HostAsInet` Transform that does what we would need for this lookup using the `start_address__host_as_inet__gte` lookup syntax. If you like this we can use it instead of the four specific lookups I did earlier (and remove those).

netbox/ipam/lookups.py
```diff
@@ -177,6 +177,11 @@ class Host(Transform):
     lookup_name = 'host'
 
 
+class HostAsInet(Transform):
```
jeremystretch, line 180: This feels redundant to the existing `Host` and `Inet` transforms. Is there a way we can utilize those instead of introducing a new transform? We could register these transforms for IPAddressField:

```python
IPAddressField.register_lookup(lookups.Host)
IPAddressField.register_lookup(lookups.Inet)
```

This allows us to chain lookups:

```
>>> IPRange.objects.filter(start_address__host__inet__lte='192.168.1.50')
<RestrictedQuerySet [<IPRange: 192.168.1.0-49/22>, <IPRange: 192.168.1.50-99/22>]>
```

jeremystretch (changes requested):

netbox/ipam/lookups.py
bctiemann: Agreed, this does seem to work fine (unit tests pass).

jeremystretch (approved):

PR #17423, opened 2024-09-09
17422 update custom field group display tag

### LINKED ISSUE #17422: custom field group label too small
### Deployment Type

Self-hosted

### NetBox Version

v.4.1.0

### Python Version

3.10

### Steps to Reproduce

Create 2 custom fields, assign them to a field group.
Go to an object's edit page where the custom fields have been applied to.

### Expected Behavior

The custom field's group 'label' should be almost the same size as the 'custom fields' font.

image

### Observed Behavior

custom field label is quite small.

custom field screen:
image

object edit screen:
image

arthanson:
### Fixes: #17422

Was set at h6 tag, but the "Custom Fields" header above is h2, so making it h3 as that is correct hierarchical wise and also displays at the correct size.

Add a new site | NetBox 2024-09-10 09-46-55

### DIFF
--- a/netbox/utilities/templates/form_helpers/render_custom_fields.html
+++ b/netbox/utilities/templates/form_helpers/render_custom_fields.html
@@ -3,7 +3,7 @@
 {% for group, fields in form.custom_field_groups.items %}
   {% if group %}
   <div class="row">
-    <h6 class="offset-sm-3 mb-3">{{ group }}</h6>
+    <h3 class="offset-sm-3 mb-3">{{ group }}</h3>
   </div>
   {% endif %}
   {% for name in fields %}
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
The custom field group heading should be visually distinct from the form field group. Currently they are both rendered as 16px text. I suggest using one of Bootstrap's [heading classes](https://getbootstrap.com/docs/5.3/content/typography/#headings) to display the `<h3>` as if it were an `<h4>` or `<h5>`.

Additionally, the `<h3>` element should have the `col-9` & `offset-3` classes to match the form group heading.

jeremystretch (approved):

PR #17456, opened 2024-09-11
Fixes: #17437 - Fix exception when saving interface-template with bridged interface

### LINKED ISSUE #17437: Error when making an interface bridge on an intarface template of module type
### Deployment Type

Self-hosted

### NetBox Version

v4.1.0

### Python Version

3.12

### Steps to Reproduce

1. On {host}/dcim/interface-templates/{id}/edit/?return_url=/dcim/module-types/{module_type_id}/interfaces/ the bridge interface is not filtering the interfaces of the {module_type_id} module type.
2. And when you select the correct interface, you get an error after clicking on save button.

### Expected Behavior

To filter the interfaces of the correct module type and to save without any error.

### Observed Behavior

There is the reported error.

<class 'django.core.exceptions.ImproperlyConfigured'>

Could not resolve URL for hyperlinked relationship using view name "". You may have failed to include the related model in your API, or incorrectly configured the `lookup_field` attribute on this field.

Python version: 3.12.3
NetBox version: 4.1.0
Plugins: None installed

### LINKED ISSUE #17452: "No reverse for interfacetemplate" when saving a bridged interface
### Deployment Type

Self-hosted

### NetBox Version

v4.1.0

### Python Version

3.12

### Steps to Reproduce

1. Create Manufacturer (NetScout)
2. Create Device Type (AED)
3. Create Interfaces (int & ext)
4. Edit ext, set bridge to int
5. Save

### Expected Behavior

Save successfully

### Observed Behavior

Outer exception:

```text
ImproperlyConfigured at /dcim/interface-templates/302/edit/
Could not resolve URL for hyperlinked relationship using view name "". You may have failed to include the related model in your API, or incorrectly configured the `lookup_field` attribute on this field.
```

Inner exception:

```
Reverse for 'interfacetemplate' not found. 'interfacetemplate' is not a valid view function or pattern name.
```

DanSheps:
### Fixes: #17437 - Fix exception when saving interface-template with bridged interface

* Return None if NoReverseMatch found from self.reverse with BaseNetBoxHyperlinkedIdentityField.get_url

### DIFF
--- a/netbox/netbox/api/serializers/fields.py
+++ b/netbox/netbox/api/serializers/fields.py
@@ -1,3 +1,4 @@
+from django.urls import NoReverseMatch
 from rest_framework import serializers
 
 __all__ = (
@@ -33,7 +34,11 @@ class BaseNetBoxHyperlinkedIdentityField(serializers.HyperlinkedIdentityField):
         model_name = self.parent.Meta.model._meta.model_name
         app_name = self.parent.Meta.model._meta.app_label
         view_name = self.get_view_name(app_name, model_name)
-        return self.reverse(view_name, kwargs=kwargs, request=request, format=format)
+
+        try:
+            return self.reverse(view_name, kwargs=kwargs, request=request, format=format)
+        except NoReverseMatch:
+            return None
 
     def get_view_name(self, app_name, model_name):
         raise NotImplementedError(_('{class_name} must implement get_view_name()').format(


jeremystretch:
`BaseNetBoxHyperlinkedIdentityField` should always return a valid URL. If it doesn't, that's an indication that it shouldn't be included on the serializer. `display_url` needs to be removed from NestedInterfaceTemplateSerializer. We also need to fix the filter names set on the `bridge` form field. I've submitted alternate PR #17470 to address both changes.

PR #17458, opened 2024-09-11
17444 fix event rules jobs enqueue

### LINKED ISSUE #17444: Event Rules not creating jobs
### Deployment Type

Self-hosted

### NetBox Version

v4.1.0

### Python Version

3.12

### Steps to Reproduce

1. Create an Event Rule to e.g. trigger a script when a device is created
2. Create a device

### Expected Behavior

Job should be created running the script

### Observed Behavior

No job is created. (Docker Container) console output:
`Cannot import events pipeline extras.events.process_event_queue error: core.models.jobs.Job.enqueue() got multiple values for keyword argument 'name'`

arthanson:
### Fixes: #17444 

Scripts pass in a name param in extras.events.process_event_rules to the enqueue function resulting in duplicate name params as there will be a name in kwargs.

### DIFF
--- a/netbox/netbox/jobs.py
+++ b/netbox/netbox/jobs.py
@@ -100,9 +100,7 @@ class JobRunner(ABC):
         This method is a wrapper of `Job.enqueue()` using `handle()` as function callback. See its documentation for
         parameters.
         """
-        name = kwargs.pop('name', None)
-        if not name:
-            name = cls.name
+        name = kwargs.pop('name', None) or cls.name
 
         return Job.enqueue(cls.handle, name=name, *args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


netbox/netbox/jobs.py
```diff
@@ -100,7 +100,11 @@ def enqueue(cls, *args, **kwargs):
         This method is a wrapper of `Job.enqueue()` using `handle()` as function callback. See its documentation for
         parameters.
         """
-        return Job.enqueue(cls.handle, name=cls.name, *args, **kwargs)
+        name = kwargs.pop('name', None)
```
bctiemann, line 103: Maybe just
`name = kwargs.pop('name') or cls.name`
?

bctiemann (approved):

PR #17493, opened 2024-09-13
Fixes: #17492 - Changes Serializer field to prevent exceptions when plugins inherit the base serializer

### LINKED ISSUE #17492: When inheriting NetBoxModelSerializer without defining an override for the url attribute, an exception is thrown
### Deployment Type

Self-hosted

### NetBox Version

4.1.1

### Python Version

3.12

### Steps to Reproduce

1. Create a plugin (per the plugin docs)
2. Add plugin to NetBox
3. View plugin's API where `url` is defined as a field but not overridden

### Expected Behavior

`url` will have a proper link

### Observed Behavior

Exception: ImproperlyConfigured at <x>

DanSheps:
Fixes: #17492 - Change methods on NetBox*HyperlinkedIdentityField to use get_viewname

* Adjust method to pass model instead of app_name and model_name
* Adjust methods to use get_viewname

### DIFF
--- a/netbox/netbox/api/serializers/fields.py
+++ b/netbox/netbox/api/serializers/fields.py
@@ -1,4 +1,7 @@
+from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
+
+from utilities.views import get_viewname
 
 __all__ = (
     'NetBoxAPIHyperlinkedIdentityField',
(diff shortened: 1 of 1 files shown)


netbox/netbox/api/serializers/fields.py
```diff
@@ -1,5 +1,8 @@
+from django.utils.translation import gettext_lazy as _
```
DanSheps, line 1: Just FYI, this is to fix an error my IDE was showing that we were using `_` without importing it

jeremystretch (approved):
Thanks @DanSheps!

PR #17523, opened 2024-09-17
Closes #16009 - Added styling to form templates to enable floating button groups

### LINKED ISSUE #16009: Add additional `save / create / cancel` buttons to top of forms (or make existing buttons stay visible)
### NetBox version

v4.0.0

### Feature type

New functionality

### Proposed functionality

Add `save / create / cancel` buttons to top of forms (or make existing buttons stay visible )

### Use case

when a model has a lot of fields/custom fields (or is being used on a small display) there can be a lot of scrolling to get to the save / create buttons. 
it would be great if there was a copy of those buttons at the top of the form or if the existing buttons were always visible on the form

### Database changes

N/A

### External dependencies

N/A

atownson:
### Fixes: #16009

Created a new style class, `btn-float-group` , and added to object_edit, bulk_edit, and filter_list templates.


netbox/project-static/styles/custom/_misc.scss
```diff
@@ -33,3 +33,11 @@ span.color-label {
 .netbox-edition {
   letter-spacing: .15rem;
 }
+
+// A floating div for form buttons
+.btn-float-group {
+  position: sticky;
+  bottom: 0px;
+  z-index: 1;
+  padding: 2px 0px 10px 0px;
```
jeremystretch, line 42: We may want to add a bit more top padding. IMO vertical spacing should be the same above and below the buttons while floating.

netbox/project-static/styles/overrides/_tabler.scss
```diff
@@ -52,7 +52,7 @@ pre {
 }
 
 // Altering background colors
-.page, .page-tabs .nav-tabs .nav-link.active {
+.page, .page-tabs .nav-tabs .nav-link.active, .btn-float-group {
```
jeremystretch, line 55: We want to avoid declaring any custom classes here, as this module should be restricted to overrides for Tabler itself. Maybe we can inherit the background color from `.page`?

netbox/templates/inc/filter_list.html
```diff
@@ -37,7 +37,7 @@
         </div>
       {% endif %}
     </div>
-    <div class="card-footer text-end d-print-none border-0">
```
jeremystretch, line 40: We may want to move these buttons outside of the card for consistency with other floating button groups. (Also, the background color is off.)

screenshot

jeremystretch (changes requested):
Nice work @atownson! We should "float' the bulk operation buttons for object list views too.

netbox/project-static/styles/overrides/_tabler.scss
atownson: This has been moved to custom/_misc.scss. Inheriting from `.page` proved to be problematic as the class inherited many undesirable attributes.

netbox/templates/inc/filter_list.html
atownson: This is done

netbox/project-static/styles/custom/_misc.scss
atownson: I went a different direction here. Let me know what you think. Essentially I removed the button group `div` background altogether, which created truly 'floating' buttons and IMO added more real estate to the form by not covering it up.
 
NetBox Button Illustration Rev

atownson:
Another thought on the list form floating buttons: What if the buttons didn't float until the user checked an item in the list, then the buttons floated? This would indicate clear intention to use the bulk buttons and provide easy accessibility to them.

atownson:
I went ahead and added the conditional floating for object list forms.
NetBox Button Illustration Rev 3

jeremystretch (approved):
Thanks @atownson!

PR #17524, opened 2024-09-17
Closes: #15239 - Allow adding/removing tagged VLANs in bulk editing of Interfaces

### LINKED ISSUE #15239: Allow bulk assignment/unassignment of VLANs to interfaces
### NetBox version

v3.7.0

### Feature type

### Proposed functionality

When bulk editing interfaces, I would like to be able to _add_ and _remove_ assignment of tagged VLANs, rather than explicitly _setting_ the list of VLANs.

This would make VLAN editing more in line with the tagging workflow.

### Use case

Adding a VLAN to a large number of ports (in my case adding a VLAN to several hundred hypervisor uplink interfaces) it would be preferable to not have to list the existing VLAN assignments when adding a new one.

### Database changes

None

### External dependencies

None

bctiemann:
### Closes: #15239

Adds "Add Tagged VLANs" and "Remove tagged VLANs" widgets to the bulk edit form of Interfaces, both when editing them standalone and when accessing them as components of a device. These fields operate in the same way as the existing "Add tags"/"Remove tags" and allow the user to add or remove the selected items to all selected objects regardless of their existing VLANs, rather than having to set all objects to the same VLAN settings in the edit operation.

Note that this PR also cleans up the `__init__` function of `InterfaceBulkEditForm`, removing the after-the-fact modification of the widgets to add query parameters, as these params can be set in the field definitions directly using the `$device` token.

Screenshot 2024-09-17 at 4 01 39 PM


netbox/netbox/views/generic/bulk_views.py
```diff
@@ -615,6 +615,12 @@ def _update_objects(self, form, request):
             if form.cleaned_data.get('remove_tags', None):
                 obj.tags.remove(*form.cleaned_data['remove_tags'])
 
+            # Add/remove tagged VLANs
+            if form.cleaned_data.get('add_tagged_vlans', None):
+                obj.tagged_vlans.add(*form.cleaned_data['add_tagged_vlans'])
+            if form.cleaned_data.get('remove_tagged_vlans', None):
+                obj.tagged_vlans.remove(*form.cleaned_data['remove_tagged_vlans'])
+
```
jeremystretch, line 623: Model-specific fields can't go into a generic view. We'll need to come up with a pattern for handling these generically, probably in a manner similar to how we handle many-to-many assignments.

jeremystretch (changes requested):
Nice work with this @bctiemann, we just need to figure out a pattern for accommodating these non-model many-to-many fields generically.

netbox/netbox/views/generic/bulk_views.py
bctiemann: Good point, how about this?

netbox/dcim/views.py
```diff
@@ -2616,6 +2616,13 @@ class InterfaceBulkEditView(generic.BulkEditView):
     table = tables.InterfaceTable
     form = forms.InterfaceBulkEditForm
 
+    def extra_object_field_operations(self, form, obj):
+        # Add/remove tagged VLANs
+        if form.cleaned_data.get('add_tagged_vlans', None):
+            obj.tagged_vlans.add(*form.cleaned_data['add_tagged_vlans'])
+        if form.cleaned_data.get('remove_tagged_vlans', None):
+            obj.tagged_vlans.remove(*form.cleaned_data['remove_tagged_vlans'])
+
```
DanSheps, line 2625: We probably want to validate the interface mode for these interfaces when bulk editing, to prevent #15924 and similar.

DanSheps: Yeah, I think for this it is a valid approach.

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -541,6 +541,9 @@ class BulkEditView(GetReturnURLMixin, BaseMultiObjectView):
     def get_required_permission(self):
         return get_permission_for_model(self.queryset.model, 'change')
 
+    def extra_object_field_operations(self, form, obj):
```
arthanson, line 544: Can you add a doc string as to what this is for, for future reference.

bctiemann: Good call; done.

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -541,6 +541,13 @@ class BulkEditView(GetReturnURLMixin, BaseMultiObjectView):
...
```
jeremystretch, line 544: I'd suggest renaming this to something like `post_save_operations()`, and moving the M2M and add/remove tag logic that currently lived near the bottom of `_update_objects()` here. That way all the post-save behavior becomes completely accessible.

jeremystretch (changes requested):

jeremystretch:
CI failure is unrelated; see #17950.

jeremystretch (approved):
Nice work @bctiemann!

PR #17525, opened 2024-09-17
Fixes: #17083 - Wrap labels in a div to reduce inadvertently clickable area to their left in forms

### LINKED ISSUE #17083: Clickable area of input fields extends too far left
### Deployment Type

NetBox Cloud

### NetBox Version

v4.0.8

### Python Version

3.10

### Steps to Reproduce

(Previous issue #16302 was closed, but I am pretty certain that this is in fact a bug, not an intended feature.)

This seems to be the case for all objects in Netbox, both in the Create New screen and in the Edit screen. I will use an IP address for this example.

Click on "Add new IP address"
Left click to the left of the name of any field (see the red X in the image below)
image

The field will now act as if it was clicked, even though it obviously was not clicked
There seems to be a limit to how far left you can go before this stops occuring.

### Expected Behavior

When clicking outside of the field, the field should not act as if it was clicked.
Clicking on the red X in the image should do nothing (and close any open drop-downs).

### Observed Behavior

The field acts as if it was clicked, and the drop down opens up.

bctiemann:
### Fixes: #17083

Wraps all form labels in a `<div class="col-sm-3">` which ensures ample dead clickable space to their left in forms, while remaining responsive.

Screenshot 2024-09-17 at 4 15 24 PM

### DIFF
--- a/netbox/utilities/templates/form_helpers/render_field.html
+++ b/netbox/utilities/templates/form_helpers/render_field.html
@@ -6,9 +6,11 @@
 
   {# Render the field label (if any), except for checkboxes #}
   {% if label and not field|widget_type == 'checkboxinput' %}
-    <label for="{{ field.id_for_label }}" class="col-sm-3 col-form-label text-lg-end{% if field.field.required %} required{% endif %}">
-      {{ label }}
-    </label>
+    <div class="col-sm-3">
+      <label for="{{ field.id_for_label }}" class="col-form-label text-lg-end{% if field.field.required %} required{% endif %}">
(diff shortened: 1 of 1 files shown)


DanSheps (approved):

jeremystretch:
I don't see any difference in behavior with this change applied: Clicking anywhere within the column to the left of the field still selects the label as before.

DanSheps (approved):
Showing label as properly constrained now

jeremystretch (approved):

PR #17551, opened 2024-09-19
create 802.15.4 interface type

### LINKED ISSUE #17550: Add 802.15.4 Wireless Interface
### NetBox version

v4.1.1

### Feature type

### Proposed functionality

Netbox currently doesn't have an option for [Zigbee](https://en.wikipedia.org/wiki/Zigbee) other [IEEE 802.15.4](https://en.wikipedia.org/wiki/IEEE_802.15.4) based RF interfaces. 

I would like to try my hand at adding this functionality.

### Use case

Outside of Wi-Fi versions, only Bluetooth and "Other" exist. The 802.15.4 stack is becoming more and more relevant and I would like to see it as an option for my Enterprise Access Points. It is already a feature that is built in to current generation Enterprise Access Points.

### Database changes

I don't believe that this will require any database changes, just an extension of the existing standards already in place.

### External dependencies

none

iamk3:
### Fixes: #17550 

Adds [IEEE 802.15.4](https://en.wikipedia.org/wiki/IEEE_802.15.4) (LR-WPAN --> basis for Zigbee, Thread, Matter, etc.) to the list of wireless interface types.

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -924,6 +924,7 @@ class InterfaceTypeChoices(ChoiceSet):
     TYPE_80211AY = 'ieee802.11ay'
     TYPE_80211BE = 'ieee802.11be'
     TYPE_802151 = 'ieee802.15.1'
+    TYPE_802154 = 'ieee802.15.4'
     TYPE_OTHER_WIRELESS = 'other-wireless'
 
     # Cellular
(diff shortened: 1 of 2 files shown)


iamk3:
@DanSheps Out of curiosity, when are pull requests typically reviewed for merger? I'd love to be able to use this interface type as soon as I can.

jeremystretch:
@iamk3 like most projects, we do not provide any guarantee or ETA regarding PR review. If you have an urgent need, you are always free to make the change locally in anticipation of it being accepted upstream.

jeremystretch (approved):

iamk3:
I definitely understand. Thank you @jeremystretch!

PR #17553, opened 2024-09-19
17419 fix MPTT issue with migrations for nested module bays

### LINKED ISSUE #17419: MPTT issues with using nested modules on pre v4.1 data
### Deployment Type

Self-hosted

### NetBox Version

v4.1.0

### Python Version

3.12

### Steps to Reproduce

1. Navigate to a device with module bays that existed before v4.1
2. Set a parent on a module bay
   a. Edit an existing module bay and set the parent to an existing module; OR
   b. Add a new module bay with the parent set to an existing module

This has been verified on local systems that have been upgraded to v4.1, but you can also see an example of this happening on the demo site if you add a new module bay "Subslot 0/0" to module "Slot 0" on device "ncsu-coreswitch1". I understand that the demo site cannot be relied on and the example is purely for reference using data included in the standard database refresh.

### Expected Behavior

The module bay appears in the list with the correct depth under the module bay for the parent module.

### Observed Behavior

When the parent is changed on an existing module bay, the error "A node may not be made a child of any of its descendants." is presented. When a new module bay is added, the new module bay has the correct depth, but does not appear in the correct location in the module bay list. 

A new module bay being added to a parent module in a new module bay appears to exhibit the expected behavior.

This looks like classic symptoms of MPTT being out of sync. When RackGroups were made nestable back in v2.8, the following migration was included:
https://github.com/netbox-community/netbox/blob/v2.8.0/netbox/dcim/migrations/0102_nested_rackgroups_rebuild.py. I do not see a similar migration accompanying https://github.com/netbox-community/netbox/blob/v4.1.0/netbox/dcim/migrations/0190_nested_modules.py.

arthanson:
### Fixes: #17419 

Note in the migration had to use the actual model and not `apps.get_model` as the manager for rebuild is needed.

### DIFF
--- a/netbox/dcim/migrations/0191_module_bay_rebuild.py
+++ b/netbox/dcim/migrations/0191_module_bay_rebuild.py
@@ -2,7 +2,7 @@ from django.db import migrations
 
 
 def rebuild_mptt(apps, schema_editor):
-    ModuleBay = apps.get_model('dcim', 'ModuleBay')
+    from dcim.models import ModuleBay
     ModuleBay.objects.rebuild()
 
 
--- a/netbox/dcim/migrations/0191_module_bay_rebuild.py
+++ b/netbox/dcim/migrations/0191_module_bay_rebuild.py
@@ -1,0 +1,20 @@
+from django.db import migrations
+
+
+def rebuild_mptt(apps, schema_editor):
+    ModuleBay = apps.get_model('dcim', 'ModuleBay')
+    ModuleBay.objects.rebuild()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dcim', '0190_nested_modules'),
+    ]
+
+    operations = [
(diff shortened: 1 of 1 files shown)


bctiemann (approved):

jeremystretch (commented):
Are we sure it's safe to import the model directly within a migration? IIRC, in the past we've had to [jump through some hoops](https://github.com/netbox-community/netbox/blob/v2.11.12/netbox/dcim/migrations/0119_inventoryitem_mptt_rebuild.py) to avoid it.

bctiemann:
Oh right, good point, the usual pattern is to do like `EventRule = apps.get_model("extras", "EventRule")`, because otherwise the model state can't be guaranteed in a later code state. I should have caught that

bctiemann (approved):

jeremystretch (approved):

PR #17576, opened 2024-09-23
Closes: #17575 - Add NullableMultipleChoiceFilter and filter form UI to match empty string

### LINKED ISSUE #17575: Add ability to filter for empty string values in an "OR" fashion
### NetBox version

v4.1.0

### Feature type

New functionality

### Proposed functionality

Currently we can filter a field on an empty string value using (for example) `type__empty=true`, using functionality present in `BaseFilterSet.get_additional_lookups`. However, this construct only allows us to apply filters with different keys in an "AND" fashion, i.e. `type=foo` and `type__empty=true` are treated as separate filter conditions that both must be true. Also this type of filter must be entered into the URL manually, it is not available in the standard FilterForm classes.

Proposal is to create a `NullableMultipleChoiceFilter` (patterned after the existing `NullableCharFieldFilter`) which treats `''` as one of several possible values for the field, e.g. `['foo', 'bar', '']`, any of which returns a result in the filter. This would also make it possible to add an "(unset)" option in the choices presented in a FilterForm, such that the user can select all the applicable values from the dropdown and search them all at once.

### Use case

It is cumbersome and obscure to use the bespoke lookups we currently support, such as `type__empty`, and it lacks flexibility. With this change, a user might select 2 or 3 different values in a multiple-choice filter input in the FilterForm, including the "empty" (unset) value (which would follow the system configurable `settings.FILTERS_NULL_CHOICE_VALUE`).

Note that this change is narrowly targeted at matching the empty string `''`; it does not address any similar treatment of other programmatically generated additional lookups (such as `__gt`, `__iexact`, etc). For that reason this might be a low-priority use case or even contribute to a pattern inconsistency. However this also might be thought of as a building block to more rich filter inputs of a similar kind later.

### Database changes

N/A

### External dependencies

N/A

bctiemann:
### Closes: #17575

Adds a `NullableMultipleChoiceFilter` class (patterned on the existing `NullableCharFieldFilter`) which can be used in a multiple-choice FilterForm input to add the "(unset)" (`''`) value to a list of acceptable choices in the filter, evaluated in an OR fashion in a single query.

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -18,7 +18,7 @@ from tenancy.models import *
 from users.models import User
 from utilities.filters import (
     ContentTypeFilter, MultiValueCharFilter, MultiValueMACAddressFilter, MultiValueNumberFilter, MultiValueWWNFilter,
-    NumericArrayFilter, TreeNodeMultipleChoiceFilter, EmptyStringMultipleChoiceFilter,
+    NumericArrayFilter, TreeNodeMultipleChoiceFilter, NullableMultipleChoiceFilter,
 )
 from virtualization.models import Cluster, ClusterGroup
 from vpn.models import L2VPN
@@ -1980,7 +1980,7 @@ class CableFilterSet(TenancyFilterSet, NetBoxModelFilterSet):
         method='_unterminated',
         label=_('Unterminated'),
     )
-    type = EmptyStringMultipleChoiceFilter(
+    type = NullableMultipleChoiceFilter(
         choices=CableTypeChoices
     )
     status = django_filters.MultipleChoiceFilter(
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -18,7 +18,7 @@ from tenancy.models import *
 from users.models import User
 from utilities.filters import (
     ContentTypeFilter, MultiValueCharFilter, MultiValueMACAddressFilter, MultiValueNumberFilter, MultiValueWWNFilter,
-    NumericArrayFilter, TreeNodeMultipleChoiceFilter, EmptyStringFilter,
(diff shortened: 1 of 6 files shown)


netbox/dcim/filtersets.py
```diff
@@ -1980,7 +1980,7 @@ class CableFilterSet(TenancyFilterSet, NetBoxModelFilterSet):
         method='_unterminated',
         label=_('Unterminated'),
     )
-    type = django_filters.MultipleChoiceFilter(
+    type = NullableMultipleChoiceFilter(
```
jeremystretch, line 1983: This addresses the `type` field on Cable specifically, but I think what we need is a general implementation which covers _all_ non-required choice fields, across all models. Ideally, this should be handled automatically for all model fields with a `choices` attribute defined, which would obviate the need to manually declare these on the FilterSets as we currently do. (Granted, I'm not sure whether that's feasible.)

netbox/utilities/filters.py
```diff
@@ -143,6 +144,16 @@ def filter(self, qs, value):
...
 
 
+class NullableMultipleChoiceFilter(django_filters.MultipleChoiceFilter):
+    """
+    Similar to NullableCharFieldFilter, but allows multiple values including the special NULL string.
+    """
+    def filter(self, qs, value):
+        if settings.FILTERS_NULL_CHOICE_VALUE in value:
+            value.append('')
+        return super().filter(qs, value)
```
jeremystretch, line 154: It feels like we need to further consider the implications of storing empty strings rather than null values for these fields. django-filter has native support for the later; replacing empty strings with `null` would obviate the need for a custom filter entirely AFAICT. Maybe that should be our preferred strategy.

jeremystretch (changes requested):

bctiemann:
Closing as this is no longer needed due to https://github.com/netbox-community/netbox/issues/17761.

PR #17579, opened 2024-09-23
Fixes: #17126 - Respect the weight unit of the DeviceType when displaying the Device detail

### LINKED ISSUE #17126: Weight unit between 'device type' and 'device' is not consistent
### Deployment Type

Self-hosted

### NetBox Version

v.4.0.8

### Python Version

3.10

### Steps to Reproduce

Create a device type
Set device type 'weight' to a value
Set device type 'weight unit' to 'pounds'
Create a new device based on the device type

image

### Expected Behavior

Device weight should match the weight unit from device type

### Observed Behavior

Device weight unit defaults to kilograms

image

bctiemann:
### Fixes: #17126

Changes `device.html` to conditionally convert a device's `total_weight` to pounds if the DeviceType defines the weight in pounds in the `weight_unit`.

Note that this compares the `weight_unit` value to the literal string `'lb'` or `'kg'`, rather than to an enum value in the `WeightUnitChoiceSet`. It would be more DRY to pass the enum to the template context, but less elegant as the data construct does not seem to allow direct access to class attrs in the template, meaning the enum would need to be transformed or deconstructed to pass into the template properly.

### DIFF
--- a/netbox/templates/dcim/device.html
+++ b/netbox/templates/dcim/device.html
@@ -324,7 +324,11 @@
                         <th scope="row">{% trans "Weight" %}</th>
                         <td>
                             {% if object.total_weight %}
-                                {{ object.total_weight|floatformat }} {% trans "Kilograms" %}
+                                {% if object.device_type.weight_unit == "lb" %}
+                                    {{ object.total_weight|kg_to_pounds|floatformat }} {% trans "Pounds" %}
+                                {% elif object.device_type.weight_unit == "kg" %}
+                                    {{ object.total_weight|floatformat }} {% trans "Kilograms" %}
+                                {% endif %}
                             {% else %}
                                 {{ ''|placeholder }}
                             {% endif %}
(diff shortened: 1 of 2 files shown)


arthanson (changes requested):
Might be cleaner - You should be able to modify total_weight in DeviceModel to return the correct converted value, It looks like it is only used in this template.  I think you could then add another property to get the weight unit which gets the value from WeightUnitChoices.

It looks like there is also grams and ounces possible though I'm not sure they would ever be used.

bctiemann:
Right, I saw there is `converters.py` which only has a couple of high-level one-way methods, and I noticed that we had templatetags for some richer and more fine-grained conversions so I opted for that.

I too would prefer to put it in a `@property` or something similar on the model but that might involve adding or refactoring a whole bunch of the conversion logic to make it more centralizable and not confined to templates.

bctiemann:
@arthanson we have this in `rack.html`:

```
            <td>
              {{ object.total_weight|floatformat }} {% trans "Kilograms" %}
              ({{ object.total_weight|kg_to_pounds|floatformat }} {% trans "Pounds" %})
            </td>
```

I've reworked it to follow this pattern, and also fixed `rack.html` to show the placeholder if the total rack weight is 0. It's redundant now but at least consistent.

jeremystretch (approved):

PR #17618, opened 2024-09-26
Fix route target naming in german

patrick7:
Updated German translation: changed "Route Ziel" back to "Route Target" to preserve the technical term.


patrick7:
Thanks. Since they want an extra account, I'll be out. 
If anyone wants to take it up, feel free!

PR #17622, opened 2024-09-26
Fixes: #17611: Fix a ValidationError message

### LINKED ISSUE #17611: Integer Custom Field shows maximum value when minimum value fails validation
### Deployment Type

Self-hosted

### NetBox Version

version-agnostic, seeing it on latest and develop

### Python Version

3.11

### Steps to Reproduce

1. Create a custom field of type Integer, with a minimum value, say **2** and a maximum value, say **1000**. Attach it in one or more Object Types, doesn't matter, just need one.
2.  Go to that Object and try to save a value for the custom_field that is less than the minimum, say, **1**. This will fail as expected, as it's less than the minimum expected.
3. The raised Validation Exception if you do so via the API will be something like:
```
RequestError: The request failed with code 400 Bad Request: {'__all__': ["Invalid value for custom field 'example': Value must be at least 1000"]}

```

The above example is via pynetbox, but it doesn't matter, the Exception Message would be the same in all cases.

### Expected Behavior

ValidationError's message should be `Invalid value for custom field 'example': Value must be at least 2"`

### Observed Behavior

Instead, the exception message reports the `validation_maximum` instead of the `validation_minimum`. It's a tiny typo in the CustomField model validation code, latest develop:
https://github.com/netbox-community/netbox/blob/85396866bca8e1831f82be8ceb17a1e897dcaa69/netbox/extras/models/customfields.py#L662-L664

costasd:
### Fixes: #17611 

Fix a ValidationError message typo

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -661,7 +661,7 @@ class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
                     raise ValidationError(_("Value must be an integer."))
                 if self.validation_minimum is not None and value < self.validation_minimum:
                     raise ValidationError(
-                        _("Value must be at least {minimum}").format(minimum=self.validation_maximum)
+                        _("Value must be at least {minimum}").format(minimum=self.validation_minimum)
                     )
                 if self.validation_maximum is not None and value > self.validation_maximum:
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @costasd!

PR #17625, opened 2024-09-26
Fixes: #17612 : Update ASN types in asn.md

### LINKED ISSUE #17612: Incorrect ANS types in asn.md
### Change Type

### Area

Other

### Proposed Changes

netbox/docs/models/ipam/**asn.md** contains incorrect ASN type - 64-bit AS number, whereas ASN types are 16- and 32-bit only :
[https://www.iana.org/assignments/as-numbers/as-numbers.xhtml](https://www.iana.org/assignments/as-numbers/as-numbers.xhtml)

Proposed cleanups :
"NetBox support both **32- and 64-** ASNs." -> "NetBox support both **16- and 32-bit** ASNs."
"The **32- or 64-bit** AS number." -> "The **16- or 32-bit** AS number."

DM3740:
### Fixes: #17612

Proposed Changes :

"NetBox support both 32- and 64- ASNs." --> "NetBox support both 16- and 32-bit ASNs."
"The 32- or 64-bit AS number." --> "The 16- or 32-bit AS number."

### DIFF
--- a/docs/models/ipam/asn.md
+++ b/docs/models/ipam/asn.md
@@ -1,6 +1,6 @@
 # ASNs
 
-An Autonomous System Number (ASN) is a numeric identifier used in the BGP protocol to identify which [autonomous system](https://en.wikipedia.org/wiki/Autonomous_system_%28Internet%29) a particular prefix is originating and transiting through. NetBox support both 16- and 32-bit ASNs.
+An Autonomous System Number (ASN) is a numeric identifier used in the Border Gateway Protocol (BGP) to identify which [autonomous system](https://en.wikipedia.org/wiki/Autonomous_system_%28Internet%29) a particular prefix is originating from or transiting through. NetBox supports both 16- and 32-bit ASNs.
 
 ASNs must be globally unique within NetBox, and may be allocated from within a [defined range](./asnrange.md). Each ASN may be assigned to multiple [sites](../dcim/site.md).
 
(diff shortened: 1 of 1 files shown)


docs/models/ipam/asn.md
```diff
@@ -1,14 +1,14 @@
 # ASNs
 
-An Autonomous System Number (ASN) is a numeric identifier used in the BGP protocol to identify which [autonomous system](https://en.wikipedia.org/wiki/Autonomous_system_%28Internet%29) a particular prefix is originating and transiting through. NetBox support both 32- and 64- ASNs.
+An Autonomous System Number (ASN) is a numeric identifier used in the BGP protocol to identify which [autonomous system](https://en.wikipedia.org/wiki/Autonomous_system_%28Internet%29) a particular prefix is originating and transiting through. NetBox support both 16- and 32-bit ASNs.
```
bluikko, line 3: While you are at it, how about fixing this invalid sentence?

Right now it says in essence "AS a particular prefix is originating" which means the prefix would be originating the AS.  
Obviously this is the other way around, the AS originates a prefix.

Additionally origination and transiting are mutually exclusive imho so change the "and" to "or".

```suggestion
An Autonomous System Number (ASN) is a numeric identifier used in the BGP protocol to identify which [autonomous system](https://en.wikipedia.org/wiki/Autonomous_system_%28Internet%29) a particular prefix is originated from or transiting through. NetBox support both 16- and 32-bit ASNs.
```

DM3740: OK

jeremystretch (approved):
Thanks @DM3740!

PR #17629, opened 2024-09-26
16547 Add distance to Circuit

### LINKED ISSUE #16547: Circuit Distance
### NetBox version

4.0.5

### Feature type

### Proposed functionality

Add optional distance field to circuit model. 

This is similar to the wireless link distance on #15106

### Use case

This would allow for continuous cable trace of dark fibre circuits to provide a full end to end route distance when connected to other cables and patch panels etc. Currently this can only be accomplished by classifying dark fibre circuit as a cable, but that loses the provider model and contact functionality so is not an ideal solution.

### Database changes

Unknown

### External dependencies

Unknown

arthanson:
### Fixes: #16547 

This has more changes as refactored WirelessLinkDistanceChoices to DistanceChoices as made it as a mixin as it is used in 3 models now so should make it DRY.


netbox/circuits/migrations/0045_circuit__abs_distance_circuit_distance_and_more.py
jeremystretch, line 1: Let's give the migration a simple, meaningful name, e.g. `0045_circuit_distance`.

netbox/circuits/tables/circuits.py
```diff
@@ -76,6 +77,10 @@ class CircuitTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
     commit_rate = CommitRateColumn(
         verbose_name=_('Commit Rate')
     )
+    distance = columns.TemplateColumn(
```
jeremystretch, line 80: It's probably worth creating a custom column for this, as identical template code is also defined for WirelessLinkTable. (This new column could likely also be used for `length` on CableTable.)

netbox/dcim/models/mixins.py
```diff
@@ -1,56 +1,12 @@
...
-from dcim.choices import *
-from utilities.conversion import to_grams
 
 __all__ = (
     'RenderConfigMixin',
-    'WeightMixin',
 )
 
 
-class WeightMixin(models.Model):
```
jeremystretch, line 13: Is there a reason to move this right now? We shouldn't be touching anything not directly related to the distance fields in this PR, as it's not relevant to the feature request. (We can reorganize the weight resources under a separate PR if needed.)

netbox/templates/circuits/circuit.html
```diff
@@ -34,6 +34,20 @@ <h2 class="card-header">{% trans "Circuit" %}</h2>
             <th scope="row">{% trans "Status" %}</th>
             <td>{% badge object.get_status_display bg_color=object.get_status_color %}</td>
           </tr>
+          <tr>
+            <th scope="row">{% trans "Description" %}</th>
+            <td>{{ object.description|placeholder }}</td>
+          </tr>
```
jeremystretch, line 40: This duplicates the `description` field.

jeremystretch (changes requested):

jeremystretch (approved):

PR #17638, opened 2024-09-27
Fixes #16024: Change `attr_type` from list to str for MultipleChoiceFilter

### LINKED ISSUE #16024: New GraphQL filters (AND, OR and NOT) don't work as intended
### Deployment Type

Self-hosted

### NetBox Version

v4.0.0

### Python Version

3.11

### Steps to Reproduce

Let's take the OR-Filter as an example: As described in [the docs](https://strawberry-graphql.github.io/strawberry-django/guide/filters/#and-or-not-distinct), I would like to filter for VMs that are in the `active` and in the `staged` status:
```
{
  virtual_machine_list(filters: {status: "staged", OR: {status: "active"}}) {
    name
    status
  }
}
```
Returns:
```
{
  "data": {
    "virtual_machine_list": []
  }
}
```

As seen, this query returns an empty result, while distinct queries for `staged` and `active` VMs return results:
```
{
  virtual_machine_list(filters: {status: "staged"}) {
    name
    status
  }
}
```
Returns:
```
{
  "data": {
    "virtual_machine_list": [
      {
        "name": "vm3",
        "status": "staged"
      },
      {
        "name": "vm4",
        "status": "staged"
      }
    ]
  }
}
```
And
```
{
  virtual_machine_list(filters: {status: "active"}) {
    name
    status
  }
}
```
Returns:
```
{
  "data": {
    "virtual_machine_list": [
      {
        "name": "vm1",
        "status": "active"
      },
      {
        "name": "vm2",
        "status": "active"
      }
    ]
  }
}
```

Or another example: the NOT-Filter: let's assume that I want to filter for all VMs that are **not** `active`:
```
{
  virtual_machine_list(filters: {NOT: {status: "active"}}) {
    name
    status
  }
}
```

But for some reason, this **only** returns `active` VMs:
```
{
  "data": {
    "virtual_machine_list": [
      {
        "name": "vm1",
        "status": "active"
      },
      {
        "name": "vm2",
        "status": "active"
      }
    ]
  }
}
```

### Expected Behavior

That the OR-Filter returns `active` and `staged` VMs and that the NOT-Filter returns everything except `active` VMs.

### Observed Behavior

Described above.

jeremystretch:
### Fixes: #16024

This seems to fix the AND/OR/NOT filtering specifically for multiple-choice filters. Additional work is likely needed but I wanted to put this out there for testing.

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -271,7 +271,7 @@ class LocationFilterSet(TenancyFilterSet, ContactModelFilterSet, OrganizationalM
 
     class Meta:
         model = Location
-        fields = ('id', 'name', 'slug', 'status', 'facility', 'description')
+        fields = ('id', 'name', 'slug', 'facility', 'description')
 
     def search(self, queryset, name, value):
         if not value.strip():
--- a/netbox/netbox/graphql/filter_mixins.py
+++ b/netbox/netbox/graphql/filter_mixins.py
@@ -1,11 +1,12 @@
-from functools import partial, partialmethod, wraps
+from functools import partialmethod
 from typing import List
 
 import django_filters
 import strawberry
 import strawberry_django
-from django.core.exceptions import FieldDoesNotExist, ValidationError
+from django.core.exceptions import FieldDoesNotExist
 from strawberry import auto
+
 from ipam.fields import ASNField
 from netbox.graphql.scalars import BigInt
 from utilities.fields import ColorField, CounterCacheField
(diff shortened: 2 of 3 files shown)


arthanson:
That does seem to fix the major issues, there are some issues with NOT handling, but I think that is potentially a separate issue.

arthanson (approved):
I think go ahead with this.  It is a work-around rather than a direct fix, but will fix it for this specific case as it bypasses using Django-filter and doesn't fix the underlying incompatibility between Strawberry and Django-filter.  But this does fix a a larger part of the filtering and I couldn't find a drawback to going forward with this.

What this basically does is not go the Django-filter function to do the filtering, but just relies on Strawberry to do a char filter, which works as this is on the model and is char field.

PR #17640, opened 2024-09-27
Add support for socks connection to Git backend

### LINKED ISSUE #17639: Add socks support to proxy settings for Git remote data source
### NetBox version

v4.1.2

### Feature type

Data model extension

### Proposed functionality

Add socks ('socks5:') support for Git remote data sources.

Currently Dulwich (the git client API library) doesn't support passing in HTTP_PROXIES with 'socks5:', this works fine on other proxy requests like WebHooks.

### Use case

This is needed if using a secure SOCKS proxy and accessing a remote Git data source.

### Database changes

n/a

### External dependencies

Requires the python_socks library if used.

arthanson:
### Fixes: #17639

**Note:** will also need to install the python_socks library along with the Dulwich library to use this:
```
pip install dulwich
pip install python_socks
```

To test locally, start a socks proxy in a docker container. Make a HTTP request and observe it using the proxy:

```
$ docker run -d --name socks5 -p 1080:1080 serjs/go-socks5-proxy
$ docker logs -f socks5
```

Specify the local proxy in Netbox configuration.py:
```
HTTP_PROXIES = {
   'http': 'socks5h://localhost:1080',
   'https': 'socks5h://localhost:1080',
}
```

Add a remote GIT datasource and you should see the traffic logs in the terminal. Setup a DataSource in Netbox, I was using the following in the form:

```
Name = GitSync
Type = Git
URL = https://github.com/bogdancordos/netbox_scripts.git
```
The URL just has a selection of scripts so you can see that it actually syncs the data source.  Press the [Sync] button and make sure it actually syncs.

For SOCKs auth scenario it can be setup as follows:

```
docker run -d --name socks5 -p 1080:1080 -e PROXY_USER=aaa -e PROXY_PASSWORD=bbb  serjs/go-socks5-proxy
```

Then in configuration.py:
```
HTTP_PROXIES = {
    'http': 'socks5h://aaa:bbb@localhost:1080',
    'https': 'socks5h://aaa:bbb@localhost:1080',
}
```

### DIFF
--- a/docs/features/synchronized-data.md
+++ b/docs/features/synchronized-data.md
@@ -13,6 +13,9 @@ To enable remote data synchronization, the NetBox administrator first designates
 !!! info
     Data backends which connect to external sources typically require the installation of one or more supporting Python libraries. The Git backend requires the [`dulwich`](https://www.dulwich.io/) package, and the S3 backend requires the [`boto3`](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) package. These must be installed within NetBox's environment to enable these backends.
 
+!!! info
+    If you are configuring Git and have HTTP_PROXIES set to use the SOCKS protocol, you will also need to install the [`python_socks`](https://pypi.org/project/python-socks/) package.
+
 Each type of remote source has its own configuration parameters. For instance, a git source will ask the user to specify a branch and authentication credentials. Once the source has been created, a synchronization job is run to automatically replicate remote files in the local database.
 
 The following NetBox models can be associated with replicated data files:
--- a/docs/features/synchronized-data.md
+++ b/docs/features/synchronized-data.md
@@ -14,7 +14,7 @@ To enable remote data synchronization, the NetBox administrator first designates
     Data backends which connect to external sources typically require the installation of one or more supporting Python libraries. The Git backend requires the [`dulwich`](https://www.dulwich.io/) package, and the S3 backend requires the [`boto3`](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) package. These must be installed within NetBox's environment to enable these backends.
 
 !!! info
-    If you are configuring Git and have HTTP_PROXIES set to use the SOCKS protocol, you will also need to install the [`python_socks`](https://pypi.org/project/python-socks/) package.
+    If you are configuring Git and have `HTTP_PROXIES` configured to use the SOCKS protocol, you will also need to install the [`python_socks`](https://pypi.org/project/python-socks/) Python library.
 
 Each type of remote source has its own configuration parameters. For instance, a git source will ask the user to specify a branch and authentication credentials. Once the source has been created, a synchronization job is run to automatically replicate remote files in the local database.
 
--- a/netbox/core/data_backends.py
+++ b/netbox/core/data_backends.py
@@ -79,7 +79,7 @@ class ProxyPoolManager(PoolManager):
         # python_socks uses rdns param to denote remote DNS parsing and
         # doesn't accept the 'h' or 'a' in the proxy URL
         cleaned_proxy_url = proxy_url
-        if use_rdns := urlparse(cleaned_proxy_url).scheme.lower() in ['socks4h', 'socks4a' 'socks5h', 'socks5a']:
+        if use_rdns := urlparse(cleaned_proxy_url).scheme in ['socks4h', 'socks4a' 'socks5h', 'socks5a']:
             cleaned_proxy_url = cleaned_proxy_url.replace('socks5h:', 'socks5:').replace('socks5a:', 'socks5:')
             cleaned_proxy_url = cleaned_proxy_url.replace('socks4h:', 'socks4:').replace('socks4a:', 'socks4:')
 
@@ -150,7 +150,7 @@ class GitBackend(DataBackend):
         if settings.HTTP_PROXIES and self.url_scheme in ('http', 'https'):
             if proxy := settings.HTTP_PROXIES.get(self.url_scheme):
                 config.set("http", "proxy", proxy)
-                if urlparse(proxy).scheme.lower() in ['socks4', 'socks4a', 'socks4h', 'socks5', 'socks5a', 'socks5h']:
+                if urlparse(proxy).scheme in ['socks4', 'socks4a', 'socks4h', 'socks5', 'socks5a', 'socks5h']:
                     self.use_socks = True
 
         return config
--- a/netbox/core/data_backends.py
+++ b/netbox/core/data_backends.py
@@ -8,10 +8,12 @@ from urllib.parse import urlparse
 
 from django import forms
 from django.conf import settings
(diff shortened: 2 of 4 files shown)


netbox/core/data_backends.py
```diff
@@ -23,6 +25,79 @@
 logger = logging.getLogger('netbox.data_backends')
 
 
+# These Proxy Methods are for handling SOCKS connection proxy
```
jeremystretch, line 28: These should live in a separate module under `utilties`.

docs/features/synchronized-data.md
```diff
@@ -13,6 +13,9 @@ To enable remote data synchronization, the NetBox administrator first designates
 !!! info
     Data backends which connect to external sources typically require the installation of one or more supporting Python libraries. The Git backend requires the [`dulwich`](https://www.dulwich.io/) package, and the S3 backend requires the [`boto3`](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) package. These must be installed within NetBox's environment to enable these backends.
 
+!!! info
+    If you are configuring Git and have HTTP_PROXIES set to use the SOCKS protocol, you will also need to install the [`python_socks`](https://pypi.org/project/python-socks/) package.
```
jeremystretch, line 17: Why is `python-socks` needed?

arthanson: It does all the handling for socks, see import in utilities/socks.py line 15

jeremystretch, line 17:
```suggestion
    If you are configuring Git and have `HTTP_PROXIES` configured to use the SOCKS protocol, you will also need to install the [`python_socks`](https://pypi.org/project/python-socks/) Python library.
```

netbox/core/data_backends.py
```diff
@@ -67,11 +68,14 @@ def init_config(self):
 
         # Initialize backend config
         config = ConfigDict()
+        self.use_socks = False
 
         # Apply HTTP proxy (if configured)
         if settings.HTTP_PROXIES and self.url_scheme in ('http', 'https'):
             if proxy := settings.HTTP_PROXIES.get(self.url_scheme):
                 config.set("http", "proxy", proxy)
+                if urlparse(proxy).scheme in ['socks4', 'socks4a', 'socks4h', 'socks5', 'socks5a', 'socks5h']:
```
jeremystretch, line 77: Feels like we should define these in `constants.py`.

netbox/utilities/socks.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class ProxyPoolManager(PoolManager):
+    def __init__(self, proxy_url, timeout=5, num_pools=10, headers=None, **connection_pool_kw):
+        # python_socks uses rdns param to denote remote DNS parsing and
+        # doesn't accept the 'h' or 'a' in the proxy URL
+        cleaned_proxy_url = proxy_url
+        if use_rdns := urlparse(cleaned_proxy_url).scheme in ['socks4h', 'socks4a', 'socks5h', 'socks5a']:
+            cleaned_proxy_url = cleaned_proxy_url.replace('socks5h:', 'socks5:').replace('socks5a:', 'socks5:')
+            cleaned_proxy_url = cleaned_proxy_url.replace('socks4h:', 'socks4:').replace('socks4a:', 'socks4:')
```
jeremystretch, line 60: Should we just raise an `ImproperlyConfigured` exception if the scheme is unsupported? I suspect it would help reduce troubleshooting effort.

netbox/utilities/socks.py
```diff
@@ -0,0 +1,76 @@
...
+
+class RdnsProxyHTTPSConnectionPool(HTTPSConnectionPool):
+    ConnectionCls = RdnsProxyHTTPSConnection
+
+
+class ProxyPoolManager(PoolManager):
+    def __init__(self, proxy_url, timeout=5, num_pools=10, headers=None, **connection_pool_kw):
+        # python_socks uses rdns param to denote remote DNS parsing and
+        # doesn't accept the 'h' or 'a' in the proxy URL
+        cleaned_proxy_url = proxy_url
```
jeremystretch, line 57: If we keep this approach, we can just refer to this variable as `proxy_url` consistently.

netbox/utilities/socks.py
```diff
@@ -0,0 +1,76 @@
...
+class ProxyHTTPConnection(HTTPConnection):
+    use_rdns = False
+
+    def __init__(self, *args, **kwargs):
+        socks_options = kwargs.pop('_socks_options')
+        self._proxy_url = socks_options['proxy_url']
+        super().__init__(*args, **kwargs)
+
+    def _new_conn(self):
+        from python_socks.sync import Proxy
```
jeremystretch, line 16: We should catch & re-raise `ImportError` here to include a helpful error message.

netbox/utilities/socks.py
```diff
@@ -0,0 +1,76 @@
+from urllib.parse import urlparse
+from urllib3 import PoolManager, HTTPConnectionPool, HTTPSConnectionPool
+from urllib3.connection import HTTPConnection, HTTPSConnection
+
+
+# These Proxy Methods are for handling SOCKS connection proxy
```
jeremystretch, line 6: Could you expand on this a bit to explain why these classes are necessary?

jeremystretch (changes requested):
I've tested this both with and without authentication, and syncing worked as expected both times. Nice work @arthanson!

netbox/utilities/socks.py
arthanson: It's in the code that is executed by the background task queue so the error is never propagated to the front-end.  In the RQ log there is the message:
```
Fetching remote data failed (ModuleNotFoundError): No module named 'python_socks'
```
Which should be similar to dulwich and any other missing library so I think a different error message might be more confusing?

netbox/utilities/constants.py
```diff
@@ -93,3 +93,7 @@
     "td": {"align"},
     "th": {"align"},
 }
+
+DATA_SOURCE_SUPPORTED_SOCK_SCHEMAS = ['socks4', 'socks4a', 'socks4h', 'socks5', 'socks5a', 'socks5h']
+DATA_SOURCE_SOCK_RDNS_SCHEMAS = ['socks4h', 'socks4a', 'socks5h', 'socks5a']
+DATA_SOURCE_SUPPORTED_SCHEMAS = ['http', 'https', 'socks4', 'socks4a', 'socks4h', 'socks5', 'socks5a', 'socks5h']
```
jeremystretch, line 99: These aren't relevant to data sources specifically; they're pertinent to the configured HTTP proxies (if any), which are used for many different functions. IMO a name like `HTTP_PROXY_*` would make more sense.

netbox/utilities/socks.py
jeremystretch: We should log an error which hints as to the root cause; something like "Configuring an HTTP proxy using SOCKS requires the python_socks library. Check that it has been installed."

jeremystretch (approved):

PR #17662, opened 2024-10-02
limits vlans on interface tables

### LINKED ISSUE #17655: Limit VLANs displayed on interface tables
### NetBox version

v4.1.2

### Feature type

### Proposed functionality

Limit number of tagged VLANs shown on interface/vminterface table to 3 VLANs. If interface is connected to more than 3 VLANs, then show a link to full list. Just like with IP addresses, that was implemented in #14645

### Use case

Make overview of interfaces more readable, if a device has some interfaces that's connected to many VLANs.

### Database changes

_No response_

### External dependencies

_No response_

thordreier:
### Fixes: #17655

Limit VLANs displayed on interface tables

### DIFF
--- a/netbox/dcim/tables/template_code.py
+++ b/netbox/dcim/tables/template_code.py
@@ -57,7 +57,7 @@ INTERFACE_FHRPGROUPS = """
 INTERFACE_TAGGED_VLANS = """
 {% if record.mode == 'tagged' %}
   {% if value.count > 3 %}
-    <a href="{% url 'ipam:vlan_list' %}?{{ record|meta:"model_name" }}_id={{ record.pk }}">{{ value.count }}</a>
+    <a href="{% url 'ipam:vlan_list' %}?{{ record|meta:"model_name" }}_id={{ record.pk }}">{{ value.count }} VLANs</a>
   {% else %}
     {% for vlan in value.all %}
         <a href="{{ vlan.get_absolute_url }}">{{ vlan }}</a><br />
--- a/netbox/dcim/tables/template_code.py
+++ b/netbox/dcim/tables/template_code.py
@@ -56,9 +56,13 @@ INTERFACE_FHRPGROUPS = """
 
 INTERFACE_TAGGED_VLANS = """
 {% if record.mode == 'tagged' %}
+  {% if value.count > 3 %}
+    <a href="{{ record.get_absolute_url }}#vlans">{{ value.count }}</a>
+  {% else %}
     {% for vlan in value.all %}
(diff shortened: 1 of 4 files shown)


netbox/dcim/tables/template_code.py
```diff
@@ -56,9 +56,13 @@
 
 INTERFACE_TAGGED_VLANS = """
 {% if record.mode == 'tagged' %}
+  {% if value.count > 3 %}
+    <a href="{{ record.get_absolute_url }}#vlans">{{ value.count }}</a>
```
jeremystretch, line 60: Instead of the interface view, let's link to the VLAN list with a filter for the current interface. This will be consistent with the way we currently handle assigned IP addresses (see `INTERFACE_IPADDRESSES`).

jeremystretch (changes requested):

netbox/dcim/tables/template_code.py
thordreier: That was also my initial plan, but VLANs doesn't seem to have interface filter like IP-address do.
So ipam/vlans/?(vm)interface_id=XXX needs to be implemented first for this to work.
That's why I implemented it with "vlan" anchor on interface page instead.

jeremystretch: Ah right. I've just opened FR #17669 to add filters for this. I'll see if I can knock that out real quick today.

thordreier: I've updated the pull request to support #17669

sleepinggenius2:
Just my two cents on this, as I know it's being implemented the same way as for IP addresses. For IP addresses, having just a number instead of a list of IPs is somewhat intuitive as to what is being conveyed, but in this context, I have a concern that having just a number may be misinterpreted as representing an actual VID and not a count that needs to be clicked on for additional details.

thordreier:
@sleepinggenius2 I've changed it to "5 VLANs" - not just "5"

jeremystretch (changes requested):
To ensure that VLAN IDs are rendered correctly when exporting e.g. to CSV, we also need to add a method to BaseInterfaceTable (similar to `value_ip_addresses`). This should work:

```python
    def value_tagged_vlans(self, value):
        return ",".join([str(obj.vid) for obj in value.all()])
```

thordreier:
I used `str(obj)` instead of `str(obj.vid)` to be consistent with previous behavior (before this change, you got "VLAN-name (VID)" when exporting to CSV, not just "VID")

jeremystretch (approved):
Thanks @thordreier!

PR #17670, opened 2024-10-03
Fixes: #17663 - Only remove extraneous attributes from extra if changing to a BooleanFilter

### LINKED ISSUE #17663: Type interface filter not working since 4.1.2
### Deployment Type

Self-hosted

### NetBox Version

V4.1.2

### Python Version

3.10

### Steps to Reproduce

Example urls which fail
- https://demo.netbox.dev/api/dcim/interfaces/?type__isw=bridge
- https://demo.netbox.dev/api/dcim/interfaces/?type__nic=bridge

### Expected Behavior

Api response to work (fails in swagger example too)

### Observed Behavior

```
HTTP 400 Bad Request
Allow: GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS, TRACE
Vary: Accept

{
    "type__nic": [
        "Select a valid choice. bridge is not one of the available choices."
    ]
}
```

Example response

bctiemann:
### Fixes: #17663

Refines https://github.com/netbox-community/netbox/pull/17574 to ensure we only remove `choices` from the filter's `extra` attributes if we are artificially changing it to a `BooleanFilter` (i.e. if the lookup is `__empty`). This fixes an issue where `choices` was not being respected in the case of `MultipleChoiceField` where it should be allowed.

Also adds a test to ensure `MultipleChoiceField` character-based lookups are exercised.
Also adds tests for `__empty` lookup to the base filter lookup tests to exercise that behavior generically.

### DIFF
--- a/netbox/netbox/filtersets.py
+++ b/netbox/netbox/filtersets.py
@@ -183,8 +183,8 @@ class BaseFilterSet(django_filters.FilterSet):
                     filter_cls = type(existing_filter)
                     if lookup_expr == 'empty':
                         filter_cls = django_filters.BooleanFilter
-                        for field_to_remove in ('choices', 'null_value'):
-                            existing_filter_extra.pop(field_to_remove, None)
+                        for param_to_remove in ('choices', 'null_value'):
+                            existing_filter_extra.pop(param_to_remove, None)
                     new_filter = filter_cls(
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @bctiemann!

PR #17692, opened 2024-10-07
Closes #6414: Enable assigning prefixes to various object types

### LINKED ISSUE #6414: Allow prefixes to be assigned to a site group, region, site, or location
### NetBox version

v2.11.3

### Feature type

New functionality

### Proposed functionality

Allow prefixes to be assigned to a site-group and not a specific site

### Use case

Lets say we have a site-group called "Singapore" which contains multiple sites e.g SG1, SG2, SG3
We have servers in each site which get assigned /31s IPs from a /24
So we'd like the ability to assign the /24 prefix to the site-group only, since it does not "belong" to only one site.
From there we can assign smaller prefixes into the relevant site as needed

### Database changes

unsure, don't think so

### External dependencies

none

jeremystretch:
### Fixes: #6414

- Add a `scope` GenericForeignKey on Prefix
- Add denormalized relations on Prefix to the scope object types (region, site group, site, and location)
- Populate scope & denormalized relations on migration
- Remove the original `site` ForeignKey field from Prefix
- Update PrefixFilterSet filters

### DIFF
--- a/netbox/ipam/forms/bulk_edit.py
+++ b/netbox/ipam/forms/bulk_edit.py
@@ -204,25 +204,6 @@ class RoleBulkEditForm(NetBoxModelBulkEditForm):
 
 
 class PrefixBulkEditForm(NetBoxModelBulkEditForm):
-    # region = DynamicModelChoiceField(
-    #     label=_('Region'),
-    #     queryset=Region.objects.all(),
-    #     required=False
-    # )
-    # site_group = DynamicModelChoiceField(
-    #     label=_('Site group'),
-    #     queryset=SiteGroup.objects.all(),
-    #     required=False
-    # )
-    # site = DynamicModelChoiceField(
-    #     label=_('Site'),
-    #     queryset=Site.objects.all(),
-    #     required=False,
-    #     query_params={
-    #         'region_id': '$region',
-    #         'group_id': '$site_group',
-    #     }
-    # )
     vlan_group = DynamicModelChoiceField(
         queryset=VLANGroup.objects.all(),
         required=False,
@@ -282,7 +263,6 @@ class PrefixBulkEditForm(NetBoxModelBulkEditForm):
     model = Prefix
     fieldsets = (
         FieldSet('tenant', 'status', 'role', 'description'),
-        # FieldSet('region', 'site_group', 'site', name=_('Site')),
         FieldSet('vrf', 'prefix_length', 'is_pool', 'mark_utilized', name=_('Addressing')),
         FieldSet('vlan_group', 'vlan', name=_('VLAN Assignment')),
     )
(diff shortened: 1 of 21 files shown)


netbox/ipam/forms/bulk_edit.py
```diff
@@ -203,25 +203,6 @@ class RoleBulkEditForm(NetBoxModelBulkEditForm):
 
 
 class PrefixBulkEditForm(NetBoxModelBulkEditForm):
```
arthanson, line 205: should we add scope here?

netbox/ipam/forms/filtersets.py
```diff
@@ -224,12 +224,13 @@ class PrefixFilterForm(TenancyFilterForm, NetBoxModelFilterSetForm):
     site_id = DynamicModelMultipleChoiceField(
         queryset=Site.objects.all(),
         required=False,
-        null_option='None',
-        query_params={
-            'region_id': '$region_id'
-        },
         label=_('Site')
     )
+    location_id = DynamicModelMultipleChoiceField(
```
arthanson, line 229: wondering if these could be a mixin ScopedFilterSet as we are going to be using this in several models and they will all have the same filter, form fields etc.. could have site_id, location_id, site_group all in the mixin.

netbox/ipam/forms/model_forms.py
```diff
@@ -228,17 +234,48 @@ class PrefixForm(TenancyForm, NetBoxModelForm):
...
 
+    def __init__(self, *args, **kwargs):
+        instance = kwargs.get('instance')
+        initial = kwargs.get('initial', {})
+
+        if instance is not None and instance.scope:
+            initial['scope'] = instance.scope
+            kwargs['initial'] = initial
+
+        super().__init__(*args, **kwargs)
```
arthanson, line 257: again seems like a good potential place for a mixin to reduce boilerplate as we are going to be using this same pattern in about 5 places.

arthanson (changes requested):

netbox/ipam/forms/bulk_edit.py
jeremystretch: I think we run into a snag using HTMXSelector on a bulk edit form: The bulk edit forms must be POSTed to capture the selected object PKs, whereas HTMXSelector uses a GET to re-render the form. We'll need to extend HTMXSelector to allow POSTing instead.

netbox/ipam/forms/filtersets.py
jeremystretch: It's not yet clear which models will support this same set of scope types. VLANGroup, for instance, supports these four but also racks, cluster groups, and clusters. We'll utilize mixins at some point for sure but it's probably still too early for that degree of optimization.

netbox/ipam/forms/bulk_edit.py
jeremystretch: Going to work on #17789 first to get a pattern established in `develop`, then port it over here.

jeremystretch:
Temporarily blocked by some prework for GFK bulk edit support in #17789

PR #17693, opened 2024-10-07
16546 define get_absolute_url method in NetBoxModel

arthanson:

### DIFF
--- a/netbox/netbox/models/__init__.py
+++ b/netbox/netbox/models/__init__.py
@@ -41,6 +41,9 @@ class NetBoxFeatureSet(
     def docs_url(self):
         return f'{settings.STATIC_URL}docs/models/{self._meta.app_label}/{self._meta.model_name}/'
 
+    def get_absolute_url(self):
+        return reverse(f'{self._meta.app_label}:{self._meta.model_name}', args=[self.pk])
+
 
 #
 # Base model classes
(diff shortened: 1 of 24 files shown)


netbox/netbox/models/__init__.py
```diff
@@ -201,3 +205,6 @@ class Meta:
 
     def __str__(self):
         return self.name
+
+    def get_absolute_url(self):
```
bctiemann, line 209: Why does it need to be redefined here if it inherits from `NetBoxModel`?

arthanson: Because it doesn't inherit from NetBoxModel, it inherits from NetBoxFeatureSet, I don't see NetBoxModel in the chain unless I'm missing something?

bctiemann: Oh, my mistake, I didn't expand upward enough and misread which class this was. Disregard.

bctiemann (approved):

jeremystretch (changes requested):
This doesn't address subclasses of OrganizationalModel. We should also avoid duplicating the method if possible. It might be cleaner to define `get_absolute_url()` on NetBoxFeatureSet instead.

jeremystretch (changes requested):
Attempting to run `makemigrations` fails:

```
django.core.exceptions.FieldError: 'action_object' cannot be specified for EventRule model form as it is a non-editable field
```

(Make sure you have the [pre-commit hook](https://netboxlabs.com/docs/netbox/en/stable/development/getting-started/#3-enable-pre-commit-hooks) enabled to catch these errors on commit.)

Edit: My mistake; this was because I still had Django 5.1 installed from a previous PR.

jeremystretch (approved):

PR #17694, opened 2024-10-08
Add EVPN-VPWS to L2VPNTypeChoices

### LINKED ISSUE #17216: Add EVPN-VPWS to the L2VPNTypeChoices Choice set.
### NetBox version

v4.0.9

### Feature type

Data model extension

### Proposed functionality

EVPN-VPWS is a point to point L2VPN service that breaks the current model for VPWS and other P2P types in the L2VPN termination list as it allows for multiple interfaces across multiple network devices to the same endpoint or location. In either an Active / Active or Active / Passive mode. 

Suggested change 

Add TYPE_EVPN_VPWS to the L2VPNTypeChoices class, but don't add it to L2VPNTypeChoices.P2P

Relevant RFC: https://datatracker.ietf.org/doc/html/rfc8214

### Use case

This would allow Netbox users to model EVPN-VPWS separately from RFC4664 type VPWS which doesn't support Active / Active across multiple devices to the same location. 

### Database changes

_No response_

### External dependencies

_No response_

caskings:
### Fixes: #17216

Updated L2VPNTypeChoices to include the EVPN-VPWS type but didn't include it in the P2P list so more than two interfaces can be assigned to it.

### DIFF
--- a/docs/models/vpn/l2vpn.md
+++ b/docs/models/vpn/l2vpn.md
@@ -28,6 +28,7 @@ The technology employed in forming and operating the L2VPN. Choices include:
 * VXLAN-EVPN
 * MPLS-EVPN
 * PBB-EVPN
+* EVPN-VPWS
 
 !!! note
     Designating the type as VPWS, EPL, EP-LAN, EP-TREE will limit the L2VPN instance to two terminations.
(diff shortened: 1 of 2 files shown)


DanSheps (approved):
Just to leave a little background on the PR.  While a VPWS is a Point-to-Point protocol, current in NetBox we have no way to properly model the relationship between two devices that are part of a MC-LAG using EVPN ESI (The chassis' don't talk to each other over a MC-LAG protocol, they use EVPN to build the MC-LAG; I believe technically a chassis could be part of an MC-LAG with one device on one bundle and then part of a different MC-LAG with a different bundle, but I haven't dove too far into it)

So you would either need to model two VPWS and link them together (maybe a parent/child relationship), build a new model to bridge the two devices for MC-LAG or allow a multi-connect like this.  I think this is the more reasonable way to handle it for no, anything else would require substantial effort.

PR #17716, opened 2024-10-09
Closes #16971: Add system jobs

### LINKED ISSUE #16971: Allow plugin authors to register housekeeping tasks
### NetBox version

v4.0.7

### Feature type

New functionality

### Proposed functionality

NetBox includes a housekeeping management command that _usually_ runs nightly. This command handles:

- Clearing expired authentication sessions from the database
- Deleting older changelog records
- Deleting older job result records
- Checking for new NetBox releases

It would be great if plugin authors could register additional tasks to run by the housekeeping process without having to write their own management commands and setup mechanisms to run those periodically.

### Use case

I often have to write periodic tasks to alter model instances in my plugins, reacting to dates, sending emails etc. and having to setup the whole management commands and related mechanisms to run those periodically is not really convenient, even more so considering that NetBox already has mechanisms built in to do so.

### Database changes

None whatsoever

### External dependencies

None

alehaa:
### Fixes: #16971

This PR allows the NetBox core and plugins to register job runners as system jobs. These are automatically scheduled without UI interaction, just before the worker process starts. This allows job runners to be used for housekeeping and synchronization tasks in future commits.


jeremystretch:
@alehaa I think we might have jumped the gun a bit on FR #16971: Because this is introducing a new plugin capability, it needs to be tagged for a minor release. (We can't ship any changes to the plugins API in a patch release.) I have no issue tagging for v4.2 though. Are okay with rebasing this against the `feature` branch? It shouldn't present much conflict.

alehaa:
That's fine by me. Do you initiate the rebase or do I have to do it locally and submit a new PR?

jeremystretch:
I should be able to tackle the rebase, just wanted to give you a heads up.

bctiemann (approved):

docs/plugins/development/background-jobs.md
```diff
@@ -46,25 +57,49 @@ As described above, jobs can be scheduled for immediate execution or at any late
...
+
+#### Example
+
 ```python title="jobs.py"
+from core.choices import JobIntervalChoices
 from netbox.jobs import JobRunner
-
+from .models import MyModel
 
 class MyHousekeepingJob(JobRunner):
```
arthanson, line 92: I think an import to MyHousekeepingJob still needs to be in __init even if there is only a system_job to run (no other background jobs?  Also, I noticed separate issue the Example in the docs line ` MyHousekeepingJob.setup(interval=60)` I think is incorrect as there is no setup() method in the JobRunner and it was throwing an error for me when I had copied the code?

That said I wasn't actually getting the job firing on schedule when I added it as a test to a plugin I had installed, not sure if I had it mis-configured. In the plugins jobs.py:

```
from core.choices import JobIntervalChoices
from netbox.jobs import JobRunner

class MyHousekeepingJob(JobRunner):
    class Meta:
        name = "My Housekeeping Job"
        system_interval = JobIntervalChoices.INTERVAL_MINUTELY

    def run(self, *args, **kwargs):
        print("in run...")

system_jobs = (
    MyHousekeepingJob,
)
```
In the __init:
```
from netbox.plugins import PluginConfig

class HealthCheckConfig(PluginConfig):
    name = 'netbox_healthcheck_plugin'
    verbose_name = 'NetBox HealthCheck Plugin'
    description = 'NetBox plugin for HealthCheck.'
    version = 'version'
    base_url = 'netbox_healthcheck_plugin'
    django_apps = [
        'health_check',
        'health_check.db',
        'health_check.contrib.migrations',
        'health_check.contrib.redis',
    ]
    min_version = "v4.0-beta1"
    events_pipeline = ["netbox_healthcheck_plugin.events.process_healthcheck_queue"]

    def ready(self):
        from .jobs import MyHousekeepingJob
        # MyHousekeepingJob.setup()

config = HealthCheckConfig

```
I was expecting the run( to get hit every minute, but I wasn't ever seeing it run. Could be operator error.

arthanson (changes requested):

docs/plugins/development/background-jobs.md
alehaa: The documentation for `setup()` was fixed in 39e5c0d24c700d251ecf5a49c2bcd068d09472b6. This was left over from removing system background jobs from the original FR.

`HealthCheckConfig.ready()` is not needed. The job should be enqueued immediately when the `rqworker` process starts. Did you restart the worker after update?

netbox/netbox/utils.py
```diff
@@ -24,3 +25,14 @@ def _wrapper(cls):
...
 
     return _wrapper
+
+
+def register_system_job():
+    """
+    Decorator for registering a `JobRunner` class as system background job.
+    """
+    def _wrapper(cls):
+        registry['system_jobs'][cls.name] = cls
```
jeremystretch, line 35: We probably want to namespace these by app or Python path. Otherwise, we risk a name collision among jobs from different plugins.

netbox/netbox/utils.py
```diff
@@ -24,3 +25,14 @@ def _wrapper(cls):
         return cls
 
     return _wrapper
+
+
+def register_system_job():
```
jeremystretch, line 30: What do you think about passing the interval (and potentially other parameters) via the `register_system_job()` wrapper instead of defining them inside the class? They would need to be stored in the registry alongside a reference to the class itself, but it might allow for more flexible reuse of jobs. You could do something like:

```python
register_system_job(interval=HOURLY, do_big_stuff=False)(MyHousekeepingJob)
register_system_job(interval=DAILY, do_big_stuff=True)(MyHousekeepingJob)
```

Just a thought; I might also be overthinking this. Happy to merge as-is.

netbox/core/management/commands/rqworker.py
```diff
@@ -14,6 +17,16 @@ class Command(_Command):
     of only the 'default' queue).
     """
     def handle(self, *args, **options):
+        # Setup system jobs.
+        for job in registry['system_jobs'].values():
+            if getattr(job.Meta, 'system_enabled', True):
```
jeremystretch, line 22: Could we do away with the `system_enabled` attribute and just check whether `system_interval` is defined? The former seems redundant since we're already registering system jobs automatically.

jeremystretch (changes requested):

netbox/netbox/utils.py
alehaa: I can see the potential in this approach. Would it be okay if the decorator is the only way to register a system job even for plugins instead of listing them in the `system_jobs` variable at the end of the file? `@register_model_view` uses the same approach, but other APIs like data source registration do not allow direct access to the decorator.

jeremystretch: I think personally I prefer the decorator approach but I don't have a strong opinion. It would probably make sense to encourage the direct use of the decorator for data backend registration in plugins as well. @arthanson @bctiemann @DanSheps do any of you have thoughts here?

jeremystretch: We discussed this today and opted to go with the decorator approach as it feels "cleaner" and offers more flexibility.

docs/plugins/development/background-jobs.md
alehaa: As registration now uses a decorator, I think this may be outdated.

netbox/core/management/commands/rqworker.py
```diff
@@ -14,6 +16,15 @@ class Command(_Command):
...
     """
     def handle(self, *args, **options):
+        # Setup system jobs.
+        for job, kwargs in registry['system_jobs'].items():
+            try:
+                interval = kwargs['interval']
+            except KeyError:
+                raise TypeError("System job must specify an interval (in minutes).")
+            logger.debug(f"Scheduling system job {job.name} (interval={interval})")
+            job.enqueue_once(**kwargs)
```
alehaa, line 26: From 6b192c9ae851fba587b43aed845f0a46b8230dfc there is a missing `if interval:`.

The reason for this `if` is to selectively ignore job registration, e.g. depending on configuration parameters. An example use case would be not to register the changelog cleanup job if the `CHANGELOG_RETENTION` setting is zero (disabled). If there are other ways to accomplish this (maybe by another decorator before `system_jon()`?) I'm open to that solution as well.

netbox/netbox/jobs.py
```diff
@@ -2,19 +2,38 @@
...
 from netbox.constants import ADVISORY_LOCK_KEYS
+from netbox.registry import registry
 
 __all__ = (
     'JobRunner',
+    'system_job',
 )
 
 
+def system_job(interval):
```
alehaa, line 21: I suggest adding `*args, **kwargs` as arguments, storing them in the registry and passing them to `enqueue_once()`. This would allow plugins to reuse jobs for different actions. Practical example: Changelog and journal cleanup could use the same job and pass model/configuration parameter names as arguments.

netbox/core/management/commands/rqworker.py
jeremystretch: Seems like a pretty rare corner case IMO. But in such a case, the job registration could be made conditional. For example:

````python
if CHANGELOG_RETENTION:
    system_job(interval=1440)(HouskeepingJob)
```

netbox/netbox/jobs.py
jeremystretch: We want to avoid caching arbitrary data in the registry, as it needs to be constructed (consuming memory) for each running process. Creating multiple subclasses of a base Job class is probably a more scalable approach.

netbox/core/management/commands/rqworker.py
DanSheps: IMO, I would say the better way to do it is as Jeremy mentioned, or even have the worker itself check to see if it needs to run and if not it can just exit cleanly.

PR #17717, opened 2024-10-09
15093 Allow plugins to register events pipeline

### LINKED ISSUE #15093: Plugin events pipeline registration
### NetBox version

v3.7.2

### Feature type

New functionality

### Proposed functionality

3.7.0 introduced the events pipeline described in https://github.com/netbox-community/netbox/issues/14132

This involves making an entry in `EVENTS_PIPELINE=[]` within `configuration.py`, it would be more ergonomic if plugins could register for events by declaring intent within their configuration.

For example:

```python
from netbox.plugins import PluginConfig

class FooBarConfig(PluginConfig):
    name = 'foo_bar'
    verbose_name = 'Foo Bar'
    description = 'An example NetBox plugin'
    version = '0.1'
    author = 'Jeremy Stretch'
    author_email = 'author@example.com'
    base_url = 'foo-bar'
    required_settings = []
    default_settings = {
        'baz': True
    }
    events_pipeline_registrations = [
         "foobar.mymodule.custom_event_handler"
    ]
    django_apps = ["foo", "bar", "baz"]

config = FooBarConfig
```

`events_pipeline_registrations` would be optional to allow for backwards compatibility and default to `[]`

It would allow plugins which require access to the events pipeline to have entries added into `EVENTS_PIPELINE` automatically. This reduces complexity on those installing Netbox plugins which need events access.

https://linear.app/netboxlabs/issue/BIZ-6/

### Use case

Simplify access to the events pipeline for plugins, reduce configuration complexity.

### Database changes

_No response_

### External dependencies

_No response_

arthanson:


docs/configuration/index.md
```diff
@@ -25,6 +25,7 @@ Some configuration parameters are primarily controlled via NetBox's admin interf
 * [`CUSTOM_VALIDATORS`](./data-validation.md#custom_validators)
 * [`DEFAULT_USER_PREFERENCES`](./default-values.md#default_user_preferences)
 * [`ENFORCE_GLOBAL_UNIQUE`](./miscellaneous.md#enforce_global_unique)
+* [`EVENTS_PIPELINE`](./miscellaneous.md#events_pipeline)
```
jeremystretch, line 28: This isn't a [dynamic config parameter](https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/config/parameters.py) (one which can be modified via the UI).
```suggestion
```

docs/configuration/miscellaneous.md
```diff
@@ -106,6 +106,16 @@ By default, NetBox will prevent the creation of duplicate prefixes and IP addres
 
 ---
 
+## EVENTS_PIPELINE
+
+!!! tip "Dynamic Configuration Parameter"
```
jeremystretch, line 111:
```suggestion
```

docs/plugins/development/index.md
```diff
@@ -114,6 +114,7 @@ NetBox looks for the `config` variable within a plugin's `__init__.py` to load i
 | `max_version`         | Maximum version of NetBox with which the plugin is compatible                                                            |
 | `middleware`          | A list of middleware classes to append after NetBox's build-in middleware                                                |
 | `queues`              | A list of custom background task queues to create                                                                        |
+| `events_pipeline`     | An optional list of functions to add to the [`EVENTS_PIPELINE`](./miscellaneous.md#events_pipeline)                                                      |
```
jeremystretch, line 117: Specify dotted paths

docs/configuration/miscellaneous.md
```diff
@@ -106,6 +106,16 @@ By default, NetBox will prevent the creation of duplicate prefixes and IP addres
...
+
+Default: ['extras.events.process_event_queue',]
+
+NetBox will call functions listed here for events (create, update, delete) on models as well as when custom EventRules are fired.
```
jeremystretch, line 115: The description should specify passing a list of dotted paths (rather than the functions themselves).

netbox/netbox/settings.py
```diff
@@ -112,9 +112,9 @@
 DEVELOPER = getattr(configuration, 'DEVELOPER', False)
 DOCS_ROOT = getattr(configuration, 'DOCS_ROOT', os.path.join(os.path.dirname(BASE_DIR), 'docs'))
 EMAIL = getattr(configuration, 'EMAIL', {})
-EVENTS_PIPELINE = getattr(configuration, 'EVENTS_PIPELINE', (
+EVENTS_PIPELINE = getattr(configuration, 'EVENTS_PIPELINE', [
     'extras.events.process_event_queue',
-))
+])
```
jeremystretch, line 117: I think we want to ensure that `process_event_queue()` is always included in the pipeline; otherwise, event rules would break. The config parameter should be treated as a list of _additional_ items to include. (I realize this deviates from how it's currently handled.)

jeremystretch (changes requested):

netbox/netbox/settings.py
```diff
@@ -859,6 +859,14 @@ def _setting(name, default=None):
         f"{plugin_name}.{queue}": RQ_PARAMS for queue in plugin_config.queues
     })
 
+    events_pipeline = plugin_config.events_pipeline
+    if events_pipeline and type(events_pipeline) in (list, tuple):
```
jeremystretch, line 863: An ImproperlyConfigured exception should be raised if `plugin_config.events_pipeline` is not an iterable.

netbox/netbox/settings.py
```diff
@@ -859,6 +859,14 @@ def _setting(name, default=None):
...
+        EVENTS_PIPELINE = list(EVENTS_PIPELINE)
```
jeremystretch, line 864: Constants should never be redefined. Any validation of the initial `EVENTS_PIPELINE` config parameter should be performed prior to plugin registration.

netbox/netbox/settings.py
```diff
@@ -859,6 +859,14 @@ def _setting(name, default=None):
...
+        if 'extras.events.process_event_queue' not in EVENTS_PIPELINE:
+            EVENTS_PIPELINE.insert(0, 'extras.events.process_event_queue')
```
jeremystretch, line 866: This doesn't need to be checked for every plugin.

jeremystretch (changes requested):
Please also add an event handler to the [dummy plugin](https://github.com/netbox-community/netbox/tree/develop/netbox/netbox/tests/dummy_plugin) and a simple test to ensure that it gets registered correctly.

netbox/netbox/settings.py
arthanson: ? We are doing that all over settings.py, just a few lines above we are doing the same to INSTALLED_APPS.  Previously this was assigned as a tuple and could be in config.py, so that is why had it convert to a list.

jeremystretch (approved):
Thanks @arthanson!

PR #17731, opened 2024-10-10
17614 Disallow removal of virtual chassis from device if set as master

### LINKED ISSUE #17614: Virtual Chassis still has master after member removed
### Deployment Type

Self-hosted

### NetBox Version

v4.1.1

### Python Version

3.12

### Steps to Reproduce

1. Create new virtual chassis
2. Create new device and assign to virtual chassis
3. Assign device as master of virtual chassis
4. Edit the device and clear the virtual chassis
5. View the virtual chassis and confirm the member is gone but the master remains

### Expected Behavior

Other pages seem to prevent the member being from being removed while it is the master, so assumedly this behaviour should be consistent from the device edit page.

### Observed Behavior

Master exists in virtual chassis without it being a member.

arthanson:
### Fixes: #17614 

Raise error if removing the virtual chassis from a device if it is set as master.  We could auto-remove it from master on the virtual-chassis, but it seems like in other places we disallow the removal so keeping it consistent.

### DIFF
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -983,7 +983,7 @@ class Device(
                 'vc_position': _("A device assigned to a virtual chassis must have its position defined.")
             })
 
-        if vc_master_for := getattr(self, 'vc_master_for', None) != self.virtual_chassis:
+        if hasattr(self, 'vc_master_for') and self.vc_master_for and self.vc_master_for != self.virtual_chassis:
             raise ValidationError({
                 'virtual_chassis': _('Device cannot be removed from virtual chassis {virtual_chassis} because it is currently designated as its master.').format(
                     virtual_chassis=self.vc_master_for
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -983,6 +983,13 @@ class Device(
                 'vc_position': _("A device assigned to a virtual chassis must have its position defined.")
             })
 
+        if hasattr(self, 'vc_master_for') and self.vc_master_for and self.vc_master_for != self.virtual_chassis:
+            raise ValidationError({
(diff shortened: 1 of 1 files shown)


netbox/dcim/models/devices.py
```diff
@@ -983,6 +983,13 @@ def clean(self):
                 'vc_position': _("A device assigned to a virtual chassis must have its position defined.")
             })
 
+        if hasattr(self, 'vc_master_for') and self.vc_master_for and self.vc_master_for != self.virtual_chassis:
```
bctiemann, line 986: This can be condensed to something like
```
if vc_master_for := getattr(self, 'vc_master_for') != self.virtual_chassis:
```

Unless `self.virtual_chassis` can be `None`, in which case disregard this

bctiemann (approved):

netbox/dcim/models/devices.py
```diff
@@ -983,6 +983,13 @@ def clean(self):
...
+            raise ValidationError({
+                'virtual_chassis': _("Cannot change to a different virtual chassis when assigned as a master. Remove it as the master of {master} first.").format(
```
jeremystretch, line 988: Can we tweak the error message to avoid confusing the user? (We're not necessarily moving the device to a _different_ virtual chassis; we may just be removing it.) Something like "Device cannot be removed from virtual chassis {virtual_chassis} because it is currently designated as its master."

jeremystretch (changes requested):

netbox/dcim/models/devices.py
arthanson: Tried a couple different permutations but it was breaking when there was no virtual_chassis / no master, so kept it as is at it works in all cases.

jeremystretch (approved):

PR #17764, opened 2024-10-15
17635 fix script AbortTransaction

### LINKED ISSUE #17635: Report raises AbortTransaction() when commit=False and log_failure() has been used
### Deployment Type

Self-hosted

### NetBox Version

v4.1.2

### Python Version

3.11

### Steps to Reproduce

Create a script that logs a failure. Run the script with commit=False
```

class LogFailure(Script):
  description = "Log one failure"

  def test_error_behavior(self):
    self.log_failure("This is a test failure, not a script failure")
```

### Expected Behavior

The script will log one 'Failure' line, rollback database changes, and complete

### Observed Behavior

The script logs appropriate, but adds an additional failure, as show below, before rolling back DB changes
```
An exception occurred: AbortTransaction:

  File "/opt/netbox/netbox/extras/jobs.py", line 47, in run_script
    raise AbortTransaction()
utilities.exceptions.AbortTransaction
```

arthanson:

### DIFF
--- a/netbox/extras/jobs.py
+++ b/netbox/extras/jobs.py
@@ -51,6 +51,9 @@ class ScriptJob(JobRunner):
                     logger.warning("Script failed")
                     raise
 
+        except AbortTransaction:
+            # this is already handled above.
+            pass
         except Exception as e:
             if type(e) is AbortScript:
                 msg = _("Script aborted with error: ") + str(e)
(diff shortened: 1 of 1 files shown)


bctiemann (approved):

netbox/extras/jobs.py
```diff
@@ -51,6 +51,9 @@ def run_script(self, script, request, data, commit):
                     logger.warning("Script failed")
                     raise
 
+        except AbortTransaction:
+            # this is already handled above.
+            pass
```
jeremystretch, line 56: This doesn't need to be caught again; we can just avoid re-raising it on line 52.

jeremystretch (changes requested):

jeremystretch (approved):

PR #17765, opened 2024-10-15
Fixes #17749: Add missing graphql fields

### LINKED ISSUE #17749: Missing GraphQL fields
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.3

### Python Version

3.12

### Steps to Reproduce

1. Open the GraphQL page and execute the following query:
```
{
  device_list { devicebays { id } }
}
```

### Expected Behavior

I receive a response with a list of all (if any) devices and no errors. This is an example from a v3.7.8 instance with no devices:
```
{
  "data": {
    "device_list": []
  }
}
```

### Observed Behavior

I receive the following response/error:
```
{
  "data": null,
  "errors": [
    {   
      "message": "Cannot query field 'devicebays' on type 'DeviceType'. Did you mean 'device_type', 'modulebays', or 'services'?",
      "locations": [
        {
          "line": 2,
          "column": 18
        }
      ]
    }
  ]
}
```

<hr>

When I first ran into this, I traced it back to the GraphQL engine switch in the v4.0 release. And because "if it happens once, it may happened twice" I grabbed the GraphQL schemas from a fresh v3.7.8 and v4.1.3 instance, and quickly wrote a script to find all fields and arguments that no longer exist; which were a lot. I had to account for general differences how these engines represent stuff, and filter out all the intended changes (like `devicetype_id` being renamed to `device_type_id`). In the end, this handful of unaccounted-for field removals was left:

```
Type ContactGroupType lost 1 field(s): children:List[ContactGroupType]
Type DeviceType lost 1 field(s): devicebays:List[DeviceBayType]
Type RackType lost 1 field(s): powerfeed_set:List[PowerFeedType]
    and gained 1 field(s): powerfeeds:List[PowerFeedType]
Type SiteType lost 1 field(s): asn:BigInt
Type TenantGroupType lost 1 field(s): children:List[TenantGroupType]
Type WirelessLANGroupType lost 1 field(s): children:List[WirelessLANGroupType]
```

While I found no matching issue for the `Rack` change, it looks like a renaming that happened as part of the engine switch, and probably is fine at it is. I am not sure what the `Site.asn` field was meant to be, because there also is a `Site.asns: List[ASN]` field; this one is probably also fine as it is. `Device` missing the `devicebays` field is the one I initially ran into, which I definitely consider a bug. And then there are the three tree-building self-referential `ContactGroup`, `TenantGroup` and `WirelessLANGroup` which are missing their respective `children` field; I would also consider this a bug, since others like `Location` still have their `children` field.

corubba:
### Fixes: #17749

Re-add GraphQL fields which were unintentionally removed during the GraphQL engine switch.

### DIFF
--- a/netbox/dcim/graphql/types.py
+++ b/netbox/dcim/graphql/types.py
@@ -243,6 +243,7 @@ class DeviceType(ConfigContextMixin, ImageAttachmentsMixin, ContactsMixin, NetBo
     consoleserverports: List[Annotated["ConsoleServerPortType", strawberry.lazy('dcim.graphql.types')]]
     poweroutlets: List[Annotated["PowerOutletType", strawberry.lazy('dcim.graphql.types')]]
     frontports: List[Annotated["FrontPortType", strawberry.lazy('dcim.graphql.types')]]
+    devicebays: List[Annotated["DeviceBayType", strawberry.lazy('dcim.graphql.types')]]
     modulebays: List[Annotated["ModuleBayType", strawberry.lazy('dcim.graphql.types')]]
     services: List[Annotated["ServiceType", strawberry.lazy('ipam.graphql.types')]]
     inventoryitems: List[Annotated["InventoryItemType", strawberry.lazy('dcim.graphql.types')]]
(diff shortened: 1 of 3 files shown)


jeremystretch (approved):
Thanks @corubba!

PR #17766, opened 2024-10-15
17754 fix per-page on version history

### LINKED ISSUE #17754: The "Version History" pane for plugins is broken
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.3

### Python Version

3.12

### Steps to Reproduce

1. Open the "Admin/System/Plugins" page via the navigation menu
2. Click on any plugin (e.g. NetBox BGP)
3. In the "Version History" pane, select the "Per Page" dropdown and klick on "25" (any other value is OK, too)

### Expected Behavior

Actually I would expect the paginator not to be there at all, as there are only the five most recent versions in the API so there won't be a longer list of versions anyway.

But since a paginator **is** shown, I would expect the last 25, 50, 100 ... most recent versions to appear in the "Version History" table.

### Observed Behavior

A narrow version of the home page appears in the pane.

Screenshot 2024-10-14 at 19 11 07

arthanson:

### DIFF
--- a/netbox/core/views.py
+++ b/netbox/core/views.py
@@ -715,6 +715,12 @@ class PluginView(BasePluginView):
         table = PluginVersionTable(plugin.release_recent_history, user=request.user)
         table.configure(request)
 
+        # If this is an HTMX request, return only the rendered table HTML
+        if htmx_partial(request):
+            return render(request, 'htmx/table.html', {
+                'table': table,
+            })
+
         return render(request, 'core/plugin.html', {
             'plugin': plugin,
             'table': table,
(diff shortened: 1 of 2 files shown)


bctiemann (approved):

netbox/core/views.py
```diff
@@ -715,6 +715,12 @@ def get(self, request, name):
         table = PluginVersionTable(plugin.release_recent_history, user=request.user)
         table.configure(request)
 
+        # If this is an HTMX request, return only the rendered table HTML
```
jeremystretch, line 718: Do we need to support a dynamic table here? It might be simpler to just render the static table without a paginator on the plugin page. It looks like `release_recent_history` is capped at five items anyway.

jeremystretch (changes requested):

jeremystretch (approved):

PR #17770, opened 2024-10-16
Fixes #17759: Add job timeout handling in JobRunner for periodic jobs

### LINKED ISSUE #17759: Custom scripts with `job_timeout` more than `RQ_DEFAULT_TIMEOUT` failed when rescheduled
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.3

### Python Version

3.12

### Steps to Reproduce

1. Create script with `job_timeout` more than `RQ_DEFAULT_TIMEOUT`, for example, the default timeout is 300 secs, setting `job_timeout` to 500 and inside script only sleep for 400 secs

```python
from time import sleep

class TestScript(Script):
    class Meta:
        job_timeout = 500

    def run(self, data, commit):
        sleep(400)
```

2. Run script with interval, for example 3 minutes

### Expected Behavior

Script runs successfully and all scheduled after scripts are successful too

### Observed Behavior

First running will be successful, but after that all scheduled jobs will fail with `JobTimeoutException`

```
[DJANGO] ERROR 2024-10-15 11:45:28,255 scripts 160778 140373397688448 An exception occurred: `JobTimeoutException: Task exceeded maximum timeout value (300 seconds)`

  File "/home/miaow/work/netbox/netbox/extras/jobs.py", line 45, in run_script
                    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/miaow/work/netbox/netbox/scripts/test_script.py", line 11, in run
    sleep(400)
  File "/home/miaow/work/netbox/venv/lib/python3.12/site-packages/rq/timeouts.py", line 63, in handle_death_penalty
    raise self._exception('Task exceeded maximum timeout value ({0} seconds)'.format(self._timeout))
rq.timeouts.JobTimeoutException: Task exceeded maximum timeout value (300 seconds)
```

This happens because `ScriptView` [passes](https://github.com/netbox-community/netbox/blob/develop/netbox/extras/views.py#L1182) `job_timeout` into `ScriptJob`,

but rescheduling mechanism does not [pass](https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/jobs.py#L71) this argument into `enqueue` method

miaow2:
### Fixes: #17759
Add check if job has assigned script object with job timeout

### DIFF
--- a/netbox/netbox/jobs.py
+++ b/netbox/netbox/jobs.py
@@ -68,6 +68,8 @@ class JobRunner(ABC):
         finally:
             if job.interval:
                 new_scheduled_time = (job.scheduled or job.started) + timedelta(minutes=job.interval)
+                if job.object and getattr(job.object, "python_class", None):
+                    kwargs["job_timeout"] = job.object.python_class.job_timeout
                 cls.enqueue(
                     instance=job.object,
                     user=job.user,


jeremystretch (approved):
Thanks @miaow2!

PR #17775, opened 2024-10-16
17464 fix margins for custom-field markdown description

### LINKED ISSUE #17464: reduce custom field padding when using descriptions
### Deployment Type

Self-hosted

### NetBox Version

v.4.1.0

### Python Version

3.10

### Steps to Reproduce

create 2 custom fields
give both custom fields a description

### Expected Behavior

custom field padding should be reduced between custom fields that have descriptions.

currently, when custom fields are next to each other and they don't have a description, the spacing is nice and tight. But as soon as you add a description to the custom field, the padding is too much.

### Observed Behavior

image

arthanson:


netbox/project-static/styles/custom/_markdown.scss
```diff
@@ -29,7 +29,7 @@
 }
 
 // Remove the bottom margin of <p> elements inside a table cell
-td > .rendered-markdown {
+td > .rendered-markdown, .form-text .rendered-markdown {
```
jeremystretch, line 32: We should break out the `p:last-of-type` rule as it applies generally, whereas the `max-height` and `overflow-y` rules apply only to Markdown content within a table cell.

jeremystretch (changes requested):

netbox/project-static/styles/custom/_markdown.scss
```diff
@@ -29,13 +29,16 @@
 }
 
 // Remove the bottom margin of <p> elements inside a table cell
```
jeremystretch, line 31: Comments need to be updated for both stanzas

jeremystretch (changes requested):

jeremystretch (approved):

PR #17792, opened 2024-10-16
17374 correct background color in dark mode for active list item

### LINKED ISSUE #17374: Low contrast with active list items (Device Page, Virtual Chassis)
### Deployment Type

Self-hosted

### NetBox Version

v4.1.0

### Python Version

3.10

### Steps to Reproduce

1. Enable the dark mode theme
2. Create a new virtual chassis
3. Add one or more Devices to the Virtual Chassis
4. Open the Device page for any device associated with a Virtual Chassis

### Expected Behavior

The highlighted Device in the Virtual Chassis pane should be readable and more consistent with other dark mode UI elements.

### Observed Behavior

The highlighted Device appears in the list with a background color that makes it very difficult to read.
PP:B117 | NetBox 2024-09-05 07-39-15

arthanson:
### Fixes: #17374 

PP:B117 | NetBox 2024-10-16 13-08-14


jeremystretch (approved):
Thanks for tracking this down @arthanson

PR #17805, opened 2024-10-17
Fixes #17802 - Added opaque background to Rename buttons

### LINKED ISSUE #17802: Update the list view bulk Rename buttons to have an opaque background
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.4

### Python Version

3.10

### Steps to Reproduce

1. Create about 20 devices and navigate to the Device list view
2. Select/check one or more of the rows so the bulk operations buttons float

Templates affected:
1. templates/dcim/device_list.html
2. templates/dcim/component_list.html

These templates were overlooked in #17523. Adding the `btn-float` class to the Rename button fixes the issue.

### Expected Behavior

The Rename button should have an opaque background to avoid clashing with the form list items.

image

### Observed Behavior

The Rename button has a transparent background.

image

atownson:
Fixes: #17802

Added the `btn-float` class to the Rename buttons in the corresponding templates.

### DIFF
--- a/netbox/templates/dcim/component_list.html
+++ b/netbox/templates/dcim/component_list.html
@@ -10,7 +10,7 @@
     {% endif %}
     {% if 'bulk_rename' in actions %}
       {% with bulk_rename_view=model|validated_viewname:"bulk_rename" %}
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @atownson!

PR #17790, opened 2024-10-16
Fixes: #17732 - Add a background-color to img elements in docs to ensure readability in dark mode

bctiemann:
### Fixes: #17732

Adds a `background-color` property to the `img` CSS style for docs, to ensure that PNGs with transparent backgrounds and black text are still readable even on a dark-mode background.

Screenshot 2024-10-16 at 2 18 10 PM

PR #17811, opened 2024-10-18
Update README.md to point to NetBox logo URL

mi3law:
The NetBox logo is referenced at https://github.com/netbox-community/netbox/blob/develop/docs/netbox_logo.svg but that no longer exists as the logo has been changed to https://github.com/netbox-community/netbox/blob/develop/docs/netbox_logo_dark.svg (or _light.svg -- should add handling for that but seems like overkill)

Updated the README to reflect this to properly render the logo.

Pending bug report creation


jeremystretch (approved):
Thanks! I missed this while working on #17790.

PR #17821, opened 2024-10-21
9604 Add Termination to CircuitTermination

arthanson:
### Fixes: #9604

As per review comments: 
* Changed from using scope to termination
* changed _sitegroup to _site_group
* changed related names to not use underscore


bctiemann (approved):

jeremystretch:
I was under the impression that ProviderNetwork would be one of the scope choices, rather than remaining as a standalone, alternative field. Is there a reason not to make this one of the options for the `scope` GFK? (The current `site` and `provider_network` ForeignKey fields are already mutually exclusive.)

docs/models/circuits/circuittermination.md
```diff
@@ -21,13 +21,13 @@ Designates the termination as forming either the A or Z end of the circuit.
 
 If selected, the circuit termination will be considered "connected" even if no cable has been connected to it in NetBox.
 
-### Site
+### Scope
 
-The [site](../dcim/site.md) with which this circuit termination is associated. Once created, a cable can be connected between the circuit termination and a device interface (or similar component).
+The [region](../dcim/region.md), [site](../dcim/site.md) or [location](../dcim/location.md) with which this circuit termination is associated. Once created, a cable can be connected between the circuit termination and a device interface (or similar component).
```
jeremystretch, line 26: Also site group

netbox/circuits/models/circuits.py
```diff
@@ -232,20 +236,22 @@ class CircuitTermination(
...
     )
-    provider_network = models.ForeignKey(
-        to='circuits.ProviderNetwork',
-        on_delete=models.PROTECT,
-        related_name='circuit_terminations',
+    scope_id = models.PositiveBigIntegerField(
         blank=True,
         null=True
     )
+    scope = GenericForeignKey(
```
jeremystretch, line 251: I'm not sure it makes sense to call this "scope;" it's more of a termination point. `termination` might be more apt, and it would be consistent with the CableTermination model.

netbox/circuits/models/circuits.py
```diff
@@ -276,6 +282,43 @@ class CircuitTermination(
...
+        null=True
+    )
+    _region = models.ForeignKey(
+        to='dcim.Region',
+        on_delete=models.CASCADE,
+        related_name='_circuit_terminations',
+        blank=True,
+        null=True
+    )
+    _sitegroup = models.ForeignKey(
```
jeremystretch, line 314: Something I wasn't sure about when working on #6414 was whether to go with `sitegroup` or `site_group`. Whichever naming scheme we choose, we should be consistent with e.g. `provider_network` above.

netbox/circuits/models/circuits.py
```diff
@@ -276,6 +282,43 @@ class CircuitTermination(
...
+        to='circuits.ProviderNetwork',
+        on_delete=models.PROTECT,
+        related_name='circuit_terminations',
+        blank=True,
+        null=True
+    )
+    _location = models.ForeignKey(
+        to='dcim.Location',
+        on_delete=models.CASCADE,
+        related_name='_circuit_terminations',
```
jeremystretch, line 296: Should the reverse relations have the leading underscore? (`related_name` for `_provider_network` above does not.)

netbox/circuits/tables/circuits.py
```diff
@@ -112,7 +112,8 @@ class CircuitTerminationTable(NetBoxTable):
     )
     site = tables.Column(
         verbose_name=_('Site'),
-        linkify=True
+        linkify=True,
+        accessor='_site'
     )
     provider_network = tables.Column(
```
jeremystretch, line 118: This needs its accessor updated as well.

jeremystretch, line 118: We should add columns for region, site group, and location as well.

jeremystretch (changes requested):

jeremystretch (approved):

gjvc:
> ### Fixes: #9604
> As per review comments:
> 
> * Changed from using scope to termination
> * changed _sitegroup to _site_group
> * changed related names to not use underscore

Does this actually add a "Location" to the "Circuit Termination" (which was the intent of #9604) ?

gjvc:
Great to see this complete, thank you all very much!

PR #17822, opened 2024-10-21
Closes #13428: Q-in-Q VLANs

### LINKED ISSUE #13428: QinQ - Basic initial functionality (no vlan translation)
### NetBox version

v3.5.7

### Feature type

Data model extension

### Proposed functionality

There would be two new sections under IPAM, svlan + svlan groups
When editing an interface there would be an additional option under 802.1 mode (Double tagged) 
When selected two additional fields would appear (svlan group + svlan) 
 
When viewing an svlan object it should list any interfaces that have that svlan assigned  
 
on the API the interface model should include the svlan object 

### Use case

This would allow for basic documentation of basic QinQ uses 

NOTE: im excluding vlan translation for the initial feature request: 
 - IMHO a larger PR needs to be had for general vlan translation options + ui features to allow cross-referencing objects)
 - In a lot of cases the interface name will be a hint to the user if there needs to be vlan translation happening and could be handled in automation tooling (which is already the way many folks handle this right now)

Documenting that svlan's exist and showing that an svlan exists on an interface would be a major improvement.

### Database changes

new svlan + svlan groups model (an exact clone of vlan + vlan groups models)

Interfaces model: 
 - Additional option under mode (double tagged) 
 - New fields: 
   - Svlan (foreign key to svlan) 

### External dependencies

N/A

jeremystretch:
### Closes: #13428

- Add Q-in-Q to InterfaceModeChoices
- Move `untagged_vlan` & `tagged_vlans` model fields to BaseInterface
- Add `qinq_svlan` field to BaseInterface
- Add `qinq_role` & `qinq_svlan` fields to VLAN


netbox/dcim/models/device_components.py
```diff
@@ -547,10 +547,41 @@ class BaseInterface(models.Model):
...
+        blank=True,
+        verbose_name=_('tagged VLANs')
+    )
+    qinq_svlan = models.ForeignKey(
+        to='ipam.VLAN',
+        on_delete=models.SET_NULL,
+        related_name='%(class)ss_svlan',
+        null=True,
+        blank=True,
+        verbose_name=_('Q-inQ SVLAN')
```
arthanson, line 570: Q-in-Q (needs an extra dash

netbox/dcim/tests/test_filtersets.py
```diff
@@ -4016,6 +4030,13 @@ def test_vdc_identifier(self):
         params = {'vdc_identifier': vdc.values_list('identifier', flat=True)}
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 3)
 
+    def test_vlan(self):
+        vlan = VLAN.objects.filter(qinq_role=VLANQinQRoleChoices.ROLE_SERVICE).first()
+        params = {'vlan_id': vlan.pk}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
+        params = {'vlan': vlan.vid}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
+
```
arthanson, line 4040: Should probably be a test for creating a service VLAN assigned to another Service VLAN to make sure it get error.

netbox/dcim/models/device_components.py
```diff
@@ -547,10 +547,41 @@ class BaseInterface(models.Model):
...
+    )
 
     class Meta:
         abstract = True
 
+    def clean(self):
```
arthanson, line 576: If a VLAN is assigned to a Site should it restrict what Q-in-Q SVLAN it can be assigned to (i.e. in the same Site)?  Not sure if that makes sense or is overkill.

netbox/ipam/graphql/types.py
```diff
@@ -250,6 +250,10 @@ class VLANType(NetBoxObjectType):
     interfaces_as_tagged: List[Annotated["InterfaceType", strawberry.lazy('dcim.graphql.types')]]
     vminterfaces_as_tagged: List[Annotated["VMInterfaceType", strawberry.lazy('virtualization.graphql.types')]]
 
+    @strawberry_django.field
```
arthanson, line 253: Can this just be done as above in one line, no need for a function?

netbox/ipam/models/vlans.py
```diff
@@ -202,6 +202,21 @@ class VLAN(PrimaryModel):
         null=True,
         help_text=_("The primary function of this VLAN")
     )
+    qinq_svlan = models.ForeignKey(
+        to='self',
+        on_delete=models.PROTECT,
+        related_name='qinq_cvlans',
```
arthanson, line 208: What is cvlans? not sure if this was a typo, or customer - any harm in expanding it out? actually can it be quinq_service_vlan to be more friendly as to what it is instead of abbreviation?

netbox/ipam/models/vlans.py
```diff
@@ -253,9 +276,18 @@ def clean(self):
                     ).format(ranges=ranges_to_string(self.group.vid_ranges), group=self.group)
                 })
 
+        # Only Q-in-Q customer VLANs may be assigned to a service VLAN
```
arthanson, line 279: Can also set Q-in-Q role without setting any Q-in-Q SVLAN, does that make any sense?

arthanson (changes requested):

netbox/dcim/models/device_components.py
jeremystretch: It might be reasonable, but I want to wait until we get some feedback on the initial implementation before making that determination.

netbox/ipam/graphql/types.py
jeremystretch: I think it needs to be a function because the type is being used recursively.

netbox/ipam/models/vlans.py
jeremystretch: CVLAN (customer) and SVLAN (service) are pretty standard terms in the context of Q-in-Q. I think it's okay to leave them as-is, but we should be sensitive to any user feedback regarding terminology during the beta.

bctiemann (approved):

PR #17847, opened 2024-10-23
Fixes #17685 : Fix job field validation

### LINKED ISSUE #17685: Job.clean() not called
### Deployment Type

Self-hosted

### NetBox Version

v4.1.3

### Python Version

3.11

### Steps to Reproduce

1. Create a new `JobRunner`.
2. Schedule the job runner without with `enqueue(interval=0)`.

### Expected Behavior

Getting a `ValidationError`: 

```Python
django.core.exceptions.ValidationError: {
  '__all__': [
    'Jobs cannot be assigned to this object type (None).'
  ]
}
```

### Observed Behavior

Nothing happens, as neither `Job.enqueue()`, `Job.start()` nor `Job.terminate()` trigger `Job.clean()`.

alehaa:
### Fixes: #17685

Previously, fields in the Job model were not validated when the job was created. Now `full_clean()` is called before saving the job to ensure valid data. In addition, test cases have been updated because using a `Job` as instance when creating a job would not pass validation.

### DIFF
--- a/netbox/core/models/jobs.py
+++ b/netbox/core/models/jobs.py
@@ -130,7 +130,7 @@ class Job(models.Model):
         super().clean()
 
         # Validate the assigned object type
-        if self.object_type not in ObjectType.objects.with_feature('jobs'):
+        if self.object_type and self.object_type not in ObjectType.objects.with_feature('jobs'):
             raise ValidationError(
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @alehaa!

PR #17848, opened 2024-10-23
7699 Add Scope to Cluster

### LINKED ISSUE #7699: Assign virtualization Clusters to a Location
### NetBox version

v3.0.8

### Feature type

Data model extension

### Proposed functionality

Virtualization clusters should be assignable to locations.

When the location function was added, clusters missed this option.

The open issue #6429 is similar and is marked as blocked by #6440 for better handling of "multi-model parent assignments". The same issue may affect this feature but I feel the use case is quite strong.

### Use case

Because Clusters are only assignable to Sites and not to Locations, the whole Locations feature is currently unusable for documentation that consist of both physical Devices and Virtual Machines.

It makes integrations very difficult if the physical location of Devices uses the Locations feature but location of Virtual Machines uses the Sites feature. So in practice both Devices and Virtual Machines then need to use Site for the physical location.

Adding Location for Clusters would allow to locate both Devices and Virtual Machines using Location.

### Database changes

Virtualization Clusters would need a parent Location.

### External dependencies

none

arthanson:
### Fixes: #7699

Added mixins as well, not sure if they should be in netbox/ or dcim/ but the filtersets would need to be in something like filterset_mixins, so not sure if that is good. Can move them if desired.

Once this is approved will make a separate PR to move the already implemented scoped models to use these mixins.


docs/models/virtualization/cluster.md
```diff
@@ -23,6 +23,6 @@ The cluster's operational status.
 !!! tip
     Additional statuses may be defined by setting `Cluster.status` under the [`FIELD_CHOICES`](../../configuration/data-validation.md#field_choices) configuration parameter.
 
-### Site
+### Scope
 
-The [site](../dcim/site.md) with which the cluster is associated.
+The [region](../dcim/region.md), [site](../dcim/site.md) or [location](../dcim/location.md) with which this cluster is associated.
```
jeremystretch, line 28: Also site group

netbox/dcim/graphql/types.py
```diff
@@ -726,9 +726,12 @@ class SiteType(VLANGroupsMixin, ImageAttachmentsMixin, ContactsMixin, NetBoxObje
     locations: List[Annotated["LocationType", strawberry.lazy('dcim.graphql.types')]]
     asns: List[Annotated["ASNType", strawberry.lazy('ipam.graphql.types')]]
     circuit_terminations: List[Annotated["CircuitTerminationType", strawberry.lazy('circuits.graphql.types')]]
-    clusters: List[Annotated["ClusterType", strawberry.lazy('virtualization.graphql.types')]]
     vlans: List[Annotated["VLANType", strawberry.lazy('ipam.graphql.types')]]
 
+    @strawberry_django.field
+    def clusters(self) -> List[Annotated["ClusterType", strawberry.lazy('virtualization.graphql.types')]]:
```
jeremystretch, line 732: We should add this for regions, site groups, and locations as well.

netbox/dcim/models/devices.py
```diff
@@ -958,10 +958,10 @@ def clean(self):
                 })
 
         # A Device can only be assigned to a Cluster in the same Site (or no Site)
-        if self.cluster and self.cluster.site is not None and self.cluster.site != self.site:
+        if self.cluster and self.cluster._site is not None and self.cluster._site != self.site:
```
jeremystretch, line 961: This should be extended to validate against the cluster's assigned region/site group/location as well.

netbox/ipam/querysets.py
```diff
@@ -148,7 +148,7 @@ def get_for_virtualmachine(self, vm):
 
         # Find all relevant VLANGroups
         q = Q()
-        site = vm.site or vm.cluster.site
+        site = vm.site or vm.cluster._site
```
jeremystretch, line 151: This should be adapted for the other scope types as well, as VLAN groups can also be assigned to different scopes.

netbox/netbox/filtersets.py
```diff
@@ -325,3 +327,60 @@ def search(self, queryset, name, value):
             models.Q(slug__icontains=value) |
             models.Q(description__icontains=value)
         )
+
+
+class ScopedFilterSet(BaseFilterSet):
```
jeremystretch, line 332: This should probably live in the `dcim` app. This module is intended for filtersets pertaining to the base model classes.

netbox/virtualization/tables/clusters.py
```diff
@@ -75,7 +75,8 @@ class ClusterTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
     )
     site = tables.Column(
         verbose_name=_('Site'),
-        linkify=True
+        linkify=True,
+        accessor='_site'
     )
```
jeremystretch, line 80: We probably want to replace this with `scope_type` & `scope` columns like we did for PrefixTable.

jeremystretch (changes requested):

netbox/ipam/querysets.py
arthanson: ? Not sure I follow this - this is for VLAN, not VLANGroup and any higher scope assignment will filter down to _site - so _site should pick up all the assignments?

netbox/virtualization/models/clusters.py
```diff
@@ -107,8 +118,8 @@ class Meta:
                 name='%(app_label)s_%(class)s_unique_group_name'
             ),
             models.UniqueConstraint(
-                fields=('site', 'name'),
-                name='%(app_label)s_%(class)s_unique_site_name'
+                fields=('_site', 'name'),
+                name='%(app_label)s_%(class)s_unique__site_name'
             ),
```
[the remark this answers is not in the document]
arthanson: I think this still works as we set all the underscored scope vars if a higher one is set in the chain region -> site group -> site -> location.  So if location is set then _site is also set correctly.  This would make the name only needing to unique to a Site and not a region, which seems to be still valid?  Or is it just un-necessary now?

arthanson:
FYI: I don't think the build breaking is from the PR, looks like from background task running...

prryplatypus:
> FYI: I don't think the build breaking is from the PR, looks like from background task running...

I've opened #17879 with the reason behind the failure.

netbox/virtualization/models/clusters.py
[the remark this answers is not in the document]
jeremystretch: I was thinking more re: location assignment, but I think you're right. We can leave it as-is.

jeremystretch (changes requested):

jeremystretch (approved):

PR #17877, opened 2024-10-28
10711 Add Scope to WirelessLAN

arthanson:


bctiemann (approved):

netbox/wireless/tests/test_views.py
jeremystretch, line 1: We should extend the CSV test data to test bulk import assignment as well.

jeremystretch (changes requested):
The wireless LAN HTML template needs to be extended to show the assigned scope.

This also needs API tests.

jeremystretch (approved):
I've extended the CSV test data.

Issue #4867, opened 2020-07-18
Multiple MAC addresses on interface


lps-rocks:
### Environment
* Python version:  3.7.6
* NetBox version:  2.8.7

### Proposed Functionality

Allow for multiple MAC addresses to be specified per physical interface. Enable the ability to tag a mac address as physical or virtual. Physical mac addresses are permanent and are unique in Netbox, Virtual mac addresses may be duplicated across devices in Netbox. 

### Use Case

In situations where devices use 'Virtual' MAC addresses (e.g. hot/standby firewall configurations). Documenting both the physical device MAC as well as the 'Virtual' MAC address to a physical interface would be beneficial for tracing. Allowing a 'Virtual' mac address to exist on multiple devices allows for searching and tying clustered hot/standby devices together transparently.

### Database Changes

Would probably require a new one->many relationship of interface->mac address as well as fields on a mac address to permit flagging as physical or virtual.

jeremystretch:
I think it'd be reasonable to change `mac_address` from a single value to an array field capable of holding zero or more MAC addresses.

>  Enable the ability to tag a mac address as physical or virtual.

This doesn't seem necessary, since it's typically implied the OUI (the first 24 bits of the address). It would also necessitate the creation of a separate MAC address model, which, while not unreasonable, would be a much more substantial change.

lps-rocks:
I totally understand and appreciate the idea of keeping the change as
minimal and simple as possible.

The advantage to creating a separate MAC address model would allow for the
model to Eventually be referenced. There’s situations where a physical
interface may have multiple MACs assigned to it and an IP assigned to each
MAC. (Shared IPMI ports as an example).

It would also allow for MAC addresses to be more efficiently searched.
Right now I can’t search by MAC address :( makes me sad. That’s one of the
primary things network admins like to search by. (Another feature request
I’m going to put in)

The downside of using the first 24 bits for implied virtualization is it
would require maintaining code that knows if it’s virtual or not  (IIRC
there’s 4 specific hunks of MAC space for Locally Administered MAC
Addressing) and a number of vendors and FOSS projects violate that
specification.

On Jul 19, 2020, at 7:23 PM, Jeremy Stretch <notifications@github.com>
wrote:

﻿

I think it'd be reasonable to change mac_address from a single value to an
array field capable of holding zero or more MAC addresses.

Enable the ability to tag a mac address as physical or virtual.

This doesn't seem necessary, since it's typically implied the OUI (the
[...]

PR #17902, opened 2024-10-30
4867 multiple mac addresses

bctiemann:
### Closes: #4867 

This PR changes the `mac_address` of an Interface or VMInterface from a simple model field to a many-to-one relationship with a new `MACAddress` model. This allows many MAC addresses to be associated with a single Interface or VMInterface (with one optionally designated `is_primary`).

The "MAC Addresses" pageset is grouped under Devices -> Device Components.

Interfaces and VMInterfaces now show an inline table for creation of associated MACAddress objects:

Screenshot 2024-11-05 at 1 43 19 PM

There is a data migration which converts existing `mac_address` values to concrete `MACAddress` objects. There is also now a `mac_address` property on both `Interface` and `VMInterface` which reflects the value of the primary (or only) `MACAddress` associated with the interface, preserving backward compatibility for querying and display.

Filtering by `mac_address` works as before, acting on the related object's `mac_address` field.

Note: bulk create of Interfaces and VMInterfaces via POST will create associated MACAddress objects; but bulk edit of those same objects with `mac_address` specified in the payload is not supported. This is consistent with preexisting unit test coverage of bulk create and edit operations in `test_views.py` modules; however this is an area for potential future refinement.


netbox/dcim/views.py
```diff
@@ -2610,8 +2671,9 @@ def get_extra_context(self, request, instance):
 
         return {
             'vdc_table': vdc_table,
-            'bridge_interfaces_table': bridge_interfaces_tables,
-            'child_interfaces_table': child_interfaces_tables,
+            'bridge_interfaces_table': bridge_interfaces_table,
+            'child_interfaces_table': child_interfaces_table,
+            'mac_addresses_table': mac_addresses_table,
```
jeremystretch, line 2676: This isn't needed if we're using an embedded table in the interface view template; the table is populated using a separate HTTP request.

netbox/templates/dcim/interface.html
```diff
@@ -350,7 +350,23 @@ <h2 class="card-header">
...
-        
+      </div>
+    </div>
+  </div>
+  <div class="row mb-3">
+    <div class="col col-md-12">
+      <div class="card">
+        <h2 class="card-header">
+          {% trans "MAC Addresses" %}
+          {% if perms.ipam.add_macaddress %}
```
jeremystretch, line 361:
```suggestion
          {% if perms.dcim.add_macaddress %}
```

netbox/templates/dcim/macaddress.html
```diff
@@ -0,0 +1,50 @@
...
+{% load plugins %}
+{% load render_table from django_tables2 %}
+{% load i18n %}
+
+{% block content %}
+  <div class="row">
+    <div class="col col-md-4">
+      <div class="card">
+        <h2 class="card-header">{% trans "MAC Address" %}</h2>
+        <table class="table table-hover attr-table">
```
jeremystretch, line 12: We should include the raw MAC address value as a row here, with a copy-to-clipboard button for convenience.

netbox/templates/dcim/macaddress.html
```diff
@@ -0,0 +1,50 @@
...
+            <th scope="row">{% trans "Description" %}</th>
+            <td>{{ object.description|placeholder }}</td>
+          </tr>
+          <tr>
+            <th scope="row">{% trans "Assignment" %}</th>
+            <td>
+              {% if object.assigned_object %}
+                {% if object.assigned_object.parent_object %}
+                  {{ object.assigned_object.parent_object|linkify }} /
+                {% endif %}
```
jeremystretch, line 23: `parent_object` should always be a device or VM; I think it's ok to remove the conditional.

netbox/templates/dcim/macaddress_bulk_add.html
jeremystretch, line 1: This can be removed.

netbox/dcim/forms/model_forms.py
```diff
@@ -1435,6 +1524,21 @@ class Meta:
             'mode': '802.1Q Mode',
         }
 
+    def clean_mac_address(self):
```
jeremystretch, line 1527: We can't effect the creation of related objects under field validation, because it could result in an inconsistent state should later validations fail. This needs to be handled entirely under `save()`.

jeremystretch (changes requested):

netbox/dcim/models/devices.py
```diff
@@ -1473,3 +1476,43 @@ def clean(self):
...
+    assigned_object = GenericForeignKey(
+        ct_field='assigned_object_type',
+        fk_field='assigned_object_id'
+    )
+    is_primary = models.BooleanField(
+        verbose_name=_('is primary for interface'),
+        default=False
+    )
+
+    class Meta:
```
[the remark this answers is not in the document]
bctiemann: As discussed, this is addressed through validation in the model's `clean` method.

netbox/dcim/api/serializers_/device_components.py
```diff
@@ -214,8 +214,10 @@ class InterfaceSerializer(NetBoxModelSerializer, CabledObjectSerializer, Connect
         required=False,
         default=None,
         allow_blank=True,
-        allow_null=True
+        allow_null=True,
+        read_only=True
     )
+    mac_addresses = MACAddressSerializer(many=True, read_only=True)
```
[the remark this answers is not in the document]
bctiemann: Seems like a good idea. I made sure these are `nested=True` in both Interface serializers to keep some compactness.

netbox/dcim/tables/devices.py
```diff
@@ -597,6 +598,43 @@ def value_tagged_vlans(self, value):
...
+    assigned_device = tables.Column(
+        accessor='assigned_object__device',
+        verbose_name=_('Device'),
+        linkify=True
+    )
+    assigned_vm = tables.Column(
+        accessor='assigned_object__virtual_machine',
+        verbose_name=_('Virtual Machine'),
+        linkify=True
+    )
```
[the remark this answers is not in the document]
bctiemann: Agreed, I'd rather follow the IPAddressTable pattern than break it here for no good reason. I won't add an `assigned` column though because it doesn't seem very valuable for MACAddresses (they should always be assigned).

netbox/dcim/models/devices.py
```diff
@@ -1473,3 +1476,43 @@ def clean(self):
                 raise ValidationError({
                     f'primary_ip{family}': _('Primary IP address must belong to an interface on the assigned device.')
                 })
+
+
+#
+# Addressing
+#
+
+class MACAddress(PrimaryModel):
```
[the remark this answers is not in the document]
bctiemann: I reorganized a bunch of these files to put "Addressing" in its own area below the Device Components groupings so they don't interfere as much. Maybe could warrant its own module, I'll think about it.

netbox/virtualization/api/serializers_/virtualmachines.py
```diff
@@ -96,18 +97,19 @@ class VMInterfaceSerializer(NetBoxModelSerializer):
...
     count_fhrp_groups = serializers.IntegerField(read_only=True)
     mac_address = serializers.CharField(
-        required=False,
-        default=None,
-        allow_null=True
+        allow_null=True,
+        read_only=True
     )
+    mac_addresses = MACAddressSerializer(many=True, nested=True, read_only=True, allow_null=True)
+    ip_addresses = IPAddressSerializer(many=True, nested=True, read_only=True, allow_null=True)
```
jeremystretch, line 104: Same as with InterfaceSerializer above.
```suggestion
```

netbox/dcim/tests/test_filtersets.py
jeremystretch, line 1: Needs a test case for the MAC address filter set.

netbox/dcim/forms/model_forms.py
```diff
@@ -1435,6 +1437,21 @@ class Meta:
...
```
jeremystretch, line 1440: Let's just ditch the `mac_address` field on both interface forms. There are numerous problems with trying to accommodate this workflow. It might be something to tackle in the future, but for now we should just consistently treat MAC addresses as individual from their assigned interfaces, the way we do with IP addresses.

jeremystretch (changes requested):

jeremystretch (approved):

PR #17920, opened 2024-11-02
Closes: #17871 - Add cluster on bulkedit device

### LINKED ISSUE #17871: Add Cluster to Bulk Edit Device
### NetBox version

v3.5.6

### Feature type

Data model extension

### Triage priority

N/A

### Proposed functionality

When editing a several Devices there is currently no option to change Cluster in bulk.
I think it is a simple addition that might benefit for who ever needs it. 

### Use case

My team and I have a webhook that when a device is changed it will run and trigger some automatons,
This include Vlan change and more, We do it by editing the Cluster, But there is no way to bulk edit the Cluster so we either have to do it one by one or going From the Cluster side.

### Database changes

_No response_

### External dependencies

_No response_

pl0xym0r:
### Closes: #17871

Added Cluster selection on bulk editing devices.
I kept the query param $site from the form.

### DIFF
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -6,6 +6,7 @@ from timezone_field import TimeZoneFormField
 from dcim.choices import *
 from dcim.constants import *
 from dcim.models import *
+from virtualization.models import Cluster
 from extras.models import ConfigTemplate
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @pl0xym0r!

Issue #17929, opened 2024-11-04
Add Scope Mixins to Prefix


arthanson:
### Proposed Changes

Refactor the ipam.Prefix models and forms with the new Scope mixins

### Justification

Reduces boilerplate code and makes the code more consistent.  Scope was originally added to Prefix before the Mixins were created so now they need to be retrofitted.

PR #17930, opened 2024-11-04
17929 Add Scope Mixins to Prefix

arthanson:
### Fixes: #17929

Had to move the filterset mixin to it's own file as there was a circular dependency - this requires updating previous imports that reference it.
 
Note: The mixin uses %(class)s for the related_name which resolves to "_prefixs" instead of "_prefixes" which isn't great, but it is an internal field that shouldn't really be references, so shouldn't be much of an issue.


jeremystretch:
> Note: The mixin uses %(class)s for the related_name which resolves to "_prefixs" instead of "_prefixes" which isn't great, but it is an internal field that shouldn't really be references, so shouldn't be much of an issue.

IMO this is going to cause problems in the future, as we're forcing ourselves to accommodate what is essentially a typo. I propose adopting one of two options:

1. Use Django's default related name convention (`prefix_set`).
2. Drop the related name entirely for these fields.

netbox/dcim/filterset_mixins.py
jeremystretch, line 1: Could we rename this to `base_filtersets.py` or similar as ScopedFilterSet isn't actually a mixin.

jeremystretch (changes requested):

netbox/utilities/api.py
```diff
@@ -129,7 +129,12 @@ def get_annotations_for_serializer(serializer_class, fields_to_include=None):
 
     for field_name, field in serializer_class._declared_fields.items():
         if field_name in fields_to_include and type(field) is RelatedObjectCountField:
-            related_field = model._meta.get_field(field.relation).field
+            try:
+                related_field = model._meta.get_field(field.relation).field
+            except FieldDoesNotExist:
+                # Handle RelatedObjectCountField with no related_name given
+                related_field = getattr(model, field.relation).field
```
jeremystretch, line 136: Can we just use this approach for all cases?

jeremystretch (approved):

PR #17931, opened 2024-11-04
Fixes: #17459 - Ensure help text on component create forms shows both bulk edit and substitution token instructions

### LINKED ISSUE #17459: Component Name help_text is overwritten
### Deployment Type

Self-hosted

### NetBox Version

v4.1.0

### Python Version

3.10

### Steps to Reproduce

Found while looking into: #17436 
1. create a module type and add a component
2. Look at the help text for the name field

### Expected Behavior

The help text should include "{module} is accepted as a substitution for the module bay position when attached to a module type."

### Observed Behavior

The models help text is overwritten by the ExandableNameField "Alphanumeric ranges are supported for bulk creation. Mixed cases and types within a single range are ..."

bctiemann:
### Fixes: #17459

This PR does two things:

1. Ensures that the `The string <code>{module}</code> will be replaced with the position of the assigned module, if any.` string is shown on the `name` field of modular component template create forms, in addition to the `Alphanumeric ranges are supported for bulk creation` string, instead of the latter overwriting the former

2. Moves the `{module}` help_text logic from `InterfaceCreateForm` to the generic `ComponentCreateForm` so that it applies to all modular device components which share the same bulk creation logic

Screenshot 2024-11-04 at 3 54 53 PM


arthanson (dismissed):

netbox/dcim/forms/object_create.py
```diff
@@ -52,6 +52,19 @@ class ComponentCreateForm(forms.Form):
...
+
+        # Components attached to a module need to present this standardized substitution help text.
+        if 'module' in self.fields:
+            help_text_parts = []
+            if self.fields['name'].help_text:
+                help_text_parts.append(self.fields['name'].help_text)
+            help_text_parts.append(_(
+                "The string <code>{module}</code> will be replaced with the position of the assigned module, if any."
+            ))
+            self.fields['name'].help_text = ' '.join([str(x) for x in help_text_parts])
```
jeremystretch, line 66: This seems like a really roundabout approach, and is likely to complicate future changes to the form class. Is there any reason not to just set the desired help text on the `name` field directly?

netbox/dcim/forms/object_create.py
```diff
@@ -52,6 +52,17 @@ class ComponentCreateForm(forms.Form):
     # ComponentCreateView when creating objects.
     replication_fields = ('name', 'label')
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Components attached to a module need to present this standardized substitution help text.
+        if 'module' in self.fields:
```
jeremystretch, line 59: This doesn't address the reproduction steps in #17459, which entails a module type (as opposed to a module). Both fields should get the same help text.

jeremystretch (changes requested):

netbox/dcim/forms/object_create.py
bctiemann: Centralized the help text string in `common.py` (maybe there is a better place for this?) and reused both in `ModularComponentTemplateForm` (which is used in adding component templates to module types) and `ComponentCreateForm` (which is used when creating components directly).

jeremystretch: I misspoke earlier; only module _types_ support the use of the `{module}` token in component names. (This is replaced with the position of the module being created.) It is not supported for components being created on existing modules.

jeremystretch (changes requested):

netbox/dcim/forms/object_create.py
bctiemann: From what I can understand from the documentation, the text about `{module}` should ONLY appear on the component _template_ create forms (derived from `ModularComponentTemplateForm`), concatenated with the range expansion text. This applies to both workflows where component templates are created — when attaching to `ModuleType`s and to `DeviceType`s.

The `{module}` text should NOT appear on the direct object creation forms at all, including `InterfaceCreateForm` where it was appearing prior to this PR.

Is that correct?

jeremystretch (approved):

Issue #17923, opened 2024-11-04
Custom Script CLI and API execution not working


wz4:
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.6

### Python Version

3.11

### Steps to Reproduce

Unable to execute custom scripts via the CLI and API. This is the same issue referenced in #17885. The fix implemented only applies to the GUI. 

Adding the script name as an argument to the enqueue method fixes the problem.

https://github.com/netbox-community/netbox/blob/27d15615b3590000901984bddfa7a815cd1343d8/netbox/extras/api/views.py#L281-L290

https://github.com/netbox-community/netbox/blob/27d15615b3590000901984bddfa7a815cd1343d8/netbox/extras/management/commands/runscript.py#L84-L100

### Expected Behavior

Executing a script for the CLI and API should work

### Observed Behavior

The following error is produced when attempting to use the CLI runscript command
```
  File "/opt/netbox/netbox/extras/management/commands/runscript.py", line 85, in handle
    job = ScriptJob.enqueue(
          ^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/netbox/jobs.py", line 106, in enqueue
    return Job.enqueue(cls.handle, name=name, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/core/models/jobs.py", line 236, in enqueue
    job.full_clean()
  File "/opt/netbox/venv/lib/python3.12/site-packages/django/db/models/base.py", line 1561, in full_clean
    raise ValidationError(errors)
django.core.exceptions.ValidationError: {'name': ['This field cannot be blank.']}
```

cruse1977:
+1 to this

alehaa:
I repeat my comment from #17885 and volunteer for submitting a PR:

> The bug originates from the following code. The justification of this was to not have the script name twice (as job name and related object) listed in the job table. Just having "SciptJob" as generic name seemed a bit odd to me at that time. https://github.com/netbox-community/netbox/blob/58bc3884578a98595b02d7d7412c83d41fc7bbfa/netbox/extras/jobs.py#L24-L27
> […] Reverting #17886 and removing the code linked above should fix the issue and avoid confusion. `ScriptJob` will be used as default name afterwards and the script linked as job object.

arthanson:
[...]

PR #17932, opened 2024-11-04
Fixes: #17923, #17921 - Fix non-null constraint for script execution

### LINKED ISSUE #17921: Custom Script Scheduling Broken
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.6

### Python Version

3.10

### Steps to Reproduce

In Netbox v4.1.6 schedule a recurring custom script. The first job will get scheduled, but subsequent jobs are not.

Test Script

```

class TestScript(Script):
    class Meta:
        name = "Test Script"
        description = "Hello"
        scheduling_enabled = True

    def run(self, data, commit):

        return "HELLO"
```

### Expected Behavior

After a scheduled job completes, the next scheduled job should appear with a status of 'Scheduled'.

### Observed Behavior

No observed error

alehaa:
### Fixes: #17921

With c34a0e2, validation of job object fields is enabled, so ScriptJob must not set required fields to empty strings. Changes of this PR revert b18f193 and (hopefully) fix this issue not only for UI views, but for all interactions with scripts.

### DIFF
--- a/netbox/extras/jobs.py
+++ b/netbox/extras/jobs.py
@@ -22,9 +22,7 @@ class ScriptJob(JobRunner):
     """
 
     class Meta:
-        # An explicit job name is not set because it doesn't make sense in this context. Currently, there's no scenario
-        # where jobs other than this one are used. Therefore, it is hidden, resulting in a cleaner job table overview.
-        name = ''
+        name = 'Run Script'
 
     def run_script(self, script, request, data, commit):
         """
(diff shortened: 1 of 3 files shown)


alehaa:
I've added d0ee689 to cover the name being reused when the job is rescheduled. Thanks @miaow2!

For points 1 and 2 mentioned in the comment above, I'd like some feedback from maintainers. Since this is a cosmetic change, there may not be a good or bad design, just a decision based on feeling.

jeremystretch:
IMO naming the job "run script" is acceptable, because that's what it's doing, and each job retains a relation to its corresponding script.

jeremystretch (approved):
I'm accepting this as a fix for #17921 specifically. Will re-evaluate #17923 separately to see if any additional work is required for that bug. Thanks @alehaa!

PR #17933, opened 2024-11-04
Closes #13086: Virtual circuits

### LINKED ISSUE #13086: Introduce the ability to create "Virtual" circuits
### NetBox version

v3.5.4

### Feature type

### Proposed functionality

Allow a new Circuit Termination variant which would link to a "Parent" circuit and optionally allow the user to select a virtual interface of either Parent side connected interface.

### Use case

As the operator of a service provider network, we often lease connections from third parties.
These connections often have multiple components, some of which run "on top" of other components, think many different vlans on a parent link, with each vlan being terminated to a different location or "Z" side.

This was discussed in https://github.com/netbox-community/netbox/discussions/10558

Allowing netbox to properly document these is a must for service providers such as myself and would be of great benefit to the netbox community.

### Database changes

Amongst other things, the [CircuitTermination](https://github.com/netbox-community/netbox/blob/develop/netbox/circuits/models/circuits.py#L150C7-L150C25) model would need to have an additional ForeignKey to `circuits.CircuitTermination`. 

Additionally, a ForeignKey to `dcim.Interface` would allow to track the sub-interface related to the Parent circuit termination connected interface (and thus enable traces from sub-interfaces as well).

### External dependencies

None

jeremystretch:
### Fixes: #13086

- Introduce the VirtualCircuit and VirtualCircuitTermination models
- Permit virtual circuits to terminate directly to virtual interfaces on devices


arthanson:
When add an interface it shows all interfaces, even if they are already connected to a virtual Circuit, but if you select it you will get an error.  Can we filter the list to only show non assigned interfaces?
Add a new virtual circuit termination | NetBox 2024-11-15 11-25-43

netbox/templates/circuits/virtualcircuittermination.html
```diff
@@ -0,0 +1,59 @@
...
+          <tr>
+            <th scope="row">{% trans "Provider account" %}</th>
+            <td>{{ object.virtual_circuit.provider_account|linkify|placeholder }}</td>
+          </tr>
+          <tr>
+            <th scope="row">{% trans "Virtual circuit" %}</th>
+            <td>{{ object.virtual_circuit|linkify }}</td>
+          </tr>
+          <tr>
+            <th scope="row">{% trans "Role" %}</th>
```
arthanson, line 41: Interface needs to be shown

arthanson:
Should virtual circuit terminations only go to virtual interfaces? It seems they can be connected to any interface, however if you display the interface and it's not virtual then the peer (connection) info is not shown.

arthanson (changes requested):

jeremystretch:
> In the add termination form getting a lot of duplicate interface names, I think we need a disambiguator here to show the Device. 

I've added the device as context in the dropdown selections, which should help with differentiation. However, this is the same approach we take to interface selection for e.g. adding an IP address. While it could stand to be improved, that work will need to be captured and undertaken in a separate FR.

> Can we filter the list to only show non assigned interfaces?

Yes; we can employ the `virtual_circuit_termination_id` filter for this.

> Should virtual circuit terminations only go to virtual interfaces?

No, only connections to virtual interfaces should be permitted. I'll add some validation logic on VirtualCircuitTermination to enforce this.

netbox/circuits/views.py
```diff
@@ -537,3 +537,109 @@ class CircuitGroupAssignmentBulkDeleteView(generic.BulkDeleteView):
...
+
+@register_model_view(VirtualCircuit, 'delete')
+class VirtualCircuitDeleteView(generic.ObjectDeleteView):
+    queryset = VirtualCircuit.objects.all()
+
+
+class VirtualCircuitBulkImportView(generic.BulkImportView):
+    queryset = VirtualCircuit.objects.all()
+    model_form = forms.VirtualCircuitImportForm
+    # additional_permissions = [
```
bctiemann, line 574: Should this be removed?

netbox/circuits/models/virtual_circuits.py
```diff
@@ -0,0 +1,164 @@
...
+class VirtualCircuit(PrimaryModel):
+    """
+    A virtual connection between two or more endpoints, delivered across one or more physical circuits.
+    """
+    cid = models.CharField(
+        max_length=100,
+        verbose_name=_('circuit ID'),
+        help_text=_('Unique circuit ID')
+    )
+    provider_network = models.ForeignKey(
```
bctiemann, line 27: Just for my own understanding, virtual circuits don't appear to be very congruent to regular circuits in terms of the data model -- the latter have `provider_network` as part of the CircuitTermination, whereas here they're a top-level attribute of the VirtualCircuit rather than the termination. I assume that's an intentional part of the design.

netbox/circuits/models/virtual_circuits.py
```diff
@@ -0,0 +1,164 @@
...
+        objectchange = super().to_objectchange(action)
+        objectchange.related_object = self.virtual_circuit
+        return objectchange
+
+    @property
+    def parent_object(self):
+        return self.virtual_circuit
+
+    @cached_property
+    def peer_terminations(self):
```
bctiemann, line 146: Noting that this is a `@property` whereas in `CircuitTermination` it's a method, `get_peer_terminations`. I prefer the property style just for syntactic reasons but I'm wondering if it would be better to make the usage consistent in both places?

bctiemann (approved):

netbox/circuits/views.py
jeremystretch: Actually that was meant to be a reminder that additional work is needed to support nested terminations when bulk importing virtual circuits. Good catch.

netbox/circuits/models/virtual_circuits.py
jeremystretch: Right; this is because a physical circuit provides access from a customer POP _to_ a provider network, whereas a virtual circuit provides connectivity among multiple points _across_ a provider network.

bctiemann (approved):

Issue #7848, opened 2021-11-16
Webhooks: Query RQ queue for job status via API


netgab:
### NetBox version

v3.0.10

### Feature type

New functionality

### Proposed functionality

The webhook job status (Admin UI > Background Jobs) should be queried over the netbox API.

### Use case

The goal is to make webhooks more reliable in terms of failed jobs. For example if the webhook receiver is not available for any reason (e.g. downtime), the failed jobs are not retried.
To make it more reliable, it should be possible to query netbox for failed jobs and filter based on the date/time and/or webhook name. Basically all the information in `Kwargs` in the RQ queue should be returned.

A webhook receiver (or any other monitoring system) may query netbox in regular intervals for these failed webhook events to:
- automatically resolve the missed webhook
- generate an incident for manual countermeasures (let an admin look and verify)

### Database changes

not sure - sorry

netgab:
@jeremystretch: Sure - but I'm not a developer ... so it's just a rough / best effort guess.
For a generic approach (not focussing on webhooks in particular, instead focus on queues):
**API endpoint:** `api/extras/background-tasks/<QUEUE-NAME>/{failed|finished|...}/`
For my purpose (failed webhook stuff): `api/extras/background-tasks/default/failed/`

**Response:** (in YAML, because I'm lazy) 
```
count: <NUMBER-OF-ITEMS-IN-QUEUE>
results:
  - id: "<JOB-ID>"   # example: dbcd6415-af03-4366-8d8e-7f891c0ccb3b
    queue: "<QUEUE-NAME>"    # example: default
    callable: "<CALL-NAME>"   # example: extras.webhooks_worker.process_webhook
    kwargs:    # dictionary of kwargs 
 ```

To be more use case centric (failed webhooks):
**API endpoint:** `api/extras/webhook-status/`

**Response:** (in YAML, because I'm lazy) 
```
count: <NUMBER-OF-ITEMS-IN-QUEUE>
results:
  - id: "<JOB-ID>"   # example: dbcd6415-af03-4366-8d8e-7f891c0ccb3b
    webhook: "<WEBHOOK-NAME>"   # Link to the webhook object (/apt/extras/webhook)
    event: "{create|delete| ...}
    data: <DICTIONARY-OF-WEBHOOK-DATA>
    snapshots: <PRE-POST-CHANGE-DICTIONARY>
    timestamp: <TIMESTAMP>

Personally I'm for the use case centric approach.

PR #17938, opened 2024-11-05
7848 Add RQ API

arthanson:
### Fixes: #7848

Here is a testing command to add some tasks for testing the API response, create a temporary create_task management command within core. 

Used APIView on the API's and not ViewSet as they don't fit with a ViewSet / only one operation per API can't Get, List, Delete, etc.  and the URL structure doesn't directly conform with the pk used by ViewSet, however the side-effect of this is it doesn't appear in the browsable API.

```
from django_rq.management.commands.rqworker import Command as _Command

from datetime import datetime, timedelta
from django_rq import get_queue
from rq.job import JobStatus
from rq.registry import (
    DeferredJobRegistry,
    FailedJobRegistry,
    FinishedJobRegistry,
    ScheduledJobRegistry,
    StartedJobRegistry
    )


class Command(_Command):
    # Dummy worker functions
    @staticmethod
    def dummy_job_default():
        return "Job finished"

    @staticmethod
    def dummy_job_high():
        return "Job finished"

    @staticmethod
    def dummy_job_failing():
        raise Exception("Job failed")

        queue = get_queue('default')

        job = queue.enqueue(self.dummy_job_default)
        job.set_status(JobStatus.FINISHED)
        registry = FinishedJobRegistry(queue.name, queue.connection)
        registry.add(job, -1)

        job = queue.enqueue(self.dummy_job_default)
        job.set_status(JobStatus.FAILED)
        registry = FailedJobRegistry(queue.name, queue.connection)
        registry.add(job, -1)

        job = queue.enqueue(self.dummy_job_default)
        job.set_status(JobStatus.DEFERRED)
        registry = DeferredJobRegistry(queue.name, queue.connection)
        registry.add(job, -1)

        job = queue.enqueue(self.dummy_job_default)
        job.set_status(JobStatus.STARTED)
        registry = StartedJobRegistry(queue.name, queue.connection)
        registry.add(job, -1)

        job = queue.enqueue(self.dummy_job_default)
        job.set_status(JobStatus.SCHEDULED)
        when = datetime.now() + timedelta(hours=8)
        registry = ScheduledJobRegistry(queue.name, queue.connection)
        registry.schedule(job, when)

```


netbox/core/api/views.py
```diff
@@ -71,3 +82,251 @@ class ObjectChangeViewSet(ReadOnlyModelViewSet):
...
+    """
+    Deletes the specified RQ Task.
+    """
+    permission_classes = [IsAdminUser]
+
+    def get_view_name(self):
+        return "Background Task"
+
+    @extend_schema(responses={200: OpenApiTypes.OBJECT})
+    def get(self, request, task_id, format=None):
```
bctiemann, line 205: Should these be POST operations maybe?

netbox/core/api/views.py
```diff
@@ -71,3 +82,251 @@ class ObjectChangeViewSet(ReadOnlyModelViewSet):
     queryset = ObjectChange.objects.valid_models()
     serializer_class = serializers.ObjectChangeSerializer
     filterset_class = filtersets.ObjectChangeFilterSet
+
+
+class QueueListView(APIView):
```
jeremystretch, line 87: These list views should replicate the existing API convention, which enables pagination of the results:

```json

{
    "count": 24,
    "next": null,
    "previous": null,
    "results": [...]
}
```

netbox/core/views.py
```diff
@@ -347,40 +344,12 @@ class BackgroundTaskListView(TableMixin, BaseRQView):
...
-                try:
-                    jobs.append(RQ_Job.fetch(job_id, connection=queue.connection, serializer=queue.serializer))
-                except NoSuchJobError:
-                    pass
-
-        if jobs and status == RQJobStatus.SCHEDULED:
-            for job in jobs:
-                job.scheduled_at = registry.get_scheduled_time(job)
-
+        jobs = get_rq_jobs_from_status(queue, status)
```
jeremystretch, line 352:
```suggestion
        return get_rq_jobs_from_status(queue, status)
```

netbox/core/utils.py
```diff
@@ -0,0 +1,118 @@
...
+from rq.registry import (
+    DeferredJobRegistry,
+    FailedJobRegistry,
+    FinishedJobRegistry,
+    ScheduledJobRegistry,
+    StartedJobRegistry,
+)
+
+
+def get_rq_jobs_from_status(queue, status):
```
jeremystretch, line 18: Please add docstrings for these functions.

netbox/core/utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+    return jobs
+
+
+def delete_rq_job(job_id):
+    config = QUEUES_LIST[0]
+    try:
+        job = RQ_Job.fetch(job_id, connection=get_redis_connection(config['connection_config']),)
+    except NoSuchJobError:
+        raise Http404(_("Job {job_id} not found").format(job_id=job_id))
```
jeremystretch, line 56: We might want to move the 404 handling out to the UI & API views, to ensure that these functions can be used cleanly internally e.g. within tests.

jeremystretch (changes requested):

netbox/core/tests/test_api.py
```diff
@@ -91,3 +98,191 @@ def setUpTestData(cls):
...
+        self.user.is_staff = False
+        self.user.save()
+        response = self.client.get(url, **self.header)
+        self.assertEqual(response.status_code, 403)
+
+        # Load view with permission
+        self.user.is_staff = True
+        self.user.save()
+        response = self.client.get(url, **self.header)
+        self.assertEqual(response.status_code, 200)
```
jeremystretch, line 149: We should also verify the length of the `results` list in the response.

netbox/core/tests/test_api.py
```diff
@@ -91,3 +98,191 @@ def setUpTestData(cls):
             ),
         )
         DataFile.objects.bulk_create(data_files)
+
+
+class BackgroundTaskTestCase(TestCase):
```
jeremystretch, line 103: Missing `test_background_tasks_list_started()` and `test_background_tasks_list_queued()`?

jeremystretch (changes requested):

netbox/core/tests/test_api.py
arthanson: we are checking that the queues appear directly?  so I think the check of the length is redundant?

netbox/core/api/serializers_/tasks.py
```diff
@@ -0,0 +1,84 @@
...
+        return obj.get_status()
+
+
+class BackgroundQueueSerializer(serializers.Serializer):
+    name = serializers.CharField()
+    # Note: Queue names are not url safe so not returning a url field
+    jobs = serializers.IntegerField()
+    oldest_job_timestamp = serializers.CharField()
+    index = serializers.IntegerField()
+    connection_kwargs = serializers.DictField()
```
jeremystretch, line 57: Should we omit this? Doesn't seem particularly relevant to external consumers, and it includes potentially sensitive data.

netbox/core/api/serializers_/tasks.py
```diff
@@ -0,0 +1,84 @@
...
+    scheduler_pid = serializers.CharField()
+    workers = serializers.IntegerField()
+    finished_jobs = serializers.IntegerField()
+    started_jobs = serializers.IntegerField()
+    deferred_jobs = serializers.IntegerField()
+    failed_jobs = serializers.IntegerField()
+    scheduled_jobs = serializers.IntegerField()
```
jeremystretch, line 64: Might be a bit cleaner to include a single `jobs` dictionary with keys for `total` (the current `jobs` field above) and these:

```json
"jobs": {
    "total": 10,
    "scheduled": 3,
    "started": 1,
    "finished": 4,
    "deferred": 0,
    "failed": 2
}
```

netbox/core/api/serializers_/tasks.py
```diff
@@ -0,0 +1,84 @@
...
+    def get_position(self, obj) -> int:
+        return obj.get_position()
+
+    def get_status(self, obj) -> str:
+        return obj.get_status()
+
+
+class BackgroundQueueSerializer(serializers.Serializer):
+    name = serializers.CharField()
+    # Note: Queue names are not url safe so not returning a url field
```
jeremystretch, line 53: Is this because the queue names are user-defined and thus potentially contain non-alphanumeric characters? Seems like a fairly low risk that we can mitigate with URL encoding. Ultimately we need to provide a detail endpoint to retrieve a specific RQ queue.

netbox/core/api/serializers_/tasks.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class BackgroundWorkerSerializer(serializers.Serializer):
+    name = serializers.CharField()
+    url = serializers.HyperlinkedIdentityField(
+        view_name='core-api:rqworker-detail',
+        lookup_field='name'
+    )
+    state = serializers.SerializerMethodField()
+    birth_date = serializers.CharField()
```
jeremystretch, line 74: This attribute name strikes me as odd, both because RQ workers are not "born" and more so because it's not a date, but rather a complete timestamp. Maybe we should rename it to e.g. `start_time` to avoid confusion.

netbox/core/api/serializers_/tasks.py
```diff
@@ -0,0 +1,84 @@
...
+    queue_names = serializers.ListField(
+        child=serializers.CharField()
+    )
+    pid = serializers.CharField()
+    successful_job_count = serializers.IntegerField()
+    failed_job_count = serializers.IntegerField()
```
jeremystretch, line 80: I'd suggest nesting these as well:

```json
"jobs": {
    "total": 23,
    "successful": 18,
    "failed": 5
}
```

netbox/core/api/serializers_/tasks.py
```diff
@@ -0,0 +1,84 @@
...
+    last_heartbeat = serializers.CharField()
+
+    is_finished = serializers.BooleanField()
+    is_queued = serializers.BooleanField()
+    is_failed = serializers.BooleanField()
+    is_started = serializers.BooleanField()
+    is_deferred = serializers.BooleanField()
+    is_canceled = serializers.BooleanField()
+    is_scheduled = serializers.BooleanField()
+    is_stopped = serializers.BooleanField()
```
jeremystretch, line 42: All of these (except `is_queued`) always seem to return False, even when the corresponding UI view reports the diverse statuses as expected:

screenshot

netbox/core/tests/test_api.py
```diff
@@ -91,3 +98,148 @@ def setUpTestData(cls):
...
```
jeremystretch, line 103: We also need a test for the background task list view.

jeremystretch (changes requested):

netbox/core/api/serializers_/tasks.py
bctiemann: Also shouldn't it be a DateTimeField?

arthanson: That is the name it has in RQ, if we change it we would need to change it in the front-end and also document all these fields, right now they can be documented by looking at the RQ docs but if we start changing them we would need to document what they are as it would be confusing looking at the RQ docs.

netbox/core/api/serializers_/tasks.py
arthanson: There was a previous note that status was not correctly updated in the testing script, I've updated the testing script to update the status of the jobs correctly, please use new script above.

jeremystretch (approved):

PR #17992, opened 2024-11-12
Fixes #17986: Correct label for disk size when bulk editing virtual machines

### LINKED ISSUE #17986: Incorrect disk size unit when bulk editing VMs
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.6

### Python Version

3.12

### Steps to Reproduce

1. Create VM
2. Bulk edit VM

### Expected Behavior

"MB" unit shown for disk size in bulk edit form, as this is the unit for disk size in NetBox >= 4.1

### Observed Behavior

"GB" unit shown for disk size in bulk edit form.

jeremystretch:

### DIFF
--- a/netbox/virtualization/forms/bulk_edit.py
+++ b/netbox/virtualization/forms/bulk_edit.py
@@ -331,7 +331,7 @@ class VirtualDiskBulkEditForm(NetBoxModelBulkEditForm):
     )
     size = forms.IntegerField(
         required=False,
-        label=_('Size (GB)')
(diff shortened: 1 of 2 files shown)


bctiemann (approved):

lassebm (commented):
I guess it needs fixing in line 334 as well.

arthanson (changes requested):
yeah as mentioned by @lassebm it needs the same change on line 334

PR #17995, opened 2024-11-12
Fixes #17954 - Handle CircuitTerminations in Cable Bulk Import

### LINKED ISSUE #17954: Cable bulk import when importing circuit terminations
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v.4.1.6

### Python Version

3.12

### Steps to Reproduce

When trying to use the bulk cable import feature to import cables with circuit terminations, I'm getting the error "Record 1 side_a_device: Object not found:397718".

To reproduce:
I have an existing device NTT-ODF1 with a port X01.
I have an existing circuit, 397718 and I'm trying to connect to Termination A

Import data:
397718,circuits.circuittermination,A,NTT-ODF1,dcim.rearport,X01

### Expected Behavior

Should be able to import cables connected to circuit terminations.

### Observed Behavior

Screenshot 2024-11-07 at 16 50 17
Screenshot 2024-11-07 at 16 46 55
Screenshot 2024-11-07 at 16 46 22

jsenecal:
### Add Support for Bulk Importing Cables Connected to Circuits

This pull request enhances the bulk import functionality for cables by allowing users to specify circuits as termination points on either side of a cable. Previously, only devices could be specified for cable terminations during bulk import which didn't allow for Cables to be connected to Circuit Terminations properly.

**Key Changes:**

- **Added Fields for Circuit Terminations:**
  - Introduced `side_a_circuit` and `side_b_circuit` fields in the `CableImportForm` within `dcim/forms/bulk_import.py`.
  - These fields are optional and conditional, allowing users to specify a Circuit ID (`cid`) as a termination point.
  
- **Updated CableImportForm Validation Logic:**
  - Modified the `_clean_side` method in `CableImportForm` to handle cases where a circuit is specified instead of a device.
  - Added validation to ensure that both a device and a circuit cannot be specified simultaneously for the same side.
  - Ensured that the termination object is correctly retrieved based on whether a device or circuit is provided.
  - Added checks to prevent connecting to a circuit termination that is already connected to a Provider Network.

- **Enhanced Bulk Import Template:**
  - Updated `bulk_import.html` to display a "Conditional" tag for fields that are conditionally required, improving user experience during the import process.

- **Extended CSV Field Functionality:**
  - Modified `CSVModelChoiceField` in `utilities/forms/fields/csv.py` to accept a `conditional` parameter.
  - This parameter is used to indicate fields that are conditionally required based on other form fields.

**Benefits:**

- **Improved Flexibility:** Users can now bulk import cables that connect directly to circuits, not just devices.
- **Enhanced User Experience:** Clearer form validation messages and template indicators help prevent import errors.
- **Increased Efficiency:** Streamlines the process of importing large numbers of cables connected to circuits.

**Usage Notes:**

- When importing, specify either a device or a circuit for each side of the cable—not both.
- Ensure that the circuit terminations are not already connected to a Provider Network to avoid validation errors.

### DIFF
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -8,6 +8,7 @@ from django.utils.translation import gettext_lazy as _
 from dcim.choices import *
 from dcim.constants import *
 from dcim.models import *
+from circuits.models import Circuit
 from extras.models import ConfigTemplate
 from ipam.models import VRF, IPAddress
 from netbox.forms import NetBoxModelImportForm
@@ -1171,7 +1172,17 @@ class CableImportForm(NetBoxModelImportForm):
         label=_('Side A device'),
         queryset=Device.objects.all(),
         to_field_name='name',
+        required=False,
+        conditional=True,
         help_text=_('Device name')
+    )
+    side_a_circuit = CSVModelChoiceField(
+        label=_('Side A circuit'),
+        queryset=Circuit.objects.all(),
+        to_field_name='cid',
+        required=False,
+        conditional=True,
+        help_text=_('Circuit ID'),
     )
     side_a_type = CSVContentTypeField(
         label=_('Side A type'),
@@ -1189,7 +1200,17 @@ class CableImportForm(NetBoxModelImportForm):
         label=_('Side B device'),
         queryset=Device.objects.all(),
         to_field_name='name',
+        required=False,
+        conditional=True,
         help_text=_('Device name')
+    )
+    side_b_circuit = CSVModelChoiceField(
+        label=_('Side A device'),
+        queryset=Circuit.objects.all(),
+        to_field_name='cid',
+        required=False,
+        conditional=True,
+        help_text=_('Circuit ID'),
     )
     side_b_type = CSVContentTypeField(
         label=_('Side B type'),
@@ -1232,7 +1253,7 @@ class CableImportForm(NetBoxModelImportForm):
     class Meta:
         model = Cable
         fields = [
-            'side_a_device', 'side_a_type', 'side_a_name', 'side_b_device', 'side_b_type', 'side_b_name', 'type',
+            'side_a_device', 'side_a_circuit', 'side_a_type', 'side_a_name', 'side_b_device', 'side_b_circuit', 'side_b_type', 'side_b_name', 'type',
             'status', 'tenant', 'label', 'color', 'length', 'length_unit', 'description', 'comments', 'tags',
         ]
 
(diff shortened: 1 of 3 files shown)


netbox/dcim/forms/bulk_import.py
```diff
@@ -1245,18 +1266,40 @@ def _clean_side(self, side):
...
         model = content_type.model_class()
         try:
-            if device.virtual_chassis and device.virtual_chassis.master == device and \
-                    model.objects.filter(device=device, name=name).count() == 0:
-                termination_object = model.objects.get(device__in=device.virtual_chassis.members.all(), name=name)
-            else:
-                termination_object = model.objects.get(device=device, name=name)
+            
+            # Should never happen as we return None above if we don't have a device or circuit
+            assert device or circuit
```
DanSheps, line 1285: One of the things we came up to in another PR, is asserts are not a guaranteed thing since it requires python to be compiled with debug=true.

I am actually going to defer to Jeremy on whether we want to stick with asserts or not.

netbox/dcim/forms/bulk_import.py
```diff
@@ -1189,8 +1200,18 @@ class CableImportForm(NetBoxModelImportForm):
...
+        required=False,
+        conditional=True,
         help_text=_('Device name')
     )
+    side_b_circuit = CSVModelChoiceField(
+        label=_('Side A device'),
+        queryset=Circuit.objects.all(),
+        to_field_name='cid',
+        required=False,
+        conditional=True,
```
DanSheps, line 1212: See comment re: conditional below

netbox/dcim/forms/bulk_import.py
```diff
@@ -1189,8 +1200,18 @@ class CableImportForm(NetBoxModelImportForm):
         label=_('Side B device'),
         queryset=Device.objects.all(),
         to_field_name='name',
+        required=False,
+        conditional=True,
```
DanSheps, line 1204: See comment re: conditional below

netbox/dcim/forms/bulk_import.py
```diff
@@ -1171,8 +1172,18 @@ class CableImportForm(NetBoxModelImportForm):
...
         help_text=_('Device name')
     )
+    side_a_circuit = CSVModelChoiceField(
+        label=_('Side A circuit'),
+        queryset=Circuit.objects.all(),
+        to_field_name='cid',
+        required=False,
+        conditional=True,
```
DanSheps, line 1184: See comment re: conditional below

netbox/dcim/forms/bulk_import.py
```diff
@@ -1171,8 +1172,18 @@ class CableImportForm(NetBoxModelImportForm):
         label=_('Side A device'),
         queryset=Device.objects.all(),
         to_field_name='name',
+        required=False,
+        conditional=True,
```
DanSheps, line 1176: See comment re: conditional below

netbox/templates/generic/bulk_import.html
```diff
@@ -129,6 +129,8 @@ <h2 class="card-header">{% trans "Field Options" %}</h2>
                 <td>
                   {% if field.required %}
                     {% checkmark True true="Required" %}
+                  {% elif field.conditional %}
+                    <span class="badge bg-yellow text-yellow-fg" >{% trans "Conditional" %}</span>
```
DanSheps, line 133: See comment re: conditional below

netbox/utilities/forms/fields/csv.py
```diff
@@ -57,6 +57,11 @@ class CSVModelChoiceField(forms.ModelChoiceField):
         'invalid_choice': _('Object not found: %(value)s'),
     }
 
+    def __init__(self, conditional=False, *args, **kwargs):
+        # Used to display tags for fields that are conditionally required
+        self.conditional = conditional
+        super().__init__(*args, **kwargs)
+
```
DanSheps, line 64: I am not myself prepared to make a decision on whether this is something we should consider introducing.

That said, I think that if this is something we want to persue further, there should be a FR for it and a related PR as I am sure there is more then just this.

For example, it marks it as a conditionally required value but do we want to enrich that further by telling the end-user which fields are part of this conditional group?

jeremystretch:
As mentioned in the comments on #17954, this limitation also impacts power feeds. IMO we should solve for both circuit terminations and power feeds here.

Instead of introducing additional fields, I'll suggest rename `side_x_device` to `side_x_parent`, and repurposing it to indicate the parent object (circuit, device, or power panel) depending on the selected type of termination for each side.

jsenecal:
> As mentioned in the comments on #17954, this limitation also impacts power feeds. IMO we should solve for both circuit terminations and power feeds here.

I did not tackle the other one as I wanted a consensus on the methodology first. I don't mind fixing both the same way.
> 
> Instead of introducing additional fields, I'll suggest rename `side_x_device` to `side_x_parent`, and repurposing it to indicate the parent object (circuit, device, or power panel) depending on the selected type of termination for each side.

Do you suggest that we design a replacement for `CSVModelChoiceField`? I don't know of a current CSV*Field that allows for a dynamic lookup like the one you suggest here @jeremystretch.

jsenecal:
Well, I'm waiting on @jeremystretch input at this point, not sure if it really should be marked as stale ?

jeremystretch:
I have no further opinions to offer. See what you can make work.

jsenecal:
> Do you suggest that we design a replacement for `CSVModelChoiceField`? I don't know of a current CSV*Field that allows for a dynamic lookup like the one you suggest here @jeremystretch; Also, the accessor is different for each ContentType, which makes me wonder how to properly convey its usage.

This is what I am referring to @jeremystretch, could you please comment on this ?

arthanson:
@jsenecal I think the issue here is it's hard to know if Jeremy's proposed change to fix this particular issue is workable with CSVModelChoiceField without digging into the issue and trying/fixing it.  The solution (if possible) should address both Cables and Power Feeds, which this doesn't.  

Per your question: "Do you suggest that we design a replacement for CSVModelChoiceField?".. Maybe?  But without digging in and trying it is hard to say.  IMHO I'd suggest we close this PR as it doesn't look like the best solution and see if the  side_x_device to side_x_parent is a workable solution or not and submit a new PR.

Sergii-Lucky:
Hello colleagues!
Any news about CircuitTerminations in Cable Bulk Import?

jsenecal:
Closing this PR due to lack of necessary design guidance. 

The proposed direction requires clarification on how to implement a dynamic lookup field that can reference different parent objects (circuit, device, power panel) with different accessors based on termination type. As noted in my [question](https://github.com/netbox-community/netbox/pull/17995#issuecomment-2530100020), this would likely require designing a replacement for `CSVModelChoiceField`.

Without technical direction on this architectural decision, I'm unable to properly implement the suggested approach. If there's interest in revisiting this bug with clearer specifications, please feel free to let me know or create a new issue with the necessary implementation details and I'll be happy to contribute.

PR #18011, opened 2024-11-14
17686 config option for disk divider

### LINKED ISSUE #17686: Abillity to select disk divider
### NetBox version

v4.1.3

### Feature type

### Proposed functionality

Disk size changed from GB to MB, but divider is set to 1000 instead of 1024. Import to Netbox that uses Bytes will show wrong disk size
see bug report
https://github.com/netbox-community/netbox/issues/17631 

### Use case

Field used to be GB, is now MB
Field collected from VMM is in Bytes, 1024 Bytes is 1KB, 1024KB is 1MB, 1024MB is 1GB, 1024GB is 1TB
Netbox uses 1000 Bytes as 1KB, 1000MB as 1GB
Example
Importing a value of 50000 MB on virtual disk is shown as 50GB
While converting from bytes to MB
52428800 Bytes is 51200MB

In Netbox, 51200 MB is converted to 51,2GB which is not correct

if I convert the bytes
52428800/1MB = 51200 MB which is correct, but shown wrong
(52428800/1GB)*1000 = 50000MB which is wrong, but shown correctly

Problems occur if the disk is larger than 1TB

(1 073 741 824/1GB)*1000 = 1024GB which is wrong, will be shown as 1.02 TB

Correct would be to not use humanizenumbers as conversion will be wrong
A posible solution would be to add ability in configuration to choose between 1000 and 1024

image
vs edit
image

To get value 60GB, the MB value should be 61440MB

But if I change value the converted value shown under virual disk is
image

We use netbox as source for our billing, and as a workaround they now divide the MB number by 1000
The correct way would be for billing to devide by 1024

### Database changes

Not sure, but is it possible to add a setting in configuration.py to enable selection between 1000 and 1024

### External dependencies

None that I can think of

Haeki:
### Fixes: #17686 Ability to select disk divider

I added two new config options to specify the divider used to calculate the disk and ram sizes in different units.
The default value is 1000 but can be changed to 1024 to get GiB instead of GB.
Those config values are used in the `humanize_megabytes` method. I split this method into two separate methods, one for RAM and one for disk values. Additionally I added the disk config value to the disk migration so that users with GiB values get them converted to MiB.
I tested the changes in my local environment and noticed no problems.

I'm open for feedback and I also didn't include any tests yet.


1358:
Would love to see this merged, too.

DanSheps:
Can you rebase to main?

DanSheps (changes requested):

netbox/netbox/settings.py
```diff
@@ -176,6 +176,8 @@
 STORAGE_CONFIG = getattr(configuration, 'STORAGE_CONFIG', {})
 TIME_ZONE = getattr(configuration, 'TIME_ZONE', 'UTC')
 TRANSLATION_ENABLED = getattr(configuration, 'TRANSLATION_ENABLED', True)
+DISK_UNIT_DIVISOR = getattr(configuration, 'DISK_UNIT_DIVISOR', 1000)
+RAM_UNIT_DIVISOR = getattr(configuration, 'MEMORY_UNIT_DIVISOR', 1000)
```
DanSheps, line 180: We should have a check in here to verify it is either 1000 or 1024

DanSheps (dismissed):

netbox/netbox/configuration_example.py
```diff
@@ -221,6 +221,11 @@
 # database access.) Note that the user as which NetBox runs must have read and write permissions to this path.
 SESSION_FILE_PATH = None
 
+# By default the memory and disk sizes are displayed using base 10 (e.g. 1000 MB = 1 GB).
+# If you would like to use base 2 (e.g. 1024 MB = 1 GB) set this to 1024.
+DISK_UNIT_DIVISOR  = 1024
+RAM_UNIT_DIVISOR = 1024
```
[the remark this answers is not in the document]
Haeki: 👍🏼

netbox/netbox/settings.py
Haeki: I added a simple check e21e31762799684938b855dc0c4e917c03beec13

netbox/netbox/configuration_example.py
```diff
@@ -221,6 +221,11 @@
 # database access.) Note that the user as which NetBox runs must have read and write permissions to this path.
 SESSION_FILE_PATH = None
 
+# By default the memory and disk sizes are displayed using base 10 (e.g. 1000 MB = 1 GB).
+# If you would like to use base 2 (e.g. 1024 MB = 1 GB) set this to 1024.
+DISK_BASE_UNIT = 1024
```
arthanson, line 226: would want these commented out as if you copied this over the default would change.

arthanson (changes requested):
These also need to be added to the documentation, configuration/miscellaneous.md

arthanson:
@Haeki is this ready for re-review? If so please toggle the re-request review (spinner) in the upper right next to my name.

netbox/netbox/settings.py
```diff
@@ -176,6 +176,12 @@
 STORAGE_CONFIG = getattr(configuration, 'STORAGE_CONFIG', {})
 TIME_ZONE = getattr(configuration, 'TIME_ZONE', 'UTC')
 TRANSLATION_ENABLED = getattr(configuration, 'TRANSLATION_ENABLED', True)
+DISK_BASE_UNIT = getattr(configuration, 'DISK_BASE_UNIT', 1000)
+if DISK_BASE_UNIT not in [1000, 1024]:
+    raise ImproperlyConfigured(f"DISK_BASE_UNIT must be 1000 or 1024 (found {DISK_BASE_UNIT})")
+RAM_BASE_UNIT = getattr(configuration, 'MEMORY_UNIT_DIVISOR', 1000)
```
arthanson, line 182: I think this needs to be changed to RAM_BASE_UNIT - I'm not seeing RAM get changed in the display.
vm4

arthanson (changes requested):

netbox/netbox/settings.py
Haeki: Seems like my find and replace missed that one. Thanks

arthanson (approved):

Issue #7598, opened 2021-10-21
Support for filtering on custom fields 


emil-nasso:
### NetBox version

v3.0.3

### Feature type

New functionality

### Proposed functionality

There doesn't seem to be a way to filter based on custom fields in the GraphQL-api. 

The documentation mentions "The GraphQL API employs the same filtering logic as the UI and REST API" so one approach would be to just follow the same pattern where `cf_foo_bar` etc are exposed as arguments for the list query of an entity where `foo_bar` is a custom field.

(Reference to discussion about this: https://github.com/netbox-community/netbox/discussions/7569 )

### Use case

We have the use for this when listing tenants via the graphql API. We put the customers name in the name field but we have a custom field where we have an internal identifier that is used as a reference when linking the data in netbox up with other systems though integrations.

candlerb:
Another way might be via a graphql query on `custom_fields`.  I note that you can ask for `custom_fields` in a response:

```
# curl -sS -X GET -H "Content-Type: application/json" -H "Accept: application/json" http://localhost:8001/graphql/ \
  --data '{"query": "query {device_list(id:\"5\") {id custom_fields}}"}' | python3 -m json.tool
{
    "data": {
        "device_list": [
            {
                "id": "5",
                "custom_fields": {
                    "snmp_module": [
                        "mikrotik_secret"
                    ]
                }
            }
        ]
    }
}
```

(There's also `custom_field_data` which returns a string holding JSON).  But as far as I can tell, you can't currently filter on `custom_fields` in a query:

```
# curl -sS -X GET -H "Content-Type: application/json" -H "Accept: application/json" http://localhost:8001/graphql/ \
  --data '{"query": "query {device_list(custom_fields:{snmp_module:\"mikrotik_secret\"}) {id custom_fields}}"}' | python3 -m json.tool
{
    "errors": [
        {
            "message": "Unknown argument \"custom_fields\" on field \"device_list\" of type \"Query\".",
            "locations": [
                {
                    "line": 1,
                    "column": 20
                }
            ]
        }
    ]
}
```

It would of course be good if such a query could be pushed down to a SQL condition in postgres, rather than making Netbox retrieve all objects from the database and discard the ones which don't match.

jeremystretch:
Blocked by #9856

arthanson:
Adding blocked until #9856

sirtux:
@arthanson Since #9856 is done, can you unblock this issue? Any ideas if and when this is now going to happen?

[...]

PR #18012, opened 2024-11-14
wip: Filter by custom fields in graph ql

christophbeberweil:
### Works on : #7598 

I worked on issue #7598 and made progress. Unfortunaltely, I hit some roadblocks and none of the maintainers were available to answer my questions. I open this merge request with my partial solution so that it might be of use for the project in the future.

I utilized `autotype_decorator`, which is used to communicate the filterset to graphql, in order to add `CustomField` entries with the same logic applied as in `netbox.filtersets.NetBoxModelFilterSet`. The code calls `create_attribute_and_function` for each custom field of the current model. This successfully adds a new filter to GraphQL, which is recognized in the backend. Unfortunately, there are still some problems:
- The generated Filter itereates over the filter argument, so a filter by "Foo" filters by "F" OR "o" OR "o", which is not the desired functionality.
- Additionally, filters on non-iterable datatypes such as `int` do not work.
- After adding a new CustomField, the server needs to be restarted in order for GraphQL to be aware of the new field. Since GraphQL is a schema-based tool, it is not clear if the restart requirement can be avoided.
- The lints in the GraphIQL-Frontend do not recognize the new CustomField filter. Nevertheless, applying the filter will lead to a change in the output, so it is recognized by the server. For instance, after creating a custom field `a_string` on `Device`, the following graphql statement can be used to filter the devices, but GraphIQL does not know of `cf_a_string`:
```graphql
{
 device_list(filters: {cf_a_string: "a"}) {
  id
  name
  custom_fields
  local_context_data
  device_type {
   id
   display
   slug
  }
 }
}
```

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -27,7 +27,7 @@ from wireless.models import WirelessLAN, WirelessLink
 from .choices import *
 from .constants import *
 from .models import *
-
+from django.db.models import JSONField
 __all__ = (
     'CableFilterSet',
     'CabledObjectFilterSet',
@@ -1186,6 +1186,7 @@ class DeviceFilterSet(
             'module_bay_count',
             'inventory_item_count',
         )
+
 
     def search(self, queryset, name, value):
         if not value.strip():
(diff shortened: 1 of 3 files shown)


jeremystretch:
@christophbeberweil are you still planning to work on this?

christophbeberweil:
@jeremystretch Unfortunately not, since I was not able to obtain answers for my questions from you or the community.

PR #18016, opened 2024-11-14
Closes #5858: Implement a quick-add UI widget for related objects

### LINKED ISSUE #5858: quick-add for mandatory fields
### Environment
* Python version:  2.7.16
* NetBox version: v2.10.3

### Proposed Functionality
When adding a new device for example, you always have the finish creating anything related (mandatory fields).
A "+" button would be useful, which would direct you the the specific editor (device-role in this case), you can then create the device-role - save and netbox redirects you to the original (device) editor and auto-fill the field.

image

### Use Case
Faster creating of anything that has relations (mandatory field) (e.g. devices -> device-roles)

### Database Changes
Don't know

### External Dependencies
Don't know

jeremystretch:
### Closes: #5858

- Extend the APISelect widget to support a "quick add" button, which launches an embedded modal
- Extend ObjectEditView to support a new "quick add" workflow (designated by the `_quickadd=True` URL query parameter)
- Adapt `initReslug()` to accommodate multiple "reslug" buttons on the page
- Add `quick_add=True` to various form fields as appropriate

Peek 2024-11-13 14-57


arthanson:
slug fields don't auto-populate.

netbox/utilities/forms/fields/dynamic.py
```diff
@@ -66,6 +66,7 @@ class DynamicModelChoiceMixin:
             choice (DEPRECATED: pass `context={'disabled': '$fieldname'}` instead)
         context: A mapping of <option> template variables to their API data keys (optional; see below)
         selector: Include an advanced object selection widget to assist the user in identifying the desired object
+        quick_add: Include a widget to quickly create a new related object for assignment
```
arthanson, line 69: Maybe add a comment do not set it if the folllow-on form has any quick add fields.

jeremystretch:
> * slug fields don't auto-populate if there is a slug field also on the launching form.
> * regenerate slug (spinner to the right of slug field) regenerates based on original form, not modal form
> * It would also be nice to have the quick-added field auto-selected when you create it in the form that you launched it from

This all seems to work as expected. (You might need to force a refresh of the JS.)

Peek 2024-11-18 09-25

arthanson (approved):

PR #18026, opened 2024-11-15
Hide `sensitive_parameters` from datasource view even for high privil…

### LINKED ISSUE #18007: Remove or hide password in Data Source UI
### NetBox version

v4.1.6

### Feature type

### Triage priority

### Proposed functionality

Remove or hide the plain text password that is currently displayed in the `Backend` section of the `Data Source` view.

### Use case

I'm using fine-grained personal access tokens to connect to private GitHub repositories as Data Sources. While these access tokens have very minimal permissions, I don't think they should be displayed in plain text any time I navigate to the Data Source in the UI.

At a minimum, the password / PAT should be hidden (requiring a click to expose the plain text), but I can't see any good reason to show them at all and would just remove it from the view entirely.

### Database changes

None

### External dependencies

None

mskalecki:
…ege users

### Fixes: #18007

This PR hides `sensitive_parameters` in the Data Source view's Backend Panel, even for users with sufficient permissions to set that field.

My preference would be to also hide sensitive fields in the edit form, but I don't think that is done anywhere else in NetBox, and I'm not in a position to make that design choice.

### DIFF
--- a/netbox/templates/core/datasource.html
+++ b/netbox/templates/core/datasource.html
@@ -87,7 +87,7 @@
               {% for name, field in backend.parameters.items %}
                 <tr>
                   <th scope="row">{{ field.label }}</th>
-                  {% if name in backend.sensitive_parameters and not perms.core.change_datasource %}
+                  {% if name in backend.sensitive_parameters %}
                     <td>********</td>
                   {% else %}
                     <td>{{ object.parameters|get_key:name|placeholder }}</td>


mskalecki:
This is just a partial solution to the problem. Ideally, the `sensitive_parameters` should never be sent to the client at all.

jeremystretch (approved):

PR #18029, opened 2024-11-15
Closes #18023: Employ `register_model_view()` for list views

### LINKED ISSUE #18023: Extend `register_model_view()` to handle list views
### Proposed Changes

We currently use `register_model_view()` to register detail views (edit, delete, etc.) for models, which obviates the need to declare explicit URL paths for each. This could be extended to support list views as well. For example, a common set of views might look like this:

```python
# List view
@register_model_view(MyModel, list=True)
class FooListView(generic.ObjectListView):

# Detail view
@register_model_view(MyModel)
class MyModelView(generic.ObjectView):

# Add/edit views
@register_model_view(MyModel, 'edit')
@register_model_view(MyModel, 'add', list=True)
class MyModelEditView(generic.ObjectEditView):

# Delete view
@register_model_view(MyModel, 'delete')
class MyModelDeleteView(generic.ObjectDeleteView):

# Bulk import view
@register_model_view(MyModel, 'import', list=True)
class MyModelBulkImportView(generic.BulkImportView):

# Bulk delete view
@register_model_view(MyModel, 'delete', list=True)
class MyModelBulkDeleteView(generic.BulkDeleteView):
```

The corresponding set of URL paths could then be reduced to:

```python
path('my-models/', include(get_model_urls('myapp', 'mymodel', list=True))),
path('my-models/<int:pk>/', include(get_model_urls('myapp', 'mymodel'))),
```

(This could potentially be reduced even further, but I'm also wary of over-optimizing.)

`get_model_urls()` will likewise need to be extended to differentiate between list and detail views.

### Justification

This will allow us to reduce a substantial amount of boilerplate URL patterns without any breaking changes (AFAICT).

jeremystretch:
### Closes: #18023

- Extend `register_model_view()` and `get_model_urls()` to support list views
- Register list views using `register_model_view()` as we do currently for detail views
- Remove static URL paths for list views


netbox/vpn/urls.py
```diff
@@ -1,89 +1,39 @@
 from django.urls import include, path
 
 from utilities.urls import get_model_urls
-from . import views
+from . import views  # noqa F401
```
bctiemann, line 4: I do feel like these ought to be avoided, if just because they make things less clear.

bctiemann (approved):

netbox/vpn/urls.py
jeremystretch: Agreed, but we need to import the views somehow. If not here, we could do it under the `ready()` method for each app.

PR #18033, opened 2024-11-17
Closes #16076: Extend `PROTECTION_RULES` example

### LINKED ISSUE #16076: Example for PROTECTION_RULES in WEB UI.
### Change Type

Addition

### Area

Configuration

### Proposed Changes

I have not managed to configure the protection against deletion of a device using the documentation. I would therefore like to describe an example of how to configure this for Size, Racks and Device and also know how to use it for other areas.

The experiences are based on my discussion: 
https://github.com/netbox-community/netbox/discussions/16021

I have a suggestion here:
https://github.com/netbox-community/netbox/commit/c40f0ca5271d7a4f3e24341a952d150708ed806e

LHBL2003:
Issues to change the documentation
https://github.com/netbox-community/netbox/issues/16076

Describes how to configure the PROTECTION_RULES as a user via the WEB UI.

Fixes: https://github.com/netbox-community/netbox/issues/16076

The PR was already set some time ago, but it stalled several times. This time without pictures because I had embedded them incorrectly several times.


docs/configuration/data-validation.md
```diff
@@ -94,17 +94,17 @@ The following colors are supported:
 
 !!! tip "Dynamic Configuration Parameter"
 
-This is a mapping of models to [custom validators](../customization/custom-validation.md) against which an object is evaluated immediately prior to its deletion. If validation fails, the object is not deleted. An example is provided below:
+This is a mapping of models to [custom validators](../customization/custom-validation.md) against which an object is evaluated immediately prior to its deletion. If validation fails, the object is not deleted. The configuration can be carried out in the menu under “System/Configuration History/Custom validators”. 
+In the example, a device can only be deleted when it is set to the “decommissioning” status. Paths such as “dcim.device” can be found in the top right-hand corner when you open the page of a device. For a rack it would therefore be “dcim.rack”.
 
 ```python
-PROTECTION_RULES = {
```
jeremystretch, line 100: There's no reason to remove this.

docs/configuration/data-validation.md
```diff
@@ -94,17 +94,17 @@ The following colors are supported:
...
-    "dcim.site": [
```
jeremystretch, line 101: Why change the object type?

docs/configuration/data-validation.md
```diff
@@ -94,17 +94,17 @@ The following colors are supported:
...
+{
+    "dcim.device": [
         {
             "status": {
                 "eq": "decommissioning"
             }
-        },
-        "my_plugin.validators.Validator1",
```
jeremystretch, line 107: There's no reason to remove this.

docs/configuration/data-validation.md
```diff
@@ -94,17 +94,17 @@ The following colors are supported:
 
 !!! tip "Dynamic Configuration Parameter"
 
-This is a mapping of models to [custom validators](../customization/custom-validation.md) against which an object is evaluated immediately prior to its deletion. If validation fails, the object is not deleted. An example is provided below:
+This is a mapping of models to [custom validators](../customization/custom-validation.md) against which an object is evaluated immediately prior to its deletion. If validation fails, the object is not deleted. The configuration can be carried out in the menu under “System/Configuration History/Custom validators”.
```
jeremystretch, line 97: > The configuration can be carried out in the menu under “System/Configuration History/Custom validators”.

This is implied by the "dynamic configuration parameter" note above, which appears on many similar parameters.

jeremystretch (changes requested):

docs/configuration/data-validation.md
LHBL2003: OK I have the connection between the:
!!! tip "Dynamic Configuration Parameter"

and not understood, as there is no link referencing it.
https://demo.netbox.dev/static/docs/configuration/#dynamic-configuration-parameters

It was only through your comment and Google that I found the connection again.

If there was a link to dynamic-configuration-parameters, you could probably understand it better.

I would like to adapt the tip like this:
!!! tip [`Dynamic Configuration Parameter`](./index.md#dynamic-configuration-parameters)

docs/configuration/data-validation.md
LHBL2003: I think the example is not well suited or not well described for someone who deals with the topic.

If I understand it correctly, it mixes 3 complexities together.

Example: The site can only be deleted if the status is: decommissioning. 

**Notation for Configuration.py file:**

PROTECTION_RULES = {
    ‘dcim.site": [
        {
            ‘status": {
                ‘eq": ’decommissioning’
            }
        },
    ]
}

**Notation for web frontend:**

‘dcim.site": [
    {
        ‘status": {
            ‘eq": ’decommissioning’
        }
    }
 ]

**Notation for Configuration.py with further validation of custom plugins:**

PROTECTION_RULES = {
    ‘dcim.site": [
        {
            ‘status": {
                ‘eq": ’decommissioning’
            }
        },
        ‘my_plugin.validators.Validator1’,
    ]
}

The problem with the original example is that it does not work out of the box.
Only after many conversations do you find out what is too much.

**‘my_plugin.validators.Validator1’,:** 
Does not work at all for beginners. It is something for advanced users.
**Result:** An example with this line should be listed separately. As it prevents the part from not working that would always work.

**PROTECTION_RULES = {:** 
Where does it say that this must be omitted in the web frontend? After being told that you can insert this example into the web frontend, it does not work because ‘PROTECTION_RULES = {’ must be omitted. But as a beginner you don't know that.
**Result:**: If not here, then it must be written somewhere else that the outermost bracket in the web frontend is omitted if it is inserted in the respective parameter fields.

docs/configuration/data-validation.md
LHBL2003: Yes ok you could leave it at Site. But I think the example is much more interesting for devices.
You decide if it bothers you, it can stay on site.

docs/configuration/data-validation.md
LHBL2003: Is it ok if I adjust all these tips?

Before: `!!! tip ‘Dynamic Configuration Parameter’`
After: `!!! tip [`Dynamic Configuration Parameter`](./index.md#dynamic-configuration-parameters)`

docs/configuration/data-validation.md
LHBL2003: Here I have an idea. I will undo this and create a hint text that you should leave out this line for the entry. Can you give me a short description of this line so that beginners can understand it better?
Then I would describe the meaning of this line.

Something like: This line is used to integrate a validation of its own plugin. Perhaps also a link to the topic?

docs/configuration/data-validation.md
LHBL2003: I have an idea, I will reverse this example and integrate the variant for Device. Then you can see how to integrate multiple validations.
If that's OK?

docs/configuration/data-validation.md
LHBL2003: I have an idea,

in the chapter ‘Dynamic Configuration Parameter’ I would like to point out that the outer bracket e.g. PROTECTION_RULES = { ..} is relevant within the Configuration.py. In the web frontend, however, this should be removed from the examples. As the parameter in the web front end already represents this.

Would that be ok?

PR #18039, opened 2024-11-18
Fixes: #18038 - Ensure DeviceType._abs_weight is stored as an integer

### LINKED ISSUE #18038: DeviceType._abs_weight being stored as float value though it is a PositiveBigIntegerField
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.6

### Python Version

3.11

### Steps to Reproduce

It is possible to store a non-integer value in `WeighMixin._abs_weight` fields (which are `PositiveBigIntegerField`) prior to the object being persisted to the database. This potentially causes deserialization errors when populating a new object via a core serializer which uses the field's `to_python` method.

### Expected Behavior

The value of `DeviceType._abs_weight` should be the same (an integer value) both before and after persistence and refresh from the DB.

### Observed Behavior

```
In [9]: dt.weight = 10.234567

In [10]: dt.save()

In [11]: dt._abs_weight
Out[11]: 10.234567

In [12]: dt.refresh_from_db()

In [13]: dt._abs_weight
Out[13]: 10
```

As we see here, a `DeviceType` object populated with a non-integer weight will have that non-integer value assigned to `_abs_weight` until the object is persisted to the database and refreshed. If, however, the `DeviceType` object is populated with data and not refreshed from the DB before it is used in another operation (such as deserialization into another `DeviceType` object), a `DeserializationError` is raised from the field's `to_python` code which expects an integer but receives a float.

The cause is this code:
```

        # Store the given weight (if any) in grams for use in database ordering
        if self.weight and self.weight_unit:
            self._abs_weight = to_grams(self.weight, self.weight_unit)
        else:
            self._abs_weight = None

        super().save(*args, **kwargs)
```

The output of `to_grams` should be coerced to an int before being assigned to `self._abs_weight`.

**Note**: Moved from https://github.com/netboxlabs/netbox-branching/issues/162

bctiemann:
### Fixes: #18038

`WeightMixin` provides an `_abs_weight` field which stores the normalized weight in grams. It is a `PositiveBigIntegerField`, yet the value stored in it is the unvalidated output of `to_grams`, which may be a float. This does not normally cause issues in the app, but in the case of netbox-branching, the field validation fails when the field value is synced from one branch to another because it expects an integer but receives a float.

This fix adds type annotation and consistent return values to `to_grams`.

### DIFF
--- a/netbox/utilities/conversion.py
+++ b/netbox/utilities/conversion.py
@@ -10,9 +10,9 @@ __all__ = (
 )
 
 
-def to_grams(weight, unit) -> float:
+def to_grams(weight, unit) -> int:
     """
-    Convert the given weight to kilograms.
+    Convert the given weight to integer grams.
     """
     try:
         if weight < 0:
@@ -21,13 +21,13 @@ def to_grams(weight, unit) -> float:
         raise TypeError(_("Invalid value '{weight}' for weight (must be a number)").format(weight=weight))
 
     if unit == WeightUnitChoices.UNIT_KILOGRAM:
-        return weight * 1000
+        return int(weight * 1000)
     if unit == WeightUnitChoices.UNIT_GRAM:
-        return weight
+        return int(weight)
     if unit == WeightUnitChoices.UNIT_POUND:
-        return weight * Decimal(453.592)
+        return int(weight * Decimal(453.592))
     if unit == WeightUnitChoices.UNIT_OUNCE:
-        return weight * Decimal(28.3495)
+        return int(weight * Decimal(28.3495))
     raise ValueError(
         _("Unknown unit {unit}. Must be one of the following: {valid_units}").format(
             unit=unit,
--- a/netbox/utilities/conversion.py
+++ b/netbox/utilities/conversion.py
@@ -10,7 +10,7 @@ __all__ = (
 )
 
 
-def to_grams(weight, unit):
+def to_grams(weight, unit) -> float:
     """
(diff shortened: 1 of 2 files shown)


netbox/utilities/conversion.py
```diff
@@ -10,7 +10,7 @@
 )
 
 
-def to_grams(weight, unit):
+def to_grams(weight, unit) -> float:
     """
     Convert the given weight to kilograms.
```
jeremystretch, line 15:
```suggestion
    Convert the given weight to grams.
```

netbox/utilities/conversion.py
```diff
@@ -10,7 +10,7 @@
 )
 
 
-def to_grams(weight, unit):
+def to_grams(weight, unit) -> float:
```
jeremystretch, line 13: IMO it would make more sense for `to_grams()` to return an integer value. It's only used to populate integer `_abs_weight` fields and there doesn't seem to be any practical need to store fractional grams.

jeremystretch (changes requested):

netbox/utilities/conversion.py
bctiemann: Probably right; though cleaning up `to_grams` also suggests `to_meters` needs some attention to ensure it's always returning either a float or a Decimal consistently. Probably for a different (housekeeping) ticket though.

jeremystretch (approved):

PR #18041, opened 2024-11-18
Fixes: #18037 - Bound VLANGroup VLAN ID max by `VLAN_VID_MAX`

### LINKED ISSUE #18037: vlan group with large ID makes UI crash
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.6

### Python Version

3.12

### Steps to Reproduce

1. deploy netbox (fresh install)
2. create vlan group with [2, 409411115) as bounds, save
3. visit `/ipam/vlan-groups/`

Notes:
- we deploy Netbox on Kubernetes (using official Helm chart)

### Expected Behavior

Get an overview with that vlangroup

### Observed Behavior

- at first, we observed netbox getting killed by kernel cgroup OOM killer (limited by Kubelet to 1.5GiB)
- when increased to 2GiB, the process just died quietly, without any logging
- although it's possible to remove the offending VLAN group using the API, it does not remove symptom (crash when visit VLAN group in UI)
- the problem is finally solved by deleting the offending record form `ipam_vlangroup` in PostgreSQL

DanSheps:
### Fixes: #18037 - Restrict VLANGroup VID max to 4094

* Bound VLANGroup VLAN ID max by `VLAN_VID_MAX`

### DIFF
--- a/netbox/ipam/models/vlans.py
+++ b/netbox/ipam/models/vlans.py
@@ -98,19 +98,25 @@ class VLANGroup(OrganizationalModel):
 
         # Validate VID ranges
         for vid_range in self.vid_ranges:
-            if vid_range.lower < VLAN_VID_MIN:
+            lower_vid = vid_range.lower if vid_range.lower_inc else vid_range.lower + 1
+            upper_vid = vid_range.upper if vid_range.upper_inc else vid_range.upper - 1
+            if lower_vid < VLAN_VID_MIN:
                 raise ValidationError({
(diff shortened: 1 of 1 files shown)


DanSheps:
To be clear, this does not fix the list issue, I am not sure where that code is going to come into play (Checked the annotation on the custom queryset and the table and nothing there looks to cause any issues), however this does prevent VID's > 4094 from being created.

jeremystretch (approved):

PR #18062, opened 2024-11-21
Fixes: #17950 - Handle InvalidJobOperation error in job enqueueing test

### LINKED ISSUE #17950: Fix randomly failing test `test_enqueue_once_after_enqueue`
### Proposed Changes

The test `netbox.tests.test_jobs.EnqueueTest.test_enqueue_once_after_enqueue` occasionally fails for an unknown reason (see [this example](https://github.com/netbox-community/netbox/actions/runs/11724028782/job/32656992871?pr=17916)). This needs to be investigated and resolved.

### Justification

CI tests should always pass reliably.

bctiemann:
### Fixes: #17950

This alters the `Job.delete()` method to wrap `job.cancel()` in a try-except which handles the condition where the inner `get_status` call fails to find a non-null status for the job in Redis. This should fix a recurring intermittent test failure in `test_enqueue_once_after_enqueue`.

My investigation showed that whereas the job UUID key must be in Redis in order for `fetch_job` to work (which occurs in `delete` prior to the `cancel` call), it is occasionally the case that a call to `self.connection.hget(self.key, 'status')` returns a null value, leading to the exception being raised in `rq/job.py#L400`. This never occurs locally, and may be a side effect of concurrency within the CI environment.

The risk of this fix is low, I think, because the `job.cancel` is only being used in order to clear out a job from the cache during a `delete` operation; leaving a job in undeleted or partially-deleted state (particularly specifically in CI) is unlikely to have negative side effects.

### DIFF
--- a/netbox/core/models/jobs.py
+++ b/netbox/core/models/jobs.py
@@ -9,6 +9,7 @@ from django.db import models
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext as _
+from rq.exceptions import InvalidJobOperation
 
 from core.choices import JobStatusChoices
 from core.models import ObjectType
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @bctiemann!

PR #18075, opened 2024-11-22
Show virtual ips in interfaces table

### LINKED ISSUE #18070: Add column in Interfaces tables for Virtual IPs in assigned FHRP groups
### NetBox version

v4.1.6

### Feature type

New functionality

### Triage priority

### Proposed functionality

Add an available column for the Interfaces table that displays a list of the Virtual IPs assigned to the FHRP groups assigned to the interface.

The new column would look like the existing `IP Adresses` column, expect it would show the associated virtual IPs instead of the directly assigned IPs.

### Use case

Being able to see the virtual IPs associated with an interface is useful for the same reasons it's useful to see the directly assigned IP addresses.

I think it's especially useful when modeling HA clusters in something like a Fortigate Firewall, where the individual unit interfaces don't have dedicated IPs at all. Without seeing the Virtual IPs, you have no immediate indication of what net is on the interface at all. The existing column for showing the FHRP group is not as useful as seeing the Virtual IPs for my use case.

I have a PR ready that is running on my own instance.

### Database changes

None

### External dependencies

None

mskalecki:
### Fixes: #18070

Adds a column named `Virtual IPs` to the interfaces table. 

The new column behaves almost exactly like the existing column for directly assigned IP addresses, except that it doesn't limit the number of virtual IPs displayed directly on the table. Since the virtual IPs could come from multiple FHRP Groups, I don't see an easy way to link to a filtered table of IPs that are virtual IPs of the interface.

If that is a problem, I can limit the number that are displayed and add an elipsis (or something to indicate that the list is cropped), or maybe someone with real django experience can figure it out.

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -643,8 +643,8 @@ class InterfaceTable(ModularDeviceComponentTable, BaseInterfaceTable, PathEndpoi
             'speed', 'speed_formatted', 'duplex', 'mode', 'mac_address', 'wwn', 'poe_mode', 'poe_type', 'rf_role', 'rf_channel',
             'rf_channel_frequency', 'rf_channel_width', 'tx_power', 'description', 'mark_connected', 'cable',
             'cable_color', 'wireless_link', 'wireless_lans', 'link_peer', 'connection', 'tags', 'vdcs', 'vrf', 'l2vpn',
-            'tunnel', 'ip_addresses', 'virtual_ips', 'fhrp_groups', 'untagged_vlan', 'tagged_vlans', 'inventory_items', 'created',
-            'last_updated',
+            'tunnel', 'ip_addresses', 'virtual_ips', 'fhrp_groups', 'untagged_vlan', 'tagged_vlans', 'inventory_items',
+            'created', 'last_updated',
         )
         default_columns = ('pk', 'name', 'device', 'label', 'enabled', 'type', 'description')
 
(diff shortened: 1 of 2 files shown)


netbox/dcim/tables/devices.py
```diff
@@ -557,6 +557,12 @@ class BaseInterfaceTable(NetBoxTable):
         orderable=False,
         verbose_name=_('IP Addresses')
     )
+    virtual_ips = tables.TemplateColumn(
+        accessor=Accessor('fhrp_group_assignments'),
```
jeremystretch, line 561:
```suggestion
        accessor=Accessor('fhrp_group_assignments__ip_addresses'),
```

We should reference the IP assignments directly to ensure they are prefetched when the table column is displayed.

netbox/dcim/tables/devices.py
```diff
@@ -557,6 +557,12 @@ class BaseInterfaceTable(NetBoxTable):
         orderable=False,
         verbose_name=_('IP Addresses')
     )
+    virtual_ips = tables.TemplateColumn(
```
jeremystretch, line 560: This can be a `ManyToManyColumn` with `linkify_item=True` IMO. We should avoid custom template code where possible.

jeremystretch, line 560: I suggest renaming this to `fhrp_group_ips` or similar, as "virtual" IPs may not be quite clear to the user.

netbox/dcim/tables/devices.py
```diff
@@ -675,7 +681,7 @@ class Meta(DeviceComponentTable.Meta):
             'mgmt_only', 'mtu', 'mode', 'mac_address', 'wwn', 'rf_role', 'rf_channel', 'rf_channel_frequency',
             'rf_channel_width', 'tx_power', 'description', 'mark_connected', 'cable', 'cable_color', 'wireless_link',
             'wireless_lans', 'link_peer', 'connection', 'tags', 'vdcs', 'vrf', 'l2vpn', 'tunnel', 'ip_addresses',
-            'fhrp_groups', 'untagged_vlan', 'tagged_vlans', 'actions',
+            'virtual_ips', 'fhrp_groups', 'untagged_vlan', 'tagged_vlans', 'actions',
```
jeremystretch, line 684: This also needs to be added to VirtualMachineVMInterfaceTable.

jeremystretch (changes requested):

PR #18077, opened 2024-11-23
fixes 17465 : add racktype on bulkimport and bulkedit of racks

### LINKED ISSUE #17465: Bulk Import of Racks based on Rack Type
### NetBox version

v4.1.0

### Feature type

### Proposed functionality

With the new feature to introduce Rack Types into NetBox. It'd be useful if we could then refer to these Rack Types when doing a Bulk Import of Racks. 

Currently the Field Options do not allow for the Rack Type to be used in place of specifying the Post Width and Rack Height in RU, which means that when performing a Bulk Import or Creation of racks, they all have to be defined separately. 

This both means that the Rack Types can't be used when performing a first-time database setup, and every rack needs to be defined manually. Also means that the Racks aren't associated with a Rack Type when using this method. 

### Use case

This would enable users to do a Bulk Import of Racks into their NetBox environment (either via API or the GUI) using the Rack Type as a template, rather than having to define the parameters manually for each record. This is a particular use case for new environments being managed by NetBox, who may require a large number of Racks to be populated as part of their initial setup. 

This would also enable all the Racks created under this method to be linked to a Rack Type which would help improve the consistency of NetBox data for users. 

### Database changes

_No response_

### External dependencies

_No response_

pl0xym0r:
Fixes #17465 

add racktype on bulkimport of rack

import

add racktype on bulkedit of rack

edit

### DIFF
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -272,7 +272,12 @@ class RackImportForm(NetBoxModelImportForm):
     width = forms.ChoiceField(
         label=_('Width'),
         choices=RackWidthChoices,
+        required=False,
         help_text=_('Rail-to-rail width (in inches)')
+    )
+    u_height = forms.IntegerField(
+        required=False,
+        label=_('Height (U)')
     )
     outer_unit = CSVChoiceField(
         label=_('Outer unit'),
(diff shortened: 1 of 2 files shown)


netbox/dcim/forms/bulk_import.py
```diff
@@ -256,6 +256,13 @@ class RackImportForm(NetBoxModelImportForm):
         to_field_name='name',
         help_text=_('Name of assigned role')
     )
+    rack_type = CSVModelChoiceField(
```
jeremystretch, line 259: `width` and `u_height` should no longer be required if `rack_type` is supported (although the form should have logic to check that these are set if `rack_type` is not).

jeremystretch (changes requested):

netbox/dcim/forms/bulk_import.py
pl0xym0r: I honestly don't know how to show in the view the "OR" option --> ( `width` and `u_height`  mandatory  and rack_type optional ) "OR"  (`rack_type` and ( `width` and `u_height`  optional ) )
Or you prefer to have `width` and `u_height` not required anymore in the form AND add validation/check on values passed?

jeremystretch: We just need to mark them as `required=False` and add some conditional validation logic. I've addressed this with the latest commit.

bctiemann (approved):

jeremystretch (approved):

PR #18084, opened 2024-11-25
Fixes: #17820 - Store default values from custom fields on newly created module components

### LINKED ISSUE #17820: Custom fields default value on Objects from modules are not set
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.4

### Python Version

3.12

### Steps to Reproduce

1. Create a required custom field and set a default value. Choose an object like DCIM > Interface.
```json
{
    "id": 8,
    "display": "My cf",
    "object_types": [
        "dcim.interface"
    ],
    "type": {
        "value": "text",
        "label": "Text"
    },
    "related_object_type": null,
    "data_type": "string",
    "name": "My_CF",
    "label": "",
    "group_name": "",
    "description": "",
    "required": true,
    "unique": false,
    "search_weight": 1000,
    "filter_logic": {
        "value": "loose",
        "label": "Loose"
    },
    "ui_visible": {
        "value": "always",
        "label": "Always"
    },
    "ui_editable": {
        "value": "yes",
        "label": "Yes"
    },
    "is_cloneable": false,
    "default": "default_value",
    "related_object_filter": null,
    "weight": 100,
    "validation_minimum": null,
    "validation_maximum": null,
    "validation_regex": "",
    "choice_set": null,
    "comments": "",
}
```
2.Create a module type with a few interfaces.

3. Create a device, create a module bay on this device.

4. Connect a module with the module type

Now, we have the interfaces from the type module that have been added to the device.

And if we see an Interface:

We can see that the default value has not been set.

"custom_fields": {
        "My_CF": null
    },
Note that in devices with device_type, interface custom fields (for example) are set to the default value

### Expected Behavior

We should have the default value and not null. on the same way as on a device type with interfaces, when the device is created, there will be the default value on the interfaces custom field.

### Observed Behavior

Interfaces that are created when the module is connected to a device, the default values of custom fields are not set to their default values.

bctiemann:
### Fixes: #17820

Fixes a piece of missing functionality in the `Module.save()` method where custom field default values are not applied during creation of new module-templated components.

Does not apply to update operations, only create.

### DIFF
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -1277,6 +1277,11 @@ class Module(PrimaryModel, ConfigContextModel):
                 if not disable_replication:
                     create_instances.append(template_instance)
 
+            for component in create_instances:
+                # Set default values for any applicable custom fields
+                if cf_defaults := CustomField.objects.get_defaults_for_model(component_model):
+                    component.custom_field_data = cf_defaults
+
             if component_model is not ModuleBay:
                 component_model.objects.bulk_create(create_instances)
                 # Emit the post_save signal for each newly created object
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -1277,9 +1277,9 @@ class Module(PrimaryModel, ConfigContextModel):
                 if not disable_replication:
                     create_instances.append(template_instance)
 
-            for component in create_instances:
-                # Set default values for any applicable custom fields
-                if cf_defaults := CustomField.objects.get_defaults_for_model(component_model):
+            # Set default values for any applicable custom fields
+            if cf_defaults := CustomField.objects.get_defaults_for_model(component_model):
+                for component in create_instances:
                     component.custom_field_data = cf_defaults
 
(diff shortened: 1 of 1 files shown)


netbox/dcim/models/devices.py
```diff
@@ -1277,6 +1277,11 @@ def save(self, *args, **kwargs):
                 if not disable_replication:
                     create_instances.append(template_instance)
 
+            for component in create_instances:
+                # Set default values for any applicable custom fields
+                if cf_defaults := CustomField.objects.get_defaults_for_model(component_model):
```
jeremystretch, line 1282: Can we flip this with the statement above to avoid calling it for each individual component?

jeremystretch (changes requested):

netbox/dcim/models/devices.py
bctiemann: Ooh, good call. My eye skimmed over `component_model` and read it as `component`.

jeremystretch (approved):

PR #18097, opened 2024-11-25
fix #17960 by adding 6 more tunnel encap options

### LINKED ISSUE #17960: Add additional tunnel encapsulation choices: PPTP, L2TP, EoIP, SSTP
### NetBox version

v4.1.4

### Feature type

### Triage priority

### Proposed functionality

I'd like to add PPTP, L2TP, and EoIP as Tunnel Encapsulation Choices. PPTP, and L2TP are standards based tunnel protocols while EoIP is a MikroTik proprietary protocol and SSTP is a Microsoft proprietary protocol.  

PPTP: https://datatracker.ietf.org/doc/html/rfc2637  
L2TP: https://datatracker.ietf.org/doc/html/rfc2661
EoIP: https://help.mikrotik.com/docs/spaces/ROS/pages/24805521/EoIP
SSTP: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-sstp/c50ed240-56f3-4309-8e0c-1644898f0ea8

As far as I can tell the only change here would be to add these fourto the choices.py here: https://github.com/netbox-community/netbox/blob/develop/netbox/vpn/choices.py 

### Use case

This lets users model more accurately their VPN tunnels. Right now the choices don't have an other so you have to use an incorrect Encapsulation choice.  

### Database changes

_No response_

### External dependencies

_No response_

jmcguir:
Fixes: #17960

I included all 6 types suggested in the issue. I did not include an "other" because it seems like that's against house style (there are a limited number of tunnel encap types, we could get them all in netbox). 

They should now show as an option as a tunnel encap. I wasn't sure what the proper style was for the order of the choices so I sorted them by what I thought was most common. Happy to change the PR to some other sort.

### DIFF
--- a/netbox/vpn/choices.py
+++ b/netbox/vpn/choices.py
@@ -22,16 +22,28 @@ class TunnelStatusChoices(ChoiceSet):
 
 
 class TunnelEncapsulationChoices(ChoiceSet):
+    ENCAP_EOIP = 'eoip'
     ENCAP_GRE = 'gre'
-    ENCAP_IP_IP = 'ip-ip'
     ENCAP_IPSEC_TRANSPORT = 'ipsec-transport'
     ENCAP_IPSEC_TUNNEL = 'ipsec-tunnel'
+    ENCAP_IP_IP = 'ip-ip'
+    ENCAP_L2TP = 'l2tp'
+    ENCAP_OPENVPN = 'openvpn'
+    ENCAP_PPTP = 'pptp'
+    ENCAP_SSTP = 'sstp'
+    ENCAP_WIREGUARD = 'wireguard'
 
     CHOICES = [
         (ENCAP_IPSEC_TRANSPORT, _('IPsec - Transport')),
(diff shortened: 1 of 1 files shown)


netravnen:
I would not recommend simply adding the OpenVPN and Wireguard to the choice field.

From comment https://github.com/netbox-community/netbox/discussions/14683#discussioncomment-9072788 a set of fields are necessary to specify the crypto information for WireGuard tunnels. I will expect the same to be necessary for OpenVPN tunnels using Client & Server Certificates.

jmcguir:
I don't have any feelings on OpenVPN or WireGuard. I'm happy to remove them from the commit but I don't currently use or understand them so if more work is needed on the crypto side modeling, I'm not the guy unfortunately.

jeremystretch:
@jmcguir please see my comment [here](https://github.com/netbox-community/netbox/issues/17960#issuecomment-2523396417) regarding the proposed additions.

jeremystretch (approved):
Thanks @jmcguir!

Issue #18066, opened 2024-11-21
Render Config fails with DataFile templates using dynamic include statements


ed-ud:
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.4

### Python Version

3.12

### Steps to Reproduce

1. Create a Data Source with several Jinja2 template files:

start.j2:
{%- set PATH = 'my/prefix/' %}
{%- include PATH + 'test.j2' %}

my/prefix/test.j2:
SOME RENDERED TEXT

2. Create Config Template using the Data Source and the Data File start.j2
3. Attach Config Template to a Device
4. Select the Render Config tab on the Device

### Expected Behavior

Should not fail to render.

### Observed Behavior

Fails to render with TemplateNotFound 'my/prefix/test.j2' exception.

ed-ud:
A potential fix is the recognize when the Jinja template files are using dynamic includes (the list of referenced_templates contains None entries) and in that case, include all DataFiles from the DataSource. 
In utilities/jinja2.py:
`
diff jinja2.py.orig jinja2.py
32,35c32,41
<             self.cache_templates({
<                 df.path: df.data_as_string for df in
<                 DataFile.objects.filter(source=self.data_source, path__in=referenced_templates)
<             })
---
>             if None in referenced_templates:
>                 self.cache_templates({
>                     df.path: df.data_as_string for df in
>                     DataFile.objects.filter(source=self.data_source)
>                 })
>             else:
>                 self.cache_templates({
>                     df.path: df.data_as_string for df in
>                     DataFile.objects.filter(source=self.data_source, path__in=referenced_templates)
>                 })
`

pl0xym0r:
same as #17490

jeremystretch:
Thank you for submitting this issue, however it appears that this topic has already been raised. Please see issue #17490 for further discussion.

PR #18106, opened 2024-11-26
Fixes #17490: Config Template unable to dynamically include templates

### LINKED ISSUE #17490: Config Template unable to dynamically include templates
### Deployment Type

Self-hosted

### NetBox Version

v4.0.9

### Python Version

3.10

### Steps to Reproduce

1. Create the following template in a data store, as well as child templates in the blocks directory.  In this example a template named interface.j2 is located in the blocks/TEST/ directory.  
2. Assign the template to a device
3. Attempt to render the template

Example of template.  This has been tested manually using Jinja2. The configuration blocks referenced are from a plugin.  

config system interface
{% for config_block in device.site.configblock_set.all() %}
{% include "blocks/" + custom_block.name + "/interface.j2" %}
{% endfor %}
end

### Expected Behavior

Expected behaviour is that the template renders and imports the relevant includes. 

### Observed Behavior

  File "/opt/netbox/netbox/utilities/jinja2.py", line 26, in get_source
    template_source = self._template_cache[template]
KeyError: 'blocks/TEST/interface.j2'

  File "/opt/netbox/netbox/dcim/views.py", line 2060, in get_extra_context
  File "/opt/netbox/netbox/extras/models/configs.py", line 282, in render
    output = template.render(**_context)
  File "/usr/local/lib/python3.10/site-packages/jinja2/environment.py", line 1304, in render
    self.environment.handle_exception()
  File "/usr/local/lib/python3.10/site-packages/jinja2/environment.py", line 939, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "block-config.j2", line 49, in top-level template code
  File "/opt/netbox/netbox/utilities/jinja2.py", line 28, in get_source
    raise TemplateNotFound(template)
jinja2.exceptions.TemplateNotFound: blocks/TEST/interface.j2

robduffy2010:
### Fixes: #17490

Config templates are unable to dynamically include child templates.  This is a proposed fix as suggested by @ed-ud in #18066 which was closed as a duplicate.

### DIFF
--- a/netbox/utilities/jinja2.py
+++ b/netbox/utilities/jinja2.py
@@ -29,7 +29,6 @@ class DataFileLoader(BaseLoader):
 
         # Find and pre-fetch referenced templates
         if referenced_templates := tuple(find_referenced_templates(environment.parse(template_source))):
-            if None in referenced_templates:
             related_files = DataFile.objects.filter(source=self.data_source)
             # None indicates the use of dynamic resolution. If dependent files are statically
             # defined, we can filter by path for optimization.
--- a/netbox/utilities/jinja2.py
+++ b/netbox/utilities/jinja2.py
@@ -29,10 +29,16 @@ class DataFileLoader(BaseLoader):
 
         # Find and pre-fetch referenced templates
         if referenced_templates := find_referenced_templates(environment.parse(template_source)):
-            self.cache_templates({
-                df.path: df.data_as_string for df in
-                DataFile.objects.filter(source=self.data_source, path__in=referenced_templates)
-            })
+            if None in referenced_templates:
+                self.cache_templates({
+                    df.path: df.data_as_string for df in
+                    DataFile.objects.filter(source=self.data_source)
+                })
+            else:
+                self.cache_templates({
+                    df.path: df.data_as_string for df in
+                    DataFile.objects.filter(source=self.data_source, path__in=referenced_templates)
+                })
 
         return template_source, template, lambda: True
 
--- a/netbox/utilities/jinja2.py
+++ b/netbox/utilities/jinja2.py
@@ -28,7 +28,7 @@ class DataFileLoader(BaseLoader):
             raise TemplateNotFound(template)
 
         # Find and pre-fetch referenced templates
-        if referenced_templates := find_referenced_templates(environment.parse(template_source)):
+        if referenced_templates := tuple(find_referenced_templates(environment.parse(template_source))):
             if None in referenced_templates:
                 self.cache_templates({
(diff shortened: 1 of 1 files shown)


netbox/utilities/jinja2.py
```diff
@@ -29,10 +29,16 @@ def get_source(self, environment, template):
 
         # Find and pre-fetch referenced templates
         if referenced_templates := find_referenced_templates(environment.parse(template_source)):
```
jeremystretch, line 31: There's a very odd little bug hidden here when not using dynamic resolution for related templates. We need to cast the generator returned by `find_referenced_templates()` to an iterable to avoid exhausting it on the check for None. (Otherwise it ends up empty when passed as the value for `path__in`.)
```suggestion
        if referenced_templates := tuple(find_referenced_templates(environment.parse(template_source))):
```

netbox/utilities/jinja2.py
```diff
@@ -29,10 +29,16 @@ def get_source(self, environment, template):
...
+            if None in referenced_templates:
+                self.cache_templates({
+                    df.path: df.data_as_string for df in
+                    DataFile.objects.filter(source=self.data_source)
+                })
+            else:
+                self.cache_templates({
+                    df.path: df.data_as_string for df in
+                    DataFile.objects.filter(source=self.data_source, path__in=referenced_templates)
+                })
```
jeremystretch, line 41: We can conditionally apply the `path__in` filter to avoid duplicating code. We should also explain the significance of None in the paths list as its meaning is not obvious IMO.
```suggestion
            related_files = DataFile.objects.filter(source=self.data_source)
            # None indicates the use of dynamic resolution. If dependent files are statically
            # defined, we can filter by path for optimization.
            if None not in referenced_templates:
                related_files = related_files.filter(path__in=referenced_templates)
            self.cache_templates({
                df.path: df.data_as_string for df in related_files
            })
```

jeremystretch (changes requested):
Thanks for tracking this down @robduffy2010!

netbox/utilities/jinja2.py
ed-ud: Nice catch!

robduffy2010:
Just a heads up that the workflow ran before I had finished all the changes.  Is it possible to run it again and review?

netbox/utilities/jinja2.py
jeremystretch: Django querysets are lazy; they won't actually trigger a database query until evaluation has been forced (when the dict is generated).

jeremystretch (approved):

PR #18108, opened 2024-11-26
Fixes #17810: Fix assignment of device to rack via REST API without specifying position/face

### LINKED ISSUE #17810: Can't place multiple devices in rack with non-racked device (using API)
### Deployment Type

Self-hosted

### Triage priority

This is preventing me from using NetBox

### NetBox Version

v4.1.4

### Python Version

3.12

### Steps to Reproduce

1. Create site
2. Create rack in the site (for example with id = 1)
3. Create first non-racked device in rack
4. Create second device
5. Try placing the the device in the rack as non-racked device via API

Important: I can only reproduce the problem using the REST API.

#### API Calls
```
# Create first device, non-racked in the rack (remember the id, ensure there are no devices non-racked before)
curl -X POST -H "authorization: Token TOKEN" -H 'Content-Type: application/json' -d '{"device_type":116,"role":25,"site":1,"rack":168}' https://netbox.example.org/api/dcim/devices/ | jq .id
# Create second device, floating around in the site
curl -X POST -H "authorization: Token TOKEN" -H 'Content-Type: application/json' -d '{"device_type":116,"role":25,"site":1}' https://netbox.example.org/api/dcim/devices/ | jq . | less
# Try moving the device to the rack of the first device non-racked (fails)
curl -X PATCH -H "authorization: Token TOKEN" -H 'Content-Type: application/json' -d '{"rack":168}' https://netbox.example.org/api/dcim/devices/$ID_FROM_THE_FIRST_COMMAND/ | jq . | less
# Run first command again, creating another device, doesn't fail strangely
curl -X POST -H "authorization: Token TOKEN" -H 'Content-Type: application/json' -d '{"device_type":116,"role":25,"site":1,"rack":168}' https://netbox.example.org/api/dcim/devices/ | jq .id
```

### Expected Behavior

Device should be placed in the rack as non-racked.

### Observed Behavior

HTTP 400 Error Code

```
{"non_field_errors": ["The fields rack, position, face must make a unique set."]}
```

jeremystretch:
### Fixes: #17810

This applies some conditions to the unique constraint on the Device model, which prevents DRF from raising a validation error when `rack` is set but `position` is not.

### DIFF
--- a/netbox/dcim/migrations/0192_device_adjust_rack_position_constraint.py
+++ b/netbox/dcim/migrations/0192_device_adjust_rack_position_constraint.py
@@ -1,0 +1,30 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dcim', '0191_module_bay_rebuild'),
+        ('extras', '0121_customfield_related_object_filter'),
+        ('ipam', '0070_vlangroup_vlan_id_ranges'),
+        ('tenancy', '0015_contactassignment_rename_content_type'),
+        ('virtualization', '0040_convert_disk_size'),
+    ]
+
+    operations = [
+        migrations.RemoveConstraint(
+            model_name='device',
+            name='dcim_device_unique_rack_position_face',
+        ),
+        migrations.AddConstraint(
+            model_name='device',
+            constraint=models.UniqueConstraint(
+                condition=models.Q(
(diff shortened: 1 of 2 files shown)


bctiemann (approved):

jeremystretch:
Holding on this for now as we investigate a related issue (#18132) which may require a similar fix.

PR #18127, opened 2024-11-30
Closes 18061: Hide traceback from rendered device config

### LINKED ISSUE #18061: Hide traceback in rendered config
### NetBox version

v4.1.6

### Feature type

### Triage priority

### Proposed functionality

When a rendered configuration template raises an exception, the full traceback is included in the Rendered Config pane. I suggest limiting this to the exception type and message.

### Use case

End users without a programming background might be confused by seeing the full traceback. Seeing only the error message may be more helpful to them for resolving the problem.

### Database changes

None.

### External dependencies

None.

alehaa:
### Fixes: #18061

When an exception occurs during device configuration rendering, it usually doesn't contain information about the template being rendered, but rather the trace of **how** the template was rendered. Since this could confuse users and expose internal server information, it is now hidden.

#### Before:

```
  File "/opt/netbox/netbox/dcim/views.py", line 2129, in get_extra_context
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/extras/models/configs.py", line 282, in render
    output = template.render(**_context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/jinja2/environment.py", line 1304, in render
    self.environment.handle_exception()
  File "/opt/netbox/venv/lib/python3.12/site-packages/jinja2/environment.py", line 939, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "<template>", line 1, in top-level template code
  File "/opt/netbox/venv/lib/python3.12/site-packages/jinja2/sandbox.py", line 327, in getattr
    value = getattr(obj, attribute)
            ^^^^^^^^^^^^^^^^^^^^^^^
jinja2.exceptions.UndefinedError: 'dcim.models.devices.Device object' has no attribute 'next'
```

#### After:

```
An error occurred while rendering the template: 'dcim.models.devices.Device object' has no attribute 'next'
```

This is the same message as displayed in the message notification.

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -2104,7 +2104,8 @@ class DeviceRenderConfigView(generic.ObjectView):
         # If a direct export has been requested, return the rendered template content as a
         # downloadable file.
         if request.GET.get('export'):
-            response = HttpResponse(context['rendered_config'], content_type='text')
+            content = context['rendered_config'] or context['error_message']
+            response = HttpResponse(content, content_type='text')
             filename = f"{instance.name or 'config'}.txt"
             response['Content-Disposition'] = f'attachment; filename="{filename}"'
             return response
@@ -2122,18 +2123,18 @@ class DeviceRenderConfigView(generic.ObjectView):
 
         # Render the config template
         rendered_config = None
+        error_message = None
         if config_template := instance.get_config_template():
             try:
                 rendered_config = config_template.render(context=context_data)
             except TemplateError as e:
-                msg = _("An error occurred while rendering the template: {error}").format(error=e)
-                messages.error(request, msg)
-                rendered_config = msg
+                error_message = _("An error occurred while rendering the template: {error}").format(error=e)
 
         return {
             'config_template': config_template,
             'context_data': context_data,
             'rendered_config': rendered_config,
+            'error_message': error_message,
         }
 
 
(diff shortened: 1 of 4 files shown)


bctiemann:
@alehaa Maybe we should also log the complete traceback via the regular `logger` mechanism, instead of just handling it in the UI and otherwise swallowing it? In your example the `File "/opt/netbox/netbox/dcim/views.py", line 2129, in get_extra_context` gives a much more useful place to start looking for the issue than if the ultimate error line is the only thing one has to go on.

jeremystretch (commented):
IMO it would be cleaner to introduce a separate context variable to store the error message, and return an empty string for `rendered_config`. This will allow us to more cleanly convey the error in the UI, with proper styling & formatting.

However, we also need to consider the export workflow, where the raw config is returned for download as a file. In this case maybe it's best to keep the error message in the file content.

alehaa:
I‘ll see on how to integrate changes for both UI and API. Due to some other projects an update might take until the holidays. I‘ll keep you posted.

jeremystretch (approved):

jeremystretch:
I went ahead with my proposed changes above, pretty happy with how it turned out. I've also replicated the changes for virtual machines.

bctiemann (approved):

PR #18135, opened 2024-12-02
Fixes: #18032 - Clear ordering preferences if set to a nonexistent field

### LINKED ISSUE #18032: Server error after updated object
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.0.7

### Python Version

3.11

### Steps to Reproduce

I've an issue when I try to access to Virtual Device Contexts main page : 

Server Error

Cannot resolve keyword '_name' into field. Choices are: bookmarks, comments, created, custom_field_data, description, device, device_id, id, identifier, interface_count, interfaces, journal_entries, last_updated, name, primary_ip4, primary_ip4_id, primary_ip6, primary_ip6_id, status, tagged_items, tags, tenant, tenant_id

Python version: 3.11.6
NetBox version: 4.0.7
Plugins: None installed

I can list VDC using to api call /api/dcim/virtual-device-contexts/
I can access to a specific  VDC ID using GUI and API

### Expected Behavior

I'm supposed to display the list of all VDC on GUI

### Observed Behavior

Server Error

Cannot resolve keyword '_name' into field. Choices are: bookmarks, comments, created, custom_field_data, description, device, device_id, id, identifier, interface_count, interfaces, journal_entries, last_updated, name, primary_ip4, primary_ip4_id, primary_ip6, primary_ip6_id, status, tagged_items, tags, tenant, tenant_id

Python version: 3.11.6
NetBox version: 4.0.7
Plugins: None installed

bctiemann:
### Fixes: #18032

It is possible to have a user preference set to a field which no longer exists in a new version of NetBox after an upgrade. Normally django-tables2 validates all `sort=<column>` query params against the table's `columns` and filters out any invalid values prior to storage as a preference; but if a `sort` value that was previously (but is no longer) valid has been stored in the preferences, a `FieldError` will be raised under the new version.

This change catches that `FieldError` and clears the preference accordingly.

### DIFF
--- a/netbox/netbox/tables/tables.py
+++ b/netbox/netbox/tables/tables.py
@@ -4,7 +4,7 @@ from functools import cached_property
 import django_tables2 as tables
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.contenttypes.fields import GenericForeignKey
-from django.core.exceptions import FieldDoesNotExist
+from django.core.exceptions import FieldDoesNotExist, FieldError
 from django.db.models.fields.related import RelatedField
 from django.db.models.fields.reverse_related import ManyToOneRel
 from django.urls import reverse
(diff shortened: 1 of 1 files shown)


bctiemann:
On further review I don't think this is necessary. I wasn't able to reproduce the crash in question (though I was almost positive I had seen the crash in my initial investigation). The validation in django-tables2 properly catches a saved ordering value in the preferences if it is not a valid accessor and filters it out. The only way I can see the original error happening is if there were some mismatch in the code and the db, where `_name` was still a valid field and thus `if name in self.columns and self.columns[name].orderable` returns True. Because it is not in `self.columns` in the current code, `_name` is removed from the accessors list.

I will close this PR and also close the issue as not planned.

PR #18137, opened 2024-12-02
18136 Allow Script running within a Branch

### LINKED ISSUE #18136: Add branching provider to configuration
### NetBox version

v4.1.7

### Feature type

New functionality

### Triage priority

N/A

### Proposed functionality

Add branching provider support as a configuration parameter and modify the script running code to take into account branching support.

### Use case

This will allow scripts with script parameters to be run while in the context of a branch.  This requires changes to the core of NetBox in the script running code as well as the branching plugin to support this.

### Database changes

N/A

### External dependencies

N/A

arthanson:
### Fixes: #18136 

Adds a config parameter to NetBox for BranchingBackend that is called during script execution to get and switch the active branch.  See also https://github.com/netboxlabs/netbox-branching/pull/184 for changes in the branching plugin to support this.

### DIFF
--- a/netbox/extras/jobs.py
+++ b/netbox/extras/jobs.py
@@ -39,7 +39,6 @@ class ScriptJob(JobRunner):
         """
         logger = logging.getLogger(f"netbox.scripts.{script.full_name}")
         logger.info(f"Running script (commit={commit})")
-        print("running script")
 
         try:
             try:
@@ -53,7 +52,6 @@ class ScriptJob(JobRunner):
                     logger.warning("Script failed")
 
         except Exception as e:
-            print(f"exception: {e}")
             if type(e) is AbortScript:
                 msg = _("Script aborted with error: ") + str(e)
                 if is_report(type(script)):
--- a/netbox/extras/jobs.py
+++ b/netbox/extras/jobs.py
@@ -1,6 +1,6 @@
 import logging
 import traceback
-from contextlib import nullcontext
+from contextlib import ExitStack, nullcontext
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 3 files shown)


bctiemann (dismissed):

netbox/netbox/settings.py
```diff
@@ -179,6 +179,7 @@
 STORAGE_CONFIG = getattr(configuration, 'STORAGE_CONFIG', {})
 TIME_ZONE = getattr(configuration, 'TIME_ZONE', 'UTC')
 TRANSLATION_ENABLED = getattr(configuration, 'TRANSLATION_ENABLED', True)
+BRANCHING_BACKEND = getattr(configuration, 'BRANCHING_BACKEND', None)
```
jeremystretch, line 182: IMO it does not make sense to expose this as a configuration variable, as this functionality is unique to the netbox-branching plugin.

netbox/extras/jobs.py
```diff
@@ -100,5 +103,19 @@ def run(self, data, request=None, commit=True, **kwargs):
...
+                logger = logging.getLogger(f"netbox.scripts.{script.full_name}")
+                message = _("Failed to import configured BRANCHING_BACKEND: ") + settings.BRANCHING_BACKEND
+                logger.error(message)
+                raise ImproperlyConfigured(message)
+
+            branching_backend = branching_cls()
+            branch = branching_backend.get_active_branch(request)
+
+        with branching_backend.activate_branch(branch) if branch else nullcontext():
+            with event_tracking(request) if commit else nullcontext():
```
jeremystretch, line 120: Instead of checking for a branching context manager explicitly, it would be more extensible to allow the configuration of additional arbitrary context managers to be [chained](https://stackoverflow.com/a/30981865/1516387) with `event_tracking()`. This could provide future functionality unrelated to branching.

jeremystretch (changes requested):

jeremystretch:
Closing this in favor of a similar approach in PR #18212 which replaces the configuration parameter with a dynamic registration mechanism. Thanks for your work on this @arthanson!

Issue #1234, opened 2017-05-30
Can I assign an IP address to a VLAN or VLAN group?


sinansahin:
### Issue type: feature request

**Python version: 3.5.2**
**NetBox version: v2.0.4**

Due to the small amount of IP, we are doing nat processing through a single IP on many VLANs in the network.

Can I assign an IP address to a VLAN or VLAN group?

Thank you :)

jeremystretch:
IPs cannot be mapped directly to VLANs. IPs are contained by VRFs and prefixes; each prefix may be assigned to a VLAN.

Are you referring to port address translation (PAT)? NetBox doesn't support modeling PAT and probably won't in the near future (see #788).

jeremystretch:
Closing due to lack of activity

Issue #15842, opened 2024-04-24
Add mechanism to disable form-based authentication in web app


jeffgdotorg:
### NetBox version

v3.7.6

### Feature type

Change to existing functionality

### Proposed functionality

Alter the web app so that having no local users with a password configured completely disables all elements of the default form-based authentication flow. Specifics of proposed behavior:

**Base behavior**
* When NetBox has at least one enabled local user with a password set, current behavior is unchanged
* When NetBox has zero enabled local users with a password set:
  * We never render the form-based login elements in the web app
  * We never render the form-based login elements of the Django admin UI (normally accessible at `/admin/` in NetBox 3.7)
  * POSTs to the `/login/` target in the web app either return a 400-series error (if no session is present) or are redirected with a 300-series response (if a session is present)
  * POSTs to the `/admin/login/` target in the web app either return a 400-series error (if no session is present) or are redirected with a 300-series response (if a session is present). _Although the Django admin UI is disabled by default in 4.0, it can be re-enabled and we must account for this fact._
  * UI elements enabling creation of local users through the web app are disabled
  * GET and POST requests to `/users/users/add` result in a 400-series response which, in the GET case, incorporates **Warning Copy**
  * POST, PUT, and PATCH requests to `/api/users/users` result in a 400-series response which, in the GET case, incorporates **Warning Copy**

**Warning provisions**
* Create a concise but strongly-worded bit of copy (**Warning Copy**) which cautions the user about all relevant impacts of the configuration state described in **Base behavior**, and links to the documentation section described in **Documentation coverage**
* Alter the empty-state copy for the Admin -> Users list view to include **Warning Copy**
* Alter the user individual deletion warning message for the narrow case of deleting the one and only remaining user so that it includes the **Warning Copy**
* Alter the user bulk deletion warning message for the narrow case of ALL users so that it includes **Warning Copy**
* Add a warning message – which includes **Warning Copy** – to the user individual disablement flow for the narrow case of disabling the one and only remaining user

**Documentation coverage**
* Docs explain this behavior in a longer and more thorough version of **Warning Copy**

### Use case

To clear security compliance audits commonly required in key verticals (e.g. SOC2 and ISO-27001), some NetBox users must use OIDC or SAMLv2 exclusively to authenticate users, and must disable form-based authentication completely. Deleting all local users is not sufficient for some of these use cases; removing the POST targets used by form-based login is key in those environments as auditing tools may consider its presence as an unacceptable risk in a layered defense posture.

[...]

PR #18145, opened 2024-12-03
#15842 Setting to hide login form

tobiasge:
### Fixes: #1234, #11672

This doesn't implement the full solution proposed in #15842 but enables administrators to hide the login form when users should only login with a SSO provider. To prevent a complete lockout when the SSO provider is having issues the GET parameter `skipsso` can be added to the login URL to show the form regardless.

This solution would be enough for our needs. Feel free to close, if it is not to your liking.


jeremystretch:

PR #18174, opened 2024-12-07
Fixes: #18021 - Clear Swagger/drf-spectacular API cache on startup

### LINKED ISSUE #18021: Swagger not updating when new API endpoints added in plugin
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.3

### Python Version

3.11

### Steps to Reproduce

1) add an endpoint to an API plugin
2) reload netbox

### Expected Behavior

swagger should update showing the new endpoint

### Observed Behavior

Problem is caused by the caching of the swagger in redis, this needs to be cleared every time in order for the swagger to update. It does appear to load after the cache expires but this takes 24 hours

bctiemann:
### Fixes: #18021

Clears the Swagger API schema cache upon startup, ensuring that API endpoint changes by developers are not cached and do not have to be manually refreshed upon every restart of the app.

### DIFF
--- a/netbox/core/apps.py
+++ b/netbox/core/apps.py
@@ -1,4 +1,5 @@
 from django.apps import AppConfig
+from django.conf import settings
 from django.core.cache import cache
 from django.db import models
 from django.db.migrations.operations import AlterModelOptions
@@ -24,5 +25,6 @@ class CoreConfig(AppConfig):
         # Register models
         register_models(*self.get_models())
 
-        # Clear Swagger API cache on startup
-        cache.delete("*api_schema_*")
+        # Clear Redis cache on startup in development mode
+        if settings.DEBUG:
+            cache.clear()
--- a/netbox/core/apps.py
+++ b/netbox/core/apps.py
@@ -1,4 +1,5 @@
 from django.apps import AppConfig
+from django.core.cache import cache
 from django.db import models
 from django.db.migrations.operations import AlterModelOptions
 
@@ -22,3 +23,6 @@ class CoreConfig(AppConfig):
 
         # Register models
         register_models(*self.get_models())
+
+        # Clear Swagger API cache on startup
+        cache.delete("*api_schema_*")


netbox/core/apps.py
```diff
@@ -22,3 +23,6 @@ def ready(self):
 
         # Register models
         register_models(*self.get_models())
+
+        # Clear Swagger API cache on startup
+        cache.delete("*api_schema_*")
```
jeremystretch, line 28: This does not appear to work:

```
>>> cache.keys('*')
['views.decorators.cache.cache_header.api_schema_4.1.7.13748763de6e90ce226f69bcc045cbbb.en.UTC', 'dashboard_rss_628e2b38c2cec0d4aee7cbf72ba189194251c7fc25bd5e1210115aec39a705f4', 'views.decorators.cache.cache_page.api_schema_4.1.7.GET.13748763de6e90ce226f69bcc045cbbb.5c4af095dde9a1645c6d4c7f96a07693.en.UTC']
>>> cache.delete("*api_schema_*")
False
>>> cache.keys('*')
['views.decorators.cache.cache_header.api_schema_4.1.7.13748763de6e90ce226f69bcc045cbbb.en.UTC', 'dashboard_rss_628e2b38c2cec0d4aee7cbf72ba189194251c7fc25bd5e1210115aec39a705f4', 'views.decorators.cache.cache_page.api_schema_4.1.7.GET.13748763de6e90ce226f69bcc045cbbb.5c4af095dde9a1645c6d4c7f96a07693.en.UTC']
```

I don't think wildcard matching is supported. Deleting by the complete key name works:

```
>>> cache.delete("views.decorators.cache.cache_header.api_schema_4.1.7.13748763de6e90ce226f69bcc045cbbb.en.UTC")
True
```

jeremystretch (changes requested):

netbox/core/apps.py
bctiemann: Ah, thanks. I was going by the pattern from the script in the thread and presuming that had been verified to work.

jeremystretch (approved):

PR #18185, opened 2024-12-09
Fixes 18183 - Hide Light/Dark Mode and Login Info from Printed Pages

### LINKED ISSUE #18183: Hide Light/Dark Mode and Login Info from Printed Pages
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.7

### Python Version

3.10

### Steps to Reproduce

1. Goto any object page at https://demo.netbox.dev/
2. Ctrl-P
3. View that the Lightmode/User menu exists.
image

### Expected Behavior

The site should suppress these items as they are also in the same section as the search input “d-print-none”  https://github.com/netbox-community/netbox/blob/674af4d6bcfd724aef2c1d24bcc2725cb6b14c0e/netbox/templates/base/layout.html#L60

### Observed Behavior

It looks like the desktop version print view is somehow using/trigger the mobile view. I tested on both Chrome (131.0.6778.109 (Official Build) (64-bit)) and Firefox (128.5.1esr (64-bit)).  Adding “d-print-none” to mobile view fixes the issue.
image

jchambers2012:
### Fixes: #18183 
This will add the `d-print-none` to the mobile view so objects being printed out will suppress the Light/Dark Mode and Login Info. Looking at the template, it was intended to suppress these details, but Chrome and Firefox are triggering in the mobile view for printing. I tested on both Chrome (131.0.6778.109 (Official Build) (64-bit)) and Firefox (128.5.1esr (64-bit)).

### Pre-Change:
393964173-19d8d757-40dd-406a-9424-63ec6b464072

### Post Change: 
393964481-8cfff0cc-e9b0-4bf2-87d0-f3c11009df9c

### DIFF
--- a/netbox/templates/base/layout.html
+++ b/netbox/templates/base/layout.html
@@ -19,7 +19,7 @@ Blocks:
   <div class="page">
 
     {# Sidebar #}
-    <aside class="navbar navbar-vertical navbar-expand-lg">
+    <aside class="navbar navbar-vertical navbar-expand-lg d-print-none">
 
       {% if 'commercial' in settings.RELEASE.features %}
         <img class="motif" src="{% static 'motif.svg' %}" alt="{% trans "NetBox Motif" %}">
@@ -42,7 +42,7 @@ Blocks:
         </h1>
 
         {# User menu (mobile view) #}
-        <div class="navbar-nav flex-row align-items-center d-lg-none d-print-none">
+        <div class="navbar-nav flex-row align-items-center d-lg-none">
           {% plugin_navbar %}
           {% include 'inc/light_toggle.html' %}
           {% include 'inc/user_menu.html' %}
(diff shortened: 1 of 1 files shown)


jchambers2012:
@jeremystretch - would you prefer no logo or logo? Also I'm not a CSS wizard but I can add classes....

##  Adding to aside:
> Line 22
`<aside class="navbar navbar-vertical navbar-expand-lg d-print-none">`
image

## Adding to User menu (mobile view) div:
> Line 45

image

jeremystretch:
IMO it makes sense to omit the entire top bar when printing (`<aside class="navbar navbar-vertical navbar-expand-lg d-print-none">`).

jchambers2012:
> IMO it makes sense to omit the entire top bar when printing (`<aside class="navbar navbar-vertical navbar-expand-lg d-print-none">`).

Updated

jeremystretch (approved):
Thanks @jchambers2012!

PR #18186, opened 2024-12-10
Fixes: #18184 - Gracefully handle unavailable internet connection on RSS feed dashboard widget if ISOLATED_DEPLOYMENT is set

### LINKED ISSUE #18184: Disable RSS feed when `ISOLATED_DEPLOYMENT` is true
### NetBox version

v4.1.7

### Feature type

New functionality

### Triage priority

N/A

### Proposed functionality

The default dashboard shows a "NetBox News" feed, which tries to reach out to `netbox.dev` to retrieve its news items.

In an isolated network, this will always throw an error. It would be nice to exclude the news feed automatically if `ISOLATED_DEPLOYMENT` is set.

### Use case

It would avoid a user's potential first experience with netbox having a broken widget on the dashboard.

### Database changes

_No response_

### External dependencies

_No response_

bctiemann:
### Fixes: #18184

Because the RSS feed widget ("NetBox News") will return an error in an isolated deployment, this change skips adding that widget to the default dashboard if `ISOLATED_DEPLOYMENT` is set in the configuration.

Note that this matches all `extras.RSSFeedWidget` type widgets, so any additional RSS feeds we add to `DEFAULT_DASHBOARD` in the future will also be skipped in this situation.

### DIFF
--- a/netbox/extras/constants.py
+++ b/netbox/extras/constants.py
@@ -79,6 +79,7 @@ DEFAULT_DASHBOARD = [
             'feed_url': 'http://netbox.dev/rss/',
             'max_entries': 10,
             'cache_timeout': 14400,
+            'requires_internet': True,
         }
     },
     {
--- a/netbox/extras/dashboard/utils.py
+++ b/netbox/extras/dashboard/utils.py
@@ -61,9 +61,6 @@ def get_default_dashboard(config=None):
     config = config or settings.DEFAULT_DASHBOARD or DEFAULT_DASHBOARD
 
     for widget in config:
-        if settings.ISOLATED_DEPLOYMENT and widget['widget'] == 'extras.RSSFeedWidget':
-            continue
-
         id = str(uuid.uuid4())
         dashboard.layout.append({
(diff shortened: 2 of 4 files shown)


bctiemann:
Also note that this filtering behavior only occurs when the dashboard object is first populated from `DEFAULT_DASHBOARD`, which might occur before the user knows to set `ISOLATED_DEPLOYMENT` in the config. Maybe we should filter the RSS widget(s) out of the user's already-assigned Dashboard object? That feels invasive and presumptuous however.

jeremystretch:
This works, but I worry about precluding the use of RSS feed widgets entirely for an isolated deployment: It would be valid (though admittedly rare) to want to include an RSS feed from another locally-accessible system.

Maybe we could add a configuration parameter on `RSSFeedWidget` to indicate whether external access is required. If so, and `ISOLATED_DEPLOYMENT` is true, the widget will not be displayed. But maybe I'm overthinking it.

jeremystretch (approved):

PR #18199, opened 2024-12-10
Fixes: #18192 - Use assigned_object instead of interface in display_attrs

### LINKED ISSUE #18192: Searching for a MAC address results in server error
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.2-beta1

### Python Version

3.11

### Steps to Reproduce

1. Create a MAC address.
2. Search for it.

### Expected Behavior

Address shows up.

### Observed Behavior

```
<class 'AttributeError'>

'GenericRel' object has no attribute 'verbose_name'
```

bctiemann:
### Fixes: #18192

Uses `assigned_object` instead of `interface` in `MACAddressIndex.display_attrs`.

### DIFF
--- a/netbox/dcim/search.py
+++ b/netbox/dcim/search.py
@@ -105,7 +105,7 @@ class MACAddressIndex(SearchIndex):
         ('mac_address', 100),
         ('description', 500),
     )
-    display_attrs = ('mac_address', 'interface')
+    display_attrs = ('mac_address', 'assigned_object')
 
 
 @register_search
--- a/netbox/dcim/search.py
+++ b/netbox/dcim/search.py
@@ -105,7 +105,7 @@ class MACAddressIndex(SearchIndex):
         ('mac_address', 100),
         ('description', 500),
     )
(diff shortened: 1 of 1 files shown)


netbox/dcim/search.py
```diff
@@ -105,7 +105,7 @@ class MACAddressIndex(SearchIndex):
         ('mac_address', 100),
         ('description', 500),
     )
-    display_attrs = ('mac_address', 'interface')
+    display_attrs = ('mac_address', 'assigned_object')
```
jeremystretch, line 108: We can probably omit the MAC address itself, because it's already the primary representation of the object. Including it here is redundant (see below).

```suggestion
    display_attrs = ('assigned_object',)
```

screenshot

jeremystretch (changes requested):

netbox/dcim/search.py
bctiemann: Ah, good to know. Added `description` too.

jeremystretch (approved):

PR #18212, opened 2024-12-11
Closes #18211: Enable dynamic registration of request processors

### LINKED ISSUE #18211: Enable the dynamic registration of context managers for request processing (i.e. `event_tracking()`)
### NetBox version

v4.1.8

### Feature type

### Triage priority

N/A

### Proposed functionality

NetBox currently employs its [`event_tracking()`](https://github.com/netbox-community/netbox/blob/a15ff294dd95815027575ae65e0483c8ab29ffe6/netbox/netbox/context_managers.py#L8) context manager to process events triggered by a request, e.g. for change change logging. This function is invoked explicitly by the `CoreMiddleware` class.

Instead of calling this function directly, we can register it within NetBox's [application registry](https://netboxlabs.com/docs/netbox/en/stable/development/application-registry/), and consult the registry for all registered request processors whenever necessary, including within the middleware as well as for other purposes (such as running a background job).

### Use case

This approach will enable to us to dynamically register additional request processors, such as those which may be introduced by plugins, without the need for explicit configuration. #18136 raises one such use case.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Fixes: #18211

- Enable registration of request processors in the application registry
- Adapt CoreMiddleware to apply all registered request processors

Thanks to @arthanson for knocking out some of the preliminary work here in PR #18137.

### DIFF
--- a/docs/development/application-registry.md
+++ b/docs/development/application-registry.md
@@ -51,7 +51,7 @@ This store maintains all registered items for plugins, such as navigation menus,
 
 ### `request_processors`
 
-A list of context managers to invoke when processing a request e.g. in middleware or when executing a background job. Request processors can be registered with the `register_request_processor()` decorator.
+A list of context managers to invoke when processing a request e.g. in middleware or when executing a background job. Request processors can be registered with the `@register_request_processor` decorator.
 
 ### `search`
 
--- a/docs/development/application-registry.md
+++ b/docs/development/application-registry.md
@@ -49,6 +49,10 @@ This key lists all models which have been registered in NetBox which are not des
 
 This store maintains all registered items for plugins, such as navigation menus, template extensions, etc.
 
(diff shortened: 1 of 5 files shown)


docs/development/application-registry.md
```diff
@@ -49,6 +49,10 @@ This key lists all models which have been registered in NetBox which are not des
 
 This store maintains all registered items for plugins, such as navigation menus, template extensions, etc.
 
+### `request_processors`
+
+A list of context managers to invoke when processing a request e.g. in middleware or when executing a background job. Request processors can be registered with the `register_request_processor()` decorator.
```
bctiemann, line 54: Probably don't need the parentheses if the decorator should be applied as `@register_request_processor`.

bctiemann (approved):

PR #18226, opened 2024-12-12
Fixes: #16757 - Use table_htmx.html for assigning ipaddresses

### LINKED ISSUE #16757: htmx navigation - assign ip flow - odd rendering
### Deployment Type

On prem

### NetBox Version

v4.0.6

### Python Version

3.11

### Steps to Reproduce

1. enable htmx navigation on user
2. goto ipam > ip addresses, create an ip
3. create a device
4. create an interface on that device
5. on the interfaces list view, press the + button and choose to add an IP address
6. choose assign IP tab on top left
7. Fill in the previously created IP address and press search
8. click the ip in the list
9. note that now loaded edit ip address page has rendered weirdly and has two copies of the footer.

### Expected Behavior

edit ip address page should render normally

### Observed Behavior

odd rendering + dual footers (bar in the middle is from screenshot tool scrolling)

odd rendering

bctiemann:
### Fixes: #16757

Swaps `inc/table_htmx.html` for `inc/table.html` in the `ipaddress_assign.html` template. This ensures that objects clicked on in that table return a properly targeted HTMX response, whether or not HTMX navigation is turned on.

There are many places where `inc/table.html` is used, but it appears HTMX targeting is not used in those places. As HTMX is rolled out more pervasively, these will all need to be audited.

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -1971,7 +1971,6 @@ class DeviceInterfacesView(DeviceComponentsView):
     filterset = filtersets.InterfaceFilterSet
     filterset_form = forms.InterfaceFilterForm
     template_name = 'dcim/device/interfaces.html'
-    disable_htmx = True
     tab = ViewTab(
         label=_('Interfaces'),
         badge=lambda obj: obj.vc_interfaces().count(),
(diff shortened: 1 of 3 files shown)


jeremystretch (changes requested):
This returns the IP address edit form correctly, however after saving the IP the user is redirected to a bare view of the device's interfaces table.

screenshot

bctiemann:
@jeremystretch Turns out that behavior you're seeing of the subsequent table rendering is happening in develop, it's not related to my fix. I've added another adjustment which lets us `disable_htmx` certain views (currently just `DeviceInterfacesView` because it's special) so that won't happen.

bctiemann:
Reverted the last commit as the full-page bare-view issue is preexisting and beyond the scope of this specific issue's fix.

jeremystretch (approved):

PR #18235, opened 2024-12-13
Fixes #17215: set disabled interface backgroundcolor to $gray-400

### LINKED ISSUE #17215: Change color interface background (enabled/disabled/connected)
### NetBox version

v4.0.9

### Feature type

### Proposed functionality

This is a revisit of #16075 , as it was closed.

This is revisit of #15379. There @jeremystretch mentioned that the coloring would go away in v4.0, but it seems they are still there, hence my resubmit of this FR

Can we have a poll with the community to see what they prefer ?

Currently the background color of an interface is like this :
disabled: red
virtual: blue
enabled, marked_connected: green
enabled, connected (cable status color) : green/blue/..
enabled, no cable : white
```
def get_cabletermination_row_class(record):
    if record.mark_connected:
        return 'success'
    elif record.cable:
        return record.cable.get_status_color()
    return ''

def get_interface_row_class(record):
    if not record.enabled:
        return 'danger'
    elif record.is_virtual:
        return 'primary'
    return get_cabletermination_row_class(record)
```

My proposal is to change the color coding to:

disabled: white
virtual: blue
enabled, marked_connected: green
enabled, connected (cable status color) : green/blue/...
enabled, no cable : red

```
def get_cabletermination_row_class(record):
    if record.mark_connected:
        return 'success'
    elif record.cable:
        return record.cable.get_status_color()
    return 'danger'


def get_interface_row_class(record):
    if not record.enabled:
        return ''
    elif record.is_virtual:
        return 'primary'
    return get_cabletermination_row_class(record)
```

### Use case

The 'red' color is confusing, as it tells me something is 'wrong'. but having an interface in the disable state is not wrong.
image

With the proposed change, this view would become more logic (red = wrong state )
image

Having an 'enabled' interface, without a cable is 'danger'
A disabled interface is just good, that can be white.

### Database changes

_No response_

### External dependencies

_No response_

PieterL75:


jeremystretch (approved):
Thanks @PieterL75!

PR #18237, opened 2024-12-13
Fix definition of `vid_ranges` in `VLANGroup` so it shows up in the OpenAPI schema

### LINKED ISSUE #17488: New field vid_ranges on vlan-group missing in API
### Deployment Type

Self-hosted

### NetBox Version

v4.1.1

### Python Version

3.10

### Steps to Reproduce

Open "REST API Documentation" 
Scroll down to Schemas
Open VLANGroup

### Expected Behavior

The field vid_ranges should show up

### Observed Behavior

The field vid_ranges is missing

Skjermbilde 2024-09-13 095137

mraerino:
### Fixes: https://github.com/netbox-community/netbox/issues/17488

The OpenAPI schema for `vid_ranges` cannot automatically be discovered by drf-spectacular since it's a custom serializer with no fields. Therefore we need to manually provide the OpenAPI schema for the type we expect to see in requests and responses.

On top of that there is a bug in drf-spectacular that required the workaround with the custom list class (https://github.com/tfranzel/drf-spectacular/issues/1353).

### DIFF
--- a/netbox/core/api/schema.py
+++ b/netbox/core/api/schema.py
@@ -2,12 +2,13 @@ import re
 import typing
 from collections import OrderedDict
 
-from drf_spectacular.extensions import OpenApiSerializerFieldExtension
+from drf_spectacular.extensions import OpenApiSerializerFieldExtension, OpenApiSerializerExtension, _SchemaType
 from drf_spectacular.openapi import AutoSchema
 from drf_spectacular.plumbing import (
     build_basic_type, build_choice_field, build_media_type_object, build_object_type, get_doc,
 )
 from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import Direction
 
 from netbox.api.fields import ChoiceField
 from netbox.api.serializers import WritableNestedSerializer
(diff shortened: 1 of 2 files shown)


jeremystretch:
@mraerino did you see the comment [here](https://github.com/tfranzel/drf-spectacular/issues/1353#issuecomment-2550682793) regarding the use of `extend_schema_field()` on a serializer (versus a field)? I suspect a more prudent solution might be to rewrite IntegerRangeSerializer to use IntegerField, though I'm not sure of the specifics.

mraerino:
yeah i saw. i think this is likely a little bit more complex and i wanted to get this bug fix out quickly so people can use the API again.

I can look into improving this in another PR, if you're ok with that?

mraerino:
@jeremystretch i changed the implementation to use a schema extension instead of an annotation, as recommended by the spectacular devs.

I don't think using `IntegerField` will work since we have an array of integers in an array and 2 dimensions cannot be modelled without a custom serializer

arthanson:
@mraerino Can you please rebase this to main

arthanson (approved):

PR #18244, opened 2024-12-17
Fixes: #17868 - Handle orphaned cable condition gracefully in SVG rendering

### LINKED ISSUE #17868: Cable Trace leads to "division by zero" if the last cable is not connected
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

NetBox Community v4.1.3

### Python Version

3.11

### Steps to Reproduce

Have a switch (or similar) with an interface connected through multiple passive components (ie patchpanel, wall-plugs) to a device with an interface.
From switch, do cable trace on this interface - it will display all in-between interfaces (as expected).
Delete the device without deleting or disconnecting its cable will result in an unterminated cable (might be by [design](https://github.com/netbox-community/netbox/issues/10769)).
Trace the cable again from the switch interface. 

### Expected Behavior

I expect the cable trace to walk through until the last (unterminated) cable and show either the last device or the unterminated cable to trace down the still existing cables.

### Observed Behavior

Results in an error without displaying any svg cable date:
```
<class 'ZeroDivisionError'>

division by zero

Python-Version: 3.11.2
NetBox-Version: 4.1.3
Plugins: 
  netbox_topology_views: 4.0.1
```

bctiemann:
### Fixes: #17868

If a device has been deleted, attached cables may be left orphaned, which leads to rendering errors when generating the cable trace SVG. This fix handles the case where the object list passed in to `draw_far_objects` is empty by returning empty lists of results and then handling those gracefully so that the remaining portion of the SVG can be rendered.

### DIFF
--- a/netbox/dcim/svg/cables.py
+++ b/netbox/dcim/svg/cables.py
@@ -268,6 +268,10 @@ class CableTraceSVG:
         """
         Draw the far-end objects and its terminations and return all created nodes
         """
+        # If an empty list is passed in, return empty results so this cable can be skipped
+        if len(obj_list) == 0:
+            return [], []
+
         # Make sure elements are sorted by name for readability
         objects = sorted(obj_list, key=lambda x: str(x))
         width = self.width / len(objects)
@@ -404,6 +408,10 @@ class CableTraceSVG:
                             # a and b terminations may be swapped
                             near = [term for term in near_terminations if term.object == cable.interface_b]
                             far = [term for term in far_terminations if term.object == cable.interface_a]
+
+                    # If at this point we have no near or no far terminations, this cable can't be rendered, so skip.
+                    if not (near and far):
+                        continue
 
                     # Select most-probable start and end position
                     start = near[0].bottom_center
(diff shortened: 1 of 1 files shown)


netbox/dcim/svg/cables.py
```diff
@@ -268,6 +268,10 @@ def draw_far_objects(self, obj_list, terminations):
         """
         Draw the far-end objects and its terminations and return all created nodes
         """
+        # If an empty list is passed in, return empty results so this cable can be skipped
+        if len(obj_list) == 0:
```
jeremystretch, line 272: Minor nit
```suggestion
        if not len(obj_list):
```

netbox/dcim/svg/cables.py
```diff
@@ -366,6 +370,11 @@ def render(self):
 
                 obj_list = {end.parent_object for end in far_ends}
                 parent_object_nodes, far_terminations = self.draw_far_objects(obj_list, far_ends)
+
+                # If there are no far terminations, this segment can't be rendered, so skip.
+                if not far_terminations:
+                    continue
```
jeremystretch, line 376: Can this be obviated by checking for `far_ends` in addition to `links` above?

```
if links and far_ends:
```

jeremystretch (changes requested):

netbox/dcim/svg/cables.py
bctiemann: I think you're right, that does work fine and is a lot more surgical.

jeremystretch (approved):

PR #18264, opened 2024-12-21
Fixes: #18263 - Iterate through a freshly queried set of CableTerminations to find endpoints in update_connected_endpoints

### LINKED ISSUE #18263: Cable terminations are not refreshed correctly on API PATCH of Cable
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.8

### Python Version

3.10

### Steps to Reproduce

Ideally populate the DB with a large number (1000+) of Interfaces and Cables connecting them.

API call:
```
http://127.0.0.1:8000/api/dcim/cables/1/
{
    "b_terminations": [
        {
            "object_id": 161
        }
    ]
}
```

Compare to UI edit of Cable where B-side interface is changed.

### Expected Behavior

Both operations (UI and API) should have the same effect and execute in the same amount of time.

### Observed Behavior

The UI call completes in a few hundred ms; but depending on the data size, the API call can take many seconds or minutes to complete.

bctiemann:
### Fixes: #18263

During an API PATCH update to a Cable where one or both endpoints is changed, the signal handler `update_connected_endpoints` iterates through `self.terminations.all()` which in the case of an API call has not necessarily been refreshed after the deletion of stale endpoints at https://github.com/netbox-community/netbox/blob/b6265d828526a97962e8feefa0ff57c534d2da27/netbox/dcim/models/cables.py#L224-L240

This leads to the `local_cable_terminations` being empty here (because the pks don't match): https://github.com/netbox-community/netbox/blob/b6265d828526a97962e8feefa0ff57c534d2da27/netbox/dcim/models/cables.py#L597-L611 (Step 6 of `Cable.from_origin`), which leads to `q_filter` being empty and thus the completely unfiltered iteration at L611 potentially taking O(n) time depending on the number of terminations in the DB.

This change ensures that the iteration of CableTerminations associated with a Cable after an update is freshly queried from the DB following the deletion of stale endpoints rather than relying on the (potentially stale or cached) `self.terminations` reverse relation, and thus ensuring `from_origin` is calculated correctly.

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -607,6 +607,10 @@ class CablePath(models.Model):
                     cable_end = 'A' if lct.cable_end == 'B' else 'B'
                     q_filter |= Q(cable=lct.cable, cable_end=cable_end)
 
+                # Make sure this filter has been populated; if not, we have probably been given invalid data
+                if not q_filter:
+                    break
+
                 remote_cable_terminations = CableTermination.objects.filter(q_filter)
                 remote_terminations = [ct.termination for ct in remote_cable_terminations]
             else:
--- a/netbox/dcim/signals.py
+++ b/netbox/dcim/signals.py
@@ -85,7 +85,8 @@ def update_connected_endpoints(instance, created, raw=False, **kwargs):
     if instance._terminations_modified:
         a_terminations = []
         b_terminations = []
-        for t in instance.terminations.all():
+        # Note: instance.terminations.all() is not safe to use here as it might be stale
+        for t in CableTermination.objects.filter(cable=instance):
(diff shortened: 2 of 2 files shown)


jeremystretch:
> (Step 6 of Cable.from_origin), which leads to q_filter being empty and thus the completely unfiltered iteration at L611 potentially taking O(n) time depending on the number of terminations in the DB.

This is a really good catch. IMO we should rewrite the logic under step 6 to avoid querying with an empty Q filter (in addition to the proposed change) to be safe. @bctiemann you're more familiar with the problem than I am; does that make sense?

jeremystretch (approved):

PR #18272, opened 2024-12-27
Fixes #18271: Require only encryption OR authentication algorithm when creating an IPSec proposal via REST API

### LINKED ISSUE #18271: IPSec-Proposal optional field is required when sending HTTP Post request via API
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.10

### Python Version

3.10

### Steps to Reproduce

1. Trying to send HTTP POST request to API endpoint:
/api/vpn/ipsec-proposals/
```
{
'name': 'aes-192-gcm_3600',
'encryption_algorithm': 'aes-192-gcm',
'sa_lifetime_seconds': '3600'
}
```

### Expected Behavior

Got HTTP status code 201, which means that the object was successfully created.

### Observed Behavior

Got the error:

```
  allow: GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS
  api_version: '4.1'
  connection: close
  content: '{"authentication_algorithm":["This field is required."]}'
  json:
    authentication_algorithm:
    - This field is required.
```

jeremystretch:
### Fixes: #18271

- Set `required=False` on `encryption_algorithm` & `authentication_algorithm` for IPSecProposalSerializer

### DIFF
--- a/netbox/vpn/api/serializers_/crypto.py
+++ b/netbox/vpn/api/serializers_/crypto.py
@@ -64,10 +64,12 @@ class IKEPolicySerializer(NetBoxModelSerializer):
 
 class IPSecProposalSerializer(NetBoxModelSerializer):
     encryption_algorithm = ChoiceField(
-        choices=EncryptionAlgorithmChoices
+        choices=EncryptionAlgorithmChoices,
+        required=False
     )
     authentication_algorithm = ChoiceField(
-        choices=AuthenticationAlgorithmChoices
+        choices=AuthenticationAlgorithmChoices,
+        required=False
     )
 
     class Meta:


netbox/vpn/api/serializers_/crypto.py
```diff
@@ -64,10 +64,12 @@ class Meta:
 
 class IPSecProposalSerializer(NetBoxModelSerializer):
     encryption_algorithm = ChoiceField(
-        choices=EncryptionAlgorithmChoices
```
bctiemann, line 67: This makes both fields optional, but shouldn't there also be logic added to ensure that at least one is specified? Or is "neither" a valid payload?

jeremystretch: Validation is enforced by the model's `clean()` method. This change just updates the serializer to match the model.

bctiemann (approved):

PR #18274, opened 2024-12-27
Fix duplicate on VLAN list when filtering by interface

### LINKED ISSUE #17771: Strange behaviour with generated >3 VLANs table for interfaces
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.4

### Python Version

3.11

### Steps to Reproduce

I'm seeing some weird behaviour with the new feature as implemented per #17655. I can't give exact guaranteed reproduction steps, because it might be something specific to my database that I haven't figured out yet, but basically:

1. Find an interface with more than 3 tagged VLANs and 1 untagged VLAN. (created before the update(?), see Observed Behavior)
2. Click on the created link for the tagged VLANs (of the form `/ipam/vlans/?interface_id=<ID>`).

### Expected Behavior

Tagged VLANs for that interface are shown correctly.

### Observed Behavior

What I am seeing is some of the tagged VLANs being listed multiple times (and the _untagged_ VLAN showing up as well, which I guess is to be expected). I've tried to reproduce it with new VLANs created _after_ the update to 4.1.4 and it seems to work as expected with those (some "older" VLANs are also only showing up once, however, so it might not be related to that at all), but when I attach a certain existing VLAN as tagged to the interface it results in multiple redundant rows again.

image

thordreier:
### Fixes: #17771 

Use distinct() to filter out duplicate VLANs when filtering by interface.


jeremystretch (approved):
Thanks @thordreier!

PR #18291, opened 2025-01-02
Closes #18281: Support group assignment for virtual circuits

### LINKED ISSUE #18281: Allow Virtual Circuits to be assigned to groups
### NetBox version

v4.2-beta1

### Feature type

Data model extension

### Triage priority

N/A

### Proposed functionality

Allow for Virtual Circuits to also be assigned to groups. I believe it's sufficient to use the existing Circuit Groups and only change the Circuit Group Assignment model to allow for selecting a Circuit or Virtual Circuit, as a given group might have a mix.

### Use case

We have public cloud connectivity through both a direct connection and a backup VPN connection, so it would be nice to document the (physical) Circuit for the direct connection with a primary priority and a Virtual Circuit for the VPN with a secondary priority within the same Circuit Group. In other places, we have multiple (usually two) Virtual Circuits between locations on different VLANs and diverse paths, which we designate with different priorities. Just like we can document (physical) Circuits in this way, we should be able to do the same with Virtual Circuits.

### Database changes

Either add a `virtual_circuit` field that is a `ForeignKey` to the `VirtualCircuit` model or change the existing `circuit` field to be a `GenericForeignKey` on the `CircuitGroupAssignment` model.

### External dependencies

None

jeremystretch:
### Closes: #18281

- Convert the `circuit` ForeignKey on CircuitGroupAssignment to a generic foreign key (`member`)
- Adapt Circuit & VirtualCircuit models for GFK-based group assignment
- Rearrange navigation menu to list circuit groups separately


netbox/circuits/filtersets.py
```diff
@@ -365,26 +368,57 @@ class CircuitGroupAssignmentFilterSet(NetBoxModelFilterSet):
...
-        field_name='circuit__provider',
-        queryset=Provider.objects.all(),
-        label=_('Provider (ID)'),
-    )
-    provider = django_filters.ModelMultipleChoiceFilter(
-        field_name='circuit__provider__slug',
-        queryset=Provider.objects.all(),
-        to_field_name='slug',
-        label=_('Provider (slug)'),
+    # provider_id = django_filters.ModelMultipleChoiceFilter(
```
bctiemann, line 371: Remove these commented blocks?

bctiemann (approved):

PR #18315, opened 2025-01-07
Fix #18314: Use get to avoid KeyError

### LINKED ISSUE #18314: Server Error then login with superuser Account
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.2.0

### Python Version

3.10

### Steps to Reproduce

1- Update to v4.2.0
2. Try to login with a superuser account

### Expected Behavior

Should be able to login with a superuser account

### Observed Behavior

Login with a "normal" useraccount is possible, with a superuser account we get the following error message

Screenshot 2025-01-07 093138

tobiasge:


jeremystretch (approved):
Thanks @tobiasge!

PR #18322, opened 2025-01-07
Fixes: #18316 - Fix PrefixIndex reference to 'site'

### LINKED ISSUE #18316: Server error in search
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.2.0

### Python Version

3.10

### Steps to Reproduce

1. Add an IP to IPAM
2. Seach for this IP

Example: https://demo.netbox.dev/search/?q=172.16.0.7

### Expected Behavior

IP is found

### Observed Behavior

Server error:
```
  File "/opt/netbox/current/netbox/extras/models/search.py", line 73, in display_attrs
    name = self.object._meta.get_field(attr).verbose_name
AttributeError: 'GenericRel' object has no attribute 'verbose_name'
```

bctiemann:
### Fixes: #18316

`PrefixIndex` was incorrectly referring to `site` as an indexable field following the scoping refactor; this should be `scope` now.

### DIFF
--- a/netbox/ipam/search.py
+++ b/netbox/ipam/search.py
@@ -79,7 +79,7 @@ class PrefixIndex(SearchIndex):
         ('description', 500),
         ('comments', 5000),
     )
-    display_attrs = ('site', 'vrf', 'tenant', 'vlan', 'status', 'role', 'description')
+    display_attrs = ('scope', 'vrf', 'tenant', 'vlan', 'status', 'role', 'description')
 
 
 @register_search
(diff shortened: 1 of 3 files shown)


jeremystretch (changes requested):
Could you fix this for ClusterIndex as well please? It might also make sense to add `scope` to the search index for wireless LANs (this field was added in v4.2).

jeremystretch (approved):

PR #18351, opened 2025-01-08
Fixes: #18350 - Remove 'site' and 'provider_network' from CircuitTerminationIndex.display_attrs

### LINKED ISSUE #18350: Global search which matches CircuitTermination results in "CircuitTermination has no field named 'site'"
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.2.1

### Python Version

3.10

### Steps to Reproduce

Create at least one CircuitTermination where the `xconnect_id` field is populated
Global search for the value of that field

### Expected Behavior

Successful search result page with the CircuitTermination in question listed

### Observed Behavior

`FieldDoesNotExist`: `CircuitTermination has no field named 'site'`

bctiemann:
### Fixes: #18350

As `CircuitTermination` has been refactored to move the fields `site` and `provider_network` to the new scoping model, this change uses the new internal (cached) versions of those fields in the displayed attributes in search results, resolving a crash condition.

### DIFF
--- a/netbox/circuits/search.py
+++ b/netbox/circuits/search.py
@@ -34,7 +34,7 @@ class CircuitTerminationIndex(SearchIndex):
         ('port_speed', 2000),
         ('upstream_speed', 2000),
     )
-    display_attrs = ('circuit', 'site', 'provider_network', 'description')
+    display_attrs = ('circuit', 'description')
 
 
 @register_search
--- a/netbox/circuits/search.py
+++ b/netbox/circuits/search.py
@@ -34,7 +34,7 @@ class CircuitTerminationIndex(SearchIndex):
         ('port_speed', 2000),
         ('upstream_speed', 2000),
     )
-    display_attrs = ('circuit', 'description')
+    display_attrs = ('circuit', '_site', '_provider_network', 'description')
 
 
 @register_search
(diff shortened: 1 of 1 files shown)


tyler-8:
IMO this should also display the scoping data that replaces the site and provider_network fields, not just stop the crash.

netbox/circuits/search.py
```diff
@@ -34,7 +34,7 @@ class CircuitTerminationIndex(SearchIndex):
         ('port_speed', 2000),
         ('upstream_speed', 2000),
     )
-    display_attrs = ('circuit', 'site', 'provider_network', 'description')
+    display_attrs = ('circuit', '_site', '_provider_network', 'description')
```
jeremystretch, line 37: We need to ensure that all scope types are supported (not just site and provider network). To do this, we can reference the `termination` GFK rather than the individual ForeignKey assignments.

```suggestion
    display_attrs = ('circuit', 'termination', 'description')
```

This will return whichever type of related object is assigned:

screenshot

jeremystretch (changes requested):

netbox/circuits/search.py
bctiemann: Agreed, that's much nicer.

jeremystretch (approved):

PR #18375, opened 2025-01-10
Fixes: #18373 - Fix validation of site in Assign Device to Cluster flow

### LINKED ISSUE #18373: Cannot Assign devices to Cluster
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v.4.2.1

### Python Version

3.11

### Steps to Reproduce

1. Create a new Cluster, fill out only required fields
2. Click on Assign Device to Cluster, and fill out only required fields
3. Click Add Devices

### Expected Behavior

Device(s) are assigned to Cluster

### Observed Behavior

Server Error from NetBox
```
<class 'AttributeError'>

'Cluster' object has no attribute 'site'

Python version: 3.11.2
NetBox version: 4.2.1
Plugins: 
  netbox_branching: 0.5.2
  netbox_device_view: 0.1.7
  netbox_reorder_rack: 1.1.3
```

bctiemann:
### Fixes: #18373

Assigning a Device to a Cluster goes through a validation to ensure the Site of both objects is the same. Because Cluster is now generically scoped, this causes an error as `Cluster.site` no longer exists.

This change fixes the validation to compare Site objects properly, and also conforms to a pattern that can be extended to validate Location, Site Group, etc. scopes if they are added to the Assign Device form in the future.

### DIFF
--- a/netbox/virtualization/forms/model_forms.py
+++ b/netbox/virtualization/forms/model_forms.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.apps import apps
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
I think we need to validate every potential type of cluster scope (region, site group, site, or location). Ideally, this should happen under the model validation, but that might be difficult to manage given the deviation from standard patterns in this particular view.

bctiemann:
As it turns out Device doesn't have `site_group` or `region` fields, only `site` and `location`. So I extended this validation to cover both those fields, and it means it's further a pattern outlier that isn't likely to be applicable to many other objects that aren't fully scoped.

Also note that this validation will throw an error of one of the `Device` or `Cluster` has a value specified but the other is null. Maybe we don't want that to be an error condition?

jeremystretch (approved):

PR #18388, opened 2025-01-13
Fixes: #18369 - Remove the json filter for protection rules

### LINKED ISSUE #18369: JSON serializer issue for Protection Rules defined in configuration.py
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.11

### Python Version

3.12

### Steps to Reproduce

1. Create a custom `Protection_Rule` in the configuration.py. Example:
```

# Prevent IP addresses from being deleted by a cascading delete
class IpAddressProtectionRule(CustomValidator):
    def validate(self, instance, request):
        if hasattr(instance, 'primary_ip') and instance.primary_ip is not None:
            self.fail('This instance has a related IP address(es).')


PROTECTION_RULES = {
    'dcim.device': (
        IpAddressProtectionRule(),
    ),
}
```
2. Restart NetBox services
3. Navigate to Admin -> System

### Expected Behavior

The system detail page is returned.

### Observed Behavior

An exception is thrown:
image

atownson:
Fixes: #18369

Removing the json filter resolves the TypeError exception and aligns the protection rules template with the custom validators template.

### DIFF
--- a/netbox/core/views.py
+++ b/netbox/core/views.py
@@ -570,8 +570,9 @@ class SystemView(UserPassesTestMixin, View):
             return response
 
         # Serialize any CustomValidator classes
-        if hasattr(config, 'CUSTOM_VALIDATORS') and config.CUSTOM_VALIDATORS:
-            config.CUSTOM_VALIDATORS = json.dumps(config.CUSTOM_VALIDATORS, cls=ConfigJSONEncoder, indent=4)
+        for attr in ['CUSTOM_VALIDATORS', 'PROTECTION_RULES']:
+            if hasattr(config, attr) and getattr(config, attr):
+                setattr(config, attr, json.dumps(getattr(config, attr), cls=ConfigJSONEncoder, indent=4))
 
         return render(request, 'core/system.html', {
             'stats': stats,
--- a/netbox/core/views.py
+++ b/netbox/core/views.py
@@ -571,7 +571,7 @@ class SystemView(UserPassesTestMixin, View):
 
         # Serialize any CustomValidator classes
         for attr in ['CUSTOM_VALIDATORS', 'PROTECTION_RULES']:
-            if hasattr(config, attr) and getattr(config, attr):
+            if hasattr(config, attr) and getattr(config, attr, None):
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):

netbox/templates/core/inc/config_data.html
```diff
@@ -103,7 +103,7 @@
   <tr>
     <th scope="row" class="border-0 ps-3">{% trans "Protection rules" %}</th>
     {% if config.PROTECTION_RULES %}
-      <td class="border-0"><pre>{{ config.PROTECTION_RULES|json }}</pre></td>
+      <td class="border-0"><pre>{{ config.PROTECTION_RULES }}</pre></td>
```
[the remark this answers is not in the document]
atownson: I have added a commit for the PROTECTION_RULES config attribute to use the ConfigJSONEncoder JSON serializer. This should fix the formatting issue and align with the existing CUSTOM_VALIDATORS attribute serializer.

netbox/core/views.py
```diff
@@ -570,8 +570,9 @@ def get(self, request):
             return response
 
         # Serialize any CustomValidator classes
-        if hasattr(config, 'CUSTOM_VALIDATORS') and config.CUSTOM_VALIDATORS:
-            config.CUSTOM_VALIDATORS = json.dumps(config.CUSTOM_VALIDATORS, cls=ConfigJSONEncoder, indent=4)
+        for attr in ['CUSTOM_VALIDATORS', 'PROTECTION_RULES']:
+            if hasattr(config, attr) and getattr(config, attr):
```
jeremystretch, line 574: Just have to handle the default condition here.
```suggestion
            if hasattr(config, attr) and getattr(config, attr, None):
```

jeremystretch (changes requested):

netbox/core/views.py
jeremystretch: I'll knock this out real quick if you don't mind, so we can slip it into the v4.2.2 release today.

jeremystretch (approved):
Thanks @atownson!

PR #18444, opened 2025-01-21
Fixes: #18433 - Fix missing is_primary property on MACAddress model

### LINKED ISSUE #18433: MAC address not shown as "primary for interface" in MAC address detail view
### Deployment Type

Self-hosted

### NetBox Version

v4.2.2

### Python Version

3.12

### Steps to Reproduce

1. Create a device (device1) (any device type)
2. Create an interface (eth1) (any type)
3. Create a MAC address (AA:AB:CC:DD:EE:FF) and assign it to the device1 and interface eth1
4. Go to interface eth1 on device1 and edit the device
5. Assign the MAC address AA:AB:CC:DD:EE:FF as the primary MAC address
6. Go to the MAC address view and select the AA:AB:CC:DD:EE:FF address

### Expected Behavior

In the detail view I expect the "Primary for interface" to be a green checkmark. 

### Observed Behavior

It's red to indicate it is not primary for any interface.

DanSheps:
### Fixes: #18433 - Add missing is_primary property on MACAddress model

* Added missing `is_primary` property method on `MACAddress` model


bctiemann:
I had `is_primary` on the model as a concrete field initially, then changed it to `primary_for_interface` field on the related model during development/review. I guess the template still had the vestigial mention. I'm not sure why we didn't put it back in as a property, but I seem to remember wanting to but there was some argument against it?

Approved for the moment, but there might be some discussion to be had.

bctiemann (dismissed):

DanSheps:
> I had `is_primary` on the model as a concrete field initially, then changed it to `primary_for_interface` field on the related model during development/review. I guess the template still had the vestigial mention. I'm not sure why we didn't put it back in as a property, but I seem to remember wanting to but there was some argument against it?

I think `is_primary` as a "Boolean" makes sense.  If it was `is_primary` as the reverse relation it would be kind of weird IMO.

netbox/dcim/models/devices.py
```diff
@@ -1522,3 +1522,10 @@ class Meta:
 
     def __str__(self):
         return str(self.mac_address)
+
+    @property
```
jeremystretch, line 1526: Any reason not to use `@cached_property` here?

DanSheps: I don't see why not.  Lemme make that change.

bctiemann (dismissed):

bctiemann:
@DanSheps please fix merge conflict

bctiemann (approved):

PR #18450, opened 2025-01-21
Fixes: #18447 - Fix sorting by `mac_address` field

### LINKED ISSUE #18447: Sort by MAC address crashes "interfaces" tab view...
### Deployment Type

Self-hosted

### NetBox Version

v4.2.2

### Python Version

3.11

### Steps to Reproduce

1. Once logged in, clic "virtualization",
2. clic "Virtual machines"
3. clic a VM name which owns more than one interface (in order to be able to sort)
4. clic "interfaces"
5. clic to sort by "mac address".
It's dead, you can not come back to ani "interfaces" tab anymore 👍 

### Expected Behavior

View "interfaces" tab like before MAC ADDRESS became an object.

### Observed Behavior

Server Error

Cannot resolve keyword 'mac_address' into field. Choices are: _name, bookmarks, bridge, bridge_id, bridge_interfaces, child_interfaces, created, custom_field_data, description, enabled, fhrp_group_assignments, id, ip_addresses, journal_entries, l2vpn_terminations, last_updated, mac_addresses, mode, mtu, name, parent, parent_id, primary_mac_address, primary_mac_address_id, qinq_svlan, qinq_svlan_id, subscriptions, tagged_items, tagged_vlans, tags, tunnel_terminations, untagged_vlan, untagged_vlan_id, virtual_machine, virtual_machine_id, vlan_translation_policy, vlan_translation_policy_id, vrf, vrf_id

Python version: 3.11.2
NetBox version: 4.2.2
Plugins: None installed

[Home Page](http://netbox.sud.minint.fr/)

### LINKED ISSUE #18456: Duplicate MAC address attribute
### Deployment Type

Self-hosted

### NetBox Version

v4.2.2

### Python Version

3.11

### Steps to Reproduce

1. Go to a device or VM interface table.
2. Check the available columns.

### Expected Behavior

One column: "MAC Address"

### Observed Behavior

Two columns: "MAC Address", "Mac address"

Image
Image

`dir(Interface)` in the Python shell also shows me ` ...'mac_address', 'mac_addresses',...`

So I guess the lowercase one is probably a remnant from a plugin or something? These are the plugins I have currently installed:
- NetBox Floorplan
- Reordering Racked Devices
- NetBox Interface Synchronization
- NetBox Topology Views
- NetBox Inventory
- Document Storage
- IP Calculator

As far as I can determine, none of them are the culprit, so something that _used to be_ installed maybe? In any case, how to get rid of it?

DanSheps:
### Fixes: #18447 - Fix sorting by `mac_address` field

* Disable sorting by `mac_address` for legacy `mac_address` field for Device and VM Interfaces
* Ensure `primary_mac_address` field is included in field list for Device and VM Interfaces

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -697,6 +697,10 @@ class DeviceInterfaceTable(InterfaceTable):
         linkify=True,
         verbose_name=_('LAG')
     )
+    mac_address = tables.Column(
+        verbose_name=_('MAC Address'),
+        orderable=False,
+    )
     actions = columns.ActionsColumn(
         extra_buttons=INTERFACE_BUTTONS
     )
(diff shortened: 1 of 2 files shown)


bctiemann:
@DanSheps Actually I noticed that `InterfaceTable` already has `primary_mac_address` as a fully functional and sortable field (with properly capitalized label); I think we just need to add that same field to `VMInterfaceTable`, and then `DeviceInterfacesTable` and `VirtualMachineVMInterfaceTable` need similar fields with appropriate accessors so sorting works.

I'm not sure if your solution catches all four of these cases, or if the way it is on `InterfaceTable` is the most correct, but I feel like we can preserve the functionality across the four.

DanSheps:
> @DanSheps Actually I noticed that `InterfaceTable` already has `primary_mac_address` as a fully functional and sortable field (with properly capitalized label); I think we just need to add that same field to `VMInterfaceTable`, and then `DeviceInterfacesTable` and `VirtualMachineVMInterfaceTable` need similar fields with appropriate accessors so sorting works.

That is what I did (more or less).  I didn't want to remove `mac_address` from those as they are not deprecated (yet) AFAIK, so preserving them and disabling sorting seems to be the better way to go.

If we add the accessor, it unfortunately overrides the `primary_mac_address` field (at least from what I can tell) with the verbose name set for `mac_address`.

> I'm not sure if your solution catches all four of these cases, or if the way it is on `InterfaceTable` is the most correct, but I feel like we can preserve the functionality across the four.

IMO `mac_address` should be removed in the future and we stick with `primary_mac_address`

bctiemann:
OK -- so should we disable sorting on that field in `InterfaceTable` too then? Right now that one is sortable and the other three are not.

bctiemann:
Invalid field names that are persisted in user prefs are filtered out in django-tables before being applied, I found in working on an earlier issue. I think it's safe to change field names (though it's reasonably easy to test too, just sort on that column and then rename it, see if anything breaks).

bctiemann (approved):

PR #18451, opened 2025-01-21
Allow primary key for nested models in OpenAPI request schemas

### LINKED ISSUE #17709: OpenAPI Specification Does Not Allow Numeric IDs for Related Objects, Contrary to Documentation
### Deployment Type

Self-hosted

### NetBox Version

v4.1.3

### Python Version

3.12

### Steps to Reproduce

1. Access the NetBox REST API documentation generated from the OpenAPI specification (e.g., `/api/schema/swagger-ui/`).

2. Navigate to the API endpoint for creating a new device (`POST /api/dcim/devices/`).

3. Review the schema for the request body in the OpenAPI specification.

4. Observe that foreign key fields, such as `device_role`, `device_type`, `site`, etc., are defined as requiring nested objects with specific attributes.

5. Attempt to create a new device by specifying related objects using numeric IDs in the request body:

```json
{
    "name": "MyNewDevice",
    "device_type": 1,
    "role": 2,
    "site": 3,
    "status": "active"
}
```

6. Note that according to the OpenAPI schema, this request does not conform to the expected input format.

### Expected Behavior

According to the NetBox documentation under the "Related Objects" section:

1. When performing write API actions (POST, PUT, and PATCH), related objects may be specified by either numeric ID (primary key) or by a set of attributes sufficiently unique to return the desired object.

2. For example, creating a new device by specifying its rack using a numeric ID should be acceptable:

```json
{
    "device_type": 1,
    "role": 2,
    "site": 3,
}
```

The OpenAPI specification should reflect this flexibility, allowing foreign key fields to accept either a numeric ID or a nested object with unique attributes.

Client code generated from the OpenAPI schema should support specifying related objects using numeric IDs, resulting in concise and manageable code, especially in statically typed languages.

### Observed Behavior

- The OpenAPI specification only allows for specifying related objects as nested objects with specific attributes and does not provide the option to use numeric IDs.

- Client code generated from the OpenAPI schema enforces the use of verbose nested objects for related fields, leading to cumbersome and less maintainable code.

- This discrepancy between the documentation and the OpenAPI specification creates confusion and hinders the development of client applications, particularly in statically typed languages like Go.

- For example, in Go, creating a device requires constructing nested objects for each related field:

```go
client.Api.DcimAPI.DcimDevicesCreate(context.Background()).WritableDeviceWithConfigContextRequest(
	netbox.WritableDeviceWithConfigContextRequest{
		DeviceType: *netbox.NewBriefDeviceTypeRequest(
			*netbox.NewBriefManufacturerRequest(typeMikrotik.Manufacturer.Name, typeMikrotik.Manufacturer.Slug),
			typeMikrotik.Model,
			typeMikrotik.Slug,
		),
		Role: *netbox.NewBriefDeviceRoleRequest(roleRouter.Name, roleRouter.Slug),
		Site: *netbox.NewBriefSiteRequest(site.Name, site.Slug),
	}).Execute()
```

- If the OpenAPI specification allowed for numeric IDs, the code would be much simpler:

```go
client.DcimApi.DcimDevicesCreate(context.Background()).WritableDeviceWithConfigContextRequest(
    netbox.WritableDeviceWithConfigContextRequest{
        DeviceType: 1,
        Role:       2,
        Site:       3,
    },
).Execute()
```

mraerino:
### Fixes: https://github.com/netbox-community/netbox/issues/17709

The logic for passing a primary key when specifying a nested model on creation or update is hidden in some special logic in `BaseModelSerializer` and cannot be detected by `drf-spectacular`.

I'm introducing an extension that patches the write request schemas to allow for either the primary key as a number or an object with fields (what was already present).

This aligns the schema with reality.


mraerino:
@jeremystretch any change i could get a review here?

hikhvar:
Any updates on this? I generated a Go client with these changes and it really looks very good.

netbox/core/api/schema.py
```diff
@@ -277,3 +277,23 @@ def map_serializer_field(self, auto_schema, direction):
...
+
+
+# Nested models can be passed by ID in requests
+# The logic for this is handled in `BaseModelSerializer.to_internal_value`
+class FixWritableNestedSerializerAllowPK(OpenApiSerializerFieldExtension):
+    target_class = 'netbox.api.serializers.BaseModelSerializer'
+    match_subclasses = True
+
+    def map_serializer_field(self, auto_schema, direction):
+        schema = auto_schema._map_serializer_field(self.target, direction, bypass_extensions=True)
```
arthanson, line 289: I think you can do a simple optimization here:
```
        if (schema := auto_schema._map_serializer_field(self.target, direction, bypass_extensions=True)) is None:
            return schema
```

arthanson (changes requested):
Just the one simple optimization otherwise looks good.

netbox/core/api/schema.py
mraerino: how is that better code? is that something the netbox style guide recommends?

arthanson:
@mraerino sorry one more minor, there is a merge conflict, can you please fix that.  As this code is from a separate repository I can't make the quick fix without creating a new PR.

re the syntax change for the walrus operator (:=) not really required - just more concise, in this case can be debatable if it is more readable or not.  Generally we use that in newer code (since it is a newer python addition) where applicable, but it was a suggestion so if you feel it is cleaner without I'm okay with that.

hikhvar:
@arthanson I only know this syntax from Go, and there an assignment within the condition of an if statement is only valid within that if/else block. We use `schema` later in [line 296](https://github.com/netbox-community/netbox/pull/18451/files#diff-51b2b6c8a0aca95ed0d9d121416fcfde6972e26facee45127647dbca11caac52R296) not in the scope of the if statement. Is this covered by the walrus operator? Just to understand the scoping here.

mraerino:
> @mraerino sorry one more minor, there is a merge conflict, can you please fix that. As this code is from a separate repository I can't make the quick fix without creating a new PR.

I merged main in to fix the conflict. (Funnily it was me who introduced the other change that lead to the conflict.)

Did you know that all people with access to this repo are allowed to push to my branch since i checked "Allow edits by maintainers" on this PR?

>  it was a suggestion so if you feel it is cleaner without I'm okay with that.

I'd rather keep the code like it is to avoid confusion about variable scope as pointed out by @hikhvar

arthanson (approved):

PR #18501, opened 2025-01-27
Fixes: #18241 - Script results log_threshold should default to Default 

### LINKED ISSUE #18241: Script results `log_threshold` should default to Default
### Deployment Type

NetBox Enterprise

### Triage priority

I'm a NetBox Labs customer

### NetBox Version

v4.1.7

### Python Version

3.12

### Steps to Reproduce

1. Run a custom script
2. View the results page `/extras/scripts/results/xxxxx/`

### Expected Behavior

The new "log threshold" setting should use "Default" by default.

### Observed Behavior

The default value is "Debug (All)".
image

I use user-facing custom scripts extensively, and it's important for users to quickly review the output to make sure there are no errors.  Debug output is overwhelming to look at most of the time.  I suggest making the default value an option in user and/or global preferences.

renatoalmeidaoliveira:
### Fixes: #18241 - Script results log_threshold should default to Default 
 
*  Changes the LogLevelChoices `Choices` order, putting `LOG_DEBUG` in the last position.
* Changes `script_result.html` dropdown to add the string `(All)` only to the last possition following the new `LogLevelChoices` order
* Changes `ScriptResultView` get method to return `log_threshold` with `LOG_DEFAULT` instead of `LOG_DEBUG`

### DIFF
--- a/netbox/extras/choices.py
+++ b/netbox/extras/choices.py
@@ -166,7 +166,7 @@ class LogLevelChoices(ChoiceSet):
         (LOG_SUCCESS, _('Success'), 'green'),
         (LOG_WARNING, _('Warning'), 'yellow'),
         (LOG_FAILURE, _('Failure'), 'red'),
-        
+
     )
 
     SYSTEM_LEVELS = {
--- a/netbox/extras/choices.py
+++ b/netbox/extras/choices.py
@@ -162,12 +162,12 @@ class LogLevelChoices(ChoiceSet):
     LOG_FAILURE = 'failure'
 
     CHOICES = (
-        (LOG_DEBUG, _('Debug'), 'teal'),
         (LOG_DEFAULT, _('Default'), 'gray'),
         (LOG_INFO, _('Info'), 'cyan'),
         (LOG_SUCCESS, _('Success'), 'green'),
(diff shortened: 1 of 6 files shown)


netbox/extras/choices.py
```diff
@@ -162,12 +162,12 @@ class LogLevelChoices(ChoiceSet):
     LOG_FAILURE = 'failure'
 
     CHOICES = (
-        (LOG_DEBUG, _('Debug'), 'teal'),
```
bctiemann, line 165: Why is this changed? I think the ordering is supposed to be of increasing severity. Also this should stay consistent with the order of the constants above and in `LOG_LEVEL_RANK`.

Is there something gained by changing the order here and the `.first` to `.last` in the template?

netbox/extras/views.py
```diff
@@ -1315,17 +1315,17 @@ def get_table(self, job, request, bulk_actions=True):
         index = 0
 
         try:
-            log_threshold = LOG_LEVEL_RANK[request.GET.get('log_threshold', LogLevelChoices.LOG_DEBUG)]
```
bctiemann, line 1318: Were these supposed to be changed?

bctiemann (approved):

PR #18506, opened 2025-01-28
Update default dashboard news feed URL

bctiemann:
Updates the RSS feed URL on the dashboard to the final destination of the current chain of 301 redirects.

### DIFF
--- a/netbox/extras/constants.py
+++ b/netbox/extras/constants.py
@@ -76,7 +76,7 @@ DEFAULT_DASHBOARD = [
         'height': 4,
         'title': 'NetBox News',
         'config': {
-            'feed_url': 'http://netbox.dev/rss/',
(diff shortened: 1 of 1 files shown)


jeremystretch:
Just noticed that the content type returned by this URL specifies `application/json` but the content is clearly XML. Doesn't seem to be an issue though.

jeremystretch (approved):

natm:
> Just noticed that the content type returned by this URL specifies `application/json` but the content is clearly XML. Doesn't seem to be an issue though.

fixed this

```
> GET /v1/newsfeed/ HTTP/2
> Host: api.netbox.oss.netboxlabs.com
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/2 200 
< date: Thu, 30 Jan 2025 22:22:41 GMT
< content-type: application/rss+xml
```

Issue #18535, opened 2025-01-29
Allow NetBox to start cleanly if incompatible plugins are present


bctiemann:
### NetBox version

v4.2.2

### Feature type

Data model extension

### Proposed functionality

Currently we raise an `ImproperlyConfigured` exception and refuse to start NetBox at all if any plugin has `min_version` or `max_version` specified such that the installed version of NetBox is outside the compatible range.

Proposal is to allow NetBox to recognize incompatible plugins at startup and skip loading them (while emitting an appropriate warning), rather than raising an exception.

This should probably be done with a custom exception type that inherits from `ImproperlyConfigured` so that these exceptions can be caught separately.

### Use case

It is not always easy to upgrade plugins at the same time as a NetBox installation, or a plugin might not yet be updated to the current NetBox version; this often leads to support issues where plugins must be painstakingly removed from the local configuration during an upgrade. This change would make it possible for plugins' published compatibility ranges to be enforced through the invalid plugins simply being disabled rather than interfering with the overall application's startup.

Note that plugins will need to be more diligent about defining `max_version` and pinning it to the most recent NetBox release against which it has been tested.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
This will require us to track which plugins have been successfully loaded (as opposed to just enabled in the configuration). I've opened #18540 to propose tracking active plugins in the application registry for this purpose.

PR #18537, opened 2025-01-29
Closes: #18535 - Skip incompatible plugins during startup

bctiemann:
### Closes: #18535

Changes the behavior of the `PluginConfig.validate` method so that if a plugin with incompatible min or max version specified, it raises an `IncompatiblePluginError` which is caught in `settings.py` and registering the plugin is skipped, allowing NetBox to launch cleanly with the plugin disabled.


netbox/netbox/settings.py
```diff
@@ -813,6 +815,16 @@ def _setting(name, default=None):
             f"__init__.py file and point to the PluginConfig subclass."
         )
 
+    # Validate version compatibility and user-provided configuration settings and assign defaults
+    if plugin_name not in PLUGINS_CONFIG:
+        PLUGINS_CONFIG[plugin_name] = {}
+    try:
+        plugin_config.validate(PLUGINS_CONFIG[plugin_name], RELEASE.version)
+    except IncompatiblePluginError as e:
+        print(f'Unable to load plugin {plugin_name}: {e}')
```
bctiemann, line 824: I'm not sure about adding a `print` here, as it would be the first of its kind in this file. Is there a better way to emit this warning?

a084ed22:
Just to understand, this would print a warning at netbox upgrade time? The experience of figuring out which plugin has a strict max_version specified and breaks the upgrade process even on patch upgrades during the database migration step could really use to be more explanatory.

bctiemann:
Yes, it would look like this:

```
Unable to load plugin netbox_branching: Plugin netbox_branching requires NetBox maximum version 4.2.1 (current: 4.2.2).
Watching for file changes with StatReloader
Performing system checks...
```

netbox/core/plugins.py
```diff
@@ -91,15 +91,15 @@ def get_local_plugins(plugins=None):
...
+            is_installed=plugin_name in registry['plugins']['installed'],
             installed_version=installed_version,
         )
 
     # Update catalog entries for local plugins, or add them to the list if not listed
     for k, v in local_plugins.items():
         if k in plugins:
             plugins[k].is_local = True
-            plugins[k].is_installed = True
+            plugins[k].is_installed = k in registry['plugins']['installed']
```
jeremystretch, line 102: I think we can copy this over directly?
```suggestion
            plugins[k].is_installed = v.is_installed
```

jeremystretch (changes requested):
We might want to rename the table columns. Currently, a plugin which fails to load shows false under "Installed" but confusingly still reports an installed version:

screenshot

We should also highlight any plugins which failed to load to draw attention to them.

bctiemann:
I renamed "Installed" to "Active", and I added a custom table column that allows an alert icon with a tooltip that explains the incompatibility issue:

Screenshot 2025-02-23 at 6 51 43 PM
Screenshot 2025-02-23 at 6 50 32 PM

netbox/netbox/settings.py
```diff
@@ -811,6 +812,15 @@ def _setting(name, default=None):
             f"__init__.py file and point to the PluginConfig subclass."
         )
 
+    # Validate version compatibility and user-provided configuration settings and assign defaults
+    if plugin_name not in PLUGINS_CONFIG:
+        PLUGINS_CONFIG[plugin_name] = {}
+    try:
+        plugin_config.validate(PLUGINS_CONFIG[plugin_name], RELEASE.version)
```
jeremystretch, line 819: We can simplify this a bit:
```suggestion
    try:
        plugin_config.validate(PLUGINS_CONFIG.get(plugin_name, {}), RELEASE.version)
```

jeremystretch (changes requested):
This renders an alert for every non-installed plugin. The value for `is_installed` should probably be None if the plugin has not been added to `PLUGINS`, and False if it has been added but failed to load.

screenshot

netbox/netbox/settings.py
bctiemann: Actually I don't think this can be done this way. Trying to `validate` in one fell swoop like this calls `get_plugin_config` which raises an `ImproperlyConfigured` error which I'm not trapping in this try-except (intentionally), and that fires before it checks whether the key is in the dict. It's weird and ugly but the way I had it is the only way to do it without changing `get_plugin_config` too with even more abstractions.

netbox/netbox/tables/columns.py
```diff
@@ -707,3 +707,22 @@ class DistanceColumn(TemplateColumn):
 
     def __init__(self, template_code=template_code, order_by='_abs_distance', **kwargs):
         super().__init__(template_code=template_code, order_by=order_by, **kwargs)
+
+
+class PluginActiveColumn(BooleanColumn):
```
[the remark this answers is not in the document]
jeremystretch: Could you just use a TemplateColumn then? I'd like to avoid introducing a new column class that only has one use case.

netbox/core/plugins.py
```diff
@@ -66,7 +66,10 @@ class Plugin:
     release_recent_history: list[PluginVersion] = field(default_factory=list)
     is_local: bool = False  # extra field for locally installed plugins
     is_installed: bool = False
+    failed_to_load: bool = False
```
jeremystretch, line 69: I'm having a hard time discerning the difference between `is_local` and `is_installed`. It seems that any installed plugin has `is_local` set to True. Additionally, `failed_to_load` feels backward to me.

Could we simplify these to two attributes: `is_enabled` and `is_installed` (or `is_loaded`)? These would indicate whether the plugin is listed in `settings.PLUGINS` and whether is has been successfully loaded into the registry, respectively.

netbox/core/tables/template_code.py
```diff
@@ -14,3 +14,13 @@
 OBJECTCHANGE_REQUEST_ID = """
 <a href="{% url 'core:objectchange_list' %}?request_id={{ value }}">{{ value }}</a>
 """
+
+PLUGIN_IS_INSTALLED = """
+{% if record.failed_to_load %}
```
jeremystretch, line 19: A plugin could fail to load for other reasons though, presumably. I'd avoid assuming it was due to a version incompatibility.

jeremystretch (changes requested):

netbox/core/plugins.py
bctiemann: How about this -- `is_local` (which mirrors the `local_plugins` etc verbage, as opposed to `catalog_plugins`); and `is_loaded`, which indicates whether it's loaded into the registry. This removes `is_installed` as redundant, and the `is_installed` column remains on the table class but is purely rendered via the template logic.

netbox/core/tables/template_code.py
bctiemann: Rephrased the tooltip accordingly. It's possible to do an on-the-fly validation of the version range to see whether it should be suggested at all, but that got a bit too hairy.

jeremystretch (approved):

PR #18544, opened 2025-01-30
FIX #18154 - Cannot change default table preferences for anonymous users 

### LINKED ISSUE #18154: Cannot change default table preferences for anonymous users
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.7

### Python Version

3.10

### Steps to Reproduce

Same issue as #13312 

Edit configuration.yaml to edit default user preferences table columns for IPAM>IPAddresses

DEFAULT_USER_PREFERENCES = {
    'tables': {
        'IPAddressTable': {
            'columns': ['address', 'status', 'dns_name', 'description', 'cf_Used_for_Egress', ]
        },
    },
}

image

Note: does not need custom field to cause issue. 

Open Netbox as anonymous/unauthenticated user (private browsing), and navigate to IPAM>IP Addresses
The table columns do not match the default user preferences. 

### Expected Behavior

The IP Address table columns for guests should match what is set in the default user preferences. 

### Observed Behavior

When logging in as an anonymous/unauthenticated user, the table columns are different than what is set under default user preferences.

image

This is an issue because guests do not have the option to adjust table columns, so they cannot see any non-default table fields without opening each IP address. 

Note: for authenticated, if you reset table column preferences, then it will change to the preferences set under the default user preferences.

renatoalmeidaoliveira:
### Fixes: #18154

* Change BaseTable to check if DEFAULT_USER_PREFERENCES are defined
* Change BaseTable selected_columns behaivour to be aware of DEFAULT_USER_PREFERENCES table configuration

### DIFF
--- a/netbox/netbox/tables/tables.py
+++ b/netbox/netbox/tables/tables.py
@@ -2,6 +2,7 @@ from copy import deepcopy
 from functools import cached_property
 
 import django_tables2 as tables
+from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.core.exceptions import FieldDoesNotExist
@@ -12,6 +13,7 @@ from django.urls.exceptions import NoReverseMatch
 from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy as _
 from django_tables2.data import TableQuerysetData
+
 
 from core.models import ObjectType
 from extras.choices import *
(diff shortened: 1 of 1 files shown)


netbox/netbox/tables/tables.py
```diff
@@ -64,6 +66,11 @@ def __init__(self, *args, user=None, **kwargs):
         selected_columns = None
         if user is not None and not isinstance(user, AnonymousUser):
             selected_columns = user.config.get(f"tables.{self.name}.columns")
+        elif isinstance(user, AnonymousUser) and hasattr(settings, 'DEFAULT_USER_PREFERENCES'):
+            default_user_preferences = settings.DEFAULT_USER_PREFERENCES
+            default_table = default_user_preferences.get('tables', {}).get(self.name, {}).get('columns', {})
```
bctiemann, line 71: Could probably simplify this to

```
selected_columns = settings.DEFAULT_USER_PREFERENCES.get('tables', {}).get(self.name, {}).get('columns')
```

Right? Or is there some logical case I'm missing?

renatoalmeidaoliveira: We still need to check is settings have `DEFAULT_USER_PREFERENCES` attribute, my first implementation I've missed that test and a lot of unit testa fails because the `DEFAULT_USER_PREFERENCES` aren't a required configuration. 
But the next parts is definitely simpler, gonna change to that format.

netbox/netbox/tables/tables.py
```diff
@@ -13,6 +14,7 @@
 from django.utils.translation import gettext_lazy as _
 from django_tables2.data import TableQuerysetData
 
+
```
bctiemann, line 17: This extra line should be removed

bctiemann (approved):

PR #18551, opened 2025-01-31
Fixes: #18013 - All saved filters showing up for Change Log table, regardless of defined object type

### LINKED ISSUE #18013: All saved filters showing up for Change Log table, regardless of defined object type
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.6

### Python Version

3.11

### Steps to Reproduce

1. Create a filter, selecting an Object type other than Object Change.
2. Go to Operations > Change Log
3. Open the filter dropdown

### Expected Behavior

No filter/only filters with object type Object Change show(s) up.

### Observed Behavior

The created filter shows up.

See #15638. This fix does not seem to work for the change log.

renatoalmeidaoliveira:
### Fixes: #18013 - All saved filters showing up for Change Log table, regardless of defined object type

Since ObjectChangeFilterForm doesn't inherid from NetBoxModelFilterSetForm the filter_id field doesn't get updated

Changes:

* Add `__init__` method to ObjectChangeFilterForm, and setup filter_id `query_params`

### DIFF
--- a/netbox/core/forms/filtersets.py
+++ b/netbox/core/forms/filtersets.py
@@ -62,6 +62,7 @@ class DataFileFilterForm(NetBoxModelFilterSetForm):
 
 
 class JobFilterForm(SavedFiltersMixin, FilterForm):
+    model = Job
     fieldsets = (
         FieldSet('q', 'filter_id'),
         FieldSet('object_type', 'status', name=_('Attributes')),
@@ -162,6 +163,7 @@ class ObjectChangeFilterForm(SavedFiltersMixin, FilterForm):
 
 
 class ConfigRevisionFilterForm(SavedFiltersMixin, FilterForm):
+    model = ConfigRevision
     fieldsets = (
         FieldSet('q', 'filter_id'),
     )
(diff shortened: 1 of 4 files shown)


renatoalmeidaoliveira:
That issue also affects:
`Config Contexts`
`Config Templates`
`Custom Fields`
`Custom Field Choice Sets`
`Custom Links`
`Export Templates`
`Saved Filters`
`Image Attachments`
`Jobs`
`Notification Groups`

renatoalmeidaoliveira:
The current solution, solves the issue only for ChangeLog as reported in the issue.
For a broaden solution, IMO the best approach is change the `SavedFiltersMixin` adding a `__init__` block that setups the `object_type_id` query_param in the `filter_id` widget and remove that block from `NetBoxModelFilterSetForm`
The solution also requires to add a `model` attribute for every `FilterForm` that doesn't have it already, like: `JobFilterForm`

bctiemann:
@renatoalmeidaoliveira I think you're right, this functionality should go into `SavedFiltersMixin`. I also see what you mean about the `JobFilterForm` and other similar forms -- let's go ahead and add those `model` attributes and make sure everything still works as expected.

jeremystretch:
Found a couple more forms that need to have `model` declared:

- ConfigRevisionFilterForm
- JobFilterForm

I can knock these out real quick.

jeremystretch (approved):

PR #18583, opened 2025-02-06
Fixes: #17976 DeviceType_Count on Manufacturer 

### LINKED ISSUE #17976: DeviceType_Count on Manufacturer
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.6

### Python Version

3.10

### Steps to Reproduce

GET /api/dcim/device-types/ 

```json
{
    "count": 2424,
    "next": "https://netbox.com/api/dcim/device-types/?limit=50&offset=50",
    "previous": null,
    "results": [
        {
            "id": 1112,
            "url": "https://netbox.com/api/dcim/device-types/1112/",
            "display_url": "https://netbox.com/dcim/device-types/1112/",
            "display": "Thunder 4440 ADC",
            "manufacturer": {
                "id": 47,
                "url": "https://netbox.com/api/dcim/manufacturers/47/",
                "display": "A10",
                "name": "A10",
                "slug": "a10",
                "description": ""
            },
            "default_platform": null,
            "model": "Thunder 4440 ADC",
            "slug": "a10-thunder-4440-adc",
            "part_number": "TH4440",
            "u_height": 1.0,
            "exclude_from_utilization": false,
            "is_full_depth": true,
            "subdevice_role": null,
            "airflow": null,
            "weight": null,
            "weight_unit": null,
            "front_image": null,
            "rear_image": null,
            "description": "",
            "comments": "[A10 Thunder 4440 Applcation Delivery Controller Datasheet](https://www.a10networks.com/wp-content/uploads/A10-DS-Thunder-ADC.pdf)",
            "tags": [],
            "custom_fields": {},
            "created": "2024-10-28T06:49:53.123905-04:00",
            "last_updated": "2024-10-28T06:49:53.123918-04:00",
            "device_count": 0,
            "console_port_template_count": 0,
            "console_server_port_template_count": 0,
            "power_port_template_count": 0,
            "power_outlet_template_count": 0,
            "interface_template_count": 8,
            "front_port_template_count": 0,
            "rear_port_template_count": 0,
            "device_bay_template_count": 0,
            "module_bay_template_count": 0,
            "inventory_item_template_count": 0
        }
    ]
}
```

Note `"devicetype_count"` is missing.

### Expected Behavior

The expected response, according to the OpenAPI specification for Netbox:

GET /api/dcim/device-types/ 

```json
{
    "count": 2424,
    "next": "https://netbox.com/api/dcim/device-types/?limit=50&offset=50",
    "previous": null,
    "results": [
        {
            "id": 1112,
            "url": "https://netbox.com/api/dcim/device-types/1112/",
            "display_url": "https://netbox.com/dcim/device-types/1112/",
            "display": "Thunder 4440 ADC",
            "manufacturer": {
                "id": 47,
                "url": "https://netbox.com/api/dcim/manufacturers/47/",
                "display": "A10",
                "name": "A10",
                "slug": "a10",
                "description": "",
                "devicetype_count": 1
            },
            "default_platform": null,
            "model": "Thunder 4440 ADC",
            "slug": "a10-thunder-4440-adc",
            "part_number": "TH4440",
            "u_height": 1.0,
            "exclude_from_utilization": false,
            "is_full_depth": true,
            "subdevice_role": null,
            "airflow": null,
            "weight": null,
            "weight_unit": null,
            "front_image": null,
            "rear_image": null,
            "description": "",
            "comments": "[A10 Thunder 4440 Applcation Delivery Controller Datasheet](https://www.a10networks.com/wp-content/uploads/A10-DS-Thunder-ADC.pdf)",
            "tags": [],
            "custom_fields": {},
            "created": "2024-10-28T06:49:53.123905-04:00",
            "last_updated": "2024-10-28T06:49:53.123918-04:00",
            "device_count": 0,
            "console_port_template_count": 0,
            "console_server_port_template_count": 0,
            "power_port_template_count": 0,
            "power_outlet_template_count": 0,
            "interface_template_count": 8,
            "front_port_template_count": 0,
            "rear_port_template_count": 0,
            "device_bay_template_count": 0,
            "module_bay_template_count": 0,
            "inventory_item_template_count": 0
        }
    ]
}
```

### Observed Behavior

According to the OpenAPI specification BriefManufacturer has a required field of `"devicetype_count"`. However, when querying DeviceTypes and ModuleTypes, the field is not populated and causes SDKs using the OpenAPI generator, to fail.

Go-Netbox [Creating Manufacture errors out because of missing required property devicetype_count #165](https://github.com/netbox-community/go-netbox/issues/165)

renatoalmeidaoliveira:
### Fixes: #17976 DeviceType_Count on Manufacturer 
 
* Changed `get_prefetches_for_serializer` to support nested serializers
* For each nested field select all  `RelatedObjectCountField` fields and create a `Prefetch` with the correct annotation

### DIFF
--- a/netbox/utilities/api.py
+++ b/netbox/utilities/api.py
@@ -88,7 +88,7 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None, sour
         fields_to_include = serializer_class.Meta.fields
 
     prefetch_fields = []
-    annotaded_prefetch = {}
+    annotated_prefetch = {}
     for field_name in fields_to_include:
         serializer_field = serializer_class._declared_fields.get(field_name)
 
@@ -98,10 +98,10 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None, sour
             model_field_name = serializer_field.source
 
         # If the serializer field is a RelatedObjectCountField and its a nested field
-        # Add an annotation to the annotaded_prefetch
+        # Add an annotation to the annotated_prefetch
         if isinstance(serializer_field, RelatedObjectCountField) and source_field is not None:
-            if model_field_name not in annotaded_prefetch:
-                annotaded_prefetch[model_field_name] = Count(serializer_field.relation)
+            if model_field_name not in annotated_prefetch:
+                annotated_prefetch[model_field_name] = Count(serializer_field.relation)
 
         # If the serializer field does not map to a discrete model field, skip it.
         try:
@@ -124,9 +124,9 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None, sour
                     else:
                         prefetch_fields.append(f'{field_name}__{subfield}')
 
-    # If there are annotaded_prefetch, add the annotaded prefetch to the prefetch_fields
-    if annotaded_prefetch:
-        related_prefetch = Prefetch(source_field, queryset=model.objects.all().annotate(**annotaded_prefetch))
+    # If there are annotated_prefetch, add the annotaded prefetch to the prefetch_fields
+    if annotated_prefetch:
+        related_prefetch = Prefetch(source_field, queryset=model.objects.all().annotate(**annotated_prefetch))
         prefetch_fields.append(related_prefetch)
     return prefetch_fields
 
--- a/netbox/utilities/api.py
+++ b/netbox/utilities/api.py
@@ -3,6 +3,7 @@ from django.core.exceptions import (
     FieldDoesNotExist, FieldError, MultipleObjectsReturned, ObjectDoesNotExist, ValidationError,
 )
 from django.db.models.fields.related import ManyToOneRel, RelatedField
+from django.db.models import Count, Prefetch
 from django.urls import reverse
 from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
@@ -76,7 +77,7 @@ def get_view_name(view):
     return drf_get_view_name(view)
 
 
-def get_prefetches_for_serializer(serializer_class, fields_to_include=None):
+def get_prefetches_for_serializer(serializer_class, fields_to_include=None, source_field=None):
     """
     Compile and return a list of fields which should be prefetched on the queryset for a serializer.
     """
@@ -87,6 +88,7 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None):
         fields_to_include = serializer_class.Meta.fields
 
     prefetch_fields = []
+    annotaded_prefetch = {}
     for field_name in fields_to_include:
         serializer_field = serializer_class._declared_fields.get(field_name)
 
@@ -95,23 +97,37 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None):
         if serializer_field and serializer_field.source:
             model_field_name = serializer_field.source
 
+        # If the serializer field is a RelatedObjectCountField and its a nested field
+        # Add an annotation to the annotaded_prefetch
+        if isinstance(serializer_field, RelatedObjectCountField) and source_field is not None:
+            if model_field_name not in annotaded_prefetch:
+                annotaded_prefetch[model_field_name] = Count(serializer_field.relation)
+
         # If the serializer field does not map to a discrete model field, skip it.
         try:
             field = model._meta.get_field(model_field_name)
-            if isinstance(field, (RelatedField, ManyToOneRel, GenericForeignKey)):
+            if (isinstance(field, (RelatedField, ManyToOneRel, GenericForeignKey)) and
+                 not issubclass(type(serializer_field), Serializer)):
                 prefetch_fields.append(field.name)
         except FieldDoesNotExist:
             continue
 
         # If this field is represented by a nested serializer, recurse to resolve prefetches
         # for the related object.
-        if serializer_field:
+        if serializer_field and source_field is None:
             if issubclass(type(serializer_field), Serializer):
                 # Determine which fields to prefetch for the nested object
                 subfields = serializer_field.Meta.brief_fields if serializer_field.nested else None
-                for subfield in get_prefetches_for_serializer(type(serializer_field), subfields):
-                    prefetch_fields.append(f'{field_name}__{subfield}')
+                for subfield in get_prefetches_for_serializer(type(serializer_field), subfields, field_name):
+                    if isinstance(subfield, Prefetch):
+                        prefetch_fields.append(subfield)
+                    else:
+                        prefetch_fields.append(f'{field_name}__{subfield}')
 
+    # If there are annotaded_prefetch, add the annotaded prefetch to the prefetch_fields
+    if annotaded_prefetch:
+        related_prefetch = Prefetch(source_field, queryset=model.objects.all().annotate(**annotaded_prefetch))
+        prefetch_fields.append(related_prefetch)
     return prefetch_fields
 
 
--- a/netbox/utilities/api.py
+++ b/netbox/utilities/api.py
@@ -3,7 +3,7 @@ from django.core.exceptions import (
     FieldDoesNotExist, FieldError, MultipleObjectsReturned, ObjectDoesNotExist, ValidationError,
 )
 from django.db.models.fields.related import ManyToOneRel, RelatedField
-from django.db.models import Count, Prefetch
+from django.db.models import Prefetch
 from django.urls import reverse
 from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
@@ -88,7 +88,14 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None, sour
         fields_to_include = serializer_class.Meta.fields
 
     prefetch_fields = []
-    annotated_prefetch = {}
+
+    # If this serializer is nested, get annotations and prefetches for the nested serializer
+    if source_field is not None:
+        nested_annotations = get_annotations_for_serializer(serializer_class, fields_to_include=fields_to_include)
+        if nested_annotations:
+            related_prefetch = Prefetch(source_field, queryset=model.objects.all().annotate(**nested_annotations))
+            prefetch_fields.append(related_prefetch)
+
     for field_name in fields_to_include:
         serializer_field = serializer_class._declared_fields.get(field_name)
 
@@ -96,12 +103,6 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None, sour
         model_field_name = field_name
         if serializer_field and serializer_field.source:
             model_field_name = serializer_field.source
-
-        # If the serializer field is a RelatedObjectCountField and its a nested field
-        # Add an annotation to the annotated_prefetch
-        if isinstance(serializer_field, RelatedObjectCountField) and source_field is not None:
-            if model_field_name not in annotated_prefetch:
-                annotated_prefetch[model_field_name] = Count(serializer_field.relation)
 
         # If the serializer field does not map to a discrete model field, skip it.
         try:
@@ -124,10 +125,6 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None, sour
                     else:
                         prefetch_fields.append(f'{field_name}__{subfield}')
 
-    # If there are annotated_prefetch, add the annotaded prefetch to the prefetch_fields
-    if annotated_prefetch:
-        related_prefetch = Prefetch(source_field, queryset=model.objects.all().annotate(**annotated_prefetch))
-        prefetch_fields.append(related_prefetch)
     return prefetch_fields
 
 
--- a/netbox/utilities/api.py
+++ b/netbox/utilities/api.py
@@ -1,6 +1,8 @@
+from django.contrib.contenttypes.fields import GenericForeignKey
 from django.core.exceptions import (
     FieldDoesNotExist, FieldError, MultipleObjectsReturned, ObjectDoesNotExist, ValidationError,
 )
+from django.db.models.fields.related import ManyToOneRel, RelatedField
 from django.db.models import Prefetch
 from django.urls import reverse
 from django.utils.module_loading import import_string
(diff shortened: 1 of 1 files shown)


jeremystretch:
@renatoalmeidaoliveira it would be ideal to solve this by enabling recursion through nested serializers inside `get_annotations_for_serializer()` if possible. This is because these annotations employ the `count_related()` utility function, which IIRC works a bit differently from the standard `Count()`. Did you try going down this path at all?

renatoalmeidaoliveira:
@jeremystretch that was my first try, i.e. using `get_annotations_for_serializer` I tryed that in two ways the one proposed by @arthanson in the issue and another version I made.
In both ways the annotated attribute falls inside the base model instead of the nested one and even using the nottation `{nested_model}__count_devices`.
It's possible to reprodute that behaivour inside nbshell, then I tryed to work on  `get_prefetches_for_serializer`, in such way it start to return the annotated attriibute, and that works.
Now I think it's possible to use `count_related` inside the `get_prefetches_for_serializer`, gonna try to implement that way.

renatoalmeidaoliveira:
About using `count_related` I was able to use it inside the `get_prefetches_for_serializer`.
Looking at the query at Django Debug we have the following differences:

 ## Using `count_related`

```
SELECT "dcim_manufacturer"."id",
       "dcim_manufacturer"."created",
       "dcim_manufacturer"."last_updated",
       "dcim_manufacturer"."custom_field_data",
       "dcim_manufacturer"."name",
       "dcim_manufacturer"."slug",
       "dcim_manufacturer"."description",
       COALESCE((SELECT COUNT(*) AS "c" FROM "dcim_manufacturer" U0 INNER JOIN "dcim_devicetype" U1 ON (U0."id" = U1."manufacturer_id") WHERE U1."id" = ("dcim_manufacturer"."id") GROUP BY U1."id"), 0) AS "devicetype_count"
  FROM "dcim_manufacturer"
 WHERE "dcim_manufacturer"."id" IN (3)
```
### Performance using `count_related`

count_related


## Using `Count`

```
SELECT "dcim_manufacturer"."id",
       "dcim_manufacturer"."created",
       "dcim_manufacturer"."last_updated",
       "dcim_manufacturer"."custom_field_data",
       "dcim_manufacturer"."name",
       "dcim_manufacturer"."slug",
       "dcim_manufacturer"."description",
       COUNT("dcim_devicetype"."id") AS "devicetype_count"
  FROM "dcim_manufacturer"
  LEFT OUTER JOIN "dcim_devicetype"
    ON ("dcim_manufacturer"."id" = "dcim_devicetype"."manufacturer_id")
 WHERE "dcim_manufacturer"."id" IN (3)
 GROUP BY "dcim_manufacturer"."id"
``` 

### Performance using `Count`
Count


## Summary

In my environment, using NetBox demo data, the performance using `Count` was a bit better, but both ways works.
Please let me know the preffered way.

renatoalmeidaoliveira:
About changing `get_annotations_for_serializer`, I tryed with this approach:
 ```
 def get_annotations_for_serializer(serializer_class, fields_to_include=None):
    """
    Return a mapping of field names to annotations to be applied to the queryset for a serializer.
    """
    annotations = {}
    # If specific fields are not specified, default to all
    if not fields_to_include:
        fields_to_include = serializer_class.Meta.fields

    model = serializer_class.Meta.model

    for field_name, field in serializer_class._declared_fields.items():
        if issubclass(type(field), Serializer):
            nested_fields = field.Meta.brief_fields if field.nested else []
            for subfield_name, subfield in field.get_fields().items():
                if subfield_name in nested_fields and type(subfield) is RelatedObjectCountField:
                    subfield_model = field.Meta.model
                    annotations[f'{field_name}__{subfield_name}'] = count_related(subfield_model, subfield.relation)
        if field_name in fields_to_include and type(field) is RelatedObjectCountField:
            related_field = getattr(model, field.relation).field
            annotations[field_name] = count_related(related_field.model, related_field.name)
    return annotations
```

Running with this with Debug it executes the following query:
```
SELECT "dcim_devicetype"."id",
       "dcim_devicetype"."created",
       "dcim_devicetype"."last_updated",
       "dcim_devicetype"."custom_field_data",
       "dcim_devicetype"."description",
       "dcim_devicetype"."comments",
       "dcim_devicetype"."weight",
       "dcim_devicetype"."weight_unit",
       "dcim_devicetype"."_abs_weight",
       "dcim_devicetype"."manufacturer_id",
       "dcim_devicetype"."model",
       "dcim_devicetype"."slug",
       "dcim_devicetype"."default_platform_id",
       "dcim_devicetype"."part_number",
       "dcim_devicetype"."u_height",
       "dcim_devicetype"."exclude_from_utilization",
       "dcim_devicetype"."is_full_depth",
       "dcim_devicetype"."subdevice_role",
       "dcim_devicetype"."airflow",
       "dcim_devicetype"."front_image",
       "dcim_devicetype"."rear_image",
       "dcim_devicetype"."console_port_template_count",
       "dcim_devicetype"."console_server_port_template_count",
       "dcim_devicetype"."power_port_template_count",
       "dcim_devicetype"."power_outlet_template_count",
       "dcim_devicetype"."interface_template_count",
       "dcim_devicetype"."front_port_template_count",
       "dcim_devicetype"."rear_port_template_count",
       "dcim_devicetype"."device_bay_template_count",
       "dcim_devicetype"."module_bay_template_count",
       "dcim_devicetype"."inventory_item_template_count",
       COALESCE((SELECT COUNT(*) AS "c" FROM "dcim_manufacturer" U0 INNER JOIN "dcim_devicetype" U1 ON (U0."id" = U1."manufacturer_id") WHERE U1."id" = ("dcim_devicetype"."id") GROUP BY U1."id"), 0) AS "manufacturer__devicetype_count",
       COALESCE((SELECT COUNT(*) AS "c" FROM "dcim_platform" U0 INNER JOIN "dcim_device" U1 ON (U0."id" = U1."platform_id") WHERE U1."id" = ("dcim_devicetype"."id") GROUP BY U1."id"), 0) AS "default_platform__device_count",
       COALESCE((SELECT COUNT(*) AS "c" FROM "dcim_platform" U0 INNER JOIN "virtualization_virtualmachine" U1 ON (U0."id" = U1."platform_id") WHERE U1."id" = ("dcim_devicetype"."id") GROUP BY U1."id"), 0) AS "default_platform__virtualmachine_count",
       COALESCE((SELECT COUNT(*) AS "c" FROM "dcim_device" U0 WHERE U0."device_type_id" = ("dcim_devicetype"."id") GROUP BY U0."device_type_id"), 0) AS "device_count"
  FROM "dcim_devicetype"
 WHERE "dcim_devicetype"."id" = 7
 LIMIT 21
```

And looking at the  results inside Django debug, its returning the correct results but as an attribute of `DeviceType` :


image

So even the query returning the expected results the JSON output omits the result instead of adding it inside the Nested Object

File: netbox/utilities/api.py [hunk 1]
```diff
@@ -95,23 +97,37 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None):
         if serializer_field and serializer_field.source:
             model_field_name = serializer_field.source
 
+        # If the serializer field is a RelatedObjectCountField and its a nested field
+        # Add an annotation to the annotaded_prefetch
+        if isinstance(serializer_field, RelatedObjectCountField) and source_field is not None:
+            if model_field_name not in annotaded_prefetch:
+                annotaded_prefetch[model_field_name] = Count(serializer_field.relation)
```
arthanson, line 104: Do we need to continue processing at this point?  I think we can just do a continue here?

netbox/utilities/api.py
```diff
@@ -95,23 +97,37 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None):
...
+
         # If the serializer field does not map to a discrete model field, skip it.
         try:
             field = model._meta.get_field(model_field_name)
-            if isinstance(field, (RelatedField, ManyToOneRel, GenericForeignKey)):
+            if (isinstance(field, (RelatedField, ManyToOneRel, GenericForeignKey)) and
```
arthanson, line 109: This can be moved below the try/except - if the field doesn't exist we are doing a continue so it won't get hit and field will be set.
Also - if it is `issubclass(type(serializer_field), Serializer)` do we want to continue to line 117?  Not sure - just asking, could we just do a check on that condition and continue?

arthanson (changes requested):

netbox/utilities/api.py
```diff
@@ -98,20 +105,24 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None):
         # If the serializer field does not map to a discrete model field, skip it.
         try:
             field = model._meta.get_field(model_field_name)
-            if isinstance(field, (RelatedField, ManyToOneRel, GenericForeignKey)):
-                prefetch_fields.append(field.name)
         except FieldDoesNotExist:
             continue
 
+        # Prefetch for tags
```
arthanson, line 111: This code is close but I think this is way too specific checking for a field named 'tags'. There shouldn't be anything special about tags - we should be able to check generically for M2M or FK field here.

arthanson (changes requested):

netbox/utilities/api.py
arthanson: If this is needed can you do something like:
```
if isinstance(field, models.ManyToManyField):
    prefetch_fields.append(field.name)
elif serializer_field and source_field is None:
   ...
```

netbox/utilities/api.py
```diff
@@ -87,6 +88,14 @@ def get_prefetches_for_serializer(serializer_class, fields_to_include=None):
         fields_to_include = serializer_class.Meta.fields
 
     prefetch_fields = []
+
+    # If this serializer is nested, get annotations and prefetches for the nested serializer
+    if source_field is not None:
+        nested_annotations = get_annotations_for_serializer(serializer_class, fields_to_include=fields_to_include)
+        if nested_annotations:
+            related_prefetch = Prefetch(source_field, queryset=model.objects.all().annotate(**nested_annotations))
+            prefetch_fields.append(related_prefetch)
```
arthanson, line 97: Almost there - I think you can just add a return prefetch_fields here.  If it's gotten to here then in the loop below on line 99 the only time it modifies prefetch_fields is on line 125 and that can only happen if source_field is None, which it wouldn't be in this case, so all that code is looping / processing for nothing.

If that is the case the fields_to_include doesn't matter in this case.  I'm seeing when it is called with source_field you are only getting one value returned for prefetch_fieds (which makes sense as the for loop doesn't append anything) Therefore it doesn't look like you need to recurse into this at all, the lines above can just be a separate function that is called (without the fields_to_include). 

Please let me know if that isn't clear.

arthanson (changes requested):

renatoalmeidaoliveira:
@arthanson I created a method to deal with subfield annotations, and remove the recursion, I think in most cases it gonna work well because the brief fields ussually doesn't contains serializer fields, the only exception I found was `PowerPort`.
Comparing the `prefetch` from the current implementation and this one the difference is:

#### Current Implementation:
`['device', 'module', 'module__device', 'module__module_bay', 'module__module_type', 'module__module_type__manufacturer', 'cable', 'tags']`

#### New implementation:
`['device', 'module', 'cable', 'tags']`

But even with that difference with the prefetch field the amount of queries in the list endpoint was only by one query in my environment with demo data

arthanson (changes requested):
I didn't mean take out the recursion, as you noted that removes many of the prefetch's that are needed.  At line 19 below I've added a return after the append as the code below it has no effect (since in this case `source_field is not None` 

What I wasn't sure on was if the line `get_prefetches_for_serializer(type(serializer_field), subfields, field_name)` could be simplified in the case where it is passing in source_field, but looking at it more this might be the easiest way.  So I'd suggest reverting to the code I posted below.

For more clarification: You are recursing into a function but it has two completely separate logic paths if source_field is None or not.  Wasn't sure if it would be cleaner to call two separate functions (do the recursion if source_field is None, but call another function if source_field isn't None that just does the code you have after `if source_field is not None`.  Again, it might be simpler to just keep it all in one function like you have it.

```
def get_prefetches_for_serializer(serializer_class, fields_to_include=None, source_field=None):
    """
    Compile and return a list of fields which should be prefetched on the queryset for a serializer.
    """
    model = serializer_class.Meta.model

    # If fields are not specified, default to all
    if not fields_to_include:
        fields_to_include = serializer_class.Meta.fields

    prefetch_fields = []

    # If this serializer is nested, get annotations and prefetches for the nested serializer
    if source_field is not None:
        nested_annotations = get_annotations_for_serializer(serializer_class, fields_to_include=fields_to_include)
        if nested_annotations:
            related_prefetch = Prefetch(source_field, queryset=model.objects.all().annotate(**nested_annotations))
            prefetch_fields.append(related_prefetch)
            return prefetch_fields

    for field_name in fields_to_include:
        serializer_field = serializer_class._declared_fields.get(field_name)

        # Determine the name of the model field referenced by the serializer field
        model_field_name = field_name
        if serializer_field and serializer_field.source:
            model_field_name = serializer_field.source

        # If the serializer field does not map to a discrete model field, skip it.
        try:
            field = model._meta.get_field(model_field_name)
        except FieldDoesNotExist:
            continue

        # If this field is represented by a nested serializer, recurse to resolve prefetches
        # for the related object.
        if serializer_field and source_field is None:
            if issubclass(type(serializer_field), Serializer):
                # Determine which fields to prefetch for the nested object
                subfields = serializer_field.Meta.brief_fields if serializer_field.nested else None
                for subfield in get_prefetches_for_serializer(type(serializer_field), subfields, field_name):
                    if isinstance(subfield, Prefetch):
                        prefetch_fields.append(subfield)
                    else:
                        prefetch_fields.append(f'{field_name}__{subfield}')
            elif isinstance(field, (RelatedField, ManyToOneRel, GenericForeignKey)):
                prefetch_fields.append(field.name)
    return prefetch_fields
```

renatoalmeidaoliveira:
Closing this issue since I didn't find a good solution to keep the recursion and allow Prefetches with annotations.
Described my findings in the Issue

PR #18603, opened 2025-02-07
Fixes 17309: Omit empty related models

### LINKED ISSUE #17309: Related Objects card rendering, omit empty
### NetBox version

v3.7.8

### Feature type

### Proposed functionality

Proposing a change in behaviour (toggleable or otherwise) that omits rendering Related Objects where said object relations don't exist.

I have naively achieved this (superficially?) with modification to `/netbox/templates/inc/panels/related_objects.html`
Bearing in mind that this was an educated guess at best and not necessarily considering efficient implementation, wider compatibility or design intent.

Note this code is from v3.7.8. However the behaviour looks to be unchanged on 4.0 and 4.1beta when looking at the demo instances and code.
Have also chased some of discussions/issues/pulls related to these changes in behaviour to gain insights.
e.g. #14510 

From v3.7.8 [netbox/templates/inc/panels/related_objects.html](https://github.com/netbox-community/netbox/blob/v3.7.8/netbox/templates/inc/panels/related_objects.html)
```
{% load helpers %}
... <brevity omit>
    {% for qs, filter_param in related_models %}
      {% if qs %} <-------------------- insertion
      {% with viewname=qs.model|viewname:"list" %}
        <a href="{% url viewname %}?{{ filter_param }}={{ object.pk }}" class="list-group-item list-group-item-action d-flex justify-content-between">
          {{ qs.model|meta:"verbose_name_plural"|bettertitle }}
          {% with count=qs.count %}
...<brevity omit>
          {% endwith %}
        </a>
      {% endwith %}
      {% endif %} <-------------------- insertion
    {% endfor %}
...<brevity omit>
```
EDIT: removed some snippet code I had snuck in from #16825 while testing.

Outcome on viewing an arbitrary Tenant:
netbox_related_objects

If I have a case and am barking up the right tree, then can attempt contribution.

### Use case

Reduction of scroll fatigue in page views. We only recently shifted from v3.6.9 -> v3.7.8 and had a lot of feedback internally about the change in behaviour. Our user base aren't necessarily highly technical.

### Database changes

Highly unlikely.

### External dependencies

Highly unlikely.

alehaa:
### Fixes: #17309

Related models that count as zero are omitted from the Related Models pane. If no models are related to the one being displayed, the pane displays "None" instead.

### DIFF
--- a/netbox/templates/inc/panels/related_objects.html
+++ b/netbox/templates/inc/panels/related_objects.html
@@ -19,6 +19,8 @@
         </a>
         {% endif %}
       {% endwith %}
+    {% empty %}
+      <span class="list-group-item text-muted">{% trans "None" %}</span>
     {% endfor %}
   </ul>
 </div>
--- a/netbox/utilities/views.py
+++ b/netbox/utilities/views.py
@@ -196,7 +196,10 @@ class GetRelatedModelsMixin:
         ]
         related_models.extend(extra)
 
-        return sorted(related_models, key=lambda x: x[0].model._meta.verbose_name.lower())
+        return sorted(
+            filter(lambda qs: qs[0].exists(), related_models),
+            key=lambda qs: qs[0].model._meta.verbose_name.lower(),
+        )
 
 
 class ViewTab:


jeremystretch (approved):
Thanks @alehaa!

PR #18604, opened 2025-02-07
Fixes 18208: Consolidate rendering configuration templates

### LINKED ISSUE #18208: Consolidate logic for rendering configuration templates
### Proposed Changes

Introduce a mixin class and common HTML template to handle rendering configuration templates for both devices and virtual machines.

### Justification

DeviceRenderConfigView and VirtualMachineRenderConfigView are largely redundant. Introducing a common parent mixin and inclusion template will reduce duplicate code.

alehaa:
### Fixes: #18208

The (mostly) redundant `DeviceRenderConfigView` and `VirtualMachineRenderConfigView` have been merged into a single `ObjectRenderConfigView`.

Although a common `object` rendering context variable would be a bit cleaner, the original `device` and `virtualmachine` variables have been preserved. Possible future extensions could add additional values by inheriting the view and overriding the `get_extra_context_data()` method.


jeremystretch (approved):
Thanks for tackling this @alehaa! I've just made some minor tweaks:

- Return a default value from `get_extra_context_data()`
- Moved the VIewTabs to the subclasses (mostly just so their weights are visible relative to their peer views)

PR #18617, opened 2025-02-10
Fixes 18247: Fix dark mode button classes

### LINKED ISSUE #18247: Disabled Cable Trace/Path Buttons in Device Interface View are invisible in Dark Mode
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.8

### Python Version

3.10

### Steps to Reproduce

1. Create a **Device** with One (or more) **Physical Interfaces**
2. Navigate to the **Interfaces** view
3. Toggle between Light and Dark Mode

### Expected Behavior

image

Observe in **Light Mode**, these buttons are visible, but, **disabled**.

### Observed Behavior

image

Observe in **Dark Mode**, the disabled buttons **are functionally invisible**, thus leading to a rendering-like issue.

alehaa:
### Fixes: #18247

Instead of `btn-outline-dark`, `btn-outline-secondary` will be used - just as for pagination.

### DIFF
--- a/netbox/dcim/tables/template_code.py
+++ b/netbox/dcim/tables/template_code.py
@@ -159,8 +159,8 @@ CONSOLEPORT_BUTTONS = """
         </span>
     {% endif %}
 {% elif perms.dcim.add_cable %}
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @alehaa!

PR #18631, opened 2025-02-12
Fixes: #17796 Custom Field Choices -> Create & Add Another causes IndexError

### LINKED ISSUE #17796: Custom Field Choices -> Create & Add Another causes IndexError
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.4

### Python Version

3.12

### Steps to Reproduce

1. In Custom Field Choice Sets choose + Add
2. Provide name: test and add extra choice test
3. Click Create & Add Another

### Expected Behavior

Back to Add a new custom field choice set page.

### Observed Behavior

<class 'IndexError'>

string index out of range

Python version: 3.12.3
NetBox version: 4.1.4
Plugins: 
  netbox_prometheus_sd: 0.6

renatoalmeidaoliveira:
### Fixes:  #17796 Custom Field Choices -> Create & Add Another causes IndexError

* In the current release this issue doesn't raises an Exception, but setups a wrong vaule to the form after clicking in Create & Add Another
* The fix consists in convert the array query into a string to fit the Form field that is an CharField
* It should be possible to change the behaiviour of prepare_cloned_fields, but IMO that may lead to unwanted collateral effects, and its leads to a more complex solution

### DIFF
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -178,12 +178,15 @@ class CustomFieldChoiceSetForm(forms.ModelForm):
     def __init__(self, *args, initial=None, **kwargs):
         super().__init__(*args, initial=initial, **kwargs)
 
-        # Escape colons in extra_choices
+        # Convert extra_choices Array Field from model to CharField for form
         if 'extra_choices' in self.initial and self.initial['extra_choices']:
-            choices = []
-            for choice in self.initial['extra_choices']:
-                choice = (choice[0].replace(':', '\\:'), choice[1].replace(':', '\\:'))
-                choices.append(choice)
+            extra_choices = self.initial['extra_choices']
+            if isinstance(extra_choices, str):
+                extra_choices = [extra_choices]
+            choices = ""
+            for choice in extra_choices:
+                choice_str = ":".join(choice.replace("'", "").replace(" ", "")[1:-1].split(","))
+                choices += choice_str + "\n"
 
             self.initial['extra_choices'] = choices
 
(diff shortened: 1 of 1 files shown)


arthanson (commented):
@renatoalmeidaoliveira This fixes the issue and works, but I'm a bit uncomfortable with the code, it seems like this is patching the symptom rather than the initial problem.  When we do the add-another it looks like it is getting sent in as a stringified tuple and your code here is deconstructing it.  Can we look at why the add-another code is passing it in like this and can it just be made to pass it in the same as the edit case?  I want to make sure we understand why this is different in the fist-place - if it can't be fixed there then can you please add a description in the PR why it is being passed in differently.

renatoalmeidaoliveira:
@arthanson I think the "real" problem relies in the different field type between the form and the model.
The `CustomFieldChoiceSetForm` uses a `CharField` for the `extra_choices` fields [code](https://github.com/netbox-community/netbox/blob/734468fb996cc4b47e37ecf0e01282ea3b74bc0d/netbox/extras/forms/model_forms.py#L165)
While the `CustomFieldChoiceSet` model uses an `ArrayField` type for `extra_choices`  [code](https://github.com/netbox-community/netbox/blob/ed79e3bbf4617ec0764c6e7bd1f289d7abb22106/netbox/extras/models/customfields.py#L766C21-L766C31)
Because of the type difference there must be some kind of patch that convert those types.

arthanson:
@renatoalmeidaoliveira can you add a TODO comment in the code stating what the fix is for and your note re the CharField vs ArrayField, mention that if they are refactored to use the same type then the fix could be removed.

arthanson (approved):

PR #18634, opened 2025-02-12
Fixes #18594: asn_count sort in Sites list

### LINKED ISSUE #18594: Cannot Sort By asn_count Column For Sites Table And Providers View Table
### Deployment Type

Self-hosted

### NetBox Version

V4.2.0

### Python Version

3.10

### Steps to Reproduce

1) Add the ‘ASN Count’ Column as an available/default column for the Sites table
2) Click the ‘ASN Count’ column to sort the table by ASN Count
3) Refresh the page

When doing this, I get the error:
`<class 'django.core.exceptions.FieldError'>

Cannot resolve keyword 'count' into field. Choices are: asn, bookmarks, comments, created, custom_field_data, description, device_count, id, journal_entries, last_updated, providers, rir, rir_id, sites, subscriptions, tagged_items, tags, tenant, tenant_id
`

### Expected Behavior

It would sort the table based upon ASN Count value

### Observed Behavior

It fails and can’t load the Sites Table. I also tested and verified this is an issue that occurs with the online demo system (demo.netbox.dev), so it isnt a customization or something special with my setup that is causing this.

jnovinger:
### Fixes: #18594 - Site list can now be sorted by the `asn_count` column

Changes:
- adds `asn_count` annotation
- removes now unnecessary `accessor` argument for `asn_count` column

### DIFF
--- a/netbox/circuits/tables/providers.py
+++ b/netbox/circuits/tables/providers.py
@@ -33,7 +33,6 @@ class ProviderTable(ContactsColumnMixin, NetBoxTable):
         verbose_name=_('ASNs')
     )
     asn_count = columns.LinkedCountColumn(
-        accessor=tables.A('asns__count'),
         viewname='ipam:asn_list',
         url_params={'provider_id': 'pk'},
         verbose_name=_('ASN Count')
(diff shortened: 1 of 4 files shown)


bctiemann (changes requested):
`ProviderListView` needs this same fix.

jnovinger:
> @jnovinger could you please apply the same fix for the `asn_count` column on `circuits.tables.ProviderTable`?

Fix committed, looking to see what testing opportunities we might have.

jeremystretch (approved):
Extending the test scope would be outside of this particular PR. Thanks @jnovinger!

bctiemann (approved):

PR #18636, opened 2025-02-13
Fixes: #18584 Add rack types column to manufacturers table

### LINKED ISSUE #18584: Manufacturers does not display Rack Types
### NetBox version

v4.2.2

### Feature type

Data model extension

### Proposed functionality

Manufacturers screen lists Device Types and Module Types but does not display Rack Types. It would be very helpful to have an option to include Rack Types from the “Configure Table” feature in the Manufacturers view.

This enhancement would make it easier to manage and reference all equipment types from one place, improving overall workflow and consistency.

### Use case

Currently, users can assign manufacturers when creating rack types; however, there isn’t a straightforward way to view or filter these rack types by manufacturer within the interface.

It would be very helpful if NetBox provided a column by RackType from the manufacturer’s page. This improvement would:

Enhance Usability – Quickly locate all rack types associated with a specific manufacturer.
Streamline Workflows – Reduce the time spent searching for or cross-referencing rack types.

Thank you for considering this request

### Database changes

No DB Change Require.  Presentation request only

### External dependencies

No dependencies.  Presentation request only

renatoalmeidaoliveira:
### Fixes: #18584 Add rack types column to manufacturers table

* Add Add racktype_count annotation to list view queryset
* create the LinkedCountColumn in ManufacturerTable

The result is the following image
image

### Aditional Thougts:

* The issue makes the following statement: 

> Currently, users can assign manufacturers when creating rack types; however, there isn’t a straightforward way to view or filter these rack types by manufacturer within the interface.

* IMO only adding the `Rack Types` Collumn in the `Manufacturer` List View isn't enougth to solve the user issue.
* Looking at [RackTypeFilterForm](https://github.com/netbox-community/netbox/blob/f8022040b245790796a82b38a67570335a320b74/netbox/dcim/forms/filtersets.py#L302) code the `manufacturer_id` field are defined but since it isn't referenced in any FieldSet the user are unable to select a `Manufacturer` to the filter.
* That's a really quick fix but sligtly outside the issue scope, so let me  know if I should add the `manufacturer_id` field inside some filterset
* The  `Rack Type`  FilterSet seens the best option.

### DIFF
--- a/netbox/dcim/forms/filtersets.py
+++ b/netbox/dcim/forms/filtersets.py
@@ -303,7 +303,7 @@ class RackTypeFilterForm(RackBaseFilterForm):
     model = RackType
     fieldsets = (
         FieldSet('q', 'filter_id', 'tag'),
-        FieldSet('form_factor', 'width', 'u_height', name=_('Rack Type')),
+        FieldSet('manufacturer_id', 'form_factor', 'width', 'u_height', name=_('Rack Type')),
         FieldSet('starting_unit', 'desc_units', name=_('Numbering')),
         FieldSet('weight', 'max_weight', 'weight_unit', name=_('Weight')),
     )
--- a/netbox/dcim/tables/devicetypes.py
+++ b/netbox/dcim/tables/devicetypes.py
@@ -31,6 +31,11 @@ class ManufacturerTable(ContactsColumnMixin, NetBoxTable):
         verbose_name=_('Name'),
         linkify=True
     )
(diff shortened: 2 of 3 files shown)


jeremystretch (changes requested):
@renatoalmeidaoliveira agreed; please add the manufacturer field to RackTypeFilterForm under the "Rack Type" fieldset.

jeremystretch (approved):

PR #18642, opened 2025-02-13
Fixes 18286: Replace DurationChoices by JobIntervalChoices

### LINKED ISSUE #18286: Replace DurationChoices by JobIntervalChoices
### NetBox version

v4.2-beta1

### Feature type

### Triage priority

### Proposed functionality

During the implementation of #17716, I added the `JobIntervalChoices` to simplify job scheduling. At the time of writing this PR, I wasn't aware that there was already a `DurationChoices` in the extras module used by the script form.

I suggest replacing it with the new `JobIntervalChoices` and also integrating its missing choices for "12 hours" and "30 days".

### Use case

Code deduplication and increasing maintainability.

### Database changes

None.

### External dependencies

`DurationChoices` has never been part of a public API, so I think there are no side effects.

alehaa:
### Fixes: #18286

`DurationChoices` has been replaced by `JobIntervalChoices`. Choices not often used by jobs, but available in the UI, have been merged into `CHOICES` only, without adding a new choice preset variable.

Technically, this is not an API change, just housekeeping. However, #18286 is tagged as "needs milestone", so it can be moved to the feature branch instead if needed.

### DIFF
--- a/netbox/core/choices.py
+++ b/netbox/core/choices.py
@@ -81,8 +81,10 @@ class JobIntervalChoices(ChoiceSet):
     CHOICES = (
         (INTERVAL_MINUTELY, _('Minutely')),
         (INTERVAL_HOURLY, _('Hourly')),
+        (INTERVAL_HOURLY * 12, _('12 hours')),
         (INTERVAL_DAILY, _('Daily')),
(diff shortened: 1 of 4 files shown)


jeremystretch (approved):
Thanks @alehaa!

PR #18643, opened 2025-02-14
Fixes: #17436 {module} variable is not replaced when installing a module that has submodules (modules bays) in a device

### LINKED ISSUE #17436: {module} variable is not replaced when installing a module that has submodules (modules bays) in a device
### Deployment Type

Self-hosted

### NetBox Version

v4.1.0

### Python Version

3.10

### Steps to Reproduce

1. Create a new device and add a module bay (ex: module bay name = Slot 0, position = 0)
2. 
image

3. Create a new module type and add a component 'module bayf', with name containing the expression {module} (ex: module bay name = Subslot {module}/0, position=0
4. 
image

5. Install this module on the device module bay.
image

### Expected Behavior

The submodule was created with the word {module} replaced by the position value. In the example, the expected name was "Subslot 0/0"

### Observed Behavior

The {module} word is not replaced, and the module bay name is wrong named as "Subslot {module}/0" instead of "Subslot 0/0"

renatoalmeidaoliveira:
### Fixes: #17436 {module} variable is not replaced when installing a module that has submodules (modules bays) in a device

* Update ModuleBay instance name before saving it

### DIFF
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -1298,6 +1298,7 @@ class Module(PrimaryModel, ConfigContextModel):
             else:
                 # ModuleBays must be saved individually for MPTT
                 for instance in create_instances:
+                    instance.name = instance.name.replace(MODULE_TOKEN, str(self.module_bay.position))
                     instance.save()
 
             update_fields = ['module']


jeremystretch (approved):

PR #18650, opened 2025-02-16
Close #18635: Show only the semantic version

### LINKED ISSUE #18635: Adjust netbox-version in /api/status to show version only
### NetBox version

v4.2.3

### Feature type

### Proposed functionality

Currently the `/api/status` endpoint shows a dictionary with the netbox-version key showing the "full" version.  The full version is defined in 

https://github.com/netbox-community/netbox/blob/f8022040b245790796a82b38a67570335a320b74/netbox/utilities/release.py#L37-L41

and includes specific build if needed. This is an issue when some application (in this case the NetBox ansible collection) expects the NetBox version to be the version without build info.

I think we could either move the netbox-version to a netbox-full-version key, or change the netbox-version to use the `netbox.settings.RELEASE.version` as the output.

### Use case

Allow applications to use the netbox-version key for comparing without too much extra parsing.

This issue has popped up as the netbox-docker project has added build info in the local/release.yaml, and the `netbox.settings.RELEASE.version` now looks like '4.2.3-Docker-3.2.0'.

### Database changes

_No response_

### External dependencies

_No response_

tobiasge:
### Fixes: #18635

This modifies the 'netbox-version' to only show the semantic version of NetBox and adds 'netbox-full-version' to show the full version.

Related issues:
- https://github.com/netbox-community/netbox/issues/15908
- https://github.com/netbox-community/ansible_modules/issues/1381

### DIFF
--- a/netbox/netbox/api/views.py
+++ b/netbox/netbox/api/views.py
@@ -4,15 +4,15 @@ from django import __version__ as DJANGO_VERSION
 from django.apps import apps
 from django.conf import settings
 from django_rq.queues import get_connection
-from drf_spectacular.utils import extend_schema
 from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import extend_schema
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
 from rest_framework.views import APIView
 from rq.worker import Worker
 
-from netbox.plugins.utils import get_installed_plugins
 from netbox.api.authentication import IsAuthenticatedOrLoginNotRequired
+from netbox.plugins.utils import get_installed_plugins
 
 
 class APIRootView(APIView):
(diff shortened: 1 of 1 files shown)


sc68cal (approved):

daftu (commented):
The changes do what they were intended to

sc68cal:
This is causing issues with how the Ansible modules for NetBox manages API compatibility between NetBox server versions. It would be great to find a way forward where we get a version string that we have been using for our logic, and this "fuller" version string co-exists.

sc68cal (approved):

bctiemann:
This likely needs a minor release milestone due to breaking API chages.

tobiasge:
I could retarget this to feature, but I would argue, that the breaking change when https://github.com/netbox-community/netbox/issues/15908 was merged. This PR will only change the `version` field back to it's original content.

bctiemann:
@tobiasge Regardless of how the behavior changed in the past (perhaps wrongly), this change does introduce a new change in behavior and so I think this should be retargeted for `feature`.

tobiasge:
I have retargeted this on `feature`.

bctiemann (approved):

PR #18662, opened 2025-02-18
Fixes #18619: shift-select selects hidden items

### LINKED ISSUE #18619: Shift-click selection selects hidden interfaces
### Deployment Type

Self-hosted

### NetBox Version

v4.2.3

### Python Version

3.12

### Steps to Reproduce

1. Create Device.
2. Create physical interface "et-0/0/0".
3. Create virtual interface "et-0/0/0.0".
4. Create physical interface "et-0/0/1".
5. Enable "Hide virtual interfaces".
6. Toggle selection check boxes of physical interfaces using Shift-click.
7. Show virtual interfaces again.

### Expected Behavior

Only the physical interfaces should be selected. This would be consistent with the usage of the "Select all" check box.

### Observed Behavior

The virtual interfaces "between" the selected physical interfaces are also selected.

jnovinger:
### Fixes #18619: shift-select selects hidden items

This also fixes the inverse, when a range is unselected via shift-click, previously checked checkboxes that are hidden are not changed.

Changes:
- adds tests to only change checkbox `checked` values if their parent `tr` is not hidden


jeremystretch (approved):
Thanks @jnovinger!

PR #18664, opened 2025-02-18
Fixes 18277: Cover multitable inheritance in serialization

### LINKED ISSUE #18277: Cover multi-table inheritance in change logged models
### NetBox version

v4.1.10

### Feature type

### Triage priority

### Proposed functionality

Currently the `serialize_object()` utility function checks if the model has the `custom_field_data` attribute.

https://github.com/netbox-community/netbox/blob/f03489f58ee6452f9b0c60d84364f0c67c96b41e/netbox/utilities/serialization.py#L32-L33

I suggest changing this to check if the key actually exists in the `data` dictionary.

### Use case

A plugin could extend existing NetBox models using multi-table inheritance. However, although the attribute is available on the child model class, the data is actually stored in the parent model. Serializing the object during the automatic creation of the `ObjectChange` will therefore raise an exception, as the attribute is available, but won't be serialized in the `data` dictionary.

Although NetBox should not cover serialization for this scenario, changing this would allow plugins more possibilities. NetBox simply wouldn't handle this attribute in this scenario, leaving it to the developer to implement. No exceptions will be thrown either.

### Database changes

None

### External dependencies

None

alehaa:
### Fixes: #18277

During serialization, custom fields may be available to a model due to multi-table inheritance, but might not be available in serialized data because only direct fields of the model are covered. Now this attribute is only used if available in serialized data. Models using multi-table inheritance must modify their serialize_object() method to cover parent serialization.

#### Example 

Tested example for child models:

```Python
class ChildModel(Tenant):
    def serialize_object(self, exclude=None) -> dict[str, Any]:
        data = self.tenant_ptr.serialize_object(exclude)
        data.update(super().serialize_object(exclude))
        return data
```

If not defined, only fields of the child are serialized.

### DIFF
--- a/netbox/utilities/serialization.py
+++ b/netbox/utilities/serialization.py
@@ -29,7 +29,7 @@ def serialize_object(obj, resolve_tags=True, extra=None, exclude=None):
     exclude = exclude or []
 
     # Include custom_field_data as "custom_fields"
-    if hasattr(obj, 'custom_field_data'):
+    if 'custom_field_data' in data:
         data['custom_fields'] = data.pop('custom_field_data')
 
     # Resolve any assigned tags to their names. Check for tags cached on the instance;


jeremystretch (approved):
Thanks @alehaa!

PR #18679, opened 2025-02-19
Closes #17793: Introduce a REST API endpoint for tagged objects

### LINKED ISSUE #17793: Expose tagged objects through REST api
### NetBox version

v4.1.1

### Feature type

Data model extension

### Triage priority

This is a very minor change

### Proposed functionality

Add a new endpoint to get the list of the objects using a specific tag.

### Use case

Looking up objects by their tags through the api currently requires numerous api calls to search through all object types. The underlying functionality exists as demonstrated by visiting the tag's display_url and looking under "Tagged Objects".

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #17793

- Introduce a REST API view & serializer for TaggedItem
- Add TaggedItemFilterSet
- Add relevant tests


arthanson (changes requested):
I'm a bit confused on the expected output here - The screenshot shows tag id #2 "Bravo" in the UI - It has 2 tagged objects.  I would expect that if I do the REST API for `http://127.0.0.1:8000/api/extras/tagged-objects/2/` that I would get back a list of two objects (Site: "DM-Akron" and Site: "JBB Branch 127") - but I get back Site "Butler Communication" for Tag "Lima".

Am I understanding this wrong?
Monosnap Bravo | NetBox 2025-02-24 16-58-10
Monosnap NetBox REST API 2025-02-24 16-59-28

jeremystretch:
> I would expect that if I do the REST API for `http://127.0.0.1:8000/api/extras/tagged-objects/2/` that I would get back a list of two objects (Site: "DM-Akron" and Site: "JBB Branch 127") - but I get back Site "Butler Communication" for Tag "Lima".

The REST API endpoint queries TaggedItem records; this is a mapping between a Tag and an arbitrary object (e.g. site, device, etc.). So querying for `/api/extras/tagged-objects/2/` returns the TaggedItem with ID 2.

I'd expect that typically users are not concerned with retrieving specific mappings, but will instead filter the list by tag (`?tag_id=2`) or by object type & ID (`?object_type=dcim.site&object_id=123`). However, we want to keep the detail view for the endpoint consistent with the rest of the REST API.

(Your comment did prompt me to notice, however, that I had mistakenly omitted the `object_id` field from the filter set.)

arthanson (approved):

Issue #18423, opened 2025-01-17
Replace `STORAGE_BACKEND` and `STORAGE_CONFIG` configuration parameters with Django's `STORAGES`


jeremystretch:
### NetBox version

v4.2.1

### Feature type

Change to existing functionality

### Triage priority

N/A

### Proposed functionality

NetBox currently exposes two configuration parameters for configuring file storage:

* `STORAGE_BACKEND` - Sets the storage backend (e.g. `django-storages`)
* `STORAGE_CONFIG` - A dictionary of backend-specific configuration parameters

We can replace these by instead exposing Django's [`STORAGES`](https://docs.djangoproject.com/en/5.1/ref/settings/#storages) setting directly.

### Use case

Django's native `STORAGE_BACKEND` setting was [removed in v5.1](https://docs.djangoproject.com/en/5.1/releases/5.1/#features-removed-in-5-1) and superseded by `STORAGES`. There is no longer a need to handle the backend selection and configuration via separate configuration parameters.

### Database changes

N/A

### External dependencies

N/A

RangerRick:
I've got the NBE build set up such that I can now test this on our end, and as I worked to reenable S3 storage on my side, I remembered that one of the blockers last time around was the behavior of the storages and what gets persisted to the database.

The short version is: the entire S3 URL is stored to the database, which when inside a cluster is not reachable from the outside. Setting `AWS_S3_CUSTOM_DOMAIN` is not always feasible, because it would require introspection of the ingress to know what the external hostname is, and doing that reliably is problematic.

In my previous S3 PoC, I wrote a façade storage driver that stored everything relative, but then realized I didn't have enough control over how the image URLs are used elsewhere in the system so they would still not render right if, say, you have NetBox behind a proxy using a prefix, so I was still forced to hardcode the ingress hostname to make it all work.

It feels like handling this well is more complicated than just dealing with prefixes; the current S3 storage behavior assumes your S3 host never changes.

PR #18680, opened 2025-02-19
18896 Replace STORAGE_BACKEND with STORAGES and support Script running from S3

arthanson:
### Fixes: #18896 (and #18423) 

This Allows scripts to be run from S3 as well as config changes to allow for setting it.  The changes are in two basic parts - the Script Form is changed to upload files using Django-storages (either FileSystem or S3) and the Script module loader is changed to pull from Django-storages.

* Django-storages is now a requirement
* STORAGE_BACKEND is now deprecated
* STORAGES is now set directly as a config option
* write_to_disk was moved from DataFile to ManagedFile as it is only ever used there or in classes derived from it.
* The changes in BaseScript find_source is a replacement for the inspect module to actually find and load the modules.
* extras.storage.ScriptFileSystemStorage is required for backwards compatibility with SCRIPTS_ROOT setting as otherwise with normal FileSystemStorage it will cause a path permission error going outside media root.

**Note:** the Script convenience methods `load_yaml` and `load_json` are not compatible when using S3 as they go direct to the local file system and are based on the path of the Script so wouldn't be backwards compatible.  They are convenience functions so the script author could write them directly.

RELEASE_NOTES should probably have a section about replacing STORAGE_BACKEND and STORAGE_CONFIG with STORAGES for those updating from a previous version.


netbox/core/models/files.py
```diff
@@ -104,8 +125,9 @@ def clean(self):
 
     def delete(self, *args, **kwargs):
         # Delete file from disk
+        storage = self.get_storage()
         try:
-            os.remove(self.full_path)
+            storage.delete(self.full_path)
```
jeremystretch, line 130: I think there's a bug in the path resolution preventing files from deleted from the S3 bucket when deleting a script module. In my testing, `full_path` returns the filename prefixed with `scripts/` twice.

jeremystretch (changes requested):
Looking good! Found a few things I think we can clean up.

jeremystretch:
@arthanson could you submit a retroactive FR for this work please? The scope of #18423 was just the configuration parameter change. The changes here have a far broader impact (such as introducing `django-storages` as a new mandatory dependency) that need to be captured.

netbox/extras/scripts.py
```diff
@@ -367,9 +369,46 @@ def scheduling_enabled(self):
     def filename(self):
         return inspect.getfile(self.__class__)
 
+    def findsource(self, object):
+        storage = storages.create_storage(storages.backends["scripts"])
+        with storage.open(os.path.basename(self.filename), 'r') as f:
```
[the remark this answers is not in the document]
arthanson: that's interesting/ weird, it worked with S3 when I had it configured but doesn't work with digital ocean...

netbox/netbox/settings.py
```diff
@@ -233,6 +235,21 @@
...
+    if not CONFIG_STORAGES:
+        raise ImproperlyConfigured(
+            "STORAGE_BACKEND and STORAGES are both set, remove the deprecated STORAGE_BACKEND setting."
+        )
+    else:
+        warnings.warn(
+            "STORAGE_BACKEND is deprecated, use the new STORAGES setting instead."
+        )
+
+if STORAGE_CONFIG is not None:
```
jeremystretch, line 248: `STORAGE_CONFIG` defaults to an empty dict if not defined, which triggers the warning. We should set it to None by default.

netbox/netbox/settings.py
```diff
@@ -74,6 +75,7 @@
 BASE_PATH = trailing_slash(getattr(configuration, 'BASE_PATH', ''))
 CHANGELOG_SKIP_EMPTY_CHANGES = getattr(configuration, 'CHANGELOG_SKIP_EMPTY_CHANGES', True)
 CENSUS_REPORTING_ENABLED = getattr(configuration, 'CENSUS_REPORTING_ENABLED', True)
+CONFIG_STORAGES = getattr(configuration, 'STORAGES', {})
```
jeremystretch, line 78: I'm not a fan of altering constant names from their config parameter names; it seems like it's begging for confusion (_especially_ when there's also a `STORAGE_CONFIG` parameter).

IMO we should name this `STORAGES` and default it to an empty dict. The `STORAGES` constant currently defined on line 254 can be renamed to `STORAGES_DEFAULT`, to which `STORAGES` can be applied:

```python
STORAGES = getattr(configuration, 'STORAGES', {})

STORAGES_DEFAULT = {
    "default": {
        "BACKEND": "django.core.files.storage.FileSystemStorage",
    },
    "staticfiles": {
        "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
    },
    "scripts": {
        "BACKEND": "extras.storage.ScriptFileSystemStorage",
    },
}

# Apply custom storage configuration to default
STORAGES = DEFAULT_STORAGES | STORAGES
```

netbox/netbox/settings.py
```diff
@@ -223,60 +225,58 @@
...
-        "STORAGE_CONFIG has been set in configuration.py but STORAGE_BACKEND is not defined. STORAGE_CONFIG will be "
-        "ignored."
-    )
+# Monkey-patch django-storages to fetch settings from STORAGE_CONFIG
+if STORAGE_CONFIG is not None:
+    def _setting(name, default=None):
+        if name in STORAGE_CONFIG:
+            return STORAGE_CONFIG[name]
+        return globals().get(name, default)
+    storages.utils.setting = _setting
```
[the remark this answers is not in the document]
jeremystretch: Any thoughts on this? Seems like it would be pretty easy to knock out. I can take it if you'd like.

docs/customization/custom-scripts.md
```diff
@@ -140,6 +140,8 @@ The Script class provides two convenience methods for reading data from files:
 
 These two methods will load data in YAML or JSON format, respectively, from files within the local path (i.e. `SCRIPTS_ROOT`).
 
+**Note:** These convenience methods only work if running scripts within the local path, they will not work if running scripts using AWS S3.
```
jeremystretch, line 143: Is this limitation unique to S3, or does it apply to _any_ external storage?

jeremystretch (changes requested):

netbox/netbox/settings.py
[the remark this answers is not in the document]
arthanson: I'm not sure on this one, STORAGE_CONFIG can be any group of values, I tried:
```
if STORAGE_CONFIG is not None:
    if 'OPTIONS' not in STORAGES:
        STORAGES['OPTIONS'] = {}

    for key, value in STORAGE_CONFIG.items():
        STORAGES['OPTIONS'][key] = value
```
But that doesn't work. Not sure exactly what you meant here?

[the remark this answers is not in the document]
jeremystretch: I mean that what was previously configured as
```
STORAGE_CONFIG = {
    'AWS_ACCESS_KEY_ID': 'key-id',
    'AWS_SECRET_ACCESS_KEY': 'secret-key',
    'AWS_STORAGE_BUCKET_NAME': 'storages-test',
    'AWS_S3_ENDPOINT_URL': 'https://nyc3.digitaloceanspaces.com',
    'AWS_S3_REGION_NAME': 'nyc3',
}
```
can instead be configured as
```
STORAGES = {
    ...
    "scripts": {
        "BACKEND": "storages.backends.s3boto3.S3Boto3Storage",
        "OPTIONS": {
            'access_key': 'key-id',
            'secret_key': 'secret-key',
            'bucket_name': 'storages-test',
            'endpoint_url': 'https://nyc3.digitaloceanspaces.com',
            'region_name': 'nyc3',
        }
    },
}
```
If declared, we can adapt `STORAGE_CONFIG` to provide backward compatibility by replacing the global variables with their [corresponding backend settings](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#authentication-settings). Granted there are quite a few of these parameters that would need to be mapped; it's probably better tackled under a separate FR.

jeremystretch (changes requested):

netbox/netbox/settings.py
arthanson: I'm not getting the warning if I don't have it defined.

netbox/netbox/settings.py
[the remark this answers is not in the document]
arthanson: we can add a separate FR, but IMHO it's deprecated and the existing code we know works for backward compatibility so it seems strange to be creating new code that is just going to get removed next release? Also it would be a very large list if they are using other storage backends which is possible - we would have to put in every config parameter for S3 and SWIFT at a minimum and probably all the other providers for Django-storages to be safe.

netbox/netbox/settings.py
jeremystretch: Looks like this was fixed in 4552cea

jeremystretch (dismissed):
Thanks @arthanson!

jeremystretch:
This still needs a corresponding FR which captures the full extent of changes made in the PR. #18423 is not sufficient for this in its current form.

arthanson:
> This still needs a corresponding FR which captures the full extent of changes made in the PR. #18423 is not sufficient for this in its current form.

Expanded the issue with more details.

docs/customization/custom-scripts.md
arthanson: As per discussion this will be deprecated.

arthanson:
Created issue #18896

jeremystretch (approved):

PR #18681, opened 2025-02-19
Closes #18627: Proxy routing

### LINKED ISSUE #18627: Granular proxy configuration
### NetBox version

v4.2.3

### Feature type

### Proposed functionality

Currently proxies are configured globally in Netbox. We would like the ability to specify proxies individually for each component that makes external requests:

- Authentication
- Remote datasources
- Webhooks
- News feed
- Plugins API
- Release checks

A proposed approach would offload the proxy decision making process to a platform provided function.

in `settings.py`

```
PROXY_ROUTERS = getattr(configuration, 'PROXY_ROUTERS', [])
```

in `configuration.py`

```
# Default
PROXY_ROUTERS = []

# Platform specify they want to handle proxies returned to calling functions
PROXY_ROUTERS = ["common.proxy.ProxyRouter"]
```
in `common/proxy.py`

```
class ProxyRouter:
    
    def route(self, module: str, url: Union[str,None], protocol: Union[str,None]) -> Optional[str]:
        # every router class must implement a route() method        

        # internal platform logic to decide how to handle each request
        # based on module, protocol and/or url
        # return "http://my.lovely.proxy.server:8080"
        return None
```

When making requests loop through PROXY_ROUTERS passing in the module name, protocol and url if known, use the result from the first function that does not return None

Helper functions in core:

```
def proxy_handler(module: str, url: Union[str,None], protocol: Union[str,None]) -> Optional[str]:
    # any proxy routers defined?
    if len(settings.PROXY_ROUTERS) > 0:
        # yes, delegate routing decisions to the provided function
        for router in settings.PROXY_ROUTERS:
            proxy_server = router.route(module=module, url=url, protocol=protocol)
            if proxy_server is not None:
                return proxy_server
            # no match, try the next one
         # provided routers did not match, default to no proxy
         return None
     if protocol in settings.HTTP_PROXIES:
         return settings.HTTP_PROXIES
     return None
```    

e.g.  with Sentry… https://github.com/netbox-community/netbox/blob/b91366129783d34b474d36e5dbf65e0d09016983/netbox/netbox/settings.py#L573-L580

```
    sentry_sdk.init(
        dsn=SENTRY_DSN,
        release=RELEASE.full_version,
        sample_rate=SENTRY_SAMPLE_RATE,
        traces_sample_rate=SENTRY_TRACES_SAMPLE_RATE,
        send_default_pii=SENTRY_SEND_DEFAULT_PII,
        http_proxy=HTTP_PROXIES.get('http') if HTTP_PROXIES else None,
        https_proxy=HTTP_PROXIES.get('https') if HTTP_PROXIES else None
    )
```
```
    http_proxy=proxy_handler(module="sentry", protocol="http"),
    https_proxy=proxy_handler(module="sentry", protocol="https")
)
```

e.g. webhook dispatch https://github.com/netbox-community/netbox/blob/b91366129783d34b474d36e5dbf65e0d09016983/netbox/extras/webhooks.py#L86-L91

    # Send the request
    with requests.Session() as session:
        session.verify = webhook.ssl_verification
        if webhook.ca_file_path:
            session.verify = webhook.ca_file_path
        response = session.send(prepared_request, proxies=settings.HTTP_PROXIES)

becomes

```
    # Send the request
    with requests.Session() as session:
        session.verify = webhook.ssl_verification
        if webhook.ca_file_path:
            session.verify = webhook.ca_file_path
        response = session.send(prepared_request, proxies=proxy_handler(module="webhook", url=params["url"])
```

census requests https://github.com/netbox-community/netbox/blob/b91366129783d34b474d36e5dbf65e0d09016983/netbox/core/jobs.py#L70C1-L75C14

```
            requests.get(
                url=settings.CENSUS_URL,
                params=census_data,
                timeout=3,
                proxies=settings.HTTP_PROXIES
            )
```

becomes

```
            requests.get(
                url=settings.CENSUS_URL,
                params=census_data,
                timeout=3,
                proxies=proxy_handler(module="census", url=settings.CENSUS_URL)
            )
```

other modules would be:

- `newsfeed`
- `plugins_catalog`
- `housekeeping_releasecheck`
- `datasource`

each of these would hand off proxy decision making to proxy_handler 

### Use case

- Force only webhook traffic via a SOCKS proxy into a connectivity customer VPC and out through a DX/tunnel.
- Force both webhook and remote datasources via a proxy, but all other traffic goes another route.

### Database changes

None identified

### External dependencies

None identified

jeremystretch:
### Closes: #18627

- Introduce DefaultProxyRouter and use it as the default proxy router
- Allow admins to override default routing via the new `PROXY_ROUTERS` config parameter
- Use new `resolve_proxies()` utility function as a wrapper for resolving proxies

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -64,7 +64,7 @@ Email is sent from NetBox only for critical events or if configured for [logging
 
 ## HTTP_PROXIES
 
-Default: None
+Default: Empty
 
 A dictionary of HTTP proxies to use for outbound requests originating from NetBox (e.g. when sending webhook requests). Proxies should be specified by schema (HTTP and HTTPS) as per the [Python requests library documentation](https://requests.readthedocs.io/en/latest/user/advanced/#proxies). For example:
 
@@ -74,6 +74,8 @@ HTTP_PROXIES = {
     'https': 'http://10.10.1.10:1080',
 }
 ```
+
+If more flexibility is needed in determining which proxy to use for a given request, consider implementing one or more custom proxy routers via the [`PROXY_ROUTERS`](#proxy_routers) parameter.
 
 ---
 
@@ -157,6 +159,16 @@ LOGGING = {
 Default: `$INSTALL_ROOT/netbox/media/`
 
 The file path to the location where media files (such as image attachments) are stored. By default, this is the `netbox/media/` directory within the base NetBox installation path.
+
+---
+
+## PROXY_ROUTERS
+
+Default: `["utilities.proxy.DefaultProxyRouter"]`
+
+A list of Python classes responsible for determining which proxy server(s) to use for outbound HTTP requests. Each item in the list can be the class itself or the dotted path to the class.
+
+The `route()` method on each class must return a dictionary of candidate proxies arranged by protocol (e.g. `http` and/or `https`), or None if no viable proxy can be determined. The default class, `DefaultProxyRouter`, simply returns the content of [`HTTP_PROXIES`](#http_proxies).
 
 ---
 
--- a/netbox/core/data_backends.py
+++ b/netbox/core/data_backends.py
@@ -7,13 +7,13 @@ from pathlib import Path
 from urllib.parse import urlparse
 
 from django import forms
-from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import gettext as _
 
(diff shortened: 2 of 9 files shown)


bctiemann (approved):
Tested with the sample `HTTP_PROXIES` settings in configuration.py, works as expected using the default resolver.

PR #18685, opened 2025-02-20
Fixes #18674: Fix form reset when selecting a value from a speed selection dropdown

### LINKED ISSUE #18674: Selecting speed on add Circuit Termination causes form to be wiped
### Deployment Type

Self-hosted

### NetBox Version

v4.2.3

### Python Version

3.12

### Steps to Reproduce

1. Create a new circuit, fill in any values
2. Click Add for Termination A (or Termination Z)
3. Select Termination Type > DCIM > Site
4. Select Site (any)
5. Click the down arrow next to Port speed and select a value
6. Form is refreshed and previous values are blanked out (including Port speed)

### Expected Behavior

The form should not erase previously input values.

### Observed Behavior

Form is reset to default/blank.  This happens in the NetBox demo.

jeremystretch:
### Fixes: #18674

`initForms()` must be called after re-initializing the page following an HTMX request.


arthanson (changes requested):
This fixes the issue and the code is fine but the build is failing from the static assets diff, maybe from the node-version update?

jeremystretch:
Just had to run `yarn install` locally to resolve the discrepancy.

arthanson (approved):

PR #18690, opened 2025-02-20
18296 Add Tenancy to VLAN Groups

arthanson:


netbox/ipam/models/vlans.py
```diff
@@ -65,6 +65,13 @@ class VLANGroup(OrganizationalModel):
     _total_vlan_ids = models.PositiveBigIntegerField(
         default=VLAN_VID_MAX - VLAN_VID_MIN + 1
     )
+    tenant = models.ForeignKey(
```
jeremystretch, line 68: It would be cleaned to add this before `_total_vlan_ids` IMO.

netbox/ipam/tables/vlans.py
```diff
@@ -65,9 +65,11 @@ class Meta(NetBoxTable.Meta):
         model = VLANGroup
         fields = (
             'pk', 'id', 'name', 'scope_type', 'scope', 'vid_ranges_list', 'vlan_count', 'slug', 'description',
-            'tags', 'created', 'last_updated', 'actions', 'utilization',
+            'tenant', 'tenant_group', 'tags', 'created', 'last_updated', 'actions', 'utilization',
+        )
+        default_columns = (
+            'pk', 'name', 'scope_type', 'scope', 'vlan_count', 'utilization', 'tenant', 'tenant_group', 'description'
```
jeremystretch, line 71: Suggest we omit `tenant_group` from the default column set (which I think is consistent with other tenant-enabled models).
```suggestion
            'pk', 'name', 'scope_type', 'scope', 'vlan_count', 'utilization', 'tenant', 'description',
```

netbox/ipam/tables/vlans.py
```diff
@@ -65,9 +65,11 @@ class Meta(NetBoxTable.Meta):
         model = VLANGroup
         fields = (
             'pk', 'id', 'name', 'scope_type', 'scope', 'vid_ranges_list', 'vlan_count', 'slug', 'description',
-            'tags', 'created', 'last_updated', 'actions', 'utilization',
+            'tenant', 'tenant_group', 'tags', 'created', 'last_updated', 'actions', 'utilization',
```
jeremystretch, line 68: VLANGroupTable needs to inherit TenancyColumnsMixin to ensure the `tenant` and `tenant_group` columns render correctly as hyperlinks.

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @arthanson!

Issue #17688, opened 2024-10-07
GraphQL filters (AND, OR and NOT) don't work for custom filterset fields


arthanson:
### Deployment Type

Self-hosted

### NetBox Version

v4.1.3

### Python Version

3.10

### Steps to Reproduce

Using a GraphQL filter with AND, OR, NOT for a field that has custom implementation in the filterset (or only appears in the filterset) for example asn_id on Site. Doesn't work

1. Create 4 sites with ID's 1 to 4
2. Create 4 ASNs and assign each to a single Site (1-4)
3. Use the following GraphQL query
```
{
  site_list(filters: {asn_id: "1", OR: {asn_id: "4"}}) {
    name
    asns {
      id
    }
  }
}
```  

### Expected Behavior

Will get a list of 2 sites.

### Observed Behavior

Get an empty list.

arthanson:
Copying this from issue #16024 as that was resolved for the specific case of the ChoiceField, but there is still and underling issue with Strawberry and django-filters interaction.

The basic issue is strawberry filtering uses Q objects, django-filters returns filtersets which makes them incompatible.  None of the solutions I see look very good and all would take a lot of work.

Graphene interfaces with Django-filter and uses filtersets.  Code is at https://github.com/graphql-python/graphene-django/tree/main/graphene_django/filter.  About 8 files of code.

Strawberry is all based off of Q objects.  Code is at https://github.com/strawberry-graphql/strawberry-django/blob/main/strawberry_django/filters.py. Strawberry documentation on filters is at: https://strawberry.rocks/docs/django/guide/filters

From what I can see, this is also tied into the GraphQL parsing framework, so if we tried to use Django-filter we would also probably have to make patches to that code as well.

We are also using  Strawberry Legacy Filtering which needs to be removed (https://strawberry.rocks/docs/django/guide/filters#legacy-filtering) this would need changes to the auto type decorator (https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/graphql/filter_mixins.py#L131).  

I can think of several different potential solutions (Option 4 might potentially be the best option?)

1. Basically replicate filter handling in Django-graphene - override strawberry_django.process_filters to work with django-filters basically replacing strawberry's filter handling with a new one that is compatible with Django-filters.

    * **pros:** No changes to existing NetBox filter code.  Plugin filtering / GraphQL code would remain compatible.
[...]

Issue #17681, opened 2024-10-06
Cannot filter ip range based on parent prefix


kuon:
### Deployment Type

Self-hosted

### NetBox Version

v4.1.1

### Python Version

3.12

### Steps to Reproduce

Create the following graphql query:

```
{
  ip_range_list(filters: {parent: {exact: "10.20.0.0/16"}}) {
    id
  }
}
```

### Expected Behavior

I expect the same result as this API call:

`/api/ipam/ip-ranges/?parent=10.20.0.0/16`

And it is mentioned in this issue:

https://github.com/netbox-community/netbox/issues/13313

Also the `IPRangeFilter` graphql type has a `parent: StrFilterLookup` field.

### Observed Behavior

Error is thrown:

```
{
  "data": null,
  "errors": [
    {
      "message": "Cannot resolve keyword 'parent' into field. Choices are: bookmarks, comments, contacts, created, custom_field_data, description, end_address, id, journal_entries, last_updated, mark_utilized, role, role_id, size, start_address, status, subscriptions, tagged_items, tags, tenant, tenant_id, vrf, vrf_id",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "ip_range_list"
      ]
    }
  ]
}
```

tom0010:
Also affects ip address lists too.

svluder:
Just to mention that we hit the same issue in an installation at one of our major customers in Germany.
Hoping that there is a chance that this might increase the priority of fixing this bug.

If there is anything where I can support (keep in mind that I am not a Django developer) please let me know.

loulecrivain:
> Also affects ip address lists too.

Can confirm, we are also affected. Deployment version is 4.1.4.

jeremypng:
Fixed in my branch here:
https://github.com/jeremypng/netbox/tree/refs/heads/graphql-filter-redesign

fixed query:
```graphql
query GetIPRange {
  ip_range_list (filters: {parent: [""192.168.1.0/25""]}){
    id
    start_address
    end_address
  }
}
```
results:
```json
{
  "data": {
    "ip_range_list": [
      {
        "id": "1",
        "start_address": "192.168.1.0/22",
        "end_address": "192.168.1.49/22"
      },
      {
        "id": "2",
        "start_address": "192.168.1.50/22",
        "end_address": "192.168.1.99/22"
      }
    ]
  }
}
```

If you'll assign this to me I'll tag it in the PR

[...]

Issue #16511, opened 2024-06-11
Missing lookup filters for interface object fields


cybarox:
### Deployment Type

Self-hosted

### NetBox Version

v4.0.5

### Python Version

3.11

### Linked Issue
#16261 
#16354 

#### NetBox v4.0.x - Observed Behavior

1. Setup a fresh NetBox v4.0.x installation
2. Import new Site
    ```yaml
    ---
    name: Test Site 
    slug: test_site
    status: active
    facility: Facility 1
    ```
3. Import Manufacturer
    ```yaml
    --- 
    name: Test Manufacturer
    slug: test_manufacturer
    ```
4. Import Device Type
    ```yaml
    ---
    manufacturer: Test Manufacturer
    model: Test Model
    slug: test_model
    u_height: 0
    interfaces:
    - name: eth0
      type: 1000base-t
    - name: eth1
      type: 1000base-t
    ```
5. Import Device Role
    ```yaml
    ---
    name: Test Role
    slug: test_role
    color: ffff00
    ```
6. Import Devices
    ```yaml
    ---
    - name: Test Device 1
      role: Test Role
      manufacturer: Test Manufacturer
      device_type: Test Model
      slug: test_device_1
      site: Test Site
      status: active
    - name: Test Device 2
      role: Test Role
      manufacturer: Test Manufacturer
      device_type: Test Model
      slug: test_device_1
      site: Test Site
      status: active
    ```
7. Import Interfaces
    ```yaml
    ---
    - id: 1
      mac_address: DE:AD:BE:EF:00:69
    - id: 2
      mac_address: 00:00:00:00:00:00
    - id: 3
      mac_address: DE:AD:BE:EF:00:42
    - id: 4
      mac_address: 00:00:00:00:00:01
    ```
8. Use GraphQL Webgui Query
    ##### Query:
    ```graphql
    query vendor_query($mac_vendor: [String!], $facility_name: String!, $device_role: [String!]) {
      site_list(filters: {facility: {exact: $facility_name}}) {
        name
        devices(filters: {role: $device_role}) {
          name
          interfaces(filters: {mac_address: {i_contains: $mac_vendor}}) {
            name
            mac_address
          }
        }
      }
    }
    ```
    ##### Query Variables
    ```json
    {
      "mac_vendor": "DE:AD",
      "facility_name": "Facility 1",
      "device_role": "test_role"
    }
    ```
9. Query Result
    ```json
    {
      "data": null,
      "errors": [
        {
[...]

Issue #16305, opened 2024-05-27
GraphQL filter for "ip_address_list" - Filter query tag works not as accepted


SchmidtJonathan1:
### Deployment Type

Self-hosted

### NetBox Version

v4.0.3

### Python Version

3.12

### Steps to Reproduce

When a query is created in Graphql Explorer that accesses the class "ip_address_list". Cannot be filtered by "tags" as there is only one parameter called "tag". 

```
#Graphql query:

query MyQuery {
  ip_address_list(filters: {tag: "Gateway"}) {
    tags{
      name
    }
  }
}
```

### Expected Behavior

```
{
  "data": {
    "ip_address_list": [
      {
        "id": "15",
        "tags": [
          {
            "name": "Gateway"
          }
        ]
      }
}
```

### Observed Behavior

```
{
  "data": {
    "ip_address_list": [
      {
        "id": "2",
        "tags": []
      },
      {
        "id": "3",
        "tags": []
      },
      {
        "id": "4",
        "tags": []
      },
      {
        "id": "15",
        "tags": [
          {
            "name": "Gateway"
          }
        ]
      }
}
```

fooker:
Hi, I would like to tackle this issue (CC @jeffgdotorg)

arthanson:
@fooker any updates on this?  Are you still able to work on this?

jeremypng:
Fixed in my branch here:
https://github.com/jeremypng/netbox/tree/refs/heads/graphql-filter-redesign

fixed query:
```graphql
query IPTags {
  ip_address_list(filters: {tags: {name: {exact: "Bravo"}}}) {
    id
    address
  }
}
```
results:
```json
{
  "data": {
    "ip_address_list": [
      {
        "id": "34",
        "address": "172.16.0.4/24"
      }
    ]
  }
}
```

jeremystretch:
Blocked by #7598

jeremystretch:
I believe this has been resolved by the work on #7598. For example, the following query will match all sites with to which the tag named "Alpha" has been assigned:

```graphql
query {
  site_list(filters: {tags: {name: {exact: "Alpha"}}}) {
    id
    name
  }
}
```

This change will be implemented in NetBox v4.3.

PR #18701, opened 2025-02-21
Closes #7598: Enable custom field filtering for GraphQL

### LINKED ISSUE #7598: Support for filtering on custom fields

jeremystretch:
### Closes: #7598

This PR serves to capture @jeremypng's recent work on improving the GraphQL API. Although this work primarily targets FR #7598, it includes many foundational improvements:

- Discontinuing the use of deprecated filters (`USE_DEPRECATED_FILTERS = False`)
- Defining standalone filters for GraphQL
- Reusable mixin classes for filters
- Adds the `as_enum()` method to ChoiceSet for use in GraphQL filters


jeremystretch:
We still need to update the documentation to cover:
- The new filtering syntax for GraphQL queries
- Developer instructions for adding & updating GraphQL filters

jeremystretch:
I believe this PR also resolves the following bugs related to GraphQL functionality:

* #16305
* #16511
* #17681
* #17688

jeremypng:
We have continued using this graphql filtering system on 4.0.11 and have reduced nested queries using the DRF API with pynetbox from 36s to 400ms with graphql.

The additional changes you have made look good.

netbox/circuits/graphql/filters.py
```diff
@@ -19,66 +45,165 @@
...
+        strawberry_django.filter_field()
+    )
+    termination_id: ID | None = strawberry_django.filter_field()
+    port_speed: Annotated['IntegerLookup', strawberry.lazy('netbox.graphql.filter_lookups')] | None = (
+        strawberry_django.filter_field()
+    )
+    upstream_speed: Annotated['IntegerLookup', strawberry.lazy('netbox.graphql.filter_lookups')] | None = (
+        strawberry_django.filter_field()
+    )
+    xconnect_id: FilterLookup[str] | None = strawberry_django.filter_field()
```
jeremystretch, line 71: @jeremypng do you know whether it's necessary to define these filters explicitly? Setting them to `auto` seems to yield the same schema, but I wanted to check if you had run into any problems with this approach.
```suggestion
    xconnect_id: strawberry.auto
```

jeremypng: If it can successfully infer the filter type from the field type it will work. From what I saw, this generally works for model fields that are physical database fields but not computed fields.

jeremystretch: That makes sense, thanks!

arthanson:
Adding a couple sample queries here for testing purposes:

```
{
  site_list(
    filters: {status: STATUS_PLANNED, OR: {tenant: {name: {exact: "Foo"}}}}
  ) {
    name
  }
}
```

```
{
  ip_address_list(
    filters: {parent:["172.16.0.1"]}
  ) {
    address
    status
  }
}
```

arthanson (dismissed):
Couple small suggestions, but it could go in even without those.

netbox/netbox/graphql/filter_lookups.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def get_filter(self):
+        for field in self.__strawberry_definition__.fields:
+            value = getattr(self, field.name, None)
+            if value is not strawberry.UNSET:
+                return value
+        return None
+
+    @strawberry_django.filter_field
+    def filter(self, info: Info, queryset: QuerySet, prefix: str = '') -> Tuple[QuerySet, Q]:
```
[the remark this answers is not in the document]
jeremystretch: They're very similar, but JSONFilter has a bit of additional logic (the `json_path` prefix). We might be able to simplify it in the future but I'm going to leave it alone for now just so we can get this wrapped up.

arthanson (approved):

Issue #18095, opened 2024-11-25
Display contact of parent objects


m0nkeyc0de:
### NetBox version

v4.1.7

### Feature type

Other

### Triage priority

N/A

### Proposed functionality

When contacts are assigned to an objet which can have child objects (for example Site Group, Site, Location, ...) they don't appear in the contact list of the child objects.

A contact assigned to a Site Group is very likely valid for a Sites of the Site Group (otherwise it would be assigned to the Site itself).
The same logic can be applied for the "Site > Location", "Location > Location" relationships.

The contacts of all parent objects should be visible in the contact list of the child objects.

### Use case

NetBox users will spend less time assigning Contacts and will also more quickly have access to all relevants Contacts for a Location.

### Database changes

No database change is needed as it's in the GUI.

Maybe a setting could be added to enable/disable that feature.

### External dependencies

No external dependencies

alehaa:
I like this FR and think its fairly easy to implement. If accepted, I volunteer to provide a PR for this.

1. I'd change `ObjectContactsView` to check in [`get_children()`](https://github.com/netbox-community/netbox/blob/main/netbox/tenancy/views.py#L25-L29), whether `parent` is a `NestedGroupModel`. If so, one can filter for all parents including the child, otherwise just the child as it does now.
2. If needed, views could extend this filterset as needed. I'm think about showing contacts of the site if a location is accessed (e.g. facility manager of a `Site`, also responsible for the `Location`).

Of course, the contacts should be sorted in a meaningful way, e.g. from most specific to least specific.

arthanson:
@alehaa do you want me to assign this to you?

alehaa:
@arthanson you can assign this to me. However, I'm a little confused by the "needs milestone" tag. Doesn't that mean we're waiting for confirmation if and when the maintainers allow submissions?

arthanson:
@alehaa assigned.  The needs milestone means that it will cause API changes or such that can't be done in a minor point release.  This would be included in the upcoming 4.3 release.

alehaa:
@rboucher-me, can you please elaborate on what changes you think are needed for the API? Unlike the UI views, there are currently no individual contact assignment API endpoints per object type, only the generic `/api/tenancy/contact-assignments/`. So the logic for extending returned contacts would be different for these two concepts.

alehaa:
@bctiemann I think this issue accidentally made it into the v4.2.5 changelog. As far as I know, it was only merged into the `feature` branch.

PR #18704, opened 2025-02-21
Fixes 18095: inherit contacts

alehaa:
### Fixes: #18095

Contacts for a `NestedGroupModel` will also include the contact assignments of the parents.

While at first it seemed tempting to include other related contact assignments (e.g. for a `Location` those of a `Site`), this is much more complicated than expected. The code proposed in this PR can also handle these relationships by iterating over related objects and adding constraints to the `filter`. In practice, however, it remains unclear which relationships are important to users, as these may vary between different NetBox deployments and even within a single deployment. In this case, I suggest opening a new FR with details of a possible lookup logic and adding it to `get_contacts()`.

### DIFF
--- a/netbox/tenancy/views.py
+++ b/netbox/tenancy/views.py
@@ -2,7 +2,6 @@ from django.contrib.contenttypes.models import ContentType
 from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext_lazy as _
 
-from netbox.models import NestedGroupModel
 from netbox.views import generic
 from utilities.query import count_related
 from utilities.views import GetRelatedModelsMixin, ViewTab, register_model_view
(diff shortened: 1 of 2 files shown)


netbox/tenancy/views.py
```diff
@@ -17,25 +17,13 @@ class ObjectContactsView(generic.ObjectChildrenView):
...
         weight=5000
     )
 
     def get_children(self, request, parent):
-        return ContactAssignment.objects.restrict(request.user, 'view').filter(
-            object_type=ContentType.objects.get_for_model(parent),
-            object_id=parent.pk
-        ).order_by('priority', 'contact', 'role')
-
-    def get_table(self, *args, **kwargs):
```
bctiemann, line 31: What happens to `get_table` here? Is it not needed?

alehaa: Previously, it was used to limit the columns of the contact assignments table so that the assigned object was not displayed. Now that the contacts of multiple objects are displayed, these columns are needed, so there is no need to restrict the original table.

bctiemann (approved):

PR #18710, opened 2025-02-23
Fixes 17357: Use virtual chassis name as fallback for device 

### LINKED ISSUE #17357: Use VirtualChassis name as fallback for devices
### NetBox version

v4.0.10

### Feature type

### Proposed functionality

If a device has no explicit name, but is a member of a `VirtualChassis`, this name (and position) should be used instead for all functionalities of NetBox. This especially affects the search index and the form fields to select a `Device`.

### Use case

Some virtual chassis concepts (e.g. Cisco VSS) don't allow individual host names for each hardware chassis. In this case, NetBox allows the user to set the name for the virtual chassis and leave the device name blank. NetBox will then already use a combination of the virtual chassis name and the member position as the display name in the device detail view or the rack elevation view. However, these names can't be searched for in the global search, nor in the device list view, nor when selecting a device, e.g. to connect an interface. Extending this functionality would make NetBox a little easier to use, while also increasing consistency.

### Database changes

None

### External dependencies

None

alehaa:
### Fixes: #17357

If a name is not defined for a device, it is automatically generated from other device data (i.e. virtual chassis member) if possible. This name is available as `label` attribute to not interfere with existing functionality or break APIs. This label will be used in several places like tables, while the API still returns the real name.

### DIFF
--- a/netbox/dcim/tests/test_models.py
+++ b/netbox/dcim/tests/test_models.py
@@ -590,6 +590,34 @@ class DeviceTestCase(TestCase):
         device2.full_clean()
         device2.save()
 
+    def test_device_label(self):
+        device1 = Device(
+            site=Site.objects.first(),
+            device_type=DeviceType.objects.first(),
+            role=DeviceRole.objects.first(),
+            name=None,
+        )
+        self.assertEqual(device1.label, None)
+
+        device2 = Device(
+            site=Site.objects.first(),
+            device_type=DeviceType.objects.first(),
+            role=DeviceRole.objects.first(),
+            name='Test Device 2',
+        )
+        self.assertEqual(device2.label, 'Test Device 2')
+
+        virtual_chassis = VirtualChassis.objects.create(name='VC 1')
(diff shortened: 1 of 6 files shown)


netbox/dcim/tests/test_models.py
```diff
@@ -590,6 +590,34 @@ def test_device_duplicate_names(self):
...
+        virtual_chassis = VirtualChassis.objects.create(name='VC 1')
+        device3 = Device(
+            site=Site.objects.first(),
+            device_type=DeviceType.objects.first(),
+            role=DeviceRole.objects.first(),
+            name=None,
+            virtual_chassis=virtual_chassis,
+            vc_position=2,
+        )
+        self.assertEqual(device3.label, 'VC 1:2')
```
arthanson, line 619: Should have a test if set name and virtual_chassis that it returns the name, not the generated one.

arthanson (changes requested):

arthanson (approved):

PR #18719, opened 2025-02-24
Generic prefetch on 4.2

### LINKED ISSUE #18409: Poor performance in cable trace API calls: N+1 problem
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v3.7, v4.2

### Python Version

3.11

### Steps to Reproduce

1. Fill enough data for devices with cables and connections.
2. Do API requests to load device, interfaces, ip-addresses like
http://localhost:8000/api/dcim/interfaces/?device_id=3971&limit=100&offset=0

### Expected Behavior

Data is loaded quite fast, <0.5 seconds. Number of db queries pretty small up to 10-20 per HTTP-request

### Observed Behavior

Some requests are pretty slow. Up to 3 seconds.
130 sql queries are produced on 3.7 and 160 db queries on 4.2

`CabledObjectModel.link_peers` has N+1 problem: https://github.com/netbox-community/netbox/blob/v3.7.8/netbox/dcim/models/device_components.py#L192
`CablePath._get_path` has N+1 problem as well: https://github.com/netbox-community/netbox/blob/v3.7.8/netbox/dcim/models/cables.py#L746

I tried to comment out these methods and got only 20 db queries on the same http request 

Additional info: n+1 problem should be eliminated

Tishka17:

### DIFF
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -1,3 +1,5 @@
+from django.contrib.contenttypes.prefetch import GenericPrefetch
+from django.db.models import Prefetch
 from django.http import Http404, HttpResponse
 from django.shortcuts import get_object_or_404
 from drf_spectacular.types import OpenApiTypes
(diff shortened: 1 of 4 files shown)


Tishka17:
@bctiemann thank you for your feedback, but please explain why this is issue is not yet accepted? I've found a big, provided information I thought was enough explaining the case. If it is not enough, let's discuss then. 

Closing a PR because you do not want to read it is not a good action. If you feel there is not enough information to make a decision on it - let's discuss and postpone reviewing, not close.

Tishka17:
> This PR in particular involves some fairly fundamental changes to the way querying works that would have widespread implications throughout the application

No, it doesn't. All the code is compatible. I've only added few prefetch_related annotations so speedup specific view. This feature is not required to use, the field is still accessible in old style.

PR #18723, opened 2025-02-24
Closes #18024: Add URL pattern for scripts to reference them by module.name

### LINKED ISSUE #18024: Add URL pattern for scripts to allow a specific script to be accessed by module and class in addition to ID
### NetBox version

v4.1.6

### Feature type

New functionality

### Triage priority

### Proposed functionality

Related to #16145, proposing to allow a URL pattern for scripts to reference them by `<module>.<class>`. The scope of this request is simply to allow this new URL pattern in addition to the existing `scripts/<int:pk>/`, `scripts/<int:pk>/source/`, and `scripts/<int:pk>/jobs/` patterns. The templates and `Script.get_absolute_url()` would still reference the pk. The suggested URL pattern should match the allowed pattern for the API resolved in #16145.

### Use case

When using Custom Links to link to a specific script, using the pk is problematic because it's subject to change as the scripts are updated over time. By allowing the `<module>.<class>` pattern, the custom links could be configured to use that pattern and return the correct script.

### Database changes

None

### External dependencies

None

atownson:
### Closes: #18024

Added script URL patterns to include the `module` and `name` references. Added a function to return the specified script when referenced by either the PK or the module.name pattern. The new URL patterns should match the existing URL patterns for the API.

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -1251,16 +1251,12 @@ class ScriptListView(ContentTypePermissionRequiredMixin, View):
 class BaseScriptView(generic.ObjectView):
     queryset = Script.objects.all()
 
-    def _get_script(self, **kwargs):
-        if 'pk' in kwargs:
-            pk = kwargs.get('pk')
+    def get_object(self, **kwargs):
+        if pk := kwargs.get('pk', False):
             return get_object_or_404(self.queryset, pk=pk)
-        elif 'module' in kwargs and 'name' in kwargs:
-            module = kwargs.get('module')
-            name = kwargs.get('name')
+        elif (module := kwargs.get('module')) and (name := kwargs.get('name', False)):
             return get_object_or_404(self.queryset, module__file_path=f'{module}.py', name=name)
-        else:
-            raise Http404
+        else: raise Http404
 
     def _get_script_class(self, script):
         """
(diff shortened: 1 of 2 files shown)


netbox/extras/views.py
```diff
@@ -1251,6 +1251,17 @@ def get(self, request):
 class BaseScriptView(generic.ObjectView):
     queryset = Script.objects.all()
 
+    def _get_script(self, **kwargs):
```
jnovinger, line 1254: **issue:** Please change this to `def get_object(self, **kwargs):`, which will make the changes below in the other views unnecessary.

netbox/extras/views.py
```diff
@@ -1251,6 +1251,17 @@ def get(self, request):
...
+        if 'pk' in kwargs:
+            pk = kwargs.get('pk')
+            return get_object_or_404(self.queryset, pk=pk)
+        elif 'module' in kwargs and 'name' in kwargs:
+            module = kwargs.get('module')
+            name = kwargs.get('name')
+            return get_object_or_404(self.queryset, module__file_path=f'{module}.py', name=name)
+        else:
+            raise Http404
```
jnovinger, line 1263: **note (non-blocking):** We do use the [walrus operator (assignment expression)](https://docs.python.org/3/reference/expressions.html#assignment-expressions) elsewhere in NetBox. Feel free to do so here, like so:

```python
        if pk := kwargs.get('pk', False):
            return get_object_or_404(self.queryset, pk=pk)
        elif (module := kwargs.get('module')) and (name := kwargs.get('name', False)):
            return get_object_or_404(self.queryset, module__file_path=f'{module}.py', name=name)
        else:
            raise Http404
```

jnovinger (changes requested):
Thanks for the quick action @atownson ! I just have a couple of change requests.

The note about the walrus operator is just a suggestion.

jnovinger (approved):
Thanks @atownson

PR #18725, opened 2025-02-24
Closes #17944: Allow filtering of ObjectVar and MultiObjectVar script inputs

### LINKED ISSUE #17944: Allow filtering of ObjectVar and MultiObjectVar input selections for scripts
### NetBox version

v4.1.6

### Feature type

New functionality

### Triage priority

N/A

### Proposed functionality

Proposing to extend the functionality of the object selection inputs (ObjectVar and MultiObjectVar) in scripts to be able to filter the selection in the same fashion that's available in the create/edit forms.
image
I might suggest adding the optional parameters `selector` (boolean with a default of False) and `query_params` (dictionary) to both ObjectVar and MultiObjectVar classes so that it matches the form parameters. `selector=True` would show the filter button, and the `query_params` dictionary to be used to apply the initial filtering.

### Use case

Our NetBox instance relies on a few scripts to assist in creating items. Currently we use separate inputs to control filtering (similar to the old behavior of NetBox). Adding the optional filtering would improve the user experience in the same way filtering improved the user experience when using the create/edit forms.

### Database changes

None

### External dependencies

None

atownson:
### Closes: #17944

Added the advanced object selection widget to the ObjectVar and MultiObjectVar script inputs. To render the selector button set `selector = True` on the input. The modals block was added to the script template to pass the `size='lg'`  parameter. Without it the modal was rendering thinner than normal.

I discovered an issue where the `query_params` defined for the input are not passed to the advanced object selector widget. I plan to open a separate bug issue for this.

### DIFF
--- a/docs/customization/custom-scripts.md
+++ b/docs/customization/custom-scripts.md
@@ -308,6 +308,7 @@ A particular object within NetBox. Each ObjectVar must specify a particular mode
 * `query_params` - A dictionary of query parameters to use when retrieving available options (optional)
 * `context` - A custom dictionary mapping template context variables to fields, used when rendering `<option>` elements within the dropdown menu (optional; see below)
 * `null_option` - A label representing a "null" or empty choice (optional)
+* `selector` - A boolean that, when True, includes an advanced object selection widget to assist the user in identifying the desired object (optional; False by default)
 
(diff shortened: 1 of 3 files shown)


atownson:
FYI, I referenced the associated issue for the PR in a new bug issue: #18727

jnovinger (approved):
Looks and feels solid. Thanks @atownson .

PR #18747, opened 2025-02-26
Closes #18287: Enable periodic synchronization for data sources

### LINKED ISSUE #18287: Allow periodic synchronization of data sources
### NetBox version

v4.1.10

### Feature type

New functionality

### Triage priority

### Proposed functionality

Allow a data source to be synchronized periodically. Since the `SyncDataSourceJob` uses the job runner framework, this feature is technically available, but must be added to the appropriate forms and related methods.

### Use case

Keeping data in sync automatically without user interaction.

### Database changes

None.

### External dependencies

None.

jeremystretch:
### Closes: #18287

- Add `sync_interval` field to DataSource
- Automatically enqueue a SyncDataSourceJob when the configured interval for a DataSource changes


jeremystretch:
I've noticed an issue that occurs when re-scheduling a job, but I don't think it's specific to this change. When enqueuing a new job after modifying the sync interval, the new job fill fail with an error like the following:

`IntegrityError('duplicate key value violates unique constraint "core_job_job_id_key"\nDETAIL: Key (job_id)=(1f4fdacb-1825-451e-b324-05a2f7d2de66) already exists.')`

I've seen this pop in other scenarios; I believe it's related to the background logic responsible for enqueuing Redis tasks. Will need to dig into this further.

netbox/core/signals.py
```diff
@@ -182,6 +181,20 @@ def clear_events_queue(sender, **kwargs):
...
+    """
+    When a DataSource is saved, check its sync_interval and enqueue a sync job if appropriate.
+    """
+    from .jobs import SyncDataSourceJob
+
+    if instance.sync_interval:
+        SyncDataSourceJob.enqueue_once(instance, interval=instance.sync_interval)
+    elif not created:
+        # Delete any previously scheduled recurring jobs for this DataSource
+        SyncDataSourceJob.get_jobs(instance).filter(interval__isnull=False).delete()
```
jnovinger, line 195: Ran in to what seem like edge cases while testing this:

1. when instance is marked disabled and saved, already queued recurring job is still run
2. when sync_interval is changed to None and saved, existing job records are deleted, even if the completed in one way or another. (Not sure if this one is intended or not.)
3. when sync_interval is changed to None (from Minutely in my case) saved, tasks still keeps executing in the worker every minute

jnovinger (commented):
No issues with the code, but ran in to what seem to be edge cases while testing the post-save signal handler.

netbox/core/signals.py
jeremystretch: Good catches! Appreciate the thorough review.

jnovinger: Oooh, found one last edge case I missed before. Although this may just fall under expected behavior, I think it could be confusing for the user:
1. Create a DataSource with a sync interval and is enabled
2. Manually delete the scheduled job

At this point any/all scheduled jobs are gone but the DataSource instance still shows it is enabled and scheduled (minutely in my test case). I could see that leading users to wonder why it indicates it should be running but is not.

Having said that, editing and saving the DataSource instances without making any changes does successfully re-schedule the recurring job.

So ... yeah, not sure this edge case is worth addressing. Might be better to just document that this is possible.

jnovinger (approved):
💥 all of my edge cases are resolved!

Except for the [one other case I just now noticed](https://github.com/netbox-community/netbox/pull/18747#discussion_r1976062230), which may or may not be worth addressing. I'd call it non-blocking.

PR #18756, opened 2025-02-27
Closes #17424: Add custom visibility toggle to ViewTab

### LINKED ISSUE #17424: Dynamic Tab Visibility Control in the ViewTabs Class
### NetBox version

v4.1.0

### Feature type

New functionality

### Proposed functionality

✨ Edit: Waiting for issue approval and assignment before creating a PR. The feature is already implemented.

The new visibility feature in ViewTabs class allows developers to control whether a tab is displayed based on the attributes of the object being viewed. This functionality is implemented using a callable (e.g., a function or a lambda) that accepts the object as an argument and returns a boolean indicating whether the tab should be visible.

By introducing dynamic visibility, users can fine-tune the display of tabs for specific object states, providing a more personalized and contextual navigation experience within views. If no callable is provided, the tab remains visible by default, preserving backward compatibility.

### Use case

**Conditional Display Based on Device Role**

For example, if you only want to display a tab for devices with the role "Bare Metal", you can pass a lambda function to the `visible` argument that checks the `device_role` attribute of the object. Here's how you can define it:

```python
visible=lambda obj: str(obj.device_role) == "Bare Metal"
```

In this case, the tab will only appear when the device’s role is exactly "Bare Metal". For all other roles, the tab will be hidden. This is particularly useful for scenarios where certain tabs are only relevant for specific types of devices or objects.

### Database changes

No changes.

### External dependencies

No external dependencies.

jeremystretch:
### Fixes: #17424

Adds an option `visible` parameter to ViewTab. If defined, this function will be used to control whether the tab should be rendered. If it returns true, visibility defers to the existing logic which considers the tab's badge value and `hide_if_empty` setting.

### DIFF
--- a/netbox/utilities/views.py
+++ b/netbox/utilities/views.py
@@ -206,22 +206,30 @@ class ViewTab:
 
     Args:
         label: Human-friendly text
+        visible: A callable which determines whether the tab should be displayed. This callable must accept exactly one
+            argument: the object instance. If a callable is not specified, the tab's visibility will be determined by
+            its badge (if any) and the value of `hide_if_empty`.
         badge: A static value or callable to display alongside the label (optional). If a callable is used, it must
             accept a single argument representing the object being viewed.
         weight: Numeric weight to influence ordering among other tabs (default: 1000)
         permission: The permission required to display the tab (optional).
-        hide_if_empty: If true, the tab will be displayed only if its badge has a meaningful value. (Tabs without a
-            badge are always displayed.)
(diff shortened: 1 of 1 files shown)


arthanson (approved):
This could go into main? It doesn't change the API

jeremystretch:
We tagged it for v4.3 because it introduces a new feature of the ViewTab class. Maybe a bit conservative but I'd rather avoid the risk of confusion.

PR #18784, opened 2025-02-28
Fixes #18768: allow removing secondary MACAddress from interface

### LINKED ISSUE #18768: Unable to remove a secondary MAC address from an interface
### Deployment Type

Self-hosted

### NetBox Version

v4.2.4

### Python Version

3.11

### Steps to Reproduce

All operations were done in the WebUI.

1. Create a virtual interface X and assign it to a virtual machine (virtualization/interfaces).
2. Create two MAC addresses A and B (dcim/mac-addresses).
3. Edit MAC address A and assign it to X.
4. Edit MAC address B and also assign it to X.
5. Edit virtual interface X and set A as primary MAC address.
6. Edit MAC address B and remove the assignment to X.

### Expected Behavior

I expected on step 6 that the MAC address object gets unassigned from the interface so it can be assigned to another interface.

### Observed Behavior

An Error was shown with the message "Cannot unassign MAC Address while it is designated as the primary MAC for an object".

The MAC address to be unassigned was NOT the primary MAC address.

jnovinger:
### Fixes: #18768 
- Changes check in `MACAddress.clean()` to also check if this MAC address is currently marked as the primary MAC address in the interface it is assigned to
- Adds regression tests

### DIFF
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -1550,7 +1550,10 @@ class MACAddress(PrimaryModel):
             ct = ObjectType.objects.get_for_id(self._original_assigned_object_type_id)
             original_assigned_object = ct.get_object_for_this_type(pk=self._original_assigned_object_id)
 
-            if original_assigned_object.primary_mac_address:
+            if (
+                original_assigned_object.primary_mac_address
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @jnovinger!

PR #18802, opened 2025-03-04
Fixes: #18753 Webhooks trigger on custom script dry-run

### LINKED ISSUE #18753: Webhooks trigger on custom script dry-run
### Deployment Type

Self-hosted

### NetBox Version

v4.2.4

### Python Version

3.11

### Steps to Reproduce

We have a custom script that creates a device, including setting up new IP addresses and interfaces.
Additionally, we have a webhook that triggers on ip address add/update/delete, that triggers DNS and DHCP updates. 

When the script is run with "commit changes" disabled, i.e. dry-run mode, the webhooks are triggered, even though these changes were reverted immediately.
This used to work properly (i.e. no webhook triggering) in v4.0.10.

Note that even though the webhook is configured to trigger on deletions, the dry-run only triggered `'event': 'created'` webhooks, so it's not a create-and-immediately-delete kind of situation (which IMO would be wrong, too).

### Expected Behavior

A dry-run Script should not trigger webhooks. This was the observed behaviour in  v4.0.10.

### Observed Behavior

A dry-run Script triggers webhooks in v4.2.4, even though the data was never persisted into the database.

renatoalmeidaoliveira:
### Fixes: #18753 Webhooks trigger on custom script dry-run

- Added commit test as @jeremystretch commented in the issue
- Followed the same wokring pattern as in v4.0.10 [code](https://github.com/netbox-community/netbox/blob/e93e9ac4a093f26a6fd8f6a1d6bfb0bb49de20f6/netbox/extras/scripts.py#L703)
- Tested with a simple script editing an IPAddress and the fix worked.

### DIFF
--- a/netbox/extras/jobs.py
+++ b/netbox/extras/jobs.py
@@ -100,7 +100,10 @@ class ScriptJob(JobRunner):
 
         # Execute the script. If commit is True, wrap it with the event_tracking context manager to ensure we process
         # change logging, event rules, etc.
-        with ExitStack() as stack:
-            for request_processor in registry['request_processors']:
-                stack.enter_context(request_processor(request))
+        if commit:
+            with ExitStack() as stack:
+                for request_processor in registry['request_processors']:
+                    stack.enter_context(request_processor(request))
+                self.run_script(script, request, data, commit)
+        else:
             self.run_script(script, request, data, commit)


jnovinger (approved):
Works as advertised! I tested this in the UI and via the `runscript` management command. It acted correctly in both scenarios.

PR #18803, opened 2025-03-04
Fixes: #18772 Add filter 'type' for virtual circuits 

### LINKED ISSUE #18772: Add filter 'type' for virtual circuits
### NetBox version

v4.2.3

### Feature type

### Proposed functionality

Filter function for virtual circuits is missing option 'type'. Option 'type' should correspond to virtual circuit types just as in case with regular circuits.

### Use case

With increased number of VC it would be really helpfull to be able to filter based on 'type' 

### Database changes

dont know

### External dependencies

None

renatoalmeidaoliveira:
### Fixes: #18772 Add filter 'type' for virtual circuits 

- The `type` field was already defined in `VirtualCircuitFilterForm` but with a typo when comparing with the defined fields.


jeremystretch (approved):

Issue #18820, opened 2025-03-05
Drop support for PostgreSQL 13


jeremystretch:
### Proposed Changes

Require PostgreSQL 14 or later beginning with NetBox v4.3.

### Justification

Django 5.2 [no longer supports PostgreSQL 13](https://docs.djangoproject.com/en/dev/releases/5.2/#dropped-support-for-postgresql-13). Upstream support for PostgreSQL 13 ends in November 2025.

### Impact

Users running PostgreSQL 13 will need to upgrade.

PR #18822, opened 2025-03-05
Closes #18743: Upgrade to Django 5.2

### LINKED ISSUE #18743: Upgrade to Django 5.2
### Proposed Changes

Upgrade from Django 5.1 to Django 5.2 for NetBox v4.3.

### Justification

Django 5.2 is [expected to release](https://code.djangoproject.com/wiki/Version5.2Roadmap) in early April, which coincides with the NetBox v4.3 beta period.

jeremystretch:
### Closes: #18743

- Bumps Django from 5.1 to 5.2
- Adapts `RemoteUserMiddleware` to use the newer middleware format for Django 5.2

We're currently running Django 5.2 beta1 as the final release is not yet available (ETA April 2).

Note: Django 5.2 drops support for PostgreSQL 13. This will be addressed under #18820.

### DIFF
--- a/base_requirements.txt
+++ b/base_requirements.txt
@@ -1,6 +1,6 @@
 # The Python web framework on which NetBox is built
 # https://docs.djangoproject.com/en/stable/releases/
-Django<5.2
+Django==5.2.*
 
 # Django middleware which permits cross-domain API requests
 # https://github.com/adamchainz/django-cors-headers/blob/main/CHANGELOG.rst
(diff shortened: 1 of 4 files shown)


bctiemann (changes requested):
I'm getting this, which I think has been addressed in main but we probably want to resolve just so it's got a clean set of packages at merge time:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
mkdocstrings-python 1.16.2 requires mkdocstrings>=0.28.2, but you have mkdocstrings 0.27.0 which is incompatible.
```

bctiemann (approved):

PR #18826, opened 2025-03-06
Prefetch interface data for REST API on netbox 4.2

### LINKED ISSUE #18409: Poor performance in cable trace API calls: N+1 problem

Tishka17:
### Fixes: #18409

* Do not do additional requests for `CabledObjectModel.link_peers`
* Custom field for arrays containing generic foreign keys
* Use custom field instead of `CablePath._get_path` method
* Enable prefetching for interface `_path` and `cable`


Tishka17:
I see that  I've missed `termination.link is not None` logic, I'm going to fix it

bctiemann:
@Tishka17 I've been testing this and the performance gains do look good. I'm seeing similar query and timing numbers to yours in fetching interfaces/peers/path_objects etc, and the data looks to be the same as before. A couple of questions though, mostly just around organization:

1. Now that NetBox 4.2 is on Django 5.1, does that change your earlier approach to `GenericPrefetch` as that originally had to be backported into Django 4.2 as of your [comments](https://github.com/netbox-community/netbox/issues/18409#issuecomment-2674774045) 3 weeks ago? Can the change be simplified to use anything that is now part of Django itself?
2. I see that you put `GenericArrayForeignKey` into `utilities/generics/field.py`; would `utilities/fields.py` not be an appropriate place for it? It already has the conceptually similar `RestrictedGenericForeignKey` class.

Thanks for your patience!

Tishka17:
1. In current version I am using GenericPrefetch from Django 5.x. Backport is needed only for older versions of netbox, but I guess we are not going to updated them. 

2. Ok, I will move files.

netbox/utilities/fields.py
```diff
@@ -186,3 +191,131 @@ def deconstruct(self):
         kwargs["to_model"] = self.to_model_name
         kwargs["to_field"] = self.to_field_name
         return name, path, args, kwargs
+
```
bctiemann, line 194: Could you remove this extra line?

bctiemann (approved):

PR #18827, opened 2025-03-06
Fixes: #18808 -  Fix incorrect dependencies on squashed migrations

### LINKED ISSUE #18808: Clean up nonexistent squashed migrations
### Proposed Changes

The `sqlmigrate` command fails with errors such as: `django.db.migrations.exceptions.NodeNotFoundError: Migration extras.0002_squashed_0059 dependencies reference nonexistent parent node ('dcim', '0002_auto_20160622_1821')`

This results from migration squashing where the `dependencies` list was not updated to point to the new squashed migrations.

Proposed change is to fix all `dependencies` pointers in existing migrations to ensure they point to the existing migrations containing the original (missing) parent nodes.

### Justification

Though `sqlmigrate` is rarely used (it produces the native SQL operations to be applied during migrations in case they need to be run manually), it is useful as a diagnostic tool and ought to work properly. (Note: `migrate` itself does not seem to have this issue.)

bctiemann:
### Fixes: #18808

After previously squashing migrations, many child migrations depending on those original migrations had not been updated to point to the new squashed migrations in `dependencies`. This caused errors when using `./manage.py sqlmigrate`.

This change fixes the `dependencies` of all erroneous migrations to point to the correct squashed parent migrations.

### DIFF
--- a/netbox/account/migrations/0001_initial.py
+++ b/netbox/account/migrations/0001_initial.py
@@ -8,7 +8,7 @@ class Migration(migrations.Migration):
     initial = True
 
     dependencies = [
-        ('users', '0004_netboxgroup_netboxuser'),
+        ('users', '0002_squashed_0004'),
     ]
 
     operations = [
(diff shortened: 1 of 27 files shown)


arthanson:
@bctiemann can you rebase to feature, otherwise looks good.

arthanson (approved):

PR #18830, opened 2025-03-06
Closes #18138: Add advanced object selector for custom field object and multi-object inputs

### LINKED ISSUE #18138: Allow filtering of ObjectVar and MultiObjectVar input selections for custom fields
### NetBox version

v4.1.2

### Feature type

New functionality

### Triage priority

### Proposed functionality

This is very simular to [#17944](https://github.com/netbox-community/netbox/issues/17944)

Proposing to extend the functionality of the object selection inputs (ObjectVar and MultiObjectVar) in Custom Fields to be able to filter the selection in the same fashion that's available in the create/edit forms.

image

Suggest adding the optional parameters selector (boolean with a default of False).

### Use case

I wanted to add a custom field to "Provider Network" to assign a Network to a Interface. Adding the custom field Interface worked well, but just shows all the interfaces which is unusable. A selection filter would make this much easier to use so we could filter on a device and then choose the interface.

### Database changes

None

### External dependencies

None

atownson:
### Closes: #18138

The approach here was to hard-code the `selector=True` argument for the object and multi-object inputs, thus always rendering the advanced object selector button. The advantage to this approach is that it does not require an additional attribute on the `CustomField` model or any logic to show/hide the input for such an attribute on the create form.


bctiemann:
@atownson I'm getting this error when I try to set up a custom field with object or multiobject type and then use the selector:

Screenshot 2025-03-24 at 7 38 58 PM

I wouldn't be surprised if it's an issue on my end, but could you please check in case it's a bug?

arthanson:
Working for me:
Monosnap Editing circuit 1002840283 | NetBox 2025-03-26 09-53-33
Monosnap Editing circuit 1002840283 | NetBox 2025-03-26 09-53-58

arthanson (approved):

PR #18844, opened 2025-03-08
Fixes: #18582 Bulk import prefixes with associated VLAN not working when multiple VLANs with the same vid exist. 

### LINKED ISSUE #18582: Bulk import prefixes with associated VLAN not working when multiple VLANs with the same vid exist.
### Deployment Type

Self-hosted

### NetBox Version

v4.2.3

### Python Version

3.12

### Steps to Reproduce

1. Create multiple VLAN entries with the same vid but associated to different sites. I use this to represent VLANs with the same vid, but located in different sites with different prefixes.

2. Import prefix csv with the following fields: prefix, status, role, scope_type, scope_id, vlan. Include the previously added VLANs vid in the "vlan" field.

3. The import will fail because multiple vlans with the same vid exist

Image

### Expected Behavior

Import succeeds associating the right vlan depending on the DCIM.site value. The import should use the VLAN with the specified vid contained in the specified scope (site). 

To fix this issue, it could also be possible to specify the VLAN object ID instead of the vid to avoid confusion.

### Observed Behavior

Import fails with error : 

Image

renatoalmeidaoliveira:
### Fixes: #18582 Bulk import prefixes with associated VLAN not working when multiple VLANs with the same vid exist. 

- `PrefixImportForm` already implemented a test case for VLAN `Site` and `VLANGroup` [code](https://github.com/netbox-community/netbox/blob/29c25e39fce5b549d4d960bcc7a65379059855ac/netbox/ipam/forms/bulk_import.py#L222)
- Only a `CSVModelChoiceField` for vlan_group was implemented

The fix consisted in implementing a `CSVModelChoiceField` for `Site`

Aditional observations:

It should be possible to setup the `VLAN` queryset to the same as the Prefix scope if existent, but IMO that could lead to unwanted results, like:
- User Creates a `VLAN` that doesn't belong to any `Site`, that `VLAN` gonna be missing in the queryset.

### DIFF
--- a/netbox/ipam/forms/bulk_import.py
+++ b/netbox/ipam/forms/bulk_import.py
@@ -177,6 +177,13 @@ class PrefixImportForm(ScopedImportForm, NetBoxModelImportForm):
         to_field_name='name',
         help_text=_("VLAN's group (if any)")
     )
+    site = CSVModelChoiceField(
+        label=_('VLAN Site'),
+        queryset=Site.objects.all(),
+        required=False,
+        to_field_name='name',
+        help_text=_("VLAN's site (if any)")
+    )
     vlan = CSVModelChoiceField(
         label=_('VLAN'),
         queryset=VLAN.objects.all(),
@@ -200,8 +207,8 @@ class PrefixImportForm(ScopedImportForm, NetBoxModelImportForm):
     class Meta:
         model = Prefix
         fields = (
-            'prefix', 'vrf', 'tenant', 'vlan_group', 'vlan', 'status', 'role', 'scope_type', 'scope_id', 'is_pool',
-            'mark_utilized', 'description', 'comments', 'tags',
+            'prefix', 'vrf', 'tenant', 'vlan_group', 'site', 'vlan', 'status', 'role', 'scope_type', 'scope_id',
+            'is_pool', 'mark_utilized', 'description', 'comments', 'tags',
         )
         labels = {
             'scope_id': _('Scope ID'),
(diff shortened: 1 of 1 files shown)


netbox/ipam/forms/bulk_import.py
```diff
@@ -177,6 +177,13 @@ class PrefixImportForm(ScopedImportForm, NetBoxModelImportForm):
         to_field_name='name',
         help_text=_("VLAN's group (if any)")
     )
+    site = CSVModelChoiceField(
```
jeremystretch, line 180: We should rename this to `vlan_site` to make it clear that the field is considered for only the VLAN to which the prefix is being assigned, and not the prefix itself.
```suggestion
    vlan_site = CSVModelChoiceField(
```

netbox/ipam/forms/bulk_import.py
```diff
@@ -200,8 +207,8 @@ class PrefixImportForm(ScopedImportForm, NetBoxModelImportForm):
     class Meta:
         model = Prefix
         fields = (
-            'prefix', 'vrf', 'tenant', 'vlan_group', 'vlan', 'status', 'role', 'scope_type', 'scope_id', 'is_pool',
-            'mark_utilized', 'description', 'comments', 'tags',
+            'prefix', 'vrf', 'tenant', 'vlan_group', 'site', 'vlan', 'status', 'role', 'scope_type', 'scope_id',
```
jeremystretch, line 210:
```suggestion
            'prefix', 'vrf', 'tenant', 'vlan_group', 'vlan_site', 'vlan', 'status', 'role', 'scope_type', 'scope_id',
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #18847, opened 2025-03-09
Fixes: #18744 Add Contact link as hyperlink

### LINKED ISSUE #18744: Contact Link should be hyperlinked
### NetBox version

v4.2.4

### Feature type

### Proposed functionality

In the contacts table for a Location (or any other object), the Contact Link column should be a hyperlink.

### Use case

This is how it looks now, so you have to copy/paste the URL.
Image

### Database changes

_No response_

### External dependencies

_No response_

antoinekh:
### Fixes: #18744 

Add changes similar to the already linkify_phone but for url but there may be another way to do the linkify.

### DIFF
--- a/netbox/utilities/tables.py
+++ b/netbox/utilities/tables.py
@@ -32,15 +32,6 @@ def linkify_phone(value):
     return f"tel:{value.replace(' ', '')}"
 
 
-def linkify_url(value):
-    """
-    Render a URL as a hyperlink.
-    """
-    if value is None:
-        return None
-    return value
-
-
 def register_table_column(column, name, *tables):
     """
     Register a custom column for use on one or more tables.
--- a/netbox/utilities/tables.py
+++ b/netbox/utilities/tables.py
@@ -31,6 +31,7 @@ def linkify_phone(value):
         return None
     return f"tel:{value.replace(' ', '')}"
 
+
 def linkify_url(value):
     """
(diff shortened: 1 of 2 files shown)


netbox/utilities/tables.py
```diff
@@ -32,6 +32,15 @@ def linkify_phone(value):
     return f"tel:{value.replace(' ', '')}"
 
 
+def linkify_url(value):
+    """
+    Render a URL as a hyperlink.
+    """
+    if value is None:
+        return None
+    return value
```
jnovinger, line 41: **nitpick:** There's no need for the `if` test here--in all cases, `linkify_url` is returning the value of `value` essentially unchanged (unlike `linkify_phone` above). Additionally, while the `Contact.link` field is allowed to be empty, it's not allowed to be nullable (e.g. `None`).

I think it would probably be simpler to use a lambda function in the `linkify` argument for the `ContactAssignmentTable.contact_link`, like in https://github.com/netbox-community/netbox/blob/main/netbox/ipam/tables/ip.py#L344.

jnovinger (changes requested):

netbox/utilities/tables.py
antoinekh: Thanks for the feedback. I changed with a lambda as this is much simpler

jnovinger (approved):
Thanks @antoinekh !

PR #18854, opened 2025-03-10
Fixes #18782: Misconfigured `ObjectListWidget`s now degrade gracefully

### LINKED ISSUE #18782: Dashboard broken after adding widget
### Deployment Type

Self-hosted

### NetBox Version

v4.2.3

### Python Version

3.12

### Steps to Reproduce

1. Go to main dashboard page and click Unlock.
2. Click Add Widget.
3. Select object type Notifications.
4. Click Save.

Dashboard gets the error from the screenshot below.

### Expected Behavior

We shouldn't be able to select models that don't work, or cause the dashboard to break.

### Observed Behavior

Image

jnovinger:
### Fixes: #18782 

This PR does two things to address this bug:
- Updated the check before `htmx_url` has query params added so that we don't try to add them to `None`. This has the effect of properly triggering the branch in the template that indicates this widget can not be displayed.
- It also adds validation for `ObjectListWidget.ConfigForm.model` via a `clean_model` method that checks to make sure that a model is supported by `ObjectListWidget` at the time of creation. This results in the following error for the user:
image

I'm open to adjust that error message as needed.

I also played around with disabling unsupported object choices, but ended up not including those in the PR. Will add a screenshot if anyone is interested in seeing them.

### DIFF
--- a/netbox/extras/dashboard/widgets.py
+++ b/netbox/extras/dashboard/widgets.py
@@ -9,6 +9,7 @@ import requests
 from django import forms
 from django.conf import settings
 from django.core.cache import cache
+from django.db.models import Model
 from django.template.loader import render_to_string
 from django.urls import NoReverseMatch, resolve, reverse
 from django.utils.translation import gettext as _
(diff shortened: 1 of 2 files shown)


bctiemann:
@jnovinger This works in scope, however I noticed another weird adjacent behavior which we might want to try to fix at the same time if possible, if it's in the same code area (if not, I'll open a separate bug).

Screenshot 2025-03-18 at 1 56 17 PM

If you click the X to clear the "Widget type" value (or backspace the existing value), the page immediately bombs to this error:

Screenshot 2025-03-18 at 1 57 43 PM

This behavior is in `main` too, so it isn't related to your fix. I'll probably open another issue accordingly as it's likely a separate problem, but please weigh in if it looks familiar.

bctiemann (approved):

PR #18859, opened 2025-03-10
Closes #17602: adds comments field to NestedGroupModel children

### LINKED ISSUE #17602: Markdown comments field on Locations
### NetBox version

v4.0.8

### Feature type

Data model extension

### Proposed functionality

I would like to have a `Comments` field on Locations, which supports markdown the same as we have on other objects.

### Use case

I currently have a URL custom field on each Location, linking to a folder which contains all the relevant config files and construction drawings for that room. Sometimes there is a need to include more than one link here.  I think a comments field would be a better way to store this info, using Markdown to make long URLs look more friendly.

I would also use this field to store brief operational notes on each room.

### Database changes

_No response_

### External dependencies

_No response_

jnovinger:
### Fixes: #17602
- Adds `comments` field to the following models, via their inheritance of `netbox.models.NestedGroupModel`:
  - [x] `dcim.Location`
  - [x] `dcim.Region`
  - [x] `dcim.SiteGroup`
  - [x] `tenancy.ContactGroup`
  - [x] `tenancy.TenantGroup`
  - [x] `wireless.WirelessLANGroup`
- Hooks up the new `comments` field to all the required places

#### Review Notes

I tried to keep changes for each model to its own commit, to make review easier. However, there were a couple of commits that address things I missed.


netbox/tenancy/filtersets.py
```diff
@@ -51,6 +51,16 @@ class Meta:
         model = ContactGroup
         fields = ('id', 'name', 'slug', 'description')
 
+    def search(self, queryset, name, value):
```
bctiemann, line 54: These get pretty redundant; maybe we can put this `search` variant into a mixin if it's common to all `NestedGroupModel`s?

jnovinger: No objections here. I'll see about making that happen.

jeremystretch: Just be careful to account for any deviations. LocationFilterSet, for instance, also searches on the `facility` field.

jnovinger:
Apologies for the force push, I wasn't thinking when I rebased this branch.

bctiemann (approved):

PR #18860, opened 2025-03-10
Fixes #18838: Correctly reject invalid falsy local context data

### LINKED ISSUE #18838: Local Context Data can be set to non-none falsy values
### Deployment Type

Self-hosted

### NetBox Version

v4.2.5

### Python Version

3.10

### Steps to Reproduce

1. Edit a device in the UI (e.g. https://demo.netbox.dev/dcim/devices/37/)
2. Put in 0 as the local config context and save.

### Expected Behavior

Change gets rejected for not being correct JSON (like it does for any other flat values).
Config Context also correctly rejects 0 or "" values.

### Observed Behavior

Change is accepted.

Image

Relevant code: https://github.com/netbox-community/netbox/blame/v4.2.5/netbox/extras/models/configs.py#L202-L206

Happy to put up a PR for this.

fabi125:
### Fixes: #18838

Reject falsy values like the empty string, 0 or False as local context data.

I verified that `None`/`null` still works as expected and that an empty string in UI gets correctly converted to `None` before `clean()` is called.

### DIFF
--- a/netbox/dcim/tests/test_api.py
+++ b/netbox/dcim/tests/test_api.py
@@ -1396,6 +1396,34 @@ class DeviceTest(APIViewTestCases.APIViewTestCase):
         self.assertHttpStatus(response, status.HTTP_200_OK)
         self.assertEqual(response.data['content'], f'Config for device {device.name}')
 
+    def test_valid_local_context_data(self):
+        self.add_permissions('dcim.change_device')
+        device = Device.objects.first()
+        url = reverse('dcim-api:device-detail', kwargs={'pk': device.pk})
+
+        response = self.client.patch(url, {"local_context_data": None}, format='json', **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+
+        response = self.client.patch(url, {"local_context_data": {"foo": "bar"}}, format='json', **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+
+    def test_invalid_local_context_data(self):
+        self.add_permissions('dcim.change_device')
+        device = Device.objects.first()
+        url = reverse('dcim-api:device-detail', kwargs={'pk': device.pk})
+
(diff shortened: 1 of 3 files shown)


netbox/dcim/tests/test_api.py
```diff
@@ -1396,6 +1396,34 @@ def test_render_config(self):
...
+
+        response = self.client.patch(url, {"local_context_data": 0}, format='json', **self.header)
+        self.assertContains(response, 'JSON data must be in object form.', status_code=status.HTTP_400_BAD_REQUEST)
+
+        response = self.client.patch(url, {"local_context_data": False}, format='json', **self.header)
+        self.assertContains(response, 'JSON data must be in object form.', status_code=status.HTTP_400_BAD_REQUEST)
+
+        response = self.client.patch(url, {"local_context_data": "foo"}, format='json', **self.header)
+        self.assertContains(response, 'JSON data must be in object form.', status_code=status.HTTP_400_BAD_REQUEST)
+
```
jnovinger, line 1426: Thanks for adding regression tests for this. It's very much appreciated!

However, I'd like to get the tests as close to the actual code change as possible. Let's move these to `extras/tests/test_models.py::ConfigContextTest` and adapt them just call `.clean()` directly.

Something like this:
```python
    def test_valid_local_context_data(self):
        device = Device.objects.first()
        device.local_context_data = None
        device.clean()

        device.local_context_data = {"foo": "bar"}
        device.clean()

    def test_invalid_local_context_data(self):
        device = Device.objects.first()

        device.local_context_data = ""
            device.clean()

        device.local_context_data = 0
            device.clean()

        device.local_context_data = False
            device.clean()

        device.local_context_data = 'foo'
            device.clean()
```

jnovinger (changes requested):
The change does the trick. However, let's move the tests to be as close to the code change as possible.

fabi125 (commented):
Sure! Tests moved.

jnovinger (approved):
Thanks @fabi125 !

PR #18861, opened 2025-03-10
Fixes #18845: restores sort behavior for DeviceTable.name column

### LINKED ISSUE #18845: Cannot sort Devices by device NAME
### Deployment Type

Self-hosted

### NetBox Version

v4.2.5

### Python Version

3.12

### Steps to Reproduce

On any netbox with one or more devices:

1. Go to DEVICES > Devices
2. Click the "NAME" column heading to sort by that field.

### Expected Behavior

Devices to be sorted by name.

### Observed Behavior

Nothing.  Devices are not sorted by name.

This also happens on the Demo NetBox:

https://demo.netbox.dev/dcim/devices/?sort=name

In this case, when the sort field is specified on the command line, you get:

```
Server Error

Cannot resolve keyword 'label' into field. Choices are: airflow, asset_tag, bookmarks, cabletermination, cluster, cluster_id, comments, config_template, config_template_id, console_port_count, console_server_port_count, consoleports, consoleserverports, contacts, created, custom_field_data, description, device_bay_count, device_type, device_type_id, devicebays, face, front_port_count, frontports, id, images, interface_count, interfaces, inventory_item_count, inventoryitems, journal_entries, last_updated, latitude, local_context_data, location, location_id, longitude, module_bay_count, modulebays, modules, name, oob_ip, oob_ip_id, parent_bay, platform, platform_id, position, power_outlet_count, power_port_count, poweroutlets, powerports, primary_ip4, primary_ip4_id, primary_ip6, primary_ip6_id, rack, rack_id, rear_port_count, rearports, role, role_id, serial, services, site, site_id, status, subscriptions, tagged_items, tags, tenant, tenant_id, vc_master_for, vc_position, vc_priority, vdcs, virtual_chassis, virtual_chassis_id, virtual_machines

Python version: 3.12.9
NetBox version: 4.2.4
Plugins: 
  netbox_demo: 0.5.0
```

jnovinger:
### Fixes: #18845 
- Restores sort order using `Device.name` column on `DeviceTable.name` table column.

This was a regression introduced in bf836c9bc2 (in v4.2.5).

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -143,10 +143,8 @@ class PlatformTable(NetBoxTable):
 class DeviceTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
     name = tables.TemplateColumn(
         verbose_name=_('Name'),
-        accessor=Accessor('label'),
         template_code=DEVICE_LINK,
         linkify=True,
-        order_by=('name',)
     )
     status = columns.ChoiceFieldColumn(
         verbose_name=_('Status'),
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -145,7 +145,8 @@ class DeviceTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
         verbose_name=_('Name'),
         accessor=Accessor('label'),
         template_code=DEVICE_LINK,
-        linkify=True
+        linkify=True,
+        order_by=('name',)
     )
     status = columns.ChoiceFieldColumn(
         verbose_name=_('Status'),
--- a/netbox/dcim/tables/template_code.py
+++ b/netbox/dcim/tables/template_code.py
@@ -35,7 +35,7 @@ WEIGHT = """
 """
 
 DEVICE_LINK = """
-{{ value|default:'<span class="badge text-bg-info">Unnamed device</span>' }}
+{{ record.label|default:'<span class="badge text-bg-info">Unnamed device</span>' }}
 """
 
 DEVICEBAY_STATUS = """
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -2025,7 +2025,7 @@ class PlatformBulkDeleteView(generic.BulkDeleteView):
 
 @register_model_view(Device, 'list', path='', detail=False)
 class DeviceListView(generic.ObjectListView):
-    queryset = Device.objects.all()
+    queryset = Device.objects.select_related('virtual_chassis')
     filterset = filtersets.DeviceFilterSet
     filterset_form = forms.DeviceFilterForm
     table = tables.DeviceTable


netbox/dcim/tables/devices.py
```diff
@@ -145,7 +145,8 @@ class DeviceTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
         verbose_name=_('Name'),
         accessor=Accessor('label'),
```
mraerino, line 146: i think this might also need to be changed?
```suggestion
        accessor=Accessor('name'),
```

netbox/dcim/tables/devices.py
```diff
@@ -145,7 +145,8 @@ class DeviceTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
...
         template_code=DEVICE_LINK,
-        linkify=True
+        linkify=True,
+        order_by=('name',)
```
jeremystretch, line 149: This isn't needed if we correct the accessor.

netbox/dcim/tables/devices.py
jeremystretch, line 146: As @mraerino points out, the accessor needs to be reverted to `name`.

jeremystretch (changes requested):

jnovinger:
@mraerino @jeremystretch , thanks for pointing that out. I had interpreted the [previous change (that caused this behavior)](https://github.com/netbox-community/netbox/pull/18710) to be intentional with respect to the display value of the object's `Name` column value.

Reverting the `accessor` argument reverts the behavior of using `Device.label` value for the `Name` column. See screenshots below, first is with my change, the second is with the suggested changes to this PR. In the second, you can see that devices belonging to the Tenant: NC State, Site: MDF, Location: Row 1 lose their descriptive value in the `Name` column and simply become `"Unnamed Device"`.

Behavior based on this PR as filed:
image

Behavior with PR suggested changes:
image


I just want to clarify, this is the behavior we want--that is, these devices show up with `"Unnamed Device"` in the `Name` column?

netbox/dcim/tables/devices.py
alehaa: Instead of switching the accessor to name, it should be removed entirely. It was added in bf836c9 to display a value other than `name`.

I believe removing this line and switching the template `DEVICE_LINK` from `value` to `record.label` might preserve displaying the label but fix sorting. Sorry I didn't catch this before.

alehaa:
> The "Unnamed device" label signifies that name is null for the device (which is permitted).

@jeremystretch the intention of #17357 was exactly to change this behavior and show the same display name as in the rack elevations view. I proposed a change to fix sorting but preserving #17357 in the comments above: https://github.com/netbox-community/netbox/pull/18861#discussion_r1990177112

jeremystretch:
The column accessor should be set to a concrete value in the table data. If you want the column to display something other than the accessor's value, you need to use TemplateColumn or override the columns' `render()`/`value()` methods.

However, be aware that rendering attributes of any related objects which have not been prefetched in the querysey may trigger additional queries.

netbox/dcim/tables/devices.py
```diff
@@ -143,9 +143,8 @@ class Meta(NetBoxTable.Meta):
 class DeviceTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
     name = tables.TemplateColumn(
         verbose_name=_('Name'),
-        accessor=Accessor('label'),
         template_code=DEVICE_LINK,
-        linkify=True
+        linkify=True,
```
jnovinger, line 147: Oops, forgot I added this when adding the previous `order_by` argument. I'm a monster!

jeremystretch (approved):

Issue #18729, opened 2025-02-25
Netbox API returns duplicate resources during paging with offset


dankotrajkovic:
### Deployment Type

Self-hosted

### NetBox Version

v4.2.3

### Python Version

3.12

### Steps to Reproduce

Use Python or Postman API to GET the clusters from Netbox by paging through the resources.
The real-world use case is to load a larger list of clusters using the paging mechanism.

To reproduce Run the following code
```
import requests


def main():
    """
    Pull netbox clusters for demo.netbox.dev using limit=5
    The idea of small limit is just to simulate a bigger database where we have to do multiple
    requests of 50 items to load a larger list of for example 250 clusters.

    The idea is to show that within the returned items few duplicates appear.
    :return:
    """

    cluster_list = [] # Place to store the clusters after with each requests call
    cluster_unique_ids = set() # Set to store the unique IDs of clusters loaded from netbox


    # Collect the clusters
    headers = {
        'Accept': 'application/json',
        'Authorization': 'Token 6a768e6363830a536ffa07abf261c1d64d365b9a'
    }

    parameters = {
        'limit': 5,
        'offset': 0
    }
    while True:
        response = requests.get('https://demo.netbox.dev/api/virtualization/clusters', headers=headers,
                                params=parameters)
        if response.status_code == 200:
            print(f'Collected clusters from Netbox with offset: {parameters["offset"]}, limit: {parameters["limit"]}')
            cluster_list.extend(response.json()['results'])
            parameters['offset'] += parameters['limit']
        if  not response.json()['next']:
            break

    # Check if there are any duplicates in the clusters
    for cluster in cluster_list:
        if cluster['name'] not in cluster_unique_ids:
            cluster_unique_ids.add(cluster['name'])
        else:
            print(f'Duplicate Cluster. Name: {cluster["name"]}, ID: {cluster["id"]}')


if __name__ == '__main__':
    main()
```

To Reproduce in Postman:
Issue the GET requests with following path:
https://demo.netbox.dev/api/virtualization/clusters/?limit=5&offset=5
and then issue again with
https://demo.netbox.dev/api/virtualization/clusters/?limit=5&offset=30

You will see that Cluster with ID 10 repeats in two responses. The ID might differ at times.

The Example above uses demo.netbox.dev, but the same behavior we experience on our on-prem self-hosted instance. With 150 clusters we see about 30-35 duplicates.

### Expected Behavior

Expect not to have duplicates returned as we page through the clusters.

[...]

PR #18867, opened 2025-03-11
Fixes: #18863 - Exempt MPTT-based models from centrally applying ordering on querysets

### LINKED ISSUE #18863: List views of self-nesting (recursive) object types are displayed incorrectly
### Deployment Type

Self-hosted

### NetBox Version

v4.2.5

### Python Version

3.11

### Steps to Reproduce

1. Create three regions without parent: "North America", "South America", "Europe"
2. Create three child regions under North America: "United States", "Canada", "Mexico"
3. Create three grandchildren within United States: "Wisconsin", "Oklahoma", "Tennessee"
4. Navigate to /dcim/regions/ to view all regions

### Expected Behavior

The three child regions that were created within North America are displayed directly below NA. Additionally, "Oklahoma", "Wisconsin" and "Tennessee" are indented under the US.

### Observed Behavior

Child regions "Canada" and "Mexico" are below "Europe". Grandchild "Oklahoma" is now in "Mexico".

Image

Note that this might not be the only list view that appears to be broken. Replicating the above steps as contact groups behaves in a similar way:

Image

bctiemann:
### Fixes: #18863

Because MPTT-based models rely on an implicit hierarchical ordering behavior, the central ordering of querysets of these models as introduced in https://github.com/netbox-community/netbox/issues/18729 causes its own unpredictability when using list views (which should preserve the hierarchical presentation). This includes models such as `Region`, `ContactGroup`, `TenantGroup`, etc.

This change identifies these models through checking `model._meta.local_managers` and exempts them from having `order_by` reapplied during list views in the API and UI.

### DIFF
--- a/netbox/netbox/api/viewsets/__init__.py
+++ b/netbox/netbox/api/viewsets/__init__.py
@@ -12,7 +12,7 @@ from rest_framework.viewsets import GenericViewSet
 
 from utilities.api import get_annotations_for_serializer, get_prefetches_for_serializer
 from utilities.exceptions import AbortRequest
-from utilities.mptt import TreeManager
+from utilities.query import reapply_model_ordering
 from . import mixins
 
 __all__ = (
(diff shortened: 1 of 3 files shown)


arthanson (commented):
As the code is getting more complicated, can this be put into a Mixin that is included in both places so we centralize the code?

bctiemann:
Because `get_queryset` takes a `request` in one place and not the other, and because the classes both have long inheritance trees already and the MRO causes a new mixin to just make things more confusing, I ended up deciding a util function was the best place for this logic to live. Open to suggestions of better organizational ideas, but at least now the redundancy is minimized.

arthanson (approved):

PR #18874, opened 2025-03-11
Fixes: #18833 Inventory Item Bulk Import - 'InventoryItemImportForm' has no field named 'component_id'.

### LINKED ISSUE #18833: Inventory Item Bulk Import - 'InventoryItemImportForm' has no field named 'component_id'.
### Deployment Type

Self-hosted

### NetBox Version

v4.2.4

### Python Version

3.10

### Steps to Reproduce

1. Enter Data in Direct Import tab on Inventory Item Bulk Import page
2. Click Submit

Data is in CSV format with headers:

device	name	status	role	part_id	serial	description	component_type	component_name

### Expected Behavior

Inventory items should have been imported

### Observed Behavior

Server Error

<class 'ValueError'>

'InventoryItemImportForm' has no field named 'component_id'.

Python version: 3.10.12
NetBox version: 4.2.4
Plugins: 
  netbox_topology_views: 4.2.0

renatoalmeidaoliveira:
### Root Cause:

In Django's `BaseModelForm` `_post_clean` method, an instance of the model is created with data from `self.cleaned_data`. After the model instantiation, the `full_clean()` method is executed on that model instance.  
Since `component_type` is a field of the `InventoryItem` model, the execution of `full_clean()` on the instance was failing without being captured by any form validation method.

### Solution:

In Django's form validation process, the `clean_<fieldname>()` methods are executed before the `clean_form()` method. Inside the `clean_<fieldname>()` method, there are no guarantees about which field names have already been processed and is available at `self.cleaned_data`.And to fully validate `component_type` use cases, the `device`, `component_type`, and `component_name` fields must be available.  

To fix the validation process, the `clean_component_name` method was replaced by the `clean` method, and the fields `component_name` and `component_type` were removed from `cleaned_data` when incorrect data was provided.

### DIFF
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -1164,8 +1164,8 @@ class InventoryItemImportForm(NetBoxModelImportForm):
     def clean(self):
         super().clean()
         cleaned_data = self.cleaned_data
-        component_type = cleaned_data.get('component_type', None)
-        component_name = cleaned_data.get('component_name', None)
+        component_type = cleaned_data.get('component_type')
+        component_name = cleaned_data.get('component_name')
         device = self.cleaned_data.get("device")
 
         if component_type:
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -1161,27 +1161,35 @@ class InventoryItemImportForm(NetBoxModelImportForm):
         else:
             self.fields['parent'].queryset = InventoryItem.objects.none()
 
-    def clean_component_name(self):
-        content_type = self.cleaned_data.get('component_type')
-        component_name = self.cleaned_data.get('component_name')
+    def clean(self):
+        cleaned_data = self.cleaned_data
+        content_type = cleaned_data.get('component_type')
+        component_name = cleaned_data.get('component_name')
         device = self.cleaned_data.get("device")
 
-        if not device and hasattr(self, 'instance') and hasattr(self.instance, 'device'):
-            device = self.instance.device
-
-        if not all([device, content_type, component_name]):
-            return None
-
-        model = content_type.model_class()
-        try:
-            component = model.objects.get(device=device, name=component_name)
-            self.instance.component = component
-        except ObjectDoesNotExist:
-            raise forms.ValidationError(
-                _("Component not found: {device} - {component_name}").format(
-                    device=device, component_name=component_name
+        if content_type:
+            if device is None:
+                cleaned_data.pop('component_type')
+            if component_name is None:
+                cleaned_data.pop('component_type')
+                raise forms.ValidationError(
+                    _("Component name must be specified when component type is specified")
                 )
-            )
+            if all([device, component_name]):
+                try:
+                    model = content_type.model_class()
+                    component = model.objects.get(device=device, name=component_name)
+                    self.instance.component = component
+                except ObjectDoesNotExist:
+                    cleaned_data.pop('component_type')
+                    cleaned_data.pop('component_name')
+                    raise forms.ValidationError(
+                        _("Component not found: {device} - {component_name}").format(
+                            device=device, component_name=component_name
+                        )
+                    )
+
+        return cleaned_data
 
 
 #
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -1162,12 +1162,13 @@ class InventoryItemImportForm(NetBoxModelImportForm):
             self.fields['parent'].queryset = InventoryItem.objects.none()
 
     def clean(self):
+        super().clean()
         cleaned_data = self.cleaned_data
-        content_type = cleaned_data.get('component_type')
+        component_type = cleaned_data.get('component_type')
         component_name = cleaned_data.get('component_name')
         device = self.cleaned_data.get("device")
 
-        if content_type:
+        if component_type:
             if device is None:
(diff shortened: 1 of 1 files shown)


netbox/dcim/forms/bulk_import.py
```diff
@@ -1161,27 +1161,35 @@ def __init__(self, *args, **kwargs):
         else:
             self.fields['parent'].queryset = InventoryItem.objects.none()
 
-    def clean_component_name(self):
-        content_type = self.cleaned_data.get('component_type')
-        component_name = self.cleaned_data.get('component_name')
+    def clean(self):
```
jeremystretch, line 1164: Need to call `super()` when overriding `clean()` on a form.
```suggestion
    def clean(self):
        super().clean()
```

netbox/dcim/forms/bulk_import.py
```diff
@@ -1161,27 +1161,35 @@ def __init__(self, *args, **kwargs):
...
-
-        model = content_type.model_class()
-        try:
-            component = model.objects.get(device=device, name=component_name)
-            self.instance.component = component
-        except ObjectDoesNotExist:
-            raise forms.ValidationError(
-                _("Component not found: {device} - {component_name}").format(
-                    device=device, component_name=component_name
+        if content_type:
```
jeremystretch, line 1170: We should return a validation error if `component_name` is specified but `component_type` is not.

netbox/dcim/forms/bulk_import.py
```diff
@@ -1161,27 +1161,35 @@ def __init__(self, *args, **kwargs):
         else:
             self.fields['parent'].queryset = InventoryItem.objects.none()
 
-    def clean_component_name(self):
-        content_type = self.cleaned_data.get('component_type')
-        component_name = self.cleaned_data.get('component_name')
+    def clean(self):
+        cleaned_data = self.cleaned_data
+        content_type = cleaned_data.get('component_type')
+        component_name = cleaned_data.get('component_name')
```
jeremystretch, line 1167:
```suggestion
        content_type = self.cleaned_data.get('component_type')
        component_name = self.cleaned_data.get('component_name')
```

netbox/dcim/forms/bulk_import.py
```diff
@@ -1161,27 +1161,35 @@ def __init__(self, *args, **kwargs):
         else:
             self.fields['parent'].queryset = InventoryItem.objects.none()
 
-    def clean_component_name(self):
-        content_type = self.cleaned_data.get('component_type')
-        component_name = self.cleaned_data.get('component_name')
+    def clean(self):
+        cleaned_data = self.cleaned_data
+        content_type = cleaned_data.get('component_type')
```
jeremystretch, line 1166: Suggest renaming this to `component_type` for clarity.

netbox/dcim/forms/bulk_import.py
```diff
@@ -1161,27 +1161,35 @@ def __init__(self, *args, **kwargs):
...
-        try:
-            component = model.objects.get(device=device, name=component_name)
-            self.instance.component = component
-        except ObjectDoesNotExist:
-            raise forms.ValidationError(
-                _("Component not found: {device} - {component_name}").format(
-                    device=device, component_name=component_name
+        if content_type:
+            if device is None:
+                cleaned_data.pop('component_type')
```
jeremystretch, line 1172: We generally want to avoid mutating the cleaned data. If `device` (a required field) hasn't been specified, we should be able to abort here as it will have already raised a validation error.

netbox/dcim/forms/bulk_import.py
```diff
@@ -1161,27 +1161,35 @@ def __init__(self, *args, **kwargs):
...
+                raise forms.ValidationError(
+                    _("Component name must be specified when component type is specified")
                 )
-            )
+            if all([device, component_name]):
+                try:
+                    model = content_type.model_class()
+                    component = model.objects.get(device=device, name=component_name)
+                    self.instance.component = component
```
jeremystretch, line 1182:
```suggestion
                    self.instance.component = model.objects.get(device=device, name=component_name)
```

jeremystretch (changes requested):

renatoalmeidaoliveira:
About the two remaining revisions:

When Django runs full_clean() in `BaseForm` it executes `_clean_fields`,  `_clean_form` and `_post_clean` and in both `_clean_fields` and `_clean_form` the `ValidationError` exceptions are captured, added to the form errors, and not propagated [full_clean code](https://github.com/django/django/blob/ef6a83789b310a441237a190a493c9586a4cb260/django/forms/forms.py#L324)
But inside `_post_clean` Django tries to create an instance of the form's model and run full clean in that instatiated model [_post_clean code](https://github.com/django/django/blob/ef6a83789b310a441237a190a493c9586a4cb260/django/forms/models.py#L501) 

I tryed to edit the `InventoryItem`'s `clean` method to avoid mutate the `cleaned_field` but it isn't working, even just adding a simple Exception at the beginnig of `clean` and it isn't executed, so changing that method may not work as expected
Another option I see could be changing the `InventoryItemImportForm`'s 'component_type` to something that isn't a model field

jeremystretch (changes requested):
I get a ValueError exception when attempting to import with a `component_type` but no `component_name` specified: `'InventoryItemImportForm' has no field named 'component_id'.`

Test data:

```
device,name,status,component_type
dmi01-akron-rtr01,item1,active,dcim.interface
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #18883, opened 2025-03-12
Fixes #18869: Ensure `to_meters()` always returns a clean decimal value

### LINKED ISSUE #18869: Numeric conversion utility functions do not account for floating point errors
### Deployment Type

NetBox Cloud

### NetBox Version

v4.2.5

### Python Version

3.10

### Steps to Reproduce

1. Attempt to convert feet to meters using the `to_meters()` utility function:

```python
>>> from utilities.conversion import to_meters
>>> to_meters(30, 'ft')
Decimal('9.144000000000000460964599824')
```

### Expected Behavior

The function should return the clean decimal value 9.144.

### Observed Behavior

The returned value includes a [floating point error](https://docs.python.org/3/tutorial/floatingpoint.html).

The `to_grams()` utility function is affected as well.

jeremystretch:
### Fixes: #18869

- Ensure `to_meters()` always returns a decimal value rounded to four decimal places
- Add tests for `to_meters()` & `to_grams()`

### DIFF
--- a/netbox/utilities/conversion.py
+++ b/netbox/utilities/conversion.py
@@ -37,9 +37,9 @@ def to_grams(weight, unit) -> int:
     )
 
 
-def to_meters(length, unit):
+def to_meters(length, unit) -> Decimal:
     """
-    Convert the given length to meters.
+    Convert the given length to meters, returning a Decimal value.
     """
     try:
         if length < 0:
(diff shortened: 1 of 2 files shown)


netbox/utilities/conversion.py
```diff
@@ -48,17 +48,17 @@ def to_meters(length, unit):
...
+        return round(Decimal(length / 100), 4)
     if unit == CableLengthUnitChoices.UNIT_MILE:
-        return length * Decimal(1609.344)
+        return round(length * Decimal(1609.344), 4)
     if unit == CableLengthUnitChoices.UNIT_FOOT:
-        return length * Decimal(0.3048)
+        return round(length * Decimal(0.3048), 4)
     if unit == CableLengthUnitChoices.UNIT_INCH:
-        return length * Decimal(0.0254)
+        return round(length * Decimal(0.0254), 4)
```
jnovinger, line 61: `to_meters` doesn't explicitly say `length` has to be an `int` (although I suppose it might be enforced earlier by the caller), just a "number", so I expect someone will enter a `float` at some point.

If incoming the `length` is a float, then mile, foot, and inch fail with `TypeError: unsupported operand ...`. The following tests capture it:

```python
        self.assertEqual(
            to_meters(1.0, CableLengthUnitChoices.UNIT_MILE), Decimal('1609.344')
        )
        self.assertEqual(
            to_meters(1.0, CableLengthUnitChoices.UNIT_FOOT), Decimal('0.3048')
        )
        self.assertEqual(
            to_meters(1.0, CableLengthUnitChoices.UNIT_INCH), Decimal('0.0254')
        )
```

jnovinger (changes requested):
Let me know if "length won't always be an int" is a bad assumption. This looks good otherwise.

netbox/utilities/conversion.py
jnovinger: Perhaps it makes sense to just cast `length` to a `Decimal` as soon as the early validation is done.

jeremystretch: Good catch!

jnovinger (approved):
💥 Just waiting on CI to finish

Issue #17170, opened 2024-08-15
Add ability to add contacts to multiple contact groups


arosemanhellmann:
### NetBox version

v4.0.5

### Feature type

Data model extension

### Proposed functionality

Currently it's only possible to add a contact (person) to one contact group. In our organiszation this does not work out well, as we have only few persons where this 1:1 relation would work. Most of our IT Staff works in multiple (contact)groups. As what i know, in the admin part it's possible (access groups) it's possible to add persons to multipl groups. 

It would be nice to have this abbility also in the contact groups. 

### Use case

Build the source organizational structure in the contacts / contact groups, what currently is not possible for us.

goteamkor:
literally asked for the same thing 5 days ago.
https://github.com/netbox-community/netbox/issues/17123

arosemanhellmann:
https://github.com/netbox-community/netbox/issues/17123 does also focus on sites and site groups / not sure if this is tecnically the same. I'm only focusing on Contacts and Contact Groups. Just wanted to let you know. As what i read above you're planning a change here.

goteamkor:
> #17123 does also focus on sites and site groups / not sure if this is tecnically the same. I'm only focusing on Contacts and Contact Groups. Just wanted to let you know. As what i read above you're planning a change here.

I included site groups because the overall concept is the same. The concept is allowing an object to be part of multiple groups of that objects group type.

sites in multiple site groups
contacts in multiple contact groups

the point was consistency. Especially since the documentation outlines site groups as 'role or function' without recognizing that a site or a contact can serve multiple roles or functions. I outlined some examples in https://github.com/netbox-community/netbox/issues/17123.

While sites in multiple site groups would be handy too, if I had to choose I'd choose contacts being in multiple contact groups.
I also think implementing this would flow nicely into my other FR https://github.com/netbox-community/netbox/issues/17050 which would see the ability to assign contact groups to objects instead of just contacts.

This way, if you are able to place a contact into multiple groups, and then assign contact groups to objects you can much more easily reflect how most companies actually work. (switches having a team assigned as a technical contact, etc.)

Leon-Pay:
Contacts to multiple contact groups would be a great addition

PR #18885, opened 2025-03-12
17170 Add ability to add contacts to multiple contact groups

arthanson:
### Fixes: #17170 

Changed Contacts Group field to be M2M field so a contact can belong to multiple contact groups.  There were some issues as ContactGroup is a nested model (so using TreeNode) which I don't think we have done before (allowing multiple assignments of a TreeNode field).  

This caused an issue in utilities.testing.fitersets.py as it was expecting ModelMultipleChoiceField instead of TreeNodeModelMultipleChoiceField (which is derived from ModelMultipleChoiceField) so changed the assert to assertIsInstance: https://github.com/netbox-community/netbox/pull/18885/files#diff-27b47272104ca5b929872465d014bb392aacf97b3fc957695d85ebaa82171e40

Changed group references to groups as it is plural now.

ContactFilterSet was throwing a missing filter error when using group_id, group, needed to change it to contact_group_id, contact_group. Not completely sure why that was complaining with the change.


netbox/tenancy/filtersets.py
```diff
@@ -60,17 +65,17 @@ class Meta:
 
 
 class ContactFilterSet(NetBoxModelFilterSet):
-    group_id = TreeNodeMultipleChoiceFilter(
+    contact_group_id = TreeNodeMultipleChoiceFilter(
```
jeremystretch, line 68: Leaving these as `group_id` and `group` seems to work fine AFAICT. We should avoid renaming them if possible.

netbox/tenancy/forms/filtersets.py
```diff
@@ -71,7 +71,7 @@ class ContactRoleFilterForm(NetBoxModelFilterSetForm):
 
 class ContactFilterForm(NetBoxModelFilterSetForm):
     model = Contact
-    group_id = DynamicModelMultipleChoiceField(
+    contact_group_id = DynamicModelMultipleChoiceField(
```
jeremystretch, line 74: If we keep the filter name
```suggestion
    group_id = DynamicModelMultipleChoiceField(
```

jeremystretch, line 74: Should the label of this field be "groups" (plural) now?

netbox/tenancy/forms/bulk_edit.py
```diff
@@ -90,8 +90,8 @@ class ContactRoleBulkEditForm(NetBoxModelBulkEditForm):
 
 
 class ContactBulkEditForm(NetBoxModelBulkEditForm):
-    group = DynamicModelChoiceField(
-        label=_('Group'),
+    groups = DynamicModelMultipleChoiceField(
```
jeremystretch, line 93: Because this is a many-to-many field, setting `groups` in bulk will wipe out any existing assignments. It would be ideal to include separate add & remove fields, though admittedly we don't have a great pattern for this today (aside from the assignment of tags).

netbox/tenancy/models/contacts.py
```diff
@@ -84,24 +84,28 @@ class Contact(PrimaryModel):
...
-            ),
-        )
         verbose_name = _('contact')
         verbose_name_plural = _('contacts')
 
     def __str__(self):
         return self.name
 
 
+class ContactGroupMembership(models.Model):
```
jeremystretch, line 99: We probably want to add this to `__all__` so it can be imported from `tenancy.models`.

netbox/tenancy/models/contacts.py
```diff
@@ -84,24 +84,28 @@ class Contact(PrimaryModel):
...
+    group = models.ForeignKey(ContactGroup, related_name="+", on_delete=models.CASCADE)
+    contact = models.ForeignKey(Contact, related_name="+", on_delete=models.CASCADE)
+
+    class Meta:
```
jeremystretch, line 103: Should probably add `verbose_name` and `verbose_name_plural`

netbox/tenancy/tests/test_api.py
```diff
@@ -183,20 +183,21 @@ def setUpTestData(cls):
...
         )
         Contact.objects.bulk_create(contacts)
+        contacts[0].groups.add(contact_groups[0])
+        contacts[1].groups.add(contact_groups[0])
+        contacts[2].groups.add(contact_groups[0])
 
         cls.create_data = [
             {
                 'name': 'Contact 4',
-                'group': contact_groups[1].pk,
```
jeremystretch, line 195: We should still test group assignment on creation of a contact.

netbox/tenancy/tests/test_views.py
```diff
@@ -196,26 +196,28 @@ def setUpTestData(cls):
...
             'comments': 'Some comments',
             'tags': [t.pk for t in tags],
         }
 
         cls.csv_data = (
-            "group,name",
-            "Contact Group 1,Contact 4",
-            "Contact Group 1,Contact 5",
-            "Contact Group 1,Contact 6",
+            "name",
```
jeremystretch, line 217: Need to test group assignment here as well. (It works fine AFAICT.)

netbox/tenancy/views.py
```diff
@@ -183,12 +183,14 @@ class ContactGroupListView(generic.ObjectListView):
...
     queryset = ContactGroup.objects.all()
 
+    """
     def get_extra_context(self, request, instance):
         groups = instance.get_descendants(include_self=True)
 
         return {
             'related_models': self.get_related_models(request, groups),
         }
+    """
```
jeremystretch, line 193: Was this intended to be omitted?

jeremystretch (changes requested):

netbox/tenancy/migrations/0018_contact_groups.py
```diff
@@ -0,0 +1,94 @@
...
+from django.db import migrations, models
+
+
+def migrate_contact_groups(apps, schema_editor):
+    Contacts = apps.get_model('tenancy', 'Contact')
+
+    qs = Contacts.objects.filter(group__isnull=False)
+    for contact in qs:
+        contact.groups.add(contact.group)
+        contact.save()
```
jeremystretch, line 13: I don't think we need to call `save()` on the instance if we're just updating an M2M assignment.

netbox/tenancy/migrations/0018_contact_groups.py
```diff
@@ -0,0 +1,93 @@
...
+            model_name='contact',
+            name='group',
+        ),
+        migrations.AlterField(
+            model_name='contact',
+            name='groups',
+            field=models.ManyToManyField(
+                blank=True, related_name='contacts', through='tenancy.ContactGroupMembership', to='tenancy.contactgroup'
+            ),
+        ),
```
jeremystretch, line 67: This can be merged into the operation on line 42.

netbox/tenancy/migrations/0018_contact_groups.py
```diff
@@ -0,0 +1,93 @@
...
+        migrations.AlterField(
+            model_name='contactgroupmembership',
+            name='contact',
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.CASCADE, related_name='+', to='tenancy.contact'
+            ),
+        ),
```
jeremystretch, line 74: This can be merged into the operation on line 32.

netbox/tenancy/migrations/0018_contact_groups.py
```diff
@@ -0,0 +1,93 @@
...
+        migrations.AlterField(
+            model_name='contactgroupmembership',
+            name='group',
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.CASCADE, related_name='+', to='tenancy.contactgroup'
+            ),
+        ),
```
jeremystretch, line 81: This can be merged into the operation on line 37.

netbox/tenancy/migrations/0018_contact_groups.py
```diff
@@ -0,0 +1,93 @@
...
+            model_name='contact',
+            name='groups',
+            field=models.ManyToManyField(
+                blank=True,
+                related_name='contacts',
+                related_query_name='contact',
+                through='tenancy.ContactGroupMembership',
+                to='tenancy.contactgroup',
+            ),
+        ),
```
jeremystretch, line 92: This can be merged into the operation on line 42.

netbox/tenancy/migrations/0018_contact_groups.py
```diff
@@ -0,0 +1,93 @@
+# Generated by Django 5.1.5 on 2025-03-11 20:27
+
```
jeremystretch, line 2:
```suggestion
```

netbox/tenancy/forms/bulk_edit.py
```diff
@@ -127,9 +132,11 @@ class ContactBulkEditForm(NetBoxModelBulkEditForm):
 
     model = Contact
     fieldsets = (
-        FieldSet('group', 'title', 'phone', 'email', 'address', 'link', 'description'),
+        FieldSet('add_groups', 'remove_groups', 'title', 'phone', 'email', 'address', 'link', 'description'),
```
jeremystretch, line 135: Want to move the add/remove fields under a separate "Groups" heading in the form?

netbox/tenancy/tests/test_views.py
jeremystretch: Looks like a typo: This creates a contact named "groups."

jeremystretch (changes requested):
Found a few minor cleanup items, but other than those this looks good to go!

jeremystretch (approved):

PR #18889, opened 2025-03-13
Fixes: #18783 Add a tag_id filter for all models which support tagging 

### LINKED ISSUE #18783: Add a `tag_id` filter for all models which support tagging
### NetBox version

v4.2.4

### Feature type

Data model extension

### Proposed functionality

Add a `tag_id` filter to NetBoxModelFilterSet, which will enable filtering of objects by tag ID. This should function similarly to the existing `tag` filter (which matches by tag slug). For example:

`GET /api/dcim/sites/?tag_id=100&tag_id=200`

### Use case

This provides an alternative to referencing tags by their slugs, and may be needed in some specific scenarios (such as that described in #18757).

### Database changes

N/A

### External dependencies

N/A

renatoalmeidaoliveira:
### Fixes: #18783 Add a tag_id filter for all models which support tagging 

Replicate `TagFilter` logic to `TagIDFilter` and implemented `tag_id` field to `NetBoxModelFilterSet` and  `ConfigTemplateFilterSet`

Tested the FilterSet in REST,  GraphQL, and in the UI by crafting an URL.

Since there already a Tag field in the FilterForm didn't implemented any new field.

### DIFF
--- a/docs/plugins/development/filtersets.md
+++ b/docs/plugins/development/filtersets.md
@@ -60,6 +60,10 @@ class MyModelViewSet(...):
 ### TagFilter
 
 The `TagFilter` class is available for all models which support tag assignment (those which inherit from `NetBoxModel` or `TagsMixin`). This filter subclasses django-filter's `ModelMultipleChoiceFilter` to work with NetBox's `TaggedItem` class.
+This class filters `tags` using the `slug` field. For example:
+
+`GET /api/dcim/sites/?tag=alpha&tag_id=bravo`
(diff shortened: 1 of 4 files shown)


arthanson (changes requested):
This looks very good, the only thing is add it to docs/plugins/development/filtersets.py under TagFilter heading.

arthanson (changes requested):

arthanson (approved):

PR #18930, opened 2025-03-17
Closes #17841 Allows Tags to be displayed in specified order

### LINKED ISSUE #17841: Allow `Tags` to be displayed in a specific order
### NetBox version

v4.1.0

### Feature type

New functionality

### Triage priority

N/A

### Proposed functionality

Add an integer `Index` field to the `Tag` object definition which can be used by the frontend to order tags first by `Index`, then alphabetically when displaying them.

- `Index` ordering should be from lowest to highest value, where the lowest value has the highest priority and therefore it is displayed before any other tags
- Some sort of hard limit, e.g. `-128` to `128`, could be implemented for optimization
- The default index value must be zero
- **For tags with the same index, alphabetical sorting should be the default behavior.** This is to ensure that, when migrating to this implementation existing behavior is preserved.

### Use case

There are innumerous use cases for tags, one such case is the ability to give information about an object such as a virtual machine:

`NETPY` `CLT01` `VM108` `APCHE`
```
 |     |      |      |_ Gives a clear visual hint of the service running in this machine
 |     |      |________ Shows the virtual machine name according to our virtualization system
 |     |_______________ Identifies the cluster where this machine belongs to
 |_____________________ Identifies the tenant
```

It would be lovely to display the four tags in the aforementioned order, since that is the formal name of this object and everywhere in our internal processes and documents, we refer to this virtual machine as `NETPY-CLT01-VM108-APCHE`.

Having the ability to sort tags when displaying objects in Netbox will allow for further customization and integration with other systems that rely on tags for looking up objects.

### Database changes

A new `integer` column for the `Tag` table definition.

### External dependencies

None that I am aware of.

jnovinger:
### Fixes: #17841

- Adds `Tag.weight` field so users can specify an ordering weight. Valid values are 0-32767.
- Sets default `Tag` ordering to `('weight', 'name')` so that weights are used for ordering first (lower weights get sorted first) and then names for tie breakers.
- Sets default `TaggingMixin.tags` ordering to `('weight', 'name')` so that weights are used for ordering first (lower weights get sorted first) and then names for tie breakers. **Note:** this was required so that calls through an objects `.tags` related manager would be sorted when called from a template (e.g. so that ordering would work for things like `{% for tag in Tag.objects.all %}`)
- Updates existing tests and adds tests for default model ordering and related manager ordering defaults.
- Adds `Tag.weight` in all the required places.


arthanson (approved):
Good work @jnovinger everything looks good.

PR #18932, opened 2025-03-17
Fixes: #18881 Site Groups are missing VLAN and VM related objects

### LINKED ISSUE #18881: Site Groups are missing VLAN and VM related objects
### Deployment Type

Self-hosted

### NetBox Version

v.4.2.4

### Python Version

3.10

### Steps to Reproduce

1. Create a Site Group "Example Site Group"
2. Create a VLAN group "Example VLAN Group" with scope "DCIM > Site Group"
3. Assign the VLAN group to the "Example Site Group"
4. Create the VLAN group
5. Go to Site Groups
6. See that there is no reference to any VLANs or VLAN groups even though it's literally the only parent object that has been selected for the created VLAN group

### Expected Behavior

Site groups should reference the same related objects as Sites, and not just a subset of them. That way it would be not only useful but also possible to see all:
- VLANs
- VLAN Groups
- Virtual Machines
- ASNs
- Clusters
- Devices

that are present within a Site Group.

### Observed Behavior

Here's a screenshot of a `Site's` related objects, that is part of a Site Group:

Image

Note the highlighted related objects.

Now, here's a screenshot of a `Site Group's` (where the above Site is a member of) related objects:

Image

Notice the following objects mising:
- VLANs
- VLAN Groups
- Virtual Machines
- ASNs
- Clusters
- Devices

renatoalmeidaoliveira:
### Fixes: #18881 Site Groups are missing VLAN and VM related objects

Setup Related Models tab for scoped models:

- [x] sitegroup
- [x] region
- [x] site
- [x] location
- [x] rack
- [x] clustergroup
- [x] cluster

Add models to `SiteGroupView`'s `get_extra_context`:

- [x] VLANs
- [x] VLAN Groups
- [x] Virtual Machines
- [x] ASNs
- [x] Clusters
- [x] Devices

Add models to `RegionView`'s `get_extra_context`:

- [x] VLANGroup

Add models to `RackView`'s `get_extra_context`:

- [x] VLANGroup

Add models to `ClusterGroupView`'s `get_extra_context`:

- [x] VLANGroup

- [x] Add `Circuit Terminations` Relations


renatoalmeidaoliveira:
Waiting for #18933 and #18939

renatoalmeidaoliveira:
The Cluster seens to be visible in the current version of the code. To list the `Cluster` you need to assign it to the `Site Group` scope.

image

arthanson (changes requested):
This looks good, but can you check / extend it to other scope objects - it looks like the same thing is happening for regions and racks (may be more, those are the ones I spot checked).  Do the same repro steps just assign "Example VLAN Group" to a region instead of a Site Group, then check the region assigned to and it won't be in the related objects list.

renatoalmeidaoliveira:
@corubba thanks for your observations

arthanson (approved):

PR #18946, opened 2025-03-18
Fixes #18944: Clearing widget type field no longer causes 500 error

### LINKED ISSUE #18944: Clearing "Widget type" field produces a ValueError: Unregistered widget class:
### Deployment Type

Self-hosted

### NetBox Version

v4.2.5

### Python Version

3.10

### Steps to Reproduce

1. On the Dashboard, unlock and click "Add Widget"
2. On the popup dialog, click the X to clear the "Widget type" field (or backspace over the preselected value)
3. App crashes with the below observed error

### Expected Behavior

App should not crash; form should not trigger any dynamic parsing that tries to validate an empty input value

### Observed Behavior

```
ValueError at /extras/dashboard/widgets/add/
Unregistered widget class:
Request URL: http://127.0.0.1:8000/extras/dashboard/widgets/add/?widget_class=
```

jnovinger:
### Fixes: #18944 

I don't love this fix since the feedback to the user is pretty bad--it just resets the form to defaults without explaining what happened. I didn't find an easy way to include an error message for the 'Widget type' field on GET (so that it might show in the UI) but that may just be me missing something.

I also took a quick look at our `DynamicTomSelect` code and didn't see an immediate way to fix this there.

Either way, I think it's worth getting at least this fix in for the time being to remove a potentially very visible 500 error for users with better user feedback in a follow up fix. But, I'm open to other options. Thoughts?

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -1103,7 +1103,7 @@ class DashboardWidgetAddView(LoginRequiredMixin, View):
         }
         widget_form = DashboardWidgetAddForm(initial=initial)
         widget_name = get_field_value(widget_form, 'widget_class')
-        widget_class = get_widget_class(widget_name)
+        widget_class = get_widget_class(widget_name or 'extras.NoteWidget')
         config_form = widget_class.ConfigForm(initial=widget_class.default_config, prefix='config')
 
         return render(request, self.template_name, {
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
I think this is the right approach, but we can clean it up a bit by just ensuring that the initial value for the `widget_class` form field is always set to `'extras.NoteWidget'` if not defined. We could change lines 1101-1103 to something like:

```python
initial = {
    'widget_class': request.GET.get('widget_class') or 'extras.NoteWidget',
}
```

FWIW we should probably eliminate the "clear" button on required choice fields; I'll open a separate issue for that. (Edit: #18955)

jeremystretch (approved):
Thanks @jnovinger!

PR #18947, opened 2025-03-19
Fixes: #18933 VLANGroupFilterSet site_group field doesn't match VLANGroupFilterForm sitegroup field

### LINKED ISSUE #18933: VLANGroupFilterSet site_group field doesn't match VLANGroupFilterForm  sitegroup field
### Deployment Type

Self-hosted

### NetBox Version

v4.2.5

### Python Version

3.10

### Steps to Reproduce

1. In `VLANGroupsListView` filter by `Site group`

### Expected Behavior

The queryset filter by site_group

### Observed Behavior

The queryset is filtering by sitegroup that doesn't have a FilterSet Field

renatoalmeidaoliveira:
### Fixes: #18933 VLANGroupFilterSet site_group field doesn't match VLANGroupFilterForm sitegroup field

Fixes `site_group` field typo, it was defined as `sitegroup` in the `FilterForm` and as `site_group` in the `FilterSet`
Followed `site_group` pattern as it was already used in other places like:
`ProviderFilterSet`, `CircuitTerminationFilterSet`, `LocationFilterSet` and several others

### DIFF
--- a/netbox/ipam/forms/filtersets.py
+++ b/netbox/ipam/forms/filtersets.py
@@ -418,7 +418,7 @@ class FHRPGroupFilterForm(NetBoxModelFilterSetForm):
 class VLANGroupFilterForm(NetBoxModelFilterSetForm):
     fieldsets = (
         FieldSet('q', 'filter_id', 'tag'),
-        FieldSet('region', 'sitegroup', 'site', 'location', 'rack', name=_('Location')),
+        FieldSet('region', 'site_group', 'site', 'location', 'rack', name=_('Location')),
         FieldSet('cluster_group', 'cluster', name=_('Cluster')),
         FieldSet('contains_vid', name=_('VLANs')),
     )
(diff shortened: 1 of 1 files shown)


jnovinger:
@renatoalmeidaoliveira , just to be clear, this fix should only enable filtering `VLANGroup`s with a scope directly tied to a specific `SiteGroup`, correct?

That is, it should not also be filtering on site groups via a link between a VLAN group and a particular site that _is part of a site group_, right?

renatoalmeidaoliveira:
@jnovinger, in the current implementation the `site_group` is filtering only for VlanGroups that has some `SiteGroup` as scope.
In other models, that has a site attribut, filtering by `site_group` filters all models that has a site in some `site_group`.

renatoalmeidaoliveira:
But since you can assign scopes, and a SiteGroup is a valid scope for VlanGroup model, IMO, filtering for both sites, and SiteGroup may not be the best option, but please let me know if you prefer to change the FilterSet to include sites that belongs to that SiteGroup too.
I think that culd be possible by adding a `Q` with an `OR` in the queryset.

jnovinger:
Nope, not necessarily a preference. Just wanted to make sure I understood the scope of this change correctly! I didn't want to kick it back if it wasn't part of the scope. Thanks!

jnovinger (approved):

PR #18950, opened 2025-03-19
Fixes: #18656 Unable to import IP Address and assign to FHRP Group

### LINKED ISSUE #18656: Unable to import IP Address and assign to FHRP Group
### NetBox version

v4.2.2

### Feature type

New functionality

### Proposed functionality

When importing an IP Address there does not appear to be a way to assign it to an FHRP Group during import.  Looking at the [bulk_import.py](https://raw.githubusercontent.com/netbox-community/netbox/refs/heads/main/netbox/ipam/forms/bulk_import.py) file for ipam the IPAddressImportForm class does not appear to have the ability to tie the imported IP to an FHRP Group.

### Use case

Allow importing IP Addresses and assigning them to an FHRP Group just like you would an interface on a device or vm.

### Database changes

None

### External dependencies

None

renatoalmeidaoliveira:
### Fixes: #18656 Unable to import IP Address and assign to FHRP Group

- Created the `fhrpgroup` field in `IPAddressImportForm`

#### Observations:

Since [FHRPGroup Model](https://github.com/netbox-community/netbox/blob/7c152e92340779a889bfe5ebec7d6c97c230fc72/netbox/ipam/models/fhrp.py#L54) doesn't implement uniqueness method, choosing the appropriated accessor for the `fhrpgroup` its a bit tricky.
I choose `name` instead of the `group_id` beacuse, IMO, `group_id` is more likely to get repeated, but maybe a more accurate approach could be open a new Issue to setup some uniqueness criteria for FHRPGroup, make the import form following that pattern.

### DIFF
--- a/netbox/ipam/forms/bulk_import.py
+++ b/netbox/ipam/forms/bulk_import.py
@@ -328,11 +328,11 @@ class IPAddressImportForm(NetBoxModelImportForm):
         help_text=_('Assigned interface')
     )
     fhrpgroup = CSVModelChoiceField(
-        label=_('FHRP Group (ID)'),
+        label=_('FHRP Group'),
         queryset=FHRPGroup.objects.all(),
         required=False,
-        to_field_name='group_id',
-        help_text=_('Assigned FHRP Group ID')
+        to_field_name='name',
+        help_text=_('Assigned FHRP Group name')
     )
     is_primary = forms.BooleanField(
         label=_('Is primary'),
--- a/netbox/ipam/forms/bulk_import.py
+++ b/netbox/ipam/forms/bulk_import.py
@@ -327,6 +327,13 @@ class IPAddressImportForm(NetBoxModelImportForm):
         to_field_name='name',
         help_text=_('Assigned interface')
     )
+    fhrpgroup = CSVModelChoiceField(
+        label=_('FHRP Group (ID)'),
+        queryset=FHRPGroup.objects.all(),
+        required=False,
+        to_field_name='group_id',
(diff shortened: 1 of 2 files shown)


jnovinger (changes requested):
One minor change requested, plus please add this new import field to `cls.csv_data` for the tests in `ipam.tests.test_view.py::IPAddressTestCase.setUpTestData`.

netbox/ipam/forms/bulk_import.py
```diff
@@ -327,6 +327,13 @@ class IPAddressImportForm(NetBoxModelImportForm):
         to_field_name='name',
         help_text=_('Assigned interface')
     )
+    fhrpgroup = CSVModelChoiceField(
```
jnovinger, line 330: Sorry, not sure how my previous comments got dropped, but they did.

Let's change this to `fhrp_group` (here and below) to maintain naming consistency.

renatoalmeidaoliveira:
@jnovinger about the testing cases, I updated the `cls.csv_data` with `fhrp_group` but wasn't able to update `csv_update_data`, because `test_bulk_update_objects_with_permission` lookup for model fields based in the header names, since `IPAddress` doesn't have a field `fhrp_group` that test fails  [code](https://github.com/netbox-community/netbox/blob/b1d014b5206494cd94c058f6e365bea63eca2eba/netbox/utilities/testing/views.py#L661)

jnovinger (approved):
Looks great, @renatoalmeidaoliveira !

PR #18971, opened 2025-03-20
Fixes #18955: Include the "clear" button only for non-required dropdown fields

### LINKED ISSUE #18955: Dropdown widgets for required fields should not include a "clear" button
### Deployment Type

NetBox Cloud

### NetBox Version

v4.2.5

### Python Version

3.10

### Steps to Reproduce

In many forms (the site edit form, for instance), required choice fields include a "clear" button (an X within a circle) toward the righthand side of the field which clears any selected value.

Image

### Expected Behavior

It should not be possible to clear the selected value of a field for which a value must be specified. The presence of the "clear" button may be confusing to the user.

The clear button is a [Tom Select plugin](https://tom-select.js.org/plugins/clear-button/). It can be disabled for required fields by altering the logic [here](https://github.com/netbox-community/netbox/blob/43840e6a72416a5e0c963b230f9471308a29e3e1/netbox/project-static/src/select/config.ts#L8-L12).

### Observed Behavior

See above.

jeremystretch:
### Fixes: #18955

- Enable the `clear_button` Tom-Select plugin only for fields without `required` set


jeremystretch:
Need to determine why we're not setting the `required` attribute on required `<select>` fields.

jeremystretch:
> Need to determine why we're not setting the `required` attribute on required `<select>` fields.

AFAICT it's just not included in the template. We do use a custom template for the `<select>` widget, but it's not in Django's native template either. Feels like it should be.

bctiemann (approved):

PR #18974, opened 2025-03-21
Fixes #18949: Add missing GraphQL ContactsMixin in types with ContactAssignments

### LINKED ISSUE #18949: The TenantType in GraphQL is missing the contacts property
### Deployment Type

Self-hosted

### NetBox Version

v4.2.5

### Python Version

3.10

### Steps to Reproduce

1. Create a tenant (assuming ID 1)
2. Create a contact (assuming with name "John")
3. Add a contact assignment to the tenant
4. Execute the GraphQL query:
```graphql
query {
  tenant(id: 1) {
    contacts {
      contact {
        name
      }
    }
  }
}
```

### Expected Behavior

Query returns:
```json
{
  "data": {
    "tenant": {
      "contacts": [
        {
          "contact": {
            "name": "John"
          }
        }
      ]
    }
  }
}
```

### Observed Behavior

Query returns contacts on TenantType does not exist:
```json
{
  "data": null,
  "errors": [
    {
      "message": "Cannot query field 'contacts' on type 'TenantType'. Did you mean 'comments'?",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ]
    }
  ]
}
```

YiPrograms:
### Fixes: #18949

* Add missing `ContactMixin` in GraphQL for all the types that have `ContactMixin` in `models/` but not in `graphql/types.py`.

### DIFF
--- a/netbox/circuits/graphql/types.py
+++ b/netbox/circuits/graphql/types.py
@@ -43,7 +43,7 @@ class ProviderType(NetBoxObjectType, ContactsMixin):
     fields='__all__',
     filters=ProviderAccountFilter
 )
-class ProviderAccountType(NetBoxObjectType):
+class ProviderAccountType(ContactsMixin, NetBoxObjectType):
     provider: Annotated["ProviderType", strawberry.lazy('circuits.graphql.types')]
 
     circuits: List[Annotated["CircuitType", strawberry.lazy('circuits.graphql.types')]]
--- a/netbox/ipam/graphql/types.py
+++ b/netbox/ipam/graphql/types.py
@@ -5,6 +5,7 @@ import strawberry_django
 
 from circuits.graphql.types import ProviderType
 from dcim.graphql.types import SiteType
+from extras.graphql.mixins import ContactsMixin
 from ipam import models
 from netbox.graphql.scalars import BigInt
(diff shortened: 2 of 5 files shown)


jnovinger (dismissed):
Thanks @YiPrograms !

jnovinger:
General note for when this causes a minor merge conflict when trying to incorporate this change in to the `feature` branch for 4.3, there is only a very slight conflict re: `tenancy.graphql.types.TenantType.contacts` that was introduced in #18701.

That change introduces the `TenantType.contacts` attribute directly to the type rather than using the mixin. It makes sense that we should re-use the mixin instead of a direct attribute definition.

jnovinger:
> Sorry, i've found that the import is not in alphabetical order. Updated it now.

Much appreciated, I completely missed that. Was a little surprised when the review got dismissed as I was adding a related comment!

YiPrograms:
> > Sorry, i've found that the import is not in alphabetical order. Updated it now.
> 
> Much appreciated, I completely missed that. Was a little surprised when the review got dismissed as I was adding a related comment!

What a coincidence!
Sorry about that, I should have caught it earlier before you started reviewing.

jnovinger (approved):

PR #18983, opened 2025-03-21
Closes #18980: Optimize update of object data when adding/removing custom fields

### LINKED ISSUE #18980: Optimize bulk updates of custom field values when custom fields are added/removed
### NetBox version

v4.2.5

### Feature type

### Proposed functionality

This FR intends to capture an opportunity to improve performance when a custom field is added to or removed from a model which has many thousands of records. Whenever a custom field is added, its `populate_initial_data()` method is called to automatically populate the field's default value on all objects:

https://github.com/netbox-community/netbox/blob/7db0765ed2bf7098bab17cb71e389ffbfde34ceb/netbox/extras/models/customfields.py#L279-L289

The current approach is suboptimal, as it requires loading each individual object into memory to effect the change. I believe we can optimize this by mutating the JSON data directly within the query:

```python
from django.db.models import F, Func, Value
from django.db.models import JSONField

def bulk_update_cf_value(model, field_name, new_value):
    model.objects.update(
        custom_field_data=Func(
            F('custom_field_data'),
            Value([field_name]),
            Value(new_value, JSONField()),
            function='jsonb_set'
        )
    )

bulk_update_cf_value(Site, "test1", "abc123")
```

This allows us to update the value of a specific key for all objects without first loading each into memory.

Similar approaches will be needed for the `remove_stale_data()` and `rename_object_data()` methods on CustomField as well.

### Use case

This should result in a dramatic reduction in the time required to bulk update custom field data, particularly when many thousands of records are affected.

### Database changes

N/A

### External dependencies

N/A

### LINKED ISSUE #19890: Introduce support for Python 3.14
### Proposed Changes

Support Python 3.14 (in addition to 3.12 and 3.13) beginning with NetBox v4.5.0.

### Justification

Python 3.14 is [scheduled for final release](https://peps.python.org/pep-0745/#release-schedule) in October 2025. Django 6.0, once released, will presumably support Python versions 3.12 through 3.14.

jeremystretch:
### Fixes: #18980

- `populate_initial_data()` now uses the PostgreSQL-native `jsonb_set()` function to set initial values on objects when creating a custom field
- `remove_stale_data()` now uses native JSON mutation to remove the keys for deleted custom fields
- `rename_object_data()` has likewise been optimized to rename custom field keys in object data using `jsonb_set()`

A quick test using a REST API call adding a custom field to ~11,000 interfaces took 0.543s with this change versus 5.556s in `main`.

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -9,6 +9,8 @@ from django.conf import settings
 from django.contrib.postgres.fields import ArrayField
 from django.core.validators import RegexValidator, ValidationError
 from django.db import models
+from django.db.models import F, Func, Value
+from django.db.models.expressions import RawSQL
 from django.urls import reverse
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
@@ -281,12 +283,15 @@ class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
         Populate initial custom field data upon either a) the creation of a new CustomField, or
         b) the assignment of an existing CustomField to new object types.
         """
+        if self.default is None:
+            # We have to convert None to a JSON null for jsonb_set()
+            value = RawSQL("'null'::jsonb", [])
+        else:
+            value = Value(self.default, models.JSONField())
         for ct in content_types:
-            model = ct.model_class()
-            instances = model.objects.exclude(**{'custom_field_data__contains': self.name})
-            for instance in instances:
-                instance.custom_field_data[self.name] = self.default
-            model.objects.bulk_update(instances, ['custom_field_data'], batch_size=100)
+            ct.model_class().objects.update(
+                custom_field_data=Func(F('custom_field_data'), Value([self.name]), value, function='jsonb_set')
+            )
 
     def remove_stale_data(self, content_types):
         """
@@ -294,11 +299,9 @@ class CustomField(CloningMixin, ExportTemplatesMixin, ChangeLoggedModel):
         no longer assigned to a model, or because it has been deleted).
         """
         for ct in content_types:
-            if model := ct.model_class():
-                instances = model.objects.filter(custom_field_data__has_key=self.name)
-                for instance in instances:
-                    del instance.custom_field_data[self.name]
-                model.objects.bulk_update(instances, ['custom_field_data'], batch_size=100)
+            ct.model_class().objects.update(
+                custom_field_data=F('custom_field_data') - self.name
+            )
 
     def rename_object_data(self, old_name, new_name):
         """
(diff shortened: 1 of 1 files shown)


netbox/extras/models/customfields.py
```diff
@@ -281,36 +283,48 @@ def populate_initial_data(self, content_types):
...
-                instance.custom_field_data[self.name] = self.default
-            model.objects.bulk_update(instances, ['custom_field_data'], batch_size=100)
+            ct.model_class().objects.update(
+                custom_field_data=Func(
+                    F('custom_field_data'),
+                    Value([self.name]),
+                    value,
+                    function='jsonb_set'
+                )
+            )
```
jnovinger, line 299: Previously, we weren't updating instances that already had a value for the `self.name` path in the custom field data field (I think) and only setting it on those that did not.

But now we're just flat out setting the default value on all records (either adding or updating, per `jsonb_set` docs), is that right? I think this is safe to do since this is only running as a result of adding new custom field. Is that the reasoning or is there something else I'm missing?

netbox/extras/models/customfields.py
```diff
@@ -281,36 +283,48 @@ def populate_initial_data(self, content_types):
...
         """
         for ct in content_types:
-            if model := ct.model_class():
-                instances = model.objects.filter(custom_field_data__has_key=self.name)
-                for instance in instances:
-                    del instance.custom_field_data[self.name]
-                model.objects.bulk_update(instances, ['custom_field_data'], batch_size=100)
+            ct.model_class().objects.update(
+                custom_field_data=F('custom_field_data') - self.name
+            )
```
jnovinger, line 309: Previously there was a check here (but oddly not in the other modified methods) to make sure the result of `ct.model_class()` wasn't falsy (I assume `None`, which I've seen locally once or twice).

Is the feeling that it's safe to assume that we'll never try to call `None.objects`?

jnovinger (commented):
Just had a couple of quick questions.

netbox/extras/models/customfields.py
jeremystretch: IIRC the exclusion filter was left over from an earlier implementation of custom fields and is no longer needed. An argument could be made for retaining it, however it's cleaner IMO to overwrite any data that happens to exist (somehow) when adding a new custom field on the model, because that's what I'd expect to happen as a user.

netbox/extras/models/customfields.py
jeremystretch: Good catch. I suspect it might be there to elegantly handle the case where a plugin model has been removed. I'm going to restore the check.

jnovinger (approved):

PR #18996, opened 2025-03-25
Fixes: #16144 GetReturnURLMixin Support for Plugin Views

### LINKED ISSUE #16144: GetReturnURLMixin Support for Plugin Views
### NetBox version

v3.7.3

### Feature type

### Proposed functionality

The automatic return URL generation for views (specifically object edit views) should be compatible with plugin views. 

Currently only core views are supported as `get_return_url()` in [netbox/utilities/views.py:137](https://github.com/netbox-community/netbox/blob/develop/netbox/utilities/views.py#L137)  does not include the "plugins:" namespace

### Use case

This would alleviate the need for specifying `default_return_url` for plugin views

### Database changes

_No response_

### External dependencies

_No response_

renatoalmeidaoliveira:
### Fixes: #16144 GetReturnURLMixin Support for Plugin Views

- Check for Plugin model in `queryset` use case of  `GetReturnURLMixin`

### DIFF
--- a/netbox/utilities/views.py
+++ b/netbox/utilities/views.py
@@ -149,13 +149,8 @@ class GetReturnURLMixin:
 
         # Attempt to dynamically resolve the list view for the object
         if hasattr(self, 'queryset'):
-            model_opts = self.queryset.model._meta
-            is_plugin = isinstance(self.queryset.model._meta.app_config, PluginConfig)
             try:
-                if is_plugin:
-                    return reverse(f'plugins:{model_opts.app_label}:{model_opts.model_name}_list')
-                else:
-                    return reverse(f'{model_opts.app_label}:{model_opts.model_name}_list')
+                return reverse(get_viewname(self.queryset.model, 'list'))
             except NoReverseMatch:
                 pass
 
--- a/netbox/utilities/views.py
+++ b/netbox/utilities/views.py
@@ -150,8 +150,12 @@ class GetReturnURLMixin:
         # Attempt to dynamically resolve the list view for the object
         if hasattr(self, 'queryset'):
             model_opts = self.queryset.model._meta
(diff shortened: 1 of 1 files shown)


netbox/utilities/views.py
```diff
@@ -150,8 +150,12 @@ def get_return_url(self, request, obj=None):
         # Attempt to dynamically resolve the list view for the object
         if hasattr(self, 'queryset'):
             model_opts = self.queryset.model._meta
+            is_plugin = isinstance(self.queryset.model._meta.app_config, PluginConfig)
```
jeremystretch, line 153:
```suggestion
            prefix = 'plugins:' if isinstance(self.queryset.model._meta.app_config, PluginConfig) else ''
```

netbox/utilities/views.py
```diff
@@ -150,8 +150,12 @@ def get_return_url(self, request, obj=None):
...
             try:
-                return reverse(f'{model_opts.app_label}:{model_opts.model_name}_list')
+                if is_plugin:
+                    return reverse(f'plugins:{model_opts.app_label}:{model_opts.model_name}_list')
+                else:
+                    return reverse(f'{model_opts.app_label}:{model_opts.model_name}_list')
```
jeremystretch, line 158:
```suggestion
                return reverse(f'{prefix}{model_opts.app_label}:{model_opts.model_name}_list')
```

jeremystretch (changes requested):
Can we simplify the logic here to avoid duplicating the view name?

renatoalmeidaoliveira:
@jeremystretch since `get_viewname` already implement that logic, I think that way is even cleaner and avoids repetitions.

jeremystretch (approved):

Issue #18916, opened 2025-03-16
DynamicModelChoiceField doesn't render error message on submit


chii0815:
### Deployment Type

Self-hosted

### NetBox Version

v.4.2.5

### Python Version

3.11

### Steps to Reproduce

While developing a plugin I realized that the DynamicModelChoiceField doesn't display an error if `required=True` and the nothing is selected. The form just doesn't submit and the view doesn't give a hint why.

1. create a form with two choice fields (one django.forms.ModelChoiceField and one DynamicModelChoiceField)
2. set both to be required

### Expected Behavior

When both are empty the form should show an error message on submit

### Observed Behavior

only the forms.ModelChoiceField shows an error message

![Image](https://github.com/user-attachments/assets/459d1426-4054-40b8-8482-71aba22e22e0)

chii0815:
this is my py file for this form


```
from django import forms
from circuits.models import Provider
from tenancy.models import Tenant
from netbox.forms import NetBoxModelForm
from django.utils.translation import gettext_lazy as _
from utilities.forms.fields import DynamicModelChoiceField
from ..models import CVS

class CVSForm(NetBoxModelForm):

    provider = forms.ModelChoiceField(
        queryset = Provider.objects.all(),
        required = True,
        label = _('provider')
    )

    tenant = DynamicModelChoiceField(
        queryset = Tenant.objects.all(),
        required = True,
        selector= True,
        label= _('tenant')
    )  

    class Meta:
        model = CVS
        fields = ('name', 'provider', 'tenant', 'tags')

    def clean(self):
        super().clean()
```

chii0815:
this also happens on core models like device-types:

![Image](https://github.com/user-attachments/assets/4c48cbcb-3d36-4363-bedd-6f036d26c565)

jeremystretch:
This can be reproduced natively by attempting to add a provider account without specifying a provider.

PR #19005, opened 2025-03-25
Closes #8423: Allow assigning Service to FHRP Group, in addition to Device and VirtualMachine

### LINKED ISSUE #8423: Assign service to FHRP groups
### NetBox version

v3.1.6

### Feature type

New functionality

### Proposed functionality

Be able to assign services to a FHRP group.

This can be done by either:
- assign service to FHRP group (not supported right now)
- assign service to one of the member hosts, selecting IP address of FHRP group (not possible, as these IP addresses are not shown on member hosts)

### Use case

You have a service (e.g. HTTP) that is redundantly provided by 2 servers (e.g. web1, web2). For this, you define a FHRP group (e.g. "1"), add an IP address to it (say 192.168.100.100) and assign this group to both servers. Servers may be devices or virtual machines.

Now, you want to define the service somehow. Assigning it directly to a server is impossible, as the shared IP address is not shown (this is another topic). Assigning a service to the FHRP group is impossible too. 

In my case, my monitoring system reads service definitions from NetBox, which is sadly impossible.

Workaround is to not use FHRP groups in NetBox but assign CARP IP addresses directly to server interfaces, but that's a step backwards.

### Database changes

_No response_

### External dependencies

_No response_

jnovinger:
### Closes: #8423

- Allow assigning Service to FHRP Group, in addition to Device and VirtualMachine
- Changes relation type for all three to a single GenericForeignKey that is required
  - Has the same semantics as before, that were enforced in `Service.clean()`, now enforced by the GFK and in model forms
- Patches up REST API and GraphQL representations
- Fixes up `ServiceImportForm` to work with the new `Service.parent` GFK


arthanson (changes requested):
Also - not sure if we want to change from tabbed interface for selection of assignment to dropdown - see discussion in maintainers channel chat.

netbox/ipam/filtersets.py
```diff
@@ -1137,25 +1137,35 @@ def search(self, queryset, name, value):
...
+        method='filter_device',
+        field_name='pk',
         label=_('Device (ID)'),
     )
-    device = django_filters.ModelMultipleChoiceFilter(
-        field_name='device__name',
-        queryset=Device.objects.all(),
-        to_field_name='name',
-        label=_('Device (name)'),
+    virtual_machine = MultiValueCharFilter(
```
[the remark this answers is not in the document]
jnovinger: 🤦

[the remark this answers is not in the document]
jnovinger: > I think? we can get rid of the MultiValueCharFitler fields for these with just a single assignment_type ContentTypeFilter - see VLANGroupFilterSet scope_type.

Actually, I would argue that the scope filters on VLAN Groups are buggy. For instance, using the demo data import, I can select more than one site to filter on--say 'JBB Branch 104' and 'JBB Branch 109'. However, it's only actually ever using the second one of those (guessing the last one to appear in the query params) for any filtering (see screenshot).

image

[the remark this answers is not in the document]
jnovinger: > Need to add fhrpgroup to the filter form

Fixed

jnovinger:
> FHRP Group details view should show the related services, either as related objects or a sub-table.

Fixed

arthanson:
@jnovinger test failing, not sure why: netbox/ipam/views.py:1435:9: SyntaxError: Cannot use `match` statement on Python 3.9 (syntax was added in Python 3.10)

jeremystretch:
@arthanson see #19112

jnovinger:
> @jnovinger test failing, not sure why: netbox/ipam/views.py:1435:9: SyntaxError: Cannot use `match` statement on Python 3.9 (syntax was added in Python 3.10)

Ah, this is actually a linting thing. I guess this is what the PR from Jeremy re: ruff and 3.10 python target is all about.

jnovinger:
Just realized I also missed updating the CSV import form for Services. Working on that now.

netbox/ipam/forms/bulk_import.py
```diff
@@ -572,15 +570,43 @@ class ServiceImportForm(NetBoxModelImportForm):
...
+            self.cleaned_data['parent'] = parent
+
+        for ip_address in self.cleaned_data.get('ipaddresses', []):
             if not ip_address.assigned_object or getattr(ip_address.assigned_object, 'parent_object') != parent:
                 raise forms.ValidationError(
                     _("{ip} is not assigned to this device/VM.").format(ip=ip_address)
                 )
 
-        return self.cleaned_data['ipaddresses']
+        return self.cleaned_data
```
jnovinger, line 612: All of this logic got moved up to `clean()` from `clean_ipaddresses()` because it needs access to a resolved value for the `parent` field. I tried doing that within the existing `clean_ipaddresses()`, but it was a mess and didn't feel reliable.

netbox/ipam/forms/bulk_import.py
```diff
@@ -572,15 +570,43 @@ class ServiceImportForm(NetBoxModelImportForm):
...
+
+    def clean(self):
+        super().clean()
+
+        if (parent := self.cleaned_data.get('parent')):
+            self.cleaned_data['parent_object_id'] = parent.pk
+        elif not parent and (parent_id := self.cleaned_data.get('parent_object_id')):
+            ct = self.cleaned_data.get('parent_object_type')
+            parent = ct.model_class().objects.filter(id=parent_id).first()
+            self.cleaned_data['parent'] = parent
```
jnovinger, line 604: `ServiceImportForm` supports either using the (`parent_object_type`, `parent`) field pair or the (`parent_object_type`, `parent_object_id`) field pair. This `if` and `elif` handle making sure they are both resolved when the other is used.

I don't feel fantastic about this approach. If there's pushback I'm happy to switch it to only support the (`parent_object_type`, `parent`) field pair.

netbox/ipam/forms/model_forms.py
```diff
@@ -21,7 +21,7 @@
 from utilities.forms.utils import get_field_value
```
arthanson, line 21: If you set parent type without the associated device, VM, FHRP Group and try to save there is no error given, should be a "Cannot set parent type without an associated object" message

arthanson (changes requested):
Just a couple minor issues - also needs docs in docs/models/ipam/service.md add in the parent field, can look at scope on vlan group for template.

netbox/ipam/forms/model_forms.py
jnovinger: Phew, this only took forever to figure out. After a fair amount of debugging compiled .ts, reading bootstrap docs, and finally understanding how those in-page "This field is required." messages get displayed, I discovered this is because of #18916.

The **tl;dr** is that the `APISelector`'s template changes the DOM tree structure just enough to break the way Bootstrap display's invalid feedback for fields. See [my comment on the issue](https://github.com/netbox-community/netbox/issues/18916#issuecomment-2791189429).

jnovinger: I think the work-around for this one is to just use Django's built in `Select` widget, since with the `HTMXSelect` widget on the `parent_object_type` form field, the form is being rendered and a queryset attached to the `parent` form field select, we don't necessarily need the `APISelect` widget functionality.

The downside is that means no object selector modal for the `parent` form field for the time being.

Alternatively, we solve #18916 real quick.

jnovinger: After discussing this in the maintainers call, we decided to move forward with this work as is re: the lack of validation messaging working in the `DynamicModelChoiceField`. #18916 will be a follow up in a future patch release.

jnovinger:
@arthanson , please note that the migrations have changed, for a couple of reasons:
1. there was a conflict with a migration from another branch that got merged ahead of this one
2. I had forgotten to enforce that `Service.parent` could not be null on the model itself.

You'll need at least wind back the previous IPAM `0078`, `0079`, and `0080` migrations before trying to run these.

arthanson (approved):

PR #19014, opened 2025-03-26
Closes #19002: Module type profiles

### LINKED ISSUE #19002: Extend module types to support modeling FRU components
### NetBox version

v4.2.6

### Feature type

Data model extension

### ModuleTypeProfile

Introduce a new ModuleTypeProfile model to represent a meta class of module types (e.g. power supply, NIC, or GPU). Each instance of this model will define a user-configurable schema of attributes to be assigned on module types having this profile.

```python
class ModuleTypeProfile(NetBoxModel):
    name = models.CharField()
    schema = models.JSONField()
```

The `schema` field will hold a user-defined [JSON schema](https://json-schema.org/) which defines the scalar attributes that can be defined on module types with this profile. For example, a CPU profile might have a profile specify the following:

```json
{
    "properties": {
        "architecture": {
            "type": "string"
        },
        "speed": {
            "type": "number",
            "minimum": "0.01",
            "multipleOf": "0.01"
        },
        "cores": {
            "type": "number",
            "minimum": 1,
            "multipleOf": 1
        }
    }
}
```

Note that these attributes are intentionally limited to storing the [basic types](https://cswr.github.io/JsonSchema/spec/basic_types/) supported by JSON schema (strings, numbers, booleans, and null). Any more advanced needs will require the use of a custom field on the ModelType model.

### ModuleType

Two new fields will be added to the ModuleType model:

* `profile`: An optional ForeignKey to ModuleTypeProfile
* `attributes`: A JSONField which stores the assigned key-value pairs for a module type as defined by the assigned profile

When editing a module type, the UI form will be extended to render all profile attributes as discrete form fields, with simple validation applied automatically.

### Use case

The introduction of module type profiles with configurable attributes will improve the ability to accurately model field-replaceable units (FRU) as modules in NetBox. Today, this is typically done using inventory items, however that model is considerably limited in several ways:

* Inventory items have no ability to track unoccupied bays/sockets
* Inventory items (like other device components) must be named
* Inventory items cannot contain components (interfaces, power ports, etc.)
* There is no mechanism available for storing profile-specific attributes on inventory items

Modules are better suited for this task, as only the last constraint listed above affects modules as well, and is overcome by the changes proposed in this FR. (Ultimately, we'll likely deprecate inventory items in favor of this and related improvements to modules.)

### Database changes

- Introduce the ModuleTypeProfile model
- Add a `profile` ForeignKey on ModuleType
- Add an `attributes` JSONField on ModuleType

### External dependencies

N/A

jeremystretch:
### Closes: #19002

- Move Module & ModuleType models to `modules.py`
- Introduce the new ModuleTypeProfile model
- Add `profile` ForeignKey and `attribute_data` JSONField on ModuleType
- Implement JSONSchemaProperty for rendering form fields from JSON schema properties
- Introduce MultipleOfValidator
- Extend the model form for ModuleType to render attribute fields depending on the selected profile
- Enable filtering of module types by attribute values
- Include some initial profiles to be installed during migration
- Documentation & tests

### DIFF
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -14,7 +14,9 @@ from netbox.forms import NetBoxModelBulkEditForm
 from tenancy.models import Tenant
 from users.models import User
 from utilities.forms import BulkEditForm, add_blank_choice, form_from_model
-from utilities.forms.fields import ColorField, CommentField, DynamicModelChoiceField, DynamicModelMultipleChoiceField
+from utilities.forms.fields import (
+    ColorField, CommentField, DynamicModelChoiceField, DynamicModelMultipleChoiceField, JSONField,
+)
 from utilities.forms.rendering import FieldSet, InlineFields, TabbedGroups
 from utilities.forms.widgets import BulkEditNullBooleanSelect, NumberWithOptions
 from virtualization.models import Cluster
@@ -576,7 +578,7 @@ class DeviceTypeBulkEditForm(NetBoxModelBulkEditForm):
 
 
 class ModuleTypeProfileBulkEditForm(NetBoxModelBulkEditForm):
-    schema = forms.JSONField(
+    schema = JSONField(
         label=_('Schema'),
         required=False
     )
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -641,7 +641,7 @@ class ModuleTypeBulkEditForm(NetBoxModelBulkEditForm):
             name=_('Chassis')
         ),
     )
-    nullable_fields = ('part_number', 'weight', 'weight_unit', 'description', 'comments')
+    nullable_fields = ('part_number', 'weight', 'weight_unit', 'profile', 'description', 'comments')
 
 
 class DeviceRoleBulkEditForm(NetBoxModelBulkEditForm):
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -46,6 +46,7 @@ __all__ = (
     'ModuleBayBulkEditForm',
     'ModuleBayTemplateBulkEditForm',
     'ModuleTypeBulkEditForm',
+    'ModuleTypeProfileBulkEditForm',
     'PlatformBulkEditForm',
     'PowerFeedBulkEditForm',
     'PowerOutletBulkEditForm',
@@ -574,7 +575,31 @@ class DeviceTypeBulkEditForm(NetBoxModelBulkEditForm):
     nullable_fields = ('part_number', 'airflow', 'weight', 'weight_unit', 'description', 'comments')
 
 
+class ModuleTypeProfileBulkEditForm(NetBoxModelBulkEditForm):
+    schema = forms.JSONField(
+        label=_('Schema'),
+        required=False
+    )
+    description = forms.CharField(
+        label=_('Description'),
+        max_length=200,
+        required=False
+    )
+    comments = CommentField()
+
+    model = ModuleTypeProfile
+    fieldsets = (
+        FieldSet('name', 'description', 'schema', name=_('Profile')),
+    )
+    nullable_fields = ('description', 'comments')
+
+
 class ModuleTypeBulkEditForm(NetBoxModelBulkEditForm):
+    profile = DynamicModelChoiceField(
+        label=_('Profile'),
+        queryset=ModuleTypeProfile.objects.all(),
+        required=False
+    )
     manufacturer = DynamicModelChoiceField(
         label=_('Manufacturer'),
         queryset=Manufacturer.objects.all(),
@@ -609,7 +634,7 @@ class ModuleTypeBulkEditForm(NetBoxModelBulkEditForm):
 
     model = ModuleType
     fieldsets = (
-        FieldSet('manufacturer', 'part_number', 'description', name=_('Module Type')),
+        FieldSet('profile', 'manufacturer', 'part_number', 'description', name=_('Module Type')),
         FieldSet(
             'airflow',
             InlineFields('weight', 'max_weight', 'weight_unit', label=_('Weight')),
--- a/netbox/netbox/filtersets.py
+++ b/netbox/netbox/filtersets.py
@@ -360,16 +360,20 @@ class AttributeFiltersMixin:
         # Extract JSONField-based filters from the incoming data
         if data is not None:
             for key, value in data.items():
-                if key.startswith(self.attribute_filter_prefix):
+                if field := self._get_field_lookup(key):
                     # Attempt to cast the value to a native JSON type
                     try:
-                        value = json.loads(value)
+                        self.attr_filters[field] = json.loads(value)
                     except (ValueError, json.JSONDecodeError):
                         pass
-                    field = f'{self.attributes_field_name}__{key.split(self.attribute_filter_prefix, 1)[1]}'
-                    self.attr_filters[field] = value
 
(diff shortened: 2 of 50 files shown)


jeremystretch:
@alehaa I considered that as part of the planning for this FR, but wasn't happy with the idea of having to define the constraint for _every_ module bay, and defining it on the template feels a bit hacky IMO. At any rate, I opted to punt on it for now as there's already a lot going on in this FR.

You're certainly welcome to open a new FR for it if you'd like. I almost did but wasn't sure if we'd be able to get to it in time for v4.3.

jnovinger:
Wanted to quickly note one bug I found while clicking around. When you create an module type where an attribute is an integer, then change the value definition in the associated module type profile to `"number"` and try to edit the module type, you receive `KeyError: 'attr_<field>'`.

STR:
1. Create a new hard drive module type from the prepopulated "Hard disk" module type profile (`/dcim/module-type-profiles/4/`), making sure to enter an integer for the `size` attribute--I used 524 (actual value is 524.3).
2. Modify the module type profile's `schema` to loosen the `size.type` specification from `integer` to `number`.
3. Now go back and edit the hard drive you created in step 1 to change the size from 524 to 524.3 and save.
4. Edit the module type profile's `schema` to tighten the `size.type` specification back up from `number` to `integer`.
5. Open the edit form for the hard drive created in step 1 (and edited in step 3) again. Click "Save" without changing anything.

Expected:
- Application would change the hard drive module type's `size` attribute from 524.3 to 524.

Actual:
- Application raises `KeyError: 'attr_size'`

jnovinger:
One more attribute validation bug.

STR:
1. Create a GPU module type using the GPU module type profile (`dcim/module-type-profiles/3/`)
2. Select any of the enumerated Interface values and save
3. Edit the GPU module type you just created
4. Clear the selection from the Interface field and try to save

Expected:

- Form validates changed data and saves the existing module type

Actual:

- Form return validation error indicating lack of selection is invalid, even though the field is not marked required: `Invalid schema: '' is not one of ['PCIe 4.0', 'PCIe 4.0 x8', 'PCIe 4.0 x16', 'PCIe 5.0 x16', 'SoC'] Failed validating 'enum' in schema['properties']['interface']: {'enum': ['PCIe 4.0', 'PCIe 4.0 x8', 'PCIe 4.0 x16', 'PCIe 5.0 x16', 'SoC'], 'type': 'string'} On instance['interface']: '' `

jnovinger:
Grr, sorry, one more.

STR:
1. Create module type with the GPU module type profile (`dcim/module-type-profiles/3/`)
2. Edit the newly created module type and clear out the Profile field so that no profile is selected
3. Click Save

Expected:

- Form validates successfully and saves the changes in data

Actual:

- Form performs validation as if the previously attached module type profile is still attached--it returns a form error for the Memory field (which was correctly hidden when the Profile field was cleared).

netbox/utilities/jsonschema.py
```diff
@@ -0,0 +1,133 @@
...
+                field_kwargs['validators'] = [
+                    RegexValidator(regex=self.pattern)
+                ]
+
+        # Integer/number validation
+        elif self.type in (PropertyTypeEnum.INTEGER, PropertyTypeEnum.NUMBER):
+            if self.minimum:
+                field_kwargs['min_value'] = self.minimum
+            if self.maximum:
+                field_kwargs['min_value'] = self.maximum
```
jnovinger, line 108: I feel like this should be assigning to `field_kwargs['min_value']` but somehow it still seems to work correctly, which leads me to believe that this validation is not actually being applied correctly.

jeremystretch:
> Form performs validation as if the previously attached module type profile is still attached--it returns a form error for the Memory field (which was correctly hidden when the Profile field was cleared).

This is due to bug #19023 (for which a PR is currently open). Once the fix for that is available, I'll merge it here and confirm.

jnovinger (changes requested):
Noted some bugs in regular PR comments. This review also has a couple of questions.

jnovinger:
One more weird bug around attributes and schemas. If you specify an array-type field in a module type profile, but try to use it when creating a module type, a `TypeError` is raised.

1. Add a new attribute to any schema, with `"type": "array"` and no other arguments. Save the profile.
2. Try to create a new module type with that profile.

Expected:

- Form is rendered

Actual:

- `TypeError: SimpleArrayField.__init__() missing 1 required positional argument: 'base_field'` is raised

This might be as simple as validating array-type attributes, but at the moment, you are allowed to save a schema like the following:

```json
{
    "properties": {
        "aliases": {
            "type": "array"
        },
        "architecture": {
            "type": "string"
        },
        "cores": {
            "minimum": 1,
            "multipleOf": 2,
            "type": "number"
        },
        "speed": {
            "maximum": 1,
            "minimum": 0.5,
            "multipleOf": 0.01,
            "type": "number"
        }
    }
}
```

netbox/dcim/forms/bulk_edit.py
```diff
@@ -574,7 +575,31 @@ class DeviceTypeBulkEditForm(NetBoxModelBulkEditForm):
     nullable_fields = ('part_number', 'airflow', 'weight', 'weight_unit', 'description', 'comments')
 
 
+class ModuleTypeProfileBulkEditForm(NetBoxModelBulkEditForm):
+    schema = forms.JSONField(
+        label=_('Schema'),
+        required=False
+    )
```
jnovinger, line 582: I noticed that `schema` values of `null` and `""` don't affect any change on the items selected. However, I noticed that `{}` does essentially clear the schemas of the selected items.

This might be the intended behavior, but was not super intuitive.

jnovinger: Relatedly, `true` and `false` literals were accepted as valid input for this field, even though they don't necessarily make sense.

Perhaps values should be locked down to some representation of no schema (`null` or `{}` perhaps) and some schema that is at least of type object. Not sure, though, just thinking out loud.

netbox/netbox/filtersets.py
```diff
@@ -345,3 +348,28 @@ def search(self, queryset, name, value):
...
+                if key.startswith(self.attribute_filter_prefix):
+                    # Attempt to case the value to a native JSON type
+                    try:
+                        value = json.loads(value)
+                    except (ValueError, json.JSONDecodeError):
+                        pass
+                    field = f'{self.attributes_field_name}__{key.split(self.attribute_filter_prefix, 1)[1]}'
+                    self.attr_filters[field] = value
+
+        super().__init__(data=data, queryset=queryset, request=request, prefix=prefix)
```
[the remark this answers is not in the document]
jeremystretch: I looked into this a bit but I don't think there's much we can do here realistically. The filterset can't know ahead of time what type of queries to enforce, and module types assigned different profiles may have overlapping fields. For example, one profile might declare an integer named `foo` and another profile a string named `foo`. There's no way to reconcile these when filtering for `?foo=`, other than perhaps making some assumptions based on the value passed.

The only way I can think of to lock this down reliably would be to exclude anything other than exact lookups on attributes, but that would remove a substantial amount of functionality from the feature.

jnovinger:
> This is due to bug #19023 (for which a PR is currently open). Once the fix for that is available, I'll merge it here and confirm.

I went ahead and snagged the review for that one. It's been merged.

netbox/netbox/filtersets.py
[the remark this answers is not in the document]
jnovinger: That's an incredibly fair point.

Any thoughts on the `__isnull` lookup issue? I don't think it would be as big a deal if it just silently ignored it, but raising a `ValueError` and causing a 500 seem pretty not ideal.

jnovinger (approved):
💥

PR #19028, opened 2025-03-27
Update choices.py - add MoCA (Coaxial)

### LINKED ISSUE #19018: Add MoCA (coaxial) Interface Type
### NetBox version

v4.2.6

### Feature type

Data model extension

### Proposed functionality

I am raising this FR to ask for a new coaxial "interface type" for [MoCA](https://en.wikipedia.org/wiki/Multimedia_over_Coax_Alliance) (Wikipedia link).

I am unsure as to whether the project would prefer to break out the specific MoCA "version" as multiple types? Different versions of MoCA _do_ have different rated interface throughput as well as limitations around simultaneous clients, so that would be an argument for multiple type options. **However**, the existing "DOCSIS" interface type currently doesn't get split out by version of DOCSIS, even though there are a bunch of those. So, following existing style / KISS would imply a single "MoCA" type.

I'd be willing to make an attempt at the PR to add MoCA (it would be my first PR to NetBox) if I can get clarification on whether each different MoCA revision (1.0, 1.1, 2.0, 2.5, 3.0) belongs in the list separately, or not.  AIUI, the only changes would be a few lines added in `netbox/dcim/choices.py`?

(I would think that _if_ the different MoCA revisions are broken out as different types, then DOCSIS should _also_ have similar types added for consistency, but it's unclear to me what should happen with the old "vanilla" DOCSIS type at that point...)

### Use case

Real world example of ISP/telco-type hardware that uses this interface: [InCoax in:xtnd Control C25x series](https://www.incoax.com/products/), this is one example from one vendor, but I know there are others. These systems exist as a way to re-use existing coax/CATV plant to distribute higher-speed service, typically in an apartment/MDU-type scenario.

### Database changes

None required to my knowledge

### External dependencies

n/a

ZPrimed:
### Fixes: #19018

Added MoCA (Coaxial) Interface Type

### DIFF
--- a/netbox/dcim/choices.py
+++ b/netbox/dcim/choices.py
@@ -986,6 +986,7 @@ class InterfaceTypeChoices(ChoiceSet):
 
     # Coaxial
     TYPE_DOCSIS = 'docsis'
+    TYPE_MOCA = 'moca'
 
     # PON
     TYPE_BPON = 'bpon'
@@ -1182,6 +1183,7 @@ class InterfaceTypeChoices(ChoiceSet):
             _('Coaxial'),
             (
                 (TYPE_DOCSIS, 'DOCSIS'),
+                (TYPE_MOCA, 'MoCA'),
             )
         ),
         (


jeremystretch (approved):
Thanks @ZPrimed!

Issue #18305, opened 2025-01-05
Make ContactsMixin available for plugins 


alehaa:
### NetBox version

v4.1.10

### Feature type

Other

### Triage priority

I volunteer to perform this work (if approved)

### Proposed functionality

1. Add `ContactsMixin` and related mixin classes (column, filterset, …) to the plugin documentation.
2. Add `ObjectContactsView` to [feature-view auto-registration](https://github.com/netbox-community/netbox/blob/develop/netbox/netbox/models/features.py#L639-L655)

### Use case

Plugins may want to use contacts for custom models. Making the mixin class public for use in the API will prevent plugins from reinventing the wheel and better integrate with NetBox, as they can implement a consistent UI with the NetBox core.

### Database changes

None.

### External dependencies

None.

PR #19017, opened 2025-03-27
Fixes: #18305 make contacts mixin available for plugins

renatoalmeidaoliveira:
### Fixes: #18305 make contacts mixin available for plugins

- Create `ObjectContactsView` feature view
- Deleted `tenancy.views.ObjectContactsView`
- Updated `object_contacts.html` removing `object_children` template
- Added  `netbox.views.generic.ObjectContactsView` use case to registry
- Removed `SiteContactsView`
- Removed `ProviderContactsView`
- Removed `ProviderAccountContactsView`
- Removed `CircuitContactsView`
- Removed `RegionContactsView`
- Removed `SiteGroupContactsView`
- Removed `LocationContactsView`
- Removed `ManufacturerContactsView`
- Removed `DeviceContactsView`
- Removed `PowerPanelContactsView`
- Removed `AggregateContactsView`
- Removed `PrefixContactsView`
- Removed `IPRangeContactsView`
- Removed `IPAddressContactsView`
- Removed `ServiceContactsView`
- Removed `TenantContactsView`
- Removed `ClusterGroupContactsView`
- Removed `ClusterContactsView`
- Removed `VirtualMachineContactsView`
- Removed `TunnelGroupContactsView`
- Removed `TunnelContactsView`
- Removed `L2VPNContactsView`
- Updated `docs/plugins/development/models.md` adding `ContactsMixin` as available feature

I also tested with some plugins and its working when `ContactsMixin` is added to `Model` definition class

PR #19029, opened 2025-03-27
Fixes: #18305 make contacts mixin available for plugins

renatoalmeidaoliveira:
### Fixes: #18305 make contacts mixin available for plugins

- Create `ObjectContactsView` feature view
- Deleted `tenancy.views.ObjectContactsView`
- Updated `object_contacts.html` removing `object_children` template
- Added  `netbox.views.generic.ObjectContactsView` use case to registry
- Removed `SiteContactsView`
- Removed `ProviderContactsView`
- Removed `ProviderAccountContactsView`
- Removed `CircuitContactsView`
- Removed `RegionContactsView`
- Removed `SiteGroupContactsView`
- Removed `LocationContactsView`
- Removed `ManufacturerContactsView`
- Removed `DeviceContactsView`
- Removed `PowerPanelContactsView`
- Removed `AggregateContactsView`
- Removed `PrefixContactsView`
- Removed `IPRangeContactsView`
- Removed `IPAddressContactsView`
- Removed `ServiceContactsView`
- Removed `TenantContactsView`
- Removed `ClusterGroupContactsView`
- Removed `ClusterContactsView`
- Removed `VirtualMachineContactsView`
- Removed `TunnelGroupContactsView`
- Removed `TunnelContactsView`
- Removed `L2VPNContactsView`
- Updated `docs/plugins/development/models.md` adding `ContactsMixin` as available feature

I also tested with some plugins and its working when `ContactsMixin` is added to `Model` definition class


renatoalmeidaoliveira:
@jeremystretch rebasing the previous PR #19017  genereted a lot o merge conflicts, so I decided to create a new one.

jeremystretch (commented):
I'm unclear on why you chose to rewrite ObjectContactsView. Was this necessary?

renatoalmeidaoliveira:
@jeremystretch I rewrite `ObjectContactsView` to follow `ObjectChangeLogView` and `ObjectJournalView` pattern, but it would problably be way simpler to keep it as a childview, If I keep `ObjectContactsView` as a child view I can also keep the template as is too.

jeremystretch:
Let's keep it as-is. It's better to leverage the built-in functions of ObjectChildrenView to avoid duplication of code.

netbox/netbox/views/generic/feature_views.py
```diff
@@ -12,19 +12,95 @@
...
+
+        # Gather all Contact Assignments for this object
+        content_type = ContentType.objects.get_for_model(model)
+        contactassignments = ContactAssignment.objects.restrict(request.user, 'view').prefetch_related(
+            'contact',
+        ).filter(
+            object_type=content_type,
+            object_id=obj.pk
+        ).order_by('priority', 'contact', 'role')
+        contactassignments = self.filterset(request.GET, contactassignments, request=request).qs
```
alehaa, line 69: I believe this get's direct contact assignments only. You can use `obj.get_contacts()` to get contacts for parent objects as well, like the original `ObjectContactsView` did.

netbox/netbox/views/generic/feature_views.py
```diff
@@ -12,19 +12,95 @@
...
+            obj = get_object_or_404(model, **kwargs)
+
+        # Gather all Contact Assignments for this object
+        content_type = ContentType.objects.get_for_model(model)
+        contactassignments = ContactAssignment.objects.restrict(request.user, 'view').prefetch_related(
+            'contact',
+        ).filter(
+            object_type=content_type,
+            object_id=obj.pk
+        ).order_by('priority', 'contact', 'role')
```
alehaa, line 68: Reviewing this code, I wonder if we could improve performance minimally by prefetching the related object assigned to the contact using `prefetch_related('object')`. This could be used either here or in the original view class.

renatoalmeidaoliveira: Thanks for this review, but I believe that it isn't necessary because [BaseTable](https://github.com/netbox-community/netbox/blob/12231ad71a36f7cd4a727806284ef1228510ea8a/netbox/netbox/tables/tables.py#L95) already prefetches RelatedFields, ManyToOneRel, and GenericForeignKey relation fields.
Additionally, after checking Django's debug SQL queries, I couldn't find any N+1 or related problems.

renatoalmeidaoliveira:
@jeremystretch I needed to mutate `kwargs` in `ObjectContactsView`'s `dispatch` because it's used later in `get` to retrieve the parent instance using `instance = self.get_object(**kwargs)`. Since `model` isn't a field of any model, it was raising an exception.

jeremystretch (approved):

netbox/netbox/views/generic/feature_views.py
jeremystretch: We ended up not changing any of the logic within ObjectContactsView.

PR #19061, opened 2025-04-02
Fixes #17413: Duplicate platform names allowed for different manufacturers

### LINKED ISSUE #17413: Allow Platforms with Same Name for Different Manufacturers
### NetBox version

v4.0.7

### Feature type

### Proposed functionality

In the Platform model, the system raises a validation error when trying to create a Platform with the same name even if the associated Manufacturer is different. This is problematic because two manufacturers can produce platforms with the same name. The system should allow Platform records with the same name if they are associated with different Manufacturers.

### Use case

The Platform name should only be enforced as unique within a Manufacturer. This allows two platforms with the same name but belonging to different manufacturers to be created.

**Steps to Reproduce**:

Create a Manufacturer (e.g., Manufacturer A).
Create a Platform named "Test Platform" under Manufacturer A.
Create another Manufacturer (e.g., Manufacturer B).
Attempt to create another Platform named "Test Platform" under Manufacturer B.

### Database changes

_No response_

### External dependencies

_No response_

abhi1693:


jeremystretch:
Hi @abhi1693, could you fix the linter errors please, so we can open this for review?

arthanson:
@abhi1693 unfortunately this will not make the cutoff for 4.3 since it still has some issues and hasn't been able to be reviewed yet.  We will have to push it to v4.4.

PR #19064, opened 2025-04-02
Closed #9763: Treat IP ranges as fully populated

### LINKED ISSUE #9763: Enable treating IP ranges as fully populated
### NetBox version

v3.2.6

### Feature type

New functionality

### Proposed functionality

Improve IP ranges functionality by implementing these features:

- An option to display particular IP ranges in IP address lists.
- An option to mark an IP range as restricted/reserved/fully utilized. IP addresses within such ranges will not show up as available in any IP lists.

### Use case

This would be extremely useful for DHCP ranges for instance. Right now nothing stops netbox users from using "free" IP addresses reserved for particular IP ranges. 

Example: I have an IP range defined which describes DHCP range in one of my subnets, but when browsing IP address lists nothing indicates that these addresses are reserved, thus they can be assigned to any device or VM leading to an IP conflict.

There is a workaround: bulk create IP addresses for the whole range and set a corresponding role for them all. But in this scenario users have to manually match ranges and IP addresses in case of range changes and accidents may happen due to human factor. Also a lot of unnecessary information is stored in the database and displayed in IP lists especially for large IPv6 prefixes where there can be thousands of reserved addresses in one prefix.

With my proposal implemented one could mark an IP range as reserved and users would not be allowed to allocate addresses from reserved ranges anymore or atleast will be notified that these addresses are reserved. It would also decrease amount of excessive information in IP lists, dramatically in some cases.

These features are optional and will not break existing databases.

### Database changes

Some new boolean fields for IP range description will be needed to implement this.

### External dependencies

_No response_

jeremystretch:
### Fixes: #9763

- Add a `mark_populated` boolean field to the IPRange model
- Replace the `add_available_ipaddresses()` utility function with `annotate_ip_space()`
- Introduce the AvailableIPSpace data class to represent unallocated space between two real IP objects
- Extend the `get_child_ranges()` method the Prefix model to allow passing filters

Note: Calling `get_available_ips()` on a prefix with `mark_utilized=True` no longer automatically returns an empty IPSet. This is a deviation from the current behavior that, while not explicitly addressed in the FR, is necessary to ensure the consistent treatment of both prefixes and IP ranges.

### DIFF
--- a/docs/models/ipam/iprange.md
+++ b/docs/models/ipam/iprange.md
@@ -29,9 +29,11 @@ The IP range's operational status. Note that the status of a range does _not_ ha
 !!! tip
     Additional statuses may be defined by setting `IPRange.status` under the [`FIELD_CHOICES`](../../configuration/data-validation.md#field_choices) configuration parameter.
 
-### Mark Reserved
+### Mark Populated
 
-If enabled, NetBox will prevent the creation of IP addresses which fall within the declared range (and assigned VRF, if any).
+!!! note "This field was added in NetBox v4.3."
+
+If enabled, NetBox will treat this IP range as being fully populated when calculating available IP space. It will also prevent the creation of IP addresses which fall within the declared range (and assigned VRF, if any).
 
 ### Mark Utilized
 
--- a/docs/models/ipam/iprange.md
+++ b/docs/models/ipam/iprange.md
@@ -2,11 +2,11 @@
 
 This model represents an arbitrary range of individual IPv4 or IPv6 addresses, inclusive of its starting and ending addresses. For instance, the range 192.0.2.10 to 192.0.2.20 has eleven members. (The total member count is available as the `size` property on an IPRange instance.) Like [prefixes](./prefix.md) and [IP addresses](./ipaddress.md), each IP range may optionally be assigned to a [VRF](./vrf.md).
 
-Each IP range can be marked as populated, which instructs NetBox to treat the range as though every IP address within it has been created (even though these individual IP addresses have not been created). This can be helpful in scenarios where the management of a subset of IP addresses has been deferred to an external system of record, such as a DHCP server. NetBox will prohibit the creation of individual IP addresses within a range that has been marked as populated.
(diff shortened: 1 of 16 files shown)


docs/models/ipam/iprange.md
```diff
@@ -2,6 +2,12 @@
...
 
+Each IP range can be marked as populated, which instructs NetBox to treat the range as though every IP address within it has been created (even though these individual IP addresses have not been created). This can be helpful in scenarios where the management of a subset of IP addresses has been deferred to an external system of record, such as a DHCP server. NetBox will prohibit the creation of individual IP addresses within a range that has been marked as populated.
+
+An IP range can also be marked as fully utilized. This will cause its utilization to always be reported as 100% when viewing the range or when calculating the utilization of a parent prefix. (If not enabled, a range's utilization is calculated based on the number of IP addresses which have been created within it.)
```
arthanson, line 7: Minor point: Wondering if we should either mention this as mark_utilitized or else change the form field display name to 'Mark fully utilized' as we don't ever use 'fully utilized' currently just utilized.

arthanson (commented):
I think there should be a test_filteresets.IPRangeTestCase testing for mark_populated, there should be one for mark_utilized as well, not sure why the test verifier isn't picking up those filter test missing?

There is also the case where you can initially have mark_populated off, then add an IP Address to it, then edit it and make mark_populated True.  In this case you can delete the IP Address but you can no longer edit it. Not sure if we want to dis-allow setting the flag if you already have IP addresses created under the range?

jeremystretch:
> There is also the case where you can initially have mark_populated off, then add an IP Address to it, then edit it and make mark_populated True. In this case you can delete the IP Address but you can no longer edit it. Not sure if we want to dis-allow setting the flag if you already have IP addresses created under the range?

IMO it makes sense to disallow editing the IP address, but let's see what feedback we get during the beta.

docs/models/ipam/iprange.md
jeremystretch: Dropped "fully" from the docs.

arthanson (approved):

PR #19074, opened 2025-04-03
Closes #18191: Remove duplicate SQL indexes

### LINKED ISSUE #18191: Duplicate database indexes
### Proposed Changes

As requested by @jeremystretch in discussion #18188, I am submitting this housekeeping issue to remove duplicate database indexes.

I started using the feature 'Index Tuning' of the Azure PostgreSQL flexible server for the Netbox database. Netbox version is 3.7.1 and PostgreSQL version is 14.13. After 24 hours, it found 6 rcommendations, all of them to drop duplicate indexes:

- Duplicate of "vpn_tunneltermination_termination". The equivalent index "vpn_tunneltermination_termination" is a unique index, while "vpn_tunnelt_termina_c1f04b_idx" is not.
- Duplicate of "core_managedfile_unique_root_path". The equivalent index "core_managedfile_unique_root_path" is a unique index, while "core_managedfile_root_path" is not.
- Duplicate of "core_autosyncrecord_object". The equivalent index "core_autosyncrecord_object" is a unique index, while "core_autosy_object__c17bac_idx" is not.
- Duplicate of "core_datafile_unique_source_path". The equivalent index "core_datafile_unique_source_path" is a unique index, while "core_datafile_source_path" is not.
- Duplicate of "vpn_l2vpntermination_assigned_object". The equivalent index "vpn_l2vpntermination_assigned_object" is a unique index, while "vpn_l2vpnte_assigne_9c55f8_idx" is not.
- Duplicate of "dcim_cabletermination_unique_termination". The equivalent index "dcim_cabletermination_unique_termination" is a unique index, while "dcim_cablet_termina_884752_idx" is not.

### Justification

As explained, this is likely because Netbox Team explicitly defines an index for the generic foreign key (GFK) relation termination, which Django does not do automatically. Separately, Netbox Team also declares a uniqueness constraint for this pair of fields, to ensure an object cannot have more than one termination. Hence the duplicate indexes. It should be safe to remove the former index.

jeremystretch:
### Fixes: #18191

- Remove the redundant index declarations from each model
- Generate migrations
- Add a note in the developer documentation
- Introduce a Django system check for duplicate indexes

Note: I did also perform an audit for any additional redundant indexes, but did not uncover any that weren't already listed in #18191.


jeremystretch:
I forgot; we wanted to look into adding a check for duplicate indexes as well. Let me dig into that.

jnovinger (approved):

PR #19078, opened 2025-04-04
Closes #17653 Add function to trim whitespaces in export templates via jinja environment settings

### LINKED ISSUE #17653: Add function to trim whitespaces in export templates via jinja environment settings
### NetBox version

3.7.5

### Feature type

### Proposed functionality

The goal is to achieve feature parity between the [Provisioning --> Config Templates edit] and [Customization --> Export Templates dialog]. In config templates it is possible to use jinja environment parameters, in export templates it is not.
It's also the same behavior in Netbox v4.1.2 (checked via demo application).

### Use case

I'm currently implementing a custom export template, which exports prefixes, their assigned VLANs and some custom fields. I'm aware that part of this, is also achievable via the csv export function, but I want to add some semantic sugar via the power of Jinja, like adding first host address, last host address, ...
The messy white space handling with jinja, unfortunately destroys the formatting of the csv.

### Database changes

None required?

### External dependencies

None?

renatoalmeidaoliveira:
### Fixes: #17653 Add function to trim whitespaces in export templates via jinja environment settings

- Creates `RenderMixin` model with fields: `template_code`, `environment_params`, `mime_type`, `file_name`, `file_extension`, `as_attachment`
- Add `RenderMixin` to `ExportTemplate` and `ConfigTemplate`, removing the repeated fields
- Unify the export logic inside `RenderMixin` using `ExportTemplate` as reference for exports, i.e. always using `template_code` instead of using `data_file` when exists, because `template_code` should be equals to `data_file` after `sync` and it gives the end user more control of what template is in use.

Extra Comments:

- Allowing the user to setup `file_name` and `file_extension` to `ConfigTemplate` may give the user more flexibility in ZTP use cases using NetBox as source of python ZTP configuration files,


netbox/netbox/models/features.py
```diff
@@ -337,6 +340,87 @@ class Meta:
...
+        help_text=_(
+            'Any <a href="https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.Environment">additional parameters</a>'
+            ' to pass when constructing the Jinja2 environment.'
+        )
+    )
+    mime_type = models.CharField(
+        max_length=50,
+        blank=True,
+        verbose_name=_('MIME type'),
+        help_text=_('Defaults to <code>text/plain; charset=utf-8</code>')
```
jeremystretch, line 365:
```suggestion
        help_text=_('Defaults to <code>{default}</code>').format(default=DEFAULT_MIME_TYPE)
```

netbox/extras/models/configs.py
```diff
@@ -4,16 +4,13 @@
...
 from django.utils.translation import gettext_lazy as _
-from jinja2.loaders import BaseLoader
-from jinja2.sandbox import SandboxedEnvironment
 
 from extras.querysets import ConfigContextQuerySet
-from netbox.config import get_config
 from netbox.models import ChangeLoggedModel
-from netbox.models.features import CloningMixin, CustomLinksMixin, ExportTemplatesMixin, SyncedDataMixin, TagsMixin
+from netbox.models.features import (
+    CloningMixin, CustomLinksMixin, ExportTemplatesMixin, SyncedDataMixin, TagsMixin, RenderMixin)
```
jeremystretch, line 11:
```suggestion
from netbox.models.features import (
    CloningMixin, CustomLinksMixin, ExportTemplatesMixin, SyncedDataMixin, TagsMixin, RenderMixin
)
```

netbox/extras/models/configs.py
```diff
@@ -210,7 +207,8 @@ def clean(self):
 # Config templates
 #
 
-class ConfigTemplate(SyncedDataMixin, CustomLinksMixin, ExportTemplatesMixin, TagsMixin, ChangeLoggedModel):
+class ConfigTemplate(
+    SyncedDataMixin, CustomLinksMixin, ExportTemplatesMixin, TagsMixin, ChangeLoggedModel, RenderMixin):
```
jeremystretch, line 211:
```suggestion
class ConfigTemplate(
    SyncedDataMixin, CustomLinksMixin, ExportTemplatesMixin, TagsMixin, ChangeLoggedModel, RenderMixin
):
```

netbox/extras/models/models.py
```diff
@@ -460,37 +431,11 @@ def sync_data(self):
         self.template_code = self.data_file.data_as_string
     sync_data.alters_data = True
 
-    def render(self, queryset):
-        """
-        Render the contents of the template.
-        """
-        context = {
+    def get_context(self, context=None, queryset=None):
+        export_context = {
```
jeremystretch, line 435: This can be condensed to just return the dictionary directly (`return {...}`).

netbox/extras/tables/tables.py
```diff
@@ -550,12 +550,17 @@ class ConfigTemplateTable(NetBoxTable):
...
+        false_mark=None
+    )
 
     class Meta(NetBoxTable.Meta):
         model = ConfigTemplate
         fields = (
             'pk', 'id', 'name', 'description', 'data_source', 'data_file', 'data_synced', 'role_count',
-            'platform_count', 'device_count', 'vm_count', 'created', 'last_updated', 'tags',
+            'platform_count', 'device_count', 'vm_count', 'created', 'last_updated', 'tags', 'as_attachment',
+            'mime_type', 'file_name', 'file_extension',
```
jeremystretch, line 563: Try to order primary model fields ahead of annotations & generic fields for readability.
```suggestion
            'as_attachment', 'mime_type', 'file_name', 'file_extension','platform_count', 'device_count', 'vm_count',
            'created', 'last_updated', 'tags', 
```

netbox/extras/tables/tables.py
```diff
@@ -550,12 +550,17 @@ class ConfigTemplateTable(NetBoxTable):
         url_params={'config_template_id': 'pk'},
         verbose_name=_('Virtual Machines')
     )
+    as_attachment = columns.BooleanColumn(
```
jeremystretch, line 553: We need to add columns for `file_name`, `file_extension`, and `mime_type` as well, with labels to support translation.

netbox/netbox/models/features.py
```diff
@@ -337,6 +340,87 @@ class Meta:
...
+        output = self.render(context=context, queryset=queryset)
+        mime_type = self.mime_type or 'text/plain; charset=utf-8'
+
+        # Build the response
+        response = HttpResponse(output, content_type=mime_type)
+
+        if self.as_attachment:
+            extension = f'.{self.file_extension}' if self.file_extension else ''
+            if queryset:
+                filename = self.file_name or filename_from_model(queryset.model)
```
jeremystretch, line 415: Won't `queryset.model` always be this class? We should be able to remove `queryset` as a kwarg for this method and instead reference `self._meta.model`.

netbox/netbox/models/features.py
```diff
@@ -337,6 +340,87 @@ class Meta:
...
+            elif context:
+                filename = self.file_name or filename_from_object(context)
+            full_filename = f'{filename}{extension}'
```
jeremystretch, line 418: Need to ensure `filename` is always defined.

netbox/utilities/jinja2.py
```diff
@@ -48,10 +48,10 @@ def cache_templates(self, templates):
 # Utility functions
 #
 
-def render_jinja2(template_code, context):
+def render_jinja2(template_code, context, environment_params={}):
```
jeremystretch, line 51: `render_jinja2()` needs to be added to `__all__`. Not sure why it's not already there,

jeremystretch (changes requested):
Make sure you're completing the [checklist for extending models](https://netboxlabs.com/docs/netbox/en/stable/development/extending-models/) each time you add a new field. The REST API serializers, GraphQL types, and several forms also need to be updated to reflect the new fields.

netbox/netbox/models/features.py
renatoalmeidaoliveira: At this point, `self._meta.model` refers to either `ExportTemplate` or `ConfigTemplate`, not the actual exported/rendered config `Model`. So, to use the device name or device type as the `filename`, we need to rely on the `queryset`.

I also use `queryset` at the beginning of the function to call `self.render`, which internally calls `self.get_context`. Because of this, passing `queryset` through `kwargs` is currently necessary.

While it might be possible to pass both the `context` dict and `queryset` together in a single variable for `render` and `get_context`, doing so could make the function's purpose harder to understand.

docs/models/extras/configtemplate.md
```diff
@@ -12,10 +12,6 @@ See the [configuration rendering documentation](../../features/configuration-ren
 
 A unique human-friendly name.
 
-### Weight
```
renatoalmeidaoliveira, line 15: `ConfigTemplate` doesn't have a `weight` field.

jeremystretch (approved):

Issue #19073, opened 2025-04-03
Allow Plugins to be hidden or disabled


arthanson:
### NetBox version

v4.2.6

### Feature type

Change to existing functionality

### Proposed functionality

Allow plugins to be hidden or disabled in the list of installed plugins (Admin -> plugins).

* Hidden would mean the plugin is not displayed in the list even though it is installed.
* Disabled would mean it is shown, but the plugin is not clickable in the list (i.e. not detail view or installation settings are shown).

This could be handled by a new config parameter PLUGIN_DISPLAY:

```
PLUGIN_DISPLAY = {
  "netbox-auth-extension": "hidden", 
  "netbox-other-plugin": "disabled",
}
```
This could also potentially be done in the individual plugin in the PluginConfig.

### Use case

There are some plugin cases where the plugin is just a support plugin (generally for other plugins) or for support functionality in the cloud or enterprise case where it is either confusing to show it in the plugin list, or there is no reason to show the installation information for it.

### Database changes

N/A

### External dependencies

N/A - new config param only.

PR #19087, opened 2025-04-04
19073 allow plugins to be marked as hidden or disabled in plugins table

arthanson:
### Fixes: #19073 

Defined a new config parameter PLUGINS_CATALOG_CONFIG, I kept it separate from PLUGINS_CONFIG as those settings are directly controlled by the plugin itself and I didn't want to potentially break it if there were overlap In the names or the plugin wasn't expecting additional params for some reason (probably very unlikely, but...)

```
PLUGINS_CATALOG_CONFIG = {
    'hidden': ['netbox_acls',],
    'unlinked': ['netbox_bgp',],
}
```

With this config can see netbox_acls was removed from the list and netbox_bgp is not a clickable link anymore:

Monosnap Plugins | NetBox 2025-04-04 09-09-59

In netbox/core/views.py could unfold the `if q:` lines to make one loop through the list, but thought this was clearer and list comprehension is very fast (and a short list).  I also wanted to keep the list with the hidden flags instead of removing them when it initially generates the list as we may potentially want to use the complete list for other things in the future (including hidden items).

```
        plugins = self.get_cached_plugins(request).values()
        if q:
            plugins = [obj for obj in plugins if q.casefold() in obj.title_short.casefold()]

        plugins = [plugin for plugin in plugins if not plugin.hidden]
```


docs/configuration/plugins.md
```diff
@@ -33,3 +33,21 @@ Note that a plugin must be listed in `PLUGINS` for its configuration to take eff
 
 ---
 
+## PLUGINS_CONFIG
```
jeremystretch, line 36:
```suggestion
## PLUGINS_TABLE_CONFIG
```

But also, the use of "table" here feels a little out-of-place IMO. Would `PLUGINS_CATALOG_CONFIG` make more sense?

docs/configuration/plugins.md
```diff
@@ -33,3 +33,21 @@ Note that a plugin must be listed in `PLUGINS` for its configuration to take eff
...
+
+```python
+PLUGINS_TABLE_CONFIG = {
+    'plugin1': {
+        'hidden': True,
+    },
+    'plugin2': {
+        'disabled': True,
+    },
+}
```
jeremystretch, line 50: What do you think about flipping the ordering here?

```python
PLUGINS_TABLE_CONFIG = {
    'hidden': [
        'plugin1',
    ],
    'disabled': [
        'plugin2',
    ],
}
```

I ask because I can see us extending this configuration parameter in the future for other uses (e.g. disabling the catalog entirely or specifying an alternate URL), and preserving the root-level namespace would keep our options open.

docs/configuration/plugins.md
```diff
@@ -33,3 +33,21 @@ Note that a plugin must be listed in `PLUGINS` for its configuration to take eff
 
 ---
 
+## PLUGINS_CONFIG
+
+Default: Empty
+
+This parameter holds configuration settings for how the individual plugins are displayed in the plugins table under Admin->System->Plugins.  Declaring `hidden` as True will make it so the plugin is not shown in the table.  Declaring `disabled` will make it show the plugin in the table, but it won't be linked to the details and installation page for the plugin. An example configuration is shown below:
```
jeremystretch, line 40: I worry that using the term "disabled" here invites confusion with _actually_ disabling a plugin, rather than merely removing its link in the catalog. Trying to think of a better term.

netbox/core/tables/plugins.py
```diff
@@ -86,3 +87,10 @@ class Meta(BaseTable.Meta):
         # List installed plugins first, then certified plugins, then
         # everything else (with each tranche ordered alphabetically)
         order_by = ('-is_installed', '-is_certified', 'name')
+
+    def render_title_long(self, value, record):
+        if record.disabled:
+            return f"{value}"
+
+        url = reverse('core:plugin', args=[record.config_name])
+        return mark_safe(f"<a href='{url}'>{value}</a>")
```
jeremystretch, line 96: I think we can obviate the need for a custom method by passing a lambda to `linkify` on the column:

```python
linkify=lambda record: not record.disabled
```

jeremystretch (changes requested):

docs/configuration/plugins.md
jeremystretch: Maybe we could denote these plugins as "unavailable" (with the presumption they are not installed).

arthanson: changed it to unlinked

netbox/core/tables/plugins.py
arthanson: I'm not sure this will work, it can't use get_absolute_url as it's not a model so the url accessor has to be provided (as below) - I don't think it can be overloaded with this and the test for record.disabled?

```
    title_long = tables.Column(
        linkify=('core:plugin', [tables.A('config_name')]),
        verbose_name=_('Name')
    )
```

arthanson:
Changed from disabled to unlinked as that seems more descriptive. I think unavailable would also be confusing as it is available / installed it just doesn't have installation instructions or details.  Not 100% happy with unlinked

netbox/core/tables/plugins.py
jeremystretch: Ah right, thought we had a URL method on the plugin record.

jeremystretch (approved):
One note: The detail view for a plugin is still visible even if a plugin has been declared hidden or static. It's unclear to me whether this is intended, so I'm approving this as-is.

PR #19093, opened 2025-04-06
Fixes #15310 - Ability to choose VLAN groups in VMs span over multiple sites

### LINKED ISSUE #15310: Assignment of VLANs out of VLAN Groups to VMs
### Deployment Type

Self-hosted

### NetBox Version

v3.7.3

### Python Version

3.9

### Steps to Reproduce

Create two different sites and a virtualization cluster with devices (nodes) not assigned to any site.
Create a "VLAN Group" with the scope "site group". (VLANs in global scope are not forced to be unique)
Create a virtual machine in this cluster and assign a VLAN of the VLAN Group to an interface.
This worked until a version < 3.7.? and is now not possible anymore.

This example is for a virtualization platform with vSAN and vMotion over more than one data center.

### Expected Behavior

Create two different sites in a site group and a virtualization cluster, defined with devices (nodes) in two different sites.
Assign the cluster to this site group and virtual machine to this cluster group.
Then the ability to use VLANs of the VLAN Group (scope site group) in virtual machines, that are assigned to a cluster group.

### Observed Behavior

1st: can't assign a cluster  to a site group, only to a site.
     virtual machines can't be assigned to a site group, only to a site.
2nd: because of the 1st topic we hadn't assigned the cluster to any site.
     new behavior in newer version observed: the VLAN assignment is not possible anymore but old assignments are still visible on the "802.1Q tagged" interface.
 
Due to this, not to be captious enough with the assignment of a cluster to a site and the misleading understanding of a site, we wonder how the future development of the Netbox will handle this issue.
 
In our new understanding of the Netbox we have to shift away from the idea that a [Site](https://docs.netbox.dev/en/stable/models/dcim/site/) is equal a data center. If we use a site as a logical unit wich includes different data centers as location everything works fine.
The possibility to only define a VLAN Group in the scope of a site group then looks meaningless to us.
 
We would like to know the direction of how multi-homed virtualization is intend to be handled in netbox before starting a complete restructure.
Or maybe this is only a bug, that will be fixed in upcoming releases?

kiwichrism:
Hi,

first of all I would like to apologize how I opened, handle and behaved in the original issue I opened.\
Luckily with some help I was able to adjust the codebase to resolve the problem we saw and make it possible to select VLAN Groups as described for our scenario.

### Fixes: #1234 - What I like to propose here would fix the issue mentioned, I would appreciate this pull request to be accepted and hope it will help other community members as well.

If there is anything else I can do to help or if there are question, please don't hesitate to get in touch with me.

Cheers.

### DIFF
--- a/netbox/ipam/querysets.py
+++ b/netbox/ipam/querysets.py
@@ -160,6 +160,11 @@ class VLANQuerySet(RestrictedQuerySet):
                     scope_type=ContentType.objects.get_by_natural_key('virtualization', 'clustergroup'),
                     scope_id=vm.cluster.group_id
                 )
+            if vm.cluster._site_group:
+                q |= Q(
+                    scope_type=ContentType.objects.get_by_natural_key('dcim', 'sitegroup'),
+                    scope_id__in=vm.cluster._site_group.get_ancestors(include_self=True)
+                )+            if vm.cluster._site_group:
         if site:
             # Add VLANGroups scoped to the assigned site (or its group or region)
             q |= Q(


jnovinger:

PR #19101, opened 2025-04-07
Closes #14591: Saved table configurations

### LINKED ISSUE #14591: Multiple configure table views
### NetBox version

v3.6.7

### Feature type

New functionality

### Proposed functionality

I propose a functionality that will allow saving different column layout selections to the dropdown menu of the Configure Table pop-up window. This would be useful in cases where there are multiple scenarios, and some columns are not needed to be visible for certain scenarios. For example, in the first scenario, to display a specific type of device, a larger number of columns is required, while for the second scenario, to display a different type of device, a significantly smaller number of columns is needed.

Example: In the first scenario, we list all routers, and the necessary columns are - Name, Status, Role, Type, Serial number, Asset tag, Site, Rack, Tags, Comments; in the second scenario, we list all IP phones, and only the columns - Serial number, Contacts, and the custom field "Price" are necessary.

In the current version, achieving this would require changing the column layout in the Configure Table pop-up window each time, and this would need to be done for each scenario.

The **creation** of new column layouts and their **selection** would be performed in the Configure Table pop-up window. **Editing** would be done in another location, which I will describe below.
I suggest that in the Configure Table pop-up window, desired column layouts for different scenarios be saved as dropdown menu options. This way, when we select the columns needed for display in one of the scenarios, to **save** the selection we click on a button called, for example, "Save Table View". We would do the same for each subsequent scenario. Regarding the **selection** of these saved column layouts, it would be done from that dropdown menu within the Configure Table pop-up window. **Editing** parameters for saved scenarios (found in the dropdown menu) could be similar to Saved Filters (Customization --> Saved Filters), where saved filters can be edited. Something similar could possibly be done for editing "Saved Table Views".

### Use case

Implementing this would, for example, streamline and expedite the process of displaying necessary details for different types of devices by enabling or disabling various columns.
Users would be able to swiftly choose between different levels of required details from the dropdown menu.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #14591

- Introduce the TableConfig model (similar to SavedFilter)
- Introduce SharedObjectViewMixin for queryset resolution
- Overhaul the application of table config & ordering logic
- Add the `configuration` & `config_params` properties on BaseTable
- Extend the HTMX template for tables to update the "save" link
- Update the frontend JS to provide a `select-all` CSS class


jnovinger (changes requested):
Found a couple of bugs clicking around:

- Messing up the "Ordering" field (e.g. including one field, a comma, and then nothing more) results (properly probably) in a validation error. However, it also causes the list of selected columns to be cleared.

image

- ~~It's unclear how to correctly specify a field in "Ordering" that has multiple words (e.g. "Asset tag"). I tried plain, with single and double quotes, and with a trailing comma.~~ This only happened because of the previous bug and "Asset column" was not in the list of selected columns. Oops.

- "Create and Add Another" breaks with `DoesNotExist: ObjectType matching query does not exist.` Edited to add: the original table configuration is saved. It appears the exception happens on the subsequent render of the page for the next config.

image

jeremystretch:
> "Create and Add Another" breaks with DoesNotExist: ObjectType matching query does not exist.

I've addressed this by enabling cloning for the TableConfig model.

jnovinger:
> > Messing up the "Ordering" field (e.g. including one field, a comma, and then nothing more) results (properly probably) in a validation error. However, it also causes the list of selected columns to be cleared.
> 
> I've fixed the issue with the selected columns list getting cleared out, however the validation error on the ordering field I think is technically correct. It's meant to convey a list of columns (even though we currently support only a single ordering). The validation is a natural behavior of the field, so if we want to change it that would be best handled under a separate issue as there are other places we use SimpleArrayField.

Cool, that definitely works for me. It was a passing thought as I described the actual bug.

jnovinger (changes requested):
Found one more. It looks like the first column in the "Ordering" column is not being picked up/used. You can see it in the screenshots below:

- in the first, you see the detail view of the saved config in question, with the first column in "Ordering" showing the sort direction, but not the name of the column. In the second, you see the edit view for the same saved config, showing the ordering selections.

image
image

- In the final screenshot, you can see the same saved config applied. You can see that first column in the ordering is not being sorted on, while the second and third are being sorted on correctly.

image

Edit to add one more screenshot, showing the ordering applied to the query--which I think is incorrect. Not only is it not ordering on 'virtual chassis' as configured, it's also including an ordering on 'site name' which is not selected in the saved configuration.

image

netbox/extras/api/serializers_/tableconfigs.py
```diff
@@ -0,0 +1,22 @@
...
+        queryset=ObjectType.objects.all()
+    )
+
+    class Meta:
+        model = TableConfig
+        fields = [
+            'id', 'url', 'display_url', 'display', 'object_type', 'table', 'name', 'slug', 'description', 'user',
+            'weight', 'enabled', 'shared', 'columns', 'ordering', 'created', 'last_updated',
+        ]
+        brief_fields = ('id', 'url', 'display', 'name', 'slug', 'description')
```
jnovinger, line 22: Should there be some sort of indication of what model/table this applies to in `brief_fields`?

netbox/extras/filtersets.py
```diff
@@ -326,6 +327,59 @@ def _usable(self, queryset, name, value):
...
+        method='search',
+        label=_('Search'),
+    )
+    object_type_id = django_filters.ModelMultipleChoiceFilter(
+        queryset=ObjectType.objects.all(),
+        field_name='object_type'
+    )
+    object_type = ContentTypeFilter(
+        field_name='object_type'
+    )
```
jnovinger, line 341: This one doesn't seem to be working. At this point, all of my existing saved configs are for object type dcim.Device. But, as you can see, none are returned when filtering for that object type.

image

jnovinger: Derp, this belongs on the filter form below not here, my bad.

netbox/extras/forms/model_forms.py
```diff
@@ -301,6 +304,66 @@ def __init__(self, *args, initial=None, **kwargs):
...
+    object_type = ContentTypeChoiceField(
+        label=_('Object type'),
+        queryset=ObjectType.objects.all()
+    )
+    ordering = SimpleArrayField(
+        base_field=forms.CharField(),
+        required=False,
+        label=_('Ordering'),
+        help_text=_(
+            "Enter a comma-separated list of column names. Prepend a name with a hyphen to reverse the order."
```
jnovinger, line 318: My observation here is that the name of the column in case-insensitive. Is this worth mentioning here or no?

netbox/extras/graphql/types.py
```diff
@@ -186,6 +187,15 @@ class SubscriptionType(ObjectType):
     user: Annotated["UserType", strawberry.lazy('users.graphql.types')] | None
 
 
+@strawberry_django.type(
+    models.TableConfig,
+    filters=TableConfigFilter,
+    pagination=True
+)
+class TableConfigType(ObjectType):
+    user: Annotated["UserType", strawberry.lazy('users.graphql.types')] | None
```
jnovinger, line 196: I'm only seeing `changelog`, `display`, `class_type`, and `user` as available output types for this type. Is that intentional?

netbox/extras/models/models.py
```diff
@@ -524,6 +526,98 @@ def url_params(self):
...
+
+    def __str__(self):
+        return self.name
+
+    def get_absolute_url(self):
+        return reverse('extras:tableconfig', args=[self.pk])
+
+    @property
+    def docs_url(self):
+        return f'{settings.STATIC_URL}docs/models/extras/tableconfig/'
```
jnovinger, line 598: I don't these docs actually got added.

netbox/extras/tables/tables.py
```diff
@@ -281,6 +282,36 @@ class Meta(NetBoxTable.Meta):
         )
 
 
+class TableConfigTable(NetBoxTable):
```
jnovinger, line 285: I was waiting for this one. :P

jnovinger (commented):
Had some other questions about code and one weird behavior.

netbox/extras/api/serializers_/tableconfigs.py
jeremystretch: `brief_fields` defines the set of fields returned by the serializer when `?brief=True` is passed with an API request. (It's a subset of the serializer's fields.)

netbox/extras/forms/model_forms.py
jeremystretch: The list should be case-sensitive. Each item needs to be an exact match to a column name (the name of the column itself, not its rendered verbose name/title).

netbox/extras/graphql/types.py
jeremystretch: Good catch. The type' decorator is missing `fields='__all__'`.

netbox/extras/models/models.py
jeremystretch: Added!

netbox/extras/forms/model_forms.py
jnovinger: Ah, I bet that's what was causing my issue with the first item not being picked up/used.

jnovinger:
> > Found one more. It looks like the first column in the "Ordering" column is not being picked up/used.
> 
> I'm not able to reproduce this; manually specifying multiple ordering columns in the TableConfig appears to work as expected. Did you maybe specify an invalid column name?

Yep, I think that's probably what happened. I didn't realize it needed to be the exact name of the field--so, I was using 'virtual chassis' instead of 'virtual_chassis'. Will verify that's the issue.

jnovinger (changes requested):
Doh, sorry. Found 1 more thing and narrowed the down the issue with the create/edit form accepting invalid column names for the 'ordering' column.

New issue, steps to reproduce:
1. Navigate to Table Config list view
2. Click Add button

Expected:
- Create form ready to be filled out

Actual:
- `DoesNotExist: ObjectType matching query does not exist.` exception

Edited to add: I noticed there wasn't even a `+` button in the side nav for Table Configs. Perhaps there doesn't need to be an Add button on the Table Config list view either.

jnovinger (changes requested):
Everything is looking solid and the new validation rules are working great, except that the ability to specify a reverse ordering with a `-<column_name>` is now broken.

jnovinger (approved):
💥

PR #19161, opened 2025-04-13
Fixes #18879 - Add prefix filtering by assigned VLAN group

### LINKED ISSUE #18879: Add prefix filtering by assigned VLAN group
### NetBox version

v4.2.4

### Feature type

New functionality

### Proposed functionality

Please make it possible to use both `Quick search` and `Filter` to be able to properly filter out prefixes by specific `VLAN Groups` they are assigned to.

Currently it's possible to only sort the table (asc/desc) by that property, but filtering isn't avaible.

### Use case

Having many networks, prefixes and vlans across multiple sites, it would be very helpful to be able to use a proper `filter` and `quick search` to be able to  filter for prefixes assigned to specific `vlan groups`.

Right now the only way to do so is to have the prefixes page show 1000 entries and use the browser's search tool to highlight specific vlan groups.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
#### Overview

This PR enhances NetBox by enabling filtering of prefixes by the assigned VLAN Group. Users can now filter using both quick search (by VLAN Group slug) and FilterForm (by ID).

#### Changes Introduced

- **IPAM FilterSet:** Added VLAN Group filters to support filtering by both ID and slug.
- **IPAM FilterForm:** Introduced a `vlan_group_id` filter for direct filtering by VLAN Group.
- **Testing:** Added tests to validate VLAN Group filtering.

Thank you for reviewing this PR!

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -351,6 +351,18 @@ class PrefixFilterSet(NetBoxModelFilterSet, ScopedFilterSet, TenancyFilterSet, C
         to_field_name='rd',
         label=_('VRF (RD)'),
     )
+    vlan_group_id = django_filters.ModelMultipleChoiceFilter(
+        field_name='vlan__group',
+        queryset=VLANGroup.objects.all(),
+        to_field_name="id",
+        label=_('VLAN Group (ID)'),
+    )
+    vlan_group = django_filters.ModelMultipleChoiceFilter(
+        field_name='vlan__group__slug',
+        queryset=VLANGroup.objects.all(),
+        to_field_name="slug",
+        label=_('VLAN Group (slug)'),
+    )
     vlan_id = django_filters.ModelMultipleChoiceFilter(
         queryset=VLAN.objects.all(),
         label=_('VLAN (ID)'),
--- a/netbox/ipam/tests/test_filtersets.py
+++ b/netbox/ipam/tests/test_filtersets.py
@@ -645,9 +645,16 @@ class PrefixTestCase(TestCase, ChangeLoggedFilterSetTests):
         vrfs[1].export_targets.add(route_targets[1])
         vrfs[2].export_targets.add(route_targets[2])
 
+        vlan_groups = (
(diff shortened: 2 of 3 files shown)


netbox/ipam/filtersets.py
```diff
@@ -389,6 +401,7 @@ def search(self, queryset, name, value):
             qs_filter |= Q(prefix__contains=value.strip())
         except (AddrFormatError, ValueError):
             pass
+        qs_filter |= Q(vlan__group__slug__contains=value.strip())
```
jeremystretch, line 404: `search()` methods must be limited to local fields only. This change needs to be reverted.

jeremystretch (changes requested):

netbox/ipam/filtersets.py
pheus: Thanks! I've just removed the last commit and force-pushed it.

netbox/ipam/tests/test_filtersets.py
```diff
@@ -850,6 +857,15 @@ def test_site(self):
         params = {'site': [sites[0].slug, sites[1].slug]}
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 4)
 
+    def test_vlan_group(self):
+        vlan_groups = VLANGroup.objects.all()[:2]
+        params = {'vlan_group_id': [vlan_groups[0].pk, vlan_groups[1].pk]}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 4)
+        params = {'vlan_group': [vlan_groups[0].slug]}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
```
jeremystretch, line 865: These two lines can be removed; the following two lines provide sufficient test coverage.

```suggestion
        params = {'vlan_group': [vlan_groups[0].slug]}
        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
```

jeremystretch (changes requested):

PR #19180, opened 2025-04-14
Fixes #19056 – Add Device filtering by Location slug

### LINKED ISSUE #19056: Add location by slug to device filterset
### NetBox version

v4.2.6

### Feature type

Other

### Proposed functionality

Similar to site, platform, and other fields the location field should be added to lookup by slug and not just by location ID.

### Use case

Mainly usable for API users it allows the use of unique slugs for filtering without the need to lookup the object's ID

### Database changes

_No response_

### External dependencies

_No response_

pheus:
#### Overview

This PR enhances the Device API by allowing devices to be filtered by the slug of their assigned location, in addition to the existing ID-based filtering.

#### Changes

- **DCIM FilterSet:** Added the `location` filter to support lookup by slug.
- **Tests:** Added unit tests to verify slug-based filtering functionality for locations.

This improves API usability for clients relying on human-readable slugs instead of numeric IDs.

Thanks for your time!

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1057,6 +1057,13 @@ class DeviceFilterSet(
         lookup_expr='in',
         label=_('Location (ID)'),
     )
+    location = TreeNodeMultipleChoiceFilter(
+        queryset=Location.objects.all(),
+        field_name='location',
+        lookup_expr='in',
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @pheus!

PR #19188, opened 2025-04-15
Fixes: #17794 - Setup get_field_value to return multiple values when required

### LINKED ISSUE #17794: Utility function get_field_value returns last value when used on a multi-value field
### Deployment Type

Self-hosted

### Triage priority

### NetBox Version

v4.1.4

### Python Version

3.10

### Steps to Reproduce

1. Utilize get_field_value on a form field that may produce a single value or multiple values (Ex: `tagged_vlans`)

### Expected Behavior

get_field_value to return multiple values

### Observed Behavior

get_field_valueu returns a single value

DanSheps:
### Fixes: #17794 - Setup get_field_value to return multiple values in the event of multi-values passed in a modelmultiplechoice or multiplechoice field.

* Edit tests
* Edit get_field_value to return multiple values in the event of a Multi-Value field being passed

### DIFF
--- a/netbox/utilities/forms/utils.py
+++ b/netbox/utilities/forms/utils.py
@@ -1,7 +1,8 @@
 import re
 
 from django import forms
-from django.forms.models import fields_for_model
+from django.forms import MultipleChoiceField
+from django.forms.models import fields_for_model, ModelMultipleChoiceField
 from django.utils.translation import gettext as _
 
 from utilities.choices import unpack_grouped_choices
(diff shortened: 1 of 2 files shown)


DanSheps:
A note, I could rewrite this to use a new function, get_field_values() in the case of form fields with MultipleChoice fields.  This seemed like the path of least resistance.

jeremystretch (commented):
I wasn't able to reproduce the original issue. Could you please update #17794 with complete reproduction steps?

PR #19226, opened 2025-04-17
Fixes #19134: Interfaces TX power allows negative value

### LINKED ISSUE #19134: Interface TX Power should allow negative values
### NetBox version

v4.2.6

### Feature type

Data model extension

### Proposed functionality

[Update the validations](https://github.com/netbox-community/netbox/commit/a173083e5bc0997098d513811c9283988e82a210) for Interface TX Power to allow negative values

### Use case

We run 400G ZR optics in our devices and in some cases we need to lower the transmit power to a negative because it goes through an amplifier or the path is short enough

### Database changes

Possible database validation change

### External dependencies

_No response_

yzguy:
### Fixes: #19134 

Adjust the validations for interfaces TX power to allow negative values


jeremystretch:
Thank you for your interest in contributing, but please note that FR #19134 has not yet been opened for work and thus we cannot accept a PR for it at this time.

yzguy:
No worries, I noticed some labels were added on the issue so I misunderstood that for being accepted, but now I see there is a `status:accepted` tag.

PR is still attached to the issue so whenever it gets accepted it'll be easy to reference/re-open

PR #19234, opened 2025-04-17
DOCS : fix data-validation.md code example

playerla:
Syntax fix in pseudo code to avoid silent error : add coma to define a tuple


jeremystretch:
The syntax is correct. This does not need a PR.

playerla:
Hi,

docs/configuration/data-validation.md :

- Line 41 : there is a comma
- Line 52 : another example without details. I copy this one in my configuration file and it breaks. Comma is missing.

PR #19256, opened 2025-04-22
Fixes #18959: Preserve ordering of terminations in cable traces

### LINKED ISSUE #18959: Cable Trace visual issue
### Deployment Type

Self-hosted

### NetBox Version

v.4.2.5

### Python Version

3.11

### Steps to Reproduce

Connect with each other 2 Devices with rear ports then set cable trace.

Image

### Expected Behavior

99 should be on the left side and 100 on the right
Cables should not cross each other.

### Observed Behavior

First of all the 99 should be on the left side and 100 on the right

Image

Besides that between ports 3,4 and 100,99 there are 2 cross-connections.
Unfortunately, they overlap each other.
The inscription Cable # 628 is a visual bug you can't see what is underneath.

jeremystretch:
### Fixes: #18959

Remove the call to `sorted()` as terminations should already be ordered naturally.

### DIFF
--- a/netbox/dcim/svg/cables.py
+++ b/netbox/dcim/svg/cables.py
@@ -226,7 +226,7 @@ class CableTraceSVG:
         nodes_height = 0
         nodes = []
         # Sort them by name to make renders more readable
-        for i, term in enumerate(sorted(terminations, key=lambda x: str(x))):
+        for i, term in enumerate(terminations):
             node = Node(
                 position=(offset_x + i * width, self.cursor),
                 width=width,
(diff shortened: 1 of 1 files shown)


netbox/dcim/svg/cables.py
```diff
@@ -226,7 +226,7 @@ def draw_object_terminations(self, terminations, offset_x, width):
         nodes_height = 0
         nodes = []
         # Sort them by name to make renders more readable
```
bctiemann, line 228: We should remove the comment if it no longer applies.

bctiemann (changes requested):

bctiemann (approved):

PR #19258, opened 2025-04-22
Fixes #19255: Improved documentation consistency for configuration default values

### LINKED ISSUE #19255: Improve Documentation Consistency For Configuration Default Values
### Change Type

### Area

Configuration

### Proposed Changes

In the Netbox Configuration, the Remote Authentication Settings REMOTE_AUTH_AUTO_CREATE_GROUPS value has Default: 'False' set which will have the False have a grey background, but under ALLOW_TOKEN_RETRIEVAL, the Default: value is just True which lacks the highlighting. There are a bunch of several examples, but I am attaching examples.
Image

Image

mr1716:
### Fixes: #19255

The changes ensure that the default values are highlighted the same way in the documentation

### DIFF
--- a/docs/configuration/plugins.md
+++ b/docs/configuration/plugins.md
@@ -2,7 +2,7 @@
 
 ## PLUGINS
 
-Default: `Empty`
+Default: ``
 
 A list of installed [NetBox plugins](../plugins/index.md) to enable. Plugins will not take effect unless they are listed here.
 
@@ -13,7 +13,7 @@ A list of installed [NetBox plugins](../plugins/index.md) to enable. Plugins wil
 
 ## PLUGINS_CONFIG
 
-Default: `Empty`
+Default: ``
 
 This parameter holds configuration settings for individual NetBox plugins. It is defined as a dictionary, with each key using the name of an installed plugin. The specific parameters supported are unique to each plugin: Reference the plugin's documentation to determine the supported parameters. An example configuration is shown below:
 
--- a/docs/configuration/plugins.md
+++ b/docs/configuration/plugins.md
@@ -2,7 +2,7 @@
 
 ## PLUGINS
 
-Default: ``
+Default: `[]`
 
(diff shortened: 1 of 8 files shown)


docs/configuration/plugins.md
```diff
@@ -2,7 +2,7 @@
 
 ## PLUGINS
 
-Default: Empty
+Default: `Empty`
```
jeremystretch, line 5: "Empty" isn't a literal value. We should either leave this as is, or change it to "`[]` (empty)"

docs/configuration/plugins.md
```diff
@@ -13,7 +13,7 @@ A list of installed [NetBox plugins](../plugins/index.md) to enable. Plugins wil
 
 ## PLUGINS_CONFIG
 
-Default: Empty
+Default: `Empty`
```
jeremystretch, line 16: Similar to `PLUGINS` above, but this is an empty dictionary by default.

docs/configuration/system.md
```diff
@@ -206,14 +206,14 @@ If `STORAGE_BACKEND` is not defined, this setting will be ignored.
 
 ## TIME_ZONE
 
-Default: UTC
+Default: `UTC`
```
jeremystretch, line 209: It might help to represent this explicitly as a string to avoid confusion with a global constant.
```suggestion
Default: `"UTC"`
```

jeremystretch (changes requested):

docs/configuration/plugins.md
mr1716: @jeremystretch thanks. Will switch it to []

docs/configuration/system.md
mr1716: @jeremystretch thanks. Change made

jeremystretch (approved):
Thanks @mr1716!

PR #19276, opened 2025-04-22
Fixes: #19275 - Make type choice for interfaces non-required

### LINKED ISSUE #19275: Interface Bulk Edit requires interface "type"
### Deployment Type

Self-hosted

### NetBox Version

v4.2.8

### Python Version

3.11

### Steps to Reproduce

1. Go to DCIM > Interfaces
2. Select multiple interfaces
3. Edit the interfaces
4. Enter in something for "Label" but leave other fields empty

### Expected Behavior

Interface(s) are edited properly

### Observed Behavior

Interface Edit page "jumps" to the "label" field

DanSheps:
### Fixes: #19275 - Make type choice for interfaces non-required

* Alter `form_from_model` to also mark the widget itself as non-required

### DIFF
--- a/netbox/utilities/forms/utils.py
+++ b/netbox/utilities/forms/utils.py
@@ -200,6 +200,8 @@ def form_from_model(model, fields):
     form_fields = fields_for_model(model, fields=fields)
     for field in form_fields.values():
         field.required = False
+        if field.widget and field.widget.is_required:
+            field.widget.is_required = False
 
     return type('FormFromModel', (forms.Form,), form_fields)
 


jeremystretch (commented):
I coincidentally just addressed this in #19283. Is there any reason not to just set `field.widget.is_required = False`?

DanSheps:
> I coincidentally just addressed this in #19283. Is there any reason not to just set `field.widget.is_required = False`?

Nope, I was just being extra cautious to make sure that there wasn't any issue with perhaps a null widget or anything.

PR #19284, opened 2025-04-23
Fixes #19264: Support table configs on child object list views

### LINKED ISSUE #19264: Loading of Saved Table Config in ChildrenViews is not possible
### Deployment Type

Self-hosted

### NetBox Version

v4.3.0-beta1

### Python Version

3.10

### Steps to Reproduce

1. Create a Prefix `192.168.0.0/22`
2. Create an IP address `192.168.0.1/22`
3. Navigate to the Prefix `192.168.0.0/22` → *IP Addresses* tab
4. Open **Configure Table**, add the `Tags` column
5. Save the table configuration as `Prefix-Tags`
6. Clear the user's table configuration via **User Preferences**
7. Return to the Prefix `192.168.0.0/22` → *IP Addresses* tab
8. Attempt to load the saved table config via the dropdown (**Configure Table**)

### Expected Behavior

The saved table configuration (`Prefix-Tags`) should appear in the dropdown list and be loadable.

### Observed Behavior

No saved table configuration appears in the dropdown. The previously saved config cannot be accessed.

**Additional Notes:**
- This behavior seems limited to nested or child views (e.g., the *IP Addresses* tab within a Prefix).
- Saving and loading configs on top-level list views (like global IP Address list) does work as expected.

jeremystretch:
### Fixes: #19264

- Move the logic for applying a TableConfig (specified by the `tableconfig_id` URL parameter) to TableMixin
- Move the logic for querying available TableConfigs to the new `get_table_configs()` utility function
- Pass `table_configs` as template context under ObjectChildrenView


jnovinger (commented):
Seeing a weird issue where the 'Reset' button inside the 'Configure table' modal is not clearing the selected table config. Let me know if you need details.

jnovinger (approved):

PR #19290, opened 2025-04-23
16238 map custom-link button colors to tabler

### LINKED ISSUE #16238: Custom link button colour(s) affected by dark mode
### Deployment Type

Self-hosted

### NetBox Version

v4.0.2

### Python Version

3.11

### Steps to Reproduce

1. Create a custom link with Button class "White".
2. Enable dark mode.

### Expected Behavior

Button stays white.

### Observed Behavior

Button turns black.

arthanson:
### Fixes: #16238 

Adds CSS styling for missing btn-white, btn-black, btn-grey.  Note CSS extend doesn't catch all potential styling on elements depending on how the selectors are defined (see https://stackoverflow.com/questions/13335544/how-to-alias-class-name-in-css-or-sass) but it appears all the styling is good for these buttons.

Note: ghost-dark (used for LINK type) is a tabler button but in the version of tabler we use link text doesn't display in dark mode, should be fixed in latest version of tabler.


jeremystretch (commented):
Rather than re-mapping the color values, could we just add CSS aliases to provide backward compatibility? IMO this is a less confusing approach.

```
.btn-white {
  @extend .btn-light;
}

.btn-black {
  @extend .btn-dark;
}
```

jeremystretch (approved):

PR #19294, opened 2025-04-23
Closes #17151: Use ColoredLabelColumn for Type Column in Circuits Table

### LINKED ISSUE #17151: add color backgrounds for circuit type
### NetBox version

v4.0.8

### Feature type

### Proposed functionality

resubmitting this as a FR. (original: https://github.com/netbox-community/netbox/issues/17135)
See: https://github.com/netbox-community/netbox/issues/12216. When circuit type colors were first added the 3rd screenshot shows the circuit type column having the text highlighted when viewing circuits. This never seemed to make it into the original feature, or perhaps was a regression at some point. 

Example:
image

This FR would be to have circuit types be highlighted in the same way that rack roles are highlighted.
image

### Use case

Would make circuit types easier to distinguish visually when viewing circuits. 
Circuit types already have the ability to assign a color to them, this FR would be to adjust the circuits display to reference the assigned circuit type color for the 'type' column.

### Database changes

none.

### External dependencies

none.

pheus:
### Fixes: #17151 add color backgrounds for circuit type

This PR updates the `type` column in the Circuits table to use `ColoredLabelColumn`, which applies the configured color for each Circuit Type as a background. This improves visual clarity and consistency with how Rack Roles are displayed.

As part of this change, the now-redundant `linkify=True` attribute has been removed from the column definition, since `ColoredLabelColumn` already provides clickable labels by default.

### DIFF
--- a/netbox/circuits/tables/circuits.py
+++ b/netbox/circuits/tables/circuits.py
@@ -61,9 +61,8 @@ class CircuitTable(TenancyColumnsMixin, ContactsColumnMixin, NetBoxTable):
         linkify=True,
         verbose_name=_('Account')
     )
-    type = tables.Column(
+    type = columns.ColoredLabelColumn(
         verbose_name=_('Type'),
-        linkify=True
     )
     status = columns.ChoiceFieldColumn()
     termination_a = columns.TemplateColumn(


jeremystretch (approved):
Thanks @pheus!

PR #19304, opened 2025-04-24
Fixes #19309: N+1 problem on /interfaces, /ip-addresses and /prefixes requests

### LINKED ISSUE #19309: N+1 problem in dcim/ipam API calls
### Deployment Type

Self-hosted

### NetBox Version

v4.2.8

### Python Version

3.10

### Steps to Reproduce

1. Enable SQL logging
2. Do API requests:
* api/ipam/ip-addresses/
* api/ipam/prefixes
* api/dcim/interfaces

### Expected Behavior

All required data is loaded by batch SQL requests

### Observed Behavior

In the output you will see repeated SQL calls for each object in the response set:
* `circuits_virtualcircuittermination` for interfaces
* `dcim_interface` folowing by `dcim_device` for ip-address
* `dcim_site` for prefixes

Tishka17:
### Fixes: #19309
In case of requesting `/interfaces`, `/ip-addresses` and `/prefixes` some of the fields are selected separately for each model. We should prefetch them.

### DIFF
--- a/netbox/ipam/api/views.py
+++ b/netbox/ipam/api/views.py
@@ -81,9 +81,7 @@ class RoleViewSet(NetBoxModelViewSet):
 
 
 class PrefixViewSet(NetBoxModelViewSet):
-    queryset = Prefix.objects.prefetch_related(
-        "scope",
-    ).all()
+    queryset = Prefix.objects.prefetch_related("scope")
     serializer_class = serializers.PrefixSerializer
     filterset_class = filtersets.PrefixFilterSet
 
@@ -111,7 +109,7 @@ class IPAddressViewSet(NetBoxModelViewSet):
                 Interface.objects.select_related("device"),
             ],
         ),
-    ).all()
+    )
     serializer_class = serializers.IPAddressSerializer
     filterset_class = filtersets.IPAddressFilterSet
 
--- a/netbox/ipam/api/views.py
+++ b/netbox/ipam/api/views.py
@@ -1,5 +1,6 @@
 from copy import deepcopy
 
+from django.contrib.contenttypes.prefetch import GenericPrefetch
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.db import transaction
 from django.shortcuts import get_object_or_404
@@ -13,6 +14,7 @@ from rest_framework.response import Response
 from rest_framework.routers import APIRootView
 from rest_framework.views import APIView
 
+from dcim.models import Interface
 from ipam import filtersets
 from ipam.models import *
 from ipam.utils import get_next_available_prefix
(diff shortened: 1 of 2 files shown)


netbox/ipam/api/views.py
```diff
@@ -79,7 +81,9 @@ class RoleViewSet(NetBoxModelViewSet):
 
 
 class PrefixViewSet(NetBoxModelViewSet):
-    queryset = Prefix.objects.all()
+    queryset = Prefix.objects.prefetch_related(
+        "scope",
+    ).all()
```
jnovinger, line 86: The trailing `.all()` is unnecessary here (and below on `IPAddressViewSet.queryset`).

Tishka17: removed

jnovinger: Thanks, I'll get the review finished at some point today.

netbox/ipam/api/views.py
```diff
@@ -100,7 +102,14 @@ class IPRangeViewSet(NetBoxModelViewSet):
...
 class IPAddressViewSet(NetBoxModelViewSet):
-    queryset = IPAddress.objects.all()
+    queryset = IPAddress.objects.prefetch_related(
+        GenericPrefetch(
+            "assigned_object",
+            [
+                Interface.objects.select_related("device"),
+            ],
+        ),
+    )
```
jnovinger, line 112: Sorry this took so long. I really wanted to understand why we needed to use the `GenericPrefetch` here, which meant some poking at things in my local.

So, it turns out that `prefetch_related()` can handle GFKs perfectly fine on its own, withouth the `GenericPrefetch`--which I'm guessing you already knew. I was pretty sure they did, but wanted to be sure I understood that correctly. As a matter of fact, `prefetch_related()` is also prefetching related `VMInterface` and `FHRPGroup`s that are assigned via the GFK.

The interesting thing here, is what you did with the `querysets` argument, where you used `select_related()` on any associated interfaces to cut down any more incidental N+1 issues on that relation. So, I got to digging a bit and realized this is great and we can extend it a bit more, like so:

```python
    queryset = IPAddress.objects.prefetch_related(
        GenericPrefetch(
            "assigned_object",
            [
                FHRPGroup.objects.all(),  # comment about why this is here
                Interface.objects.select_related("cable", "device"),
                VMInterface.objects.select_related("virtual_machine"),
            ],
        ),
    )
```

There are three changes here I'd like to see:
1. Add `"cable"` to the list of `.select_related(...)` arguments for Interface.
2. Add a VMInterface queryset with a `.select_related("virtual_machine")` since that also gets fetch if any IP Address is assigned to a VMInterface. 
3. Add an FHRPGroup queryset added to the list for consistency. It's actually already picked up, but I think it makes sense to list it here since the others are as well so that future code readers aren't confused about why it's missing.

jnovinger (changes requested):
Hey @Tishka17 , there's one change I'd like to see for IPAddressViewSet. Other than that, this looks great. Sorry for the lengthy review time!

netbox/ipam/api/views.py
Tishka17: Sorry for long pause, I've updated prefetch according to your comment

jnovinger: No worries. Thanks! I'm looking at it now.

jnovinger (approved):
💥

PR #19308, opened 2025-04-24
Fixes: #18717 - On delete signal handling, manually save the related object in a ManyToOneRel to trigger a change record

### LINKED ISSUE #18717: DCIM Interface: ObjectChange missing for deleted untagged VLANs
### Deployment Type

NetBox Cloud (demo), Self-hosted

### NetBox Version

v4.2.0 (demo.netbox.dev) & v4.1.11 (locally)

### Python Version

3.11

### Steps to Reproduce

1. Have an Device with an Interface
2. Have two VLANs: A & B
3. Set Device Interface mode to "tagged"
4. Add VLAN A as untagged VLAN to Interface
5. Add VLAN B as tagged VLAN to Interface
6. Delete VLAN B
7. Delete VLAN A

### Expected Behavior

Interface ObjectChange `update` for each VLAN deletion:
1. `tagged_vlans` to empty list
2. `untagged_vlan` to `null`/`None`

### Observed Behavior

Only one Interface ObjectChange `update` is created for VLAN deletion: `tagged_vlans`.

No Interface ObjectChange is created for `untagged_vlan` when related VLAN is deleted. According to Interface Changelog the Interface still has an `untagged_vlan` when in fact it does not exist anymore.

bctiemann:
### Fixes: #18717

The `handle_deleted_object` signal handler ensures that change records are saved for the object being deleted, but also manually triggers a `m2m_changed` signal to ensure that related objects through many-to-many relationships are also change-logged through their own signal handlers. However, we do not handle the case where the related object is linked to the deleted object through a many-to-one relationship.

This change handles `ManyToOneRel` fields in the same area of code, by resolving the related object in the same way as with M2M fields, but instead setting the related object's field value to `None` and saving, thus ensuring a change record is created for the related object.

**Note** the check on L164 which ensures the related field is `null=True`. This is primarily to ensure that unit tests pass, as without it one of the tests fails due to trying to set a non-nullable field to `None`. However I think in practice this will not happen as in the case of any FK relationship where `null=False`, trying to delete an object which is pointed to by a related object via a many-to-one relationship will be caught by object-level validation before the deleted object handler is ever called.

### DIFF
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -2,7 +2,7 @@ import logging
 
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
-from django.db.models.fields.reverse_related import ManyToManyRel
+from django.db.models.fields.reverse_related import ManyToManyRel, ManyToOneRel
 from django.db.models.signals import m2m_changed, post_save, pre_delete
 from django.dispatch import receiver, Signal
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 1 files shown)


jnovinger (approved):
Spent some time thinking on this and trying to poke holes in it, but I got nothing.

I think in an ideal world, we'd have a test suite for `handle_deleted_object` and could add tests to make sure this behaves as expected. But, that seems out of scope for this fix.

PR #19312, opened 2025-04-24
18334 add location, device, site to module filters

### LINKED ISSUE #18334: Allow filtering modules by device, location, tenant, or site
### NetBox version

v4.2.0

### Feature type

New functionality

### Triage priority

I'm a NetBox Labs customer

### Proposed functionality

In the modules view under filter you would be able to sort by devices, locations, sites, or tenants. There are currently only 5 fields to filter the table. 
image

### Use case

The current filtering on modules is very limited. This would allow you to see where modules are located/assigned to instead of having to use an external script. Expansion of https://github.com/netbox-community/netbox/issues/8810 

### Database changes

_No response_

### External dependencies

_No response_

arthanson:
### Fixes: #18334 

Monosnap Modules | NetBox 2025-04-24 09-42-38

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1337,9 +1337,61 @@ class ModuleFilterSet(NetBoxModelFilterSet):
         lookup_expr='in',
         label=_('Module bay (ID)'),
     )
+    region_id = TreeNodeMultipleChoiceFilter(
+        queryset=Region.objects.all(),
+        field_name='device__site__region',
+        lookup_expr='in',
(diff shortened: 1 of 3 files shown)


jeremystretch (changes requested):
Please add tests for the new filters. Other than that this looks good.

jeremystretch (changes requested):
Sorry, missed this earlier but if we're going to include a Region filter we should probably have one for SiteGroup as well. That said, the FR doesn't actually call for either.

jeremystretch (approved):

PR #19316, opened 2025-04-24
Fixes: #19187 - Management command for fetching UN/LOCODE data and regenerating UN_LOCODE values

### LINKED ISSUE #19187: Wrong UN/LOCODEs in predefined base choice
### Deployment Type

Self-hosted

### NetBox Version

v4.2.3

### Python Version

3.12

### Steps to Reproduce

1. Create a custom field choice
2. Use UN/LOCODE as base choice
3. Create an object and select the unlocode

### Expected Behavior

The description/name of the unlocodes should be correct.

### Observed Behavior

Some or maybe many unlocodes in the predefined base choice list are wrong.
For example IT-BLQ (Bolladello, Varese, Italy). It is not Bolladello, it's Bologna.
Same for IT-BSI (Basciano, Teramo, Italy). It is not Basciano, it's Basiglio.

E.g if I search for Basciano:

Image

or Nieuwolda:

Image

bctiemann:
### Fixes: #19187

This adds a management command, `refresh_un_locode.py`, which fetches the latest UN/LOCODE data and formats it into the appropriate ChoiceSet tuple to assign to `UN_LOCODE`.

This is still a draft because:
- This PR currently only adds the management command, the `UN_LOCODE` values in `un_locode.py` have not yet been updated;
- There are still some values that are encoded strangely. The source files seem to be encoded in Windows-1252/`cp1252`, and that is how the script treats them; the vast majority of names imported from the Subdivisions file are displayed correctly, but some (particularly in Vietnamese and some others) appear to still be wrong. If I open the CSV files directly on my Mac all the values are interpreted and displayed correctly, so either MacOS is able to do better on-the-fly encoding detection than `chardet` or Python's `csv` module can do, or there is some yet unknown character encoding that can be used to read the files.

### DIFF
--- a/netbox/extras/management/commands/refresh_un_locode.py
+++ b/netbox/extras/management/commands/refresh_un_locode.py
@@ -1,0 +1,75 @@
+import csv
+import os
+import pprint
+from urllib.request import urlopen
+from io import BytesIO
+from zipfile import ZipFile
+
+from django.core.management.base import BaseCommand
+
+from extras.data.iso_3166 import ISO_3166
+
+
+class Command(BaseCommand):
+    help = "Import UN/LOCODE codes from the UNECE source data set and regenerate UN_LOCODES ChoiceSet values"
+    subdivisions = {}
+    countries = {}
+    code_list = []
+    subdivisions_file = None
+    code_list_files = []
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--url',
+            help="URL of CSV file archive (.zip) to download",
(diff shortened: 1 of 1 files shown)


jeremystretch:
I'm not sure it makes sense to implement this as a management command. As this functionality is for development purposes only and (presumably) doesn't interact with NetBox data directly, IMO it should be a standalone script.

arthanson (changes requested):
See Jeremy's comment

PR #19330, opened 2025-04-25
Closes #17243: Include example of image upload using REST API

### LINKED ISSUE #17243: Extend REST API documentation with an example of uploading an image attachment using `curl`
### Change Type

Addition

### Area

Integrations/API

### Proposed Changes

The process for uploading files via the REST API differs a bit from regular JSON data. A simple copy-and-paste example using `curl` would be helpful for reference and troubleshooting. For example:

```
curl -X POST \
-H "Authorization: Token $TOKEN" \
-F "object_type=dcim.site" \
-F "object_id=2" \
-F "name=attachment1.png" \
-F "image=@image.png" \
http://netbox:8000/api/extras/image-attachments/
```

jeremystretch:


jnovinger:
@jeremystretch , your issue number is off--it should be 17243 instead of 17423 (which is actually a PR).

jnovinger (approved):

PR #19332, opened 2025-04-25
18706 Fix VLAN Assignment checking

### LINKED ISSUE #18706: Changing scope of VLAN group breaks possible VLAN association
### Deployment Type

NetBox Cloud

### NetBox Version

v4.2.2

### Python Version

3.12

### Steps to Reproduce

1. create a Site "New York"
2. create as Site Group "USA"
3. assign site "New York" to Site Group "USA"
4. create a VLAN group "example group" with scope Site and and site "New York"
5. create a VLAN "example VLAN" and assign both site "New York" and and group "example group"
6. change scope of VLAN group "example group" to Site-Group and select group "USA"
7. try to create a new VLAN "example with issue" and assign both site "New York" and and group "example group"

### Expected Behavior

That the VLAN "example with issue" can be assigned to VLAN group "example group" and site "New York" at the same time.

### Observed Behavior

Step 7 will fail, even though "example VLAN"  is assigned to exactly the same set of site and group

Error message:
```
VLAN is assigned to group example group (scope: USA); cannot also assign to site New York.
```

This could be solved by:
* changing VLAN to use scope as well: `dcim.site` and `ipam.vlangroup`
* permit both but add add constrains checking during add of a VLAN and at the change of the VLAN group scope to not violate the conditions of the currently assigned VLANs to that group
* make the usage of site and group exclusive for a VLAN. Either site or group, not both.

arthanson:
### Fixes: #18706 

VLANs can currently be assigned to both a Site and SiteGroup. The SiteGroup has a scope assignment that can be to a Site or SiteGroup.  If a VLAN is assigned to both a Site and SiteGroup an error is raised if the SiteGroup assigned site doesn't match the VLAN site. However the check assumed the SiteGroup was only scoped to a Site, this fixes it to account that it can be scoped to either a Site or SiteGroup.

### DIFF
--- a/netbox/ipam/models/vlans.py
+++ b/netbox/ipam/models/vlans.py
@@ -291,7 +291,7 @@ class VLAN(PrimaryModel):
             if self.site not in self.group.scope.sites.all():
                 raise ValidationError(
                     _(
-                        "VLAN is assigned to group {group} (scope: {scope}); cannot also assign to site {site}."
+                        "The assigned site {site} is not a member of the assigned group {group} (scope: {scope})."
                     ).format(group=self.group, scope=self.group.scope, site=self.site)
                 )
 
--- a/netbox/ipam/models/vlans.py
+++ b/netbox/ipam/models/vlans.py
@@ -1,4 +1,5 @@
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
+from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres.fields import ArrayField, IntegerRangeField
 from django.core.exceptions import ValidationError
 from django.core.validators import MaxValueValidator, MinValueValidator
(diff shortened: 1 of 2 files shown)


netbox/ipam/models/vlans.py
```diff
@@ -279,12 +280,20 @@ def clean(self):
...
+                raise ValidationError(
+                    _(
+                        "VLAN is assigned to group {group} (scope: {scope}); cannot also assign to site {site}."
+                    ).format(group=self.group, scope=self.group.scope, site=self.site)
+                )
+        if self.group and self.site and self.group.scope_type == ContentType.objects.get_for_model(SiteGroup):
+            if self.site not in self.group.scope.sites.all():
+                raise ValidationError(
+                    _(
+                        "VLAN is assigned to group {group} (scope: {scope}); cannot also assign to site {site}."
```
jeremystretch, line 294: This error message should specifically indicate the issue is that the assigned site is not a member of the assigned group.

netbox/ipam/tests/test_models.py
```diff
@@ -645,3 +647,54 @@ def test_qinq_role(self):
...
+            ),
+        ))
+        vlan = VLAN(
+            name='VLAN 1',
+            vid=1,
+            group=vlangroups[0],
+            site=sites[0],
+        )
+
+        # VLAN Group 0 and 1 should be valid
```
jeremystretch, line 694:
```suggestion
        # VLAN Group 1 and 2 should be valid
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #19333, opened 2025-04-25
Fixes #17405: Added plugin icon to plugin list/detail

### LINKED ISSUE #17405: Include the plugin's icon (if any) when displaying plugins from the catalog
### NetBox version

v4.1.0

### Feature type

New functionality

### Proposed functionality

The [plugins catalog API](https://api.netbox.oss.netboxlabs.com/v1/plugins/) now includes a URL to an icon for each plugin (or a placeholder, for plugins without one). We can display this icon next to the plugin name.

### Use case

Icons help to visually distinguish icons.

### Database changes

None

### External dependencies

The plugins catalog API was recently updated to support this.

jnovinger:
### Fixes: #17405
- uses new `icon` URL value from the [plugins catalog API](https://api.netbox.oss.netboxlabs.com/v1/plugins/)
- applies minimal styling to keep image display from being to large relative to nearby text
- assumes icons are black and white and inverts the image via CSS for dark mode


netbox/project-static/styles/custom/_misc.scss
```diff
@@ -75,3 +75,15 @@ span.color-label {
 .btn-grey, .btn-gray {
   @extend .btn-secondary;
 }
+
+img.plugin-icon {
+  max-width: 1.4285em;
```
jnovinger, line 80: This was eye-balled. I didn't see any immediate variables to re-use here.

netbox/project-static/styles/custom/_misc.scss
```diff
@@ -75,3 +75,15 @@ span.color-label {
...
+  height: auto;
+}
+
+body[data-bs-theme=dark] {
+  // Assuming icon is black/white line art, invert it and tone down brightness
+  img.plugin-icon {
+    filter: invert(100%) brightness(80%);
+  }
```
jnovinger, line 88: Not loving this approach, unless we'll only ever accept black/white line-drawing-ish icons.

pheus: Just a quick drive-by thought: you could consider adding `grayscale()` to the filter as well:
```css
filter: grayscale(100%) invert() brightness(80%);
```

Also, small note - you don't need to specify `100%` for `invert`; `invert()`, `invert(1)`, and `invert(100%)` are equivalent. (See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/invert) if helpful.)

bctiemann:
I think this looks fine code-wise, but the icon looks a little too big for the text:

Screenshot 2025-04-30 at 8 45 19 AM

However I don't think that's a problem with the icon, but rather that the text in the table rows is too small / not vertically centered in the table rows throughout the UI. That's a separate issue that should be raised.

bctiemann (approved):

PR #19360, opened 2025-04-29
Closes #19358: Move release info from footer to nav menu

### LINKED ISSUE #19358: Display the version at the bottom of the navigation sidebar
### NetBox version

v4.2.8

### Feature type

Data model extension

### Proposed functionality

Move the version from the right side of the page footer to the bottom of the navigation menu.

### Use case

The NetBox version is frequently referenced for myriad reasons, and is currently accessible only at the very end of each page. Moving it to the navigation sidebar ensures that the release info is always immediately available.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:


netbox/templates/base/layout.html
```diff
@@ -51,8 +51,20 @@ <h1 class="navbar-brand pb-0">
...
         <div class="collapse navbar-collapse" id="sidebar-menu">
           {% nav %}
-        </div>
 
+          {# Release info #}
+          <div class="text-muted text-center fs-5 my-3">
+            {{ settings.RELEASE.name }}
+            {% if not settings.RELEASE.features.commercial %}
+              <div>
+                {% trans "Get" %}
```
arthanson, line 60: very minor nit: I'd either put a colon after Get or else include it in the link as it's not obvious they are links and I first hovered over Get.  Otherwise everything is good.

arthanson (dismissed):

arthanson (approved):

PR #19369, opened 2025-05-01
Closes: #19200 Add Virtual Chassis name to pane on Device View

renatoalmeidaoliveira:
### Closes: #19200 Add Virtual Chassis name to pane on Device View

- Added the Virtual Chassis information under the `Device` card.

**Observations**  

When I added that data to the `Virtual Chassis` card, it looked very different from the general UI. I mean, users normally don't expect to see a device name in the card header.  

If I add that information there to follow the usual behavior, I should add it as a clickable link. However, this would be redundant, considering the `View Virtual Chassis` link already present in that pane.  

So, adding that information in the `Device` card next to the `Rack` information makes more sense to me.


renatoalmeidaoliveira:
image

netbox/templates/dcim/device.html
```diff
@@ -26,6 +26,12 @@ <h2 class="card-header">{% trans "Device" %}</h2>
                         <th scope="row">{% trans "Location" %}</th>
                         <td>{% nested_tree object.location %}</td>
                     </tr>
+                    {% if vc_members %}
```
jeremystretch, line 29: This should be conditional on the value of `object.virtual_chassis`, rather than `vc_members`.

netbox/templates/dcim/device.html
```diff
@@ -26,6 +26,12 @@ <h2 class="card-header">{% trans "Device" %}</h2>
...
+                    <tr>
+                      <th scope="row">{% trans "Virtual Chassis" %}</th>
+                      <td>{{ object.virtual_chassis|linkify }}</td>
+                    </tr>
```
jeremystretch, line 33: Remember to indent code within `{% if %}` blocks.

jeremystretch (changes requested):

jeremystretch (approved):

PR #19388, opened 2025-05-02
Skip login for API requests when using REMOTE_AUTH_HEADER

### LINKED ISSUE #18914: API POST requests fail when REMOTE_AUTH_HEADER is enabled
### Deployment Type

Self-hosted

### NetBox Version

v4.2.4

### Python Version

3.12

### Steps to Reproduce

1. Run Netbox behind a reverse proxy which adds a custom username header
2. Enable `REMOTE_AUTH_ENABLED` and `REMOTE_AUTH_HEADER` in Netbox config
3. Make a POST API request like `curl http://localhost:8000/api/dcim/sites/ -d '{"name": "site1", "slug": "site1"}'`

### Expected Behavior

The request should succeed without "logging in".  The same as how API requests work when you're not using `REMOTE_AUTH_HEADER` and you provide an API token in the Authorization header.

### Observed Behavior

The request fails with HTTP 403:
```
{"detail":"CSRF Failed: CSRF cookie not set."}
```

The problem lies in Netbox customized version of `RemoteUserMiddleware`.  It runs `auth.login()` which causes Django to enforce CSRF.  This of course fails, because a webhook cannot contain a CSRF token.

I can workaround this by writing my own custom `REMOTE_AUTH_BACKEND` class, and using this logic to skip API requests:
```
if request.META['PATH_INFO'].startswith('/api'):
    return None
```

This avoids the `login()` and the request succeeds, but this causes another minor problem:  Every API request triggers a `user_login_failed` signal which prints this message to the log:
> Failed login attempt for username: None from 2607:fb10:7011:1::e82

Furthermore, it should be possible to use `REMOTE_AUTH_HEADER` without writing a custom class.

llamafilm:


jnovinger:
I'm closing this as I don't believe the reported issue is valid. I've asked the reporter to clarify there and will re-open this if I misunderstood.

PR #19401, opened 2025-05-05
Fixes #19396 : Set qinq_role allow_null to True

### LINKED ISSUE #19396: QinQ role not nullable
### Deployment Type

Self-hosted

### NetBox Version

v4.2.7

### Python Version

3.10

### Steps to Reproduce

1. Create (or update) a vlan like this (with the qinq_role as null):
```json
{
  "name": "Test",
  "vid":8,
  "status": "active",
  "description": "Hey",
  "qinq_role": null
}
```

### Expected Behavior

Normally this would set the role to null 

### Observed Behavior

400 / Error: Bad Request

```json
{
  "qinq_role": [
    "This field may not be blank."
  ]
}
```

Etibru:
### Fixes: #19396

- Allow vlan `qinq_role` to be set as null with the API matching the model definition.


jeremystretch (approved):
Thanks @Etibru!

PR #19418, opened 2025-05-06
19381 fix data file script sync

### LINKED ISSUE #19381: Unable to add script
### Deployment Type

Self-hosted

### NetBox Version

v4.3.0

### Python Version

3.12

### Steps to Reproduce

After upgrading to 4.3.0 I'm unable to add a custom script to Netbox. Syncing the datasource works as expected but adding the script throws an exception. When I downgrade to 4.2.9 everything works again.

Seems to be related to the changes introduced with: https://github.com/netbox-community/netbox/pull/18680/

### Expected Behavior

Script was added and can be executed from UI.

### Observed Behavior

Image

arthanson:
### Fixes: #19381 

Fix could be a one liner:
`new_file.write(self.data)` ->  `new_file.write(self.data_file.data)`

However the _write_to_disk is only ever called in sync_data if a data_file so just collapsed the function to be inline and then it can be simplified, the line `if storage.exists(path) and not overwrite:` will never evaluate to True as overwrite was always passed as True so this can be removed.

Note: This only effects the data_file path so should have no effect on S3 storage or local file storage for scripts.

### DIFF
--- a/netbox/core/models/files.py
+++ b/netbox/core/models/files.py
@@ -88,19 +88,11 @@ class ManagedFile(SyncedDataMixin, models.Model):
     def sync_data(self):
         if self.data_file:
             self.file_path = os.path.basename(self.data_path)
-            self._write_to_disk(self.full_path, overwrite=True)
 
-    def _write_to_disk(self, path, overwrite=False):
-        """
-        Write the object's data to disk at the specified path
-        """
-        # Check whether file already exists
-        storage = self.storage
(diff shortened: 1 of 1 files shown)


karimone:
I wonder why we don’t have a test covering this issue. I created a management command in my plugin to consolidate the datasource across different instances. The datasource comes from the filesystem, and when I upgraded NetBox, all the datasource command tests I wrote started failing.

bctiemann (approved):

PR #19424, opened 2025-05-07
feat(core): Add queue_name parameter to Job.enqueue() method

dianbofa:
- Added the queue_name parameter to Job.enqueue(), allowing users to specify an existing queue name.

- If queue_name is not provided, the default model queue will be used.

- This enhancement improves job scheduling flexibility by enabling targeted execution in specific queues.

PR #19433, opened 2025-05-08
Issue 19424

dianbofa:


jeremystretch:
We can accept PRs only for valid, assigned issues. It looks like you're intending to adapt your work on PR #19424. If this is the case, please update that PR directly.

PR #19447, opened 2025-05-08
Fixes #17073 - Add search index for tags

### LINKED ISSUE #17073: Include Tags in the global search.
### NetBox version

v4.0.7

### Feature type

Data model extension

### Proposed functionality

I propose that Tags are included in the global search function.

#16147 mentions this but it seems like a feature request was never opened.

### Use case

Users may decide to use Tags to document certain aspects of their environment that does not fit into existing fields in Netbox, or if they consider Tags to be more apt for their specific use case. As an example, we use the Tag "Out of Band" on both devices and circuits that are used for out of band management. 

Netbox has many ways to document ones environment, and it is difficult for users to remember exactly how everything is documented, so often times the global search will be the first step to find relevant documentation. In this case, a user would search for "OoB" or "Out of Band" and find 0 results in the global search, despite the fact that we have a tag called "Out of Band" and have several devices and circuits with this tag. Allowing Tags to be searchable would allow users to find where and how things are documented much quicker. 

### Database changes

I don't think any database changes are necessary here? 

### External dependencies

I am not aware of any new external dependencies.

pheus:
### Fixes: #17073 - Include Tags in the global search

Introduces a search index for the Tag model to enable global search for Tags. Includes fields for name, slug, and description with corresponding weight values. Display attributes are limited to the description field.

### DIFF
--- a/netbox/extras/search.py
+++ b/netbox/extras/search.py
@@ -25,6 +25,17 @@ class JournalEntryIndex(SearchIndex):
 
 
 @register_search
+class TagIndex(SearchIndex):
+    model = models.Tag
+    fields = (
+        ('name', 100),
+        ('slug', 110),
+        ('description', 500),
+    )
+    display_attrs = ('description',)
+
+
+@register_search
 class WebhookEntryIndex(SearchIndex):
     model = models.Webhook
     fields = (


jeremystretch (approved):
Thanks @pheus!

PR #19456, opened 2025-05-09
Fixes #19444: Fix change logging for contact group assignments

### LINKED ISSUE #19444: Contact group assignments are not recorded in changelog
### Deployment Type

Self-hosted

### NetBox Version

v4.3.0

### Python Version

3.10

### Steps to Reproduce

1. Create a contact group named Group 1
2. Create a contact named Contact 1 and assign it to the group
3. Inspect the creation record for the contact in the changelog

### Expected Behavior

The assigned group should appear in a `groups` list in the serialized representation of the contact.

### Observed Behavior

There is no `groups` field in the changelog record.

jeremystretch:
### Fixes: #19444

- Remove the custom ContactGroupMembership "through" model for contact group assignments
- Introduce a crafted migration to transparently handle the database changes

### DIFF
--- a/netbox/tenancy/migrations/0020_remove_contactgroupmembership.py
+++ b/netbox/tenancy/migrations/0020_remove_contactgroupmembership.py
@@ -1,0 +1,71 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tenancy', '0019_contactgroup_comments_tenantgroup_comments'),
+    ]
+
+    operations = [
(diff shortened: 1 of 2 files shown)


bctiemann:
I noticed that adding a contact to a group as well as a sub-group of that group causes the total count of contacts in the group to be +1 from what you would expect, i.e. it counts total assignments rather than distinct contacts; but that is not new behavior and is out of scope for this fix.

bctiemann (approved):

Issue #19466, opened 2025-05-12
unnecessary save action device instance when it is removed


Taurdil:
### Deployment Type

NetBox Cloud

### NetBox Version

v4.3.0

### Python Version

3.12

### Steps to Reproduce

1. Create Device
2. Create Interface of device
3. Create IPAddress from for interface
4. Set "Make this the primary IP for the device/VM" for IPAddress
5. Delete Device

### Expected Behavior

Only operations to delete all related objects (Device, Interface, IPAddress)

### Observed Behavior

Before deleting the Device, it is updated

jnovinger:
I'm accepting this as it does make sense--if we can remove extraneous queries and still accomplish the same outcome, then we should do it.

The code responsible for doing this (in this case) appears to be `ipam.signals.clear_primary_ip` and was put in place purposefully--to handle clearing the "primary IP {family}" for any devices/VMs this IP address is attached to, particularly when the device/VM is not also being deleted.

I'm not at all certain how we would go about telling whether the linked device/VM is also being deleted and therefore does not need to be updated.

jnovinger:
Apologies @Taurdil , I noticed your commit in your fork after my comment above. Is this something you like to take ownership of?

jeremystretch:
While I can appreciate wanting to optimize this process, it is working as intended. Any conditional logic implement with the intent of bypassing the update introduces inherent complexity, increasing both development burden and risk of breakage. The negligible gains that might be realized by this change warrant neither the effort nor the risk IMO.

jeremystretch:
After further consideration, I'm going to close this as the risk associated with interfering with the current workflow does not justify whatever modest performance gains might be recognized.

PR #19467, opened 2025-05-12
Fixes: #19466. Remove unnecessary saving

Taurdil:
Before the main signal code, a check is made of the initiator of the deletion process. If this is a device or a virtual machine, then further operation of the signal is interrupted.


jeremystretch:

PR #19477, opened 2025-05-13
Fixes #19464: Fix bulk editing of inventory items from device view

### LINKED ISSUE #19464: Bulk editing inventory items fails because no device is selected by default
### Deployment Type

Self-hosted

### NetBox Version

v4.3.0

### Python Version

3.12

### Steps to Reproduce

1. Create a new device
2. Create a new inventory item (Add components -> Inventory items, add some name)
3. On the device inventory item list, select the item you created and click "edit selected"
4. Do any edit to the item data (for example, set the "description" field to "test")
5. Click "Apply"

### Expected Behavior

Edits are saved

### Observed Behavior

An error pops up in the bottom right corner ("This field cannot be null.")

I believe this error happens because the "Device" field is for some reason not being populated on the bulk edit form by default. In addition to fixing the form to populate the field, it'd be nice if the error message would actually identify the field with the problem.

jeremystretch:
### Fixes: #19464

Remove the `device` initial data after instantiating the form to avoid flagging the actual `device` field as having changed data. (Ultimately we need to figure out a better way to pass contextual data under BulkEditView, but this suffices for now.)

### DIFF
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -1779,6 +1779,13 @@ class InventoryItemBulkEditForm(
     )
     nullable_fields = ('label', 'role', 'manufacturer', 'part_id', 'description')
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Remove parent device passed as context to avoid conflicts with the actual device field
+        # on this form (see bug #19464)
+        self.initial.pop('device', None)
+
 
 #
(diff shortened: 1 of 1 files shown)


arthanson (commented):
Why is the device field needed here, on all the other components the device filed does not appear - we don't allow that to be changed on any other component bulk edit that I see?

jeremystretch:
@arthanson The `device` field was added to permit moving items from one device to another.

arthanson (approved):

PR #19495, opened 2025-05-14
Fixes #19415: Increased Circuit/WirelessLink distance upper limit

### LINKED ISSUE #19415: Circuit Distances = Django Error: A field with precision 10, scale 4 must round to an absolute value less than 10^6.
### Deployment Type

NetBox Cloud

### NetBox Version

v4.2.3

### Python Version

3.10

### Steps to Reproduce

On an existing Circuit, click on "edit".
Enter a distance value of 999.99, select any unit (km, mi, etc.) and click save and it results in 1000 for distance.
Perform the same steps as above and try to save anything higher, and you will get the following error:

Server Error

<class 'django.db.utils.DataError'>

numeric field overflow
DETAIL:  A field with precision 10, scale 4 must round to an absolute value less than 10^6.

Python version: 3.11.0rc1
NetBox version: 4.2.3
Plugins: 
  nea_netbox_utils_plugin: 1.1.0
  netbox_attachments: 7.0.0
  netbox_bgp: 0.15.0
  netbox_more_metrics: 0.3.2
  netbox_napalm_plugin: 2.1.1

[Home Page](https://netbox01.point-broadband.net/)

### Expected Behavior

Unable to save any distance beyond 999.99 in any circuit distance field, when the system at 10^6 should allow for values upto 1M, and the feature should allow for a larger range to represent the overall length (in km or miles) from a circuit provider.

### Observed Behavior

This is the error you receive when attempting to save a distance value >999.99 in any unit of measure.

Server Error

<class 'django.db.utils.DataError'>

numeric field overflow
DETAIL:  A field with precision 10, scale 4 must round to an absolute value less than 10^6.

Python version: 3.11.0rc1
NetBox version: 4.2.3
Plugins: 
  nea_netbox_utils_plugin: 1.1.0
  netbox_attachments: 7.0.0
  netbox_bgp: 0.15.0
  netbox_more_metrics: 0.3.2
  netbox_napalm_plugin: 2.1.1

[Home Page](https://netbox01.point-broadband.net/)

jnovinger:
### Fixes: #19415
- Increases the longest possible distance for Circuits and WirelessLink to accommodate distances up to 100,000 miles (normalize to an upper limit of 160,934,000 meters)
- Adds form validation to provide useful feedback to user when exceeding distance upper limit
- Fixes a bug discovered during development where field errors for items in an InlineFields object were not displayed correctly


jnovinger:
For testing purposes, the new upper limit is ~10^14m, which is enough to handle the farthest man-made object from Earth with which we still have a data connection.

netbox/netbox/models/mixins.py
```diff
@@ -55,7 +55,7 @@ def clean(self):
 class DistanceMixin(models.Model):
     distance = models.DecimalField(
         verbose_name=_('distance'),
-        max_digits=8,
+        max_digits=16,
```
jeremystretch, line 58: I don't follow why this needs to be increased. `distance` is the "native" unit (e.g. miles, kilometers, etc.); six digits with a precision of two decimal points seems sufficient. Only the absolute distance (in meters) might need to be increased to support a high native value (e.g. 9,999 miles).

netbox/utilities/forms/mixins.py
```diff
@@ -44,3 +46,33 @@ def clean(self):
                 "This object has been modified since the form was rendered. Please consult the object's change "
                 "log for details."
             ))
+
+
+class DistanceValidationMixin(forms.Form):
```
jeremystretch, line 51: Can we instead just enforce a static maximum value of 100,000 (or similar) on `distance`? So long as `_abs_distance` is adjusted to support the maximum possible _absolute_ value, this should be sufficient.

For example, assume a maximum value of 100,000. The longest available unit is mile; 100,000 miles equals 160,934,000 meters. We just need to ensure that `_abs_distance` can store this theoretical maximum value.

jeremystretch (changes requested):

netbox/circuits/migrations/0052_alter_circuit__abs_distance_alter_circuit_distance.py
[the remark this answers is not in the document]
jnovinger: Can do

netbox/netbox/models/mixins.py
jnovinger: My expectation here was that this also needed to accommodate entering something in much smaller units. For instance 100,000 miles is 528,000,000 feet, which would overflow the 10^6 limit that currently exists. I can see why it wouldn't make sense to enter such a large magnitude of a smaller unit, when larger units are available.

netbox/utilities/forms/mixins.py
jnovinger: Fair enough.

The thinking behind making such a large change in potential distance upper limit was that, since there appears to be a negligible difference to storage requirements, why not set some ridiculous upper bound in absolute distance that could represent any potential circuit or wireless link possibly constructed. I suppose that qualifies as a premature optimization.

netbox/utilities/templatetags/form_helpers.py
```diff
@@ -51,8 +51,11 @@ def widget_type(field):
 def render_fieldset(form, fieldset):
     """
     Render a group set of fields.
+
+    The signature for row tuples is (layout, title, items, has_field_errors).
```
jeremystretch, line 55: This is pretty firmly beyond the scope of the bug fix. If needed, let's capture this work under a separate issue, as changes to these general-purpose tools affect the entire application.

netbox/utilities/forms/mixins.py
```diff
@@ -44,3 +46,13 @@ def clean(self):
...
```
jeremystretch, line 51: Please add DistanceValidationMixin to `__all__` above.

netbox/netbox/models/mixins.py
```diff
@@ -1,3 +1,4 @@
+
```
jeremystretch, line 1: I suspect this line break was meant to go between lines 3 and 4?

jeremystretch (changes requested):

netbox/netbox/models/mixins.py
jnovinger: It was actually because of an import I removed in a subsequent commit and forgot to remove the line break.

netbox/utilities/templatetags/form_helpers.py
jnovinger: I was mistaken that this was even needed. When I first tried to address this, the simple template modification for just InlineFields was not working without the larger change. After removing the changes to `render_fieldset()`, they still worked--which surprised me. I clearly had some other issue there that I subsequently fixed without realizing it.

jeremystretch (approved):

PR #19563, opened 2025-05-22
#19561 fix changelog table on profile page

arthanson:
### Fixes: #19561

Removes the split from the queryset for the changelog on the profile page, I don't see any reason to keep it there.  Added the pagination so can get all the info. 
Monosnap User Profile | NetBox 2025-05-22 14-33-01


jeremystretch:
Closing this in favor of PR #19630 for bug #19599.

@jakobsen-lrz thanks for flagging the above issue; it is also addressed by PR #19630.

PR #262, opened 2016-07-12
Centos 7 Install doc

aoyawale:
Helo,

I apologize for the issues before. Hopefully this document its fine. I'm not sure how you want to merge this file with the rest of the documentation but, I have added it to the installation folder.

PR #19628, opened 2025-06-02
Closes #19627: Object change migrators

jeremystretch:
### Closes: #19627

Provide ObjectChange data migrators for schema migrations starting with NetBox v4.2.0:

* circuits.0047_circuittermination__termination
* circuits.0051_virtualcircuit_group_assignment
* dcim.0200_populate_mac_addresses
* ipam.0071_prefix_scope
* ipam.0080_populate_service_parent
* tenancy.0018_contact_groups
* virtualization.0044_cluster_scope
* virtualization.0048_populate_mac_addresses

The order of operations necessary for testing this PR in conjunction with the netbox-branching plugin as as follows.

1. Install NetBox v4.1.9 and populate some [data](https://github.com/netbox-community/netbox-demo-data/blob/master/sql/netbox-demo-v4.1.sql).
2. Install and enable [netbox-branching](https://github.com/netboxlabs/netbox-branching) by checking out the `main` branch of the repo. (We need FR [#262](https://github.com/netboxlabs/netbox-branching/issues/262) but v0.5.6 hasn't been released yet.)
3. Create branches and perform object changes to generate branch data for testing.
4. Upgrade to NetBox v4.3.1, then switch to the branch for this PR (`19627-object-change-migrators`).
5. Check out the `1-branch-migrations` branch of netbox-branching. (This emulates upgrading to v0.6.0 of the plugin.)
6. Test migrating and merging open branches that were created under NetBox v4.1.9.

### DIFF
--- a/netbox/circuits/migrations/0047_circuittermination__termination.py
+++ b/netbox/circuits/migrations/0047_circuittermination__termination.py
@@ -1,4 +1,5 @@
 import django.db.models.deletion
+from django.contrib.contenttypes.models import ContentType
 from django.db import migrations, models
 
 
@@ -49,3 +50,28 @@ class Migration(migrations.Migration):
         # Copy over existing site assignments
         migrations.RunPython(code=copy_site_assignments, reverse_code=migrations.RunPython.noop),
     ]
+
+
+def oc_circuittermination_termination(objectchange, revert):
+    site_ct = ContentType.objects.get_by_natural_key('dcim', 'site').pk
+    provider_network_ct = ContentType.objects.get_by_natural_key('circuits', 'providernetwork').pk
+    for data in (objectchange.prechange_data, objectchange.postchange_data):
+        if data is None:
+            continue
+        if site_id := data.get('site'):
+            data.update({
+                'termination_type': site_ct,
+                'termination_id': site_id,
+            })
+        elif provider_network_id := data.get('provider_network'):
+            data.update({
+                'termination_type': provider_network_ct,
+                'termination_id': provider_network_id,
+            })
+        data.pop('site')
+        data.pop('provider_network')
+
+
+objectchange_migrators = {
+    'circuits.circuittermination': oc_circuittermination_termination,
+}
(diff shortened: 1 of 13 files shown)


netbox/circuits/migrations/0047_circuittermination__termination.py
```diff
@@ -49,3 +50,28 @@ class Migration(migrations.Migration):
...
+                'termination_type': site_ct,
+                'termination_id': site_id,
+            })
+        elif provider_network_id := data.get('provider_network'):
+            data.update({
+                'termination_type': provider_network_ct,
+                'termination_id': provider_network_id,
+            })
+        data.pop('site', None)
+        data.pop('provider_network', None)
```
jnovinger: Just want to clarify that we really do want to do these `data.pop(...)` actions in this migrator, even though the corresponding model migrations (`RemoveField`) are in another migration.

netbox/dcim/migrations/0200_populate_mac_addresses.py
```diff
@@ -51,3 +53,42 @@ class Migration(migrations.Migration):
             name='mac_address',
         ),
     ]
+
+
+def oc_interface_primary_mac_address(objectchange, reverting):
```
jnovinger, line 59: Might be worth leaving a note for future maintainers to look at the migrator for VMInterface as well if making changes here.

netbox/circuits/migrations/0047_circuittermination__termination.py
jeremystretch: That's a good callout. I might be proven wrong, but I think it's safe to keep them here. I'd rather not split apart the migrator logic if we can avoid it.

jeremystretch: Actually yeah, let's split it out to ensure the logic lives alongside the specific operation with which it's associated.

jnovinger (approved):

PR #19635, opened 2025-06-03
#19634 Update values to ensure consistency when referencing values set

### LINKED ISSUE #19634: Improve Configuration Setting Consistency with Values
### Change Type

### Area

Configuration

### Proposed Changes

Found that there are additional configuration items that should be updated accordingly as was done last time. For example, under the section titled REMOTE_AUTH_AUTO_CREATE_GROUPS has the text .."If true," which should be changed to "If `True`,"

mr1716:
### Fixes: #19634

Fixes inconsistencies in how documentation values set are referenced.

### DIFF
--- a/docs/configuration/error-reporting.md
+++ b/docs/configuration/error-reporting.md
@@ -4,7 +4,7 @@
 
 Default: `None`
 
-Defines a Sentry data source name (DSN) for automated error reporting. `SENTRY_ENABLED` must be True for this parameter to take effect. For example:
+Defines a Sentry data source name (DSN) for automated error reporting. `SENTRY_ENABLED` must be `True` for this parameter to take effect. For example:
 
 ```
 SENTRY_DSN = "https://examplePublicKey@o0.ingest.sentry.io/0"
@@ -16,7 +16,7 @@ SENTRY_DSN = "https://examplePublicKey@o0.ingest.sentry.io/0"
 
 Default: `False`
 
-Set to True to enable automatic error reporting via [Sentry](https://sentry.io/).
+Set to `True` to enable automatic error reporting via [Sentry](https://sentry.io/).
 
 !!! note
     The `sentry-sdk` Python package is required to enable Sentry integration.
@@ -27,7 +27,7 @@ Set to True to enable automatic error reporting via [Sentry](https://sentry.io/)
 
 Default: `1.0` (all)
 
-The sampling rate for errors. Must be a value between 0 (disabled) and 1.0 (report on all errors).
+The sampling rate for errors. Must be a value between `0` (disabled) and `1.0` (report on all errors).
 
 ---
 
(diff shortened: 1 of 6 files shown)


docs/configuration/error-reporting.md
```diff
@@ -27,7 +27,7 @@ Set to True to enable automatic error reporting via [Sentry](https://sentry.io/)
 
 Default: `1.0` (all)
 
-The sampling rate for errors. Must be a value between 0 (disabled) and 1.0 (report on all errors).
+The sampling rate for errors. Must be a value between `0` (disabled) and `1.0` (report on all errors).
```
jeremystretch: I can understand the argument for explicitly formatting literals such as True and False as code, however this is unnecessary for numeric values. Please revert these changes.

docs/configuration/required-parameters.md
```diff
@@ -2,12 +2,12 @@
 
 ## ALLOWED_HOSTS
 
-This is a list of valid fully-qualified domain names (FQDNs) and/or IP addresses that can be used to reach the NetBox service. Usually this is the same as the hostname for the NetBox server, but can also be different; for example, when using a reverse proxy serving the NetBox website under a different FQDN than the hostname of the NetBox server. To help guard against [HTTP Host header attacks](https://docs.djangoproject.com/en/3.0/topics/security/#host-headers-virtual-hosting), NetBox will not permit access to the server via any other hostnames (or IPs).
+This is a list of valid fully-qualified domain names (FQDNs) and/or IP addresses that can be used to reach the NetBox service. Usually this is the same as the hostname for the NetBox server, but can also be different; for example, when using a reverse proxy serving the NetBox website under a different FQDN than the hostname of the NetBox server. To help guard against [HTTP Host header attacks](https://docs.djangoproject.com/en/5.2/topics/security/#host-headers-virtual-hosting), NetBox will not permit access to the server via any other hostnames (or IPs).
```
jeremystretch: Except where necessary to reference a specific release, links to the Django documentation should point to `https://docs.djangoproject.com/en/stable/...`. This is an evergreen link that redirects to the documentation for the current stable release.
```suggestion
This is a list of valid fully-qualified domain names (FQDNs) and/or IP addresses that can be used to reach the NetBox service. Usually this is the same as the hostname for the NetBox server, but can also be different; for example, when using a reverse proxy serving the NetBox website under a different FQDN than the hostname of the NetBox server. To help guard against [HTTP Host header attacks](https://docs.djangoproject.com/en/stable/topics/security/#host-headers-virtual-hosting), NetBox will not permit access to the server via any other hostnames (or IPs).
```

jeremystretch (changes requested):

docs/configuration/required-parameters.md
mr1716: Fixed. Thanks.

docs/configuration/error-reporting.md
mr1716: Change made

mr1716:
@jeremystretch thanks for the feedback. Changes made. Please let me know if other changes are necessary and needed

docs/configuration/error-reporting.md
```diff
@@ -62,7 +62,7 @@ SENTRY_TAGS = {
 
 Default: `0` (disabled)
 
-The sampling rate for transactions. Must be a value between 0 (disabled) and 1.0 (report on all transactions).
+The sampling rate for transactions. Must be a value between `0` (disabled) and `1.0` (report on all transactions).
```
jeremystretch:
```suggestion
The sampling rate for transactions. Must be a value between 0 (disabled) and 1.0 (report on all transactions).
```

jeremystretch (approved):

PR #19641, opened 2025-06-04
Closes #19545 - Suppress housekeeping cron output

### LINKED ISSUE #19545: Housekeeping task output triggers emails via anacron
### Deployment Type

Self-hosted

### NetBox Version

v4.3.0

### Python Version

3.10

### Steps to Reproduce

With the housekeeping script that's default to netbox install, the script is echoing text to stdout, since all cron's output is redirected to email, I am getting a daily mail with the result of the housekeeping (because I have a mailer configured to my address)

https://github.com/netbox-community/netbox/blob/88565e8f680a5b64f1a22b67f007f15acf8f1373/contrib/netbox-housekeeping.sh#L2

If the script was outputting to /dev/null it would only sent emails when something is throwing errors, otherwise discard them

> /opt/netbox/venv/bin/python /opt/netbox/netbox/manage.py housekeeping >dev/null

Another option would be to write the output to a log file:

> /opt/netbox/venv/bin/python /opt/netbox/netbox/manage.py housekeeping >/var/log/netbox/housekeeping.log

### Expected Behavior

There should be only emails when there are errors, otherwise the output should be logged or discarded only.

### Observed Behavior

A mail is being sent every time the script is run through crontab

pheus:
### Fixes: #19545 - Housekeeping task output triggers emails via anacron

Enhances the NetBox housekeeping cron job by adding optional logging functionality. By default, the cron job now suppresses its output for quieter operation. Users can optionally direct output to a log file for monitoring or troubleshooting as needed.

### DIFF
--- a/contrib/netbox-housekeeping.sh
+++ b/contrib/netbox-housekeeping.sh
@@ -6,4 +6,16 @@
 #
 # If NetBox has been installed into a nonstandard location, update the paths
 # below.
-/opt/netbox/venv/bin/python /opt/netbox/netbox/manage.py housekeeping
+# By default, all output is discarded. Uncomment the log lines below to enable
+# logging.
+
+# Run quietly (no log)
+# --------------------
+/opt/netbox/venv/bin/python /opt/netbox/netbox/manage.py housekeeping > /dev/null 2>&1
+
+# Run with logging (append output to log file)
+# --------------------------------------------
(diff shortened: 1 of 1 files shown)


jeremystretch:
Thanks for tackling this @pheus. Unfortunately it should not have been classified as a bug and no changes are needed; please see my comment [here](https://github.com/netbox-community/netbox/issues/19545#issuecomment-2944047973).

PR #19646, opened 2025-06-04
Fixes: #19492: Add Save Button to Script Output Window

### LINKED ISSUE #19492: Add Save Button to Script Output Window
### NetBox version

v4.2.6

### Feature type

New functionality

### Proposed functionality

It would be an improvement to user experience if there were a Save button on a custom's script output panel so that it can be readily saved to a file.  The button, ideally, would prompt a user to enter a filename and then have the file downloaded after submitting the filename.

### Use case

I need to create CSV files from the data in the database and having the option to save the data would be convenient.  Yes, the output can be selected, copied, and then pasted in a file, and then saved, but having the save option is a much better user experience.

### Database changes

None

### External dependencies

None

Omripresent:
### Fixes: #19492: Add Save Button to Script Output Window
* Added download button to `script_result.html` template following the same model of `object_render_config.html`
* Added logic to `ScriptResultView.get` function to trigger content download when matching log or output.
    * Output export responds with the raw content of the script result
    * Log export responds with a CSV formatted content of the log table matching the visible table columns (works for both scripts and reports)

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -1482,31 +1482,13 @@ class ScriptResultView(TableMixin, generic.ObjectView):
         if job.completed:
             table = self.get_table(job, request, bulk_actions=False)
 
-        if job.completed and job.data:
+        if job.completed:
             # If a direct export output has been requested, return the job data content as a
             # downloadable file.
-            if request.GET.get('export') == 'output' and job.data.get('output'):
-                content = job.data['output'].encode()
+            if request.GET.get('export') == 'output':
+                content = (job.data.get("output") or "").encode()
                 response = HttpResponse(content, content_type='text')
                 filename = f"{job.object.name or 'script-output'}_{job.completed.strftime('%Y-%m-%d-%H-%M-%S')}.txt"
-                response['Content-Disposition'] = f'attachment; filename="{filename}"'
-                return response
-
-            # If a direct export log has been requested, return the job log content as a
-            # downloadable CSV file.
-            if request.GET.get('export') == 'log' and table:
-                # Filter generator for visible columns
-                def column_filter(data):
-                    for item in data:
-                        yield dict((k, v) for k, v, in item.items() if k in table.columns.names())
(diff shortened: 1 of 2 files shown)


netbox/extras/views.py
```diff
@@ -1479,6 +1482,34 @@ def get(self, request, **kwargs):
...
+                # Write a CSV to a string buffer for content
+                buf = StringIO()
+                writer = csv.DictWriter(buf, fieldnames=table.columns.names())
+                writer.writeheader()
+                writer.writerows(column_filter(table.data))
+                buf.seek(0)
+                response = HttpResponse(buf.read().encode(), content_type='text/csv')
+                filename = f"{job.object.name or 'script-log'}_{job.completed.strftime('%Y-%m-%d-%H-%M-%S')}.csv"
+                response['Content-Disposition'] = f'attachment; filename="{filename}"'
+                return response
```
jeremystretch: The FR only calls for saving the raw output of a custom script (from the "Output" tab in the UI). Recompiling the entire job log falls outside this scope, so let's omit it. (This would necessitate a discussion around exporting log data for all types of jobs, under a separate FR.)

netbox/extras/views.py
```diff
@@ -1479,6 +1482,34 @@ def get(self, request, **kwargs):
         if job.completed:
             table = self.get_table(job, request, bulk_actions=False)
 
+        if job.completed and job.data:
```
jeremystretch: The export function should still work even if the data is empty. Otherwise, the user will not be able to determine if the request is being processed as intended.

jeremystretch (changes requested):

netbox/extras/views.py
Omripresent: Noted, will remove.

netbox/extras/views.py
Omripresent: Will do, should I update the download button display on job status completed instead of checking the output?

PR #19649, opened 2025-06-05
#19638 Update Django URLs To Stable Version

### LINKED ISSUE #19638: Update Documentation Django URL Links From Outdated Version
### Change Type

Correction

### Area

Other

### Proposed Changes

To help guard against [HTTP Host header attacks](https://docs.djangoproject.com/en/3.0/topics/security/#host-headers-virtual-hosting), NetBox will not permit access to the server via any other hostnames (or IPs). But when visiting that link, it says "This document is for an insecure version of Django that is no longer supported. Please upgrade to a newer release!" There are several other links like this as well. This ticket is to fix that.

And instead of referencing specific Django versions in the documentation, this ticket will use the …/en/stable/… in the URL. This is so that these links won’t need to get changed in the future and remain consistent.

mr1716:
### Fixes: #19638

Changes to stable versions of Django URL

### DIFF
--- a/docs/configuration/security.md
+++ b/docs/configuration/security.md
@@ -92,7 +92,7 @@ If `True`, the cookie employed for cross-site request forgery (CSRF) protection 
 
 Default: `[]`
 
-Defines a list of trusted origins for unsafe (e.g. `POST`) requests. This is a pass-through to Django's [`CSRF_TRUSTED_ORIGINS`](https://docs.djangoproject.com/en/stable/ref/settings/#std:setting-CSRF_TRUSTED_ORIGINS) setting. Note that each host listed must specify a scheme (e.g. `http://` or `https://).
+Defines a list of trusted origins for unsafe (e.g. `POST`) requests. This is a pass-through to Django's [`CSRF_TRUSTED_ORIGINS`](https://docs.djangoproject.com/en/stable/ref/settings/#csrf-trusted-origins) setting. Note that each host listed must specify a scheme (e.g. `http://` or `https://).
 
 ```python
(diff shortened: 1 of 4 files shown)


docs/configuration/security.md
```diff
@@ -92,7 +92,7 @@ If `True`, the cookie employed for cross-site request forgery (CSRF) protection
 
 Default: `[]`
 
-Defines a list of trusted origins for unsafe (e.g. `POST`) requests. This is a pass-through to Django's [`CSRF_TRUSTED_ORIGINS`](https://docs.djangoproject.com/en/4.0/ref/settings/#std:setting-CSRF_TRUSTED_ORIGINS) setting. Note that each host listed must specify a scheme (e.g. `http://` or `https://).
+Defines a list of trusted origins for unsafe (e.g. `POST`) requests. This is a pass-through to Django's [`CSRF_TRUSTED_ORIGINS`](https://docs.djangoproject.com/en/stable/ref/settings/#std:setting-CSRF_TRUSTED_ORIGINS) setting. Note that each host listed must specify a scheme (e.g. `http://` or `https://).
```
jeremystretch: Updating the named anchor
```suggestion
Defines a list of trusted origins for unsafe (e.g. `POST`) requests. This is a pass-through to Django's [`CSRF_TRUSTED_ORIGINS`](https://docs.djangoproject.com/en/stable/ref/settings/#csrf-trusted-origins) setting. Note that each host listed must specify a scheme (e.g. `http://` or `https://).
```

jeremystretch (changes requested):

jeremystretch (approved):
Thanks @mr1716!

PR #19651, opened 2025-06-05
19644 Make atomic use correct database instead of default

arthanson:
### Fixes: #19644 

transaction.atomic will go to the default database, this is normally not a problem but if multiple schemas are databases are configured using a router (like with the branching plugin) this will be incorrect as the atomic wrapper should be done on the database the changes are made on.

In jobs.py the script running code is a special case as the script could potentially modify models in multiple schemas so both the default database needs a wrapper and any model specific ones - arbitrarily used device as the model to key off, but any ChangeLogged model could be used.


netbox/dcim/views.py
```diff
@@ -124,7 +124,7 @@ def post(self, request):
 
             if form.is_valid():
 
-                with transaction.atomic():
+                with transaction.atomic(using=router.db_for_write(self.queryset.model)):
```
jeremystretch: I think we should pass the Cable model here, as those are what's being deleted within the transaction.
```suggestion
                with transaction.atomic(using=router.db_for_write(Cable)):
```

netbox/dcim/utils.py
```diff
@@ -1,6 +1,6 @@
 from django.apps import apps
 from django.contrib.contenttypes.models import ContentType
-from django.db import transaction
+from django.db import transaction, router
```
jeremystretch:
```suggestion
from django.db import router, transaction
```

netbox/ipam/api/views.py
```diff
@@ -2,7 +2,7 @@
 
 from django.contrib.contenttypes.prefetch import GenericPrefetch
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
-from django.db import transaction
+from django.db import transaction, router
```
jeremystretch:
```suggestion
from django.db import router, transaction
```

netbox/circuits/views.py
```diff
@@ -1,5 +1,5 @@
 from django.contrib import messages
-from django.db import transaction
+from django.db import transaction, router
```
jeremystretch:
```suggestion
from django.db import router, transaction
```

netbox/dcim/views.py
```diff
@@ -1,7 +1,7 @@
 from django.contrib import messages
 from django.contrib.contenttypes.models import ContentType
 from django.core.paginator import EmptyPage, PageNotAnInteger
-from django.db import transaction
+from django.db import transaction, router
```
jeremystretch:
```suggestion
from django.db import router, transaction
```

netbox/netbox/api/viewsets/__init__.py
```diff
@@ -2,7 +2,7 @@
 from functools import cached_property
 
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
-from django.db import transaction
+from django.db import transaction, router
```
jeremystretch:
```suggestion
from django.db import router, transaction
```

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -1,5 +1,5 @@
 from django.core.exceptions import ObjectDoesNotExist
-from django.db import transaction
+from django.db import transaction, router
```
jeremystretch:
```suggestion
from django.db import router, transaction
```

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -6,7 +6,7 @@
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRel
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import FieldDoesNotExist, ObjectDoesNotExist, ValidationError
-from django.db import transaction, IntegrityError
+from django.db import transaction, IntegrityError, router
```
jeremystretch:
```suggestion
from django.db import IntegrityError, router, transaction
```

netbox/netbox/views/generic/feature_views.py
```diff
@@ -1,7 +1,7 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.contenttypes.models import ContentType
 from django.contrib import messages
-from django.db import transaction
+from django.db import transaction, router
```
jeremystretch:
```suggestion
from django.db import router, transaction
```

netbox/virtualization/views.py
```diff
@@ -297,7 +297,7 @@ def post(self, request, pk):
         if form.is_valid():
 
             device_pks = form.cleaned_data['devices']
-            with transaction.atomic():
+            with transaction.atomic(using=router.db_for_write(self.queryset.model)):
```
jeremystretch:
```suggestion
            with transaction.atomic(using=router.db_for_write(Device)):
```

netbox/virtualization/views.py
```diff
@@ -332,7 +332,7 @@ def post(self, request, pk):
             if form.is_valid():
 
                 device_pks = form.cleaned_data['pk']
-                with transaction.atomic():
+                with transaction.atomic(using=router.db_for_write(self.queryset.model)):
```
jeremystretch:
```suggestion
                with transaction.atomic(using=router.db_for_write(Device)):
```

netbox/extras/jobs.py
```diff
@@ -39,10 +40,15 @@ def run_script(self, script, request, data, commit):
 
         try:
             try:
+                # A script can modify multiple models so need to do an atomic lock on
+                # both the default database (for non ChangeLogged models) and potentially
+                # any other database (for ChangeLogged models) - choosing Device as
+                # the model to use as it has ChangeLoggingMixin
```
jeremystretch: I'm not sure this is something we can reasonably tackle here; it may need to be offloaded to the script itself, as it's not possible to know what models the script might mutate, or the database routers that might be in place. IMO we should err on the safe side and leave it as-is unless we can identify a more robust solution.

jeremystretch (changes requested):

netbox/extras/jobs.py
```diff
@@ -39,6 +39,10 @@ def run_script(self, script, request, data, commit):
...
```
jeremystretch: We should keep this comment but remove the bit referencing the device model.

jeremystretch (changes requested):

jeremystretch (approved):

PR #19678, opened 2025-06-06
Fixes #19677 - Call none() when VRF is None in IPAddress and Prefix `filter_present_in_vrf`filters

### LINKED ISSUE #19677: IPAM FilterSet: `filter_present_in_vrf` method returns queryset.none instead of calling it when VRF is None
### Deployment Type

Self-hosted

### NetBox Version

v4.3.2

### Python Version

3.11

### Steps to Reproduce

1. Launch the NetBox shell:

    ```bash
    python3 manage.py nbshell
    ```

2. Run the following:

    ```python
    from ipam.models import IPAddress
    from ipam.filtersets import IPAddressFilterSet

    qs = IPAddress.objects.all()
    filterset = IPAddressFilterSet()

    # Intentionally passing `None` as the `vrf` parameter
    result = filterset.filter_present_in_vrf(qs, 'present_in_vrf', None)
    print(result)
    ```

3. Observe the result when `vrf` is `None`.

### Expected Behavior

When `vrf` is `None`, the method should return an **empty queryset** like:

```python
<RestrictedQuerySet []>
```

### Observed Behavior

The method instead returns a **method reference**:

```python
<bound method QuerySet.none of <RestrictedQuerySet [...]>>
```

This happens because `queryset.none` is returned without being called - i.e., it's missing parentheses. This causes unexpected behavior when the return value is used downstream, potentially leading to runtime issues or incorrect query evaluation.

https://github.com/netbox-community/netbox/blob/2680f855ffaef0a876bac811d4cd56198ae3f329/netbox/ipam/filtersets.py#L449-L456

https://github.com/netbox-community/netbox/blob/2680f855ffaef0a876bac811d4cd56198ae3f329/netbox/ipam/filtersets.py#L730-L736

pheus:
### Fixes: #19677 - IPAM FilterSet: filter_present_in_vrf method returns queryset.none instead of calling it when VRF is None

Ensures the `queryset.none()` method is called properly with parentheses. This fixes a potential issue where the method would not execute as intended, improving the stability and correctness of the filter logic.

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -449,7 +449,7 @@ class PrefixFilterSet(NetBoxModelFilterSet, ScopedFilterSet, TenancyFilterSet, C
     @extend_schema_field(OpenApiTypes.STR)
     def filter_present_in_vrf(self, queryset, name, vrf):
         if vrf is None:
-            return queryset.none
+            return queryset.none()
         return queryset.filter(
             Q(vrf=vrf) |
             Q(vrf__export_targets__in=vrf.import_targets.all())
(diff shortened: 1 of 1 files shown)


pheus:
I'm sorry for opening this PR before the issue was triaged — I understand that may be a bit premature. This is more of a small drive-by fix that I wanted to wrap up while I had a moment, as I may be quite busy in the coming days.

I hope that's okay, and I really appreciate your understanding.
Thank you for the fantastic work on this project - it's deeply appreciated!

jnovinger (approved):

PR #19679, opened 2025-06-06
Closes #19499 - Add WirelessLink Bulk Import Support by Device and Interface Names

### LINKED ISSUE #19499: Wireless Link Bulk Import-Device/interface
### NetBox version

v4.1.11

### Feature type

### Proposed functionality

The bulk import functionality for wireless links (https://netbox-demo.netboxlabs.com/wireless/wireless-links/import/) has interface_a and interface_z. What it should have is device_a, interface_a, device_z, interface_z. That would allow us to bulk-import our wireless links.

### Use case

I am documenting a wireless ISP where all our radios have the similar interface names (they are unique to the device, not globally unique). I'd like to import our wireless links rather than building them manually. I'm willing to use json / yaml but I cannot figure how to do it even with that. I believe that this functionality would need to be added.

### Database changes

None

### External dependencies

None

pheus:
### Fixes: #19499 - Wireless Link Bulk Import-Device/interface

* Enhances the WirelessLink bulk import form to support specifying site and device for both terminations (A and B), with dynamic queryset filtering based on selection. Also allows interfaces to be referenced by name (instead of ID) during import, improving usability for CSV workflows.
* Modifies the CSV data structure in WirelessLink test cases to include device names alongside interface names for both terminations.

### DIFF
--- a/netbox/wireless/forms/bulk_import.py
+++ b/netbox/wireless/forms/bulk_import.py
@@ -2,7 +2,7 @@ from django.utils.translation import gettext_lazy as _
 
 from dcim.choices import LinkStatusChoices
 from dcim.forms.mixins import ScopedImportForm
-from dcim.models import Interface
+from dcim.models import Device, Interface, Site
 from ipam.models import VLAN
 from netbox.choices import *
 from netbox.forms import NetBoxModelImportForm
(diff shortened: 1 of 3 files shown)


arthanson (changes requested):
I created the device and interfaces and did the import correctly, that all worked fine. I removed the WirelessLink that was created and re-did the import but I changed device_a to a different device that didn't have the interface and imported and got the crash below.

May be an existing issue that was exposed by these changes but looks like some more validation might be needed.

```
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/views/generic/bulk_views.py", line 505, in post
    new_objs = self.create_and_update_objects(form, request)
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/views/generic/bulk_views.py", line 463, in create_and_update_objects
    if model_form.is_valid():
       ~~~~~~~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/forms/forms.py", line 206, in is_valid
                                 ^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/forms/forms.py", line 201, in errors
    self.full_clean()
    ~~~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/forms/forms.py", line 339, in full_clean
    self._post_clean()
    ~~~~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/forms/base.py", line 78, in _post_clean
    return super()._post_clean()
           ~~~~~~~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/forms/models.py", line 498, in _post_clean
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/db/models/base.py", line 1654, in full_clean
    self.clean()
    ~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/netbox/wireless/models.py", line 201, in clean
    if self.interface_a.type not in WIRELESS_IFACE_TYPES:
       ^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/db/models/fields/related_descriptors.py", line 271, in __get__
    raise self.RelatedObjectDoesNotExist(
        "%s has no %s." % (self.field.model.__name__, self.field.name)
    )

```

pheus:
Thanks for testing that out and for the detailed feedback!

You're absolutely right - that crash shouldn’t happen. The issue you encountered when `device_a` didn’t have the specified interface is definitely concerning.

I’ll take a closer look at the interface validation. Adding an explicit check to ensure the interface exists *on the given device* before attempting to create the `WirelessLink` would definitely improve robustness here.

Really appreciate you catching that!

pheus:
I tracked down the root cause of the crash.

The issue stems from the model’s `clean()` method, which assumes that `interface_a` and `interface_b` are always set. However, during bulk import, it’s possible that one or both interfaces are missing (e.g., due to a typo or referencing a non-existent interface on the specified device). This causes an exception to be raised before the form validation can properly handle the error.

To resolve this, I’ve updated the `clean()` method to first check whether `interface_a` and `interface_b` are present before performing further validation. With this change, the crash no longer occurs, and the import form now correctly flags missing or invalid interfaces with a validation error instead of an unhandled exception.

Let me know if you'd prefer a different approach, but this seems like a safe and targeted fix that improves robustness.

arthanson (approved):
Thanks @pheus

mblumber1:
Thank you!

PR #19681, opened 2025-06-06
#19680 fix deletion dependency order for GenericRelations

### LINKED ISSUE #19680: ObjectChange incorrect ordering for GenericRelation references
### Deployment Type

Self-hosted

### NetBox Version

v4.3.2

### Python Version

3.12

### Steps to Reproduce

1. Create a device with an interface and connect a cable to it to another interface (you can use the imports below)
2. delete the device
3. Check the ChangeLog for the ordering of the Change Logs

```
Create Device Role "test":
name,slug,color
"test","test",888888

Create Manufacturer "test":
name,slug
"test","test"

Create Device Type "test":
manufacturer,model,slug,u_height
"test","test","test",1.0

Create Site "test":
name,slug,status
"test","test",active

Create Device "test" and "test2":
role,manufacturer,device_type,status,site,name
"test","test","test",active,"test","test"
"test","test","test",active,"test","test2"

Create Interfaces on the 2 test devices:
device,name,type
test,1,1000base-tx
test2,1,1000base-tx

Create Cable:
side_a_device,side_a_type,side_a_name,side_b_device,side_b_type,side_b_name,label
test,dcim.interface,1,test2,dcim.interface,1,"test:1,test2:1"
```

### Expected Behavior

The changelog for this deletion (by date-time) should be:
Device
Interface
CableTermination

This is because of the dependency chain: the CableTermination relies on the Interface and the Interface relies on the Device.  This is required when playing back the ChangeLog to reconstruct data, otherwise the CableTermination will get created before the Interface and fail.

### Observed Behavior

You will see the order (by date-time) of the ChangeLog records will be:
Device
CableTermination
Interface
Image

arthanson:

### DIFF
--- a/netbox/netbox/models/features.py
+++ b/netbox/netbox/models/features.py
@@ -16,6 +16,7 @@ from extras.choices import *
 from extras.constants import CUSTOMFIELD_EMPTY_VALUES
 from extras.utils import is_taggable
 from netbox.config import get_config
+from netbox.models.deletion import DeleteMixin
 from netbox.registry import registry
 from netbox.signals import post_clean
 from utilities.json import CustomFieldJSONEncoder
@@ -45,7 +46,7 @@ __all__ = (
 # Feature mixins
 #
 
-class ChangeLoggingMixin(models.Model):
+class ChangeLoggingMixin(DeleteMixin, models.Model):
     """
     Provides change logging support for a model. Adds the `created` and `last_updated` fields.
     """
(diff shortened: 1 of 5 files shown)


netbox/netbox/models/features.py
```diff
@@ -45,7 +46,7 @@
 # Feature mixins
 #
 
-class ChangeLoggingMixin(models.Model):
+class ChangeLoggingMixin(DeleteMixin, models.Model):
```
jnovinger, line 49: @arthanson , just to be clear, we only want to employee this mechanism for `ChangedLoggingMixin` and no other, correct?

jnovinger (commented):
Just had the one question. Otherwise, this looks pretty sane and well contained.

netbox/netbox/models/features.py
arthanson: Yes, it only matters in that case.  This changes the order that the delete signal handlers are sent out, but the only reason we care about that is what order the ObjectChange records are written out, so this only effects models derived from ChangeLoggingMixin.

jnovinger (approved):

PR #19689, opened 2025-06-10
Closes #19571: Create expansion_card.json

### LINKED ISSUE #19571: Module Type Profiles for interfaces
### NetBox version

v.4.3.1

### Feature type

Data model extension

### Proposed functionality

To add as part of the default module type profiles one for interfaces either for servers or switches. Not sure it would need any special attributes but for completeness.

### Use case

Should allow for all module types to be profiled under one of the default options, which would allow us in the future to enforce a profile for all module types within the devicetype-library repo

### Database changes

_No response_

### External dependencies

_No response_

harryajc:
### Adds Module Type as discussed in: #19571

Adds the Expansion card default module profile type

### DIFF
--- a/netbox/dcim/migrations/initial_data/module_type_profiles/expansion_card.json
+++ b/netbox/dcim/migrations/initial_data/module_type_profiles/expansion_card.json
@@ -2,7 +2,7 @@
     "name": "Expansion card",
     "schema": {
         "properties": {
-            "Connector type": {
+            "connector_type": {
                 "type": "string",
                 "description": "Connector type e.g. PCIe x4"
             },
--- a/netbox/dcim/migrations/initial_data/module_type_profiles/expansion_card.json
+++ b/netbox/dcim/migrations/initial_data/module_type_profiles/expansion_card.json
@@ -1,0 +1,15 @@
+{
+    "name": "Expansion card",
+    "schema": {
+        "properties": {
+            "Connector type": {
+                "type": "string",
+                "description": "Connector type e.g. PCIe x4"
+            },
+            "bandwidth": {
+                "type": "integer",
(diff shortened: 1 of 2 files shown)


jeremystretch:
The proposed change will only populate the addition profile type for NetBox installations which have yet been upgraded to v4.3.0 or later. Is that the intended effect?

I don't want to get into a pattern of introducing additional profile types sporadically. The few which are provided for a default installation exist solely to provide examples.

harryajc:
Yes, this was as discussed in the issue #19571

jnovinger (changes requested):
@harryajc , apologies for letting this languish.

We need to add `exapansion_card.json` to the `initial_profiles` tuple in `0206_load_module_type_profiles.load_initial_data`, otherwise the migration will not pick the new profile up.

harryajc:
> @harryajc , apologies for letting this languish.
> 
> We need to add `exapansion_card.json` to the `initial_profiles` tuple in `0206_load_module_type_profiles.load_initial_data`, otherwise the migration will not pick the new profile up.

@jnovinger All done

netbox/dcim/migrations/initial_data/module_type_profiles/expansion_card.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Expansion card",
+    "schema": {
+        "properties": {
+            "Connector type": {
```
jnovinger: Let's make this `connector_type`.

jnovinger (changes requested):
@harryajc , one more change then this is ready to go. Apologies for not catching that earlier.

harryajc:
@jnovinger Completed

jnovinger (approved):

PR #19691, opened 2025-06-10
Fixes #19687 - Treat cellular interface type as not connectable

### LINKED ISSUE #19687: Treat cellular interface type and not connectable
### Deployment Type

Self-hosted

### NetBox Version

v4.2.6

### Python Version

3.12

### Steps to Reproduce

1. Create interface type LTE
2. Via nbshell retrieve the interface
3. call the is_wired property

### Expected Behavior

is_wired should return False

### Observed Behavior

is_wired returns True

Omripresent:
### Fixes: #19687

Add cellular interface types to WIRELESS_IFACE_TYPES const
Add cable termination test for cellular interface

### DIFF
--- a/netbox/dcim/tests/test_models.py
+++ b/netbox/dcim/tests/test_models.py
@@ -954,6 +954,7 @@ class CableTestCase(TestCase):
         with self.assertRaises(ValidationError):
             cable.clean()
 
+    @tag('regression')
     def test_cable_cannot_terminate_to_a_cellular_interface(self):
         """
         A cable cannot terminate to a cellular interface
--- a/netbox/dcim/tests/test_models.py
+++ b/netbox/dcim/tests/test_models.py
@@ -954,6 +954,18 @@ class CableTestCase(TestCase):
         with self.assertRaises(ValidationError):
             cable.clean()
 
+    def test_cable_cannot_terminate_to_a_cellular_interface(self):
+        """
+        A cable cannot terminate to a cellular interface
+        """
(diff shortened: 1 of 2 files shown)


netbox/dcim/tests/test_models.py
```diff
@@ -954,6 +954,18 @@ def test_cable_cannot_terminate_to_a_wireless_interface(self):
         with self.assertRaises(ValidationError):
             cable.clean()
 
+    def test_cable_cannot_terminate_to_a_cellular_interface(self):
```
jnovinger, line 958: Thanks for adding the regression test!

jnovinger (dismissed):

netbox/dcim/tests/test_models.py
jnovinger, line 958: Derp, meant to also ask ... can you add a 'regression' tag here, like so?

```python
@tag('regression')
def test_cable_cannot_terminate_to_a_cellular_interface(self):
    ...

Comment (jnovinger) [reply netbox/dcim/tests/test_models.py:958]: You can import it like

```python
from django.test import tag
```

Omripresent: My bad I misunderstood the comment on the issue. I'll add and update the PR.

jnovinger: It couldn't have helped that it took me three comments to say what I meant to say. 🤦 

Thanks!

jnovinger (approved):

PR #19721, opened 2025-06-18
Fixes: #19492: Add Save Button to Script Output Window

### LINKED ISSUE #19492: Add Save Button to Script Output Window

Omripresent:
### Fixes: #19492: Add Save Button to Script Output Window

* Added download button to script_result.html template following the same model of object_render_config.html
* Added logic to ScriptResultView.get function to trigger content download when matching log or output.
   * Output export responds with the raw content of the script result
   * Log export responds with a CSV formatted content of the log table matching the visible table columns (works for both scripts and reports)

### DIFF
--- a/netbox/extras/views.py
+++ b/netbox/extras/views.py
@@ -1479,6 +1479,16 @@ class ScriptResultView(TableMixin, generic.ObjectView):
         if job.completed:
             table = self.get_table(job, request, bulk_actions=False)
 
+        if job.completed:
+            # If a direct export output has been requested, return the job data content as a
+            # downloadable file.
+            if request.GET.get('export') == 'output':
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @Omripresent!

PR #19725, opened 2025-06-19
Closes #19692: Unnecessary DB queries for API Endpoints

JCWasmx86:
### Fixes: #19692

Just adds a few prefetches for the API endpoints

- ipam/vlan-groups
- ipam/services
- virtualization/clusters

### DIFF
--- a/netbox/ipam/api/views.py
+++ b/netbox/ipam/api/views.py
@@ -143,7 +143,7 @@ class FHRPGroupAssignmentViewSet(NetBoxModelViewSet):
 
 
 class VLANGroupViewSet(NetBoxModelViewSet):
-    queryset = VLANGroup.objects.annotate_utilization()
+    queryset = VLANGroup.objects.annotate_utilization().prefetch_related('scope')
     serializer_class = serializers.VLANGroupSerializer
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):
Thanks for taking a stab at this @JCWasmx86. Unfortunately the root issue here (GFK relationships not being prefetched) requires a more robust solution.

First, a solution is needed for _every_ such instance within the application, not just the three relationships captured in #19692.

Second, we need a solution which ensures that the related objects are prefetched only where necessary: API serializers are dynamic, and users can request that arbitrary fields be omitted from the response. For example, a request for `/api/virtualization/clusters/?fields=id,name,status` will include only the ID, name, and status fields for each object: There's no need to prefetch the scope in this case. Generally, we want to avoid explicitly prefetching related objects on the view querysets and let the API handle it automatically.

JCWasmx86:
So you mean, e.g. somewhere around the NetBoxModelViewSet class?

jeremystretch:
It will probably require a new type of serializer field for GFKs, which can be inspected to automatically apply the prefetch, similar to how regular foreign keys are handled.

PR #19794, opened 2025-06-30
Closes: #19793 - Nav menu link customization

### LINKED ISSUE #19793: Support callables for nav menu
### NetBox version

v4.3.3

### Feature type

New functionality

### Proposed functionality

Support nav menu items which can be rendered dynamically via callables, rather than populated statically at startup time.

### Use case

Plugins such as Custom Objects may need to be able to extend the nav menu with arbitrary headings/subheadings during runtime without depending on startup configuration.

### Database changes

N/A

### External dependencies

N/A

bctiemann:
### Closes: #19793

Changes the `PluginMenuItem` and `PluginMenuButton` classes to handle an alternate pre-rendered literal `url` init parameter, rather than just supporting `link` which must be a view name suitable for `{% url %}`.

This allows plugin authors to supply link target URLs that might require additional (arbitrary) kwargs for resolution, rather than relying on built-in NetBox view names.

Note that Django does not provide a supported way of passing arbitrary kwargs to `{% url %}`; the params must be known in advance and specified explicitly (i.e. `{% url 'view_name' arg1='foo' arg2='bar' %}`). This prevents us from implementing this functionality in a more elegant way, i.e. being able to pass a `link_kwargs` dict which could then be plugged into the template directly like `{% url item.link **item.link_kwargs %}`.


jeremystretch:
I'm not convinced we need to (or should) support arbitrary executables in place of menu items. Where dynamic iteration is needed, a plugin author can pass an arbitrary object as `items` within a MenuGroup, and define its `__iter__()` method to return whatever MenuItems are prudent.

I tested this approach with the custom objects plugin and it works well. The only blocker I ran into there was the rigid requirement for passing a static view name as a MenuItem's `link`. Rather than introducing a separate, conflicting `url` parameter maybe it would make sense to introduce a `link_kwargs` argument that can be passed along with `link` for URL resolution in the menu item.

netbox/netbox/navigation/__init__.py
```diff
@@ -22,20 +24,46 @@ class MenuItemButton:
...
+            self._url = reverse(self.link)
+
+    @property
+    def url(self):
+        return self._url
+
+    @url.setter
+    def url(self, value):
+        self._url = value
+
```
jeremystretch: Want to move these to a base class to avoid the redundancy? It could be reused for both core and plugin nav objects.

jeremystretch (dismissed):

jeremystretch (approved):

PR #19795, opened 2025-06-30
Closes #19231: Add bulk renaming support for all models

### LINKED ISSUE #19231: Ability to bulk rename all relevant models
### NetBox version

v4.2.7

### Feature type

New functionality

### Proposed functionality

The table at `/ipam/services/` lists all created services attached to a Device or VM.

If you want to rename several of them, you have to edit each of them, one at a time.

It's already possible for interfaces (`/dcim/interfaces/`, button *Rename Selected*), so I propose the same feature for services too.

### Use case

1. You named a bunch of services "DNS", but now you need to distinguish between "DNS Resolver" and "DNS Authoritative".
2. You use [Ansible netbox.netbox.nb_inventory](https://docs.ansible.com/ansible/latest/collections/netbox/netbox/nb_inventory_inventory.html) with `group_by: [ "services" ]`, and therefore needs a coherent naming (I feel this use case could be improved by #17196)

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #19231

- Introduce the `getattr` template filter (needed for the bulk rename template)
- Add `BulkRename` to the default set of actions on `ObjectListView`
- Introduce a subclass of `BulkRenameView` for all relevant models; disable bulk renaming for others
- Officially support BulkRenameView for plugins
- Update ObjectAction URL resolution to fail silently

### DIFF
--- a/docs/plugins/development/views.md
+++ b/docs/plugins/development/views.md
@@ -64,6 +64,7 @@ Generic view classes (documented below) facilitate common operations, such as cr
 | `ObjectListView`     | View a list of objects                                 |
 | `BulkImportView`     | Import a set of new objects                            |
 | `BulkEditView`       | Edit multiple objects                                  |
+| `BulkRenameView`     | Rename multiple objects                                |
 | `BulkDeleteView`     | Delete multiple objects                                |
 
 !!! warning
@@ -168,6 +169,10 @@ Below are the class definitions for NetBox's multi-object views. These views han
         - save_object
 
 ::: netbox.views.generic.BulkEditView
+    options:
+      members: false
+
+::: netbox.views.generic.BulkRenameView
     options:
       members: false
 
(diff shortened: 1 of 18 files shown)


jnovinger (changes requested):
I'm seeing some issues with the bulk actions buttons. On this branch in my local, some actions on the page seem to make the bulk action buttons disappear:
- Using the quick filter
- Changing the page size
- Clicking another page of results

In all of these cases, a page refresh restores the bulk action buttons.

Also, it seems like the buttons are no longer hovering in the bottom left when vertical scrolling is required.

I can provide screenshots/video if needed.

jeremystretch:
> I'm seeing some issues with the bulk actions buttons.

This was an issue with the HTMX template. Should be fixed.

> Also, it seems like the buttons are no longer hovering in the bottom left when vertical scrolling is required.

The buttons start floating when one or more checkboxes are ticked, which I _think_ is the intended behavior.

jnovinger (approved):
Looks solid! I was not able to get the buttons to glitch at all.

PR #19803, opened 2025-07-02
Fixes #19800: ModuleType import supports associating ModuleTypeProfile 

### LINKED ISSUE #19800: Module type import does not assign profile
### Deployment Type

Self-hosted

### NetBox Version

v4.3.1

### Python Version

3.10

### Steps to Reproduce

When importing a model type that defines a profile, the profile is not assigned after the import succeeded.
This can be reproduced with the following test data:
```yaml
---
manufacturer: Generic
model: Fan1
part_number: JF1H
description: Front to Back Fan Tray
airflow: front-to-rear
profile: Fan
```

The `Fan` module type profile exists in a default installation.

1. On the Module Types page click "Import".
2. Stay on the "Direct import" tab, which opens by default. 
3. Paste the test data above into the "Data" field.
4. Click Submit

### Expected Behavior

The created module type should have the specified profile assigned. The import page lists `profile` as a field option. The import does check if the specified profile exists and complains when a non-existing profile was given in the YAML data.

### Observed Behavior

The module type is created from the import data, but the profile specified in the import is not assigned to the created module type.

Either this is expected behaviour because the import doesn't allow for specifying the profiles attribute data. Or, the profile shouldn't be a field option, since it does nothing.

jnovinger:
### Fixes: #19800
- Fixes issue by adding `profile` to the list of fields the import form should recognize
- Fixes up module type bulk import testing by adding necessary permissions for related models and adding inheritance of `BulkImportObjectsViewTestCase`
- Adds an additional regression assertion to make sure that this fix does not regress

### DIFF
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -470,8 +470,8 @@ class ModuleTypeImportForm(NetBoxModelImportForm):
     class Meta:
         model = ModuleType
         fields = [
-            'manufacturer', 'model', 'part_number', 'description', 'airflow', 'weight', 'weight_unit', 'comments',
-            'tags', 'profile',
+            'manufacturer', 'model', 'part_number', 'description', 'airflow', 'weight', 'weight_unit', 'profile',
+            'comments', 'tags'
         ]
 
 
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -471,7 +471,7 @@ class ModuleTypeImportForm(NetBoxModelImportForm):
         model = ModuleType
         fields = [
             'manufacturer', 'model', 'part_number', 'description', 'airflow', 'weight', 'weight_unit', 'comments',
-            'tags',
+            'tags', 'profile',
         ]
 
 
--- a/netbox/dcim/tests/test_views.py
+++ b/netbox/dcim/tests/test_views.py
@@ -3,17 +3,18 @@ from decimal import Decimal
 from zoneinfo import ZoneInfo
 
 import yaml
-from django.test import override_settings
+from django.test import override_settings, tag
 from django.urls import reverse
 from netaddr import EUI
 
+from core.models import ObjectType
 from dcim.choices import *
 from dcim.constants import *
 from dcim.models import *
 from ipam.models import ASN, RIR, VLAN, VRF
 from netbox.choices import CSVDelimiterChoices, ImportFormatChoices, WeightUnitChoices
 from tenancy.models import Tenant
-from users.models import User
+from users.models import ObjectPermission, User
 from utilities.testing import ViewTestCases, create_tags, create_test_device, post_data
 from wireless.models import WirelessLAN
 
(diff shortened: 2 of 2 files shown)


netbox/dcim/tests/test_views.py
```diff
@@ -1044,6 +1036,62 @@ def setUpTestData(cls):
...
+    def _set_module_type_bulk_import_permissions(self, **permission_definition):
+        obj_perm = ObjectPermission(**permission_definition)
+        obj_perm.save()
+        obj_perm.users.add(self.user)
+        additional_permissions = [
+            ConsolePortTemplate, ConsoleServerPortTemplate, PowerPortTemplate, PowerOutletTemplate,
+            InterfaceTemplate, FrontPortTemplate, RearPortTemplate, ModuleBayTemplate,
+        ]
+        for model in additional_permissions:
+            obj_perm.object_types.add(ObjectType.objects.get_for_model(model))
```
jnovinger: Open to suggestions on this approach.

jeremystretch: There's an [`add_permissions()` method](https://github.com/netbox-community/netbox/blob/e33793dc820c26c002fa018e557cceebeb98ed91/netbox/utilities/testing/base.py#L43) on the test case to facilitate this; will that work?

netbox/dcim/tests/test_views.py
```diff
@@ -1044,6 +1036,62 @@ def setUpTestData(cls):
...
+        )
+
+        # run base test
+        super().test_bulk_import_objects_with_permission()
+
+        # extra regression asserts
+        fan_module_type = ModuleType.objects.get(part_number='generic-fan')
+        fan_module_type_profile = ModuleTypeProfile.objects.get(name='Fan')
+
+        assert fan_module_type.profile == fan_module_type_profile
```
jeremystretch, line 1085: I don't think this is needed. The test should be checking that each declared field was set to the correct value on import.

netbox/dcim/forms/bulk_import.py
```diff
@@ -471,7 +471,7 @@ class Meta:
         model = ModuleType
         fields = [
             'manufacturer', 'model', 'part_number', 'description', 'airflow', 'weight', 'weight_unit', 'comments',
-            'tags',
+            'tags', 'profile',
```
jeremystretch: Could you move `profile` before `comments` in the list? It affects the ordering of the fields in the import form.

jeremystretch (changes requested):

netbox/dcim/tests/test_views.py
jnovinger: Copy, will verify. My memory was that it was just checking object count relative to pre-test.

jnovinger: It turns out that the parent test _is_ actually only testing that the count is correct.

It feels like a FR to expand the scope of the parent test, especially when it could very well break a bunch of other tests.

Open to feedback on that, though.

jeremystretch: Roger. I think I was confusing it for a different test.

jeremystretch (approved):

PR #19804, opened 2025-07-02
Closes #19589: Background job for bulk operations

### LINKED ISSUE #19589: Use a background job to handle bulk imports
### NetBox version

v4.3.1

### Feature type

### Proposed functionality

Rather than processing bulk import synchronously, NetBox should offload them to a background job and provide the user a link (or automatically redirect) to the job where the progress of the import can be checked.

### Use case

This will mitigate timeout errors during very large imports by removing the actual upload process from the request-response cycle.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #19589

- Introduce AsyncViewJob to handle the background execution of arbitrary views
- Introduce the `apply_request_processors()` utility context manager for applying request processors
- Extend BulkImportView to support execution via a background job
- Add the `background_job` checkbox to BulkImportForm
- The string representation of a job now shows its name (rather than its task ID)

### DIFF
--- a/netbox/core/models/jobs.py
+++ b/netbox/core/models/jobs.py
@@ -116,7 +116,7 @@ class Job(models.Model):
         verbose_name_plural = _('jobs')
 
     def __str__(self):
-        return str(self.job_id)
+        return self.name
 
     def get_absolute_url(self):
         # TODO: Employ dynamic registration
--- a/netbox/netbox/jobs.py
+++ b/netbox/netbox/jobs.py
@@ -8,11 +8,15 @@ from django_pglocks import advisory_lock
 from rq.timeouts import JobTimeoutException
 
 from core.choices import JobStatusChoices
+from core.events import JOB_COMPLETED, JOB_FAILED
 from core.models import Job, ObjectType
+from extras.models import Notification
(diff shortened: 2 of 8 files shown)


jnovinger (commented):
@jeremystretch , quick question, should a background import via pasted CSV (or via DataSource) also return a URL that points to the items modified by the request, e.g. `http://127.0.0.1:8000/ipam/services/?modified_by_request=<request_id>`? Or should that only be happening on the non-background job imports?

Apologies if I missed you describing that behavior before.

jeremystretch:
I'm not sure it makes sense for background jobs to redirect to the filtered view because it's typically going to show no results (because the job hasn't completed yet). On the other hand, it allows the user to easily refresh the page to check for results without having to click through the notification.

jnovinger:
> I'm not sure it makes sense for background jobs to redirect to the filtered view because it's typically going to show no results (because the job hasn't completed yet). On the other hand, it allows the user to easily refresh the page to check for results without having to click through the notification.

That works for me, I just wanted to make sure I had the right expectation.

jnovinger (approved):

PR #19820, opened 2025-07-07
19633 fix in condition aval

ravenrs:
### Fixes: #19633

Fixed the ValueError error that appeared in the status evaluation function when the value is None, it may be an error on the attribute path. I also added tests to cover this case. All of them were completed.


jeremystretch (changes requested):
Please see my guidance [here](https://github.com/netbox-community/netbox/issues/19633#issuecomment-3049273704) for the changes needed to address this bug.

PR #19856, opened 2025-07-10
Fix 19633 interrupting event processing in the process_event_rules() function

ravenrs:
### Fixes: #19633 

1. **condition.py**
- Added condition to to check when value is None. It should return False or not False if self.negative is True.
- Change condition with self.negate to result XOR self.negate for compactness.
2. **test_condition.py**
- Added test to cover cases when Value is None.
3.  **context_manager.py**
- Added context manager suppress_event_exp to prevent missing exceptions in cycles related to sequential event handling
4. **event.py**
- Using cm suppress_event_exp.
5. **test_event_rules.py**
- Added test to emulate the exception in process_event_function and verify that exception was logged properly.

### DIFF
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -81,7 +81,7 @@ class Condition:
             # Invalid key path
             value = None
         if value is None:
-            # Hande comparison case when value is None.
+            # Handle comparison case when value is None.
             if self.op in (self.GT, self.GTE, self.LT, self.LTE, self.IN, self.CONTAINS, self.REGEX):
                 result = False
         else:
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -80,16 +80,15 @@ class Condition:
         except TypeError:
             # Invalid key path
             value = None
+
         if value is None:
             # Handle comparison case when value is None.
             if self.op in (self.GT, self.GTE, self.LT, self.LTE, self.IN, self.CONTAINS, self.REGEX):
-                result = False
-        else:
-            result = self.eval_func(value)
+                return False ^ self.negate
 
-        if self.negate:
-            return not result
-        return result
+        result = self.eval_func(value)
+
+        return result ^ self.negate
 
     # Equivalency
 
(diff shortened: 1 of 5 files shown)


jeremystretch:
@ravenrs please update the body of your PR above to reference the correct issue number.

ravenrs:
@jeremystretch done.

ravenrs:
@jeremystretch What are the next steps? Do I need to fix something?

jeremystretch (changes requested):
@ravenrs per my comment [here](https://github.com/netbox-community/netbox/issues/19633#issuecomment-3049273704), the root issue is in the `flush_events()` function. There is no need to introduce a context manager to suppress exceptions.

I'm afraid I don't have time to go into more detail than that. If you would prefer someone else own this issue, I'll be happy to reassign it.

ravenrs:
@jeremystretch Thank you for your patience. I believe there might be some misunderstanding regarding the issue. When you have time, could you please review [my detailed comment in the issue thread](https://github.com/netbox-community/netbox/issues/19633#issuecomment-3073023376)?

If you're unavailable, could you recommend someone who could:

1. Reproduce this in a development environment
2. Verify the core problem:

- Event processing silently fails (no webhook failures shown)
- Subsequent events for the same object type get dropped

Quick Reproduction Steps (already set up on demo.netbox.dev):
I've pre-configured the test event - no setup needed
Simply update any device
Observe background tasks - you'll see no events on queue with the same object type (dcim.devices) event type (update object)

Key Points:
This isn't about context managers vs try/except (I'm happy to adjust implementation)
This is a critical reliability issue where:
Event processing halts for entire object types

The [original comment](https://github.com/netbox-community/netbox/issues/19633#issuecomment-3073023376) contains technical details, but the core problem is dangerously silent data loss.

Please let me know if you'd like me to:
Create a more minimal reproduction case
Provide additional debugging data

jeremystretch:
@ravenrs thanks for your work on this. Per my comment [here](https://github.com/netbox-community/netbox/issues/19633#issuecomment-3073406917) I'll take ownership of this issue. Hopefully the PR I submit will make clear my intended guidance.

PR #19873, opened 2025-07-14
Closes #18811: Match full-form IPv6 addresses in global search

### LINKED ISSUE #18811: Searching for IPv6 addresses in the global search should search for the compressed format
### NetBox version

v4.2.4

### Feature type

### Proposed functionality

Currently, when searching for ipv6 addresses in the global search, the user may only use the compressed ipv6 form (no leading zeros, multiple `:0:` blocks replaced by `::`). My suggestion is that if the search term correctly parses as an ipv6 address, it should be internally normalized to the compressed form before sending it off to the search engine. Btw a similar correction already is in place with capital letters in ipv6 addresses, which are transparently converted to lower case letters.

### Use case

When copying+pasting addresses or constructing search urls from other systems these systems may not always work with the compressed form, which causes the user to have to manually change the search term or being in the wrong impression that this ip address is not present in netbox, causing confusion or errors.

### Database changes

None

### External dependencies

None

jeremystretch:
### Closes: #18811

- Introduce an equivalent NetHost lookup specifically for cached IP address values
- Extend the search backend to employ this filter under certain conditions
- Add a normalized `ipv6_full` property on the Aggregate, Prefix, and IPAddress models (for general use)

### DIFF
--- a/netbox/netbox/search/backends.py
+++ b/netbox/netbox/search/backends.py
@@ -115,11 +115,13 @@ class CachedValueSearchBackend(SearchBackend):
         if lookup in (LookupTypes.STARTSWITH, LookupTypes.ENDSWITH):
             # "Starts/ends with" matches are valid only on string values
             query_filter &= Q(type=FieldTypes.STRING)
-        elif lookup == LookupTypes.PARTIAL:
+        elif lookup in (LookupTypes.PARTIAL, LookupTypes.EXACT):
             try:
-                # If the value looks like an IP address, add an extra match for CIDR values
+                # If the value looks like an IP address, add an extra filters for CIDR/INET values
                 address = str(netaddr.IPNetwork(value.strip()).cidr)
-                query_filter |= Q(type=FieldTypes.CIDR) & Q(value__net_contains_or_equals=address)
+                query_filter |= Q(type=FieldTypes.INET) & Q(value__net_host=address)
+                if lookup == LookupTypes.PARTIAL:
+                    query_filter |= Q(type=FieldTypes.CIDR) & Q(value__net_contains_or_equals=address)
             except (AddrFormatError, ValueError):
                 pass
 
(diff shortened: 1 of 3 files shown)


netbox/netbox/search/backends.py
```diff
@@ -115,11 +115,13 @@ def search(self, value, user=None, object_types=None, lookup=DEFAULT_LOOKUP_TYPE
         if lookup in (LookupTypes.STARTSWITH, LookupTypes.ENDSWITH):
             # "Starts/ends with" matches are valid only on string values
             query_filter &= Q(type=FieldTypes.STRING)
-        elif lookup == LookupTypes.PARTIAL:
+        elif lookup in (LookupTypes.PARTIAL, LookupTypes.EXACT):
             try:
-                # If the value looks like an IP address, add an extra match for CIDR values
+                # If the value looks like an IP address, add an extra filters for CIDR/INET values
```
jnovinger: tiny quibble: 'add extra filters' rather than 'add an extra filters'

jnovinger (commented):
I don't think this is in scope, but when I do search for the IPv6 address in expanded form, it does not highlight the matched object, like it does when I search for and find it using the compressed format.

Addresses used for testing:
- Expanded: 2001:0db8:0000:0000:0000:8a2e:0370:7334/64
- Abbreviated: 2001:db8::8a2e:370:7334/64

Other than that (and one tiny grammar quibble in a comment), this all looks great.

jnovinger (dismissed):
Derp, meant to approve this.

jeremystretch:
> it does not highlight the matched object

Yeah, unfortunately I don't think this is feasible. The highlighting works on simple text matching; we would need to somehow expose the IPv6 addressing logic to support it. IMO it's fine to leave this as-is.

jnovinger (approved):

PR #19881, opened 2025-07-14
Fixes #19413: Group custom fields in filter tab

### LINKED ISSUE #19413: Filter tab: Custom fields are not grouped
### Deployment Type

Self-hosted

### NetBox Version

v4.3.0

### Python Version

3.12

### Steps to Reproduce

1. Create several custom fields, grouped by several groups, assigned to *Virtual Machines*
2. Go to https://netbox.siaga.re/virtualization/virtual-machines/
3. Open *Filters* tab
4. Scroll down to *Custom Fields*

### Expected Behavior

CF should be grouped just like they are when you see or edit a VM (here, groups GLPI and VM):

Image

### Observed Behavior

They are not grouped, so you cannot tell exactly what are these *ID* or *Folder*:

Image

jnovinger:
### Fixes: #19413

Replaced manual rendering of custom fields in the filter tab with the `render_custom_fields` template tag. This change ensures that custom fields are properly grouped, addressing the issue where they were previously displayed without their associated groups.

### DIFF
--- a/netbox/templates/inc/filter_list.html
+++ b/netbox/templates/inc/filter_list.html
@@ -29,11 +29,7 @@
           <div class="hr-text">
             <span>{% trans "Custom Fields" %}</span>
           </div>
-          {% for name in filter_form.custom_fields %}
-            {% with field=filter_form|get_item:name %}
-              {% render_field field %}
-            {% endwith %}
-          {% endfor %}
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Nice

PR #19895, opened 2025-07-16
Closes #19893: Include hostname in REST API status endpoint

### LINKED ISSUE #19893: Report the system hostname in `/api/status/`
### NetBox version

v4.3.4

### Feature type

New functionality

### Proposed functionality

Include the system hostname in the data returned by the `/api/status/` REST API endpoint.

### Use case

The system hostname may be a valuable piece of information to evaluate along with other status information, especially now that #18528 (designated for the v4.4 release) allows it to be configured.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #19893

Include the system hostname in REST API status output.

### DIFF
--- a/netbox/netbox/api/views.py
+++ b/netbox/netbox/api/views.py
@@ -65,6 +65,7 @@ class StatusView(APIView):
 
         return Response({
             'django-version': DJANGO_VERSION,
+            'hostname': settings.HOSTNAME,
             'installed-apps': installed_apps,
             'netbox-version': settings.RELEASE.version,
             'netbox-full-version': settings.RELEASE.full_version,


jnovinger (approved):
Two quick thoughts for follow-up:
- Since we don't actually document the /api/status/ endpoint anywhere (which hasn't been a problem), maybe we should file a housekeeping ticket to add it to the API schema? Would help all the existing fields too.
- Also worth considering a simple test for this - even though it's straightforward, having automated coverage for API fields helps catch regressions without relying on manual checks during releases.

Both are follow-up items though - this is ready to ship as-is!

Issue #19806, opened 2025-07-02
No entry point exists to mark a background job as failed


jeremystretch:
### Deployment Type

Self-hosted

### NetBox Version

v4.3.3

### Python Version

3.10

### Steps to Reproduce

This issue concerns the current implementation of JobRunner, specifically the way the `run()` method of a job is executed:

https://github.com/netbox-community/netbox/blob/ea4c205a37baa3e58e6e481158c15c6154cceeff/netbox/netbox/jobs.py#L77-L79

### Expected Behavior

Setting the status of a job to "failed" (`STATUS_FAILED`) within the `run()` method should mark the completed job as failed.

### Observed Behavior

The job's status is overwritten with the default "completed" status so long as no exceptions are raised within `run()`. This is because the `terminate()` method, called immediately after `run()`, sets the job's status to "completed" by default.

Issue #19735, opened 2025-06-20
Implement the ability to register arbitrary bulk actions for a model


jeremystretch:
### NetBox version

v4.3.2

### Feature type

New functionality

### Proposed functionality

Provide a mechanism for canonically declaring bulk action support for individual models. Currently, NetBox supports bulk edit and bulk delete functionality by default. Certain models, such as device components, support additional actions such as bulk renaming and bulk disconnecting. However, this is accomplished using customized templates in a nonstandard way.

We can implement a form of lightweight API to support arbitrary bulk operations for any model, obviating the need for customization in views and templates. This could be as simple as inheriting a base class:

```python
class BulkEdit(BulkOperation):
    verb = _('Edit')
    button_class = 'warning'
    view_name = 'bulk_edit'
```

This class could be responsible for evaluating user permissions (based on the declared set of actions for a model), resolving the view, and rendering the UI button. Bulk operations could be declared under each generic view, with the ability to be overridden where needed:

```
class SiteListView(generic.ObjectListView):
    bulk_actions = (BulkEdit, BulkRename, BulkDelete)
```

It might also be feasible to extend view registration (`register_model_view()`) to automatically enable bulk operations.

### Use case

This will unlock the implementation of bulk operations beyond the defaults (bulk edit & bulk delete) for more models.

### Database changes

None

### External dependencies

None

PR #19897, opened 2025-07-16
Closes #19891: Bulk operation jobs

jeremystretch:
### Closes: #19891

- Adopt fix for bug #19806 in v4.3.4
- Refactor base bulk edit forms
- Add `background_job` toggle to BulkEditForm and BulkDeleteForm via BackgroundJobMixin
- Extend BulkEditView & BulkDeleteView to support background jobs
- Move the attachment of a success message to `process_request_as_job()`

### DIFF
--- a/netbox/netbox/forms/base.py
+++ b/netbox/netbox/forms/base.py
@@ -8,7 +8,7 @@ from django.utils.translation import gettext_lazy as _
 from core.models import ObjectType
 from extras.choices import *
 from extras.models import CustomField, Tag
-from utilities.forms import CSVModelForm
+from utilities.forms import BulkEditForm, CSVModelForm
 from utilities.forms.fields import CSVModelMultipleChoiceField, DynamicModelMultipleChoiceField
 from utilities.forms.mixins import CheckLastUpdatedMixin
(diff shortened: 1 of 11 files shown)


jnovinger (commented):
Sorry, this has a fair amount of surface area, so wanted to give it some attention while poking at it. Everything seems to work very well. I only noticed one potentially unexpected behavior.

When invoking bulk edit or bulk delete for for objects in a sub-table (think: viewing interfaces of a device, via the device's detail view), you're dumped back in the generic list view for the item type being edited/deleted.

So, I have a switch. I select multiple interfaces from `/dcim/devices/<device pk>/interfaces/` and click bulk edit/delete. I then select background job and submit. I get redirected to `/dcim/interfaces/`.

jeremystretch:
> When invoking bulk edit or bulk delete for for objects in a sub-table (think: viewing interfaces of a device, via the device's detail view), you're dumped back in the generic list view for the item type being edited/deleted.

I think this is actually a regression introduced under #19735; it occurs in the current `feature` branch as well. I considered tackling it as part of this PR, but it probably requires changes to form controls outside the scope of the associated FR. I've opened bug #19911 to address this separately.

Issue #18588, opened 2025-02-06
Tweak label for services to better represent intended use


jeremystretch:
### NetBox version

v4.2.3

### Feature type

Other

### Proposed functionality

Change the canonical name for the Service model to something like "L7 Service" or similar.

(Note: This does _not_ propose actually renaming the model itself of any API endpoints.)

### Use case

This would better convey the model's intended use and hopefully avoid confusion with the myriad other definitions for the term "service."

### Database changes

N/A

PR #19900, opened 2025-07-16
Closes: #18588: Relabel Service to Application Service

jnovinger:
### Closes: #18588
This PR relabels the Service model to "Application Service" (and ServiceTemplate to "Application Service Template") throughout the UI to make it clearer what this model is actually for.

"Application Service" was chosen because it best describes the model's intent: to represent pretty much anything running above the transport layer (L4). This includes classic examples like SSH, HTTPS, and NTP, but also custom applications, APIs, or any other process defined by a protocol and port. The name "L7 Service" was considered, but it felt a little too restrictive and could be confusing for users who also interact with programmatic aspects of NetBox.

To ensure backward compatibility, this is a UI-only change. The underlying models, REST API endpoints, and GraphQL schema remain unchanged.
TODO:

- [x] Update Service and ServiceTemplate verbose_name and adapt test suites (django-filter, GraphQL) to assert API stability.
- [x] Rename Service and Service template in the IPAM sub-menu.
- [x] Update hardcoded labels for buttons, titles, and messages across all relevant templates, views, forms, and tables.
- [x] Update docs as necessary.
- ~~Optional: work to make sure translations exist~~ This should be covered by routine Transifex maintenance.

### DIFF
--- a/docs/models/virtualization/vminterface.md
+++ b/docs/models/virtualization/vminterface.md
@@ -1,6 +1,6 @@
 ## Interfaces
 
-[Virtual machine](./virtualmachine.md) interfaces behave similarly to device [interfaces](../dcim/interface.md): They can be assigned to VRFs, may have IP addresses, VLANs, and services attached to them, and so on. However, given their virtual nature, they lack properties pertaining to physical attributes. For example, VM interfaces do not have a physical type and cannot have cables attached to them.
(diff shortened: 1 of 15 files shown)


docs/models/virtualization/vminterface.md
```diff
@@ -1,6 +1,6 @@
 ## Interfaces
 
-[Virtual machine](./virtualmachine.md) interfaces behave similarly to device [interfaces](../dcim/interface.md): They can be assigned to VRFs, may have IP addresses, VLANs, and services attached to them, and so on. However, given their virtual nature, they lack properties pertaining to physical attributes. For example, VM interfaces do not have a physical type and cannot have cables attached to them.
```
jnovinger, line 3: I removed this bit here because I could not find any direct relationship between application services and VM interfaces. If this is some other concept of service, let me know and I can add it back.

jeremystretch (approved):
Nice work accommodating for the field M2M filter name deviation in the test suite.

Issue #18204, opened 2024-12-11
Image Attachments section redesign


artiomello:
### NetBox version

v4.1.6

### Feature type

Change to existing functionality

### Triage priority

N/A

### Proposed functionality

I believe the `Image Attachment` section needs to be revised to have it reflect data a bit more concisely, and precisely at that.

1. Image size
<img width="1061" alt="image" src="https://github.com/user-attachments/assets/c8708aba-5a0e-4e12-bef0-db7b5190355d">

Size needs to be shown in MB, not in bytes. I'm almost 40 now and even I can't recall a single time in my life where image size in bytes would make any sense to me. Let alone now.

---
2. Name and Image columns
<img width="412" alt="image" src="https://github.com/user-attachments/assets/60f768f9-1fe2-466b-97cf-a269ec8e3949">

I think only one of them actually needs to stay. If you give a name to an image during upload - it's renamed accordingly, and the new name displayed in both columns. If you don't give a name, then the file name is taken and is displayed only in the `Image` column. 

You can't have a situation where these two properties would be different, rather (and weirdly enough) omitted in some cases. So, they're pretty much the same. Hence having both of them doesn't make much sense. 

---
3. Referencing image attachments in other objects and views

So, if we open any object that has image attachments like this one, that has two:
<img width="1053" alt="image" src="https://github.com/user-attachments/assets/0dd5088a-a49c-4fa6-b858-b0f2a1ed5d25">

We see that it likely filters that same original image table with itself, and we get the same columns as in the original table:
<img width="583" alt="image" src="https://github.com/user-attachments/assets/b5b68cba-9f48-4aa8-8b0f-c7a0c1e1523b">
Only here this information is completely redundant:
- We already know its name and that it's a site. Showing these properties here would make sense only in case it would show images for the site's child objects (which it should definitely not).
- The image/name pair is redundant, as described above.
- The size in bytes looks more like a phone number than anything else.
- All in all, this redundant information just adds more clutter to the object's page.

What really should be displayed here are:
- Name (serving as a pop-up image window).
- Size (in MB, obviously).
- Creation timestamp.

### Use case

Updating this table, its object properties and how they are presented in other views/objects would make certain netbox pages less cluttery and more concise, showing only that which has value. This is especially true when viewing some more data rich objects, such as sites and racks. 

 Less is more, as they say.

jeremystretch:
Blocked by #18990

PR #19907, opened 2025-07-17
Closes #18990: Add description field to ImageAttachment model

### LINKED ISSUE #18990: Comment/description for image attachments
### NetBox version

v4.2.5

### Feature type

Data model extension

### Proposed functionality

Other objects support a free-floating comment/description field, however image attachments do not.

We found this upon importing images from RackTables, which does have a comment field for attachments.

### Use case

This would be useful to:

- have a text reference about what the image is about before opening it
- have a text caption about the what is displayed on the image for visually impaired users

Using the `Name` field for this purpose not ideal, as that one alters the display of the image file name as well.

### Database changes

It requires an additional field.

### External dependencies

_No response_

jeremystretch:
### Closes: #18990

- Add an optional `description` field to the ImageAttachment model
- Enable global search for image attachments by name & description
- Establish a dedicated UI view for an image attachment to display all attributes


arthanson (changes requested):
This all works and potentially an existing issue but the detail view shown after you add an image is great, but is only accessible when you add an image.  There is no way to get back to it from either the sub-table (like on a device) or from the main Image Attachment list view, maybe have the name field link to the detail view?

jeremystretch:
Yeah, I went a little outside the scope of the FR honestly, and didn't want to push too far. I'm planning on doing some additional cleanup under #18204, which I expect to include linking to the detail view for individual image attachments.

arthanson (approved):

PR #19908, opened 2025-07-17
Closes #19713: Enable recording user messages in the change log

### LINKED ISSUE #19713: Add ability to add a comment to change log entries
### NetBox version

v4.3.2

### Feature type

Data model extension

### Proposed functionality

Support an additional "netbox-change-comment: some comment goes here" header for changes made via the api that gets stored in the change log together with the changes. 

### Use case

We have a lot of automated systems that make changes to our NetBox instance. It would be great if we could send additional information about what caused a certain change to NetBox that would be included in the change log entry.

### Database changes

Additional text field on the changelog model.

### External dependencies

None.

jeremystretch:
### Closes: #19713

- Add a `message` field on the ObjectChange model
- Add a `changelog_message` field to the object edit & delete forms
- Capture the `changelog_message` value on the resulting ObjectChange record when an object is saved or deleted via the web UI
- Expand the search filter for ObjectChange to match against the `message` value
- Expand all model forms & serializers to support changelog messages as appropriate
- Add the `changelog_message` field to the necessary view templates
- Define `fieldsets` on VirtualChassisCreateForm & discontinue use of the custom HTML template (to ensure consistent display of the changelog message field form)
- Extend object & bulk views to support changelog messages
- Extend the view and REST API test suite to validate the changelog message functionality for all create/edit/delete operations
- Document the new changelog message functionality


netbox/ipam/tests/test_api.py
```diff
@@ -1152,7 +1156,8 @@ def setUpTestData(cls):
         ]
 
         cls.bulk_update_data = {
-            'policy': vlan_translation_policies[1].pk,
+            'policy': vlan_translation_policies[2].pk,
```
jeremystretch, line 1159: Before this change, the test was not producing the expected number of ObjectChange records, which failed a check in the updated test suite.

jeremystretch:
The test failure appears to be unrelated to these changes. Extension of the test functionality has uncovered an unrelated bug; see #19956.

jeremystretch:
The CI failure should be addressed by the fix in PR #19957.

netbox/core/filtersets.py
```diff
@@ -186,7 +186,8 @@ def search(self, queryset, name, value):
             return queryset
         return queryset.filter(
             Q(user_name__icontains=value) |
-            Q(object_repr__icontains=value)
+            Q(object_repr__icontains=value) |
+            Q(message__icontains=value)
```
jnovinger, line 190: If we're going to make this a default queried field, should we add an index to the field? (He says not having looked at other similar fields yet ...)

netbox/utilities/testing/views.py
```diff
@@ -804,13 +858,17 @@ def test_bulk_delete_objects_without_permission(self):
                 self.assertHttpStatus(self.client.post(self._get_url('bulk_delete'), data), 403)
 
         def test_bulk_delete_objects_with_permission(self):
-            pk_list = self._get_queryset().values_list('pk', flat=True)
+            pk_list = list(self._get_queryset().values_list('pk', flat=True))[:3]
```
jeremystretch, line 861: Limit this test to three objects (as above) to avoid failing on cascading deletions for some test data.

docs/integrations/rest-api.md
```diff
@@ -608,6 +608,28 @@ http://netbox/api/dcim/sites/ \
...
+
+Most objects in NetBox support [change logging](../features/change-logging.md), which generates a detailed record each time an object is created, modified, or deleted. Beginning in NetBox v4.4, users can attach a message to the change record as well. This is accomplished via the REST API by including a `changelog_message` field in the object representation.
+
+For example, the following API request will create a new site and record a message in the resulting changelog entry:
+
+```no-highlight
+curl -s -X POST \
+-H "Authorization: Token $TOKEN" \
+-H "Content-Type: application/json" \
+http://netbox/api/dcim/site/ \
```
jnovinger:
```suggestion
http://netbox/api/dcim/sites/ \
```

jnovinger (commented):
One change requested and one question.

netbox/core/filtersets.py
jeremystretch: Probably not as part of this change. `message` here is analogous to `description` on most other models. I'm not sure if it makes sense to add an index for those fields, but it does it IMO it should be undertaken as part of an application-wide audit. For now I think it makes sense to continue the current convention.

jnovinger: Yep, got ahead of myself here. Just got done looking around and realized the same thing about it being a bigger, separate chunk of work.

PR #19913, opened 2025-07-18
Closes #19902: add clip path to avoid overflow of device name

MarcoSpiz:
Add text truncation to avoid device name overflow in SVG rendering, and add clipPath method as fallback because text_truncate is an approximation


arthanson (approved):
Thanks @MarcoSpiz !

PR #19914, opened 2025-07-18
Closes #18204: Miscellaneous improvements to the display of image attachments

### LINKED ISSUE #18204: Image Attachments section redesign

jeremystretch:
### Closes: #18204

- Enable filtering & global search by image attachment filename
- Simplify default table configuration for image attachment list
- Primary link now goes to the detail view for individual image attachments
- Introduce optional columns to include additional data

### DIFF
--- a/netbox/extras/tables/tables.py
+++ b/netbox/extras/tables/tables.py
@@ -37,14 +37,6 @@ __all__ = (
     'WebhookTable',
 )
 
-IMAGEATTACHMENT_IMAGE = """
-{% if record.image %}
-  <a class="image-preview" href="{{ record.image.url }}" target="_blank">{{ record }}</a>
-{% else %}
-  &mdash;
-{% endif %}
-"""
-
 NOTIFICATION_ICON = """
 <span class="text-{{ value.color }} fs-3"><i class="{{ value.icon }}"></i></span>
 """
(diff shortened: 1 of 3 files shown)


netbox/extras/tables/tables.py
```diff
@@ -230,16 +232,31 @@ class ImageAttachmentTable(NetBoxTable):
...
         verbose_name=_('Parent'),
-        linkify=True
-    )
-    image = tables.TemplateColumn(
-        verbose_name=_('Image'),
-        template_code=IMAGEATTACHMENT_IMAGE,
+        linkify=True,
+        orderable=False,
     )
     size = tables.Column(
```
jnovinger, line 261: One tiny nit on this one, size's verbose name still says 'Bytes' even though this is now rendered by `filesizeformat()` and includes a size unit.

jnovinger (changes requested):
One little detail on this one, everything else looks good.

Issue #17972, opened 2024-11-08
Media files are reachable unauthenticated


Andres1357:
### Deployment Type

Self-hosted

### Triage priority

N/A

### NetBox Version

v4.1.6

### Python Version

3.12

### Steps to Reproduce

1. Install NetBox v4.1.0 - 4.1.6 and ensure configuration variable LOGIN_REQUIRED = True.
2. Add an image attachment to any object and get the URL to the image
(https://yournetbox.com/media/image-attachments/yourimage.png)
3. Attempt to access that URL while logged out of NetBox.

I've reproduced this in various NetBox versions (4.1.0, 4.1.1, 4.1.5, 4.1.6) both as standalone and in Docker.

It is reproducible using the Documents plugin as well as it also stores files in the media directory (https://yournetbox.com/media/netbox-documents/doc.pdf)

I've also reproduced it on https://netbox-demo.netboxlabs.com, though I can't confirm if that instance has LOGIN_REQUIRED = True.

### Expected Behavior

It is my understanding that NetBox should not display the file and instead redirect to the login page if the variable LOGIN_REQUIRED = True. This was the behavior seen on NetBox v4.0.3.

### Observed Behavior

NetBox displays the file just as if you were logged in.
![image](https://github.com/user-attachments/assets/4c4633bf-b37d-474e-a49c-19b62da79ad8)

Andres1357:
I believe change #16580 may be the cause.

Azmodeszer:
Can confirm this.

Andres1357:
Thanks Jeremy!

PR #19919, opened 2025-07-21
Closes #19591: Establish dedicated tab for image attachments

### LINKED ISSUE #19591: Dedicated tab on object detail view for image attachments
### NetBox version

V4.1.10

### Feature type

Other

### Proposed functionality

Adding a Tab to DCIM objects where you can view thumbnails of images attached to said objects.

I dont think there would be a huge amount of development involved in this, and I had thought about trying to develop a plugin with a template extension to do it, but feel like it might be something that others using netbox might appreciate.

### Use case

We are using pictures to "site audits" currently, then using those pictures to reference in the future instead of having to send a technician out to the site to see exactly what is going on.  
We have also proposed having technician take pictures of individual devices after they make changes to the physical interfaces, and try and use computer vision to determine which interfaces have cables and which don't, but until we can get a AI computer vision solution working, it would be nice if we could view these images in a thumbnail.  

### Database changes

no db changes needed

### External dependencies

_No response_

jeremystretch:
### Closes: #19591

- Introduce [sorl-thumbnail](https://github.com/jazzband/sorl-thumbnail) as a new dependency
- Register ObjectImageAttachmentsView for objects which support image attachments
- Render thumbnails for each attached image under the new "Images" tab
- Add an `html_tag` property to ImageAttachment for convenience


jnovinger:
`sorl-thumbnail` is a package name I haven't heard for a fairly long time. Didn't realize it was still being maintained.

netbox/templates/extras/imageattachment.html
```diff
@@ -56,8 +56,13 @@ <h2 class="card-header">{% trans "File" %}</h2>
...
         <h2 class="card-header">{% trans "Image" %}</h2>
         <div class="card-body">
-          <a href="{{ object.image.url }}">
-            <img src="{{ object.image.url }}" height="{{ image.height }}" width="{{ image.width }}" alt="{{ object }}" />
+          <a href="{{ object.image.url }}" title="{{ object.description }}">
+            <img
+              src="{{ object.image.url }}"
+              height="{{ image.height }}"
+              width="{{ image.width }}"
+              alt="{{ object.name }}"
```
jnovinger: Noticed a couple of things here:

1. `image.height` and `image.width` don't seem to actually produce anything, but `object.image_height` and `object.image_width` do, at least for my example image. Perhaps I missed something.

2. I would argue that we should switch what attributes we're using for the `<a title=...>` and `<img alt=...`.

    My reasoning is that image alt attributes should describe the image, not name it. Conversely, a link title should represent what happens if you click the link, and the image title better represents that.

    Having said that, we do already display the description in page and I'm seeing multiple places say that you shouldn't repeat nearby text in alt attributes. So ... yeah.

3. Given nothing more than a gut feeling that images are more likely to have names than descriptions (or both names and descriptions), perhaps the alt attribute (the more important from an accessibility perspective) should do some fallback, like `alt="{{ object.description|default:object.name }}"`.

netbox/templates/extras/object_imageattachments.html
```diff
@@ -0,0 +1,35 @@
...
+    {% for object in image_attachments %}
+      <div class="thumbnail m-2">
+        {% thumbnail object.image "200x200" crop="center" as tn %}
+          <a href="{{ object.get_absolute_url }}" class="d-block" title="{{ object.description }}">
+            <img
+              src="{{ tn.url }}"
+              width="{{ tn.width }}"
+              height="{{ tn.height }}"
+              class="rounded"
+              alt="{{ object.name }}"
```
jnovinger: Same as above about title/alt attributes. Although, here, I would argue more forcefully for description as the alt attribute since the description text is not otherwise displayed on the page.

netbox/templates/extras/object_imageattachments.html
```diff
@@ -0,0 +1,35 @@
...
+{% endblock %}
+
+{% block content %}
+  <div class="d-flex flex-wrap">
+    {% for object in image_attachments %}
+      <div class="thumbnail m-2">
+        {% thumbnail object.image "200x200" crop="center" as tn %}
+          <a href="{{ object.get_absolute_url }}" class="d-block" title="{{ object.description }}">
+            <img
+              src="{{ tn.url }}"
```
jnovinger: We recently [enforced authentication for viewing images](https://github.com/netbox-community/netbox/issues/17972).

However, even though these are fairly small images without a ton of detail, using this thumbnail URL as is makes it possible to see a representation of the image outside of being logged in.

.gitignore
```diff
@@ -9,6 +9,7 @@ yarn-error.log*
 /netbox/netbox/configuration.py
 /netbox/netbox/ldap_config.py
 /netbox/local/*
+/netbox/media/cache
```
jnovinger: I was surprised that `/netbox/media` wasn't already in this list. Then I noticed that sub-folders of `/netbox/media` have their own `.gitignore` files.

Any particular reason we didn't use that approach here?

netbox/templates/extras/object_imageattachments.html
```diff
@@ -0,0 +1,35 @@
...
+              class="rounded"
+              alt="{{ object.name }}"
+            />
+          </a>
+        {% endthumbnail %}
+        <div class="text-center text-secondary text-truncate fs-5">
+          {{ object }}
+        </div>
+      </div>
+    {% endfor %}
```
jnovinger: Perhaps we should have an `{% empty %}` clause here noting that there are no image attachments.

I was a little surprised when I uploaded a front image for a device type (derp) and then viewed this tab and saw nothing. Admittedly, front/back images for device types are not the intended use for this tab.

jnovinger (commented):
Raised some concerns/questions, although I could probably be convinced otherwise on most of them.

.gitignore
jeremystretch: IMO it's cleaner and less confusing to keep everything in the root `.gitignore`. I don't think there's any particular reason the other media paths have their own files.

netbox/templates/extras/object_imageattachments.html
jeremystretch: AFAICT the thumbnails (with URLs like `/media/cache/a3/92/a392eccb4e72730f96e4eddda867eb05.jpg`) are also protected, as authentication is enforced for the entire `/media/` path. Were you able to access them without being authenticated?

.gitignore
jeremystretch: I've collapsed them all into the root `.gitignore` per our conversation.

netbox/templates/extras/object_imageattachments.html
jnovinger: I'm pretty sure I was. But I will double check.

jnovinger: Ah, I was incorrect. I had been using different configuration files and accidentally used one with `LOGIN_REQUIRED=False`. My bad.

netbox/templates/extras/object_imageattachments.html
```diff
@@ -0,0 +1,46 @@
...
+      {% for object in image_attachments %}
+        <div class="thumbnail m-2">
+          {% thumbnail object.image "200x200" crop="center" as tn %}
+            <a href="{{ object.get_absolute_url }}" class="d-block" title="{{ object.name }}">
+              <img
+                src="{{ tn.url }}"
+                width="{{ tn.width }}"
+                height="{{ tn.height }}"
+                class="rounded"
+                alt="{{ object.description or object.name }}"
```
jnovinger: `{{ object.description or object.name }}` seems to be invalid--throws a `TemplateSyntaxError` exception for me.

```suggestion
                alt="{{ object.description|default:object.name }}"
```

jnovinger (changes requested):

jnovinger (approved):

.gitignore
tobiasge: This caused the deletion of those directories the repository and with that failures in the Docker image build.
I can easily fix it in the Dockerfile. Just wanted to be sure if the deletion of the directories was intentional.

jeremystretch: AFAIK there's no need to preserve the directories in the repo: Django will create them automatically as needed. Does that impose a challenge for the Docker image? Apologies for the surprise change.

tobiasge: It's an easy fix for the image. I just create the media directory in the build process.

PR #19942, opened 2025-07-23
Closes: #18936 import cable color names

### LINKED ISSUE #18936: Import cables with color name instead of hex code
### NetBox version

v.4.2.5

### Feature type

### Proposed functionality

Make it possible to import cables with the color name instead of the hex code. 

Currently, color names get rejected and a hex code is demanded.

Image

### Use case

Would allow our Data Center Staff to enter more human-friendly color names instead of having to deal with hex codes.

### Database changes

Possibly, yes - from what I can see of the API data, only the hex code is stored, not the color name

### External dependencies

None

Jathn:
### Fixes: #18936 

The changes made allows for colors to be imported as either hex code or with the label name.
Changes made:

- Make "color" a custom field instead of using the built in Django field. It now allows longer names for the colors (previously only 6 chars matching that of a hex code.)
- The given color is then matched against both the labels and hex code of ColorChoices. This is more in line with how the color is selected through the GUI as well, as previously any hex code could be imported.
-
Note: It seems that color choices in general are limited to the set defined in ColorChoices, however I can't seem to find these documented anywhere. Didn't include it as it is out of scope for this issue.


Jathn:
Hello,
I'll check the failing tests and get back to you.

PR #19949, opened 2025-07-24
Closes #18936: add color name support for cable bulk import

### LINKED ISSUE #18936: Import cables with color name instead of hex code

Jathn:
### Closes: #18936

The changes made allows for colors to be imported as either hex code or with the label name.

Changes made:
- Make "color" a custom field instead of using the built in Django field. It now allows longer names for the colors (previously only 6 chars matching that of a hex code.)
- The given color is matched against the labels of ColorChoices. In case of a hex code, the behavior is the same as previous implementation and all test cases still pass.
- The exception raised if the color is over six characters and not in ColorChoices is now separate from that of the base case.
 
*Note: It seems that color choices in general are limited to the set defined in ColorChoices, however I can't seem to find these documented anywhere. Didn't include it as it is out of scope for this issue.*

### DIFF
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -1410,6 +1410,12 @@ class CableImportForm(NetBoxModelImportForm):
         required=False,
         help_text=_('Length unit')
     )
+    color = forms.CharField(
+        label=_('Color'),
+        required=False,
+        max_length=16,
+        help_text=_('Color name (e.g. "Red") or hex code (e.g. "f44336")')
+    )
 
     class Meta:
         model = Cable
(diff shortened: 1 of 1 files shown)


arthanson (approved):
Thanks @Jathn !

Issue #18396, opened 2025-01-15
Get deviceTypes front_image and rear_image from the API


Janl1:
### NetBox version

v4.2.1

### Feature type

New functionality

### Triage priority

N/A

### Proposed functionality

Currently, I cannot retrieve the stored front_image and rear_image of a device type from the API. The API response from `GET /api/dcim/device-types/` returns the URLs which are only accessible with a valid Netbox session but not with the API credentials.

If there isn't already a way to get the images with the API key, it would be nice to be able to do so.

### Use case

Currently, providing the front_image and rear_image in the API response is a dead end, as it cannot be used. For a deeper integration into other systems it would be nice to be able to retrieve the stored images with a API token.

jeremystretch:
Related to #19669

jeremystretch:
Marking this as blocked by #19669 for now as the solution for that will likely solve for this FR as well.

PR #19950, opened 2025-07-24
Fixes: #19669 - Add an API endpoint to download image attachments

### LINKED ISSUE #19669: Media assets not reachable anymore through api requests since media requires Authentication
### Deployment Type

Self-hosted

### NetBox Version

since 4.1.7

### Python Version

3.12

### Steps to Reproduce

1. Use the api to retrieve image attachments
2. Try to retrieve the attachment by the url provided with the api key
3. this gives a failure when LOGIN_REQUIRED is set to true - related to https://github.com/netbox-community/netbox/pull/17990

### Expected Behavior

Allow a way to retrieve an image by url with an API key. 

### Observed Behavior

an error occurs because the request is not authenticated.

DanSheps:
### Fixes: #19669 - Add an API endpoint to download image attachments

* Add a detailed API endpoint for downloading image attachment(s)

### DIFF
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -229,7 +229,6 @@ class JournalEntryViewSet(NetBoxModelViewSet):
 # Config contexts
 #
 
-
 class ConfigContextViewSet(SyncedDataMixin, NetBoxModelViewSet):
     queryset = ConfigContext.objects.all()
     serializer_class = serializers.ConfigContextSerializer
@@ -336,7 +335,6 @@ class ObjectTypeViewSet(ReadOnlyModelViewSet):
     """
     Read-only list of ObjectTypes.
     """
-
     permission_classes = [IsAuthenticatedOrLoginNotRequired]
     queryset = ObjectType.objects.order_by('app_label', 'model')
     serializer_class = serializers.ObjectTypeSerializer
@@ -346,7 +344,6 @@ class ObjectTypeViewSet(ReadOnlyModelViewSet):
 #
 # User dashboard
 #
-
 
 class DashboardView(RetrieveUpdateDestroyAPIView):
     queryset = Dashboard.objects.all()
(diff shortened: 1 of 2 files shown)


netbox/extras/api/views.py
```diff
@@ -200,6 +202,17 @@ class ImageAttachmentViewSet(NetBoxModelViewSet):
...
+    @action(
+        methods=['GET'],
+        detail=True,
+        url_path='download',
+        url_name='download',
+    )
+    def download(self, request, pk, *args, **kwargs):
+        obj = get_object_or_404(self.queryset, pk=pk)
+        # Render and return the elevation as an SVG drawing with the correct content type
+        return serve(request, obj.image.path, document_root=settings.MEDIA_ROOT)
```
bctiemann: Is this path construction correct? When I run this I get:

```
GET /api/extras/image-attachments/1/download/
HTTP 404 Not Found
Allow: GET, HEAD, OPTIONS
Vary: Accept

{
    "detail": "“/Users/btman/Development/netbox-community/netbox/netbox/media/Users/btman/Development/netbox-community/netbox/netbox/media/image-attachments/device_27_20240730_212632-51361-scaled.jpg” does not exist"
}
```

My `MEDIA_ROOT` is `/Users/btman/Development/netbox-community/netbox/netbox/media`

It works if I set `document_root="/"`

bctiemann (changes requested):
See inline comment.

netbox/extras/api/views.py
DanSheps: Odd, it did work when I tested it.

DanSheps: I am going to try and construct a test for this as well.  Just need to figure out the best way to do that.

jeremystretch:
We should also consider FR #18396, which concerns front & rear images for device types. The proposed new endpoint serves image attachments, but not image fields in general. What should we do for other images? A consistent approach would be preferable.

DanSheps:
Per discussion in maintainers meeting, we will be allowing Token Authentication against the /media/* URL

PR #19951, opened 2025-07-24
Fix #19910: Add conditional to hide internet dependent links in an isolated deployment

### LINKED ISSUE #19910: Remove internet dependent links when running in an air gapped environment
### NetBox version

v4.3.4

### Feature type

### Proposed functionality

Hide links dependant on internet access when running in an air gapped environment (as indicated by using the ISOLATED_DEPLOYMENT system parameter).

This would be the 'Get Cloud', 'Get Enterprise', and the links to the GitHub repository and NetDev slack.

### Use case

Users can't open the links anyway, so why show them? 

### Database changes

None

### External dependencies

_No response_

rodvand:
### Fixes: #19910

In the base template, add conditionals on internet links to hide them when the setting ISOLATED_DEPLOYMENT is set to True.

### DIFF
--- a/netbox/templates/base/layout.html
+++ b/netbox/templates/base/layout.html
@@ -199,21 +199,19 @@ Blocks:
                   </li>
 
                 {# Community links #}
-                {% else %}
-                  {% if not settings.ISOLATED_DEPLOYMENT %}
-                    {# GitHub #}
-                    <li class="list-inline-item">
-                      <a href="https://github.com/netbox-community/netbox" target="_blank" class="link-secondary" rel="noopener" aria-label="{% trans "Source Code" %}">
-                        <i title="{% trans "Source Code" %}" class="mdi mdi-github text-primary" data-bs-placement="top" data-bs-toggle="tooltip"></i>
-                      </a>
-                    </li>
-                    {# NetDev Slack #}
-                    <li class="list-inline-item">
-                      <a href="https://netdev.chat" target="_blank" class="link-secondary" rel="noopener" aria-label="{% trans "Community" %}">
-                        <i title="{% trans "Community" %}" class="mdi mdi-slack text-primary" data-bs-placement="top" data-bs-toggle="tooltip"></i>
-                      </a>
-                    </li>
-                  {% endif %}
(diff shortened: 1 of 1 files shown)


netbox/templates/base/layout.html
```diff
@@ -200,18 +200,20 @@ <h1 class="navbar-brand pb-0">
...
-                      <i title="{% trans "Source Code" %}" class="mdi mdi-github text-primary" data-bs-placement="top" data-bs-toggle="tooltip"></i>
-                    </a>
-                  </li>
-                  {# NetDev Slack #}
-                  <li class="list-inline-item">
-                    <a href="https://netdev.chat" target="_blank" class="link-secondary" rel="noopener" aria-label="{% trans "Community" %}">
-                      <i title="{% trans "Community" %}" class="mdi mdi-slack text-primary" data-bs-placement="top" data-bs-toggle="tooltip"></i>
-                    </a>
-                  </li>
+                  {% if not settings.ISOLATED_DEPLOYMENT %}
```
jnovinger: Let's combine this `{% if not ... %}` into the `{% else %}` above it with `{% elif not ... %}`. That will allow us to remove the level of indentation that would be necessary otherwise and keep the template more readable.

netbox/templates/base/layout.html
```diff
@@ -184,7 +184,7 @@ <h1 class="navbar-brand pb-0">
                 {% endif %}
 
                 {# Commercial links #}
-                {% if settings.RELEASE.features.commercial %}
+                {% if settings.RELEASE.features.commercial and not settings.ISOLATED_DEPLOYMENT %}
```
jnovinger, line 187: ~These might be mutually exclusive, but I'm not positive. I'm checking now and will have an answer for you ASAP.~ 

Never mind on this one, I was incorrect and your approach is.

jnovinger (changes requested):
Hey @rodvand , thanks for the quick turnaround! I marked one item for change and possibly another. I'll update the comment as soon as I can.

jnovinger (approved):

PR #19961, opened 2025-07-28
Fixes: #19917 - Fix MAC address pagination duplicates by adding 'pk' to model ordering

### LINKED ISSUE #19917: Inconsistent data being returned from new dcim.mac_addresses endpoint
### Deployment Type

Self-hosted

### NetBox Version

v4.3.4

### Python Version

3.11

### Steps to Reproduce

BUG: Total interfaces from `dcim.mac_addresses` endpoint (count 7978) does not match total interfaces from `dcim.interfaces` endpoint (count 8628)

After having stumbled upon a bug present in `v4.1.5` with `IPAM`, which I successfully reproduced here https://github.com/jseifeddine/netbox-ipam-bug/ 

Realizing that it was fixed in latest version from v4.3.3 - we upgraded to the newest `v4.3.4` - and now experiencing a very similar issue with the new MAC addresses objects on the new `dcim.mac_addresses` endpoint. (mac objects and endpoint introduced in v4.2.0)

We use `pynetbox` to maintain "synchronization" between devices and NetBox.

Syncing interface information (Status, Description, IPs, speed, type, mac address etc) 

Our sync script only makes changes when it needs to, doing a delta check before hand to figure out what in NetBox needs updating/deleting/creating. 

IP addresses were an issue, as per the bug repository attached above - causing unnecessary operations.

That was fixed with IP addresses, but MAC addresses now have similar issues. I haven't yet been able to reproduce it exactly how I did with the IPAM repo above, but in our dev environment, which is a clone of production - I can see inconsistent data being returned from the `dcim.mac_addresses` endpoint, here is the best way I can display that:

NetBox v4.3.4

```bash
Found 8748 MAC addresses in NetBox for device 13756 from dcim.mac_addresses endpoint.
Found 9262 interfaces in NetBox for device 13756 from dcim.interfaces endpoint.
Found 8748 MAC addresses on 7978 interfaces with from dcim.mac_addresses endpoint.
Found 8748 MAC addresses on 8628 interfaces from dcim.interfaces endpoint.
BUG: Total interfaces from dcim.mac_addresses endpoint (7978) does not match total interfaces from dcim.interfaces endpoint (8628)

Test complete
```

Running this script;
```python
#!venv/bin/python

from dotenv import load_dotenv
import os
import pynetbox

load_dotenv()

nb = pynetbox.api(
    url=os.getenv("NETBOX_API_URL"),
    token=os.getenv("NETBOX_API_TOKEN"),
    threading=True
)

DEVICE_ID = 13756

all_netbox_macs = list(nb.dcim.mac_addresses.filter(device_id=DEVICE_ID))
all_netbox_interfaces = list(nb.dcim.interfaces.filter(device_id=DEVICE_ID))

print(f"Found {len(all_netbox_macs)} MAC addresses in NetBox for device {DEVICE_ID} from dcim.mac_addresses endpoint.")
print(f"Found {len(all_netbox_interfaces)} interfaces in NetBox for device {DEVICE_ID} from dcim.interfaces endpoint.")

mac_interfaces = {}
interface_macs = {}

for mac in all_netbox_macs:
    if mac.assigned_object and mac.assigned_object.name:
        if not mac_interfaces.get(mac.assigned_object.name):
            mac_interfaces[mac.assigned_object.name] = []
        mac_interfaces[mac.assigned_object.name].append(mac)

total_macs = sum(len(macs) for macs in mac_interfaces.values())
print(f"Found {total_macs} MAC addresses on {len(mac_interfaces)} interfaces with from dcim.mac_addresses endpoint.")

for interface in all_netbox_interfaces:
    if interface.mac_addresses:
        if not interface_macs.get(interface.name):
            interface_macs[interface.name] = []
        for mac in interface.mac_addresses:
            interface_macs[interface.name].append(mac)

total_macs_from_interface = sum(len(macs) for macs in interface_macs.values())
print(f"Found {total_macs_from_interface} MAC addresses on {len(interface_macs)} interfaces from dcim.interfaces endpoint.")

# Expose bug in mac / interface counts returned from different endpoints

# total_macs is the total number of MAC addresses on the device from the dcim.mac_addresses endpoint
# total_macs_from_interface is the total number of MAC addresses on all interfaces from the dcim.interfaces endpoint
if total_macs != total_macs_from_interface:
    print(f"BUG: Total MAC addresses from dcim.mac_addresses endpoint ({total_macs}) does not match total MAC addresses from dcim.interfaces endpoint ({total_macs_from_interface})")

# mac_interfaces[mac.assigned_object.name] -> list of MAC addresses from the dcim.mac_addresses endpoint
# interface_macs[interface.name] -> list of MAC addresses from the dcim.interfaces endpoint
if len(mac_interfaces) != len(interface_macs):
    print(f"BUG: Total interfaces from dcim.mac_addresses endpoint ({len(mac_interfaces)}) does not match total interfaces from dcim.interfaces endpoint ({len(interface_macs)})")

print()
print("Test complete")
```

### Expected Behavior

Consistent data from the new MAC Addresses endpoint 

For example, correct interface counts derived from assigned_object 

### Observed Behavior

BUG: Total interfaces from dcim.mac_addresses endpoint (7978) does not match total interfaces from dcim.interfaces endpoint (8628)

```python
# mac_interfaces[mac.assigned_object.name] -> list of MAC addresses from the dcim.mac_addresses endpoint
# interface_macs[interface.name] -> list of MAC addresses from the dcim.interfaces endpoint
if len(mac_interfaces) != len(interface_macs):
    print(f"BUG: Total interfaces from dcim.mac_addresses endpoint ({len(mac_interfaces)}) does not match total interfaces from dcim.interfaces endpoint ({len(interface_macs)})")
```

jseifeddine:
### Fixes: #19917

Fix MAC address pagination by ensuring deterministic ordering
Add `pk` to MACAddress model ordering to ensure deterministic results
when multiple MAC addresses have the same value. This prevents the same
MAC address from appearing on multiple pages during pagination.

The issue occurred because Django's default ordering by 'mac_address'
alone is non-deterministic when multiple records share the same MAC
address value, causing inconsistent pagination results.


netbox/dcim/models/devices.py
```diff
@@ -1272,7 +1272,7 @@ class MACAddress(PrimaryModel):
     )
 
     class Meta:
-        ordering = ('mac_address',)
+        ordering = ('mac_address', 'id',)
```
DanSheps:
```suggestion
        ordering = ('mac_address', 'pk',)
```

I believe we use `pk` instead of `id` in all other locations where we have done this.

netbox/dcim/migrations/0209_alter_macaddress_options.py
```diff
@@ -0,0 +1,15 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dcim', '0208_devicerole_uniqueness'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='macaddress',
+            options={'ordering': ('mac_address', 'id'), 'verbose_name': 'MAC address', 'verbose_name_plural': 'MAC addresses'},
```
DanSheps:
```suggestion
            options={'ordering': ('mac_address', 'pk'), 'verbose_name': 'MAC address', 'verbose_name_plural': 'MAC addresses'},
```

DanSheps (changes requested):

jseifeddine:
Thanks @DanSheps 

Done https://github.com/jseifeddine/netbox/commit/83b312a3dcbdb194b0af8c0b5e30f5007d77c5ce

jseifeddine:
> Thanks @DanSheps
> 
> Done [jseifeddine@83b312a](https://github.com/jseifeddine/netbox/commit/83b312a3dcbdb194b0af8c0b5e30f5007d77c5ce)

sorry, just saw the failed builds, 
Will address this

jseifeddine:
should be good to go... thanks

jseifeddine:
Cleaned up the tests and able to reproduce in both `dcim.tests.test_filtersets` and `dcim.tests.test_api`

```bash
System check identified no issues (0 silenced).
test_mac_address_pagination_duplicates (dcim.tests.test_filtersets.MACAddressTestCase.test_mac_address_pagination_duplicates) ... FAIL
test_mac_address_pagination_duplicates (dcim.tests.test_api.MACAddressTest.test_mac_address_pagination_duplicates) ... FAIL

FAIL: test_mac_address_pagination_duplicates (dcim.tests.test_filtersets.MACAddressTestCase.test_mac_address_pagination_duplicates)
  File "/opt/netbox/netbox/dcim/tests/test_filtersets.py", line 6959, in test_mac_address_pagination_duplicates
    self.assertEqual(len(all_ids & page_ids), 0)
AssertionError: 1 != 0

FAIL: test_mac_address_pagination_duplicates (dcim.tests.test_api.MACAddressTest.test_mac_address_pagination_duplicates)
  File "/opt/netbox/netbox/dcim/tests/test_api.py", line 2696, in test_mac_address_pagination_duplicates
    self.assertEqual(len(all_mac_ids & page_ids), 0)
AssertionError: 1 != 0

Ran 2 tests in 0.271s

FAILED (failures=2)
```

with the proposed `pk` order fix:

```bash
test_mac_address_pagination_duplicates (dcim.tests.test_filtersets.MACAddressTestCase.test_mac_address_pagination_duplicates) ... ok
test_mac_address_pagination_duplicates (dcim.tests.test_api.MACAddressTest.test_mac_address_pagination_duplicates) ... ok

Ran 2 tests in 0.421s

OK
```

netbox/dcim/tests/test_filtersets.py
```diff
@@ -6940,3 +6940,22 @@ def test_vminterface(self):
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
         params = {'vminterface': [vm_interfaces[0].name, vm_interfaces[1].name]}
         self.assertEqual(self.filterset(params, self.queryset).qs.count(), 2)
+
+    def test_mac_address_pagination_duplicates(self):
```
jeremystretch: Please remove this as it's not a filterset test.

netbox/dcim/tests/test_api.py
```diff
@@ -2674,3 +2674,25 @@ def setUpTestData(cls):
                 'mac_address': '00:00:00:00:00:06',
             },
         ]
+
+    def test_mac_address_pagination_duplicates(self):
```
jeremystretch: This test can be simplified to check ordering of the base queryset; it doesn't need to invoke API pagination. For example, you could create ten duplicate MAC addresses, then check that `MACAddress.objects.all()` returns them in order of numeric ID.

Also, this test should be moved to `test_models` as it doesn't involve the REST API.

jeremystretch (changes requested):

netbox/dcim/tests/test_api.py
jseifeddine: @jeremystretch 

`netbox/dcim/tests/test_models.py`
```python
class MACAddressTestCase(TestCase):
    @classmethod
    def setUpTestData(cls):
...        
        cls.device2 = Device.objects.create(
            name='Device 2', device_type=device_type, role=device_role, site=site,
        )
        
        shared_mac = '1234567890cd'
        interfaces = []
        for i in range(10000):
            interfaces.append(Interface(
                device=device, name=f'Interface {i:03d}', type='1000base-t'
            ))
        Interface.objects.bulk_create(interfaces)
        
        mac_addresses = []
        for interface in interfaces:
            mac_addresses.append(MACAddress(mac_address=shared_mac, assigned_object=interface))
        MACAddress.objects.bulk_create(mac_addresses)
...

    @tag('regression')
    def test_mac_address_ordering_with_duplicates(self):
        """Test that MAC addresses with the same value are ordered by their numeric ID."""
        mac_addresses = list(MACAddress.objects.all())
        for i in range(1, len(mac_addresses)):
            self.assertLess(mac_addresses[i-1].pk, mac_addresses[i].pk)
            self.assertGreater(mac_addresses[i].pk, mac_addresses[i-1].pk)
```

I can't get this to fail easily (without `pk` ordering fix)

even passed with 10,000 duplicate macs 
```bash
test_mac_address_ordering_with_duplicates (dcim.tests.test_models.MACAddressTestCase.test_mac_address_ordering_with_duplicates)
Test that MAC addresses with the same value are ordered by their numeric ID. ... ok

Ran 1 test in 22.340s
```

it did **FAIL** with 100,000
```bash
Test that MAC addresses with the same value are ordered by their numeric ID. ... FAIL

FAIL: test_mac_address_ordering_with_duplicates (dcim.tests.test_models.MACAddressTestCase.test_mac_address_ordering_with_duplicates)
Test that MAC addresses with the same value are ordered by their numeric ID.
  File "/opt/netbox/netbox/dcim/tests/test_models.py", line 75, in test_mac_address_ordering_with_duplicates
    self.assertLess(mac_addresses[i-1].pk, mac_addresses[i].pk)
AssertionError: 100002 not less than 43691

Ran 1 test in 67.364s
```

with the `pk` ordering fix, it passes - even with 100,000
```bash
Test that MAC addresses with the same value are ordered by their numeric ID. ... ok

Ran 1 test in 36.533s

OK
```

I don't want to commit the 100,000 test and blow out the netbox tests duration

Please advise

jseifeddine:
@jeremystretch 

I've removed the tests completely just keeping the actual fix, 

```
netbox (19917-fix-mac-address-pagination-duplicates) ✔ ruff check netbox/ 
All checks passed!
```

jseifeddine:
OK @jeremystretch 

I added the appropriate test, with 10 identical mac address values; 

It now fails without the `pk` ordering fix;
```bash
Test that MAC addresses with the same value are ordered by their primary key (pk). ... FAIL

FAIL: test_mac_address_ordering_with_duplicates (dcim.tests.test_models.MACAddressTestCase.test_mac_address_ordering_with_duplicates)
Test that MAC addresses with the same value are ordered by their primary key (pk).
  File "/opt/netbox/netbox/dcim/tests/test_models.py", line 82, in test_mac_address_ordering_with_duplicates
    self.assertLess(
AssertionError: 12 not less than 1 : MAC addresses with value 12:34:56:78:90:AB are not ordered by primary key

Ran 1 test in 0.027s
```

and passes with the `pk` ordering fix;
```bash
Test that MAC addresses with the same value are ordered by their primary key (pk). ... ok

Ran 1 test in 0.028s

OK
```

jeremystretch (dismissed):
Thanks @jseifeddine. I've removed the test as I don't feel it's necessary; there are many other models with a similar pattern for which we don't include an explicit regression test for ordering. With `pk` added to the model's default ordering this shouldn't be an issue.

jeremystretch (approved):

DanSheps (approved):

PR #19962, opened 2025-07-28
Closes #18797: Support path import for certain Jinja environment parameters

### LINKED ISSUE #18797: Config Template Rendering: Option for jinja2.StrictUndefined
### NetBox version

v4.1.11 (code is untouched since 2023 and still active in v4.2.4)

### Feature type

Data model extension

### Proposed functionality

tl;dr: Being able to make Jinja2 Template variables `StrictUndefined` instead of default `Undefined`, to error out instead of silently rendering an invalid config.

Currently Config Templates for Devices are Jinja2 Templates.
On a Config Template Model the User can configure "Environment Parameters" in JSON format.
These are passed as kwargs to the Jinja2 Renderer:
https://github.com/netbox-community/netbox/blob/v4.2.4/netbox/extras/models/configs.py#L302

However Jinja2 Environment Parameters are not completely compatible to JSON Format, as they're Python.
This leads to one of the most important settings not being usable: `undefined`

See https://jinja.palletsprojects.com/en/stable/api/#jinja2.Environment
* default: https://jinja.palletsprojects.com/en/stable/api/#jinja2.Undefined
* alternative: https://jinja.palletsprojects.com/en/stable/api/#jinja2.StrictUndefined

As both are python classes, one would have to properly pass that python class to the environment, which is not possible in JSON format.

To enable the User to set `jinja2.StrictUndefined` there could be
* a global setting
  * which would be enough, as Jinja2 has the `default` filter to work with undefined variables
* a specifically JSON-parsed & Python-importlib handled Environment Parameter
* a dedicated attribute on an Config Template

UI can already deal with this, however API cannot, as it tries to provide a nice error message relying on a missing attribute:
Image

### Use case

One could use NetBox own documentation on the feature as an example: https://netbox.readthedocs.io/en/stable/features/configuration-rendering/#configuration-templates
* if a device name is `None`
* if ntp_servers is undefined

The config would look like:
```
    system {
        host-name ;
        domain-name example.com;
        time-zone UTC;
        authentication-order [ password radius ];
        ntp {
            
        }
    }
```
On a hypothetical network device the ntp section would probably not cause an error while an empty host-name would. This is not cleanly predictable.

My Use Case is: I want to know if things are broken and not having to debug the rendered configuration that may well be accepted by the network device, applied and cause an outage.

If a var is missing, it should not render a configuration.

An incomplete workaround is to always test with an `{% if x is defined %}...{% endif %}` which is very verbose and still does not provide any error-functionality. The rendered configuration will still seem ok unless any parser/the target network device rejects it.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #18797

Adds support for dotted path imports for Jinja environment params `undefined` and `finalize`, for both config templates and export templates.

- Introduce and utilize the new `get_environment_params()` method on RenderTemplateMixin

### DIFF
--- a/netbox/extras/models/mixins.py
+++ b/netbox/extras/models/mixins.py
@@ -2,15 +2,16 @@ import importlib.abc
 import importlib.util
 import os
 import sys
+
 from django.core.files.storage import storages
 from django.db import models
-from django.utils.translation import gettext_lazy as _
 from django.http import HttpResponse
+from django.utils.module_loading import import_string
+from django.utils.translation import gettext_lazy as _
 
-from extras.constants import DEFAULT_MIME_TYPE
+from extras.constants import DEFAULT_MIME_TYPE, JINJA_ENV_PARAMS_WITH_PATH_IMPORT
 from extras.utils import filename_from_model, filename_from_object
 from utilities.jinja2 import render_jinja2
-
 
 __all__ = (
     'PythonModuleMixin',
@@ -125,12 +126,22 @@ class RenderTemplateMixin(models.Model):
             class_name=self.__class__
         ))
 
+    def get_environment_params(self):
+        """
+        Pre-processing of any defined Jinja environment parameters (e.g. to support path resolution).
+        """
+        params = self.environment_params or {}
+        for name, value in params.items():
+            if name in JINJA_ENV_PARAMS_WITH_PATH_IMPORT and type(value) is str:
+                params[name] = import_string(value)
+        return params
+
     def render(self, context=None, queryset=None):
         """
         Render the template with the provided context. The context is passed to the Jinja2 environment as a dictionary.
         """
         context = self.get_context(context=context, queryset=queryset)
-        env_params = self.environment_params or {}
+        env_params = self.get_environment_params()
         output = render_jinja2(self.template_code, context, env_params, getattr(self, 'data_file', None))
 
         # Replace CRLF-style line terminators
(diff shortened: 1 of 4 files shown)


netbox/extras/models/mixins.py
```diff
@@ -125,12 +126,22 @@ def get_context(self, context=None, queryset=None):
...
 
+    def get_environment_params(self):
+        """
+        Pre-processing of any defined Jinja environment parameters (e.g. to support path resolution).
+        """
+        params = self.environment_params or {}
+        for name, value in params.items():
+            if name in JINJA_ENV_PARAMS_WITH_PATH_IMPORT and type(value) is str:
+                params[name] = import_string(value)
+        return params
```
jnovinger, line 137: Couple of questions here:

1. Should we do any validation of `value`? E.g. make sure it's within a set of accepted values?
2. I assume `import_string()` might raise some exceptions, especially around dotted-paths that don't exist. Is the intention to let those bubble up to the user?

jnovinger (commented):
Had a couple of questions. Given new behavior, would love to see some testing. Lastly, is it worth mentioning this in the docs somewhere? Perhaps even in just the `ConfigTemplate` model docs for the [environment params](https://netbox.readthedocs.io/en/stable/models/extras/configtemplate/#environment-parameters) field.

netbox/extras/models/mixins.py
jeremystretch: 1. What would the valid values be? I don't think we have a way of knowing, unless we limited it to a very small, prescribed set of choices.
2. Yes; exceptions will be raised to surface the invalid configuration.

jeremystretch:
> Given new behavior, would love to see some testing.

Unfortunately there's no existing test case for this functionality which can be expanded to cover the modest addition in functionality. I've opened #19971 for new test cases to be added, but it's beyond the scope of this FR IMO.

netbox/extras/models/mixins.py
jnovinger: Good point.

jnovinger (approved):

PR #19963, opened 2025-07-28
Closes #19945: Create DecimalVar class for custom script input

### LINKED ISSUE #19945: Create DecimalVar class for custom script input
### NetBox version

v4.2.9

### Feature type

Data model extension

### Proposed functionality

Create a new class in `netbox/extras/scripts.py` called `DecimalVar` that enables custom scripts to take in decimal numbers as user input. This would be functionally identical to IntegerVar except it would allow decimal numbers.

```python
class DecimalVar(ScriptVariable):
    """
    Decimal representation. Can enforce minimum/maximum values.
    """

    form_field = forms.DecimalField

    def __init__(self, min_value=None, max_value=None, *args, **kwargs):

        # Optional minimum/maximum values
        if min_value:
            self.field_attrs["min_value"] = min_value
        if max_value:
            self.field_attrs["max_value"] = max_value
```

### Use case

There is currently no built-in way to get decimal input from a user in a custom script (unless I'm just overlooking something obvious). IntegerVar exists for whole number input, but it cannot be used to get, for example, GPS coordinates from a user trying to create a site from a custom script.

Adding DecimalVar would enable the following:

```python
# get data from user:

site_lat = DecimalVar(
    label="Latitude",
)

site_long = DecimalVar(
    label="Longitude",
)

# run:
site = Site(
    name="name",
    slug="slug",
    region="region",
    latitude=data["site_lat"],
    longitude=data["site_long"],
)
```

### Database changes

None

### External dependencies

None

kyerlasswell:
### Fixes: #19945

- Adds a new class called `DecimalVar` to `netbox/extras/scripts.py` to take in decimal input from a form. DecimalVar takes the same arguments as `django.forms.DecimalField`: `min_value`, `max_value`, `max_digits` and `decimal_places`.

- Adds 5 new tests to `netbox/extras/tests/test_scripts.py`. One to isolate each of the four parameters and one to validate valid data. Two objects are created with different parameters because `max_value` will always be hit before `max_digits` can come into play so they have to be tested separately (e.g. if `max_value = 999` but `max_digits = 4`, you can't test `max_digits` because a value of `1000` is already invalid).

- Adds DecimalVar to custom script documentation.

### DIFF
--- a/docs/customization/custom-scripts.md
+++ b/docs/customization/custom-scripts.md
@@ -275,6 +275,15 @@ Stores a numeric integer. Options include:
 * `min_value` - Minimum value
 * `max_value` - Maximum value
 
+### DecimalVar
+
+Stores a numeric decimal. Options include:
+
+* `min_value` - Minimum value
+* `max_value` - Maximum value
+* `max_digits` - Maximum number of digits, including decimal places
+* `decimal_places` - Number of decimal places
+
 ### BooleanVar
 
 A true/false flag. This field has no options beyond the defaults listed above.
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -31,6 +31,7 @@ __all__ = (
     'DateTimeVar',
     'FileVar',
     'IntegerVar',
+    'DecimalVar',
     'IPAddressVar',
     'IPAddressWithMaskVar',
     'IPNetworkVar',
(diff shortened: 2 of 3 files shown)


jeremystretch (approved):
Looks good, thanks @kyerlasswell!

kyerlasswell:
Happy I could contribute, @jeremystretch!

PR #19984, opened 2025-07-30
Closes #19977: Denormalize device relationships on component models

### LINKED ISSUE #19977: Cache the assigned site, location, and rack of device components
### NetBox version

v4.3.5

### Feature type

New functionality

### Proposed functionality

For all device components, cache the assigned site and location of the parent device. This assignment will be read-only, and updated automatically whenever the device's assignment is altered.

There is an [existing pattern](https://github.com/netbox-community/netbox/blob/6ca390871525c248b56f9e941bc4f1fa28b86a69/netbox/dcim/signals.py#L19) for how this can be handled using a receiver for the `post_save` signal.

### Use case

Saving these assignments directly on each component will obviate the need for SQL table joins when filtering device components by parent site or location. For example, consider the following REST API query:

```
GET /api/dcim/interfaces/?site_id=123
```

This currently requires a `JOIN` of the `dcim_interface` table with the `dcim_device` table to query the site. Copying this relationship to the `dcim_interface` table allows the filter to query that table directly.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #19977

- Adds `_site`, `_location`, and `_rack` ForeignKeys on all device component models
  - The included database migration automatically populates these new fields for existing objects
- Introduce the `handle_device_site_change()` `post_save` receiver to automatically update child components when a device is modified
- Update tests
  - Manually populate these fields where necessary when `bulk_create()` is employed
  - Tweak GraphQLTestCase to ignore private ForeignKey fields when constructing test queries

### DIFF
--- a/netbox/dcim/migrations/0209_device_component_denorm_site_location.py
+++ b/netbox/dcim/migrations/0209_device_component_denorm_site_location.py
@@ -1,0 +1,247 @@
+import django.db.models.deletion
+from django.db import migrations, models
+from django.db.models import OuterRef, Subquery
+
+
+def populate_denormalized_data(apps, schema_editor):
+    Device = apps.get_model('dcim', 'Device')
+    component_models = (
+        apps.get_model('dcim', 'ConsolePort'),
+        apps.get_model('dcim', 'ConsoleServerPort'),
+        apps.get_model('dcim', 'PowerPort'),
+        apps.get_model('dcim', 'PowerOutlet'),
+        apps.get_model('dcim', 'Interface'),
+        apps.get_model('dcim', 'FrontPort'),
+        apps.get_model('dcim', 'RearPort'),
+        apps.get_model('dcim', 'DeviceBay'),
+        apps.get_model('dcim', 'ModuleBay'),
+        apps.get_model('dcim', 'InventoryItem'),
+    )
+
+    for model in component_models:
+        subquery = Device.objects.filter(pk=OuterRef('device_id'))
+        model.objects.update(
+            _site=Subquery(subquery.values('site_id')[:1]),
+            _location=Subquery(subquery.values('location_id')[:1]),
+            _rack=Subquery(subquery.values('rack_id')[:1]),
+        )
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('dcim', '0208_devicerole_uniqueness'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='consoleport',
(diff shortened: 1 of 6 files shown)


netbox/dcim/models/device_components.py
```diff
@@ -100,6 +123,14 @@ def clean(self):
...
             })
 
+    def save(self, *args, **kwargs):
+        # Save denormalized references
+        self._site = self.device.site
+        self._location = self.device.location
+        self._rack = self.device.rack
+
+        super().save(*args, **kwargs)
+
```
jnovinger, line 133: Out of curiosity, any reason to not use the signal handler approach for this as well?

netbox/dcim/migrations/0209_device_component_denorm_site_location.py
```diff
@@ -0,0 +1,287 @@
...
+        apps.get_model('dcim', 'InventoryItem'),
+    )
+
+    for model in component_models:
+        subquery = Device.objects.filter(pk=OuterRef('device_id'))
+        model.objects.update(
+            _site=Subquery(subquery.values('site_id')[:1]),
+            _location=Subquery(subquery.values('location_id')[:1]),
+            _rack=Subquery(subquery.values('rack_id')[:1]),
+        )
```
jnovinger, line 27: Any concern here about how long these updates might lock up a whole table, particularly `dcim_interface`? I would imagine a series of batched updates would results more, but shorter locks, meaning that other operations on the tables can happen, if needed.

I'd be curious if you had a chance to run this against a DB with 1M+ interfaces.

Having said all that, I'm far from certain this is really an issue.

jnovinger (commented):
One question about the migrations and one (much less important) question about saving denormalized values via `.save()`. Please and thank you.

netbox/dcim/models/device_components.py
jeremystretch: Not really. Maybe slightly cleaner, more direct?

netbox/dcim/migrations/0209_device_component_denorm_site_location.py
jeremystretch: During testing with ~500K interfaces, the migration took only a few seconds to run, so I'm not _too_ concerned about it. It's a valid point, but I'm not sure how we'd meaningfully split up the query to batch it. I wanted to use subqueries to avoid handling any of the bulk data in Python.

jnovinger (approved):

PR #19991, opened 2025-07-31
Closes #19968: Use  multiple selection lists for the assignment of object types when editing a permission

jeremystretch:
### Closes: #19968

- Introduced SplitMultiSelectWidget to render side-by-side multi-select widgets (one for available choices and one for selected choices)
  - Also introduced the AvailableOptions and SelectedOptions subwidgets, which are not intended for general use
- Update and extend TypeScript to support moving options between the two widgets
- Changed the `object_types` field on ObjectPermissionForm to use SplitMultiSelectWidget
- Updated the table configuration form templates to work with the updated TypeScript


jeremystretch:
CodeQL failure is unrelated to these changes

jnovinger (approved):
Looks awesome!

PR #19994, opened 2025-07-31
Fixes #19987: Show changelog_message field only for models which support change logging

### LINKED ISSUE #19987: Changelog message appears on delete confirmation dialog even if not a change logged model
### Deployment Type

Self-hosted

### NetBox Version

`feature` @ 40dd36812c63ca1af43a4feb442dbce0900cf883

### Python Version

3.10

### Steps to Reproduce

1. Use the feature branch
2. create or go to an existing user
3. delete the user

### Expected Behavior

The Confirm Deletion dialog should not have a Changelog message entry at the bottom as it is not a change-logged model and the message won't be saved anywhere.

### Observed Behavior

The changelog message entry appears on the form:

Image

jeremystretch:
### Fixes: #19987

- Rename the ChangeLoggingMixin _form_ mixin to ChangelogMessageMixin to avoid confusion
- Move BulkDeleteForm to a proper class, defined outside the view
- Alter signature of DeleteForm to support passing a model instance
- Alter signature of BulkDeleteForm to require passing a model class
- Tweak both forms to omit the `changelog_message` field is the associated model does not support change logging

### DIFF
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -11,7 +11,7 @@ from ipam.choices import VLANQinQRoleChoices
 from ipam.models import ASN, VLAN, VLANGroup, VRF
 from netbox.choices import *
 from netbox.forms import NetBoxModelBulkEditForm
-from netbox.forms.mixins import ChangeLoggingMixin
+from netbox.forms.mixins import ChangelogMessageMixin
 from tenancy.models import Tenant
 from users.models import User
 from utilities.forms import BulkEditForm, add_blank_choice, form_from_model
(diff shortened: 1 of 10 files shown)


bctiemann:
@jeremystretch I noticed that if you enter a change log message while operating on a "parent" object (like a Device), the message you enter is only applied to that object in the change logs, and all the child objects that are cascade-deleted get empty change log messages. Is this intentional? Or should the messages cascade down too?

Screenshot 2025-08-01 at 2 35 32 PM

Otherwise approved.

bctiemann (approved):

jeremystretch:
> Is this intentional? Or should the messages cascade down too?

I was on the fence about this; there are valid arguments for both behaviors. However, there's also a technical limitation that prevents us from replicating the changelog message from the primary object being deleted to objects deleted via cascade.

Given this limitation and the fact that all the resulting change records are correlated via request ID, I think it's reasonable to leave as-is.

PR #20000, opened 2025-08-01
Configure CodeQL to ignore compiled JS resources

jeremystretch:
- Introduces a basic CodeQL workflow (simplified from the default GitHub provides)
- Introduces a CodeQL configuration file to exempt the `netbox/project-static/dist/` path from scanning

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,9 +2,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "main" ]
+    branches: [ "main", "feature" ]
   pull_request:
-    branches: [ "main" ]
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,42 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
jnovinger: Should we also include push and PR events for `feature`?

jeremystretch: :thinking: It wasn't in the original config GitHub gave me, but makes sense to add it.

jnovinger (approved):

PR #20022, opened 2025-08-05
Closes #20008: Job logging for bulk operation background jobs

jeremystretch:
### Closes: #20008

- Extend the `terminate()` method on Job to automatically notify the assigned user (if any) on completion
- AsyncViewJob no longer records any returned data from the view being executed
- AsyncViewJob now attaches the job being run to the request object to support logging (`is_background_request()` has been updated accordingly)
- Extend BulkImportView, BulkEditView, and BulkDeleteView to record success & error messages in the job log when executed as background jobs
- Introduce `_compile_form_errors()` on BulkImportView for consistent form error handling

### DIFF
--- a/netbox/core/models/jobs.py
+++ b/netbox/core/models/jobs.py
@@ -18,8 +18,10 @@ from rq.exceptions import InvalidJobOperation
 
 from core.choices import JobStatusChoices
 from core.dataclasses import JobLogEntry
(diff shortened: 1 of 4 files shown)


jnovinger (approved):
Very solid, I wasn't able to find any holes at all!

PR #20024, opened 2025-08-05
Fixes #19379: allow standalone id in vlan-ids range list

### LINKED ISSUE #19379: Unable to specify single VLAN IDs in range when creating a VLAN group
### NetBox version

v4.2.6

### Feature type

### Proposed functionality

When selecting Vlan IDs within a Vlan Group add the option of single vlan IDs rather than just ranges

### Use case

For adding a single vlan id 1900 currently requires 1900-1900, 

allow the use of just 1900,

### Database changes

_No response_

### External dependencies

_No response_

Jathn:
### Fixes #19379: allow standalone id in vlan-ids range list

The changes made allow for VLAN id ranges to be created using both ranges, e.g. 10-15 and single numbers, e.g. 20 and any combination of these. I've also included broader test coverage and fixed a bug where the range "a-b" would previously break the UI.

Changes made:
- Changed the instructions in NumericRangeArrayField to indicate that single numbers can be given.
- Changed utility data function: string_to_ranges  (utilities/data.py line 153) so that it properly handles all of these cases.
- Improved test coverage of utilities.tests.test_data.RangeFunctionsTestCase.test_string_to_ranges

### DIFF
--- a/netbox/utilities/data.py
+++ b/netbox/utilities/data.py
@@ -160,9 +160,19 @@ def string_to_ranges(value):
         return None
     value.replace(' ', '')  # Remove whitespace
     values = []
-    for dash_range in value.split(','):
-        if '-' not in dash_range:
+    for data in value.split(','):
+        dash_range = data.strip().split('-')
+        lower, upper = '', ''
+        if len(dash_range) == 1 and str(dash_range[0]).isdigit():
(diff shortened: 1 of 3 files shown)


jeremystretch (approved):
Thanks @Jathn!

PR #20025, opened 2025-08-05
Closes #20003: Introduce mechanism to register callbacks for webhook context

### LINKED ISSUE #20003: Provide a mechanism for plugins to attach arbitrary context to outgoing webhooks
### NetBox version

v4.3.5

### Feature type

New functionality

### Proposed functionality

Implement a mechanism by which plugins can contribute additional data to outbound webhooks generated in response to an event rule. This mechanism would enable a plugin to register a callback function to be called whenever a webhook is being generated. The callback would accept the object and current request, and return any data to be included in the webhook's payload.

Something like this:

```python
@webhook_data(models=(Site, Location))
def my_plugins_callback(instance, request):
    if instance.tenant is not None:
        return {
            'billing_id': Customer.objects.get(tenant=instance.tenant)
        }
```

Any data returned by the callback will be attached to the webhook payload, e.g. within a `context` key on the JSON object root. (An empty response will be ignored.)

This should be implemented in a way so as not to break the current format of a webhook payload.

### Use case

One example where this would be useful is in the netbox-branching plugin, where it would be helpful to attach the branch in which a change is being made as context. (See netboxlabs/netbox-branching#132 for some discussion around this use case.) However, it's not difficult to image other scenarios where this would be convenient.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #20003

- Simplify the signature of `enqueue_event()` to pass the request object directly (instead of the user & request ID)
- Attach the request object to the event when it is enqueued for processing
- Alter the signature of `process_event_rules()` to take a request object rather than a request ID
- Alter the parameters recorded for `send_webhook` tasks
    - Replace `model_name` with `object_type`
    - Add `request` if present
- Introduce the `process_event_rules()` function for use by plugins
    - Add `webhook_callbacks` to the NetBox registry
- Extend `send_webhook()` to gather additional context data from registered webhook callbacks
- Update event tests


jeremystretch:
One bit I'm still not sure about: Should we namespace the context data to avoid collisions among callbacks from different plugins?

mrmrcoleman:
Capturing some points here (I would like to be corrected on anything that I have misunderstood)

- In the namespacing scenario one would reference variables like `payload.context.pluginname.key`
- In the non-namespacing scenario one would reference variables like `payload.context.key`

The risk is that two plugins define the same key. My instinct tells me to namespace it because even if the risk of collisions is small the user impact is, presulably, large e.g. I get very difficult to solve issues that require me to introspect how NetBox works to figure out

jeremystretch:
I have two counterpoints to share:

1. Plugins can always voluntarily namespace keys by attaching an arbitrary prefix (e.g. `my_plugin_foo`).
2. There may be cases where namespacing is undesirable. For example, support a webhook receiver wants to check for an `error` key set by _any_ plugin. It is much simpler to check for `context.error` than to iterate through all (or some) dictionaries under `context` to determine if any of them contain `error`.

That said, I have no strong opinion either way.

arthanson:
Should this be documented, maybe in the webhooks docs: https://netboxlabs.com/docs/netbox/integrations/webhooks/

arthanson (approved):
Looks good and works, just wondering about documenting this.

docs/plugins/development/webhooks.md
```diff
@@ -0,0 +1,63 @@
...
+
+```python
+from extras.webhooks import register_webhook_callback
+from my_plugin.utilities import get_foo_status
+
+@register_webhook_callback
+def set_foo_status(object_type, event_type, data, request):
+    if status := get_foo_status():
+        return {
+            'foo': status
```
arthanson, line 21: I think we probably want to default to name-spacing here and Put in a note that you are encouraged to namespace the context variables to avoid the chance of collision between plugins - guide them towards best practices.

arthanson (approved):
Re comment on giving the example with namespacing and a note about why you should do this.

PR #20026, opened 2025-08-05
Fixes: #19998 - Add changelog entry when clearing M2M fields

### LINKED ISSUE #19998: No changelogs entry for deleted tags
### NetBox Version

v4.35

### Python Version

3.12

### Steps to Reproduce

Here are the curl api calls to reproduce:
(Assuming you have set a NETBOX_DEMO_TOKEN environment variable)

1. Create Tag

```
curl -X POST "https://demo.netbox.dev/api/extras/tags/" \
  -H "Authorization: Token $NETBOX_DEMO_TOKEN" \
  -d '{"name": "test", "slug": "test"}'
```

2. Add tag to interface

```
curl -X PATCH "https://demo.netbox.dev/api/dcim/interfaces/1/" \
  -H "Authorization: Token $NETBOX_DEMO_TOKEN" \
  -d '{"tags": [{"slug": "test"}]}'
```

3. Remove tag from interface

```
curl -X PATCH "https://demo.netbox.dev/api/dcim/interfaces/1/" \
  -H "Authorization: Token $NETBOX_DEMO_TOKEN" \
  -d '{"tags": []}'
```

### Expected Behavior

After executing the commands above, I'd expect two changelog entries for the target interface. 
However there's is only a single change entry.

Image

The  patch removal of the tag worked though, as it results in an interface without tags.

### Observed Behavior

There should be two changelog entries. 
One for the addition of the tag and another one for the removal.

DanSheps:
### Fixes: #19998

* Add check for clearing of M2M fields to generate object_change records

### DIFF
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -76,7 +76,8 @@ def handle_changed_object(sender, instance, **kwargs):
     elif kwargs.get('action') == 'post_clear':
         # Handle clearing of an M2M field
         if isinstance(Tag, kwargs.get('model')) and getattr(instance, '_prechange_snapshot', {}).get('tags'):
-            # Handle tags as it is a Generic M2M
+            # Handle generation of M2M changes for Tags which have a previous value (ignoring changes where the
+            # prechange snapshot is empty)
             m2m_changed = True
             event_type = OBJECT_UPDATED
         else:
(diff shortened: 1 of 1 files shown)


bctiemann:
@DanSheps I tested this using the repro steps in the bug and the behavior is still unchanged. When I clear the tags from the interface via the API, the second changelog entry is not created.

bctiemann (changes requested):
Tested -- clearing tags on interface via API does not create the second changelog entry (same behavior as in bug report).

DanSheps:
Hey Brian,

I must have made a change to using isinstance() after I tested as part of code cleanup.  Thanks for the catch.  Should be good now

bctiemann (approved):

PR #20041, opened 2025-08-07
Fixes: #19645 - Correct Interface selection for Cable add when VC master is the selected device

### LINKED ISSUE #19645: VC Member Interfaces Missing from Cable Add View
### Deployment Type

Self-hosted

### NetBox Version

v4.2.6

### Python Version

3.11

### Steps to Reproduce

1. Create a Device Type (Test Device Type)
2. Add a few Interfaces to the Device Type
3. Create a Device from the new Device Type (Test Device 1)
4. Create two more Devices from the Device Type (Test Device 2, Test Device 3)
5. Rename the interfaces on Test Device 3 for uniqueness
6. Create a Virtual Chassis from Test Devices 2 and 3 (start position 0, position 0 is VC master)
7. Go to the Test Device 1 Device, open the Interfaces tab
8. Click the 'add cable' icon for an interface
9. Select the VC Master (Test Device 2) for the B-side interface
10. Open the interface list

### Expected Behavior

The interface list should contain all VC member interfaces when selecting the VC master device for the B-side device.

### Observed Behavior

The interface list only shows interfaces directly associated with the VC master device.

DanSheps:
### Fixes: #19645 - Correct Interface selection for Cable add when VC master is the selected device

* Add new filter to `Interface` filterset: `virtual_chassis_member_or_master` which will only return interfaces in the full VC if the device is a master device
* Set cable creation form to utilize the new filter

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1888,12 +1888,12 @@ class InterfaceFilterSet(
     virtual_chassis_member_or_master = MultiValueCharFilter(
         method='filter_virtual_chassis_member_or_master',
         field_name='name',
-        label=_('Virtual Chassis Interfaces for Device')
+        label=_('Virtual Chassis Interfaces for Device when device is master')
     )
     virtual_chassis_member_or_master_id = MultiValueNumberFilter(
         method='filter_virtual_chassis_member_or_master',
         field_name='pk',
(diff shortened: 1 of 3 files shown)


jnovinger (changes requested):
We need tests for any new filterset filters.

jnovinger (approved):

PR #20046, opened 2025-08-07
Fixes: #19669 & #18396 - Allow Token Authentication against Media view

### LINKED ISSUE #18396: Get deviceTypes front_image and rear_image from the API

### LINKED ISSUE #19669: Media assets not reachable anymore through api requests since media requires Authentication

DanSheps:
### Fixes: #19669 & #18396 - Allow Token Authentication against Media view

* Add reusable `TokenConditionalLoginRequiredMixin` (inheriting from `ConditionalLoginRequiredMixin`)
* Change MediaView to utilize new Mixin

### DIFF
--- a/netbox/utilities/views.py
+++ b/netbox/utilities/views.py
@@ -6,7 +6,6 @@ from django.core.exceptions import ImproperlyConfigured
 from django.urls import reverse
 from django.urls.exceptions import NoReverseMatch
 from django.utils.translation import gettext_lazy as _
-from rest_framework.exceptions import AuthenticationFailed
 
 from netbox.api.authentication import TokenAuthentication
 from netbox.plugins import PluginConfig
@@ -47,12 +46,9 @@ class TokenConditionalLoginRequiredMixin(ConditionalLoginRequiredMixin):
         # Attempt to authenticate the user using a DRF token, if provided
         if settings.LOGIN_REQUIRED and not request.user.is_authenticated:
             authenticator = TokenAuthentication()
-            try:
-                auth_info = authenticator.authenticate(request)
-                if auth_info is not None:
-                    request.user = auth_info[0]  # User object
-            except AuthenticationFailed:
-                pass
+            auth_info = authenticator.authenticate(request)
+            if auth_info is not None:
+                request.user = auth_info[0]  # User object
 
         return super().dispatch(request, *args, **kwargs)
 
--- a/netbox/utilities/views.py
+++ b/netbox/utilities/views.py
@@ -6,7 +6,9 @@ from django.core.exceptions import ImproperlyConfigured
 from django.urls import reverse
 from django.urls.exceptions import NoReverseMatch
 from django.utils.translation import gettext_lazy as _
+from rest_framework.exceptions import AuthenticationFailed
 
+from netbox.api.authentication import TokenAuthentication
 from netbox.plugins import PluginConfig
 from netbox.registry import registry
 from utilities.relations import get_related_models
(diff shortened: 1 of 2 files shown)


netbox/utilities/views.py
```diff
@@ -39,6 +42,21 @@ def dispatch(self, request, *args, **kwargs):
...
+    def dispatch(self, request, *args, **kwargs):
+        # Attempt to authenticate the user using a DRF token, if provided
+        if settings.LOGIN_REQUIRED and not request.user.is_authenticated:
+            authenticator = TokenAuthentication()
+            try:
+                auth_info = authenticator.authenticate(request)
+                if auth_info is not None:
+                    request.user = auth_info[0]  # User object
+            except AuthenticationFailed:
+                pass
```
jnovinger: Couple of thoughts on this:

1. While this isn't a DRF view, the DRF convention is to set both `request.user` and `request.auth` (where `.auth` is the token instance) when using one of DRF's authentication mechanisms (which NetBox's `TokenAuthentication` is based on. I don't know if that's necessary here, but wanted to mention it, as convention. (FWIW, we also fail to set it elsewhere in NetBox).

2. Any particular reason we're silencing `AuthenticationFailed` exceptions? Looking at `TokenAuthentication.authenticate()`, those exceptions provide useful information. Not including it might lead to confusion for users trying to debug why the token they're using is not working.

DanSheps: > Any particular reason we're silencing AuthenticationFailed exceptions? Looking at TokenAuthentication.authenticate(), those exceptions provide useful information. Not including it might lead to confusion for users trying to debug why the token they're using is not working.

You raise a good point, I was thinking that because this chains into the next auth with the super().dispatch() which should raise the AuthenticationFailed as well if both Token and User Authentication fail.

I don't see a compelling need to do this though as it should still work as the user will be authenticated prior to this (typically).

netbox/utilities/views.py
```diff
@@ -39,6 +41,22 @@ def dispatch(self, request, *args, **kwargs):
...
+        # Attempt to authenticate the user using a DRF token, if provided
+        if settings.LOGIN_REQUIRED and not request.user.is_authenticated:
+            authenticator = TokenAuthentication()
+            auth_info = authenticator.authenticate(request)
+            if auth_info is not None:
+                request.user = auth_info[0]  # User object
+                try:
+                    request.auth = auth_info[1]
+                except IndexError:
+                    pass
```
jnovinger: If `auth_info` is not None, I think it should be safe to assume that the second item will be there. Granted we've over-ridden that method, but we shouldn't ever be changing the return signature.

But, now I'm beginning to wonder if you ran in to something unexpected that led to this ...

DanSheps: Did a double check, I think I did this out of habit.  Checking it, it should include the token information so like you said, it should be safe.

jnovinger (approved):

arthanson (approved):

Issue #20048, opened 2025-08-07
Cleanup get_viewname URL resolution and make extensible


arthanson:
### Proposed Changes

Create a new utility function alongside `get_viewname` tentatively called `get_action_url` that combines the functionality of get_viewname and reversing the URL.  Also allow this function to be overridden at the model level by plugins.

### Justification

`utilities.views.get_viewname` is used throughout the code doing basically the following:

```
viewname = get_viewname(instance, 'action')
url = reverse(viewname, kwargs={'pk': instance.pk})
```

In almost every case the URL is the only thing needed so this code can be made DRY by a combined function.  

In addition, this function requires the viewname to conform to the `'{app_label}:{model_name}_{action}` format with a single param being the object_id.  While this is valid for most plugins and models, there are cases where the plugin would need a different format (for example a URL with extra kwargs being passed in).  It would be fairly easy to make the URL resolution for this function be overridable at the model-class level, for example:

```
def get_action_url(model, action=None, rest_api=False, kwargs=None):
    if hasattr(model, '_get_action_url'):
        return model._get_action_url(action, rest_api, kwargs)

    return reverse(get_viewname(model, action, rest_api), kwargs=kwargs)
```

PR #20050, opened 2025-08-07
20048 cleanup get_viewname URL resolution

arthanson:
### Fixes: #20048 

adds a `get_action_url` to replace the `get_viewname` / reverse sequence in the app, and allows it to be replaceable by models that need a different URL pattern then the standard.

`get_viewname` is used in two filters `viewname` and `validated_viewname` which gets the viewname and then passes this to the url template tag so added in an `action_url` template tag (in netbox/utilities/templatetags/helpers.py) to use the get_action_url and have cleaner syntax goes from `{% url object|viewname:'list' %}` to `{% action_url object 'list' %}`.  As a template tag it also supports `as` syntax so allows the following replacement:

```
{% with viewname=qs.model|validated_viewname:"list" %}
<a href="{% url viewname %}
``` 

with cleaner:

```
{% action_url qs.model 'list' as list_url %}
<a href="{{ list_url }}
```

`netbox/extras/dashboard/widgets.py` needed the returned context changed as viewname isn't returned anymore but the template just uses it in an error message, so just updated the error message.

NetBoxFeatureSet still uses get_viewname in the get_absolute_url function, but that is fine as any model that is overriding _get_action_url would also override get_absoulte_url as well so it doesn't need to be changed there.

NetBoxAPIHyperlinkedIdentityField and NetBoxURLHyperlinkedIdentityField also use get_viewname and there isn't a good way around that, but if needed any plugin should be able to override those as needed.

### DIFF
--- a/netbox/extras/dashboard/widgets.py
+++ b/netbox/extras/dashboard/widgets.py
@@ -21,7 +21,7 @@ from utilities.permissions import get_permission_for_model
 from utilities.proxy import resolve_proxies
 from utilities.querydict import dict_to_querydict
 from utilities.templatetags.builtins.filters import render_markdown
-from utilities.views import get_viewname
+from utilities.views import get_action_url
 from .utils import register_widget
 
 __all__ = (
@@ -53,7 +53,7 @@ def object_list_widget_supports_model(model: Model) -> bool:
     """
     def can_resolve_model_list_view(model: Model) -> bool:
         try:
-            reverse(get_viewname(model, action='list'))
+            reverse(get_action_url(model, action='list'))
             return True
         except Exception:
             return False
@@ -206,7 +206,7 @@ class ObjectCountsWidget(DashboardWidget):
             permission = get_permission_for_model(model, 'view')
             if request.user.has_perm(permission):
                 try:
-                    url = reverse(get_viewname(model, 'list'))
+                    url = get_action_url(model, action='list')
                 except NoReverseMatch:
                     url = None
                 qs = model.objects.restrict(request.user, 'view')
@@ -275,15 +275,13 @@ class ObjectListWidget(DashboardWidget):
             logger.debug(f"Dashboard Widget model_class not found: {app_label}:{model_name}")
             return
 
-        viewname = get_viewname(model, action='list')
-
         # Evaluate user's permission. Note that this controls only whether the HTMX element is
         # embedded on the page: The view itself will also evaluate permissions separately.
         permission = get_permission_for_model(model, 'view')
         has_permission = request.user.has_perm(permission)
 
         try:
-            htmx_url = reverse(viewname)
+            htmx_url = get_action_url(model, action='list')
         except NoReverseMatch:
             htmx_url = None
         parameters = self.config.get('url_params') or {}
@@ -297,7 +295,7 @@ class ObjectListWidget(DashboardWidget):
             except ValueError:
                 pass
         return render_to_string(self.template_name, {
-            'viewname': viewname,
+            'model_name': model_name,
             'has_permission': has_permission,
             'htmx_url': htmx_url,
         })
--- a/netbox/extras/dashboard/widgets.py
+++ b/netbox/extras/dashboard/widgets.py
@@ -11,7 +11,7 @@ from django.conf import settings
 from django.core.cache import cache
 from django.db.models import Model
 from django.template.loader import render_to_string
-from django.urls import NoReverseMatch, resolve, reverse
+from django.urls import NoReverseMatch, resolve
 from django.utils.translation import gettext as _
 
 from core.models import ObjectType
@@ -53,9 +53,9 @@ def object_list_widget_supports_model(model: Model) -> bool:
     """
     def can_resolve_model_list_view(model: Model) -> bool:
         try:
-            reverse(get_action_url(model, action='list'))
+            get_action_url(model, action='list')
             return True
-        except Exception:
+        except NoReverseMatch:
             return False
 
     tests = [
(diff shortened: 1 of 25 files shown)


netbox/utilities/views.py
jeremystretch, line 1: Need to add `get_action_url` to `__all__`

netbox/utilities/views.py
```diff
@@ -282,6 +282,21 @@ def get_viewname(model, action=None, rest_api=False):
     return viewname
 
 
+def get_action_url(model, action=None, rest_api=False, kwargs=None):
+    """
+    Return the url for the given model and action, if valid.
```
jeremystretch:
```suggestion
    Return the URL for the given model and action, if valid; otherwise raise NoReverseMatch.
```

netbox/utilities/views.py
```diff
@@ -282,6 +282,21 @@ def get_viewname(model, action=None, rest_api=False):
...
+
+    :param model: The model or instance to which the view applies
+    :param action: A string indicating the desired action (if any); e.g. "add" or "list"
+    :param rest_api: A boolean indicating whether this is a REST API view
+    :param kwargs: A dictionary of keyword arguments for the view to include when registering its URL path (optional).
```
jeremystretch:
```suggestion
    :param model: The model or instance to which the URL belongs
    :param action: A string indicating the desired action (if any); e.g. "add" or "list"
    :param rest_api: A boolean indicating whether this is a REST API action
    :param kwargs: A dictionary of keyword arguments for the view to include when resolving its URL path (optional)
```

netbox/extras/dashboard/widgets.py
```diff
@@ -53,7 +53,7 @@ def object_list_widget_supports_model(model: Model) -> bool:
     """
     def can_resolve_model_list_view(model: Model) -> bool:
         try:
-            reverse(get_viewname(model, action='list'))
+            reverse(get_action_url(model, action='list'))
```
jeremystretch:
```suggestion
            get_action_url(model, action='list')
```

netbox/extras/dashboard/widgets.py
```diff
@@ -53,7 +53,7 @@ def object_list_widget_supports_model(model: Model) -> bool:
...
             return True
         except Exception:
```
jeremystretch: We should limit this to catching NoReverseMatch to avoid swallowing unanticipated exceptions.

```suggestion
        except NoReverseMatch:
```

netbox/templates/extras/dashboard/widgets/objectlist.html
```diff
@@ -7,6 +7,6 @@
   </div>
 {% else %}
   <div class="text-danger text-center">
-    <i class="mdi mdi-alert"></i> {% trans "Unable to load content. Invalid view name" %}: <span class="font-monospace">{{ viewname }}</span>
+    <i class="mdi mdi-alert"></i> {% trans "Unable to load content. Invalid list viewname for:" %} <span class="font-monospace">{{ model_name }}</span>
```
jeremystretch:
```suggestion
    <i class="mdi mdi-alert"></i> {% trans "Unable to load content. Could not resolve list URL for:" %} <span class="font-monospace">{{ model_name }}</span>
```

netbox/extras/tests/test_dashboard.py
```diff
@@ -45,4 +45,4 @@ def has_perm(self, *args, **kwargs):
         mock_request = Request()
         widget = ObjectListWidget(id='2829fd9b-5dee-4c9a-81f2-5bd84c350a27', **config)
         rendered = widget.render(mock_request)
-        self.assertTrue('Unable to load content. Invalid view name:' in rendered)
+        self.assertTrue('Unable to load content. Invalid list viewname for:' in rendered)
```
jeremystretch:
```suggestion
        self.assertTrue('Unable to load content. Could not resolve list URL for:' in rendered)
```

netbox/netbox/tables/columns.py
```diff
@@ -285,7 +285,7 @@ def render(self, record, table, **kwargs):
         for idx, (action, attrs) in enumerate(self.actions.items()):
             permission = get_permission_for_model(model, attrs.permission)
             if attrs.permission is None or user.has_perm(permission):
-                url = reverse(get_viewname(model, action), kwargs={'pk': record.pk})
+                url = get_action_url(record, action=action, kwargs={'pk': record.pk})
```
jeremystretch: I'd keep this as passing `model` unless there's a need to change it.

```suggestion
                url = get_action_url(model, action=action, kwargs={'pk': record.pk})
```

netbox/utilities/templatetags/helpers.py
```diff
@@ -63,6 +64,125 @@ def validated_viewname(model, action):
         return None
 
 
+@register.tag
+def action_url(parser, token):
```
jeremystretch, line 122: Need to add `action_url` to `__all__`

netbox/utilities/templatetags/helpers.py
```diff
@@ -63,6 +64,125 @@ def validated_viewname(model, action):
...
+
+        or
+
+        {% action_url model "action_name" pk=object.pk %}
+
+        or
+
+        {% action_url model "action_name" pk=object.pk as variable_name %}
+
+    The first argument is a model instance. The second argument is the action name.
```
jeremystretch:
```suggestion
    The first argument is a model or instance. The second argument is the action name.
```

netbox/utilities/templatetags/helpers.py
```diff
@@ -63,6 +64,125 @@ def validated_viewname(model, action):
...
+    You can also pass additional parameters::
+
+        {% action_url device "journal" pk=device.pk %}
+
+    Or assign the URL to a variable::
+
+        {% action_url device "edit" as edit_url %}
+    """
+
+    class ActionURLNode(template.Node):
```
jeremystretch: Can we move this outside of `action_url()`?

netbox/utilities/views.py
jeremystretch: We should also call out that the function will defer to `_get_action_url()` on the model if it exists.

jeremystretch (changes requested):

jeremystretch (approved):

PR #20058, opened 2025-08-08
Closes #19377: Introduce config context profiles

### LINKED ISSUE #19377: Add JSON schema validation for Context Data
### NetBox version

v4.3.0

### Feature type

New functionality

### Proposed functionality

Similar to the proposal outline in https://github.com/netbox-community/netbox/issues/19025 it would be really useful if a user could specify a json schema along with a [config context](https://netboxlabs.com/docs/netbox/en/stable/features/context-data/) entry. 

### Use case

The structure of config context can be crucial for consuming systems (e.g. automated provisioning). We currently implement JSON schema validation with [custom validation](https://netboxlabs.com/docs/netbox/en/stable/customization/custom-validation/). However, it is always difficult to identify which config context a certain validation should be applied to.

Being able to assign a schema directly to a config context entry would remove that complexity.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #19377

- Introduce the ConfigContextProfile model to hold JSON schema definitions for config contexts
- Add a `profile` ForeignKey on ConfigContext
- Tweak `validate_schema()` to not raise an error on empty values
- Tweak ModuleTypeProfile to use `validate_schema()` as a validator on its `schema` field
- Introduce SyncedDataMixin for GraphQL types for models which support data syncing


netbox/extras/models/configs.py
```diff
@@ -24,6 +28,39 @@
...
+
+    clone_fields = ('schema',)
+
+    class Meta:
+        ordering = ('name',)
+        verbose_name = _('config context profile')
+        verbose_name_plural = _('config context profiles')
+
+    def __str__(self):
+        return self.name
```
arthanson, line 61: I can add a profile without any data, I think a schema or a data file should be required.
also, you can give both a schema and a data-file and it's not clear which one will take precedence - I would think you would allow one or the other but not both.

netbox/extras/forms/filtersets.py
```diff
@@ -354,16 +355,29 @@ class TagFilterForm(SavedFiltersMixin, FilterForm):
     )
 
 
+class ConfigContextProfileFilterForm(SavedFiltersMixin, FilterForm):
```
arthanson, line 358: The "Data" section with Data source and Data file should appear in the filter form.

netbox/extras/tables/tables.py
```diff
@@ -546,7 +547,28 @@ class Meta(NetBoxTable.Meta):
...
+        linkify=True
+    )
+    tags = columns.TagColumn(
+        url_name='extras:configcontextprofile_list'
+    )
+
+    class Meta(NetBoxTable.Meta):
+        model = ConfigContextProfile
+        fields = (
+            'pk', 'id', 'name', 'description', 'comments', 'tags', 'created', 'last_updated',
```
arthanson: Data File, Data source, 'Date Synced should all be in the available columns.

netbox/extras/models/configs.py
```diff
@@ -147,6 +190,13 @@ def clean(self):
                 {'data': _('JSON data must be in object form. Example:') + ' {"foo": 123}'}
             )
 
+        # Validate config data against the assigned profile's schema (if any)
```
arthanson, line 200: I think this will only validate against the schema field, not if there is a data-source and data-file?  If I add a data file and click save I get the following error:

```
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/core/handlers/base.py", line 197, in _get_response
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/views/generic/base.py", line 105, in view
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/views/generic/object_views.py", line 192, in dispatch
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/views/generic/base.py", line 26, in dispatch
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/netbox/utilities/views.py", line 125, in dispatch
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/netbox/utilities/views.py", line 39, in dispatch
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/views/generic/base.py", line 144, in dispatch
    return handler(request, *args, **kwargs)
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/views/generic/object_views.py", line 288, in post
    if form.is_valid():
       ~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/forms/forms.py", line 206, in is_valid
                                 ^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/forms/forms.py", line 201, in errors
    self.full_clean()
    ~~~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/forms/forms.py", line 339, in full_clean
    self._post_clean()
    ~~~~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/forms/base.py", line 78, in _post_clean
    return super()._post_clean()
           ~~~~~~~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/forms/models.py", line 498, in _post_clean
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahanson/dev/work/netbox/venv/lib/python3.13/site-packages/django/db/models/base.py", line 1654, in full_clean
    self.clean()
    ~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/models/features.py", line 546, in clean
    self.sync()
    ~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/models/features.py", line 611, in sync
    self.sync_data()
    ~~~~~~~~~~~~~~^^
  File "/Users/ahanson/dev/work/netbox/netbox/netbox/models/features.py", line 622, in sync_data
    raise NotImplementedError(_("{class_name} must implement a sync_data() method.").format(
        class_name=self.__class__
    ))
NotImplementedError: <class 'extras.models.configs.ConfigContextProfile'> must implement a sync_data() method.

```

arthanson (changes requested):
Not sure if Data Source and Data file should be there - but they don't look like the completely work currently:
1. you can have both schema and data-source which I think should be one or the other.
2. adding data-file and saving will cause error
3. The UI should probably be a dropdown for the type and allow one or the other to be entered - this is different then the current Config Context entry and Config Templates - but those should probably be changed as well? (maybe separate PR)
4. data source / data file are missing from filters, tables, detail view
5. The validation I think only works against schema, not against data file.

It also doesn't validate the reverse way, i.e. if you have a profile assigned to a context then change the profile so the validation will fail on the context - Not sure if we want to put up a warning or not?

netbox/extras/models/configs.py
jeremystretch: > I think a schema or a data file should be required.

This was my original intent, but I now think there's value in assigning profiles solely from an organizational perspective, without requiring a schema definition.

netbox/extras/models/configs.py
jeremystretch: ConfigContextProfile was missing a `sync_data()` method to copy the DataFile content to its `schema` field.

netbox/extras/models/configs.py
jeremystretch: > also, you can give both a schema and a data-file and it's not clear which one will take precedence - I would think you would allow one or the other but not both.

The `sync_data()` method was missing; this is responsible for copying the DataFile content to the `schema` field. (We're following the same pattern here as with ConfigContext and ExportTemplate.)

jeremystretch:
> It also doesn't validate the reverse way, i.e. if you have a profile assigned to a context then change the profile so the validation will fail on the context - Not sure if we want to put up a warning or not?

This is acceptable IMO. The schema was validated against the current profile at the time it was last modified; we can't reasonably ensure anything beyond that.

arthanson (approved):

PR #20072, opened 2025-08-11
Fixes #20012: Fix support for `empty` filter for custom fields

### LINKED ISSUE #20012: Custom fields of type 'object' don't work with 'empty' filter expression
### Deployment Type

Self-hosted

### NetBox Version

v4.3.2

### Python Version

3.12

### Steps to Reproduce

1. Create a custom field called 'location_served' with type 'object (location)' and filter logic 'exact' and object types 'device'. (We use this when a device serves a room other than where it's physically located, for example an AV Closet next to a screening room.)
2. Add some values to this custom field on a few devices.
3. Try to filter `/dcim/devices/?cf_location_served__empty=true`

### Expected Behavior

Since the custom field stores a numeric value, representing the ID of the linked object, it should respect all the numeric lookup expressions.  Some expressions do work correctly, for example `/dcim/devices/?cf_location_served__gt=18600`.

### Observed Behavior

It returns all objects, as if the filter were not present.

jeremystretch:
### Fixes: #20012

- Introduce `JSONEmpty` specifically for JSON fields
- Ensure that a `BooleanFilter` is used for the `empty` lookup expression

### DIFF
--- a/netbox/netbox/filtersets.py
+++ b/netbox/netbox/filtersets.py
@@ -29,6 +29,13 @@ __all__ = (
     'OrganizationalModelFilterSet',
 )
 
+STANDARD_LOOKUPS = (
+    'exact',
+    'iexact',
+    'in',
+    'contains',
+)
+
 
 #
 # FilterSets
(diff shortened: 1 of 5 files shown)


netbox/netbox/filtersets.py
```diff
@@ -29,6 +29,13 @@
     'OrganizationalModelFilterSet',
 )
 
+STANDARD_LOOKUPS = (
+    'exact',
+    'iexact',
+    'in',
+    'contains',
+)
```
jnovinger, line 37: I was originally going to suggest that maybe this should be a set instead of a tuple, since it only gets instantiated once but accessed a lot later (on the order of ~2500 times within 10 list view requests). However, some quick measurements showed that with only 4 items, the tuple is actually a little faster for lookup than the set.

jnovinger (approved):

PR #20080, opened 2025-08-12
Closes #18984: Add status field to Rack model

### LINKED ISSUE #18984: Rack reservations - add status field
### NetBox version

v4.2.5

### Feature type

Data model extension

### Proposed functionality

add status field choice to rack reservations. (allow customisation of this choices list like many other status fields have)

internally in our org we would have uses for the following statuses:
- active
- pending
- staged
- proposed
- hold
- cancelled

### Use case

it would be great if reservations had native status fields.

being able to track active vs proposed vs cancelled colocation reservations would be very useful

### Database changes

add status field to rack reservations model

### External dependencies

_No response_

jeremystretch:
### Closes: #18984

- Add a `status` ChoiceField and `get_status_color()` method to dcim.Rack
- Clean up the `rackreservation.html` template

### DIFF
--- a/docs/models/dcim/rackreservation.md
+++ b/docs/models/dcim/rackreservation.md
@@ -12,6 +12,13 @@ The [rack](./rack.md) being reserved.
 
 The rack unit or units being reserved. Multiple units can be expressed using commas and/or hyphens. For example, `1,3,5-7` specifies units 1, 3, 5, 6, and 7.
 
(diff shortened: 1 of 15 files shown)


jnovinger (approved):
Note to self: `dcim.RackReserveration.status+` is not a valid `FIELD_CHOICES` key. 🤦

PR #20082, opened 2025-08-12
Fixes #20042: Support Bridge-Interfaces in Interface-Template Import

m-hau:
### Fixes: #20042

When importing device- or module-types, allow their interface-templates to contain the `bridge` attribute and properly save it.

<hr>

The unexpectedly hardest part was finding the right place to hook into the existing process, which in this cases required to create an entirely new hook-point. While developing and testing I also found some other minor bugs in the generalized import validation, and attempted to fix them. Please see the individual commit messages for more information.

### DIFF
--- a/netbox/dcim/forms/object_import.py
+++ b/netbox/dcim/forms/object_import.py
@@ -84,6 +84,12 @@ class InterfaceTemplateImportForm(forms.ModelForm):
         label=_('Type'),
         choices=InterfaceTypeChoices.CHOICES
     )
+    bridge = forms.ModelChoiceField(
+        label=_('Bridge'),
+        queryset=InterfaceTemplate.objects.all(),
+        to_field_name='name',
+        required=False
+    )
     poe_mode = forms.ChoiceField(
         choices=InterfacePoEModeChoices,
         required=False,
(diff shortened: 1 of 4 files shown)


m-hau:
This is a first draft to show the general idea and implementation, not ready to be merged, with a few parts missing and questions/TODOs still open. I would kindly request some first feedback from the more knowledgeable and codebase-familiar people around here, if this is a valid and acceptable way of doing this.

jeremystretch (commented):
@m-hau while it's clear you've invested a lot of effort in this, I'm afraid the complexity of the proposed implementation is disproportionate to the value of the functionality it provides. Have you considered any alternative approaches which don't require such extensive modifications to the existing code?

m-hau:
I am ignoring the list/dict validation fixes here (the first two commits), since those are general fixes that should be applied anyway. Would it be good to extract them into their own issue/PR?

The simplest solution would be to just add the field to the form (what commit 3 does). This generally allows that field to be imported, but since there is no sorting happening the interfaces in the input data already need to be properly sorted, such that bridge interfaces come before their bridged interfaces, which may be acceptable from a UX perspective.
The biggest gripe I have is the error if violated will be `interfaces[x] bridge: Select a valid choice. That choice is not one of the available choices.`, which is not wrong but also not very specific that the cause may simply be a sorting issue. To fix that you would still need the new hook to validate the interface list as a whole, including checking their relations.

The next step up would be the new hook to sort the list, ensuring a simple sort order like "first all non-bridged-interfaces, then the rest". This works for all non-recursive cases (where a bridge interface is never also a bridged interface), and should also work for simple cases if in future "parent interface" relations are also importable. If validation isn't explicitly handled, it suffers from the same unspecific validation error message as above.

And the imho "correct" solution is the full blown implementation I did using a dependency graph, which covers all current (and future) cases and allows specific and actionable user feedback.

EDIT: There is also the variant of splitting the interface creation into two phases. First create all interfaces but without any relations (bridge, parent, ...), and in phase two add all the relations. In that case the sort order is irrelevant, but requires a special handling/implementation for interface imports. It would also allow for circular dependencies, which may or may not be desired.

m-hau:
@jeremystretch Apologies for the ping, just a quick reminder that I am still waiting for feedback on how to proceed further.

jeremystretch:
@m-hau thanks for your work on this. While I certainly appreciate all the effort you've put in here, after conferring with the other maintainers we're not comfortable accepting responsibility for such complex logic given its relatively small impact in terms of functionality delivered. Unfortunately we have to decline your proposed changes. It's likely that FR #20042 simply won't be implemented due to the technical challenges involved.

PR #20090, opened 2025-08-13
20089 use get_queryset function for valid_models

arthanson:

### DIFF
--- a/netbox/core/api/urls.py
+++ b/netbox/core/api/urls.py
@@ -9,7 +9,7 @@ router.APIRootView = views.CoreRootView
 router.register('data-sources', views.DataSourceViewSet)
 router.register('data-files', views.DataFileViewSet)
 router.register('jobs', views.JobViewSet)
-router.register('object-changes', views.ObjectChangeViewSet)
+router.register('object-changes', views.ObjectChangeViewSet, basename='objectchange')
 router.register('object-types', views.ObjectTypeViewSet)
 router.register('background-queues', views.BackgroundQueueViewSet, basename='rqqueue')
 router.register('background-workers', views.BackgroundWorkerViewSet, basename='rqworker')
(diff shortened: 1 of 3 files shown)


netbox/core/api/urls.py
```diff
@@ -9,7 +9,7 @@
 router.register('data-sources', views.DataSourceViewSet)
 router.register('data-files', views.DataFileViewSet)
 router.register('jobs', views.JobViewSet)
-router.register('object-changes', views.ObjectChangeViewSet)
+router.register('object-changes', views.ObjectChangeViewSet, basename='objectchange')
```
jeremystretch, line 12: Why set `basename`?

jeremystretch (commented):
Rather than changing the views, can we address the root issue in the `valid_models()` method? Seems like it would be possible by removing the forced resolution of the ContentType queryset.

netbox/core/api/urls.py
arthanson: When you remove queryset (change to function) from API view it throws an error unless basename is set.

arthanson: I looked at trying to move to Q objects, like the following, however the issue is get_models returns a dict and going like this still causes the issue where the content types are only resolved at instantiation.  As discussed an alternative solution might be to put it into the housekeeping task or such but that would be a breaking change:

```
    def valid_models(self):
        # Exclude any change records which refer to an instance of a model that's no longer installed. This
        # can happen when a plugin is removed but its data remains in the database, for example.
        from django.db.models import Q
        
        # Get all currently registered models at query time
        current_models = apps.get_models()
        
        # Build a Q object that matches any ContentType that corresponds to a currently registered model
        valid_content_type_conditions = Q()
        for model in current_models:
            valid_content_type_conditions |= Q(
                changed_object_type__app_label=model._meta.app_label,
                changed_object_type__model=model._meta.model_name
            )
        
        return self.filter(valid_content_type_conditions)
```

jeremystretch (approved):

PR #20097, opened 2025-08-14
Closes #20013: Syntax highlighting on Data Files

### LINKED ISSUE #20013: Syntax highlighting for data files
### NetBox version

v4.3.5

### Feature type

New functionality

### Proposed functionality

Syntax highlighting on the data files details page.

### Use case

Better readability of the content from data files.

### Database changes

- None

### External dependencies

[Python Pygments](https://pygments.org/#) as a direct dependency. It's already installed as a dependency of `mkdocs-material`.

tobiasge:
### Fixes: #20013

Add syntax highlighting on the details page for Data Files.


tobiasge:
Example using https://github.com/netbox-community/customizations/blob/master/reports/dcim-reports/CheckConsoleOOBPower.py 
dark
light

jnovinger:
Thanks for the contribution, @tobiasge.

While this is technically a new feature, it's really more of a UI enhancement that improves the presentation of existing functionality rather than adding new capabilities to NetBox. Changes like this are appropriate for patch releases and should target `main` rather than the `feature` branch, which allows us to get the changes to users sooner.

Also, while we don't have thorough tests for template tags currently, it's something we would like to improve. Please add some basic tests, something like:

```py
# core/tests/test_templatetags.py
from django.test import TestCase
from core.templatetags.highlight_code import highlight_code

class HighlightCodeTestCase(TestCase):
    def test_python_highlighting(self):
        # Test that Python code gets highlighted with pygments classes

    def test_unknown_extension_fallback(self):
        # Test fallback to plain <pre> tags for unknown file types

    def test_empty_content(self):
        # Test handling of empty strings and None values
```

Please add tests and then resubmit this PR targeting `main`.

tobiasge:
@jnovinger No need to close, you can just change the base branch in this PR; I have rebased in on the current main.
I will add some test tomorrow.

PR #20099, opened 2025-08-14
Fixes: #20098 - Handle empty object_types field in Tag bulk import

### LINKED ISSUE #20098: AttributeError during Tag Bulk Import
### Deployment Type

Self-hosted

### NetBox Version

v4.3.6

### Python Version

3.12

### Steps to Reproduce

1. Open "Tag Bulk Import" (`/extras/tags/import/`)
2. Paste data

Example data:
```csv
name,slug,color
test,test,00ff00
```

3. Submit import

### Expected Behavior

The tag should be imported.

### Observed Behavior

```python
<class 'AttributeError'>

'NoneType' object has no attribute 'app_label'

Python version: 3.12.3
NetBox version: 4.3.6-Docker-3.3.0
Plugins: None installed
```

bctiemann:
### Fixes: #20098

Handles the case in `CSVMultipleContentTypeField.prepare_value` where `value` is null (which only occurs in Tag bulk import, where the `object_types` field is optional).

### DIFF
--- a/netbox/utilities/forms/fields/csv.py
+++ b/netbox/utilities/forms/fields/csv.py
@@ -120,4 +120,6 @@ class CSVMultipleContentTypeField(forms.ModelMultipleChoiceField):
                 app_label, model = name.split('.')
                 ct_filter |= Q(app_label=app_label, model=model)
             return list(ContentType.objects.filter(ct_filter).values_list('pk', flat=True))
-        return object_type_identifier(value)
+        if value:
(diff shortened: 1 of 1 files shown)


arthanson (changes requested):
I think it would be cleaner/clearer right at the top of the function to have:

```
if not value:
    return None
```
as none of the other processing needs to be done.

arthanson (approved):

PR #20111, opened 2025-08-16
Closes #20013: Syntax highlighting on Data Files

### LINKED ISSUE #20013: Syntax highlighting for data files

tobiasge:
Fixes: https://github.com/netbox-community/netbox/issues/20013

Add syntax highlighting on the details page for Data Files.


netbox/core/templatetags/highlight_code.py
```diff
@@ -0,0 +1,30 @@
...
+        lexer = get_lexer_for_filename(filename)
+    except ClassNotFound:
+        return mark_safe(f'<pre>{value}</pre>')  # Fallback to plain text if no lexer was found
+    return mark_safe(
+        highlight(
+            value,
+            lexer,
+            HtmlFormatter(linenos='inline', classprefix='pygments-', style='solarized-light'),
+        )
+    )
```
jnovinger, line 31: In the fallback cases of `highlight_code()`, we need to ensure that we are escaping `value` before marking it safe, to avoid XSS vulnerabilities. Previously, this was handled by Django's auto-escaping behavior.

I think something like this is the best approach:

```py
from django.utils.html import escape

@register.simple_tag
def highlight_code(value, filename: str):
    ...

	if not filename:
    	return mark_safe(f'<pre>{escape(value)}</pre>')
```

netbox/core/templatetags/highlight_code.py
```diff
@@ -0,0 +1,30 @@
...
+    try:
+        lexer = get_lexer_for_filename(filename)
+    except ClassNotFound:
+        return mark_safe(f'<pre>{value}</pre>')  # Fallback to plain text if no lexer was found
+    return mark_safe(
+        highlight(
+            value,
+            lexer,
+            HtmlFormatter(linenos='inline', classprefix='pygments-', style='solarized-light'),
+        )
```
jnovinger, line 30: My testing indicates that the `style='solarized-light'` argument is not needed. Am I maybe missing something?

netbox/core/tests/test_templatetags.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertTrue(FAKE_PLAIN_TEXT_CONTENT in result)
+
+    def test_empty_content(self):
+        result = highlight_code('', 'FAKE_PLAIN_TEXT_NAME')
+        self.assertTrue(result.startswith('<pre></pre>'))
+        self.assertTrue(len(result) == 11)
+
+        result = highlight_code(None, 'FAKE_PLAIN_TEXT_NAME')
+        self.assertTrue(result.startswith('<pre></pre>'))
+        self.assertTrue(len(result) == 11)
```
jnovinger: If we're going to assert the length of the result, then why do we need to assert against `.startswith()`? Wouldn't it be easier to just replace both asserts with one that just asserts equality with the expected output?

netbox/core/tests/test_templatetags.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def test_empty_filename(self):
+        result = highlight_code(' ', '')
+        self.assertTrue(result.startswith('<pre> </pre>'))
+        self.assertTrue(len(result) == 12)
+
+        result = highlight_code(' ', None)
+        self.assertTrue(result.startswith('<pre> </pre>'))
+        self.assertTrue(len(result) == 12)
```
jnovinger: In addition to the note above about escaping data file values in fall-back cases in the template tag, once that is added, we should also add tests to ensure that it works as expected.

Something like:

```py
def test_html_escaping_in_fallbacks(self):
    malicious_content = '<script>alert("XSS")</script><b>Bold</b>'

    # Test empty filename fallback
    ...

    # Test unknown extension fallback
    ...
```

jnovinger (changes requested):

netbox/core/templatetags/highlight_code.py
tobiasge: The different themes use different class names for the styling. `solarized-light` and `solarized-dark` use the same names. This ensures that the switch from dark to light can be done without a reload of the page.

jnovinger: Interesting, I was unaware of that. However, when I run this without the `style` param, I'm not able to observe what you describe--I can easily switch from light to dark and back without reloading the page.

Can you point me to the docs that describe this behavior? Thanks!

tobiasge: You can run this snippet with the different styles from here https://pygments.org/styles/:
```python
from pygments.formatters import HtmlFormatter
h = HtmlFormatter(linenos="inline", classprefix="pygments-", style="solarized-dark")
print(h.get_style_defs())
```

There you see, that not all styles use the same classes. I used this method to have a good starting point for the style. If more customization is wanted, it would be necessary to create a `NetboxStyle`.

jnovinger: Thanks for the explanation, @tobiasge. I think we might be talking about two different things here.

You're absolutely right that different Pygments styles generate different CSS class names. However, my point was more specific: looking at the CSS styles you've generated and included in this PR, the `style='solarized-light'` parameter doesn't appear to change the actual visual result.

When I test this locally with your generated styles - both with and without the style parameter in the `HTMLFormatter` instance used in the template tag - the syntax highlighting looks identical and theme switching works the same way. This suggests that the CSS you've already generated covers the highlighting regardless of which specific Pygments style classes the HTMLFormatter produces.

Could you help me understand what specific visual difference you're seeing when you include vs exclude the style parameter? That would help us better understand whether this parameter is actually needed given the styles you've already committed in this PR.

tobiasge: Without the parameter Pygments uses it's default style. That the default style and `solarized-*` are using the same class names is more or less random. Because I created the style in the CSS files from `solarized-*` I added the parameter, to force Pygements to use the known classes.
This is only needed because the dark and light mode switching in NetBox is build in JS rather than a page reload. If it was a page reload we could generate the styles directly from Pygments while rendering the HTML.

jnovinger:
Thank you for the work on this feature. While the syntax highlighting functionality is valuable, the implementation has become more complex than originally anticipated, requiring security fixes and extended technical discussion. I have decided to close this PR for now. The feature request remains open for future consideration with a potentially different approach.

tobiasge:
I don't understand why having a technical discussion about the PR and a fixed security issue are a reason to close it.

PR #20128, opened 2025-08-19
add optional return_url parameter to "Add" button for missing prerequisites

### LINKED ISSUE #19990: add return_url to Add button for missing prerequisite.
### NetBox version

v4.3.4

### Feature type

New functionality

### Proposed functionality

Setting a return_url to the current page when adding a new model through the Add button in the missing prerequisite dialog box would easily streamline the workflow filling an empty database, without having to navigate through the menu after adding each missing prerequisite model.

Image

### Use case

Remove unnecessary menu navigation when trying to add a model with missing prerequisites. For example, adding a device on a fresh, empty database requires me to add a device role, device type and site. If I don't remember each prerequisite for the object I want to add, I have to navigate back to the Add Device page after creating each object. This is even more annoying when the prerequisites have prerequisites on their own.

### Database changes

_No response_

### External dependencies

_No response_

9er-tom:
### Fixes: #19990

add optional return_url parameter to "Add" button, used in missing_prerequisites.html to return user back to initial page after adding an prerequisite object

### DIFF
--- a/netbox/utilities/templates/buttons/add.html
+++ b/netbox/utilities/templates/buttons/add.html
@@ -1,10 +1,6 @@
 {% if url %}
     {% load i18n %}
-    {% if return_url %}
-        <a href="{{ url }}?return_url={{ return_url }}" type="button" class="btn btn-primary">
-    {% else %}
-        <a href="{{ url }}" type="button" class="btn btn-primary">
-    {% endif %}
-<i class="mdi mdi-plus-thick"></i> {% trans "Add" %}
-</a>
+    <a href="{{ url }}{% if return_url %}?return_url={{ return_url }}{% endif %}" type="button" class="btn btn-primary">
+        <i class="mdi mdi-plus-thick"></i> {% trans "Add" %}
+    </a>
 {% endif %}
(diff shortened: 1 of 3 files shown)


netbox/utilities/templates/buttons/add.html
```diff
@@ -1,6 +1,10 @@
...
-{% endif %}
+    {% load i18n %}
+    {% if return_url %}
+        <a href="{{ url }}?return_url={{ return_url }}" type="button" class="btn btn-primary">
+    {% else %}
+        <a href="{{ url }}" type="button" class="btn btn-primary">
+    {% endif %}
+<i class="mdi mdi-plus-thick"></i> {% trans "Add" %}
+</a>
+{% endif %}
```
jnovinger: This can be simplified to avoid repeating ourselves:
```suggestion
{% if url %}
{% load i18n %}
  <a href="{{ url }}{% if return_url %}?return_url={{ return_url }}{% endif %}" type="button" class="btn btn-primary">
    <i class="mdi mdi-plus-thick"></i> {% trans "Add" %}
  </a>
{% endif %}
```

jnovinger (changes requested):
One suggestion on the change to `add.html`.

netbox/utilities/templates/buttons/add.html
9er-tom: ah of course, makes sense. django templates still feel a bit strange to me.
i commited the fix

jnovinger (approved):
Thanks @9er-tom !

PR #20130, opened 2025-08-19
Closes #20129: Enable dynamic model feature registration

### LINKED ISSUE #20129: Enable plugins to register custom model features
### NetBox version

v4.4.0-beta1

### Feature type

New functionality

### Proposed functionality

NetBox support various _model features_ such as tags, custom fields, change logging, etc. which enhance the application's functionality on a per-model basis. These features are enabled for models by inheriting a designated mixin class: For example, tagging is enabled for all models which inherit the `netbox.models.features.TagsMixin` class. These feature-to-class mappings are defined statically in the [`FEATURES_MAP`](https://github.com/netbox-community/netbox/blob/9580ac2946e4b702e249c37aaf9ec81a85229e19/netbox/netbox/models/features.py#L618) constant:

```python
FEATURES_MAP = {
    'bookmarks': BookmarksMixin,
    'change_logging': ChangeLoggingMixin,
    'cloning': CloningMixin,
    ...
}
```

This static mapping can be replaced with a dynamic registration system which can be employed by both core NetBox apps and plugins to register model features. For example, the tagging feature mentioned above would be registered like this:

```python
@register_model_feature('tags')
def support_tagging(model):
    return issubclass(model, TagsMixin)
```

This approach allows for complex logic within the litmus function bound to the declared feature, however in many cases it could be simplified to a lambda invoking `issubclass()` directly:

```python
register_model_feature('tags', lambda model: issubclass(model, TagsMixin))
```

The result is that `registry['model_features']` will now hold a mapping of feature names to evaluation functions, rather than to mixin classes.

### Use case

This approach provides two benefits over the current implementation:

1. Plugins can introduce new features and apply them to core models.
2. The logic for determining whether a feature is supported for a given model can be extended beyond simple inheritance checking.

For example, the [netbox-branching plugin](https://github.com/netboxlabs/netbox-branching) would register `branching` as a model feature, using a function it provides to determine eligibility on a per-model basis.

While this has previously been possible by modifying the `model_features` registry directly, FR #19924 in NetBox v4.4 now records model features on ObjectType records in the database, so registry hacking is no longer feasible.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #20129

- Introduce the `register_model_feature()` function for registering individual model features by name
- Replace `FEATURES_MAP` with calls to `register_model_feature()`
- Update development documentation

### DIFF
--- a/docs/plugins/development/models.md
+++ b/docs/plugins/development/models.md
@@ -124,7 +124,7 @@ For more information about database migrations, see the [Django documentation](h
 
 ## Custom Model Features
 
-In addition to utilizing the model features provided natively by NetBox (listed above), plugins can register their own model features. This is done using the `register_model_feature()` function from `netbox.models.features`. This function takes two arguments: a feature name, and a callable which accepts a model class. The callable must return a boolean value indicting whether the given model supports the named feature.
+In addition to utilizing the model features provided natively by NetBox (listed above), plugins can register their own model features. This is done using the `register_model_feature()` function from `netbox.utils`. This function takes two arguments: a feature name, and a callable which accepts a model class. The callable must return a boolean value indicting whether the given model supports the named feature.
 
 This function can be used as a decorator:
 
--- a/docs/plugins/development/models.md
+++ b/docs/plugins/development/models.md
@@ -24,20 +24,7 @@ Every model includes by default a numeric primary key. This value is generated a
 
 ## Enabling NetBox Features
 
-Plugin models can leverage certain NetBox features by inheriting from NetBox's `NetBoxModel` class. This class extends the plugin model to enable features unique to NetBox, including:
-
-* Bookmarks
-* Change logging
-* Cloning
-* Custom fields
-* Custom links
-* Custom validation
-* Export templates
-* Journaling
-* Tags
-* Webhooks
-
-This class performs two crucial functions:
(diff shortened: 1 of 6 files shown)


docs/plugins/development/models.md
```diff
@@ -135,6 +122,27 @@ For more information about database migrations, see the [Django documentation](h
 
 ::: netbox.models.features.TagsMixin
 
+## Custom Model Features
+
+In addition to utilizing the model features provided natively by NetBox (listed above), plugins can register their own model features. This is done using the `register_model_feature()` function from `netbox.models.features`. This function takes two arguments: a feature name, and a callable which accepts a model class. The callable must return a boolean value indicting whether the given model supports the named feature.
```
jnovinger: Is `netbox.models.features` the correct module to import from, given that `register_model_feature()` is actually defined in `netbox.utils`?

It would work as is, since it's also imported in `netbox.models.features`, just wanted to clarify.

jnovinger (approved):
Looks great. I had one little question about wording in the updated docs, but that's it.

docs/plugins/development/models.md
jeremystretch: Good catch, thanks!

PR #20158, opened 2025-08-22
Closes #18147: Include device & VM interfaces in VRF related objects

### LINKED ISSUE #18147: Add Interfaces to related objects in VRF view.
### NetBox version

v4.1.7

### Feature type

Data model extension

### Triage priority

N/A

### Proposed functionality

When a user clicks on a VRF object they are met with the following list of related objects:
image

I ask that an additional row be included for related interfaces. When clicked, this would open a filtered search for that VRF on the interface search page (netbox/dcim/interfaces/?vrf_id=XXXX).

### Use case

This is useful to track interface assigment to different VRFs straight from the VRF screen, without having to go to the interface search and manually filter for the VRF. 

### Database changes

Not sure.

### External dependencies

No

jeremystretch:
### Fixes: #18147

- Introduce the RelatedObjectCount dataclass to accept an optional label
- Update the `related_objects.html` inclusion template to use RelatedObjectCount
- Extend `VRFView.get_extra_context()` to include device & VM interface counts
- Tweak the signature of `get_related_models()` to avoid passing mutable values for `omit` and `extra`

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -53,8 +53,26 @@ class VRFView(GetRelatedModelsMixin, generic.ObjectView):
         )
         export_targets_table.configure(request)
 
+        related_models = self.get_related_models(
+            request,
+            instance,
+            omit=(Interface, VMInterface),
+            extra=(
+                (
+                    Interface.objects.restrict(request.user, 'view').filter(vrf=instance),
+                    'vrf_id',
+                    _('Device Interfaces')
+                ),
+                (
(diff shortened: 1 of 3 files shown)


arthanson (changes requested):
The Devices Interfaces shows up but if you click on it the resulting list view is not correctly filtered, this is actually true for all the Related Objects - I had prefixes assigned and clicking on that will go to a list of all prefixes.  Switching back to main branch and the filtering works correctly.

arthanson (approved):

PR #20171, opened 2025-08-25
Fixes #20164 - Add status field to PowerOutlet bulk create form

### LINKED ISSUE #20164: Adding Power outlets in bulk
### Deployment Type

Self-hosted

### NetBox Version

v4.3.6

### Python Version

3.12

### Steps to Reproduce

1. Select multiple devices when you are on the Devices page
2. Click Add Components, then Power Outlets
3. Fill in required field Name with [1-9]
4. Click Create

Image

### Expected Behavior

9 New power outlets should have been created for each device that was selected

### Observed Behavior

See attached screenshot
It looks like the Status field is missing when doing a bulk create for power outlets

Image

pheus:
### Fixes: #20164

Includes the `status` field in the PowerOutlet bulk create form to allow configuration during bulk creation.

### Summary
- Add `status` to the DCIM PowerOutlet bulk create form.
- Brings bulk create in line with single-object create/edit behavior.

### Validation
1) On **Devices**, select multiple devices.
2) Click **Add Components → Power Outlets**.
3) Enter a Name pattern (e.g., `[1-9]`), choose a **Status**, and click **Create**.
4) Verify the expected outlets are created for each device with the selected status.

### Notes
- UI-only change; no model or API impact.

### DIFF
--- a/netbox/dcim/forms/bulk_create.py
+++ b/netbox/dcim/forms/bulk_create.py
@@ -69,11 +69,14 @@ class PowerPortBulkCreateForm(
 
 
 class PowerOutletBulkCreateForm(
-    form_from_model(PowerOutlet, ['type', 'color', 'feed_leg', 'mark_connected']),
+    form_from_model(PowerOutlet, ['type', 'status', 'color', 'feed_leg', 'mark_connected']),
     DeviceBulkAddComponentForm
 ):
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @pheus!

PR #20190, opened 2025-08-27
Merge main into feature

arthanson:
Merge main into feature, Several of the template files had merge conflicts:
```
	both modified:   netbox/templates/inc/panels/related_objects.html
	both modified:   netbox/utilities/templates/buttons/add.html
	both modified:   netbox/utilities/templatetags/buttons.py
```


netbox/templates/inc/panels/related_objects.html
jeremystretch, line 1: This needs to be adapted to accommodate the change in a585bc044ebdb9eebf88df5d91883b1f47f302d7. (We need to use `action_url` now.)

netbox/utilities/templatetags/buttons.py
```diff
@@ -164,7 +164,7 @@ def sync_button(instance):
 
 # TODO: Remove in NetBox v4.7
 @register.inclusion_tag('buttons/add.html')
-def add_button(model, action='add'):
+def add_button(model, return_url=None, action='add'):
```
jeremystretch: `return_url` should be positioned after `action` to avoid altering the function's signature in a potentially breaking way.

jeremystretch (changes requested):

jeremystretch (approved):

PR #20207, opened 2025-08-29
Fixes #19896: cf minmax mustbe int

Jathn:
### Fixes #19896: Allow decimal number boundaries for custom fields

The changes made allows storing decimal values as boundary values for custom fields. This fixes the previous bug that caused values to be stored as whole numbers.

Changes made:
- Changes customfields model to use models.DecimalField to store validation_maximum and validation_minimum
- Changes customfields filterset to use forms.DecimalField for respective fields
- Database migration 0130 in extras application to allow above changes


arthanson (changes requested):
The min/max is different then the decimal field definition also I think larger then integer could be 12 seems excessive I think 10 (non decimal) places works for max integer 2147483647, wondering if 16 max and 4 decimal would work better here?

```
           field = forms.DecimalField(
                required=required,
                initial=initial,
                max_digits=12,
                decimal_places=4,
                min_value=self.validation_minimum,
                max_value=self.validation_maximum
            )
```

Jathn:
I agree. I originally looked mainly at the storage and that it was a BigIntegerField, however to my understanding it seems only Integers could be entered anyway. Please see new changes:
- max_digits set to 16 and decimal_fields set to 4 on both min/max and the field. [276dd6c](https://github.com/netbox-community/netbox/pull/20207/commits/276dd6c94e367b6d5a8154f807419d00144edef9)
- I also noted that the graphql still used IntegerLookup so changed it to FloatLookup in line with other DecimalField implementations. [017b173](https://github.com/netbox-community/netbox/pull/20207/commits/017b17376178efb9fb81d03b27e19ae44505ef09)
- Fixed history of migrations: 0130 -> 0133 [28be6a7](https://github.com/netbox-community/netbox/pull/20207/commits/28be6a726d738af0fbcba78baa58cfdce220f397)

Edit: added note about migrations

arthanson (approved):

arthanson:
Thanks @Jathn !

PR #20219, opened 2025-09-01
Fixes: #19825 - Prevent cache for config revisions from being overwritten when in debug mode when not intended

DanSheps:
### Fixes: #19825 - Prevent cache for config revisions from being overwritten when in debug mode when not intended

* Obtain the current config revision information
* Reinitialize that information when app is restarted in debug mode


DanSheps:
Looks like this was changed in #18021 to automatically clear cache on startup.

We may want to revisit this completely as this could hit way more then just swagger.

jeremystretch (dismissed):
I don't think this is the optimal way to approach this. `cache.clear()` might be called elsewhere (including by plugins), so we need to ensure that the active configuration is always reset correctly in case the entire cache is cleared. This will probably require tweaking the initialization logic for the `Config` class.

DanSheps:
> I don't think this is the optimal way to approach this. `cache.clear()` might be called elsewhere (including by plugins), so we need to ensure that the active configuration is always reset correctly in case the entire cache is cleared. This will probably require tweaking the initialization logic for the `Config` class.

I was kind of wrestling with this as well.  As the `Config` class is supposed to be immutable I didn't want to directly set an `status` parameter on the class.

DanSheps:
Thoughts on adding a `active = models.BooleanField(default=False)` on the ConfigRevision model to properly track which ConfigRevision is active using the `.update()` or `.bulk_update()` method to bypass the signals and update the fields directly with an appropriate database constraint to ensure only one revision active at a time?

This would remove some of the non-concrete logic we have going on (based on the last config saved).  Alternatively, have a date which we utilize instead.

jnovinger:
Hey @DanSheps ,

I've been thinking about this PR and Jeremy's feedback about needing to address the broader architectural issue rather than just the debug mode cache clearing. And really, I think the issue as reported is a bit off--I think the caching issue has just be papering over the flawed "assume the most recent config revision is active" approach.

I think you have identified the core of a clean solution that addresses Jeremy's concerns while being doable for v4.4.1. Thinking about it seems like we would need to do at least the following:

1. Add the `active` field as you described, with something like `UniqueConstraint(fields=['active'], condition=Q(active=True))`, along with the required schema migration
2. A data migration that does its best to intuit the current active config revision before fallback to the "most recently" created method of determing which is active
3. Update `_populate_from_db()` to use `.get(active=True)` and update edge-case handling accordingly
4. Update `activate()` method to manage the boolean field
5. Update `is_active` property to use database field

Given the Tuesday deadline for including this in v4.4.1, I wanted to check: do you still want to implement this approach, or would you prefer I take ownership?

DanSheps:
I think I can get that done by Tuesday

DanSheps:
Turns out, I had most of this done...  lol

netbox/core/migrations/0019_configrevision_active.py
```diff
@@ -0,0 +1,46 @@
...
+    version = None
+    revision = None
+
+    try:
+        version = cache.get('config_version')
+    except Exception:
+        pass
+
+    if version:
+        revision = ConfigRevision.objects.filter(pk=version).first()
```
jnovinger: I'm tempted to say that this should be a `.get(pk=version)` which encodes our expectation that only one most recent config revision can exist.

But, then there's just another edge-case to handle. And, I'm not sure that's likely enough to ever happen and how we'd even handle it. Given that this is the approach we had before of getting the most recent config from the DB in the past and we haven't seen complaints about it not being deterministic, then it's safe.

netbox/netbox/config/__init__.py
```diff
@@ -78,7 +78,9 @@ def _populate_from_db(self):
         from core.models import ConfigRevision
 
         try:
```
jnovinger, line 80: I was just thinking this through and trying decide if a mostly try/except control flow here is better or not. I think I finally convinced myself. This is kind of what I had in mind:

```python
  try:
      revision = ConfigRevision.objects.get(active=True)
      logger.debug(f"Loaded active configuration revision #{revision.pk}")
  except ConfigRevision.DoesNotExist:
      logger.warning("No active configuration revision found - falling back to most recent")
      revision = ConfigRevision.objects.order_by('-created').first()
      if revision is None:
          logger.debug("No configuration revisions exist; using default values")
          return
      logger.debug(f"Using fallback configuration revision #{revision.pk}")
  except DatabaseError:
      logger.warning("Database unavailable during config initialization")
      return
```

Here's my reasoning:
1. The use of `.get()` clearly communicates "we expect one and exactly one config revision to be active
2. Catching `DoesNotExist` allows us to fall back to the old approach. We should expect that this should not happen very often at all, but does provide confidence in the safety of the new mechanism.
3. The inclusion of extra logging helps in the cases where something unexpected happens.

Thoughts?

netbox/core/apps.py
```diff
@@ -42,6 +42,13 @@ def ready(self):
...
         if settings.DEBUG:
             try:
+                config = cache.get('config')
+                config_version = cache.get('config_version')
                 cache.clear()
+                if config_version:
+                    # Activate the current config revision
+                    # Do not query DB due to apps still initializing
+                    cache.set('config', config, None)
+                    cache.set('config_version', config_version, None)
```
jnovinger: What's the value of keeping this around?

netbox/core/migrations/0019_configrevision_active.py
DanSheps: > I'm tempted to say that this should be a `.get(pk=version)` which encodes our expectation that only one most recent config revision can exist.

I could have used get, but I didn't want a needless `try: except ConfigRevision.DoesNotExist:` in the migration.

I do have an idea of how to clean that up though because I do see at least one edge case which shouldn't happen but might want to protect against now that you mention it:

* The version in cache doesn't exist for whatever reason as an actual item in ConfigRevision anymore

Either way, I am going to update the code and add comments so it is easier to understand.

netbox/netbox/config/__init__.py
DanSheps: I am fine with moving to a try/except.  However the fall back is probably going to happen more often then you anticipate as it will be happening anywhere where someone has not defined a configuration revision.

Wonder if it would make more sense to short-circuit it first with a ConfigRevision.objects.exists() to see if there are even any records in the DB yet.  Thoughts on that?

netbox/core/apps.py
DanSheps: No, not needed, didn't hurt anything (and maybe prevented an extra DB query) but not needed.  Will get rid of it now.

netbox/netbox/config/__init__.py
jnovinger: Yeah, you could do that, but in the positive case where one does exist (whether it's `active = True` or not, you're going to do at least two database round trips.

Even though exception handling is a little slower than simple `if`s, I think doing the optimistic query with `.get(...)` and catching the `DoesNotExist` accomplishes the same thing with a max of one database round trip.

jnovinger (approved):
Apologies for letting this sit for so long.

PR #20230, opened 2025-09-02
Closes #20203: Add a pre-commit check for OpenAPI schema changes

### LINKED ISSUE #20203: Add pre-commit check for changes to OpenAPI schema
### Proposed Changes

1. Commit a static copy of NetBox's OpenAPI schema (the JSON data published at `/api/schema/`)\*
2. Introduce a `pre-commit` check that regenerates this file and fails if any changes are detected

\* Use of this static file for other purposes should be discouraged as it is limited to the base API schema, and does not include any plugins installed in the user's own NetBox deployment.

### Justification

The goal here is to implement a check for NetBox's OpenAPI schema to mitigate unexpected changes. Should the pre-commit check fail, the author will need to either update the schema as part of the commit, or make adjustments to avoid the unexpected change. (Changes are generally anticipated only for breaking releases.)

jeremystretch:
### Closes: #20203

- Add a static record of the OpenAPI schema at `contrib/openapi.json`
- Introduce the `verify-openapi.sh` script to generate the OpenAPI schema & compare it to the static record
- Add a `pre-commit` rule to run the OpenAPI verification check
- Update the release documentation to include updating `contrib/openapi.json`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,14 @@ repos:
       language: system
       pass_filenames: false
       types: [python]
+    - id: openapi-check
+      name: "Validate OpenAPI schema"
+      description: "Check for any unexpected changes to the OpenAPI schema"
+      files: api/.*\.py$
+      entry: scripts/verify-openapi.sh
+      language: system
+      pass_filenames: false
+      types: [python]
     - id: mkdocs-build
       name: "Build documentation"
       description: "Build the documentation with mkdocs"
--- a/docs/development/release-checklist.md
+++ b/docs/development/release-checklist.md
@@ -123,16 +123,6 @@ $ node bundle.js
 Done in 1.00s.
 ```
 
-### Rebuild the Device Type Definition Schema
-
-Run the following command to update the device type definition validation schema:
-
-```nohighlight
-./manage.py buildschema --write
-```
-
-This will automatically update the schema file at `contrib/generated_schema.json`.
-
 ### Update & Compile Translations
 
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -21,6 +21,14 @@ repos:
       language: system
       pass_filenames: false
       types: [python]
+    - id: openapi-check
+      name: "Validate OpenAPI schema"
+      description: "Check for any unexpected changes to the OpenAPI schema"
+      files: api/.*\.py$
```
jeremystretch, line 27: Lazy regex to avoid triggering the rule unless it's likely that a modification to the REST API was made.

docs/development/release-checklist.md
```diff
@@ -160,6 +150,24 @@ Then, compile these portable (`.po`) files for use in the application:
...
+```
+
+This will automatically update the schema file at `contrib/generated_schema.json`.
+
+### Update the OpenAPI Schema
+
+Update the static OpenAPI schema definition at `contrib/openapi.json` with the management command below. If the schema file is up-to-date, only the NetBox version will be changed.
+
+```nohighlight
+./manage.py spectacular --format openapi-json > ../contrib/openapi.json
```
arthanson, line 168: I think that should be `./manage.py spectacular --format openapi-json > contrib/openapi.json` the ../ before contrib causes an error.

arthanson (changes requested):

docs/development/release-checklist.md
jeremystretch: The command assumes you're in the path where `manage.py` exists; `contrib/` is one up from that.

arthanson (approved):

PR #20246, opened 2025-09-04
Closes #20241: Record A & B terminations on cable changelog records

### LINKED ISSUE #20241: Record A & B terminations on changelog record for cable
### NetBox version

v4.4.0

### Feature type

New functionality

### Proposed functionality

Currently, when creating or modifying a cable, NetBox does not record any data about its associated terminations in the changelog record for the cable itself. (Each termination gets its own changelog record.) We should be able to store the numeric IDs of each termination on the A and B sides of a cable in the changelog to complement the individual CableTermination records.

We can extend the `save()` method on the Cable model to perform the following tasks:

1. Call `super().save()` initially if the cable is being created. (This ensures that it exists in the database prior to creating any new CableTerminations.)
2. Create/delete any associated CableTermination records in the database as needed.
3. Call `super().save()` to capture the updated cable terminations.

This will result in two changelog records when creating a new cable, but I don't expect that to be problematic (and may in fact be necessary to accurately reflect what happens in the database).

The resulting serialized post-change data would look like this:

```
{
    label: "",
    status: "connected",
    a_terminations: [
        223
    ],
    b_terminations: [
        224
    ],
    ...
}
```

### Use case

Including `a_terminations` and `b_terminations` on the change record for a cable allows a user to determine immediately whether the terminations for a cable were changed, without needing to reference the associated change records for each termination. This would also enable us to address some issues with cabling in the `netbox-branching` plugin (see netboxlabs/netbox-branching#150 for example).

### Database changes

None anticipated

### External dependencies

N/A

jeremystretch:
### Closes: #20241

- Call `save()` twice for new cables (before and after updating the terminations)
- Consolidate the logic for getting/setting A/B terminations into the new `_get_x_terminations()` and `_set_x_terminations()` methods
- Add `get_terminations()` to Cable to return a mapping of all current CableTerminations
- Move the logic for updating CableTerminations to a new `update_terminations()` method on Cable
- Override `serialize_object()` on Cable to add `a_terminations` and `b_terminations` to the change data
- Add a complementary `deserialize_object()` method (for completeness)
- Update some testing utilities to specifically check for creation records (ignoring subsequent update records)

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -18,6 +18,7 @@ from utilities.conversion import to_meters
 from utilities.exceptions import AbortRequest
 from utilities.fields import ColorField, GenericArrayForeignKey
 from utilities.querysets import RestrictedQuerySet
+from utilities.serialization import serialize_object
 from wireless.models import WirelessLink
 from .device_components import FrontPort, RearPort, PathEndpoint
 
@@ -119,6 +120,9 @@ class Cable(PrimaryModel):
         pk = self.pk or self._pk
         return self.label or f'#{pk}'
 
+    def get_status_color(self):
+        return LinkStatusChoices.colors.get(self.status)
+
     @property
     def a_terminations(self):
         if hasattr(self, '_a_terminations'):
@@ -208,7 +212,7 @@ class Cable(PrimaryModel):
             for termination in self.b_terminations:
                 CableTermination(cable=self, cable_end='B', termination=termination).clean()
 
-    def save(self, *args, **kwargs):
+    def save(self, *args, force_insert=False, force_update=False, using=None, update_fields=None):
         _created = self.pk is None
 
         # Store the given length (if any) in meters for use in database ordering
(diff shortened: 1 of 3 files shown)


netbox/dcim/models/cables.py
```diff
@@ -119,43 +120,61 @@ def __str__(self):
...
-            return self._a_terminations
+    def get_status_color(self):
+        return LinkStatusChoices.colors.get(self.status)
+
+    def _get_x_terminations(self, side):
+        """
+        Return the terminating objects for the given cable end (A or B).
+        """
+        if side not in (CableEndChoices.SIDE_A, CableEndChoices.SIDE_B):
+            raise ValueError("Unknown cable side: {side")
```
jnovinger: It looks like this should have been an f-string? Also, looks like the trailing curly-brace is missing.

netbox/dcim/models/cables.py
```diff
@@ -119,43 +120,61 @@ def __str__(self):
...
 
-    @a_terminations.setter
-    def a_terminations(self, value):
-        if not self.pk or self.a_terminations != list(value):
+    def _set_x_terminations(self, side, value):
+        """
+        Set the terminating objects for the given cable end (A or B).
+        """
+        if side not in (CableEndChoices.SIDE_A, CableEndChoices.SIDE_B):
+            raise ValueError("Unknown cable side: {side")
```
jnovinger: Same as above

jnovinger (commented):
Is there any concern that providing multiple avenues to modify cable terminations could create conflicts when both approaches operate on the same cable? For example:

```py
  # Request 1: UI user editing cable via new approach
  cable = Cable.objects.get(pk=1)  # On branch X
  cable.a_terminations = [interface1]
  # ... during double-save process

  # Request 2: Plugin or custom script using old approach on same branch
  CableTermination.objects.create(
      cable_id=1,
      cable_end='A',
      termination=interface2
  )

  # Request 1 continues: update_terminations() reconciles based on stale cable state
```

This seems like it could happen with simultaneous UI/API operations, plugins, or custom scripts that manipulate `CableTerminations` directly during the save window.

I realize this is probably rare since most API usage would go through the Cable endpoints anyway, but wanted to check if there should be any safeguards or if this is considered acceptable risk given the benefits.

jeremystretch:
All modifications typically occur within a transaction bound to the request, so I don't think that's much of a concern. At least, I don't think it's any _more_ of a concern than prior to the change.

Tangentially related, there's an effort to discourage the direct modification of cable terminations (see #20295), a mantra which should be extended to custom scripts as well.

jnovinger:
> All modifications typically occur within a transaction bound to the request, so I don't think that's much of a concern. At least, I don't think it's any _more_ of a concern than prior to the change.
> 
> Tangentially related, there's an effort to discourage the direct modification of cable terminations (see #20295), a mantra which should be extended to custom scripts as well.

Ah, right, I forgot about that! Thanks for the clarification.

jnovinger (approved):

PR #20255, opened 2025-09-04
Closes #20237 - Add search index for TunnelGroup

### LINKED ISSUE #20237: Include VPN Tunnel Groups in the global search.
### NetBox version

v4.2.9

### Feature type

Data model extension

### Proposed functionality

I suggest we add support for Tunnel Groups to show up in the global Netbox search, in the same way VLAN Groups currently show up.

### Use case

We use VPN Tunnel Groups to group several IPsec tunnels together. One group serves one tenant and the name of the group is used both in our configuration and communication with clients. Unfortunately Tunnel Groups do not show up in the Netbox global search, which makes it harder to find these tunnels unless you know to search in the specific Tunnel Groups table.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Fixes: #20237

Adds `TunnelGroupIndex` so VPN Tunnel Groups appear in the global search (parity with VLAN Groups). Indexes `name` (high weight) and `description` (lower weight). No schema or API changes.

**Tested:** Created a sample group; global search now hits on name/description and links to the object view as expected.

### DIFF
--- a/netbox/vpn/search.py
+++ b/netbox/vpn/search.py
@@ -15,6 +15,17 @@ class TunnelIndex(SearchIndex):
 
 
 @register_search
+class TunnelGroupIndex(SearchIndex):
+    model = models.TunnelGroup
+    fields = (
+        ('name', 100),
+        ('slug', 110),
+        ('description', 500),
+    )
+    display_attrs = ('description',)
+
+
+@register_search
 class IKEProposalIndex(SearchIndex):
     model = models.IKEProposal
     fields = (


netbox/vpn/search.py
```diff
@@ -14,6 +14,16 @@ class TunnelIndex(SearchIndex):
     display_attrs = ('group', 'status', 'encapsulation', 'tenant', 'tunnel_id', 'description')
 
 
+@register_search
+class TunnelGroupIndex(SearchIndex):
+    model = models.TunnelGroup
+    fields = (
+        ('name', 100),
```
jeremystretch, line 21: Suggest adding `slug` as well for consistency with similar models.

```suggestion
        ('name', 100),
        ('slug', 110),
```

jeremystretch (changes requested):
Thanks @pheus!

netbox/vpn/search.py
pheus: Thanks for the fast review and for the suggestions!

jeremystretch (approved):

PR #20261, opened 2025-09-05
Fixes #20217: Fix '0 VLANs available' in the VLANs table in VLAN Groups

### LINKED ISSUE #20217: VLAN Groups show "0 VLANS Available" following a range if the last VLAN in the range exists
### Deployment Type

NetBox Cloud

### NetBox Version

v4.3.7

### Python Version

3.10

### Steps to Reproduce

1. Add a new VLAN Group with one or more child VLAN ID ranges (something like "5-6,10-20,30-35" or "1-4094")
2. Add one or more VLANs that corresponds to the last VLAN ID in one or more of those ranges. Make sure that they are assigned to that new VLAN Group. So for "5-6,10-20,30-35" add VLANs 6, 20, 35, for "1-4094" add 4094, etc.
3. Open the VLAN Group, go to the VLANs tab. Notice that NetBox shows "0 VLANs available" after the last defined VLAN in those ranges.

### Expected Behavior

NetBox should only display "x VLANs Available" if there is at least 1 or more availble VLANs ID at the end a range, matching the behavior with IP addresses in a prefix.

### Observed Behavior

The VLANs table in the VLAN Group contains an entry that displays "0 VLANs available" at the end of a range if the last VLAN in the range exists. While this is technically true, this can be confusing for a user -- at a quick glance a user may believe there is room for another VLAN, when in actuality this is not true. Furthermore, if the user clicks the "0 VLANs available" button, NetBox will try to add the next VLAN sequentially, which falls outside of the VLAN range. If the user then tries to save this page then they will get an error that this VLAN is outside of the range.

Here I created a VLAN Group with range of "5,10,15,20,25" and created and associated the five VLANs:

Image

aq5747:
### Fixes: #20217 

Fixes an issue where the VLANs table under a VLAN Group will display a clickable `0 VLANs available` button when the final VLAN in the range is in use.

### DIFF
--- a/netbox/ipam/utils.py
+++ b/netbox/ipam/utils.py
@@ -164,11 +164,11 @@ def available_vlans_from_range(vlans, vlan_group, vid_range):
         prev_vid = vlan.vid
 
     # Annotate any remaining available VLANs
-    if prev_vid < max_vid:
+    if prev_vid < max_vid and (available := max_vid - prev_vid - 1) > 0:
         new_vlans.append({
             'vid': prev_vid + 1,
             'vlan_group': vlan_group,
-            'available': max_vid - prev_vid - 1,
+            'available': available,
         })
 
     return new_vlans
--- a/netbox/ipam/utils.py
+++ b/netbox/ipam/utils.py
@@ -164,11 +164,11 @@ def available_vlans_from_range(vlans, vlan_group, vid_range):
         prev_vid = vlan.vid
 
     # Annotate any remaining available VLANs
-    if prev_vid < max_vid and (available := max_vid - prev_vid - 1) > 0:
+    if prev_vid < max_vid - 1:
         new_vlans.append({
             'vid': prev_vid + 1,
             'vlan_group': vlan_group,
-            'available': available,
+            'available': max_vid - prev_vid - 1,
         })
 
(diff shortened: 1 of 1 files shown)


netbox/ipam/utils.py
```diff
@@ -164,11 +164,11 @@ def available_vlans_from_range(vlans, vlan_group, vid_range):
         prev_vid = vlan.vid
 
     # Annotate any remaining available VLANs
-    if prev_vid < max_vid:
+    if prev_vid < max_vid and (available := max_vid - prev_vid - 1) > 0:
```
jeremystretch: While this works, it's a bit confusing IMO. Would it not be sufficient to just account for the off-by-one error (and leave the value of `available` as is)?

```suggestion
    if prev_vid < max_vid - 1:
```

jeremystretch (changes requested):

netbox/ipam/utils.py
aq5747: Fair point! I must have had some tunnel vision going on and didn't see the simpler, cleaner fix. 

I'll revise it.

jeremystretch (approved):
Thanks @aq5747!

PR #20262, opened 2025-09-05
Closes #20222: Enable HttpOnly flag for the CSRF cookie

### LINKED ISSUE #20222: Enable HttpOnly flag for the CSRF cookie
### NetBox version

v4.2.9

### Feature type

Other

### Proposed functionality

Enable the Django setting [`CSRF_COOKIE_HTTPONLY`](https://docs.djangoproject.com/en/stable/ref/settings/#csrf-cookie-httponly).

If there are any plugins that access the CSRF cookie from JS code instead of using the `window.CSRF_TOKEN` property (see https://github.com/netbox-community/netbox/pull/10205), they might break. So alternatively make the setting configurable by adding to `settings.py`:

    CSRF_COOKIE_HTTPONLY = getattr(configuration, 'CSRF_COOKIE_HTTPONLY', False)

I have tested `CSRF_COOKIE_HTTPONLY = True` in our environment according to steps described in https://github.com/netbox-community/netbox/issues/9963 and found no issues. I’ll be happy to submit a pull request if this proposal is accepted. Thanks!

### Use case

Mainly to appease our auditors. The security benefit is marginal, but there should be no downside (except as described above if the setting is enabled unconditionally).

### Database changes

None.

### External dependencies

None.

jetomit:
### Fixes: #20222

This prevents JS code from reading the CSRF token from the cookie store. If needed, the `window.CSRF_TOKEN` property can be used instead (see also commit cdea302).

Actual security benefit of this change is minimal, but it is sometimes requested by auditors.

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -84,6 +84,7 @@ CORS_ORIGIN_REGEX_WHITELIST = getattr(configuration, 'CORS_ORIGIN_REGEX_WHITELIS
 CORS_ORIGIN_WHITELIST = getattr(configuration, 'CORS_ORIGIN_WHITELIST', [])
 CSRF_COOKIE_NAME = getattr(configuration, 'CSRF_COOKIE_NAME', 'csrftoken')
 CSRF_COOKIE_PATH = f'/{BASE_PATH.rstrip("/")}'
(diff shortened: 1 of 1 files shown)


arthanson (approved):
Thanks @jetomit !

PR #20267, opened 2025-09-05
Fixes #19744: Add accessor for is_loaded in TemplateColumn

### LINKED ISSUE #19744: Unable to sort plugins list by Active column
### Deployment Type

Self-hosted

### NetBox Version

v4.3.2

### Python Version

3.11

### Steps to Reproduce

1. Install a local plugin (like https://github.com/markkuleinio/netbox-hello-world-page for a small example)
2. Go to Admin - Plugins
3. Click the **Local** column two times and see that the plugins list is sorted accordingly (= this works)
4. Now click the **Active** column two times

### Expected Behavior

The plugins list is sorted by the Active status

### Observed Behavior

The plugins list is sorted so that the Active plugin is always at the end of the list.

pheus:
### Fixes: #19744

**Summary**  
Fix sorting for the **Active** column on **Admin → Plugins**.

**Change**  
Add `accessor=tables.A('is_loaded')` to the Active `TemplateColumn` so django‑tables2 orders by the boolean value rather than the rendered HTML.

**Result**  
Clicking **Active** now correctly toggles:
- **Ascending:** inactive → active  
- **Descending:** active → inactive

### DIFF
--- a/netbox/core/tables/plugins.py
+++ b/netbox/core/tables/plugins.py
@@ -61,6 +61,7 @@ class CatalogPluginTable(BaseTable):
         verbose_name=_('Local')
     )
     is_installed = columns.TemplateColumn(
+        accessor=tables.A('is_loaded'),
         verbose_name=_('Active'),
         template_code=PLUGIN_IS_INSTALLED
     )


pheus:
Alternatively, we could rename the table column to `is_loaded`. If you prefer that approach, I’m happy to update the PR.

bctiemann (approved):

PR #20269, opened 2025-09-05
Closes #19547: Add Sync button for DataSource ListView

### LINKED ISSUE #19547: Add 'Sync' button to DataSource Listview
### NetBox version

v4.1.11

### Feature type

### Proposed functionality

Add a 'Sync' button to the listview of the DataSources

Image

### Use case

Quick access to the sync of one or more datasources.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Summary
Adds a **Sync** button to the Data Sources list view so a sync can be triggered directly from the table (no need to open the detail page).

### Details
- Introduces a per‑row **Sync** action in the list view.
- Visibility/enabled state respects user permissions and current job status.
- Reuses the existing sync mechanism/queue; no new API surface.

### UX
- Button sits alongside existing row actions.
- Disabled while a sync is pending or running.

### Notes
Open to feedback on approach.

### DIFF
--- a/netbox/core/tables/template_code.py
+++ b/netbox/core/tables/template_code.py
@@ -26,3 +26,19 @@ PLUGIN_IS_INSTALLED = """
     <span class="text-muted">&mdash;</span>
 {% endif %}
 """
+
+DATA_SOURCE_SYNC_BUTTON = """
+{% load helpers %}
+{% load i18n %}
+{% if perms.core.sync_datasource %}
+    {% if record.ready_for_sync %}
+        <button class="btn btn-primary btn-sm" type="submit" formaction="{% url 'core:datasource_sync' pk=record.pk %}?return_url={{ request.get_full_path|urlencode }}" formmethod="post">
+            <i class="mdi mdi-sync" aria-hidden="true"></i> {% trans "Sync" %}
+        </button>
+    {% else %}
+        <button class="btn btn-primary btn-sm" disabled>
+            <i class="mdi mdi-sync" aria-hidden="true"></i> {% trans "Sync" %}
+        </button>
+    {% endif %}
+{% endif %}
+"""
--- a/netbox/core/views.py
+++ b/netbox/core/views.py
@@ -33,7 +33,13 @@ from utilities.forms import ConfirmationForm
 from utilities.htmx import htmx_partial
 from utilities.json import ConfigJSONEncoder
 from utilities.query import count_related
-from utilities.views import ContentTypePermissionRequiredMixin, GetRelatedModelsMixin, ViewTab, register_model_view
+from utilities.views import (
+    ContentTypePermissionRequiredMixin,
+    GetRelatedModelsMixin,
+    GetReturnURLMixin,
+    ViewTab,
+    register_model_view,
+)
 from . import filtersets, forms, tables
 from .jobs import SyncDataSourceJob
 from .models import *
@@ -66,7 +72,7 @@ class DataSourceView(GetRelatedModelsMixin, generic.ObjectView):
 
 
 @register_model_view(DataSource, 'sync')
-class DataSourceSyncView(BaseObjectView):
+class DataSourceSyncView(GetReturnURLMixin, BaseObjectView):
     queryset = DataSource.objects.all()
 
     def get_required_permission(self):
@@ -85,7 +91,7 @@ class DataSourceSyncView(BaseObjectView):
             request,
             _("Queued job #{id} to sync {datasource}").format(id=job.pk, datasource=datasource)
         )
-        return redirect(datasource.get_absolute_url())
+        return redirect(self.get_return_url(request, datasource))
 
 
 @register_model_view(DataSource, 'add', detail=False)
--- a/netbox/core/tables/data.py
+++ b/netbox/core/tables/data.py
@@ -4,6 +4,7 @@ import django_tables2 as tables
 from core.models import *
 from netbox.tables import NetBoxTable, columns
 from .columns import BackendTypeColumn
+from .template_code import DATA_SOURCE_SYNC_BUTTON
 
 __all__ = (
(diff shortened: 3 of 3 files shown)


netbox/core/tables/template_code.py
```diff
@@ -26,3 +26,21 @@
...
+{% load helpers %}
+{% load i18n %}
+  {% if perms.core.sync_datasource %}
+    {% if record.ready_for_sync %}
+      <button type="submit" class="btn btn-primary btn-sm" formaction="{% url 'core:datasource_sync' pk=record.pk %}?return_url={{ request.get_full_path|urlencode }}" formmethod="post">
+        <i class="mdi mdi-sync" aria-hidden="true"></i> {% trans "Sync" %}
+      </button>
+    {% else %}
+      <span class="inline-block" tabindex="0" data-bs-toggle="tooltip" data-bs-delay="100" data-bs-placement="bottom">
+        <button class="btn btn-primary" disabled>
```
jeremystretch: Need to add the `btn-sm` class for consistent sizing.

```suggestion
        <button class="btn btn-primary btn-sm" disabled>
```

netbox/core/tables/template_code.py
```diff
@@ -26,3 +26,21 @@
...
+DATA_SOURCE_SYNC_BUTTONS = """
+{% load helpers %}
+{% load i18n %}
+  {% if perms.core.sync_datasource %}
+    {% if record.ready_for_sync %}
+      <button type="submit" class="btn btn-primary btn-sm" formaction="{% url 'core:datasource_sync' pk=record.pk %}?return_url={{ request.get_full_path|urlencode }}" formmethod="post">
+        <i class="mdi mdi-sync" aria-hidden="true"></i> {% trans "Sync" %}
+      </button>
+    {% else %}
+      <span class="inline-block" tabindex="0" data-bs-toggle="tooltip" data-bs-delay="100" data-bs-placement="bottom">
```
jeremystretch: Is this `<span>` needed? It doesn't seem to have any effect.

netbox/core/views.py
```diff
@@ -85,6 +85,8 @@
             request,
             _("Queued job #{id} to sync {datasource}").format(id=job.pk, datasource=datasource)
         )
+        if return_url := request.POST.get('return_url'):
+            return redirect(return_url)
```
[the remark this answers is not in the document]
jeremystretch: The `return_url` value needs to be sanitized. We can have the view inherit [`GetReturnURLMixin`](https://github.com/netbox-community/netbox/blob/8311f457b5474a57fc31a5583a390c9404a3acd7/netbox/utilities/views.py#L147) and call `get_return_url()` to handle this.

Note: Even after making this change, CodeQL might still flag this. There are numerous similar false positives we're working on cleaning up.

jeremystretch (changes requested):

netbox/core/tables/template_code.py
pheus: Good catch! I've added `btn-sm` for consistent. Thanks!

netbox/core/tables/template_code.py
pheus: You’re right, the wrapper isn’t needed here. I’ve removed the `<span>` and the unused tooltip attributes.

On the markup, I can either keep two explicit button tags (clearer) or collapse into one with conditional attributes (less duplication). I have a slight preference for the two‑tag version for readability, but happy to switch if you prefer the single‑tag approach. Do you have a preference?

**Option A: two explicit buttons (readable)**

```python
DATA_SOURCE_SYNC_BUTTON = """
{% load helpers %}
{% load i18n %}
{% if perms.core.sync_datasource %}
    {% if record.ready_for_sync %}
        <button class="btn btn-primary btn-sm" type="submit" formaction="{% url 'core:datasource_sync' pk=record.pk %}?return_url={{ request.get_full_path|urlencode }}" formmethod="post">
            <i class="mdi mdi-sync" aria-hidden="true"></i> {% trans "Sync" %}
        </button>
    {% else %}
        <button class="btn btn-primary btn-sm" disabled>
            <i class="mdi mdi-sync" aria-hidden="true"></i> {% trans "Sync" %}
        </button>
    {% endif %}
{% endif %}
"""
```

**Option B: single button with conditional attributes (compact)**


```python
DATA_SOURCE_SYNC_BUTTON = """
{% load helpers %}
{% load i18n %}
{% if perms.core.sync_datasource %}
    <button class="btn btn-primary btn-sm" {% if record.ready_for_sync %}type="submit" formaction="{% url 'core:datasource_sync' pk=record.pk %}?return_url={{ request.get_full_path|urlencode }}" formmethod="post"{% else %}disabled{% endif %}>
        <i class="mdi mdi-sync" aria-hidden="true"></i> {% trans "Sync" %}
    </button>
{% endif %}
"""
```

netbox/core/views.py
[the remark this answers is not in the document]
pheus: Agreed. I’ve updated the view to inherit `GetReturnURLMixin` and now use `get_return_url()` instead of reading `return_url` directly. Thanks for the pointer!

netbox/core/tables/template_code.py
pheus: I went with **Option A** (two explicit buttons) for readability and clearer diffs, but I’m happy to switch to **Option B** (single button with conditional attributes) if that’s preferred. Thanks!

pheus:
Thank you for taking the time to review this PR. I appreciate the thoughtful feedback and attention to detail.

netbox/core/tables/template_code.py
jeremystretch: IMO the way you have it now is fine, and consistent with how we've defined alternate button states in templates elsewhere. Thanks.

jeremystretch (approved):

PR #20280, opened 2025-09-05
Fixes #20264: Update plugin title rendering with default icon

pheus:
### Fixes: #20264

**Summary**
- Replace inline plugin title markup with a reusable template in `template_code.py`.
- Add a **default core SVG icon** used when no plugin icon is provided.
- Keep `title_long` as a `TemplateColumn` using the template; remove redundant `render_title_long()`.
- Correct `order_by` from `name` → `title_long`.

**Impact**
- UI-only; no API/DB changes.
- Deduplicates logic and ensures consistent icon rendering.
- Catalog/author-supplied `icon_url` still overrides the default.


netbox/core/tables/plugins.py
```diff
@@ -93,10 +85,4 @@ class Meta(BaseTable.Meta):
...
         # List installed plugins first, then certified plugins, then
         # everything else (with each tranche ordered alphabetically)
-        order_by = ('-is_installed', '-is_certified', 'name')
-
-    def render_title_long(self, value, record):
-        if record.static:
-            return value
-        url = reverse('core:plugin', args=[record.config_name])
-        return mark_safe(f"<a href='{url}'>{value}</a>")
+        order_by = ('-is_installed', '-is_certified', 'title_long')
```
bctiemann, line 88: I'm not sure about the behavior of `is_installed`. Since it is a field that only renders a template fragment based on the values of `is_local` and `is_loaded`, and there is no `is_installed` concrete field on the Plugin (anymore), is it meaningful to use it as an ordering column?

bctiemann (changes requested):

bctiemann:
I'm also not sure about the icon_url; it seems like all catalog plugins have a default icon, but actual installed plugins (such as Custom Objects and Branching) typically don't have icons, so now the table looks like this:

Screenshot 2025-09-08 at 8 48 51 AM

If the icons haven't been getting displayed for a long time and plugins have been developed without concern for an icon, maybe we need to add a default icon for cases where none is supplied?

netbox/core/tables/plugins.py
bctiemann: Actually I guess this depends on https://github.com/netbox-community/netbox/pull/20267, with which it would make sense, right?

pheus: Thanks for the review!

You’re right: `order_by="-is_installed"` predates this PR and is currently broken because there’s no concrete field. That will be fixed by #20267.  
This PR only corrects the bad reference by changing `order_by="name"` to `order_by="title_long"`.

pheus:
Thanks for calling this out!

A few options to make this consistent:

1) **Fallback default (preferred):** ship a small default SVG “box” in NetBox core and render it when a plugin provides no icon. Consistent look, no new dependency, easy to size.
2) **Standardize on MDI:** use the Material Design Icon as the default for all plugins, letting plugins override with their own icon, but sizing questions remain.
3) **No icon:** hide the icon like it's currently implemented.

Happy to implement whichever you prefer.

bctiemann:
I like the idea of option 1 I think.

It's interesting that the current default "box+" icon suggests that you're supposed to click on it to install the plugin, but it's not clickable and does nothing; maybe the default "box" you're suggesting should take its place? I assume a catalog plugin can override the icon even if it's not installed?

pheus:
Thanks! One nuance to call out:

The current **“box+”** icon doesn’t come from NetBox core; it’s delivered via the **Plugin Catalog `icon_url`** (managed by **NetBox Labs**). So swapping that graphic needs to happen on the catalog side, not in this PR.

In this PR I can:
- Ship a neutral default “box” SVG in core and use it **only when no `icon_url` is provided**.

And yes, catalog plugins can still override the icon at any time (installed or not) by supplying their own `icon_url`. If the catalog keeps sending “box+”, that’s what will render until NetBox Labs updates it.

bctiemann:
Sure, this works for me. It should definitely spur us to make a better default icon for the catalog plugins, but now at least we have the flexibility for that.

bctiemann (approved):

PR #20288, opened 2025-09-06
Fixes #20253: GraphQL filter by contacts

### LINKED ISSUE #20253: GraphQL filter location list by contacts
### NetBox version

v4.3.2

### Feature type

New functionality

### Proposed functionality

Using GraphQL, I would like to be able to filter the `location_list` by assigned contacts.  For example:

```
{
  location_list(
    filters:{
      site: {name: {exact:"LA-LAI"}}
      contacts: {name: {exact:"MSE"}}
    }) {
    id
  }
}
```

This query fails with error
> Unsupported lookup 'name__exact' for GenericRel or join on the field not permitted.

### Use case

We use contacts to define which team is responsible for supporting each location.  I want to build a Grafana dashboard to visualize how many rooms are supported by each team.

### Database changes

_No response_

### External dependencies

_No response_

llamafilm:
### Fixes: #20253

Define a custom filter method for `contacts` using the `@strawberry_django.filter_field` decorator.


netbox/tenancy/graphql/filter_mixins.py
```diff
@@ -17,11 +17,21 @@
 )
 
 
-@dataclass
+@strawberry.type
 class ContactFilterMixin(BaseFilterMixin):
-    contacts: Annotated['ContactFilter', strawberry.lazy('tenancy.graphql.filters')] | None = (
-        strawberry_django.filter_field()
-    )
+    @strawberry_django.filter_field
```
jeremystretch:
```suggestion
    @strawberry_django.filter_field()
```

netbox/tenancy/graphql/filter_mixins.py
```diff
@@ -17,11 +17,21 @@
...
+    def contacts(
+        self,
+        queryset,
```
jeremystretch:
```suggestion
```

netbox/tenancy/graphql/filter_mixins.py
```diff
@@ -17,11 +17,21 @@
...
+        value: Annotated['ContactFilter', strawberry.lazy('tenancy.graphql.filters')],
+        prefix: str,
+    ):
+        return strawberry_django.process_filters(
```
jeremystretch: This should return a Q object to filter the queryset. Here's an example from `IPAddressFilter`:

https://github.com/netbox-community/netbox/blob/1b83d32f4a99faddc52a341877a9d93d39578a0f/netbox/ipam/graphql/filters.py#L188-L194

jeremystretch (changes requested):

netbox/tenancy/graphql/filter_mixins.py
llamafilm: Are you sure? All the examples in strawberry-django docs show using this without parentheses. I tested and it works both ways; I don't mind changing it, but I'm not exactly sure what the difference is.

netbox/tenancy/graphql/filter_mixins.py
llamafilm: What's wrong with `process_filters`?  I tried but could not figure out how to make this work with a Q object. My code works in both 4.3 and 4.4.

jeremystretch: Maintainability demands that we stay reasonably consistent in our approach across the application. Introducing a deviation from an established pattern must be avoided wherever possible.

netbox/tenancy/graphql/filter_mixins.py
jeremystretch: I'm not clear on it myself TBH. Most (but not all) of the existing similar usages invoke the instantiated form (`@strawberry_django.filter_field()`). The two which don't (on [`InterfaceFilter`](https://github.com/netbox-community/netbox/blob/81401b9e1742de4c19fc8511f8ffd5fc08e9b92d/netbox/dcim/graphql/filters.py#L491)) actually trigger a warning in my IDE complaining of an unfilled `resolver` parameter, but that might be inconsequential.

The example usage in the docstring on [`filter_field`](https://github.com/strawberry-graphql/strawberry-django/blob/05a534f555960ec8fc7d00bb9c732b917ffd25ba/strawberry_django/fields/filter_order.py#L217) implies that it should be instantiated when employed as a decorator, so IMO we should be consistent in that approach unless conflicting guidance is presented.

llamafilm:
I found a simpler / better solution — the `contacts` field should actually be a `ContactAssignmentFilter`.  This way I can write more complex queries like this:
```
{
  site_list(
    filters: {
      contacts: {
        contact:{name:{contains:"Dwight"}}
        role:{name:{exact:"Operational"}}
        priority:PRIORITY_PRIMARY
      }
    }) {
    name
  }
}
```

jeremystretch:
I think this is actually a bug, as it doesn't seem possible to filter by contacts at all using the current implementation of `ContactFilterMixin`. I suspect the mixin was intended to use `ContactAssignmentFilter` originally.

jeremystretch (approved):
Thanks @llamafilm!

PR #20291, opened 2025-09-07
Fixes #20272: Inherit ConfigContext from ancestors locations

pheus:
### Fixes: #20272

Make `ConfigContext` on a **parent Location** apply to devices in its **child Locations**.

- Aligns Locations with regions/site groups/roles.
- Updates querysets to include ancestor Locations
  (`get_ancestors(include_self=True)` and an MPTT-based join for devices).
- No schema or API changes.

**Before:** Context on Location A didn’t apply to a device in child Location B.  
**After:** It does.


arthanson (approved):
thanks @pheus

Issue #7604, opened 2021-10-21
Extend filter forms to support filter modifiers


jeremystretch:
### NetBox version

v3.0.8

### Feature type

Change to existing functionality

### Proposed functionality

When viewing a list of objects in the NetBox UI, a "filters" tab is present and lists all of the available filters for the relevant model. However, these filters utilize only the default behavior of each filter, and don't allow for modifications to the filter expressions defined [here](https://netbox.readthedocs.io/en/stable/rest-api/filtering/#lookup-expressions).

This FR proposes to extend the filter forms to include a dropdown widget for each appropriate field, allowing the user to select the desired logic. For example, when filtering by an integer value, the default behavior will be "equals," but options for "less than" (`lt`) and "greater than" (`gt`) will also be available.

### Use case

Although these filter expressions are technically available and working in the current release, this addition to the user interface would make them much more convenient to use.

jeremystretch:
Per #7587 this should also include the ability to negate/invert a filter.

jeremystretch:
We should explore what's feasible for #10352 first, as we can likely build on that solution to implement this.

jsenecal:
So, I was thinking of something like this:
![image](https://user-images.githubusercontent.com/1564902/197538664-be055ac7-941e-4c83-93e5-174cc7355b17.png)

I still have yet to find the proper way of doing it within Django but this is the gist of it. I was thinking of adding this under some sort of "advanced search" accordion.

Does this makes sense to you ?

jeremystretch:
I would prefer to keep each of the filters displayed individually under the "filters" tab, rather than requiring the use to select the desired filter(s) from a dropdown. Additionally, this sort of matrix only seems to makes sense where the user needs to supply a value. When filtering by related objects, an option to invert the logic (excluding the selected objects) seems sufficient.

jeremystretch:
Per @jsenecal unfortunately this won't be done in time for the v3.4 beta, so I'm going to drop it from the milestone.

jeremystretch:
@jsenecal do you still want to tackle this?

jsenecal:
Not sure I'm up to the task, work has the better part of me lately, I'll unassign myself.

jnovinger:
I took a stab at this, with the goals being:

- Maintain consistency with the rest of the NetBox UI/UX
- Have as little JS enhancement as possible
- Use the existing dependencies we have, nothing new
- Leverage as much of the existing infrastructure as possible
[...]

PR #20292, opened 2025-09-07
Closes #19523: Add instance count filter to Module and Device Types

### LINKED ISSUE #19523: Device Types - Filter - With Instances (Yes / No)
### NetBox version

v4.3.0

### Feature type

### Proposed functionality

Expand the filter view so that you can filter by device types that have or do not have instances.

### Use case

We use the Netbox Git project ‘Device-Type-Library-Import’ to automatically import all device types from the official library. This way we avoid creating device types that already exist in good form.

There are currently a total of 3,938 device types in our NetBox instance, including our own. However, we would like to be able to filter the device types actually used without using the sort function in the ‘Insances’ column. Because then it restricts the actual sorting and filtering.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Fixes: #19523

**Summary**  

Add filters to quickly find used/unused types and query by count:
- `has_instances` (bool) — **UI**, **REST**, **GraphQL**
- `instance_count` (int with `=, gt, gte, lt, lte, range`) — **REST**, **GraphQL**

**Implementation**
- `DeviceTypeFilterSet` & `ModuleTypeFilterSet`: add `has_instances` and `instance_count` using `count_related()`; a single method handles integer lookups.
- GraphQL: Strawberry filters via `@filter_field`; annotate `instance_count` and delegate lookups with `process_filters(prefix="…instance_count__")`.
- REST examples:
  - `/api/dcim/device-types/?has_instances=true`
  - `/api/dcim/device-types/?instance_count__gte=10`
  - `/api/dcim/module-types/?has_instances=false`
- Tests: filterset tests for boolean & numeric cases.

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -18,8 +18,8 @@ from tenancy.filtersets import TenancyFilterSet, ContactModelFilterSet
 from tenancy.models import *
 from users.models import User
 from utilities.filters import (
-    ContentTypeFilter, MultiValueCharFilter, MultiValueMACAddressFilter, MultiValueNumberFilter, MultiValueWWNFilter,
-    NumericArrayFilter, TreeNodeMultipleChoiceFilter,
+    AnnotatedCountFilter, ContentTypeFilter, MultiValueCharFilter, MultiValueMACAddressFilter, MultiValueNumberFilter,
+    MultiValueWWNFilter, NumericArrayFilter, TreeNodeMultipleChoiceFilter,
 )
 from virtualization.models import Cluster, ClusterGroup, VMInterface, VirtualMachine
 from vpn.models import L2VPN
@@ -608,6 +608,10 @@ class DeviceTypeFilterSet(NetBoxModelFilterSet):
         method='_inventory_items',
         label=_('Has inventory items'),
     )
+    instance_count = AnnotatedCountFilter(
+        field_name='instance_count',
+        label=_('Instance count'),
+    )
 
     class Meta:
         model = DeviceType
@@ -742,6 +746,10 @@ class ModuleTypeFilterSet(AttributeFiltersMixin, NetBoxModelFilterSet):
     pass_through_ports = django_filters.BooleanFilter(
         method='_pass_through_ports',
         label=_('Has pass-through ports'),
+    )
+    instance_count = AnnotatedCountFilter(
+        field_name='instance_count',
+        label=_('Instance count'),
     )
 
     class Meta:
--- a/netbox/dcim/api/serializers_/devicetypes.py
+++ b/netbox/dcim/api/serializers_/devicetypes.py
@@ -45,8 +45,9 @@ class DeviceTypeSerializer(NetBoxModelSerializer):
     device_bay_template_count = serializers.IntegerField(read_only=True)
     module_bay_template_count = serializers.IntegerField(read_only=True)
     inventory_item_template_count = serializers.IntegerField(read_only=True)
+    instance_count = serializers.IntegerField(read_only=True)
 
-    # Related object counts
+    # Related object counts (TODO: Remove in v4.5)
     device_count = RelatedObjectCountField('instances')
 
     class Meta:
@@ -58,7 +59,7 @@ class DeviceTypeSerializer(NetBoxModelSerializer):
             'created', 'last_updated', 'device_count', 'console_port_template_count',
             'console_server_port_template_count', 'power_port_template_count', 'power_outlet_template_count',
             'interface_template_count', 'front_port_template_count', 'rear_port_template_count',
-            'device_bay_template_count', 'module_bay_template_count', 'inventory_item_template_count',
+            'device_bay_template_count', 'module_bay_template_count', 'inventory_item_template_count', 'instance_count',
         ]
         brief_fields = ('id', 'url', 'display', 'manufacturer', 'model', 'slug', 'description', 'device_count')
 
@@ -101,11 +102,14 @@ class ModuleTypeSerializer(NetBoxModelSerializer):
         allow_null=True
     )
 
+    # Counter fields
+    instance_count = serializers.IntegerField(read_only=True)
+
     class Meta:
         model = ModuleType
         fields = [
             'id', 'url', 'display_url', 'display', 'profile', 'manufacturer', 'model', 'part_number', 'airflow',
             'weight', 'weight_unit', 'description', 'attributes', 'comments', 'tags', 'custom_fields', 'created',
-            'last_updated',
+            'last_updated', 'instance_count',
         ]
         brief_fields = ('id', 'url', 'display', 'profile', 'manufacturer', 'model', 'description')
--- a/netbox/dcim/api/views.py
+++ b/netbox/dcim/api/views.py
@@ -20,6 +20,7 @@ from netbox.api.viewsets import NetBoxModelViewSet, MPTTLockedMixin
 from netbox.api.viewsets.mixins import SequentialBulkCreatesMixin
 from utilities.api import get_serializer_for_model
 from utilities.query_functions import CollateAsChar
+from utilities.query import count_related
 from virtualization.models import VirtualMachine
 from . import serializers
 from .exceptions import MissingFilterException
@@ -266,7 +267,9 @@ class ManufacturerViewSet(NetBoxModelViewSet):
 #
 
 class DeviceTypeViewSet(NetBoxModelViewSet):
-    queryset = DeviceType.objects.all()
+    queryset = DeviceType.objects.annotate(instance_count=count_related(Device, 'device_type')).prefetch_related(
+        'manufacturer', 'default_platform'
+    )
     serializer_class = serializers.DeviceTypeSerializer
     filterset_class = filtersets.DeviceTypeFilterSet
 
@@ -278,7 +281,7 @@ class ModuleTypeProfileViewSet(NetBoxModelViewSet):
 
 
 class ModuleTypeViewSet(NetBoxModelViewSet):
-    queryset = ModuleType.objects.all()
+    queryset = ModuleType.objects.annotate(instance_count=count_related(Module, 'module_type'))
     serializer_class = serializers.ModuleTypeSerializer
     filterset_class = filtersets.ModuleTypeFilterSet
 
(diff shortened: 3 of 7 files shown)


netbox/dcim/filtersets.py
```diff
@@ -639,6 +643,9 @@ def search(self, queryset, name, value):
             Q(comments__icontains=value)
         )
 
+    def _instances(self, queryset, name, value):
+        return queryset.exclude(instances__isnull=value)
```
jeremystretch: I don't think `isnull` is supported for reverse relations. While it appears to suffice in this case, note that the queryset it returns is impacted by the total number of related objects:

```python
>>> dcim.DeviceType.objects.count()
14
>>> dcim.DeviceType.objects.filter(instances__isnull=True).count()
4
>>> dcim.DeviceType.objects.filter(instances__isnull=False).count()
74
```

netbox/dcim/filtersets.py
```diff
@@ -564,6 +564,10 @@ class DeviceTypeFilterSet(NetBoxModelFilterSet):
         lookup_expr='in',
         label=_('Default platform (slug)'),
     )
+    instances = django_filters.BooleanFilter(
```
jeremystretch: Do you think it would make sense to extend this functionality by providing an integer filter to match on the number of instances, rather than merely whether any instances exist? This could be achieved by annotating the queryset with the instance count, which we do already for the UI:

```
    queryset = DeviceType.objects.annotate(
        instance_count=count_related(Device, 'device_type')
    )
```

This would need to be enabled for the REST & GraphQL APIs as well.

jeremystretch (changes requested):

netbox/dcim/filtersets.py
pheus: Thanks for the review and the heads‑up!
I had the same concern about `__isnull` on reverse relations. For consistency, I mirrored the pattern already used elsewhere in `dcim.filtersets` (see [dcim/filtersets.py#L654–L667](https://github.com/netbox-community/netbox/blob/9d0e80571cac3ef3cc541d4c027f900d4c100e92/netbox/dcim/filtersets.py#L654-L667)).

Using `.exclude(...)` plus `.distinct()` keeps the result to one row per type:

```python
def _instances(self, queryset, name, value: bool):
    return queryset.exclude(instances__isnull=value).distinct()
```

**Trade‑off:** `.distinct()` can be a bit heavier than an `EXISTS` subquery on very large datasets. If we ever see a slow query, we can switch to an `Exists` version.

Quick check from my local data:

```python
>>> DeviceType.objects.count()
3
>>> DeviceType.objects.exclude(instances__isnull=False).distinct()
<RestrictedQuerySet [<DeviceType: Catalyst 9300-48U>]>
>>> DeviceType.objects.exclude(instances__isnull=False).distinct().count()
1
>>> DeviceType.objects.exclude(instances__isnull=True).distinct()
<RestrictedQuerySet [<DeviceType: Catalyst 9407R>, <DeviceType: Nexus 93180YC-FX3>]>
>>> DeviceType.objects.exclude(instances__isnull=True).distinct().count()
2
```

netbox/dcim/filtersets.py
pheus: Thanks again! Great suggestion. An integer count filter would be very useful from an API standpoint (e.g., `?instances__gte=10`) and we can expose the same in GraphQL.

On the UI side, list views don’t currently support advanced numeric filter inputs; we’d need the enhancements tracked in **#7604** to surface this there. With that in mind, I’m happy to implement both:

- **`has_instances`** (bool) — works immediately in UI/REST/GraphQL.
- **`instances`** (int) with lookups (`=`, `gte`, `lte`) enabled for REST & GraphQL now; UI can follow once #7604 lands.

The integer filter will take me a bit longer.

pheus:
Thanks for the thoughtful review and patience. I’ve updated the PR to support querying by **instance count**. While working on the GraphQL side I used `ComparisonFilterLookup` so we can support `exact/gt/gte/lt/lte/range`, which seems like a good fit for counts.

Would it be useful to migrate other numeric counters (e.g., `front_port_template_count`, `interface_template_count`, etc.) to `ComparisonFilterLookup` for consistency? If so, would you prefer a follow‑up issue/PR, or should I include it here?

One question (happy to move to a discussion if you’d prefer): booleans in GraphQL are exposed as `FilterLookup[bool]` rather than a plain `bool`, which introduces additional lookups (`contains`, `ends_with`, `exact`, `in_list`, `is_null`, …). For `has_instances` I used `BaseFilterLookup[bool]` to limit it to `exact/in_list/is_null`. Is that acceptable, or would you rather I switch to `bool` or `FilterLookup[bool]`?

Happy to update the PR either way.

jeremystretch:
Thanks @pheus, it's clear you've put a lot of work into this.

One thing to always keep in mind when working on a new feature is to ask "Does this approach scale?" In this case, for instance, what if we had to add the same filters to ten other models? Or 100 other models? That would introduce a _ton_ of boilerplate, which would be impractical to maintain long-term.

There are a few things we can do to simplify the implementation here. First, I want to point out that we typically don't explicitly define filters for individual lookups (`gt`, lt`, etc.`) on the FilterSet classes: These are generated automatically by [`BaseFilterSet`](https://github.com/netbox-community/netbox/blob/192440a4d3e157637afb09473d2fc1b0b9fcdaa1/netbox/netbox/filtersets.py#L44) to avoid thousands of lines of boilerplate.

Next, if we look at the UI view for the DeviceType list, we can see that `instance_count` is already annotated on the queryset:

https://github.com/netbox-community/netbox/blob/192440a4d3e157637afb09473d2fc1b0b9fcdaa1/netbox/dcim/views.py#L1105-L1108

Our `instance_count` filter can reference this value directly; there's no need for a custom method, at least as far as the UI is concerned.

Now `DeviceTypeViewSet`, which serves the REST API does _not_ explicitly annotate this value, but `DeviceTypeSerializer` does include a `device_count` RelatedObjectCountField field which adds the annotation (unless the field has been omitted by request):

https://github.com/netbox-community/netbox/blob/192440a4d3e157637afb09473d2fc1b0b9fcdaa1/netbox/dcim/api/serializers_/devicetypes.py#L50

We could convert this to a normal read-only field and move the annotation to the base queryset to ensure the filter always works with the REST API.

I'm not sure offhand what the preferred approach would be for the GraphQL API, but presumably we have a pattern elsewhere for handling this sort of field. (And if not, that in itself might be something worth exploring.)

We should also consider whether there's a need for an explicit boolean filter. In this case, `has_instances=True` is effectively identical to `instance_count__gt=0`. The only place where I can see the explicit boolean being helpful is in the UI, where its presence would simplify the inclusion of a form field.

pheus:
First of all, thank you so much for taking the time to review the code and explain all these details to me. This is really helpful. I really appreciate it! 💟 

Second, the more I dig into the inner workings of NetBox, the more impressed I am by how much thoughtful helper code lives in the background to make per‑model implementations straightforward.

---

I completely understand your point about reducing duplication and boilerplate. I had assumed this might be an edge case without a more generic solution.

I’ll take the time to read through everything and work out a cleaner, simplified approach.

I know it takes a lot of time to review code and write such thorough explanations, so thank you again for all of this!

pheus:
Thanks for your patience!

I tried to find a common NetBox pattern that would let a FilterSet reuse the **view‑level annotation**. Since FilterSets operate on model fields or custom methods, I couldn’t find a clean way to say “this field is already annotated.” As a stop‑gap, I added an **`AnnotatedCountFilter`** that sets a private `_is_annotated` flag so `BaseFilterSet` skips model‑field checks. It works, but I’m not a fan of introducing a special‑case filter for what seems like a niche need.

As an alternative, I explored a **`CounterCacheField`**. Because it’s a real model field, all FilterSet lookups just work, and reads should be faster (with a small write hit). I pushed a WIP here:  
**19523-add-instance-filter-for-devicetype-and-moduletype-countcachefield** → <https://github.com/pheus/netbox/tree/19523-add-instance-filter-for-devicetype-and-moduletype-countcachefield>  
Open items: add `ModuleType`, a migration to backfill counts, and fix an issue where creating a `Device` fires two `post_save` signals (counter goes +2). I haven’t pinned down the cause yet.

**Naming:** do you prefer `instance_count` or `device_count`? The REST API currently exposes `device_count`, so switching to `instance_count` would be breaking. I left a TODO to align in v4.5, but I’m happy to standardize on `device_count` now and update the PR.

I may be missing an obvious approach here. Any guidance on the preferred direction (reuse annotation vs. small helper vs. counter cache) would be really appreciated. And if this FR is creating more work than it saves, I’m happy to step back.

jeremystretch:
Sorry @pheus I know I owe you a follow-up here. It keeps getting bounced to the bottom of my to-do. Will try to get to it this week.

pheus:
No worries at all. Thanks for the update! This turned out more involved than it first looked, and I know it’s not urgent. Whenever you have time is perfectly fine. I appreciate your time and guidance.

jeremystretch:
> As an alternative, I explored a `CounterCacheField`.

Given the challenges you cite around filters relying on concrete fields, this seems like the preferable approach. (I was actually a little surprised to find that this hasn't been a more common problem in general.) Want to close this and submit an alternative PR with your other branch?

> Naming: do you prefer instance_count or device_count?

I'd prefer to keep the existing `device_count` for DeviceType and just change the underlying annotation to a CounterCacheField, as this would be the least disruptive approach. We should also introduce a `module_count` CounterCacheField on ModuleType to be consistent.

Assuming we move forward with the CounterCacheField approach, we'll need to tag FR #19523 for v4.5 and target the new PR to the `feature` branch. Does that sound good?

pheus:
Thanks so much for reviewing all the options!

> Given the challenges you cite around filters relying on concrete fields, this seems like the preferable approach. (I was actually a little surprised to find that this hasn't been a more common problem in general.)

I’m a bit surprised this hasn’t come up more often too!

> Want to close this and submit an alternative PR with your other branch?

Will do. I’ll close this PR and open a new one based on the counter‑cache branch.

> I'd prefer to keep the existing `device_count` for DeviceType and just change the underlying annotation to a CounterCacheField, as this would be the least disruptive approach. We should also introduce a `module_count` CounterCacheField on ModuleType to be consistent.

As I wasn't sure if we want to follow this path, I skipped the additional work on `ModuleType`. But sure, I will add it in the new PR.

> Assuming we move forward with the CounterCacheField approach, we'll need to tag FR #19523 for v4.5 and target the new PR to the `feature` branch. Does that sound good?

Sounds great. I’ll aim the new PR at `feature`.

Thanks again for the clear guidance - much appreciated!

PR #20302, opened 2025-09-09
Closes #19944: Add multi-scenario CSV import testing support with cleanup

jnovinger:
Closes #19944

- Enhanced `BulkImportObjectsViewTestCase` to support multiple CSV import scenarios via dictionary format, where each scenario runs as a separate subtest with automatic cleanup. This enables testing different import configurations (e.g., with/without optional fields) in a single test run with clear output showing which scenario is being tested.
- Introduces `cleanupSubTest()` context manager that uses database savepoints to automatically roll back changes between subtests, providing test isolation similar to separate test methods. This allows subtests to create/modify objects without affecting subsequent subtests in the same test method.

### DIFF
--- a/netbox/utilities/testing/views.py
+++ b/netbox/utilities/testing/views.py
@@ -152,7 +152,6 @@ class ViewTestCases:
 
         @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'], EXEMPT_EXCLUDE_MODELS=[])
         def test_create_object_with_permission(self):
-
             # Assign unconstrained permission
             obj_perm = ObjectPermission(
                 name='Test permission',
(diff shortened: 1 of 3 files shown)


netbox/utilities/testing/views.py
```diff
@@ -701,35 +755,58 @@ def test_bulk_update_objects_with_permission(self):
...
 
-            # Import permitted objects
+            # Import permitted objects (should succeed)
             self.assertHttpStatus(self.client.post(self._get_url('bulk_import'), data), 302)
-            self.assertEqual(self._get_queryset().count(), initial_count + len(self.csv_data) - 1)
+            self.assertEqual(self._get_queryset().count(), initial_count + expected_new_objects)
+
+            # Reset permission constraints for next scenario
+            obj_perm.constraints = {'pk': 0}  # Reset to deny all
+            obj_perm.save()
```
jeremystretch: Can these lines be removed? They appear unnecessary as `cleanupSubTest()` rolls back the transaction for each test. (`CableTestCase` passes with these lines removed.)
```suggestion
```

netbox/utilities/testing/views.py
```diff
@@ -620,10 +656,38 @@ def test_bulk_import_objects_without_permission(self):
                 self.assertHttpStatus(response, 403)
 
         @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'], EXEMPT_EXCLUDE_MODELS=[])
-        def test_bulk_import_objects_with_permission(self):
+        def test_bulk_import_objects_with_permission(self, post_import_callback=None):
+            scenarios = self.csv_data.keys() if isinstance(self.csv_data, dict) else ['default']
```
jeremystretch: Would it be worthwhile to move this logic to a `get_scenarios()` method to ensure any future changes are captured for both test methods?

jeremystretch (dismissed):
Fixed a minor merge conflict. Had a couple questions but approving as-is.

netbox/utilities/testing/views.py
jnovinger: Yep, that's a good idea.

netbox/utilities/testing/views.py
jnovinger: Yep, I forgot to go back and clean those up. Thanks.

jeremystretch (approved):

PR #20306, opened 2025-09-09
Closes #16137: Remove `is_staff` boolean from User model

### LINKED ISSUE #16137: Remove the `is_staff` field from the user model
### NetBox version

v4.0.2

### Feature type

### Proposed functionality

Remove the `is_staff` boolean field from our custom user model. (This will involve no longer subclassing the `AbstractUser` parent class.)

### Use case

This field was carried over from Django's stock user model in NetBox v4.0. It was previously used to control access to the Django admin UI, but will no longer have any significance once support for the admin UI is [completely removed](https://github.com/netbox-community/netbox/issues/16136).

### Database changes

Removes the `is_staff` field from users.User.

### External dependencies

None

jeremystretch:
### Closes: #16137

- Remove the `is_staff` boolean field from our custom User model
- Remove the `REMOTE_AUTH_STAFF_GROUPS` and `REMOTE_AUTH_STAFF_USERS` configuration parameters
- Introduced the `IsSuperuser` REST API permission to replace DRF's `IsAdminUser` (which evaluates `is_staff`)
- Accessing UI & REST API views for RQ tasks now requires superuser permission
- Accessing the installed plugins API view now requires superuser permission
- Extend the RQ tasks view tests to check permissions enforcement consistently
- `RemoteUserBackend` now infers `is_staff` from `is_superuser`
- `PluginMenuItem` still supports the `staff_only` attribute, but it now maps to the `is_superuser` attribute on a user
- Remove references to `is_staff` from the LDAP authentication docs

### DIFF
--- a/netbox/netbox/authentication/__init__.py
+++ b/netbox/netbox/authentication/__init__.py
@@ -191,11 +191,6 @@ class RemoteUserBackend(_RemoteUserBackend):
         logger.debug(
             f"User {user} should be Superuser: {self._is_superuser(user)}")
 
-        # Set is_staff attribute for compatibility with pre-v4.5
-        user.is_staff = user.is_superuser
-        if user.is_staff:
-            logger.debug(f"Marked user {user} as staff due to superuser status")
-
         user.save()
         return user
 
--- a/netbox/netbox/authentication/__init__.py
+++ b/netbox/netbox/authentication/__init__.py
@@ -184,14 +184,18 @@ class RemoteUserBackend(_RemoteUserBackend):
         else:
             user.groups.clear()
             logger.debug(f"Stripping user {user} from Groups")
+
+        # Evaluate superuser status
         user.is_superuser = self._is_superuser(user)
         logger.debug(f"User {user} is Superuser: {user.is_superuser}")
         logger.debug(
             f"User {user} should be Superuser: {self._is_superuser(user)}")
 
-        user.is_staff = self._is_staff(user)
-        logger.debug(f"User {user} is Staff: {user.is_staff}")
-        logger.debug(f"User {user} should be Staff: {self._is_staff(user)}")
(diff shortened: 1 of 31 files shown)


netbox/netbox/authentication/__init__.py
```diff
@@ -184,14 +184,18 @@ def configure_groups(self, user, remote_groups):
...
         logger.debug(
             f"User {user} should be Superuser: {self._is_superuser(user)}")
 
-        user.is_staff = self._is_staff(user)
-        logger.debug(f"User {user} is Staff: {user.is_staff}")
-        logger.debug(f"User {user} should be Staff: {self._is_staff(user)}")
+        # Set is_staff attribute for compatibility with pre-v4.5
+        user.is_staff = user.is_superuser
+        if user.is_staff:
+            logger.debug(f"Marked user {user} as staff due to superuser status")
```
jnovinger: Hopefully it's just my lack of familiarity with our use of `RemoteUserBackend`, but I don't understand why we need to do this (temporarily setting `user.is_staff`) here, now.

jnovinger (commented):
Had one question about why were still setting `user.is_staff` in `RemoteUserAuth`, but this looks good otherwise.

netbox/netbox/authentication/__init__.py
jeremystretch: Honestly I'm not sure that we do, but I figured it might be less disruptive to keep it. That said, I'm okay removing it for v4.5.

jnovinger: I'm +0 on removing it. Happy to defer if you think it's best to leave it, but otherwise it's just one more thing for us to remember to get back to later.

Going to approve this PR and let you decide.

jnovinger (approved):

PR #20307, opened 2025-09-09
Format timestamp in script results table

### LINKED ISSUE #16381: User preference for time format in script logs
### NetBox version

v4.0.3

### Feature type

### Proposed functionality

I'd like an option to format the timestamps on the `/extras/scripts/results/12345/` page like `2024-06-02 22:26:50`.  Currently it looks like `2024-06-02T22:26:50.483238+00:00`.  The timezone can be assumed to match what is shown everywhere else in the UI.

### Use case

I use the log output in custom scripts quite a bit, to display important information to the user when running in a browser.  I like the new feature in v4 to display timestamps on this page, but it occupies too much screen real estate.

### Database changes

maybe a new user preference?

### External dependencies

_No response_

llamafilm:
### Fixes: #16381

I thought about adding a try/except block in case `fromisoformat` fails to parse, but I could not think of a case where that would happen.

### DIFF
--- a/netbox/extras/tables/tables.py
+++ b/netbox/extras/tables/tables.py
@@ -725,8 +725,9 @@ class ScriptResultsTable(BaseTable):
     index = tables.Column(
         verbose_name=_('Line')
     )
-    time = tables.Column(
-        verbose_name=_('Time')
+    time = columns.DateTimeColumn(
+        verbose_name=_('Time'),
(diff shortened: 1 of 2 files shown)


arthanson (approved):

arthanson:
Thanks @llamafilm

PR #20315, opened 2025-09-10
Fixes #20236: Improve file naming and upload handling

pheus:
### Fixes: #20236

**Summary**
Silently normalize the **on‑disk filename** derived from image attachments to prevent nested directories and path traversal, while preserving the **user‑entered `name`** in the database/UI unchanged.

**What’s changed**
- In `upload_to`:
  - Normalize browser paths (e.g. `C:\fakepath\photo.jpg`) and extract the real uploaded extension.
  - Use the user‑entered `name` when present; sanitize via Django’s filename utility.
  - Prefix with `{object_type.model}_{object_id}` and validate the final relative path with `validate_file_name(..., allow_relative_path=True)`.
  - Preserve only known image extensions (`bmp`, `gif`, `jpeg`, `jpg`, `png`, `webp`); otherwise save without an extension.
- (Hardening) Guard filename parsing to avoid `IndexError` if older files don’t follow the expected underscore pattern.

**Behavior before**
- Supplying slashes in the `name` could create subdirectories and trigger `IndexError` in code that assumes a fixed filename layout.

**Behavior after**
- Files are always stored under `image-attachments/` with a flat, safe basename.  
- The value shown/stored in the `name` field is **not altered**; only the on‑disk filename is normalized.

**Notes for reviewers**
- This follows the maintainer guidance to **silently normalize** rather than reject slashes.
- No user‑visible changes to the `name` field or forms.


jnovinger (commented):
Hey, @pheus , this looks great and achieves exactly what the original issue and discussion required. I do have one ask, though, which is to include some minimal testing of `ImageAttachment.filename` and `image_upload` to establish future expectations.

Having said that, I think it's important that we ship this in the upcoming v4.4.1 release coming next Tuesday. So, if there's not enough time to add the tests before then, we can merge this as is and add tests as a follow up. Thoughts?

pheus:
Thanks for the quick review and for aiming this at v4.4.1. This is much appreciated!

I’ll add a minimal test and aim to push these by Monday to leave time for review. If timing gets tight, I’m fine merging as-is and following up with a tests-only PR.

pheus:
I’ve added minimal tests covering:

- `image_upload` normalization (slashes/backslashes; no nested paths) and extension handling
- `ImageAttachment.filename` prefix handling and legacy names (no `IndexError`)
- Windows-style inputs (`C:\fake_path\...`)

I hope this aligns with what you had in mind. Thanks for the review and guidance. And for maintaining NetBox.

jnovinger (approved):
Thanks, @pheus . Those are exactly the sorts of tests I had in mind.

PR #20323, opened 2025-09-11
#20206: Clarify `django-storages` configuration from env vars

### LINKED ISSUE #20206: Document NetBox configuration approach with environment variables and third-party libraries
### Change Type

Addition

### Area

Configuration

### Proposed Changes

Add guidance explaining NetBox's configuration approach and why environment variables don't work as expected with third-party libraries like `django-storage`s. Many Django libraries expect to use environment variable patterns, but NetBox's explicit configuration approach might affect how these libraries function.

Issue Context: Users following third-party library documentation expect environment variables to be automatically detected. This leads to confusion when standard patterns don't work as documented.

Documentation Updates Needed:

1. In `docs/configuration/index.md`: explain NetBox's explicit configuration approach:
    - NetBox uses explicit Python configuration rather than environment variable detection
    - Benefits: Clear configuration, version control, explicit parameter definition
    - Impacts: Third-party libraries that rely on automatic environment variable detection require explicit configuration
2. Environment Variable Usage Patterns: Document how to use environment variables in NetBox:
    

    ```py
    # Explicitly reading environment variables when needed
    import os

    SECRET_KEY = os.environ.get('NETBOX_SECRET_KEY')

    # Third-party library configuration requires explicit parameter passing
    STORAGES = {
        'staticfiles': {
            'BACKEND': 'storages.backends.s3.S3Storage',
            'OPTIONS': {
                'bucket_name': os.environ.get('AWS_STORAGE_BUCKET_NAME'),
              'location': 'static/',
            }
        }
    }
    ```

3. Common Third-Party Integration Gotchas: Document that libraries expecting automatic environment variable detection need explicit configuration in NetBox context

This addresses recurring support questions about why documented patterns from Django libraries don't work in NetBox and provides clear guidance on NetBox's configuration approach.

jnovinger:
### Closes: #20206

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -294,7 +294,7 @@ STORAGES = {
 
 This approach works because the environment variables are resolved during NetBox's configuration processing, before the third-party library attempts its own environment variable detection.
 
-!!! warning "Common Gotcha"
+!!! warning "Configuration Behavior"
     Simply setting environment variables like `AWS_STORAGE_BUCKET_NAME` without explicitly reading them in your configuration will not work. The variables must be read using `os.environ.get()` within your `configuration.py` file.
 
 ---
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -257,6 +257,46 @@ The specific configuration settings for each storage backend can be found in the
 !!! note
     Any keys defined in the `STORAGES` configuration parameter replace those in the default configuration. It is only necessary to define keys within the `STORAGES` for the specific backend(s) you wish to configure.
 
+### Environment Variables and Third-Party Libraries
+
+NetBox uses an explicit Python configuration approach rather than automatic environment variable detection. While this provides clear configuration management and version control capabilities, it affects how some third-party libraries like `django-storages` function within NetBox's context.
+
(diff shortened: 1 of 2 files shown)


docs/configuration/system.md
```diff
@@ -257,6 +257,46 @@ The specific configuration settings for each storage backend can be found in the
...
+            'endpoint_url': os.environ.get('AWS_S3_ENDPOINT_URL'),
+            'location': 'static/',
+        }
+    },
+}
+```
+
+This approach works because the environment variables are resolved during NetBox's configuration processing, before the third-party library attempts its own environment variable detection.
+
+!!! warning "Common Gotcha"
```
bctiemann: While I am all for breaking down formalities, I think this would be the first time we've used a phrase like "Common Gotcha" as a warning label. Precedent has always been for a descriptive noun about the subject or failing that, "Important".

I don't really object to this per se, just calling out that it would be a bit novel compared to something like "Important" or just leaving it unlabeled.

bctiemann (dismissed):
Approved but with one piece of feedback that you may or may not want to act on

docs/configuration/system.md
jnovinger: Hehe, no, I think you're right. I think by that point I was just in the flow of writing and didn't even realize I did it. Thanks.

bctiemann (approved):

Issue #20029, opened 2025-08-06
Include complete representation of object type in outbound webhook payload


jeremystretch:
### NetBox version

v4.3.5

### Feature type

New functionality

### Proposed functionality

Add new `object_type` and `object_id` fields to the default JSON payload of outbound webhooks:

```json
{
    "event": "updated",
    "timestamp": "2025-08-05T14:24:22.300153+00:00",
    "object_type": "dcim.site",
    "object_id": 123,
    "model": "site",
    "username": "admin",
    ...
}
```

Note: This is not a breaking change, as the currently `model` field will be retained to ensure backward compatibility.

### Use case

Currently, only the bare model name (e.g. `"site"`) is included. This is ambiguous and represents an opportunity for naming collisions. Additionally, providing the object ID at the JSON data root simplifies the process of extracting the data, as it obviates the need to also parse the `data` object.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
Blocked by FR #20003

PR #20325, opened 2025-09-11
Closes #20088: Remove `model` from webhook context (replaced by `object_type`)

jeremystretch:
### Closes: #20088

The fully-qualified key `object_type` was added in NetBox v4.4 under FR #20029 to supersede `model`, which is prone to naming collisions.


arthanson (changes requested):
Couple test failures.

arthanson (approved):

Issue #5387, opened 2020-11-29
Showing config context with multiple tags assigned fails with MultipleObjectsReturned


krombel:
### Environment
* Python version: 3.8.6
* NetBox version: 2.9.10

### Steps to Reproduce
1. create a virtual machine
2. add two tags (which result in adding data to config context)
3. Open Config context of that VM

### Expected Behavior
See config context

### Observed Behavior
See an error
```
<class 'virtualization.models.VirtualMachine.MultipleObjectsReturned'>

get() returned more than one VirtualMachine -- it returned 2!
```
```
netbox_1         | Internal Server Error: /virtualization/virtual-machines/70/config-context/
netbox_1         | Traceback (most recent call last):
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
netbox_1         |     response = get_response(request)
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 179, in _get_response
netbox_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/views/generic/base.py", line 73, in view
netbox_1         |     return self.dispatch(request, *args, **kwargs)
netbox_1         |   File "/opt/netbox/netbox/utilities/views.py", line 124, in dispatch
netbox_1         |     return super().dispatch(request, *args, **kwargs)
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/views/generic/base.py", line 101, in dispatch
netbox_1         |     return handler(request, *args, **kwargs)
netbox_1         |   File "/opt/netbox/netbox/extras/views.py", line 146, in get
netbox_1         |     obj = get_object_or_404(self.queryset, pk=pk)
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/shortcuts.py", line 76, in get_object_or_404
netbox_1         |     return queryset.get(*args, **kwargs)
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/cacheops/query.py", line 353, in get
netbox_1         |     return qs._no_monkey.get(qs, *args, **kwargs)
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 433, in get
netbox_1         |     raise self.model.MultipleObjectsReturned(
[...]

Issue #5314, opened 2020-11-06
Config Context error in Virtual Machine and Device


maximebrun:
### Environment
* Python version: 3.7.9 (from docker image python:3.7-alpine)
* NetBox version: 2.9.8 (migrate from 2.6.9)

### Steps to Reproduce
1. Create a VM.
2. Add 2 tags to the VM previously created.
3. Go to Config Context tab of this VM.

### Expected Behavior
Display config context information.

### Observed Behavior
An Error message :
```
MultipleObjectsReturned at /virtualization/virtual-machines/399/config-context/

get() returned more than one VirtualMachine -- it returned 2!
```
2 is the number of tag attached to the VM. A vm with 4 tags will return a error message "-- it returned 4!".

### Comment
Problem exist with virtual machine and device.

I'm not very confident it is the way to really fix the issue but perhaps it can help :

Problem seems to be in ConfigContextModelQuerySet class. Adding a distinct in queryset of class VirtualMachineConfigContextView fix the error (I don't use config context so i can't test more).

```
 class VirtualMachineConfigContextView(ObjectConfigContextView):
-    queryset = VirtualMachine.objects.annotate_config_context_data()
+    queryset = VirtualMachine.objects.distinct().annotate_config_context_data()
     base_template = 'virtualization/virtualmachine.html'
```

krombel:
We are running on v2.9.9 now but still have the issue:
```
<class 'virtualization.models.VirtualMachine.MultipleObjectsReturned'>

get() returned more than one VirtualMachine -- it returned 2!

Python version: 3.8.6
NetBox version: 2.9.9
```
This is the stack trace:
```
netbox_1         | Internal Server Error: /virtualization/virtual-machines/11/config-context/
netbox_1         | Traceback (most recent call last):
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
netbox_1         |     response = get_response(request)
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 179, in _get_response
netbox_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
netbox_1         |   File "/usr/local/lib/python3.8/site-packages/django/views/generic/base.py", line 73, in view
netbox_1         |     return self.dispatch(request, *args, **kwargs)
netbox_1         |   File "/opt/netbox/netbox/utilities/views.py", line 124, in dispatch
[...]

PR #20346, opened 2025-09-12
#20327: Device queries now faster when including ConfigContexts

jnovinger:
### Fixes: #20327

Move `.distinct()` from main queryset to tag subquery to eliminate performance bottleneck when querying devices with config contexts.

The `.distinct()` call on the main device queryset was causing PostgreSQL to sort all devices before pagination, resulting in 15x slower API responses for large installations (10k+ devices, 100+ config contexts).

Moving `.distinct()` to the tag subquery eliminates duplicates at their source (`GenericForeignKey` tag relationships) while preserving the fix for issues #5314 and #5387 without impacting overall query performance.

This is also likely to have positive impacts in other places that make use of these queryset methods, including:
- /api/virtualization/virtual-machines
- /dcim/devices/
- /virtualization/virtual-machines/

### DIFF
--- a/netbox/extras/tests/test_models.py
+++ b/netbox/extras/tests/test_models.py
@@ -1,16 +1,14 @@
-import re
 import tempfile
 from pathlib import Path
 
 from django.contrib.contenttypes.models import ContentType
 from django.core.files.uploadedfile import SimpleUploadedFile
-from django.db import connection
 from django.forms import ValidationError
 from django.test import tag, TestCase
 
 from core.models import DataSource, ObjectType
 from dcim.models import Device, DeviceRole, DeviceType, Location, Manufacturer, Platform, Region, Site, SiteGroup
-from extras.models import ConfigContext, ConfigContextProfile, ConfigTemplate, ImageAttachment, Tag
+from extras.models import ConfigContext, ConfigContextProfile, ConfigTemplate, ImageAttachment, Tag, TaggedItem
 from tenancy.models import Tenant, TenantGroup
 from utilities.exceptions import AbortRequest
 from virtualization.models import Cluster, ClusterGroup, ClusterType, VirtualMachine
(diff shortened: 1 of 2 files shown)


netbox/extras/tests/test_models.py
```diff
@@ -458,6 +460,34 @@ def test_virtualmachine_site_context(self):
...
+
+        device.local_context_data = False
+        with self.assertRaises(ValidationError):
+            device.clean()
+
+        device.local_context_data = 'foo'
+        with self.assertRaises(ValidationError):
+            device.clean()
+
+    @tag('regression')
```
jnovinger, line 566: Just rearranging the test to keep all of the regression tests together at the end of the test class.

netbox/extras/tests/test_models.py
```diff
@@ -543,32 +574,35 @@ def test_multiple_tags_return_distinct_objects_with_seperate_config_contexts(sel
...
+        for query in device_queries:
+            # Main device query should NOT use DISTINCT
+            self.assertNotIn('SELECT DISTINCT "dcim_device"', query)
+
+            # Tag sub-query _should_ use DISTINCT
+            tag_distinct_pattern = re.compile(
+                r'SELECT DISTINCT \w+\."tag_id".*FROM "extras_taggeditem" \w+ INNER JOIN "django_content_type"'
+            )
+            distinct_tag_pattern_found = any(tag_distinct_pattern.search(query) for query in device_queries)
+            self.assertTrue(distinct_tag_pattern_found)
```
jnovinger: There may very well be better ways to assert the thing we want here. I'm open to ideas.

netbox/extras/tests/test_models.py
```diff
@@ -621,32 +652,35 @@ def test_multiple_tags_return_distinct_objects_with_separate_config_contexts(sel
...
+        Verifies that DISTINCT is only used in tag subquery where needed,
+        not on the main device query which is expensive for large datasets.
+        """
+        # Use existing test device from ConfigContextTest setup
         device = Device.objects.first()
 
-        device.local_context_data = ""
-        with self.assertRaises(ValidationError):
-            device.clean()
+        # connection.queries_log.clear()
```
jeremystretch: Can we delete this?

jeremystretch (dismissed):

netbox/extras/tests/test_models.py
jeremystretch: IIRC you can check for the application of `distinct()` on the parent query, but I'm not sure about the subquery. IMO this is fine as-is; we can adjust it in the future if needed.

fabi125:
I tested this fix with our dataset, and a query that returns ~16k devices still took ~24s to finish, while my workaround in  #20327 finishes in ~0.5s. Returned data is exactly the same.

netbox/extras/tests/test_models.py
jnovinger: Doh, thanks.

jeremystretch:
@fabi125 unfortunately you're proposed change deviates from supported patterns, and addresses only the REST API. We should attempt to devise a solution which addresses the inefficiency within the queryset itself.

fabi125:
> @fabi125 unfortunately you're proposed change deviates from supported patterns, and addresses only the REST API. We should attempt to devise a solution which addresses the inefficiency within the queryset itself.

Of course! Fully supportive of fixing this in a proper and supported way instead of my hack :)

jnovinger:
~~@fabi125 , thanks again for the bug report and testing this fix! Since the subquery is actually part of the where clause, I'm not sure that the double-query approach you mentioned would actually return the correct results, since the first query used to gather the PKs would not have the annotated fields to filter against.~~

My apologies! In retrospect, I was clearly confusing the WHERE clauses of the main query and the subquery used to retrieve the config contexts. Please disregard the above.

jeremystretch (approved):

PR #20350, opened 2025-09-13
add global search hotkey

### LINKED ISSUE #20308: Global search hotkey shortcut
### NetBox version

v4.4.0

### Feature type

New functionality

### Proposed functionality

Pressing the slash `/` key from any webpage should move the cursor to the global search box, unless another input element already has focus. This is a common convention across many web services including Gmail and GitHub.

This is a duplicate of #17824 which was auto-closed because I didn't have time to work on it, but now I do.

Also very similar to #12749 which was accepted but also got auto-closed.

### Use case

Nine times out of 10, the first thing I do when opening Netbox is click on global search. I've arrived the page using the keyboard (typing the URL), so it's un-ergonomic to move to the mouse for one click and then move back to the keyboard to type.

### Database changes

_No response_

### External dependencies

_No response_

llamafilm:
### Fixes: #20308

Add hotkeys typescript module to handle `/` keypress, jumping focus to the global search input.


jnovinger:
@llamafilm , please update the PR to address the merge conflicts noted above.

llamafilm:
I didn't see any merge conflicts except for the generated JS files.  But I spotted a mistake and just pushed a commit, let me know if you're still seeing any conflict.

bctiemann (approved):

bctiemann:
I can't see the source of the conflict either, but this is clean now and works well. Thanks @llamafilm !

PR #20351, opened 2025-09-14
Fixes #20342: Override create_superuser to drop is_staff

pheus:
### Fixes: #20342

Override `UserManager.create_superuser()` to strip `is_staff` from `extra_fields` and enforce `is_superuser=True`, fixing the `TypeError` during `createsuperuser` with the custom `User` model.


jeremystretch (approved):
Thanks for adding `acreate_user()` as well.

Reviewing this, I caught another oversight in the initial implementation: We neglected to redefine `alters_data = True` on the overridden user creation methods. I've just added them in a separate commit.

PR #20354, opened 2025-09-14
Closes #19262: Remove FHRP IP prefix constraint

### LINKED ISSUE #19262: Disassociate between interface FHRP group assignment options and interface IP address
### NetBox version

v4.2.5

### Feature type

### Proposed functionality

Existing functionality :  When trying to assign FHRP groups to interfaces, only FHRP groups with a related IP address assigned are displayed as options. (I didn't really read the code, but from the system behaviour, "related" means the FHRP group must be assigned an IP address of the same subnet prefix as the interface IP)

Proposed functionality : When trying to assign FHRP groups to interfaces, all FHRP groups should be displayed as options regardless of any relationship between the FHRP group IP and interface IP

There isn't a need to worry about being too hard to find the right FHRP group when hundreds of groups being displayed as mentioned in https://github.com/netbox-community/netbox/issues/8435 . The purpose of "Avoiding the user from having to select from hundreds of potential groups" is already achieved by another functionality where users are allowed to type and quick search within the column.

Related issues :
https://github.com/netbox-community/netbox/issues/18975
https://github.com/netbox-community/netbox/issues/18696

### Use case

This is a removal of unnecessary existing restriction. It's not a must that each interface's IP address is under the same parent prefix as the FHRP group's VIP, there could be additional mechanisms like NAT in the middle. For example : We have a whole dedicated subnet specifically used as F5 load-balanced service VIPs, and the VM interface IPs are not from the same subnet.

There is an existing workaround; Assigning FHRP groups to the interface before assigning a physical IP, but this is not clean and not scalable. Once the interface IP is specified, it's painful when there is a need to modify its FHRP group assignments.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Fixes: #19262

- Lift the “same prefix” requirement when assigning FHRP groups to interfaces.
- Remove `FHRPGroupAssignmentForm.__init__` filtering; show all groups regardless of VIP subnet.
- Add `group_id` to the `q` filter for matching by group ID (per maintainer feedback).

**Reason:** NAT/shared VIP subnets mean interface IP and VIP prefixes can differ; type‑ahead search keeps selection manageable.

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -804,6 +804,7 @@ class FHRPGroupFilterSet(NetBoxModelFilterSet):
             return queryset
         return queryset.filter(
             Q(description__icontains=value) |
+            Q(group_id__contains=value) |
             Q(name__icontains=value)
         )
 
--- a/netbox/ipam/forms/model_forms.py
+++ b/netbox/ipam/forms/model_forms.py
@@ -580,13 +580,6 @@ class FHRPGroupAssignmentForm(forms.ModelForm):
         model = FHRPGroupAssignment
         fields = ('group', 'priority')
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
-        ipaddresses = self.instance.interface.ip_addresses.all()
-        for ipaddress in ipaddresses:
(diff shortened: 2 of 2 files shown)


jeremystretch (approved):
Thanks @pheus!

PR #20386, opened 2025-09-18
Closes #17010: Show admin navigation menu items only for staff & superusers

### LINKED ISSUE #17010: Hide Admin menu section for non-admins
### NetBox version

v4.0.8

### Feature type

### Proposed functionality

If a user is not an administrator, the `Admin > System` section of the menu should not be visible. If the user also lacks permissions for any menu item of `Admin > Authentication`, the `Admin` section should be hidden completely.

### Use case

A user with no permissions can't see *API Tokens*, *System* and *Background Tasks*. However, these menu items are shown. Hiding them makes the UI a bit cleaner.

### Database changes

None

### External dependencies

None

jeremystretch:
### Closes: #17010

- Set `staff_only=True` on all MenuItems under the Admin section
- Tweak the the logic of the `nav()` template tag so that users with `is_staff` or `is_superuser` satisfied the `staff_only` condition

### DIFF
--- a/netbox/netbox/navigation/menu.py
+++ b/netbox/netbox/navigation/menu.py
@@ -412,7 +412,7 @@ ADMIN_MENU = Menu(
                 MenuItem(
                     link='users:user_list',
                     link_text=_('Users'),
-                    auth_required=True,
+                    staff_only=True,
                     permissions=['users.view_user'],
                     buttons=(
                         MenuItemButton(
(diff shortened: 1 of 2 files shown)


arthanson:
@jeremystretch how will this play with the removal of is_staff flag for user in 4.5?  Wondering if this should just be moved to 4.5 as those changes will probably effect this?

jeremystretch:
@arthanson I think I've addressed this sufficiently (for now, at least) under #20306; `staff_only` will check for `is_superuser` instead. It might be worth tweaking the argument, but we'll need to keep support for the current one in place for a few releases anyway.

I'd like to fix #17010 in v4.4 as it _is_ a fairly annoying bug.

arthanson (approved):

PR #20387, opened 2025-09-18
Fixes #20375: Preserve filter params when performing bulk operations

jeremystretch:
### Fixes: #20375

- Add the `get_url_params()` method on ObjectAction to return a QueryDict of URL query parameters
- Provide this QueryDict as `url_params` in the button template context
- Update all button templates to reference `url_params`

### DIFF
--- a/netbox/templates/core/buttons/bulk_sync.html
+++ b/netbox/templates/core/buttons/bulk_sync.html
@@ -1,3 +1,3 @@
-<button type="submit" name="_sync" {% formaction %}="{{ url }}{% if return_url %}?return_url={{ return_url }}{% endif %}" class="btn btn-primary">
+<button type="submit" name="_sync" {% formaction %}="{{ url }}{% if url_params %}?{{ url_params.urlencode }}{% endif %}{% endif %}" class="btn btn-primary">
   <i class="mdi mdi-sync" aria-hidden="true"></i> {{ label }}
 </button>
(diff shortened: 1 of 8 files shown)


netbox/templates/core/buttons/bulk_sync.html
```diff
@@ -1,3 +1,3 @@
-<button type="submit" name="_sync" {% formaction %}="{{ url }}{% if return_url %}?return_url={{ return_url }}{% endif %}" class="btn btn-primary">
+<button type="submit" name="_sync" {% formaction %}="{{ url }}{% if url_params %}?{{ url_params.urlencode }}{% endif %}{% endif %}" class="btn btn-primary">
```
jnovinger, line 1: You've got an extra `{% endif %}` here.

jnovinger (changes requested):
Looks and acts pretty solid, except for any list view that uses the `BulkSync` action class, as noted below.

Also, I noticed that bulk rename has the inverse bug, where when you select the "select all xx <blah>s" checkbox, it only seems to act on the ones from the current page--so if pagination is at 25 and there are 37 items that match the filter, only 25 items are acted on for renaming. This appears to be the same behavior across all models.

It looks like this bulk rename issue predated this fix though, so not really a problem here. I couldn't find a bug report for that, so will file one shortly.

jnovinger (approved):

PR #20401, opened 2025-09-19
Fixes #20398: Rely on browser-native form field validation

jeremystretch:
### Fixes: #20398

- Remove the TypeScript which applies the `is-invalid` CSS class to form fields with invalid values on form submit
- Remove the call to `event.preventDefault()` to allow the native browser validation to take over
- Remove the now-obsolete generic "This field is required" warning from the form template


jnovinger (approved):
Very nice!

PR #20455, opened 2025-09-29
19615 append extra query params to static template tag

arthanson:


netbox/utilities/templatetags/builtins/tags.py
```diff
@@ -127,3 +130,38 @@ def formaction(context):
...
+    # Get the base static URL
+    static_url = static(path)
+
+    # Parse the URL to extract existing query parameters
+    parsed = urlparse(static_url)
+    existing_params = parse_qs(parsed.query)
+
+    # Add new parameters to existing ones
+    for key, value in params.items():
+        existing_params[key] = [str(value)]
```
jnovinger, line 177: What's the expectation of what should happen if a param is passed with a key that matches an existing param key?

It doesn't seem like there's a large chance of collisions with just the `v` param key, but I'm concerned that this template tag will get re-used in the future without that behavior being understood.

jnovinger (commented):
I assume we only needed to change from `{% static %}` to `{% static_with_params %}` for these specific cases in `base.html` and not elsewhere?

netbox/utilities/templatetags/builtins/tags.py
arthanson: Added logging in if dupe params

netbox/templates/base/base.html
```diff
@@ -26,7 +26,7 @@
     {# Initialize color mode #}
     <script
       type="text/javascript"
-      src="{% static 'setmode.js' %}?v={{ settings.RELEASE.version }}"
+      src="{% static_with_params 'setmode.js' v=settings.RELEASE.version %}"
```
jnovinger, line 29: > I assume we only needed to change from {% static %} to {% static_with_params %} for these specific cases in base.html and not elsewhere?

netbox/utilities/templatetags/builtins/tags.py
```diff
@@ -127,3 +132,44 @@ def formaction(context):
...
+    CDNs that already include query parameters in the URL. It properly appends additional
+    query parameters without creating double question marks.
+
+    Args:
+        path: The static file path (e.g., 'setmode.js')
+        **params: Query parameters to append (e.g., v='4.3.1')
+
+    Returns:
+        A properly formatted URL with query parameters.
+    """
```
jnovinger, line 161: Apologies for being nitpicky, but I am concerned about a couple of things relating to the maintenance of this tag over time:

1. The docstring mentions S3, but that was only the motivating use case, while the tag name is generic and invites re-use.
2. The docstring doesn't mention the duplicate key behavior, which might surprise some developers and is only noticeable at runtime.
3. Ideally we'd have tests to lock in the expected behavior.

arthanson: Not nitpicky at all.  Updated doc stings and added tests.

netbox/templates/base/base.html
arthanson: yes, static is fine to use without query params

jnovinger (approved):

PR #20473, opened 2025-10-02
RFC: Fixes #20290: Prevent ObjectType queries when table doesn't exist

### LINKED ISSUE #20290: Upgrade from v3.7.8 to v4.4.0 fails on migration `users.0005_alter_user_table` (`core_objecttype` missing)
### Deployment Type

Self-hosted

### NetBox Version

v4.4.0

### Python Version

3.10

### Steps to Reproduce

1. Start with a fresh PostgreSQL database (no existing NetBox tables).  
2. Clone the NetBox repository and check out **v3.7.8**:
   ~~~bash
   git clone https://github.com/netbox-community/netbox.git /opt/netbox
   cd /opt/netbox
   git checkout v3.7.8
   ~~~
3. Configure NetBox per the installation guide (minimal config with DB/Redis and `ALLOWED_HOSTS`) and initialize using the provided script:
   ~~~bash
   ./upgrade.sh
   ~~~
4. Upgrade the working tree to **v4.4.0**:
   ~~~bash
   git checkout v4.4.0
   ~~~
5. Run the upgrade script again to apply migrations for v4.4.0:
   ~~~bash
   ./upgrade.sh
   ~~~

### Expected Behavior

The upgrade should complete successfully. All migrations should apply without errors and NetBox should start normally on v4.4.0.

### Observed Behavior

The migration fails while applying `users.0005_alter_user_table` with an error indicating that `core_objecttype` does not exist:

~~~bash
Applying users.0005_alter_user_table...Traceback (most recent call last):
  File "/opt/netbox/venv/lib/python3.10/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox-/venv/lib/python3.10/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.UndefinedTable: relation "core_objecttype" does not exist
LINE 1: ...type"."public", "core_objecttype"."features" FROM "core_obje...
                                            ^
~~~

After this exception, the upgrade process aborts.

jnovinger:
### Fixes: #20290

In v4.4.0, `ObjectType` became a concrete model with a database table. Signal handlers query this table during migrations, but the table doesn't exist yet during 3.7.x→4.4.0 upgrades. The query failures poison the transaction and abort migrations.

This commits adds a `objecttype_table_exists()` helper and and uses it in three places:
- `has_feature()` in netbox/models/features.py
- `update_object_types()` in core/signals.py
- `Search backend cache()` in search/backends.py

If `core_objecttype` table doesn't exist, operations return early instead of querying. Once table exists, normal operation resumes.

Pros
- no migration changes
- one helper function, three identical checks
- easy to test, verify, and remove later if needed
- queries PostgreSQL system catalogs for table existence - won't poison transactions

Cons
- fixes symptoms, not root cause
- touches 3 different files instead of one surgical fix
- future code querying `ObjectType` during migrations could still fail/future developers might not know to add this check in new code
- adds table introspection overhead (could cache if needed) to every `has_feature()` call
- could introspect table name from model instead of hard-coding

### DIFF
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -13,6 +13,7 @@ from django_prometheus.models import model_deletes, model_inserts, model_updates
 from core.choices import JobStatusChoices, ObjectChangeActionChoices
 from core.events import *
 from core.models import ObjectType
+from utilities.object_types import objecttype_table_exists
 from extras.events import enqueue_event
 from extras.models import Tag
 from extras.utils import run_validators
@@ -51,6 +52,10 @@ def update_object_types(sender, **kwargs):
     """
     Create or update the corresponding ObjectType for each model within the migrated app.
     """
+    # Skip ObjectType operations if the table doesn't exist yet (during migrations)
+    if not objecttype_table_exists():
+        return
+
     for model in sender.get_models():
         app_label, model_name = model._meta.label_lower.split('.')
 
--- a/netbox/netbox/models/features.py
+++ b/netbox/netbox/models/features.py
@@ -24,6 +24,7 @@ from netbox.registry import registry
 from netbox.signals import post_clean
 from netbox.utils import register_model_feature
 from utilities.json import CustomFieldJSONEncoder
+from utilities.object_types import objecttype_table_exists
 from utilities.serialization import serialize_object
 
 __all__ = (
@@ -670,6 +671,10 @@ def has_feature(model_or_ct, feature):
     """
     Returns True if the model supports the specified feature.
     """
+    # Check if ObjectType table exists before attempting queries
+    if not objecttype_table_exists():
+        return False
+
     # If an ObjectType was passed, we can use it directly
     if type(model_or_ct) is ObjectType:
         ot = model_or_ct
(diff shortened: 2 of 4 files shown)


jnovinger:
Other approaches considered:

1. Squashed migrations with updated dependencies (per @pheus's comment in #20290) - Changes already-applied migrations, breaks 4.3.7 -> 4.4.0 upgrades
2. Earlier `ObjectType` table creation via dependency restructuring - Same issues as squashing, multiple upgrade paths make this impractical
3. Global signal disconnection during migrations - Too broad, loses legitimate functionality
4. Dual feature detection (DB + in-memory fallback) - Re-adds code we just removed in v4.4.0
5. Lazy `ObjectType` evaluation - Doesn't address direct `ObjectType` queries in `core/signals.py` and `search/backends.py`
6. Document intermediate upgrade step requirement - Still viable, could complement this fix

jeremystretch:
I've opened PR #20518 as an alternative to this solution. Its approach is very similar, but hooks into `get_for_model()` and `has_feature()` directly to avoid querying the `core_objecttype` table as necessary.

jnovinger:
Closing in favor of #20518.

PR #20477, opened 2025-10-02
Closes #20210: Implement new version of API token

### LINKED ISSUE #20210: New token authentication mechanism
### NetBox version

v4.3.7

### Feature type

New functionality

### Proposed functionality

This FR proposes a new type of token for authenticating REST & GraphQL API requests. Unlike current tokens, which are monolithic strings, the new format includes both a unique public ID and a secret plaintext in the format `<id>.<token>`.

#### 1. Extend the `Token` model

* Add a `version` PositiveSmallIntegerField to track the token version. Legacy tokens are v1; new, hashed tokens are v2.
* Add a nullable `pepper` PositiveSmallIntegerField to track the pepper used to generate to hash (see below).
* Add a nullable `key` fixed-length CharField. This will hold the alphanumeric ID for v2 tokens.
* Add a nullable `diegst` CharField. This will hold the SHA-256 digest for v2 tokens.

#### 2. Extend `TokenAuthentication` backend

* Update the `TokenAuthentication` class to support two header formats:
  * `Authorization: Token ...` will be retained for v1 tokens.
  * The new format `Authorization: Bearer ...` will be used for v2 tokens.

#### 3. Introduce a configuration parameter to store versions peppers

* Introduce the `API_TOKEN_PEPPERS` configuration parameter to store versioned peppers.
  * Peppers will be versioned to allow rotating token hashes over time.
  * Pepper values can be populated either from the configuration file or from environment variables.

#### Token creation

The process for creating a token is as follows:

1. A new random token ID and plaintext are generated.
2. The plaintext is hashed along with the currently configured global pepper.
3. The ID and digest are stored in the database.
4. The plaintext is immediately returned to the user. This is the only time the plaintext token is sent from the server.

#### Token authentication

1. An API request is sent to the server with the header `Authorization: Bearer <id>.<token>`.
2. The token (if one exists) is retrieved from the database by its ID.
3. The presented plaintext is hashed along with the pepper specified by the database record.
4. The resulting digest is compared to the stored digest. If they are the same, authentication passes.

#### Token retrieval

As with user passwords, the retrieval of existing v2 tokens is not possible, as only cryptographic digests are stored.

All other token features (expiration times, IP restrictions, etc.) will continue to be supported for both v1 and v2 tokens. Support for v1 tokens will be discontinued in a future NetBox release.

### Use case

This new generation of API tokens provides a migration path away from storing plaintext token values in the database (see, for example, #20125), while retaining backward compatibility for legacy API tokens. The employment of HMAC-SHA-256 with peppers stored outside the database embody current best practices with regard to secure token storage. Additionally, the move to the `Bearer` header format ensures compliance with [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1).

### Database changes

Add the `version`, `pepper`, `key`, and `digest` fields to the Token model as described above. No modification to existing fields is anticipated.

### External dependencies

N/A

jeremystretch:
### Closes: #20210

- Extend the Token model
  - Rename `key` to `plaintext` (used only for v1 tokens)
  - Add an integer field `version` to denote v1/v2 tokens
  - Add `key`, `pepper_id`, and `hmac_digest` fields to store v2 token attributes
- Introduce the `API_TOKEN_PEPPERS` configuration parameter
- Extend `TokenAuthentication` to recognize both token versions and to validate v2 tokens
- Introduce the `get_current_pepper()` utility function to retrieve the most recent pepper
- Simplify the `token.html` template for the account view
- Adapt all test suites to use v2 API tokens by default
- Update the documentation


netbox/netbox/tables/columns.py
```diff
@@ -270,7 +270,7 @@ def render(self, record, table, **kwargs):
         if not (self.actions or self.extra_buttons):
             return ''
         # Skip dummy records (e.g. available VLANs or IP ranges replacing individual IPs)
-        if type(record) is not model or not getattr(record, 'pk', None):
+        if not isinstance(record, model) or not getattr(record, 'pk', None):
```
jeremystretch, line 273: Fix this to work with proxy models as well as concrete. (Needed to support UserToken)

jeremystretch:
Moving this back to draft for now. Going to work on adding a token prefix as well.

bctiemann (approved):

PR #20478, opened 2025-10-02
Fixes #20471: Update NumericRange handling to use half-open intervals

### LINKED ISSUE #20471: Saving a VLANGroup without changes creates a new changelog entry
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.2

### Python Version

3.12

### Steps to Reproduce

1. Navigate to **IPAM → VLAN Groups**; create a VLAN group (or open an existing one).
2. In **VLAN ID Ranges**, enter a value such as `100-200, 300-300` and click **Save**.
3. Reopen the same VLAN group, **do not change any fields**, and click **Save** again.
4. Open the object’s **Changelog** tab (or **Other → Change Log**) and filter by this object.

### Expected Behavior

No new changelog entry should be created if no fields were changed.

### Observed Behavior

A new changelog entry is recorded on each save, even when the form or API payload is identical in meaning. The difference appears isolated to `vid_ranges` and stems from representation drift:
- UI/forms (and CSV import) return **inclusive** ranges (e.g., `NumericRange(lo, hi, bounds='[]')`).
- PostgreSQL `int4range` canonicalizes to **half‑open** `[lo, hi)` on round‑trip.
- Change logging compares serialized pre/post snapshots textually, so these representations don’t match even though the set of VLAN IDs is unchanged.

pheus:
### Fixes: #20471

**Summary**

Normalize integer range handling to use half‑open intervals `[lower, upper)` across the app. Inclusive inputs are still accepted (forms/CSV/API) but are converted to a canonical half‑open form on save and serialization. This yields stable pre/post snapshots and prevents spurious change log entries when saving an unchanged `VLANGroup`.

**Changes**

- Canonicalize all `NumericRange` values for `VLANGroup.vid_ranges` to `[lo, hi)` in forms, serializers, and utilities.
- Ensure change‑logging compares normalized values to avoid no‑op diffs.
- Update tests for round‑trip stability and “no new changelog on no‑op save.”

**Notes**

- No database schema changes.
- Backward‑compatible: existing data remains valid; user‑facing behavior is unchanged.

### DIFF
--- a/netbox/netbox/api/fields.py
+++ b/netbox/netbox/api/fields.py
@@ -169,7 +169,7 @@ class IntegerRangeSerializer(serializers.Serializer):
         if type(data[0]) is not int or type(data[1]) is not int:
             raise ValidationError(_("Range boundaries must be defined as integers."))
 
-        return NumericRange(data[0], data[1], bounds='[]')
+        return NumericRange(data[0], data[1] + 1, bounds='[)')
 
     def to_representation(self, instance):
         return instance.lower, instance.upper - 1
--- a/netbox/utilities/data.py
+++ b/netbox/utilities/data.py
@@ -152,9 +152,16 @@ def ranges_to_string(ranges):
 
 def string_to_ranges(value):
     """
-    Given a string in the format "1-100, 200-300" return an list of NumericRanges. Intended for use with ArrayField.
-    For example:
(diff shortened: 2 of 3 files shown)


arthanson (approved):
Thanks @pheus

PR #20489, opened 2025-10-03
Improved docs on background jobs on instances

### LINKED ISSUE #20488: Mention, that the JobsMixin is required for running jobs on a model instance
### Change Type

Addition

### Area

Plugins

### Proposed Changes

If a Job is executed with an object passed in the instance variable a Validation error is raised, stating

> Jobs cannot be assigned to this object type $OBJECTTYPE

All objects that have to be passed to a job need the JobsMixin. This is not clearly stated in the Plugin Development documentation.

991jo:

### DIFF
--- a/docs/plugins/development/background-jobs.md
+++ b/docs/plugins/development/background-jobs.md
@@ -60,6 +60,13 @@ Four of the standard Python logging levels are supported:
 
 Log entries recorded using the runner's logger will be saved in the job's log in the database in addition to being processed by other [system logging handlers](../../configuration/system.md#logging).
 
+### Jobs running for Model instances
+
+A Job can be run for a certain instance of a Model.
+To support this, the `JobsMixin` is required for this model.
+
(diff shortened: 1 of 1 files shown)


arthanson (changes requested):
@991jo can you please provide an example for this, also lower in the file under Scheduled Jobs the example there does enqueue with instance=self and not derived from JobsMixin which either the example is wrong or the wording here is not clear - can you please take a look.

Also, here is a bit reworded description that might be a bit clearer?

A Job can be executed for a specific instance of a Model. To enable this functionality, the model must include the JobsMixin.

When enqueuing a Job, you can associate it with a particular instance by passing that instance to the instance parameter.

991jo:
Yes, the JobsMixin is missing in the example. I added it there.

I think your description is good. I replaced mine with yours.

arthanson (approved):
Thanks @991jo

PR #20500, opened 2025-10-04
Documented ObjectListView quick search feature for plugins

### LINKED ISSUE #20499: The Plugins development documentation does not explain the ObjectListView quick search function.
### Change Type

Addition

### Area

Plugins

### Proposed Changes

The documentation never mentions how the "Quick Search" Field in the ObjectListView works.

991jo:
### Fixes: #20499


docs/plugins/development/filtersets.md
```diff
@@ -55,6 +55,28 @@ class MyModelViewSet(...):
     filterset_class = filtersets.MyModelFilterSet
 ```
 
+### Search function for use by the ObjectListView quick search
```
jnovinger:
```suggestion
### Implementing Quick Search
```

docs/plugins/development/filtersets.md
```diff
@@ -55,6 +55,28 @@ class MyModelViewSet(...):
...
+    ...
+    def search(self, queryset, name, value):
+        if not value.strip():
+            return queryset
+        return queryset.filter(
+            Q(name__icontains=value) |
+            Q(description__icontains=value)
+        )
+```
+
```
jnovinger, line 76: Some things that I think are worth mentioning:

1. `NetBoxModelFilterSet.search()` is already implemented and just returns the queryset unmodified, which is why it needs to be over-ridden here.
2. For context, the `search()` method is called as a result of the `q` filter, which is already wire up in `NetBoxModelFilterSet` and does not need to be over-ridden here unless different behavior is required.

docs/plugins/development/filtersets.md
```diff
@@ -55,6 +55,28 @@ class MyModelViewSet(...):
     filterset_class = filtersets.MyModelFilterSet
 ```
 
+### Search function for use by the ObjectListView quick search
+
+The `ObjectListView` has a field called Quick Search.
+For this field to work the corresponding FilterSet has to implement the `search` function.
```
jnovinger:
```suggestion
For Quick Search to work the corresponding FilterSet has to implement the `search` method.
```

jnovinger (changes requested):
Thanks for the doc additions, @991jo . It's much appreciated!

I made some notes here and below:

1. Let's clarify that the "Quick Search" field is the UI field, not the field in the filter or the method name.
2. Your sentences appear to all be on a separate line, is there some reason to not have them on the same line and let MkDocs figure out the wrapping?

991jo:
Regarding each sentence being on it's own line: That's mainly a convention I am sticking to, because it makes changes in diffs clearer, because you only get 1 sentence highlighted instead of the whole paragraph.

Regarding the control of linebreaks: It is complicated. This is primarily an issue of Markdown. CommonMark (an attempt at a markdown spec) seperates a paragraph only if there is a blank line inbetween (see https://spec.commonmark.org/0.31.2/#paragraphs ).
There could be explicit line breaks with two spaces at the end of the line (https://spec.commonmark.org/0.31.2/#hard-line-breaks)

For me this generates one `<p>` elememt:
```html
<p>The <code>ObjectListView</code> has a field called Quick Search.
For Quick Search to work the corresponding FilterSet has to implement the <code>search</code> method.
This function takes a queryset and can perform arbitrary operations on it and return it.
A common use-case is to search for the given search value in multiple fields:</p>
```

-> there is no special line break handling in mkdocs afaik and it's all HTML or whatever the other output options of mkdocs are.
However I am also fine with putting the whole paragraph on one line to match the style of the remaining document.

docs/plugins/development/filtersets.md
991jo: I incorporated your comments in [6593f1b](https://github.com/netbox-community/netbox/pull/20500/commits/6593f1b4e0f4c2d042b31095ed7858a8d7932c1a)

jnovinger (changes requested):
I understand the diff-clarity rationale for semantic line breaks, but we do need to maintain consistency with the existing documentation style.

991jo:
Okay, I updated the file accordingly.

jnovinger (approved):
Thank you, @991jo!

PR #20502, opened 2025-10-04
Fixed wrong link in plugin filtersets documentation

### LINKED ISSUE #20501: The link to django-filter in the plugin dev docs is wrong.
### Change Type

Correction

### Area

Plugins

### Proposed Changes

The link to django-filter in the plugin dev docs is wrong.

991jo:


jeremystretch:
The link is correct. You're confusing two separate packages: `django-filter` and `django-tables2`.

jeremystretch:
My mistake, you're right, the link should point to `django-filter` instead of `django-tables2` given the context.

jeremystretch (approved):

PR #20514, opened 2025-10-06
Fixes #20497: Add `range_contains` lookup and fix VLANGroup VID range filtering

### LINKED ISSUE #20497: Filtering `VLANGroup` by `vid_ranges.contains` in GraphQL fails with “cannot cast type integer to int4range”
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.2

### Python Version

3.12

### Steps to Reproduce

Filtering `vlan_group_list` by a scalar VLAN ID using the `vid_ranges.contains` GraphQL filter returns an error instead of matching groups whose VID ranges include the value.

1. Ensure a `VLANGroup` exists with `vid_ranges` that include VLAN 10 (e.g., `[1,11)`).
2. Run the GraphQL query below against `/graphql/`.
3. The request fails with the error shown instead of returning the matching `VLANGroup`.

```graphql
query MyQuery {
  vlan_group_list(filters: { vid_ranges: { contains: 10 } }) {
    id
    name
    vid_ranges
  }
}
```

### Expected Behavior

The query should succeed and return all `VLANGroup` records whose `vid_ranges` include the given VLAN ID (e.g., `10`), without errors.  
If no groups match, the API should return an empty list. The response should include the requested fields for each matching group.

### Observed Behavior

```json
{
  "data": null,
  "errors": [
    {
      "message": "cannot cast type integer to int4range\nLINE 1: ... WHERE \"ipam_vlangroup\".\"vid_ranges\" @> (ARRAY[10])::int4...\n                                                             ^",
      "path": ["vlan_group_list"]
    }
  ]
}
```

*Server traceback* (excerpt)
```text
django.db.utils.ProgrammingError: cannot cast type integer to int4range
LINE 1: ... "ipam_vlangroup"."vid_ranges" @> (ARRAY[10])::int4...
                                                             ^
```

pheus:
### Fixes: #20497

**Summary**
Introduce a generic ORM lookup `range_contains` for `ArrayField(RangeField)` to correctly match rows where a scalar value is contained by **any** range in the array (e.g., `VLANGroup.vid_ranges`). This replaces the raw‑SQL helper used by the VLAN group filter and aligns REST and GraphQL semantics with PostgreSQL range operators.

**Changes**
- **extras**: Add and register the `range_contains` lookup (guarded to `ArrayField(RangeField)`).
- **ipam/filtersets**: Refactor `VLANGroup` FilterSet (`contains_vid`) to use the lookup (remove raw SQL).
- **GraphQL schema**:
  - Add an `IntegerRangeArrayLookup` input and wire `VLANGroup.vid_ranges` to it.
  - Keep the public API friendly: clients use `vid_ranges: { contains: <Int> }`, which maps internally to `__range_contains`.

**Tests**
- Add ORM unit tests for the lookup (positive, negative, and boundary cases).
- Add REST FilterSet tests for `?contains_vid=<int>` (hit/miss and exclusive upper bound).

### DIFF
--- a/netbox/extras/lookups.py
+++ b/netbox/extras/lookups.py
@@ -1,7 +1,37 @@
+from django.contrib.postgres.fields import ArrayField
+from django.contrib.postgres.fields.ranges import RangeField
 from django.db.models import CharField, JSONField, Lookup
 from django.db.models.fields.json import KeyTextTransform
 
 from .fields import CachedValueField
+
+
+class RangeContains(Lookup):
+    """
+    Filter ArrayField(RangeField) columns where ANY element-range contains the scalar RHS.
+
+    Usage (ORM):
+        Model.objects.filter(<range_array_field>__range_contains=<scalar>)
+
+    Works with int4range[], int8range[], daterange[], tstzrange[], etc.
+    """
+
+    lookup_name = 'range_contains'
+
+    def as_sql(self, compiler, connection):
+        # Compile LHS (the array-of-ranges column/expression) and RHS (scalar)
+        lhs, lhs_params = self.process_lhs(compiler, connection)
+        rhs, rhs_params = self.process_rhs(compiler, connection)
+
+        # Guard: only allow ArrayField whose base_field is a PostgreSQL RangeField
+        field = getattr(self.lhs, 'output_field', None)
+        if not (isinstance(field, ArrayField) and isinstance(field.base_field, RangeField)):
+            raise TypeError('range_contains is only valid for ArrayField(RangeField) columns')
+
+        # Range-contains-element using EXISTS + UNNEST keeps the range on the LHS: r @> value
+        sql = f"EXISTS (SELECT 1 FROM unnest({lhs}) AS r WHERE r @> {rhs})"
+        params = lhs_params + rhs_params
+        return sql, params
 
 
 class Empty(Lookup):
@@ -25,7 +55,7 @@ class JSONEmpty(Lookup):
 
     A key is considered empty if it is "", null, or does not exist.
     """
-    lookup_name = "empty"
+    lookup_name = 'empty'
 
     def as_sql(self, compiler, connection):
         # self.lhs.lhs is the parent expression (could be a JSONField or another KeyTransform)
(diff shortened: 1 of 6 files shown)


pheus:
**Naming note:** I used `any_range_contains` for the lookup (ArrayField(RangeField) → “value is contained by any range”). If the team prefers a different name (e.g., `in_any_range`), I’m happy to rename it and update tests, the REST FilterSet, and the GraphQL mapping accordingly.

netbox/extras/lookups.py
```diff
@@ -1,9 +1,39 @@
...
+    """
+    Filter ArrayField(RangeField) columns where ANY element-range contains the scalar RHS.
+
+    Usage (ORM):
+        Model.objects.filter(<range_array_field>__any_range_contains=<scalar>)
+
+    Works with int4range[], int8range[], daterange[], tstzrange[], etc.
+    """
+
+    lookup_name = 'any_range_contains'
```
jeremystretch: What do you think about shortening this a bit?

```suggestion
    lookup_name = 'range_contains'
```

netbox/ipam/tests/test_filtersets.py
```diff
@@ -1724,6 +1724,15 @@ def test_contains_vid(self):
...
 
+    def test_contains_vid_upper_bound_exclusive(self):
+        # 11 is NOT in [1,11)
+        params = {'contains_vid': 12}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 1)
+
+    def test_contains_vid_miss(self):
+        params = {'contains_vid': 4095}
+        self.assertEqual(self.filterset(params, self.queryset).qs.count(), 0)
+
```
jeremystretch: IMO we should wrap these into `test_contains_vid()` above as they all test the same filter.

jeremystretch (commented):
Awesome work @pheus! Just had a couple minor suggestions.

netbox/ipam/tests/test_filtersets.py
pheus: Will do. Thanks!

netbox/extras/lookups.py
pheus: That's great, thanks for this suggestion! I will update the PR accordingly.

jeremystretch (approved):
Thanks @pheus!

PR #20530, opened 2025-10-08
Add color to PowerOutletTemplate

blackphantom39:


arthanson (changes requested):
@blackphantom39 this looks good, only minor change needed on the list view it is showing the color hex value, just needs to be updated to show the color (see screenshot).  
Monosnap Image 2025-10-09 12-53-35

arthanson (approved):

PR #20531, opened 2025-10-08
Update base.html

### LINKED ISSUE #20523: templates/account/base.html doesn't account for passwordless options other than LDAP.
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.1

### Python Version

3.10

### Steps to Reproduce

1. Log into Netbox using any SSO solution. (Keycloak SSO auth 2 was used originally.)
2. Click on profile in the top right hand corner
3. See password tab.

### Expected Behavior

For there to be no Password tab, as it makes no sense for SSO.

### Observed Behavior

There is a Password tab.

Amir-Bakar:
Update base.html to account for other cases where passwords are not used, other than LDAP. (SSO solutions, for example.)

### Change summary:
Changed the check in `templates/account/base.html` from `if not request.user.ldap_username` to `if request.user.has_usable_password` in line 21.

### DIFF
--- a/netbox/templates/account/base.html
+++ b/netbox/templates/account/base.html
@@ -18,7 +18,7 @@
     <li class="nav-item">
       <a class="nav-link{% if active_tab == 'preferences' %} active{% endif %}" href="{% url 'account:preferences' %}">{% trans "Preferences" %}</a>
     </li>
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @Amir-Bakar!

Issue #17571, opened 2024-09-23
Refine HTMX Navigation Strategy


arthanson:
### Proposed Changes

The HTMX Navigation (currently an optional setting) needs to have research and a systematic redesign as the current system was added on as individual changes to the static navigation.

### Justification

The current HTMX navigation was added on to the static navigation without a systematic design which has lead to a series of bugs.  The whole system needs to be researched and designed as a whole to prevent the current edge case bugs that are showing up.

ITJamie:
another one https://github.com/netbox-community/netbox/issues/18868

PR #20460, opened 2025-09-30
Closes #20449: Add user preferences documentation

pheus:
### Fixes: #20449

Adds a new **Features → User Preferences** page and integrates a short, actionable guide for **clearing table preferences**. Also adds cross‑links to help users find the steps from related pages.

**Changes**
- New page: `docs/netbox/features/user-preferences.md`
- Nav: add entry in `mkdocs.yml`
- Features → Customization: point Bookmarks to the new page
- Configuration → `DEFAULT_USER_PREFERENCES`: link to “Clearing table preferences”
- Development → User Preferences: link to the user‑facing guide
- Data Model → Extras → Table Configs: note distinction and link to clearing steps

PR #20537, opened 2025-10-08
#17571 - Remove HTMX navigation

bctiemann:
### Closes: #17571

Removes experimental HTMX navigation.

Note: this leaves in place the `{% formaction %}` template tag as a hook for later customization, rather than removing all usages of it from templates.

Also does not remove existing non-experimental HTMX usages in forms, notifications, etc.

### DIFF
--- a/docs/features/user-preferences.md
+++ b/docs/features/user-preferences.md
@@ -34,9 +34,6 @@ Sets the default number of rows displayed on paginated tables.
 ### Paginator placement
 Controls where pagination controls are rendered relative to a table.
 
-### HTMX navigation (experimental)
-Enables partial‑page navigation for supported views. Disable this preference if unexpected behavior is observed.
-
 ### Striped table rows
 Toggles alternating row backgrounds on tables.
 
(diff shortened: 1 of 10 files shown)


pheus:
Drive‑by suggestion _(not a reviewer/maintainer)_: In #20460 I added the docs for user preferences. Since this PR removes HTMX navigation, could we also remove it from the docs?

Source lines for reference:
https://github.com/netbox-community/netbox/blob/18a308ae3ab1f4cc3ce945f47c8ec8af1e808add/docs/features/user-preferences.md?plain=1#L37-L38

Happy to open a small docs PR if that helps. Thanks!

bctiemann:
Nice catch, thanks! I've added that to this PR.

jeremystretch (approved):
I'm on the fence about retaining the `formaction` tag. While I see potential future benefit in keeping it in place, I worry that it'll only introduce confusion in the near term.

Approving this as-is; your call whether to keep it.

bctiemann:
I'm going to go ahead and keep it as-is. We can remove it as tech debt later, meanwhile I think the comment is clear enough about the intent that a future dev ought not to be too confused.

PR #20548, opened 2025-10-09
Closes #20516: Add `ranges_to_string_list` and render VLAN Group VID ranges with `ArrayColumn`

### LINKED ISSUE #20516: Render VLAN Group VID ranges with `ArrayColumn`
### NetBox version

v4.4.2

### Feature type

### Proposed functionality

Introduce a helper that returns a **list of human‑readable range strings** so callers can control spacing and presentation without re‑implementing range logic:

- `ranges_to_text_list(ranges: Iterable[NumericRange]) -> list[str]`  
  Returns items like `["5-10", "25-30"]` using the same inclusive semantics that `ranges_to_string()` currently applies.

Refactor the existing `ranges_to_string()` implementation to build on the new helper while **preserving its current output** (comma‑separated, **no spaces**), e.g.:

```py
def ranges_to_string(ranges):
    return ",".join(ranges_to_text_list(ranges))
```

Finally, update the VLAN Group list to **render VID ranges using `ArrayColumn`** so the UI shows the standard comma‑and‑space formatting (`", "`), improving readability and aligning with other list displays.

### Use case

Today, `utilities.data.ranges_to_string()` emits no spaces between items (e.g., `"5-10,25-30"`). This is correct for some contexts, but it makes it hard to reuse when a space‑separated presentation is desired (for example, table displays or human‑facing messages). Adding `ranges_to_text_list()` lets call sites choose their own joiner, and lets tables use `ArrayColumn` (which already renders with `", "`) for consistent UI.

### Database changes

None.

### External dependencies

None.

## Proposed change (implementation notes)

1) **Add helper in `utilities/data.py`**
```py
from typing import Iterable
# Use the appropriate range type for the project (e.g., psycopg's NumericRange)
# from psycopg2.extras import NumericRange  # psycopg2
# from psycopg.types.range import Range as NumericRange  # psycopg3

def ranges_to_text_list(ranges: Iterable["NumericRange"]) -> list[str]:
    """
    Convert NumericRange values to human-friendly inclusive strings ["lo-hi", ...].
    Mirrors ranges_to_string() semantics.
    """
    if not ranges:
        return []

    items: list[str] = []
    for r in ranges:
        # Compute inclusive bounds regardless of how the DB range is stored.
        lower = r.lower if getattr(r, "lower_inc", True) else r.lower + 1
        upper = r.upper if getattr(r, "upper_inc", False) else r.upper - 1

        # If you prefer singletons like "5" instead of "5-5", uncomment:
        # if lower == upper:
        #     items.append(f"{lower}")
        # else:
        items.append(f"{lower}-{upper}")
    return items
```

2) **Refactor `ranges_to_string()`** (backward compatible)
```py
def ranges_to_string(ranges):
    if not ranges:
        return ""
    return ",".join(ranges_to_text_list(ranges))
```

3) **VLAN Group model**  
   Keep the existing string property (e.g., `vid_ranges_list`) **to avoid breaking any consumers**. Add a new list‑returning property for table use:
```py
# ipam/models/vlans.py
from utilities.data import ranges_to_text_list

@property
def vid_ranges_items(self) -> list[str]:
    return ranges_to_text_list(self.vid_ranges)
```

4) **VLAN Group table**  
   Switch the column to `ArrayColumn` but **keep the column name** `vid_ranges_list` to preserve saved table configs; point it at the new list property:
```py
# ipam/tables/vlans.py
from netbox.tables import columns

# Old:
# vid_ranges_list = tables.Column(verbose_name=_('VID Ranges'), orderable=False)

# New:
vid_ranges_list = columns.ArrayColumn(
    accessor='vid_ranges_items',
    verbose_name=_('VID Ranges'),
    orderable=False,
)
```

---

## Backwards compatibility

- `ranges_to_string()` output and signature remain unchanged (still returns a `str` with **no spaces**).
- `VLANGroup.vid_ranges_list` remains available (still a `str`) to avoid breaking any code or saved table configurations that reference its column name. The table change keeps the same column name while sourcing from a new `vid_ranges_items` list property.

---

## Additional context / references

- Background plugin work that motivated this (spacing control): `ranges_to_text_list` implementation in **netbox‑acls**:  
  https://github.com/pheus/netbox-acls/commit/da406f918ce904b1c8d2e71e2165f4b55896a19d

---

### Before / After (table rendering example)

- **Before:** `VID Ranges` → `1-99,200-299`  
- **After:** `VID Ranges` → `1-99, 200-299` (space after comma via `ArrayColumn`)

pheus:
### Fixes: #20516

**Summary**
- Add `ranges_to_string_list(ranges) -> list[str]`, which formats numeric ranges as human‑readable strings (e.g., `["1-5", "8", "10-12"]`).
- Refactor `ranges_to_string(ranges)` to delegate to the new helper:
  ```py
  if not ranges:
      return ''
  return ','.join(ranges_to_string_list(ranges))
  ```
  This preserves the existing output (comma‑separated with **no spaces**) while enabling callers to control separators when needed.
- Improve VLAN Group rendering by switching the table column to use a list accessor with `ArrayColumn`, yielding `", "` spacing in the UI without changing underlying data or the string helper’s behavior.

**Implementation notes**
- `ranges_to_string` remains fully backward‑compatible in both signature and output.
- VLAN Group table uses a list accessor (e.g., `vid_ranges_items`) with `ArrayColumn`, while keeping the existing `vid_ranges_list` string property available to avoid breaking saved table configs.

**Testing**
- Add unit tests covering:
  - Singleton ranges (`"8"`) vs. hyphen ranges (`"1-5"`)
  - `ranges_to_string_list()`

**Backwards compatibility**
- No breaking changes. Existing uses of `ranges_to_string()` and `VLANGroup.vid_ranges_list` continue to work as before.

### DIFF
--- a/netbox/ipam/models/vlans.py
+++ b/netbox/ipam/models/vlans.py
@@ -10,9 +10,9 @@ from django.utils.translation import gettext_lazy as _
 from dcim.models import Interface, Site, SiteGroup
 from ipam.choices import *
 from ipam.constants import *
-from ipam.querysets import VLANQuerySet, VLANGroupQuerySet
+from ipam.querysets import VLANGroupQuerySet, VLANQuerySet
 from netbox.models import OrganizationalModel, PrimaryModel, NetBoxModel
-from utilities.data import check_ranges_overlap, ranges_to_string
+from utilities.data import check_ranges_overlap, ranges_to_string, ranges_to_string_list
 from virtualization.models import VMInterface
 
 __all__ = (
@@ -165,7 +165,17 @@ class VLANGroup(OrganizationalModel):
         return VLAN.objects.filter(group=self).order_by('vid')
 
     @property
+    def vid_ranges_items(self):
+        """
+        Property that converts VID ranges to a list of string representations.
+        """
+        return ranges_to_string_list(self.vid_ranges)
+
+    @property
     def vid_ranges_list(self):
+        """
+        Property that converts VID ranges into a string representation.
+        """
         return ranges_to_string(self.vid_ranges)
 
 
--- a/netbox/ipam/tables/vlans.py
+++ b/netbox/ipam/tables/vlans.py
@@ -41,7 +41,8 @@ class VLANGroupTable(TenancyColumnsMixin, NetBoxTable):
         linkify=True,
         orderable=False
     )
-    vid_ranges_list = tables.Column(
+    vid_ranges_list = columns.ArrayColumn(
+        accessor='vid_ranges_items',
         verbose_name=_('VID Ranges'),
         orderable=False
     )
--- a/netbox/templates/ipam/vlangroup.html
+++ b/netbox/templates/ipam/vlangroup.html
@@ -40,7 +40,7 @@
         </tr>
         <tr>
           <th scope="row">{% trans "VLAN IDs" %}</th>
-          <td>{{ object.vid_ranges_list }}</td>
+          <td>{{ object.vid_ranges_items|join:", " }}</td>
         </tr>
         <tr>
           <th scope="row">Utilization</th>
--- a/netbox/utilities/data.py
+++ b/netbox/utilities/data.py
@@ -1,6 +1,7 @@
 import decimal
-from django.db.backends.postgresql.psycopg_any import NumericRange
 from itertools import count, groupby
+
+from django.db.backends.postgresql.psycopg_any import NumericRange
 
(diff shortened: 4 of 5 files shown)


jeremystretch (approved):
Thanks @pheus! We may want to revisit the table column at some point in the future to normalize the name, but I think this is the best we can do to avoid a breaking change.

PR #20555, opened 2025-10-13
Closes #16681: Introduce render_config permission for configuration rendering

### LINKED ISSUE #16681: Introduce a custom permission action for rendering configuration templates
### NetBox version

v4.0.5

### Feature type

### Proposed functionality

Rendering a configuration template for a device or virtual machine will require a new custom permission action, `render_config`.

Currently, rendering a configuration template via the UI does not require any permissions beyond `view` for the relevant object type (e.g. `dcim.view_device`). However, rendering a configuration template via the REST API currently requires the `add` permission (e.g. `dcim.add_device`). While the requirement for write permission makes sense for API calls (because we're accepting and processing arbitrary user data), the specific permission for creating devices does not and has been captured as a bug in #14634.

### Use case

This new permission action can be used to allow config rendering via the REST API without needing to grant users permission to create new devices/VMs.

### Database changes

None

### External dependencies

None

jnovinger:
### Closes: #16681

Add `render_config` permission action for rendering device, virtual machine, and config template configurations via the REST API.

**Changes:**
- Add permission checks to `/api/dcim/devices/{id}/render-config/`, `/api/virtualization/virtual-machines/{id}/render-config/`, and `/api/extras/config-templates/{id}/render/`
- Introduce `TokenWritePermission` to enforce token write ability independent of model permissions
- Replace POST='add' requirement with `render_config` permission
- Add test coverage and documentation

Per #16681: rendering configs requires accepting arbitrary user data (requiring write-enabled tokens), but doesn't require permission to create devices/VMs. The `render_config` permission separates these concerns.

**Note:** Optional data migration could grant `render_config` to existing users with `view` or `add` permissions to ease upgrade path. Can add if desired.

### DIFF
--- a/docs/features/configuration-rendering.md
+++ b/docs/features/configuration-rendering.md
@@ -92,8 +92,8 @@ http://netbox:8000/api/extras/config-templates/123/render/ \
 ```
 
 !!! note "Permissions"
-    Rendering configuration templates via the REST API requires the `render_config` permission for the relevant object type:
+    Rendering configuration templates via the REST API requires appropriate permissions for the relevant object type:
 
     * To render a device's configuration via `/api/dcim/devices/{id}/render-config/`, assign a permission for "DCIM > Device" with the `render_config` action
     * To render a virtual machine's configuration via `/api/virtualization/virtual-machines/{id}/render-config/`, assign a permission for "Virtualization > Virtual Machine" with the `render_config` action
-    * To render a config template directly via `/api/extras/config-templates/{id}/render/`, assign a permission for "Extras > Config Template" with the `render_config` action
+    * To render a config template directly via `/api/extras/config-templates/{id}/render/`, assign a permission for "Extras > Config Template" with the `render` action
--- a/docs/features/configuration-rendering.md
+++ b/docs/features/configuration-rendering.md
@@ -90,3 +90,6 @@ http://netbox:8000/api/extras/config-templates/123/render/ \
   "bar": 123
 }'
 ```
+
+!!! note "Permissions"
+    Rendering device or virtual machine configurations via the REST API requires the `render_config` permission for the relevant object type. For example, to render a device's configuration via `/api/dcim/devices/{id}/render-config/`, a user must be assigned a permission for the "DCIM > Device" object type with the `render_config` action.
--- a/docs/features/configuration-rendering.md
+++ b/docs/features/configuration-rendering.md
@@ -94,6 +94,6 @@ http://netbox:8000/api/extras/config-templates/123/render/ \
 !!! note "Permissions"
     Rendering configuration templates via the REST API requires appropriate permissions for the relevant object type:
 
-    * To render a device's configuration via `/api/dcim/devices/{id}/render-config/`, assign a permission for "DCIM > Device" with the `render_config` action
-    * To render a virtual machine's configuration via `/api/virtualization/virtual-machines/{id}/render-config/`, assign a permission for "Virtualization > Virtual Machine" with the `render_config` action
-    * To render a config template directly via `/api/extras/config-templates/{id}/render/`, assign a permission for "Extras > Config Template" with the `render` action
+    * To render a device's configuration via `/api/dcim/devices/{id}/render-config/`, assign a permission for "DCIM > Device" with the `render_config` action.
+    * To render a virtual machine's configuration via `/api/virtualization/virtual-machines/{id}/render-config/`, assign a permission for "Virtualization > Virtual Machine" with the `render_config` action.
+    * To render a config template directly via `/api/extras/config-templates/{id}/render/`, assign a permission for "Extras > Config Template" with the `render` action.
--- a/docs/features/configuration-rendering.md
+++ b/docs/features/configuration-rendering.md
@@ -92,4 +92,8 @@ http://netbox:8000/api/extras/config-templates/123/render/ \
 ```
 
 !!! note "Permissions"
-    Rendering device or virtual machine configurations via the REST API requires the `render_config` permission for the relevant object type. For example, to render a device's configuration via `/api/dcim/devices/{id}/render-config/`, a user must be assigned a permission for the "DCIM > Device" object type with the `render_config` action.
+    Rendering configuration templates via the REST API requires the `render_config` permission for the relevant object type:
+
+    * To render a device's configuration via `/api/dcim/devices/{id}/render-config/`, assign a permission for "DCIM > Device" with the `render_config` action
+    * To render a virtual machine's configuration via `/api/virtualization/virtual-machines/{id}/render-config/`, assign a permission for "Virtualization > Virtual Machine" with the `render_config` action
+    * To render a config template directly via `/api/extras/config-templates/{id}/render/`, assign a permission for "Extras > Config Template" with the `render_config` action
--- a/netbox/dcim/tests/test_api.py
+++ b/netbox/dcim/tests/test_api.py
@@ -13,7 +13,8 @@ from ipam.choices import VLANQinQRoleChoices
 from ipam.models import ASN, RIR, VLAN, VRF
 from netbox.api.serializers import GenericObjectSerializer
 from tenancy.models import Tenant
-from users.models import User
+from users.constants import TOKEN_PREFIX
+from users.models import Token, User
 from utilities.testing import APITestCase, APIViewTestCases, create_test_device, disable_logging
 from virtualization.models import Cluster, ClusterType
 from wireless.choices import WirelessChannelChoices
@@ -1485,8 +1486,8 @@ class DeviceTest(APIViewTestCases.APIViewTestCase):
         device.config_template = configtemplate
         device.save()
 
-        self.add_permissions('dcim.render_config_device', 'dcim.view_device', 'extras.view_configtemplate')
-        url = reverse('dcim-api:device-detail', kwargs={'pk': device.pk}) + 'render-config/'
+        self.add_permissions('dcim.render_config_device', 'dcim.view_device')
+        url = reverse('dcim-api:device-render-config', kwargs={'pk': device.pk})
         response = self.client.post(url, {}, format='json', **self.header)
         self.assertHttpStatus(response, status.HTTP_200_OK)
         self.assertEqual(response.data['content'], f'Config for device {device.name}')
@@ -1502,9 +1503,40 @@ class DeviceTest(APIViewTestCases.APIViewTestCase):
         device.save()
 
         # No permissions added - user has no render_config permission
-        url = reverse('dcim-api:device-detail', kwargs={'pk': device.pk}) + 'render-config/'
+        url = reverse('dcim-api:device-render-config', kwargs={'pk': device.pk})
         response = self.client.post(url, {}, format='json', **self.header)
         self.assertHttpStatus(response, status.HTTP_404_NOT_FOUND)
+
+    def test_render_config_token_write_enabled(self):
+        configtemplate = ConfigTemplate.objects.create(
+            name='Config Template 1',
+            template_code='Config for device {{ device.name }}'
+        )
+
+        device = Device.objects.first()
+        device.config_template = configtemplate
+        device.save()
+
+        self.add_permissions('dcim.render_config_device', 'dcim.view_device')
+        url = reverse('dcim-api:device-render-config', kwargs={'pk': device.pk})
+
+        # Request without token auth should fail with PermissionDenied
+        response = self.client.post(url, {}, format='json')
+        self.assertHttpStatus(response, status.HTTP_403_FORBIDDEN)
+
+        # Create token with write_enabled=False
+        token = Token.objects.create(version=2, user=self.user, write_enabled=False)
+        token_header = f'Bearer {TOKEN_PREFIX}{token.key}.{token.token}'
+
+        # Request with write-disabled token should fail
+        response = self.client.post(url, {}, format='json', HTTP_AUTHORIZATION=token_header)
+        self.assertHttpStatus(response, status.HTTP_403_FORBIDDEN)
+
+        # Enable write and retry
+        token.write_enabled = True
+        token.save()
+        response = self.client.post(url, {}, format='json', HTTP_AUTHORIZATION=token_header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
 
 
 class ModuleTest(APIViewTestCases.APIViewTestCase):
--- a/netbox/dcim/tests/test_api.py
+++ b/netbox/dcim/tests/test_api.py
@@ -1487,7 +1487,6 @@ class DeviceTest(APIViewTestCases.APIViewTestCase):
         device.save()
 
         self.add_permissions('dcim.render_config_device')
-        self.add_permissions('dcim.add_device')
         url = reverse('dcim-api:device-detail', kwargs={'pk': device.pk}) + 'render-config/'
         response = self.client.post(url, {}, format='json', **self.header)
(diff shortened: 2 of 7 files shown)


netbox/extras/api/views.py
```diff
@@ -238,13 +240,26 @@ class ConfigTemplateViewSet(SyncedDataMixin, ConfigTemplateRenderMixin, NetBoxMo
     serializer_class = serializers.ConfigTemplateSerializer
     filterset_class = filtersets.ConfigTemplateFilterSet
 
+    def get_permissions(self):
+        # For render action, check only token write ability (not model permissions)
+        if self.action == 'render':
+            return [TokenWritePermission()]
```
jeremystretch, line 244: I think we still need to enforce view permissions for the ConfigTemplate itself.

netbox/extras/api/mixins.py
```diff
@@ -64,12 +68,26 @@ class RenderConfigMixin(ConfigTemplateRenderMixin):
     """
     Provides a /render-config/ endpoint for REST API views whose model may have a ConfigTemplate assigned.
     """
+
+    def get_permissions(self):
+        # For render_config action, check only token write ability (not model permissions)
+        if self.action == 'render_config':
+            return [TokenWritePermission()]
```
jeremystretch, line 72: As with ConfigTemplateViewSet, I think we still need to enforce view permission for the device/VM for which the template is being rendered.

netbox/extras/api/views.py
```diff
@@ -238,13 +240,26 @@ class ConfigTemplateViewSet(SyncedDataMixin, ConfigTemplateRenderMixin, NetBoxMo
...
     def render(self, request, pk):
         """
         Render a ConfigTemplate using the context data provided (if any). If the client requests "text/plain" data,
         return the raw rendered content, rather than serialized JSON.
         """
+        self.queryset = self.queryset.model.objects.all().restrict(request.user, 'render_config')
         configtemplate = self.get_object()
+
+        # Check render_config permission
+        perm = get_permission_for_model(configtemplate, 'render_config')
```
jeremystretch: Should this be `render` for the ConfigTemplate?
```suggestion
        perm = get_permission_for_model(configtemplate, 'render')
```

netbox/extras/api/mixins.py
```diff
@@ -64,12 +68,26 @@ class RenderConfigMixin(ConfigTemplateRenderMixin):
...
     def render_config(self, request, pk):
         """
         Resolve and render the preferred ConfigTemplate for this Device.
         """
+        self.queryset = self.queryset.model.objects.all().restrict(request.user, 'render_config')
         instance = self.get_object()
+
+        # Check render_config permission
+        perm = get_permission_for_model(instance, 'render_config')
+        if not request.user.has_perm(perm, obj=instance):
```
jeremystretch: Here we're restricting the initial queryset and also checking for the permission on the retrieved object. If the user doesn't have permission for the object, it won't be in the queryset, and the request will yield a 404. If we want to check the permission explicitly instead (i.e. to return a more helpful error message), we should remove the restriction from the queryset (unless I've missed something).

netbox/extras/api/mixins.py
```diff
@@ -64,12 +68,26 @@ class RenderConfigMixin(ConfigTemplateRenderMixin):
...
+        if self.action == 'render_config':
+            return [TokenWritePermission()]
+        return super().get_permissions()
+
     @action(detail=True, methods=['post'], url_path='render-config', renderer_classes=[JSONRenderer, TextRenderer])
     def render_config(self, request, pk):
         """
         Resolve and render the preferred ConfigTemplate for this Device.
         """
+        self.queryset = self.queryset.model.objects.all().restrict(request.user, 'render_config')
```
jeremystretch: `all()` isn't needed here.
```suggestion
        self.queryset = self.queryset.model.objects.restrict(request.user, 'render_config')
```

jnovinger: Doh

netbox/extras/api/views.py
jnovinger: I hemmed a bit on this one, but convinced myself you are right.

jnovinger:
I may have over-corrected a little on enforcing the view permissions, enforcing them on both the object in question and for the associated config template. Also, trying to be consistent, I enforced the view permissions by chaining another `.restrict(request.user, 'view')` for enforcing it for the object in question and with `has_perm(...)` for the associated config template.

If any of these aren't the best way to do this, please let me know.

netbox/extras/api/mixins.py
```diff
@@ -64,19 +67,34 @@ class RenderConfigMixin(ConfigTemplateRenderMixin):
...
         configtemplate = instance.get_config_template()
         if not configtemplate:
             return Response({
                 'error': f'No config template found for this {object_type}.'
             }, status=HTTP_400_BAD_REQUEST)
 
+        # Check view permission for ConfigTemplate
+        if not request.user.has_perm('extras.view_configtemplate', obj=configtemplate):
+            raise PermissionDenied(_("This user does not have permission to view this configuration template."))
+
```
jeremystretch: I don't think we need to enforce `view` permission for the template itself, because it's not being displayed to the user. (We're not enforcing this currently.) The scenario which immediately jumps to my mind is an automated provisioning system that has permission to view devices/VMs and render their configs, but has no other permissions.

netbox/netbox/api/authentication.py
```diff
@@ -164,6 +164,17 @@ def has_object_permission(self, request, view, obj):
...
 
+class TokenWritePermission(BasePermission):
+    """
+    Verify the token has write_enabled for unsafe methods, without requiring specific model permissions.
+    Used for custom actions that accept user data but don't map to standard CRUD operations.
+    """
+    def has_permission(self, request, view):
+        if isinstance(request.auth, Token):
+            return request.method in SAFE_METHODS or request.auth.write_enabled
+        return True
```
jeremystretch: It feels like this should raise an error if the authentication mechanism is anything other than a token, as a sanity check. Session auth should never be used in the few places we might use this.

netbox/extras/api/mixins.py
```diff
@@ -64,19 +67,34 @@ class RenderConfigMixin(ConfigTemplateRenderMixin):
...
+        return super().get_permissions()
+
     @action(detail=True, methods=['post'], url_path='render-config', renderer_classes=[JSONRenderer, TextRenderer])
     def render_config(self, request, pk):
         """
         Resolve and render the preferred ConfigTemplate for this Device.
         """
+        self.queryset = self.queryset.model.objects.restrict(request.user, 'render_config').restrict(
+            request.user, 'view'
+        )
```
jeremystretch, line 83:
```suggestion
        self.queryset = self.queryset.restrict(request.user, 'render_config').restrict(request.user, 'view')
```

netbox/extras/tests/test_api.py
```diff
@@ -854,6 +855,23 @@ def setUpTestData(cls):
         )
         ConfigTemplate.objects.bulk_create(config_templates)
 
+    def test_render(self):
+        configtemplate = ConfigTemplate.objects.first()
+
+        self.add_permissions('extras.render_configtemplate', 'extras.view_configtemplate')
+        url = reverse('extras-api:configtemplate-detail', kwargs={'pk': configtemplate.pk}) + 'render/'
```
jeremystretch:
```suggestion
        url = reverse('extras-api:configtemplate-render', kwargs={'pk': configtemplate.pk})
```

netbox/extras/tests/test_api.py
```diff
@@ -854,6 +855,23 @@ def setUpTestData(cls):
...
+        response = self.client.post(url, {'foo': 'bar'}, format='json', **self.header)
+        self.assertHttpStatus(response, status.HTTP_200_OK)
+        self.assertEqual(response.data['content'], 'Foo: bar')
+
+    def test_render_without_permission(self):
+        configtemplate = ConfigTemplate.objects.first()
+
+        # No permissions added - user has no render permission
+        url = reverse('extras-api:configtemplate-detail', kwargs={'pk': configtemplate.pk}) + 'render/'
```
jeremystretch:
```suggestion
        url = reverse('extras-api:configtemplate-render', kwargs={'pk': configtemplate.pk})
```

netbox/dcim/tests/test_api.py
```diff
@@ -1486,12 +1485,27 @@ def test_render_config(self):
         device.config_template = configtemplate
         device.save()
 
-        self.add_permissions('dcim.add_device')
+        self.add_permissions('dcim.render_config_device', 'dcim.view_device', 'extras.view_configtemplate')
```
jeremystretch: Assuming we drop this requirement:
```suggestion
        self.add_permissions('dcim.render_config_device', 'dcim.view_device')
```

netbox/virtualization/tests/test_api.py
```diff
@@ -281,12 +281,30 @@ def test_render_config(self):
         vm.config_template = configtemplate
         vm.save()
 
-        self.add_permissions('virtualization.add_virtualmachine')
+        self.add_permissions(
+            'virtualization.render_config_virtualmachine', 'virtualization.view_virtualmachine',
+            'extras.view_configtemplate'
+        )
```
jeremystretch, line 288: Assuming we drop this requirement:
```suggestion
        self.add_permissions(
            'virtualization.render_config_virtualmachine', 'virtualization.view_virtualmachine',
        )
```

netbox/virtualization/tests/test_api.py
```diff
@@ -281,12 +281,30 @@ def test_render_config(self):
...
+            name='Config Template 1',
+            template_code='Config for virtual machine {{ virtualmachine.name }}'
+        )
+
+        vm = VirtualMachine.objects.first()
+        vm.config_template = configtemplate
+        vm.save()
+
+        # No permissions added - user has no render_config permission
+        url = reverse('virtualization-api:virtualmachine-detail', kwargs={'pk': vm.pk}) + 'render-config/'
```
jeremystretch:
```suggestion
        url = reverse('virtualization-api:virtualmachine-render-config', kwargs={'pk': vm.pk})
```

netbox/dcim/tests/test_api.py
```diff
@@ -1486,12 +1485,27 @@ def test_render_config(self):
...
+            name='Config Template 1',
+            template_code='Config for device {{ device.name }}'
+        )
+
+        device = Device.objects.first()
+        device.config_template = configtemplate
+        device.save()
+
+        # No permissions added - user has no render_config permission
+        url = reverse('dcim-api:device-detail', kwargs={'pk': device.pk}) + 'render-config/'
```
jeremystretch:
```suggestion
        url = reverse('dcim-api:device-render-config', kwargs={'pk': device.pk})
```

netbox/extras/api/views.py
jeremystretch, line 1: Want to add a test that checks for enforcement of `write_enabled` on the token?

jeremystretch (changes requested):

netbox/extras/api/mixins.py
jnovinger: I'm hesitant to change `self.queryset.model.objects.restrict()` to `self.queryset.restrict()` because of how `BaseViewSet.initial()` handles POST requests.

From [`netbox/api/viewsets/__init__.py:BaseViewSet.initial()`](https://github.com/netbox-community/netbox/blob/5334c8143c6b8a2e0962f266fa6f2b9d5ac4383e/netbox/netbox/api/viewsets/__init__.py#L42-L48):

```python
def initial(self, request, *args, **kwargs):

    if request.user.is_authenticated:
        if action := HTTP_ACTIONS[request.method]:
            self.queryset = self.queryset.restrict(request.user, action)
```

With `HTTP_ACTIONS = {'POST': 'add', ...}`, by the time `render_config()` executes, `self.queryset` has already been restricted to the `add` permission by `BaseViewSet.initial()`.

If we use `self.queryset.restrict(request.user, 'render_config')` to directly restrict the existing queryset, it would chain onto the existing `add` restriction, requiring users to have both `add` AND `render_config` permissions, which is contrary to what the original issue calls for.

Using `self.queryset.model.objects.restrict()` starts fresh from the model manager, bypassing the default POST='add' restriction and applying only the permissions we want: `render_config` + `view`.

Am I missing something about how this should work?

docs/features/configuration-rendering.md
```diff
@@ -90,3 +90,10 @@ http://netbox:8000/api/extras/config-templates/123/render/ \
   "bar": 123
 }'
 ```
+
+!!! note "Permissions"
+    Rendering configuration templates via the REST API requires appropriate permissions for the relevant object type:
+
+    * To render a device's configuration via `/api/dcim/devices/{id}/render-config/`, assign a permission for "DCIM > Device" with the `render_config` action
+    * To render a virtual machine's configuration via `/api/virtualization/virtual-machines/{id}/render-config/`, assign a permission for "Virtualization > Virtual Machine" with the `render_config` action
+    * To render a config template directly via `/api/extras/config-templates/{id}/render/`, assign a permission for "Extras > Config Template" with the `render` action
```
jeremystretch:
```suggestion
    * To render a device's configuration via `/api/dcim/devices/{id}/render-config/`, assign a permission for "DCIM > Device" with the `render_config` action.
    * To render a virtual machine's configuration via `/api/virtualization/virtual-machines/{id}/render-config/`, assign a permission for "Virtualization > Virtual Machine" with the `render_config` action.
    * To render a config template directly via `/api/extras/config-templates/{id}/render/`, assign a permission for "Extras > Config Template" with the `render` action.
```

netbox/extras/api/mixins.py
jeremystretch: Ah, I see. It's probably worth including a comment to this effect then because it's very easy to overlook the override.

netbox/extras/api/views.py
```diff
@@ -238,13 +238,21 @@ class ConfigTemplateViewSet(SyncedDataMixin, ConfigTemplateRenderMixin, NetBoxMo
...
+            return [TokenWritePermission()]
+        return super().get_permissions()
+
     @action(detail=True, methods=['post'], renderer_classes=[JSONRenderer, TextRenderer])
     def render(self, request, pk):
         """
         Render a ConfigTemplate using the context data provided (if any). If the client requests "text/plain" data,
         return the raw rendered content, rather than serialized JSON.
         """
+        self.queryset = self.queryset.model.objects.restrict(request.user, 'render').restrict(request.user, 'view')
```
jeremystretch, line 254:
```suggestion
        # Override restrict() on the default queryset to enforce the render & view actions
        self.queryset = self.queryset.model.objects.restrict(request.user, 'render').restrict(request.user, 'view')
```

netbox/extras/api/mixins.py
```diff
@@ -64,12 +65,23 @@ class RenderConfigMixin(ConfigTemplateRenderMixin):
...
+        if self.action == 'render_config':
+            return [TokenWritePermission()]
+        return super().get_permissions()
+
     @action(detail=True, methods=['post'], url_path='render-config', renderer_classes=[JSONRenderer, TextRenderer])
     def render_config(self, request, pk):
         """
         Resolve and render the preferred ConfigTemplate for this Device.
         """
+        self.queryset = self.queryset.model.objects.restrict(request.user, 'render_config').restrict(
```
jeremystretch, line 81:
```suggestion
        # Override restrict() on the default queryset to enforce the render_config & view actions
        self.queryset = self.queryset.model.objects.restrict(request.user, 'render_config').restrict(
```

jeremystretch (approved):
Thanks @jnovinger!

PR #20556, opened 2025-10-13
Fixes #20156: rack elevation labels not using previous setting

aq5747:
### Fixes: #20156

- Fixes a regression that prevented rack elevation images from loading in with the user's last selected view setting


jnovinger (approved):
Thanks, @aq5747!

Issue #19724, opened 2025-06-19
Pagination metadata for graphql


patrickfnielsen:
### NetBox version

v4.3.2

### Feature type

Change to existing functionality

### Proposed functionality

NetBox v4.3.0 introduced pagination in GraphQL.
I would like to propose that we include support for a "metadata" object containing record counts, like there is in the rest api.

It could look like this:
```
    page_metdata {
      count
    }
```

### Use case

Allows getting the total number of records for pagination to display total number of records to users / Provide accurate "Showing X of Y results".

arthanson:
This is actually referenced in the Strawberry docs as an optional way to present the information: https://strawberry.rocks/docs/guides/pagination/overview

patrickfnielsen:
@arthanson strawberry-django provides an generic way to implement it: https://strawberry.rocks/docs/django/guide/pagination

But as I read it, it would be a breaking change as items would be wrapped in a `results` container. I assume that is something you want to avoid? If so I'll can have a look into how an custom implementation might look

jeremystretch:
> But as I read it, it would be a breaking change as items would be wrapped in a `results` container. I assume that is something you want to avoid?

Yes; we need to avoid any breaking changes to the GraphQL API in v4.4. Let's tag this for v4.5.0.

tyler-8:
This is important for effective pagination usage. With a total count and a limit, you can pre-calculate the number of required pages. From what I can tell in the current implementation, you end up having to paginate until you get no results back - which leads to one more API call than is required and no way to know the size of the dataset through GraphQL.

patrickfnielsen:
@arthanson I am curios as I see you created an implementation for the v2 api that was closed - Is that because v2 of the GraphQL API wont be in NetBox 4.5.0?

If that is the case I can understand it as it's a large breaking change for v1 - but if you are introducing a new version, now would be the perfect time for a large breaking change ?

arthanson:
@patrickfnielsen yes, it and v2 of the API were too much of a breaking change for 4.5.  We will consider it for a later release.

PR #20578, opened 2025-10-14
19724 Add pagination metadata to GraphQL queries

arthanson:
### Fixes: #19724 

Adds new OffsetPaginated Generic to the v2 of the GraphQL API.  For information on OffsetPaginated Generic see: https://strawberry.rocks/docs/django/guide/pagination. This returns metadata of the record counts and affects both the return of GraphQL queries as well as submission for paginated results.  This PR updates the documentation on the new formats and puts in tests for the new v2 version of the queries.


jeremystretch:
Holding on this until we've added GraphQL API v2 per #20603. Once that's done, this can be rebased and we'll make the necessary adaptations to move the metadata to the v2 API.

jeremystretch:
Following the change in #20603 (which introduced the v2 API), the GraphQL API now defaults to the original v1 format. Seems like it would be most prudent to refactor and duplicate the existing `test_graphql_get_object()`, `test_graphql_list_objects()`, and `test_graphql_filter_objects()` methods to test both the v1 and v2 APIs.

arthanson:
@jeremystretch ready for re-review.

arthanson:
Closing as we decided this is too much of a breaking change to include currently as it change the format of the GraphQL query as well as the output.

Issue #20466, opened 2025-10-02
Unable to query devices with primary-ip set in GraphQL - "Cannot resolve keyword 'assigned_object_id' into field"


dxks:
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.1

### Python Version

3.12

### Steps to Reproduce

1. Create a simple GraphQL-query, i.e.
`query {
  device_list(filters:{primary_ip4:{assigned:true}}) {
    id
    primary_ip4 {
      assigned_object {__typename}
    }
  }
}`

2. encounter the error:
`{
  "data": null,
  "errors": [
    {
      "message": "Cannot resolve keyword 'assigned_object_id' into field. Choices are: airflow, asset_tag, bgpsession, bookmarks, cabletermination, cluster, cluster_id, comments, config_template, config_template_id, console_port_count, console_server_port_count, consoleports, consoleserverports, contacts, created, custom_field_data, description, designated_arp_router_peering_services, device_bay_count, device_type, device_type_id, devicebays, face, front_port_count, frontports, id, images, interface_count, interfaces, internet_exchange_points, inventory_item_count, inventoryitems, journal_entries, last_updated, latitude, local_context_data, location, location_id, longitude, module_bay_count, modulebays, modules, name, oob_ip, oob_ip_id, parent_bay, peering_services, platform, platform_id, position, power_outlet_count, power_port_count, poweroutlets, powerports, primary_ip4, primary_ip4_id, primary_ip6, primary_ip6_id, rack, rack_id, rear_port_count, rearports, role, role_id, sbfd_discriminator, serial, services, site, site_id, status, subscriptions, tagged_items, tags, tenant, tenant_id, termination_a_device, termination_z_device, vc_master_for, vc_position, vc_priority, vdcs, virtual_chassis, virtual_chassis_id, virtual_machines",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "device_list"
      ]
    }
  ]
}`

### Expected Behavior

As the filter `has_primary_ip` is no longer available (like cabled within InterfaceFilter) I tried the assigned-filter like I do for the connected as a replacement for cabled.

I expected that this would work in the same way and provides a list of devices with a primary ip set.

### Observed Behavior

Unfortunately the assigned-filter is applied to the device instead of the Interface and causes the error that it cannot be resolved to a field - as the assigned_object_id is on the interface, not the device.

pheus:
[...]

PR #20579, opened 2025-10-15
Fixes #20541: Enhance GraphQL filter methods with dynamic prefixing

### LINKED ISSUE #20541: Ensure all custom GraphQL filter methods apply `prefix` in `Q(...)` for nested filters
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.2

### Python Version

3.12

### Steps to Reproduce

1. Run a nested GraphQL query that relies on a custom filter method returning a raw `Q(...)` without applying the Strawberry‑Django `prefix` (illustrative example from #20466):
   ```graphql
   query {
     device_list(filters: { primary_ip4: { assigned: true } }) {
       id
       primary_ip4 { address }
     }
   }
   ```
2. Observe that the generated lookup key is unqualified (e.g., `assigned_object_id__isnull`) and is applied to the *outer* model rather than the nested relation, producing an error like “Cannot resolve keyword ‘assigned_object_id’ into field”.

**Note**: #20466 proposes/fixes one concrete instance in `IPAddressFilter.assigned` by using `return Q(**{f"{prefix}assigned_object_id__isnull": not value})`. This issue tracks auditing **other** custom filter methods for the same class of bug.

### Expected Behavior

All custom GraphQL filter methods that return `Q` qualify their lookups using the provided `prefix`, so nested filters target the correct related model and do not error.

### Observed Behavior

At least one confirmed case (#20466) shows a custom filter method constructing `Q(...)` without `prefix`, causing nested filters to mis-target the outer model and fail with “Cannot resolve keyword …”. Other custom filter methods that build raw `Q(...)` may be similarly affected.

---

#### Proposed Fix / Scope
- **Audit** all `@strawberry_django.filter_field()` methods across GraphQL filters and update any `return Q(...)` to include `prefix` in the lookup keys. Example pattern:
  ```python
  @strawberry_django.filter_field()
  def <name>(self, value: T, prefix: str) -> Q:
      return Q(**{f"{prefix}<field_path>": value})
  ```
- Where a method returns `(queryset, Q)`, ensure any annotated keys or aliases also incorporate `prefix`.

#### References
- Example bug & reproduction: #20466
- Strawberry‑Django docs show `prefix` is **required** for nested filters and demonstrate qualified `Q(**{f"{prefix}…": …})` lookups.
  - https://strawberry.rocks/docs/django/guide/filters#custom-filter-methods
  - https://strawberry.rocks/docs/django/guide/filters#resolver-arguments

pheus:
### Fixes: #20541

Refactors filter methods to enable dynamic query prefixing for greater flexibility in nested queries. Improves error handling for invalid network inputs and ensures consistency across IPAM filter implementations.

---

### Context

This follows the discussion in **Bug #20466** and builds on **PR #20538** (which addressed `IPAddressFilter.assigned`). The goal here is to complete the prefixing/validation sweep within **`ipam.graphql.filters` only** and ensure nested filters consistently target the correct related model.

### Summary of changes

- **Prefix-aware Q objects (nested filtering):**  
  Standardize construction of lookups as `Q(**{f"{prefix}…": …})` in custom filter methods so nested queries are resolved against the intended model path.
- **Input safeguards:**  
  `contains`/`parent`-style helpers now guard network parsing with `try/except`, **skipping invalid entries** rather than failing the entire filter.
- **Targeted refactors for clarity/consistency:**  
  Minor internal tidy-ups (e.g., shared helper for device-based lookups) while preserving existing behavior.

### 🚨 Breaking change (GraphQL): `AggregateFilter`

`Aggregate.prefix` is a **concrete field**, not a relation. To avoid double-qualifying lookups in nested contexts, this PR proposes:

- Replace the nested filter with a field lookup:
  ```py
  prefix: FilterLookup[str] | None
  ```
- Add a field-level `contains(value: [CIDR])` that applies `prefix__net_contains` to the **Aggregate** field itself.

**Client impact (input shape changed for Aggregates only):**

- **Before**  
  ```graphql
  aggregate_list(filters: { prefix: { contains: "10.0.0.0/8" } }) { prefix }
  ```
- **After**  
  - containment:  
    ```graphql
    aggregate_list(filters: { contains: ["10.0.0.0/8"] }) { prefix }
    ```
  - equality:  
    ```graphql
    aggregate_list(filters: { prefix: { exact: "10.0.0.0/8" } }) { prefix }
    ```

No other GraphQL inputs are changed.

### Files/areas touched (IPAM only)

- `ipam.graphql.filters.AggregateFilter`  
  - `prefix` now a `FilterLookup[str]` on the **field**.  
  - New `contains([...])` uses `prefix__net_contains`.  
- `ipam.graphql.filters.PrefixFilter`  
  - `contains([...])` remains relation-aware; lookups are prefixed via `f"{prefix}prefix__net_contains"`.  
- `ipam.graphql.filters.IPRangeFilter`  
  - `parent([...])` and `contains([...])` use prefixed lookups with input validation.  
- `ipam.graphql.filters.IPAddressFilter`  
  - `parent([...])` uses prefixed lookups with input validation.  
  - (Note: `assigned` was handled in **PR #20538** per **Bug #20466**.)

### Upgrade notes

- **Breaking (Aggregates only):** Update GraphQL clients to use:
  - `aggregate_list(filters: { contains: ["<cidr>"] })` for containment, and  
  - `aggregate_list(filters: { prefix: { exact: "<cidr>" } })` for equality.  
- No changes required for other IPAM GraphQL filters.

### Related

- Bug: #20466  
- Prior fix: #20538

### DIFF
--- a/netbox/ipam/graphql/filters.py
+++ b/netbox/ipam/graphql/filters.py
@@ -79,11 +79,27 @@ class ASNRangeFilter(TenancyFilterMixin, OrganizationalModelFilterMixin):
 
 @strawberry_django.filter_type(models.Aggregate, lookups=True)
 class AggregateFilter(ContactFilterMixin, TenancyFilterMixin, PrimaryModelFilterMixin):
-    prefix: Annotated['PrefixFilter', strawberry.lazy('ipam.graphql.filters')] | None = strawberry_django.filter_field()
-    prefix_id: ID | None = strawberry_django.filter_field()
+    prefix: FilterLookup[str] | None = strawberry_django.filter_field()
     rir: Annotated['RIRFilter', strawberry.lazy('ipam.graphql.filters')] | None = strawberry_django.filter_field()
     rir_id: ID | None = strawberry_django.filter_field()
     date_added: DateFilterLookup[date] | None = strawberry_django.filter_field()
+
+    @strawberry_django.filter_field()
+    def contains(self, value: list[str], prefix) -> Q:
+        """
+        Return aggregates whose `prefix` contains any of the supplied networks.
+        Mirrors PrefixFilter.contains but operates on the Aggregate.prefix field itself.
+        """
+        if not value:
+            return Q()
+        q = Q()
+        for subnet in value:
+            try:
+                query = str(netaddr.IPNetwork(subnet.strip()).cidr)
+            except (AddrFormatError, ValueError):
+                continue
+            q |= Q(**{f"{prefix}prefix__net_contains": query})
+        return q
 
 
 @strawberry_django.filter_type(models.FHRPGroup, lookups=True)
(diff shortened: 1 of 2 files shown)


jnovinger (commented):
Thanks for the thorough audit, @pheus. The fixes look good and I agree the breaking change for `AggregateFilter` is the correct choice.

Please add tests covering the nested query/filter scenarios that these fixes enable as well as the fixes from #20538 (which I merged and you probably want to rebase on). Please also test the error handling improvement where invalid networks are skipped rather than breaking the entire filter.

You can add custom test methods to the appropriate model test classes in [`netbox/ipam/tests/test_api.py`](https://github.com/netbox-community/netbox/blob/main/netbox/ipam/tests/test_api.py) (e.g., `IPAddressTest`, `IPRangeTest`, `PrefixTest`, `AggregateTest`). These inherit from `APIViewTestCase` which provides `self.client`, `self.user`, and other helpers. Please mark these tests with `@tag('regression')`, which you see elsewhere in the same file. For examples, see the templated-tests that are defined in  [`netbox/netbox/tests/test_graphql.py`](https://github.com/netbox-community/netbox/blob/main/netbox/netbox/tests/test_graphql.py).

pheus:
Thanks for the thoughtful review — the guidance is super helpful!

I’m working on this next, but it’ll take me a bit to put together all the fixtures and assertions. I’ll push an update as soon as it’s ready. Thanks again for the review!

jnovinger (approved):
Thank you for adding the tests, @pheus.

PR #20586, opened 2025-10-15
Closes #19751: Disable occupied Module Bays in form dropdowns

### LINKED ISSUE #19751: Provide feature parity in module bay dropdown
### NetBox version

v4.3.2

### Feature type

New functionality

### Proposed functionality

When assigning/creating a module on a device, currently there is no indication if a module bay is occupied or not. I propose a indication like with interfaces or front/rear ports, where bays in the drop down menu are grayed out and you are not able to select them.

### Use case

When assign a lot of modules through the GUI to a device with a lot of module bays this will help the user by being able to see the which bays are already occupied and also not being able to select them.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Fixes: #19751

Implements the occupied‑bay UI behavior without adding a private `_occupied` field to the API.

Update the `module_bay` form field to use context with `disabled_indicator: 'installed_module'` so occupied bays are disabled in dropdowns

**Notes**
- Backward compatible; no database or migration changes
- Keeps the REST surface clean while providing the needed UI parity

### DIFF
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -755,7 +755,10 @@ class ModuleForm(ModuleCommonForm, NetBoxModelForm):
         queryset=ModuleBay.objects.all(),
         query_params={
             'device_id': '$device'
-        }
+        },
+        context={
+            'disabled': 'installed_module',
+        },
     )
     module_type = DynamicModelChoiceField(
         label=_('Module type'),


arthanson (changes requested):
This is showing up in the regular REST API (see screenshot) which with an underscore seems like it is an internal / private parameter? 
Monosnap Image 2025-10-17 14-25-32

pheus:

You’re right! The leading underscore makes it look internal. I initially mirrored `CabledObjectSerializer` (which exposes `_occupied`), but I agree it’s not ideal here and a brief‑only approach would still require listing it in `Meta.fields`.

I’ve reworked this to keep the API surface clean:

- Drop `_occupied` entirely
- Use `installed_module` as the occupancy signal (included in brief)
- Update the form field context to disable options via `installed_module`

This preserves the UI behavior without adding a private‑looking field. If you’d prefer an explicit boolean, I can expose a read‑only `occupied` derived from `installed_module`. Happy to adjust. Thanks again!

pheus:
Apologies for the noise! I tried to rename the feature branch in GitHub’s UI to better reflect the updated scope, and that inadvertently auto‑closed this PR. I’ve restored the branch and reopened the PR—no code changes were introduced as part of this. Sorry for the disruption!

arthanson (approved):
Thanks @pheus! much cleaner implementation.

PR #20611, opened 2025-10-17
Fixes #20610: Correct VLAN ID range calculation logic

### LINKED ISSUE #20610: VLAN Group utilization is not being calculated correctly
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.4

### Python Version

3.10

### Steps to Reproduce

1. Create a new VLAN group with VLAN IDs value of `1-2`.
2. Create two new VLANs (1 and 2) and assign them to the new VLAN group.
3. Notice that the VLAN group shows as 66.7% utilized, but it should display as 100%. 
4. Next update the VLAN group to use VLAN IDs value of `1,2`
5. Notice now that the utilization is showing as 50% instead of 100%.

### Expected Behavior

VLAN Group utilization should be accurate.

### Observed Behavior

VLAN Group utilization is no longer being calculated correctly, as described above.

pheus:
### Fixes: #20610

Correct an off‑by‑one in `VLANGroup._total_vlan_ids` by computing lengths with half‑open `[lo, hi)` semantics during `save()`. Removes the `+1`, adds a test and includes a data migration to backfill existing `_total_vlan_ids`.

No schema changes.

### DIFF
--- a/netbox/ipam/migrations/0083_vlangroup_populate_total_vlan_ids.py
+++ b/netbox/ipam/migrations/0083_vlangroup_populate_total_vlan_ids.py
@@ -1,0 +1,27 @@
+from django.db import migrations
+
+
+def populate_vlangroup_total_vlan_ids(apps, schema_editor):
+    VLANGroup = apps.get_model('ipam', 'VLANGroup')
+    db_alias = schema_editor.connection.alias
+
+    vlan_groups = VLANGroup.objects.using(db_alias).only('id', 'vid_ranges')
+    for group in vlan_groups:
+        total_vlan_ids = 0
+        if group.vid_ranges:
+            for r in group.vid_ranges:
+                # Half-open [lo, hi): length is (hi - lo).
+                if r is not None and r.lower is not None and r.upper is not None:
+                    total_vlan_ids += r.upper - r.lower
+        group._total_vlan_ids = total_vlan_ids
+    VLANGroup.objects.using(db_alias).bulk_update(vlan_groups, ['_total_vlan_ids'], batch_size=100)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('ipam', '0082_add_prefix_network_containment_indexes'),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_vlangroup_total_vlan_ids, migrations.RunPython.noop),
+    ]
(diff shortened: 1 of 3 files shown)


pheus:
Thanks for all the release work!
If it’s still feasible, could this be included in v4.4.5? It fixes a regression I introduced in v4.4.3 (#20610) that miscounts `_total_vlan_ids`. I’m sorry for the slip.

No pressure if the window has closed. I completely understand. I can rebase or adjust anything right away if that helps.

Thanks again for your time and for everything you do.

jeremystretch:
I'll try to review this tomorrow ahead of the v4.4.5 release if @bctiemann hasn't had a chance yet.

netbox/ipam/migrations/0083_vlangroup_populate_total_vlan_ids.py
```diff
@@ -0,0 +1,26 @@
...
+
+    qs = VLANGroup.objects.using(db).only('id', 'vid_ranges')
+    for group in qs.iterator():
+        total_vlan_ids = 0
+        if group.vid_ranges:
+            for r in group.vid_ranges:
+                # Half-open [lo, hi): length is (hi - lo).
+                if r is not None and r.lower is not None and r.upper is not None:
+                    total_vlan_ids += r.upper - r.lower
+        VLANGroup.objects.using(db).filter(pk=group.pk).update(_total_vlan_ids=total_vlan_ids)
```
jeremystretch: A minor concern, but have you considered setting the updated value on each VLANRange instance in the queryset, and then calling `bulk_update()` (with perhaps `batch=100`) to update the database more efficiently? Should be safe AFAICT.

pheus: Great suggestion! Thank you! I was trying to keep the DB work light by selecting only the needed fields, but I missed the chance to use `bulk_update()` here.

Agreed that batching with `bulk_update(batch_size=100)` is the better approach and should be safe since we’re only updating the computed `_total_vlan_ids` in a data migration (no signals or `save()` hooks).

I’ll switch to this approach.

Thanks again for the review!

pheus:
I’ve rebased and switched the migration to use `bulk_update()` as suggested. Please take another look when you have a moment. Happy to adjust anything further.

jeremystretch (approved):
Thanks @pheus!

PR #20616, opened 2025-10-17
Fixes #20524: Enhance API script scheduling validation

pheus:
### Fixes: #20524

Tightens script scheduling validation:

- Rejects `schedule_at` values in the past.
- When only `interval` is provided, starts now and repeats.
- Uses `local_now()` for accurate, timezone‑aware comparisons.

Small, backward‑compatible fix to make scheduling behave as expected.

### DIFF
--- a/netbox/extras/api/serializers_/scripts.py
+++ b/netbox/extras/api/serializers_/scripts.py
@@ -5,6 +5,7 @@ from rest_framework import serializers
 from core.api.serializers_.jobs import JobSerializer
 from extras.models import Script
 from netbox.api.serializers import ValidatedModelSerializer
+from utilities.datetime import local_now
 
 __all__ = (
     'ScriptDetailSerializer',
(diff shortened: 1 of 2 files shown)


netbox/extras/api/serializers_/scripts.py
```diff
@@ -66,11 +67,30 @@ class ScriptInputSerializer(serializers.Serializer):
...
-            raise serializers.ValidationError(_("Scheduling is not enabled for this script."))
+            raise serializers.ValidationError(_('Scheduling is not enabled for this script.'))
         return value
+
+    def validate(self, data):
+        """
+        Validates the given data and ensures the necessary fields are populated.
+        """
+        # Set the schedule_at time to now if only an interval is provided.
+        if 'interval' in data and 'schedule_at' not in data:
```
jnovinger:
```suggestion
        if 'interval' in data and not data.get('schedule_at'):
```

It seems there's a case not noticed in the original issue, where the API caller might pass something like `{"scheduled": null, "interval": 60}` and they'd end up with a recurring job that has no set start time. All other edge-case-y values that I could think of that might satisfy this conditional fail because they aren't a valid datetime format.

jnovinger (commented):
Thanks @pheus. This fix looks good, although I did note one other edge case (w/ a suggestion) that the original bug report did not mention. Can you please address that, since we're already making changes there?

Could you please also add the following tests for this? Probably in `ScriptTest` around line 910 in `netbox/extras/tests/test_api.py`:

```python
def test_schedule_script_past_time_rejected(self):
    """Scheduling with past schedule_at should 
fail"""
    # POST with schedule_at = timezone.now() - timedelta(hours=1)
    # Expect 400 with "must be in the future" error

def test_schedule_script_interval_only(self):
    """Interval without schedule_at should auto-set 
schedule_at to now"""
    # POST with interval=60, no schedule_at
    # Verify job created with schedule_at set

def test_schedule_script_when_disabled(self):
    """Scheduling should fail when 
script.scheduling_enabled=False"""
    # POST with schedule_at or interval
```

Let me know if you have any questions.

netbox/extras/api/serializers_/scripts.py
pheus: Great point, and I agree. I’ll use `.get()` for both `schedule_at` and `interval`. Using `.get()` treats an explicit `null` the same as a missing key, so the “interval‑only” behavior still applies when `schedule_at` is `null`.

I’ll normalize & validate accordingly:
- If both are `None` → no scheduling (run now).
- If `interval` is set and `schedule_at` is falsy → `schedule_at = local_now()`.
- If `schedule_at` is set → must be in the future.
- If either is set and `scheduling_enabled` is `False` → reject.

netbox/extras/tests/test_api.py
```diff
@@ -878,14 +879,16 @@ def run(self, data, commit=True):
...
             module=module,
-            name="Test script",
+            name='Test script',
             is_executable=True,
         )
+        cls.url = reverse('extras-api:script-detail', kwargs={'pk': script.pk})
 
+    @property
     def python_class(self):
         return self.TestScriptClass
```
pheus, line 893: Make this a property to mirror the model’s `cached_property`. This returns the class object (not a bound function), which fixes the test setup and matches production behavior.

pheus:

I’ve pushed updates to address your suggestion and added tests:

- Use `.get()` so `{"schedule_at": null, "interval": 60}` starts now (no start‑less recurring job).
- Tests added: past `schedule_at` → 400; `interval`‑only → auto‑sets `schedule_at`; scheduling disabled → reject.
- Fixed the test setup to return the class from `python_class`, matching production behavior.

Happy to tweak anything.

jnovinger (approved):

PR #20618, opened 2025-10-17
Closes #20617: Introduce BaseModel

### LINKED ISSUE #20617: Introduce `BaseModel` to hold universal overrides for Django's `models.Model`
### Proposed Changes

1. Introduce a new abstract model named BaseModel.
2. Move the default manager definitions from ChangeLoggedModel and NetBoxModel to BaseModel.
3. Move the `clean()` method from NetBoxModel to BaseModel.
4. Update NetBoxModel & ChangeLoggedModel to inherit from BaseModel instead of `models.Model`.

### Justification

This approach does a better job of organizing stock functionality, and provides a convenient hook for (mostly) global model overrides. It will also allow for a slightly cleaner implementation of #20304.

jeremystretch:
### Closes: #20617

- Introduces the new abstract BaseModel class
- ChangeLoggedModel & NetBoxModel now inherit from BaseModel
- OrganizationalModel now inherits from NetBoxModel (and indirectly from BaseModel)

### DIFF
--- a/netbox/netbox/models/__init__.py
+++ b/netbox/netbox/models/__init__.py
@@ -50,21 +50,15 @@ class NetBoxFeatureSet(
 # Base model classes
 #
 
-class ChangeLoggedModel(ChangeLoggingMixin, CustomValidationMixin, EventRulesMixin, models.Model):
+class BaseModel(models.Model):
     """
-    Base model for ancillary models; provides limited functionality for models which don't
-    support NetBox's full feature set.
-    """
-    objects = RestrictedQuerySet.as_manager()
+    A global base model for all NetBox objects.
 
-    class Meta:
-        abstract = True
-
-
-class NetBoxModel(NetBoxFeatureSet, models.Model):
+    This class provides some important overrides to Django's default functionality, such as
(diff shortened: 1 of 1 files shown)


netbox/netbox/models/__init__.py
```diff
@@ -103,6 +97,25 @@ def clean(self):
                     setattr(self, field.name, obj)
 
 
+class ChangeLoggedModel(ChangeLoggingMixin, CustomValidationMixin, EventRulesMixin, BaseModel):
```
arthanson, line 100: Does it matter that the clean method from BaseModel is on ChangeLoggedModel (it wasn't before)?

arthanson (approved):
Just the one question on the clean method.

netbox/netbox/models/__init__.py
jeremystretch: I believe it was an oversight. The additional GFK validation may not have originally been needed on ChangeLoggedModel, but we do now have models which inherit from that class with GFKs. I've moved it to BaseModel to ensure consistent coverage.

PR #20624, opened 2025-10-19
Fixes #20551: Support quick-add form prefix in automatic slug generation

jnovinger:
Fixes: #20551

The slug generation logic in `reslug.ts` looks for form fields using hard-coded ID selectors like `#id_slug` and `#id_name`. In quick-add modals, Django applies a `quickadd` prefix to form fields (introduced in #20542), resulting in IDs like `#id_quickadd-slug` and `#id_quickadd-name`. The logic couldn't find these prefixed fields, so automatic slug generation failed silently in quick-add modals.

This fix updates the field selectors to try both unprefixed and prefixed patterns using the nullish coalescing operator (`??`). The logic checks for the standard field ID first and falls back to the quickadd-prefixed ID if the standard one isn't found.

Manually tested automatic slug generation in both direct add forms and quick-add modals. Both the blur event (tab out of name field) and the explicit slug button click work correctly.


netbox/project-static/src/buttons/reslug.ts
```diff
@@ -23,10 +23,17 @@ export function initReslug(): void {
   for (const slugButton of getElements<HTMLButtonElement>('button#reslug')) {
     const form = slugButton.form;
     if (form == null) continue;
-    const slugField = form.querySelector('#id_slug') as HTMLInputElement;
+
+    // Try without prefix first, fallback to quickadd prefix for quick-add modals
+    const slugField = (form.querySelector('#id_slug') ??
+      form.querySelector('#id_quickadd-slug')) as HTMLInputElement;
```
jeremystretch: We should avoid relying on fallback behavior as it is unreliable. Instead, we can set the `slug-source` attribute on the slug field to the exact ID of the source field. (This is done on the [`SlugField`](https://github.com/netbox-community/netbox/blob/531ea342074c51448275e43079fc7ae462466904/netbox/utilities/forms/fields/fields.py#L40) form field.)

netbox/project-static/src/buttons/reslug.ts
```diff
@@ -23,10 +23,17 @@ export function initReslug(): void {
   for (const slugButton of getElements<HTMLButtonElement>('button#reslug')) {
```
jeremystretch: We should take this opportunity to use a class here, to avoid duplicated element IDs. This can be changed on the HTML widget in [`render_field.html`](https://github.com/netbox-community/netbox/blob/531ea342074c51448275e43079fc7ae462466904/netbox/utilities/templates/form_helpers/render_field.html#L22).
```suggestion
  for (const slugButton of getElements<HTMLButtonElement>('button.reslug')) {
```

jeremystretch (changes requested):

jeremystretch (approved):
Nice, very elegant

PR #20626, opened 2025-10-20
Fixes #20422: Allow Aggregate and Prefix to filter by family in GraphQL

rinna11:
### Fixes: #20422

This allows `AggregateFilter` and `PrefixFilter` to filter by their family similar with `IPAddressFilter`.
I've tested the following queries manually:

```graphql
aggregate_v6: aggregate_list(filters: {family: FAMILY_6}) {
    prefix
}
prefixes_v4: prefix_list(filters: {family: FAMILY_4}) {
    prefix
    role {
      name
    }
    vrf {
      name
    }
    vlan {
      vid
    }
}
```

### DIFF
--- a/netbox/ipam/graphql/filters.py
+++ b/netbox/ipam/graphql/filters.py
@@ -101,6 +101,13 @@ class AggregateFilter(ContactFilterMixin, TenancyFilterMixin, PrimaryModelFilter
             q |= Q(**{f"{prefix}prefix__net_contains": query})
         return q
 
+    def family(
+        self,
+        value: Annotated['IPAddressFamilyEnum', strawberry.lazy('ipam.graphql.enums')],
+        prefix,
+    ) -> Q:
+        return Q(**{f"{prefix}prefix__family": value.value})
+
 
 @strawberry_django.filter_type(models.FHRPGroup, lookups=True)
 class FHRPGroupFilter(PrimaryModelFilterMixin):
@@ -291,6 +298,14 @@ class PrefixFilter(ContactFilterMixin, ScopedFilterMixin, TenancyFilterMixin, Pr
                 continue
             q |= Q(**{f"{prefix}prefix__net_contains": query})
         return q
+
+    @strawberry_django.filter_field()
+    def family(
+        self,
+        value: Annotated['IPAddressFamilyEnum', strawberry.lazy('ipam.graphql.enums')],
+        prefix,
+    ) -> Q:
+        return Q(**{f"{prefix}prefix__family": value.value})
 
 
 @strawberry_django.filter_type(models.RIR, lookups=True)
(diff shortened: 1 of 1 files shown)


jnovinger (changes requested):
Thanks for the work, @rinna11. The approach looks good and correctly follows the existing pattern.

Please add tests covering the family filter for both `AggregateFilter` and `PrefixFilter`, with both `FAMILY_4` and `FAMILY_6` values to verify the filter returns the correct address family results.

You can add custom test methods to the `AggregateTest` and `PrefixTest` classes in [ipam/tests/test_api.py](https://github.com/netbox-community/netbox/blob/main/netbox/ipam/tests/test_api.py). These inherit from `APIViewTestCase` which provides `self.client`, `self.user`, and other helpers. Please mark these tests with `@tag('regression')`, which you can see elsewhere in the same file. For examples, see the templated tests that are defined in [netbox/tests/test_graphql.py](https://github.com/netbox-community/netbox/blob/main/netbox/netbox/tests/test_graphql.py#L48-L128.).

rinna11:
Hi @jnovinger , thank you for the suggestion!
I had a quick look at `ipam/tests/test_api.py`, and it seems there aren’t any existing GraphQL filter tests for `AggregateTest` and `PrefixTest` there. I did find relevant filtering tests in `ipam/tests/test_filtersets.py`, and those already include conditions for filtering by `family`. These tests are passing both in the current codebase and with this PR.
Also, when the `family` field was enabled for `IPAddressFilter` in #19621 , adding specific filter tests wasn’t required at that time either.

Given that, I would like to kindly ask if we could proceed with merging this PR without adding new tests. Please let me know if you're still up for it after all that—happy to adjust!
Thanks again for your time and review.

jnovinger:
@rinna11 , yeah, let's accept this without new tests. I have some work in mind to improve GraphQL filter testing that will eventually cover all of this. If you are interested in seeing some GraphQL tests that were just added, check out #20579. Apologies for pointing you in the wrong direction with the earlier comment.

It looks like you do now have conflicts, though. Please address those and I'll be happy to merge. Thanks for hanging with me!

rinna11:
@jnovinger , I have rebased the main branch and believe the conflicts have been resolved. Thank you for your patience!

netbox/ipam/graphql/filters.py
```diff
@@ -101,6 +101,13 @@ def contains(self, value: list[str], prefix) -> Q:
             q |= Q(**{f"{prefix}prefix__net_contains": query})
         return q
 
+    def family(
```
jnovinger, line 105: I think you lost the `@strawberry_django.filter_field()` decorator application during rebase.

jnovinger (changes requested):

jnovinger (approved):

PR #20650, opened 2025-10-21
Closes #20459 : add is_oob and is_primary on IPAddress bulk_edit

### LINKED ISSUE #20459: is_oob and/or is_primary do not update to false in bulk update
### NetBox Edition

NetBox Community

### NetBox Version

v4.3.7

### Python Version

3.11

### Steps to Reproduce

1. Choose an IP Address with Primary (and/or OOB) status as true
2. Go to bulk import of IP Addresses
3. Import id, interface, device, is_oob (and/or is_primary) setting as false
For instance:
id, interface, device, is_oob
744, Eth1, HOSTNAME, false

### Expected Behavior

IP Address updated with OOB (and/or Primary) status as False

### Observed Behavior

IP Address updated without errors but the status remains True.

It works when the status is False and it is updated to True.

pl0xym0r:
### Closes: #20459 

Added parameters is_oob and is_primary  on the IPAddress bulk_edit.
Logic + validation is done in the view.
Added tests.

Capture

### DIFF
--- a/netbox/ipam/forms/bulk_edit.py
+++ b/netbox/ipam/forms/bulk_edit.py
@@ -359,11 +359,21 @@ class IPAddressBulkEditForm(NetBoxModelBulkEditForm):
         max_length=200,
         required=False
     )
+    is_oob = forms.NullBooleanField(
+        required=False,
+        widget=BulkEditNullBooleanSelect(),
+        label=_('Is Oob')
+    )
+    is_primary = forms.NullBooleanField(
+        required=False,
+        widget=BulkEditNullBooleanSelect(),
+        label=_('Is Primary')
+    )
     comments = CommentField()
 
     model = IPAddress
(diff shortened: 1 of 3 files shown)


jnovinger:
Thanks for the work, @pl0xym0r. However, #20459 reports that bulk import (CSV) cannot clear `is_primary`/`is_oob` flags to false. This PR adds those fields to bulk edit (the web form) instead, which wasn't requested in the issue.

The original bug still exists: `netbox/ipam/forms/bulk_import.py:414` only handles `if self.cleaned_data.get('is_primary'):`, which doesn't evaluate when the value is false-y.

Are you still interested in fixing the original bug? If you want to pursue the bulk edit functionality, that will require a separate, accepted feature request first.

Either way, I think it's appropriate to close this particular PR.

PR #20657, opened 2025-10-22
Closes #20459 : clean is_oob and is_primary on bulk_import

### LINKED ISSUE #20459: is_oob and/or is_primary do not update to false in bulk update

pl0xym0r:
Closes: #20459 

I used the same logic as https://github.com/netbox-community/netbox/blob/main/netbox/dcim/forms/bulk_import.py#L768 in order to clean the parameters coming from a CSV.
Then, once data is cleaned, apply if parameter are true or false on assigned_objects.
Test are ok

### DIFF
--- a/netbox/ipam/forms/bulk_import.py
+++ b/netbox/ipam/forms/bulk_import.py
@@ -369,6 +369,20 @@ class IPAddressImportForm(NetBoxModelImportForm):
                     **{f"virtual_machine__{self.fields['virtual_machine'].to_field_name}": data['virtual_machine']}
                 )
 
+    def clean_is_primary(self):
+        # Make sure is_primary is None when it's not included in the uploaded data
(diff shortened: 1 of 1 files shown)


jnovinger (approved):
Thanks, @pl0xym0r

Issue #20613, opened 2025-10-17
Align GraphQL lookup types for booleans & numerics (replace `FilterLookup`)


pheus:
### NetBox version

v4.4.4

### Feature type

Change to existing functionality

### Proposed functionality

To keep GraphQL filter inputs consistent and aligned with Strawberry‑Django, I’d like to suggest:

- **Booleans:**

  **Option A (preferred): use a plain `bool` on *all* boolean filter fields**  
  This keeps input shapes simple and ergonomic for typical use.

  ```py
  # v2 boolean filters – applied consistently across all fields
  is_active: bool | None = strawberry_django.filter_field()
  ```

  **Option B (alternative): use `BaseFilterLookup[bool]` on *all* boolean filter fields**  
  Choose this if we want to expose `exact`, `isNull`, and `inList` uniformly for tri‑state scenarios.

  ```py
  from strawberry_django import BaseFilterLookup

  # v2 boolean filters – applied consistently across all fields
  is_active: BaseFilterLookup[bool] | None = strawberry_django.filter_field()
  ```

  > The intent is **consistency**: adopt **Option A** (recommended) *or* **Option B** across the entire v2 schema, but don’t mix them by field.

- **Numerics (e.g., counts/lengths):** adopt `ComparisonFilterLookup[...]` so v2 clients can use `exact/gt/gte/lt/lte/range`.

  ```py
  from strawberry_django import ComparisonFilterLookup

  device_bay_count: ComparisonFilterLookup[int] | None = strawberry_django.filter_field()
  ```

- **Strings:** continue using `FilterLookup[str]` where substring/regex matching is desired.

### Use case

- With **Option A (preferred)**, boolean queries remain compact and easy to read:

  ```graphql
  # exact match using plain bool
  query {
    device_type_list(filters: { is_full_depth: true }) { id }
  }
  ```

- If **Option B** is selected, nullable/tri‑state patterns remain idiomatic:

  ```graphql
  # exact true
  query {
    device_type_list(filters: { is_full_depth: { exact: true } }) { id }
  }

  # explicit null check
  query {
    device_type_list(filters: { is_full_depth: { is_null: true } }) { id }
  }

  # exclude nulls (generic builder pattern)
  query {
    device_type_list(filters: { is_full_depth: { in_list: [true, false] } }) { id }
  }
  ```

- Numeric comparisons become first‑class:

  ```graphql
  query {
[...]

Issue #20153, opened 2025-08-21
GraphQL Ordering Support


sleepinggenius2:
### NetBox version

v4.3.6

### Feature type

New functionality

### Proposed functionality

With the implementation of pagination in #16224, I thought that might have also introduced ordering, but that does not seem to be the case. Therefore, I propose `ordering` as an additional argument alongside `filters` and `pagination` when accessing non-scalar types. It would work the same as the `ordering` argument in the REST API, but should probably be turned into an array of strings instead of a single comma-separated string.

### Use case

I need to pull nested data from NetBox and using the REST API does not allow that without N+1 queries. Initially, I wrote an export template to solve this problem, but as the output format needs to be JSON, that started becoming tedious, so I decided to explore the possibility of using GraphQL instead. It ended up being fairly straightforward to get the shape of the data I needed, but I was disappointed to find that ordering was not possible. In my case, I am grabbing the members of a virtual chassis and I need them in order of `vc_position`, not `name`, as they seem to be by default. My hope was that I could unmarshal the JSON response directly without additional processing.

### Database changes

None

### External dependencies

None

arthanson:
Strawberry does have ordering support https://strawberry.rocks/docs/django/guide/ordering this would be a fairly substantial change as ordering enum and input would need to be be defined for all model types.

jeremystretch:
I think we have a fairly good understanding of what this entails, however it's still to be determined whether this could effect a breaking change.

Issue #19736, opened 2025-06-20
Namespace GraphQL queries by app


jeremystretch:
### NetBox version

v4.3.2

### Feature type

Change to existing functionality

### Proposed functionality

Currently, the GraphQL API is flat: All model queries are available under the root query. For example, querying sites:

```
{
    site_list {
        id
        name
    }
}
```

This FR proposes nesting queries within their core app or plugin, to effectively namespace queries and prevent collisions. The above query would become:

```
{
    dcim {
        site_list {
            id
            name
        }
    }
}
```

### Use case

This change will protect against name collisions in the case where two plugins have a model with the same name, or where a plugin model has the same name as a core model.

### Database changes

None

### External dependencies

None

cruse1977:
Q - will this also namespace types; as a type collision also came up recently (I /think/ its the same thing)

jeremystretch:
@cruse1977 I think it's reasonable to include that within this scope, and a good callout. There wouldn't be much benefit to namespacing the entry points if we still leave the potential for naming collisions among GraphQL types.

PR #20667, opened 2025-10-24
Closes #20679 - Duplicate all GraphQL hierarchy objects to separate V1 and V2 namespaces

### LINKED ISSUE #20679: Duplicate all GraphQL component objects into V1 and V2 namespaces
### Proposed Changes

Duplicate all GraphQL hierarchy objects (Schema -> Query -> Type -> Filter) to separately named `V1` versions.

### Justification

Because strawberry requires all objects in the GraphQL hierarchy (Schema -> Query -> Type -> Filter) to be uniquely defined, and complicated by the fact that most of the fields in Type classes are defined using `strawberry.lazy` which looks in specific files for resolving objects, separating V1 and V2 objects by shortcuts such as subclassing or conditional imports is not feasible. The only reasonable way to keep the V1 and V2 schemas separate within the strawberry object space is to duplicate the entire hierarchy and differentiate them with `V1` suffixes, defined in duplicate files with `_v1.py`.

This will add extra work if GraphQL changes that need to apply to both V1 and V2 arise, but it will incentivize us to deprecate V1.

The upside of this is that when V1 is deprecated, cleaning it up will be almost entirely a matter of deleting the `_v1.py` files.

bctiemann:
## Closes: #20679

This has been tested such that the V1 and V2 configurations work as expected as per the `GRAPHQL_DEFAULT_VERSION` setting. A thorough review would be to look at all the field type definitions and ensure that I didn't miss any `V1` suffixes or `_v1` module paths in the `strawberry.lazy` statements.

Note that enums and lookups are not namespaced, though types and filter types are.


jeremystretch:
I've added a basic test to sanity-check the v1/v2 types. It's currently failing on the dummy plugin, which raises the question: How should we handle plugins?

bctiemann:
After some experimentation here's what I think would be a reasonable way to handle plugins, though feedback is welcome.

We populate the schema by automatically pulling in plugin types from the registry:

```
@strawberry.type
class QueryV2(
    UsersQuery,
    ....
    VPNQuery,
    WirelessQuery,
    *registry['plugins']['graphql_schemas'],  # Append plugin schemas
):
```

This last line doesn't distinguish between V1 and V2 types, but it could. I'm thinking we could just presume that any Query object found in the registry with the `V1` suffix is a V1 type and would be added to `QueryV1`, and otherwise would be added to `QueryV2`.

```
In [3]: registry['plugins']['graphql_schemas']
Out[3]: 
[netbox.tests.dummy_plugin.graphql.DummyQueryV1,    # Gets added to QueryV1
 netbox.tests.dummy_plugin.graphql.DummyQuery]      # Gets added to QueryV2
```

This means existing plugins, without doing anything, would have their schemas behave in the V2 way (pagination, `in_list`, etc) and be available only in the V2 schema. This might be disruptive.

To alleviate that, another way to do this (though less consistent) would be to treat plugins in the opposite sense of core NetBox as regards the suffixes, in other words:

```
In [3]: registry['plugins']['graphql_schemas']
Out[3]: 
[netbox.tests.dummy_plugin.graphql.DummyQuery,      # Gets added to QueryV1
 netbox.tests.dummy_plugin.graphql.DummyQueryV2]    # Gets added to QueryV2
```

This would allow existing plugins to retain their current behavior and they could opt-in to the V2 behavior by creating `V2` objects in the GraphQL schema hierarchy.

However, the downside of this approach (aside from the inconsistency) is that it would generate no pressure to deprecate the old style, and when V1 disappears the unchanged plugins would then "become" V2 schemas with the same disruption as we would have now. But at least it gives plugin authors enough runway to prepare one way or the other.

And another downside is that once we deprecate V1, the plugin GraphQL schemas would have a lot of messy `V2` suffixes left over, and we would be facing another technical challenge at that point of how to distinguish between `PluginObjectType` and `PluginObjectTypeV2` when core NetBox will (presumably) no longer be attempting to distinguish between the versions based on the suffix and thus would try to load both versions in plugins that have them. (We would probably have to establish another deprecation schedule where we keep looking for the `V2` suffixes in plugins only, up to a certain sunset date when we require plugins to have removed their old versions.)

Is all the rigmarole of this second approach worthwhile weighed against the disruption of existing plugins silently switching to the V2 behavior as in the first approach?

Comments welcome.

pheus:
Thanks for pushing the V1/V2 namespacing work and for asking about plugin handling - much appreciated.

_Small note:_ Since this PR was opened, there have been several GraphQL changes on `main` (#20422, #20466, #20541). It’d be great to ensure we don’t regress any of those changes as we wire up the versioned schemas. Thanks!

From the plugin side, I ported two plugins from NetBox v4.2 GraphQL to v4.3, which was fairly disruptive. I think it helps to separate two concerns:

- **API users** shouldn’t see mixed behavior under V2.
- **Plugin authors** deserve a clear migration path.

To keep V2 clean, I’d suggest exposing **non‑ported plugin schemas only under V1**. Once a plugin is updated, it can add a proper V2 schema. That way, installing a not‑yet‑ported plugin doesn’t create a mixed V1/V2 API surface within V2.

Based on the open PRs around GraphQL V2, it looks like plugin authors will need to adopt at least the following:

- Schema: `OffsetPaginated[Type] = strawberry_django.offset_paginated()`
- Filters: `BaseFilterLookup` for enums
- Support for #19736
- Support for #20153
- Possibly support for #20613

For discovery, keeping an explicit **`V2` suffix (or another clear indicator)** for plugin Query types seems workable. If a plugin isn’t ported, it simply won’t appear in the V2 schema; when V1 is eventually removed, those plugins naturally drop out until they’re updated.

Happy to help test with my plugins if that’s useful. If I’ve misunderstood anything above, please correct me - and thanks again for driving this forward!

bctiemann:
Superseded by https://github.com/netbox-community/netbox/pull/20829.

PR #20674, opened 2025-10-24
Fixes #20653: Add `object_type_id` filter for Jobs

### LINKED ISSUE #20653: Filtering Jobs by Object Type return empty set
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.4

### Python Version

3.10

### Steps to Reproduce

1. Go to core/jobs
2. The form lists all your jobs
3. Filter by Object Type
4. The form is now empty 

### Expected Behavior

Filtering by a valid object type should return all matching jobs.

### Observed Behavior

Image

After filter

Image

pheus:
#### Summary
Fix Jobs list returning empty results when filtering by **Object Type** by introducing a proper `object_type_id` filter and exposing it in the Jobs filter form/fieldsets.

#### Changes
- Add `object_type_id` to the Jobs FilterSet.
- Update the Jobs filter form and fieldsets to include it.
- No model/schema changes or migrations.

### DIFF
--- a/netbox/core/api/serializers_/jobs.py
+++ b/netbox/core/api/serializers_/jobs.py
@@ -1,8 +1,13 @@
+from drf_spectacular.utils import extend_schema_field
+from rest_framework import serializers
+
 from core.choices import *
 from core.models import Job
+from netbox.api.exceptions import SerializerNotFound
 from netbox.api.fields import ChoiceField, ContentTypeField
 from netbox.api.serializers import BaseModelSerializer
 from users.api.serializers_.users import UserSerializer
+from utilities.api import get_serializer_for_model
 
 __all__ = (
     'JobSerializer',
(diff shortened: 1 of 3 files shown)


arthanson (changes requested):
Good work, but I think this is still broken in the REST API - would be good to fix that in this PR.  There is an object_type_id in the REST filter but sending the same filter param as in the UI returned 0 results instead of the two that I had in the REST API.

pheus:
Thanks for the thorough review - much appreciated!

I tried to reproduce the REST issue but I’m seeing consistent results between the UI and the API. On my dev instance (301 jobs total), filtering for `extras.Script` yields 291 matches both ways:

- **UI:** Object Type → *Extras / Script* → 291
- **API:**
  1) Get the ContentType ID  
     `GET /api/extras/object-types/?app_label=extras&model=script` → id=2
  2) Use it to filter  
     `GET /api/core/jobs/?object_type_id=2` → 291  
     (Also tried `GET /api/core/jobs/?object_type=extras.script` with the same result.)

Could you share the exact request you used, the Object Type (e.g. `app_label.Model` or the ContentType ID)?

I’ll reproduce your case and update the PR accordingly. Thanks again for taking a close look!

arthanson (approved):
Thanks @pheus

PR #20680, opened 2025-10-24
Closes #19338 - Add in_list and exact support on all id and enum fields in GraphQL v2

### LINKED ISSUE #19338: Add "in_list" operator to GraphQL API filters for IDs and Enums
### NetBox version

v4.3.0

### Feature type

### Proposed functionality

Introduce support for the in_list operator across all GraphQL filters for ID and Enum fields (like status or interface types) in NetBox's v4.3.0, aligning with most other filter options for other fields in the NetBox GraphQL API.

This operator allows filtering objects by specifying multiple values simultaneously using a concise syntax:

```graphql
query {
  circuit_list {
    terminations(filter: { site: { id: { in_list: [1, 2] } } }) {
      id
      site { name }
    }
  }
}
```

### Use case

This feature significantly simplifies and optimizes common automation tasks and data integrations. Examples include:

### Batch Object Retrieval

Quickly query multiple objects by IDs or statuses within one GraphQL request. While GraphQL does offer other logical operators like OR, they don't cleanly fit every use case where filtering by multiple values is needed.

```graphql
query {
  device_list(filter: { status: { in_list: [ACTIVE, PLANNED, STAGED] } }) {
    id
    name
    status
  }
}
```

### Dashboard Filtering

Simplify dashboards displaying entities filtered by various criteria, making queries cleaner and easier to maintain.

The REST API currently supports similar functionality, and the v4.3.0 GraphQL API also provides similar capability on most endpoints/field filters. Adding in_list to the current GraphQL filtering mechanism for IDs and Enums continues this existing theme, eliminating the need for chaining multiple "OR" filters - which are difficult to produce dynamically without awkward/cumbersome string manipulation.

The ability to combine in_list, OR filters, and pagination would provide maximum API flexibility; serving diverse use cases.

### Database changes

N/A

### External dependencies

N/A

bctiemann:
### Closes: #19338

Converts all `id` fields and all enum-based fields to `FilterLookup`s, which support `in_list` and `exact` querying.

- Changes `BaseObjectTypeFilterMixin` and `ChangeLogFilterMixin` (which are used by most if not all field types) to define `id` using a `FilterLookup` instead of `UNSET`.
- In all apps' `filters.py`, all fields defined as a choice of enum values has similarly been redefined using a `FilterLookup`, with the same new lookup support.

Note: The linter really does not like this kind of typing syntax, but I wasn't able to find a better way to do it:

```
    authentication_algorithm: (
        FilterLookup[Annotated['AuthenticationAlgorithmEnum', strawberry.lazy('vpn.graphql.enums')]] | None
    ) = (
    )
```

**Note:** Depends on https://github.com/netbox-community/netbox/pull/20667


pheus:
_Not a maintainer or reviewer - just dropping a quick thought. Thanks for the PR!_

From my understanding, `FilterLookup` is mostly aimed at string fields and exposes string‑specific operators (e.g., `contains`, `icontains`, `startswith`, `endswith`). If the goal here is to limit the available filters for this field, would `BaseFilterLookup` be a better fit so we only surface the more generic lookups (like `exact`, `in`, `isnull`) and avoid string‑specific ones that may not apply?

If I’m off base or there’s a use case for the string lookups here, please feel free to ignore. Appreciate the work on this!

bctiemann:
@pheus You could be right; I only grabbed `FilterLookup` because it seemed more the recommended route (in more generic use cases than this one). Thanks for the pointer!

jeremystretch:
Marking this as blocked by #20679 temporarily

bctiemann:
Superseded by https://github.com/netbox-community/netbox/pull/20829

PR #20681, opened 2025-10-24
Closes #12318: Ignore case when determining uniquess of name & slug fields

### LINKED ISSUE #12318: Extend case-insensitive uniqueness constraints
### NetBox version

v3.4.8

### Feature type

### Proposed functionality

Introduced in v3.4.0: _"Device and virtual machine names are no longer case-sensitive"_

Should this be extended to other models which have unique constraints on name, in particular Site and Tag?

### Use case

Same reason as doing this on devices and VMs: for data validation purposes.

It seems weird to allow site "foo bar" and "Foo bar" as two different sites, when site names are supposedly "unique".

### Database changes

I believe enforcement is via a database btree index

### External dependencies

None

jeremystretch:
### Closes: #12318

- Introduce two new database collations for `und-u-ks-level2` (case-insensitive) and `und-u-kn-true-ks-level2` (case-insensitive natural sort)
- Alter all model CharFields which are referenced for uniqueness to use a case-insensitive collation in the PostgreSQL database
- Remove all `varchar_pattern_ops` indexes attached to these fields (not supported with non-deterministic collations)
- Alter the UniqueConstraints on the Device and VirtualMachine models to reference the `name` field without calling `Lower()`


jeremystretch:
Well, it seems this approach isn't going to work. The non-deterministic collation causes issue not only for the indexes: It also breaks any kind of `LIKE` query (e.g. `name__icontains='foo'`). This limitation unfortunately disqualifies the collation approach to ensuring case-insensitive uniqueness. I'm going to abandon this work and submit an alternative PR using `Lower()` to normalize field values for uniqueness evaluation.

PR #20687, opened 2025-10-26
Closes #20685: add cloneable custom fields to templates

### LINKED ISSUE #20685: Clone Custom Fields on Component Template Instantiation
### NetBox version

v4.4.4

### Feature type

### Proposed functionality

It seems that none of the component templates have a "custom fields" section. As someone who performs large scale ingests of data centers, I would like the ability to define "templated custom fields" that can be cloned to the instantiated object. 

### Use case

We illustrate the use case with a practical example: suppose that I am instantiating a leaf-spine topology in Netbox. I've created a "interface roles" custom field which describes how each templated interface is meant to behave (i.e., this is an uplink/downlink/BGP speaker, etc.). When I'm instantiating this large fabric, I'd like the child devices to initialize with these roles.

This extends to any "enrichment" of a template with some context of the component.

### Database changes

This will require that the component template models add field for custom fields, most likely. This will correspond to "all cloneable custom fields for the component".

### External dependencies

Nope

anshulkamath:
### Closes: #20685

adds the ability to assign custom fields to component templates. any template with a cloneable custom field will be cloned to the child instance of the template.


jeremystretch:

PR #20695, opened 2025-10-27
Add contacts field to ip addresses table view

### LINKED ISSUE #20692: Add contacts field to ip addresses table view
### NetBox version

v.4.4.4

### Feature type

### Proposed functionality

I would like to suggest adding the Contacts field to the IP address table view settings. Currently, this field is unavailable for selection in the table configuration.

### Use case

When viewing or exporting a list of IP addresses, it would be nice to see contact persons.

### Database changes

_No response_

### External dependencies

_No response_

stalker-nn:


jeremystretch (approved):
Thanks @stalker-nn!

PR #20702, opened 2025-10-28
Closes #20647: add cleanup for interface import

### LINKED ISSUE #20647: Interface Bulk Upload can create invalid WWN values
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.4

### Python Version

3.10

### Steps to Reproduce

1. Create a device example - "dmi-akron-sw01", 
2. Bulk Upload an interface with the following data

```
device, name, type, wwn
"dmi01-akron-sw01", "test-interface2", "25gbase-x-sfp28", ""
```

### Expected Behavior

 In the changelog, wwn should either be set to null, or disallowed due to an empty string

### Observed Behavior

 Interface allows to be created - however this then prevents branches from merging:

```
        "Applying change DCIM | interface Eth1/1/1 created by admin using default",
        "['Invalid WWN format: ']: (dcim.interface:pk=81563) field_value was ''"
```

  Leaving this blank in the UI and creating an interface WWN is always set to null:

Image

 However with the CSV its "" - causing branch merges to fail:

Image

cmdrrobin:
### Fixes: #20647 

When importing interfaces and no `wwn` or `mode` are set, by default it uses an empty string. This should be None.

### DIFF
--- a/netbox/netbox/forms/base.py
+++ b/netbox/netbox/forms/base.py
@@ -2,6 +2,7 @@ import json
 
 from django import forms
 from django.contrib.contenttypes.models import ContentType
+from django.db import models
 from django.db.models import Q
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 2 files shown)


cmdrrobin:
I have added this to the import, but I was thinking maybe it would be better to add this to the model, `netbox/dcim/models/device_components.py`, under `clean()` function of `BaseInterface` class model.

cmdrrobin:
scratch this PR, I need to find a better way.

After further investigation, I noticed that the following fields are set to "" when importing an interface:

```
wwn: ""
mode: ""
duplex: ""
rf_role: ""
poe_mode: ""
poe_type: ""
```

UPDATE: Once I have discovered this, @pheus came with the same conclusion and has participated with the PR.

cmdrrobin:
Thanks to @pheus for participating into this PR and bringing the correct fix on all nullable fields.

jnovinger (approved):
Thanks @cmdrrobin and @pheus!

PR #20704, opened 2025-10-28
Fixes #20697: Ensure unique signal handlers for counter models

### LINKED ISSUE #20697: Per‑sender duplicate signal registration causes CounterCacheField double‑counting
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.4

### Python Version

3.10

### Steps to Reproduce

1. Have two `CounterCacheField` registrations that both use **`dcim.Device`** as the child model, for example:
   - `DeviceType.device_count` (count devices by `device_type`)
   - `VirtualChassis.member_count` (count devices by `virtual_chassis`)
2. Create or delete a single `Device`.

### Expected Behavior

Each create/delete updates the relevant counters **once** (e.g., `device_count += 1` on create, `device_count -= 1` on delete).

### Observed Behavior

Counters update **twice** (e.g., `+2` on create, `-2` on delete).

### Root Cause
In the counters registry (`utilities/counters.py`), the `dispatch_uid` for signal connections is currently formed from the **model and field** (conceptually `{model._meta.label}.{field.name}`).  
When multiple counters share the **same sender** (e.g., `Device` via `device_type_id` and `virtual_chassis_id`), this yields **different UIDs**, so Django connects the **same** receiver **multiple times** for that sender. A single `post_save` emission for `Device` is then handled twice, doubling the counter updates.

### Related Work
This was uncovered while developing FR **#19523** (instance/usage filtering), which depends on accurate `device_count`/`module_count` values.

pheus:
### Fixes: #20697

**Summary**  
Prevent double counting in `CounterCacheField` by ensuring signal receivers are connected **once per sender**.

**Implementation**
- Update `utilities.counters.connect_counters()` to use per‑sender `dispatch_uid` for `post_save`, `pre_delete`, and `post_delete`.
- Guard against reconnecting by tracking models already processed during registration.
- Leave a single receiver to update all registered counters for the same sender (e.g., `device_count` and `member_count`).

**Impact**
- Fixes `+2/−2` updates when multiple counters share the same child model (e.g., `Device`).
- No database changes; backward‑compatible internal wiring change.

### DIFF
--- a/netbox/utilities/counters.py
+++ b/netbox/utilities/counters.py
@@ -77,7 +77,7 @@ def post_delete_receiver(sender, instance, origin, **kwargs):
         parent_pk = getattr(instance, field_name, None)
 
         # Decrement the parent's counter by one
-        if parent_pk is not None and not hasattr(instance, "_previously_removed"):
+        if parent_pk is not None and not hasattr(instance, '_previously_removed'):
             update_counter(parent_model, parent_pk, counter_name, -1)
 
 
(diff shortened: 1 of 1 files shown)


pheus:
Note: This bug only manifests when multiple counters are registered for the **same** model/sender. That isn’t the case in core yet, so I didn’t add a failing test here without introducing test‑only fields. Once the PR for **#19523** is ready, it will include a second counter for `Device`, which gives us a clean repro to test against. I’m happy to add/enable the test in that PR (or as a follow‑up). I hope that’s acceptable.

pheus:
I can push my current WIP for `device_count` and `module_count` as a draft PR to make it easier to review this change. Let me know if that would be helpful.

jeremystretch (approved):
Thanks @pheus!

Issue #17936, opened 2024-11-05
Introduce a serializer field for GenericForeignKeys


jeremystretch:
### Proposed Changes

Devise and implement a custom serializer field to represent GenericForeignKeys on models. This is currently handled using a read-only [`SerializerMethodField`](https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield), which requires a boilerplate method be included on the serializer class. For instance, the code below is from the serializer for circuit terminations:

```python
    @extend_schema_field(serializers.JSONField(allow_null=True))
    def get_termination(self, obj):
        if obj.termination_id is None:
            return None
        serializer = get_serializer_for_model(obj.termination)
        context = {'request': self.context['request']}
        return serializer(obj.termination, nested=True, context=context).data
```

### Justification

Introducing a custom field to represent GFK relationships allows us to enforce a standard implementation and reduce boilerplate.

jeremystretch:
As part of this change, we now expose the `object_repr` field on the `/api/core/object-changes/` endpoint.

PR #20706, opened 2025-10-28
Closes: #17936 - GFK serializer field

bctiemann:


jeremystretch (changes requested):
I found two serializers with generic foreign keys where GFKSerializerField is not being employed, presumably due to variations in the resolution logic.

- `ObjectChangeSerializer.changed_object`: Suggest we replace the SerializerMethodField with GFKSerializerField, and expose `object_repr` as a regular read-only field. IMO this cleaner than the current fallback behavior.
- `EventRuleSerializer.action_object`: I don't think the custom logic under `get_action_object()` is necessary anymore as there is a now a Script model: The automatic resolution should work.

bctiemann:
Added those two; tested successfully, works nice. Added `ObjectChangeSerializer.object_repr` as a read-only `CharField`.

jeremystretch (approved):

PR #20718, opened 2025-10-30
Closes #19523: Add counter‑cached counts & filters for DeviceType, ModuleType, and RackType

### LINKED ISSUE #19523: Device Types - Filter - With Instances (Yes / No)

pheus:
### Fixes: #19523

**Summary**  
Extend usage-count filtering via counter caches to cover **DeviceType**, **ModuleType**, and **RackType**.

**Implementation**
- Add counter‑cached integer fields:
  - `device_count` on **DeviceType**
  - `module_count` on **ModuleType**
  - `rack_count` on **RackType**
- Expose numeric lookups in **REST** and **GraphQL** (e.g., `device_count__gte=10`, `module_count=0`, `{rack_count: {gt: 2}`).
- Include migration to add fields and backfill counts.
- Add a test.

**Database Changes**  
Adds integer fields on the three models listed above.

### DIFF
--- a/netbox/dcim/migrations/0218_devicetype_device_count.py
+++ b/netbox/dcim/migrations/0218_devicetype_device_count.py
@@ -1,0 +1,66 @@
+import utilities.fields
+from django.db import migrations
+from django.db.models import Count, OuterRef, Subquery
+
+
+def _populate_count_for_type(
+    apps, schema_editor, app_name: str, model_name: str, target_field: str, related_name: str = 'instances'
+):
+    """
+    Update a CounterCache field on the specified model by annotating the count of related instances.
+    """
+    Model = apps.get_model(app_name, model_name)
+    db_alias = schema_editor.connection.alias
+
+    count_subquery = (
+        Model.objects.using(db_alias)
+        .filter(pk=OuterRef('pk'))
+        .annotate(_instance_count=Count(related_name))
+        .values('_instance_count')
+    )
+    Model.objects.using(db_alias).update(**{target_field: Subquery(count_subquery)})
+
+
+def populate_device_type_device_count(apps, schema_editor):
+    _populate_count_for_type(apps, schema_editor, 'dcim', 'DeviceType', 'device_count')
+
+
+def populate_module_type_module_count(apps, schema_editor):
+    _populate_count_for_type(apps, schema_editor, 'dcim', 'ModuleType', 'module_count')
+
+
+def populate_rack_type_rack_count(apps, schema_editor):
+    _populate_count_for_type(apps, schema_editor, 'dcim', 'RackType', 'rack_count', related_name='racks')
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('dcim', '0217_owner'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='devicetype',
+            name='device_count',
+            field=utilities.fields.CounterCacheField(
+                default=0, editable=False, to_field='device_type', to_model='dcim.Device'
+            ),
+        ),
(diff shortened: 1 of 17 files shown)


netbox/dcim/migrations/0218_devicetype_device_count.py
```diff
@@ -0,0 +1,25 @@
+import utilities.fields
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
jeremystretch, line 36: We need to populate the field values as part of the migration as well (using a function called via `RunPython`), to ensure the counts are accurate. Without this, the counts all default to zero for existing objects.

jeremystretch (changes requested):
Nice work @pheus!

netbox/dcim/migrations/0218_devicetype_device_count.py
pheus: Thanks! Will do.

pheus:
Thanks for the thoughtful review and the migration tip - much appreciated!

I’ve added a **data migration** to prefill `device_count` and `module_count`, and included a test to ensure we don’t double‑register the signal receivers. Happy to adjust anything further.

pheus:
I just noticed an earlier FR comment asking to extend this to **RackType** as well - sorry I missed that. I’m happy to apply the same approach to `RackType` (counter cache + filters). Would you prefer I add it here or open a follow‑up PR? Thanks again!

jeremystretch:
@pheus I think it makes sense to fold that in here as well, since it's a continuation of the same pattern. Thanks!

netbox/dcim/models/racks.py
```diff
@@ -19,9 +19,11 @@
 from netbox.models.features import ContactsMixin, ImageAttachmentsMixin
 from utilities.conversion import to_grams
 from utilities.data import array_to_string, drange
-from utilities.fields import ColorField
+from utilities.fields import ColorField, CounterCacheField
+from utilities.tracking import TrackingModelMixin
 from .device_components import PowerPort
-from .devices import Device, Module
+from .devices import Device
+from .modules import Module
```
pheus, line 26: _Note_: Unrelated import fix. Happy to drop this if you'd prefer to keep the PR focused.

netbox/dcim/models/racks.py
```diff
@@ -509,7 +515,7 @@ def get_rack_units(self, user=None, face=DeviceFaceChoices.FACE_FRONT, exclude=N
 
         return [u for u in elevation.values()]
 
-    def get_available_units(self, u_height=1, rack_face=None, exclude=None, ignore_excluded_devices=False):
+    def get_available_units(self, u_height=1.0, rack_face=None, exclude=None, ignore_excluded_devices=False):
```
pheus, line 518: _Note_: Unrelated cleanup. Happy to drop this if you'd prefer to keep the PR focused.

pheus:
Thanks! I’ve applied the same approach to `RackType` as well. This is ready for review now. Happy to adjust anything further.

jeremystretch (approved):
Thanks @pheus, great work!

PR #20737, opened 2025-11-03
Closes #20204: Introduce modular template components

jeremystretch:
### Closes: #20204

- Implement new Layout, Row, and Column components to replace boilerplate HTML in templates
- Introduce a set of Panels to render cards containing content
- Introduce a set of ObjectAttributes for use within ObjectAttributePanels
- Define programmatic layouts for a number of object views under `dcim`

#### TODO

- [x] Write plugin development documentation for UI components
- [x] Delete the relevant HTML from templates which have been replaced with layouts
- [x] Annotate start & end of panel HTML within templates


netbox/netbox/ui/attrs.py
```diff
@@ -0,0 +1,395 @@
...
+    template_name = 'ui/attrs/text.html'
+
+    def __init__(self, *args, style=None, format_string=None, copy_button=False, **kwargs):
+        """
+        Instantiate a new TextAttr.
+
+        Parameters:
+             accessor: The dotted path to the attribute being rendered (e.g. "site.region.name")
+             label: Human-friendly label for the rendered attribute
+             template_name: The name of the template to render
```
bctiemann: I'm not sure about these parameter signatures in `__init__` for these classes and the ones in `panels.py`. Is `template_name` a valid thing to list as a parameter here when it isn't being consumed in `__init__` and can't be set directly on the class at instantiation?

bctiemann (approved):
Approving but with a couple of general comments --

- I had the thought of adding inline comments in the template snippets so someone debugging from the rendered source would be able to track down where something is coming from. Maybe it isn't that important though, since it'll be obvious that these are panels and someone ought to be able to infer which panel snippet it is
- I'm not seeing any deletions or removal of existing HTML in this PR I believe. Is there a second-phase PR which will shift all the individual views over to the new style? (I think that was the plan if I recall)

jeremystretch:
> I had the thought of adding inline comments

Good idea; should help with troubleshooting.

> I'm not seeing any deletions or removal of existing HTML in this PR

I captured the deletion of the legacy HTML context under the TODO items (along with writing the docs). I wanted to keep it around until the end for easier testing.

bctiemann:
Looks good to me, aside from the failing tests. I was going to rerun them but it looks like they're actual errors about missing HTML templates?

jeremystretch:
There must be something left that depends on the named object template (e.g. `racktype.html`) existing. I'll restore the two deleted templates and just have them extend `generic/object.html` without any additional content for now.

PR #20747, opened 2025-11-04
Closes #7604: Add filter modifier dropdowns for advanced lookup operators

### LINKED ISSUE #7604: Extend filter forms to support filter modifiers

jnovinger:
### Closes: #7604

Implements dynamic filter modifier UI that allows users to select lookup operators (exact, contains, starts with, regex, negation, empty/not empty) directly in filter forms without manual URL parameter editing.

Supports filters for all scalar types and strings, as well as some related object filters. Explicitly does not support filters on fields that use `APIWidget`. That has been broken out in to follow up work.

#### How It Works

```
User selects "contains" ─────────────────┐
                                         │
                                         ▼
┌──────────────────────────────────────────────────────────┐
│  Filter Form (Browser)                                   │
│  ┌────────────────┐  ┌──────────────────────────┐        │
│  │ [Contains ▼]   │  │ Input: ABC               │        │
│  └────────────────┘  └──────────────────────────┘        │
│         │                                                │
│         │ TypeScript handler (filterModifiers.ts)        │
│         ▼                                                │
│  Updates input name: "serial" -> "serial__ic"            │
│  Updates URL param:  ?serial__ic=ABC                     │
└──────────────────────────────────────────────────────────┘
                         │
                         │ Form submission
                         ▼
┌──────────────────────────────────────────────────────────┐
│  Django Backend                                          │
│                                                          │
│  FilterModifierWidget                                    │
│    ↓ Renders modifier dropdown + original widget         │
│                                                          │
│  FilterModifierMixin                                     │
│    ↓ Enhances filterset fields with appropriate lookups  │
│                                                          │
│  FilterSet processes: serial__ic=ABC                     │
│    ↓ Django ORM: .filter(serial__icontains='ABC')        │
│                                                          │
│  Results + Filter Pills                                  │
│    ↓ "Serial: contains ABC [×]"                          │
└──────────────────────────────────────────────────────────┘
```

##### Backend: `FilterModifierWidget`

Wraps any Django form widget with a modifier dropdown. Key method:

```python
    # Check all possible lookup variants (exact, ic, isw, n, etc.)
    for lookup_code, _ in self.lookups:
        param_name = f"{name}__{lookup_code}" if lookup_code != 'exact' else name
        if param_name in data:
            return data.get(param_name)
```

This allows the widget to find its value regardless of which lookup modifier is active in the URL.

##### Backend: `FilterModifierMixin`

Automatically enhances filterset form fields based on their type:

- **`CharField`** -> `exact`, `contains` (`ic`), `startswith` (`isw`), `endswith` (`iew`), `iexact` (`ie`), `negation` (`n`), `regex`, `iregex`, `empty`
- **`IntegerField`** -> `exact`, `gte`, `lte`, `gt`, `lt`, `negation` (`n`), `empty`
- **`DecimalField`** -> `exact`, `gte`, `lte`, `gt`, `lt`, `negation` (`n`), `empty`
- **`DateField`** -> `exact`, `gte`, `lte`, `gt`, `lt`, `negation` (`n`), `empty`
- **`ChoiceField`** -> `exact`, `negation` (`n`), `empty`
- **`MultipleChoiceField`** -> `exact`, `negation` (`n`), `empty`
- **`ModelChoiceField`** -> `exact`, `negation` (`n`), `empty`
- **`ColorField`** -> `exact`, `negation` (`n`), `empty`
- **`TagFilterField`** -> `exact`, `negation` (`n`), `empty`

##### Frontend: `filterModifiers.ts`

Typescript handler that:
1. Detects modifier selection change
2. Updates the associated input's `name` attribute
3. Syncs URL parameters when navigating with existing filters

##### Filter Pills Enhancement

Modified `applied_filters` template tag to:
1. Detect lookup modifier in parameter name (`serial__ic`)
2. Display human-readable label ("contains") instead of raw lookup code
3. Maintain correct removal URL (preserves other filters)

#### Scope & Compatibility

**What's changed:**
- All `FilterForm` subclasses now have modifier dropdowns
- Filter pills show lookup type
- URL parameters include lookup suffixes (`__ic`, `__n`, etc.)

**What's NOT changed:**
- Existing URL parameters still work (backward compatible)
- Django `FilterSet` behavior unchanged
- No database schema changes
- No API changes


netbox/utilities/forms/filterset_mappings.py
```diff
@@ -0,0 +1,11 @@
...
+# This enables the FilterModifierMixin to verify which lookups are actually supported
+# by checking the FilterSet's auto-generated lookup filters.
+#
+# Usage:
+#   from utilities.forms.filterset_mappings import FILTERSET_MAPPINGS
+#   from .forms.filtersets import XFilterForm
+#   from .filtersets import XFilterSet
+#   FILTERSET_MAPPINGS[XFilterForm] = XFilterSet
+
+FILTERSET_MAPPINGS = {}
```
jeremystretch: We should avoid introducing one-off globals to be used for making associations in this manner. Instead, we can leverage the [application registry](https://netboxlabs.com/docs/netbox/development/application-registry/) to correlate FilterSets to their corresponding models; the models are already set on the filter forms. Then you could do something like this:

```python
@register_filterset
class CircuitFilterSet(PrimaryModelFilterSet, TenancyFilterSet, ContactModelFilterSet):
```

```python
from netbox.registry import registry
filterset = registry['filtersets'].get(f'{app_label}.{model}')
```

We take a similar approach to map models to their search indexes (`registry['search']`).

netbox/circuits/forms/filtersets.py
```diff
@@ -366,7 +374,7 @@ class VirtualCircuitFilterForm(TenancyFilterForm, ContactModelFilterForm, Primar
     tag = TagFilterField(model)
 
 
-class VirtualCircuitTerminationFilterForm(NetBoxModelFilterSetForm):
+class VirtualCircuitTerminationFilterForm(FilterModifierMixin, NetBoxModelFilterSetForm):
```
jeremystretch: Any reason not to have NetBoxModelFilterSetForm inherit FilterModifierMixin upstream?

netbox/utilities/forms/mixins.py
```diff
@@ -75,3 +79,177 @@ class DistanceValidationMixin(forms.Form):
...
+class FilterModifierMixin:
+    """
+    Mixin that enhances filter form fields with lookup modifier dropdowns.
+
+    Automatically detects fields that could benefit from multiple lookup options
+    and wraps their widgets with FilterModifierWidget.
+    """
+
+    # Mapping of form field types to their supported lookups
+    FORM_FIELD_LOOKUPS = {
```
jeremystretch: Suggest moving this to a constant outside the mixin class.

netbox/utilities/forms/mixins.py
```diff
@@ -75,3 +79,177 @@ class DistanceValidationMixin(forms.Form):
...
+            if filterset:
+                lookups = self._verify_lookups_with_filterset(field_name, lookups, filterset)
+
+            if len(lookups) > 1:
+                field.widget = FilterModifierWidget(
+                    original_widget=field.widget,
+                    lookups=lookups
+                )
+
+    def _should_skip_field(self, field_name, field):
```
jeremystretch: Could we move this logic into `_get_lookup_choices()`, and just have it return an empty list for inapplicable fields?

netbox/utilities/forms/mixins.py
```diff
@@ -75,3 +79,177 @@ class DistanceValidationMixin(forms.Form):
...
+        """Determine if a field should be skipped for enhancement."""
+        # Skip the global search field
+        if field_name == 'q':
```
jeremystretch: The baked-in assumption here gives me pause. Should we consider converting `q` on NetBoxModelFilterSetForm to a custom subclass of CharField (e.g. `QueryField` or something), and exclude the field based on its type instead, as we do with boolean fields below? Trying to avoid a potential conflict with a hypothetical plugin which introduces a `q` filter for some other purpose.

netbox/utilities/forms/mixins.py
```diff
@@ -75,3 +79,177 @@ class DistanceValidationMixin(forms.Form):
...
+            # Check if this filter exists in the FilterSet
+            if filter_key in filterset.filters:
+                verified_lookups.append((lookup_code, lookup_label))
+
+        return verified_lookups
+
+    def _is_api_widget_field(self, field):
+        """Check if a field uses an API-based widget."""
+        # Check field class name
+        if 'Dynamic' in field.__class__.__name__:
```
jeremystretch: I'd prefer to avoid making any assumptions about a field based on its name alone. Is this check needed? I would assume the attributes check below is sufficient to exclude API-backed fields.

netbox/utilities/forms/mixins.py
```diff
@@ -75,3 +79,177 @@ class DistanceValidationMixin(forms.Form):
...
+        # MultipleChoiceField and TagFilterField are now supported
+        # (no longer skipped)
+
+        # Skip API widget fields
+        if self._is_api_widget_field(field):
+            return True
+
+        return False
+
+    def _get_lookup_choices(self, field, field_name=None):
```
jeremystretch: `field_name` is unused?
```suggestion
    def _get_lookup_choices(self, field):
```

netbox/utilities/templatetags/helpers.py
```diff
@@ -418,22 +419,69 @@ def applied_filters(context, model, form, query_params):
...
 
+        # Special handling for empty lookup (boolean value)
+        if modifier == 'empty':
+            if display_value.lower() in ('true', '1'):
+                link_text = f'{bound_field.label} {_("is empty")}'
+            else:
+                link_text = f'{bound_field.label} {_("is not empty")}'
+        else:
+            # Add friendly lookup label for other modifier-enhanced fields
+            lookup_labels = {
```
jeremystretch: Feels redundant to `FORM_FIELD_LOOKUPS`. Can we reuse that here?

netbox/utilities/templatetags/helpers.py
```diff
@@ -418,22 +419,69 @@ def applied_filters(context, model, form, query_params):
...
+                'iew': _('ends with'),
+                'ie': _('equals (case-insensitive)'),
+                'regex': _('matches pattern'),
+                'iregex': _('matches pattern (case-insensitive)'),
+                'gt': _('>'),
+                'gte': _('≥'),
+                'lt': _('<'),
+                'lte': _('≤'),
+            }
+            if modifier != 'exact' and modifier in lookup_labels:
```
jeremystretch: Can we simplify this a bit to check for `label := lookup_labels.get(modifier)`?

netbox/utilities/forms/widgets/modifiers.py
```diff
@@ -0,0 +1,105 @@
...
+class FilterModifierWidget(forms.Widget):
+    """
+    Wraps an existing widget to add a modifier dropdown for filter lookups.
+
+    The original widget's semantics (name, id, attributes) are preserved.
+    The modifier dropdown controls which lookup type is used (exact, contains, etc.).
+    """
+    template_name = 'widgets/filter_modifier.html'
+
+    def __init__(self, original_widget, lookups, attrs=None):
```
jeremystretch: Maybe just call this `widget` in the signature? Seems a bit cleaner.
```suggestion
    def __init__(self, widget, lookups, attrs=None):
```

jeremystretch (changes requested):
I haven't quite completed a review (still need to go through the Typescript), but wanted to deliver some suggestions and clarify some pieces. Great work!

jnovinger:
Just realized I misread your comment about the registry _**twice**_. 🤦‍♂️ 

Will get that sorted shortly.

jnovinger:
Just realized I'll also need to update the 'Application Registry' section of the docs for the new `filtersets` store.

netbox/circuits/forms/filtersets.py
jnovinger: Nope, done.

netbox/utilities/forms/mixins.py
jnovinger: Done

netbox/utilities/forms/mixins.py
jnovinger: Surprisingly, something like that was still required, yes. I ended up changing it to recognize the fields based by type-checking the widget. FWIW, this will be completely removed by the follow up work for API-based fields.

netbox/utilities/filtersets.py
```diff
@@ -0,0 +1,13 @@
+from netbox.registry import registry
+
+
+def register_filterset(filterset_class):
```
jeremystretch, line 8: Suggest we move this to `netbox.registration` with the other `register_foo()` functions. IMO it makes sense to expose this to plugins at some point.

netbox/utilities/forms/widgets/modifiers.py
```diff
@@ -0,0 +1,105 @@
+from django import forms
+from django.utils.translation import gettext_lazy as _
+
+__all__ = ('FilterModifierWidget',)
```
jeremystretch: These are referenced externally (by `utilities.forms.mixins`).

```suggestion
__all__ = (
    'FilterModifierWidget',
    'MODIFIER_EMPTY_FALSE',
    'MODIFIER_EMPTY_TRUE',
)
```

netbox/utilities/forms/mixins.py
```diff
@@ -5,13 +5,98 @@
...
     'BackgroundJobMixin',
     'CheckLastUpdatedMixin',
     'DistanceValidationMixin',
+    'FilterModifierMixin',
+    'FORM_FIELD_LOOKUPS',
 )
 
 
+# Mapping of form field types to their supported lookups
+FORM_FIELD_LOOKUPS = {
```
jeremystretch, line 23: How would you feel about moving this to `utilities.constants`? There may be some opportunity to couple this with the filter lookup maps (but we don't have to do that now).

netbox/utilities/forms/mixins.py
```diff
@@ -75,3 +160,89 @@ class DistanceValidationMixin(forms.Form):
...
+
+    def _get_lookup_choices(self, field):
+        """Determine the available lookup choices for a given field.
+
+        Returns an empty list for fields that should not be enhanced.
+        """
+        if isinstance(field, QueryField):
+            return []
+
+        if isinstance(field, (forms.BooleanField, forms.NullBooleanField)):
```
jeremystretch: Is this needed? AFAICT only fields listed in `FORM_FIELD_LOOKUPS` can return anything.

netbox/utilities/forms/mixins.py
```diff
@@ -5,13 +5,98 @@
...
+        ('exact', _('Is')),
+        ('n', _('Is Not')),
+        ('ic', _('Contains')),
+        ('isw', _('Starts With')),
+        ('iew', _('Ends With')),
+        ('ie', _('Equals (case-insensitive)')),
+        ('regex', _('Matches Pattern')),
+        ('iregex', _('Matches Pattern (case-insensitive)')),
+        (MODIFIER_EMPTY_TRUE, _('Is Empty')),
+        (MODIFIER_EMPTY_FALSE, _('Is Not Empty')),
```
jeremystretch: Very minor nit here, but IMO we shouldn't use title case for these labels

netbox/utilities/forms/mixins.py
```diff
@@ -75,3 +160,89 @@ class DistanceValidationMixin(forms.Form):
...
+
+        if isinstance(field, (forms.BooleanField, forms.NullBooleanField)):
+            return []
+
+        if self._is_api_widget_field(field):
+            return []
+
+        for field_class in field.__class__.__mro__:
+            if field_class in FORM_FIELD_LOOKUPS:
+                return FORM_FIELD_LOOKUPS[field_class]
```
jeremystretch: One thing I've noticed is that the empty/not empty lookups are added even for mandatory model fields (e.g. `Site.status`). While not necessarily problematic, this can be confusing to the user. However, I don't see any direct path to glean that context via the model's filterset. Not sure it's worth tackling at this time.

netbox/utilities/templatetags/helpers.py
```diff
@@ -412,28 +414,67 @@ def applied_filters(context, model, form, query_params):
     user = context['request'].user
     form.is_valid()  # Ensure cleaned_data has been set
 
+    # Build lookup labels from FORM_FIELD_LOOKUPS
+    lookup_labels = {}
+    for field_lookups in FORM_FIELD_LOOKUPS.values():
```
jeremystretch: Unfortunately we can't use raw values here, because the labels for certain lookups differ depending on the field type. For instance, `gt` is labeled as `greater than` for numeric fields but as `after` for date/time fields. (This can be observed when filtering racks by `u_height__gt`.)

IMO the "most correct" fix would be to rename the `lt` and `gt` lookups for date/time fields to `before` and `after`, but obviously that's a separate breaking change. (Also, there likely exist similar overlaps with other lookups.) Happy to split that off a separate work item if you're in agreement, but open to alternative approaches.

jeremystretch (changes requested):

netbox/utilities/forms/mixins.py
jnovinger: Works for me

jnovinger:
Just finished auditing all of the filter fields exposed in filter forms to identify all of those that don't support the filter modifier UI at the moment. 56 (out of around 670) fields across NetBox don't receive filter modifier dropdowns due to their filterset implementations.

### Custom Filter Methods (31 fields)

| Form | Fields | Method |
|------|--------|--------|
| MACAddressFilterForm | `device_id`, `virtual_machine_id` | `filter_device`, `filter_virtual_machine` |
| CableFilterForm | `site_id`, `location_id`, `rack_id`, `device_id` | `filter_by_termination` |
| IPAddressFilterForm | `device_id`, `virtual_machine_id` | `filter_device`, `filter_virtual_machine` |
| VLANGroupFilterForm | `region`, `site_group`, `site`, `location`, `rack`, `cluster_group`, `cluster` | `filter_scope` |
| ServiceFilterForm | `device_id`, `virtual_machine_id`, `fhrpgroup_id` | `filter_device`, `filter_virtual_machine`, `filter_fhrp_group` |
| L2VPNTerminationFilterForm | `region_id`, `site_id` | `filter_region`, `filter_site` |
| CircuitGroupAssignmentFilterForm | `provider_id` | `filter_provider` |
| CustomFieldChoiceSetFilterForm | `choice` | `filter_by_choice` |
| TagFilterForm | `content_type_id`, `for_object_type_id` | `_content_type_id`, `_for_object_type` |
| EventRuleFilterForm | `event_type` | `filter_event_type` |

### Special Lookup Expressions (20 fields)

| Form | Fields | Lookup Expression |
|------|--------|-------------------|
| IPAddressFilterForm | `family`, `mask_length` | `family`, `net_mask_length` (PostgreSQL) |
| IPRangeFilterForm | `family` | `family` (PostgreSQL) |
| AggregateFilterForm | `family` | `family` (PostgreSQL) |
| VLANGroupFilterForm | `contains_vid` | `range_contains` (PostgreSQL) |
| WebhookFilterForm | `payload_url` | `icontains` |
| JobFilterForm | `created__before/after`, `scheduled__before/after`, `started__before/after`, `completed__before/after` | `lte`/`gte` |
| TokenFilterForm | `expires__gte/lte`, `last_used__gte/lte` | `gte`/`lte` |

### Range Filters (4 fields)

| Form | Fields | Filter Type |
|------|--------|-------------|
| JournalEntryFilterForm | `created_after`, `created_before` | `DateTimeFromToRangeFilter` |
| ObjectChangeFilterForm | `time_after`, `time_before` | `DateTimeFromToRangeFilter` |

### Other Exceptions (3 fields)

| Form | Field | Issue |
|------|-------|-------|
| L2VPNTerminationFilterForm | `assigned_object_type_id` | ContentTypeFilter (custom class) |
| IKEPolicyFilterForm | `proposal_id` | Form uses `proposal_id`, filterset has `ike_proposal_id` |
| IPSecPolicyFilterForm | `proposal_id` | Form uses `proposal_id`, filterset has `ipsec_proposal_id` |

These fields are excluded because:

1. Their filterset implementations don't support standard lookup patterns (`__n`, `__empty`, etc.)
2. Adding support requires implementing custom logic in filterset methods/expressions
3. Supporting these edge cases should probably be deferred for potential future work

netbox/utilities/forms/mixins.py
jnovinger: Yeah, that's an excellent point. I think this is leftover from before I had an explicit mapping. Thanks for catching it.

netbox/utilities/filtersets.py
jnovinger: Do you mean `netbox/utils.py`? That's where I'm seeing three other registration funcs. I also see `netbox/plugins/registration.py`, which contains other registrations for plugins. Otherwise, it seems like app registry functions seems to be scattered around a bit.

netbox/utilities/forms/mixins.py
jnovinger: I ran into an import issue trying to move `FORM_FIELD_LOOKUPS` to `utilities/constants.py`. The problem is that `utilities/constants.py` gets imported very early in Django's startup sequence (via `netbox/choices.py` -> `utilities/constants.py`) before the app registry is ready.

`FORM_FIELD_LOOKUPS` has dependencies on things from Django that don't exist at that point or aren't ready. When these imports are added to `utilities/constants.py`, it triggers a chain that tries to load Django models before the app registry is ready, resulting in `AppRegistryNotReady: Apps aren't loaded yet`.

I could create a new module like `utilities/forms/constants.py` specifically for form-related constants like this, but I think `netbox/constants.py` isn't going to work without some hacky lazy-import machinery.

netbox/utilities/templatetags/helpers.py
jnovinger: Yeah, that feels like a housekeeping thing that we can get in v4.5 in a separate issue.

jnovinger: I did make an attempt to make sure each filter gets the right description here: ~~[`11dbe0d` (#20747)](https://github.com/netbox-community/netbox/pull/20747/commits/11dbe0df955d3f6ada54060351c26f5709b87df5)~~ [`7681450` (#20747)](https://github.com/netbox-community/netbox/pull/20747/commits/768145054971b1d80ad23f99434218e3279bd488)

netbox/utilities/forms/mixins.py
jnovinger: I took a stab at this, by trying to check the the field on form and then on the model. I didn't love it.

netbox/utilities/filtersets.py
jeremystretch: Sorry, I meant [`netbox.plugins.registration`](https://github.com/netbox-community/netbox/blob/main/netbox/netbox/plugins/registration.py), where we have similar functions like `register_graphql_schema()` and `register_menu()`. IMO we should go ahead and make this part of the plugins API. It won't be a mandatory change, but plugin authors can opt to register their filtersets now to leverage the new functionality.

netbox/utilities/forms/mixins.py
jeremystretch: I'm fine leaving it as-is; not worth the effort IMO.

netbox/utilities/forms/mixins.py
```diff
@@ -5,13 +5,100 @@
...
+        (MODIFIER_EMPTY_TRUE, _('is empty')),
+        (MODIFIER_EMPTY_FALSE, _('is not empty')),
+    ],
+    forms.IntegerField: [
+        ('exact', _('is')),
+        ('n', _('is not')),
+        ('gt', _('greater than (>)')),
+        ('gte', _('at least (≥)')),
+        ('lt', _('less than (<)')),
+        ('lte', _('at most (≤)')),
```
jeremystretch: IMO this would look cleaner without including the comparison symbols.

Image

netbox/utilities/templates/widgets/filter_modifier.html
```diff
@@ -0,0 +1,16 @@
+<div class="d-flex filter-modifier-group">
+  {# Modifier dropdown - NO name attribute, just a UI control #}
+  <select class="form-select modifier-select"
+          data-field="{{ widget.field_name }}"
+          data-empty-placeholder="{{ widget.empty_placeholder }}"
+          aria-label="Modifier">
+    {% for lookup, label in widget.lookups %}
+      <option value="{{ lookup }}"{% if widget.current_modifier == lookup %} selected{% endif %}>{{ label }}</option>
+    {% endfor %}
+  </select>
```
jeremystretch: Should we wrap this in a `{% if widget.lookups %}` as a failsafe? Currently it looks like FilterModifierWidget will initialize even with no lookups specified.

netbox/utilities/forms/mixins.py
```diff
@@ -75,3 +162,68 @@ class DistanceValidationMixin(forms.Form):
...
+            key = f'{model._meta.app_label}.{model._meta.model_name}'
+            filterset_class = registry['filtersets'].get(key)
+
+        filterset = filterset_class() if filterset_class else None
+
+        for field_name, field in self.fields.items():
+            lookups = self._get_lookup_choices(field)
+
+            if filterset:
+                lookups = self._verify_lookups_with_filterset(field_name, lookups, filterset)
```
jeremystretch, line 197: Currently, lookups are validated against the underlying filterset only if the filterset is known; otherwise, we assume that the lookups are all valid.

I'm concerned about plugins here, which we should assume likely won't register their filtersets via `register_filterset()` (at least in the short term). It would be safer to populate the additional lookups only if we can validate them against the filterset, and fallback to simple filters if the filterset is unknown.

netbox/utilities/tests/test_filter_modifiers.py
```diff
@@ -0,0 +1,298 @@
...
+            lookups=[('exact', 'Is'), ('ic', 'Contains')]
+        )
+
+        html = widget.render('serial', 'test', {})
+
+        # Should contain modifier dropdown
+        self.assertIn('class="form-select modifier-select"', html)
+        self.assertIn('data-field="serial"', html)
+        self.assertIn('value="exact"', html)
+        self.assertIn('>Is</option>', html)
```
jeremystretch: Can we match by the complete `<option>` tags to make these a bit more sturdy and clear?

netbox/utilities/tests/test_filter_modifiers.py
```diff
@@ -0,0 +1,298 @@
...
+            name = forms.CharField(required=False)
+
+        form = TestForm()
+
+        self.assertIsInstance(form.fields['name'].widget, FilterModifierWidget)
+        self.assertGreater(len(form.fields['name'].widget.lookups), 1)
+        # Should have exact, ic, isw, iew
+        lookup_codes = [lookup[0] for lookup in form.fields['name'].widget.lookups]
+        self.assertIn('exact', lookup_codes)
+        self.assertIn('ic', lookup_codes)
```
jeremystretch: Is there any harm in checking for _all_ of the expected lookups for each field rather than just a couple? This would mitigate accidental omissions.

netbox/utilities/tests/test_filter_modifiers.py
```diff
@@ -0,0 +1,298 @@
...
+        class TestForm(FilterModifierMixin, forms.Form):
+            created = forms.DateField(required=False)
+
+        form = TestForm()
+
+        date_lookups = [lookup[0] for lookup in form.fields['created'].widget.lookups]
+        self.assertIn('n', date_lookups)
+
+
+class ExtendedLookupFilterPillsTest(TestCase):
```
jeremystretch, line 203: Would it be worthwhile to add a test for exact (no lookup) here as well?

jeremystretch (changes requested):

netbox/utilities/tests/test_filter_modifiers.py
jnovinger: Nope, not at all.

netbox/utilities/templates/widgets/filter_modifier.html
jnovinger: I can do that as a safety precaution. Generally speaking, this template should only be rendered if the original widget is wrapped with `FilterModifierWidget`, which only happens when there is more than a single lookup available.

jeremystretch:
I'm going to walk back my comment [above](https://github.com/netbox-community/netbox/pull/20747#discussion_r2578249056). I realized that all the registration functions within the `netbox.plugins.registration` module apply specifically to the `plugins` registry. Other registration functions (e.g. `register_model_view()`) are scattered elsewhere. So I think your initial instinct to have `register_filterset()` live in `utilities.filtersets` was correct.

jeremystretch (approved):

docs/plugins/development/filtersets.md
```diff
@@ -6,12 +6,17 @@ Filter sets define the mechanisms available for filtering or searching through a
 
 To support additional functionality standard to NetBox models, such as tag assignment and custom field support, the `NetBoxModelFilterSet` class is available for use by plugins. This should be used as the base filter set class for plugin models which inherit from `NetBoxModel`. Within this class, individual filters can be declared as directed by the `django-filters` documentation. An example is provided below.
 
+!!! info "New in NetBox v4.5: FilterSet Registration"
+    NetBox v4.5 introduced the `register_filterset()` utility function. This enables plugins to register their filtersets to receive advanced functionality, such as the automatic attachment of field-specific lookup modifiers on the filter form. Registration is optional: Unregistered filtersets will continue to work as before, but will not receive the enhanced functionality.
+
```
jnovinger, line 11: Doh! Thanks for this.

PR #20749, opened 2025-11-05
Closes #18658: Add start on boot field to VirtualMachine model

### LINKED ISSUE #18658: virtual machine - start at boot option
### NetBox version

v4.2.3

### Feature type

Data model extension

### Proposed functionality

add a choice field which is to decide if a vm is meant to start at boot.
choices:
  - On
  - Off
  - Last State

### Use case

most hypervisors have the concept of if a vm is meant to start automatically when a host is powered up that the vm is located on.
usually the options are On or Off (proxmox) but also some other systems (hyperv) have the concept of last state.

### Database changes

_No response_

### External dependencies

_No response_

RobertH1993:
### Fixes: #18658

Added start_on_boot field to model VirtualMachine


RobertH1993:
Running ruff localy doesnt give any errors, so i cant fix it with --fix. Any ideas?

PR #20751, opened 2025-11-05
Closes  #18658: Add start on boot field to VirtualMachine model

### LINKED ISSUE #18658: virtual machine - start at boot option

RobertH1993:
### Closes: #18658

Adds start on boot field to VirtualMachine model


RobertH1993:
For reviewers

I had doubts about how I would do the displaying of the start on boot status. The 'status' field in the model VirtualMachine uses a badge. I decided to not use a badge but plaintext or a placeholder when the 'start_on_boot' field is empty. This was a personal choice but can be changed ofc. I did specify colors in the choices model for when a badge is preferred.

I took inspiration from another merge request that added a status field to some model, so I guess I got everything covered, bulk_imports, csv_imports, etc. I added this to the tests too and they run succesfully.

netbox/virtualization/migrations/0051_alter_virtualmachine_start_on_boot.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 5.2.7 on 2025-11-05 13:45
```
arthanson: These migrations should be collapsed together

docs/models/virtualization/virtualmachine.md
```diff
@@ -21,6 +21,13 @@ The VM's operational status.
 !!! tip
     Additional statuses may be defined by setting `VirtualMachine.status` under the [`FIELD_CHOICES`](../../configuration/data-validation.md#field_choices) configuration parameter.
 
+### Start on boot
+
+The start on boot setting from the hypervisor.
+
+!!! tip
+    Additional status may be defined by setting `VirtualMachine.start_on_boot` under the [`FIELD_CHOICES`](../../configuration/data-validation.md#field_choices) configuration parameter.
```
arthanson: Additional statuses

netbox/virtualization/choices.py
```diff
@@ -49,3 +49,17 @@ class VirtualMachineStatusChoices(ChoiceSet):
...
+
+class VirtualMachineStartOnBootChoices(ChoiceSet):
+    key = 'VirtualMachine.start_on_boot'
+
+    STATUS_ON = 'on'
+    STATUS_OFF = 'off'
+    STATUS_LAST_STATE = 'laststate'
+
+    CHOICES = [
+        (STATUS_ON, _('On'), 'green'),
```
arthanson, line 62: You are defining color here but it isn't used, it is also inconsistent - in the list view it is always shown as a grey badge no matter the status, in the detail view it is shown as text.  If there is a color/badge then it should be shown as such in both places.

netbox/virtualization/models/virtualmachines.py
```diff
@@ -79,6 +79,14 @@ class VirtualMachine(ContactsMixin, ImageAttachmentsMixin, RenderConfigMixin, Co
         default=VirtualMachineStatusChoices.STATUS_ACTIVE,
         verbose_name=_('status')
     )
+    start_on_boot = models.CharField(
```
arthanson, line 82: This is assigning a default and then allowing blank=True and null=True you probably don't want the blank and null=True there, check the status field for an example.  Is there a reason to have an unset state?

arthanson (changes requested):
Good work, just some cleanup needed mostly on the color and merge the migrations together.

netbox/virtualization/migrations/0051_alter_virtualmachine_start_on_boot.py
RobertH1993: I removed this migration, like you said in another feedback item there isn't really a case where the field should be empty. There is always some kind of behaviour when the hypervisor reboots

netbox/virtualization/choices.py
RobertH1993: Fixed this by adding a get_start_on_boot_color method to VirtualMachine model.

arthanson (approved):
Thanks @RobertH1993 !

PR #20761, opened 2025-11-06
Fixes #20750: Normalize actions in cloned objects init

### LINKED ISSUE #20750: Permission Cloning fails with single action selected
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.5

### Python Version

3.12

### Steps to Reproduce

The following assumes you are logged in with an administrative account of type "superuser".

1. Navigate to "Permissions"
2. Create a new basic permission (give it a name and object type[s])
3. Assign any single action and "Save"
4. Once redirected to the permission view, attempt to "Clone". This will cause the error below.

This has been reproduced in the Netbox Demo environment and a clone of our local production environment with all plugins disabled.

### Expected Behavior

   Attempting to clone a permission object which contains a single action should succeed. This has been tested to be reproducible with any single action selected, but appears to work when more than one is toggled.

Workaround: Currently, we have worked around this by removing all assignments on the permission, adding the "add" action, cloning, then removing the "add" action and re-assigning groups/users. This causes excess tedium. 

   Use case: Permissions in our environment need to limit what is visible on a least-privilege basis. For example, Asset Management only needs to see the hardware and serial numbers, while Facilities may need to see power infrastructure. This is required for local compliance regulations.Within departments, only certain people should be allowed to modify the documentation (henceforth 'department editors', i.e., the head of Facilities can update power panel information, but the workers can only see information). This is intended to enforce changes being documented as part of our workflow procedures before work is physically completed. NetBox represents our 'desired state' in this scenario.
   Currently, all departments are limited to specific "view only" permissions, but the trusted department editors have full edit permissions to everything via superuser. To enforce better security and privilege management, we want to clone the current "view only" permissions and set the "add/change/delete" permissions, then assign to the required groups/users.

### Observed Behavior

```
<class 'AttributeError'>

'str' object has no attribute 'remove'

Python version: 3.12.12
NetBox version: 4.4.5
Plugins: 
  netbox_topology_views: 4.4.0
```

pheus:
### Fixes: #20750

Normalize `actions` to a list of strings when initializing cloned `ObjectPermission` forms.
Prevents `AttributeError` caused by single action values being passed as a string instead of a list.

### DIFF
--- a/netbox/users/forms/model_forms.py
+++ b/netbox/users/forms/model_forms.py
@@ -372,6 +372,9 @@ class ObjectPermissionForm(forms.ModelForm):
         elif self.initial:
             # Handle cloned objects - actions come from initial data (URL parameters)
             if 'actions' in self.initial:
+                # Normalize actions to a list of strings
+                if isinstance(self.initial['actions'], str):
+                    self.initial['actions'] = [self.initial['actions']]
                 if cloned_actions := self.initial['actions']:
                     for action in ['view', 'add', 'change', 'delete']:
                         if action in cloned_actions:


jeremystretch (approved):
Thanks @pheus!

PR #20765, opened 2025-11-07
20465 fix script re-upload

### LINKED ISSUE #20465: Custom Script “Edit → Upload” creates a new file but NetBox continues to execute the old version
### NetBox Edition

NetBox Community

### NetBox Version

v4.3.6

### Python Version

3.12

### Steps to Reproduce

1. Upload Custom Script "log_test.py" to log "Foo"
2. Click run, log's "Foo"
3. Click Edit on script and upload modified file to log "Bar"
4. Click run, still logs "Foo"

### Expected Behavior

Should have logged "Bar" on second run.

### Observed Behavior

If you click into the script and examine the source you will see the code is unaltered.
If you look into /opt/netbox/netbox/scripts on the server a new file is created next to "log_test.py" called "log_test_tRlNd5h.py" though the UI still points towards old one.
Found an open discussion on this [issue #19716](https://github.com/netbox-community/netbox/discussions/19716)

arthanson:
### Fixes: #20465

On re-upload deletes the old file so storage_save won't create a new filename.

### DIFF
--- a/netbox/extras/forms/scripts.py
+++ b/netbox/extras/forms/scripts.py
@@ -79,8 +79,8 @@ class ScriptFileForm(ManagedFileForm):
                     pass
 
             # Save the new file and capture the actual filename
-            actual_filename = storage.save(filename, data)
-            self.instance.file_path = actual_filename
+            self.instance.file_path = filename
+            storage.save(filename, data)
 
         # need to skip ManagedFileForm save method
         return super(ManagedFileForm, self).save(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
Rather than handling this manually, we should be able to leverage the [`allow_overwrite`](https://docs.djangoproject.com/en/5.2/ref/files/storage/#django.core.files.storage.FileSystemStorage.allow_overwrite) argument when invoking the ScriptFileSystemStorage backend. AFAICT this can be accomplished by adding an option under the `DEFAULT_STORAGES` setting:

```python
DEFAULT_STORAGES = {
    "scripts": {
        "BACKEND": "extras.storage.ScriptFileSystemStorage",
        "OPTIONS": {
            "allow_overwrite": True,
        },
    },
}
```

PR #20777, opened 2025-11-10
Fixes #20638: Document bulk create support in OpenAPI schema

jnovinger:
### Fixes: #20638

POST operations on `NetBoxModelViewSet` endpoints accept both single objects and arrays, but the schema only documented single objects. This prevented REST API client generators from producing correct code.

Add explicit bulk_create_enabled flag to `NetBoxModelViewSet` and update schema generation to emit `oneOf` for these endpoints.

FWIW, I don't believe the `PUT` issue reported is a bug. The schema seems to correctly document `PUT` as array-only. NetBox's PUT is a bulk operation by design. For single updates, use `PATCH /api/model/{id}/`.

### DIFF
--- a/netbox/core/api/schema.py
+++ b/netbox/core/api/schema.py
@@ -12,6 +12,7 @@ from drf_spectacular.utils import Direction
 
 from netbox.api.fields import ChoiceField
 from netbox.api.serializers import WritableNestedSerializer
+from netbox.api.viewsets import NetBoxModelViewSet
 
 # see netbox.api.routers.NetBoxRouter
 BULK_ACTIONS = ("bulk_destroy", "bulk_partial_update", "bulk_update")
@@ -50,15 +51,8 @@ class ChoiceFieldFix(OpenApiSerializerFieldExtension):
 
 
 def viewset_handles_bulk_create(view):
-    """
-    Check if viewset explicitly declares bulk create support.
-
-    Viewsets opt-in to bulk create by setting bulk_create_enabled = True.
-    This allows POST operations to accept either single objects or arrays.
-
-    Refs: #20638
-    """
-    return getattr(view, 'bulk_create_enabled', False)
+    """Check if view automatically provides list-based bulk create"""
+    return isinstance(view, NetBoxModelViewSet)
 
 
 class NetBoxAutoSchema(AutoSchema):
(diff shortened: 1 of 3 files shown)


netbox/netbox/api/viewsets/__init__.py
```diff
@@ -128,6 +128,10 @@ def get_queryset(self):
         qs = super().get_queryset()
         return reapply_model_ordering(qs)
 
+    # Declare that this viewset accepts either single objects or arrays for creation
+    # Refs: #20638
+    bulk_create_enabled = True
```
jeremystretch: IMO we shouldn't rely on an explicit declaration of functionality as this can be unreliable. Rather, we should infer bulk creation ability from the existence of some inherent indicator. For instance, we could detect bulk update support by the existence of a `bulk_update()` method.

We might be able to achieve this by declaring a separate `bulk_create()` method and extending `NetBoxRouter` to route to it for `POST` operations on the list view. I suspect this would simplify some of the logic under NetBoxAutoSchema as well.

netbox/core/api/schema.py
```diff
@@ -128,6 +159,35 @@ def get_response_serializers(self) -> typing.Any:
 
         return response_serializers
 
+    def _get_request_for_media_type(self, serializer, direction):
```
jeremystretch: To preserve the default
```suggestion
    def _get_request_for_media_type(self, serializer, direction='request'):
```

netbox/netbox/api/viewsets/__init__.py
jnovinger: I gave the `bulk_create()` method and re-routing `POST` requests to it in `NetBoxRouter`, but ran in to several issues that just seemed to be complicating the issue. For instance, the schema reported that each of these endpoints now returned HTTP status 200 on create rather than the correct 201.

Given that the `bulk_create()` method was really doing anything besides passing to `create()`, the method itself wasn't really anything more than a flag itself. I finally just decided that, since the up-converting of single item `POST` action to list of items `POST` action happens in `NetBoxModelViewSet.get_serializer()`, that the simplest detection of functionality is to just verify the view in question is based on `NetBoxModelViewSet`.

jeremystretch: That seems like a reasonable approach. Can we remove `bulk_create_enabled` then?

netbox/netbox/api/viewsets/__init__.py
```diff
@@ -128,6 +128,10 @@ def get_queryset(self):
...
+
```
jeremystretch:
```suggestion
```

jeremystretch (changes requested):
Assuming we can remove the now-unused `bulk_create_enabled` var I think this is good to go.

jeremystretch (approved):

PR #20783, opened 2025-11-10
Fixes #20660: Optimize loading of custom script modules from remote storage

### LINKED ISSUE #20660: CustomStoragesLoader creates storage on every exec_module
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.4

### Python Version

3.10

### Steps to Reproduce

In the `CustomStoragesLoader` class in exec_module it inits an instance of `storage` for every call to the function (for every script reference - even on a Script list view).  In the case of storage backend like S3, this can add significant and linearly increasing overhead for every script file on the page.

Note: Follow on to Issue #20574 

### Expected Behavior

Constant (low) page load time on `/extras/scripts`

### Observed Behavior

Significant (and linearly increasing with number of script files) page load times, i.e. ~7s per script file.

jeremystretch:
### Fixes: #20660

- Tweak `exec_module()` to avoid forcing the creation of a new storage connection on each call
- Convert `module_scripts` to a cached property, to avoid triggering a new fetch for each script within a module file

Some quick testing using three modules comprising five total scripts shows total rendering time for the script list view reduced from ~4.15s to ~0.80s. There is still some linear increment per script module, as each file must be fetched from remote storage, but it should be greatly reduced.

### DIFF
--- a/netbox/extras/models/mixins.py
+++ b/netbox/extras/models/mixins.py
@@ -30,8 +30,7 @@ class CustomStoragesLoader(importlib.abc.Loader):
         return None  # Use default module creation
 
     def exec_module(self, module):
-        storage = storages.create_storage(storages.backends["scripts"])
-        with storage.open(self.filename, 'rb') as f:
+        with storages["scripts"].open(self.filename, 'rb') as f:
             code = f.read()
         exec(code, module.__dict__)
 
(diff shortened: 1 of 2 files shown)


jnovinger (approved):
I saw very similar improvements in loading time.

Issue #20564, opened 2025-10-13
Many-to-many mappings for pass-through ports


jeremystretch:
### NetBox version

v4.4.2

### Feature type

New functionality

### Proposed functionality

NetBox currently permits assigning multiple front ports to a single rear port. This FR proposes extending the data model to support the reverse as well (multiple rear ports to a single front port), enabling many-to-many assignment among front and rear ports. The current implementation, which employs a ForeignKey from FrontPort to RearPort and an Integerfield would be replaced by a bi-directional many-to-many assignment.

Once implemented, it will become possible to condense of expand the "stack" of positions in both directions across a pair of pass-through ports.

### Use case

Extending the data model to support multi-port mappings in either direction would enable modeling complex pass-through paths. These are commonly seen e.g. in datacenter Clos fabrics which utilize crossover fiber modules to simplify spine-to-leaf connections. An example is show below:

<img width="550" height="266" alt="Image" src="https://github.com/user-attachments/assets/5a585a32-a627-4c9c-8261-124c5552a50c" />

The resulting M2M table for the above topology would look like this:

| FrontPort | Position | RearPort | Position |
|---|---|---|---|
| 1 | 1 | 1 | 1 |
| 1 | 2 | 1 | 2 |
| 1 | 3 | 2 | 1 |
| 1 | 4 | 2 | 2 |
| 2 | 1 | 1 | 3 |
| 2 | 2 | 1 | 4 |
| 2 | 3 | 2 | 3 |
| 2 | 4 | 2 | 4 |

### Database changes

There are two relevant fields currently defined on the FrontPort model:

```python
class FrontPort:
    rear_port = models.ForeignKey(
        to='dcim.RearPort',
        on_delete=models.CASCADE,
        related_name='frontports'
    )
    rear_port_position = models.PositiveSmallIntegerField(
        verbose_name=_('rear port position'),
        default=1,
        validators=[
            MinValueValidator(REARPORT_POSITIONS_MIN),
            MaxValueValidator(REARPORT_POSITIONS_MAX)
        ],
        help_text=_('Mapped position on corresponding rear port')
    )
```

These would be replaced by a ManyToManyField on the FrontPort model. This M2M assignment would utilize a custom "through" table which annotates the numeric position of each port:

```python
class PortMapping:
    front_port = models.ForeignKey(
        to='dcim.FrontPort',
        on_delete=models.CASCADE,
    )
    front_port_position = models.PositiveSmallIntegerField(
        default=1,
    )
[...]

PR #20802, opened 2025-11-13
Closes #20788: Cable profiles and and position mapping

### LINKED ISSUE #20788: Cable profiles and connector/position mapping
### NetBox version

v4.4.6

### Feature type

Data model extension

### Proposed functionality

This FR supersedes #20562 and expands upon the core idea it proposes. In conjunction with FR #20564, its goal is to support modeling complex cabling in NetBox. There are two components in play here: cable profiles and connector/position tracking for cable terminations.

#### Cable termination connector & position

A cable termination represents the association of one end of a cable (A or B) with a particular object (e.g. an interface or console port). This FR would add two new pieces of data to the CableTermination model, to better represent real-world connections:

* Connector grouping: A numeric ID grouping cable terminations delivered via a common connector. For example, a 12-strand MPO fiber connector typically delivers up to six discrete connections (one fiber pair per connection).
* Connector position: The position of a termination within a connector relative to its peers. Continuing the MPO example above, available positions would be 1 through 6. This enables us to correlate a particular "lane" on the cable between two ends.

#### Cable profile

A cable profile holds the connector/position mapping between A- and B-side terminations. For example, a profile representing a 4x10GE breakout cable would look like this:

| A connector | A position | B connector | B position |
|-------------|------------|-------------|------------|
| 1           | 1          | 1           | 1          |
| 1           | 2          | 2           | 1          |
| 1           | 3          | 3           | 1          |
| 1           | 4          | 4           | 1          |

(Note that this concept of mapping positions is intentionally very similar to the approach proposed in FR #20564.)

A profile would be assigned to a cable via a ChoiceField on the Cable model. The current intent is for profiles to be declared statically and live in-memory, rather than being customizable: It is expected that a limited number of such real-world profiles exist. However, that may be reconsidered if additional flexibility is needed.

### Use case

While NetBox is capable of modeling complex cabling (cables which provide more than two endpoints) to a limited degree, it does not possess sufficient context to accurately trace these connections end-to-end. By associating explicit connector & positioning details with each termination, as well as a mapping profile with the cable itself, NetBox will be able to trace end-to-end paths across complex cables with precision.

### Database changes

On the `dcim.Cable` model, introduce a new optional CharField to associate the cable profile. Cables without a profile assigned will not support complex tracing.

```python
class Cable(...):
    profile = models.CharField(
        choices=CableProfileChoices,
        blank=True,
        null=True,
    )
)
```

One the `dcim.CableTermination` model, add two integer fields to track connector and position:

```python
class CableTermination(...):
    connector = models.PositiveSmallIntegerField()
    position = models.PositiveSmallIntegerField()
```

(We'll either make these fields nullable or default to `1`; whichever is found to work best once the implementation is fleshed out.)

### External dependencies

N/A

jeremystretch:
### Closes: #20788

- Introduce cable profiles to define mapping & validation logic for various classes of cables (e.g. straight-through vs. breakout)
- Add an optional `profile` choice field on dcim.Cable
- Add a nullable `cable_position` integer field on all cabled objects (e.g. interfaces)
- Rename `create_cablepath()` to `create_cablepaths()` and add position support
- Extend cable tracing logic to consider the position of a path relative to its cable
- Enable drag-and-drop of items within multi-select widgets
- Refactor cable path tests
  - Split out base CablePathTestCase class
  - Introduced new profile-based tests in parallel to legacy (non-profile) tests
  - Did not alter any of the legacy tests, to ensure backward compatibility is retained
- Update documentation for the Cable model


netbox/dcim/models/device_components.py
```diff
@@ -194,14 +203,23 @@ class Meta:
...
+                    "cable_position": _("Must specify cable termination position when attaching a cable.")
+                })
         if self.cable_end and not self.cable:
             raise ValidationError({
                 "cable_end": _("Cable end must not be set without a cable.")
             })
+        if self.cable_position and not self.cable:
+            raise ValidationError({
+                "cable_position": _("Cable termination position must not be set without a cable.")
+            })
```
jnovinger, line 222: Since `cable_position` is never settable by a user, what's the purpose of raising a `ValidationError` here (and above)? Just being proactive?

netbox/dcim/models/cables.py
```diff
@@ -347,6 +370,14 @@ class CableTermination(ChangeLoggedModel):
...
         fk_field='termination_id'
     )
+    position = models.PositiveIntegerField(
+        blank=True,
+        null=True,
+        validators=(
+            MinValueValidator(CABLE_POSITION_MIN),
+            MaxValueValidator(CABLE_POSITION_MAX)
+        )
+    )
```
jnovinger, line 380: Just to be clear, the `connector` field was turned out to not be needed, correct?

netbox/dcim/models/cables.py
```diff
@@ -315,12 +336,14 @@ def update_terminations(self):
...
+        for i, termination in enumerate(self.a_terminations, start=1):
             if not termination.pk or termination not in a_terminations:
-                CableTermination(cable=self, cable_end='A', termination=termination).save()
-        for termination in self.b_terminations:
+                position = i if self.profile and isinstance(termination, PathEndpoint) else None
+                CableTermination(cable=self, cable_end='A', position=position, termination=termination).save()
+        for i, termination in enumerate(self.b_terminations, start=1):
             if not termination.pk or termination not in b_terminations:
-                CableTermination(cable=self, cable_end='B', termination=termination).save()
+                position = i if self.profile and isinstance(termination, PathEndpoint) else None
```
jnovinger, line 345: Are we only doing this on `PathEndpoint`-inheriting classes because the pass-through cabled objects don't need/shouldn't see `position`?

netbox/project-static/src/select/config.ts
```diff
@@ -20,6 +20,11 @@ export function getPlugins(element: HTMLSelectElement): object {
     };
   }
 
+  // Enable drag-and-drop reordering of items on multi-select fields
+  if (element.hasAttribute('multiple')) {
+    plugins.drag_drop = {};
+  }
```
jnovinger, line 26: I don't remember seeing this in the FR. Is this included intentionally?

netbox/dcim/tests/test_api.py
```diff
@@ -2427,7 +2427,7 @@ class CableTerminationTest(
     APIViewTestCases.ListObjectsViewTestCase,
```
jnovinger, line 2430: Apologies for the location of this comment, could not comment on line 2422 above. Can we add an entry to `CableTest.create_data`, just to encode that creation via the API is tested and supported? I tried the following successfully locally:

```python
{
    'a_terminations': [{
        'object_type': 'dcim.interface',
        'object_id': interfaces[7].pk,
    }],
    'b_terminations': [{
        'object_type': 'dcim.interface',
        'object_id': interfaces[17].pk,
    }],
    'label': 'Cable 7',
    'profile': CableProfileChoices.STRAIGHT_SINGLE,
},
```

jnovinger (approved):
All in all, this looks excellent and well-thought through. I wasn't able to find any sharp edges while poking at it.

I had some questions, but they're more clarifications than anything. Approved as is, but feel free to re-request if you want to make some additions.

netbox/dcim/models/device_components.py
jeremystretch: Yeah, it's just a sanity check.

netbox/dcim/models/cables.py
jeremystretch: Correct

netbox/project-static/src/select/config.ts
jeremystretch: It was included in the original FR (#20562) to support drag-and-drop for cable terminations:

> This may require enabling Tom-Select's [drag-and-drop](https://tom-select.js.org/plugins/drag-drop/) functionality to allow reordering interfaces within the multi-select list.

netbox/dcim/models/cables.py
jeremystretch: As far as _this_ PR is concerned, yes. This will change as we complete work on FR #20564.

netbox/project-static/src/select/config.ts
jnovinger: :facepalm:

jnovinger (approved):

PR #20830, opened 2025-11-19
fix: check permission constraints when running script by API

### LINKED ISSUE #20740: Permissions constraints are ignored when running script by API
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.1

### Python Version

3.12

### Steps to Reproduce

This is a follow-up to #15587.
When permission constraints are used to restrict users ability to run specific scripts, it works as intended in the GUI, but API requests allow running all scripts, ignoring the constraint.

1. Create two custom scripts, called "safe" and "dangerous"
2. Create three permissions:  
  - `readonly` can **view** all objects
  - `read safe scripts` can **view** `Core | managed file` with constraints `{"data_path": "scripts/safe_script.py"}`
  - `run safe scripts` can **view** and **run** `Extras > Script` with constraints `{"name": "SafeScript"}`
3. Create a user called Alice with those permissions
4. Create an API token for Alice
5. Run the dangerous script by API. `curl  -H "Content-Type: application/json" http://localhost:8000/api/extras/scripts/2/ --data '{"data": {}, "commit": true}' -H "Authorization: Token $ALICE_TOKEN"`

### Expected Behavior

When Alice makes an API request to run the "dangerous script" it should fail.

### Observed Behavior

Alice can run both scripts by API.
In the GUI, she can run the safe script, but when she tries to run the dangerous script, the run button is disabled.

llamafilm:
### Fixes: #20740

Now the REST API enforces the same permission constraints as the webpage, when running a custom script.

### DIFF
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -290,10 +290,12 @@ class ScriptViewSet(ModelViewSet):
         """
         Run a Script identified by its numeric PK or module & name and return the pending Job as the result
         """
-        if not request.user.has_perm('extras.run_script'):
-            raise PermissionDenied("This user does not have permission to run scripts.")
 
         script = self._get_script(pk)
+
+        if not request.user.has_perm('extras.run_script', obj=script):
+            raise PermissionDenied("This user does not have permission to run this script.")
+
         input_serializer = serializers.ScriptInputSerializer(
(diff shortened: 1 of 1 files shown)


arthanson (approved):
thanks @llamafilm

PR #20835, opened 2025-11-19
Fixes #20827: fix theme toggle visibility for logo and buttons

### LINKED ISSUE #20827: Light/Dark Theme toggle: Logo, button icon, and tooltip don't update until page refresh
## Bug Description

When clicking the theme toggle button to switch between light and dark mode, the theme colors change immediately but the logo, button icon, and tooltip text do not update until the page is refreshed.

https://github.com/user-attachments/assets/6b8b2bb8-1b29-4689-b96e-f2729b156165

 ## Steps to Reproduce

 1. Load any NetBox page in light mode
 2. Hover over the theme toggle button - tooltip says "Enable dark mode" ✓
 3. Click the theme toggle button to switch to dark mode
 4. **BUG**: The page switches to dark mode colors, but:
    - The NetBox logo doesn't switch to the bright teal version (stays dark teal)
    - The toggle button icon doesn't update
    - Hovering shows tooltip still says "Enable dark mode" (should say "Enable light mode")
 5. Refresh the page
 6. Now the logo, icon, and tooltip are correct

 The same issue occurs when switching from dark mode back to light mode.

 ## Expected Behavior

 When clicking the theme toggle:
 - Theme colors should change ✓
 - Logo should immediately switch between dark teal (light mode) and bright teal (dark mode) versions
 - Toggle button icon should update
 - Tooltip should immediately show the correct text for the current mode

 ## Suspected Root Cause

 Tabler's CSS has broken `.hide-theme-*` utility classes that check for `data-bs-theme` on `:root`, but NetBox sets this attribute on `<body>`.
 This causes elements to be shown/hidden incorrectly during theme switching.

 ```css
 /* Tabler's broken rule - checks :root but NetBox uses body */
 :root:not([data-bs-theme=dark]) .hide-theme-light {
   display: none !important;
 }
 ```

 Since `:root` never has the `data-bs-theme` attribute, the selector behavior is incorrect, preventing the logo and buttons from switching
 visibility properly.

 ## Solution

 Add CSS override rules in `netbox/project-static/styles/overrides/_tabler.scss` with higher specificity that correctly target
 `body[data-bs-theme]` where the attribute actually exists.

 ## Environment

 - NetBox Version: 4.6.0
 - Browser: Tested in Chrome
 - Theme: Both light and dark mode switches affected

abubnalitic-nbl:
Fixes #20827

Before fix:

https://github.com/user-attachments/assets/0e331506-dac5-4379-a3e0-ad24bbc770c3

After fix:

https://github.com/user-attachments/assets/9bd825fa-c093-4662-8d28-e9bedbede2c4


jeremystretch (approved):
Thanks @abubnalitic-nbl!

PR #20837, opened 2025-11-19
Closes 20817 fix datasource sync broken when cron is set

### LINKED ISSUE #20817: Data Sources remove sync interval
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.6

### Python Version

3.10

### Steps to Reproduce

1.  Create new Data Source (git)
2. Leave the Sync Interval blank
3. After creation the "Sync" button is active and functional.
4. Edit the sync internal and set it to anything.
5. Save
6. Sync button becomes inactive.
7. Edit the sync internal and remove the existing sync interval
8. Save
9. Status still shows as "Queued" and sync button is still grayed out.
10. The job was removed from the Jobs Queue.  
11. The data source never changes out of Queue state, nor does the Sync button ever become active.

### Expected Behavior

Removing Sync Internal should re-enable the Sync button and change the status to Active.

### Observed Behavior

Status still shows as Queued and Sync button is grayed out.

ifoughal:
### Fixes: #20817 

Datasources sync was broken when the user sets a sync interval.

The fix introduces 2 new states:

- Ready 
- Scheduled

Ready state is set when the data-source object is not new. ie: either updated or changed state recently.
Scheduled is set when the user sets a sync interval.

This PR includes the following logic enhancements:

- by removing ```DataSourceStatusChoices.QUEUED``` from the ```ready_for_sync``` method, we let the user sync the data-source anytime even when a daily interval is set, as long as the data-source is enabled.

- Include the 2 new states (Scheduled, Ready) in the clean & save properties of data-source form model, to set the status correctly depending on the user inputs/
  - If the user creates a new data-source object, then the state is new.
  - if the user update the object with a defined interval, then the states changes to scheduled
  - if the user removed the sync interval, then the state changes to Ready.
  - if a sync is completed either successfully or with a failure, then the status will get update with the job's status as before without impact.


jnovinger:
@ifoughal, CI appears to be failing with a test failure.

ifoughal:
@jnovinger 

Initially I thought there was an issue with the view rendering class, but after some further TS, and checking the actual value of status on the DB entry, it seems the status field got overwritten somehow. 
```
 id |            created            |         last_updated         | custom_field_data | description | comments | name | type |                           source_url                            | status | enabled | ignore_rules |                   parameters                   | last_synced | sync_interval 
----+-------------------------------+------------------------------+-------------------+-------------+----------+------+------+-----------------------------------------------------------------+--------+---------+--------------+------------------------------------------------+-------------+---------------
 10 | 2025-11-20 10:02:12.763951+01 | 2025-11-20 10:02:19.22684+01 | {}                |             |          | test | git  | https://gitlab.pic.services.prod/orion/IaC/config-templates.git | queued | t       |              | {"branch": "", "password": "", "username": ""} |             |           720
 ```
 
To circumvent this, I run the instance.save() method in the save override of dataSourcesForm. but this generates 2 commits and breaks the test. I'll continue looking into a fix.

**Comment by ifoughal** [comment]
@jnovinger 

The issue was coming from :

```
@classmethod
    def enqueue(cls, *args, **kwargs):
        job = super().enqueue(*args, **kwargs)

        # Update the DataSource's synchronization status to queued
        if datasource := job.object:
            datasource.status = DataSourceStatusChoices.QUEUED
            DataSource.objects.filter(pk=datasource.pk).update(status=datasource.status)

        return job
```

The enqueue method was overriding the state, I have removed ```datasource.status = DataSourceStatusChoices.QUEUED```, the state should then be updated on the clean() method of the post of the datasource form.

File: netbox/core/jobs.py [hunk 1]
```diff
@@ -31,7 +31,6 @@ def enqueue(cls, *args, **kwargs):
 
         # Update the DataSource's synchronization status to queued
         if datasource := job.object:
```
arthanson, line 33: You've removed the set to datasource.status but it is still being used in the update clause which means it will get set to whatever job.object currently has. I think we want to set it to a set value here?

netbox/core/models/data.py
```diff
@@ -112,7 +112,6 @@ def backend_class(self):
     @property
     def ready_for_sync(self):
         return self.enabled and self.status not in (
```
arthanson: if a single value can just be an == statement and doesn't need an in.

arthanson (changes requested):
@ifoughal Is QUEUED status used anymore? I don't see it ever getting set anywhere. 

The QUEUED state was that it was submitted to be run (so pending being run). I think this is should still be a valid status, i.e. get set when it is pending being run (enqueued) and cleared once it is run.

The description for the PR says that two new states are added, but only one is added here, is the description old?

ifoughal:
> @ifoughal Is QUEUED status used anymore? I don't see it ever getting set anywhere.
> 
> The QUEUED state was that it was submitted to be run (so pending being run). I think this is should still be a valid status, i.e. get set when it is pending being run (enqueued) and cleared once it is run.
> 
> The description for the PR says that two new states are added, but only one is added here, is the description old?

I will be reverting the removed of queued as it was indeed needed for when the rq workers as busy. It should then reflect the queued state.

The two new states are ```READY``` & ```SCHEDULED```.
Ready is used when the datasource object is not ```NEW```, meaning if the user removes the schedule, so when it changes from interval to None, then the state becomes Ready (ready for sync).
as for ```scheduled```, as its name suggests, is used when the user sets a schedule for the datasource object.

netbox/core/choices.py
```diff
@@ -13,13 +13,15 @@ class DataSourceStatusChoices(ChoiceSet):
...
     FAILED = 'failed'
+    READY = 'ready'
 
     CHOICES = (
         (NEW, _('New'), 'blue'),
         (QUEUED, _('Queued'), 'orange'),
         (SYNCING, _('Syncing'), 'cyan'),
         (COMPLETED, _('Completed'), 'green'),
         (FAILED, _('Failed'), 'red'),
+        (READY, _('Ready'), 'green'),
```
arthanson, line 24: You say SCHEDULED status is added but there is no scheduled in the status or code anywhere?  Still confused as you state "The two new states are READY & SCHEDULED."

netbox/core/forms/model_forms.py
```diff
@@ -79,14 +80,24 @@ def __init__(self, *args, **kwargs):
                 if self.instance and self.instance.parameters:
                     self.fields[field_name].initial = self.instance.parameters.get(name)
 
-    def save(self, *args, **kwargs):
+    def clean(self):
+        super().clean()
+        if not self.instance.pk:
+            self.cleaned_data['status'] = DataSourceStatusChoices.NEW
+        else:
+            if not self.data.get('sync_interval'):
```
arthanson: shouldn't this be cleaned_data?

netbox/core/forms/model_forms.py
```diff
@@ -79,14 +80,24 @@ def __init__(self, *args, **kwargs):
                 if self.instance and self.instance.parameters:
                     self.fields[field_name].initial = self.instance.parameters.get(name)
 
-    def save(self, *args, **kwargs):
+    def clean(self):
+        super().clean()
+        if not self.instance.pk:
+            self.cleaned_data['status'] = DataSourceStatusChoices.NEW
```
arthanson: It's sort of strange to be setting this in cleaned_data as that field doesn't exist.  I think you are only using this in save so why not just set it there?

netbox/core/models/data.py
```diff
@@ -111,10 +111,7 @@ def backend_class(self):
 
     @property
     def ready_for_sync(self):
-        return self.enabled and self.status not in (
-            DataSourceStatusChoices.QUEUED,
-            DataSourceStatusChoices.SYNCING
-        )
+        return self.enabled and self.status != DataSourceStatusChoices.SYNCING
```
arthanson, line 114: Question: If we are adding the READY state and other changes here do we still need to skip QUEUED check here?  Just wondering if your other changes will allow it to be kept as I think if it is actually queued and hasn't finished we probably don't want to allow sync again?

arthanson (changes requested):

netbox/core/choices.py
ifoughal: I ended up removing it, so there is only one state that has been added in the end, which is the ```READY``` state. 

The reasoning behind was that ```scheduled``` would interfere with the post sync action trigger status (completed/failed/syncing). Therefore, I decided to keep the ```ready``` state only, which has the same effect as ```new```, the only difference is that ```new``` is only applied when the object gets created, whereas ```ready``` is applied when there there is no effective sync interval.

netbox/core/forms/model_forms.py
ifoughal: correct, it is safer!

netbox/core/models/data.py
ifoughal: that's why I had initially added the ```scheduled``` state initially, so that when it's queued, I would disable the syncing button, but allow it only when the state is scheduled. The issue was that scheduled was completely redundant with Queued. As they could be set only during the save action of the data-source object. They would also get overwritten when the job finishes. so queued, just as before reflects when the job is queued when the user hits sync, but also when the sync_interval is set. 

- The proper way to fix it would be to add a new field, for example ```last-status``` and use that for the latest job status.
- keep the current status field, and use it only for the workers Q state (new, ready, queued) and maybe re-add the ```scheduled``` state as well.

what do you think?

netbox/core/forms/model_forms.py
ifoughal: moved logic to save method.

jeremystretch:
@ifoughal thanks for taking a stab at this, but I think we just need to clear out any pending jobs and reset the data source's status. We don't need to introduce a new status for this. I've just submitted PR #21055 with a slightly simplified approach as I want to make sure we ship a fix for this in time for the upcoming v4.4.10 release.

Thanks again for your work on this!

PR #20838, opened 2025-11-19
Closes #20820: Add Object Type Filter to CustomField

### LINKED ISSUE #20820: Custom fields - add ability to filter and sort by object type
### NetBox version

v4.4.6

### Feature type

New functionality

### Proposed functionality

Add ability to filter Custom fields list (`/extras/custom-fields/`) by their assigned Object Types.

Also make Object types column sortable. Though I suspect there is a good reason that sorting is disabled. Just filtering would go a long way to make the list more manageable.

### Use case

You could for example filter all custom fields that are assigned to Interface.

Once you get to a few dozen and more custom fields in a netbox installation it becomes difficult to quickly see what fields belong where.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Fixes: #20820

Add `object_type_id` to the CustomField filter form to filter by assigned object types. Move type‑specific controls into a new “Type Options” FieldSet (`choice_set_id`, `related_object_type_id`). This clarifies the difference between a field’s assigned `ObjectType` and its related `ObjectType` (for `object` fields), improving discoverability and consistency.

UI only; no API or database changes.

### DIFF
--- a/netbox/extras/forms/filtersets.py
+++ b/netbox/extras/forms/filtersets.py
@@ -42,17 +42,20 @@ class CustomFieldFilterForm(SavedFiltersMixin, FilterForm):
     model = CustomField
     fieldsets = (
         FieldSet('q', 'filter_id'),
-        FieldSet(
-            'type', 'related_object_type_id', 'group_name', 'weight', 'required', 'unique', 'choice_set_id',
-            name=_('Attributes')
-        ),
+        FieldSet('object_type_id', 'type', 'group_name', 'weight', 'required', 'unique', name=_('Attributes')),
+        FieldSet('choice_set_id', 'related_object_type_id', name=_('Type Options')),
         FieldSet('ui_visible', 'ui_editable', 'is_cloneable', name=_('Behavior')),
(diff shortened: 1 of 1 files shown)


netbox/extras/forms/filtersets.py
```diff
@@ -42,17 +42,20 @@ class CustomFieldFilterForm(SavedFiltersMixin, FilterForm):
...
     )
-    related_object_type_id = ContentTypeMultipleChoiceField(
+    object_type_id = ContentTypeMultipleChoiceField(
         queryset=ObjectType.objects.with_feature('custom_fields'),
         required=False,
-        label=_('Related object type')
+        label=_('Object types'),
+    )
+    related_object_type_id = ContentTypeMultipleChoiceField(
+        queryset=ObjectType.objects.public(),
```
pheus, line 56: _Note_: I’ve updated the queryset for `related_object_type_id` to use `ObjectType.objects.public()`, matching the queryset used on the model. Previously it used `ObjectType.objects.with_feature('custom_fields')`, which excluded valid targets for object-type custom fields.

bctiemann (approved):

PR #20850, opened 2025-11-20
Closes #20841: Add selector widget to RackType field on the Rack EditForm

### LINKED ISSUE #20841: Filter `RackType` by Manufacturer on the Rack EditForm
### NetBox version

v4.4.6

### Feature type

### Proposed functionality

When creating or editing a Rack, allow filtering the **Rack type** (`RackType`) choices by **Manufacturer**, aligning with the device/module workflow.

Two possible approaches (either would work):

1) **Manufacturer field + query params**  
Add a `Manufacturer` ModelChoiceField to the Rack form and use its `query_params` to constrain the `Rack type` field to types from the selected manufacturer.

2) **Selector modal on Rack type**  
Add/enable the **Selector** modal button for the `Rack type` field. The modal should expose a `Manufacturer` filter so the user can pick a vendor first, then select a Rack type from the filtered list.

### Use case

Environments often deploy multiple rack models from different manufacturers. Filtering rack types by manufacturer shortens long mixed lists, speeds data entry, prevents vendor mix-ups, and keeps the UX consistent with devices and modules.

### Database changes

None.

### External dependencies

None.

pheus:
### Fixes: #20841

This PR configures the `RackType` field on the rack add/edit form to use the selector widget.

Using the selector widget improves usability by:

- Allowing users to search rack types instead of scrolling a long dropdown
- Making it easier to filter and choose the correct rack type (e.g. by manufacturer)
- Keeping the rack workflow consistent with other objects that already use selector modals

No model or API changes are introduced; this only updates the rack form UI.

### DIFF
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -269,7 +269,8 @@ class RackForm(TenancyForm, NetBoxModelForm):
         label=_('Rack Type'),
         queryset=RackType.objects.all(),
         required=False,
-        help_text=_("Select a pre-defined rack type, or set physical characteristics below.")
+        selector=True,
+        help_text=_("Select a pre-defined rack type, or set physical characteristics below."),
     )
     comments = CommentField()
 


jeremystretch (approved):
Thanks @pheus!

PR #20851, opened 2025-11-20
Closes #20564: Many-to-many pass-through port mappings

### LINKED ISSUE #20564: Many-to-many mappings for pass-through ports

jeremystretch:
### Closes: #20564

**Note:** Any changes made for FrontPort or RearPort are also made for their corresponding template models, FrontPortTemplate and RearPortTemplate, unless otherwise notes.

- Introduce the `PortMapping` model to map port & position pairs between FrontPorts and RearPorts
  - This model establishes a `mappings` relation on both FrontPort & RearPort
- Add a `positions` IntegerField on FrontPort to match the existing `positions` field on RearPort
- Remove the `rear_port` ForeignKey and `rear_port_position` IntegerField from FrontPort
- Update REST API serializers
  - Add `front_ports` on RearPortSerializer
  - Add `rear_ports` on FrontPortSerializer
- Extend device instantiation logic to replicate all front/rear port mappings from a DeviceType to a newly created Device
  - Introduce the `create_port_mappings()` function for this purpose
- Update cable tracing logic to respect positions mapped between front & rear ports
- Update FrontPort & RearPort tables
  - Remove `rear_port` and `rear_port_position` columns from FrontPortTable
  - Add `mappings` column to both FrontPortTable and RearPortTable
- Add a `front_port_id` filter for RearPort and update FrontPort's `rear_port_id` filter
- Add `positions` filter for FrontPort and FrontPortTemplate
- GraphQL API
  - Replace `rear_port` field on FrontPortType with `mappings`
  - Add `mappings` to RearPortType
- Update all tests to use PortMapping for front/rear port mapping
- Add debug logging for cable path tracing

### TODO

- [x] Add FKs direct to Device/DeviceType on port mappings
- [x] Finish adapting GraphQL API (e.g. add PortMappingTemplateFilter)
- [ ] Support mapping of front/rear ports via bulk import
- [x] Support port template mapping via device type import
- [x] Update logic for detecting split paths
- [x] Test setting front/rear mappings via REST API
- [x] Add tests to verify cable paths are updated correctly when front/rear port mappings change
- [x] Clean up debugging code under `Cable.from_origin()`


netbox/dcim/tests/test_cablepaths.py
jeremystretch, line 1: The only changes made in this file are to update the test data to use PortMappings. None of the test methods themselves have been significantly modified.

jeremystretch:
Aside from the incomplete items noted in the **TODO** list above, I believe this is ready for review.

arthanson:
This is the Rear Ports detail view, it is showing "Rear Ports" as the included table, should be "Front Ports"
Monosnap Port 1 | NetBox 2025-12-04 12-04-20

arthanson:
We May want to move Rear Ports out of the right-hand-side to below all the other columns - with this demo data it is pushing connection and inventory items way down so not very usable.
Monosnap Rear Splice | NetBox 2025-12-04 12-06-12

arthanson:
On cable trace in the Related Paths I think it would be useful to not link the current path, for example this one is showing the end path "xe-0/0/3" to "xe-0/0/3" but it is still linked in the table and not obvious which one you are on except by looking at the name in the trace box.  Having the current one not be linked would just another nice UI  indicator of which one.
cable-trace

arthanson (changes requested):
Looks good - the main item is fixing the confusing "rear ports" reference, should be "front ports" on the rear ports detail page. Couple other suggested UI tweaks.

jeremystretch:
> On cable trace in the Related Paths I think it would be useful to not link the current path

Agreed, but this improvement would be out of scope for #20564. Do you want to open a separate FR for it?

arthanson (approved):

PR #20862, opened 2025-11-21
Closes #20823: Validate Token expiration date on creation

### LINKED ISSUE #20823: Prevent Creation Of API Token With Expiration In Past
### NetBox version

v4.4.6

### Feature type

### Proposed functionality

When users try to create an API Token with an expiration date in the past, the system should throw an error and not create the token.

### Use case

Creating API Tokens with expiration dates in the past serves few to any benefits. Setting tokens with future values makes sense and has benefits. 

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Fixes: #20823

This PR adds model-level validation to prevent creating API tokens with an expiration date in the past, while still allowing updates to existing tokens (including already-expired ones).

Changes:

- Updates `Token.clean()` to raise a `ValidationError` when creating a new token (`pk is None`) whose `expires` timestamp is already in the past.
- Leaves updates to existing tokens unchanged so that expired tokens can still be modified as needed.
- Adds tests in `users/tests/test_models.py` covering:
  - the `Token.is_expired` property for `None`, future, and past expiration times
  - creation of tokens with a past expiration date
  - updates to existing expired tokens

No database or API schema changes are introduced by this PR.

### DIFF
--- a/netbox/users/models/tokens.py
+++ b/netbox/users/models/tokens.py
@@ -1,8 +1,10 @@
 import binascii
 import os
+import zoneinfo
 
 from django.conf import settings
 from django.contrib.postgres.fields import ArrayField
+from django.core.exceptions import ValidationError
 from django.core.validators import MinLengthValidator
 from django.db import models
 from django.urls import reverse
@@ -85,6 +87,24 @@ class Token(models.Model):
     @property
     def partial(self):
         return f'**********************************{self.key[-6:]}' if self.key else ''
+
+    def clean(self):
+        super().clean()
+
+        # Prevent creating a token with a past expiration date
+        # while allowing updates to existing tokens.
+        if self.pk is None and self.is_expired:
+            current_tz = zoneinfo.ZoneInfo(settings.TIME_ZONE)
+            now = timezone.now().astimezone(current_tz)
+            current_time_str = f'{now.date().isoformat()} {now.time().isoformat(timespec="seconds")}'
+
+            # Translators: {current_time} is the current server date and time in ISO format,
+            # {timezone} is the configured server time zone (for example, "UTC" or "Europe/Berlin").
+            message = _('Expiration time must be in the future. '
+                        'Current server time is {current_time} ({timezone}).'
+                        ).format(current_time=current_time_str, timezone=current_tz.key)
+
+            raise ValidationError({'expires': message})
 
     def save(self, *args, **kwargs):
         if not self.key:
--- a/netbox/users/tests/test_models.py
+++ b/netbox/users/tests/test_models.py
@@ -1,6 +1,72 @@
-from django.test import TestCase
+from datetime import timedelta
 
-from users.models import User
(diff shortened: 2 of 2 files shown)


netbox/users/models/tokens.py
```diff
@@ -86,6 +87,14 @@ def get_absolute_url(self):
     def partial(self):
         return f'**********************************{self.key[-6:]}' if self.key else ''
 
+    def clean(self):
+        super().clean()
+
+        # Prevent creating a token with a past expiration date
+        # while allowing updates to existing tokens.
+        if self.pk is None and self.is_expired:
+            raise ValidationError({'expires': _('Expiration time must be in the future.')})
```
jnovinger: I think we should consider giving the user some feedback here on what the current date, time, and timezone the server expects are. NetBox's `settings.py` defaults to `UTC` if an installation's config does not set a custom `TIME_ZONE` setting, which actually tripped me up when I was testing this locally as I didn't have it set and so was confused why this wasn't working for something just 15 mins ahead of the current time (I'm UTC-6:00 at the moment).

I'm not arguing that we should include it by default in the UI, but perhaps just when this validation error is raised. You can look at [`netbox.tables.columns.DateTimeColumn`](https://github.com/netbox-community/netbox/blob/066b78777798b60be7215f7355ac7ff00f20e466/netbox/netbox/tables/columns.py#L87-L91) to see how we've accessed the `TIME_ZONE` setting and displayed it in table columns, although you may need to format it a little differently here. I would also review Django's [docs on internationalization](https://docs.djangoproject.com/en/5.2/topics/i18n/) to find the best way to include a dynamic element like that in a translatable string.

pheus: Thanks for the detailed review and for pointing out the timezone behavior! That context about the default UTC setting and `DateTimeColumn` is really helpful.

I’ll look into surfacing the server’s current date/time and timezone when this validation error is raised and make sure it’s done in a way that plays nicely with Django’s i18n and translations.

pheus:
Thanks again for the detailed review and the pointers around timezones and i18n 🙏

I’ve updated the `ValidationError` to include both the current server time and the configured timezone, following the pattern from `DateTimeColumn` and using named placeholders so it’s translation‑friendly.

For local verification, I regenerated the English and German message catalogs and confirmed that the new string and the translators’ hint are picked up correctly. The entry looks like this in `django.po`:

```po
#. Translators: {current_time} is the current server date and time in ISO format,
#. {timezone} is the configured server time zone (for example, "UTC" or "Europe/Berlin").
#: users/models/tokens.py:78
#, python-brace-format
msgid ""
"Expiration time must be in the future. Current server time is {current_time} "
"({timezone})."
msgstr "Das Ablaufdatum muss in der Zukunft liegen. Aktuelle Serverzeit ist {current_time} ({timezone})."
```

jnovinger (approved):

PR #20864, opened 2025-11-21
Closes #20834: Add support for enabling/disabling Tokens

### LINKED ISSUE #20834: Enable field for API tokens
### NetBox version

v4.4.6

### Feature type

New functionality

### Proposed functionality

Have a button that users would be able to click which would enable and disable API Tokens.

At the time that the users create and edit the individual API Tokens, there would either be an Enable checkbox that checked means enabled this unchecked means disabled or a status drop-down that users can change.

### Use case

If an API Token needs to be disabled or re-enabled, the only way to do that would be to change the Expiration Date or restrict the IPs. It’s less than ideal to be changing those values to simply enable or disable the tokens.

### Database changes

A new field as either Status of either Enabled or Disabled or enabled: True for the API to do the comparison when authentication and use of an API Token is done. The idea being that if Enabled is set to True or Status is set to enabled, it would allow authentication. And this value would be able to be set by administrators at time of creation as well as part of normal work. Moreover, the value could be set to False or disabled when the Expiration Date passes.

### External dependencies

Unsure. 

Sorry if this is not useful, just thought it may be something that may help users with administering the tool

pheus:
### Fixes: #20834

This PR introduces an `enabled` field on the `Token` model to allow API tokens to be temporarily disabled without being deleted.

Key changes:

- Add an `enabled` Boolean field to `Token` (defaulting to `True` for backwards compatibility)
- Enforce the `enabled` flag in token authentication so disabled tokens are rejected
- Expose the new field in token forms, bulk edit, and tables to make it easy to toggle status from the UI
- Expose `enabled` via the token API serializers (including provisioning), allowing tokens to be created and managed in a disabled state
- Add tests for the Token model, API, and authentication to cover the new behavior

Existing tokens remain enabled by default, so there is no change in behavior until a token is explicitly disabled.

Thanks in advance for reviewing!

### DIFF
--- a/netbox/netbox/tests/test_authentication.py
+++ b/netbox/netbox/tests/test_authentication.py
@@ -67,6 +67,32 @@ class TokenAuthenticationTestCase(APITestCase):
         self.assertEqual(response.data['detail'], "Invalid v2 token")
 
     @override_settings(LOGIN_REQUIRED=True, EXEMPT_VIEW_PERMISSIONS=['*'])
+    def test_token_enabled(self):
+        url = reverse('dcim-api:site-list')
+
+        # Create v1 & v2 tokens
+        token1 = Token.objects.create(version=1, user=self.user, enabled=True)
+        token2 = Token.objects.create(version=2, user=self.user, enabled=True)
+
+        # Request with an enabled token should succeed
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Token {token1.token}')
+        self.assertEqual(response.status_code, 200)
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Bearer {TOKEN_PREFIX}{token2.key}.{token2.token}')
+        self.assertEqual(response.status_code, 200)
+
+        # Request with a disabled token should fail
+        token1.enabled = False
+        token1.save()
+        token2.enabled = False
+        token2.save()
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Token {token1.key}')
+        self.assertEqual(response.status_code, 403)
+        self.assertEqual(response.data['detail'], 'Token disabled')
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Bearer {TOKEN_PREFIX}{token2.key}')
+        self.assertEqual(response.status_code, 403)
+        self.assertEqual(response.data['detail'], 'Token disabled')
+
+    @override_settings(LOGIN_REQUIRED=True, EXEMPT_VIEW_PERMISSIONS=['*'])
     def test_token_expiration(self):
         url = reverse('dcim-api:site-list')
 
--- a/netbox/netbox/tests/test_authentication.py
+++ b/netbox/netbox/tests/test_authentication.py
@@ -85,10 +85,10 @@ class TokenAuthenticationTestCase(APITestCase):
         token1.save()
         token2.enabled = False
         token2.save()
-        response = self.client.get(url, HTTP_AUTHORIZATION=f'Token {token1.key}')
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Token {token1.token}')
         self.assertEqual(response.status_code, 403)
         self.assertEqual(response.data['detail'], 'Token disabled')
-        response = self.client.get(url, HTTP_AUTHORIZATION=f'Bearer {TOKEN_PREFIX}{token2.key}')
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Bearer {TOKEN_PREFIX}{token2.key}.{token2.token}')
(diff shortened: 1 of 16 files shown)


pheus:
I’ve opened this as a draft PR to get an initial round of feedback on the approach and implementation.

Once the changes from #20823 are merged from `main` into the `feature` branch, I’ll rebase this work to align with the updated code and reduce potential merge conflicts.

In the meantime, any early feedback on the model changes, API surface, or tests is very welcome. Thanks in advance for taking a look!

pheus:
I’ve just rebased this PR onto the newly merged branch and resolved the merge conflicts. The changes are now up to date and ready for review.

Thank you very much for taking a look!

netbox/users/models/tokens.py
```diff
@@ -180,6 +185,26 @@ def token(self, value):
...
+    @property
+    def is_active(self):
+        """
+        Check whether the token is active (enabled and not expired).
+        """
+        if not self.enabled:
+            return False
+        if self.is_expired:
+            return False
+        return True
```
jeremystretch: We can simplify this a bit.
```suggestion
        return self.enabled and not self.is_expired
```

netbox/netbox/tests/test_authentication.py
```diff
@@ -66,6 +66,30 @@ def test_v2_token_invalid(self):
...
+
+        # Request with a disabled token should fail
+        token1.enabled = False
+        token1.save()
+        token2.enabled = False
+        token2.save()
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Token {token1.key}')
+        self.assertEqual(response.status_code, 403)
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Bearer {TOKEN_PREFIX}{token2.key}')
+        self.assertEqual(response.status_code, 403)
```
jeremystretch: We can include some additional validation here to ensure that auth fails for the expected reason:
```suggestion
        response = self.client.get(url, HTTP_AUTHORIZATION=f'Token {token1.key}')
        self.assertEqual(response.status_code, 403)
        self.assertEqual(response.data['detail'], "Token disabled")
        response = self.client.get(url, HTTP_AUTHORIZATION=f'Bearer {TOKEN_PREFIX}{token2.key}')
        self.assertEqual(response.status_code, 403)
        self.assertEqual(response.data['detail'], "Token disabled")
```

netbox/users/tests/test_models.py
```diff
@@ -20,6 +20,26 @@ def setUpTestData(cls):
...
+        # Token with future expiration
+        token.expires = timezone.now() + timedelta(days=1)
+        self.assertTrue(token.is_active)
+
+        # Token with past expiration
+        token.expires = timezone.now() - timedelta(days=1)
+        self.assertFalse(token.is_active)
+
+        # Token with disabled status
+        token.enabled = False
```
jeremystretch: I suggest creating a new Token instance to evaluate each combination of `enabled` and `expires`. Here, for instance, `expires` is already set to a past time at line 36, so `is_active` will be false regardless of the value of `enabled`.

netbox/users/tests/test_api.py
```diff
@@ -323,6 +329,19 @@ def test_reassign_token(self):
         token1.refresh_from_db()
         self.assertEqual(token1.user, user1, "Token's user should not have changed")
 
+    def test_retrieve_includes_enabled(self):
```
jeremystretch: I don't think we need a separate test for the individual field. This should be covered by `test_create_object()`, where the serialized object is evaluated using `assertInstanceEqual()`.

jeremystretch (commented):
Nice work @pheus! Just had a few minor suggestions.

netbox/users/models/tokens.py
pheus: 💯 
This is way better! Thank you!

netbox/users/tests/test_models.py
pheus: Agreed. I've added all possible combination in the updated commit. Thank you!

netbox/users/tests/test_api.py
pheus: I've removed it. Thanks

jeremystretch (approved):

pheus:
Thank you very much for the thoughtful review and suggestions! I really appreciate you taking the time to go through this. Your feedback is super helpful and will definitely help me improve!

pheus:
I’m really sorry about that. I ran the `users` tests locally but completely overlooked the authentication tests under `netbox`. I should have caught this earlier.

Thank you for pushing the fix and for your patience. I’ll make sure to run the full set of relevant tests next time. 🙏

jeremystretch:
No worries!

PR #20877, opened 2025-11-25
Fixes #20371: Add Power Port Type (Molex Micro-Fit 2x3)

Etibru:
Fixes: https://github.com/netbox-community/netbox/issues/20371
- Add Molex Micro-Fit 2x3 on PowerPortTypeChoices and PowerOutletTypeChoices


jeremystretch (approved):
Thanks @Etibru!

PR #20879, opened 2025-11-25
20878 use database routing when running script

### LINKED ISSUE #20878: Script run does not use database routing
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.6

### Python Version

3.10

### Steps to Reproduce

The easiest way to see this is with branching:

1. Have a custom script that modifies some data (example script below).
2. Create a branch and activate the branch
3. Run the script without the "Commit Changes" box checked

You can see the issue here: https://github.com/netbox-community/netbox/blob/main/netbox/extras/jobs.py#L45

```
                with transaction.atomic():
                    if not commit:
                        raise AbortTransaction()
```

In all other places in the codebase we use the correct `with transaction.atomic(using=router.db_for_write(model)):`

See NetBox branching issue: [#270](https://github.com/netboxlabs/netbox-branching/issues/270)

```
from typing import Any

from extras.scripts import Script, ObjectVar
from dcim.models import Device


class BranchDummy(Script):
    class Meta:
        name = "Branch Dummy"
        description = "Test for branching"
        commit_default = True
        scheduling_enabled = False

    device = ObjectVar(
        model=Device,
        label="Device",
    )

    def run(self, data: Any, commit: bool) -> str:  # type: ignore
        device = data["device"]

        device.name = f"Branch Dummy {device.name}"

        device.full_clean()
        device.save()

        self.log_info(f"Device [{device.name}]({device.get_absolute_url()}) updated")
```

### Expected Behavior

The script should revert the changes, so no changes are made.  This is the case if you run it on the main branch.

### Observed Behavior

The changes are not reversed.

arthanson:
### Fixes: #20878 

Two main changes were needed:
* The atomic transaction wrapping the script needs to be done on default database and also the branch database as the script could potentially modify data in either.  Device was picked as a branch aware model to get the correct db_for_write.
* The request processor code needs to be run even if not committing in order to pick up the branch processor that actually switches to the branch.   However the event_tracking request_processor needs to be skipped so it doesn't process events (like a web hook setup on data being modified).

### DIFF
--- a/netbox/extras/jobs.py
+++ b/netbox/extras/jobs.py
@@ -44,10 +44,18 @@ class ScriptJob(JobRunner):
                 # A script can modify multiple models so need to do an atomic lock on
                 # both the default database (for non ChangeLogged models) and potentially
                 # any other database (for ChangeLogged models)
-                with transaction.atomic(using=router.db_for_write(Device)):
-                    script.output = script.run(data, commit)
-                    if not commit:
-                        raise AbortTransaction()
+                branch_db = router.db_for_write(Device)
+                with transaction.atomic(using='default'):
+                    # If branch database is different from default, wrap in a second atomic transaction
+                    if branch_db != 'default':
+                        with transaction.atomic(using=branch_db):
+                            script.output = script.run(data, commit)
+                            if not commit:
+                                raise AbortTransaction()
+                    else:
+                        script.output = script.run(data, commit)
+                        if not commit:
+                            raise AbortTransaction()
             except AbortTransaction:
                 script.log_info(message=_("Database changes have been reverted automatically."))
                 if script.failed:
--- a/netbox/extras/jobs.py
+++ b/netbox/extras/jobs.py
@@ -109,9 +109,6 @@ class ScriptJob(JobRunner):
         script.request = request
         self.logger.debug(f"Request ID: {request.id if request else None}")
 
-        # Execute the script. Always use request processors to set up the execution context (e.g. branch activation).
-        # The event_tracking context manager is one of the request processors and will be activated regardless of
-        # the commit flag, but it only records changes when commit=True.
         if commit:
             self.logger.info("Executing script (commit enabled)")
         else:
--- a/netbox/extras/jobs.py
+++ b/netbox/extras/jobs.py
@@ -3,6 +3,7 @@ import traceback
 from contextlib import ExitStack
 
 from django.db import router, transaction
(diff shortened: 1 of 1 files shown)


netbox/extras/jobs.py
```diff
@@ -42,10 +44,18 @@ def run_script(self, script, request, data, commit):
...
                 # both the default database (for non ChangeLogged models) and potentially
                 # any other database (for ChangeLogged models)
-                with transaction.atomic():
-                    script.output = script.run(data, commit)
-                    if not commit:
-                        raise AbortTransaction()
+                branch_db = router.db_for_write(Device)
+                with transaction.atomic(using='default'):
+                    # If branch database is different from default, wrap in a second atomic transaction
+                    if branch_db != 'default':
```
jnovinger: Can we make this more generic, rather than talking specifically about a `branch_db`?

netbox/extras/jobs.py
```diff
@@ -42,10 +44,18 @@ def run_script(self, script, request, data, commit):
...
+                        with transaction.atomic(using=branch_db):
+                            script.output = script.run(data, commit)
+                            if not commit:
+                                raise AbortTransaction()
+                    else:
+                        script.output = script.run(data, commit)
+                        if not commit:
+                            raise AbortTransaction()
```
jnovinger, line 62: Just to be clear, I don't think this guarantees true cross-connection atomicity, even though they are talking to the same DB, I'm pretty sure they are doing so via two separate connections.

As it stands, outside of exceptional circumstances, I think that this is pretty safe and handles the the failure cases (where something failure in either the outer `atomic()` block or the inner `atomic()` block). But, I think that potentially changes if anything is ever added after the `script.run(...)` invocation but before the end of the outer `atomic()` block. I think in that case, it's possible that the inner block can succeed and commit while the outer block can still fail.

Do you think it would be worthwhile to add a comment here making it clear that (outside of the `not commit` cases) no other logic should ever exist here (or should be _very_ carefully thought out), after the `script.run(...)` invocation and before exiting the outer block?

jnovinger (commented):
I had a couple of questions:

1. Is there any way to include a regression test for this new behavior? Given the particular order of operations and specific conditions being checked, it seems like it might be easy to break in future work.

2. A general question about the nested `atomic()` context manager blocks and potential edge-cases (see the comment/question in the code).

jnovinger (approved):

PR #20890, opened 2025-11-26
Closes #20639: Inherit ConfigContext from ancestor Platforms

### LINKED ISSUE #20639: Ability to apply parent platform config context values to device in children platforms
### NetBox version

v4.4.2

### Feature type

### Proposed functionality

Config contexts can use `Platforms` as conditions to apply given context to a particular device. With introduction of parent field for the platforms I believe it would be beneficial for Config Contexts to apply data to device based on parent as well and not just a child.

Currently, only if device is assigned to exactly that specific platform config contexts would be applied, but we have following platform hierarchy:
- SROS:
  - SROS-21
  - SROS-24

Config contexts defined for parent SROS will be rendered/applied for devices that have platforms SROS-21 and SROS-24.

### Use case

This will remove duplication of values between children platforms, that could have been defined at parent level.

### Database changes

N/A

### External dependencies

N/A

pheus:
### Fixes: #20639

This change applies ConfigContext to objects whose platforms descend from any assigned platform. This aligns platform behavior with regions, site groups, locations, and roles.

### DIFF
--- a/netbox/extras/querysets.py
+++ b/netbox/extras/querysets.py
@@ -46,6 +46,10 @@ class ConfigContextQuerySet(RestrictedQuerySet):
         # Match against the directly assigned role as well as any parent roles.
         device_roles = obj.role.get_ancestors(include_self=True) if obj.role else []
 
+        # Match against the directly assigned platform as well as any parent platforms.
+        platform = getattr(obj, 'platform', None)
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @pheus!

PR #20931, opened 2025-12-05
Fixes #20759: Group object types by app in permission form

### LINKED ISSUE #20759: Unable to see full object type names when creating new permissions
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.5

### Python Version

3.10

### Steps to Reproduce

- Install `Custom Objects` or `Changes` plugin
- Go to `Admin` -> `Permissions` -> `+ Add`
- Under `Objects` -> `Available` scroll down the entries for `Custom Objects` or `Changes`

### Expected Behavior

I should be able to see the name of the Object Types I'm adding to the `Permission`

### Observed Behavior

I am unable to see the Object Types because they extend off the side of the form

Image

jnovinger:
### Fixes: #20759

Modified the `ObjectPermissionForm` to use `optgroups` for organizing object types by application. This shortens the display names (e.g., "permission" instead of "Authentication and Authorization | permission") while maintaining clear organization through visual grouping.

Changes:
- Updated `get_object_types_choices()` to return nested `optgroup` structure
- Changed `AvailableOptions` and `SelectedOptions` widgets to handle `optgroups`
- Modified TypeScript `moveOptions` to preserve `optgroup` structure
- Added hover text showing full model names


netbox/users/forms/model_forms.py
```diff
@@ -283,10 +283,41 @@ def save(self, *args, **kwargs):
...
+        if current_app != app_label:
+            if current_group:
+                choices.append((current_app, current_group))
+            current_app = app_label
+            current_group = []
+
+        # Add model to current group using just the model's verbose name
+        model_class = ot.model_class()
+        model_name = model_class._meta.verbose_name if model_class else ot.model
+        current_group.append((ot.pk, model_name.title()))
```
jeremystretch: Use `utilities.string.title` here instead of the builtin to avoid camel-casing acronyms in model names (e.g. FHRP Group).

netbox/users/forms/model_forms.py
```diff
@@ -283,10 +283,41 @@ def save(self, *args, **kwargs):
...
-    ]
+    """
+    Generate choices for object types grouped by app label using optgroups.
+    Returns nested structure: [(app_label, [(id, model_name), ...]), ...]
+    """
+    from django.apps import apps
+
+    choices = []
+    current_app = None
+    current_group = []
```
jeremystretch: I'd suggest using a `defaultdict` to compile the choices so you don't need to track the current group for each iteration. Then when the `for` loop has completed you can convert the dict to a nested two-tuple. Just my preference though.

netbox/project-static/styles/transitional/_forms.scss
```diff
@@ -32,3 +32,14 @@ form.object-edit {
...
 }
+
+// Make optgroup labels sticky when scrolling through select elements
+select[multiple] {
+  optgroup {
+    position: sticky;
+    top: 0;
+    background-color: var(--bs-body-bg);
+    font-weight: bold;
+    padding: 0.25rem 0.5rem;
```
jeremystretch: I'd argue for removing both X and Y padding from `<optgroup>` elements to make better use of the space within a multi-select field. I'd also suggest reducing the left-padding of child `<option>` elements from the default of 20px, which seems excessive IMO.

netbox/utilities/forms/widgets/select.py
```diff
@@ -72,9 +72,22 @@ class AvailableOptions(forms.SelectMultiple):
     will be empty.) Employed by SplitMultiSelectWidget.
     """
     def optgroups(self, name, value, attrs=None):
```
jeremystretch, line 76: I notice that the logic under `optgroups()` and `create_option()` is nearly identical for AvailableOptions and SelectedOptions. Do you think we could move these to a base class that infers the appropriate behavior from a class attribute? Trying to avoid the redundancy if we can.

jeremystretch (changes requested):
Nice work!

netbox/utilities/forms/widgets/select.py
jnovinger: I'll take a look, thanks.

jeremystretch (approved):

PR #20956, opened 2025-12-10
#20875 fix updating of denormalized fields (_site, _location, _rack) for component models

### LINKED ISSUE #20875: Interface does not appear under interfaces' site filter
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.5

### Python Version

3.10

### Steps to Reproduce

Load the interfaces with a site id filter. (URL//dcim/interfaces/?site_id=2)
Insert a module to the device which is allocated to the site. (The module has to have interface template)
Load the interfaces with a site id filter. (URL//dcim/interfaces/?site_id=2) again.

### Expected Behavior

With module creation the interface should be able to be filtered with site.

### Observed Behavior

The interfaces are created, but cannot filter them by site. If I edit (with nothing) and save the interface on page the edited interface appears.

created interfaces:

Image

found interfaces with site id:

Image

found interfaces with device ids:

Image

After editing the first interface:

Image

arthanson:
### Fixes: #20875 

Several cases were previously missed:
* netbox/dcim/models/devices.py - when components are instantiated from the device-template and the denormalized fields were never getting updated.
* netbox/dcim/models/modules.py - when creating a module and instantiating components
* netbox/dcim/signals.py - when changing the site for a location, when changing the site for a rack

Also checked the API and it looks like the API and this should also affect those.  Bulk updates on the API do serializer.save() on each object which will cause the signal handler to run.

### DIFF
--- a/netbox/dcim/signals.py
+++ b/netbox/dcim/signals.py
@@ -45,7 +45,6 @@ def handle_location_site_change(instance, created, **kwargs):
         PowerPanel.objects.filter(location__in=locations).update(site=instance.site)
         CableTermination.objects.filter(_location__in=locations).update(_site=instance.site)
         # Update component models for devices in these locations
-        # (since Device.objects.update() doesn't trigger post_save signals)
         for model in COMPONENT_MODELS:
             model.objects.filter(device__location__in=locations).update(_site=instance.site)
 
@@ -58,7 +57,6 @@ def handle_rack_site_change(instance, created, **kwargs):
     if not created:
         Device.objects.filter(rack=instance).update(site=instance.site, location=instance.location)
         # Update component models for devices in this rack
-        # (since Device.objects.update() doesn't trigger post_save signals)
         for model in COMPONENT_MODELS:
             model.objects.filter(device__rack=instance).update(
                 _site=instance.site,
--- a/netbox/dcim/signals.py
+++ b/netbox/dcim/signals.py
@@ -6,11 +6,25 @@ from django.dispatch import receiver
 from dcim.choices import CableEndChoices, LinkStatusChoices
 from virtualization.models import VMInterface
 from .models import (
-    Cable, CablePath, CableTermination, Device, FrontPort, Interface, PathEndpoint, PowerPanel, Rack, Location,
+    Cable, CablePath, CableTermination, ConsolePort, ConsoleServerPort, Device, DeviceBay, FrontPort, Interface,
+    InventoryItem, ModuleBay, PathEndpoint, PowerOutlet, PowerPanel, PowerPort, Rack, RearPort, Location,
     VirtualChassis,
 )
 from .models.cables import trace_paths
-from .utils import create_cablepath, rebuild_paths, update_device_components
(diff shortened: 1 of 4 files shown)


netbox/dcim/utils.py
```diff
@@ -76,3 +76,36 @@ def update_interface_bridges(device, interface_templates, module=None):
             )
             interface.full_clean()
             interface.save()
+
+
+def update_device_components(device):
```
jeremystretch: Rather than introducing a new function to handle this, could we just extend the `instantiate()` method on each component to populate the necessary fields prior to calling `bulk_create()`?

arthanson: It's used in the signal handler as well which if we did instantiate would mean separate code and I wanted to keep the update in one place in-case we ever add additional fields here or need to change the logic in any way.

netbox/dcim/signals.py
```diff
@@ -53,6 +42,9 @@ def handle_rack_site_change(instance, created, **kwargs):
     """
     if not created:
         Device.objects.filter(rack=instance).update(site=instance.site, location=instance.location)
+        # Update component models for devices in this rack
+        for device in Device.objects.filter(rack=instance):
+            update_device_components(device)
```
jnovinger: While this is safe below, in that it doesn't introduce a large number of new queries, I don't think it's so safe here (or in the `post_save` handler for `Location` above), the way it's constructed. Imagine having a site with 1000+ devices, since there are 10 component models, we've now queued up 10,000+ SQL queries (`n` devices in the site x `m` component models.

I think if you reconstructed this something like the following, then you should only ever need `m + 1` queries.

```
devices_in_site = Device.objects.filter(rack=instance)
devices_in_site.update(site=instance.site, location=instance.location)

for model in COMPONENT_MODELS:
	model.objects.filter(device__in=devices_in_site).update(
		_site=instance.site,
		_location=instance.location,
		_rack=instance.rack,
	)
```

jnovinger (changes requested):
I think I agree with @jeremystretch's suggestion about populating fields in `instantiate()`. I see two distinct problems here that may not need the same solution:

1. Device/Module creation

When components are instantiated via `bulk_create()`, the denormalized fields are NULL. Jeremy's suggestion handles this populate `_site`, `_location`, and `_rack` in each template's `instantiate()` method so the fields are set correctly before `bulk_create()` is called. No post-processing `UPDATE` needed.

2. Location/Rack changes (signal handlers)

The helper function creates an `O(m * n)` query pattern here. When a location's site changes with 100 devices, this executes 1,000 `UPDATE` queries (100 devices × 10 component models). The signal handlers could instead bulk-update per component model (see suggestion in other comment for how I'd approach that).

Given that needed change of behavior, I don't think the helper function provides enough value to justify abstracting these two distinct patterns. Case 1 can be handled in `instantiate()`, and Case 2 needs efficient bulk updates.

Although, perhaps I'm missing something here. Let me know.

jnovinger (approved):

PR #20973, opened 2025-12-12
Additional work for FR #20788

jeremystretch:
This PR captures some follow-on work for v4.5 FR #20788 that was identified following the initial implementation.

- CableTermination
  - Add the 'connector' integer field to index terminations on either end of a cable
  - Rename the `position` integer field to `cable_positions` (plural) and convert it to an array of integers
- All cabled models (device components, power feeds, circuit terminations)
  - Introduce the `cable_connector` integer field
  - Rename the `cable_position` integer field to `cable_positions` (plural) and convert it to an array of integers
  - Introduce `set_cable_termination()` and `clear_cable_termination()` methods to simplify setting/clearing of cable termination attributes
- Rework cable profiles to enumerate A & B connectors
- Standardize cable profile names
- Add cable profiles for specific cable types
- Rename migration files
- Extend cable path test suite


netbox/dcim/cable_profiles.py
```diff
@@ -1,108 +1,387 @@
...
+        (4, 2): (2, 4),
+        (4, 3): (3, 4),
+        (4, 4): (4, 4),
+    }
+
+
+class ShuffleBreakout2x8CableProfile(BaseCableProfile):
+    """
+    Temporary solution for mapping 2 front/rear ports to 8 discrete interfaces
+    """
```
jnovinger: I'm curious, what makes this temporary? Is this just something that we need time to iterate on more thoroughly?

jnovinger (approved):
Had one question and one suggestion: the happy path seems well tested, would be good to confirm the error paths behave as expected too. I think a follow-up housekeeping issue could add tests that explicitly verify validation errors are raised when constraints are violated (e.g., too many terminations for a profile, mismatched connector counts).

netbox/dcim/cable_profiles.py
jeremystretch: It was originally a healf-measure to address niche cases in the absence of FR #20972, but I've reorganized and renamed the cable profiles now so that it more correctly represents a shuffled breakout cable.

jnovinger (approved):

PR #20975, opened 2025-12-12
Fixes #20929: Require render_config permission for UI config rendering

### LINKED ISSUE #20929: Expand permissions to include "Rendered Config" contexts.
### NetBox version

v4.4.7

### Feature type

Data model extension

### Proposed functionality

Feature request. In short, our customer would like to be able to limit whether a user of NetBox can view "Rendered Config" and other config contexts from within the "Devices" model in NetBox. 

### Use case

This is to ensure that when configs are rendered along with their passwords and other secrets, these are not openly visible to users whom they don't want to be able to view this information.

You can sort do this to a limited degree by denying access to Extras>Contexts, however the "Rendered Config" is visible regardless of this still on the device itself.

Could we expand permissions to include this?

### Database changes

_No response_

### External dependencies

_No response_

jnovinger:
### Fixes: #20929 

- Modified `ObjectRenderConfigView.has_permission()` to require both view and render_config permissions
- Added `remove_permissions()` test helper to remove permissions from existing ObjectPermission objects
- Added regression tests for Device and VirtualMachine render-config permission enforcement

The `render_config` permission action was introduced in #16681 for API endpoints. This extends PR_7604_description to the UI render-config tabs, preventing users from viewing rendered configurations without explicit permission.

### DIFF
--- a/netbox/dcim/tests/test_views.py
+++ b/netbox/dcim/tests/test_views.py
@@ -11,6 +11,7 @@ from core.models import ObjectType
 from dcim.choices import *
 from dcim.constants import *
 from dcim.models import *
+from extras.models import ConfigTemplate
 from ipam.models import ASN, RIR, VLAN, VRF
 from netbox.choices import CSVDelimiterChoices, ImportFormatChoices, WeightUnitChoices
 from tenancy.models import Tenant
@@ -2338,6 +2339,29 @@ class DeviceTestCase(ViewTestCases.PrimaryObjectViewTestCase):
 
         url = reverse('dcim:device_inventory', kwargs={'pk': device.pk})
         self.assertHttpStatus(self.client.get(url), 200)
+
+    @tag('regression')  # #20929
+    def test_device_renderconfig_permission(self):
+        configtemplate = ConfigTemplate.objects.create(
+            name='Test Config Template',
+            template_code='Config for device {{ device.name }}'
+        )
+        device = Device.objects.first()
+        device.config_template = configtemplate
+        device.save()
+        url = reverse('dcim:device_render-config', kwargs={'pk': device.pk})
+
+        # User with only view permission should NOT be able to render config
+        self.add_permissions('dcim.view_device')
+        self.assertHttpStatus(self.client.get(url), 403)
+
+        # With render_config permission added should be able to render config
+        self.add_permissions('dcim.render_config_device')
+        self.assertHttpStatus(self.client.get(url), 200)
+
+        # With view permission removed should NOT be able to render config
+        self.remove_permissions('dcim.view_device')
+        self.assertHttpStatus(self.client.get(url), 403)
 
 
 class ModuleTestCase(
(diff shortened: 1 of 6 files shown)


netbox/dcim/tests/test_views.py
```diff
@@ -2339,6 +2340,29 @@ def test_device_inventory(self):
         url = reverse('dcim:device_inventory', kwargs={'pk': device.pk})
         self.assertHttpStatus(self.client.get(url), 200)
 
+    @tag('regression')  # #20929
+    def test_device_renderconfig_permission(self):
```
jeremystretch: Could we make this a standard view test?
```suggestion
    def test_device_renderconfig(self):
```

netbox/virtualization/tests/test_views.py
```diff
@@ -326,6 +327,29 @@ def test_virtualmachine_interfaces(self):
         url = reverse('virtualization:virtualmachine_interfaces', kwargs={'pk': virtualmachine.pk})
         self.assertHttpStatus(self.client.get(url), 200)
 
+    @tag('regression')  # #20929
+    def test_virtualmachine_renderconfig_permission(self):
```
jeremystretch: Same question as above.

netbox/utilities/testing/base.py
```diff
@@ -67,6 +67,18 @@ def add_permissions(self, *names):
...
+        """
+        Remove a set of permissions from the test user. Accepts permission names in the form <app>.<action>_<model>.
+        """
+        for name in names:
+            object_type, action = resolve_permission_type(name)
+            obj_perms = ObjectPermission.objects.filter(
+                actions__contains=[action], object_types=object_type, users=self.user
+            )
+            for obj_perm in obj_perms:
+                obj_perm.users.remove(self.user)
```
jeremystretch: `add_permissions()` creates a new ObjectPermission instance for each permission passed, so it might be prudent to simply delete the instance here instead of removing the assigned user. (This approach could also create a name conflict if calling `add_permissions()` after `remove_permissions()` for the same model & action.)

netbox/extras/views.py
```diff
@@ -1082,6 +1083,14 @@ class ObjectRenderConfigView(generic.ObjectView):
     base_template = None
     template_name = 'extras/object_render_config.html'
 
+    def has_permission(self):
+        if super().has_permission():  # enforce base required permission
+            perm = get_permission_for_model(self.queryset.model, 'render_config')
+            if self.request.user.has_perm(perm):
+                self.queryset = self.queryset.restrict(self.request.user, 'render_config')
```
jeremystretch: I'd prefer to avoid altering the queryset directly here if we can. Instead, how about just adding the necessary permission by name to `additional_permissions` under each view?

jeremystretch (changes requested):

netbox/extras/views.py
jnovinger: My goal here was to ensure that the queryset is also restricted relative to the permission action. Without this, I don' think that will happen, right?

But, I suppose that's a bit outside of the ask of original FR.

jeremystretch (approved):

PR #20986, opened 2025-12-15
Fixed #20944: Ensure cached scope fields stay consistent when Region, Site, or Location changes

### LINKED ISSUE #20944: Related objects "clusters" not updated for site-group
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.7

### Python Version

3.10

### Steps to Reproduce

1. Create a site, Contoso-HQ and attach it to a site group Contoso-SiteGroup
2. Create a cluster scoped to site Contoso-HQ

Related objects for site group Contoso-SiteGroup display one site and one cluster (expected part)

3. Now attach Contoso-Site to another site group "Headquarters"
4. Site group Headquarters related objects doesn't show the cluster (only the site)
5. Contoso-SiteGroup related objects still shows the cluster (but not the site)
6. Editing the cluster and saving it solves the issue

### Expected Behavior

The related object should be updated for clusters

### Observed Behavior

Related object still show the previous value (but the cluster is still attached to his parent site)

kprince28:

### DIFF
--- a/netbox/dcim/signals.py
+++ b/netbox/dcim/signals.py
@@ -7,9 +7,11 @@ from dcim.choices import CableEndChoices, LinkStatusChoices
 from ipam.models import Prefix
 from virtualization.models import Cluster, VMInterface
 from wireless.models import WirelessLAN
-from .models import (Cable, CablePath, CableTermination, ConsolePort, ConsoleServerPort, Device, DeviceBay, FrontPort,
-                     Interface, InventoryItem, Location, ModuleBay, PathEndpoint, PowerOutlet, PowerPanel, PowerPort,
-                     Rack, RearPort, Site, VirtualChassis)
+from .models import (
+    Cable, CablePath, CableTermination, ConsolePort, ConsoleServerPort, Device, DeviceBay, FrontPort, Interface,
+    InventoryItem, Location, ModuleBay, PathEndpoint, PowerOutlet, PowerPanel, PowerPort, Rack, RearPort, Site,
+    VirtualChassis,
+)
 from .models.cables import trace_paths
 from .utils import create_cablepath, rebuild_paths
 
--- a/netbox/dcim/signals.py
+++ b/netbox/dcim/signals.py
@@ -186,7 +186,7 @@ def update_mac_address_interface(instance, created, raw, **kwargs):
 
 @receiver(post_save, sender=Location)
 @receiver(post_save, sender=Site)
-def sync_cached_scope_fields(sender, instance, **kwargs):
+def sync_cached_scope_fields(instance, created, **kwargs):
     """
     Rebuild cached scope fields for all CachedScopeMixin-based models
     affected by a change in a Region, SiteGroup, Site, or Location.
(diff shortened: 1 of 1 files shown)


netbox/dcim/signals.py
```diff
@@ -1,15 +1,15 @@
...
-    Cable, CablePath, CableTermination, ConsolePort, ConsoleServerPort, Device, DeviceBay, FrontPort, Interface,
-    InventoryItem, ModuleBay, PathEndpoint, PowerOutlet, PowerPanel, PowerPort, Rack, RearPort, Location,
-    VirtualChassis,
-)
+from ipam.models import Prefix
+from virtualization.models import Cluster, VMInterface
+from wireless.models import WirelessLAN
+from .models import (Cable, CablePath, CableTermination, ConsolePort, ConsoleServerPort, Device, DeviceBay, FrontPort,
+                     Interface, InventoryItem, Location, ModuleBay, PathEndpoint, PowerOutlet, PowerPanel, PowerPort,
+                     Rack, RearPort, Site, VirtualChassis)
```
jeremystretch: Revert formatting change
```suggestion
from .models import (
    Cable, CablePath, CableTermination, ConsolePort, ConsoleServerPort, Device, DeviceBay, FrontPort, Interface,
    InventoryItem, Location, ModuleBay, PathEndpoint, PowerOutlet, PowerPanel, PowerPort, Rack, RearPort, Site,
    VirtualChassis,
)
```

netbox/dcim/signals.py
```diff
@@ -180,3 +180,36 @@ def update_mac_address_interface(instance, created, raw, **kwargs):
...
+def sync_cached_scope_fields(sender, instance, **kwargs):
+    """
+    Rebuild cached scope fields for all CachedScopeMixin-based models
+    affected by a change in a Region, SiteGroup, Site, or Location.
+
+    This method is safe to run for objects created in the past and does
+    not rely on incremental updates. Cached fields are recomputed from
+    authoritative relationships.
+    """
+
```
jeremystretch, line 200: We can check whether the triggering instance is newly created, and if so skip it because we know it can't yet be referenced by other objects. (The function signature also needs to be updated to capture the `created` argument.)
```suggestion
    if created:
        return

```

netbox/dcim/signals.py
```diff
@@ -180,3 +180,36 @@ def update_mac_address_interface(instance, created, raw, **kwargs):
...
+    if isinstance(instance, Location):
+        filters = {'_location': instance}
+    elif isinstance(instance, Site):
+        filters = {'_site': instance}
+    else:
+        return
+
+    for model in (Prefix, Cluster, WirelessLAN):
```
jeremystretch, line 210: We should add a comment here that these models are included because they're all subclasses of CachedScopeMixin.

jeremystretch (changes requested):
Thanks @kprince28! Just had a couple minor tweaks.

jeremystretch (approved):
Thanks @kprince28!

PR #20994, opened 2025-12-16
Fix JSON serialization error in get_installed_apps()

### LINKED ISSUE #20993: /api/status/ returns JSON serialization error with django-health-check installed
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.8

### Python Version

3.12

### Steps to Reproduce

1. Install NetBox 4.4.8
2. Install `django-health-check` package (version 3.20.0)
3. Add `health_check` to Django's `INSTALLED_APPS`
4. Access `/api/status/` endpoint

### Expected Behavior

The `/api/status/` endpoint should return a valid JSON response containing version information for all installed apps.

### Observed Behavior

The endpoint returns:

```
TypeError: Object of type type is not JSON serializable
```

**Root Cause Analysis:**

The `django-health-check` package uses `setuptools-scm` for versioning. In their `health_check/__init__.py`:

```python
from health_check import _version
VERSION = _version.VERSION_TUPLE
```

Due to how `setuptools-scm` generates `_version.py`, `VERSION_TUPLE` is defined as:

```python
VERSION_TUPLE: Tuple[int, int, int] = object  # type placeholder at runtime
```

At runtime, `VERSION_TUPLE` is actually `<class 'object'>` rather than an actual tuple like `(3, 20, 0)`.

The `get_installed_apps()` function in `netbox/utilities/apps.py` retrieves this `VERSION` attribute and attempts to include it in the response dictionary. Since `<class 'object'>` is not JSON-serializable, the API call fails.

**Suggested Fix:**

Add a check in `get_installed_apps()` to skip VERSION values that are neither tuples nor strings:

```python
if version := getattr(app, 'VERSION', getattr(app, '__version__', None)):
    if type(version) is tuple:
        version = '.'.join(str(n) for n in version)
    elif not isinstance(version, str):
        # Skip non-serializable version types
        continue
    installed_apps[app_config.name] = version
```

This defensive approach ensures that any third-party package with a non-standard VERSION attribute won't break the `/api/status/` endpoint.

I have a fix ready and would be happy to submit a PR if this issue is assigned to me.

dantte-lp:
### Summary

This PR adds a defensive check in `get_installed_apps()` to skip VERSION attributes that are neither tuples nor strings, preventing JSON serialization errors in the `/api/status/` endpoint.

**The Problem:**

Some packages (like `django-health-check` using `setuptools-scm`) define their `VERSION` attribute as a type placeholder at runtime rather than an actual version value:

```python
# In django-health-check's _version.py (generated by setuptools-scm)
VERSION_TUPLE: Tuple[int, int, int] = object  # This is <class 'object'> at runtime!
```

When `get_installed_apps()` encounters this, it includes `<class 'object'>` in the dictionary, which fails JSON serialization.

**The Fix:**

Add an `elif` branch to skip non-serializable version types:

```python
if type(version) is tuple:
    version = '.'.join(str(n) for n in version)
elif not isinstance(version, str):
    # Skip non-serializable version types (e.g. setuptools-scm placeholders)
    continue
```

This is a minimal, defensive change that:
- Maintains existing behavior for tuple and string versions
- Gracefully handles edge cases from third-party packages
- Has no impact on correctly-behaving packages

**Testing:**

Verified fix resolves the issue with `django-health-check` 3.20.0 installed.

### DIFF
--- a/netbox/utilities/apps.py
+++ b/netbox/utilities/apps.py
@@ -11,6 +11,9 @@ def get_installed_apps():
         if version := getattr(app, 'VERSION', getattr(app, '__version__', None)):
             if type(version) is tuple:
                 version = '.'.join(str(n) for n in version)
+            elif not isinstance(version, str):
+                # Skip non-serializable version types (e.g. setuptools-scm placeholders)
+                continue
             installed_apps[app_config.name] = version
     return {
         k: v for k, v in sorted(installed_apps.items())


jnovinger:
Please do not submit PRs for issues until they have been reviewed and assigned to you for work. This is noted in our [contributing guide](https://github.com/netbox-community/netbox/blob/main/CONTRIBUTING.md#arrow_heading_up-submitting-pull-requests).

PR #20996, opened 2025-12-17
Fixes #19506: DCIM Filtering on Templates custom fields

### LINKED ISSUE #19506: Filtering MultiObjectVar for Custom Field does not works for related object = Module Bay Template
### Deployment Type

NetBox Cloud

### NetBox Version

v4.3.1

### Python Version

3.10

### Steps to Reproduce

1. Create a new custom field with object type = DCIM -> Module Type, type = Multiple Objects and related object type = DCM -> Module Bay Template

Image

2. Create a new module type, and press the object filtering button near the custom field

Image

### Expected Behavior

A new popup widget shows allowing the user select "Manufacturer" and "Device Type", listing all filtered modules bays:

Image

This screenshot is a Microsoft Paint concept, based on another working select filter in netbox

### Observed Behavior

The UI greys out and freezes

Image

callumau:
### Summary

Fixes issue #19506 by enabling filtering on DCIM template models’ custom fields and attributes.
Files Changed: [filtersets.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

### Impact

UI/API filtering: Adds comprehensive filtering capabilities for template objects, aligning template-level filters with their concrete counterparts.
Scope: Form-layer changes; no schema or migrations implied.

### DIFF
--- a/netbox/dcim/forms/filtersets.py
+++ b/netbox/dcim/forms/filtersets.py
@@ -24,35 +24,45 @@ __all__ = (
     'CableFilterForm',
     'ConsoleConnectionFilterForm',
     'ConsolePortFilterForm',
+    'ConsolePortTemplateFilterForm',
     'ConsoleServerPortFilterForm',
+    'ConsoleServerPortTemplateFilterForm',
     'DeviceBayFilterForm',
+    'DeviceBayTemplateFilterForm',
     'DeviceFilterForm',
     'DeviceRoleFilterForm',
     'DeviceTypeFilterForm',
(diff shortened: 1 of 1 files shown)


arthanson (changes requested):
I think you would probably want a DeviceComponentTemplateFilterForm that has device_type_id and module_type_id and have all these filter forms derive from that?  For some of these you have device_type_id in the fieldset but you don't have them defined.  I think you would want device_type_id and module_type_id on all of these filter forms?

PR #21005, opened 2025-12-18
Fixed #20950: Add missing module and device properties in module-bay

### LINKED ISSUE #20950: ModuleBay API GET - Missing Properties
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.7

### Python Version

3.10

### Steps to Reproduce

1. Create a new device type.
2. Create a new module type.
3. Create a new device.
4. Add a `Module Bays` component to the device.
5. Install a module in the created module bay.
6. API GET: `/api/dcim/module-bays/?device_id=<device_id>`

### Expected Behavior

Assuming the API spec is correct I would expect that the `installed_module` property contains a reference to the `device` and `module_bay`. 

### Observed Behavior

The `installed_module` property does not contain any `device` or `module_bay` references. In my case I use the Golang [client](https://github.com/netbox-community/go-netbox) (which is generated based on the API spec) that validates the returned API result and errors out because no value was given for the required properties `device` and `module_bay`.

I also think the same applies to the `module` property but I don't have a use-case where this value is set so I can't test it. 

```
{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 2913,
            "url": "http://localhost:8585/api/dcim/module-bays/2913/",
            "display_url": "http://localhost:8585/dcim/module-bays/2913/",
            "display": "bay0",
            "device": {
                "id": 5645,
                "url": "http://localhost:8585/api/dcim/devices/5645/",
                "display": "Something-101",
                "name": "Something-101",
                "description": ""
            },
            "module": null,
            "name": "bay0",
            "installed_module": {
                "id": 291,
                "url": "http://localhost:8585/api/dcim/modules/291/",
                "display": "bay0: SomeModule01 (291)",
                "serial": "",
                "description": ""
            },
            "label": "",
            "position": "",
            "description": "",
            "tags": [],
            "custom_fields": {},
            "created": "2025-12-08T16:45:10.445185Z",
            "last_updated": "2025-12-08T16:45:10.445200Z"
        }
    ]
}
```

kprince28:

### DIFF
--- a/netbox/dcim/api/serializers_/device_components.py
+++ b/netbox/dcim/api/serializers_/device_components.py
@@ -357,7 +357,7 @@ class ModuleBaySerializer(NetBoxModelSerializer):
     )
     installed_module = ModuleSerializer(
         nested=True,
-        fields=('id', 'url', 'display', 'device', 'module_bay'),
+        fields=('id', 'url', 'display', 'device', 'module_bay', 'serial', 'description'),
         required=False,
         allow_null=True
     )
--- a/netbox/dcim/api/serializers_/device_components.py
+++ b/netbox/dcim/api/serializers_/device_components.py
@@ -350,14 +350,14 @@ class ModuleBaySerializer(NetBoxModelSerializer):
     device = DeviceSerializer(nested=True)
     module = ModuleSerializer(
         nested=True,
-        fields=('id', 'url', 'display'),
+        fields=('id', 'url', 'display', 'device', 'module_bay'),
         required=False,
         allow_null=True,
         default=None
     )
     installed_module = ModuleSerializer(
         nested=True,
-        fields=('id', 'url', 'display', 'serial', 'description'),
+        fields=('id', 'url', 'display', 'device', 'module_bay'),
         required=False,
         allow_null=True
     )


netbox/dcim/api/serializers_/device_components.py
```diff
@@ -350,14 +350,14 @@ class ModuleBaySerializer(NetBoxModelSerializer):
...
-        fields=('id', 'url', 'display'),
+        fields=('id', 'url', 'display', 'device', 'module_bay'),
         required=False,
         allow_null=True,
         default=None
     )
     installed_module = ModuleSerializer(
         nested=True,
-        fields=('id', 'url', 'display', 'serial', 'description'),
+        fields=('id', 'url', 'display', 'device', 'module_bay'),
```
jnovinger: Why did we lose `serial` and `description` here?

kprince28: because I just follow the API specification in this API, so serial and description are not present in the API spec.

jnovinger: Understood. Unfortunately, removing fields from an API serializer like this is a breaking change that will need to be present in a minor release, as opposed to this, which will ship in a patch release.

It's also very possible that those fields were intended to be there all along and that the OpenAPI schema that's being generated is just incorrect.

Please restore these fields in this PR. We will need to follow up on this in a new issue. If you have a moment, we would very much appreciate if you could file a bug for this.

kprince28: I have added them now

jnovinger (approved):

PR #21006, opened 2025-12-18
Fixes #20320: Include parent PK in BulkEdit context for child objects

pheus:
### Fixes: #20320

This PR addresses the issue where the “Related Interfaces” selectors (Parent/Bridge/LAG) are disabled when bulk-editing interfaces from an object child view (e.g. device → interfaces) in v4.4.0.

#### Summary / Rationale
When bulk actions are invoked from an object child view (e.g. `/dcim/devices/<pk>/interfaces/`), the bulk edit action URL no longer carries the parent-scoping query parameter (e.g. `device=<pk>`). Without that context, the bulk edit form can’t scope/populate the related-interface fields, leaving them grayed out.

#### Proposed Changes
- Enhance `BulkEdit.get_context()` to detect the “child view” case (presence of a parent `object` in the template context).
- When applicable, inject the parent object’s primary key into the bulk edit action URL parameters (e.g. `device=<parent.pk>` / `device_type=<parent.pk>`), while preserving any existing `request.GET` parameters (and `return_url` when present).
- Only add the parameter when it’s not already provided, to avoid altering behavior on global list views.

#### Tests Added
- Added a unit test to confirm the parent object's PK is included in `url_params` when rendering a bulk action in a child-object context:
  - `test_bulk_edit_get_context_child_object` creates a `DeviceType` parent and verifies `BulkEdit.get_context(context, Device)` injects `device_type=<parent.pk>` into `url_params` when the template context includes `object=<DeviceType>`.
- Also adjusted a plugin URL test to ensure the `@skipIf` decorator behaves correctly:
  - Moved the import of the dummy plugin model inside `test_get_url_plugin_model` so the module import doesn’t occur before the `@skipIf` condition is evaluated.

#### Notes / Open Question
I’m not 100% certain this is the preferred place to implement the fix (action context vs. view/template), but this approach keeps the change small and avoids duplicating templates.

#### How To Test
1. Go to a device → Interfaces tab.
2. Select multiple interfaces.
3. Click **Edit Selected**.
4. Confirm the Parent/Bridge/LAG fields are enabled and show relevant interface choices.
5. Confirm bulk edit from the global Interfaces list behaves as before (no unintended scoping).
6. Run the test suite (including plugin tests when `netbox.tests.dummy_plugin` is enabled) and confirm all tests pass.

### DIFF
--- a/netbox/netbox/object_actions.py
+++ b/netbox/netbox/object_actions.py
@@ -1,3 +1,4 @@
+from django.db.models import ForeignKey
 from django.template import loader
 from django.urls.exceptions import NoReverseMatch
 from django.utils.translation import gettext_lazy as _
@@ -174,6 +175,21 @@ class BulkEdit(ObjectAction):
     multi = True
     permissions_required = {'change'}
     template_name = 'buttons/bulk_edit.html'
+
+    @classmethod
+    def get_context(cls, context, model):
+        url_params = super().get_url_params(context)
+
+        # If this is a child object, pass the parent's PK as a URL parameter
+        if parent := context.get('object'):
+            for field in model._meta.get_fields():
+                if isinstance(field, ForeignKey) and field.remote_field.model == parent.__class__:
+                    url_params[field.name] = parent.pk
+                    break
+
+        return {
+            'url_params': url_params,
+        }
 
 
 class BulkRename(ObjectAction):
--- a/netbox/netbox/tests/test_object_actions.py
+++ b/netbox/netbox/tests/test_object_actions.py
@@ -1,11 +1,10 @@
 from unittest import skipIf
 
 from django.conf import settings
-from django.test import TestCase
+from django.test import RequestFactory, TestCase
 
-from dcim.models import Device
-from netbox.object_actions import AddObject, BulkImport
-from netbox.tests.dummy_plugin.models import DummyNetBoxModel
+from dcim.models import Device, DeviceType, Manufacturer
+from netbox.object_actions import AddObject, BulkEdit, BulkImport
 
 
 class ObjectActionTest(TestCase):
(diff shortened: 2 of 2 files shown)


pheus:
Hi everyone,

Is there any chance this could be included in the final v4.4 release?

I don’t usually ask for this, but since this appears to be a regression introduced in 4.4.0, I thought it might be worth considering for the last 4.4.x release, assuming the approach and implementation look acceptable to the maintainers.

Thanks a lot for taking a look, and for all the work that goes into these releases!

jeremystretch:
@pheus I've tagged it for v4.4.10

pheus:
Awesome, thank you!

netbox/netbox/object_actions.py
```diff
@@ -175,6 +176,21 @@ class BulkEdit(ObjectAction):
...
+    @classmethod
+    def get_context(cls, context, model):
+        url_params = super().get_url_params(context)
+
+        # If this is a child object, pass the parent's PK as a URL parameter
+        if parent := context.get('object'):
+            for field in model._meta.get_fields():
+                if isinstance(field, ForeignKey) and field.remote_field.model == parent.__class__:
+                    url_params[field.name] = parent.pk
+                    break
```
jnovinger, line 188: I initially wondered if this `break` might come back to bite us, if we ever had a model with multiple parents for some reason. However, if that's the case, then a whole lot needs to be reworked anyway. I think this is pretty pragmatic and correct for this case.

jnovinger (approved):

netbox/netbox/object_actions.py
pheus: Thanks for taking a close look at that and for the thoughtful review!

I had a similar concern about the `break` in case a model ever had multiple relevant parent relationships. My thinking was the same as yours: if we ever get to that point, we’d likely need more explicit scoping/context handling anyway. For the current child-view use case, stopping at the first matching FK felt like the most pragmatic approach.

Really appreciate you calling it out and confirming the reasoning.

PR #21010, opened 2025-12-18
Closes #20953: Show interfaces bridged to an interface in the UI

### LINKED ISSUE #20953: Show reverse bridge relationships on interface detail pages
### NetBox version

v4.4.8

### Feature type

### Proposed functionality

When an interface’s Bridged interface is set to another interface, only the source interface shows this relationship. The target interface has no indication that anything is bridged to it.

1. Create two interfaces "int-A" and "int-B"
2. on "int-A" set Bridged Interface to "int-B"
3. Look on the detail view of Int-A and under Related Interfaces it shows Bridge pointing to Int-B
4. Go to the detail page of Int-B and under Related Interfaces there is no indication that it has a Bridge

### Use case

Operators visiting the physical NIC cannot see which virtual interfaces depend on it without API queries or manually checking other interfaces.

### Database changes

None

### External dependencies

None

kprince28:

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -2856,10 +2856,7 @@ class InterfaceView(generic.ObjectView):
         vdc_table.configure(request)
 
         # Get bridge interfaces
-        bridge_interfaces = Interface.objects.restrict(request.user, 'view').filter(bridge=instance).select_related(
-            'device', 'module', 'parent', 'bridge', 'lag', 'vrf', 'untagged_vlan', 'qinq_svlan',
-            'primary_mac_address'
-        ).prefetch_related('tagged_vlans', 'wireless_lans', 'vdcs')
+        bridge_interfaces = Interface.objects.restrict(request.user, 'view').filter(bridge=instance)
         bridge_interfaces_table = tables.InterfaceTable(
             bridge_interfaces,
             exclude=('device', 'parent'),
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -13,7 +13,7 @@ from django.views.generic import View
 
 from circuits.models import Circuit, CircuitTermination
 from extras.views import ObjectConfigContextView, ObjectRenderConfigView
-from ipam.models import ASN, IPAddress, Prefix, VLANGroup, VLAN
+from ipam.models import ASN, IPAddress, Prefix, VLAN, VLANGroup
 from ipam.tables import InterfaceVLANTable, VLANTranslationRuleTable
 from netbox.object_actions import *
 from netbox.views import generic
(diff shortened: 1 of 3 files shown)


netbox/dcim/views.py
```diff
@@ -2856,7 +2856,10 @@ def get_extra_context(self, request, instance):
         vdc_table.configure(request)
 
         # Get bridge interfaces
-        bridge_interfaces = Interface.objects.restrict(request.user, 'view').filter(bridge=instance)
+        bridge_interfaces = Interface.objects.restrict(request.user, 'view').filter(bridge=instance).select_related(
+            'device', 'module', 'parent', 'bridge', 'lag', 'vrf', 'untagged_vlan', 'qinq_svlan',
+            'primary_mac_address'
+        ).prefetch_related('tagged_vlans', 'wireless_lans', 'vdcs')
```
jnovinger: Please remove these `select_related()` and `prefetch_related()` calls, they are already handled via `BaseTable.__init__()`.

netbox/templates/dcim/interface.html
```diff
@@ -112,6 +112,19 @@ <h2 class="card-header">{% trans "Related Interfaces" %}</h2>
             <th scope="row">{% trans "Bridge" %}</th>
             <td>{{ object.bridge|linkify|placeholder }}</td>
           </tr>
+          <tr>
+            <th scope="row">{% trans "Bridged Interfaces" %}</th>
+            <td>
+              {% if object.is_bridge %}
```
jnovinger: This check fails when the interface is not of type `TYPE_BRIDGE`.

netbox/templates/dcim/interface.html
```diff
@@ -438,7 +451,7 @@ <h2 class="card-header">
   {% if object.is_bridge %}
```
jnovinger: This check also fails if interface is not of type `TYPE_BRIDGE`.

jnovinger (changes requested):

netbox/templates/dcim/interface.html
kprince28: actually this is the property which is defined in the `Interface` model to check the `TYPE_BRIDGE`, so I reused that code

netbox/templates/dcim/interface.html
kprince28: same here

netbox/templates/dcim/interface.html
kprince28: This property checks whether the interface type is `TYPE_BRIDGE` or another type.

jnovinger: Yes, but an interface is [not required](https://netboxlabs.com/docs/netbox/models/dcim/interface/#bridged-interface) to be of type `TYPE_BRIDGE` in order to be a bridge(d interface).

kprince28: I have removed the check

jnovinger (approved):

PR #21040, opened 2025-12-26
Fix nested serializer validation for PATCH requests with {"id": X} format

casibbald:
When a nested serializer receives {"id": X} in a PATCH request, DRF was attempting to validate it as a full object, requiring all required fields (e.g., 'name' and 'group' for Tenant). This caused validation errors like 'group and name are required' even though the nested serializer should only look up the object by ID.

The fix overrides run_validation() to skip field-level validation for nested serializers, allowing to_internal_value() to properly handle {"id": X} format by looking up the object via get_related_object_by_attrs().

Fixes issue where PATCH requests to update Site with tenant field using {"id": X} format would fail with validation errors.

### DIFF
--- a/netbox/netbox/api/serializers/base.py
+++ b/netbox/netbox/api/serializers/base.py
@@ -41,6 +41,19 @@ class BaseModelSerializer(serializers.ModelSerializer):
 
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


casibbald:
Seems we were assuming some fields can be empty

Issue #20474, opened 2025-10-02
Cannot install module with placeholder values in a module bay tree 2 in tree but 1 placeholders given.


Stingo77:
### NetBox Edition

NetBox Community

### NetBox Version

v.4.4.2

### Python Version

3.11

### Steps to Reproduce

1. Create device with module bay Gi0/0/[1-24] position [1-24] for sfp module with intarface {module} and one module bay "Rear card slot" Position 1 for rear card.
2. Create rear card  with module bay Gi0/1/[1-4] position [1-4] for sfp module.
3. Insert rear card into device.
4. Insert sfp module in slot rear card.

### Expected Behavior

The module will be inserted into the raer card and  interface will be added.
Gi0/0/1		—	—	—	—	Gi0/0/1	 
Gi0/0/2		—	—	—	—	Gi0/0/2	 
Gi0/0/3		—	—	—	—	Gi0/0/3
Rear card slot	—	Generic-card-sfp		—	1	 
 Gi0/1/1	—	—	—	—	Gi0/1/1	 
 Gi0/1/2	—	—	—	—	Gi0/1/2

### Observed Behavior

The module is not added, an error is returned "Cannot install module with placeholder values in a module bay tree 2 in tree but 1 placeholders given."

arthanson:
@Stingo77 can you please expand on these steps (more details) esp step 1 "Create device with module bay Gi0/0/[1-24] position [1-24] for sfp module with intarface {module} and one module bay "Rear card slot" Position 1 for rear card." is not clear, but all the steps could use further detail.

Stingo77:
> [@Stingo77](https://github.com/Stingo77) can you please expand on these steps (more details) esp step 1 "Create device with module bay Gi0/0/[1-24] position [1-24] for sfp module with intarface {module} and one module bay "Rear card slot" Position 1 for rear card." is not clear, but all the steps could use further detail.

<img width="1134" height="860" alt="Image" src="https://github.com/user-attachments/assets/56511381-e604-40bf-8299-6de5b4799891" />
I've created a sw-test device with module bay "Rear card slot" (1) and module bays Gi/0/1- Gi0/0/8.

<img width="739" height="453" alt="Image" src="https://github.com/user-attachments/assets/090e865c-5b32-4b79-986e-4e3985a26558" />
I've created a module Generic-card-sfp with module bays Gi/1/1- Gi0/1/4. Installed the module into the device sw-test.

[...]

Issue #20467, opened 2025-10-02
The "Position" field doesn't resolve the {module} variable in nested modules


tpx-mk:
### NetBox Edition

NetBox Community

### NetBox Version

4.3.7

### Python Version

3.10

### Steps to Reproduce

- Create a module type, e.g. _TapeLib Extension module_
- Within the module type create one (or more) module bay(s) with the {module} variable in the _position_ field 
- Create another module type, e.g. _TapeLib Drive module_
- Add a module of the first type into some device (e.g. TapeLibrary)
- Add a module of the second type into a module bay of the first module

### Expected Behavior

the {module} variable gets expended not only in the names of the modules and interfaces and the _label_ field but also in the _position_ field

### Observed Behavior

the position in the nested moduled shows _{module}_ as text without resolving the variable

mrmrcoleman:
Hey @tpx-mk I've now reproduced this and it seems like the position field was never intended to be able to use the `{module}` approach. Could you share a bit more about your workflow/use case for relying on the `position` field instead of the `name` field please?

Separately the resolution of [this issue](https://github.com/netbox-community/netbox/issues/20474) may allow for the correct resolution of `{module}` in in the `name` field for any depth of module nesting which could be an alernative solution to your problem.

tpx-mk:
Hey @mrmrcoleman,

happy new year... ;-)

My use case:
I have a TapeLibrary that could be extended by modules to fit in more tapes. As well as the base unit each extension module also allows to plug in up to 2 TapeDrives for writing/reading the tapes. So ideally I'd like to get a position on a TapeDrive like 1/1 (means module 1, drive 1) or 1/2 or 3/1, etc.

I hope this helps for getting my idea.

Thanks a lot

mrmrcoleman:
Thanks @tpx-mk couple questions:

1. If we could make this work with the `name` field instead of the `position` field, would that be sufficient?
2. Please take a look at the worked examples below and tell me if any of them do not match your expected behaviour:

If I understand you correctly, the following would be true:

### Example 1 — Single module, single drive
**GIVEN**
- TapeLibrary `TL-01`
- Extension module 1
- One TapeDrive in drive bay 1

**THEN**
- TapeDrive position: `1/1`

---

### Example 2 — Single module, two drives
**GIVEN**
- TapeLibrary `TL-01`
- Extension module 1
- TapeDrives in drive bays 1 and 2

**THEN**
- Drive in bay 1: `1/1`  
[...]

Issue #19796, opened 2025-06-30
Extend the behaviour of the {module} variable in components of module types to account for different module bay layouts


sol1-matt:
### NetBox version

v4.2.8

### Feature type

Change to existing functionality

### Proposed functionality

This is a FR related to #19414 to fulfil the request by [arthanson](https://github.com/netbox-community/netbox/issues/19414#issuecomment-2920128512) . 

The proposed functionality would allow for the definition of multiple naming conventions using `{module}` such that the same module type can be used with devices with in different levels of module bay configurations. 

The implementation of this could be managed in different ways with the desired outcome to be the creation of values like 
```
1 level:  port-1
2 levels: port-1-1
3 levels: port-1-1-1
```
#### Possible Implementations 
**List all possible combinations**
```
port-{module},port-{module}-{module},port-{module}-{module}-{module}

```

**Assume stop once all modules values are consumed**
```
port-{module}-{module}-{module}
```

**Regex optional repeat**
```
port-{module}(-{module}){0,2}
```

### Use case

The current functionality of module bays and module types correctly allows for nested designs

eg: device -> module bays -> module type -> module bays -> module type -> interface

A practical example of this would be a Server with a OCP bay (device -> module bay) which is filled with a dual port SFP nic (module type -> module bays), each port is then filled with a SFP module (module type -> interface).

For this to work in Netbox the SFP module needs to have the interface name set with 2 `{module}` variables 
eg: `port-{module}-{module}`

However the same SFP module can not be used in another device which doesn't have this nested structure.

eg: device -> module bays -> module type -> interface

A practical example of this would be a Switch with SFP ports (module type -> module bays), each port is then filled with a SFP module (module type -> interface).

For this to work in Netbox the SFP module needs to have the interface name set with 1 `{module}` variable 
eg: `port-{module}`

The Server and the Switch require different variables for the SFP Module which currently isn't possible. 

The error message seen when trying to add a module type with the incorrect number of `{module}` vars for a given device configuration is
```Cannot install module with placeholder values in a module bay tree <x> in tree but <y> placeholders given.```

#### Workarounds
**Duplicate Module Types**
A work around to this would be to have multiple SFP Modules of the same type with different nesting vars for each one. 

eg: 
SFP Module (nest 1) = `port-{module}`
SFP Module (nest 2) = `port-{module}-{module}`
SFP Module (nest 3) = `port-{module}-{module}-{module}`

[...]

PR #21073, opened 2026-01-05
Fix {module} placeholder resolution + add {module_path} for nested modules

mrmrcoleman:
## Summary

Fixes the `{module}` placeholder resolution for nested module bays and introduces a new `{module_path}` placeholder for automatic full-path expansion.

**Fixes:** #20474, #20467, #19796

## Problem

The current implementation requires an **exact match** between the number of `{module}` tokens in template names and the depth of the module bay tree. This forces users to create duplicate ModuleTypes for each nesting level, breaking the "pluggable transceivers" use case where a single SFP/QSFP module should install at any depth.

Additionally, the `{module}` placeholder was not resolved in the `position` field of module bays, leaving literal `{module}` strings in positions when installing nested modules.

## Solution

This PR introduces **two distinct placeholders**:

### `{module_path}` (NEW)
- Expands to the **full path** from root to current position, joined by `/`
- Example: At depth 3 with positions "1", "2", "3" → expands to "1/2/3"
- **Use case:** Generic modules (SFPs, QSFPs) that work at any nesting depth without modification

### `{module}` (UPDATED BEHAVIOR)
- **Single token:** Expands to the **immediate parent's position only** (not full path)
- **Multiple tokens:** Expands **level-by-level** (unchanged for backwards compatibility)
- **Use case:** Building custom paths via position field with user-controlled separators

### Why Two Placeholders?

This design supports two complementary workflows:

1. **Simple case:** Use `{module_path}` in component names for automatic path joining
2. **Custom separators:** Use `{module}` in position field to build paths like `eth{module}-{position}` where users control the separator in the position field itself

### Validation Rules
- Only one `{module_path}` allowed per template attribute
- Cannot mix `{module}` and `{module_path}` in the same attribute
- Multiple `{module}` tokens must match tree depth exactly (existing behavior)

## Testing

- All **8,375 tests pass** locally
- Added comprehensive test coverage for:
  - `{module_path}` at depths 1, 2, and 3
  - Single `{module}` parent-only behavior
  - Position field resolution with `{module}` and `{module_path}`
  - Validation rules for both placeholders
  - Edge case: nested position resolution without duplication (sigprof's scenario)

## Why This Targets `feature` Branch

This PR targets `feature` because it introduces new functionality (`{module_path}` placeholder). However, **all changes are backwards compatible**:

- Existing single `{module}` usage now gives parent position (previously gave full path for depth-1 only)
- Multiple `{module}` behavior is unchanged
- No existing workflows break

**To target `main` for 4.5.x:** The behavioral change for single `{module}` (parent-only vs. full-path) is technically a change, but since the previous behavior was problematic for nested modules anyway, this could be considered a bug fix. Maintainers can decide if this warrants backporting.

## Changes

- **`dcim/constants.py`:** Added `MODULE_PATH_TOKEN`, `resolve_module_placeholders()` helper
- **`dcim/forms/common.py`:** Updated validation for both placeholder types
- **`dcim/models/device_component_templates.py`:** Uses centralized helper for name/label/position resolution
- **`dcim/tests/test_models.py`:** Comprehensive test coverage for all scenarios

### DIFF
--- a/docs/models/dcim/moduletype.md
+++ b/docs/models/dcim/moduletype.md
@@ -16,9 +16,33 @@ Note that device bays and module bays may _not_ be added to modules.
 
 ## Automatic Component Renaming
 
-When adding component templates to a module type, the string `{module}` can be used to reference the `position` field of the module bay into which an instance of the module type is being installed.
+When adding component templates to a module type, placeholders can be used to dynamically incorporate the module bay's `position` field into component names. Two placeholders are available:
 
-For example, you can create a module type with interface templates named `Gi{module}/0/[1-48]`. When a new module of this type is "installed" to a module bay with a position of "3", NetBox will automatically name these interfaces `Gi3/0/[1-48]`.
+### `{module}` Placeholder
+
+The `{module}` placeholder references the position of the parent module bay:
+
+* **Single use**: Expands to the immediate parent's position only
+* **Multiple uses**: Each `{module}` token is replaced level-by-level (the number of tokens must match the nesting depth)
+
+For example, a module type with interface templates named `Gi{module}/0/[1-48]`, when installed in a module bay with position "3", will create interfaces named `Gi3/0/[1-48]`.
+
+### `{module_path}` Placeholder
+
+The `{module_path}` placeholder expands to the full path from the root device to the current module, with positions joined by `/`. This is useful for modules that can be installed at any nesting depth without modification.
+
+For example, consider an SFP module type with an interface template named `eth{module_path}`:
+
+* Installed directly in slot 2: creates interface `eth2`
+* Installed in slot 1's nested bay 1: creates interface `eth1/1`
+* Installed in slot 1's nested bay 2's sub-bay 3: creates interface `eth1/2/3`
+
+!!! note
+    `{module_path}` can only be used once per template attribute, and cannot be mixed with `{module}` in the same attribute.
+
+### Position Field Resolution
+
+The `{module}` placeholder can also be used in the `position` field of [module bay templates](./modulebaytemplate.md) defined on a module type. This allows nested module bays to build hierarchical position values. For example, a module bay template with `position="{module}/1"`, when its parent module is installed in a bay with position "2", will have its position resolved to "2/1".
 
 Automatic renaming is supported for all modular component types (those listed above).
 
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -126,18 +126,26 @@ class ModuleCommonForm(forms.Form):
                             _("Cannot install module with placeholder values in a module bay with no position defined.")
                         )
 
-                    if len(module_bays) != template.name.count(MODULE_TOKEN):
+                    token_count = template.name.count(MODULE_TOKEN)
+                    # Validate: depth must be >= token_count (can't expand tokens without context)
+                    if len(module_bays) < token_count:
                         raise forms.ValidationError(
                             _(
                                 "Cannot install module with placeholder values in a module bay tree {level} in tree "
                                 "but {tokens} placeholders given."
                             ).format(
-                                level=len(module_bays), tokens=template.name.count(MODULE_TOKEN)
+                                level=len(module_bays), tokens=token_count
                             )
                         )
 
-                    for module_bay in module_bays:
-                        resolved_name = resolved_name.replace(MODULE_TOKEN, module_bay.position, 1)
+                    if token_count == 1:
+                        # Single token: substitute with full path (e.g., "1/1" for depth 2)
+                        full_path = '/'.join([mb.position for mb in module_bays])
+                        resolved_name = resolved_name.replace(MODULE_TOKEN, full_path, 1)
+                    else:
+                        # Multiple tokens: substitute level-by-level (existing behavior)
+                        for mb in module_bays:
+                            resolved_name = resolved_name.replace(MODULE_TOKEN, mb.position, 1)
 
                 existing_item = installed_components.get(resolved_name)
 
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -130,6 +130,12 @@ class ModuleCommonForm(forms.Form):
                             _("Cannot install module with placeholder values in a module bay with no position defined.")
                         )
 
+                    # Cannot mix {module} and {module_path} in the same attribute
+                    if has_module_token and has_module_path_token:
+                        raise forms.ValidationError(
+                            _("Cannot mix {module} and {module_path} placeholders in the same template attribute.")
+                        )
+
                     # Validate {module_path} - can only appear once
                     if has_module_path_token:
                         path_token_count = template.name.count(MODULE_PATH_TOKEN)
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -3,6 +3,7 @@ from django.utils.translation import gettext_lazy as _
 
 from dcim.choices import *
 from dcim.constants import *
+from dcim.utils import resolve_module_placeholders
 from utilities.forms import get_field_value
 
 __all__ = (
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -139,14 +139,9 @@ class ModuleCommonForm(forms.Form):
                             )
                         )
 
-                    if token_count == 1:
-                        # Single token: substitute with full path (e.g., "1/1" for depth 2)
-                        full_path = '/'.join([mb.position for mb in module_bays])
-                        resolved_name = resolved_name.replace(MODULE_TOKEN, full_path, 1)
-                    else:
-                        # Multiple tokens: substitute level-by-level (existing behavior)
-                        for mb in module_bays:
-                            resolved_name = resolved_name.replace(MODULE_TOKEN, mb.position, 1)
+                    # Use centralized helper for token substitution
+                    positions = [mb.position for mb in module_bays]
+                    resolved_name = resolve_module_token(resolved_name, positions)
 
                 existing_item = installed_components.get(resolved_name)
 
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -127,8 +127,9 @@ class ModuleCommonForm(forms.Form):
                         )
 
                     token_count = template.name.count(MODULE_TOKEN)
-                    # Validate: depth must be >= token_count (can't expand tokens without context)
-                    if len(module_bays) < token_count:
+                    # A single token which gets expanded to the full path is always
+                    # allowed; otherwise the number of tokens needs to match the path length.
+                    if token_count != 1 and token_count != len(module_bays):
                         raise forms.ValidationError(
                             _(
                                 "Cannot install module with placeholder values in a module bay tree {level} in tree "
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -119,29 +119,41 @@ class ModuleCommonForm(forms.Form):
             # Get the templates for the module type.
             for template in getattr(module_type, templates).all():
                 resolved_name = template.name
+                has_module_token = MODULE_TOKEN in template.name
+                has_module_path_token = MODULE_PATH_TOKEN in template.name
+
                 # Installing modules with placeholders require that the bay has a position value
-                if MODULE_TOKEN in template.name:
+                if has_module_token or has_module_path_token:
                     if not module_bay.position:
                         raise forms.ValidationError(
                             _("Cannot install module with placeholder values in a module bay with no position defined.")
                         )
 
-                    token_count = template.name.count(MODULE_TOKEN)
-                    # A single token which gets expanded to the full path is always
-                    # allowed; otherwise the number of tokens needs to match the path length.
-                    if token_count != 1 and token_count != len(module_bays):
-                        raise forms.ValidationError(
-                            _(
-                                "Cannot install module with placeholder values in a module bay tree {level} in tree "
-                                "but {tokens} placeholders given."
-                            ).format(
-                                level=len(module_bays), tokens=token_count
+                    # Validate {module_path} - can only appear once
+                    if has_module_path_token:
+                        path_token_count = template.name.count(MODULE_PATH_TOKEN)
+                        if path_token_count > 1:
+                            raise forms.ValidationError(
+                                _("The {module_path} placeholder can only be used once per template.")
                             )
-                        )
 
-                    # Use centralized helper for token substitution
+                    # Validate {module} - multi-token must match depth exactly
+                    if has_module_token:
+                        token_count = template.name.count(MODULE_TOKEN)
+                        # Multiple {module} tokens must match the tree depth exactly
+                        if token_count > 1 and token_count != len(module_bays):
+                            raise forms.ValidationError(
+                                _(
+                                    "Cannot install module with placeholder values in a module bay tree {level} deep "
+                                    "but {tokens} placeholders given."
+                                ).format(
+                                    level=len(module_bays), tokens=token_count
+                                )
+                            )
+
+                    # Use centralized helper for placeholder substitution
                     positions = [mb.position for mb in module_bays]
-                    resolved_name = resolve_module_token(resolved_name, positions)
+                    resolved_name = resolve_module_placeholders(resolved_name, positions)
 
                 existing_item = installed_components.get(resolved_name)
 
--- a/netbox/dcim/models/device_component_templates.py
+++ b/netbox/dcim/models/device_component_templates.py
@@ -175,9 +175,16 @@ class ModularComponentTemplateModel(ComponentTemplateModel):
 
         if module:
             modules = self._get_module_tree(module)
+            token_count = self.name.count(MODULE_TOKEN)
             name = self.name
-            for module in modules:
-                name = name.replace(MODULE_TOKEN, module.module_bay.position, 1)
+            if token_count == 1:
+                # Single token: substitute with full path (e.g., "1/1" for depth 2)
+                full_path = '/'.join([m.module_bay.position for m in modules])
+                name = name.replace(MODULE_TOKEN, full_path, 1)
+            else:
+                # Multiple tokens: substitute level-by-level (existing behavior)
+                for m in modules:
+                    name = name.replace(MODULE_TOKEN, m.module_bay.position, 1)
             return name
         return self.name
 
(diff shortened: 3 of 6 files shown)


netbox/dcim/forms/common.py
```diff
@@ -126,18 +126,26 @@ def clean(self):
                             _("Cannot install module with placeholder values in a module bay with no position defined.")
                         )
 
-                    if len(module_bays) != template.name.count(MODULE_TOKEN):
+                    token_count = template.name.count(MODULE_TOKEN)
+                    # Validate: depth must be >= token_count (can't expand tokens without context)
+                    if len(module_bays) < token_count:
```
sigprof: This also enables a questionable case where `token_count` is above 1, but less than the tree depth; in this case some `position` values would be lost.  Maybe having a more strict check would be better:
```suggestion
                    token_count = template.name.count(MODULE_TOKEN)
                    # A single token which gets expanded to the full path is always allowed; otherwise the number of tokens needs to match the path length.
                    if token_count != 1 and token_count != len(module_bays):
```

netbox/dcim/forms/common.py
```diff
@@ -126,18 +126,26 @@ def clean(self):
...
-                                level=len(module_bays), tokens=template.name.count(MODULE_TOKEN)
+                                level=len(module_bays), tokens=token_count
                             )
                         )
 
-                    for module_bay in module_bays:
-                        resolved_name = resolved_name.replace(MODULE_TOKEN, module_bay.position, 1)
+                    if token_count == 1:
+                        # Single token: substitute with full path (e.g., "1/1" for depth 2)
+                        full_path = '/'.join([mb.position for mb in module_bays])
```
sigprof: The `/` separator between the module tree levels is hardcoded.  This could be enough for most current switches which use the `<UNIT>/<SLOT>/<PORT>` scheme for interface names, but some older hardware could use different separators, and there are also cases like channelized interfaces for which some vendors use other separator characters like `:`.

netbox/dcim/models/device_component_templates.py
```diff
@@ -175,9 +175,16 @@ def resolve_name(self, module):
...
-                name = name.replace(MODULE_TOKEN, module.module_bay.position, 1)
+            if token_count == 1:
+                # Single token: substitute with full path (e.g., "1/1" for depth 2)
+                full_path = '/'.join([m.module_bay.position for m in modules])
+                name = name.replace(MODULE_TOKEN, full_path, 1)
+            else:
+                # Multiple tokens: substitute level-by-level (existing behavior)
+                for m in modules:
+                    name = name.replace(MODULE_TOKEN, m.module_bay.position, 1)
             return name
```
sigprof: This almost duplicates the code in `ModuleCommonForm.clean()`, except for the `_get_module_bay_tree`/`_get_module_tree` difference.  Maybe it could be possible to refactor it, so that issues like the module path separator may be handled in a single place?

netbox/dcim/models/device_component_templates.py
```diff
@@ -187,9 +194,16 @@ def resolve_label(self, module):
...
-                label = label.replace(MODULE_TOKEN, module.module_bay.position, 1)
+            if token_count == 1:
+                # Single token: substitute with full path (e.g., "1/1" for depth 2)
+                full_path = '/'.join([m.module_bay.position for m in modules])
+                label = label.replace(MODULE_TOKEN, full_path, 1)
+            else:
+                # Multiple tokens: substitute level-by-level (existing behavior)
+                for m in modules:
+                    label = label.replace(MODULE_TOKEN, m.module_bay.position, 1)
             return label
```
sigprof: And this is also almost a straight duplicate of the code in `resolve_name()`, just for a different attribute.

sigprof (commented):
Issues to discuss:
1. Should we redefine the meaning of `{module}` (so that a single use of `{module}` means the whole path, but multiple uses of `{module}` in the same attribute each expand to a single level of `position`), or define a new placeholder like `{module_path}` which must be used only once in the string (and then keep the existing restrictions for `{module}`)?  The latter may be better for backwards compatibility, but some people would need to replace `{module}` with `{module_path}` in their module types when they encounter devices with multilevel module hierarchy.
2. How should the separators in the `{module_path}` expansion be handled?  In general, the separator may need to be specified by the device (and in some cases multiple kinds of separators may be needed, like `<UNIT>/<SLOT>/<PORT>:<CHANNEL>`, although in this case the `:<CHANNEL>` part may need to be supplied by the transceiver module).
3. More generally, should the `position` attribute for a module bay contain only the parent-relative position, or the full path from the device?  The current code assumes the first interpretation (therefore expanding `{module_path}` requires concatenating the `position` values along the path and adding separators); however, if we choose the second interpretation (which would also require fixing #20467, so that the `position` attribute for the module bay on a module could reference the position of the parent module bay), we could avoid at least some issues with vendor-specific separators: the modules which correspond to interface cards (which are inevitably vendor-specific) could include all separators in their `position` values, so that the module bays which accept generic transceiver modules could have their `position` values always equal to the interface names.

netbox/dcim/forms/common.py
mrmrcoleman: Please updated note on PR

sigprof:
The suggested fixes for #20474/#19796 and #20467 are somewhat conflicting:
- for #20474/#19796 the intended way seems to be “use a single `{module}` placeholder, which automatically gets expanded to the `/`-separated path in case of nested modules”;
- but for #20467 a way to refer to just the position of the parent module would be needed, which does not seem to be present in the suggested changes.

E.g., look at the intended usage in #20467:
- device
  - module bay (position=`2`)
    - extension module (name=`Extension module {module}`, expanded to `Extension module 2`)
      - module bay (position=`{module}/1`, expanded to `2/1`)
        - tape drive module with name = ?:
          - name=`{module}` would expand to `2/2/1`, duplicating the position which got inserted into the position of the nested module bay :-1:
          - name=`{module}/{module}` would expand to `2/2/1` again :-1:

Possible options here:
1. Decide that position=`{module}/something` is wrong, and a single `{module}` expanding to a `/`-separated path is the only supported way.  This would mean rejecting #20467 as WONTFIX (the intended hierarchy would still be implementable as the way 1 below, just using `{module}` instead of `{module_path}`).
2. Introduce separate placeholders to make two ways of building the module hierarchy possible:
   - `{module}` — if used only once in the value, expands to `position` of just the parent module bay even if there are more nesting levels, otherwise the number of `{module}` placeholders needs to be equal to the number of nesting levels;
   - `{module_path}` — expands to the `/`-separated path of all `position` values up to the top level of the module tree.
3. Decide the position=`{module}/something` is the only correct way and change `{module}` to expand to `position` of just the parent module bay if used only once in the value.

In the first case modules which can be used at any depth would have the `/` separator hardcoded (the alternative of specifying multiple `{module}` placeholders with custom separators would lock the module to the specific number of hierarchy levels).

In the second case there would be two ways to build the hierarchy:
1. Set the `position` values for nested module bays to contain only the position on the particular level (do not use the `{module}` placeholder in `position`), then use `{module_path}` in generic modules.  This hardcodes the `/` separators again.
Example:
	- device
	  - module bay (position=`1/1`) *tape drive slot on the base unit*
	    - tape drive module with name = `Tape drive {module_path}`, expanded to `Tape drive 1/1`
	  - module bay (position=`2`) *extension module slot*
	    - extension module (name=`Extension module {module}`, expanded to `Extension module 2`)
	      - module bay (position=`1`) *tape drive slot on the extension unit*
	        - tape drive module with name = `Tape drive {module_path}`, expanded to `Tape drive 2/1`
	  - module bay (position=`3`) *extension module slot*
	    - extension module (name=`Extension module {module}`, expanded to `Extension module 3`)
	      - module bay (position=`1`) *tape drive slot on the extension unit*
	        - tape drive module with name = `Tape drive {module_path}`, expanded to `Tape drive 3/1`
2. Set the `position` values for nested module bays to contain the whole path from the root device (use the `{module}` placeholder in `position` to get the path to the containing module and append the relative position of the bay), then use `{module}` in generic modules.  This does not hardcode the `/` separators (each module bay can specify a different separator if needed).
Example:
	- device
	  - module bay (position=`1/1`) *tape drive slot on the base unit*
	    - tape drive module with name = `Tape drive {module}`, expanded to `Tape drive 1/1`
	  - module bay (position=`2`) *extension module slot*
	    - extension module (name=`Extension module {module}`, expanded to `Extension module 2`)
	      - module bay (position=`{module}/1`, expanded to `2/1`) *tape drive slot on the extension unit*
	        - tape drive module with name = `Tape drive {module}`, expanded to `Tape drive 2/1`
	  - module bay (position=`3`) *extension module slot*
	    - extension module (name=`Extension module {module}`, expanded to `Extension module 3`)
	      - module bay (position=`{module}/1`, expanded to `3/1`) *tape drive slot on the extension unit*
	        - tape drive module with name = `Tape drive {module}`, expanded to `Tape drive 3/1`

One problem here is that the choice between these two ways to build the hierarchy needs to be made at the start — if you already have some module types set up for one way, you can't reuse those types without changes for the other way.

In the third case only the way 2 above would be possible if you want to build a multilevel hiearchy with generic modules that accept any number of nested levels.

mrmrcoleman:
@sigprof Your input is appreciated. Please take a look at the recent changes and updated PR description

netbox/dcim/forms/common.py
```diff
@@ -119,25 +120,41 @@ def clean(self):
...
+                        token_count = template.name.count(MODULE_TOKEN)
+                        # Multiple {module} tokens must match the tree depth exactly
+                        if token_count > 1 and token_count != len(module_bays):
+                            raise forms.ValidationError(
+                                _(
+                                    "Cannot install module with placeholder values in a module bay tree {level} deep "
+                                    "but {tokens} placeholders given."
+                                ).format(
+                                    level=len(module_bays), tokens=token_count
+                                )
```
sigprof, line 158: The description says that `{module}` and `{module_path}` cannot be mixed in the same attribute, but the validation code to reject that case seems to be missing.

netbox/dcim/models/device_component_templates.py
```diff
@@ -170,29 +171,33 @@ def _get_module_tree(self, module):
...
+        This is used by ModuleBayTemplate to resolve positions like "{module}/1"
+        to actual values like "A/1" when the parent module is installed in bay "A".
+
+        Fixes Issue #20467.
+        """
+        if module:
+            positions = [m.module_bay.position for m in self._get_module_tree(module)]
+            return resolve_module_placeholders(position, positions)
+        return position
+
```
sigprof: Why does this method receive the `position` value as argument, unlike `resolve_name()` and `resolve_label()`?  Looks like it should be called `resolve_module_placeholders()` and then used to implement `resolve_name()` and `resolve_label()` too:
```python
    def resolve_name(self, module):
        return self.resolve_module_placeholders(self.name, module)
    def resolve_label(self, module):
        return self.resolve_module_placeholders(self.label, module)
    def resolve_position(self, module):
        return self.resolve_module_placeholders(self.position, module)
```

netbox/dcim/models/device_component_templates.py
```diff
@@ -722,10 +727,11 @@ class Meta(ModularComponentTemplateModel.Meta):
...
 
     def instantiate(self, **kwargs):
+        module = kwargs.get('module')
         return self.component_model(
-            name=self.resolve_name(kwargs.get('module')),
-            label=self.resolve_label(kwargs.get('module')),
-            position=self.position,
+            name=self.resolve_name(module),
+            label=self.resolve_label(module),
+            position=self.resolve_position(self.position, module),
```
sigprof: Here that `self.position` looks out of place (see the comment above).

mrmrcoleman:
@sigprof Thanks for your review. Your feedback is addressed in https://github.com/netbox-community/netbox/pull/21073/commits/898fe8b3d8dc593a9fe2c2492215a8eb3e274e24

arthanson (changes requested):
Can you add documentation for {module_path} to docs/models/dcim/moduletype.md under ## Automatic Component Renaming

netbox/dcim/tests/test_models.py
```diff
@@ -875,6 +875,1231 @@ def test_module_bay_parent_cleared_when_module_removed(self):
...
+            'replicate_components': True,
+            'adopt_components': False,
+        })
+
+        self.assertFalse(form.is_valid())
+
+    def test_mixed_module_and_module_path_fails_validation(self):
+        """
+        Test that mixing {module} and {module_path} in the same template attribute
+        fails validation. Per sigprof's feedback, these cannot be combined.
```
arthanson, line 2051: can probably get rid of the "Per sigprof's feedback..."

netbox/dcim/tests/test_models.py
```diff
@@ -875,6 +875,1231 @@ def test_module_bay_parent_cleared_when_module_removed(self):
...
+        Test that {module_path} can only appear once in a template.
+        Using it multiple times should fail validation.
+        """
+        from dcim.forms import ModuleForm
+
+        manufacturer = Manufacturer.objects.first()
+        site = Site.objects.first()
+        device_role = DeviceRole.objects.first()
+
+        device_type = DeviceType.objects.create(
```
arthanson, line 2006: would it make sense to move device_type creation into setup, it seems to be used in almost every test.

netbox/dcim/tests/test_models.py
```diff
@@ -875,6 +875,1231 @@ def test_module_bay_parent_cleared_when_module_removed(self):
...
+        line_card = Module.objects.create(device=device, module_bay=bay1, module_type=line_card_type)
+
+        bay2 = line_card.modulebays.get(name='Nested Bay')
+        leaf = Module.objects.create(device=device, module_bay=bay2, module_type=leaf_type)
+
+        interface = leaf.interfaces.first()
+        # Single {module} should give PARENT position only: Y (not X/Y)
+        self.assertEqual(interface.name, 'Port Y')
+
+    def test_sigprof_nested_position_no_duplication(self):
```
arthanson, line 1925: I think the tests are good but a few things to check:
1. There looks like some duplication, like checking for depth 1, then 2, then 3, I think just 1 and 3 would work potentially. Also test_nested_module_bay_label_resolution looks like it could be folded into one of the new tests.
2. I'd re-order these to make them easier to check in the future - put all the {module} tests first, then the {module_path} test next, then the combined ones after.  keep the simple 'test single {module}' first then test multi module after it, it is sort of like this at start but then the tests get jumbled together.  Move test_single_module_token down with the other {module} tests.

docs/models/dcim/moduletype.md
```diff
@@ -16,9 +16,33 @@ Note that device bays and module bays may _not_ be added to modules.
 
 ## Automatic Component Renaming
 
-When adding component templates to a module type, the string `{module}` can be used to reference the `position` field of the module bay into which an instance of the module type is being installed.
+When adding component templates to a module type, placeholders can be used to dynamically incorporate the module bay's `position` field into component names. Two placeholders are available:
```
arthanson, line 19: This is a bit confusing - can these both be used in both the module_name and position fields?  "Position Field Resolution" says only {module} can be there, but resolve_position function makes it sound like {module_name} can be their also?  Can multiple {module} be in the position field?

If not the validators should be updated for position field and tests for it as well.

netbox/dcim/forms/common.py
```diff
@@ -119,25 +120,47 @@ def clean(self):
...
-                                "but {tokens} placeholders given."
-                            ).format(
-                                level=len(module_bays), tokens=template.name.count(MODULE_TOKEN)
-                            )
+                            _("Cannot mix {module} and {module_path} placeholders in the same template attribute.")
                         )
 
-                    for module_bay in module_bays:
-                        resolved_name = resolved_name.replace(MODULE_TOKEN, module_bay.position, 1)
+                    # Validate {module_path} - can only appear once
```
arthanson, line 139: maybe move the token validator to a separate private function to make clearer.

netbox/dcim/models/device_component_templates.py
```diff
@@ -170,27 +171,17 @@ def _get_module_tree(self, module):
...
-            return self.name
-
+        """Resolve {module} and {module_path} placeholders in component name."""
         if module:
-            modules = self._get_module_tree(module)
-            name = self.name
-            for module in modules:
-                name = name.replace(MODULE_TOKEN, module.module_bay.position, 1)
-            return name
+            positions = [m.module_bay.position for m in self._get_module_tree(module)]
```
arthanson, line 176: this has common code with resolve_label, can just make a private function - something like:

```
    def _resolve_placeholders(self, text, module):
        """Resolve {module} and {module_path} placeholders in text (module must be set)."""
        positions = [m.module_bay.position for m in self._get_module_tree(module)]
        return resolve_module_placeholders(text, positions)

    def resolve_name(self, module):
        if not module:
            return self.name
        return self._resolve_placeholders(self.name, module)

    def resolve_label(self, module):
        if not module:
            return self.label
        return self._resolve_placeholders(self.label, module)

```

netbox/dcim/models/device_component_templates.py
```diff
@@ -721,11 +712,26 @@ class Meta(ModularComponentTemplateModel.Meta):
...
+        """
+        Resolve {module} and {module_path} placeholders in position field.
+
+        This allows positions like "{module}/1" to resolve to "A/1" when
+        the parent module is installed in bay "A".
+
+        Fixes Issue #20467.
+        """
+        if module:
+            positions = [m.module_bay.position for m in self._get_module_tree(module)]
```
arthanson, line 725: actually this seems duplicated in the my suggestion above - can the position and resolve_module_placeholders just be made into a common function. Currently duplicated three times.

arthanson (changes requested):

marcinpsk:
> 2. How should the separators in the `{module_path}` expansion be handled?  In general, the separator may need to be specified by the device (and in some cases multiple kinds of separators may be needed, like `<UNIT>/<SLOT>/<PORT>:<CHANNEL>`, although in this case the `:<CHANNEL>` part may need to be supplied by the transceiver module).

Maybe something like below could be possible? but that would mean adding yet another field. And there's added complexity because numbering starts from 0 on Juniper, but Cisco I think it's 1, so is Cumulus I think, so yet another 'offset' field. Not sure how else to model this.

Device/line card YAML — Juniper:
  module-bays:
  - name: Xcvr 3
    position: '3'
    name_format: '{module_path}:{name}'    # Juniper uses ':'

  Device/line card YAML — Cisco:
  module-bays:
  - name: Port 8
    position: '8'
    name_format: '{module_path}/{name}'    # Cisco uses '/'

  Device/line card YAML — Cumulus:
  module-bays:
  - name: Port 1
    position: '1'
    name_format: '{module_path}s{name}'    # Cumulus uses 's'

  Generic breakout module:
  model: QSFP-4X10G-SR
  interfaces:
  - name: '0'
    type: 10gbase-x-sfpp
  - name: '1'
    type: 10gbase-x-sfpp
  - name: '2'
    type: 10gbase-x-sfpp
  - name: '3'
    type: 10gbase-x-sfpp

  Non-breakout bays (no name_format — so no change):
  module-bays:
  - name: Port 3
    position: 'et-0/0/3'
no name_format → module uses {module_path} - at least when I was trying to check some devices on this PR branch, that's what I was always using.

mrmrcoleman:
Superseded by #21436, which is a cleaner rewrite rebased onto `main`.

Key differences from this PR:
- `{module}` behavior is **completely unchanged** (no more "parent-only" semantic shift)
- Only adds `{module_path}` as a new placeholder
- Does **not** address #20467 (position field resolution) — that's a separate issue
- Targets `main` instead of `feature`

PR #21092, opened 2026-01-08
Introduce a new issue type for feature removals

jeremystretch:
- Introduce a new issue type & label specifically for the removal of a feature
- Tweak the existing deprecation issue type to limit its scope to deprecation only

This change is being made to clearly delineate between the _deprecation_ of a feature and its actual removal from the application, which typically happens at a later point in time.

### DIFF
--- a/.github/ISSUE_TEMPLATE/07-feature_removal.yaml
+++ b/.github/ISSUE_TEMPLATE/07-feature_removal.yaml
@@ -1,0 +1,20 @@
+---
+name: 🗑️ Feature Removal
+type: Deprecation
+description: The removal of a deprecated feature or resource
+labels: ["netbox", "type: removal"]
+body:
+  - type: input
+    attributes:
+      label: Deprecation Issue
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/07-feature_removal.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: 🗑️ Feature Removal
+type: Deprecation
```
jnovinger:
```suggestion
type: Removal
```

jnovinger (changes requested):

.github/ISSUE_TEMPLATE/07-feature_removal.yaml
jeremystretch: Thanks. I've also created the "Removal" issue type in GitHub.

jnovinger (approved):

PR #21101, opened 2026-01-09
20911 Fix sorting in dropdown

### LINKED ISSUE #20911: change 'module bay' sorting to be alphabetical when installing module
### NetBox version

v4.4.7

### Feature type

### Proposed functionality

Currently when installing a module, the sorting in the 'module bay' dropdown appears to be by module id. 
Can we get this changed to be alphabetical?

### Use case

When installing a module, let's suppose you created 3 modules called "fan1", "fan2", "fan3". Then you create 2 modules called "psu1", and "psu2".

Now let's say you did an oopsie and actually need 4 fan modules. So you go back and add 1 more fan module "fan4". 

Now when you go to install modules, the drop down selector for "module bay" has "fan4" at the bottom of the list. Rather than having fan1, fan2, fan3, fan4, psu1, psu2 ... it actually is listed as fan1, fan2, fan3, psu1, psu2, fan4.

For devices with lots of modules, it can be confusing for users.

Image

I'll add, that this seems to only be an issue if you manually create modules on a device after the device has been created from a device type.
It seems during the device creation process, modules are created in alphabetical order, so they would naturally be alphabetical in the dropdown. But if you create modules after the fact, the alphabetical sorting is disrupted.

So now you end up with instances where maybe on some devices it IS alphabetical (by nature of device id's having been created that way), and other devices where it ISN'T, causing even further confusion.

### Database changes

none

### External dependencies

none

arthanson:
### Fixes: #20911 

Two issues:
1. The sorting order wasn't alphabetical
2. Even with corrected sorting order, the selected item was show in the wrong ordering in the list (i.e. When you install a module into a Module Bay the select has a pre-set Module-bay, but when you do the dropdown - this pre-selected item shows up in the wrong order in the dropdown).

This fixes item 1, item 2 still appears but would need a front-end fix so keeping that for a separate ticket.  

As noted in the issue discussion, the change to order_insertion_by will fix the ordering of newly inserted items, there were other ordering problems in bulk edit and bulk rename which this also fixes.  I tried either looping and doing a save or doing a bulk_update and then a rebuild at the end, for small data-sets the bulk_update if much faster, for large data-sets they are almost the same perf but bulk_update still wins out.

Just changing order_insertion_by revealed an existing bug in the CSV import when using the order_insertion_by.  We don't run into it in the tests for NestedGroupModel (the other model using it) because of the way the imports are ordered in the test case as they don't cause a re-arrangement in the tree.  The ordering of the ModuleBay test actually does cause a re-ordering and hits what looks like a bug in MPTT code.  The additional changes in bulk_views.py so the CSV processing if done on models that inherit form MPTTModel uses a delay_mptt_updates() and rebuild at the end fixes this issue.


netbox/dcim/forms/model_forms.py
```diff
@@ -733,9 +733,10 @@ class ModuleForm(ModuleCommonForm, PrimaryModelForm):
     )
     module_bay = DynamicModelChoiceField(
         label=_('Module bay'),
-        queryset=ModuleBay.objects.all(),
+        queryset=ModuleBay.objects.order_by('name'),
```
jeremystretch: This change has no effect on DynamicModelChoiceFields, because the results are populated via REST API call from the frontend. (The queryset is used only to specify the related model.)

jeremystretch (changes requested):
This looks like an issue with the underlying model ordering, and shouldn't require any changes to the frontend widget (unless I'm missing something). For example, we see the ordering of the REST API endpoint is incorrect:

```
GET /api/dcim/module-bays/?device_id=27&fields=id,name

{
    "count": 6,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 29,
            "name": "fan1"
        },
        {
            "id": 30,
            "name": "fan2"
        },
        {
            "id": 31,
            "name": "fan3"
        },
        {
            "id": 32,
            "name": "psu1"
        },
        {
            "id": 33,
            "name": "psu2"
        },
        {
            "id": 34,
            "name": "fan4"
        }
    ]
}
```

The API endpoint should be returning these in the expected order, and the form widget should just populate them in sequence. This will probably require altering `order_insertion_by` on the model's `MPTTMeta` class (which I suppose should be `'name'` rather than `'module'`).

netbox/dcim/migrations/0226_modulebay_rebuild_tree.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # Set MPTTMeta with the correct order_insertion_by
+    class MPTTMeta:
+        order_insertion_by = ('module', 'name',)
+
+    ModuleBay.MPTTMeta = MPTTMeta
+    ModuleBay._mptt_meta = mptt.models.MPTTOptions(MPTTMeta)
+
+    manager = mptt.managers.TreeManager()
+    manager.model = ModuleBay
```
jnovinger, line 20: Other migrations use the `mptt.register()` helper for this. Seems like a good idea to be consistent.

jnovinger (commented):
Any concerns about any of the other bulk action views at all?

netbox/dcim/migrations/0226_modulebay_rebuild_tree.py
jnovinger: > Other migrations use the `mptt.register()` helper for this. Seems like a good idea to be consistent.

Just realized that you did use `mptt.register()` in an earlier commit. Why change it?

arthanson: Tried that, it wasn't actually respecting the order_insertion_by, it looked like the MPTTMeta wasn't correct.  This is the only way I found to make it work through trial / error.

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -526,6 +508,38 @@ def create_and_update_objects(self, form, request):
...
+        prefetched_objects = {
+            obj.pk: obj
+            for obj in self.queryset.model.objects.filter(id__in=prefetch_ids)
+        } if prefetch_ids else {}
+
+        # For MPTT models, delay tree updates until all saves are complete
+        if issubclass(self.queryset.model, MPTTModel):
+            with self.queryset.model.objects.delay_mptt_updates():
+                saved_objects = self._process_import_records(form, request, records, prefetched_objects)
+            self.queryset.model.objects.rebuild()
```
jeremystretch: The [docs](https://django-mptt.readthedocs.io/en/latest/mptt.managers.html#mptt.managers.TreeManager.delay_mptt_updates) seem to imply that calling `rebuild()` after the context manager exits is not necessary (unlike for `disable_mptt_updates()`).

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -895,9 +909,18 @@ def post(self, request):
...
-                                setattr(obj, self.field_name, obj.new_name)
-                                obj.save()
+                            # For MPTT models, delay tree updates until all saves are complete
+                            if issubclass(self.queryset.model, MPTTModel):
+                                with self.queryset.model.objects.delay_mptt_updates():
+                                    for obj in selected_objects:
+                                        setattr(obj, self.field_name, obj.new_name)
+                                        obj.save()
+
+                                self.queryset.model.objects.rebuild()
```
jeremystretch: Same as above; `rebuild()` doesn't seem to be needed.

netbox/dcim/models/modules.py
```diff
@@ -329,7 +330,7 @@ def save(self, *args, **kwargs):
                 component._location = self.device.location
                 component._rack = self.device.rack
 
-            if component_model is not ModuleBay:
+            if not issubclass(component_model, MPTTModel):
```
jeremystretch, line 337: Tangential, but since you're here could you add comments at line 333 and 352 calling out that we're handling _creates_ and _updates_ separately. I was confused for a moment reading through the existing code.

netbox/dcim/migrations/0226_modulebay_rebuild_tree.py
jeremystretch: Let's dig into this some more, because the corollary of that conclusion is that all existing MPTT rebuild migrations are potentially broken. If the below pattern doesn't work in this case, let's identify why and determine whether changes are needed for the other migrations.

```python
def rebuild_mptt(apps, schema_editor):
    manager = mptt.managers.TreeManager()
    DeviceRole = apps.get_model('dcim', 'DeviceRole')
    manager.model = DeviceRole
    mptt.register(DeviceRole)
    manager.contribute_to_class(DeviceRole, 'objects')
    manager.rebuild()
```

jeremystretch (changes requested):

netbox/dcim/migrations/0226_modulebay_rebuild_tree.py
arthanson: From what I can see it looks like this is from changing the order_insertion_by.  The normal migration above is fine in the other places we used it as it just needs to populate the tree fields (lft, rght, tree_id, level).  But this doesn't work when we change order_insertion_by.

When you call mptt.register(ModuleBay) on this historical model, it looks like it uses the previously registered ordering - so when you rebuild it ends up doing nothing.  

For changing MPTT ordering it looks like you must explicitly set the new MPTTMeta on the historical model. If we change ordering on any models in the future we will probably need to replicate this migration.

jeremystretch: I think the disparity may be that this is the first instance where `order_insertion_by` differs significantly from the model's native ordering, so it hasn't been a problem before.

jeremystretch (changes requested):
This seems to still be broken. Starting with the following list of module bays:

Image

When I add a new module bay named `fan5` via the UI, it appears at the end of the list, both in the UI and in the REST API. However, after I rebuild the tree manually via `dcim.ModuleBay.objects.rebuild()` the ordering is now correct.

netbox/dcim/models/device_components.py
```diff
@@ -1273,7 +1273,7 @@ class Meta(ModularComponentModel.Meta):
         verbose_name_plural = _('module bays')
 
     class MPTTMeta:
-        order_insertion_by = ('module',)
+        order_insertion_by = ('module', 'name',)
```
jeremystretch: I think this should just include `name`. `module` is the parent FK against which the tree is constructed.

```suggestion
        order_insertion_by = ('name',)
```

netbox/dcim/api/views.py
```diff
@@ -516,7 +516,7 @@ class RearPortViewSet(PassThroughPortMixin, NetBoxModelViewSet):
 
 
 class ModuleBayViewSet(NetBoxModelViewSet):
-    queryset = ModuleBay.objects.all()
+    queryset = ModuleBay.objects.order_by('device', 'module', 'name')
```
jeremystretch: This shouldn't be needed once the MPTT ordering is corrected.
```suggestion
    queryset = ModuleBay.objects.all()
```

netbox/dcim/forms/model_forms.py
```diff
@@ -745,7 +745,8 @@ class ModuleForm(ModuleCommonForm, PrimaryModelForm):
         label=_('Module bay'),
         queryset=ModuleBay.objects.all(),
         query_params={
-            'device_id': '$device'
+            'device_id': '$device',
+            'ordering': 'module,name',
```
jeremystretch: This doesn't seem necessary as the module bays should have this ordering applied natively.
```suggestion
```

jeremystretch (changes requested):

arthanson:
@jeremystretch Side effect of just using order_insertion_by to ('name') is the ModuleBay view is default sorted by name, where previously it was by module (as the migration needs to update the existing records to match).  Previously I had order_insertion_by ('module', 'name') which was requested to change.  Not sure if that is an issue, you can just click on 'Device' column header to see the old sorting.

**Note:** When I had order_insertion_by('module', 'name') it appeared to have the side-effect of requiring the ordering on form for the dropdown and the API, if it is just ('name') these weren't needed.

jeremystretch (approved):

PR #21106, opened 2026-01-09
Update help text for token field on API page.

### LINKED ISSUE #21105: Terminology Change for API Token help text
### NetBox version

v4.5.0

### Feature type

Other

### Proposed functionality

The help text for the API token states you need to keep a copy of the key however the terminology for the field above is Token and the term token is used throughout, with new V2 bearers there is now also a key field so this could be confusing to some users.

### Use case

Just a simple terminology change.

### Database changes

_No response_

### External dependencies

_No response_

adamboutcher:
### Fixes: #21105

The help text for the API token states you need to keep a copy of the key however the terminology above is Token, with new V2 bearers there is now also a key field so this could be confusing to some users.

### DIFF
--- a/netbox/users/forms/model_forms.py
+++ b/netbox/users/forms/model_forms.py
@@ -123,7 +123,7 @@ class UserTokenForm(forms.ModelForm):
     token = forms.CharField(
         label=_('Token'),
         help_text=_(
-            'Tokens must be at least 40 characters in length. <strong>Be sure to record your key</strong> prior to '
+            'Tokens must be at least 40 characters in length. <strong>Be sure to record your token</strong> prior to '
             'submitting this form, as it will no longer be accessible once the token has been created.'
         ),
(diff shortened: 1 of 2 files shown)


jnovinger:
Thank you for your interest in contributing to NetBox, however it appears there is no accepted issue that correlates to this pull request. Before pull requests are opened, we require an accepted issue as per our [contributing guide](https://github.com/digitalocean/netbox/blob/master/CONTRIBUTING.md). Please first open an issue and wait for it to be accepted before further work is done on this pull request.

jnovinger (changes requested):
Since we're making this change, let's also fix this in [`users/models/tokens.py:72`](https://github.com/netbox-community/netbox/blob/3624b88c3fa6cb677756823db1413081b3a6607a/netbox/users/models/tokens.py#L72).

jnovinger (approved):

Issue #21121, opened 2026-01-11
Expand User Messages documentation in Change Logging


mrmrcoleman:
### Change Type

Addition

### Area

Features

### Proposed Changes

The "User Messages" section in `docs/features/change-logging.md` currently states that users can record a message but does not explain how. This leaves users to discover the feature on their own.

Propose expanding this section to:

- Explain that the "Changelog message" field appears at the bottom of object create/edit forms and delete confirmation dialogs
- Note that it is optional and limited to 200 characters
- Clarify it works for individual and bulk operations
- Add a cross-reference to the REST API documentation for programmatic usage

Additionally, add a back-reference from the REST API "Changelog Messages" section to the change logging feature documentation.

PR #21122, opened 2026-01-11
Expand changelog message documentation and add cross-references

mrmrcoleman:
Fixes: https://github.com/netbox-community/netbox/issues/21121

### DIFF
--- a/docs/features/change-logging.md
+++ b/docs/features/change-logging.md
@@ -10,9 +10,11 @@ Change records are exposed in the API via the read-only endpoint `/api/extras/ob
 
 ## User Messages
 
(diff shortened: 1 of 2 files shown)


jnovinger:
@mrmrcoleman, this looks good. But I think you probably want to target the `main` branch with this, otherwise it will be waiting to be published until v4.6 comes out, in late April or early May.

mrmrcoleman:
> @mrmrcoleman, this looks good. But I think you probably want to target the `main` branch with this, otherwise it will be waiting to be published until v4.6 comes out, in late April or early May.

@jnovinger Done: https://github.com/netbox-community/netbox/pull/21138

PR #21182, opened 2026-01-15
Clear Rack Face when clear Rack

arthanson:
### Fixes: #20383 

We could just clear this field in clean, but it still leaves the UI field set when clearing the rack which is different then the behavior of position field. 

The Position field is cleared because it has rack in the api_url but rack_face is just a regular select and doesn't need any URL params.  It would be useful in other places to allow clearing related fields like this so made a ts-clear-field attribute, like the existing ts-disabled-field to allow setting a field to be auto-cleared on another field being cleared.


netbox/project-static/src/forms/elements.ts
```diff
@@ -28,4 +60,7 @@ export function initFormElements(): void {
       });
     }
   }
+
+  // Initialize clear-field dependencies
+  initClearFieldDependencies();
```
jeremystretch: This probably shouldn't live inside `initFormElements()`. Can we move it into its own module, similar to what we did with `initSpeedSelector()`?

netbox/dcim/forms/model_forms.py
```diff
@@ -722,6 +722,9 @@ def __init__(self, *args, **kwargs):
         if position:
             self.fields['position'].widget.choices = [(position, f'U{position}')]
 
+        # Clear face field when rack is cleared
+        self.fields['face'].widget.attrs['ts-clear-field'] = 'rack'
```
jeremystretch: I'm not a huge fan of hacking widget attributes like this. Could we introduce a new custom subclass of the stock Select widget that takes an argument e.g. `requires_fields` perhaps?

```python
face = forms.ChoiceField(
    choices=DeviceFaceChoices,
    widget=CustomSelect(
        requires_fields=['rack']
    )
```

Something like that. This would attach a list of field names to the rendered HTML element, which we could have `initClearFieldDependencies()` iterate through accordingly. (Hopefully that makes sense.)

jeremystretch (changes requested):

netbox/utilities/forms/widgets/select.py
```diff
@@ -28,6 +29,24 @@ def __init__(self, *args, **kwargs):
...
+                        are cleared, this field will also be cleared automatically via JavaScript.
+    """
+
+    def __init__(self, *args, requires_fields=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        if requires_fields:
+            # Convert list to comma-separated string for the data attribute
+            if isinstance(requires_fields, (list, tuple)):
+                requires_fields = ','.join(requires_fields)
+            self.attrs['data-requires-fields'] = requires_fields
```
jeremystretch: IMO we should just require this to be an iterable. Up to you.
```suggestion
            self.attrs['data-requires-fields'] = ','.join(requires_fields)
```

jeremystretch (dismissed):
Approved with a minor suggestion. Thanks @arthanson!

netbox/utilities/forms/widgets/select.py
arthanson: Made this change, unfortunately requires re-review.

jeremystretch (approved):

PR #21184, opened 2026-01-15
Closes #21039: Add AVIF support for image attachments

### LINKED ISSUE #21039: avif and potential other image formats missing in list of acceptable extensions for handling filenames in image_upload
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.9

### Python Version

3.12

### Steps to Reproduce

1. Upload avif image ("avif-sample.avif") as Image Attachment for example from https://convertico.com/samples/avif/
2. Type something as name or leave name blank, does not matter

### Expected Behavior

Image is accessible as https://demo.netbox.dev/media/image-attachments/device_27_avif-sample.avif

### Observed Behavior

Image is accessible as https://demo.netbox.dev/media/image-attachments/device_27_avif-sample

### Related

* #17788
* #19198

pheus:
### Fixes: #21039

Preserve `.avif` extensions when uploading Image Attachments (so the attachment is served with the expected filename), and ensure AVIF files are selectable in the UI.

* Add AVIF (`.avif` / `image/avif`) to the list of acceptable image extensions used when generating attachment filenames, so the stored file keeps its extension.
* Update the image upload widget to explicitly include AVIF in the `accept` attribute (required for Firefox to offer AVIF files in the file picker).
* Manual test: upload a AVIF file as an Image Attachment and confirm it’s served at `/media/image-attachments/...avif`.

Thanks for the review!

### DIFF
--- a/netbox/extras/constants.py
+++ b/netbox/extras/constants.py
@@ -4,6 +4,17 @@ from extras.choices import LogLevelChoices
 # Custom fields
 CUSTOMFIELD_EMPTY_VALUES = (None, '', [])
 
+# ImageAttachment
+IMAGE_ATTACHMENT_IMAGE_FORMATS = {
+    'avif': 'image/avif',
+    'bmp': 'image/bmp',
+    'gif': 'image/gif',
+    'jpeg': 'image/jpeg',
+    'jpg': 'image/jpeg',
+    'png': 'image/png',
+    'webp': 'image/webp',
+}
+
 # Template Export
 DEFAULT_MIME_TYPE = 'text/plain; charset=utf-8'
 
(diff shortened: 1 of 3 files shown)


arthanson (dismissed):
This works but the upload form is weird (at least on Safari)

This is before, how it should look:
Monosnap Add a new image attachment | NetBox 2026-01-16 14-48-51

This is after:

Monosnap Add a new image attachment | NetBox 2026-01-16 14-49-13

pheus:
Thanks for catching the FileInput layout change! I really appreciate it.

_You’re right_: it ended up being a side effect of a last-minute switch to NetBox’s custom widget. I initially tested with Django’s default file input, then swapped widgets right before pushing without doing a full UI check. I’ll make sure to re-test any UI-related changes across the major browsers going forward.

I’ve now reverted that widget change so the layout is back to normal. While I was in there, I also took the opportunity to clean things up by centralizing the supported image formats in a single constants mapping, and updating the widgets/utilities to use it (including explicitly adding AVIF to the `accept` attribute for Firefox).

Thanks again for taking a thoughtful look, and apologies for the unintended UI change.

jeremystretch (approved):
Nicely done @pheus!

PR #21186, opened 2026-01-15
Fixes #21166: Fix support for filtering on unsigned 32-bit integer values in GraphQL API

### LINKED ISSUE #21166: 4 Byte ASN filtering doesn't work via GraphQL
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.2

### Python Version

3.12

### Steps to Reproduce

Run following graphql query or simliary for any 4 byte ASN greater than 2,147,483,647:

```
query MyQuery {
  asn_list(filters: {asn: {filter_lookup: {exact: 4277815416}}}) {
    id
  }
}
```

### Expected Behavior

Return matching ASN from netbox

### Observed Behavior

This returns the error:

`Int cannot represent non 32-bit signed integer value: 4277815416`

This occurs because the int type is a signed 32 bit number in graphql (where as 4 byte asn are unsigned 32bits) and graphql doesn't have a native type for 64 bit integers or similar "longint" equivalent. 

I would suggest there are a couple of options here:

1. Change the graphql filter of asn to string (this is breaking change to existing graphql behaviour for 2 byte ASNs)

i.e. from

```
asn: Annotated['IntegerLookup', strawberry.lazy('netbox.graphql.filter_lookups')] | None = (
    )
```

to:

`asn: FilterLookup[str] | None = strawberry_django.filter_field()`

2. A new type is created which caters for long ints, this would mean creating a new filter etc (this would add complexity)

I have tested option 1 and it works (and honestly is my preference), but I can understand why the project would probably prefer option 2 despite the added complexity. 

Reference: https://strawberry.rocks/docs/types/scalars#:~:text=)-,BigInt%20(64%2Dbit%20integers),-Python%20integers%20have

jeremystretch:
### Fixes: #21166

- Introduce BigIntegerLookup (replicated from IntegerLookup)
- Employ it on the following filters:
    - ASNFilter.asn
    - ASNRangeFilter.start
    - ASNRangeFilter.end
    - L2VPNFilter.identifier

### DIFF
--- a/netbox/ipam/graphql/filters.py
+++ b/netbox/ipam/graphql/filters.py
@@ -20,7 +20,7 @@ from tenancy.graphql.filter_mixins import ContactFilterMixin, TenancyFilterMixin
 from virtualization.models import VMInterface
 
 if TYPE_CHECKING:
-    from netbox.graphql.filter_lookups import IntegerLookup, IntegerRangeArrayLookup
+    from netbox.graphql.filter_lookups import BigIntegerLookup, IntegerLookup, IntegerRangeArrayLookup
     from circuits.graphql.filters import ProviderFilter
     from core.graphql.filters import ContentTypeFilter
     from dcim.graphql.filters import SiteFilter
@@ -53,7 +53,7 @@ __all__ = (
 class ASNFilter(TenancyFilterMixin, PrimaryModelFilter):
     rir: Annotated['RIRFilter', strawberry.lazy('ipam.graphql.filters')] | None = strawberry_django.filter_field()
     rir_id: ID | None = strawberry_django.filter_field()
-    asn: Annotated['IntegerLookup', strawberry.lazy('netbox.graphql.filter_lookups')] | None = (
+    asn: Annotated['BigIntegerLookup', strawberry.lazy('netbox.graphql.filter_lookups')] | None = (
         strawberry_django.filter_field()
     )
     sites: (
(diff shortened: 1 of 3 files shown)


jnovinger (commented):
This look like a good solution and does resolve the original issue. However, I notice that `range_lookup` is not working as I would expect (although, it's very possible my expectation is wrong) when used with something like the following query:

```gql
query MyQuery {
  asn_list(filters: {asn: {range_lookup: {start: 0, end: 4277815416}}}) {
    id
  }
}
```

jnovinger (commented):
Ah, also just did a quick check for big integer sized fields that might not be covered and found `Tunnel.tunnel_id` which uses `PositiveBigInteger`.

I realize it's not mentioned in the original issue, but it seems worth fixing with this as well.

jeremystretch:
Ah, I missed `tunnel_id` in my review because it looked like a foreign key. Yeah, let me fix that one too.

jnovinger:
> This look like a good solution and does resolve the original issue. However, I notice that `range_lookup` is not working as I would expect (although, it's very possible my expectation is wrong) when used with something like the following query:
> 
> ```graphql
> query MyQuery {
>   asn_list(filters: {asn: {range_lookup: {start: 0, end: 4277815416}}}) {
>     id
>   }
> }
> ```

@jeremystretch, just wanted to make sure you saw this comment as well. Happy to open a follow up bug for it if you prefer.

jnovinger (approved):
Approved, but not merged. I had another question about a `range_filter` query. If you want to tackle that in a follow up, feel free to merge this.

jeremystretch:
> Happy to open a follow up bug for it if you prefer.

@jnovinger this would be out of scope for this PR as the error is not limited to BigIntegerLookup. Please open a separate bug report for that issue.

cnemo-cenic:
Wait, how is this supposed to work? My query broke because of this.

```
Variable '$asn' of type 'Int!' used in position expecting type 'Union'., Locations: [{Line:1 Column:8} {Line:1 Column:67}], Extensions: map[internal:map[request:map[body:{\"query\":\"query ($asn:Int!){asn_list(filters: {asn: {filter_lookup: {exact: $asn}}}){asn,sites{name,tenant{name}}}}\",\"variables\":{\"asn\":23309}}"
```

Edit: Wow, your GraphQL schema has a literal type named "Union" that represents the BigInt type... gross, and more importantly a breaking change that should have been announced. Second time your GraphQL API has broken us in six months. https://github.com/netbox-community/netbox/issues/15569#issuecomment-3451699127

Screenshot 2026-02-10 at 10 32 33

PR #21187, opened 2026-01-15
Add Max Depth and Max Length dropdowns for child prefix pages

### LINKED ISSUE #21088: Add Max Depth and Max Length dropdowns for child prefix pages
### NetBox version

v4.4.2

### Feature type

New functionality

### Proposed functionality

Currently the aggregate and prefixes child prefixes pages allow `mask_length__lte` and `depth__lte` parameters when added to the url, but do not have buttons/dropdowns to support this the web interface. I propose adding those buttons (aligning to how it works for prefixes). 

Example below:

Image

Note: I am happy to create a PR for this if this feature is accepted. 

### Use case

This allows people to create a "simpler" top down view in particular when they have substantial nesting under aggregates or prefixes. 

### Database changes

N/A - This would be a html change only.

### External dependencies

None

matthew-242:
### Fixes: #21088 

This adds the max_length and max_depth drop down filters to the child "prefix" pages for both aggregates and prefixes. It achieves this by creating 2 new templates which can be then be included wherever this functionality is required. This PR also standardises the existing prefix_list template to include these new templates.

### DIFF
--- a/netbox/templates/ipam/aggregate/prefixes.html
+++ b/netbox/templates/ipam/aggregate/prefixes.html
@@ -3,6 +3,8 @@
 
 {% block extra_controls %}
   {% include 'ipam/inc/toggle_available.html' %}
+  {% include 'ipam/inc/max_depth.html' %}
+  {% include 'ipam/inc/max_length.html' %}  
   {% if perms.ipam.add_prefix and first_available_prefix %}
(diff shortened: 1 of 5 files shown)


jeremystretch (approved):
Thanks @matthew-242!

PR #21188, opened 2026-01-15
Fixes #21051: Fix ObjectPermission object type selector to exclude internal/third-party models

### LINKED ISSUE #21051: ObjectPermission object type selector includes internal/third-party models
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.8

### Python Version

3.12

### Steps to Reproduce

1. Navigate to **Admin → Permissions** (or go to `/users/permissions/add/`)
2. Click **Add** (create a new Object Permission)
3. In the **Object types** selector, scroll/search for terms like `ObjectType`, `cached`, `queue`, etc.
4. Observe that internal/third-party models appear in the selectable list

### Expected Behavior

The Object types selector should only include object types that make sense for NetBox ObjectPermissions, for example:

- user-facing NetBox models, and
- plugin models intended to participate in NetBox’s object permission system

Internal metadata models and third-party dependency/tooling models (like the examples above) should not be offered as selectable object types.

### Observed Behavior

The selector includes internal/third-party models such as:

- `core.ObjectType`
- `debug_toolbar.historyentry` *
- `django_rq.queue`
- `extras.cached_value`
- `migrations.migration` *
- `thunmbnail.kv_store`

*) only listed in a local dev environment

---

#### Additional Notes / Suggestion

_First off_: thank you for the recent improvements around the ObjectPermission UI (especially the object type grouping introduced around #20759). It makes the selector much easier to navigate.

From a quick look, this seems to be related to the queryset/filter used to build the choices. The current filter appears to be:

    OBJECTPERMISSION_OBJECT_TYPES = Q(
        ~Q(app_label__in=['account', 'admin', 'auth', 'contenttypes', 'sessions', 'taggit', 'users']) |
        Q(app_label='users', model__in=['objectpermission', 'token', 'group', 'user'])
    )

As written, this behaves like a small denylist (plus a special-case allowlist for `users.*`), which means *any other installed Django app* (including dev tooling / background worker apps) can show up as selectable object types.

I’d like to suggest a review of **all** object types that are currently eligible for ObjectPermissions. A short-term fix could be to exclude known internal and common third-party tooling apps/models.

Longer-term, it might be more robust to base the selector on a “user-facing/public models” concept (while still supporting plugins), rather than relying on a denylist of a few Django apps.

Happy to help test any changes.

pheus:
#### Summary

Thank you for reviewing/triaging this issue and for the earlier guidance on treating this as a short-term bugfix.

This PR tightens the object type selector for **Object Permissions** to avoid listing internal-only and third-party/dev tooling models. Instead of maintaining a growing denylist of app labels, it leans on NetBox’s existing `ObjectType.public` flag and adds a small set of explicit exceptions for private-but-permission-relevant models.

#### Changes

- Replace the current `OBJECTPERMISSION_OBJECT_TYPES` denylist-style filter with:

  ```python
  OBJECTPERMISSION_OBJECT_TYPES = (
      (Q(public=True) & ~Q(app_label='core', model='objecttype'))
      | Q(app_label='core', model__in=['managedfile'])
      | Q(app_label='extras', model__in=['scriptmodule', 'taggeditem'])
  )
  ```

- Explicitly exclude **`Core > Object Type`** (`core.objecttype`) from being selectable.
- Explicitly include the private models that are expected to remain permissionable:
  - **Core > Managed File**
  - **Extras > Script Module**
  - **Extras > Tagged Item**

#### Resulting behavior (high level)

This removes several non-user-facing/internal or third-party object types which were previously shown in the selector, including:

- `Core > Auto Sync Record`
- `Core > Object Type`
- `DCIM > Cable Path` *
- `DCIM > Port Mapping` *
- `DCIM > Port Template Mapping` *
- `Django_Rq > Queue`
- `Extras > Cached Value`
- `Python Social Auth > *`
- `Thumbnail > Kv Store`

\* These models don’t use `RestrictedQuerySet` today and therefore don’t participate in object-permission filtering.

It also adds:

- `Users > Owner Group`

#### Rationale

- The prior approach was effectively “include everything except a small denylist”, which allowed unrelated installed apps (e.g. background worker tooling, debug/thumbnail/social-auth dependencies) to appear as selectable ObjectPermission targets.
- Using `public=True` makes the selector align with NetBox’s concept of “user-facing” models and avoids continually expanding a denylist.
- A few private models still need to remain selectable for real workflows (e.g. scripts), so they are explicitly included.

#### Alternative considered

As a short-term fix, an alternative would have been to expand the existing denylist and re-allow a small set of models:

```python
OBJECTPERMISSION_OBJECT_TYPES = Q(
    ~Q(app_label__in=[
        'account', 'admin', 'auth', 'contenttypes', 'core', 'django_rq', 'extras', 'migrations', 'sessions', 'taggit',
        'thumbnail', 'users',
    ]) |
    Q(app_label='core', model__in=['configrevision', 'datafile', 'datasource', 'job', 'managedfile', 'objectchange']) |
    Q(app_label='users', model__in=['objectpermission', 'token', 'group', 'user', 'owner', 'ownergroup']) |
    (Q(app_label='extras') & ~Q(model='cachedvalue'))
)
```

I opted for the `public=True` approach (with explicit exceptions) to reduce the likelihood of new unrelated installed apps/models showing up in the selector over time.

#### Testing

Quick manual check in `nbshell`:

```python
from users.constants import OBJECTPERMISSION_OBJECT_TYPES

for ot in ObjectType.objects.filter(OBJECTPERMISSION_OBJECT_TYPES).order_by('app_label', 'model'):
    print(ot.app_labeled_name)
```

#### Follow-up

Per maintainer guidance, this PR focuses on the short-term bugfix. I plan to open a separate housekeeping issue to track the longer-term solution (i.e., defining a more explicit and/or feature-based mechanism for “permissionable object types”).

### DIFF
--- a/netbox/users/constants.py
+++ b/netbox/users/constants.py
@@ -3,9 +3,10 @@ import string
 from django.db.models import Q
 
 
-OBJECTPERMISSION_OBJECT_TYPES = Q(
-    ~Q(app_label__in=['account', 'admin', 'auth', 'contenttypes', 'sessions', 'taggit', 'users']) |
-    Q(app_label='users', model__in=['objectpermission', 'token', 'group', 'user', 'owner'])
+OBJECTPERMISSION_OBJECT_TYPES = (
+    (Q(public=True) & ~Q(app_label='core', model='objecttype'))
+    | Q(app_label='core', model__in=['managedfile'])
+    | Q(app_label='extras', model__in=['scriptmodule', 'taggeditem'])
 )
 
 CONSTRAINT_TOKEN_USER = '$user'


jeremystretch (approved):
Confirmed that this removes the following object types:

- Core > Auto Sync Record
- Core > Object Type
- DCIM > Cable Path
- DCIM > Port Mapping
- DCIM > Port Mapping Template
- Django RQ > Queue
- Extras > Cached Value
- Python Social Auth > * (All)
- Thumbnail > * (All)

Additionally, it adds the following:

- Users > Owner Group

PR #21193, opened 2026-01-16
Enable specifying mask length when creating IP addresses via available-ips endpoint

### LINKED ISSUE #21144: Enable specifying a mask length when creating a new IP address via the "available IPs" REST API endpoint
### NetBox version

v4.5.0

### Feature type

New functionality

### Proposed functionality

When creating an IP address from a parent prefix or range using the "next available" REST API endpoint (e.g. `/api/ipam/prefixes/123/available-ips/`), permit the client to specify an arbitrary mask length. For example:

```
curl -X POST \
-H "Authorization: Bearer <token>" \
http://netbox/api/ipam/prefixes/7/available-ips/ \
--data '{"prefix_length": 32}'
```

This can be achieved using PrefixLengthSerializer, which currently provides this functionality when creating new child prefixes.

### Use case

It may be desirable in some cases to create IP addresses with a mask length greater than that of the parent prefix, e.g. when creating loopback IP addresses.

### Database changes

N/A

### External dependencies

N/A

adionit7:
## Summary
Enable specifying a mask length when creating IP addresses via the available-ips REST API endpoint.

## Changes
- Updated `AvailableIPAddressesView` to use `PrefixLengthSerializer` as `write_serializer_class`
- Enhanced `PrefixLengthSerializer` to support both 'prefix' and 'parent' context keys
- Added validation to ensure requested `prefix_length >= parent.mask_length`
- Updated `prep_object_data` to use requested `prefix_length` if provided, otherwise fall back to `parent.mask_length` for backwards compatibility
- Updated API schema documentation

## Example Usage
curl -X POST \
-H "Authorization: Bearer <token>" \
http://netbox/api/ipam/prefixes/123/available-ips/ \
--data '{"prefix_length": 32}'

### DIFF
--- a/netbox/ipam/api/serializers_/ip.py
+++ b/netbox/ipam/api/serializers_/ip.py
@@ -19,6 +19,7 @@ from ..field_serializers import IPAddressField, IPNetworkField
 __all__ = (
     'AggregateSerializer',
     'AvailableIPSerializer',
+    'AvailableIPRequestSerializer',
     'AvailablePrefixSerializer',
     'IPAddressSerializer',
     'IPRangeSerializer',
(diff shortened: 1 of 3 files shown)


netbox/ipam/api/serializers_/ip.py
```diff
@@ -75,20 +75,26 @@ class Meta:
 
 class PrefixLengthSerializer(serializers.Serializer):
 
-    prefix_length = serializers.IntegerField()
+    prefix_length = serializers.IntegerField(required=False)
```
jeremystretch: `prefix_length` is always required for prefix allocation. It probably doesn't make sense to use the same serializer for both endpoints.

jeremystretch (changes requested):

netbox/ipam/api/serializers_/ip.py
adionit7: Good catch, thanks Jeremy!
I’ve reverted `PrefixLengthSerializer` to remain strict for the available-prefixes endpoint and introduced a separate `AvailableIPRequestSerializer `for the available-ips endpoints, where `prefix_length `is optional.
The `available-ips `view now uses this new request serializer and applies the requested `prefix_length` (or falls back to the parent mask length) when creating IP addresses, without changing the behavior of available-prefixes.

jeremystretch: Could you revert the changes to PrefixLengthSerializer please?

jeremystretch (changes requested):

netbox/ipam/api/serializers_/ip.py
adionit7: Done @jeremystretch. Reverted the changes.

jeremystretch (approved):
Thanks @adionit7!

PR #21210, opened 2026-01-17
Virtual Machine Resources Use Decimal MB/GB Instead of Binary MiB/GiB

### LINKED ISSUE #21095: Virtual Machine Resources Use Decimal MB/GB Instead of Binary MiB/GiB, Causing Memory Mismatch
### NetBox version

v4.5.0

### Feature type

### Proposed functionality

Memory (and disk) values under Virtualization → Virtual Machine → Resources are currently stored and displayed using decimal units (MB/GB), but the UI labels them simply as “GB”. In practice, most users interpret this as binary units (GiB), which is what operating systems and hypervisors commonly use.

The proposal is to adjust how units are handled and/or displayed to avoid this ambiguity. Possible approaches include:

- Switching memory and disk resources to use binary units (MiB/GiB) consistently throughout the UI and data model, or
- Clearly labeling values as decimal (MB/GB) wherever they are shown, or
- Providing a configurable option to display resources using either decimal or binary units
At a minimum, the UI should make it explicit which unit system is being used so users can accurately allocate resources.

This change would primarily affect display and input handling in the VM resources UI, without changing the underlying numeric values unless binary units are adopted.

### Use case

When configuring virtual machines, users often allocate memory in round binary values (e.g. 4 GiB, 8 GiB). Currently, setting memory to 4000 MB results in the VM summary showing 4.00 GB, which looks correct at a glance but actually equates to ~3.9 GiB inside the guest OS.

This causes confusion, makes it appear that memory is missing, and makes capacity planning less precise. Since most operating systems, hypervisors, and monitoring tools report memory in binary units, the current behavior doesn’t align with user expectations.

Clarifying or standardizing units would reduce confusion, prevent accidental under-allocation, and make NetBox’s virtualization resource modeling easier to reason about for operators managing large numbers of VMs.

Image
Image
Image
Image

### Database changes

_No response_

### External dependencies

_No response_

kakathian:
### Fixes: #21095 
This change would primarily affect display and input handling in the VM resources UI, without changing the underlying numeric values unless binary units are adopted.

### DIFF
--- a/netbox/virtualization/forms/model_forms.py
+++ b/netbox/virtualization/forms/model_forms.py
@@ -221,7 +221,7 @@ class VirtualMachineForm(TenancyForm, PrimaryModelForm):
         FieldSet('site', 'cluster', 'device', name=_('Site/Cluster')),
         FieldSet('tenant_group', 'tenant', name=_('Tenancy')),
         FieldSet('platform', 'primary_ip4', 'primary_ip6', 'config_template', name=_('Management')),
-        FieldSet('vcpus', 'memory', 'disk', name=_('Resources')),
+        FieldSet('vcpus', 'memory', 'disk', name=_('Resources (MiB)')),
         FieldSet('local_context_data', name=_('Config Context')),
     )
 
@@ -393,7 +393,7 @@ class VMInterfaceForm(InterfaceCommonForm, VMComponentForm):
 class VirtualDiskForm(VMComponentForm):
 
     fieldsets = (
-        FieldSet('virtual_machine', 'name', 'size', 'description', 'tags', name=_('Disk')),
+        FieldSet('virtual_machine', 'name', 'size', 'description', 'tags', name=_('Disk (GiB)')),
     )
 
     class Meta:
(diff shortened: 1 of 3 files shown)


jnovinger:
@kakathian, our contribution process specifically calls for not working on issues or filing PRs until an issue has been assigned to you. In order for us to assign it to you, you must comment on the original issue indicating you would like to work on it.

jnovinger:
@kakathian , can you please address the CI failures?

kakathian:
Thanks. This PR is ready for reviews

bctiemann:
Hi @kakathian, looking at this PR it appears that all it does it change the labeling in the UI to reflect a presumption that the stored values are binary (MiB/GiB). It doesn't seem to take into account the setting of `DISK_BASE_UNIT`/`RAM_BASE_UNIT` (which might be set to something else, and indeed by default are set to base 10, i.e. MB/GB or 1000 MB = 1 GB).

The proposal in the issue was meant to cover different users' needs by flexibly adapting to whether they are using base-10 or base-2 units, or else by adopting a comprehensive, system-wide enforcement of a standard storage unit and then reflecting the chosen display unit at the UI level.

bctiemann (changes requested):
Please see general comments.

PR #21212, opened 2026-01-18
Fixes: #20490 - Add filtering of Script objects based on object permissions with custom constraints

### LINKED ISSUE #20490: Do not display scripts in the scripts list view if the user does not have permission to view them
### NetBox version

v4.4.2

### Feature type

New functionality

### Proposed functionality

Currently, if you add a permission constraint to the `Script` model limiting the scripts a user can view, the script is still displayed in the list view and returns a 404 when the hyperlink is clicked. I'm proposing to add a permissions check when rendering the scripts list view that will only display the scripts a user has permission to view. This should be able to be accomplished by adding a permissions check in the [script_list_content.html](https://github.com/netbox-community/netbox/blob/main/netbox/templates/extras/inc/script_list_content.html)  template.

Example `Script` model permission constraint:
```
{
  "name__in": [
    "MyScript",
    "YourScript"
  ]
}
```

### Use case

Limiting the scripts in the list view based on permissions would improve the use experience by consolidating the scripts within the list and preventing 404 responses when the hyperlinks are clicked.

### Database changes

None

### External dependencies

None

bctiemann:
### Fixes: #20490 

Adds the ability to filter Scripts based on custom object-level permissions with `constraints` such as:

```
    {
        "name__in": [
            "MyScript"
        ]
    }
```

Note that this permission is applied against `Script` objects, allowing permissions to be applied against `Script` or `ScriptModule` objects via ORM lookups; however, access to the overall Scripts page is given via the `ManagedFile` object type.


docs/customization/custom-scripts.md
```diff
@@ -18,7 +18,17 @@ They can also be used as a mechanism for validating the integrity of data within
...
 
 !!! danger "Only install trusted scripts"
-    Custom scripts have unrestricted access to change anything in the databse and are inherently unsafe and should only be installed and run from trusted sources.  You should also review and set permissions for who can run scripts if the script can modify any data.
+    Custom scripts have unrestricted access to change anything in the database and are inherently unsafe and should only be installed and run from trusted sources.  You should also review and set permissions for who can run scripts if the script can modify any data.
+
+!!! tip "Permissions for Custom Scripts"
+    A user can be granted permissions on all Custom Scripts via the "Managed File" object-level permission. To further restrict a user to only be able to access certain scripts, create an additional permission on the "Script" object type, with appropriate queryset-style constraints matching fields available on Script. For example:
```
jeremystretch: This seems very confusing, as "script module" appears as an available object type in the permission form. I don't think we should rely on the ManagedFile model for permissions here, as it's not intended to be a "public" model. Or, if we can't get permissions to work using ScriptModule (a proxy model), we should remove it from the list of available models.

jeremystretch (changes requested):

docs/customization/custom-scripts.md
bctiemann: I reworked and simplified this so that the permission that should be applied is on the `Script` object, which is concrete.

This may have some performance implications since the prefetch is now gone and we have a new iteration, but I feel like if there are enough scripts/modules that this is a significant concern the overall layout/navigation of the page is too rudimentary to support lists that big anyway (we'd need to add pagination etc).

netbox/extras/views.py
```diff
@@ -1441,12 +1441,13 @@ def get_required_permission(self):
         return 'extras.view_script'
 
     def get(self, request):
-        script_modules = ScriptModule.objects.restrict(request.user).prefetch_related(
-            'data_source', 'data_file', 'jobs'
-        )
+        available_scripts = Script.objects.restrict(request.user)
+        script_modules = [s.module for s in available_scripts]
```
jeremystretch: Let's keep the prefetch in place if we can. We may have users with an inordinate number of custom scripts.

```suggestion
        module_ids = [script.module_id for script in available_scripts]
        script_modules = ScriptModule.objects.restrict(request.user).filter(pk__in=module_ids).prefetch_related(
            'data_source', 'data_file',
        )
```

(I omitted `jobs` here because it doesn't seem to be used.)

jeremystretch (changes requested):

netbox/extras/views.py
bctiemann: I had the same lunch thought and was coming back here to do just that (also deduping the IDs with a set)

jeremystretch (approved):

PR #21218, opened 2026-01-19
Fixes #21213: Make Tag weight field required in forms

adionit7:
The weight field was explicitly declared with `required=False` in `TagForm` and `TagImportForm`, allowing empty submissions that would crash with a database `IntegrityError` since the column is `NOT NULL`.

By removing the explicit field override, Django now auto-generates the form field from the model, which has `default=1000` and is required. This ensures the UI matches the database constraint and prevents invalid submissions before they reach the database.

### Summary

- Remove the explicit `weight` field overrides from `TagForm` and `TagImportForm` that set `required=False`.
- Allow Django to use the model’s `PositiveSmallIntegerField(default=1000)` directly, making the field required at the form level.
- Align form validation with the database `NOT NULL` constraint to avoid `IntegrityError` when the weight is left empty.

### DIFF
--- a/netbox/extras/forms/bulk_import.py
+++ b/netbox/extras/forms/bulk_import.py
@@ -271,10 +271,6 @@ class EventRuleImportForm(OwnerCSVMixin, NetBoxModelImportForm):
 
 class TagImportForm(OwnerCSVMixin, CSVModelForm):
     slug = SlugField()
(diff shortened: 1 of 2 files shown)


arthanson (approved):
Thanks @adionit7 !

PR #21219, opened 2026-01-19
Fixes #21214: Clean up AutoSyncRecord when detaching from DataSource

### LINKED ISSUE #21214: AutoSyncRecord not cleaned up when detaching Config Template from Data Source
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.7

### Python Version

3.12

### Steps to Reproduce

1. Create a Git Data Source containing at least two Jinja2 files (e.g., a.j2, b.j2)
2. Run sync on the Data Source
3. Create two Config Templates:
   - CT-A → Data Source = <git>, Data File = a.j2, Auto sync enabled
   - CT-B → Data Source = <git>, Data File = b.j2, Auto sync enabled
4. Verify sync works by changing b.j2, commit/push, run Data Source sync, confirm CT-B updates
5. Edit CT-A and detach it from the Data Source:
   - Set data_source = NULL
   - Set data_file = NULL  
   - Set auto_sync_enabled = False
   - Save the Config Template
6. Make a manual content edit to CT-A
7. Change b.j2 again in git, commit/push, run Data Source sync

### Expected Behavior

When a Config Template is detached from a Data Source (data_source and data_file cleared, auto_sync_enabled disabled), the associated AutoSyncRecord should be automatically cleaned up. The Data Source sync should continue successfully, syncing other valid objects (like CT-B).

### Observed Behavior

The Data Source sync job fails with an error (e.g., 'NoneType' object has no attribute 'data_as_string') and prevents other objects from syncing.

The stale AutoSyncRecord remains even though the Config Template is detached:

```python
from extras.models import ConfigTemplate
from core.models import AutoSyncRecord

ct = ConfigTemplate.objects.get(name="CT-A")
ct.data_source_id, ct.data_file_id, ct.auto_sync_enabled   
# (None, None, False)

ct_type = ContentType.objects.get_for_model(ConfigTemplate)
AutoSyncRecord.objects.filter(object_type=ct_type, object_id=ct.pk).values()
# Returns record with datafile_id still set
```

Deleting the stale AutoSyncRecord manually resolves the issue:

```python
AutoSyncRecord.objects.filter(object_type=ct_type, object_id=ct.pk).delete()
```

After deletion, Data Source sync succeeds and remaining templates sync as expected.

adionit7:
When a synced object (e.g., `ConfigTemplate`) is detached from a `DataSource`, the `clean()` method sets `data_file` to `None` before `save()` is called. The previous code attempted to delete the `AutoSyncRecord` by filtering with `datafile=self.data_file`, which in this case is `None`, so it never matched the existing record and left a stale `AutoSyncRecord` behind.

This caused subsequent DataSource sync jobs to fail when trying to sync objects that no longer had a valid `data_file`. The fix removes the `datafile` filter from the delete query and relies on the existing `UniqueConstraint` on `(object_type, object_id)` to correctly identify and delete the associated `AutoSyncRecord`.

### Summary

- Update `SyncedDataMixin.save()` and `.delete()` to delete `AutoSyncRecord` instances by `(object_type, object_id)` only.
- Remove the `datafile=self.data_file` filter, which could be `None` after `clean()` has cleared `data_file` when detaching from a `DataSource`.
- Ensure that stale `AutoSyncRecord` rows are cleaned up when a synced object is detached, allowing subsequent DataSource syncs to proceed without errors.

### DIFF
--- a/netbox/extras/tests/test_models.py
+++ b/netbox/extras/tests/test_models.py
@@ -6,7 +6,7 @@ from django.core.files.uploadedfile import SimpleUploadedFile
 from django.forms import ValidationError
 from django.test import tag, TestCase
 
-from core.models import DataSource, ObjectType
+from core.models import AutoSyncRecord, DataSource, ObjectType
 from dcim.models import Device, DeviceRole, DeviceType, Location, Manufacturer, Platform, Region, Site, SiteGroup
 from extras.models import ConfigContext, ConfigContextProfile, ConfigTemplate, ImageAttachment, Tag, TaggedItem
 from tenancy.models import Tenant, TenantGroup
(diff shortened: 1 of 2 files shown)


jeremystretch:
@adionit7 is looks like you've captured commits here that were intended for your other PR, #21193.

adionit7:
> @adionit7 is looks like you've captured commits here that were intended for your other PR, #21193.

Yes Jeremy, that was pushed by mistake. I've cleaned up the branch and reverted those changes.. the PR now only contains the AutoSyncRecord fix.

jnovinger (changes requested):
Hey @adionit7, appreciate the quick fix. Can you please add a quick regression test, something like:

```python
def test_autosyncrecord_cleanup_on_detach(self):
    """Test that AutoSyncRecord is deleted when detaching from DataSource."""
    # Create a ConfigTemplate with data_file set and auto_sync_enabled=True

    # Verify AutoSyncRecord was created

    # Detach from DataSource by setting data_file=None

    # Call clean() and save()

    # Verify AutoSyncRecord was deleted
```

adionit7:
> Hey @adionit7, appreciate the quick fix. Can you please add a quick regression test, something like:
> 
> ```python
> def test_autosyncrecord_cleanup_on_detach(self):
>     """Test that AutoSyncRecord is deleted when detaching from DataSource."""
>     # Create a ConfigTemplate with data_file set and auto_sync_enabled=True
> 
>     # Verify AutoSyncRecord was created
> 
>     # Detach from DataSource by setting data_file=None
> 
>     # Call clean() and save()
> 
>     # Verify AutoSyncRecord was deleted
> ```

Hi @jnovinger, Sure.. I've added the regression test. It verifies AutoSyncRecord creation and cleanup on detach. 
It's ready for your review!

jnovinger (approved):

jnovinger:
Thanks, @adionit7!

adionit7:
> Thanks, @adionit7!

Thanks for the review @jnovinger! Happy to help.

PR #21224, opened 2026-01-19
Fixes #21181: Handle AuthenticationFailed exception on /media endpoint

jeremystretch:
### Fixes: #21181

Extend TokenConditionalLoginRequiredMixin to handle the AuthenticationFailed exception and return a 403 response

### DIFF
--- a/netbox/utilities/views.py
+++ b/netbox/utilities/views.py
@@ -53,7 +53,7 @@ class TokenConditionalLoginRequiredMixin(ConditionalLoginRequiredMixin):
         if settings.LOGIN_REQUIRED and not request.user.is_authenticated:
             authenticator = TokenAuthentication()
             try:
-                if auth_info := authenticator.authenticate(request) is not None:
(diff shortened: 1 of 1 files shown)


netbox/utilities/views.py
```diff
@@ -50,10 +52,12 @@ def dispatch(self, request, *args, **kwargs):
         # Attempt to authenticate the user using a DRF token, if provided
         if settings.LOGIN_REQUIRED and not request.user.is_authenticated:
             authenticator = TokenAuthentication()
-            auth_info = authenticator.authenticate(request)
-            if auth_info is not None:
-                request.user = auth_info[0]  # User object
-                request.auth = auth_info[1]
+            try:
+                if auth_info := authenticator.authenticate(request) is not None:
```
joernheissler:
```suggestion
                if (auth_info := authenticator.authenticate(request)) is not None:
```

arthanson (approved):

PR #21237, opened 2026-01-20
Fixes #21231: Cache table existence for ObjectType checks

### LINKED ISSUE #21231: ObjectTypeManager introspection not cached
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.0

### Python Version

3.12

### Steps to Reproduce

1. Enable DB logging:
```sql
ALTER SYSTEM SET log_min_duration_statement = 0;
SELECT pg_reload_conf();
```
2. Load basically any page with data, for example http://localhost:8080/dcim/racks/ or http://localhost:8080/api/dcim/racks/ (GraphQL not tested)
3. Check Postgres logs for queries on table `pg_catalog.pg_class`

### Expected Behavior

I'd expect only one introspection query to happen per worker, and the results to be cached until a restart or a migration

### Observed Behavior

Per HTTP request multiple introspection calls are made to the database. For REST calls this is usually two or three, for a frontend page load I've observed up to 11.

The issue seems to come from `netbox/core/models/object_types.py`. Patching out the below lines on our staging environment completely eliminated the introspection calls to our database.
https://github.com/netbox-community/netbox/blob/3efa23cf8f4c5de84617072a19833b46ad255b12/netbox/core/models/object_types.py#L72-L75

After analysis with PgBadger these queries were both the most frequent query and also the heaviest query for our Postgres instance. Some quick and dirty benchmarks of `get_for_model` showed ~44ms with the introspection in place, and ~3ms with it patched out.

pheus:
### Summary
This PR introduces a process-level cache for the `core_objecttype` table existence check within `ObjectTypeManager.get_for_model()`, based on the incredible findings of `@jetseverschuren-novoserve`.

Currently, NetBox performs database introspection on every `get_for_model()` call to determine if it should fall back to `ContentType` (a check retained for backward compatibility during v4.4 migrations). This hits `pg_catalog.pg_class` repeatedly, causing significant performance degradation on data-heavy pages and API endpoints.

**Proposed Changes:**
- Added a `_table_exists` boolean flag to `ObjectTypeManager`.
- The manager now performs introspection at most once per worker process.
- Once the table is detected, the flag is set to `True`, bypassing all future introspection calls for that process.
- The logic explicitly avoids caching a `False` result to ensure safety during active migrations or upgrades.

**Performance Impact:**
Benchmarks identified by the community show a reduction in lookup time from ~44ms to ~3ms per call in fully migrated environments by eliminating the high-frequency introspection queries.

**Testing:**
I have verified this solution by successfully migrating an empty v3.7.8 database to the most recent commit in `main`. The fallback logic correctly handles the pre-migration state and transitions seamlessly once the `core_objecttype` table is provisioned.

Special thanks to @jetseverschuren-novoserve for identifying this bottleneck and providing the detailed analysis!

### DIFF
--- a/netbox/core/models/object_types.py
+++ b/netbox/core/models/object_types.py
@@ -35,6 +35,10 @@ class ObjectTypeQuerySet(models.QuerySet):
 
 class ObjectTypeManager(models.Manager):
 
+    # TODO: Remove this in NetBox v5.0
+    # Cache the result of introspection to avoid repeated queries.
+    _table_exists = False
+
     def get_queryset(self):
         return ObjectTypeQuerySet(self.model, using=self._db)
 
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks for the quick turnaround @pheus!

PR #21248, opened 2026-01-21
Fixes #21173: Fix plugin menu registration order timing issue

### LINKED ISSUE #21173: Plugin Menu Registration Order
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.0 (and likely earlier versions)

### Python Version

3.12

### Steps to Reproduce

1. Install multiple plugins that register menus via `navigation.py` and `PluginConfig.ready()`:
   - netbox_dns
   - netbox_topology_views
   - ...

2. Ensure each plugin has a `navigation.py` file with a `menu` variable:
   ```python
   from netbox.plugins import PluginMenu, PluginMenuButton, PluginMenuItem

   menu = PluginMenu(
       label='Plugin',
       icon_class='mdi mdi-lan',
       groups=(...)
   )
   ```

3. Configure plugins in `configuration/plugins.py`:
   ```python
   PLUGINS = [
       'netbox_dns',
       'netbox_topology_views',
       'netbox_....'
   ]
   ```

4. Start NetBox and view the navigation sidebar
5. Observe which plugin menus appear in the sidebar

### Expected Behavior

All plugin menus should appear in the navigation sidebar regardless of plugin load order:
- DNS
- Topology Views
- ...

### Observed Behavior

Only some plugin menus appear in the navigation sidebar:
- DNS (appears)
- Topology Views (MISSING)
- ... (MISSING)

The missing menus are registered but not included in the final MENUS list.

---
In `netbox/netbox/navigation/menu.py`, the MENUS list is built at import time:

```python
MENUS = [
    ORGANIZATION_MENU,
    RACKS_MENU,
    # ... other base menus ...
]

# Add top-level plugin menus
for menu in registry['plugins']['menus']:
    MENUS.append(menu)
```

This code executes when navigation/menu.py is imported during Django startup (in settings.py). At that point, only the first few plugins have called PluginConfig.ready() and registered their menus.

1. settings.py imports plugins and calls `PluginConfig.ready()` for each
2. `PluginConfig.ready()` imports the plugin's `navigation.menu` module
3. `navigation.menu` calls `register_menu()` to add the menu to `registry['plugins']['menus']`
4. But `netbox/navigation/menu.py` is imported BEFORE all plugins complete `ready()`
5. The MENUS list is built with only the menus registered so far

adionit7:
Fixes #21173: Fix plugin menu registration order timing issue

## Summary

This PR fixes an issue where plugin menus registered later in the Django app initialization process were not appearing in the NetBox navigation sidebar. The problem occurred because the `MENUS` list was built at module import time, before all plugins had completed their `ready()` methods.

## Problem

When multiple plugins register menus via `navigation.py` and `PluginConfig.ready()`, only the first few plugins' menus would appear in the navigation sidebar. The missing menus were properly registered in the plugin registry, but were not included in the final `MENUS` list.

**Root Cause:**
- `netbox/navigation/menu.py` was imported during Django startup (in `settings.py`)
- At import time, the `MENUS` list was built by iterating over `registry['plugins']['menus']`
- However, only the first few plugins had completed their `ready()` methods at that point
- Later plugins' menus were registered after `MENUS` was already built, so they were excluded

## Solution

Converted the static `MENUS` list to a dynamic `get_menus()` function that builds the menu list on-demand at request time (when the navigation template tag is rendered). This ensures all plugins have completed their `ready()` methods before the menus are assembled.

**Changes Made:**
- Replaced static `MENUS` list with `get_menus()` function in `netbox/navigation/menu.py`
- Updated `netbox/utilities/templatetags/navigation.py` to call `get_menus()` instead of importing `MENUS`
- Function builds the same menu structure, just dynamically at request time

## Impact

- **Fixed:** All plugin menus now appear in the navigation sidebar regardless of plugin load order
- **Backward Compatible:** No changes to plugin registration API or menu structure
- **Performance:** Negligible impact - building menu list is microseconds per request
- **No Breaking Changes:** Function returns the same list structure as before

## Testing

- Verified that all plugin menus appear in navigation sidebar
- Confirmed existing plugin menu registration continues to work
- No changes required to plugin code - existing plugins work without modification

### DIFF
--- a/netbox/netbox/navigation/menu.py
+++ b/netbox/netbox/navigation/menu.py
@@ -501,40 +501,46 @@ ADMIN_MENU = Menu(
     ),
 )
 
-MENUS = [
-    ORGANIZATION_MENU,
-    RACKS_MENU,
-    DEVICES_MENU,
-    CONNECTIONS_MENU,
-    WIRELESS_MENU,
-    IPAM_MENU,
-    VPN_MENU,
-    VIRTUALIZATION_MENU,
-    CIRCUITS_MENU,
-    POWER_MENU,
-    PROVISIONING_MENU,
-    CUSTOMIZATION_MENU,
-    OPERATIONS_MENU,
-]
-
-# Add top-level plugin menus
-for menu in registry['plugins']['menus']:
-    MENUS.append(menu)
-
-# Add the default "plugins" menu
-if registry['plugins']['menu_items']:
-
-    # Build the default plugins menu
-    groups = [
-        MenuGroup(label=label, items=items)
-        for label, items in registry['plugins']['menu_items'].items()
+def get_menus():
+    """
+    Dynamically build and return the list of navigation menus.
+    This ensures plugin menus registered during app initialization are included.
+    """
+    menus = [
+        ORGANIZATION_MENU,
+        RACKS_MENU,
+        DEVICES_MENU,
+        CONNECTIONS_MENU,
+        WIRELESS_MENU,
+        IPAM_MENU,
+        VPN_MENU,
+        VIRTUALIZATION_MENU,
+        CIRCUITS_MENU,
+        POWER_MENU,
+        PROVISIONING_MENU,
+        CUSTOMIZATION_MENU,
+        OPERATIONS_MENU,
     ]
-    plugins_menu = Menu(
-        label=_("Plugins"),
-        icon_class="mdi mdi-puzzle",
-        groups=groups
-    )
-    MENUS.append(plugins_menu)
 
-# Add the admin menu last
-MENUS.append(ADMIN_MENU)
+    # Add top-level plugin menus
+    for menu in registry['plugins']['menus']:
+        menus.append(menu)
+
+    # Add the default "plugins" menu
+    if registry['plugins']['menu_items']:
+        # Build the default plugins menu
+        groups = [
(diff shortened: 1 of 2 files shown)


arthanson (changes requested):
This looks good - just checking if you looked at using a cache for this, something like:?

```
from functools import cache
 @cache
 def get_menus():
```

It's only loaded once per request, but might be easy to shave off a bit of processing each time as it shouldn't ever change without a Django restart.

adionit7:
> This looks good - just checking if you looked at using a cache for this, something like:?
> 
> ```
> from functools import cache
>  @cache
>  def get_menus():
> ```
> 
> It's only loaded once per request, but might be easy to shave off a bit of processing each time as it shouldn't ever change without a Django restart.

Thanks for the suggestion @arthanson! Great catch.

I've added the `@cache` decorator from `functools` to cache the result. Since the menu structure doesn't change without a Django restart, this eliminates redundant list building on each request.

Changes made:
- Added `from functools import cache` import
- Applied `@cache` decorator to `get_menus()`
- Updated docstring to document the caching behavior

arthanson (approved):
Thanks @adionit7 !

adionit7:
> Thanks @adionit7 !

Thanks for the review @arthanson! Happy to help.

PR #21250, opened 2026-01-21
Fixes #21249: Avoid unneeded user query when no event rules are present

### LINKED ISSUE #21249: `process_event_rules()` forces user lookup even when no event rules are defined
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.1

### Python Version

3.12

### Steps to Reproduce

1. Ensure no event rules exist 
2. Enable debugging & SQL logging
3. Create 100 objects via the REST API via a bulk request
4. Inspect the end of the SQL log (or use the debug toolbar)

### Expected Behavior

There should be no SQL queries issued relating to event rules.

### Observed Behavior

There are 100 instances of the following query at the end of the SQL log:

`SELECT ••• FROM "users_user" WHERE "users_user"."username" = 'admin' LIMIT 21`

This is caused by the following code:

https://github.com/netbox-community/netbox/blob/339ad455e4e635080f098cc275f0a1c56937777c/netbox/extras/events.py#L88-L91

It is necessary to resolve the user from the passed username only if there exists an event rule, and further, only if that event rule executes a custom script.

jeremystretch:
### Fixes: #21249

Avoid resolving `user` unless needed by an event rule

### DIFF
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -134,7 +134,7 @@ def process_event_rules(event_rules, object_type, event_type, data, username=Non
             # Resolve the script from action parameters
             script = event_rule.action_object.python_class()
 
-            # Retrieve the User if nto already resolved
+            # Retrieve the User if not already resolved
             if user is None:
                 user = User.objects.get(username=username)
 
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -86,7 +86,7 @@ def enqueue_event(queue, instance, request, event_type):
 
 
 def process_event_rules(event_rules, object_type, event_type, data, username=None, snapshots=None, request=None):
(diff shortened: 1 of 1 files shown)


netbox/extras/events.py
```diff
@@ -134,6 +134,10 @@ def process_event_rules(event_rules, object_type, event_type, data, username=Non
             # Resolve the script from action parameters
             script = event_rule.action_object.python_class()
 
+            # Retrieve the User if nto already resolved
```
jnovinger:
```suggestion
            # Retrieve the User if not already resolved
```

jnovinger (dismissed):
Approved with one minor suggestion

jnovinger (approved):

PR #21251, opened 2026-01-21
Fixes #21081: Add Owner Group column & filters across all supported object list views

### LINKED ISSUE #21081: Add "Owner Group" as a filter option and table column
### NetBox version

v4.5.0

### Feature type

New functionality

### Proposed functionality

The new owners are organized in groups. It would be nice to have these groups display as a column in the overview of devices and virtual machines. 
It would also be nice to have it as a filter for those views.

### Use case

We have specific departments which are responsible for their virtual machines. Until now, we had the department as a tenant for a vm and a custom field for the specific administrators.
With the new owner and owner group option we could get rid of these options and use the build in feature.
I can then filter for all VMs assigned to one/multiple departments.

### Database changes

I think none.

### External dependencies

None.

pheus:
### Fixes: #21081

This PR builds on the ownership model introduced in v4.5 by exposing **Owner Group** in the same places users already rely on for day-to-day workflows:

- Adds **Owner Group** as an available table column on all relevant list views (i.e., models which support ownership)
- Adds **Owner Group** as a filter option for those views (supporting selection of one or more groups)
- Updates the relevant dynamic model fields to better handle the `owner_group` ⇄ `owner` relationship in filter forms

### DIFF
--- a/netbox/dcim/forms/filtersets.py
+++ b/netbox/dcim/forms/filtersets.py
@@ -12,11 +12,12 @@ from netbox.forms import (
     NestedGroupModelFilterSetForm, NetBoxModelFilterSetForm, OrganizationalModelFilterSetForm,
     PrimaryModelFilterSetForm,
 )
+from netbox.forms.mixins import OwnerFilterMixin
 from tenancy.forms import ContactModelFilterForm, TenancyFilterForm
 from tenancy.models import Tenant
-from users.models import Owner, User
+from users.models import User
 from utilities.forms import BOOLEAN_WITH_BLANK_CHOICES, FilterForm, add_blank_choice
-from utilities.forms.fields import ColorField, DynamicModelChoiceField, DynamicModelMultipleChoiceField, TagFilterField
+from utilities.forms.fields import ColorField, DynamicModelMultipleChoiceField, TagFilterField
 from utilities.forms.rendering import FieldSet
 from utilities.forms.widgets import NumberWithOptions
 from virtualization.models import Cluster, ClusterGroup, VirtualMachine
@@ -70,11 +71,11 @@ __all__ = (
     'SiteFilterForm',
     'SiteGroupFilterForm',
     'VirtualChassisFilterForm',
-    'VirtualDeviceContextFilterForm'
+    'VirtualDeviceContextFilterForm',
 )
 
 
-class DeviceComponentFilterForm(NetBoxModelFilterSetForm):
+class DeviceComponentFilterForm(OwnerFilterMixin, NetBoxModelFilterSetForm):
     name = forms.CharField(
         label=_('Name'),
         required=False
@@ -157,18 +158,14 @@ class DeviceComponentFilterForm(NetBoxModelFilterSetForm):
         required=False,
         label=_('Device Status'),
     )
-    owner_id = DynamicModelChoiceField(
-        queryset=Owner.objects.all(),
-        required=False,
-        label=_('Owner'),
-    )
 
 
 class RegionFilterForm(ContactModelFilterForm, NestedGroupModelFilterSetForm):
     model = Region
     fieldsets = (
-        FieldSet('q', 'filter_id', 'tag', 'owner_id'),
+        FieldSet('q', 'filter_id', 'tag'),
         FieldSet('parent_id', name=_('Region')),
+        FieldSet('owner_group_id', 'owner_id', name=_('Ownership')),
         FieldSet('contact', 'contact_role', 'contact_group', name=_('Contacts'))
     )
     parent_id = DynamicModelMultipleChoiceField(
@@ -182,8 +179,9 @@ class RegionFilterForm(ContactModelFilterForm, NestedGroupModelFilterSetForm):
 class SiteGroupFilterForm(ContactModelFilterForm, NestedGroupModelFilterSetForm):
     model = SiteGroup
     fieldsets = (
-        FieldSet('q', 'filter_id', 'tag', 'owner_id'),
+        FieldSet('q', 'filter_id', 'tag'),
         FieldSet('parent_id', name=_('Site Group')),
+        FieldSet('owner_group_id', 'owner_id', name=_('Ownership')),
         FieldSet('contact', 'contact_role', 'contact_group', name=_('Contacts'))
     )
     parent_id = DynamicModelMultipleChoiceField(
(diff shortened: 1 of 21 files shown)


netbox/dcim/forms/filtersets.py
```diff
@@ -157,18 +157,29 @@ class DeviceComponentFilterForm(NetBoxModelFilterSetForm):
...
+    )
+    owner_id = DynamicModelMultipleChoiceField(
         queryset=Owner.objects.all(),
         required=False,
+        null_option='None',
+        query_params={
+            'group_id': '$owner_group_id'
+        },
         label=_('Owner'),
     )
```
jeremystretch: Can we inherit from `OwnerFilterMixin` here instead of defining these two fields locally? (I'm not sure why this wasn't done for `owner_id` originally.)

netbox/extras/forms/filtersets.py
```diff
@@ -119,9 +119,19 @@ class CustomFieldFilterForm(SavedFiltersMixin, FilterForm):
...
+    )
+    owner_id = DynamicModelMultipleChoiceField(
         queryset=Owner.objects.all(),
         required=False,
+        null_option='None',
+        query_params={
+            'group_id': '$owner_group_id'
+        },
         label=_('Owner'),
     )
```
jeremystretch: It looks like we're defining the `owner_group_id` and `owner_id` fields locally on several forms in this module. Seems like we should be to just inherit from OwnerFilterMixin instead. (Again, I'm not sure why this wasn't done originally.)

netbox/virtualization/forms/filtersets.py
```diff
@@ -259,9 +264,19 @@ class VMInterfaceFilterForm(NetBoxModelFilterSetForm):
         required=False,
         label=_('VLAN Translation Policy')
     )
-    owner_id = DynamicModelChoiceField(
+    owner_group_id = DynamicModelMultipleChoiceField(
```
jeremystretch: Another candidate for OwnerFilterMixin?

netbox/virtualization/forms/filtersets.py
```diff
@@ -284,9 +300,19 @@ class VirtualDiskFilterForm(NetBoxModelFilterSetForm):
         required=False,
         min_value=1
     )
-    owner_id = DynamicModelChoiceField(
+    owner_group_id = DynamicModelMultipleChoiceField(
```
jeremystretch: Another candidate for OwnerFilterMixin?

netbox/users/filterset_mixins.py
```diff
@@ -12,6 +12,17 @@ class OwnerFilterMixin(django_filters.FilterSet):
...
     Adds owner & owner_id filters for models which inherit from OwnerMixin.
     """
+    owner_group_id = django_filters.ModelMultipleChoiceFilter(
+        queryset=OwnerGroup.objects.all(),
+        field_name='owner__group',
+        label=_('Owner Group (ID)'),
+    )
+    owner_group = django_filters.ModelMultipleChoiceFilter(
+        queryset=OwnerGroup.objects.all(),
+        field_name='owner__group',
```
jeremystretch: `field_name` needs to target the `name` field specifically. (I'm not sure why this isn't inferred from `to_field_name`.)
```suggestion
        field_name='owner__group__name',
```

netbox/templates/dcim/virtualchassis_edit.html
```diff
@@ -38,6 +38,7 @@ <h2 class="col-9 offset-3">{% trans "Virtual Chassis" %}</h2>
         <div class="row">
           <h2 class="col-9 offset-3">{% trans "Owner" %}</h2>
```
jeremystretch: Just to stay consistent
```suggestion
          <h2 class="col-9 offset-3">{% trans "Ownership" %}</h2>
```

jeremystretch (changes requested):
Nice work @pheus! I think there are a few places we can avoid declaring the filter fields by inheriting from `OwnerFilterMixin`.

netbox/dcim/forms/filtersets.py
pheus: Thanks for the review! Good catch.

I did look at reusing `OwnerFilterMixin` from `netbox.forms.filtersets` during development. The main reason I didn’t lean on it everywhere is that it’s **not** currently exported via `__all__`, and I wasn’t sure whether making it part of the public surface area was intended.

From a Django app organization standpoint, it also feels like this mixin conceptually belongs closer to `users` (e.g. something like an `OwnerFilterFormMixin`), so I assumed there might be historical context here (imports/circular deps, etc.) and opted for a small amount of duplication to keep things safe.

That said, I’m happy to refactor these filtersets to inherit from `OwnerFilterMixin` (and/or introduce a _dedicated_ mixin under `users` if that’s preferred). Let me know which direction you’d like and I’ll update the PR accordingly.

netbox/users/filterset_mixins.py
pheus: Oh, that shouldn’t have happened. Sorry about that, and thanks for catching it!
I’ll update it accordingly.

netbox/templates/dcim/virtualchassis_edit.html
pheus: That's definitely an oversight. Thanks for catching it!

netbox/dcim/forms/filtersets.py
jeremystretch: Agreed, moving `OwnerFilterMixin` makes sense. Let's do that and try to use it consistently for all the forms.

pheus: Thanks! I tried moving it to `users.forms.mixins` as discussed.

Unfortunately that approach introduces a circular import in this case. To avoid that, I moved `OwnerFilterMixin` to `netbox.forms.mixins`, alongside `SavedFiltersMixin`, `OwnerMixin`, etc., and then updated the affected filter forms to use it consistently.

Hope this aligns with what you had in mind, but I’m happy to adjust if you’d prefer a different structure.

netbox/templates/generic/bulk_edit.html
```diff
@@ -62,7 +62,7 @@
           {% if form.owner %}
             <div class="field-group mb-5">
               <div class="row">
-                <h2 class="col-9 offset-3">{% trans "Owner" %}</h2>
+                <h2 class="col-9 offset-3">{% trans "Ownership" %}</h2>
               </div>
               {% render_field form.owner bulk_nullable=True %}
```
jeremystretch, line 70: Just noticed that we didn't add the `owner_group` field here to include the selector.

jeremystretch (changes requested):
Thanks @pheus, great work! Just one other little nit I just noticed while testing again.

netbox/templates/generic/bulk_edit.html
pheus: Good catch! Thanks.
I’ll add `owner_group` alongside `owner` in `generic/bulk_edit.html` so the selector can scope owners by group, consistent with the other forms.

jeremystretch (approved):

Issue #21262, opened 2026-01-22
Scope and Scope type does not  get clone to child prefix


pradiptama:
### NetBox Edition

NetBox Community

### NetBox Version

v4.2.9

### Python Version

3.12

### Steps to Reproduce

1. Create a Site named DC1 (Organization → Sites → Add).
2. Create a new Prefix (IPAM → Prefixes → Add), as below:
       Prefix: 10.0.0.0/20
       Status: Active (or any valid status)
       Scope type: Site
       Scope: DC1
3. Save the Prefix.
4. Create another Prefix (IPAM → Prefixes → Add), as below:
       Prefix: 10.0.0.0/24
       Status: Active (or any valid status)
       Scope type: Site
       Scope: DC1
5. Open the newly created Prefix’s 10.0.0.0/20 detail view.
6. Go to "Child Prefixes" tab
7. Click any available Prefix for example 10.0.1.0/24.
8. On the prefix form (initial GET):
        Scope type and scope is blank

### Expected Behavior

The new prefix form scope type and scope should be inherit from the parent prefix.

### Observed Behavior

The new prefix creation link is currently being generated as: /ipam/prefixes/add/?prefix=10.0.1.0/24\&vrf=\&site=.
It appears that the &site= parameter is a leftover from an older Netbox version and should be updated, as "site" has been replaced with "scope" in the current version.

Issue #21201, opened 2026-01-16
Pre-populate GenericForeignKey form fields when cloning


pheus:
### NetBox version

v4.5.0

### Feature type

Change to existing functionality

### Proposed functionality

When cloning an object via `CloningMixin`, only the raw database fields are currently copied into the clone “initial” data. For models which use a `GenericForeignKey` (GFK), this means the mixin clones fields like `parent_object_type` and `parent_object_id`, but **NetBox forms typically expose only a single field** named after the GFK attribute (e.g. `parent`) using a `DynamicModelChoiceField`.

As a result, the “Parent” selector remains empty when cloning, even though the underlying content type + object ID values were copied.

Enhance `CloningMixin.clone()` to automatically support `GenericForeignKey` attributes “out of the box”:

- Detect `GenericForeignKey` descriptors on the model (e.g. via Django’s model metadata / private fields).
- For each detected GFK:
  - If both constituent fields (content type field and object ID field) are present in `clone_fields` (e.g. `parent_object_type` and `parent_object_id`),
  - and the GFK attribute name is **not already** present in the clone dict,
  - then inject the GFK attribute name into the returned clone data:
    - Example: add `{"parent": instance.parent}`

This would allow forms which expose only the “combined” field (`parent`) to correctly pre-populate during cloning, without requiring per-form custom logic.

### Use case

Cloning is commonly used to speed up repetitive data entry. For any model that uses a `GenericForeignKey` exposed via a single form field (e.g. a “Parent” selector), the current cloning behavior forces users to manually re-select the parent every time.

Automatically populating the GFK attribute in `CloningMixin.clone()` would make cloning behave as users expect across all GFK-enabled models, consistently and without extra per-model code.

### Database changes

None.

### External dependencies

None.

PR #21265, opened 2026-01-22
Fixes #21202: Fix scoped form cloning clearing the Scope field when Scope Type changes

### LINKED ISSUE #21202: ScopedForm clears `scope` on clone because it doesn’t distinguish cloning from editing
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.0

### Python Version

3.12

### Steps to Reproduce

1. Create a Site named `DC1` (Organization → Sites → Add).
2. Create a new Prefix (IPAM → Prefixes → Add), for example:
   - Prefix: `10.0.0.0/24` (any prefix works)
   - Status: Active (or any valid status)
   - **Scope type**: `Site`
   - **Scope**: `DC1`
   - Save the Prefix.
3. Open the newly created Prefix’s detail view.
4. Click **Clone**.
5. On the clone form (initial **GET**):
   - **Scope type** remains set to `Site`
   - but **Scope** is cleared/blank (the original value `DC1` is not pre-filled), forcing the user to re-select `DC1` manually.

### Expected Behavior

When opening a clone form, if the URL includes valid `scope_type` and `scope` values, the form should preserve both initial values on the initial GET request. Cloning should not require the user to manually re-select the scope if it’s already provided.

### Observed Behavior

During the initial GET request of a **clone** operation, `ScopedForm._set_scoped_values()` clears `self.initial['scope']` whenever it detects a `scope_type` change. Because the logic doesn’t differentiate between **editing** an existing instance and **cloning** (a new instance), the scope is wiped even when the URL provides a correct `scope_type` and scope ID.

This forces the user to re-select the scope manually, which is easy to miss and makes cloning scoped objects more error-prone.

pheus:
### Fixes: #21202

This PR adjusts `ScopedForm._set_scoped_values()` so it no longer clears `initial['scope']` during clone form initialization.

Previously, when cloning an object with scoped fields (e.g., a Prefix scoped to `Site: DC1`), the form would treat the `scope_type` mismatch like an edit and wipe the pre-filled scope on the initial GET, forcing the user to re-select it manually.

With this change, the scope is only cleared when **editing an existing object** (`instance.pk` is set) *and* the `scope_type` has actually changed.

Thanks for taking a look! This also helps unblock the follow-up fix in #21262.

### DIFF
--- a/netbox/dcim/forms/mixins.py
+++ b/netbox/dcim/forms/mixins.py
@@ -75,7 +75,7 @@ class ScopedForm(forms.Form):
             except ObjectDoesNotExist:
                 pass
 
-            if self.instance and scope_type_id != self.instance.scope_type_id:
+            if self.instance and self.instance.pk and scope_type_id != self.instance.scope_type_id:
                 self.initial['scope'] = None
 
         else:


pheus:
Small follow-up note (so expectations are clear): this PR fixes the **form-side** behavior where `ScopedForm._set_scoped_values()` could wipe `initial['scope']` during clone initialization.

However, cloning a scoped Prefix still won’t fully pre-populate the **Scope** selector yet. The underlying reason is that `CloningMixin` currently copies only the raw database fields into the clone “initial” data (e.g. `scope_type` + `scope_id`). Since `scope` is a `GenericForeignKey` and NetBox forms typically expose only the combined field (`scope`) via a `DynamicModelChoiceField`, the form has nothing to pre-fill and the selector remains empty even though the constituent fields were cloned.

A generic fix to make `CloningMixin.clone()` inject the GFK attribute (e.g. `{"scope": instance.scope}` when `scope_type` + `scope_id` are present) is tracked in #21201.

Thanks again for taking a look!

jnovinger (approved):

PR #21270, opened 2026-01-23
Fixes: #19129 - Richer display of MAC addresses in InterfaceTable when multiple MACs are present

### LINKED ISSUE #19129: Rework multiple MAC UI representation
### NetBox version

v4.2.6

### Feature type

### Proposed functionality

Add way to know from UI list, when interface have any MACs, even if it have no "primary" MAC.
Add way to know if interface have multiple MACs, with highlighted "primary" MAC.

Add comprehensive way to remove mac from edit interface UI. Without long alternatives: API or MAC list filter.

### Use case

Right now, user have no way to know which interface have multiple MACs. And can't figure out on-the-go how to remove one.

Example:
Image

Image

### Database changes

_No response_

### External dependencies

_No response_

bctiemann:
### Fixes: #19129

On the list view for Interface and VMInterface, the presence of multiple MAC addresses on an interface is now communicated in various ways depending on whether one is assigned:

Screenshot 2026-01-22 at 9 58 26 PM

```
        11:22:33:44:55:66      <-- Single MAC address on interface, assigned as primary
        11:22:33:44:55:66 (2)  <-- Multiple MAC addresses on interface, one assigned as primary
        2 available            <-- 1 or more MAC addresses on interface, none assigned as primary
        -                      <-- No MAC addresses on interface
```

Note: this same logic is reflected on the interface detail page.


netbox/dcim/tables/devices.py
```diff
@@ -616,6 +616,7 @@ class BaseInterfaceTable(NetBoxTable):
     )
     primary_mac_address = tables.Column(
         verbose_name=_('MAC Address'),
+        accessor=Accessor('mac_address_display'),
```
jeremystretch: This adds significant overhead as currently implemented, triggering a count of MAC addresses for each interface in the table. Why don't we just introduce a non-default column that lists all assigned MAC addresses (perhaps annotating the primary), like we do for IP addresses? That seems much cleaner IMO, and it ensures the MAC addresses are efficiently prefetched.

netbox/templates/dcim/interface.html
```diff
@@ -143,9 +143,11 @@ <h2 class="card-header">{% trans "Addressing" %}</h2>
           <tr>
             <th scope="row">{% trans "MAC Address" %}</th>
             <td>
-              {% if object.primary_mac_address %}
-                <span class="font-monospace">{{ object.primary_mac_address|linkify }}</span>
-                <span class="badge text-bg-primary">{% trans "Primary" %}</span>
+              {% if object.mac_address_display %}
```
jeremystretch: Same general note as above. Rather than only showing the primary MAC and hinting whether there are other MACs assigned, let's just display all the MACs and annotate the primary.

jeremystretch (changes requested):

netbox/dcim/tables/devices.py
bctiemann: Agreed, that's a better pattern. Done, also synchronized the `vminterface.html` MAC Address field HTML with `interface.html` which was overlooked before.

netbox/dcim/tables/devices.py
```diff
@@ -583,6 +583,15 @@ class BaseInterfaceTable(NetBoxTable):
         orderable=False,
         verbose_name=_('IP Addresses')
     )
+    primary_mac_address = tables.Column(
+        verbose_name=_('MAC Address'),
```
jeremystretch: We should annotate this as primary somehow to avoid confusion with the multi-MAC column.
```suggestion
        verbose_name=_('MAC Address (Primary)'),
```

netbox/dcim/tables/devices.py
```diff
@@ -583,6 +583,15 @@ class BaseInterfaceTable(NetBoxTable):
...
+        linkify=True
+    )
+    mac_addresses = tables.TemplateColumn(
```
jeremystretch: IMO we should just make this a [`ManyToManyColumn`](https://github.com/netbox-community/netbox/blob/8e620ef325bc7ce09e357a313cca01aa93d7dd15/netbox/netbox/tables/columns.py#L121).

netbox/dcim/tables/template_code.py
```diff
@@ -62,6 +62,16 @@
   {% endif %}
 """
 
+INTERFACE_MACADDRESSES = """
+  {% if value.count > 3 %}
+    <a href="{% url 'ipam:macaddress_list' %}?{{ record|meta:"model_name" }}_id={{ record.pk }}">{{ value.count }}</a>
```
jeremystretch: If we keep the TemplateColumn
```suggestion
    <a href="{% url 'dcim:macaddress_list' %}?{{ record|meta:"model_name" }}_id={{ record.pk }}">{{ value.count }}</a>
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #21275, opened 2026-01-23
Closes #21209: Support real model names in configuration parameters

### LINKED ISSUE #21209: Use real model names in dashboard widget and other configurations
### NetBox version

v4.5.0

### Feature type

### Proposed functionality

Currently [`DEFAULT_DASHBOARD`](https://netbox.readthedocs.io/en/stable/configuration/default-values/#default_dashboard) and [`PROTECTION_RULES`](https://netbox.readthedocs.io/en/stable/configuration/data-validation/#protection_rules) (probably some others as well) configurations take lowercase model names (like `"ipam.ipaddress"`), while [`FIELD_CHOICES`](https://netbox.readthedocs.io/en/stable/configuration/data-validation/#field_choices) is built on top of the real model names like `"ipam.IPAddress"`.

I'd suggest unifying the syntaxes, maybe by using the real model names, as that syntax needs to be used in NetBox scripts anyway.

### Use case

When configuring NetBox and also creating scripts, it's confusing that the model names are referenced in different syntax in different places.

### Database changes

None

### External dependencies

None

adionit7:
## Summary

This PR adds support for real model names (e.g., `dcim.Site`) in `DEFAULT_DASHBOARD`, `CUSTOM_VALIDATORS`, and `PROTECTION_RULES` configuration parameters, matching the syntax used by `FIELD_CHOICES`.

### Changes:
- Add `normalize_model_identifier()` utility function to convert real model names to lowercase format used by ObjectType
- Update `get_models_from_content_types()` in dashboard widgets to normalize model identifiers
- Update `CUSTOM_VALIDATORS` and `PROTECTION_RULES` lookups to check both real model names and lowercase names
- Update documentation with real model name examples and backward compatibility notes

### Backward Compatibility
Both formats are supported:
- Real model names: `dcim.Site`, `ipam.IPAddress` 
- Lowercase names: `dcim.site`, `ipam.ipaddress` (existing behavior preserved)

Fixes #21209

### DIFF
--- a/docs/configuration/data-validation.md
+++ b/docs/configuration/data-validation.md
@@ -8,7 +8,7 @@ This is a mapping of models to [custom validators](../customization/custom-valid
 
 ```python
 CUSTOM_VALIDATORS = {
-    "dcim.site": [
+    "dcim.Site": [
         {
             "name": {
                 "min_length": 5,
@@ -17,11 +17,14 @@ CUSTOM_VALIDATORS = {
         },
         "my_plugin.validators.Validator1"
     ],
-    "dcim.device": [
+    "dcim.Device": [
         "my_plugin.validators.Validator1"
     ]
 }
 ```
+
+!!! info "Case-Insensitive Model Names"
+    Model identifiers are case-insensitive. Both `dcim.site` and `dcim.Site` are valid and equivalent.
 
 ---
 
@@ -52,6 +55,9 @@ FIELD_CHOICES = {
     )
 }
 ```
+
+!!! info "Case-Insensitive Field Identifiers"
+    Field identifiers are case-insensitive. Both `dcim.Site.status` and `dcim.site.status` are valid and equivalent.
 
 The following model fields support configurable choices:
 
@@ -98,7 +104,7 @@ This is a mapping of models to [custom validators](../customization/custom-valid
 
 ```python
 PROTECTION_RULES = {
-    "dcim.site": [
+    "dcim.Site": [
         {
             "status": {
                 "eq": "decommissioning"
@@ -108,3 +114,6 @@ PROTECTION_RULES = {
     ]
 }
 ```
+
+!!! info "Case-Insensitive Model Names"
+    Model identifiers are case-insensitive. Both `dcim.site` and `dcim.Site` are valid and equivalent.
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -16,8 +16,9 @@ from core.events import *
 from core.models import ObjectType
 from extras.events import enqueue_event
 from extras.models import Tag
-from extras.utils import get_validators_for_model, run_validators
+from extras.utils import run_validators
 from netbox.config import get_config
+from utilities.data import get_config_value_ci
 from netbox.context import current_request, events_queue
 from netbox.models.features import ChangeLoggingMixin, get_model_features, model_is_public
 from utilities.exceptions import AbortRequest
@@ -168,7 +169,7 @@ def handle_deleted_object(sender, instance, **kwargs):
     # to queueing any events for the object being deleted, in case a validation error is
     # raised, causing the deletion to fail.
     model_name = f'{sender._meta.app_label}.{sender._meta.model_name}'
-    validators = get_validators_for_model(get_config().PROTECTION_RULES, model_name)
+    validators = get_config_value_ci(get_config().PROTECTION_RULES, model_name, default=[])
     try:
         run_validators(instance, validators)
     except ValidationError as e:
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -23,6 +23,22 @@ from netbox.models.features import ChangeLoggingMixin, get_model_features, model
 from utilities.exceptions import AbortRequest
 from .models import ConfigRevision, DataSource, ObjectChange
 
+
+def _get_validators_for_model(config_dict, model_name):
+    """
+    Retrieve validators from config dict with case-insensitive model name lookup.
+    Supports both lowercase (e.g. "dcim.site") and PascalCase (e.g. "dcim.Site") keys.
+    """
+    # Try exact match first
+    if model_name in config_dict:
+        return config_dict[model_name]
+    # Try case-insensitive match
+    for key, value in config_dict.items():
+        if key.lower() == model_name:
+            return value
+    return []
+
+
 __all__ = (
     'clear_events',
     'job_end',
@@ -168,7 +184,7 @@ def handle_deleted_object(sender, instance, **kwargs):
     # to queueing any events for the object being deleted, in case a validation error is
     # raised, causing the deletion to fail.
     model_name = f'{sender._meta.app_label}.{sender._meta.model_name}'
-    validators = get_config().PROTECTION_RULES.get(model_name, [])
+    validators = _get_validators_for_model(get_config().PROTECTION_RULES, model_name)
     try:
         run_validators(instance, validators)
     except ValidationError as e:
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -16,28 +16,12 @@ from core.events import *
 from core.models import ObjectType
 from extras.events import enqueue_event
 from extras.models import Tag
-from extras.utils import run_validators
+from extras.utils import get_validators_for_model, run_validators
 from netbox.config import get_config
 from netbox.context import current_request, events_queue
 from netbox.models.features import ChangeLoggingMixin, get_model_features, model_is_public
 from utilities.exceptions import AbortRequest
 from .models import ConfigRevision, DataSource, ObjectChange
-
-
-def _get_validators_for_model(config_dict, model_name):
-    """
-    Retrieve validators from config dict with case-insensitive model name lookup.
-    Supports both lowercase (e.g. "dcim.site") and PascalCase (e.g. "dcim.Site") keys.
-    """
-    # Try exact match first
-    if model_name in config_dict:
-        return config_dict[model_name]
-    # Try case-insensitive match
-    for key, value in config_dict.items():
-        if key.lower() == model_name:
-            return value
-    return []
-
 
 __all__ = (
     'clear_events',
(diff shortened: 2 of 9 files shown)


netbox/core/signals.py
```diff
@@ -23,6 +23,22 @@
...
+    # Try exact match first
+    if model_name in config_dict:
+        return config_dict[model_name]
+    # Try case-insensitive match
+    for key, value in config_dict.items():
+        if key.lower() == model_name:
+            return value
+    return []
+
+
```
pheus: Just a quick drive-by observation: I noticed this code block is duplicated. We usually try to avoid duplication where possible, since it can drift over time and makes future changes a bit harder.

pheus:
Thanks a lot for working on this! 🙌

While reviewing the feature request, I noticed the scope is a bit broader than what’s currently covered in this PR. The request is to allow ContentType/ObjectType strings (e.g. `dcim.Device`) to be case-insensitive across *all* relevant configuration parameters. The FR author specifically called out three areas:

- `PROTECTION_RULES`
- `DEFAULT_DASHBOARD`
- `FIELD_CHOICES`

From what I can see, this PR currently addresses only the `PROTECTION_RULES` and the `DEFAULT_DASHBOARD` case.

Would you mind taking another pass to review the other config paths as well, and extend the PR to include any additional configuration options where these model strings are accepted?

adionit7:
> Thanks a lot for working on this! 🙌
> 
> While reviewing the feature request, I noticed the scope is a bit broader than what’s currently covered in this PR. The request is to allow ContentType/ObjectType strings (e.g. `dcim.Device`) to be case-insensitive across _all_ relevant configuration parameters. The FR author specifically called out three areas:
> 
> * `PROTECTION_RULES`
> * `DEFAULT_DASHBOARD`
> * `FIELD_CHOICES`
> 
> From what I can see, this PR currently addresses only the `PROTECTION_RULES` and the `DEFAULT_DASHBOARD` case.
> 
> Would you mind taking another pass to review the other config paths as well, and extend the PR to include any additional configuration options where these model strings are accepted?

Thanks for the review @pheus! Really appreciate it! 🙌
You're right.. we initially missed FIELD_CHOICES. I've added case-insensitive support for it in utilities/choices.py using the same pattern.

The PR now covers all three areas mentioned in the FR:
- DEFAULT_DASHBOARD — accepts both dcim.site and dcim.Site
-  PROTECTION_RULES — case-insensitive lookup via helper function
-  FIELD_CHOICES — case-insensitive lookup for both dcim.Site.status and dcim.site.status (and the + extension variant)

As a bonus, I also updated CUSTOM_VALIDATORS to use the same case-insensitive pattern for consistency.

All changes follow the same approach: try an exact match first (O(1)), then fall back to a case-insensitive search if needed. This maintains backward compatibility while allowing users to use either format.

The PR is ready for another review.

netbox/extras/dashboard/widgets.py
```diff
@@ -75,10 +75,11 @@ def get_bookmarks_object_type_choices():
 def get_models_from_content_types(content_types):
     """
     Return a list of models corresponding to the given content types, identified by natural key.
+    Accepts both lowercase (e.g. "dcim.site") and PascalCase (e.g. "dcim.Site") model names.
```
jnovinger, line 78: Little nit, I think it's more accurate to say this is case-insensitive, give that `DCIM.sItE` would also now work, correct?

Maybe note something to publicize, but maybe worth noting in a code comment.

netbox/utilities/choices.py
```diff
@@ -11,6 +11,21 @@
...
+    Supports both lowercase (e.g. "dcim.site.status") and PascalCase (e.g. "dcim.Site.status") keys.
+    """
+    # Try exact match first
+    if key in config_dict:
+        return config_dict[key]
+    # Try case-insensitive match
+    for config_key, value in config_dict.items():
+        if config_key.lower() == key.lower():
+            return value
+    return None
```
jnovinger: This appears to be almost exactly the same as `extras.get_validators_for_model()`, except for the return value.

I think it would make sense to combine these, with a configurable default return value. My concern stems from when (in the future) we need to change one and miss changing the other to match.

jnovinger: FWIW, let's stick with keeping the util function in this module, with a suitably generic but still informative name.

jnovinger: We should also add tests for this.

jnovinger (dismissed):
Left some notes, but this is a great start. Also, you need to make sure you update the docs (as noted in the PR description).

netbox/extras/dashboard/widgets.py
jnovinger: Not an issue anymore, assuming we merge the util functions.

netbox/utilities/choices.py
adionit7: Thanks for the review @jnovinger! I've addressed your comments:

- Created get_config_value_ci() in utilities/choices.py with a configurable default parameter
- Updated get_validators_for_model() in extras/utils.py to use the shared function
- Added tests for the utility function
- Updated the docs with case-insensitivity notes for CUSTOM_VALIDATORS, FIELD_CHOICES, and PROTECTION_RULES

ready for your review!

docs/configuration/data-validation.md
pheus: Could you please unify the style throughout this document?
`dcim.site` -> `dcim.Site`

netbox/extras/utils.py
```diff
@@ -131,6 +132,13 @@ def is_report(obj):
         return False
 
 
+def get_validators_for_model(config_dict, model_name):
+    """
+    Retrieve validators from config dict with case-insensitive model name lookup.
+    """
+    return get_config_value_ci(config_dict, model_name, default=[])
+
+
```
pheus: Right now `get_validators_for_model()` appears to be a very thin wrapper around `get_config_value_ci(..., default=[])` without adding extra semantics (normalization, validation, error handling, etc.). That makes it harder to justify the extra function + import surface.

Could we either:
- (preferred) **Inline** the `get_config_value_ci(..., default=[])` call at the call sites and drop `get_validators_for_model()`, **or**
- **Enhance** `get_validators_for_model()` so it adds real value (e.g. ensures a list is returned, validates expected types/shape, centralizes key construction, raises a clearer error for invalid config values).

netbox/utilities/choices.py
```diff
@@ -7,10 +7,24 @@
...
+    Retrieve a value from a config dict with case-insensitive key lookup.
+    """
+    if key in config_dict:
+        return config_dict[key]
+    key_lower = key.lower()
+    for config_key, value in config_dict.items():
+        if config_key.lower() == key_lower:
+            return value
+    return default
+
```
pheus: `get_config_value_ci()` is a general-purpose “case-insensitive dict get” helper and is now used outside the choices context. Placing it in `choices.py` makes dependencies a bit surprising.

I’d suggest **move it** to a more general utilities module.

netbox/utilities/choices.py
```diff
@@ -24,13 +38,15 @@ def __new__(mcs, name, bases, attrs):
...
+            extend_choices = get_config_value_ci(settings.FIELD_CHOICES, extend_key)
+            if replace_choices is not None:
                 # Replace the stock choices
-                attrs['CHOICES'] = settings.FIELD_CHOICES[replace_key]
-            elif extend_key and extend_key in settings.FIELD_CHOICES:
+                attrs['CHOICES'] = replace_choices
+            elif extend_choices is not None:
                 # Extend the stock choices
-                attrs['CHOICES'].extend(settings.FIELD_CHOICES[extend_key])
+                attrs['CHOICES'].extend(extend_choices)
```
pheus, line 36: Minor readability: avoid looking up `extend_choices` if `replace_choices` matched
If `replace_choices` is present, we never use `extend_choices`, so it reads cleaner to compute it only in the `else` branch. For example:

```python
replace_choices = get_config_value_ci(settings.FIELD_CHOICES, replace_key)
if replace_choices is not None:
    attrs["CHOICES"] = replace_choices
else:
    extend_choices = get_config_value_ci(settings.FIELD_CHOICES, extend_key)
    if extend_choices is not None:
        attrs["CHOICES"].extend(extend_choices)
```

pheus (changes requested):
Thanks for pushing this forward! I appreciate that you’re applying it consistently across `FIELD_CHOICES` and `CUSTOM_VALIDATORS`.

A few follow-ups I’d like to see addressed

#### Tests: please add an integration-style test for `FIELD_CHOICES` casing + `+` suffix
The unit tests for `get_config_value_ci()` are helpful, but I’d like to see at least one test that proves:
- replacement works when config key casing differs (e.g. `dcim.Site.status`), and
- extension works with the `+` suffix under casing differences (e.g. `dcim.Site.status+`).

That will ensure the original reported issue stays fixed and avoids future regressions.

docs/configuration/data-validation.md
adionit7: Done

netbox/extras/utils.py
adionit7: Inlined at both call sites (`extras/signals.py` and `core/signals.py`) and removed the wrapper function.

netbox/utilities/choices.py
adionit7: Moved to utilities/data.py under the "Dictionary utilities" section.

netbox/utilities/choices.py
adionit7: Restructured to only compute `extend_key` and `extend_choices` inside the else branch.

adionit7:
Thanks for the feedback @pheus! Appreciate that.

I've added integration tests in `test_choices.py` that verify:
- Replacement works when config key casing differs (lowercase config key `utilities.teststatus `found when code looks up PascalCase `utilities.TestStatus`)
- Extension works with + suffix under casing differences.

Other comments are also resolved as per the feedback. Ready for your review! :)

pheus (approved):
Thanks for the updates! I re-reviewed the latest changes.

Nice cleanup 👍

Overall, LGTM — thanks for iterating on this and for adding solid tests.

adionit7:
> Thanks for the updates! I re-reviewed the latest changes.
> 
> Nice cleanup 👍
> 
> Overall, LGTM — thanks for iterating on this and for adding solid tests.

Thanks a lot for the review @pheus! 
Happy to help :)

PR #21276, opened 2026-01-23
Closes #21228: Add image attachments support to RackType model

### LINKED ISSUE #21228: Add image attachments to Rack Type model
### NetBox version

v4.4.6

### Feature type

### Proposed functionality

Rack Types should accept image attachments, similarly to Device Types and other models.

### Use case

It would be helpful to show details of specialized racks, such as sound-proofing and flat racks, and dimensional drawings, etc.  I think this is a one-line change, just adding ImageAttachmentMixin.

### Database changes

_No response_

### External dependencies

_No response_

adionit7:
## Summary

Enable image attachments for the RackType model, similar to other models like DeviceType, ModuleType, Site, and Location.

### Changes:
- Add `ImageAttachmentsMixin` to `RackType` model
- Add `ImageAttachmentsMixin` to `RackTypeType` GraphQL type  
- Add `ImageAttachmentFilterMixin` to `RackTypeFilter` GraphQL filter

Fixes #21228

### DIFF
--- a/netbox/dcim/graphql/filters.py
+++ b/netbox/dcim/graphql/filters.py
@@ -889,7 +889,7 @@ class PowerPortTemplateFilter(ModularComponentTemplateFilterMixin, ChangeLoggedM
 
 
 @strawberry_django.filter_type(models.RackType, lookups=True)
-class RackTypeFilter(RackFilterMixin, WeightFilterMixin, PrimaryModelFilter):
+class RackTypeFilter(ImageAttachmentFilterMixin, RackFilterMixin, WeightFilterMixin, PrimaryModelFilter):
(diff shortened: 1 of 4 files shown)


arthanson (changes requested):
in views.py you need to add in the:
            ImageAttachmentsPanel(),
so it appears in the detail view.

adionit7:
> in views.py you need to add in the: ImageAttachmentsPanel(), so it appears in the detail view.

Good catch @arthanson, thanks for the review!

I've added `ImageAttachmentsPanel()` to the `right_panels` in `RackTypeView` so image attachments will now appear in the detail view, consistent with other models like DeviceType and Rack.

arthanson (approved):
Thanks @adionit7 !

PR #21282, opened 2026-01-23
Fixes #21254: Fix release check failure when stale `latest_release` cache can't be unpickled

### LINKED ISSUE #21254: AttributeError: 'Version' object has no attribute 'dict' with packaging 26.0
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.1

### Python Version

3.12

### Steps to Reproduce

1. Upgrade to NetBox 4.5.1 from 4.5.0 (no plugins installed)
2. Run ./upgrade.sh (which installs latest dependencies, including packaging==26.0 as a transitive dependency)
3. Restart NetBox services
4. Access NetBox in browser and the following error occurs:
```
<class 'AttributeError'>
'Version' object has no attribute `__dict__`
```

### Expected Behavior

NetBox loads normally after upgrade without errors.

### Observed Behavior

Error:  
```
<class 'AttributeError'>
'Version' object has no attribute `__dict__`
```

The packaging library version 26.0 added `__slots__` to the Version class (https://github.com/pypa/packaging/pull/1001), which removes the `__dict__`  attribute. This breaks any code that attempts to access Version().`__dict__`.

packaging is a transitive dependency installed via gunicorn, mkdocs, and/or strawberry-graphql.

The workaround I used to get it working:
```
cd /opt/netbox
source venv/bin/activate
pip install 'packaging<26.0'
sudo systemctl restart netbox netbox-rq
```

pheus:
### Fixes: #21254

This PR prevents a 500 on the home page when reading the `latest_release` cache entry fails to unpickle after dependency upgrades (e.g. `packaging>=26` with a value cached under `<26`).

**Summary of changes**
- Wrap `cache.get('latest_release')` in a `try/except` during the superuser release check
- On failure, log at debug level, delete the stale cache key, and continue as if no cached release is available

This keeps the UI working and allows the cache to be repopulated normally by housekeeping/release checks.

### DIFF
--- a/netbox/netbox/views/misc.py
+++ b/netbox/netbox/views/misc.py
@@ -1,5 +1,6 @@
 import re
 from collections import namedtuple
+import logging
 
 from django.conf import settings
 from django.contrib import messages
@@ -28,6 +29,8 @@ __all__ = (
     'SearchView',
 )
 
+logger = logging.getLogger(f'netbox.{__name__}')
+
 Link = namedtuple('Link', ('label', 'viewname', 'permission', 'count'))
 
 
@@ -50,7 +53,14 @@ class HomeView(ConditionalLoginRequiredMixin, View):
         # Check whether a new release is available. (Only for superusers.)
         new_release = None
         if request.user.is_superuser:
-            latest_release = cache.get('latest_release')
+            # cache.get() can raise if the cached value can't be unpickled after dependency upgrades
+            try:
+                latest_release = cache.get('latest_release')
+            except Exception:
+                logger.debug("Failed to read 'latest_release' from cache; deleting key", exc_info=True)
+                cache.delete('latest_release')
+                latest_release = None
+
             if latest_release:
                 release_version, release_url = latest_release
                 if release_version > version.parse(settings.RELEASE.version):
--- a/netbox/netbox/views/misc.py
+++ b/netbox/netbox/views/misc.py
@@ -53,7 +53,7 @@ class HomeView(ConditionalLoginRequiredMixin, View):
         # Check whether a new release is available. (Only for superusers.)
         new_release = None
         if request.user.is_superuser:
-            # cache.get() can raise if the cached value can't be unpickled after dependency upgrades
+            # cache.get() can raise an exception if the cached value can't be unpickled after dependency upgrades
             try:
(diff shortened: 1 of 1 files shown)


netbox/netbox/views/misc.py
```diff
@@ -50,7 +53,14 @@ def get(self, request):
         # Check whether a new release is available. (Only for superusers.)
         new_release = None
         if request.user.is_superuser:
-            latest_release = cache.get('latest_release')
+            # cache.get() can raise if the cached value can't be unpickled after dependency upgrades
```
jeremystretch:
```suggestion
            # cache.get() can raise an exception if the cached value can't be unpickled after dependency upgrades
```

netbox/netbox/views/misc.py
```diff
@@ -50,7 +53,14 @@ def get(self, request):
...
+            try:
+                latest_release = cache.get('latest_release')
+            except Exception:
```
jeremystretch, line 59: Is there a more specific exception we can catch here?

jeremystretch (changes requested):

netbox/netbox/views/misc.py
pheus: Thanks for the review!

In the reports I’ve seen, the failure shows up as an `AttributeError` (`'Version' object has no attribute '__dict__'`) during unpickling, so we can definitely narrow the `except` to that.

That said, since the exception is raised while deserializing the cached value, I think it’s also plausible to hit other unpickling/deserialization errors depending on the backend/serializer/state (e.g. `pickle.UnpicklingError`, `EOFError`, `TypeError`, `ValueError`). My intent with catching `Exception` was simply to ensure a stale/bad `latest_release` entry can’t take down `HomeView`: if we can’t read it, treating it as a cache miss (and deleting the key) seems like the safest behavior.

Happy to adjust either way. Would you prefer we catch only `AttributeError`, or expand it to a small explicit set like:
`(AttributeError, pickle.UnpicklingError, EOFError, TypeError, ValueError)`?

jeremystretch: You're right to point out that we could see a variety of exceptions here. I'm fine to leave it as-is.

jeremystretch (approved):

PR #21286, opened 2026-01-23
Closes #21260: Defer object serialization for events pipeline

### LINKED ISSUE #21260: Defer object serialization for events pipeline
### NetBox Version

v4.5.1

### Python Version

3.12

### Area(s) of Concern

- [x] User Interface
- [x] REST API
- [x] GraphQL API
- [ ] Python ORM
- [ ] Other

### Details

When an object is created or modified in NetBox, its changes are captured and queued for event processing. The flow is outlined below. (Note that although we're focusing on adds & changes here for simplicity, a very similar flow happens for object deletions as well.)

1. [`handle_changed_object()`](https://github.com/netbox-community/netbox/blob/077d9b11294028b5e444a4c088058d18008533db/netbox/core/signals.py#L86) is triggered by the `post_save()` signal.
2. An ObjectChange record is created and saved to the database. This contains only a minimal serialized copy of the object (foreign keys are note resolved to related objects).
3. [`enqueue_event()`](https://github.com/netbox-community/netbox/blob/077d9b11294028b5e444a4c088058d18008533db/netbox/extras/events.py#L54) is called to add the instance to the request's event queue.
4. The instance is serialized by [`serialize_for_event()`](https://github.com/netbox-community/netbox/blob/077d9b11294028b5e444a4c088058d18008533db/netbox/extras/events.py#L26) to its _complete_ JSON representation (using the its REST API serializer). This representation and associated metadata is added to the events queue.
5. Once the response has been generated, the events queue is [flushed](https://github.com/netbox-community/netbox/blob/077d9b11294028b5e444a4c088058d18008533db/netbox/extras/events.py#L199) to the handlers registered in [`EVENTS_PIPELINE`](https://netboxlabs.com/docs/netbox/configuration/miscellaneous/#events_pipeline). (By default, this includes only `extras.events.process_event_queue`.)
6. [`process_event_queue()`](https://github.com/netbox-community/netbox/blob/077d9b11294028b5e444a4c088058d18008533db/netbox/extras/events.py#L169) iterates through the events in the queue and executed any applicable event rules for each.

The JSON serialization performed at step 4 can be quite expensive, as the object's complete representation (including related objects) is assembled from the database. This occurs even if the serialized data is ultimately never needed, i.e. because there are no event rules registered for the object type.

We can skip the serialization process and its related database queries by deferring it until `process_event_queue()` executes and finds at least one suitable event rule. This would be a breaking change as `event['data']` would presumably no longer be defined in the queue for other processors, but we might be able to devise a backward-compatible solution.

jeremystretch:
### Closes: #21260

- Introduce `EventContext` custom dictionary to populate serialized data on-demand
- Update `enqueue_event()`
    - Use `EventContext` instances instead of raw dictionaries
    - Add `object` and `user` as additional context
    - Force object serialization for deletion events
- Simplify signature of `process_event_rules()` to accept an event dict

### DIFF
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -90,6 +90,7 @@ def enqueue_event(queue, instance, request, event_type):
             event_type=event_type,
             snapshots=get_snapshots(instance, event_type),
             request=request,
+            user=request.user,
             # Legacy request attributes for backward compatibility
             username=request.user.username,
             request_id=request.id,
@@ -152,7 +153,7 @@ def process_event_rules(event_rules, object_type, event):
             params = {
                 "instance": event_rule.action_object,
                 "name": script.name,
-                "user": event['request'].user if 'request' in event else None,
+                "user": event['user'],
                 "data": event_data
             }
             if 'snapshots' in event:
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -27,11 +27,13 @@ class EventContext(UserDict):
     A custom dictionary that automatically serializes its associated object on demand.
     """
 
+    # We're emulating a dictionary here (rather than using a custom class) because prior to NetBox v4.5.2, events were
+    # queued as dictionaries for processing by handles in EVENTS_PIPELINE. We need to avoid introducing any breaking
+    # changes until a suitable minor release.
     def __getitem__(self, item):
         if item == 'data' and 'data' not in self:
             data = serialize_for_event(self['object'])
             self.__setitem__('data', data)
-            return data
         return super().__getitem__(item)
 
 
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -130,13 +130,11 @@ def process_event_rules(event_rules, object_type, event):
                 "object_type": object_type,
                 "event_type": event['event_type'],
                 "data": event_data,
-                "snapshots": event['snapshots'],
+                "snapshots": event.get('snapshots'),
                 "timestamp": timezone.now().isoformat(),
                 "username": event['username'],
                 "retry": get_rq_retry()
             }
-            if 'snapshots' in event:
-                params['snapshots'] = event['snapshots']
             if 'request' in event:
                 # Exclude FILES - webhooks don't need uploaded files,
                 # which can cause pickle errors with Pillow.
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -1,5 +1,5 @@
 import logging
-from collections import defaultdict
+from collections import UserDict, defaultdict
 
 from django.conf import settings
 from django.utils import timezone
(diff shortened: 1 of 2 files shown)


jeremystretch:
Some quick benchmarks creating 100 IP addresses via the REST API:

| | `main` | `a6c6a58` |
|-|-|-|
| Response time | 2.457s | 1.999s |
| Total queries | 3058 | 2657 |

netbox/extras/events.py
```diff
@@ -66,37 +80,42 @@ def enqueue_event(queue, instance, request, event_type):
...
+    if event_type == OBJECT_DELETED:
+        queue[key]['data'] = serialize_for_event(instance)
 
 
-def process_event_rules(event_rules, object_type, event_type, data, username=None, snapshots=None, request=None):
-    user = None  # To be resolved from the username if needed
+def process_event_rules(event_rules, object_type, event):
+    """
+    Process a list of EventRules against an event.
+    """
```
jnovinger, line 108: Just want to be real clear about this change, this is not part of a "published" or supported API, right? That is, we're not expecting this to cause problems for plugins or script authors, correct?

netbox/extras/events.py
```diff
@@ -109,50 +128,43 @@ def process_event_rules(event_rules, object_type, event_type, data, username=Non
...
                 "timestamp": timezone.now().isoformat(),
-                "username": username,
+                "username": event['username'],
                 "retry": get_rq_retry()
             }
-            if snapshots:
-                params["snapshots"] = snapshots
-            if request:
+            if 'snapshots' in event:
+                params['snapshots'] = event['snapshots']
```
jnovinger: Are these lines redundant with line 133 above?

netbox/extras/events.py
jeremystretch: Correct: `process_event_rules()` is for internal use only (for the "event rules" core feature). It's only called by `process_event_queue()` (the default `EVENTS_PIPELINE` handler) and by the `job_start`/`job_end` handlers.

netbox/extras/events.py
jeremystretch: It would appear so. Good catch. I was being so careful not to fiddle with the existing logic I didn't even see that.

netbox/extras/events.py
```diff
@@ -66,37 +80,42 @@ def enqueue_event(queue, instance, request, event_type):
...
+            snapshots=get_snapshots(instance, event_type),
+            request=request,
+            user=request.user,
             # Legacy request attributes for backward compatibility
-            'username': request.user.username,
-            'request_id': request.id,
-        }
+            username=request.user.username,
+            request_id=request.id,
+        )
```
jnovinger, line 99: This one seems like a really low likelihood of happening, but do you think we should be concerned about the addition of the `object` and `user` keys here, given that in theory it could change behavior of custom event handlers?

The only case I can see is if they iterate the keys in the item for some reason, and even then it seems unlikely that finding a key they didn't expect before would cause a problem.

jeremystretch: IMO it shouldn't be considered breaking to _introduce_ new keys to a data structure that we own, so I think it's okay. It's good to call it out though.

jnovinger (approved):

PR #21287, opened 2026-01-23
Closes #21259: Cache ObjectType results for the duration of a request

jeremystretch:
### Closes: #21259

- Introduce the `object_types_cache` context variable to
- Extend the `get_for_model()` method on ObjectTypeManager to check & populate the cache when resolving ObjectTypes
- Extend the `event_tracking()` context manager to initialize and clear `object_types_cache` for each request

### DIFF
--- a/netbox/core/models/object_types.py
+++ b/netbox/core/models/object_types.py
@@ -9,7 +9,7 @@ from django.db import connection, models
 from django.db.models import Q
 from django.utils.translation import gettext as _
 
-from netbox.context import object_types_cache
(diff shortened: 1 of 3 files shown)


jeremystretch:
Some quick benchmarks creating 100 IP addresses via the REST API:

| | `main` | `75193ee` |
|-|-|-|
| Response time | 2.457s | 2.027s |
| Total queries | 3058 | 2357 |

arthanson (approved):

PR #21295, opened 2026-01-26
Fixes #20977: Missing script variables use their default value if defined

### LINKED ISSUE #20977: Run Again button does not respect script variable defaults
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.6

### Python Version

3.12

### Steps to Reproduce

1. Create a custom script with `BooleanVar(default=True)`
2. Run that script using the "Run again" button on the script list page.

image

### Expected Behavior

It should run the script with all variables set to their default value.

### Observed Behavior

The BooleanVar value is False

llamafilm:
### Fixes: #20977

* When a variable is missing from a custom script submission, it will be filled with its default value, if defined. This is most useful for BooleanVars when clicking the "Run Again" button, but would also apply to other variables if left blank in the UI.

* Add a test case for this.

* Remove the default value from `test_script_validation_error_displays_message` so it properly displays a validation error.

### DIFF
--- a/netbox/extras/scripts.py
+++ b/netbox/extras/scripts.py
@@ -61,7 +61,7 @@ class ScriptVariable:
             self.field_attrs['label'] = label
         if description:
             self.field_attrs['help_text'] = description
-        if default:
+        if default is not None:
             self.field_attrs['initial'] = default
         if widget:
             self.field_attrs['widget'] = widget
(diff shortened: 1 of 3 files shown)


pheus (approved):
Thanks, @llamafilm !

PR #21306, opened 2026-01-27
Fixes #21262: Fix scope inheritance when creating child prefixes

### LINKED ISSUE #21262: Scope and Scope type does not  get clone to child prefix

pheus:
### Fixes: #21262

This PR fixes an issue where the **"Add Prefix"** button in the **Child Prefixes** view was still generating a legacy `site` query parameter, which caused the prefix creation form to open with **blank Scope Type / Scope**.

#### Summary of changes
- Update the "Add Prefix" button URL to pass `scope_type` and `scope` instead of the deprecated `site` parameter

Thanks again for your time and for reviewing!

### DIFF
--- a/netbox/ipam/tables/template_code.py
+++ b/netbox/ipam/tables/template_code.py
@@ -6,7 +6,7 @@ PREFIX_LINK = """
 {% if record.pk %}
   <a href="{{ record.get_absolute_url }}" id="prefix_{{ record.pk }}">{{ record.prefix }}</a>
 {% else %}
-  <a href="{% url 'ipam:prefix_add' %}?prefix={{ record }}{% if object.vrf %}&vrf={{ object.vrf.pk }}{% endif %}{% if object.site %}&site={{ object.site.pk }}{% endif %}{% if object.tenant %}&tenant_group={{ object.tenant.group.pk }}&tenant={{ object.tenant.pk }}{% endif %}">{{ record.prefix }}</a>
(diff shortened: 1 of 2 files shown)


jeremystretch (changes requested):
This fixes the "add prefix" button but we also need to address the in-table links to create available prefixes:

Image

I believe this can be fixed by updating the [`PREFIX_LINK`](https://github.com/netbox-community/netbox/blob/5a36e79215a419ef82bf8fe34a579e1c733f5042/netbox/ipam/tables/template_code.py#L5) template code. (I _think_ that's the only other place where it needs to be fixed.)

pheus:
Thanks for the review and sorry I didn’t catch that initially. I checked the templates where `CachedScopeMixin` is applied, but I missed the shared template code used for those in-table links.

I went back through everything and you’re absolutely right - that was the last missing piece. I’ve updated the `PREFIX_LINK` template accordingly.

Thanks a lot for pointing it out!

jeremystretch (approved):
Thanks @pheus!

PR #21309, opened 2026-01-27
21129 Store queue_name in Job so correctly deleted in RQ

### LINKED ISSUE #21129: RQ Jobs in non-model/default queues are kept when Jobs are deleted in DB and reinsert themselves
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.0

### Python Version

3.13

### Steps to Reproduce

Use-Cases:
* Canceling a queued/scheduled job
* Re-Scheduling an existing Job (currently only possible by delete + enqueue)

Steps:
1. Have a plugin with a Job class inheriting from `JobRunner`
2. Enqueue a scheduled Job for this Job class and pass a `queue_name` that is
  2.1 not the default queue
  2.2 not the queue assigned via `QUEUE_MAPPING`, if any was defined
3. Delete the Job instance before its RQ Job runs, e.g. by using a scheduled Job while enqueueing or by not having enough workers to run it directly

### Expected Behavior

Non-running RQ Job of deleted Job should be canceled and not start.

### Observed Behavior

RQ Job of deleted Job stays queued/scheduled and is started by workers.

When that RQ Job starts, it uses the pickled Job Instance of its kwargs and saves it to the database, effectively re-inserting it with a new created datetime but under the same primary key.

This happens because feature of PR #19424 is not fully implemented - the delete logic in model `Job` is not passed this queue name and may not find the RQ Job to cancel it. It then silently ignores this issue without logging it and exits, leaving the RQ Job queued.

The delete logic only looks in one of two queues - either the queue defined by model label (app label is not differentiated) in `QUEUE_MAPPINGS` or otherwise in `default`:
https://github.com/netbox-community/netbox/blob/3624b88c3fa6cb677756823db1413081b3a6607a/netbox/core/models/jobs.py#L184-L193

A workaround for plugins is to override the `handle()` function in the Job class and check if the job instance actually exists, before doing anything:
```python
class SomeJob(JobRunner):
    @classmethod
    def handle(cls, job, *args, **kwargs):
        try:
            job.refresh_from_db()
        except Job.DoesNotExist:
            plugin_logger.info(f'{job} #{job.pk}/RQ Job {job.job_id} is not present in DB anymore, exit early')
            return

        return super().handle(job=job, *args, **kwargs)
```

PS: I do not know why the Job instance is pickled, but it seems odd - an RQ Job uses potentially outdated data as there is no refresh being done in current handle function and it leads to reinserting itself, as we can see. Instead of this it would be safer to just pass primary key, app label and model label to then fetch the Job instance on start.

arthanson:
### Fixes: #21129 

Issue was PR https://github.com/netbox-community/netbox/pull/19424 added a queue_name but it isn't saved to the Job record so if you don't use one of the standard queues it doesn't know where to cancel it from.  Added the queue_name to the job record and associated changes for filter, tables, REST_API, etc.. Also slight change to the delete function to remove it from the correct queue.

Note: After talking with Jeremy as this is a follow up to a change, even thought this has a modification to the REST API it is okay for a patch release.


netbox/core/filtersets.py
```diff
@@ -129,10 +129,16 @@ class JobFilterSet(BaseFilterSet):
         choices=JobStatusChoices,
         null_value=None
     )
+    queue_name = django_filters.CharFilter(
```
jeremystretch: We should probably leave this with the default `iexact` lookup, and let the user apply a lookup modifier if they want to filter on partial matches.

netbox/core/filtersets.py
jeremystretch, line 145: Should we add `queue_name` here?

jeremystretch (changes requested):
Nice solution, just a couple very minor notes.

netbox/core/filtersets.py
arthanson: IMHO: I don't think so - I just don't think it would be useful to globally search for it and is potentially extra noise.  You would just want to go to Jobs and filter on it I would think.

jeremystretch (approved):
Thanks @arthanson!

PR #21312, opened 2026-01-28
Closes #21244: Introduce ability to omit specific fields from REST API responses

### LINKED ISSUE #21244: Enable omitting specific fields from REST API responses (inverse of `?fields=`)
### NetBox version

v4.5.1

### Feature type

New functionality

### Proposed functionality

It's currently possible to limit the fields returned by the REST API by passing a list of fields names appended to the `fields` URL parameter. For example, `GET /api/dcim/sites/?fields=id,name,status` will return the following fields for each site:

```json
{
    "id": 24,
    "name": "Butler Communications",
    "status": {
        "value": "active",
        "label": "Active"
    }
}
```

This FR proposes the introduction of an inverse parameter (i.e. `exclude` or `omit`) that allows the client to specify only the fields to _exclude_ from the response. For example, `GET /api/dcim/sites/?exclude=device_count,virtualmachine_count` would omit the counts of assigned devices and VMs, but retain all other fields.

### Use case

This parameter would be convenient in scenarios where the client wants to omit certain fields for performance reasons, but otherwise retain all default serializer fields.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #21244

- Extend BaseModelSerializer to accept the `omit` keyword argument on initialization.
- Replace `requested_fields()` on BaseViewSet with `field_kwargs()`, which returns a dictionary to pass as kwargs when instantiating the serializer (e.g. `{'fields': [...]}`).
- Update the `get_prefetches_for_serializer()` and `get_annotations_for_serializer()` functions to accept `fields` or `omit`.
- Add documentation for the `fields` and `omit` parameters.

### DIFF
--- a/docs/integrations/rest-api.md
+++ b/docs/integrations/rest-api.md
@@ -250,13 +250,16 @@ Similarly, you can opt to omit only specific fields by passing the `omit` parame
 GET /api/dcim/sites/?omit=circuit_count,device_count,virtualmachine_count
 ```
 
-!!! note "The `omit` paramater was introduced in NetBox v4.5.2."
+!!! note "The `omit` parameter was introduced in NetBox v4.5.2."
 
-Strategic use of the `fields` and `omit` parameters can drastically improve REST API performance, as the exclusion of fields which reference related objects reduces the number and compelxity of underlying database queries needed to generate the response.
+Strategic use of the `fields` and `omit` parameters can drastically improve REST API performance, as the exclusion of fields which reference related objects reduces the number and complexity of underlying database queries needed to generate the response.
+
+!!! note
+    The `fields` and `omit` parameters should be considered mutually exclusive. If both are passed, `fields` takes precedence.
 
 #### Brief Format
 
-Most API endpoints support an optional "brief" format, which returns only a minimal representation of each object in the response. This is useful when you need only a list of available objects without any related data, such as when populating a drop-down list in a form. It's also more convenient than listing out individual fileds via the `fields` or `omit` parameters. As an example, the default (complete) format of a prefix looks like this:
+Most API endpoints support an optional "brief" format, which returns only a minimal representation of each object in the response. This is useful when you need only a list of available objects without any related data, such as when populating a drop-down list in a form. It's also more convenient than listing out individual fields via the `fields` or `omit` parameters. As an example, the default (complete) format of a prefix looks like this:
 
 ```no-highlight
 GET /api/ipam/prefixes/13980/
--- a/docs/integrations/rest-api.md
+++ b/docs/integrations/rest-api.md
@@ -215,9 +215,48 @@ http://netbox/api/ipam/ip-addresses/ \
 
 If we wanted to assign this IP address to a virtual machine interface instead, we would have set `assigned_object_type` to `virtualization.vminterface` and updated the object ID appropriately.
 
-### Brief Format
+### Specifying Fields
 
-Most API endpoints support an optional "brief" format, which returns only a minimal representation of each object in the response. This is useful when you need only a list of available objects without any related data, such as when populating a drop-down list in a form. As an example, the default (complete) format of a prefix looks like this:
+A REST API response will include all available fields for the object type by default. If you wish to return only a subset of the available fields, you can append `?fields=` to the URL followed by a comma-separated list of field names. For example, the following request will return only the `id`, `name`, `status`, and `region` fields for each site in the response.
+
+```
+GET /api/dcim/sites/?fields=id,name,status,region
+```
+
+```json
+{
+    "id": 1,
+    "name": "DM-NYC",
+    "status": {
+        "value": "active",
+        "label": "Active"
+    },
+    "region": {
+        "id": 43,
+        "url": "http://netbox:8000/api/dcim/regions/43/",
+        "display": "New York",
+        "name": "New York",
+        "slug": "us-ny",
+        "description": "",
+        "site_count": 0,
+        "_depth": 2
+    }
+}
+```
+
+Similarly, you can opt to omit only specific fields by passing the `omit` parameter:
+
+```
+GET /api/dcim/sites/?omit=circuit_count,device_count,virtualmachine_count
+```
+
+!!! note "The `omit` paramater was introduced in NetBox v4.5.2."
+
+Strategic use of the `fields` and `omit` parameters can drastically improve REST API performance, as the exclusion of fields which reference related objects reduces the number and compelxity of underlying database queries needed to generate the response.
+
+#### Brief Format
+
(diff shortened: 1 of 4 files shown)


docs/integrations/rest-api.md
```diff
@@ -215,9 +215,48 @@ http://netbox/api/ipam/ip-addresses/ \
...
+}
+```
+
+Similarly, you can opt to omit only specific fields by passing the `omit` parameter:
+
+```
+GET /api/dcim/sites/?omit=circuit_count,device_count,virtualmachine_count
+```
+
+!!! note "The `omit` paramater was introduced in NetBox v4.5.2."
```
pheus:
```suggestion
!!! note "The `omit` parameter was introduced in NetBox v4.5.2."
```

docs/integrations/rest-api.md
```diff
@@ -215,9 +215,48 @@ http://netbox/api/ipam/ip-addresses/ \
...
+
+Strategic use of the `fields` and `omit` parameters can drastically improve REST API performance, as the exclusion of fields which reference related objects reduces the number and compelxity of underlying database queries needed to generate the response.
```
pheus:
```suggestion
Strategic use of the `fields` and `omit` parameters can drastically improve REST API performance, as the exclusion of fields which reference related objects reduces the number and complexity of underlying database queries needed to generate the response.
```

docs/integrations/rest-api.md
```diff
@@ -215,9 +215,48 @@ http://netbox/api/ipam/ip-addresses/ \
...
+
+#### Brief Format
+
+Most API endpoints support an optional "brief" format, which returns only a minimal representation of each object in the response. This is useful when you need only a list of available objects without any related data, such as when populating a drop-down list in a form. It's also more convenient than listing out individual fileds via the `fields` or `omit` parameters. As an example, the default (complete) format of a prefix looks like this:
```
pheus:
```suggestion
Most API endpoints support an optional "brief" format, which returns only a minimal representation of each object in the response. This is useful when you need only a list of available objects without any related data, such as when populating a drop-down list in a form. It's also more convenient than listing out individual fields via the `fields` or `omit` parameters. As an example, the default (complete) format of a prefix looks like this:
```

netbox/netbox/api/serializers/base.py
pheus, line 29:
```suggestion
        :param omit: An iterable of fields to omit from the serialized object
        """
```

netbox/netbox/api/viewsets/__init__.py
```diff
@@ -59,33 +59,37 @@ def get_queryset(self):
...
-    def requested_fields(self):
+    def field_kwargs(self):
         # An explicit list of fields was requested
         if requested_fields := self.request.query_params.get('fields'):
-            return requested_fields.split(',')
+            return {'fields': requested_fields.split(',')}
+
+        # An explicit list of fields to omit was requested
+        if omit_fields := self.request.query_params.get('omit'):
+            return {'omit': omit_fields.split(',')}
```
pheus, line 85: The early return of `fields` silently ignores the an additional `omit`.

netbox/netbox/api/serializers/base.py
```diff
@@ -41,6 +41,15 @@ def __init__(self, *args, nested=False, fields=None, **kwargs):
 
         super().__init__(*args, **kwargs)
 
+        # If fields have been specified, remove any default fields not in the list
+        if fields is not None:
+            for field_name in set(fields) - set(self.fields):
+                self.fields.pop(field_name, None)
```
pheus: It looks like we already fully apply the `fields` filtering in the overridden `fields` property. Because of that, removing default fields in `__init__()` ends up being redundant (and can effectively become a no-op since the final field set is rebuilt in `fields` anyway).

To keep the include/omit logic in one place (and avoid mutating `self.fields` twice), I’d suggest moving the handling back into `fields`, like so:

```python
def __init__(self, *args, nested=False, fields=None, omit=None, **kwargs):
    ...
    self.nested = nested
    self._requested_fields = fields
    self._omit_fields = omit or []
    ...

@cached_property
def fields(self):
    # Start with either "all" fields or the explicitly requested subset
    if not self._requested_fields:
        fields = super().fields
    else:
        fields = BindingDict(self)
        for key, value in self.get_fields().items():
            if key in self._requested_fields:
                fields[key] = value

    # Apply omission
    for field_name in set(self._omit_fields):
        fields.pop(field_name, None)

    return fields
```

pheus (changes requested):
Thanks for tackling this!

A few things I’d suggest
- **Conflicting params (`fields` + `omit`)**: right now it looks like one will “win” implicitly. I think we should explicitly reject requests that supply both (ideally as a 400 with a clear message), similar to the guard already present in the helper functions. That avoids surprising behavior.
- **Tests**: could we add at least one API test for:
  - `?omit=<field>` removes the field from the response
  - `?fields=...&omit=...` returns a 400 (if we implement the explicit rejection)

Happy to re-review.

netbox/netbox/api/serializers/base.py
jeremystretch: Great catch! I went back and forth a number of times during the implementation and miss this completely.

netbox/netbox/api/viewsets/__init__.py
jeremystretch: This seemed like the cleanest way to handle the condition where both `fields` and `omit` are passed. Otherwise, we'd need to trigger a 400 response, which I'm not sure makes sense. Maybe we can just specify in the docs that the `fields` parameter takes precedence over `omit`.

pheus: That's fine. I wanted to make sure that this is an intended behavior. (Sorry, I've should have clarified my comment here.)

netbox/netbox/api/serializers/base.py
pheus: Thanks for the fast update of the PR, but I think those lines are no longer needed.

netbox/netbox/api/serializers/base.py
```diff
@@ -63,16 +64,20 @@ def to_internal_value(self, data):
...
-        if not self._requested_fields:
-            return super().fields
+        fields = super().fields
+
+        # Include only requested fields
+        if self._include_fields:
+            fields = {
+                k: v for k, v in fields.items() if k in self._include_fields
+            }
+
```
pheus, line 66: This is a nicely readable solution, but I'm a little worried about changing `fields` from `BindingDict` to a plain `dict` as I don't know the side effects.
Would you accept something like your initial solution (to be on the safe side)?

```python
if self._include_fields:
    for field_name in set(fields) - set(self._include_fields):
        fields.pop(field_name, None)
```

pheus (changes requested):
Thanks for the update!

pheus (approved):
Great! Thanks, @jeremystretch !

PR #21321, opened 2026-01-29
Fixes #21320: Prevent Rack validation errors when site or optional fields are missing during import

### LINKED ISSUE #21320: Rack Direct Import - "Rack has no site."
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.1

### Python Version

3.12

### Steps to Reproduce

0. Site allready created and used allready  
1. Go to "Rack"
2. Export current view
3. Add Racks
4. Paste edited CSV into "Direct Import" -> "Data"
5. Leave everything on Auto-detect (same effect on setting CSV)
6. Error -> Rack has no site.

### Expected Behavior

Bulk create new Racks

### Observed Behavior

Error: 
```
<class 'dcim.models.racks.Rack.site.RelatedObjectDoesNotExist'>

Rack has no site.

Python version: 3.12.3
NetBox version: 4.5.1
Plugins: 
  netbox_floorplan: 0.9.0
  netbox_topology_views: 4.5.0
```
Both CSVs attached.

[netbox_racks_exported.csv](https://github.com/user-attachments/files/24936274/netbox_racks_exported.csv)
[netbox_racks_tobeimported.csv](https://github.com/user-attachments/files/24936273/netbox_racks_tobeimported.csv)

pheus:
### Fixes: #21320

This PR prevents an uncaught exception during Rack direct import/validation when the `site` relation isn’t set (e.g. due to missing/incorrect CSV headers), which previously could raise `RelatedObjectDoesNotExist`.

#### Summary of changes
- Update Rack validation to use the FK ID fields (`site_id` / `location_id`) instead of accessing related objects directly.
- Avoid triggering related-object lookups during validation, ensuring we return normal, user-facing validation errors rather than a backend exception.
- Aligns the validation logic with Django’s underlying foreign key fields and behavior when relations are unset.

Thanks again to @cpt-kernel-afk for the report and sample CSVs, and to @jnovinger for the review/suggestion to use FK IDs here.

### DIFF
--- a/netbox/dcim/models/racks.py
+++ b/netbox/dcim/models/racks.py
@@ -373,7 +373,7 @@ class Rack(ContactsMixin, ImageAttachmentsMixin, TrackingModelMixin, RackBase):
         super().clean()
 
         # Validate location/site assignment
-        if self.site and self.location and self.location.site != self.site:
+        if self.site_id and self.location_id and self.location.site_id != self.site_id:
             raise ValidationError(_("Assigned location must belong to parent site ({site}).").format(site=self.site))
 
         # Validate outer dimensions and unit


netbox/dcim/models/racks.py
```diff
@@ -373,15 +373,21 @@ def clean(self):
         super().clean()
 
         # Validate location/site assignment
-        if self.site and self.location and self.location.site != self.site:
+        if getattr(self, 'site', None) and getattr(self, 'location', None) and self.location.site != self.site:
```
jnovinger:
```suggestion
        if self.site_id and self.location_id and self.location.site != self.site:
```

Since we're checking whether the model instance has a relation, before then checking if the sites match, I think we can avoid using `getattr()` here. I generally think of using `getattr()` when we're not sure if an instance even has the attribute we want to check, as opposed to checking to see if there's a relation set.

Because of the way Django sets foreign keys, we have access to the relation field (e.g. `self.site`) as the related instance. But, we also have access to the foreign key ID field underlying that relation (e.g. `self.site_id`), which Django defaults to creating as `{field_name}_id`.

jnovinger: Alternatively, we could wrap this in a try/except that looks for `RelatedObjectDoesNotExist`. But, I like avoiding exception catching as part of flow control.

netbox/dcim/models/racks.py
```diff
@@ -373,15 +373,21 @@ def clean(self):
...
 
         # Validate outer dimensions and unit
-        if any([self.outer_width, self.outer_depth, self.outer_height]) and not self.outer_unit:
+        if any(
+                [
+                    getattr(self, 'outer_width', None),
+                    getattr(self, 'outer_depth', None),
+                    getattr(self, 'outer_height', None),
+                ]
+        ) and not getattr(self, 'outer_unit', None):
```
jnovinger: I'm not sure we need this safe checking here, since this is validation on the model and we can rely on the fact that Django will have set default values if they were not passed in.

netbox/dcim/models/racks.py
```diff
@@ -373,15 +373,21 @@ def clean(self):
...
             raise ValidationError(_("Must specify a unit when setting an outer dimension"))
 
         # Validate max_weight and weight_unit
-        if self.max_weight and not self.weight_unit:
+        if getattr(self, 'max_weight', None) and not getattr(self, 'weight_unit', None):
```
jnovinger: Same, I'm not convinced these safe checks are required. Can we verify by removing them and seeing if the STR is still able to trigger the bug?

jnovinger (changes requested):

jnovinger (changes requested):
Left some thoughts on how you're checking those values safely.

netbox/dcim/models/racks.py
pheus: Good point on using the FK ID fields here. I’ll replace the `getattr()` checks with `self.site_id` / `self.location_id` to avoid hitting the related-object descriptor at all.

netbox/dcim/models/racks.py
pheus: Also agreed on the outer dimension and max weight checks. I’ll revert those back to direct attribute checks!
Thanks!

jnovinger (approved):

PR #21322, opened 2026-01-29
Closes #21110: Support for cursor-based pagination in GraphQL API

### LINKED ISSUE #21110: Support cursor-based pagination in GraphQL API
### NetBox version

v4.5.0

### Feature type

New functionality

### Proposed functionality

The GraphQL API currently supporting paginating through results by passing `offset` and `limit` parameters with a request. For example:

```graphql
query {
    device_list(
        pagination: {
            offset: $offset,
            limit: $limit
        }
    ) {
        id
        name
    }
}
```

This FR proposes an alternative to the `offset` parameter which specifies the primary key of the first object to fetch. For now, let's call this parameter `start`. Passing this parameter forces results to be ordered by primary key only, ensuring a deterministic set of results. The existing `limit` parameter can also be included to limit the number of results returned. For example:

```graphql
query {
    device_list(
        pagination: {
            start: 4629,
            limit: 100
        }
    ) {
        id
        name
    }
}
```

The above query would yield (roughly) the following Django queryset:

```python
Device.objects.order_by('pk').filter(pk__gte=4629)[:100]
```

The pagination strategy to be applied for a request is inferred from the parameter passed. The `offset` and `start` parameters are to be mutually exclusive.

#### Rough implementation plan

Subclass `OffsetPaginationInput` to introduce the `start` parameter:

```python
import strawberry
from strawberry_django.pagination import OffsetPaginationInput

@strawberry.input
class StartIdPaginationInput(OffsetPaginationInput):
    start: int | None = None
```

Subclass `StrawberryDjangoField` to create a custom GraphQL field which employs our custom pagination by default:

```python
from strawberry_django.fields.field import StrawberryDjangoField

class StartIdStrawberryDjangoField(StrawberryDjangoField):
    def apply_pagination(self, queryset, pagination=None):
        if pagination is not None:
            if start = getattr(pagination, "start", None):
                # TODO: Raise error if `offset` is also present
                queryset = queryset.filter(pk__gte=start).order_by("pk")

        return super().apply_pagination(queryset, pagination)
```

Use this field in place of `strawberry_django.field()` for all GraphQL lists:

```python
# netbox/netbox/graphql/fields.py (new, or wherever NetBox keeps shared GraphQL helpers)
from functools import partial
import strawberry_django

from .pagination import StartIdPaginationInput, StartIdStrawberryDjangoField

netbox_field = partial(
    strawberry_django.field,
    pagination=StartIdPaginationInput,
    field_cls=StartIdStrawberryDjangoField,
)
```

```python
from netbox.graphql.fields import netbox_field

foo_list: list[FooType] = netbox_field()
```

### Use case

The current pagination approach relies on a relative offset. This struggles with very large result sets, because the offset must be calculated by scanning the entire table up to the offset position. Employing an object's unique primary key as a cursor avoids this, as we can leverage simple numeric comparison to filter for all PKs greater than or equal to a given value. Combined with the existing `limit` capability, this affords very efficient pagination at the cost of a fixed (and not very meaningful) ordering.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #21110

- Monkey-patch a custom `OffsetPaginationInput` class to support a `start` pagination parameter
- Monkey-patch a customized `apply_pagination()` function to replace the stock method on `StrawberryDjangoField`
- Add tests for both offset- and cursor-based pagination
- Update the pagination docs for GraphQL
    - Also tidy up `test_graphql_filter_objects()`

### DIFF
--- a/docs/integrations/graphql-api.md
+++ b/docs/integrations/graphql-api.md
@@ -137,6 +137,8 @@ The GraphQL API supports two types of pagination. Offset-based pagination operat
 
 The alternative approach is cursor-based pagination, which operates using absolute (rather than relative) primary key values. (These are the numeric IDs assigned to each object in the database.) When using cursor-based pagination, the response will contain records with a primary key greater than or equal to the specified start value, up to the maximum number of results. This strategy requires keeping track of the last seen primary key from each response when paginating through data, but is extremely performant. The cursor is specified by passing the starting object ID via the `start` parameter.
 
+To ensure consistent ordering, objects will always be ordered by their primary keys when cursor-based pagination is used.
+
 !!! note "Cursor-based pagination was introduced in NetBox v4.5.2."
 
 Both pagination strategies support passing an optional `limit` parameter. In both approaches, this specifies the maximum number of objects to include in the response. If no limit is specified, a default value of 100 is used.
--- a/docs/integrations/graphql-api.md
+++ b/docs/integrations/graphql-api.md
@@ -133,23 +133,65 @@ The field "class_type" is an easy way to distinguish what type of object it is w
 
 ## Pagination
 
-Queries can be paginated by specifying pagination in the query and supplying an offset and optionaly a limit in the query.  If no limit is given, a default of 100 is used.  Queries are not paginated unless requested in the query. An example paginated query is shown below:
+The GraphQL API supports two types of pagination. Offset-based pagination operates using an offset relative to the first record in a set, specified by the `offset` parameter. For example, the response to a request specifying an offset of 100 will contain the 101st and later matching records. Offset-based pagination feels very natural, but its performance can suffer when dealing with large data sets due to the overhead involved in calculating the relative offset.
+
+The alternative approach is cursor-based pagination, which operates using absolute (rather than relative) primary key values. (These are the numeric IDs assigned to each object in the database.) When using cursor-based pagination, the response will contain records with a primary key greater than or equal to the specified start value, up to the maximum number of results. This strategy requires keeping track of the last seen primary key from each response when paginating through data, but is extremely performant. The cursor is specified by passing the starting object ID via the `start` parameter.
+
+!!! note "Cursor-based pagination was introduced in NetBox v4.5.2."
+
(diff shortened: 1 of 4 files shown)


netbox/netbox/graphql/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+    related_field_id: str | None = None,
+) -> _QS:
+    """
+    Replacement for the `apply_pagination()` method on StrawberryDjangoField to support cursor-based pagination.
+    """
+    if pagination is not None:
+        if pagination.start and pagination.offset:
+            raise ValueError('Cannot specify both `start` and `offset` in pagination.')
+
+        if pagination.start not in (None, UNSET):
```
arthanson: Couple additional validations needed:
1. validate that start isn't negative, offset correctly checks this.
2. if you have start: 0, offset: 0 the logic doesn't catch that.

docs/integrations/graphql-api.md
```diff
@@ -133,23 +133,65 @@ The field "class_type" is an easy way to distinguish what type of object it is w
...
+The GraphQL API supports two types of pagination. Offset-based pagination operates using an offset relative to the first record in a set, specified by the `offset` parameter. For example, the response to a request specifying an offset of 100 will contain the 101st and later matching records. Offset-based pagination feels very natural, but its performance can suffer when dealing with large data sets due to the overhead involved in calculating the relative offset.
+
+The alternative approach is cursor-based pagination, which operates using absolute (rather than relative) primary key values. (These are the numeric IDs assigned to each object in the database.) When using cursor-based pagination, the response will contain records with a primary key greater than or equal to the specified start value, up to the maximum number of results. This strategy requires keeping track of the last seen primary key from each response when paginating through data, but is extremely performant. The cursor is specified by passing the starting object ID via the `start` parameter.
+
```
arthanson, line 139: Do we want to mention that ordering will be different.  By default the GraphQL isn't ordered, using the cursor-based pagination it will be ordered by pk.

arthanson (changes requested):

arthanson (approved):

PR #21324, opened 2026-01-29
Closes #21201: Support GenericForeignKey fields when cloning objects

### LINKED ISSUE #21201: Pre-populate GenericForeignKey form fields when cloning

pheus:
### Fixes: #21201

This PR improves cloning support for models that use **GenericForeignKey (GFK)** fields.

#### Problem
When cloning an object, `CloningMixin.clone()` currently copies only the underlying database fields (e.g. `<name>_type` and `<name>_id`). In the UI, however, NetBox forms typically expose a single combined `DynamicModelChoiceField` named after the GFK attribute (e.g. `parent`, `scope`). As a result, the selector field is not pre-populated during cloning even though the CT+ID values are present in the URL parameters.

#### Changes
- Enhance `CloningMixin.clone()` to detect `GenericForeignKey` fields and, **when both the content type field and ID field are included in `clone_fields`**, inject the GFK attribute name into the returned clone dict (unless already set).
- Add tests covering:
  - successful injection when both constituent fields are present
  - no injection when only one constituent field is present

#### Result
Cloning now pre-populates form fields backed by `GenericForeignKey` consistently across the UI without requiring model- or form-specific overrides.

Thanks for reviewing!

### DIFF
--- a/netbox/netbox/models/features.py
+++ b/netbox/netbox/models/features.py
@@ -2,7 +2,7 @@ import json
 from collections import defaultdict
 from functools import cached_property
 
-from django.contrib.contenttypes.fields import GenericRelation
+from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
 from django.core.validators import ValidationError
 from django.db import models
@@ -158,6 +158,13 @@ class CloningMixin(models.Model):
                 attrs[field_name] = json.dumps(field_value)
             elif field_value not in (None, ''):
                 attrs[field_name] = field_value
+
+        # Handle GenericForeignKeys. If the CT and ID fields are being cloned, also
+        # include the name of the GFK attribute itself, as this is what forms expect.
+        for field in self._meta.private_fields:
+            if isinstance(field, GenericForeignKey):
+                if field.ct_field in attrs and field.fk_field in attrs:
+                    attrs[field.name] = attrs[field.fk_field]
 
         # Include tags (if applicable)
         if is_taggable(self):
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @pheus!

PR #21325, opened 2026-01-29
Closes #21303: Cache serialized post-change data on object

### LINKED ISSUE #21303: Cache post-change data on the instance after serialization
### NetBox Version

v4.5.1

### Python Version

3.12

### Area(s) of Concern

- [x] User Interface
- [x] REST API
- [ ] GraphQL API
- [ ] Python ORM
- [ ] Other

### Details

When calling the `to_objectchange()` method on an instance (to create a new ObjectChange), pre-change data is sourced from the instance's `_prechange_snapshot` attribute (if set). Post-change data is calculated on-demand (via `serialize_object()`) for the ObjectChange but is not cached on the instance being serialized.

When an object event (i.e. create, update, or delete) is queued for processing during the request-response cycle, its pre- and post-change snapshots are included by calling [`get_snapshots()`](https://github.com/netbox-community/netbox/blob/aa69e96818a92918c4e77a0fdef45edfed820dae/netbox/extras/events.py#L39). This utility will leverage an existing pre-change snapshot if one exists, but always force a new call to `serialize_object()` to generate the post-change data (except for deletion events).

We can take the opportunity presented by the first call to `serialize_object()` to cache the resulting data on the instance, and tweak `get_snapshots()` to leverage the cached data if it exists. Although the serialization overhead should be negligible, this will at least avoid a second SQL query for assigned tags for object which support tagging.

jeremystretch:
### Closes: #21303

- Extend `to_objectchange()` to cache the serialized data as `_postchange_snapshot` on the instance being serialized.
- Modify `get_snapshots()` to check for cached serialized data before running serialization.

### DIFF
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -53,9 +53,9 @@ def serialize_for_event(instance):
 def get_snapshots(instance, event_type):
     snapshots = {
         'prechange': getattr(instance, '_prechange_snapshot', None),
-        'postchange': None,
+        'postchange': getattr(instance, '_postchange_snapshot', None),
     }
-    if event_type != OBJECT_DELETED:
+    if snapshots['postchange'] is None and event_type != OBJECT_DELETED:
         # Use model's serialize_object() method if defined; fall back to serialize_object() utility function
         if hasattr(instance, 'serialize_object'):
             snapshots['postchange'] = instance.serialize_object()
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -51,18 +51,26 @@ def serialize_for_event(instance):
 
 
 def get_snapshots(instance, event_type):
-    snapshots = {
-        'prechange': getattr(instance, '_prechange_snapshot', None),
-        'postchange': getattr(instance, '_postchange_snapshot', None),
-    }
-    if snapshots['postchange'] is None and event_type != OBJECT_DELETED:
-        # Use model's serialize_object() method if defined; fall back to serialize_object() utility function
-        if hasattr(instance, 'serialize_object'):
-            snapshots['postchange'] = instance.serialize_object()
-        else:
-            snapshots['postchange'] = serialize_object(instance)
+    """
+    Return a dictionary of pre- and post-change snapshots for the given instance.
+    """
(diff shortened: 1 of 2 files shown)


netbox/netbox/models/features.py
```diff
@@ -121,7 +121,8 @@ def to_objectchange(self, action):
         if hasattr(self, '_prechange_snapshot'):
             objectchange.prechange_data = self._prechange_snapshot
         if action in (ObjectChangeActionChoices.ACTION_CREATE, ObjectChangeActionChoices.ACTION_UPDATE):
-            objectchange.postchange_data = self.serialize_object(exclude=exclude)
+            self._postchange_snapshot = self.serialize_object(exclude=exclude)
+            objectchange.postchange_data = self._postchange_snapshot
 
         return objectchange
```
pheus, line 129:
```suggestion
    snapshot.alters_data = True
```
Small nit: would it make sense to mark `to_objectchange()` with `alters_data = True`? Since it now sets `_postchange_snapshot` on the instance, it’s mutating state (even if only in-memory).

netbox/extras/events.py
pheus: _Probably an edge case, but worth calling out:_

One thing to double-check: in `get_snapshots()`, `postchange` is now initialized from `getattr(instance, '_postchange_snapshot', None)`. For `OBJECT_DELETED`, we’ve historically/semantically treated `postchange` as always `None`.

In the unlikely case that the same instance already has `_postchange_snapshot` populated (e.g. saved earlier in the request/transaction and then deleted), we might end up returning a non-null `postchange` snapshot for a delete event. It might be worth explicitly forcing `postchange = None` when `event_type == OBJECT_DELETED` to preserve the invariant.

pheus (commented):
Thanks! Just two small nits below.

I’m not strongly opinionated about either of them, so I’m leaving this as a comment-only review and I’m happy with the PR as-is.

netbox/extras/events.py
jeremystretch: That's a really good catch, thanks!

pheus (approved):
Awesome! Thanks, @jeremystretch !

PR #21329, opened 2026-01-30
Closes #21263: Prefetch related objects after creating/updating objects via REST API

### LINKED ISSUE #21263: Related objects are not prefetched when returning data from a REST API `POST` request
### NetBox Version

v4.5.1

### Python Version

3.12

### Area(s) of Concern

- [ ] User Interface
- [x] REST API
- [ ] GraphQL API
- [ ] Python ORM
- [ ] Other

### Details

When creating a number of objects in bulk using a single `POST` request to the REST API endpoint for the model (e.g. `/api/dcim/sites/`), any objects to those created by the request are queried individually from the database.

For example, when creating a number of IP addresses via the `/api/ipam/ip-addresses/` endpoint, the interface to which each IP address is assigned (if any) and that interface's parent device are queried individually for each IP address that was created.

Ideally, these should all be prefetched and attached to the root objects in memory to minimize the number of database queries.

jeremystretch:
### Closes: #21263

Extend the `create()`, `update()`, and `bulk_update()` methods on NetBoxModelViewSet to replace the original request data with a queryset. This ensures that all related objects are prefetched efficiently. It also ensures that the response data accurately reflects the state of the data as written.

(Note: There is no `bulk_create()` method; `create()` handles both individual and bulk writes.)

### DIFF
--- a/netbox/netbox/api/viewsets/__init__.py
+++ b/netbox/netbox/api/viewsets/__init__.py
@@ -170,6 +170,26 @@ class NetBoxModelViewSet(
 
     # Creates
 
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        bulk_create = getattr(serializer, 'many', False)
+        self.perform_create(serializer)
+
+        # After creating the instance(s), re-initialize the serializer with a queryset
+        # to ensure related objects are prefetched.
+        if bulk_create:
+            instance_pks = [obj.pk for obj in serializer.instance]
+            qs = self.get_queryset().filter(pk__in=instance_pks).order_by('pk')
+        else:
+            qs = self.get_queryset().get(pk=serializer.instance.pk)
+
+        # Re-serialize the instance(s) with prefetched data
+        serializer = self.get_serializer(qs, many=bulk_create)
+
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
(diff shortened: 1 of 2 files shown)


netbox/netbox/api/viewsets/__init__.py
```diff
@@ -170,6 +170,26 @@ def dispatch(self, request, *args, **kwargs):
...
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        bulk_create = getattr(serializer, 'many', False)
+        self.perform_create(serializer)
+
+        # After creating the instance(s), re-initialize the serializer with a queryset
+        # to ensure related objects are prefetched.
+        if bulk_create:
+            instance_pks = [obj.pk for obj in serializer.instance]
+            qs = self.get_queryset().filter(pk__in=instance_pks).order_by('pk')
```
jnovinger, line 185: I'm curious about the necessity of the `.order_by()` clause. I assume there's a good reason for that, maybe because that's how the data comes in?

jnovinger (dismissed):
I looked over this but couldn't find anything that would indicate a regression, outside of my one question below. Approved, but not merged, in case we want to talk through that question.

netbox/netbox/api/viewsets/__init__.py
jeremystretch: It ensures that objects are returned in the order they are created. A few objects (e.g. journal entries) default to reverse chronological ordering: Ordering by PK ensures that ordering stays consistent with the request data (and avoids breaking a few tests).

jnovinger (approved):

PR #21333, opened 2026-01-30
Closes #21302: Avoid redundant uniqueness checks in REST API serializers

### LINKED ISSUE #21302: Avoid redundant uniqueness checks in ValidatedModelSerializer
### NetBox Version

v4.5.1

### Python Version

3.12

### Area(s) of Concern

- [ ] User Interface
- [x] REST API
- [ ] GraphQL API
- [ ] Python ORM
- [ ] Other

### Details

When creating a new object via the REST API, unique fields are currently validated twice: First by the serializer field's validator, and again by our custom [`validate()` method](https://github.com/netbox-community/netbox/blob/a9e50238ebba8fda7462ed1b0723d9012205fd28/netbox/netbox/api/serializers/base.py#L85).

IIRC, the custom `validate()` method was originally added to ensure that unique constraints were enforced properly. (However, that change was made years ago and probably deserves to be re-evaluated.) It may suffice to just call `validate_constraints()` directly rather than the entire set of model validation. At a minimum, it should be acceptable to disable uniqueness validation specifically, as that is already handled at the field level:

```python
instance.full_clean(validate_unique=False)
```

jeremystretch:
### Closes: #21302

Pass `validate_unique=False` when calling `full_clean()` on the instance. The serializer already validates fields with individual unique constraints.

### DIFF
--- a/netbox/netbox/api/serializers/base.py
+++ b/netbox/netbox/api/serializers/base.py
@@ -112,6 +112,7 @@ class ValidatedModelSerializer(BaseModelSerializer):
             for k, v in attrs.items():
                 setattr(instance, k, v)
         instance._m2m_values = m2m_values
-        instance.full_clean()
+        # Skip uniqueness validation of individual fields inside `full_clean()` (this is handled by the serializer)
+        instance.full_clean(validate_unique=False)
 
         return data


pheus (approved):
Thanks!

PR #21344, opened 2026-02-02
Closes #21168: Fix Application Service cloning to preserve parent object

### LINKED ISSUE #21168: change "create & add another" behavior for application services
### NetBox version

v.4.5

### Feature type

### Proposed functionality

Currently, when creating application services from a device's context, the **Parent type**, and **Parent** are pre-filled.
However, when you click "Create & Add another", the **Parent type**, and **Parent** reset to null.

The proposal would be to keep **Parent type**, and **Parent** pre-filled with the previous values when clicking "Create & Add another"

### Use case

When attempting to add multiple application services to a device, it's very cumbersome to have to re-select the **Parent type**, and **Parent** for every application service you wish to add to a device.

### Database changes

none expected.

### External dependencies

none expected.

pheus:
### Fixes: #21168

Thanks for taking the time to review this PR!

When creating multiple application services from a parent object context, the parent fields are pre-filled on the initial form, but clicking **Create & Add another** clears them. This makes adding several services to the same parent unnecessarily repetitive.

This change adds `parent_object_type` and `parent_object_id` to the service form’s `clone_fields`, so the parent context is preserved between consecutive creations and object cloning behaves more consistently for models that reference a parent object.

### DIFF
--- a/netbox/ipam/models/services.py
+++ b/netbox/ipam/models/services.py
@@ -87,7 +87,9 @@ class Service(ContactsMixin, ServiceBase, PrimaryModel):
         help_text=_("The specific IP addresses (if any) to which this application service is bound")
     )
 
-    clone_fields = ['protocol', 'ports', 'description', 'parent', 'ipaddresses', ]
+    clone_fields = (
+        'protocol', 'ports', 'description', 'parent_object_type', 'parent_object_id', 'ipaddresses',
+    )
 
     class Meta:
         indexes = (


jeremystretch (approved):
Thanks @pheus!

jeremystretch:
I'm okay merging this without the complete CI suite passing (due to a GitHub service disruption) given how minimal the change is.

PR #21348, opened 2026-02-03
Closes #21266: Add installed device table columns to DeviceBay table 

### LINKED ISSUE #21266: Device Bays view to allow Child device properties as table columns
### NetBox version

v4.4.5

### Feature type

New functionality

### Proposed functionality

When viewing the device-bays for a Device that is a parent, it would be super handy to be able to have the Configure Table button allow for picking Child devices properties as viewable in this table. The particular one we would like is Child Description. 
While the Description field is for the Bay, the Child Description would be a super handy column to display all the info at once. Other select Child properties would be nice too if at all possible. 

### Use case

The use case here is specifically to be able to see more info than just the name for the child devices. For broadcast industry devices, they can have many bays, and the child devices all do different things. Helping the user have a single view where they can see what does what in this parent device would be super handy.

### Database changes

Not sure if we need DB changes - this is more a frontend query perhaps - to render the Child properties on that line into the Device-Bays view.

### External dependencies

nil

DylanLucci:
### Fixes: #21266

This PR updates the **DeviceBay** table to include more columns related to installed devices.
- Some existing columns were prefixed with `installed_` for child devices whereas other child device columns had no prefix. For clarity and consistency, other child device columns have been prefixed with `installed_`.
- DeviceBay table has been updated to include more columns related to installed devices.
- Capitalised existing table column names to be consistent with existing capitalisation.

### DIFF
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -917,8 +917,8 @@ class DeviceBayTable(DeviceComponentTable):
     class Meta(DeviceComponentTable.Meta):
         model = models.DeviceBay
         fields = (
-            'pk', 'id', 'name', 'device', 'label', 'status', 'installed_role', 'installed_type', 'installed_device',
-            'description', 'installed_description', 'installed_serial', 'installed_asset_tag', 'tags', 'created',
+            'pk', 'id', 'name', 'device', 'label', 'status', 'description', 'installed_device', 'installed_role',
+            'installed_type', 'installed_description', 'installed_serial', 'installed_asset_tag', 'tags', 'created',
             'last_updated',
         )
 
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -891,24 +891,24 @@ class DeviceBayTable(DeviceComponentTable):
     )
     installed_role = columns.ColoredLabelColumn(
         accessor=Accessor('installed_device__role'),
-        verbose_name=_('Installed Device Role')
+        verbose_name=_('Installed Role')
     )
-    installed_device_type = tables.Column(
+    installed_type = tables.Column(
         accessor=Accessor('installed_device__device_type'),
         linkify=True,
-        verbose_name=_('Installed Device Type')
+        verbose_name=_('Installed Type')
     )
     installed_description = tables.Column(
         accessor=Accessor('installed_device__description'),
-        verbose_name=_('Installed Device Description')
+        verbose_name=_('Installed Description')
     )
     installed_serial = tables.Column(
         accessor=Accessor('installed_device__serial'),
-        verbose_name=_('Installed Device Serial')
+        verbose_name=_('Installed Serial')
     )
     installed_asset_tag = tables.Column(
         accessor=Accessor('installed_device__asset_tag'),
-        verbose_name=_('Installed Device Asset Tag')
+        verbose_name=_('Installed Asset Tag')
     )
     tags = columns.TagColumn(
         url_name='dcim:devicebay_list'
@@ -917,9 +917,9 @@ class DeviceBayTable(DeviceComponentTable):
     class Meta(DeviceComponentTable.Meta):
         model = models.DeviceBay
         fields = (
-            'pk', 'id', 'name', 'device', 'label', 'status', 'installed_role', 'installed_device_type',
-            'installed_device', 'description', 'installed_description', 'installed_serial', 'installed_asset_tag',
-            'tags', 'created', 'last_updated',
+            'pk', 'id', 'name', 'device', 'label', 'status', 'installed_role', 'installed_type', 'installed_device',
+            'description', 'installed_description', 'installed_serial', 'installed_asset_tag', 'tags', 'created',
+            'last_updated',
         )
 
         default_columns = ('pk', 'name', 'device', 'label', 'status', 'installed_device', 'description')
--- a/netbox/dcim/tables/devices.py
+++ b/netbox/dcim/tables/devices.py
@@ -893,7 +893,7 @@ class DeviceBayTable(DeviceComponentTable):
         accessor=Accessor('installed_device__role'),
         verbose_name=_('Installed Role')
     )
-    installed_type = tables.Column(
+    installed_device_type = tables.Column(
         accessor=Accessor('installed_device__device_type'),
         linkify=True,
         verbose_name=_('Installed Type')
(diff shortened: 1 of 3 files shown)


pheus (changes requested):
Thank you for your work here, @DylanLucci!
It’s clear you put a lot of effort into this, and the core idea (exposing *child device* info in the parent’s Device Bays table) makes sense and aligns with the original FR.

That said, after reviewing the FR scope and the changes in this PR, it looks like the implementation goes quite a bit beyond what was requested/accepted. To keep this change focused, could you please trim this PR back to the minimum required for the FR?

Specifically:

#### Please drop the filtering additions
- Remove the added `FilterSet` and `FilterForm` changes for `DeviceBay`.

Rationale: filtering by device attributes is already covered in the device list views (e.g. via `DeviceFilterForm`). Adding dedicated filters on the Device Bays table feels redundant for the stated use case and increases ongoing maintenance for relatively little benefit.

#### Please remove a couple of the heavier table columns
- `installed_primary_ip`
- `installed_platform`

Rationale: these require additional joins and/or related lookups (and, in the case of primary IP, an extra level of indirection), which can add noticeable query cost on a view that may need to render many bays at once. We generally try to keep table columns here lightweight, since this view is primarily about bays + the installed device relationship.

If you’d like, a follow-up FR could be proposed separately for any broader “child device metadata” expansion (or additional filtering), where we can evaluate the use case and the tradeoffs. But for this PR, I’d like to keep it centered on the FR’s intent: *adding one or a small set of child device properties as configurable columns* (e.g. child description, serial, asset_tag).

Thanks again! Happy to take another look once it’s updated. 🙂

DylanLucci:
Hi @pheus!

Thank you very much for your review and proposed changes to keep the PR more focused. I've just implemented those now and await approval.

Thanks again!

netbox/dcim/tables/devices.py
```diff
@@ -880,32 +880,45 @@ class DeviceBayTable(DeviceComponentTable):
...
     tags = columns.TagColumn(
         url_name='dcim:devicebay_list'
     )
 
     class Meta(DeviceComponentTable.Meta):
         model = models.DeviceBay
         fields = (
-            'pk', 'id', 'name', 'device', 'label', 'status', 'role', 'device_type', 'installed_device', 'description',
+            'pk', 'id', 'name', 'device', 'label', 'status', 'installed_role', 'installed_device_type',
+            'installed_device', 'description', 'installed_description', 'installed_serial', 'installed_asset_tag',
```
pheus:
```suggestion
            'pk', 'id', 'name', 'device', 'label', 'status', 'description', 'installed_device',
            'installed_role', 'installed_device_type', 'installed_description', 'installed_serial',
            'installed_asset_tag',
```
In the `fields`, could we keep the `installed_*` fields together (as you already started doing)?

pheus (changes requested):
Thanks for the adjustments, @DylanLucci! This is much appreciated!

Small follow-ups to better match NetBox’s table conventions and keep things tidy:

**Column header naming**
While being explicit is totally understandable, `verbose_name` values are usually kept fairly short. The current `Installed Device …` labels feel a bit wordy in the table config picker.

Would you be open to shortening them to something like:
- **Installed Role**
- **Installed Type**
- **Installed Description**
- **Installed Serial**
- **Installed Asset Tag**

netbox/dcim/tables/devices.py
DylanLucci: @pheus No problem, just finished up with those changes now and awaiting approval.

Thank you!

netbox/dcim/tables/devices.py
```diff
@@ -880,33 +880,46 @@ class DeviceBayTable(DeviceComponentTable):
...
     installed_device = tables.Column(
-        verbose_name=_('Installed device'),
+        verbose_name=_('Installed Device'),
         linkify=True
     )
+    installed_role = columns.ColoredLabelColumn(
+        accessor=Accessor('installed_device__role'),
+        verbose_name=_('Installed Role')
+    )
+    installed_type = tables.Column(
```
pheus:
```suggestion
    installed_device_type = tables.Column(
```
Thanks for the quick update! I think this rename might be a small miscommunication on my side: when I suggested “Installed Type”, I meant the **UI header** (`verbose_name`), not the underlying table field name.

Since the related model field is `device_type`, it probably makes sense to keep the column key as `installed_device_type` (and just use `verbose_name=_("Installed Type")` for the display label).

Sorry for the confusion, and thanks again for iterating on this! 🙂

netbox/dcim/tables/devices.py
```diff
@@ -880,33 +880,46 @@ class DeviceBayTable(DeviceComponentTable):
...
         url_name='dcim:devicebay_list'
     )
 
     class Meta(DeviceComponentTable.Meta):
         model = models.DeviceBay
         fields = (
-            'pk', 'id', 'name', 'device', 'label', 'status', 'role', 'device_type', 'installed_device', 'description',
-            'tags', 'created', 'last_updated',
+            'pk', 'id', 'name', 'device', 'label', 'status', 'description', 'installed_device', 'installed_role',
+            'installed_type', 'installed_description', 'installed_serial', 'installed_asset_tag', 'tags', 'created',
```
pheus:
```suggestion
            'installed_device_type', 'installed_description', 'installed_serial', 'installed_asset_tag', 'tags', 'created',
```

pheus (changes requested):
Thanks for the quick update!

netbox/dcim/tables/devices.py
DylanLucci: Sorry for the misunderstanding, I've just updated this!

pheus: Thanks!

netbox/dcim/tables/devices.py
```diff
@@ -880,33 +880,46 @@ class DeviceBayTable(DeviceComponentTable):
...
     )
 
     class Meta(DeviceComponentTable.Meta):
         model = models.DeviceBay
         fields = (
-            'pk', 'id', 'name', 'device', 'label', 'status', 'role', 'device_type', 'installed_device', 'description',
-            'tags', 'created', 'last_updated',
+            'pk', 'id', 'name', 'device', 'label', 'status', 'description', 'installed_device', 'installed_role',
+            'installed_device_type', 'installed_description', 'installed_serial', 'installed_asset_tag', 'tags', 'created',
+            'last_updated',
```
pheus:
```suggestion
            'pk', 'id', 'name', 'device', 'label', 'status', 'description', 'installed_device',
            'installed_role', 'installed_device_type', 'installed_description', 'installed_serial',
            'installed_asset_tag',
            'tags', 'created', 'last_updated',
```

pheus (approved):
Thanks, @DylanLucci !

Issue #20923, opened 2025-12-04
Replace object view HTML templates with declarative layouts


jeremystretch:
### Proposed Changes

NetBox v4.5 introduced modular template components under FR [#20204](https://github.com/netbox-community/netbox/issues/20204). However, the vast majority of object view templates have not yet been migrated. This issue has been created to track template updates across all relevant apps within NetBox.

- [x] Circuits
- [x] Core
- [x] DCIM
- [ ] Extras
- [x] IPAM
- [x] Tenancy
- [X] Users
- [X] Virtualization
- [X] VPN
- [X] Wireless

### Justification

Replacing legacy HTML with modular components will drastically reduce template complexity and ensure a consistent look and feel across the entire application.

PR #21372, opened 2026-02-06
#20923: Migrate DCIM view templates

jeremystretch:
Within the scope of #20923, this migrates *some of* the remaining DCIM views to the new UI layout mechanism.


jeremystretch:
@claude /review check for any inconsistencies between the original HTML templates and the new layouts defined in `views.py`

jeremystretch:
Fixed items 1 & 2 above. 3 and 4 are false positives.

netbox/dcim/ui/panels.py
```diff
@@ -153,11 +159,36 @@ class DeviceTypePanel(panels.ObjectAttributesPanel):
     rear_image = attrs.ImageAttr('rear_image')
 
 
+class ModulePanel(panels.ObjectAttributesPanel):
+    device = attrs.RelatedObjectAttr('device', linkify=True)
+    device_type = attrs.RelatedObjectAttr('device.device_type', linkify=True, grouped_by='manufacturer')
+    module_bay = attrs.NestedObjectAttr('module_bay')
```
pheus:
```suggestion
    module_bay = attrs.NestedObjectAttr('module_bay', linkify=True)
```
_Small nit_: This was a link before, therefore I suggest to enable linkify.

pheus (commented):
Thanks for continuing the #20923 migration work !

_One quick clarification:_ I noticed the attribute label capitalization shifts from **Title Case** to **sentence case** in a few places (e.g. `Device Role` → `Device role`, `Serial Number` → `Serial number`). That _seems_ intentional and may be aligned with the new UI/layout styling, but I wanted to confirm: is this change deliberate as part of the migration/standardization effort?

Just want to make sure we’re being consistent with the intended label style across the UI. If not, we might want to keep the previous Title Case labels to avoid inconsistencies between pages.

I also left one small nit inline, but otherwise this looks good to me. Thanks again!

jeremystretch:
> is this change deliberate as part of the migration/standardization effort?

Yes; we're standardizing on normal case (e.g. "Device type" rather than "Device Type") for the attribute panels.

pheus (approved):
Thanks for confirmation and the fast fix!

PR #21373, opened 2026-02-06
Closes #21337: Refactor VirtualMachine view to UI layout

### LINKED ISSUE #21337: Add platform parent on virtual machines
### NetBox version

v4.5.1

### Feature type

### Proposed functionality

On virtual machines, when you select a platform that has a parent, only the child is displayed. This is not how platforms work on physical devices, where it displays the parent and the child.

Proposal is to make virtual machines display parent / child platform just like devices do.

**Physical Device:**
Image

**Virtual Machine:**
Image

### Use case

Make platform parent/child consistent with devices.

### Database changes

none.

### External dependencies

none.

pheus:
### Fixes: #21337

This PR migrates the **VirtualMachine** detail view to the new modular UI layout approach introduced in v4.5.0, bringing platform rendering in line with the parent/child behavior of devices.

#### Changes
- Refactors the VirtualMachine detail view to use `SimpleLayout`
- Adds standardized panels for **attributes**, **cluster**, and **resources**
- Splits the legacy template into reusable components for improved maintainability

Thanks for the guidance and review on the approach!


jeremystretch (approved):
Nice work @pheus!

PR #21401, opened 2026-02-11
Fixes #21196: `q` filter should match on primary IP only for IP address values

### LINKED ISSUE #21196: The device search filter (`q`) is too slow when handling thousands of devices with primary IPs assigned
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.0

### Python Version

3.12

### Steps to Reproduce

1. Begin with a database having several thousand devices populated, each with a primary IPv4 and IPv6 address assigned.
2. Attempt to search for a device using the `q` filter, e.g. `GET /api/dcim/devices/?q=foo`

### Expected Behavior

Matching results should be returned quickly.

### Observed Behavior

Depending on the number of devices with primary IPs assigned, this can take an inordinate amount of time. (Real world results have shown delays in excess of 30 seconds.)

This is due to the inclusion of the `primary_ip4` and `primary_ip6` IPAddresses in the underlying filter query, to match the `q` value against the assigned IP addresses. While clearly valuable functionality, this incurs a heavy performance cost in very large databases.

https://github.com/netbox-community/netbox/blob/3e2a26984fe07f394812664d89f82efd85f6d595/netbox/dcim/filtersets.py#L1332-L1341

jeremystretch:
### Fixes: #21196

Update the `search()` methods on the filtersets for Device and VirtualMachine to query against the `primary_ip4` & `primary_ip6` filters only if the given search value is a valid IP address.

### DIFF
--- a/netbox/virtualization/filtersets.py
+++ b/netbox/virtualization/filtersets.py
@@ -231,7 +231,7 @@ class VirtualMachineFilterSet(
     def search(self, queryset, name, value):
         if not value.strip():
             return queryset
-        qs_filter = queryset.filter(
+        qs_filter = Q(
             Q(name__icontains=value) |
             Q(description__icontains=value) |
             Q(comments__icontains=value) |
@@ -246,7 +246,7 @@ class VirtualMachineFilterSet(
                 qs_filter |= Q(primary_ip6__address__host__inet=ipaddress.ip)
         except (AddrFormatError, ValueError):
             pass
-        return qs_filter
+        return queryset.filter(qs_filter)
 
     def _has_primary_ip(self, queryset, name, value):
         params = Q(primary_ip4__isnull=False) | Q(primary_ip6__isnull=False)
--- a/netbox/virtualization/filtersets.py
+++ b/netbox/virtualization/filtersets.py
@@ -1,6 +1,8 @@
 import django_filters
+import netaddr
 from django.db.models import Q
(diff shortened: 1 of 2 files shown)


netbox/virtualization/filtersets.py
```diff
@@ -229,14 +231,22 @@ class Meta:
     def search(self, queryset, name, value):
         if not value.strip():
             return queryset
-        return queryset.filter(
+        qs_filter = queryset.filter(
```
jnovinger: This appears to be different than the logic you used in `DeviceFilterSet.search()`, in that there you're composing the `Q` objects all together with a logical "OR" before passing any to `.filter()`.

But, here, you're starting the initial set of `Q` filters directly in a `.filter()` call, meaning that `qs_filter` here is actually a `QuerySet` object rather than composed `Q` object.

I assume that's not intentional.

jnovinger (commented):
Left a couple of comments, but also curious about the apparent change of behavior resulting from switching from `__startswith` matching to exact IP address matching. Is that intentional? Any side-effects we should be aware of? Should we consider this a breaking change?

jeremystretch:
The fix for #21196 is that we're now only pulling in related IP addresses if the search value is a valid IPv4 or IPv6 address. Therefore we can no longer support the "starts with" lookups.

jnovinger (approved):

PR #21415, opened 2026-02-12
Fixes #21412: Defer monkey-patching until after settings have been loaded

jeremystretch:
### Fixes: #21412

This moves all the imports necessary for monkey-patching to the end of the `settings` module, to avoid interfering with the loading of configuration parameters.

This fix can be tested by checking out the PR and installing & enabling netbox-changes v0.4.2

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -11,14 +11,10 @@ from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.core.validators import URLValidator
 from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
-from rest_framework.utils import field_mapping
-from strawberry_django import pagination
-from strawberry_django.fields.field import StrawberryDjangoField
 
 from core.exceptions import IncompatiblePluginError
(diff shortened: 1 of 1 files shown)


pheus:
I noticed an additional way to reproduce / validate the underlying issue:

If `configuration.py` lists a plugin in `PLUGINS` that isn’t installed in the current venv (e.g. `netbox_acls`), I’d expect NetBox to fail fast with the plugin import error, e.g.:

```text
  File "/home/vagrant/netbox-dev/netbox/netbox/netbox/settings.py", line 875, in <module>
    raise ImproperlyConfigured(
django.core.exceptions.ImproperlyConfigured: Unable to import plugin netbox_acls: Module not found. Check that the plugin module has been installed within the correct Python environment.
```

However, without the fix in this PR, I instead hit:

```text
CommandError: You must set settings.ALLOWED_HOSTS if DEBUG is False.
```

With this PR applied, the expected “Unable to import plugin …” error is raised again (i.e. the real root cause isn’t masked).

pheus (approved):
Thanks for the fix!
This looks good to me.

PR #21421, opened 2026-02-12
Closes #21419: Improve query efficiency for MultipleChoiceFilter

### LINKED ISSUE #21419: Audit filters for unintentional application of `distinct=True`
### NetBox Version

v4.5.2

### Python Version

3.12

### Area(s) of Concern

- [x] User Interface
- [x] REST API
- [ ] GraphQL API
- [ ] Python ORM
- [ ] Other

### Details

We recently discovered while investigating a performance issue that django-filter's MultipleChoiceFilter (and by extension ModelMultipleChoiceFilter) passes `distinct=True` [by default](https://django-filter.readthedocs.io/en/stable/ref/filters.html#django_filters.filters.MultipleChoiceFilter). This introduces unnecessary overhead to the query in many cases: Typically, `distinct=True` is needed only for many-to-many relationships.

jeremystretch:
### Closes: #21419

Pass `distinct=False` to

* All MultipleChoiceFilters associated with a concrete model field
* All ModelMultipleChoiceFilters associated with a ForeignKey field

There may exist additional opportunities for performance improvements, but I want to keep the scope of this initial work limited as the changes are quite broad.

### DIFF
--- a/netbox/circuits/filtersets.py
+++ b/netbox/circuits/filtersets.py
@@ -209,6 +209,7 @@ class CircuitFilterSet(PrimaryModelFilterSet, TenancyFilterSet, ContactModelFilt
     )
     status = django_filters.MultipleChoiceFilter(
         choices=CircuitStatusChoices,
+        distinct=False,
         null_value=None
     )
     region_id = TreeNodeMultipleChoiceFilter(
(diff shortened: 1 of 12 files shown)


pheus (approved):
Nice enhancements, @jeremystretch !

PR #21423, opened 2026-02-12
21390 skip m2m processing for internal models to avoid extraneous ObectChange records

### LINKED ISSUE #21390: ChangeLog for second interface after connect circuit
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.0

### Python Version

3.12

### Steps to Reproduce

1. create 2 device ("device A" and "device B") with interface
2. create circuit with termination in same site/location with devices
3. connect first termination to "device A"
4. connect second termination to "device B" (this generate 2 changes to log for interface of "device A")

### Expected Behavior

no changes (or only one change) for interface of "device A" after connect circuit termination to second "device B"

### Observed Behavior

Netbox generate 2 changes for interface of "device A" (in this attachment is "eth0") with 1 requestID after connect termination

Image

arthanson:
### Fixes: #21390 

Issue is from CablePath and it having a SET_NULL on the FK relationship which in the `handle_deleted_object` m2m processing causes a save to the related object, but it has already been handled by Django.  CablePath is an internal model and for those (_netbox_private = True) models the m2m processing should probably be skipped.

Skips the m2m handling in `handle_deleted_object`  Django's on_delete handlers already take care of the database integrity. This appears to only effect CablePath as other _netbox_private models - CablePath is the only _netbox_private model that has reverse FK relationships from ChangeLoggingMixin models with on_delete=SET_NULL.

### DIFF
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -209,22 +209,28 @@ def handle_deleted_object(sender, instance, **kwargs):
     # for the forward direction of the relationship, ensuring that the change is recorded.
     # Similarly, for many-to-one relationships, we set the value on the related object to None
     # and save it to trigger a change record on that object.
-    for relation in instance._meta.related_objects:
-        if type(relation) not in [ManyToManyRel, ManyToOneRel]:
-            continue
-        related_model = relation.related_model
-        related_field_name = relation.remote_field.name
-        if not issubclass(related_model, ChangeLoggingMixin):
-            # We only care about triggering the m2m_changed signal for models which support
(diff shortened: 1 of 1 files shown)


arthanson:
@jeremystretch also tagging you on this review as this is a pretty core change to the m2m code in the delete handler that you had implemented.  Wanted to have it verified.

jeremystretch (approved):
Seems to be safe to omit private models from the M2M handling. I can't find any instances where their omission is concerning.

PR #21432, opened 2026-02-13
Closes #21016: Add missing MPTT tree indexes

### LINKED ISSUE #21016: Restore missing SQL indexes for MPTT fields
### Proposed Changes

All models which employ MPTT (from `django-mptt`) should have SQL indexes for `('tree_id', 'left')`, however these are not created by the database migrations. This seems to have been partially solved by django-mptt/django-mptt#845 in v0.18.0, however the indexes are not automatically created for most models.

This was originally tracked under #20954 but has been split into a separate issue for investigation.

### Justification

Adding these indexes should improve overall query performance for MPTT-enabled models.

jnovinger:
### Fixes: #21016

Upgrade django-mptt to 0.18.0 and add the missing composite `(tree_id, lft)` indexes to all MPTT models.

### Root Cause

django-mptt dynamically adds indexes to `_meta.indexes` via its metaclass, but doesn't update `_meta.original_attrs['indexes']`. Django's migration autodetector only checks `original_attrs`, so it never detected the missing indexes for models without an explicit `indexes` declaration in their Meta class.

This is a known issue: django-mptt/django-mptt#682

### Solution

Add an empty `indexes = ()` tuple to each affected model's Meta class. This triggers Django's migration detection for the dynamically-added MPTT indexes. We cannot define the indexes explicitly because the MPTT fields (`tree_id`, `lft`) don't exist when the Meta class is evaluated.

### Affected Models (11)

- Region, SiteGroup, Location
- DeviceRole, Platform
- ModuleBay, InventoryItem, InventoryItemTemplate
- TenantGroup, ContactGroup
- WirelessLANGroup

### Changes

- Upgrade django-mptt 0.17.0 → 0.18.0
- Add `indexes = ()` with explanatory comment to 9 model Meta classes
- 3 new migrations adding composite indexes to all 11 MPTT models


pheus (dismissed):
This looks good! Thanks, @jnovinger !

One small suggestion: it could be helpful to add a brief comment near `NestedGroupModel` to capture the reasoning for future maintainers. For example:

> **Note:** django-mptt injects the `(tree_id, lft)` index dynamically, but Django’s migration autodetector won’t reliably detect it unless the *concrete* model explicitly declares `Meta.indexes` (even as an empty tuple). See #21016 / django-mptt#682.

Just to be clear, this wouldn’t replace the per-model `indexes = ()` additions, but it may save someone time the next time we introduce a new `NestedGroupModel` subclass.

pheus (approved):
Awesome! Thanks!

PR #21436, opened 2026-02-13
Add {module_path} placeholder for nested module component templates

mrmrcoleman:
## Summary

Introduces a new `{module_path}` placeholder for module component templates that automatically expands to the full position path for nested modules, joined by `/`.

**Fixes:** #20474, #19796

## Problem

When module types are nested (e.g. a line card containing SFP bays), users need component names that reflect the full position path — for example `SFP 1/2` for an SFP in slot 2 of line card 1.

The existing `{module}` placeholder requires one token per nesting level (`{module}/{module}`), which means the same ModuleType cannot be reused at different depths. An SFP module type designed for depth 1 (`SFP {module}`) fails validation at depth 2 with: *"Cannot install module with placeholder values in a module bay tree 2 in tree but 1 placeholders given."*

This forces users to create duplicate ModuleTypes for every possible nesting level, which defeats the purpose of modular design.

## Solution

A new `{module_path}` placeholder that expands to the full path regardless of depth:

| Depth | Positions | `{module_path}` expands to |
|-------|-----------|----------------------------|
| 1     | "1"       | `1`                        |
| 2     | "1", "2"  | `1/2`                      |
| 3     | "1", "2", "3" | `1/2/3`              |

A single ModuleType with template name `SFP {module_path}` now works at any nesting depth without modification.

### `{module}` behavior is unchanged

The existing `{module}` placeholder continues to work exactly as before — level-by-level replacement where the token count must match the nesting depth. No existing workflows are affected.

### Validation rules (form layer)

- `{module_path}` can only appear once per template attribute
- `{module}` and `{module_path}` cannot be mixed in the same attribute
- Multiple `{module}` tokens must still match tree depth exactly

## Changes

| File | Change |
|------|--------|
| `dcim/constants.py` | Added `MODULE_PATH_TOKEN` and `MODULE_TOKEN_SEPARATOR` constants |
| `dcim/utils.py` | New `resolve_module_placeholders()` — centralized substitution logic for both placeholders |
| `dcim/models/device_component_templates.py` | Refactored `resolve_name()` and `resolve_label()` to use shared `_resolve_placeholders()` helper |
| `dcim/forms/common.py` | Extracted `_validate_module_tokens()` method; updated `clean()` to validate and resolve both placeholder types |
| `dcim/tests/test_models.py` | Comprehensive test coverage (see below) |
| `docs/models/dcim/moduletype.md` | Documentation for `{module_path}` |

## Testing

All existing tests pass. Added tests covering:

- `{module}` at depth 1 (backward compat, single token → position)
- `{module}/{module}` at depth 2 (backward compat, level-by-level)
- `{module_path}` at depths 1, 2, and 3
- Label substitution with `{module_path}`
- Non-interface components (ConsolePort) with `{module_path}`
- Positions containing slashes (`0/1` + `2` → `0/1/2`)
- Validation: single `{module}` rejected at depth > 1
- Validation: token count mismatch (too many / too few)
- Validation: mixed `{module}` + `{module_path}` rejected
- Validation: duplicate `{module_path}` rejected

## Not addressed

- #20467 (`{module}` in the `position` field of ModuleBayTemplates) — this is a separate issue requiring changes to position resolution, which is outside the scope of this PR.


rhartmann:
I am absolute looking forward to a solution for nested modules, as I have issues modelling sub-modules on several cases. However, `MODULE_TOKEN_SEPARATOR` should definitely not be hardcoded. Some vendors use `/`, some use `-` or `.`.

There's probably a couple of ways to define the module path separator. Some that come to mind:

- Extend the syntax to `{module_path:-}` where `-` is the separator to use. This mimics the [Python Format String Syntax](https://docs.python.org/3/library/string.html#format-string-syntax)
- Introduce a new field `module_path_separator` to the data model. Either on `dcim.moduletype` (use same separator for all module bays) or `dcim.modulebay` / `dcim.modulebaytemplate` (more flexibility, each bay may define a separator). Or maybe on both to allow overlays

A completely different approach (though I don't know if that even works): Add a boolean field `inherit_parent_position` (default `false` for backwards compatibility) to `dcim.modulebay`. This would only apply to bays added to modules (not devices). When a module has bays with this enabled, any module type inserted into that sub-bay can simply use `{module}` like before, which then resolves to whatever the module bay position resolved to before (which itself could use `{module}` with any or no separator as desired. This should give maximum flexibility and does not require a new token `{module_path}`.

netbox/dcim/utils.py
```diff
@@ -85,13 +123,13 @@ def update_interface_bridges(device, interface_templates, module=None):
             interface.save()
 
 
-def create_port_mappings(device, device_or_module_type, module=None):
+def create_port_mappings(device, device_type, module=None):
```
arthanson: Is this rename and change to the doc-string correct?  It is still called from [modules.py](https://github.com/mrmrcoleman/netbox/blob/fa2facd118467d3245e1d274a04ae7721ef106ba/netbox/dcim/models/modules.py#L365) with a module_type. The old comment and naming seem correct.

netbox/dcim/forms/common.py
```diff
@@ -82,6 +83,50 @@ def _get_module_bay_tree(self, module_bay):
...
+        if not has_module_token and not has_module_path_token:
+            return
+
+        # Installing modules with placeholders require that the bay has a position value
+        if not module_bay.position:
+            raise forms.ValidationError(
+                _("Cannot install module with placeholder values in a module bay with no position defined.")
+            )
+
+        # Cannot mix {module} and {module_path} in the same attribute
```
arthanson: The below can be cleaned up and simplified:
```
        # Cannot mix {module} and {module_path} in the same attribute; validate {module_path} usage
        if has_module_path_token:
            if has_module_token:
                raise forms.ValidationError(
                    _("Cannot mix {module} and {module_path} placeholders in the same template attribute.")
                )
            if template_name.count(MODULE_PATH_TOKEN) > 1:
                raise forms.ValidationError(
                    _("The {module_path} placeholder can only be used once per template.")
                )

        # Validate {module} - token count must match depth exactly
        elif has_module_token:
            if (token_count := template_name.count(MODULE_TOKEN)) != depth:
                raise forms.ValidationError(
                    _(
                        "Cannot install module with placeholder values in a module bay tree {level} deep "
                        "but {tokens} placeholders given."
                    ).format(
                        level=depth, tokens=token_count
                    )
                )

```

netbox/dcim/utils.py
```diff
@@ -4,6 +4,44 @@
...
+    if not text:
+        return text
+
+    result = text
+
+    # Handle {module_path} first — it must be resolved before {module}
+    # because '{module}' is a substring of '{module_path}'.
+    if MODULE_PATH_TOKEN in result:
+        full_path = MODULE_TOKEN_SEPARATOR.join(positions) if positions else ''
+        result = result.replace(MODULE_PATH_TOKEN, full_path)
```
arthanson: Since both MODULE_PATH_TOKEN and MODULE_TOKEN Cannot be in the same string it is probably safer to just return the value here and not fall-through.

netbox/dcim/utils.py
```diff
@@ -4,6 +4,44 @@
...
+    """
+    if not text:
+        return text
+
+    result = text
+
+    # Handle {module_path} first — it must be resolved before {module}
+    # because '{module}' is a substring of '{module_path}'.
+    if MODULE_PATH_TOKEN in result:
+        full_path = MODULE_TOKEN_SEPARATOR.join(positions) if positions else ''
```
arthanson, line 35: I'd make a separate FR to make MODULE_TOKEN_SEPARATOR a configurable param.

netbox/dcim/tests/test_models.py
```diff
@@ -772,244 +777,855 @@ def test_module_bay_recursion(self):
...
+            module_bay=sfp_bay,
+            module_type=sfp_type
+        )
+
+        # Verify interface name resolves to just "1"
+        interface = sfp_module.interfaces.first()
+        self.assertEqual(interface.name, 'SFP 1')
+        self.assertEqual(interface.label, '1')
+
+    def test_depth_1_single_token_no_extra_slashes(self):
```
arthanson: I think this is already tested in the test above?

arthanson (changes requested):
Very close, couple cleanup items.

sleepinggenius2:
Definitely agree with `MODULE_TOKEN_SEPARATOR` not being hardcoded. We have devices that use `/`, `-`, and `.` and they are all able to use the same pluggables, so this would only solve the problem for 1 out of 6+ duplicates for each pluggable that we have to model today.

rhartmann:
I did some tests regarding inheritance of parent module position. I came up with two variants. Both worked for my tests (which certainly didn't cover all possible cases/combinations). Maybe someone is interested and willing to put the pieces together and write some docs and tests.

# New attribute: Inherit parent position

This defaults to `True` for inheritance. If an attribute `inherit_parent_position` was added to the Module form, it would use that instead.

```patch
diff --git a/netbox/dcim/forms/common.py b/netbox/dcim/forms/common.py
index a3a781be5..3c28a1c06 100644
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -70,11 +70,14 @@ class InterfaceCommonForm(forms.Form):

 class ModuleCommonForm(forms.Form):

-    def _get_module_bay_tree(self, module_bay):
+    def _get_module_bay_tree(self, module_bay, inherit_parent_position):
         module_bays = []
         while module_bay:
             module_bays.append(module_bay)
             if module_bay.module:
+                if inherit_parent_position:
+                    # Read parent position, replace token, pass to child
+                    module_bay.position = module_bay.position.replace(MODULE_TOKEN, module_bay.module.module_bay.position, 1)
                 module_bay = module_bay.module.module_bay
             else:
                 module_bay = None
@@ -87,6 +90,7 @@ class ModuleCommonForm(forms.Form):

         replicate_components = self.cleaned_data.get('replicate_components')
         adopt_components = self.cleaned_data.get('adopt_components')
+        inherit_parent_position = self.cleaned_data.get('inherit_parent_position', True)
         device = self.cleaned_data.get('device')
         module_type = self.cleaned_data.get('module_type')
         module_bay = self.cleaned_data.get('module_bay')
@@ -100,7 +104,7 @@ class ModuleCommonForm(forms.Form):
             self.instance._disable_replication = True
             return

-        module_bays = self._get_module_bay_tree(module_bay)
+        module_bays = self._get_module_bay_tree(module_bay, inherit_parent_position)

         for templates, component_attribute in [
                 ("consoleporttemplates", "consoleports"),
@@ -126,18 +130,21 @@ class ModuleCommonForm(forms.Form):
                             _("Cannot install module with placeholder values in a module bay with no position defined.")
                         )

-                    if len(module_bays) != template.name.count(MODULE_TOKEN):
-                        raise forms.ValidationError(
-                            _(
-                                "Cannot install module with placeholder values in a module bay tree {level} in tree "
-                                "but {tokens} placeholders given."
-                            ).format(
-                                level=len(module_bays), tokens=template.name.count(MODULE_TOKEN)
+                    if inherit_parent_position:
+                        resolved_name = resolved_name.replace(MODULE_TOKEN, module_bays[-1].position)
+                    else:
+                        if len(module_bays) != template.name.count(MODULE_TOKEN):
+                            raise forms.ValidationError(
+                                _(
+                                    "Cannot install module with placeholder values in a module bay tree {level} in tree "
+                                    "but {tokens} placeholders given."
+                                ).format(
+                                    level=len(module_bays), tokens=template.name.count(MODULE_TOKEN)
+                                )
                             )
-                        )

-                    for module_bay in module_bays:
-                        resolved_name = resolved_name.replace(MODULE_TOKEN, module_bay.position, 1)
+                        for module_bay in module_bays:
+                            resolved_name = resolved_name.replace(MODULE_TOKEN, module_bay.position, 1)

                 existing_item = installed_components.get(resolved_name)

diff --git a/netbox/dcim/models/device_component_templates.py b/netbox/dcim/models/device_component_templates.py
index 0d54331d8..ab9eafb20 100644
--- a/netbox/dcim/models/device_component_templates.py
+++ b/netbox/dcim/models/device_component_templates.py
@@ -177,16 +177,23 @@ class ModularComponentTemplateModel(ComponentTemplateModel):
         modules.reverse()
         return modules

+    def _resolve_module_position(self, module, value):
+        modules = self._get_module_tree(module)
+        if not module or not modules:
+            return value
+        if getattr(module, 'inherit_parent_position', True):
+            return value.replace(MODULE_TOKEN, modules[-1].module_bay.position)
+        else:
+            for module in modules:
+                value = value.replace(MODULE_TOKEN, module.module_bay.position, 1)
+            return value
+
     def resolve_name(self, module):
         if MODULE_TOKEN not in self.name:
             return self.name

         if module:
-            name = self.name
-                name = name.replace(MODULE_TOKEN, module.module_bay.position, 1)
-            return name
+            return self._resolve_module_position(module, self.name)
         return self.name

     def resolve_label(self, module):
@@ -194,11 +201,7 @@ class ModularComponentTemplateModel(ComponentTemplateModel):
             return self.label

         if module:
-            return label
+            return self._resolve_module_position(module, self.label)
         return self.label

```

# Always inherit parent position

This is surprisingly simple. It replaces `MODULE_TOKEN` in module name, label and position fields when building the module bay tree, where each replacement uses the parent module bay position. Finally, when adopting or cloning components, the `MODULE_TOKEN` is also replaced (only the name attribute of the components).

```patch
diff --git a/netbox/dcim/forms/common.py b/netbox/dcim/forms/common.py
index a3a781be5..4608542de 100644
--- a/netbox/dcim/forms/common.py
+++ b/netbox/dcim/forms/common.py
@@ -75,6 +75,11 @@ class ModuleCommonForm(forms.Form):
         while module_bay:
             module_bays.append(module_bay)
             if module_bay.module:
+                parent_position = module_bay.module.module_bay.position:
+                if parent_position:
+                    module_bay.name = module_bay.name.replace(MODULE_TOKEN, parent_position, 1)
+                    if module_bay.position:
+                        module_bay.position = module_bay.position.replace(MODULE_TOKEN, parent_position, 1)
+                    if module_bay.label:
+                        module_bay.label = module_bay.label.replace(MODULE_TOKEN, parent_position, 1)
                 module_bay = module_bay.module.module_bay
             else:
                 module_bay = None
@@ -126,18 +131,7 @@ class ModuleCommonForm(forms.Form):
                             _("Cannot install module with placeholder values in a module bay with no position defined.")
                         )

-                    if len(module_bays) != template.name.count(MODULE_TOKEN):
-                        raise forms.ValidationError(
-                            _(
-                                "Cannot install module with placeholder values in a module bay tree {level} in tree "
-                                "but {tokens} placeholders given."
-                            ).format(
-                                level=len(module_bays), tokens=template.name.count(MODULE_TOKEN)
-                            )
-                        )
-
-                    for module_bay in module_bays:
-                        resolved_name = resolved_name.replace(MODULE_TOKEN, module_bay.position, 1)
+                    resolved_name = resolved_name.replace(MODULE_TOKEN, module_bay.position)

                 existing_item = installed_components.get(resolved_name)

diff --git a/netbox/dcim/models/device_component_templates.py b/netbox/dcim/models/device_component_templates.py
index 0d54331d8..ba0031aae 100644
--- a/netbox/dcim/models/device_component_templates.py
+++ b/netbox/dcim/models/device_component_templates.py
@@ -177,16 +177,20 @@ class ModularComponentTemplateModel(ComponentTemplateModel):
         modules.reverse()
         return modules

+    def _resolve_module_position(self, module, value):
+        modules = self._get_module_tree(module)
+        if not module or not modules:
+            return value
+        for module in modules:
+            value = value.replace(MODULE_TOKEN, module.module_bay.position, 1)
+        return value
+
     def resolve_name(self, module):
         if MODULE_TOKEN not in self.name:
             return self.name

         if module:
-            name = self.name
-                name = name.replace(MODULE_TOKEN, module.module_bay.position, 1)
-            return name
+            return self._resolve_module_position(module, self.name)
         return self.name

     def resolve_label(self, module):
@@ -194,11 +198,7 @@ class ModularComponentTemplateModel(ComponentTemplateModel):
             return self.label

         if module:
-            return label
+            return self._resolve_module_position(module, self.label)
         return self.label

```

mrmrcoleman:
@sleepinggenius2 see separate FR here: https://github.com/netbox-community/netbox/issues/21667

arthanson (changes requested):
Should retarget against feature. Also, you said you might have a different implementation.

mrmrcoleman:
Closing this PR to split the work into separate PRs: a bug fix for #20474 against main (v4.5.6) and a feature PR for #19796 against feature (v4.6.0).

PR #21460, opened 2026-02-17
Closes #21459: Avoid prefetching data for hidden table columns

### LINKED ISSUE #21459: Avoid prefetching data for hidden table columns
### NetBox Version

v4.5.3

### Python Version

3.12

### Area(s) of Concern

- [x] User Interface
- [ ] REST API
- [ ] GraphQL API
- [ ] Python ORM
- [ ] Other

### Details

It seems that when we introduced saved table configs in NetBox v4.3 (see #14591), dynamic prefetching for table columns was broken in a way that forced prefetching even for hidden columns. his means that tables still rendered correctly, but extraneous database queries are issued in some cases when rendering a table.

The implementation of saved table configs can be adapted to avoid these extraneous queries.

jeremystretch:
### Closes: #21459

- Move the logic for determining pre-fetched fields to _after_ columns have been hidden (in the new `_apply_prefetching()` method)
- Call `_apply_prefetching()` from `configure()`
- Add a test for dynamic table prefetching
- Stop passing `user` when initializing the table (no longer needed)

### DIFF
--- a/netbox/core/views.py
+++ b/netbox/core/views.py
@@ -682,7 +682,7 @@ class PluginListView(BasePluginView):
 
         plugins = [plugin for plugin in plugins if not plugin.hidden]
 
-        table = CatalogPluginTable(plugins, user=request.user)
+        table = CatalogPluginTable(plugins)
         table.configure(request)
 
         # If this is an HTMX request, return only the rendered table HTML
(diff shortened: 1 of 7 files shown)


arthanson (changes requested):
This all looks good, I think there are a few places where user needs to be removed from the tables:

netbox/netbox/views/generic/feature_views.py
   * line 154 - JournalEntryTable
   * line 221 JobTable

netbox/extras/views.py
   * line 1629 ScriptResultsTable
   * line 1653 reportResultsTable

netbox/core/views.py
   * line 685 CatalogPluginTable
   * line 708 PluginVersionTable

arthanson (approved):

PR #21474, opened 2026-02-18
Fixes #21411: Enable Ruff RET rules for consistent return behavior

pheus:
### Fixes: #21411

This PR enables Ruff’s `RET` rules to make return behavior more explicit and consistent across the codebase.

- Simplifies control flow by removing redundant `else` blocks after `return`
- Adds explicit fallback returns (`return None` / equivalent) where functions previously relied on implicit fallthrough

No functional changes are intended.

Thanks for reviewing!

### DIFF
--- a/netbox/core/api/schema.py
+++ b/netbox/core/api/schema.py
@@ -39,7 +39,7 @@ class ChoiceFieldFix(OpenApiSerializerFieldExtension):
         if direction == 'request':
             return build_cf
 
-        elif direction == "response":
+        if direction == "response":
             value = build_cf
             label = {
                 **build_basic_type(OpenApiTypes.STR),
@@ -52,6 +52,10 @@ class ChoiceFieldFix(OpenApiSerializerFieldExtension):
                     "label": label
                 }
             )
+
+        # TODO: This function should never implicitly/explicitly return `None`
+        # The fallback should be well-defined (drf-spectacular expects request/response naming).
+        return None
 
 
 def viewset_handles_bulk_create(view):
@@ -75,8 +79,7 @@ class NetBoxAutoSchema(AutoSchema):
     def is_bulk_action(self):
         if hasattr(self.view, "action") and self.view.action in BULK_ACTIONS:
             return True
-        else:
-            return False
+        return False
 
     def get_operation_id(self):
         """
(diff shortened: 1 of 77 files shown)


netbox/core/api/schema.py
```diff
@@ -52,6 +52,8 @@ def map_serializer_field(self, auto_schema, direction):
                     "label": label
                 }
             )
+        # Fallback (drf-spectacular uses "request"/"response")
+        return build_cf
```
jeremystretch: Functional change

netbox/dcim/api/views.py
```diff
@@ -222,24 +222,26 @@ def elevation(self, request, pk=None):
...
+        if page is not None:
+            rack_units = serializers.RackUnitSerializer(page, many=True, context={'request': request})
+            return self.get_paginated_response(rack_units.data)
 
-            page = self.paginate_queryset(elevation)
-            if page is not None:
-                rack_units = serializers.RackUnitSerializer(page, many=True, context={'request': request})
-                return self.get_paginated_response(rack_units.data)
+        rack_units = serializers.RackUnitSerializer(elevation, many=True, context={'request': request})
+        return Response(rack_units.data)
```
jeremystretch: Functional change

netbox/dcim/graphql/gfk_mixins.py
```diff
@@ -117,7 +119,7 @@ class Meta:
     def resolve_type(cls, instance, info: Info):
         if type(instance) is CircuitTermination:
             return CircuitTerminationType
-        if type(instance) is ConsolePortType:
+        if type(instance) is ConsolePort:
```
jeremystretch: We should probably capture this as a bug and ship the fix separately.

jeremystretch (changes requested):
I noticed a few instances where a change results in a material alteration to the application logic. Let's avoid these and stick with purely stylistic updates so as not to risk inadvertently introducing unexpected behavior.

We could open separate issues for these fixes if warranted, or just drop a `TODO` comment for each citing the recommended tweak.

netbox/dcim/graphql/gfk_mixins.py
pheus: I'm going to remove this unrelated fix and open a bug report for this. Thanks!

pheus:

I’ve removed the three changes that materially altered application logic so this PR stays focused on purely stylistic `RET` cleanups and avoids any risk of unintended behavior changes.
For the spots where a missing return seems like it might warrant a follow-up, I added TODO comments noting the suggested adjustment (keeping the current behavior by returning `None`).

Appreciate you calling these out.

jeremystretch (approved):

netbox/dcim/graphql/gfk_mixins.py
pheus: For future reference: #21478

PR #21476, opened 2026-02-18
Closes #21331: Emit deprecation warning on use of querystring template tag

### LINKED ISSUE #21331: Deprecate our custom `querystring` template tag
### Deprecated Functionality

Deprecate the [`querystring`](https://github.com/netbox-community/netbox/blob/bec5ecf6a905945aa9ee2572f74dfe14b003b228/netbox/utilities/templatetags/helpers.py#L363) template tag.

### Scheduled removal

v4.7.0

### Justification

Django 5.1 introduced a new built-in [`querystring` template tag](https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#querystring).

### Impact

Any plugins which import NetBox's local `querystring` template tag will need to be updated to import from Django instead.

bctiemann:
### Fixes: #21331 

Emits a warning if the custom `querystring` template tag is used:
```
The querystring template tag is deprecated and will be removed in a future release. Use the built-in Django querystring tag instead.
```


netbox/utilities/templatetags/helpers.py
```diff
@@ -363,6 +366,10 @@ def querystring(request, **kwargs):
     """
     Append or update the page number in a querystring.
     """
+    logger.warning(
```
jnovinger: I think you want to be using `warnings.warn()` instead of `logger.warn()`, yeah? Like in [`settings.py`](https://github.com/netbox-community/netbox/blob/main/netbox/netbox/settings.py#L275-L278).

jnovinger (changes requested):

netbox/utilities/templatetags/helpers.py
bctiemann: Good call; fixed. Note I'm using `FutureWarning` instead of `DeprecationWarning` as the latter is suppressed unless explicitly using python development mode, and we want this to be highly visible.

bctiemann:
@claude /review

jnovinger (approved):

PR #21482, opened 2026-02-19
Fixes #21481: Fix Rack detail view to display Facility ID

### LINKED ISSUE #21481: Facility ID doesn't show in rack page.
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.3

### Python Version

3.12

### Steps to Reproduce

1. Create a rack with 'Facility ID'

### Expected Behavior

The 'Facility ID' field on the rack view page must be filled in.

### Observed Behavior

The 'Facility ID' field on the rack view page is empty.

pheus:
### Fixes: #21481

Updates the rack detail view to reference the correct `facility_id` model field when rendering “Facility ID”. This restores the expected behavior so racks created with a Facility ID show it properly on the rack view page.

### DIFF
--- a/netbox/dcim/ui/panels.py
+++ b/netbox/dcim/ui/panels.py
@@ -44,7 +44,7 @@ class RackPanel(panels.ObjectAttributesPanel):
     site = attrs.RelatedObjectAttr('site', linkify=True, grouped_by='group')
     location = attrs.NestedObjectAttr('location', linkify=True)
     name = attrs.TextAttr('name')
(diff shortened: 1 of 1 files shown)


netbox/dcim/ui/panels.py
kukabu, line 12: also?

kukabu, line 24: also?

arthanson (approved):

Issue #21304, opened 2026-01-26
Deprecate the `housekeeping` management command


jeremystretch:
### Deprecated Functionality

Formally deprecate the `housekeeping` management command and note its planned removal in NetBox v4.7.

### Scheduled removal

v4.7.0

### Justification

The housekeeping command was superseded by the system housekeeping job in NetBox v4.4.0 and is no longer needed.

### Impact

Any deployments which still utilize the `housekeeping` management command need to be updated to not use it. (The system housekeeping job runs automatically.)

PR #21483, opened 2026-02-19
Closes: #21304 - Add stronger deprecation warning on use of housekeeping management command

bctiemann:
### Closes: #21304

The `housekeeping` command was already deprecated in that it printed a message warning the user that the command is no longer necessary and will be removed in a future release. This change strengthens it to explicitly mention `DEPRECATION WARNING` and should be included in official release notes as a warning of scheduled removal in NetBox v4.7.0.

### DIFF
--- a/netbox/extras/management/commands/housekeeping.py
+++ b/netbox/extras/management/commands/housekeeping.py
@@ -18,6 +18,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         self.stdout.write(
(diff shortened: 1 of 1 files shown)


jeremystretch (changes requested):
Let's raise a [deprecation warning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) to ensure that the deprecation notice is escalated sufficiently, rather than hoping someone notices the std output.

jeremystretch (approved):

PR #21491, opened 2026-02-19
Closes: #21284 -  Add deprecation note to webhooks documentation

bctiemann:
### Closes: #21284

Adds a deprecation note to the webhooks documentation.

### DIFF
--- a/docs/plugins/development/webhooks.md
+++ b/docs/plugins/development/webhooks.md
@@ -43,6 +43,9 @@ The resulting webhook payload will look like the following:
 }
 ```
 
+!!! note "Deprecation of legacy fields"
+    The "request_id" and "username" fields in the webhook payload above are deprecated and should no longer be used. Support for them will be removed in a future release.
+
 !!! note "Consider namespacing webhook data"
     The data returned from all webhook callbacks will be compiled into a single `context` dictionary. Any existing keys within this dictionary will be overwritten by subsequent callbacks which include those keys. To avoid collisions with webhook data provided by other plugins, consider namespacing your plugin's data within a nested dictionary as such:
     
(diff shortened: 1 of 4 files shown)


docs/plugins/development/webhooks.md
```diff
@@ -43,6 +43,9 @@ The resulting webhook payload will look like the following:
 }
 ```
 
+!!! note "Deprecation of legacy fields"
+    The "request_id" and "username" fields in the webhook payload above are deprecated and should no longer be used. Support for them will be removed in a future release.
+
```
pheus, line 48: Thanks for adding this deprecation callout.

Since these fields are scheduled for removal in **v4.7.0**, would it make sense to switch this from a `note` to a `warning` so it’s harder to miss? Also, it might help readers if we point to the replacements on the `request` object (as noted in the issue), e.g. `request.user.username` and the request ID.

Something like:

```markdown
!!! warning "Deprecation of legacy fields"
    The `request_id` and `username` fields in the webhook payload above are deprecated and will be removed in NetBox v4.7.0.

    Use `request.user.username` and `request.request_id` from the `request` object included in the callback context instead.
```

pheus (changes requested):
Thanks!

I found two other docs locations that still reference these fields, so it’d be great if we could update them as part of this PR:

- https://github.com/netbox-community/netbox/blob/main/docs/integrations/webhooks.md#available-context
- https://github.com/netbox-community/netbox/blob/main/docs/models/extras/webhook.md#context-data

The goal would be the same: mark `username` / `request_id` as deprecated (ideally as a warning given the v4.7.0 removal), and point readers to `request.user.username` and `request.request_id` on the `request` object instead.

docs/plugins/development/webhooks.md
bctiemann: Good call, thanks.

pheus (approved):
Thanks, @bctiemann !

PR #21495, opened 2026-02-19
Fixes: #21456 - Improve config_context rendering with GraphQL

grische:
### Fixes: #21456

It seems that GraphQL API, unlike the REST API, does not make use of the performance improvement from using ConfigContextModelQuerySet (https://github.com/netbox-community/netbox/pull/5266). This PR adds the necessary config_context annotation to GraphQL queries that request config_context.

 It speeds up queries by a factor of >10x with a simple netbox-demo-data setup and even surpases the REST API performance (due it's much smaller overall query) when using:
```graphql
{
  device_list {
    id
    config_context
  }
}
```

### DIFF
--- a/netbox/extras/graphql/mixins.py
+++ b/netbox/extras/graphql/mixins.py
@@ -21,6 +21,13 @@ if TYPE_CHECKING:
 @strawberry.type
 class ConfigContextMixin:
 
+    @classmethod
+    def get_queryset(cls, queryset, info: Info, **kwargs):
+        queryset = super().get_queryset(queryset, info, **kwargs)
+        if hasattr(queryset, 'annotate_config_context_data'):
+            return queryset.annotate_config_context_data()
+        return queryset
+
     @strawberry_django.field
     def config_context(self) -> strawberry.scalars.JSON:
         return self.get_config_context()
--- a/netbox/extras/graphql/mixins.py
+++ b/netbox/extras/graphql/mixins.py
@@ -24,11 +24,16 @@ class ConfigContextMixin:
     @classmethod
     def get_queryset(cls, queryset, info: Info, **kwargs):
         queryset = super().get_queryset(queryset, info, **kwargs)
-        if hasattr(queryset, 'annotate_config_context_data'):
+
+        # If `config_context` is requested, call annotate_config_context_data() on the queryset
+        selected = {f.name for f in info.selected_fields[0].selections}
+        if 'config_context' in selected and hasattr(queryset, 'annotate_config_context_data'):
             return queryset.annotate_config_context_data()
+
         return queryset
 
-    @strawberry_django.field
+    # Ensure `local_context_data` is fetched when `config_context` is requested
+    @strawberry_django.field(only=['local_context_data'])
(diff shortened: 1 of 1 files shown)


netbox/extras/graphql/mixins.py
```diff
@@ -14,14 +14,20 @@
...
 
 @strawberry.type
 class ConfigContextMixin:
 
+    @classmethod
+    def get_queryset(cls, queryset, info: Info, **kwargs):
+        queryset = super().get_queryset(queryset, info, **kwargs)
+        if hasattr(queryset, 'annotate_config_context_data'):
+            return queryset.annotate_config_context_data()
+        return queryset
```
jeremystretch, line 33: Claude pointed out that our current approach _always_ compiles config context data, even when not requested. We can adapt this slightly to call `annotate_config_context_data()` only when `config_context` is requested. This improves performance for non-context-data requests as well.

```suggestion
        queryset = super().get_queryset(queryset, info, **kwargs)

        # If `config_context` is requested, call annotate_config_context_data() on the queryset
        selected = {f.name for f in info.selected_fields[0].selections}
        if 'config_context' in selected and hasattr(queryset, 'annotate_config_context_data'):
            return queryset.annotate_config_context_data()

        return queryset
```

netbox/extras/graphql/mixins.py
```diff
@@ -14,14 +14,20 @@
...
+
     @strawberry_django.field
```
jeremystretch: I noticed that it's still triggering additional queries for each device for `local_context_data` if that field isn't included in the GraphQL request. We can add a hint to ensure this field is always fetched when `config_context` is requested.

```suggestion
    # Ensure `local_context_data` is fetched when `config_context` is requested
    @strawberry_django.field(only=['local_context_data'])
```

jeremystretch (changes requested):
Thanks for this @grische, it's a huge improvement! No issues with your approach, but after a bit of analysis I have two suggestions for improving performance a bit further. LMK what you think.

netbox/extras/graphql/mixins.py
grische: That's a tremendous improvement. I don't fully understand it though. 

Without the fix:
- If I do not include `local_context_data` in my query, it takes ~500-550ms on a test set
- if I **do include** `local_context_data` in a simple query, it takes ~200-250ms on the same test set ❓ 

With your suggested fix:
- If I do not include `local_context_data` in my query, it takes ~70ms on a test set
- If I do include `local_context_data` in my query, it has no measurable difference (~70ms)

netbox/extras/graphql/mixins.py
grische: That's a nice way, thanks a lot for that!
I tried something similar but I wasn't sure how to properly find out if config_context was included and accepted it because the performance improvement was so vast and the impact so little.

grische:
Do you want me to squash the commits with you as a co-author? Or will this happen anyway as part of the merge?

jeremystretch (approved):
Thanks @grische! (I'll squash the commits on merge.)

PR #21519, opened 2026-02-25
clarify E501 enforcement

kartik-lamkhade:
Fixes #21517

Update the style guide to clarify that Ruff rule E501 (line length) is enforced with a maximum of 120 characters, aligning the documentation with the current ruff.toml configuration


jeremystretch (approved):
Thanks @kartik-lamkhade!

PR #21520, opened 2026-02-25
Add new Ethernet types for 10GE and 40GE

### LINKED ISSUE #21394: Add missing interface types
### NetBox version

4.4.8

### Feature type

Data model extension

### Proposed functionality

I've run into two types of interfaces that aren't included in the list of available interfaces
10gbase-cu  
40gbase-sr4-bd

### Use case

10gbase-cu is different from 10gbase-cx4, but cx4 is the only 10g DAC available
There are plenty of other versions of BiDi interfaces, but there's no option for 40g BiDi.

### Database changes

_No response_

### External dependencies

_No response_

bevand10:
Closes #21394

### Fixes: #21394

Added requested types.


jeremystretch (approved):
Thanks @bevand10!

PR #21530, opened 2026-02-26
Fixes #21527: UI Bug with Displaying Primary IP Address with NAT IP o…

### LINKED ISSUE #21527: UI Bug with Displaying Primary IP Address with NAT IP on a Device
### NetBox Edition

NetBox Community

### NetBox Version

v4,5,3

### Python Version

3.12

### Steps to Reproduce

1. Create an IP
2. Create another IP and assign it as the outside NAT IP for the first IP
3. Create a Device
4. Assign the first IP to the device
5. View the device with attention given to the Primary IPv4 NAT address.  The brackets are spread very far apart on either side of the NAT IP

### Expected Behavior

There should be no extra spacing before and after the brackets surrounding the NAT IP. 

### Observed Behavior

There is a large amount of extra space surrounding the NAT IP before and after the brackets. 

It looks like this 

`Primary IPv4  10.10.100.129    (                     NAT: 10.20.30.30                      )`

But should look like this

`Primary IPv4  10.10.100.129    (NAT: 10.20.30.30)`

I can submit a PR with a fix.

robduffy2010:
### Fixes: #21527

UI Bug with Displaying Primary IP Address with NAT IP on a Device

### DIFF
--- a/netbox/templates/dcim/device/attrs/ipaddress.html
+++ b/netbox/templates/dcim/device/attrs/ipaddress.html
@@ -1,10 +1,12 @@
 {% load i18n %}
-<a href="{{ value.get_absolute_url }}"{% if name %} id="attr_{{ name }}"{% endif %}>{{ value.address.ip }}</a>
-{% if value.nat_inside %}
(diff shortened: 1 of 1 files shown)


jeremystretch (approved):
Thanks @robduffy2010!

PR #21546, opened 2026-03-02
Fixes #21533: Fix missing `family`/`mask_length` in API when creating IP-related objects

### LINKED ISSUE #21533: API responses missing read-only fields (e.g., `family`) upon creation
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.3

### Python Version

3.12

### Steps to Reproduce

When creating IP-related objects (`Prefix`, `IPAddress`, `Aggregate`, `IPRange`) via the REST API, read-only fields like `family` are unexpectedly omitted from the returned JSON response.

1. Submit a `POST` request to create a new `Prefix` (e.g., `{"prefix": "192.168.0.0/24"}`).
2. Inspect the HTTP 201 response payload.
3. Note that the `family` attribute is missing from the response, even though it should be a read-only field included in the serialized output.

### Expected Behavior

The API response should contain the `family` attribute (and any other similarly calculated fields) immediately upon object creation.

### Observed Behavior

The API response is missing `family` attribute (and any other similarly calculated fields) immediately upon object creation.

chbndrhnns:
### Fixes: #21533

This PR addresses an issue where read-only calculated fields (like `family` and `mask_length`) are silently omitted from API responses when creating IP-related objects (`Prefix`, `IPAddress`, `Aggregate`, `IPRange`).

### Problem
During DRF's model instantiation and validation lifecycle, IP network fields are temporarily held as strings rather than `netaddr` objects. When the serializer attempts to build the response representation, evaluating properties like `family` raises an `AttributeError` (e.g., `'str' object has no attribute 'version'`). DRF's `Field.get_attribute()` catches this exception for read-only fields and silently ignores it by raising a `SkipField` exception, dropping the key from the API response.

### Solution
- Modified the `family` and `mask_length` properties on IP-related models to detect if the underlying value is a string and cast it to a `netaddr` object before evaluating.
- This ensures these properties can be safely evaluated at any point in the object's lifecycle, including during API validation and serialization.
- Added unit tests in `ipam.tests.test_models` to explicitly verify that the models correctly parse the IP version and mask length when instantiated with raw string inputs.

### DIFF
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -159,9 +159,11 @@ class Aggregate(ContactsMixin, GetAvailablePrefixesMixin, PrimaryModel):
 
     @property
     def family(self):
-        if self.prefix:
-            return self.prefix.version
-        return None
+        if not self.prefix:
+            return None
+        if isinstance(self.prefix, str):
+            return netaddr.IPNetwork(self.prefix).version
+        return self.prefix.version
 
     @property
     def ipv6_full(self):
@@ -335,11 +337,19 @@ class Prefix(ContactsMixin, GetAvailablePrefixesMixin, CachedScopeMixin, Primary
 
     @property
     def family(self):
-        return self.prefix.version if self.prefix else None
+        if not self.prefix:
+            return None
+        if isinstance(self.prefix, str):
+            return netaddr.IPNetwork(self.prefix).version
+        return self.prefix.version
 
     @property
     def mask_length(self):
-        return self.prefix.prefixlen if self.prefix else None
+        if not self.prefix:
+            return None
+        if isinstance(self.prefix, str):
+            return netaddr.IPNetwork(self.prefix).prefixlen
+        return self.prefix.prefixlen
 
     @property
     def ipv6_full(self):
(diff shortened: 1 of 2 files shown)


chbndrhnns:
I don't know how to fix this:

> Attempt 3 failed: Could not fetch an OIDC token. Did you remember to add `id-token: write` to your workflow permissions?

pheus:
> I don't know how to fix this:
>
> Attempt 3 failed: Could not fetch an OIDC token. Did you remember to add `id-token: write` to your workflow permissions?

Thanks for flagging this (and for the PR)! This failure is coming from a new **optional** Claude AI review workflow we recently enabled, and it’s currently misconfigured on our side. Nothing you need to change in your PR! Please feel free to ignore it while we get the workflow fixed. Sorry for the noise!

netbox/ipam/models/ip.py
```diff
@@ -630,7 +640,11 @@ def save(self, *args, **kwargs):
 
     @property
     def family(self):
-        return self.start_address.version if self.start_address else None
+        if not self.start_address:
+            return None
+        if isinstance(self.start_address, str):
+            return netaddr.IPAddress(self.start_address.split('/')[0]).version
```
jnovinger, line 646: I'm not sure the `.split('/')` is needed here. Can you please verify?

chbndrhnns: The test fails without the split.
When self.start_address is a string like "192.168.1.1/24", passing it directly to netaddr.IPAddress() raises:

   ValueError: IPAddress() does not support netmasks or subnet prefixes! See documentation for details.

The .split('/')[0] is necessary to strip the prefix length e.g., /24 before passing the bare IP "192.168.1.1" to netaddr.IPAddress(). Without it, errors out on ip_range.family.

jnovinger (approved):

PR #21560, opened 2026-03-03
Fixes #21357: Add support for registering custom model actions

jnovinger:
### Fixes: #21357

Adds `register_model_actions()` API allowing plugins and core to register custom permission actions that appear as checkboxes in ObjectPermissionForm.

**Changes:**
- Add `model_actions` registry store
- Add `ModelAction` dataclass and `register_model_actions()` function in `utilities.permissions`
- Add `ObjectTypeSplitMultiSelectWidget` with data attributes for JS targeting
- Add `RegisteredActionsWidget` for rendering grouped action checkboxes
- Update `ObjectPermissionForm` with consolidated Actions fieldset
- Add JavaScript to show/hide actions based on selected object types
- Register core actions: `DataSource.sync`, `Device.render_config`, `VirtualMachine.render_config`
- Add unit tests for `ModelAction` and `register_model_actions`

**Plugin usage:**
```python
# In plugin's apps.py ready() method
from utilities.permissions import ModelAction, register_model_actions
from .models import MyModel

register_model_actions(MyModel, [
    ModelAction('custom_action', help_text='Description of the action'),
])
```

image

**Testing instructions:**
1. Navigate to Admin > Permissions > Object Permissions > Add
2. In the "Object types" field, select models that have registered actions:
   - Data Source (Core)
   - Device (DCIM)
   - Virtual Machine (Virtualization)
3. Observe the "Custom actions" section appears with checkboxes grouped by model
4. Remove all models with registered actions - the "Custom actions" field should hide completely
5. Select a mix of models with and without registered actions - only applicable action groups should show


jnovinger:
@claude /review

jnovinger:
@claude /review

jnovinger:
@claude /review again. See comments below about some things I decided to not address.

Addressed the dedup bug, template substring check, shared action pre-selection, duplicate registration guard, and stale comment in earlier commits. Additional fixes since the review:

- `SplitMultiSelectWidget` now uses class attributes for widget classes (`7ef77a0`), so subclasses override `available_widget_class`/`selected_widget_class` instead of duplicating `__init__`.
- `register_model_actions()` now raises `ValueError` for reserved action names (`2c1ee47`).

For the remaining items:

- **JS `setTimeout` / hardcoded DOM ID**: `getElementById('id_object_types_1')` follows the same convention used throughout NetBox's JS (e.g. `moveOptions.ts`). `MutationObserver` isn't used anywhere in the codebase. The 50ms timeout matches the existing pattern in `search.ts`.

- **`ModelAction.__eq__` asymmetry**: Tested — Python's reflected `__eq__` fallback handles both directions correctly.

- **Additional form-level tests**: `test_shared_action_preselection`, `test_duplicate_registration_ignored`, and `test_reserved_action_rejected` cover the trickiest scenarios.

jnovinger:
> ### One Remaining Minor Item
> 
> **UX: action names display as raw identifiers**
> 
> **`netbox/utilities/templates/widgets/registered_actions.html:15`**
> 
> The checkbox label renders `action.name` directly (e.g. `render_config`, `sync`). Compare to NetBox's standard CRUD checkboxes which use proper labels ("View", "Add", etc.).
> 
> ```
> {{ action.name }}   {# Renders as "render_config" #}
> ```
> 
> `ModelAction` already has a `help_text` field for supplementary description, but no `verbose_name` for a display-friendly primary label.

This feels like scope creep to me. Let's tie it off here and revisit with a followup issue if needed.

netbox/utilities/permissions.py
```diff
@@ -1,19 +1,64 @@
...
+@dataclass
+class ModelAction:
+    """
+    Represents a custom permission action for a model.
+
+    Attributes:
+        name: The action identifier (e.g. 'sync', 'render_config')
+        help_text: Optional description displayed in the ObjectPermission form
+    """
+    name: str
```
jeremystretch, line 30: We should probably check that `name` is not an empty string.

netbox/users/constants.py
```diff
@@ -10,6 +10,10 @@
 
 CONSTRAINT_TOKEN_USER = '$user'
 
+# Built-in actions that receive special handling (dedicated checkboxes, model properties)
```
jeremystretch: It might be worth calling out that these come from Django.

netbox/utilities/permissions.py
```diff
@@ -1,19 +1,64 @@
...
     'permission_is_exempt',
     'qs_filter_from_constraints',
+    'register_model_actions',
     'resolve_permission',
     'resolve_permission_type',
 )
 
 
+@dataclass
+class ModelAction:
```
jeremystretch, line 22: I'm curious if you've looked into declaring custom permissions under each model's `Meta` per the [Django convention](https://docs.djangoproject.com/en/6.0/topics/auth/customizing/#custom-permissions). I couldn't say why we don't do this already - there might be a good reason - but we should consider that approach if we haven't already.

netbox/core/apps.py
```diff
@@ -25,12 +25,19 @@ def ready(self):
...
+        from utilities.permissions import ModelAction, register_model_actions
 
         from . import data_backends, events, search  # noqa: F401
+        from .models import DataSource
 
         # Register models
         register_models(*self.get_models())
 
+        # Register custom permission actions
+        register_model_actions(DataSource, [
```
jeremystretch: If we determine that it's feasible to declare permissions on each model directly, we can probably move this logic into `register_models()`.

jeremystretch (changes requested):
The UI is a little scattered IMO: It's not made clear to the user the differences among actions, custom actions, and additional actions.

Image

Instead of dynamically adding/removing custom actions based on the selected objects, what do you think about starting by showing checkboxes for _all_ available actions (both built-in and custom), and merely enabling/disabling them based on the selected objects? This would simplify the UI considerably, and it looks like you have the bulk of the necessary scripting in place already.

We should also extend the help text of the "additional actions" field to clarify that it is needed only when dealing with plugins which don't yet register their actions.

netbox/utilities/permissions.py
jnovinger: Looked into this. Django validates `Meta` attributes against an internal allowlist and raises an error for anything unknown, so we can't add custom attributes there without patching `Options`.

Using Django's `Meta.permissions` would work in theory but creates `auth.Permission` DB objects we'd never use, since `ObjectPermission.actions` is its own flat `ArrayField` completely outside Django's permission framework.

A regular class attribute on the model (not `Meta`) could work, and `register_models()` could pick it up. Happy to refactor in that direction if you'd prefer, or keep the current imperative approach.

netbox/utilities/permissions.py
jnovinger: Done, thanks

jnovinger:
Just saw the merge conflicts, fixing those now.

jnovinger:
This is ready for review now, as soon as CI finishes.

netbox/utilities/permissions.py
jeremystretch: My intent was to use `Meta.permissions`.

> Using Django's Meta.permissions would work in theory but creates auth.Permission DB objects we'd never use, since ObjectPermission.actions is its own flat ArrayField completely outside Django's permission framework.

That's okay; it's already the case with the four built-in actions. The benefit of this approach is that we leverage Django's existing mechanism for declaring new permission actions rather than reinventing the wheel.

jeremystretch (changes requested):
To clarify my earlier suggestion, let's consolidate these into a single list. Users don't know or care that a particular action is "custom" so there's no need to separate them out.

Image

It's also a little confusing to list duplicate actions, e.g. `render_config`. There's no significant different between them: So long as the `render_config` action is present for an ObjectPermission, it doesn't matter which model it came from.

jnovinger:
@claude /review. Focus on things that are concerns. Be sure that all of Jeremy's PR feedback is addressed.

jnovinger:
@claude /review, notes below. As before, don't bother mentioning positives, just mention items of concern.

- `registered_actions` field label renamed from "Custom actions" to "Registered actions" (fixes validation error messages)
- `data-models` attribute values are now sorted to avoid non-deterministic ordering in rendered HTML
- Replaced hardcoded `getElementById('id_object_types_1')` with `document.querySelector('select[data-object-types-selected]')`; added `data-object-types-selected` to `ObjectTypeSelectedOptions.get_context()` to support this
- `setTimeout(updateState, 50)` changed to `setTimeout(updateState, 0)`; not switching to `MutationObserver` since it's not used anywhere in the codebase and `0` matches the existing pattern in `search.ts`
- Detail view consolidated into a single "Actions" panel; CRUD and registered actions now appear in one table, consistent with the form layout
- Plugin docs updated to show `Meta.permissions` as the primary registration approach

jnovinger:
> **Claude finished @jnovinger's task in 3m 53s** —— [View job](https://github.com/netbox-community/netbox/actions/runs/23873880556)
> ### Review
> ...

Another round:

* Admin docs still referenced "Custom actions" and "grouped by model"; updated to reflect the flat "Registered actions" layout
* "Registered actions" label intentionally omitted from the form since the fieldset is already titled "Actions"
* Raw model keys in the detail view (e.g. `dcim.device`) left as-is; not something Jeremy asked for
* `clean()` now collects all validation errors before raising instead of stopping at the first
* This one is wrong, `help_text` is already displayed.	
* Double-click path for moving options is out of scope; the existing `.move-option` listener covers the supported interaction

netbox/core/migrations/0023_datasource_sync_permission.py
jeremystretch, line 1: Can we rename these migrations to something like `foo_x_permission`?

netbox/utilities/permissions.py
```diff
@@ -1,19 +1,84 @@
...
+        actions: A list of ModelAction instances or action name strings
+    """
+    label = f'{model._meta.app_label}.{model._meta.model_name}'
+    for action in actions:
+        if isinstance(action, str):
+            action = ModelAction(name=action)
+        if not action.name:
+            raise ValueError("Action name must not be empty.")
+        if action.name in RESERVED_ACTIONS:
+            raise ValueError(f"'{action.name}' is a reserved action and cannot be registered.")
```
jeremystretch: Can we move these validation checks into ModelAction itself to ensure they're always applied?

```python
class ModelAction:
    def __post_init__(self):
        if not self.name:
            raise ValueError("Action name must not be empty.")
        if self.name in RESERVED_ACTIONS:
            raise ValueError(f"'{action.name}' is a reserved action and cannot be registered.")
```

netbox/netbox/registry.py
```diff
@@ -28,6 +28,7 @@ def __delitem__(self, key):
     'denormalized_fields': collections.defaultdict(list),
     'event_types': dict(),
     'filtersets': dict(),
+    'model_actions': collections.defaultdict(list),
```
jeremystretch: Could make this a set as order shouldn't matter.
```suggestion
    'model_actions': collections.defaultdict(set),
```

netbox/utilities/permissions.py
```diff
@@ -1,19 +1,84 @@
...
+        if action not in registry['model_actions'][label]:
```
jeremystretch: We can skip this check if we change `registry['model_actions']` to `defaultdict(set)`.

netbox/dcim/models/devices.py
```diff
@@ -759,6 +759,9 @@ class Meta:
         )
         verbose_name = _('device')
         verbose_name_plural = _('devices')
+        permissions = [
+            ('render_config', 'Render device configuration'),
```
jeremystretch: Suggest dropping the model name from these to avoid confusion in the UI.
```suggestion
            ('render_config', 'Render configuration'),
```

netbox/virtualization/models/virtualmachines.py
```diff
@@ -276,6 +276,9 @@ class Meta:
         )
         verbose_name = _('virtual machine')
         verbose_name_plural = _('virtual machines')
+        permissions = [
+            ('render_config', 'Render VM configuration'),
```
jeremystretch:
```suggestion
            ('render_config', 'Render configuration'),
```

netbox/users/ui/panels.py
```diff
@@ -45,13 +47,43 @@ class ObjectPermissionPanel(panels.ObjectAttributesPanel):
...
+        crud_actions = [
+            (_('View'), 'view' in obj.actions),
+            (_('Add'), 'add' in obj.actions),
+            (_('Change'), 'change' in obj.actions),
+            (_('Delete'), 'delete' in obj.actions),
+        ]
+
+        enabled_actions = set(obj.actions) - set(RESERVED_ACTIONS)
+
+        # Collect all registered actions from the full registry, deduplicating by name.
```
jeremystretch: This seems like a lot of logic to bake into a UI panel. Suggest moving the logic for resolving registered actions to the ObjectPermission class.

netbox/utilities/forms/widgets/select.py
```diff
@@ -180,3 +183,53 @@ def get_context(self, name, value, attrs):
     def value_from_datadict(self, data, files, name):
         # Return only the choices from the SelectedOptions widget
         return super().value_from_datadict(data, files, name)[1]
+
+
+#
+# ObjectType-specific widgets for ObjectPermissionForm
```
jeremystretch: This seems like it's getting out-of-hand. We should avoid creating custom widgets for specific fields. This splits off from the original SplitMultiSelectWidget class, which is now unused.

jeremystretch (changes requested):
While the UI functionality is pretty slick, I'm afraid there's way too much logic baked into the form and widgets to justify it. There's too much going on in the frontend for me to even review in a reasonable amount of time.

In the interest of moving this forward, I'm going to suggest a different approach: Rip out the fancy UI bits and just implement a regular form with a checkbox for each available permission. (Don't worry about enabling/disabling them dynamically.) IMO that's plenty sufficient to meet the needs of FR #21357. We can always revisit the advanced functionality at a later date, after the upcoming v4.6 release.

netbox/users/models/permissions.py
```diff
@@ -82,5 +84,22 @@ def list_constraints(self):
...
+        """
+        enabled_actions = set(self.actions) - set(RESERVED_ACTIONS)
+
+        action_models = {}
+        for model_key, model_actions in registry['model_actions'].items():
+            for action in model_actions:
+                action_models.setdefault(action.name, []).append(model_key)
+
+        return [
+            (name, name in enabled_actions, ', '.join(sorted(action_models[name])))
```
jeremystretch: Suggest we return the sorted list of models natively, and do string joining in the template if needed.

netbox/templates/users/panels/actions.html
```diff
@@ -0,0 +1,26 @@
...
+      </tr>
+    {% endfor %}
+    {% for action, enabled, models in registered_actions %}
+      <tr>
+        <th scope="row">{{ action }}</th>
+        <td>
+          <div class="d-flex justify-content-between align-items-start">
+            {% checkmark enabled %}
+            {% if models %}
+              <small class="text-muted">{{ models }}</small>
```
jeremystretch: This seems a little out of place: It's not immediately evident to the user what significance the models has. It might be more helpful to include a human-friendly description of each action in this table (maybe including the models list).

jeremystretch:
@claude /review

netbox/users/ui/panels.py
```diff
@@ -45,13 +45,25 @@ class ObjectPermissionPanel(panels.ObjectAttributesPanel):
     enabled = attrs.BooleanAttr('enabled')
 
 
-class ObjectPermissionActionsPanel(panels.ObjectAttributesPanel):
+class ObjectPermissionActionsPanel(panels.ObjectPanel):
```
jeremystretch, line 48: Per Claude's feedback, we should also show non-registered actions here. Maybe an "additional actions" row at the bottom of the panel?

jeremystretch (changes requested):

jeremystretch (approved):

PR #21568, opened 2026-03-03
#20923: Migrate Users object to declarative layouts

arthanson:
This PR continues the migration of object views in the user app to NetBox v4.5’s declarative layouts (#20923).

Replace legacy object view templates with declarative layouts for:

   - Users
   - Groups
   - API Tokens
   - Permissions
   - Owner Groups
   - Owners


arthanson:
@claude /review

netbox/templates/users/token.html
```diff
@@ -1,4 +1,3 @@
 {% extends 'generic/object.html' %}
-{% load i18n %}
 
-{% block title %}{% trans "Token" %} {{ object }}{% endblock %}
```
pheus, line 4: Would you be open to keeping the **“Token …”** prefix in the title?

netbox/templates/users/objectpermission.html
```diff
@@ -1,93 +1,3 @@
 {% extends 'generic/object.html' %}
-{% load i18n %}
-{% load helpers %}
-{% load render_table from django_tables2 %}
-
-{% block title %}{% trans "Permission" %} {{ object.name }}{% endblock %}
```
pheus, line 6: This is a personal preference:
But I like that the header / title reads "Permission app.model" instead of just "app.model".

netbox/templates/users/token.html
```diff
@@ -1,4 +1,3 @@
...
+{% block subtitle %}{% endblock %}
```
pheus:
```suggestion

```
Could we avoid blanking the subtitle for tokens, so the created timestamp remains visible?

netbox/users/views.py
```diff
@@ -86,7 +95,40 @@ class UserListView(generic.ObjectListView):
...
+        bottom_panels=[
+            ContextTablePanel(
+                'changelog_table',
+                title=_('Recent Activity'),
+                actions=[
+                    actions.LinkAction(
+                        view_name='core:objectchange_list',
+                        url_params={'user_id': lambda ctx: ctx['object'].pk},
+                        label=_('View All'),
+                        button_class='ghost-primary',
```
pheus:
```suggestion

```
The custom styling here stands out a bit from the other object views and the standard action styling we use elsewhere.  
Could we drop the custom look and keep this action consistent with the default button styling?

netbox/users/views.py
```diff
@@ -326,7 +410,19 @@ class OwnerListView(generic.ObjectListView):
 @register_model_view(Owner)
 class OwnerView(GetRelatedModelsMixin, generic.ObjectView):
     queryset = Owner.objects.all()
+    # Custom template is required to override the breadcrumbs block and include
+    # the owner's group in the breadcrumb trail when present.
     template_name = 'users/owner.html'
```
pheus:
```suggestion

```
Thanks for calling this out in-code. I agree the comment is factually correct, but we do have several cases where we intentionally keep a template around for customization (breadcrumbs are a good example). Because of that, explicitly setting `template_name` and adding the comment feels a bit unusual compared to similar views.  
Would you mind removing the comment (and the explicit `template_name` assignment), and just relying on the existing template override where needed?

pheus (changes requested):
Thanks for the conversion! Overall this looks really good.
I only have a couple of small nits to address before this is ready.

Also, while reviewing the changes I noticed there are already at least two distinct use cases for rendering date/datetime attributes. That makes me suspect we’ll run into more of these over time.  
Would you be open to introducing a dedicated `DateTimeAttr` (with options to control the exact rendering/format) instead of adding multiple `TemplateAttr` variants for each formatting case?

_One more process note_: for this work we should be referencing the main tracking issue (#20923) rather than creating sub-issues. Could we close the newly created issue as a duplicate, and update this PR to **reference** #20923 (rather than “fixes/closes” it), so the overall migration progress stays centralized in the main tracker?

pheus (approved):
Thanks for adding `DateTimeAttr`!

I also thought about moving the rendering logic into the Python class, but I don’t think importing template filters into `attrs.py` would be the right trade-off either. For now, I’m okay with the small dedicated template.

As a follow-up improvement, it might be worth extracting the ISO date/datetime formatting into a shared Python helper so both the template filters and `DateTimeAttr` can reuse the same logic without crossing layers.

PR #21570, opened 2026-03-04
Limit auto-review workflow to GitHub org members

jeremystretch:
Modifies the `claude-code-review` workflow (currently disabled) so that it applies only to PRs submitted by members of the `netbox-community` organization, as a safeguard against abuse.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -8,8 +8,9 @@ jobs:
   claude-review:
     # Only run for PRs submitted by organization members or owners
     if: |
-      github.event.pull_request.author_association == 'MEMBER' ||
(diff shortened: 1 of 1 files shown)


pheus:
A couple of (optional) thoughts:

- If the goal is “only run when the PR branch is in this repo (not a fork)”, we could also gate on:
  `github.event.pull_request.head.repo.full_name == github.repository`

- If we enable this workflow, it might be worth pinning `anthropics/claude-code-action` to a specific commit SHA (as we often do with third‑party actions), just to reduce supply-chain risk.

Just suggestions for future hardening.

arthanson (dismissed):
I'd also suggest pinning as Martin suggested

arthanson (approved):

PR #21574, opened 2026-03-04
Closes #20698: Expose total_vlan_ids on VLAN groups

### LINKED ISSUE #20698: Add a read-only "Total VLAN IDs" count to the VLANGroup model
### NetBox version

v4.4.4

### Feature type

New functionality

### Proposed functionality

Add a public integer field to the `ipam.VLANGroup` model showing the total number of VLAN IDs contained in the range. This is tracked internally by the private field `_total_vlan_ids`, which can be renamed to be public.

The count should be included in API representations of a VLAN group, and should be available as a table column in the UI.

### Use case

Displaying an absolute count of VLANs would be helpful in scenarios where a range contains a complex set of sub-ranges, e.g. 1-50,100,200,1000-1999.

### Database changes

Presumably, the existing field `_total_vlan_ids` would be renamed to e.g. `total_vlan_ids`.

### External dependencies

N/A

pheus:
### Fixes: #20698

This PR exposes the total number of VLAN IDs contained in a `VLANGroup`.

It renames the internal `_total_vlan_ids` field to `total_vlan_ids` (preserving existing data via a migration), and makes the value available as a read-only integer in the VLANGroup API. The VLANGroup table view has also been updated to include a **Total VLAN IDs** column, with relevant tests adjusted accordingly.

### DIFF
--- a/docs/models/ipam/vlangroup.md
+++ b/docs/models/ipam/vlangroup.md
@@ -18,6 +18,10 @@ A unique URL-friendly identifier. (This value can be used for filtering.)
 
 The set of VLAN IDs which are encompassed by the group. By default, this will be the entire range of valid IEEE 802.1Q VLAN IDs (1 to 4094, inclusive). VLANs created within a group must have a VID that falls within one of these ranges. Ranges may not overlap.
 
+### Total VLAN IDs
+
+A read-only integer indicating the total count of VLAN IDs available within the group, calculated from the configured VLAN ID Ranges. For example, a group with ranges `100-199` and `300-399` would have a total of 200 VLAN IDs. This value is automatically computed and updated whenever the VLAN ID ranges are modified.
+
 ### Scope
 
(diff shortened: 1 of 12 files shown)


jnovinger (changes requested):
Two things I think we need to address before merging:

1. `docs/models/ipam/vlangroup.md` should get a `total_vlan_ids` entry. The other fields are documented there and this one should be too, even though it's read-only/computed.

2. Missing a `test_total_vlan_ids()` in `VLANGroupTestCase` in `test_filtersets.py`. Note that the test setup uses `bulk_create()` which skips `save()`, so you'll need to either set `total_vlan_ids` explicitly on the test objects or use individual `.save()` calls to get varying values for meaningful filter assertions.

pheus:
Thanks for the thoughtful review, and sorry for missing these on the first pass.

I've now added the documentation entry, a filterset test for `total_vlan_ids`, and the corresponding FilterForm field.

Thanks again for the helpful feedback!

jnovinger (approved):

PR #21581, opened 2026-03-04
Closes #20916: Record a stack trace in the job log for unhandled exceptions

### LINKED ISSUE #20916: Include stack trace for unhandled exceptions on 500 responses and in job logs
### NetBox version

v4.4.7

### Feature type

### Proposed functionality

Currently, when dealing with an unhandled exception NetBox displays only the exception message. This FR proposes including the complete stack trace as well.

As the stack trace may be considered mildly sensitive due to the disclosure of file paths, we should sanitize file paths by omitting anything higher than the application's root path.

### Use case

Stack traces are often necessary to expedite the troubleshooting process, and making these available to the user immediately is expected to simplify the bug reporting process for all parties.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #20916

Record a traceback when capturing an unhandled exception in the log of a background job, truncating file paths outside the installation root.

### DIFF
--- a/netbox/core/tables/jobs.py
+++ b/netbox/core/tables/jobs.py
@@ -1,4 +1,6 @@
 import django_tables2 as tables
+from django.utils.html import conditional_escape
+from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy as _
 
 from core.constants import JOB_LOG_ENTRY_LEVELS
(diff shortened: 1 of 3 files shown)


jeremystretch:
@claude /review

bctiemann (approved):

PR #21582, opened 2026-03-04
Fix #17654: Add Role to ASN

### LINKED ISSUE #17654: Enhance ASN model
### NetBox version

v4.1.2

### Feature type

Data model extension

### Proposed functionality

Extend ASN model with new "Role" and "Name" fields

### Use case

BGP ASNs have roles and names, being able to capture that information natively in Netbox instead of using custom fields can help to document networks.

For example many companies I worked with name their ASNs like CompanyXYZ_public_ASN or LAB123_ASN or customer_foo_asn. 

At the same time there can be more then single ASN used by individual device using local-as feature, in that case being able to capture ASN role becomes also helpful, like DEFAULT role - device global ASN for BGP process or VRF_XYZ role - ASN for specific VRF use on the device using local-as

### Database changes

ASN model need to have these new fields added:
role - reference to Prefi&Vlan roles objects
name - string

### External dependencies

None

arthanson:


arthanson:
@claude /review

arthanson:
@claude /review

jnovinger:
@claude /review

Notes:
1. `select_related()` is generally an anti-pattern in NetBox. We prefer the use of `prefetch_related()`. If you review the codebase, you will note that there is machinery to apply `prefetch_related()` opportunistically for django-tables2 based views and for API views.

2. Release notes do not happen in individual issue PRs, they are added in release PRs.

3. Re: the `brief_fields` bit, if this is an existing pattern, there's likely a good reason for it.

As a general note, it's helpful to review the `docs/development/`docs: particularly `adding-models.md`, `extending-models.md`, and `release-checklist.md`.

jnovinger (changes requested):
Feedback:

- View tests in `test_views.py` should probably be updated to include `role`. `ASNTestCase` is missing Role fixtures in `setUpTestData`, and `role` isn't included in `form_data`, `csv_data`, or `bulk_edit_data`.

- Agree with Claude's review, in that `role` should be added to the `SearchIndex`.

jnovinger (approved):

PR #21584, opened 2026-03-04
Closes #21409: Add option to retain create & last update changelog records when pruning

### LINKED ISSUE #21409: Introduce an option to retain the original "create" and latest "update" change record for each object
### NetBox version

v4.5.2

### Feature type

New functionality

### Proposed functionality

Introduce a configuration parameter (name TBD) that enables preserving the original "create" and most recent "update" change record for each object in NetBox. These records will be exempt from the purge of expired changelog records (per the [`CHANGELOG_RETENTION`](https://netboxlabs.com/docs/netbox/configuration/miscellaneous/#changelog_retention) setting).

When this configuration parameter is enabled, the housekeeping task responsible for purging old changelog records will ignore

1. All creation records (for which there can only be one per object)
2. The most recent update record for each object

### Use case

This FR is being submitted as an alternative to the proposal in #15323, which seeks to capture information about the user(s) who created and last modified each object. Rather than replicating this information onto each object, we can leverage the data which already exists in the changelog for this purpose.

### Database changes

N/A

### External dependencies

N/A

pheus:
### Fixes: #21409

Thanks for reviewing this PR!

This introduces a new dynamic configuration parameter, `CHANGELOG_RETAIN_CREATE_LAST_UPDATE`, which ensures each **non-deleted** object’s original `create` change record and most recent `update` change record are retained when pruning expired changelog entries per `CHANGELOG_RETENTION`. For objects with a `delete` change record, changelog entries are pruned normally according to `CHANGELOG_RETENTION`.

Changes included:
- Add `CHANGELOG_RETAIN_CREATE_LAST_UPDATE` (enabled by **default**)
- Update changelog pruning logic (system housekeeping job + `housekeeping` management command) to respect the new behavior, including excluding objects with a `delete` change record from the retention exemption
- Add tests covering pruning behavior, including deleted-object scenarios
- Update docs/templates/forms to surface the behavior in the UI and configuration reference (including an upgrade warning)

#### Why this is enabled by default
I’ve set the default value to `True` because NetBox defaults to a 90-day changelog retention period, which can remove valuable provenance data for new users (especially “who created this?” / “who last changed this?”) once the instance has been running for a while. In my experience it’s preferable to retain this minimal set of high-value records by default, even if the behavior change may be surprising for existing users, rather than silently losing it.

For existing deployments that prefer the previous pruning behavior, this can be reverted by setting
`CHANGELOG_RETAIN_CREATE_LAST_UPDATE = False`. Any previously retained records will then be eligible for pruning on the next housekeeping run.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -21,6 +21,7 @@ Some configuration parameters are primarily controlled via NetBox's admin interf
 * [`BANNER_BOTTOM`](./miscellaneous.md#banner_bottom)
 * [`BANNER_LOGIN`](./miscellaneous.md#banner_login)
 * [`BANNER_TOP`](./miscellaneous.md#banner_top)
+* [`CHANGELOG_RETAIN_CREATE_LAST_UPDATE`](./miscellaneous.md#changelog_retain_create_last_update)
 * [`CHANGELOG_RETENTION`](./miscellaneous.md#changelog_retention)
 * [`CUSTOM_VALIDATORS`](./data-validation.md#custom_validators)
 * [`DEFAULT_USER_PREFERENCES`](./default-values.md#default_user_preferences)
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -73,6 +73,27 @@ This data enables the project maintainers to estimate how many NetBox deployment
 
 ---
 
+## CHANGELOG_RETAIN_CREATE_LAST_UPDATE
+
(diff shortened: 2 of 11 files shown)


pheus:
@claude /review

pheus:
@claude /review

pheus:
@claude /review

pheus:
@claude /review

bctiemann (approved):

PR #21588, opened 2026-03-05
Fixes #19867: Retain the `per_page` URL parameter after editing an object

### LINKED ISSUE #19867: Elements per page setting is discarded after editing an object
### Deployment Type

Self-hosted

### NetBox Version

v4.3.3

### Python Version

3.12

### Steps to Reproduce

1. Go into the "Interfaces" overview. This assumes that you have more than 25 interfaces. 
2. In the lower right corner, press the drop down "Per Page" and select that you wish to see 100 entries. 
3. Select two elements in this longer list, press "Edit Selected"
4. Do the necessary changes and save. 
5. You're now being taken back to the "Interfaces" overview. You're back in the default view of 25 entries instead of 100. 

### Expected Behavior

I would expect that the setting remains active until I decide to leave the overview by explicitly navigating somewhere else. 

### Observed Behavior

As mentioned in "Steps to reproduce"

jeremystretch:
### Fixes: #19867

Update `return_url` to capture any additional query parameters.

### DIFF
--- a/netbox/templates/generic/object_list.html
+++ b/netbox/templates/generic/object_list.html
@@ -92,7 +92,7 @@ Context:
 
         <div class="form form-horizontal">
           {% csrf_token %}
-          <input type="hidden" name="return_url" value="{% if return_url %}{{ return_url }}{% else %}{{ request.path }}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}{% endif %}" />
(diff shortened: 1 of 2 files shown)


jeremystretch:
@claude /review

bctiemann (approved):

PR #21590, opened 2026-03-05
Fixes #20915: Ensure preferred language is applied during SSO login

### LINKED ISSUE #20915: Language not re-applied on login with social-app-django
### NetBox Edition

NetBox Community

### NetBox Version

v4.2.9

### Python Version

3.10

### Steps to Reproduce

1. Configure django-social-app with any provider
2. Set your language under `Preferences > Language` to any language
3. Log out
4. Log in again

### Expected Behavior

The language setting should be re-applied upon login

### Observed Behavior

The language setting is not re-applied upon login.

jeremystretch:
### Fixes: #20915

- Introduce a signal handler (`set_language_on_login()`) to attach the user's preferred language (if any) to the request during login
- Extend the middleware logic to apply the language if set on the request

### DIFF
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -40,15 +40,24 @@ class CoreMiddleware:
         with apply_request_processors(request):
             response = self.get_response(request)
 
-        # Check if language cookie should be renewed
-        if request.user.is_authenticated and settings.SESSION_SAVE_EVERY_REQUEST:
-            if language := request.user.config.get('locale.language'):
-                response.set_cookie(
-                    key=settings.LANGUAGE_COOKIE_NAME,
-                    value=language,
-                    max_age=request.session.get_expiry_age(),
-                    secure=settings.SESSION_COOKIE_SECURE,
-                )
+        # Set or renew the language cookie based on the user's preference. This handles two cases:
+        # 1. The user just logged in (via any auth backend): the user_logged_in signal stores the preferred language on
+        #    the request so we set the cookie here on the login response.
+        # 2. SESSION_SAVE_EVERY_REQUEST is enabled: renew the language cookie on every request to keep it in sync with
+        #    the session expiry.
+        if hasattr(request, '_language_cookie'):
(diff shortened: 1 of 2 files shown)


jeremystretch:
@claude /review

jnovinger (approved):

jnovinger:
Looks good, although I did notice, while looking at this, that `user_last_login` gets disconnected during maintenance mode but never reconnected (that I can tell). Probably not a huge issue and definitely out of scope for this issue, but maybe something to follow up on.

Re: Claude's review above, this is what my local Claude had to say:

> Here's my take on each of the three "minor observations":
>
> **1. "No tests added"** — The suggestion to test that `auth.login()` sets `request._language_cookie` is fine in theory but low-value. The signal handler is 3 lines of trivially-correct code, and such a test would mostly verify that Django's signal dispatch works. The Claude bot hedges correctly ("not a blocker") but the framing "given the project's convention of requiring tests for new functionality" is a stretch — NetBox doesn't have signal handler unit tests elsewhere. INFERRED: `rg "user_logged_in\|user_login_failed" netbox/*/tests/` would likely return nothing.
>
> **2. "Double-set on normal form login"** — This observation is correct and I flagged the same thing. But calling it "actually better behavior" is wrong. The previous code in `LoginView.post()` already set the cookie on every form login — that's not new. The "double-set" is just redundant (middleware and view both set the same cookie on the same response). For non-SSO users with `SESSION_SAVE_EVERY_REQUEST=False`, there's no behavior change at all — the LoginView was already handling it.
>
> **3. "`set_cookie` parameters"** — This is noise. The bot correctly says "unchanged from the pre-existing code" and "not a new issue" but then suggests a follow-up anyway. It's padding the review with a pre-existing concern unrelated to the PR. A reviewer should either file a separate issue or not mention it.
>
> Overall: the bot's review is directionally correct and doesn't contain anything wrong that would mislead, but it inflates minor points to seem more substantive. The "double-set" analysis in particular mischaracterizes what actually changed.

PR #21595, opened 2026-03-05
#21330 optimize the assignment of tags when saving an object

### LINKED ISSUE #21330: Optimize the assignment of tags when saving an object
### NetBox Version

v4.5.1

### Python Version

3.12

### Area(s) of Concern

- [ ] User Interface
- [ ] REST API
- [ ] GraphQL API
- [x] Python ORM
- [ ] Other

### Details

Assigning tags to a newly created object is currently done in an inefficient manner. For example, creating a new site with five tags assigned generates ~23 additional SQL queries (4 per tag):

Image

AFAICT these are attributable to the default behavior of the django-taggit manager, but we should explore potential options for optimizing the creation of tag assignments in the database.

arthanson:
### Fixes: #21330 

Taggit was causing N+1 queries here. Created a new NetBoxTaggableManager replacing the add function on taggit and generally follows what it does but does a single bulk-insert of all tags instead of one at a time.  Also a small optimization on the serializer.  Could probably be upstreamed into Taggit, will look at opening an optimization PR there.

Created a new site with 12 tag assignments - checking the POST Request:

main branch: 76 SQL queries
this branch: 29 SQL queries

**Note:** After adding site, with debug toolbar, go to the history menu, find the post request and choose 'switch' to that then look at SQL queries.

### DIFF
--- a/netbox/extras/managers.py
+++ b/netbox/extras/managers.py
@@ -1,0 +1,65 @@
+from django.db.models import signals
+from django.db import router
+from taggit.managers import _TaggableManager
+
+__all__ = (
+    'NetBoxTaggableManager',
+)
+
+
+class NetBoxTaggableManager(_TaggableManager):
+    """
+    Extends taggit's _TaggableManager to replace the per-tag get_or_create loop in add() with a
+    single bulk_create() call, reducing SQL queries from O(N) to O(1) when assigning tags.
+    """
+
+    def add(self, *tags, through_defaults=None, tag_kwargs=None, **kwargs):
+        self._remove_prefetched_objects()
+        if tag_kwargs is None:
+            tag_kwargs = {}
+
+        tag_objs = self._to_tag_model_instances(tags, tag_kwargs)
+        new_ids = {t.pk for t in tag_objs}
+
+        # Determine which tags are not already assigned to this object
+        vals = set(
+            self.through._default_manager.using(db)
+            .values_list("tag_id", flat=True)
+            .filter(**self._lookup_kwargs())
+        )
+        new_ids -= vals
+
+        if not new_ids:
+            return
+
+        db = router.db_for_write(self.through, instance=self.instance)
+        signals.m2m_changed.send(
+            sender=self.through,
+            action="pre_add",
+            instance=self.instance,
+            reverse=False,
+            model=self.through.tag_model(),
+            pk_set=new_ids,
+            using=db,
+        )
+
+        # Use a single bulk INSERT instead of one get_or_create per tag.
+        lookup = self._lookup_kwargs()
+        self.through._default_manager.using(db).bulk_create(
+            [
+                self.through(tag=tag, **lookup, **(through_defaults or {}))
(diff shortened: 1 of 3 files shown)


arthanson:
@claude /review

netbox/extras/managers.py
```diff
@@ -0,0 +1,65 @@
...
+)
+
+
+class NetBoxTaggableManager(_TaggableManager):
+    """
+    Extends taggit's _TaggableManager to replace the per-tag get_or_create loop in add() with a
+    single bulk_create() call, reducing SQL queries from O(N) to O(1) when assigning tags.
+    """
+
+    def add(self, *tags, through_defaults=None, tag_kwargs=None, **kwargs):
```
jeremystretch, line 18: We probably want to wrap this with `require_instance_manager()` as _TaggableManager does.

```suggestion
    @require_instance_manager
    def add(self, *tags, through_defaults=None, tag_kwargs=None, **kwargs):
```

netbox/extras/managers.py
```diff
@@ -0,0 +1,65 @@
...
+        self._remove_prefetched_objects()
+        if tag_kwargs is None:
+            tag_kwargs = {}
+
+        tag_objs = self._to_tag_model_instances(tags, tag_kwargs)
+        new_ids = {t.pk for t in tag_objs}
+
+        # Determine which tags are not already assigned to this object
+        db = router.db_for_write(self.through, instance=self.instance)
```
jeremystretch: Very minor nit, but could we move this up to line 20, just to minimize the diff with `_TaggableManager.add()`?

netbox/extras/managers.py
```diff
@@ -0,0 +1,65 @@
...
+            action="pre_add",
+            instance=self.instance,
+            reverse=False,
+            model=self.through.tag_model(),
+            pk_set=new_ids,
+            using=db,
+        )
+
+        # Use a single bulk INSERT instead of one get_or_create per tag.
+        lookup = self._lookup_kwargs()
```
jeremystretch: `self._lookup_kwargs()` is called once above (line 29). Can we capture and reuse its value here?

jeremystretch (changes requested):

netbox/extras/managers.py
```diff
@@ -0,0 +1,67 @@
...
+
+        tag_objs = self._to_tag_model_instances(tags, tag_kwargs)
+        new_ids = {t.pk for t in tag_objs}
+
+        # Determine which tags are not already assigned to this object
+        lookup = self._lookup_kwargs()
+        vals = set(
+            self.through._default_manager.using(db)
+            .values_list("tag_id", flat=True)
+            .filter(**(lookup), tag_id__in=new_ids)
```
jeremystretch: Redundant parentheses
```suggestion
            .filter(**lookup, tag_id__in=new_ids)
```

jeremystretch (dismissed):
Approved with a nit

jeremystretch (approved):

Issue #20123, opened 2025-08-18
Expose modules adopt_components and replicate_components via the API


TheFlyingCorpse:
### NetBox version

v4.2.6

### Feature type

Change to existing functionality

### Proposed functionality

Expose «adopt_components» and «replicate_components» via the API, existing functionality already present in the Web GUI.

I found an old issue #10925 which ended up closed before merged which could have resolved this.

### Use case

Helping to adapt the full Netbox model for environments with existing devices and cables.

For those with large environments that don’t want to reconnect thousands of cables, please expand so the API can expose these parameters, or allow us to submit a PR for it.

### Database changes

No schema changes required.

### External dependencies

None.

CADbloke:
This would certainly ease the adoption of the move to model SPFs as Modules with an Interface instead of just as raw interfaces as per https://netboxlabs.com/blog/sfp-modeling-modules-over-inventory-items/ 
I presume we're not the only ones trying to figure out a workflow towwards this. 
... or am I misunderstanding and the default is to adopt components?

PR #21600, opened 2026-03-06
Fixes: #20123 - Add replicate_components and adopt_components write_only fields to ModuleSerializer

bctiemann:
### Fixes: #20123

Adds `replicate_components` and `adopt_components` fields to `ModuleSerializer`. These are write-only fields, meaning they do not appear in API GET payloads. However, they are used during create operations to create (new) or adopt (existing) components automatically, which is functionality already present in the UI.

Note that the logic is somewhat redundant with `ModuleCommonForm.clean` and `Module.save`, but the consolidation of similar pieces of code might be too intricate to be worthwhile as only small parts are truly identical.

### DIFF
--- a/netbox/dcim/api/serializers_/devices.py
+++ b/netbox/dcim/api/serializers_/devices.py
@@ -175,6 +175,11 @@ class ModuleSerializer(PrimaryModelSerializer):
         brief_fields = ('id', 'url', 'display', 'device', 'module_bay', 'module_type', 'description')
 
     def validate(self, data):
+        # When used as a nested serializer (e.g. as the `module` field on device component
+        # serializers), `data` is already a resolved Module instance — skip our custom logic.
+        if self.nested:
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review

bctiemann:
@claude /review

bctiemann:
@claude /review

bctiemann:
@claude /review

bctiemann:
@arthanson This is ready for review.

arthanson (approved):

krisiasty:
This patch doesn't seem to be included in any release (yet), and the API still ignores `adopt_components` field. This is breaking any automation I'm trying to build around it to add transceiver modules to existing switches.

I have two questions:
1. will this patch actually solve the problem of API ignoring adopt_components
2. if so, when it will be included in official release?

PR #21601, opened 2026-03-06
Closes #21575: Implement `{vc_position}` template variable on component template name/label

Etibru:
### Closes #21575 

* Added the possibility of having a `{vc_position}` token as `{module}` for component template name/label **at the time of component creation**.
* Possibility of having a fallback value if the device is not in a virtual chassis case.

### DIFF
--- a/docs/models/dcim/devicetype.md
+++ b/docs/models/dcim/devicetype.md
@@ -7,6 +7,18 @@ Device types are instantiated as devices installed within sites and/or equipment
 !!! note
     This parent/child relationship is **not** suitable for modeling chassis-based devices, wherein child members share a common control plane. Instead, line cards and similarly non-autonomous hardware should be modeled as modules or inventory items within a device.
 
+## Automatic Component Renaming
+
+When adding component templates to a module type, the string `{vc_position}` can be used in component template names to reference the
+`vc_position` field of the device being provisioned, when that device is a member of a Virtual Chassis.
+
+For example, an interface template named `Gi{vc_position}/0/0` installed on a Virtual Chassis
+member with position `2` will be rendered as `Gi2/0/0`.
+
+If the device is not a member of a Virtual Chassis, `{vc_position}` defaults to `0`. A custom
+fallback value can be specified using the syntax `{vc_position:X}`, where `X` is the desired default.
+For example, `{vc_position:1}` will render as `1` when no Virtual Chassis position is set.
+
 ## Fields
 
 ### Manufacturer
--- a/docs/models/dcim/devicetype.md
+++ b/docs/models/dcim/devicetype.md
@@ -9,7 +9,7 @@ Device types are instantiated as devices installed within sites and/or equipment
 
 ## Automatic Component Renaming
 
-When adding component templates to a module type, the string `{vc_position}` can be used in component template names to reference the
+When adding component templates to a device type, the string `{vc_position}` can be used in component template names to reference the
 `vc_position` field of the device being provisioned, when that device is a member of a Virtual Chassis.
 
 For example, an interface template named `Gi{vc_position}/0/0` installed on a Virtual Chassis
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -1039,7 +1039,7 @@ class ModularComponentTemplateForm(ComponentTemplateForm):
             "supported (example: <code>[ge,xe]-0/0/[0-9]</code>). The token <code>{module}</code>, if present, will be "
             "automatically replaced with the position value when creating a new module. "
             "The token <code>{vc_position}</code> will be replaced with the device's Virtual Chassis position "
-            "(use <code>{vc_position:1}</code> to specify a fallback (default is 0)"
+            "(use <code>{vc_position:1}</code> to specify a fallback (default is 0))"
         )
 
 
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -1037,7 +1037,9 @@ class ModularComponentTemplateForm(ComponentTemplateForm):
         self.fields['name'].help_text = _(
             "Alphanumeric ranges are supported for bulk creation. Mixed cases and types within a single range are not "
             "supported (example: <code>[ge,xe]-0/0/[0-9]</code>). The token <code>{module}</code>, if present, will be "
-            "automatically replaced with the position value when creating a new module."
+            "automatically replaced with the position value when creating a new module. "
+            "The token <code>{vc_position}</code> will be replaced with the device's Virtual Chassis position "
+            "(use <code>{vc_position:1}</code> to specify a fallback (default is 0)"
         )
 
 
--- a/netbox/dcim/models/device_component_templates.py
+++ b/netbox/dcim/models/device_component_templates.py
@@ -221,8 +221,6 @@ class ModularComponentTemplateModel(ComponentTemplateModel):
         has_vc = VC_POSITION_RE.search(self.label) is not None
         if not has_module and not has_vc:
             return self.label
-        if not has_module and not has_vc:
-            return self.label
 
         label = self.label
 
@@ -230,9 +228,9 @@ class ModularComponentTemplateModel(ComponentTemplateModel):
             modules = self._get_module_tree(module)
             for m in modules:
                 label = label.replace(MODULE_TOKEN, m.module_bay.position, 1)
-
-        resolved_device = (module.device if module else None) or device
-        label = self._resolve_vc_position(label, resolved_device)
+        if has_vc:
+            resolved_device = (module.device if module else None) or device
+            label = self._resolve_vc_position(label, resolved_device)
 
         return label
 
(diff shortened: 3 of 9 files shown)


pheus:
Thanks for working on this.

I’m not reviewing the actual code changes here, but I did have one drive-by thought about the overall approach.

From a quick look, this seems to cover `{vc_position}` when components are instantiated from templates, for example when a module is inserted or when a device is created with a VC position already set.

What I’m not sure about is the case where a `DeviceType` already has interfaces defined, a device has already been instantiated from it, and that device is only later assigned to a `VirtualChassis`.

Likewise, removing a device that is already assigned to a `VirtualChassis` seems to raise the same concern in reverse.

In both cases, the interfaces already exist by the time the VC membership changes, so it feels like the `VirtualChassis` use case may need a different approach from the `{module}`-style solution.

I may be missing something, but it seems worth confirming whether those add/remove cases are intentionally out of scope for this PR.

netbox/dcim/models/device_component_templates.py
```diff
@@ -177,29 +197,44 @@ def _get_module_tree(self, module):
...
+            name = self._resolve_vc_position(name, resolved_device)
+
+        return name
+
+    def resolve_label(self, module=None, device=None):
+        has_module = MODULE_TOKEN in self.label
+        has_vc = VC_POSITION_RE.search(self.label) is not None
+        if not has_module and not has_vc:
+            return self.label
+        if not has_module and not has_vc:
```
arthanson: duplicated if statement, this second one will never get hit as the first one with this condition will return.

netbox/dcim/models/device_component_templates.py
```diff
@@ -177,29 +197,44 @@ def _get_module_tree(self, module):
...
-            label = self.label
-            for module in modules:
-                label = label.replace(MODULE_TOKEN, module.module_bay.position, 1)
-            return label
-        return self.label
+            for m in modules:
+                label = label.replace(MODULE_TOKEN, m.module_bay.position, 1)
+
+        resolved_device = (module.device if module else None) or device
+        label = self._resolve_vc_position(label, resolved_device)
```
arthanson: should this be called if there is no has_vc? I think you want same handling for has_vc as in resolve_name?

netbox/dcim/models/devices.py
```diff
@@ -962,6 +963,16 @@ def _instantiate_components(self, queryset, bulk_create=True):
             components = [obj.instantiate(device=self) for obj in queryset]
             if not components:
                 return
+
+            # Check for duplicate names after resolution {vc_position}
+            names = [c.name for c in components]
+            duplicates = {n for n in names if names.count(n) > 1}
+            if duplicates:
+                raise AbortRequest(
+                    _("Interface name conflict after resolving {{vc_position}}: {names}").format(
```
arthanson: This should probably say `Component name` I think it can be called for all components, not just interfaces?

netbox/dcim/forms/model_forms.py
```diff
@@ -1037,7 +1037,9 @@ def __init__(self, *args, **kwargs):
         self.fields['name'].help_text = _(
             "Alphanumeric ranges are supported for bulk creation. Mixed cases and types within a single range are not "
             "supported (example: <code>[ge,xe]-0/0/[0-9]</code>). The token <code>{module}</code>, if present, will be "
-            "automatically replaced with the position value when creating a new module."
+            "automatically replaced with the position value when creating a new module. "
+            "The token <code>{vc_position}</code> will be replaced with the device's Virtual Chassis position "
+            "(use <code>{vc_position:1}</code> to specify a fallback (default is 0)"
```
arthanson: minor nit - missing closing paren: "...(default is 0))"

arthanson (changes requested):
See @pheus comments above.  

In device_components_templates.py line 431:
```
    def instantiate(self, **kwargs):
        if self.power_port:
            power_port_name = self.power_port.resolve_name(kwargs.get('module'))
            power_port = PowerPort.objects.get(name=power_port_name, **kwargs)
```
Does the power_port_name line need to pass device as well?
```
power_port_name = self.power_port.resolve_name(kwargs.get('module'), kwargs.get('device'))
```

netbox/dcim/models/device_component_templates.py
Etibru: You're right, it must be like `resolve_name`, it's a mistake.

Etibru:
Thanks for the revew @arthanson !
And thanks for the feedback @pheus :)

You're right, this PR is essentially focused on template instantiation time. When a device is created with a `vc_position` already set, or when a module is installed into a device that is already a VC member, the `{vc_position}` token will be resolved correctly.
However, the cases you describe, (assigning a device to a VC after creation, or removing it), I think this is out of scope for this PR.
In those cases, existing components remain unchanged and would require manual renaming. This is consistent with how `{module}` works: it resolves at instantiation time and does not retroactively rename components.

In practice, my personal workflow is to always assign a device to a VC at creation time, and to remove the entire device rather than just detaching it from the VC (except for the VC master, I affect it later). So these edge cases don't affect my use case.

That said, you raise a valid point about renaming components when a VC position changes. It would indeed be interesting to handle that automatically, but it feels like a separate concern for a follow-up issue. The main blocker I see is that it would require comparing existing interface names against the device type templates to identify which ones were generated from `{vc_position}`, and that breaks down as soon as an interface has been manually renamed (e.g. `ge-1/0/8` -> `xe-1/0/8`), making it impossible to reliably detect which names are still template-derived.
So I'd suggest keeping that as a known limitation for now and tracking it separately.

pheus:
Thanks for taking the time to reflect on this, and for the detailed explanation.

Your reasoning makes sense to me, and I agree that automatic renaming on VC membership changes is probably a separate concern from this PR.

Would you mind adding this limitation to the FR issue and noting that a follow-up issue will likely be needed?

That said, I’d like to highlight that, from my experience, adding or removing a device from a `VirtualChassis` is not really an edge case. I think this differs a bit from the `{module}` case, because a module cannot exist without already being placed into a device, whereas devices can be assigned to or removed from a `VirtualChassis` more freely over time.

I just wanted to help clarify the scope and make sure this limitation is documented clearly.

Etibru:
You're right to highlight this. I've updated the FR issue to document the scope limitation and note that automatic renaming on VC membership changes will need a follow-up issue.

Thank you for your comments ! :)

arthanson:
@Etibru We've decided this would be good to go into NetBox 4.6 along side #21436 the {module_path} changes.  Would you mind switching this to feature branch and rebasing?  Hopefully fairly straightforward.

Etibru:
That's all good, thanks @arthanson !

netbox/dcim/models/devices.py
```diff
@@ -962,6 +963,16 @@ def _instantiate_components(self, queryset, bulk_create=True):
             components = [obj.instantiate(device=self) for obj in queryset]
             if not components:
                 return
+
+            # Check for duplicate names after resolution {vc_position}
```
arthanson, line 981: This duplicate name check only runs If in bulk create, I think it should run in both, maybe refactor that out into a private function.

netbox/dcim/models/device_component_templates.py
```diff
@@ -110,6 +110,26 @@ def clean(self):
                 "device_type": _("Component templates cannot be moved to a different device type.")
             })
 
+    @staticmethod
+    def _resolve_vc_position(value: str, device) -> str:
```
arthanson: I think this can be moved to ModularComponentTemplateModel, InventoryItemTemplate inherit from ComponentTemplateModel derive from ComponentTemplateModel and don't ever use this.

docs/models/dcim/devicetype.md
```diff
@@ -7,6 +7,18 @@ Device types are instantiated as devices installed within sites and/or equipment
 !!! note
     This parent/child relationship is **not** suitable for modeling chassis-based devices, wherein child members share a common control plane. Instead, line cards and similarly non-autonomous hardware should be modeled as modules or inventory items within a device.
 
+## Automatic Component Renaming
+
+When adding component templates to a module type, the string `{vc_position}` can be used in component template names to reference the
```
arthanson: should be device type instead of module type here.

netbox/dcim/models/device_component_templates.py
```diff
@@ -110,6 +110,26 @@ def clean(self):
...
+            explicit_fallback = match.group(1)
+            if (
+                device is not None
+                and device.virtual_chassis is not None
+                and device.vc_position is not None
+            ):
+                return str(device.vc_position)
+            return explicit_fallback if explicit_fallback is not None else '0'
+
+        return VC_POSITION_RE.sub(replacer, value)
```
arthanson: Note - this will replace all tokens so you can have multiple vc_position tokens in the name, wanted to make sure this was intended behavior.

arthanson (changes requested):
Great work, just a few small changes and I think it is ready.

netbox/dcim/models/device_component_templates.py
Etibru: It's not necessarily intentional. Leaving it like this leaves open the possibility of names like this: `switch-{vc_position}-port-{vc_position}-8`
I don't know if that might be useful

arthanson (approved):
Thanks @Etibru ! nice work

PR #21613, opened 2026-03-07
docs: custom rack/device/module airflow FIELD_CHOICES

### LINKED ISSUE #21509: Custom rack/device/module airflow choices not documented
### Change Type

Addition

### Area

Data model

### Proposed Changes

With #18785 support was added for custom airflow choices but I do not find where I can configure them. #19054 also does not tell me. I played around with the latest version (v4.5.3), grepped over the docs for "airflow" but I cannot figure this out. Could you add a sentence to the docs how this is supposed to work?

ypid:
### Fixes: #21509

Was forgotten in: #19054

I tested those customization, all three work as expected.

### DIFF
--- a/docs/configuration/data-validation.md
+++ b/docs/configuration/data-validation.md
@@ -62,10 +62,13 @@ FIELD_CHOICES = {
 The following model fields support configurable choices:
 
 * `circuits.Circuit.status`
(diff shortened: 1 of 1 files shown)


jeremystretch:

PR #21617, opened 2026-03-07
Fixes #20385: Ensure GraphQL API respects `MAX_PAGE_SIZE`

jeremystretch:
### Fixes: #20385

- Extend `apply_pagination()` to check for and apply `MAX_PAGE_SIZE`
- Add a test


jeremystretch:
@claude /review

pheus (commented):
Thanks for working on this! The overall approach looks good to me, but I think there’s still one small gap worth mentioning:
`limit < 0` can still bypass the new cap.

```graphql
{
  site_list(pagination: {limit: -1}) {
    id
    name
  }
}
```

This feels more like an edge case than something users are likely to hit in practice, but I still wanted to point it out for completeness. I’d be okay with keeping it as is if you don’t see this as a practical concern.

pheus (approved):
Thanks!

PR #21623, opened 2026-03-09
#20923: Convert `vpn` views to new UI layout

jeremystretch:
As part of #20923, this PR migrates all views under the `vpn` app from bespoke HTML templates to the new declarative UI layout.


jeremystretch:
@claude /review check for any inconsistencies between the new layout and the original HTML templates.

jeremystretch:
Addressing the inconsistencies flagged above:

1. Not an issue (handled natively by the panel)
2. Known and accepted
3. Deleted `attrs/proposals.html`
4. Not an issue (handled natively by the panel)
5. Incorrect; `tunnelgroup.html` is still in use and "add" button still appears

bctiemann (approved):

PR #21625, opened 2026-03-09
Fixes #21618: Preserve cable terminations when bulk-editing cable profile

GeertJohan:
## Summary
- Cache termination objects in `update_terminations()` before deleting `CableTermination` records when `force=True`
- Without this fix, the `a_terminations`/`b_terminations` properties query the DB after deletion and return empty lists, causing all terminations to be silently lost
- Removes a leftover debug `print()` statement from PR #21069

## Root Cause
PR #21069 added `update_terminations(force=True)` to handle profile changes. This method first deletes all existing `CableTermination` records, then reads `self.a_terminations`/`self.b_terminations` to recreate them. When the single-object edit form is used, these properties are populated from form data (`_a_terminations`/`_b_terminations` private attrs). But during bulk edit (or any `save()` without explicitly setting terminations), the properties fall back to `self.terminations.all()` — which returns empty results because the records were just deleted.

The fix caches the termination objects from `get_terminations()` into `_a_terminations`/`_b_terminations` before the delete step, so they're available for recreation.

## Test plan
- [x] Added `test_cable_profile_change_preserves_terminations` to `CableTestCase`
- [x] Verified test **fails** without the fix (`AssertionError: 0 != 2`)
- [x] Verified test **passes** with the fix
- [x] All existing `CableTestCase` tests pass
- [x] Cable path tests (`test_cablepaths2`) unaffected

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -293,7 +293,6 @@ class Cable(PrimaryModel):
             self._pk = self.pk
 
         if self._orig_profile != self.profile:
-            print(f'profile changed from {self._orig_profile} to {self.profile}')
             self.update_terminations(force=True)
         elif self._terminations_modified:
             self.update_terminations()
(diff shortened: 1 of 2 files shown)


jeremystretch:
Thank you for your interest in contributing to NetBox, however it appears there is no accepted issue that correlates to this pull request.  Before pull requests are opened, we require an accepted issue as per our [contributing guide](https://github.com/digitalocean/netbox/blob/main/CONTRIBUTING.md).  Please wait for the issue to be accepted before submitting a pull request.

If the issue is assigned for work, this PR can be re-opened.

jeremystretch:
@claude /review

Edit: Still broken it seems (not related to PR changes)

jeremystretch:
@claude /review

jeremystretch:
@GeertJohan please be sure to retain the `### Fixes: ` header in future PR descriptions, as GitHub relies on this to automatically close issues upon merge.

jeremystretch (approved):
Thanks @GeertJohan!

GeertJohan:
Thanks for reviewing and merging :)

PR #21628, opened 2026-03-10
Fixes #21579: Display 'add script' button only if user has sufficient permission

### LINKED ISSUE #21579: Scripts: "‘Add" button is displayed even without required permissions
### NetBox Edition

NetBox Community

### NetBox Version

4.5.4

### Python Version

3.12

### Steps to Reproduce

1. Create a script
2. Give user permission to view script

### Expected Behavior

"Add" button should not be displayed

### Observed Behavior

"Add" button is displayed

jeremystretch:

### DIFF
--- a/netbox/templates/extras/script_list.html
+++ b/netbox/templates/extras/script_list.html
@@ -15,7 +15,7 @@
 {% endblock tabs %}
 
 {% block controls %}
-  {% if perms.extras.add_scriptmodule %}
+  {% if perms.core.add_managedfile and perms.extras.add_scriptmodule %}
     {% add_button model %}
   {% endif %}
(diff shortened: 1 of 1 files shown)


pheus (commented):
Thanks for the fast fix on this.

One thing I think this may still miss: should the button also check `perms.core.add_managedfile`? Since script creation depends on ManagedFile as well, checking only `perms.extras.add_scriptmodule` could still expose an **Add** button that the user can’t actually use.

I think requiring both permissions here would keep the UI consistent with the real access check.

pheus (changes requested):

pheus (approved):

PR #21630, opened 2026-03-10
#21114 Allow specifying exclude directories for Data Sources

### LINKED ISSUE #21114: Allow specifying exclude directories, or possibly include/exclude directories for Data Source
### NetBox version

v4.4.10

### Feature type

Data model extension

### Proposed functionality

Add the ability to sync only selected subdirectories from a repo instead of pulling the entire repo.  Ignore Rules operate per file which means you can't exclude directories.

### Use case

Many repos contain a large amount of development content, while the needed production files live in a single subdirectory. Excluding everything else is impractical due to the number of directories involved. An “include-only these paths” option would make it easy to:

-Sync only the required production folder
-Optionally include certain dev folders when needed
-Avoid syncing unrelated files

### Database changes

N/A

### External dependencies

N/A

arthanson:
### Fixes: #21114 

Data sources support ignore rules to filter files during synchronization, but previously these rules could only match against bare filenames — making it impossible to exclude entire directories.

This change passes the full relative path to the pattern matcher, enabling path-based glob patterns like subdir/* or Wireguard-tunnels/*. Existing filename-only rules (e.g. *.txt).

Previously, _walk() called _ignore(file_name) with only the bare filename, so subdirectory patterns never matched.

### DIFF
--- a/docs/models/core/datasource.md
+++ b/docs/models/core/datasource.md
@@ -36,13 +36,16 @@ If false, synchronization will be disabled.
 
 ### Ignore Rules
 
-A set of rules (one per line) identifying filenames to ignore during synchronization. Some examples are provided below. See Python's [`fnmatch()` documentation](https://docs.python.org/3/library/fnmatch.html) for a complete reference.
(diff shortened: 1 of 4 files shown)


arthanson:
@claude /review

bctiemann (approved):

PR #21637, opened 2026-03-10
Closes #21157: Add public models to export template context

### LINKED ISSUE #21157: Add public models to export template context
### NetBox version

v4.4.10

### Feature type

New functionality

### Proposed functionality

As part of the context data available when rendering config templates, #12814 added NetBox model classes that can be accessed by the app or plugin in which they reside, e.g. `There are {{ dcim.Site.objects.count() }} sites.`. I propose adding this same functionality for export templates. Borrowing from the `get_context` method in `ConfigTemplate`, updating the `get_context` method in `ExportTemplate` as shown below seems to work as expected.

```python
from collections import defaultdict
...
...
class ExportTemplate(...):
...
    def get_context(self, context=None, queryset=None):
        _context = defaultdict(dict, {
            'queryset': queryset,
        }) 

        # Populate all public models for reference within the template
        for object_type in ObjectType.objects.public():
            if model := object_type.model_class():
                _context[object_type.app_label][model.__name__] = model

        # Apply the provided context data, if any
        if context is not None:
            _context.update(context)

        return _context
```

### Use case

I was trying to write an export template for IPAM data and since there is currently no direct relationship between an `IPAddress` and its parent `Prefix` object, I was finding it difficult to make efficient queries (not N+1) without access to either the `IPAddress` or `Prefix` models in the template.

### Database changes

None

### External dependencies

None

jnovinger:
### Fixes: #21157

Move shared `get_context()` logic from `ConfigTemplate` into `RenderTemplateMixin` so `ExportTemplate` also gets access to all public model classes. This enables export templates to perform cross-model lookups (e.g. resolving a parent `Prefix` from an `IPAddress`).

`ConfigTemplate` already populated these models via `ObjectType.objects.public()`. This was identical boilerplate that belonged in the shared base class. `ConfigTemplate`'s override is removed (it inherits the base behavior), and `ExportTemplate` now calls `super().get_context()` before adding `queryset`.


jnovinger:
Ugh, closing this as I was sloppy with the base branch. New PR coming shortly.

PR #21645, opened 2026-03-11
Fixes #20077: Fix form field focus bug on Edge

### LINKED ISSUE #20077: GUI bugging out when using saved info
### Deployment Type

Self-hosted

### NetBox Version

v4.3.2

### Python Version

3.10

### Steps to Reproduce

This issues seems to only be present in Microsoft Edge, I could not replicate the problem when using Google Chrome.

This issue is replicable, by doing the following:
1. devices > devices > add a new device > then adding a new device
2. devices > devices > add a new device > click on "Name" > try using the saved info

### Expected Behavior

Saved information to be filled in.

### Observed Behavior

While trying to add new devices to our NetBox, there seems to be a recurring problem whever the "saved info"/"last used" in Microsoft Edge is clicked.
The GUI bugs out and seems to jumping through all textbox, i.e. if you were to spam "tab".
Only reloading the website stops it from continuously scrolling through text boxes.

jeremystretch:
### Fixes: #20077                                                                                                                                                           
                                                                                                                                                                        
Root cause: TomSelect's open() method calls focus(), which:    
                                                                                                       
1. Synchronously calls `control_input.focus()` (moving browser focus to this instance)
2. Schedules `setTimeout(() => { ignoreFocus = false; onFocus(); }, 0)`
                                                                                                                                                                        
When Edge's autofill rapidly focuses multiple inputs, the sequence becomes a ping-pong loop:
- Edge focuses select1 → open() → focus() → control_input.focus() → setTimeout A
- Edge focuses select2 → select1 gets onBlur() → select1 closes (isOpen=false)
- setTimeout A fires → select1 calls onFocus() → open() (since isOpen=false again!) → focus() → steals focus back from select2
- select2 closes → select2's pending setTimeout fires → steals focus back from select1 → infinite loop

Changes:
- New file `netboxTomSelect.ts`: NetBoxTomSelect extends TomSelect with the patched `focus()`
- `dynamicTomSelect.ts`: DynamicTomSelect now extends NetBoxTomSelect instead of TomSelect
- `static.ts`: `initStaticSelects()` and `initColorSelects()` use NetBoxTomSelect instead of TomSelect (static selects with pre-loaded options are equally susceptible to the loop)

I've verified that this addresses the problematic behavior using the Edge browser.


jeremystretch:
@claude /review

arthanson (approved):
There is still some weirdness on edge, if I select the item from the dropdown, it disables that edit field and then goes to the bottom of the page to the last tom-select. This seems just edge specific weird behavior, so this makes it usable - stopping the screen freaking out.  The follow-on I think is just customizing for edge, which we probably don't want to do.

PR #21648, opened 2026-03-11
Closes #20152: Add support for disabling Device and Module bays

### LINKED ISSUE #20152: Support for marking module bays and device bays as disabled
### NetBox version

v4.3.6

### Feature type

Data model extension

### Proposed functionality

Currently, NetBox allows interfaces to be marked as disabled, which helps indicate that they are not available for use. It would be useful to have the same functionality for module and device bays, allowing them to be marked as disabled when they are unavailable, decommissioned, or reserved.

### Use case

Devices can have module bays that physically exist, but are disabled in specific circumstances. An example of that is switches that disable ports based on the feature set used.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Fixes: #20152

This PR adds an `enabled` boolean field to `ModuleBay`, `ModuleBayTemplate`, `DeviceBay`, and `DeviceBayTemplate`.

The intent is to model bay availability as a simple on/off state, which feels like the best fit for this feature and keeps the implementation lightweight. Disabled bays are now reflected in the UI and cannot be used for new device or module installation. Forms, filters, serializers, and tables have also been updated so the new field is supported consistently across NetBox.

### DIFF
--- a/docs/models/dcim/devicebay.md
+++ b/docs/models/dcim/devicebay.md
@@ -23,3 +23,8 @@ The device bay's name. Must be unique to the parent device.
 ### Label
 
 An alternative physical label identifying the device bay.
+
+### Enabled
+
+Whether this device bay is enabled. Disabled device bays are not available for installation.
+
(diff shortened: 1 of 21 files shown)


pheus:
@claude /review

pheus:
@claude /review

1. has been addressed
2. has been addressed
3. can't be implemented, ModuleBay is on the reverse side of the relation (Module points to ModuleBay) → there's no installed_module_id column on ModuleBay
4. is correct
5. no issue
6. has been addressed
7. no issue

bctiemann (approved):

PR #21649, opened 2026-03-11
Fixes #21556: Restore previous value (if applicable) after clearing related dropdown

jeremystretch:
### Fixes: #21556

Extend the `load()` method on DynamicTomSelect to preserve the previously-selected value (if any) after updating the choices for a select field if it's still a valid choice.


jeremystretch:
@claude /review

bctiemann (approved):

PR #21665, opened 2026-03-13
Closes #19034: Add calculated `RackReservation.unit_count`, with min/max filtering

### LINKED ISSUE #19034: Rack Reservations - Add a count of U's to reservations (list and single view)
### NetBox version

v4.2.5

### Feature type

Data model extension

### Proposed functionality

have a counter reservations for how many u's that reservation is for (list and ui views + api)

### Use case

when viewing rack reservations list single reservation view it would be great if it had a counter of how many U's are in the reservation.

being able to sort and filter list views by that number would also be useful

being able to find reservations bigger than X u or smaller than X u would be great

### Database changes

_No response_

### External dependencies

_No response_

bctiemann:
### Closes: #19034

Adds a calculated `unit_count` field to `RackReservation` which is reflected in list and detail views and the API, and can be used for filtering with min/max parameters (`unit_count_min`, `unit_count_max`) in both UI and API views.


bctiemann:
@claude /review

bctiemann:
Addressed GraphQL and API unit test gaps and clarified the documentation text.

jnovinger (approved):
One thing I noticed: `unit_count` is in `default_columns` now, but the bulk edit and bulk delete views (`RackReservationBulkEditView`, `RackReservationBulkDeleteView`) use `RackReservation.objects.all()` as their queryset. The confirmation table renders the column as dashes since those querysets don't have the `CARDINALITY` annotation.

The `RackRackReservationsView` is fine because it passes through the filterset, which adds the annotation in `filter_queryset()`.

Not a blocker as the confirmation screens are transient. But maybe worth annotating those querysets too for consistency.

jnovinger (approved):

PR #21671, opened 2026-03-13
Fixes #21578: Enable assignment of scope object by name when bulk importing prefixes/VLAN groups

jeremystretch:
### Fixes: #21578

- Add `scope_name` field on ScopedImportForm
- Extend ScopedImportForm validation
- VLANGroupImportForm now inherits from ScopedImportForm
- Add `scope_name` to field lists on subclasses


jeremystretch:
@claude /review

arthanson (approved):

PR #21685, opened 2026-03-16
Fixes #21673: UI Bug with Displaying Primary IP Address with NAT IP on a VM

### LINKED ISSUE #21673: UI Bug with Displaying Primary IP Address with NAT IP on a Virtual Machine
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.4

### Python Version

3.12

### Steps to Reproduce

1. Create an IP
2. Create another IP and assign it as the outside NAT IP for the first IP
3. Create a Virtual Machine
4. Assign the first IP to the virtual machine
5. View the virtual machine with attention given to the Primary IPv4 NAT address. The brackets are spread very far apart on either side of the NAT IP

### Expected Behavior

There should be no extra spacing before and after the brackets surrounding the NAT IP.

### Observed Behavior

There is a large amount of extra space surrounding the NAT IP before and after the brackets.

It looks like this

Primary IPv4  10.10.100.129    (                     NAT: 10.20.30.30                      )

But should look like this

Primary IPv4  10.10.100.129    (NAT: 10.20.30.30)

I can submit a PR with a fix.  It's almost identical to [the bug](https://github.com/netbox-community/netbox/issues/21527) I previously submitted for this problem on devices.

robduffy2010:
### Fixes: #21673

UI Bug affecting displaying of Primary IP Address with NAT IP on a Virtual Machine.

### DIFF
--- a/netbox/templates/virtualization/virtualmachine/attrs/ipaddress.html
+++ b/netbox/templates/virtualization/virtualmachine/attrs/ipaddress.html
@@ -1,10 +1,12 @@
 {% load i18n %}
-<a href="{{ value.get_absolute_url }}"{% if name %} id="attr_{{ name }}"{% endif %}>{{ value.address.ip }}</a>
-{% if value.nat_inside %}
-  ({% trans "NAT for" %} <a href="{{ value.nat_inside.get_absolute_url }}">{{ value.nat_inside.address.ip }}</a>)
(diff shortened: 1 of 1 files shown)


pheus (approved):
Thanks, @robduffy2010 !

PR #21689, opened 2026-03-17
Fixes #21686: Fix stale Circuit Termination association when reassigning a Termination to another Circuit

### LINKED ISSUE #21686: Editing Circuit Termination Circuit leaves orphaned association.
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.3

### Python Version

3.13

### Steps to Reproduce

Initial state, Circuit X with A Termination set assigned to a site. A Term Cabled to Device Interface.
Circuit Y newly created with no terminations set.
1. Select Edit on Circuit X A Termination.
2. Change Circuit value to Circuit Y
3. Save

### Expected Behavior

Circuit X - A Termination should be set to Null,
A Termination Object should have a circuit value of Circuit Y
Circuit Y - A Termination should be set to A Termination Object.

### Observed Behavior

A Termination Obj data only references Circuit Y as expected.
Circuit Y shows A term as expected.

Circuit X retains A Termination value pointing at A Termination Object.

pheus:
### Fixes: #21686

When a circuit termination is reassigned to a different circuit, the previous circuit could retain a stale cached association. This change moves the cache update logic from the signal handler into `CircuitTermination.save()`, where the original `circuit_id` and `term_side` values can be tracked so the old cache is cleared correctly when either value changes.

Tests have also been added to cover the relevant cache update scenarios and help prevent regressions.

### DIFF
--- a/netbox/circuits/models/circuits.py
+++ b/netbox/circuits/models/circuits.py
@@ -347,6 +347,13 @@ class CircuitTermination(
         verbose_name = _('circuit termination')
         verbose_name_plural = _('circuit terminations')
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Cache original values to detect changes
+        self._orig_circuit_id = self.__dict__.get('circuit_id')
+        self._orig_term_side = self.__dict__.get('term_side')
+
     def __str__(self):
         return f'{self.circuit}: Termination {self.term_side}'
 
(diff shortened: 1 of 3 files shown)


pheus:
@claude /review

pheus:
@claude /review All three issues have been addressed.

pheus:
Reverted the `is not None` change back to truthiness checks. Turns out `term_side` can be an empty string in some code paths, which caused:

```
FieldDoesNotExist("Circuit has no field named 'termination_'")
```

The truthiness check handles both `None` and `''` correctly, so it was actually the safer option all along.

netbox/circuits/signals.py
```diff
@@ -6,21 +6,10 @@
...
-    instance.circuit.refresh_from_db()
-    setattr(instance.circuit, termination_name, instance)
-    instance.circuit.save()
-
-
 @receiver((post_save, post_delete), sender=CircuitTermination)
 def rebuild_cablepaths(instance, raw=False, **kwargs):
     """
-    Rebuild any CablePaths which traverse the peer CircuitTermination.
+    Rebuild any CablePaths, which traverse the peer CircuitTermination.
```
jeremystretch: We're not rebuilding _any_ CablePaths; only those which traverse the peer CircuitTermination.
```suggestion
    Rebuild any CablePaths which traverse the peer CircuitTermination.
```

netbox/circuits/models/circuits.py
```diff
@@ -360,11 +367,39 @@ def clean(self):
             raise ValidationError(_("A circuit termination must attach to a terminating object."))
 
     def save(self, *args, **kwargs):
+        is_new = self.pk is None
```
jeremystretch: `self.pk is None` works fine, but I'm trying to slowly move us toward this Django convention.
```suggestion
        is_new = self._state.adding
```

jeremystretch (changes requested):
Just some minor nits

pheus:
> Just some minor nits

Thanks for the fast and thoughtful review! I've made the suggested changes.

jeremystretch (approved):

PR #21690, opened 2026-03-17
#21662 Increase rf_channel_frequency Precision

arthanson:


netbox/dcim/models/device_components.py
```diff
@@ -808,7 +808,7 @@ class Interface(
     )
     rf_channel_frequency = models.DecimalField(
         max_digits=7,
```
pheus: This probably needs `max_digits=8` as well. With `decimal_places=3` and `max_digits=7`, the field max drops from `99999.99` to `9999.999`, which would regress existing high-frequency channels.

pheus (changes requested):
Thanks!

I think we need one more change before this is ready: `decimal_places` is being increased to `3`, but `max_digits` stays at `7`. That reduces the maximum whole-number portion from five digits to four, so values above `9999.999` would no longer validate.

That looks like a regression for existing wireless channels, since NetBox already has built-in 60 GHz channel frequencies in the `58320`+ MHz range.

pheus (approved):
Thanks!

PR #21691, opened 2026-03-17
#14329 Improve diffs for custom_fields

arthanson:
### Fixes: #14329 

Needed to add a deep_compare_dict to process the diffs instead of the current shallow_compare_dict, as custom_fields is a nested dict inside the diff.  Two places for change - one modifying the "Difference" panel and the other for the "Pre-Change Data" and "Post-Change Data"

I created 6 custom fields on a Tenant and modified one of them. On the original note all the custom fields are highlighted, even though only modified one of them:

Monosnap Tenancy | tenant Cyberdyne Systems updated by ahanson | NetBox 2026-03-17 09-04-29

On this version, only the custom field that actually got modified is highlighted:

Monosnap Tenancy | tenant Cyberdyne Systems updated by ahanson | NetBox 2026-03-17 09-01-09

### DIFF
--- a/netbox/core/models/change_logging.py
+++ b/netbox/core/models/change_logging.py
@@ -11,7 +11,7 @@ from mptt.models import MPTTModel
 from core.choices import ObjectChangeActionChoices
 from core.querysets import ObjectChangeQuerySet
 from netbox.models.features import ChangeLoggingMixin, has_feature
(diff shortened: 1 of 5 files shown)


arthanson:
@claude /review

bctiemann (approved):

PR #21725, opened 2026-03-20
#20923: Convert `circuits` views to new UI layout

pheus:
As part of https://github.com/netbox-community/netbox/issues/20923, this PR migrates all views under the `circuits` app from bespoke HTML templates to the new declarative UI layout.


pheus:
@claude /review check for any inconsistencies between the new layout and the original HTML templates.

pheus:
@claude /review

1. CircuitPanel — provider_account label mismatch: This is the new style.
2. CircuitGroupAssignmentPanel — member label mismatch: This panel used multiple times, therefore the field should be more generic.
3. ProviderPanel — adds name field not present in original: This fixes an old bug.
4. CircuitTerminationView — new panel title not in original: This fixes an old bug.
5. circuit_termination.html — redundant {% if object %} guard: I've fixed that in the recent push.

pheus:
@claude /review

1. Missing {% trans %} for custom fields fallback label — untranslatable string: has been fixed in the newest push.
2. fs-5 on content-type label is larger than body text: `netbox.css` overwrite this to `.75rem !important`, so no problem.
3. Indentation misalignment after {% include %} in circuit_circuit_termination.html: has been fixed in the newest push.

jeremystretch (approved):
Great work @pheus!

PR #21726, opened 2026-03-20
Closes #21702: Include originating HTTP request in outbound webhook context data

jeremystretch:
### Closes: #21702

Adds a `request` key to the webhook data if a request is associated with the origination of the webhook.

Note: We're not attaching a complete representation of the request in the interest of both security and brevity.

### DIFF
--- a/docs/models/extras/webhook.md
+++ b/docs/models/extras/webhook.md
@@ -90,4 +90,4 @@ The following context variables are available to the text and link templates.
 | `snapshots`   | Pre- and post-change snapshots of the object         |
 
 !!! warning "Deprecation of legacy fields"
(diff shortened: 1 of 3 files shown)


jeremystretch:
@claude /review

pheus (changes requested):
Looks good overall!

My only suggestion is to sync the docs with the schema introduced here. The current webhook docs still point users toward `request.user.username` / `request.request_id`, while this change exposes a lightweight serialized `request` object with `id`, `method`, `path`, and `user`.

It might also be worth mentioning that `request` may be absent when there is no originating HTTP request.

pheus (approved):
Thanks!

PR #21728, opened 2026-03-23
fix: escape exception messages in mark_safe calls and fix SQL interpolation

Aprazor:
Multiple XSS and security fixes:
1. truncate_middle filter: escape first_part/second_part before mark_safe
2. CustomLinkColumn.render: escape exception in title attribute
3. render_widget dashboard tag: escape exception message
4. custom_links tag: escape exception and cl.name in both ungrouped
   and grouped link error paths
5. core/views.py: use current_database() directly instead of f-string
   interpolation of db_name into SQL
6. core/jobs.py: guard max(releases) against empty list

Signed-off-by: Aprazors <Aprazors@gmail.com>


jeremystretch:

PR #21731, opened 2026-03-23
Closes #12024: Allow Virtual Machines to be assigned directly to Devices

### LINKED ISSUE #12024: Possibility to assign virtual machines (containers) directly to devices.
### NetBox version

v3.4.6

### Feature type

### Proposed functionality

I am submitting for consideration the idea of enabling the assignment of virtual machines (rather in the sense of containers) directly to devices.

Currently, if you want to assign a VM to a device it is necessary to create a cluster and assign that device to it. When trying to add a VM to a device without specifying a cluster, an error is displayed:
image

If I want to bypass this by adding multiple devices to a cluster, it results in the fact that I cannot have two VMs with the same name in a given cluster, one active and the other offline. In the use cases I am dealing with, this situation occurs.

In short, the above would only require:
- allowing a virtual machine to be assigned to a device without being assigned to a cluster,
- checking the uniqueness of virtual machine names within a single device,

### Use case

In the case where I'm dealing with containers (e.g. LXC or Docker), the cluster is a single device so it would be very practical to allow such a direct assignment of a virtual machine to a device and to keep an eye on the uniqueness of the virtual machine's name only within the device, so that on another device it can be of the same name in a different strangeness (offline, planned, etc.).

In cases where containers are used on devices and migrated frequently, it is useful to have information that a container is offline on another device.

I hope it would be useful not only for me. Maybe someone else will comment on whether they use Netbox in such use cases?

### Database changes

Probably none, unless a foreign key in some table?

### External dependencies

None

pheus:
### Summary

This PR allows virtual machines to be assigned directly to a standalone device, without requiring a cluster. Previously, a VM could only be placed on a site or within a cluster (with an optional device pin). This adds a third placement mode: direct assignment to a device that doesn't belong to any cluster.

### Changes

**Model (`VirtualMachine`)**
- A VM can now be assigned to a site, a cluster, a standalone device, or a combination (e.g. cluster + device pin)
- Site is automatically inherited from the assigned cluster or device when not explicitly set
- If a device belongs to a cluster, the cluster must still be specified on the VM
- Added device-scoped uniqueness constraints (name must be unique per device and tenant for standalone assignments)

**Forms**
- Updated `VirtualMachineForm`, `VirtualMachineImportForm`, and `VirtualMachineBulkEditForm` with descriptive help texts for the site, cluster, and device fields

**Documentation**
- Updated model docstring, `virtualmachine.md` model docs, and `virtualization.md` feature page to describe all three placement modes
- Added "New in NetBox v4.6" callout for the standalone device feature

**Tests**
- Added comprehensive test coverage for all valid and invalid placement combinations, standalone device assignment, name uniqueness per device, site auto-inheritance, tenant scoping, and case-insensitive name checks

### DIFF
--- a/docs/features/virtualization.md
+++ b/docs/features/virtualization.md
@@ -1,17 +1,19 @@
 # Virtualization
 
-Virtual machines and clusters can be modeled in NetBox alongside physical infrastructure. IP addresses and other resources are assigned to these objects just like physical objects, providing a seamless integration between physical and virtual networks.
+Virtual machines, clusters, and standalone hypervisors can be modeled in NetBox alongside physical infrastructure. IP addresses and other resources are assigned to these objects just like physical objects, providing a seamless integration between physical and virtual networks.
 
 ```mermaid
 flowchart TD
     ClusterGroup & ClusterType --> Cluster
     Cluster --> VirtualMachine
+    Device --> VirtualMachine
     Platform --> VirtualMachine
     VirtualMachine --> VMInterface
 
 click Cluster "../../models/virtualization/cluster/"
 click ClusterGroup "../../models/virtualization/clustergroup/"
 click ClusterType "../../models/virtualization/clustertype/"
+click Device "../../models/dcim/device/"
 click Platform "../../models/dcim/platform/"
 click VirtualMachine "../../models/virtualization/virtualmachine/"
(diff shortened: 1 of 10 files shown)


pheus:
@claude /review

pheus:

arthanson (approved):

PR #21736, opened 2026-03-23
Closes #21735: Replace deprecated Strawberry scalar for `BigInt`

### LINKED ISSUE #21735: Replace deprecated Strawberry scalar registration for `BigInt`
### Proposed Changes

NetBox defines the GraphQL `BigInt` scalar in `netbox/netbox/graphql/scalars.py` by passing a class directly to `strawberry.scalar()`. When the development server is run with warnings enabled, Strawberry emits a deprecation warning during startup (currently twice):

```bash
python3 -Wd netbox/manage.py runserver 0.0.0.0:8000
```

```text
/netbox/netbox/graphql/scalars.py:5: DeprecationWarning: Passing a class to strawberry.scalar() is deprecated. Use StrawberryConfig.scalar_map instead for better type checking support. See: https://strawberry.rocks/docs/types/scalars
  BigInt = strawberry.scalar(
```

This scalar registration should be updated to use Strawberry's supported configuration mechanism, such as `StrawberryConfig.scalar_map` or the current upstream-recommended equivalent, instead of the deprecated `strawberry.scalar()` pattern. The goal is to preserve the existing GraphQL behavior for `BigInt` while eliminating the warning from development and test runs.

No user-facing changes to workflows, data models, or the UI are expected.

### Justification

This is a codebase-only cleanup that keeps NetBox aligned with Strawberry's supported API and reduces warning noise for developers running NetBox with warnings enabled. Addressing it now should help avoid future breakage once the deprecated behavior is removed upstream, and it keeps local development and CI output cleaner when deprecations are surfaced more aggressively.

pheus:
### Fixes: #21735

Replace the deprecated `strawberry.scalar()` class-based registration of `BigInt` with the approach recommended by Strawberry's current API.

`BigInt` is now defined as a `NewType` in `scalars.py`, and the corresponding GraphQL scalar (`BigIntScalar`) is registered via `StrawberryConfig.scalar_map` in `schema.py`. The exported name `BigInt` is unchanged, so existing field annotations throughout the codebase continue to work without modification.

This eliminates the `DeprecationWarning` that currently appears (twice) on startup when running with warnings enabled, and keeps NetBox aligned with Strawberry's supported API ahead of the eventual removal of the old pattern.

### DIFF
--- a/netbox/netbox/graphql/scalars.py
+++ b/netbox/netbox/graphql/scalars.py
@@ -1,10 +1,12 @@
-from typing import Union
+from typing import NewType
 
 import strawberry
 
-BigInt = strawberry.scalar(
-    Union[int, str],  # type: ignore
+BigInt = NewType('BigInt', int)
+
+BigIntScalar = strawberry.scalar(
+    name='BigInt',
     serialize=lambda v: int(v),
     parse_value=lambda v: str(v),
-    description="BigInt field",
+    description='BigInt field',
 )
--- a/netbox/netbox/graphql/schema.py
+++ b/netbox/netbox/graphql/schema.py
@@ -16,6 +16,8 @@ from virtualization.graphql.schema import VirtualizationQuery
 from vpn.graphql.schema import VPNQuery
 from wireless.graphql.schema import WirelessQuery
 
+from .scalars import BigInt, BigIntScalar
+
 
 @strawberry.type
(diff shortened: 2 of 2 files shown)


pheus:
@claude /review

jnovinger (approved):

PR #21744, opened 2026-03-24
Closes #20808: Show occupying device in rack position selector

### LINKED ISSUE #20808: UX enhancement Device racking position selection - Device names
### NetBox version

v4.4.5

### Feature type

Data model extension

### Proposed functionality

In Device model, when assigning a Device to a Rack. In the list view to select the position, already occupied Us are greyed out. Add the Device name that is occupying the position.

### Use case

Makes it more intuitive to position a Device in the correct position.

### Database changes

_No response_

### External dependencies

_No response_

Guibi1:
### Fixes: #20808 

Added the name and device type of the racked device inside the device rack position select.

image

### DIFF
--- a/netbox/dcim/api/serializers_/rackunits.py
+++ b/netbox/dcim/api/serializers_/rackunits.py
@@ -1,6 +1,7 @@
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
+from django.utils.translation import gettext as _
 
 from dcim.choices import *
 from netbox.api.fields import ChoiceField
@@ -25,7 +26,17 @@ class RackUnitSerializer(serializers.Serializer):
     device = DeviceSerializer(nested=True, read_only=True)
     occupied = serializers.BooleanField(read_only=True)
     display = serializers.SerializerMethodField(read_only=True)
+    description = serializers.SerializerMethodField(read_only=True)
 
     @extend_schema_field(OpenApiTypes.STR)
     def get_display(self, obj):
-        return obj['name']
+        display = obj['name']
+        if obj['occupied'] and (device := obj['device']):
+            return _('{rack_unit} - {device}').format(rack_unit=display, device=device)
+
+        return display
+
+    @extend_schema_field(OpenApiTypes.STR)
+    def get_description(self, obj):
+        if obj['occupied'] and (device := obj['device']):
+            return str(device.device_type)
--- a/netbox/dcim/api/serializers_/rackunits.py
+++ b/netbox/dcim/api/serializers_/rackunits.py
@@ -1,4 +1,3 @@
-from django.utils.translation import gettext as _
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
@@ -26,20 +25,12 @@ class RackUnitSerializer(serializers.Serializer):
     device = DeviceSerializer(nested=True, read_only=True)
     occupied = serializers.BooleanField(read_only=True)
     display = serializers.SerializerMethodField(read_only=True)
-    device_display = serializers.SerializerMethodField(read_only=True)
-    device_type = serializers.SerializerMethodField(read_only=True)
+    description = serializers.SerializerMethodField(read_only=True)
 
     @extend_schema_field(OpenApiTypes.STR)
     def get_display(self, obj):
         return obj['name']
 
-    @extend_schema_field(OpenApiTypes.STR)
-    def get_device_display(self, obj):
-        if obj['occupied'] and (device := obj['device']):
-            return _('{rack_unit} - {device}').format(rack_unit=obj['name'], device=device)
-        return obj['name']
-
-    @extend_schema_field(OpenApiTypes.STR)
-    def get_device_type(self, obj):
-        if obj['occupied'] and (device := obj['device']):
-            return str(device.device_type)
+    @extend_schema_field(OpenApiTypes.STR)  
+    def get_description(self, obj):  
+        return f'{obj["device"]}' if obj['device'] else None  
--- a/netbox/dcim/api/serializers_/rackunits.py
+++ b/netbox/dcim/api/serializers_/rackunits.py
@@ -1,7 +1,7 @@
+from django.utils.translation import gettext as _
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
-from django.utils.translation import gettext as _
 
 from dcim.choices import *
 from netbox.api.fields import ChoiceField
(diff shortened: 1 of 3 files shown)


pheus (changes requested):
Thanks for working on this. I agree with the goal, but I don't think this implementation works in its current form.

For me, the main issue is not the specific string formatting, but that this changes the public rack-elevation API payload itself. `display` is being repurposed from the rack-unit label into selector-specific UI text, and `description` is also being introduced for that same purpose. As a result, every caller of this endpoint gets different semantics, not just the device position dropdown.

Because of that, I think this needs a different approach.

The rack position field is the consumer we want to improve, so I don't think the general API response should change in order to support that.

I also think the test should reflect that. Right now it locks in the API contract change by asserting the modified `display` and new `description` values. I'd expect the existing rack-elevation response to remain stable, so from my side this needs a rework.

Thanks again for the PR.

Guibi1:
I see that makes sense. I will add the device information to another api field then. Im pretty sure that will require me to add JS though?

pheus:
Thanks for looking into this.

I don't think adding custom JS should be necessary, since we already have API-backed widgets for this kind of use case, although I haven't looked into the exact implementation path myself.

Could you please take another look through the existing codebase and see whether there is a solution that keeps the current API contract intact and avoids introducing additional code we would need to maintain?

Guibi1:
The API backed widget is what is currently being used, and it simply uses the display and description fields of the objects it receives. We can either
- Change what the api returns (my current impl)
- Change how the widget shows the data on the client (in the JS)
- Create a new api endpoint 
- Create a new widget (probably a better idea than writing JS)

If we dont want to modify the API, i feel like making a new path is the better option.
I might be missing an option that would make it easier... (like an attr on the widget to select another field so we can preserve display)

Guibi1:
> I might be missing an option that would make it easier... (like an attr on the widget to select another field so we can preserve display)

Welp i just found `ts-label-field`, which allows us to use another field on the data. I can add a `device-display` field on the api that contains the name of the device, and undo my changes to the `display` field.
Would that work?
This PR would then just add two fields on the rack unit serializer. (description and device-display)

pheus:
You might also want to take a look at the widget `context` option, depending on how you plan to pass the additional data through.

I haven't had the bandwidth to work through all implementation details on my side, so I can't fully validate every possible approach, but this direction sounds reasonable to me. Thanks for digging into it.

netbox/dcim/api/serializers_/rackunits.py
```diff
@@ -26,7 +26,20 @@ class RackUnitSerializer(serializers.Serializer):
...
 
     @extend_schema_field(OpenApiTypes.STR)
     def get_display(self, obj):
         return obj['name']
+
+    @extend_schema_field(OpenApiTypes.STR)
+    def get_device_display(self, obj):
+        if obj['occupied'] and (device := obj['device']):
+            return _('{rack_unit} - {device}').format(rack_unit=obj['name'], device=device)
+        return obj['name']
```
pheus, line 40:
```suggestion

```

netbox/dcim/api/serializers_/rackunits.py
```diff
@@ -26,7 +26,20 @@ class RackUnitSerializer(serializers.Serializer):
...
+
+    @extend_schema_field(OpenApiTypes.STR)
+    def get_device_type(self, obj):
+        if obj['occupied'] and (device := obj['device']):
+            return str(device.device_type)
```
pheus, line 45:
```suggestion
    @extend_schema_field(OpenApiTypes.STR)
    def get_description(self, obj):
        return f'{obj["device"]}' if obj['device'] else None
```

netbox/dcim/api/serializers_/rackunits.py
```diff
@@ -26,7 +26,20 @@ class RackUnitSerializer(serializers.Serializer):
     device = DeviceSerializer(nested=True, read_only=True)
     occupied = serializers.BooleanField(read_only=True)
     display = serializers.SerializerMethodField(read_only=True)
+    device_display = serializers.SerializerMethodField(read_only=True)
+    device_type = serializers.SerializerMethodField(read_only=True)
```
pheus, line 30:
```suggestion
    description = serializers.SerializerMethodField(read_only=True)
```

netbox/dcim/api/serializers_/rackunits.py
```diff
@@ -1,10 +1,10 @@
+from django.utils.translation import gettext as _
```
pheus, line 1:
```suggestion

```

netbox/dcim/forms/model_forms.py
```diff
@@ -600,6 +590,8 @@ class DeviceForm(TenancyForm, PrimaryModelForm):
             api_url='/api/dcim/racks/{{rack}}/elevation/',
             attrs={
                 'ts-disabled-field': 'device',
+                'ts-label-field': 'device_display',
+                'ts-description-field': 'device_type',
```
pheus, line 594:
```suggestion

```

netbox/dcim/tests/test_api.py
```diff
@@ -451,6 +451,36 @@ def test_get_rack_elevation(self):
         response = self.client.get(f'{url}?q=U10', **self.header)
         self.assertEqual(response.data['count'], 2)
 
+    def test_get_rack_elevation_display_includes_occupying_device(self):
```
pheus, line 454: This needs to be adopted to the new API scheme.

pheus (changes requested):
Thanks for reworking this and for taking another pass at it.

I had a chance to look at it more closely, and I think we are getting much closer.

Looking at other parts of the NetBox UI, additional information in dropdowns is typically shown underneath each option via the `description` field. In the current implementation, the dropdown also includes the device type, which was not part of the original feature request.

Because of that, I think we can keep this implementation simpler.

From my side, I’d prefer to keep `display` unchanged and add a single `description` field on the rack unit serializer for the occupying device information, rather than introducing separate `device_display` / `device_type` fields plus additional widget configuration.

That would keep the existing `display` semantics intact, avoid the extra widget-specific code, and still address the UI problem we want to solve here.

Could you please rework it in that direction? Thanks again for the update.

xkilian:
The device type is integral in showing hardware that do not have device names. It is complementary. I can update the issue with the information to support this.

xkilian:
Hardware are shelves, cable management, blank covers, etc.

pheus:
Thanks for taking the time to share that context, @xkilian !

I understand the use case you are describing. That said, PRs are not the right place to expand or redefine the scope of an already accepted feature request.

If you would like to cover that additional use case, please open a separate discussion or feature request with the proposed behavior and implementation direction. For this PR, we need to stay aligned with the scope that was originally accepted, and updating the feature request at this stage would not change the scope of this review.

To keep this PR review focused, I’m going to mark this thread, including my own reply, as off-topic. Thanks for your understanding.

pheus (changes requested):
Thanks for the rework.

This is still not ready for review while CI is failing. Please fix the lint/test failures and make sure CI is green before requesting another review.

Also, please update the new test name/docstring so it matches the current implementation: the device information is now exposed via `description`, not `display`.

Guibi1:
Sorry about that, but i need approval to run the CI...
This is out of scope, but running ruff locally changes hundreds of files. I am probably missing something?

netbox/dcim/tests/test_api.py
```diff
@@ -451,6 +451,34 @@ def test_get_rack_elevation(self):
         response = self.client.get(f'{url}?q=U10', **self.header)
         self.assertEqual(response.data['count'], 2)
 
+    def test_get_rack_elevation_description_is_occupying_device_name(self):
+        """
+        Verify occupied rack units include the occupying device in their display label.
```
pheus:
```suggestion
        Verify occupied rack units include the occupying device in their description.
```

pheus (changes requested):
Thanks for sticking with this and reworking it.

This looks good to me now. The implementation keeps `display` unchanged, adds the occupying device information via `description`, and includes API coverage for occupied and unoccupied rack units.

One small nit: the new test docstring still mentions the display label, but the implementation now uses `description`.

Thanks again

pheus (approved):
Thanks for the rework and for getting CI green.

This looks good to me now. The implementation keeps `display` unchanged, exposes the occupying device information via `description`, and includes coverage for both occupied and unoccupied rack units.

Approved from my side. Thanks again.

PR #21746, opened 2026-03-24
Closes #19025: Add schema validation for JSON custom fields

### LINKED ISSUE #19025: Add JSON schema validation for JSON custom fields
### NetBox version

v4.2.6

### Feature type

Data model extension

### Proposed functionality

With the JSON schema infrastructure added to support #19002, I feel like this is finally an appropriate time to make this feature request. A proposal to use regex for the same purpose was previously made in #9236, but I agree that was not a good option. Therefore, my proposal is to add a new `schema` JSONField to the `CustomField` data model and add JSONSchema validation when cleaning custom fields. This would be an optional field, just like other custom field validation, to allow for data structure validation only if desired. The addition of the dynamic form rendering when a schema is present would be an interesting further enhancement. A modal or other reference to the schema when entering data would also be a useful solution.

### Use case

We have found in our environment a growing interest in leveraging JSON type custom fields to capture some more complex data structures. These are often then fed into scripts or other integrations, so their structure is critical. While custom validators can be used, we have not found that solution to be scalable or user-friendly, as it can often be difficult when working in the UI to remember the exact syntax that is expected and it requires backend administration when changes need to be made.

### Database changes

Add an optional `schema` JSONField to the `CustomField` data model.

### External dependencies

This should only need to leverage the newly introduced `jsonschema` dependency.

jeremystretch:
### Closes: #19025

Adds an optional `validation_schema` JSON field on the CustomField model for validating JSON-type custom field values.

### DIFF
--- a/docs/customization/custom-fields.md
+++ b/docs/customization/custom-fields.md
@@ -63,6 +63,7 @@ NetBox supports limited custom validation for custom field values. Following are
 * Text: Regular expression (optional)
 * Integer: Minimum and/or maximum value (optional)
 * Selection: Must exactly match one of the prescribed choices
+* JSON: Must adhere to the defined validation schema (if any)
 
 ### Custom Selection Fields
 
--- a/docs/models/extras/customfield.md
+++ b/docs/models/extras/customfield.md
@@ -118,3 +118,7 @@ For numeric custom fields only. The maximum valid value (optional).
 ### Validation Regex
 
 For string-based custom fields only. A regular expression used to validate the field's value (optional).
+
+### Validation Schema
+
+For JSON custom fields, users have the option of defining a [validation schema](https://json-schema.org). Any value applied to this custom field on a model will be validated against the provided schema, if any.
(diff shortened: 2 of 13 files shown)


jeremystretch:
This is waiting on a fix for [#21747](https://github.com/netbox-community/netbox/issues/21747)

jeremystretch:
@claude /review

jeremystretch:
1. Fixed: Removed the check for `properties` from `validate_schema()`. This shouldn't cause any issues with ModuleTypeProfile.
2. Fixed
3. Fixed
4. Ignored; this is consistent with e.g. `ModuleTypeProfileForm.schema`
5. Ignored; this is intentional

jeremystretch:
@claude /review

jeremystretch:
1. Invalid feedback AFAICT; the behavior is intentional
2. This is an acceptable limitation
3. IMO a test for this is unnecessary
4. This is intentional

jnovinger (approved):
Approved as is, but one possible addition if you want: `docs/customization/custom-fields.md` (the Validation section) and `docs/models/extras/customfield.md` are both missing entries for `validation_schema`.

jnovinger (approved):

PR #21748, opened 2026-03-24
Fixes #21747: Skip search caching when encountering an invalid schema during migrations

jeremystretch:
### Fixes: #21747

The search backend should silently capture and ignore exceptions relating to the database schema when running in a database migration.

### DIFF
--- a/netbox/netbox/search/backends.py
+++ b/netbox/netbox/search/backends.py
@@ -5,7 +5,7 @@ import netaddr
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ImproperlyConfigured
-from django.db import OperationalError, ProgrammingError
+from django.db import ProgrammingError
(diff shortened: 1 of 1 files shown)


arthanson:
@claude /review

arthanson (changes requested):
The logging in the exception case is probably a good idea, could save some debugging time just-in-case this causes any issue.  

I don't think the comment re the removal_handler is valid as it is in reverse so this would be run after, but worth considering.

jeremystretch:
1. I think this is safe to ignore as deletions are not schema-specific, and there's no chance of deleting an object before the model has been created.
2. Agree with logging implementation
3. Decided to catch only ProgrammingError exceptions to avoid conflating other potential migration issues
4. Thanks Claude, sorry I'm not perfect like you.

arthanson (approved):

PR #21750, opened 2026-03-25
Fixes #21655: Fix duplicate SQL queries on serializing custom fields

miaow2:
### Closes: #21655

Added `CustomFieldListSerializer` that caches `object` and `multiobject` types custom fields data

### DIFF
--- a/netbox/extras/api/customfields.py
+++ b/netbox/extras/api/customfields.py
@@ -49,7 +49,7 @@ class CustomFieldsDataField(Field):
         # TODO: Fix circular import
         from utilities.api import get_serializer_for_model
         data = {}
-        cache = getattr(self, '_cf_object_cache', None)
+        cache = self.parent.context.get('cf_object_cache')
 
         for cf in self._get_custom_fields():
             if cache is not None and cf.type in (
@@ -131,5 +131,5 @@ class CustomFieldListSerializer(ListSerializer):
                             pks.add(raw)
                 for obj in model.objects.filter(pk__in=pks):
                     cache[(model, obj.pk)] = obj
-            cf_field._cf_object_cache = cache
+            self.child.context['cf_object_cache'] = cache
         return super().to_representation(data)
--- a/netbox/extras/api/customfields.py
+++ b/netbox/extras/api/customfields.py
@@ -2,7 +2,7 @@ from django.utils.translation import gettext as _
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
 from rest_framework.fields import Field
-from rest_framework.serializers import ValidationError
+from rest_framework.serializers import ListSerializer, ValidationError
 
 from extras.choices import CustomFieldTypeChoices
 from extras.constants import CUSTOMFIELD_EMPTY_VALUES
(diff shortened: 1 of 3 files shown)


jnovinger:
@miaow2 , wanted to let you know I'm looking at this, between other work. But, given the trickiness of custom fields, I want to give myself some time to think on this.

netbox/netbox/api/serializers/features.py
```diff
@@ -23,6 +23,29 @@ class CustomFieldModelSerializer(serializers.Serializer):
...
     )
 
+    @classmethod
+    def many_init(cls, *args, **kwargs):
+        """
+        We can't call super().many_init() and change the outcome because by the time it returns,
+        the plain ListSerializer is already instantiated.
+        Because every NetBox serializer defines its own Meta which doesn't inherit from a parent Meta,
+        so this would silently not apply to any real serializer.
+        Thats why this method replicated many_init from parent and changed the default value for list_serializer_class.
```
jnovinger:
```suggestion
        this would silently not apply to any real serializer.
        Thats why this method replicates many_init from parent and changed the default value for list_serializer_class.
```

netbox/extras/api/customfields.py
```diff
@@ -87,3 +104,32 @@ def to_internal_value(self, data):
...
+                    raw = item.custom_field_data.get(cf.name)
+                    if raw is not None:
+                        if cf.type == CustomFieldTypeChoices.TYPE_MULTIOBJECT:
+                            pks.update(raw)
+                        else:
+                            pks.add(raw)
+                for obj in model.objects.filter(pk__in=pks):
+                    cache[(model, obj.pk)] = obj
+            cf_field._cf_object_cache = cache
+        return super().to_representation(data)
```
jnovinger, line 135: I'm concerned with how `CustomFieldListSerializer` communicates the cache to `CustomFieldsDataField`. Writing `cf_field._cf_object_cache = cache` directly onto a field instance creates implicit coupling between two classes with no interface or contract. If either is refactored independently, nothing signals the dependency exists. This is a maintenance burden just waiting to bite us.

DRF's context dict is designed for passing data down through the serializer tree, and we already use it for that in this file (see [customfields.py:56](https://github.com/netbox-community/netbox/blob/main/netbox/extras/api/customfields.py#L56) and [gfk_fields.py:18](https://github.com/netbox-community/netbox/blob/main/netbox/netbox/api/gfk_fields.py#L18)). In a `ListSerializer`, `self.child.context` is the same dict as `self.context`, so writing into it before calling `super()` makes it available to child fields via `self.parent.context`:

```python
# In CustomFieldListSerializer.to_representation():
self.child.context['cf_object_cache'] = cache
return super().to_representation(data)

# In CustomFieldsDataField.to_representation():
cache = self.parent.context.get('cf_object_cache')
```

This makes the communication explicit and uses the established channel rather than an attribute side channel.

jnovinger (changes requested):
Hey @miaow2 , thanks for waiting and apologies again for the time it took to digest this. I've left a couple of comments below that I think we need to address.

netbox/extras/api/customfields.py
miaow2: Hi @jnovinger, thanks for your time and review. Now I know more about DRF!

I also had concerns about how to store cache, and did not find better options.

I've pushed the code with your suggested `context` solution, please check.

jnovinger (approved):

PR #21752, opened 2026-03-25
Fix {module} placeholder resolution in module bay position field

### LINKED ISSUE #20467: The "Position" field doesn't resolve the {module} variable in nested modules

mrmrcoleman:
## Summary

Fixes a bug where the `{module}` placeholder in a `ModuleBayTemplate`'s position field was not being resolved when a module was installed, leaving the literal string `{module}` in the position value.

The `{module}` placeholder already worked correctly in the `name` and `label` fields — this brings `position` into parity.

## Changes

| File | Change |
|------|--------|
| `dcim/models/device_component_templates.py` | Add `resolve_position()` method; consolidate shared resolution logic into `_resolve_module_placeholder()` to eliminate duplication across `resolve_name`, `resolve_label`, and `resolve_position` |
| `dcim/tests/test_models.py` | Regression test: module bay template with `position='{module}-1'` resolves to `1-1` when installed under a parent bay with position `1` |

## Testing

All existing dcim model and form tests pass (60 tests). Added one regression test that verifies:

1. Create a device type with a module bay (position `1`)
2. Create a module type with a sub-bay template using `position='{module}-1'`
3. Install the module — verify the sub-bay's position resolves to `1-1`

### DIFF
--- a/netbox/dcim/models/device_component_templates.py
+++ b/netbox/dcim/models/device_component_templates.py
@@ -191,9 +191,6 @@ class ModularComponentTemplateModel(ComponentTemplateModel):
     def resolve_label(self, module):
         return self._resolve_module_placeholder(self.label, module)
 
-    def resolve_position(self, module):
-        return self._resolve_module_placeholder(self.position, module)
-
 
 class ConsolePortTemplate(ModularComponentTemplateModel):
     """
@@ -721,6 +718,9 @@ class ModuleBayTemplate(ModularComponentTemplateModel):
     class Meta(ModularComponentTemplateModel.Meta):
         verbose_name = _('module bay template')
         verbose_name_plural = _('module bay templates')
+
+    def resolve_position(self, module):
+        return self._resolve_module_placeholder(self.position, module)
 
     def instantiate(self, **kwargs):
         return self.component_model(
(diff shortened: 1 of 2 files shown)


jeremystretch:
@claude /review

netbox/dcim/models/device_component_templates.py
```diff
@@ -177,29 +177,22 @@ def _get_module_tree(self, module):
...
+        return self._resolve_module_placeholder(self.label, module)
 
-        if module:
-            modules = self._get_module_tree(module)
-            label = self.label
-            for module in modules:
-                label = label.replace(MODULE_TOKEN, module.module_bay.position, 1)
-            return label
-        return self.label
+    def resolve_position(self, module):
```
jeremystretch: Let's move this method to ModuleBayTemplate as the `position` field is exclusive to that model.

jeremystretch (changes requested):

jeremystretch (approved):

PR #21753, opened 2026-03-25
Implement {module} position inheritance for nested module bays

mrmrcoleman:
## Summary

Enables a single ModuleType to produce correctly named components at any nesting depth by resolving `{module}` in module bay position fields during tree traversal. The user controls the separator through the position field template itself.

**Fixes:** #19796

**Depends on:** #21752 (`{module}` resolution in position field, included in this branch)

## Problem

A ModuleType with a single `{module}` placeholder (e.g., an SFP with interface `SFP {module}`) cannot be reused at different nesting depths. Installing it at depth 2 requires `{module}/{module}` in the template name, forcing users to create duplicate ModuleTypes for every possible nesting level.

## Solution: Position Inheritance

Instead of introducing a new placeholder token, this PR resolves `{module}` in each module bay's **position field** during tree traversal, propagating parent positions through the hierarchy. The user controls the separator by defining it directly in the position field template.

### Example

```
Device bay:      position = "1"
Line card bay:   position = "{module}/2"   --> resolves to "1/2"
SFP template:    name = "SFP {module}"     --> resolves to "SFP 1/2"
```

Using `-` instead of `/`:

```
Device bay:      position = "1"
Line card bay:   position = "{module}-2"   --> resolves to "1-2"
SFP template:    name = "eth{module}"      --> resolves to "eth1-2"
```

### Why not `{module_path}`?

The previous approach (PR #21436) introduced a new `{module_path}` token with a hardcoded `/` separator. This approach is simpler and more flexible:

- No new tokens or constants
- No database migration
- Separator is user-controlled, not hardcoded (addresses feedback from #21667)
- Backwards compatible — inheritance only activates when positions contain `{module}`

Credit to @rhartmann for suggesting this approach in the #21436 discussion.

## Changes

| File | Change |
|------|--------|
| `dcim/models/device_component_templates.py` | Add `_get_inherited_positions()` to resolve positions during tree traversal; update `_resolve_module_placeholder()` with single-token logic (resolves to leaf's inherited position); add `resolve_position()` for position field resolution (#20467); consolidate into `_resolve_placeholders()` with `{vc_position}` support |
| `dcim/forms/common.py` | Add `_get_inherited_positions()` static method; extract `_validate_module_tokens()` private method; update `clean()` to use inherited positions |
| `dcim/tests/test_models.py` | Tests for position inheritance at depth 2 and 3, custom separators, multi-token backwards compatibility, position field resolution |
| `docs/models/dcim/moduletype.md` | Document position inheritance behavior |

## Testing

All 1103 dcim tests pass (model, form, API). Added 5 new tests:

- **Position field resolution** (#20467): `{module}-1` resolves to `1-1`
- **Inheritance depth 2**: position `{module}/2` resolves to `1/2`, interface `SFP {module}` resolves to `SFP 1/2`
- **Inheritance depth 3**: chained inheritance through 3 levels
- **Custom separator**: `.` separator produces `eth1.1`
- **Multi-token backwards compat**: `Gi{module}/{module}` at matching depth still resolves level-by-level

## Design decisions

- **No object mutation during form validation** — `_get_inherited_positions()` builds a separate list of resolved position strings without modifying module bay model instances
- **Single token = leaf position**: a single `{module}` resolves to the leaf bay's inherited position regardless of depth
- **Multi token = level-by-level**: multiple `{module}` tokens must match tree depth exactly (unchanged behavior)
- **`{vc_position}` integration**: position inheritance works alongside the `{vc_position}` template variable

### DIFF
--- a/docs/models/dcim/moduletype.md
+++ b/docs/models/dcim/moduletype.md
@@ -32,6 +32,26 @@ For example, `{vc_position:1}` will render as `1` when no Virtual Chassis positi
 
 Automatic renaming is supported for all modular component types (those listed above).
 
+### Position Inheritance for Nested Modules
+
+When using nested module bays (modules installed inside other modules), the `{module}` placeholder
+can also be used in the **position** field of module bay templates to inherit the parent bay's
+position. This allows a single module type to produce correctly named components at any nesting
+depth, with a user-controlled separator.
+
+For example, a line card module type might define sub-bay positions as `{module}/1`, `{module}/2`,
+etc. When the line card is installed in a device bay with position `3`, these sub-bay positions
+resolve to `3/1`, `3/2`, etc. An SFP module type with interface template `SFP {module}` installed
+in sub-bay `3/2` then produces interface `SFP 3/2`.
+
+The separator between levels is defined by the user in the position field template itself. Using
+`{module}-1` produces positions like `3-1`, while `{module}.1` produces `3.1`. This provides
+full flexibility without requiring a global separator configuration.
+
+!!! note
+    If the position field does not contain `{module}`, no inheritance occurs and behavior is
+    unchanged from previous versions.
+
 ## Fields
 
 ### Manufacturer
--- a/netbox/dcim/api/serializers_/devices.py
+++ b/netbox/dcim/api/serializers_/devices.py
@@ -8,6 +8,7 @@ from rest_framework import serializers
 from dcim.choices import *
 from dcim.constants import MACADDRESS_ASSIGNMENT_MODELS, MODULE_TOKEN
 from dcim.models import Device, DeviceBay, MACAddress, Module, VirtualDeviceContext
+from dcim.utils import get_module_bay_positions, resolve_module_placeholder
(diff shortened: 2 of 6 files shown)


arthanson:
@claude /review

arthanson (changes requested):
As per other module code, the API also needs to be updated for this - both the validator and actual serializer.  I think these should be common routines if possible that both the UI and API calls as we have pretty common routines in two places right now that need to be kept in sync - actually three places right now (model, form, API).

LDTips:
Very cool PR, looking forward to it as this is the main issue I have with new SFP modelling approach.

Any confirmation whether this approach to having {module} in positions to solve this problem (proposed in this PR), rather than use the proposed `{module_path}` from now closed #21436 will be the one decided to be shipped to next major release (v4.6.0), or there are still on-going design decisions that may impact the way the module inheritance will be handled?

arthanson:
@claude /review

arthanson (approved):

PR #21754, opened 2026-03-25
#20923: Migrate core app to the new UI layouts

jeremystretch:
This migrates all object detail views within the `core` app.


jeremystretch:
@claude /review Does this PR cover all object detail views within the `core` app? Are there any discrepancies or inconsistencies with other apps that have been migrated (`circuits`, `dcim`, etc.)?

jeremystretch:
1. Converted to a single-column layout
2. Addressed in the `inc/panel_table.html` template
3. This is intentional
4. This is intentional

bctiemann (approved):

PR #21756, opened 2026-03-25
21701 Allow scripts to be uploaded via post to API

arthanson:
### Fixes: #21701 

Pre review comments created a separate script-upload endpoint off of Scripts.

I looked at a general script-module API but that has overlap with scripts which can be confusing (if we support get, list, put, delete) which can be confusing, then we have two APIs affecting the same thing, and would arguably need to expose it via GraphQL. Also as it is a proxy model it gets into permission issues as NetBox currently has an issue directly supporting that case.


arthanson:
@claude /review

jeremystretch:
@claude /review

jeremystretch:
@arthanson please address the above items and let me know when this is ready for review.

Also, we should probably include an example showing how to upload a custom script via the REST API to the custom scripts documentation.

arthanson:
@claude /review

arthanson:
claude review:
1. same as in the UI - can add if needed in a separate PR that refactors the checking to be more DRY between API/UI
2. this is incorrect
3. fixed
4. fixed
5. by design
6. that gets messy - would need to temporary ScriptModule in _sync_data_file or a large scale refactor of this and UI code.
7. no idea where it got this.

netbox/extras/api/views.py
```diff
@@ -281,9 +283,43 @@ def initial(self, request, *args, **kwargs):
         super().initial(request, *args, **kwargs)
 
         # Restrict the view's QuerySet to allow only the permitted objects
-        if request.user.is_authenticated:
-            action = 'run' if request.method == 'POST' else 'view'
-            self.queryset = self.queryset.restrict(request.user, action)
+        if request.user.is_authenticated and self.action != 'create':
+            perm_action = 'run' if request.method == 'POST' else 'view'
+            self.queryset = self.queryset.restrict(request.user, perm_action)
```
jeremystretch: `.restrict()` needs to be applied to the queryset regardless of the action.

netbox/netbox/authentication/__init__.py
```diff
@@ -137,12 +138,18 @@ def has_perm(self, user_obj, perm, obj=None):
...
-                permission=perm, model=model
-            ))
+            try:
+                perm_model = apps.get_model(app_label, model_name)
+            except LookupError:
+                perm_model = None
+            if not perm_model or perm_model._meta.concrete_model != model:
+                raise ValueError(_("Invalid permission {permission} for model {model}").format(
+                    permission=perm, model=model
+                ))
```
jeremystretch: I'm not comfortable with the breadth of this change, as it modifies a core part of NetBox's authentication system.

netbox/scripts/__init__.py
jeremystretch, line 1: Why was this file deleted?

jeremystretch (changes requested):
I think there are several issues with this approach that arise from attempting to serve two different models (Script and ScriptModule) via the same REST API endpoint. IMO it would be significantly simpler and safer to introduce a separate endpoint (e.g. `/api/extras/script-modules/`) to exclusively handle script modules. Granted, this is type of change normally reserved for a minor release. What do you think?

netbox/scripts/__init__.py
arthanson: it is not ready for re-review yet, still working on it.

arthanson:
@claude /review

arthanson:
RE claude review:

1. I think script-modules is confusing as it implies full support of the list, get, post.  Doing that opens up a can of worms as the permission system would require a tweak in authentication.py as ScriptModule is a proxy model that the permission doesn't currently handle.  Also IMHO it is confusing as it overlaps Scripts in functionality and greatly increases the scope of this PR.
2. by design
3. it does do permission checks, this is not really valid.
4. Would require refactoring classes in the normal script code, didn't feel it was warranted in this case.
5. That is currently how scripts work

netbox/extras/api/views.py
```diff
@@ -264,10 +263,11 @@ def render(self, request, pk):
 # Scripts
 #
 
-@extend_schema_view(
-    update=extend_schema(request=serializers.ScriptInputSerializer),
-    partial_update=extend_schema(request=serializers.ScriptInputSerializer),
-)
```
jeremystretch, line 270: Why was this removed? This defines the schema for PUT/PATCH to the script endpoint.

netbox/extras/api/urls.py
```diff
@@ -26,6 +26,7 @@
 router.register('config-contexts', views.ConfigContextViewSet)
 router.register('config-context-profiles', views.ConfigContextProfileViewSet)
 router.register('config-templates', views.ConfigTemplateViewSet)
+router.register('script-upload', views.ScriptModuleViewSet)
 router.register('scripts', views.ScriptViewSet, basename='script')
```
jeremystretch: Can we nest the `upload` endpoint under the `scripts/` path?
```suggestion
router.register('scripts/upload', views.ScriptModuleViewSet)
router.register('scripts', views.ScriptViewSet, basename='script')
```

docs/customization/custom-scripts.md
```diff
@@ -384,6 +384,29 @@ A calendar date. Returns a `datetime.date` object.
...
+
+```no-highlight
+curl -X POST \
+-H "Authorization: Token $TOKEN" \
+-H "Accept: application/json; indent=4" \
+-F "upload_file=@/path/to/myscript.py" \
+http://netbox/api/extras/script-upload/
+```
+
+Alternatively, a script module can be linked to an existing data source and data file instead of uploading a file directly:
```
jeremystretch: IMO this is conflating two different functions and will likely lead to confusion. By naming the endpoint some form of the word "upload" implies that the endpoint exists solely to upload custom script files. If we want to support creating modules from data sources via the REST API (and I'm not sure we _do_), a more generic endpoint (e.g. `/api/extras/script-modules/`) would be appropriate.

netbox/extras/api/serializers_/scripts.py
```diff
@@ -1,19 +1,118 @@
...
     'ScriptInputSerializer',
+    'ScriptModuleSerializer',
     'ScriptSerializer',
 )
 
 
+class ScriptModuleSerializer(ValidatedModelSerializer):
+    data_source = DataSourceSerializer(nested=True, required=False, allow_null=True)
+    data_file = DataFileSerializer(nested=True, required=False, allow_null=True)
+    upload_file = serializers.FileField(write_only=True, required=False, allow_null=True)
```
jeremystretch: If we opt to shed the data source behavior, I suggest renaming `upload_file` to `file`.

jeremystretch (changes requested):
I find myself strongly in favor of simply removing the data source functionality, and limiting this to a simple `/scripts/upload/` endpoint that merely allows uploading a file, as that is the minimal scope of the FR.

arthanson:
@claude /review

arthanson:
For claude:

1. fixed
2. not valid as we don't actually return any views data, no get or list
3. just a safety check. in filesystem we rename so this will never get this, in storages there is a bug where we should be doing the rename (separate issue).
4. fixed - logged issue.
5. removed as just a safety check see 3
6. this is the way the UI works, if we want to change separate issue.

jeremystretch (approved):

PR #21759, opened 2026-03-26
fix regression from #14329

arthanson:
Fixes regression for viewing created changelog diffs introduced in #14329, scenario is simple:

1. create an object (like site)
2. view the changelog for the create


arthanson:
@claude /review

pheus (approved):
Thanks for the quick fix!

PR #21761, opened 2026-03-26
Closes #5795: Add Virtual Machine Type model

### LINKED ISSUE #5795: Add Virtual Machine Types
### Environment
* Python version: 3.6.11
* NetBox version: 2.10.4

### Proposed Functionality
Would be handy to be able to track different types of Virtual Machines like Device Types.

Already using Platform to track the Virtual Machine OS, but would also like to track the vm/instance type without having to use custom fields.

### Use Case
Tracking a mix of on-prem devices and AWS EC2 instances, and would like to track the EC2 instance type as can be done for make/model of on-prem devices.

### Database Changes
New Virtual Machine Type model, or an abstract model that can be shared between Devices and Virtual Machines for Types

### External Dependencies
N/A

pheus:
### Fixes: #5795

This PR introduces `VirtualMachineType` to classify virtual machines and provide default values for platform, vCPUs, and memory when a VM is created.

We chose the **classification + create-time defaults** approach here because it adds practical value beyond metadata alone, while keeping the scope well-defined. It solves the immediate need to model VM types natively in NetBox and avoids overloading `Platform`, without jumping straight into full `DeviceType`-style templating for interfaces and disks.

Defaults are only applied at creation time, explicit values on the VM always take precedence, and updating a `VirtualMachineType` does not retroactively change existing virtual machines.

The implementation includes the new model and its relationship to `VirtualMachine`, along with support across forms, filtersets, tables, views, the REST API, GraphQL, navigation, search, documentation, and tests.

The design is also intended to leave the door open for a more template-oriented approach in the future if there is a stronger need to model `VMInterface` and `VirtualDisk` templates as well.


pheus:
@claude /review

pheus:
@claude /review

1. Bug: Duplicate GraphQL field: This is an unrelated bug, since this PR adds `VirtualMachineTypeType`.
2. Design: on_delete=PROTECT on the VM → type FK: From my experience, I expect that unintentional data deletion should be prevented. Bulk edit VMs is easy to do.
3. UX: Type defaults are silent — not surfaced in the form: That's an UX question. Keeping it empty means the user didn't specify anything specific, whereas prefilling leads to confusion.
4. Minor: apply_type_defaults uses FK accessor instead of _id: That is fixed in the newest push.
5. Minor: No select_related on VM list queryset for type: This is handled by NetBox's BaseTable.
6. Minor: Missing default_disk field: This is intentional, since NetBox has moved this its own model. I don't want to keep the current legacy field alive.

netbox/netbox/navigation/menu.py
```diff
@@ -270,6 +270,12 @@
                 get_model_item('virtualization', 'virtualdisk', _('Virtual Disks')),
             ),
         ),
+        MenuGroup(
```
jeremystretch: IMO we can just put this under the existing "Virtual Machines" group. (Unlike device types, there aren't any other relevant links to justify a separate menu group.)

netbox/virtualization/graphql/filters.py
```diff
@@ -129,10 +147,11 @@ class VirtualMachineFilter(
...
     start_on_boot: (
-        BaseFilterLookup[Annotated['VirtualMachineStartOnBootEnum', strawberry.lazy('virtualization.graphql.enums')]
-    ] | None) = (
-        strawberry_django.filter_field()
-    )
+        BaseFilterLookup[Annotated['VirtualMachineStartOnBootEnum', strawberry.lazy('virtualization.graphql.enums')]]
+        | None
+    ) = strawberry_django.filter_field()
+    interface_count: ComparisonFilterLookup[int] | None = strawberry_django.filter_field()
+    virtual_disk_count: ComparisonFilterLookup[int] | None = strawberry_django.filter_field()
```
jeremystretch: We should probably back these out and capture them separately under a bug report.

mkdocs.yml
```diff
@@ -285,9 +285,10 @@ nav:
             - Cluster: 'models/virtualization/cluster.md'
             - ClusterGroup: 'models/virtualization/clustergroup.md'
             - ClusterType: 'models/virtualization/clustertype.md'
-            - VMInterface: 'models/virtualization/vminterface.md'
             - VirtualDisk: 'models/virtualization/virtualdisk.md'
             - VirtualMachine: 'models/virtualization/virtualmachine.md'
+            - VirtualMachineType: 'models/virtualization/virtualmachinetype.md'
+            - VMInterface: 'models/virtualization/vminterface.md'
```
jeremystretch: Minor nit, but VMInterface should be listed above VirtualDisk because we treat uppercase letters as preceding lowercase letters. (For example, ASNRange comes before Aggregate under IPAM.)

netbox/virtualization/models/virtualmachines.py
```diff
@@ -282,8 +364,29 @@ def save(self, *args, **kwargs):
             elif self.device and self.device.site:
                 self.site = self.device.site
 
+        if self._state.adding:
+            self.apply_type_defaults()
```
jeremystretch, line 368: I think I agree with Claude's initial assessment that we should avoid assigning these values in a manner that's opaque to the user. Instead, we could modify the edit form to automatically re-render with the default platform, vCPU, and memory values pre-populated after selecting a VM type. (ModuleTypeForm under `dcim` is a good example of this approach.)

netbox/virtualization/tests/test_api.py
```diff
@@ -224,26 +340,106 @@ def setUpTestData(cls):
...
             {
                 'name': 'Virtual Machine 7',
-                'cluster': clusters[2].pk,
+                'cluster': cls.clusters[2].pk,
+                'virtual_machine_type': cls.vm_types[0].pk,
                 'start_on_boot': VirtualMachineStartOnBootChoices.STATUS_ON,
             },
         ]
 
+    def test_filter_by_virtual_machine_type(self):
```
jeremystretch: This should not be needed as we have a separate filterset test.

netbox/virtualization/tests/test_models.py
```diff
@@ -1,19 +1,188 @@
...
+        Verify that the string representation of a VirtualMachineType returns
+        its name, and that all default fields (platform, vcpus, memory) are
+        stored correctly after creation.
+        """
+        self.assertEqual(str(self.virtual_machine_type), 'Small Linux')
+        self.assertEqual(self.virtual_machine_type.default_platform, self.platform)
+        self.assertEqual(self.virtual_machine_type.default_vcpus, Decimal('2.00'))
+        self.assertEqual(self.virtual_machine_type.default_memory, 4096)
+
+    def test_virtual_machine_type_duplicate_name_case_insensitive(self):
```
jeremystretch: IMO we don't need individual tests for unique constraints. That seems excessive.

netbox/virtualization/tests/test_models.py
```diff
@@ -1,19 +1,188 @@
...
+        """
+        virtual_machine_type = VirtualMachineType(
+            name='SMALL LINUX',
+            slug='small-linux-2',
+        )
+
+        with self.assertRaises(ValidationError):
+            virtual_machine_type.full_clean()
+
+    def test_virtual_machine_type_duplicate_slug(self):
```
jeremystretch: Same as above.

netbox/virtualization/tests/test_models.py
```diff
@@ -1,19 +1,188 @@
...
+        self.virtual_machine_type.refresh_from_db()
+        self.assertEqual(self.virtual_machine_type.virtual_machine_count, 1)
+
+        # Reassign the remaining VM to no type — count should drop to zero
+        vm2.virtual_machine_type = None
+        vm2.save()
+        self.virtual_machine_type.refresh_from_db()
+        self.assertEqual(self.virtual_machine_type.virtual_machine_count, 0)
+
+    def test_virtual_machine_type_deletion_protected(self):
```
jeremystretch: Same as above; deletion protection is a core Django feature.

netbox/virtualization/tests/test_models.py
```diff
@@ -1,19 +1,188 @@
...
+            cluster=self.cluster,
+            virtual_machine_type=self.virtual_machine_type,
+        )
+
+        from django.db import models
+
+        with self.assertRaises(models.ProtectedError):
+            self.virtual_machine_type.delete()
+
+    def test_virtual_machine_type_deletion_without_vms(self):
```
jeremystretch: Same as above

netbox/virtualization/tests/test_models.py
```diff
@@ -1,19 +1,188 @@
...
+        """
+        vmt = VirtualMachineType(
+            name='Zero vCPU Type',
+            slug='zero-vcpu-type',
+            default_vcpus=Decimal('0.00'),
+        )
+        with self.assertRaises(ValidationError):
+            vmt.full_clean()
+
+    def test_virtual_machine_type_minimal_fields(self):
```
jeremystretch: Not needed

jeremystretch (changes requested):

netbox/virtualization/graphql/filters.py
pheus: These changes are fairly small: I only moved them to the end and replaced `FilterLookup[int]` with `ComparisonFilterLookup[int]` to align with `virtual_machine_count` on `VirtualMachineTypeFilter`.

Because of that, my preference would be to keep them here, but I’m also fine splitting them out into a separate bug report if you’d prefer.

netbox/netbox/navigation/menu.py
pheus: I went back and forth on this a few times while implementing it. I ended up giving it its own group because the existing **Virtual Machines** group currently contains objects attached to a VM, whereas `VirtualMachineType` sits one level above `VirtualMachine`.

That said, I don’t feel strongly about it, and I’m happy to move it under the existing **Virtual Machines** group if that feels more consistent.

netbox/virtualization/models/virtualmachines.py
pheus: Thanks, that makes sense. I agree the web UI should make these defaults visible rather than applying them in a way that feels surprising.

I’ve updated the create form to re-render when the VM type changes and to pre-populate `platform`, `vcpus`, and `memory` there, following the same general UX pattern as other dynamic forms in NetBox. I’ve intentionally left the edit form alone so that changing the type on an existing VM does not rewrite saved values.

I did keep the model-level defaulting in place for non-UI workflows, though. If a `VirtualMachineType` is assigned via the REST API, `nbshell`, or a script, I think it’s still important that omitted values are populated consistently there as well.

netbox/virtualization/graphql/filters.py
jeremystretch: They're small changes, but they correct an incorrect behavior in the current release, so we need to capture and document them in the release notes. They should also be fixed in the next available v4.5 release as opposed to v4.6.0.

pheus: I've removed the unrelated changes. Thanks!

netbox/virtualization/tests/test_api.py
jeremystretch: Can we delete this test?

jeremystretch (changes requested):

netbox/virtualization/tests/test_api.py
pheus: I'm sorry. I was quite sure that I've deleted all of the irrelevant tests.

jeremystretch (approved):

PR #21769, opened 2026-03-27
Closes #20163: Add changelog message support to bulk component creation

### LINKED ISSUE #20163: Changelog message support for bulk device component creation
### NetBox version

v4.4.0-beta1

### Feature type

New functionality

### Proposed functionality

When adding components to multiple devices in bulk, include a form field to record a changelog message, as exists in similar forms.

### Use case

This will establish parity with the workflow where components are added to a single device.

### Database changes

N/A

### External dependencies

N/A

pheus:
### Fixes: #20163

This PR adds changelog message support to bulk device-component creation so it behaves consistently with the single-device workflow. It includes `ChangelogMessageMixin` in `DeviceBulkAddComponentForm`, captures `changelog_message` during bulk creation, and applies it to each created component instance.

A regression test is also included to verify changelog message propagation during bulk creation.

Thanks for taking the time to review this change.

### DIFF
--- a/netbox/dcim/forms/bulk_create.py
+++ b/netbox/dcim/forms/bulk_create.py
@@ -3,7 +3,7 @@ from django.utils.translation import gettext_lazy as _
 
 from dcim.models import *
 from extras.models import Tag
-from netbox.forms.mixins import CustomFieldsMixin
+from netbox.forms.mixins import ChangelogMessageMixin, CustomFieldsMixin
 from utilities.forms import form_from_model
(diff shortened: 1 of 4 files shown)


pheus:
@claude /review

arthanson (changes requested):
I think the styling needs to change to be consistent.  When you do the bulk creation here the Changelog message looks like all the other form fields, but in all the other places it is separate with a greenish background:

Image
Image

arthanson (approved):

PR #21787, opened 2026-03-30
Fixes #21786: Revert failed fix for #21747

jeremystretch:
### Fixes: #21786

Reverts the change to the search backend caching behavior made under #21747


pheus (approved):
Thanks!

PR #21789, opened 2026-03-31
Fixes #21095: Add IEC unit labels support and rename humanize helpers to be unit-agnostic

### LINKED ISSUE #21095: Virtual Machine Resources Use Decimal MB/GB Instead of Binary MiB/GiB, Causing Memory Mismatch

jsenecal:
## Summary

  - `_humanize_capacity()` (renamed from `_humanize_megabytes()`) now emits IEC labels
    (MiB, GiB, TiB, PiB) when the divisor is 1024, and SI labels (MB, GB, TB, PB) when 1000
  - Renamed `humanize_disk_megabytes` / `humanize_ram_megabytes` template filters to
    `humanize_disk_capacity` / `humanize_ram_capacity` to reflect that stored values
    are unit-agnostic, not strictly megabytes
  - Form labels in bulk edit and filter forms dynamically reflect the configured
    `RAM_BASE_UNIT` / `DISK_BASE_UNIT` setting via `get_capacity_unit_label()`
  - VM table memory column now humanized via `render_memory()` (disk already was)

  No conversion logic is needed on save. The database stores a raw integer whose
  interpretation as MB or MiB is determined by the `RAM_BASE_UNIT` / `DISK_BASE_UNIT`
  setting. The same setting controls both input labels and display humanization,
  keeping the roundtrip consistent.

  ## Tests

  - [x] Unit tests for `_humanize_capacity` with both SI (1000) and IEC (1024) divisors
  - [x] Unit tests for `get_capacity_unit_label` helper
  - [x] Full virtualization test suite passes (410 tests)
  - [x] Django system check and migration check pass

### DIFF
--- a/netbox/templates/virtualization/panels/cluster_resources.html
+++ b/netbox/templates/virtualization/panels/cluster_resources.html
@@ -12,7 +12,7 @@
       <th scope="row"><i class="mdi mdi-chip"></i> {% trans "Memory" %}</th>
       <td>
         {% if memory_sum %}
-          <span title={{ memory_sum }}>{{ memory_sum|humanize_ram_megabytes }}</span>
+          <span title={{ memory_sum }}>{{ memory_sum|humanize_ram_capacity }}</span>
         {% else %}
           {{ ''|placeholder }}
         {% endif %}
@@ -24,7 +24,7 @@
       </th>
       <td>
         {% if disk_sum %}
-          {{ disk_sum|humanize_disk_megabytes }}
+          {{ disk_sum|humanize_disk_capacity }}
         {% else %}
           {{ ''|placeholder }}
         {% endif %}
--- a/netbox/templates/virtualization/panels/virtual_machine_resources.html
+++ b/netbox/templates/virtualization/panels/virtual_machine_resources.html
@@ -12,7 +12,7 @@
       <th scope="row"><i class="mdi mdi-chip"></i> {% trans "Memory" %}</th>
       <td>
         {% if object.memory %}
-          <span title={{ object.memory }}>{{ object.memory|humanize_ram_megabytes }}</span>
+          <span title={{ object.memory }}>{{ object.memory|humanize_ram_capacity }}</span>
         {% else %}
           {{ ''|placeholder }}
         {% endif %}
@@ -24,7 +24,7 @@
       </th>
       <td>
         {% if object.disk %}
-          {{ object.disk|humanize_disk_megabytes }}
+          {{ object.disk|humanize_disk_capacity }}
         {% else %}
           {{ ''|placeholder }}
         {% endif %}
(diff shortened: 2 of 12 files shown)


pheus:
@claude /review

pheus:
Thanks for opening this PR!

jsenecal:
Thanks for the review, I somehow managed to forget about the model forms, thanks Claude :D

1. I'll make the model `verbose_name` fields unit-agnostic (drop the (MB) suffix) and add `__init__` label overrides to VirtualMachineForm and VirtualDiskForm to dynamically display the correct unit. ~~This will generate a migration since Django tracks `verbose_name` changes.~~ Apparently, Django doesn't consider `verbose_name` changes as migration-worthy. No migration needed.

2. get_capacity_unit_label isn't a template tag and doesn't actually belong there, all things considered. See (3)

3. `get_capacity_unit_label` is only used by forms, so it belongs with form helpers. I'll move it to utilities/forms/utils.py alongside `add_blank_choice`, `get_field_value`, etc. 
`_humanize_capacity` stays in templatetags/helpers.py since it powers the template filters.

jsenecal:
> Template title attribute missing quotes

Want me to address this ? Pretty off-scope however.

pheus:
Thanks for the corrections and for the additional context.

> Want me to address this? Pretty off-scope however.

I’d agree that this seems a bit out of scope for the PR, but I’ll leave that to the reviewer to decide.

arthanson:
> > Template title attribute missing quotes
> 
> Want me to address this ? Pretty off-scope however.

no - I think it would be a separate housekeeping issue if we wanted to do that

arthanson (approved):
Thanks @jsenecal !

PR #21804, opened 2026-03-31
Fixes #21709: Use `@register_search` in plugin search docs

pheus:
### Fixes: #21709

Align the plugin search example with the recommended registration pattern used in the general search documentation and NetBox core.

Replace the legacy `indexes = [...]` example with decorator-based registration to make the preferred approach clearer for plugin authors.


jeremystretch (approved):
Thanks @pheus!

PR #21805, opened 2026-03-31
Fixes: #21696 - Upgrade to django-rq==4.0.1

bctiemann:
### Fixes: #21696

This PR upgrades the pinned version of django-rq to 4.0.1. It makes a few changes to queue map lookup behavior and autocommit defaults (as in recommended usage of the current API) in order to ensure unit tests pass correctly.

### DIFF
--- a/netbox/core/api/views.py
+++ b/netbox/core/api/views.py
@@ -2,7 +2,7 @@ from django.http import Http404, HttpResponse
 from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext_lazy as _
 from django_rq.queues import get_redis_connection
-from django_rq.settings import QUEUES_LIST
+from django_rq.settings import get_queues_list
 from django_rq.utils import get_statistics
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter, extend_schema
@@ -195,7 +195,7 @@ class BackgroundWorkerViewSet(BaseRQViewSet):
         return 'Background Workers'
 
     def get_data(self):
-        config = QUEUES_LIST[0]
+        config = get_queues_list()[0]
         return Worker.all(get_redis_connection(config['connection_config']))
 
     @extend_schema(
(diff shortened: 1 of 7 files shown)


bctiemann:
@claude /review

bctiemann:
From local testing prior to PR creation:

❯ Does the COMMIT_MODE fix need to be applied to the regular configuration (and/or added to configuration_example.py)? Or is this behavior only relevant to tests?     
                                                                                                                                                                       
⏺ The on_db_commit default is actually correct and better for production. It prevents jobs from being enqueued if the triggering DB transaction rolls back (e.g., due  
  to an exception mid-request). This is a deliberate improvement in django-rq 4.0.                                                                                     
                                                                                                                                                                       
  The auto mode is only needed for tests because Django's TestCase wraps each test in a transaction that is never committed — it just gets rolled back at teardown.    
  That's the special case that breaks things.                                                                                                                          
                                                                                                                                                                       
  Let me quickly verify there are no NetBox code paths that enqueue jobs outside of a request/DB-transaction context where on_db_commit might swallow jobs silently:   
   
  Searched for 2 patterns (ctrl+o to expand)                                                                                                                           
                  
⏺ All production enqueue sites are either within request handlers (which run in DB transactions that commit normally) or management commands (which have no active     
  transaction, in which case on_db_commit fires immediately per Django's on_commit() semantics).
                                                                                                                                                                       
  Answer: No changes needed to production configuration or configuration_example.py.                                                                                   
                                                                                                                                                                       
  COMMIT_MODE: 'on_db_commit' is the right default for production — it's actually safer, since jobs won't be dispatched if the triggering transaction rolls back. The  
  auto override is only needed in tests because Django's TestCase uses a permanent, never-committed wrapping transaction.

arthanson (changes requested):
QUEUES_LIST is also deprecated (https://github.com/rq/django-rq/blob/master/django_rq/settings.py#L49) should probably move to get_queueus_list()

bctiemann:
Good call, thanks. Now they're both in line.

arthanson (approved):

PR #21806, opened 2026-03-31
Closes #21771: Add `add_tags` & `remove_tags` fields for taggable objects

jeremystretch:
### Closes: #21771

- Add write-only `add_tags` and `remove_tags` fields to every REST API serializer which supports tag assignment
- Add API tests


jeremystretch:
@claude /review Check that this satisfies the requirements of FR #21771 and note any potential issues with the implementation. Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration.

jeremystretch:
1. We now raise a validation error if `remove_tags` is set when creating an object.
2. This appears to be harmless AFAICT. It should not impact change logging.
3. We now raise a validation error if the same tag is present in both `add_tags` and `remove_tags`
4. Appears to be safe

bctiemann (approved):

PR #21811, opened 2026-04-01
Fixes #21808: Prevent errors when viewing Virtual Interfaces without a Virtual Circuit Termination

pheus:
### Fixes: #21808

This fixes a regression in the virtual interface detail view introduced during the UI layout migration.

`VirtualCircuitPanel` accessed `object.virtual_circuit_termination` directly, which can raise `RelatedObjectDoesNotExist` for virtual interfaces without a related termination. This change handles that case gracefully and skips rendering the panel when no termination exists, restoring the previous behavior and preventing a server error when opening the interface detail page.

### DIFF
--- a/netbox/dcim/ui/panels.py
+++ b/netbox/dcim/ui/panels.py
@@ -532,7 +532,7 @@ class VirtualCircuitPanel(panels.ObjectPanel):
 
     def render(self, context):
         obj = context.get('object')
-        if not obj or not obj.is_virtual or not obj.virtual_circuit_termination:
+        if not obj or not obj.is_virtual or not hasattr(obj, 'virtual_circuit_termination'):
             return ''
         ctx = self.get_context(context)
         return render_to_string(self.template_name, ctx, request=ctx.get('request'))


pheus:

pheus:

netbox/dcim/ui/panels.py
```diff
@@ -532,8 +533,14 @@ class VirtualCircuitPanel(panels.ObjectPanel):
...
         obj = context.get('object')
-        if not obj or not obj.is_virtual or not obj.virtual_circuit_termination:
+        if not obj or not obj.is_virtual:
             return ''
+
+        try:
+            obj.virtual_circuit_termination
+        except ObjectDoesNotExist:
+            return ''
+
```
jnovinger: One thought: the model itself uses `hasattr(self, 'virtual_circuit_termination')` for this same check at `dcim/models/device_components.py:1100`. Using the same pattern here would keep the guard clause structure consistent with the rest of the file:

```python
if not obj or not obj.is_virtual or not hasattr(obj, 'virtual_circuit_termination'):
    return ''
```

Not blocking, just a consistency nit. Either way this looks good.

jnovinger (dismissed):

jnovinger (approved):

PR #21812, opened 2026-04-01
Fixes #21810: Prevent errors when viewing Device pages for Virtual Chassis Members

pheus:
### Fixes: #21810

This fixes a regression in the device detail view introduced during the UI layout migration.

`VirtualChassisMembersPanel` was building its add-device action from `ctx['object']`, but on the device view `object` is a `Device`, not a `VirtualChassis`. This change uses the panel's `virtual_chassis` context instead and checks for `master_id` before populating the default site and rack parameters, preventing an `AttributeError` when viewing a virtual chassis member that is not the master.


pheus:

pheus:

jeremystretch (approved):

PR #21815, opened 2026-04-01
Fixes #21498: Fix Exception when changing a Cable Termination with an Interface Event Rule

### LINKED ISSUE #21498: Eventrules in v4.5.2 and v4.5.3 result in server error
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.3

### Python Version

3.12

### Steps to Reproduce

1. Create an event rule von Interfaces triggering on Create/Update and calling an arbitrary custom script or notification.
2. Create a cable connection from Device A Ethernet1 to Ethernet2, let call that cable "Cable ID 3374"
3. Edit the just created "Cable ID 3374" and alter the termination from Ethernet2 to Ethernet3.
4. Server Error Message

### Expected Behavior

The cabling changing as intended without any server errors.

### Observed Behavior

Here is a screenshot of said error:

Image

Despite the error, changes on the cable will result in proper new cabling as intended, with "Cable ID 3374" now connected from Device A Ethernet1 to Device A Ethernet3.

Disabling the event rule and you will not run into the error shown.

So far I was able to reproduce this behavior in:

- my lab environment using NetBox v4.5.3 and v4.5.2.
- the current netbox demo 
  - Python version: 3.12.3
  - NetBox version: 4.5.3

Everything works as intended in v4.5.1 in my lab. So I think something related to the v4.5.2 changes seem to be the issue.

And logs from /var/log/netbox/netbox.log

`Internal Server Error: /dcim/cables/3374/edit/
  File "/opt/netbox/venv/lib/python3.12/site-packages/django/db/models/fields/related_descriptors.py", line 243, in __get__
    rel_obj = self.field.get_cached_value(instance)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/django/db/models/fields/mixins.py", line 37, in get_cached_value
    return instance._state.fields_cache[self.cache_name]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
KeyError: '_path'

  File "/opt/netbox/venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/netbox/middleware.py", line 40, in __call__
    with apply_request_processors(request):
  File "/usr/lib/python3.12/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/opt/netbox/netbox/utilities/request.py", line 109, in apply_request_processors
    with ExitStack() as stack:
  File "/usr/lib/python3.12/contextlib.py", line 610, in __exit__
    raise exc_details[1]
  File "/usr/lib/python3.12/contextlib.py", line 595, in __exit__
    if cb(*exc_details):
       ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/opt/netbox/netbox/netbox/context_managers.py", line 26, in event_tracking
    flush_events(events)
  File "/opt/netbox/netbox/extras/events.py", line 245, in flush_events
    func(events)
  File "/opt/netbox/netbox/extras/events.py", line 230, in process_event_queue
    process_event_rules(
  File "/opt/netbox/netbox/extras/events.py", line 132, in process_event_rules
    if not event_rule.eval_conditions(event['data']):
                                      ~~~~~^^^^^^^^
  File "/opt/netbox/netbox/extras/events.py", line 35, in __getitem__
    data = serialize_for_event(self['object'])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/extras/events.py", line 50, in serialize_for_event
    return serializer.data
           ^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/rest_framework/serializers.py", line 573, in data
    ret = super().data
          ^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/rest_framework/serializers.py", line 251, in data
    self._data = self.to_representation(self.instance)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/rest_framework/serializers.py", line 540, in to_representation
    ret[field.field_name] = field.to_representation(attribute)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/rest_framework/fields.py", line 1870, in to_representation
    return method(value)
           ^^^^^^^^^^^^^
  File "/opt/netbox/netbox/dcim/api/serializers_/base.py", line 32, in get_connected_endpoints
    if endpoints := obj.connected_endpoints:
                    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/functools.py", line 995, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/dcim/models/device_components.py", line 1102, in connected_endpoints
    return super().connected_endpoints
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/functools.py", line 995, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/dcim/models/device_components.py", line 358, in connected_endpoints
    return self._path.destinations if self._path else []
                                      ^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/django/db/models/fields/related_descriptors.py", line 261, in __get__
    rel_obj = self.get_object(instance)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/django/db/models/fields/related_descriptors.py", line 224, in get_object
    return qs.get(self.field.get_reverse_related_filter(instance))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/django/db/models/query.py", line 635, in get
    raise self.model.DoesNotExist(
dcim.models.cables.CablePath.DoesNotExist: CablePath matching query does not exist.`

pheus:
### Fixes: #21498

This PR fixes a regression in deferred event serialization during cable edits.

When a cable termination is changed, `CablePath` rows can be deleted and recreated while endpoint instances are still in memory. If an event rule is active, the deferred serialization introduced in v4.5.2 can then hit a stale `_path` reference and raise `CablePath.DoesNotExist`.

The DCIM-side fix makes `PathEndpoint.path` recover from a stale `_path` reference with a narrow refresh of that field, and routes the remaining internal callers through `path` instead of accessing `_path` directly. This prevents the crash and keeps the fix scoped to the denormalized relation involved in path rebuilds.

This PR also includes the `EventContext` changes intentionally. While the `PathEndpoint` fix prevents the exception, it does not by itself address stale event payloads when the same object is enqueued multiple times within one request. `EventContext` now tracks the latest serialization source for coalesced duplicate enqueues, so deferred serialization uses the most recent queued instance while keeping the public `object` entry unchanged. Delete events are still serialized eagerly before the row becomes unavailable.

Finally, `process_event_rules()` now avoids mutating `event_rule.action_data` in place when merging event payload data.

Together, these changes fix the traceback from #21498, preserve the deferred-serialization optimization, and ensure event consumers receive the final queued state for coalesced updates.

### DIFF
--- a/netbox/dcim/models/device_components.py
+++ b/netbox/dcim/models/device_components.py
@@ -2,7 +2,7 @@ from functools import cached_property
 
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.postgres.fields import ArrayField
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
 from django.db.models import Sum
@@ -307,11 +307,12 @@ class PathEndpoint(models.Model):
 
     `connected_endpoints()` is a convenience method for returning the destination of the associated CablePath, if any.
     """
+
     _path = models.ForeignKey(
         to='dcim.CablePath',
         on_delete=models.SET_NULL,
         null=True,
-        blank=True
+        blank=True,
     )
 
     class Meta:
@@ -323,11 +324,14 @@ class PathEndpoint(models.Model):
 
         # Construct the complete path (including e.g. bridged interfaces)
         while origin is not None:
-
-            if origin._path is None:
+            # Go through the public accessor rather than dereferencing `_path`
+            # directly. During cable edits, CablePath rows can be deleted and
+            # recreated while this endpoint instance is still in memory.
+            cable_path = origin.path
+            if cable_path is None:
                 break
 
-            path.extend(origin._path.path_objects)
+            path.extend(cable_path.path_objects)
 
             # If the path ends at a non-connected pass-through port, pad out the link and far-end terminations
             if len(path) % 3 == 1:
@@ -336,8 +340,8 @@ class PathEndpoint(models.Model):
             elif len(path) % 3 == 2:
                 path.insert(-1, [])
 
-            # Check for a bridged relationship to continue the trace
-            destinations = origin._path.destinations
+            # Check for a bridged relationship to continue the trace.
+            destinations = cable_path.destinations
             if len(destinations) == 1:
                 origin = getattr(destinations[0], 'bridge', None)
             else:
@@ -348,14 +352,42 @@ class PathEndpoint(models.Model):
 
     @property
     def path(self):
-        return self._path
+        """
+        Return this endpoint's current CablePath, if any.
+
+        `_path` is a denormalized reference that is updated from CablePath
+        save/delete handlers, including queryset.update() calls on origin
+        endpoints. That means an already-instantiated endpoint can briefly hold
+        a stale in-memory `_path` relation while the database already points to
+        a different CablePath (or to no path at all).
+
+        If the cached relation points to a CablePath that has just been
+        deleted, refresh only the `_path` field from the database and retry.
+        This keeps the fix cheap and narrowly scoped to the denormalized FK.
+        """
+        if self._path_id is None:
+            return None
+
+        try:
+            return self._path
+        except ObjectDoesNotExist:
+            # Refresh only the denormalized FK instead of the whole model.
+            # The expected problem here is in-memory staleness during path
+            # rebuilds, not persistent database corruption.
+            self.refresh_from_db(fields=['_path'])
+            return self._path if self._path_id else None
 
     @cached_property
     def connected_endpoints(self):
         """
-        Caching accessor for the attached CablePath's destination (if any)
+        Caching accessor for the attached CablePath's destinations (if any).
+
+        Always route through `path` so stale in-memory `_path` references are
+        repaired before we cache the result for the lifetime of this instance.
         """
-        return self._path.destinations if self._path else []
+        if cable_path := self.path:
+            return cable_path.destinations
+        return []
 
 
 #
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -25,16 +25,54 @@ logger = logging.getLogger('netbox.events_processor')
 
 class EventContext(UserDict):
     """
-    A custom dictionary that automatically serializes its associated object on demand.
+    Dictionary-compatible wrapper for queued events that lazily serializes
+    ``event['data']`` on first access.
+
+    Backward-compatible with the plain-dict interface expected by existing
+    EVENTS_PIPELINE consumers. When the same object is enqueued more than once
+    in a single request, the serialization source is updated so consumers see
+    the latest state.
     """
 
-    # We're emulating a dictionary here (rather than using a custom class) because prior to NetBox v4.5.2, events were
-    # queued as dictionaries for processing by handles in EVENTS_PIPELINE. We need to avoid introducing any breaking
-    # changes until a suitable minor release.
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Track which model instance should be serialized if/when `data` is
+        # requested. This may be refreshed on duplicate enqueue, while leaving
+        # the public `object` entry untouched for compatibility.
+        self._serialization_source = None
+        if 'object' in self:
+            self._serialization_source = super().__getitem__('object')
+
+    def refresh_serialization_source(self, instance):
+        """
+        Point lazy serialization at a fresher instance, invalidating any
+        already-materialized ``data``.
+        """
+        self._serialization_source = instance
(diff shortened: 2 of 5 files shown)


pheus:

pheus:
Thanks for the review.

**1. `@skipIf` on `test_send_webhook`:** Intentional, `test_send_webhook` already depends on the dummy plugin implicitly; the decorator makes that explicit. Happy to defer to maintainers on whether to keep it in this PR.

**2. Unhandled `ObjectDoesNotExist` after refresh:** The described scenario would require two consecutive path deletions targeting the same endpoint between `refresh_from_db` and the FK access. The actual rebuild sequence (delete -> create -> update FK) means the refreshed `_path_id` is either valid or `NULL`. A second try/except would obscure intent without guarding against a realistic failure.

**3. Docstring:** Fair nit. The inline comment explaining *why* `path` is used instead of `_path` is the valuable part. Leaving as-is for maintainer preference.

netbox/dcim/models/device_components.py
```diff
@@ -323,39 +327,74 @@ def trace(self):
...
+            # recreated while this endpoint instance is still in memory.
+            cable_path = origin.path
+            if cable_path is None:
                 break
 
-            path.extend(origin._path.path_objects)
+            path.extend(cable_path.path_objects)
 
-            # If the path ends at a non-connected pass-through port, pad out the link and far-end terminations
+            # If the path ends at a non-connected pass-through port, pad out
```
arthanson: Is your line length set short - these can still be just be on same line. a lot of these changes are just splitting the comments for no discernible reason.

netbox/dcim/models/device_components.py
```diff
@@ -323,39 +327,74 @@ def trace(self):
...
+        save/delete handlers, including queryset.update() calls on origin
+        endpoints. That means an already-instantiated endpoint can briefly hold
+        a stale in-memory `_path` relation while the database already points to
+        a different CablePath (or to no path at all).
+
+        If the cached relation points to a CablePath that has just been
+        deleted, refresh only the `_path` field from the database and retry.
+        This keeps the fix cheap and narrowly scoped to the denormalized FK.
+        """
+        if self._path_id is None:
```
arthanson, line 368: I don't think this check is needed, the try below should return None if _path_id is None

netbox/extras/events.py
```diff
@@ -25,22 +25,59 @@
...
-    # We're emulating a dictionary here (rather than using a custom class) because prior to NetBox v4.5.2, events were
-    # queued as dictionaries for processing by handles in EVENTS_PIPELINE. We need to avoid introducing any breaking
-    # changes until a suitable minor release.
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Track which model instance should be serialized if/when `data` is
+        # requested. This may be refreshed on duplicate enqueue, while leaving
+        # the public `object` entry untouched for compatibility.
+        self._serialization_source = self.get('object')
```
arthanson: may want to use super().__getitem__('object') to be safe, I think self.get( will call down int __getitem.

arthanson (changes requested):
great work @pheus just some minor comments, - can we not reformat the comments for just line length changes unless there is a specific reason? quite a few of the changes are just reformatting comments without any wording changes.

netbox/dcim/models/device_components.py
pheus: I’ll drop the comment-only/docstring cleanup to keep this PR tighter.

netbox/dcim/models/device_components.py
pheus: For `PathEndpoint.path`, I’d like to keep the early `_path_id is None` branch as a small fast-path/readability check, but I agree the important part is the second guarded lookup after `refresh_from_db()`.

arthanson (approved):

PR #21817, opened 2026-04-01
Fix AttributeError when an AnonymousUser tries to sort a table

fabi125:
### Fixes: #21784

Check if a user is logged in before trying to save table sorting preferences.

Also added a test case for this.


pheus:

pheus (approved):
Thanks, @fabi125 ! This looks good to me.

PR #21826, opened 2026-04-02
Fixes #21825: Fix sorting for multiple incorrectly configured table columns

pheus:
### Fixes: #21825

This PR fixes several broken sortable columns caused by incorrect accessors, invalid field references, and columns being marked sortable when sorting is not actually supported.

It updates the affected table definitions by correcting accessor paths and field names, removing the deprecated `name` field from `VLANTranslationRule`, and marking computed or template-only columns as non-orderable where needed.

Related test coverage is being handled as part of #21766, so this PR focuses on the fixes themselves.


pheus:

pheus:
@claude /review
I've your first point.

jeremystretch (approved):
Thanks for taking the time to clean all this up @pheus!

PR #21827, opened 2026-04-02
Closes #20924: Ready UI components for use by plugins

### LINKED ISSUE #20924: Support the use of modular template components in plugins
### NetBox version

v4.4.7

### Feature type

New functionality

### Proposed functionality

NetBox v4.5 introduced modular template components under FR #20204, which we've begun to implement internally. Once that work has been completed and the set of components has been fully fleshed out, we will include them as part of the plugins API.

Note that this change will not preclude plugin authors from continuing to utilize custom HTML templates; it merely makes the components available as part of the plugins API.

### Use case

Supporting the use of these components within plugins will greatly simplify the process of designing plugin UIs and help ensure that UX remains consistent with core NetBox.

### Database changes

_No response_

### External dependencies

_No response_

jeremystretch:
### Closes: #20924

The goal of this PR is to ready the various UI components initially implemented in NetBox v4.5 for use by plugins in NetBox v4.6 by ensuring they are robust and reasonably complete. Although this PR is fairly broad, all changes herein are driven by this common goal.

- Enable iterating directly against layouts/rows/columns to access child objects
- Support optional column grid width assignment
- Introduce the `should_render()` method on Panel to simplify conditional rendering
- Fix potential issues with resolving models before the application has fully initialized
- Standardize the `get_config()` and `render()` signatures for Panel subclasses
- Replace bare instances of the base Panel class with model-specific subclasses
- Add missing PluginContentPanel inclusions
- Cleanly handle exceptions that arise when rendering plugin content embedded in core views
- Update the plugins UI documentation
- Complete object attribute tests


jeremystretch:
@claude /review Audit all the changes in this PR and note any inconsistencies or potential bugs. Also check for any breaking changes as compared to the current `main` branch. Don't summarize all the changes in your response: Just note items that might need further attention.

jeremystretch:
#### Breaking Changes

- Removing the `template_name` kwarg from the base Panel class is intentional. Plugin authors should use TemplatePanel instead.
- Deferring model resolution errors until load time is intentional and necessary to avoid attempting to resolve a plugin model before it has been registered in the application.

#### Bugs / Inconsistencies

- Tweaked PluginContentPanel to not call `should_render()` (the inclusion of plugin content is not conditional at this level)
- Cleaned up AddObject
- Output is limited to the exception type and message, which is crucial for troubleshooting
- Tweaked `speed.html` to reference `value` instead of `port_speed`

netbox/netbox/ui/attrs.py
```diff
@@ -366,11 +384,13 @@ def __init__(self, *args, linkify=None, max_depth=None, **kwargs):
...
-        nodes = value.get_ancestors(include_self=True)
-        if self.max_depth:
-            nodes = list(nodes)[-self.max_depth:]
+    def get_context(self, obj, attr, value, context):
+        if value is not None:
+            nodes = value.get_ancestors(include_self=True)
+            if self.max_depth:
+                nodes = list(nodes)[-self.max_depth:]
+        else:
+            nodes = []
```
pheus:
```suggestion
    def get_context(self, obj, attr, value, context):
        nodes = []
        if value is not None:
            nodes = value.get_ancestors(include_self=True)
            if self.max_depth:
                nodes = list(nodes)[-self.max_depth:]
```

docs/plugins/development/ui-components.md
```diff
@@ -3,10 +3,7 @@
 !!! note "New in NetBox v4.5"
     All UI components described here were introduced in NetBox v4.5. Be sure to set the minimum NetBox version to 4.5.0 for your plugin before incorporating any of these resources.
```
pheus: I think we should explicitly call out that plugins using this API require NetBox v4.6.
This page still says 4.5.0, but with the breaking layout API changes in this PR, it might be worth updating.

netbox/netbox/ui/attrs.py
```diff
@@ -64,26 +80,29 @@ def get_value(self, obj):
...
-    def get_context(self, obj, context):
+    def get_context(self, obj, attr, value, context):
         """
         Return any additional template context used to render the attribute value.
 
         Parameters:
             obj (object): The object for which the attribute is being rendered
+            attr (str): The name of the attribute being rendered
+            value: The value of the attribute on the object
             context (dict): The root template context
```
pheus: Small question here: `ObjectAttribute.get_context()` describes `context` as the root template context, but `ObjectAttributesPanel` currently only passes a reduced context (`name` / `perms`) into the attrs.

Do we want to keep the full request/view context hidden from attributes? If so, could we update the parameter docstring to reflect the narrower context here?

netbox/templates/generic/object.html
```diff
@@ -124,11 +124,11 @@
...
 {% block content %}
   {# Render panel layout declared on view class #}
-  {% for row in layout.rows %}
+  {% for row in layout %}
     <div class="row">
-      {% for column in row.columns %}
-        <div class="col">
-          {% for panel in column.panels %}
+      {% for column in row %}
+        <div class="col{% if column.width %}-{{ column.width }}{% endif %}">
```
pheus: I think we lost the responsive breakpoint here.

Right now this renders `col` / `col-<width>`, so sibling columns stay side by side on small screens. Would it make sense to restore the mobile-friendly behavior and only apply the configured width from `md` upward?

```suggestion
        <div class="col-12 col-md{% if column.width %}-{{ column.width }}{% endif %}">
```

docs/plugins/development/ui-components.md
```diff
@@ -77,34 +74,14 @@ class RecentChangesPanel(Panel):
         }
```
pheus, line 74: Small docs suggestion: since `Panel.should_render()` is now the hook for conditional rendering, would it make sense to add a short example for it here as well?

The current section shows how to override `get_context()`, but a `should_render()` example would also help plugin authors discover the intended pattern for conditionally hiding a panel.

docs/plugins/development/ui-components.md
```diff
@@ -77,34 +74,14 @@ class RecentChangesPanel(Panel):
...
 ```
 
-NetBox also includes a set of panels suite for specific uses, such as display object details or embedding a table of related objects. These are listed below.
+NetBox also includes a set of panels suited for specific uses, such as display object details or embedding a table of related objects. These are listed below.
```
pheus:
```suggestion
NetBox also includes a set of panels suited for specific uses, such as displaying object details or embedding a table of related objects. These are listed below.
```

pheus (changes requested):
Thanks, this looks great!
I left a handful of small nits inline, mostly around the docs and one template tweak. Nothing blocking, just a few things that might be worth a second look.

pheus (approved):
Thank you very much, @jeremystretch !

Issue #21541, opened 2026-03-02
"is [not] empty" filter expects value


ColdRogue26:
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.1

### Python Version

3.12

### Steps to Reproduce

1. Go to filters in any list
2. Set any filter (e.g. "Speed" for interfaces) to "is empty" or "is not empty"
3. Click Search
4. Open Filter tab again

### Expected Behavior

Expected behavior would be the filter applying and filtering out any object applying to the filter. 3

### Observed Behavior

The filter "applies" without throwing an error, but it doesn't filter anything. Objects not matching the "is empty" or "is not empty" parameter still appear in the list. 
Going back to the filter settings, it now shows "Enter a whole number" (Interfaces > Speed "is [not] empty") or "Enter a list of values" (Interfaces > Duplex "is [not] empty")

https://github.com/user-attachments/assets/c2557e8c-370c-4295-86f9-dc3e90361be1

bctiemann:
It looks like this is going to be fixed by [PR #21829](https://github.com/netbox-community/netbox/pull/21829).

pheus:
Thank you for submitting this issue, however it appears that this topic has already been raised. Please see issue [#21535](https://github.com/netbox-community/netbox/issues/21535) for further discussion.

PR #21829, opened 2026-04-02
Fixes: #21535 - Fix filtering of object-type custom fields when "is empty" is selected

### LINKED ISSUE #21535: Exception raised when search a custom field related to object with "is empty" filter
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.3

### Python Version

3.12

### Steps to Reproduce

1. Create a custom field on any object, type =Object,  relating another any object type

Image

2. Use a custom search on the object and select 'is empty' filter on the custom field
Image

3. Click search

### Expected Behavior

The search returns with a table showing objects that has empty related custom field

### Observed Behavior

An exception is raised, with the message <class 'ValueError'> Field 'id' expected a number but got 'true'.

Image

bctiemann:
### Fixes: #21535 

The "is empty" filter option on custom fields pointing to objects was incorrectly passing the string `true`, which caused a ValueError. This PR fixes that behavior by properly returning a None value that is handled correctly in the filter and in the form validation.

### DIFF
--- a/netbox/utilities/forms/widgets/modifiers.py
+++ b/netbox/utilities/forms/widgets/modifiers.py
@@ -48,11 +48,13 @@ class FilterModifierWidget(forms.Widget):
             Just the value string for form validation. The modifier is reconstructed
             during rendering from the query parameter names.
         """
-        # Special handling for empty - check if field__empty exists
+        # Special handling for empty modifier: return None so the underlying field does not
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review

bctiemann:
This appears to fix [#21541](https://github.com/netbox-community/netbox/issues/21541) as well

arthanson (approved):

PR #21834, opened 2026-04-03
Fixes #21542: Increase supported interface speed values above 2.1 Tbps

pheus:
### Fixes: #21542

This PR changes `Interface.speed` from `PositiveIntegerField` to `PositiveBigIntegerField`, allowing higher speed values to be stored without hitting the current 32-bit limit.

It also updates the related GraphQL definitions, REST/OpenAPI filter typing, and regression tests to keep API behavior consistent.

Automatic calculation of a LAG's speed from its member interfaces was discussed separately, but is intentionally left out of scope for this fix.


pheus:

pheus:
@claude /review
Good catch on the remaining manual form/filter validation path.

I’m going to address this, but not via `forms.BigIntegerField`. Instead, I’ll add a small shared bigint-aware form field in NetBox, use it for the two manual `speed` form declarations, and wire `MultiValueBigNumberFilter` to use it as well. That keeps the bigint bounds centralized and avoids handling them as one-off inline values.

jnovinger (dismissed):
Looks good, approving (when CI passes) with a couple non-blocking observations:

`CircuitTermination.port_speed` and `upstream_speed` have the same `PositiveIntegerField` type, worth a follow-up issue.

The filterset and view tests still use speed values within the 32-bit range. The API test covers the boundary well with the 16 Tbps value, so this isn't critical, but adding a >32-bit speed value to the filterset test data and/or view test form_data would exercise the new capacity more thoroughly.

jnovinger (approved):

PR #21836, opened 2026-04-03
Closes #21766: Add reusable tests for sortable table columns

pheus:
### Fixes: #21766

Introduce `TableTestCases.OrderableColumnsTestCase`, a reusable base class for table ordering smoke tests.

The new helper discovers sortable columns from list-view querysets and verifies that each one renders without exceptions when ordered in both ascending and descending directions. This gives us a consistent way to catch unsupported ordering across model-backed tables without duplicating the same test logic in each app.

This PR also adds per-table smoke tests across circuits, core, dcim, extras, ipam, tenancy, users, virtualization, vpn, and wireless.

### DIFF
--- a/netbox/circuits/tests/test_tables.py
+++ b/netbox/circuits/tests/test_tables.py
@@ -1,48 +1,46 @@
-from django.test import RequestFactory, TestCase, tag
-
-from circuits.models import CircuitGroupAssignment, CircuitTermination
-from circuits.tables import CircuitGroupAssignmentTable, CircuitTerminationTable
+from circuits.tables import *
+from utilities.testing import TableTestCases
 
 
-@tag('regression')
-class CircuitTerminationTableTest(TestCase):
-    def test_every_orderable_field_does_not_throw_exception(self):
-        terminations = CircuitTermination.objects.all()
-        disallowed = {
-            'actions',
-        }
-
-        orderable_columns = [
-            column.name
-            for column in CircuitTerminationTable(terminations).columns
-            if column.orderable and column.name not in disallowed
-        ]
-        fake_request = RequestFactory().get('/')
-
-        for col in orderable_columns:
-            for direction in ('-', ''):
-                table = CircuitTerminationTable(terminations)
-                table.order_by = f'{direction}{col}'
-                table.as_html(fake_request)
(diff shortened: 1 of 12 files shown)


pheus:

pheus:
@claude /review

1. Good point on the empty-queryset limitation. The new helper is mainly intended as a smoke test for unsupported ordering, not as a full data-backed rendering test, so I have made that tradeoff explicit in the helper docstring.

2. Agreed on the regression tag.

3/4. Also agreed on the manual queryset labels. I’ll rename those to reflect the actual source more clearly, or use neutral queryset labels where there isn’t a real list view.

5. Good call on `cleanupSubTest()`.

netbox/circuits/tests/test_tables.py
```diff
@@ -1,48 +1,50 @@
...
+
+class CircuitTypeTableTest(TableTestCases.OrderableColumnsTestCase):
+    table = CircuitTypeTable
+
+
+class CircuitTableTest(TableTestCases.OrderableColumnsTestCase):
+    table = CircuitTable
+
+
+@tag("regression")
```
jeremystretch: Given that we're moving to a new testing mechanism for tables, IMO it makes sense to remove these tags. (It's safe to ignore Claude's earlier feedback.)

netbox/utilities/testing/tables.py
```diff
@@ -0,0 +1,157 @@
...
+            yield column.name
+
+
+class TableTestCases:
+    """
+    Keep test_* methods nested to avoid unittest auto-discovering the reusable
+    base classes directly.
+    """
+
+    class OrderableColumnsTestCase(ModelTableTestCase):
```
jeremystretch: Should we give this a more generic name? Seems reasonable to assume we'll add tests unrelated to ordering in the future.

jeremystretch (dismissed):
Nice work @pheus, really clean! Approving this with two minor administrative points for your consideration.

pheus:
That makes sense to me on both points.

I’ll drop the regression tags from the migrated table tests.

I’ll also rename `OrderableColumnsTestCase` to `StandardTableTestCase` so the shared class has room to grow as we add more common table tests in the future.

jeremystretch (approved):

PR #21837, opened 2026-04-03
Closes #21795: Improve humanize_speed formatting for decimal Gbps/Tbps values

### LINKED ISSUE #21795: Update `humanize_speed` to support decimal Gbps/Tbps output
### NetBox version

v4.5.5

### Feature type

### Proposed functionality

I'd like to propose updating the `humanize_speed` template filter so that it always formats a value using the largest appropriate unit, even when the converted result is not a whole number.

Today, the filter only promotes values to Gbps or Tbps when they divide evenly by `1000000` or `1000000000` Kbps. As a result, values such as `2500000` are rendered as `2500 Mbps` instead of `2.5 Gbps`, and `1600000000` is rendered as `1600 Gbps` instead of `1.6 Tbps`.

A practical implementation would keep the existing Kbps input, but allow decimal conversion for Mbps, Gbps, and Tbps while trimming trailing `.0` for whole-number results. For example:

- `2500000` → `2.5 Gbps`
- `50000000` → `50 Gbps`
- `800000000` → `800 Gbps`
- `1600000000` → `1.6 Tbps`

This would also complement `InterfaceSpeedChoices` updates by keeping rendered interface speeds consistent across the UI.

### Use case

This would make speed formatting in NetBox more consistent with the wording already used for modern interface speeds elsewhere in the application.

It also improves readability for users working with current hardware, where values such as `2.5 Gbps` and `1.6 Tbps` are much more natural than `2500 Mbps` or `1600 Gbps`.

### Database changes

None.

### External dependencies

None.

pheus:
### Fixes: #21795

This updates the `humanize_speed` template filter to always display speeds using the largest appropriate unit, including decimal values where needed. For example, `2500000` Kbps now renders as `2.5 Gbps` instead of `2500 Mbps`, and `1600000000` Kbps renders as `1.6 Tbps` instead of `1600 Gbps`.

This brings the filter in line with the speed notation already used elsewhere in NetBox and adds test coverage for whole-number and fractional values across Mbps, Gbps, and Tbps.

### DIFF
--- a/netbox/utilities/templatetags/helpers.py
+++ b/netbox/utilities/templatetags/helpers.py
@@ -186,26 +186,52 @@ def action_url(parser, token):
     return ActionURLNode(model, action, kwargs, asvar)
 
 
+def _format_speed(speed, divisor, unit):
+    """
+    Format a speed value with a given divisor and unit.
+
+    Handles decimal values and strips trailing zeros for clean output.
+    """
+    whole, remainder = divmod(speed, divisor)
+    if remainder == 0:
+        return f'{whole} {unit}'
+
+    # Divisors are powers of 10, so len(str(divisor)) - 1 matches the decimal precision.
+    precision = len(str(divisor)) - 1
+    fraction = f'{remainder:0{precision}d}'.rstrip('0')
+    return f'{whole}.{fraction} {unit}'
+
+
 @register.filter()
 def humanize_speed(speed):
     """
-    Humanize speeds given in Kbps. Examples:
+    Humanize speeds given in Kbps, always using the largest appropriate unit.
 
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
@claude /review

For the first point, the longer fractional output is intentional here. The goal of this change is to always render the largest appropriate unit without rounding uncommon values away, so I’d prefer to preserve the exact converted value rather than silently shorten it.

For the second point, agreed that the `precision` calculation assumes powers-of-10 divisors. I have added a short inline comment there to make that clearer for future readers.

bctiemann (approved):

PR #21842, opened 2026-04-03
Closes #21455: Add SQL indexes for default ordering

jeremystretch:
### Closes: #21455

Introduce a SQL index for each model that doesn't already have one suitable for its default ordering. Skip MPTT-enabled models because they're weird.

Some anecdata from querying devices at `OFFSET 49950`:

| Without index | With index | Improvement
|---|---|---|
| ~90ms | ~21ms | 77% reduction |


jeremystretch:
@claude /review check that all the new SQL indexes accurately reflect the default ordering of their corresponding models

bctiemann (approved):

PR #21846, opened 2026-04-06
Closes #21780: Add changelog message support for bulk creation of IP Addresses and Prefixes

pheus:
### Fixes: #21780

This PR adds changelog message support to bulk creation for both prefixes and IP addresses, bringing these workflows in line with single-object creation.

To support this, `IPAddressBulkAddForm` now uses `PrimaryModelForm`, field ordering has been updated, and bulk-add template rendering has been consolidated so the changelog message field is displayed consistently. Test coverage is also included to verify changelog message propagation during bulk creation.

Thanks for taking the time to review this change.


pheus:

pheus:
For (1), I left it as-is because it matches existing NetBox patterns: `ObjectChange` reads from `self._changelog_message`, and other generic bulk paths already set that attribute directly on the instance.

For (2), this is intentional with the shared renderer. `htmx/form.html` already handles `owner`, `comments`, and `changelog_message` outside declared fieldsets, so keeping those out of `IPAddressBulkAddForm.fieldsets` stays aligned with that common template behavior.

For (3), I agree it’s true in the abstract, but it doesn’t affect this PR because both bulk-add forms define `fieldsets`, so the fallback branch isn’t exercised here.

For (4), I don’t think there’s an actual risk here, since `ObjectPermission.name` is not unique.

bctiemann (approved):

PR #21856, opened 2026-04-07
Fixes #21430: Display Roles and Circuit Types as colored badges in detail views

pheus:
### Fixes: #21430

This PR adds reusable support for rendering color-capable related objects as colored badges in detail views.

It introduces an opt-in `colored` parameter for `RelatedObjectAttr`, `NestedObjectAttr`, and `RelatedObjectListAttr`, updates the `badge` template tag to support hex colors and optional URLs, and applies the new behavior to device roles, rack roles, inventory item roles, VM roles, circuit types, and virtual circuit types.

This keeps the change backward compatible while addressing the original device-role request and bringing similar detail views in line with the existing colored label presentation used elsewhere in the UI. Tests are included for both simple and nested related-object rendering.


pheus:

pheus:
For the duplication in `object.html`, I agree that extracting the colored badge block into a shared include would reduce repetition. That said, I’d prefer to keep it inline for now, since the grouped and ungrouped branches already differ structurally and pulling out this small section would add another layer of template indirection for a fairly limited amount of duplicated code.

On the test coverage point, I could add a very narrow test that effectively only verifies that `colored=colored` is passed through the include in `object_list.html`. At the moment, that feels a bit too implementation-specific to me for the value it would add, so I’d prefer to wait for maintainer feedback before expanding the patch in that direction.

If a reviewer would like either of these follow-ups, I’m very happy to add them.

jeremystretch (approved):

PR #21860, opened 2026-04-08
Closes #19138: Expose NAT details for primary IPs in the Device and Virtual Machine API

### LINKED ISSUE #19138: Show both "real" and NAT IP addresses when viewing a device in the API
### NetBox version

v4.2.6

### Feature type

### Proposed functionality

Display both the real primary IP address and corresponding NAT IP when viewing a device in the API.  This should be simple change to the brief fields on the IPAddress serializer.  I can take ownership and submit a PR. 

### Use case

When viewing a device object via the API, only the real OR NAT IP is shown, depending on which has been selected as the primary IP.  We use several different tools, some of which can use a device's NAT IP and others which must use the real IP address.  These tools use Netbox as an inventory and don't support additional logic to perform extra API lookups to retrieve the desired IP.  As a result, in for these tools to work simultaneously, both the primary real IP and NAT IP should be displayed in the API.  

### Database changes

NA

### External dependencies

NA

pheus:
### Fixes: #19138

This PR updates the nested IP representations used by `Device` and `VirtualMachine` so API consumers can see NAT relationships directly on primary IP fields. For `Device`, the same NAT details are also exposed on `oob_ip` for consistency.

To support the inline `fields=` approach cleanly, it also updates `get_prefetches_for_serializer()` to honor field-level `fields=` and `omit=` on nested serializers, so queryset prefetching stays aligned with the fields that are actually rendered.

Finally, this PR adds a small shared test helper for creating NAT IP pairs and extends API and utility test coverage around the new behavior.

### DIFF
--- a/netbox/dcim/api/serializers_/devices.py
+++ b/netbox/dcim/api/serializers_/devices.py
@@ -58,10 +58,30 @@ class DeviceSerializer(PrimaryModelSerializer):
     )
     status = ChoiceField(choices=DeviceStatusChoices, required=False)
     airflow = ChoiceField(choices=DeviceAirflowChoices, allow_blank=True, required=False)
-    primary_ip = IPAddressSerializer(nested=True, read_only=True, allow_null=True)
-    primary_ip4 = IPAddressSerializer(nested=True, required=False, allow_null=True)
-    primary_ip6 = IPAddressSerializer(nested=True, required=False, allow_null=True)
-    oob_ip = IPAddressSerializer(nested=True, required=False, allow_null=True)
+    primary_ip = IPAddressSerializer(
+        nested=True,
+        read_only=True,
+        allow_null=True,
+        fields=[*IPAddressSerializer.Meta.brief_fields, 'nat_inside', 'nat_outside'],
+    )
+    primary_ip4 = IPAddressSerializer(
+        nested=True,
+        required=False,
+        allow_null=True,
+        fields=[*IPAddressSerializer.Meta.brief_fields, 'nat_inside', 'nat_outside'],
+    )
+    primary_ip6 = IPAddressSerializer(
+        nested=True,
+        required=False,
+        allow_null=True,
+        fields=[*IPAddressSerializer.Meta.brief_fields, 'nat_inside', 'nat_outside'],
+    )
+    oob_ip = IPAddressSerializer(
+        nested=True,
+        required=False,
+        allow_null=True,
+        fields=[*IPAddressSerializer.Meta.brief_fields, 'nat_inside', 'nat_outside'],
+    )
     parent_device = serializers.SerializerMethodField()
     cluster = ClusterSerializer(nested=True, required=False, allow_null=True)
     virtual_chassis = VirtualChassisSerializer(nested=True, required=False, allow_null=True, default=None)
--- a/netbox/utilities/api.py
+++ b/netbox/utilities/api.py
@@ -11,7 +11,7 @@ from django.urls import reverse
 from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
 from rest_framework.permissions import BasePermission
-from rest_framework.serializers import Serializer
+from rest_framework.serializers import ListSerializer, Serializer
 from rest_framework.views import get_view_name as drf_get_view_name
 
 from extras.constants import HTTP_CONTENT_TYPE_JSON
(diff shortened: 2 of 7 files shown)


pheus:

pheus:
@claude /review

I’ve narrowed the scope of the PR and removed the `VirtualDeviceContext` change. After looking at it more closely, `Device` and `VirtualMachine` both support using a NAT outside IP as the primary IP when its inside IP is assigned to the object, but `VirtualDeviceContext` does not. Keeping VDC in this PR would have made the API surface broader without matching model behavior, so restricting this to `Device` and `VirtualMachine` felt cleaner.

I’ve also updated the `IPAddressWithNATSerializer` docstring accordingly.

On the prefetch point, agreed there is some additional overhead because these NAT fields are now part of the nested serializer’s brief representation. I haven’t measured a concrete query delta, so I’d rather not overstate the exact impact, but the extra work is limited to the NAT relations on `primary_ip4` / `primary_ip6`. Given the FR and the narrow serializer-scoped approach, I think that tradeoff is reasonable here.

Keeping `oob_ip` unchanged was intentional so this stays focused on primary IPs only.

netbox/dcim/api/serializers_/devices.py
```diff
@@ -58,9 +58,9 @@ class DeviceSerializer(PrimaryModelSerializer):
     )
     status = ChoiceField(choices=DeviceStatusChoices, required=False)
     airflow = ChoiceField(choices=DeviceAirflowChoices, allow_blank=True, required=False)
-    primary_ip = IPAddressSerializer(nested=True, read_only=True, allow_null=True)
-    primary_ip4 = IPAddressSerializer(nested=True, required=False, allow_null=True)
-    primary_ip6 = IPAddressSerializer(nested=True, required=False, allow_null=True)
+    primary_ip = IPAddressWithNATSerializer(nested=True, read_only=True, allow_null=True)
```
jeremystretch: Rather than introducing a new serializer class, we should be able to pass a list of fields to the nested serializer (a fairly recent addition to NetBox's capabilities):

```python
    primary_ip = IPAddressSerializer(
        nested=True,
        read_only=True,
        allow_null=True,
        fields=[*IPAddressSerializer.Meta.brief_fields, 'nat_inside', 'nat_outside']
    )
```

netbox/dcim/api/serializers_/devices.py
```diff
@@ -58,9 +58,9 @@ class DeviceSerializer(PrimaryModelSerializer):
...
+    primary_ip4 = IPAddressWithNATSerializer(nested=True, required=False, allow_null=True)
+    primary_ip6 = IPAddressWithNATSerializer(nested=True, required=False, allow_null=True)
     oob_ip = IPAddressSerializer(nested=True, required=False, allow_null=True)
```
jeremystretch: Although it wasn't mentioned in the FR, we should probably do the same for the OOB IP as well.

jeremystretch (changes requested):

netbox/dcim/api/serializers_/devices.py
pheus: I went with a dedicated serializer here because the nested serializer’s `brief_fields` also drive NetBox’s automatic queryset prefetching. With an inline `fields=` override, the NAT fields would be rendered, but they would not be picked up by the current prefetch logic, so this felt like the safer option from a performance perspective.

I did consider extending the prefetch mechanism to honor field-level `fields=`, but given the narrow scope of this change, a dedicated serializer seemed like the smaller and more targeted approach.

That said, if you’d prefer the inline `fields=` style here, I’m happy to rework it.

netbox/dcim/api/serializers_/devices.py
pheus: That’s fair. I left `oob_ip` unchanged initially to keep the scope aligned with the FR and because NAT on OOB addresses is much less common in practice, but I agree that handling it the same way would be more consistent. I can include `oob_ip` in this PR as well.

pheus:

pheus:
@claude /review
1. This is actually the correct behavior. If someone explicitly passes omit=, they want the full field list minus the omitted ones, not brief_fields. No action needed.
2. Should be fixed now.

netbox/utilities/api.py
```diff
@@ -130,14 +154,13 @@ def get_prefetches_for_serializer(serializer_class, fields=None, omit=None):
...
-                subfields = serializer_field.Meta.brief_fields if serializer_field.nested else None
-                for subfield in get_prefetches_for_serializer(type(serializer_field), subfields):
-                    prefetch_fields.append(f'{field_name}__{subfield}')
+        # If this field is represented by a nested serializer, recurse to resolve
+        # prefetches for the related object, honoring any field-level fields=/omit=
+        # constraints set on that serializer field instance.
+        if nested_serializer := _get_nested_serializer(serializer_field):
+            subfields = _get_serializer_fields(nested_serializer)
+            for subfield in get_prefetches_for_serializer(type(nested_serializer), fields=subfields):
+                prefetch_fields.append(f'{field.name}__{subfield}')
```
jeremystretch, line 163: Nice, thanks for tackling this!

jeremystretch (approved):

PR #21861, opened 2026-04-08
Fixes #21529: Prevent unknown custom field data from being persisted or shown in change logs

pheus:
### Fixes: #21529

This PR fixes handling of unknown custom field names during API updates.

It rejects unknown custom field keys submitted via the API, preserves `model.clean()` normalization during serializer validation, and filters stale custom field keys during serialization. Together, this prevents invalid custom field data from being persisted or appearing in change logs.

It also adds regression tests covering both rejection of unknown submitted keys and cleanup of stale custom field data during updates.

### DIFF
--- a/netbox/extras/api/customfields.py
+++ b/netbox/extras/api/customfields.py
@@ -85,8 +85,18 @@ class CustomFieldsDataField(Field):
                 "values."
             )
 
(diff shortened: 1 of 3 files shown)


pheus:

pheus:

pheus:
@claude /review
Both points should be addressed now.

arthanson (approved):

PR #21863, opened 2026-04-08
Fixes #21801: Ensure unique Image Attachment filenames when using S3 storage

### LINKED ISSUE #21801: Duplicate filename allowed when upload files using S3 storage
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.5

### Python Version

3.12

### Steps to Reproduce

1. switch to using S3 storage for image upload
2. go to site and add an image attachment for a given file and choose "Create and Add Another"
3. upload the same file again.
4. check the uploaded image filenames in both image attachment and in Object Change

### Expected Behavior

The images should have different filenames, when you do this using local file storage a unique value will be appended onto the filename to make it unique, for example:

site_28_action-buttons.png
site_28_action-buttons_sdmmer4.png

While having these names be the same causes no problem in the UI, the Object Change having the same name will cause issues with branching and is different behavior then when using local storage.

### Observed Behavior

The file is uploaded in both places with the same name.  Here is the object change records for both files:

```
name: ""
image: "image-attachments/site_28_action-buttons.png"
object_id: 28
description: ""
image_width: 1698
object_type: 57
image_height: 342
```

```
name: ""
image: "image-attachments/site_28_action-buttons.png"
object_id: 28
description: ""
image_width: 1698
object_type: 57
image_height: 342
```

as can be seen they are duplicated. On local file storage the second image will be:

```
name: ""
image: "image-attachments/site_28_action-buttons_sdmmer4.png"
object_id: 28
description: ""
image_width: 1698
object_type: 57
image_height: 342
```

pheus:
### Fixes: #21801

This updates image attachment filename generation to use Django's base collision handling so overwrite-style storage backends, such as S3, behave consistently with local file storage.

With this change, the first upload keeps its original filename and only a colliding upload receives a suffixed name. This avoids duplicate image paths being recorded for separate attachments in `ObjectChange`, which was causing problems for branching.

I've also added regression tests covering both path generation and collision handling.


jeremystretch:

pheus:
@claude /review

The direct call to `Storage.get_available_name()` is intentional here, because calling `field.storage.get_available_name()` would preserve overwrite-style backend behavior and reintroduce the bug.

I've added a short comment to make that explicit and to note that we still want the configured storage instance to handle `exists()` / alternative name generation.

Agreed on the test concern. Rather than documenting the Django import path assumption, I updated the fake storage classes to return a deterministic `get_alternative_name()` and removed the `get_random_string` patching entirely.

That keeps the tests deterministic while staying focused on NetBox's behavior instead of Django's internal module layout.

bctiemann (approved):

PR #21867, opened 2026-04-08
Closes #21862: Stabilize ScriptModule tests and reduce CI noise

### LINKED ISSUE #21862: Reduce CI noise from script tests that reference non-existent files
### Proposed Changes

Some tests currently create `ScriptModule` objects with file paths such as `script.py` and `test_script.py` without creating the corresponding files on disk. This causes NetBox to log `Failed to load script: ...` during CI, even though the tests later create the `Script` records explicitly and monkey-patch `Script.python_class`.

A small cleanup would be to avoid calling `sync_classes()` during setup for these tests, for example by patching `ScriptModule.sync_classes()` around `ScriptModule.objects.create(...)` in the affected test cases. This would keep the test intent the same while preventing unnecessary script load attempts during setup.

### Justification

The test suite passes successfully today, so this does not appear to be a functional bug. However, the current behavior adds avoidable error output to CI logs, which makes the test output noisier and can make real failures harder to spot.

Since these tests are not actually validating on-disk script loading, preventing the setup-time import attempt seems like the cleanest option. It keeps production error logging intact for genuine script load failures while making CI output a bit easier to read.

pheus:
### Fixes: #21862

This PR cleans up a few `ScriptModule` tests that were producing noisy `Failed to load script` messages in CI even though the tests themselves were passing.

It patches `ScriptModule.sync_classes()` in tests that intentionally create `ScriptModule` records without backing files, avoiding setup-time import attempts for non-existent modules.

For the upload API test, it mocks both serializer and module storage access so the uploaded script can be saved and imported during the test, and adds an assertion to confirm that the related `Script` object is created.

This keeps the test coverage and intent intact while making CI output a bit cleaner and easier to review.

### DIFF
--- a/netbox/extras/tests/test_api.py
+++ b/netbox/extras/tests/test_api.py
@@ -1,5 +1,6 @@
 import datetime
 import hashlib
+import io
 from unittest.mock import MagicMock, patch
 
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
Good catch on the duplicate `sync_classes()` calls. That behavior is already present on `main`, and this PR only makes it more visible by letting the upload test exercise the real import path, so I’d prefer to keep that as a separate follow-up rather than expand the scope here.

The `BytesIO` use was intentional so the mocked storage can be reopened during import, and I left `sync_classes()` enabled in the upload test on purpose to cover the save/import flow end to end.

For the `from extras.scripts import Script` import, that is the documented import path for custom scripts, so this felt like a reasonable dependency for the test.

jnovinger (approved):

PR #21871, opened 2026-04-09
Closes #21865: Display debug toolbar if `INTERNAL_IPS` is empty

jeremystretch:
### Closes: #21865

This has deviated slightly from the original proposal in the interest of preserving existing capability. Instead of ignoring `INTERNAL_IPS` entirely, we allow for it to be empty, and if so display the debug toolbar for every request (provided debugging is enabled).

- Introduce a custom `show_toolbar()` function in `utilities.debug` to return True if debugging is enabled and `INTERNAL_IPS` is falsy
- Define `DEBUG_TOOLBAR_CONFIG` in `settings.py` and configure `SHOW_TOOLBAR_CALLBACK`
- Remove `INTERNAL_IPS` from the example configuration file as it's a fairly niche parameter
- Document the new debug toolbar behavior

### DIFF
--- a/docs/configuration/development.md
+++ b/docs/configuration/development.md
@@ -6,7 +6,7 @@ Default: `False`
 
 This setting enables debugging and displays a debugging toolbar in the user interface. Debugging should be enabled only during development or troubleshooting.
 
-Note that the debugging toolbar will be displayed only for requests originating from an [internal IP address](./system.md#internal_ips), if defined. If no internal IPs are defined, the toolbar will be displayed for all requests.
+Note that the debugging toolbar will be displayed only for requests originating from [internal IP addresses](./system.md#internal_ips), if defined. If no internal IPs are defined, the toolbar will be displayed for all requests.
 
 !!! warning
     Never enable debugging on a production system, as it can expose sensitive data to unauthenticated users and impose a
--- a/docs/configuration/development.md
+++ b/docs/configuration/development.md
@@ -4,9 +4,9 @@
 
 Default: `False`
 
-This setting enables debugging. Debugging should be enabled only during development or troubleshooting. Note that only
-clients which access NetBox from a recognized [internal IP address](./system.md#internal_ips) will see debugging tools in the user
-interface.
+This setting enables debugging and displays a debugging toolbar in the user interface. Debugging should be enabled only during development or troubleshooting.
+
+Note that the debugging toolbar will be displayed only for requests originating from an [internal IP address](./system.md#internal_ips), if defined. If no internal IPs are defined, the toolbar will be displayed for all requests.
 
 !!! warning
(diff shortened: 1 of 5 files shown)


jeremystretch:
@claude /review

pheus (changes requested):
This is really great! Thank you!
The implementation itself looks good, and I appreciate the effort to preserve the existing `INTERNAL_IPS` behavior.

My only remaining concern is around discoverability and alignment with the original FR. I understand the reasoning for preserving the current `INTERNAL_IPS` behavior, but for the support-team use case I think the relevant settings should stay very easy to find and configure.

Because of that, I’d be happy with either of these directions:

- align fully with the FR and enable the toolbar by default whenever `DEBUG=True`; or
- keep the current behavior, but group the debug-related settings under a dedicated header in `configuration_example.py` and retain a commented `INTERNAL_IPS` example with an explanatory note.

Something like this near the end of the example config would make the current behavior much more discoverable:

```python
#########################
#                       #
#   Debug settings      #
#                       #
#########################

# Set to True to enable server debugging.
# WARNING: Debugging introduces a significant performance penalty and may reveal sensitive information about your
# installation. Only enable debugging while performing testing. Never enable debugging on a production system.
DEBUG = False

# IP addresses recognized as internal to the system.
# If defined, the debugging toolbar will be displayed only to clients accessing NetBox from a listed IP address.
# If set to an empty list, the toolbar will be available for all requests when DEBUG is True.
# INTERNAL_IPS = []
```

(Adding `DEVELOPER` to the same section would also make sense to me, but `DEBUG` + `INTERNAL_IPS` are the important bits for this PR.)

jeremystretch:
> align fully with the FR and enable the toolbar by default whenever DEBUG=True

I decided against this approach because it removes the existing ability to limit the debug panel to only certain clients, which can be valuable in certain circumstances.

> keep the current behavior, but group the debug-related settings under a dedicated header in configuration_example.py and retain a commented INTERNAL_IPS example with an explanatory note.

The example configuration file isn't meant to serve as documentation: It's intended merely to illustrate a subset of commonly customized parameters. That's why I opted to remove `INTERNAL_IPS` - primarily a development setting - entirely. In the actual configuration docs (where users _should_ be looking for guidance), we do specifically reference the `INTERNAL_IPs` parameter in relation to the `DEBUG` parameter, which I think is sufficient:

---

## DEBUG

Default: `False`

This setting enables debugging and displays a debugging toolbar in the user interface. Debugging should be enabled only during development or troubleshooting.

Note that the debugging toolbar will be displayed only for requests originating from [internal IP addresses](./system.md#internal_ips), if defined. If no internal IPs are defined, the toolbar will be displayed for all requests.

pheus (approved):
Thanks for taking the time to clarify your point! I appreciate that!

PR #21880, opened 2026-04-10
Add post_raw_create signal hook for running post-save logic after raw object creation

### LINKED ISSUE #21879: Add post_raw_create signal hook for models requiring post-save logic after raw object creation
### NetBox version

v4.5.7

### Feature type

Data model extension

### Proposed functionality

Introduce a post_raw_create signal in netbox.signals that is sent after one or more objects of a given model are created using a raw save (i.e. save_base(raw=True)). Unlike Django's built-in post_save, this signal is designed for bulk notification — it receives a pks list rather than a single instance — and is intended for callers that create objects outside the normal model save() path.

The signal follows the same sender-filtering pattern as other Django signals, allowing receivers to subscribe only to the model type they care about:

```
  @receiver(post_raw_create, sender=Cable)
  def retrace_cable_paths_after_raw_create(sender, pks, **kwargs):
      ...
```

NetBox itself connects a receiver for Cable in dcim/signals.py to retrace cable paths after cables are created this way, since Cable.save() (which fires trace_paths) is bypassed during raw creation.

### Use case

Django's `save_base(raw=True)` is commonly used when restoring or importing serialized object data - for example, during data migrations, fixture loading, or integrations that replay an audit trail of changes. In these scenarios, model-level logic in save() is intentionally skipped to avoid side effects during bulk application.

However, some models (Cable specifically) rely on their `save()` method to trigger necessary secondary operations. Cable is a concrete example: `Cable.save()` fires the `trace_paths signal`, which causes CablePath records to be created or updated. When cables are created via raw save, no CablePath records are produced, leaving the physical connectivity graph incomplete.

Without a hook like post_raw_create, any code that creates objects via raw save must either contain model-specific workarounds or accept that model-level post-save logic will not run. The post_raw_create signal provides a standard extension point: models can register receivers to perform whatever cleanup or secondary operations are needed after raw creation, keeping that logic in the model layer where it belongs rather than in the calling code.

### Database changes

N/A

### External dependencies

N/A

arthanson:
### Fixes: #21879 

Triggers cable path retracing after a Cable is created outside the normal save() path (e.g., via bulk_create()). Callers must dispatch Django's built-in post_save signal with post_raw_create=True after all CableTermination records are in place. Without the flag the handler is a no-op, so normal saves are unaffected.

We need a new parameter on the signal-handler instead of using raw=True. Using a handler gated on raw=True would fire at exactly the wrong time. Django sets raw=True during loaddata and fixture imports - scenarios where the database is being populated in dependency order and CableTermination records for a given cable may not exist yet.

### DIFF
--- a/netbox/dcim/signals.py
+++ b/netbox/dcim/signals.py
@@ -167,13 +167,13 @@ def retrace_cable_paths(instance, **kwargs):
 
 
 @receiver(post_save, sender=Cable)
-def retrace_cable_paths_on_update_dependencies(sender, instance, update_dependencies=False, **kwargs):
+def retrace_cable_paths_on_raw_create(sender, instance, post_raw_create=False, **kwargs):
     """
     Trigger cable path retracing for a Cable instance when dependencies need to be
-    recalculated. Callers should fire post_save with update_dependencies=True after
+    recalculated. Callers should fire post_save with post_raw_create=True after
     all CableTerminations are in place.
     """
-    if not update_dependencies:
+    if not post_raw_create:
         return
     instance._terminations_modified = True
     trace_paths.send(Cable, instance=instance, created=True)
--- a/netbox/dcim/signals.py
+++ b/netbox/dcim/signals.py
@@ -4,6 +4,8 @@ from django.db.models import Q
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 
+from netbox.signals import post_raw_create
+
 from dcim.choices import CableEndChoices, LinkStatusChoices
 from ipam.models import Prefix
 from virtualization.models import Cluster, VMInterface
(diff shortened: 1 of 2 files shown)


arthanson:
@claude /review

arthanson:
re claude review - updated documentation for the signal handler regarding raw and that terminations need to be set already.

jeremystretch (commented):
I'm not sure I understand the justification for introducing a new signal. Isn't this already possible using Django's stock [`post_save` signal](https://docs.djangoproject.com/en/6.0/ref/signals/#post-save) by filtering on the `raw` argument? I understand that the proposed implementation utilizes a list of primary keys, but it's just operating on a `for` loop that isn't substantially different from call the receiver function one per object. I'd much rather leverage the existing signal if feasible.

netbox/dcim/signals.py
```diff
@@ -166,6 +166,19 @@ def retrace_cable_paths(instance, **kwargs):
         cablepath.retrace()
 
 
+@receiver(post_save, sender=Cable)
+def retrace_cable_paths_on_raw_create(sender, instance, post_raw_create=False, **kwargs):
```
jeremystretch: We should probably stick with the [documented signature](https://docs.djangoproject.com/en/6.0/ref/signals/#post-save) for `post_save` receivers.
```suggestion
def retrace_cable_paths_on_raw_create(sender, instance, raw=False, **kwargs):
```

jeremystretch (changes requested):

netbox/dcim/signals.py
```diff
@@ -166,6 +166,20 @@ def retrace_cable_paths(instance, **kwargs):
...
 
 
+@receiver(post_save, sender=Cable)
+def retrace_cable_paths_on_raw_create(sender, instance, raw=False, **kwargs):
+    """
+    Trigger cable path retracing for a Cable instance when dependencies need to be
+    recalculated. Callers should fire post_save with post_raw_create=True after
+    all CableTerminations are in place.
+    """
+    post_raw_create = kwargs.get('post_raw_create', False)
```
jeremystretch, line 176: Sorry, my earlier feedback wasn't clear. I don't think we need a custom argument at all. `post_save` already passes `created` and `raw` booleans: It should suffice to check whether both are true.

jeremystretch (changes requested):

arthanson:
See comment at top: "We need a new parameter on the signal-handler instead of using raw=True. Using a handler gated on raw=True would fire at exactly the wrong time. Django sets raw=True during loaddata and fixture imports - scenarios where the database is being populated in dependency order and CableTermination records for a given cable may not exist yet."

Since we want this not just for cable but potentially use similar if another model in the future has something similar I'm very leery of changing the meaning of the signal handler raw=True as that has a different meaning in Django and we could have side-effects to loaddata (among others, not sure what all Django might use it for).

PR #21892, opened 2026-04-10
Closes #21890: Deprecate the `models` registry key

### LINKED ISSUE #21890: Deprecate the `models` registry key
### Deprecated Functionality

Deprecate references to the `models` registry key, i.e. `registry['models']`. This will be removed in NetBox v4.7.0.

### Scheduled removal

v4.7.0

### Justification

Maintaining a list of public models in the registry is no longer necessary since the implementation of ObjectType feature flags in NetBox v4.4.0 (#19924).

### Impact

Any plugins or custom scripts still utilizing this registry key will need to be adapted. Object types for public models are now returned via

```python
ObjectType.objects.public()
```

jeremystretch:
### Closes: #21890

Raise a DeprecationWarning on access to `registry['models']`

### DIFF
--- a/docs/development/application-registry.md
+++ b/docs/development/application-registry.md
@@ -32,6 +32,9 @@ Core model features are listed in the [features matrix](./models.md#features-mat
 
 ### `models`
 
+!!! warning "Deprecated"
+    Usage of this key has been deprecated and will be removed in NetBox v4.7. Use `ObjectType.objects.public()` to find registered models.
+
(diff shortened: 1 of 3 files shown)


pheus (changes requested):
Looks good overall!
The direct `dict.__getitem__()` access in `register_models()` is super nice!

I just have one small follow-ups I'd love to see:

Could we also update the relevant docs (`docs/development/application-registry.md`)? The application registry docs still present `models` as a normal store, so it would be helpful to mark it deprecated there and point readers to `ObjectType.objects.public()`.

pheus (approved):
Thanks, @jeremystretch !

PR #21899, opened 2026-04-13
Fixes #21869: Remove redundant ScriptModule class synchronization on save

### LINKED ISSUE #21869: `ScriptModule.save()` triggers `sync_classes()` twice for a single save
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.7

### Python Version

3.12

### Steps to Reproduce

1. Start a Django shell with `./manage.py nbshell`.
2. Run the following to create a small script file and count how many times `sync_classes()` is called during a single save:
   ```python
   from pathlib import Path
   from django.conf import settings

   script_path = Path(settings.SCRIPTS_ROOT) / "example.py"
   script_path.parent.mkdir(parents=True, exist_ok=True)
   script_path.write_text(
       "from extras.scripts import Script\n"
       "class Example(Script):\n"
       "    pass\n"
   )

   calls = []
   original = ScriptModule.sync_classes

   def counted(self, *args, **kwargs):
       calls.append(self.file_path)
       return original(self, *args, **kwargs)

   ScriptModule.sync_classes = counted
   module = ScriptModule(file_path="example.py")
   module.save()

   print(len(calls))
   ```
3. Observe that the output is `2`.

### Expected Behavior

Saving a `ScriptModule` should synchronize its classes once for a single save operation.

### Observed Behavior

`sync_classes()` is invoked twice for one `ScriptModule.save()`: once directly in `save()` and again from the `post_save` signal handler. The second call appears to be redundant.

pheus:
### Fixes: #21869

This PR removes the `post_save` signal handler that calls `ScriptModule.sync_classes()`, leaving the existing explicit call in `ScriptModule.save()` as the single synchronization path.

Since `save()` already performs the sync after the model has been written, the additional signal-based call is redundant. Removing it avoids duplicate work during a single save and keeps the behavior a bit simpler and easier to follow.

### DIFF
--- a/netbox/extras/models/scripts.py
+++ b/netbox/extras/models/scripts.py
@@ -5,8 +5,6 @@ from functools import cached_property
 from django.contrib.contenttypes.fields import GenericRelation
 from django.db import models
 from django.db.models import Q
-from django.db.models.signals import post_save
-from django.dispatch import receiver
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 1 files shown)


jnovinger (dismissed):
Approved as-is, but perhaps we also want to put a comment where `sync_classes()` is triggered to make sure it's not removed accidentally. Or a regression test ensuring it gets called.

jnovinger (approved):

PR #21900, opened 2026-04-13
Fixes #21711: Added support for filtering and viewing modules by their module type profile

ITJamie:
### Fixes: #21711

  Added support for filtering `/dcim/modules/` by module type profile, matching the existing module type profile relationship on `ModuleType`.

  ## Changes

  - Added `profile` and `profile_id` filters to the module list filterset.
  - Exposed module profile filtering in the module list UI, including support for `None` / unprofiled module types.
  - Added module profile as an available table field in `/dcim/modules/`.
  - Added REST API test coverage for filtering modules by profile ID and profile name.
  - Added UI filterset, form, table, and utility widget regression tests.
  - Updated the module model documentation to describe profile-based filtering.
  - Updated Module Panel to display Module Profile

### DIFF
--- a/docs/models/dcim/module.md
+++ b/docs/models/dcim/module.md
@@ -18,6 +18,10 @@ The [module bay](./modulebay.md) into which the module is installed.
 
 The [module type](./moduletype.md) which represents the physical make & model of hardware. By default, module components will be instantiated automatically from the module type when creating a new module.
 
+### Profile
+
+The [module type profile](./moduletypeprofile.md) associated with the selected module type. Module list views and the REST API can be filtered by this related field.
+
 ### Status
 
 The module's operational status.
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1537,6 +1537,19 @@ class VirtualDeviceContextFilterSet(PrimaryModelFilterSet, TenancyFilterSet, Pri
 
 @register_filterset
 class ModuleFilterSet(PrimaryModelFilterSet):
+    profile_id = django_filters.ModelMultipleChoiceFilter(
+        field_name='module_type__profile',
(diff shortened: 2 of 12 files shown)


pheus:

ITJamie:
Re 1

This was a pain point when I was testing locally. I can drop it if its too far out of scope. 

Re 2

Is the preference to lookup the settings reference or to use the hardcoded string

arthanson (changes requested):
Looks good, a few changes:

1. Please take out the null option widget change - if desired open up a separate FR and PR for that as it is outside the scope of this issue.
2. Use settings.FILTERS_NULL_CHOICE_VALUE in tests rather than the literal string. see test_filter_modifiers.py:110,131 - null_value = settings.FILTERS_NULL_CHOICE_VALUE, and test_filters.py:81,89 - same pattern
3. I think it makes potentially makes sense to add this to the detail view as well, under Module Type. As we are filtering on it and in the table seems like it should probably be in the detail view for the module?

ITJamie:
Changes made as requested

arthanson (approved):
Thanks @ITJamie

docs/models/dcim/module.md
```diff
@@ -18,6 +18,10 @@ The [module bay](./modulebay.md) into which the module is installed.
 
 The [module type](./moduletype.md) which represents the physical make & model of hardware. By default, module components will be instantiated automatically from the module type when creating a new module.
 
+### Profile
+
+The [module type profile](./moduletypeprofile.md) associated with the selected module type. Module list views and the REST API can be filtered by this related field.
+
```
jeremystretch, line 24: This should not have been added. There is no `profile` field on the Module model.

jeremystretch: This is fixed under #21955.

PR #21901, opened 2026-04-13
21866 Include the PostgreSQL database schema within system details

arthanson:
### Fixes: #21866 

Added "Database" tab to System view, also refactored the SystemView class get function to make it more modular / readable.

Monosnap System | NetBox 2026-04-13 09-06-22
Monosnap System | NetBox 2026-04-13 09-39-10


arthanson:
@claude /review

arthanson:
claude review, #1, #3 fixed.  Others not really needed.

netbox/templates/core/system.html
```diff
@@ -173,4 +179,133 @@ <h2 class="card-header">{% trans "Object Counts" %}</h2>
...
+      {# Tables grouped by app prefix #}
+      {% for group in db_schema_groups %}
+        <div class="card mb-3">
+          <h2 class="card-header">
+            <button class="accordion-button collapsed p-0 w-100" type="button"
+                    data-bs-toggle="collapse" data-bs-target="#db-group-body-{{ group.name }}"
+                    aria-expanded="false" aria-controls="db-group-body-{{ group.name }}">
+              {{ group.name }}
+              {% if group.is_plugin %}<span class="badge bg-purple text-white ms-1">{% trans "plugin" %}</span>{% endif %}
+              <span class="badge bg-secondary text-white ms-1">{{ group.tables|length }} {% trans "tables" %}</span>
```
pheus:
```suggestion
              <span class="badge bg-secondary text-bg-gray ms-1">{{ group.tables|length }} {% trans "tables" %}</span>
```
This CSS class ensures that the contrast between background and foreground color works in light and dark mode.

netbox/templates/core/system.html
```diff
@@ -173,4 +179,133 @@ <h2 class="card-header">{% trans "Object Counts" %}</h2>
...
+              <span class="badge bg-secondary text-white ms-1">{{ group.index_count }} {% trans "indexes" %}</span>
```
pheus:
```suggestion
              <span class="badge bg-secondary text-bg-gray ms-1">{{ group.index_count }} {% trans "indexes" %}</span>
```

netbox/templates/core/system.html
```diff
@@ -173,4 +179,133 @@ <h2 class="card-header">{% trans "Object Counts" %}</h2>
...
+              <span class="accordion-button-toggle"><i class="mdi mdi-chevron-down"></i></span>
+            </button>
+          </h2>
+          <div id="db-group-body-{{ group.name }}" class="collapse">
+            <div class="accordion accordion-flush" id="db-group-{{ group.name }}">
+              {% for table in group.tables %}
+                <div class="accordion-item">
+                  <h3 class="accordion-header" id="table-heading-{{ group.name }}-{{ forloop.counter }}">
```
pheus:
```suggestion
                  <h5 class="accordion-header" id="table-heading-{{ group.name }}-{{ forloop.counter }}">
```

netbox/templates/core/system.html
```diff
@@ -173,4 +179,133 @@ <h2 class="card-header">{% trans "Object Counts" %}</h2>
...
+                            data-bs-toggle="collapse" data-bs-target="#table-collapse-{{ group.name }}-{{ forloop.counter }}"
+                            aria-expanded="false" aria-controls="table-collapse-{{ group.name }}-{{ forloop.counter }}">
+                      {{ table.name }}
+                      <span class="badge bg-secondary text-white ms-2">{{ table.columns|length }} {% trans "columns" %}</span>
+                      {% if table.indexes %}
+                        <span class="badge bg-secondary text-white ms-1">{{ table.indexes|length }} {% trans "indexes" %}</span>
+                      {% endif %}
+                      <span class="accordion-button-toggle"><i class="mdi mdi-chevron-down"></i></span>
+                    </button>
+                  </h3>
```
pheus:
```suggestion
                  </h5>
```

netbox/templates/core/system.html
```diff
@@ -173,4 +179,133 @@ <h2 class="card-header">{% trans "Object Counts" %}</h2>
...
+      </div>
+      {# Tables grouped by app prefix #}
+      {% for group in db_schema_groups %}
+        <div class="card mb-3">
+          <h2 class="card-header">
+            <button class="accordion-button collapsed p-0 w-100" type="button"
+                    data-bs-toggle="collapse" data-bs-target="#db-group-body-{{ group.name }}"
+                    aria-expanded="false" aria-controls="db-group-body-{{ group.name }}">
+              {{ group.name }}
+              {% if group.is_plugin %}<span class="badge bg-purple text-white ms-1">{% trans "plugin" %}</span>{% endif %}
```
pheus:
```suggestion
              {% if group.is_plugin %}<span class="badge text-bg-purple ms-1">{% trans "plugin" %}</span>{% endif %}
```

netbox/templates/core/system.html
```diff
@@ -173,4 +179,133 @@ <h2 class="card-header">{% trans "Object Counts" %}</h2>
...
+              <span class="badge bg-secondary text-white ms-1">{{ group.index_count }} {% trans "indexes" %}</span>
+              <span class="accordion-button-toggle"><i class="mdi mdi-chevron-down"></i></span>
+            </button>
+          </h2>
+          <div id="db-group-body-{{ group.name }}" class="collapse">
+            <div class="accordion accordion-flush" id="db-group-{{ group.name }}">
+              {% for table in group.tables %}
+                <div class="accordion-item">
+                  <h3 class="accordion-header" id="table-heading-{{ group.name }}-{{ forloop.counter }}">
+                    <button class="accordion-button collapsed font-monospace" type="button"
```
pheus:
```suggestion
                    <button class="accordion-button border-bottom collapsed font-monospace" type="button"
```

netbox/core/views.py
```diff
@@ -686,20 +686,102 @@ def get(self, request):
...
+
+    def _get_db_schema(self):
+        db_schema = []
+        try:
+            with connection.cursor() as cursor:
+                cursor.execute("""
+                    SELECT table_name, column_name, data_type, is_nullable, column_default
+                    FROM information_schema.columns
+                    WHERE table_schema = 'public'
+                    ORDER BY table_name, ordinal_position
```
pheus: I think hardcoding `public` here is going to bite us. NetBox can run against a non-default PostgreSQL schema, and in those deployments this will return the wrong result (or nothing at all) even though the connection itself is fine.

Could we switch these filters to use the active schema (`current_schema()` or similar) rather than assuming `public`?

netbox/core/views.py
```diff
@@ -686,20 +686,102 @@ def get(self, request):
...
+                        'nullable': is_nullable == 'YES',
+                        'default': column_default,
+                    })
+
+                cursor.execute("""
+                    SELECT tablename, indexname, indexdef
+                    FROM pg_indexes
+                    WHERE schemaname = 'public'
+                    ORDER BY tablename, indexname
+                """)
```
pheus: Same as above.

pheus (changes requested):
Thanks for putting this together. This looks amazing! I’ve left a few small UI suggestions inline, but overall this looks great.

One thing I do think we should address before merge is that `_get_db_schema()` currently hardcodes `table_schema = 'public'` / `schemaname = 'public'`. That means the new tab and export could miss the actual NetBox tables for deployments using a non-default PostgreSQL schema. Since this view is meant to reflect the live schema during troubleshooting, could we make those queries use the active schema instead of assuming `public`?

As a follow-up, I think HTMX could be a really nice fit for this panel. Right now we build `db_schema` on every `/core/system/` request, even if the Database tab is never opened. Lazy-loading the panel on first activation would avoid that work unless the user actually opens it, while still keeping the data fresh. I’d keep the current eager behavior for `?export=true`, since the feature request calls for including the full schema in the export.

arthanson:
@pheus added in the HTMX for the database tab.

netbox/templates/core/htmx/system_db_schema.html
```diff
@@ -0,0 +1,128 @@
...
+              <h5 class="accordion-header" id="table-heading-{{ group.name }}-{{ forloop.counter }}">
+                <button class="accordion-button border-bottom collapsed font-monospace" type="button"
+                        data-bs-toggle="collapse" data-bs-target="#table-collapse-{{ group.name }}-{{ forloop.counter }}"
+                        aria-expanded="false" aria-controls="table-collapse-{{ group.name }}-{{ forloop.counter }}">
+                  {{ table.name }}
+                  <span class="badge bg-secondary text-white ms-2">{{ table.columns|length }} {% trans "columns" %}</span>
+                  {% if table.indexes %}
+                    <span class="badge bg-secondary text-white ms-1">{{ table.indexes|length }} {% trans "indexes" %}</span>
+                  {% endif %}
+                  <span class="accordion-button-toggle"><i class="mdi mdi-chevron-down"></i></span>
```
pheus, line 58: _One small UI nit I noticed_: the nested accordion chevron still looks a little off-center when it rotates. Since this seems to come from the shared `.accordion-button-toggle` styling rather than this specific template, I think a small global SCSS tweak in `project-static/styles/overrides/_tabler.scss` to center the icon inside the toggle box would be a nice cleanup.

```css
// Center the accordion chevron inside Tabler's toggle box.
.accordion-button-toggle {
  align-items: center;
  justify-content: center;
  flex: 0 0 auto;

  > i {
    display: block;
    line-height: 1;
  }
}
```

pheus (changes requested):
Thanks again for iterating on this! I really appreciate the HTMX lazy-loading in this PR. 

The one thing I’d still like to see adjusted before merge is moving `get_db_schema()` / `get_db_schema_groups()` out of `core/views.py` and into `core/utils.py`. Functionally this looks good, but I think those helpers would fit more naturally alongside the other core utility helpers there and keep `views.py` a bit easier to scan.

I also left one small inline UI note, but otherwise this is looking really good. Thanks again for iterating on this.

netbox/templates/core/htmx/system_db_schema.html
arthanson: Added this.

pheus (approved):
Nice work! Thanks!

Issue #21783, opened 2026-03-30
Cable bulk csv import powerfeed to powerport


WGBSP:
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.1

### Python Version

3.12

### Steps to Reproduce

1. Create power panel "DB123"
2. Create powerfeed "1"
3. Create Main Circuit Breaker device "MCB1"
4. Create power port "Input"
5. Create CSV file side_a_device,side_a_type,side_a_name,side_b_device,side_b_type,side_b_name
DB123,dcim.powerfeed,1,MCB1,dcim.powerport,Input
6. Attempt import.  It will fail with "Record 1 side_a_device: Object not found: DB123" even though it is there, this is because dcim.powerfeed is not a device type.  This was reported by a user in #18008 but incorrectly dismissed.

### Expected Behavior

When creating a CSV with termination_types of dcim.powerfeed and dcim.powerport, they should be able to be connected as they are both valid options. 

### Observed Behavior

Action fails with error "Record 1 side_a_device: Object not found: DB123" .  even though the power panel is there and correctly named.

PR #21903, opened 2026-04-13
Release v4.5.8

jeremystretch:
### Enhancements

* [#21430](https://github.com/netbox-community/netbox/issues/21430) - Display the device role's color in the device view
* [#21795](https://github.com/netbox-community/netbox/issues/21795) - Update `humanize_speed` template filter to support decimal Gbps/Tbps values

### Bug Fixes

* [#21529](https://github.com/netbox-community/netbox/issues/21529) - Exclude non-existent custom fields from object changelog data returned via the REST API
* [#21542](https://github.com/netbox-community/netbox/issues/21542) - Expand interface speed field to 64-bit integer to prevent overflow for LAG interfaces exceeding ~2.1 Tbps
* [#21704](https://github.com/netbox-community/netbox/issues/21704) - Fix missing port mappings in device type YAML export
* [#21783](https://github.com/netbox-community/netbox/issues/21783) - Fix support for bulk import of cables connected to power feeds
* [#21801](https://github.com/netbox-community/netbox/issues/21801) - Prevent duplicate filename collision when uploading files using S3 storage
* [#21814](https://github.com/netbox-community/netbox/issues/21814) - Fix custom script "last run" time to reflect job start time rather than creation time
* [#21835](https://github.com/netbox-community/netbox/issues/21835) - Correct help text for color selection form fields
* [#21841](https://github.com/netbox-community/netbox/issues/21841) - Restore visibility of the edit button for script modules to non-superusers
* [#21845](https://github.com/netbox-community/netbox/issues/21845) - Fix CSV export of connection columns rendering template whitespace instead of a formatted value
* [#21869](https://github.com/netbox-community/netbox/issues/21869) - Remove redundant `ScriptModule` class synchronization triggered on save


base_requirements.txt
```diff
@@ -55,7 +55,8 @@ django-storages
 
 # Abstraction models for rendering and paginating HTML tables
 # https://github.com/jieter/django-tables2/blob/master/CHANGELOG.md
-django-tables2
+# See #21902 for upgrading to django-tables2 v3.0+
+django-tables2<3.0
```
pheus, line 60: While looking at #21902, I've noticed that `django-tables2` v2.9 does introduce breaking changes, which should have been in v3.0 only. So we should limit the version to `django-tables2<2.9`

jeremystretch: IIRC the v2.9 release was yanked after it was determined to be "breaking," but yeah no harm in being explicit here.

pheus (approved):
Thanks!

PR #21905, opened 2026-04-13
#21361 Expand unit tests for ObjectChange and testing asserts

arthanson:
### Fixes: #21361 

Adds a new assert `assertObjectChangeData` and checks this in the standard API/View test cases.  It is just extra assert so the overhead is minimal.  The new assertions added to every ObjectChange verification block across views.py and api.py check:

* **CREATE** - prechange_data is None, postchange_data is populated
* **UPDATE** - both fields are populated, and they differ from each other
* **DELETE** - prechange_data is populated, postchange_data is None

This applies to single-object operations (create, edit, delete) and bulk operations (bulk import, bulk edit, bulk delete) in both the UI view tests and API tests.

The primary bug class caught: any code path where snapshot() is not called before a modification, which would leave prechange_data = None on an UPDATE or DELETE record.

### DIFF
--- a/netbox/utilities/testing/api.py
+++ b/netbox/utilities/testing/api.py
@@ -254,6 +254,7 @@ class APIViewTestCases:
                     action=ObjectChangeActionChoices.ACTION_CREATE,
                 )
                 self.assertEqual(objectchange.message, data['changelog_message'])
+                self.assertObjectChangeData(objectchange, prechange_data=None, postchange_data=True)
 
         def test_bulk_create_objects(self):
             """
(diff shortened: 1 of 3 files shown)


arthanson:
@claude /review

arthanson:
re claude review: fixed #1, #2 - others are being pedantic.

jeremystretch (commented):
This feels like a half-measure IMO: It's odd that we'd have a dedicated method solely to evaluate pre-/post-change data and not other commonly tested attributes of a change record. Have you considered expanding the method to also facilitate testing e.g. the action and changelog message?

Personally, I'd prefer we either implement a convenience method to test all of the prudent attributes, or skip it and test each individually with the built-in assertions as we do currently for `action` and `message`.

jeremystretch (approved):

PR #21922, opened 2026-04-15
Closes #20881: `get_filterset_for_model()` should reference application registry

### LINKED ISSUE #20881: Update FilterSetClassesTestCase to utilize filterset register
### Proposed Changes

Update the `get_filterset_for_model()` method under FilterSetClassesTestCase to return the filterset registered for a given model, rather than relying on obtuse name-based resolution.

Filterset registration was introduced as part of [NB-15](https://linear.app/netboxlabs/issue/NB-15/extend-filter-forms-to-support-filter-modifiers) and will ship in NetBox v4.5.0.

### Justification

This approach ensures more reliable filter registration, and will uncover unregistered filtersets.

jeremystretch:
### Closes: #20881

- Update `get_filterset_for_model()` on FilterSetClassesTestCase to reference the application registry when resolving the filter set for a given model instead of assuming its class name

### DIFF
--- a/netbox/netbox/tests/test_base_classes.py
+++ b/netbox/netbox/tests/test_base_classes.py
@@ -45,6 +45,7 @@ from netbox.graphql.types import (
     PrimaryObjectType,
 )
 from netbox.models import NestedGroupModel, NetBoxModel, OrganizationalModel, PrimaryModel
+from netbox.registry import registry
 from netbox.tables import (
     NestedGroupModelTable,
     NetBoxTable,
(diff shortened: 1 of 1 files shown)


jeremystretch:
@claude /review This is a simple PR. Just call out any concerns.

jeremystretch:
Addressed the minor concern noted above by raising an explicit error if a filterset cannot be resolved for a model.

pheus (approved):

PR #21925, opened 2026-04-15
Closes #21688: Improve performance of profiled Cable Path Tracing

pheus:
### Fixes: #21688

Improve profiled cable path tracing by adding `get_peer_terminations()` to resolve multiple peer terminations in a batched query, instead of issuing one lookup per `(termination, position)` pair.

`CablePath.from_origin()` is updated to use the batched lookup path, reducing N+1 queries in this hot path while preserving existing tracing behavior. This should especially help multi-position and multi-connector profiled cables.

### DIFF
--- a/netbox/dcim/cable_profiles.py
+++ b/netbox/dcim/cable_profiles.py
@@ -1,8 +1,13 @@
+from collections import defaultdict, namedtuple
+
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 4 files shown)


pheus:

pheus:

jeremystretch (approved):
Nice work!

PR #21930, opened 2026-04-15
Closes #21929: Eliminate redundant object view templates

### LINKED ISSUE #21929: Eliminate redundant object view templates
### Proposed Changes

Identify all object view templates which inherit from `object.html` without extending any blocks. Remove them, and alter their views to use the base template directly:

```python
class FooView:
    template_name = 'generic/object.html'
```

### Justification

Since the introduction of declarative layouts, a number of object views no longer have any need for custom templates. We can eliminate the redundancy by rendering the base generic template directly.

jeremystretch:
### Closes: #21929

- Introduce the `get_default_template()` utility function to safely resolve the conventional template name for a model
- Update several feature views to employ `get_default_template()`
- Delete all "empty" object view templates
- Update the applicable view classes to render `generic/object.html` directly

### DIFF
--- a/netbox/dcim/views.py
+++ b/netbox/dcim/views.py
@@ -4060,6 +4060,7 @@ class InventoryItemRoleListView(generic.ObjectListView):
 @register_model_view(InventoryItemRole)
 class InventoryItemRoleView(GetRelatedModelsMixin, generic.ObjectView):
     queryset = InventoryItemRole.objects.all()
+    template_name = 'generic/object.html'
     layout = layout.SimpleLayout(
         left_panels=[
             panels.InventoryItemRolePanel(),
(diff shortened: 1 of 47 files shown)


jeremystretch:
@claude /review check for any inconsistencies or missing bits. Adopt the persona of Homer Simpson in your response.

jeremystretch:
@claude /review Highlight any inconsistencies or gaps

jeremystretch:
Fixed the above bug concerning ObjectChildrenView

arthanson (changes requested):
Can we remove all the assignments of ` template_name = 'generic/object.html'` and just put it into `netbox.views.generic.object_views.ObjectView` it should still work the same and would be consistent with the other Generic object views and less boilerplate with the new `get_default_template`

Also the doc comment on netbox.views.generic.object_views.ObjectView needs to be changed, currently

```
    Note: If `template_name` is not specified, it will be determined automatically based on the queryset model.
```

With the new logic the is reversed, it will always try to check if the {app}/{model} template exists first and then fallback to the template_name.

jeremystretch:
@arthanson That would potentially break plugin usage. We'll probably head that way eventually, but for now we have to be very careful not to alter the base view's behavior.

arthanson (approved):

PR #21932, opened 2026-04-15
21782 Enable optional config template override in URL

### LINKED ISSUE #21782: Enable optional config template selection on Device "Render Config"
### NetBox version

v4.5.3

### Feature type

### Proposed functionality

The proposal here is to read the url for a variable that specifies a config template to use; if none is provided, default to current behaviour, i.e. use the template assigned to the device/role/platform.  However, if a config template is specified, use that one instead.  This behaviour should also apply to the APIs as well.

/dcim/devices/10185/render-config/
/dcim/devices/10185/render-config/?config_template_id=23

This would require the following changes to ObjectRenderConfigView::get_extra_context() to read the request object and, if a template is specified in the request, use that rather than the default one attached to the device instance.

        # Compile context data
        context_data = instance.get_config_context()
        context_data.update(self.get_extra_context_data(request, instance))

        # Render the config template
        rendered_config = None
        error_message = ''

        # if an optional config template has been specified, use that
        if config_template_id := request.GET.get('config_template_id'):
            config_template = ConfigTemplate.objects.get(pk = config_template_id)
        else:
            config_template = instance.get_config_template()

        if config_template:
            try:
            except TemplateError as e:
                error_message = _("An error occurred while rendering the template: {error}").format(error=e)

        return {
            'base_template': self.base_template,
            'config_template': config_template,
            'context_data': context_data,
            'rendered_config': rendered_config,
            'error_message': error_message,
        }

Additionally, changes required to the UI are required to allow the user to select which template to use.  Default behaviour could remain as is, selecting the Render Config tab renders the default config template; a simple form could be included on the Render tab to allow the user to optionally select a different config template.

### Use case

Sometimes it is desirable to be able to apply a partial config template to a device context and get the resulting rendered config, for example, changing only ntp server configs where the variables depend on the region that the device is in.  This is especially needed where the modelled network does not equal the actual network and so the fully rendered configs cannot be blindly applied, where a partial config can be (as in the ntp change example).

This will typically be used to drive network automations to close network drift, e.g. the ntp changes above would be driven by Ansible automations that retrieve a device config from netbox by specifying both the device and the template to use.

### Database changes

None required

### External dependencies

_No response_

arthanson:
### Fixes: #21782 

Adds the ability to override the config template used for rendering.  For the UI you can tack on the query param `config_template_id` to override:

```
http://127.0.0.1:8000/dcim/devices/27/render-config/?config_template_id=2
```

In the API you need to pass it as a data param (see below)

Can use the following curl commands to check (update device, token, etc..) also need a config template (`Provisioning > Config Template`) setup.

```
curl -s -X POST http://127.0.0.1:8000/api/dcim/devices/27/render-config/ \
    -H "Authorization: Bearer nbt_<TOKEN>.<TOKEN>" \
    -d '{}' | python3 -m json.tool
```

And to test the PR override with template ID:

```
  curl -s -X POST http://127.0.0.1:8000/api/dcim/devices/27/render-config/ \
    -H "Authorization: Bearer nbt_<TOKEN>.<TOKEN>" \
    -d '{"config_template_id": 3}' | python3 -m json.tool
```


arthanson:
@claude /review

arthanson:
claude review - added documentation line re the UI.

bctiemann (approved):

PR #21939, opened 2026-04-16
Fixes #21895 Log output for job limited to a certain number of rows

### LINKED ISSUE #21895: Log output for job limited to 50 rows
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.7

### Python Version

3.12

### Steps to Reproduce

1. Run a job with lots of log messages
2. Navigate to Operations -> Jobs -> (job in question) -> Logs
3. Scroll to bottom of page

### Expected Behavior

At the bottom of the page there should be a current page number, a "showing (x) of (y)", and a Per Page dropdown so the user can view the rest of the logs for that job.

Image

### Observed Behavior

None of those options exist, nor does the bar where you would normally find them.  Current workaround is to append "?page=(x)" to the end of the URL.

Image

sergiolpzgmz:
Fixes #21895

Include pagination controls in `context_table.html`

### DIFF
--- a/netbox/core/views.py
+++ b/netbox/core/views.py
@@ -39,9 +39,9 @@ from netbox.plugins.utils import get_installed_plugins
 from netbox.ui import layout
 from netbox.ui.panels import (
     CommentsPanel,
-    ContextTablePanel,
     JSONPanel,
     ObjectsTablePanel,
+    PaginatedContextTablePanel,
     PluginContentPanel,
     RelatedObjectsPanel,
     TemplatePanel,
@@ -267,7 +267,7 @@ class JobLogView(generic.ObjectView):
     layout = layout.Layout(
         layout.Row(
             layout.Column(
-                ContextTablePanel('table', title=_('Log Entries')),
+                PaginatedContextTablePanel('table', title=_('Log Entries')),
                 PluginContentPanel('left_page'),
             ),
         ),
--- a/netbox/netbox/ui/panels.py
+++ b/netbox/netbox/ui/panels.py
@@ -20,6 +20,7 @@ __all__ = (
     'ObjectPanel',
     'ObjectsTablePanel',
     'OrganizationalObjectPanel',
+    'PaginatedContextTablePanel',
     'Panel',
     'PluginContentPanel',
     'RelatedObjectsPanel',
@@ -425,3 +426,11 @@ class ContextTablePanel(ObjectPanel):
         if table is None:
             return ''
         return super().render(context)
+
+
+class PaginatedContextTablePanel(ContextTablePanel):
+    """
+    A panel which extends ContextTablePanel to also include
+    pagination controls for the rendered table.
+    """
+    template_name = 'ui/panels/paginated_context_table.html'
--- a/netbox/templates/ui/panels/paginated_context_table.html
+++ b/netbox/templates/ui/panels/paginated_context_table.html
@@ -1,0 +1,7 @@
+{% extends "ui/panels/_base.html" %}
+{% load render_table from django_tables2 %}
+
+{% block panel_content %}
+  {% render_table table 'inc/table.html' %}
+  {% include 'inc/paginator.html' with paginator=table.paginator page=table.page %}
+{% endblock panel_content %}


pheus (changes requested):
Thanks for working on this. I agree the missing pagination on the job log page should be fixed, but I don't think `ui/panels/context_table.html` is the right place to do it.

`ContextTablePanel` is a shared declarative component that’s already used in several other views, and it currently just renders the table it is given. Adding paginator controls there changes the behavior of that reusable component globally to address one specific page.

Could we revisit this with a more thoughtful implementation that keeps `ContextTablePanel` generic and handles pagination specifically for the job log view, or introduces a dedicated paginated variant if we want this pattern more broadly?

sergiolpzgmz:
Hi @pheus ,

Thank you for your review. You’re  right; I agree that my initial proposal isn’t correct I shouldn’t be modifying the behavior of such a generic view. 

After reviewing it again and looking at other examples that have already been implemented, what do you think of this as a possible solution: editing core/job/log.html with something like this?

~~~html
{% extends 'core/job/base.html' %}
{% load render_table from django_tables2 %}

{% block content %}
  <div class="row mb-3">
    <div class="col col-md-12">
      <div class="card">
        <div class="card-body table-responsive">
          {% render_table table 'inc/table.html' %}
          {% include 'inc/paginator.html' with paginator=table.paginator page=table.page %}
        </div>
      </div>
    </div>
  </div>
{% endblock %}
~~~

When you mention creating a variant with dedicated pagination, do you mean, for example, sending a flag variable from get_extra_context() and, based on that, loading the paginated or normal variant?

Once again, thank you for your feedback and your time. This is my second contribution to the project, and I still have a lot to learn, but these comments are really helpful!

pheus:
Thanks for the follow-up, and for taking the feedback on board.

Since you're working on this issue, I'd like you to take the lead on finding a suitable implementation. I don't want to prescribe the exact solution in too much detail here, otherwise I'd be moving beyond review and into implementing the fix for you.

From a first look, your latest suggestion might work, but that depends on the details. When I mentioned a dedicated variant, I didn’t mean a specific mechanism such as passing a flag from `get_extra_context()`; the key point is that pagination should be enabled deliberately where needed, rather than changing the behavior of a generic component globally. One possible direction would be to extend `ContextTablePanel` with an explicit opt-in attribute for pagination.

Please take another pass with that in mind, and I’ll be happy to review the next iteration.

pheus (changes requested):
Thanks for the update. I still think making pagination opt-in is the right direction, but this implementation is not safe yet.

`JobLogView` is a pluggable page and also renders `PluginContentPanel('left_page')`, so other tables can exist on the same page. Reusing the shared paginator there still drives the generic `page` / `per_page` query parameters, which can collide with other paginated tables on the page. That can lead to another table picking up the same `page` value and raising `EmptyPage`.

I also don’t think we should introduce a separate paginated panel/template if we can avoid it. This feels like duplicate code for what is really just a small behavioral variation. I’d much rather see this handled by extending the existing component with an explicit opt-in, e.g. `ContextTablePanel('table', title=_('Log Entries'), pagination=True)`.

So I don’t think this is ready yet. The pagination needs to be isolated from any other table on the page, and the implementation should stay simple and avoid duplicating panel or template logic.

sergiolpzgmz:
> Thanks for the update. I still think making pagination opt-in is the right direction, but this implementation is not safe yet.
> 
> `JobLogView` is a pluggable page and also renders `PluginContentPanel('left_page')`, so other tables can exist on the same page. Reusing the shared paginator there still drives the generic `page` / `per_page` query parameters, which can collide with other paginated tables on the page. That can lead to another table picking up the same `page` value and raising `EmptyPage`.
> 
> I also don’t think we should introduce a separate paginated panel/template if we can avoid it. This feels like duplicate code for what is really just a small behavioral variation. I’d much rather see this handled by extending the existing component with an explicit opt-in, e.g. `ContextTablePanel('table', title=_('Log Entries'), pagination=True)`.
> 
> So I don’t think this is ready yet. The pagination needs to be isolated from any other table on the page, and the implementation should stay simple and avoid duplicating panel or template logic.

Thanks for the feedback, your time and patience. I've implemented the opt-in approach you suggested (pagination=True), but I'm stuck on isolating the query strings.

Here's what I've found:
The shared `inc/paginator.html` uses `{% querystring request page=... %}` `with page` and `per_page` hardcoded, so it doesn't respect the table's prefix. Even when I set `prefix='log-'` on the table, the pagination links still generate ?page=2 instead of ?log-page=2.
I've considered a few options:
1. Make `paginator.html` prefix-aware pass the table and use `table.prefixed_page_field`. I'm not comfortable making this change to a shared template.
2. Duplicate pagination in `context_table.html` use django-tables2's {% querystring %} which supports `table.prefixed_page_field`. This works, but it's duplicating code.
3. Modify the querystring tag in `helpers.py` to accept an optional table parameter and auto-use prefixed field names when provided. Backward compatible I think, but still a core change.
4. Create a new template tag like `{% prefixed_querystring table page=p %}`. Isolated change, but adds new code.

I'd really appreciate your guidance on which approach you'd prefer.

PR #21940, opened 2026-04-16
Fixes #21935: Document MAX_PAGE_SIZE effect on GraphQl

### LINKED ISSUE #21935: GraphQL queries are now limited to MAX_PAGE_SIZE
### Change Type

Correction

### Area

Integrations/API

### Proposed Changes

Followup of [NB-2672](https://linear.app/netboxlabs/issue/NB-2672/graphql-list-queries-only-show-the-first-1000-entries)

v4.5.4 and before:<br>Basic GraphQL list queries were unlimited because MAX_PAGE_SIZE had no effect.

v4.5.5 and later:<br>Basic GraphQL queries (even without pagination) are limited to MAX_PAGE_SIZE

The documentation still refers only the REST API to MAX_PAGE_SIZE:

* [https://netboxlabs.com/docs/netbox/configuration/miscellaneous/#max_page_size](<https://netboxlabs.com/docs/netbox/configuration/miscellaneous/#max_page_size>)
* [https://netboxlabs.com/docs/netbox/best-practices/performance-handbook/](<https://netboxlabs.com/docs/netbox/best-practices/performance-handbook/>)

MAX_PAGE_SIZE completely missing from GraphQL docs:

* [https://netboxlabs.com/docs/netbox/integrations/graphql-api/](<https://netboxlabs.com/docs/netbox/integrations/graphql-api/>)
* But present here: [https://netboxlabs.com/docs/netbox/integrations/rest-api/](<https://netboxlabs.com/docs/netbox/integrations/rest-api/>)

Please update the documentation to reflect that MAX_PAGE_SIZE also applies to GraphQL now.

grische:
### Fixes: #21935

Documenting the effect of MAX_PAGE_SIZE on GraphQL introduced in #21617

I highlighted in particular that REST API's `?limit=0` has a different behaviour than GraphQL's `limit: 0`.

### DIFF
--- a/docs/best-practices/performance-handbook.md
+++ b/docs/best-practices/performance-handbook.md
@@ -34,7 +34,7 @@ NetBox ships with a reasonable default configuration for most environments, but 
 
 #### Reduce the Maximum Page Size
 
-NetBox paginates large result sets to reduce the overall response size. The [`MAX_PAGE_SIZE`](../configuration/miscellaneous.md#max_page_size) parameter specifies the maximum number of results per page that a client can request. This is set to 1,000 by default. Consider lowering this number if you find that API clients are frequently requesting very large result sets.
+NetBox paginates large result sets to reduce the overall response size. The [`MAX_PAGE_SIZE`](../configuration/miscellaneous.md#max_page_size) parameter specifies the maximum number of results per page that a client can request. This is set to 1,000 by default. Consider lowering this number if you find that API clients are frequently requesting very large result sets. `MAX_PAGE_SIZE` applies to both the REST API (`?limit=`) and the GraphQL API (`pagination: {limit: …}`), so lowering it reduces the maximum size of responses from either API.
 
 #### Limit GraphQL Aliases
 
@@ -185,3 +185,5 @@ Like the REST API, the GraphQL API supports pagination. Queries which return a l
   }
 }
 ```
+
+The requested `limit` is capped by [`MAX_PAGE_SIZE`](../configuration/miscellaneous.md#max_page_size).
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -45,7 +45,7 @@ Sets content for the top banner in the user interface.
 
 !!! tip
     If you'd like the top and bottom banners to match, set the following:
-    
+
     ```python
     BANNER_TOP = 'Your banner text'
     BANNER_BOTTOM = BANNER_TOP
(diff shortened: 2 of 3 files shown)


pheus (changes requested):
Thanks for working on this.

I think the changes in `docs/configuration/miscellaneous.md` go a bit beyond what is needed here. The configuration docs should explain the purpose and implications of the parameter, but they do not need to cover every possible combination. I’d prefer to keep the core text mostly unchanged and add a short GraphQL pagination example, along with any notable differences. As it stands, the section feels a bit too verbose.

The same applies to `docs/integrations/graphql-api.md`. The existing text already notes that “If no limit is specified, a default value of 100 is used,” so I don’t think that needs to be repeated in `docs/configuration/miscellaneous.md`. I’m also not sure it is helpful to document all possible combinations there, as those are more advanced details and may confuse regular users.

grische:
You're right, I am duplicating info on the miscellaneous.md. I drastically reduced this and I am pretty happy with the new version.

I tried to reword the graphql-api.md to make it more streamline, but I did not remove the bullet list for the MAX_PAGE_SIZE=0 case. Personally, I think it is crucial to be able to look up this edge case as it behaves slightly different than the REST API.

What do you think about the new version?

pheus (changes requested):
Thanks for the update! This is much closer.

I still think we should keep the fact that `MAX_PAGE_SIZE` also affects the Web UI. In `docs/configuration/miscellaneous.md`, the new wording reads as if it only applies to the REST and GraphQL APIs, but this setting also limits the table paginator in the UI.

Could we tweak that sentence so it keeps the broader scope, and leave the REST/GraphQL-specific details to the integration docs? That would keep the configuration docs concise while still being accurate.

grische:
Removing the Web UI was an oversight, I will add it back.
I am not sure I understand the "and leave the REST/GraphQL-specific details to the integration docs".
Is this what you are looking for?

```diff
diff --git a/docs/configuration/miscellaneous.md b/docs/configuration/miscellaneous.md
index 9a512a9c4..382cc8620 100644
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -171,7 +171,7 @@ This specifies the URL to use when presenting a map of a physical location by st
 
 Default: `1000`
 
-Defines the maximum number of objects that may be returned in a single page for both the REST API and the GraphQL API. Setting `MAX_PAGE_SIZE` to `0` or `None` removes the limit.
+Defines the maximum number of objects that may be returned in a single page across the web UI, REST API, and GraphQL API. Setting `MAX_PAGE_SIZE` to `0` or `None` removes the limit.
 
 See the [REST API](../integrations/rest-api.md#pagination) and [GraphQL API](../integrations/graphql-api.md#pagination) pagination documentation for details.
```

pheus (approved):
Thanks for the update! This addresses my earlier concern.

The wording in `docs/configuration/miscellaneous.md` now makes it clear that `MAX_PAGE_SIZE` also applies to the Web UI, which was the main thing I wanted to preserve. I think the split is much better now: the configuration docs stay concise, while the GraphQL-specific behavior is documented in the integration docs.

From my side, this looks good now.

PR #21954, opened 2026-04-17
Fixes #21870: Optimize prefix view

Miyoshi-Ryota:
### Fixes: #21870

Optimize the prefix view for better responsiveness in large-scale environments. It speeds up prefix utilization calculations when no child ranges are marked as utilized. It also caches available IP calculations on the model to avoid duplicate work during rendering. Finally, it lazy-loads the Addressing panel so the prefix detail page can render sooner.


netbox/ipam/tests/test_models.py
```diff
@@ -290,6 +290,7 @@ def test_get_first_available_ip(self):
         self.assertEqual(parent_prefix.get_first_available_ip(), '10.0.0.4/24')
 
         IPAddress.objects.create(address=IPNetwork('10.0.0.4/24'))
+        parent_prefix = Prefix.objects.get(pk=parent_prefix.pk)
         self.assertEqual(parent_prefix.get_first_available_ip(), '10.0.0.5/24')
```
Miyoshi-Ryota, line 294: I believe the change in behavior caused by adding the cache is acceptable for the reasons described in the commit message below. 
https://github.com/netbox-community/netbox/pull/21954/commits/8253d5008cc46b2b65093bd4e9e6eb9546447e5c

However, if this is not acceptable to you, please let me know and I will revert the changes

Miyoshi-Ryota:
### Appendix: optimization results
#### /ipam/prefixes/:
* The page load now takes approximately 2–3 seconds (previously 5+ minutes).
スクリーンショット 2026-04-17 23 07 29

#### /ipam/prefixes/<id of 10.0.0.0/8>: 
* The initial load now takes less than 1 second (previously 5+ minutes).
スクリーンショット 2026-04-17 23 07 38

* The entire page finishes loading in about 1.2 minutes (previously 5+ minutes).
スクリーンショット 2026-04-17 23 08 51

pheus:

pheus:
Thanks for opening this PR!

Miyoshi-Ryota:
@pheus 
Thank you! 

I have fixed points 1 and 3. Regarding points 2 and 4, I believe the current implementations are acceptable. I’d like to keep them as they are because:

-----------
#### 2. cached_property behavior change (ip.py:442–478)
I mentioned this in commit 8253d50, but I will describe it with more detail here.

- Impact on existing code
I (and Codex) reviewed the repository and did not find any existing code affected by this change. Since most objects only live during a single request/response cycle, the "cache -> modify -> read stale cache" scenario is unlikely to occur in the current codebase.

- Consistency with existing logic
Existing parts of the code, such as IPRange.first_available_ip, already use this same caching behavior. Because it is possible to produce stale results manually, this change is consistent with the current existing patterns.

e.g. `./manage.py shell` then paste following code.
```python
from netaddr import IPNetwork
from ipam.models import IPRange, IPAddress

iprange = IPRange.objects.create(
    start_address=IPNetwork("198.51.100.10/24"),
    end_address=IPNetwork("198.51.100.12/24"),
)

# 198.51.100.10/24
print(iprange.first_available_ip)

IPAddress.objects.create(address=IPNetwork("198.51.100.10/24"))

# Still 198.51.100.10/24 (stale)
print(iprange.first_available_ip)

# 198.51.100.11/24
print(IPRange.objects.get(pk=iprange.pk).first_available_ip)

# cleanup
IPAddress.objects.filter(address=IPNetwork("198.51.100.10/24")).delete()
iprange.delete()
```

- Performance benefits
In environments with a large number of IPs (e.g., 1 million), this caching improves performance by approximately 1.2 minutes. I believe this performance gain is more valuable than the risk of stale data in these specific edge cases.




#### 4. HTMX partial — no loading state after initial render
This is intended behavior, and acceptable. Showing older content after using the browser back button is not new with this change. NetBox already behaves like that in other cases.

For example:
- Tab 1 opens the 10.0.0.0/8 prefix page.
- Tab 2 edits the description of the same prefix.
- Then tab 1 goes to another page and comes back with the browser back button.
- Tab 1 can still show the old description.

I think this Addressing panel is the same kind of browser history behavior, not a new issue caused by this change.

pheus (changes requested):
Thanks for working on this! The overall direction makes sense.

That said, I still have a few blocking concerns with the current shape of the change.

1. The lazy-loading part feels too ad hoc for the declarative UI work. Right now this is implemented as a Prefix-specific HTMX branch in the view plus HTMX transport markup embedded directly in the panel template. I think this would be better introduced as a reusable panel-level capability (for example, lazy/deferred rendering on `TemplatePanel`) instead of a one-off just for this panel.

2. Related to that, I don’t think the `hx-*` wrapper belongs in `ipam/panels/prefix_addressing.html`. That template should ideally describe panel content, not transport/loading behavior. If we want deferred rendering here, I think the wrapper should live one layer up, or the panel framework should gain first-class support for it.

3. I’m also still uneasy about caching available IPs on the model instance. That changes the freshness semantics of `get_available_ips()` / `get_first_available_ip()` outside this specific detail-page render path, and the updated test needing to re-fetch the `Prefix` is a good indication that behavior has changed.

4. Looking at the actual optimization, the substantive algorithmic win here is fairly narrow: the new `get_utilization()` fast path applies when there are no child ranges marked as utilized. Without the model cache, the expensive available-IP path is still computed twice for the Addressing panel; with lazy loading, that work is mostly deferred rather than fundamentally solved.

5. The appendix reinforces that impression for me. The initial prefix detail page becomes much faster, but the full page still takes a long time to finish loading. So this reads more like a fine tuning of the no-utilized-range case plus deferred rendering, rather than a broader fix for the general large-prefix addressing bottleneck.

Because of that, my preference would be to keep the `get_utilization()` improvement, avoid changing model-helper freshness semantics, and, if we want lazy panel rendering here, add it at the panel/declarative-UI layer rather than baking a Prefix-specific HTMX path into this view/template pair.

Miyoshi-Ryota:
Thank you for the review!

Quick update:
- I'll revert the cache changes.
- For lazy load, I'm looking for a better way to implement it.
  - I'm thinking of something like this **for now**.:
    - Add `class LazyLoadTemplatePanel(TemplatePanel):` or something like a those panel class to netbox/netbox/ui/panels/panels.py. (But)
    - Use it in `PrefixView`.
    - (I’m trying some things in the code to find a good implementation, but it doesn't feel right yet—especially regarding the views to handling the hx-* request to get the addressing panel content. So, the approach might change.)

Miyoshi-Ryota:
### Appendix: Images of Behavior After Fix

#### During Loading
スクリーンショット 2026-04-28 1 21 54

#### After Loading Completes
スクリーンショット 2026-04-28 1 24 06

netbox/netbox/views/generic/object_views.py
```diff
@@ -78,14 +90,23 @@ def get(self, request, **kwargs):
...
-
-        return render(request, self.get_template_name(), {
+        context = {
             'object': instance,
             'actions': actions,
             'tab': self.tab,
             'layout': self.layout,
             **self.get_extra_context(request, instance),
-        })
+        }
```
Miyoshi-Ryota, line 99: The Claude might say that calling self.get_extra_context is a waste of time when htmx_partial is True.

It’s true that the Addressing panel doesn't need it right now. However, other panels in the future might need it. I think adding a "skip" option would make the code messy, so I decided to keep it simple and call it every time.

For PrefixView, this part is not heavy at all, so calling it twice won't cause any issues.

Miyoshi-Ryota:
@pheus 
I’ve updated the code based on the review feedback.
* Created a foundation to allow a Panel to be lazy-loaded.
* Applied this to lazy-load the Addressing panel in PrefixView.

Regarding the scope: I previously mentioned this would be limited to `TemplatePanel` or `LazyLoadTemplatePanel`. However, as I generalized the implementation, I realized that making all Panels lazy-loadable required almost the same amount of effort, and the code is much cleaner. Therefore, I decided to support Panels in general.
~~

### Memo for Review 

To generalize lazy loading, we need to address two key points: how to represent it in the declarative UI and where to handle the partial requests.

I have implemented them as follows:

* UI: Use LazyLoadPanel(some_panel_instance, partial_key='xxx') within the Layout.
* View: ObjectView now automatically detects if a Panel with partial_key exists in the Layout and returns only that Panel for partial request. LazyLoadPanel then fetches this content via an HTMX request after the initial page load.

jeremystretch:
@Miyoshi-Ryota thanks for your work on this. Unfortunately, I don't think the approach here sufficiently addresses the underlying performance concern. PR #22211 has also been submitted as a potential resolution for #21870, and aims to improve the underlying database performance for utilization calculations considerably. Although it hasn't yet been approved, it's very likely that we're going to adopt the strategy it proposes, so I'm going to close out this PR. Thanks again for taking a whack at this!

Miyoshi-Ryota:
@jeremystretch 
Thank you!

I've been following PR #22211, and I also think it's a much better solution than my PR.
( To be honest, I wanted to solve it with SQL COUNT like #22211 does, but I couldn't find out a way because there are two models involved ;( )

I really appreciate the work of @jeremystretch and @pheus. I and my users are looking forward to seeing #22211 merged and released!

PR #21967, opened 2026-04-20
Fixes #21906: Return a 404 for REST API writes to endpoints with no trailing slash

### LINKED ISSUE #21906: REST API `POST`/`PATCH` request without a trailing slash causes an exception
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.7

### Python Version

3.12

### Steps to Reproduce

Make a valid, authenticated `PATCH` request to e.g. `/api/dcim/devices/1` (no trailing slash on the URL):

```
curl -v -X PATCH \
http://netbox/api/dcim/devices/1
```

### Expected Behavior

I'm actually not sure. We can't automatically redirect (as noted in the error message below) without losing the request data. A 404 (Not Found) might be the most appropriate response.

### Observed Behavior

A `RuntimeError` exception is raised:

```
  File "/home/jstretch/.venv/netbox/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jstretch/.venv/netbox/lib/python3.12/site-packages/django/utils/deprecation.py", line 311, in __call__
    response = self.process_response(request, response)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jstretch/.venv/netbox/lib/python3.12/site-packages/django/middleware/common.py", line 109, in process_response
    self.get_full_path_with_slash(request)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jstretch/.venv/netbox/lib/python3.12/site-packages/django/middleware/common.py", line 87, in get_full_path_with_slash
    raise RuntimeError(
    ^

Exception Type: RuntimeError at /api/dcim/devices/1
Exception Value: You called this URL via PATCH, but the URL doesn't end in a slash and you have APPEND_SLASH set. Django can't redirect to the slash URL while maintaining PATCH data. Change your form to point to netbox:8000/api/dcim/devices/1/ (note the trailing slash), or set APPEND_SLASH=False in your Django settings.
```

jeremystretch:
### Fixes: #21906

Return a well-formed 404 response to `POST`/`PATCH`/`PUT`/`DELETE` REST API requests with no trailing slash on the endpoint. This resolves the current behavior of raising an error. (We opt not to return a 302, because a redirect would result in a `GET` request by the client.)

### DIFF
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -8,6 +8,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.db import ProgrammingError, connection
 from django.db.utils import InternalError
 from django.http import Http404, HttpResponseRedirect
+from django.middleware.common import CommonMiddleware as DjangoCommonMiddleware
 from django_prometheus import middleware
 
 from netbox.config import clear_config, get_config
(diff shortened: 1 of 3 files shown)


jeremystretch:
@claude /review Perform a thorough review but act mildly annoyed that you have to deal with this.

arthanson (approved):

PR #21969, opened 2026-04-20
Closes #21924: Refactor sticky bulk actions and form bars

### LINKED ISSUE #21924: Improve styling and consistency of floating bulk actions
### NetBox version

v4.5.8

### Feature type

### Proposed functionality

I’d like to propose a small refinement to the existing floating bulk-action UI for tabular views.

NetBox already supports floating bulk actions, but the current behavior and styling are a bit inconsistent across templates. This proposal is to keep the existing functionality, while improving the presentation and making the front-end implementation more explicit and less dependent on fragile DOM structure.

Concretely, this would:
- add an explicit marker on bulk-action containers in the relevant templates (for example, `data-bulk-float="left|right"`)
- update `floatBulk.ts` to toggle the existing `btn-float-group-left` / `btn-float-group-right` classes based on selection state within the same form
- keep `.bulk-action-buttons` as the HTMX swap target
- refine the sticky bar styling so spacing, alignment, background, and overlap behavior are consistent across views

This would be a UI/UX improvement to existing behavior only.

Screenshots:

EditForm:
Image

ListView (light):
Image

ListView (dark):
Image

### Use case

The current semi-transparent floating action buttons can lack contrast, allowing table content to show through and making the controls harder to read. Floating bulk actions are especially useful when working with long tables, but their behavior and presentation currently feel inconsistent between standard list views and child-object tables. Refining this UI would improve clarity and consistency for users who frequently perform bulk edits or deletes, while also making the front-end behavior easier to maintain.

### Database changes

None.

### External dependencies

None.

pheus:
### Fixes: #21924

This PR refines sticky action buttons across list and form views by introducing a shared, data-attribute-driven `sticky-actions` pattern.

Selection-based bulk action bars now use a stable wrapper and become active only when objects are selected, which avoids the previous visual jump while keeping the controls easy to reach. As part of this update, bulk action buttons in list views are moved from the left to the right for consistency with child-object views. Always-visible form action bars use the same mechanism with a full-width sticky footer presentation for create/edit-style forms.

This also cleans up the implementation by removing fragile DOM assumptions, dropping obsolete `.btn-float` usage from footer templates, styling disabled bulk-action buttons more consistently, and removing a redundant `{% csrf_token %}` from `object_list.html`.


pheus:

pheus:
1. The dual .bulk-action-buttons target is pre-existing and both render the same buttons, so the OOB swap keeps them in sync; no regression there.
2. The footer's dependency on .sticky-actions for position: sticky is a fair call; I've added a comment noting the required co-application.
3. The is-sticky-active branch in the footer is intentionally kept as defensive CSS for potential future selection-driven footers.
4. And the missing mt-5 on preferences is intentional: that page doesn't use a tab bar, so the spacing comes from the base layout instead.

bctiemann (approved):

PR #21974, opened 2026-04-20
Closes #21788: Return CSV export as a streaming response

### LINKED ISSUE #21788: Exporting large querysets leads to 504 timeout errors
### NetBox Version

v4.4.0

### Python Version

3.12

### Area(s) of Concern

- [x] User Interface
- [ ] REST API
- [ ] GraphQL API
- [X] Python ORM
- [ ] Other

### Details

For context, our environment has over 1 million device objects. When exporting objects as CSV (via the "Export" button on any list view), the response blocks entirely until all matching objects have been fetched, serialized, and buffered in memory before a single byte is sent to the client. For large tables (tens of thousands of objects), this results in the browser appearing to hang with no download activity, timeout errors before the response completes, and high peak memory consumption on the server. In addition, from a UX perspective, the end user is unable to navigate away from the page while the export is processing, and once the timeout is hit, the web server returns a 504 error while the query is still processing on the server.

Image

Image

As one potential solution, I switched the export functionality locally to use a `StreamingHttpResponse` so that the HTTP server isn't waiting on all rows to be loaded into memory. Instead, the export function fetches results in batches of 200 rows and streams the response to the client, yielding each row to the CSV writer as it's processed. This means the first CSV row reaches the client within milliseconds of the request, the end-user is able to navigate away from the page as the file downloads, memory stays bounded to ~200 rows at a time, and timeouts are no longer a concern regardless of table size.

jeremystretch:
### Fixes: #21788

- Introduce `stream_table_csv_response()` to replace TableExport
- Employ Django's StreamingHttpResponse to stream generated CSV content
- Update tests

### DIFF
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -123,6 +123,18 @@ The maximum size (in bytes) of an incoming HTTP request (i.e. `GET` or `POST` da
 
 ---
 
+## STREAMING_EXPORTS
+
+!!! note "This parameter was introduced in NetBox v4.6."
+
+Default: `False`
+
+When set to `True`, CSV bulk exports are returned as a streaming HTTP response, emitting rows to the client as they are rendered rather than buffering the entire dataset in memory first. This can significantly reduce memory usage and time-to-first-byte for very large exports.
+
+Because streaming responses do not have a `Content-Length` header and defer errors until after the response has begun, this behavior is opt-in.
+
+---
+
 ## ENFORCE_GLOBAL_UNIQUE
 
 !!! tip "Dynamic Configuration Parameter"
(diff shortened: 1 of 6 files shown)


jeremystretch:
@claude /review

jeremystretch:
@claude /review

jeremystretch:
1. I don't think this is a risk as the table itself correctly handles null values. Confirmed that an export still renders empty strings in place of `None`.
2. This is a well-understood trade-off.
3. Appears to be an incorrect observation. The [Django docs](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#iterator) confirm `prefetch_related()` support for `iterator()`:

> `iterator()` is compatible with previous calls to `prefetch_related()` as long as `chunk_size` is given. Larger values will necessitate fewer queries to accomplish the prefetching at the cost of greater memory usage.

I've also verified this manually by overriding the response with a normal HttpResponse and inspecting the underlying queries.

jeremystretch:
@claude /review

jeremystretch:
1. Addressed
2. Addressed
3. Leaving this out for now

arthanson (approved):

PR #21976, opened 2026-04-20
Fixes #21975: Prefetch all related data during CSV bulk export

### LINKED ISSUE #21975: CSV bulk export does not fully employ prefetching of related objects
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.8

### Python Version

3.12

### Steps to Reproduce

1. Enable debugging to capture SQL queries.
2. Navigate to a list view with at least a few dozen objects for illustration.
3. Click Export > All Data (CSV) and save or discard the file.
4. Open the debugging toolbar, select the history tab, and switch to the most recent request (ending in `?export`)
5. Open the toolbar's SQL panel and inspect the queries

### Expected Behavior

Data for all related objects should be prefetched, minimizing the number of SQL queries required to generate the response data.

### Observed Behavior

Only data for columns which are visible in the UI table support prefetching. Data for any hidden columns is fetched individually per row.

jeremystretch:
### Fixes: #21975

- Extend `_apply_prefetching()` on BaseTable to accept an optional list of columns to prefetch
- Tweak `export_table()` on ObjectListView to pass a list of all relevant columns when calling `_apply_prefetching()`
- Add a test

### DIFF
--- a/netbox/netbox/tables/tables.py
+++ b/netbox/netbox/tables/tables.py
@@ -116,16 +116,23 @@ class BaseTable(tables.Table):
             self.sequence.remove('actions')
             self.sequence.append('actions')
 
-    def _apply_prefetching(self):
+    def _apply_prefetching(self, columns=None):
         """
-        Dynamically update the table's QuerySet to ensure related fields are pre-fetched
+        Dynamically update the table's QuerySet to ensure related fields are pre-fetched.
+
+        Args:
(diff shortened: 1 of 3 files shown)


jeremystretch:
@claude /review Confirm that this fixes the reported behavior and check for any unintended consequences.

arthanson (approved):

PR #21981, opened 2026-04-21
Add support for DynamicModelMultipleChoiceField in custom fields filterset forms.

miaow2:
### Closes: #21854

In the method `_get_form_field` of the class `NetBoxModelFilterSetForm`, the attribute `for_filterset_form` is [passed](https://github.com/miaow2/netbox/blob/main/netbox/netbox/forms/filtersets.py#L48) for rendering custom fields filter forms. 
But the condition for `for_filterset_form` is [missing](https://github.com/miaow2/netbox/blob/main/netbox/extras/models/customfields.py#L585) for object custom fields, and `DynamicModelChoiceField` is rendered, and the user can't choose several objects for filtering.
I have added this condition.
Also added condition `not for_filterset_form` to disable the selector button, because it does not work in filter forms.

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -593,7 +593,7 @@ class CustomField(CloningMixin, ExportTemplatesMixin, OwnerMixin, ChangeLoggedMo
                 'required': required,
                 'initial': initial,
             }
-            if not for_csv_import and not for_filterset_form:
+            if not for_csv_import:
                 kwargs['query_params'] = self.related_object_filter
                 kwargs['selector'] = True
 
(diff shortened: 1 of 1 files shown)


pheus:

pheus (changes requested):
Thanks for working on this!

I agree with the main fix here: using `DynamicModelMultipleChoiceField` for `TYPE_OBJECT` in filter forms seems reasonable.

However, I’d like to ask that we remove the unrelated `selector` change from this PR. As written, the new `not for_filterset_form` guard does more than just hide the selector: it also stops passing `related_object_filter` as `query_params`, and it changes the existing `TYPE_MULTIOBJECT` path at the same time.

To me, that makes the scope broader than the issue this PR is trying to solve. If the selector does not work correctly in filter forms, that sounds like a separate bug which should be reported and tracked independently. Hiding the selector could be a workaround, but I think we should aim to fix that behavior rather than fold it into this change.

So I’d prefer to keep this PR focused on the `TYPE_OBJECT` multi-select fix, preserve the current `query_params` behavior, and handle selector follow-up separately.

miaow2:
@pheus, thanks for your review. I have pushed the requested changes.

pheus (approved):
Thanks, @miaow2 !

PR #21984, opened 2026-04-22
Closes #19648: Add support for colored Custom Field Choice Set values

### LINKED ISSUE #19648: Add a to field type based on ColorChoices with associated widget to Custom Fields
### NetBox version

v4.3.1

### Feature type

### Proposed functionality

It's not currently possible to use colors effectively in Custom Fields. You can use create a select field and add a custom selection set, but it isn't currently possible to view those colors using the existing form widget.

I'm proposing to add an additional custom field type for color selection using the existing color choices making it possible to add a color selection to custom fields.

As an alternative approach, allow for overriding of which widget is used for a field and make the color choices available as a custom field base choice set (e.g. like IATA codes etc). This would provide greater flexibility and permit users to use custom color sets.

### Use case

I have a requirement to assign color classifications to VRFs and some other models to enable some reporting outside of netbox.

This is presently done using a custom field, and I've manually extracted all the colours used by NetBox into extra field set, however, there is no rendering of the colors in the UI.

### Database changes

None

### External dependencies

None

pheus:
### Fixes: #19648

This adds optional per-choice color bindings to `CustomFieldChoiceSet` without changing the existing `value[:label]` syntax for `extra_choices`.

When a color is defined for a choice value, selection and multiple-selection custom fields render that value as a colored badge on object detail pages. The `CustomFieldChoiceSet` detail view also displays the configured color for each choice. Dropdown rendering is unchanged.

This implementation includes validation, form/import support, REST and GraphQL support, documentation updates, and tests.

### DIFF
--- a/docs/models/extras/customfield.md
+++ b/docs/models/extras/customfield.md
@@ -103,6 +103,8 @@ The default value to populate for the custom field when creating new objects (op
 
 For selection and multi-select custom fields only, this is the [set of choices](./customfieldchoiceset.md) which are valid for the field.
 
+Choice sets may optionally define colors for individual values. Colored choices are rendered as badges on object detail pages for selection and multiple-selection custom fields.
+
 ### Cloneable
 
 If enabled, values from this field will be automatically pre-populated when cloning existing objects.
--- a/docs/models/extras/customfieldchoiceset.md
+++ b/docs/models/extras/customfieldchoiceset.md
@@ -22,6 +22,28 @@ The set of pre-defined choices to include. Available sets are listed below. This
 
 A set of custom choices that will be appended to the base choice set (if any).
 
+### Choice Colors
+
+Optional color bindings for individual choice values. Each color is bound to a choice by its value rather than its label.
+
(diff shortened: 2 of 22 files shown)


pheus:

pheus:
I looked at these again, and I don't think any follow-up changes are needed here.

1. **Filter performance**
   Fair point, but this is an intentional ORM-only tradeoff. I don't want to introduce raw SQL for this, and Django doesn't really offer a cleaner exact-value JSON lookup for unknown keys.

2. **Form-level color validation**
   The model already validates allowed colors and invalid choice references, and that runs through the `ModelForm` as well. Adding the same check again in the form would mostly be duplicate validation.

3. **`_colors` cache**
   True in theory, but this follows the same caching pattern already used elsewhere on the model. I don't think it justifies special handling here.

4. **`removed_choices` logic**
   This change is intentional. A value that still exists through `base_choices` has not really been removed from the effective choice set, even if it is no longer present in `extra_choices`.

So overall, I think these are good observations, but not things I’d change in this PR.

jnovinger (approved):

PR #21987, opened 2026-04-22
Closes #21924: Add sticky bulk actions to account templates (follow-up)

### LINKED ISSUE #21924: Improve styling and consistency of floating bulk actions

pheus:
### Fixes: #21924

Follow-up to #21924 to apply the new sticky-actions pattern to account list templates that were missed in the original refactor. This updates bookmarks, notifications, and subscriptions so their bulk delete controls use the same sticky behavior and styling as other list views.

### DIFF
--- a/netbox/templates/account/bookmarks.html
+++ b/netbox/templates/account/bookmarks.html
@@ -23,8 +23,10 @@
     </div>
 
     {# Form buttons #}
-    <div class="btn-list d-print-none mt-2">
-      {% bulk_delete_button model query_params=request.GET %}
+    <div class="card btn-list sticky-actions d-print-none" data-sticky-position="right" data-sticky-when="selection">
+      <div class="btn-list bulk-action-buttons">
+        {% bulk_delete_button model query_params=request.GET %}
+      </div>
     </div>
   </form>
 {% endblock %}
--- a/netbox/templates/account/notifications.html
+++ b/netbox/templates/account/notifications.html
@@ -23,8 +23,10 @@
     </div>
 
     {# Form buttons #}
-    <div class="btn-list d-print-none mt-2">
-      {% bulk_delete_button model query_params=request.GET %}
+    <div class="card btn-list sticky-actions d-print-none" data-sticky-position="right" data-sticky-when="selection">
+      <div class="btn-list bulk-action-buttons">
+        {% bulk_delete_button model query_params=request.GET %}
+      </div>
     </div>
   </form>
 {% endblock %}
--- a/netbox/templates/account/subscriptions.html
+++ b/netbox/templates/account/subscriptions.html
@@ -23,8 +23,10 @@
     </div>
 
     {# Form buttons #}
-    <div class="btn-list d-print-none mt-2">
-      {% bulk_delete_button model query_params=request.GET %}
+    <div class="card btn-list sticky-actions d-print-none" data-sticky-position="right" data-sticky-when="selection">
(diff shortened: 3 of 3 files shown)


netbox/templates/account/bookmarks.html
```diff
@@ -23,8 +23,9 @@
     </div>
 
     {# Form buttons #}
-    <div class="btn-list d-print-none mt-2">
-      {% bulk_delete_button model query_params=request.GET %}
-    </div>
+    <div class="sticky-actions card btn-list d-print-none" data-sticky-position="right" data-sticky-when="selection">
+      <div class="bulk-action-buttons">
+        {% bulk_delete_button model query_params=request.GET %}
+      </div>
```
jnovinger, line 29: I think we're missing a closing `</div>` here.

netbox/templates/account/notifications.html
```diff
@@ -23,8 +23,9 @@
     </div>
 
     {# Form buttons #}
-    <div class="btn-list d-print-none mt-2">
-      {% bulk_delete_button model query_params=request.GET %}
-    </div>
+    <div class="sticky-actions card btn-list d-print-none" data-sticky-position="right" data-sticky-when="selection">
+      <div class="bulk-action-buttons">
+        {% bulk_delete_button model query_params=request.GET %}
+      </div>
```
jnovinger, line 29: Same, missing closing `</div>`.

jnovinger:

jnovinger (changes requested):

jnovinger:
> **2. CSS class order inconsistency (minor)**
> 
> The new account templates use `"sticky-actions card btn-list d-print-none"`, but the canonical pattern in `generic/object_list.html` and `generic/object_children.html` is `"card btn-list sticky-actions d-print-none"`. Functionally equivalent, but inconsistent with the established ordering.

I don't think this is worth pursuing.

jnovinger:
Whoops, hit the close button on accident.

pheus:

pheus:
Thanks for catching these!

jnovinger (approved):

PR #21994, opened 2026-04-24
Fixes #21989: Validate EventRule action data to prevent event processing failures

### LINKED ISSUE #21989: EventRule.action_data accepts non-object JSON, causing 500 on every object write
### NetBox Edition

NetBox Community

### NetBox Version

4.5.7

### Python Version

3.12

### Steps to Reproduce

1. Operations → Event Rules → Add.
2. Set an **Object type** (e.g. `dcim.device`), an **Event type** (e.g. `object_updated`), and an **Action** (e.g. a webhook).
3. In the **Data** field, enter a bare JSON value that is valid JSON but is not an object. For example:
   - `"test"`
   - `42`
4. Leave the rule **enabled** and save.
5. Perform a matching write against a subscribed object

### Expected Behavior

Either:
- `action_data` is rejected at save time (must be a JSON object or null), or
- The event dispatcher tolerates any JSON value at runtime.

Writes to unrelated objects should never fail because of a malformed rule.

### Observed Behavior

The form accepts the bare scalar; the rule saves without complaint. On the next matching write, `flush_events` raises and the request returns HTTP 500.

pheus:
### Fixes: #21989

This PR adds validation to `EventRule.clean()` so `action_data` must be a JSON object or `null`.

It also hardens event processing for legacy rows that may already contain invalid `action_data`: instead of raising an error during event handling, NetBox now logs a warning and falls back to an empty dictionary.

Together, this prevents malformed `action_data` from breaking unrelated object writes while ensuring new data is validated correctly.

### DIFF
--- a/netbox/extras/events.py
+++ b/netbox/extras/events.py
@@ -181,9 +181,26 @@ def process_event_rules(event_rules, object_type, event):
         if not event_rule.eval_conditions(event['data']):
             continue
 
+        # Guard against action_data that is valid JSON but not a dict
+        # (e.g. a bare string or number). Existing rows with bad data are
+        # tolerated at runtime; validation on EventRule.clean() prevents
+        # new ones.
+        if event_rule.action_data is None:
+            action_data = {}
+        elif isinstance(event_rule.action_data, dict):
+            action_data = event_rule.action_data
+        else:
+            logger.warning(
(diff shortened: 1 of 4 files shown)


pheus:

jeremystretch (approved):

PR #22006, opened 2026-04-27
Fixes #21917: Fix link peer resolution for cable profiles

### LINKED ISSUE #21917: Rear port has multiple link peers when connected to an interface using trunk cable profile
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.8

### Python Version

3.12

### Steps to Reproduce

1. Create device `switch1` and make sure it has four interfaces
2. Create device `patch panel 1` and make sure it has four rear ports, each with one position
3. Create a new cable with the following spec:
    - Status: Connected
    - Profile: 4C1P trunk
    - Terminations:
        - **A Side**
           Type: Interface
           Device: switch 1
           Interface: 1, 2, 3 and 4
        - **B side**
           Type: Rear Port
           Device: patch panel 1
           Rear Port: 1, 2, 3, 4
4. Go to the Rear Ports tab of the device `patch panel 1`

### Expected Behavior

- `Rear Port 1` should have `switch1 Interface 1` as it's link peer
- `Rear Port 2` should have `switch1 Interface 2` as it's link peer
- `Rear Port 3` should have `switch1 Interface 3` as it's link peer
- `Rear Port 4` should have `switch1 Interface 4` as it's link peer

### Observed Behavior

All four Rear Ports have all four interfaces as their link peers.

Image

Tracing works as expected:

Image

pheus:
### Fixes: #21917

Updates link peer resolution for profiled cables so peers are matched by their cable profile connector/position mapping instead of returning all terminations on the opposite cable end.

This adds a profile-aware `_get_profile_link_peers()` helper while preserving the existing behavior for non-profile cables. It also includes a regression test for the `TRUNK_4C1P` profile to verify that interfaces and rear ports resolve only their correct mapped peers.

### DIFF
--- a/netbox/dcim/models/device_components.py
+++ b/netbox/dcim/models/device_components.py
@@ -255,13 +255,37 @@ class CabledObjectModel(models.Model):
 
     @cached_property
     def link_peers(self):
-        if self.cable:
-            return [
-                peer.termination
-                for peer in self.cable.terminations.all()
-                if peer.cable_end != self.cable_end
-            ]
-        return []
+        if not self.cable:
+            return []
+
+        if self.cable.profile:
+            return self._get_profile_link_peers()
+
+        return [peer.termination for peer in self.cable.terminations.all() if peer.cable_end != self.cable_end]
+
+    def _get_profile_link_peers(self):
+        if self.cable_end is None or self.cable_connector is None or not self.cable_positions:
+            return []
+
+        profile = self.cable.profile_class()
+        peer_terminations = {
+            (peer.connector, position): peer.termination
+            for peer in self.cable.terminations.all()
+            if peer.cable_end == self.opposite_cable_end and peer.connector is not None
+            for position in peer.positions or []
+        }
+        link_peers = []
+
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
@claude /review
Thanks, both points make sense.

For the first one, I updated the logic so profiled cables always use the profile-aware path. If a profiled cable is missing connector/position metadata, `link_peers` now returns an empty list instead of falling back to all opposite-side terminations. I went with returning `[]` rather than raising because this accessor is used for display/API output, and returning all peers would be more misleading than returning none in an invalid/stale state.

For the test coverage, I kept the `TRUNK_4C1P` regression because it matches #21917 directly, but added an asymmetric breakout/shuffle case as well. I used `BREAKOUT_2C4P_8C1P_SHUFFLE` since it verifies that connector remapping works when each A-side termination maps to only a subset of B-side terminations, and vice versa. That gives better coverage of the `peer.connector == peer_connector` matching logic than a purely symmetric trunk case.

jeremystretch (approved):

PR #22009, opened 2026-04-27
Fixes #22002: Enable horizontal scrolling for Context Table Panels

### LINKED ISSUE #22002: UI: Horizontal scroll missing for Parent Prefixes table on IP Address view
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.8

### Python Version

3.12

### Steps to Reproduce

1. Configure DEFAULT_USER_PREFERENCES in configuration.py to include multiple columns (including custom fields) for the PrefixTable.
2. Navigate to an IP Address view that has associated Parent Prefixes.
3. Observe the Parent Prefixes panel.

The issue is easily reproducible when a large number of columns are enabled by default for users via configuration.py:
```
DEFAULT_USER_PREFERENCES = {
    'tables': {
        'PrefixTable': {
            'columns': [
                'prefix',
                'vlan',
                'status',
                'tags',
                ... ,
                'last_updated'
            ],
            'ordering': [
                '-last_updated'
            ]
        }
    }
}
```

### Expected Behavior

The table should be contained within the parent card, providing a horizontal scrollbar to access all columns without breaking the layout.
The  **table** element within the Parent Prefixes card should be wrapped in a div with the Bootstrap class table-responsive. 

```
<div class="table-responsive">
    <table class="table table-hover object-list">
        </table>
</div>
```
Image

### Observed Behavior

The table overflows the parent container and the screen boundaries. No horizontal scrollbar is provided, rendering rightmost columns and action buttons inaccessible.
Image

pheus:
### Fixes: #22002

This adds a `table-responsive` wrapper around tables rendered by the `ContextTablePanel` template.

This ensures wide context tables can scroll horizontally instead of overflowing their panel, especially on smaller viewports or when many table columns are enabled. The change is scoped to the context table panel template so generic table rendering remains unchanged.

### DIFF
--- a/netbox/templates/ui/panels/context_table.html
+++ b/netbox/templates/ui/panels/context_table.html
@@ -2,5 +2,7 @@
 {% load render_table from django_tables2 %}
 
 {% block panel_content %}
-  {% render_table table 'inc/table.html' %}
+  <div class="table-responsive">
+    {% render_table table 'inc/table.html' %}
+  </div>
 {% endblock panel_content %}


jnovinger:

jnovinger (approved):

PR #22013, opened 2026-04-27
Fixes #21988: Enforce object permissions for nested related objects in the REST API

### LINKED ISSUE #21988: Authorization bypass in nested object resolution via unrestricted nested lookup filters
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.8

### Python Version

3.12

### Steps to Reproduce

I previously reported this vulnerability through the private security contact security@netboxlabs.com but have not yet received a response, so I am opening this issue as a minimal public tracking stub.

1.  Deploy a current NetBox instance using the official `netbox-docker` release image.
2. Create:
   - one administrator account
   - one low-privileged user with permission to add/view `dcim.device` only
   - at least one `Site`
   - at least one `Manufacturer`
   - at least one `DeviceType` object that the low-privileged user is not permitted to view directly

3. Confirm the permission boundary:
   - an administrator can access `GET /api/dcim/device-types/`
   - the low-privileged user receives `403` for `GET /api/dcim/device-types/`
   - the low-privileged user can still submit `POST /api/dcim/devices/`

4. As the low-privileged user, submit a device creation request using a nested `device_type` object in the request body.

5. Observe that nested related-object resolution is performed during request validation, and that the behavior differs depending on whether the supplied nested attributes match restricted related objects.

6. Compare this behavior with the direct read endpoint, where the same user is correctly denied access to those related objects.

### Expected Behavior

Nested object resolution during API write operations should enforce the same object-level permission restrictions as the corresponding read endpoints.

A user who cannot view a related object directly should not be able to distinguish or resolve that object indirectly through nested serializer input.

### Observed Behavior

During API write operations, nested related-object resolution does not appear to apply the same permission restrictions as the direct API read path.

As a result, a low-privileged authenticated user can interact with nested related-object references in a way that reveals whether restricted related objects match the supplied attributes, even though the user cannot access those objects directly through their normal API endpoints.

pheus:
### Fixes: #21988

This updates nested serializer handling to respect object-level view permissions when resolving related objects from REST API write input.

Nested create and update operations now resolve related objects from a permission-restricted queryset, so objects the user cannot view fail validation the same way as nonexistent objects. This prevents users from resolving or assigning hidden related objects indirectly through nested API input.

Normal `GET` response rendering is unchanged by this PR.

Tests have been added to cover both nested serializer input paths, including visible and hidden related objects by ID and attributes.

### DIFF
--- a/netbox/core/tests/test_changelog.py
+++ b/netbox/core/tests/test_changelog.py
@@ -149,7 +149,14 @@ class ChangeLogViewTestCase(ModelViewTestCase):
             'path': self._get_url('delete', instance=site),
             'data': post_data({'confirm': True}),
         }
-        self.add_permissions('dcim.delete_site')
+        self.add_permissions(
+            'dcim.add_module',
+            'dcim.delete_site',
+            'dcim.view_device',
+            'dcim.view_modulebay',
+            'dcim.view_moduletype',
+            'extras.view_tag',
+        )
         response = self.client.post(**request)
         self.assertHttpStatus(response, 302)
 
@@ -449,7 +456,7 @@ class ChangeLogAPITestCase(APITestCase):
         }
         self.assertEqual(ObjectChange.objects.count(), 0)
         url = reverse('dcim-api:site-list')
-        self.add_permissions('dcim.add_site')
+        self.add_permissions('dcim.add_site', 'extras.view_tag')
 
         response = self.client.post(url, data, format='json', **self.header)
         self.assertHttpStatus(response, status.HTTP_201_CREATED)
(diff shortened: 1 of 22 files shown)


pheus:

pheus:

pheus:

pheus:
@claude /review

I added an end-to-end API test for the constrained related-object case: a user with view access limited to Tenant 1 can create a Prefix referencing Tenant 1, but receives a validation error when referencing Tenant 2. That should cover the gap left by the shared CRUD helpers.

I also checked the query-count change. The `12 → 13` increase is isolated to the user list baseline and comes from one fixed permission-resolution query for the `permissions` field. It is not N+1 behavior and does not affect write operations.

The `GenericObjectSerializer` behavior change is intentional: an unresolved PK now returns `400` instead of leaking an uncaught `ObjectDoesNotExist` as a `500`, which is consistent with other invalid related-object references.

For `JournalEntrySerializer.created_by`, I left the field as-is. `User` does support `restrict()`, but this field intentionally remains a plain `PrimaryKeyRelatedField`; the normal path uses `CurrentUserDefault`, so only explicitly supplied `created_by` values go through the field queryset.

Finally, the non-`RestrictedQuerySet` case is intentional. Models without `RestrictedQuerySet` do not participate in NetBox object-level permission scoping, so there is no restriction to apply there. I added a docstring note rather than a runtime warning, since warning on every such resolution would be noisy.

bctiemann (approved):

PR #22015, opened 2026-04-27
#21990 fix deletion of device from Virtual Machines

### LINKED ISSUE #21990: Device is being deleted from virtual machines
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.6

### Python Version

3.12

### Steps to Reproduce

Devices -> Device -> Virtual Mashines -> select virtual mashines -> Edit Selected -> any change -> Apply

### Expected Behavior

After saving the changes, the device is saved for the virtual machine.

### Observed Behavior

After saving the changes, the device is deleted for the VM.

arthanson:
### Fixes: #21990

When navigating to a bulk edit via a URL containing ?device=<id> (e.g. Device → Virtual Machines tab → Edit Selected), the view passes the device ID as initial data on the form. Django's changed_data then sees the device field as modified (from the initial device ID to the empty submitted value) causing it to be nullified on save.

Fixed by popping device from self.initial in VirtualMachineBulkEditForm and VirtualDeviceContextBulkEditForm. The ?device GET param is navigation/filter context, not an intent to edit the field.

Regression tests added for both models.

### DIFF
--- a/netbox/dcim/forms/bulk_edit.py
+++ b/netbox/dcim/forms/bulk_edit.py
@@ -1754,6 +1754,13 @@ class VirtualDeviceContextBulkEditForm(PrimaryModelBulkEditForm):
     )
     nullable_fields = ('device', 'tenant', )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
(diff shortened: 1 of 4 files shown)


arthanson:
@claude /review

bctiemann (approved):

PR #22017, opened 2026-04-27
#21995 Don't copy unique fields when adding another Contact

### LINKED ISSUE #21995: Contacts - Create & Add Another does not clear inputs
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.8

### Python Version

3.12

### Steps to Reproduce

1. Create a new contact with the web form
2. Click **Create & Add Another**

### Expected Behavior

Only certain fields should be preserved, like Groups. Fields like Name and Email should not, as that is generally unique per contact.

### Observed Behavior

All fields are preserved

arthanson:
### Fixes: #21995

removes name, title, phone, email, address, and link from Contact.clone_fields, keeping only groups. Now when a user clicks "Create & Add Another", only the contact group selection carries over

### DIFF
--- a/netbox/tenancy/models/contacts.py
+++ b/netbox/tenancy/models/contacts.py
@@ -111,7 +111,7 @@ class Contact(PrimaryModel):
     )
 
     clone_fields = (
-        'groups', 'name', 'title', 'phone', 'email', 'address', 'link',
+        'groups',
     )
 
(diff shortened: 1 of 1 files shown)


arthanson:
@claude /review

pheus (approved):

PR #22018, opened 2026-04-27
#21934 allow highlight override table stripping

### LINKED ISSUE #21934: Striped table rows override interface-type row highlighting on Device Interfaces (Virtual/LAG) – inconsistent styling
### NetBox Edition

NetBox Community

### NetBox Version

v.4.5.5

### Python Version

3.12

### Steps to Reproduce

1. Navigate to Devices → (select a device) → Interfaces (e.g. /dcim/devices/<id>/interfaces/).
2. Ensure the device has Virtual interfaces and/or LAG interfaces in the list.
3. Go to User Preferences and enable “Striped table rows”.
4. Return to the device’s Interfaces table.

### Expected Behavior

Interface-type highlighting for Virtual interfaces and LAGs should remain consistent and clearly visible regardless of whether “Striped table rows” is enabled.
Ideally, striping should either:

- be applied in a way that preserves the type highlight (e.g. subtle overlay/tint), or
- not override explicit row highlight styling for these interface types.

### Observed Behavior

- Zebra striping (alternating row backgrounds) overrides or visually dominates the row background styling intended to highlight Virtual interfaces and LAGs.
- Virtual/LAG rows no longer stand out consistently; depending on odd/even row position, the stripe background changes the appearance.

Striping disabled (Virtual rows have uniform highlight): Image

Striping enabled (Virtual rows become zebra-striped and lose consistent highlight):
Image

arthanson:
### Fixes: #21934 

The old code set background-color on the tr, but cell backgrounds cover row backgrounds in CSS - so the stripe's box-shadow always won.

Fix: Replace background-color on the rows with setting `--tblr-table-bg-state` (using SCSS interpolation #{} to force expression evaluation).

Monosnap dmi01-akron-sw01 | NetBox 2026-04-27 13-10-18


jeremystretch (changes requested):
We seem to have lost the row borders, at least for virtual interfaces. This makes it difficult to scan tables horizontally.

Image

arthanson:
> We seem to have lost the row borders, at least for virtual interfaces. This makes it difficult to scan tables horizontally.

Looks like it was chrome specific, safari was fine. I think it was showing just too light to see, I darkened it a little bit and looks good now.

jeremystretch (approved):
Nice!

PR #22022, opened 2026-04-28
Fixes #22005: Fix connected interface filter for incomplete cable paths

### LINKED ISSUE #22005: Query for connected interfaces shown cabled interfaces
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.5

### Python Version

3.12

### Steps to Reproduce

1. Connect a device interface to an front port of a patchpanel where the last end is not connected to an interface
2. Query https://netbox/api/dcim/interfaces/?device_id=deviceid&cabled=true or https://netbox/api/dcim/interfaces/?device_id=deviceid&connected=true
Expected Behavior

Observed Behavior

### Expected Behavior

https://netbox/api/dcim/interfaces/?device_id=deviceid&cabled=true should show one interface more than https://netbox/api/dcim/interfaces/?device_id=deviceid&connected=true. The last one between the taken interface from steps to reproduce and the patchpanel should be not shown as it is not connected:
````json
{
            "id": 274,
            "url": "https://netbox/api/dcim/interfaces/274/",
            "display_url": "https://netbox/dcim/interfaces/274/",
            "display": "GigabitEthernet1/0/3",
            ...
            "link_peers": [
                {
                    "id": 721,
                    "url": "https://netbox/api/dcim/front-ports/721/",
                    "display": "Port 1",
              ...
            "connected_endpoints": null,
            "connected_endpoints_type": null,
            "connected_endpoints_reachable": false,
              ...
}
````

### Observed Behavior

There is no difference between https://netbox/api/dcim/interfaces/?device_id=deviceid&cabled=true or https://netbox/api/dcim/interfaces/?device_id=deviceid&connected=true even if the third port that don't have a connection (no endpoint) is made.

pheus:
### Fixes: #22005

This updates the `connected` filter used by interfaces to require a cable path to be both active and complete.

Previously, an interface cabled to a patch-panel front port could be returned by `connected=true` even when the mapped rear port had no onward connection. In that state, the path is active but incomplete, so the interface should be considered cabled but not connected.

The change keeps `cabled=true` behavior unchanged while ensuring `connected=true` only returns interfaces with a complete reachable path.

Regression coverage has been added to the existing interface filterset tests for an active but incomplete pass-through path.

### DIFF
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -1874,8 +1874,8 @@ class PathEndpointFilterSet(django_filters.FilterSet):
 
     def filter_connected(self, queryset, name, value):
         if value:
-            return queryset.filter(_path__is_active=True)
-        return queryset.filter(Q(_path__isnull=True) | Q(_path__is_active=False))
+            return queryset.filter(_path__is_active=True, _path__is_complete=True)
+        return queryset.filter(Q(_path__isnull=True) | Q(_path__is_active=False) | Q(_path__is_complete=False))
 
 
 @register_filterset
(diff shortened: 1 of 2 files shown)


pheus:

arthanson (approved):

PR #22026, opened 2026-04-28
Permit Claude triage workflow for users without write permission

jeremystretch:


pheus (changes requested):
Thanks!

_One small concern_: since `allowed_non_write_users: "*"` means this can be triggered by arbitrary public issue content, I think we should update the pinned `anthropics/claude-code-action` SHA before enabling it.

Newer versions include additional hardening for this mode, including subprocess environment scrubbing for workflows using `allowed_non_write_users`. **Could we pin to v1.0.77 or newer here?**

_Optional thought_: since this will now run for all newly opened issues, we might want to add a small `--max-turns` limit to keep the workflow predictable and avoid unexpectedly long runs.

Not blocking from my side, but it could be a nice safety guard.

pheus (approved):
Thanks!

PR #22035, opened 2026-04-29
Fixes #22029: Recast empty string values on unique nullable fields as null

### LINKED ISSUE #22029: Setting an `asset_tag` to an empty string on more than one device raises an IntegrityError
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.9

### Python Version

3.12

### Steps to Reproduce

1. Create two devices
2. Make a REST API request to set the `asset_tag` of the first device to an empty string:

```
curl -X PATCH \
http://netbox:8000/api/dcim/devices/1/ \
--data '{"asset_tag": ""}'
```

3. Make a REST API request to set the `asset_tag` of the second device to an empty string:

```
curl -X PATCH \
http://netbox:8000/api/dcim/devices/2/ \
--data '{"asset_tag": ""}'
```

I've not tested this yet, but it likely applies to the `asset_tag` fields on other models, and may apply to nullable CharFields generally throughout the application.

### Expected Behavior

One of two things should probably happen:

1. A ValidationError should be raised when attempting to set the asset tag of the second device, caught, and returned as an error. (This happens as expected when attempting to set the same non-empty string value on two devices.)
2. The initial request should fail validation because `asset_tag` enforces uniqueness: Empty strings should be prohibited and only `null` permitted to represent an empty value.

The second option seems like it would be the most robust solution.

### Observed Behavior

An IntegrityError exception is raised.

Note this occurs only with duplicate empty string values: Attempting to set a duplicate non-empty string returns a well-formed error as expected.

jeremystretch:
### Fixes: #22029

Extends BaseModel to automatically recast empty string values on nullable CharFields as null to avoid triggering a uniqueness violation when attempting to set an "empty" value.

### DIFF
--- a/netbox/dcim/tests/test_models.py
+++ b/netbox/dcim/tests/test_models.py
@@ -638,11 +638,26 @@ class DeviceTestCase(TestCase):
         device2.full_clean()
         device2.save()
 
-    def test_empty_asset_tag_coerced_to_null(self):
+    def test_empty_asset_tag_coerced_to_null_on_clean(self):
         """
         An empty string assigned to a unique nullable CharField (e.g. asset_tag) must be coerced
         to None on save so that multiple objects can be saved without violating the unique
-        constraint.
+        constraint. Test that this is done on clean().
+        """
+        common_kwargs = {
+            'site': Site.objects.first(),
+            'device_type': DeviceType.objects.first(),
+            'role': DeviceRole.objects.first(),
+        }
+        device1 = Device(name='Device 1', asset_tag='', **common_kwargs)
+        device1.clean()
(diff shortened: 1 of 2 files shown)


jeremystretch:
@claude /review Consider whether this approach is safe given its wide breadth in touching practically every model. We're aiming for a solution that doesn't require e.g. plugin authors to apply a fix themselves, but also want to avoid introducing any unintended behavior.

jeremystretch:
- Consolidated the coercion logic under a single private helper method
- Added a test to check that the coercion is applied during `clean()`

jnovinger (approved):

PR #22037, opened 2026-04-29
Fixes #22031: Fix error when adding a Prefix from a VLAN with no Tenant or Site

### LINKED ISSUE #22031: Add prefix to VLAN "Field 'id' expected a number but got 'None'."
### NetBox Edition

NetBox Community

### NetBox Version

4.5.8-Docker-4.0.2

### Python Version

3.12

### Steps to Reproduce

1. Click on 'IPAM' -> 'VLANs' -> choose a vlan -> click on '+ Add a Prefix'

Using [Netbox-sync](https://github.com/bb-ricardo/netbox-sync) to sync vlans, vm's etc from vsphere/vmware.
Have tried Netbox 4.6.0-beta1-Docker-4.0.2. Same error.

2. Workaround.
Prefixes -> Add -> Add prefix and set "VLAN Assignment"

### Expected Behavior

Should be able to add prefix.

### Observed Behavior

Server Error

<class 'ValueError'>

Field 'id' expected a number but got 'None'.

Python version: 3.12.3
NetBox version: 4.5.8-Docker-4.0.2
Plugins: None installed
The request which yielded the above error is shown below:

GET https://[REDACTED URL]/ipam/prefixes/add/?tenant=None&site=None&vlan=105&return_url=%2Fipam%2Fvlans%2F105%2F

pheus:
### Fixes: #22031

This PR fixes the "Add a Prefix" action on VLAN detail pages when optional values such as tenant or site are not set.

Previously, URL parameters that resolved to `None` were serialized into the query string as the literal value `"None"`, for example `tenant=None`. This could cause the Prefix add form to fail while trying to interpret `"None"` as an object ID.

The change updates `LinkAction.get_url()` to skip parameters that resolve to `None`, so optional parameters are simply omitted from the generated URL.

This PR also updates the VLAN-to-Prefix action to pass the Prefix scope using `scope_type` and `scope` instead of the old `site` parameter, while still preselecting the VLAN.

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -1636,8 +1636,11 @@ class VLANView(generic.ObjectView):
                     actions.AddObject(
                         'ipam.prefix',
                         url_params={
-                            'tenant': lambda ctx: ctx['object'].tenant.pk if ctx['object'].tenant else None,
-                            'site': lambda ctx: ctx['object'].site.pk if ctx['object'].site else None,
+                            'tenant': lambda ctx: ctx['object'].tenant_id,
+                            'scope_type': lambda ctx: (
+                                ContentType.objects.get_for_model(Site).pk if ctx['object'].site_id else None
+                            ),
+                            'scope': lambda ctx: ctx['object'].site_id,
                             'vlan': lambda ctx: ctx['object'].pk,
                         },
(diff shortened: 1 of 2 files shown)


pheus:

bctiemann (approved):

PR #22064, opened 2026-04-30
Closes #22062: Display API token ID & plaintext one time immediately upon creation

### LINKED ISSUE #22062: One-time display of the v2 API token ID & plaintext immediately upon creation
### NetBox version

v4.5.9

### Feature type

### Proposed functionality

Currently, the process for creating a v2 API token via the web UI is as follows.

1. Navigate to the token creation form. A random plaintext value is pre-populated in the form.
2. Copy the plaintext and save it locally.
3. Submit the form.

Upon creation, the token detail view now shows the token ID, and the plaintext is no longer known to the server.

This FR proposes two closely related changes to improve the user's workflow:

1. Defer random plaintext generation until the form has been submitted.
2. Present the ID and plaintext one time for convenient reference by the user. After refreshing or navigating away, the plaintext permanently disappears (it is never stored on the server), but the ID will persist as it does currently.

### Use case

This affords the user an opportunity to copy the entire authorization string, rather than the two-part process as implemented today. This saves time and avoids confusion.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #22062

- When creating an API token via the web UI, temporarily attach its plaintext to the current session for display on the token detail view.
- Remove the `token` field from the form
- Add tests

### DIFF
--- a/netbox/templates/users/inc/new_token_banner.html
+++ b/netbox/templates/users/inc/new_token_banner.html
@@ -1,0 +1,24 @@
+{% load i18n %}
+<div class="alert alert-success bg-success-subtle mb-3" role="alert">
+  <div class="d-flex">
+    <div>
+      <i class="mdi mdi-check-bold pe-2"></i>
+    </div>
+    <div class="flex-grow-1">
+      <h4 class="alert-title">{% trans "Save this token now &mdash; it cannot be retrieved later." %}</h4>
+      <div class="text-secondary mb-2">
+        {% blocktrans %}
+          The full HTTP Authorization header value is shown below. NetBox does not store the token plaintext,
+          so this is your only chance to copy it. It will be lost when navigating away from this page.
+        {% endblocktrans %}
+      </div>
+      <div class="d-flex align-items-center gap-2">
+        <code id="new-token-auth-string" class="flex-grow-1 p-2 bg-body border rounded font-monospace text-break">{{ token_auth_string }}</code>
+        <a class="btn btn-primary copy-content" data-clipboard-target="#new-token-auth-string" title="{% trans "Copy to clipboard" %}">
+          <i class="mdi mdi-content-copy"></i>
+          {% trans "Copy" %}
+        </a>
+      </div>
+    </div>
+  </div>
+</div>
--- a/netbox/templates/users/inc/new_token_banner.html
+++ b/netbox/templates/users/inc/new_token_banner.html
@@ -5,7 +5,7 @@
       <i class="mdi mdi-check-bold pe-2"></i>
     </div>
     <div class="flex-grow-1">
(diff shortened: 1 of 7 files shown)


jeremystretch:
@claude /review

jeremystretch:
@claude /review I've addressed the previous concerns:

1. Fixed
2. Fixed
3. Removed the —
4. Lack of test coverage specifically for account views is consistent
5. Out of scope for the PR

netbox/users/views.py
```diff
@@ -55,6 +64,11 @@ class TokenEditView(generic.ObjectEditView):
     form = forms.TokenForm
     template_name = 'users/token_edit.html'
 
+    def alter_object(self, obj, request, url_args, url_kwargs):
+        # Attach the request so that UserTokenForm.save() can stash the newly-generated plaintext on the session.
+        obj._request = request
+        return obj
+
```
pheus, line 71: Could we handle the `Create & Add Another` workflow?

This view still uses the generic object edit flow, so token creation can be submitted with `_addanother`. In that case, the user is redirected back to the add form instead of the newly-created token’s detail view, which means the one-time plaintext banner is never shown immediately after creation.

netbox/templates/users/inc/new_token_banner.html
```diff
@@ -0,0 +1,24 @@
...
+  <div class="d-flex">
+    <div>
+      <i class="mdi mdi-check-bold pe-2"></i>
+    </div>
+    <div class="flex-grow-1">
+      <h4 class="alert-title">{% trans "Save this token now. It cannot be retrieved later." %}</h4>
+      <div class="text-secondary mb-2">
+        {% blocktrans %}
+          The full HTTP Authorization header value is shown below. NetBox does not store the token plaintext,
+          so this is your only chance to copy it. It will be lost when navigating away from this page.
```
pheus, line 12: _Minor wording suggestion_: since this flow temporarily stores the plaintext in the session until it is displayed, I’d avoid the absolute “does not store” wording here.

Maybe:

```suggestion
          The full HTTP Authorization header value is shown below. This is the only time this token value will be displayed.
          Copy it now and store it securely. It will be lost when navigating away from this page.
```

That keeps the important warning without making a stronger storage claim than we need.

pheus (changes requested):
Thanks! This is a nice improvement overall.

I’m requesting changes for one workflow issue: `Create & Add Another` can bypass the token detail view, so the generated plaintext may not be shown immediately after creation.

netbox/templates/users/inc/new_token_banner.html
jeremystretch: I prefer to keep the wording as-is. I believe it's accurate.

netbox/users/views.py
jeremystretch: Good catch!

pheus (approved):
Looks good to me! Thanks!

PR #22083, opened 2026-05-03
Fixes #21960: Allow cable trace to continue through multi-position FrontPorts without position context

### LINKED ISSUE #21960: Cable trace fails to resolve connected_endpoints when a front port with multiple positions maps to multiple rear ports that converge on the same remote endpoint
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.8

### Python Version

3.12

### Steps to Reproduce

When a front port has 2 positions, each mapped to a separate rear port, and both rear ports are cabled through to the same remote interface, the cable trace fails to resolve `connected_endpoints` from the multi-position side. The trace works correctly in the reverse direction.

This is a problem when modelling duplex optical connections — for example, a QSFP28 transceiver module where a single edge connector (front port) has TX and RX paths (two rear ports) that both ultimately connect to the same remote interface.

Create two devices with pluggable optic modules connected back-to-back. The devices are identically connected at the fibre level — the only difference is how the QSFP28 edge connector is modelled (one front port with 2 positions vs two front ports with 1 position each).

### Device A (test-router-a)

Module has **two separate front ports**, each mapping to one rear port:

| Component | Name | Details |
|---|---|---|
| Interface | ce0 | 100GBASE-LR4 |
| Front port | ce0-tx | 1 position, maps to rear port ce0-TX |
| Front port | ce0-rx | 1 position, maps to rear port ce0-RX |
| Rear port | ce0-TX | LC/UPC, 1 position |
| Rear port | ce0-RX | LC/UPC, 1 position |

### Device B (test-router-b)

Module has **one front port with 2 positions**, mapping to the same two rear ports:

| Component | Name | Details |
|---|---|---|
| Interface | ce0 | 100GBASE-LR4 |
| Front port | ce0 | 2 positions: pos 1 → ce0-TX, pos 2 → ce0-RX |
| Rear port | ce0-TX | LC/UPC, 1 position |
| Rear port | ce0-RX | LC/UPC, 1 position |

### Cabling

1. **Device A internal:** Interface ce0 → multi-termination cable → front ports ce0-tx + ce0-rx
2. **Device B internal:** Interface ce0 → cable → front port ce0
3. **Fibre between devices:** Device A ce0-TX (rear) ↔ Device B ce0-RX (rear)
4. **Fibre between devices:** Device A ce0-RX (rear) ↔ Device B ce0-TX (rear)

### Use case

This models a duplex optical connection — a QSFP28 transceiver module where a single edge connector (front port) has TX and RX fibre paths (two rear ports) that both ultimately connect to the same remote interface. Device A uses a workaround (two front ports), Device B uses the natural model (one duplex front port with 2 positions).

### Expected Behavior

Tracing from either device's interface should resolve `connected_endpoints` to the other device's interface, since both fibre paths converge on the same remote endpoint.

Both positions on Device B's front port lead through separate cables to separate rear ports on Device A, but both paths converge on the **same** interface (test-router-a:ce0). Since the endpoint is unambiguous — both branches of the split arrive at the same destination — `connected_endpoints` should resolve to that endpoint.

A path split is only genuinely ambiguous when the branches lead to *different* endpoints.

### Observed Behavior

**Trace from Device A (two front ports) → Device B: WORKS**
- 3 hops, `connected_endpoints: [test-router-b:ce0]`

**Trace from Device B (single front port, 2 positions) → Device A: FAILS**
- 1 hop only (interface → front port), `connected_endpoints: null`
- The trace stops at the multi-position front port — it encounters a path split and does not follow either path

The UI shows "No connected interface" on Device B's interface. When manually tracing in the UI, it asks to select a cable (the path split), and then correctly shows the full trace to Device A's interface. So the trace engine *can* follow the path — it just doesn't resolve `connected_endpoints` automatically.

Image
Image
Image
Image

Note that interface test-router-a:ce0, shows a connection to test-router-b:ce0, but test-router-b:ce0 fails to show the connection in the other direction (instead reporting non-reachable).

attoba:
### Fixes: #21960

When tracing encounters a multi-position FrontPort with an empty position_stack, instead of aborting unconditionally, perform a one-hop lookahead: inspect what each mapped RearPort is cabled to on the far end and check whether all far-end terminations share the same parent device/module.

- If they share a parent, the split is convergent (all strands lead to the same endpoint) and traversal continues correctly.
- If they diverge to different parents, the path is genuinely ambiguous and is flagged as is_split=True.

The existing abort behavior for multi-position RearPorts is preserved unchanged.

Tests added:
- test_225: convergent duplex QSFP model resolves connected_endpoints in both directions.
- test_226: divergent multi-position FrontPort still flags is_split=True.

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -936,6 +936,7 @@ class CablePath(models.Model):
                 break
 
             if isinstance(remote_terminations[0], FrontPort):
+
                 # Follow FrontPorts to their corresponding RearPorts
                 if remote_terminations[0].positions > 1 and position_stack:
                     positions = position_stack.pop()
@@ -943,15 +944,44 @@ class CablePath(models.Model):
                     for rt in remote_terminations:
                         q_filter |= Q(front_port=rt, front_port_position__in=positions)
                     port_mappings = PortMapping.objects.filter(q_filter)
-                elif remote_terminations[0].positions > 1:
-                    is_split = True
-                    logger.debug(
-                        'Encountered front port mapped to multiple rear ports but position stack is empty; aborting '
-                        'trace.'
-                    )
-                    break
                 else:
                     port_mappings = PortMapping.objects.filter(front_port__in=remote_terminations)
+
+                    # For multi-position FrontPorts with no positional context, check whether
+                    # all mapped RearPorts converge on the same remote parent (device/module).
+                    # If they diverge to different parents the split is genuinely ambiguous;
+                    # if they share a parent the split is convergent and safe to follow.
+                    if any(rt.positions > 1 for rt in remote_terminations):
+                        if not port_mappings:
+                            break
+                        rear_ports = [m.rear_port for m in port_mappings]
+                        termination_type = ObjectType.objects.get_for_model(rear_ports[0])
+                        local_cts = CableTermination.objects.filter(
+                            termination_type=termination_type,
+                            termination_id__in=[rp.pk for rp in rear_ports]
+                        )
+
+                        q_filter = Q()
+                        for lct in local_cts:
+                            far_end = 'A' if lct.cable_end == 'B' else 'B'
+                            q_filter |= Q(cable=lct.cable, cable_end=far_end)
+
+                        if q_filter:
+                            far_end_cts = CableTermination.objects.filter(q_filter).prefetch_related('termination')
+                            far_end_parents = set()
+                            for ct in far_end_cts:
+                                t = ct.termination
+                                far_end_parents.add(
+                                    ('module', t.module_id) if getattr(t, 'module_id', None)
+                                    else ('device', getattr(t, 'device_id', None))
+                                )
+                            if len(far_end_parents) > 1:
+                                is_split = True
+                                logger.debug(
+                                    'Multi-position FrontPort maps to RearPorts connected to different '
+                                    'remote parents; divergent path, aborting trace.'
+                                )
+                                break
                 if not port_mappings:
                     break
 
--- a/netbox/dcim/tests/test_cablepaths.py
+++ b/netbox/dcim/tests/test_cablepaths.py
@@ -2608,6 +2608,7 @@ class LegacyCablePathTests(CablePathTestCase):
         self.assertPathIsSet(interface2, path2)
 
     def test_224_single_path_via_multiple_pass_throughs_with_breakouts(self):
+
         """
         [IF1] --C1-- [FP1] [RP1] --C2-- [IF3]
         [IF2]        [FP2] [RP2]        [IF4]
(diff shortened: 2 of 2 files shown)


jeremystretch:

jeremystretch (changes requested):
@attoba could you please include tests to ensure that this change doesn't introduce a breaking change to existing behavior?

jeremystretch:
@attoba are you still interested in working on this?

netbox/dcim/models/cables.py
```diff
@@ -937,53 +864,32 @@ def from_origin(cls, terminations):
                 # Profile-based tracing
                 if links[0].profile:
                     cable_profile = links[0].profile_class()
-                    if position_stack:
-                        positions = position_stack.pop()
-                    else:
-                        # When the position stack is empty (e.g. the trace reached this
-                        # profiled cable after crossing single-position pass-through ports
-                        # which don't push onto the stack), derive positions from each
-                        # termination's own cable_positions — which were set by this
```
jeremystretch, line 947: Removing this branch fails to address the scenario where two front ports map to _different_ remote endpoints:

> With the abort gone and the rear ports being single-position, nothing pushes onto `position_stack` (line 1056), so the split-detection guard at line 887 (`position_stack and …`) is skipped on the next hop. If the two positions of the front port lead to *different* remote endpoints, the trace will now complete with two `connected_endpoints` and `is_split = False`, rather than being flagged as a split. The issue text explicitly frames the desired behavior as "a path split is only genuinely ambiguous when the branches lead to different endpoints" — but this PR doesn't actually distinguish convergent from divergent; it just stops aborting.

Rather than removing it, consider checking that the mapped ports all share the same parent.

jeremystretch (changes requested):

jeremystretch:

jeremystretch (changes requested):
@attoba thanks for following up on this. Please see if you can address the concerns raised above.

jeremystretch:

jeremystretch (changes requested):
Thanks for keeping at this @attoba. Unfortunately this doesn't sufficiently validate that the diverged paths rejoin. I've included some analysis from Claude that breaks down the specific concerns.

**1. `netbox/dcim/models/cables.py` ~1043-1058 — parent-equality is necessary but not sufficient; strands landing on different endpoints of the *same* device are merged into one wrong path.** (CONFIRMED)
The check compares only the immediate far-end termination's parent, never that the strands reconverge on one endpoint. Topology: duplex `fp_b` (positions=2) → `rp_b_1`,`rp_b_2`, cabled to `rp_a_1`,`rp_a_2` on the same device A, which fan out to two *different* interfaces `if_a_1`/`if_a_2`. `far_end_parents == {('device', A)}` (len 1) → not split → trace continues and terminates at `[if_a_1, if_a_2]`, all same type → `is_complete=True, is_split=False`. Result: `if_b.connected_endpoints == [if_a_1, if_a_2]` (two endpoints presented as a complete link), while the reverse trace from either `if_a_*` returns a single peer `if_b` — an asymmetric, physically-invalid merge that the removed `break` prevented. This propagates to `connected_endpoints_reachable`, `destinations`, tables, panels, and GraphQL, none of which consult `is_split`.

**2. `netbox/dcim/models/cables.py` ~1046-1051 — far-end terminations with no `device_id`/`module_id` all collapse to `('device', None)`, so divergent circuits/provider-networks are treated as convergent.** (CONFIRMED)
`CircuitTermination` (and site/provider-network terminations) have neither `device_id` nor `module_id`, so `('module', t.module_id) if getattr(t,'module_id',None) else ('device', getattr(t,'device_id',None))` yields `('device', None)` for every one. Topology: duplex `fp_b` → `rp_b_1`,`rp_b_2`, cabled to CircuitTerminations on two *different* circuits. `far_end_parents == {('device', None)}` (len 1) → not split → the trace merges two unrelated circuits into one path instead of flagging `is_split=True`.

**3. `netbox/dcim/models/cables.py` ~1057-1058 — the convergent continuation is now reachable with an empty stack and pushes a flat `rear_port_position` list mixing positions from distinct rear ports.** (PLAUSIBLE)
When the converged rear ports themselves have `positions > 1`, `position_stack.append([mapping.rear_port_position for mapping in port_mappings])` pushes one flat list gathered across multiple different rear ports. The next hop applies it uniformly via `rear_port_position__in=...` to every rear port, so a rear port that legitimately uses only positions 1–2 also matches phantom positions 3–4 from a sibling rear port → extra front ports traced. Under the old code this path aborted as split, so the diff newly exposes this. (Not exercised by the added tests, whose rear ports are all `positions=1`.)

**4. `netbox/dcim/models/cables.py` ~1046-1051 — the same parent-key expression produces a *false* split for two components on one device when one is module-mounted and the other is not.** (CONFIRMED, low severity)
A module-installed component keys as `('module', module_id)` while a directly-mounted one on the same device keys as `('device', device_id)`, so a genuinely convergent path (both far ends on device X) is flagged `is_split=True`. This is the conservative direction (matches pre-PR behavior, no new wrong output), but it shows the parent key should resolve to the owning *device* uniformly rather than device-or-module.

**5. `netbox/dcim/models/cables.py` ~1043 — N+1 query: `[m.rear_port for m in port_mappings]` loads each RearPort with a separate query.** (CONFIRMED)
`PortMapping.objects.filter(front_port__in=...)` has no `select_related('rear_port')`, so the new full iteration issues one query per mapping on a hot path run for every FrontPort hop. Add `.select_related('rear_port')` (also benefits the existing `mapping.rear_port` use at line 1056).

**6. `netbox/dcim/models/cables.py` ~965-975 — the far-end lookup duplicates the legacy positionless branch at lines 992-1010.** (cleanup)
The `get_for_model` → `CableTermination.filter(termination_type, termination_id__in)` → cable-end flip → `filter(q_filter).prefetch_related('termination')` sequence is a near-verbatim copy of lines 992-1010 (the `far_end = 'A' if lct.cable_end == 'B' else 'B'` block copies 998-1001). Extract a shared "far-end terminations for a set of local terminations" helper and reuse it in both places.

Unfortunately I don't have sufficient time to invest in this right now. You're welcome to continue working at it, but I'm not comfortable altering the tracing logic unless we can verify with sufficient confidence that it won't introduce unintended behavior.

jeremystretch:
@attoba I'm closing this PR and unassigning you from the associated issue as there's been no response to my review above. If you'd like to revisit this work, please let us know before submitting a new PR.

PR #22086, opened 2026-05-04
Fixes #22084: Fix OpenAPI schema for nullable `cable_end` values on cabled objects

### LINKED ISSUE #22084: OpenAPI `Interface.cable_end` is required and non-nullable but returns `null` for unconnected interfaces
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.9

### Python Version

3.12

### Steps to Reproduce

1. Start a fresh NetBox v4.5.9 instance (the official `netboxcommunity/netbox:v4.5.9` Docker image suffices).

2. Provision an API token. Then create the prerequisites for an interface (site, manufacturer, device-type, device-role, device) via these REST API calls:

   curl -fsS -H "Authorization: Token $TOKEN" -H 'Content-Type: application/json' \
     -X POST http://netbox/api/dcim/sites/ \
     -d '{"name":"DC1","slug":"dc1"}'

   curl -fsS -H "Authorization: Token $TOKEN" -H 'Content-Type: application/json' \
     -X POST http://netbox/api/dcim/manufacturers/ \
     -d '{"name":"Acme","slug":"acme"}'

   curl -fsS -H "Authorization: Token $TOKEN" -H 'Content-Type: application/json' \
     -X POST http://netbox/api/dcim/device-types/ \
     -d '{"manufacturer":{"slug":"acme"},"model":"Server","slug":"server"}'

   curl -fsS -H "Authorization: Token $TOKEN" -H 'Content-Type: application/json' \
     -X POST http://netbox/api/dcim/device-roles/ \
     -d '{"name":"Server","slug":"server","color":"3366ff"}'

   curl -fsS -H "Authorization: Token $TOKEN" -H 'Content-Type: application/json' \
     -X POST http://netbox/api/dcim/devices/ \
     -d '{"name":"srv01","site":1,"device_type":1,"role":1}'

3. Create an interface with no cable attached:

   curl -fsS -H "Authorization: Token $TOKEN" -H 'Content-Type: application/json' \
     -X POST http://netbox/api/dcim/interfaces/ \
     -d '{"device":1,"name":"Ethernet1","type":"1000base-t"}'

4. Read the interface back:

   curl -fsS -H "Authorization: Token $TOKEN" http://netbox/api/dcim/interfaces/1/

5. Compare against the schema declaration for `cable_end`:

   curl -fsS http://netbox/api/schema/ | jq '.components.schemas.Interface.properties.cable_end, (.components.schemas.Interface.required | map(select(. == "cable_end")))'

### Expected Behavior

The schema declares `cable_end` on `Interface` as a required, non-nullable string:

```json
"cable_end": {
  "type": "string",
  "readOnly": true
}
```

with cable_end listed in Interface.required. By the OpenAPI spec, the field should therefore appear in every Interface response as a string.

The behavior I expect is one of:

(a) The schema is corrected to mark cable_end nullable: true (matching the actual database model, where CabledObjectModel.cable_end is CharField(blank=True, null=True) in netbox/dcim/models/device_components.py), so strict clients accept the null. This mirrors the fix already applied to the neighboring fields connected_endpoints_type and link_peers_type in PR #16627.

(b) Or the API always serializes cable_end to a non-null string (e.g., empty string) when no cable is attached.

Either resolution makes the schema match the runtime response.

### Observed Behavior

The runtime response from step 4 contains "cable_end": null:

{
  "id": 1,
  "name": "Ethernet1",
  "device": { ... },
  "cable": null,
  "cable_end": null,
  "_occupied": false,
  ...
}

This contradicts the schema's type: string declaration on a required field. Strict typed clients generated from the schema (e.g., via openapi-generator, typify, or any OpenAPI-strict consumer) reject this response with an error along the lines of invalid type: null, expected a string for the cable_end field. Every Interface response with no cable attached fails validation.

This is the same class of bug as #16624 (parent_device), fixed by #16627. PR #16627 specifically added allow_null=True to:

- connected_endpoints_type
- link_peers_type
- parent_device (NestedDeviceSerializer)

The author of #16627 wrote: "Went through all uses of SerializerMethodField and added allow_null=True if the method could return None." cable_end was missed because it's a model CharField rather than a SerializerMethodField, so the same sweep didn't catch it.

The fix likely lives in whichever serializer represents CabledObjectModel fields — adding allow_null=True on the cable_end field declaration, or extra_kwargs = {"cable_end": {"allow_null": True}} on the relevant serializer Meta.

This is closely related to the broader #15588 ("Nullable fields not nullable in the OpenAPI spec").

pheus:
### Fixes: #22084

This updates `CabledObjectSerializer` so `cable_end` is correctly represented in the generated API schema.

Uncabled objects can already return `"cable_end": null`, but the schema currently describes the field as a non-null string. This change marks the field as nullable and uses DRF’s `ChoiceField` so the schema also reflects the allowed values while preserving the existing raw `"A"` / `"B"` API output.

### DIFF
--- a/netbox/dcim/api/serializers_/cables.py
+++ b/netbox/dcim/api/serializers_/cables.py
@@ -95,7 +95,8 @@ class CablePathSerializer(serializers.ModelSerializer):
 
 class CabledObjectSerializer(serializers.ModelSerializer):
     cable = CableSerializer(nested=True, read_only=True, allow_null=True)
-    cable_end = serializers.CharField(read_only=True)
+    # Use DRF's ChoiceField; NetBox's ChoiceField would return a value/label object.
+    cable_end = serializers.ChoiceField(choices=CableEndChoices, read_only=True, allow_null=True)
     link_peers_type = serializers.SerializerMethodField(read_only=True, allow_null=True)
     link_peers = serializers.SerializerMethodField(read_only=True)
     _occupied = serializers.SerializerMethodField(read_only=True)


pheus:

pheus:

jeremystretch (approved):

PR #22094, opened 2026-05-04
Closes #21938: Display ancestors for hierarchical Generic Foreign Key targets

### LINKED ISSUE #21938: Adding context to Circuit Terminations
### NetBox version

v4.5.8

### Feature type

Data model extension

### Proposed functionality

Currently, the interface for Circuit Terminations only displays the immediate termination point, which obscures the broader context that hierarchies provide. 

This proposal seeks to improve geographic visibility for circuits terminated at Sites, Locations, or Regions within the UI. 

### Use case

The most effective way is to explain this through a "Rapid Troubleshooting" scenario where it is important the information is available. If an engineer needed to know the direction to be travelling in, they shouldn't have to 'click to find out.' 

Without the full lineage, a circuit assigned to 'Location X' is not useful information. 

Image

We should update the UI to utilize the `get_ancestors` method for displaying nested parent hierarchies of the same type (region > region > termination_type or location > location > termination_type)

If possible, we should explore beyond the same-type recursion and provide the full lineage (region > ... > site > location > ... > termination_type) without needing to do additional lookups.

Image

By surfacing the path, our engineers will be able to have full geographical context and be on their way. 

### Database changes

None

### External dependencies

None

pheus:
### Fixes: #21938

This adds optional `nested` and `max_depth` support to `GenericForeignKeyAttr`, allowing tree-like GFK targets to render their ancestors as breadcrumbs when they expose `get_ancestors(include_self=True)`.

The behavior is opt-in, so non-hierarchical GFK targets continue to render as before. This is applied to circuit termination points, including the template-backed circuit termination panel, as well as relevant scoped fields in IPAM and wireless to provide clearer context for hierarchical objects.

### DIFF
--- a/netbox/circuits/ui/panels.py
+++ b/netbox/circuits/ui/panels.py
@@ -12,16 +12,43 @@ class CircuitCircuitTerminationPanel(panels.ObjectPanel):
 
     template_name = 'circuits/panels/circuit_circuit_termination.html'
     title = _('Termination')
+    termination_ancestor_max_depth = 3
 
     def __init__(self, side, accessor=None, **kwargs):
         super().__init__(accessor=accessor, **kwargs)
         self.side = side
 
+    def _get_termination_nodes(self, termination):
+        """
+        Return the termination target's ancestors, including itself, when the
+        target is tree-like.
+
+        Non-tree GFK targets return None, so the template preserves the current
+        single-object rendering.
+        """
+        target = getattr(termination, 'termination', None)
+        if target is None:
+            return None
+
+        get_ancestors = getattr(target, 'get_ancestors', None)
+        if not callable(get_ancestors):
+            return None
+
+        nodes = list(get_ancestors(include_self=True))
+
+        if self.termination_ancestor_max_depth is not None:
+            nodes = nodes[-self.termination_ancestor_max_depth:]
+
+        return nodes
+
     def get_context(self, context):
+        termination = resolve_attr_path(context, f'{self.accessor}.termination_{self.side.lower()}')
+
         return {
             **super().get_context(context),
             'side': self.side,
-            'termination': resolve_attr_path(context, f'{self.accessor}.termination_{self.side.lower()}'),
+            'termination': termination,
+            'termination_nodes': self._get_termination_nodes(termination),
         }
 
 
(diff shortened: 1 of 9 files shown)


pheus:

pheus:
@claude /review 

I've addressed point 1 and 3.

I’m going to keep the small duplicated helper in `CircuitCircuitTerminationPanel` for now. That panel is still template-backed during the declarative UI migration, so keeping the panel-specific glue local feels clearer for this PR.

netbox/netbox/ui/attrs.py
```diff
@@ -412,19 +412,48 @@ class GenericForeignKeyAttr(ObjectAttribute):
...
+             `nested` is enabled. Ignored otherwise.
     """
     template_name = 'ui/attrs/generic_object.html'
 
-    def __init__(self, *args, linkify=None, **kwargs):
+    def __init__(self, *args, linkify=None, nested=False, max_depth=None, **kwargs):
         super().__init__(*args, **kwargs)
         self.linkify = linkify
+        self.nested = nested
+        self.max_depth = max_depth
```
jeremystretch, line 429: Might be prudent to add some lightweight validation, checking that `max_depth` (if specified) is a positive integer to avoid a potentially confusing exception when an invalid value is passed.

netbox/circuits/ui/panels.py
```diff
@@ -58,7 +85,9 @@ class CircuitTerminationPanel(panels.ObjectAttributesPanel):
     title = _('Circuit Termination')
     circuit = attrs.RelatedObjectAttr('circuit', linkify=True)
     provider = attrs.RelatedObjectAttr('circuit.provider', linkify=True)
-    termination = attrs.GenericForeignKeyAttr('termination', linkify=True, label=_('Termination point'))
+    termination = attrs.GenericForeignKeyAttr(
+        'termination', linkify=True, max_depth=3, nested=True, label=_('Termination point')
```
jeremystretch: Super minor nit, but could we switch the order of the kwargs to match the signature? Feels more natural IMO.
```suggestion
        'termination', linkify=True, nested=True, max_depth=3, label=_('Termination point')
```

jeremystretch (changes requested):

netbox/circuits/ui/panels.py
pheus: Thanks, I’ve applied the kwarg-order change.

netbox/netbox/ui/attrs.py
pheus: I’d prefer not to add `max_depth` validation in this PR. This option mirrors `NestedObjectAttr.max_depth`, which does not currently validate the value, and it is declarative UI configuration rather than user-provided input. If an invalid value is introduced in code, I think it is fine for that to fail during development/tests.

jeremystretch (commented):
Thanks!

jeremystretch (approved):

PR #22100, opened 2026-05-04
Fixes #20776: Add changelog message to bulk rename process

### LINKED ISSUE #20776: Add changelog message to bulk rename process
### NetBox version

v4.4.5

### Feature type

### Proposed functionality

The changelog message box should be added to the object bulk rename process.

### Use case

Seems to be one area that allows changes to be made without provision for a changelog message.

### Database changes

_No response_

### External dependencies

_No response_

jnovinger:
### Fixes: #20776

Adds changelog message support to the bulk rename view, bringing it in line with bulk edit, bulk delete, and bulk import.

#### Changes

- `BulkRenameForm` now inherits from `ChangelogMessageMixin`, adding an optional changelog message field
- `BulkRenameView` sets `obj._changelog_message` before save (both MPTT and non-MPTT paths)
- Conditional field removal in `BulkRenameView.__init__` for models that don't support change logging
- New test `test_bulk_rename_objects_with_changelog_message` in the standard view test mixin

No template changes were needed: `render_form.html` already handles `meta_fields` rendering.

#### Form relocation

`BulkRenameForm` was moved from `utilities/forms/forms.py` to `netbox/forms/bulk_rename.py`. This was necessary because `utilities/forms/forms.py` loads early in Django's boot chain (via `extras/models` → `utilities/fields` → `utilities/forms`). Adding `ChangelogMessageMixin` as a parent there creates a circular import since the mixin's module imports `extras.models`, which hasn't finished loading at that point. The new location in `netbox/forms/` loads after the app registry is populated, so the import works cleanly.

### DIFF
--- a/netbox/netbox/forms/__init__.py
+++ b/netbox/netbox/forms/__init__.py
@@ -1,5 +1,6 @@
 from .bulk_edit import *
 from .bulk_import import *
+from .bulk_rename import *
 from .filtersets import *
 from .model_forms import *
 from .search import *
--- a/netbox/netbox/forms/bulk_rename.py
+++ b/netbox/netbox/forms/bulk_rename.py
@@ -1,0 +1,40 @@
+import re
+
+from django import forms
+from django.utils.translation import gettext as _
+
+from .mixins import ChangelogMessageMixin
+
+__all__ = (
+    'BulkRenameForm',
+)
+
+
+class BulkRenameForm(ChangelogMessageMixin, forms.Form):
+    """
+    An extendable form to be used for renaming objects in bulk.
+    """
+    find = forms.CharField(
+        strip=False
+    )
+    replace = forms.CharField(
+        strip=False,
+        required=False
+    )
+    use_regex = forms.BooleanField(
+        required=False,
+        initial=True,
+        label=_('Use regular expressions')
+    )
+
+    def clean(self):
+        super().clean()
+
+        # Validate regular expression in "find" field
+        if self.cleaned_data['use_regex']:
+            try:
+                re.compile(self.cleaned_data['find'])
+            except re.error:
+                raise forms.ValidationError({
+                    'find': "Invalid regular expression"
+                })
--- a/netbox/netbox/forms/bulk_rename.py
+++ b/netbox/netbox/forms/bulk_rename.py
@@ -1,40 +1,14 @@
-import re
-
-from django import forms
-from django.utils.translation import gettext as _
+from utilities.forms import BulkRenameForm
 
 from .mixins import ChangelogMessageMixin
 
 __all__ = (
-    'BulkRenameForm',
+    'NetBoxModelBulkRenameForm',
 )
 
 
-class BulkRenameForm(ChangelogMessageMixin, forms.Form):
+class NetBoxModelBulkRenameForm(ChangelogMessageMixin, BulkRenameForm):
     """
-    An extendable form to be used for renaming objects in bulk.
+    Extends BulkRenameForm with a changelog message field for NetBox models that support change logging.
     """
-    find = forms.CharField(
-        strip=False
-    )
-    replace = forms.CharField(
-        strip=False,
-        required=False
-    )
-    use_regex = forms.BooleanField(
-        required=False,
-        initial=True,
-        label=_('Use regular expressions')
-    )
-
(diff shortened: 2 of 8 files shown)


jnovinger:

jnovinger:

> #### 1. `BulkRenameView` ignores the `form` class attribute — `VMInterfaceBulkRenameForm` / `VirtualDiskBulkRenameForm` are dead code

Out of scope, will file a follow up housekeeping issue for this.

> #### 2. `base_fields.pop()` mutates the class-level dict
Added a comment indicating why this is safe.

> #### 3. New test does not skip for models without change logging support

Added.

pheus (commented):
Thanks for explaining the circular import! That helped a lot while reviewing.

That said, I don’t think we should move `BulkRenameForm` out of `utilities.forms` in a patch release. That changes an existing import path that plugins/custom code may already rely on, e.g. `from utilities.forms import BulkRenameForm`.

It also changes the current separation between generic utility forms in `utilities` and NetBox model-specific forms in `netbox`. I’m not opposed to moving or reorganizing these forms, but I think that should wait for a release where breaking import-path changes are allowed.

The cleanest patch-release-safe alternative I can think of is to keep `utilities.forms.BulkRenameForm` as the generic form and add a NetBox-specific wrapper, similar to the existing bulk edit pattern:

```python
class NetBoxModelBulkRenameForm(ChangelogMessageMixin, BulkRenameForm):
    pass
```

I’d also suggest making `BulkRenameView.form` an explicit extension point, while still injecting the `pk` field dynamically from the view. The `pk` field depends on the scoped queryset, so I think it should remain view-owned rather than becoming part of the generic form.

For example:

```python
class BulkRenameView(GetReturnURLMixin, BaseMultiObjectView):
    """
    An extendable view for renaming objects in bulk.

    Attributes:
        field_name: The name of the object attribute for which the value is being updated (defaults to "name")
        form: Optional base form class used for the rename operation
    """
    field_name = 'name'
    template_name = 'generic/bulk_rename.html'
    filterset = None
    form = None

        base_form = self.form
        if base_form is None:
            base_form = (
                NetBoxModelBulkRenameForm
                if issubclass(self.queryset.model, ChangeLoggingMixin)
                else BulkRenameForm
            )

        class _Form(base_form):
            pk = ModelMultipleChoiceField(
                queryset=self.queryset,
                widget=MultipleHiddenInput()
            )

        self.form = _Form
```

This keeps `utilities.forms.BulkRenameForm` generic and backwards-compatible, avoids importing `ChangelogMessageMixin` into `utilities.forms`, and removes the need to mutate `base_fields` to strip `changelog_message` for models that do not support change logging.

One thing to watch: `VMInterfaceBulkRenameView` and `VirtualDiskBulkRenameView` already set `form`, but `BulkRenameView` currently overwrites it, so those form classes have effectively been unused. If we make `form` supported now, those two should either be updated to inherit from the new NetBox-specific bulk rename form, or removed/cleaned up separately if they are not needed.

jnovinger:
> Thanks for explaining the circular import! That helped a lot while reviewing.
> 
> ...

Thanks for the thorough review, @pheus. The feedback about preserving the `utilities.forms` import path is well taken.

I've pushed two commits:

**Commit 1** (`92181faa4`) addresses the core feedback:

- Restored `BulkRenameForm` to `utilities/forms/forms.py`, preserving the existing import path
- Created `NetBoxModelBulkRenameForm(ChangelogMessageMixin, BulkRenameForm)` in `netbox/forms/bulk_rename.py`, following the established `NetBoxModelBulkEditForm` pattern
- `BulkRenameView` now conditionally selects the base form based on whether the model supports change logging, which eliminates the `base_fields.pop()` mutation entirely

I intentionally did not add the `form` extension point in this PR. The view still overwrites `self.form` unconditionally, keeping existing behavior unchanged for a patch release.

**Commit 2** (`c3374bc3b`) is an optional cleanup: removes `VMInterfaceBulkRenameForm` and `VirtualDiskBulkRenameForm`, which have been dead code since `BulkRenameView` was refactored to create its form dynamically in `__init__` (1dbae5b64). Happy to split this to a separate housekeeping issue if you'd prefer to keep it out of this PR.

pheus (approved):
Thanks! I really appreciate that!

PR #22105, opened 2026-05-04
return v2 token for API Token create

### LINKED ISSUE #22081: TokenViewSet.create() does not return v2 token plaintext
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.9

### Python Version

3.12

### Steps to Reproduce

Create a v2 API token for another user via the REST API:

```
POST /api/users/tokens/
Authorization: Bearer <admin-v2-token>

{"user": 2, "write_enabled": false}
```

### Expected Behavior

The HTTP response should contain an usable token. 

The `users.grant_token` permission and the documentation for "Creating Tokens for Other Users" imply that admins should be able to create tokens for other users via the REST API and receive a usable token. With v2 tokens, this is broken — the plaintext is lost in the re-serialization step.

This blocks use cases like:
- Automated token provisioning for service accounts
- Short-lived per-tenant scoped tokens created on demand

### Observed Behavior

The "token" is null and cannot be retrieved:
```json
{
  "id": 9,
  "version": 2,
  "key": "nk7cPTUrEr9w",
  "token": null,
  ...
}
```

The token field is always null, making it impossible to use the created token — the full v2 plaintext (nbt_<key>.<token>) is only available at creation time and cannot be recovered.

The cause is that `NetBoxModelViewSet.create()` in `netbox/api/viewsets/__init__.py` re-serializes from the database after `perform_create()`:

```python
    serializer = self.get_serializer(data=request.data)
    serializer.is_valid(raise_exception=True)

    # Re-fetch from DB — this loses _token
    qs = self.get_queryset().get(pk=serializer.instance.pk)
    serializer = self.get_serializer(qs, many=bulk_create)

    return Response(serializer.data, ...)
```

The `Token._token` attribute is only set in memory after `save()` and is never persisted (by design — it's hashed). Re-fetching from DB returns `_token = None`, which the `TokenSerializer.token` field reads via the `Token.token` property.

This is in contrast with TokenProvisionView. The `/api/users/tokens/provision/` endpoint (`TokenProvisionView`) works correctly because it returns `serializer.data` directly after `save()` without re-fetching:

```python
def post(self, request):
    serializer = serializers.TokenProvisionSerializer(data=request.data, context={'request': request})
    serializer.is_valid(raise_exception=True)
    return Response(serializer.data, status=HTTP_201_CREATED)  # preserves _token
```

[token_v2_fix.py](https://github.com/user-attachments/files/27321331/token_v2_fix.py)

arthanson:
### Fixes: #22081 

NetBoxModelViewSet.create() re-fetches the instance from the database after perform_create() to attach prefetched relations, which discards the in-memory _token plaintext - and v2 tokens store only an HMAC digest, so the plaintext can't be recovered.

Bit of a surgical fix to avoid having to change the base create with an override function. TokenViewSet.perform_create() stashes the in-memory plaintext(s) keyed by pk, and get_serializer() re-attaches them to the re-fetched instance(s) before the response is serialized.  This does instantiate a list of the data but tokens are going to be small amount created so should be fine.

Originally looked at overriding create and copying the code, but that is about 15 lines of duplicated code and if any changes are made in the base they won't be picked up in this one (like the recent etag change).


arthanson:
@claude /review

jeremystretch (commented):
I'm worried this approach might be too fragile. I've submitted an alternate approach in [PR #22160](https://github.com/netbox-community/netbox/pull/22160), which more closely mirrors the solution we use for conveying the plaintext in the UI. Please take a look and let me know what you think of that approach.

PR #22106, opened 2026-05-04
Fixes #22055: Report API exceptions to Monitoring Services

### LINKED ISSUE #22055: API exceptions swallowed by middleware — never reported to Sentry
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.9

### Python Version

3.12

### Steps to Reproduce

1. Setup Sentry
2. Cause and API request exception (easiest is to modify the API to force an exception)

### Expected Behavior

API Error is reported to Sentry

### Observed Behavior

There is no error reported to Sentry.

`CoreMiddleware.process_exception()` in `netbox/netbox/middleware.py:105-106` intercepts all API request exceptions and returns a JSON 500 response via `handle_rest_api_exception()` (netbox/utilities/error_handlers.py:48-59). By returning a response, Django considers the exception handled, so Sentry's Django integration never sees it.

UI exceptions (form POSTs, etc.) fall through the middleware and remain unhandled, so Sentry captures them normally.

**Result**: We are blind to all API-originated 500 errors in Sentry.

### Proposed Fix

Two options:

**Option A:** Capture before swallowing (minimal change)

```
def handle_rest_api_exception(request, *args, **kwargs):
    type_, error = sys.exc_info()[:2]
    # Report to error tracking before returning JSON response
    try:
        import sentry_sdk
        sentry_sdk.capture_exception()
    except ImportError:
        pass
    ...
```

Downside: Hard-couples to Sentry in OSS code. Though it's already an optional dependency used in settings.py.

**Option B:** Use DRF's exception handler instead (better)

Remove the middleware interception for API requests entirely. Instead, configure a custom DRF EXCEPTION_HANDLER that formats the JSON response. DRF exception handlers don't suppress the exception from Django's perspective — Sentry (and any other error tracking) still sees it.

This is the more correct approach architecturally, but is a larger change.

arthanson:


arthanson:
@claude /review

arthanson:
addressed issue 1 in claude review.

pheus (approved):
Thanks!

PR #22117, opened 2026-05-05
Closes #22093: Run coverage only on a single matrix entry

### LINKED ISSUE #22093: Enable coverage calculation for only a single test matrix entry
### Proposed Changes

Our CI workflow currently runs `coverage run --source="netbox/" netbox/manage.py test netbox/ --parallel` for each matrix entry (one per Python version). This is needed only for one runner, as the coverage report should be effectively identical for all Python versions.

### Justification

This will reduce the overall text run time for runners not charged with reporting coverage.

pheus:
### Fixes: #22093

Add explicit CI job names showing the Python and Node versions, with the coverage job clearly marked in the GitHub Actions UI.

Run coverage only for the designated coverage matrix entry to avoid redundant coverage collection and reporting across the full test matrix.

Also add the YAML document marker and clean up trailing whitespace.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,3 +1,4 @@
+---
 name: CI
 
 on:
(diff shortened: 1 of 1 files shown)


pheus:

pheus:
This PR keeps that behavior unchanged and only limits the existing coverage run to a single matrix entry, which is the scope of #22093. So while the coverage accuracy concern may be real, I don’t think this PR introduces the issue or should expand to solve it.

I’ll open a separate housekeeping issue to track improving coverage collection for parallel test runs.

jnovinger (approved):

PR #22119, opened 2026-05-05
Closes #22118: Collect coverage from parallel test workers

### LINKED ISSUE #22118: Improve coverage accuracy for parallel test runs
### Proposed Changes

Review and improve how CI collects coverage when running the Django test suite with `--parallel`.

The current coverage command runs the test suite through Django's parallel test runner. Since parallel test execution uses worker processes, coverage collected only from the parent process may not include all code exercised by the test workers.

A follow-up change could address this in one of two ways:

* run the coverage job without Django's `--parallel` option; or
* configure [coverage.py](<http://coverage.py>) to collect subprocess/multiprocessing coverage data and   add a `coverage combine` step before generating the report.

### Justification

This would make the CI coverage report more accurate and avoid giving an incomplete picture of test coverage.

This is separate from netbox-community/netbox#22093, which only reduces redundant coverage runs across the CI matrix. Once coverage is run on a single matrix entry, it would be good to also ensure that the remaining coverage job captures the full test execution as accurately as possible.

pheus:
### Fixes: #22118

Configure coverage.py to collect data from Django's parallel test workers.

The coverage job currently runs the Django test suite with `--parallel`, which uses worker processes. This change adds coverage.py multiprocessing configuration, moves the measured source/report settings into `pyproject.toml`, and combines the per-process coverage data before generating the report.

This keeps the faster parallel test run while making the remaining coverage job more accurate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,12 +122,12 @@ jobs:
 
     - name: Run tests with coverage
       if: ${{ matrix.coverage }}
-      run: >-
-        coverage run --source="netbox/"
-        netbox/manage.py test netbox/ --parallel
+      run: coverage run netbox/manage.py test netbox/ --parallel
+
+    - name: Combine coverage data
+      if: ${{ matrix.coverage }}
+      run: coverage combine
 
     - name: Show coverage report
       if: ${{ matrix.coverage }}
-      run: >-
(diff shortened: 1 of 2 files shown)


pheus:

jeremystretch (approved):
Nice work!

PR #22133, opened 2026-05-07
Remove prohibition on AI-generated PRs; add guidance to `AGENTS.md`

jeremystretch:
- Remove the blanket prohibition on AI-generated pull requests. (AI-generated issues are still prohibited.)
- Add guidance to `AGENTS.md` conveying our PR policy.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -273,6 +273,26 @@ GitHub Actions workflows in `.github/workflows/`:
 - Every PR must reference an approved GitHub issue.
 - PRs must include tests for new functionality.
 
+## PR Submission Requirements
+
+**Do not open a PR unless all the following conditions are met:**
+
+1. **Issue reference required** — The PR body must include a `Closes: #<number>` line identifying the associated GitHub issue. PRs without this line must not be submitted.
+2. **Issue must be open** — Before opening a PR, verify via `gh issue view <number>` that the referenced issue is currently open. Do not submit a PR against a closed issue.
+3. **Issue must be assigned to you** — Verify that the referenced issue is assigned to the submitting user. Do not open a PR for an issue that is unassigned or assigned to someone else.
+4. **No exceptions without maintainer status** — These three requirements are waived only for project maintainers (members of the `netboxlabs` GitHub organization). All other contributors must satisfy all three checks before a PR is opened.
+
+**Pre-submission checklist for AI agents:**
+
+```bash
+# Confirm the issue is open and assigned before opening a PR
+gh issue view <number> --json state,assignees
+```
+
+Reject the PR submission and report the problem if the issue is closed, unassigned, or assigned to a different user.
+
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -273,6 +273,26 @@ GitHub Actions workflows in `.github/workflows/`:
 - Every PR must reference an approved GitHub issue.
 - PRs must include tests for new functionality.
 
+## PR Submission Requirements
+
+**Do not open a PR unless all the following conditions are met:**
+
+1. **Issue reference required** — The PR body must include a `Closes: #<number>` line identifying the associated GitHub issue. PRs without this line must not be submitted.
```
pheus, line 280: The new text requires a PR body line of exactly `Closes: #<number>`, but the repository’s PR template currently starts with `Fixes: #1234`. Since both are valid GitHub closing keywords, could we avoid requiring one exact keyword here? Otherwise agents following `AGENTS.md` may treat the existing PR template as invalid.

Maybe something like:

```suggestion
1. **Issue reference required** — The PR body must include an issue-closing reference identifying the associated GitHub issue, following the repository PR template, e.g. `Fixes: #<number>`. PRs without such a reference must not be submitted.
```

AGENTS.md
```diff
@@ -273,6 +273,26 @@ GitHub Actions workflows in `.github/workflows/`:
...
+2. **Issue must be open** — Before opening a PR, verify via `gh issue view <number>` that the referenced issue is currently open. Do not submit a PR against a closed issue.
+3. **Issue must be assigned to you** — Verify that the referenced issue is assigned to the submitting user. Do not open a PR for an issue that is unassigned or assigned to someone else.
+4. **No exceptions without maintainer status** — These three requirements are waived only for project maintainers (members of the `netboxlabs` GitHub organization). All other contributors must satisfy all three checks before a PR is opened.
```
pheus, line 283: The new text equates “project maintainers” with members of the `netboxlabs` GitHub organization. That feels a bit too broad/narrow for an operational agent instruction: org membership is not necessarily the same as maintainer status for `netbox-community/netbox`, and community maintainers may not map cleanly to that org.

Could we phrase this in terms of NetBox project maintainer approval instead, and avoid asking agents to infer maintainer status from org membership?

```suggestion
4. **No exceptions without maintainer approval** — These three requirements are waived only when explicitly approved by a NetBox project maintainer. All other contributors must satisfy all three checks before a PR is opened.
```

pheus (changes requested):

AGENTS.md
jeremystretch: We use both "closes" and "fixes" (the later typically for bugs). Either works, an arguably we should just standardize on "closes." The important part is that the PR gets associated to the issue by GitHub.

AGENTS.md
jeremystretch: I think that's complicating it a bit too much. There are very few scenarios in which a non-maintainer is permitted to submit a PR without an assigned issue to begin with (typically only e.g. typo fixes in the documentation). And we want to discourage these drive-by PRs anyway.

pheus (approved):

PR #22145, opened 2026-05-07
Introduce additional Claude skills

jeremystretch:
This introduces Claude skills to help with:
- Removing a model
- Removing a field from a model
- Adding & removing configuration parameters


jeremystretch:

jeremystretch:
Addressed both concerns above

bctiemann (approved):

PR #22147, opened 2026-05-07
Closes #22146: Avoid renumbering MPTT trees when creating module bays

### LINKED ISSUE #22146: Avoid renumbering MPTT trees when creating module bays
### NetBox Version

v4.6.0

### Python Version

3.12

### Area(s) of Concern

- [x] User Interface
- [x] REST API
- [ ] GraphQL API
- [x] Python ORM
- [ ] Other

### Observations

Currently, when creating a new module bay on a device, a new tree ID is assigned based on the module bay's ordering, and all tree IDs "after" it are renumbered. This is necessary to ensure proper native ordering, but causes performance issues at scale.

### Proposed Changes

Drop `order_insertion_by` from ModuleBay.MPTTMeta and override the default queryset for ModuleBay to order by parent device & root module bay name:

```python
    def get_queryset(self, *args, **kwargs):
        root_name = self.model._objects_raw.filter(
            tree_id=models.OuterRef('tree_id'),
            level=0,
        ).values('name')[:1]
        return super().get_queryset(*args, **kwargs).annotate(
            _root_name=models.Subquery(
                root_name,
                output_field=models.CharField(db_collation='natural_sort'),
            )
        ).order_by('_root_name', 'lft')
```

This is moderately more expensive for reads but avoids the renumbering overhead entirely. This approach hasn't been fully vetted yet at scale but appears promising.

jeremystretch:
### Closes: #22146

- Introduce a custom manager for the ModuleBay model to override the default queryset and tweak ordering
- Set `tree_id` to `Max('tree_id') + 1` for new ModuleBay instances to avoid renumbering existing records
- Avoid renumbering tree roots when renaming a ModuleBay

This is very much a hack, but unfortunately appears to be our only option short of dropping alphanumeric ordering for module bays. This adaptation will be needed only until MPTT can be removed under #21418.

### DIFF
--- a/netbox/dcim/models/device_components.py
+++ b/netbox/dcim/models/device_components.py
@@ -1313,6 +1313,29 @@ class RearPort(ModularComponentModel, CabledObjectModel, TrackingModelMixin):
 # Bays
 #
 
+class ModuleBayManager(TreeManager):
+    """
+    Order ModuleBays by the natural-sort name of each tree's root, then by MPTT
+    left value within the tree. Combined with the root-insert bypass in
+    ModuleBay.save(), this lets us keep MPTTMeta.order_insertion_by for cheap
+    intra-tree sibling placement while skipping the global tree_id renumbering
+    it would otherwise perform on every root insert.
+    """
+    def get_queryset(self, *args, **kwargs):
+        # Use the raw manager to avoid recursing through this get_queryset() when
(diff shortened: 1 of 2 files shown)


jeremystretch:
@claude /review

bctiemann (approved):

PR #22151, opened 2026-05-08
Closes #22150: Run collectstatic before Python tests

### LINKED ISSUE #22150: Ensure backend test jobs have required SVG CSS assets
### Proposed Changes

Update the CI workflow so backend test jobs which exercise SVG rendering have access to the CSS assets required by those tests.

After the recent split between test and frontend-related jobs, the backend test job no longer appears to have the following files available under `STATIC_ROOT`:

```text
static/rack_elevation.css
static/cable_trace.css
```

These files are currently opened directly while rendering SVG output:

```python
# dcim/svg/racks.py
with open(f'{settings.STATIC_ROOT}/rack_elevation.css') as css_file:
```

```python
# dcim/svg/cables.py
with open(f'{settings.STATIC_ROOT}/cable_trace.css') as css_file:
```

As a result, CI fails with `FileNotFoundError` in tests such as:

```text
dcim.tests.test_api.RackTest.test_get_rack_elevation_svg
dcim.tests.test_cablepaths.LegacyCablePathTests
dcim.tests.test_cablepaths2.CablePathTests
```

Example failure:

```text
FileNotFoundError: [Errno 2] No such file or directory:
'/home/runner/work/netbox/netbox/netbox/static/rack_elevation.css'
```

and:

```text
FileNotFoundError: [Errno 2] No such file or directory:
'/home/runner/work/netbox/netbox/netbox/static/cable_trace.css'
```

Possible approaches:

* Ensure the backend test job builds or collects the minimal static assets required by these tests.
* Have the frontend/static-assets job provide the generated CSS files as artifacts consumed by the backend test job.
* Keep these specific CSS files available in the test environment independently of the frontend job split.

### Justification

The CI split should reduce unnecessary work while preserving the existing test coverage and keeping backend tests self-contained.

At the moment, tests which validate rack elevation SVG rendering and cable trace SVG rendering depend on CSS files that are no longer present in the backend test environment. This causes otherwise valid CI runs to fail even though the tests themselves have not changed.

This is related to netbox-community/netbox#22114, which aimed to skip CI when no applicable files are modified, but the current split appears to leave some backend tests without the static assets they require.

pheus:
### Fixes: #22150

Run `collectstatic` before the Python test suite so committed static assets are available under `STATIC_ROOT`.

Some SVG rendering code reads CSS files such as `rack_elevation.css` and `cable_trace.css` directly from `STATIC_ROOT`. These files already ship with NetBox under `project-static/dist/`, but after the recent CI split the Python test job no longer prepares collected static files before running the tests.

This keeps the application code unchanged and avoids making the Python test matrix depend on frontend build artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: read
 
 # Add concurrency group to control job running
 concurrency:
@@ -124,6 +125,11 @@ jobs:
 
       - name: Check for missing migrations
         run: python netbox/manage.py makemigrations --check
+
+      # Copy frontend-generated files into STATIC_ROOT before SVG rendering
+      # tests read their CSS directly.
+      - name: Collect static files
+        run: python netbox/manage.py collectstatic --no-input
 
       - name: Run tests
         if: ${{ ! matrix.coverage }}


pheus:

pheus:
@claude /review

Point 1: For the artifact action versions: I think this pairing is intentional. The current `actions/upload-artifact` documentation uses `actions/upload-artifact@v7`, while `actions/download-artifact` documents `actions/download-artifact@v8`. These actions appear to have independent major versions now, so I don't think they need to match one-for-one.

Point 2: It's addressed.

jeremystretch (commented):
Could we instead just update the code to reference the CSS stylesheets that ship in `project-static/dist/`? The CSS files already ship with NetBox so we shouldn't need to rely on the CI job to generate them.

pheus:
Thanks, agreed that we should not need to generate or pass frontend artifacts around here.

I took another look and I think the simplest fix is to run `collectstatic` in the Python test job instead. The SVG rendering code currently reads these stylesheets from `STATIC_ROOT`, and the CSS files already ship in `project-static/dist/`, so `collectstatic` will copy the committed assets into the expected location before the tests run.

That keeps the application code unchanged, avoids adding a frontend artifact dependency to the Python test matrix, and makes the CI environment closer to a normal deployment where static files have been collected.

pheus:

pheus:
> Could we instead just update the code to reference the CSS stylesheets that ship in `project-static/dist/`? The CSS files already ship with NetBox so we shouldn't need to rely on the CI job to generate them.

Thanks, I considered that approach, but I think keeping the application code unchanged is the better fit for this PR.

The SVG rendering code currently reads these stylesheets from `STATIC_ROOT`, which is the collected static files location. The CSS files already ship in `project-static/dist/`; the CI failure is that the Python test job no longer populates `STATIC_ROOT` before exercising production SVG rendering code.

I’m going to simplify this PR to just run `collectstatic` before the Python tests. That keeps the CI environment closer to a normal deployment, avoids adding frontend build/artifact handling, and avoids changing application logic to address a CI setup issue.

pheus:
I’m going to keep `pull-requests: read` here because the workflow relies on `dorny/paths-filter` for `pull_request` events, and that action needs access to the PR file list to evaluate the filters correctly. While the new `collectstatic` step itself does not require this permission, the scope of this PR is fixing the CI workflow behavior around the changed-file filtering and Python test setup.

jeremystretch (approved):
Nice, very clean.

PR #22153, opened 2026-05-08
Closes #22077: Move API test authentication setup to setUpTestData

### LINKED ISSUE #22077: Move APITestCase user/token to `setUpTestData`
### Proposed Changes

**Problem**: `setUp` creates user, permissions wiring, and token per test. Class-level fixturing is the correct shape; only per-test `add_permissions(*self.user_permissions)` need stay in `setUp` (those rows must roll back per test).

**Change**: In `netbox/utilities/testing/api.py:47`, hoist `cls.user`, `cls.token`, and `cls.header` into `setUpTestData`. Keep `add_permissions` call in `setUp`. Audit subclasses that override `setUp` — none should mutate `self.user`.

### Justification

Avoid repeated user provisioning for individual tests.

pheus:
### Fixes: #22077

This PR updates the API test setup flow so the shared test user, token, and authorization header are created once in the base `APITestCase.setUpTestData()` instead of being recreated before every test.

To preserve the base setup behavior, this also adds `super().setUpTestData()` to API test classes that override `setUpTestData()`. The calls are added after each class’s existing fixture setup to keep the current object creation order intact.

The per-test permission setup remains in `setUp()`, so permissions continue to be isolated and rolled back between individual tests.

### DIFF
--- a/netbox/circuits/tests/test_api.py
+++ b/netbox/circuits/tests/test_api.py
@@ -58,6 +58,8 @@ class ProviderTest(APIViewTestCases.APIViewTestCase):
             },
         ]
 
+        super().setUpTestData()
+
 
 class CircuitTypeTest(APIViewTestCases.APIViewTestCase):
     model = CircuitType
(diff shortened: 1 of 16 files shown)


pheus:

pheus:
After benchmarking the API test suite with 3 runs on both `main` and this branch, the results were within noise (~2m19s avg on `main` vs ~2m24s avg on this branch), with median timings nearly identical. Given that the change adds complexity to the test setup without a measurable performance benefit in practice, closing this PR.

PR #22160, opened 2026-05-11
Closes #22081: Include plaintext when creating v2 tokens via REST API

### LINKED ISSUE #22081: TokenViewSet.create() does not return v2 token plaintext

jeremystretch:
### Closes: #22081

- Override `create()` on TokenSerializer to attach the plaintext for each new token to the request on creation
- Override `to_representation()` to populate the plaintext (saved on the request) in the response data
- Introduce tests for individual & bulk token creation

### DIFF
--- a/netbox/users/api/serializers_/tokens.py
+++ b/netbox/users/api/serializers_/tokens.py
@@ -57,6 +57,24 @@ class TokenSerializer(ValidatedModelSerializer):
 
         return super().validate(data)
 
+    def create(self, validated_data):
+        instance = super().create(validated_data)
+        # The plaintext token is only available in memory after save(); v2 tokens persist only an
+        # HMAC digest, so it can't be recovered later. Stash it on the request so to_representation()
+        # can return it even after the viewset re-fetches the instance from the database.
+        if request := self.context.get('request'):
+            if not hasattr(request, '_token_plaintexts'):
+                request._token_plaintexts = {}
+            request._token_plaintexts[instance.pk] = instance.token
+        return instance
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        if not data.get('token') and (request := self.context.get('request')):
+            if plaintext := getattr(request, '_token_plaintexts', {}).get(instance.pk):
+                data['token'] = plaintext
+        return data
+
 
 class TokenProvisionSerializer(TokenSerializer):
     user = UserSerializer(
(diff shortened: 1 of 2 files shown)


jeremystretch:

jeremystretch:
@claude /review I've addressed the concern above. Please give this another review.

arthanson (approved):

PR #22166, opened 2026-05-11
Fixes #22163: Require Circuit Termination target when Termination Type is selected

### LINKED ISSUE #22163: CircuitTerminationForm raises ValueError: has no field named 'termination_id' when termination_type is set but the target object is blank
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.5

### Python Version

3.12

### Steps to Reproduce

 NetBox version: 4.5.5 (reproduces on current 4.5.x; same form code is unchanged through v4.5.10)

  Steps to reproduce
  1. Circuits → Circuit Terminations → Add
  2. Select a Circuit, set Termination side = A
  3. Set Termination type = Circuits > Provider Network
  4. Leave the Provider Network field blank
  5. Click Create

### Expected Behavior

 Inline validation error on the Provider Network field (e.g. "This field is required").

### Observed Behavior

HTTP 500 with:
<class 'ValueError'>
'CircuitTerminationForm' has no field named 'termination_id'

pheus:
### Fixes: #22163

This PR updates circuit termination validation so a target object is required whenever a termination type is selected.

A `CircuitTermination` without a termination object is not useful, and the previous behavior could fall through to generic GFK validation and produce an error for `termination_id` instead of the visible form field. With this change, the form now reports the missing target inline, and the model includes the same guard to prevent the invalid state outside the form.

### DIFF
--- a/netbox/circuits/forms/model_forms.py
+++ b/netbox/circuits/forms/model_forms.py
@@ -1,6 +1,6 @@
 from django import forms
 from django.contrib.contenttypes.models import ContentType
-from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.utils.translation import gettext_lazy as _
 
 from circuits.choices import (
@@ -24,7 +24,7 @@ from utilities.forms.fields import (
 from utilities.forms.mixins import DistanceValidationMixin
 from utilities.forms.rendering import FieldSet, InlineFields, M2MAddRemoveFields
 from utilities.forms.widgets import DatePicker, HTMXSelect, NumberWithOptions
-from utilities.templatetags.builtins.filters import bettertitle
+from utilities.string import title
 
 __all__ = (
     'CircuitForm',
@@ -195,13 +195,11 @@ class CircuitTerminationForm(NetBoxModelForm):
     termination_type = ContentTypeChoiceField(
         queryset=ContentType.objects.filter(model__in=CIRCUIT_TERMINATION_TERMINATION_TYPES),
         widget=HTMXSelect(),
-        required=False,
         label=_('Termination type')
     )
     termination = DynamicModelChoiceField(
         label=_('Termination'),
         queryset=Site.objects.none(),  # Initial queryset
-        required=False,
         disabled=True,
         selector=True
     )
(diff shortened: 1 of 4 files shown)


pheus:

pheus:
@claude /review I've addressed both points

netbox/circuits/forms/model_forms.py
```diff
@@ -195,13 +195,11 @@ class CircuitTerminationForm(NetBoxModelForm):
     termination_type = ContentTypeChoiceField(
         queryset=ContentType.objects.filter(model__in=CIRCUIT_TERMINATION_TERMINATION_TYPES),
         widget=HTMXSelect(),
-        required=False,
```
jeremystretch, line 198: As Claude points out, this introduces a regression. NetBox supports creating a circuit termination with no terminating object assigned.

netbox/circuits/models/circuits.py
```diff
@@ -17,6 +17,7 @@
     TagsMixin,
 )
 from netbox.models.mixins import DistanceMixin
+from utilities.templatetags.builtins.filters import bettertitle
```
jeremystretch:
```suggestion
from utilities.string import title
```

netbox/circuits/models/circuits.py
```diff
@@ -367,6 +368,13 @@ def get_absolute_url(self):
         return reverse('circuits:circuittermination', args=[self.pk])
 
     def clean(self):
+        if self.termination_type and not (self.termination or self.termination_id):
+            termination_type = self.termination_type.model_class()
+            raise ValidationError(
+                _("Please select a {termination_type}.").format(
+                    termination_type=_(bettertitle(termination_type._meta.verbose_name))
```
jeremystretch: We can avoid proxying through the template tag and just call `title()` directly.
```suggestion
                    termination_type=_(title(termination_type._meta.verbose_name))
```

jeremystretch (changes requested):

netbox/circuits/forms/model_forms.py
pheus: Thanks, I double-checked this. The DB fields are nullable, so a raw ORM save can technically persist a termination with no type/object because Django does not call `full_clean()` on `save()`. Through the normal NetBox form/API validation paths, though, `CircuitTermination.clean()` still rejects `termination is None`.

https://github.com/netbox-community/netbox/blob/64d3b114bc68e152869b964d54b220ecf1d50880/netbox/circuits/models/circuits.py#L369-L373

jeremystretch (approved):

PR #22167, opened 2026-05-11
Closes #22104: Avoid retracing paths when deleting cables

### LINKED ISSUE #22104: Skip retracing cable paths during cable deletion
### NetBox Version

v4.5.10

### Python Version

3.12

### Area(s) of Concern

- [ ] User Interface
- [ ] REST API
- [ ] GraphQL API
- [x] Python ORM
- [ ] Other

### Observations

When a cable is deleted, `nullify_connected_endpoints()` fires per cascaded CableTermination, retracing every CablePath through the cable. `retrace_cable_paths()` then fires for the Cable itself, retracing the same paths again. For a 2-termination cable, each affected CablePath is retraced up to three times, and each retrace re-walks the path from origin (multiple queries).

### Proposed Changes

We can tweak `nullify_connected_endpoints()` to skip the retrace when the parent Cable is also being deleted in the same operation. This can be accomplished by setting a temporary flag (e.g. `_is_being_deleted`) on the Cable instance in memory for reference by the signal handler.

jeremystretch:

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -1,5 +1,6 @@
 import itertools
 import logging
+import threading
 from collections import Counter
 
 from django.contrib.contenttypes.fields import GenericForeignKey
@@ -76,6 +77,11 @@ class Cable(PrimaryModel):
     """
     A physical connection between two endpoints.
     """
+    # Per-thread tracking of Cable PKs currently in delete(); referenced by
+    # dcim.signals.nullify_connected_endpoints to skip per-CableTermination
+    # cable path retracing during cascade (retrace_cable_paths handles it once).
+    _deletion_tracking = threading.local()
+
     type = models.CharField(
         verbose_name=_('type'),
         max_length=50,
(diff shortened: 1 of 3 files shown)


jeremystretch:
@claude /review

jeremystretch:
@claude /review The earlier concern regarding thread safety should now be addressed.

pheus (changes requested):
Thanks! The change looks good to me overall!

Since this touches cable deletion/path invalidation, could you please expand `test_101_interface_to_interface()` with a small assertion that the connected interfaces are fully cleaned up after deleting the cable?

```python
interface1.refresh_from_db()
interface2.refresh_from_db()

self.assertIsNone(interface1.cable_id)
self.assertEqual(interface1.cable_end, '')
self.assertPathIsNotSet(interface1)

self.assertIsNone(interface2.cable_id)
self.assertEqual(interface2.cable_end, '')
self.assertPathIsNotSet(interface2)
```

That should help guard against stale endpoint or `_path` data.

pheus (approved):
Nice! Thanks!

PR #22170, opened 2026-05-11
Fixes #22079: Restrict environment_params import_string to an allowlist

### LINKED ISSUE #22079: ExportTemplate `environment_params` can be configured to bypass template sandbox
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.4 (and current `main`, byte-identical for `extras/models/mixins.py` since the introduction commit `063d1fef` on 2025-07-29)

### Python Version

3.12

### Steps to Reproduce

This vulnerability was coordinated through VulnCheck (CNA), who reached out to `security@netboxlabs.com` on 2026-03-10 on my behalf. There has been no maintainer response since, so I am opening this issue as a public tracking stub. Tracked as CVE-2026-29514 (CVSS 8.8 High, CWE-94).

1. Deploy a current NetBox instance using the official `netbox-docker` release image.
2. Create one low-privileged user (non-staff, non-superuser) with `extras.add_exporttemplate`, `extras.view_exporttemplate`, and view permission on any object type with existing data (e.g. `dcim.site`).
3. Authenticate and obtain an API token for that user.
4. Create an `ExportTemplate` via the API:
   ```json
   POST /api/extras/export-templates/
   {
     "name": "poc",
     "object_types": ["dcim.site"],
     "template_code": "{{ \"id\" }}",
     "environment_params": {"finalize": "subprocess.getoutput"}
   }
   ```
5. Trigger render: `GET /api/dcim/sites/?export=poc`.

The same root cause applies to `ConfigTemplate` via `POST /api/extras/config-templates/<id>/render/` (`extras.add_configtemplate` + `extras.render_configtemplate`).

### Expected Behavior

`environment_params['finalize']` should not be able to resolve to arbitrary callables via `import_string()`. Only known-safe values (e.g. the four jinja2 undefined classes for `undefined`) should be importable; any other value should be rejected with a `ValidationError`.

### Observed Behavior

The HTTP response body contains the output of the shell command (`uid=...(netbox) ...`) executed as the NetBox service user.

`RenderTemplateMixin.get_environment_params()` (in `netbox/extras/models/mixins.py`) calls `import_string(value)` for any string value of the `finalize` or `undefined` keys with no allowlist. The resolved callable is passed as `finalize=` to `SandboxedEnvironment`, where it is invoked on every rendered `{{ expression }}` outside the sandbox's call interception, enabling arbitrary code execution at render time.

A fix is proposed in #22078 (allowlist on path-import params, validation at `clean()` and at render time, with tests).

jnovinger:
### Fixes: #22079

Replaces `JINJA_ENV_PARAMS_WITH_PATH_IMPORT` with `JINJA_ENV_PARAMS_ALLOWED`, a whitelist of permitted Jinja2 Environment parameters. Only keys in the allowlist may appear in `environment_params`; unknown keys are rejected by `clean()` and stripped at render time.

The `undefined` parameter resolves values via direct class mapping (no `import_string()`). All safe scalar and string Jinja2 Environment parameters are included in the allowlist to avoid breaking existing templates.

The `finalize` parameter is deprecated and blocked from new use via `clean()`. Existing stored `finalize` values continue to resolve via `import_string()` at render time to preserve backward compatibility. No data migrations; no existing template data is altered on upgrade.

### Behavioral notes

- **Editing templates with `finalize`**: `clean()` rejects `finalize` on every save, including saves of existing templates that already have it stored. This means a user editing an unrelated field (name, description) on such a template would need to remove `finalize` first. This is a deliberate deprecation nudge, but if a gentler transition is preferred (only validate newly-added or changed keys), we can adjust `clean()` to skip validation for pre-existing stored values.

- **`extensions`, `loader`, `bytecode_cache`**: These are excluded from the allowlist because they accept callables or trigger internal imports. Existing templates using these keys will have them silently stripped at render time. `loader` and `bytecode_cache` are not JSON-serializable so are unlikely to exist in stored data. We don't have visibility into whether `extensions` is used in practice.

- **`finalize` removal**: `finalize` will be removed in a future release.

### DIFF
--- a/netbox/extras/constants.py
+++ b/netbox/extras/constants.py
@@ -1,3 +1,5 @@
+from jinja2 import ChainableUndefined, DebugUndefined, StrictUndefined, Undefined
+
 from core.events import *
 from extras.choices import LogLevelChoices
 
@@ -32,11 +34,47 @@ WEBHOOK_EVENT_TYPES = {
     JOB_ERRORED: 'job_ended',
 }
 
-# Jinja environment parameters which support path imports
-JINJA_ENV_PARAMS_WITH_PATH_IMPORT = (
-    'undefined',
-    'finalize',
-)
+# Allowed Jinja2 environment parameters and their permitted values.
+# Only keys listed here may appear in a template's environment_params.
+#   None  = any JSON-serializable value accepted (scalars, booleans, etc.)
+#   dict  = only dict keys accepted; dict values are the resolved Python objects
+#
+# Note: 'finalize' is intentionally absent. It is deprecated and handled as a
+# legacy carve-out in RenderTemplateMixin (blocked from new use, but existing
+# stored values continue to resolve via import_string at render time).
+JINJA_ENV_PARAMS_ALLOWED = {
+    # Boolean / scalar params (accept any JSON-serializable value)
+    'auto_reload': None,
+    'autoescape': None,
+    'cache_size': None,
+    'enable_async': None,
+    'keep_trailing_newline': None,
+    'lstrip_blocks': None,
+    'optimized': None,
+    'trim_blocks': None,
+    # String params (template syntax delimiters)
+    'block_start_string': None,
+    'block_end_string': None,
+    'comment_start_string': None,
+    'comment_end_string': None,
+    'line_comment_prefix': None,
+    'line_statement_prefix': None,
+    'newline_sequence': None,
+    'variable_start_string': None,
+    'variable_end_string': None,
+    # Mapped params (value must be a key in the dict; resolved to the dict value)
+    'undefined': {
+        'jinja2.ChainableUndefined': ChainableUndefined,
+        'jinja2.DebugUndefined': DebugUndefined,
+        'jinja2.StrictUndefined': StrictUndefined,
+        'jinja2.Undefined': Undefined,
+    },
+    # Excluded (dangerous — accept callables or trigger imports):
+    #   'bytecode_cache' — accepts arbitrary object
+    #   'extensions'     — Jinja2 internally calls import_string() on string entries
+    #   'finalize'       — deprecated; legacy carve-out in RenderTemplateMixin
+    #   'loader'         — accepts arbitrary object
+}
 
 # Dashboard
 DEFAULT_DASHBOARD = [
--- a/netbox/extras/models/mixins.py
+++ b/netbox/extras/models/mixins.py
@@ -4,6 +4,7 @@ import os
 import sys
 from collections import defaultdict
 
+from django.core.exceptions import ValidationError
 from django.core.files.storage import storages
 from django.db import models
 from django.http import HttpResponse
@@ -11,7 +12,7 @@ from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
 
 from core.models import ObjectType
-from extras.constants import DEFAULT_MIME_TYPE, JINJA_ENV_PARAMS_WITH_PATH_IMPORT
+from extras.constants import DEFAULT_MIME_TYPE, JINJA_ENV_PARAMS_ALLOWED
 from extras.utils import filename_from_model, filename_from_object
 from utilities.jinja2 import render_jinja2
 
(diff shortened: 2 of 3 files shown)


jnovinger:

jnovinger:
> **1. `DummyTemplate` registered in Django's app registry without isolation**

Switched to `@isolate_apps('extras')` with `DummyTemplate` nested inside the test class.

> **2. `finalize` with non-string JSON values bypasses `clean()` silently**

Moved `finalize` from `JINJA_ENV_PARAM_IMPORT_ALLOWLIST` to `JINJA_ENV_PARAMS_BLOCKED_KEYS` so it's rejected regardless of value type. Added a test for non-string values.

> **3. Dict mutated while iterating in `get_environment_params()`**

Changed to a two-pass approach: collect resolved imports into a separate dict, then apply after iteration.

All three fixes have been squashed into the original commits rather than added as separate fixup commits.

jnovinger:

In particular, see my response to your points from the last review in https://github.com/netbox-community/netbox/pull/22170#issuecomment-4431718892.

netbox/extras/migrations/0140_block_unsafe_environment_params.py
jeremystretch, line 1: Can these be merged into a single migration? We can probably just hit on all the keys at once per object.

netbox/extras/migrations/0139_sanitize_environment_params.py
```diff
@@ -0,0 +1,35 @@
...
+    'jinja2.Undefined',
+}
+
+
+def sanitize_environment_params(apps, schema_editor):
+    for model_name in ('ExportTemplate', 'ConfigTemplate'):
+        Model = apps.get_model('extras', model_name)
+
+        for obj in Model.objects.filter(environment_params__has_key='finalize'):
+            del obj.environment_params['finalize']
```
jeremystretch: We can't blindly remove this parameter, as it is highly likely to break legitimate uses of the functionality. Consider an automation workflow which feeds a rendered export template into some other system that depends on some specific `finalize` behavior to adapt the data.

netbox/extras/migrations/0139_sanitize_environment_params.py
```diff
@@ -0,0 +1,35 @@
...
+            obj.save(update_fields=['environment_params'])
+
+        for obj in Model.objects.filter(environment_params__has_key='undefined'):
+            value = obj.environment_params.get('undefined')
+            if type(value) is not str or value not in ALLOWED_UNDEFINED_VALUES:
+                del obj.environment_params['undefined']
```
jeremystretch: This is also a potential breaking change, granted the risk is far lower as the default handlers provided by Jinja suffice for most use cases.

netbox/extras/migrations/0140_block_unsafe_environment_params.py
```diff
@@ -0,0 +1,24 @@
...
+
+BLOCKED_KEYS = ('autoescape', 'bytecode_cache', 'extensions', 'loader')
+
+
+def remove_blocked_keys(apps, schema_editor):
+    for model_name in ('ExportTemplate', 'ConfigTemplate'):
+        Model = apps.get_model('extras', model_name)
+        for key in BLOCKED_KEYS:
+            for obj in Model.objects.filter(environment_params__has_key=key):
+                del obj.environment_params[key]
```
jeremystretch: We definitely can't blanket remove parameters like `autoescape`. We should at least support boolean values.

netbox/extras/constants.py
```diff
@@ -32,11 +32,29 @@
...
+}
+
+# Keys in environment_params that accept callables or trigger import resolution
+# inside Jinja2 itself. These are blocked from user input entirely, regardless
+# of value type.
+JINJA_ENV_PARAMS_BLOCKED_KEYS = frozenset({
+    'autoescape',     # Accepts Union[bool, Callable]
+    'bytecode_cache',  # Accepts cache object
+    'extensions',     # Jinja2 internally calls import_string() on string entries
+    'finalize',       # Callable invoked on every expression, outside sandbox (CVE-2026-29514)
```
jeremystretch: If the intent is to permanently remove support for `finalize`, it needs to be deprecated first so that users have time to adapt between releases.

netbox/extras/constants.py
```diff
@@ -32,11 +32,29 @@
...
+        'jinja2.DebugUndefined',
+        'jinja2.StrictUndefined',
+        'jinja2.Undefined',
+    }),
+}
+
+# Keys in environment_params that accept callables or trigger import resolution
+# inside Jinja2 itself. These are blocked from user input entirely, regardless
+# of value type.
+JINJA_ENV_PARAMS_BLOCKED_KEYS = frozenset({
```
jeremystretch: It would probably be safer to define a whitelist of parameters that we _do_ allow rather than blacklisting those forbidden. Jinja could introduce additional potentially dangerous parameters at any time.

netbox/extras/constants.py
```diff
@@ -32,11 +32,29 @@
...
-# Jinja environment parameters which support path imports
-JINJA_ENV_PARAMS_WITH_PATH_IMPORT = (
-    'undefined',
-    'finalize',
-)
+# Jinja environment parameters which support path imports, mapped to allowed values.
+# Only dotted paths in the frozenset may be resolved via import_string().
+# An empty frozenset means no values are permitted (the key is effectively blocked
+# from import resolution).
+JINJA_ENV_PARAM_IMPORT_ALLOWLIST = {
```
jeremystretch: It might be simpler to just maintain a dictionary mapping names to the actual Python classes and allowing only dictionary keys in the param data, to avoid the dynamic import.

jeremystretch (changes requested):
We can't ship migrations that introduce a breaking change in a patch release, especially with the inherent risk of silently altering the data returned by an export template. It's arguably reasonable to stop supporting these keys in new or modified templates, but we can't alter data in place.

At the extreme, we could potentially _disable_ an offending export template: That would trigger an obvious error when attempting to use the template rather than altering the data being returned. Unfortunately, the ConfigTemplate and ExportTemplate models lack an `enabled` toggle.

jnovinger:

Significant rewrite since last review. Key changes:

1. **Finalize legacy carve-out**: `finalize` is no longer in `JINJA_ENV_PARAMS_ALLOWED`. It's handled as a separate legacy path: `clean()` blocks new use, but `get_environment_params()` still resolves existing stored values via `import_string()` to preserve backward compatibility. The maintainer requires that existing templates not break or produce wrong output.

2. **Complete allowlist**: Added all safe Jinja2 Environment scalar/string params (`line_statement_prefix`, `line_comment_prefix`, `optimized`, `cache_size`, `auto_reload`, `enable_async`) to avoid silently stripping params that existing templates may use.

3. **Refactored `get_environment_params()`**: Split into three testable static methods (`_filter_environment_params`, `_resolve_mapped_params`, `_resolve_finalize`) with dedicated test classes for each.

4. **Simplified `clean()` check**: Replaced `allowed is None and key not in ...` with straightforward `key not in JINJA_ENV_PARAMS_ALLOWED` membership check.

5. **Consistent return semantics**: All three helper methods return new dicts instead of mutating input.

jnovinger:
Rewrote from scratch based on your review. Single commit, no data migrations.

> It would probably be safer to define a whitelist of parameters that we _do_ allow rather than blacklisting those forbidden.

Done. `JINJA_ENV_PARAMS_ALLOWED` is a whitelist of every safe Jinja2 Environment parameter. Unknown keys are rejected by `clean()` and stripped at render time.

> It might be simpler to just maintain a dictionary mapping names to the actual Python classes and allowing only dictionary keys in the param data, to avoid the dynamic import.

Values map directly to Python classes via dict lookup. No `import_string()` for allowlisted params.

> If the intent is to permanently remove support for `finalize`, it needs to be deprecated first so that users have time to adapt between releases.

`finalize` is deprecated: `clean()` blocks new use, but existing stored values continue to resolve via `import_string()` at render time. This preserves backward compatibility for existing templates. The residual `import_string()` call is limited to this one legacy path; exploiting it requires pre-existing DB data or direct database access, neither of which is reachable through the API or UI.

> We can't blindly remove this parameter / We can't ship migrations that introduce a breaking change

No data migrations. No existing data is altered on upgrade. Existing `finalize` templates continue to work.

> We definitely can't blanket remove parameters like `autoescape`. We should at least support boolean values.

`autoescape` and all other safe Jinja2 Environment params are in the allowlist.

> Can these be merged into a single migration?

No migrations in this version.

jnovinger:
> One issue worth addressing:
> 
> **`clean()` makes existing templates with `finalize` (or other previously-used non-allowlisted keys) unsaveable after upgrade**
> 
> `mixins.py:144` raises a `ValidationError` for any template whose `environment_params` contains `finalize`, with no exception for templates that already have it stored. `NetBoxModelSerializer.validate()` calls `full_clean()` on the instance (`base.py:114`), and UI forms do too. This means any existing template with `finalize` (or any key not in `JINJA_ENV_PARAMS_ALLOWED`, such as `extensions` or `bytecode_cache`) will fail every save attempt after upgrade — including routine edits to unrelated fields like `name` or `template_code`.
> 
> This contradicts the stated backward-compatibility requirement from the maintainer: _"we can't ship migrations that introduce a breaking change"_ and _"existing templates not break."_ Templates that worked before the upgrade silently become unsaveable.
> 
> The `get_environment_params()` path handles this correctly — it strips unknown keys and preserves `finalize` for render. The inconsistency is only in `clean()`.
 
This was an intentional decision in this particular iteration. But, I'm happy to change tack if required.

jeremystretch (approved):
Thanks @jnovinger!

PR #22185, opened 2026-05-13
Closes #22181: Clarify plugin package installation and removal instructions

### LINKED ISSUE #22181: Clarify plugin installation permissions, root shell usage, and local_requirements.txt
### Change Type

Addition

### Area

Plugins

### Proposed Changes

The current plugin installation documentation instructs users to activate the NetBox virtual environment and run `pip install <package>` directly:

    $ source /opt/netbox/venv/bin/activate
    (venv) $ pip install <package>

That is technically correct when the current user has write access to the virtual environment. However, it can be confusing for users following the official installation guide, where NetBox is installed under `/opt/netbox` and many installation steps are performed with elevated privileges.

In that scenario, a user may have the NetBox virtual environment activated and see the `(venv)` prompt, but still receive a pip error such as:

    ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied:
    '/opt/netbox/venv/lib/pythonX.Y/site-packages/...'
    Check the permissions.

I would like to propose adding a short clarification to the plugin installation documentation covering three points:

1. Installing packages into the NetBox virtual environment requires write permissions to the virtual environment. For installations under `/opt/netbox`, this commonly means using root privileges.
2. If using a root shell, the virtual environment should be activated from within that root shell.
3. For persistent plugin installations, users should be encouraged to add the plugin package to `/opt/netbox/local_requirements.txt` and then run the normal NetBox upgrade process.

It may also help to make the shell examples more explicit by using `$` for an unprivileged user prompt and `#` for a root prompt.

Suggested wording for `docs/plugins/installation.md`, under "Install the Python Package":

> For production installations under `/opt/netbox`, the NetBox virtual environment may not be writable by a regular user. Activating the virtual environment does not change file permissions, so running `pip install` as an unprivileged user may result in a `Permission denied` error.
>
> One option is to switch to a root shell first, then activate the NetBox virtual environment and install the plugin from there:
>
>     $ sudo -i
>     # cd /opt/netbox
>     # source venv/bin/activate
>     (venv) # pip install <package>
>
> Alternatively, pip can be run directly using the Python executable inside the NetBox virtual environment:
>
>     $ sudo /opt/netbox/venv/bin/python3 -m pip install <package>
>
> For persistent installations, it is recommended to add the plugin package to `/opt/netbox/local_requirements.txt` and then run the normal NetBox upgrade script:
>
>     $ sudo sh -c "echo '<package>' >> /opt/netbox/local_requirements.txt"
>     $ sudo /opt/netbox/upgrade.sh
>
> This ensures the plugin is installed as part of the standard NetBox installation and upgrade process, and that it is reinstalled if the virtual environment is rebuilt.
>
> In the examples above, `$` indicates a regular user shell and `#` indicates a root shell.

This would help avoid confusion where users follow the plugin installation guide, confirm they are in the virtual environment, but still encounter permission errors from pip.

It would also align the plugin installation documentation with the existing installation and upgrade flow, where optional Python packages can be managed through `local_requirements.txt`.

Related links:

- https://github.com/netbox-community/netbox/discussions/6551
- https://github.com/netbox-community/netbox/discussions/19305

pheus:
### Fixes: #22181

This updates the plugin documentation to make the recommended installation and removal flow clearer and more durable.

The plugin installation docs now lead with `/opt/netbox/local_requirements.txt` and NetBox's `upgrade.sh` script for production installs, which helps ensure plugins are reinstalled if the virtual environment is rebuilt. The update also clarifies virtual environment write permissions, explains the `$` and `#` shell prompt conventions, and provides safe direct-install examples using `sudo`.

Additionally, this normalizes related command examples across the plugin docs and adds guidance for removing plugin entries from `local_requirements.txt` during uninstall so they are not reinstalled on the next upgrade.

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -57,7 +57,7 @@ In order to send email, NetBox needs an email server configured. The following i
 Email is sent from NetBox only for critical events or if configured for [logging](#logging). If you would like to test the email server configuration, Django provides a convenient [send_mail()](https://docs.djangoproject.com/en/stable/topics/email/#send-mail) function accessible within the NetBox shell:
 
 ```no-highlight
-# python ./manage.py nbshell
+(venv) $ python3 ./manage.py nbshell
 >>> from django.core.mail import send_mail
 >>> send_mail(
   'Test Email Subject',
--- a/docs/plugins/installation.md
+++ b/docs/plugins/installation.md
@@ -7,12 +7,32 @@
 
 Download and install the plugin's Python package per its installation instructions. Plugins published via PyPI are typically installed using the [`pip`](https://packaging.python.org/en/latest/tutorials/installing-packages/) command line utility. Be sure to install the plugin within NetBox's virtual environment.
 
+For production installations, the recommended method is to add the package to `/opt/netbox/local_requirements.txt` and then run NetBox's upgrade script. This installs the plugin as part of the standard NetBox installation and upgrade process and ensures that it will be reinstalled if the virtual environment is rebuilt.
+
 ```no-highlight
-$ source /opt/netbox/venv/bin/activate
-(venv) $ pip install <package>
+$ sudo sh -c "echo '<package>' >> /opt/netbox/local_requirements.txt"
+$ sudo /opt/netbox/upgrade.sh
 ```
 
-Alternatively, you may wish to install the plugin manually by running `python setup.py install`. If you are developing a plugin and want to install it only temporarily, run `python setup.py develop` instead.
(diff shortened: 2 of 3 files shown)


pheus:

pheus:
@claude /review

I agree with the first point, with one nuance: `upgrade.sh` only applies migrations and collects static files for the plugin if the plugin is already enabled in `configuration.py` when the script runs. Since the current flow may have users run `upgrade.sh` only to install the package before enabling the plugin, I added a note clarifying when the later migration/static-file steps can be skipped and when they should still be run manually.

For the second point, I don’t think `sudo` should be added for `nbshell`. The permission issue is specific to modifying the virtual environment with `pip`; running management commands does not require write access to the venv. To avoid ambiguity,

arthanson (approved):

PR #22186, opened 2026-05-13
Fixes #22183: Fix Wireless Link signal causing unnecessary Interface Change-Log Entries

### LINKED ISSUE #22183: Saving an unchanged WirelessLink creates spurious change-log entries for `interface_b`
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

1. Create two Interface objects of a wireless type on a Device.
2. Create a `WirelessLink` between them and save it.
3. Save the `WirelessLink` again without changing anything (e.g. via the UI edit form, or `link.save()` in nbshell).
4. Look at the change log for `interface_b`.

### Expected Behavior

Re-saving an unchanged `WirelessLink` should not create a change-log entry on either connected interface, the same way `interface_a` is already handled.

### Observed Behavior

`interface_b` gets re-saved on every `WirelessLink` save, producing a spurious change-log entry each time. `interface_a` is unaffected.

The cause is in [netbox/wireless/signals.py](https://github.com/netbox-community/netbox/blob/main/netbox/wireless/signals.py), line 32:

```
if instance.interface_a.wireless_link != instance:   # line 28 — correct
    ...
    instance.interface_a.save()
if instance.interface_b.cable != instance:           # line 32 — should be wireless_link
    ...
    instance.interface_b.wireless_link = instance
    instance.interface_b.save()
```

instance is a `WirelessLink`, but `interface_b.cable` is a FK to `Cable`. The two model types can never be equal, so the condition is always `true` and the `save()` always runs.

pheus:
### Fixes: #22183

This fixes the `update_connected_interfaces` post-save signal for wireless links so that `interface_b` checks `wireless_link` instead of `cable`, matching the existing `interface_a` logic.

Without this change, saving an unchanged `WirelessLink` unnecessarily re-saves `interface_b`, which can create spurious change-log entries.

A regression test has been added to confirm that re-saving an unchanged wireless link does not create change-log entries for either connected interface.

### DIFF
--- a/netbox/wireless/signals.py
+++ b/netbox/wireless/signals.py
@@ -29,7 +29,7 @@ def update_connected_interfaces(instance, created, raw=False, **kwargs):
         logger.debug(f"Updating interface A for wireless link {instance}")
         instance.interface_a.wireless_link = instance
         instance.interface_a.save()
-    if instance.interface_b.cable != instance:
+    if instance.interface_b.wireless_link != instance:
         logger.debug(f"Updating interface B for wireless link {instance}")
         instance.interface_b.wireless_link = instance
         instance.interface_b.save()
--- a/netbox/wireless/tests/test_models.py
+++ b/netbox/wireless/tests/test_models.py
@@ -1,0 +1,61 @@
+from uuid import uuid4
+
+from django.contrib.contenttypes.models import ContentType
+from django.test import RequestFactory, TestCase
+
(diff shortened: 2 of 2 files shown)


pheus:

jeremystretch (approved):
Thanks @pheus!

Issue #19158, opened 2025-04-11
IP Range's `get_child_ips` does not return IP Addresses with different prefix lengths


mlow:
### Deployment Type

Self-hosted

### NetBox Version

v4.2.7

### Python Version

3.11

### Steps to Reproduce

```console
$ ./manage.py shell
>>> from netaddr import IPNetwork
>>> from ipam.models import IPRange, IPAddress
>>> range = IPRange.objects.create(start_address=IPNetwork('10.0.0.2/24'), end_address=IPNetwork('10.0.0.254/24'))
>>> IPAddress.objects.create(address='10.0.0.2/32')
>>> IPAddress.objects.create(address='10.0.0.3/24')
>>> range.get_child_ips()
<RestrictedQuerySet [<IPAddress: 10.0.0.3/24>]>
```

### Expected Behavior

I would expect `get_child_ips` to return all IP addresses whose host part falls within the IP range. It seems odd that the prefix length should be considered at all when talking about IP Ranges.

Updating `get_child_ips` as follows would bring the behaviour I expect:

```diff
diff --git a/netbox/ipam/models/ip.py b/netbox/ipam/models/ip.py
index e1a8d91..ae1f498 100644
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -654,8 +654,8 @@ class IPRange(ContactsMixin, PrimaryModel):
         Return all IPAddresses within this IPRange and VRF.
         """
         return IPAddress.objects.filter(
-            address__gte=self.start_address,
-            address__lte=self.end_address,
+            address__host__inet__gte=self.start_address.ip,
+            address__host__inet__lte=self.end_address.ip,
             vrf=self.vrf
         )
 
```

### Observed Behavior

Only IP Addresses with prefix length matching that of the IP Range are returned from `IPRange.get_child_ips()`

PR #22188, opened 2026-05-13
Closes #21993: Allow IP Ranges to contain a single IP Address

### LINKED ISSUE #21993: Allow IP ranges comprising a single IP address
### NetBox version

v4.5.8

### Feature type

### Proposed functionality

This is a renewal of request [#11117](https://github.com/netbox-community/netbox/issues/11117)

Currently, the IPRange model in NetBox enforces a validation check requiring the end_address to be greater than the start_address. This prevents the creation of a range consisting of a single IP. 

**Proposed Change**

Modify the validation logic in [ipam/models/ip.py](https://github.com/netbox-community/netbox/blob/b1a810164a42e467d7a84caf043a6f5e2f369c45/netbox/ipam/models/ip.py#L585) to allow the end_address to be greater than or equal to the start_address.

### Use case

**DHCP Pools**: In many environments (such as DIA customer /30 or /31 subnets), a DHCP pool may technically consist of only one available host address.

**NAT Pools**: Non-contiguous or arbitrary NAT ranges that do not align with CIDR boundaries are modeled as IPRanges. If a specific policy requires a single-IP pool, it cannot currently be modeled the same way.

**Consistency in Automation:** External plugins or custom scripts may reference the IPRange model to represent pools. If a pool happens to be a single IP, then the Prefix or IPAddress model must be used. This breaks consistency, as these models have different Roles and API endpoints.

### Database changes

None

### External dependencies

None

pheus:
### Fixes: #21993

Allow IP ranges where `start_address` equals `end_address`, so a range can represent a single-address pool or reservation, such as a DHCP or NAT pool containing one usable address.

This keeps `IPRange` semantics unchanged:

- `end_address` remains required
- reversed ranges are still invalid
- single-address ranges are rendered with both endpoints, e.g. `192.0.2.10-192.0.2.10/24`, to avoid confusing them with `IPAddress` objects
- documentation clarifies that a single-address `IPRange` is not a replacement for an `IPAddress`

This PR intentionally does **not** change existing CIDR/mask-aware range membership behavior. During development, I've considered switching some range checks to host-only comparisons, but that would alter the existing behavior for addresses with different prefix lengths and overlaps with previously declined behavior in #19158. To keep this change focused on #21993, this PR only allows equal start/end boundaries and leaves CIDR containment behavior unchanged.

Tests were added/updated for model validation, rendering, API creation, filtersets, form creation, CSV import, and the mixed prefix IP Addresses tab.

### DIFF
--- a/docs/models/ipam/iprange.md
+++ b/docs/models/ipam/iprange.md
@@ -21,6 +21,13 @@ The [Virtual Routing and Forwarding](./vrf.md) instance in which this IP range e
 
 The beginning and ending IP addresses (inclusive) which define the boundaries of the range. Both IP addresses must specify the correct mask.
 
+A range may contain a single IP address by setting the start and end address to the same value, including the same mask length. This is useful when a pool-like construct, such as a DHCP or NAT pool, contains only one usable address.
+
+!!! note
+    A single-address IP range is not a replacement for an [IP address](./ipaddress.md). Use an IP address object when modeling an address configured on an interface, assigned as a primary IP, or otherwise participating in IP-address-specific relationships. Use an IP range only when the address is being modeled as a pool, reservation range, or similar range-oriented construct.
+
+    The end address is still required. To create a single-address range, enter the same address and mask for both the start and end address.
+
 !!! note
     The maximum supported size of an IP range is 2^32 - 1.
 
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -605,11 +605,12 @@ class IPRange(ContactsMixin, PrimaryModel):
                     'end_address': _("Starting and ending IP address masks must match")
                 })
 
-            # Check that the ending address is greater than the starting address
-            if not self.end_address > self.start_address:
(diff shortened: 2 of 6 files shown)


pheus:

pheus:
@claude /review

I've update the validation comment.

For the `@cached_property` staleness concern, I agree this can happen when an `IPRange` instance is mutated programmatically after `name` has already been accessed. I’m going to leave that unchanged in this PR because it is pre-existing behavior for `IPRange.name`, and similar cached properties such as `first_available_ip` and `utilization` have the same general caveat. Adding save-time cache invalidation would be a broader model lifecycle cleanup than this issue needs.

This PR is intentionally limited to allowing `start_address == end_address` while preserving the existing CIDR/mask-aware range behavior. In particular, it does not change `get_child_ips()`, `mark_populated` enforcement, or `contains` filtering to host-only comparisons, since that overlaps with the behavior discussed in #19158.

jeremystretch (approved):
Thanks @pheus!

PR #22193, opened 2026-05-14
Closes #22124: Add test coverage for custom management commands

### LINKED ISSUE #22124: Add tests for management commands
### Proposed Changes

Add a test case for each management command (excluding the deprecated `housekeeping` command) across all apps.

### Justification

Extends test coverage to the standalone management commands.

pheus:
### Fixes: #22124

This PR adds test coverage for NetBox’s custom Django management commands across the `core`, `dcim`, `extras`, `ipam`, and `utilities` apps.

The new tests cover command argument handling, expected error cases, and successful execution paths. Long-running or external behavior, such as RQ workers, webhook serving, search indexing, and script job enqueueing, is mocked to keep the tests focused and safe.

The deprecated `housekeeping` command is intentionally excluded, as requested in the issue.

### DIFF
--- a/netbox/core/management/commands/nbshell.py
+++ b/netbox/core/management/commands/nbshell.py
@@ -150,7 +150,7 @@ class Command(BaseCommand):
         try:
             import readline
             import rlcompleter
-        except ModuleNotFoundError:
+        except ModuleNotFoundError:  # pragma: no cover
             pass
         else:
             readline.set_completer(rlcompleter.Completer(namespace).complete)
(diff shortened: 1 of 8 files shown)


pheus:

pheus:

pheus:

pheus:
I noticed the initial coverage for some of the management command modules was still lower than expected, so I added a follow-up set of targeted tests. The PR now reaches 100% coverage for all active custom management commands, while keeping the deprecated `housekeeping` command excluded as requested.

This should now be ready for review.

bctiemann (approved):

PR #22194, opened 2026-05-14
Closes #20804: Add field selector to BulkRenameView for label field

bctiemann:
## Summary

- Adds a **Field** dropdown to the bulk rename form for device/module component models (Interface, ConsolePort, FrontPort, RearPort, PowerPort, PowerOutlet, ModuleBay, DeviceBay, InventoryItem, and their template counterparts — 20 models in total)
- When the target model has both a `name` and a `label` field, the selector lets users choose which field the find/replace pattern applies to; the dropdown defaults to **Name**
- Models that have only one renameable field (e.g. Device with only `name`, or Cable with only `label`) are unaffected — no selector is rendered

## Preview table layout

For models with both `name` and `label`, the preview table always shows **four columns** regardless of which field is selected:

| Current Name | New Name | Current Label | New Label |
|---|---|---|---|
| GigabitEthernet0/0/0 | GigabitEthernet0/1/0 | | |
| Cellular0/2/0 | Cellular0/3/0 | | |

The **New** column for the non-targeted field is left blank, making it visually clear which field the find/replace pattern applies to. The name is always visible as an identifier when renaming labels, and the label is always visible when renaming names.

For models without a `label` field (Device, Rack, Prefix, etc.), the original two-column layout is unchanged.

Screenshot 2026-05-19 at 9 27 39 AM

## Implementation details

- `BulkRenameView._get_rename_fields()` returns choices only when `self.field_name == 'name'` (the default) and the model has a `label` field in `_meta.fields`; views that already hardcode a non-default `field_name` (Cable → `label`, DeviceType → `model`, etc.) return an empty list
- The `field_name` ChoiceField is injected into the form class dynamically via `type()` so the form metaclass processes it normally
- `required=False` lets existing callers (tests, scripts) omit the field; an empty value falls back to `self.field_name`
- `BulkRenameObjectsViewTestCase.test_bulk_rename_label_field` added to the shared test mixin; auto-skips for models without a `label` field

## Test plan

- [x] Existing bulk rename tests pass (Interface, ConsolePort, PowerPort, InterfaceTemplate — 128 tests)
- [x] `test_bulk_rename_label_field` passes for Interface, ConsolePort, ConsolePortTemplate
- [x] New test verifies that `name` is unchanged when renaming `label`, and vice versa

Closes: #20804


bctiemann:
@claude /review

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -879,6 +879,15 @@ class BulkRenameView(GetReturnURLMixin, BaseMultiObjectView):
     # so "Select all N matching query" can expand across the full queryset.
     filterset = None
 
+    def _get_rename_fields(self):
```
jeremystretch: I think the root issue here is converting the view from supporting a single field to supporting multiple fields. Explicitly handling `name` and `label` feels like an anti-pattern: What if we want to support a third field for some models in the future?

I suggest tweaking the strategy here a bit to instead declare a set of supported fields on the view, e.g.

```python
class FooBulkRenameView:
    rename_fields = ('name', 'label')
```

We should keep support for `field_name` for now to ensure backward compatibility for plugins, but otherwise migrate all subclasses of BulkRenameView to use the list-based approach.

netbox/netbox/views/generic/bulk_views.py
```diff
@@ -889,19 +898,31 @@ def __init__(self, *args, **kwargs):
...
+        self.has_label_field = bool(rename_fields)
+        form_attrs = {
+            'pk': ModelMultipleChoiceField(
                 queryset=self.queryset,
-                widget=MultipleHiddenInput()
+                widget=MultipleHiddenInput(),
+            ),
+        }
+        if rename_fields:
+            form_attrs['field_name'] = ChoiceField(
```
jeremystretch: It's very likely a user might want to apply the same find/replace filter to multiple fields at once. Instead of a single-choice dropdown, we could present a series of checkboxes (perhaps all ticked by default) so the user can specify the desired field(s).

netbox/templates/generic/bulk_rename.html
```diff
@@ -34,12 +34,36 @@
 <div class="row mb-3">
     <div class="col col-md-7">
         <table class="table">
+            {% if has_label_field %}
```
jeremystretch: With the multi-field set approach suggested above, we can simplify the template logic and just iterate through the selected fields to render before & after columns. To avoid confusion, we probably want to include only fields which the user has selected.

jeremystretch (changes requested):

bctiemann:
@jeremystretch I do like that more extensible approach better:

Screenshot 2026-05-19 at 9 27 39 AM

mkarel:
Thanks for taking the time to implement this feature. I like the change to have the option to select multiple fields, curious if from a UI perspective if using a multi-select drop down for the **Fields** would be "cleaner". 

Just a thought and thanks again. looking forward to seeing this in a future release.

jeremystretch:

netbox/templates/generic/bulk_rename.html
```diff
@@ -71,4 +107,4 @@ <h2 class="card-header">{% trans "Rename" %}</h2>
         </form>
     </div>
 </div>
-{% endblock content %}
+{% endblock content %}
```
jeremystretch: Don't forget the trailing newline

netbox/templates/generic/bulk_rename.html
```diff
@@ -58,7 +64,37 @@
...
+                    {% for field in form.hidden_fields %}{{ field }}{% endfor %}
+                    {# Standard fields (find, replace, use_regex) #}
+                    {% for field in form.visible_fields %}
+                      {% if not form.meta_fields or field.name not in form.meta_fields %}
+                        {% render_field field %}
+                      {% endif %}
+                    {% endfor %}
+                    {# Per-field rename checkboxes, rendered before the changelog fieldset #}
+                    {% if rename_fields %}
+                    <div class="row mb-3">
```
jeremystretch: Needs to be indented under the `{% if %}` block above.

netbox/templates/generic/bulk_rename.html
```diff
@@ -58,7 +64,37 @@
...
+                    {# Hidden fields #}
+                    {% for field in form.hidden_fields %}{{ field }}{% endfor %}
+                    {# Standard fields (find, replace, use_regex) #}
+                    {% for field in form.visible_fields %}
+                      {% if not form.meta_fields or field.name not in form.meta_fields %}
+                        {% render_field field %}
+                      {% endif %}
+                    {% endfor %}
+                    {# Per-field rename checkboxes, rendered before the changelog fieldset #}
+                    {% if rename_fields %}
```
jeremystretch, line 70: Suggest placing the field checkboxes first in the form to avoid conflation with the "use regex" checkbox.

Image

netbox/utilities/testing/views.py
```diff
@@ -1117,6 +1117,32 @@ def test_bulk_rename_objects_with_constrained_permission(self):
             for i, instance in enumerate(self._get_queryset().filter(pk__in=pk_list)):
                 self.assertEqual(instance.name, f'{objects[i].name}X')
 
+        @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])
+        def test_bulk_rename_label_field(self):
```
jeremystretch, line 1124: Should probably add a test for renaming both the `name` and `label` fields simultaneously.

jeremystretch (changes requested):
Please also see issues 1 and 3 from Claude's review above.

bctiemann:
Addressed the "no fields selected" issue and removed `obj.new_name`. Also this does look better:

Screenshot 2026-05-21 at 5 56 56 AM

I think this is preferable to a multi-select picker too.

jeremystretch (approved):
Nice work!

PR #22197, opened 2026-05-14
Closes #22192: Introduce `HTTP_CLIENT_IP_HEADERS` configuration parameter

### LINKED ISSUE #22192: Support customization of HTTP headers used to infer client IP address
### NetBox version

v4.6.0

### Feature type

New functionality

### Proposed functionality

Introduce a new configuration parameter (e.g. `HTTP_CLIENT_IP_HEADERS`) which lists the HTTP headers that are inspected to determine the client IP address for a request. Currently, these are all hard-coded. The default list should contain the hard-coded headers:

* `HTTP_X_REAL_IP`
* `HTTP_X_FORWARDED_FOR`
* `REMOTE_ADDR`

The configured list will be passed to the [`get_client_ip()`](https://github.com/netbox-community/netbox/blob/df5bc85b482287ab05d375fe528de5d742791f96/netbox/utilities/request.py#L72) utility function.

### Use case

Some reverse proxies inject their own HTTP headers which should take precedence over the defaults. For example, Cloudflare introduces its proprietary [`CF-Connecting-IP`](https://developers.cloudflare.com/fundamentals/reference/http-headers/#cf-connecting-ip) and [`CF-Connectipng-IPv6`](https://developers.cloudflare.com/fundamentals/reference/http-headers/#cf-connecting-ipv6) headers.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #22192

- Introduce the `HTTP_CLIENT_IP_HEADERS` configuration parameter
- Update `get_client_ip()` to reference the configured headers

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -80,6 +80,26 @@ The hostname displayed in the user interface identifying the system on which Net
 
 ---
 
+## HTTP_CLIENT_IP_HEADERS
+
+!!! info "This parameter was introduced in NetBox v4.6.1."
+
+Default:
+
+```python
+(
+    'HTTP_X_REAL_IP',
+    'HTTP_X_FORWARDED_FOR',
+    'REMOTE_ADDR',
+)
+```
+
+An ordered list of HTTP request headers inspected to determine the source IP address of a client request. The first header in the list which is present on the request is used; if none are found, the client IP cannot be determined. This is most commonly required when NetBox is deployed behind a reverse proxy which injects a proprietary client IP header (e.g. `HTTP_CF_CONNECTING_IP` for Cloudflare).
+
+The client IP is used for source-address restrictions on API tokens and for logging failed login attempts.
+
+---
+
 ## HTTP_PROXIES
 
 Default: `None`
--- a/netbox/utilities/request.py
+++ b/netbox/utilities/request.py
@@ -2,6 +2,7 @@ import warnings
 from contextlib import ExitStack, contextmanager
 from urllib.parse import urlparse
 
+from django.conf import settings
 from django.utils.http import url_has_allowed_host_and_scheme
 from django.utils.translation import gettext_lazy as _
 from netaddr import AddrFormatError, IPAddress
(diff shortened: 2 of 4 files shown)


jeremystretch:

jeremystretch:
IMO we don't need a warning about `REMOTE_ADDR`. There are scenarios where it may even be desirable to ignore it.

docs/configuration/system.md
```diff
@@ -80,6 +80,26 @@ The hostname displayed in the user interface identifying the system on which Net
 
 ---
 
+## HTTP_CLIENT_IP_HEADERS
+
+!!! info "This parameter was introduced in NetBox v4.6."
```
pheus:
```suggestion
!!! info "This parameter was introduced in NetBox v4.6.1"
```

netbox/utilities/request.py
```diff
@@ -69,17 +70,14 @@ def copy_safe_request(request, include_files=True):
...
-    HTTP_HEADERS = (
-        'HTTP_X_REAL_IP',
-        'HTTP_X_FORWARDED_FOR',
-        'REMOTE_ADDR',
-        *additional_headers
-    )
-    for header in HTTP_HEADERS:
+    if headers is None:
+        headers = settings.HTTP_CLIENT_IP_HEADERS
+    for header in headers:
```
pheus: This changes `get_client_ip()` from accepting `additional_headers=()` to `headers=None`. Existing plugin/custom code calling `get_client_ip(request, additional_headers=(...))` would now raise `TypeError`, and positional callers such as `get_client_ip(request, ("HTTP_CF_CONNECTING_IP",))` would silently change from “append these after the defaults” to “use only these headers.”

Could we preserve the existing helper contract while still adding the new override behavior? One option would be to keep `additional_headers` and add `headers` as a keyword-only override:

```suggestion
def get_client_ip(request, additional_headers=(), *, headers=None):
    """
    Return the client (source) IP address of the given request. The list of HTTP headers to inspect (in order) may be
    passed explicitly; otherwise the HTTP_CLIENT_IP_HEADERS configuration parameter is used.
    """
    if headers is None:
        headers = (*settings.HTTP_CLIENT_IP_HEADERS, *additional_headers)

    for header in headers:
        ...
```

pheus (changes requested):

jeremystretch:

jeremystretch:
Ok, I think I've addressed the concerns re: backward compatibility.

pheus (approved):
Nice solution!

PR #22201, opened 2026-05-14
Closes #22125: Add test coverage for system housekeeping, data source sync, and script jobs

### LINKED ISSUE #22125: Extend test coverage of background jobs
### Proposed Changes

Extend the current test coverage for background jobs defined in the `core` and `extras` apps. Current coverage is insufficient:

```
Name                                                              Stmts   Miss  Cover
-------------------------------------------------------------------------------------
netbox/core/jobs.py                                                 128     81    37%
netbox/extras/jobs.py                                                74     56    24%
```

We should aim for at least 95% coverage of both modules.

### Justification

Increase overall test coverage to mitigate regressions and breaking changes.

pheus:
### Fixes: #22125

This PR adds focused test coverage for the jobs highlighted in the issue, including `SystemHousekeepingJob`, `SyncDataSourceJob`, and `ScriptJob`.

The new tests cover housekeeping execution and skip paths, data source synchronization behavior, script rollback and exception handling, and request processor integration. It also includes small test helpers for safely instantiating job runners without accumulating log handlers, along with a `DummyScript` test double to keep the script job tests lightweight and focused.

### DIFF
--- a/netbox/core/tests/test_jobs.py
+++ b/netbox/core/tests/test_jobs.py
@@ -1,0 +1,367 @@
+import uuid
+from datetime import timedelta
+from unittest.mock import MagicMock, patch
+
+import requests
+from django.conf import settings
+from django.test import TestCase, override_settings
+from django.utils import timezone
+
+from core.choices import DataSourceStatusChoices
+from core.jobs import SyncDataSourceJob, SystemHousekeepingJob
+from core.models import DataFile, DataSource, Job
+
+
+def _make_runner(cls, **job_attrs):
+    """
+    Build a JobRunner without going through ``__init__``.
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
@claude /review 

I’ve opened #22202 to track the `check_for_new_releases()` empty stable-release case separately, since fixing that would require an implementation change outside the scope of this coverage PR.

For this PR, I’ll address the `SyncDataSourceJob.enqueue()` test gap by making the no-object case assert that no datasource status update is attempted and that an existing datasource remains unchanged.

bctiemann (approved):

PR #22211, opened 2026-05-15
Closes #21870: Optimized Prefix and IP Range availability and utilization calculations

pheus:
### Fixes: #21870

This PR improves IPAM prefix availability and utilization performance by replacing IPSet-heavy calculations with database-backed host queries and interval-based counting.

The main goal is to avoid materializing large address sets in Python when calculating prefix utilization, available IP counts, and first-available IPs. Instead, the implementation uses indexed host lookups, distinct host counts, merged IP range intervals, and a lazy generator for first-available IP discovery.

#### Prefix availability and utilization

- Reworks `Prefix.get_available_ip_count()` to count distinct child IP hosts with range exclusions instead of building large `IPSet` objects.
- Keeps `Prefix.get_utilization()` on a utilization-only path so Prefix ListView does not pay for availability-specific work.
- Adds `Prefix.get_ip_usage_summary()` for Prefix DetailView, where both utilization and available count are needed and can share the same distinct-host scan.
- Centralizes utilization denominator logic so IPv4 pool/non-pool and IPv6 behavior are handled consistently.
- Normalizes usable prefix bounds defensively for string-assigned prefixes.

#### First available IPs

- Replaces IPSet-based first-available logic with a lazy interval sweep.
- Streams distinct occupied child IP hosts in batches.
- Merges populated/utilized ranges into occupied intervals.
- Finds the first gap without walking the whole address space.
- Validates batch size and handles `batch_size=1` correctly.

#### Queryset and lookup improvements

- Adds queryset-level helpers for distinct host counting and available interval generation.
- Adds a `host_between` lookup for inclusive mask-insensitive host-bound filtering.
- Updates child IP/range filtering to use host-bound lookups.
- Adds host expression indexes for `IPRange.start_address` and `IPRange.end_address`.

#### Correctness fixes

- Fixes child IP matching when an IP address has a different mask length from the containing range/prefix bounds.
- Fixes `mark_populated` validation for the same mask-length mismatch case.
- Fixes `BaseIPField.get_prep_value()` handling for zero addresses such as `0.0.0.0` and `::`.
- Makes interval merging address-family aware.
- Normalizes unsorted or overlapping exclude intervals before availability calculations.
- Keeps `rebuild_prefixes()` scoped to the existing VRF/global API while using `apps.get_model()` to avoid circular imports.

#### Deprecation

- Deprecates `IPRange.first_available_ip` in favor of `IPRange.get_first_available_ip()`.

#### Performance

The new implementation avoids materializing large `IPSet` objects and shifts the expensive host work to indexed database queries.

In local testing on a `/16` with approximately 18k child IPs, one utilized range, and one populated range, the optimized paths produced identical results while significantly reducing runtime:

| Operation | Before | After |
|---|---:|---:|
| `Prefix.get_utilization()` | ~1429 ms | ~13.9 ms |
| Available IP count | ~893 ms | ~13.7 ms |
| `get_first_available_ip()` | ~897 ms | ~25.1 ms |
| Detail-view usage summary | n/a | ~13.1 ms, 3 queries |

The previous v4 ListView regression was addressed by splitting the cheap utilization-only path from the detail-view summary path.

#### Testing

The patch includes regression coverage for:

- prefix utilization and available IP count equivalence
- first available IP generation
- lazy available interval batching
- duplicate child IP hosts
- populated/utilized range exclusion
- mixed-family interval merging
- invalid batch sizes
- zero-address preparation
- mask-insensitive child IP/range matching
- `rebuild_prefixes()` legacy VRF/global behavior
- API/view/template paths using the rewritten availability logic

Additional related suites were checked for regressions around API behavior, filtersets, views, tables, ordering, and the `rebuild_prefixes` management command.

### DIFF
--- a/netbox/ipam/fields.py
+++ b/netbox/ipam/fields.py
@@ -42,7 +42,10 @@ class BaseIPField(models.Field):
             raise ValidationError(e)
 
     def get_prep_value(self, value):
-        if not value:
+        # Membership check; `not value` incorrectly treats the valid zero addresses
+        # 0.0.0.0 and :: as empty. netaddr objects compare unequal to all three
+        # sentinels; raw int 0 stays "empty" for backward compatibility.
+        if value in (None, '', 0):
             return None
         if isinstance(value, list):
             return [str(self.to_python(v)) for v in value]
@@ -107,6 +110,7 @@ IPAddressField.register_lookup(lookups.NetContainsOrEquals)
 IPAddressField.register_lookup(lookups.NetHost)
 IPAddressField.register_lookup(lookups.NetIn)
 IPAddressField.register_lookup(lookups.NetHostContained)
+IPAddressField.register_lookup(lookups.NetHostBetween)
 IPAddressField.register_lookup(lookups.NetFamily)
 IPAddressField.register_lookup(lookups.NetMaskLength)
 IPAddressField.register_lookup(lookups.Host)
--- a/netbox/ipam/models/ip.py
+++ b/netbox/ipam/models/ip.py
@@ -15,7 +15,7 @@ from ipam.constants import *
 from ipam.fields import IPAddressField, IPNetworkField
 from ipam.lookups import Host
 from ipam.managers import IPAddressManager
-from ipam.querysets import PrefixQuerySet
+from ipam.querysets import IPRangeQuerySet, PrefixQuerySet
 from ipam.validators import DNSValidator
 from netbox.config import get_config
 from netbox.models import OrganizationalModel, PrimaryModel
@@ -425,14 +425,63 @@ class Prefix(ContactsMixin, GetAvailablePrefixesMixin, CachedScopeMixin, Primary
             return Prefix.objects.filter(prefix__net_contained=str(self.prefix))
         return Prefix.objects.filter(prefix__net_contained=str(self.prefix), vrf=self.vrf)
 
+    @property
+    def usable_ip_bounds(self):
+        """
+        Return the first and last IPs considered usable for available-IP calculations.
+
+        Pools and IPv4 /31-/32 / IPv6 /127-/128 are fully usable; otherwise IPv4 excludes
+        network and broadcast, IPv6 excludes the subnet-router anycast address.
+        """
+        network = netaddr.IPNetwork(self.prefix)
+        family = network.version
+        first = network.first
+        last = network.last
+        mask_length = network.prefixlen
+
+        if (
+            self.is_pool
+            or (family == 4 and mask_length >= 31)
+            or (family == 6 and mask_length >= 127)
+        ):
+            return (
+                netaddr.IPAddress(first, version=family),
+                netaddr.IPAddress(last, version=family),
+            )
+
+        if family == 4:
+            return (
+                netaddr.IPAddress(first + 1, version=family),
+                netaddr.IPAddress(last - 1, version=family),
+            )
+
+        return (
+            netaddr.IPAddress(first + 1, version=family),
+            netaddr.IPAddress(last, version=family),
+        )
+
+    @property
+    def usable_size(self):
+        """
+        The number of usable host addresses within the prefix (excludes reserved addresses).
+        """
+        first_ip, last_ip = self.usable_ip_bounds
+        return int(last_ip) - int(first_ip) + 1
+
     def get_child_ranges(self, **kwargs):
         """
         Return all IPRanges within this Prefix and VRF.
         """
+        # A host BETWEEN over the prefix span uses the ipam_iprange_*_host btree indexes.
+        prefix = netaddr.IPNetwork(self.prefix)
+        bounds = (
+            netaddr.IPAddress(prefix.first, version=prefix.version),
+            netaddr.IPAddress(prefix.last, version=prefix.version),
+        )
         return IPRange.objects.filter(
             vrf=self.vrf,
-            start_address__net_host_contained=str(self.prefix),
-            end_address__net_host_contained=str(self.prefix),
+            start_address__host_between=bounds,
+            end_address__host_between=bounds,
             **kwargs
         )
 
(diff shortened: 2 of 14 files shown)


pheus:

pheus:
@claude /review

I agree on the `IPRange.first_available_ip` template issue. I’d prefer not to restore `@cached_property` for `first_available_ip`, since this value may be used for allocation and should stay fresh. I’ll instead wrap the template access in a `{% with first_available_ip=object.first_available_ip %}` block so it is only evaluated once during rendering.

The index/query expression mismatch is also a good catch. I checked the existing IPAM lookups and NetBox already has `Host` and `Inet` transforms, so rather than adding another transform or using repeated aliases, I think the cleaner fix is to align the new indexes with the existing `__host__inet` query pattern. In other words, define the `IPRange` host indexes using `Inet(Host(...))` so they match the filters using `start_address__host__inet` and `end_address__host__inet`. I’ll also verify this with `EXPLAIN`.

For the IPv6 utilization behavior, I agree with your assessment. It looks like a pre-existing semantic distinction rather than a regression from this change: availability is calculated against usable bounds, while utilization intentionally preserves the existing behavior. I’d keep that out of scope for this performance-focused PR.

The `_to_ipaddress()` point is fair as a micro-optimization. I don’t think it is critical compared to the larger allocation/query improvements, but I can simplify the hot path if it stays readable.

pheus:

pheus:
@claude /review

I agree with points 2, 4, and 5.

For point 2, I’ve added equivalence coverage between `available_ip_count` and `get_available_ips().size` for the important availability cases. That should help guard against future semantic drift while still keeping the optimized path separate from the existing `IPSet` implementation.

For `BaseIPField.get_prep_value()`, the raw integer `0` edge case is fair. I added a focused test and preserve the previous behavior for that unsupported/raw input path, while still allowing valid zero IP addresses such as `0.0.0.0` and `::`.

For the `find_first_available_ip()` fast-path comment, I agree the logic is correct but the wording can be clearer. I've adjusted the comment to make it explicit that this only covers the case where one merged occupied interval spans the full usable range.

For point 1, I think the current interval-exclusion approach is still the right trade-off for this PR. The intervals are merged before building the predicate, and the expected hot path is many IPs with comparatively few populated/utilized ranges. A very high number of disjoint ranges could justify a different strategy later, but I’d prefer to keep this implementation readable unless benchmarks show that case is a real bottleneck.

For point 3, the template already avoids calling `get_first_available_ip()` when `available_count` is zero. When there is available space, the count and first-available calculations use different query shapes, and I’d rather not introduce a combined/cached summary object in this PR. That can be revisited separately if benchmarks show the remaining duplicate work is still significant.

netbox/ipam/fields.py
```diff
@@ -42,7 +42,10 @@ def to_python(self, value):
             raise ValidationError(e)
 
     def get_prep_value(self, value):
-        if not value:
+        # Use an explicit None / empty-string check; `not value` incorrectly treats
+        # the valid zero addresses 0.0.0.0 and :: as empty. Raw int 0 is preserved
+        # as "empty" for backward compatibility (Django's ORM does not pass it here).
+        if value is None or value == '' or (type(value) is int and value == 0):
```
jeremystretch: This seems equivalent but simpler:
```suggestion
        if value in (None, '', 0):
```

netbox/ipam/models/ip.py
```diff
@@ -479,14 +489,49 @@ def get_available_ips(self):
 
         return available_ips
 
+    @property
```
jeremystretch: Given that accessing this property triggers potentially expensive database queries (`get_child_ranges()`, `get_child_ips()`), we should probably implement it as a regular method, e.g. `get_available_ip_count()`. (Alternatively, we could use `@cached_property` but then we need to worry about stale value.)

netbox/ipam/models/ip.py
```diff
@@ -479,14 +489,49 @@ def get_available_ips(self):
...
+    def available_ip_count(self):
+        """
+        Return the number of available IPs without constructing the full IPSet.
+
+        Intended for summary display (e.g. the Prefix detail Addressing panel).
+        get_available_ips() remains available for callers that need the actual set.
+        """
+        first_ip, last_ip = get_usable_ip_bounds(self)
+        usable_size = int(last_ip) - int(first_ip) + 1
```
jeremystretch, line 527: Might be worth exposing this as a cached property on Prefix, but we also want to be careful to avoid confusion with `self.prefix.size`.

netbox/ipam/utils.py
```diff
@@ -270,3 +316,185 @@ def get_next_available_prefix(ipset, prefix_size):
...
+    merged.append((current_start, current_end))
+    return merged
+
+
+def count_ip_intervals(intervals):
+    """Count addresses covered by the supplied intervals. Callers should pass merged intervals."""
+    return sum(end - start + 1 for start, end in intervals)
+
+
+def count_distinct_ip_hosts_outside_intervals(queryset, intervals, version):
```
jeremystretch: Can this be collapsed into `count_distinct_ip_hosts()` above by making `intervals` and `version` optional?

netbox/ipam/utils.py
```diff
@@ -270,3 +316,185 @@ def get_next_available_prefix(ipset, prefix_size):
...
+    # netaddr raises AddrFormatError for malformed input and ValueError for
+    # otherwise-valid CIDR strings; both fall through to the IPNetwork path.
+    text = str(value)
+    try:
+        return netaddr.IPAddress(text)
+    except (netaddr.AddrFormatError, ValueError):
+        return netaddr.IPNetwork(text).ip
+
+
+def annotate_host_address(queryset):
```
jeremystretch: It might make sense to introduce a custom subclass of QuerySet for the IPAddress model to hold some of these functions, so that they can be executed on a queryset directly.

netbox/ipam/utils.py
```diff
@@ -270,3 +316,185 @@ def get_next_available_prefix(ipset, prefix_size):
...
+    """Annotate an IPAddress queryset with its host address (mask ignored)."""
+    return queryset.order_by().annotate(
+        host_address=Cast(
+            Host('address'),
+            output_field=IPAddressField(),
+        )
+    )
+
+
+def filter_ip_hosts_between(queryset, first_ip, last_ip):
```
jeremystretch: Did you happen to explore implementing this as a lookup? It would be nice to do e.g. `.filter(address__host_between=[x, y])` if feasible.

netbox/ipam/utils.py
```diff
@@ -33,27 +47,64 @@ def title(self):
...
     :param show_available: Include available prefixes.
     :param show_assigned: Show assigned prefixes.
     """
     child_prefixes = []
 
     # Add available prefixes to the table if requested
     if prefix_list and show_available:
+        # Infer the Prefix model from the queryset/list so this helper does not need
+        # a model import at module scope.
+        prefix_model = getattr(prefix_list, 'model', None) or prefix_list[0].__class__
```
jeremystretch: Could we use `apps.get_model()` here instead?

netbox/ipam/models/ip.py
```diff
@@ -731,30 +792,44 @@ def get_available_ips(self):
...
+    def available_ip_count(self):
+        """
+        Return the number of available IPs without constructing the full IPSet.
+        """
+        if self.mark_populated:
+            return 0
+
+        return max(self.size - count_distinct_ip_hosts(self.get_child_ips()), 0)
+
+    @property
```
jeremystretch: We probably want to keep this as a method, both to avoid a breaking change and because it effects a database query.

netbox/ipam/models/ip.py
```diff
@@ -731,30 +792,44 @@ def get_available_ips(self):
 
         return netaddr.IPSet(range) - child_ips
 
-    @cached_property
+    @property
+    def available_ip_count(self):
```
jeremystretch: Same concern here as with `available_ip_count()` on IPRange. Just calling out for consistency.

jeremystretch (changes requested):
I've done an initial review, but I'm finding it difficult to keep straight the hierarchy of all the new utility functions. It seems like a lot of them could be moved to model or queryset methods to keep things better organized. What do you think?

pheus:
Thanks for the review! I haven't thought about the lookup solution, but that's way better!

I’ve pushed a new version with the requested cleanup and a few additional safeguards.

The main changes since the earlier revision are:

- Split the Prefix ListView and DetailView performance paths:
  - `get_utilization()` is utilization-specific again and should avoid the extra availability-side work.
  - `get_ip_usage_summary()` is used for the detail-view case where both utilization and available IP count are needed.
- Added/kept queryset-level helpers for the IPAM availability logic instead of leaving the logic in standalone utility functions.
- Added the `host_between` lookup for mask-insensitive inclusive host-bound filtering.
- Updated child IP/range filtering to use host-bound lookups consistently.
- Added host expression indexes for `IPRange.start_address` and `IPRange.end_address`.
- Normalized and validated `NetHostBetween` bounds.
- Made interval merging address-family aware.
- Made `available_intervals()` normalize unsorted/overlapping exclude intervals internally.
- Added validation for invalid `batch_size` values.
- Fixed the batch resume logic so `batch_size=1` is safe.
- Normalized usable prefix bounds defensively.
- Kept `rebuild_prefixes()` scoped to the existing `None` / VRF-pk API while still using `apps.get_model()` to avoid the circular import.
- Deprecated `IPRange.first_available_ip` in favor of `get_first_available_ip()`.

The v4.7 deprecation target for `IPRange.first_available_ip` is intentional.

Performance-wise, the current approach keeps the interval/generator design: SQL handles indexed distinct-host scans and Python handles the merged interval sweep. This avoids the previous IPSet materialization cost while keeping the implementation easier to review and maintain than a large SQL/window-function rewrite.

In local testing on a `/16` with around 18k child IPs, the optimized paths produced identical results with much lower runtimes:

| Operation | Before | After |
|---|---:|---:|
| `Prefix.get_utilization()` | ~1429 ms | ~13.9 ms |
| Available IP count | ~893 ms | ~13.7 ms |
| `get_first_available_ip()` | ~897 ms | ~25.1 ms |

I also added regression coverage around the availability/counting behavior, mask-insensitive child matching, zero-address handling, family-safe interval merging, lazy batching, invalid batch sizes, and `rebuild_prefixes()` behavior.

netbox/ipam/utils.py
```diff
@@ -33,27 +34,63 @@ def title(self):
...
+    network and broadcast, IPv6 excludes the subnet-router anycast address.
+    """
+    network = netaddr.IPNetwork(prefix.prefix)
+    family = network.version
+    first = network.first
+    last = network.last
+    mask_length = network.prefixlen
+
+    if (
+        prefix.is_pool
```
jeremystretch: Given that this utility relies on the `is_pool` field, maybe it should be a property of the Prefix model?

netbox/ipam/utils.py
```diff
@@ -204,6 +226,9 @@ def rebuild_prefixes(vrf):
     """
     Rebuild the prefix hierarchy for all prefixes in the specified VRF (or global table).
     """
+    prefix_model = apps.get_model('ipam', 'Prefix')
```
jeremystretch: IMO giving the model its proper name makes for more readable usage. Up to you.
```suggestion
    Prefix = apps.get_model('ipam', 'Prefix')
```

netbox/ipam/models/ip.py
```diff
@@ -714,48 +820,96 @@ def get_child_ips(self):
...
+        # TODO: Remove in NetBox v4.7
+        warnings.warn(
+            'IPRange.first_available_ip is deprecated and will be removed in NetBox v4.7. Use '
+            'get_first_available_ip() instead.',
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.get_first_available_ip()
+
+    @property
```
jeremystretch: Any reason not keep this a `cached_property`?

jeremystretch (dismissed):
Had some minor suggestions but I'm okay merging this as-is. Awesome work @pheus!

netbox/ipam/models/ip.py
pheus: Good question. The deprecation was mainly about API clarity rather than a runtime issue in this PR.

`first_available_ip` does database-backed availability work and depends on related `IPAddress` / `IPRange` rows, so exposing it as an attribute can make it look like cheap model state. A method like `get_first_available_ip()` makes that cost and behavior more explicit. The `cached_property` also has the usual freshness caveat: once accessed, the value is cached for the lifetime of that model instance, even if related IPs or ranges change later in the same request/script.

But I agree that deprecating an existing public attribute should be handled more deliberately. I’ll remove the warning/TODO from this PR and keep `first_available_ip` as a `cached_property` compatibility wrapper for now. If we still want to move callers to `get_first_available_ip()`, we can track that separately with a proper deprecation issue/release note.

netbox/ipam/utils.py
pheus: That makes sense. Thanks!

pheus:
Thank you very much for the thoughtful review and guidance! I’ve adopted your suggestions and squashed the commits.

`Prefix` now exposes `usable_ip_bounds` as a property, `prefix_model` has been renamed to `Prefix`, and I’ve removed the deprecation from this PR so it can be tracked separately if we decide to pursue it.

Since this PR changes quite a few `Prefix` and `IPRange` methods, I also took the opportunity to reorder the methods to better follow the common model layout used across NetBox: properties are grouped after `clean()` / `save()`, public `get_*` behavior is grouped by concern, and private helpers are kept close to the methods that use them. I hope that makes the larger diff a bit easier to review and maintain.

jeremystretch (approved):

PR #22213, opened 2026-05-15
Fixes #22206: Fix TypeError when validating Table Configurations with unset ordering

### LINKED ISSUE #22206: TableConfig.clean() raises TypeError when `ordering` is None
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

1. Open a NetBox shell:

   ```bash
   python manage.py nbshell
   ```
2. Create a `TableConfig` without setting the `ordering` field:

   ```python
   
   from dcim.models import Site
   from extras.models import TableConfig
   from utilities.tables import get_table_for_model
   
   tc = TableConfig(
       object_type=ContentType.objects.get_for_model(Site),
       table=get_table_for_model(Site).__name__,
       name='Example',
       columns=['name', 'status'],
       # ordering left unset; the field allows null/blank
   )
   tc.full_clean()
   ```
3. The same error can surface in normal create/update paths that run `full_clean()` when no ordering is supplied.

### Expected Behavior

`full_clean()` accepts an unset `ordering` field and treats it as "no ordering configured", matching the model definition where `ordering` allows `null=True` and `blank=True`.

### Observed Behavior

`TableConfig.clean()` raises:

```text
TypeError: 'NoneType' object is not iterable
```

The error is raised while validating ordering columns:

```python
# Validate ordering columns
for name in self.ordering:
    ...
```

The nearby `ordering_items` property already handles the same field defensively:

```python
return [
    (name, get_col(name.lstrip('-')))
    for name in self.ordering or []
]
```

A similar guard in `clean()`, such as `for name in self.ordering or []:`, should preserve current behavior for callers that do provide ordering while allowing the field to remain unset.

This was surfaced while adding standard test coverage for `extras.tableconfig` in netbox-community/netbox#22088. The housekeeping PR currently exempts the affected API/view coverage with a self-cleaning marker that will prompt removal once this issue is fixed.

pheus:
### Fixes: #22206

This PR prevents `TableConfig.clean()` from raising a `TypeError` when `ordering` is unset.

It adds an explicit null-safe check so missing ordering is treated as “no ordering configured,” while preserving the existing validation behavior when ordering values are provided. It also includes a regression test to ensure `full_clean()` succeeds when the ordering field is left unset.

### DIFF
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -658,7 +658,7 @@ class TableConfig(CloningMixin, ChangeLoggedModel):
         table = self.table_class([])
 
         # Validate ordering columns
-        for name in self.ordering:
+        for name in self.ordering or []:
             if name.startswith('-'):
                 name = name[1:]  # Strip leading hyphen
             if name not in table.columns:
--- a/netbox/extras/tests/test_models.py
+++ b/netbox/extras/tests/test_models.py
@@ -24,6 +24,7 @@ from extras.models import (
     EventRule,
     ExportTemplate,
     ImageAttachment,
+    TableConfig,
     Tag,
     TaggedItem,
 )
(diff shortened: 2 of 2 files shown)


pheus:

pheus:

jeremystretch (approved):
Thanks @pheus!

PR #22214, opened 2026-05-15
Fixes #22207: Fix GraphQL `object_type` field for table configurations

### LINKED ISSUE #22207: TableConfigType GraphQL type missing explicit `object_type` annotation
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

1. Create a `TableConfig` instance bound to any content type.
2. Issue a GraphQL query that selects the related `object_type`:

   ```graphql
   {
     table_config(id: 1) {
       id
       name
       table
       object_type {
         id
       }
     }
   }
   ```

### Expected Behavior

The query returns the `object_type` field as a `ContentTypeType`, consistent with peer GraphQL types that explicitly annotate content-type fields.

For example, peer types such as `ImageAttachmentType` and `JournalEntryType` explicitly annotate their content-type fields.

### Observed Behavior

The GraphQL query fails with:

```text
Cannot query field 'id' on type 'DjangoModelType'.
```

`TableConfigType` uses `fields='__all__'` but does not annotate `object_type` explicitly. As a result, the auto-generated schema falls back to the generic `DjangoModelType`, which does not expose `id`.

A fix would be to use the same annotation pattern already used by peer types:

```python
@strawberry_django.type(
    models.TableConfig,
    fields='__all__',
    filters=TableConfigFilter,
    pagination=True
)
class TableConfigType(ObjectType):
    object_type: Annotated["ContentTypeType", strawberry.lazy('netbox.graphql.types')] | None
    user: Annotated["UserType", strawberry.lazy('users.graphql.types')] | None
```

This was surfaced while adding standard test coverage for `extras.tableconfig` in netbox-community/netbox#22088. The auto-built GraphQL query in `APIViewTestCases.GraphQLTestCase` is one of the blockers behind the API exemption documented there.

pheus:
### Fixes: #22207

This explicitly annotates the `object_type` field on `TableConfigType` as `ContentTypeType`.

Without this annotation, the GraphQL field falls back to `DjangoModelType`, which does not expose `id` and causes nested queries such as `object_type { id }` to fail. This mirrors the existing `ImageAttachmentType` pattern for content type fields.

### DIFF
--- a/netbox/extras/graphql/types.py
+++ b/netbox/extras/graphql/types.py
@@ -204,6 +204,7 @@ class SubscriptionType(ObjectType):
     pagination=True
 )
 class TableConfigType(ObjectType):
+    object_type: Annotated["ContentTypeType", strawberry.lazy('netbox.graphql.types')] | None
     user: Annotated["UserType", strawberry.lazy('users.graphql.types')] | None
 
 
--- a/netbox/netbox/tests/test_graphql.py
+++ b/netbox/netbox/tests/test_graphql.py
@@ -1,6 +1,7 @@
 import json
 
 import strawberry
+from django.contrib.contenttypes.models import ContentType
 from django.test import override_settings
 from django.urls import reverse
 from rest_framework import status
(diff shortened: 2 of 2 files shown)


pheus:

jnovinger (approved):

PR #22215, opened 2026-05-15
Fixes #22208: Add User ID and Username Filters for Jobs

### LINKED ISSUE #22208: JobFilterSet missing `user_id` FK filter
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

1. Inspect `JobFilterSet` in `netbox/core/filtersets.py`.
2. Confirm that `user` is included via `Meta.fields`, but there is no explicit `user_id` companion filter.
3. Attempt to filter jobs by user ID through the REST API or filterset:

   ```http
   GET /api/core/jobs/?user_id=1
   ```

### Expected Behavior

`JobFilterSet` exposes an ID-based filter for the `user` FK, consistent with the standard NetBox filterset convention enforced by `BaseFilterSetTests`.

A peer example is `ObjectChangeFilterSet`, which defines both:

* `user_id` for filtering by user ID
* `user` for filtering by username

### Observed Behavior

`JobFilterSet` does not expose `user_id`, so standard filterset coverage fails with a missing FK companion filter:

```text
No filter defined for user_id (user)!
```

A fix would be to define `user_id` explicitly and, if desired, override `user` to filter by username, matching the pattern used elsewhere:

```python
user_id = django_filters.ModelMultipleChoiceFilter(
    queryset=User.objects.all(),
    distinct=False,
    label=_('User (ID)'),
)
user = django_filters.ModelMultipleChoiceFilter(
    field_name='user__username',
    queryset=User.objects.all(),
    distinct=False,
    to_field_name='username',
    label=_('User (name)'),
)
```

This was surfaced while adding standard filterset test coverage for `core.job` in netbox-community/netbox#22088. The housekeeping PR currently exempts the affected filterset coverage with a self-cleaning marker that will prompt removal once this issue is fixed.

pheus:
### Fixes: #22208

Adds the missing `user_id` companion filter to `JobFilterSet`, allowing jobs to be filtered by the related user’s ID.

Also defines an explicit `user` filter for username-based filtering, matching the pattern used by `ObjectChangeFilterSet` and keeping `JobFilterSet` consistent with NetBox’s standard FK companion filter behavior.

### DIFF
--- a/netbox/core/filtersets.py
+++ b/netbox/core/filtersets.py
@@ -137,7 +137,24 @@ class JobFilterSet(BaseFilterSet):
         distinct=False,
         null_value=None
     )
+    notifications = django_filters.MultipleChoiceFilter(
+        choices=JobNotificationChoices,
+        distinct=False,
+        null_value=None
+    )
     queue_name = django_filters.CharFilter()
+    user_id = django_filters.ModelMultipleChoiceFilter(
+        queryset=User.objects.all(),
+        distinct=False,
+        label=_('User (ID)'),
+    )
+    user = django_filters.ModelMultipleChoiceFilter(
+        field_name='user__username',
+        queryset=User.objects.all(),
+        distinct=False,
+        to_field_name='username',
+        label=_('User name'),
+    )
 
     class Meta:
         model = Job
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
@claude /review

Reviewer concerns addressed:
1. JobTestCase(TestCase, BaseFilterSetTests) — picks up test_id and test_missing_filters
2. Label _('User (name)') → _('User name') (aligns with ObjectChangeFilterSet)

bctiemann (approved):

PR #22220, opened 2026-05-15
Fixes #22202: Handle empty release list in check_for_new_releases

### LINKED ISSUE #22202: SystemHousekeepingJob.check_for_new_releases() crashes when no stable releases are returned
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

1. Configure NetBox with `ISOLATED_DEPLOYMENT = False` and a valid `RELEASE_CHECK_URL`.
2. Cause the configured release-check endpoint to return a JSON response in which every entry has `prerelease=True`, `devrelease=True`, or lacks `tag_name` (for example, by configuring a fork's GitHub releases URL where no stable releases have been published yet).
3. Trigger the daily `SystemHousekeepingJob` (or call `SystemHousekeepingJob.check_for_new_releases()` directly via `nbshell`).

### Expected Behavior

The job should gracefully handle the absence of stable releases, log an info-level message that no usable releases were found and return without raising.

### Observed Behavior

core/jobs.py:225 calls `max(releases)` on an unconditionally-populated list. When the filter loop (`for release in response.json(): ...`) excludes every entry, releases is empty and `max([])` raises:

`ValueError: max() arg is an empty sequence`

The exception escapes `check_for_new_releases()` and propagates into the surrounding `SystemHousekeepingJob.run()`, which the `JobRunner.handle()` framework then surfaces as a job error. The cache is never updated.

Suggested Fix

Add an early-return guard before `max()`:

```
if not releases:
    self.logger.info("No stable releases found; skipping")
    return
latest_release = max(releases)
```

A test for this case belongs in `CheckForNewReleasesTestCase` in `core/tests/test_jobs.py`.

pheus:
### Fixes: #22202

Guard max(releases) against an empty iterable to prevent a ValueError when the release-check endpoint returns only prereleases, dev releases, or entries lacking a tag_name.

### DIFF
--- a/netbox/core/jobs.py
+++ b/netbox/core/jobs.py
@@ -222,8 +222,11 @@ class SystemHousekeepingJob(JobRunner):
             if 'tag_name' not in release or release.get('devrelease') or release.get('prerelease'):
                 continue
             releases.append((version.parse(release['tag_name']), release.get('html_url')))
-        latest_release = max(releases)
         self.logger.debug(f"Found {len(response.json())} releases; {len(releases)} usable")
+        if not releases:
+            self.logger.info("No usable releases found; skipping")
+            return
+        latest_release = max(releases)
         self.logger.info(f"Latest release: {latest_release[0]}")
 
         # Cache the most recent release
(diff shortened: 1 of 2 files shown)


pheus:

jeremystretch (approved):
Thanks @pheus!

PR #22223, opened 2026-05-18
Closes #22110: Add validation for model table test classes

### LINKED ISSUE #22110: Add coverage check for model table test cases
### Proposed Changes

Add a coverage test to ensure each canonical model-backed table has a corresponding table test case.

The new check should follow the existing convention-based approach used in `test_base_classes.py`: for each model with a standard `{ModelName}Table`, verify that `{ModelName}TableTest` exists in the app’s `tests.test_tables` module, inherits from `TableTestCases.StandardTableTestCase`, and points at the expected table class.

Non-standard or intentionally exempt tables can be handled with explicit exemptions where needed.

### Justification

We now have a reusable table smoke-test helper, but there is not yet a safety net to ensure new or existing tables actually use it.

Adding a coverage check will help prevent table test coverage from drifting over time, similar to how NetBox already validates other model-related companion classes. This should make future table changes safer, reduce the chance of missing unsupported sortable columns, and give contributors a clear pattern to follow when adding new model-backed tables.

pheus:
### Closes: #22110

Add test ensuring each model-backed table has a corresponding test case inheriting from StandardTableTestCase with correct table attribute set. Includes helper method to import table test classes by model.

### DIFF
--- a/netbox/dcim/tests/test_tables.py
+++ b/netbox/dcim/tests/test_tables.py
@@ -31,6 +31,10 @@ class RackRoleTableTestCase(TableTestCases.StandardTableTestCase):
     table = RackRoleTable
 
 
+class RackGroupTableTestCase(TableTestCases.StandardTableTestCase):
+    table = RackGroupTable
+
+
 class RackTypeTableTestCase(TableTestCases.StandardTableTestCase):
     table = RackTypeTable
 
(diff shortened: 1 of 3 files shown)


pheus:

pheus:
@claude /review Nice catch. I've fixed it.

arthanson (approved):

PR #22226, opened 2026-05-18
Closes #22091: Replace `override_settings` with explicit permission grants

### LINKED ISSUE #22091: Audit usage of `@override_settings` throughout test suite
### Proposed Changes

Audit the entire test suite for usage of `@override_settings`. Adapt as appropriate by removing the decorator or updating the testing configuration file.

### Justification

There are likely many instances where the use of `@override_settings` can be avoided by modifying the test configuration. This will simplify the test suite overall.

pheus:
### Closes: #22091

Audits and reduces the use of `@override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])` (sometimes paired with `EXEMPT_EXCLUDE_MODELS=[]`) across the test suite.

#### Background

The audit found 213 occurrences of `@override_settings` in the test suite. Of these, 117 wildcard-exempted view permissions purely as a setup convenience. They applied to tests that exercise a specific URL or form, where the cleaner alternative is to grant the test user the exact permissions the view actually needs.

#### Changes

In nine per-app test files, 81 occurrences of `@override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])` were replaced with explicit `self.add_permissions('<app>.view_<model>', ...)` calls listing the specific models each view reads. The `from django.test import override_settings` import was dropped where it became unused.

Three replacement shapes appear, depending on what the test does:
- **Detail-tab / child-listing views**: replaced with `view_<parent>` plus `view_<child>` for whatever models the page renders.
- **Tests that already grant `add`/`change`/`delete` via an `ObjectPermission`**: the override existed only so the form could render its FK dropdowns; added the matching `view_*` permissions for those FK targets.
- **Bulk-import tests** (the ones that used `EXEMPT_EXCLUDE_MODELS=[]`): granted explicit `view_*` for every FK referenced by name in the import payload, including the normally-excluded models (`users.*`, `extras.configrevision`) where applicable.

After the change, every removed override is replaced with permissions that match exactly what the view under test reads. Tests now express their permission dependencies instead of bypassing them globally.

Result: `EXEMPT_VIEW_PERMISSIONS=['*']` overrides drop from 117 → 27 in the test suite. The full suite (9462 tests) passes.

#### Not changed (intentionally)

- **`netbox/utilities/testing/views.py` and `netbox/utilities/testing/api.py`**: the generic `ViewTestCases.*` / `APIViewTestCases.*` base classes inherited by most app test cases also use `@override_settings(EXEMPT_VIEW_PERMISSIONS=['*'],
EXEMPT_EXCLUDE_MODELS=[])` on their `*_with_permission` / changelog / list-anonymous tests. These need view access to an unbounded set of FK-related models (the form-render dependencies vary per subclass), so replacing them safely would require either every subclass to declare its FK view dependencies, or a broader refactor of the base helpers. Out of scope for this PR.
- **`netbox/netbox/tests/test_authentication.py`**: this module's purpose is to test the authentication / permission system itself, including the `LOGIN_REQUIRED` and `EXEMPT_VIEW_PERMISSIONS` settings as features. The overrides here are part of the test, not setup.
- **Non-permission `@override_settings` (`CUSTOM_VALIDATORS`, `LOGIN_REQUIRED`, `ISOLATED_DEPLOYMENT`, `GRAPHQL_*`, `CACHES`, `DEBUG`, `STREAMING_EXPORTS`, etc.)**: these tests exist specifically to verify the behavior of the setting being overridden. Removing them would change what the test is testing.

### DIFF
--- a/netbox/circuits/tests/test_views.py
+++ b/netbox/circuits/tests/test_views.py
@@ -1,7 +1,6 @@
 import datetime
 
 from django.contrib.contenttypes.models import ContentType
-from django.test import override_settings
 from django.urls import reverse
 
 from circuits.choices import *
@@ -210,8 +209,13 @@ class CircuitTestCase(ViewTestCases.PrimaryObjectViewTestCase):
         self.assertContains(response, circuit_type.name)
         self.assertContains(response, 'background-color: #12ab34')
 
-    @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'], EXEMPT_EXCLUDE_MODELS=[])
     def test_bulk_import_objects_with_terminations(self):
+        self.add_permissions(
+            'circuits.view_circuit',
+            'circuits.view_provider',
+            'circuits.view_circuittype',
+            'dcim.view_site',
+        )
         site = Site.objects.first()
         json_data = f"""
             [
@@ -420,8 +424,13 @@ class CircuitTerminationTestCase(ViewTestCases.PrimaryObjectViewTestCase):
             'description': 'New description',
         }
 
-    @override_settings(EXEMPT_VIEW_PERMISSIONS=['*'])
     def test_trace(self):
+        self.add_permissions(
+            'circuits.view_circuittermination',
+            'dcim.view_cable',
+            'dcim.view_interface',
+            'dcim.view_device',
+        )
         device = create_test_device('Device 1')
 
         circuittermination = CircuitTermination.objects.first()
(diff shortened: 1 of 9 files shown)


pheus:

pheus:
@claude /review I've address your concern.

jeremystretch (approved):
Nice work @pheus!

PR #22230, opened 2026-05-18
Fixes #22227: Limit NAT (outside) list to 10 with link to filtered view

### LINKED ISSUE #22227: Detailed IP-address view doesn't seem right when an IP has multiple NAT assignments
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

When multiple public IPs are NAT‑ed to the same inside IP (example: 192.168.0.1/32), the IP detail view renders all NAT entries in one continuous inline row in the NAT (inside) field rather than placing each entry on its own line. It appears to be a CSS/layout issue.

Steps to reproduce:
1. Create inside IP: 192.168.0.1/32.
2. Create external IP 1: 1.1.1.1/32 and configure NAT: map 1.1.1.1/32 → inside 192.168.0.1/32. Assign 1.1.1.1 to any interface.
3. Create external IP 2: 2.2.2.2/32 and configure NAT: map 2.2.2.2/32 → inside 192.168.0.1/32. Assign 2.2.2.2 to any interface.
4. Create external IP 3: 3.3.3.3/32 and configure NAT: map 3.3.3.3/32 → inside 192.168.0.1/32. Assign 3.3.3.3 to any interface.
5. Open detailed view for inside IP 192.168.0.1/32.

### Expected Behavior

Image

### Observed Behavior

Image

pheus:
### Closes: #22227

Render NAT (outside) entries as a vertical list capped at 10 items on the IP detail view; append a muted "…" link routing to the IP address list filtered by nat_inside_id when more exist. Expose nat_inside_id in IPAddressFilterForm so the applied filter is visible in the sidebar of the resulting list view, and add a filterset test for the existing nat_inside_id declaration.

### DIFF
--- a/netbox/ipam/forms/filtersets.py
+++ b/netbox/ipam/forms/filtersets.py
@@ -352,6 +352,7 @@ class IPAddressFilterForm(ContactModelFilterForm, TenancyFilterForm, PrimaryMode
         ),
         FieldSet('vrf_id', 'present_in_vrf_id', name=_('VRF')),
         FieldSet('device_id', 'virtual_machine_id', name=_('Device/VM')),
+        FieldSet('nat_inside_id', name=_('NAT')),
         FieldSet('tenant_group_id', 'tenant_id', name=_('Tenant')),
         FieldSet('owner_group_id', 'owner_id', name=_('Ownership')),
         FieldSet('contact', 'contact_role', 'contact_group', name=_('Contacts')),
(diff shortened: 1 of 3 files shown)


pheus:

pheus:
- Point 1: incorrect: verified empirically with CaptureQueriesContext; {% if ips %} populates _result_cache, so ips|slice:":10" returns a list slice, not a new query.
- Point 2: low-priority style preference; current code matches the file's existing test convention.

jnovinger (approved):

PR #22234, opened 2026-05-18
Closes #21952: Improve robustness of RQ worker check

### LINKED ISSUE #21952: Improve robustness of RQ worker check

jeremystretch:
### Closes: #21952

- Introduce NetBoxRQWorker, a Worker subclass which overrides `heartbeat()` to re-register the worker if needed to recover from a Redis crash
- Alter `get_workers_for_queue()` to include only workers with a valid heartbeat

### DIFF
--- a/netbox/netbox/api/views.py
+++ b/netbox/netbox/api/views.py
@@ -2,19 +2,18 @@ import platform
 
 from django import __version__ as DJANGO_VERSION
 from django.conf import settings
-from django_rq.queues import get_connection
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
 from rest_framework.views import APIView
-from rq.worker import Worker
 
 from netbox.api.authentication import IsAuthenticatedOrLoginNotRequired
 from netbox.plugins.utils import get_installed_plugins
 from users.api.serializers import UserSerializer
 from utilities.apps import get_installed_apps
+from utilities.rqworker import get_all_workers
 
 
 class APIRootView(APIView):
@@ -62,7 +61,7 @@ class StatusView(APIView):
             'netbox-full-version': settings.RELEASE.full_version,
             'plugins': get_installed_plugins(),
             'python-version': platform.python_version(),
-            'rq-workers-running': Worker.count(get_connection('default')),
+            'rq-workers-running': len(get_all_workers()),
         })
 
 
--- a/netbox/utilities/rqworker.py
+++ b/netbox/utilities/rqworker.py
@@ -1,14 +1,50 @@
+import logging
+
 from django_rq.queues import get_connection
 from rq import Retry, Worker
+from rq.worker_registration import REDIS_WORKER_KEYS
+from rq.worker_registration import register as register_worker
 
 from netbox.config import get_config
 from netbox.constants import RQ_QUEUE_DEFAULT
 
 __all__ = (
+    'NetBoxRQWorker',
+    'any_workers_for_queue',
+    'get_all_workers',
     'get_queue_for_model',
     'get_rq_retry',
     'get_workers_for_queue',
 )
+
+logger = logging.getLogger('netbox.rqworker')
+
+
+class NetBoxRQWorker(Worker):
+    """
+    RQ worker subclass which self-heals its registration. If the worker's
+    registration is missing from Redis (e.g. because the tasks Redis database
+    was lost and rebuilt while the worker was running), the next heartbeat
+    will re-register the worker so that Worker.all() / Worker.find_by_key()
+    can locate it again.
+    """
+
+    def heartbeat(self, *args, **kwargs):
+        try:
+            if not self.connection.sismember(REDIS_WORKER_KEYS, self.key):
+                logger.warning(f"Worker {self.name} not found in registry; re-registering.")
+                # If the worker hash still exists (partial Redis data loss),
+                # register_birth() would raise because rq treats an existing,
+                # non-dead hash as an active worker. Re-add to the registry
+                # sets directly in that case; the heartbeat below will refresh
+                # the hash TTL.
+                if self.connection.exists(self.key) and not self.connection.hexists(self.key, 'death'):
+                    register_worker(self, self.connection)
+                else:
+                    self.register_birth()
+        except Exception:
+            logger.exception("Failed to verify worker registration.")
+        super().heartbeat(*args, **kwargs)
 
 
 def get_queue_for_model(model):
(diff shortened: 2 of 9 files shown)


jeremystretch:

jeremystretch:

netbox/utilities/rqworker.py
```diff
@@ -18,11 +59,53 @@ def get_queue_for_model(model):
...
+    """
+    Return True if the given Worker is currently servicing queue_name and its
+    last heartbeat is within (worker_ttl + 60s) of threshold.
+    """
+    if queue_name not in worker.queue_names():
+        return False
+    last = worker.last_heartbeat
+    if last is None:
+        return False
+    ttl = (worker.worker_ttl or DEFAULT_WORKER_TTL) + 60
```
pheus: **question / possible blocker:** I think this liveness check may be wrong for workers started with a non-default `worker_ttl`.

`Worker.all()` / `Worker.find_by_key()` reconstruct worker objects from Redis, but RQ does not appear to persist the original `worker_ttl` in the worker hash. That means this line will usually evaluate against the constructor/default TTL, not necessarily the TTL of the running worker:

```python
ttl = (worker.worker_ttl or DEFAULT_WORKER_TTL) + 60
```

For example, an idle worker with a longer custom TTL may legitimately heartbeat less often than the default threshold, but NetBox could report it as not running. Could we add coverage for a custom `worker_ttl`, and either persist the TTL from `NetBoxRQWorker` or base the freshness check on the Redis key TTL/fallback threshold instead?

netbox/netbox/api/views.py
```diff
@@ -62,7 +61,7 @@ def get(self, request):
             'netbox-full-version': settings.RELEASE.full_version,
             'plugins': get_installed_plugins(),
             'python-version': platform.python_version(),
-            'rq-workers-running': Worker.count(get_connection('default')),
+            'rq-workers-running': len(get_workers_for_queue('default')),
```
pheus: **question:** Is it intentional that the dashboard/API worker counts now only count workers for the `default` queue?

The previous `Worker.count(get_connection('default'))` counted workers registered on the tasks Redis connection, while `len(get_workers_for_queue('default'))` now filters to workers listening on the `default` queue. That seems right for script execution, but the dashboard/API root fields read more like overall RQ worker status.

Could we either preserve the previous queue-agnostic count for the stats/API paths, or add a note/test showing that these fields are intended to mean “workers servicing the default queue” rather than “all RQ workers”?

netbox/utilities/tests/test_rqworker.py
```diff
@@ -0,0 +1,237 @@
...
+        super_heartbeat.assert_called_once()
+
+    def test_heartbeat_tolerates_redis_exception(self):
+        worker = self._make_subject(is_member=False)
+        worker.connection.sismember.side_effect = RuntimeError('redis down')
+        with patch('rq.Worker.heartbeat') as super_heartbeat:
+            # Must not raise
+            NetBoxRQWorker.heartbeat(worker)
+        worker.register_birth.assert_not_called()
+        super_heartbeat.assert_called_once()
```
pheus: **nit:** `test_heartbeat_tolerates_redis_exception` might be slightly stronger than what the test proves.

Because `rq.Worker.heartbeat` is patched, this verifies that an exception during the new registration check does not prevent delegating to the parent heartbeat. In a real Redis outage, the parent heartbeat can still fail when it writes the actual heartbeat.

Could we rename this to something like `test_registration_check_exception_still_delegates_to_super_heartbeat()` so the expected behavior is clearer?

pheus (changes requested):
Overall, I like the approach and the new boolean helper is a nice improvement for the script-execution path. I’d ask for clarification or a small fix on the TTL handling before approving.

jeremystretch:

jeremystretch:
I've addressed point 2 and 3 above. Point 1 we'll just have to live with.

netbox/utilities/rqworker.py
```diff
@@ -1,15 +1,51 @@
...
+                # non-dead hash as an active worker. Re-add to the registry
+                # sets directly in that case; the heartbeat below will refresh
+                # the hash TTL.
+                if self.connection.exists(self.key) and not self.connection.hexists(self.key, 'death'):
+                    register_worker(self, self.connection)
+                else:
+                    self.register_birth()
+        except Exception:
+            logger.exception("Failed to verify worker registration.")
+        super().heartbeat(*args, **kwargs)
```
pheus, line 47: I think `heartbeat()` still misses one partial-loss state: the worker is still present in `rq:workers`, but its worker hash no longer exists.

Right now we only re-register if the worker key is missing from the global registry:

```python
if not self.connection.sismember(REDIS_WORKER_KEYS, self.key):
    ...
```

But the stale-registration case we’re trying to defend against can also be `rq:workers` still containing the worker key while `rq:worker:<name>` has expired or been lost. In that state this branch is skipped, and `super().heartbeat()` runs against a missing hash.

RQ’s heartbeat does `expire(self.key, timeout)` before `hset(self.key, 'last_heartbeat', ...)`; if the hash is missing, the `expire()` has no effect and the `hset()` creates a new hash containing only `last_heartbeat`, with no TTL and no `queues` field. That would make the worker hard to discover for queue-specific checks, and may also leave a non-expiring partial worker hash.

Could we treat “registry membership exists but worker hash is missing” as needing `register_birth()` too? Something along these lines would cover both sides of the partial-loss case:

```python
is_registered = self.connection.sismember(REDIS_WORKER_KEYS, self.key)
hash_exists = self.connection.exists(self.key)

if not is_registered or not hash_exists:
    logger.warning(f"Worker {self.name} registration is incomplete; re-registering.")

    if hash_exists and not self.connection.hexists(self.key, 'death'):
        register_worker(self, self.connection)
    else:
        self.register_birth()
```

I’d also update `test_heartbeat_skips_register_when_present` so the healthy case has both `is_member=True` and `hash_exists=True`, then add a new test for `is_member=True, hash_exists=False` asserting `register_birth()` is called.

pheus (commented):
Thanks for the update! The the TTL and queue-agnostic dashboard/API count concerns look addressed to me. I think there’s one remaining partial Redis-loss edge case in `NetBoxRQWorker.heartbeat()` where the global worker registry entry survives but the worker hash is gone; I left a comment with details.

Other than that, the direction looks good. But I might want to someone else take a second look.

pheus (approved):

Issue #22087, opened 2026-05-04
Employ FactoryBoy for object creation in tests


jeremystretch:
### Proposed Changes

Employ [FactoryBoy](https://github.com/FactoryBoy/factory_boy) throughout our test suite to simplify the creation of objects when running tests. This is accomplished by creating a factory class (e.g. `SiteFactory`) for each relevant model, to be reused whenever a new instance of that model is needed.

### Justification

Our test suite currently comprises many redundant lines of code effecting the creation of objects for testing. We can leverage FactoryBoy to simplify and optimize these tasks.

jnovinger:
I've been looking into the state of the tooling before we start work on this. Here's where things stand.

**factory_boy** (the original proposal): Last release was v3.3.3 in Feb 2025. The primary maintainer (rbarrois) hasn't interacted with the project since Aug 2024; the secondary maintainer (francoisfreitag) occasionally merges ready PRs but isn't triaging the 162 open issues or 46 open PRs. Someone opened [FactoryBoy/factory_boy#1153](https://github.com/FactoryBoy/factory_boy/issues/1153) a week ago asking if the project is still maintained. No response. There's no Django 6.x support in CI or classifiers, and no indication anyone with merge access is working toward it. Downloads are still strong (~22M/month), so it's not going away tomorrow, but there's real risk adopting a dependency with no active maintainer when we're tracking modern Django.

**Model Bakery** ([model-bakers/model_bakery](https://github.com/model-bakers/model_bakery)): The strongest alternative. v1.23.4 released Mar 2026, actively maintained, and the only library in this space with Django 6.0 support. ~1.6M downloads/month. Takes a different approach: auto-generates valid model instances by introspecting Django field types (`baker.make(Site)`) rather than requiring explicit factory class declarations. Less boilerplate per model, but also less explicit about what values are set.

Other options I looked at (Mixer, django-seed, polyfactory) are either dead or don't have Django ORM integration.

The original issue assumes factory_boy, but the maintenance situation has turned out to be less than ideal.

jnovinger:
factory_boy turned out to not be viable: the primary maintainer has been inactive since Aug 2024, there are 162 open issues with no triage, and there's no Django 6.x support. The PoC in #22235 evaluated model_bakery as the strongest alternative, covering both bare `baker.make()` and model_bakery's Recipe system across the circuits app. Full findings are in the PR description; the short version is that the conversion cost outweighs the value.

[...]

PR #22235, opened 2026-05-18
Ref #22087: Evaluate model_bakery for test fixture creation

jnovinger:
### Ref: #22087

Proof-of-concept converting the circuits app test suite to model_bakery. Two approaches are included for comparison: bare `baker.make()` calls (commits 1-2) and model_bakery's Recipe system (commit 3).

### Background

The original issue proposed FactoryBoy to simplify test object creation. The goal goes beyond line reduction: factory classes act as a single source of truth for test data, encode domain scenarios as reusable vocabulary (traits, lazy attributes), and make FK dependency graphs explicit and overridable. These properties matter more as AI tooling writes and maintains tests, since factory definitions are effectively machine-readable documentation of what valid test data looks like for a given model.

factory_boy turned out to be effectively unmaintained (primary maintainer inactive since Aug 2024, 162 open issues with no active triage, no Django 6.x support). model_bakery (v1.23.4, actively maintained, Django 6.0 support) is the strongest alternative. This PR evaluates it against both goals: reducing boilerplate and providing structural guardrails.

### What changed

**Commits 1-2: `baker.make()` conversion**
- All 6 circuits test files converted to `baker.make()` (708 tests, all passing)
- Net reduction of ~580 lines of test setup code
- Added `MPTT_ALLOW_TESTING_GENERATORS` setting for model_bakery to auto-create objects in MPTT foreign key chains
- No test assertions changed; only fixture creation was replaced
- model_bakery added to requirements (pinned at v1.23.4)

**Commit 3: Recipe-based conversion of `CircuitTestCase`**
- Added `baker_recipes.py` files for circuits, dcim, tenancy, and ipam apps
- Recipes define named, reusable model definitions with `seq()` for unique fields and `foreign_key()` for relationship composition
- Status variants (`active_circuit`, `planned_circuit`, `offline_circuit`) encode domain scenarios by name
- `CircuitTestCase` in `test_filtersets.py` converted to use recipes instead of bare `baker.make()`
- All 127 filterset tests pass

#### Line reduction (baker.make)

The `baker.make()` conversion cut ~580 lines from circuits test code (1085 deletions, 506 insertions). `setUpTestData` methods are shorter with less copy-paste across test classes. This compounds across the remaining apps.

#### MPTT friction (both approaches)

django-mptt detects model_bakery in the call stack and raises unless `MPTT_ALLOW_TESTING_GENERATORS` is enabled. Even with that setting, MPTT tree models (Region, SiteGroup, TenantGroup, ContactGroup, etc.) still need manual `.save()` calls because mptt manages tree fields itself. 11 MPTT models out of ~163 total (~7%), so the mixed approach affects a bounded set. This friction is identical for both `baker.make()` and recipes.

#### Ordering assumptions (baker.make)

Several tests relied on `Model.objects.all()[:N]` or `.objects.first()` with deterministic ordering by name. Baker's auto-generated names broke these. Each required individual diagnosis and fixing. Recipes with `seq()` mitigate this by producing predictable, incrementing values.

#### Recipes: modest structural improvement

model_bakery's Recipe system provides some of the structural benefits the original issue was after, though not all.

**What recipes provide:**
- `baker_recipes.py` per app gives a canonical location for reusable object definitions that AI tools and contributors can reference
- `foreign_key()` documents FK dependency chains explicitly, even when overridden at call time
- `extend()` enables named scenario variants (e.g., `planned_circuit = circuit.extend(status='planned')`)
- `seq()` handles unique field generation, avoiding the ordering fragility found with bare `baker.make()`

**What recipes lack compared to factory_boy:**
- No traits (composable named bundles of overrides); `extend()` covers simple variants but doesn't compose the way traits do
- No `LazyAttribute` (fields computed from sibling fields on the same object)
- No post-generation hooks (declarative M2M/tag setup after creation)
- Neither library calls `full_clean()`, so neither provides model-level validation guardrails

**What the recipe PoC showed in practice:**
- Status variants (`active_circuit.make(...)`, `planned_circuit.make(...)`) make each circuit's intent readable at the call site
- Filterset tests still need explicit field overrides for filter assertions, so recipe savings in this test class are modest
- The recipe benefit is more visible in `test_models.py` or `test_api.py` where tests just need "a valid circuit" without specific field values
- Recipe files are small (76 lines for circuits) but need to be created and maintained for each app

### Recommendation

model_bakery with recipes delivers line reduction plus a modest structural improvement over bare `baker.make()`. The `baker_recipes.py` files provide a per-app canonical location for test data definitions, `foreign_key()` documents relationships, and `extend()` enables named scenario variants. These are real benefits for AI tooling and contributor onboarding, though they fall short of factory_boy's full vocabulary (traits, lazy attributes, post-generation hooks).

factory_boy would have been the more complete solution for structural guardrails, but its maintenance situation rules it out.

Possible paths forward:

1. **Proceed with model_bakery using recipes as the primary pattern.** Convert app by app, creating `baker_recipes.py` files that document model relationships and common test scenarios. Use bare `baker.make()` for simple cases where a recipe adds no value.
2. **Build lightweight factory functions in-house** (per the [Luke Plant pattern](https://lukeplant.me.uk/blog/posts/test-factory-functions-in-django/)). Provides the single source of truth and explicit dependency benefits without an external dependency, at the cost of more upfront authoring per model and solving the same edge cases (MPTT, M2M, GFKs) that model_bakery already handles.
3. **Close this out** and revisit if factory_boy's maintenance situation improves or a new library emerges.


jnovinger:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Note that the scope has changed from the original issue, which called for using Factory Boy. Research found that Factory Boy is essentially unmaintained. The new scope is to evaluate model_factory by providing a PR converting the `circuits` app.

jnovinger:
> #### 1. `model_bakery` missing from requirements

Added to both `base_requirements.txt` and `requirements.txt` (pinned at 1.23.4).

> #### 2. `MPTT_ALLOW_TESTING_GENERATORS` added to production `settings.py`

Added a comment explaining the setting's purpose. The MPTT models (Region, SiteGroup, TenantGroup) are still created with manual `.save()` because django-mptt's tree field management requires it regardless of this setting. The setting prevents mptt from raising when baker auto-creates an MPTT model as part of a foreign key chain (e.g., `baker.make('dcim.Site')` auto-populates nullable `region` FK, which triggers MPTT's model_bakery guard without this setting).

> #### 3. Incomplete conversion creates internal inconsistency

Converted the remaining `bulk_create` calls in `test_filtersets.py` to `baker.make()`. The two exceptions that remain are `Location.objects.create()` (MPTT model, requires manual `.save()`) and `Interface.objects.create()` (complex parent/type relationship).

> #### 4. JSON import test now depends on `.objects.first()` ordering

Replaced with stored class attributes (`cls.providers`, `cls.circuittypes`, `cls.provider_networks`, `cls.virtual_circuit_types`, `cls.site`) set in `setUpTestData` and referenced directly in the test methods.

> #### 5. `baker.make(_quantity=N)` returns a list; without `_quantity` returns a single object

This is standard model_bakery API. No changes made.

PR #22243, opened 2026-05-19
Closes #19460: Support {lat}/{lon} placeholders in MAPS_URL

### LINKED ISSUE #19460: Allow more complex URLs in the map URL
### NetBox version

v4.3.0

### Feature type

### Proposed functionality

Hello,

I'd like to suggest a small improvement for the GPS Coordinate / MAPS_URL feature.
Currently NetBox renders the site/device coordinate as "{MAPS_URL}{lat},{lon}" when I click the map button.
My favorite online map is OpenStreetMap and with above scheme all I can do is set:
```
MAPS_URL = "https://www.openstreetmap.org/search?query="
```
which results in a link like: https://www.openstreetmap.org/search?query=48.858258,2.294498

OSM doesn't know if that is lat,lon or lon,lat. And it doesn't show a marker, doesn't allow me to set a different layer, zoom level, etc.

Instead I'd like to get a link like
https://www.openstreetmap.org/?mlat=48.858258&mlon=2.294498#map=16/48.858258/2.294498&layers=T

To implement this, the configured URL would be formatted, e.g.:
```
MAPS_URL_COORD = "https://www.openstreetmap.org/?mlat={lat}&mlon={lon}#map=16/{lat}/{lon}&layers=T"
```

The formatting could be done with a property in the Site/Device classes:
```python
@property
def maps_url_coord(self) -> str:
  fmt = ConfigItem("MAPS_URL_COORD")()
  return fmt.format(lat=self.latitude, lon=self.longitude)
```

Alternatively, the same could be achived with a user specified jinja template.

For compatibility with the current `MAPS_URL` and the "map address" feature, a new configuration option might be needed, hence I used `MAPS_URL_COORD` above.

The templates in `device.html` and `site.html` need to be aware of the new config option. If it's not set (default), `MAPS_URL` is used:

```
{% if config.MAPS_URL or config.MAPS_URL_COORD %}
  <a href="{% if config.MAPS_URL_COORD %}{{ object.maps_url_coord }}{% else %}{{ config.MAPS_URL }}{{ object.physical_address|urlencode }}{% endif %}" target="_blank" class="btn btn-primary btn-sm d-print-none">
    <i class="mdi mdi-map-marker"></i> {% trans "Map" %}
  </a>
{% endif %}
```
That template doesn't look too nice anymore, so rendering MAPS_URL could be done in the same python function.

### Use case

Better user experience when opening a map by allowing more flexibility in the map url.

### Database changes

No database changes.

### External dependencies

No new dependencies.

bctiemann:
## Summary

- When `MAPS_URL` contains `{lat}` or `{lon}` placeholders, it is treated as a Python format string and interpolated with the object's latitude and longitude
- When neither placeholder is present, existing behavior is preserved: coordinates are appended as a comma-separated `lat,lon` suffix
- The full URL is resolved in `GPSCoordinatesAttr.render()` before being passed to the template

**Example (OpenStreetMap with marker, zoom, and layer control):**
```python
MAPS_URL = "https://www.openstreetmap.org/?mlat={lat}&mlon={lon}#map=16/{lat}/{lon}&layers=T"
```

## Test plan

- [x] 4 new unit tests in `GPSCoordinatesAttrTestCase` covering legacy prefix, both placeholders, lat-only, lon-only (7 total, all pass)

Closes: #19460

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -197,11 +197,14 @@ This specifies the URL to use when presenting a map of a physical location by st
 **For GPS coordinates**, two formats are supported:
 
 * **Simple prefix** (default behavior): The latitude and longitude are appended as a comma-separated pair. For example, `https://maps.google.com/?q=` produces `https://maps.google.com/?q=48.858,2.294`.
-* **Format string**: Include `{lat}` and/or `{lon}` placeholders anywhere in the URL for full control over placement. For example:
+* **Coordinate placeholders**: Include `{lat}` and/or `{lon}` anywhere in the URL. Only these two literal placeholders are supported. For example:
 
 ```
 MAPS_URL = "https://www.openstreetmap.org/?mlat={lat}&mlon={lon}#map=16/{lat}/{lon}"
 ```
+
+!!! note
+    When `MAPS_URL` contains `{lat}` or `{lon}` placeholders, the "map it" button will only appear on pages with GPS coordinates — address-based map links will be suppressed, since the coordinate-format URL cannot be used with a plain address string.
 
 ---
 
--- a/docs/configuration/miscellaneous.md
+++ b/docs/configuration/miscellaneous.md
@@ -190,7 +190,18 @@ Setting this to `True` will display a "maintenance mode" banner at the top of ev
 
 Default: `https://maps.google.com/?q=` (Google Maps)
 
-This specifies the URL to use when presenting a map of a physical location by street address or GPS coordinates. The URL must accept either a free-form street address or a comma-separated pair of numeric coordinates appended to it. Set this to `None` to disable the "map it" button within the UI.
+This specifies the URL to use when presenting a map of a physical location by street address or GPS coordinates. Set this to `None` to disable the "map it" button within the UI.
+
+**For street addresses**, the URL must accept a free-form address string appended directly to it.
+
+**For GPS coordinates**, two formats are supported:
+
+* **Simple prefix** (default behavior): The latitude and longitude are appended as a comma-separated pair. For example, `https://maps.google.com/?q=` produces `https://maps.google.com/?q=48.858,2.294`.
+* **Format string**: Include `{lat}` and/or `{lon}` placeholders anywhere in the URL for full control over placement. For example:
+
+```
+MAPS_URL = "https://www.openstreetmap.org/?mlat={lat}&mlon={lon}#map=16/{lat}/{lon}"
+```
 
 ---
 
--- a/netbox/netbox/config/parameters.py
+++ b/netbox/netbox/config/parameters.py
@@ -229,7 +229,10 @@ PARAMS = (
         name='MAPS_URL',
         label=_('Maps URL'),
         default='https://maps.google.com/?q=',
-        description=_("Base URL for mapping geographic locations")
+        description=_(
+            "URL for mapping geographic locations. For GPS coordinates, include {lat} and {lon} placeholders, "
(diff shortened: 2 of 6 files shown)


bctiemann:
@claude /review

pheus (changes requested):
Thanks for moving the GPS URL-building logic out of the template.

I have a few concerns/suggestions before this is merged:

1. `MAPS_URL` is currently shared by both GPS coordinates and address rendering. With a coordinate-specific value like one using `{lat}` / `{lon}`, GPS links would work, but address map links may break because the address template still treats `MAPS_URL` as a prefix for the address string. I think we should either introduce a separate coordinate-specific setting, or make the map URL system explicitly support both address and coordinate placeholders.

2. `_build_coords_url()` is currently a static method on `GPSCoordinatesAttr`, but it receives `map_url` explicitly rather than using instance state. Could we either make it an instance method that uses `self.map_url`, or move it out as a helper into a local `utils.py` / suitable utilities module? The current form feels a little awkward because it looks class-specific, but does not depend on the class instance.

3. The URL formatting should probably be fail-safe. Since this is admin-controlled config, a malformed format string should not be able to break object detail pages. A simple `{lat}` / `{lon}` replacement may be safer than applying Python formatting to the whole URL, unless we explicitly want to support full format-string syntax and handle errors.

It would also be helpful to add tests for the shared `MAPS_URL` behavior with `AddressAttr`, and to clarify in the docs whether only literal `{lat}` / `{lon}` placeholders are supported or whether Python format strings are intended.

bctiemann:
Thanks @pheus , implemented your suggestions.

pheus (changes requested):
Thanks for the follow-up!

One remaining design concern I have is around `MAPS_URL` still being shared by both coordinate and address rendering. Suppressing address-based map links when `MAPS_URL` contains `{lat}` / `{lon}` is safer than rendering a broken link, but it also means configuring a coordinate-specific URL globally disables address map links. That may be acceptable, but I think we should be very deliberate about that behavior.

I still think the cleaner long-term option would be either:

- a separate coordinate-specific setting, e.g. `MAPS_COORDINATES_URL` / `MAPS_URL_COORD`; or
- explicit placeholder support for both address and coordinate use cases, e.g. `{address}`, `{lat}`, and `{lon}`.

If we keep the current suppression approach, could we also make that behavior clear in the dynamic config parameter description, not only in the docs? Admins changing `MAPS_URL` from the UI should be able to see that address map links will be disabled when coordinate placeholders are used.

One small code-organization suggestion: since this logic is now shared conceptually between `GPSCoordinatesAttr` and `AddressAttr`, could we move the map URL helpers into something like `netbox/ui/utils.py`? For example, a small `is_coordinate_map_url()` helper plus `build_coords_url()` would avoid duplicating the `{lat}` / `{lon}` detection and keep `attrs.py` focused on the attribute classes themselves.

bctiemann:
Agreed on the organizational separation of `utils.py` and `attrs.py`.

I think either option (separate `MAPS_URL` and `MAPS_COORDINATES_URL` settings which would have to each have their own form input and defined behavior if both are specified, or keeping the suppression behavior and adding more inline documentation) is bound to be inelegant; but the latter is probably more to my liking because it keeps the setting consolidated to a single field. The user is only ever going to see this setting rarely so a verbose piece of help text isn't a big concern, and they're probably only ever going to configure it once and then leave it alone forever.

pheus (approved):
Thanks for the update! This looks much better to me.

PR #22244, opened 2026-05-19
Fixes #22228: Correct VLAN Group total VLAN ID calculation for non-canonical VID ranges

### LINKED ISSUE #22228: VLANGroup.save() doesn't validate vid_ranges bounds metadata — bounds='[]' callers get silent miscounts and a hard 500 on singleton ranges
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.9

### Python Version

3.12

### Steps to Reproduce

Although we run version 4.4.9, as far as I can tell from the source code, the issue is still present in the latest release.

`VLANGroup.save()` since #20611 (NetBox 4.4.5) computes `_total_vlan_ids` with a formula that only works for canonical half-open ranges. Both `psycopg.NumericRange` and Django's `IntegerRangeField` accept any valid bounds string (`'[)'`, `'[]'`, `'()'`, `'(]'`), so a Python caller that explicitly passes `bounds='[]'` — which is a legitimate way to express "VIDs `lo` through `hi` inclusive" — silently triggers:

1. A per-range off-by-one undercount stored in `_total_vlan_ids` (cosmetic but wrong).
2. For a singleton range like `NumericRange(n, n, bounds='[]')`, a stored zero that crashes the entire VLAN-Groups API the next time anything reads through `annotate_utilization` — including the IPAM list view, per-row detail, every REST `PATCH`/`DELETE` (DRF reads first), bulk operations, and the OpenAPI swagger page.

The bug doesn't fire with defaults (psycopg's `bounds='[)'`, which is also what `string_to_ranges` and `IntegerRangeSerializer.to_internal_value` produce). It only fires when a caller overrides them. **NetBox itself was a caller** — `default_vid_ranges()` used `NumericRange(VLAN_VID_MIN, VLAN_VID_MAX, bounds='[]')` from v4.4.x through v4.5.x and was quietly switched to canonical form in v4.6.0 without a release note. Any custom script or plugin that took the same approach gets the same bug today.

The bug fires from the most natural Python construction. `NumericRange(100, 199, bounds='[]')` reads as "VIDs 100 through 199 inclusive" — and that *is* what it means in psycopg semantics — but `save()` silently stores `99` instead of `100`. To get the count right, the caller has to use the canonical form `NumericRange(100, 200)` (or `NumericRange(100, 200, bounds='[)')` explicitly), which requires knowing that the upper bound must be `vid_max + 1`. Nothing in the user-facing docs or model surface points that out; the only mention is an inline comment in `save()` itself.

Singletons make the friction worse. To express "just VID 100" using bounds metadata, the only construction that both survives `clean()` and reads as a one-VID range to a human is `NumericRange(100, 100, bounds='[]')`. The bare form `NumericRange(100, 100)` (default `bounds='[)'`) is the empty set `[100, 100)` and is rejected by `clean()` with a misleading "Ending VLAN ID in range must be greater than or equal to the starting VLAN ID (100-99)". The canonical-form construction `NumericRange(100, 101)` works, but requires the caller to mentally translate "VID 100" into "lower 100, upper 101" — exactly the translation `IntegerRangeSerializer.to_internal_value` does at the REST boundary, and exactly the thing a script author writing inclusive ranges does not realise they have to do.

So the path of least resistance lands on `bounds='[]'`, which is the only one that triggers the crash.

## Reproduction

```python
from django.db.backends.postgresql.psycopg_any import NumericRange
from ipam.models import VLANGroup

# Caller overrides default bounds with the (valid) inclusive-inclusive form:
g = VLANGroup.objects.create(
    name='single', slug='single',
    vid_ranges=[NumericRange(100, 100, bounds='[]')],
)
# save() runs `upper - lower` = 0
assert g._total_vlan_ids == 1   # FAILS: actual 0

# Now visit IPAM → VLAN Groups in the UI, or:
#     GET /api/ipam/vlan-groups/<id>/
# → 500, psycopg.errors.DivisionByZero
# The row is also unreachable via PATCH/DELETE (DRF's get_object() uses the
# same annotated queryset), so the standard API can't repair it.
```

Same setup with `bounds='[)'` (the default) stores `_total_vlan_ids = 1` and behaves correctly.

### Expected Behavior

Adding a VLAN group from a script with the range _100, 199, inclusive_ should give a group with 100 `total_vlan_ids`
Adding a group from a script with the range _100, 100, inclusive_ should give a group with 1 `total_vlan_ids`

And IF you manage to create a group with a range of 0 vlans, you should be able to fix the problem using either the UI or at least an API-request. 

### Observed Behavior

- The crash takes down a whole API endpoint — not just the bad row. One row with `_total_vlan_ids = 0` makes `/api/ipam/vlan-groups/` 500, regardless of why it got that way. That's a much bigger blast radius than the original misuse.
- The row becomes unreachable through the REST API: every write path (`PATCH`, `DELETE`, `bulk_destroy`, `bulk_update`) starts with `get_object()` against `VLANGroupQuerySet.annotate_utilization()`, which hits the same div/0. Repair currently requires Django shell or a custom NetBox script using `.update()` to bypass `save()`.
- The REST input layer (`IntegerRangeSerializer`) accepts `[100, 100]` as valid input and translates it to canonical form internally. A Python-level caller writing the same data shape (`bounds='[]'`, `NumericRange(100, 100, bounds='[]')`) gets surprisingly different behaviour from the same intent.

Just to be clear:

`NumericRange(100, 199)` gives a range from 100-198, with `total_vlan_ids=99` - Unexpected, but "correct"

`NumericRange(100, 200)` gives a range from 100-199, with `total_vlan_ids=100` - "Correct", but not an intuitive way of defining the VLANs in a range

`NumericRange(100, 199, bounds='[]')` gives a range from 100-199, with `total_vlan_ids=99` - Expected and intuitive range, but incorrect total_vlan_ids.

For singleton VLAN-ranges:

`NumericRange(100, 100)` gives an error messsage that "max vlan id must be >= min vlan id" - Unexpected.

`NumericRange(100, 101)` gives a range of 100-100 with total_vlan_ids=1 - Unintuititve.

`NumericRange(100, 100, bounds='[]')` gives a range of 100-100 with total_vlan_ids=0 and error 500 on pages and API-calls for vlan groups.

pheus:
### Closes: #22228

This fixes a VLANGroup range handling bug where `total_vlan_ids` could be calculated incorrectly when `vid_ranges` were provided with valid but non-canonical `NumericRange` bounds, such as `bounds='[]'`.

The existing validation logic already interpreted inclusive/exclusive bounds correctly in `clean()`, but `save()` still calculated `total_vlan_ids` from the raw in-memory range values. This could affect direct model usage, including Custom Scripts, where callers may create and save a `VLANGroup` instance and continue working with the same object without refreshing it from the database.

This change adds shared integer range helpers, normalizes VLAN group ranges to the database-expected `[)` form during `save()`, and recalculates `total_vlan_ids` from the normalized values. It also updates range display/serialization paths to derive inclusive bounds from the range metadata, so non-canonical in-memory ranges are represented correctly.

A migration recomputes existing cached totals, and utilization annotations now guard against division-by-zero for legacy rows with incorrect totals.

Note that this normalization is intentionally save-bound. Direct queryset updates such as `VLANGroup.objects.filter(...).update(vid_ranges=...)` still bypass model `save()` and validation, as with other direct ORM updates.

### DIFF
--- a/docs/models/ipam/vlangroup.md
+++ b/docs/models/ipam/vlangroup.md
@@ -18,6 +18,16 @@ A unique URL-friendly identifier. (This value can be used for filtering.)
 
 The set of VLAN IDs which are encompassed by the group. By default, this will be the entire range of valid IEEE 802.1Q VLAN IDs (1 to 4094, inclusive). VLANs created within a group must have a VID that falls within one of these ranges. Ranges may not overlap.
 
+Internally, each range is stored in PostgreSQL as a canonical half-open `[start, end)` interval. The `start` value is the first VID included in the range; the `end` value sits one above the last VID included.
+
+The REST API and UI both present ranges using inclusive bounds, so most users never see the half-open form. Users working with the stored value directly, such as through psql, raw SQL, Django ORM access in Custom Scripts or plugins, or third-party tools, should expect this canonical representation.
+
+For the range covering VLANs 100 through 200:
+
+* UI input: `100-200`
+* REST API range item: `[100, 200]`
+* Database row: `[100, 201)`
+
 ### Total VLAN IDs
 
 A read-only integer indicating the total count of VLAN IDs available within the group, calculated from the configured VLAN ID Ranges. For example, a group with ranges `100-199` and `300-399` would have a total of 200 VLAN IDs. This value is automatically computed and updated whenever the VLAN ID ranges are modified.
--- a/netbox/ipam/migrations/0090_vlangroup_recompute_total_vlan_ids.py
+++ b/netbox/ipam/migrations/0090_vlangroup_recompute_total_vlan_ids.py
@@ -1,0 +1,32 @@
+from django.db import migrations
+
+
+def recompute_vlangroup_total_vlan_ids(apps, schema_editor):
+    # PostgreSQL canonicalizes int4range values to '[)' on insert, so vid_ranges
+    # at rest is already canonical. Only total_vlan_ids needs fixing: the
+    # pre-fix VLANGroup.save() miscounted it for ranges supplied with
+    # non-canonical bounds (e.g. NumericRange(lo, hi, bounds='[]')).
+    VLANGroup = apps.get_model('ipam', 'VLANGroup')
+    db_alias = schema_editor.connection.alias
+
+    stale_groups = []
+    for group in VLANGroup.objects.using(db_alias).only('id', 'vid_ranges', 'total_vlan_ids').iterator(chunk_size=500):
+        total_vlan_ids = sum(
+            r.upper - r.lower for r in (group.vid_ranges or []) if r.lower is not None and r.upper is not None
+        )
+        if group.total_vlan_ids != total_vlan_ids:
+            group.total_vlan_ids = total_vlan_ids
+            stale_groups.append(group)
+
+    if stale_groups:
+        VLANGroup.objects.using(db_alias).bulk_update(stale_groups, ['total_vlan_ids'], batch_size=100)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('ipam', '0089_default_ordering_indexes'),
+    ]
+
+    operations = [
+        migrations.RunPython(recompute_vlangroup_total_vlan_ids, migrations.RunPython.noop),
+    ]
--- a/netbox/ipam/models/vlans.py
+++ b/netbox/ipam/models/vlans.py
@@ -12,7 +12,13 @@ from ipam.choices import *
 from ipam.constants import *
 from ipam.querysets import VLANGroupQuerySet, VLANQuerySet
 from netbox.models import NetBoxModel, OrganizationalModel, PrimaryModel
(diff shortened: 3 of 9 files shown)


pheus:

pheus:
@claude /review I’ve applied the small follow-ups for the migration and range utility:

* The migration now uses `.iterator(chunk_size=500)` for the VLANGroup scan.
* `check_ranges_overlap()` now uses `sorted()` instead of sorting the caller’s list in place.
* I also added a short comment in the `get_available_vids()` test to make it clear that the reassignment is intentional and exercises the unsaved in-memory path.

I’m keeping the `save()` behavior as-is. The in-memory canonicalization is intentional so callers, especially Custom Scripts, see the same canonical range representation after `save()` that PostgreSQL stores. The `update_fields` handling already persists `total_vlan_ids` together with `vid_ranges` when `vid_ranges` is being updated.

I also updated the PR description to call out the serializer representation fix and to note that direct queryset `update()` calls still bypass `save()` by design.

bctiemann (approved):

PR #22246, opened 2026-05-20
Closes #17127: Add user preference for metric/imperial measurements

### LINKED ISSUE #17127: Add user preference for metric or imperial measurements.
### NetBox version

v.4.0.8

### Feature type

New functionality

### Proposed functionality

Hot on the heels of https://github.com/netbox-community/netbox/issues/17126, it got me thinking. It would be great for users to be able to see weight units according to regional expectations.

### Use case

For global companies, you might have a user (in the US) create a device type using the weight unit 'pounds' because in the United States that is the generally accepted weight measurement.
But we must acknowledge that 99% of the rest of the world uses the metric system.
Fixing https://github.com/netbox-community/netbox/issues/17126 would make the device follow the same weight unit as the device type, but it would be nice if users had a choice in how the weight unit is rendered based on their preference.

How this might be implemented would be:
User Preferences -> Preferences -> User Interface -> create a new dropdown option named "Unit Measurement"
The available options in this dropdown would be:
Inherit (default option)
Metric
Imperial

This way, you provide maximum flexibility. The default option would be to match whatever the upstream device type weight measurement unit is.
But a user could override that by selecting either Metric or Imperial. Then NetBox would convert the weight during the page render. (which appears what its already doing. See https://github.com/netbox-community/netbox/issues/17126 for existing behavior.

### Database changes

Additional key for storing the new user preference.

### External dependencies

none.

bctiemann:
## Summary

- Adds a `ui.measurement_system` user preference (Metric / Imperial / Inherited) exposed in Profile → Preferences under the User Interface section
- Adds `WeightAttr` and `DistanceAttr` panel attribute classes that convert stored values to the preferred unit system, using the normalized gram/meter values from `WeightMixin` and `DistanceMixin`
- Adds public `abs_weight` / `abs_distance` property aliases on `WeightMixin` / `DistanceMixin` so templates can access normalized values without hitting Django's underscore restriction
- Adds `display_weight` and `display_distance` context-aware template tags used by `WEIGHT` table template code and `DistanceColumn`
- Updates rack and device `total_weight` templates to respect the preference, with correct singular/plural unit labels
- Passes `preferences` through `ObjectAttributesPanel` attr render context so conversion attrs can read the setting

Screenshot 2026-05-19 at 8 11 41 PM

Screenshot 2026-05-19 at 8 11 11 PM

## Test plan

- [ ] Set preference to **Imperial**: weights stored in kg display in lbs; distances stored in km/m display in mi/ft
- [ ] Set preference to **Metric**: weights stored in lb/oz display in kg; distances stored in mi/ft display in km/m
- [ ] Set preference to **Inherited**: stored value and unit shown as-is; total weight shows both kg and lb
- [ ] Verify pluralization: "1 lb", "2 lbs", "1 Kilogram", "2 Kilograms", "1 Pound", "2 Pounds"
- [ ] Run unit tests: `python manage.py test netbox.tests.test_ui.WeightAttrTestCase netbox.tests.test_ui.DistanceAttrTestCase`

Closes #17127


bctiemann:
@claude /review

bctiemann:
Addressed all actionable issues raised by Claude review.

jnovinger (approved):

PR #22252, opened 2026-05-20
Fixes #21895: Fix missing pagination controls for Job Log entries

### LINKED ISSUE #21895: Log output for job limited to 50 rows

pheus:
### Closes: #21895

This fixes missing pagination controls for job log entries by rendering the log entries table through NetBox’s embedded django-tables2/HTMX table path.

The job log view now constructs the log entries table as an embedded table and returns the standard `htmx/table.html` partial for HTMX pagination and per-page requests. This keeps the fix scoped to the job log page and avoids changing `ContextTablePanel`, `inc/paginator.html`, or the shared query-string behavior used by other views.

Tests have been added for the full job log page render, the embedded HTMX table partial, page navigation, per-page handling, and ensuring the embedded table does not push pagination parameters into the browser URL.

### DIFF
--- a/netbox/core/tests/test_views.py
+++ b/netbox/core/tests/test_views.py
@@ -1,7 +1,7 @@
 import json
 import urllib.parse
 import uuid
-from datetime import datetime
+from datetime import UTC, datetime
 
 from django.contrib.contenttypes.models import ContentType
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


pheus:

pheus:

pheus:
I’ve addressed the two actionable test points:

- Strengthened `test_log_page_prefixed_navigation()` so it now verifies that `?log-page=2` actually moves away from page 1 by asserting that `log line 0` is absent.
- Updated `_set_pagination_placement()` to merge into the existing user config instead of replacing it, preserving the pinned `pagination.per_page` value from `setUp()`.

For the other notes:

- `ContextTablePanel.should_render()` intentionally checks `context['table']`, because it receives the prepared panel context after `get_context()` has normalized the resolved table under that key.
- The `hx-push-url` behavior is pre-existing shared paginator behavior and does not affect this change.
- The django-tables2 variable-key querystring behavior is intentional here and covered by the paginator tests.

The targeted job log view tests pass locally; I’ll let CI confirm the full suite.

jnovinger (commented):
The pagination fix itself looks good and the test coverage is solid, including the backward-compatibility cases for callers that don't pass a `table` to the paginator partial.

One thing I want to flag: the switch from `{% load helpers %}` to `{% load django_tables2 %}` in `paginator.html` replaces the querystring tag implementation for every view that uses that partial, not just the job log page. The helpers version uses `QueryDict.urlencode(safe='/')`, while the django_tables2 version uses `urllib.parse.urlencode(params, doseq=True)` wrapped in `escape()`. For standard pagination parameters these produce identical output, but for views with more complex existing query strings (filter parameters with special characters, multi-value params) there could be subtle encoding differences.

Was this considered and deemed acceptable, or would it make sense to scope the tag change more narrowly (e.g., only in the `context_table.html` partial)? Given the wide blast radius, is this an appropriate change for a patch release targeting `main`?

Also: there's a merge conflict that needs resolving before this can be merged.

jnovinger (changes requested):

pheus:

pheus:

pheus:
Thanks for the review, @jnovinger !

I’ve reworked the implementation based on the feedback from you and the other maintainers. Instead of extending `ContextTablePanel` or changing `inc/paginator.html`, the job log entries table now uses NetBox’s existing embedded django-tables2/HTMX table rendering path.

With this approach:

- `ContextTablePanel` remains unchanged.
- `inc/paginator.html` remains unchanged.
- The query-string tag implementation is no longer changed globally.
- The job log table is marked as embedded, so pagination updates only the table content and does not push `page` / `per_page` into the browser URL.
- HTMX requests to the job log view return the standard `htmx/table.html` partial.

I’ve also updated the tests to match the new approach, covering the full page render, HTMX partial rendering, page navigation, per-page handling, and the embedded-table behavior.

jnovinger (approved):

PR #22253, opened 2026-05-20
Implement #22255 - Allow plugins to register custom serializer resolvers

### LINKED ISSUE #22255: Allow plugins to register custom serializer resolvers for get_serializer_for_model()
### NetBox version

v4.6.1

### Feature type

New functionality

### Proposed functionality

Provide a supported extension point that lets plugins override how NetBox resolves a REST API serializer for a given model. NetBox should consult plugin-supplied resolvers before falling back to its default import-path-based lookup, with deterministic precedence when multiple resolvers are registered. Default behavior is unchanged when no plugin registers a resolver.

### Use case

Plugins that generate models at runtime (or otherwise don't follow NetBox's serializer naming convention) currently have no supported way to plug into get_serializer_for_model() and must resort to monkey-patching - fragile, and easy for new core import sites to silently bypass. A first-class extension point gives these plugins a stable, supported integration path.

### Database changes

N/A

### External dependencies

N/A

arthanson:
### Closes: #22255

Adds `register_serializer_resolver()` as a supported plugin extension point for overriding REST API serializer resolution. Plugins can now register a callable that `utilities.api.get_serializer_for_model()` consults before its `default {app_label}.api.serializers.{Model}Serializer import-path lookup` — enabling clean integration for plugins with dynamically generated models, which previously had to monkey-patch.

Resolvers are tried in registration order; the first non-None return wins. If no resolver matches (or none are registered), the existing default lookup runs unchanged.


arthanson:
@claude /review

arthanson:
claude review, issues 1 and 2 are current conventions, issue 3 doc comment shortened.

netbox/netbox/registry.py
```diff
@@ -44,6 +44,7 @@ def __delitem__(self, key):
     'plugins': dict(),
     'request_processors': list(),
     'search': dict(),
+    'serializer_resolvers': list(),
```
jeremystretch: IMO we should limit plugins (and arguably even internal apps) to registering resolvers for their own models. It would probably make sense to register resolvers using a dictionary mapping them by app.

netbox/netbox/plugins/__init__.py
```diff
@@ -32,6 +32,7 @@
     'graphql_schema': 'graphql.schema',
     'menu': 'navigation.menu',
     'menu_items': 'navigation.menu_items',
+    'serializer_resolver': 'api.serializers.serializer_resolver',
```
jeremystretch: Given that we anticipate this being a fairly niche bit of functionality that arguably doesn't have any clear natural path, it might be prudent to not introduce a default path and only attempt to load the resolver if one is explicitly defined by the plugin.

jeremystretch (changes requested):

netbox/netbox/plugins/__init__.py
```diff
@@ -134,6 +135,13 @@ def ready(self):
         if user_preferences := self._load_resource('user_preferences'):
             register_user_preferences(plugin_name, user_preferences)
 
+        # Register serializer resolver (if explicitly defined). No default path is attempted.
+        if self.serializer_resolver:
+            register_serializer_resolver(
+                self.label,
+                import_string(f"{self.__module__}.{self.serializer_resolver}"),
```
jeremystretch: It's probably worth attempting the import first inside a dedicated `try`/`except` and raising an explicit "Invalid resolver path" exception in the event of an ImportError to better convey the problem to the user.

netbox/netbox/plugins/__init__.py
```diff
@@ -134,6 +135,13 @@ def ready(self):
         if user_preferences := self._load_resource('user_preferences'):
             register_user_preferences(plugin_name, user_preferences)
 
+        # Register serializer resolver (if explicitly defined). No default path is attempted.
```
jeremystretch:
```suggestion
        # Register serializer resolver (if defined)
```

netbox/netbox/plugins/registration.py
```diff
@@ -82,3 +83,24 @@ def register_user_preferences(plugin_name, preferences):
...
+    This is the supported extension point for plugins whose models are
+    generated dynamically (and therefore have no importable serializer at
+    the {app_label}.api.serializers.{Model}Serializer path) or that need
+    to override serializer resolution for specific models.
+
+    Resolver signature: resolver(model, prefix='') -> serializer class or None
+    """
+    if not callable(resolver):
+        raise TypeError(_("Serializer resolver must be callable"))
+    registry['serializer_resolvers'][app_label] = resolver
```
jeremystretch, line 110: We should probably check whether any resolver has already been registered for the app, and if so raise an error.

jeremystretch (changes requested):

jeremystretch (approved):

PR #22257, opened 2026-05-20
Fixes #22219: Label inline fields as required if any individual field is required

### LINKED ISSUE #22219: InlineFields rows missing required field indicator (asterisk)
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.13

### Steps to Reproduce

1. Navigate to /dcim/rack-types/add/
2. Observe the "Outer Dimensions" row, which uses `InlineFields('outer_width', 'outer_height', 'outer_depth', 'outer_unit', label=_('Outer Dimensions'))` in `dcim/forms/model_forms.py:276`
3. Compare the label styling with standard required fields like "Manufacturer" or "Model" on the same form

### Expected Behavior

Required fields rendered inside an `InlineFields` group should display the red asterisk (`required` CSS class) on the shared label, consistent with how `render_field` marks required fields.

### Observed Behavior

The shared label for `InlineFields` rows never gets the `required` CSS class, even when one or more of the inline fields are required. This is because `render_fieldset.html` line 29 renders the inline label without checking any field's `required` attribute:

```html
<label class="col col-3 col-form-label text-lg-end">{{ title|default:'' }}</label>
```

Compare with `render_field.html` line 10 which does check:

```html
<label ... class="col-form-label d-inline-block{% if field.field.required %} required{% endif %}">
```

jeremystretch:
### Closes: #22219

- Introduce the `any_required` template filter for check for a required form field within an iterable.
- Update `render_fieldset.html` to invoke `any_required` when rendering inline fields.
- Add tests

### DIFF
--- a/netbox/utilities/templates/form_helpers/render_fieldset.html
+++ b/netbox/utilities/templates/form_helpers/render_fieldset.html
@@ -26,7 +26,7 @@
     {% elif layout == 'inline' %}
       {# Multiple form fields on the same line #}
       <div class="row mb-3"{% if title %} role="group" aria-label="{{ title }}"{% endif %}>
-        <label class="col col-3 col-form-label text-lg-end">{{ title|default:'' }}</label>
+        <label class="col col-3 col-form-label text-lg-end{% if items|any_required %} required{% endif %}">{{ title|default:'' }}</label>
         {% for field in items %}
           <div class="col mb-1">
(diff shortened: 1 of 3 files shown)


jeremystretch:

bctiemann (approved):

PR #22258, opened 2026-05-20
Fixes #22232: Avoid enqueuing duplicate housekeeping jobs on startup

### LINKED ISSUE #22232: Duplicate background jobs
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

1. Start netbox
2. Wait
3. Go to /core/jobs
4. There are multiple scheduled 'System Housekeeping' jobs

Same problem with jobs registered via custom plugins (netbox 4.5.9)
Reproduced on demo.netbox.dev

Jobs page:
Image

Background tasks page: (for some reason it shows 3 jobs scheduled, while there are only 2 in jobs page)
Image

Background jobs details:
Image

Image

Image

### Expected Behavior

Should be single scheduled 'System Housekeeping' job

### Observed Behavior

There are multipe scheduled 'System Housekeeping' jobs

jeremystretch:
### Closes: #22232

Wrap the call to `enqueue()` with an advisory lock to avoid the race condition presented when multiple workers start up simultaneously. Note that we're not using `enqueue_once()` here because the jobs aren't guaranteed to have identical parameters.

### DIFF
--- a/netbox/netbox/jobs.py
+++ b/netbox/netbox/jobs.py
@@ -136,7 +136,8 @@ class JobRunner(ABC):
                 )
                 if job.object and getattr(job.object, "python_class", None):
                     kwargs["job_timeout"] = job.object.python_class.job_timeout
-                cls.enqueue(
+
+                enqueue_kwargs = dict(
                     instance=job.object,
                     name=job.name,
                     user=job.user,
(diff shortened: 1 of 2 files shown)


jeremystretch:

jeremystretch:
@claude /review Please verify that the concern raised above has been addressed.

pheus (changes requested):
The advisory lock looks like the right direction for the startup race, but I think this successor check may be broader than the system-job/housekeeping case it is fixing.

`JobRunner.handle()` is generic, and this query treats any enqueued job with the same runner/object/interval as the successor of the job that just completed. That seems safe for registered system jobs, but I’m not sure it is safe for generic recurring jobs whose runtime parameters differ.

A concrete case to check is scheduled custom scripts. The same `ScriptJob`/script object can be scheduled with the same `interval` but different `data`. Since those kwargs are not included in this successor lookup, schedule A could finish while schedule B is still enqueued, see B as its “successor,” and then skip rescheduling A. That would effectively collapse two legitimate recurring schedules into one.

Would it be safer to scope this duplicate-suppression branch to system jobs only, or otherwise include a stable schedule identity/fingerprint before deduping generic recurring jobs?

A regression test with two same-interval recurring jobs for the same runner/object but different kwargs would also make this behavior much clearer.

jeremystretch:
I've adapted the solution so that the advisory lock applies only to registered system jobs.

jeremystretch:

jeremystretch:
IMO the above concern is acceptable and arguably even preferred behavior.

pheus (approved):
Thanks for the update! This resolves my concern.
Nice work!

PR #22263, opened 2026-05-21
Closes: #19336 - Replace JS interface table toggles with server-side URL filters

### LINKED ISSUE #19336: Reword filtering of tabbed list views - convert from javascript to HTMX
### NetBox version

v4.2.8

### Feature type

### Proposed functionality

Currently for related objects we present some of them as tabs under the detail view (for example look at a Device that has interfaces and click on the interfaces tab).  We have some javascript filters under the Quick Search dropdown (see screenshot).  The proposal is to potentially convert these to HTMX as these predate HTMX being in NetBox.  But also look at re-working the UI to potentially provide a way to get to the normal filters to the list view.

These tabbed lists are a convenience and replicate functionality in the main list views, for example you can got to the Interfaces list and filter by a specific Device to get this list.  Potential solutions:

* Make the tabbed list views have all the filtering functionality of the main list views - so there is a button? that brings up the filtering dialog like on the main list views, also show the filter chits and allow saved filter functionality.
* Use the current UX (but remove the saved filter dropdown) but convert the JS filters to HTMX. This would fix the display issues with the pagination buttons.
* Instead of displaying the lists in tabs do what related objects do and just go the main list view that is filtered by the Device.

Image

### Use case

There are several issues with the current implementation:

* The drop-down are in javascript which is more difficult to change and replicates code that is much easier to implement in HTMX
* The UI is limited (a dropdown without being able to enter in any parameters like the normal filtering), not very obvious as it is on the Quick Search.
* The Saved filter dropdown appears on this page which does not work correctly.
* If you do filter the items it can cause UX confusion as the pagination does not reflect the filtering, for example if you have more then one page of interfaces and don't have any interfaces that are disabled and you click hide disabled, your lists will show as empty but you will still have the pagination showing.

### Database changes

None

### External dependencies

None

bctiemann:
## Summary

- Removes `interfaceTable.ts` (252 lines of JS that hid/showed table rows client-side via CSS), which caused incorrect pagination counts because the server queryset was never filtered
- Replaces the four toggle buttons (Hide Disabled, Hide Enabled, Hide Virtual, Hide Disconnected) with URL-based links using Django's `{% querystring %}` tag, passing `enabled`, `kind`, and `connected` params to the existing `InterfaceFilterSet`
- Removes unused `data-enabled`, `data-type`, `data-connected`, and `data-virtual` row attributes from `DeviceInterfaceTable`

### Intentional behavioral changes

**"Hide Virtual" now hides LAG, bridge, and wireless interfaces** — The old JS filtered rows by `data-type="virtual"`, which only matched the literal `virtual` interface type. The new `kind=physical` filter uses `InterfaceFilterSet.filter_kind()`, which excludes all non-connectable types (virtual, LAG, bridge, and all wireless variants). This is more consistent and useful: users hiding virtual interfaces almost certainly also want to hide LAG, bridge, and wireless types.

**"Hide Disconnected" uses path-completion semantics** — The old JS considered an interface "connected" if `cable IS NOT NULL OR mark_connected = True`. The new `connected=True` filter requires a complete, active cable path (`_path__is_active=True AND _path__is_complete=True`). This aligns with how "connected" is defined everywhere else in NetBox and correctly excludes interfaces whose cable terminates mid-path (e.g., attached to an unpatched panel).

Closes: #19336

## Test plan

- [ ] On a device with mixed enabled/disabled interfaces, click Hide Disabled — confirm only enabled interfaces appear and pagination reflects the filtered count
- [ ] Click Hide Disabled again (active state) — confirm it toggles off and all interfaces return
- [ ] Verify Hide Enabled, Hide Virtual, and Hide Disconnected filters all work correctly with accurate pagination
- [ ] Verify active filter shows a checkmark and the dropdown item is highlighted
- [ ] Verify no JS console errors on the interfaces tab

Generated with [Claude Code](https://claude.com/claude-code)


bctiemann:
The **Frontend** CI job failure is a pre-existing issue unrelated to this PR's changes. The graphiql and CSS dist files under `netbox/project-static/dist/` are stale relative to the current `yarn.lock`, so `scripts/verify-bundles.sh` fails whenever any `project-static/` file is touched. Tracked in #22264.

bctiemann:
@claude /review and pay particular attention to whether any gaps in existing functionality are introduced by this PR.

bctiemann:
Fixed issue #1 (restored `filterset_form` which was overzealously removed during development).

Two behavioral differences from the old JS implementation are intentional and kept:

**"Hide Virtual" scope** — The old toggle hid rows by matching `data-type="virtual"` on table row attributes, which only matched the literal `virtual` interface type. The new `kind=physical` filter uses `InterfaceFilterSet.filter_kind()`, which excludes all non-connectable types: virtual, LAG, bridge, and all wireless variants. The broader scope is intentional — a user filtering out virtual interfaces almost certainly wants LAG, bridge, and wireless gone too, and the old behaviour (showing LAG/bridge while hiding virtual) was inconsistent.

**"Hide Disconnected" strictness** — The old toggle set `data-connected="disconnected"` when `cable IS NULL AND NOT mark_connected`, so any cabled interface (even mid-path) was treated as connected. The new `connected=True` filter requires a complete, active cable path (`_path__is_active=True AND _path__is_complete=True`), consistent with how "connected" is defined throughout the rest of NetBox. An interface cabled to an unpatched panel was never truly connected and should appear under "Disconnected".

netbox/dcim/tables/devices.py
```diff
@@ -785,12 +785,8 @@ class Meta(DeviceComponentTable.Meta):
         )
         row_attrs = {
             'data-name': lambda record: record.name,
-            'data-enabled': lambda record: "enabled" if record.enabled else "disabled",
-            'data-virtual': lambda record: "true" if record.is_virtual else "false",
```
jeremystretch, line 789: `data-enabled` and `data-virtual` are also used for CSS highlighting. Suggest we leave these classes in place. (`data-type` and `data-connected` appear to be unused.)

jeremystretch (changes requested):
Other than the styling fix this looks good.

jeremystretch (approved):

PR #22268, opened 2026-05-21
Closes #22090: Extend test cases to analyze the number of SQL queries executed

### LINKED ISSUE #22090: Extend test cases to analyze the number of SQL queries executed
### Proposed Changes

Extend the API and view tests cases to call [`assertNumQueries`](https://docs.djangoproject.com/en/6.0/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) on relevant list views and compare them to records from prior executions (stored in JSON as persistent test data).

The implementation should include a mechanism (e.g. presence of an environment variable) to automatically update the records when appropriate (i.e. due to an expected change in the application).

### Justification

These will guard against the inadvertent introduction of new SQL queries due to the modification of some queryset, and help catch e.g. n+1 queries.

jeremystretch:


jeremystretch:

jeremystretch:

jeremystretch:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Point 2 from the previous review can be disregarded.

netbox/utilities/testing/query_counts.py
```diff
@@ -0,0 +1,139 @@
...
+            if fcntl is not None:
+                fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+            f.seek(0)
+            content = f.read()
+            existing = {}
+            if content:
+                try:
+                    existing = json.loads(content)
+                except json.JSONDecodeError:
+                    existing = {}
```
pheus: Tiny concern: should we let invalid existing `query_counts.json` fail loudly here?

Falling back to `{}` seems like it could accidentally wipe the other baselines for the app if the file was corrupted or merge-conflicted. Since the empty-file case is already covered, raising on invalid JSON might be the safer behavior.

pheus (changes requested):
Thanks for adding this! This looks like a really useful!

I left one tiny concern around invalid JSON handling in update mode, but otherwise this looks good to me.

pheus (approved):
Thanks! This is a really nice addition!

PR #22275, opened 2026-05-22
Closes: #22245: Fix OpenAPI request schemas for bulk update endpoints

### LINKED ISSUE #22245: Bulk Patch/Put OpenAPI Schema Doesn't Include ID Field
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1 (I have tested this as far back as v4.2.5)

### Python Version

3.12

### Steps to Reproduce

1. Spin up a new NetBox development instance running v4.6.1, following the developer documentation.
2. Run the local development server with `manage.py runserver`.
3. Navigate to `<url>/api/schema/swagger-ui/#/dcim/dcim_cables_bulk_partial_update`.
4. Expand the endpoint. Under "Request body (required)", next to "Example Value", click "Schema".
5. Observe that the request body is documented as `CableRequest[]`, and the item schema does not require or include an `id` field. (note: this is true for all bulk patch and put endpoints). 

### Expected Behavior

Bulk `PUT` and `PATCH` endpoints should document that each object in the request body must include the target object's numeric `id`, e.g.
```json
[
  {"id": 123, "description": "Updated description"}
]
```

The OpenAPI schema should include `id` as a required field for each item in the bulk update request body.

### Observed Behavior

The generated OpenAPI schema for bulk update endpoints uses the normal request model, e.g. `CableRequest`, which omits `id`. This does not match the runtime API contract.

At runtime, `BulkUpdateModelMixin.bulk_update()` first validates the request data with `BulkOperationSerializer`:

https://github.com/netbox-community/netbox/blob/64d3b114bc68e152869b964d54b220ecf1d50880/netbox/netbox/api/viewsets/mixins.py#L104

If `id` is omitted, NetBox returns `400 Bad Request`:
```json
[
  {
    "id": [
      "This field is required."
    ]
  }
]
```

jniec-js:
NetBox's bulk PUT/PATCH endpoints require each request item to include an `id`, as enforced by `BulkOperationSerializer`, but the generated OpenAPI schema reused the normal request serializer and omitted `id`. This caused generated clients to produce invalid request types. This change adds dedicated bulk update request schemas that include required `id` while preserving the usual writable fields, with PATCH variants keeping non-id fields optional.

Closes: #22245

### DIFF
--- a/netbox/core/api/schema.py
+++ b/netbox/core/api/schema.py
@@ -138,14 +138,30 @@ class NetBoxAutoSchema(AutoSchema):
         return super().get_operation_id()
 
     def get_request_serializer(self) -> typing.Any:
-        # bulk operations should specify a list
         serializer = super().get_request_serializer()
 
+        # Bulk update/partial-update has a special request shape: a list of
+        # writable objects plus a required `id` field. The normal writable
+        # serializer omits `id` because it is read-only, so don't use the generic
+        # bulk handling for these actions.
+        action = getattr(self.view, 'action', None)
+        if action in ('bulk_update', 'bulk_partial_update'):
+            get_bulk_update_request_serializer = getattr(
+                self.view,
+                'get_bulk_update_request_serializer',
+                None,
+            )
+            if get_bulk_update_request_serializer is not None:
+                return get_bulk_update_request_serializer(
+                    partial=(action == 'bulk_partial_update' or self.method == 'PATCH')
+                )
+
+        # Bulk creates/deletes should specify a list.
         if self.is_bulk_action:
             return type(serializer)(many=True)
 
-        # handle mapping for Writable serializers - adapted from dansheps original code
-        # for drf-yasg
+        # handle mapping for Writable serializers - adapted from dansheps original
+        # code for drf-yasg.
         if serializer is not None and self.method in WRITABLE_ACTIONS:
             writable_class = self.get_writable_class(serializer)
             if writable_class is not None:
--- a/netbox/netbox/api/serializers/bulk.py
+++ b/netbox/netbox/api/serializers/bulk.py
@@ -68,6 +68,3 @@ def get_bulk_update_serializer_class(serializer_class, *, partial=False):
 
     prefix = 'PatchedBulk' if partial else 'Bulk'
     return type(f'{prefix}{serializer_class.__name__}', bases, attrs)
-
-
-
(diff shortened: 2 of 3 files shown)


bctiemann:

bctiemann:
Thanks for the fix — the approach is solid. Two medium-priority issues worth addressing before merge:

**1. `meta.fields == '__all__'` will silently corrupt the schema**

```python
class Meta(meta):
    fields = ('id', *[f for f in meta.fields if f != 'id'])
```

If any serializer declares `Meta.fields = '__all__'` (valid DRF), this iterates over the characters of the string rather than the actual field names, producing a nonsensical `fields` tuple and a broken schema. Suggest guarding against it:

```python
if meta.fields == '__all__':
    fields = '__all__'
else:
    fields = ('id', *[f for f in meta.fields if f != 'id'])
```

**2. `id` is injected twice for the partial case**

`get_bulk_update_serializer_class` always includes `id` in `attrs` regardless of `partial`, and `BulkPartialUpdateSchemaMixin.get_fields()` also adds it. For PATCH serializers `id` ends up defined at the class level and then overwritten in `get_fields()`. The mixin wins so behaviour is correct, but the class-level `id` in `attrs` is dead weight on the partial path and makes the logic harder to follow. Either skip adding `id` to `attrs` when `partial=True` (since the mixin handles it), or drop the mixin and rely solely on `attrs` + `get_fields()` on the base class.

bctiemann (changes requested):
See comment in conversation.

jniec-js:
@bctiemann thanks for the feedback!

I believe I've addressed both concerns.

For the `Meta.fields = '__all__'` case, I added the suggested guard so we don't accidentally iterate over the string and generate a bogus field list. I also added a short comment around the dynamic `Meta` construction, since moving the `fields` assignment into the class body is tempting but runs into Python class-scope name lookup behavior.

For the duplicate `id` injection, I moved the `id` handling into a single shared `BulkUpdateSchemaMixin`. Both bulk PUT and bulk PATCH schemas now get their `id` field from the same place, copied from `BulkOperationSerializer`, so the schema stays aligned with the runtime validation. The PATCH-specific mixin now only makes non-`id` fields optional.

bctiemann:
@jniec-js sorry, a couple more things --

1. Minor, but worthwhile — please add `BulkUpdateSchemaMixin` to `__all__`
2. `@lru_cache(maxsize=None)` would be better to avoid inadvertend cache evictions under load

Thanks --

bctiemann (changes requested):
See above

jniec-js:
No worries! I wasn't aware of that behavior of lru_cache, thanks for flagging that. I've made the requested changes. 

Ruff flagged that we should use `@functools.cache` over `@functools.lru_cache(maxsize=None)`, so I have changed the decorator to satisfy the ruff recommendation.

bctiemann (approved):

PR #22284, opened 2026-05-26
Fixes #22282: Improve resolution of file path for imported S3 objects

### LINKED ISSUE #22282: `fetch()` method on S3Backend does not reliably resolve object keys
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

The `fetch()` method on the S3Backend class relies on `os.path.join()` to concatenate the base path (a temporary directory) with the S3 object key. For certain keys, this can result in erroneous behavior.

### Expected Behavior

The full path for the local file should be computed as a POSIX path.

### Observed Behavior

Only the basic `join()` logic is implemented.

jeremystretch:
### Closes: #22282

- Introduce the private method `_resolve_local_path()` on S3Backend to provide more robust local path resolution for new objects
- Ensure `local_path.cleanup()` is executed in the event of an exception
- Add tests

### DIFF
--- a/netbox/core/data_backends.py
+++ b/netbox/core/data_backends.py
@@ -1,9 +1,8 @@
 import logging
-import os
 import re
 import tempfile
 from contextlib import contextmanager
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from urllib.parse import urlparse
 
 from django import forms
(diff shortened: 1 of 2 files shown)


jeremystretch:

jeremystretch:
Added the above proposed cleanup logic.

arthanson (approved):

PR #22285, opened 2026-05-26
Fixes #22283: Restrict ScriptResultView queryset to current user

### LINKED ISSUE #22283: ScriptResultView does not adequately restrict Job queryset
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

The `get()` method on ScriptResultView (the UI view for viewing the result of a custom script execution) does not properly enforce view permissions on the underlying queryset:

```python
job = get_object_or_404(Job.objects.all(), pk=kwargs.get('job_pk'))
```

### Expected Behavior

`.restrict()` must be called on the queryset to limit matching objects based on the requesting user's assigned permissions.

### Observed Behavior

`.restrict()` is not applied to the queryset.

jeremystretch:
### Closes: #22283

- Fix the view queryset
- Add view tests

### DIFF
--- a/netbox/extras/tests/test_views.py
+++ b/netbox/extras/tests/test_views.py
@@ -1,18 +1,19 @@
+import uuid
 from unittest.mock import PropertyMock, patch
 
 from django.contrib.contenttypes.models import ContentType
 from django.test import tag
 from django.urls import reverse
 
-from core.choices import ManagedFileRootPathChoices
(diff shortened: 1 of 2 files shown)


jeremystretch:

jeremystretch:

bctiemann (approved):

Issue #22288, opened 2026-05-26
Rename `JINJA2_FILTERS` config parameter to `JINJA_FILTERS`


jeremystretch:
### Deprecated Functionality

Rename the `JINJA2_FILTERS` configuration parameter to `JINJA_FILTERS`. Preserve backward compatibility until a future breaking release.

### Scheduled removal

v5.0

### Justification

The Jinja project dropped the "2" suffix some time ago, and new config parameters (e.g. `JINJA_ENVIRONMENT_PARAMS`) are now prefixed with only `JINJA`. This rename achieves consistency and avoids confusion.

### Impact

Existing deployments making use of the `JINJA2_FILTERS` config parameter will need to update their configuration files (eventually).

PR #22289, opened 2026-05-26
Closes #22212: Support for exposing environment parameters in Jinja template context

### LINKED ISSUE #22212: Make designated environment parameters available within Jinja templates
### NetBox version

v4.6.0

### Feature type

New functionality

### Proposed functionality

1. Introduce a config parameter to whitelist environment variables that should be accessible within a Jinja template. This list should support wildcard matching. For example:

```python
JINJA_ENVIRONMENT_PARAMS = [
    "WEBHOOK_TOKEN_*",
    "DEFAULT_SECRET_ID",
]
```

2. Register by default a simple Jinja filter named `env`, which returns any specified environment parameter which matches the whitelist. (Nonexistent or non-matching parameters will return None.) This can be used within a Jinja template as:

```
Authorization: Bearer {{ env('WEBHOOK_TOKEN_3') }}
```

Note: To avoid a potential breaking change, any existing `JINJA2_FILTERS` configuration should override the default filter set. Administrators of existing deployments will be able to add the new `env` filter if desired.

### Use case

This will simplify the process of "import" system environment parameters into the Jinja template rendering context.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #22212

- Introduce the `JINJA_ENVIRONMENT_PARAMS` config parameter to whitelist (with glob support) specific environment parameters
- Introduce the `env` filter (`env_filter()`)
- Establish a list of default Jinja filters that is extended/overridden by `JINJA2_FILTERS`
- Add JinjaEnvFilterTestCase

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -145,6 +145,24 @@ Set this configuration parameter to `True` for NetBox deployments which do not h
 
 ---
 
+## JINJA_ENVIRONMENT_PARAMS
+
+Default: `[]`
+
+A list of system environment variable names which may be referenced from within Jinja2 templates via the built-in [`env`](#jinja2_filters) filter. Patterns may include wildcards (matched using Python's `fnmatch` syntax). Any variable whose name does not match an entry in this list cannot be referenced from a template. For example:
+
+```python
+JINJA_ENVIRONMENT_PARAMS = [
+    'WEBHOOK_TOKEN_*',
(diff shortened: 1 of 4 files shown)


jeremystretch:

jeremystretch:
1. We're going to rename `JINJA2_FILTERS` (see #22288).
2. Added a note in the config parameter's documentation.

jnovinger (approved):

PR #22294, opened 2026-05-26
#21025: Optimize rendering of config context data

### LINKED ISSUE #21025: Pre-render and cache config context data
### NetBox version

v4.4.8

### Feature type

Data model extension

### Proposed functionality

Currently, config context data is compiled for a device/VM on demand from the set of applicable ConfigContext instances that have been defined. In scenarios involving a large number of these instances, this can result in very complex, slow queries leading to long response times, as noted in #20358.

This FR proposes instead pre-calculating the config context data for all relevant objects (devices and VMs) via non-blocking background tasks whenever context data or devices/VMs are modified. This pre-rendered data could be cached locally on each device/VM to maximize query performance.

The primary challenge of this approach will be in reliably detecting all triggering changes. These include

1. The creation, modification, or deletion of a ConfigContext object
2. The creation or modification of a device/VM which impacts an associated ConfigContext object (e.g. changing the role of a device)

All updates to pre-rendered context data should be done via non-blocking asynchronous background jobs which are triggered after the relevant objects have been updated. Note that this does introduce a small window in which the rendered data for an object may be out of date. This could potentially be addressed by immediately nullifying the cached data whenever an update is queued, which would force an on-demand rendering of the context data for the object.

### Use case

This approach shifts the delay of rendering config contexts to the background, optimizing for fast read queries and greatly reduced rendering times for device and VM config templates.

### Database changes

A "private" nullable JSON field will likely be added to the Device and VirtualMachine models to store the pre-rendered config context data.

### External dependencies

N/A

jeremystretch:
### Closes: #21025

Optimizes rendering of config context data for devices and virtual machines by pre-rendering and caching each object's merged context, eliminating the expensive per-request aggregation in the common (warm-cache) case.

#### Cache fields & read path

- Add two fields to Device and VirtualMachine (via `ConfigContextModel`):
  - `_config_context_data` — JSON field storing the cached, pre-rendered context data (NULL means "invalidated; render on demand")
  - `_config_context_generation` — monotonic counter used to serialize recalculations and avoid race conditions (compare-and-set)
- Split `get_config_context()` into a cached read path and a new `render_config_context()` method. `get_config_context()` returns the cached data when present and falls back to on-demand rendering when the cache is NULL.
- `get_config_context()` returns a deep copy of the cached blob, so callers (e.g. `ObjectRenderConfigView`) may safely mutate the result, matching the fresh-dict guarantee of the legacy render path.
- Exclude the cache fields from change-log snapshots (`serialize_object()`) to avoid noisy diffs.

#### Invalidation

- Add `ConfigContext.get_affected_objects()`, the inverse of `ConfigContextQuerySet.get_for_object()`, returning the devices/VMs covered by a context's scope.
- Introduce signal handlers (`extras/signals.py`, `extras/cache.py`) that invalidate cached data upon any applicable change: ConfigContext create/update/delete and scope M2M changes; device/VM scope-field changes; upstream related-object FK changes and MPTT reparents; and SET_NULL deletes of upstream objects.

#### Background rendering & population

- Introduce `RenderConfigContextJob`, a coalesced background job that repopulates all objects with a NULL cache (with a compare-and-set generation guard and a re-scan loop to catch caches invalidated mid-sweep).
- Introduce the `rebuild_config_context_cache` management command (with `--force`), wired into `upgrade.sh` for initial population.

#### Read-path query optimization

- Add a partial index on Device/VirtualMachine covering only rows whose `_config_context_data` is NULL, keeping the renderer's invalidated-row scan cheap on large tables.
- Add `annotate_config_context_data(only_invalidated=True)`, which wraps the aggregation subquery in a `CASE` guarded on the NULL-cache condition. PostgreSQL short-circuits the `CASE`, so the correlated subquery executes only for invalidated rows and is skipped entirely for warm rows. This collapses the cold-cache list path from N+1 queries to a single query while preserving the warm-cache fast path. (Verified via `EXPLAIN ANALYZE`: 0 subquery executions when all rows are warm, one per invalidated row otherwise.)
- Reintroduce `ConfigContextQuerySetMixin` (now applying the conditional annotation) on the Device/VM REST viewsets, and restore `ConfigContextMixin.get_queryset()` on the GraphQL types with the same conditional annotation (only when `config_context` is requested).

#### API changes

- Remove the conditional `*WithConfigContextSerializer` REST serializers; `config_context` is now a field on the base device/VM serializers.
- Remove the `ConfigContextQuerySetMixin`-based serializer selection and the `annotate_config_context_data()` calls from UI & GraphQL querysets in favor of the cache + conditional annotation.
- Drop support for suppressing `config_context` via `?exclude=config_context`; config context data is now always returned for devices and virtual machines (the parameter is silently ignored).


jeremystretch:
@claude /review

jeremystretch:
@claude /review Please review again, considering the following responses to the above concerns:

1. Not an issue. (Bizarre that this was even flagged.)
2. Fixed
3. Acceptable for now. We'll likely explore improving config context "plugability" in the future.
4. Known; will flag this in the release notes.
5. Added tests
6. Added tests
7. Fixed
8. Acceptable for now. We might adopt a more robust approach after testing at scale.

netbox/extras/models/configs.py
```diff
@@ -215,12 +216,131 @@ def sync_data(self):
...
+    def _get_affected_object_filters(self):
+        """
+        Build the Q expressions matching Devices and VirtualMachines in this context's scope.
+        Returns (device_q, vm_q). Does NOT consider `is_active` — callers that need that should
+        check it separately. For invalidation purposes, we want the scope set regardless of
+        whether the context is currently active (toggling is_active also requires invalidation).
+        """
+        from extras.models.tags import TaggedItem
+
+        def _mptt_descendants(m2m):
```
arthanson: will need to double-check this with the Ltree stuff once merged. at least rename it

netbox/extras/models/configs.py
```diff
@@ -233,9 +353,20 @@ class Meta:
         abstract = True
 
     def get_config_context(self):
+        """
+        Return the merged config context for this object. If a pre-rendered cache is present
+        (`_config_context_data`), return it directly. Otherwise, fall back to rendering on demand.
+        """
+        cached = getattr(self, '_config_context_data', None)
+        if cached is not None:
+            return cached
```
arthanson: callers are mutating this (extras/views.py:1288) I think me may want to return a new copy here instead of the direct cached object?

arthanson (changes requested):
Everything looks good, just one small item I'm not sure is a bug with returning the cached item that gets mutated, it is a change from previous and probably should return a copy to avoid inadvertent side-effects.

netbox/extras/models/configs.py
jeremystretch: Agreed; will mark this work as blocked pending completion of #21418.

PR #22295, opened 2026-05-26
Closes #22012: Allow rack starting unit to be zero

### LINKED ISSUE #22012: Allow rack starting unit of 0
### NetBox version

4.5.2

### Feature type

### Proposed functionality

Currently, Netbox only allows starting unit of 1 or greater in Racks and Rack Types. Netbox should allow a starting unit of 0, to match rack unit numbering from certain manufacturers which start at 0 rather than 1.

### Use case

Accurately model device positioning and numbering, avoid confusion when numbering does not match at the site.

### Database changes

_No response_

### External dependencies

_No response_

mburggraf:
### Closes: #22012

Files were changed to allow a rack to have a starting unit of zero.

The constant `RACK_STARTING_UNIT_DEFAULT` was kept as 1, because I guess that even though 0 should be possible as starting unit, the default should stay at one. 
If this assumption is not correct, the default value can be changed as well.

### DIFF
--- a/docs/models/dcim/racktype.md
+++ b/docs/models/dcim/racktype.md
@@ -36,7 +36,7 @@ The height of the rack, measured in units.
 
 ### Starting Unit
 
-The number of the numerically lowest unit in the rack. This value defaults to one, but may be higher in certain situations. For example, you may want to model only a select range of units within a shared physical rack (e.g. U13 through U24).
+The number of the numerically lowest unit in the rack. This value defaults to one, but may differ in certain situations (minimal valid value is zero). For example, you may want to model only a select range of units within a shared physical rack (e.g. U13 through U24).
 
 ### Outer Dimensions
 
--- a/docs/models/dcim/racktype.md
+++ b/docs/models/dcim/racktype.md
@@ -36,7 +36,7 @@ The height of the rack, measured in units.
 
 ### Starting Unit
 
-The number of the numerically lowest unit in the rack. This value defaults to one, but may differ in certain situations (minimal valid value is zero). For example, you may want to model only a select range of units within a shared physical rack (e.g. U13 through U24).
+The number of the numerically lowest unit in the rack. This value defaults to one, but may differ depending on the numbering scheme used in the rack. The minimum valid value is zero. For example, you may want to model only a select range of units within a shared physical rack (e.g. U13 through U24).
 
 ### Outer Dimensions
 
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -1,3 +1,5 @@
+import decimal
+
 from django import forms
 from django.contrib.contenttypes.models import ContentType
 from django.core.validators import EMPTY_VALUES
@@ -785,7 +787,7 @@ class DeviceForm(TenancyForm, PrimaryModelForm):
 
         # Rack position
         position = self.data.get('position') if self.data.get('position') is not None else self.initial.get('position')
-        if position is not None:
+        if isinstance(position, (int, decimal.Decimal)):
             self.fields['position'].widget.choices = [(position, f'U{position}')]
 
 
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -1,5 +1,3 @@
-import decimal
-
 from django import forms
 from django.contrib.contenttypes.models import ContentType
 from django.core.validators import EMPTY_VALUES
@@ -786,8 +784,8 @@ class DeviceForm(TenancyForm, PrimaryModelForm):
             self.fields['oob_ip'].widget.attrs['readonly'] = True
 
         # Rack position
-        position = self.data.get('position') if self.data.get('position') is not None else self.initial.get('position')
-        if isinstance(position, (int, decimal.Decimal)):
+        position = self.data.get('position')
+        if position not in EMPTY_VALUES or (position := self.initial.get('position')) not in EMPTY_VALUES:
             self.fields['position'].widget.choices = [(position, f'U{position}')]
 
 
--- a/netbox/dcim/forms/model_forms.py
+++ b/netbox/dcim/forms/model_forms.py
@@ -784,8 +784,8 @@ class DeviceForm(TenancyForm, PrimaryModelForm):
             self.fields['oob_ip'].widget.attrs['readonly'] = True
 
         # Rack position
-        position = self.data.get('position') or self.initial.get('position')
-        if position:
+        position = self.data.get('position') if self.data.get('position') is not None else self.initial.get('position')
+        if position is not None:
             self.fields['position'].widget.choices = [(position, f'U{position}')]
 
 
--- a/netbox/dcim/migrations/0237_allow_rack_starting_unit_zero.py
+++ b/netbox/dcim/migrations/0237_allow_rack_starting_unit_zero.py
@@ -8,6 +8,13 @@ class Migration(migrations.Migration):
         ('dcim', '0236_moduletype_component_counts'),
     ]
 
+    def clear_legacy_0u_positions(apps, schema_editor):
+        Device = apps.get_model('dcim', 'Device')
+        Device.objects.filter(
+            device_type__u_height=0,
+            position=0,
+        ).update(position=None)
+
     operations = [
         migrations.AlterField(
             model_name='device',
@@ -32,4 +39,5 @@ class Migration(migrations.Migration):
             name='starting_unit',
             field=models.PositiveSmallIntegerField(default=1, validators=[django.core.validators.MinValueValidator(0)]),
         ),
+        migrations.RunPython(clear_legacy_0u_positions, migrations.RunPython.noop),
     ]
--- a/netbox/dcim/migrations/0237_allow_rack_starting_unit_zero.py
+++ b/netbox/dcim/migrations/0237_allow_rack_starting_unit_zero.py
@@ -5,7 +5,7 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('dcim', '0235_cabletermination_circuit_site_cache'),
+        ('dcim', '0236_moduletype_component_counts'),
     ]
 
     operations = [
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -847,7 +847,7 @@ class Device(
                             "device."
                         )
                     })
-                if self.device_type.is_child_device and self.position:
+                if self.device_type.is_child_device and self.position is not None:
                     raise ValidationError({
                         'position': _(
                             "Child device types cannot be assigned to a rack position. This is an attribute of the "
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -584,7 +584,7 @@ class Device(
         decimal_places=1,
         blank=True,
         null=True,
-        validators=[MinValueValidator(1), MaxValueValidator(RACK_U_HEIGHT_MAX + 0.5)],
+        validators=[MinValueValidator(0), MaxValueValidator(RACK_U_HEIGHT_MAX + 0.5)],
         verbose_name=_('position (U)'),
         help_text=_('The lowest-numbered unit occupied by the device')
     )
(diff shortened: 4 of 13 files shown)


netbox/dcim/migrations/0237_allow_rack_starting_unit_zero.py
```diff
@@ -0,0 +1,35 @@
+import django.core.validators
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dcim', '0235_cabletermination_circuit_site_cache'),
+    ]
```
pheus, line 9: This migration needs to be regenerated/renumbered after rebasing on current `main`.

Current `main` already has a `dcim` migration numbered `0236`, so adding another `0236_*` migration here will create conflicting migration leaves. Please rebase and regenerate this as the next migration, with the dependency pointing at the current `dcim` leaf migration.

netbox/dcim/models/devices.py
```diff
@@ -812,24 +812,24 @@ def clean(self):
...
-        if self.position and not self.face:
+        if self.position is not None and not self.face:
             raise ValidationError({
                 'face': _("Must specify rack face when defining rack position."),
             })
 
         # Prevent 0U devices from being assigned to a specific position
         if hasattr(self, 'device_type'):
-            if self.position and self.device_type.u_height == 0:
+            if self.position is not None and self.device_type.u_height == 0:
```
pheus, line 832: I think we need to untangle the meaning of `position=0` before this can merge.

With this change, `position=0` becomes a valid mounted rack unit for racks whose `starting_unit` is zero. But elsewhere, `Rack.get_0u_devices()` still uses `position=0` to identify 0U devices:

```python
return self.devices.filter(position=0)
```

That means U0-mounted devices and 0U devices can now collide semantically. It also looks like this validation will reject existing/intended 0U devices that use `position=0`.

Could we please audit the 0U handling and add tests for both cases?

```text
- a normal device mounted in U0 is treated as a mounted rack-unit device
- an actual 0U device still appears wherever NetBox expects 0U devices
```

Depending on the intended model semantics, 0U devices probably need to be identified by `device_type__u_height=0` and/or another explicit condition rather than by `position=0` alone.

netbox/dcim/models/racks.py
```diff
@@ -559,7 +559,7 @@ def get_available_units(self, u_height=1.0, rack_face=None, exclude=None, ignore
         :param ignore_excluded_devices: Ignore devices that are marked to exclude from utilization calculations
         """
         # Gather all devices which consume U space within the rack
-        devices = self.devices.prefetch_related('device_type').filter(position__gte=1)
+        devices = self.devices.prefetch_related('device_type').filter(position__isnull=False)
```
pheus, line 562: This looks directionally correct for racks starting at U0, but I think it exposes the existing ambiguity around `position=0`.

`get_rack_units()` / `get_available_units()` now include devices at any non-null position, while `get_0u_devices()` still identifies 0U devices with `position=0`. Could we please update the 0U device logic as part of this PR so U0-mounted devices and 0U devices cannot be confused?

netbox/dcim/forms/model_forms.py
```diff
@@ -784,8 +784,8 @@ def __init__(self, *args, **kwargs):
             self.fields['oob_ip'].widget.attrs['readonly'] = True
 
         # Rack position
-        position = self.data.get('position') or self.initial.get('position')
-        if position:
+        position = self.data.get('position') if self.data.get('position') is not None else self.initial.get('position')
+        if position is not None:
```
pheus: Good catch preserving `0` here. One edge case: on a bound form, a blank submitted value can be `''`, which is not `None`. With the current check, that blank value can be treated as an actual position choice.

Could we distinguish `0` from blank values explicitly?

The important bit is that `0` remains valid, while `None`/`''` are still treated as empty.

netbox/dcim/tests/test_models.py
```diff
@@ -370,6 +370,50 @@ def test_utilization(self):
...
+
+        # Test device in slot U0
+        Device(
+            name='Device 1',
+            device_type=DeviceType.objects.first(),
+            role=DeviceRole.objects.first(),
+            site=site,
+            rack=rack0,
+            position=0
+        ).save()
```
pheus: Could we adjust this test to exercise the validation paths users actually hit?

This creates devices with `.save()` directly, including a U0-mounted device without `face`. That bypasses `full_clean()` and can hide regressions in `Device.clean()`, forms, and the API serializer.

Suggested coverage:

```text
- `Rack.starting_unit = 0` accepts a device at `position=0` with a valid `face`
- the default `starting_unit = 1` still rejects `position=0`
- for a 42U rack starting at 0, `position=41` is valid and `position=42` is invalid
- rack utilization / available units include U0 correctly
- a mounted U0 device is not returned/displayed as a 0U device
- an actual 0U device still behaves as expected
```

netbox/dcim/tests/test_models.py
```diff
@@ -370,6 +370,50 @@ def test_utilization(self):
...
+            device_type=DeviceType.objects.first(),
+            role=DeviceRole.objects.first(),
+            site=site,
+            rack=rack0,
+            position=42,
+            face=DeviceFaceChoices.FACE_FRONT,
+        ).save()
+
+        with self.assertRaises(ValidationError):
+            rack0.clean()
```
pheus: I think this should validate the device rather than the rack.

The invalid object in this setup is the device placed at `position=42`, but the assertion calls `rack0.clean()`. Could we keep a reference to the device and call `full_clean()` on it instead? That would prove the device/rack-position validation catches the out-of-range U position through the same path used by forms/API validation.

netbox/dcim/models/devices.py
pheus: Minor consistency nit: this still uses a truthiness check for `position`.

```python
if self.device_type.is_child_device and self.position:
```

Since this PR makes `0` a meaningful value, it would be clearer and safer to use the same explicit style as the rest of this method:

```python
if self.device_type.is_child_device and self.position is not None:
```

Even if another validation branch currently catches this first, keeping the checks consistent will make future changes less surprising.

docs/models/dcim/racktype.md
```diff
@@ -36,7 +36,7 @@ The height of the rack, measured in units.
 
 ### Starting Unit
 
-The number of the numerically lowest unit in the rack. This value defaults to one, but may be higher in certain situations. For example, you may want to model only a select range of units within a shared physical rack (e.g. U13 through U24).
+The number of the numerically lowest unit in the rack. This value defaults to one, but may differ in certain situations (minimal valid value is zero). For example, you may want to model only a select range of units within a shared physical rack (e.g. U13 through U24).
```
pheus: Tiny wording suggestion:

```suggestion
The number of the numerically lowest unit in the rack. This value defaults to one, but may differ depending on the numbering scheme used in the rack. The minimum valid value is zero. For example, you may want to model only a select range of units within a shared physical rack (e.g. U13 through U24).
```

pheus (changes requested):
Thanks for working on this! The overall direction makes sense to me, and keeping the default `starting_unit` at `1` seems correct for backward compatibility.

I’d like to request changes mainly for two blockers:

```text
1. The migration needs to be regenerated after rebasing because current `main` already has a `0236` migration.
2. `position=0` now represents a real rack unit, but parts of the existing code still use `position=0` as the 0U-device marker.
```

Once those are addressed and the tests cover both U0-mounted devices and actual 0U devices through the normal validation paths, this should be much safer to merge.

netbox/dcim/models/devices.py
mburggraf: Maybe I'm misunderstanding something here, but before applying the new database migration, a position value of 0 shouldn't have been possible resulting in a ValidationError – therefore the function `Rack.get_0u_devices()` should have always returned an empty QuerySet.
I can't find any occurence in the main application code, where this function is actually used. Do you have an example where this function is used and gives back valid objects? Could it be possible, that the function never worked to begin with and it just never was detected?
Nevertheless, as suggested it's definately worth to change the function content to:
```python
return self.devices.filter(device_type__u_height=0)
```

netbox/dcim/models/devices.py
mburggraf: Must have missed this one, thanks.

netbox/dcim/tests/test_models.py
mburggraf: Actually I copied the code from the `test_utilization` function above, thinking about keeping a consistent look to the tests. But the comment is valid, so I'm going to rewrite the test.

netbox/dcim/forms/model_forms.py
```diff
@@ -784,8 +786,8 @@ def __init__(self, *args, **kwargs):
             self.fields['oob_ip'].widget.attrs['readonly'] = True
 
         # Rack position
-        position = self.data.get('position') or self.initial.get('position')
-        if position:
+        position = self.data.get('position') if self.data.get('position') is not None else self.initial.get('position')
+        if isinstance(position, (int, decimal.Decimal)):
```
pheus: Thanks for addressing the blank-position edge case. I think this version may regress bound form submissions, though.

On a submitted Django form, `self.data.get('position')` will usually be a string such as `'0'`, `'1'`, or `'41'`, not an `int`/`Decimal`. That means this branch won’t add the current rack position back to the APISelect widget for any submitted position, not just blank values.

Could we use `EMPTY_VALUES` instead so blank values are excluded but submitted position strings are still preserved?

```python
position = self.data.get('position')
if position in EMPTY_VALUES:
    position = self.initial.get('position')

if position not in EMPTY_VALUES:
    self.fields['position'].widget.choices = [(position, f'U{position}')]
```

That keeps `0`/`'0'` valid while still treating `None` and `''` as empty.

netbox/dcim/migrations/0237_allow_rack_starting_unit_zero.py
```diff
@@ -0,0 +1,35 @@
+import django.core.validators
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dcim', '0236_moduletype_component_counts'),
+    ]
+
```
pheus, line 10: I think we may need a small data migration for existing 0U devices.

Before this change, `get_0u_devices()` identified 0U devices using `position=0`. This PR changes that semantic so `position=0` can be a real rack unit, and `Device.clean()` now rejects 0U device types with any non-null position. Existing installations may therefore have 0U devices stored with `position=0`, which would become invalid on the next edit/API update and could be picked up by rack-space calculations.

Could we normalize those legacy rows as part of this migration?

```python
def clear_legacy_0u_positions(apps, schema_editor):
    Device = apps.get_model('dcim', 'Device')
    Device.objects.filter(
        device_type__u_height=0,
        position=0,
    ).update(position=None)
```

Then add a `RunPython(clear_legacy_0u_positions, migrations.RunPython.noop)` operation.

netbox/dcim/tests/test_models.py
```diff
@@ -435,6 +439,117 @@ def test_utilization(self):
...
+            u_height=42
+        )
+        rack0.save()
+
+        self.assertEqual(rack0.starting_unit, 0)
+
+        # Test device in slot U0
+        device1 = Device(
+            name='Device 1',
+            device_type=DeviceType.objects.first(),
```
pheus: Tiny test-readability nit: since this test is specifically checking a mounted 1U device in U0, could we use an explicit device type here?

```python
device_type=DeviceType.objects.get(u_height=1),
```

`objects.first()` happens to be a 1U type in the current setup, but making the height explicit would make the test less sensitive to future fixture/order changes.

pheus (changes requested):
Thanks for the follow-up changes! This is much closer.

I’d still like to request changes for two follow-ups:

```text
1. The form change now skips bound submitted string values like `'0'`/`'41'` when restoring the rack-position widget choice.
2. Existing 0U devices that were stored with `position=0` should probably be normalized, and `get_available_units()` should continue excluding 0U device types from rack-space calculations.
```

pheus:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Mind the already raised review points.

netbox/dcim/migrations/0237_allow_rack_starting_unit_zero.py
mburggraf: I'm still not use how there could even be devices with postion U0, when the validators should have prevented this, but there should be no harm in taking care of this case.

netbox/dcim/tests/test_models.py
mburggraf: Good point.
However, using the same logic, I decided to go for `.filter(u_height=1).first()`, so in case there may be more device_types with height 1U in the future, it doesn't fail there.

netbox/dcim/forms/model_forms.py
mburggraf: Implemented this logic, though I decided for a stylistically different approach:
To avoid the value beeing checked twice, I combined both if-clauses to a single one with an or condition, so in case the value is present, the check can short-circuit.

jeremystretch:
Upon further review, we've determined that the proposed change will require the introduction of a new field on the Rack model to indicate the minimum valid unit number, per my comment [here](https://github.com/netbox-community/netbox/issues/22012#issuecomment-4594041320).

I'm going to close out this PR for now, although we'll likely be able to continue the work completed thus far once the FR has been assigned for a minor release. Thanks for your work on this @mburggraf, and sorry again for not calling this out earlier.

mburggraf:
@jeremystretch No problem. I'd suggest however, to at use the part of the PR, which fixes the broken `get_0u_devices()` function, and the corresponding test case.

PR #22296, opened 2026-05-27
#21418 - Replace MPTT wtih PostgreSQL Ltree

arthanson:
### Closes: #21418 

Replaces django-mptt with PostgreSQL ltree for all hierarchical models in NetBox core. Eliminates the read/write/maintenance costs of MPTT's lft/rght/tree_id/level columns and the table-wide lock MPTT takes on every insert/reparent, in favor of a single GiST-indexed path column maintained entirely by database triggers.

## Expected Performance Improvements

The biggest gains are on writes and concurrency, and they scale with tree size - so the speedup is data-dependent rather than a fixed factor. An MPTT insert/move is O(N) in the tree (it renumbers lft/rght across much of the tree, and shifts tree_id across trees on a root insert), whereas ltree only recomputes the affected node and its subtree (≈O(1) for a leaf insert, O(subtree) for a move) and maintains one path column instead of MPTT's four. 

For small/shallow hierarchies the two are roughly on par; for large hierarchies a single write touches one row instead of thousands, so it can be an order of magnitude or more faster. Separately, MPTT's per-model lock serialized every write to the model; ltree locks per tree, so writes to different trees (e.g. components on different devices) run concurrently - bulk imports and automation that previously queued on one lock should see throughput improve roughly in proportion to how many independent trees are written in parallel.

Read performance for ancestor/descendant lookups is comparable (both index-backed), so queries are unaffected. 

## New infrastructure (netbox/netbox/models/ltree.py)

- **LtreeField** - Django field backed by PostgreSQL's ltree type with ancestor / descendant / descendant_or_equal lookups.
- **LtreeModel** - abstract base providing the MPTT-compatible API surface: get_ancestors, get_descendants, get_children, get_root, get_family, get_siblings, get_descendant_count, level, is_root_node, is_leaf_node, is_child_node, move_to, insert_at, plus tree_id/level properties.
- **LtreeManager** / **LtreeQuerySet** - extends RestrictedQuerySet and mirrors TreeManager.add_related_count for cumulative descendant counts (FK, M2M, and the scope_type/scope_id GenericFK pattern).
- **Triggers** - installs per-table BEFORE INSERT OR UPDATE OF parent_id (computes NEW.path) and AFTER UPDATE OF parent_id, path (cascades to descendants) triggers. Optional sort_path text column carries chr(1)-separated ancestor names, preserving MPTT's order_insertion_by=('name',) flatten ordering.

`NestedLtreeGroupModel` is the ltree-backed base for hierarchical group models (Region, Location, Platform, etc.), carrying two trigger-maintained columns: path, an ltree of ancestor PKs for fast ancestor/descendant queries, and sort_path, a chr(9)-joined chain of ancestor names. sort_path exists because ordering by path sorts siblings by PK (insertion order), whereas ORDER BY sort_path reproduces MPTT's order_insertion_by=('name',) - depth-first tree-flatten with siblings in name order (the TAB separator keeps each subtree grouped under its parent), and its collation tracks name's automatically.

## Base model changes (netbox/netbox/models/__init__.py)

- **NestedGroupModelMixin** - shared fields (name, slug, description, comments) and self-as-descendant clean() check.
- **NestedLtreeGroupModel** - new ltree-backed base used by all NetBox core hierarchical models.
- **NestedGroupModel** - kept as deprecated MPTT-backed base for plugin backwards compatibility (will be removed in a future release).

## Models converted

- **dcim**: Region, SiteGroup, Location, DeviceRole, Platform, ModuleBay, InventoryItem, InventoryItemTemplate
- **tenancy**: ContactGroup, TenantGroup
- **wireless**: WirelessLANGroup

ModuleBay, InventoryItem, and InventoryItemTemplate use LtreeModel directly (not the nested-group base) since they aren't NestedGroupModel subclasses.

## Behavioral changes vs MPTT

- **Cycle validation** - assigning a node's own self or a descendant as its parent now raises `ValidationError` both in `clean()` (forms/serializers) and in `save()` (scripts/bulk), mirroring MPTT's save-time `InvalidMove`. `ModuleBay` opts out (its parent is derived from its installed module).
- **Renames** now update `sort_path` immediately and cascade to descendants. MPTT's `order_insertion_by` left descendants stale until a manual rebuild. `LtreeModel.rebuild_sort_paths()` remains for repairing raw writes that bypass the triggers.
- **Concurrency** - MPTT's per-model advisory lock (which serialized every write to the entire table) is replaced by a per-tree lock: each insert/move takes a transaction advisory lock keyed on the tree's root (both roots for a cross-tree move, in sorted order). Different trees proceed in parallel; within a tree, inserts/moves serialize against a concurrent ancestor reparent so no stale path is persisted. Crossing/ancestor-descendant concurrent moves may deadlock (retryable); plain inserts never do.
- **bulk_create** - within a single batch, a child row must appear after its parent (the BEFORE trigger resolves the parent's path by querying the table).

## GraphQL

The MPTT-internal `lft`/`rght`/`tree_id` columns are no longer part of the schema. `level` is preserved as a resolved field (`nlevel(path) - 1`).

## Removed

`MPTTLockedMixin` and the per-model MPTT tree locks - i.e. the MPTT-specific entries (region, sitegroup, location, tenantgroup, contactgroup, wirelesslangroup, inventoryitem, inventoryitemtemplate, platform) in `ADVISORY_LOCK_KEYS`; the delay_mptt_updates() / rebuild() wrappers in the bulk import/edit/rename views; and utilities/templatetags/mptt.py.

django-pglocks and ADVISORY_LOCK_KEYS themselves are retained — they're still used for available-prefix/IP/VLAN/ASN allocation and job scheduling. django-mptt is retained only for historical migrations and the deprecated NestedGroupModel.

## Migrations

Per-app migrations enable the `ltree` extension, add the `path` (and `sort_path` where applicable) columns, backfill existing rows via a recursive CTE, add GiST indexes, and install the triggers. No operator action is required beyond running migrations.

And the inventory-item ordering line (separate, in case you place it under the conversion notes):

## Test Script

Below is a script that adds nested items to ModuleBays, Inventory Items, etc.. to check screens on MPTT and LTree. So run it once on MPTT version, open all the tabs, then run it on LTree version and in new browser window check same tabs. can do command like:

`python netbox/manage.py shell < ./pr_reviews/demo_tree_ordering.py > /tmp/ltree.out 2>&1`

 <details>
  <summary>demo_tree_ordering.py</summary>

  ```python
"""
Create demo data for every tree-ordered model so list ordering can be compared
between the MPTT branch (main) and the ltree branch.

Run identically on either branch, then diff the output:

    python manage.py shell < /Users/ahanson/dev/work/netbox/pr_reviews/demo_tree_ordering.py

Idempotent (removes prior demo rows first); data is committed so it shows in the UI.

All nodes/bays/items are created in a deliberately scrambled order (e.g. Slot 2, 11,
1, 10 — not pre-sorted), so a correctly sorted listing proves the ordering comes from
the tree (sort_path / MPTT), not from insertion order.

WHAT EACH SECTION EXERCISES
---------------------------
Group models — single hierarchy each, run for ALL the nested-group models the PR
converts: Region, SiteGroup, DeviceRole, Platform (dcim) and TenantGroup,
ContactGroup, WirelessLANGroup (tenancy/wireless). Expected to be PARITY: main's
MPTT TreeManager emits tree-flatten (tree_id, lft) with siblings name-sorted via
order_insertion_by, and ltree reproduces it via sort_path. The alpha nodes plus a
numbered block (Demo 1/2/10/11, with 'Demo 1' having a child) check two edge cases:

  * natural-sort collation — TenantGroup, WirelessLANGroup and ModuleBay use the
    `natural_sort` collation, so their numbered roots must order 1, 2, 10, 11. The
    plain-collation models (Region/SiteGroup/DeviceRole/Platform/ContactGroup/
    Location) order them lexically 1, 10, 11, 2. Each model should look the SAME on
    both branches; the contrast is between models, not between branches.
  * prefix / separator edge — 'Demo 1' has a child, and 'Demo 11' is a sibling root.
    sort_path joins ancestor names with chr(9) (TAB); because TAB sorts below any
    printable char, 'Demo 1's subtree must appear right after 'Demo 1' and BEFORE
    'Demo 11' on both branches. If the separator were wrong the child would
    interleave past 'Demo 11'.

Location — TWO sites; root locations are named so location-name order is the
REVERSE of site-creation order. PARITY: the global list is alphabetical on both.
    main  -> ORDER BY tree_id, lft  : root locations by NAME across sites
             -> Site B's "Demo Alpha" tree before Site A's "Demo Zulu".
    ltree -> ORDER BY sort_path, pk : sort_path begins with the root name, so also
             alphabetical across sites -> "Demo Alpha" before "Demo Zulu".

ModuleBay — TWO devices, nested bays. PARITY (natural sort + nesting + prefix edge,
and the cross-device global ordering):
    * Device A (created first) gets numbered top-level bays Slot 1/2/10/11, and a
      module installed in 'Slot 1' brings nested bays Sub 1/2/10 (children of Slot 1).
    * Device B (created second) gets one bay 'Slot 0', whose name sorts BEFORE every
      bay on device A.
    main  -> ModuleBayManager ORDER BY root-name, lft.
    ltree -> ORDER BY sort_path, pk. sort_path begins with the tree's root-bay name
             (natural_sort collation), so this reproduces main: the global list groups
             by root-bay name ACROSS devices ('Slot 0' from every device first, then
             'Slot 1' + its Sub children, ...), not device-by-device.

InventoryItem — nested items on one device. InventoryItem has NO sort_path and is
intentionally NOT tree-ordered under ltree:
    main  -> TreeManager forces tree-flatten (tree_id, lft) globally.
    ltree -> Meta.ordering ('device', 'parent', 'name') — a flat sort (roots, whose
             parent is NULL, sort LAST). The indented device Inventory tab is kept
             correct on ltree via get_children().order_by('path').
"""
from django.db import transaction
from django.utils.text import slugify

from dcim.models import (
    Device,
    DeviceRole,
    DeviceType,
    InventoryItem,
    Location,
    Manufacturer,
    Module,
    ModuleBay,
    ModuleBayTemplate,
    ModuleType,
    Platform,
    Region,
    Site,
    SiteGroup,
)
from tenancy.models import ContactGroup, TenantGroup
from wireless.models import WirelessLANGroup

# ---------------------------------------------------------------------------
# Group-model hierarchy (parent precedes children). Alpha roots + a numbered
# block to exercise natural-sort collation and the prefix/separator edge.
# ---------------------------------------------------------------------------
# Nodes are created deliberately OUT OF sorted order so the listing proves the
# ordering comes from name/sort_path, not insertion order. Expected listed order:
#   natural-sort models: Demo 1, [Demo 1 Child], Demo 2, Demo 10, Demo 11, Demo Mike, Demo Zulu
#   plain-collation:     Demo 1, [Demo 1 Child], Demo 10, Demo 11, Demo 2, Demo Mike, Demo Zulu
NODES = [
    # Roots, created Zulu→Mike and 10→2→11→1 (neither in sorted order).
    ('Demo Zulu', None),
    ('Demo Mike', None),
    ('Demo 10', None),
    ('Demo 2', None),
    ('Demo 11', None),
    ('Demo 1', None),
    # Children, also created out of order within a parent (Yankee before Alpha).
    ('Demo Yankee', 'Demo Zulu'),
    ('Demo Alpha', 'Demo Zulu'),
    ('Demo Sierra', 'Demo Alpha'),
    ('Demo Bravo', 'Demo Mike'),
    ('Demo 1 Child', 'Demo 1'),   # must sort right after 'Demo 1' and before 'Demo 11'
]
DEMO_SLUGS = [slugify(name) for name, _ in NODES]

# Group models that own a self-nesting hierarchy with name/slug/parent.
GROUP_MODELS = [
    ('Region', Region),
    ('SiteGroup', SiteGroup),
    ('DeviceRole', DeviceRole),
    ('Platform', Platform),
    ('TenantGroup', TenantGroup),       # natural_sort name
    ('ContactGroup', ContactGroup),
    ('WirelessLANGroup', WirelessLANGroup),  # natural_sort name
]

# ---------------------------------------------------------------------------
# Location: two sites (Site A created first -> lower site_id).
# ---------------------------------------------------------------------------
DEMO_SITE_SLUGS = ['demo-site-a', 'demo-site-b']
LOCATION_SITES = [
    ('Demo Site A', 'demo-site-a', [
        ('Demo Zulu', None),         # sorts LAST by name, but this site is created FIRST
        ('Demo Yankee', 'Demo Zulu'),
    ]),
    ('Demo Site B', 'demo-site-b', [
        ('Demo Alpha', None),        # sorts FIRST by name, but this site is created SECOND
        ('Demo Bravo', 'Demo Alpha'),
    ]),
]

# ---------------------------------------------------------------------------
# Devices for ModuleBay + InventoryItem demos.
# ---------------------------------------------------------------------------
DEV_A = 'Demo ModBay A'           # created first  -> lower device_id
DEV_B = 'Demo ModBay B'           # created second -> higher device_id
DEV_INV = 'Demo Inventory Device'
DEVICE_NAMES = [DEV_A, DEV_B, DEV_INV]

# Created out of order; both must LIST in natural order (Slot 1, 2, 10, 11 / Sub 1, 2, 10).
DEV_A_BAYS = ['Slot 2', 'Slot 11', 'Slot 1', 'Slot 10']  # nested under 'Slot 1' below
MODULE_SUBBAYS = ['Sub 10', 'Sub 1', 'Sub 2']            # nested under 'Slot 1' via a module
DEV_B_BAYS = ['Slot 0']                                  # sorts before every Device A bay

INV_NODES = [
    ('Item Zulu', None),
    ('Item Mike', None),
    ('Item Yankee', 'Item Zulu'),   # created before 'Item Alpha' (tests sibling order)
    ('Item Alpha', 'Item Zulu'),
    ('Item Bravo', 'Item Mike'),
]


def make_tree(model, **extra):
    created = {}
    for name, parent_name in NODES:
        obj = model(
            name=name,
            slug=slugify(name),
            parent=created.get(parent_name) if parent_name else None,
            **extra,
        )
        obj.save()
        created[name] = obj


def show_indented(label, queryset, suffix=lambda obj: ''):
    print(f"\n=== {label} ===")
    for obj in queryset:
        print(f"  {'    ' * obj.level}{obj.name}{suffix(obj)}")


with transaction.atomic():
    # --- Clean up prior demo rows (devices first; cascades bays/modules/items) ---
    Device.objects.filter(name__in=DEVICE_NAMES).delete()
    DeviceType.objects.filter(slug='demo-device-type').delete()
    ModuleType.objects.filter(model='Demo Module Type').delete()
    Manufacturer.objects.filter(slug='demo-mfr').delete()
    DeviceRole.objects.filter(slug='demo-equipment-role').delete()
    Location.objects.filter(site__slug__in=DEMO_SITE_SLUGS + ['demo-device-site']).delete()
    Site.objects.filter(slug__in=DEMO_SITE_SLUGS + ['demo-device-site']).delete()
    for _, model in GROUP_MODELS:
        model.objects.filter(slug__in=DEMO_SLUGS).delete()

    # --- Group models: single hierarchy each ---
    for _, model in GROUP_MODELS:
        make_tree(model)

    # --- Location: two sites ---
    for site_name, site_slug, nodes in LOCATION_SITES:
        site = Site.objects.create(name=site_name, slug=site_slug)
        created = {}
        for name, parent_name in nodes:
            obj = Location(
                name=name,
                slug=slugify(f'{site_slug}-{name}'),
                site=site,
                parent=created.get(parent_name) if parent_name else None,
            )
            obj.save()
            created[name] = obj

    # --- Shared scaffolding for the device demos ---
    mfr = Manufacturer.objects.create(name='Demo Mfr', slug='demo-mfr')
    dtype = DeviceType.objects.create(manufacturer=mfr, model='Demo Device Type', slug='demo-device-type')
    erole = DeviceRole.objects.create(name='Demo Equipment Role', slug='demo-equipment-role')
    dsite = Site.objects.create(name='Demo Device Site', slug='demo-device-site')

    # ModuleType carrying nested module-bay templates (instantiated on install).
    mtype = ModuleType.objects.create(manufacturer=mfr, model='Demo Module Type')
    for n in MODULE_SUBBAYS:
        ModuleBayTemplate.objects.create(module_type=mtype, name=n)

    # Device A (first): numbered top-level bays; install a module in 'Slot 1' to
    # create the nested Sub bays beneath it.
    dev_a = Device.objects.create(name=DEV_A, device_type=dtype, role=erole, site=dsite, status='active')
    a_bays = {}
    for n in DEV_A_BAYS:
        bay = ModuleBay(device=dev_a, name=n)
        bay.save()
        a_bays[n] = bay
    Module.objects.create(device=dev_a, module_type=mtype, module_bay=a_bays['Slot 1'], status='active')

    # Device B (second): one bay that name-sorts before every Device A bay.
    dev_b = Device.objects.create(name=DEV_B, device_type=dtype, role=erole, site=dsite, status='active')
    for n in DEV_B_BAYS:
        ModuleBay(device=dev_b, name=n).save()

    # InventoryItem device.
    dev_inv = Device.objects.create(name=DEV_INV, device_type=dtype, role=erole, site=dsite, status='active')
    inv = {}
    for name, parent_name in INV_NODES:
        item = InventoryItem(device=dev_inv, name=name, parent=inv.get(parent_name) if parent_name else None)
        item.save()
        inv[name] = item


# --- Verification (mirrors the UI's default ordering) ---------------------
NATURAL = {'TenantGroup', 'WirelessLANGroup'}
for label, model in GROUP_MODELS:
    collation = 'natural_sort name' if label in NATURAL else 'plain name'
    show_indented(f'{label}: default list ordering [{collation}]', model.objects.filter(slug__in=DEMO_SLUGS))

show_indented(
    'Location: default list ordering (parity: alphabetical across sites on both)',
    Location.objects.filter(site__slug__in=DEMO_SITE_SLUGS),
    suffix=lambda obj: f'  [{obj.site.name}]',
)

show_indented(
    'ModuleBay: GLOBAL list ordering across devices '
    "(parity: grouped by root-bay name across devices — 'Slot 0' first, then 'Slot 1' ...)",
    ModuleBay.objects.filter(device__name__in=[DEV_A, DEV_B]),
    suffix=lambda obj: f'  [{obj.device.name}]',
)
for dev in (DEV_A, DEV_B):
    show_indented(f'ModuleBay: device tab for {dev}', ModuleBay.objects.filter(device__name=dev))

# InventoryItem: default order differs (tree-flatten on main, flat parent/name on
# ltree); the device tab is path-ordered on ltree, tree-ordered by default on main.
show_indented(
    'InventoryItem: default list ordering (tree-flatten on main; parent/name on ltree)',
    InventoryItem.objects.filter(device__name=DEV_INV),
)
inv_qs = InventoryItem.objects.filter(device__name=DEV_INV)
if any(f.name == 'path' for f in InventoryItem._meta.get_fields()):
    inv_qs = inv_qs.order_by('path')   # ltree device-tab ordering
show_indented('InventoryItem: device tab ordering (indented hierarchy)', inv_qs)

print(f"\nDevices created: {DEV_A!r}, {DEV_B!r}, {DEV_INV!r}")
print('Done.')

  ````python


arthanson:

arthanson:

arthanson:

arthanson:
Claide review comments addressed.

netbox/dcim/models/sites.py
```diff
@@ -323,10 +328,15 @@ class Location(ContactsMixin, ImageAttachmentsMixin, NestedGroupModel):
...
     class Meta:
-        ordering = ['site', 'name']
-        # Empty tuple triggers Django migration detection for MPTT indexes
-        # (see #21016, django-mptt/django-mptt#682)
-        indexes = ()
+        # Alphabetical tree-flatten in the global list: sort_path begins with the
+        # root location's name, so ORDER BY sort_path lists root locations by name
+        # across sites (descendants following their root). `pk` is a deterministic
+        # tie-break for same-named roots in different sites.
+        ordering = ('sort_path', 'pk')
```
jeremystretch: Should we instead order by `('site', 'sort_path')`?

netbox/dcim/models/device_components.py
```diff
@@ -1313,34 +1313,11 @@ def clean(self):
...
-        return super().get_queryset(*args, **kwargs).annotate(
-            _root_name=models.Subquery(
-                root_name,
-                output_field=models.CharField(db_collation='natural_sort'),
-            )
-        ).order_by('_root_name', 'lft')
-
-
-class ModuleBay(ModularComponentModel, TrackingModelMixin, MPTTModel):
+class ModuleBay(ModularComponentModel, TrackingModelMixin, LtreeModel):
```
jeremystretch, line 1316: Why does ModuleBay inherit from LtreeModel instead of NestedLtreeGroupModel?

netbox/dcim/models/device_components.py
```diff
@@ -1620,13 +1585,7 @@ class Meta:
...
 
     def clean(self):
-        super().clean()
-
-        # An InventoryItem cannot be its own parent
-        if self.pk and self.parent_id == self.pk:
-            raise ValidationError({
-                "parent": _("Cannot assign self as parent.")
-            })
+        super().clean()  # LtreeModel.clean() rejects self/descendant-as-parent cycles
```
jeremystretch: Suggest removing this comment as it might be confusing in the future.
```suggestion
        super().clean()
```

netbox/dcim/models/device_component_templates.py
```diff
@@ -877,6 +885,8 @@ class Meta:
         verbose_name = _('inventory item template')
         verbose_name_plural = _('inventory item templates')
 
+    # Self/descendant-as-parent cycles are rejected by the inherited LtreeModel.clean().
+
```
jeremystretch: Can probably remove this.

base_requirements.txt
```diff
@@ -26,7 +26,9 @@ django-graphiql-debug-toolbar
 # https://django-htmx.readthedocs.io/en/latest/changelog.html
 django-htmx
 
-# Modified Preorder Tree Traversal (recursive nesting of objects)
+# Modified Preorder Tree Traversal (required only for historical migrations
```
jeremystretch: We should probably call out backward compatibility for plugins as the primary motivation for keeping MPTT around.

netbox/wireless/migrations/0020_ltree_paths.py
```diff
@@ -0,0 +1,108 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='wirelesslangroup', name='parent',
+            field=models.ForeignKey(
+                blank=True, null=True, on_delete=django.db.models.deletion.CASCADE,
+                related_name='children', to='wireless.wirelesslangroup',
+            ),
+        ),
+
+        CreateExtension('ltree'),
```
jeremystretch: Suggest moving `CreateExtension()` to the beginning of `operations` within each migration file to ensure fast failure should the extension be missing.

netbox/wireless/migrations/0020_ltree_paths.py
```diff
@@ -0,0 +1,108 @@
...
+        migrations.AddField(
+            model_name=MODEL, name='sort_path',
+            field=models.TextField(
+                blank=True, default='', editable=False, db_collation='natural_sort',
+            ),
+        ),
+
+        InstallLtreeTriggers(TABLE, name_column='name'),
+
+        migrations.RunSQL(
```
jeremystretch, line 54: We should ensure that each call to `RunSQL()` includes an explanatory comment.

netbox/dcim/migrations/0238_ltree_paths.py
```diff
@@ -0,0 +1,339 @@
...
+    'dcim_devicerole',
+    'dcim_platform',
+    'dcim_inventoryitem',
+    'dcim_inventoryitemtemplate',
+    'dcim_modulebay',
+)
+
+# Subset that previously declared `MPTTMeta.order_insertion_by = ('name',)` and
+# therefore needs a `sort_path` text column maintained alongside `path`.
+SORT_MODELS = ('region', 'sitegroup', 'location', 'devicerole', 'platform', 'modulebay')
```
jeremystretch, line 63: We should discuss what would be necessary to standardize the InventoryItem and InventoryItemTemplate models as well. Or, acknowledging that they're due to be removed in v5.0, we could opt to just leave them as MPTT for now (given that we need to retain MPTT for some time anyway).

netbox/dcim/migrations/0238_ltree_paths.py
```diff
@@ -0,0 +1,339 @@
...
+    'dcim_location',
+    'dcim_devicerole',
+    'dcim_platform',
+    'dcim_modulebay',
+)
+
+LEGACY_FIELDS = ('lft', 'rght', 'tree_id', 'level')
+
+
+def _populate_paths_sql():
```
jeremystretch: While we normally try to keep migrations as self-contained as possible, I think there would be a benefit to consolidating the logic within _populate_paths_sql() and _assert_paths_populated_sql() as standalone utility functions. This would both reduce redundant code and ensure that any potential future fixes get applied consistently. We'll also need an answer for plugin maintainers who want to migrate from MPTT to Ltree.

IMO it would be reasonably safe to contain these in a purpose-built module (e.g. utilities.mptt_to_ltree).

netbox/netbox/api/viewsets/__init__.py
```diff
@@ -335,22 +333,3 @@ def perform_destroy(self, instance):
                 super().perform_destroy(instance)
         except ObjectDoesNotExist:
             raise PermissionDenied()
-
-
-class MPTTLockedMixin:
```
jeremystretch, line 340: We probably need to keep this around for now to avoid plugin breakage.

netbox/netbox/tables/columns.py
```diff
@@ -628,6 +629,10 @@ def value(self, value):
         return value
 
 
+# Deprecated alias for plugin compatibility; use TreeColumn going forward.
```
jeremystretch, line 632:
```suggestion
# Deprecated alias for plugin compatibility; use TreeColumn going forward.
# TODO: Remove this in NetBox v5.0
```

netbox/netbox/views/generic/bulk_views.py
jeremystretch, line 1: I think we still need to retain MPTT support in all generic views to avoid plugin breakage.

netbox/utilities/query.py
```diff
@@ -64,8 +62,13 @@ def reapply_model_ordering(queryset: QuerySet) -> QuerySet:
     Reapply model-level ordering in case it has been lost through .annotate().
     https://code.djangoproject.com/ticket/32811
     """
-    # MPTT-based models are exempt from this; use caution when annotating querysets of these models
-    if any(isinstance(manager, TreeManager) for manager in queryset.model._meta.local_managers):
```
jeremystretch, line 68: We probably need to keep this in place too, for any plugins that still use MPTT.

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+    def db_type(self, connection):
+        return 'ltree'
+
+    def get_prep_value(self, value):
+        if value is None:
+            return value
+        return str(value)
+
+
+@LtreeField.register_lookup
```
jeremystretch: Maybe move the lookup classes to a separate `lookups.py` module for consistency with other apps.

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+    lookup_name = 'descendant'
+
+    def as_sql(self, compiler, connection):
+        lhs, lhs_params = self.process_lhs(compiler, connection)
+        rhs, rhs_params = self.process_rhs(compiler, connection)
+        return f'({lhs} <@ {rhs} AND {lhs} <> {rhs})', lhs_params + rhs_params + lhs_params + rhs_params
+
+
+@LtreeField.register_lookup
+class DescendantOrEqual(Lookup):
```
jeremystretch: Is there any scenario where we might want an `AncestorOrEqual` lookup?

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+        ).exists()
+
+    @classmethod
+    def _has_sort_path(cls):
+        """
+        Whether this model carries the optional trigger-maintained `sort_path`
+        column (the MPTT `order_insertion_by=('name',)` equivalent). Single source
+        of truth for clean(), save(), and _tree_order_field().
+        """
+        return any(f.attname == 'sort_path' for f in cls._meta.concrete_fields)
```
jeremystretch: `self.get_field()` feels more canonical here, but is there a reason we avoid it?

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+        name_changed = (
+            (not is_insert) and has_sort_path and name_written
+            and self.__dict__.get('name') != self._loaded_name
+        )
+
+        # Reject cyclic moves before writing, mirroring django-mptt's save-time
+        # guard so scripts / bulk callers (which bypass form & serializer clean())
+        # cannot silently corrupt the tree.
+        if parent_changed and self._parent_creates_cycle():
+            raise ValidationError(_("Cannot assign self or a descendant as parent."))
```
jeremystretch: This feels like it should be moved up above line 407, where `parent_changed` is determined.

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+
+        if is_insert or parent_written:
+            self._loaded_parent_id = self.parent_id
+        if is_insert or name_written:
+            self._loaded_name = self.__dict__.get('name')
+
+    # -- MPTT-compatible API ------------------------------------------------
+
+    @property
+    def level(self):
```
jeremystretch, line 465: I'm curious whether we could store in this the database and update it automatically alongside `path`/`sort_path`.

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+    def _root_pk(self):
+        """
+        Integer PK of the root node, parsed from the first (zero-padded) path label.
+        Returns None for a node with no path.
+        """
+        if not self.path:
+            return None
+        return int(str(self.path).split('.', 1)[0].lstrip('0') or '0')
+
+    @property
```
jeremystretch: Seems like it would be safe to cache this?
```suggestion
    @cached_property
```

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+            return self
+        if not self.path:
+            # Unsaved (no path computed yet). Walk up the in-memory parent
+            # chain if available; otherwise we have no way to resolve the root.
+            parent = getattr(self, 'parent', None)
+            return parent.get_root() if parent is not None else None
+        return type(self)._default_manager.get(pk=self._root_pk())
+
+    def get_parent(self):
+        return self.parent
```
jeremystretch: Is this for MPTT compatability? Doesn't seem to be a documented method.

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
+"""
+Ltree-based hierarchical model support - drop-in replacement for django-mptt.
+
+LtreeModel provides the same public API as django-mptt's MPTTModel (get_ancestors,
```
jeremystretch: We should caveat this to clarify that we don't _fully_ replicate MPTT's API (nor do we care to); e.g. we don't provide `get_leafnodes()`, `get_next_sibling()`, etc.

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+            Q(path__ancestor=self.path) | Q(path__descendant_or_equal=self.path)
+        ).order_by(self._tree_order_field())
+
+    def get_siblings(self, include_self=False):
+        qs = type(self)._default_manager.filter(parent_id=self.parent_id)
+        if not include_self:
+            qs = qs.exclude(pk=self.pk)
+        return qs.order_by(self._tree_order_field())
+
+    def move_to(self, target, position='last-child'):
```
jeremystretch: I'm not sure how much we care about replicating MPTT's API. IMO we can do without `move_to()` and `insert_at()`, but open to keeping them.

netbox/netbox/models/ltree.py
```diff
@@ -0,0 +1,901 @@
...
+                FROM {table} r INNER JOIN t ON r.parent_id = t.id
+            )
+            UPDATE {table} SET sort_path = t.sort_path FROM t WHERE {table}.id = t.id;
+        '''
+        with connection.cursor() as cursor:
+            cursor.execute(sql)
+
+
+#
+# Migration operation
```
jeremystretch: Let's move these into a different module, e.g. `utilities.ltree`.

jeremystretch (commented):
Amazing work! I haven't dug into the triggers or reviewed any of the new tests yet, but wanted to push through some general feedback/questions. A couple general thoughts:

- I don't know how closely we care to mimic the MPTT API. IMO we should provide only methods/properties that have a solid use case. (We can always extend these later, but will need to maintain indefinitely whatever we initially expose.)
- We'll need to do a ton of testing at scale to catch performance issues, especially concerning the initial migrations.
- We should probably break up the `netbox.models.ltree` module as it's quite large.
- This is the most optimistic I've ever felt about finally ditching MPTT. 🙂

netbox/dcim/models/device_components.py
arthanson: LtreeModel is the equivalent of MPTTModel, ModuleBay never inherited from NestedGroupModel?

netbox/netbox/models/ltree.py
arthanson: This isn't safe - path is rewritten on reparent and a cached value wouldn't be invalidated

arthanson:
@jeremystretch removed un-used MPTT methods, cleaned up comments and refactored.  Please note two comments on your suggestions above.

netbox/dcim/migrations/0238_ltree_paths.py
arthanson: Discussion resolution was to move them to LTree as the work is already done

netbox/netbox/models/ltree.py
arthanson: added

netbox/netbox/models/ltree.py
arthanson: I'd push back on this. it duplicates information path already encodes and the triggers would have to compute and cascade level on every insert/move/reparent which can be expensive and would be redundant.

arthanson:
@jeremystretch other review comments addressed.

jeremystretch (commented):
This all looks clean AFAICT. However, Claude did raise a performance concern during review, which I'll share below. The proposed modification to `_LOCK_TREE_ROOTS_SQL` seems reasonable, but I wanted to get your take @arthanson.

---

## Advisory-lock pool exhaustion on large single-transaction bulk inserts (worth confirming)

The `BEFORE` trigger takes a `pg_advisory_xact_lock` keyed per-tree, held until commit. For new root nodes, dest_root is the row's own label, so each new root is a distinct key. A single atomic transaction that bulk-creates many roots (e.g. a large CSV import of Regions/TenantGroups, or a script inside `transaction.atomic()`) accumulates one advisory lock per root, all held simultaneously, drawing from the shared lock table (`max_locks_per_transaction` × (`max_connections` + `…`)).

MPTT took a single per-model lock; this is strictly more locks for the wide-root case, and at sufficient volume can fail with out of shared memory / You might need to increase `max_locks_per_transaction`. The per-tree concurrency win is real and worth it — but has this been load-tested with a large single-transaction import of many roots? If it's a known limit, a note in the docs/release notes would help operators.

### Primary fix: don't lock on root inserts

The exhaustion comes from one fact: a brand-new root node takes an advisory lock keyed on its own label, held until commit — but that lock can never contend with 
anything. A root being INSERTed with parent_id IS NULL starts its own singleton tree that no concurrent transaction can see or reference yet (MVCC — the uncommitted
row isn't visible, and nobody else has its PK). So:

- It can't be reparented concurrently (doesn't exist for other txns).
- Its subtree is empty.
- No other transaction's AFTER cascade can target it (it's in nobody's tree).

The lock the trigger's own comment justifies — "serialize an insert against a concurrent reparent of one of its ancestors" — is meaningless for a node that has no
ancestor. So the lock on a new-root key is pure overhead and the entire pool-consumption vector, since each root is a distinct key.

The change is a guard at the top of `_LOCK_TREE_ROOTS_SQL`:

```sql
-- A brand-new root (INSERT, no parent) starts its own singleton tree that no
-- concurrent transaction can yet see or reference, so there is nothing to
-- serialize against. Skipping the lock here is what keeps a bulk import of many
-- top-level objects from accumulating one xact-lock per root and exhausting the
-- shared lock table (max_locks_per_transaction).
IF TG_OP = 'INSERT' AND NEW.parent_id IS NULL THEN
  -- no lock required
ELSE
  -- ... existing dest_root / old_root / pg_advisory_xact_lock logic ...
END IF;
```

This is safe and narrow:
- A reparent-to-root (`UPDATE … parent_id = NULL`) still locks — the node exists, has descendants, and the cascade rewrites a real subtree.
- A child insert into an existing tree still locks.
- Only the provably race-free case is exempted.

Critically, this kills the exact scenario I flagged: importing N top-level objects (Regions, TenantGroups, or one inventory item per device — InventoryItem trees are
rooted at top-level items, so each is its own root) now takes zero advisory locks instead of N. The dominant bulk pattern becomes lock-free.

#### Residual case + operator fallback

The primary fix doesn't cover bulk-inserting children into many distinct existing trees in one transaction (e.g. one child under each of 50k existing trees) — that
still accumulates one lock per distinct tree. This is a much rarer pattern, and for it the right answer is operator guidance:

- Document that very large hierarchical bulk writes scale with the number of distinct existing trees touched per transaction, and that `max_locks_per_transaction` is
the tuning knob if out of shared memory appears. This is the standard Postgres remedy and a one-line config bump.

I'd put that in the migration/upgrade notes alongside the existing operator warning.

### What I'd avoid

- Releasing the lock mid-transaction (session-level `pg_advisory_lock` + early unlock) is not safe: the correctness argument requires holding until commit, or a
concurrent reparent could interleave between the path computation and commit and leave a stale path this txn can no longer fix. Don't go there.
- Per-batch commits in bulk import would release locks between batches but breaks NetBox's all-or-nothing atomic import semantics — not worth it.
- Coarse whole-table lock for bulk defeats the per-tree concurrency that's the whole point of the PR.

### Verification

Add a regression test that pins the behavior: inside a transaction, create a root and assert no advisory locks are held —

```python
with transaction.atomic():
  Region.objects.create(name='Root', slug='root')
  with connection.cursor() as cur:
      cur.execute("SELECT count(*) FROM pg_locks WHERE locktype = 'advisory'")
      self.assertEqual(cur.fetchone()[0], 0)
```

and a companion asserting a child insert (or reparent-to-root) does hold exactly one, so the optimization can't silently over-broaden later.

arthanson:
Updated the lock code.

jeremystretch (approved):

PR #22304, opened 2026-05-27
Closes #21945: Introduce a test for duplicate migration indexes

### LINKED ISSUE #21945: Test for duplicate migration indexes
### Proposed Changes

Add a simple CI test that checks all apps for duplicate migration indexes, e.g.

```
0123_foo_add_fields.py
0123_bar_remoe_fields.py
```

### Justification

While technically permissible, duplicate indexes tend toward confusion and are thus best avoided. This has always been our (and Django's) convention, but we currently have no means of enforcement.

jeremystretch:
### Closes: #21945

Introduce DuplicateMigrationIndexTestCase, which fails if any migrations within an app (beside those exempt) contain the same numeric index.

### DIFF
--- a/netbox/netbox/tests/test_migrations.py
+++ b/netbox/netbox/tests/test_migrations.py
@@ -1,0 +1,89 @@
+import re
+from collections import defaultdict
+
+from django.apps import apps
+from django.conf import settings
+from django.db.migrations.loader import MigrationLoader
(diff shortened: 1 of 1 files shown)


jeremystretch:

jeremystretch:
Addressed the feedback above.

bctiemann (approved):

PR #22306, opened 2026-05-27
Closes #22305: Allow test cases to declare a stable query-count key prefix

### LINKED ISSUE #22305: Allow test cases to declare a stable query-count key prefix
### Proposed Changes

Add an optional `query_count_model_label` class attribute to the `assert_expected_query_count` contract. When present on the test case, its value is used as the key prefix in `query_counts.json` instead of `model._meta.model_name`.

Current behaviour in `netbox/utilities/testing/query_counts.py`:

```python
model = test_case.model
app_label = model._meta.app_label
model_name = model._meta.model_name
key = f'{model_name}:{name}'
```

Proposed behaviour:

```python
model = test_case.model
app_label = model._meta.app_label
model_name = getattr(test_case, 'query_count_model_label', None) or model._meta.model_name
key = f'{model_name}:{name}'
```

No existing test cases are affected — the attribute is optional and the fallback is identical to today's behaviour. Plugin or third-party test suites that need a stable label can set the attribute on their test class:

```python
class MyDynamicModelViewTestCase(ViewTestCases.PrimaryObjectViewTestCase):
    query_count_model_label = 'my-stable-label'
    ...
```

The docstring for `assert_expected_query_count` should be updated to document the attribute.

### Justification

`model._meta.model_name` is stable for every built-in NetBox model because the class name is a compile-time constant. However, plugins that generate Django models at runtime (e.g. from database records) derive `model_name` from a runtime value such as a database primary-key sequence number. A typical key looks like `table291model:list_objects_with_permission`.

This causes two practical problems:

**1. Baseline keys drift when unrelated tests are added.**
PostgreSQL sequences are not transactional. Any test method (in any module that alphabetically precedes `test_views.py`) that creates and then rolls back a row in the dynamic-model table permanently increments the sequence counter. Each such test shifts every subsequent dynamic-model ID by 1, invalidating every key in `query_counts.json` for that plugin. Adding a single new model-level test can require regenerating the baseline file even though no view code changed.

**2. Keys are environment-dependent.**
The sequence state when the view tests run depends on how many rows were inserted earlier in the same test run. A baseline generated on a developer's `--keepdb` database will have different keys from a baseline generated against CI's fresh database — the same code produces different key names in different environments.

Both problems disappear when the key is based on a stable, human-assigned label (such as a slug) rather than a runtime sequence number. The one-line change to `assert_expected_query_count` gives plugin authors an escape hatch without touching any existing NetBox test infrastructure.

Source references:
- `netbox/utilities/testing/query_counts.py:95-98` — the lines that derive the key
- `netbox/utilities/testing/views.py:474` — the call site that passes `test_case` to `assert_expected_query_count`

bctiemann:
### Closes: #22305

Add an optional `query_count_model_label` class attribute to the `assert_expected_query_count` contract. When set on a test case, its value is used as the baseline key prefix in `query_counts.json` instead of `model._meta.model_name`.

The change is a single line in `query_counts.py` plus an expanded docstring. No existing test cases are affected — the attribute is optional and the fallback is identical to the current behaviour.

**Motivation**

`model._meta.model_name` is stable for all built-in NetBox models because their class names are compile-time constants. Plugins that generate Django models at runtime (for example, from database records) derive `model_name` from a runtime value such as a PostgreSQL primary-key sequence number, producing keys like `table291model:list_objects_with_permission`.

PostgreSQL sequences are not transactional: a rolled-back test that creates a row still permanently increments the counter. Any new test method in a module that runs before `test_views.py` will shift every subsequent dynamic-model ID by 1, invalidating the entire `query_counts.json` baseline for that plugin — even when no view code changed.

With this change, an affected plugin can declare a stable label on its test class and the problem goes away entirely:

```python
class ComplexObjectViewTestCase(ViewTestCases.PrimaryObjectViewTestCase):
    query_count_model_label = 'complex-objects'   # COT slug — never shifts
    ...
```

### DIFF
--- a/netbox/utilities/testing/query_counts.py
+++ b/netbox/utilities/testing/query_counts.py
@@ -101,7 +101,8 @@ def assert_expected_query_count(test_case, name):
     """
     model = test_case.model
     app_label = model._meta.app_label
-    model_name = getattr(test_case, 'query_count_model_label', None) or model._meta.model_name
+    label = getattr(test_case, 'query_count_model_label', None)
+    model_name = label if label is not None else model._meta.model_name
     key = f'{model_name}:{name}'
 
     if _is_update_mode():
--- a/netbox/utilities/testing/query_counts.py
+++ b/netbox/utilities/testing/query_counts.py
@@ -86,7 +86,14 @@ def assert_expected_query_count(test_case, name):
     Assert that the wrapped block performs the number of SQL queries recorded
     in the per-app baseline file (`<app>/tests/query_counts.json`).
 
-    The baseline key is `<model_name>:<name>`, derived from `test_case.model`.
+    The baseline key is `<model_label>:<name>`. By default `<model_label>` is
+    derived from `test_case.model._meta.model_name`. Test cases that use
+    runtime-generated models with unstable names (e.g. names derived from a
+    database primary-key sequence) can declare a ``query_count_model_label``
+    class attribute to provide a stable, human-assigned label instead::
+
+        class MyViewTestCase(ViewTestCases.PrimaryObjectViewTestCase):
+            query_count_model_label = 'my-stable-label'
 
     When the `UPDATE_QUERY_COUNTS` environment variable is set, the assertion
     is skipped and the observed count is written back to the baseline file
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review and limit response only to areas of potential concern.

bctiemann:
Items 1 and 2 are addressed.

jnovinger (approved):

PR #22308, opened 2026-05-27
Fixes #22307: Ensure consistent treatment of grant_token in web UI

### LINKED ISSUE #22307: `grant_token` enforcement in UI is inconsistent with REST API
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Grant a non-superuser the `add_token` permission. Ensure the user does not have the `grant_token` permission.
2. As the user, attempt to create a token for _another_ user via the REST API.
3. The API returns an error: `This user does not have permission to create tokens for other users.`
4. Again as the user, attempt to create a token for another user via the web UI.

### Expected Behavior

Creating a token for another user via the web UI should be prohibited even with the `add_token` permission assigned if the user does not also possess the `grant_token` permission.

### Observed Behavior

The user is able to create a token for another user via the web UI.

jeremystretch:
### Fixes: #22307

- Extend UserTokenForm and TokenBulkImportView to consistently enforce the `users.grant_token` permission.
- Add tests

### DIFF
--- a/netbox/users/forms/bulk_import.py
+++ b/netbox/users/forms/bulk_import.py
@@ -54,6 +54,14 @@ class TokenImportForm(CSVModelForm):
         model = Token
         fields = ('user', 'version', 'token', 'enabled', 'write_enabled', 'expires', 'description',)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Disable the user field when updating an existing Token, so a Token's owner cannot be reassigned via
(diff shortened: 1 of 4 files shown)


jeremystretch:

jeremystretch:

jeremystretch:

jeremystretch:

jeremystretch:
Addressed point 1 above. Left point 2 as-is.

pheus (approved):
Looks good to me! Thanks!

PR #22315, opened 2026-05-28
Closes #22290: Use cleanup handlers for config test teardown

### LINKED ISSUE #22290: `ConfigTestCase` cleanup can be skipped, leaving stale Redis-backed config state
### Proposed Changes

In `netbox/netbox/tests/test_config.py`, register the config/cache cleanup at the beginning of each `ConfigTestCase` method using `self.addCleanup(...)`:

```python
self.addCleanup(clear_config)
self.addCleanup(cache.clear)
```

Keep the existing initial `cache.clear()` call so each test starts from a clean cache, but also register `cache.clear` as a cleanup so the test leaves the cache clean after it finishes. Once these cleanups are registered, the trailing `clear_config()` calls can be removed.

Current pattern:

```python
@override_settings(CACHES=CACHES)
def test_config_init_from_cache(self):
    CONFIG_DATA = {'BANNER_TOP': 'B'}
    cache.clear()

    # Create a config and load it into the cache
    configrevision = ConfigRevision.objects.create(data=CONFIG_DATA)
    configrevision.activate()

    config = get_config()
    self.assertEqual(config.config, CONFIG_DATA)        # if this raises...
    self.assertEqual(config.version, configrevision.pk)

    clear_config()                                      # ...this never runs
```

Proposed pattern:

```python
@override_settings(CACHES=CACHES)
def test_config_init_from_cache(self):
    self.addCleanup(clear_config)
    self.addCleanup(cache.clear)

    CONFIG_DATA = {'BANNER_TOP': 'B'}
    cache.clear()

    configrevision = ConfigRevision.objects.create(data=CONFIG_DATA)
    configrevision.activate()

    config = get_config()
    self.assertEqual(config.config, CONFIG_DATA)
    self.assertEqual(config.version, configrevision.pk)
```

Apply the same change to all four methods in `ConfigTestCase`:

- `test_config_init_empty`
- `test_config_init_from_db`
- `test_config_init_from_cache`
- `test_settings_override`

### Justification

`ConfigRevision.activate()` writes the revision's `data` blob into the Django cache using the `config` and `config_version` cache keys. The existing trailing `clear_config()` calls only clear NetBox's in-process `_thread_locals.config`; they do not remove the Redis-backed cache entries.

Because the cleanup currently appears as the final statement in each test method, any assertion failure before that point skips the cleanup entirely. The next `ConfigTestCase` method defensively calls `cache.clear()` at its start, but that still leaves a window where stale cache state can leak into later tests or across local test invocations using the same Redis-backed cache.

This can surface as unrelated API failures when the leaked config blob contains a `CUSTOM_VALIDATORS` rule such as:

```python
{
    'ipam.asn': [
        {'tenant': {'required': True}}
    ]
}
```

In that case, subsequent ASN create tests can unexpectedly fail with HTTP 400 instead of HTTP 201:

```python
AssertionError: 400 != 201 : Expected HTTP status 201; received 400:
{'__all__': [ErrorDetail(string="Custom validation failed for tenant: ['This field must not be empty.']", code='invalid')]}
```

Observed failure pattern, reproducible after a prior `ConfigTestCase` assertion failure:

- `ipam.tests.test_api.ASNRangeTestCase.test_create_multiple_available_asns`
- `ipam.tests.test_api.ASNRangeTestCase.test_create_single_available_asn`
- `ipam.tests.test_api.ASNTestCase.test_bulk_create_objects`
- `ipam.tests.test_api.ASNTestCase.test_create_object`

The same failures are not reproducible when running `ipam.tests.test_api` in isolation, which points to leaked config/cache state from an earlier test run. Registering both `clear_config` and `cache.clear` with `addCleanup()` closes this off because the cleanups are executed even if the test fails before reaching the end of the method.

Source references:

- `netbox/netbox/tests/test_config.py:15-67` — the four affected tests
- `netbox/netbox/config/__init__.py:34-40` — `clear_config()` only clears the in-process thread-local config
- `netbox/core/models/config.py` — `ConfigRevision.activate()` writes `config` and `config_version` into the cache
- `netbox/extras/validators.py:33-60,119` — origin of the `"Custom validation failed for tenant"` error format

pheus:
### Closes: #22290

This updates `ConfigTestCase` to register `clear_config()` and `cache.clear()` with `addCleanup()` from `setUp()`, so the cleanup behavior is applied consistently across the whole test case class.

This avoids relying on manual cleanup calls at the end of individual tests and ensures cleanup still runs if a test fails early, helping prevent stale Redis-backed config state from leaking into later tests.

### DIFF
--- a/netbox/netbox/tests/test_config.py
+++ b/netbox/netbox/tests/test_config.py
@@ -10,22 +10,28 @@ CACHES = settings.CACHES
 CACHES['default'].update({'KEY_PREFIX': 'TEST-'})
 
 
+@override_settings(CACHES=CACHES)
 class ConfigTestCase(TestCase):
 
-    @override_settings(CACHES=CACHES)
-    def test_config_init_empty(self):
+    def setUp(self):
+        super().setUp()
+        # Register cleanup so it runs after each test, even on assertion failure.
+        # clear_config drops the thread-local Config; cache.clear wipes the Redis
+        # entries that configrevision.activate writes.
+        self.addCleanup(cache.clear)
+        self.addCleanup(clear_config)
+
+        # Defensive reset against pre-existing state from outside this class.
+        clear_config()
         cache.clear()
 
+    def test_config_init_empty(self):
         config = get_config()
         self.assertEqual(config.config, {})
         self.assertEqual(config.version, None)
 
-        clear_config()
-
-    @override_settings(CACHES=CACHES)
     def test_config_init_from_db(self):
         CONFIG_DATA = {'BANNER_TOP': 'A'}
-        cache.clear()
 
         # Create a config but don't load it into the cache
         configrevision = ConfigRevision.objects.create(data=CONFIG_DATA)
@@ -34,12 +40,8 @@ class ConfigTestCase(TestCase):
         self.assertEqual(config.config, CONFIG_DATA)
         self.assertEqual(config.version, configrevision.pk)
 
-        clear_config()
-
-    @override_settings(CACHES=CACHES)
     def test_config_init_from_cache(self):
         CONFIG_DATA = {'BANNER_TOP': 'B'}
-        cache.clear()
 
         # Create a config and load it into the cache
         configrevision = ConfigRevision.objects.create(data=CONFIG_DATA)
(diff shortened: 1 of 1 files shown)


pheus:

jeremystretch (approved):
Nice, thanks for taking the time to clean this up!

PR #22316, opened 2026-05-28
Fixes #22187: Fix cable trace when entering profiled cable via single-position pass-through ports

### LINKED ISSUE #22187: Cable Profile - Cable retrace bug
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.9

### Python Version

3.12

### Steps to Reproduce

Create the following:
SW1 -> generic switch
PP1 -> Patch Panel with Front / Rear ports (2)
DeviceX with two interfaces, Int1 / Int2.
 

The flow should be 
SW1 Switchport 48 -> Breakout cable **using the cable profile 1C2P-2C1P** -> Front port 1 / Front port 2 on a device -> rear port 1 / Rear port 2 -> DeviceX-Int1 / DeviceX-Int2 

Image

NOTE - this works **without** a profile selected on the cable. 

I ran into this trying to really understand what the purpose of connectors / positions are in the `dcim_cabletermination` table, when you can easily map these via the UI with selecting a profile and they seem to map positions properly based on the order of what you select.

### Expected Behavior

You shouldn't get an error and you should be able to trace this
DeviceX-Int1 traces back to SW1 Interface 48 via RearPort1 → FrontPort1 → breakout cable.
DeviceX-Int2 traces back to SW1 Interface 48 via RearPort2 → FrontPort2 → breakout cable.

### Observed Behavior

```
Error <class 'ValueError'>
Could not map connector 1 position None on side B
No cable is created. No CablePath is stored.
```

jeremystretch:
## Summary

- Fixes a `ValueError: Could not map connector N position None on side X` raised when saving an unprofiled cable whose far end leads back through single-position FrontPort/RearPort pairs into a profiled cable (e.g. the `1C2P:2C1P` breakout).
- Root cause: when the trace reaches a profiled cable with an empty `position_stack`, it defaulted to `[None]`, which the profile's mapping lookup cannot resolve. Single-position pass-through ports never push onto the stack, so this state is reachable from any unprofiled cable on the back side of a patch panel feeding into a profiled cable.
- The fix falls back to deriving positions from the local terminations' own `cable_positions` (which are populated by the profiled cable when it's saved), giving the profile lookup the correct local position to map.

## Test plan

- [x] Added `dcim.tests.test_cablepaths2.CablePathTestCase.test_203_breakout_1c2p_2c1p_to_single_position_passthroughs`, which reproduces the exact topology from the issue (1C2P:2C1P breakout into two single-position FP/RP pairs, each connected to its own interface via an unprofiled cable). Confirmed the test fails on `main` with the original `ValueError` and passes with this fix.
- [x] All 64 existing cable path / profile tests pass (`dcim.tests.test_cablepaths`, `dcim.tests.test_cablepaths2`, `dcim.tests.test_cable_profiles`).
- [x] Broader DCIM suite (`dcim.tests.test_models`, `dcim.tests.test_views`) — 1408 tests pass with no regressions.

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -934,7 +934,17 @@ class CablePath(models.Model):
                 # Profile-based tracing
                 if links[0].profile:
                     cable_profile = links[0].profile_class()
-                    positions = position_stack.pop() if position_stack else [None]
+                    if position_stack:
+                        positions = position_stack.pop()
+                    else:
+                        # When the position stack is empty (e.g. the trace reached this
+                        # profiled cable after crossing single-position pass-through ports
+                        # which don't push onto the stack), derive positions from each
+                        # termination's own cable_positions — which were set by this
+                        # profiled cable when it was saved.
+                        positions = [
+                            pos for term in terminations for pos in (term.cable_positions or [])
+                        ]
                     remote_terminations = []
                     new_positions = []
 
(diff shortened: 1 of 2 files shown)


jeremystretch:
@claude /review

jnovinger (approved):

Issue #22317, opened 2026-05-28
`BackgroundTaskTestCase` lacks teardown, leaking django-rq state into later test suites


pheus:
### Proposed Changes

In `netbox/core/tests/test_views.py`, add a `tearDown` to `BackgroundTaskTestCase` that mirrors the existing `setUp` and clears the `default`/`high`/`low` queues after each test. The current implementation only clears queues in `setUp`, so the queue state written by the *last* test in the class persists in Redis after the class finishes.

Current pattern (lines 191-216):

```python
class BackgroundTaskTestCase(TestCase):
    user_permissions = ()
    ...
    def setUp(self):
        super().setUp()
        self.user.is_superuser = True
        self.user.is_active = True
        self.user.save()

        # Clear all queues prior to running each test
        get_queue('default').connection.flushall()
        get_queue('high').connection.flushall()
        get_queue('low').connection.flushall()
```

Proposed pattern (add immediately below `setUp`):

```python
    def tearDown(self):
        super().tearDown()

        # Clear all queues after each test so no leftover jobs leak into the
        # next test suite. Mirrors JobRunnerTestCase.tearDown.
        get_queue('default').connection.flushall()
        get_queue('high').connection.flushall()
        get_queue('low').connection.flushall()
```

### Justification

`netbox/netbox/tests/test_jobs.py:36-43` (`JobRunnerTestCase`) already establishes the pattern for Redis-touching test classes: a `tearDown` that mirrors the `setUp` queue-clear. `BackgroundTaskTestCase` clears only in `setUp`, so within-class isolation holds (each test's `setUp` wipes everything with `flushall()`), but across suites the last enqueued job survives in Redis until something else clears it. Redis is not torn down by `--keepdb`, so the stale jobs survive across test invocations.

No specific cross-suite failure has been reported; this is preventive cleanup motivated by the inconsistency with the established pattern. The concrete risk it closes: any downstream test that asserts a queue is empty at start would see the leftover jobs.

Source references:

- `netbox/core/tests/test_views.py:191-216`
- `netbox/netbox/tests/test_jobs.py:36-43` (reference `JobRunnerTestCase.tearDown` pattern)

PR #22320, opened 2026-05-28
Closes #22318: Add tearDown to clear RQ queues after tests

### LINKED ISSUE #22318: `BackgroundTaskTestCase` (API tests) lacks teardown, leaking django-rq state into later test suites
### Proposed Changes

In `netbox/core/tests/test_api.py`, add a `tearDown` to `BackgroundTaskTestCase` that mirrors the existing `setUp` and clears the `default`/`high`/`low` queues after each test. This is the API-side sibling of the same class name in `netbox/core/tests/test_views.py`; the fix shape is identical.

Current pattern (lines 171-194):

```python
class BackgroundTaskTestCase(TestCase):
    user_permissions = ()
    ...
    def setUp(self):
        """
        Create a user and token for API calls.
        """
        # Create the test user and assign permissions
        self.user = User.objects.create_user(username='testuser', is_active=True)
        self.token = Token.objects.create(user=self.user)
        self.header = {'HTTP_AUTHORIZATION': f'Bearer {TOKEN_PREFIX}{self.token.key}.{self.token.token}'}

        # Clear all queues prior to running each test
        get_queue('default').connection.flushall()
        get_queue('high').connection.flushall()
        get_queue('low').connection.flushall()
```

Proposed pattern (add immediately below `setUp`):

```python
    def tearDown(self):
        super().tearDown()

        # Clear all queues after each test so no leftover jobs leak into the
        # next test suite. Mirrors JobRunnerTestCase.tearDown.
        get_queue('default').connection.flushall()
        get_queue('high').connection.flushall()
        get_queue('low').connection.flushall()
```

### Justification

`netbox/netbox/tests/test_jobs.py:36-43` (`JobRunnerTestCase`) already establishes the pattern for Redis-touching test classes: a `tearDown` that mirrors the `setUp` queue-clear. `BackgroundTaskTestCase` clears only in `setUp`, so within-class isolation holds (each test's `setUp` wipes the queues with `flushall()`), but across suites the last enqueued job survives in Redis until cleared.

No specific cross-suite failure has been reported; this is preventive cleanup motivated by the inconsistency with the established pattern. The concrete risk it closes: any downstream test that asserts a queue is empty at start would see the leftover jobs.

Source references:

- `netbox/core/tests/test_api.py:171-194`
- `netbox/netbox/tests/test_jobs.py:36-43`

pheus:
### Closes: #22318

Add `RQQueueTestMixin` to centralize RQ queue cleanup for test cases that interact with background jobs.

The mixin clears all RQ queues in both `setUp()` and `tearDown()`, preventing jobs created by one test from leaking into later unrelated test runs.

This also replaces duplicated queue cleanup logic in the core and netbox tests with the shared mixin for better maintainability.

### DIFF
--- a/netbox/core/tests/test_api.py
+++ b/netbox/core/tests/test_api.py
@@ -11,8 +11,9 @@ from rq.job import JobStatus
 from rq.registry import FailedJobRegistry, StartedJobRegistry
 
 from users.constants import TOKEN_PREFIX
-from users.models import Token, User
+from users.models import Token
 from utilities.testing import APITestCase, APIViewTestCases, TestCase
+from utilities.testing.mixins import RQQueueTestMixin
 from utilities.testing.utils import disable_logging
 
 from ..models import *
@@ -168,7 +169,7 @@ class JobTestCase(
         )
 
 
-class BackgroundTaskTestCase(TestCase):
+class BackgroundTaskTestCase(RQQueueTestMixin, TestCase):
     user_permissions = ()
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


arthanson (changes requested):
I noticed that same setup code is in test_views, wondering if you looked at adding it there as well?  

Also test_jobs has the teardown but doesn't clear in the setup - also wondering if it would make sense to just be a mixin for the tests that has the setup and teardown logic for clearing the queues, maybe a bit more DRY.

pheus:
Thanks for the review! The view-side cleanup was handled separately in #22317, but I agree that extracting the shared RQ cleanup into a mixin is the right call here.

I’ve updated this PR to add `RQQueueTestMixin` and use it across the queue-related tests, including `test_jobs`, so the queues are cleared consistently in both `setUp()` and `tearDown()`.

pheus:

pheus:
I agree this is worth investigating, but I think changing the Redis/RQ cleanup strategy is broader than the original scope of this issue.

The current `flushall()` behavior was inherited from the existing tests, so this PR is focused on making the existing cleanup consistent and DRY. I’ve opened #22341 as a follow-up housekeeping issue to look into whether we can replace it with a less destructive RQ-specific cleanup.

arthanson (approved):

PR #22323, opened 2026-05-28
Fixes #22301: Avoid reverse relation name collision among tagged models

### LINKED ISSUE #22301: Name conflict when multiple plugins introduce taggable models of the same name
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Install two plugins which have a model of the same name, both of which inherit from TagsMixin.
2. Attempt to start NetBox.

(In the example below, `plugin_a` and `plugin_b` both introduce a model named `Foo`.)

### Expected Behavior

The redundant names should not cause an issue as they should be namespaced as separate Django apps.

### Observed Behavior

Django complains about a conflict in manager names on the Tag model:

```
plugin_a.Foo.tags: (fields.E304) Reverse accessor 'Tag.foo_set' for 'plugin_a.Foo.tags'
clashes with reverse accessor for 'plugin_b.Foo.tags'.
    HINT: Add or change a related_name argument to the definition for 'plugin_a.Foo.tags'
    or 'plugin_b.Foo.tags'.
```

(Note, however, that the Tag model doesn't actually have a `foo_set` attribute when only one plugin is installed.)

jeremystretch:
### Fixes: #22301

- Introduce NetBoxTaggableManagerField subclassing TaggableManager to override `contribute_to_class()` to properly suppress the reverse relation
- Add TagsMixinCollisionTestCase

### DIFF
--- a/netbox/extras/managers.py
+++ b/netbox/extras/managers.py
@@ -1,10 +1,11 @@
 from django.db import router
 from django.db.models import signals
-from taggit.managers import _TaggableManager
+from taggit.managers import TaggableManager, _TaggableManager
 from taggit.utils import require_instance_manager
 
 __all__ = (
     'NetBoxTaggableManager',
+    'NetBoxTaggableManagerField',
 )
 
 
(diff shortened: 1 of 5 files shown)


jeremystretch:

jeremystretch:
I'm okay with points 1 and 2 above.

pheus (approved):
Very nice! Thanks!

Alef-Burzmali:
Hi. I do believe this change should have been considered a breaking change, even if I welcome it, as the reverse relation is a standard Django API and breaks plugins using it.

I'm also wondering why you are suppressing the related_name with +, in addition to prefixing it with the app_label. It would have been easier to fix our code as a plugin author if we only had to rename the reverse field.

PR #22326, opened 2026-05-28
Fixes #22325: AttributeError when creating choice set with base choices

### LINKED ISSUE #22325: AttributeError when creating a Custom Field Choice Set with base choices
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Navigate to **Customization → Custom Field Choice Sets → Add**
2. Select any built-in base choice set (e.g. IATA, ISO 3166, or UN/LOCODE)
3. Click **Create**

### Expected Behavior

The choice set is created successfully.

### Observed Behavior

An `Internal Server Error` is returned with the following traceback:

```
File "netbox/extras/models/customfields.py", line 975, in clean
    valid_choice_values.update(CHOICE_SETS.get(self.base_choices).values())
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'values'
```

`CHOICE_SETS` values (IATA, ISO_3166, UN_LOCODE) are lists of `(value, label)` tuples, not dicts. The call to `.values()` was introduced in #21984 and treats them as dicts.

**Fix:** replace the `.values()` call with an iteration that extracts the first element from each tuple:

```python
valid_choice_values.update(value for value, _ in CHOICE_SETS.get(self.base_choices))
```

bctiemann:
## Summary

`CustomFieldChoiceSet.clean()` calls `.values()` on the result of `CHOICE_SETS.get(self.base_choices)`, treating it as a dict. However, the built-in choice sets (`IATA`, `ISO_3166`, `UN_LOCODE`) are lists of `(value, label)` tuples, not dicts. This raises `AttributeError: 'list' object has no attribute 'values'` whenever a choice set is saved with `base_choices` set.

The bug was introduced by #21984. The fix replaces the `.values()` call with a generator expression that extracts the first element from each tuple — matching the same pattern already used elsewhere in the model (e.g. line 924 and 1011).

## Test plan

- Added `test_choice_set_with_base_choices_validates_without_error` to `CustomFieldTestCase` — verifies that `full_clean()` succeeds for each built-in base choice set (`IATA`, `ISO_3166`, `UN_LOCODE`).
- Manually reproduced: navigating to **Customization → Custom Field Choice Sets → Add**, selecting a base choice set, and submitting now creates the object successfully.

Closes: #22325

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -972,7 +972,7 @@ class CustomFieldChoiceSet(CloningMixin, ExportTemplatesMixin, OwnerMixin, Chang
         extra_choice_values = set()
 
         if self.base_choices:
-            valid_choice_values.update(CHOICE_SETS.get(self.base_choices).values())
+            valid_choice_values.update(value for value, _ in CHOICE_SETS.get(self.base_choices))
 
         if self.extra_choices:
             for value, _label in self.extra_choices:
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -1031,7 +1031,7 @@ class CustomFieldChoiceSet(CloningMixin, ExportTemplatesMixin, OwnerMixin, Chang
     def save(self, *args, **kwargs):
 
         # Sort choices if alphabetical ordering is enforced
-        if self.order_alphabetically:
+        if self.order_alphabetically and self.extra_choices:
             self.extra_choices = sorted(self.extra_choices, key=lambda x: x[0])
 
         return super().save(*args, **kwargs)
--- a/netbox/extras/tests/test_customfields.py
+++ b/netbox/extras/tests/test_customfields.py
@@ -466,10 +466,9 @@ class CustomFieldTestCase(TestCase):
 
         self.assertIn('choice_colors', cm.exception.message_dict)
 
+    @tag('regression')
     def test_choice_set_with_base_choices_validates_without_error(self):
-        """Regression test for #22325: creating a choice set with only base_choices
-        set must not raise AttributeError from clean().  CHOICE_SETS values are lists
-        of (value, label) tuples, not dicts, so .values() must not be called on them."""
+    """Regression test for #22325: base-only choice sets must validate."""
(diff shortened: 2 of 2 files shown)


netbox/extras/tests/test_customfields.py
```diff
@@ -466,6 +466,15 @@ def test_choice_set_invalid_color_mapping_structure(self):
 
         self.assertIn('choice_colors', cm.exception.message_dict)
 
+    def test_choice_set_with_base_choices_validates_without_error(self):
+        """Regression test for #22325: creating a choice set with only base_choices
+        set must not raise AttributeError from clean().  CHOICE_SETS values are lists
+        of (value, label) tuples, not dicts, so .values() must not be called on them."""
```
pheus:
```suggestion
    @tag('regression')
    def test_choice_set_with_base_choices_validates_without_error(self):
    """Regression test for #22325: base-only choice sets must validate."""
```

netbox/extras/models/customfields.py
pheus: Possible adjacent edge case: `CustomFieldChoiceSet.save()` currently sorts `self.extra_choices` whenever `order_alphabetically` is true. For a base-only choice set, `extra_choices` can be `None`, so a user selecting a base choice set and enabling alphabetical ordering may still hit a `TypeError` during save.

Could we guard this as well?

```python
if self.order_alphabetically and self.extra_choices:
    self.extra_choices = sorted(self.extra_choices, key=lambda x: x[0])
```

It would also be good to cover this with the regression test, since the reported workflow is creating the object rather than only calling `full_clean()`.

pheus (changes requested):
Thanks for fixing this!

One small required change: since this is covering a bug regression, please tag the new test with `@tag('regression')`, consistent with the other regression tests in this file.

bctiemann:
Good catch — both addressed in the latest commit: guarded the `save()` path with `and self.extra_choices`, and extended the regression test to call `save()` with `order_alphabetically=True` so both code paths are covered.

pheus (approved):
Thanks for the follow-up! This looks good to me now.

PR #22327, opened 2026-05-28
Fixes #22210: Respect filters when rendering IPAM child availability views

### LINKED ISSUE #22210: Saved Filter doesn't work in additional tabs
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

1. Open https://demo.netbox.dev/ (version 4.6.0)
2. Add IPAM-Prefix 100.64.64.0/24
3. Add IPAM-"IPAddresses"  100.64.64.1/24 with tenant "Cyberdyne Systems" (or other)
4. Add IPAM-"IPAddresses"  100.64.64.2/24 with tenant "Dunder-Mifflin, Inc." (or other)
5. Add Customization-"Saved Filters" new filter with:
- Name "Cyberdyne_tenant_Filter"
- Parameters. With the filtering parameters of the tenant using above in item 3 ("Cyberdyne Systems"). For example (in demo instance):
```
{
    "tenant_id": [
        "7"
    ]
}
```
Tenant "Cyberdyne Systems" has id=7 in demo instance.
- Assigned Models:
```
IPAM | IPAdresses
IPAM | Prefix
```
6. Go to IPAM-Prefix, find prefix 100.64.64.0/24 (it was created in item 2 above) and click on it.
Open "IP Adresses tab". You'll see two IPadresses (it was created in item 3,4 above)
7. Choose "Cyberdyne_tenant_Filter" in SavedFilter list.

### Expected Behavior

The IP Adresses list is filtered according to the filter conditions.

### Observed Behavior

The IP Adresses list is not filtered according to the filter conditions.
Image

But it works fine in IPAM-IPAddress section:
Image

pheus:
### Fixes: #22210

This PR fixes filtering in IPAM child availability views by detecting when the child queryset has been narrowed by a direct filter or saved filter.

Synthetic “available” rows are still rendered for unfiltered views, preserving the existing default behavior. When filters are active, those synthetic rows are suppressed so filtered-out child objects are not represented as available space.

The change is scoped to the relevant IPAM child views via a small `ChildAvailabilityMixin`, leaving the generic `ObjectChildrenView` behavior unchanged.

### DIFF
--- a/netbox/netbox/views/generic/object_views.py
+++ b/netbox/netbox/views/generic/object_views.py
@@ -11,7 +11,6 @@ from django.urls import reverse
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
 from django.utils.translation import gettext as _
-from django_filters.constants import EMPTY_VALUES
 
 from core.signals import clear_events
 from netbox.object_actions import BulkDelete, BulkEdit, CloneObject, DeleteObject, EditObject
@@ -105,7 +104,6 @@ class ObjectChildrenView(ObjectView, ActionsMixin, TableMixin):
     table = None
     filterset = None
     filterset_form = None
-    filterset_instance = None
     actions = (CloneObject, EditObject, DeleteObject, BulkEdit, BulkDelete)
     template_name = 'generic/object_children.html'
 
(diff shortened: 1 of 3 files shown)


pheus:

pheus:
@claude /review

Point 1 and 3 addressed.
Point 2 is intentional.

netbox/netbox/views/generic/object_views.py
```diff
@@ -119,6 +121,66 @@ def get_children(self, request, parent):
             class_name=self.__class__.__name__
         ))
 
+    def get_filterset(self, request, queryset):
```
jeremystretch: We should avoid introducing new methods on generic views in a patch release, as it introduces the risk of a breaking change for subclasses. The logic within `get_filterset()` and `filter_queryset()` can be collapsed into `get()`, at least until we have an opportunity to break them out separately. `has_active_filters()` (if needed) should probably be moved to a separate mixin class specifically for the relevant views.

netbox/netbox/views/generic/object_views.py
```diff
@@ -119,6 +121,66 @@ def get_children(self, request, parent):
...
+        if self.filterset is None:
+            return None
+
+        return self.filterset(request.GET, queryset, request=request)
+
+    def filter_queryset(self, request, queryset):
+        """
+        Apply the configured FilterSet to the child object queryset.
+        """
+        self.filterset_instance = self.get_filterset(request, queryset)
```
jeremystretch: As an alternative to attaching the filterset instance to the view for inspection by `has_active_filters()`, can we introspect the filtered queryset (i.e. `qs.query.where`) to determine whether filters have been applied? I'm sure there are caveats to this approach but it seems _reasonably_ safe given the use case here (disabling the annotation of available space).

```python
>>> qs = dcim.Site.objects.all()
>>> qs.query.where
<WhereNode: (AND: )>
>>> bool(qs.query.where)
False
```
```python
>>> qs = dcim.Site.objects.filter(status='active')
>>> qs.query.where
<WhereNode: (AND: Exact(Col(dcim_site, dcim.Site.status), 'active'))>
>>> bool(qs.query.where)
True
```

jeremystretch (changes requested):

pheus:

pheus:
@claude /review
Point 1 is addressed.
Point 2 got an comment.

jeremystretch (approved):
Thanks @pheus!

PR #22331, opened 2026-05-28
#22270 - Skip cascade-deleted objects when clearing reverse SET_NUL relations

### LINKED ISSUE #22270: Deleting an object with reverse SET_NULL relations records UPDATE ObjectChange after DELETE
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Create a device with two physical interfaces (e.g. eth0, eth1).
2. Add a LAG interface (lag0) to the same device.
3. Assign both physical interfaces to the LAG (set lag=lag0 on each, save).
4. Delete the device.
5. Inspect the ObjectChange records produced during the delete:

Having the update after delete will break branching when it replays the ObjectChange records.

**Note:** Can paste the following into django shell to repro the issue

<details>
  <summary>Script to show issue (click to expand)</summary>

  ```python
def repro():
    from uuid import uuid4
    from django.test import RequestFactory
    from core.models import ObjectChange
    from dcim.choices import InterfaceTypeChoices
    from dcim.models import Device, DeviceRole, DeviceType, Interface, Manufacturer, Site
    from netbox.context_managers import event_tracking
    from users.models import User
    site, _ = Site.objects.get_or_create(name='Repro Site', slug='repro-site')
    mfr, _ = Manufacturer.objects.get_or_create(name='Repro Mfr', slug='repro-mfr')
    dtype, _ = DeviceType.objects.get_or_create(manufacturer=mfr, model='Repro Type', slug='repro-type')
    role, _ = DeviceRole.objects.get_or_create(name='Repro Role', slug='repro-role')
    user, _ = User.objects.get_or_create(username='repro-user')
    Device.objects.filter(name='Repro Device', site=site).delete()
    device = Device.objects.create(name='Repro Device', site=site, device_type=dtype, role=role)
    m1 = Interface.objects.create(device=device, name='eth0', type=InterfaceTypeChoices.TYPE_1GE_FIXED)
    m2 = Interface.objects.create(device=device, name='eth1', type=InterfaceTypeChoices.TYPE_1GE_FIXED)
    lag = Interface.objects.create(device=device, name='lag0', type=InterfaceTypeChoices.TYPE_LAG)
    m1.lag = lag; m1.save()
    m2.lag = lag; m2.save()
    labels = {m1.pk: 'member1', m2.pk: 'member2', lag.pk: 'lag'}
    request = RequestFactory().post('/')
    request.id = uuid4()
    request.user = user
    with event_tracking(request):
        device.delete()
    print('\nObjectChange records produced by device.delete():')
    print('-' * 72)
    for c in ObjectChange.objects.filter(request_id=request.id).order_by('time'):
        label = labels.get(c.changed_object_id, '')
        print(f"  {c.time}  {c.changed_object_type.model:>10}.{c.changed_object_id:<4} {c.action:<7} {label}")
    print('-' * 72)
    bug = False
    for pk in (m1.pk, m2.pk):
        actions = [c.action for c in ObjectChange.objects.filter(request_id=request.id, changed_object_id=pk).order_by('time')]
        if 'delete' in actions and 'update' in actions[actions.index('delete') + 1:]:
            bug = True
            print(f"BUG REPRODUCED: interface pk={pk} actions in order = {actions}")
    if not bug:
        print('Bug did NOT reproduce — check that members were created before the LAG.')

repro()
```
</details>

### Expected Behavior

Each interface's last ObjectChange record is delete. No update record is written for an interface after its delete record.

### Observed Behavior

For each physical interface, an update ObjectChange is recorded after its delete record:

```
interface  <member1>  delete
interface  <member2>  delete
interface  <lag>      delete
interface  <member1>  update   ← after DELETE
interface  <member2>  update   ← after DELETE
device     <device>   delete
```

The trailing update originates from core.signals.handle_deleted_object (netbox/core/signals.py:218-234): when the LAG's pre_delete fires, it iterates the reverse member_interfaces relation and calls setattr(obj, 'lag', None); obj.save() on each member. Because all interfaces are deleted in one cascade, every pre_delete runs before any SQL DELETE, so the members still exist in the DB at that point and the save records an UPDATE for rows whose DELETE has already been logged. The ordering only manifests when member PKs are lower than the LAG's PK (the common case when a LAG is added on top of existing ports).

arthanson:
### Closes: #22270 

Reuse the _signals_received.pre_delete bookkeeping NetBox already maintains for de-duping cascade pre_delete signals. Before nulling-and-saving (or .remove()-ing) a related object, skip it if it's itself part of this same delete operation.

### DIFF
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -225,7 +225,15 @@ def handle_deleted_object(sender, instance, **kwargs):
                 # We only care about triggering the m2m_changed signal for models which support
                 # change logging
                 continue
+            related_object_type = ContentType.objects.get_for_model(related_model)
             for obj in related_model.objects.filter(**{related_field_name: instance.pk}):
+                # Skip any related object that is itself being deleted as part of this same
+                # operation (e.g. a sibling caught up in the same cascade). Its deletion has
+                # already been recorded, so nulling the FK and saving here would record an
+                # UPDATE ObjectChange *after* the object's DELETE, corrupting the changelog and
+                # breaking branch replay. (Ref: #22270)
+                if (related_object_type, obj.pk) in _signals_received.pre_delete:
+                    continue
                 obj.snapshot()  # Ensure the change record includes the "before" state
                 if type(relation) is ManyToManyRel:
                     getattr(obj, related_field_name).remove(instance)
--- a/netbox/core/signals.py
+++ b/netbox/core/signals.py
@@ -232,6 +232,14 @@ def handle_deleted_object(sender, instance, **kwargs):
                 # already been recorded, so nulling the FK and saving here would record an
                 # UPDATE ObjectChange *after* the object's DELETE, corrupting the changelog and
                 # breaking branch replay. (Ref: #22270)
+                #
+                # Note this is order-dependent: it only fires once the related object's own
+                # pre_delete has run (adding it to the set). If the cascade happens to delete
+                # this instance *before* the related object, the guard won't trigger and the
+                # related object still gets an UPDATE — but in the harmless UPDATE-then-DELETE
+                # order, not the corrupting DELETE-then-UPDATE order. Fully suppressing it in
+                # every ordering would require the complete deletion set, which isn't available
+                # from a pre_delete signal.
                 if (related_object_type, obj.pk) in _signals_received.pre_delete:
                     continue
                 obj.snapshot()  # Ensure the change record includes the "before" state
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
per claud review:

- Ordering dependency: Expanded the code comment to document that the guard is order-dependent
- Redundant test assertions: Collapsed the three overlapping assertions into a single exact-sequence check.

bctiemann (approved):

PR #22336, opened 2026-05-29
Fixes #22251: Re-parent child ModuleBays when a Module is moved to a new bay

### LINKED ISSUE #22251: Moving a multi-bay module does not move its children
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1 (demo.netbox.dev)

### Python Version

3.12

### Steps to Reproduce

1. create a multi-bay module type - e.g. something that represents a 4 x SFP pci card
2. create a multi-bay device type - e.g. a motherboard with 2 pci slots
3. install the pci card into a pci slot and review - it looks like this at demo.netbox.dev

Image

4. use the edit module feature to change the slot... e.g. https://demo.netbox.dev/dcim/modules/151/edit/

Image

5. return to the device and review the installed modules... https://demo.netbox.dev/dcim/devices/215/module-bays/

Image

Here observe that while the installed module has indeed moved from slot 01 to slot 02, it's children devices have not moved along with it - they have stayed "parented" to the previous/inital slot.

## General synopsis:

Moving a module between module bays updates Module.module_bay_id, but descendant ModuleBay.parent_id values are not updated. This leaves ModuleBay.module_id pointing to the moved module while MPTT placement remains under the old parent bay. Rebuilding the tree seems to not help because the stored parent_id is stale.

### Expected Behavior

Children (module bays) of a module should move with the module, not stay stuck to their "initial parent". 

### Observed Behavior

 No errors are raised, just unexpected behaviour.

bctiemann:
## Summary

Closes: #22251

When `Module.module_bay_id` is updated (the module is moved to a different bay), the child `ModuleBay` records belonging to that module were not having their MPTT `parent` pointers updated. They stayed anchored to the old bay in the tree even though the module had moved, producing the display anomaly reported in the issue.

**Root cause:** `ModuleBay.save()` derives `self.parent` from `self.module.module_bay`. On initial creation this runs automatically, but when only the *module's* `module_bay_id` changes there is nothing to trigger a re-save of the child bays. Their stored `parent_id` goes stale.

**Fix:** In `Module.save()`, snapshot the current `module_bay_id` from the DB before saving. If it changed, iterate and re-save each child `ModuleBay`. The existing `ModuleBay.save()` logic then recalculates `parent` from the now-updated `module.module_bay` and MPTT moves each node to the correct position in the tree.

## Test plan

- [x] New regression test `ModuleBayTestCase.test_moving_module_reparents_child_module_bays` creates a module with two child bays, moves the module to a different top-level bay, and asserts both `parent_id` and `tree_id` on each child reflect the new location.
- [x] Full `ModuleBayTestCase` (20 tests) passes.

### DIFF
--- a/netbox/dcim/models/modules.py
+++ b/netbox/dcim/models/modules.py
@@ -326,8 +326,8 @@ class Module(TrackingModelMixin, PrimaryModel, ConfigContextModel):
 
     def save(self, *args, **kwargs):
         is_new = self.pk is None
+        old_module_bay_id = None
 
-        # Capture the current bay before saving so we can detect a move.
         if not is_new:
             old_module_bay_id = Module.objects.filter(pk=self.pk).values_list(
                 'module_bay_id', flat=True
@@ -335,10 +335,7 @@ class Module(TrackingModelMixin, PrimaryModel, ConfigContextModel):
 
         super().save(*args, **kwargs)
 
-        # When the module is moved to a different bay, re-save each child
-        # ModuleBay so ModuleBay.save() can update its MPTT parent to the
-        # new bay (ModuleBay.parent is derived from module.module_bay).
-        if not is_new and old_module_bay_id != self.module_bay_id:
+        if old_module_bay_id is not None and old_module_bay_id != self.module_bay_id:
             for child_bay in self.modulebays.all():
                 child_bay.save()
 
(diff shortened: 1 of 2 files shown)


bctiemann:

bctiemann:
Both actionable items are addressed.

arthanson (approved):

Issue #22056, opened 2026-04-30
Clean up `.gitignore`


jeremystretch:
### Proposed Changes

Remove obsolete entries in the project's `.gitignore`, and comment what remains to explain why it's needed.

### Justification

This file has accumulated a lot of cruft over the years and can likely be pared down significantly.

PR #22337, opened 2026-06-01
Fixes #22310: Restore tracked placeholder in project-static/docs to prevent staticfiles warning

### LINKED ISSUE #22310: Warning about missing project-static/docs shown during upgrade.sh
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.13

### Steps to Reproduce

Prepare a NetBox installation (based on NetBox installation docs):

```
sudo -i
mkdir -p /opt/nbtest/
cd /opt/nbtest/
git clone https://github.com/netbox-community/netbox.git .
git checkout v4.6.1
adduser --system --group netbox
chown --recursive netbox netbox/media/
chown --recursive netbox netbox/reports/
chown --recursive netbox netbox/scripts/
cp netbox/netbox/configuration_example.py netbox/netbox/configuration.py
python3 netbox/generate_secret_key.py
vi netbox/netbox/configuration.py (set SECRET_KEY)
./upgrade.sh
```
(Note: This is not a **full** NetBox setup but enough to demonstrate the warning below)

### Expected Behavior

No warnings during the upgrade script run.

### Observed Behavior

Warning about missing `/opt/nbtest/netbox/project-static/docs`:

```
System check identified some issues:

WARNINGS:
?: (staticfiles.W004) The directory '/opt/nbtest/netbox/project-static/docs' in the STATICFILES_DIRS setting does not exist.
```

(In a real NetBox install this message is shown once more later in the process)

jnovinger:
### Closes: #22310

On a fresh clone, `upgrade.sh` emits `staticfiles.W004` because `netbox/project-static/docs` doesn't exist yet. `settings.py` registers that directory in `STATICFILES_DIRS`, and Django runs system checks on every `manage.py` command. In `upgrade.sh` the `migrate` and `trace_paths` commands run before `zensical build`, so the checks fire against a directory that hasn't been built.

Before 4.6 a tracked `.info` placeholder kept the directory present. #21742 deleted that file and #22056 collapsed the `.gitignore` rule to the directory form, so nothing keeps the directory in a fresh checkout anymore.

This restores a tracked placeholder, switching the ignore rule back to the `dir/*` + `!dir/.gitkeep` form already used for `media/`, `reports/`, and `scripts/` in the same file, and adds an empty `netbox/project-static/docs/.gitkeep`.

Verified with `manage.py check`: with the directory absent, `staticfiles.W004` fires; with the placeholder present, it's gone (only the unrelated `netbox.W001` Postgres warning remains).

One wrinkle worth flagging: `zensical build` clears `site_dir` (`netbox/project-static/docs`) and deletes `.gitkeep` along with it, so after any local build `git status` shows the placeholder as deleted (`AD`). This does not affect the fix, since W004 only fires on the pre-build `migrate`/`trace_paths` commands when the placeholder still exists.

This differs from the pre-4.6 behavior. mkdocs' `clean_directory()` skips dotfiles (`entry.startswith('.')`), so the old `.info` placeholder survived builds; zensical clears them. The directory-handling side of this was reported upstream as zensical#621 and partially fixed in 0.0.40, but the dotfile deletion specifically remains: verified on zensical 0.0.43 that a dotfile placed in `site_dir` is still removed by `zensical build`. Restoring the placeholder is the correct fix for the warning regardless; full alignment with mkdocs' dotfile handling would need to land upstream in zensical.


pheus (approved):
Thanks!

jnovinger:
Filed https://github.com/zensical/zensical/issues/728 to track the "wrinkle" regression noted above.

PR #22342, opened 2026-06-01
Fixes #22158: Cache empty config revision state to avoid per-request queries

### LINKED ISSUE #22158: ConfigRevision queries stay in DB, expending all connections
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.14

### Steps to Reproduce

1. Set `CONN_MAX_AGE` above 0
2. Deploy NetBox v4.6.0 with `netboxlabs-netbox-branching==1.0.2` (Branching might not be needed)
3. Wait few minutes
4. Observe DB state and logs
```
SELECT COUNT(*) FROM pg_stat_activity WHERE query LIKE '%core_configrevision%';
```

Note: Could not reproduce on v4.5.10, appears newly on v4.6.0 even on fresh DB.

### Expected Behavior

`core_configrevision` query count never gets above 3

### Observed Behavior

`core_configrevision` query count keeps increasing steadily until it hits DB connection limit (100 in my case)

```
netbox=# SELECT * FROM pg_stat_activity;
 datid | datname | pid | leader_pid | usesysid | usename | application_name | client_addr | client_hostname | client_port |         backend_start         |          xact_start           |          query_start          |         state_change          | wait_event_type |     wait_event      | state  | backend_xid | backend_xmin | query_id |                                                                                                                       query                                                                                                                        |         backend_type
-------+---------+-----+------------+----------+---------+------------------+-------------+-----------------+-------------+-------------------------------+-------------------------------+-------------------------------+-------------------------------+-----------------+---------------------+--------+-------------+--------------+----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------------------------
       |         |  31 |            |          |         |                  |             |                 |             | 2026-05-11 08:17:13.403801+00 |                               |                               |                               | Activity        | AutoVacuumMain      |        |             |              |          |                                                                                                                                                                                                                                                    | autovacuum launcher
       |         |  32 |            |       10 | netbox  |                  |             |                 |             | 2026-05-11 08:17:13.40405+00  |                               |                               |                               | Activity        | LogicalLauncherMain |        |             |              |          |                                                                                                                                                                                                                                                    | logical replication launcher
 16384 | netbox  | 106 |            |       10 | netbox  |                  | 172.18.0.5  |                 |       47238 | 2026-05-11 08:22:17.197878+00 |                               | 2026-05-11 08:22:17.205861+00 | 2026-05-11 08:22:17.206877+00 | Client          | ClientRead          | idle   |             |              |          | SELECT "core_configrevision"."id", "core_configrevision"."active", "core_configrevision"."created", "core_configrevision"."comment", "core_configrevision"."data" FROM "core_configrevision" ORDER BY "core_configrevision"."created" DESC LIMIT 1 | client backend
 16384 | netbox  | 107 |            |       10 | netbox  |                  | 172.18.0.5  |                 |       36786 | 2026-05-11 08:22:22.262909+00 |                               | 2026-05-11 08:22:22.268876+00 | 2026-05-11 08:22:22.269517+00 | Client          | ClientRead          | idle   |             |              |          | SELECT "core_configrevision"."id", "core_configrevision"."active", "core_configrevision"."created", "core_configrevision"."comment", "core_configrevision"."data" FROM "core_configrevision" ORDER BY "core_configrevision"."created" DESC LIMIT 1 | client backend
 16384 | netbox  | 108 |            |       10 | netbox  |                  | 172.18.0.5  |                 |       36790 | 2026-05-11 08:22:27.316628+00 |                               | 2026-05-11 08:22:27.322778+00 | 2026-05-11 08:22:27.323382+00 | Client          | ClientRead          | idle   |             |              |          | SELECT "core_configrevision"."id", "core_configrevision"."active", "core_configrevision"."created", "core_configrevision"."comment", "core_configrevision"."data" FROM "core_configrevision" ORDER BY "core_configrevision"."created" DESC LIMIT 1 | client backend
 16384 | netbox  | 109 |            |       10 | netbox  |                  | 172.18.0.5  |                 |       41086 | 2026-05-11 08:22:32.373678+00 |                               | 2026-05-11 08:22:32.380469+00 | 2026-05-11 08:22:32.381643+00 | Client          | ClientRead          | idle   |             |              |          | SELECT "core_configrevision"."id", "core_configrevision"."active", "core_configrevision"."created", "core_configrevision"."comment", "core_configrevision"."data" FROM "core_configrevision" ORDER BY "core_configrevision"."created" DESC LIMIT 1 | client backend

[snipped]
```

```
netbox=#  SELECT COUNT(*) FROM pg_stat_activity WHERE query LIKE '%core_configrevision%';
 count
-------
   100
(1 row)

netbox=#  SELECT COUNT(*) FROM pg_stat_activity;
 count
-------
   105
(1 row)
```

No ConfigRevision exists:
```
netbox=# SELECT * FROM core_configrevision;
 id | created | comment | data | active 
----+---------+---------+------+--------
(0 rows)
```

Docker logs:
```
postgres-1       | 2026-05-11 08:51:42.701 UTC [552] FATAL:  sorry, too many clients already
netbox-1         | WARNING 2026-05-11 08:51:42,701 Skipping config initialization (database unavailable)
netbox-1         | INFO 2026-05-11 08:51:42,702 "GET /api/ HTTP/1.1" 200 549
postgres-1       | 2026-05-11 08:51:47.749 UTC [553] FATAL:  sorry, too many clients already
netbox-1         | WARNING 2026-05-11 08:51:47,749 Skipping config initialization (database unavailable)
netbox-1         | INFO 2026-05-11 08:51:47,750 "GET /api/ HTTP/1.1" 200 549
```

jnovinger:
### Closes: #22158

When no `ConfigRevision` exists in the database, NetBox re-queried `core_configrevision` on every request. `Config._populate_from_db()` hit the "no revision" branch and returned without caching anything, so the cache was never seeded and each request (e.g. a health check polling `/api/`) repeated the lookup. With `CONN_MAX_AGE > 0`, the idle connections left behind made the query appear to accumulate in `pg_stat_activity`.

This caches the empty config state on first load. A sentinel distinguishes a genuine cache miss from a cached-empty state, so the database is consulted only when the cache has never been populated. Creating the first `ConfigRevision` overwrites the empty state via the existing `post_save` handler. The `DatabaseError` branch intentionally still does not cache, so management commands run before the database is reachable will retry.

Verified with new regression tests (the configrevision query fires once on a cold cache and not on subsequent requests) and confirmed against a running server with no `ConfigRevision` and `CONN_MAX_AGE` set: the query runs once, then requests are served from cache.

Notes for reviewers:

- The branching plugin's separate per-schema connection accumulation (dynamic `schema_*` aliases bypassing Django's `close_old_connections`) is a distinct issue, already fixed in netbox-branching v1.0.2 ([netboxlabs/netbox-branching#358](https://github.com/netboxlabs/netbox-branching/issues/358)). It is not addressed here.
- Not re-reading the database while the cache is warm is the intended design (the cache is the source of truth, synced on `ConfigRevision` save). This PR extends that contract to the empty state and does not change behavior on revision deletion.

### DIFF
--- a/netbox/netbox/config/__init__.py
+++ b/netbox/netbox/config/__init__.py
@@ -77,7 +77,9 @@ class Config:
         """Populate config data from Redis cache"""
         cached = cache.get('config', _MISSING)
         self._cache_miss = cached is _MISSING
-        self.config = {} if self._cache_miss else cached
+        # A cached value of None (ConfigRevision.data is nullable) or {} is a legitimate empty
+        # config and must not be treated as a miss; normalize it to an empty dict.
+        self.config = {} if self._cache_miss else (cached or {})
         self.version = cache.get('config_version')
         if self.config:
             logger.debug("Loaded configuration data from cache")
--- a/netbox/netbox/config/__init__.py
+++ b/netbox/netbox/config/__init__.py
@@ -20,6 +20,10 @@ _thread_locals = threading.local()
 
 logger = logging.getLogger('netbox.config')
 
+# Sentinel used to distinguish a cache miss from a cached "empty" config (an empty dict is a
+# legitimate cached value when no ConfigRevision exists).
+_MISSING = object()
+
 
 def get_config():
     """
@@ -47,7 +51,9 @@ class Config:
     """
     def __init__(self):
         self._populate_from_cache()
-        if not self.config or not self.version:
+        # Only consult the database when the cache has genuinely never been populated. A cached
+        # empty config (no ConfigRevision) is authoritative and must not trigger a re-query.
+        if self._cache_miss:
             self._populate_from_db()
         self.defaults = {param.name: param.default for param in PARAMS}
 
@@ -69,7 +75,9 @@ class Config:
 
     def _populate_from_cache(self):
         """Populate config data from Redis cache"""
-        self.config = cache.get('config') or {}
+        cached = cache.get('config', _MISSING)
+        self._cache_miss = cached is _MISSING
+        self.config = {} if self._cache_miss else cached
         self.version = cache.get('config_version')
         if self.config:
             logger.debug("Loaded configuration data from cache")
@@ -86,10 +94,17 @@ class Config:
             revision = ConfigRevision.objects.order_by('-created').first()
             if revision is None:
                 logger.debug("No configuration found in database; proceeding with default values")
+                # Cache the empty state so subsequent requests are served from the cache rather than
+                # re-querying the database on every request (#22158). Creating the first
+                # ConfigRevision overwrites this via the post_save handler.
+                cache.set('config', {}, None)
+                cache.set('config_version', None, None)
+                self._populate_from_cache()
                 return
             logger.debug(f"No active configuration revision found; falling back to most recent (#{revision.pk})")
         except DatabaseError:
-            # The database may not be available yet (e.g. when running a management command)
+            # The database may not be available yet (e.g. when running a management command). Do NOT
+            # cache anything here, so the next instantiation re-queries once the database is reachable.
             logger.warning("Skipping config initialization (database unavailable)")
             return
 
--- a/netbox/netbox/config/__init__.py
+++ b/netbox/netbox/config/__init__.py
@@ -75,12 +75,19 @@ class Config:
 
     def _populate_from_cache(self):
         """Populate config data from Redis cache"""
-        cached = cache.get('config', _MISSING)
-        self._cache_miss = cached is _MISSING
+        cached_config = cache.get('config', _MISSING)
+        cached_version = cache.get('config_version', _MISSING)
+
+        # Treat the cache as warm only when both keys are present. A missing 'config_version'
+        # (e.g. evicted or never written) must re-query the database, even if 'config' is cached.
+        # The no-revision branch writes both keys (config={}, config_version=None), so the
+        # intentional empty state is still a cache hit.
+        self._cache_miss = cached_config is _MISSING or cached_version is _MISSING
+
         # A cached value of None (ConfigRevision.data is nullable) or {} is a legitimate empty
-        # config and must not be treated as a miss; normalize it to an empty dict.
-        self.config = {} if self._cache_miss else (cached or {})
-        self.version = cache.get('config_version')
+        # config and must not crash attribute access; normalize it to an empty dict.
+        self.config = {} if cached_config is _MISSING else (cached_config or {})
+        self.version = None if cached_version is _MISSING else cached_version
         if self.config:
             logger.debug("Loaded configuration data from cache")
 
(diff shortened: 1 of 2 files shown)


jnovinger:

jnovinger:
Confirmed and fixed in d5e832c.

> With the new code, `cached is not _MISSING` (it's `None`), so `self.config` is set to `None`. Any subsequent attribute access via `Config.__getattr__` then does `if item in self.config` # TypeError

Right. `ConfigRevision.data` is nullable, so a revision saved with `data=None` caches `None`, and the old `cache.get('config') or {}` quietly normalized that away. The sentinel rewrite dropped the normalization. The miss check has to stay separate from the value check, so:

```python
self.config = {} if self._cache_miss else (cached or {})
```

A genuine miss still routes to `_populate_from_db()`; a cached `None`/`{}` is treated as an empty config.

> The new regression test `test_empty_then_create_first_revision` uses `data={'BANNER_TOP': 'A'}`, so it doesn't exercise the `data=None` path.

Added `test_config_init_from_cache_null_data`, which activates a revision with `data=None` and asserts config access falls back to defaults without raising. Verified it fails against the unfixed line (`None != {}`) and passes with the fix.

jnovinger:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Previous comment notes how the previous review was addressed.

pheus (changes requested):
Thanks!

I think there’s one cache-integrity edge case worth tightening before merge: `_populate_from_cache()` now treats the cache as warm solely based on whether the `config` key exists. That correctly handles the intended cached-empty state, but it also means a partial cache state like this would skip `_populate_from_db()`:

- `config` exists, possibly `{}` or `None`
- `config_version` is missing because it was evicted, manually deleted, or not written

In that case `Config.version` would remain `None` even if a `ConfigRevision` exists. The previous `if not self.config or not self.version:` path recovered from a missing version key, so I think we should preserve that protection while still allowing the intentional no-revision state where `config_version` is present and set to `None`.

Could we sentinel-check both cache keys?

```python
def _populate_from_cache(self):
    cached_config = cache.get('config', _MISSING)
    cached_version = cache.get('config_version', _MISSING)

    self._cache_miss = cached_config is _MISSING or cached_version is _MISSING
    self.config = {} if cached_config is _MISSING else (cached_config or {})
    self.version = None if cached_version is _MISSING else cached_version

    if self.config:
        logger.debug("Loaded configuration data from cache")
```

The no-`ConfigRevision` branch already writes both `config={}` and `config_version=None`, so a cached `None` version would still be treated as a valid cache hit. Only a missing key would re-query the database.

A small regression test for `cache.delete('config_version')` after activating a revision would lock this down.

Also, CI is currently failing and needs to be addressed before this is merged.

pheus (approved):
Thanks for the update! This looks good to me!

PR #22344, opened 2026-06-01
Closes #18663: Replace assertions with proper error handling

### LINKED ISSUE #18663: Check asserts and replace with RaiseError where appropriate
### Proposed Changes

There are several places in the code where asserts are used (cables.py, events, bulk_views).  There should be a code review of the places where asserts are used in the codebase and replaced where appropriate with a raised error.

### Justification

Asserts are removed when not running in debug mode.  It was noticed in the cabling code that the asserts were used in places where actually errors should have been raised to catch it in production environments.  So an code review should be done on all these usages and make sure an assert is appropriate or if a RaiseError should be used.

pheus:
### Closes: #18663

Replace assert statements with explicit ValueError exceptions throughout the codebase. Add validation tests for all changed error paths to ensure proper exception handling.

### DIFF
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -1583,7 +1583,8 @@ class CableImportForm(PrimaryModelImportForm):
 
         :param side: 'a' or 'b'
         """
-        assert side in 'ab', f"Invalid side designation: {side}"
+        if side not in ('a', 'b'):
+            raise ValueError(_("Invalid side designation: {side}").format(side=side))
 
         device = self.cleaned_data.get(f'side_{side}_device')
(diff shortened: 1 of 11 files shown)


pheus:

pheus:
@claude /review 
Good catches. I agree on both points.

I’ve updated `CableTermination.cache_related_objects()` to raise `ValueError` instead of `ValidationError`, since this is an internal precondition/state check rather than a user-facing validation path. I also updated the corresponding test expectation.

I’ve also restored the eager `gettext` imports in `extras/events.py` and `netbox/views/generic/bulk_views.py` so this PR only changes the new error-handling paths and does not re-characterize the existing runtime messages in those modules.

jnovinger (approved):

PR #22345, opened 2026-06-01
Closes: #15165 - HTMX partial fieldset re-rendering for HTMXSelect forms

### LINKED ISSUE #15165: HTMX Allow updating sub-parts of forms without re-rendering entire form
### NetBox version

v3.7.2

### Feature type

### Proposed functionality

Currently we can re-render forms via HTMX when you change a select field to change the fields that are displayed (for example if you add an interface and change the 802.1Q selection).  

What happens is the entire form is re-rendered via HTMX (causing a slight flash as the entire form is re-rendered). The proposal would be to allow specifying and updating just a sub-set of the fields, not just the whole form.  

### Use case

The issue with the current implementation is that on larger forms (like the interface form) There is a UI 'flash' as the entire form is re-rendered.  Just updating the subset of fields instead of the whole form should make for a smoother experience.

Discussing with Jeremy this could probably be done by either tying the fields to a fieldset and having HTMX handle rendering of the fieldset, or else using https://htmx.org/attributes/hx-select/ Tying it to a fieldset might also make the implementation of forms more DRY as you could potentially just mark fieldsets as handled via HTMX and the underlying form-view class could auto handle the return info instead of currently how we handle the return in each form-view individually.

### Database changes

None

### External dependencies

None

bctiemann:
## Summary

Replaces the full `#form_fields` swap with targeted per-fieldset swaps for all `HTMXSelect` fields where it is safe to do so, eliminating the visible flash on large forms when a select value changes.

### How it works

- `FieldSet` gains an optional `fieldset_id=` parameter that renders as the HTML `id` on the fieldset `<div>`.
- `HTMXSelect` gains an optional `hx_fieldset_id=` parameter. When set, the widget emits `hx-select` and retargets `hx-target` to the named fieldset instead of `#form_fields`. `hx-include` remains `#form_fields` so all field values are submitted and server-side dependent-field resolution continues to work correctly.
- No view or server-side template changes are required — the server still returns the full form HTML; HTMX's `hx-select` extracts only the matching element on the client side.

### Forms updated

| Form | Trigger field | Target fieldset |
|---|---|---|
| `InterfaceForm` | `mode` | `dot1q-switching` |
| `VMInterfaceForm` | `mode` | `dot1q-switching` |
| `CableForm` | `a_terminations_type` / `b_terminations_type` | `cable-side-a` / `cable-side-b` (hand-rolled template) |
| `ModuleTypeForm` | `profile` | `profile-attributes` (dynamic fieldset) |
| `ScopedForm` mixin | `scope_type` | `scope` (applied to `ClusterForm`, `WirelessLANForm`, `PrefixForm`) |
| `VLANGroupForm` | `scope_type` | `scope` |
| `ServiceForm` | `parent_object_type` | `service` |
| `CircuitTerminationForm` | `termination_type` | `circuit-termination` |
| `CircuitGroupAssignmentForm` | `member_type` | `circuit-group-assignment` |
| `VirtualCircuitTerminationForm` | `role` | `virtual-circuit-termination` |
| `TunnelCreateForm` | `termination1_type` / `termination2_type` | `tunnel-termination1` / `tunnel-termination2` |
| `TunnelTerminationForm` | `type` | `tunnel-termination` |
| `EventRuleForm` | `action_type` | `event-rule-action` |

### Forms intentionally left on full-form swap

Three forms keep `HTMXSelect()` without `hx_fieldset_id` because changing their trigger field adds or removes entire fieldsets dynamically — a partial swap would silently miss those additions. Each is annotated with an inline comment explaining the constraint:

- **`CustomFieldForm`** — `type` change dynamically adds/removes Validation, Related Object, and Choices fieldsets via `self.fieldsets = ...` in `__init__`
- **`DataSourceForm`** — `type` change conditionally appends a Backend Parameters fieldset
- **`VirtualMachineForm`** — `virtual_machine_type` populates defaults across both the Virtual Machine and Resources fieldsets

### Notes

- `FieldSet.id` must be a valid CSS identifier (start with a letter). The inline docstring notes this constraint.
- The `CableForm` change targets `<div>` elements in `dcim/htmx/cable_edit.html` directly (that form uses a hand-rolled HTMX template rather than the fieldset rendering system).
- Bulk-edit forms (`ScopedBulkEditForm`, `CircuitTerminationBulkEditForm`, `VLANGroupBulkEditForm`) already use an explicit `hx-select` targeting `#form_fields` for a different purpose and are unchanged.

## Test plan

- [x] `InterfaceForm`: change 802.1Q Mode — only the 802.1Q Switching fieldset updates; no full-form flash
- [x] `CableForm`: change A Side type — only A Side updates; B Side is unaffected
- [x] `CableForm`: change B Side type independently
- [x] `ModuleTypeForm`: switch profile — only Profile & Attributes fieldset re-renders; dynamically added attribute fields appear correctly
- [x] `utilities` test suite (314 tests): all pass
- [x] `dcim` form tests (30 tests): all pass

---

## Addendum — Review feedback (Jun 9)

Two naming improvements made in response to Jeremy's review:

### `FieldSet`: `fieldset_id` → `html_id`

The parameter name was redundant given it lives on a class already called `FieldSet`. Renamed to `html_id` throughout — rendering.py, form_helpers templatetag, render_fieldset.html, all call sites in forms, and tests.

### `HTMXSelect`: split `hx_target_id` / `hx_fieldset_id` into `hx_include_id` / `hx_target_id`

Jeremy asked why `hx_target_id` couldn't simply be overridden instead of needing a separate `hx_fieldset_id`. The reason: the original `hx_target_id` controlled *both* `hx-include` (the form data source) *and* the default `hx-target`. Overriding it would also scope `hx-include` to just the fieldset, dropping the rest of the form data and breaking server-side dependent-field resolution.

Fixed by separating the two concerns explicitly:

| Parameter | Controls | Default |
|---|---|---|
| `hx_include_id` | `hx-include` — which container's fields are submitted | `'form_fields'` |
| `hx_target_id` | `hx-target` + `hx-select` — which element to swap | `None` (falls back to `hx_include_id`) |

Jeremy's instinct is now correct: `HTMXSelect(hx_target_id='dot1q-switching')` does exactly what was expected — `hx-include` stays `#form_fields` while `hx-target` and `hx-select` narrow to `#dot1q-switching`.

All call sites, tests, and the `ScopedForm` runtime guard updated accordingly. The inline comments referencing the old `hx_fieldset_id` name updated to `hx_target_id`.


bctiemann:
@claude /review

bctiemann:
Addressed the two actionable items.

netbox/utilities/forms/rendering.py
```diff
@@ -25,10 +28,16 @@ class FieldSet:
     Parameters:
         items: An iterable of items to be rendered (one per row)
         name: The fieldset's name, displayed as a heading (optional)
+        fieldset_id: An HTML id for the rendered fieldset div, enabling HTMX partial swaps (optional).
+            Must be a valid CSS identifier: start with a letter, use only letters, digits, hyphens, underscores.
     """
-    def __init__(self, *items, name=None):
+    def __init__(self, *items, name=None, fieldset_id=None):
```
jeremystretch: What you think about renaming `fieldset_id` to something like `html_id` or `element_id`? The current name feels a bit redundant.

netbox/utilities/forms/widgets/select.py
```diff
@@ -61,15 +61,18 @@ class HTMXSelect(forms.Select):
     """
     Selection widget that will re-generate the HTML form upon the selection of a new option.
     """
-    def __init__(self, method='get', hx_url='.', hx_target_id='form_fields', attrs=None, **kwargs):
+    def __init__(self, method='get', hx_url='.', hx_target_id='form_fields', hx_fieldset_id=None, attrs=None, **kwargs):
```
jeremystretch: Can you elaborate on why the new `hx_fieldset_id` parameter is needed? It seems like we could just override `hx_target_id` instead, but I'm sure there's a reason not to.

netbox/utilities/forms/rendering.py
bctiemann: Agreed; done.

netbox/utilities/forms/widgets/select.py
bctiemann: Reworked this to separate the form data source (`hx-include`) and the targeted item (`hx-target`) more explicitly. Please see the addendum on the PR description.

netbox/utilities/forms/rendering.py
```diff
@@ -25,10 +28,16 @@ class FieldSet:
     Parameters:
         items: An iterable of items to be rendered (one per row)
         name: The fieldset's name, displayed as a heading (optional)
+        html_id: An HTML id for the rendered fieldset div, enabling HTMX partial swaps (optional).
+            Must be a valid CSS identifier: start with a letter, use only letters, digits, hyphens, underscores.
     """
-    def __init__(self, *items, name=None):
+    def __init__(self, *items, name=None, html_id=None):
+        if html_id is not None and settings.DEBUG:
```
jeremystretch, line 35: Do you think it makes sense to generate a slugified element ID by default for all FieldSets? It would obviate the need to define each one manually at the cost of being less explicit. We'd also probably want to attach a prefix to the ID (e.g. `fieldset-*`) to avoid conflicts. I'm fine either way.

netbox/circuits/forms/model_forms.py
```diff
@@ -395,7 +400,7 @@ class VirtualCircuitTerminationForm(NetBoxModelForm):
     )
     role = forms.ChoiceField(
         choices=VirtualCircuitTerminationRoleChoices,
-        widget=HTMXSelect(),
+        widget=HTMXSelect(hx_target_id='virtual-circuit-termination'),
```
jeremystretch: This doesn't appear to be needed; the form renders the same regardless of which role is selected. I suspect it was an errant copy-paste from CircuitTerminationForm.

netbox/core/forms/model_forms.py
```diff
@@ -29,6 +29,7 @@
 class DataSourceForm(PrimaryModelForm):
     type = forms.ChoiceField(
         choices=get_data_backend_choices,
+        # No hx_fieldset_id: changing type adds/removes the Backend Parameters fieldset entirely.
```
jeremystretch:
```suggestion
        # No hx_target_id: changing type adds/removes the Backend Parameters fieldset entirely.
```

netbox/extras/forms/model_forms.py
```diff
@@ -107,7 +107,8 @@ class Meta:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        # Mimic HTMXSelect()
+        # Mimic HTMXSelect() — no hx_fieldset_id because changing type adds/removes
```
jeremystretch:
```suggestion
        # Mimic HTMXSelect() — no hx_target_id because changing type adds/removes
```

netbox/virtualization/forms/model_forms.py
```diff
@@ -198,6 +198,8 @@ class VirtualMachineForm(TenancyForm, PrimaryModelForm):
         label=_('Type'),
         queryset=VirtualMachineType.objects.all(),
         required=False,
+        # No hx_fieldset_id: type change populates defaults across both the Virtual Machine
```
jeremystretch:
```suggestion
        # No hx_target_id: type change populates defaults across both the Virtual Machine
```

netbox/circuits/forms/model_forms.py
```diff
@@ -194,7 +194,8 @@ class CircuitTerminationForm(NetBoxModelForm):
     )
     termination_type = ContentTypeChoiceField(
         queryset=ContentType.objects.filter(model__in=CIRCUIT_TERMINATION_TERMINATION_TYPES),
-        widget=HTMXSelect(),
+        widget=HTMXSelect(hx_target_id='circuit-termination'),
+        required=False,
```
jeremystretch: Should this not be required? Seems odd to require `termination` but not `termination_type`.

netbox/ipam/forms/model_forms.py
```diff
@@ -809,7 +809,8 @@ class ServiceForm(PrimaryModelForm):
         FieldSet(
             'parent_object_type', 'parent', 'name',
             InlineFields('protocol', 'ports', label=_('Port(s)')),
-            'ipaddresses', 'description', 'tags', name=_('Application Service')
+            'ipaddresses', 'description', 'tags', name=_('Application Service'),
+            html_id='service',
```
jeremystretch, line 813: `html_id` needs to be defined on the fieldset under ServiceCreateForm below as well.

netbox/ipam/forms/model_forms.py
```diff
@@ -229,7 +229,7 @@ class PrefixForm(TenancyForm, ScopedForm, PrimaryModelForm):
         FieldSet(
             'prefix', 'status', 'vrf', 'role', 'is_pool', 'mark_utilized', 'description', 'tags', name=_('Prefix')
         ),
-        FieldSet('scope_type', 'scope', name=_('Scope')),
+        FieldSet('scope_type', 'scope', name=_('Scope'), html_id='scope'),
```
jeremystretch, line 232: `html_id` needs to be set on the fieldset under PrefixBulkAddForm as well.

jeremystretch (changes requested):

bctiemann (commented):
Thanks for the thorough review — addressed in the latest push:

- Dropped the errant `HTMXSelect` from `VirtualCircuitTerminationForm.role`
- Fixed the three stale `hx_fieldset_id` comments → `hx_target_id`
- Added `html_id='scope'` to `PrefixBulkAddForm`'s Scope fieldset and `html_id='service'` to `ServiceCreateForm`'s Application Service fieldset

Two open points worth a word:

**Auto-generating slugified IDs for FieldSets:** I lean toward keeping them explicit for this PR. Auto-generation is appealing, but it would require updating every `hx_target_id` value in every form *and* every HTMX `hx-target` attribute in templates to use the new `fieldset-*` prefix — a wider blast radius than seems right for a follow-up. Happy to do it as a separate cleanup if you'd prefer consistency over explicitness.

**`termination_type required=False`:** The underlying model field is a nullable FK (`null=True, blank=True` on `CircuitTermination.termination_type`), so a circuit termination can legitimately be saved with no termination target at all. The `required=False` corrects what was a latent mismatch with the model — the form was previously `required=True` by default even though the DB allows null. The form's `clean()` already handles the "type selected but no termination" case; the only new behaviour is allowing both to be blank.

netbox/circuits/forms/model_forms.py
bctiemann: See main comment

netbox/utilities/forms/rendering.py
bctiemann: See main comment

netbox/circuits/forms/model_forms.py
```diff
@@ -412,7 +416,10 @@ class VirtualCircuitTerminationForm(NetBoxModelForm):
     )
 
     fieldsets = (
-        FieldSet('virtual_circuit', 'role', 'interface', 'description', 'tags'),
+        FieldSet(
+            'virtual_circuit', 'role', 'interface', 'description', 'tags',
+            html_id='virtual-circuit-termination',
```
jeremystretch: We can remove `html_id` too.
```suggestion
```

netbox/circuits/forms/model_forms.py
jeremystretch: IMO we should revert this, for two reasons:

1. The `termination` field below is still required, which is confusing (my point above).
2. It's out of scope for the feature being implemented.

jeremystretch (changes requested):

bctiemann:
Applied both remaining suggestions.

jeremystretch (approved):

PR #22348, opened 2026-06-02
Closes #22180: Validate scripts added via a data source

### LINKED ISSUE #22180: Adding Custom Scripts via Data Source bypasses validate_script_content creating a broken script that cannot be updated.
### NetBox Edition

NetBox Community

### NetBox Version

4.5.10 (also reproduced on 4.6.0)

### Python Version

3.13

### Steps to Reproduce

1. Create a bad script in a datasource.
e.g.

```
import imnotarealmoduleicreateerrors

class TestBadBlank(Script):

    class Meta(Script.Meta):
        name = "Test Bad Blank"
        author = "Your Name"
        description = "A blank script to test bad functionality."

    def run(self):
        self.log_error("Testing bad blank...")
```
2. Add script via Data Source.
Image

```
2026-05-13 09:58:24,282 netbox.data_backends ERROR: Failed to load script: test_badblank error: No module named 'imnotarealmoduleicreateerrors'
2026-05-13 09:58:24,284 netbox.views.ObjectEditView INFO: Created script module test_badblank (PK: 131)
```

3.  Remove bad import from script, and resync datasource.

### Expected Behavior

Incorrect script should be blocked by validate_script_content and module should not be created.
Alternatively, Corrected script should then show up on scripts page.

### Observed Behavior

Script remains Missing 
Attempting to re-add a new script module errors  
`A script module with this file path already exists (scripts/test_badblank.py). `

jnovinger:
### Closes: #22180

Scripts added via a data source bypassed `validate_script_content`. `ScriptFileForm.clean()` only validated an uploaded file, so selecting an existing data file skipped validation entirely. A script that failed to load was then committed as a broken module that showed as missing and could not be corrected, since re-adding it tripped the `(file_root, file_path)` uniqueness constraint.

`ScriptFileForm.clean()` now validates a selected data file's content the same way it already validates an upload, raising a form error before any module is created. A corrected script syncs and appears normally.

The fix is scoped to the interactive form path (the path reported in the issue). The REST API has no data-source creation path for script modules, so it needs no change. Validation of scripts created via background data-source synchronization is tracked separately.

### DIFF
--- a/netbox/extras/forms/scripts.py
+++ b/netbox/extras/forms/scripts.py
@@ -90,6 +90,15 @@ class ScriptFileForm(ManagedFileForm):
                 raise forms.ValidationError(
                     _("Error loading script: {error}").format(error=e)
                 )
+        elif data_file := self.cleaned_data.get('data_file'):
+            # Validate scripts synced from a data source as well, to avoid creating a broken
+            # script module that cannot be loaded or corrected (see #22180)
+            try:
+                validate_script_content(data_file.data, data_file.path)
+            except Exception as e:
+                raise forms.ValidationError(
+                    _("Error loading script: {error}").format(error=e)
+                )
 
         return self.cleaned_data
 
--- a/netbox/extras/forms/scripts.py
+++ b/netbox/extras/forms/scripts.py
@@ -92,7 +92,7 @@ class ScriptFileForm(ManagedFileForm):
                 )
         elif data_file := self.cleaned_data.get('data_file'):
             # Validate scripts synced from a data source as well, to avoid creating a broken
-            # script module that cannot be loaded or corrected (see #22180)
+            # script module that cannot be loaded or corrected
             try:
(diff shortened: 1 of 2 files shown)


jnovinger:

jnovinger:
> `validate_script_content` calls `exec()`, so any module-level side effects in a script file run synchronously during the form `clean()` step — in the web-server process. [...] a user with write access to the backing data source could craft a file that executes in the web process when any admin opens the *Add Script Module* form and selects that file.

Correct, and the trust boundary is unchanged. Write access to a scripts data source already implies arbitrary code execution in NetBox: that's what running a script does. Validating via `exec()` at form time only moves that execution slightly earlier and into the web process instead of the RQ worker. The existing upload path already execs in the web process during `clean()`, so this PR adds no new capability.

We considered it and are accepting it as-is for this narrowly scoped fix.

bctiemann (approved):

PR #22360, opened 2026-06-02
Fixes #21338: Fix missing connected endpoint data in Interface Webhooks during Cable creation

### LINKED ISSUE #21338: Interface webhook data does not include connected endpoints nor link_peers when a connection is made
### NetBox Edition

NetBox Community

### NetBox Version

v4.4.10

### Python Version

3.12

### Steps to Reproduce

1. In the Netbox web UI, create a webhook `localhost testing webhook` using URL: `127.0.0.1:9000`, Method: `POST` and Content type `application/json`
2. Create an event rule `localhost testing event rule` with Object types `DCIM > Interface` and Event Types `Object updated`, Action type `webhook` and Webhook `localhost testing webhook` (the one created in step one)
3. SSH or exec to netbox underlying OS, activate virtual environment and run the webhook listener with `python3 netbox/manage.py webhook_receiver`
4. Back in the Netbox web UI, create a device `mydevice1` off some device type. Make sure there is an interface on the device, otherwise create one. Select a Device role and a Site.
5. Create a device `mydevice2` off some device type. Make sure there is an interface on the device, otherwise create one. Select a Device role and a Site.
6. On an interface on `mydevice1`, create a cable connecting to an interface on `mydevice2`
7. On the webhook listener, observe the output. There should be two completed requests, one for each interface at the cable endpoints.

### Expected Behavior

- On `request #1`, `connected_endpoints` and `link_peers` will be lists containing json data describing the interface from `mydevice2`
- On `request #2`, `connected_endpoints` and `link_peers` will be lists containing json data describing the interface from `mydevice1`

### Observed Behavior

- On `request #1`, `connected_endpoints` is `null` and `link_peers` is `[]`
- On `request #2`. `connected_endpoints` is `null` and `link_peers` is a list containing json data describing the interface from `mydevice1`

pheus:
### Fixes: #21338

Fixes stale in-memory `_path` state for cabled endpoints during cable creation.

When a cable is created, endpoint update events can be queued before path tracing has populated the denormalized CablePath reference. This could cause webhook payloads to omit connected endpoint details. The `path` accessor now refreshes `_path` from the database when an endpoint is already linked but has no path cached, ensuring payloads include the correct connection data without adding webhook-specific logic.

### DIFF
--- a/netbox/dcim/models/device_components.py
+++ b/netbox/dcim/models/device_components.py
@@ -384,12 +384,27 @@ class PathEndpoint(models.Model):
         a stale in-memory `_path` relation while the database already points to
         a different CablePath (or to no path at all).
 
-        If the cached relation points to a CablePath that has just been
-        deleted, refresh only the `_path` field from the database and retry.
-        This keeps the fix cheap and narrowly scoped to the denormalized FK.
+        Two stale cases are repaired by refreshing only the `_path` field
+        from the database:
+
+        1. The endpoint is linked (by cable or wireless link) but `_path` is
+           unset, because the instance was loaded before its path was traced
+           (e.g. while queued for event serialization during link creation).
+        2. The cached relation points to a CablePath row that has just been
+           deleted.
+
+        Repairing case 1 costs one query per access for a linked endpoint
+        whose path is genuinely absent in the database. That state is
+        transient outside of tracing failures, so no result caching is
+        attempted here.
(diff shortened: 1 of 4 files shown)


pheus:

pheus:

jeremystretch (approved):
Nice work tracking this down!

PR #22363, opened 2026-06-02
Closes #22351: Add jinja2_filters plugin hook and get_jinja2_context() for config template extensibility

### LINKED ISSUE #22351: Add jinja2_filters plugin hook and get_jinja2_context() to PluginConfig for config template extensibility
## NetBox version

v4.6.x

## Feature type

New functionality

## Proposed functionality

Add two new extension points to `PluginConfig` that allow plugins to participate in Jinja2 config template rendering.

### 1. `jinja2_filters` plugin resource

A `jinja2_filters` class attribute on `PluginConfig`, mirroring the existing `template_extensions` pattern. It points to a module path whose exported `filters` dict maps filter names to callables. The filters are registered via a new `register_jinja2_filters()` function into `registry['plugins']['jinja2_filters']` and merged into every `render_jinja2()` call, between `DEFAULT_JINJA2_FILTERS` and the instance-level `JINJA2_FILTERS` setting. Instance config always wins.

Default auto-discovery path: `jinja2_env.filters` (consistent with how `template_extensions` auto-discovers from `template_content.template_extensions`).

### 2. `PluginConfig.get_jinja2_context()` method

A new method on `PluginConfig` with a default `return {}` implementation. Subclasses override it to inject additional variables into the Jinja2 template context. `RenderTemplateMixin.get_context()` calls it on every installed plugin after the standard `ObjectType.objects.public()` model-class population and before the caller-supplied context (so callers can still override plugin-injected keys).

### Concrete changes

- `netbox/plugins/registration.py`: `register_jinja2_filters(filters: dict)` added to `__all__`
- `netbox/plugins/__init__.py`: `'jinja2_filters': {}` registry slot; `jinja2_filters = None` class attribute; `'jinja2_env.filters'` default resource path; `get_jinja2_context()` base method; loading in `ready()`
- `utilities/jinja2.py`: `render_jinja2()` merges `registry['plugins']['jinja2_filters']` between defaults and instance config
- `extras/models/mixins.py`: `get_context()` iterates installed plugins and calls `get_jinja2_context()`

## Use case

Config templates (used to render device configurations for DCIM automation) currently have access to the device/VM object, merged config context data, and all public model classes via `ObjectType.objects.public()`. However, plugins that manage their own model data — whether dynamically-generated models, proxy types, or computed aggregates — have no supported path to surface that data to config templates.

Today, exposing plugin-managed data requires a site administrator to manually add Python functions to `JINJA2_FILTERS` in `configuration.py`. This means:

- The feature cannot be delivered by the plugin itself; it requires per-instance admin action on every installation
- Template authors must know internal implementation details rather than using plugin-defined abstractions
- There is no documentation or discoverability path — users must read plugin source code to know what to add to their config

With `jinja2_filters`, a plugin registers a filter once in its own codebase and it is available in every config template on every installation without any instance-level configuration. With `get_jinja2_context()`, a plugin can inject a namespace object or named variables that give template authors a clean, documented interface:

```jinja2
{# jinja2_filters hook — filter syntax #}
{% for peer in 'BGPPeer' | bgp_peers_for_device(device) %}

{# get_jinja2_context() hook — namespace syntax #}
{% for peer in bgp.peers_for(device) %}
```

Both patterns are used widely in the Jinja2 ecosystem (Ansible, Salt, Nautobot). NetBox's current gap makes it harder for the plugin ecosystem to deliver the "source of truth → config generation" workflow that is a primary use case for the platform.

The `serializer_resolver` hook (added in v4.6) establishes the precedent that plugin API surfaces can be extended incrementally without breaking changes; these two hooks follow the same pattern and have similar scope.

The proximate driver is the netbox-custom-objects plugin, whose dynamically-generated model classes have machine-generated internal names (`Table3Model`). Even though they are reachable via `ObjectType.objects.public()`, template authors cannot use them ergonomically without a plugin-provided abstraction layer over the top.

## Database changes

None.

## External dependencies

None. The implementation uses only existing Python stdlib, Django, and Jinja2 facilities already present in NetBox.

bctiemann:
## Summary

- Adds `jinja2_filters` as a new `PluginConfig` resource, mirroring the `template_extensions` pattern. Plugins export a `filters` dict (auto-discovered from `jinja2_env.filters`) that is registered into `registry['plugins']['jinja2_filters']` and merged into every `render_jinja2()` call. Instance-level `JINJA2_FILTERS` always takes precedence.
- Adds `PluginConfig.get_jinja2_context()` — a method with a default `return {}` implementation that subclasses override to inject named variables into every `ConfigTemplate` render context. Called from `RenderTemplateMixin.get_context()` after the standard `ObjectType.objects.public()` population and before caller-supplied context.

## Test plan

- [ ] `extras.tests.test_models` passes unchanged (no regressions to existing config template or render behaviour)
- [ ] A plugin that exports `jinja2_env.filters = {'my_filter': fn}` has `my_filter` available in config templates after installation
- [ ] A plugin that overrides `get_jinja2_context()` returning `{'ns': obj}` has `ns` available as a top-level variable in config templates
- [ ] Instance `JINJA2_FILTERS` setting overrides a plugin-registered filter of the same name

> **Note:** Replaces #22352, which was inadvertently auto-merged when the head branch briefly matched the base.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


bctiemann:
See https://github.com/netbox-community/netbox/pull/22352 for Copilot review results.

arthanson (approved):

PR #22364, opened 2026-06-02
Fixes #22357: Remove unused `local_context_data` field from dcim.Module

### LINKED ISSUE #22357: dcim.Module should not inherit from ConfigContextModel
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

Inspect the definition of the `dcim.Module` model in `netbox/dcim/models/modules.py`.

### Expected Behavior

The `Module` class should inherit only from `TrackingModelMixin` and `PrimaryModel`.

### Observed Behavior

It also inherits from `ConfigContextModel` despite having no support for config context functionality. This is harmless and was likely an oversight during the initial implementation, but should be cleaned up to remove the extraneous `local_context_data` field from the model's database table.

jeremystretch:
### Fixes: #22357

- Remove ConfigContextModel from the list of parent classes for Module
- Add a migration to remove the database table field

### DIFF
--- a/netbox/dcim/migrations/0237_module_remove_local_context_data.py
+++ b/netbox/dcim/migrations/0237_module_remove_local_context_data.py
@@ -1,0 +1,15 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dcim', '0236_moduletype_component_counts'),
+    ]
+
+    operations = [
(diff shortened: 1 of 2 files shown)


jeremystretch:

jnovinger (approved):

PR #22366, opened 2026-06-02
Fixes #22324: Fix GraphQL filtering of Custom Field Choice Set extra choices

### LINKED ISSUE #22324: `CustomFieldChoiceSet.extra_choices` GraphQL filter uses a flat array lookup for a nested array field
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Create a `CustomFieldChoiceSet` with `extra_choices`, for example via the REST API:

   ```http
   POST /api/extras/custom-field-choice-sets/
   
   {
     "name": "Choice Set 1",
     "extra_choices": [["1A", "Choice 1A"], ["2B", "Choice 2B"]]
   }
   ```
2. Query GraphQL using the currently advertised filter shape:

   ```graphql
   query {
     custom_field_choice_set_list(filters: { extra_choices: { contains: ["1A"] } }) {
       id
       name
     }
   }
   ```

   This validates because `CustomFieldChoiceSetFilter.extra_choices` is currently declared as `StringArrayLookup`, whose `contains` operand is a flat list of strings.
3. Try the shape that matches the stored model value:

   ```graphql
   query {
     custom_field_choice_set_list(
       filters: { extra_choices: { contains: [["1A", "Choice 1A"]] } }
     ) {
       id
       name
     }
   }
   ```

### Expected Behavior

The GraphQL schema should not expose a filter shape that cannot represent the underlying model field shape.

Any of these outcomes would be reasonable:

* expose a nested-list filter for `extra_choices`, so clients can pass nested operands such as `contains: [["1A", "Choice 1A"]]`;
* expose explicit value/label filters, if the intended use case is value-only or label-only filtering; or
* remove `extra_choices` from the GraphQL filter type until a meaningful filter shape is supported.

### Observed Behavior

The current schema only accepts a flat string array for `extra_choices.contains`.

A schema-following query such as:

```graphql
filters: { extra_choices: { contains: ["1A"] } }
```

does not represent the stored `extra_choices` value shape, which is a list of `[value, label]` pairs.

A query using the stored value shape:

```graphql
filters: { extra_choices: { contains: [["1A", "Choice 1A"]] } }
```

fails GraphQL validation with an error similar to:

```text
String cannot represent a non string value: ["1A", "Choice 1A"]
```

The issue also surfaces in the GraphQL filter test framework when the test attempts to use the model’s actual nested `extra_choices` value as the lookup operand:

```text
FAIL: test_graphql_filter_objects
(extras.tests.test_api.CustomFieldChoiceSetTestCase.test_graphql_filter_objects)
(filter='extra_choices__contains')

AssertionError: 'errors' unexpectedly found in
{'data': None, 'errors': [{'message': 'String cannot represent a non string value: ["1A", "1A"]', ...}]}
```

### Root Cause

`CustomFieldChoiceSetFilter.extra_choices` is declared as a flat string-array lookup:

```python
extra_choices: Annotated['StringArrayLookup', strawberry.lazy('netbox.graphql.filter_lookups')] | None = (
)
```

`StringArrayLookup` is `ArrayLookup[str]`, so its lookup operands are flat `list[str]` values.

However, the model field is a nested array:

```python
extra_choices = ArrayField(
    ArrayField(
        base_field=models.CharField(max_length=100),
        size=2,
    ),
    blank=True,
    null=True,
)
```

So the model stores values shaped like:

```python
[
    ["1A", "Choice 1A"],
    ["2B", "Choice 2B"],
]
```

The filter declaration therefore does not match the model field’s Python/storage shape.

### Proposed Fix

Preferred minimal fix: replace `StringArrayLookup` with a lookup type that accepts a nested string list, for example a `NestedStringArrayLookup` / `ArrayLookup[list[str]]` equivalent, and use that for `CustomFieldChoiceSetFilter.extra_choices`.

That would make the schema match the field shape and allow clients to express nested operands such as:

```graphql
contains: [["1A", "Choice 1A"]]
```

One caveat: PostgreSQL array containment has element-based semantics, so this should be treated primarily as a schema/type-shape fix. If NetBox needs strict “value + label in the same pair” matching, or value-only/label-only matching, a dedicated filter implementation would be clearer.

Possible dedicated follow-up filters:

* `extra_choices_value`
* `extra_choices_label`

Those could use a purpose-built PostgreSQL expression/helper for the `[value, label]` pairs rather than relying on the generic array lookup.

Fallback: if filtering on `extra_choices` is not a common need, remove it from the GraphQL filter type to avoid advertising a misleading schema shape.

### Related

This is another GraphQL filter declaration mismatch surfaced by the filter test framework in the `15569-add-better-tests-for-graphql-filtering-and-lookup` branch.

Related issues:

* netbox-community/netbox#22286
* netbox-community/netbox#22287

pheus:
### Fixes: #22324

`CustomFieldChoiceSet.extra_choices` stores a list of `[value, label]` pairs, but its GraphQL filter was declared as a flat `StringArrayLookup`. That advertised a shape the data can never match (`contains: ["1A"]`), and passing the value as actually stored (`contains: [["1A", "Choice 1A"]]`) failed schema validation entirely.

Rather than reshape the filter to mirror the raw two-dimensional storage, this PR exposes the filter people reach for in practice: matching by choice **value**. `extra_choices` now accepts `contains` (is there a choice with this value?) and `length` (how many choices are defined?). Value-only filtering is one of the resolutions called out in the issue, and it keeps the schema intuitive without leaking the internal `[value, label]` layout into the API. Matching against the value element means comparing the first item of each pair, which the built-in array lookups can't express, so we add a `choice_value` lookup backed by a small PostgreSQL array slice (`= ANY(field[:][1:1])`).

**Why a dedicated field type.** That lookup only makes sense for choice-set columns. Registering it on `ArrayField` would attach `choice_value` to *every* `ArrayField` in NetBox, where it's meaningless and misleading. To keep it scoped, we introduce `ChoiceSetField` — a thin `ArrayField` subclass that fixes the `[value, label]` shape and carries the `choice_value` lookup on itself alone. Every other array field is left untouched (there's a test covering exactly that).

The rest is supporting work: a migration switching `extra_choices` to `ChoiceSetField` (no change to the underlying column — it stays `varchar(100)[]`), and a one-line change in the test helper so it recognizes `ArrayField` subclasses, not just `ArrayField` itself.

### DIFF
--- a/netbox/extras/fields.py
+++ b/netbox/extras/fields.py
@@ -1,4 +1,10 @@
-from django.db.models import TextField
+from django.contrib.postgres.fields import ArrayField
+from django.db.models import CharField, TextField
+
+__all__ = (
+    'CachedValueField',
+    'ChoiceSetField',
+)
 
 
 class CachedValueField(TextField):
@@ -6,3 +12,18 @@ class CachedValueField(TextField):
     Currently a dummy field to prevent custom lookups being applied globally to TextField.
     """
     pass
+
+
+class ChoiceSetField(ArrayField):
+    """
+    An ArrayField of two-element [value, label] string pairs representing custom field choices.
+    """
+    def __init__(self, **kwargs):
+        kwargs['base_field'] = ArrayField(base_field=CharField(max_length=100), size=2)
+        super().__init__(**kwargs)
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        # base_field is fixed by __init__ and omitted from migrations
+        del kwargs['base_field']
+        return name, path, args, kwargs
--- a/netbox/extras/graphql/filter_lookups.py
+++ b/netbox/extras/graphql/filter_lookups.py
@@ -1,0 +1,30 @@
+import strawberry
+import strawberry_django
+from django.db.models import Q, QuerySet
+from strawberry.directive import DirectiveValue
+from strawberry.types import Info
+
+__all__ = (
+    'ExtraChoicesLookup',
+)
+
+
+@strawberry.input(
+    one_of=True,
+    description='Lookup for extra choices defined on a choice set. Only one of the lookup fields can be set.',
+)
+class ExtraChoicesLookup:
+    contains: str | None = strawberry.field(
+        default=strawberry.UNSET, description='Has an extra choice with this value'
+    )
+    length: int | None = strawberry.field(
+        default=strawberry.UNSET, description='Number of extra choices'
+    )
+
+    @strawberry_django.filter_field
+    def filter(self, info: Info, queryset: QuerySet, prefix: DirectiveValue[str] = '') -> tuple[QuerySet, Q]:
+        if self.contains is not strawberry.UNSET and self.contains is not None:
+            return queryset, Q(**{f'{prefix}choice_value': self.contains})
+        if self.length is not strawberry.UNSET and self.length is not None:
+            return queryset, Q(**{f'{prefix}len': self.length})
+        return queryset, Q()
--- a/netbox/extras/graphql/filters.py
+++ b/netbox/extras/graphql/filters.py
@@ -23,6 +23,7 @@ if TYPE_CHECKING:
         SiteFilter,
         SiteGroupFilter,
     )
+    from extras.graphql.filter_lookups import ExtraChoicesLookup
     from netbox.graphql.enums import ColorEnum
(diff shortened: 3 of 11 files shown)


pheus:

pheus:

arthanson:
Just a note, this is the query that now works:
```
query {
  custom_field_choice_set_list(filters: { extra_choices: { contains: "1A" } }) {
    id
    name
  }
}
```

arthanson (approved):
nice!

PR #22368, opened 2026-06-02
Fixes #22358: Remove broken and unused function get_0u_devices

### LINKED ISSUE #22358: fix broken get_0u_devices() function
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.13

### Steps to Reproduce

While working on https://github.com/netbox-community/netbox/pull/22295 I noticed, that the function `get_0u_devices()` in `netbox/dcim/models/racks.py` is broken and always returns an empty set.

The current implementation is: https://github.com/netbox-community/netbox/blob/742f4b433048e9aa63bf644e0cb4a97737c408b5/netbox/dcim/models/racks.py#L641-L643

Because of the Model Definition, it's not possible to save a 0 as position, therefore, this QuerySet is always empty.

### Expected Behavior

Function should return a valid QuerySet containing the requested information.

I suggest to rewrite this function:
```python
def get_0u_devices(self):
    return self.devices.filter(device_type__u_height=0)
```

Additionally adding a test case, to make sure the function works:
```python
    def test_0u_devices(self):
        """
        Check that get_0u_devices() successfully return 0U devices.
        """
        site = Site.objects.first()
        rack = Rack.objects.first()

        device1 = Device.objects.create(
            name='Device 1',
            role=DeviceRole.objects.first(),
            device_type=DeviceType.objects.get(u_height=0),
            site=site,
            rack=rack
        )
        device1.full_clean()
        device1.save()
        device2 = Device.objects.create(
            name='Device 2',
            role=DeviceRole.objects.first(),
            device_type=DeviceType.objects.get(u_height=0),
            site=site,
            rack=rack
        )
        device2.full_clean()
        device2.save()

        self.assertQuerySetEqual(rack.get_0u_devices(), [device1, device2])
```

**An Alternative** would be, to remove this function altogether. There doesn't seem to be any place in the code, which uses this function.

### Observed Behavior

Function returns empty Queryset.

---

If this issue is accepted, I can take the task and make a PR out of the code above.

mburggraf:
### Closes: #22358

Removed the broken function.

### DIFF
--- a/netbox/dcim/models/racks.py
+++ b/netbox/dcim/models/racks.py
@@ -638,9 +638,6 @@ class Rack(ContactsMixin, ImageAttachmentsMixin, TrackingModelMixin, RackBase):
 
         return elevation.render(face)
 
-    def get_0u_devices(self):
-        return self.devices.filter(position=0)
-
     def get_utilization(self):
         """
         Determine the utilization rate of the rack and return it as a percentage. Occupied and reserved units both count


pheus (approved):
Thanks for cleaning this up!

PR #22377, opened 2026-06-03
Closes #17598: Add bulk creation for VLANs

### LINKED ISSUE #17598: Bulk create of VLANs
### NetBox version

v4.1.1

### Feature type

New functionality

### Proposed functionality

Create multiple VLANs if a range (i.e 3100-3299) is given in "VLAN ID" of "/ipam/vlans/add/"

### Use case

If you want to create a VLAN-pool for some orchestrator to use for dynamic VLAN provisioning you could use a VLAN Group for that, but you still have to create the containing VLANs to see in the GUI, that those VLANs are not "free-to-use". These could have the name pattern "Name-VLAN_ID"

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Closes: #17598

This adds bulk VLAN creation using NetBox’s existing bulk-create pattern.

Users can now create multiple VLANs from a comma-separated list or range of VLAN IDs, for example `100,200-210,3100-3299`. The VLAN name field also supports a `{vid}` placeholder, allowing generated names such as `Customer-3100`, `Customer-3101`, and so on.

The implementation keeps the single VLAN create form unchanged and reuses the generic `BulkCreateView` flow, with a small generic extension for applying generated pattern values to additional form fields.

### DIFF
--- a/docs/models/ipam/vlan.md
+++ b/docs/models/ipam/vlan.md
@@ -2,6 +2,12 @@
 
 A Virtual LAN (VLAN) represents an isolated layer two domain, identified by a name and a numeric ID (1-4094) as defined in [IEEE 802.1Q](https://en.wikipedia.org/wiki/IEEE_802.1Q). VLANs are arranged into [VLAN groups](./vlangroup.md) to define scope and to enforce uniqueness.
 
+## Bulk Creation
+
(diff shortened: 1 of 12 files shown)


pheus:

pheus:
@claude /review

Point 1, 2 and 4 are addressed.

For the batch-size concern, I agree it’s worth considering, but I’m inclined not to add a hard limit in this PR unless review prefer one. The possible VLAN range is bounded by 1-4094, and adding a generic limit to `BulkCreateView` could affect existing prefix/IP address bulk-create behavior. If we do want a limit, I’d suggest adding it as a generic opt-in, e.g. `max_pattern_values = None`, and setting it only where desired.

jnovinger:
@pheus , this PR looks good, but appears to have a merge conflict.

pheus:
@jnovinger Thanks! I've rebased the branch.

jnovinger (approved):

Issue #22287, opened 2026-05-26
GraphQL `EventRuleFilter.action_object_type` typed as string lookup against a ContentType FK


pheus:
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Query the GraphQL endpoint:
  ```graphql
  { event_rule_list(filters: {action_object_type: {i_contains: "device"}}) { id } }
  ```

### Expected Behavior

Either the schema rejects the lookup as inapplicable to a ForeignKey, or the filter returns matching event rules.

### Observed Behavior

Unsupported lookup 'icontains' for ForeignKey or join on the field not permitted.
The schema accepts the input (the type validates) but execution raises at the ORM layer.

Issue #22286, opened 2026-05-26
GraphQL ConfigContextProfileFilter name and description should be Optional


pheus:
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Query the GraphQL endpoint for `config_context_profile_list` with any single filter, e.g.:
  ```graphql
  { config_context_profile_list(filters: {id: {exact: 1}}) { id } }
  ```
2. Observe the response.

### Expected Behavior

Filter accepted; objects matching the filter returned.

### Observed Behavior

Field 'ConfigContextProfileFilter.description' of required type 'StrFilterLookup!' was not provided.
Field 'ConfigContextProfileFilter.name' of required type 'StrFilterLookup!' was not provided.

PR #22384, opened 2026-06-04
Closes #15569: Auto-generate GraphQL filter tests for API test cases

### LINKED ISSUE #15569: Add better tests for GraphQL filtering and lookup

pheus:
### Closes: #15569

This extends `APIViewTestCases.GraphQLTestCase` to automatically generate GraphQL filter regression tests for model API test cases.

The goal is to make filter coverage harder to miss by default. For each model test case, the framework discovers the conventional GraphQL filter class, inspects its declared lookup annotations, and compares GraphQL filter results against the equivalent ORM queryset.

This also adds a loud-fail gate so models with no generated or explicit GraphQL filter tests fail clearly instead of silently losing coverage.

The PR also introduces two declarative extension points:

- `graphql_filter_tests` for explicit field-level filter cases
- `graphql_query_tests` for nested or full-query GraphQL cases

`SiteTestCase` and `LocationTestCase` include worked examples, including coverage for nested filtered fields such as related locations and tags.

#### What it surfaced

While developing this, the new framework caught two real GraphQL filter declaration issues. Those were fixed separately and have already been merged:

- #22286
- #22287
- #22324

#### Query count baseline

`site:api_list_objects` 16 -> 17: the Site fixture now assigns a tenant to Site 1. The API list view prefetches serializer FK relations dynamically (`prefetch_related`; there is no `select_related` path), and Django skips a prefetch query entirely when every FK value is NULL (the empty `__in` lookup short-circuits via `EmptyResultSet`). With one tenant populated, the tenant prefetch issues one real query. The bump is the prefetch becoming active, not a regression.

#### Verification

- `utilities.tests.test_api_graphql`: 9 OK
- All apps' `test_api` modules: 3036 OK / 8 skipped, 0 failures
- Full suite with `manage.py test netbox/ --parallel`: 9927 OK / 13 skipped, 0 failures
- `ruff check` clean on all changed files

#### Follow-up work

A few improvements were discussed and intentionally left for follow-up PRs:

- Split the GraphQL test framework out of `utilities/testing/api.py`
- Add an explicit `BaseFilterLookup` enum-literal test
- Add nested OR/NOT combinator tests

### DIFF
--- a/netbox/dcim/tests/query_counts.json
+++ b/netbox/dcim/tests/query_counts.json
@@ -72,7 +72,7 @@
   "rearporttemplate:api_list_objects": 12,
   "region:api_list_objects": 13,
   "region:list_objects_with_permission": 20,
-  "site:api_list_objects": 16,
+  "site:api_list_objects": 17,
   "site:list_objects_with_permission": 22,
   "sitegroup:api_list_objects": 13,
   "sitegroup:list_objects_with_permission": 20,
--- a/netbox/dcim/tests/test_api.py
+++ b/netbox/dcim/tests/test_api.py
@@ -20,6 +20,8 @@ from users.models import ObjectPermission, Token, User
 from utilities.testing import (
     APITestCase,
     APIViewTestCases,
+    GraphQLFilterTest,
+    GraphQLQueryTest,
     create_test_device,
     create_test_nat_ip_pair,
     disable_logging,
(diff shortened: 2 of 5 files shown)


pheus:

pheus:
@claude /review
Item 1: is noted in the PR text.
Item 2 + 3 are fixed.
Item 4 (sys.modules mutation): one-line comment added noting the process-based --parallel assumption.

bctiemann (approved):

PR #22385, opened 2026-06-04
Fixes #22237: Prevent direct access to TableConfig create view

### LINKED ISSUE #22237: Server error when opening the standalone "Add Table Configuration" page
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.0

### Python Version

3.12

### Steps to Reproduce

1. Log in as a user with the `extras.add_tableconfig` permission.
2. Browse directly to the standalone Add URL, without going through a  <br>"Save table configuration" button on an existing table list view:

   ```text
   GET /extras/table-configs/add/
   ```

### Expected Behavior

The Add page should load without a server error. The form should either prompt the user to first choose an `object_type` and `table` (then populate the column selectors based on the chosen table), or redirect the user back to the parent table's list view where the Save Configuration action lives. Forms that derive widget choices from a related field should guard the related lookup so that the form is still constructible before the user has  <br>picked a value.

### Observed Behavior

The view returns HTTP 500 and the server log shows `core.models.object_types.ObjectType.DoesNotExist` raised during form rendering:

```text
core.models.object_types.ObjectType.DoesNotExist: ObjectType matching query does not exist.
```

The traceback ends in `TableConfigForm.__init__`:

```python
# netbox/extras/forms/model_forms.py
class TableConfigForm(forms.ModelForm):
    ...
    def __init__(self, data=None, *args, **kwargs):
        super().__init__(data, *args, **kwargs)

        object_type = ObjectType.objects.get(pk=get_field_value(self, 'object_type'))  # ← raises
        model = object_type.model_class()
        ...
```

The lookup is unconditional. When the form has no bound POST data, no initial data, and no instance, `get_field_value(self, 'object_type')` returns `None` and the `get()` call raises `DoesNotExist`.

A small guard preserves current behavior for the "Save table configuration" flow while making the form usable from the standalone Add URL:

```python
def __init__(self, data=None, *args, **kwargs):
    super().__init__(data, *args, **kwargs)

    object_type_pk = get_field_value(self, 'object_type')
    if not object_type_pk:
        return
    object_type = ObjectType.objects.get(pk=object_type_pk)
    ...
```

This was surfaced while adding standard test coverage for `extras.tableconfig` in netbox-community/netbox#22088. The housekeeping PR currently exempts the affected view coverage with a self-cleaning marker that will prompt removal once this issue is fixed.

pheus:
### Closes: #22237

Add GET handler to TableConfigEditView that redirects users to home with a warning if they attempt to access the create form directly without required object_type and table parameters from a source list view.

### DIFF
--- a/netbox/extras/forms/bulk_edit.py
+++ b/netbox/extras/forms/bulk_edit.py
@@ -212,7 +212,7 @@ class SavedFilterBulkEditForm(ChangelogMessageMixin, OwnerMixin, BulkEditForm):
     nullable_fields = ('description',)
 
 
-class TableConfigBulkEditForm(BulkEditForm):
+class TableConfigBulkEditForm(ChangelogMessageMixin, BulkEditForm):
     pk = forms.ModelMultipleChoiceField(
         queryset=TableConfig.objects.all(),
         widget=forms.MultipleHiddenInput
--- a/netbox/extras/forms/model_forms.py
+++ b/netbox/extras/forms/model_forms.py
@@ -402,7 +402,7 @@ class SavedFilterForm(ChangelogMessageMixin, OwnerMixin, forms.ModelForm):
         super().__init__(*args, initial=initial, **kwargs)
 
 
-class TableConfigForm(forms.ModelForm):
+class TableConfigForm(ChangelogMessageMixin, forms.ModelForm):
     object_type = ContentTypeChoiceField(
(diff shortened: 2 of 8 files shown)


pheus:

pheus:
@claude /review I've addressed your concern.

jeremystretch (approved):

PR #22390, opened 2026-06-04
Closes #18987: Preserve restricted related values on edit forms

### LINKED ISSUE #18987: Loss of Object Properties When Editing with Restricted View Permissions
### Deployment Type

Self-hosted

### NetBox Version

v4.2.5

### Python Version

3.12

### Steps to Reproduce

As Superuser:

1. Create site s1, devicetype dt1, device role dr1
3. Create Tag tag1
4. Create custom field cf1 with object type dcim>device
5. Create device device1 with tag1, tag2 and every required field (dr1, dt1, s1)
6. Create user user1 with the following permissions
    - permission1: view DCIM > device, device role, device type, site
    - permission2: view Extras > tag constraints { "slug": "tag1"}
    - permission3: change DCIM > device

As user1

1. Go to device1 into edit mode
2. Change anything or nothing as you like
3. Hit save
4. tag2 is lost

**This happens with every object property the user is not allowed to view!** 

### Expected Behavior

The user should not be able to delete entries on the object just because he cannot see them. This leads to missing data. 
**In the object overview, the user sees all data, which should also be visible in the edit view, but not changeable unless explicit permission has been granted.**

### Observed Behavior

The user deletes data unintentionally and is not even aware of it.

Image

Image

Image

pheus:
### Closes: netbox-community/netbox#18987

Add read-only preservation of related object values hidden by object permissions. Single-value fields are disabled entirely; multi-value fields remain editable for visible values while restricted current values appear as disabled options and are merged back on save.

Fixes data loss when users edit objects with related values they cannot view, such as tags or tenants constrained by permissions.

### DIFF
--- a/docs/administration/permissions.md
+++ b/docs/administration/permissions.md
@@ -113,3 +113,7 @@ Device.objects.filter(
 ### Creating and Modifying Objects
 
 The same sort of logic is in play when a user attempts to create or modify an object in NetBox, with a twist. Once validation has completed, NetBox starts an atomic database transaction to facilitate the change, and the object is created or saved normally. Next, still within the transaction, NetBox issues a second query to retrieve the newly created/updated object, filtering the restricted queryset with the object's primary key. If this query fails to return the object, NetBox knows that the new revision does not match the constraints imposed by the permission. The transaction is then rolled back, leaving the database in its original state prior to the change, and the user is informed of the violation.
+
+#### Preserving Restricted Related Values
+
+When editing an object that references related objects the user is not permitted to view (for example a tag, tenant, or custom field value constrained away by a permission), those current values are shown read-only on the edit form and preserved when the form is saved. Single-value fields are disabled entirely; multi-value fields (such as tags) remain editable for the values the user can see, while the restricted current values are rendered as disabled options. This prevents a user from unintentionally clearing related objects they cannot see, without exposing any other restricted objects or allowing the restricted values to be replaced.
--- a/netbox/circuits/forms/model_forms.py
+++ b/netbox/circuits/forms/model_forms.py
@@ -213,6 +213,10 @@ class CircuitTerminationForm(NetBoxModelForm):
         FieldSet('port_speed', 'upstream_speed', 'xconnect_id', 'pp_info', name=_('Termination Details')),
     )
 
+    restricted_related_selectors = {
+        'termination': {'path': 'termination', 'lock_fields': ('termination_type',)},
+    }
+
     class Meta:
         model = CircuitTermination
         fields = [
@@ -306,6 +310,10 @@ class CircuitGroupAssignmentForm(NetBoxModelForm):
     fieldsets = (
         FieldSet('group', 'member_type', 'member', 'priority', 'tags', name=_('Group Assignment')),
     )
+
+    restricted_related_selectors = {
+        'member': {'path': 'member', 'lock_fields': ('member_type',)},
+    }
 
     class Meta:
         model = CircuitGroupAssignment
--- a/netbox/dcim/forms/mixins.py
+++ b/netbox/dcim/forms/mixins.py
@@ -38,6 +38,11 @@ class ScopedForm(forms.Form):
         selector=True
     )
 
+    # Duplicated by ipam.VLANGroupForm, which needs broader scope_type choices and cannot inherit ScopedForm.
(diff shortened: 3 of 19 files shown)


pheus:

pheus:

pheus:

The widget.label.disabled convention is now documented in the template. I still think it is a subtle convention, but I would treat it as acceptable and non-blocking.

pheus:

I did a refactor and expanded the restriction to more forms.

pheus:
@claude /review I've addressed all 3 points.

jeremystretch:
#18987 is currently qualified as a bug, however this PR targets the `feature` branch for inclusion into NetBox v4.7. As bug fixes need to ship in a patch release, we have two options:

1. Re-target this PR to `main` for inclusion in a v4.6.x release, and ensure that the fix doesn't have any undesired side-effects that would effect breaking changes to existing functionality.
2. Reject the bug as expected behavior, and open a new FR for v4.7 proposing an improvement to that behavior that may incur some intentional breaking changes.

Aside from that, I think we need to identify a solution which doesn't require the addition of boilerplate on each form, e.g.

```python
    restricted_related_selectors = {
        # The selectors are stored as the assigned_object GenericForeignKey; the model picks the matching one.
        'interface': {'path': 'assigned_object', 'model': Interface},
        'vminterface': {'path': 'assigned_object', 'model': VMInterface},
        'fhrpgroup': {'path': 'assigned_object', 'model': FHRPGroup},
    }
```

This is very brittle and likely to incur additional bugs over time as models and forms evolve.

We should also consider how the potential implementation of FR #22525 would impact this. While that initiative has not yet been fully fleshed out, it could potentially side-step this problem entirely: Returning at least a minimal representation of _all_ objects (i.e. just a primary key for those the user has no permission to view) would retain all selected related objects in the form. IMO we should explore that path further before committing to a solution here, especially one this complex.

jeremystretch (changes requested):
Please see the [above](https://github.com/netbox-community/netbox/pull/22390#issuecomment-4877282149)

jeremystretch (changes requested):
@pheus would you mind rebasing this onto `main` (instead of `feature`)? Let me know if that would be too disruptive and we can take a more surgical approach, replicating your solution in a new branch.

jeremystretch (changes requested):
@pheus you've done some truly remarkable work here. However, I'm not sure we've arrived at a solution suitable for shipping. I have several concerns with the approach.

#### 1. Preventing the user from editing a foreign key assignment

Assume the user is editing a site assigned to region A, and wants to reassign the site to region B. The user has permission to view only region B. In this case, the "region" field of the site edit form conveys the current assignment to region A, but does not permit the user to reassign the site to region B. This prevents the accidental removal of the current assignment, but also prevents the user from making a new, valid assignment.

An alternative solution might be to leave the field enabled and manually populate the currently-assigned object regardless of permission. We might be able to tweak the dynamic select widget to always preserve the initially assigned object as an avilable option to cover the scenario where the user accidentally clears the current assignment.

#### 2. Explicit declaration of `restricted_related_selectors` on each form

IMO this is an anti-pattern we need to avoid, as it's very fragile and easy to forget when adding new fields. I realize this is a significant challenge, but we'll need to devise some way to ensure these relations among form fields are derived automatically.

#### 3. Lots of complex logic baked into forms

While the degree or work undertaken here is admirable, the sheer amount of new logic being introduced to mitigate what is a fairly niche concern gives me pause. There are several hundred lines of code under RestrictedRelatedFieldsMixin used to surgically manipulate the rendering of individual form fields; even in the best of cases this is likely to introduce issues in the future given the complexities of form rendering in general. IMO it introduces too much risk given the relatively niche scenarios being addressed.

I get a strong sense that we should be looking to solve this problem at layer somewhere beneath the form itself, though to be fair I couldn't say where. I'm just not comfortable introducing this degree of fragility, especially given that it's applied so broadly throughout the application.

pheus:
Thanks everyone for taking the time to review this large and complex PR. I agree that the current approach introduces too much complexity and maintenance burden for solving this particular permissions-related problem, and may become brittle as forms continue to evolve.

I’m going to close the PR for now. The underlying problem is still worth addressing, and I may revisit it in the future if a simpler and more general solution emerges.

Thank you again for the thoughtful and open feedback. I really appreciate it.

PR #22400, opened 2026-06-05
Fixes #22399: Enforce object permissions for relevant static media

### LINKED ISSUE #22399: MediaView should enforce permissions on the related object when serving static content
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. Create an image attachment (on any object) and record the URL path to the image file (not the object).
2. Re-authenticate as a user without permission to view image attachments.
3. Fetch the image URL.

### Expected Behavior

NetBox should return a 404 as the user does not have permission to view the associated ImageAttachment object.

### Observed Behavior

The file is served, provided as the user is authenticated. Note that this requires knowing the exact URL path to the file content.

Device type images are similarly affected.

jeremystretch:
### Closes: #22399

- Extend the `get()` method on MediaView to enforce object permissions specifically for ImageAttachments and DeviceTypes
- Add the `Content-Disposition` and `X-Content-Type-Options` response headers
- Add tests

### DIFF
--- a/netbox/netbox/tests/test_views.py
+++ b/netbox/netbox/tests/test_views.py
@@ -1,9 +1,13 @@
 import urllib.parse
+from unittest.mock import patch
 
+from django.contrib.contenttypes.models import ContentType
+from django.http import HttpResponse
 from django.test import Client, override_settings
 from django.urls import reverse
 
-from dcim.models import Site
+from dcim.models import DeviceType, Manufacturer, Site
+from extras.models import ImageAttachment
 from netbox.constants import EMPTY_TABLE_TEXT
 from netbox.search.backends import search_backend
 from utilities.testing import TestCase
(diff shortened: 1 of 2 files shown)


jeremystretch:

pheus (changes requested):
This looks good, but I have strong suggestion:

Could we normalize `path` before applying the protected-directory checks?

The current checks run against the raw captured URL path, but `django.views.static.serve()` normalizes the path before resolving the file. That means a non-canonical path such as `foo/../image-attachments/<filename>` or `/image-attachments/<filename>` may not match `path.startswith("image-attachments/")`, but may still be normalized by `serve()` to `image-attachments/<filename>` and served from `MEDIA_ROOT`.

```python
import posixpath

def get(self, request, path):
    path = posixpath.normpath(path).lstrip('/')

    if path.startswith('image-attachments/'):
        ...
    elif path.startswith('devicetype-images/'):
        ...

    response = serve(request, path, document_root=settings.MEDIA_ROOT)
    ...
```

A regression test with `foo/../image-attachments/<filename>` would make this behavior explicit.

pheus (approved):
Thanks!

PR #22401, opened 2026-06-07
Fixes #22273 replace broken ArrayField index for service model

### LINKED ISSUE #22273: Application Service: OperationalError when adding a new service with more than approx. 2000 ports
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Go to /ipam/services/ on a NetBox instance running >= v4.6.0, e.g. https://demo.netbox.dev/ipam/services/
2. Add a new application service:
   - Select any Parent type / Parent
   - Select "Custom"
     - Name: foo
     - Protocol: TCP  (or any other)
     - Ports: 10001-20000  (>= 2050 ports should sufficient to trigger the error)
3. Click the "create" button

### Expected Behavior

A new application service should have been created with the specified port range.

### Observed Behavior

An OperationalError was raised:

```

<class 'django.db.utils.OperationalError'>

index row requires 40048 bytes, maximum size is 8191

Python version: 3.12.3
NetBox version: 4.6.1
Plugins: 
  netbox_demo: 0.5.0
The request which yielded the above error is shown below:

POST https://demo.netbox.dev/ipam/services/add/

If further assistance is required, please post to the NetBox discussion forum on GitHub.
```

mburggraf:
### Closes: #22273

- Replace the broken index in `0089_default_ordering_indexes.py` with a dummy to allow a smooth migration for users who are upgrading with a large set of ports
- Add internal _min_port field to hold the smallest port. It gets automatically set in the `save()` method
- Change the default ordering to Name, Protocol, Min Port and add a matching index
- Create a migration for the new database scheme
- Add Testcase for maximum number of ports to check if it everything gets saved without any problems

### DIFF
--- a/netbox/ipam/graphql/types.py
+++ b/netbox/ipam/graphql/types.py
@@ -264,7 +264,7 @@ class ServiceType(ContactsMixin, PrimaryObjectType):
 
 @strawberry_django.type(
     models.ServiceTemplate,
-    exclude=('_ports_lowest'),
+    exclude=('_ports_lowest',),
     filters=ServiceTemplateFilter,
     pagination=True
 )
--- a/netbox/ipam/graphql/types.py
+++ b/netbox/ipam/graphql/types.py
@@ -244,7 +244,7 @@ class RouteTargetType(PrimaryObjectType):
 
 @strawberry_django.type(
     models.Service,
-    exclude=('parent_object_type', 'parent_object_id'),
+    exclude=('_min_port', 'parent_object_type', 'parent_object_id'),
     filters=ServiceFilter,
     pagination=True
 )
--- a/netbox/ipam/graphql/types.py
+++ b/netbox/ipam/graphql/types.py
@@ -244,7 +244,7 @@ class RouteTargetType(PrimaryObjectType):
 
 @strawberry_django.type(
     models.Service,
-    exclude=('_min_port', 'parent_object_type', 'parent_object_id'),
+    exclude=('_ports_lowest', 'parent_object_type', 'parent_object_id'),
     filters=ServiceFilter,
     pagination=True
 )
@@ -264,7 +264,7 @@ class ServiceType(ContactsMixin, PrimaryObjectType):
 
 @strawberry_django.type(
     models.ServiceTemplate,
-    fields='__all__',
+    exclude=('_ports_lowest'),
     filters=ServiceTemplateFilter,
     pagination=True
 )
--- a/netbox/ipam/migrations/0091_alter_service_index_and_ordering.py
+++ b/netbox/ipam/migrations/0091_alter_service_index_and_ordering.py
@@ -1,0 +1,51 @@
+from django.db import migrations, models
+
+
+def populate__min_ports(apps, schema_editor):
+    Service = apps.get_model('ipam', 'Service')
+    CHUNK_SIZE = 500
+    chunk = []
+    service_objects = Service.objects.filter(_min_port__isnull=True).only('id', 'ports', '_min_port')
+    for service in service_objects.iterator(chunk_size=CHUNK_SIZE):
+        if service.ports:
+            service._min_port = min(service.ports)
+            chunk.append(service)
+
+        if len(chunk) >= CHUNK_SIZE:
+            Service.objects.bulk_update(chunk, ['_min_port'])
+            chunk = []
+    if chunk:
+        Service.objects.bulk_update(chunk, ['_min_port'])
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ipam', '0090_vlangroup_recompute_total_vlan_ids'),
+    ]
+
+    operations = [
+        migrations.RemoveIndex(
+            model_name='service',
+            name='ipam_servic_protoco_687d13_idx',
+        ),
+        migrations.AddField(
+            model_name='service',
+            name='_min_port',
+            field=models.PositiveIntegerField(blank=True, null=True),
+        ),
+        migrations.RunPython(populate__min_ports, migrations.RunPython.noop),
+        migrations.AddIndex(
+            model_name='service',
+            index=models.Index(
+                fields=['name', 'protocol', '_min_port', 'id'],
+                name='ipam_servic_name_ef1f28_idx'
+            ),
+        ),
+        migrations.AlterModelOptions(
+            name='service',
+            options={
+                'ordering': ('name', 'protocol', '_min_port', 'id')
+            },
+        ),
+    ]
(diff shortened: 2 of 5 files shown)


jeremystretch:

netbox/ipam/models/services.py
```diff
@@ -86,16 +85,29 @@ class Service(ContactsMixin, ServiceBase, PrimaryModel):
         verbose_name=_('IP addresses'),
         help_text=_("The specific IP addresses (if any) to which this application service is bound")
     )
+    _min_port = models.PositiveIntegerField(
+        null=True,
+        blank=True,
+    )
```
jeremystretch: Whatever we do for the Service model, we probably want to do for ServiceTemplate as well (via ServiceBase). The migration and tests will need to be extended as well.

netbox/ipam/models/services.py
```diff
@@ -86,16 +85,29 @@ class Service(ContactsMixin, ServiceBase, PrimaryModel):
         verbose_name=_('IP addresses'),
         help_text=_("The specific IP addresses (if any) to which this application service is bound")
     )
+    _min_port = models.PositiveIntegerField(
```
jeremystretch: I suggest renaming this to something like `_ports_first` or `_ports_lowest` to avoid confusion. ("min" might be inferred as declaring a minimum value.)

netbox/ipam/models/services.py
```diff
@@ -86,16 +85,29 @@ class Service(ContactsMixin, ServiceBase, PrimaryModel):
...
     clone_fields = (
         'protocol', 'ports', 'description', 'parent_object_type', 'parent_object_id', 'ipaddresses',
     )
 
+    def save(self, *args, **kwargs):
+        # On saving find the smallest port and save for default ordering
+        if self.ports:
+            self._min_port = min(self.ports)
+        else:
+            self._min_port = None
```
jeremystretch: We can condense this a bit.
```suggestion
        self._min_port = min(self.ports) if self.ports else None
```

netbox/ipam/models/services.py
```diff
@@ -86,16 +85,29 @@ class Service(ContactsMixin, ServiceBase, PrimaryModel):
...
+
+        super().save(*args, **kwargs)
+
     class Meta:
         indexes = (
-            models.Index(fields=('protocol', 'ports', 'id')),  # Default ordering
+            models.Index(fields=('name', 'protocol', '_min_port', 'id')),  # Default ordering
```
jeremystretch: IMO changing the default ordering to include the service name is out of scope here (though I understand why you did it). Let's leave it out for now, as this could be a break changing; some users might rely on the current numeric ordering.

netbox/ipam/models/services.py
jeremystretch: We'll need to add this field to the `exclude` list for the GraphQL types for Service and ServiceTemplate to avoid exposing it.

netbox/ipam/models/services.py
```diff
@@ -86,16 +85,29 @@ class Service(ContactsMixin, ServiceBase, PrimaryModel):
...
+    _min_port = models.PositiveIntegerField(
+        null=True,
+        blank=True,
+    )
 
     clone_fields = (
         'protocol', 'ports', 'description', 'parent_object_type', 'parent_object_id', 'ipaddresses',
     )
 
+    def save(self, *args, **kwargs):
```
jeremystretch: See Claude's feedback [here](https://github.com/netbox-community/netbox/pull/22401#issuecomment-4650586001) (point 2) regarding the handling `update_fields`.

jeremystretch (changes requested):
Thanks for working on this @mburggraf! I have some suggestions that I think will deliver a more robust solution.

netbox/ipam/models/services.py
mburggraf: Ah okay, then I probably misunderstood your comment in the original issue thread.
No problem, going to change it.

netbox/ipam/models/services.py
mburggraf: Since the ServiceTemplate has a completely different default ordering and doesn't use the problematic index, I didn't consider it necessary to apply those changes there as well.
However, considering this may change in the future, I guess it's wise to implement it in the common parent class.

netbox/ipam/migrations/0091_alter_service_index_and_ordering.py
```diff
@@ -0,0 +1,72 @@
...
+    for template in template_objects.iterator(chunk_size=CHUNK_SIZE):
+        if template.ports:
+            template._ports_lowest = min(template.ports)
+            chunk.append(template)
+
+        if len(chunk) >= CHUNK_SIZE:
+            ServiceTemplate.objects.bulk_update(chunk, ['_ports_lowest'])
+            chunk = []
+    if chunk:
+        ServiceTemplate.objects.bulk_update(chunk, ['_ports_lowest'])
```
jeremystretch: I think we can simplify this a bit by iterating through the models instead of duplicating the logic. This will ensure that any future modifications get applied consistently to both models.
```suggestion
def populate__ports_lowest(apps, schema_editor):
    Service = apps.get_model('ipam', 'Service')
    ServiceTemplate = apps.get_model('ipam', 'ServiceTemplate')
    CHUNK_SIZE = 500

    for model in (Service, ServiceTemplate):
        chunk = []
        qs = model.objects.filter(_ports_lowest__isnull=True).only('id', 'ports', '_ports_lowest')
        for obj in qs.iterator(chunk_size=CHUNK_SIZE):
            if obj.ports:
                obj._ports_lowest = min(service.ports)
                chunk.append(obj)
            if len(chunk) >= CHUNK_SIZE:
                model.objects.bulk_update(chunk, ['_ports_lowest'])
                chunk = []
        if chunk:
            model.objects.bulk_update(chunk, ['_ports_lowest'])
```

netbox/ipam/graphql/types.py
```diff
@@ -264,7 +264,7 @@ def parent(self) -> Annotated[
 
 @strawberry_django.type(
     models.ServiceTemplate,
-    fields='__all__',
+    exclude=('_ports_lowest'),
```
jeremystretch: Need to make sure this gets passed as a tuple rather than a string.
```suggestion
    exclude=('_ports_lowest',),
```

jeremystretch (commented):
Nice work @mburggraf! Just have to two minor follow-ups.

mburggraf:
I applied the follow-ups and fixed a typo you had from copying.

jeremystretch (approved):
Thanks @mburggraf!

PR #22405, opened 2026-06-08
Closes #22392: Deprecate support for Redis 5.x

### LINKED ISSUE #22392: Deprecate support for Redis <6.0
### Deprecated Functionality

Deprecate support for Redis releases earlier than v6.0.

### Scheduled removal

v4.7.0

### Justification

Redis versions older than 6.2 are all [end of life](https://endoflife.date/redis). Additionally, `py-redis` v8.0 now defaults to RESP3, which requires Redis 6.0 or later.

### Impact

NetBox deployments running Redis older than 6.0 will need to upgrade.

jeremystretch:
### Closes: #22392

- Register the new `check_redis_version()` function as a Django check to warn when the Redis version is <6.0
- Update the documentation

### DIFF
--- a/docs/installation/2-redis.md
+++ b/docs/installation/2-redis.md
@@ -8,7 +8,10 @@
 sudo apt install -y redis-server
 ```
 
-Before continuing, verify that your installed version of Redis is at least v5.0:
+Before continuing, verify that your installed version of Redis is at least v6.0:
+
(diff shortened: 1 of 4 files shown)


jeremystretch:

jeremystretch:
IMO only checking the caching backend is sufficient for a best effort deprecation warning.

arthanson (approved):

PR #22406, opened 2026-06-08
Fixes #22397: Don't get user configs for unauthenticated users during…

### LINKED ISSUE #22397: AttributeError when logged out user tries to export devices
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. Allow view access to NetBox for unauthenticated users.
2. Open up a random UI list view, e.g. http://netbox/dcim/devices/
3. Click on Export -> Current View

### Expected Behavior

CSV export works.

### Observed Behavior

You get a 500:

```
<class 'AttributeError'>

'AnonymousUser' object has no attribute 'config'

Python version: 3.12.13
NetBox version: 4.6.2
```

Code location: https://github.com/netbox-community/netbox/blob/v4.6.2/netbox/netbox/views/generic/bulk_views.py#L184

The error is similar to the one reported in #21784 and similar to there, I'd be happy to put up a PR to fix this by guarding it behind `request.user.is_authenticated` (I might also check if there are other places in the code that have the same pattern so maybe we can fix them all together).

fabi125:
### Fixes: #22397

Don't try to get user config if the user is not logged in. Also added a test case for it.

### DIFF
--- a/netbox/extras/tests/test_views.py
+++ b/netbox/extras/tests/test_views.py
@@ -408,6 +408,9 @@ class BookmarkTestCase(
     def test_list_objects_anonymous(self):
         return
 
+    def test_export_objects_anonymous(self):
+        return
+
     def test_list_objects_with_constrained_permission(self):
         return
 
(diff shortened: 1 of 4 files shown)


fabi125:
will look at the test failures, sorry about that

jeremystretch:

jeremystretch:
1. Valid, but out of scope for this fix.
2. The approach here is consistent with the (arguably suboptimal) existing convention.

jeremystretch (approved):
Thanks @fabi125!

PR #22410, opened 2026-06-08
Closes #22409: Disallow chosen-plaintext API tokens

### LINKED ISSUE #22409: Disallow chosen-plaintext API tokens via REST API
### NetBox version

v4.6.2

### Feature type

### Proposed functionality

Prevent REST API clients from passing a chosen plaintext value when creating a new API token. Only randomly-generated token values will be permitted. (This functionality has already been removed from the web UI, but had been retained in the REST API for backward compatibility.)

### Use case

Permitting only randomly-generated API tokens enhances application security in general.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #22409

- Change the `token` field on TokenSerializer to read-only
- Populate a random token under `clean()` on the Token model (in addition to `save()`)


jeremystretch:

jeremystretch:

jeremystretch:
The above feedback is out of scope for the PR.

bctiemann (approved):

PR #22412, opened 2026-06-08
#22411 - Enforce token write ability when running scripts via the REST API

### LINKED ISSUE #22411: Enforce token write ability on the REST API script execution endpoint
### NetBox version

v4.6.2

### Feature type

### Proposed functionality

Require the calling API token to have write ability enabled (write_enabled=True) in order to run a script via the REST API (POST /api/extras/scripts/{id}/).

ScriptViewSet overrides permission_classes with [IsAuthenticatedOrLoginNotRequired], which removes TokenPermissions from the permission chain. Because TokenPermissions is the only permission class that consults the token's write_enabled flag, the override means script execution is permitted by read-only tokens, provided the user holds the run_script permission.

The endpoint should enforce token write ability for script execution. Note that this endpoint also supports session authentication (via IsAuthenticatedOrLoginNotRequired), so the enforcement must apply only to token-authenticated requests and must not block session-authenticated users.

### Use case

API tokens can be created with write ability disabled so that they may be used for read-only automation and monitoring. Operators reasonably expect such tokens to be unable to make changes or trigger state-changing operations. Today a read-only token can still execute scripts, which contradicts that expectation and the principle of least privilege.

### Database changes

N/A

### External dependencies

N/A

arthanson:
### Closes: #22411 

The script execution endpoint (`POST /api/extras/scripts/{id}/`) did not honor the calling token's `write_enabled` flag. `ScriptViewSet` overrides `permission_classes` with `[IsAuthenticatedOrLoginNotRequired]`, which removes `TokenPermissions` (the only permission class that consults `request.auth.write_enabled`) from the chain. As a result, a read-only token (`write_enabled=False`) whose user holds the `run_script` permission could still trigger script execution, bypassing the read-only restriction that operators expect such tokens to enforce.

### Changes

* Added an explicit token write-ability check in `ScriptViewSet.post()`: when the request is token-authenticated (`request.auth` is a `Token`) and the token has `write_enabled=False`, the request is rejected with `403 Forbidden`. The existing object-level `run_script` permission check is retained.
* The check is intentionally scoped to token authentication. This endpoint also supports session authentication via `IsAuthenticatedOrLoginNotRequired`, so session-authenticated requests (where `request.auth` is not a `Token`) are unaffected and continue to work as before.
* Added `ScriptTestCase.test_run_token_write_enabled`: a `write_enabled=False` token (with the `run_script` permission) receives `403 Forbidden`, and `200 OK` once the token's write ability is enabled.
* Added `ScriptTestCase.test_run_session_auth`: a session-authenticated request (no token) is still permitted to run scripts, guarding against a regression in the session-auth path.

### Notes

The other REST API viewsets that override `permission_classes` with `IsAuthenticatedOrLoginNotRequired` (`StatusView`, `ObjectTypeViewSet`, `ConnectedDeviceViewSet`) expose only read-only (`GET`) methods, so the missing write check has no effect there — `ScriptViewSet` was the sole affected endpoint.

### DIFF
--- a/netbox/extras/api/views.py
+++ b/netbox/extras/api/views.py
@@ -7,7 +7,6 @@ from rest_framework.decorators import action
 from rest_framework.exceptions import PermissionDenied
 from rest_framework.generics import RetrieveUpdateDestroyAPIView
 from rest_framework.mixins import CreateModelMixin, ListModelMixin, RetrieveModelMixin
-from rest_framework.permissions import SAFE_METHODS
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
 from rest_framework.routers import APIRootView
@@ -22,6 +21,7 @@ from netbox.api.metadata import ContentTypeMetadata
 from netbox.api.renderers import TextRenderer
 from netbox.api.viewsets import BaseViewSet, NetBoxModelViewSet
 from netbox.api.viewsets.mixins import ObjectValidationMixin
+from users.models import Token
 from utilities.exceptions import RQWorkerNotRunningException
 from utilities.request import copy_safe_request
 from utilities.rqworker import any_workers_for_queue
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
Per claude review, addressed 1, 2 is a direct consequence of these changes - collapsing them back to one pattern would reintroduce the bug.

jeremystretch (commented):
While I agree generally with the assertion that executing a custom script should require write permissions, this would be a breaking change. IMO we should target it for NetBox v4.7. (Consider a scenario where a read-only API token is used to execute a custom script that performs only read operations, e.g. to generate some report.)

arthanson:
rebased to feature branch

jeremystretch (approved):

PR #22418, opened 2026-06-09
Closes #22288: Rename `JINJA2_FILTERS` to `JINJA_FILTERS`

### LINKED ISSUE #22288: Rename `JINJA2_FILTERS` config parameter to `JINJA_FILTERS`

jeremystretch:
### Closes: #22288

* Rename `JINJA2_FILTERS` to `JINJA_FILTERS`, maintaining backward compatibility
* Update references to Jinja2 in documentation

### DIFF
--- a/netbox/netbox/settings.py
+++ b/netbox/netbox/settings.py
@@ -141,7 +141,7 @@ HTTP_PROXIES = getattr(configuration, 'HTTP_PROXIES', {})
 INTERNAL_IPS = getattr(configuration, 'INTERNAL_IPS', ('127.0.0.1', '::1'))
 ISOLATED_DEPLOYMENT = getattr(configuration, 'ISOLATED_DEPLOYMENT', False)
 JINJA_ENVIRONMENT_PARAMS = getattr(configuration, 'JINJA_ENVIRONMENT_PARAMS', [])
-JINJA2_FILTERS = getattr(configuration, 'JINJA2_FILTERS', {})
+JINJA_FILTERS = getattr(configuration, 'JINJA_FILTERS', getattr(configuration, 'JINJA2_FILTERS', {}))
 LANGUAGE_CODE = getattr(configuration, 'DEFAULT_LANGUAGE', 'en-us')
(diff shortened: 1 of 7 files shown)


jeremystretch:

jeremystretch:
IMO the above are out of scope for this change, but will be handled separately.

netbox/netbox/settings.py
```diff
@@ -263,6 +263,12 @@
         "configuration file.",
         DeprecationWarning,
     )
+if hasattr(configuration, 'JINJA2_FILTERS'):
+    warnings.warn(
+        "JINJA2_FILTERS has been renamed to JINJA_FILTERS and the old name will be removed in NetBox v5.0. Please "
+        "update your configuration file to use JINJA_FILTERS instead.",
+        DeprecationWarning,
```
bctiemann, line 270: I know this is our usual pattern, but should this and other deprecation warnings use `FutureWarning` instead of `DeprecationWarning`? The difference being that unless you're in developer mode you're never going to see the latter, but it's mostly going to be sysadmins who will need to see these configuration-related warnings and thus `FutureWarning` would seem more appropriate.

bctiemann (changes requested):
See inline, OK to approve this as long as we've considered the `DeprecationWarning` issue.

jeremystretch:
@bctiemann good callout. I'm going to merge this as-is but have opened #22419 to audit and replace all instances of DeprecationWarning as appropriate, because there are likely many other places we should make this change.

PR #22420, opened 2026-06-09
Fixes #22346: Return a clean error message & redirect on SSO auth failure

### LINKED ISSUE #22346: Render SSO/SAML authentication failures as a login-page message instead of an HTTP 500 "Server Error"
### NetBox version

V4.6.1

### Feature type

### Proposed functionality

When a python-social-auth backend raises a SocialAuthBaseException (e.g. AuthFailed: SAML login failed: ..., AuthCanceled, AuthMissingParameter), NetBox should catch it and redirect the user back to the login page with the error surfaced via Django's messages framework, rather than letting it bubble up to the generic 500 handler.

Concretely:
- Add social_django.middleware.SocialAuthExceptionMiddleware to MIDDLEWARE in netbox/netbox/settings.py.
- Set SOCIAL_AUTH_LOGIN_ERROR_URL to the NetBox login view.
- Keep SOCIAL_AUTH_RAISE_EXCEPTIONS defaulting to DEBUG, so full exceptions still surface in development but not in production.
- Optionally map common exceptions to a generic, non-internal message (e.g. "Single sign-on failed. Please try again or contact your administrator.") so the raw social_core exception class is not shown to unauthenticated users.

### Use case

A user whose SSO/SAML login is rejected currently gets a page reading "Server Error" with the social_core.exceptions.AuthFailed: ... SAML login failed text. 

This looks like the application crashed and exposes an internal exception class to an unauthenticated visitor. Redirecting to the login page with a clear message lets the user retry and stops a normal, expected event from presenting as an outage.

### Database changes

none

### External dependencies

None new

jeremystretch:
### Closes: #22346

- Introduce & employ SocialAuthExceptionMiddleware to return a clean error message in the event of an SSO authentication failure
- Set `SOCIAL_AUTH_LOGIN_ERROR_URL` and `SOCIAL_AUTH_RAISE_EXCEPTIONS` in `settings.py
- Add tests

### DIFF
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -10,7 +10,9 @@ from django.db import ProgrammingError, connection
 from django.db.utils import InternalError
 from django.http import Http404, HttpResponseRedirect
 from django.middleware.common import CommonMiddleware as DjangoCommonMiddleware
+from django.utils.translation import gettext_lazy as _
 from django_prometheus import middleware
+from social_django.middleware import SocialAuthExceptionMiddleware as SocialAuthExceptionMiddleware_
 
 from netbox.config import clear_config, get_config
 from netbox.metrics import Metrics
@@ -26,6 +28,7 @@ __all__ = (
     'PrometheusAfterMiddleware',
     'PrometheusBeforeMiddleware',
     'RemoteUserMiddleware',
+    'SocialAuthExceptionMiddleware',
 )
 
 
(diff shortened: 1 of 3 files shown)


jeremystretch:

jeremystretch:
1. Set `SOCIAL_AUTH_LOGIN_ERROR_URL` to `LOGIN_URL` (the path does get reversed).
2. The test mock now reads the real `SOCIAL_AUTH_*` settings so it'll catch a misconfigured URL.

jnovinger (approved):

PR #22422, opened 2026-06-09
Fixes #22421: GraphQLTestCase should support relative imports

### LINKED ISSUE #22421: Adapt GraphQLTestCase to support relative package references
### Proposed Changes

Tweak `import_string(f'{lazy_module}.{name}')` to support relative imports from other packages.

### Justification

This adjustment will better adapt the tooling to eventually suit plugin testing.

jeremystretch:


jeremystretch:

arthanson (approved):

PR #22423, opened 2026-06-09
Closes #22428: Sanitize ImageAttachment uploads by re-encoding through Pillow

### LINKED ISSUE #22428: ImageAttachment upload accepts polyglot files that can carry non-image payloads
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. As an authenticated user with `extras.add_imageattachment`, attach a file to any object that supports image attachments (e.g. a Site).
2. Upload a polyglot PNG — a file that is simultaneously a valid PNG (passes Pillow's dimension check) and contains an HTML/JavaScript payload in its IDAT or trailer bytes:

```python
import io
from PIL import Image

buf = io.BytesIO()
Image.new('RGB', (1, 1)).save(buf, format='PNG')
polyglot = buf.getvalue() + b'<script>alert(document.domain)</script>'

with open('polyglot.png', 'wb') as f:
    f.write(polyglot)
```

3. The upload succeeds. The stored file contains the injected payload.

### Expected Behavior

Uploaded images should be sanitised before storage. Any non-image bytes embedded in an otherwise valid image file should be stripped so that the stored file contains only the re-encoded image data.

### Observed Behavior

`ImageField` validates image uploads using Pillow's dimension-extraction check, which confirms the file is a parseable image but does not sanitise the content. A polyglot PNG with an HTML or JavaScript payload in its trailer section passes this check and is stored as-is.

When a sufficiently privileged user (or an attacker who has compromised such credentials) later accesses the media URL, the stored file is served. Prior to the `MediaView` hardening in #22400, a browser could content-sniff the response as HTML, leading to same-origin XSS. Even with the `nosniff` and `Content-Disposition: attachment` headers added by #22400, storing attacker-controlled non-image bytes is an unnecessary risk.

Fixed in PR #22423.

bctiemann:
## Summary

Re-encodes uploaded images through Pillow before storage, stripping any embedded non-image payload (e.g. polyglot HTML/PNG files) before they reach the filesystem.

**Note:** `X-Content-Type-Options: nosniff` and `Content-Disposition: attachment` response headers were independently added to `MediaView` by #22400 (merged). This PR provides the complementary **upload-time** defence.

### Root cause

`ImageField` validates that uploaded files are images (via Pillow dimension extraction) but does not sanitise the file content. A polyglot PNG with an HTML payload in its IDAT/trailer section passes the dimension check unchanged and is stored as-is.

### Fix — `ImageAttachment.clean()` (`models.py`)

For new (uncommitted) uploads, `clean()` now:

1. Validates the file extension against `IMAGE_ATTACHMENT_IMAGE_FORMATS`, so the API path cannot bypass the form-level `accept=` filter.
2. Re-encodes the image through Pillow into a fresh buffer. `Pillow.save()` produces output containing only the image data — any non-image trailer bytes are discarded. GIF animation is preserved via `ImageSequence.Iterator`. AVIF/JPEG2000 failures (missing native codec library) surface a clear error instead of a generic "unable to process" message.
3. Replaces the in-memory upload with the sanitised `ContentFile` so Django writes the clean version to storage on model save.

The `_committed` guard ensures already-stored images are not re-encoded on every save.

Pillow is already a required dependency (`ImageField` uses it for dimension extraction), so the top-level import adds no new dependency.

Ref: SR-001 / VM-326 (internal security review, not an externally-reported disclosure)

### DIFF
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.postgres.fields import ArrayField
+from django.core.files.base import ContentFile
 from django.core.validators import ValidationError
 from django.db import models
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review

jeremystretch:

netbox/extras/tests/test_models.py
```diff
@@ -189,7 +189,102 @@ def test_duplicate_uploaded_names_get_suffixed_with_overwrite_style_storage(self
         self.assertEqual(first.filename, 'action-buttons.png')
         self.assertEqual(second.filename, 'action-buttons_sdmmer4.png')
 
-        self.assertCountEqual(storage.files.keys(), {base_name, suffixed_name})
```
jeremystretch, line 192: Seems like this was removed by mistake?

netbox/extras/models/models.py
```diff
@@ -741,6 +745,65 @@ def clean(self):
                 _("Image attachments cannot be assigned to this object type ({type}).").format(type=self.object_type)
             )
 
+        # Re-encode new image uploads through Pillow to strip any embedded
+        # non-image payloads (e.g. polyglot HTML/PNG files).
+        # Only process uncommitted (new) uploads; skip already-stored images.
+        if self.image and not self.image._committed:
```
jeremystretch, line 751: I don't think this covers the REST API: The model serializer performs validation on a copy of the instance, rather than the instance itself.

netbox/extras/models/models.py
```diff
@@ -741,6 +745,65 @@ def clean(self):
...
+                detected_mime = PillowImage.MIME.get(img_format, '').lower()
+                expected_mime = IMAGE_ATTACHMENT_IMAGE_FORMATS.get(ext, '').lower()
+                if detected_mime and expected_mime and detected_mime != expected_mime:
+                    raise ValidationError(
+                        _("File content does not match the declared extension (.{ext}).").format(ext=ext)
+                    )
+                # Re-encode into a fresh buffer; any non-image trailer bytes are discarded.
+                # Animated formats (GIF, WebP) use ImageSequence to preserve all frames.
+                buf = io.BytesIO()
+                if img_format in ('GIF', 'WEBP') and getattr(img, 'n_frames', 1) > 1:
```
jeremystretch, line 775: The logic in here seems fairly brittle, and I'm worried that we're taking on too much risk arbitrarily re-encoding various image formats. Given the relatively low potential avenue for attack, I have to question whether this is really worth undertaking.

jeremystretch (changes requested):
This approach only covers the ImageAttachment model. Presumably we would want to ensure global coverage for _all_ image field (e.g. front/rear images for DeviceTypes). I'm not sure whether it's feasible to implement something for ImageField directly, but we should at least move the re-encoding logic to a model mixin so that it can be reused.

That said, this seems like a lot of induced risk to address what is (IMO) a minimal potential threat, especially in light of the safeguards shipped under #22399. I'd be more comfortable with it if we could employ some known-good sanitization library (similar to `nh3`) instead of handling the re-encoding ourselves.

bctiemann:
I've closed the ticket and I'm going to withdraw this PR.

PR #22424, opened 2026-06-10
Closes #22429: Enforce ObjectPermission constraints on grant_token

### LINKED ISSUE #22429: Constrained grant_token ObjectPermissions are silently bypassed in the REST API
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. As an administrator, create an ObjectPermission granting `grant` on `users.token` with a constraint that limits which users tokens may be created for — for example `{"user__is_superuser": false}` (operator may only create tokens for non-superuser accounts).
2. Assign this permission to a non-admin user.
3. As that non-admin user, POST to `/api/users/tokens/` supplying a superuser's `user` PK in the request body.
4. The request succeeds and a token is created for the superuser, bypassing the constraint.

### Expected Behavior

The constraint `{"user__is_superuser": false}` should be enforced. Attempting to create a token for a superuser should return HTTP 403.

### Observed Behavior

`TokenSerializer.validate()` calls `request.user.has_perm('users.grant_token')` with no `obj` argument. In `ObjectPermissionMixin.has_perm`, `obj=None` causes an unconditional early return of `True` the moment the permission is found in cache, without ever evaluating constraints. The intended constraint is silently ignored and any cross-user token creation is permitted.

This allows a constrained operator to mint tokens for privileged accounts (including administrators), which can be used to escalate privilege.

Fixed in PR #22424.

bctiemann:
## Summary

Fixes a privilege escalation where constrained `grant_token` ObjectPermissions were silently bypassed in the REST API (SR-001 / VM-322, CWE-863).

### Root cause

`TokenSerializer.validate()` called `request.user.has_perm('users.grant_token')` with no `obj` argument. In `ObjectPermissionMixin.has_perm`, `obj=None` causes an unconditional early return of `True` once the permission is found in cache — constraints are never evaluated.

An operator holding a constrained `grant_token` ObjectPermission (e.g. `{"user__is_superuser": False}`) could supply any `user` in a token-creation request and the constraint would never fire, allowing tokens to be minted for privileged users.

### Fix

Replace `has_perm(obj=None)` with a new helper `_user_may_grant_token(requesting_user, token_user)` that:

1. Returns `True` immediately for superusers (mirrors `ObjectPermissionMixin.has_perm`'s bypass).
2. Confirms the requesting user holds `users.grant_token` at all.
3. Treats an empty/null constraint as unconstrained (preserves existing semantics).
4. For constrained permissions, remaps Token-model constraint keys to User-model keys (`user__<field>` → `<field>`, `user` → `pk`) and evaluates them directly against the target User record — the only variable in a new-token creation scenario. Non-user Token fields are treated as fail-closed (deny).

We cannot use `has_perm(obj=token_instance)` with an existing Token because the token does not exist yet, and passing an unsaved instance returns empty from `model.objects.filter(pk=None)`.

### Tests

- `test_grant_token_constrained_permission_is_enforced` — `{"user__is_superuser": False}` denies superuser target, allows non-superuser target.
- `test_grant_token_superuser_always_allowed` — superuser can grant tokens without an explicit ObjectPermission.
- `test_grant_token_self_only_constraint` — `{"user": "$user"}` blocks cross-user grants.
- `test_grant_token_non_user_field_constraint_fails_closed` — non-user Token field constraints fail closed for unsaved tokens.

Ref: SR-001 / VM-322 (internal security review, not an externally-reported disclosure)

### DIFF
--- a/netbox/users/api/serializers_/tokens.py
+++ b/netbox/users/api/serializers_/tokens.py
@@ -34,13 +34,18 @@ def _user_may_grant_token(requesting_user, token_user):
     Constraints referencing other Token fields cannot be evaluated for an unsaved
     token; the function returns False (deny) for those.
     """
+    # Mirrors ObjectPermissionMixin.has_perm: superusers implicitly have all permissions.
+    if requesting_user.is_active and requesting_user.is_superuser:
+        return True
+
     perm = 'users.grant_token'
 
     # get_all_permissions() populates _object_perm_cache as a side effect.
+    # Guard against missing cache key in case a non-standard backend is in use.
     if perm not in requesting_user.get_all_permissions():
         return False
 
-    constraints = requesting_user._object_perm_cache[perm]
+    constraints = getattr(requesting_user, '_object_perm_cache', {}).get(perm, [])
 
     # An empty/null constraint means "no restriction" — allow any target.
     if any(not c for c in constraints):
@@ -54,14 +59,13 @@ def _user_may_grant_token(requesting_user, token_user):
     for constraint in constraints:
         user_constraint = {}
         for key, raw_val in constraint.items():
-            # Resolve $user placeholder
+            # Resolve $user placeholder; constraint values are JSON primitives so
+            # no .pk extraction is needed.
             val = resolved_user_id if raw_val == user_token else raw_val
-            if hasattr(val, 'pk'):
-                val = val.pk
             if key == 'user':
                 user_constraint['pk'] = val
             elif key.startswith('user__'):
-                user_constraint[key[len('user__'):]] = val
+                user_constraint[key.removeprefix('user__')] = val
             else:
                 # Non-user Token field — cannot evaluate for a new (unsaved) token.
                 # Fail closed.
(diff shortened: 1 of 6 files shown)


bctiemann:
@claude /review

arthanson (changes requested):
This works bug Claude is saying the same has_perm('users.grant_token')-with-no-obj pattern exists at two other token-creation sites, and both have the identical privilege-escalation bug the PR is fixing:

- users/views.py:89 - TokenBulkImportView.save_object() (bulk CSV import)
- users/forms/model_forms.py:198 - TokenForm.clean() (web UI create form)

An operator with a constrained grant_token (e.g. {"user__is_superuser": False}) can still mint a token for a superuser through the UI or bulk import, exactly as described in #22429 - just not through the REST API anymore.

bctiemann:
Addressed @arthanson 's feedback by adding the same pattern to the UI form and bulk import paths.

arthanson (approved):

PR #22426, opened 2026-06-10
Closes #22427: Validate JSONFilter.path and remove regex from JSONLookup

### LINKED ISSUE #22427: GraphQL JSONFilter.path allows ORM operator injection over JSONField contents
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. As an authenticated user with read access to any object that has a `JSONFilter`-decorated field (e.g. `ConfigContext` for `local_context_data`, `ObjectChange` for `prechange_data`/`postchange_data`, or any model with custom field data), submit a GraphQL query of the form:

```graphql
{
  config_context_list(
    filters: {
      local_context_data: {
        path: "secret_key__regex"
        lookup: { string_lookup: { regex: "^s" } }
      }
    }
  ) { id }
}
```

2. Observe that the filter is accepted and applied. Vary the regex pattern to probe the value of `local_context_data.secret_key` one character at a time.

3. GraphQL alias multiplication allows hundreds of probes per request.

### Expected Behavior

`JSONFilter.path` should only accept valid JSON key paths. Operator-keyword segments (e.g. `regex`, `iregex`, `exact`) must be rejected. The `string_lookup` sub-filter must not expose `regex` or `i_regex` operators on JSON paths.

### Observed Behavior

`JSONFilter.path` accepts arbitrary strings that are concatenated unsanitised into the Django ORM lookup prefix (`f'{prefix}{self.path}__'`). Combined with `StrFilterLookup.regex` / `i_regex` this forms a per-character boolean oracle over every JSONField accessible to the querying user. GraphQL alias multiplication amplifies throughput to hundreds of bits per request.

Two attack primitives:

1. **Path injection**: `path="secret__regex"` appends `__regex` as an ORM lookup suffix.
2. **Direct regex operator**: `string_lookup.regex` on any valid path performs full regex matching against JSON field contents.

Fixed in PR #22426.

bctiemann:
## Summary

Fixes two related attack primitives that together form a per-character boolean oracle over every JSONField accessible via GraphQL.

### Attack path

An authenticated user with read access to any model that exposes a `JSONFilter`-decorated field (`custom_field_data`, `local_context_data`, `prechange_data`, `postchange_data`) can:

1. Supply an arbitrary `path` string (e.g. `"secret__regex"`) that injects a Django ORM operator suffix into the JSON lookup.
2. Use `string_lookup.regex` / `string_lookup.i_regex` in `JSONLookup` to perform full regex matching against JSON field contents.
3. Multiply queries via GraphQL aliases to extract secrets bit-by-bit — the same primitive class identified in NB-2703, but in the `JSONFilter` layer.

### Fix 1 — Path validation (`_validate_json_path`)

`JSONFilter.path` is now validated before being concatenated into the ORM lookup prefix. Each `__`-separated segment must:
- Match `[A-Za-z0-9][A-Za-z0-9_-]*` (alphanumerics, hyphens, underscores; must start with alphanumeric)
- Not match any Django ORM operator keyword (`regex`, `iregex`, `exact`, `contains`, etc.)

An invalid path returns `Q(pk__in=[])` (empty result set) rather than propagating user input into the ORM.

### Fix 2 — `JSONStringLookup` without regex operators

`JSONLookup.string_lookup` previously used `StrFilterLookup`, which exposes `regex` and `i_regex`. A new `JSONStringLookup` type provides all other string operators while omitting those two. The omission is documented with an inline comment. `JSONLookup` and `JSONStringLookup` are added to `__all__`.

### Scope

The single changed file (`netbox/graphql/filter_lookups.py`) is the only location where `JSONFilter` is defined. All call sites in `extras/graphql/filter_mixins.py`, `core/graphql/filters.py`, and `extras/graphql/filters.py` benefit automatically.

Ref: SR-001 / VM-323 (internal security review, not an externally-reported disclosure)

### DIFF
--- a/netbox/netbox/graphql/filter_lookups.py
+++ b/netbox/netbox/graphql/filter_lookups.py
@@ -1,3 +1,4 @@
+import re
 from enum import Enum
 from typing import Generic, TypeVar
 
@@ -13,11 +14,45 @@ from strawberry_django import (
     ComparisonFilterLookup,
     FilterLookup,
     RangeLookup,
-    StrFilterLookup,
     process_filters,
 )
 
 from netbox.graphql.scalars import BigInt
+
+# ------------------------------------------------------------------
+# JSON path validation (VM-323)
+# ------------------------------------------------------------------
+
+# Each segment of a JSON path may only contain alphanumerics, underscores, and
+# hyphens.  Hyphens are included because JSON keys commonly use them; leading
+# underscores are permitted (e.g. _foo is a valid key name).
+_JSON_PATH_SEGMENT_RE = re.compile(r'^[A-Za-z0-9_][A-Za-z0-9_-]*$')
+
+
+def _validate_json_path(path: str) -> str:
+    """Validate a JSON traversal path for use in ORM lookups.
+
+    Each ``__``-separated segment must match ``[A-Za-z0-9_][A-Za-z0-9_-]*``.
+    Raises ``ValueError`` on an empty path, empty segment, or segment with
+    disallowed characters.
+
+    ORM operator names (``date``, ``regex``, etc.) are intentionally *not*
+    blocked here: ``JSONFilter.filter()`` always appends ``__`` to the path
+    before handing it to ``process_filters``, so a segment named ``regex``
+    becomes another level of JSON key traversal (``data__key__regex__exact``),
+    not the ORM regex transform (``data__key__regex=…``).
+    """
+    if not path:
+        raise ValueError("JSON path cannot be empty")
+
+    for segment in path.split('__'):
+        if not segment:
+            raise ValueError("JSON path contains consecutive or trailing '__'")
+        if not _JSON_PATH_SEGMENT_RE.match(segment):
+            raise ValueError(f"Invalid JSON path segment: {segment!r}")
+
+    return path
+
 
 __all__ = (
     'ArrayLookup',
@@ -28,6 +63,8 @@ __all__ = (
     'IntegerLookup',
     'IntegerRangeArrayLookup',
     'JSONFilter',
+    'JSONLookup',
+    'JSONStringLookup',
     'StringArrayLookup',
     'TreeNodeFilter',
 )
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review

netbox/netbox/graphql/filter_lookups.py
```diff
@@ -15,13 +16,56 @@
...
 
 from netbox.graphql.scalars import BigInt
 
+# ------------------------------------------------------------------
+# JSON path validation (VM-323)
+# ------------------------------------------------------------------
+
+# Each segment of a JSON path may only contain alphanumerics, hyphens, and
+# underscores.  Hyphens are included because JSON keys commonly use them.
+_JSON_PATH_SEGMENT_RE = re.compile(r'^[A-Za-z0-9][A-Za-z0-9_-]*$')
```
jeremystretch: This regex would not match on fields named with a leading prefix (e.g. `_foo`), which AFAICT should be allowed.
```suggestion
_JSON_PATH_SEGMENT_RE = re.compile(r'^[A-Za-z0-9_][A-Za-z0-9_-]*$')
```

netbox/netbox/graphql/filter_lookups.py
```diff
@@ -15,13 +16,56 @@
...
+    """
+    if not path:
+        raise ValueError("JSON path cannot be empty")
+
+    for segment in path.split('__'):
+        if not segment:
+            raise ValueError("JSON path contains consecutive or trailing '__'")
+        if not _JSON_PATH_SEGMENT_RE.match(segment):
+            raise ValueError(f"Invalid JSON path segment: {segment!r}")
+        if segment.lower() in _ORM_OPERATOR_NAMES:
```
jeremystretch: I think this breaks filtering on legitimate data keys. For example, a config context data blob might have a literal key `date`, which is valid.

netbox/netbox/graphql/filter_lookups.py
```diff
@@ -39,9 +85,31 @@
...
+    i_exact: str | None = strawberry_django.filter_field()
+    contains: str | None = strawberry_django.filter_field()
+    i_contains: str | None = strawberry_django.filter_field()
+    starts_with: str | None = strawberry_django.filter_field()
+    i_starts_with: str | None = strawberry_django.filter_field()
+    ends_with: str | None = strawberry_django.filter_field()
+    i_ends_with: str | None = strawberry_django.filter_field()
+    in_: list[str] | None = strawberry_django.filter_field()
+    isnull: bool | None = strawberry_django.filter_field()
+    # regex / i_regex intentionally omitted (oracle prevention, VM-323)
```
jeremystretch: I don't follow what disabling regex lookups buy us. The hypothetical attack cited in #22427 uses a `^` regex, which is effectively identical to the `starts_with` lookup.

jeremystretch (changes requested):

netbox/netbox/graphql/filter_lookups.py
bctiemann: Point, no reason to remove these if `starts_with` is supported. Done.

netbox/netbox/graphql/filter_lookups.py
bctiemann: Agreed, done.

netbox/netbox/graphql/filter_lookups.py
bctiemann: Yeah, this is unnecessary/redundant. Removed.

jeremystretch:
@claude /review Does this resolve the security concern raised in #22427?

jeremystretch (approved):
Thanks @bctiemann!

PR #22432, opened 2026-06-10
Closes #22431: Replace random.choice with secrets.choice in Token.generate()

### LINKED ISSUE #22431: Token.generate() uses non-CSPRNG random.choice() — tokens are predictable
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. Note that `Token.generate()` in `users/models/tokens.py` uses `random.choice(TOKEN_CHARSET)` in a generator expression.
2. `random` in Python uses the Mersenne Twister PRNG, which is not cryptographically secure. Observing approximately 624 32-bit outputs from the same process allows complete state recovery via standard MT19937 untemper attacks (well-documented in CWE-338 references).
3. An authenticated user who can observe token plaintexts (e.g. by creating tokens via the API — which returns the plaintext on creation) can accumulate enough outputs to recover the PRNG state.
4. Once the state is recovered, all future token values generated by the same worker process are predictable, including tokens created for other users (including administrators).

### Expected Behavior

Token generation must use a CSPRNG. Python's `secrets` module is the standard library answer: it is backed by `os.urandom()` / `getrandom()` and is immune to state-recovery attacks.

### Observed Behavior

`Token.generate()` uses `random.choice()` (Mersenne Twister). Sufficient observations of token values generated by the same worker allow full PRNG state recovery and prediction of subsequently generated tokens.

Fixed in PR #22432.

bctiemann:
## Summary

Replaces the non-CSPRNG `random.choice()` call in `Token.generate()` with `secrets.choice()` (SR-001 / VM-317, CWE-338).

### Root cause

`Token.generate()` used `random.choice(TOKEN_CHARSET)` to build token strings. Python's `random` module uses the Mersenne Twister PRNG, which is not cryptographically secure. Observing approximately 624 32-bit outputs from the same worker process is sufficient to recover the full PRNG state and predict all subsequent outputs — including tokens minted for other users (potentially administrators) by the same worker.

### Fix

Replace `import random` with `import secrets` and change `random.choice` to `secrets.choice`. `secrets` is backed by `os.urandom()` / `getrandom()` (OS-level CSPRNG) and is immune to state-recovery attacks. This is a one-line logic change with no behavioural impact beyond improved entropy.

The fix plan also requested an audit of other `random.*` usages in auth-adjacent code; none were found.

### Tests

- `test_generate_uses_csprng` — patches `secrets.choice` with `wraps=` to verify it is called exactly `TOKEN_DEFAULT_LENGTH` times and that the output only contains `TOKEN_CHARSET` characters.
- `test_generate_length_parameter` — verifies the `length=` parameter is respected across several values.

Ref: SR-001 / VM-317 (internal security review, not an externally-reported disclosure)

### DIFF
--- a/netbox/users/models/tokens.py
+++ b/netbox/users/models/tokens.py
@@ -1,6 +1,6 @@
 import hashlib
 import hmac
-import random
+import secrets
 import zoneinfo
 
 from django.conf import settings
@@ -260,7 +260,7 @@ class Token(models.Model):
         """
         Generate and return a random token value of the given length.
         """
-        return ''.join(random.choice(TOKEN_CHARSET) for _ in range(length))
+        return ''.join(secrets.choice(TOKEN_CHARSET) for _ in range(length))
 
     def update_digest(self):
         """
--- a/netbox/users/tests/test_models.py
+++ b/netbox/users/tests/test_models.py
@@ -1,10 +1,13 @@
+import secrets
 from datetime import timedelta
+from unittest.mock import patch
 
 from django.core.exceptions import ValidationError
 from django.test import TestCase, override_settings
 from django.utils import timezone
 
 from users.choices import TokenVersionChoices
+from users.constants import TOKEN_CHARSET, TOKEN_DEFAULT_LENGTH
 from users.models import Token, User
 from utilities.testing import create_test_user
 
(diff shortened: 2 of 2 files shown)


bctiemann:
@claude /review and limit response only to areas of potential concern.

jeremystretch (approved):

PR #22434, opened 2026-06-10
Fixes #22303: Annotate fields & omit parameters in OpenAPI schema

### LINKED ISSUE #22303: OpenAPI schema is missing `fields` param
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.1

### Python Version

3.12

### Steps to Reproduce

1. Download schema from `/api/schema`
2. Search for `fields` string. Notice that it's not there
```
> grep fields ./NetBox\ REST\ API\ \(4.6\).yaml | sort -u
          description: Custom fields within the same group will be displayed together
          description: Maximum allowed value (for numeric fields)
          description: Minimum allowed value (for numeric fields)
        custom_fields:
      description: Adds support for custom fields and tags.
      operationId: extras_custom_fields_bulk_destroy
      operationId: extras_custom_fields_bulk_partial_update
      operationId: extras_custom_fields_bulk_update
      operationId: extras_custom_fields_create
      operationId: extras_custom_fields_destroy
      operationId: extras_custom_fields_list
      operationId: extras_custom_fields_partial_update
      operationId: extras_custom_fields_retrieve
      operationId: extras_custom_fields_update
  /api/extras/custom-fields/:
  /api/extras/custom-fields/{id}/:
```

### Expected Behavior

I would expect the `fields` param to exist in the schema, since it is a valid parameter.

### Observed Behavior

The `fields` param is not in the OpenAPI spec.

# Impact

`pynetbox` recently added a [`strict_filters` param](https://github.com/netbox-community/pynetbox#filters-validation) which we would like to use. That feature relies on the OpenAPI spec. Since it isn't in the spec, it prevents the user from using the `fields` param

jeremystretch:
### Fixes: #22303

Extend NetBoxAutoSchema to annotate the optional `fields` and `omit` query parameters.

### DIFF
--- a/netbox/core/api/schema.py
+++ b/netbox/core/api/schema.py
@@ -14,7 +14,7 @@ from drf_spectacular.plumbing import (
     get_doc,
 )
 from drf_spectacular.types import OpenApiTypes
-from drf_spectacular.utils import Direction
+from drf_spectacular.utils import Direction, OpenApiParameter
 
 from netbox.api.fields import ChoiceField
 from netbox.api.serializers import WritableNestedSerializer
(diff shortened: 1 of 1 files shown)


jeremystretch:

jeremystretch:
Addressed the above.

Didn't update `contrib/openapi.json` as part of this PR (that will happen during the release process), but did verify that it makes the intended changes.

bctiemann (approved):

PR #22445, opened 2026-06-13
Fixes #22444: Fix KeyError exception on device view with non-English locale

### LINKED ISSUE #22444: KeyError 'V' on PowerFeed detail view when locale is not en-us
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. Set LANGUAGE_CODE or user language to any non-English locale (e.g. 'fr')
2. Create a PowerFeed object (DCIM > Power Feeds > Add)
3. Navigate to the PowerFeed detail page (/dcim/power-feeds/<id>/)

### Expected Behavior

The PowerFeed detail page renders correctly, displaying voltage (e.g. "230V") 
and amperage (e.g. "10A") in the Electrical Characteristics panel.

### Observed Behavior

HTTP 500 error with KeyError: 'V'

Exception Location: netbox/ui/attrs.py, line 134, in get_value
    return self.format_string.format(value)

Root cause: in dcim/ui/panels.py, the format strings '{}V' and '{}A' are 
wrapped in Django's gettext_lazy _() function. When using a non-English 
locale, the translation engine interprets the letter after {} as a named 
format key, transforming '{}V' into '{V}'. When .format(230) is called, 
Python raises KeyError: 'V'.

The fix is to remove _() from format_string='{}V' and format_string='{}A' 
since Volt and Ampere are universal SI unit symbols that do not require 
translation. For format_string=_('{} millimeters'), replacing {} with {0} 
prevents the same issue while preserving translatability.

Affected lines in dcim/ui/panels.py:
- Line 35:  format_string=_('{} millimeters')  → format_string=_('{0} millimeters')
- Line 358: format_string=_('{}V')             → format_string='{}V'
- Line 359: format_string=_('{}A')             → format_string='{}A'

jeremystretch:
### Fixes: #22444

Tweak certain format strings to prevent erroneous formatting during translations

### DIFF
--- a/netbox/dcim/ui/panels.py
+++ b/netbox/dcim/ui/panels.py
@@ -32,7 +32,7 @@ class RackDimensionsPanel(panels.ObjectAttributesPanel):
     outer_width = attrs.NumericAttr('outer_width', unit_accessor='get_outer_unit_display')
     outer_height = attrs.NumericAttr('outer_height', unit_accessor='get_outer_unit_display')
     outer_depth = attrs.NumericAttr('outer_depth', unit_accessor='get_outer_unit_display')
-    mounting_depth = attrs.TextAttr('mounting_depth', format_string=_('{} millimeters'))
+    mounting_depth = attrs.TextAttr('mounting_depth', format_string=_('{0} millimeters'))
 
 
 class RackNumberingPanel(panels.ObjectAttributesPanel):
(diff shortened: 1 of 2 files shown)


jeremystretch:

pheus (approved):

PR #22450, opened 2026-06-15
Closes #22280: Set 91% test coverage threshold and exclude non-testable paths

### LINKED ISSUE #22280: Enforce a minimum test coverage threshold in CI
### Proposed Changes

Configure coverage.py to enforce a minimum total coverage percentage for NetBox’s test suite.

This can be implemented by adding `fail_under` to the existing `[tool.coverage.report]` section in `pyproject.toml`, using the current CI coverage result as the initial baseline. For example:

```toml
[tool.coverage.report]
skip_covered = true
fail_under = <current baseline>
omit = [
    "*/migrations/*",
    "*/tests/*",
]
```

The existing CI workflow already runs `coverage report`, so coverage.py will automatically fail the job if the reported total coverage drops below the configured threshold.

### Justification

NetBox already collects and reports test coverage in CI, but the report is currently informational only. Enforcing a minimum threshold would help prevent accidental decreases in overall test coverage and make regressions easier to catch during review.

Starting with the current baseline should keep this low-risk while giving maintainers a clear signal when future changes reduce coverage.

pheus:
### Closes: #22280

Exclude non-application code from coverage reports and set fail_under to 91%. Omits scripts, configuration files, WSGI entrypoint, secret key generator, debug utilities, and deprecated management commands that lack testable logic or are outside the application scope.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,9 +34,17 @@ sigterm = true
 
 [tool.coverage.report]
 skip_covered = true
+fail_under = 91
 omit = [
     "*/migrations/*",
     "*/tests/*",
+    # Non-application code (no testable logic / not part of the app)
+    "netbox/scripts/*",  # SCRIPTS_ROOT: user/generated scripts
+    "*/netbox/configuration*.py",  # settings/config files (template, testing, local)
+    "*/netbox/wsgi.py",  # WSGI entrypoint
+    "*/generate_secret_key.py",  # standalone CLI helper
+    "*/utilities/debug.py",  # debug-toolbar hook, active only when DEBUG=True
+    "*/extras/management/commands/housekeeping.py",  # deprecated; will not be tested
 ]
 
 [tool.pyright]


pheus:

pheus:
1. The current coverage hits 91%, and the intent is to use that as the minimum floor so coverage does not regress below the current level. I’ll clarify this in the PR description.
2. The command is already deprecated, and we decided not to add new tests for it while it is pending removal. The exclusion is intended specifically for this deprecated command, not as a general precedent for active application code.
3. The current `scripts/` directory is empty, and the only scripts present during coverage are generated test scripts, so they do not need coverage.
4. This is developer-only debug toolbar logic and not user-facing application behavior.
5. I’ll leave the CI command as-is since `coverage.py` picks up `fail_under` from `pyproject.toml` and will fail correctly when the threshold is not met.

pyproject.toml
```diff
@@ -34,9 +34,17 @@ sigterm = true
 
 [tool.coverage.report]
 skip_covered = true
+fail_under = 91
 omit = [
     "*/migrations/*",
     "*/tests/*",
+    # Non-application code (no testable logic / not part of the app)
+    "*/scripts/*",  # SCRIPTS_ROOT: user/test scripts
```
jnovinger: I think I agree with claudebot here. Use the anchored pattern instead:

```toml
"netbox/scripts/*",  # SCRIPTS_ROOT: user/generated scripts
```

This matches the one real location with no wildcards that can drift.

Worth tightening because `*/scripts/*` is only safe today by accident: nothing under `source` has a `scripts/` subdirectory yet. Add one later (`dcim/scripts/`, etc.) and it gets silently excluded from coverage, which defeats the regression-catching this PR exists to provide.

jnovinger (changes requested):

pyproject.toml
pheus: Good catch! Thanks!

jnovinger (approved):

PR #22452, opened 2026-06-15
Closes #21992: Enable background job support for REST API bulk requests

### LINKED ISSUE #21992: Enable background job support for REST API requests
### NetBox version

v4.5.8

### Feature type

New functionality

### Proposed functionality

NetBox v4.4 introduced background job support for bulk UI functions, such as creating or deleting a series of objects. This FR proposes extending that ability to REST API requests as well.

This could be implemented using either a query parameter (e.g. `?background=true`) or an HTTP header. Upon receipt of an API request indicating background processing is desired, NetBox should enqueue a background job to perform the task and immediately return a JSON payload containing the job ID and URL.

### Use case

This will enable REST API clients to frontload time-intensive bulk operations without having to wait for the task to complete before receiving a response.

### Database changes

N/A

### External dependencies

N/A

jnovinger:
### Closes: #21992

Extends the background-job support introduced in v4.4 for UI bulk operations to the REST API. A bulk write (creating, updating, or deleting a JSON list of objects via a model's list endpoint) can opt into background processing with the `?background=true` query parameter: the request is validated synchronously and, if accepted, an `AsyncAPIJob` is enqueued and `202 Accepted` is returned with the job's ID and poll URL. The write is performed later by a worker that re-invokes the same viewset action, so behavior matches the synchronous path, including all-or-nothing transaction semantics.

#### Behavior

- Applies only to **bulk** operations (a JSON list) on list endpoints. Single-object writes, GET requests, and non-list payloads ignore the parameter and run synchronously.
- The job result is captured in `job.data` as `{status_code, data}`, mirroring the response the synchronous request would have returned. Clients poll `/api/core/jobs/{id}/` for the outcome.
- Handled rejections are translated to match the synchronous API and terminate the job as `failed` (not `errored`): DRF `APIException` via `handle_exception()`, and `AbortRequest` / `ProtectedError` / `RestrictedError` via a new `NetBoxModelViewSet.exception_to_response()` helper (e.g. a protected-dependent delete records a 409 with the dependent list).
- A background request is refused with `503` when no worker is servicing the queue (mirroring the scripts API), and rejected with `400` when combined with an `If-Match` precondition, which cannot be honored once execution is deferred.

#### Notes for reviewers

- `exception_to_response()` intentionally duplicates the translation logic in `NetBoxModelViewSet.dispatch()` rather than `dispatch()` being refactored to call it. This keeps the change off the synchronous write hot path; consolidating the two into a single source of truth is a reasonable follow-up.
- Two behaviors differ from a synchronous request and are documented as such: `fields`/`omit` field selection is not applied to the stored result, and the authorization captured at acceptance is not re-checked if the token is later disabled/expires before the job runs. These are noted as candidates for follow-up.
- Manual end-to-end testing (against a live `runserver` + `rqworker`) is summarized in a follow-up comment.

#### Testing

New tests in `netbox/netbox/tests/test_api_background.py` cover bulk create/update (PATCH and PUT)/delete, the 202 contract and `Location` header, all-or-nothing rollback with structured error capture, synchronous validation rejection, the `If-Match` and no-worker guards, object-permission parity with the synchronous path, change-log `request_id` propagation, single-object/GET pass-through, and the deferred-execution (non-immediate) scheduling path.

### DIFF
--- a/netbox/netbox/api/viewsets/__init__.py
+++ b/netbox/netbox/api/viewsets/__init__.py
@@ -151,6 +151,7 @@ class NetBoxReadOnlyModelViewSet(
 
 class NetBoxModelViewSet(
     ETagMixin,
+    mixins.BackgroundOperationMixin,
     mixins.BulkUpdateModelMixin,
     mixins.BulkDestroyModelMixin,
     mixins.ObjectValidationMixin,
@@ -215,9 +216,37 @@ class NetBoxModelViewSet(
                 **kwargs
             )
 
+    def exception_to_response(self, exc):
+        """
+        Translate a NetBox/Django exception that is not a DRF APIException into the same
+        Response that dispatch() would return for it. Returns None if the exception is not
+        one this method handles (the caller should then re-raise or defer to DRF).
+
+        This mirrors the except clauses in dispatch(); it is also called by the background
+        job runner (netbox.jobs.AsyncAPIJob), which executes action methods directly and so
+        bypasses dispatch(). NOTE: dispatch() does not yet call this helper itself — see the
+        PR description for the proposed consolidation.
+        """
+        if isinstance(exc, (ProtectedError, RestrictedError)):
+            if type(exc) is ProtectedError:
+                protected_objects = list(exc.protected_objects)
+            else:
+                protected_objects = list(exc.restricted_objects)
+            msg = f'Unable to delete object. {len(protected_objects)} dependent objects were found: '
+            msg += ', '.join([f'{obj} ({obj.pk})' for obj in protected_objects])
+            return Response({'detail': msg}, status=409)
+        if isinstance(exc, AbortRequest):
+            return Response({'detail': exc.message}, status=400)
+        return None
+
     # Creates
 
     def create(self, request, *args, **kwargs):
+        # If background processing was requested for a bulk (list) create, validate and enqueue.
+        # Single-object creates always run synchronously.
+        if (response := self._maybe_background_bulk_create(request)) is not None:
+            return response
+
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
         bulk_create = getattr(serializer, 'many', False)
--- a/netbox/netbox/api/viewsets/__init__.py
+++ b/netbox/netbox/api/viewsets/__init__.py
@@ -227,6 +227,7 @@ class NetBoxModelViewSet(
         bypasses dispatch(). NOTE: dispatch() does not yet call this helper itself — see the
         PR description for the proposed consolidation.
         """
+        logger = logging.getLogger(f'netbox.api.views.{self.__class__.__name__}')
         if isinstance(exc, (ProtectedError, RestrictedError)):
             if type(exc) is ProtectedError:
                 protected_objects = list(exc.protected_objects)
@@ -234,8 +235,10 @@ class NetBoxModelViewSet(
                 protected_objects = list(exc.restricted_objects)
             msg = f'Unable to delete object. {len(protected_objects)} dependent objects were found: '
             msg += ', '.join([f'{obj} ({obj.pk})' for obj in protected_objects])
+            logger.warning(msg)
             return Response({'detail': msg}, status=409)
         if isinstance(exc, AbortRequest):
+            logger.debug(exc.message)
             return Response({'detail': exc.message}, status=400)
         return None
 
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -152,7 +152,10 @@ class SequentialBulkCreatesMixin:
     """
     def create(self, request, *args, **kwargs):
         # If background processing was requested for a bulk (list) create, validate and enqueue.
-        if (response := self._maybe_background_bulk_create(request)) is not None:
+        # _maybe_background_bulk_create() comes from BackgroundOperationMixin; fall back to "no
+        # background" so this mixin remains usable on its own (e.g. in custom viewsets).
+        maybe_background = getattr(self, '_maybe_background_bulk_create', lambda request: None)
+        if (response := maybe_background(request)) is not None:
             return response
 
         with transaction.atomic(using=router.db_for_write(self.queryset.model)):
@@ -200,7 +203,9 @@ class BulkUpdateModelMixin:
 
         # If background processing was requested, enqueue a job and return immediately.
         # The payload is captured here, before the request.data mutation below.
-        if self._background_requested(request):
+        # _background_requested() comes from BackgroundOperationMixin; fall back to "no
+        # background" so this mixin remains usable on its own (e.g. in custom viewsets).
+        if getattr(self, '_background_requested', lambda request: False)(request):
             action = 'bulk_partial_update' if partial else 'bulk_update'
             return self._enqueue_bulk_job(request, action, payload=list(request.data))
 
@@ -259,7 +264,9 @@ class BulkDestroyModelMixin:
         serializer.is_valid(raise_exception=True)
 
         # If background processing was requested, enqueue a job and return immediately.
-        if self._background_requested(request):
+        # _background_requested() comes from BackgroundOperationMixin; fall back to "no
+        # background" so this mixin remains usable on its own (e.g. in custom viewsets).
+        if getattr(self, '_background_requested', lambda request: False)(request):
             return self._enqueue_bulk_job(request, 'bulk_destroy', payload=list(request.data))
 
         qs = self.get_bulk_destroy_queryset().filter(
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -10,6 +10,7 @@ from rest_framework.reverse import reverse
 from core.models import ObjectType
 from extras.models import ExportTemplate
 from netbox.api.serializers import BulkOperationSerializer
+from netbox.jobs import AsyncAPIJob
 from utilities.exceptions import RQWorkerNotRunningException
 from utilities.rqworker import any_workers_for_queue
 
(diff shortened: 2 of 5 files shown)


jnovinger:
## Manual testing

Verified end-to-end against a live `runserver` + `rqworker` (the real deferred path, not the test runner's inline mode).

### Setup

```bash
python netbox/manage.py runserver 0.0.0.0:8005
python netbox/manage.py rqworker        # separate shell; must be running

TOKEN='<write-enabled token>'
BASE='http://localhost:8005'
```

Poll a job by GETting its URL until `status.value` is terminal.

### Scenarios

**1. Bulk create → completed**
```bash
curl -s -D - -X POST -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  "$BASE/api/dcim/regions/?background=true" \
  --data '[{"name":"Manual One","slug":"manual-one"},{"name":"Manual Two","slug":"manual-two"}]'
```
→ `202 Accepted` + `Location` header + `{"job":{"id":N,"url":...,"status":"pending"}}`. Poll the job → `completed`, `status_code: 201`, two objects, URLs begin with `http://`. GET an id to confirm it exists.

**2. Bulk update (PATCH) → completed, DB changed**
```bash
curl -s -X PATCH -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  "$BASE/api/dcim/regions/?background=true" \
  --data '[{"id":<id1>,"description":"patched"},{"id":<id2>,"description":"patched"}]' | jq '.job'
```
→ poll → `completed`, `status_code: 200`; GET confirms `description` changed. (PUT with a full representation behaves the same.)

**3. All-or-nothing failure → failed + rollback**
```bash
curl -s -X PATCH -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  "$BASE/api/dcim/regions/?background=true" \
  --data '[{"id":<id1>,"description":"shouldnotstick"},{"id":<id2>,"slug":""}]' | jq '.job'
```
→ `202`, then poll:
```json
{
  "status": "failed",
  "status_code": 400,
  "data": {"slug": ["This field may not be blank."]},
  "error": "Request failed with status 400."
}
```
GET `<id1>` confirms its description is unchanged — the whole batch rolled back.

**4. Protected-dependent delete → failed with 409**
Background-deleting an object with a `PROTECT`ed dependent (e.g. a Manufacturer referenced by a DeviceType) → `failed`, `status_code: 409`, `data.detail` lists the dependent(s); the object is not deleted.

**5. `If-Match` + background → synchronous 400** (no job created).

**6. Single-object create / GET with `?background=true`** → parameter ignored, processed synchronously (`201` / `200`), no job.

**7. No worker running** → `503 Service Unavailable` (no job enqueued).

All scenarios passed; job results mirror the synchronous response shape, handled 4xx rejections terminate as `failed` (not `errored`), and bulk operations preserve all-or-nothing semantics.

jnovinger:

jnovinger:
All five addressed in `eb5c9e0`.

> Hardcoded `localhost` in the reconstructed `WSGIRequest` makes object URLs in `job.data` read `http://localhost/...` instead of the real server.

Agreed, fixed. We now capture `request.scheme` and `request.get_host()` at enqueue time and pass them to the worker, which populates `SERVER_NAME`/`SERVER_PORT`/`wsgi.url_scheme` from the real values. Added a test asserting result URLs reflect the request host.

> `SequentialBulkCreatesMixin` has an implicit, unenforced dependency on `BackgroundOperationMixin` (AttributeError if a viewset inherits one without the other).

Real coupling, but we're not adding a guard. It's the same shape as the bulk mixins' existing reliance on the surrounding `NetBoxModelViewSet`, and a fail-fast `hasattr`/`__init_subclass__` check is the same cross-mixin coupling we pulled out earlier in review. Our preference is to keep the mixins free of that. These mixins are intended for use within `NetBoxModelViewSet`, which always includes `BackgroundOperationMixin`; treating that as a documented assumption.

> `exception_to_response()` drops the WARNING-level log that `dispatch()` emits for `ProtectedError`/`RestrictedError`.

Agreed, fixed. `exception_to_response()` now emits the same warning for protected/restricted deletes, and the same debug log for `AbortRequest`, that `dispatch()` does.

> `drf_request._authenticator = None` sets a non-existent DRF attribute / no-op; should be removed.

Removed, though the premise is off: `_authenticator` is a real `Request` attribute (`rest_framework/request.py`), read by `successful_authenticator` and set by `_authenticate()`; DRF itself sets it to `None` when no authenticator succeeds. The conclusion holds anyway. Setting `request.user` already populates the user cache so DRF never lazily authenticates, and nothing on the worker's action path reads the authenticator (we bypass `check_permissions` and enforce object permissions via `queryset.restrict()`). It was inert, so there's no reason to touch a private attribute we don't need.

> Minor: double `job.save()` on the success path.

Agreed, fixed. Dropped the redundant success-path save; `job.data` is persisted by `terminate()`.

All of the above is in `eb5c9e0`; tests pass.

jnovinger:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. See reply comment above for context on how the review items were resolved.

jnovinger:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Fixed the IPv6 issue.

pheus (changes requested):
Thanks for the work on this. This looks good to me, but I  have two concerns I’d like to see addressed before merge:

1. **The bulk mixins now have an implicit dependency on `BackgroundOperationMixin`.**  
   `BulkUpdateModelMixin` and `BulkDestroyModelMixin` call `_background_requested()` directly, which means those mixins can now raise `AttributeError` if used without `BackgroundOperationMixin`. That may be fine for core `NetBoxModelViewSet`, but it feels a little fragile for plugin/custom viewset composition. I think a small fallback via `getattr(..., lambda request: False)` would preserve the previous standalone behavior and make this safer.

2. **The custom exception-to-response path should have automated coverage.**  
   Since the background worker bypasses some of the normal request dispatch/exception handling behavior, the new `exception_to_response()` helper is important. I’d feel better with a focused test for something like `ProtectedError`/`RestrictedError` during a background bulk delete, asserting that the job records the expected API-shaped response instead of surfacing as an unexpected job error.

jnovinger:
> Thanks for the work on this. This looks good to me, but I have two concerns I’d like to see addressed before merge:
> 
>     1. **The bulk mixins now have an implicit dependency on `BackgroundOperationMixin`.**
>        `BulkUpdateModelMixin` and `BulkDestroyModelMixin` call `_background_requested()` directly, which means those mixins can now raise `AttributeError` if used without `BackgroundOperationMixin`. That may be fine for core `NetBoxModelViewSet`, but it feels a little fragile for plugin/custom viewset composition. I think a small fallback via `getattr(..., lambda request: False)` would preserve the previous standalone behavior and make this safer.
> 
>     2. **The custom exception-to-response path should have automated coverage.**
>        Since the background worker bypasses some of the normal request dispatch/exception handling behavior, the new `exception_to_response()` helper is important. I’d feel better with a focused test for something like `ProtectedError`/`RestrictedError` during a background bulk delete, asserting that the job records the expected API-shaped response instead of surfacing as an unexpected job error.

Done and done!

pheus (approved):
Thanks for the quick follow-up!

netbox/netbox/api/viewsets/__init__.py
```diff
@@ -215,9 +216,40 @@ def dispatch(self, request, *args, **kwargs):
...
 
+    def exception_to_response(self, exc):
+        """
+        Translate a NetBox/Django exception that is not a DRF APIException into the same
+        Response that dispatch() would return for it. Returns None if the exception is not
+        one this method handles (the caller should then re-raise or defer to DRF).
+
+        This mirrors the except clauses in dispatch(); it is also called by the background
+        job runner (netbox.jobs.AsyncAPIJob), which executes action methods directly and so
+        bypasses dispatch(). NOTE: dispatch() does not yet call this helper itself — see the
```
jeremystretch, line 227: In-code comments and docstrings should never reference a PR description. (The comment doesn't even indicate _which_ PR.) If it's important, it needs to be included in the docstring.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -18,6 +25,90 @@
...
+    This mixin overrides no framework methods; the bulk action methods call its helpers.
+    """
+
+    def _background_requested(self, request):
+        """Return True if background processing was requested for this write."""
+        if request.method not in ('POST', 'PUT', 'PATCH', 'DELETE'):
+            return False
+        return request.query_params.get('background', '').lower() == 'true'
+
+    def _maybe_background_bulk_create(self, request):
```
jeremystretch, line 45: The name of this method is very confusing. Something like `_handle_background_request()` would better convey its function.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -60,6 +151,13 @@ class SequentialBulkCreatesMixin:
     appropriately.
     """
     def create(self, request, *args, **kwargs):
+        # If background processing was requested for a bulk (list) create, validate and enqueue.
+        # _maybe_background_bulk_create() comes from BackgroundOperationMixin; fall back to "no
+        # background" so this mixin remains usable on its own (e.g. in custom viewsets).
+        maybe_background = getattr(self, '_maybe_background_bulk_create', lambda request: None)
```
jeremystretch, line 157: We're pre-validating for creates but not updates or deletes; this results in an inconsistent experience for the API client. We should handle all three operations consistently.

Additionally, validation itself might incur a substantial overhead, partially negating the benefit of deferring an operation to a background job. IMO we should defer validation entirely.

netbox/netbox/jobs.py
```diff
@@ -243,3 +254,127 @@ def run(self, view_cls, request, **kwargs):
...
+    The viewset's action method is re-invoked inside the worker against a reconstructed
+    request, so the synchronous and background code paths are identical (validation,
+    transaction semantics, object permissions, and change logging all behave the same).
+    The action's serialized Response is captured into the job's data.
+    """
+    class Meta:
+        name = 'Async API Request'
+
+    @staticmethod
+    def _build_request(payload, method, request_id, scheme, host):
```
jeremystretch, line 272: NetBox already provides a `copy_safe_request()` utility for this, which should be employed here.

jeremystretch (commented):
@jnovinger @pheus I realize this has already been merged, but I wanted to raise some additional considerations that I think are worth addressing prior to release.

PR #22453, opened 2026-06-15
Closes #22353: Remove deprecated type parameters from GraphQL filter lookups

### LINKED ISSUE #22353: Remove extraneous type parameters from GraphQL lookups
### Proposed Changes

Remove the type parameters from `StrFilterLookup`, `DateFilterLookup`, `TimeFilterLookup`, and `DatetimeFilterLookup` classes in Django filters.

### Justification

[strawberry-graphql-django v0.86.0](https://github.com/strawberry-graphql/strawberry-django/releases/tag/0.86.0) now raises a warning when unnecessary type parameters are present on these lookups.

pheus:
### Closes: #22353

This updates `strawberry-graphql-django` from `0.85.0` to `0.86.1` and removes explicit type parameters from `StrFilterLookup`, `DateFilterLookup`, `TimeFilterLookup`, and `DatetimeFilterLookup` annotations across GraphQL filter classes.

The dependency bump is intentional and required for this change. Strawberry-Django `0.86.x` treats these lookup classes as concrete lookup types, so the old parameterized form is now redundant and emits deprecation warnings. The previously pinned version does not support the unparameterized date/time lookup annotations used by this change, which is why the library update is included in this PR.

Existing optional annotations and filter field declarations are otherwise left unchanged.

#### JSON lookup handling

`JSONLookup` is handled separately because JSON values are string-backed and are not model-backed `DateField`, `DateTimeField`, or `TimeField` values.

Previously, NetBox used annotations such as `DateFilterLookup[str]`, `DatetimeFilterLookup[str]`, and `TimeFilterLookup[str]` for JSON lookups. That kept these filters string-backed while still exposing the existing date/time lookup fields. With Strawberry-Django `0.86.x`, those upstream lookup classes are now concrete, so parameterizing them no longer preserves the old JSON-specific behavior and also triggers deprecation warnings.

To keep the JSON filter API backward-compatible, this PR uses NetBox-local string-backed date/time lookup input types for `date_lookup`, `datetime_lookup`, and `time_lookup`. This preserves the previous JSON lookup behavior and field shape without relying on deprecated upstream generic annotations.

#### Schema impact

For model-backed date/time filters, generated lookup input type names now use the concrete Strawberry-Django names. For example, `DateDateFilterLookup` becomes `DateFilterLookup`. The field shapes and scalar types remain aligned with the underlying model fields.

For JSON lookups, `date_lookup`, `datetime_lookup`, and `time_lookup` remain string-backed to match the previous behavior for JSON-stored values such as `parameters`, `prechange_data`, `postchange_data`, `custom_field_data`, and `local_context_data`.

Clients that reference generated GraphQL input type names directly, such as codegen or persisted-query tooling, should refresh their schema.

### DIFF
--- a/netbox/circuits/graphql/filters.py
+++ b/netbox/circuits/graphql/filters.py
@@ -1,4 +1,3 @@
-from datetime import date
 from typing import TYPE_CHECKING, Annotated
 
 import strawberry
@@ -62,9 +61,9 @@ class CircuitTerminationFilter(
     upstream_speed: Annotated['IntegerLookup', strawberry.lazy('netbox.graphql.filter_lookups')] | None = (
         strawberry_django.filter_field()
     )
-    xconnect_id: StrFilterLookup[str] | None = strawberry_django.filter_field()
-    pp_info: StrFilterLookup[str] | None = strawberry_django.filter_field()
-    description: StrFilterLookup[str] | None = strawberry_django.filter_field()
+    xconnect_id: StrFilterLookup | None = strawberry_django.filter_field()
+    pp_info: StrFilterLookup | None = strawberry_django.filter_field()
+    description: StrFilterLookup | None = strawberry_django.filter_field()
 
     # Cached relations
     _provider_network: Annotated['ProviderNetworkFilter', strawberry.lazy('circuits.graphql.filters')] | None = (
@@ -92,7 +91,7 @@ class CircuitFilter(
     TenancyFilterMixin,
     PrimaryModelFilter
 ):
-    cid: StrFilterLookup[str] | None = strawberry_django.filter_field()
+    cid: StrFilterLookup | None = strawberry_django.filter_field()
     provider: Annotated['ProviderFilter', strawberry.lazy('circuits.graphql.filters')] | None = (
         strawberry_django.filter_field()
     )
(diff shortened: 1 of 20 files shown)


pheus:

pheus:

pheus:
@claude /review

I checked this against the old `strawberry-graphql-django==0.85.0` lookup definitions and the generated NetBox schema, and I think both JSON-specific points are false positives for this PR’s compatibility goal.

The local JSON lookup types are intentionally preserving the old `DateFilterLookup[str]`, `TimeFilterLookup[str]`, and `DatetimeFilterLookup[str]` schema, rather than adopting the new model-backed `0.86.x` date/time lookup semantics.

In `0.85.0`, `TimeFilterLookup[T]` included:

```python
hour: ComparisonFilterLookup[int]
minute: ComparisonFilterLookup[int]
second: ComparisonFilterLookup[int]
date: ComparisonFilterLookup[int]
time: ComparisonFilterLookup[int]
```

and `DatetimeFilterLookup[T]` inherited from both `DateFilterLookup[T]` and `TimeFilterLookup[T]`. That means the legacy `DatetimeFilterLookup[str]` schema exposed `date` and `time` as integer comparison lookups, not as nested date/time lookup inputs.

For the same reason, `JSONTimeFilterLookup.date` and `.time` are kept intentionally. They are odd for model-backed `TimeField` usage, which is why upstream removed them in `0.86.x`, but they were present in the old schema that NetBox exposed for JSON lookups.

I also verified this by diffing the generated `JSONLookup` and the three `Str*FilterLookup` input types against the schema generated with `strawberry-graphql-django==0.85.0`; the JSON lookup schema is byte-identical. So changing these fields now would make the PR less backward-compatible.

The model-backed filters still use the new concrete strawberry-django lookup classes, while the JSON-specific local types preserve the previous string-backed JSON schema without triggering the deprecated generic lookup warnings.

jeremystretch (approved):

PR #22455, opened 2026-06-15
Closes #22451: Use factories for GraphQL schema extension initialization

### LINKED ISSUE #22451: Pass Strawberry GraphQL extension factories instead of extension instances
### Proposed Changes

Update NetBox's GraphQL schema extension configuration to pass Strawberry extension factories instead of pre-constructed extension instances.

With `strawberry-graphql` 0.316.0, passing an extension instance to `Schema(extensions=[...])` is deprecated and emits the following warning:

```python
netbox/netbox/graphql/schema.py:50: DeprecationWarning: Passing an extension instance to `extensions=[...]` is deprecated and will be removed in a future release. Pass the class itself, or a factory callable (e.g. `lambda: MyExtension(arg=...)`), so that a fresh extension is constructed for each request. See https://github.com/strawberry-graphql/strawberry/issues/4369.
```

NetBox currently constructs these GraphQL extensions eagerly in `get_schema_extensions()`:

```python
DjangoOptimizerExtension(prefetch_custom_queryset=True)
MaxAliasesLimiter(max_alias_count=settings.GRAPHQL_MAX_ALIASES)
QueryDepthLimiter(max_depth=max_depth)
```

Instead, return zero-argument factory callables, for example:

```python
lambda: DjangoOptimizerExtension(prefetch_custom_queryset=True)
lambda: MaxAliasesLimiter(max_alias_count=settings.GRAPHQL_MAX_ALIASES)
lambda: QueryDepthLimiter(max_depth=max_depth)
```

This preserves the current GraphQL behavior while aligning with Strawberry's updated extension API.

The existing GraphQL tests that inspect `get_schema_extensions()` should also be updated to account for the fact that the function now returns factories rather than instances.

### Justification

Strawberry 0.316.0 deprecates passing `SchemaExtension` instances directly to `Schema(extensions=...)`. Passing instances is still accepted for backward compatibility, but Strawberry now warns because the same extension object can be reused across requests, allowing request-scoped state such as `execution_context` to leak between concurrent executions.

Using factories ensures Strawberry creates a fresh extension instance per request, removes the deprecation warning, and avoids future breakage when support for passing instances is removed.

This is a small internal maintenance change with no intended user-facing behavior change.

pheus:
### Closes: #22451

This updates NetBox's GraphQL schema extension configuration to pass Strawberry extension factories instead of pre-constructed extension instances.

Strawberry 0.316.0 now deprecates passing extension instances to `Schema(extensions=...)`, as those instances may be reused across requests. Returning zero-argument factories preserves the existing extension configuration while ensuring Strawberry creates a fresh extension instance per request.

#### Changes

* Changed `get_schema_extensions()` to return extension factories.
* Preserved the current `DjangoOptimizerExtension`, `MaxAliasesLimiter`, and optional `QueryDepthLimiter` configuration.
* Updated GraphQL tests to instantiate returned factories only for assertion purposes.
* Continued passing factories into temporary Strawberry schemas.

### DIFF
--- a/netbox/netbox/graphql/schema.py
+++ b/netbox/netbox/graphql/schema.py
@@ -1,3 +1,5 @@
+from collections.abc import Callable
+
 import strawberry
 from django.conf import settings
 from strawberry.extensions import MaxAliasesLimiter, QueryDepthLimiter, SchemaExtension
@@ -18,6 +20,8 @@ from wireless.graphql.schema import WirelessQuery
 
 from .scalars import BigInt, BigIntScalar
 
+SchemaExtensionFactory = type[SchemaExtension] | Callable[[], SchemaExtension]
+
 
 @strawberry.type
 class Query(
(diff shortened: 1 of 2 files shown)


pheus:

pheus:

bctiemann (approved):

PR #22456, opened 2026-06-15
Closes #22395: Remove unused save() override on ManagedFileForm

### LINKED ISSUE #22395: Remove unused save() override on core.forms.ManagedFileForm
### Proposed Changes

Remove the `save()` method from `ManagedFileForm` (`netbox/core/forms/model_forms.py`, lines 117-124). The base class will then inherit `NetBoxModelForm.save()`, which is already what the one subclass invokes.

The method writes an uploaded file to disk with a raw `open(self.instance.full_path, 'wb+')`. Its only subclass, `ScriptFileForm` (`netbox/extras/forms/scripts.py:76`), overrides `save()` to write through the django-storages backend instead, and explicitly skips the base implementation with `super(ManagedFileForm, self).save(...)` (the inline comment at `scripts.py:115` reads "need to skip ManagedFileForm save method").

Once the override is gone, `ScriptFileForm` can drop the two-argument `super(ManagedFileForm, self).save()` back to a plain `super().save()`.

### Justification

`ManagedFileForm.save()` has no live caller. The only view touching either form, `ScriptModuleCreateView`, uses `ScriptFileForm`, and the REST API upload path (`ScriptModuleSerializer.create()`) writes through `storage.save()` as well. Nothing in the views, serializers, tests, or the documented plugin API reaches the base method (`ManagedFile` is internal, marked `_netbox_private`).

It is a remnant of #18680 (March 2025), which moved both upload paths onto django-storages but left the old form-level write in place. Removing it drops a redundant, divergent file-write path and the maintenance surface that comes with it. The change is behavior-preserving, since no code path executes the method today.

jnovinger:
### Closes: #22395

Removes the `save()` override on `core.forms.ManagedFileForm`, which wrote uploaded files to disk via a raw `open(self.instance.full_path, 'wb+')`. No code path reached it: its only subclass, `ScriptFileForm`, overrides `save()` to write through the django-storages backend and explicitly skipped the base implementation with `super(ManagedFileForm, self).save()`. With the override gone, that call simplifies back to a plain `super().save()`.

The override was a leftover from #18680, which moved both upload paths (the form and `ScriptModuleSerializer.create()`) onto django-storages but left the old form-level write in place. The change is behavior-preserving, since the method had no live caller.

### DIFF
--- a/netbox/core/forms/model_forms.py
+++ b/netbox/core/forms/model_forms.py
@@ -114,15 +114,6 @@ class ManagedFileForm(SyncedDataMixin, NetBoxModelForm):
 
         return self.cleaned_data
 
-    def save(self, *args, **kwargs):
-        # If a file was uploaded, save it to disk
-        if self.cleaned_data['upload_file']:
-            self.instance.file_path = self.cleaned_data['upload_file'].name
-            with open(self.instance.full_path, 'wb+') as new_file:
-                new_file.write(self.cleaned_data['upload_file'].read())
-
-        return super().save(*args, **kwargs)
-
 
 class ConfigFormMetaclass(forms.models.ModelFormMetaclass):
 
--- a/netbox/extras/forms/scripts.py
+++ b/netbox/extras/forms/scripts.py
@@ -112,5 +112,4 @@ class ScriptFileForm(ManagedFileForm):
             data = self.cleaned_data['upload_file']
             storage.save(filename, data)
 
-        # need to skip ManagedFileForm save method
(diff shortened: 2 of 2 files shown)


jnovinger:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Focus on whether removing ManagedFileForm.save() leaves any live caller or breaks the ScriptFileForm super() chain.

jeremystretch (approved):

PR #22458, opened 2026-06-15
Closes #22457: Use `hmac.compare_digest()` to authenticate API tokens

### LINKED ISSUE #22457: Use `hmac.compare_digest()` to authenticate API tokens
### NetBox version

v4.6.2

### Feature type

### Proposed functionality

Currently, token authenticate uses regular Python evaluation when authenticating a token value (e.g. `token == self.token`). We should instead use Python's [`hmac.compare_digest()`](https://docs.python.org/3/library/hmac.html#hmac.compare_digest) function for this.

### Use case

`hmac.compare_digest()` is constructed to mitigate timing attacks, which are theoretically possible using plain Python equality checking.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:

### DIFF
--- a/netbox/users/models/tokens.py
+++ b/netbox/users/models/tokens.py
@@ -282,7 +282,7 @@ class Token(models.Model):
         digest.
         """
         if self.v1:
-            return token == self.token
+            return hmac.compare_digest(token, self.plaintext)
         if self.v2:
             token = token.removeprefix(TOKEN_PREFIX)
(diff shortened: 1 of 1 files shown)


jeremystretch:

jeremystretch:

jeremystretch:

jnovinger (approved):

PR #22459, opened 2026-06-15
#21355 - Handle updates to denormalized data via PostgreSQL triggers

arthanson:
### Closes: #21355 

Replaces the Python `post_save`-based denormalization in `netbox.denormalized` and `dcim.signals` with PostgreSQL triggers, following the raw-SQL trigger pattern introduced by the ltree migration.

- Add a reusable `InstallDenormalizationTrigger` migration operation in `utilities/migration.py` that installs an `AFTER UPDATE` trigger on the source table to propagate mapped columns into a dependent table's cached columns. Supports one-hop related lookups (e.g. a Location's Site change also refreshes the dependent's region/site-group).
- Add migrations installing triggers that replace the `denormalized.register()` calls and the `dcim.signals` cache handlers:
    - CircuitTermination ← Site, Location
    - CableTermination ← Device, Rack, Location
    - Prefix / Cluster / WirelessLAN ← Site, Location (CachedScopeMixin scope columns)
    - Device components (Interface, etc.) ← Device (`_site`/`_location`/`_rack`)
- Remove `netbox/denormalized.py`, the `register()` calls, the `denormalized_fields` registry slot and docs entry, and the cache-maintenance signals (`sync_cached_scope_fields`, `handle_device_site_change`, and the component/cable-termination updates in the location/rack handlers). The authoritative FK cascades remain as signals and now drive the triggers.
- Add trigger tests for each app, including the bulk `QuerySet.update()` path.


arthanson:

arthanson:
pre claude review:

fix 1, 3, 4.

2: verified Django's base Operation already auto-captures constructor args and round-trips correctly, so squashmigrations works without it (same as the merged InstallLtreeTriggers). Adding it would be redundant.

bctiemann (approved):

netbox/utilities/migration.py
```diff
@@ -32,3 +34,146 @@ def custom_deconstruct(field):
...
+SITE_SCOPE_RELATED_MAPPINGS = (
+    {
+        'table': 'dcim_site',
+        'source_fk': 'site_id',
+        'mappings': {'_region_id': 'region_id', '_site_group_id': 'group_id'},
+    },
+)
+
+
+def cached_scope_triggers(dependent_table):
```
jeremystretch, line 154: I'd be more comfortable unwinding this into explicit InstallDenormalizationTrigger instances within each migration. I'd rather not risk complications in future migrations just to save a few lines of code. (Otherwise, we'd effectively never be able to modify the function anyway.)

netbox/dcim/signals.py
```diff
@@ -5,99 +5,53 @@
...
+
 @receiver(post_save, sender=Location)
 def handle_location_site_change(instance, created, **kwargs):
     """
-    Update child objects if Site assignment has changed. We intentionally recurse through each child
-    object instead of calling update() on the QuerySet to ensure the proper change records get created for each.
+    Cascade a Location's Site assignment down to the Racks, Devices, and PowerPanels it contains
+    (and to descendant Locations). The denormalized cache columns on cable terminations and device
+    components are maintained by PostgreSQL triggers, which fire on these Site/Location/Rack/Device
+    column writes.
```
jeremystretch, line 37: Omit irrelevant detail
```suggestion
    Cascade a Location's Site assignment down to the Racks, Devices, and PowerPanels it contains
    (and to descendant Locations).
```

netbox/dcim/signals.py
```diff
@@ -5,99 +5,53 @@
...
-        for model in COMPONENT_MODELS:
-            model.objects.filter(device__location__in=locations).update(_site=instance.site)
 
 
 @receiver(post_save, sender=Rack)
 def handle_rack_site_change(instance, created, **kwargs):
     """
-    Update child Devices if Site or Location assignment has changed.
+    Cascade a Rack's Site/Location assignment down to the Devices it contains. The denormalized cache
+    columns on those devices' components are maintained by PostgreSQL triggers.
```
jeremystretch, line 51: Omit irrelevant detail
```suggestion
    Cascade a Rack's Site/Location assignment down to the Devices it contains.
```

PR #22465, opened 2026-06-16
Closes #22169: Cache image file size on ImageAttachment

### LINKED ISSUE #22169: ImageAttachments view is slow with S3 storage
### NetBox Version

v4.5.9

### Python Version

3.12

### Area(s) of Concern

- [x] User Interface
- [ ] REST API
- [ ] GraphQL API
- [ ] Python ORM
- [ ] Other

### Observations

When images are stored in S3, the `/extras/image-attachments/?page=4&per_page=50` webpage takes about 4-13 seconds to load each page.  This is because it sends a HEAD to each object to read its size.

### Proposed Changes

Store the file size as a new `image_size` field on the `ImageAttachment` model.

jnovinger:
### Closes: #22169

The image attachment list view issues a storage request for every image on each render, because the `size` column reads `ImageAttachment.size`, which calls `self.image.size`. On S3 that is a HEAD request per object, so a page of 50 attachments takes 4-13 seconds to load.

This caches the file size in a new `image_size` field, populated in `save()` when the image is created or replaced. The `size` property now returns the cached value, falling back to a storage read only for rows where `image_size` has not yet been set. This follows the existing `image_height`/`image_width` fields, with one difference: `image_size` is nullable. There is no native `ImageField` `size_field`, existing rows predate the field, and storage reads can fail, so a null value means "not yet computed" rather than "zero".

Existing attachments keep the old (slow) fallback behavior until they are next saved. There were three options for backfilling them: a data migration, a management command, or nothing. A data migration would read every file during the upgrade, reintroducing the same storage round-trips I'm eliminating, at a point where the operator can't control timing. Doing nothing leaves existing installs slow with no remedy. I chose a management command, `populate_image_sizes`, so operators decide when to absorb the bulk read: it reads each file once, writes in batches, is safe to run on a live system, and is re-runnable (any file that can't be read is left null and retried next run). Operators with many existing attachments (particularly on remote storage) should run it once after upgrading; this is documented under Administration > Management Commands. The command reads sizes sequentially, one file at a time. It issues the same per-file request the list view did (`storage.size()`, a single HEAD on S3), just once per file instead of on every render. The storage API has no batch-metadata call, so files can't be read in bulk; if the sequential backfill proves too slow on very large datasets, adding concurrency (a thread pool of independent reads) is a straightforward follow-up. I'm open to revisiting the overall approach (for example, folding the backfill into the migration) if reviewers prefer.

Verified manually against a running instance: creating an attachment populates `image_size`, replacing the image recomputes it, the list and detail views render the cached value, legacy null rows render an em-dash rather than "0 bytes", and the backfill command populates readable files while skipping (and retrying) unreadable ones.

---

**Note on filtering:** `image_size` is filterable in both the REST API (`ImageAttachmentFilterSet`) and GraphQL (`ImageAttachmentFilter`), matching the existing `image_height`/`image_width` fields. This isn't core to the performance fix, but NetBox's `test_missing_filters` requires every model field to have a corresponding filter, so omitting it isn't an option; adding it to both interfaces keeps REST and GraphQL consistent. The GraphQL lookup uses `BigIntegerLookup` (not `IntegerLookup`) because `image_size` is a `PositiveBigIntegerField` and can exceed the 32-bit range.

Rebased on `main` to resolve a conflict with the GraphQL filter refactor (the `BigIntegerLookup` type it introduced is what the size filter now uses).

### DIFF
--- a/docs/administration/management-commands.md
+++ b/docs/administration/management-commands.md
@@ -1,0 +1,21 @@
+# Management Commands
+
+In addition to Django's built-in management commands, NetBox provides several commands of its own. These are run using `manage.py`:
+
+```
+cd /opt/netbox
+source /opt/netbox/venv/bin/activate
+python3 netbox/manage.py <command>
+```
+
+## populate_image_sizes
+
+Image attachments cache the size of their file in the database so that list views do not need to query the storage backend for each image when rendered. Attachments created or updated on a recent NetBox release have this value populated automatically; those created on an earlier release do not, and will fall back to querying the storage backend until they are next saved.
+
+This command populates the cached size for any image attachments that are missing it. Running it once after upgrading is recommended for deployments with many existing image attachments, particularly when using a remote storage backend (such as S3), where the per-image queries are most expensive.
+
+```
+python3 netbox/manage.py populate_image_sizes
+```
+
+This command is safe to run on a live system. It reads each file's size from the storage backend, so it may take some time to complete on a deployment with a large number of attachments; progress is reported as it runs. The command may also be run again safely: any attachments whose file could not be read (for example, due to a temporary storage outage) are left unchanged and retried on a subsequent run.
--- a/docs/administration/management-commands.md
+++ b/docs/administration/management-commands.md
@@ -8,6 +8,24 @@ source /opt/netbox/venv/bin/activate
 python3 netbox/manage.py <command>
 ```
 
+Run any command with `--help` to see its full set of arguments.
+
+## calculate_cached_counts
+
+Force a recalculation of all cached counter fields (for example, the device count shown on a site). NetBox keeps these counters current automatically; this command is useful to repair them if they have drifted.
+
+```
+python3 netbox/manage.py calculate_cached_counts
+```
+
+## nbshell
+
+Start the Django shell with all NetBox models already imported. See [NetBox Shell](./netbox-shell.md) for details.
+
+```
+python3 netbox/manage.py nbshell
+```
+
 ## populate_image_sizes
 
 Image attachments cache the size of their file in the database so that list views do not need to query the storage backend for each image when rendered. Attachments created or updated on a recent NetBox release have this value populated automatically; those created on an earlier release do not, and will fall back to querying the storage backend until they are next saved.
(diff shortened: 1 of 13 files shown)


jnovinger:

jnovinger:

> When a file is replaced but `_read_image_size()` fails, `image_size` retains the previous file's size, not a sentinel indicating "unknown". [...] Setting `image_size = None` in the replacement+failure case would be more honest and would make the management command retry it automatically on next run.

I went a different way than set-None. Assigning `None` on a failed read isn't safe here: the same `save()` block is entered on non-replacement saves where the cached size is still correct (notably deferred / `.only()` loads, where the change-detection key looks different but the file is unchanged), so a `None` there clobbers a good value.

Instead I tightened the detection itself. The recompute key is now `(name, image_height, image_width)` rather than name alone, so a same-name replacement whose dimensions differ is detected and the size recomputed (`dc637b7`). The one case left is same name AND same dimensions AND different bytes, which can't be distinguished without reading the file, the storage round-trip this whole change exists to avoid. The `size` property's lazy fallback still returns a correct value for any null row in the meantime.

> The command calls a single-underscore method from a different file. [...] Making `_read_image_size` public (`read_image_size`) or adding a small dedicated method (`fetch_size()`) would be cleaner.

Good catch. The command now calls the public `size` property. For the rows it touches (`image_size IS NULL`) the property does the same storage read, so the helper stays private and there's no new public surface (`320cad2`).

> `image_size` is added to `ImageAttachmentFilterSet.Meta.fields` [...] but `test_filtersets.py` has no `test_image_size` or `test_image_size_range` test case.

Added both, exact multi-value match and a `__gte`/`__lte` range (`b6cfc46`).

All three are pushed.

arthanson (changes requested):
I think having the dedicated document here raises a bit of a can of worms.. I'd expect this to list all the management commands available and that would probably be a good thing, but worth discussing in standup.

nb-shell has it's own page, runscript is documented in customization/custom-scripts.md, reindex in plugins/removal.md.  syncdatasource, trace_paths, rebuild_prefixes, calculate_cached_counts, renaturalize aren't documented.  We should probably have all of these listed in one place with at least a brief one-liner for them.

jnovinger:
> I think having the dedicated document here raises a bit of a can of worms.. I'd expect this to list all the management commands available and that would probably be a good thing, but worth discussing in standup.
> 
> nb-shell has it's own page, runscript is documented in customization/custom-scripts.md, reindex in plugins/removal.md. syncdatasource, trace_paths, rebuild_prefixes, calculate_cached_counts, renaturalize aren't documented. We should probably have all of these listed in one place with at least a brief one-liner for them.

Noted, I had actually shied away from that as a scope expansion thing, but happy to do it with a one-liner for each, if you'd like.

docs/administration/management-commands.md
```diff
@@ -0,0 +1,102 @@
...
+
+Start the Django shell with all NetBox models already imported. See [NetBox Shell](./netbox-shell.md) for details.
+
+```
+python3 netbox/manage.py nbshell
+```
+
+## populate_image_sizes
+
+!!! info "This command was introduced in NetBox v4.7."
```
arthanson: This references 4.7 but this is targeting main?

arthanson (changes requested):

docs/administration/management-commands.md
jnovinger: 🤦

arthanson (approved):

PR #22467, opened 2026-06-16
Fixes #22466: Fix test failure against SSL-enabled PosgtreSQL

### LINKED ISSUE #22466: Tests error when using a PostgreSQL database with SSL enabled
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. Enable SSL on the PostgreSQL database.
2. Run the current test suite.

(Note that the test suite in CI is currently passing as the test database does not support SSL.)

### Expected Behavior

The test suite should fully pass.

### Observed Behavior

Several tests under raise an error: `extras.tests.test_event_rules.EventRuleTestCase`: `django.db.utils.OperationalError: the connection is closed`. This is due to the introduction of `test_eventrule_script_action_with_object_image_files` under PR #22398, which spawns a new RQ worker:

```python
django_rq.get_worker().work(burst=True)
```

The chain of events:

1. `get_worker()` returns a `NetBoxRQWorker`, a subclass of `rq.Worker` — the **forking** worker. On `work()`, RQ `fork()`s a work-horse child to execute the queued job.
2. The child inherits the parent's already-open PostgreSQL connection — the live socket from the test's open transaction.
3. **Locally that socket is TLS-encrypted** (I confirmed: `TLSv1.3 / TLS_AES_256_GCM_SHA384`, against your local Postgres 16 which has SSL on). Parent and child then both read/write the *same* TLS session, whose state (record sequence numbers + MAC keys) cannot be shared across two processes.
4. The TLS stream desynchronizes → `OperationalError: consuming input failed: SSL error: decryption failed or bad record mac`. The connection is now dead.
5. Django marks it closed, so every alphabetically-later test in the class fails in `setUp` at `User.objects.create_user(...)` with *"the connection is closed."*

jeremystretch:
### Fixes: #22466

- Introduce a `run_rq_jobs()` utility method on RQQueueTestMixin for manually processing background jobs via a non-forking RQ worker (SimpleWorker)
- Replace independent calls to `worker.work()` with `run_rq_jobs()`

### DIFF
--- a/netbox/core/tests/test_api.py
+++ b/netbox/core/tests/test_api.py
@@ -321,9 +321,8 @@ class BackgroundTaskTestCase(RQQueueTestMixin, TestCase):
         # Enqueue & run a job that will fail
         queue = get_queue('default')
         job = queue.enqueue(self.dummy_job_failing)
-        worker = get_worker('default')
         with disable_logging():
-            worker.work(burst=True)
+            self.run_rq_jobs('default')
         self.assertTrue(job.is_failed)
         url = reverse('core-api:rqtask-requeue', args=[job.id])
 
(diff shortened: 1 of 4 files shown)


jeremystretch:

pheus (approved):

PR #22469, opened 2026-06-16
#22439 - Restrict custom links by object-level permissions

### LINKED ISSUE #22439: custom_links templatetag queries CustomLink without .restrict(), bypassing object-level permissions
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.12

### Steps to Reproduce

1. As an admin, create a CustomLink for an object type (e.g. Site) with link text/URL that renders visibly.
2. Configure a Permission granting a constrained role extras.view_customlink with an object-level constraint that excludes that CustomLink.
3. Assign that role to a non-admin user.
4. As that user, navigate to a Device detail view (or a Device list with the custom-link column enabled).

### Expected Behavior

The custom link should not be shown

### Observed Behavior

The custom link is shown

arthanson:
### Closes: #22439 

Applies restrict() to the custom_links template-tag queryset.

### DIFF
--- a/netbox/extras/templatetags/custom_links.py
+++ b/netbox/extras/templatetags/custom_links.py
@@ -33,8 +33,9 @@ def custom_links(context, obj):
     """
     Render all applicable links for the given object.
     """
+    user = context['request'].user
     object_type = ObjectType.objects.get_for_model(obj)
-    custom_links = CustomLink.objects.filter(object_types=object_type, enabled=True)
+    custom_links = CustomLink.objects.restrict(user, 'view').filter(object_types=object_type, enabled=True)
     if not custom_links:
         return ''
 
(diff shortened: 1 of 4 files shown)


arthanson:

arthanson:
per claude's review the flagged behavior is consistent with NetBox's standard model - no code change is warranted.

pheus (changes requested):
Thanks!

I think the table column path may still need the same treatment. The issue also mentions custom links shown as list/table columns, and those columns still seem to be built from all enabled `CustomLink` objects rather than the user-restricted queryset.

Could we also filter custom link columns by the current user’s `view_customlink` permissions, and add a small regression test for the list/table case?

arthanson:
" think the table column path may still need the same treatment. ' that took a bit more rework then I originally thought but working now.

pheus (approved):
Thanks for the great follow-up!

PR #22471, opened 2026-06-16
Closes #22393: Drop support for Redis 5.x

### LINKED ISSUE #22393: Drop support for Redis <6.0
### Deprecation Issue

netbox-community/netbox#22392

### Summary of Changes

Redis databases older than v6.0 will no loner be supported.

jeremystretch:
### Closes: #22393

- Bump the documented minimum version of Redis from 5.0 to 6.0
- Update the `check_redis_version()` system check to raise an error (not a warning) if Redis is not 6.0+

### DIFF
--- a/docs/installation/2-redis.md
+++ b/docs/installation/2-redis.md
@@ -10,9 +10,6 @@ sudo apt install -y redis-server
 
 Before continuing, verify that your installed version of Redis is at least v6.0:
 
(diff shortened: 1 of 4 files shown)


jeremystretch:

jeremystretch:
IMO the concern above can be disregarded: Our intention here is merely to raise an error _if_ the version can be determined.

jnovinger (approved):

PR #22473, opened 2026-06-16
#21367 - Add is_primary field to MAC address REST API serializer

arthanson:
### Closes: #21367 

adds is_primary to the mac_address serializer.

Optimizes MACAddress.is_primary to compare primary_mac_address_id (a column already loaded on the assigned interface) instead of dereferencing the related primary_mac_address object, eliminating an extra query per row in list contexts (REST API, UI tables, panels, GraphQL). Behavior is unchanged.


arthanson:

arthanson:
pre claude's review - optimized the cached_property lookup to avoid n+1 queries

bctiemann (approved):

Issue #22300, opened 2026-05-27
Drop the automatic reverse relationship defined by OwnerMixin


jeremystretch:
### NetBox version

v4.6.1

### Feature type

Change to existing functionality

### Proposed functionality

The OwnerMixin class adds an `owner` ForeignKey field to subclasses. This field automatically creates a reverse relationship named in the form `FOO_set`, e.g. `site_set` for the dcim.Site model.

```python
class OwnerMixin(models.Model):
    owner = models.ForeignKey(
        to='users.Owner',
        on_delete=models.PROTECT,
        blank=True,
        null=True
    )
```

This FR proposes removing this automatic reverse relationship via `related_name`:

```python
class OwnerMixin(models.Model):
    owner = models.ForeignKey(
        ...
        related_name='+',
    )
```

### Use case

The reverse relationships formed via the foreign key are not generally useful and can introduce name conflicts where two plugins introduce models with the same name which both subclass OwnerMixin.

### Database changes

This change will generate a migration, however it impacts only the ORM and does not effect any concrete changes in the database.

### External dependencies

N/A

PR #22474, opened 2026-06-16
#22300 - Drop reverse relationship defined by OwnerMixin

arthanson:
### Closes: #22300 

Drops the automatic reverse relationship defined by OwnerMixin and adds the associated migrations.


arthanson:

jeremystretch (changes requested):
Rather than introducing a bunch of new migrations, we can alter the existing migrations to set `related_name='+'`. This is safe because we're not actually altering the underlying schema: It's purely an application-level change.

jeremystretch (approved):

PR #22475, opened 2026-06-16
Closes #20897: Expose selection custom field labels in the REST API

arthanson:
### Closes: #20897 

Selection and multiple-selection custom fields are now rendered in the REST API as objects exposing both the stored value and its resolved label ({"value": …, "label": …}), matching the convention already used by built-in choice fields. The change affects read access only; values are still written by passing the raw value.

### DIFF
--- a/docs/customization/custom-fields.md
+++ b/docs/customization/custom-fields.md
@@ -104,9 +104,9 @@ Selection and multiple selection fields are returned as objects exposing both th
 
 ```json
     "custom_fields": {
-        "cluster": {
-            "value": 1,
-            "label": "cluster1"
+        "site_type": {
+            "value": "datacenter",
+            "label": "Data Center"
         },
         "regions": [
             {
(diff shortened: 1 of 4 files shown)


arthanson:

arthanson:
pre classed review 1 is known it's a breaking change.  2 fixed - now consistent across the board: read returns {value, label}; write takes the raw value; the object form is rejected with a 400 for both selection types - same as built-in choice fields.

docs/customization/custom-fields.md
```diff
@@ -100,6 +100,28 @@ When retrieving an object via the REST API, all of its custom data will be inclu
     ...
 ```
 
+Selection and multiple selection fields are returned as objects exposing both the stored value and its human-friendly label, following the same convention used by NetBox's built-in choice fields:
+
+```json
+    "custom_fields": {
+        "cluster": {
+            "value": 1,
```
jnovinger: This reads like a PK integer, but I think this would actually be a database stored string, yeah?

docs/customization/custom-fields.md
```diff
@@ -111,3 +133,5 @@ To set or change these values, simply include nested JSON data. For example:
     }
 }
 ```
+
+As with built-in choice fields, selection custom fields are written by passing the raw value (e.g. `"cluster": 1`), not the `{value, label}` object returned on read.
```
jnovinger: As with built-in choice fields, selection custom fields are written by passing the raw value (e.g. `"site_type": "datacenter"`), not the `{value, label}` object returned on read, right?

jnovinger (changes requested):

jnovinger (approved):

Issue #22231, opened 2026-05-18
Add "nulls first" parameter for custom field ordering


jeremystretch:
### NetBox version

v4.6.0

### Feature type

New functionality

### Proposed functionality

Add a new field (e.g. `nulls_first`) on the CustomField model which enables the user to specify whether null values appear first or last when ordering objects according to the field. Nulls are currently sorted first; we should probably retain this as the default behavior.

### Use case

This will provide an additional degree of control over the ordering logic which may be needed for certain use cases.

A use case for this was originally raised under FR netbox-community/netbox#19422 but did not include an implementation plan.

### Database changes

Add a boolean field named `nulls_first` to the CustomField model. This should default to True for all existing custom fields to retain the current behavior.

### External dependencies

N/A

PR #22476, opened 2026-06-16
#22231 - Add nulls-first parameter for custom field ordering

arthanson:
### Closes: #22231 

This PR adds a nulls_first boolean field to the CustomField model (defaulting to True to preserve current behavior), letting users control whether objects with no value for a custom field sort before or after populated values when a list is ordered by that field. 

Because custom field values are stored in a JSONB blob where an unset value is JSON null rather than SQL NULL, PostgreSQL's NULLS FIRST/LAST clause has no effect; the implementation works around this in CustomFieldColumn.order() by extracting the value as text (which yields SQL NULL for JSON null) and ordering by an explicit null-rank ahead of the JSONB value. 

The field is wired through the model, migration, API serializer, filterset, GraphQL filter, table column, and all four forms (model, bulk-edit, bulk-import, filter), with tests covering filtering and null-placement ordering, plus updated documentation.


arthanson:

arthanson:
1, 3 for Claude review fixed. 2 is an observation, 4 - custom_filed name is 50 char max so name is within bounds.

docs/models/extras/customfield.md
```diff
@@ -109,6 +109,10 @@ Choice sets may optionally define colors for individual values. Colored choices
 
 If enabled, values from this field will be automatically pre-populated when cloning existing objects.
 
+### Nulls First
+
+When ordering objects by this custom field, controls whether objects with no value (null) are sorted before (enabled, the default) or after (disabled) objects that have a value.
```
jeremystretch: Minor suggestion to improve readability:
```suggestion
When ordering objects by this custom field, controls whether objects with no value (null) are sorted before or after objects that have a value. This option is enabled by default.
```

netbox/netbox/tables/columns.py
```diff
@@ -514,15 +515,38 @@ def __init__(self, customfield, *args, **kwargs):
         kwargs['accessor'] = Accessor(f'custom_field_data__{customfield.name}')
         if 'verbose_name' not in kwargs:
             kwargs['verbose_name'] = customfield.label or customfield.name
-        # We can't logically sort on FK values
+        # We can't logically sort on FK values, nor on multi-value (array) fields
         if customfield.type in (
             CustomFieldTypeChoices.TYPE_OBJECT,
-            CustomFieldTypeChoices.TYPE_MULTIOBJECT
+            CustomFieldTypeChoices.TYPE_MULTIOBJECT,
+            CustomFieldTypeChoices.TYPE_MULTISELECT,
```
jeremystretch: While I understand the motivation for the change, we should _probably_ revert this as it's not technically in-scope for the FR.

jeremystretch (approved):
Nice work. Two minor suggestions I'll leave to your discretion.

PR #22490, opened 2026-06-18
Closes #22487: Remove release notes for NetBox v2.x

### LINKED ISSUE #22487: Drop release notes for NetBox v2.x
### Change Type

Removal

### Area

Other

### Proposed Changes

Remove the release notes for NetBox v2.0 through v2.11. These are all 5+ years old and should no longer be needed for reference when upgrading. (They will still be available by checking out a prior NetBox release.)

mburggraf:
Removed the release notes from index.md and removed the stand-alone files.


jeremystretch (changes requested):
Thanks @mburggraf. We also need to remove the navigation links from `mkdocs.yml`.

mburggraf:
Sorry, must have missed those.
I removed them as well.

jeremystretch (approved):
Thanks @mburggraf!

PR #22492, opened 2026-06-18
Fixes #21310: Fix IntegrityError exception when `AUTH_LDAP_MIRROR_GROUPS` is enabled

### LINKED ISSUE #21310: LDAP group-lookup fails with row containing null
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.1

### Python Version

3.13

### Steps to Reproduce

1. Create user-groups in netbox with permissions attributed to them
2. Have a LDAP server that contains groupOfNames with your users in it under a lower ou
3. Create a ldap_config.py that contains these options
```
AUTH_LDAP_GROUP_SEARCH = LDAPSearch("ou=groups,dc=example,dc=com",
                                    ldap.SCOPE_SUBTREE,
                                    "(objectClass=groupOfNames)")
AUTH_LDAP_GROUP_TYPE = GroupOfNamesType()

AUTH_LDAP_MIRROR_GROUPS = True
```
4. Log in with a user that matches one of the groups

### Expected Behavior

As the group search returns multiple groups, some of them are matching the names of local groups on the netbox instance, I expect the user to be inherited those permissions

### Observed Behavior

I get errors that say : 

```
null value in column "name" of relation "users_group" violates not-null constraint
DETAIL:  Failing row contains (48, null, ). while authenticating altf4

```

Errors go away when removing `AUTH_LDAP_MIRROR_GROUPS = True`. But the groups don't match

Even in DEBUG=True, it seems to just miss a name to pass through from Django-LDAP (I have confirmed that the data is well received from the LDAP)

When debug mode is disabled, I get this on the web-page
```

null value in column "name" of relation "users_group" violates not-null constraint
DETAIL:  Failing row contains (60, null, ).

Python version: 3.13.5
NetBox version: 4.5.1
Plugins: None installed
```

jeremystretch:
### Fixes: #21310

- Replicate upstream defensive fix to our custom `_mirror_groups()` method
- Add a test

### DIFF
--- a/netbox/netbox/authentication/misc.py
+++ b/netbox/netbox/authentication/misc.py
@@ -31,7 +31,9 @@ def _mirror_groups(self):
     Mirrors the user's LDAP groups in the Django database and updates the
     user's membership.
     """
-    target_group_names = frozenset(self._get_groups().get_group_names())
+    target_group_names = frozenset(
+        filter(None, self._get_groups().get_group_names())
+    )
     current_group_names = frozenset(
         self._user.groups.values_list("name", flat=True).iterator()
     )
--- a/netbox/netbox/tests/test_authentication.py
+++ b/netbox/netbox/tests/test_authentication.py
@@ -1,4 +1,5 @@
 import datetime
+from unittest.mock import MagicMock
 
 from django.conf import settings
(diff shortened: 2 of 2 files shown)


jeremystretch:

jnovinger (approved):

PR #22493, opened 2026-06-18
Fixes #22480: Revert forced handling of image attachments as downloads

### LINKED ISSUE #22480: Front/rear images of devices no longer open in browser, only download
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.2

### Python Version

3.13

### Steps to Reproduce

1. Lookup a device or device-type that already has a front/rear image attached to it.
2. Click on the image.

### Expected Behavior

The image should pop-up in full size in the browser window.

### Observed Behavior

The browser starts downloading a copy of the image.

jeremystretch:
### Fixes: #22480

- Remove the `Content-Disposition: attachment` response header from MediaView
- Add the `Content-Security-Policy: sandbox; default-src 'none'` response header as a secondary layer of protection

### DIFF
--- a/netbox/netbox/tests/test_views.py
+++ b/netbox/netbox/tests/test_views.py
@@ -124,8 +124,8 @@ class MediaViewTestCase(TestCase):
         with patch('netbox.views.misc.serve', return_value=HttpResponse(status=200)):
             response = self.client.get(url)
         self.assertHttpStatus(response, 200)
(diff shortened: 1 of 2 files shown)


jeremystretch:

jeremystretch:

jeremystretch:
The above concern can be disregarded.

arthanson (approved):

PR #22495, opened 2026-06-19
Closes #21710: Support multi-select fields for module type profile array enums

ITJamie:
### Closes: #21710

Implemented support for JSON Schema array properties whose items define an enum:
- Those fields now render as MultipleChoiceField instead of the old comma-separated array text field.
- Non-enum arrays still use the existing array input behavior.
- Module and module type detail panels now display stored list values as comma-separated text.
- Added tests for:
  - array + items.enum field generation
  - plain arrays still using the old path
  - form clean/save round-trip
  - list rendering helper
  - module type view rendering for stored enum-array values

### DIFF
--- a/netbox/utilities/templatetags/builtins/filters.py
+++ b/netbox/utilities/templatetags/builtins/filters.py
@@ -1,7 +1,6 @@
 import datetime
 import json
 import re
-from collections.abc import Iterable, Mapping
 
 import yaml
 from django import template
@@ -24,7 +23,6 @@ __all__ = (
     'content_type_id',
     'fgcolor',
     'getattr_',
-    'is_iterable',
     'isodate',
     'isodatetime',
     'isotime',
@@ -121,14 +119,6 @@ def placeholder(value):
         return value
 
     return mark_safe('<span class="text-muted">&mdash;</span>')
-
-
-@register.filter()
-def is_iterable(value):
-    """
-    Return whether the value is an iterable collection suitable for list rendering.
-    """
-    return isinstance(value, Iterable) and not isinstance(value, (str, bytes, Mapping))
 
 
 @register.filter()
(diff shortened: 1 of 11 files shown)


ITJamie:
example screenshots 
Screenshot 2026-06-19 at 23 24 09

Screenshot 2026-06-19 at 23 24 15

ITJamie:
The `moduletype:list_objects_with_permission` query count increased by 1 because the new regression fixture adds a `ModuleTypeProfile` with JSON Schema plus a `ModuleType` that carries `attribute_data`. 

That extra profile-backed object causes the module type list view to issue one additional `extras_tag` query, so the baseline in `netbox/dcim/tests/query_counts.json` was updated from `21` to `22`.

jeremystretch:

netbox/utilities/templatetags/builtins/filters.py
```diff
@@ -121,6 +123,14 @@ def placeholder(value):
     return mark_safe('<span class="text-muted">&mdash;</span>')
 
 
+@register.filter()
+def is_iterable(value):
```
jeremystretch: Rather than introduce a new template filter specifically for this case, let's tweak `parameters()` on ModuleType to present the iterable as a string:

```python
@property
def attributes(self):
    ...
    for name, options in self.profile.schema.get('properties', {}).items():
        key = options.get('title', title(name))
        value = self.attribute_data.get(name)
        if isinstance(value, (list, tuple)):
            value = ', '.join(str(v) for v in value)
        attrs[key] = value
    return dict(sorted(attrs.items()))
```

This will clean up the panel displays as well as other instances, such as inside a table column:

Image

jeremystretch (changes requested):
Thanks for tackling this @ITJamie. I just have one request regarding the template cleanup.

jeremystretch:

jeremystretch:
1. This is fine.
2. Not an issue AFAICT.
3. I'm okay with leaving the old path in place as a safeguard.
4. This is fine.

jeremystretch (approved):

PR #22505, opened 2026-06-22
Fixes #22470: Restrict TaggedItem API to visible target objects

### LINKED ISSUE #22470: Tagged-objects REST API should respect target object view permissions
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

Minimal precondition: a non-superuser with `extras.view_taggeditem`, but without view permission for a tagged target object, can read a brief representation of that object through the tagged-objects endpoint.

1. Start from a clean NetBox Community v4.6.3 installation.
2. Create a target object the test user must not be able to view directly. For example, create a device named `device-hidden` assigned to a tenant or scope outside the user's permitted set.
3. Apply any tag to `device-hidden`.
4. Create a non-superuser account and grant only:
   - `extras.view_taggeditem`
   - no view permission for `device-hidden`, either by omitting `dcim.view_device`, or by granting constrained `dcim.view_device` permission that excludes `device-hidden`
5. Authenticate as that user.
6. Confirm that the object is not visible through its own endpoint:
   ```http
   GET /api/dcim/devices/<device-hidden-id>/ HTTP/1.1
   Host: netbox.example.test
   Authorization: Token <redacted>
   Accept: application/json
   ```
   The expected result for this request is that direct access is denied, for example with `403`, or that the object is not returned because it is outside the user's constrained permissions.
7. Query the tagged-objects endpoint as the same user:
   ```http
   GET /api/extras/tagged-objects/?tag=<tag-slug> HTTP/1.1
   Host: netbox.example.test
   Authorization: Token <redacted>
   Accept: application/json
   ```
8. Inspect the response entry whose `object_id` matches `device-hidden`.

### Expected Behavior

The tagged-objects API should not disclose a representation of a tagged object that the requesting user cannot view through that object's own model API endpoint.

A user who can view `extras.TaggedItem`, but cannot view the underlying tagged object, should not be able to learn that object's identity or attributes from the tagged-objects response.

### Observed Behavior

The response entry for `device-hidden` contains a populated `object` field with the brief nested serialization of the hidden object.

For a device, this brief representation includes `id`, `url`, `display`, `name`, and `description`, disclosing the object's existence, name, description, and direct API URL to a user who does not have view permission for that device.

`TaggedItemViewSet` restricts access to the `TaggedItem` queryset. `TaggedItemSerializer.get_object()` then renders the related `content_object` with that object's nested serializer, but the target object's own view restrictions are not applied in this serialization path.

pheus:
### Closes: #22470

This updates the TaggedItem API to respect object-level view permissions for each tagged object.

Tagged items are now filtered against the permissions of their target objects, preventing users from discovering objects they are not allowed to view through the tags endpoint. It also handles missing or deleted tagged targets more safely during serialization.

### DIFF
--- a/docs/plugins/development/models.md
+++ b/docs/plugins/development/models.md
@@ -162,6 +162,13 @@ For more information about database migrations, see the [Django documentation](h
 
 ::: netbox.models.features.TagsMixin
 
+!!! note "Tagged objects and the REST API"
+    The `/api/extras/tagged-objects/` endpoint returns only tagged objects the requesting user is
+    permitted to view, which requires the target model's manager to provide `RestrictedQuerySet` (as
+    `NetBoxModel` does). A model that uses `TagsMixin` without `NetBoxModel` has a plain manager, so
+    its tagged objects are omitted from that endpoint unless its view permission is exempted via
+    [`EXEMPT_VIEW_PERMISSIONS`](../../configuration/security.md#exempt_view_permissions).
+
 ## Custom Model Features
 
 In addition to utilizing the model features provided natively by NetBox (listed above), plugins can register their own model features. This is done using the `register_model_feature()` function from `netbox.utils`. This function takes two arguments: a feature name, and a callable which accepts a model class. The callable must return a boolean value indicting whether the given model supports the named feature.
--- a/netbox/extras/api/serializers_/tags.py
+++ b/netbox/extras/api/serializers_/tags.py
@@ -58,11 +58,15 @@ class TaggedItemSerializer(BaseModelSerializer):
     def get_object(self, obj):
         """
         Serialize a nested representation of the tagged object.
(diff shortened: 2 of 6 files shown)


pheus:

pheus:
@claude /review

1. Addressed with a docs note. I kept the fail-closed behavior because if a plugin model has no `restrict()`, we can’t safely check object-level permissions.
2. Fixed. Exempt models now match by `content_type_id` directly, avoiding the no-op subquery. Added test coverage for specific and blanket exemptions.
3. Left as-is. The race can only cause under-inclusion for one response, not data exposure.
4. Left as-is. The fallback paths are intentional, and the new `None` guard avoids a possible 500.

PR #22506, opened 2026-06-22
Fixes #22489: Update member Device search cache when renaming a Virtual Chassis

### LINKED ISSUE #22489: Virtual Chassis reference on Device is not updated when Virtual Chassi changes name
### NetBox Edition

NetBox Community

### NetBox Version

v.4.6.3

### Python Version

3.12

### Steps to Reproduce

1. Create a Virtual Chassis "VC1".
2. Create a Device "Switch-1" and assign it to VC1 in position 1.
3. Change the name of VC1 to VC-test.
4. Search for "VC1" in global search, note that you will see Switch-1 with "VC1" as value in the Virtual Chassis field, even though that is not the name of the Virtual Chassis anymore.

### Expected Behavior

I expect the global search of "VC1" to not return any results. 

### Observed Behavior

The global search of "VC1" returns a result for Switch-1, suggesting that the value is not updated when the Virtual Chassis is renamed.

pheus:
### Closes: #22489

This PR adds a targeted `post_save` signal handler for `VirtualChassis` updates. When a virtual chassis is renamed, the handler refreshes the cached `virtual_chassis` search field for its member devices so global search no longer returns stale matches for the old VC name.

The handler skips creates, raw saves, and targeted saves that do not update the `name` field, keeping the cache refresh limited to the cases where the indexed value can actually change.

### DIFF
--- a/netbox/dcim/signals.py
+++ b/netbox/dcim/signals.py
@@ -6,6 +6,7 @@ from django.dispatch import receiver
 
 from dcim.choices import CableEndChoices, LinkStatusChoices
 from ipam.models import Prefix
+from netbox.search.backends import search_backend
 from virtualization.models import Cluster, VMInterface
 from wireless.models import WirelessLAN
 
@@ -33,6 +34,7 @@ from .models import (
     VirtualChassis,
 )
 from .models.cables import trace_paths
+from .search import DeviceIndex
 from .utils import create_cablepaths, rebuild_paths
 
 COMPONENT_MODELS = (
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
Re: performance concern  
A virtual chassis is essentially a switch stack, which typically has only a handful of members. Given that, I think it’s reasonable to keep this as-is.

jeremystretch (approved):

PR #22508, opened 2026-06-22
Closes #22507: Check is_active in restrict() and IsSuperuser superuser bypass

### LINKED ISSUE #22507: `restrict()` and `IsSuperuser` superuser bypass should also check `is_active`
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

`RestrictedQuerySet.restrict()` and `IsSuperuser` short-circuit on `is_superuser` without also checking `is_active`, so a deactivated superuser account is treated as a full superuser by these two checks.

Demonstrated at the queryset level (`Token` uses a plain `RestrictedQuerySet` manager):

```python
from users.models import Token, User

# A deactivated superuser holding no explicit view permission
user = User.objects.create(username='inactive_su', is_superuser=True, is_active=False)

# Some tokens exist
Token.objects.create(user=User.objects.create(username='owner'))

# restrict() should return an empty queryset for a user with no view permission,
# but the is_superuser-only bypass returns every Token instead
Token.objects.restrict(user, 'view').count()
```

The same omission is in `IsSuperuser.has_permission()` (`netbox/utilities/api.py:47`), which returns `True` for a deactivated superuser.

### Expected Behavior

A deactivated superuser (`is_active=False`) is not granted the superuser bypass. `restrict()` returns `none()` (no explicit permission), and `IsSuperuser` returns `False`, matching the existing `is_active` guard in `ObjectPermissionMixin.has_perm` (`netbox/authentication/__init__.py:118`).

### Observed Behavior

`restrict()` (`netbox/utilities/querysets.py:52`) returns the unrestricted queryset and `IsSuperuser` returns `True` for a deactivated superuser, because both check `is_superuser` only.

jnovinger:
### Closes: #22507

`RestrictedQuerySet.restrict()` and `IsSuperuser` short-circuited on `is_superuser` without checking `is_active`, granting a deactivated superuser the full bypass. `restrict()` fails open on that branch, returning the unrestricted queryset.

- `restrict()` (`utilities/querysets.py`): guard the superuser branch with `is_active`; kept on that branch only so the `permission_is_exempt()` short-circuit stays reachable for anonymous/exempt views; parenthesized the `and`/`or`
- `IsSuperuser.has_permission()` (`utilities/api.py`): also require `is_active`
- Both now match the existing guard in `ObjectPermissionMixin.has_perm`
- Added regression test (`users.tests.test_models.RestrictedQuerySetTestCase`): active superuser still bypasses, deactivated superuser does not

### DIFF
--- a/netbox/users/tests/test_models.py
+++ b/netbox/users/tests/test_models.py
@@ -232,3 +232,29 @@ class UserConfigTestCase(TestCase):
 
         # Clear a non-existing value; should fail silently
         userconfig.clear('invalid')
+
+
+class RestrictedQuerySetTestCase(TestCase):
+    """
+    Test the is_active handling of RestrictedQuerySet.restrict()'s superuser bypass.
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        # Token uses a plain RestrictedQuerySet manager, so its objects() exercises restrict()
+        # directly without requiring any object permissions to be configured.
+        cls.token = Token.objects.create(user=create_test_user('Token Owner'))
+
+    def test_active_superuser_bypasses_restriction(self):
+        user = User.objects.create(username='active_su', is_superuser=True, is_active=True)
+        self.assertIn(self.token, Token.objects.restrict(user, 'view'))
+
+    def test_inactive_superuser_does_not_bypass_restriction(self):
+        """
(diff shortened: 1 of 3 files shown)


pheus (approved):
Thanks!

Issue #22447, opened 2026-06-14
Cooling infrastructure modeling


jeremystretch:
### NetBox version

v4.6.2

### Feature type

Data model extension

### Proposed functionality

NetBox models power distribution end-to-end — from a building-level `PowerPanel`,
through a `PowerFeed`, into a PDU's `PowerOutlet`s, and on to the `PowerPort` of a
downstream device, all stitched together with cables that can be traced. It has no
equivalent for **cooling**. As liquid-cooled and hybrid-cooled infrastructure becomes
the norm for AI/GPU deployments, this is an increasingly significant gap.

This proposal adds a cooling data model that deliberately mirrors the power model, so
the concepts, views, and workflows feel familiar to existing users. The new surface
breaks down into three parts:

### 1. Cooling attributes on existing models

Add lightweight descriptive fields so users can record cooling characteristics without
necessarily modeling the full plumbing:

- **`DeviceType` / `Device`** — a new `cooling_method` field (`air`, `liquid`,
  `hybrid`, `immersion`). `Device` inherits the value from its `DeviceType` unless
  overridden. The existing `airflow` field is retained and remains meaningful for
  air-cooled and hybrid devices (e.g. a GPU server can be `cooling_method=hybrid` with
  `airflow=front-to-rear`).
- **`Rack`** — `cooling_capable` (`air-only`, `liquid-capable`, `liquid-required`),
  a `has_rdhx` boolean (Rear Door Heat Exchanger), and a `cooling_capacity` in kW.

These alone address the most common asks: "which devices are liquid cooled?", "which
racks can host liquid-cooled equipment?", and "which racks have an RDHx?"

### 2. Facility-to-rack cooling distribution

Two new top-level models that parallel `PowerPanel` and `PowerFeed`:

- **`CoolingSource`** — facility-level cooling infrastructure (chiller, cooling tower,
  dry cooler, or facility water system), scoped to a site/location. Carries a capacity
  (kW) and design supply/return temperature targets. Like `PowerPanel`, it is **not** a
  `Device` — this is facility plant that may live outside the data hall.
- **`CoolingFeed`** — a cooling loop from a `CoolingSource` to a rack/CDU, analogous to
  `PowerFeed`. Each feed has an explicit `type` of **supply** (cold) or **return**
  (warm), plus status, fluid type, and design flow-rate/pressure. Modeling supply and
  return as distinct objects reflects physical reality (separate pipes) and lets the two
  paths be traced independently.

[...]

Issue #6987, opened 2021-08-18
Cooling Section


adamboutcher:
### NetBox version

v2.11.10

### Feature type

New functionality

### Proposed functionality

Add a section for cooling. Much like Power connections, water cooled data centres have lots of infrastructure relating to cooling. We require water cooled racks however some people have water cooled servers.

You need a loops similar in high-level function to the current power panel and you need a connection, similar in high-level function to the current power feed.

A rack/device would need an in/out connection but you could combine these into a single database "connection" as long as there was a space to write down notes to flow/return (in/out) details, typically these are a labelled valve, so it would require some freeform text.

For DCs that have water cooled servers, you have cooling into the rack as above but then you distribute this into a server (with a manifold), much like a PDU and so the functionality could be mirrored, although again a free-form text-input for flow/return valve numbering might be required (I've not used a manifold system as I'm unsure to how I'd document it exactly).

### Use case

If its a "green" water cooled DC, then it add extra functionality to maintain connections/locations related to cooling, much like the power section, at the moment this has to be manually extended (we rather simply use custom inputs to include the valve numbers into the rack).

### Database changes

Unknown

### External dependencies

None.

bellwood:
This would essentially be the exact same thing as power. 

Perhaps the way to look at this would be modify the existing power models to be "service" models and let the user flag if the panels, feeds and components are either water or power?

adamboutcher:
That could work, the major difference to power is there's physically two connections that would need documenting in some way, you always have two connections, never just one or three etc.

The key data to capture is cooling capacity (typically in killawatts) , preassure (psi) connection type, flow/return connection (valve) id and maybe operating temp range on the actual loop.

bellwood:
> That could work, the major difference to power is there's physically two connections that would need documenting in some way, you always have two connections, never just one or three etc.

This would be just be another feed no?

bellwood:
> The key data to capture is cooling capacity (typically in killawatts) , preassure (psi) connection type, flow/return connection (valve) id and maybe operating temp range on the actual loop.

This would definitely be a modification/expansion of the existing model.

jeremystretch:
This sounds like the sort of thing that would be best implemented as a plugin initially. Then, once it's seen some real-world use, we can revisit the idea of adding it to NetBox.

adamboutcher:
Thanks! I didn't know plugins were a thing. Will see if I can do it.

PR #22517, opened 2026-06-23
22447 Add Cooling infrastructure modeling

arthanson:
### Closes: #22447 

Adds cooling infrastructure modeling to DCIM, mirroring the *shape* of the power-distribution stack and reusing the standard NetBox model features (contacts and image attachments on the facility plant, plus change logging, custom fields, and tags). This lets operators model liquid- and hybrid-cooled environments (CDUs, manifolds, RDHx, facility plant) as a source of truth.

NetBox models **design intent and physical topology, not live conditions** — flow and capacity are design/rated specifications, not telemetry.

The topology parallels power, with two deliberate differences:

1. **Coolant connections are direct foreign keys, not cables.** Hoses aren't structured cabling, so there's no `Cable`/`CablePath` machinery; tracing a loop is an FK walk.
2. **A feed is the whole loop.** One `CoolingFeed` represents both the supply (cold) and return (warm) paths of a single loop, rather than modeling each direction as a separate feed.

```text
POWER:    PowerPanel    → PowerFeed    →[cable]→ PowerPort      ↔ PowerOutlet
COOLING:  CoolingSource → CoolingFeed  (serves a rack)          CoolingIntake ↔(FK) CoolingOutflow
```

## End-to-end picture

```text
 CoolingSource   (chiller / cooling tower / dry cooler / CRAC / CRAH)
     │
     ▼
  CoolingFeed     (entire loop — supply + return; serves a rack)
     │
     ▼
  Rack            (devices derive their serving feed from the rack)
     │
     ├─ Device  (e.g. CDU)
     │     CoolingIntake    ◀── chilled water in (facility side)
     │       │  through the device
     │       ▼
     │     CoolingOutflow ──┐
     │                      │  coolant
     ├─ Device  (e.g. RDHx / row manifold)
     │     CoolingIntake ◀──┘
     │       │  through the device
     │       ▼
     │     CoolingOutflow ──┐
     │                      │  coolant
     └─ Device  (e.g. server)
           CoolingIntake ◀──┘   (cold-plate inlet)
```

Coolant connectivity is a chain of direct FKs stored on the **downstream (consuming)** side:

| Field | Points to | Meaning |
|---|---|---|
| `CoolingIntake.cooling_outflow` | one upstream `CoolingOutflow` | the outflow feeding this intake |
| `CoolingOutflow.cooling_intake` | parent `CoolingIntake` (same device) | coolant enters via the intake, leaves via the outflow |

A device's serving `CoolingFeed` is **derived from its rack**, not stored on each intake.

In-rack equipment (CDUs, manifolds, RDHx) is modeled generically as (typically zero-U) `Device`s carrying these components — the same way a PDU is a device with power ports/outlets. No bespoke CDU/manifold/RDHx models were added.

## New models

- **CoolingSource** — facility plant (chiller / cooling tower / dry cooler / CRAC / CRAH), scoped to a site/location; carries the loop `fluid_type` and a rated cooling capacity. Reuses contacts and image attachments.
- **CoolingFeed** — a single coolant loop (supply + return) delivered from a source to a rack; rated cooling capacity and rated (design) flow rate.
- **CoolingIntake** — a device coolant intake (CDU facility intake, server cold-plate inlet); connector type, diameter, rated maximum flow. Fed by **one** upstream `CoolingOutflow`.
- **CoolingOutflow** — a device coolant supply point (CDU / manifold outlet); connector type, diameter, optionally linked to a parent `CoolingIntake` on the same device.
- **CoolingIntakeTemplate** / **CoolingOutflowTemplate** — component templates, auto-instantiated when a device or module is created.

## Changes to existing models

- **Device / DeviceType / ModuleType**: new `cooling_method` (air / liquid / hybrid / immersion). A `Device` inherits the value from its `DeviceType` on creation and may override it (like `airflow`); `DeviceType` and `ModuleType` carry it as a product-level attribute (`ModuleType` mirrors `airflow` — type-level only, no per-instance override). On the edit and filter forms, `cooling_method` and `airflow` are grouped under a dedicated **Cooling** section. Each also gains cooling component counter caches (`cooling_intake_count` / `cooling_outflow_count`, or the `_template_count` variants on the types).
- Rack / RackType: cooling_capability (air-only / hybrid / liquid-only) and cooling_capacity (kW) defined on the shared RackBase — set on the RackType and propagated to racks (like max_weight).
- **No `Cable` changes** — coolant connectivity is FK-based (`CoolingIntake.cooling_outflow`, `CoolingOutflow.cooling_intake`); no new cable type or cable-termination registration.

## Field conventions on the new cooling models

Several quantities are modeled as **value + unit** with a normalized field for correct cross-unit sorting (the same pattern as weight). All are **design/rated specifications**, not live readings:

- **Connector type** — industry-standard set: UQD, UQDB, QDC, camlock, NPT, BSP, proprietary (no single-vendor names).
- **Diameter** (intakes/outflows) — mm / cm / in, normalized to mm (`DiameterMixin`).
- **Flow rate** — `CoolingFeed.rated_flow_rate` and `CoolingIntake.maximum_flow` — L/min / m³/h / GPM, normalized to L/min (`MaximumFlowMixin`; the feed inlines the same value/unit/normalized pattern).
- **Fluid type** — recorded once on the `CoolingSource` (water / water-glycol / dielectric / refrigerant / other); the whole loop shares it.
- **Capacity** (`cooling_capacity`) stays single-unit kW — the universal standard for IT cooling load, so no unit selector.

> No `temperature` fields: supply/return temperatures are operating conditions that go stale immediately, so they belong in a monitoring stack rather than the source of truth.
>
> No `pressure` field: "operating pressure" has no meaning as a *design* value independent of live conditions, so it's omitted rather than inviting a stored telemetry reading. If a rated pressure spec is needed later, it can be added deliberately.

## Design rationale

- **No live metrics.** A source of truth shouldn't store point-in-time telemetry (temperature, pressure, instantaneous flow) — it goes stale immediately and belongs in a monitoring stack (Prometheus/BMS). Retained values are stable specs: rated flow, rated capacity, fluid type, connector geometry.
- **Coolant isn't cabling.** Hoses lack structured-cabling semantics (no pass-through/`CablePath`), so connections are direct FKs; individual hoses aren't inventoried.
- **One feed per loop.** A loop's supply and return are physically inseparable, so a single `CoolingFeed` represents the whole loop rather than splitting it into a supply feed and a return feed.
- **Feed derived, not denormalized.** A `CoolingIntake` doesn't store its serving `CoolingFeed`; the feed is resolved from the device's rack. This avoids a denormalized FK that could drift out of sync when a device is moved.
- **Simple FKs, no through model.** `CoolingIntake → CoolingOutflow` is many-to-one, so there's no M2M requiring a `through`; the intake row is already the per-connection record. A dedicated connection model would only be warranted for multi-source intakes (redundancy is modeled as multiple intakes, like dual-PSU) or physical-hose inventory.
- **In-rack equipment is a device.** RDHx, CDUs, and manifolds are real appliances with a make/model, so they're modeled as ordinary (typically zero-U) `Device`s rather than rack booleans — mirroring how NetBox has no `has_pdu`/`has_ups` flag. Rack-level liquid readiness is still expressed via `Rack.cooling_capability`.

## Deferred (non-schema) decisions

- **Vertical/rail-mounted manifolds and RDHx** — model as zero-U (`u_height=0`) devices, reusing the vertical-PDU pattern. A generic mounting-type field, if wanted, is a DCIM-wide enhancement, not cooling-specific.
- **Cooling-path visualization** — a graphical trace widget over the FK chain is a separable UI enhancement; the MVP is an FK-walk trace.

## Data model (fields & relationships)

New models list **all declared fields**; existing models list **only the fields added** by this feature. `FK →` marks a foreign key; `_abs_*` are internal columns holding the normalized value (L/min, mm) for consistent sorting/filtering.

**New models**

```text
CoolingSource — facility plant
  name
  site                     FK → Site
  location                 FK → Location          (optional)
  type                     chiller | cooling-tower | dry-cooler | crac | crah
  status                   offline | active | planned | failed
  fluid_type               water | water-glycol | dielectric | refrigerant | other
  cooling_capacity         rated kW
  (also: contacts, image attachments, description, comments, tags, custom fields)

CoolingFeed — coolant loop (source → rack); one feed = supply + return
  cooling_source           FK → CoolingSource
  rack                     FK → Rack              (optional)
  tenant                   FK → Tenant            (optional)
  name
  status                   offline | active | planned | failed
  cooling_capacity         rated kW
  rated_flow_rate          design  + rated_flow_rate_unit / _abs_rated_flow_rate   (L/min)

CoolingIntake — device coolant intake
  device                   FK → Device
  module                   FK → Module            (optional)
  name / label / description
  type                     connector: uqd | uqdb | qdc | camlock | npt | bsp | proprietary
  diameter                 + diameter_unit / _abs_diameter        (mm)
  maximum_flow             rated  + maximum_flow_unit / _abs_maximum_flow   (L/min)
  cooling_outflow          FK → CoolingOutflow    (optional, upstream — the feeding outflow)

CoolingOutflow — device coolant supply
  device                   FK → Device
  module                   FK → Module            (optional)
  name / label / description
  type                     connector type
  diameter                 + diameter_unit / _abs_diameter        (mm)
  cooling_intake           FK → CoolingIntake     (optional, parent; same device)

CoolingIntakeTemplate — instantiated as a CoolingIntake on device/module creation
  device_type              FK → DeviceType        (optional)
  module_type              FK → ModuleType        (optional)
  name / label / description
  type (connector), diameter (+unit, mm), maximum_flow (rated; +unit, L/min)

CoolingOutflowTemplate — instantiated as a CoolingOutflow on device/module creation
  device_type              FK → DeviceType        (optional)
  module_type              FK → ModuleType        (optional)
  name / label / description
  type (connector), diameter (+unit, mm)
  cooling_intake           FK → CoolingIntakeTemplate  (optional, parent)
```

**Existing models — added fields only**

```text
Device       cooling_method (air|liquid|hybrid|immersion; inherited from DeviceType)
             cooling_intake_count, cooling_outflow_count            (counter caches)
DeviceType   cooling_method
             cooling_intake_template_count, cooling_outflow_template_count
ModuleType   cooling_method
             cooling_intake_template_count, cooling_outflow_template_count
Rack         cooling_capability (air-only|hybrid|liquid-only)
              cooling_capacity (kW)
RackType     cooling_capability, cooling_capacity (kW)
```

**Relationships** (all foreign keys — no new many-to-many)

```text
Site           1 ──▶ N  CoolingSource
Location       1 ──▶ N  CoolingSource                                        (optional)
CoolingSource  1 ──▶ N  CoolingFeed
Rack           1 ──▶ N  CoolingFeed                                          (optional)
Tenant         1 ──▶ N  CoolingFeed                                          (optional)
CoolingOutflow 1 ──▶ N  CoolingIntake   via CoolingIntake.cooling_outflow    (an intake is fed by one outflow)
CoolingIntake  1 ──▶ N  CoolingOutflow  via CoolingOutflow.cooling_intake    (parent intake; same device)
Device         1 ──▶ N  CoolingIntake, CoolingOutflow
Module         1 ──▶ N  CoolingIntake, CoolingOutflow                        (optional)
DeviceType /
ModuleType     1 ──▶ N  CoolingIntakeTemplate, CoolingOutflowTemplate
CoolingIntakeTemplate  1 ──▶ N  CoolingOutflowTemplate                       (parent)
```

A device's serving `CoolingFeed` is not a stored FK; it is derived from the device's rack (`Rack.cooling_feeds`).

## Scope

Full surface area for every new model: choice sets, filtersets, forms (model/bulk-edit/bulk-import/filter/create), tables, REST API (serializers, viewsets, routes), GraphQL (types, filters, queries), search indexes, navigation (new **Cooling** menu), UI panels/templates, and database migrations.


arthanson:

arthanson:

arthanson:
Per Claude Review:

## Fixed

1. CoolingMethodChoices key → 'Device.cooling_method' (was 'CoolingMethod'). Now matches the airflow convention (Device.airflow/Rack.airflow), so extending via FIELD_CHOICES = {'dcim.Device.cooling_method': [...]} works as documented.
2. CoolingFeedTestCase.create_data → bumped from 2 to 3 items for proper bulk-create coverage.
3. _abs_* GraphQL exposure

## Declined (consistency with NetBox precedent)

4. No keys on CoolingSourceType/CoolingFeedType/FluidType/CoolingConnectorType. I checked: power's type/supply/connector sets (PowerFeedTypeChoices, PowerFeedSupplyChoices, PowerPortTypeChoices, PowerOutletTypeChoices) are all keyless — NetBox makes only status (and airflow) field-choices-extensible.
5. CoolingFeedForm.rack query_params — left as-is; identical to PowerFeedForm (server-side clean() enforces the cross-site rule). Adding live filtering isn't clean here (there's no site form field to bind query_params to — only cooling_source), and it'd diverge from the power form. Noted as a possible future UX nicety.

## Verified, no change

6. GraphQL filter tests - not missing: the base APIViewTestCase auto-generates and runs test_graphql_filter_objects (and fails loudly if none exist, graphql_auto_filter_required=True). The power tests rely on the same auto mechanism; explicit graphql_filter_tests (as in SiteTestCase) are only for special cases.

arthanson:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Also see comments on previous review of Declined requested changes and reasoning.

arthanson:
Both Claude review issues addressed.

netbox/dcim/models/cooling.py
```diff
@@ -0,0 +1,205 @@
...
+        choices=CoolingFeedStatusChoices,
+        default=CoolingFeedStatusChoices.STATUS_ACTIVE
+    )
+    type = models.CharField(
+        verbose_name=_('type'),
+        max_length=50,
+        choices=CoolingFeedTypeChoices,
+        default=CoolingFeedTypeChoices.TYPE_SUPPLY
+    )
+    fluid_type = models.CharField(
```
jeremystretch: Seems like this would be set on the parent CoolingSource. Have we identified use case for defining it separately for each feed?

netbox/netbox/models/mixins.py
```diff
@@ -75,6 +88,300 @@ def clean(self):
             raise ValidationError(_("Must specify a unit when setting a weight"))
 
 
+class DiameterMixin(models.Model):
```
jeremystretch: These mixin classes should probably live in `dcim.models.mixins` unless there's a reasonable expectation that they'll be shared among apps (which seems unlikely IMO).

netbox/netbox/models/mixins.py
```diff
@@ -75,6 +88,300 @@ def clean(self):
...
+        # Validate pressure and pressure_unit
+        if self.pressure is not None and not self.pressure_unit:
+            raise ValidationError(_("Must specify a unit when setting a pressure"))
+
+
+class CoolingTemperatureMixin(models.Model):
+    """
+    Adds supply/return temperatures sharing a single unit, with normalized (°C) fields for ordering.
+    """
+    supply_temperature = models.DecimalField(
```
jeremystretch: Are these intended to store thresholds or live data? If the former, we should be clear about the intent; if the later, we probably don't want to record that data statically as it's purely environmental. (Same concern re: PressureMixin and FlowRateMixin)

netbox/netbox/models/mixins.py
```diff
@@ -75,6 +88,300 @@ def clean(self):
...
+
+    def clean(self):
+        super().clean()
+
+        # Validate maximum_flow and maximum_flow_unit
+        if self.maximum_flow is not None and not self.maximum_flow_unit:
+            raise ValidationError(_("Must specify a unit when setting a maximum flow"))
+
+
+class PressureMixin(models.Model):
```
jeremystretch: This is inherited only by CoolingFeed; does it need to be a mixin?

netbox/dcim/models/device_components.py
```diff
@@ -683,6 +685,111 @@ def get_status_color(self):
         return PowerOutletStatusChoices.colors.get(self.status)
 
 
+#
+# Cooling components
+#
+
+class CoolingPort(
+    DiameterMixin, MaximumFlowMixin, ModularComponentModel, CabledObjectModel, PathEndpoint, TrackingModelMixin
```
jeremystretch: Let's not try to shoehorn the Cable model into representing the dispersal of liquid coolant, as the two concepts have very different constraints. (I realize this was prescribed in the original FR; my fault for not catching it in the source document.)

Arguably we may not even want to model the plumbing between a port and an outlet. I imagine these connections (which presumably can be one-to-many) can grow quite complex. It might make more sense to just map the indirect dependencies via a ForeignKey from CoolingPort to CoolingOutlet.

netbox/dcim/models/device_components.py
```diff
@@ -683,6 +685,111 @@ def get_status_color(self):
...
+    """
+    type = models.CharField(
+        verbose_name=_('type'),
+        max_length=50,
+        choices=CoolingFeedTypeChoices,
+        blank=True,
+        null=True,
+        help_text=_('Supply (cold) or return (warm)')
+    )
+    connector_type = models.CharField(
```
jeremystretch: This seems misleading IMO. I would expect "type" to refer to the physical termination type (which we seem to have captured as diameter), not the direction of flow. I suggest conveying the direction (inflow/exhaust or supply/return) explicitly, and potentially use "type" to indicate the diameter and other physical attributes of each connector.

jeremystretch (changes requested):
This is a good start, but let's rethink the data model some. Specifically consider:

1. Do we need to model individual hoses? If so, how do we handle multi-point connectors (e.g. a hose with a 1-to-n splitter)?
2. Do we need to explicitly model the return path in addition to the supply path?
3. What are the canonical types of physical inlet & outlet connectors that we need to capture? Is diameter alone sufficient to convey these?
4. What temperature information belongs in the data model vs. observability metrics?

The description in FR #22447 was lifted from an initial feature draft; it probably needs a lot more consideration regarding real-world modeling and constraints. It might help to investigate various real-world cooling distribution implementations and determine how we would optimally model those specific instances in NetBox.

adamboutcher:
Historical related issue https://github.com/netbox-community/netbox/issues/6987 useful for searching etc.

netbox/dcim/models/device_components.py
```diff
@@ -29,6 +29,8 @@
     'CabledObjectModel',
     'ConsolePort',
     'ConsoleServerPort',
+    'CoolingOutlet',
+    'CoolingPort',
```
jeremystretch: These sound a bit strange IMO. I realize they were replicated from PowerPort & PowerOutlet, but those terms are far more common in the context of power distribution. I suggest renaming these to something more intuitive; maybe CoolingIngress & CoolingEgress? Or Intake and Outflow?

netbox/dcim/models/cooling.py
jeremystretch: Sorry, want to be clear on this: What is the use case for defining this per feed?

netbox/dcim/models/device_components.py
```diff
@@ -683,6 +685,135 @@ def get_status_color(self):
...
+#
+# Cooling components
+#
+
+class CoolingPort(DiameterMixin, MaximumFlowMixin, ModularComponentModel, TrackingModelMixin):
+    """
+    A coolant intake/outlet port within a Device (e.g. a server cold-plate inlet or CDU intake). A
+    CoolingPort is supplied by an upstream CoolingOutlet or CoolingFeed.
+    """
+    flow_direction = models.CharField(
```
jeremystretch: If we tie the direction of flow to the type of component, we can omit this field entirely. I.e. we can assert that coolant always flows from a CoolingOutflow to a CoolingIntake (see my other comment re: component naming).

netbox/dcim/models/mixins.py
```diff
@@ -150,3 +161,250 @@ def clean(self):
...
+
+    def clean(self):
+        super().clean()
+
+        # Validate maximum_flow and maximum_flow_unit
+        if self.maximum_flow is not None and not self.maximum_flow_unit:
+            raise ValidationError(_("Must specify a unit when setting a maximum flow"))
+
+
+class CoolingTemperatureMixin(models.Model):
```
jeremystretch: This seems like it would be useful only for tracking live environmental data, which I don't think we want to include in the scope of FR #22447. Suggest we omit temperature tracking entirely, at least for the initial implementation.

netbox/dcim/models/mixins.py
```diff
@@ -150,3 +161,250 @@ def clean(self):
...
+
+    def clean(self):
+        super().clean()
+
+        # Validate rated_flow_rate and rated_flow_rate_unit
+        if self.rated_flow_rate is not None and not self.rated_flow_rate_unit:
+            raise ValidationError(_("Must specify a unit when setting a rated flow rate"))
+
+
+class MaximumFlowMixin(models.Model):
```
jeremystretch: Is this significantly different from RatedFlowRateMixin? Could CoolingPort and CoolingFeed use the same mixin class?

netbox/dcim/models/mixins.py
```diff
@@ -150,3 +161,250 @@ def clean(self):
         # RF role may be set only for wireless interfaces
         if self.rf_role and self.type not in WIRELESS_IFACE_TYPES:
             raise ValidationError({'rf_role': _("Wireless role may be set only on wireless interfaces.")})
+
+
+class DiameterMixin(models.Model):
```
jeremystretch, line 288: I suspect users will typically deal with just a handful of connector diameters, and possibly other form factors. Maybe we should convey this as a fixed `type` choice field, similar to power ports. We can research and populate all the well-known connector types as a convenience for the user.

netbox/dcim/models/modules.py
```diff
@@ -91,6 +91,13 @@ class ModuleType(ImageAttachmentsMixin, PrimaryModel, WeightMixin):
         blank=True,
         null=True
     )
+    cooling_method = models.CharField(
```
jeremystretch, line 149: We might want to validate this against the parent device. For instance, a module can't have a cooling method of "immersion" if its parent device is air or hybrid.

docs/models/dcim/rack.md
```diff
@@ -51,5 +51,17 @@ The unique physical serial number assigned to this rack.
 
 A unique, locally-administered label used to identify hardware resources.
 
+### Cooling Capability
```
jeremystretch, line 57: I'm not sure I understand the meaning of these choices. Let's provide a brief description of each.

netbox/dcim/models/racks.py
```diff
@@ -359,6 +378,11 @@ class Rack(ContactsMixin, ImageAttachmentsMixin, TrackingModelMixin, RackBase):
         blank=True,
         null=True
     )
+    has_rdhx = models.BooleanField(
```
jeremystretch: Is indicating this as a boolean sufficient? What if the user wants to track the make & model of the installed RDHX? It would provide more flexibility to direct the user to model the RDHX the same as any other device associated with the rack. (It might still be worth keeping the `has_rdhx` boolean, but it's worth further consideration.)

If we keep `has_rdhex`, we should validate it against the rack type. Two-post racks, for instance, presumably can't have an RDHX installed.

netbox/dcim/models/device_components.py
```diff
@@ -683,6 +685,135 @@ def get_status_color(self):
...
+        verbose_name=_('type'),
+        max_length=50,
+        choices=CoolingConnectorTypeChoices,
+        blank=True,
+        null=True,
+        help_text=_('Physical connector type')
+    )
+    # diameter, diameter_unit, _abs_diameter provided by DiameterMixin
+    # maximum_flow, maximum_flow_unit, _abs_maximum_flow provided by MaximumFlowMixin
+    heat_capacity = models.DecimalField(
```
jeremystretch: I'm not sure this makes sense to define at the component level. The aggregate feed capacity is shared by all connected intakes, so it will vary in proportion to the demand induced by peer loads.

If we do keep this, let's standardize on the name `cooling_capacity` (used for CoolingSource).

netbox/dcim/models/device_components.py
```diff
@@ -683,6 +685,135 @@ def get_status_color(self):
...
+        help_text=_('The upstream cooling outlet supplying this port')
+    )
+    cooling_feed = models.ForeignKey(
+        to='dcim.CoolingFeed',
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+        related_name='cooling_ports',
+        help_text=_('The upstream cooling feed supplying this port')
+    )
```
jeremystretch: Is there a need to track the cooling feed relationship explicitly? If so, can we change this to a denormalized relationship?

netbox/dcim/models/device_component_templates.py
```diff
@@ -432,6 +436,160 @@ def to_yaml(self):
...
+    )
+    type = models.CharField(
+        verbose_name=_('type'),
+        max_length=50,
+        choices=CoolingConnectorTypeChoices,
+        blank=True,
+        null=True
+    )
+    # diameter, diameter_unit, _abs_diameter provided by DiameterMixin
+    color = ColorField(
```
jeremystretch: Not sure it makes sense to assign a color to cooling outlets. We support this on power outlets because color is often used to signify redundant feeds; this isn't commonplace for cooling AFAIK. I could be wrong.

netbox/dcim/models/cooling.py
```diff
@@ -0,0 +1,206 @@
...
+        max_length=100,
+        db_collation="natural_sort"
+    )
+    status = models.CharField(
+        verbose_name=_('status'),
+        max_length=50,
+        choices=CoolingFeedStatusChoices,
+        default=CoolingFeedStatusChoices.STATUS_ACTIVE
+    )
+    flow_direction = models.CharField(
```
jeremystretch: Does it make sense to specify the flow direction on a cooling feed? My assumption was that one feed would handle both the distribution and return paths, but maybe not. If we want to use one "feed" for each direction, we should reconsider the name.

jeremystretch (changes requested):

netbox/dcim/models/device_components.py
adamboutcher: In/Out or even better, Flow/Return would be my vote(s).

The Port/Outlet isn't really "normal" for cooling but anyone familiar with Netbox and Power connections would likely understand them.

Ingress/Egress may not be inherently obvious on direction as does it apply to the pipework or device, plus plumbers don't usually use these terms.

docs/models/dcim/rack.md
adamboutcher: Possibly a way to stop on chip water cooled kit being installed in to an air cooled rack? Or air cooled things into immersion?

mrmrcoleman:
Overall a solid keeping the power-model shape. Most deviations from FR #22447 are reasonable simplifications (several requested in review). Notes below, kept short.

## Decisions

- **Design supply/return temperatures — defer.** Not telemetry (they're design setpoints, like rated capacity), but a clean two-way door: adding nullable °C fields later is a purely additive migration, no backfill. Fine to leave out of v1 and add as a fast-follow.
- **RDHx — defer, with notes below.** Keep the current "model it as a zero-U device" direction; don't add a `Rack.has_rdhx` boolean now (it'd create a second source of truth that drifts from the device).

## RDHx notes (for whoever picks it up)

- An RDHx is **1:1 with a rack** — a water-cooled coil that replaces the rear door of one cabinet. It isn't shared or free-standing, and it doesn't occupy a U-slot, so a zero-U device is a slightly forced fit (no rear-door mount concept exists).
- **"Which racks have an RDHx?" is not free with the device approach.** Verified: `RackFilterSet` has no `devices__` traversal filter, and there's no built-in RDHx device role/type. Today you can only filter the *Device* list by role — that returns devices, not a rack list, and can't do the inverse ("racks missing one").
- If we later add RDHx-as-device, plan to also add a small **`RackFilterSet` filter** traversing `devices__role` (needs `distinct=True`). Otherwise the FR's "RDHx tracking" reporting need is quietly dropped.
- Also unaddressed (defer, but flag): partial-U-coverage RDHx and active-vs-passive (raised by @adamboutcher on the FR).

## Endorsed simplifications (no change requested)

- **FK instead of `cooling-hose` cables** — reasonable; handles 1-to-n splitters and dual-feed redundancy (multiple intakes), reviewer-endorsed.
- **One feed = whole loop (supply + return combined)** — fine simplification; return is modeled implicitly.

## Documentation suggestions

- Add a short mapping from the new names/relationships back to power so existing users' mental model transfers: `CoolingIntake`≈`PowerPort`, `CoolingOutflow`≈`PowerOutlet`, FK chain ≈ cable.
- Explicitly document the deliberate v1 limitations so they read as conscious cuts, not gaps: return path implicit, path tracing / impact-analysis deferred, no per-hose inventory, no design pressure/temperature yet.

## Deferred but acceptable (noted, not blocking)

- Design/rated pressure on the feed — likely fast-follow depending on feedback.
- Independent supply/return modeling (asymmetric routing, separate return failure).
- Path tracing + impact analysis ("if this loop fails, what loses cooling?") — an FR headline use case, deferred pending model approval; make sure it lands soon after.

docs/features/cooling.md
```diff
@@ -0,0 +1,54 @@
+# Cooling
+
+As part of its DCIM feature set, NetBox supports modeling data center cooling infrastructure, from facility plant down to the coolant connections on individual devices. This is used to document liquid- and hybrid-cooled environments (chillers, cooling distribution units, manifolds, rear-door heat exchangers, and cold-plate servers) as a source of truth.
+
+## Relationship to Power
+
+The cooling stack deliberately mirrors the [power distribution](./power-tracking.md) model:
```
jeremystretch: I'm not sure this is accurate any longer, given that connections among power ports & outlets are represented using cables. I would avoid referencing the power data model, and just present cooling as its own independent entity.

docs/features/cooling.md
```diff
@@ -0,0 +1,54 @@
...
+| Cooling Outflow | Power Outlet | A device-side supply (feeds downstream equipment) |
+
+There are two intentional differences from the power model:
+
+- **Coolant connections are direct references, not cables.** Hoses are not structured cabling, so an intake references the outflow that supplies it directly, rather than through a cable. Tracing a loop is a walk along these references.
+- **A feed represents the entire loop.** A single cooling feed covers both the supply (cold) and return (warm) paths of a loop, rather than modeling each direction separately.
+
+## Cooling Sources
+
+A [cooling source](../models/dcim/coolingsource.md) is the furthest upstream cooling element modeled in NetBox, representing facility plant such as a chiller, cooling tower, dry cooler, or CRAC/CRAH unit. Each source is associated with a site, and may optionally be associated with a particular location within that site. A cooling source is not a device; it represents external facility plant, and records the coolant (fluid type) and total rated cooling capacity for the loops it originates.
```
jeremystretch:
```suggestion
A [cooling source](../models/dcim/coolingsource.md) is the furthest upstream cooling element modeled in NetBox, representing a chiller, cooling tower, dry cooler, or CRAC/CRAH unit. Each source is associated with a site, and may optionally be associated with a particular location within that site. A cooling source is not a device; it represents external facility plant, and records the coolant (fluid type) and total rated cooling capacity for the loops it originates.
```

docs/features/cooling.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Cooling Feeds
+
+A [cooling feed](../models/dcim/coolingfeed.md) represents a coolant loop originating from a cooling source and serving a particular rack. Each feed records an operational status, a rated cooling capacity, and a rated (design) flow rate.
```
jeremystretch:
```suggestion
A [cooling feed](../models/dcim/coolingfeed.md) represents a coolant loop running between a cooling source and a particular rack. Each feed records an operational status, a rated cooling capacity, and a rated (design) flow rate.
```

docs/features/cooling.md
```diff
@@ -0,0 +1,54 @@
...
+
+The [cooling intakes](../models/dcim/coolingintake.md) a feed supplies are not listed on the feed explicitly; they are derived from the devices installed in the rack the feed serves. Moving a device to a different rack therefore reassigns it to that rack's feed automatically.
+
```
jeremystretch: Seems redundant to the section below.
```suggestion
```

docs/features/cooling.md
```diff
@@ -0,0 +1,54 @@
...
+- A [rack](../models/dcim/rack.md) records a **cooling capability** (air-only, liquid-capable, or liquid-required) and a **cooling capacity** in kilowatts, typically inherited from its rack type.
+- A [device](../models/dcim/device.md) records a **cooling method** (air, liquid, hybrid, or immersion), inherited from its device type and overridable per device.
+
+!!! note "Scope of the initial version"
+    The cooling model captures topology and design specifications. A few capabilities are intentionally left out at this point:
+
+    - **The return path is implicit.** A single feed models the whole loop; the supply and return legs are not traced as separate objects.
+    - **Path tracing and impact analysis are not yet surfaced.** The reference chain supports tracing a loop, but there is no graphical trace or "what does this source feed?" view yet.
+    - **Individual hoses are not inventoried.** Connections are modeled as direct references, not as first-class cable objects.
+    - **No design pressure or temperature.** These are operating conditions with no stable design value independent of live readings, so they are recorded in a monitoring system rather than in NetBox.
```
jeremystretch: The scope of the documentation is limited to what the application currently supports.
```suggestion
```

docs/models/dcim/coolingfeed.md
```diff
@@ -0,0 +1,39 @@
+# Cooling Feed
+
+A cooling feed represents a coolant loop delivered from a [cooling source](./coolingsource.md) to a particular rack or coolant distribution unit (CDU). It is the cooling equivalent of a [power feed](./powerfeed.md). The [cooling intakes](./coolingintake.md) a feed supplies are derived from the devices installed in the rack it serves, rather than referenced explicitly.
```
jeremystretch:
```suggestion
A cooling feed represents a coolant loop delivered from a [cooling source](./coolingsource.md) to a particular rack or coolant distribution unit (CDU). The [cooling intakes](./coolingintake.md) a feed supplies are derived from the devices installed in the rack it serves, rather than referenced explicitly.
```

docs/models/dcim/coolingfeed.md
```diff
@@ -0,0 +1,39 @@
...
+
+A single feed represents the entire loop, covering both the supply (cold) and return (warm) paths.
+
+The rated flow rate recorded on a feed is a design specification (the intended operating envelope), not live telemetry; runtime readings belong in an external monitoring system.
+
```
jeremystretch: Redundant to the field description.
```suggestion
```

docs/models/dcim/coolingfeed.md
```diff
@@ -0,0 +1,39 @@
...
+!!! tip
+    Additional statuses may be defined by setting `CoolingFeed.status` under the [`FIELD_CHOICES`](../../configuration/data-validation.md#field_choices) configuration parameter.
+
+### Cooling Capacity
+
+The heat-removal capacity of the feed, in kilowatts (kW).
+
+### Rated Flow Rate
+
+The rated (design) coolant flow rate, expressed as a numeric value with a selectable unit (L/min, m³/h, or GPM).
```
jeremystretch:
```suggestion
The rate of coolant flow supported by the feed, expressed as a numeric value with a selectable unit (L/min, m³/h, or GPM).
```

docs/models/dcim/coolingintake.md
```diff
@@ -0,0 +1,36 @@
+# Cooling Intakes
+
+A cooling intake is a device component which represents a coolant intake on a device, such as a server cold-plate inlet or a coolant distribution unit (CDU) intake. A cooling intake optionally references the upstream [cooling outflow](./coolingoutflow.md) which supplies it. The [cooling feed](./coolingfeed.md) an intake belongs to is derived from the rack in which its device is installed, rather than referenced directly.
```
jeremystretch:
```suggestion
A cooling intake is a device component which consumes coolant, such as a server cold-plate inlet or a coolant distribution unit (CDU) intake. A cooling intake optionally references the upstream [cooling outflow](./coolingoutflow.md) which supplies it. The [cooling feed](./coolingfeed.md) an intake belongs to is derived from the rack in which its device is installed, rather than referenced directly.
```

docs/models/dcim/coolingoutflow.md
```diff
@@ -0,0 +1,44 @@
+# Cooling Outflows
+
+A cooling outflow is a device component which delivers coolant to a downstream [cooling intake](./coolingintake.md), and generally represents an outlet on a coolant distribution unit (CDU) or manifold. A cooling outflow may optionally be associated with an upstream cooling intake on the same device for path tracing.
```
jeremystretch, line 3: It's not clear to me whether a CoolingOutflow is intended to represent the return of warmed coolant from a device back to the upstream cooling source.

docs/models/dcim/coolingsource.md
```diff
@@ -0,0 +1,36 @@
+# Cooling Source
+
+A cooling source represents a facility-level source of cooling, such as a chiller, cooling tower, or dry cooler. It is the cooling equivalent of a [power panel](./powerpanel.md): it serves as the upstream origin for one or more [cooling feeds](./coolingfeed.md) which distribute coolant to racks and devices. A cooling source is not modeled as a device; it represents external facility plant.
```
jeremystretch:
```suggestion
A cooling source represents a facility-level source of cooling, such as a chiller, cooling tower, or dry cooler. It serves as the upstream origin for one or more [cooling feeds](./coolingfeed.md) which distribute coolant to racks and devices. A cooling source is not modeled as a device; it represents external facility plant.
```

docs/models/dcim/coolingsource.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Fields
+
+### Site
+
+The [site](../../models/dcim/site.md) at which the cooling source is located.
```
jeremystretch:
```suggestion
The [site](./site.md) at which the cooling source is located.
```

docs/models/dcim/rack.md
```diff
@@ -51,5 +51,19 @@ The unique physical serial number assigned to this rack.
...
+
+Describes how the rack is able to cool the equipment installed in it, which indicates what kind of equipment it can accommodate:
+
+- **Air-only**: The rack is cooled by airflow only; no coolant is delivered to it. Only air-cooled equipment can be installed.
+- **Liquid-capable**: Coolant can be delivered to the rack (e.g. via a [cooling feed](./coolingfeed.md)), but it can also house air-cooled equipment. Suitable for mixed or hybrid deployments.
+- **Liquid-required**: The rack is intended exclusively for liquid-cooled equipment (such as direct-to-chip or immersion systems) and does not provide adequate air cooling on its own.
```
jeremystretch: Can we rename "liquid-required" to "liquid-only," and "liquid-capable" to "hybrid"?

netbox/dcim/forms/object_import.py
```diff
@@ -80,6 +82,46 @@ def clean_module_type(self):
         return module_type
 
 
+class CoolingIntakeTemplateImportForm(forms.ModelForm):
+
+    class Meta:
+        model = CoolingIntakeTemplate
+        fields = [
+            'device_type', 'module_type', 'name', 'label', 'type', 'diameter', 'maximum_flow',
```
jeremystretch: Missing `diameter_unit`

netbox/dcim/forms/object_import.py
```diff
@@ -80,6 +82,46 @@ def clean_module_type(self):
...
+        label=_('Cooling intake'),
+        queryset=CoolingIntakeTemplate.objects.all(),
+        to_field_name='name',
+        required=False
+    )
+
+    class Meta:
+        model = CoolingOutflowTemplate
+        fields = [
+            'device_type', 'module_type', 'name', 'label', 'type', 'diameter',
```
jeremystretch: Missing `diameter_unit`

netbox/utilities/conversion.py
```diff
@@ -66,3 +76,101 @@ def to_meters(length, unit) -> Decimal:
...
+        return round(flow_rate * Decimal('3.785411784'), 4)
+    raise ValueError(
+        _("Unknown unit {unit}. Must be one of the following: {valid_units}").format(
+            unit=unit,
+            valid_units=', '.join(FlowRateUnitChoices.values())
+        )
+    )
+
+
+def to_kilopascals(pressure, unit) -> Decimal:
```
jeremystretch: No longer used; should we remove it?

netbox/utilities/conversion.py
```diff
@@ -66,3 +76,101 @@ def to_meters(length, unit) -> Decimal:
...
+        return round(pressure * Decimal('6.894757'), 4)
+    raise ValueError(
+        _("Unknown unit {unit}. Must be one of the following: {valid_units}").format(
+            unit=unit,
+            valid_units=', '.join(PressureUnitChoices.values())
+        )
+    )
+
+
+def to_celsius(temperature, unit) -> Decimal:
```
jeremystretch: No longer used; should we remove it?

netbox/utilities/conversion.py
```diff
@@ -66,3 +76,101 @@ def to_meters(length, unit) -> Decimal:
...
+        return round((Decimal(temperature) - 32) * 5 / 9, 4)
+    raise ValueError(
+        _("Unknown unit {unit}. Must be one of the following: {valid_units}").format(
+            unit=unit,
+            valid_units=', '.join(TemperatureUnitChoices.values())
+        )
+    )
+
+
+def to_millimeters(diameter, unit) -> Decimal:
```
jeremystretch, line 74: Suggest moving this under `to_meters()` just for the sake of organization.

netbox/netbox/choices.py
```diff
@@ -197,3 +201,65 @@ class WeightUnitChoices(ChoiceSet):
...
+    UNIT_INCH = 'in'
+
+    CHOICES = (
+        (UNIT_MILLIMETER, _('Millimeters')),
+        (UNIT_CENTIMETER, _('Centimeters')),
+        (UNIT_INCH, _('Inches')),
+    )
+
+
+class TemperatureUnitChoices(ChoiceSet):
```
jeremystretch: Let's remove this if we remove `to_celsius()`.

netbox/netbox/choices.py
```diff
@@ -197,3 +201,65 @@ class WeightUnitChoices(ChoiceSet):
...
+    UNIT_GALLONS_PER_MINUTE = 'gpm'
+
+    CHOICES = (
+        (UNIT_LITERS_PER_MINUTE, _('Liters per minute (L/min)')),
+        (UNIT_CUBIC_METERS_PER_HOUR, _('Cubic meters per hour (m³/h)')),
+        (UNIT_GALLONS_PER_MINUTE, _('Gallons per minute (GPM)')),
+    )
+
+
+class PressureUnitChoices(ChoiceSet):
```
jeremystretch: Let's remove this if we remove `to_kilopascals()`.

netbox/netbox/choices.py
```diff
@@ -197,3 +201,65 @@ class WeightUnitChoices(ChoiceSet):
...
+        (UNIT_CELSIUS, _('Celsius')),
+        (UNIT_FAHRENHEIT, _('Fahrenheit')),
+    )
+
+
+class FlowRateUnitChoices(ChoiceSet):
+
+    # Metric
+    UNIT_LITERS_PER_MINUTE = 'lpm'
+    UNIT_CUBIC_METERS_PER_HOUR = 'm3h'
```
jeremystretch: We should arguably include the "p" for consistency with `lpm` and `gpm`.
```suggestion
    UNIT_CUBIC_METERS_PER_HOUR = 'm3ph'
```

docs/models/dcim/coolingoutflow.md
```diff
@@ -0,0 +1,44 @@
...
+The name of the cooling outflow. Must be unique to the parent device.
+
+### Label
+
+An alternative physical label identifying the cooling outflow.
+
+### Type
+
+Indicates whether the outlet carries supply (cold) or return (warm) coolant.
+
```
jeremystretch: I think this is a remnant from an earlier implementation.
```suggestion
```

netbox/dcim/api/serializers_/device_components.py
```diff
@@ -199,6 +204,84 @@ class Meta:
...
+        required=False,
+        allow_null=True
+    )
+    maximum_flow_unit = ChoiceField(
+        choices=FlowRateUnitChoices,
+        allow_blank=True,
+        required=False,
+        allow_null=True
+    )
+    cooling_outflow = serializers.PrimaryKeyRelatedField(
```
jeremystretch: This needs to be some form of a nested `CoolingOutflowSerializer`. (You might need to create a standalone `NestedCoolingOutflowSerializer` class to resolve the circular dependency.)

netbox/dcim/models/device_components.py
```diff
@@ -684,6 +686,83 @@ def get_status_color(self):
         return PowerOutletStatusChoices.colors.get(self.status)
 
 
+#
+# Cooling components
+#
+
+class CoolingIntake(DiameterMixin, MaximumFlowMixin, ModularComponentModel, TrackingModelMixin):
+    """
+    A coolant intake/outlet port within a Device (e.g. a server cold-plate inlet or CDU intake). A
```
jeremystretch:
```suggestion
    A coolant intake port within a Device (e.g. a server cold-plate inlet or CDU intake). A
```

jeremystretch (changes requested):
I'm unclear on the intended use of CoolingIntake. Is it meant to model the _return_ path of warm coolant from a device back to its originating cooling feed/source, or is it to be used solely for the distribution of cold coolant (and the return path is implied to mirror the cold path)?

docs/features/cooling.md
adamboutcher: So not in my data centre but there is certainly scope where the cold (in) and warm (out) may not mirror.

I will add that the in and out feed in my case are not numbered the same so referencing to plumbers I couldn't say connection 22 is a cold and warm.

arthanson:
> I'm unclear on the intended use of CoolingIntake. Is it meant to model the _return_ path of warm coolant from a device back to its originating cooling feed/source, or is it to be used solely for the distribution of cold coolant (and the return path is implied to mirror the cold path)?

Coolant enters a device via its `CoolingIntake`, leaves via its `CoolingOutflow`, and flows on to the next device's `CoolingIntake`. The warm return isn't modeled per-component; it's implied by the CoolingFeed, which represents the entire loop (both supply and return paths).

netbox/dcim/forms/model_forms.py
```diff
@@ -1812,6 +1944,59 @@ class Meta:
...
 
 
+class CoolingIntakeForm(ModularDeviceComponentForm):
+    cooling_outflow = DynamicModelChoiceField(
+        label=_('Cooling outflow'),
+        queryset=CoolingOutflow.objects.all(),
+        required=False,
+        query_params={
+            'device_id': '$device',
+        }
```
jeremystretch: This limits the connected outflow to the same device as the intake. Shouldn't they typically be on different devices?

netbox/dcim/models/device_component_templates.py
```diff
@@ -438,6 +442,125 @@ def to_yaml(self):
...
+    # diameter, diameter_unit, _abs_diameter provided by DiameterMixin
+    # maximum_flow, maximum_flow_unit, _abs_maximum_flow provided by MaximumFlowMixin
+
+    component_model = CoolingIntake
+
+    class Meta(ModularComponentTemplateModel.Meta):
+        verbose_name = _('cooling intake template')
+        verbose_name_plural = _('cooling intake templates')
+
+    def instantiate(self, **kwargs):
```
jeremystretch, line 465: Need to add `_abs_diameter` and `_abs_maximum_flow`.

netbox/dcim/models/device_component_templates.py
```diff
@@ -438,6 +442,125 @@ def to_yaml(self):
...
+                    )
+                )
+            if self.module_type and self.cooling_intake.module_type != self.module_type:
+                raise ValidationError(
+                    _("Parent cooling intake ({cooling_intake}) must belong to the same module type").format(
+                        cooling_intake=self.cooling_intake
+                    )
+                )
+
+    def instantiate(self, **kwargs):
```
jeremystretch, line 539: Need to add `_abs_diameter`.

netbox/dcim/models/device_components.py
```diff
@@ -684,6 +686,83 @@ def get_status_color(self):
...
+        verbose_name=_('type'),
+        max_length=50,
+        choices=CoolingConnectorTypeChoices,
+        blank=True,
+        null=True,
+        help_text=_('Physical connector type')
+    )
+    # diameter, diameter_unit, _abs_diameter provided by DiameterMixin
+    # maximum_flow, maximum_flow_unit, _abs_maximum_flow provided by MaximumFlowMixin
+    cooling_outflow = models.ForeignKey(
```
jeremystretch, line 715: We currently don't have any form of loop protection in place. It's probably worth putting something lightweight in place, maybe similar to what we do for nested modules.

netbox/dcim/api/urls.py
```diff
@@ -30,6 +30,8 @@
 router.register('console-server-port-templates', views.ConsoleServerPortTemplateViewSet)
 router.register('power-port-templates', views.PowerPortTemplateViewSet)
 router.register('power-outlet-templates', views.PowerOutletTemplateViewSet)
+router.register('cooling-port-templates', views.CoolingIntakeTemplateViewSet)
+router.register('cooling-outlet-templates', views.CoolingOutflowTemplateViewSet)
```
jeremystretch:
```suggestion
router.register('cooling-intake-templates', views.CoolingIntakeTemplateViewSet)
router.register('cooling-outflow-templates', views.CoolingOutflowTemplateViewSet)
```

netbox/dcim/api/urls.py
```diff
@@ -49,6 +51,8 @@
 router.register('console-server-ports', views.ConsoleServerPortViewSet)
 router.register('power-ports', views.PowerPortViewSet)
 router.register('power-outlets', views.PowerOutletViewSet)
+router.register('cooling-ports', views.CoolingIntakeViewSet)
+router.register('cooling-outlets', views.CoolingOutflowViewSet)
```
jeremystretch:
```suggestion
router.register('cooling-intakes', views.CoolingIntakeViewSet)
router.register('cooling-outflows', views.CoolingOutflowViewSet)
```

netbox/dcim/urls.py
```diff
@@ -62,6 +62,12 @@
     path('power-outlet-templates/', include(get_model_urls('dcim', 'poweroutlettemplate', detail=False))),
     path('power-outlet-templates/<int:pk>/', include(get_model_urls('dcim', 'poweroutlettemplate'))),
 
+    path('cooling-port-templates/', include(get_model_urls('dcim', 'coolingintaketemplate', detail=False))),
+    path('cooling-port-templates/<int:pk>/', include(get_model_urls('dcim', 'coolingintaketemplate'))),
+
+    path('cooling-outlet-templates/', include(get_model_urls('dcim', 'coolingoutflowtemplate', detail=False))),
+    path('cooling-outlet-templates/<int:pk>/', include(get_model_urls('dcim', 'coolingoutflowtemplate'))),
```
jeremystretch:
```suggestion
    path('cooling-intake-templates/', include(get_model_urls('dcim', 'coolingintaketemplate', detail=False))),
    path('cooling-intake-templates/<int:pk>/', include(get_model_urls('dcim', 'coolingintaketemplate'))),

    path('cooling-outflow-templates/', include(get_model_urls('dcim', 'coolingoutflowtemplate', detail=False))),
    path('cooling-outflow-templates/<int:pk>/', include(get_model_urls('dcim', 'coolingoutflowtemplate'))),
```

netbox/dcim/urls.py
```diff
@@ -123,6 +129,22 @@
...
+        name='device_bulk_add_coolingintake'
+    ),
+
+    path('cooling-outlets/', include(get_model_urls('dcim', 'coolingoutflow', detail=False))),
+    path('cooling-outlets/<int:pk>/', include(get_model_urls('dcim', 'coolingoutflow'))),
+    path(
+        'devices/cooling-outlets/add/',
+        views.DeviceBulkAddCoolingOutflowView.as_view(),
+        name='device_bulk_add_coolingoutflow'
+    ),
```
jeremystretch:
```suggestion
    path('cooling-intakes/', include(get_model_urls('dcim', 'coolingintake', detail=False))),
    path('cooling-intakes/<int:pk>/', include(get_model_urls('dcim', 'coolingintake'))),
    path(
        'devices/cooling-intakes/add/',
        views.DeviceBulkAddCoolingIntakeView.as_view(),
        name='device_bulk_add_coolingintake'
    ),

    path('cooling-outflows/', include(get_model_urls('dcim', 'coolingoutflow', detail=False))),
    path('cooling-outflows/<int:pk>/', include(get_model_urls('dcim', 'coolingoutflow'))),
    path(
        'devices/cooling-outflows/add/',
        views.DeviceBulkAddCoolingOutflowView.as_view(),
        name='device_bulk_add_coolingoutflow'
    ),
```

netbox/templates/dcim/device/base.html
```diff
@@ -33,6 +33,12 @@
                 {% if perms.dcim.add_poweroutlet %}
                     <li><a class="dropdown-item" href="{% url 'dcim:poweroutlet_add' %}?device={{ object.pk }}&return_url={% url 'dcim:device_poweroutlets' pk=object.pk %}">{% trans "Power Outlets" %}</a></li>
                 {% endif %}
+                {% if perms.dcim.add_coolingintake %}
+                    <li><a class="dropdown-item" href="{% url 'dcim:coolingintake_add' %}?device={{ object.pk }}&return_url={% url 'dcim:device_coolingintakes' pk=object.pk %}">{% trans "Cooling Ports" %}</a></li>
```
jeremystretch:
```suggestion
                    <li><a class="dropdown-item" href="{% url 'dcim:coolingintake_add' %}?device={{ object.pk }}&return_url={% url 'dcim:device_coolingintakes' pk=object.pk %}">{% trans "Cooling Intakes" %}</a></li>
```

netbox/templates/dcim/device/base.html
```diff
@@ -33,6 +33,12 @@
...
+                {% endif %}
+                {% if perms.dcim.add_coolingoutflow %}
+                    <li><a class="dropdown-item" href="{% url 'dcim:coolingoutflow_add' %}?device={{ object.pk }}&return_url={% url 'dcim:device_coolingoutflows' pk=object.pk %}">{% trans "Cooling Outlets" %}</a></li>
```
jeremystretch:
```suggestion
                    <li><a class="dropdown-item" href="{% url 'dcim:coolingoutflow_add' %}?device={{ object.pk }}&return_url={% url 'dcim:device_coolingoutflows' pk=object.pk %}">{% trans "Cooling Outflows" %}</a></li>
```

netbox/templates/dcim/devicetype/base.html
```diff
@@ -25,6 +25,12 @@
         {% if perms.dcim.add_poweroutlettemplate %}
           <li><a class="dropdown-item" href="{% url 'dcim:poweroutlettemplate_add' %}?device_type={{ object.pk }}&return_url={% url 'dcim:devicetype_poweroutlets' pk=object.pk %}">{% trans "Power Outlets" %}</a></li>
         {% endif %}
+        {% if perms.dcim.add_coolingintaketemplate %}
+          <li><a class="dropdown-item" href="{% url 'dcim:coolingintaketemplate_add' %}?device_type={{ object.pk }}&return_url={% url 'dcim:devicetype_coolingintakes' pk=object.pk %}">{% trans "Cooling Ports" %}</a></li>
```
jeremystretch:
```suggestion
          <li><a class="dropdown-item" href="{% url 'dcim:coolingintaketemplate_add' %}?device_type={{ object.pk }}&return_url={% url 'dcim:devicetype_coolingintakes' pk=object.pk %}">{% trans "Cooling Intakes" %}</a></li>
```

netbox/templates/dcim/devicetype/base.html
```diff
@@ -25,6 +25,12 @@
...
+        {% endif %}
+        {% if perms.dcim.add_coolingoutflowtemplate %}
+          <li><a class="dropdown-item" href="{% url 'dcim:coolingoutflowtemplate_add' %}?device_type={{ object.pk }}&return_url={% url 'dcim:devicetype_coolingoutflows' pk=object.pk %}">{% trans "Cooling Outlets" %}</a></li>
```
jeremystretch:
```suggestion
          <li><a class="dropdown-item" href="{% url 'dcim:coolingoutflowtemplate_add' %}?device_type={{ object.pk }}&return_url={% url 'dcim:devicetype_coolingoutflows' pk=object.pk %}">{% trans "Cooling Outflows" %}</a></li>
```

netbox/dcim/models/devices.py
```diff
@@ -272,6 +288,14 @@ def to_yaml(self):
             data['power-outlets'] = [
                 c.to_yaml() for c in self.poweroutlettemplates.all()
             ]
+        if self.coolingintaketemplates.exists():
+            data['cooling-ports'] = [
```
jeremystretch:
```suggestion
            data['cooling-intakes'] = [
```

netbox/dcim/models/devices.py
```diff
@@ -272,6 +288,14 @@ def to_yaml(self):
...
+                c.to_yaml() for c in self.coolingintaketemplates.all()
+            ]
+        if self.coolingoutflowtemplates.exists():
+            data['cooling-outlets'] = [
```
jeremystretch:
```suggestion
            data['cooling-outflows'] = [
```

netbox/dcim/views.py
```diff
@@ -1655,6 +1690,8 @@ class DeviceTypeImportView(generic.BulkImportView):
         'console-server-ports': forms.ConsoleServerPortTemplateImportForm,
         'power-ports': forms.PowerPortTemplateImportForm,
         'power-outlets': forms.PowerOutletTemplateImportForm,
+        'cooling-ports': forms.CoolingIntakeTemplateImportForm,
+        'cooling-outlets': forms.CoolingOutflowTemplateImportForm,
```
jeremystretch:
```suggestion
        'cooling-intakes': forms.CoolingIntakeTemplateImportForm,
        'cooling-outflows': forms.CoolingOutflowTemplateImportForm,
```

netbox/dcim/views.py
```diff
@@ -4476,6 +4807,28 @@ class DeviceBulkAddPowerOutletView(generic.BulkComponentCreateView):
     default_return_url = 'dcim:device_list'
 
 
+class DeviceBulkAddCoolingIntakeView(generic.BulkComponentCreateView):
```
jeremystretch, line 4815: Need to add links to these two views in `netbox/templates/dcim/buttons/bulk_add_components.html`.

netbox/dcim/views.py
```diff
@@ -2918,6 +3071,38 @@ class DevicePowerOutletsView(DeviceComponentsView):
...
+    child_model = CoolingIntake
+    table = tables.DeviceCoolingIntakeTable
+    filterset = filtersets.CoolingIntakeFilterSet
+    filterset_form = forms.CoolingIntakeFilterForm
+    actions = (EditObject, DeleteObject, BulkEdit, BulkRename, BulkDelete)
+    tab = ViewTab(
+        label=_('Cooling Intakes'),
+        badge=lambda obj: obj.cooling_intake_count,
+        permission='dcim.view_coolingintake',
+        weight=570,
```
jeremystretch: Tab weight conflicts with DevicePowerPortsView.

netbox/dcim/views.py
```diff
@@ -2918,6 +3071,38 @@ class DevicePowerOutletsView(DeviceComponentsView):
...
+    child_model = CoolingOutflow
+    table = tables.DeviceCoolingOutflowTable
+    filterset = filtersets.CoolingOutflowFilterSet
+    filterset_form = forms.CoolingOutflowFilterForm
+    actions = (EditObject, DeleteObject, BulkEdit, BulkRename, BulkDelete)
+    tab = ViewTab(
+        label=_('Cooling Outflows'),
+        badge=lambda obj: obj.cooling_outflow_count,
+        permission='dcim.view_coolingoutflow',
+        weight=580,
```
jeremystretch: Tab weight conflicts with DevicePowerOutletsView.

docs/models/dcim/coolingoutflow.md
```diff
@@ -0,0 +1,42 @@
...
+The connector diameter, expressed as a numeric value with a selectable unit (millimeters, centimeters, or inches).
+
+### Cooling Intake
+
+The upstream [cooling intake](./coolingintake.md) on the same device which feeds this outlet (optional).
+
+### Mark Connected
+
+If selected, this component will be treated as if a cable has been connected.
```
jeremystretch: This field doesn't exist.
```suggestion
```

docs/models/dcim/coolingintake.md
jeremystretch, line 1: Model documentation for CoolingIntake is missing the cooling outflow assocation.

docs/models/dcim/coolingintake.md
```diff
@@ -0,0 +1,36 @@
+# Cooling Intakes
+
+A cooling intake is a device component which consumes coolant, such as a server cold-plate inlet or a coolant distribution unit (CDU) intake. It **receives** coolant from the cold, supply side of a loop (see [cooling](../../features/cooling.md) for the overall flow model). A cooling intake optionally references the upstream [cooling outflow](./coolingoutflow.md) which supplies it. The [cooling feed](./coolingfeed.md) an intake belongs to is derived from the rack in which its device is installed, rather than referenced directly.
```
jeremystretch: > The cooling feed an intake belongs to is derived from the rack in which its device is installed, rather than referenced directly.

Delete this?

mkdocs.yml
```diff
@@ -201,6 +202,12 @@ nav:
             - ConsolePortTemplate: 'models/dcim/consoleporttemplate.md'
             - ConsoleServerPort: 'models/dcim/consoleserverport.md'
             - ConsoleServerPortTemplate: 'models/dcim/consoleserverporttemplate.md'
+            - CoolingFeed: 'models/dcim/coolingfeed.md'
+            - CoolingOutflow: 'models/dcim/coolingoutflow.md'
+            - CoolingOutflowTemplate: 'models/dcim/coolingoutflowtemplate.md'
+            - CoolingIntake: 'models/dcim/coolingintake.md'
+            - CoolingIntakeTemplate: 'models/dcim/coolingintaketemplate.md'
```
jeremystretch: Correct alphabetical ordering
```suggestion
            - CoolingIntake: 'models/dcim/coolingintake.md'
            - CoolingIntakeTemplate: 'models/dcim/coolingintaketemplate.md'
            - CoolingOutflow: 'models/dcim/coolingoutflow.md'
            - CoolingOutflowTemplate: 'models/dcim/coolingoutflowtemplate.md'
```

netbox/templates/dcim/coolingfeed.html
jeremystretch, line 1: The four templates that exist solely to define breadcrumbs can likely be removed in favor of declaring breadcrumbs directly on their corresponding view classes. (See `PowerPanelView` for reference.)

netbox/templates/dcim/coolingfeed/attrs/connected_device.html
jeremystretch, line 1: Unused template?

netbox/dcim/ui/panels.py
```diff
@@ -399,6 +429,31 @@ class PowerFeedElectricalPanel(panels.ObjectAttributesPanel):
...
+
+class CoolingFeedPanel(panels.ObjectAttributesPanel):
+    cooling_source = attrs.RelatedObjectAttr('cooling_source', linkify=True)
+    rack = attrs.RelatedObjectAttr('rack', linkify=True)
+    status = attrs.ChoiceAttr('status')
+    description = attrs.TextAttr('description')
+    tenant = attrs.RelatedObjectAttr('tenant', linkify=True, grouped_by='group')
+
+
+class CoolingFeedElectricalPanel(panels.ObjectAttributesPanel):
```
jeremystretch: Panel class should be renamed or just merged into CoolingFeedPanel.

netbox/dcim/models/device_components.py
```diff
@@ -684,6 +686,83 @@ def get_status_color(self):
...
+        null=True,
+        help_text=_('Physical connector type')
+    )
+    # diameter, diameter_unit, _abs_diameter provided by DiameterMixin
+    cooling_intake = models.ForeignKey(
+        to='dcim.CoolingIntake',
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+        related_name='coolingoutflows'
```
jeremystretch, line 760: Inconsistent with `related_name='cooling_intakes'` on `CoolingIntake.cooling_outflow`. Not sure which is correct TBH. (Same issue with `CoolingOutflowTemplate`.)

netbox/dcim/api/views.py
```diff
@@ -477,6 +489,22 @@ class PowerOutletViewSet(PathEndpointMixin, NetBoxModelViewSet):
     filterset_class = filtersets.PowerOutletFilterSet
 
 
+class CoolingIntakeViewSet(NetBoxModelViewSet):
+    queryset = CoolingIntake.objects.prefetch_related(
+        'cooling_outflow',
+    )
```
jeremystretch: Prefetch should be handled dynamically

netbox/dcim/api/views.py
```diff
@@ -477,6 +489,22 @@ class PowerOutletViewSet(PathEndpointMixin, NetBoxModelViewSet):
...
+    serializer_class = serializers.CoolingIntakeSerializer
+    filterset_class = filtersets.CoolingIntakeFilterSet
+
+
+class CoolingOutflowViewSet(NetBoxModelViewSet):
+    queryset = CoolingOutflow.objects.prefetch_related(
+        'cooling_intake',
+    )
```
jeremystretch: Prefetch should be handled dynamically

netbox/dcim/api/views.py
```diff
@@ -621,6 +649,28 @@ class PowerFeedViewSet(PathEndpointMixin, NetBoxModelViewSet):
...
+
+
+#
+# Cooling feeds
+#
+
+class CoolingFeedViewSet(NetBoxModelViewSet):
+    queryset = CoolingFeed.objects.prefetch_related(
+        'cooling_source', 'rack', 'tenant',
+    )
```
jeremystretch: Prefetch should be handled dynamically

netbox/dcim/models/racks.py
```diff
@@ -135,6 +135,23 @@ class RackBase(WeightMixin, PrimaryModel):
...
 
+    # Cooling
+    cooling_capability = models.CharField(
+        verbose_name=_('cooling capability'),
+        max_length=50,
+        choices=RackCoolingCapabilityChoices,
+        blank=True,
+        null=True
+    )
+    cooling_capacity = models.DecimalField(
```
jeremystretch, line 146: Needs MinValueValidator(0) to reject negative values

netbox/dcim/models/racks.py
```diff
@@ -135,6 +135,23 @@ class RackBase(WeightMixin, PrimaryModel):
...
+        verbose_name=_('cooling capacity'),
+        max_digits=8,
```
jeremystretch: Should be 10 to match `cooling_capacity` on CoolingSource and CoolingFeed.
```suggestion
        max_digits=10,
```

netbox/dcim/choices.py
```diff
@@ -2116,6 +2116,126 @@ class PowerOutletStatusChoices(ChoiceSet):
...
+
+    CHOICES = [
+        (METHOD_AIR, _('Air'), 'cyan'),
+        (METHOD_LIQUID, _('Liquid'), 'blue'),
+        (METHOD_HYBRID, _('Hybrid'), 'purple'),
+        (METHOD_IMMERSION, _('Immersion'), 'indigo'),
+    ]
+
+
+class CoolingSourceTypeChoices(ChoiceSet):
```
jeremystretch, line 2137: Should we add a `key` to enable admins to add custom choices?

netbox/dcim/choices.py
```diff
@@ -2116,6 +2116,126 @@ class PowerOutletStatusChoices(ChoiceSet):
...
+
+    CHOICES = [
+        (STATUS_OFFLINE, _('Offline'), 'gray'),
+        (STATUS_ACTIVE, _('Active'), 'green'),
+        (STATUS_PLANNED, _('Planned'), 'blue'),
+        (STATUS_FAILED, _('Failed'), 'red'),
+    ]
+
+
+class FluidTypeChoices(ChoiceSet):
```
jeremystretch, line 2187: Suggest adding a `key` to allow custom choices and removing `FLUID_OTHER`.

netbox/dcim/filtersets.py
```diff
@@ -3121,6 +3220,172 @@ def search(self, queryset, name, value):
...
+        )
+
+    def search(self, queryset, name, value):
+        if not value.strip():
+            return queryset
+        qs_filter = (
+            Q(name__icontains=value) |
+            Q(description__icontains=value) |
+            Q(cooling_source__name__icontains=value) |
+            Q(comments__icontains=value)
```
jeremystretch, line 3415: I think we typically omit `comments` from the quick search filter. (If not, this is still inconsistent with `CoolingSourceFilterSet`.)

jeremystretch (changes requested):
Remember to run a Claude `/review` once all changes have been addressed.

netbox/dcim/filtersets.py
jeremystretch: Disregard; we need to add `comments` on CoolingSourceFilterSet instead.

netbox/dcim/filtersets.py
```diff
@@ -3121,6 +3220,172 @@ def search(self, queryset, name, value):
...
+        fields = (
+            'id', 'name', 'fluid_type', 'cooling_capacity', 'description',
+        )
+
+    def search(self, queryset, name, value):
+        if not value.strip():
+            return queryset
+        qs_filter = (
+            Q(name__icontains=value) |
+            Q(description__icontains=value)
```
jeremystretch: Add `comments`
```suggestion
            Q(description__icontains=value) |
            Q(comments__icontains=value)
```

arthanson:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Also see comments on previous review of Declined requested changes and reasoning.

arthanson:
Per Claudes review:

1. fixed
2. There is no way to know like hybrid can have air or water, also they are extensible choices so no good way to restrict this.

netbox/dcim/models/device_components.py
```diff
@@ -684,6 +686,125 @@ def get_status_color(self):
         return PowerOutletStatusChoices.colors.get(self.status)
 
 
+#
+# Cooling components
+#
+
+def validate_cooling_loop(component):
```
jeremystretch: Can this be a method on CoolingIntake?

netbox/dcim/models/mixins.py
```diff
@@ -229,3 +239,115 @@ def clean(self):
         # RF role may be set only for wireless interfaces
         if self.rf_role and self.type not in WIRELESS_IFACE_TYPES:
             raise ValidationError({'rf_role': _("Wireless role may be set only on wireless interfaces.")})
+
+
+def normalize_measurement_field(instance, value_field, unit_field, abs_field, converter):
```
jeremystretch: Move to a model method

netbox/dcim/models/mixins.py
```diff
@@ -229,3 +239,115 @@ def clean(self):
...
+    bulk_create), so the normalization logic has a single source of truth.
+    """
+    value = getattr(instance, value_field)
+    unit = getattr(instance, unit_field)
+    setattr(instance, abs_field, converter(value, unit) if value is not None and unit else None)
+    if value is None:
+        setattr(instance, unit_field, None)
+
+
+def validate_measurement_unit(instance, value_field, unit_field, error_message):
```
jeremystretch: Move to a model method

netbox/dcim/views.py
```diff
@@ -65,10 +65,13 @@
     'dcim.consoleserverport': ConsoleServerPort,
     'dcim.powerport': PowerPort,
     'dcim.poweroutlet': PowerOutlet,
+    'dcim.coolingintake': CoolingIntake,
+    'dcim.coolingoutflow': CoolingOutflow,
     'dcim.interface': Interface,
     'dcim.frontport': FrontPort,
     'dcim.rearport': RearPort,
     'dcim.powerfeed': PowerFeed,
+    'dcim.coolingfeed': CoolingFeed,
```
jeremystretch: Remove; no longer needed.

netbox/dcim/graphql/filters.py
```diff
@@ -979,6 +994,129 @@ class PowerPortTemplateFilter(ModularComponentTemplateFilterMixin, ChangeLoggedM
...
+    diameter_unit: BaseFilterLookup[Annotated['DiameterUnitEnum', strawberry.lazy('dcim.graphql.enums')]] | None = (
+        strawberry_django.filter_field()
+    )
+    cooling_intake: Annotated['CoolingIntakeFilter', strawberry.lazy('dcim.graphql.filters')] | None = (
+        strawberry_django.filter_field()
+    )
+    cooling_intake_id: ID | None = strawberry_django.filter_field()
+    color: BaseFilterLookup[Annotated['ColorEnum', strawberry.lazy('netbox.graphql.enums')]] | None = (
+        strawberry_django.filter_field()
+    )
```
jeremystretch: Remove; there is no `color` field on CoolingOutflow

netbox/dcim/forms/filtersets.py
```diff
@@ -1683,6 +1850,133 @@ class PowerOutletTemplateFilterForm(ModularDeviceComponentTemplateFilterForm):
...
+        label=_('Diameter unit'),
+        choices=add_blank_choice(DiameterUnitChoices),
+        required=False
+    )
+    maximum_flow_unit = forms.MultipleChoiceField(
+        label=_('Maximum flow unit'),
+        choices=FlowRateUnitChoices,
+        required=False
+    )
+    cooling_outflow = DynamicModelChoiceField(
```
jeremystretch: Correct filter name & change to DynamicModelMultipleChoiceField
```suggestion
    cooling_outflow_id = DynamicModelMultipleChoiceField(
```

jeremystretch (changes requested):

jeremystretch (approved):

PR #22520, opened 2026-06-24
Closes #18821: Simplify setting/updating primary MAC through interface model

### LINKED ISSUE #18821: Simplify setting/updating primary MAC through interface model.
### NetBox version

v4.2.4

### Feature type

### Proposed functionality

Hello! I'd like to see the ability to set & update the primary MAC address in one step through actions on the interface (or VM interface) model directly. See related discussion #18807. This proposal relates to both GUI & API.

### GUI
Change the interface creation & update request processing to create/update/delete MAC address models in the background once we hit submit on the form. Something like this:
1. Create/update interface
2. Create/update MAC assigned to new/updated interface
3. Update interface to set the MAC as primary.

I think the "quick add" MAC address mini form could remain (minus the interface selection) but the related request wouldn't happen until after hitting submit on the main form.

### API
Add a similar behind-the-scenes workflow when handling a request on `dcim.interface` or `virtualization.vminterface` that basically does the same as above.

### Use case

The majority of interfaces have only one MAC so I think ease of use for that use case is important. Unfortunately, the new MAC address model additions make it a little harder to manage this.

The current GUI workflow for managing interfaces & primary MACs is not smooth when creating new interfaces. The MAC "quick add" function does not allow assigning the new MAC to the new interface you are creating (because the MAC is created before the interface exists). The "quick add" works fine when updating existing interfaces to add a primary MAC.

The fastest existing process is:
1. Create new interface without a MAC
2. Edit the interface
a. "Quick add" a primary MAC (now defaults to assigning it to this interface)
b. Save interface with new MAC

When using the API, the fastest existing process is:
1. Create interface object without a MAC.
2. Create MAC address object w/ assigned interface ID.
3. Update interface to set the MAC address as primary.

I'm essentially suggesting that these steps be put under the hood to allow for a single-request experience as it was before `v4.2`.

Thanks for any consideration or ideas.

### Database changes

_No response_

### External dependencies

_No response_

bctiemann:
## Summary

- **API**: Makes `mac_address` writable on `InterfaceSerializer` and `VMInterfaceSerializer`. Posting a MAC string creates or promotes an existing `MACAddress` object as the primary in a single request; posting `null` clears the primary; omitting the field is a no-op. Find-or-create logic prevents duplicates when a matching MAC already exists on the interface.
- **Forms**: Replaces the `primary_mac_address` DynamicModelChoiceField (a FK picker) in `InterfaceForm` / `VMInterfaceForm` with a plain text `mac_address` field shared via `InterfaceCommonForm`. On save, the same find-or-create logic applies.
- **UI**: Adds a "Set as primary" action to the `MACAddressTable` actions dropdown (visible only for non-primary, assigned MACs), backed by `MACAddressSetPrimaryView`. Adds a pre-scoped "Add" button to the MAC Addresses inline panel on the Interface detail view.

Screenshot 2026-06-23 at 9 32 20 PM

### What was removed and why

The previous `InterfaceForm` included a `primary_mac_address` DynamicModelChoiceField — a searchable picker that listed all `MACAddress` objects already assigned to the interface and let the user select one as primary. It also had a **Quick Add** inline widget that could create a new `MACAddress` directly from the dropdown.

This design had several problems:

- Creating an interface with a primary MAC required three steps: create the interface, create a `MACAddress`, then PATCH `primary_mac_address`. The FR goal was to collapse this to one.
- The picker only exposed a single workflow (assign an *existing* MAC as primary) and the quick-add widget was a blunt workaround for the missing create-in-one-step flow.
- The picker could not show MACs that do not exist yet, so "create and assign" still required leaving the form.

The replacement — a plain text `mac_address` field — directly addresses the FR: the user types a MAC string, submits the form, and both the `MACAddress` record and the primary assignment are created in one round-trip.

### What was not replaced (and the mitigation)

The picker gave the user a visual list of MACs already on the interface, making it easy to switch which one is primary. The plain text field requires the user to know (or copy) the MAC string they want to promote.

The mitigation is the **"Set as primary" action** added to the `MACAddressTable` dropdown on the Interface detail view. A user who wants to change which MAC is primary can:

1. Navigate to the Interface detail view.
2. Find the desired MAC in the inline table.
3. Click **Actions → Set as primary**.

Screenshot 2026-06-23 at 9 32 09 PM

This is arguably a better UX for that workflow than the form picker, since the table shows all assigned MACs with context (including the `is_primary` indicator column), whereas the picker in the edit form only showed raw MAC strings with no surrounding context.

### Remaining UX gap

The edit form no longer surfaces which MACs are available to promote. A user editing an interface must either know the MAC string in advance or switch to the detail view to use the "Set as primary" action. For interfaces with many MACs this is a workflow interruption. A future improvement could add a read-only annotation to the `mac_address` field (e.g., "other MACs assigned: …") or restore a non-editable list of assigned MACs alongside the text field.

### API backward compatibility

The `mac_address` field on both serializers was already returned read-only (reflecting the primary MAC string). This PR makes it writable while preserving the same read behavior, so existing API clients that only read the field are unaffected. Clients that explicitly POST `mac_address: null` will now clear the primary assignment (previously this was silently ignored).

## Test plan

- [ ] `dcim.tests.test_api.InterfaceTestCase` — covers `mac_address` create, update (including existing-MAC find-or-create path), clear, and invalid format
- [ ] `virtualization.tests.test_api.VMInterfaceTestCase` — same coverage for `VMInterface`
- [ ] Manual: create an Interface via the edit form with a new MAC string → MAC and primary assignment created
- [ ] Manual: edit an Interface with an existing primary MAC → field pre-populated, value round-trips
- [ ] Manual: on Interface detail view, click **Actions → Set as primary** on a non-primary MAC → primary updated, redirected to interface
- [ ] Manual: "Set as primary" absent from dropdown for already-primary MACs and unassigned MACs
- [ ] Manual: "Add" button in MAC Addresses panel opens MACAddress form pre-scoped to the current interface


bctiemann:
@claude /review

bctiemann:
Addressed all findings from the review. Here's what changed in the latest two commits:

### Security — form-layer permission check

`InterfaceCommonForm.clean()` now checks `dcim.add_macaddress` before accepting a MAC address value, using NetBox's `current_request` ContextVar (set for every request via the `event_tracking` context manager). Raises a proper form validation error shown inline, rather than a 403 page.

The check is in `clean()` rather than `save()` because the generic `ObjectEditView` does not pass `request` to forms — `current_request.get()` works either way, but a clean-time error gives better UX. The check is slightly conservative (fires even if the MAC already exists on the interface and would be found rather than created), which is the right trade-off.

### Correctness — `snapshot()` and `transaction.atomic()` in form layer

`InterfaceCommonForm.save()` now calls `instance.snapshot()` immediately before each secondary `instance.save()` that sets or clears `primary_mac_address`, matching the pattern already in both serializers. The MAC writes are also wrapped in `transaction.atomic()`.

### Code quality — `MACAddressShortcutMixin`

Extracted the identical `create()` and `update()` methods from `InterfaceSerializer` and `VMInterfaceSerializer` into a new `dcim/api/serializers_/mixins.py`. `VMInterfaceSerializer` shed its `transaction`, `PermissionDenied`, and `MACAddress` imports entirely (~90 lines of net duplication removed).

The `validate()` methods were left in each serializer: `InterfaceSerializer.validate()` has ~60 lines of mode/VLAN/Q-in-Q logic that differs substantially from `VMInterfaceSerializer.validate()`'s simpler VM-site check, and a hook-based approach would add abstraction overhead without enough clarity benefit.

### Code quality — `format_html` in `MACAddressActionsColumn`

Replaced the f-string `form_li` construction with `format_html()`, which escapes each argument automatically. The `mark_safe()` on the final `str.replace()` result remains necessary — string manipulation on a `SafeString` produces a plain `str` — but the individual inputs are now properly escaped.

### Test coverage — find-or-create path

Added `test_mac_address_find_or_create`: creates two `MACAddress` records assigned to the same interface, PATCHes with the second MAC's address string, and asserts (a) the second MAC is now primary and (b) the total MAC count on the interface is unchanged — no duplicate was created.

bctiemann:
@claude /review latest changes and compare against previous review.

bctiemann:
Follow-up fixes from the re-review:

### Medium — view tests were missing `dcim.add_macaddress`

`test_mac_address_shortcut_create` and `test_mac_address_shortcut_edit` in both `dcim/tests/test_views.py` and `virtualization/tests/test_views.py` only granted the interface add/change permission. The `clean()` security check fires in tests — `current_request` is set via middleware for all Django test client requests — so those tests were now correctly failing with 200 (form validation error) instead of 302. Added `'dcim.add_macaddress'` to `add_permissions` in all four tests.

### Minor — spurious `save()` in form clear branch

`InterfaceCommonForm.save()` was calling `instance.save()` unconditionally in the "clear primary MAC" branch, even when `primary_mac_address_id` was already `None`. This fired a no-op `post_save` signal and generated a spurious changelog entry when a user saved the form with an empty MAC field that was already empty. Fixed by moving the assignment and save inside the `if instance.primary_mac_address_id is not None` guard.

### Minor — missing `test_mac_address_find_or_create` for VMInterface

Added the find-or-create test to `VMInterfaceTestCase` in `virtualization/tests/test_api.py`, mirroring the one added earlier for `InterfaceTestCase`.

### Nit — lazy import of `redirect_to_login`

Moved `from django.contrib.auth.views import redirect_to_login` from inside `MACAddressSetPrimaryView.post()` to the module-level imports in `dcim/views.py`.

arthanson (changes requested):
The edit of Interface is a bit confusing, if you already have a MAC address assigned nothing shows up in the edit form and if you type a MAC address it just gets appended to the list. The comment says it will be assigned as primary, but IMHO it is confusing as everyone else in edit shows if something is already set.  Maybe we could show teh currently set MAC address above the edit field or something, or show the currently existing primary MAC address in the field

bctiemann:
I'm not sure what you mean -- the edit form for the Interface shows the primary MAC address filled in, and if you enter one that is already assigned to that Interface it switches Primary to that one rather than creating one, at least for me

Screenshot 2026-06-29 at 7 15 02 AM

arthanson (approved):

PR #22524, opened 2026-06-24
Closes #22441: Add execution_time to background jobs

### LINKED ISSUE #22441: Add execution time to the background jobs table
### NetBox version

v4.6.2

### Feature type

New functionality

### Proposed functionality

Add a table column showing the execution time for completed jobs to the background jobs table. This is calculated as `completed - started`. Ideally, this column should support ordering.

### Use case

This will enable users to quickly determine the time each job took to execute, and long-running jobs will be more prominent.

### Database changes

This is probably best implemented by adding a read-only `execution_time` column to the Job model, to ensure that it can be used for ordering.

### External dependencies

N/A

jnovinger:
### Closes: #22441

Adds an `execution_time` field to the `Job` model, computed as `completed - started` when a job finishes.

**What's added:**

- Orderable `Execution Time` column in the jobs table (non-default; enable via Configure Table)
- `Execution time` row in the job detail Scheduling panel
- `execution_time` field in the REST API (read-only)
- `execution_time`, `execution_time__gte`, `execution_time__lte` filters in the UI and API

**Implementation notes:**

- Plain nullable `DurationField(editable=False)`, set in `Job.terminate()` — not a `GeneratedField`. A stored generated column requires a full table rewrite on `core_job` on migration; a plain field keeps it metadata-only (`ADD COLUMN ... NULL`).
- No historical backfill. Only jobs completing after migration get a value; matches the convention for every prior Job field addition.
- `Job.duration` is deliberately untouched. It computes `completed - (started or created)` and returns a formatted string consumed by the script-result template and `runscript` command. These are distinct quantities.
- REST API renders `execution_time` as Django's default duration format (`HH:MM:SS.ffffff`), not ISO-8601.

**Testing:**

- Model: `terminate()` sets field; null before completion; null when job never started
- Filterset: exact filter + `__gte`/`__lte` range; `test_missing_filters` passes
- API: field present; `?ordering=execution_time`; `?execution_time__gte=` filter
- `humanize_duration` unit tests

### DIFF
--- a/docs/models/core/job.md
+++ b/docs/models/core/job.md
@@ -28,6 +28,10 @@ The interval (in minutes) at which a scheduled job should re-execute.
 
 The date and time at which the job completed (if complete).
 
+### Execution Time
+
+The amount of time the job spent executing, calculated as the difference between its start and completion times. This is populated only once a started job has completed.
+
 ### User
 
 The user who created the job.
--- a/netbox/core/api/serializers_/jobs.py
+++ b/netbox/core/api/serializers_/jobs.py
@@ -32,8 +32,8 @@ class JobSerializer(BaseModelSerializer):
         model = Job
         fields = [
             'id', 'url', 'display_url', 'display', 'object_type', 'object_id', 'object', 'name', 'status', 'created',
-            'scheduled', 'interval', 'started', 'completed', 'user', 'data', 'error', 'job_id', 'queue_name',
-            'notifications', 'log_entries',
+            'scheduled', 'interval', 'started', 'completed', 'execution_time', 'user', 'data', 'error', 'job_id',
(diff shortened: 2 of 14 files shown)


jnovinger:

jnovinger:

Addressed since your last review:

- `humanize_duration` now guards on `value is None` instead of falsiness, so `timedelta(0)` renders as `0s` rather than an empty cell
- Added `label=_('Execution time')` to the bare `execution_time` filter

One finding not addressed: `execution_time` appears in both the explicit class attribute and `Meta.fields`. Left as-is because `status`, `queue_name`, and `user` have the same redundancy in this filterset. Removing it from `Meta.fields` would make it the exception, not the rule.

bctiemann (approved):

PR #22535, opened 2026-06-24
Fixes #22528: Fix keyboard navigation for object list tabs

### LINKED ISSUE #22528: Results/Filters tabs not keyboard operable (A11Y-4487)
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.14

### Steps to Reproduce

1. Navigate to an objects list.
2. Attempt to switch between the "Results" and "Filters" tabs using only the keyboard.

### Expected Behavior

The arrow keys can be used to navigate between the tabs.

### Observed Behavior

The Results and Filters tabs cannot be reached/operated with the keyboard alone.

**Suggested fix:** Implement the WAI-ARIA [Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) — proper `role="tablist"/"tab"/"tabpanel"`, `aria-selected`, roving `tabindex`, and arrow-key navigation. (As these are Bootstrap tabs, ensure the keyboard JS is wired up and not suppressed.)

jeremystretch:
### Fixes: #22528

The Results/Filters tabs on object list pages were not keyboard operable — arrow keys did nothing and the Results tab could not receive focus.

The "Results" tab in `generic/object_list.html` was rendered as an `<a>` element with no `href`. Anchors without an `href` are not in the keyboard tab order, so the tab could never be focused. Bootstrap 5.3 binds its arrow-key `keydown` handler to the active tab (auto-initialized on page load), so with an unfocusable active tab that handler never fired and arrow-key navigation was dead. The element even carried an invalid `type="button"` attribute, indicating it was always meant to be a button. The adjacent "Filters" tab was already a `<button>`.

This change converts the "Results" tab from `<a>` to `<button>`, matching the "Filters" tab. A `<button>` is natively focusable, so keyboard users can Tab into the tablist and Bootstrap's roving-tabindex / arrow-key / Home / End navigation works as expected. The existing ARIA roles and attributes (`tablist`/`tab`/`tabpanel`, `aria-selected`, `aria-controls`) are unchanged.

This is a template-only change; no JavaScript rebuild is required.

### DIFF
--- a/netbox/templates/generic/object_list.html
+++ b/netbox/templates/generic/object_list.html
@@ -38,10 +38,10 @@ Context:
 {% block tabs %}
   <ul class="nav nav-tabs" role="tablist">
     <li class="nav-item" role="presentation">
-      <a class="nav-link active" id="object-list-tab" data-bs-toggle="tab" data-bs-target="#object-list" type="button" role="tab" aria-controls="object-list" aria-selected="true">
+      <button class="nav-link active" id="object-list-tab" data-bs-toggle="tab" data-bs-target="#object-list" type="button" role="tab" aria-controls="object-list" aria-selected="true">
         {% trans "Results" %}
         <span class="badge text-bg-secondary total-object-count">{% if table.page.paginator.count %}{{ table.page.paginator.count }}{% else %}{{ total_count|default:"0" }}{% endif %}</span>
-      </a>
+      </button>
     </li>
     {% if filter_form %}
       <li class="nav-item" role="presentation">


jeremystretch:

jnovinger (approved):
Approving this as-is, since it's what the issue called for. However, there were some follow-up a11y issues found while reviewing this PR:

- `core/system.html` has five interactive tabs (Status, Config, Plugins, Object Counts, Database) all using `<a>` with `data-bs-toggle="tab"` instead of `<button>`. Same keyboard navigation failure as the issue fixed here — highest priority.
- `core/plugin.html` mixes `<a>` (Overview tab) and `<button>` (Install tab) in the same tablist.
- Several single-tab decorative headers (`core/rq_task_list.html`, `rq_worker_list.html`, `rq_queue_list.html`, `rq_worker.html`, `rq_task.html`, `plugin_list.html`, `extras/script_list.html`) use `<a role="tab">` with no Bootstrap toggle wiring. No broken navigation since there are no sibling tabs, but semantically incorrect.

PR #22537, opened 2026-06-25
Closes #19821: Consolidate GFK form handling with GenericObjectChoiceField

### LINKED ISSUE #19821: Introduce a custom form field to represent generic foreign key relations
### NetBox version

v4.3.3

### Feature type

New functionality

### Proposed functionality

Devise a new type of form field and widget capable of setting a GFK relation on an object. This is currently achieved by using two independent form fields which map to the two concrete model fields supporting the GFK (e.g. `content_type` and `object_id`).

Some quick prototyping suggests the new widget would look similar to the following:

```python
from django import forms

class GenericForeignKeyWidget(forms.MultiWidget):
    def __init__(self, content_type_choices=None, attrs=None):
        widgets = [
            forms.Select(choices=content_type_choices),
            forms.TextInput(attrs={'placeholder': 'Object ID'}),
        ]
        super().__init__(widgets, attrs)

    def decompress(self, value):
        if value and hasattr(value, '_meta'):
            ct_id = ContentType.objects.get_for_model(value).pk
            return [ct_id, value.pk]
        return [None, None]
```

And the corresponding form field might look like this:

```python
from django import forms
from django.core.exceptions import ValidationError

class GenericForeignKeyFormField(forms.MultiValueField):
    def __init__(self, content_type_choices=None, **kwargs):
        fields = [
            forms.ChoiceField(choices=content_type_choices),
            forms.CharField()
        ]
        widget = GenericForeignKeyWidget(content_type_choices)
        super().__init__(fields=fields, widget=widget, **kwargs)

    def compress(self, data_list):
        if data_list and all(data_list):
            ct_id, obj_id = data_list
            try:
                content_type = ContentType.objects.get(pk=ct_id)
                model_class = content_type.model_class()
                return model_class.objects.get(pk=obj_id)
            except Exception as e:
                raise ValidationError(f"Invalid GenericForeignKey: {e}")
        return None
```

These examples will need to be modified to employ NetBox's REST API for object selection.

### Use case

If proved viable, this approach will greatly simplify GFK relation handling within model forms by obviating the need to declare two form fields, and by reducing the amount of custom logic needed within each form's `__init__()` and `save()` methods.

### Database changes

N/A

### External dependencies

N/A

pheus:
### Closes: #19821

Replace separate scope_type/scope and parent_object_type/parent field pairs with unified GenericObjectChoiceField. Introduce GenericObjectFormMixin to handle GFK descriptor initialization and assignment.

This removes redundant HTMX/queryset setup logic from ScopedForm, VLANGroupForm, and ServiceForm by delegating GFK presentation to a single reusable field and mixin pair. Field query param references now use `$scope_object_id` instead of `$scope` to match the subwidget name.


pheus:

pheus:

pheus:
@claude /review
I've addressed your points.

The pre-existing VLANGroupBulkEditForm dangling 'site' fieldset is left untouched for a separate follow-up.

netbox/utilities/forms/fields/generic.py
```diff
@@ -0,0 +1,239 @@
...
+        self.selected_model = model
+
+        # Narrow the queryset to the current value to avoid loading the entire table for rendering
+        queryset = self._get_object_queryset(model)
+        if object_value in self.empty_values:
+            queryset = queryset.none()
+        else:
+            lookup = getattr(self.object_field, 'to_field_name', None) or 'pk'
+            try:
+                queryset = django_filters.ModelChoiceFilter(field_name=lookup).filter(queryset, object_value)
```
jeremystretch: Seems odd to use ModelChoiceFilter here. Would `queryset.filter(**{lookup: object_value})` be sufficient?

netbox/utilities/forms/fields/generic.py
```diff
@@ -0,0 +1,239 @@
...
+    @property
+    def queryset(self):
+        # Exposed at the top level so restrict_form_fields() (which only inspects top-level fields) can apply
+        # object-permission restrictions to the nested object selector.
+        return self.object_field.queryset
+
+    @queryset.setter
+    def queryset(self, queryset):
+        self.object_field.queryset = queryset
+        self.object_field.widget.choices = self.object_field.choices
```
jeremystretch, line 103: Per Claude: `queryset` setter writes `choices` to a widget that `_sync_widget_refs()` later re-points. When `restrict_form_fields()` assigns `field.queryset` before `_sync_widget_refs()` runs, the choices land on the field's orphan widget copy and are only recovered because `_sync_widget_refs()` re-derives them onto `self.widget.widgets[1]` afterward. Works today, but the correctness depends on `_sync_widget_refs` always running after any queryset assignment — fragile redundant work.

Not a major concern, but potentially worth exploring.

netbox/utilities/forms/fields/generic.py
```diff
@@ -0,0 +1,239 @@
...
+
+        # Clear the paired object selection client-side when the content type changes, so a stale object_id
+        # cannot cross model boundaries on the HTMX re-render (a Site pk must not resurface as a Region pk).
+        self.content_type_field.widget.attrs['hx-on::config-request'] = (
+            f"event.detail.parameters['{field_name}_object_id'] = ''"
+        )
+
+        content_type_value = self._resolve_subvalue(form, field_name, 'content_type')
+        object_value = self._resolve_subvalue(form, field_name, 'object_id')
+        self._configure_object_field(self._get_content_type(content_type_value), object_value)
```
jeremystretch, line 204: We end up querying the ContentType twice: Once in `GenericObjectFormMixin.__init__` (so forms like `PrefixForm` can read `selected_model` early) and again at render via `get_bound_field()`. Would it be possible to fetch the CT via `ContentType.objects.get_for_id()` to ensure the CT is cached, or do we need to constrain it to the prescribed queryset?

netbox/circuits/forms/bulk_edit.py
```diff
@@ -179,24 +176,18 @@ class CircuitBulkEditForm(PrimaryModelBulkEditForm):
...
-    termination = DynamicModelChoiceField(
+    termination = GenericObjectChoiceField(
         label=_('Termination'),
-        queryset=Site.objects.none(),  # Initial queryset
+        content_type_queryset=ContentType.objects.filter(model__in=CIRCUIT_TERMINATION_TERMINATION_TYPES),
         required=False,
-        disabled=True,
-        selector=True
+        selector=True,
+        hx_method='post',
```
jeremystretch, line 190: This seems to cause an issue: After selecting an object type, the HTMX request is treated as a form submission, and a validation error appears: "Please select a [type]."

jeremystretch (approved):
Had some minor feedback, but approving this as-is in case it can all be safely disregarded. Awesome work!

netbox/utilities/forms/fields/generic.py
pheus: Good catch! Thanks

pheus:
Thanks for the review! This is much appreciated.

I pushed a follow-up for the actionable items. The main change is that bulk-edit HTMX requests without `_apply` are now treated as field refreshes rather than submissions, so the object selector can be rebuilt without showing validation errors too early. The actual Apply path still validates normally.

I also cleaned up the queryset handling, removed the `ModelChoiceFilter` indirection, added a small content type cache, and added tests for the bulk-edit behavior.

I kept the field validation strict intentionally: selecting an object type without an object should still be invalid on submit, just not during the intermediate HTMX refresh.

pheus:

pheus:
@claude /review

The broader HTMX behavior is intentional here: in bulk edit, an HTMX request without `_apply` is only a field refresh, not a submission. That should be true for GFK fields and for other dependent fields as well, including plugin-provided ones. The actual Apply path still uses a bound form and validates normally.

I’m also comfortable keeping the full `normalize_querydict(post_data)` state for now. Some fields, including this generic object field, depend on decomposed subwidget keys like `scope_content_type` and `scope_object_id`, so narrowing this too aggressively could be more fragile than helpful.

Good point on the test data though. I've switched `_apply` from `True` to a string value to better match normal POST data.

netbox/ipam/forms/model_forms.py
```diff
@@ -230,24 +228,24 @@ class PrefixForm(TenancyForm, ScopedForm, PrimaryModelForm):
         FieldSet(
             'prefix', 'status', 'vrf', 'role', 'is_pool', 'mark_utilized', 'description', 'tags', name=_('Prefix')
         ),
-        FieldSet('scope_type', 'scope', name=_('Scope'), html_id='scope'),
+        FieldSet('scope', name=_('Scope'), html_id='scope'),
```
jeremystretch, line 231: The template snippets [here](https://github.com/netbox-community/netbox/blob/15d7de65daa635129ac1b41f65b6e0e3fcb06548/netbox/templates/ipam/prefix/prefixes.html#L9) and [here](https://github.com/netbox-community/netbox/blob/15d7de65daa635129ac1b41f65b6e0e3fcb06548/netbox/ipam/tables/template_code.py#L9) still pass `object_type=` as a URL query parameter. Could we update those?

jeremystretch (changes requested):
One last little gap that popped up.

netbox/ipam/forms/model_forms.py
pheus: Thanks! Good catch. This is valid, and it also lines up with a clone regression I noticed while testing.

jeremystretch:
Fixed a few more pre-populated links that came up.

jeremystretch (approved):

PR #22541, opened 2026-06-25
Fixes #22530: Remove hidden select inputs from the accessibility tree

### LINKED ISSUE #22530: Hidden `<select>` inputs announced by screen reader (A11Y-4493)
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.14

### Steps to Reproduce

This issue was identified during an accessibility audit.

### Expected Behavior

Hidden form elements should not be announced by a screen reader.

### Observed Behavior

Hidden select inputs (the source elements behind the styled comboboxes, from Saved Filter through Contact Group) remain in the DOM without accessible names. JAWS announces a stray "listbox" then "combobox" when navigating the Filters form. Likely a Select2/Tom-Select widget initialization issue affecting all dynamic filter dropdowns.

**Suggested fix:** Remove the hidden native `<select>` from the accessibility tree once the enhanced combobox is initialized (`aria-hidden="true"` / remove from DOM / `display:none`), or give each a proper accessible name.

jeremystretch:
### Fixes: #22530

- Set `aria-hidden="true"` on `<select>` elements wrapped & hidden by Tom-Select


jeremystretch:

arthanson (approved):

PR #22542, opened 2026-06-25
Fixes #22531: Ensure saved filter selector has unique element ID

### LINKED ISSUE #22531: "Saved Filter" combobox label not announced by screen readers (A11Y-4492)
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.14

### Steps to Reproduce

1. Navigate to an objects list using a screen reader.

### Expected Behavior

All interactive elements of the page should be announced.

### Observed Behavior

The Saved Filter combobox's label is not announced by the screen reader, so users can't tell what the field is for. (This likely results from duplicate ids breaking the association — audit for duplicate IDs on the page and make them unique.)

**Suggested fix:** Ensure the label's `for` attribute matches the combobox's `id` (or use `aria-labelledby`).

jeremystretch:
### Problem

The "Saved Filter" combobox's label is not announced by screen readers, so users can't tell what the field is for.

The `filter_id` field is rendered twice on every object-list page from the same form instance: once as the dedicated selector in the list controls (`inc/table_controls_htmx.html`) and again inside the filter drawer via `inc/filter_list.html` fieldsets (`FieldSet('q', 'filter_id', …)`). Both emitted `<select id="id_filter_id">` and `<label for="id_filter_id">`.

The combobox is built by Tom Select, which exposes the field's accessible name by locating `label[for='<select-id>']` and wiring up `aria-labelledby` on the visible combobox. That depends on the select's `id` being unique. The duplicate IDs cascade into duplicate `*-ts-control` / `*-ts-label` IDs once Tom Select runs, so the label association no longer resolves reliably and the field's name is not announced. (NetBox's dynamic Tom Select does not set `controlInput`, so the `aria-label` set on the `<select>` is not copied onto the visible combobox — the label association is the only thing carrying the accessible name.)

### Fix

- `inc/table_controls_htmx.html`: render the list-controls selector with a distinct id (`saved-filter-select`) and point its `<label for>` at it, so it no longer collides with the drawer's copy.
- `form_helpers.py`: add an optional `element_id` argument to the existing `render_field_with_aria` tag to override the widget's HTML id, and use that tag for the selector.
- `savedFiltersSelect.ts`: update the navigate-on-change handler to target `saved-filter-select` (it previously used `getElementById('id_filter_id')`, which would otherwise resolve to the drawer's copy after the rename). Rebuilt `dist/netbox.js`.
- Added a `render_field_with_aria` test case for the `element_id` override.

### Verification

Rendering an object-list page confirms the saved-filter IDs/labels are now unique and correctly paired (one `saved-filter-select`, one `id_filter_id`, each with a matching label). Lint (ruff), prettier, and the templatetags test module pass.


jeremystretch:

jeremystretch:
1. This is acceptable.
2. Does not appear to be an issue.
3. The test coverage is sufficient IMO (`element_id` is not dependent on field type).

pheus (approved):

PR #22545, opened 2026-06-25
Fixes #22532: Ensure all empty table headers have an ARIA label

### LINKED ISSUE #22532: Empty table header cell in Results (A11Y-4491)
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.14

### Steps to Reproduce

1. Navigate to an objects list using a screen reader.

### Expected Behavior

All columns of the results table should be annotated with a header.

### Observed Behavior

The results table contains an empty `<th>` header cell (the checkbox/select column), so data cells aren't properly associated with a header.

**Suggested fix:** Provide an accessible name for the header cell — e.g. visually-hidden text like "Select" inside the `<th>`, or `aria-label`. Applies to the shared object-list table template, so the fix propagates broadly.

jeremystretch:
### Summary

Screen readers encountered empty `<th>` header cells in object tables, so data cells weren't properly associated with a column header. This adds accessible names to all such cells across the application.

### Changes

**Shared table columns** (propagates to every object-list table):
- `ActionsColumn` — the trailing edit/delete/changelog column had an empty header; now `aria-label="Actions"`. This was the empty `<th>` on the results table referenced in the issue.
- `NotificationTable.icon` — icon-only column; now `aria-label="Type"`.

**Hand-written panel/detail tables** — added an `aria-label` to each remaining empty `<th>`:
- `ipam/panels/fhrp_groups.html`, `ipam/inc/panels/fhrp_groups.html` → "Actions"
- `extras/inc/script_list_content.html` → "Actions"
- `dcim/virtualchassis_edit.html` → "Actions"
- `dcim/panels/component_inventory_items.html`, `dcim/inc/panels/inventory_items.html` → "Actions"
- `core/configrevision_restore.html` → "Changed" (change-indicator column)
- `dcim/panels/interface_wireless.html` → "Attribute" (corner cell of the Local/Peer comparison matrix)

The select/checkbox column named in the issue was already accessible (`aria-label="Select all"`, added in #17069); the genuinely empty header on the results table was the trailing actions column.

### Verification

- Programmatic scan of all 168 `BaseTable` subclasses: 0 columns with a blank header and no accessible name (down from 2).
- Template grep: no remaining empty `<th></th>` without an `aria-label`.
- All new labels are i18n-wrapped (`gettext_lazy` / `{% trans %}`).
- `ruff` clean; affected view tests pass.

### DIFF
--- a/netbox/extras/tables/tables.py
+++ b/netbox/extras/tables/tables.py
@@ -424,7 +424,7 @@ class NotificationTable(NetBoxTable):
         orderable=False,
         attrs={
             'td': {'class': 'w-1'},
-            'th': {'class': 'w-1'},
+            'th': {'class': 'w-1', 'aria-label': _('Type')},
         },
         verbose_name=''
     )
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -248,6 +248,9 @@ class ActionsColumn(tables.Column):
         direct button link and icon (default: True)
     """
     attrs = {
+        'th': {
+            'aria-label': _('Actions'),
+        },
         'td': {
             'class': 'text-end text-nowrap noprint p-1'
         }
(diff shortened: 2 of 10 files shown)


jeremystretch:

jnovinger (approved):

PR #22546, opened 2026-06-25
Closes #22446: Add breadcrumbs support for Layouts

### LINKED ISSUE #22446: Add breadcrumbs support for Layouts
### NetBox version

v4.6.2

### Feature type

New functionality

### Proposed functionality

Introduce a `breadcrumbs()` method on `netbox.ui.Layout` that subclasses can override to return an iterable of navigation breadcrumbs that will render at the top of the page. This method will accept the object being viewed and return an iterable of Breadcrumb instances, each of which renders as a label and (optionally) a linked URL for navigation.

The Breadcrumb class should look something like this:

```python
class Breadcrumb:
    template_name = 'ui/breadcrumb.html'

    def __init__(self, instance, url=None):
        self.object = instance
        self.url = url

    def render(self):
        if self.object is None:
            return ''
        if self.url is not None:
            url = self.url
        elif hasattr(self.object, 'get_absolute_url'):
            url = self.object.get_absolute_url()
        else:
            url = None
        context = {
            'url': url,
            'label': str(self.object),
        }
        return render_to_string(self.template_name, context)
```

And the `breadcumbs()` method on Layout would typically be overridden like this:

```python
def breadcrumbs(self, instance):
    return [
        Breadcrumb(instance.site),
        Breadcrumb(instance.location),
        Breadcrumb(instance.rack),
    ]
```

If a rendered Breadcrumb returns an empty string, it is omitted from the list. This helps simplify conditional logic (e.g. where a device may or may not be assigned to a rack).

### Use case

Currently, breadcrumbs must be defined statically within a custom HTML template. Implementing this ability for programmatic layouts will allow us to shed a number of object HTML templates.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #22446

- Introduce `ui.Breadcrumb` for rendering one or multiple breadcrumb objects
- Extend the `Layout` and `SimpleLayout` classes to accept declarative breadcrumbs
- Introduce the `render_breadcrumbs` template tag
- Introduce the `get_view()` utility function to assist in resolving views for a given model/instance
- Update most object detail views to pass breadcrumbs directly and delete/reduce their custom templates
- Removed an incorrect breadcrumb from CircuitGroupView
- Add tests & documentation for Breadcrumb


jeremystretch:

jeremystretch:

jeremystretch:
@claude /review

jeremystretch:
1. This is inaccurate. The view will attempt to load `<app>/<model>.html` (and fail if it doesn't exist) unless `template_name` is defined.
2. Invalid; the object is linkified by dfault.
3. This is intentional (we're including the object's group as "self").
4. No changes were made to these views; out of scope for this FR.
5. This is necessary to support the conditional logic, and acceptable as a corner case.
6. Not an issue
7. Not an issue
8. Not an issue
9. C'mon, Claude...

jnovinger (commented):
A couple of things my local claude found:

> Two small things:
>
> The deleted `ipam/vlan/base.html` also carried `{% block title %}{% trans "VLAN" %} {{ object }}{% endblock %}`, so the VLAN page title changes from "VLAN Data (100)" to "Data (100)" (`generic/object.html` renders just `{{ object }}`). It is the only deleted template that had a title override. `rack/base.html`, `devicetype/base.html`, and `users/token.html` all kept theirs. Intentional?
>
>The old `asn.html` also had a second crumb linking `?range_id={{ object.range.pk }}`. `ASNFilterSet` declares no `range_id` filter, so this looks like the same dead-parameter cleanup as `CircuitGroupView`, just not mentioned in the description. Confirming that was deliberate rather than an accidental drop, since the range label disappears from the trail too.

jeremystretch:
1. This was an intentional removal to standardize the VLAN representation in the view.
2. This was dead code: The ASN model has no `range` attribute, so the breadcrumb would never render anyway.

jnovinger (approved):

PR #22547, opened 2026-06-25
Fixes #22513: Make JournalEntry.created_by immutable after creation

### LINKED ISSUE #22513: JournalEntry.created_by can be reassigned on update via REST API and bulk import (audit-trail spoofing)
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

`JournalEntry.created_by` is writable and is not protected on update, so an existing entry's recorded author can be reassigned to an arbitrary user. Two surfaces:

REST API (`PATCH`):

1. As a user with `extras.change_journalentry`, PATCH an existing entry:
   `PATCH /api/extras/journal-entries/<id>/  {"created_by": <any user id>}`
2. The response shows `created_by` changed to the supplied user.

Bulk import (update mode):

1. As a user with `extras.add_journalentry`, import a CSV with an `id` column referencing an existing entry plus a `created_by` column:

   ```
   id,created_by,comments
   <existing id>,<any username>,unchanged
   ```

2. The existing entry's `created_by` is overwritten with the supplied user.

`created_by` (`extras/api/serializers_/journaling.py`) is a writable `PrimaryKeyRelatedField(queryset=User.objects.all(), default=CurrentUserDefault())`, and `JournalEntryImportForm` (`extras/forms/bulk_import.py`) includes `created_by` in its fields. Neither path restricts changing the value on update; `CurrentUserDefault` only populates on create.

### Expected Behavior

`created_by` is set from the requesting user at creation and is immutable thereafter. Setting it at create time stays supported (this is intentional, added in #6850 so API clients and imports can self-attribute and preserve authorship when migrating historical entries). Updates should not be able to change it.

### Observed Behavior

Both the REST API update path and the bulk-import update path accept a client-supplied `created_by` on an existing entry and overwrite the recorded author, allowing authorship to be reassigned after the fact.

Note: the `ObjectChange` changelog still records the actual user who performed the edit, so this does not forge the underlying audit log; it falsifies the journal entry's own displayed authorship. The single-object UI edit view and the bulk-edit form are not affected (they force or omit `created_by`).

mburggraf:
Closes #22513
Made the field created_by immutable on model level.
This prevents the REST API from changing the field, as well as the import form


pheus:
@mburggraf Thanks for opening this PR!

Could you please update the PR description to match the required PR template? For this PR, that should include:

```markdown
### Closes: #1234

<!--
    Please include a summary of the proposed changes below.
-->

```


This helps keep issue linking consistent for review and merging. Thanks!

mburggraf:
> @mburggraf Thanks for opening this PR!
> 
> Could you please update the PR description to match the required PR template? For this PR, that should include:
> 
> ```
> ### Closes: #1234
> 
> 
> ```
> 
> This helps keep issue linking consistent for review and merging. Thanks!

Sorry, didn't notice, that the linking doesn't work on the subject.

arthanson (changes requested):
Thanks for the work on this, but a few changes needed.

This returns a 500 error if you do an API call. The guard raises `django.core.exceptions.ValidationError` from save(). On the API path, perform_update() (`netbox/netbox/api/viewsets/__init__.py:301`) calls serializer.save() → instance.save(). A Django core ValidationError raised there is not translated by DRF's exception handler (which only handles DRF's APIException/Http404/PermissionDenied), so it propagates as a 500 Internal Server Error.  I did a call like:

```
curl -sS -X PATCH "$NETBOX_URL/api/extras/journal-entries/1/" \
  -H "Authorization: Token $NETBOX_TOKEN" \
  -d '{"created_by": 1}' \
  -w '\nHTTP %{http_code}\n'
```

This also increases the query count, if you compare the id instead, something like:
```
if self.created_by_id != self.__original_created_by_id:
```
It won't have the extra query.

Also, there should be a test added for this.

mburggraf:
I changed the serializer, so there won't be a 500 Internal Server Error anymore. A Testcase was added as well and the comparision now works with the id, to avoid unnecessary queries.

arthanson (changes requested):
Thanks for the revisions, @mburggraf - the 400 on the API path and the test are good improvements. Before this goes in did you consider the approach in #20576, as the model-level guard reintroduces a couple of issues.

This is the same bug that #20576 fixed for Token.user, and I think it's worth mirroring that fix directly rather than guarding at the model level. It's lighter and avoids two problems the current version has:

1. PUT requests fail even when they don't touch created_by. Since the field has default=CurrentUserDefault(), a PUT that omits it gets created_by filled in as request.user - so any user editing someone else's entry trips the "immutable" error without ever sending the field. 

(PATCH is unaffected; defaults don't apply to partial updates.) Making the field read-only sidesteps this entirely - read-only fields are dropped before the default is applied.

2. The list query count went 15 to 16. The __init__ override reads self.created_by_id on every instantiation, pinning that column into queries that would otherwise defer it (the brief serializer doesn't include created_by).

Mirroring #20576 — serializer (tokens.py:42-49) something like:

```
  def get_fields(self):
      fields = super().get_fields()
      # Make created_by read-only when updating an existing entry.
      if self.instance is not None:
          fields['created_by'].read_only = True
      return fields
```

and the import form (users/forms/bulk_import.py:57-63):

```
      if self.instance.pk:
          self.fields['created_by'].disabled = True
```

I think this covers both vectors in #22513 and lets you drop the __init__ and save() overrides on the model entirely.

mburggraf:
@arthanson I applied your suggestion, so that the changes now reflect the changes done in #20576 for Tokens.
Just pointing out that before, the user got an error message when trying to change the `created_by`, while now any changes are just silently discarded.

arthanson (approved):
Thanks @mburggraf !

PR #22549, opened 2026-06-25
Closes #20547: Consolidate unique constraints comprising nullable fields

jeremystretch:
### Closes: #20547

- Replace all pairs of "redundant" unique constraints with a single UniqueConstraint passing `nulls_distinct=False`.
- Fix a uniqueness enforcement gap on ContactGroup by passing `nulls_distinct=False` on the single existing constraint.
- Remove an extraneous UniqueConstraint on WirelessLAN Group (both its `name` and `slug` have `unique=True`).

### DIFF
--- a/netbox/vpn/models/tunnels.py
+++ b/netbox/vpn/models/tunnels.py
@@ -1,7 +1,6 @@
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import Q
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
@@ -78,17 +77,6 @@ class Tunnel(ContactsMixin, PrimaryModel):
 
     class Meta:
         ordering = ('name',)
-        constraints = (
-            models.UniqueConstraint(
-                fields=('group', 'name'),
-                name='%(app_label)s_%(class)s_group_name'
-            ),
-            models.UniqueConstraint(
-                fields=('name',),
-                name='%(app_label)s_%(class)s_name',
-                condition=Q(group__isnull=True)
-            ),
-        )
         verbose_name = _('tunnel')
         verbose_name_plural = _('tunnels')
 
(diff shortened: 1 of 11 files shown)


jeremystretch:

jeremystretch:

jeremystretch:

netbox/wireless/models.py
```diff
@@ -70,12 +70,6 @@ class Meta:
             GistIndex(fields=['path'], name='wireless_lan_grp_path_gist'),
             models.Index(fields=['sort_path'], name='wireless_lan_grp_sort_idx'),
         )
-        constraints = (
```
jeremystretch, line 73: We're dropping this constraint because `name` and `slug` are both required to be globally unique: The per-parent constraint is redundant.

jeremystretch:

jeremystretch:
The scope of #20547 is limited to consolidating _existing_ uniqueness constraints without modifying any existing behavior. Altering the UniqueConstraint in place for the ContactGroup model by adding `nulls_distinct=False` would result in a material change: The migration would fail on the presence of any ContactGroups with the same name and no parent assigned.

This is worth revisiting separately, but needs to be handled as a potentially breaking change.

pheus (changes requested):
Thanks for this cleanup!

I noticed one small thing:
In `vpn.Tunnel`, `name` is still globally unique, so the new `(group, name)` constraint cannot currently have any effect beyond the existing field-level uniqueness. Was the intent to drop `unique=True` from `name` and make tunnel names unique per group instead?

Other than that, the consolidation looks good to me.

jeremystretch:
We can't make any material changes to uniqueness enforcement in this PR, so it probably makes sense to simply drop the extraneous constraint for Tunnel.

netbox/vpn/models/tunnels.py
```diff
@@ -78,17 +77,6 @@ class Tunnel(ContactsMixin, PrimaryModel):
 
     class Meta:
         ordering = ('name',)
-        constraints = (
```
jeremystretch, line 81: Dropping this one too, as `name` is already unique.

pheus (approved):
Thanks! Nice!

PR #22550, opened 2026-06-25
Fixes #22521: Respect RAM_BASE_UNIT for VirtualMachineType

### LINKED ISSUE #22521: Virtual Machine Type: Use RAM_BASE_UNIT for default memory
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

- Set RAM_BASE_UNIT = 1024 in configuration and restart netbox
- create new Virtual Machine Type (VMT) with 1000 MB (notice it should mention MiB here...)
- Create new VM from this Virtual Machine Type and set default Memory.

### Expected Behavior

If set "RAM_BASE_UNIT = 1024" the VMT should display MiB

### Observed Behavior

- Compare VM and VMT. One shows 1000MB and the other 1000MiB. This is a mismatch.

mburggraf:
Closes #22521: Make VirtualMachineType respect the RAM_BASE_UNIT setting.

Note: The translation files seem to contain a translation for the string 'Default Memory (MB)', but not for 'Default Memory (MiB)'. So translations need to be updated as well.

### DIFF
--- a/netbox/virtualization/forms/bulk_edit.py
+++ b/netbox/virtualization/forms/bulk_edit.py
@@ -93,6 +93,9 @@ class VirtualMachineTypeBulkEditForm(PrimaryModelBulkEditForm):
     default_memory = forms.IntegerField(
         label=_('Default Memory (MB)'),
         required=False,
+    default_memory = forms.IntegerField(
+        label=_('Default memory)'),
+        required=False,
     )
 
     model = VirtualMachineType
@@ -108,7 +111,7 @@ class VirtualMachineTypeBulkEditForm(PrimaryModelBulkEditForm):
         super().__init__(*args, **kwargs)
 
         # Set unit label based on configured RAM_BASE_UNIT (MB vs MiB)
-        self.fields['default_memory'].label = _('Default Memory ({unit})').format(
+        self.fields['default_memory'].label = _('Default memory ({unit})').format(
             unit=get_capacity_unit_label(settings.RAM_BASE_UNIT)
         )
 
--- a/netbox/virtualization/forms/bulk_edit.py
+++ b/netbox/virtualization/forms/bulk_edit.py
@@ -104,6 +104,14 @@ class VirtualMachineTypeBulkEditForm(PrimaryModelBulkEditForm):
         'default_platform', 'default_vcpus', 'default_memory', 'description', 'comments',
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Set unit label based on configured RAM_BASE_UNIT (MB vs MiB)
+        self.fields['default_memory'].label = _('Default Memory ({unit})').format(
+            unit=get_capacity_unit_label(settings.RAM_BASE_UNIT)
+        )
+
 
 class VirtualMachineBulkEditForm(PrimaryModelBulkEditForm):
     virtual_machine_type = DynamicModelChoiceField(
--- a/netbox/virtualization/forms/bulk_edit.py
+++ b/netbox/virtualization/forms/bulk_edit.py
@@ -91,7 +91,7 @@ class VirtualMachineTypeBulkEditForm(PrimaryModelBulkEditForm):
         required=False,
     )
     default_memory = forms.IntegerField(
-        label=_('Default memory)'),
+        label=_('Default memory'),
         required=False,
     )
 
--- a/netbox/virtualization/forms/bulk_edit.py
+++ b/netbox/virtualization/forms/bulk_edit.py
@@ -91,9 +91,6 @@ class VirtualMachineTypeBulkEditForm(PrimaryModelBulkEditForm):
         required=False,
     )
     default_memory = forms.IntegerField(
-        label=_('Default Memory (MB)'),
-        required=False,
-    default_memory = forms.IntegerField(
         label=_('Default memory)'),
         required=False,
     )
--- a/netbox/virtualization/forms/filtersets.py
+++ b/netbox/virtualization/forms/filtersets.py
@@ -128,7 +128,9 @@ class VirtualMachineTypeFilterForm(PrimaryModelFilterSetForm):
         required=False,
     )
     default_memory = forms.IntegerField(
-        label=_(f'Default memory ({get_capacity_unit_label(settings.RAM_BASE_UNIT)})'),
+        label=_('Default memory ({unit})').format(
+            unit=get_capacity_unit_label(settings.RAM_BASE_UNIT)
+        ),
         required=False,
         min_value=0,
     )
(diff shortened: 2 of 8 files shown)


pheus (changes requested):
@mburggraf Thanks for opening this PR!

Could you please update the PR description to match the PR template? In particular, please include the `Closes: #22521` line and a short summary of the proposed changes.

```markdown
### Closes: #1234

<!--
    Please include a summary of the proposed changes below.
-->

```

I also noticed a couple of implementation details that we should adjust before merging:

1. Please avoid using f-strings inside translated strings. With gettext, the f-string is evaluated before the string reaches the translation machinery, so this will not work as expected for i18n.

   Instead of:

   ```python
   _(f'Default memory ({get_capacity_unit_label(settings.RAM_BASE_UNIT)})')
   ```

   please use a static translatable string with a placeholder, for example:

   ```python
   _('Default memory ({unit})').format(
       unit=get_capacity_unit_label(settings.RAM_BASE_UNIT)
   )
   ```

2. I think we should align this with the existing `VirtualMachine` / `VirtualDisk` handling from #21789 rather than patching individual labels only. That change made the model `verbose_name`s unit-agnostic and then applied the configured unit at the form/display layer.

   For `VirtualMachineType.default_memory`, I think that would mean:
   - making the model `verbose_name` unit-agnostic, e.g. `default memory`
   - setting the dynamic `RAM_BASE_UNIT` label in the model form, bulk edit form, and filter form
   - making the table/object display follow the same display pattern as VM memory
   - adding/adjusting tests for both `RAM_BASE_UNIT = 1000` and `RAM_BASE_UNIT = 1024`

This should keep VirtualMachineType consistent with the rest of the virtualization app and avoid leaving any hard-coded `(MB)` metadata behind.

Thanks again!

mburggraf:
I removed the hardcoded MB reference and the f-String translation.
It now looks the same like in the `VirtualMachine` views.

pheus (changes requested):
@mburggraf Thanks for the update! This is definitely closer, and I appreciate you iterating on it!

The form and table changes now look much more aligned with the existing `VirtualMachine` handling. I think the remaining part is the object/detail display: it still formats `default_memory` manually and also still has an f-string inside a translated string.

Could we make this follow the same humanized RAM display pattern used by `VirtualMachine` instead of appending the configured unit directly? That should keep the behavior consistent across the virtualization app and avoid the i18n issue at the same time.

Small consistency nits: I’d also suggest moving the filter-form label setup into `__init__`, and keeping the model `verbose_name` lowercase as `default memory`.

Thanks again! This is close!

mburggraf:
@pheus 
I removed the translation. Come to think of it there is absolutely no use in calling the translation function, since the string only contains a value and the unit… there is nothing to translate in it at all.

The filter-form label setup was moved into `__init__`, which is now consistent with the other files, even though it's more code.

About the `verbose_name` in the model: The reason I went for a mixed case was, that the translation didn't work on the all-lowercase version.

The detail display for the `VirtualMachine` is designed quite different, putting the CPUs, memory and Disk values into a separate 'Ressources'-Block. Not sure if this is needed for `VirtualMachineType` as well.

netbox/virtualization/forms/bulk_edit.py
```diff
@@ -104,6 +104,14 @@ class VirtualMachineTypeBulkEditForm(PrimaryModelBulkEditForm):
         'default_platform', 'default_vcpus', 'default_memory', 'description', 'comments',
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Set unit label based on configured RAM_BASE_UNIT (MB vs MiB)
+        self.fields['default_memory'].label = _('Default Memory ({unit})').format(
```
pheus:
```suggestion
        self.fields['default_memory'].label = _('Default memory ({unit})').format(
```

netbox/virtualization/forms/filtersets.py
```diff
@@ -140,6 +140,14 @@ class VirtualMachineTypeFilterForm(PrimaryModelFilterSetForm):
 
     tag = TagFilterField(model)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Set unit label based on configured RAM_BASE_UNIT (MB vs MiB)
+        self.fields['default_memory'].label = _('Default Memory ({unit})').format(
```
pheus:
```suggestion
        self.fields['default_memory'].label = _('Default memory ({unit})').format(
```

netbox/virtualization/forms/model_forms.py
```diff
@@ -192,6 +192,14 @@ class Meta:
             'owner', 'comments', 'tags',
         )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Set unit label based on configured RAM_BASE_UNIT (MB vs MiB)
+        self.fields['default_memory'].label = _('Default Memory ({unit})').format(
```
pheus:
```suggestion
        self.fields['default_memory'].label = _('Default memory ({unit})').format(
```

netbox/virtualization/forms/bulk_edit.py
pheus:
```suggestion
    default_memory = forms.IntegerField(
        label=_('Default memory)'),
        required=False,
    )
```

pheus (changes requested):
@mburggraf Thanks for the follow-up and for addressing the previous comments. This is looking much better!

I think there is one consistency issue left: the object/detail panel still appends the configured base unit to the **raw stored value**, while `VirtualMachine` displays memory using the **humanized RAM helper**.
Could we make the object panel use the same humanized display as well? That would avoid showing something like `2048 MiB` in one place and `2.00 GiB` in another.

Small nit while touching this: could we also keep the label casing consistent as `Default memory ({unit})` / `Default memory` rather than `Default Memory`?

Thanks again. this is very close!

pheus (changes requested):
@mburggraf Thanks for the update!

I think this version is broader than needed, though. Adding `format_callable` to `TextAttr` changes the generic UI attribute API for one field, while we already have a local pattern for this.

Please use the same approach as `VirtualDiskPanel.size` instead:

```python
# netbox/virtualization/ui/panels.py

default_memory = attrs.TemplatedAttr(
    'default_memory',
    template_name='virtualization/virtualmachinetype/attrs/default_memory.html',
    label=_('Default memory'),
)
```

and add the matching template here:

```django
{% load helpers %}

{{ value|humanize_ram_capacity }}
```

That keeps this scoped to `VirtualMachineType.default_memory` and avoids changing the generic `TextAttr` API.

Small cleanups:
- please drop the `TextAttr` / `format_callable` change from `netbox/netbox/ui/attrs.py`
- please fix the bulk edit label typo in `netbox/virtualization/forms/bulk_edit.py`: `Default memory)`
- please keep the model `verbose_name` lowercase in `netbox/virtualization/models/virtualmachines.py`: `default memory`

Thanks for sticking with this!

mburggraf:
@pheus: I applied the requested changes.

Just pointing out: Translation doesn't currently work for the lowercase name in the model.

pheus (approved):
@mburggraf Thanks for sticking with this and for working through the review comments! This looks good to me now.

PR #22552, opened 2026-06-26
Fixes #22501: Fix GraphQL API exceptions falling back to HTML error responses

### LINKED ISSUE #22501: GraphQL 500 responses render HTML instead of JSON
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

1. Run NetBox with `DEBUG = False`.
2. Issue a GraphQL query that triggers an unhandled server-side exception (any 500-class error raised within a resolver that Strawberry doesn't catch).

Reconstructed request:

```
POST /graphql/ HTTP/1.1
Host: netbox.example.com
Authorization: Token <token>

{"query": "query { ... }"}
```

### Expected Behavior

A JSON response, consistent with how the REST API handles the same class of failure. A GraphQL client expects JSON, not HTML.

### Observed Behavior

`CoreMiddleware.process_exception()` (`netbox/netbox/middleware.py:106`) routes exceptions for REST API requests through `handle_rest_api_exception()`, but only when `is_api_request(request)` is true:

```python
if is_api_request(request):
    got_request_exception.send(sender=self.__class__, request=request)
    return handle_rest_api_exception(request)
```

`is_api_request()` (`netbox/utilities/api.py`) matches only paths under the REST API root. GraphQL requests hit `/graphql/` and fail this check, despite `is_graphql_request()` already existing in the same module. For a GraphQL request, `process_exception()` returns `None`, so the request falls through to Django's default 500 handler and returns `text/html` to a JSON client.

Confirmed by running both paths through the middleware with `DEBUG = False` and the same exception in flight:

```
GraphQL request (/graphql/):
  is_api_request:     False
  is_graphql_request: True
  process_exception -> None (falls through to Django HTML 500 handler)

REST API request (/api/):
  is_api_request:     True
  process_exception -> application/json status 500
```

A second consequence: GraphQL exceptions never reach the `got_request_exception` signal sent on the REST API path, so error-tracking integrations (e.g. Sentry) don't capture them. This is the same class of gap addressed for the REST API in netbox-community/netbox#22055.

pheus:
### Closes: #22501

This updates NetBox’s exception handling so JSON GraphQL requests are handled the same way as REST API requests when `DEBUG=False`.

Previously, exceptions raised while processing `/graphql/` requests could fall through to Django’s default HTML 500 response handling. With this change, GraphQL API exceptions now use NetBox’s existing JSON API error response path and emit `got_request_exception`, ensuring error-tracking integrations such as Sentry can capture them consistently.

The change is intentionally narrow: REST API behavior is preserved, GraphiQL/HTML requests are left unchanged, and the existing API error handler is reused for GraphQL JSON requests.

Tests have been added to cover both the existing REST API behavior and the new GraphQL API exception handling.

### DIFF
--- a/netbox/netbox/middleware.py
+++ b/netbox/netbox/middleware.py
@@ -105,8 +105,8 @@ class CoreMiddleware:
         if settings.DEBUG:
             return None
 
-        # Cleanly handle exceptions that occur from REST API requests
-        if is_api_request(request):
+        # Cleanly handle exceptions that occur from REST or GraphQL API requests
+        if is_api_request(request) or is_graphql_request(request):
             # Fire Django's got_request_exception signal so error-tracking
             # integrations (e.g. Sentry) capture the exception.
             got_request_exception.send(sender=self.__class__, request=request)
@@ -283,7 +283,7 @@ class MaintenanceModeMiddleware:
             error_message = 'NetBox is currently operating in maintenance mode and is unable to perform write ' \
                             'operations. Please try again later.'
 
-            if is_api_request(request):
+            if is_api_request(request) or is_graphql_request(request):
                 return handle_rest_api_exception(request, error=error_message)
 
             messages.error(request, error_message)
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
@claude /review
I've addressed your points.

jnovinger (approved):

PR #22554, opened 2026-06-26
Fixes #22543: Prevent duplicate Script Module uploads via REST API from corrupting existing files

### LINKED ISSUE #22543: Re-uploading a same-named custom script via REST deletes the existing script's file
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

1. Upload a new script module via the API:
```
  curl -k -X POST https://<netbox>/api/extras/scripts/upload/ \
    -H "Authorization: Token <token>" \
    -F "file=@zz_probe.py"
```
zz_probe.py:
```

  class ProbeScript(Script):
      def run(self, data, commit):
          return "v1"
```
2. Upload again with the same filename but different content (e.g. change the return string to "v2"):
```
  curl -k -X POST https://<netbox>/api/extras/scripts/upload/ \
    -H "Authorization: Token <token>" \
    -F "file=@zz_probe.py"
```
2. → 400 Bad Request — ["A script module with this file name already exists."]

### Expected Behavior

Receive the error but the original file should still be there.

### Observed Behavior

On disk, SCRIPTS_ROOT/zz_probe.py has been deleted, even though the original ScriptModule DB record still exists and points at it.

- The script list still reports is_executable: true for a short time (registry caching masks the breakage).
- After the next registry refresh, run the original script (POST /api/extras/scripts/<id>/) → HTTP 500: "'NoneType' object has no attribute 'job_timeout'" (its python_class is now None because the backing file is gone).

pheus:
### Closes: #22543

This updates script module uploads to reject duplicate filenames before writing the uploaded file to storage. Previously, a failed duplicate upload could overwrite or delete the existing script file while leaving its `ScriptModule` record behind.

The cleanup path now also checks whether a file is still referenced before deleting it, which helps avoid removing files created or retained by another successful upload.

### DIFF
--- a/netbox/extras/api/serializers_/scripts.py
+++ b/netbox/extras/api/serializers_/scripts.py
@@ -37,6 +37,13 @@ class ScriptModuleSerializer(ValidatedModelSerializer):
         # Pop 'file' before model instantiation — ScriptModule has no such field.
         file = data.pop('file', None)
         data['file_root'] = ManagedFileRootPathChoices.SCRIPTS
+
+        # Reject duplicates before writing to storage so a failed upload can't touch the existing file
+        if file is not None and ScriptModule.objects.filter(
+            file_root=ManagedFileRootPathChoices.SCRIPTS, file_path=file.name
+        ).exists():
+            raise serializers.ValidationError(_("A script module with this file name already exists."))
+
         data = super().validate(data)
         data.pop('file_root', None)
         if file is not None:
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
I agree with both points.

This PR is focused on the issue reported here: a normal duplicate upload through the REST API should be rejected before it touches storage. The concurrent upload race is a broader pre-existing problem and probably needs a more involved fix.

For `file.name`, my understanding is that the default scripts storage keeps the same path because it uses `allow_overwrite=True`. The existing `IntegrityError` handling should still cover unexpected/custom storage behavior.

arthanson (approved):

PR #22556, opened 2026-06-26
Closes #22526: Avoid query timeouts when updating custom fields on a large number of objects

### LINKED ISSUE #22526: Chunk bulk updates to custom field data to better handle a large number of records
### NetBox version

v4.6.3

### Feature type

### Proposed functionality

When a custom field is created or assigned to one or more object types, NetBox synchronously populates the field's default (or JSON null) into the custom_field_data of every existing object of those types. 

Offload this population to a background job (JobRunner) rather than running it inline in the request. Ideally the update should also be chunked (batched by PK) so each statement touches a bounded number of rows. The same applies to the sibling operations remove_stale_data() (field unassigned) and rename_object_data() (field renamed), which use the identical unbatched pattern.

This happens in CustomField.populate_initial_data() (extras/models/customfields.py), triggered by the post_add m2m signal in extras/signals.py, and runs inside the request's transaction.atomic() block as a single unbounded UPDATE ... jsonb_set(...) per content type.

```
netbox/extras/models/customfields.py:325
def populate_initial_data(self, content_types):
    ...
    for ct in content_types:
        ct.model_class().objects.update(
            custom_field_data=Func(
                F('custom_field_data'),
                Value([self.name]),
                value,
                function='jsonb_set'
            )
        )
```

### Use case

On large installations, the affected models can have millions of rows. PostgreSQL must rewrite every row (JSONB updates are not in-place), so the single full-table UPDATE frequently exceeds statement_timeout and the operation fails with canceling statement due to statement timeout. 

Moving the work to a chunked background job would let the custom field be created immediately, keep individual statements within timeout limits, reduce lock contention, and give users visibility into progress/failure via the existing jobs UI.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #22526

- Set `CUSTOMFIELD_DATA_BATCH_SIZE` to 5000.
- Introduce the `_update_object_data()` private method on CustomField to apply bulk data changes in chunks.
- Add a test

### DIFF
--- a/netbox/extras/constants.py
+++ b/netbox/extras/constants.py
@@ -6,6 +6,14 @@ from extras.choices import LogLevelChoices
 # Custom fields
 CUSTOMFIELD_EMPTY_VALUES = (None, '', [])
 
+# Maximum number of objects to update per query when provisioning, removing, or renaming custom
+# field data. Bounding the number of rows touched by each statement prevents very large tables from
+# exceeding the database statement timeout (JSONB updates rewrite each affected row). This value
+# sits at the throughput "knee": benchmarking jsonb_set() across a 1M-row table showed throughput
+# plateaus by ~5K rows/statement (raising it further yields no meaningful speedup), while keeping
+# each statement orders of magnitude below a typical statement timeout.
+CUSTOMFIELD_DATA_BATCH_SIZE = 5000
+
 # ImageAttachment
 IMAGE_ATTACHMENT_IMAGE_FORMATS = {
     'avif': 'image/avif',
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -8,7 +8,7 @@ import jsonschema
 from django import forms
 from django.conf import settings
 from django.core.validators import RegexValidator, ValidationError
-from django.db import models
+from django.db import models, transaction
 from django.db.models import F, Func, Value
 from django.db.models.expressions import RawSQL
 from django.urls import reverse
@@ -19,6 +19,7 @@ from jsonschema.exceptions import ValidationError as JSONValidationError
 
 from core.models import ObjectType
 from extras.choices import *
+from extras.constants import CUSTOMFIELD_DATA_BATCH_SIZE
 from extras.data import CHOICE_SETS
 from extras.fields import ChoiceSetField
 from netbox.context import query_cache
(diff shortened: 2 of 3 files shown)


jeremystretch:

jeremystretch:
Here are some benchmark results using different values for `CUSTOMFIELD_DATA_BATCH_SIZE` (1M rows, PostgreSQL 16.14, adding one key to `custom_field_data` on every row via the exact keyset-batched `jsonb_set` pattern):

| batch      | statements | total wall | rows/sec | % of max | p50 stmt | p95 stmt | max stmt  |
| ---------: | ---------: | ---------: | -------: | -------: | -------: | -------: | --------: |
| 100        | 10,000     | 33.0 s     | 30,314   | 35%      | 2.8 ms   | 5.5 ms   | 57 ms     |
| 500        | 2,000      | 16.8 s     | 59,508   | 69%      | 7.6 ms   | 11.3 ms  | 67 ms     |
| **1,000**  | 1,000      | 14.5 s     | 69,064   | 80%      | 13.5 ms  | 16.4 ms  | 87 ms     |
| 2,500      | 400        | 13.7 s     | 73,098   | 84%      | 31.7 ms  | 38.1 ms  | 179 ms    |
| **5,000**  | 200        | 13.3 s     | 75,085   | 87%      | 62.1 ms  | 72.4 ms  | 216 ms    |
| 10,000     | 100        | 13.2 s     | 75,551   | 87%      | 122 ms   | 169 ms   | 381 ms    |
| 25,000     | 40         | 12.6 s     | 79,168   | 91%      | 301 ms   | 398 ms   | 436 ms    |
| 50,000     | 20         | 12.8 s     | 78,005   | 90%      | 603 ms   | 777 ms   | 782 ms    |
| 100,000    | 10         | 12.6 s     | 79,095   | 91%      | 1,204 ms | 1,342 ms | 1,342 ms  |
| **SINGLE** | 1          | 11.5 s     | 86,630   | 100%     | —        | —        | 11,543 ms |

## What the data says

**1. The knee is sharp and early — around 2,500–5,000 rows.** Throughput nearly doubles from 100→500, gains +16% to 1,000, then flattens hard: +5.8% to 2,500, +2.7% to 5,000, and essentially **nothing** (<1–5%, within noise) beyond that. By 5,000 you're at 87% of the theoretical max (the single unbounded query). Past 5,000 the curve is flat — bigger batches buy no throughput.

**2. Statement duration is dead-linear in batch size: ~12 µs/row** here (p50 holds at 12–13.5 µs/row from 1k all the way to 100k). This is the number that matters for the timeout, and it's the knob: duration = batch_size × per-row-cost.

**3. The binding constraint is *far* from the knee.** Your 5-min (300 s) `statement_timeout`, even budgeting a conservative 25% (75 s), allows ~6 *million* rows per statement at this per-row cost. The throughput knee (~5k) sits **three orders of magnitude** below that. So on any sane hardware the batch size should be chosen at the *throughput knee*, not derived from the timeout — the timeout headroom is automatically gigantic there.

**4. Why 1,000 is indeed a touch conservative:** it leaves ~20% throughput on the table (69k vs 75–79k rows/sec) and runs 5× more statements than necessary, for no safety benefit.

## Recommendation: bump the default to **5,000**

It sits right on the plateau (87% of max throughput, ≈15% slower wall-clock than a single query), with a p95 statement of ~72 ms locally. Even assuming production is **20× slower** per row (network-attached storage, concurrent load, wider rows, more indexes), that's ~1.4 s per statement — still ~200× under a 5-min timeout. Huge margin, near-optimal throughput, and 1/5 the round-trips of 1,000.

**Caveats (all of which make 5,000 *safer*, not riskier):** this box is local NVMe, single-client, no contention; synthetic rows are narrower than real NetBox rows; per-row cost on production will be higher. A higher per-row cost pushes the throughput knee *lower* (fixed overhead amortizes sooner), so ~5k stays optimal while the timeout margin only grows. The case for still exposing it as a setting holds: a pathological environment (very tight timeout + very slow rows) can dial it down, but the default belongs at the knee.

jeremystretch:
Bumped the value of `CUSTOMFIELD_DATA_BATCH_SIZE` from 1000 to 5000 and addressed the concern regarding atomicity noted above.

jnovinger (approved):

PR #22557, opened 2026-06-26
Closes #22548: Improve sidebar action button styling

### LINKED ISSUE #22548: Tweak styling of action buttons in navigation menu
### NetBox version

v4.6.3

### Feature type

Other

### Proposed functionality

Change the styling of the action buttons alongside navigation menu items to be more subtle.

Before:

Image

After:

Image

This is accomplished by applying the `btn-secondary` and `btn-ghost` classes to each button in the `menu.html` template.

Note that we'll also want to clean up the spacing and sizing a bit. There seems to be a roughly half-pixel overflow beneath the buttons (barely visible in the second screenshot above).

### Use case

IMO the borderless buttons blend in much more nicely with the navigation links.

### Database changes

N/A

### External dependencies

N/A

pheus:
### Closes: #22548

This updates the sidebar menu action buttons to use the subtler `btn-ghost-secondary` variant and tightens their line-height to reduce extra vertical spacing.

It also vertically centers dropdown item contents and scopes the dropdown link color rules to direct child anchors only. This prevents the menu link styles from cascading into nested action buttons, allowing those buttons to keep their intended hover and focus contrast.

Light (add :hover):
Screenshot 2026-06-26 at 18-57-30 L2VPN Terminations NetBox

Dark (add :hover):
Screenshot 2026-06-26 at 18-56-52 L2VPN Terminations NetBox


pheus:

pheus:
@claude /review
I've addressed your point.

netbox/utilities/templates/navigation/menu.html
```diff
@@ -40,12 +40,12 @@
...
               </div>
               {% for item, buttons in items %}
-                <div class="dropdown-item d-flex justify-content-between ps-3 py-0">
+                <div class="dropdown-item d-flex align-items-center justify-content-between ps-3 py-0">
                   <a href="{{ item.url }}" class="d-inline-flex flex-fill py-1">{{ item.link_text }}</a>
                   {% if buttons %}
                     <div class="btn-group ms-1">
                       {% for button in buttons %}
-                        <a href="{{ button.url }}" class="btn btn-sm btn-{{ button.color|default:"outline" }} lh-2 px-2" title="{{ button.title }}" aria-label="{{ button.title }}">
+                        <a href="{{ button.url }}" class="btn btn-sm btn-ghost-secondary lh-1 px-2" title="{{ button.title }}" aria-label="{{ button.title }}">
```
jeremystretch: Can we preserve color support? Tabler supports the "ghost" style for the primary theme colors:

Image

jeremystretch (changes requested):

jeremystretch (changes requested):
There's a very faint hint of shadow from the parent `btn-group`. We can set `box-shadow: none` on the `.btn-group` to remove it for a nice, clean display. (This should probably be set irrespective of the visibility toggle.)

Actually, it might make sense to remove the `.btn-group` entirely: There's no point in grouping the buttons anymore with the ghost effect. The partially-rounded corners on either button independently look weird IMO. (We could keep `.btn-group` but remove the radius on the child elements.)

pheus:
Thanks for catching that!

I’ve adjusted the approach and dropped the `.btn-group` wrapper as suggested, since these buttons no longer need grouped button styling. The wrapper now uses a small semantic `dropdown-item-buttons` class for the sidebar visibility behavior, while each action keeps the regular `btn-sm btn-ghost btn-{{ button.color|default:"secondary" }}` styling so existing color support is preserved.

I also kept the direct child selector change for the menu item link styles so they don’t cascade into the nested action buttons.

jeremystretch (approved):

PR #22562, opened 2026-06-29
Fixes #22500: Use passed error kwarg in handle_rest_api_exception()

### LINKED ISSUE #22500: Maintenance mode message discarded for REST API requests
### Proposed Changes

`handle_rest_api_exception()` (`netbox/utilities/error_handlers.py:48`) accepts `*args, **kwargs` but never reads them. Its body builds the response solely from `sys.exc_info()`:

```python
def handle_rest_api_exception(request, *args, **kwargs):
    type_, error = sys.exc_info()[:2]
    data = {
        'error': str(error),
        ...
    }
```

`MaintenanceModeMiddleware.process_exception()` (`netbox/netbox/middleware.py:284`) calls it with a custom message that is silently dropped:

```python
return handle_rest_api_exception(request, error=error_message)
```

Update `handle_rest_api_exception()` to use a passed `error` message when provided, falling back to `sys.exc_info()` otherwise.

### Justification

When NetBox is in maintenance mode and an API client triggers a write, the middleware intends to return "NetBox is currently operating in maintenance mode and is unable to perform write operations. Please try again later." Instead the client receives the raw database error, because the message argument is discarded.

Confirmed by exercising the call path directly. With an `InternalError` in flight, `handle_rest_api_exception(request, error=<maintenance message>)` returns:

```json
{
  "error": "cannot execute UPDATE in a read-only transaction",
  "exception": "InternalError",
  "netbox_version": "4.6.2",
  "python_version": "3.12.12"
}
```

The intended maintenance message never appears. The UI path returns it (via `messages.error()`); the API path does not. This aligns the two.

g-builder-0:
### Closes: #22500

`handle_rest_api_exception()` accepted `**kwargs` but never read them, silently 
discarding the `error` kwarg passed by `MaintenanceModeMiddleware.process_exception()`. 
API clients received the raw database error instead of the human-readable maintenance 
mode message.

The fix checks for an `error` kwarg first and uses it when present, falling back to 
`sys.exc_info()` otherwise. Also adds `test_error_handlers.py` with two tests covering 
both code paths.

### DIFF
--- a/netbox/utilities/error_handlers.py
+++ b/netbox/utilities/error_handlers.py
@@ -28,9 +28,9 @@ def handle_protectederror(obj_list, request, e):
         raise e
 
     # Formulate the error message
-    err_message = _("Unable to delete <strong>{objects}</strong>. {count} dependent objects were found: ").format(
+    err_message = _('Unable to delete <strong>{objects}</strong>. {count} dependent objects were found: ').format(
         objects=', '.join(escape(obj) for obj in obj_list),
-        count=len(protected_objects) if len(protected_objects) <= 50 else _('More than 50')
+        count=len(protected_objects) if len(protected_objects) <= 50 else _('More than 50'),
     )
 
     # Append dependent objects to error message
@@ -49,10 +49,17 @@ def handle_rest_api_exception(request, *args, **kwargs):
     """
     Handle exceptions and return a useful error message for REST API requests.
     """
-    type_, error = sys.exc_info()[:2]
+    if 'error' in kwargs:
+        error_message = str(kwargs['error'])
+        type_, _ = sys.exc_info()[:2]
+        exception_name = type_.__name__ if type_ else 'Exception'
+    else:
+        type_, error = sys.exc_info()[:2]
+        error_message = str(error)
+        exception_name = type_.__name__
     data = {
-        'error': str(error),
-        'exception': type_.__name__,
+        'error': error_message,
+        'exception': exception_name,
         'netbox_version': settings.RELEASE.full_version,
         'python_version': platform.python_version(),
     }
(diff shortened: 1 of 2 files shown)


netbox/utilities/error_handlers.py
```diff
@@ -49,10 +49,17 @@ def handle_rest_api_exception(request, *args, **kwargs):
     """
     Handle exceptions and return a useful error message for REST API requests.
     """
-    type_, error = sys.exc_info()[:2]
+    if 'error' in kwargs:
+        error_message = str(kwargs['error'])
+        type_, _ = sys.exc_info()[:2]
```
jnovinger: Let's be careful to not use `_` here as the temporary, unnamed variable, as it shadows the `gettext_lazy` import on line 10 above.

jnovinger (changes requested):

netbox/utilities/error_handlers.py
g-builder-0: Ok. I see the problem. I will replace `_` with a different temporary variable.

g-builder-0: @jnovinger Updated version of error_handlers.py is ready for review.

g-builder-0: @jnovinger Would you like me to make any other changes to the Python? Looks like the current testing failure involves [Javascript-Typescript + CodeQL](https://github.com/netbox-community/netbox/actions/runs/28465078901/job/84363744333?pr=22562).

jnovinger:
> @jnovinger Would you like me to make any other changes to the Python? Looks like the current testing failure involves [Javascript-Typescript + CodeQL](https://github.com/netbox-community/netbox/actions/runs/28465078901/job/84363744333?pr=22562).

Looks like the CI fail was just a random infra thing lagging. Re-running it now.

I don't think any other changes are needed.

jnovinger (approved):

PR #22574, opened 2026-06-30
22378 reshape customscript object column

### LINKED ISSUE #22378: Reshape object column for custom script log view
### NetBox version

v4.6.2

### Feature type

### Proposed functionality

When writing custom scripts, the `log_*()` functions take a parameter obj which will create a link to the corresponding object in the log view of the respective script.
However, object names aren't unique most of the time, especially with some of the nested models, which make the log view hardly useful. For example when working with interfaces, we may get something like this:  
Image

It would be nicer, if the object column of the view could show more detailed information, to make the objects more easily identifiable. It could look for example like this (edited picture):

Image

Since `obj` may contain any kind of model depending on the contents of the custom script, it may even be worth considering to make the output configurable, for example by some sort of format string, so different scripts could make the Link look different

### Use case

Users of custom scripts could benefit from a clearer overview with objects more easily identifiable.

### Database changes

No database changes should be made. The propose only targets the display of the existing information.

### External dependencies

No changes.

mburggraf:
### Closes: #22378

I went for a simple approach, using `parent_object` when present. I was thinking about doing the same for `.device` and `.group`, but wasn't convinced it would help much.

I was wondering, if there is some sort of convention on how a parent object should be referred to (see discussion #22403).

Removed the unused `render_url()` function as well.

### DIFF
--- a/netbox/extras/tables/tables.py
+++ b/netbox/extras/tables/tables.py
@@ -816,6 +816,28 @@ class ScriptResultsTable(BaseTable):
         )
 
     def render_object(self, value, record):
+        try:
+            if record['url']:
+                from django.urls import resolve
+                match = resolve(record['url'])
+                view_class = match.func.view_class
+                if not (hasattr(view_class, "queryset") and hasattr(view_class.queryset, "model")):
+                    return format_html("<a href='{}'>{}</a>", record['url'], value)
+                model = view_class.queryset.model
+                obj = model.objects.get(**match.kwargs)
+
+                if hasattr(obj, "parent_object") and obj.parent_object:
+                    return format_html(
+                        "<a href='{}'>{}</a> / <a href='{}'>{}</a>",
+                        obj.parent_object.get_absolute_url(),
+                        obj.parent_object,
+                        record['url'],
+                        value,
+                    )
+        except RuntimeException:
+            # in case anything goes wrong at all, we always can fall back to the previous style
+            pass
+
         return format_html("<a href='{}'>{}</a>", record['url'], value)
 
 
--- a/netbox/extras/tables/tables.py
+++ b/netbox/extras/tables/tables.py
@@ -2,6 +2,7 @@ import json
 
 import django_tables2 as tables
 from django.template.defaultfilters import filesizeformat
+from django.urls import resolve
 from django.utils.html import format_html
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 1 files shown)


jeremystretch:

netbox/extras/tables/tables.py
```diff
@@ -816,10 +816,29 @@ class Meta(BaseTable.Meta):
...
+
+                if hasattr(obj, "parent_object") and obj.parent_object:
+                    return format_html(
+                        "<a href='{}'>{}</a> / <a href='{}'>{}</a>",
+                        obj.parent_object.get_absolute_url(),
+                        obj.parent_object,
+                        record['url'],
+                        value,
+                    )
+        except RuntimeException:
```
jeremystretch: This is an invalid exception, but it also doesn't adequately contain a potential failure. The method will still attempt to reference `record['url']`, which (per line 820) might not exist.

netbox/extras/tables/tables.py
```diff
@@ -816,10 +816,29 @@ class Meta(BaseTable.Meta):
         )
 
     def render_object(self, value, record):
-        return format_html("<a href='{}'>{}</a>", record['url'], value)
+        try:
+            if record['url']:
+                from django.urls import resolve
```
jeremystretch: Imports should go at the top level unless there's a specific need to nest them.

netbox/extras/tables/tables.py
```diff
@@ -816,10 +816,29 @@ class Meta(BaseTable.Meta):
...
+                match = resolve(record['url'])
+                view_class = match.func.view_class
+                if not (hasattr(view_class, "queryset") and hasattr(view_class.queryset, "model")):
+                    return format_html("<a href='{}'>{}</a>", record['url'], value)
+                model = view_class.queryset.model
+                obj = model.objects.get(**match.kwargs)
```
jeremystretch: We need to avoid potentially triggering an individual query per log line.

jeremystretch (changes requested):
I appreciate what you're trying to achieve with this change, but I'm not sure it' feasible from a performance standpoint.

netbox/extras/tables/tables.py
mburggraf: Oh, must have missed it.
Wonder why the linter didn't complain… fixed.

netbox/extras/tables/tables.py
mburggraf: My intention was, that if anything goes wrong, the whole function would execute in the same way, it does currently, before my PR.
So if `record['url']` is empty, it currently generates a link with empty `href`…
I fixed the Exception name.
But I noticed, that the `if` statement is kinda useless if I accept running into an exception anyways, so I removed the `if`-line for performance reasons.

netbox/extras/tables/tables.py
mburggraf: Actually I wanted to go with a change, which has as minimal impact on the current structures as possible.

The suggestion from Claude to injected the parent object information into the log record at write time would go against this, since it would impact the whole log format and therefore breaking backwards compatibility. (Scripts which were run before the change don't possess this information)

The additional query is only done for those log lines, who have set an object in the first place (since `render_object()` isn't even called when there is no object), therefore IMHO the trade-off regarding the performance does not weigh as heavily in comparison.

mburggraf:
Technically I'm still waiting for feedback…

jeremystretch:
@mburggraf I'm not able to offer any further assistance. I'm going to close this out.

PR #22576, opened 2026-06-30
Closes #22548: Improve sidebar action button styling (followup)

### LINKED ISSUE #22548: Tweak styling of action buttons in navigation menu

pheus:
### Closes: #22548

This PR includes two small UI fixes.

First, it normalizes navigation menu button styling so buttons using the default color are rendered with the secondary ghost style. This fixes plugin menu buttons that set their color to `"default"` instead of leaving it unset, while still preserving explicitly configured colors.

Second, it updates the saved filter dropdown in table controls so it is only rendered when the filter form includes a `filter_id` field. This prevents errors on views that provide a filter form but do not support saved filters, such as the connection list views.

### DIFF
--- a/netbox/templates/inc/table_controls_htmx.html
+++ b/netbox/templates/inc/table_controls_htmx.html
@@ -18,7 +18,7 @@
     </div>
   </div>
 
-  {% if filter_form %}
+  {% if filter_form and filter_form.filter_id %}
     <div class="col-auto d-print-none">
       <div class="input-group">
         <label for="saved-filter-select" class="input-group-text">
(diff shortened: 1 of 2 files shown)


jeremystretch (approved):
Thanks @pheus!

PR #22580, opened 2026-06-30
Closes #21712: Support description annotations for static choice form fields

### LINKED ISSUE #21712: Support description annotations in static select dropdowns (parity with DynamicModelChoiceField)
### NetBox version

v4.5.2

### Feature type

### Proposed functionality

Add support for rendering **description annotations** (subtitle text below each option label) in static `<select>` dropdowns, matching the existing behavior of `DynamicModelChoiceField`'s AJAX-powered dropdowns.

Currently, NetBox's TomSelect integration renders descriptions only for dynamic (API-powered) selects via `DynamicTomSelect` in `dynamic.ts`. Static selects initialized by `initStaticSelects()` in `static.ts` use `NetBoxTomSelect` without a custom render function, so there is no way to display descriptions.

**Proposed implementation:**

1. **New widget or enhancement to `APISelect`**: Add a `DescriptionSelect` widget (or extend the existing `Select` widget) that accepts a description mapping and renders `data-description` attributes on `<option>` elements:

```python
# Example widget
class DescriptionSelect(forms.Select):
    def create_option(self, name, value, label, selected, index, subindex=None, attrs=None):
        option = super().create_option(name, value, label, selected, index, subindex, attrs)
        if description := self.descriptions.get(value, ''):
            option['attrs']['data-description'] = description
        return option
```

2. **Update `initStaticSelects()` in `static.ts`**: Add a custom render function that checks for `data-description` on `<option>` elements and renders them as `<br /><small class="text-secondary">...</small>` — the same HTML pattern already used in `dynamic.ts`:

```typescript
// In static.ts — detect and render descriptions for static selects
function renderOption(data: TomOption, escape: typeof escape_html) {
  const option = select.querySelector(`option[value="${data.value}"]`);
  const description = option?.dataset.description;
  let html = `<div>${escape(data.text)}`;
  if (description) {
    html += `<br /><small class="text-secondary">${escape(description)}</small>`;
  }
  html += `</div>`;
  return html;
}
```

3. **Optional: A `ModelChoiceField` subclass** (e.g., `DescriptionModelChoiceField`) that automatically populates the widget's description mapping from the queryset's `.description` field, similar to how `DynamicModelChoiceField` works with the API.

### Use case

Plugin developers who need **filtered dropdowns with description annotations** currently have no clean path to achieve this. The core issue is that `DynamicModelChoiceField` ignores the `.queryset` set in `__init__()` for the dropdown UI — it fetches all options from the REST API via AJAX. The queryset only affects server-side validation on form submission.

This means:
- `DynamicModelChoiceField` with a filtered queryset: dropdown shows **all** objects (queryset ignored for UI), descriptions work
- `ModelChoiceField` with a filtered queryset: dropdown correctly filters, but **no description annotations**

**Concrete example**: Our plugin restricts which `DeviceRole` values are available per tenant hierarchy. The filtering logic (tenant group ancestry, cross-tenant intersection) cannot be expressed via `query_params` on the standard API endpoint. We need a `ModelChoiceField` with a custom queryset — but then we lose the description subtitle styling.

**Current workaround**: Build a custom REST API endpoint that duplicates the filtering logic, register a URL for it, and use `DynamicModelChoiceField` with `widget=APISelect(api_url='/api/plugins/.../custom-endpoint/')`. This works but requires significant boilerplate:
- A custom DRF `APIView` returning `{count, results}` with `id`, `display`, `description` fields
- URL registration for the endpoint
- Using `query_params` with `$field` references for dynamic dependencies
- Redundant server-side validation in `clean()` as a safety net

With native static description support, all of this would be replaced by a `ModelChoiceField` with a `DescriptionSelect` widget — a single line change.

### Database changes

None. This is a frontend/widget change only.

### External dependencies

None. Uses existing TomSelect library already bundled with NetBox.

jeremystretch:
### Closes: #21712

- Introduce a new `Choice` dataclass to be passed in place of 2- or 3-tuples to `ChoiceSet` subclasses
- Introduce two new custom form widgets to render annotated descriptions: `Select` and `SelectMultiple` (which shadow Django's built-in widgets)
- Introduce two new custom form fields to support description annotations on static choice fields (employing the new widgets): `ChoiceField` and `MultipleChoiceField` (which shadow Django's built-in fields)
- Extend the frontend widgets to annotate description text (see below)
- Update all applicable choice and multi-choice form fields to employ the new fields
- Add relevant tests

2026-07-01 14 54 20 netbox 1f72dfe1f776


jeremystretch:

jeremystretch:

jeremystretch:
#### Item 1 — `AttrChoiceMixin` hardcoded `forms.ChoiceField.choices.fset` ✅ Fixed
Replaced the hardcoded base-class reference with a small `_parent_choices_property(cls)` helper that walks `cls.__mro__` to find the `choices` property on the first class *after* `AttrChoiceMixin` (the Django field's own property), and delegates both the getter and setter through it. Now `ChoiceField`, `TypedChoiceField`, and `MultipleChoiceField` each resolve to their respective Django base, and a setter override on any intermediate class is honored rather than silently bypassed. I proved the fix with a synthetic subclass whose overriding setter now *is* invoked (it wasn't before), and added a `test_choices_setter_delegates_through_mro` regression test.

#### Item 2 — `TagFilterField` `Choice` wrapping was dead code ✅ Reverted
`TagFilterField` extends Django's plain `forms.MultipleChoiceField` (not NetBox's description-aware `MultipleChoiceField`), so wrapping its options in `Choice` could never render descriptions. I reverted those to plain tuples and dropped the now-unused `Choice` import from `fields.py`.

#### Item 3 — Silent description loss for callable choices ⏸️ Declined (with rationale)
`_map_choice_attr` receives the *unevaluated* callable, because Django intentionally defers callable choices (e.g. `DataSourceForm.type = ChoiceField(choices=get_data_backend_choices)` passes the bare callable so it's resolved at render time, not construction). Eagerly invoking it to detect descriptions would break that deferral. And since callable-based choices legitimately carry no descriptions in essentially all real usage, a runtime warning would fire constantly for correct code (`DataSourceForm.type`, `event_types`, tag filters, …) — pure noise. The behavior is already documented in the docstring, so I left it as-is.

bctiemann (approved):

PR #22590, opened 2026-07-02
Closes #22589: Cache serializers

### LINKED ISSUE #22589: Serializer reinstantiation for the cables list API endpoint causes performance issues
### NetBox Version

4.6.4

### Python Version

3.14

### Area(s) of Concern

- [ ] User Interface
- [X] REST API
- [ ] GraphQL API
- [ ] Python ORM
- [ ] Other

### Observations

(Note: This is similar to netbox-community/netbox#22442)

Use a fully empty netbox, generate a key and run this script: (AI-Generated, but anything else is written by human)  <br>[cbl.py  ](<https://github.com/user-attachments/files/29602635/cbl.py>)<br>It just creates < />1700 cables for better testing.

Enable the pyinstrument profiler and go to [http://localhost:8000/api/dcim/cables/?limit=1000&profile1&format=json](<http://localhost:8000/api/dcim/cables/?limit=1000&profile1&format=json>) (profile1 as PYINSTRUMENT_URL_ARGUMENT, as otherwise it collides with the profile field)

Image

If we checkout GenericObjectSerializer, we can see, that the serializer is always reused:

```
class GenericObjectSerializer(serializers.Serializer):
    """
    Minimal representation of some generic object identified by ContentType and PK.
    """
    object_type = ContentTypeField(
        queryset=ContentType.objects.all()
    )
    object_id = serializers.IntegerField()
    object = serializers.SerializerMethodField(read_only=True)

    def to_internal_value(self, data):
        data = super().to_internal_value(data)
        model = data['object_type'].model_class()
        return model.objects.get(pk=data['object_id'])

    def to_representation(self, instance):
        object_type = ObjectType.objects.get_for_model(instance)
        data = {
            'object_type': object_type_identifier(object_type),
            'object_id': instance.pk,
        }
        if 'request' in self.context:
            data['object'] = self.get_object(instance)

        return data

    @extend_schema_field(serializers.JSONField(allow_null=True))
    def get_object(self, obj):
        serializer = get_serializer_for_model(obj)
        return serializer(obj, nested=True, context=self.context).data
```

Using the same command for checking the time:

```
time for i in {0..9}; do echo $i; curl -s -H "Authorization: Token nbt_XXX.XXX" http://127.0.0.1:8000/api/dcim/cables/?limit=1000 >/dev/null; done
0
1
2
3
4
5
6
7
8
9

real	2m45.588s
user	0m0.021s
sys	0m0.031s
```

(curl is not that precise, but shows at least the generic "does it get faster?")

### Proposed Changes

Basically the exactly the same fix, as for netbox-community/netbox#22442  (feel free to assign me, if you accept this ticket)

Update `GenericObjectSerializer::get_object` to:

```python
    @extend_schema_field(serializers.JSONField(allow_null=True))
    def get_object(self, obj):
        if obj.__class__ not in self._serializer_cache:
            serializer = get_serializer_for_model(obj)(nested=True, context=self.context)
            self._serializer_cache[obj.__class__] = serializer
        else:
            serializer = self._serializer_cache[obj.__class__]
        return serializer.to_representation(obj)
```

This yields a faster endpoint:

```
time for i in {0..9}; do echo $i; curl -s -H "Authorization: Token XXX" http://127.0.0.1:8000/api/dcim/cables/?limit=1000 >/dev/null; done
0
1
2
3
4
5
6
7
8
9

real	0m53.233s
user	0m0.016s
sys	0m0.032s
```

Image

This seems to be the last endpoint with such issues

JCWasmx86:
### Closes: #22589

This change introduces caching for serializers to avoid the overhead of enumerating the fields. This will improve the performance of the cables API listing endpoint.

### DIFF
--- a/netbox/netbox/api/serializers/generic.py
+++ b/netbox/netbox/api/serializers/generic.py
@@ -22,6 +22,10 @@ class GenericObjectSerializer(serializers.Serializer):
     object_id = serializers.IntegerField()
     object = serializers.SerializerMethodField(read_only=True)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._serializer_cache = {}
+
     def to_internal_value(self, data):
         data = super().to_internal_value(data)
         model = data['object_type'].model_class()
@@ -40,5 +44,7 @@ class GenericObjectSerializer(serializers.Serializer):
 
     @extend_schema_field(serializers.JSONField(allow_null=True))
     def get_object(self, obj):
-        serializer = get_serializer_for_model(obj)
-        return serializer(obj, nested=True, context=self.context).data
+        if obj.__class__ not in self._serializer_cache:
+            self._serializer_cache[obj.__class__] = get_serializer_for_model(obj)(nested=True, context=self.context)
+        serializer = self._serializer_cache[obj.__class__]
+        return serializer.to_representation(obj)


jeremystretch:

JCWasmx86:
> Plugin serializers registered via register_serializer_resolver() are third-party code with no guarantee they're safe for multi-call reuse. A plugin serializer that stores per-object state during to_representation or __init__ (e.g. caching a computed attribute on self) would silently produce incorrect output for the second and subsequent objects of the same type.

I think it doesn't apply here, as all possible models are stored here: https://github.com/netbox-community/netbox/blob/main/netbox/dcim/constants.py#L118C1-L118C25 and belong to netbox. (And I think it's not configurable)

Furthermore serializers can be probably reused, as `@cached_property` is added to the fields property used by to_representation: https://github.com/encode/django-rest-framework/blob/main/rest_framework/serializers.py#L377


> The cached serializer has self.instance = None: the original code passed obj as the positional arg to the serializer constructor, which sets self.instance = obj. The new code does not. While none of the current NetBox serializers rely on self.instance inside to_representation, any future serializer that does will break silently here.

to_representation is documented as 'Object instance -> Dict of primitive datatypes.', so it doesn't depend on the instance passed in the constructor

> A safer alternative that still avoids repeated get_serializer_for_model lookups (which involve module imports) would be to cache the serializer class rather than the instance:

The slow part is not the extraction of serializer classes, instead it's the extraction of fields

netbox/netbox/api/serializers/generic.py
```diff
@@ -40,5 +44,9 @@ def to_representation(self, instance):
 
     @extend_schema_field(serializers.JSONField(allow_null=True))
     def get_object(self, obj):
-        serializer = get_serializer_for_model(obj)
-        return serializer(obj, nested=True, context=self.context).data
+        if obj.__class__ not in self._serializer_cache:
+            serializer = get_serializer_for_model(obj)(nested=True, context=self.context)
+            self._serializer_cache[obj.__class__] = serializer
+        else:
+            serializer = self._serializer_cache[obj.__class__]
```
jeremystretch: We can simplify this a bit.
```suggestion
        if obj.__class__ not in self._serializer_cache:
            self._serializer_cache[obj.__class__] = get_serializer_for_model(obj)(nested=True, context=self.context)
        serializer = self._serializer_cache[obj.__class__]
```

jeremystretch (changes requested):
Agree that this appears to be safe, though I have one suggestion for cleanup. Attached some Claude analysis below for reference.

---

**What it does:** `GenericObjectSerializer.get_object()` previously constructed a fresh nested serializer for every object it rendered. This PR adds a per-instance `_serializer_cache` keyed by the object's class, so a nested serializer is built once per model type and reused via `to_representation(obj)`. This targets the cables API list endpoint, where `a_terminations`/`b_terminations` (`GenericObjectSerializer(many=True)`) render many objects of the same few types.

**Findings:** Nothing survived verification.

I checked the mechanism against the ways it could plausibly break:

- **Reuse safety** — Sharing one serializer instance across many objects of the same type is exactly what DRF's `ListSerializer` does internally; `to_representation(instance)` receives its object as an argument and reads no per-object serializer state. I confirmed all `self.instance` usages in NetBox serializers (`base.py`, `features.py`, `device_components.py`, `devices.py`) live on the write/validation path (`validate`/`to_internal_value`), never in `to_representation`. Read-path reuse is safe.
- **The `.data` → `to_representation()` switch** — Correct and deliberate. Calling `.data` on a reused instance would memoize the first object's result in `_data` and return it forever; calling `to_representation(obj)` directly avoids that trap. `.data` is equivalent to `to_representation(instance)` anyway (returns a `ReturnDict` vs. plain dict, but both serialize to identical JSON).
- **Cache lifetime / leak** — `_serializer_cache` lives on the `GenericObjectSerializer` instance (one per field per request), bounded by the number of distinct termination model classes. It's garbage-collected with the request. No unbounded growth.
- **`many=True` interaction** — `super().__init__()` runs on the child serializer created by DRF's `many_init`, so `_serializer_cache` is initialized on the single shared child and correctly persists across all list items. Context (`self.context`) is stable for the instance's lifetime, so caching the serializer with its context at first construction is sound.

The change is correct and well-scoped.

netbox/netbox/api/serializers/generic.py
JCWasmx86: Thanks. I've applied your change and added you as patch co-author

jeremystretch (approved):

PR #22594, opened 2026-07-02
#21992: Additional cleanup

jeremystretch:
This PR addresses some additional feedback I raised on PR #22452.

**1. Docstring no longer references the PR** (`netbox/netbox/api/viewsets/__init__.py`)
`exception_to_response()`'s docstring now states the consolidation-with-`dispatch()` note inline as a plain forward-looking note, with no PR reference.

**2. Confusing method name renamed** (`mixins.py`)
`_maybe_background_bulk_create()` is gone. There's now a single generic `_handle_background_request(request, action, action_kwargs=None)` used by all three bulk actions.

**3. Validation deferred entirely and made consistent** (`mixins.py` + `__init__.py`)
- The synchronous full-serializer validation on the create path is removed.
- The background check now runs at the **top** of `create()`, `bulk_update()`, and `bulk_destroy()` — before any validation — so all three behave identically.
- The `If-Match` and no-worker guards remain synchronous (they're request preconditions, not payload validation), so those still 400/503 immediately.
- The `BackgroundOperationMixin` docstring and the `getattr(...)` fallbacks were updated accordingly.

**4. `copy_safe_request()` employed** (`jobs.py` + `mixins.py`)
- The enqueue side now passes `request=copy_safe_request(request, include_files=False)` (plus `scheme`, which the util doesn't capture), dropping the separate `host`/`method`/`request_id` kwargs.
- `AsyncAPIJob._build_request()` seeds the reconstructed `WSGIRequest` from the snapshot's `META`. Since `SERVER_NAME`/`SERVER_PORT`/`HTTP_HOST` are already correctly separated by the original WSGI layer, **all the `urlsplit`/IPv6 special-casing is gone** (the `urllib.parse.urlsplit` import was removed too).

**Behavioral change (intended):** a malformed bulk payload with `?background=true` now returns `202` and the job fails at execution (error captured in `job.data`), rather than a synchronous `400`. Tests and `docs/integrations/rest-api.md` were updated to reflect this.

### DIFF
--- a/docs/integrations/rest-api.md
+++ b/docs/integrations/rest-api.md
@@ -744,7 +744,7 @@ http://netbox/api/dcim/sites/ \
 
 Bulk write operations (creating, updating, or deleting multiple objects via a model's list endpoint) can optionally be processed as a [background job](../features/background-jobs.md) rather than synchronously. This is useful for large batches that would otherwise hold the connection open long enough to risk a proxy or gateway timeout.
 
-To request background processing, append the `background=true` query parameter to a bulk write request. NetBox validates the request immediately and, if it is well-formed and authorized, enqueues a job and returns an `HTTP 202 Accepted` response containing the job's ID and URL. The actual write is performed later by a worker, running the same logic (and preserving the same all-or-none transaction semantics) as the synchronous path.
+To request background processing, append the `background=true` query parameter to a bulk write request. NetBox enqueues a job and returns an `HTTP 202 Accepted` response containing the job's ID and URL. The actual write is performed later by a worker, running the same logic (and preserving the same all-or-none transaction semantics) as the synchronous path. Note that the request payload is **not** validated before the job is enqueued; validation is deferred to the worker (see below).
 
 ```no-highlight
 curl -s -X PATCH \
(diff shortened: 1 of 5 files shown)


jeremystretch:

jeremystretch:

jeremystretch:
**Point 1 (plugin breaking change) — no action.** `AsyncAPIJob` is new, unreleased feature-branch code (#22452, for 4.7). The `(request_id, method, host)` signature it's compared against also only ever existed in that unreleased code, and the internals aren't a documented public plugin API. Nothing has shipped that a plugin could break against, so no release note or migration path is warranted.

**Point 2 (redundant User serialization) — fixed.** Added `request_copy.user = None` in `_enqueue_bulk_job()` (`netbox/netbox/api/viewsets/mixins.py`), alongside the existing `COOKIES = {}` drop, and extended the comment to explain why. The worker re-fetches the user by PK and never reads the copied user, so this avoids pickling a full `User` instance (with its permission cache and relations) into Redis for the job's lifetime.

jnovinger (approved):

PR #22602, opened 2026-07-03
Fixes #22568: Fix ValueError exception when receiving an invalid `filter_id` value

### LINKED ISSUE #22568: Non-integer `filter_id` query parameter raises an uncaught ValueError (HTTP 500) on object lists
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

1. Go to any object list, e.g. `/dcim/devices/`.
2. Manually edit the URL to append a non-integer `filter_id`: `/dcim/devices/?filter_id=abc`.

Note: this is not reachable through normal UI use. The `filter_id` parameter is emitted by the Saved Filter selector, which only ever submits integer PKs of existing saved filters. Reaching the error requires editing the URL by hand (or following a malformed/stale link).

### Expected Behavior

The invalid `filter_id` is ignored and the list renders (same as a non-existent integer like `?filter_id=999999`, which works).

### Observed Behavior

HTTP 500 — uncaught `ValueError: Field 'id' expected a number but got 'abc'`.

jeremystretch:
### Fixes: #22568

Normalize received ID values before passing them to the SavedFilter queryest

### DIFF
--- a/netbox/netbox/filtersets.py
+++ b/netbox/netbox/filtersets.py
@@ -108,9 +108,18 @@ class BaseFilterSet(django_filters.FilterSet):
         # Apply any referenced SavedFilters
         if data and ('filter' in data or 'filter_id' in data):
             data = data.copy()  # Get a mutable copy
+
+            # Coerce filter_id values to integers, ignoring any which are not valid (see #22568)
+            filter_ids = []
+            for f_id in data.pop('filter_id', []):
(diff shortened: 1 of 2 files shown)


jeremystretch:

jnovinger (approved):

PR #22605, opened 2026-07-03
Closes #22603: Add experimental Python packaging support for NetBox

### LINKED ISSUE #22603: Add experimental Python packaging support for NetBox
### Proposed Changes

Add the code and release infrastructure needed to build NetBox as an official Python package.

This work introduces an experimental packaged installation path while keeping the existing archive and git-clone installation workflows intact.

The initial publishing target for this work is TestPyPI only. Production PyPI publishing is intentionally deferred until the packaged workflow is documented and promoted.

The work should cover:

- building NetBox as standard Python package artifacts;
- keeping package metadata aligned with NetBox’s existing release metadata;
- keeping package dependencies aligned with NetBox’s tested release dependency set;
- ensuring published artifacts contain the required application code and read-only runtime assets;
- ensuring local configuration, secrets, and deployment-owned files are not included in published artifacts;
- supporting a clear runtime layout for packaged installs;
- keeping local deployment state outside the installed package and virtual environment;
- supporting flexible virtual environment locations;
- supporting configuration discovery for packaged installs without changing existing source/archive behavior;
- providing one package-friendly `netbox` command;
- supporting small pre-Django commands for setup, secret key generation, and version output;
- supporting upgrade tasks for packaged deployments through `netbox upgrade`;
- preserving the existing upgrade workflow for archive and git-clone installs;
- adding CI checks for package build, verification, smoke testing, and TestPyPI publishing.

The packaged install workflow should be treated as experimental during the TestPyPI validation phase. It should not be recommended for production deployments yet.

Out of scope:

- replacing archive or git-clone installs;
- changing NetBox’s plugin model;
- automatically enabling plugins;
- installing or managing external services such as PostgreSQL, Redis, Nginx, or Apache;
- automating service or web server configuration;
- writing system-level deployment files as a side effect of package installation;
- publishing production releases to PyPI during the initial validation phase;
- adding a separate `netbox install` command;
- adding a second bootstrap binary;
- adding user-facing installation documentation;
- changing the long-term dependency management policy beyond what is needed for this packaging work.

### Justification

Publishing NetBox as a Python package gives maintainers a standard artifact to build, verify, and test before making the workflow available to users.

This should be treated as an application distribution for NetBox, not a change to NetBox’s application architecture. The packaged install path should preserve the current Django and plugin behavior as much as possible.

Starting with TestPyPI gives maintainers a safe way to validate packaging behavior, artifact contents, metadata, dependency handling, and installed-package smoke tests before making the workflow available through production PyPI.

The code changes should stay focused on what is needed to support Python package installation. General deployment automation should remain outside the scope of this work.

pheus:
### Closes: #22603

This PR adds the code and release infrastructure needed to build and test NetBox as an official Python package.

The packaged install path is experimental during the TestPyPI validation phase. Existing archive and git-clone workflows remain unchanged and continue to be the supported production installation methods until the packaged workflow is promoted.

This PR targets TestPyPI only. Production PyPI publishing is intentionally deferred until the packaged installation workflow is documented and promoted.

This PR covers:

- package build support;
- package metadata generation;
- package artifact layout;
- packaged-install runtime support;
- configuration discovery for packaged installs;
- one installed `netbox` command;
- pre-Django command handling for version output, setup, and secret key generation;
- upgrade task support for packaged deployments through `netbox upgrade`;
- package verification and smoke testing;
- TestPyPI publishing workflow.

This PR does not:

- replace the current archive or git-clone installation workflows;
- change NetBox’s plugin model;
- install or configure external services;
- automate service or web server configuration;
- add a separate `netbox install` command;
- add a second bootstrap binary;
- add user-facing installation documentation;
- enable production PyPI publishing.

The goal is to make the package artifacts buildable, testable, and safe to iterate on before the packaged install workflow is documented and promoted.


pheus:

pheus:
1. `zensical build` is intentional. NetBox switched from MkDocs to Zensical in v4.6.
2. Agreed, these are false positives. Both commands intentionally print a newly generated, unused secret key for the admin to copy.
3. I’d keep `Production/Stable`. NetBox itself is stable; only this new pip installation path is experimental.
4. Fair point. The publish job is still gated by the smoke test, but we can address PR smoke-test coverage or document why it’s skipped.
5. Good catch. I’ll update this to remove only the `sys.path` entry inserted by the helper.

.github/workflows/release.yml
```diff
@@ -0,0 +1,265 @@
...
+  publish-testpypi:
+    name: Publish package to Test PyPI
+    runs-on: ubuntu-latest
+    needs: [smoke-test, verify-dependencies, verify-sdist]
+    # Publishing always requires a v* tag ref: a tag push publishes to Test PyPI
+    # automatically, and a manual dispatch does the same when the chosen ref is a v* tag.
+    # Branch dispatches still run the build, verify, and smoke-test jobs (a useful dry run)
+    # but the publish job is skipped. Production PyPI publishing is intentionally absent
+    # during the v4.6.x preview; it arrives with the v4.7.0 feature branch.
+    if: startsWith(github.ref, 'refs/tags/v') && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
```
jeremystretch, line 287: It would be ideal to replace `startsWith` with a more exact match on `v\d+\.\d+\.\d+\` but it doesn't look like we have access to a regex filter here.

netbox/netbox/cli.py
```diff
@@ -0,0 +1,92 @@
...
+    from utilities.secret_key import generate_secret_key
+    print(generate_secret_key())
+    return 0
+
+
+def main(argv=None):
+    prog = _prog()
+    args = list(sys.argv[1:] if argv is None else argv)
+
+    if not args or args[0] in ('-h', '--help'):
```
jeremystretch: Can we use `argparse` here to register parameters programatically? This gets us most of the help text for free.

docs/development/building-the-package.md
```diff
@@ -0,0 +1,198 @@
...
+
+`python -m build` produces both an sdist and a wheel (the wheel is built from the sdist). The
+release workflow's `verify-sdist` job rebuilds a wheel from the published sdist and runs
+`scripts/verify_wheel_metadata.py` and `scripts/verify_wheel_contents.py` against it, so a
+missing build input (for example the metadata hook or `base_requirements.txt`) cannot regress
+unnoticed.
+
+### Wheel data layout
+
+Source assets that are not Python modules are force-included with a `netbox/netbox/_data/`
```
jeremystretch: Could you clean up the line wrapping in this paragraph and the ones below? Not sure what happened there.

netbox/utilities/upgrade_tasks.py
jeremystretch, line 1: Why put these in a separate module under `utilities` instead of just keeping them in `core/managemnet/commands/upgrade.py`? IMO it doesn't make sense to position these as reusable utilities.

netbox/netbox/scaffold.py
jeremystretch, line 1: I feel this is out of scope for the goal of simply shipping a Python package. We should not take on administration of deployment-specific configurations. Once the package has been installed, the administrator will still be responsible for configuring related services (gunicorn, nginx, etc.) as they have been.

netbox/netbox/settings.py
```diff
@@ -119,7 +135,10 @@
     'users.delete_token': ({'user': '$user'},),
 })
 DEVELOPER = getattr(configuration, 'DEVELOPER', False)
-DOCS_ROOT = getattr(configuration, 'DOCS_ROOT', os.path.join(os.path.dirname(BASE_DIR), 'docs'))
+_DEFAULT_DOCS_ROOT = os.path.join(os.path.dirname(BASE_DIR), 'docs')
+if not os.path.isdir(_DEFAULT_DOCS_ROOT):
+    _DEFAULT_DOCS_ROOT = None  # pragma: no cover
```
jeremystretch: Shouldn't the `docs/` source root always exist?

netbox/netbox/settings.py
```diff
@@ -30,8 +31,25 @@
 
 RELEASE = load_release_data()
 VERSION = RELEASE.full_version  # Retained for backward compatibility
-# Set the base directory two levels up
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+# Settings package directory (settings.py lives here in both checkout & wheel).
+_SETTINGS_DIR = os.path.dirname(os.path.abspath(__file__))
+
+# In a wheel install, package data is bundled under netbox/_data. In a checkout,
+# the data directories remain siblings of the settings package's parent.
```
jeremystretch: I'd like to dig into why this is. IMO it feels very fragile to accommodate two different file hierarchies.

netbox/netbox/settings_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        if sys.modules.get(module_name) is module:
+            del sys.modules[module_name]
+        raise
+    finally:
+        if module_dir in sys.path:
+            sys.path.remove(module_dir)
+    return module
+
+
+def configuration_dir(module):
```
jeremystretch: Suggest renaming this for clarity (i.e. avoid the assumption that this is creating a directory).
```suggestion
def get_configuration_dir(module):
```

netbox/utilities/upgrade_tasks.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def _docs_source_root():
+    # mkdocs.yml sits beside the application root (one level above BASE_DIR in a checkout);
+    # a wheel install has no documentation sources, so this returns None there.
+    candidate = os.path.dirname(settings.BASE_DIR)
+    return candidate if os.path.isfile(os.path.join(candidate, 'mkdocs.yml')) else None
+
+
+def run_upgrade_tasks(command, *, no_input=False, readonly=False,
```
jeremystretch: This seems largely redundant to `upgrade.sh`, which is retained in its original form. Should we update `upgrade.sh` to wrap the new `manage.py upgrade` command? I'd like to avoid having to maintain both avenues in parallel.

jeremystretch (changes requested):
Nice work @pheus! I think we should talk through the wheel vs. checkout layout considerations; I'm sure there's a good reason for the disparity but it seems very fragile IMO.

.github/workflows/release.yml
pheus: Good point.

The stricter validation is handled by the release tag verification step before publishing. I can update the comment to make that clear, or add a small shell-side check if you’d prefer the workflow itself to fail earlier for tags that do not match the expected release format.

netbox/netbox/cli.py
pheus: Yes, that makes sense.

netbox/netbox/scaffold.py
pheus: That’s a good point. I agree NetBox should not take ownership of administering deployment-specific service configuration.

My intent was to make the existing contrib examples available for wheel installs in roughly the same way they are available in a source/archive install, not to automate service configuration. But I agree the current `setup` command may blur that line.

netbox/netbox/settings.py
pheus: For checkout/source installs, yes, the docs root should exist.

For wheel installs, the docs source tree is not shipped and the wheel cannot build or provide embedded docs. That is why this falls back to `None` when the directory is absent.

netbox/netbox/settings.py
pheus: The split comes from the wheel layout: immutable application/package data lives in the installed Python package, while mutable instance-local files need to live outside the wheel under the instance root. For checkout installs, the existing source tree layout remains unchanged.

I can try to reduce the fragility here by centralizing the install-mode/path decisions more clearly instead of spreading wheel-vs-checkout checks through settings.

netbox/utilities/upgrade_tasks.py
pheus: I originally split this out to share the main logic for `install` and `upgrade` commands. Later I dropped the `install` command, but I didn’t move the shared code back.

I’ll move this logic into `core/management/commands/upgrade.py`.

netbox/utilities/upgrade_tasks.py
pheus: Good point. My intention is not to alter `upgrade.sh` during the NetBox v4.6.x release series, and instead switch the script to `manage.py upgrade` in NetBox v4.7.0.

That keeps the existing source/archive upgrade path unchanged for now, while still allowing us to test the management command and adopt it later.

pheus:
Thank you for the thoughtful review! I appreciate you taking the time to go through the packaging approach in detail.

#### Wheel vs. checkout layout

The main reason for the split is that a wheel installs NetBox’s application code and bundled package data into the Python environment, while instance-local files like `configuration.py`, media, reports, scripts, and collected static files still need to live outside the package. This separates immutable application data in `site-packages/` from mutable instance data.

There is also an import-path issue with the current configuration loading model. Today, NetBox loads its default configuration as the Python module `netbox.configuration`, which maps to the historical source/archive file path `netbox/netbox/configuration.py`. For a wheel install, however, `netbox` is the installed package from the virtual environment, so that import resolves against `site-packages`, not against an instance-local `/opt/netbox/netbox/configuration.py` file.

That is why the wheel install uses an instance-local `conf/configuration.py` file instead. Because this file is intentionally outside the installed `netbox` package, it has to be loaded by path rather than through the normal package import mechanism.

I did keep a fallback for the legacy source-style path for a grace period, but making that the default for wheel installs would require users to create a mostly empty `netbox/netbox/` hierarchy under the instance root just to hold configuration. I wanted to avoid introducing that artificial structure for new pip-installed instances.

My goal was to keep the current checkout/archive layout unchanged and only introduce the separate instance root for wheel installs. Changing the configuration path for all installation methods could be discussed separately, but I think that would be better handled as a breaking-change release rather than as part of this packaging PR.

netbox/netbox/settings.py
jeremystretch: We need to figure out a way to include the docs to ensure support for the UI-embedded documentation.

pheus:

pheus:

1. The wheel no longer builds documentation inside `site-packages`. The documentation is rendered before the package artifacts are built, and the rendered site is included in the wheel and sdist. A wheel install now uses the bundled rendered docs, which are picked up by `collectstatic`, so there is no runtime write into the installed package directory.

2. Agreed this is a false positive. I have dismissed the CodeQL alert. Both commands intentionally print a freshly generated, unused secret key for the administrator to copy.

3. I have reduced the scope of `netbox setup`. It now only creates the minimal local instance files and copies the bundled contrib examples verbatim. It does not rewrite service files, install systemd units, configure a web server, or otherwise take ownership of deployment-specific administration. Existing files are also left untouched.

4. This behavior is now intentional. Wheel installs do not need to build documentation anymore because the rendered docs are bundled in the package. `--build-docs` remains useful for checkout/source installs, but for wheel installs there is no documentation source tree to build from.

netbox/netbox/settings.py
```diff
@@ -594,13 +603,15 @@ def _setting(name, default=None):
 X_FRAME_OPTIONS = 'SAMEORIGIN'
 
 # Static files (CSS, JavaScript, Images)
-STATIC_ROOT = BASE_DIR + '/static'
+STATIC_ROOT = getattr(configuration, 'STATIC_ROOT', os.path.join(NETBOX_ROOT, 'static')).rstrip('/')
```
jeremystretch: Is there a need to expose `STATIC_ROOT` as a configuration parameter? If so, it needs to be documented in `docs/configuration/system.md`. Would prefer to keep this fixed if we can, as allowing administrators to change the path feels like it's inviting breakage.

jeremystretch (changes requested):
Just one more minor note.

netbox/netbox/settings.py
pheus: You're right, there is no need to expose this as a configuration parameter. I've removed it and kept `STATIC_ROOT` fixed. Thanks for the pushback.

jeremystretch (approved):

PR #22606, opened 2026-07-03
Fixes #22578: Ensure shared objects are treated consistently across the UI and APIs

### LINKED ISSUE #22578: REST API does not honor the shared/owner visibility that the UI enforces for SavedFilter and TableConfig
### NetBox Edition

NetBox Community

### NetBox Version

4.6.3

### Python Version

3.14

### Steps to Reproduce

1. Two non-superuser accounts. As victim, create a SavedFilter with shared=false and a recognizable filter. The admin has granted both users extras.view_savedfilter without a {'user':'$user'} constraint (the common case).
2. As attacker: GET /api/extras/saved-filters/ returns the victim's private SavedFilter, including its stored filter parameters. The same filter is hidden from the attacker in the UI.

### Expected Behavior

NetBox's UI restricts SavedFilter and TableConfig to "shared, or owned by the current user" in code (SharedObjectViewMixin).

### Observed Behavior

REST API applies no such filter, so a user with an unconstrained view permission can read other users' private (shared=false) saved filters and table configs through the API.

jeremystretch:
### Fixes: #22578

- Introduce SharedObjectQuerySet, which provides a `restrict_to_shared()` method for filtering shared objects by user
- Alter the SavedFilter and TableConfig models to derive their managers from SharedObjectQuerySet
- Introduce SharedObjectMixin and SharedObjectQuerySetMixin to call `restrict_to_shared()`
- Update SharedObjectViewMixin to call `restrict_to_shared()`
- Add tests

### DIFF
--- a/netbox/extras/api/mixins.py
+++ b/netbox/extras/api/mixins.py
@@ -15,7 +15,17 @@ __all__ = (
     'ConfigContextQuerySetMixin',
     'ConfigTemplateRenderMixin',
     'RenderConfigMixin',
+    'SharedObjectQuerySetMixin',
 )
+
+
+class SharedObjectQuerySetMixin:
+    """
+    Restrict the queryset to shared objects, or those owned by the current user, unless the user is a superuser.
+    This mirrors the visibility enforced in the UI by extras.utils.SharedObjectViewMixin.
+    """
+    def get_queryset(self):
+        return super().get_queryset().restrict_to_shared(self.request.user)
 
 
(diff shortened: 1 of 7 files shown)


jeremystretch:

jeremystretch:

jeremystretch:
1. This is acceptable given the context in which the method is used.
2. No test is necessary for this.

pheus (approved):
This looks good to me. Thanks!

PR #22609, opened 2026-07-04
Fixes #22598: Patch `get_scheduler_pid()` to resolve RQ discrepancy

### LINKED ISSUE #22598: ValueError when viewing Background Tasks with rq 2.10
### NetBox Edition

NetBox Community

### NetBox Version

4.6.4

### Python Version

3.14

### Steps to Reproduce

- add any RQ job using `queue.enqueue_at(...)` to run later
  - this will create an rq:scheduler:&lt;hex string&gt; (with the rq version in the netbox community docker image)  
     (https://github.com/rq/rq/releases/tag/v2.10 mentions that rq  now stores scheduler metadata even without without rq_scheduler)
- open /core/background-queues/

### Expected Behavior

netbox shows a list of redis queues

### Observed Behavior

ValueError at /core/background-queues/
invalid literal for int() with base 10: '2c209b9f487443e6b5acbaaede582117'

jeremystretch:
### Fixes: #22598

The bug was in `django_rq==4.1.0`'s `get_scheduler_pid()`, called via `get_statistics()` from both the UI (`core/views.py:525`) and REST (`core/api/views.py:167`) background-queue views. It cast the scheduler lock value to `int()`, but RQ 2.10 stores the scheduler's hex `name` there instead of a PID → `ValueError` → 500.

The shim mirrors the (unreleased) upstream fix by delegating to RQ 2.10's native `Queue.scheduler_pid` property:

- **`netbox/core/utils.py`** — added `get_scheduler_pid(queue)`: returns `False` if the external `rq-scheduler` is in use, otherwise `queue.scheduler_pid` (which correctly resolves hex name → scheduler hash → int PID), and `None` on any other error.
- **`netbox/core/apps.py`** — in `CoreConfig.ready()`, rebind `django_rq.utils.get_scheduler_pid` to our version. `get_statistics()` looks up that name from its module globals at call time, so both call sites are covered without touching either view. A `TODO` marks it for removal once `django_rq` ships a fixed release.
- **`netbox/core/tests/test_views.py`** — added `test_background_queue_list_with_scheduler_lock`, which plants an RQ-2.10-style hex lock + scheduler hash and asserts the view returns 200 and the PID resolves to `12345`. I confirmed this test fails with the exact issue `ValueError` when the fix is disabled.

### DIFF
--- a/netbox/core/apps.py
+++ b/netbox/core/apps.py
@@ -21,8 +21,11 @@ class CoreConfig(AppConfig):
     name = "core"
 
     def ready(self):
+        import django_rq.utils
+
         from core.api import schema  # noqa: F401
         from core.checks import check_duplicate_indexes, check_postgresql_version, check_redis_version  # noqa: F401
+        from core.utils import get_scheduler_pid
         from netbox import context_managers  # noqa: F401
         from netbox.models.features import register_models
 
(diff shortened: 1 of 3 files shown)


jeremystretch:

jeremystretch:

jeremystretch:
Going to cancel this work as `django-rq` now has a v4.1.1 release which should resolve the issue upstream.

PR #22614, opened 2026-07-06
Closes #22538: Add security note for HTTP_CLIENT_IP_HEADERS client-IP trust

### LINKED ISSUE #22538: Add a security consideration for HTTP_CLIENT_IP_HEADERS client-IP trust
### Change Type

Addition

### Area

Configuration

### Proposed Changes

Add a security note to the `HTTP_CLIENT_IP_HEADERS` section of `docs/configuration/system.md` explaining that the configured headers are trusted as the source of the client IP address. The default list (`HTTP_X_REAL_IP`, `HTTP_X_FORWARDED_FOR`, `REMOTE_ADDR`) trusts the forwarded headers first, and since the resolved client IP is used for token `allowed_ips` restrictions and login logging, an administrator should understand the trust requirement. Something like:

> The headers listed in `HTTP_CLIENT_IP_HEADERS` are trusted as the source of the client IP address. Trusting `X-Forwarded-For` or `X-Real-IP` is only safe when NetBox is deployed behind a reverse proxy that overwrites these headers with the real client address. If NetBox is reachable directly, or the proxy passes a client-supplied value through unchanged, a client can spoof its apparent IP and defeat token `allowed_ips` restrictions. Deployments without a trusted proxy should set `HTTP_CLIENT_IP_HEADERS = ('REMOTE_ADDR',)`.

jnovinger:
### Closes: #22538

Adds a security warning to the `HTTP_CLIENT_IP_HEADERS` section of the system configuration docs. The warning covers three points:

- The configured headers are trusted as the source of the client IP address.
- Trusting `X-Forwarded-For` or `X-Real-IP` is only safe behind a reverse proxy that overwrites those headers. A proxy that appends instead preserves the client-supplied value, because NetBox uses the leftmost address in the list.
- Deployments without a trusted proxy should set the parameter to `('REMOTE_ADDR',)`.

Also adds a pointer to this note from the "Client IP Restriction" section of the REST API docs, where token IP restrictions are documented.

### DIFF
--- a/docs/configuration/system.md
+++ b/docs/configuration/system.md
@@ -99,7 +99,7 @@ An ordered list of HTTP request headers inspected to determine the source IP add
 The client IP is used for source-address restrictions on API tokens and for logging failed login attempts.
 
 !!! warning "Client IP trust"
-    The headers listed here are trusted as the source of the client IP address. Trusting `X-Forwarded-For` or `X-Real-IP` is safe only when NetBox is deployed behind a reverse proxy that overwrites these headers with the real client address. If NetBox is reachable directly, or the proxy appends to or passes through a client-supplied value, a client can spoof its apparent IP address and defeat API token client IP restrictions. Deployments without a trusted proxy should set `HTTP_CLIENT_IP_HEADERS = ('REMOTE_ADDR',)`.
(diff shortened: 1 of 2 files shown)


jnovinger:

arthanson (approved):

PR #22616, opened 2026-07-06
Closes #22607: Sanitize HTTP requests passed to template contexts for custom links

### LINKED ISSUE #22607: Sanitize HTTP requests passed to template contexts for custom links
### NetBox version

v4.6.4

### Feature type

New functionality

### Proposed functionality

The current request is passed as context when rendering a custom link, so that attributes of the request (e.g. the name of the current user) can be embedded in the rendered link. While useful, we should sanitize the request object to prevent exposure to potential sensitive attributes (the specific set of which is TBD).

Note: This is potentially a breaking change, as removing access to certain attributes of the request might break existing custom links.

### Use case

This enacts an additional safeguard around custom link rendering, which is already tightly controlled via the permissions framework.

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #22607

- Introduce the `get_safe_request_context()` utility function, which returns only "safe" requests of an HTTP request object
- Update the `{% custom_links %}` template tag to call `get_safe_request_context()`
- Update `send_webhook()` to employ `get_safe_request_context()` for consistency
- Update the custom link and webhook docs
- Add tests

### DIFF
--- a/docs/customization/custom-links.md
+++ b/docs/customization/custom-links.md
@@ -28,9 +28,12 @@ The following context data is available within the template when rendering a cus
 |-----------|-------------------------------------------------------------------------------------------------------------------|
 | `object`  | The NetBox object being displayed                                                                                 |
 | `debug`   | A boolean indicating whether debugging is enabled                                                                 |
-| `request` | The current WSGI request                                                                                          |
-| `user`    | The current user (if authenticated)                                                                               |
+| `request` | A sanitized subset of the current request (see below)                                                             |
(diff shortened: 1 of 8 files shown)


jeremystretch:

jeremystretch:
1. Reverted the errant docs change
2. Tweaked `get_safe_request_context()` to cast `id` as a string

jnovinger (changes requested):
There appears to be a code-related CI failure.

jnovinger (approved):

PR #22618, opened 2026-07-06
Fixes #22617: Editing objects via bulk import form requires "change" permission

### LINKED ISSUE #22617: Bulk import incorrectly asserts "add" permission for object mutations
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.4

### Python Version

3.14

### Steps to Reproduce

1. Create a permission granting a user only the `add` action for a model (e.g. `dcim.add_device`) — no `change` action. Assign it to a test user.
2. As that user, note the ID and current field values of an existing object of that type.
3. Navigate to the model's list view and choose **Import**.
4. Submit a CSV whose header includes `id` and at least one other field, with one data row containing the existing object's `id` and a changed value, e.g.:
   ```
   id,description
   123,modified-without-change-permission
   ```
5. Submit the import.

### Expected Behavior

Updating an existing object via bulk import should require the `change` permission on that model, consistent with the REST API (where `PUT`/`PATCH` map to `change`). A user holding only `add` should not be able to modify existing records; the update rows should be rejected.

### Observed Behavior

The import succeeds and the existing object is overwritten with the submitted values, despite the user having no `change` permission. `BulkImportView.get_required_permission()` returns only `add`, and the update path in `_process_import_records()` loads the existing instance by `id` and saves it without any `change` check. The object-level enforcement runs against a queryset restricted to the `add` action, so `change`-scoped constraints are never applied. This diverges from the REST API, which requires `change` for the equivalent update. The current test suite encodes this behavior as expected (`test_bulk_update_objects_with_permission` grants `add` only and asserts the update succeeds).

jeremystretch:
### Fixes: #22617

The bulk import view is gated only on the `add` permission, but a record carrying an `id` column updates an existing object rather than creating one — which should require `change` (as the REST API enforces). This corrects the gap and consolidates object-level permission enforcement:

1. **`create_and_update_objects()` — model-level gate, object-level prefetch, and aggregate enforcement.**
   - When any record references an existing object by ID, require the `change` permission at the model level (`request.user.has_perm(...)`, i.e. `obj=None`); otherwise raise a `ValidationError` instructing the user to remove the ID column.
   - Prefetch update targets through `objects.restrict(user, 'change')`, so objects outside the user's `change` scope are simply not found (surfacing the existing "does not exist" error).
   - After all records are processed, enforce object-level permissions in aggregate: created objects against the `add`-restricted `self.queryset`, updated objects against the `change`-restricted queryset. This runs inside the caller's atomic transaction, so any violation rolls back the entire import, and it also catches an update that moves an object *out of* the user's `change` constraint.

2. **`_save_object()` — removed the per-object object-level check.** The check that was previously performed here for every saved object is now handled by the aggregate enforcement in `create_and_update_objects()`.

3. **`post()` — removed the inline aggregate check.** The object-level permission check that previously lived in `post()` has moved into `create_and_update_objects()`, so all enforcement lives in one place.

The missing-`change` case surfaces as a `ValidationError` (clear message, form re-rendered, transaction rolled back), consistent with how bulk import reports other problems.

**Testing.** The shared bulk-update view test is split into two:
- `test_bulk_update_objects_without_change_permission` — grants only `add` and asserts the import is rejected with the permissions error and that no target object is modified.
- `test_bulk_update_objects_with_permission` — grants `add` + `change` and verifies the objects are actually updated. This also fixes a pre-existing no-op assertion (`assertEqual(value, value)`) that verified nothing; scalar field values are now compared against the CSV input.

### DIFF
--- a/netbox/dcim/tests/test_views.py
+++ b/netbox/dcim/tests/test_views.py
@@ -1303,6 +1303,22 @@ class ModuleTypeTestCase(ViewTestCases.PrimaryObjectViewTestCase):
         # run base test
         super().test_bulk_update_objects_with_permission()
 
+    def test_bulk_update_objects_without_change_permission(self):
+        # ModuleTypeImportView declares these as additional_permissions, so they're required to reach the view
+        self.add_permissions(
+            'dcim.add_consoleporttemplate',
+            'dcim.add_consoleserverporttemplate',
+            'dcim.add_powerporttemplate',
+            'dcim.add_poweroutlettemplate',
+            'dcim.add_interfacetemplate',
+            'dcim.add_frontporttemplate',
+            'dcim.add_rearporttemplate',
+            'dcim.add_modulebaytemplate',
+        )
+
+        # run base test
+        super().test_bulk_update_objects_without_change_permission()
+
     @tag('regression')
     def test_bulk_import_objects_with_permission(self):
         self.add_permissions(
--- a/netbox/netbox/views/generic/bulk_views.py
+++ b/netbox/netbox/views/generic/bulk_views.py
@@ -489,12 +489,9 @@ class BulkImportView(GetReturnURLMixin, BaseMultiObjectView):
     def _save_object(self, model_form, request, parent_idx):
         _action = 'Updated' if model_form.instance.pk else 'Created'
 
-        # Save the primary object
+        # Save the primary object. Object-level permissions are enforced in aggregate by
+        # create_and_update_objects() once all records have been processed.
         obj = self.save_object(model_form, request)
-
-        # Enforce object-level permissions
-        if not self.queryset.filter(pk=obj.pk).first():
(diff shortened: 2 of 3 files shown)


jeremystretch:

jeremystretch:

jeremystretch:

jeremystretch:
1. Agreed that this could be confusing, however we don't currently have a better alternative due to how permission constraints work.
2. This is acceptable.

jnovinger (commented):
If I'm reading this correctly, changes/updates via bulk import are still gated on `add` permission, as well as now on `change` permission. Was that intentional?

jeremystretch:
Yes, we're keeping the blanket requirement for the `add` permission in place for now.

jnovinger (approved):

PR #22624, opened 2026-07-07
Closes #22577: Add test coverage for Connection list views

### LINKED ISSUE #22577: Add tests for connections list UI views
### Proposed Changes

Add tests for:

- `/dcim/interface-connections/`
- `/dcim/console-connections/`
- `/dcim/power-connections/`

### Justification

These views currently lack any test coverage.

pheus:
### Closes: #22577

Add test cases for Console, Power, and Interface Connection list views. Include query count baselines and shared mixin for read-only connection views that filter by complete cable paths.


pheus:

pheus:
I did consider this, but since it’s a small test-only mixin, I’m okay with letting a missing `url_base` fail as a test authoring error rather than adding a guard for it.

jeremystretch (approved):
Thanks @pheus!

PR #22625, opened 2026-07-07
Fixes #22566: Avoid name collisions when custom script name matches core module

### LINKED ISSUE #22566: Custom script whose filename matches a core app label shadows that app in sys.modules
### NetBox Edition

NetBox Community

### NetBox Version

4.6.2

### Python Version

3.12

### Steps to Reproduce

1. Create a custom script in a file named circuits.py, containing any Script class.
2. Trigger a Django migration-graph rebuild after the script has been loaded. By viewing or syncing a branch with the netbox-branching plugin installed. Equivalently, in the NetBox shell:
```
ScriptModule(file_root='scripts', file_path='circuits.py').get_module()
from django.db import connections
from django.db.migrations.executor import MigrationExecutor
MigrationExecutor(connections['default']).loader.graph.leaf_nodes()
```

### Expected Behavior

A custom script's filename does not collide with core app modules. Loading it has no effect on app or migration resolution.

### Observed Behavior

ScriptModule.get_module() registers the script in sys.modules under the bare name circuits, replacing the real app (circuits/__init__.py) with the script module (which has no __path__). 

The next migration-graph rebuild then fails:
- django.db.migrations.exceptions.NodeNotFoundError: Migration dcim.0160_squashed_0166 dependencies reference nonexistent parent node ('circuits', '0003_squashed_0037')
- A direct import circuits.migrations fails with ModuleNotFoundError: No module named 'circuits.migrations'; 'circuits' is not a package (on NetBox 4.3.7 / Python 3.12 the same condition surfaces as AttributeError: module 'circuits' has no attribute '__path__'; wording varies by Python version, root cause is identical).

jeremystretch:
### Fixes: #22566

- Include a prefix when registering the script module to avoid naming collisions
- Add a test

### DIFF
--- a/netbox/extras/constants.py
+++ b/netbox/extras/constants.py
@@ -28,6 +28,11 @@ IMAGE_ATTACHMENT_IMAGE_FORMATS = {
 # Template Export
 DEFAULT_MIME_TYPE = 'text/plain; charset=utf-8'
 
+# Scripts
+# Prefix applied to dynamically-loaded script/report module names so that a script whose filename
+# matches a core app or package (e.g. "circuits.py") cannot shadow it in sys.modules.
+SCRIPT_MODULE_NAME_PREFIX = '_netbox_script_module_'
+
 # Webhooks
 HTTP_CONTENT_TYPE_JSON = 'application/json'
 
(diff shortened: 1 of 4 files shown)


jeremystretch:

jeremystretch:

jeremystretch:

jeremystretch:
Both points above can be disregarded as existing behavior and unlikely to present any issues.

arthanson (approved):

PR #22627, opened 2026-07-07
Fixes #22626: Ensure custom link names are escaped when rendering fails

### LINKED ISSUE #22626: Custom link name not escaped properly when render fails
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.4

### Python Version

3.14

### Steps to Reproduce

1. As a user holding the `extras.add_customlink` permission, create a Custom Link with:
   - **Name:** ``
   - **Link text** (or **Link URL**) template: content that raises during rendering so the fallback path is taken — e.g. `{{ object.does_not_exist.attr }}` or any Jinja/Django expression guaranteed to throw.
   - Assign it to an object type you can view (e.g. Device).
2. Navigate to an object page of that type where the custom link is rendered (e.g. any Device detail page).
3. Because `CustomLink.render()` raises, the `except Exception` branch in `custom_links()` builds the error button, interpolating the raw custom-link `name` into the HTML.

### Expected Behavior

The custom link's `name` should be HTML-escaped before being interpolated into markup returned via `mark_safe()` — consistent with the existing `escape(group)` usage in the same function. A malicious name should render as inert text, not execute as markup.

### Observed Behavior

The raw, unescaped `name` is emitted into the page and returned via `mark_safe()`, so an HTML/JS payload stored in a custom link's name executes in the browser of any authenticated user (including admins) who views an object page where that custom link's render fails. This is stored XSS across a user trust boundary.

jeremystretch:
### Fixes: #22626

Ensure custom link representations are wrapped with `escape()` when handling exceptions.

### DIFF
--- a/netbox/extras/templatetags/custom_links.py
+++ b/netbox/extras/templatetags/custom_links.py
@@ -67,8 +67,8 @@ def custom_links(context, obj):
                         rendered['link'], rendered['link_target'], button_class, rendered['text']
                     )
             except Exception as e:
-                template_code += f'<a class="btn btn-sm btn-outline-secondary" disabled="disabled" title="{e}">' \
-                                 f'<i class="mdi mdi-alert"></i> {cl.name}</a>\n'
+                template_code += f'<a class="btn btn-sm btn-outline-secondary" disabled="disabled" ' \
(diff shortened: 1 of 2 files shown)


jeremystretch:

pheus (approved):

PR #22628, opened 2026-07-07
Fixes #22539: Restore available IP display for users with constrained…

### LINKED ISSUE #22539: Button showing available IPs in prefix is missing if user permissions has constraints
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

1. Create a user
2. Create a prefix `192.168.0.0/24'
3. Create permission granting `view`, `add`, `change` and `delete`  on object type `IPAM | IP Address` with constraints (see below) for the user.
4. Logon using the created user
5. Navigate to Prefix `192.168.0.0/24`
6. Click on `IP Addresses` tab

Constraint is this:
```
{
    "address__startswith": "192.168.0."
}
```
I used simpler constraints than the ones we use in production but could reproduce the bug nevertheless.

### Expected Behavior

Seeing button(s) showing available IP(s).

### Observed Behavior

The button(s) `X IPs available` are missing. The buttons are visible for users without a sonstraint. A rollback to 4.5.10 using the same database fixes the problem so I figure this was introduced in 4.6.x.

Could be related to [13913](https://github.com/netbox-community/netbox/issues/13913).

Amir-Bakar:
### Closes: #22539

Changes explained in this issue comment:
https://github.com/netbox-community/netbox/issues/22539#issuecomment-4904122860

### DIFF
--- a/netbox/ipam/views.py
+++ b/netbox/ipam/views.py
@@ -1,4 +1,5 @@
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import EmptyResultSet
 from django.db.models import Prefetch
 from django.db.models.expressions import RawSQL
 from django.shortcuts import get_object_or_404, redirect, render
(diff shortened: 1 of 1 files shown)


jeremystretch:

bctiemann (approved):

PR #22630, opened 2026-07-07
Closes #22629: Lower maximum uploaded image size to 50M pixels

### LINKED ISSUE #22629: Enforce a lower maximum image size than the Pillow default
### NetBox version

v4.6.4

### Feature type

### Proposed functionality

The default permissible image size for uploaded images, enforced by Pillow's [`Image.MAX_IMAGE_PIXELS`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.MAX_IMAGE_PIXELS) is 89,478,485. This FR proposes lowering the default value to 25,000,000.

This would be achieved via patching the Image class:

```python
from PIL import Image
Image.MAX_IMAGE_PIXELS = 25_000_000
```

Note that Pillow actually permits images up to 2x the declared `MAX_IMAGE_PIXELS` value (i.e. 50,000,000 in this case). (A `DecompressionBombWarning` is raised between 1x and 2x the max, but is not surfaced in NetBox.) So, the effective ceiling is 50 million pixels (roughly 7000x7000), which should be plenty for practical uses.

### Use case

Reducing the maximum image size mitigates potential "decompression bomb" attacks, wherein a malicious image of relatively small file size consumes an inordinate amount of memory when rendered by the Pillow library; e.g. a 68-byte PNG declaring dimensions of 20000×20000 (400M pixels).

### Database changes

N/A

### External dependencies

N/A

jeremystretch:
### Closes: #22629

Set `Image.MAX_IMAGE_PIXELS` to 25,000,000, effecting a maximum size of 50M pixels for uploaded images.

### DIFF
--- a/netbox/core/apps.py
+++ b/netbox/core/apps.py
@@ -4,6 +4,7 @@ from django.core.cache import cache
 from django.db import models
 from django.db.migrations.operations import AlterModelOptions
 from django.utils.translation import gettext as _
+from PIL import Image
 
 from core.events import *
 from netbox.events import EVENT_TYPE_KIND_DANGER, EVENT_TYPE_KIND_SUCCESS, EVENT_TYPE_KIND_WARNING, EventType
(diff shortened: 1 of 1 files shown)


jeremystretch:

pheus (approved):

PR #22633, opened 2026-07-07
Fixes #22632: Fix Object Permission checks for proxy models

### LINKED ISSUE #22632: Object-level permission check raises `ValueError` for cross-app proxy models
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.4

### Python Version

3.12

### Steps to Reproduce

1. Create a `ManagedFile` which is also exposed through the `extras.ScriptModule` proxy model:
   ```python
   from core.choices import ManagedFileRootPathChoices
   from core.models import ManagedFile

   managed_file = ManagedFile.objects.create(
       file_root=ManagedFileRootPathChoices.SCRIPTS,
       file_path="proxy_permission_test.py",
   )

   script_module = ScriptModule.objects.get(pk=managed_file.pk)
   ```
2. Create a non-superuser and grant it an object permission for the `extras.ScriptModule` proxy model:
   ```python
   from users.models import ObjectPermission, User

   user = User.objects.create(username="proxy-test-user")

   object_permission = ObjectPermission.objects.create(
       name="Can change script modules",
       actions=["change"],
       constraints={},
   )
   object_permission.users.add(user)
   object_permission.object_types.add(
       ObjectType.objects.get_for_model(ScriptModule, for_concrete_model=False)
   )
   ```
3. Check the object-level permission against the `ScriptModule` instance:

   ```python
   user.has_perm("extras.change_scriptmodule", script_module)
   ```

### Expected Behavior

The permission check should complete without raising an exception.

Because the user has an applicable object permission for `extras.ScriptModule`, the call above should return `True`.

If the permission/object pair is invalid, or if the object does not match the permission constraints, the check should return `False` rather than raising an unhandled exception.

### Observed Behavior

The permission check raises a `ValueError`:

```text
ValueError: Invalid permission extras.change_scriptmodule for model <class 'core.models.files.ManagedFile'>
```

This appears to happen because `extras.ScriptModule` is a proxy model over `core.ManagedFile`.

During object-level permission evaluation, NetBox currently compares the permission label from the requested permission, `extras.scriptmodule`, with the object's concrete model label, `core.managedfile`. For a cross-app proxy model, those labels do not match even though the permission is valid for the proxy model.

This can surface as an HTTP 500 if a DRF viewset performs its normal object-level permission check against a cross-app proxy model instance.

pheus:
### Closes: #22632

This fixes object-level permission checks for proxy models, including cross-app proxy models such as `extras.ScriptModule` over `core.ManagedFile`.

Previously, `ObjectPermissionMixin.has_perm()` compared the requested permission's app/model label directly with the object's concrete model label. For proxy models, the permission can correctly refer to the proxy model while the object resolves to a different concrete model. In that case, the sanity check raised a `ValueError`, which could surface as an unexpected HTTP 500 during normal permission evaluation.

The updated logic resolves the model named by the permission, verifies that it shares the same concrete model as the object, and evaluates object permission constraints through the permission model's manager. This keeps proxy model permissions valid while still denying permissions that do not apply to the given object.

Invalid model references or mismatched permission/object pairs now return `False` instead of raising an exception. This keeps the authorization path fail-closed: the permission is denied gracefully rather than turning a bad permission string, stale plugin reference, or invalid object pairing into a server error. A warning is logged for nonexistent models, and a debug message is logged for model mismatches.

The logger lookup is intentionally kept local to the exceptional branches. `has_perm()` is a hot path, so this avoids doing even a small amount of extra logger-related work when the permission check succeeds normally.

### DIFF
--- a/netbox/netbox/authentication/__init__.py
+++ b/netbox/netbox/authentication/__init__.py
@@ -1,13 +1,13 @@
 import logging
 from collections import defaultdict
 
+from django.apps import apps
 from django.conf import settings
 from django.contrib.auth.backends import ModelBackend
 from django.contrib.auth.backends import RemoteUserBackend as _RemoteUserBackend
 from django.contrib.auth.models import AnonymousUser
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import Q
-from django.utils.translation import gettext_lazy as _
 
 from users.constants import CONSTRAINT_TOKEN_USER
 from users.models import Group, ObjectPermission, User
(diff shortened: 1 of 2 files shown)


pheus:

jeremystretch (approved):
Thanks @pheus!

Issue #22205, opened 2026-05-14
EOS/EOL dates for Device Types


mw-bowman:
### NetBox version

4.5.7

### Feature type

Data model extension

### Proposed functionality

Something that's come up during the rollout of Netbox in our environment is the concept of lifecycle management, and how Netbox might be useful in that capacity. It occurs to me that it would be enormously useful if, as part of creating a device type for a given model of switch or firewall, you could enter an ISO 8601 format End of Support/End of Life date.

This would allow us to then filter our device list by this field. For example, you could make a custom filter that shows all devices with a device type with a defined EOS/EOL date that falls within the next six months. This gives you an easy way to see what device replacements need to be planned for the near future.

### Use case

Allows Netbox to aid in the management of hardware lifecycling.

### Database changes

I imagine this would just be a new field in the Device Type model.

ITJamie:
Arguably the same should be considered for platform for os's

jeremystretch:
Whatever field(s) we add to DeviceType for this purpose should be added to ModuleType as well.

mw-bowman:
Stoked to see this added as a release milestone, this will be super useful for my team. Agree with both of the above comments re: including this field elsewhere.

mw-bowman:
Probably safest to use "EOL date" or "End of Life Date" for the field name, calling it "EOS date" can be confusing because the term is used somewhat interchangeably for two different things: End of Sale (ie. when it's no longer available to purchase new) and End of Support (ie. end of life/no longer patched by the manufacturer). The date we care about in almost all cases is the latter one, so calling it EOL is less ambiguous.

PR #22634, opened 2026-07-07
#22205 - Add EOL to DeviceType, ModuleType

arthanson:
### Closes: #22205 

Adds an **End of Life** date field to `DeviceType` and `ModuleType` to support hardware lifecycle management. The field is optional and supports date-range filtering (`end_of_life__gte` / `end_of_life__lte`), allowing users to identify device/module types approaching or past their manufacturer support horizon.

Wired up across the model, migration, REST API, GraphQL, forms (create/edit, bulk edit, bulk import, filter), filterset, tables, detail panels, and docs, with test coverage added for the API, filtersets, and views.


arthanson:

arthanson:
Fixed Claude issues it brought up.

bctiemann (approved):

PR #22637, opened 2026-07-08
Closes #18159: Expose snapshots to event rule condition evaluation

### LINKED ISSUE #18159: Add more Event Rule Conditions
### NetBox version

v4.1.7

### Feature type

### Triage priority

### Proposed functionality

The Event Rule condition only checks the objects data, not the value it had before. https://github.com/netbox-community/netbox/blob/develop/netbox/extras/events.py#L90. This proposal could save a lot of unnecessary Webhooks or Scripts, if the data, that the condition checks doesn't change.

### Use case

I want to create a device 1x in an external application, if the device has status active.
1. I create a device with status Planned.
2. I update the device and set the status to active.
3. I update the device and give it a description.
4. I Update the device and set the status to Offline.

I create an Event Rule, `on update` send a Webhook if the status of the device is active.
My Event Rule condition states:
```json
{
    "attr": "status.value",
    "value": "active"
}
```

The event hook is sent on step 2 & 3. But on step 3 the webhook is executed unnecessary because the Device was created in step 2 in the external system.

Possible new condition syntax:
```json
{
  "and": [
    {
      "attr": "status.value",
      "value": "active"
    },
    {
      "attr": "snapshots.prechange.status.value",
      "value": "snapshots.postchange.status.value",
      "op": "eq",
      "negate": true
    }
  ]
}
```
or
```json
{
    "attr": "status.value",
    "value": "active",
    "changed": true
}
```
This rule should check if the status is active, and check if the value of the status has changed. Only if the status has been changed and set to active, the webhook is executed.

### Database changes

None

### External dependencies

None

bctiemann:
## Summary

Event rule conditions previously evaluated only the current (post-change) REST API representation of the object. Snapshots — the pre- and post-change data captured at event time — were present in the event queue but never exposed to condition logic, making it impossible to express rules like "fire only when status *changes* to active."

This PR adds two complementary mechanisms:

**1. `changed` / `unchanged` operators**

New snapshot-comparison operators that check whether an attribute's value differs between the prechange and postchange snapshots. No `value` key is accepted.

```json
{
  "and": [
    { "attr": "status.value", "value": "active" },
    { "attr": "status", "op": "changed" }
  ]
}
```

This is the canonical solution for the issue's use case: the webhook fires only on the transition to active, not on subsequent updates that leave status unchanged.

**2. Direct snapshot path access**

Snapshots are merged into the condition evaluation context, so any existing operator can reference pre/post values via `snapshots.prechange.<attr>` and `snapshots.postchange.<attr>` dot-paths.

```json
{ "attr": "snapshots.prechange.status", "value": "planned" }
```

> **Note on serialization format:** Snapshot data uses the model serializer format (raw field values), not the REST API format. Choice fields like `status` appear as `"active"` in snapshots, not `{"value": "active", "label": "Active"}`. Use `attr: "status"` — not `"status.value"` — when referencing snapshot attributes. The `changed`/`unchanged` operators compare the same format on both sides, so they are unaffected by this distinction.

## Changes

| File | Change |
|---|---|
| `extras/conditions.py` | Add `CHANGED`/`UNCHANGED` operators; `_MISSING` sentinel; make `value` optional; `_resolve_snapshot_attr()` helper; `eval_changed()`/`eval_unchanged()` methods |
| `extras/events.py` | Merge snapshots into condition context in `process_event_rules()` |
| `extras/tests/test_conditions.py` | 16 new tests covering all new functionality and edge cases |
| `extras/tests/test_event_rules.py` | 1 new integration test exercising the full `process_event_rules()` path |
| `docs/reference/conditions.md` | Document new operators, snapshot path syntax, format caveat, and availability note |

## Comparison with community PR

A community implementation exists at https://github.com/kenancakir/netbox/pull/1/changes. Key differences from this PR:

| | Community PR | This PR |
|---|---|---|
| `changed`/`unchanged` operators | ✓ | ✓ |
| Snapshot access syntax | `prechange.status` (intercepts path prefix) | `snapshots.prechange.status` (transparent path merge) |
| `ConditionSet.eval()` signature | Modified (passes snapshots arg through chain) | Unchanged |
| Implementation complexity | Larger (routing logic, signature changes) | Minimal (one merge in `process_event_rules`, two eval methods) |
| Format inconsistency | Unclear | Documented; `changed`/`unchanged` are immune |
| Unit tests added | 0 | 17 |

The main tradeoff: `prechange.status` is marginally more ergonomic than `snapshots.prechange.status`, but requires special-casing the path prefix throughout the eval chain. Our approach falls out naturally from the existing dot-path traversal with no routing logic needed.

### DIFF
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -78,6 +78,13 @@ class Condition:
                 raise ValueError(_(
                     "The '{op}' operator compares snapshots and does not accept a value."
                 ).format(op=op))
+            if attr.startswith('snapshots.'):
+                raise ValueError(_(
+                    "The '{op}' operator resolves '{attr}' within each snapshot dict, not the "
+                    "top-level condition context. Use the bare attribute name (e.g. 'status') "
+                    "rather than a snapshot path (e.g. 'snapshots.prechange.status'), which is "
+                    "only valid with standard operators."
+                ).format(op=op, attr=attr))
             self.value = _MISSING
         else:
             if value is _MISSING:
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -131,7 +131,7 @@ class Condition:
                 else:
                     obj = operator.getitem(obj or {}, key)
             return obj
-        except KeyError:
+        except (KeyError, TypeError):
             return _MISSING
 
     def eval(self, data):
@@ -196,8 +196,11 @@ class Condition:
     # Snapshot comparison operators
     # These resolve self.attr in both the prechange and postchange snapshots and
     # compare the resulting values.  _MISSING is used when a snapshot is absent
-    # or does not contain the attribute, so that two absent values compare equal
-    # (nothing changed) and a present value differs from an absent one (changed).
+    # or does not contain the attribute.
+    #
+    # Fail-closed semantics:
+    #   changed:   False when attr is absent from both snapshots (field never existed)
+    #   unchanged: False when attr is absent from both snapshots (avoids silent pass on typos)
 
     def eval_changed(self, snapshots):
         pre = self._resolve_snapshot_attr(snapshots.get('prechange'))
@@ -207,6 +210,8 @@ class Condition:
     def eval_unchanged(self, snapshots):
         pre = self._resolve_snapshot_attr(snapshots.get('prechange'))
         post = self._resolve_snapshot_attr(snapshots.get('postchange'))
+        if pre is _MISSING and post is _MISSING:
+            return False
         return pre == post
 
 
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -133,7 +133,7 @@ class Condition:
         """
         if self.op in self.SNAPSHOT_OPERATORS:
             snapshots = data.get('snapshots') if isinstance(data, dict) else None
-            if not snapshots:
+            if snapshots is None:
                 raise InvalidCondition(
                     f"No snapshot data available for '{self.op}' operator. "
                     f"Snapshot operators are only meaningful on update and delete events."
@@ -141,10 +141,7 @@ class Condition:
             result = self.eval_func(snapshots)
             return not result if self.negate else result
 
-        try:
-            value = self._resolve_attr(data)
-        except InvalidCondition:
-            raise
+        value = self._resolve_attr(data)
         try:
             result = self.eval_func(value)
         except TypeError as e:
--- a/netbox/extras/conditions.py
+++ b/netbox/extras/conditions.py
@@ -103,7 +103,8 @@ class Condition:
     def _resolve_attr(self, data):
         """
         Walk self.attr as a dotted key path through data. Raises InvalidCondition on
-        missing keys.
+        missing keys, or when an intermediate value can't be indexed by key (e.g. a
+        REST API-style path like 'status.value' applied to a raw snapshot value).
         """
         def _get(obj, key):
             if isinstance(obj, list):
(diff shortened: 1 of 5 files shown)


bctiemann:

bctiemann:
Addressed items 2 and 3, and also added a test for the validation on 3. Fixed the duplicate site names in the fixtures as well.

netbox/extras/conditions.py
```diff
@@ -55,35 +67,88 @@ class Condition:
...
+        try:
+            obj = snapshot
+            for key in self.attr.split('.'):
+                if isinstance(obj, list):
+                    obj = [operator.getitem(item or {}, key) for item in obj]
+                else:
+                    obj = operator.getitem(obj or {}, key)
+            return obj
+        except KeyError:
+            return _MISSING
```
pheus, line 138: `_resolve_snapshot_attr()` currently only catches `KeyError`, but snapshot paths can also fail with `TypeError` when a path traverses through a scalar value.

For example, because snapshot choice fields are serialized as raw strings, a rule like this is an easy mistake to make:

`{"attr": "status.value", "op": "changed"}`

That will try to evaluate something equivalent to `"active"["value"]`, which raises `TypeError`. Since this is not converted into `InvalidCondition` or treated as a missing path, it can bubble out of event rule processing rather than simply making the condition evaluate false.

Could we make this fail closed, either by treating `TypeError` as `_MISSING` in `_resolve_snapshot_attr()` or by converting it to `InvalidCondition`? A regression test with a snapshot condition like `status.value changed` would be helpful too.

netbox/extras/conditions.py
```diff
@@ -128,6 +193,22 @@ def eval_contains(self, value):
...
+
+    def eval_changed(self, snapshots):
+        pre = self._resolve_snapshot_attr(snapshots.get('prechange'))
+        post = self._resolve_snapshot_attr(snapshots.get('postchange'))
+        return pre != post
+
+    def eval_unchanged(self, snapshots):
+        pre = self._resolve_snapshot_attr(snapshots.get('prechange'))
+        post = self._resolve_snapshot_attr(snapshots.get('postchange'))
+        return pre == post
```
pheus, line 218: `eval_unchanged()` currently returns `True` when the attribute is absent from both snapshots, because both sides resolve to the shared `_MISSING` sentinel.

That means a typo like this can unexpectedly pass:

`{"attr": "statsu", "op": "unchanged"}`

I think `unchanged` should probably only return true when the attribute was actually resolved, for example by requiring both `pre` and `post` to be different from `_MISSING` before comparing them. If the current behavior is intentional, could we document it explicitly and add a test for the typo/missing-field case? As written, this feels surprising and less fail-closed than the existing condition behavior.

pheus (changes requested):
Thanks for working on this! The overall approach looks solid to me.

I think there are a few edge cases worth tightening before merge, mostly around fail-closed behavior for invalid snapshot paths.

bctiemann:
Addressed both suggestions. Thanks!

pheus (changes requested):
Thanks for the follow-up!

I think there is one closely related edge case left before merge: direct snapshot paths using standard operators still go through `_resolve_attr()`, not `_resolve_snapshot_attr()`.

For example, this is an easy mistake because REST API choice fields normally use `.value`, while snapshots store raw strings:

```json
{
  "attr": "snapshots.prechange.status.value",
  "value": "planned"
}
```

With snapshot data like:

```python
{
    "snapshots": {
        "prechange": {"status": "planned"},
        "postchange": {"status": "active"},
    }
}
```

`_resolve_attr()` will traverse into `"planned"` and call `operator.getitem("planned", "value")`, which raises `TypeError`. Since `_resolve_attr()` only catches `KeyError`, and `Condition.eval()` calls `_resolve_attr()` before the `TypeError` guard around `eval_func()`, this can still bubble out of event rule evaluation instead of failing closed.

Could we also catch `TypeError` in `_resolve_attr()` and convert it to `InvalidCondition`, similar to the snapshot-operator helper?

bctiemann:
Good call, implemented (including tests). Thanks!

pheus (approved):
Thanks for the follow-ups! This looks good to me!

PR #22641, opened 2026-07-08
Closes #18645: Add support for bulk importing multi-termination cables

### LINKED ISSUE #18645: Bulk import of multi-termination cables
### NetBox version

v4.2.2

### Feature type

New functionality

### Proposed functionality

Please add the feature to be able to bulk import cables that have multiple termination points on one side (a or b).

### Use case

Will provide clarity and no need to use work-around with front and rear ports, which cause confusion when traced (since still being shown/treated as non split cable). Useful for data center scale deployments which use split cables.

### Database changes

_No response_

### External dependencies

_No response_

pheus:
### Closes: #18645

This adds support for importing cables with multiple terminations on either side via CSV, JSON, and YAML bulk import.

Cable import now accepts comma-separated values for device, power panel, and termination name fields. Each side can specify either one parent for all terminations or one parent per termination name. Submitted order is preserved so breakout cable profiles can assign connector positions correctly.

This also updates the relevant cable parent fields from `CSVModelChoiceField` to `CSVModelMultipleChoiceField`, refreshes their help text, and adds validation for duplicate terminations, empty termination names, parent/name count mismatches, already-connected terminations, and ambiguous parent lookups.

### DIFF
--- a/netbox/dcim/forms/bulk_import.py
+++ b/netbox/dcim/forms/bulk_import.py
@@ -1,7 +1,8 @@
 from django import forms
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres.forms.array import SimpleArrayField
-from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
+from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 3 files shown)


pheus:

pheus:
@claude /review

I agree with the `mark_safe()` point and I've updated the code.

For `_resolve_side_parent_objects()`, I’d like to keep the current explicit lookup. The extra queries are intentional here: the cleaned `CSVModelMultipleChoiceField` value is a queryset, does not preserve submitted order, deduplicates repeated values, and does not give us the same clear `MultipleObjectsReturned` handling for non-unique `to_field_name` values. Since cable sides usually have only a small number of terminations, the trade-off seems acceptable.

The whitespace behavior in `CSVModelMultipleChoiceField` is a valid broader UX issue, but I’d prefer not to change that shared utility field as part of this PR.

I’ll also add a test covering multi-parent import with `side_x_site` queryset filtering, to make sure an otherwise valid parent name outside the selected site is rejected on the parent field.

arthanson (approved):

PR #22645, opened 2026-07-08
#22644 Add ObjectChange to PortMapping

### LINKED ISSUE #22644: Front/rear port mappings are not recorded in the changelog
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.4

### Python Version

3.12

### Steps to Reproduce

1. Create a device
2. add a rear port to the device "rp1"
3. add a front port to the device "fp1" and map it to the device "rp1"
4. Go to look at the Change Log

### Expected Behavior

The mapping of fp1 to rp1 should appear in the changelog.

### Observed Behavior

There is nothing in the changelog related to the port assignment.  The ObjectChange for fp1 does not contain any port mappings and there is also no separate changelog for the port mapping.

arthanson:
### Closes: #22644

PortMapping and PortTemplateMapping now inherit ChangeLoggingMixin so their create/update/delete are recorded like any other object. 

Because the mixin sits ahead of PortMappingBase in the MRO, each model re-declares class Meta(PortMappingBase.Meta) to preserve the inherited unique constraints.

Both write paths for front/rear port mappings — FrontPortFormMixin._save_m2m (UI/forms) and PortSerializer.create/update (REST API) — previously deleted all of a port's mappings and recreated them on every save. Now that PortMapping/PortTemplateMapping are change-logged, that pattern minted a new PK for every mapping on every save, producing:

- Changelog churn - re-saving a front port without touching its wiring emitted a DELETE + CREATE for every mapping.
- Broken branch merges - because the PK is the object's identity for changelog/branching, two branches (or a branch and main) that merely re-saved the same front port replayed DELETE(old-pk) + CREATE(new-pk) against rows whose PKs had already changed, colliding on the (front_port, front_port_position) unique constraint and failing the merge - even though neither side changed the wiring.

Both paths now call a shared dcim.utils.reconcile_port_mappings() that writes only the difference: unchanged mappings keep their PK and emit no changelog entry, and changed/removed rows are deleted before replacements are created (keeping single-request swaps safe against transient constraint violations). No-op re-saves become no-ops; genuine concurrent edits to the same slot remain a real merge conflict.


netbox/dcim/migrations/0238_add_portmapping_objectchange.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 6.0.6 on 2026-07-08 20:46
+
```
jeremystretch:
```suggestion
```

netbox/dcim/models/device_components.py
```diff
@@ -1216,6 +1224,12 @@ class PortMapping(PortMappingBase):
         related_name='mappings',
     )
 
+    class Meta(PortMappingBase.Meta):
+        # Inherit the unique constraints from PortMappingBase.Meta. An explicit Meta is
+        # required now that ChangeLoggingMixin precedes PortMappingBase in the MRO;
+        # without it, ChangeLoggingMixin's (constraint-less) Meta would shadow the base's.
```
jeremystretch:
```suggestion
        # Inherit the unique constraints from PortMappingBase.Meta.
```

netbox/dcim/models/device_components.py
```diff
@@ -1196,9 +1197,16 @@ def connected_endpoints(self):
...
+class PortMapping(ChangeLoggingMixin, PortMappingBase):
     """
     Maps a FrontPort & position to a RearPort & position.
+
+    Change logging is enabled (via ChangeLoggingMixin) so that creating, editing, or
+    removing a mapping is recorded in the object change log. The model remains "private"
+    (_netbox_private, inherited from PortMappingBase): it has no public API/URL of its own
+    and is still treated as an implementation detail of its FrontPort/RearPort. Recording
+    an ObjectChange ensures the mapping is captured in the audit trail and can be replayed
+    by consumers of the change log (e.g. branch merges). See netbox-branching issue #611.
```
jeremystretch: Excessive commentary
```suggestion
```

netbox/dcim/models/device_components.py
```diff
@@ -1196,9 +1197,16 @@ def connected_endpoints(self):
 # Pass-through ports
 #
 
-class PortMapping(PortMappingBase):
+class PortMapping(ChangeLoggingMixin, PortMappingBase):
```
jeremystretch, line 1200: Presumably this is needed for PortTemplateMapping as well?

jeremystretch (changes requested):

arthanson:

arthanson:

arthanson:
@jeremystretch addressed review comments, also noticed a follow on issue that all port mappings were getting deleted and re-applied when the port was updated which would cause excessive ObjectChange records, there was a comment in the code to find a better solution, implemented `_reconcile_mappings` which will only update port mappings that have changed.

netbox/dcim/migrations/0238_add_portmapping_objectchange.py
```diff
@@ -0,0 +1,31 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("dcim", "0237_module_remove_local_context_data"),
```
jeremystretch: We need to renumber this migration to `0239` to avoid a conflict in `main`.
```suggestion
        ("dcim", "0238_alter_cable__abs_length"),
```

jeremystretch (changes requested):
LGTM, just needs a migrations tweak.

jeremystretch (approved):

PR #22646, opened 2026-07-08
Closes #20054: Return per-object error details for failed bulk operations

### LINKED ISSUE #20054: Better BULK operation error response, providing offending object -> error correlation
### NetBox version

v4.3.4

### Feature type

New functionality

### Proposed functionality

- For bulk list-endpoint updates (e.g., PATCH to `.../api/dcim/interfaces/`), return a structured error payload that correlates each submitted object with its validation outcome.
- Return per-object validation errors so clients can identify and retry only the failing items.
- Response shape (example) when any item fails:
  ```json
  {
    "detail": "One or more objects failed validation",
    "results": [
      { "id": 1, "status": "ok" },
      {
        "id": 2,
        "status": "error",
        "errors": {
          "description": ["Ensure this field has no more than 200 characters."]
        }
      },
      { "id": 3, "status": "ok" }
    ],
    "succeeded": [1, 3],
    "failed": [2]
  }
  ```
- Success case (all pass) remains unchanged (200 with updated objects).
- Error status remains 400 to preserve backward compatibility; the new body adds per-object detail.
- Optional enhancement (non-breaking, future-friendly): support an HTTP 207 Multi-Status alternative or a query flag (e.g., `?verbose_errors=true`) to opt into the detailed error shape.

### Use case

- When performing large bulk updates (e.g., 100–1000 objects), a single invalid item currently returns an aggregate field error without indicating which object(s) failed (e.g., `{"description": ["Ensure this field has no more than 200 characters."]}`).
- Clients must then bisect or iterate requests to isolate the bad object(s), incurring extra API calls and complexity.
- With per-object error correlation, clients can:
  - Immediately identify the offending ID(s)
  - Drop or correct invalid items
  - Retry only the problematic subset
- This significantly reduces operational time and complexity for workflows that rely on bulk operations (inventory sync, onboarding, mass edits).

### Database changes

- None. This is an API response enhancement in the bulk update path (e.g., `BulkUpdateModelMixin`) and serializers; no model or schema changes required.

### External dependencies

- None

bctiemann:
## Summary

Bulk write endpoints now return a structured, per-object error payload instead of stopping at the first failure. The entire batch is still rolled back atomically when any object fails, but the response identifies exactly which objects failed and why — enabling clients to correct and retry the specific failing items without guessing.

### Response shape (error case)

```json
{
  "detail": "1 of 3 objects failed validation.",
  "results": [
    {"id": 1},
    {"id": 2, "errors": {"name": ["This field is required."]}},
    {"id": 3}
  ]
}
```

For bulk creates via `SequentialBulkCreatesMixin` the correlator is `"index"` (zero-based position in the request list, since no IDs exist yet). For bulk deletes the status code remains 409.

### Operations changed

| Operation | Mixin | Success | Partial/full failure |
|---|---|---|---|
| `PATCH` (bulk update) | `BulkUpdateModelMixin` | 200 – unchanged | 400 with `results` |
| `POST` (bulk create) | `SequentialBulkCreatesMixin` | 201 – unchanged | 400 with `results` |
| `DELETE` (bulk delete) | `BulkDestroyModelMixin` | 204 – unchanged | 409 with `results` |

**Breaking change**: the bulk delete error response body has changed from a bare `{"detail": "..."}` string to the structured format above.

### Implementation notes

- `perform_bulk_update`: two-pass — validate all objects first (no writes), then update all if all pass; `set_rollback(True)` on any failure
- `SequentialBulkCreatesMixin.create`: sequential validate-and-create (preserves rack-space and other cross-object validators), collect errors, `set_rollback(True)` on any failure
- `BulkDestroyModelMixin.perform_bulk_destroy`: catches `ProtectedError`/`RestrictedError` per object; `set_rollback(True)` rolls back any partial deletes
- `rollback()` is always called by exiting the `with transaction.atomic()` block normally (never via `return`-from-inside), so `TransactionManagementError` is not raised in production

## Test plan

- [x] `SiteTestCase.test_bulk_update_objects_validation_error` — mixed valid/invalid PATCH returns 400 with per-object results; object 0 passes validation, object 1 fails
- [x] `SiteTestCase.test_bulk_delete_objects_protected` — bulk DELETE where one site has a dependent Device; 409 response with per-object results, both sites still exist (rollback verified)
- [x] `DeviceTestCase.test_bulk_create_objects_validation_error` — empty objects via `SequentialBulkCreatesMixin`; 400 with index-correlated results, no objects created
- [x] Existing `DeviceTestCase.test_rack_fit` — sequential rack-space validation preserved (overlapping bulk create → 400)
- [x] Full `dcim.tests.test_api` suite — 1146 tests, 44 skipped, 0 failures
- [x] `ipam.tests.test_api` + `extras.tests.test_api` — pass without changes

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -389,7 +389,7 @@ class BulkDestroyModelMixin:
                         'errors': {
                             '__all__': _(
                                 'Unable to delete: {n} dependent object(s) prevent deletion.'
-                            ).format(n=n),
+                            ).format(n=len(protected)),
                         },
                     })
             if errors:
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -183,9 +183,9 @@ class SequentialBulkCreatesMixin:
                     # All creates are rolled back together if any item in the batch fails.
                     self.perform_create(serializer)
                     return_data.append(serializer.data)
-                    results.append({'index': i, 'status': 'ok'})
+                    results.append({'index': i})
                 else:
-                    results.append({'index': i, 'status': 'error', 'errors': serializer.errors})
+                    results.append({'index': i, 'errors': serializer.errors})
                     error_count += 1
 
             if error_count:
@@ -288,9 +288,9 @@ class BulkUpdateModelMixin:
                 if serializer.is_valid():
                     self.perform_update(serializer)
                     updated_pks.append(obj.pk)
-                    results.append({'id': obj.pk, 'status': 'ok'})
+                    results.append({'id': obj.pk})
                 else:
-                    results.append({'id': obj.pk, 'status': 'error', 'errors': serializer.errors})
+                    results.append({'id': obj.pk, 'errors': serializer.errors})
                     error_count += 1
             if error_count:
                 transaction.set_rollback(True)
@@ -378,7 +378,7 @@ class BulkDestroyModelMixin:
                 pk = obj.pk  # Django sets obj.pk = None after deletion; capture it first
                 try:
                     self.perform_destroy(obj)
-                    results.append({'id': pk, 'status': 'ok'})
+                    results.append({'id': pk})
                 except (ProtectedError, RestrictedError) as e:
                     protected = list(
                         e.protected_objects if isinstance(e, ProtectedError) else e.restricted_objects
@@ -388,7 +388,6 @@ class BulkDestroyModelMixin:
                     # the caller may not have permission to view.
                     results.append({
                         'id': pk,
-                        'status': 'error',
                         'errors': {
                             'detail': _(
                                 'Unable to delete: {n} dependent object(s) prevent deletion.'
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -247,6 +247,8 @@ class BulkUpdateModelMixin:
 
         object_pks, results = self.perform_bulk_update(qs, update_data, partial=partial)
 
+        # perform_bulk_update returns an empty list on full success; non-empty means at least one
+        # object failed validation and the full results list (with per-object status) is populated.
         if results:
             failed_count = sum(1 for r in results if r['status'] == 'error')
             return Response(
@@ -380,7 +382,7 @@ class BulkDestroyModelMixin:
                     if n > 10:
                         objects_str += f', and {n - 10} more'
                     results.append({
-                        'id': obj.pk,
+                        'id': pk,
                         'status': 'error',
                         'errors': {'detail': f'Unable to delete. {n} dependent object(s): {objects_str}'},
                     })
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -177,6 +177,9 @@ class SequentialBulkCreatesMixin:
             for i, data in enumerate(request.data):
                 serializer = self.get_serializer(data=data)
                 if serializer.is_valid():
+                    # Provisionally create even when a prior item failed, so subsequent
+                    # cross-object validators (e.g. rack space checks) see a realistic state.
+                    # All creates are rolled back together if any item in the batch fails.
                     self.perform_create(serializer)
                     return_data.append(serializer.data)
                     results.append({'index': i, 'status': 'ok'})
@@ -190,7 +193,10 @@ class SequentialBulkCreatesMixin:
             failed_count = sum(1 for r in results if r['status'] == 'error')
             return Response(
                 {
-                    'detail': f'{failed_count} of {len(results)} objects failed validation.',
+                    'detail': _('{failed_count} of {total} objects failed validation.').format(
+                        failed_count=failed_count,
+                        total=len(results),
+                    ),
                     'results': results,
                 },
                 status=status.HTTP_400_BAD_REQUEST,
@@ -253,7 +259,10 @@ class BulkUpdateModelMixin:
             failed_count = sum(1 for r in results if r['status'] == 'error')
             return Response(
                 {
-                    'detail': f'{failed_count} of {len(results)} objects failed validation.',
+                    'detail': _('{failed_count} of {total} objects failed validation.').format(
+                        failed_count=failed_count,
+                        total=len(results),
+                    ),
                     'results': results,
                 },
                 status=status.HTTP_400_BAD_REQUEST,
@@ -349,11 +358,14 @@ class BulkDestroyModelMixin:
 
         results = self.perform_bulk_destroy(qs, changelog_messages)
 
-        if results and any(r['status'] == 'error' for r in results):
+        if any(r['status'] == 'error' for r in results):
             failed_count = sum(1 for r in results if r['status'] == 'error')
             return Response(
                 {
-                    'detail': f'{failed_count} of {len(results)} objects could not be deleted.',
+                    'detail': _('{failed_count} of {total} objects could not be deleted.').format(
+                        failed_count=failed_count,
+                        total=len(results),
+                    ),
                     'results': results,
                 },
                 status=status.HTTP_409_CONFLICT,
@@ -378,13 +390,16 @@ class BulkDestroyModelMixin:
                         e.protected_objects if isinstance(e, ProtectedError) else e.restricted_objects
                     )
                     n = len(protected)
-                    objects_str = ', '.join(f'{o} ({o.pk})' for o in protected[:10])
-                    if n > 10:
-                        objects_str += f', and {n - 10} more'
+                    # Report only the count — not names or PKs — to avoid exposing objects
+                    # the caller may not have permission to view.
                     results.append({
                         'id': pk,
                         'status': 'error',
-                        'errors': {'detail': f'Unable to delete. {n} dependent object(s): {objects_str}'},
+                        'errors': {
+                            'detail': _(
+                                'Unable to delete: {n} dependent object(s) prevent deletion.'
+                            ).format(n=n),
+                        },
                     })
             if any(r['status'] == 'error' for r in results):
                 transaction.set_rollback(True)
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -378,7 +378,6 @@ class BulkDestroyModelMixin:
                     protected = list(
                         e.protected_objects if isinstance(e, ProtectedError) else e.restricted_objects
                     )
-                    n = len(protected)
                     # Report only the count, not names or PKs, to keep each per-object error
                     # entry small in a batch response. Note: the single-object delete endpoint
                     # (NetBoxModelViewSet.dispatch()) does include names and PKs of dependent
--- a/netbox/netbox/api/viewsets/mixins.py
+++ b/netbox/netbox/api/viewsets/mixins.py
@@ -167,14 +167,14 @@ class SequentialBulkCreatesMixin:
 
         # Create objects sequentially so each validation sees the state left by prior creates
         # (e.g. rack space checks). Collect per-object errors instead of failing on the first.
-        results = []
+        errors = []
         return_data = []
-        error_count = 0
         with transaction.atomic(using=router.db_for_write(self.queryset.model)):
             if not isinstance(request.data, list):
                 # Creating a single object
                 return super().create(request, *args, **kwargs)
 
+            total = len(request.data)
             for i, data in enumerate(request.data):
                 serializer = self.get_serializer(data=data)
                 if serializer.is_valid():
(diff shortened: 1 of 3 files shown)


bctiemann:

bctiemann:
Addressed items 1, 3, 4, 5. Item 2 is fine as-is; "ok" is better semantics than something like "valid" which just gets confusing.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -164,21 +165,45 @@ def create(self, request, *args, **kwargs):
         if (response := handle_background(request, 'create')) is not None:
             return response
 
-        with transaction.atomic(using=router.db_for_write(self.queryset.model)):
-            if not isinstance(request.data, list):
-                # Creating a single object
-                return super().create(request, *args, **kwargs)
+        if not isinstance(request.data, list):
```
jeremystretch: Are we sure this is safe to remove from the atomic transaction? IIRC it may be necessary in certain situations for signal handlers. Would prefer not to alter this unless we can be 100% confident doing so is safe.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -164,21 +165,45 @@ def create(self, request, *args, **kwargs):
...
-                return_data.append(serializer.data)
-
-            headers = self.get_success_headers(serializer.data)
+                if serializer.is_valid():
+                    # Provisionally create even when a prior item failed, so subsequent
+                    # cross-object validators (e.g. rack space checks) see a realistic state.
+                    # All creates are rolled back together if any item in the batch fails.
+                    self.perform_create(serializer)
+                    return_data.append(serializer.data)
+                    results.append({'index': i, 'status': 'ok'})
```
jeremystretch: Do we need an explicit `status` key? Can we just infer success when `errors` is empty?

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -164,21 +165,45 @@ def create(self, request, *args, **kwargs):
...
+                else:
+                    results.append({'index': i, 'status': 'error', 'errors': serializer.errors})
+
+            if any(r['status'] == 'error' for r in results):
```
jeremystretch: Rather than iterating through the results list, we could set a counter locally (e.g. `errors = 0`) that gets incremented with each failure. Rollback can then be inferred as needed of `errors > 0`, and we can reuse the counter in the response.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -236,17 +276,31 @@ def bulk_update(self, request, *args, **kwargs):
...
                     obj.snapshot()
                 serializer = self.get_serializer(obj, data=data, partial=partial)
-                serializer.is_valid(raise_exception=True)
-                self.perform_update(serializer)
-                updated_pks.append(obj.pk)
-
-        return updated_pks
+                prepared.append((obj, serializer, serializer.is_valid()))
+
+            if any(not valid for _, _, valid in prepared):
```
jeremystretch: Same thing here: We can use a local counter/flag to track failures. (Also, try to avoid using `_` for throwaways in general as it's typically aliased to `gettext()`; use `__` instead where needed.)

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -236,17 +276,31 @@ def bulk_update(self, request, *args, **kwargs):
...
+                results = [
+                    {'id': obj.pk, 'status': 'error', 'errors': ser.errors} if not valid
+                    else {'id': obj.pk, 'status': 'ok'}
+                    for obj, ser, valid in prepared
+                ]
+                transaction.set_rollback(True)
+            else:
+                # Pass 2: all objects are valid — perform updates
+                for obj, serializer, _ in prepared:
+                    self.perform_update(serializer)
```
jeremystretch, line 286: We need to perform each update prior to validating the next object. Consider for instance changing two sites to the same name: Each will validate independently, but the second one will throw an exception when we try to save it.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -302,18 +356,54 @@ def bulk_destroy(self, request, *args, **kwargs):
...
+                    results.append({
+                        'id': pk,
+                        'status': 'error',
+                        'errors': {
+                            'detail': _(
+                                'Unable to delete: {n} dependent object(s) prevent deletion.'
+                            ).format(n=n),
+                        },
+                    })
+            if any(r['status'] == 'error' for r in results):
```
jeremystretch: Another use case for a counter.

jeremystretch (changes requested):

netbox/netbox/api/viewsets/mixins.py
bctiemann: Single-object `create()` path moved back inside `transaction.atomic()` to preserve signal-handler safety.

netbox/netbox/api/viewsets/mixins.py
bctiemann: `create()` now uses a local `error_count` counter; no more double-iteration of results.

netbox/netbox/api/viewsets/mixins.py
bctiemann: `perform_bulk_update()` uses a local counter; the `_` vs `__` issue is moot because the two-pass prepared list (with its throwaway unpacking) is gone entirely.

netbox/netbox/api/viewsets/mixins.py
bctiemann: `perform_bulk_update()` is now single-pass sequential: each object is validated then saved before moving to the next. Cross-object unique constraint conflicts (e.g., two sites renamed to the same name) are caught at validation time rather than as integrity errors on save.

netbox/netbox/api/viewsets/mixins.py
bctiemann: `perform_bulk_destroy()` uses a local `error_count` counter; callers unpack the new `(results, error_count)` tuple.

netbox/netbox/api/viewsets/mixins.py
bctiemann: Went back and forth on this a bunch but landed on agreeing it's just cleaner at the code level to remove the `status` key, and less prone to maintenance mistakes. Done.

bctiemann:
Updated PR description according to new payload.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -236,17 +274,27 @@ def bulk_update(self, request, *args, **kwargs):
...
                 serializer = self.get_serializer(obj, data=data, partial=partial)
-                serializer.is_valid(raise_exception=True)
-                self.perform_update(serializer)
-                updated_pks.append(obj.pk)
-
-        return updated_pks
+                if serializer.is_valid():
+                    self.perform_update(serializer)
+                    updated_pks.append(obj.pk)
+                    results.append({'id': obj.pk})
```
jeremystretch: With the removal of the explicit `status` field, it probably makes sense to omit successful objects entirely, and return only the objects with errors associated. Similarly, it probably makes sense to rename `results` to `errors`. This also obviates the need for a separate error counter: `len(errors)` is sufficient.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -302,18 +350,54 @@ def bulk_destroy(self, request, *args, **kwargs):
...
+                    protected = list(
+                        e.protected_objects if isinstance(e, ProtectedError) else e.restricted_objects
+                    )
+                    n = len(protected)
+                    # Report only the count — not names or PKs — to avoid exposing objects
+                    # the caller may not have permission to view.
+                    results.append({
+                        'id': pk,
+                        'errors': {
+                            'detail': _(
```
jeremystretch: Can we tweak this a bit to better match the format of errors returned by creates and updates? (These are field-based.)
```suggestion
                            '__all__': _(
```

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -302,18 +350,54 @@ def bulk_destroy(self, request, *args, **kwargs):
...
+                try:
+                    self.perform_destroy(obj)
+                    results.append({'id': pk})
+                except (ProtectedError, RestrictedError) as e:
+                    protected = list(
+                        e.protected_objects if isinstance(e, ProtectedError) else e.restricted_objects
+                    )
+                    n = len(protected)
+                    # Report only the count — not names or PKs — to avoid exposing objects
+                    # the caller may not have permission to view.
```
jeremystretch: This differs from the single-object deletion endpoint, which _does_ convey the specific objects preventing deletion. If this is a genuine security concern, the detail endpoint should be modified to match this behavior. (This would require a separate bug report.) If it's merely a performance concern, the comment should be corrected.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -164,21 +165,46 @@ def create(self, request, *args, **kwargs):
...
-                return_data.append(serializer.data)
-
-            headers = self.get_success_headers(serializer.data)
+                if serializer.is_valid():
+                    # Provisionally create even when a prior item failed, so subsequent
+                    # cross-object validators (e.g. rack space checks) see a realistic state.
+                    # All creates are rolled back together if any item in the batch fails.
+                    self.perform_create(serializer)
+                    return_data.append(serializer.data)
+                    results.append({'index': i})
```
jeremystretch: Suggest omitting this as under `perform_bulk_update()`
```suggestion
```

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -164,21 +165,46 @@ def create(self, request, *args, **kwargs):
...
-            return_data = []
-            for data in request.data:
+            for i, data in enumerate(request.data):
                 serializer = self.get_serializer(data=data)
-                serializer.is_valid(raise_exception=True)
-                self.perform_create(serializer)
-                return_data.append(serializer.data)
-
-            headers = self.get_success_headers(serializer.data)
+                if serializer.is_valid():
```
jeremystretch, line 180: Removing `raise_exception=True` appears to no longer catch certain uniqueness violations. For example, try creating two sites with the same name & slug:
```
POST /api/dcim/sites/
[
    {"name": "Site 1", "slug": "site-1"},
    {"name": "Site 1", "slug": "site-1"}
]
```

On the current release, this returns a clean validation failure:

```json
{
    "error": "duplicate key value violates unique constraint \"dcim_site_name_key\"\nDETAIL:  Key (name)=(Site 1) already exists.",
    "exception": "IntegrityError",
    "netbox_version": "4.6.4",
    "python_version": "3.12.3"
}
```

However, the PR no longer catches the IntegrityError exception:

```
IntegrityError at /api/dcim/sites/
duplicate key value violates unique constraint "dcim_site_name_key"
DETAIL:  Key (name)=(Site 1) already exists.
```

Although it limits the utility of conveying errors for multiple objects in some scenarios, the safer path IMO would be to continue raising exceptions to trigger an immediate failure on the first such error. This improvement is incidental to the PR's primary focus anyway.

jeremystretch (changes requested):

netbox/netbox/api/viewsets/mixins.py
bctiemann: Thanks for flagging this — I reproduced the exact scenario (duplicate name+slug in one POST batch) against both this branch and the pre-PR base commit, using an isolated test DB with `DEBUG=False`.

Both branches return the identical response:

```
STATUS: 500
BODY: {"error": "duplicate key value violates unique constraint \"dcim_site_slug_key\"\nDETAIL:  Key (slug)=(repro-site) already exists.", "exception": "IntegrityError", "netbox_version": "4.6.4", "python_version": "3.12.1"}
```

`CoreMiddleware.process_exception()` catches this via `handle_rest_api_exception()` and returns the structured JSON 500 you saw on "the current release" — this is pre-existing behavior, unrelated to the `raise_exception=True` → `if serializer.is_valid():` change in this PR. The root cause is that `dcim_site`'s uniqueness constraint isn't covered by DRF's automatic `UniqueValidator` (it's presumably enforced only at the DB level), so `serializer.is_valid()` doesn't catch it either before or after this PR — the failure always happens at `.save()` time in both versions.

My guess is the raw traceback you saw locally came from testing with `DEBUG=True` (a common local dev setting), which causes `process_exception()` to explicitly bail out early (`if settings.DEBUG: return None`) and let Django's interactive debug page take over — that would happen identically on `main` too under the same local `DEBUG=True` conditions.

Since this isn't a regression introduced by this PR, I haven't changed anything for this one — let me know if you disagree or want to dig further.

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -302,18 +345,56 @@ def bulk_destroy(self, request, *args, **kwargs):
...
+                    # entry small in a batch response. Note: the single-object delete endpoint
+                    # (NetBoxModelViewSet.dispatch()) does include names and PKs of dependent
+                    # objects, so this is not a hard security boundary — just a narrower
+                    # response shape for the bulk case.
+                    errors.append({
+                        'id': pk,
+                        'errors': {
+                            '__all__': _(
+                                'Unable to delete: {n} dependent object(s) prevent deletion.'
+                            ).format(n=n),
```
jeremystretch:
```suggestion
                            ).format(n=len(protected)),
```

netbox/netbox/api/viewsets/mixins.py
```diff
@@ -302,18 +345,56 @@ def bulk_destroy(self, request, *args, **kwargs):
...
                 obj._changelog_message = changelog_messages.get(obj.pk)
-                self.perform_destroy(obj)
+                pk = obj.pk  # Django sets obj.pk = None after deletion; capture it first
+                try:
+                    self.perform_destroy(obj)
+                except (ProtectedError, RestrictedError) as e:
+                    protected = list(
+                        e.protected_objects if isinstance(e, ProtectedError) else e.restricted_objects
+                    )
+                    n = len(protected)
```
jeremystretch:
```suggestion
```

jeremystretch (approved):
Approved with a tiny suggested cleanup.

PR #22647, opened 2026-07-08
Closes #20972: Add support for channelized subinterfaces

jeremystretch:
### Closes: #20972

- New model fields:
    - Add `channels` and `channel_id` to Interface.
    - Add `parent`, `channels`, and `channel_id` to InterfaceTemplate.
- Add a unique constraint for `(parent, channel_id`) to Interface and InterfaceTemplate.
- Override `set_cable_termination()` and `clear_cable_termination()` on the Interface model to replicate cable assignment to channelized subinterfaces.
- Adapt the cable tracing logic & rendering to support paths terminated to subinterfaces for a cable connected to the physical parent interface.

### DIFF
--- a/docs/models/dcim/interface.md
+++ b/docs/models/dcim/interface.md
@@ -28,10 +28,16 @@ An alternative physical label identifying the interface.
 
 ### Type
 
-The type of interface. Interfaces may be physical or virtual in nature, but only physical interfaces may be connected via cables.
+The type of interface. Interfaces may be physical or virtual in nature, but only physical interfaces may be connected via cables. The generic **channel** type identifies a [channelized subinterface](#channel-id) bound to a parent interface.
 
 !!! note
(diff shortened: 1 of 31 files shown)


jeremystretch:

jeremystretch:
1. I'm going to keep this as-is for now, as it can be argued that it's desirable behavior. We can re-evaluate once we get feedback from the v4.7 beta.
2. Addressed
3. Addressed (by adding `selected_related('cable')`)
4. This seems like a narrow, acceptable risk. Going to leave it for now, as mitigating it would likely require a substantial change to the instantiation mechanism.

jnovinger (approved):
Claude noticed that the stale-channel cleanup in `update_channelized_cable_paths` (`signals.py:218-219`) deletes only paths where the channel is an origin, so it tears down the forward (channel→far) direction but not the reverse (far→channel). As far as I can tell, every normal workflow seems safe: cable deletion goes through `nullify_connected_endpoints`, and a parent reassignment triggers the old-parent rebuild that clears both directions.

I could only strand a reverse path by clearing the parent's cable out-of-band and then re-saving the child, which isn't a user-reachable path.

PR #22650, opened 2026-07-09
Closes: #22544: Add support for updating Custom Script Modules via REST API

### LINKED ISSUE #22544: Provide a REST API method to update/overwrite an existing custom script module
### NetBox version

v4.6.3

### Feature type

New functionality

### Proposed functionality

Provide a supported, idempotent way to update an existing script module via REST. Add PUT/PATCH to the upload viewset (by id or file path) to replace content + re-sync. possibly providing an explicit overwrite=true flag on the POST to perform get-or-update + re-sync.

### Use case

The REST API can create a custom script module (POST /api/extras/scripts/upload/) but offers no supported way to update / overwrite an existing one. 

On NetBox Cloud, with no filesystem or management-command access, REST is the only programmatic deployment path - so there is currently no way to do idempotent, automated (CI/CD) re-deployment of an updated script.

### Database changes

N/A

### External dependencies

N/A

pheus:
### Closes: #22544

This adds `PUT`/`PATCH` support to the script module upload endpoint, allowing an existing custom script module to be replaced in place.

Script modules can now be addressed by either numeric ID or file name. During update, the uploaded file name must match the existing module's `file_path`, which keeps the module identity stable and avoids accidentally replacing one script module with a differently named file.

After a successful replacement, the module is saved normally so its script classes are re-synchronized from the new file content. This allows removed, renamed, or newly added script classes to be reflected without deleting and recreating the script module.

The existing `POST` behavior remains create-only: duplicate uploads are still rejected rather than silently overwriting an existing module. The update path also uses the normal object permission check, now that the proxy-model permission issue has been fixed separately.

### DIFF
--- a/docs/customization/custom-scripts.md
+++ b/docs/customization/custom-scripts.md
@@ -443,6 +443,18 @@ curl -X POST \
 http://netbox/api/extras/scripts/upload/
 ```
 
+### Updating an Uploaded Script
+
+An existing script module can be replaced in place by sending a `multipart/form-data` PUT or PATCH request to the module's detail URL. The module may be identified by its numeric ID or by its file name. The uploaded file name must match the existing module's file path, and the caller must have the `extras.change_scriptmodule` and `core.change_managedfile` permissions. The module's scripts are re-synchronized from the new content.
+
+```no-highlight
+curl -X PUT \
+-H "Authorization: Bearer $TOKEN" \
+-H "Accept: application/json; indent=4" \
+-F "file=@/path/to/myscript.py" \
+http://netbox/api/extras/scripts/upload/myscript.py/
+```
+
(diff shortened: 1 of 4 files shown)


pheus:

pheus:
@claude /review

On 1: fair concern, but restoring the old file would mean reading the whole object on every update (storage can be S3), and it still wouldn't be truly atomic. Went with `transaction.atomic()` around the save instead, so the Script rows can't end up half-synced, plus a regression test for that. Retrying the update heals the rest.

On 2: false positive. `ScriptModule` proxies `ManagedFile`, whose `last_updated` has no `auto_now`, so the manual assignment is doing real work.

bctiemann (approved):

PR #22653, opened 2026-07-09
Closes #22652: Force autoescape=False on ConfigTemplate rendering

### LINKED ISSUE #22652: SandboxedEnvironment in render_jinja2() does not explicitly disable autoescape
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

1. Call `render_jinja2()` (or render a ConfigTemplate) with an `environment_params` dict that includes `autoescape=True` or `autoescape=select_autoescape(['html'])`.
2. Observe that the `SandboxedEnvironment` is constructed with autoescape enabled, which was not the intended behaviour.

### Expected Behavior

`render_jinja2()` renders config templates (network configs, scripts) as plain text. Autoescape should always be `False` regardless of what is passed in `environment_params`. A caller should not be able to enable HTML autoescaping through that dict.

### Observed Behavior

`SandboxedEnvironment(**environment_params)` is called without any post-construction override of `autoescape`. If a caller (or a future feature) passes `autoescape=True` via `environment_params`, the environment will HTML-escape template output. If that output is later rendered in a web context without additional escaping (on the assumption that config templates are plain text), it becomes a latent XSS sink.

Jinja2's default is `autoescape=False`, so there is no current exploit — but the absence of an explicit stamp means the invariant is not enforced in code and could silently break if the call site or a plugin passes unexpected params.

bctiemann:
## Summary

Forces `autoescape=False` for `ConfigTemplate` rendering by overriding `get_environment_params()` on the model.

### Root cause

`render_jinja2()` unpacks a caller-supplied `environment_params` dict directly into `SandboxedEnvironment(...)` with no override of `autoescape`. Jinja2's default is `False`, so there is no current exploit, but the invariant is not enforced in code. A caller (or plugin) supplying `autoescape=True` in `environment_params` would produce an environment that HTML-escapes config template output. If that output is later rendered in a web context under the assumption it is already plain text, this creates a latent XSS sink.

### Fix

`render_jinja2()` is shared by both `ConfigTemplate` and `ExportTemplate` (via `RenderTemplateMixin`). An earlier version of this PR forced `autoescape=False` unconditionally inside `render_jinja2()` itself, which also silently overrode `ExportTemplate`'s `autoescape` setting — an explicitly supported, documented option (`JINJA_ENV_PARAMS_ALLOWED['autoescape']`) that export templates may legitimately set to `True` when producing HTML output.

The fix is now scoped to `ConfigTemplate.get_environment_params()`, which calls `super().get_environment_params()` and then forces `params['autoescape'] = False`. Config templates render plain text — network configurations, scripts — not HTML, so autoescaping must always be off there. `ExportTemplate` is untouched and continues to honor whatever `autoescape` value is supplied.

### Tests

Added two regression tests in `extras/tests/test_models.py`:
- `test_configtemplate_autoescape_always_disabled` — `ConfigTemplate` ignores an explicit `environment_params={'autoescape': True}` and never escapes output.
- `test_exporttemplate_autoescape_is_configurable` — `ExportTemplate` still honors an explicit `autoescape=True`.

Also updated three pre-existing tests (`test_get_environment_params_handles_none`, `test_none_environment_params`, `test_empty_environment_params`) whose exact-dict-equality assertions needed the new `autoescape: False` key that `ConfigTemplate.get_environment_params()` now always includes.

### DIFF
--- a/netbox/extras/models/configs.py
+++ b/netbox/extras/models/configs.py
@@ -316,6 +316,16 @@ class ConfigTemplate(
         self.template_code = self.data_file.data_as_string
     sync_data.alters_data = True
 
+    def get_environment_params(self):
+        """
+        Config templates render plain text (network configs, scripts), not HTML. Force
+        autoescape off so environment_params cannot enable it and create a latent XSS sink
+        if output is ever rendered in an HTML context.
+        """
+        params = super().get_environment_params()
+        params['autoescape'] = False
+        return params
+
     def format_render_error(self, exc):
         """
         Return a formatted error string for a rendering exception. When debug is enabled, the full
--- a/netbox/extras/tests/test_models.py
+++ b/netbox/extras/tests/test_models.py
@@ -1310,6 +1310,24 @@ class RenderTemplateMixinRenderTestCase(TestCase):
         self.assertNotEqual(plain.render(ctx), trimmed.render(ctx))
         self.assertEqual(trimmed.render(ctx).strip(), 'VALUE')
 
+    def test_configtemplate_autoescape_always_disabled(self):
+        """
+        ConfigTemplate renders plain text (network configs, scripts); autoescape must stay off
+        even if environment_params explicitly requests it (#22652).
+        """
+        t = ConfigTemplate(name='autoescape', template_code='{{ value }}', environment_params={'autoescape': True})
+        self.assertEqual(t.render({'value': '<script>'}), '<script>')
+
+    def test_exporttemplate_autoescape_is_configurable(self):
+        """
+        Unlike ConfigTemplate, ExportTemplate output may legitimately be HTML, so an explicit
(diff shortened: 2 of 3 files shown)


pheus (changes requested):
Thanks for working on this!

One concern: `render_jinja2()` is shared by both configuration and export templates, so forcing `autoescape=False` here also disables escaping for inline HTML exports that explicitly enable it. Could we preserve the supplied value, perhaps with `setdefault()`, or scope this behavior to `ConfigTemplate` and add a small regression test?

bctiemann:
Thanks -- I went with a narrower scope to avoid breaking `ExportTemplate` functionality or other uses of `render_jinja2` that are supported.

pheus (approved):
Thanks! This looks good to me now!

PR #22655, opened 2026-07-09
Closes #22654: Redact install paths from debug tracebacks

### LINKED ISSUE #22654: ConfigTemplate debug mode exposes server filesystem paths in rendered error output
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.3

### Python Version

3.12

### Steps to Reproduce

1. Create a ConfigTemplate with **Debug** enabled.
2. Render the template against an object in a way that causes a Jinja2 rendering error.
3. Observe the error response body returned by the render endpoint.

### Expected Behavior

Tracebacks shown in debug mode should not reveal the absolute filesystem layout of the NetBox installation (e.g. `/home/netbox/opt/netbox/...`). Paths should be expressed relative to the repository root or stripped to the filename-and-line-number level.

Additionally, the **Debug** field should only be settable by superusers, since enabling it leaks internal server details.

### Observed Behavior

`ConfigTemplate.format_render_error()` calls `traceback.format_exception()` and returns the full traceback string verbatim when `debug=True`. This string includes `File "/abs/path/to/netbox/..."` lines exposing the absolute install-path prefix of every Python module in the traceback.

The `debug` field is also writable by any user who has permission to edit a ConfigTemplate, allowing non-privileged users to deliberately trigger detailed error output.

bctiemann:
## Summary

- `ConfigTemplate.format_render_error()` now strips the absolute install-path prefix from all `File "..."` lines in the Jinja2 traceback before returning it, so internal filesystem layout is not disclosed to users (CWE-209).

### Root cause

`ConfigTemplate.format_render_error()` returned `traceback.format_exception(exc)` verbatim when `debug=True`. The output includes `File "/abs/install/path/..."` entries for every Python frame in the traceback, revealing the absolute filesystem layout of the NetBox installation to whoever can trigger a render error (CWE-209).

### Fix

**Path redaction** — after calling `traceback.format_exception()`, apply `re.sub` with a pattern anchored to `re.escape(install_root)` (computed as `os.path.dirname(settings.BASE_DIR) + os.sep`) to replace the absolute prefix with an empty string. The result is a traceback with paths relative to the repo root, which is safe to show. When the venv lives outside the repo, its root is stripped separately as well.

### Tests

Added `test_format_render_error_debug_redacts_install_path` and `test_format_render_error_non_debug_returns_concise_message` to `ConfigTemplateDebugTestCase` in `extras/tests/test_models.py`. Existing `test_render_jinja2_*` and `ConfigTemplateTestCase` tests continue to pass.

### DIFF
--- a/netbox/extras/api/serializers_/configtemplates.py
+++ b/netbox/extras/api/serializers_/configtemplates.py
@@ -1,7 +1,6 @@
 from django.utils.translation import gettext_lazy as _
 from drf_spectacular.extensions import OpenApiSerializerExtension
 from rest_framework import serializers
-from rest_framework.exceptions import ValidationError
 
 from core.api.serializers_.data import DataFileSerializer, DataSourceSerializer
 from extras.models import ConfigTemplate
@@ -39,12 +38,6 @@ class ConfigTemplateSerializer(
             'data_file', 'auto_sync_enabled', 'data_synced', 'owner', 'tags', 'created', 'last_updated',
         ]
         brief_fields = ('id', 'url', 'display', 'name', 'description')
-
-    def validate_debug(self, value):
-        # Restrict debug mode to superusers only (CWE-209: exposes install paths via traceback)
-        if value and not self.context['request'].user.is_superuser:
-            raise ValidationError(_("Debug mode may only be enabled by superusers."))
-        return value
 
 
 class RenderConfigInputSerializer(serializers.Serializer):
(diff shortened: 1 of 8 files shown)


bctiemann:

bctiemann:
Addressed items 1 and 3. 2 and 4 are minor/below the bar for this fix.

jeremystretch (changes requested):
> The `debug` field is now restricted to superusers

This cannot be included in the scope of a bug fix, as it represents a change to intended functionality. It needs to be a separate FR. Please revert the related changes so that the PR is limited to the sanitization of file paths.

bctiemann:
Much simpler change now. Also opened https://github.com/netbox-community/netbox/issues/22664 to track the superuser gate.

jeremystretch (approved):

PR #22663, opened 2026-07-10
Closes #22522: Render colored badges for Custom Field Choices in tables

### LINKED ISSUE #22522: Render colored badges for select/multiselect custom fields in list views
### NetBox version

v4.6.3

### Feature type

### Proposed functionality

Custom fields of type `select` and `multiselect` already render as colored badges on object detail pages (when colors are defined on the choice set).
However, in list/table views, `CustomFieldColumn` renders these values as plain text, ignoring any configured colors.
Other fields like `tags` or `status` shows with a badge like style.

In [netbox/netbox/tables/columns.py:538](https://github.com/netbox-community/netbox/blob/f25a1c7474e4b0584246675a9a37eb5355bc128c/netbox/netbox/tables/columns.py#L538), `CustomFieldColumn.render()` handles `TYPE_SELECT` and `TYPE_MULTISELECT` as:
```py
if self.customfield.type == CustomFieldTypeChoices.TYPE_SELECT:
    return self.customfield.get_choice_label(value)
if self.customfield.type == CustomFieldTypeChoices.TYPE_MULTISELECT:
    return ', '.join(self.customfield.get_choice_label(v) for v in value)
```

To be consistent with the detail page, it could be implemented as follows :
```py
if self.customfield.type == CustomFieldTypeChoices.TYPE_SELECT:
    label = self.customfield.get_choice_label(value)
    color = self.customfield.get_choice_color(value)
    if color:
        return mark_safe(f'<span class="badge text-bg-{color}">{escape(label)}</span>')
    return label
if self.customfield.type == CustomFieldTypeChoices.TYPE_MULTISELECT:
    parts = []
    for v in value:
        label = self.customfield.get_choice_label(v)
        color = self.customfield.get_choice_color(v)
        if color:
            parts.append(f'<span class="badge text-bg-{color}">{escape(label)}</span>')
        else:
            parts.append(escape(label))
    return mark_safe(' '.join(parts))
```
PS: The implementation may need to be adjusted to align with the project's conventions (e.g. using the existing `{% badge %}` template tag instead of inline HTML if possible).

### Use case

We have added an `Environment` field to categorize Virtual Machines by their end-usage (Production, Staging, Dev) and it would make the table visually clearer as for the use of the machine :

Image

### Database changes

_No response_

### External dependencies

_No response_

studioussagar:
Render select and multiselect custom field values as colored badges in table views when a color is defined on the associated choice set.

This aligns table rendering with object detail pages, which already display colored choice values as badges.

##

### DIFF
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -539,27 +539,35 @@ class CustomFieldColumn(tables.Column):
         if self.customfield.type == CustomFieldTypeChoices.TYPE_URL:
             return mark_safe(f'<a href="{escape(value)}">{escape(value)}</a>')
         if self.customfield.type == CustomFieldTypeChoices.TYPE_SELECT:
+            if value is None:
+                return self.default
             label = self.customfield.get_choice_label(value)
             color = self.customfield.get_choice_color(value)
-            if color:
-                return mark_safe(
-                    f'<span class="badge text-bg-{escape(color)}">{escape(label)}</span>'
+            return mark_safe(
+                    f'<span class="badge text-bg-{escape(color or "secondary")}">{escape(label)}</span>'
                 )
-            return label
         if self.customfield.type == CustomFieldTypeChoices.TYPE_MULTISELECT:
+            has_color = False
             if not value:
-                return ''
+                return self.default
             parts = []
+            
             for v in value:
                 label = self.customfield.get_choice_label(v)
                 color = self.customfield.get_choice_color(v)
                 if color:
-                    parts.append(
-                        f'<span class="badge text-bg-{escape(color)}">{escape(label)}</span>'
-                    )
-                else:
-                    parts.append(escape(label))
-            return mark_safe(', '.join(parts))
+                    has_color = True
+                parts.append( (escape(label), color) )
+            if has_color:
+                badges = []
+                for label, color in parts:
+                    badges.append(
+                        f'<span class="badge text-bg-{escape(color or "secondary")}">{escape(label)}</span>'
+                        )
+                return mark_safe(' '.join(badges))
+            else:
+                return ', '.join(label for label, _ in parts)
+            
         if self.customfield.type == CustomFieldTypeChoices.TYPE_MULTIOBJECT:
             return mark_safe(', '.join(
                 self._linkify_item(obj) for obj in self.customfield.deserialize(value)
--- a/netbox/netbox/tables/columns.py
+++ b/netbox/netbox/tables/columns.py
@@ -53,11 +53,13 @@ __all__ = (
 # Django-tables2 overrides
 #
 
+
 @library.register
 class DateColumn(tables.Column):
     """
     Render a datetime.date in ISO 8601 format.
     """
+
     def render(self, value):
         if value:
             return value.isoformat()
@@ -83,6 +85,7 @@ class DateTimeColumn(tables.Column):
     Args:
         timespec: Granularity specification; passed through to datetime.isoformat()
     """
+
     def __init__(self, *args, timespec='seconds', **kwargs):
         self.timespec = timespec
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


pheus:

pheus (changes requested):
Thanks for opening this PR!

pheus (changes requested):
Thanks for the follow-up work! This still needs a small adjustment to match the existing custom field rendering behavior:

1. Not every choice is required to define a color. When values are rendered as badges, any choice without an explicit color should fall back to `secondary`.
2. For a multiselect field, the rendering mode should apply to the complete set of selected values. If at least one selected value has a color, all selected values should be rendered as badges, without commas between them. Only when none of the selected values has a color should they be rendered as comma-separated plain text.

Could you please update the logic accordingly?

pheus (changes requested):
Thanks for the follow-up! The overall multiselect behavior now looks right.

There are still a few details to address:

1. For a single-select field, an uncolored choice should remain plain text. I realize my previous wording could be read as applying the `secondary` fallback here too, but that fallback is intended only for uncolored values within a multiselect that is already being rendered as badges. Please retain the previous `if color: ...; return label` behavior for single-select fields.

2. Please store the unescaped label in `parts` and escape it only when inserting it into the badge HTML. The label is currently escaped when added to `parts` and then escaped again in the badge branch, which will display encoded entities for labels containing characters such as `&` or `<`. The plain-text branch should return the normal labels and allow the table template to escape them.

3. Please preserve the existing behavior for an empty multiselect by returning an empty string. `self.default` here is the column default rather than the table-level display default, and will normally be `None`.

Could you also add focused tests in `netbox/netbox/tests/test_tables.py` covering:

- colored and uncolored single-select values;
- an empty multiselect;
- a multiselect with no colored values;
- a mixed colored/uncolored multiselect, including the `secondary` fallback and absence of commas;
- a label containing HTML-sensitive characters, to guard against both missing and double escaping.

pheus (changes requested):
Thanks for the update! The rendering behavior now looks correct, and the escaping issue has been addressed.

Just two final things before this is ready:

1. Please run Ruff and the formatter over the changed files. There are still a few small style issues, such as the missing space in `parts.append((label,color))`, the redundant `else`, import ordering, and the missing final newline.

2. Could you strengthen the multiselect tests slightly?
   - Test uncolored selected values using a choice set that also contains a colored choice, to confirm the decision is based on the selected values.
   - Move the HTML-sensitive label test to the multiselect path, since that is where the earlier double-escaping issue occurred.

Once those are cleaned up, this should be good to go. Thanks!

pheus (changes requested):
Thanks for the update! The rendering logic and the main test coverage now look good.

I have two final requests:

- The latest commit reformats quite a lot of unrelated code in both files. I realize my previous request to run the formatter was too broad. Sorry about that. Could you please revert the formatting-only changes outside `CustomFieldColumn` and the new tests so the PR stays focused?
- In `test_html_sensitive_multiselect_labels`, could you also assert that the `<script>` label is escaped and that raw `<script>` is not present? The value is rendered, but the test currently only checks the `<b>` label.

After that, this should be ready to merge. Thanks for sticking with the review!

netbox/netbox/tables/columns.py
```diff
@@ -539,9 +539,36 @@ def render(self, value):
...
+                return mark_safe(
+                    f'<span class="badge text-bg-{escape(color)}">{escape(label)}</span>'
+                )
+            return label
         if self.customfield.type == CustomFieldTypeChoices.TYPE_MULTISELECT:
-            return ', '.join(self.customfield.get_choice_label(v) for v in value)
+            has_color = False
+            if not value:
+                return ""
+            parts = []
```
pheus:
```suggestion
            if not value:
                return ""

            has_color = False
            parts = []
```

pheus:

pheus (changes requested):
Thanks for the update! The implementation and tests now look good, and the unrelated formatting changes have been removed.

There’s just one small cleanup left. Could you keep it limited to the new lines and fix the import ordering, the indentation of the closing parenthesis in `badges.append()`, the remaining double-quoted strings, and the missing final newline in `test_tables.py`? Please run `ruff check` once more afterward.

After that, this looks ready from my side. Thanks for sticking with the review!

pheus (approved):
Thanks for the cleanup! The remaining review points have been addressed, and the implementation and tests now look good.

The select and multiselect rendering matches the expected behavior, including the `secondary` fallback and proper escaping of choice labels and colors. I don’t see any further concerns from my side.

Thanks for sticking with the review!

PR #22666, opened 2026-07-12
Fixes #22565: Include Circuit distance in Cable Path length calculations

### LINKED ISSUE #22565: Cable tracers calculated total length, does not take Circuit lenght in account
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.7

### Python Version

3.12

### Steps to Reproduce

1. Create Circuit of Dark Fiber type with Distance 10km
2. Create respective A and Z sides for the circuit
3. Attach cables of 200 meter length on both sides of circuit (for example single mode fiber) to network device ports
4. Perform cable trace

### Expected Behavior

Total length for cable trace should be 10.4 km. Correct calculation is important for optical budgeting.

### Observed Behavior

Cable trace only shows total of 400 meters (length of cables, ignores dark fiber circuit distance).

pheus:
### Closes: netbox-community/netbox#22565

This PR updates cable path length calculations to include the configured distance of circuits crossed between opposing terminations.

Circuit terminations and their related circuits are preloaded, and crossings are identified from adjacent path hops. For paths with parallel circuits, the longest distance is included, matching the existing handling of parallel cables. A missing cable length or circuit distance continues to mark the calculated length as non-definitive.

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -1169,10 +1169,14 @@ class CablePath(models.Model):
 
     def get_total_length(self):
         """
-        Return a tuple containing the sum of the length of each cable in the path
-        and a flag indicating whether the length is definitive.
+        Return a tuple containing the sum of the length of each cable and the distance of each circuit
+        crossed by the path, and a flag indicating whether the length is definitive.
         """
-        cable_ct = ObjectType.objects.get_for_model(Cable).pk
+        from circuits.models import CircuitTermination
+
+        object_types = ObjectType.objects.get_for_models(Cable, CircuitTermination)
+        cable_ct = object_types[Cable].pk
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
@claude /review

1. Parallel-circuit test: This has been added.
2. `is_definitive` check: This is intentional. If a referenced termination cannot be resolved, we cannot determine whether it belongs to a crossing, so the result should not be considered definitive. I'll add a comment to clarify this.
3. Iterating `_nodes` twice: I'll leave this as-is. `_nodes` is already materialized, and reusing `get_cable_ids()` keeps the logic simpler.
4. `else: break`: This relies on the existing invariant that path node sets are homogeneous and matches the previous behavior, so no change is needed.

jnovinger (approved):

PR #22668, opened 2026-07-13
Fixes #22662: Fix database overflow when saving Cables with large lengths

### LINKED ISSUE #22662: Cable length exceeds field limit causes 500 error instead of validation
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.4

### Python Version

3.12

### Steps to Reproduce

1. Edit a cable and set Length to 1234, Length unit to Kilometers.
2. Click Save.

### Expected Behavior

Either the length saves, or the form rejects it with a validation error (e.g. "Length exceeds the maximum supported value").

### Observed Behavior

500 Internal Server Error: `DataError: numeric field overflow DETAIL: A field with precision 10, scale 4 must round to an absolute value less than 10^6.`

pheus:
### Closes: netbox-community/netbox#22662

This PR increases `Cable._abs_length.max_digits` from 10 to 14, allowing the normalized field to accommodate the full range already supported by `Cable.length`, including large values expressed in kilometers or miles.

It also adds regression tests covering the reported large kilometer value and the maximum supported value in miles, ensuring both can be normalized without causing a database overflow.

### DIFF
--- a/netbox/dcim/migrations/0238_alter_cable__abs_length.py
+++ b/netbox/dcim/migrations/0238_alter_cable__abs_length.py
@@ -1,0 +1,16 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dcim', '0237_module_remove_local_context_data'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='cable',
+            name='_abs_length',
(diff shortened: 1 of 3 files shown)


pheus:

pheus:
`14` is intentional, as it covers the full range currently supported by `Cable.length`, including the maximum value in miles. Since the boundary test documents that contract, I’d prefer to keep it sized to the current requirement rather than add arbitrary headroom.

jnovinger (approved):

PR #22676, opened 2026-07-14
#22675 Validate RSS feed entry link schemes to prevent javascript: XSS

### LINKED ISSUE #22675: RSS feed widget renders unvalidated entry.link scheme, enabling javascript: DOM-XSS
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.5

### Python Version

3.14

### Steps to Reproduce

1. Host an RSS/Atom feed you control (e.g. https://attacker.example/feed.xml) containing at least one entry whose <link> uses a javascript: scheme:
  <item>
    <title>Click me</title>
    <link>javascript:alert(document.cookie)</link>
    <description>Latest news</description>
  </item>
2. Go to your dashboard and click Add Widget.
3. Choose RSS Feed as the widget type.
4. Set Feed URL to https://attacker.example/feed.xml and save the widget.
6. Observe the entry rendered

### Expected Behavior

Feed entry links should be validated against ALLOWED_URL_SCHEMES before rendering (as CustomLink already does). Any link with a disallowed scheme (e.g. javascript:, data:) should be blanked so it cannot execute script, while legitimate http/https links continue to render normally.

### Observed Behavior

The RSS feed widget interpolates the feed-controlled entry.link directly into <a href="{{ entry.link }}"> with no scheme validation.

arthanson:
### Closes: #22675 

Sanitizes the RSS fields before they go into cache.

### DIFF
--- a/netbox/extras/dashboard/widgets.py
+++ b/netbox/extras/dashboard/widgets.py
@@ -358,7 +358,9 @@ class RSSFeedWidget(DashboardWidget):
     def cache_key(self):
         url = self.config['feed_url']
         url_checksum = sha256(url.encode('utf-8')).hexdigest()
-        return f'dashboard_rss_{url_checksum}'
+        # The version segment invalidates entries cached by a pre-sanitization release: such
+        # entries live under the old key and are never read, so they can't be served unsanitized.
+        return f'dashboard_rss_2_{url_checksum}'
 
     def get_feed(self):
         if self.config.get('requires_internet') and settings.ISOLATED_DEPLOYMENT:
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
per claude review:

1. No behavior change - clarified via comment.
2. Removed per-render sanitization - now sanitize only on write and versioned the cache key, so cached feeds are clean without repeated work and pre-fix entries are never served. 
3. This isn't used or displayed.

bctiemann (approved):

PR #22679, opened 2026-07-14
#22678 - Add security note for Redis broker trust / RQ task deserialization

### LINKED ISSUE #22678: Add a security note for RQ task deserialization / Redis broker trust
### Change Type

Addition

### Area

Configuration

### Proposed Changes

Add a security note to the REDIS section of `docs/configuration/required-parameters.md` explaining that the tasks Redis database is a trusted component. NetBox's background workers deserialize and execute jobs read from this database (RQ's default pickle serializer), so any party with write access to it can achieve code execution on a worker. 

This restates the trust boundary already declared in THREAT_MODEL.md (Redis is a trusted actor) at the point where an operator actually configures Redis. Something like:

> NetBox's background workers deserialize and execute jobs read from the tasks Redis database, so any party with write access to it can run arbitrary code on a worker process. Redis must therefore be treated as trusted infrastructure, on par with the PostgreSQL database: keep it bound to localhost or a private network and require authentication. See NetBox's threat model for details.

A matching one-line note may also be added to `docs/installation/2-redis.md`.

arthanson:
### Closes: #22678

Added security note re Redis is a trusted component.

### DIFF
--- a/docs/configuration/required-parameters.md
+++ b/docs/configuration/required-parameters.md
@@ -146,6 +146,9 @@ REDIS = {
     It is highly recommended to keep the task and cache databases separate. Using the same database number on the
     same Redis instance for both may result in queued background tasks being lost during cache flushing events.
 
+!!! danger "Redis is a trusted component"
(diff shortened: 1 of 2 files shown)


pheus (approved):
Thanks!

PR #22689, opened 2026-07-15
Closes #22687: Fix queryset truthiness check in RenderTemplateMixin.render_to_response()

### LINKED ISSUE #22687: Truthiness check against queryset existence causes performance hit in RenderTemplateMixin using export templates
### NetBox Version

v4.6.4

### Python Version

3.12

### Area(s) of Concern

- [ ] User Interface
- [ ] REST API
- [ ] GraphQL API
- [x] Python ORM
- [ ] Other

### Observations

Performed an audit of the NetBox codebase looking for occurrences of the pattern where a truthiness comparison of a queryset (e.g. `if self.queryset:`) is used instead of a stricter comparison against None (`if self.queryset is not None:`).

Found one such occurrence:
https://github.com/netbox-community/netbox/blob/b61c232305a34374f2aa5cc27e41fe6ee157aa96/netbox/extras/models/mixins.py#L254-L259

This construct forces the evaluation of the entire queryset, which (because this mixin is used in every list view in NetBox) might be applied to any exposed model in the system, with unbounded size.

I traced the call chain and confirmed real-world impact:

- `netbox/api/viewsets/mixins.py:148` — REST API `?export=<template>` action
- `netbox/views/generic/bulk_views.py:166` — the UI "Export" button when using a custom Export Template

Both pass the entire current list-view queryset (e.g., every Device or Site matching the active filter — potentially the whole table). `ExportTemplate.as_attachment` defaults to `True`, and `file_name` is an optional field most users leave blank — so this branch is the common path. `elif queryset:` triggers `QuerySet.__bool__()` → `_fetch_all()`, fully materializing and caching the entire queryset in memory just to read `queryset.model` for a default filename — before the template even renders. 

### Proposed Changes

Change the `elif queryset` to `elif queryset is not None`.

bctiemann:
## Summary

`RenderTemplateMixin.render_to_response()` (`netbox/extras/models/mixins.py`) checked `elif queryset:` to decide whether a queryset was supplied when deriving a default export filename. This is both a performance bug and a functional correctness bug.

## Root cause

`elif queryset:` relies on Python truthiness, which for a Django `QuerySet` calls `__bool__()` → `_fetch_all()` — forcing a full, unbounded evaluation of every matching row into memory, purely to check whether a queryset was provided at all.

`render_to_response()` is reachable from any model's list view via a configured `ExportTemplate`:
- `netbox/api/viewsets/mixins.py` — the REST API's `?export=<template>` action
- `netbox/views/generic/bulk_views.py` — the UI's "Export" button

Both pass the entire current (filtered) list-view queryset. `ExportTemplate.as_attachment` defaults to `True`, and `file_name` is an optional field most users leave blank, so this branch is the common path — meaning a large, unfiltered export could materialize an entire table's worth of rows in memory just to read `queryset.model`.

This is also a **functional bug**: an empty (but valid, non-`None`) queryset is falsy, so the branch was incorrectly skipped in favor of the `"output"` filename fallback even when a real queryset was supplied — producing the wrong default filename for any export that happens to match zero rows.

## Fix

Check `queryset is not None` instead of relying on truthiness.

## Tests

Added two regression tests to `RenderTemplateMixinResponseTestCase` in `extras/tests/test_models.py`:
- `test_response_attachment_filename_from_empty_queryset` — an empty queryset must still yield a model-derived filename, not the `"output"` fallback.
- `test_response_attachment_does_not_force_queryset_evaluation` — using a template that never references `queryset`, asserts via `CaptureQueriesContext` that no query against the underlying table is issued at all.

I verified both tests fail against the pre-fix code (one with the wrong filename, one with an unbounded full-table `SELECT`) and pass with the fix applied. Full `extras.tests.test_models` suite (109 tests) passes on a fresh test database; `ruff check` passes.

### DIFF
--- a/netbox/extras/models/mixins.py
+++ b/netbox/extras/models/mixins.py
@@ -244,7 +244,7 @@ class RenderTemplateMixin(models.Model):
             extension = f'.{self.file_extension}' if self.file_extension else ''
             if self.file_name:
                 filename = self.file_name
-            elif queryset:
+            elif queryset is not None:
                 filename = filename_from_model(queryset.model)
             elif context:
                 filename = filename_from_object(context)
--- a/netbox/extras/tests/test_models.py
+++ b/netbox/extras/tests/test_models.py
@@ -1449,13 +1449,7 @@ class RenderTemplateMixinResponseTestCase(TestCase):
         self.assertEqual(response['Content-Disposition'], 'attachment; filename="netbox_sites.txt"')
 
     def test_response_attachment_filename_from_empty_queryset(self):
-        """
-        Regression test (#22687): a non-None but empty queryset must still yield a
-        model-derived filename. A prior bug checked `elif queryset:` (truthiness)
-        instead of `elif queryset is not None:`; QuerySet.__bool__() returns False for
-        an empty queryset, so this branch was incorrectly skipped in favor of the
-        "output" fallback even though a valid queryset was provided.
-        """
+        """An empty (but non-None) queryset must still yield a model-derived filename."""
         t = ExportTemplate(
             name='t',
             template_code='{% for obj in queryset %}{{ obj.name }}{% endfor %}',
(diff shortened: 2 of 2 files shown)


bctiemann:

bctiemann:
Item 1 doesn't seem relevant; we're keeping this fix narrowly on the queryset comparison, not other truthiness checks. Fixed items 2 and 3.

jeremystretch (approved):

PR #22691, opened 2026-07-15
Closes #22569: Add explicit path for Script POST detail route

### LINKED ISSUE #22569: POST `/api/extras/scripts/{id}/` missing from OpenAPI schema
### NetBox Edition

NetBox Community

### NetBox Version

v4.5.9

### Python Version

3.12

### Steps to Reproduce

The POST `/api/extras/scripts/{id}/` endpoint (used to run a script) is absent from the generated OpenAPI schema. The endpoint works correctly at runtime, but it is invisible to any tooling that relies on the schema — including auto-generated clients, documentation browsers, and API tool registries.  So those tools are unable to run a custom script.

1. Start NetBox and navigate to /api/schema/
2. Search for `/api/extras/scripts/{id}/` in the output

### Expected Behavior

`POST /api/extras/scripts/{id}/` should appear in the schema, matching the actual runtime behavior.

### Observed Behavior

Only GET, PUT, PATCH, and DELETE are listed — POST is absent

g-builder-0:
Closes: netbox-community/netbox#22569

## Fix

DRF's router never maps `post` into a detail route's `actions` dict by<br>default, so `ScriptViewSet.post()` — reachable at runtime via Django's<br>generic method dispatch, not via the router — was invisible to<br>drf-spectacular's schema generation. Added an explicit `path()` in<br>`extras/api/urls.py` declaring all five methods for this route, including<br>`post`, placed before the router's `include()` so it takes precedence.

## Also fixed

`ScriptViewSet.post()`'s docstring claimed it returns "the pending Job,"<br>but it returns a Script object with the Job nested under `result` (per<br>`ScriptDetailSerializer`). Corrected to match.

## Testing

* Added `test_script_post_operation_present`, confirming POST appears in<br>the schema for this path.
* `extras.tests.test_api.ScriptTestCase` and<br>`core.tests.test_openapi_schema.OpenAPISchemaTestCase` both pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ __pycache__/
 
 # Python virtual environment created by the installation/upgrade workflow
 /venv/
+/.venv/
 
 # Frontend dependencies and Yarn logs generated during asset development/builds
 /netbox/project-static/node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ __pycache__/
 
 # Python virtual environment created by the installation/upgrade workflow
 /venv/
-/.venv/
 
 # Frontend dependencies and Yarn logs generated during asset development/builds
 /netbox/project-static/node_modules/
--- a/netbox/extras/api/urls.py
+++ b/netbox/extras/api/urls.py
@@ -32,5 +32,18 @@ router.register('scripts', views.ScriptViewSet, basename='script')
 app_name = 'extras-api'
 urlpatterns = [
     path('dashboard/', views.DashboardView.as_view(), name='dashboard'),
+    path(
+        'scripts/<str:pk>/',
+        views.ScriptViewSet.as_view(
+            {
+                'get': 'retrieve',
+                'put': 'update',
+                'patch': 'partial_update',
+                'delete': 'destroy',
+                'post': 'post',
+            }
+        ),
+        name='script-detail',
+    ),
     path('', include(router.urls)),
 ]
--- a/netbox/extras/api/urls.py
+++ b/netbox/extras/api/urls.py
@@ -33,6 +33,14 @@ app_name = 'extras-api'
 urlpatterns = [
     path('dashboard/', views.DashboardView.as_view(), name='dashboard'),
     path(
+        'scripts/upload/',
+        views.ScriptModuleViewSet.as_view(
+            {
+                'post': 'create',
+            }
+        ),
+    ),
+    path(
         'scripts/<str:pk>/',
         views.ScriptViewSet.as_view(
             {
--- a/netbox/extras/api/urls.py
+++ b/netbox/extras/api/urls.py
@@ -1,4 +1,5 @@
 from django.urls import include, path
+from rest_framework.routers import Route
 
 from netbox.api.routers import NetBoxRouter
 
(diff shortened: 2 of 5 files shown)


jeremystretch:
@g-builder-0 could you address all the test failures please?

g-builder-0:
> @g-builder-0 could you address all the test failures please?

Fixed. The path() I added for scripts/{id}/ was broad enough to also match scripts/upload/ (since <str:pk> matches any single path segment, including "upload"), shadowing ScriptModuleViewSet's routes. Added an explicit path('scripts/upload/', ...) before the broader pattern so the specific route takes precedence. All three relevant test suites pass, and CI is green.

pheus:

netbox/extras/api/urls.py
```diff
@@ -32,5 +32,26 @@
...
+            {
+                'get': 'retrieve',
+                'put': 'update',
+                'patch': 'partial_update',
+                'delete': 'destroy',
+                'post': 'post',
+            }
+        ),
+        name='script-detail',
+    ),
```
pheus: **Blocking:** Could we keep this endpoint on the router-generated detail route instead of shadowing it with hand-written `path()` entries?

The router registrations are still present, so these paths duplicate them and rely on URL ordering to take precedence. They also bypass the router-provided `basename`, `detail`, display suffix, route naming, and format-suffix variants.

There is a concrete regression here: `POST /api/extras/scripts/upload.json/` previously resolved to the router-generated `ScriptModuleViewSet.create` route. The exact `scripts/upload/` path does not match the suffixed URL, so the broader `scripts/<str:pk>/` path now captures it as `pk='upload.json'` and dispatches it to `ScriptViewSet.post()`.

Could we instead extend the router-generated Script detail mapping, for example, using a local custom router/`Route` that adds `'post': 'post'`? That would preserve the existing route metadata and suffix behavior without introducing precedence-sensitive duplicate paths.

netbox/extras/api/views.py
```diff
@@ -349,18 +363,16 @@ def retrieve(self, request, pk):
 
     def post(self, request, pk):
```
pheus, line 355: **Blocking:** The operation is now present in the schema, but its generated request and response schemas still do not match the endpoint’s runtime contract.

This method accepts `ScriptInputSerializer` and returns `ScriptDetailSerializer`, whereas automatic introspection derives the schema from `ScriptViewSet.serializer_class`.

The default operation ID also collides with the existing list `POST /api/extras/scripts/`: both POST operations are assigned a `create`-based ID after the `{id}` path variable is removed, so one receives a numeric suffix.

Could we explicitly annotate this operation, for example:

```python
@extend_schema(
    operation_id='extras_scripts_run',
    request=serializers.ScriptInputSerializer,
    responses={200: serializers.ScriptDetailSerializer},
)
def post(self, request, pk):
    ...
```

The exact operation ID can follow the project’s preferred naming, but it should be stable and distinct from the list-create operation.

.gitignore
```diff
@@ -5,6 +5,7 @@ __pycache__/
 
 # Python virtual environment created by the installation/upgrade workflow
 /venv/
+/.venv/
```
pheus: **Blocking cleanup:** Could you revert the unrelated `/.venv/` ignore entry and the formatting-only changes outside the lines needed for this fix? Keeping the diff focused will make the routing and schema behavior easier to review and avoid unnecessary merge conflicts.

pheus (changes requested):
Thanks for working on the missing OpenAPI operation. I think this needs another revision before it can be merged.

I found four blocking concerns:

1. The manually declared paths duplicate and shadow the router-generated routes. This also introduces a regression for `POST /api/extras/scripts/upload.json/`.
2. The new POST operation is documented with the wrong request and response serializers, and its default operation ID collides with that of the existing list POST operation.
3. The PR does not include tests for the new behavior. Please add schema and routing regression tests, including coverage for the request and response schemas, the operation ID, and the `scripts/upload.json/` route.
4. The PR includes unrelated code and style-only changes that should be removed to keep the change focused.

I think the cleanest approach would be to keep the detail URL router-generated, add the `POST -> post` mapping through a custom `Route` on a local router, explicitly annotate the custom operation, and add the necessary schema and routing regression tests.

.gitignore
g-builder-0: @pheus 

Thanks for the detailed review — all four points addressed:

1. Replaced the hand-written path() entries with a ScriptRouter subclass that extends the router's own detail route mapping to include post, so it's no longer a separate, shadowing pattern. Confirmed scripts/upload/ and scripts/upload.json/ both correctly dispatch to ScriptModuleViewSet again.
2. Added @extend_schema(operation_id='extras_scripts_run', request=serializers.ScriptInputSerializer, responses={200: serializers.ScriptDetailSerializer}) on post() — correct request/response schemas, no more operationId collision.
3. Added a schema-accuracy test asserting the operationId and both $ref values, plus a routing regression test covering scripts/upload/ and scripts/upload.json/.
4. Reverted the unrelated .gitignore entry and all stray formatting in views.py — the diff is now limited to the @extend_schema decorator and the docstring fix.

pheus:
Thanks for all the work you’ve put into this, and for patiently addressing the review feedback.

After looking at the different approaches, we don’t currently see a solution that improves the schema without adding routing complexity or requiring a broader API change. Rather than ask you to rework this again without a clear direction from us, I’m going to close the PR.

This is not a reflection on the quality of your contribution. We can revisit the issue once we have agreed on the API design we want to maintain going forward. Thanks again!

g-builder-0:
@pheus 
I understand. Thanks for the explanation. I'd be happy to revisit this issue once the design decisions have been worked out.

Issue #20285, opened 2025-09-06
Add new protocol type on Application service template


m0rp30:
### NetBox version

v.4.4.0-3.4.0

### Feature type

Change to existing functionality

### Proposed functionality

Considering that many services use both TCP and UDP protocols, it would be useful to have an option such as “TCP/UDP.” Even more interesting would be the ability to add the desired protocol type so that it can be customized as desired without having to create new issues.

### Use case

In a service like DNS the protocols as both TCP and UDP for the same port 53

jnovinger:
Related (if not duplicate) issues:
- #17395
- #11888
- #16841
- #17196

jeremystretch:
Rather than adding a new combined "TCP/UDP" option, it would probably be more robust to convert the field to a multi-choice field, allowing the user to select TCP and/or UDP (and potentially other protocols).

jeremystretch:
Talked about this more with the other maintainers. To provide even more flexibility, we could merge the `protocol` and `ports` fields into a single ArrayField denoting individual items as protocol & port combinations. For example: `tcp/80, tcp/443, udp/999`.

This will require a bit more work on the frontend, but will allow users to clearly define the protocols and ports in use. It will also enable more efficient filtering for services.

m0rp30:
It's an excellent idea in that the concept of protocol and port are linked.

ghost:
does this mean we'll also get the ability to assign multiple port/protocol combinations to a single service?

PR #22692, opened 2026-07-15
#20285 - Support multiple protocols per application service via port mappings

arthanson:
### Closes: #20285

This PR replaces the separate `protocol`/`ports` fields with a single **`port_mappings`** array of `protocol/port` strings (e.g. `['tcp/80', 'tcp/443', 'udp/53']`), letting a service expose any combination of protocols and ports.

Monosnap Add a new application service template | NetBox 2026-07-15 13-20-46

## Changes

- **Model**: `Service`/`ServiceTemplate` drop `protocol`, `ports`, and `_ports_lowest` in favor of a `port_mappings` `ArrayField`. At least one mapping is required; entries are validated for format, known protocol, port range, and uniqueness.
- **UI**: a new dynamic port-mapping widget (one row per protocol, comma/range port entry; e.g. `80,443,8000-8010`). Protocols already chosen are disabled in other rows.
- **REST API & GraphQL**: `port_mappings` is exposed as a list of strings; `protocol` and `port` filters are retained on both.
- **Bulk import (CSV)**: the `port_mappings` column uses `protocol:ports` pairs separated by semicolons, e.g. `tcp:80,443;udp:53`.
- **Data migration**: existing `protocol`/`ports` values are converted to `port_mappings`.
- **Extend GraphQL test:** Extended the shared GraphQL test query builder (utilities/testing/api.py) to auto-select subfields for nested object-list fields (e.g. port_mappings { protocol ports }); without it the auto-generated get/list queries emitted port_mappings with no selection set, which is invalid GraphQL for an object type.

## REST API backward compatibility

The legacy single-protocol `protocol`/`ports` fields are retained on the `Service` and `ServiceTemplate` REST serializers alongside the new `port_mappings` field. **They are deprecated and will be removed in a future release.**

**Reads** — `port_mappings` is always authoritative. The legacy fields are derived from it:

| Service state | `protocol` | `ports` |
|---|---|---|
| Single protocol (e.g. `tcp/80`, `tcp/443`) | `"tcp"` | `[80, 443]` |
| No mappings | `null` | `[]` |
| Multiple protocols | `null` | `null` |

`ports: null` specifically signals "multiple protocols — not representable in the legacy format; read `port_mappings`." An empty service returns `ports: []` (as the old API always did), keeping the two cases distinguishable.

Below `service-template-1` has the multiple protocols so only returns the new format.  `service-template-2` has only a single protocol so returns both new and old format. 
```
 {
      "count": 2,
      "next": null,
      "previous": null,
      "results": [
          {
              "id": 1,
              ...
              "name": "service-template-1",
              "port_mappings": [
                  "tcp/20",
                  "udp/30",
              ],
              "protocol": null,
              "ports": null,
              ...
          },
          {
              "id": 2,
              ...
              "name": "service-template-2",
              "port_mappings": [
                  "tcp/20"
              ],
              "protocol": "tcp",
              "ports": [80],
              ...
          }
      ]
  }
```

**Writes** — either format is accepted. If `port_mappings` is supplied it takes precedence; otherwise `protocol` + `ports` is translated into `port_mappings` and validated identically (protocol, port range, duplicates). Values submitted via the legacy fields are subject to the same "at least one mapping required" rule.

GraphQL exposes `port_mappings` only (no legacy fields).

## Behavior changes to note (release notes)

- **Bulk edit**: direct editing of `protocol`/`ports` is replaced by **add/remove port mappings** fields (like tags) - bulk-add a mapping to, or bulk-remove one from, many services at once, rather than overwriting.
- **Default ordering** of the service list changed from `(protocol, port)` to `name`; the port column is no longer sortable.
- Any pre-existing services with an **empty `ports`** array migrate to empty `port_mappings` and will require a mapping to be added before they can next be saved.


arthanson:

arthanson:
Per claude feedback:

1. Not an issue
2.  fixed
3. valid observation - not a bug
4. as noted this is okay

netbox/ipam/forms/fields.py
jeremystretch, line 1: Let's split this into generic `fields.py` and `widgets.py` modules, so they can be cleanly extended in the future if needed.

netbox/ipam/templates/ipam/widgets/port_mappings.html
```diff
@@ -0,0 +1,59 @@
...
+    </tbody>
+  </table>
+  <button type="button" class="btn btn-sm btn-outline-primary" data-port-mapping-add>
+    <i class="mdi mdi-plus-thick"></i> {% trans "Add mapping" %}
+  </button>
+
+  <template data-port-mapping-template>
+    <tr data-port-mapping-row>
+      <td>
+        <select class="form-select form-select-sm no-ts port-mapping-protocol">
```
jeremystretch: The dropdown is missing our standard styling:

Image

netbox/ipam/api/serializers_/services.py
```diff
@@ -45,7 +63,7 @@ class Meta:
         model = Service
         fields = [
             'id', 'url', 'display_url', 'display', 'parent_object_type', 'parent_object_id', 'parent', 'name',
-            'protocol', 'ports', 'ipaddresses', 'description', 'owner', 'comments', 'tags', 'custom_fields',
+            'port_mappings', 'ipaddresses', 'description', 'owner', 'comments', 'tags', 'custom_fields',
```
jeremystretch: Calling this out as a breaking API change. Per our discussion earlier, please see if you can devise a way to support backward compatibility for single-protocol services.

netbox/ipam/models/services.py
```diff
@@ -99,14 +139,16 @@ class Service(ContactsMixin, ServiceBase, PrimaryModel):
...
     class Meta:
         indexes = (
-            models.Index(fields=('protocol', '_ports_lowest', 'id')),  # Default ordering
+            models.Index(fields=('name', 'id')),  # Default ordering
             models.Index(fields=('parent_object_type', 'parent_object_id')),
+            # Supports exact protocol/port containment lookups (port_mappings @> ['tcp/80'])
+            GinIndex(fields=('port_mappings',)),
         )
-        ordering = ('protocol', '_ports_lowest', 'id')
+        ordering = ('name', 'id')
```
jeremystretch, line 183: While I understand the motivation for this change and generally support it, we should be careful to capture the change to default ordering in the release notes, particularly given there was a recent relevant bug fix (see #22273).

jeremystretch (changes requested):
Had a couple quick observations. Will hold off on further review until we've determined whether avoiding a breaking API changes is feasible. Thanks!

netbox/ipam/models/services.py
arthanson: Not sure what to do on this one - the PR description has the "release notes" section - the version-4.7.md file was deliberately removed so didn't want to re-create it.

jeremystretch: This was just for confirmation and to serve as a reminder.

netbox/ipam/migrations/0095_multi_protocol_services.py
jeremystretch, line 1: Please choose a more concise, descriptive name for the migration.

docs/models/ipam/service.md
```diff
@@ -23,13 +23,22 @@ The parent object to which the application service is assigned. This must be one
...
+```json
+[
+  {"protocol": "tcp", "ports": [80, 443]},
+  {"protocol": "udp", "ports": [53]}
+]
+```
+
+!!! note "Changed in NetBox v4.7"
+
+    The single-protocol `protocol` and `ports` fields have been replaced by the unified `port_mappings` field (which supports multiple protocols per service, in the grouped form shown above). For backward compatibility, the REST and GraphQL APIs still expose the legacy `protocol` and `ports` fields, and the REST API still accepts them on write as an alternative to `port_mappings`. They are populated for single-protocol services; a service with multiple protocols cannot be represented in the legacy format and returns `null` for both, while a service with no mappings returns `protocol: null` and `ports: []`. In other words, `ports: null` specifically signals "multiple protocols — read `port_mappings` instead." **These legacy fields are deprecated and will be removed in a future release; use `port_mappings` instead.**
```
jeremystretch: Let's go ahead and lock in removal of the backward compatibility for v5.0.
```suggestion
    The single-protocol `protocol` and `ports` fields have been replaced by the unified `port_mappings` field (which supports multiple protocols per service, in the grouped form shown above). For backward compatibility, the REST and GraphQL APIs still expose the legacy `protocol` and `ports` fields, and the REST API still accepts them on write as an alternative to `port_mappings`. They are populated for single-protocol services; a service with multiple protocols cannot be represented in the legacy format and returns `null` for both, while a service with no mappings returns `protocol: null` and `ports: []`. In other words, `ports: null` specifically signals "multiple protocols — read `port_mappings` instead." **These legacy fields are deprecated and will be removed in NetBox v5.0; use `port_mappings` instead.**
```

docs/models/ipam/servicetemplate.md
```diff
@@ -12,10 +12,23 @@ Application service templates can be used to instantiate [application services](
 
 A service or protocol name.
 
-### Protocol
+### Port Mappings
```
jeremystretch, line 15: Can we just link to the same section in `service.md` to avoid the redundant text?

netbox/ipam/templates/ipam/widgets/port_mappings.html
```diff
@@ -0,0 +1,59 @@
...
+    <thead>
+      <tr>
+        <th scope="col" style="width: 40%;">{% trans "Protocol" %}</th>
+        <th scope="col">{% trans "Ports" %}</th>
+        <th scope="col" style="width: 1%;"></th>
+      </tr>
+    </thead>
+    <tbody data-port-mapping-rows>
+      {% for row in widget.rows %}
+        <tr data-port-mapping-row>
```
jeremystretch, line 15: The height of the protocol dropdown and port input should match.

We should also avoid rendering the horizontal line beneath the two fields unless another row has been added, but that's more of a nit.

Image

netbox/ipam/models/services.py
```diff
@@ -17,42 +18,75 @@
...
-        update_fields = kwargs.get('update_fields')
-        if update_fields is not None and '_ports_lowest' not in update_fields:
-            kwargs['update_fields'] = list(update_fields) + ['_ports_lowest']
-        super().save(*args, **kwargs)
-
     def __str__(self):
-        return f'{self.name} ({self.get_protocol_display()}/{self.port_list})'
+        if self.port_mappings:
+            return f'{self.name} ({self.port_list})'
+        return self.name
```
jeremystretch: IMO it probably makes sense to just return the name. Otherwise, the object representation could become too verbose.
```suggestion
        return self.name
```

netbox/ipam/models/services.py
```diff
@@ -17,42 +18,75 @@
...
+        if add:
+            mappings += [mapping for mapping in add if mapping not in mappings]
+        if remove:
+            mappings = [mapping for mapping in mappings if mapping not in remove]
+        self.port_mappings = mappings
 
     @property
     def port_list(self):
-        return array_to_string(self.ports)
+        # Group ports by protocol for a compact display, e.g. "TCP/80,443, UDP/53". Ports are sorted
```
jeremystretch: IMO this can be more difficult to interpret than just listing each protocol/port pair individually when multiple protocols are present. Compare this

Image

to this:

Image

This also avoids needing to reformat the raw data on read.

netbox/ipam/models/services.py
```diff
@@ -17,42 +18,75 @@
...
+        # numerically within each protocol so the display is stable regardless of stored order.
+        return ', '.join(
+            # Guard the numeric sort so a malformed entry that bypassed validation (e.g. a raw SQL
+            # insert) degrades gracefully instead of raising ValueError when the service is rendered.
+            f'{protocol.upper()}/{",".join(sorted(ports, key=lambda p: int(p) if p.isdigit() else 0))}'
```
jeremystretch: Be careful with `upper()`. It works currently, but if we ever introduce a protocol whose natural representation is not fully capitalized, it will break.

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +25,152 @@
...
+    """A single protocol and its associated ports, e.g. ``{"protocol": "tcp", "ports": [80, 443]}``."""
+    protocol = serializers.ChoiceField(choices=ServiceProtocolChoices)
+    ports = serializers.ListField(
+        child=serializers.IntegerField(min_value=SERVICE_PORT_MIN, max_value=SERVICE_PORT_MAX),
+        allow_empty=False,
+    )
+
+
+@extend_schema_field(PortMappingSerializer(many=True))
+class PortMappingsField(serializers.Field):
```
jeremystretch: This feels like it might be unnecessarily complex. We're storing protocol/port combos as atomic items in the database, e.g.

```json
["tcp/80", "tcp/443", "udp/53"]
```

but representing them via the API as a hierarchy listing ports by protocol:

```json
"port_mappings": [
    {
        "protocol": "tcp",
        "ports": [
            80,
            443
        ]
    },
    {
        "protocol": "udp",
        "ports": [
            53
        ]
    }
]
```

Considering the perspective of a hypothetical REST API client, the former feels more natural and portable IMO. We still need to validate each item individually of course, but the flat list seems easier to work with in general, and it allows us to simplify the serialization logic significantly.

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +25,152 @@
...
+            return validate_port_mappings(mappings)
+        except DjangoValidationError as exc:
+            raise serializers.ValidationError(exc.messages)
+
+
+# Legacy single-protocol fields, retained on the serializers for backward compatibility. Declared via
+# factories (rather than on the shared mixin) because DRF's serializer metaclass only collects declared
+# fields from the class itself and other serializers — not from a plain mixin. default=None keeps them
+# from being sourced off the (now nonexistent) model attributes; the real values are filled in by
+# PortMappingsSerializerMixin.to_representation().
```
jeremystretch: Can we avoid this by turning `PortMappingsSerializerMixin` into a subclass of `serializers.Serializer`?

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +25,152 @@
...
+                    "Specify either 'port_mappings' or the deprecated 'protocol'/'ports' fields, not both."
+                ))
+            # The legacy fields only form a mapping as a pair. Supplying just one is ambiguous (we can't
+            # infer the other), so reject it explicitly rather than silently ignoring the input — the old
+            # behavior of updating a single field is no longer possible now that they're derived.
+            if not (legacy_protocol and legacy_ports):
+                raise serializers.ValidationError(_(
+                    "Both 'protocol' and 'ports' are required when writing via the deprecated legacy "
+                    "format; use port_mappings instead."
+                ))
```
jeremystretch, line 120: Are both strictly required? This breaks e.g. a `PATCH` that merely adds or removes an individual port within a list.

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +25,152 @@
...
+                    [f'{legacy_protocol}/{port}' for port in legacy_ports]
+                )
+            except DjangoValidationError as exc:
+                raise serializers.ValidationError({'ports': exc.messages})
+
+        # Require at least one mapping whenever they're being written (a create, or an update that sets
+        # them). A partial update that touches neither format leaves the existing mappings intact.
+        writing_mappings = 'port_mappings' in data or not self.partial
+        if writing_mappings and not data.get('port_mappings'):
+            raise serializers.ValidationError({'port_mappings': _("At least one port mapping is required.")})
```
jeremystretch: This feels like it should be handled by model validation.

netbox/ipam/forms/bulk_import.py
```diff
@@ -586,19 +589,63 @@ class Meta:
...
+    port_mappings = forms.CharField(
+        label=_('Port mappings'),
+        required=True,
+        help_text=_(
+            'Protocol/port pairs in the form "tcp:80,443;udp:53" (pairs separated by semicolons; ports '
+            'may use commas and hyphenated ranges).'
+        )
     )
 
+    def clean_port_mappings(self):
```
jeremystretch, line 604: Similar to my earlier comment, IMO we should keep this simple and just accept a flat comma-separated list of protocol & port combos, e.g. `tcp/80,tcp/443,udp/53`.

netbox/ipam/models/services.py
```diff
@@ -17,42 +18,75 @@
...
-    )
-    ports = ArrayField(
-        base_field=models.PositiveIntegerField(
-            validators=[
-                MinValueValidator(SERVICE_PORT_MIN),
-                MaxValueValidator(SERVICE_PORT_MAX)
-            ]
-        ),
-        verbose_name=_('port numbers')
-    )
```
jeremystretch, line 33: It might make sense to retain these as properties (similar to the REST API serializer, but read-only) until backward compatibility is dropped in v5.0.

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +25,152 @@
 )
 
 
-class ServiceTemplateSerializer(PrimaryModelSerializer):
-    protocol = ChoiceField(choices=ServiceProtocolChoices, required=False)
+class PortMappingSerializer(serializers.Serializer):
```
jeremystretch: Let's add a comment flagging this for removal in v5.0.

netbox/ipam/graphql/types.py
```diff
@@ -270,12 +311,24 @@ def parent(self) -> Annotated[
...
+    exclude=('port_mappings',),
     filters=ServiceTemplateFilter,
     pagination=True
 )
 class ServiceTemplateType(PrimaryObjectType):
-    ports: list[int]
+
+    @strawberry_django.field
+    def port_mappings(self) -> list[ServicePortMappingType]:
+        return _grouped_port_mappings(self)
```
jeremystretch: Similar to my comment re: the REST API, it might be preferable to stick with a flat list of protocol & port pairings.

jeremystretch (changes requested):
Great work on supporting backward compatibility; this will help smooth the transition for users significantly. Beyond that, I think we can simplify the representation of protocols and ports within the REST and GraphQL APIs substantially. IMO this would improve the client experience and allow us to shed some fairly complex validation logic.

netbox/ipam/validators.py
```diff
@@ -3,6 +3,129 @@
...
+    the model form field (``PortMappingField``), the CSV import form, and the REST API serializers so
+    all paths enforce identical rules.
+    """
+    # Imported lazily to avoid a circular import during settings load (this module is imported by
+    # ipam.models, and ipam.constants pulls in ipam.choices, which reads settings.FIELD_CHOICES).
+    from ipam.choices import ServiceProtocolChoices
+    from ipam.constants import SERVICE_PORT_MAX, SERVICE_PORT_MIN
+
+    valid_protocols = ServiceProtocolChoices.values()
+    seen = set()
```
jeremystretch, line 28: Instead of tracking a "seen" set in addition to the list of validated port mappings, can we just declare `normalized_mappings` as a set, and sort it into a list immediately before returning? This would avoid the need for the separate tracking object and ensure a truly normalized return value.

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +19,123 @@
...
+        help_text=_("Deprecated; use port_mappings. Reported only for single-protocol services."),
+    )
+
+    def validate(self, data):
+        # Consume the legacy fields and translate them into port_mappings *before* calling super(),
+        # which instantiates the model (via full_clean()) and would choke on these now-nonexistent kwargs.
+        legacy_protocol = data.pop('protocol', None)
+        legacy_ports = data.pop('ports', None)
+        legacy_supplied = legacy_protocol is not None or legacy_ports is not None
+        if legacy_supplied:
```
jeremystretch:
```suggestion
        if any([legacy_protocol, legacy_ports]):
```

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +19,123 @@
...
+            # The two formats are mutually exclusive. Rather than silently dropping one when they conflict,
+            # reject the request so the caller picks a single representation.
```
jeremystretch: For clarity
```suggestion
            # `port_mappings` and `protocol`/`ports` are mutually exclusive. Rather than silently dropping
            # one when they conflict, reject the request so the caller picks a single representation.
```

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +19,123 @@
...
+            # The legacy API let either field be updated on its own (e.g. a PATCH that adjusts only the
+            # port list). Preserve that by backfilling the omitted field from the instance's current
+            # single-protocol representation.
+            if not (legacy_protocol and legacy_ports):
+                existing_protocol, existing_ports = (
+                    legacy_protocol_and_ports(self.instance.port_mappings) if self.instance else (None, None)
+                )
+                legacy_protocol = legacy_protocol or existing_protocol
+                if legacy_ports is None:
+                    legacy_ports = existing_ports
```
jeremystretch: Can we just reference `instance.legacy_protocol` and `instance.legacy_ports` here?

netbox/ipam/api/serializers_/services.py
```diff
@@ -15,20 +19,123 @@
...
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+
+        # Populate the legacy single-protocol representation for backward compatibility. Skipped in
+        # brief mode, where these fields are not exposed. The empty and multi-protocol cases are kept
+        # distinct: an empty service reports ports=[] (as the old API always did), whereas a service
+        # with multiple protocols reports ports=null to signal "not representable in the legacy format;
+        # use port_mappings".
+        if 'protocol' in self.fields and 'ports' in self.fields:
+            data['protocol'], data['ports'] = legacy_protocol_and_ports(instance.port_mappings)
```
jeremystretch:
```suggestion
            data['protocol'] = instance.legacy_protocol
            data['ports'] = instance.legacy_ports
```

netbox/ipam/forms/bulk_import.py
```diff
@@ -586,19 +588,54 @@ class Meta:
...
-    protocol = CSVChoiceField(
-        label=_('Protocol'),
-        choices=ServiceProtocolChoices,
-        help_text=_('IP protocol')
+class ServicePortMappingsImportMixin(forms.Form):
+    """
+    Adds a ``port_mappings`` CSV column parsed from a compact string (e.g. "tcp:80,443;udp:53") into the
+    model's flat ``['tcp/80', 'tcp/443', 'udp/53']`` list.
+    """
+    port_mappings = forms.CharField(
```
jeremystretch: `port_mappings` can now be a SimpleArrayField. The list of port mappings should be a comma-separated string encased in double quotes (e.g. `"tcp/80,udp/53"`).

We might also want to force the input string to lowercase for better flexibility in specifying protocols.

netbox/ipam/filtersets.py
```diff
@@ -1206,16 +1206,107 @@ def search(self, queryset, name, value):
...
+                qs_filter |= Q(port_mappings__contains=[f'{protocol}/{port}'])
+        return False, qs_filter
+    # Protocol-only: ports are unbounded, so containment can't enumerate them; match the protocol prefix
+    # against the annotated string form.
+    for protocol in protocols:
+        qs_filter |= Q(_port_mappings_str__contains=f',{protocol}/')
+    return True, qs_filter
+
+
+def port_mapping_filter_qs(queryset, protocols, ports):
```
jeremystretch: It seems like we could eliminate a lot of this filtering cruft by storing a denormalized array of protocols (and maybe ports?) on the Service model directly, e.g.

```python
    _protocols = ArrayField(
        base_field=models.CharField(),
        default=list,
    )
```

This field would get populated with the set of all protocols specified by at least one port mapping, and can then be queried directly.

netbox/ipam/templates/ipam/widgets/port_mappings.html
```diff
@@ -0,0 +1,59 @@
+{% load i18n %}
+<div class="port-mapping-widget" data-name="{{ widget.name }}">
+  <input type="hidden" name="{{ widget.name }}" value="{{ widget.value }}">
```
jeremystretch, line 4:
```suggestion
  <input id="{{ widget.attrs.id }} type="hidden" name="{{ widget.name }}" value="{{ widget.value }}">
```

netbox/ipam/validators.py
```diff
@@ -3,6 +3,129 @@
 from django.utils.translation import gettext_lazy as _
 
 
+def split_port_mapping(mapping):
```
jeremystretch: Let's move anything not directly related to validation to the `ipam.utils` module.

netbox/ipam/models/services.py
```diff
@@ -17,42 +20,86 @@
...
+    @property
+    def protocol(self):
+        return self._legacy_protocol_ports[0]
+
+    @property
+    def ports(self):
+        return self._legacy_protocol_ports[1]
 
     @property
     def port_list(self):
```
jeremystretch: We might want to rename this to `port_mappings_list` for clarity, given that it returns both protocols & ports.

netbox/ipam/models/services.py
```diff
@@ -17,42 +20,86 @@
...
+    def clean(self):
+        super().clean()
+        # validate_port_mappings returns the canonical form (integer ports), so storing its result
+        # normalizes any entry that bypassed the form field (e.g. a raw REST payload of 'tcp/080').
+        self.port_mappings = validate_port_mappings(self.port_mappings)
+        self.__dict__.pop('_legacy_protocol_ports', None)  # invalidate the cached legacy view
+        if not self.port_mappings:
+            raise ValidationError({'port_mappings': _("At least one port mapping is required.")})
+
+    def apply_port_mapping_delta(self, add=None, remove=None):
```
jeremystretch: Can we split this into two methods, `add_port_mappings()` and `remove_port_mappings()`?

jeremystretch (changes requested):

arthanson:
@claude /review Only report on issues of potential concern; we don't need commentary on anything that doesn't require further consideration. Also see comments on previous review of Declined requested changes and reasoning.

arthanson:
Per Claudes comments, fixed 1-4.

netbox/ipam/validators.py
arthanson: The seen set exists to raise on duplicates (not silently dedupe) and to preserve input order - a set-return would drop both.

netbox/ipam/migrations/0095_add_multiple_service_port_mappings.py
```diff
@@ -0,0 +1,277 @@
...
+# Maintain the denormalized Service/ServiceTemplate._protocols column (the distinct set of protocols
+# present in port_mappings) via a BEFORE INSERT/UPDATE trigger, so a protocol-only filter can hit a GIN
+# index (_protocols && ['tcp']) instead of scanning a computed string form of port_mappings. A trigger
+# (rather than a Python save/clean hook) keeps the column correct for every write path, including
+# bulk_create and raw SQL. Both tables share one trigger function, which reads only NEW.port_mappings /
+# NEW._protocols — columns common to both. Existing rows are backfilled with a single set-based UPDATE
+# (computing _protocols directly, the same expression the trigger uses) *before* the triggers are
+# installed, so the backfill doesn't rewrite every row through the trigger. It runs after
+# populate_port_mappings, so port_mappings is already populated.
+INSTALL_PROTOCOLS_TRIGGER_SQL = """
```
jeremystretch: We should question whether the overhead here is really worth the limited value of retaining the `protocol` filter for the Service and ServiceTemplate models.

netbox/ipam/filtersets.py
```diff
@@ -1214,16 +1214,60 @@ def search(self, queryset, name, value):
         return queryset.filter(qs_filter)
 
 
+class ServicePortMappingFilterMixin(django_filters.FilterSet):
```
jeremystretch, line 1228: It feels like we're overcomplicating this. It's not clear to me how a user would filter for e.g. services which specify both `tcp/80` and `udp/53`, for example. Presumably the filters would be `?protocol=tcp&protocol=udp&port=80&port=53`, but that would also match on `tcp/53` and `udp/80`, correct

Also, the current implementation doesn't offer any mechanism for filtering by individual protocol & port mappings (e.g. `?mapping=tcp/80&mapping=udp/53`), which IMO is a far more common use case.

netbox/ipam/models/services.py
```diff
@@ -1,58 +1,142 @@
...
+        blank=True,
+        default=list,
     )
-    _ports_lowest = models.PositiveIntegerField(
-        null=True,
+    # Denormalized set of the distinct protocols present in port_mappings, maintained by a PostgreSQL
+    # trigger (see the migration) so a protocol-only filter can hit a GIN index (_protocols && ['tcp'])
+    # instead of scanning a computed string form of port_mappings. Ports are unbounded, so only protocols
+    # are denormalized; port and protocol+port filters query port_mappings directly.
+    _protocols = ArrayField(
```
jeremystretch: I'm not sure it makes sense to maintain a denormalized field for protocol but not for ports. Without it, we lose the ability to do numeric filtering on ports (e.g. `?protocol=tcp&port__gt=1000&port__lt=2000`).

netbox/ipam/filtersets.py
```diff
@@ -1214,16 +1214,60 @@ def search(self, queryset, name, value):
...
+        method='filter_protocol',
+    )
+    port = MultiValueNumberFilter(
+        method='filter_port',
+    )
+    # Negation lookup retained from when `protocol` was a model field: method-based filters don't get
+    # the char-based lookups (protocol__n, __ic, ...) auto-generated, and silently dropping protocol__n
+    # would widen existing saved filters/scripts rather than error. The __ic/__nic/__empty variants were
+    # never meaningful on a small fixed choice set and are intentionally left gone.
+    protocol__n = django_filters.MultipleChoiceFilter(
```
jeremystretch, line 1255: We've added a negation filter for `protocol` but not for `port`.

netbox/ipam/graphql/filters.py
```diff
@@ -345,8 +345,65 @@ class RouteTargetFilter(TenancyFilterMixin, PrimaryModelFilter):
...
+
+def _make_port_mapping_filters(model):
+    # strawberry_django only collects filter_field methods declared on the filter_type class itself (not
+    # from a mixin), so the two Service/ServiceTemplate filters are produced by this factory and assigned
+    # into each class body. This keeps the protocol/port correlation logic in a single place.
+    @strawberry_django.filter_field
+    def protocol(
+        self,
+        queryset,
+        value: list[Annotated['ServiceProtocolEnum', strawberry.lazy('ipam.graphql.enums')]],
```
jeremystretch, line 438: ServiceProtocolEnum is defined with a prefix of "role" for some reason. I don't think it should have a prefix.

```
ServiceProtocolEnum = strawberry.enum(ServiceProtocolChoices.as_enum(prefix='role'))
```

jeremystretch (changes requested):
I'm going to take a stab at cleaning up the filtering limitations I've noted below. I think we can get away without a denormalized field for protocols or ports.

jeremystretch:

jeremystretch:
1. Addressed
2. Addressed
3. Let's see how this plays out during evaluation. I appreciate the utility for it but I'm also concerned about UX.
4. This is an understood and accepted tradeoff
5. These are not a concern

jeremystretch (approved):

PR #22693, opened 2026-07-16
Closes #22682: Fix CachedScopeMixin cache fields cascading on ancestor deletion

### LINKED ISSUE #22682: Deleting a SiteGroup cascade-deletes prefixes scoped to member sites
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.4

### Python Version

3.12

### Steps to Reproduce

Description:
Deleting a `SiteGroup` unexpectedly deletes any `Prefix` scoped to a `Site` that belonged to that group. This is silent data loss — the user only deletes a grouping object, but loses prefixes scoped to the member sites.
Steps to reproduce (minimal):

Create a site group (`SG1`)

Create a site (`Site A`) and assign it to `SG1`

Create a prefix (`10.0.0.0/24`) and set its scope to `Site A`

Delete `SG1`
****

### Expected Behavior

Deleting the site group unassigns the group from its member sites (`Site.group` → null). Sites and their scoped prefixes remain intact.

### Observed Behavior

`Site A` survives but is unassigned from the group (correct). However, `10.0.0.0/24` is deleted, despite being scoped to the site — not the group.

### Root Cause
`CachedScopeMixin` (`netbox/dcim/models/mixins.py`) denormalizes a prefix's scope hierarchy into hidden cache columns for filtering. When a prefix is scoped to a site, `cache_related_objects()` populates `_site_group` with the site's parent group:

```
elif scope_type == apps.get_model('dcim', 'site'):
    self._region = self.scope.region
    self._site_group = self.scope.group   # cached from the site's parent group
    self._site = self.scope
```

But the `_site_group` cache field is declared with `on_delete=models.CASCADE`:

```
_site_group = models.ForeignKey(
    to='dcim.SiteGroup',
    on_delete=models.CASCADE,   # <-- causes the cascade delete
    blank=True,
    null=True
)
```

So Django's cascade collector deletes every prefix whose `_site_group` matches the group being deleted — even prefixes that were only ever scoped to a site. Compare with `Site.group` (`netbox/dcim/models/sites.py`), which is correctly `on_delete=models.SET_NULL`.

All four cache FKs (`_region`, `_site_group`, `_site`, `_location`) use `CASCADE`. `_site` / `_location` cascading is arguably defensible (the prefix's actual scope target is gone), but `_region` and `_site_group` cascading is not — those are ancestor grouping objects, not the prefix's scope target.

### Suggested fix
 Change the cache-only ancestor FKs (`_region`, `_site_group`) to `on_delete=models.SET_NULL`, or re-run `cache_related_objects()` on ancestor deletion rather than cascading. These fields exist purely for filter performance and should never drive deletion of the owning object.

bctiemann:
## Summary

Deleting a `SiteGroup` (or `Region`) was cascade-deleting `Prefix`/`Cluster`/`WirelessLAN` objects scoped to a `Site` that merely belonged to that group — even though the `Site` itself correctly survives (`Site.region`/`Site.group` are both `SET_NULL`). This is silent data loss: a user deleting a grouping object loses objects scoped to its member sites.

## Root cause

`CachedScopeMixin` (`netbox/dcim/models/mixins.py`) denormalizes a scoped object's ancestor hierarchy into cache columns (`_region`, `_site_group`, `_site`, `_location`) for filtering performance. When scope is a `Site`, `cache_related_objects()` populates `_region`/`_site_group` from the site's **ancestors** (`self.scope.region`, `self.scope.group`), not from the actual scope target. But both fields were declared `on_delete=models.CASCADE`, so Django's deletion collector deleted every object whose `_region`/`_site_group` matched the ancestor being deleted — even when that ancestor was never the object's actual scope.

## Fix

Change `_region` and `_site_group` to `on_delete=models.SET_NULL`. `_site` and `_location` are intentionally left as `CASCADE` — `Location.site` is itself `CASCADE`, so there's no scenario where an ancestor `Site` is deleted while a `Location` scope target survives it (unlike `Region`/`SiteGroup`, which use `SET_NULL` on `Site`, creating the surviving-ancestor state that exposed this bug).

Deletion of a `Region`/`SiteGroup` that **is** the actual scope (not just an ancestor) continues to work correctly — that's handled independently via the scoped model's own `GenericRelation` (e.g. `Region.prefixes` in `dcim/models/sites.py`), which is unaffected by the cache field's `on_delete` setting. I confirmed this by temporarily reverting only the field change and observing that the "scoped directly to the Region/SiteGroup" sanity tests still passed while the "ancestor deleted" tests failed — reproducing the exact reported bug and confirming the fix doesn't regress legitimate cascade behavior.

## Migrations

`CachedScopeMixin` is used by three models across three apps: `Prefix` (ipam), `Cluster` (virtualization), and `WirelessLAN` (wireless). Generated a migration for each. Verified via `sqlmigrate` that all three are no-ops at the database level — Django's `on_delete` is enforced entirely in the ORM's deletion collector, not via a native FK constraint action, so applying these requires no table rewrite or lock.

## Tests

Added regression tests to `ipam/tests/test_models.py`, `virtualization/tests/test_models.py`, and `wireless/tests/test_models.py` (8 new tests total):
- For each of `Prefix`/`Cluster`/`WirelessLAN`: deleting a `SiteGroup` or `Region` that a scoped `Site` belongs to must not delete the scoped object.
- For `Prefix` only: a sanity check that an object scoped **directly** to a `Region`/`SiteGroup` is still correctly deleted when that Region/SiteGroup is deleted (confirming the legitimate cascade path, driven by `GenericRelation`, is untouched).

Verified:
- All 8 new tests pass with the fix.
- Reverting only the `on_delete` change reproduces the bug: the 6 "ancestor deleted" tests fail, while the 2 "scoped directly to it" tests still pass — confirming the tests catch the actual reported issue.
- `ipam.tests.test_models` (167), `virtualization.tests.test_models`, `wireless.tests.test_models` (167 combined), `dcim.tests.test_models` (82), and `dcim.tests.test_signals` (21, includes existing coverage for the cache-resync signal on Site/Location rename) all pass with no regressions.
- `makemigrations --check` reports no missing migrations.
- `ruff check` passes.


bctiemann:

bctiemann:
@claude /review changes since previous review.

pheus (changes requested):
Thanks for addressing the incorrect behavior. The underlying fix looks reasonable, but I’m requesting changes to bring the implementation and tests in line with NetBox’s conventions.

Please:

1. Remove the issue reference from the production comment and reduce it to a concise explanation of the invariant.
2. Organize the tests by model rather than under `*CachedScopeTestCase` classes:
   - move the Prefix tests into the existing `PrefixTestCase`;
   - use `ClusterTestCase` for the Cluster tests;
   - use `WirelessLANTestCase` for the WirelessLAN tests.
3. Remove the lengthy class and method docstrings. They mostly repeat the issue, implementation, test setup, and assertions.
4. Make the remaining regression comments consistent. A single short `# Regression test for #22682` marker is sufficient where the issue reference is useful.

The tests themselves provide useful coverage, but their current organization and commentary add unnecessary noise and make the model test suites harder to scan.

pheus (approved):
Thanks for the follow-up! This looks good to me now!

PR #22696, opened 2026-07-16
Fixes #21988: Enforce view permissions when referencing related object by attributes in REST API

### LINKED ISSUE #21988: Authorization bypass in nested object resolution via unrestricted nested lookup filters

jeremystretch:
### Fixes: #21988

- Tweak `get_related_object_by_attrs()` to enforce the user's view permission on the underlying queryset when referencing an object by any attribute other than its numeric ID
- Update `to_internal_value()` on BaseModelSerializer and WritableNestedSerializer accordingly
- Add relevant tests & extend assigned permissions where needed
- Add a note in the REST API documentation

### DIFF
--- a/docs/integrations/rest-api.md
+++ b/docs/integrations/rest-api.md
@@ -168,6 +168,9 @@ Or by a set of attributes which uniquely identify the rack:
 
 Note that if the provided parameters do not return exactly one object, a validation error is raised.
 
+!!! note "Permissions"
+    When a related object is referenced by a set of attributes, the lookup is restricted to only those objects which the requesting user has permission to view. This prevents the enumeration of objects by their attributes. Referencing a related object directly by its numeric ID is always permitted, regardless of the user's view permissions for that object.
+
 ### Generic Relations
 
 Some objects within NetBox have attributes which can reference an object of multiple types, known as _generic relations_. For example, an IP address can be assigned to either a device interface _or_ a virtual machine interface. When making this assignment via the REST API, we must specify two attributes:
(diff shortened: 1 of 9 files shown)


jeremystretch:

jeremystretch:
1. IIRC we don't support the `pk` alias for IDs in the REST API. At least, the PK for an object is always conveyed as its `id` attribute.
2. This should be okay to ignore, as serializers are always initialized with the request context in production usage. (Permissions enforcement is not possible at all without a context present.)

bctiemann (approved):

PR #22706, opened 2026-07-17
Add optional Front Image, Rear Image, and Image Count columns to Device Type and Module Type tables

### LINKED ISSUE #22030: Add optional Image Indicators/Counters to Device Type and Module Type Tables
### NetBox version

v4.5.8

### Feature type

New functionality

### Proposed functionality

In the Device Type and Module Type views today, there are a number of configurable columns to filter on. (e.g Airflow, Full Depth, Number of Interfaces)

I'd like to see whether it's possible to add optional columns (which can be added to the Device Types and Module Types tables) for the following:

- Front Image (Y/N)
- Rear Image (Y/N)
- Images (Counter)

These columns would operate in the same way that a "Full Depth" or "Interfaces" counter exist today. Users could choose to add these columns to the table or not as they wish. 

### Use case

Currently there's no mechanism to see what device/module types have images or not, or to see whether there's any image attachments for a device/module type. 

We are currently in the process of updating our NetBox instance to add elevation images and module images where possible to all of our device types (~500 devices) and it would be convenient to be able to filter in the Device Type view to have a visual indicator as to whether a Front View, Rear View or other Images are present. Right now the only way to check is via individually selecting each device type. 

### Database changes

N/A - Do not believe so

### External dependencies

N/A

TallesH13:
## What this PR does

Adds three optional columns to the Device Type list view (Front Image, Rear Image and Images as the attached images counter) and one optional column to the Module Type list view (Images column, as the attached images counter)

## Changes
- Added `front_image`, `rear_image` and `image_count` columns to the Device Type table and `image_count` for Module Type table
- Updated the corresponding filterset and view logic to support these columns
- Columns are opt-in via the existing "Configure table" UI, consistent with other optional columns in Netbox

## Testing 
Verified manually and via local Cypress tests that:
- The three columns appear as available options in "Configure Table"
- Selecting and applying them renders the correct data in the table
- Existing table behavior is unaffected

## Screenshots
- Device Type with changes
WhatsApp Image 2026-07-17 at 11 46 44
WhatsApp Image 2026-07-17 at 11 45 59

- Module Type with changes
WhatsApp Image 2026-07-17 at 11 47 08
WhatsApp Image 2026-07-17 at 11 43 48

---

Co-authored by: Christiane Giestas <christianegiestas@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ yarn-error.log*
 /dist/
 /build/
 *.egg-info/
+
+# Ruff's locally generated cache
+.ruff_cache
--- a/netbox/dcim/filtersets.py
+++ b/netbox/dcim/filtersets.py
@@ -722,6 +722,10 @@ class DeviceTypeFilterSet(PrimaryModelFilterSet):
         method='_inventory_items',
         label=_('Has inventory items'),
     )
+    has_images = django_filters.BooleanFilter(
+        method='_has_images',
+        label=('Has images'),
+    )
 
     class Meta:
         model = DeviceType
@@ -793,6 +797,9 @@ class DeviceTypeFilterSet(PrimaryModelFilterSet):
 
     def _inventory_items(self, queryset, name, value):
         return queryset.exclude(inventoryitemtemplates__isnull=value)
+
+    def _has_images(self, queryset, name, value):
+        return queryset.exclude(images__isnull=value)
 
 
 @register_filterset
@@ -866,6 +873,10 @@ class ModuleTypeFilterSet(AttributeFiltersMixin, PrimaryModelFilterSet):
         method='_module_bays',
         label=_('Has module bays'),
     )
+    has_images = django_filters.BooleanFilter(
+        method='_has_images',
+        label=_('Has images'),
+    )
 
     class Meta:
         model = ModuleType
@@ -918,6 +929,9 @@ class ModuleTypeFilterSet(AttributeFiltersMixin, PrimaryModelFilterSet):
 
     def _module_bays(self, queryset, name, value):
         return queryset.exclude(modulebaytemplates__isnull=value)
+
+    def _has_images(self, queryset, name, value):
+        return queryset.exclude(images__isnull=value)
 
 
 class DeviceTypeComponentFilterSet(django_filters.FilterSet):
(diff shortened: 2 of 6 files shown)


jeremystretch:

netbox/dcim/filtersets.py
```diff
@@ -722,6 +722,10 @@ class DeviceTypeFilterSet(PrimaryModelFilterSet):
         method='_inventory_items',
         label=_('Has inventory items'),
     )
+    has_images = django_filters.BooleanFilter(
+        method='_has_images',
+        label=('Has images'),
```
arthanson, line 727: Need _( internationalization here

netbox/dcim/views.py
```diff
@@ -1409,6 +1409,22 @@ class DeviceTypeListView(generic.ObjectListView):
     filterset_form = forms.DeviceTypeFilterForm
     table = tables.DeviceTypeTable
 
+    def get_table(self, data, request, bulk_actions=True):
```
arthanson, line 1412: This is duplicated below, can we extract this out into a common place, maybe an ImageCountAnnotationMixin to keep it DRY

.gitignore
```diff
@@ -69,3 +69,6 @@ yarn-error.log*
 /dist/
 /build/
 *.egg-info/
+
+# Ruff's locally generated cache
+.ruff_cache
```
arthanson, line 74: This is out of scope for the PR, can we remove this

arthanson (changes requested):
Thanks @TallesH13 a couple issues to clean up. 

Please add some tests: The PR adds a filter, filter form fields, and table columns with zero test coverage:
- has_images filter tests for both models, mirroring test_has_front_image / test_has_rear_image at netbox/dcim/tests/test_filtersets.py:1520-1529.
- A table-render test exercising image_count when the column is selected (this is the path with the most logic).

Column-visibility detection in get_table() is incomplete?  To replicate:
1. Add a front and rear image and image attachment to a device type.
2. Go to configure table and add these columns. (note the display the checkmarks and count)
3. In the configure table dropdown, save this configuration.
4. Go to configure table and remove these fields.
5. I the Configure table dropdown load your saved configuration.
6. Note the columns show but they appear with a dash (no checkmark or count)
 
Table.configure() resolves visible columns from more sources (netbox/netbox/tables/tables.py:189-200 and mixins.py:54):
- settings.DEFAULT_USER_PREFERENCES
- ?include_columns=image_count (query param)
- ?tableconfig_id=.

In all three paths the image_count column is displayed but the queryset isn't annotated.

jeremystretch:
@TallesH13 are you still interested in working on this? If not, please close your PR so the issue can be reassigned.

PR #22708, opened 2026-07-18
Fix #22671 : Enable multi-selection for Site Groups and other scopes in VLAN Groups filter

### LINKED ISSUE #22671: Selecting multiple Site Groups in the VLAN Groups filter does not display the full output.
### NetBox Edition

NetBox Community

### NetBox Version

NetBox Community v4.6.3-Docker-5.0.1 

### Python Version

3.12

### Steps to Reproduce

1. Create site group "POP"
2. Create site group "A" with parent "POP"
3. Create site group "B" with parent "POP"
4. Create Vlan Group "A VLANS" with Scope->Site group "A"
5. Create Vlan Group "B VLANS" with Scope->Site group "B"

Tested at local installation + demo netbox
1. Open https://xxx/ipam/vlan-groups/
2. Click on "Filters" tab which is near 'Results" tab
3. Choose in "Site Group" field both groups - "A" and "B" at the same time.

### Expected Behavior

Expect to see VLAN groups "A VLANS" and "B VLANS" in output table according to site groups.

### Observed Behavior

I see only only the last chosen VLAN group in the  field, but not both.

LujainYasser1:
### Description

The filter form for VLAN Groups sends arrays of IDs for site groups and other scopes, but `VLANGroupFilterSet` was implementing them using `NumberFilter`. This caused the backend to drop multiple selections or fail to process them properly.

I replaced `NumberFilter` with `MultiValueNumberFilter` across all scope fields in `VLANGroupFilterSet` (including `site_group`, `region`, `site`, `location`, `rack_group`, `rack`, `cluster_group`, and `cluster`). I also refactored the `filter_scope` method to use `scope_id__in=value` so it can cleanly evaluate multiple lookups at once.

### Checklist

- [X] My code follows the code style of this project (self-documenting, no unnecessary comments).
- [X] Syntax checking passed successfully.

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -1227,7 +1227,6 @@ class ServiceTemplateFilterSet(PrimaryModelFilterSet):
         )
         return queryset.filter(qs_filter)
 
-
 @register_filterset
 class ServiceFilterSet(ContactModelFilterSet, PrimaryModelFilterSet):
     parent_object_type = MultiValueContentTypeFilter()
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -946,7 +946,6 @@ class FHRPGroupAssignmentFilterSet(ChangeLoggedModelFilterSet):
 @register_filterset
 class VLANGroupFilterSet(OrganizationalModelFilterSet, TenancyFilterSet):
     scope_type = MultiValueContentTypeFilter()
-
     region = MultiValueNumberFilter(
         method='filter_scope'
     )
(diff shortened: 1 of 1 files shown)


jeremystretch:
@LujainYasser1 could you please drop a comment on #22671 so the issue can be assigned to you? Thanks.

pheus:

pheus (changes requested):
Thanks for opening this PR!

Could you please take a look at Claude's review comments and either address the points it raised or reply with a short explanation where you think a comment doesn’t apply? That context would be really helpful for the review.
Finally, please address the tests and lint failures. Thanks!

jeremystretch:
@LujainYasser1 are you still interested in working on this? If not, please close your PR so the issue can be reassigned.

PR #22710, opened 2026-07-19
Fixes #22707: Fix `{vc_position}` resolution for port mappings and interface bridges

### LINKED ISSUE #22707: Port mapping causes error when {vc_position} is used on device type templates
### NetBox Edition

NetBox Community

### NetBox Version

4.6.5

### Python Version

3.12

### Steps to Reproduce

1. Create a Manufacturer with:
   * Name = Test
   * Slug = test
2. Create a Device Type with:
   * Manufacturer = Test
   * Model = Test
   * Slug = test
3. Create a Rear Port Template with:
   * Name = {vc_position}
   * Type = LC
4. Create a Front Port Template with:
   * Name = {vc_position}
   * Type = LC
   * Positions = 1
   * Rear ports = {vc_position}:1
5. Create a Device Role with:
   * Name = Test
   * Slug = test
6. Create a Site with:
   * Name = Test
   * Slug = test
7. Create a Virtual Chassis with:
   * Name = Test
8. Create a Device with:
   * Device role = Test
   * Device type = Test
   * Site = Test
   * Virtual chassis = Test
   * Position = 1

### Expected Behavior

The device is created successfully and the front port is mapped to the rear port.

### Attribute Error at /dcim/devices/add/

'NoneType' object has no attribute 'device_id'

**Request Method:**	POST
**Request URL:**	http://netbox/dcim/devices/add/
**Django Version:**	6.0.7
**Exception Type:**	AttributeError
**Exception Value:**	
`'NoneType' object has no attribute 'device_id'`
**Exception Location:**	/path/to/netbox/netbox/dcim/utils.py, line 164, in create_port_mappings
**Raised during:**	dcim.views.DeviceEditView
**Python Executable:**	/path/to/netbox/venv/bin/python3
**Python Version:**	3.12.3
**Python Path:**	`['/path/to/netbox/netbox',
 '/usr/lib/python312.zip',
 '/usr/lib/python3.12',
 '/usr/lib/python3.12/lib-dynload',
 '/path/to/netbox/venv/lib/python3.12/site-packages']`
**Server time:**	Fri, 17 Jul 2026 15:57:31 -0400

Since the templates are instantiated directly on the device, `module` is `None` and `device` is not being passed to the `resolve_name` call, so it is also defaulting to `None`, which causes the `{vc_position}` placeholder to always be replaced with the fallback value, `0` in this case, and the Front Port is not found in the cache.

https://github.com/netbox-community/netbox/blob/ebee3578b90901ba69ea646815f9b0662f627726/netbox/dcim/utils.py#L160-L161

pheus:
### Closes: netbox-community/netbox#22707

This PR ensures that `{vc_position}` is resolved consistently when replicating port mappings and assigning interface bridges.

It passes the device context to `resolve_name()` in both code paths, ensuring that relationship lookups use the same names as the instantiated device components. This prevents device creation from failing when component template names reference the device’s virtual chassis position.

### DIFF
--- a/netbox/dcim/tests/test_models.py
+++ b/netbox/dcim/tests/test_models.py
@@ -2570,6 +2570,70 @@ class VCPositionTokenTestCase(TestCase):
         interface = device.interfaces.get(name='ge-2/1/0')
         self.assertEqual(interface.label, 'Member 2 / Slot 1')
 
+    @tag('regression')  # Ref: #22707
+    def test_vc_position_token_interface_bridge_device_type_template(self):
+        site = Site.objects.first()
+        device_type = DeviceType.objects.first()
+        device_role = DeviceRole.objects.first()
+
+        bridge_template = InterfaceTemplate.objects.create(
+            device_type=device_type,
+            name='br-{vc_position}',
+            type='bridge',
+        )
+        InterfaceTemplate.objects.create(
+            device_type=device_type,
+            name='ge-{vc_position}/0/1',
+            type='1000base-t',
+            bridge=bridge_template,
+        )
+        vc = VirtualChassis.objects.create(name='Test VC 5')
+        device = Device.objects.create(
+            name='Device VC 5', device_type=device_type, role=device_role,
+            site=site, virtual_chassis=vc, vc_position=5,
+        )
+
+        interface = device.interfaces.get(name='ge-5/0/1')
+        self.assertEqual(interface.bridge, device.interfaces.get(name='br-5'))
+
+    @tag('regression')  # Ref: #22707
+    def test_vc_position_token_port_mapping_device_type_template(self):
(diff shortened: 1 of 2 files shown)


pheus:

pheus:
@claude /review

You're right that the existing `dict.get()` calls would result in a less descriptive exception if an instantiated component were unexpectedly missing.

However, that behavior predates this PR, and this change does not introduce a new missing-component path. The front and rear ports are instantiated using the device context, and this change makes port-mapping replication use the same resolution context. Also, each `PortTemplateMapping` requires both a front and rear port template, so a missing side is not a valid mapping state.

I'd prefer to keep this PR focused on resolving the issue. More explicit invariant or error reporting could be considered separately if needed.

arthanson (approved):

PR #22713, opened 2026-07-19
Fixes #22712: Fix validation error highlighting for dropdown fields

### LINKED ISSUE #22712: Dropdown fields are not highlighted when validation fails
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.5

### Python Version

3.12

### Steps to Reproduce

1. Create an FHRP group.
2. Open a device or virtual machine interface and assign the FHRP group to it.
3. From the same interface, click **Assign Group** again.
4. Select the same FHRP group and submit the form.

### Expected Behavior

The **Assignment already exists** validation message should be displayed, and the Group dropdown should be highlighted with NetBox's standard red error border.

### Observed Behavior

The validation message is displayed below the Group dropdown, but the visible dropdown is not highlighted.

pheus:
### Closes: netbox-community/netbox#22712

This updates Tom Select dropdowns to show NetBox's standard red error border when server-side validation marks the underlying field with `aria-invalid="true"`.

The existing `.is-invalid` handling remains unchanged for widgets that explicitly use Bootstrap's validation state. This keeps validation feedback consistent with regular form fields without adding Bootstrap's additional validation icon.


pheus:

pheus:
@claude /review

NetBox's standard select template, including `APISelect`, adds the `.form-select` class to the native `<select>`. Tom Select copies that class to the generated `.ts-wrapper`.

For a wrapper with `.form-select`, Tom Select's Bootstrap theme explicitly applies `border: none !important` to the child `.ts-control`, so the visible border belongs to the wrapper. The new rule therefore changes the existing wrapper border rather than adding a second border.

Targeting `.ts-control` with `border-color` would not have a visible effect in this case because its border is disabled with `!important`. Applying the border to `.ts-wrapper` also matches how Tom Select's own Bootstrap validation styling targets invalid controls.

I've verified the current styling against the reported server-side validation case, so I believe the wrapper remains the correct target here.

jeremystretch (approved):

PR #22715, opened 2026-07-20
Closes #19273: Enable selection of VLANs scoped to a Device's Cluster or Cluster Group

### LINKED ISSUE #19273: Unable to select VLANs on Device Interfaces where they are scoped by Cluster or Cluster Groups
### NetBox version

v4.2.4

### Feature type

### Proposed functionality

It was suggested in [issue #12764](https://github.com/netbox-community/netbox/issues/12764) to make it a formal feature request. This was also reported in [issue #12690](https://github.com/netbox-community/netbox/issues/12690).

Assigning VLANs from a VLAN groups should work in all context in the Netbox UI. At the moment, when using the API or Netbox Scripts we can properly assign VLAN(s) to interface(s) based on specific VLAN Groups, but we can't from [dcim/interfaces/](https://demo.netbox.dev/dcim/interfaces).

Here's the workflow:
1. Create a device.
2. Assign the device to a virtualization cluster.
3. Create a VLAN group attached to the cluster.
4. Create VLAN(s) in the VLAN group.
5. Can't assign VLAN(s) from the VLAN group on the device interfaces manually in the GUI.
6. Can assign VLAN(s) using the API or a Netbox Scripts.

When we modify an interface and want to assign a VLAN based on a VLAN group, the "dropdown menu/list" of Untagged VLAN and Tagged VLANs is empty even if VLANs exist in the selected VLAN group.

Image

### Use case

We are allowed to create VLANs and VLAN groups, we should be able to assign any of those VLANs in the Netbox Web UI.

Creating VLAN groups is common in virtualization or when we group devices that shares VLANs: VTP domain (people still uses those), any Network Fabric or cluster of devices (or servers).

### Database changes

I don't think any database change is required, the API already allows proper assignment of VLANs that are member of a group. This is actually how we work around the issue at the moment: we have created a form with a Netbox Scripts that allows us to add/remove VLANs to an interface manually. We override the UI limitation.

I feel the current limitation, is mainly due to the Web UI vlan lookup logic: when selecting a VLAN group, the list/dropdown is empty (the query may need to be fine tuned).

### External dependencies

N/A

pheus:
### Closes: netbox-community/netbox#19273

This PR updates VLAN availability checks for devices assigned to a cluster. VLANs belonging to groups scoped to the device's cluster or cluster group are now included, bringing this behavior in line with the existing site, location, and rack scope handling.

### DIFF
--- a/netbox/ipam/filtersets.py
+++ b/netbox/ipam/filtersets.py
@@ -1070,7 +1070,7 @@ class VLANFilterSet(PrimaryModelFilterSet, TenancyFilterSet):
         method='get_for_site'
     )
     available_on_device = django_filters.ModelChoiceFilter(
-        queryset=Device.objects.all(),
+        queryset=Device.objects.select_related('cluster'),
         method='get_for_device'
     )
     available_on_virtualmachine = django_filters.ModelChoiceFilter(
--- a/netbox/ipam/querysets.py
+++ b/netbox/ipam/querysets.py
@@ -298,6 +298,18 @@ class VLANQuerySet(RestrictedQuerySet):
 
         # Find all relevant VLANGroups
         q = Q()
+        if device.cluster_id:
+            # The Device's physical scope is evaluated below. For valid assignments,
+            # the Cluster's physical scope is already represented by that hierarchy.
+            q |= Q(
+                scope_type=ContentType.objects.get_by_natural_key('virtualization', 'cluster'),
+                scope_id=device.cluster_id
+            )
+            if device.cluster.group_id:
+                q |= Q(
+                    scope_type=ContentType.objects.get_by_natural_key('virtualization', 'clustergroup'),
+                    scope_id=device.cluster.group_id
+                )
         if device.site.region:
             q |= Q(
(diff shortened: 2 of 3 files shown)


pheus:

pheus:
@claude /review

The difference from `get_for_virtualmachine()` is intentional. A device's physical scope is evaluated directly from the device immediately below this block. NetBox also validates that a device assigned to a site- or location-scoped cluster occupies that same site or location, so the different-location example would be an invalid assignment. Region and site-group scopes containing the device are likewise already covered by the device's site hierarchy.

Virtual machines need the additional cluster-scope handling because their physical scope can be derived from the cluster itself. I'll add a short comment here to make that distinction clearer for future readers.

The additional cluster lookup is a fair optimization point. This is a singular `ModelChoiceFilter`, rather than an N+1 query, and only `cluster.group_id` is accessed, so selecting the `ClusterGroup` itself is unnecessary. I've updated the queryset to use `select_related('cluster')`.

The virtual-machine filter uses a separate queryset, so any equivalent optimization for that path would need to be handled independently.

jnovinger (approved):

Issue #22592, opened 2026-07-02
Enable plugins to extend core GraphQL API entry points


jeremystretch:
### NetBox version

v4.6.4

### Feature type

New functionality

### Proposed functionality

Provide a supported plugin extension point for **adding fields and filters to existing core GraphQL types** (e.g. the `Device` type), so a plugin's related data can be traversed within a single query rooted at a core object rather than requiring a separate top-level query.

Today, plugins can only contribute *new top-level query fields* — `PluginConfig.graphql_schema` classes are spread into the root `Query` via multiple inheritance in `netbox/netbox/graphql/schema.py`. There is no way for a plugin to add a field/relation to an already-defined core type. A plugin model with an FK/GFK to `dcim.Device` therefore cannot surface its reverse relation on the `Device` GraphQL type; clients must issue a second query against the plugin's own type.

**Proposed mechanism** (modeled on the existing `PluginTemplateExtension` pattern):

1. Plugins ship **output-type mixins** and **filter mixins**, each declaring a `models` attribute (list of lowercased `app_label.model` labels), in `graphql/type_extensions.py` and `graphql/filter_extensions.py`:

   ```python
   # my_plugin/graphql/type_extensions.py
   @strawberry.type
   class DeviceGraphQLExtension:
       models = ['dcim.device']

       @strawberry_django.field(prefetch_related='widgets')
       def widgets(self) -> list[Annotated['WidgetType', strawberry.lazy('my_plugin.graphql.types')]]:
           return self.widgets.all()

   type_extensions = [DeviceGraphQLExtension]
   ```

2. New `PluginConfig` resources (`graphql_type_extensions`, `graphql_filter_extensions`) with default paths `graphql.type_extensions` / `graphql.filter_extensions`, loaded in `PluginConfig.ready()` (alongside the existing `graphql_schema` handling in `netbox/netbox/plugins/__init__.py`).

3. New `register_graphql_type_extensions()` / `register_graphql_filter_extensions()` functions in `netbox/netbox/plugins/registration.py` (mirroring `register_template_extensions`) that collect classes into per-model registry buckets keyed by model label.

[...]

PR #22721, opened 2026-07-20
Enable plugins to extend core GraphQL API

arthanson:
### Closes: #22592 

Adds a supported plugin extension point for injecting fields and filters onto existing core GraphQL types, so a plugin's related data can be traversed within a single query rooted at a core object rather than requiring a separate top-level query.

## Changes

- New PluginConfig resources graphql_type_extensions / graphql_filter_extensions.
- New register_type() / register_filter() wrappers splice registered mixins into core types.
- Extensions are strictly additive: mixins are appended after the core bases. Two-extension collisions resolve to the first-loaded plugin. Both are logged.
- Tests and documentation updated.

Plugins ship @strawberry.type mixins declaring a models attribute; related-object resolvers scope the prefetch with RestrictedPrefetch to enforce object permissions:

```
  @strawberry.type
  class DeviceTypeExtension:
      models = ['dcim.device']

      @strawberry_django.field(
          prefetch_related=lambda info: RestrictedPrefetch(
              'widgets', info.context.request.user, 'view', queryset=Widget.objects.all()
          ),
      )
      def widgets(self) -> list[Annotated['WidgetType', strawberry.lazy('my_plugin.graphql.types')]]:
          return self.widgets.all()

  type_extensions = [DeviceTypeExtension]
```


arthanson:

arthanson:
Claude review issues fixed.

bctiemann (approved):

PR #22722, opened 2026-07-20
Speed up bulk-delete objects

### LINKED ISSUE #22497: Bulk-deleting many objects is extremely slow due to per-object cascade and N+1 counter updates
### NetBox Edition

NetBox Community

### NetBox Version

4.5.1 (reproduced; relevant code still present on `main` as of June 2026)

### Python Version

3.12

### Steps to Reproduce

> Environment (for reproduction context only — **this is not Docker- or****<br>****Kubernetes-specific**, the issue is in NetBox's Python/ORM code paths and<br>reproduces on any installation including bare-metal): netbox-docker images on<br>Kubernetes, PostgreSQL + Redis, async deletes handled by an RQ worker pod.<br>Because the bulk delete runs as a background RQ job, the symptom most operators<br>hit first is the job exceeding `RQ_DEFAULT_TIMEOUT` and rolling back.

1. Have a reasonably large DCIM dataset (in our case \~1264 devices / \~88k interfaces, i.e. \~70–88 interfaces per device).
2. Select \~100 devices in the device list and run the **Delete selected** bulk action.
3. Observe the background RQ job duration (and `RQ_DEFAULT_TIMEOUT`).

### Expected Behavior      

Deleting 100 devices should complete in a few seconds. The actual SQL work is tiny.

### Observed Behavior

The delete takes **\~8 minutes** for 100 devices and frequently exceeds the default<br>`RQ_DEFAULT_TIMEOUT` of 300 s, causing `JobTimeoutException` and a full rollback<br>(0 objects deleted). Profiling shows the time is **not** spent in heavy SQL — every<br>individual query is fast — but in the sheer number of round-trips and Python signal<br>handling.

Short version of the cause (full analysis in the linked PR): deleting 100 devices<br>cascades to \~8149 objects, and two N+1 patterns dominate — (1) `BulkDeleteView.post()`<br>deletes one object at a time (100 separate cascades instead of one), and (2)<br>`post_delete_receiver` in `utilities/counters.py` issues one `UPDATE` per cascaded<br>child to decrement a parent counter (`_interface_count`) even when that parent is<br>itself being deleted in the same batch (\~8886 no-op UPDATEs for 100 devices). Both are<br>code-level, dataset-independent (any install reproduces them with enough child objects).

arthanson:
### Closes: #22497 

Bulk-deleting objects that own many counter-tracked children took several minutes and could exceed RQ_DEFAULT_TIMEOUT, rolling back. The cost was DB round-trips, not query execution:

- post_delete_receiver issued a counter UPDATE on the parent for every child deleted — even when the parent itself was being deleted in the same operation (a no-op UPDATE).
- pre_delete_receiver issued an EXISTS SELECT per child to guard against double-deletion.

For 100 devices this was ~8,800 pointless UPDATEs plus thousands of redundant SELECTs.

## Fix

Use the delete signal's origin to detect when a tracked parent is itself being deleted, and skip both the counter UPDATE and the existence check in that case.

- utilities/counters.py: add _parent_is_being_deleted(); skip the redundant UPDATE in post_delete_receiver and the EXISTS query in pre_delete_receiver when the parent is going away. The double-delete guard (_previously_removed, #14081) is preserved for surviving parents.
- netbox/models/deletion.py: DeleteMixin.delete() now passes origin=self to CustomCollector (matching Django's Model.delete), so the skip applies to every delete path — UI bulk delete, REST API bulk delete, and single-object delete.
- Added a regression test asserting a Device delete issues no per-Interface counter UPDATEs while the surviving DeviceType.device_count still decrements.

## Out of scope

The per-object delete loop in BulkDeleteView was intentionally left as-is. Batching it through a single collector would bypass custom Model.delete() methods (e.g. Device.delete() removes uploaded image files from disk; also Cable, VirtualChassis, etc.), silently dropping those side effects. The counter fix removes the dominant cost without that risk.

### DIFF
--- a/netbox/netbox/models/deletion.py
+++ b/netbox/netbox/models/deletion.py
@@ -79,7 +79,9 @@ class DeleteMixin:
                 )
             )
 
-        collector = CustomCollector(using=using)
+        # Pass origin=self (matching Django's Model.delete) so signal receivers can tell that
+        # cascaded child objects are being deleted as part of deleting this object.
+        collector = CustomCollector(using=using, origin=self)
         collector.collect([self], keep_parents=keep_parents)
 
         return collector.delete()
--- a/netbox/utilities/counters.py
+++ b/netbox/utilities/counters.py
@@ -70,6 +70,11 @@ def _parent_is_being_deleted(origin, parent_model, parent_pk):
     operation. In that case, decrementing its counter is wasted work: the parent row is going away,
     so the UPDATE would be a no-op. Skipping it avoids an N+1 storm of pointless UPDATEs when a
     parent with many tracked children is deleted (e.g. a Device with thousands of Interfaces).
+
+    Note: only the *direct* parent is detected, since `origin` is just the top-level object/queryset
+    delete() was called on. In a deeper cascade (DeviceType -> Device -> Interface) `origin` stays
+    the DeviceType, so intermediate Devices' interface counters still get the (harmless) no-op
+    UPDATE. Suppressing that would require the full deletion set, which the signals don't expose.
     """
     if origin is None:
         return False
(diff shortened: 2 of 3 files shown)


arthanson:

PR #22724, opened 2026-07-21
Fixes #22719: Correct malformed IP value validation

### LINKED ISSUE #22719: REST API returns 500 instead of a validation error for malformed IP address and prefix values
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.5

### Python Version

3.12

### Steps to Reproduce

1. Send a POST request to the IP addresses endpoint with a malformed address value:

```
curl -X POST http://localhost:8000/api/ipam/ip-addresses/ \
   -d '{"address": "invalid"}'
```

2. Observe 500 error

Same issue with the `/api/ipam/prefixes/` POST endpoint:

   ```
   curl -X POST http://localhost:8000/api/ipam/prefixes/ \
     -d '{"prefix": "invalid"}'
   ```

### Expected Behavior

Both requests return HTTP 400 with the intended validation messages: "Invalid IP address format: invalid"

### Observed Behavior

Both requests return HTTP 500:

```
File ".../ipam/api/field_serializers.py", line 43, in to_internal_value
    raise serializers.ValidationError(_("Invalid IP prefix format: {data}").format(data))
KeyError: 'data'
```

pheus:
### Closes: netbox-community/netbox#22719

Corrects REST API validation for malformed IP address and prefix values.

The validation messages use a named `{data}` placeholder but supplied the value as a positional argument. This raised `KeyError` before DRF could return the intended validation error, resulting in an HTTP 500 response.

#### Changes

* Pass `data` as a named formatting argument in `IPAddressField`.
* Pass `data` as a named formatting argument in `IPNetworkField`.
* Add regression tests confirming malformed values return HTTP 400 with the<br>expected validation messages.

### DIFF
--- a/netbox/ipam/api/field_serializers.py
+++ b/netbox/ipam/api/field_serializers.py
@@ -20,7 +20,7 @@ class IPAddressField(serializers.CharField):
         try:
             return IPNetwork(data)
         except AddrFormatError:
-            raise serializers.ValidationError(_("Invalid IP address format: {data}").format(data))
+            raise serializers.ValidationError(_("Invalid IP address format: {data}").format(data=data))
         except (TypeError, ValueError) as e:
             raise serializers.ValidationError(e)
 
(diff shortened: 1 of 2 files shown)


pheus:

bctiemann (approved):

PR #22725, opened 2026-07-21
Fixes #22717: Correct Device Cluster Location validation

### LINKED ISSUE #22717: Device validation raises KeyError when assigned cluster is scoped to a different location
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.5

### Python Version

3.12

### Steps to Reproduce

1. Create a site "Site 1".
2. Create two locations under Site 1: "Location A" and "Location B".
3. Create a cluster type and a cluster "Cluster 1" with its scope set to Location A.
4. Create a device in Site 1 at Location B (any device type and role).
5. Edit the device and assign it to Cluster 1, then save.

### Expected Behavior

The form rejects the assignment with a validation error: "The assigned cluster belongs to a different location (Location A)".

### Observed Behavior

The save fails with a server error instead of a validation message:

```
KeyError: 'location'
```

The ValidationError message at `dcim/models/devices.py:947` interpolates a `{location}` placeholder but passes the value as `site=`, so building the message raises `KeyError` before the validation error can be shown.

pheus:
### Closes: netbox-community/netbox#22717

Corrects device validation when its assigned cluster is scoped to a different location.

The validation message expects a `location` formatting argument, but the value was supplied using the `site` keyword. This caused a `KeyError` before the intended field validation error could be raised.

Adds a regression test covering two locations within the same site and verifies the expected cluster validation message.

### DIFF
--- a/netbox/dcim/models/devices.py
+++ b/netbox/dcim/models/devices.py
@@ -946,7 +946,7 @@ class Device(
         if self.cluster and self.cluster._location is not None and self.cluster._location != self.location:
             raise ValidationError({
                 'cluster': _("The assigned cluster belongs to a different location ({location})").format(
-                    site=self.cluster._location
+                    location=self.cluster._location
                 )
             })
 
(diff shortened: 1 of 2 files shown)


pheus:

arthanson (approved):

PR #22728, opened 2026-07-21
Closes #22595: Introduce `BULK_UPDATE_CHUNK_SIZE` to limit max number of rows per bulk update

### LINKED ISSUE #22595: Implement a configuration setting to chunk `update()` calls defensively
### NetBox Version

v4.6.4

### Python Version

3.14

### Area(s) of Concern

- [ ] User Interface
- [ ] REST API
- [ ] GraphQL API
- [X] Python ORM
- [ ] Other

### Observations

There are several places within NetBox where calls to the `update()` ORM method may result in excessively long query times when many millions of objects are present in a SQL table. (netbox-community/netbox#22526 is one such example we encountered recently.) Queries that exceed certain thresholds will fail, in which case it may become impossible to perform certain operations without adjusting the database configuration.

### Proposed Changes

The solution for netbox-community/netbox#22526 was to chunk `update()` calls when applying bulk updates to custom field data. Although this does not significantly improve overall performance, it mitigates the risk that any individual bulk update query will violate the maximum time permitted by PostgreSQL. Although this solution was specific to custom field data, it can be applied more generally across the application.

A new configuration parameter (e.g. `BULK_UPDATE_CHUNK_SIZE`) is exposed. It default value is None, retaining the current behavior. Setting it to an integer value will limit the maximum number of rows affected per query. Some rough code stolen from netbox-community/netbox#22556:

```python
with transaction.atomic():
    last_pk = 0
    while True:
        pks = list(
            model.objects.filter(pk__gt=last_pk).order_by('pk')
            .values_list('pk', flat=True)[:CHUNK_SIZE]
        )
        if not pks:
            break
        model.objects.filter(pk__in=pks).update(**data)
        last_pk = pks[-1]
```

I see two options for adopting the chunked approach:

* Introduce a utility function to be invoked instead of calling `update()` on a queryset directly (more boilerplate, safer)
* Override Django's stock `update()` method (more concise, somewhat risky)

We'll also probably want to sanity-check the configured value, perhaps enforcing a range of sane values. (`5000` was identified as a reasonably ideal row count in netbox-community/netbox#22556.)

jeremystretch:
### Closes: #22595

- Introduce the `BULK_UPDATE_CHUNK_SIZE` configuration parameter (default: `None`)
- Introduce the `chunked_update()` utility function to enforce the chunk size on queryset calls to `update()`
- Wrap relevant calls to `update()` with `chunked_update()` (exempting migrations & tests)
- Pass `BULK_UPDATE_CHUNK_SIZE` as `batch_size` for certain calls to `bulk_update()` as appropriate

### DIFF
--- a/netbox/dcim/models/cables.py
+++ b/netbox/dcim/models/cables.py
@@ -25,7 +25,7 @@ from netbox.models import ChangeLoggedModel, PrimaryModel
 from utilities.conversion import to_meters
 from utilities.exceptions import AbortRequest
 from utilities.fields import ColorField, GenericArrayForeignKey
-from utilities.querysets import RestrictedQuerySet
+from utilities.querysets import RestrictedQuerySet, chunked_update
 from utilities.serialization import deserialize_object, serialize_object
 from wireless.models import WirelessLink
 
@@ -783,7 +783,7 @@ class CablePath(models.Model):
         # Record a direct reference to this CablePath on its originating object(s)
         origin_model = self.origin_type.model_class()
         origin_ids = [decompile_path_node(node)[1] for node in self.path[0]]
-        origin_model.objects.filter(pk__in=origin_ids).update(_path=self.pk)
+        chunked_update(origin_model.objects.filter(pk__in=origin_ids), _path=self.pk)
 
     def delete(self, *args, **kwargs):
         # Mirror save() - clear _path on origins to prevent stale references
@@ -791,7 +791,7 @@ class CablePath(models.Model):
         if self.path:
             origin_model = self.origin_type.model_class()
             origin_ids = [decompile_path_node(node)[1] for node in self.path[0]]
-            origin_model.objects.filter(pk__in=origin_ids, _path=self.pk).update(_path=None)
+            chunked_update(origin_model.objects.filter(pk__in=origin_ids, _path=self.pk), _path=None)
 
         super().delete(*args, **kwargs)
 
--- a/netbox/utilities/querysets.py
+++ b/netbox/utilities/querysets.py
@@ -1,4 +1,6 @@
-from django.db.models import Prefetch, QuerySet
+from django.conf import settings
+from django.db import router, transaction
+from django.db.models import Max, Prefetch, QuerySet
 
 from users.constants import CONSTRAINT_TOKEN_USER
 from utilities.permissions import get_permission_for_model, permission_is_exempt, qs_filter_from_constraints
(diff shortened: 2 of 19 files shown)


jeremystretch:

jeremystretch:
1. I was on the fence about this originally, but setting a default chunk size of e.g. 5000 is probably reasonable.
2. Added some simple validation.

netbox/dcim/models/cables.py
```diff
@@ -783,15 +783,15 @@ def save(self, *args, **kwargs):
...
+        chunked_update(origin_model.objects.filter(pk__in=origin_ids), _path=self.pk)
 
     def delete(self, *args, **kwargs):
         # Mirror save() - clear _path on origins to prevent stale references
         # in table views that render _path.destinations
         if self.path:
             origin_model = self.origin_type.model_class()
             origin_ids = [decompile_path_node(node)[1] for node in self.path[0]]
-            origin_model.objects.filter(pk__in=origin_ids, _path=self.pk).update(_path=None)
+            chunked_update(origin_model.objects.filter(pk__in=origin_ids, _path=self.pk), _path=None)
```
pheus, line 794: Could this update through `queryset.filter(pk__in=pks)` instead? Using `_default_manager` drops the queryset's other filters. In `CablePath.delete()`, that means a concurrent `_path` change could still be cleared.

netbox/utilities/querysets.py
```diff
@@ -6,9 +8,48 @@
...
+    """
+    if chunk_size is None:
+        chunk_size = settings.BULK_UPDATE_CHUNK_SIZE
+    if not chunk_size:
+        return queryset.update(**kwargs)
+
+    model = queryset.model
+    count = 0
+    last_pk = 0
+    with transaction.atomic():
```
pheus: Could we pin this operation to a single write database? Right now `atomic()` uses `default`, the PK lookup can use a read replica, and rebuilding the update from `_default_manager` drops an explicit `.using()`. Using the same alias for the lookup, updates, and transaction would preserve the queryset's database routing and keep all batches atomic.

netbox/utilities/querysets.py
```diff
@@ -6,9 +8,48 @@
...
+        while True:
+            pks = list(
+                queryset.filter(pk__gt=last_pk).order_by('pk').values_list('pk', flat=True)[:chunk_size]
```
pheus: Could we capture the initial maximum matching PK and use it as an upper bound? Later batches can otherwise pick up newly inserted rows, so a busy table may keep extending the operation.

netbox/utilities/querysets.py
```diff
@@ -6,9 +8,48 @@
...
+    When that is also None (the default), a single unbounded UPDATE is issued, identical to calling
+    queryset.update(**kwargs) directly.
+
+    :param queryset: The QuerySet identifying the rows to update
+    :param chunk_size: The maximum number of rows to update per statement (defaults to
+        settings.BULK_UPDATE_CHUNK_SIZE)
+    """
+    if chunk_size is None:
+        chunk_size = settings.BULK_UPDATE_CHUNK_SIZE
+    if not chunk_size:
```
pheus: The setting is validated at startup, but direct callers can still pass `0`, a negative value, or a non-integer here. Could the helper validate `chunk_size` as a positive integer or `None` too?

pheus (changes requested):
Thanks for putting this together! I think the helper is a useful improvement, but I’m requesting changes for two correctness concerns:

**Blocking**
- Preserve the original queryset filters in each `UPDATE`; rebuilding the query from `_default_manager` can overwrite concurrent changes that no longer match those filters.
- Use one resolved write database for the PK lookup, updates, and `atomic()` block so `.using()` and database routers remain safe and the full operation is genuinely atomic.

**Non-blocking**
- Consider capturing an initial maximum PK so newly inserted rows cannot keep extending the operation.
- Validate directly supplied `chunk_size` values as a positive integer or `None`.
- Update the docstring to reflect the new default of `5000`.

The latter items can be handled separately, but I believe the first two should be addressed before merging. Thanks again!

netbox/utilities/querysets.py
jeremystretch: Good catch. I was operating under the assumption that the atomic transaction would obviate the need for that, but  that's not true for `READ COMMITTED`.

netbox/dcim/models/cables.py
jeremystretch: Not sure I follow the proposal. Is this still needed considering the other changes that have been made?

pheus: You’re right, this is no longer needed. My comment was based on the earlier implementation that rebuilt the update from `_default_manager`.

pheus (changes requested):
Thanks for the update!

One blocker remains in `ConfigRevision.activate()`: `atomic()` still uses the default database, while `chunked_update()` may use the routed write database. Could we resolve the write alias once and use it for the transaction and both querysets?

Once that’s addressed, this looks good to me.

pheus (approved):
Thanks for the updates! This looks good to me now!

PR #22730, opened 2026-07-21
Fixes #22729: Escape names of file attachments in HTTP responses

### LINKED ISSUE #22729: Content-Disposition filename injection via unescaped object names
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.5

### Python Version

3.14

### Steps to Reproduce

1. Create an object whose name is used to build an export filename — for example an Export Template, or the source object of a config-render / report-output view — and set its **name** to a value containing a double quote and an extra header directive, e.g.:
   ```
   report"; filename="example.txt
   ```
2. As an authenticated user with the appropriate `view_<model>` permission, request the direct export of the rendered content. For a rendered config this is the object's render-config view with `?export=1`; for a report/script it is the job results view with `?export=output`.
3. Inspect the `Content-Disposition` response header returned by NetBox.

### Expected Behavior

The object name should be safely encoded into the `Content-Disposition` header. The embedded double quote should be escaped (or the filename encoded per RFC 5987, e.g. `filename*=UTF-8''<urlencoded>`), so that the entire name is treated as a single opaque filename value and no additional header directives can be introduced.

### Observed Behavior

The object name is interpolated directly into the header via an f-string with no quoting or escaping.

jeremystretch:
### Fixes: #22729

- Introduce the `add_file_attachment()` utility function
- Update all instances where the `Content-Disposition` header is added to an HTTP response
- Add tests

### DIFF
--- a/netbox/core/views.py
+++ b/netbox/core/views.py
@@ -12,6 +12,7 @@ from django.db import DatabaseError, connection
 from django.http import Http404, HttpResponse, HttpResponseForbidden
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
+from django.utils.http import content_disposition_header
 from django.utils.translation import gettext_lazy as _
 from django.views.generic import View
 from django_rq.queues import get_queue_by_index, get_redis_connection
@@ -761,7 +762,7 @@ class SystemView(UserPassesTestMixin, View):
                 },
             }
             response = HttpResponse(json.dumps(data, cls=ConfigJSONEncoder, indent=4), content_type='text/json')
-            response['Content-Disposition'] = 'attachment; filename="netbox.json"'
+            response['Content-Disposition'] = content_disposition_header(as_attachment=True, filename='netbox.json')
             return response
 
         # Serialize any JSON-based classes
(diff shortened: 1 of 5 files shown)


jeremystretch:

jeremystretch:

jeremystretch:

jnovinger (dismissed):
A couple of things I noticed, neither blocking:

1. The non-streaming export path in `bulk_views.py` still builds the header the old way. It's currently safe only because that `filename` is `verbose_name_plural`-derived and no caller passes user data.

2. Did you consider using `django.utils.http.content_disposition_header`? Its escaping logic is pretty similar to yours.

jeremystretch:

jeremystretch:
> A couple of things I noticed, neither blocking:
> 
> 1. The non-streaming export path in `bulk_views.py` still builds the header the old way. It's currently safe only because that `filename` is `verbose_name_plural`-derived and no caller passes user data.
> 2. Did you consider using `django.utils.http.content_disposition_header`? Its escaping logic is pretty similar to yours.

1. Yep, ok with this.
2. I've adapted the PR to use this, thanks.

jnovinger (approved):

Issue #22640, opened 2026-07-08
Enforce `ALLOWED_URL_SCHEMES` for URLs in custom fields


jeremystretch:
### NetBox version

v4.6.4

### Feature type

Change to existing functionality

### Proposed functionality

Restrict custom field URLs to the schemes listed in [`ALLOWED_URL_SCHEMES`](https://netboxlabs.com/docs/netbox/configuration/security/#allowed_url_schemes). The whitelist of permitted schemes would be applied in addition to any validation regex associated with the field.

### Use case

It may be desirable for `ALLOWED_URL_SCHEMES` to be enforced for custom fields as well. However, this is a potentially breaking change for many users, and should be considered carefully.

### Database changes

N/A

### External dependencies

N/A

PR #22732, opened 2026-07-21
#22640 - Enforce ALLOWED_URL_SCHEMES for URLs in custom fields

arthanson:
### Closes: #22640

Enforce `ALLOWED_URL_SCHEMES` for URL custom fields at two layers:

- **Entry**: `CustomField.validate()` runs `EnhancedURLValidator`. The UI and REST API now have the same validation.
- **Render**: the `customfield_value` tag and `CustomFieldColumn` only emit a clickable `<a href>` for permitted schemes, otherwise showing the value as plain text.

**Note:** Previously REST and UI allowed different URLs, REST was more permissive.

### Notes

This is a breaking change: URL custom field values using a scheme outside `ALLOWED_URL_SCHEMES` are now rejected on save.

### DIFF
--- a/netbox/extras/models/customfields.py
+++ b/netbox/extras/models/customfields.py
@@ -46,7 +46,7 @@ from utilities.forms.widgets import APISelect, APISelectMultiple, DatePicker, Da
 from utilities.jsonschema import validate_schema
 from utilities.querysets import RestrictedQuerySet
 from utilities.templatetags.builtins.filters import render_markdown
-from utilities.validators import EnhancedURLValidator, validate_regex
+from utilities.validators import url_scheme_is_allowed, validate_regex
 
 __all__ = (
     'CustomField',
@@ -796,9 +796,14 @@ class CustomField(CloningMixin, ExportTemplatesMixin, OwnerMixin, ChangeLoggedMo
             elif self.type == CustomFieldTypeChoices.TYPE_URL:
                 if type(value) is not str:
                     raise ValidationError(_("Value must be a string."))
-                # Validate the URL and its scheme against ALLOWED_URL_SCHEMES, as the UI does (LaxURLField).
-                # Instantiated per call so a runtime change to the (dynamic) setting is honored.
-                EnhancedURLValidator()(value)
+                # Enforce ALLOWED_URL_SCHEMES to guard against dangerous schemes (e.g. javascript:), using
+                # the same check applied when rendering the value. The UI and REST API both normalize a
+                # schemeless value to an absolute URL (LaxURLField with assume_scheme='https'); a schemeless
+                # value reaching this point directly (e.g. via a script) is permitted and treated as relative.
+                if not url_scheme_is_allowed(value):
+                    raise ValidationError(
+                        _("URLs must use a scheme permitted by ALLOWED_URL_SCHEMES.")
+                    )
                 if self.validation_regex and not re.match(self.validation_regex, value):
                     raise ValidationError(_("Value must match regex '{regex}'").format(regex=self.validation_regex))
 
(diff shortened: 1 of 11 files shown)


arthanson:

arthanson:
per claude review:

1. makes UI and REST consistent. They shouldn't allow different values.
2. fixed - ALLOWED_URL_SCHEMES is admin editable so shouldn't be cached.

netbox/utilities/validators.py
```diff
@@ -72,6 +75,24 @@ def __call__(self, value):
             )
 
 
+def get_url_scheme(value):
```
jeremystretch: Suggest collapsing this into `is_url_scheme_allowed()` if there's nowhere else we intend to use it right now.

netbox/utilities/validators.py
```diff
@@ -72,6 +75,24 @@ def __call__(self, value):
...
+    """
+    Return the (lower-cased) scheme of a URL, or an empty string if it has none. A percent-encoded
+    scheme (e.g. "javascript%3A…") yields no scheme, matching browser behavior: a browser does not
+    decode the scheme portion of an href, so such a value is inert and is treated as relative.
+    """
+    return urlparse(value).scheme.lower()
+
+
+def is_url_scheme_allowed(value):
```
jeremystretch: Suggest renaming this to `url_scheme_is_allowed()` for better readability.

netbox/utilities/validators.py
```diff
@@ -72,6 +75,24 @@ def __call__(self, value):
             )
 
 
+def get_url_scheme(value):
+    """
+    Return the (lower-cased) scheme of a URL, or an empty string if it has none. A percent-encoded
+    scheme (e.g. "javascript%3A…") yields no scheme, matching browser behavior: a browser does not
+    decode the scheme portion of an href, so such a value is inert and is treated as relative.
+    """
+    return urlparse(value).scheme.lower()
```
jeremystretch: We probably need to catch `ValueError` exceptions from `urlparse()`, which can result from invalid URLs (e.g. `urlparse('http://[::1/foo')`).

netbox/extras/models/customfields.py
```diff
@@ -796,6 +796,9 @@ def validate(self, value):
             elif self.type == CustomFieldTypeChoices.TYPE_URL:
                 if type(value) is not str:
                     raise ValidationError(_("Value must be a string."))
+                # Validate the URL and its scheme against ALLOWED_URL_SCHEMES, as the UI does (LaxURLField).
+                # Instantiated per call so a runtime change to the (dynamic) setting is honored.
```
jeremystretch: Claude flagged a subtle distinction re: validation. I'm not sure how big a concern this is in practice, but wanted to raise it. I would _not_ recommend removing the `assume_scheme` behavior from our usage of LaxURLField here just to achieve parity between the UI and API.

---

The UI's `LaxURLField` is instantiated with `assume_scheme='https'`, so a schemeless value is normalized *before* validation:
```
URLField(assume_scheme='https').to_python('example.com') -> 'https://example.com'
```
The new REST path calls `EnhancedURLValidator()(value)` on the **raw** value, and that validator's regex requires `://`. Net result:
- UI: user enters `example.com` → stored as `https://example.com` (accepted)
- REST: client sends `{"url_field": "example.com"}` → **400 rejected**

So the layers still diverge, contradicting the PR body's claim that "the UI and REST API now have the same validation." API clients migrating existing bare-hostname data will hit this. If parity is the goal, consider normalizing the scheme in `validate()` the same way (or documenting that the API requires an explicit scheme). The new `test_url_scheme_validation` doesn't cover the schemeless case, which is why the divergence isn't surfaced.

jeremystretch (changes requested):
Nice work. Have some minor suggestions. The only real concern is about catching a potential ValueError from `urlparse()`.

jeremystretch (approved):

PR #22733, opened 2026-07-21
22677 - Fix cable form validation errors

### LINKED ISSUE #22677: Form validation errors are silently dropped for fields without HTML5 constraints
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.5 (post netbox-community/netbox#22662)

### Python Version

3.12

### Steps to Reproduce

1. Edit a cable (e.g. `/dcim/cables/1/edit/`).
2. Set Length to `1234000` and Length unit to Kilometers
3. Click Save.

### Expected Behavior

The form redisplays with a validation error on the Length field, e.g. "Ensure that there are no more than 6 digits before the decimal point."

### Observed Behavior

The form redisplays with the entered value, no error message, and no indication of why it wasn't saved. The value is rejected server-side (the record is not updated), but the error is never shown to the user.

arthanson:
### Closes: #22677

The Cable HTMX form template (dcim/htmx/cable_edit.html) renders length and length_unit manually. Fix was to inline error display for length and length_unit, matching the render_field.html pattern. The Cable form was the only one that looked like it had this issue.

### DIFF
--- a/netbox/templates/dcim/htmx/cable_edit.html
+++ b/netbox/templates/dcim/htmx/cable_edit.html
@@ -59,13 +59,23 @@
   {% render_field form.label %}
   {% render_field form.description %}
   {% render_field form.color %}
-  <div class="row mb-3">
+  <div class="row mb-3{% if form.length.errors or form.length_unit.errors %} has-errors{% endif %}">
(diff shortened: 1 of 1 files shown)


arthanson:

arthanson:
claude issue fixed.

pheus (approved):
Thanks!

Issue #22486, opened 2026-06-18
Configurable timeout for webhooks


bctiemann:
### NetBox version

v4.6.3

### Feature type

New functionality

### Proposed functionality

Feature request for a configurable timeout parameter for webhooks.

### Use case

Without the timeout, a webhook request (with a target URL thats timing out) will hold open an RQ worker for 5 minutes (or whatever the RQ job timeout is), potentially blocking other tasks, subject to worker count.

When this happens we see a log message like this emitted:

```
2026-06-18 10:01:35,972 rq.worker ERROR Worker 101f836f12624c70a808b3054d767e8b: job 4f4f4ad1-8d9b-4ab8-ad5d-df7aefc77f5f: exception raised while executing (extras.webhooks.send_webhook)
Traceback (most recent call last):
  File "/opt/netbox/venv/lib/python3.12/site-packages/rq/worker/base.py", line 1522, in perform_job
    return_value = job.perform()
                   ^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/rq/job.py", line 1342, in perform
    self._result = self._execute()
                   ^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/rq/job.py", line 1402, in _execute
    result = self.func(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/netbox/extras/webhooks.py", line 122, in send_webhook
    response = session.send(prepared_request, proxies=proxies)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/requests/sessions.py", line 706, in send
    r = adapter.send(request, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/requests/adapters.py", line 645, in send
    resp = conn.urlopen(
           ^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 788, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 464, in _make_request
[...]

PR #22734, opened 2026-07-21
22486 - Add Configurable timeout for webhooks

arthanson:
### Closes: #22486 

This adds a configurable timeout at two levels:

- WEBHOOK_DEFAULT_TIMEOUT — new global config parameter (default 30s) applied to all webhooks.
- Per-webhook timeout — optional field on the Webhook model that overrides the global default when set.

**Note:** using `https://httpbin.org/delay/10` as the web hook URL is very nice for testing the timeout (this delays for 10 seconds) do a GET web hook request to it.

send_webhook() now passes the resolved timeout to the outgoing request (per-webhook value if set, otherwise the global default).

## Changes

- New WEBHOOK_DEFAULT_TIMEOUT setting + docs
- New nullable Webhook.timeout field (validated 1–3600s) wired through serializer, forms, filterset, table, detail panel, and GraphQL
- Timeout applied in extras/webhooks.py
- Tests for global default and per-webhook override


arthanson:

arthanson:
both claude review issues fixed.

netbox/extras/migrations/0142_webhook_timeout.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 6.0.7 on 2026-07-21 23:28
```
bctiemann: I've been removing these autogenerated headers on migrations; probably a good practice even if we've been inconsistent about it in the past.

docs/configuration/miscellaneous.md
```diff
@@ -306,3 +306,11 @@ The base unit for disk sizes. Set this to `1024` to use binary prefixes (MiB, Gi
 Default: `1000`
 
 The base unit for RAM sizes. Set this to `1024` to use binary prefixes (MiB, GiB, etc.) instead of decimal prefixes (MB, GB, etc.).
+
+---
+
+## WEBHOOK_DEFAULT_TIMEOUT
```
bctiemann, line 312: I assume it's not worth it to call it `WEBHOOK_DEFAULT_TIMEOUT_SECONDS` or similar? I see our precedent has a lot of `TIMEOUT` variables where seconds is presumed, so no issue here.

bctiemann (changes requested):
Just a small comment cleanup.

bctiemann (approved):

PR #22735, opened 2026-07-22
Closes #22685 - Add "has any of these tags" filter mode

### LINKED ISSUE #22685: Allow "has any of these tags" during search filtering
### NetBox version

4.6.4

### Feature type

New functionality

### Proposed functionality

When searching in netbox (for instance for an IP address) the user can filter for tags using the following two functions:

- Has these tags
- Does not have these tags

This gives users the ability to find items with ALL the following tags, or NONE of the following tags, but neglects to find items with ANY of the following tags. 

I suggest adding "Has any of these tags" to the search filter options. 

### Use case

This would allow users to find items in netbox that have one or more of a subset of tags, which is useful during compliance work or when trying to find information when you are not sure how the item was tagged originally. 

### Database changes

Not sure

### External dependencies

I don't think so

bctiemann:
## Summary

- Adds a `tag__any` / `tag_id__any` filter lookup expression that matches objects tagged with **any** of the specified tags (logical OR), complementing the existing default (AND / `conjoined=True`) and `n` (negation) modes.
- Exposed in the UI filter builder as a new "has any of these tags" option alongside "has these tags" / "does not have these tags".

Screenshot 2026-07-21 at 9 55 11 PM

Example:
```
http://127.0.0.1:8000/api/dcim/devices/?tag__any=alpha&tag__any=bravo
```

## Design notes (per discussion on the issue)

`TagFilter`/`TagIDFilter` hard-code `conjoined=True` in `__init__`, and NetBox's dynamic lookup-expression framework (`BaseFilterSet.get_additional_lookups`) only varies `lookup_expr` and toggles `.exclude` for negation — it has no mechanism to flip `.conjoined`. So this couldn't be a simple lookup swap:

- Added a dedicated `FILTER_TAG_LOOKUP_MAP` (`utilities/constants.py`) rather than reusing the shared `FILTER_NEGATION_LOOKUP_MAP` (which also serves plain FK/M2M filters that are already OR-by-default and don't need an "any" mode).
- `_get_filter_lookup_dict` routes `TagFilter`/`TagIDFilter` to this new map.
- `get_additional_lookups` sets `new_filter.conjoined = False` when constructing the `any` variant — the missing plumbing.
- Scoped narrowly to the positive "has any" case, not expanded into a full any/all/negation matrix.

## Test plan

- [x] Extended `test_tag_filter` (`utilities/tests/test_filters.py`) to assert `.conjoined` on the `exact`, `n`, and new `any` filter variants.
- [x] Added `TagFilterTestCase`, exercising real `Site` objects through `SiteFilterSet` to verify AND (`tag`), OR (`tag__any`), and NOR (`tag__n`) queryset behavior end-to-end — there was previously no behavioral test for `TagFilter`'s AND semantics at all.
- [x] Verified both new tests fail without the fix (`KeyError` on `tagfield__any`; `tag__any` param silently ignored, returning all objects) and pass with it.
- [x] Ran the broader `extras.tests.test_filtersets` + `dcim.tests.test_filtersets` suites (939 tests) to confirm no regression in existing tag AND/negation behavior across taggable models.
- [x] `ruff check` passes; pre-commit hooks (lint, Django checks, docs build) pass.
- [x] Documented the new lookup expressions in `docs/reference/filtering.md`.

### DIFF
--- a/docs/reference/filtering.md
+++ b/docs/reference/filtering.md
@@ -110,6 +110,21 @@ expression: `n`. Here is an example of a lookup expression on a foreign key, it 
 GET /api/ipam/vlans/?group_id__n=3203
 ```
 
+### Tags
+
+The `tag` and `tag_id` filters support negation (`n`) as well as an `any` lookup expression:
+
+| Filter | Description                                       |
+|--------|----------------------------------------------------|
+| `n`    | Does not have any of these tags                    |
+| `any`  | Has any of these tags (logical OR)                  |
+
+Passing multiple values for `tag`/`tag_id` without a lookup expression uses a logical AND: `GET /api/dcim/sites/?tag=foo&tag=bar` returns only sites tagged with both `foo` _and_ `bar`. To instead match sites tagged with `foo` _or_ `bar`, use the `any` lookup expression:
+
+```no-highlight
+GET /api/dcim/sites/?tag__any=foo&tag__any=bar
+```
+
 ## Ordering Objects
 
 To order results by a particular field, include the `ordering` query parameter. For example, order the list of sites according to their facility values:
--- a/docs/reference/filtering.md
+++ b/docs/reference/filtering.md
@@ -125,6 +125,8 @@ Passing multiple values for `tag`/`tag_id` without a lookup expression uses a lo
 GET /api/dcim/sites/?tag__any=foo&tag__any=bar
 ```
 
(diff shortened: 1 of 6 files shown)


bctiemann:

bctiemann:
Addressed all three concerns.

jeremystretch (approved):
Nice solution!

PR #22741, opened 2026-07-22
Fixes #22697: Return to Scripts list when canceling ScriptModule creation

### LINKED ISSUE #22697: Using Cancel button in Add Script screen returns to root page
### NetBox Edition

NetBox Community

### NetBox Version

4.6.5

### Python Version

3.13

### Steps to Reproduce

1. Go to Customization -> Scripts
2. Click `+Add` button
3. Click `Cancel` button

### Expected Behavior

Browser should return to http://netbox.example/extras/scripts/

Several other things that I have tested typically return to the page where the +Add button would normally be found.   Probably everything does that.

### Observed Behavior

Browser returns to http://netbox.example/

pheus:
### Closes: netbox-community/netbox#22697

Updates the ScriptModule add view to use the Scripts list as its default return URL when no explicit return URL is provided. This ensures that cancelling the form returns the user to the expected page instead of the NetBox home page.

Adds a regression test covering the default return URL.

### DIFF
--- a/netbox/extras/tests/test_views.py
+++ b/netbox/extras/tests/test_views.py
@@ -1148,6 +1148,20 @@ class ScriptListViewTestCase(TestCase):
         self.assertTemplateUsed(response, 'extras/inc/script_list_content.html')
 
 
+class ScriptModuleCreateViewTestCase(TestCase):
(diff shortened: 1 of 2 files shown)


pheus:

bctiemann (approved):

PR #22744, opened 2026-07-22
Closes #22485: Break the search signal wiring  import cycle

### LINKED ISSUE #22485: Move netbox.search signal wiring into AppConfig.ready() and break the backends/deferred/tasks import cycle
### Proposed Changes

The `netbox.search` subsystem connects its `post_save`/`post_delete` receivers as an import side effect, and carries an import cycle that's currently held together with a function-local import. Two coupled cleanups:

1. **Make signal connection deterministic.** Today the receivers are wired at the bottom of `netbox/search/backends.py`:

   ```python
   search_backend = get_backend()
   
   post_save.connect(search_backend.caching_handler)
   post_delete.connect(search_backend.removal_handler)
   ```

   These `.connect()` calls run whenever `backends.py` is first imported. There's no `AppConfig.ready()` doing the wiring, and `netbox` is the project package rather than a Django app, so it has no native `ready()`. Connection therefore fires at whatever point some consumer first imports `backends.py` (`netbox/forms/search.py`, `netbox/views/misc.py`, `extras/management/commands/reindex.py`, `netbox/search/tasks.py`, `core/jobs.py`), which is import-order dependent rather than fixed at startup.

   Move the receivers (`caching_handler`, `removal_handler`) and the `.connect()` calls into a dedicated `netbox/search/signals.py`, wired from an appropriate `AppConfig.ready()` (e.g. `CoreConfig`). This matches the 9 apps that already keep their handlers in `signals.py`.
2. **Resolve the layering inversion / import cycle.** The receivers live as methods on the low-level `SearchBackend` (`backends.py`), but as of netbox-community/netbox#21326 they call up into the deferral orchestration: `netbox/search/deferred.py` `mark_for_deferred_indexing` dispatches to `netbox/search/tasks.py`, which imports the `search_backend` singleton back from `backends.py`. That `backends -> deferred -> tasks -> backends` cycle is currently worked around with a function-local import in `deferred._flush` (`netbox/search/deferred.py`, documented in the comment pointing here).

   Likely extract the `search_backend` singleton into its own small module so `tasks.py` no longer imports back into `backends.py`. With the receivers out of `backends.py` as well, the cycle dissolves and the remaining function-local import can move back to module level.

Net effect: signal connection is deterministic at startup, the subsystem gets a normal `signals.py`, and the import cycle goes away.

### Justification

Noticed during review of netbox-community/netbox#21326 (PR netbox-community/netbox#22481, deferring global search cache updates). The performance change left the receivers calling up into the deferral layer, which exposed both the import-order-dependent signal wiring and the `backends -> deferred -> tasks -> backends` cycle.

This is a deliberate follow-up, not urgent. Moving signal wiring into an `AppConfig.ready()` shifts import and execution ordering at init time, with a blast radius too large to fold into a performance PR. It should be done on its own, with full test-suite verification.

bctiemann:
## Summary

Two coupled cleanups to the `netbox.search` subsystem, per the issue's own breakdown:

**1. Deterministic signal wiring.** `post_save`/`post_delete` were connected as a side effect of `netbox/search/backends.py` being imported for the first time, whichever of this subsystem's several consumers (`netbox.forms.search`, `netbox.views.misc`, `extras.management.commands.reindex`, `netbox.search.jobs`, `core.jobs`, `dcim.signals`) happened to trigger that. Moved the `.connect()` calls into a new `netbox/search/signals.py`, wired explicitly from `CoreConfig.ready()` — matching the convention already used by the 9 apps that keep their handlers in `signals.py`.

**2. Resolved the `backends -> deferred -> jobs -> backends` import cycle.** The actual cycle came from `backends.py`'s own module-level `from .deferred import ...` (needed by `CachedValueSearchBackend.caching_handler`/`removal_handler`) executing *before* `search_backend` gets defined further down that same file, while `deferred.py`/`jobs.py` both need that singleton at their own module level. Scoping the `deferred.py` import to just the two methods that need it (rather than importing it at module level) breaks the cycle: nothing forces `deferred.py` to load during `backends.py`'s own module-level execution anymore, so `deferred.py`'s own imports of `search_backend`/`SearchCacheJob` can move back to module level, removing the previous workaround (and its comment pointing at this issue).

`get_backend()`/`search_backend` stay defined in `backends.py` exactly where they always were — an earlier version of this PR extracted them into a new module, but that turned out to be unnecessary (the function-local import above is sufficient on its own) and would have forced an import-path compatibility shim on at least one plugin that imports `get_backend` directly from `netbox.search.backends`. No reason to move a symbol for no benefit.

## Files changed

- `netbox/search/signals.py` (**new**) — `@receiver`-decorated `caching_handler`/`removal_handler` wrappers connecting `post_save`/`post_delete`, delegating to whichever backend is configured.
- `netbox/search/backends.py` — removed the `.connect()` calls (moved to `signals.py`); scoped `CachedValueSearchBackend`'s `mark_for_deferred_indexing` import to the two methods that need it, with an updated comment explaining why.
- `netbox/search/deferred.py` — moved the two previously function-local imports (`search_backend`, `SearchCacheJob`) to module level; removed the stale workaround comment.
- `core/apps.py` — `CoreConfig.ready()` now imports `netbox.search.signals`, connecting the receivers deterministically at startup.
- ~~`netbox/tests/test_search.py` — added `SearchModuleImportCycleTestCase`, a regression test that clears the three affected modules from `sys.modules` and re-imports them fresh to catch a reintroduced cycle for real (verified it fails with the expected `ImportError` against the pre-fix import graph, and passes against the fix); updated two stale comments referencing the old "connected at import" behavior.~~ Removed at Jason's request.

No behavior change: this is a pure internal refactor. The public `SearchBackend` contract (`cache()`/`remove()`/`clear()`/`search()`/`count()`/`size`) and the `caching_handler`/`removal_handler` override point for custom backends are unchanged. `get_backend`/`search_backend` remain importable from `netbox.search.backends` exactly as before — no import path changes anywhere outside the `netbox.search` package itself.

## Test plan

- [x] Added `SearchModuleImportCycleTestCase` and verified it fails with the exact `ImportError: cannot import name 'search_backend' from partially initialized module...` against the pre-fix import graph, and passes with the fix.
- [x] Ran `netbox.tests.test_search`, `core.tests.test_jobs`, `extras.tests.test_management_commands`, `dcim.tests.test_signals`, `netbox.tests.test_views` (120 tests) — only the 3 pre-existing, unrelated `SearchViewTestCase` failures remain (a `querystring_replace` template tag `TemplateSyntaxError` in `table_htmx.html`, confirmed identical against unmodified `feature` with a byte-for-byte diff of failure names).
- [x] Ran the broader `core` + `netbox` app suites (794 tests) and confirmed the failure set is byte-for-byte identical to unmodified `feature` (26 errors, all the same pre-existing `table_htmx.html` issue affecting any list-view test).
- [x] `ruff check` passes on all changed files.
- [x] `manage.py check` passes.


bctiemann:

netbox/netbox/tests/test_search.py
```diff
@@ -757,10 +760,50 @@ def test_custom_backend_indexes_synchronously(self):
...
+        # reaches the SAME caching_handler call path -- but its override defers instead of indexing
+        # inline. This contrasts with the custom backend above: identical dispatch, polymorphic
+        # behavior.
         with transaction.atomic():
             Site.objects.create(name='Default Defers', slug='default-defers')
             scheduled = scheduled_search_flushes()
             self.assertEqual(len(scheduled), 1)
+
+
+class SearchModuleImportCycleTestCase(TestCase):
```
jnovinger: I love this as a regression-first style test while fixing the issue. But given the hard-coded module names and specifics, I don't think this is something we want to commit to maintaining long term.

I'd support a generic import-cycle test, but I also have no idea of how best to do that. I think the cleanest path forward is to just drop this test from the commit and treat it as something that was useful during work on the issue, but not long-term.

jnovinger (changes requested):

netbox/netbox/tests/test_search.py
bctiemann: Agreed and done. There are still changes to the comments but they are largely just to do with correcting the import signatures to match the change.

jnovinger (approved):

PR #22747, opened 2026-07-22
Fixes #22746: Sanitize rendered HTTP headers for outbound webhooks

### LINKED ISSUE #22746: Outbound webhook headers do not escape control characters
### NetBox Edition

NetBox Community

### NetBox Version

v4.6.5

### Python Version

3.14

### Steps to Reproduce

1. Create a **Webhook** with:
   - **Payload URL:** an endpoint you control / can observe (e.g. a request-capture service)
   - **Additional headers:**
     ```
     X-Object-Name: {{ data.name }}
     ```
   - Attach it to an **Event Rule** that fires on object creation for, say, Tags.
2. Create an object whose name embeds a newline and a forged header, e.g. a Tag named:
   ```
   legit-name
   X-Injected-Header: malicious-value
   ```
   (Submit via the REST API to reliably include the literal newline in the `name` field.)
3. Trigger the event so the webhook fires.
4. Inspect the HTTP request received at the payload URL.

### Expected Behavior

Header rendering should treat the interpolated value as a single opaque header value. A value containing CR/LF or control characters (or a resulting header name that is not a valid RFC 7230 token) should be escaped or omitted.

### Observed Behavior

The rendered output is split on line boundaries and each fragment is parsed as its own header, so the outbound request contains an `X-Injected-Header: malicious-value` header that the webhook author never defined.

jeremystretch:

### DIFF
--- a/netbox/extras/models/models.py
+++ b/netbox/extras/models/models.py
@@ -1,4 +1,5 @@
 import json
+import re
 import urllib.parse
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


jeremystretch:

jeremystretch:
> 1. Breaking change for multi-line Jinja2 templates

This is going to be a sticking point, I think. I don't see how we can effectively sanitize the rendered headers without infringing on the flexibility afforded by Jinja rendering. I see two options:

1. Sanitize raw lines individually, breaking the current behavior. Likely to introduce more harm than benefit.
2. Sanitize on the rendered output as a whole. This would require establishing a whitelist of approved HTTP headers (and is also likely to result in breakage initially).

Given the fairly niche scenario in which this bug might be exploited, it may be reasonable to simply note the concern for webhook authors in the UI and provide guidance for constructing hardened templates. For example, the webhook header template cited in #22746 can be improved to mitigate this concern:

```
X-Object-Name: {{ data.name | replace('\n', '') }}
```

It might make sense to introduce a custom Jinja filter specifically for sanitizing HTTP header values.

jeremystretch:
As noted [above](https://github.com/netbox-community/netbox/pull/22747#issuecomment-5058137255), it's not feasible to address this holistically. I've opened FR #22753 to enable webhook template authors to conveniently validate header values directly.

